PR #1525, opened 2017-08-18
Remove unused decodeOctalEscaped

wingo:
Besides being unused, this function has a bug that it will incorrectly
decode the path component Ubuntu\04016.04.2\040LTS\040amd64 as
"Ubuntu.04.2 LTS amd64" instead of "Ubuntu 16.04.2 LTS amd64".

Found because Guix's copy of nix-daemon still used this function.


edolstra:
Thanks Andy!

Issue #1299, opened 2017-04-04
Backport fetchTarball hash


FRidh:
Could we have 06bbfb6004942bfcddd930e746ee7a2bfe5c3872 in 1.11.x?

taktoa:
Seconded; we could really use this here at Awake Security.

taktoa:
In the mean time, you can use

```nix
import <nix/fetchurl.nix> {
  url = "https://example.com/foo.tar.gz";
  sha256 = "0hk0zxxibl7r62nmzaf4qbdrqack5xs7skcys82bqi8glf2qjr7z";
  system = builtins.currentSystem;
  executable = false;
  unpack = true;
}
```

FRidh:
Interesting. I knew `fetchTarball` shared code path and the only difference was the `unpack` parameter, but I didn't know you could just import it.

We could create a `fetchTarball` function in Nixpkgs that utilizes this until Nix 1.12 is stable. @edolstra, is there any reason against doing that?

edolstra:
@FRidh 06bbfb6 does not cherry-pick straightforwardly on the 1.11 branch, so I'd prefer not to, sorry.

`builtins.fetchTarball` and `<nix/fetchurl.nix>` are very different in that the former doesn't rely on import-from-derivation. Import-from-derivation should generally be avoided in Nixpkgs because it causes builds at evaluation time (which fails in read-only mode, e.g. in `nix-env -qa`).

FRidh:
Thank you for the clarification. I'm a happy user of 1.12 so I don't *really* mind.

PR #1538, opened 2017-08-29
libexpr: add sha256 noop argument to fetch

zimbatm:
Nix 1.12 is going to support that argument fully. This addition is only
there to allow forward-compatibility.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1709,6 +1709,8 @@ void fetch(EvalState & state, const Pos & pos, Value * * args, Value & v,
             string n(attr.name);
             if (n == "url")
                 url = state.forceStringNoCtx(*attr.value, *attr.pos);
+            else if (n == "sha256")
+                // noop, added for forward-compatiblity
             else if (n == "name")
                 name = state.forceStringNoCtx(*attr.value, *attr.pos);
             else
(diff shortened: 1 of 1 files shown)


edolstra:
This seems dangerous to me. It creates a false impression that the hash is verified.

zimbatm:
I'm trying to find a solution because 1.11 is going to be around for a long time. The most ideal would to backport, but that was also rejected in #1299.

The main use-case I am trying to solve is to pin nixpkgs. Something like:
```nix
let
  repo = builtins.fromJSON (builtins.readFile ./src.json);
in
  builtins.fetchTarball {
    url = "https://github.com/${repo.owner}/${repo.repo}/tarball/${repo.rev}";
    #sha256 = repo.sha256;
  }
```

(also any solution that imports `<nixpkgs>` doesn't work when NIX_PATH is being set to that file).

But because the sha256 is missing, github.com gets hit on each invocation. There is also a workaround with the cache TTL that has it's own set of problems.

Even if it's introduced in 1.12, not everybody is going to upgrade overnight. Also the release date for 1.12 is unknown. I want a backport that can be pushed to the 17.03 and 17.09 channels so we can start use this argument, even if it's just a noop.

zimbatm:
Would it be acceptable if that argument would emit some sort of warning instead?

zimbatm:
Closing this as @taktoa's solution is a viable workaround: https://gist.github.com/taktoa/4ec8b0cb6d07857400b07a830ad99558

PR #1540, opened 2017-08-29
1.11: Replace stat / uid comparison with test -O

grahamc:
cc @matthewbauer @edolstra


domenkozar:
This fixes `stat: cannot read file system information for '%u': No such file or directory` on darwin

copumpkin:
Awesome, thanks!

PR #1547, opened 2017-09-01
manual: import needs root or trusted signature

atondwal:
`nix-store --import` needs to be either run as root, or on a
closure that's been signed by a trusted cache.

Currently this is mentioned nowhere in the manual, so this
makes it explicit

### DIFF
--- a/doc/manual/command-ref/nix-store.xml
+++ b/doc/manual/command-ref/nix-store.xml
@@ -1123,7 +1123,7 @@ $ nix-store --export $(nix-store -qR <replaceable>paths</replaceable>) > out</sc
 To import the whole closure again, run:
 
 <screen>
-$ nix-store --import &lt; out</screen>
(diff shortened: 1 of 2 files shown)


doc/manual/command-ref/nix-store.xml
```diff
@@ -1155,6 +1155,11 @@ that already exist in the Nix store are ignored.  If a path refers to
 another path that doesn’t exist in the Nix store, the import
 fails.</para>
 
+<para>
+Note: this needs to be either run as root, or on a seralization with
+signatures from a trusted cache.
```
edolstra: This is not quite correct. Import needs to be run as a user who has write access to the Nix store, or (when using the Nix daemon) a trusted user (one listed in the `trusted-users` option).

Also, there are no binary cache signatures in the export/import file format. There was a legacy signing mechanism, but it has been removed.

PR #1554, opened 2017-09-03
install-nix: Adding HOME checks

Anton-Latukha:
### Core part: It is possible HOME can be set even by system config, but there can be no directory in reality.
I found no bugreports. But I solve error I did couple of times myself through years. And making installation robust by enabling requirement checks for the right installation process is a great idea. I try to not interfere with any functional use cases.

How HOME var can be set, but directory not exist:
Example, admin in `adduser` don't provided '-m' key.
```
man adduser
...
-d, --home HOME_DIR
    ... The directory HOME_DIR does not have to exist but will not be created if it is missing.
...
-m, --create-home
    Create the user's home directory if it does not exist.
```

This is possible to mixup, because `usermod -d` creates HOME directories, but `adduser -d` - not.

### `echo` messages - self-explanatory of checks they do.

In this same script we do:
use HOME - so it needs to exist, be directory, and be executable
create $HOME/.nix-profile - so it needs to be writable
since script is `echo "performing a single-user installation of Nix..." >&2` - we can require that HOME is user owned.
in 99% read permissions also needed, but technically it is possible to operate and have a case when HOME can not have read permissions for user though (if admin not trust even user itself). HOME -wx is enough to go to known path ($HOME/.nix-profile) and it should have rwx for user, so user can operate as long, as he knows path.

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -22,6 +22,31 @@ if [ -z "$HOME" ]; then
     exit 1
 fi
 
+if [ ! -e "$HOME" ]; then
+    echo "$0: Home directory $HOME does not exist" >&2
+    exit 1
+fi
+
+if [ ! -d "$HOME" ]; then
+    echo "$0: Home directory $HOME is not a directory" >&2
+    exit 1
+fi
+
+if [ ! -w "$HOME" ]; then
+    echo "$0: Home directory $HOME is not writable for user $USER. No deal" >&2
+    exit 1
+fi
+
+if [ ! -O "$HOME" ]; then
+    echo "$0: Home directory $HOME is not owned by user $USER, users HOME probably must be owned by user" >&2
+    exit 1
+fi
+
+if [ ! -x "$HOME" ]; then
+    echo "$0: Home directory $HOME is not marked as executable for user $USER, how we are going to go into it?" >&2
+    exit 1
+fi
+
 # macOS support for 10.10 or higher
 if [ "$(uname -s)" = "Darwin" ]; then
     if [ $(($(sw_vers -productVersion | cut -d '.' -f 2))) -lt 10 ]; then


edolstra:
To be honest, if the user has a non-existent or otherwise invalid `$HOME`, they have bigger problems, and a lot more than the Nix installer is going to fail. We don't have to do 5 different checks to validate `$HOME`.

Anton-Latukha:
I've made a rewrite of script at this stage.

I look at it like this:
To make it fully robust and portable.
That on any environment it going to work.

Users on some systems can not have HOME, plethora of users can have single HOME, there are a lot of occasions.
If that are required conditions, they need to be checked, and installer needs to fallback/rollback. It is how I see it.

This is simple checks that does need no maintenance, and I am ready to pour time and wrote them.

I aim to make installer work in unknown Linux conditions. So it will work on any distribution/environment. And also work on already running servers. This is a bar I set for myself.

Anton-Latukha:
One time I done secure system that created a lot of users, for every client of the company.
And their home was: `/scp` folder with additional jailing for users.
`/scp` was home to big number of users.

Anton-Latukha:
@ edolstra 
Sorry, I am tired. I can sound not appreciative.
I just try to make installer an ask you to say, is root install is possible.

It should ease the migration of available admins/systems to Nix.

Thank you for everything.

Anton-Latukha:
Please, let me unveil some `shell-fu` on Linux installation.

I am not making 4 000+ shell script here. Thou I've could make much portable, advanced and safe install, using `shellfire` shell framework. But it going to have huge entrance level to modify it for anyone else.
And even 4 000+ shell script is not any performance hick-up for today systems, words from `shellfire` developer itself.

But the aim here is that - sh - already installed on any even the most bare system, and compatible with itself (almost).
sh is everywhere, even on Windows now, but I not try to make it for Windows now.

Anton-Latukha:
So it is bad, to do requirements checks and report to the user the status before doing work on his setup?

I am trying to do unattended Linux install on any distribution and setup. The more portable, polished and user friendly it will be - the bigger fame, community and project you will get.

Since Nix is perfectly fitting to do portable script (it installs separate infrastructure in `/nix` and `HOME/.nix-profile`). It is very real goal.

I just see, that you developed initial version of the script, and since than contribution to it was sporadic.

I am trying to send a little by little pull requests to get chain of process going.

wmertens:
@Anton-Latukha actually, the only thing that matters is that you can write the requisite files into $HOME. It doesn't matter where $HOME is or who owns it, just that you can write, and only when you need to do the write.

So I think the only useful test is to check if $HOME is defined, and then let errors during installation just happen.

Anton-Latukha:
@wmertens 

1. Thank you on input.

2. Doesn't HOME must have also 'x', to be able to open it:
```shell
$ mkdir /tmp/test_folder
$ chmod 0200 /tmp/test_folder
$ touch /tmp/test_folder/file
touch: cannot touch '/tmp/test_folder/file': Permission denied

$ chmod 0300 /tmp/test_folder
$ echo "Hello, I'm here!" >> /tmp/test_folder/file
$ cat /tmp/test_folder/file
Hello, I'm here!
```

3. I though to do all requirement checks at start.
And if they pass - do changes on the system.
It is by design more clean installer this way.
As I started in "install-nix: Full progress of Linux installation #1565". All checks go before doing any actions.

4. That is why also I do additional checks. Does HOME is folder or link to folder, because it can be file.

5. And check if HOME owned by user. Because it can be a community folder with wide access rights.

And when people would come with bugreport of esoteric case, when HOME is not owned by user, - it can be modified to also permit that legitimate but esoteric use.

I can agree that last point is debatable. But, check, does HOME is folder, and about also needing 'x' to write anything to folder - I'm solid on that.

Anton-Latukha:
Ok.

I close this.

Issue #539, opened 2015-05-19
installation fails while asking `sudo`, _after_ downloading the 69Mb archive


vn971:
The problem is with unnecessary resource consumption and a surprising UX. It happens because the tarball script requires `sudo` but the "install" script does not check its availability.

Current behavior:
On a clean Debian 8 distribution, the install script will not check `sudo` availability and will eagerly download & execute the tarball script. The tarball script will fail, causing the user to re-download the archive to try again...

Expected behavior:
Early failure with the install script after seeing that `sudo` is not available.

Notes:
I'm actually not sure whether the "install" script should know the internals of the tarball script. But for a default path of a Debian user, the installation will fail, and that's not cool I think.

vn971:
I tried to generalize the issue in #540.
For a script that caches the tarball, there would be no issue in running the script once again (practically rendering this ticket and a couple of others to fixed).

tomberek:
Closing as stale. Re-open if needed.

PR #1555, opened 2017-09-03
install-nix: Advancement of destination folder creation means

Anton-Latukha:
This commit concentrates on streamlining `/nix` folder creation.

As Nix installation can be launched from any user, on any distro, with `sudo` or not - here is universal code.

1. At first it tries to create folder from current user. If it is root, or user can create directory - no need to use sudo or su or disturb user with additional messages.
2. Detects does `sudo` installed. Before this we could not distinct when `sudo` access failed itself and when `sudo` is not installed. And now we can say that we looked and `sudo` is not installed, and it is understandable why we ask him root password itself.
3. Looks does `sudo` access is passwordless, or not.
3a. Passwordless: we can inform user that we used `sudo`.
3b. With password: we explain to user why we need sudo access, and ask to enter password for that.
4. If `sudo` fails, we mention that it also can be due to user has no sudoers access. Nowadays `sudo` pretends that password do not match, while in reality user has no sudoers access. The One `sudo` developer, - millert - made that change years ago to not inform hackers that user has no `sudo` access, but those reports is written to logs. But admins/users also can be puzzled by that, especially if passwords/infrastructure is complex.
5. If folder still not created - use `su`.
5a.And big volume of clean distributions (minimalist ones) doesn't have `sudo` supplied with a distro (Alpine, Arch, even Debian I think).
5a. And we explain why installer asks for root password itself.
6. Without informing user when root access closed - he doesn't sure how many operations was under root after he gave access, and what Nix beast will do with his system. We now inform user, in both sudo and su, when we closed root assess, that he kindly entrusted to us.
7. Script errors only when all merits was used - only then. So user has less hassle, we are trying in his benefit.
8. More vocal installation process overall. Which is important for smooth experience and good impression.
9. Yes, I also changed:
```sh
if ! [ -e $dest ]; then # it can also be file or link to file.
# to
if ! [ -d $dest ]; then # it can be only directory/link to directory
```

And, yes; solves #539. Now script not falls because `sudo` is not present, - it uses root, user rights or `su`.
I looked though all 'install' bugs and that's all I found related, but I by the way triaged a handful for people bugs, old ones also.

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -40,13 +40,51 @@ fi
 
 echo "performing a single-user installation of Nix..." >&2
 
-if ! [ -e $dest ]; then
-    cmd="mkdir -m 0755 $dest && chown $USER $dest"
-    echo "directory $dest does not exist; creating it by running '$cmd' using sudo" >&2
-    if ! sudo sh -c "$cmd"; then
-        echo "$0: please manually run '$cmd' as root to create $dest" >&2
-        exit 1
+
+if [ -d "$dest" ] ; then
+
+    echo "Directory $dest already exists; skipping creation"
+
+else
+    echo "Directory $dest does not exist;"
+
+    # Shell parses variable with current user, so value under sudo/su not changes.
+    # It is const, but don't move it above USER set/checks.
+    readonly cmd="mkdir -m 0755 $dest && chown $USER $dest"
+
+    # Trying to create $dest by all means
+
+    # For case user is root, or has access to create $cmd. Else -> do 'then' section
+    if sh -c "$cmd" > /dev/null 2>&1 ; then
+        echo "Creating $dest directory by running $cmd using $USER rights"
+    else
+        if which sudo > /dev/null 2>&1 || type sudo > /dev/null 2>&1 ; then
+            echo "Creating $dest directory by running $cmd using sudo:"
+            if sudo -n true 2> /dev/null ; then
+                echo "Checked: sudo access does not require password."
+            else
+                echo "Checked: sudo access requires to enter password:"
+            fi
+            # Using sudo to create directory
+            if ! sudo sh -c "$cmd" && echo "Exiting priviliged mode." ; then
+                echo "Password don't match or user $USER not in sudoers." >&2
+            fi
+        else
+            echo "'sudo' seems not installed"
+        fi
+
     fi
+
+    # if dest still not created (with user or sudo) - use su to access root
+    if [ -d "$dest" ] ; then
+        echo "Trying to get privileges using su and root password"
+        if ! su sh -c "$cmd" && echo "Exiting priviliged mode."; then
+            # all means failed, Exit!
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
There are 555 open issues.
My PR is 1555.

Thug Life!

danbst:
I think, every case when `$cmd` fails, should end with

`You can skip this check by manually running '$cmd' as root to create $dest`

Also, I don't think that script that writes WHY does it need sudo-permissions will be more trusted then current one (I'm speaking for myself here). I don't trust random curl|sh scripts, even if they post to `/dev/stderr` they do only good.

Triying `su` when `sudo` not found may be a good change though

Anton-Latukha:
When you run remotely on deployment:
`curl https://nixos.org/nix/install | sh`
You don't wait on message:
```shell
You can skip this check by manually running '$cmd' as root to create $dest
```
That shows that install is not production ready.

Run:
`curl -L https://get.docker.com/ | sh`
`curl -L https://bootstrap.saltstack.com | sudo sh`

Does they ask to do something manually before install?

They created to be automatic.

danbst:
> This is our worm welcome to world of Nix.
This is great, but let me repeat my point again, in detail:

I'm talking about situation, when actual install script failed. Don't know why, probably user chose not to enter sudo password (`hey man, you're a script from internet that wants root access, why would I trust you? no-no ^C`). They've already read that Nix doesn't need root in single-user usage, but why the hell does it need root for install?

In that case, when actual install failure happened, it is good to explain user that if he can create and own `/nix`, then script will run without `sudo`, `su` or any other access to root.

I'm also 👍 on using su when sudo not found.

Anton-Latukha:
1. If user runs install script on Linux, - user expects that superuser can be used.
    Doesn't he?
    On what occasions you was installing applications on Linux without superuser access?
    When he installed on Linux something, that was not requiring superuser access, especially package manager.

    It it really astonishing and not expected for user that `hey man, you're a script from internet that wants root access, why would I trust you?`

2. Once more best production oneliners that I mentioned above - expect user is knows when he runs it. They are unattended.

3. If user runs `curl https://nixos.org/nix/install | sh` directly, - that means he trust in script or don't bother.

4. User will look at `https://nixos.org/nix/install`, or download it and check-out.

    For me - this is defence from the fool. People are decide for them-self to look at it and launch it.

    That can be mentioned in official installation. It is not role of the script. To do the job, and not do the job, - it is not realistic.

5. User can just not enter password.
When `sudo` or `su` access asked, user can do ctlr-c, to exit, it is no biggy.

6. Script tells why it needs superuser access.

7. My pull request does not changed point you object to. It existed before me.
Currently main way code runs on master - it runs `sudo` directly. So your note must be pointed at master code.

6. In fact I even work in favour of your objection here.
Before using `sudo` or `su,- my code tries to use user rights itself.
So if we can do it from user, and not use `sudo` or `su` - it does it without them.

Anton-Latukha:
Ok. I close this.

PR #1562, opened 2017-09-08
docker: update to Nix 1.11.14

peti:
- Use the latest Nix version 1.11.14.

- Attempts to download the Nix installation tarball from http://nixos.org
  redirect to https these days, which wget doesn't support unless OpenSSL is
  available.

- Use addgroup and adduser commands to create the Nix build users.

- Split image generation into multiple steps instead so that intermediate
  images can be cached.

- Link the Nix profile script into /etc/profile.d, where it's run
  automatically.

Use the command "docker build -t nix . && docker run -it --rm nix sh -"
to build and run the Nix docker container.

### DIFF
--- a/misc/docker/Dockerfile
+++ b/misc/docker/Dockerfile
@@ -1,13 +1,16 @@
 FROM alpine
 
-RUN wget -O- http://nixos.org/releases/nix/nix-1.11.7/nix-1.11.7-x86_64-linux.tar.bz2 | bzcat - | tar xf - \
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Image split is appreciated by me.
Also using link to `/etc/profile.d/` is a more proper approach.

msegado:
FYI, each of those `RUN` commands adds an immutable layer to the filesystem... the last two commands for example ("Clean up files that we no longer need") will not make the image any smaller since they can't change the layers before them, and in fact they'll make it slightly _larger_ since they add two layers themselves. More info on this: https://beenje.github.io/blog/posts/dockerfile-anti-patterns-and-best-practices/

Given that most of those RUN commands should be fast (there's not much saved by caching the result of, e.g., `addgroup -g 30000 -S nixbld`) and that the caches would be invalidated any time you changed the version in the first command anyway, I expect it makes sense to keep them all into a single command as in the original Dockerfile.

peti:
@msegado, thank you very much for pointing that out. I no idea that every RUN execution creates a new dependency that cannot be removed. I assumed that the final image would stand on its own -- but clearly it does not. I've adoped the PR accordingly.

domenkozar:
Thanks!

PR #1571, opened 2017-09-21
Improve error message for conflicting priorities

teto:
I find the error message 'nix-env --set-flag priority NUMBER PKGNAME'
not as helpful as it could be :
- doesn't share the current priorities
- doesn't say that the command must be run on the already installed
PKGNAME (which is confusing the first time)
- the doc needs careful reading:
"If there are multiple derivations matching a name in args that have the same name (e.g., gcc-3.3.6 and gcc-4.1.1), then the derivation with the highest priority is used."
if one stops reading there, he is screwed. Salvation comes with reading "A derivation can define a priority by declaring the meta.priority attribute. This attribute should be a number, with a higher value denoting a lower priority. The default priority is 0."
To sum it up, lower number wins. I tried to convey this idea in the
message too.

I haven't used this in production because nixUnstable tends to break my install and I wanted to check for agreement first.

### DIFF
--- a/src/buildenv/buildenv.cc
+++ b/src/buildenv/buildenv.cc
@@ -74,10 +74,11 @@ static void createLinks(const Path & srcDir, const Path & dstDir, int priority)
                     auto prevPriority = priorities[dstFile];
                     if (prevPriority == priority)
                         throw Error(format(
-                                "collision between '%1%' and '%2%'; "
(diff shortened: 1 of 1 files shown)


domenkozar:
Thank you!

teto:
Thanks for the merge. The previous message did confuse me at the beginning (I know it's true for others too) and I hope this one might help beginners.

PR #1578, opened 2017-09-25
nix-channel: remind user to update after adding

lheckemann:
This reduces the likelihood of a common pitfall; I've often forgotten to run --update after adding a channel or changing a channel URL.

### DIFF
--- a/src/nix-channel/nix-channel.cc
+++ b/src/nix-channel/nix-channel.cc
@@ -50,6 +50,7 @@ static void addChannel(const string & url, const string & name)
     readChannels();
     channels[name] = url;
     writeChannels();
+    std::cerr << "Remember to run nix-channel --update to apply the changes!\n";
 }
 
 static auto profile = Path{};


edolstra:
Arguably this violates the Unix philosophy of not spamming the user with non-error messages: http://www.faqs.org/docs/artu/ch11s09.html

lheckemann:
- interoperability: it's on standard error so it shouldn't break anything in that respect
- vertical screen space: it's a single line
- bandwidth: not really applicable here as it's a practically instantaneous operation

Personally I've found myself forgetting to update channels after modifying them frequently enough that to me this change is justified; @infinisil had the idea after a newbie came asking, having the same issue, on IRC. I won't insist on the change, but personally think it would be for the better.

edolstra:
Unnecessary messages harm composability. For example, a script that does `nix-channel --add; nix-channel --update` will still print `Remember to run nix-channel --update`, thereby confusing the user.

Issue #1573, opened 2017-09-21
Really annoying apparent deadlock during `nix-store -r` in Nix 1.12


copumpkin:
I'm on a Linux box running Nix unstable at 1.12pre5511_c94f3d55 and am seeing a hard-to-reproduce/minimize freeze during `nix-store -r`. The multi-thread stack trace is at [this gist](https://gist.github.com/copumpkin/d1ccd6ef9e66cc74cb295fed395359b7) and I'm trying to figure out what's confusing it.

The Nix on that box is still frozen if you have ideas for other things to try while it's frozen before I kill it.

Were there any recent changes in the concurrency code? I was on a `nixUnstable` a couple months ago that I never observed this problem on, but perhaps I was just lucky?

cc @edolstra

copumpkin:
What I've found out so far:

I have a thread calling `download()` which waits on a future to signal completion/failure from the downloader. That future never signals, which suggests that https://github.com/NixOS/nix/blob/aca4f7dff0ee615164668e923deaf5cc96ab1c8a/src/libstore/download.cc#L271-L279 is never happening.

The other thing I've found by poking around the stack trace and memory is that the download thread appears to be stuck on an infinite poll of our signalling FD:

https://github.com/NixOS/nix/blob/aca4f7dff0ee615164668e923deaf5cc96ab1c8a/src/libstore/download.cc#L424-L433

Meaning that the `CURLM` multi handle has no FDs of its own, and is just polling on our extra `wakeupPipe` one. I'm guessing we never signal it. Perhaps some sort of deadlock around that?

dezgeg:
The 'never signaling' theory should be verifiable by checking if `state->incoming` is non-empty.

copumpkin:
I actually realized that I was stuck on 1fe1976e0d68b520f04f8ad4f619bb52c40cef78 with this symptom, and when I updated to c94f3d55, it appears (hard to be sure given how nondeterministic this is) that all the hangs turned into timeout errors from CURL: 

```
error: unable to download ‘https://some-cache.s3.amazonaws.com/nix-cache-info’: Timeout was reached (28)
```

copumpkin:
Since switching to c94f3d5 I indeed haven't had any deadlocks, but am getting a lot more of those timeout errors. I'm wondering if something got restructured in the threading code, and maybe in the past the timeouts from curl led to a deadlock?

Anyway, here's a bunch of Nix logs of the deadlock with increased verbosity from back in 1fe1976:

```
[...]

PR #1580, opened 2017-09-25
Retry in all error cases but a few

copumpkin:
~~I could make this optional but it doesn't seem like too weird of a thing to enable in general.~~

As we discussed on IRC

This is what I'm currently running into in #1573, even though it no longer causes deadlocks. I'm honestly not sure if it was what was causing deadlocks before, but since updating Nix I'm now getting timeout errors with the same frequency as I was getting deadlocks before.

I'm unsure what to do about timeouts. The behavior in this PR will retry on curl connect timeouts (which happens to be the behavior I want, since I'm getting occasional timeouts from dud S3 backend servers), but that doesn't seem like something everyone would want.

cc @edolstra

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,16 +99,20 @@ perl/Makefile.config
 /misc/systemd/nix-daemon.socket
 /misc/upstart/nix-daemon.conf
 
+/src/resolve-system-dependencies/resolve-system-dependencies
+
 inst/
 
 *.a
 *.o
 *.so
+*.dylib
 *.dll
 *.exe
 *.dep
 *~
 *.pc
+*.plist
 
 # GNU Global
 GPATH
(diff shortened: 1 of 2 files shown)


edolstra:
Maybe it would be better to set the default value for `connect-timeout` to something much lower (say 10 seconds) and then retry timeouts unconditionally.

copumpkin:
@edolstra I considered that, but I don't think the retry count is fixed or known ahead of time, right? Different download invocations can pass in different values for it I think.

copumpkin:
@edolstra I updated it as discussed on IRC. Let me know what you think about the timeout concern.

copumpkin:
@edolstra any thoughts on this? I'm happy to make further changes if you see fit, but ultimately just want to stop trying to catch every little error that flaky servers throw my way.

PR #1586, opened 2017-10-03
install-darwin-multi-user: relax assumption check

acowley:
The installer will error out if a user's shell configuration includes any mention of `nix-profile`, even if this is in a comment. This change is designed to do the bare minimum to ignore lines beginning with a `#`.


acowley:
Friendly ping to @grahamc

grahamc:
LGTM, it'll need to be backported to 1.11-maintenance

PR #1588, opened 2017-10-03
More elegant test output

### LINKED ISSUE #880: Test suite is noisy enough to make it hard to figure out what failed
I tried running `make installcheck` on darwin and got a failure. Now I'm trying to parse through the test output and am seeing dozen of red "error" appearances, but I think those are expected. The final message doesn't tell me which test failed, so I'm struggling to figure out what's wrong.

copumpkin:
I got sick of trying to find the failures in the sea of debug output, so we now:

- Hide test output unless it fails
- Sprinkle in some simple color if we're in a terminal that supports it (so Hydra/CI doesn't get cluttered up with escape codes)
- Pad results for a more tabular look

If Nix is getting a more friendly user interface, we might as well get a friendlier developer interface, right? :)

I realize we don't want to add a dependency on a third-party test framework so I figured a few lines of make/bash could make this a lot more pleasant to work with.

Fixes #880 

cc @edolstra @domenkozar @shlevy

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -7,20 +7,38 @@ define run-install-test
 
 endef
 
+# Color code from https://unix.stackexchange.com/a/10065
 installcheck:
-	@total=0; failed=0; for i in $(_installcheck-list); do \
+	@total=0; failed=0; \
+	pad="                           "; \
+	red=""; \
+	green=""; \
+	normal=""; \
+	if [ -t 1 ]; then \
+		ncolors="$$(tput colors)"; \
+		if [[ -n "$$ncolors" && $$ncolors -ge 8 ]]; then \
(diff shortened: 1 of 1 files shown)


domenkozar:
Ideally we'd a sample of output, but looks good! :)

edolstra:
Thanks!

Mic92:
this broke `nix-build -A deb_ubuntu1604x86_64 release.nix`:

```
GEN    tests/common.sh
/bin/sh: 7: tput: not found
/bin/sh: 8: [[: not found
/bin/sh: 16: Bad substitution
```

Mic92:
https://github.com/NixOS/nix/pull/1593

PR #1593, opened 2017-10-06
fixing bashisms in test code

Mic92:
This fixed the build on ubuntu/debian, where dash is the sh.

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -10,13 +10,12 @@ endef
 # Color code from https://unix.stackexchange.com/a/10065
 installcheck:
 	@total=0; failed=0; \
-	pad="                           "; \
 	red=""; \
 	green=""; \
 	normal=""; \
 	if [ -t 1 ]; then \
 		ncolors="$$(tput colors)"; \
-		if [[ -n "$$ncolors" && $$ncolors -ge 8 ]]; then \
+		if [ -n "$$ncolors" ] && [ "$$ncolors" -ge 8 ]; then \
 			red="$$(tput setaf 1)"; \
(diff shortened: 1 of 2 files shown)


copumpkin:
Damn Ubuntu, that `dash` decision they made has screwed me over so often (and not just with stuff I write 😦)

Makes me sad to see the padding go but I guess this is better than not working at all.

Mic92:
I have not find an elegant way to do this with posix shell.

copumpkin:
Can we just somehow tell Make to use /bin/bash?

Mic92:
then everybody needs to remember to use make SH=/bin/bash

orivej:
> Can we just somehow tell Make to use /bin/bash?

[Yes](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html), with `SHELL = /bin/bash` in the Makefile. But there is no `/bin/bash` in NixOS by default. This can be accounted for with:
```make
SHELL := $(wildcard /bin/bash)
ifneq ($(SHELL), /bin/bash)
SHELL := /bin/sh
endif
```

bjornfor:
Or `SHELL = bash`? (Does it have to be absolute path?)

orivej:
@bjornfor That's perfect!

bjornfor:
The linux kernel has this Makefile snippet, but I figure we can skip right to the `$PATH` lookup part :-)

```
CONFIG_SHELL := $(shell if [ -x "$$BASH" ]; then echo $$BASH; \
	  else if [ -x /bin/bash ]; then echo /bin/bash; \
	  else echo sh; fi ; fi)
```

PR #1594, opened 2017-10-07
release-common: Fix busybox builtins (busybox >= 1.27)

dtzWill:
See https://github.com/NixOS/nixpkgs/pull/28261

If this is fixed here, it should be fixed in nixpkgs as well.


dtzWill:
It does seem curious that things haven't broken, perhaps they're not needed after all? :stuck_out_tongue_winking_eye:

PR #1601, opened 2017-10-16
fetchgit: Remove incomplete/unneeded isURI check.

shlevy:
This check spuriously fails for e.g. git@github.com:NixOS/nixpkgs.git,
and even for ssh://git@github.com/NixOS/nixpkgs.git, and is made
redundant by the checks git itself will do when fetching the repo.

### DIFF
--- a/src/libexpr/primops/fetchgit.cc
+++ b/src/libexpr/primops/fetchgit.cc
@@ -13,9 +13,6 @@ namespace nix {
 Path exportGit(ref<Store> store, const std::string & uri,
     const std::string & ref, const std::string & rev)
 {
-    if (!isUri(uri))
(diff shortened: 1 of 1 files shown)


edolstra:
Well, some sort of validation is in order to prevent a command line injection...

shlevy:
Why? Does runProgram use a shell?

edolstra:
E.g. if the URI begins with a dash, git will interpret it as a flag.

shlevy:
Can't we just use a `--` before the repo/rev?

edolstra:
Yeah that's probably good enough.

shlevy:
OK incoming shortly

shlevy:
@edolstra good now?

PR #1605, opened 2017-10-16
Pull nix-profile-daemon from 1.11

shlevy:


edolstra:
👎 on this. It means that we now have to maintain two divergent profile scripts, in addition to the one in NixOS. 

As an example of divergence: nix-profile.sh tests various values for `NIX_SSL_CERT_FILE`, nix-profile-daemon.sh doesn't.

Also, this file is not referenced anywhere so I don't really understand what it's for.

shlevy:
@edolstra It gets installed as `nix-daemon.sh`, which is sourced by the multi-user installer. @grahamc How hard would it be to have a single profile script that can handle single and multi user, do you think?

PR #1611, opened 2017-10-17
fetchgit: Comprehensive valid reference check.

shlevy:
This ensures we're matching git's definition of a valid reference.


edolstra:
:-1: on doing an uncached call to an external program.

Also, why not just call `runProgram`?

shlevy:
Is caching in memory OK?

I didnt use runProgram vecau

shlevy:
@edolstra updated with caching and using runProgram

shlevy:
Hmm, not sure why PR isn't updating, maybe because I accidentally closed it earlier? Will open a new one.

PR #1612, opened 2017-10-17
fetchgit: Comprehensive valid reference check.

shlevy:
This ensures we're matching git's definition of a valid reference.


shlevy:
Replaces #1611

edolstra:
Why is it important that we match git's notion of a reference? Won't we just get an error eventually if we pass an invalid ref?

shlevy:
Probably. I'm not sure what will happen if the passed in ref has a : or some other special git characters though, maybe it's fine

shlevy:
#1614 just removes the check.

PR #1617, opened 2017-10-18
[Nasty sketch] Stop crashing High Sierra

copumpkin:
This is a nasty approach and I don't like it, but given that the other `kill(-1, SIGKILL)` approach crashes macOS 10.13 and we don't know when Apple will fix their kernel, we're left with shitty options.

I threw this together rather hastily to get people thinking about it. It's untested (I don't have a High Sierra or even a multi-user install handy right now) and I'm on a train and can't do much, but I figured I'd throw up the sketch and see what y'all thought of the approach.

cc @shlevy @edolstra @LnL7 @grahamc @matthewbauer

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -23,7 +23,7 @@
 #include <unistd.h>
 
 #ifdef __APPLE__
-#include <sys/syscall.h>
+#include <sys/sysctl.h>
 #endif
 
 #ifdef __linux__
@@ -786,13 +786,64 @@ pid_t Pid::release()
     return p;
 }
 
-
 void killUser(uid_t uid)
 {
     debug(format("killing all processes running under uid '%1%'") % uid);
 
     assert(uid != 0); /* just to be safe... */
 
+#ifdef __APPLE__
+    /* We originally had very similar logic here as on the Linux side, but
+       then got bitten hard by a system crashing bug as described in here:
+       https://github.com/NixOS/nix/issues/1583. So now we adopt this other
+       approach, but we should go back to the kill approach once Apple fixes
+       their kernel. */
+
+    // Borrowed from https://stackoverflow.com/a/7733928
+    int err = 0;
+    size_t length = 0, newlength = 0;
+
+    // XXX: Should this be KERN_PROC_RUID? EUID vs. real UID
+    static const int name[4] = { CTL_KERN, KERN_PROC, KERN_PROC_UID, static_cast<int>(uid) };
+
+    int tries = 100;
+    bool done = false;
+
+    while (!done) {
+        tries--;
+
+        // First call with NULL to get number of processes. Yeah yeah TOCTOU is fun, but we check
+        // that it hasn't changed later and keep going until it's 0
+        err = sysctl((int *)name, 4, NULL, &length, NULL, 0);
+        if (err)
+            throw SysError("calling sysctl to get size");
+
+        vector<struct kinfo_proc> proc_list(length);
+
+        newlength = length;
+        err = sysctl((int *)name, 4, (void *)proc_list.data(), &newlength, NULL, 0);
+        if (err)
+            throw SysError("calling sysctl to get process list");
+
+        // @grahamc points out https://utcc.utoronto.ca/~cks/space/blog/unix/ProcessKillingTrick
+        for (auto proc : proc_list) {
+            kill(proc.kp_proc.p_pid, SIGSTOP);
+        }
+
+        for (auto proc : proc_list) {
+            kill(proc.kp_proc.p_pid, SIGKILL);
+        }
+
+        if (newlength == 0)
+            break;
+
+        if (tries <= 0)
+            break;
+    }
+
+    if (newlength != 0)
+        throw SysError("ran out of tries but there were processes left. I've created a monster!!!");
+#else
     /* The system call kill(-1, sig) sends the signal `sig' to all
        users to which the current process can send signals.  So we
        fork a process, switch to uid, and send a mass kill. */
@@ -806,16 +857,7 @@ void killUser(uid_t uid)
             throw SysError("setting uid");
 
         while (true) {
-#ifdef __APPLE__
-            /* OSX's kill syscall takes a third parameter that, among
-               other things, determines if kill(-1, signo) affects the
-               calling process. In the OSX libc, it's set to true,
-               which means "follow POSIX", which we don't want here
-                 */
-            if (syscall(SYS_kill, -1, SIGKILL, false) == 0) break;
-#else
             if (kill(-1, SIGKILL) == 0) break;
-#endif
             if (errno == ESRCH) break; /* no more processes */
             if (errno != EINTR)
                 throw SysError(format("cannot kill processes for uid '%1%'") % uid);
@@ -827,12 +869,14 @@ void killUser(uid_t uid)
     int status = pid.wait();
     if (status != 0)
         throw Error(format("cannot kill processes for uid '%1%': %2%") % uid % statusToString(status));
-
+ 
+#endif
     /* !!! We should really do some check to make sure that there are
        no processes left running under `uid', but there is no portable
        way to do so (I think).  The most reliable way may be `ps -eo
        uid | grep -q $uid'. */
 }
+
 
 
 //////////////////////////////////////////////////////////////////////


copumpkin:
This compiles so ship it, right? :smile:

src/libutil/util.cc
```diff
@@ -786,13 +786,59 @@ pid_t Pid::release()
...
+        // First call with NULL to get number of processes. Yeah yeah TOCTOU is fun, but we check
+        // that it hasn't changed later and keep going until it's 0
+        err = sysctl((int *)name, 4, NULL, &length, NULL, 0);
+        if (err)
+            throw SysError("calling sysctl to get size");
+
+        vector<struct kinfo_proc> proc_list(length);
+
+        newlength = length;
+        err = sysctl((int *)name, 4, (void *)proc_list.data(), &newlength, NULL, 0);
```
copumpkin: I was originally going to check if `length == newlength` but then couldn't see a reason to...

src/libutil/util.cc
```diff
@@ -786,13 +786,64 @@ pid_t Pid::release()
...
+       then got bitten hard by a system crashing bug as described in here:
+       https://github.com/NixOS/nix/issues/1583. So now we adopt this other
+       approach, but we should go back to the kill approach once Apple fixes
+       their kernel. */
+
+    // Borrowed from https://stackoverflow.com/a/7733928
+    int err = 0;
+    size_t length = 0, newlength = 0;
+
+    // XXX: Should this be KERN_PROC_RUID? EUID vs. real UID
```
shlevy: They should be identical for normal programs run by the user, so whichever one is set to the nixbld user for a setuid program, right?

src/libutil/util.cc
```diff
@@ -786,13 +786,64 @@ pid_t Pid::release()
...
+
+        newlength = length;
+        err = sysctl((int *)name, 4, (void *)proc_list.data(), &newlength, NULL, 0);
+        if (err)
+            throw SysError("calling sysctl to get process list");
+
+        // @grahamc points out https://utcc.utoronto.ca/~cks/space/blog/unix/ProcessKillingTrick
+        for (auto proc : proc_list) {
+            kill(proc.kp_proc.p_pid, SIGSTOP);
+        }
```
grahamc: It is possible that between listing the processes and the SIGSTOP, the process dies and the PID is reused by another user. After the SIGSTOP, we can take a look to see if the PID is owned by the build user we're wanting to kill. If it is our builder, SIGKILL it. If not, SIGCONT it.

@Dezgeg was kind enough to point out SIGCONT to a SIGSTOP'd process, which was SIGSTOP'd before we got to it is rude... which is true, but jeeze, that means the processes exited, the pid recycled, started, and SIGSTOP'd within such a brief period of time. Still rude, but we tried (?)

src/libutil/util.cc
shlevy: Can you use a `const_cast` here with a comment reassuring us that the `name` argument won't be modified based on some manpage or whatever?

src/libutil/util.cc
```diff
@@ -786,13 +786,64 @@ pid_t Pid::release()
...
+    static const int name[4] = { CTL_KERN, KERN_PROC, KERN_PROC_UID, static_cast<int>(uid) };
+
+    int tries = 100;
+    bool done = false;
+
+    while (!done) {
+        tries--;
+
+        // First call with NULL to get number of processes. Yeah yeah TOCTOU is fun, but we check
+        // that it hasn't changed later and keep going until it's 0
```
shlevy: We don't check that it hasn't changed later, right?

src/libutil/util.cc
```diff
@@ -786,13 +786,64 @@ pid_t Pid::release()
...
+            throw SysError("calling sysctl to get size");
+
+        vector<struct kinfo_proc> proc_list(length);
+
+        newlength = length;
+        err = sysctl((int *)name, 4, (void *)proc_list.data(), &newlength, NULL, 0);
+        if (err)
+            throw SysError("calling sysctl to get process list");
+
+        // @grahamc points out https://utcc.utoronto.ca/~cks/space/blog/unix/ProcessKillingTrick
```
shlevy: A brief description of why this helps inline would be nice.

src/libutil/util.cc
shlevy: Maybe we should do the `kill`'s as the nixbld user instead?

shlevy: Then if it fails due to EPERM, we just continue on our way

grahamc: 🔥

shlevy: @copumpkin Probably this whole loop should just be done as the `nixbld` user (excluding the current process, of course)

grahamc: It'd be really helpful if we could SIGKILL -1 and only kill things we own.

copumpkin:
Thanks for all the feedback! Will fix when train wifi wakes up again or tomorrow

src/libutil/util.cc
copumpkin: Yessir!

src/libutil/util.cc
copumpkin: 👍

src/libutil/util.cc
copumpkin: Sounds good, I'll run as subprocess again

copumpkin: But actually, if I run as subprocess, we need to start worrying about our siblings killing us again...

src/libutil/util.cc
copumpkin: Yeah I don't. Should I? My first thought was that I'd need to, but then I thought some more and convinced myself I wouldn't.

matthewbauer:
I think you’re also going to need to handle this SIGKILL:

https://github.com/NixOS/nix/blob/1dd29d7aebae706f3e90a18bbfae727f2ed03c70/src/libstore/build.cc#L1013

src/libutil/util.cc
shlevy: But if our siblings kill us we know there's a rogue

src/libutil/util.cc
shlevy: Nah, shouldn't be necessary, just need to update the comment

copumpkin:
@matthewbauer I was never able to get it to die when killing a process group. Were you? I left it running for a while but only ever saw crashes from `kill(-1, ...)`

matthewbauer:
@copumpkin Ok maybe not... I guess I was thinking based on 3cc1b57ac0030287a6f0806b0a21e44db21787d2 still leading to crashes for some people. That was before we saw the issue with plain old ```kill(-1, SIGKILL)``` though.

matthewbauer:
@copumpkin I've got this branch installed and running on High Sierra. So far so good. Will report back if I get a system lock again.

copumpkin:
Nice!

src/libutil/util.cc
```diff
@@ -786,13 +786,64 @@ pid_t Pid::release()
...
+
+    int tries = 100;
+    bool done = false;
+
+    while (!done) {
+        tries--;
+
+        // First call with NULL to get number of processes. Yeah yeah TOCTOU is fun, but we check
+        // that it hasn't changed later and keep going until it's 0
+        err = sysctl((int *)name, 4, NULL, &length, NULL, 0);
```
matthewbauer: Does this work for you @copumpkin ?

Trying this little program is giving really big length values:

```
#include <sys/sysctl.h>
#include <stdio.h>
#include <string.h>

#define UID 30001

int main() {
        size_t length = 0;
        int errno = 0;
        static const int name[4] = { CTL_KERN, KERN_PROC, KERN_PROC_UID, UID };
        errno = sysctl((int *)name, 4, NULL, &length, NULL, 0);
        printf("uid:%i length:%zu err:%s\n", UID, length, strerror(errno));
}
```

copumpkin: Maybe I screwed it up! I haven't actually run this yet 😄

matthewbauer: Haha that's fine! I'll go through it and see whats works. Before now I had hoped apple would just fix the issue but 10.13.1 still has the issue.

copumpkin: Well each process entry is actually a struct so that might be fairly large. How big are the values you're seeing? Also, now that I think about it, my vector might be too big, since I should probably divide it by the size of the struct.

matthewbauer: Yep that looks more like it (ceacd3e565931797e3e4447a2649c78dfc3b3d45):

Although it looks like it's the same regardless of the user:

```
euid:501 uid:501 length:344 err:Undefined error: 0
euid:501 uid:30001 length:344 err:Undefined error: 0
```

copumpkin: Weird, maybe we'll just need to filter by hand based on the struct?

copumpkin: Or maybe try `KERN_PROC_RUID`

matthewbauer:
Hey @copumpkin,
I've updated a few things here: https://github.com/matthewbauer/nix/tree/shitty-kill-processes

So far so good. Going to use it for a little while until I say it's worth merging but haven't had the hanging yet. (lot's of weird High Sierra stuff though- but like I've said I'd blame it on High Sierra's bugginess first).

Issue #1402, opened 2017-06-10
How to remove nix


jarry9:
Hi,

I am using nix on Ubuntu 16. I installed nix by these commands:

bash <(curl https://nixos.org/nix/install)
source ~/.nix-profile/etc/profile.d/nix.sh

I have not started nix-daemon. Or I did not use any extra command...

I will not remove the nix package manager now. But If I should remove it, what should I type on command line? Is there any ready script or something?

I found this topic: https://github.com/NixOS/nix/issues/458 But it is not completely clear about uninstallation. Because there is not only /nix folder. There are link files on HOME + nix has set some directories to PATH variable?

How can I make sure that I remove it completely?

Thank you

zimbatm:
To uninstall nix you will need to do the following things:
* remove the `. "$HOME/.nix-profile/etc/profile.d/nix.sh"` in your `~/.profile` or `~/.bash_profile`
* `rm -rf $HOME/{.nix-channels,.nix-defexpr,.nix-profile,.config/nixpkgs}`
* `sudo rm -rf /nix`
On a multi-user install, also remove all the `nixbld[0-9]` users

jarry9:
Thank you

pynixwang:
no offical uninstaller?

pynixwang:
I find a new way, install again and installer will tell you how to remove it, all included but deleting user.

SimSez:
Hi there, 

I've completed the third step (sudo -fr/nix)

How does one go about these first two steps on OSX, is this from within the terminal? I can't seem to find the .profile or .bash_profile 

>To uninstall nix you will need to do the following things:

remove the . "$HOME/.nix-profile/etc/profile.d/nix.sh" line in your ~/.profile or ~/.bash_profile

rm -rf $HOME/{.nix-channels,.nix-defexpr,.nix-profile,.config/nixpkgs}

sudo rm -rf /nix

JamesMcMahon:
> I find a new way, install again and installer will tell you how to remove it, all included but deleting user.

@pynixwang This didn't work for me under macOS. What command did you run to see the uninstall options?

jbyler:
I followed @zimbatm's instructions, with some tweaks:
- my personal dotfiles didn't have anything to load nix, but I did have to clean up `/etc/{profile,bashrc,zshrc}`
- I didn't bother to remove the `nixbld[0-9]` users (oops?)

[...]

Issue #458, opened 2015-01-24
How to completely remove Nix from OS X


chamini2:
Hi, I want to do a *clean* install of Nix on my computer, but I can't find how to uninstall it. In the manual, in chapter 4, it states the following:

> Nix can be uninstalled using rpm -e nix or dpkg -r nix on RPM- and Dpkg-based systems, respectively. After this you should manually remove the Nix store and other auxiliary data, if desired:
> ~~~
> $ rm -rf /nix
> ~~~

But I'm on OS X. What steps should I follow?

wmertens:
Rm -rf /nix

caius:
Just an extra note for anyone finding this issue via google in future, to completely remove nix on OS X, I found I had to tidy up a couple of (now broken) symlinks in my home folder with `rm -rf ~/.nix*` as well. (After doing `sudo rm -rf /nix` that is.)

lynnard:
Also don't forget to remove that `. $HOME/.nix-profile/etc/profile.d/nix.sh` from .profile.

Cireo:
You may also have persistent information in `/etc/nix/nix.conf` (http://nixos.org/nix/manual/#ch-files) that might taint a new install.

copumpkin:
These tidbits should be added to the manual. Anyone want to submit a PR?

JamesMcMahon:
On multi user systems you may want to check out the additional instructions at https://github.com/NixOS/nix/issues/1402.

Instead of the manual it would be great to have an uninstall script that took care of everything.

toraritte:
A probably naïve idea: what if the install script would generate an uninstall script and (try to) restore all that has been touched? 

As far as I can tell, https://nixos.org/nix/install is generated from https://github.com/NixOS/nix/blob/master/scripts/install.in (somehow).

tscholak:
From the multiuser installation script:
```
Uninstalling nix:
1. Delete the files Nix added to your system:

  sudo rm -rf /etc/nix /nix /var/root/.nix-profile /var/root/.nix-defexpr /var/root/.nix-channels /Users/tscholak/.nix-profile /Users/tscholak/.nix-defexpr /Users/tscholak/.nix-channels

and that is it.
```

expelledboy:
I had a crack at writing an uninstall script for [OSx](https://gist.github.com/expelledboy/c00aebb004b178cf78b2c9b344526ff6). Hope someone finds this helpful!

PaulRBerg:
[...]

PR #1624, opened 2017-10-20
Add 'Uninstalling Nix' section to the documentation

### LINKED ISSUE #1623: Add uninstall information to the manual
The uninstall information in the manual is currently both incomplete and hard to find. I think we should add a separate section to describe the steps necessary to perform a complete uninstall.

Related issues: #1608, #1551. Also discussed in #1402 and #458.

Munksgaard:
The purpose of this PR is to make it easier to figure out how to uninstall Nix. It's very likely that I'm missing some details, especially with regards to multi-user installs, but I think this PR should serve as a good starting point. It condenses information from #458 and #1402 and adds a new section in the manual.

Fixes #1623

### DIFF
--- a/doc/manual/installation/uninstalling-nix.xml
+++ b/doc/manual/installation/uninstalling-nix.xml
@@ -1,0 +1,56 @@
+<chapter xmlns="http://docbook.org/ns/docbook"
+      xmlns:xlink="http://www.w3.org/1999/xlink"
+      xmlns:xi="http://www.w3.org/2001/XInclude"
+      version="5.0"
+      xml:id="ch-uninstalling-nix">
+
+<title>Uninstalling Nix</title>
+
+<para>The following steps describes how to remove Nix from your system.</para>
+
+<para>First of all you need to remove the <filename>/nix</filename>
+directory:</para>
+
+<screen>sudo rm -rf /nix</screen>
+
+<para>You also need to remove some files from your home directory:</para>
+
+<screen>
+rm -rf $HOME/.nix-profile
+rm -rf $HOME/.nix-defexpr
+rm -rf $HOME/.nix-channels
+rm -rf $HOME/.config/nixpkgs
+</screen>
+
+<para>Then you need to remove the line that references Nix from your
+<filename><envar>$HOME</envar>/.profile</filename>. Assuming that your username
+is <literal>user</literal> it should look something like this:</para>
+
+<screen>
+if [ -e /home/user/.nix-profile/etc/profile.d/nix.sh ]; then . /home/user/.nix-profile/etc/profile.d/nix.sh; fi # added by Nix installer
+</screen>
+
+<para>There might also be some configuration files in
+<filename>/etc/nix</filename> that you need to delete:</para>
+
+<screen>
+sudo rm -rf /etc/nix
+</screen>
+
+<para>Finally, on a multi-user installation of Nix, you will also need to remove
+the <literal>nixbld[0-9]</literal> users.</para>
+
+<screen>
+sudo deluser nixbld0
+sudo deluser nixbld1
+sudo deluser nixbld2
+sudo deluser nixbld3
+sudo deluser nixbld4
+sudo deluser nixbld5
+sudo deluser nixbld6
+sudo deluser nixbld7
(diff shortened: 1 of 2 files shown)


Munksgaard:
Note that `nix-build release.nix -A build.x86_64-linux` fails on my local machine at the moment (with [this output](https://pastebin.com/vYvSFzJq)), so someone should probably make sure it works.

doc/manual/installation/uninstalling-nix.xml
```diff
@@ -0,0 +1,56 @@
...
+
+<screen>sudo rm -rf /nix</screen>
+
+<para>You also need to remove some files from your home directory:</para>
+
+<screen>
+rm -rf $HOME/.nix-profile
+rm -rf $HOME/.nix-defexpr
+rm -rf $HOME/.nix-channels
+rm -rf $HOME/.config/nixpkgs
```
edolstra: These files do not need to be removed.

doc/manual/installation/uninstalling-nix.xml
```diff
@@ -0,0 +1,56 @@
...
+
+<screen>
+if [ -e /home/user/.nix-profile/etc/profile.d/nix.sh ]; then . /home/user/.nix-profile/etc/profile.d/nix.sh; fi # added by Nix installer
+</screen>
+
+<para>There might also be some configuration files in
+<filename>/etc/nix</filename> that you need to delete:</para>
+
+<screen>
+sudo rm -rf /etc/nix
```
edolstra: Likewise, these files do not need to be deleted.

doc/manual/installation/uninstalling-nix.xml
```diff
@@ -0,0 +1,56 @@
...
+      xml:id="ch-uninstalling-nix">
+
+<title>Uninstalling Nix</title>
+
+<para>The following steps describes how to remove Nix from your system.</para>
+
+<para>First of all you need to remove the <filename>/nix</filename>
+directory:</para>
+
+<screen>sudo rm -rf /nix</screen>
```
edolstra: This and removing the sourcing of nix-profile.sh are really the only steps that are required.

doc/manual/installation/uninstalling-nix.xml
```diff
@@ -0,0 +1,56 @@
...
+
+<screen>
+sudo rm -rf /etc/nix
+</screen>
+
+<para>Finally, on a multi-user installation of Nix, you will also need to remove
+the <literal>nixbld[0-9]</literal> users.</para>
+
+<screen>
+sudo deluser nixbld0
```
edolstra: This is system-specific. On NixOS (and I think most Linux distributions) the command is `userdel`. Also it can be more elegantly expressed using a for loop, e.g. `for $((i = 0; i < 10; i++)); do deluser nixbld$i; done`.

doc/manual/installation/uninstalling-nix.xml
Munksgaard: But shouldn't they be? If I was testing out some new software on my machine, I wouldn't want it to leave behind unnecessary files or directories. Perhaps we could move these commands into a sub-section describing how to completely purge Nix from your system?

doc/manual/installation/uninstalling-nix.xml
Munksgaard: Good catch, I'll update the PR.

Munksgaard: Although, strictly speaking, that for loop is also system-specific, isn't it? It doesn't work in my shell (fish).

doc/manual/installation/uninstalling-nix.xml
SimSez: How does one remove the nix-profile.sh on osx, how do I locate it from the terminal (if that's what I am using)

doc/manual/installation/uninstalling-nix.xml
FRidh: It may be better to have a section which lists all files that are (typically) existing/generated by Nix and what their purpose is. Based on that a reader can easily decide what they need to remove.

Issue #833, opened 2016-03-04
builtins.readFile should preserve path context


abbradar:
If you do `builtins.readFile "${drv}/foo"`, `"${drv}/foo"`'s context would be lost -- that is, this file can reflect to something in `drv` or its inputs but Nix would lose this information. I've stumbled upon this in a very nasty way:
1. Have a service configuration option, `service.foo.config`, `type = types.str`.
2. Have its configuration in a separate file which is pre-processed by `substituteAll` adding references to several other derivations.
3. Use `builtins.readFile (pkgs.substituteAll { src = ./foo; bar = ./bar; })`.
4. After building this and running garbage collection `bar` would be gone from Nix store.

A simple test case is:
1. foo.nix

```
{ pkgs ? import <nixpkgs> {} }:

pkgs.writeText "test" (builtins.readFile (pkgs.substituteAll {
  src = ./foo;
  bar = ./bar;
}))
```
1. foo

```
@bar@
```
1. bar

```
hello!
```

After running `nix-build foo.nix` and `nix-collect-garbage` then, observe error when trying to do `cat $(cat ./result)`

abbradar:
Should this be reopened now that https://github.com/NixOS/nix/commit/4e6a2fbc561f3e692a5644b3816a205ba769f39e and https://github.com/NixOS/nix/commit/fa738e50bcd79e5e536d7453f5571e2b13b01bda landed?

abbradar:
I have just got a report from user who stumbled spectacularly upon this -- he uses similar code to one in the OP to generate a piece of configuration for Postfix server. Because the file was garbage collected all mail directed to this server was bounced away by Postfix for nearly two days!

Can we try to debug Hydra issue instead? Can it be reproduced locally?

PR #1643, opened 2017-10-30
Preserve file context across readFile, second take

abbradar:
This rebases old patches to fix https://github.com/NixOS/nix/issues/833 and fixes Hydra test failure too. Closes https://github.com/NixOS/nix/issues/833.

cc @shlevy

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1470,7 +1470,11 @@ static void prim_readFile(EvalState & state, const Pos & pos, Value * * args, Va
     string s = readFile(path);
     if (s.find((char) 0) != string::npos)
         throw Error("the contents of the file '%1%' cannot be represented as a Nix string", path);
-    v.mkString(s);
+    auto refs = state.store->isInStore(path) ?
+        state.store->queryPathInfo(state.store->toStorePath(path).first)->references :
+        StorePathSet{};
+    auto context = state.store->printStorePathSet(refs);
+    v.mkString(s, context);
 }
 
 static RegisterPrimOp primop_readFile({
--- a/tests/local.mk
+++ b/tests/local.mk
@@ -59,7 +59,8 @@ nix_tests = \
   ca/recursive.sh \
   ca/concurrent-builds.sh \
   ca/nix-copy.sh \
-  eval-store.sh
+  eval-store.sh \
+  readfile-context.sh
   # parallel.sh
 
 ifeq ($(HAVE_LIBCPUID), 1)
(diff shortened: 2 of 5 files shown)


shlevy:
LGTM. I'll set up a hydra jobset

shlevy:
https://hydra.nixos.org/jobset/nix/PR-1643

domenkozar:
Could we have a test? :)

abbradar:
@domenkozar Sorry for taking a long time, I've added a test!

abbradar:
I've rebased the patches but for some reason `nix-copy-closure` test [fails](https://hydra.nixos.org/build/63104473). Any ideas why? I'd very much like to see this fixed in Nix 2.0.

shlevy:
@abbradar That's a bug in the test that I thought @edolstra had fixed... @edolstra should that block merging this?

edolstra:
@abbradar You need the `nix-2.0` branch of Nixpkgs.

Don't know if changing the behaviour of `readFile` just before release is a good idea though.

abbradar:
@edolstra I see this as a bug because it can silently break users' systems after `nix-collect-garbage` is they have fancy enough configuration ([example](https://github.com/NixOS/nix/issues/833#issuecomment-305802085)). If I understand correctly this fix (in theory) just adds more paths into closure of some outputs so it should be fairly harmless. However I don't feel qualified to say this after I have already broken things with an older fix ;)

shlevy:
@edolstra I'd really prefer we branch off 2.0 if we're going to start delaying PRs because they shouldn't be in 2.0... That way we can keep moving despite being feature frozen (or perhaps feature chilled)

abbradar:
Can we proceed with this now that 2.0 release has been branched?

abbradar:
So, any movement here? I've run into this bug yet again, it's pretty annoying.

abbradar:
I have rebased this and resolved conflicts. Not sure if using `printStorePathSet` [here](https://github.com/NixOS/nix/pull/1643/commits/c66865dff1747ecd55915ef8c549dc1a088ff168#diff-0f59bb6f197822ef9f19ceae9624989499d170c84dfdc1f486a8959bb4588cafR1476) is a good idea.

I understand that life and other priorities can get in the way, but it still feels frustrating to keep rebasing this patchset for so many years without any movement. Maybe now that 2.0 is released long time ago we can move forward with this?

edolstra:
The risk with this PR is that it can cause evaluation to produce a different result depending on whether the input is in the Nix store (where it could have references). In the past we had issues like that with Hydra (which copies sources to the Nix store). However, it should be okay so long as the source tree has no references.

abbradar:
Thanks! Let's hope this won't break corner cases for people like it did before @shlevy helped fix it. If anything arises, ping me.

PR #1645, opened 2017-10-30
Fix (highly unlikely) race condition in readLink

twhitehead:
Used to determine symlink size with stat and value with readlink.  This could technically result in garbage if symlink changed between calls.  Also gets around the broken stat implementation in our network filesystem (returns size + 1 giving a byte of garbage).

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -216,18 +216,17 @@ bool pathExists(const Path & path)
 Path readLink(const Path & path)
 {
     checkInterrupt();
-    struct stat st = lstat(path);
-    if (!S_ISLNK(st.st_mode))
-        throw Error(format("'%1%' is not a symlink") % path);
-    auto bufSize = std::max(st.st_size, (off_t) PATH_MAX + 1);
-    char buf[bufSize];
-    ssize_t rlsize = readlink(path.c_str(), buf, bufSize);
-    if (rlsize == -1)
-        throw SysError(format("reading symbolic link '%1%'") % path);
-    else if (rlsize > bufSize)
-        throw Error(format("symbolic link '%1%' size overflow %2% > %3%")
-            % path % rlsize % bufSize);
-    return string(buf, rlsize);
+    for (ssize_t bufSize = PATH_MAX/4; true; bufSize += bufSize/2) {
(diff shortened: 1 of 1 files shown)


copumpkin:
Isn't this case the motivation for `readlinkat`?

twhitehead:
I'm not sure I understand what you are meaning?

From reading the man page of `readlinkat` it seems it's difference is that you can provide a directory file descriptor from which to interpreted a passed relative path from instead of a implicitly relative to the processes working directory.

The race condition I was referring to was that, in the original code, the length of the symlink was determined by a call to `stat` while the contents of the symlink was determined by a later call to `readlink`.  This can go wrong as things can technically change between the two calls.

The value used for the length of the returned symlink should be the value that `readlink` returns, as it is the one that also provides the symlink contents, not the earlier `stat` value.

Cheers!  -Tyson

PS:  My motivation for this change isn't so much that I'm worried about this race condition, although it should be fixed as it is technically wrong and can be, but that the stat call on our (vendor supplied) filesystem is broken and returns the length +1 which screws up the current code.

copumpkin:
Oh, I see, sorry. Ignore me!

dezgeg:
Since we're allocating the temporary value on the stack anyway, I would just directly allocate a `PATH_MAX` sized buffer.

twhitehead:
Apparently PATH_MAX is a bit of a dubious thing

http://www.gnu.org/software/libc/manual/html_node/Limits-for-Files.html

and isn't even defined on some OSs.

orivej:
Golang implements readlink as in this PR: https://sourcegraph.com/github.com/golang/go@go1.9.2/-/blob/src/os/file_posix.go#L16-29

twhitehead:
@orivej  Interesting.  Looks exactly like what I did except they start with a length of 128 and do 2x on failure while I started with 1024 (PATH_MAX/4) and do 1.5x on failure.

Perhaps picking a value like 128 is better though as on some systems PATH_MAX is not defined.

According to `git blame`, the code originally just used the `stat` size, but was later amended to use `max(stat size,PATH_MAX)` as links in */proc* returned a `stat` size of 0.

PR #1646, opened 2017-10-30
Allow optional localhost network access to sandboxed derivations

copumpkin:
This will allow bind and connect to 127.0.0.1, which can reduce purity/security (if you're running a vulnerable service on localhost) but is also needed for a ton of test suites, so I'm leaving it turned off by default but allowing certain derivations to turn it on as needed.

It also allows DNS resolution of arbitrary hostnames but I haven't found a way to avoid that. In principle I'd just want to allow resolving `localhost` but that doesn't seem to be possible.

I don't think this belongs under `build-use-sandbox = relaxed` because we want it on Hydra and I don't think it's the end of the world.

cc @edolstra @domenkozar @shlevy

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2774,10 +2774,10 @@ void DerivationGoal::runChild()
                     sandboxProfile += "(deny default (with no-log))\n";
                 }
 
-                sandboxProfile += "(import \"sandbox-defaults.sb\")";
+                sandboxProfile += "(import \"sandbox-defaults.sb\")\n";
 
                 if (fixedOutput)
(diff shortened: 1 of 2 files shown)


copumpkin:
Whoops, bug, please don't merge

copumpkin:
Okay, this is now safe to merge

shlevy:
We can't use network namespaces? :(

shlevy:
Oh, darwin only

copumpkin:
If only... 😄 Linux doesn't need to worry about this crap because it already has network namespaces

PR #1649, opened 2017-10-31
search.cc: sort attribute names with `std::map`

Ma27:


edolstra:
Maybe they should be sorted by relevance, e.g. matches in the attribute name having a greater weight than matches in the description.

Ma27:
@edolstra so you mean we should put it all into some kind of `std::vector<SearchResult>` and build a custom sort behavior which weights descriptions **and** attribute names (I only care about attribute names at the moment)?

Ma27:
@edolstra I should've some time in the next days, do you agree with the approach I described in https://github.com/NixOS/nix/pull/1649#issuecomment-340811054 for would you suggest something else? :)

Ma27:
@edolstra I'll resolve the merge conflict tonight.

However it would be cool if you could describe what exactly you would change in the sort behavior before merging it as I'd really like to get it ready for Nix 1.12 :)

shlevy:
@NixOS/nix-core (doesn't exist yet, but a man can dream) @edolstra this looks reasonable to me, can we get this in and think about better sorting algorithms separately?

edolstra:
Yup.

Ma27:
@edolstra @shlevy do you agree with @lheckemann regarding the sort by attribute path thing? In that case I'd submit another PR :)

PR #1650, opened 2017-10-31
Always allow builds to use unix domain sockets in Darwin sandbox

copumpkin:
No reason not to allow this, and quite a few test suites use unix sockets

### DIFF
--- a/src/libstore/sandbox-defaults.sb
+++ b/src/libstore/sandbox-defaults.sb
@@ -22,7 +22,14 @@
 (allow signal (target same-sandbox))
 
 ; Access to /tmp.
-(allow file* process-exec (literal "/tmp") (subpath TMPDIR))
+; The network-outbound/network-inbound ones are for unix domain sockets, which
+; we allow access to in TMPDIR (but if we allow them more broadly, you could in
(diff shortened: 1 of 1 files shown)


copumpkin:
cc @edolstra @LnL7

src/libstore/sandbox-defaults.sb
```diff
@@ -22,7 +22,14 @@
...
 
 ; Access to /tmp.
-(allow file* process-exec (literal "/tmp") (subpath TMPDIR))
+; The network-outbound/network-inbound ones are for unix domain sockets, which
+; we allow access to in TMPDIR (but if we allow them more broadly, you could in
+; theory escape the sandbox)
+(allow file* process-exec network-outbound network-inbound
+       (literal "/tmp") (subpath TMPDIR))
+
+; Always allow unix domain sockets, since they can't hurt purity or security
```
copumpkin: Whoops, not sure where this came from

edolstra: I removed it.

copumpkin: Thanks!

PR #1651, opened 2017-10-31
Allow getpwuid in the darwin sandbox

LnL7:
/cc @copumpkin


LnL7:
@edolstra I removed the parameter.

LnL7:
Thanks!

Issue #228, opened 2014-03-15
Inline docstring support


offlinehacker:
Is there any possibility that nix syntax could be extended and inline docstring support associated with functions would be added like in python? This would be really useful for documentation generation.

qknight:
i would like to **implement this** if someone can **sponsor the effort**!

would be a great to have autogenerated documentation in assistant, see:
  http://blog.lastlog.de/posts/nix-assistant/

(if interested, just contact me via email: js@lastlog.de)

greedy:
I've been thinking about this as well, but was thinking more about doccomments.

I've also been thinking about the possibility of using functors for documented functions so that the documentation is actually part of the value. This makes it easier to deal with the fact that the attribute path used to access a function can't be easily determined. The documentation tool could evaluate stdenv (or any other set) and directly get the documentation for all of its function attributes. Non-function values would have to work some other way though, hmm.

Ericson2314:
I thought https://github.com/tweag/nickel had something along the lines of docstrings.

thiagokokada:
I think this is still important. Having access to `:doc builtins.filter` inside the `nix repl` already helps, but I would like this to be extended to any function.

edolstra:
@thiagokokada See #5527 for a possible way to do this.

thiagokokada:
> @thiagokokada See #5527 for a possible way to do this.

Nice. Hope that we can eventually adopt this on `nixpkgs`.

mightyiam:
I, @101v and @a-kenji are currently working on _a design_ for in-code documentation of Nix values.

This issue seems to be the correct place for providing updates on that work. So, subscribe to this one, if you're curious or are willing to provide feedback and/or answer our occasional question.

Also, we seek funding for this activity. I've contacted NLnet to that end and haven't heard back yet.

piegamesde:
Where does the discussion take place? Can everybody who is interested participate?

(Also, depending on the size of the feature prepare yourselves for going through the RFC process)

mightyiam:
> Where does the discussion take place? Can everybody who is interested participate?

We work in mob programming format on a regular schedule. Is this something you'd like to try participating in?

piegamesde:
No, I'm only available for async discussion.

blaggacao:
@mightyiam already got an idea about schema? Making unit tests part of the doc (based on examples)? Some ideas can also be observed in `styx`.

mightyiam:
[...]

PR #1652, opened 2017-10-31
nix repl: Provide documentation from comment when evaluating to lambda

roberth:
This provides limited support for python-like docstrings in the `nix repl`. When the user evaluates an expression to a lambda, `nix repl` will now print the contents of a documentation comment, as long the comment written right before the attribute, and the attribute value is an actual lambda.

Current limitations:
 - Overriding the documentation requires redundant lambdas (eta abstraction to be precise)
 - No docstring support for anything other than functions

Demo:

```
nix-repl> pkgs.lib.concatMapStringsSep ":"
«lambda @ /nix/store/ksi72626r14035xkndbn584pmb7l703r-nixos-17.09.1535.1fdca25ee8/nixos/lib/strings.nix:64:30»

| concatMapStringsSep
| -------------------
| 
| NOTE: This function has already been applied!
|       You should ignore the first 1 parameter(s) in this documentation,
|       because they have already been applied.
|
| First maps over the list and then concatenates it.
| 
| Example:
|    concatMapStringsSep "-" (x: toUpper x)  ["foo" "bar" "baz"]
|    => "FOO-BAR-BAZ"


nix-repl>
```

Changes:
 - A new test set, `nix-repl.sh`
 - A new module, `comment.cc` for the documentation retrieval logic
 - Additions to `repl.cc`

Todo:
 - [ ] choose a comment syntax
 - [ ] hide the documentation behind a hint that `:doc` displays the documentation for the function

### DIFF
--- a/src/nix/comment.cc
+++ b/src/nix/comment.cc
@@ -1,0 +1,231 @@
+#include <iostream>
+#include <fstream>
+#include <sstream>
+#include <regex>
+#include <climits>
+#include <algorithm>
+
+#include "comment.hh"
+#include "util.hh"
+
+/* This module looks for documentation comments in the source code.
+
+   Documentation is not retained during parsing, and it should not be,
+   for performance reasons. Because of this the code has to jump
+   through some hoops, to perform its task.
+
+   Adapting the parser was not considered an option, so this code
+   parses the comments from scratch, using regular expressions. These
+   do not support all syntactic constructs, so in rare cases, they
+   will fail and the code will report no documentation.
+
+   One such situation is where documentation is requested for a
+   partially applied function, where the outer lambda pattern
+   matches an attribute set. This is not supported in the regexes
+   because it potentially requires (almost?) the entire grammar.
+
+   This module has been designed not to report the wrong
+   documentation; considering that the wrong documentation is worse
+   than no documentation. The regular expressions will only match
+   simple, well understood syntactic structures, or not match at all.
+
+   This approach to finding documentation does not cause extra runtime
+   overhead, until used.
+
+   This module does not support tab ('\t') characters. In some places
+   they are treated as single spaces. They should be avoided.
+*/
+namespace nix::Comment {
+
+struct Doc emptyDoc("", "", "", 0);
+
+/* parseDoc will try to recover a Doc by looking at the text that leads up to a term
+   definition.*/
+static struct Doc parseDoc(std::string sourcePrefix);
+
+/* stripComment unpacks a comment, by unindenting and stripping " * " prefixes as
+   applicable. The argument should include any preceding whitespace. */
+static std::string stripComment(std::string rawComment);
+
+/* Consistent unindenting. It will only remove entire columns. */
+static std::string unindent(std::string s);
+
+static std::string trimUnindent(std::string s) {
+    return trim(unindent(s));
+}
+
+static std::string stripPrefix(std::string prefix, std::string s) {
+    std::string::size_type index = s.find(prefix);
+    return (index == 0) ? s.erase(0, prefix.length()) : s;
+}
+
+static std::string readFileUpToPos(const Pos & pos) {
+
+    std::ifstream ifs(static_cast<const std::string>(pos.file));
+    std::stringstream ret;
+    size_t lineNum = 1;
+    std::string line;
+
+    while (getline(ifs, line) && lineNum <= pos.line) {
+        if (lineNum < pos.line) {
+            ret << line << "\n";
+        } else if (lineNum == pos.line) {
+            ret << line.substr(0, pos.column-1);
+        }
+        lineNum++;
+    }
+
+    return ret.str();
+}
+
+struct Doc lookupDoc(const Pos & pos) {
+    try {
+        return parseDoc(readFileUpToPos(pos));
+    } catch (std::exception e) {
+        ignoreException();
+        return emptyDoc;
+    }
+}
+
+/* See lambdas in parseDoc */
+static int countLambdas(std::string piece) {
+    return std::count(piece.begin(), piece.end(), ':');
+}
+
+/* Try to recover a Doc by looking at the text that leads up to a term
+   definition */
+static struct Doc parseDoc(std::string sourcePrefix) {
+
+    std::string wss("[ \t\r\n]*");
+    std::string spaces("[ \t]*");
+
+    std::string singleLineComment(spaces + "#[^\r\n]*(?:\n|\r\n)");
+    std::string multiSingleLineComment("(?:" + singleLineComment + ")*");
+    std::string multiLineComment("\\/\\*(?:[^*]|\\*+[^*/])*\\*+\\/");
+    std::string commentUnit("(" + multiSingleLineComment + "|" + spaces + multiLineComment + ")" + wss);
+
+    std::string ident("[a-zA-Z_][a-zA-Z0-9_'-]*" + wss);
+    std::string identKeep("([a-zA-Z_][a-zA-Z0-9_'-]*)" + wss);
+
+    /* lvalue for nested attrset construction, but not matching
+       quoted identifiers or ${...} or comments inbetween etc */
+    std::string simplePath("(?:" + wss + ident + "\\.)*" + identKeep);
+
+    std::string lambda(ident + wss + ":" + wss);
+
+    /* see countLambdas */
+    std::string lambdas("((:?" + lambda + ")*)");
+
+    std::string assign("=" + wss);
+
+    std::string re(commentUnit + simplePath + assign + lambdas + "$");
+    std::regex e(re);
+
+    #define REGEX_GROUP_COMMENT 1
+    #define REGEX_GROUP_NAME 2
+    #define REGEX_GROUP_LAMBDAS 3
+    #define REGEX_GROUP_MAX 4
+
+    std::smatch matches;
+    regex_search(sourcePrefix, matches, e);
+
+    std::stringstream buffer;
+    if (matches.length() < REGEX_GROUP_MAX) {
+        return emptyDoc;
+    }
+
+    std::string rawComment = matches[REGEX_GROUP_COMMENT];
+    std::string name = matches[REGEX_GROUP_NAME];
+    int timesApplied = countLambdas(matches[REGEX_GROUP_LAMBDAS]);
+    return Doc(rawComment, stripComment(rawComment), name, timesApplied);
+}
+
+static std::string stripComment(std::string rawComment) {
+    rawComment.erase(rawComment.find_last_not_of("\n")+1);
+
+    std::string s(trimUnindent(rawComment));
+
+    if (s[0] == '/' && s[1] == '*') {
+        // Remove the "/*"
+        // Indentation will be removed consistently later on
+        s[0] = ' ';
+        s[1] = ' ';
+
+        // Remove the "*/"
+        if (!s.empty() && *(--s.end()) == '/')
+            s.pop_back();
+        if (!s.empty() && *(--s.end()) == '*')
+            s.pop_back();
+
+        s = trimUnindent(s);
+
+        std::istringstream inStream(s);
+        std::ostringstream stripped;
+
+        std::string line;
+
+        /* at first, we assume a comment
+         * that is formatted like this
+         * with '*' characters at the beginning
+         * of the line.
+         */
+        bool hasStars = true;
+
+        while(std::getline(inStream,line,'\n')){
+            if (hasStars && (
+                    (!line.empty() && line[0] == '*')
+                 || (line.length() >= 2 && line[0] == ' ' && line[1] == '*')
+                    )) {
+                if (line[0] == ' ') {
+                    line = stripPrefix(" *", line);
+                } else {
+                    line = stripPrefix("*", line);
+                }
+            } else {
+                hasStars = false;
+            }
+
+            stripped << line << std::endl;
+        }
+        return trimUnindent(stripped.str());
+    }
+    else {
+        std::istringstream inStream(s);
+        std::ostringstream stripped;
+
+        std::string line;
+        while(std::getline(inStream, line, '\n')) {
+            line.erase(0, line.find("#") + 1);
+            stripped << line << std::endl;
+        }
+        return trimUnindent(stripped.str());
+    }
+}
+
+static std::string unindent(std::string s) {
+    size_t maxIndent = 1000;
+    {
+        std::istringstream inStream(s);
+        for (std::string line; std::getline(inStream, line); ) {
+            size_t firstNonWS = line.find_first_not_of(" \t\r\n");
+            if (firstNonWS != std::string::npos) {
+                maxIndent = std::min(firstNonWS, maxIndent);
+            }
+        }
+    }
+
+    std::ostringstream unindentedStream;
+    {
+        std::istringstream inStream(s);
+        for (std::string line; std::getline(inStream, line); ) {
+            if (line.length() >= maxIndent) {
+                unindentedStream << line.substr(maxIndent) << std::endl;
+            } else {
+                unindentedStream << std::endl;
+            }
+        }
+    }
+    return unindentedStream.str();
+}
+
+}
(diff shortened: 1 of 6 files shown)


regnat:
Really nice improvement :)

It may be more natural to have this printed only with the `:t` command (or maybe a new `:doc` one) rather than when evaluating, what do you think of this?

grahamc:
!!!

gilligan:
Oops.. i didn't realize the PR was already open and commented on your fork instead.. well you can see it anyway ;)

src/nix/comment.cc
```diff
@@ -0,0 +1,224 @@
...
+    regex_search(sourcePrefix, matches, e);
+
+    std::stringstream buffer;
+    if (matches.length() < 3) {
+        return emptyDoc;
+    }
+
+    std::string rawComment = matches[1];
+    std::string name = matches[2];
+    int timesApplied = countLambdas(matches[3]);
```
gilligan: I think matches could be of length 3 here and you are accessing the 4th element? Since you check for < 3 in line 126..

src/nix/comment.cc
```diff
@@ -0,0 +1,224 @@
...
+
+// Consistent unindenting. It will only remove entire columns.
+std::string unindent(std::string s);
+
+std::string trimUnindent(std::string s) {
+    return trim(unindent(s));
+}
+
+// SLOW, probably O(n^2)
+std::string stripPrefix(std::string prefix, std::string s) {
```
gilligan: How about something like this:

```c++
std::string stripPrefix(std::string prefix, std::string s) {
  std::string::size_type index = s.find(prefix);
  return (index == 0) ? s.erase(0, prefix.length()) : s;
}
```

Untested but I guess that should work?! ;-)

src/nix/comment.cc
```diff
@@ -0,0 +1,224 @@
...
+
+    std::string assign("=" + wss);
+
+    std::string re(commentUnit + simplePath + assign + lambdas + "$");
+    std::regex e(re);
+    std::smatch matches;
+    regex_search(sourcePrefix, matches, e);
+
+    std::stringstream buffer;
+    if (matches.length() < 3) {
```
gilligan: I really don't know what kind of standards the nix code base has in general.. i usually try to avoid "magic numbers" and would introduce some `#define` here or so.. same for the indices to access matches below. Just a thought.

gilligan:
@roberth I wonder if this should perhaps restricted to not just any kind of `/* .. */` delimited block but that we instead introduce some marker ?

```nix
{
  /*nixdoc
  Proofs that P = NP
  */
  proofThingy = p: np: 42;
}
```
Otherwise we might end up with arbitrary things like `TODO: rewrite this` or `FIXME` or whatever else. Since we (NixOsDocs folks) are starting off with improving nix functions documentation we could pick this up from the get go.

This would definitely also help for automatically extracting docstrings like these from nixpkgs. In fact - maybe we could go even further and use `nixdoc:<function>` (So `nixdoc:proofThingy` for the example above). That would make it really easy to create things like ctags files for navigating nixpkgs. I would *love* that.

roberth:
@gilligan , thank you for your review. I have implemented all of your suggestions except the `nixdoc:`.

I'm also a bit concerned about the ugly comments that might show up. I'm not sure what the special syntax should be. `nixdoc:myfunc` will inevitably lead to copy paste errors and bitrot. A 'small syntax' like `/** */` might be better than `nixdoc`, but that leaves no room for metadata in the comment that isn't normally shown...

gilligan:
@roberth right, having the function name duplicated in the comment might easily go out of sync. Then again: doesn’t that also apply to any kind of comment written for any function anyway?

So the function documentation can *always* be wrong but if we have some tag we could maybe limit the output to the “right” kind of comment blocks?

In short: IMHO looking for “/** nixdoc” might be preferable.

roberth:
@gilligan I agree that we need some kind of marker to indicate that it's a documentation comment.
I think `/**` or `##` should be sufficient, but it's also nice to be forward compatible, so we can switch to RST/Markdown/whatever when we want to generate HTML docs. My suggestion:

     * `/**` followed by plain text, to be shown verbatim in the repl
     * `/**sometoken ` followed by text in some format

We can remove the `sometoken` in the repl, for forward compatibility. When we're going to generate HTML docs, we can decide on a format without causing breakage when switching. So in the future, docs may look like

    `/**md
     # Introduction

     This function bla bla
    ` */

What I like about this is that it is very clear that we're using markdown here, it will be shown in the repl without the `md` part, and can be processed in a future repl, knowing that it's the 'Markdown+Nix' format.

I'm not saying that we should do Markdown in the future, though. There's no standard for it. At least Restructured Text has a better specification (from what I've heard). Markdown is rather popular though.
 Let's not go for docbook at least :)

dtzWill:
Sounds good to me! Can't wait! ^_^

roberth:
Before the holidays, I put out a twitter poll about the syntax for documentation comments. The results:

It got 24 votes:

> 33% /** always plain text */  (8 people)
> 33% /** always markdown */
> 17% /**md to pick md format  (4 people)
> 17% ## let's discuss\n=======
> One person mentioned doxygen, but didn't gain support from others.
> On IRC, grahamc mentioned docbook

So, there is no clear winner, but a combination of `/** plain text` by default and `/**md to 'upgrade' to markdown` can be argued to have 50% of the votes (ignoring the possibility that some plain text voters hate markup languages at all cost)

If anyone is going to decide the color of this bike shed I think it should be @edolstra because I don't want to ruin his awesome creation with ugly comments.

edolstra:
I'm not in favor of Markdown.

roberth:
@edolstra good.

What about an extensible syntax marked by a keyword right after `/**`?
Or do you want to 'disallow' plain text (most comments now) and standardize on a single format? If so, which one? Some possibilities are restructured text (has a specification; python doc format), docbook and doxygen.

src/nix/comment.cc
```diff
@@ -0,0 +1,227 @@
...
+        for (size_t i = 0;
+             getline(ifs, line) && i < pos.line-1;
+             i++) {
+            buffer << line << "\n";
+        }
+        buffer << line.substr(0, pos.column-1);
+
+        return parseDoc(buffer.str());
+    } catch (std::exception e) {
+        std::cout << "Caught exception: " << e.what() << std::endl;
```
edolstra: => `ignoreException()`.

src/nix/comment.cc
```diff
@@ -0,0 +1,227 @@
...
+
+        // read up to line and column with unix newlines
+
+        std::string line;
+        for (size_t i = 0;
+             getline(ifs, line) && i < pos.line-1;
+             i++) {
+            buffer << line << "\n";
+        }
+        buffer << line.substr(0, pos.column-1);
```
edolstra: This can be simplified to something like `for (auto & line : tokenizeString(readFile(pos.file), "\n") { ... }`.

src/nix/comment.hh
```diff
@@ -0,0 +1,41 @@
...
+};
+
+extern struct Doc emptyDoc;
+
+// lookupDoc will try to recover a Doc. This will perform perform I/O,
+// because documentation is not retained by the parser.
+//
+// Will return empty values if nothing can be found.
+// For its limitations, see the docs of the implementation.
+struct Doc lookupDoc(Pos & pos);
```
edolstra: Should be `const Pos & pos`.

src/nix/comment.cc
```diff
@@ -0,0 +1,227 @@
...
+#include <fstream>
+#include <sstream>
+#include <regex>
+#include <climits>
+#include <algorithm>
+
+#include "comment.hh"
+#include "util.hh"
+
+// This module looks for documentation comments in the source code.
```
edolstra: We don't use `//` in the Nix code base except for single-line comments.

src/nix/comment.cc
roberth: `tokenizeString` didn't work because it doesn't give back the empty lines. I didn't find another suitable function in util either, so I have factored the thing out instead. The other review items are now solved.

roberth:
Before this can be merged we have to make decisions about syntax, because this code will return FIXME comments and such.

edolstra:
I'm not really fond of abusing comments for documentation. That's what you do when you don't control the language and don't have a way to add documentation in a more structured way (e.g. C++ and doxygen). But we have the freedom to add documentation in a "proper" way (e.g. as annotations).

roberth:
What makes use abuse? Aren't comments equally controllable?

By turning comments into annotations, you enable _static_ introspection, so that you don't have to burden the interpreter with it. Evaluation performance is already a problem and `__functor` has an overhead.

Adding documentation at the value level is what you do when you don't control the language tooling and don't have a way to add documentation in a way that works without having to evaluate code. After all, documentation is not a runtime concept but one that relates directly to the source code.

piegamesde:
I'm really in favor of this change (yay for more static and less evaluation, yay for less functors), but I really think we should have a dedicated team of people that discuss the syntax, independently of the implementation.

edolstra:
Shoving documentation into comments and then trying to correlate the comment with some nearby identifier is the sort of hack you do when you can't add documentation as first-class syntax (e.g. Python annotations). At the least, documentation comments should have some distinguishing syntax, since it's not obvious in 
```
  # Just a function.
  f = x: x;
```
that the comment `# Just a function` is intended as documentation. (E.g. it could also be `# FIXME: bla bla`.)

In Nix, documentation can *only* be done at the value level (as in the NixOS module system) because we don't have a proper module system. So we can't actually statically extract from a source file how the functions in that file are supposed to be called. E.g. we have no way to figure out from an attribute like
```
  /* Bla */
  concatMapStringsSep = ...;
```
that this defines a function that can be called as `lib.concatMapStringsSep`.

piegamesde:
Even if you want to do an implementation that is more value/run time based, there are alternative proposals that would not make it a functor-exclusive feature: https://github.com/NixOS/nix/pull/5527#issuecomment-967236780

blaggacao:
We need this (native doc support). A diversity of solutions isn't adding value any more at this point in time.

Ericson2314:
We should do this. We were also talking in the documentation team about need for better API docs *in the code* so they do not get out of sync CC @fricklerhandwerk @infiisil.

@edolstra I understand your concerns about hijacking comments being ugly, but I think it is a good place for iteration. Consider these examples:

1. In Rust, a `///` variation on comments is common place, thought there are other annotation options
2. Typescript, mypy, and other post-hoc type checkers evolved from special comments to dedicated syntax.

Basically, I think it will take a few rounds of experimentation to figure out exactly the system we want. The fact that we are so expression-oriented means that we cannot easily grab steal ideas from other languages with a "static top-level". If we go for dedicated syntax immediately, tie our hands with comparability concerns. Conversely, if we "steal" some comment syntax we can be sure that we are backwards compatible.

In this manner we can have a low-impact unstable feature will *both* allow us to start written *extremely important* API documentation immediately, and *also* give us plenty of flexibility to experiment with different variations. Once we decide what we want, we can introduce dedicated syntax, and only stabilize that. Existing API documentation (which is hopefully much more comprehensive at that point than it is today!) can then be transitioned over to using the new system.

This is the lowest risk plan which allows us to accelerate documentation contribution the soonest. Let's do it!

Ericson2314:
CC @mightyiam who was looking to work on this very problem!

mightyiam:
Thank you, @Ericson2314 .

Tracking issue seems to be https://github.com/NixOS/nix/issues/228#issuecomment-1277024527. Will provide updates there. If anyone is willing to answer some questions we may have, please subscribe to that issue.

fricklerhandwerk:
Discussed in the Nix team meeting 2023-01-20:

  - @fricklerhandwerk: there are multiple approaches to the broad problem of documenting Nix language code
  - @thufschmitt: while the implementation is small, it's a substantial change to the language
  - @roberth: don't have time to work on this
  - agreement: this should be an RFC
      - @NixOS/documentation-team should coordinate this

src/nix/comment.hh
```diff
@@ -0,0 +1,41 @@
...
+
+    // Number of times the curried function must be applied to get the value
+    // that this structure documents.
+    //
+    // This is useful when showing the documentation for a partially applied
+    // curried function. The documentation is for the unapplied function, so
+    // this is crucial information.
+    int timesApplied;
+
+    Doc(std::string rawComment, std::string comment, std::string name, int timesApplied) {
```
inclyc, line 23: would like to prefer the C++ way writing ctors
```suggestion
    Doc(std::string rawComment, std::string comment, std::string name, int timesApplied) : rawComment(std::move(rawComment), comment(std::move(comment)), name(std::move(name)), timesApplied(timesApplied) {
```

and remove `this-> ... = ...` assignments. 

We are using copy constructors for `std::string`, this is inefficient because C++ string are really "value-like" strings, not references, the buffer will be copied each time if we do

`this->name = name`

src/nix/comment.cc
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/* See lambdas in parseDoc */
+static int countLambdas(std::string piece) {
+    return std::count(piece.begin(), piece.end(), ':');
+}
+
+/* Try to recover a Doc by looking at the text that leads up to a term
+   definition */
+static struct Doc parseDoc(std::string sourcePrefix) {
```
inclyc, line 97: I'd rather like to preserve comment information in our lexer, not using regexes here. That is, we store pointers to comments here, and construct & strip indentation after parsing state.

https://github.com/NixOS/nix/blob/9428d7dcd1699266189873197565923872e0563f/src/libexpr/lexer.l#L305-L306

This has no overhead but we must carefully deal with the lifetime of these references/pointers.

I think we should do this in the lexer because it will be flexible for further changes & keep consistent with the lexer.

src/nix/comment.cc
```diff
@@ -0,0 +1,231 @@
...
+   than no documentation. The regular expressions will only match
+   simple, well understood syntactic structures, or not match at all.
+
+   This approach to finding documentation does not cause extra runtime
+   overhead, until used.
+
+   This module does not support tab ('\t') characters. In some places
+   they are treated as single spaces. They should be avoided.
+*/
+namespace nix::Comment {
```
inclyc, line 38:
```suggestion
namespace nix::comment {
```

nit pick (NFC)

src/nix/comment.cc
```diff
@@ -0,0 +1,231 @@
...
+    try {
+        return parseDoc(readFileUpToPos(pos));
+    } catch (std::exception e) {
+        ignoreException();
+        return emptyDoc;
+    }
+}
+
+/* See lambdas in parseDoc */
+static int countLambdas(std::string piece) {
```
inclyc, line 91:
```suggestion
static int countLambdas(const std::string &piece) {
```

src/nix/repl.cc
```diff
@@ -59,7 +61,11 @@ struct NixRepl
 
     typedef set<Value *> ValuesSeen;
     std::ostream &  printValue(std::ostream & str, Value & v, unsigned int maxDepth);
+    std::ostream &  printValueAndDoc(std::ostream & str, Value & v, unsigned int maxDepth);
     std::ostream &  printValue(std::ostream & str, Value & v, unsigned int maxDepth, ValuesSeen & seen);
+
+    // Only prints if a comment is found preceding the position.
```
inclyc, line 67:
```suggestion
    /// Only prints if a comment is found preceding the position.
```

src/nix/comment.hh
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include "nixexpr.hh"
+
+namespace nix::Comment {
+
+struct Doc {
+
+    // Name that the term is assigned to
```
inclyc, line 9:
```suggestion
    /// Name that the term is assigned to
```

src/nix/comment.cc
```diff
@@ -0,0 +1,231 @@
...
+}
+
+static std::string stripPrefix(std::string prefix, std::string s) {
+    std::string::size_type index = s.find(prefix);
+    return (index == 0) ? s.erase(0, prefix.length()) : s;
+}
+
+static std::string readFileUpToPos(const Pos & pos) {
+
+    std::ifstream ifs(static_cast<const std::string>(pos.file));
```
inclyc, line 64: can we read from the buffer instead? Because for static analysis tooling, actually there is no such file on the filesystem, this is a problem we deal with #6530

PR #1655, opened 2017-11-01
Don't freak out if we get a 403 from S3

copumpkin:
As far as we're concerned, not being able to access a file just means the file is missing. Plus, AWS explicitly goes out of its way to return a 403 if the file is missing and the requester doesn't have permission to list the bucket.

cc @edolstra @jammerful

### DIFF
--- a/src/libstore/s3-binary-cache-store.cc
+++ b/src/libstore/s3-binary-cache-store.cc
@@ -241,8 +241,8 @@ struct S3BinaryCacheStoreImpl : public S3BinaryCacheStore
             auto & error = res.GetError();
             if (error.GetErrorType() == Aws::S3::S3Errors::RESOURCE_NOT_FOUND
                 || error.GetErrorType() == Aws::S3::S3Errors::NO_SUCH_KEY
-                || (error.GetErrorType() == Aws::S3::S3Errors::UNKNOWN // FIXME
-                    && error.GetMessage().find("404") != std::string::npos))
+                // If bucket listing is disabled, 404s turn into 403s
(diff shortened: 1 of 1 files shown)


src/libstore/s3-binary-cache-store.cc
```diff
@@ -241,8 +241,11 @@ struct S3BinaryCacheStoreImpl : public S3BinaryCacheStore
             auto & error = res.GetError();
             if (error.GetErrorType() == Aws::S3::S3Errors::RESOURCE_NOT_FOUND
                 || error.GetErrorType() == Aws::S3::S3Errors::NO_SUCH_KEY
+                // If bucket listing is disabled, 404s turn into 403s
+                || error.GetErrorType() == Aws::S3::S3Errors::ACCESS_DENIED
                 || (error.GetErrorType() == Aws::S3::S3Errors::UNKNOWN // FIXME
-                    && error.GetMessage().find("404") != std::string::npos))
+                    && (error.GetMessage().find("404") != std::string::npos
+                    || error.GetMessage().find("403") != std::string::npos)))
```
edolstra: Is checking for the 403 string necessary?

copumpkin: Not really sure! I figured that you had some sort of weird edge case that led you to checking for `UNKNOWN` with 404, and had no confidence it wouldn't strike on a 403 too. I can remove it if you don't think it's necessary.

copumpkin:
@edolstra okay I got rid of the string checks. How's this?

copumpkin:
Thanks!

PR #1662, opened 2017-11-03
Remove errant `-r`s from `nix copy --help`

bgamari:
This option seems to be neither necessary nor supported. Presumably the documentation is simply out-of-date.


bgamari:
Ahem. Ignore this noise.

PR #1664, opened 2017-11-05
Setup NIX_PATH correctly in nix-profile-daemon

matthewbauer:
We need nixpkgs to be set in NIX_PATH for Nix 1.12 to work correctly.

### DIFF
--- a/scripts/nix-profile-daemon.sh.in
+++ b/scripts/nix-profile-daemon.sh.in
@@ -50,5 +50,5 @@ if test -w $HOME; then
 fi
 
 export NIX_SSL_CERT_FILE="@localstatedir@/nix/profiles/default/etc/ssl/certs/ca-bundle.crt"
-export NIX_PATH="@localstatedir@/nix/profiles/per-user/root/channels"
+export NIX_PATH="nixpkgs=@localstatedir@/nix/profiles/per-user/root/channels/nixpkgs:@localstatedir@/nix/profiles/per-user/root/channels"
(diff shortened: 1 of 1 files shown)


matthewbauer:
/cc @grahamc

Does this look like a good idea? The issue right now is Nix 1.12 will not recognize the path that’s setup by default (nixpkgs is not found) so something like:

nix build nixpkgs.hello

Will fail

FRidh:
What would happen if root would remove its channel?

matthewbauer:
> What would happen if root would remove its channel?

I think it should just give a warning like:

```
warning: Nix search path entry ‘asdf’ does not exist, ignoring
```

So if you have Nix 2.0 you can try this right now with the default value for NIX_PATH:

```
NIX_PATH=/nix/var/nix/profiles/per-user/root/channels nix run nixpkgs.hello
```

/cc @edolstra @grahamc

matthewbauer:
> What would happen if root would remove its channel?

I'm pretty sure Nix just falls back to previous path?

Also @edolstra Reopening this because some people have been hitting it. This should partially fix #2016. I think it should work okay- not sure if there are any objections.

grahamc:
This is the same I was thinking, but I think the rest of the patch is good.

matthewbauer:
Okay removed that line. Note that it is still in nix-profile.sh.in:

https://github.com/NixOS/nix/blob/master/scripts/nix-profile.sh.in#L57

LnL7:
This should only get sourced once anyway, if we export [__ETC_PROFILE_NIX_SOURCED](https://github.com/matthewbauer/nix/blob/d7a84d330cc677d3c9e2c5553a861ab9f82245f6/scripts/nix-profile-daemon.sh.in#L3) paths can be extended without introducing duplicates.

matthewbauer:
Pinging

PR #1678, opened 2017-11-14
anchor nix::Exit exception

dtzWill:
Fixes problem in my toolchain and generally helps compatibility
to anchor into a single place so there is only 1 typeinfo when throwing cross-DSO.


dtzWill:
`nm -A result/bin/*|c++filt|grep Exit` before and after shows the difference.

PR #1713, opened 2017-12-04
only send overridden settings to the remote store

LnL7:
Sending all settings breaks builders with another platform,
since that includes 'system' and other local settings like
'sandbox-paths' that might be different.

Explicit overrides like `nix-build --store ssh-ng://foo --option build-use-sandbox true` still work.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -221,6 +221,10 @@ let
       nix = build.x86_64-linux; system = "x86_64-linux";
     });
 
+    tests.sshStoreBuilds = (import ./tests/ssh-store-builds.nix rec {
+      nix = build.x86_64-linux; system = "x86_64-linux";
+    });
+
     tests.nix-copy-closure = (import ./tests/nix-copy-closure.nix rec {
       nix = build.x86_64-linux; system = "x86_64-linux";
     });
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -664,7 +664,8 @@ HookInstance::HookInstance()
     toHook.readSide = -1;
 
     sink = FdSink(toHook.writeSide.get());
-    for (auto & setting : settings.getSettings())
+    /* Send overridden settings. */
+    for (auto & setting : settings.getSettings(true))
         sink << 1 << setting.first << setting.second;
     sink << 0;
 }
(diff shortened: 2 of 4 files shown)


LnL7:
@edolstra ping

edolstra:
Isn't the real issue here that we shouldn't send options to ssh-ng? Only sending overriden settings is a bit ambiguous, e.g. should settings in the user's `~/.config/nix/nix.conf` be respected?

LnL7:
That's a very good point, however I must say that I've found the behaviour pretty useful. Let me check how this behaves with non trusted users.

If you think that's not a good idea I can close this and look at the other solution.

LnL7:
I forgot that this doesn't go through the local daemon. Overriding options only works if the local user has direct ssh access to a trusted user, just like a local client.

puffnfresh:
When I try to build using the ssh-ng remote builder, I get these errors:

```
warning: unknown setting 'allow-new-privileges'
warning: unknown setting 'filter-syscalls'
warning: unknown setting 'sandbox-build-dir'
warning: unknown setting 'sandbox-dev-shm-size'
```

Would this PR also fix these warnings?

puffnfresh:
Yeah, it fixes the warnings and allows using ssh-ng as a remote builder. @edolstra what's the more appropriate fix to make ssh-ng work as a remote builder?

LnL7:
Yeah, those are some platform specific settings.

This basically makes the client behave the same way with ssh-ng as with a local daemon, by only sending settings from the cli and the user’s nix.conf. Ignoring everything would also work but that kind of reverts the functionality from https://github.com/NixOS/nix/commit/37fbfffd8e23cf9ca4038e6c4145583a811e91aa, I’m not sure what implications that has.

LnL7:
Perhaps I should clarify, the main reason for this pr was to get build-remote to work with ssh-ng.

shlevy:
Copying comment from #nix-darwin:

> <shlevy> LnL: I feel like "overridden" is the wrong concept here  [09:05]
> <shlevy> Can you add an additional field to BaseSetting or similar to mark
> 	 whether it should be forwarded?

shlevy:
Possible fix at https://github.com/shlevy/nix/tree/forwardable-settings . Testing now.

PR #1717, opened 2017-12-08
upload-release.pl: support aarch64-linux

grahamc:
I haven't / can't really test this, but it'd be neat to support installing Nix on aarch64 via the website.


grahamc:
I can't seem to find this script in 1.11-maintenance, so I haven't backported. If I should backport it somewhere, I'd be happy to!

PR #1721, opened 2017-12-09
Escape left angle brackets in XML documentation

expipiplus1:


bhipple:
Thanks!  I noticed the other day that master fails to build because of this; glad to see someone else has already fixed it by the time I got back around to hacking on nix  :+1:

PR #1722, opened 2017-12-10
Fix for builds with system libcurl < 7.30

bhipple:
CentOS 7.4 and RHEL 7.4 ship with `libcurl-devel-7.29.0-42.el7.x86_64`; this flag
was added in 7.30.0
https://curl.haxx.se/libcurl/c/CURLMOPT_MAX_TOTAL_CONNECTIONS.html

Tangentially related to https://github.com/NixOS/nixpkgs/issues/32320, which tracks some efforts to get the latest Nix working on CentOS 7.4.  To fully get there we need to figure out what to do with c++14 and libsodium, but in the meantime this is a generic fix for any enterprise distribution maintaining an older libcurl for binary compatibility.

### DIFF
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -369,11 +369,13 @@ struct CurlDownloader : public Downloader
 
         curlm = curl_multi_init();
 
-        #if LIBCURL_VERSION_NUM >= 0x072b00 // correct?
+        #if LIBCURL_VERSION_NUM >= 0x072b00 // Multiplex requires >= 7.43.0
         curl_multi_setopt(curlm, CURLMOPT_PIPELINING, CURLPIPE_MULTIPLEX);
         #endif
+        #if LIBCURL_VERSION_NUM >= 0x071e00 // Max connections requires >= 7.30.0
         curl_multi_setopt(curlm, CURLMOPT_MAX_TOTAL_CONNECTIONS,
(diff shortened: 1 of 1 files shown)


src/libstore/download.cc
```diff
@@ -369,11 +369,13 @@ struct CurlDownloader : public Downloader
 
         curlm = curl_multi_init();
 
-        #if LIBCURL_VERSION_NUM >= 0x072b00 // correct?
```
bhipple: The answer to your question is that it was indeed correct, @edolstra :)
https://curl.haxx.se/libcurl/c/CURLMOPT_PIPELINING.html

I can remove/reformat/change the comments if you prefer.

edolstra:
Disabling multiplexing will kill HTTP/2 performance. But then, maybe those curl versions don't support HTTP/2 anyway?

bhipple:
Indeed, if you have an older libcurl the choices are either fail to compile or build without HTTP/2; the libcurl used on CentOS/RHEL [came out in early 2013](https://curl.haxx.se/mail/lib-2013-02/0086.html).

Probably the best thing to do is use an existing nix installation to bootstrap a statically linked nix for deployment.  I believe @abbradar mentioned doing something to this effect for CentOS 7.4, but that said, it doesn't hurt to have the option of building Nix by itself on these platforms.

PR #1723, opened 2017-12-11
Fix forward-decl of Bindings as "class", match definition.

dtzWill:
(appease clang -Wmismatched-tags warning)

NFCI.


dtzWill:
(bundling this change into #1724)

PR #1729, opened 2017-12-12
tests.setuid: only on i686 and x86_64 linuxs

grahamc:
Please backport to 1.11 as well.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -225,10 +225,12 @@ let
       nix = build.x86_64-linux; system = "x86_64-linux";
     });
 
-    tests.setuid = pkgs.lib.genAttrs (pkgs.lib.filter (pkgs.lib.hasSuffix "-linux") systems) (system:
-      import ./tests/setuid.nix rec {
-        nix = build.${system}; inherit system;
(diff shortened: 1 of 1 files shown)


grahamc:
Note: We'll want to fix the test itself and reenable on aarch64 once Nixpkgs' aarch64 VM test support is improved.

release.nix
```diff
@@ -225,10 +225,12 @@ let
       nix = build.x86_64-linux; system = "x86_64-linux";
     });
 
-    tests.setuid = pkgs.lib.genAttrs (pkgs.lib.filter (pkgs.lib.hasSuffix "-linux") systems) (system:
-      import ./tests/setuid.nix rec {
-        nix = build.${system}; inherit system;
-      });
+    tests.setuid = pkgs.lib.genAttrs
+      (pkgs.lib.filter (system: system == "x86_64-linux" || system == "i686-linux") systems)
```
edolstra: This can be written more concisely as `genAttrs ["i686-linux" "x86_64-linux"]`. :-)

PR #1736, opened 2017-12-15
Gracefully handle exceptions from stoi

bgamari:
I encountered an exception from `stoi` which I believe was from the JSON parser.


shlevy:
@edolstra barring objections I'll merge early next week.

PR #1745, opened 2017-12-17
Add an issue template to cut down on the confusion

### LINKED ISSUE #1008: Add contributing.md or issue template to make it clear that this is the Nix repository
We regularly get tickets in here that probably belong in the nixpkgs repo. GitHub gives us issue templates and contributing.md to clarify exactly what the issue is about, and it seems like we could save some wasted time by just pointing people to the right issue tracker in the first place.

copumpkin:
This should probably be a CONTRIBUTING.md or a SUPPORT.md but those are hidden behind a link, whereas doing it this way forces people's eyeballs onto it.

Fixes #1008

cc @edolstra @shlevy @domenkozar @grahamc

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,27 @@
+<!--
+
+# Filing a Nix issue
+
+*WAIT* Are you sure you're filing your issue in the right repository?
+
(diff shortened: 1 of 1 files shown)


copumpkin:
Now that I've tested it, perhaps this blurb is too long? Maybe a quick note about it in the issue body telling people to read the contributing link at the top level?

PR #1746, opened 2017-12-19
linenoise.cpp: allow completions from empty input

### LINKED ISSUE #1742: nixUnstable doesn't tab-complete unless a character has been entered
One of the nice uses of tab-completion in `nix-repl` is to get a sense for which bindings are currently in scope. Unfortunately, it seems that `nix repl` preclude this use since it only tab-completes if at least one character has been entered at the prompt.

dtzWill:
Fixes #1742.


dtzWill:
Was using `expect` to test this, here's the script I put together in case it's useful:

https://gist.github.com/dtzWill/23987bf3b4160dc41711670a33adc603

bgamari:
Thanks @dtzWill!

PR #1748, opened 2017-12-19
release.nix: use 'nixpkgs' argument, even if nixpkgs in NIX_PATH

dtzWill:
When trying to use custom Nix (upstream with my outstanding PR's optimistically merged),
I noticed that overriding the `nixpkgs` argument to `release.nix` doesn't seem to have any impact.
At least not if `nixpkgs` is on `NIX_PATH` (which it usually isn't on Hydra).

Concretely this caused Nix to fail to build since my system's `nixpkgs` doesn't have the new aws bits (I'm on 17.09).

The default value for the argument reads from NIX_PATH and seems to work (happily no recursive definition problems?),
so thought I'd submit it to y'all in case it seems like a good change.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -6,7 +6,7 @@
 
 let
 
-  pkgs = import <nixpkgs> {};
+  pkgs = import nixpkgs {};
 
   jobs = rec {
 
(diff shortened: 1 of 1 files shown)


domenkozar:
+1 cc @edolstra

edolstra:
I had a more complete version of this in my git stash that I've pushed instead (6d8087083278b078d0db6238fb16929163388acd).

PR #619, opened 2015-08-26
Fix adding large paths to the store

Mathnerd314:
This requires a patched version of the latest boost (sed 's|explicit basic_format|basic_format|' boost/format_class.hpp) compiled with -std=c++14.

PR #1754, opened 2017-12-22
Fix adding paths to the store (try 3)

bgamari:
This is a rebased version of the patch originally authored by @Mathnerd314 and proposed for merge in #619. It refactors the `addToStore` logic to stream data from the file system, avoiding the need to read potentially large files into memory. Note that this patch, unlike the version initially put forth in #619, incurs no dependency on Boost.

So far this is only build tested. However, I would like review on the overall idea.

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1098,13 +1098,12 @@ Path LocalStore::addToStore(const string & name, const Path & _srcPath,
             copyPath(srcPath, dstPath, filter, recursive);
 
             canonicalisePathMetaData(dstPath, -1);
-            HashResult hash = hashPath(htSHA256, dstPath);
             optimisePath(dstPath); // FIXME: combine with hashPath()
 
             ValidPathInfo info;
             info.path = dstPath;
-            info.narHash = hash.first;
-            info.narSize = hash.second;
+            info.narHash = h.first;
+            info.narSize = h.second;
             registerValidPath(info);
         }
 
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1017,13 +1017,31 @@ void LocalStore::addToStore(const ValidPathInfo & info, const ref<std::string> &
 }
 
 
-Path LocalStore::addToStoreFromDump(const string & dump, const string & name,
-    bool recursive, HashType hashAlgo, RepairFlag repair)
+struct HashAndReadSource : Source
 {
-    Hash h = hashString(hashAlgo, dump);
+    Source & readSource;
+    HashSink hashSink;
+    bool hashing;
+    HashAndReadSource(Source & readSource) : readSource(readSource), hashSink(htSHA256)
+    {
+        hashing = true;
+    }
+    size_t read(unsigned char * data, size_t len)
+    {
+        size_t n = readSource.read(data, len);
+        if (hashing) hashSink(data, n);
+        return n;
+    }
+    HashResult finish()
+    {
+        hashing = false;
+        return hashSink.finish();
+    }
+};
 
-    Path dstPath = makeFixedOutputPath(recursive, h, name);
-
+Path LocalStore::addToStoreFromDump(Source & source, const Path & dstPath,
+    bool recursive, bool repair)
+{
     addTempRoot(dstPath);
 
     if (repair || !isValidPath(dstPath)) {
@@ -1041,32 +1059,17 @@ Path LocalStore::addToStoreFromDump(const string & dump, const string & name,
 
             autoGC();
 
-            if (recursive) {
-                StringSource source(dump);
-                restorePath(realPath, source);
-            } else
-                writeFile(realPath, dump);
-
+            HashAndReadSource hashSource(source);
+            restorePath(realPath, hashSource, recursive);
+            HashResult hash = hashSource.finish();
             canonicalisePathMetaData(realPath, -1);
-
-            /* Register the SHA-256 hash of the NAR serialisation of
-               the path in the database.  We may just have computed it
-               above (if called with recursive == true and hashAlgo ==
-               sha256); otherwise, compute it here. */
-            HashResult hash;
-            if (recursive) {
-                hash.first = hashAlgo == htSHA256 ? h : hashString(htSHA256, dump);
-                hash.second = dump.size();
-            } else
-                hash = hashPath(htSHA256, realPath);
-
             optimisePath(realPath); // FIXME: combine with hashPath()
 
             ValidPathInfo info;
             info.path = dstPath;
             info.narHash = hash.first;
             info.narSize = hash.second;
-            info.ca = makeFixedOutputCA(recursive, h);
+            info.ca = makeFixedOutputCA(recursive, hash.first);
             registerValidPath(info);
         }
 
@@ -1081,17 +1084,34 @@ Path LocalStore::addToStore(const string & name, const Path & _srcPath,
     bool recursive, HashType hashAlgo, PathFilter & filter, RepairFlag repair)
 {
     Path srcPath(absPath(_srcPath));
+    Hash h = recursive ? hashPath(hashAlgo, srcPath, filter).first : hashFile(hashAlgo, srcPath);
+    Path dstPath = makeFixedOutputPath(recursive, h, name);
+    addTempRoot(dstPath);
+    if (repair || !isValidPath(dstPath)) {
 
-    /* Read the whole path into memory. This is not a very scalable
-       method for very large paths, but `copyPath' is mainly used for
-       small files. */
-    StringSink sink;
-    if (recursive)
-        dumpPath(srcPath, sink, filter);
-    else
-        sink.s = make_ref<std::string>(readFile(srcPath));
+        PathLocks outputLock({dstPath});
 
-    return addToStoreFromDump(*sink.s, name, recursive, hashAlgo, repair);
+        if (repair || !isValidPath(dstPath)) {
+
+            if (pathExists(dstPath)) deletePath(dstPath);
+
+            copyPath(srcPath, dstPath, filter, recursive);
+
+            canonicalisePathMetaData(dstPath, -1);
+            HashResult hash = hashPath(htSHA256, dstPath);
+            optimisePath(dstPath); // FIXME: combine with hashPath()
+
+            ValidPathInfo info;
+            info.path = dstPath;
+            info.narHash = hash.first;
+            info.narSize = hash.second;
+            registerValidPath(info);
+        }
+
+        outputLock.setDeletion(true);
+    }
+
+    return dstPath;
 }
 
 
@@ -1139,7 +1159,6 @@ Path LocalStore::addTextToStore(const string & name, const string & s,
 
     return dstPath;
 }
-
 
 /* Create a temporary directory in the store that won't be
    garbage-collected. */
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -412,9 +412,16 @@ Path RemoteStore::addToStore(const string & name, const Path & _srcPath,
 
     auto conn(connections->get());
 
+    if (GET_PROTOCOL_MINOR(conn->daemonVersion) < 0x15) {
+        throw Error("adding to the store is not supported with an old Nix daemon");
+    }
+
     Path srcPath(absPath(_srcPath));
 
-    conn->to << wopAddToStore << name
+    Hash h = recursive ? hashPath(hashAlgo, srcPath, filter).first : hashFile(hashAlgo, srcPath);
+    Path dstPath = makeFixedOutputPath(recursive, h, name);
+
+    conn->to << wopAddToStore << dstPath
        << ((hashAlgo == htSHA256 && recursive) ? 0 : 1) /* backwards compatibility hack */
        << (recursive ? 1 : 0)
        << printHashType(hashAlgo);
--- a/src/libstore/local-store.hh
+++ b/src/libstore/local-store.hh
@@ -152,8 +152,8 @@ public:
        in `dump', which is either a NAR serialisation (if recursive ==
        true) or simply the contents of a regular file (if recursive ==
        false). */
-    Path addToStoreFromDump(const string & dump, const string & name,
-        bool recursive = true, HashType hashAlgo = htSHA256, RepairFlag repair = NoRepair);
+    Path addToStoreFromDump(Source& dump, const Path& dstPath,
+        bool recursive = true, bool repair = false);
 
     Path addTextToStore(const string & name, const string & s,
         const PathSet & references, RepairFlag repair) override;
--- a/src/libstore/worker-protocol.hh
+++ b/src/libstore/worker-protocol.hh
@@ -6,7 +6,7 @@ namespace nix {
 #define WORKER_MAGIC_1 0x6e697863
 #define WORKER_MAGIC_2 0x6478696f
 
-#define PROTOCOL_VERSION 0x114
+#define PROTOCOL_VERSION 0x115
 #define GET_PROTOCOL_MAJOR(x) ((x) & 0xff00)
 #define GET_PROTOCOL_MINOR(x) ((x) & 0x00ff)
 
--- a/src/libutil/archive.cc
+++ b/src/libutil/archive.cc
@@ -253,7 +253,7 @@ static void parse(ParseSink & sink, Source & source, const Path & path)
                             names[name] = 0;
                     }
                 } else if (s == "node") {
-                    if (s.empty()) throw badArchive("entry name missing");
+                    if (name.empty()) throw badArchive("entry name missing");
                     parse(sink, source, path + "/" + name);
                 } else
                     throw badArchive("unknown field " + s);
@@ -289,10 +289,17 @@ void parseDump(ParseSink & sink, Source & source)
 struct RestoreSink : ParseSink
 {
     Path dstPath;
+    bool recursive;
     AutoCloseFD fd;
+
+    RestoreSink(const Path& dstPath, bool recursive)
+     : dstPath(dstPath), recursive(recursive)
+    {
+    }
 
     void createDirectory(const Path & path)
     {
+        if(!recursive) throw Error("regular file expected");
         Path p = dstPath + path;
         if (mkdir(p.c_str(), 0777) == -1)
             throw SysError(format("creating directory '%1%'") % p);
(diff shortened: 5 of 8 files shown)


src/libstore/local-store.cc
```diff
@@ -1081,17 +1084,34 @@ Path LocalStore::addToStore(const string & name, const Path & _srcPath,
...
-    else
-        sink.s = make_ref<std::string>(readFile(srcPath));
+        if (repair || !isValidPath(dstPath)) {
+
+            if (pathExists(dstPath)) deletePath(dstPath);
+
+            copyPath(srcPath, dstPath, filter, recursive);
+
+            canonicalisePathMetaData(dstPath, -1);
+            HashResult hash = hashPath(htSHA256, dstPath);
```
shlevy: So we copy then hash? How does this differ from the `hashPath` call above?

src/libstore/remote-store.cc
```diff
@@ -412,9 +412,16 @@ Path RemoteStore::addToStore(const string & name, const Path & _srcPath,
 
     auto conn(connections->get());
 
+    if (GET_PROTOCOL_MINOR(conn->daemonVersion) < 0x15) {
```
shlevy: That's a pretty tight compatibility requirement :(

bgamari: Right, I honestly don't know whether we could improve the compatibility story here. Ideas are certainly welcome.

src/libstore/local-store.cc
bgamari: I suspect this `hashPath` is redundant. Assuming that `canonicalisePathMetaData` doesn't affect the hash, it seems like we should be able to use the hash `h` computed above.

bgamari: @Mathnerd314, do you recall why this second hash was needed?

Mathnerd314: See the comment [over here](https://github.com/NixOS/nix/pull/1754/files#diff-dcbcac55a634031f9cc73707da6e4b18L1052). The first hash computes the output path, the second is the SHA256 of the NAR serialization which is used for validity checking. I guess the code could re-use the hash in the case of recursive SHA256, but hashing isn't that expensive.

src/libstore/local-store.cc
```diff
@@ -1081,17 +1084,33 @@ Path LocalStore::addToStore(const string & name, const Path & _srcPath,
...
+
+            copyPath(srcPath, dstPath, filter, recursive);
+
+            canonicalisePathMetaData(dstPath, -1);
+            optimisePath(dstPath); // FIXME: combine with hashPath()
+
+            ValidPathInfo info;
+            info.path = dstPath;
+            info.narHash = h.first;
+            info.narSize = h.second;
```
volth: This fails:

volth:
Also two tests fail (I used only [the first commit](https://github.com/NixOS/nix/pull/1754/commits/1eb1b2b0014f04cbacdb910a7a1784b9f49b31ad), because [the second](https://github.com/NixOS/nix/pull/1754/commits/d7023eddb821da376e3fa40cc228a7796bcde5bf) [breaks compilability](https://github.com/NixOS/nix/pull/1754#discussion_r170350734)):

```
running test tests/linux-sandbox.sh... [FAIL]
    + clearStore
    + echo 'clearing store...'
    clearing store...
    + chmod -R +w /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store
    + mkdir /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix
    + mkdir /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix
    + nix-store --init
    + clearProfiles
    + profiles=/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix/profiles
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix/profiles
    + canUseSandbox
    ++ uname
    + [[ Linux != Linux ]]
    + '[' '!' -L /proc/self/ns/user ']'
    + '[' -e /proc/sys/kernel/unprivileged_userns_clone ']'
    + return 0
    + [[ ! /nix/store/zqh3l3lyw32q1ayb15bnvg9f24j5v2p0-bash-4.4-p12/bin/bash =~ /nix/store ]]
    + chmod -R u+w /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store0
    chmod: cannot access '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store0': No such file or directory
    + true
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store0
    + export NIX_STORE_DIR=/my/store
    + NIX_STORE_DIR=/my/store
    + export NIX_REMOTE=/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store0
    + NIX_REMOTE=/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store0
    ++ nix-build dependencies.nix --no-out-link --sandbox-paths /nix/store
    error: opening lock file '/my/store/nk93jzqyamdzapyi4hl4yf8sfyz1r4ij-dependencies.builder0.sh.lock': No such file or directory
    + outPath=

```

----

```
running test tests/signing.sh... [FAIL]
    + clearStore
    + echo 'clearing store...'
    clearing store...
    + chmod -R +w /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store
    + mkdir /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix
    + mkdir /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix
    + nix-store --init
    + clearProfiles
    + profiles=/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix/profiles
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/var/nix/profiles
    + clearCache
    + rm -rf /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/binary-cache
    + nix-store --generate-binary-cache-key cache1.example.org /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/sk1 /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/pk1
    ++ cat /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/pk1
    + pk1=cache1.example.org:kKxzV+LmViZbeyFnMDH+9BQgTypP8dMoirrUSSrF6uQ=
    + nix-store --generate-binary-cache-key cache2.example.org /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/sk2 /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/pk2
    ++ cat /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/pk2
    + pk2=cache2.example.org:m9ogRTUnTuAMLFXcgLLNaOMcbaJkpnATz2j7biD28fc=
    ++ nix-build dependencies.nix --no-out-link --secret-key-files '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/sk1 /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/sk2'
    error (ignored): closing file descriptor 9: Bad file descriptor
    error (ignored): closing file descriptor 9: Bad file descriptor
    error (ignored): closing file descriptor 9: Bad file descriptor
      /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/8328d8bgfrfhnj6hzr7xh2ba11vkqs27-dependencies-input-1.drv
      /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/gb079vg7gbjcmbxgwp1fvql2a44qmg8j-dependencies-input-2.drv
      /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/v6wfk2iglacnnl63k4pl62xwvhacqkdf-dependencies.drv
    building '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/8328d8bgfrfhnj6hzr7xh2ba11vkqs27-dependencies-input-1.drv'...
    building '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/gb079vg7gbjcmbxgwp1fvql2a44qmg8j-dependencies-input-2.drv'...
    building '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/v6wfk2iglacnnl63k4pl62xwvhacqkdf-dependencies.drv'...
    FOO
    + outPath=/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies
    ++ nix path-info --json /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies
    + info='[{"path":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies","narHash":"sha256:0v3z2rzfwbl7v76gnhvacdv1shpwg3lj72xr2hzw4ljv1xzj
lvhc","narSize":1112,"references":["/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/1px6lmbwalajvh5n2ddj52i0qp5fv5p5-dependencies-input-2","/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-
1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies"],"deriver":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/v6wfk2iglacnnl63k4pl62xwvhacqkdf-dependencies.drv","registr
ationTime":1519415826,"ultimate":true,"signatures":["cache1.example.org:1476lqWQDF41Yu61XuTjotr531kjTVma6Q1yZMZ7dK/DAk0FoLPhGD7HEpodnVLuirf32Fgioq0+OVvfrMntCA==","cache2.example.org:oCMW7HN5CNqCh
Q19ZqllAOB5R7C5plE7Cul/wfm1e2FADluCLSei0ax/AGTzpSVzUzdlTVZhE+CMI/1hwwc2Aw=="]}]'
    + [[ [{"path":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies","narHash":"sha256:0v3z2rzfwbl7v76gnhvacdv1shpwg3lj72xr2hzw4ljv1xzjlvh
c","narSize":1112,"references":["/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/1px6lmbwalajvh5n2ddj52i0qp5fv5p5-dependencies-input-2","/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/n
ix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies"],"deriver":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/v6wfk2iglacnnl63k4pl62xwvhacqkdf-dependencies.drv","registrati
onTime":1519415826,"ultimate":true,"signatures":["cache1.example.org:1476lqWQDF41Yu61XuTjotr531kjTVma6Q1yZMZ7dK/DAk0FoLPhGD7HEpodnVLuirf32Fgioq0+OVvfrMntCA==","cache2.example.org:oCMW7HN5CNqChQ19
ZqllAOB5R7C5plE7Cul/wfm1e2FADluCLSei0ax/AGTzpSVzUzdlTVZhE+CMI/1hwwc2Aw=="]}] =~ "ultimate":true ]]
    + [[ [{"path":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies","narHash":"sha256:0v3z2rzfwbl7v76gnhvacdv1shpwg3lj72xr2hzw4ljv1xzjlvh
c","narSize":1112,"references":["/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/1px6lmbwalajvh5n2ddj52i0qp5fv5p5-dependencies-input-2","/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/n
ix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies"],"deriver":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/v6wfk2iglacnnl63k4pl62xwvhacqkdf-dependencies.drv","registrati
onTime":1519415826,"ultimate":true,"signatures":["cache1.example.org:1476lqWQDF41Yu61XuTjotr531kjTVma6Q1yZMZ7dK/DAk0FoLPhGD7HEpodnVLuirf32Fgioq0+OVvfrMntCA==","cache2.example.org:oCMW7HN5CNqChQ19
ZqllAOB5R7C5plE7Cul/wfm1e2FADluCLSei0ax/AGTzpSVzUzdlTVZhE+CMI/1hwwc2Aw=="]}] =~ cache1\.example\.org ]]
    + [[ [{"path":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies","narHash":"sha256:0v3z2rzfwbl7v76gnhvacdv1shpwg3lj72xr2hzw4ljv1xzjlvh
c","narSize":1112,"references":["/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/1px6lmbwalajvh5n2ddj52i0qp5fv5p5-dependencies-input-2","/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/n
ix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies"],"deriver":"/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/v6wfk2iglacnnl63k4pl62xwvhacqkdf-dependencies.drv","registrati
onTime":1519415826,"ultimate":true,"signatures":["cache1.example.org:1476lqWQDF41Yu61XuTjotr531kjTVma6Q1yZMZ7dK/DAk0FoLPhGD7HEpodnVLuirf32Fgioq0+OVvfrMntCA==","cache2.example.org:oCMW7HN5CNqChQ19
ZqllAOB5R7C5plE7Cul/wfm1e2FADluCLSei0ax/AGTzpSVzUzdlTVZhE+CMI/1hwwc2Aw=="]}] =~ cache2\.example\.org ]]
    + nix verify -r /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies
    + expect 2 nix verify -r /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies --sigs-needed 1
    + local expected res
    + expected=2
    + shift
    + set +e
    + nix verify -r /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies --sigs-needed 1
    path '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/1px6lmbwalajvh5n2ddj52i0qp5fv5p5-dependencies-input-2' is untrusted
    path '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies' is untrusted
    + res=2
    + set -e
    + [[ 2 -eq 2 ]]
    + nix verify -r /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies --sigs-needed 1 --trusted-public-keys cache1.example.org:kKxzV+LmViZb
eyFnMDH+9BQgTypP8dMoirrUSSrF6uQ=
    + expect 2 nix verify -r /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies --sigs-needed 2 --trusted-public-keys cache1.example.org:kKx
zV+LmViZbeyFnMDH+9BQgTypP8dMoirrUSSrF6uQ=
    + local expected res
    + expected=2
    + shift
    + set +e
    + nix verify -r /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies --sigs-needed 2 --trusted-public-keys cache1.example.org:kKxzV+LmViZb
eyFnMDH+9BQgTypP8dMoirrUSSrF6uQ=
    path '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/1px6lmbwalajvh5n2ddj52i0qp5fv5p5-dependencies-input-2' is untrusted
    path '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies' is untrusted
    + res=2
    + set -e
    + [[ 2 -eq 2 ]]
    + nix verify -r /tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/xfvbz5fj2zqzgr0rjcaw70x0sj86r6r9-dependencies --sigs-needed 2 --trusted-public-keys 'cache1.example.org:kKxzV+LmViZ
beyFnMDH+9BQgTypP8dMoirrUSSrF6uQ= cache2.example.org:m9ogRTUnTuAMLFXcgLLNaOMcbaJkpnATz2j7biD28fc='
    + nix verify --all --sigs-needed 2 --trusted-public-keys 'cache1.example.org:kKxzV+LmViZbeyFnMDH+9BQgTypP8dMoirrUSSrF6uQ= cache2.example.org:m9ogRTUnTuAMLFXcgLLNaOMcbaJkpnATz2j7biD28fc='
    path '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/i2kwqvs2gvvcp84swmcmpcf0xfka606q-dependencies.builder0.sh' is untrusted
    path '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/mggcgq4bwcvcjrhhpzhc01zdw98dw5xq-dependencies.builder1.sh' is untrusted
    path '/tmp/nix-build-nix-2.0pre5968_a6c0b773.drv-1/nix-test/store/viv58n9vsdk25zim6qclg0qbvmgh2v2f-dependencies.builder2.sh' is untrusted
```

volth:
I plan to use this patch or any patch which would solve the high memory consumption of ```nix-store```.
200Mb of ```binutils``` and 400Mb of ```oraclejdk8``` makes NixOS not upgradable on small VPSes.

Well, I can prepare a disk image on a powerful server, but cannot upload a new closure to an existing instance because it likely contains a big package which causes "out of memory" error

dtzWill:
Recent commits (based on earlier version of this I think?) should alleviate much of this, can this be closed or is it still relevant?

volth:
I'd close, it is broken anyway

bgamari:
I will happily find time to look into the failures; however, I would first like some feedback on the acceptability of the general approach.

copumpkin:
In general it looks fine to me, but I think only @shlevy and @edolstra are really qualified to evaluate deeply.

shlevy:
@bgamari In a general sense I'm very in favor of this, but can you provide a high level view of what this is doing on top of the recent streaming improvements already on master?

edolstra:
@shlevy Those improve copying paths between stores (including substitution), they don't improve `addToStore` from a (possibly filtered) path.

If I read this PR correctly, it reads the path twice: first to compute the hash / store path, second to copy it to the destination store path. This has a race: if the contents of the source path changes in between, we end up with a corrupt store path. (Note that this is a security bug.)

In general, we have a tension here between performance and memory usage. There are several solutions:

* The current implementation: read once into memory, hash, write to the store if the destination store path is not already valid. O(n) memory.

* Hash first, then copy if the destination store path is not already valid (this PR). This requires reading and hashing twice which is obviously slower.

* Copy to a temporary location in the store, then rename() if the destination store path is not already valid. This is single-pass but does unnecessary write I/O in the common case where the path is already valid.

* Hybrid: do 1) if the path is small, otherwise do 2) or 3).

However, maybe this should be a "will not fix" since Nix is not really intended for use cases where you have enormous path dependencies, because it will always be slow. (Even in the best case, it needs to read/hash the entire path on every evaluation.) If you have a giant dependency, you're better off using `fetchurl` or something similar.

shlevy:
Hmm, IMO a hybrid approach is a good idea, especially since we already have the "large path" warning so we already have a sense of "small". The problem with fetchurl or similar is the need to know the hash in advance and it doesn't have the nice relative path resolution, without that I think that'd be a good solution here.

edolstra:
Well, you *have* to specify the hash to prevent having to read the file on every evaluation. Though I suppose `builtins.fetchurl` could have a store path cache similar to `builtins.fetchGit` (keyed on e.g. inode + mtime + file size).

shlevy:
Right, I'm saying there's a reasonable use case space for taking the hit of a long read on every eval to avoid manually specifying the hash.

Was actually going to separately suggest an mtime-based cache as well, so definitely let's do that :grinning:

edolstra:
Well, we don't have to support every reasonable use case if the cost in terms of complexity is too high. The hybrid approach is pretty complex, especially if you take RemoteStore into account.

volth:
Doesn't it cover also the case when huge files added to the store on
the destination side of "nix-copy-closure"?
Or that is already fixed with the recent changes?

shlevy:
@edolstra I guess the problem with the fetchurl approach is filters etc. Maybe what we want is a new option on `builtins.path`, so you have to opt-in to the stream-to-temp-file approach, and we can add caching to that or not.

I do agree we don't need to support all use cases of course, but I do think this is an (unfortunately) common one in using Nix for development. These repos grow huge and it's not always trivial to filter them (and sometimes the size is in actual source code...).

edolstra:
@volth Yeah, copy closure is largely fixed.

Ericson2314:
> If you have a giant dependency, you're better off using `fetchurl` or something similar.

But how does one *get* the hash in the first place to write the `fetch*`? Maybe the new Nix 2.0 handling of fixed-output hash mismatch failures helps?

> Copy to a temporary location in the store, then rename...

Renaming is really cheap on any half-decent file-system. Let's just do that. In fact, don't we already do that approach to support that new behavior of avoiding the second fetch on hash mismatch which I mentioned above? If not, FWIW the intentional store also has to do the exact same thing, so I'd hope we'd end up writing this anyways. 

-----

Taking a step back, it's *really* important that we fix this. People new to Nix regularly hit this as they incrementally nixify some preexisting build process, as that usually means larger files than usual. When I evangelize Nix to someone, I need to offer up tons of disclaimers for things like this so that when they inevitably hit this, Nix and I don't instantly loose all credibility. A lot of these issues like this one wouldn't require adding too much complexity to fix, but fixing them would greatly improve Nix's image, and the experience for Nix users.

edolstra:
> Renaming is really cheap on any half-decent file-system.

The problem is not renaming, it's writing it to the store in the first place. This causes a huge amount of unnecessary I/O in the common case where the target path already exists. (Also you can run out of disk space.)

edolstra:
Okay, I took a stab at implementing hybrid option 3: https://github.com/edolstra/nix/commit/c94b4fc7ee0c7b322a5f3c7ee784063b47a11d98

Ericson2314:
@edolstra 

> This causes a huge amount of unnecessary I/O in the common case where the target path already exists. (Also you can run out of disk space.)

First of all, very glad you are attempting a solution here! If we are concerned memory usage and disk usage, and secondarily complexity, @kmicklas give me the idea for this.

1. `hash . toNar` everything in place
2. If we don't have path, "cp -r" everything to temp location
3. `hash . toNar` temp location
4. Move to proper location (optional warning if hash changed, store path is not corrupt either way so doesn't matter too much)

Compared to the existing implementation, this *always* uses <= disk space and way smaller small O(1) memory, at the cost of 2x read I/O in the less common case where path is indeed new. For the small file common case that should be a negligible cost. 

Compared to your https://github.com/edolstra/nix/commit/c94b4fc7ee0c7b322a5f3c7ee784063b47a11d98, besides the 2x reads, it avoids the complexity / discontinuity of an arbitrary cutoff, and avoids a `fromNar . toNar` round trip, giving the underlying filesystem a chance to optimize the `cp -r`.

What do you think?

shlevy:
@Ericson2314 I like that approach, though ultimately I think the tradeoff should be decided with profiling. In any case, for small files we needn't take the 2x I/O hit, as we can just continue reading it all into memory.

shlevy:
And we should possibly consider a cache (based on inode+size+mtime or whatever), and make it easy for a hash to be provided up-front to avoid this concern altogether.

PR #1765, opened 2017-12-27
match and split discard string content

volth:
I am not sure if there is a real issue, I have not hit one, just noticed during code review
(although in ```nixpkgs``` these functions are really called with strings with context)


shlevy:
I think the problem here is that it's not clear whether the matched string/which element of the split actually contains the reference. I'd be hesitant to include it without clear examples of where this semantics is desired.

volth:
At the same time, ```substring``` always copies the context.
Even if the result is empty string.

PR #1766, opened 2017-12-29
Do not export ASPELL_CONF

FRidh:
This does not belong in Nix. Setting this env var is already done by the aspell derivation found in Nixpkgs.


FRidh:
In my opinion this should be backported as well; it does not belong in Nix, and I highly doubt anyone actually depends on it.

PR #1768, opened 2017-12-29
use libbrotli directly when available

dtzWill:
Tried to match coding style / design decisions done by other compression code.

Basic there-and-back-again test included, seems to properly fail if I inject extra bytes or corrupt a file.

Speed and size settings are the same as those in the tool (defaults);
tried copying `okular` from my system to brotli-compressed store before and after these changes and the resulting stores are close in size (183MB vs 184MB) and can be read using both methods.

--------

* Look for both 'brotli' and 'bro' as external command,
  since upstream has renamed it in newer versions.
  If neither are found, current runtime behavior
  is preserved: try to find 'bro' on PATH.
* Limit amount handed to BrotliEncoderCompressStream
  to ensure interrupts are processed in a timely manner.
  Testing shows negligible performance impact.
  (Other compression sinks don't seem to require this)


dtzWill:
Updated:
* Better interrupt behavior when compressing large strings
* brotli not hard dependency, due to ubuntu/fedora missing packages for it in the releases used
* General formatting and other cleanups

edolstra:
Thanks @dtzWill!

PR #1771, opened 2017-12-31
nix-env.cc: make `--query` evaluate `.outPath`s

oxij:
NixOS/nixpkgs#33057 delays `stdenv.mkDerivation` derivation checks (e.g.
meta, brokennes, etc) to `.drvPath` and `.outPath` evaluation leaving
other fields accessible. That has a side-effect of making `nix-env -qa`
list broken packages.

This patch forces those checks.

### DIFF
--- a/doc/manual/command-ref/nix-env.xml
+++ b/doc/manual/command-ref/nix-env.xml
@@ -877,11 +877,6 @@ $ nix-env --set-flag priority 10 gcc</screen>
     <arg choice='plain'><option>--available</option></arg>
     <arg choice='plain'><option>-a</option></arg>
   </group>
-  <group choice='opt'>
-    <arg choice='plain'><option>--all</option></arg>
-    <arg choice='plain'><option>--approximate</option></arg>
-    <arg choice='plain'><option>--strict</option></arg>
-  </group>
 
   <sbr />
 
--- a/doc/manual/command-ref/nix-env.xml
+++ b/doc/manual/command-ref/nix-env.xml
@@ -877,6 +877,11 @@ $ nix-env --set-flag priority 10 gcc</screen>
     <arg choice='plain'><option>--available</option></arg>
     <arg choice='plain'><option>-a</option></arg>
   </group>
+  <group choice='opt'>
+    <arg choice='plain'><option>--all</option></arg>
+    <arg choice='plain'><option>--approximate</option></arg>
+    <arg choice='plain'><option>--strict</option></arg>
+  </group>
 
   <sbr />
 
--- a/src/nix-env/nix-env.cc
+++ b/src/nix-env/nix-env.cc
@@ -897,6 +897,7 @@ static void opQuery(Globals & globals, Strings opFlags, Strings opArgs)
     bool jsonOutput = false;
 
     enum { sInstalled, sAvailable } source = sInstalled;
+    enum { eAll, eApproximate, eStrict } evaluate = eApproximate;
 
     settings.readOnlyMode = true; /* makes evaluation a bit faster */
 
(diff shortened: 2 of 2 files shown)


vcunat:
Well, that's possible, but due to the recursive nature, my previous tests suggested that forcing `.out` leads to a significant slowdown of `nix-env -qa` (roughly five-times in my year-old test).  On the other hand, it improves semantics &ndash; even closure would get checked during `-qa`.

vcunat:
Except perhaps that `queryOutPath` seems overkill, as it queries the binary cache if I understand it correctly.

oxij:
Okay, so I hacked it a little bit more and simplified the first patch considerably.

> Well, that's possible, but due to the recursive nature, my previous tests suggested that forcing `.out` leads to a significant slowdown of `nix-env -qa` (roughly five-times in my year-old test).  On the other hand, it improves semantics &ndash; even closure would get checked during `-qa`.

On my tests the first patch gives 7x slowdown (unless `--fast` is given).

However, now this patchset has the second patch, with it and the new patch in `nixpkgs` PR the slowdown is less than 2x.

I have to note, that 7x slowdown produced a nice effect: I found a couple of new bugs in expressions in `nixpkgs` on `-darwin` from `-linux` (because I could evaluate their `.outPath`s even though I couldn't build them) so I'm now seriously considering adding an option to skip setting `meta.skipme` in the other PR to get that 7x slowdown back.

> Except perhaps that `queryOutPath` seems overkill, as it queries the binary cache if I understand it correctly.

I`m looking at `src/libexpr/get-drvs.cc:78` and see nothing that seems to be related to binary caches.
It seems that `.outPath` is just simply too expensive to compute.

oxij:
After some more hacking it's now gives 4% slowdown on my machine with 99% of the end result (a small number of broken packages get listed).
Pedantic result does require 7x slowdown.

oxij:
I think this and the sibling in `nixpkgs` are perfect now, with both
patches applied its now indistinguishable by performance (with default
options) and is much better w.r.t. semantics.

Will add more documentation to the manual if there's a consensus on
merging this.

edolstra:
This behaviour (`-qa` not evaluating `outPath`) is intentional, see edd145d2fb2960fab414bfe2c4a1221b976815ea.

oxij:
> This behaviour (`-qa` not evaluating `outPath`) is intentional, see edd145d2fb2960fab414bfe2c4a1221b976815ea.

I don't understand. Is there any other motivation except speed there?

If not, then merging this and the sibling PR (NixOS/nixpkgs#33057) for `nixpkgs` makes things strictly not worse, and merging both and then setting `config.checkMetaRecursively` makes things much better (at the cost of 2x slowdown).

NixOS/nixpkgs#33057 can't be merged without this because it will make `nix-env -qa` list broken packages and you didn't want to have that in #22277.
Not merging NixOS/nixpkgs#33057 breaks NixOS/nixpkgs#32424 (see NixOS/nixpkgs#33006) and many people want to have that feature since NixOS/nixpkgs#9306.

edolstra:
Yes, the motivation is speed.

oxij:
Updated and rebased onto master. Tested to work with 2.0.4 too.

Ekleog:
@edolstra (triage) If I read correctly, currently this PR is a 4% slowdown in exchange for fixing https://github.com/NixOS/nix/issues/1963 and https://github.com/NixOS/nixpkgs/issues/37350. If so, this sounds like a net gain to me?

@oxij Can you confirm this version is the 4% slowdown you mentioned in https://github.com/NixOS/nix/pull/1771#issuecomment-354706012 and not the 7x slowdown indeed?

oxij:
The first patch here is no longer needed with `nix-env -qa` filtering semantics change introduced by https://github.com/NixOS/nixpkgs/commit/ae16dd1a15dfbe29c6a1de61c4cc23e2cd6487f0, i.e. listing broken and unfree packages by default is now an expected behavior (which is fine, IMHO). The second patch is still needed, though, and I'll make a separate PR for that.

Issue #1759, opened 2017-12-23
Nix on Mac OS 10.13.2


muji786:
Nix continues to have issues on 10.13.2 ... I am receiving the following error with nix-shell

Databrarys-Air:databrary databraryadmin$ nix-shell --version
nix-shell (Nix) 1.11.16
```
Databrarys-Air:databrary databraryadmin$ nix-shell --show-trace
building path(s) ‘/nix/store/wrl2sxy6yqgm7w7ii1515jxndncc7545-Libsystem-osx-10.11.6’
building path(s) ‘/nix/store/fy95rqnbkgs29hbgvpbab4fgcfrqm843-ICU-osx-10.10.5’
unpacking sources
unpacking source archive /nix/store/sx6iznjbjm99n8nahx4q1bx9bd5c02pd-Libsystem-1226.10.1.tar.gz
unpacking sources
unpacking source archive /nix/store/7n6jr1p3im2kbbf6zhwq1smkycyrcjaj-ICU-531.48.tar.gz
source root is Libsystem-1226.10.1
/nix/store/j1ly0zzpnzb8y8ry6k4vi4prk8vwq233-set-source-date-epoch-to-latest.sh: line 7: warning: command substitution: ignored null byte in input
setting SOURCE_DATE_EPOCH to timestamp 1449637899 of file Libsystem-1226.10.1/init.c
installing
6906 blocks
1177 blocks
93 blocks
347 blocks
182 blocks
93 blocks
24 blocks
114 blocks
603 blocks
651 blocks
352 blocks
9 blocks
42 blocks
108 blocks
33 blocks
cpio: /nix/store/wrl2sxy6yqgm7w7ii1515jxndncc7545-Libsystem-osx-10.11.6/include/./setjmp.h not created: newer or same age version exists
cpio: /nix/store/wrl2sxy6yqgm7w7ii1515jxndncc7545-Libsystem-osx-10.11.6/include/./libkern/OSCacheControl.h not created: newer or same age version exists
cpio: /nix/store/wrl2sxy6yqgm7w7ii1515jxndncc7545-Libsystem-osx-10.11.6/include/./libkern/OSAtomic.h not created: newer or same age version exists
cpio: /nix/store/wrl2sxy6yqgm7w7ii1515jxndncc7545-Libsystem-osx-10.11.6/include/./ucontext.h not created: newer or same age version exists
271 blocks
253 blocks
154 blocks
31 blocks
163 blocks
376 blocks
[...]

PR #1775, opened 2018-01-03
installer: create 'enough' build users

LnL7:
Fixes issues like https://github.com/NixOS/nix/issues/1759#issuecomment-355060143


grahamc:
How do we know 32 is `enough`? Why is that user experiencing that error / how does this fix it?

edolstra:
As an aside, in the future Nix may allocate UIDs dynamically: https://github.com/NixOS/nix/commit/ad1c827c0d7c96075e8e820fc66be5ea849497c9

copumpkin:
That'll only work in Linux though, so we'll still need something like this on Darwin

PR #1778, opened 2018-01-05
do not make Thunks for ExprLambda

volth:
It is senseless to make Thunks for ExprLambda, as a Lambda Value does nothing else than a Thunk Value: it stores pointers to Expr and Env.
Eager evaluation of ExprLambda saves some memory and CPU clocks.


volth:
oops, sorry.
it wont work with stack-allocated values.
so it cannot be backported

PR #1782, opened 2018-01-07
nix-daemon.sh: Export MANPATH

bobbypriambodo:
As Darwin installation defaults to multi-user, `nix-daemon.sh` is sourced instead of `nix.sh`. However MANPATHs are not exported, causing not-found failures when invoking `nix-env --help` (or other `nix-*` commands) and `man <cmd>` where `<cmd>` is any program installed via nix. Related issue [here](https://github.com/NixOS/nixpkgs/issues/23601#issuecomment-355757630).

Suggestions welcome as I am currently just starting out with nix :)

### DIFF
--- a/scripts/nix-profile-daemon.sh.in
+++ b/scripts/nix-profile-daemon.sh.in
@@ -52,3 +52,7 @@ fi
 export NIX_SSL_CERT_FILE="@localstatedir@/nix/profiles/default/etc/ssl/certs/ca-bundle.crt"
 export NIX_PATH="@localstatedir@/nix/profiles/per-user/root/channels"
 export PATH="$HOME/.nix-profile/bin:$HOME/.nix-profile/lib/kde4/libexec:@localstatedir@/nix/profiles/default/bin:@localstatedir@/nix/profiles/default:@localstatedir@/nix/profiles/default/lib/kde4/libexec:$PATH"
+
+if [ -n $MANPATH ]; then
+    export MANPATH="$HOME/.nix-profile/share/man:@localstatedir@/nix/var/nix/profiles/default/share/man:$MANPATH"
+fi


matthewbauer:
I like trying to get this to work out-of-the-box but this might open up a can of worms. There's lots of environment variables like this but I don't think we want to update Nix for all of them (INFOPATH, DICPATH, etc.)

I have a custom way to set this that works pretty well but involves using ```buildEnv``` and some custom Emacs stuff.

I think a more Nix-compatible approach might be possible if we made "man" look for man pages in ```$HOME/.nix-profile/share/man``` by default? (maybe also ```/nix/var/nix/profiles/default/share/man```)

bobbypriambodo:
> I think a more Nix-compatible approach might be possible if we made "man" look for man pages in $HOME/.nix-profile/share/man by default? (maybe also /nix/var/nix/profiles/default/share/man)

Do you have a suggestion on how to achieve this? I'm fine with either updating or dropping this PR if there's a better way.

bjornfor:
> I think a more Nix-compatible approach might be possible if we made "man" look for man pages in $HOME/.nix-profile/share/man by default? (maybe also /nix/var/nix/profiles/default/share/man)

I.e. iterate over $NIX_PROFILES. (Reminds me, single user Nix doesn't setup $NIX_PROFILES.)

FRidh:
[We just got rid of `ASPELL_CONF`](https://github.com/NixOS/nix/commit/ab8ba712054e170dfbef30f53b4a5a88b93aad6f). I'm of the opinion that we should provide derivations for this in Nixpkgs instead.

dtzWill:
`MANPATH` is tricky.  Appending to it only if it is already set is "safe", but many systems do not set this variable and so will still be unable to read man pages for Nix-managed software.

On Linux at least, we could just set `MANPATH="/nix/.….:$MANPATH"`, since the presence of a trailing semicolon  means that normal system paths are searched after.

Unfortunately this isn't true (apparently) on BSD or OSX, and so perhaps is not something we can use.

Two ideas:
* introduce a "NIX_MANPATH" and teach our man (man-db on Linux, not sure elsewhere) to respect it.  This way we don't break system behavior and users get expected behavior if they install man via Nix, which seems reasonable (err assuming Nix's man works as well as system "man" in most cases?)
* Try something like what is described here: https://unix.stackexchange.com/a/344606 .  Unfortunately this requires relying on `manpath`.

Finally, man by default actually searches PATH and looks for pages nearby.  I think this is fairly standard.

Unfortunately it only looks in "./man" and "../man" (and man1, etc) -- which is a bummer since we store man pages in /share/man ("../share/man" relative to PATH entries).  I had the silly idea of just adding /share/bin to PATH but apparently man only searches around if the path exists, how reasonable of it.

Creating a dummy set of directories like:

/dummy/bin
/dummy/man -> ~/.nix-profile/share/man

And then

PATH=/dummy/bin:$PATH man

Does the trick.  It's a kludge for sure, though :).

dtzWill:
Oh, I see much of this was discussed in the linked issue.  I suppose we might as well append to it if already set.

@FRidh: what did you have in mind re:derivations? A "append-to-manpath" package you install?
If so, I'd suggest installing it by default... lack of man pages (and --help giving up if they aren't present) is a critical problem especially for new users.

FRidh:
@dtzWill Basically a wrapper around `man` that sets `MANPATH` as is done here.

FRidh:
> If so, I'd suggest installing it by default... lack of man pages (and --help giving up if they aren't present) is a critical problem especially for new users.

https://github.com/NixOS/nixpkgs/issues/23601#issuecomment-357532024

bobbypriambodo:
Thank you for all the replies. From what I gathered, the preferred solution for this is a bit outside of my knowledge right now (I don't even understand much yet about what a derivation is, just trying out Nix for a short moment), and I don't think I have the time to invest on this. For now I can just work around this by manually appending the MANPATH directly to my shell config files.

@FRidh if you deem that this PR is not the way to go then feel free to close this! I really appreciate the input.

FRidh:
@bobbypriambodo I am not a maintainer of this repo, just a contributor, and what I gave is just my view on the matter ;) I think this is best discussed further in https://github.com/NixOS/nixpkgs/issues/23601.

PR #1784, opened 2018-01-08
 Add setPathName primop to copy paths to the store with a given name.

shlevy:
In some situations, relative paths in nix expressions resolve to paths
with names that are uninformative or, worse, unrepresentable in the
nix store. For example, one jenkins plugin for parallel building
starts each build in a directory named like `foo@1`, which then can't
causes nix to fail to add the path to the store when referenced with
`./.`. With setPathName, this can be fixed without modifying the
surrounding build infrastructure.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -84,7 +84,11 @@ static void printValue(std::ostream & str, std::set<const Value *> & active, con
         str << "\"";
         break;
     case tPath:
-        str << v.path; // !!! escaping?
+        if (v.path.name)
+            str << "(builtins.setPathName ";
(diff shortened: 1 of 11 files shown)


domenkozar:
Yes!! Needs a Nix tests.

shlevy:
@domenkozar done

edolstra:
I would prefer adding a function that extends `filterSource` with optional `name` and `sha256` arguments. The latter would allow local paths to be used in pure evaluation mode. E.g.
```
fetchPath {
  path = ./foo;
  name = "bla";
  sha256 = "abcdef...";
  filter = name: type: ... boolean expression ...;
}
```

PR #1786, opened 2018-01-09
release: appease checkMeta by using string for schedulingPriority

dtzWill:
See https://github.com/NixOS/nixpkgs/blob/ae9a7c4969c70950f7c92ba487069ca5091ecca2/pkgs/stdenv/generic/check-meta.nix#L165 .

Either that's wrong or this is :).

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -314,7 +314,7 @@ let
             [ "sqlite" "sqlite-devel" "bzip2-devel" "libcurl-devel" "openssl-devel" "xz-devel" "libseccomp-devel" ]
             ++ extraPackages; };
       memSize = 2048;
-      meta.schedulingPriority = 50;
+      meta.schedulingPriority = "50";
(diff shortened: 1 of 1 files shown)


grahamc:
Hmm sure seems like nixpkgs is wrong.

dtzWill:
It was probably made a string because the only instance of it in-tree is also a string: https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/release/source-tarball.nix#L126

Curiously Nix's release.nix uses this as well, so ... neither type is right?! :package: 

Just kidding, `int` clearly makes more sense here.  Time to go fix it over on nixpkgs, then! :)

edolstra:
IIRC, once upon a time, meta attributes had to be strings, but that's no longer the case. `DrvInfo::queryMetaInt()` still accepts strings for backward compatibility.

PR #1801, opened 2018-01-15
remote_store: register for NIX_REMOTE=unix://path

### LINKED ISSUE #1800: Allow overriding nixDaemonSocketFile by setting NIX_REMOTE
It would be nice to be able to set `NIX_REMOTE=daemon?path=foo` to connect to a daemon listening on an unusual socket. This would allow for flexibility in proxying remote daemons, using different Nix prefixes, other things like that.

catern:
This allows overriding the socket path so the daemon may be listening at
an arbitrary Unix domain socket location.

Fixes #1800

### DIFF
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -78,9 +78,22 @@ UDSRemoteStore::UDSRemoteStore(const Params & params)
 }
 
 
+UDSRemoteStore::UDSRemoteStore(std::string socket_path, const Params & params)
+    : Store(params)
+    , LocalFSStore(params)
+    , RemoteStore(params)
+    , path(socket_path)
+{
+}
+
+
 std::string UDSRemoteStore::getUri()
 {
-    return "daemon";
+    if (path) {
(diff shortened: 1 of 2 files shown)


src/libstore/remote-store.hh
```diff
@@ -145,6 +146,7 @@ private:
     };
 
     ref<RemoteStore::Connection> openConnection() override;
+    std::experimental::optional<std::string> path = std::experimental::nullopt;
```
edolstra: The initializer is unnecessary.

src/libstore/remote-store.cc
```diff
@@ -98,7 +111,12 @@ ref<RemoteStore::Connection> UDSRemoteStore::openConnection()
         throw SysError("cannot create Unix domain socket");
     closeOnExec(conn->fd.get());
 
-    string socketPath = settings.nixDaemonSocketFile;
+    string socketPath;
```
shlevy: Nit: `socketPath = path ? *path : settings.nixDaemonSocketFile`

PR #1813, opened 2018-01-23
Fix obscure corner case in name resolution for builtin:fetchurl in sandboxed environments

copumpkin:
Unfortunately this currently displays a few warnings like the following:

```
warning: unable to download 'http://this.pre-initializes.the.dns.resolvers.invalid': Couldn't resolve host name (6); retrying in 285 ms
warning: unable to download 'http://this.pre-initializes.the.dns.resolvers.invalid': Couldn't resolve host name (6); retrying in 668 ms
warning: unable to download 'http://this.pre-initializes.the.dns.resolvers.invalid': Couldn't resolve host name (6); retrying in 1234 ms
warning: unable to download 'http://this.pre-initializes.the.dns.resolvers.invalid': Couldn't resolve host name (6); retrying in 2106 ms
```

which I don't know how to suppress. @edolstra, any ideas?

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -6,6 +6,7 @@
 #include "archive.hh"
 #include "affinity.hh"
 #include "builtins.hh"
+#include "download.hh"
 #include "finally.hh"
 #include "compression.hh"
 #include "json.hh"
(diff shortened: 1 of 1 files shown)


copumpkin:
I suppose I could add an `bool stfu` field to `DownloadRequest` to shut up the warning, but that feels messy.

copumpkin:
Just verified that `tries = 1` silences the warnings too.

vcunat:
Right, `**.invalid` query shouldn't even make it past the first (reasonable) resolver.

PR #1815, opened 2018-01-25
Fix restricted mode when installing in non-canonical data dir

shlevy:


shlevy:
Ping

edolstra:
Why does it matter whether the install location is non-canonical? (What's the canonical location anyway?)

shlevy:
Canonical as-in `canonPath`

shlevy:
It matters because NIX_PATH lookup canonicalizes the path first, so it's disallowed in restricted mode.

LnL7:
I think there's a similar problem when `$(pwd)` is a symlink, or is this related?

shlevy:
This is related, yeah.

shlevy:
@edolstra ping

src/libexpr/eval.cc
```diff
@@ -311,7 +311,7 @@ EvalState::EvalState(const Strings & _searchPath, ref<Store> store)
     addToSearchPath("nix=" + settings.nixDataDir + "/nix/corepkgs");
 
     if (settings.restrictEval || settings.pureEval) {
-        allowedPaths = PathSet();
+        allowedPaths = {canonPath(settings.nixDataDir + "/nix/corepkgs", true)};
```
edolstra: Isn't the correct fix to do
```
addToSearchPath("nix=" + canonPath(settings.nixDataDir + "/nix/corepkgs"));
 ```
?

shlevy: Sure, I'll switch to that.

shlevy:
@edolstra Fixed

PR #1816, opened 2018-01-25
Add path primop.

shlevy:
buiiltins.path allows specifying the name of a path (which makes paths
with store-illegal names now addable), allows adding paths with flat
instead of recursive hashes, allows specifying a filter (so is a
generalization of filterSource), and allows specifying an expected
hash (enabling safe path adding in pure mode).


shlevy:
Tested locally, will add tests and docs if this is deemed OK

copumpkin:
This is a new take on #1784, right?

shlevy:
Yep!

shlevy:
Ping

grahamc:
@shlevy would you be able to write up a bit of docs for this function?

shlevy:
If it's otherwise acceptable yeah

src/libexpr/primops.cc
```diff
@@ -1009,20 +1009,16 @@ static void prim_toFile(EvalState & state, const Pos & pos, Value * * args, Valu
...
-    PathFilter filter = [&](const Path & path) {
+static void addPath(EvalState & state, const Pos & pos, const string & name, const Path & path_,
+    Value * filterFun, bool recursive, const Hash & expectedHash, Value & v)
+{
+    /* !!! Possibly unsafe to bypass restricted eval even with
+       expected hash... you can confirm some file with known hash
+       exists at an arbitrary path. */
+    const auto path = expectedHash ?
+        path_ :
+        state.checkSourcePath(path_);
```
edolstra: This looks wrong. In `--restricted-eval`, the path should be checked even if a hash is given. In `--pure-eval`, the path doesn't need to be checked but it requires a hash.

src/libexpr/primops.cc
```diff
@@ -2071,6 +2129,7 @@ void EvalState::createBaseEnv()
     addPrimOp("__fromJSON", 1, prim_fromJSON);
     addPrimOp("__toFile", 2, prim_toFile);
     addPrimOp("__filterSource", 2, prim_filterSource);
+    addPrimOp("__addPath", 1, prim_addPath);
```
edolstra: I'm not a fan of the name `addPath` (add to what?). Maybe it should just be `builtins.path`? That has a less imperative ring to it.

edolstra:
Looks good apart from 2 comments.

shlevy:
@edolstra @grahamc Updated, added tests and docs.

PR #1817, opened 2018-01-26
HttpBinaryCacheStore: Support upsertFile with PUT.

adelbertc:
Some servers, such as Artifactory, allow uploading with PUT and BASIC
auth. This allows nix copy to work to upload binaries to those
servers.

Worked on together with @adelbertc


shlevy:
WIP, should switch to curl multi and possibly combine with download.cc

shlevy:
OK, this is good now

Issue #179, opened 2013-12-02
Chroot builds are slow


edolstra:
Chroot builds have a significant overhead. For instance, this expression:

```
with import <nixpkgs> {};
with lib;
let deps = map (n: runCommand "depM-${toString n}" {} "touch $out") (range 1 100);
in runCommand "foo" { x = deps; } "touch $out"
```

(i.e. a trivial build with 100 trivial dependencies) takes 4.7s to build on my laptop without chroot, but 39.6s with chroot.

The main overhead seems to be in setting up the chroot (i.e. all the bind mounts), but the various CLONE_\* flags also have significant overhead.

Unfortunately, there is not much we can do about this since it's all in the kernel, but it does mean we can't enable chroot builds by default on NixOS.

This is on Linux 3.4.70.

aristidb:
Computer have become faster in the past 2 years. We should re-evaluate whether the speed is really worth the significant impurities.

Note that the fact that NixOS default Hydra not using chroot leads to packages "randomly" failing to build locally for those who do use it.

So at least Hydra should enable it.

edolstra:
Hydra does use it. It's the other way around, users like me might not have it enabled and think that a package builds properly when it doesn't. (Happened today with a PHP update, which turns out to do a download during its build.)

iElectric:
Yes, leaving our deterministic promise aside for a sake of some small overhead.

benley:
When nix sets up chroots, is most of the time spent setting up bind mounts?  Or does it do a lot of file copying too?  If the latter, have you considered using something like Cowdancer (http://www.netfort.gr.jp/~dancer/software/cowdancer.html.en) to get copy-on-write bind mounts?  It's low-overhead and fast to set up.  Debian uses it in cowbuilder/pbuilder, which makes for an excellent ephemeral-chrooted build system.

vcunat:
@benley: COW isn't needed, as all accessible in the chroot is read-only anyway. From the comments it seems noone has analyzed precisely what's the main cost, but bind mounts are suspected (and they probably were never meant to be used so massively).

copumpkin:
Has anyone looked into [proot](http://proot.me/) for this purpose?

edolstra:
"PRoot uses the CPU emulator QEMU user-mode to execute transparently guest programs." I doubt that's faster than bind mounts :-)

vcunat:
[...]

PR #1821, opened 2018-01-26
Turn sandbox on by default on Darwin

copumpkin:
Is this premature? @edolstra @LnL7 @grahamc 

I think we still have an outstanding `nix-env` bug to fix (the sandbox profile in the buildenv.nix corepkg), but otherwise I think 1.12 is fine on Darwin with nixpkgs master.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -15,6 +15,13 @@ typedef enum { smEnabled, smRelaxed, smDisabled } SandboxMode;
 
 extern bool useCaseHack; // FIXME
 
+/* There's no real downside to us enabling the sandbox by default on Darwin */
+#if __APPLE__
+#define smDefault smEnabled
+#else
+#define smDefault smDisabled
+#endif
+
 struct CaseHackSetting : public BaseSetting<bool>
 {
     CaseHackSetting(Config * options,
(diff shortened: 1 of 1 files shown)


LnL7:
I'm afraid there are still some issues with propagating impure dependencies, https://github.com/NixOS/nixpkgs/pull/33756#issuecomment-357065326

copumpkin:
Dammit. I'll leave this here until we address those issues 😄

dtzWill:
Meanwhile Linux-land sandbox is disabled by default despite working wonderfully :sob: .

I know, I know, performance concerns but still it's unfortunate. ("performance!!111oneone")

copumpkin:
I think last time he looked into it, @edolstra found that it was actually the network namespaces that took most of the time. I was noodling about that it seems like we might be able to keep two persistent network namespaces: one empty one, and one that can talk to the internet for fixed-output derivations. Then we just keep reusing them and avoid paying the penalty.

dtzWill:
Interesting!  That does seem rather useful, especially if the cost is their creation/destruction.

(cc #179)

shlevy:
@copumpkin Can two builders talk to each other in the same empty network namespace? I suppose it's fine as long as both ends need to opt-in, but that'd be my onlly concern.

copumpkin:
@shlevy I don't think they can unless we give them capabilities to create interfaces, right?

shlevy:
Well we need loopback for testing right?

copumpkin:
Oh, yeah, that might throw a wrench in things

dtzWill:
Well we could have a pool of namespaces then?

copumpkin:
Yeah, that could work. Or we just accept that builds could technically talk to each other if they listen on loopback, sort of like what we have on Darwin

shlevy:
As long as they can't actually interfere without cooperating I think it's fine. Or we could have as many namespaces as we have build users, whatever

PR #1822, opened 2018-01-27
Improve manual on inheriting attributes

typetetris:
Expands first paragraph a bit
Adds a more comprehensive example


grahamc:
👍 tested and built ok here.

PR #1824, opened 2018-01-28
configure: Use $CPP instead of cpp directly

bgamari:
The latter breaks in the case of cross-compilation, when `cpp` bears a target prefix.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -61,6 +61,7 @@ CFLAGS=
 CXXFLAGS=
 AC_PROG_CC
 AC_PROG_CXX
+AC_PROG_CPP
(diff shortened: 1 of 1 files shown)


shlevy:
Wouldn't it be better to use `AC_PROG_CPP` and `$CPP`?

bgamari:
Yes, probably. Fixed although I'm waiting for a build to test it.

dtzWill:
Woo this fixes nixUnstable cross-compilation for me!

Well, coupled with a few other things (disabling tests on aws-sdk-cpp, etc.).

Result has a few undesirable references but it builds! :)

PR #1827, opened 2018-01-31
Add docstrings to Derivation/SimpleDerivation WIP

Profpatsch:
This should make deciphering the internal representation of a derivation a lot easier.

There’s still a `TODO` and it will probably have to be adapted a bit for 1.12.

cc @edolstra @aszlig

### DIFF
--- a/src/libstore/derivations.hh
+++ b/src/libstore/derivations.hh
@@ -40,18 +40,36 @@ typedef std::map<Path, StringSet> DerivationInputs;
 
 typedef std::map<string, string> StringPairs;
 
+
+/* A Derivation without its closure (sub-derivations).
+   This is used to prevent having to send the closure of a drv file
+   (which could be very large), e.g. for remote build operations. */
 struct BasicDerivation
 {
-    DerivationOutputs outputs; /* keyed on symbolic IDs */
-    PathSet inputSrcs; /* inputs that are sources */
+    /* Keyed on symbolic IDs, like `"out"` or `"doc"`. */
+    DerivationOutputs outputs;
+
+    /* Inputs that are sources. <- TODO what’s a source? */
+    PathSet inputSrcs;
+
+    /* The system architecture this derivation can be built on,
+       as returned by `config/config.guess`, e.g. `"x86_64-linux"` */
     string platform;
+
+    /* Program that executes the build. Either a path to a derivation
+       or a script. */
     Path builder;
+
+    /* Command line arguments passed to the builder. */
     Strings args;
+
+    /* POSIX environment passed to the builder. */
     StringPairs env;
 
     virtual ~BasicDerivation() { };
 };
 
+/* Description of a build action, input for the builder. */
 struct Derivation : BasicDerivation
 {
     DerivationInputs inputDrvs; /* inputs that are sub-derivations */


grahamc:
Eelco would prefer if you sent this to master first, then adapted it for 1.11 in a cherry-picked backport.

src/libstore/derivations.hh
```diff
@@ -40,18 +40,36 @@ typedef std::map<Path, StringSet> DerivationInputs;
...
+   (which could be very large), e.g. for remote build operations. */
 struct BasicDerivation
 {
-    DerivationOutputs outputs; /* keyed on symbolic IDs */
-    PathSet inputSrcs; /* inputs that are sources */
+    /* Keyed on symbolic IDs, like `"out"` or `"doc"`. */
+    DerivationOutputs outputs;
+
+    /* Inputs that are sources. <- TODO what’s a source? */
+    PathSet inputSrcs;
```
grahamc: By process of elimination, if I have `cp ${./foobar} $out` then `./foobar` would be moved to the Nix store, and the Nix store path would be in the srcs.

grahamc:
Just a few comments, overall looks good to me.

src/libstore/derivations.hh
```diff
@@ -40,18 +40,36 @@ typedef std::map<Path, StringSet> DerivationInputs;
...
+
+    /* The system architecture this derivation can be built on,
+       as returned by `config/config.guess`, e.g. `"x86_64-linux"` */
```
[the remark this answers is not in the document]
Profpatsch: Yes, it’s worded ambiguously. The strings follow the naming of the possible values returned by the `config.guess` script. How could that be rephrased?

edolstra:
I don't really see a need to add docstrings to the 1.11 branch.

Profpatsch:
What about the contents of the docstrings? Are they correct? What about the TODO?

src/libstore/derivations.hh
shlevy: Yes, this is correct.

edolstra:
The semantics of store derivations are described in https://nixos.org/~eelco/pubs/phd-thesis.pdf, especially page 40 and 101.

Profpatsch:
Does `DerivationOutputs outputs;` contain a map from output names to `DerivationOutput`? That’s not mentioned in the paper.

Since the paper is from a few years back, I’d expect the data structures and even some semantics to have changed quite drastically, especially in nix `2.0`.

grahamc:
Derivation files are forward / backward compatible, as far as I can tell.
This is (what I assume) allows Nix 2.0 to be deployed to all the Hydra
builders.

copumpkin:
Having said all that, having the canonical documentation for a piece of code be an immutable PDF from 12 years ago is probably not the best way to encourage documentation to improve 😄 

Perhaps we can have a slightly more "live" version of this information in a place that can evolve quickly (even if the underlying concepts don't change much, the clarity of their explanations can obviously evolve independently).

It seems like getting something like this into 1.12 documentation, probably with a link back to the dissertation, would be good, but probably not worth making further changes to the 1.11 codebase over. Does that make sense? At least that's my understanding of @edolstra's comment above about putting this against the maintenance branch.

Profpatsch:
> It seems like getting something like this into 1.12 documentation, probably with a link back to the dissertation, would be good, but probably not worth making further changes to the 1.11 codebase over.

Yes, I’d have pushed the docs to both versions (for people like me who want to study 1.11 source while it’s still in use on their system), but if it’s not worth the effort I’m okay with just pushing to 1.12.

PR #1828, opened 2018-01-31
Add isPath primop

zimbatm:
this is added for completeness' sake since all the other possible
`builtins.typeOf` results have a corresponding `builtins.is<Type>`


zimbatm:
ping

Issue #1841, opened 2018-02-05
Safe definition of extra builtins outside of nix proper


shlevy:
We currently have several impure or potentially impure builtins hard-coded into the nix codebase: fetch{Tarball,Git,Mercurial,url}, paths, etc. Having them defined in nix itself makes it very hard to iterate on them (e.g. the open issues to improve/fix fetchGit, adding addPath) and effectively impossible to add more niche uses (e.g. fetchFromHackage, which is pure so long as you trust Hackage's promise to keep tarballs immutable).

I propose we add (modulo bikeshedding) `builtins.extraBuiltins`, which is populated by importing `$NIX_CONF_DIR/extra-builtins.nix` and passing in `{ exec, importNative, settings }`, where `exec` and `importNative` are the relevant builtins even if `enableNativeCode` is off and `settings` is some nix-usable serialization of the current evaluation's settings (the settings are needed so the user builtins can switch off of `pure-eval` etc. and could be passed just by adding a generally available `builtins.settings`). The user could then implement arbitrarily complex policies (ranging from a fixed list of builtins imported from a fixed nixpkgs, which could be safely usable even by hydra in restricted mode, to a fancy whitelist scheme listing which projects can define their own builtins) without simply turning on `builtins.exec/importNative` globally.

I'm happy to implement this, and I'd really like this in 2.x if possible so we have much less need to have our users rely on a pre-release or a fork if we need to improve/add a builtin.

shlevy:
@edolstra Please let me know

shlevy:
@copumpkin I think this is a poor hacker's version of #520

grahamc:
I've had similar thoughts along these lines, but depending on the PATH of the caller to Nix. This is a more heavy-weight implementation, and I'm not sure the trade-offs it makes, but programs in the PATH with the name `nix-fetcher-git` would be exposed at `builtins.fetcher.git`.

In a Nix file, the user would write:

```nix
builtins.fetcher.hdfs {
  url = "my-url";`
  arbitrary = {
    other = "attributes";
  };
  sha256 = "..."; # perhaps mandated by Nix 
}
```

Nix would then effectively run:

```sh
echo '{"url": "my-url", "arbitrary": { "other": "attributes" }, "sha256": ".."; }' | nix-fetcher-hdfs
```

nix-fetcher-hdfs could be:

```sh
#!/bin/sh
url=$(cat | jq .url);
scratch=$(mktemp -d)
[...]

PR #1844, opened 2018-02-06
Add user-defined extra builtins.

shlevy:
To enable quick iteration on builtin functions, or adding builtins
that aren't suitable for inclusion upstream (due to limited audience
or proprietary code), without the security concerns entailed by
enabling 'allow-unsafe-native-code-during-evaluation', users can now
define a function in a nix expression at the path specified by the
'extra-builtins-file' nix setting (default
$NIX_CONF_DIR/extra-builtins.nix). That function takes a set
containing at least the 'exec' and 'importNative' attributes,
corresponding to the relevant builtins, even when
'allow-unsafe-native-code-during-evaluation' is false, and the result
is available as 'builtins.extraBuiltins'.

See tests/extra-builtins/extra-builtins.nix for an example, and the
nix manual for full details.

### DIFF
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -741,6 +741,34 @@ builtins.fetchurl {
 
   </varlistentry>
 
+  <varlistentry xml:id="conf-enable-native-code"><term><literal>allow-unsafe-native-code-during-evaluation</literal></term>
+
+    <listitem>
+
+      <para>If set to <literal>true</literal>, extra functions that allow
+      nix expressions to extend the language are added to the <literal>
+      builtins</literal> set. Because these functions entail running arbitrary
+      code as the evaluating user, this option should only be enabled when
+      all evaluated nix expressions are fully trusted! Defaults to <literal>
+      false</literal>.</para>
+
+    </listitem>
+
+  </varlistentry>
+
+  <varlistentry xml:id="conf-extra-builtins-file"><term><literal>extra-builtins-file</literal></term>
+
+    <listitem>
+
+      <para>An absolute path to a Nix expression specifying
+      <link linkend="builtin-extraBuiltins">extraBuiltins</link>. See
+      the documentation of that builtin for more details. Defaults to
+      <filename>$NIX_CONF_DIR/extra-builtins.nix</filename>.</para>
+
+    </listitem>
+
+  </varlistentry>
+
 
 </variablelist>
 
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -1269,7 +1269,65 @@ stdenv.mkDerivation (rec {
 
   </varlistentry>
 
+  <varlistentry><term><function>builtins.exec</function>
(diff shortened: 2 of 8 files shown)


grahamc:
- Can you provide an example implementation of a fetcher?
 - Are fetchers adding things to the store subject to hash checks?

grahamc:
re #1841

shlevy:
Fetchers aren't trivial... I'm not sure what value would be added by having another example here over what's in `primops{.cc/*.cc}` or in nixpkgs. Can you expand more?

Fetchers defined in `extra-builtins` can do hash checks or not, they can respect `pure-eval` or not, whatever. If you want a more limited interface you can build it on top of this. (of course, anything added to the store will be hash-addressed).

grahamc:
For  the first point, I didn't mean to say fetcher, but also I completely missed (due to how short it is!) the example in the test. Very cool!

For the second point, I wanted to be sure a user builtin couldn't lie about the contents of a file and put data in the store where it doesn't belong. From IRC, the answer is no.

Nice!

edolstra:
I much prefer having a `plugin` config option (see https://github.com/NixOS/nix/issues/1841#issuecomment-363107363). That's also more general because it supports other kinds of extensions (such as new Store subclasses, e.g. S3BinaryCacheStore could be moved into a plugin).

copumpkin:
@edolstra it does seem like the laziness @shlevy is talking about in the comments below the one you linked to is appealing. My dream is for `builtins.fetchGit` to actually depend on `git` (so the relevant plugin and its git dependency are expressed in Nix), but for Nix not to. A static loadable plugins architecture would force the `fetchGit` plugin (and its git dependency) to be loaded whenever Nix is loaded.

Couldn't we just support both mechanisms? It does seem appealing to split out the `S3BinaryCacheStore` as you say, but nothing says that needs to work the same way as the extra builtins, if we don't make a mess in the code by supporting both.

edolstra:
Another issue is that there is no guarantee that the plugins from `extra-builtins.nix` are compatible. It's up to the user to ensure that the Nix version used by `extra-builtins.nix` is the same as the caller's. OTOH, with a NixOS option `nix.plugins`, it's guaranteed that Nix and the plugins come from the same Nixpkgs.

shlevy:
Working on plugins now, @copumpkin my first plugin will be one to support extraBuiltins 😉

shlevy:
@edolstra True, though that doesn't apply to extraBuiltins defined with builtins.exec.

shlevy:
And we have use cases to support besides NixOS

shlevy:
@copumpkin https://github.com/shlevy/nix-plugins/tree/nix-2x-plugins has an implementation of extra-builtins on top of #1855, though it doesn't add the setting (you have to use $NIX_CONF_DIR/extra-builtins.nix, but you can set NIX_CONF_DIR so...)

copumpkin:
For anyone following along confused, we got this through https://github.com/NixOS/nix/pull/1854 instead.

shlevy:
Oops, sorry, thanks!

PR #1847, opened 2018-02-06
Added documentation generation with doxygen

taktoa:
This doesn't integrate the Doxygen generation into the build, but it does make it a bit easier to generate the Doxygen documentation. Should be mergeable without much review, as it only adds files.

### DIFF
--- a/doc/doxygen/Doxyfile
+++ b/doc/doxygen/Doxyfile
@@ -1,0 +1,8 @@
+RECURSIVE = YES
+INPUT = ../../src/
+PROJECT_NAME = nix
+
+HTML_HEADER = include/header.html
+HTML_FOOTER = include/footer.html
+HTML_EXTRA_STYLESHEET = include/extra.css
+HTML_EXTRA_FILES = include/extra.js
--- a/doc/doxygen/include/extra.css
+++ b/doc/doxygen/include/extra.css
@@ -1,0 +1,255 @@
+h1, .h1, h2, .h2, h3, .h3 {
+    font-weight: 200 !important;
+}
+
+#navrow1, #navrow2, #navrow3, #navrow4, #navrow5{
+    border-bottom: 1px solid #EEEEEE;
+}
+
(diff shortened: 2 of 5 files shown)


grahamc:
I'm surprised to see this as part of Nix itself and not a website about nix. Does it really belong here?

taktoa:
It's useful for people hacking on Nix, so you will often want to generate it from your modified checkout of the source.

copumpkin:
Where did those files come from? They seem generated or standard, right? If generated, can Nix generate them for us and not pollute the repo with generated code? If they're from somewhere else, can we just make a fetcher?

taktoa:
They're neither generated nor standard; they're modified from https://github.com/Velron/doxygen-bootstrapped

taktoa:
Oh, wow, it's been a while since I made these files; it seems that they have modified `doxygen-bootstrapped` to the extent that it is superior to my modified version of one of their old commits.
A fetcher is fairly reasonable in that case, I guess.

copumpkin:
Did you want to switch it to a fetcher-based doxygen model then? That might be quite appealing.

dunnl:
I am interested in producing good nix documentation. But [https://github.com/Velron/doxygen-bootstrapped](https://github.com/Velron/doxygen-bootstrapped) hasn't been updated in 2 years and is broken w.r.t. doxygen >= 1.8.12 (current version is 1.8.14). This [misnamed branch](https://github.com/Velron/doxygen-bootstrapped/tree/feature/support-doxygen-1.1.12+) looks experimental and equally stale.

Bigger issue, I think the nix source isn't doxygen-friendly. I can only get halfway useful doxygen docs with EXTRACT_ALL = YES.

Am I mistaken to think this requires some non-trivial work on the actual source annotation? Besides work on the bootstrap stuff and builder integration.

domenkozar:
Closing as we're now going with markdown and probably another option for code/doc generation.

PR #1851, opened 2018-02-07
More completely fix recursive nix, unbreak tests

dtzWill:
See:
https://github.com/NixOS/nix/commit/88b5d0c8e89afefbc547b6243c5aa5a3ec8176e9#commitcomment-27406365


shlevy:
Does this actually unbreak the tests?

dtzWill:
It sure does locally! But by all means please confirm :+1:.

(I didn't expect it, by itself, would fix the tests either, but.... since it did hooray!)

PR #1854, opened 2018-02-08
Add plugins support

shlevy:
Alternative solution to #1841 / #1844.

### DIFF
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -64,6 +64,8 @@ int main (int argc, char * * argv)
 
         settings.maxBuildJobs.set("1"); // hack to make tests with local?root= work
 
+        initPlugins();
+
         auto store = openStore().cast<LocalStore>();
 
         /* It would be more appropriate to use $XDG_RUNTIME_DIR, since
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ perl/Makefile.config
 /scripts/nix-copy-closure
 /scripts/nix-reduce-build
 /scripts/nix-http-export.cgi
+/scripts/nix-profile-daemon.sh
 
 # /src/libexpr/
 /src/libexpr/lexer-tab.cc
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,8 @@ makefiles = \
   misc/launchd/local.mk \
   misc/upstart/local.mk \
   doc/manual/local.mk \
-  tests/local.mk
+  tests/local.mk \
+  tests/plugins/local.mk
 
 GLOBAL_CXXFLAGS += -std=c++14 -g -Wall -include config.h
 
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -742,6 +742,33 @@ builtins.fetchurl {
   </varlistentry>
 
 
+  <varlistentry xml:id="conf-plugin-files">
+    <term><literal>plugin-files</literal></term>
+    <listitem>
+      <para>
+        A list of plugin files to be loaded by Nix. Each of these
+        files will be dlopened by Nix, allowing them to affect
+        execution through static initialization. In particular, these
+        plugins may construct static instances of RegisterPrimOp to
+        add new primops to the expression language,
+        RegisterStoreImplementation to add new store implementations,
+        and RegisterCommand to add new subcommands to the
+        <literal>nix</literal> command. See the constructors for those
+        types for more details.
+      </para>
+      <para>
+	Since these files are loaded into the same address space as
+        Nix itself, they must be DSOs compatible with the instance of
+        Nix running at the time (i.e. compiled against the same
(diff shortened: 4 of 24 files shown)


src/build-remote/build-remote.cc
```diff
@@ -64,6 +65,8 @@ int main (int argc, char * * argv)
 
         settings.maxBuildJobs.set("1"); // hack to make tests with local?root= work
 
+        initPlugins();
```
edolstra: Why are all these `initPlugins()` calls necessary? Can't that be done in `initNix()` or some other common location?

shlevy: They need to be called after `--option` flags are parsed, which is done in a bunch of different ways by different programs.

src/libexpr/primops/fetchMercurial.cc
```diff
@@ -204,6 +203,11 @@ static void prim_fetchMercurial(EvalState & state, const Pos & pos, Value * * ar
         state.allowedPaths->insert(hgInfo.storePath);
 }
 
-static RegisterPrimOp r("fetchMercurial", 1, prim_fetchMercurial);
+static void reg(EvalState & state)
+{
+    state.addPrimOp("fetchMercurial", 1, prim_fetchMercurial);
+}
+
+static RegisterPlugin r(Plugin(PluginTag<PluginType::EvalPlugin>(), reg));
```
edolstra: I don't understand why these changes are necessary. Doesn't the `RegisterPrimOp` constructor run automatically when the plugin is loaded? Why the need for a `RegisterPlugin` with a lot of template crap?

shlevy: It's not necessary, this is all contained in the second commit that I explicitly said could be dropped ;) This was just to drop duplicate capabilities.

Also, a single template is hardly "a lot of template crap".

shlevy: ... Though I just realized an easy way to avoid the template altogether, patch incoming.

shlevy:
@edolstra Template crap deleted, second commit simplified (though still not needed)

src/libexpr/primops/fetchMercurial.cc
edolstra: Well, I don't understand the need for all this over-engineering (e.g. `Plugin(PluginTag<PluginType::EvalPlugin>()`), plus all that complexity in `plugins.hh`, which for some reason has a forward-decl to `EvalState`...

src/libexpr/primops/fetchGit.cc
```diff
@@ -238,6 +237,11 @@ static void prim_fetchGit(EvalState & state, const Pos & pos, Value * * args, Va
         state.allowedPaths->insert(gitInfo.storePath);
 }
 
-static RegisterPrimOp r("fetchGit", 1, prim_fetchGit);
+static void reg(EvalState & state)
```
shlevy: This could be inlined with a lambda if desired.

src/libexpr/primops/fetchMercurial.cc
shlevy: I removed the tag stuff, was a holdover from an earlier approach. `plugins.hh` needs to reference `EvalState &` to definte the `EvalPlugin` type, but #including `eval.hh` this low in the header chain could easily lead to accidental references to libnixexpr in libnixstore, if it even worked otherwise.

shlevy:
@edolstra Continuing out of the source  comments: `plugins.hh` is ready to support multiple plugin types for two reasons: 1) You mentioned wanting to move s3binarycachestore to plugins and 2) it's very cheap forward compatibility.

edolstra:
Well, I still don't understand why RegisterPlugin is needed at all (and why plugin.cc needs to know about EvalState). The *only* thing that should be needed for plugin support is to do `dlopen` on the .so files, the `RegisterPrimOp` and `RegisterStoreImplementation` constructors should take care of the registration.

shlevy:
OK, I'll switch tothat approach.

src/buildenv/buildenv.cc
```diff
@@ -135,6 +136,7 @@ int main(int argc, char ** argv)
 {
     return handleExceptions(argv[0], [&]() {
         initNix();
+        initPlugins();
```
edolstra: `buildenv` should not need plugins.

src/resolve-system-dependencies/resolve-system-dependencies.cc
```diff
@@ -158,6 +158,7 @@ int main(int argc, char ** argv)
 {
     return handleExceptions(argv[0], [&]() {
         initNix();
+        initPlugins();
```
edolstra: This command probably does not need plugins.

shlevy:
@edolstra updated

edolstra:
Merged, thanks!

Feature request: it would be useful to support loading all .so's in a directory, e.g.
```
plugin-files = ${config.system.path}/lib/nix/plugins
```
to load all plugins provided by installed packages.

PR #1857, opened 2018-02-10
configure.ac: check if lzma has MT support, fix deb build/etc.

dtzWill:
Check is simple but seems to do the job on debian at least,
more complicated checks that include linking may be required
but it seems best to keep this simple for now.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -175,6 +175,8 @@ AC_SUBST(HAVE_SODIUM, [$have_sodium])
 
 # Look for liblzma, a required dependency.
 PKG_CHECK_MODULES([LIBLZMA], [liblzma], [CXXFLAGS="$LIBLZMA_CFLAGS $CXXFLAGS"])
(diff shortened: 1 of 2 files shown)


shlevy:
Should this error out if the user requests parallel decompression but it's not available? Probably not, but want to raise the question...

copumpkin:
Maybe issue a warning in that case, since we're now making an effort to be friendly when people ask for nonsense, rather than just silently ignoring them

dtzWill:
Okay, now emits warnings.  Was a bit clumsy and redundant, so refactored a bit to try to improve things...

Not sure I love what I ended up with, but can't spend more on this right now;
included commit that just adds warnings, maybe grab that if preferred.

Also warns when requesting parallel compression with a method other than "xz".

edolstra:
Ping @AmineChikhaoui

dtzWill:
Beep boop pingaling @AmineChikhaoul

PR #1862, opened 2018-02-13
Enable specifying directories in plugin-files.

shlevy:


src/libstore/globals.cc
```diff
@@ -142,12 +142,24 @@ void MaxBuildJobsSetting::set(const std::string & str)
...
-           DSO needed by the action of the plugin. */
-        void *handle =
-            dlopen(pluginFile.c_str(), RTLD_LAZY | RTLD_LOCAL);
-        if (!handle)
-            throw Error(format("could not dynamically open plugin file '%1%': %2%") % pluginFile % dlerror());
+        Paths pluginFiles;
+        try {
+            auto ents = readDirectory(pluginFile);
+            for (auto & ent : ents)
+                pluginFiles.emplace_back(pluginFile + std::move(ent.name));
```
edolstra: Does this move actually accomplish anything? Isn't it a const argument to `+` anyway?

src/libstore/globals.cc
```diff
@@ -142,12 +142,24 @@ void MaxBuildJobsSetting::set(const std::string & str)
...
+                throw;
+            pluginFiles.emplace_back(pluginFile);
+        }
+        for (const auto & file : pluginFiles) {
+            /* handle is purposefully leaked as there may be state in the
+               DSO needed by the action of the plugin. */
+            void *handle =
+                dlopen(file.c_str(), RTLD_LAZY | RTLD_LOCAL);
+            if (!handle)
+                throw Error(format("could not dynamically open plugin file '%1%': %2%") % file % dlerror());
```
edolstra: Note that nowadays you can write
```
throw Error("could not dynamically open plugin file '%s': %s", file, dlerror());
```

src/libstore/globals.cc
shlevy: Erm, right, this was from before I remembered the `ent` was relative. Will fix.

shlevy:
@edolstra fixed

PR #1863, opened 2018-02-13
Allow includes from nix.conf

shlevy:


edolstra:
What's the use case for this?

shlevy:
Providing a package of nix + a common base set of settings (e.g. plugins, binary caches, etc.) in the store while still allowing users to specify their own settings on top.

PR #1867, opened 2018-02-13
Allow plugins to define new settings.

shlevy:


shlevy:
(example use case is implementation of extra builtins that allows the user to specify the path to the extra builtins file)

shlevy:
@edolstra thoughts?

puffnfresh:
Related is https://github.com/NixOS/nix/issues/1875 - maybe this PR fixes it.

shlevy:
Hmm not sure if that's related?

src/libutil/config.cc
```diff
@@ -21,34 +23,37 @@ void Config::addSetting(AbstractSetting * setting)
...
-                initials.erase(i);
+                extras.erase(i);
                 set = true;
             }
         }
     }
 }
 
-void Config::warnUnknownSettings()
+void Config::handleUnknownSettings(bool fatal)
```
puffnfresh: Is this called with `fatal = true` anywhere now?

shlevy: @puffnfresh It wasn't before this commit either, but @edolstra removed the flag ed73d40c3b19dc0581bbf28ef29aad50cab3aaf2

PR #1872, opened 2018-02-14
monitor-fds: Fix on macOS.

### LINKED ISSUE #1871: Daemon doesn't notice client disconnects on macOS
Not sure how to reproduce this exactly, but I've seen several people hit a case on macOS with 1.12pre5788_e3013543 where they cancelled all running builds but build locks were still held and needed to restart the daemon.

shlevy:
Fixes #1871.

### DIFF
--- a/src/libutil/monitor-fd.hh
+++ b/src/libutil/monitor-fd.hh
@@ -21,13 +21,29 @@ public:
     MonitorFdHup(int fd)
     {
         thread = std::thread([fd]() {
-            /* Wait indefinitely until a POLLHUP occurs. */
-            struct pollfd fds[1];
-            fds[0].fd = fd;
-            fds[0].events = 0;
-            if (poll(fds, 1, -1) == -1) abort(); // can't happen
-            assert(fds[0].revents & POLLHUP);
-            triggerInterrupt();
+            while (true) {
+              /* Wait indefinitely until a POLLHUP occurs. */
+              struct pollfd fds[1];
+              fds[0].fd = fd;
+              /* This shouldn't be necessary, but macOS doesn't seem to
+                 like a zeroed out events field.
+                 See rdar://37537852.
+              */
+              fds[0].events = POLLHUP;
+              auto count = poll(fds, 1, -1);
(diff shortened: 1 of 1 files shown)


shlevy:
Testing this on my box now, but based on the simplified tests I fully expect it to work.

src/libutil/monitor-fd.hh
```diff
@@ -24,7 +24,12 @@ public:
             /* Wait indefinitely until a POLLHUP occurs. */
             struct pollfd fds[1];
             fds[0].fd = fd;
-            fds[0].events = 0;
+            /* This shouldn't be necessary, but macOS doesn't seem to
+               like a zeroed out events field.
+               See https://openradar.appspot.com/radar?id=5047895313088512
```
copumpkin: Lovely, thanks!

shlevy:
Ugh, hitting the assertion on the next line, digging in...

shlevy:
poll is timing out even with a `-1` timeout. Going to try select...

shlevy:
Seems related... https://github.com/curl/curl/issues/1057

shlevy:
OK, this is tested and working. Ugly, but so is the problem domain.

src/libutil/monitor-fd.hh
```diff
@@ -21,13 +21,23 @@ public:
...
+              fds[0].fd = fd;
+              /* This shouldn't be necessary, but macOS doesn't seem to
+                 like a zeroed out events field.
+                 See https://openradar.appspot.com/radar?id=5047895313088512
+                 and https://bugreport.apple.com/web/?problemID=37537852.
+              */
+              fds[0].events = POLLHUP;
+              auto count = poll(fds, 1, -1);
+              if (count == -1) abort(); // can't happen
+              if (count == 0) continue; // shouldn't happen, but macOS is terrbile
```
shlevy: We may want to insert a delay here if this turns out to cause lots of spinning.

shlevy:
Going to file another mac bug and will update the comment with the links, but we can merge this whenever and I'll add it when it's available.

Issue #1627, opened 2017-10-23
nix-env --help No manual entry for nix-env


doublehp:
(bad characters are console locale issue; try to ignore them)

```
1$ nix-env -h
error: no operation specified
Try ânix-env --helpâ for more information.
2$ nix-env --help
No manual entry for nix-env
3$ . /home/dhp/.nix-profile/etc/profile.d/nix.sh
4$ nix-env -h
error: no operation specified
Try ânix-env --helpâ for more information.
5$ nix-env --help
No manual entry for nix-env
```

I was expecting at the very least ... a list of valid commands !!!

vyp:
When I run `nix-env --help`, the manpage for nix-env opens for me. My `nix-env --version` is `nix-env (Nix) 1.11.15`.

sgraf812:
Hitting this, too on a fresh nix install on Ubuntu.

I started out by installing through `curl | sh`, which starts out with Ubuntu's `man`, which has no files for `nix-env`. Then I `nix-env -i man` to no avail, followed by `nix-env -i nix` to no avail. Between each invocation I re`source`d my profile. `nix-env --version` is 1.11.16.

matthewbauer:
Running this should fix it:

```
export MANPATH=/nix/var/nix/profiles/default/share/man:$HOME/.nix-profile/share/man:$MANPATH
```

Nix should set this as a backup before running "man" in "showManPage":

https://github.com/NixOS/nix/blob/6ddfe9a999027f9867bbf61ae92c19c591c89a86/src/libmain/shared.cc#L262-L267

/cc @edolstra

svetlyak40wt:
I still have the same problem on a fresh Nix installation. Setting MANPATH does not help.

System info:

 - system: `"x86_64-linux"`
 - host os: `Linux 6.8.0-90-generic, Ubuntu, 24.04.3 LTS (Noble Numbat), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.33.2`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/store/5x63v8377qlcjl7lvaszb3q66hfylp4r-nixpkgs/nixpkgs`

PR #1873, opened 2018-02-15
Set MANPATH in case man path isn’t set correctly.

matthewbauer:
Previously, this would fail at startup for non-NixOS installs:

nix-env --help

The fix for this is to just use "nixManDir" as the value for MANPATH when spawning "man".

To test this, I’m using the following:

$ nix-build release.nix -A build
$ MANPATH= ./result/bin/nix-env --help

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -262,6 +262,7 @@ void printVersion(const string & programName)
 void showManPage(const string & name)
 {
     restoreSignals();
+    setenv("MANPATH", settings.nixManDir.c_str(), 1);
     execlp("man", "man", name.c_str(), NULL);
     throw SysError(format("command 'man %1%' failed") % name.c_str());
 }
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -38,6 +38,7 @@ Settings::Settings()
     , nixConfDir(canonPath(getEnv("NIX_CONF_DIR", NIX_CONF_DIR)))
     , nixLibexecDir(canonPath(getEnv("NIX_LIBEXEC_DIR", NIX_LIBEXEC_DIR)))
     , nixBinDir(canonPath(getEnv("NIX_BIN_DIR", NIX_BIN_DIR)))
+    , nixManDir(canonPath(NIX_MAN_DIR))
     , nixDaemonSocketFile(canonPath(nixStateDir + DEFAULT_SOCKET_PATH))
 {
     buildUsersGroup = getuid() == 0 ? "nixbld" : "";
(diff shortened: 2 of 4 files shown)


edolstra:
Not sure if we should introduce another environment variable (`NIX_MAN_DIR`). Otherwise looks good.

matthewbauer:
> Not sure if we should introduce another environment variable (NIX_MAN_DIR)

Fixed. Just removed that environment variable check now.

catern:
Hmm, this might be wrong... some pagers (which might be used as MANPAGER) are capable of switching to displaying other man pages after starting up. By setting MANPATH here, we prevent that (for any non-nix man page). Also, this pollutes the environment under man, so if those pagers want to spawn subprocesses, those subprocesses will also carry that polluted environment.

It would be better to pass -M (which overrides manpath) to man. Then it wouldn't pollute the environment of pagers, nor prevent switching to non-Nix man pages. Unfortunately it wouldn't help with switching within a pager to other Nix man pages, if the manpath configuration is indeed set up incorrectly. I think that would be fine though.

(BTW, I don't think that it should normally be true that nix-env --help fails on non-NixOS installs. If the Nix bin directory is on your path, then man knows to add the sibling directory for share/man. So what kind of install did you have, that that did not happen?)

matthewbauer:
> It would be better to pass -M (which overrides manpath) to man.

I guess switch to that if it's supported on most versions of the 'man' command. It seemed to me like it should have a similar effect. Maybe it doesn't if you are launching subprocesses from man though?

> If the Nix bin directory is on your path, then man knows to add the sibling directory for share/man

That might be true if MANPATH isn't set at all but definitely not in all cases. At least it's working incorrectly for some people (#1627). I thought it might have something to do with the multiple outputs of Nix (man and bin have separate drvs). Also possibly because of different versions of "man" (BSD vs GNU).

Anyway MANPATH is set in nix-profile.sh for Nix 2.0 but not for Nix 1.12 (see https://github.com/NixOS/nix/blob/1.11-maintenance/scripts/nix-profile.sh.in). In addition multi-user installs are not setting MANPATH at all (#1782).

catern:
> Maybe it doesn't if you are launching subprocesses from man though?
Well, man always launches a subprocess: The pager. Like I said, -M works better in the scenario where you view other manpages from your pager after viewing the Nix manpage, as well as when your pager launches subprocesses. (For example, your pager might be emacsclient, which starts up emacs --daemon on demand when first run.)

>So it's probably dependent on configs.
If you've set `MANPATH=/usr/share/man nix-env --help` then you have set MANPATH incorrectly :) You need to set `/usr/share/man:` (note the trailing `:`) so that manpath appends its calculated value. See `manpath(1)` and `manpath(5)` for details about this.

I'm not sure of all the details of this computation in BSD vs GNU `man`. I'm confident now that this change to set MANPATH in nix-env --help is wrong, though.

PR #1884, opened 2018-02-19
libexpr: Don't create lots of temporary strings in Bindings::lexicographicOrder

dezgeg:
Avoids ~180,000 string temporaries created when evaluating a headless
NixOS system.


shlevy:
:astonished:

LnL7:
\o/

PR #1886, opened 2018-02-20
AFL-fuzz hang/crash fixes

dezgeg:
Fix three issues AFL found in ~30 hours.

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -1216,7 +1216,7 @@ std::string filterANSIEscapes(const std::string & s, unsigned int width)
 
         else if (*i == '\r')
             // do nothing for now
-            ;
+            i++;
 
         else {
             t += *i++; w++;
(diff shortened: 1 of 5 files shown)


copumpkin:
Nice! Have any information on your AFL setup for Nix? Might also be fun to try on macOS due to the different codepaths we exercise.

dtzWill:
This is great!

Next stop.... [OSS-Fuzz](https://github.com/google/oss-fuzz)?

src/libutil/util.cc
```diff
@@ -1216,7 +1216,7 @@ std::string filterANSIEscapes(const std::string & s, unsigned int width)
 
         else if (*i == '\r')
             // do nothing for now
-            ;
+            i++;
```
shlevy: Nice, would love a C++ unit testing framework to put this in.

edolstra:
The fix for an infinite loop in replaceString caused an infinite loop: e2d71bd1862cdda65f957c6a6c41016258d9e003.

dezgeg:
Eep. Will take a look.

dezgeg:
@copumpkin The setup is at https://github.com/dezgeg/fuzz-nix. However I don't actually try to build the (potential) generated derivations, just evaluate them with `nix-instantiate --dry-run --read-only-mode`.

PR #1916, opened 2018-02-27
Support extra compatible architectures

lheckemann:
This is based on @cleverca22's patch (cleverca22/nix-misc). It allows building for additional architectures locally. This can be useful with binfmt-misc or compatible architectures (e.g. ARM32-capable aarch64 machines). Additionally, it enables 32-bit ARM seccomp support.

Questions:
 - Thoughts on disabling i686-linux on x86_64-linux by default?
 - What about the seccomp setting? I experienced crashes caused by seccomp while testing on an ARM32-capable aarch64 machine, but I'm not sure how this will behave on non-ARM32-capable ones — in case it fails, I didn't throw an error. I'm not sure if this is appropriate, feedback would be appreciated.

### DIFF
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -254,6 +254,25 @@ false</literal>.</para>
   </varlistentry>
 
 
+  <varlistentry><term><literal>extra-platforms</literal></term>
+
+    <listitem><para>Platforms other than the native one which
+    this machine is capable of building for. This can be useful for
+    supporting additional architectures on compatible machines:
+    i686-linux can be built on x86_64-linux machines (and the default
+    for this setting reflects this); armv7 is backwards-compatible with
+    armv6 and armv5tel; some aarch64 machines can also natively run
+    32-bit ARM code; and qemu-user may be used to support non-native
+    platforms (though this may be slow and buggy). Most values for this
+    are not enabled by default because build systems will often
+    misdetect the target platform and generate incompatible code, so you
+    may wish to cross-check the results of using this option against
+    proper natively-built versions of your
+    derivations.</para></listitem>
+
+  </varlistentry>
+
+
   <varlistentry><term><literal>extra-substituters</literal></term>
 
     <listitem><para>Additional binary caches appended to those
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -295,6 +295,9 @@ public:
         "Nix store has a valid signature (that is, one signed using a key "
         "listed in 'trusted-public-keys'."};
 
+    Setting<StringSet> extraPlatforms{this, StringSet{}, "build-extra-platforms",
+        "Additional platforms that can be built on the local system, e.g. using qemu-user."};
+
     Setting<Strings> substituters{this,
         nixStore == "/nix/store" ? Strings{"https://cache.nixos.org/"} : Strings(),
         "substituters",
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -295,8 +295,12 @@ public:
         "Nix store has a valid signature (that is, one signed using a key "
         "listed in 'trusted-public-keys'."};
 
-    Setting<StringSet> extraPlatforms{this, StringSet{}, "build-extra-platforms",
-        "Additional platforms that can be built on the local system, e.g. using qemu-user."};
+    Setting<StringSet> extraPlatforms{this,
+        SYSTEM == "x86_64-linux" ? StringSet{"i686-linux"} : StringSet{},
+        "extra-platforms",
+        "Additional platforms that can be built on the local system. "
+        "These may be supported natively (e.g. armv7 on some aarch64 CPUs "
+        "or using hacks like qemu-user."};
 
     Setting<Strings> substituters{this,
         nixStore == "/nix/store" ? Strings{"https://cache.nixos.org/"} : Strings(),
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -98,7 +98,9 @@ int main (int argc, char * * argv)
             source >> drvPath;
             auto requiredFeatures = readStrings<std::set<std::string>>(source);
 
-            auto canBuildLocally = amWilling && (neededSystem == settings.thisSystem);
+            auto canBuildLocally = amWilling
+                &&  (  neededSystem == settings.thisSystem
+                    || settings.extraPlatforms.get().count(neededSystem) > 0);
 
             /* Error ignored here, will be caught later */
             mkdir(currentLoad.c_str(), 0777);
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2491,6 +2491,10 @@ void setupSeccomp()
         seccomp_arch_add(ctx, SCMP_ARCH_X32) != 0)
         throw SysError("unable to add X32 seccomp architecture");
 
+    if (settings.thisSystem == "aarch64-linux" &&
+        seccomp_arch_add(ctx, SCMP_ARCH_ARM) != 0)
+        printError("unsable to add ARM seccomp architecture; this may result in spurious build failures if running 32-bit ARM processes.");
+
     /* Prevent builders from creating setuid/setgid binaries. */
     for (int perm : { S_ISUID, S_ISGID }) {
         if (seccomp_rule_add(ctx, SCMP_ACT_ERRNO(EPERM), SCMP_SYS(chmod), 1,
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -57,16 +57,8 @@ bool BasicDerivation::isBuiltin() const
 bool BasicDerivation::canBuildLocally() const
 {
     return platform == settings.thisSystem
-        || isBuiltin()
-#if __linux__
-        || (platform == "i686-linux" && settings.thisSystem == "x86_64-linux")
-        || (platform == "armv6l-linux" && settings.thisSystem == "armv7l-linux")
-        || (platform == "armv5tel-linux" && (settings.thisSystem == "armv7l-linux" || settings.thisSystem == "armv6l-linux"))
-#elif __FreeBSD__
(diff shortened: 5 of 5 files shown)


src/libstore/globals.hh
```diff
@@ -295,6 +295,9 @@ public:
         "Nix store has a valid signature (that is, one signed using a key "
         "listed in 'trusted-public-keys'."};
 
+    Setting<StringSet> extraPlatforms{this, StringSet{}, "build-extra-platforms",
```
shlevy: IMO this should default ti `i686-linux` on `x86_64-linux`

bjornfor:
> partly because I don't really trust qemu-user enough (though that's a gut feeling that's not well-founded).

That is my impression too. Relevant: https://github.com/NixOS/nixpkgs/issues/33553.

src/libstore/globals.hh
lheckemann: Makes sense.

lheckemann:
Oh yes, the docs. Whoops :)

As for its actual use — I'm not saying this absolutely should be used with qemu-user; my primary motivation was to allow ARM32-capable aarch64 machines to build ARM32 stuff. Another option for this would of course be to have both 32-bit and 64-bit nix installed on the machine, but at that point working out which to use, especially in automated invocations, could get hairy.

shlevy:
Oh, yeah, that's a solid use case :+1:

src/libstore/globals.hh
Ericson2314: Shouldn't we have all the platforms from https://github.com/NixOS/nix/pull/1916/files#diff-81a5a969aed0f3f9d6dc4e99b92925cdL60 ?

lheckemann: I'm not sure having all of them is a good idea, given how unreliable actually building them is: building armv6 stuff on an armv7 machine will often result in code that doesn't actually run on armv6 because of annoying autodetection in various build systems. Users who want to risk that can easily set this setting as appropriate.

cleverca22:
another edgecase this can fix, some of the kernel options in nixos lack 32bit x86 support, and there is currently no way to tell nix about that

but if the automatic 64bit supports 32bit too code is removed, we can declare the 32bit support in the config file

lheckemann:
Added some documentation, renamed the option, added i686-linux on x86_64-linux by default. I'm not 100% happy with the description, advice on improving it would be appreciated. I'll also squash the commits before this gets merged.

edolstra:
I don't really see the use case for this option. Nix needs to know how to build something for a particular platform type. For example, on `x86_64-linux`, we can support `i686-linux` because there is a `personality()` syscall to do so. Merely adding a platform to an `extra-platforms` option isn't enough.

doc/manual/command-ref/conf-file.xml
```diff
@@ -254,6 +254,25 @@ false</literal>.</para>
...
+    i686-linux can be built on x86_64-linux machines (and the default
+    for this setting reflects this); armv7 is backwards-compatible with
+    armv6 and armv5tel; some aarch64 machines can also natively run
+    32-bit ARM code; and qemu-user may be used to support non-native
+    platforms (though this may be slow and buggy). Most values for this
+    are not enabled by default because build systems will often
+    misdetect the target platform and generate incompatible code, so you
+    may wish to cross-check the results of using this option against
+    proper natively-built versions of your
+    derivations.</para></listitem>
```
edolstra: This warning suggests that `extra-platforms` is a really bad idea. I mean, it allows you to completely lie to Nix about the platform type (e.g. do a `x86_64-darwin` build on `x86_64-linux`), thus producing a completely wrong build result, which could then end up in a binary cache.

lheckemann:
I've tested this and it works on an aarch64 machine with 32-bit support. Nixpkgs does need quite a few modifications for this to work (nixos/nixpkgs#21471 should fix this properly) correctly, but I have got it producing binaries that work fine on a 32-bit only ARM machine successfully.

edolstra:
Why does Nixpkgs need modification?

lheckemann:
```
localhost:/tmp# nix build -f '<nixpkgs>' --argstr system armv7l-linux hello
[1 built, 1 copied (0.7 MiB), 0.0 MiB DL]
localhost:/tmp# uname -a
Linux localhost 4.9.20 #1 SMP Tue Sep 12 19:34:09 UTC 2017 aarch64 GNU/Linux
localhost:/tmp# file result/
bin/    share/
localhost:/tmp# file result/bin/hello
result/bin/hello: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /nix/store/lhsrzki5bx9bj2ccxibxw5scfckh3p01-glibc-2.27/lib/ld-linux-armhf.so.3, for GNU/Linux 2.6.32, not stripped
localhost:/tmp# result/bin/hello
Hello, world!
```

The reason nixpkgs needs modification is that a lot of packages will incorrectly detect the system (using uname AFAIK) to build for as aarch64 when we actually want to build for armv7 using armv7 tools.

edolstra:
Right, that demonstrates exactly why this option is a bad idea. You could do a armv7 build and end up with a aarch64 binary, a rather significant impurity. For x86_64-linux/i686-linux we have `personality()` to prevent that. Maybe aarch64 has something similar?

IMHO this PR should be reverted. Or at the very least `extra-platforms` should be undocumented.

lheckemann:
>Maybe aarch64 has something similar?

It doesn't, as far as I know.

>You could do a armv7 build and end up with a aarch64 binary, a rather significant impurity.

Exactly the same thing can still be achieved by using an armv7 nix on the same machine, so the absence of this option doesn't eliminate this impurity risk.

edolstra:
> It doesn't, as far as I know.

Maybe seccomp can be used to fake the uname result?

> Exactly the same thing can still be achieved by using an armv7 nix on the same machine, so the absence of this option doesn't eliminate this impurity risk.

Yes, there is the same problem when running i686-linux on x86_64-linux since then it doesn't know that it should use `personality()`. I guess we should move the build-time determination of the platform to runtime.

lheckemann:
>Maybe seccomp can be used to fake the uname result?

Sure, if that works!

>Yes, there is the same problem when running i686-linux on x86_64-linux since then it doesn't know that it should use personality(). I guess we should move the build-time determination of the platform to runtime.

Yes, but I'm also strongly in favour of having nixpkgs tell the build systems what to build for explicitly as per nixos/nixpkgs#21471.

What I don't want is to end up having machines (probably more powerful than any "pure" 32-bit ARM machine) that can run a 32-bit user space but not allowing nix to build 32-bit arm stuff on them, just because they also support aarch64.

Ericson2314:
> Maybe seccomp can be used to fake the uname result?

That + https://github.com/NixOS/nixpkgs/issues/21471 seems like a good thing.

dezgeg:
There is already some discussion on the uname thing at https://github.com/NixOS/nixpkgs/pull/25240. So, there's no `personality()` flag that is good enough. And seccomp (except until this `SECCOMP_RET_USER_NOTIF` someone proposed on LKML goes in) would mean using `ptrace()`, which is pretty much a no-go.  I also don't expect https://github.com/NixOS/nixpkgs/issues/21471 to ever happen (if it means that `configure` scripts will start spewing those `configure: error: cannot run test program while cross compiling.` things at users building native stuff).

What to me sounds like the best next plan is to try exporting `build_alias` to the host triplet in native builds, which should make autotools not call `uname` without hopefully breaking anything else. Sadly, I think some of the known offenders (on the ARMv6-on-ARMv7 case at least) like `perl` and `openssl` don't use autotools...

doc/manual/command-ref/conf-file.xml
cleverca22: if you lie and dont configure things properly, it will simply fail to execute the darwin bash/gcc/bootstrap tools, and never get anywhere

dezgeg:
(Getting a bit off-topic, but...) Well, the condition for that spew is literally:
````
AS_IF([test "$cross_compiling" = yes],
  [m4_default([$4],
	   [AC_MSG_FAILURE([cannot run test program while cross compiling])])],
  [_AC_RUN_IFELSE($@)])
])
````
so there is no way to simultaneously have autotools in cross-compilation mode and have those compile-time tests getting run.

dezgeg:
I tried the quick hack of doing a `export build_alias='armv6l-unknown-linux-gnueabihf'` in stdenv and the results were quite good; Ruby created a `lib/ruby/vendor_ruby/2.4.0/armv6l-linux-eabihf` as expected and GMP stopped looking in `/proc/cpuinfo` and enabling NEON instructions (IIRC GMP currently has an impurity problem on x86 as well where it does similar things). Perl and OpenSSL didn't get fixed as expected (since they don't use autotools).

lopsided98:
I built an armv6l system on armv7l, and it seems to be working well. So far, I've only had to modify openssl and I haven't run into any other illegal instructions yet.

@dezgeg Where did you set `build_alias`? In `mkDerivation`, or somewhere else? I was thinking about setting `configurePlatforms` to `[ "build" ]` by default, but this fails because it passes `--build` to all configure scripts, which often fails if the package does not use autotools. Should `configurePlatforms` set `build/host/target_alias` instead?

dezgeg:
See https://github.com/NixOS/nixpkgs/compare/master...dezgeg:uname-hack.

dezgeg:
Seems like this `SECCOMP_RET_USER_NOTIF` is gaining some traction (https://lwn.net/SubscriberLink/756233/1f23df75d146c393/) so hopefully we get to use it for `uname()` emulation.

In the meantime I've experimented with a dirty kernel hack to look at the `system` environment variable  and base the `uname -m` result on that: https://github.com/dezgeg/nixpkgs/blob/arm-work/pkgs/os-specific/linux/kernel/uname-hack.patch

Don't use yet, it still has some problems with the huge environment variables that happen in nixpkgs stdenv.

lopsided98:
I've been building for armv6l on armv7l using `build_alias` and a fix for openssl to prevent it from detecting the platform. I'm sure there are packages that still have problems, but it has been working well for my use cases. My armv6l machines are so slow that this is the only way I can feasibly run NixOS on them, so I would like to see this configuration be better supported in nixpkgs.

PR #1920, opened 2018-02-28
launchd: enable keepalive for the nix-daemon service

LnL7:
Without this the daemon won't be restarted if the process ever dies, for
example when sending a SIGHUP to reload nix.conf.


LnL7:
Thanks!

PR #1938, opened 2018-03-02
keep-paths: keep store paths alive without touching the filesystem.

shlevy:
Before this, scripts that do a nix build whose validity needn't extend
past the script's lifetime would have to either hope there was no
garbage collection between the build and the usage of it or create a
temporary directory and manage cleanup. With nix keep-paths, path
lifetimes can track script lifetimes (or shorter) without any races or
possibly interrupted cleanup steps.

### DIFF
--- a/src/nix/keep-paths.cc
+++ b/src/nix/keep-paths.cc
@@ -1,0 +1,48 @@
+#include "command.hh"
+#include "util.hh"
+#include "store-api.hh"
+
+using namespace nix;
+
+struct CmdKeepPaths : StoreCommand
+{
+    std::string name() override
+    {
+        return "keep-paths";
+    }
+
+    std::string description() override
+    {
+        return "keeps store paths alive as long as nix keep-paths is alive";
+    }
+
+    void printHelp(const string & programName, std::ostream & out) override
+    {
+        StoreCommand::printHelp(programName, out);
+
+        out << "\n";
+        out << "Reads in newline-separated store-paths over stdin.\n";
+        out << "A newline is printed after the path is registered.\n";
+    }
+
+    void run(ref<Store> store) override
+    {
+        try {
+            while (true) {
+                const auto s = readLine(0);
+                if (s.empty())
+                    continue;
+                if (!store->isStorePath(s)) {
+                    throw Error("%s is not a store path", s);
+                }
+
+                store->addTempRoot(s);
+                writeLine(1, "");
+            }
+        } catch (EndOfFile & e) {
+            return;
+        }
+    }
+};
+
+static RegisterCommand r1(make_ref<CmdKeepPaths>());
--- a/tests/gc.sh
+++ b/tests/gc.sh
@@ -34,6 +34,14 @@ if test -e $drvPath; then false; fi
 
 rm "$NIX_STATE_DIR"/gcroots/foo
 
+coproc nix keep-paths
+echo "$outPath" >&${COPROC[1]}
+read <&${COPROC[0]}
+if nix-store --delete $outPath; then false; fi
+test -e $outPath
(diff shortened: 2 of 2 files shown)


edolstra:
This does not necessarily seem better than using `--out-link` to prevent the result from being garbage-collected. It seems pretty heavy to keep an entire process in memory (along with a nix-daemon worker process) just to keep track of a root.

shlevy:
With `--out-link` though you need to do cleanup. I've found `result` symlinks that weren't cleaned up when a script was interrupted to hard tons of times...

edolstra:
With `nix keep-paths` you have to clean up a process. Leaking a process is potentially much worse than leaking a symlink.

shlevy:
@edolstra You don't have to do anything to clean up the process. If your shell dies, it closes the fd *and* keep-paths gets a HUP.

shlevy:
(my test only closes the fd manually because I wanted to validate gc worked afterwards)

7c6f434c:
On the other hand, doesn't this solution require work on the daemon side while not fully eliminating the race condition? Maybe just nohup a process that (independently of Nix) waits for its `stdin` to get closed, then deletes a directory on `tmpfs`? In that case `--link` logic is supposed to guarantee that the symlink is created before GC has any chance to run.

shlevy:
@7c6f434c What race condition is there? Note that the daemon *already* supports process-lifetime temporary roots (this is how there's no chance of GC between evaluation and building in nix-build).

shlevy:
@7c6f434c If we're spawning a background process already, why not just have something that *actually* tracks roots with process lifetime, rather than hopefully having a chance to clean up a tempdir?

7c6f434c:
@shlevy You need to find `outPath` anyway, and `outPath="$(nix-build --no-out-link something.nix)"` is so convenient it will actually get used. Or `nix build` when it starts to print the `outPath`. You can evaluate `outPath` without instantiating, but that's enough of additional verbosity (and I think you cannot just build the resulting `outPath` without a re-evaluation) that `nix keep-paths` risks to be a feature that people use but usually incorrectly.

shlevy:
@7c6f434c Well if this gets merged the next step would be a `Store` implementation that punts indirect root creation off to `keep-paths`, then you can just do `nix-build` and it Just Works.

But, there's actually no risk even with your approach if you simply re-run nix-build after you register the root. Agreed it's verbose though.

7c6f434c:
I am not sure it will Just Work, as you have to specify the FD at some point as there is no safe FD number to assume (if we include scripts running under ulimit). But yes, it will be nice.

What is a rough outline of the needed flags/open FDs/communication with daemon?

Double-build is one of the multiple available ways that are verbose enough to incentivise Using It Wrong.

shlevy:
```bash
coproc nix keep-paths
nix-build --store keep-paths://auto?in=${COPROC[1]}&out=${COPROC[0]}
```

shlevy:
(obviously the `keep-paths://` protocol doesn't exist yet)

7c6f434c:
And some quoting is also needed.

Frankly, that looks scary enough to wrap, and if I wrap it anyway, I can wrap using a solution that works now: `nix-build` with a link into a temprorary directory, open the output or change directory there, wait for parent to die.

xaverdh:
Still it would be nice to be able to keep store paths in a races free way alive without using the filesystem (e.g. for `nixos-rebuild`). Temporary directories are quite hard to use without introducing new races.

Maybe something like a "file descriptor api" for nix would be handy? Something like this:

Allow passing a file descriptor from the parent process, which is used as a communication channel. The nix process communicates when the build is done and outputs the build result over the `fd`, and then waits for the caller to communicate, that the gc lock on store path can be lifted, before exiting.

PR #1939, opened 2018-03-02
libexpr: Recognize newline in more places in lexer

dezgeg:
Flex's regexes have an annoying feature: the dot matches everything
except a newline. This causes problems for expressions like:

````
"${0}\
"
````

where the backslash-newline combination matches this rule instead of the
intended one mentioned in the comment:

````
    <STRING>\$|\\|\$\\ {
                    /* This can only occur when we reach EOF, otherwise the above
                    (...|\$[^\{\"\\]|\\.|\$\\.)+ would have triggered.
                    This is technically invalid, but we leave the problem to the
                    parser who fails with exact location. */
                    return STR;
                }
````
However, the parser actually accepts the resulting token sequence
(`'"' DOLLAR_CURLY 0 '}' STR '"'`), which is a problem because the lexer
rule didn't assign anything to yylval. Ultimately this leads to a crash
when dereferencing a NULL pointer in ExprConcatStrings::bindVars().

The fix does change the syntax of the language in some corner cases
but I think it's only turning previously invalid (or crashing) syntax
to valid syntax. E.g.

````
"a\
b"
````

and

````
''a''\
b''
````
were previously syntax errors but now both result in "a\nb".

Found by afl-fuzz.


shlevy:
LGTM, would like a second set of eyes to make sure we're not breaking any valid expression with this (or, if we are, that it's worth it).

dezgeg:
cc @edolstra

PR #1940, opened 2018-03-02
Move large buffer allocations to heap instead of stack

dtzWill:
Significantly reduces stack usage and avoids
running over the default limit when using musl.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -4154,9 +4154,9 @@ void Worker::waitForInput()
         assert(goal);
 
         set<int> fds2(j->fds);
+        std::vector<unsigned char> buffer(4096);
         for (auto & k : fds2) {
             if (FD_ISSET(k, &fds)) {
-                std::vector<unsigned char> buffer(4096);
                 ssize_t rd = read(k, buffer.data(), buffer.size());
                 if (rd == -1) {
                     if (errno != EINTR)
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -4156,8 +4156,8 @@ void Worker::waitForInput()
         set<int> fds2(j->fds);
         for (auto & k : fds2) {
             if (FD_ISSET(k, &fds)) {
-                unsigned char buffer[4096];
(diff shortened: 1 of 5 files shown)


edolstra:
Hm, 4 KB and 64 KB doesn't really count as a large buffer IMHO...

edolstra:
Also, if the stack is so small that this is a problem, then you'll have a much bigger issue with  Nix expression evaluation, which recurses very deeply.

dtzWill:
Okay.  glibc defaults to 8MB IIRC, musl only 80k.  So 64k is a lot in that context :).

An alternative to moving these off the stack is to use `pthread_getattr_default_np` to explicitly request/configure a larger default stack size.

Would that be preferred? I'll check if that does the trick re:hitting limit on threads.

shlevy:
I'd love to get to a place where we can fit in 80k stacks, but I think given where we are explicitly asking for a larger one is the way to go.

shlevy:
On the other hand if this is all you need to get things really working and profiling doesn't show a major issue this doesn't seem too bad.

edolstra:
I don't strongly object to this patch per se, but it does illustrate my concerns about adding musl support to Nixpkgs: you get all these little patches that provide absolutely no benefit to the user but are just to work around quirks/limitations of the alternative libc. (I mean, why in 2018, given 64-bit address spaces, would you have a 80 KiB stack by default...)

shlevy:
I think it's probably a good idea anyway to have a sense of what stack space we need and ask for it explicitly. But I agree that if there are a lot of permanent patches like this it's much less attractive.

dtzWill:
Alternative that uses "non-portable" pthread functions to configure default stack size used:

https://github.com/dtzWill/nix/tree/fix/default-stacksize

Fixes problems in nix tools AFAICT, switching my system to it to test nix-daemon now.... yep that seems to do the trick as well.

edolstra:
Merged because it also fixes segfaults on older Boost versions (in Nixpkgs 17.09 but not 18.03). Probably Boost had a smaller default stack size for coroutines.

PR #1941, opened 2018-03-02
anchor vtables of exceptions, helps throwing across DSO boundaries

dtzWill:
At least it does on my special platform :).

As-is the vtables will be anchored in every translation unit that
includes them, so this should reduce size a bit everywhere.

### DIFF
--- a/src/libexpr/json-to-value.cc
+++ b/src/libexpr/json-to-value.cc
@@ -4,6 +4,7 @@
 
 namespace nix {
 
+void JSONParseError::anchor() {}
 
 static void skipWhitespace(const char * & s)
 {
--- a/src/libexpr/nixexpr.cc
+++ b/src/libexpr/nixexpr.cc
@@ -7,6 +7,16 @@
 
 namespace nix {
 
+// anchor vtables
+void EvalError::anchor() {}
+void ParseError::anchor() {}
+void IncompleteParseError::anchor() {}
+void AssertionError::anchor() {}
+void ThrownError::anchor() {}
+void Abort::anchor() {}
+void TypeError::anchor() {}
+void UndefinedVarError::anchor() {}
+void RestrictedPathError::anchor() {}
 
 /* Displaying abstract syntax trees. */
 
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1473,6 +1473,7 @@ void replaceValidPath(const Path & storePath, const Path tmpPath)
 
 
 MakeError(NotDeterministic, BuildError)
+void NotDeterministic::anchor() {}
 
 
 void DerivationGoal::buildDone()
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -29,6 +29,8 @@ using namespace std::string_literals;
 
 namespace nix {
 
+void DownloadError::anchor() {}
+
 double getTime()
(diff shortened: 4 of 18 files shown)


shlevy:
I think we definitely need some in-code documentation on what these are doing and what problem they solve. Maybe a detailed description in `maintainers/` and a pointer there above each declaration?

shlevy (changes requested):
See previous comment.

edolstra:
Hm, I wonder why this hasn't caused problems before. I remember that we had a bug at work with exceptions / RTTI across DSOs, but I can't remember if that was Linux or macOS...

dtzWill:
#1678 anchored `nix::Exit`, FWIW.  But that was due to issues I was seeing as well :).

dtzWill:
clang's `-Wweak-vtable` helps here, FWIW.

Avoiding weak vtables makes things work more readily, situations that might not work when relying on them re:exceptions/dynamic_cast:
* dlopen with RTLD_LOCAL
* Using `-fvisibility=hidden`
* Using gcc before 4.5? Or so.
* Some situations with libc++ (since it doesn't do the string name comparison of type_info that makes duplicate type_info's "work")
* Various linker flags re:exporting dynamic symbols

Anchoring vtables (no weak vtables) is part of LLVM's coding standard, FWIW, although AFAIK not for these reasons as LLVM avoids both RTTI and exceptions:

https://llvm.org/docs/CodingStandards.html#provide-a-virtual-method-anchor-for-classes-in-headers 

So basically anchoring them helps code size a bit and helps avoid the need to ensure things work right, instead of relying on type_info merging we just have one and done.

I'll try to come up with how to best reference/document this in the code.

shlevy:
Plugins rely on RTLD_LOCAL, just FYI

dtzWill:
I'm not sure how well this sort of detail is "documented", but here's a post that seems to describe it well enough: https://marcmutz.wordpress.com/2010/08/04/fun-with-exceptions/ .

That said much of this is implementation-specific with various compilers/toolchains having different behavior and workarounds-- even the approach here is only required and indeed AFAIK only effective due to implementation choices on various platforms and compilers.  AFAIK the C++ standard simply requires that whatever the compiler/linker do that there only be one definition (the so-called "one definition rule").

I like the explanation here, hopefully it helps this make sense: https://stackoverflow.com/a/23749273

As for plugins, I'm not sure that RTLD_LOCAL + exceptions will work? Will get back to you.

dtzWill:
Looks like doesn't work with libc++, which makes sense since IIRC it has stricter behavior re:comparing addresses of type_info's.  Here's a quick repo I put together for demonstrating/testing this behavior; https://github.com/dtzWill/throw_from_share .

dtzWill:
Updated to demonstrate with and without anchor!

shlevy:
@dtzWill That link doesn't work for me. What precisely doesn't work with libc++? We can't throw from RTLD_LOCAL DSOs at all?

shlevy:
@dtzWill Can you give a bit more information on how this impacts plugins, if at all?

xokdvium:
This has been done in https://github.com/NixOS/nix/pull/15770 and the follow-up from that which enables Werror=weak-vtables. At least less than 10 years since this PR 🫠

PR #1942, opened 2018-03-03
installables: Support unprefixed NIX_PATH lookups.

### LINKED ISSUE #1865: attribute 'nixpkgs' in selection path 'nixpkgs.hello' not found
Try running the following with Nix 2.0:

```
$ NIX_PATH=/nix/var/nix/profiles/per-user/root/channels nix run nixpkgs.hello
error: attribute 'nixpkgs' in selection path 'nixpkgs.hello' not found
```

This is different than how previous version of Nix worked.

Oddly settings 'nixpkgs' to anything in NIX_PATH (even something that doesn't exist) will fix this issue:

```
NIX_PATH='nixpkgs=asdf:/nix/var/nix/profiles/per-user/root/channels' nix run nixpkgs.hello
```

shlevy:
Fixes #1865

### DIFF
--- a/src/nix/installables.cc
+++ b/src/nix/installables.cc
@@ -39,29 +39,36 @@ Value * SourceExprCommand::getSourceExpr(EvalState & state)
 
         auto searchPath = state.getSearchPath();
 
-        state.mkAttrs(*vSourceExpr, searchPath.size() + 1);
+        std::unordered_set<std::string> pathAttrNames;
+        pathAttrNames.reserve(searchPath.size());
+        for (const auto & i : searchPath) {
+            if (i.first == "") {
+                DirEntries ents;
+                try {
+                    ents = readDirectory(i.second);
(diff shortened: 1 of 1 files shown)


shlevy:
@edolstra Is this OK?

edolstra:
No, because adding every file in every directory listed in `NIX_PATH` sounds like it could seriously pollute the set of expressions evaluated by `nix`, potentially adding expressions that are not intended to be package sources.

For example, with the `NIX_PATH` on NixOS, this adds every package in Nixpkgs as a `nixos.*` set, doubling evaluation time for `nix search` and creating ambiguity.  It also adds files like `nixos-config` which is also not exactly elegant.

In short `NIX_PATH` is not intended for this. The current use of `nixpkgs=` is a placeholder to be replaced when we have a proper `nix-channel` replacement.

PR #1943, opened 2018-03-03
release.nix: don't try to use nix-2.0 branch, no longer exists

dtzWill:
Probably should point at the 18.03 release branch once that's made.

Currently evaluating this throws an error.


vcunat:
Oh, I had no idea the branch is referenced anywhere, so I simply deleted it on merging to master.

PR #1944, opened 2018-03-04
Don't throw in destructor, happens in nix-daemon when client disappears.

dtzWill:
Wasn't sure how to best handle the ignoreException decl re:layering.


dtzWill:
Moved into logging.cc instead, minimal org impact :).

PR #1946, opened 2018-03-05
ssh-ng: Only forward settings that make sense to forward.

### LINKED ISSUE #1935: Remote builds via ssh-ng try to use local sandbox shell
error: build of '/nix/store/4306ksk71wpvrvw23pfhsplfnrzhsfri-binutils-2.30.drv' on 'ssh-ng://nix-ssh@<redacted>?ssh-key=<redacted>' failed: while setting up the build environment: getting attributes of path '/nix/store/mqy29jzrp7dqcivwb919ksg4g75m5lfm-busybox-1.28.0/bin/busybox': No such file or directory

shlevy:
Fixes #1713 and fixes #1935

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -95,11 +95,11 @@ public:
         "Whether to keep temporary directories of failed builds."};
 
     Setting<bool> keepGoing{this, false, "keep-going",
-        "Whether to keep building derivations when another build fails."};
+        "Whether to keep building derivations when another build fails.", {}, true};
 
     Setting<bool> tryFallback{this, false, "fallback",
         "Whether to fall back to building when substitution fails.",
-        {"build-fallback"}};
+        {"build-fallback"}, true};
 
     /* Whether to show build log output in real time. */
     bool verboseBuild = true;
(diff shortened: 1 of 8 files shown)


puffnfresh:
Just pointing out for anybody else; this is not specific to ssh-ng, which is really good for things I've been playing around with.

edolstra:
IMHO this is not the right approach since it will lead to endless confusion and bikeshedding over what settings need to be forwarded. For example, should `--cores` affect the remote system? Generally not, except when I want to disable parallelism. So it might be better to allow passing arbitrary settings in the store URI (e.g. `--store ssh-ng://foo?cores=1`).

Regarding build-related flags, polluting `config.hh` with an ad-hoc flag for whether to forward certain settings seems ugly to me. In fact it's not clear why we can't just use the existing code for forwarding build-related settings, i.e.
```
    conn.to << wopSetOptions
       << settings.keepFailed
       << settings.keepGoing
    ...
```
in `RemoteStore::setOptions()`.

shlevy:
@edolstra We can't reuse the existing code because it forwards things like `keepFailed` and `buildCores` unconditionally. Closing in favor of #1948, which gives us a minimal sane base on top of which we can iterate if need be.

PR #1965, opened 2018-03-12
Wrap thread local in function for Cygwin

### LINKED ISSUE #1826: Nix fails to build on cygwin64 (thread_local)
Building nix master 0167eac571d6d92f18640d05cfb7fa10a4cd0fd9 on cygwin64 fails, with multiple lines like this:

```
src/libstore/build.o:/home/xxxx/3rd/nix/src/libstore/build.cc:4326:(.text+0x                    1553): relocation truncated to fit: R_X86_64_PC32 against undefined symbol `TLS                     init function for nix::curActivity'
```

That seems related to the only thread_local c++11 symbol, and this seems to be a gcc issue since 4.8: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64697

The last nix on cygwin64 that we use fine is 1.11.3. I remember that 1.11.7 fails because the libexec/ executables cannot find some dlls.

cc @cpages

masaeedu:
Fixes #1826. See #1352 for a previous instance of a similar change.

### DIFF
--- a/src/libstore/download.hh
+++ b/src/libstore/download.hh
@@ -22,7 +22,7 @@ struct DownloadRequest
     std::string mimeType;
 
     DownloadRequest(const std::string & uri)
-        : uri(uri), parentAct(curActivity) { }
+        : uri(uri), parentAct(getCurActivity()) { }
 };
 
 struct DownloadResult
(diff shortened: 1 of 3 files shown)


shlevy:
Seems fine to me. @viric can you confirm this fixes it?

dtzWill:
LGTM.

PR #1983, opened 2018-03-16
nix-shell: allow symlinks to .drvs

lheckemann:
This makes persistent shell environments easier to use.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -271,10 +271,14 @@ void mainWrapped(int argc, char * * argv)
         exprs = {state.parseStdin()};
     else
         for (auto i : left) {
+            auto absolute = i;
+            try {
+                absolute = canonPath(absPath(i), true);
+            } catch (Error e) {};
             if (fromArgs)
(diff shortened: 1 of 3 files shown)


src/nix-build/nix-build.cc
```diff
@@ -271,10 +271,14 @@ void mainWrapped(int argc, char * * argv)
         exprs = {state.parseStdin()};
     else
         for (auto i : left) {
+            auto absolute = i;
+            try {
+                absolute = canonPath(absPath(i), true);
```
shlevy: Do we care that this will also resolve symlinks in the store?

lheckemann: I don't think so, since this only affects the .drv case.

7c6f434c: And what happens if there are symlinks in the store path (i.e. `/nix/` is a symlink)? Does it move from «purity not guaranteed» to «doesn't work»?

shlevy: /nix is not allowed to be a symlink anyway

PR #1998, opened 2018-03-20
execl: cast NULL sentinel to (char *), per man page and compiler warning

dtzWill:
From exec(3):

> The list of arguments must be terminated by a null pointer, and, since these
> are variadic functions, this pointer must be cast (char *) NULL

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -263,7 +263,7 @@ void showManPage(const string & name)
 {
     restoreSignals();
     setenv("MANPATH", settings.nixManDir.c_str(), 1);
-    execlp("man", "man", name.c_str(), NULL);
+    execlp("man", "man", name.c_str(), (char *)NULL);
     throw SysError(format("command 'man %1%' failed") % name.c_str());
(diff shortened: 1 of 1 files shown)


shlevy:
Eh. Can we use `nullptr` instead?

dtzWill:
I'd imagine so, but not sure! Since I wasn't sure I went with what was explicitly mentioned :).

shlevy:
Does it fix the warning?

shlevy:
(`nullptr` without a cast, I mean)

dtzWill:
Honestly I don't remember what env was producing the error, but I **think** `nullptr` is legit.  I'll update with that and see if I can't quickly pull up whatever compiler it was.

7c6f434c:
Does Clang support it? I think I have seen a `poppler` build on Darwin fail because of `nullptr` today. (I don't use anything Darwin myself)

shlevy:
@7c6f434c yes, clang definitely supports `nullptr` when compiling C++.

PR #2010, opened 2018-03-26
libexpr: Make unsafeGetAttrPos not crash on noPos

dezgeg:
Currently e.g. `builtins.unsafeGetAttrPos "path" builtins` will
eventually segfault because pos->file is an unset Symbol.

Found by afl-fuzz.


dezgeg:
Not sure what's the best fix here, just picked one.

shlevy:
I have a mild preference for `null` instead of a special string

dezgeg:
I did think of `null`, but I have the prediction that whoever calls this will immediately do something like:
````
let pos = builtins.unsafeGetAttrPos foo bar; in "${pos.file}:${pos.line}:${pos.column}"
````
which will then blow up on `null`.

shlevy:
It should IMO :grinning:

shlevy:
Less facetiously: What is the "right" behavior for that code in the noPos case?

dezgeg:
Well I don't think `"(unknown file):0:0"` is too bad of an output, but I guess it could be beautified to handle `noPos` like this:
````
let
    pos = builtins.unsafeGetAttrPos foo bar;
in if pos.line == 0
     then "(undefined position)"
     else "${pos.file}:${toString pos.line}:${toString pos.column}"
````

shlevy:
If you're doing that check then why not just check for null (or attr non-existence?)

dezgeg:
Oh I had somehow misinterpreted that question as returning `{ column = 0; file = null; line = 0; }` instead of returning `null`.

IMHO the problem of `null` there is that now `null` can mean both "the attribute didn't exist" and "position of that attribute is not known" while it previously could only be the former. So before this change, one could reasonably write code like:
````
{ name, version, sha256 } @ args:
let
    pos = builtins.unsafeGetAttrPos args "name";
in mkDerivation (args // ...)
````
and with the expectation that `pos` is always non-`null`.

edolstra:
`unsafeGetAttrPos` should probably just throw an exception if the position is undefined.

dezgeg:
Throwing an exception sounds pretty inconsistent given that the function already returns `null` in one error case.

I guess returning `null` in the crashing case is mostly fine given that nothing in nixpkgs would break on a sudden `null`. I've updated the commit.

Issue #2001, opened 2018-03-21
Store is corrupt (missing hashes) in clean install of Nix 2.0


bchallenor:
## Steps to reproduce

Build this Dockerfile:

    FROM debian:stable

    RUN apt-get update && apt-get install -y curl bzip2

    RUN curl --remote-name https://nixos.org/releases/nix/nix-2.0/nix-2.0-x86_64-linux.tar.bz2
    RUN echo "6312837aee33306cdbb351b75ba1638b89d21b30f0caf0346f9a742425f197ee  nix-2.0-x86_64-linux.tar.bz2" | sha256sum -c
    RUN tar xjf nix-*-x86_64-linux.tar.bz2

    RUN useradd --create-home --shell /bin/bash --password "*" nix

    RUN mkdir -m 0755 /nix
    RUN chown nix:nix /nix /mnt

    USER nix
    ENV USER=nix

    RUN nix-*-x86_64-linux/install

    ENV PATH=/home/nix/.nix-profile/bin:/usr/sbin:/usr/bin:/sbin:/bin

    RUN nix-store --verify --check-contents

## Expected behaviour

`nix-store --verify --check-contents` should return simply:

    reading the Nix store...
    checking path existence...
    checking hashes...

## Actual behaviour

`nix-store --verify --check-contents` returns:

    reading the Nix store...
    checking path existence...
    checking hashes...
    fixing missing hash on '/nix/store/108i1xswxgmvrzrsrx277892gpw6ix53-zlib-1.2.11'
    updating size field on '/nix/store/108i1xswxgmvrzrsrx277892gpw6ix53-zlib-1.2.11' to 104328
    fixing missing hash on '/nix/store/1b5l524r05qzgws8k2v7f0p52i0y4nq6-openssl-1.0.2n'
    updating size field on '/nix/store/1b5l524r05qzgws8k2v7f0p52i0y4nq6-openssl-1.0.2n' to 3443112
    fixing missing hash on '/nix/store/1jn750iav5rfbm05345ysi8dvxcz7j26-gnutar-1.30'
    updating size field on '/nix/store/1jn750iav5rfbm05345ysi8dvxcz7j26-gnutar-1.30' to 2623000
    fixing missing hash on '/nix/store/22hw10naf4flf4czcay460fky7kmh89v-libkrb5-1.15.2'
    updating size field on '/nix/store/22hw10naf4flf4czcay460fky7kmh89v-libkrb5-1.15.2' to 2289448
[...]

PR #2011, opened 2018-03-26
release.nix: Use pkgs.closureInfo for binaryTarball

dezgeg:
pathsFromGraph is legacy since Nix 2.0.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -115,16 +115,15 @@ let
       let
         toplevel = builtins.getAttr system jobs.build;
         version = toplevel.src.version;
+        installerClosureInfo = closureInfo { rootPaths = [ toplevel cacert ]; };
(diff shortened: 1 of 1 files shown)


dezgeg:
Note that so far this is essentially untested by me - I only checked that the `.reginfo` file gets the necessary stuff required to fix https://github.com/NixOS/nix/issues/2001.

grahamc:
I'd really like to avoid using Nix 2.0 features for a while longer, it is already causing a very painful experience for NixOS Stable users.

dezgeg:
It's already using `builtins.fetchGit`. But actually yes, that one can be worked around by building inside `nix-shell -p nixUnstable` while `closureInfo` can't.

PR #2023, opened 2018-03-28
Keep LC_ALL

matthewbauer:
LC_ALL is often needed by programs to detect what locale to output in. This can be safely passed to pure nix-build.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -368,7 +368,7 @@ void mainWrapped(int argc, char * * argv)
         auto tmp = getEnv("TMPDIR", getEnv("XDG_RUNTIME_DIR", "/tmp"));
 
         if (pure) {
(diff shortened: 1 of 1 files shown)


edolstra:
I don't think that this is a good idea in pure mode. The whole point of pure mode is *not* to be infected by those kinds of environment variables. A quick grep of Nixpkgs shows dozens of packages that set LC_ALL, so there is plenty of evidence of LC_ALL having a material effect on build results.

7c6f434c:
From that point of view, HOME, USER and DISPLAY also sound like something that will affect the build.

shlevy:
In my experience the only effect LC_ALL seems to have is whether it's set or not (or sometimes whether it's set to a utf-8 locale or not)... Are there counterexamples?

globin:
There are quite a few python packages that depend on a UTF-8 locale for running tests, most of those are set because of that.

shlevy:
Is there a strong reason not to just set a UTF-8 locale in stdenv? Then it seems like respecting the user's LC_ALL would be much less of a divergence.

matthewbauer:
I wonder if it really is a python issue then... It should be able to default to the "C" locale - but for whatever reason that never happens. Anyway, I'm convinced this probably isn't the place to fix this issue.

PR #2025, opened 2018-03-29
nix copy: add an example with S3

zimbatm:
I couldn't find a good example how to use it with non-us-east-1 buckets.

I think these options should be documented more thoroughly but I didn't find the right place to do it. If you point me to the right place I'm happy to write some more docs.


src/nix/copy.cc
```diff
@@ -67,6 +67,10 @@ struct CmdCopy : StorePathsCommand
                 "To copy a closure from another machine via SSH:",
                 "nix copy --from ssh://server /nix/store/a6cnl93nk1wxnq84brbbwr6hxw9gp2w9-blender-2.79-rc2"
             },
+            Example{
```
edolstra: There should probably be an `#ifdef ENABLE_S3` around this.

zimbatm: Fixed!

PR #2026, opened 2018-03-29
Expand the multi-user installer to support Linuxes with systemd

grahamc:
- darwin installer: delete hardware report, not necessary
 - moves os-specific code from the darwin installer to to `poly_*`
   functions
 - adds profile.d support to the profile targets, which automatically
   handles many distros which don't have a /etc/bashrc but do have an
   /etc/profile.d
 - /bin/bash -> /usr/bin/env bash
 - document why each excluded shellcheck check is excluded
 - rename the multi-user to Daemon-based

Tested on:

 - Arch: Works
 - CentOS 7: Works
 - Debian Jessie: Works
 - Debian Stretch: Works
 - Debian Wheezy: falls back to the single user installer (not systemd)
 - Ubuntu Artful: Works
 - Ubuntu Wily: Works
 - Ubuntu Xenial: Works
 - Ubuntu Trusty: falls back to the single user installer (not systemd)
 - MacOS Sierra: Still works :)

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -127,17 +127,39 @@ let
           substitute ${./scripts/install-nix-from-closure.sh} $TMPDIR/install \
             --subst-var-by nix ${toplevel} \
             --subst-var-by cacert ${cacert}
-          substitute ${./scripts/install-darwin-multi-user.sh} $TMPDIR/install-darwin-multi-user \
+
+          substitute ${./scripts/install-darwin-multi-user.sh} $TMPDIR/install-darwin-multi-user.sh \
+            --subst-var-by nix ${toplevel} \
+            --subst-var-by cacert ${cacert}
+          substitute ${./scripts/install-systemd-multi-user.sh} $TMPDIR/install-systemd-multi-user.sh \
+            --subst-var-by nix ${toplevel} \
+            --subst-var-by cacert ${cacert}
+          substitute ${./scripts/install-multi-user.sh} $TMPDIR/install-multi-user \
             --subst-var-by nix ${toplevel} \
             --subst-var-by cacert ${cacert}
 
           if type -p shellcheck; then
-            shellcheck -e SC1090 $TMPDIR/install
-            shellcheck -e SC1091,SC2002 $TMPDIR/install-darwin-multi-user
+            # SC1090: Don't worry about not being able to find
+            #         $nix/etc/profile.d/nix.sh
+            shellcheck --exclude SC1090 $TMPDIR/install
+            shellcheck $TMPDIR/install-darwin-multi-user.sh
+            shellcheck $TMPDIR/install-systemd-multi-user.sh
+
+            # SC1091: Don't panic about not being able to source
+            #         /etc/profile
+            # SC2002: Ignore "useless cat" "error", when loading
+            #         .reginfo, as the cat is a much cleaner
+            #         implementation, even though it is "useless"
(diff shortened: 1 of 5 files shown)


grahamc:
The diff is a bit tedious to read, but essentially the install-multi-user script is calls `poly_*` functions which implement platform-specific behavior, like creating a user, fetching a user's attribute, or setting a user's attribute.

shlevy:
Reviewed by downloading install-multi-user and diffing with a local copy of install-darwin-multi-user

grahamc:
@shlevy would you mind posting that diff?

edolstra:
Can we do something about the extreme verbosity of the installer? Why is it necessary to tell the user about every sudo command separately? What useful information do messages like 'in order to give $1 a safe home directory' convey? ("Rule of Silence: When a program has nothing surprising to say, it should say nothing.")

In fact, why don't we just re-exec most of the installer under sudo, rather than invoke sudo a gazillion times?

grahamc:
I've heard from a number of people that they really like the verbosity, because they feel like they can look back and know exactly what happened, should they want to undo it. I've even been contracted to make other installers for other tools be similarly chatty... I have also heard the opposite, of course. Maybe there is a happy medium?

Maybe we could cut down the # of lines per sudo? or a `--quiet`?

grahamc:
After some non-blocking cleanup patches, it is confirmed working on (this is a todo list):

 - [x] macOS High Sierra
 - [ ] macOS Sierra
 - [ ] Ubuntu 14
 - [ ] Arch
 - [ ] CentOS 7
 - [ ] Debian Jessie
 - [ ] Debian Stretch
 - [ ] Debian Wheezy (falls back to single user)
 - [ ] Ubuntu Artful
 - [ ] Ubuntu Wily
 - [ ] Ubuntu Xenial
 - [ ] Ubuntu Trusty (falls back to single user)

grahamc:
This is ready to go as far as my testing reveals, and I've opened #2027 to backport it to the maintenance branch.

dezgeg:
To me doing this much stuff in a curl-piped-to-bash style script sounds scary when it comes to uninstallation, upgrads and such. E.g. currently the uninstall process is a dead simple `rm -rf /nix` but after this you need to manually remove the added users, etc...

I wonder if anybody has looked into https://github.com/jordansissel/fpm for building multi-user installer packages for Deb/RPM/Pacman?

dtzWill:
> To me doing this much stuff in a curl-piped-to-bash style script sounds scary when it comes to uninstallation, upgrads and such. E.g. currently the uninstall process is a dead simple rm -rf /nix but after this you need to manually remove the added users, etc...

Good call: this does make it very unclear how to uninstall!  I don't think the answer is an equally magic "uninstaller" (there be dragons) so exploring how to leverage package manager on each system would be very valuable.

(the package manager solutions might help provide an upgrade path as well?)

------


".... who would uninstall Nix?!" :grin:

Continuing such thoughts.... "We've upgraded you to NixOS as well, you're welcome" :wink:
(insert automatic kexec magic here)

Mic92:
@dezgeg https://github.com/NixOS/nix/pull/1141 <- @abbradar invested some effort in improving rpm/deb packages.

Mic92:
@dezgeg I think it would make sense to also provide pre-build packages. Otherwise there is a bootstrap issue to build packages without having nix.

PR #2027, opened 2018-03-30
Backport #2026: Expand the multi-user installer to support Linuxes with systemd

grahamc:
- darwin installer: delete hardware report, not necessary
 - moves os-specific code from the darwin installer to to `poly_*`
   functions
 - adds profile.d support to the profile targets, which automatically
   handles many distros which don't have a /etc/bashrc but do have an
   /etc/profile.d
 - /bin/bash -> /usr/bin/env bash
 - document why each excluded shellcheck check is excluded
 - rename the multi-user to Daemon-based

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -145,17 +145,39 @@ let
           substitute ${./scripts/install-nix-from-closure.sh} $TMPDIR/install \
             --subst-var-by nix ${toplevel} \
             --subst-var-by cacert ${cacert}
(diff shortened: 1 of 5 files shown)


edolstra:
I'm not sure that this is a good idea, since IMHO a patch release (2.0.1) should not radically change the behaviour of the installer (by switching to doing a multi-user install on Linux).

PR #2028, opened 2018-03-30
Add test eval-okay-nested-with

ryantrinkle:
This test ensures that `with`s nest in static order (even though it is a separate static order from the one used by lexically-scoped names)


ryantrinkle:
Thanks!

PR #2036, opened 2018-04-02
Make the TTL for disk cache configurable

AmineChikhaoui:
This should allow modifying the disk cache TTL or completely disabling it to force another binary cache query.

### DIFF
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -788,6 +788,30 @@ password <replaceable>my-password</replaceable>
 
   </varlistentry>
 
+  <varlistentry xml:id="conf-negative-disk-cache-ttl"><term><literal>negative-disk-cache-ttl</literal></term>
+
+    <listitem>
+
+      <para>The TTL in seconds for negative lookups. If a store path is queried from a substituer but
+      was not found, there will be a negative lookup cached in the local disk cache database for the specified
+      duration.</para>
+
+    </listitem>
+
+  </varlistentry>
+
+  <varlistentry xml:id="conf-positive-disk-cache-ttl"><term><literal>positive-disk-cache-ttl</literal></term>
+
+    <listitem>
+
+      <para>The TTL in seconds for positive lookups. If a store path is queried from a substituer, the result of
+      the query will be cached in the local disk cache database including some of the NAR metadata. Setting a TTL
+      for positive lookups can be useful in case of builds that aren't reproducible, in which case having a more
+      frequent cache invalidation would prevent hash mismatch issues.</para>
+
+    </listitem>
+
+  </varlistentry>
 
 </variablelist>
 
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -456,7 +456,7 @@ builtins.fetchurl {
 
   </varlistentry>
 
-  <varlistentry xml:id="conf-negative-disk-cache-ttl"><term><literal>negative-disk-cache-ttl</literal></term>
+  <varlistentry xml:id="conf-narinfo-cache-negative-ttl"><term><literal>narinfo-cache-negative-ttl</literal></term>
 
     <listitem>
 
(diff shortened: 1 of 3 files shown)


AmineChikhaoui:
@shlevy yeah sure

AmineChikhaoui:
@shlevy done let me know if anything should be modified/added.

volth:
I do not like timeout values, it adds some non-determinism.
That means, today we have "hash mismatch" error but one day the problem will magically disappear because the timeout expired.

What makes the timeout 30 days, not 30 hours or 30 minutes or 30 seconds?

AmineChikhaoui:
Well setting `--option positive-disk-cache-ttl 0` in all deployment scripts is probably not a good idea as you'll be querying every single path in the closure (also much more load on cloudfront side) but it's more of a convenience config for when someone needs to just disable the disk cache for pulling one store path or at least set a smaller TTL value without removing the whole binary-cache*.sqlite db to get around an issue. It also doesn't make much sense to me in general to have this config hard-coded.

volth:
> as you'll be querying every single path in the closure 

Not actually.
More correct "every single path in the closure which does not exist on the target". After the ```nix copy``` (or another operation which queried the cache) the path will likely land on the target and won't queried from the cache again (at least until it removed by GC)

volth:
It would be interesting to log the hit/miss usage of ```NarInfoDiskCache``` to check is it useful or it is only a source of problem.

doc/manual/command-ref/conf-file.xml
```diff
@@ -788,6 +788,30 @@ password <replaceable>my-password</replaceable>
 
   </varlistentry>
 
+  <varlistentry xml:id="conf-negative-disk-cache-ttl"><term><literal>negative-disk-cache-ttl</literal></term>
+
+    <listitem>
+
+      <para>The TTL in seconds for negative lookups. If a store path is queried from a substituer but
```
edolstra: => `substituter`.

doc/manual/command-ref/conf-file.xml
```diff
@@ -788,6 +788,30 @@ password <replaceable>my-password</replaceable>
...
+
+    </listitem>
+
+  </varlistentry>
+
+  <varlistentry xml:id="conf-positive-disk-cache-ttl"><term><literal>positive-disk-cache-ttl</literal></term>
+
+    <listitem>
+
+      <para>The TTL in seconds for positive lookups. If a store path is queried from a substituer, the result of
```
edolstra: => `substituter`.

doc/manual/command-ref/conf-file.xml
```diff
@@ -788,6 +788,30 @@ password <replaceable>my-password</replaceable>
...
+      the query will be cached in the local disk cache database including some of the NAR metadata. Setting a TTL
+      for positive lookups can be useful in case of builds that aren't reproducible, in which case having a more
+      frequent cache invalidation would prevent hash mismatch issues.</para>
```
edolstra: `Setting a TTL..." implies that there is no TTL by default, which is not the case. Would be better to say something like "The defautl TTL is bla bla. Setting a shorter TTL can be useful in case...".

BTW this scenario may not be completely clear. I assume this is when the binary cache has frequent garbage collections (in which case it doesn't really matter whether the builds are reproducible).

doc/manual/command-ref/conf-file.xml
AmineChikhaoui: oops.

doc/manual/command-ref/conf-file.xml
AmineChikhaoui: @edolstra yeah that's also another case where this is useful, I'll mention that instead.

src/libstore/globals.hh
```diff
@@ -313,6 +313,14 @@ public:
     Setting<Strings> trustedUsers{this, {"root"}, "trusted-users",
         "Which users or groups are trusted to ask the daemon to do unsafe things."};
 
+    Setting<unsigned int> ttlNegativeDiskCache{this, 3600, "negative-disk-cache-ttl",
```
edolstra: One more thing: "disk cache" is a bit ambiguous as there can be many disk cache (e.g. `nix search` also has a disk cache). Maybe it should be called `narinfo cache` (e.g. `narinfo-cache-negative-ttl`).

AmineChikhaoui: sounds good

AmineChikhaoui:
@edolstra options name updated.

Issue #1221, opened 2017-02-07
Remove the build hook mechanism


edolstra:
Like the substituter mechanism, we can use the Store abstraction instead. I.e. we call `remoteStore->buildPaths(...)` directly rather than calling an external process that does the same thing.

expipiplus1:
Would this just be an internal change to Nix? If not, how would this affect those of us who are using `NIX_BUILD_HOOK` with things other than `build-remote.pl`?

edolstra:
It would break your build hook. What are you using `NIX_BUILD_HOOK` for?

expipiplus1:
I'm using `NIX_BUILD_HOOK` with `nix-instantiate` and `nix-store --realise` in order to easily get machine readable output of the build in progress. By using a build hook with these tools I can have precise control over what's being built when, and the output of the builder.

It's also been useful in getting some better `--dry-run` functionality as described [here](https://github.com/NixOS/nix/issues/666#issuecomment-277542784) too.

veprbl:
@expipiplus1 
Which entity executes ```NIX_BUILD_HOOK``` in your scenario? I'm asking because both ```nix-build``` and  ```nix-store -r``` will try to lock the output path, so you can't have one just call another.

ledettwy:
Would doing this prevent implementation of use cases like https://github.com/NixOS/nix/issues/1914?

edolstra:
@ledettwy No, it shouldn't affect that.

volth:
This also un-fixes https://github.com/NixOS/nix/issues/1118
Since ```--no-build-hook``` is removed there is no way to force local build

edolstra:
@volth You can force a local build by passing `--builders ''`.

volth:
Yes, but it requres to be a trusted user.
I understand that adding builders requires privilegies, but removing?

edolstra:
You're right, it requires a trusted user. We could require the untrusted user's `builders` to be a subset of the daemons `builders` though.

srghma:
I've come from here https://github.com/NixOS/nixpkgs/issues/28816 and I'm looking for a way to pass impure env variable to `nix-build`

Building locally should help, but I can't make the example work, can someone help me

```
code=`<<'EOF'
with import <nixpkgs> {};
stdenv.mkDerivation {
  name="impure";
  phases=["buildPhase"];
  impureEnvVars=["IMPURE"];
  buildPhase=''
    env | grep IMPURE
[...]

PR #2037, opened 2018-04-02
List the features a derivation requires when it cannot be built locally

grahamc:
> error: a 'aarch64-linux' is required to build '/nix/store/2r9r8m2pf8j0kjwfxk02gx5siyl4m9ks-hello-2.10.drv', but I am a 'x86_64-linux'

okay .... but ...

```
$ cat /etc/nix/machines
root@myaarch64 aarch64-linux /root/foobar 96 1 big-parallel
```

_applies patch_

> error: a 'aarch64-linux' with the system feature 'magic-pixie-dust' is required to build '/nix/store/2r9r8m2pf8j0kjwfxk02gx5siyl4m9ks-hello-2.10.drv', but I am a 'x86_64-linux'

ah ha!

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1786,9 +1786,24 @@ void DerivationGoal::startBuilder()
 {
     /* Right platform? */
     if (!drv->canBuildLocally()) {
-        throw Error(
-            format("a '%1%' is required to build '%3%', but I am a '%2%'")
-            % drv->platform % settings.thisSystem % drvPath);
+        Strings features = tokenizeString<Strings>(get(drv->env, "requiredSystemFeatures"));
+
+        if (features.size() == 1) {
+            throw Error(
+                "a '%s' system with the feature '%s' is required to build '%s'",
+                drv->platform, concatStringsSep("', '", features), drvPath
+                );
+        } else if (features.size() > 1) {
+            throw Error(
+                "a '%s' system with the features '%s' is required to build '%s'",
+                drv->platform, concatStringsSep("', '", features), drvPath
+                );
+        } else {
+            throw Error(
+                "a '%s' system is required to build '%s'",
+                drv->platform, drvPath
+                );
+        }
     }
 
     if (drv->isBuiltin())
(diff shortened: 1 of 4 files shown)


grahamc:
P.S. this is literally the first C++ I've ever written, so ...

shlevy:
Hmm... So one thought is that the requred system features stuff is technically part of the build-remote build hook, and as of today build.cc is technically build-hook agnostic. There's #1221 which may be the right way to go here, or maybethis should instead be part of `build-remote` when it fails to find a matching machine?

grahamc:
I think you're right, it is a bit outside of the responsibilities here. Perhaps a more C++-skilled dev would like to take over the idea and carry it forward :)

shlevy:
@edolstra Thoughts on this? Should we remove the concept of build-hooks separate from build-remote?

src/libstore/build.cc
```diff
@@ -1783,9 +1783,21 @@ void DerivationGoal::startBuilder()
...
     if (!drv->canBuildLocally()) {
-        throw Error(
-            format("a '%1%' is required to build '%3%', but I am a '%2%'")
-            % drv->platform % settings.thisSystem % drvPath);
+        Strings features = tokenizeString<Strings>(get(drv->env, "requiredSystemFeatures"));
+
+        if (features.size() > 1) {
+            throw Error(
+                format("a '%1%' with the system features '%4%' is required to build '%3%', but I am a '%2%'")
+                % drv->platform % settings.thisSystem % drvPath % concatStringsSep("', '", features));
```
edolstra: Note that `format` is obsolete, you can just write:
```
throw Error("a machine of type '%s' is required", drv->platform);
```

grahamc:
@edolstra I just updated to use the new way, and also added tests to assert the messages come out properly.

grahamc:
If this is acceptable and merges, ideally, we would also backport it.

grahamc:
Shea proposes I move the impl to https://github.com/NixOS/nix/blob/1df32c7d7c5ce0a4d10fa1e13f557ff525a62efa/src/build-remote/build-remote.cc#L117

domenkozar:
This was actually committed in https://github.com/NixOS/nix/commit/1e7b8deea7e052ed9ebf47d1411bcaf542054b41#diff-267a6391980cbd9a743958945ff44ef9R1796

PR #2049, opened 2018-04-06
Add missing eval-okay-regex-split.exp test file

mightybyte:
I noticed that this file was absent for some reason.


shlevy:
I think the way the tests work is that if there's no `.exp` it just checks that the evaluation succeeds?

mightybyte:
Well, this test case has one: https://github.com/NixOS/nix/blob/master/tests/lang/eval-okay-regex-match.exp

shlevy:
Is this causing any issues? Test suite seems to run fine here.

mightybyte:
I'm working on tools that import and run these tests automatically, so it would be nice to have uniformity.

PR #2056, opened 2018-04-08
nix-shell: Fixes use with ruby shebangs.

samueldr:
The ported code in 80ebc553eca19dafc64c47420cd49ddd506bc9b7 was incorrectly ported.

```
-            $envCommand = "exec $execArgs $interpreter -e 'load(\"$script\")' -- ${\(join ' ', (map shellEscape, @savedArgs))}";
...
+                    envCommand = (format("exec %1% %2% -e 'load(\"%3%\") -- %4%") % execArgs % interpreter % script % joined.str()).str();
```

The single-quote finishing the small ruby snippet was lost in
translation.

* * *

Symptoms:

```
/tmp/nix-shell-1953-0/rc: line 1: unexpected EOF while looking for matching `''
/tmp/nix-shell-1953-0/rc: line 2: syntax error: unexpected end of file
```

* * *

~~In addition, I will be looking at how tests work, but I would really like to add a test for this regression. Any help is welcome.~~

### DIFF
--- a/tests/nix-shell.sh
+++ b/tests/nix-shell.sh
@@ -39,3 +39,12 @@ chmod a+rx $TEST_ROOT/shell.shebang.sh
 
 output=$($TEST_ROOT/shell.shebang.sh abc def)
 [ "$output" = "foo bar abc def" ]
+
+# Test nix-shell shebang mode for ruby
+# This uses a fake interpreter that returns the arguments passed
+# This, in turn, verifies the `rc` script is valid and the `load()` script (given using `-e`) is as expected.
+sed -e "s|@ENV_PROG@|$(type -p env)|" shell.shebang.rb > $TEST_ROOT/shell.shebang.rb
+chmod a+rx $TEST_ROOT/shell.shebang.rb
+
+output=$($TEST_ROOT/shell.shebang.rb abc ruby)
+[ "$output" = '-e load("'"$TEST_ROOT"'/shell.shebang.rb") -- abc ruby' ]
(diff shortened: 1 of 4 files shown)


samueldr:
I have added a (partially dubious) test.

The test only verifies the command line that would be passed to the ruby interpreter and does not actually test with a ruby interpreter.

For the purpose of this PR, the test is sufficient. Executing the test after reverting the fix commit gives me these errors:

```
    /run/user/1000/nix-shell-30420-0/rc: line 1: unexpected EOF while looking for matching `''
    /run/user/1000/nix-shell-30420-0/rc: line 2: syntax error: unexpected end of file
```

This matches with what was observable without tests.

A similar test could be added for the perl shebang handling if pertinent.

tests/nix-shell.sh
```diff
@@ -39,3 +39,12 @@ chmod a+rx $TEST_ROOT/shell.shebang.sh
 
 output=$($TEST_ROOT/shell.shebang.sh abc def)
 [ "$output" = "foo bar abc def" ]
+
+# Test nix-shell shebang mode for ruby
+# This uses a fake interpreter that returns the arguments passed
+# This, in turn, verifies the `rc` script is valid and the `load()` script (given using `-e`) is as expected.
+sed -e "s|@ENV_PROG@|$(type -p env)|" shell.shebang.rb > $TEST_ROOT/shell.shebang.rb
```
shlevy: Can you just hard-code the path to nix-shell instead?

samueldr:
It has been amended so it uses the path to `nix-shell`.

While I did make the change, I would like to know why shoult it use the path to `nix-shell` instead of the equivalent of `/usr/bin/env` for that test, while the previous test uses the equivalent to `/usr/bin/env`?

shlevy:
Mostly just to remove the unnecessary indirection and need to resolve `env`.. I didn't realize we already had a test doing this, sorry!

samueldr:
No problem! Thanks!

PR #2060, opened 2018-04-09
Add all dependencies to source prerequisites

IohannesArnold:
I was trying to bootstrap the 2.0 tarball on an embedded system I was putting together and the configure script required more dependencies than the manual implies. It's entirely possible I'm doing something wrong though.

### DIFF
--- a/doc/manual/installation/prerequisites-source.xml
+++ b/doc/manual/installation/prerequisites-source.xml
@@ -9,6 +9,9 @@
 <itemizedlist>
 
   <listitem><para>GNU Make.</para></listitem>
+  
+  <listitem><para>Bash Shell. Bash, and no other shell, is required to
+  run the <literal>./configure</literal> script.</para></listitem>
 
   <listitem><para>A version of GCC or Clang that supports C++14.</para></listitem>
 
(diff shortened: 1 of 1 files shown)


doc/manual/installation/prerequisites-source.xml
```diff
@@ -9,6 +9,9 @@
 <itemizedlist>
 
   <listitem><para>GNU Make.</para></listitem>
+  
+  <listitem><para>Bash Shell. Bash, and no other shell, is required to
```
shlevy: "and no other shell" reads kind of oddly to me, can you  say something like "the configure script relies on bashisms" or some such?

shlevy:
Thanks! Do you mind squashing these, or if I squash when I merge?

IohannesArnold:
I don't have the git repo downloaded right now, just the tarball. I made the edits over Github's interface and don't think I can squash commits with just the website. So if you could squash them while merging, that would be convenient, thanks.

IohannesArnold:
If this change is merged into master will it show up on the website soon or does it have to wait until the next release? If the latter, should it also get merged into the 2.0 branch? I wrote it in reaction to the 2.0 release tarball; if anyone else is building from the 2.0 source they might appreciate having the live manual agree with the configure script...

PR #2068, opened 2018-04-10
bump version to 2.1

LnL7:
I assume master is 2.1pre now.


shlevy:
@edolstra `nix-env` won't work with a 2.x daemon now that buildenv is a builtin, right? So probably we should be at 3.0?

copumpkin:
Damn, are we going to turn into Chrome and be at version 81 by next year? 😄

edolstra:
Well, this is the problem with semver I mentioned in the release RFC: every non-bug-fix release will have *some* minor incompatibility, so they will all cause a major version bump. At which point the minor version no longer conveys useful information (since there will never be a x.1 release).

BTW, I think we should require the daemon to be the same version as the client. That would make life a lot simpler.

shlevy:
@edolstra How would you like to proceed here then?

Ericson2314:
> BTW, I think we should require the daemon to be the same version as the client. That would make life a lot simpler.

Simple enough so that "every non-bug-fix release will have some minor incompatibility" is no longer true? I'd love to invalidate that fatalism somehow :D.

LnL7:
Is this important for `nixUnstable`? The 2.1 release could be based on master but eg. include some changes to make buildenv backwards compatible.

I would like to bump nixUnstable in nixpkgs, without having it claim to be 2.0pre. 😄

shlevy:
@edolstra Thoughts?

Ericson2314:
`2.1pre` is strictly more accurate than `2.0pre`, is it not? I say merge this now and worry about what the actual release version will be later.

PR #2071, opened 2018-04-12
Add to glossary and clarify garbage collection

sseefried:
While trying to understand garbage collection it was not immediately
clear that only the runtime dependency closure of output paths
would be kept (instead of the build-time dependency closure).
This commit attempts to clarify this by expanding some of the
glossary definitions and extending the Garbage Collection
section.

### DIFF
--- a/doc/manual/glossary/glossary.xml
+++ b/doc/manual/glossary/glossary.xml
@@ -85,29 +85,48 @@
 
 <glossentry xml:id="gloss-reference"><glossterm>reference</glossterm>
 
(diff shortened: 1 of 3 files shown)


sseefried:
@7c6f434c Thanks! 

What is the procedure now? Are you the one that merges it?

7c6f434c:
I don't have commit access to the Nix repo, though, just wanted to record that I have verified that the change is pure improvement.

Hm, I cannot even request reviews by GitHub means…

Sigh. @shlevy is it small enough to merge quickly?

Issue #2076, opened 2018-04-14
nix repl does not interrupt partially typed in expressions on Ctrl-C


obadz:
For instance:

```
$ nix repl
Welcome to Nix version 2.0. Type :? for help.

nix-repl> let a = 1
          
          ^C
          ^C
          ^C
```
I have to know to end this expression with "`; in a`" in order to get a prompt again.

(Possibly related to #1757)

Using Nix 2.0.

NinjaTrappeur:
Indeed, I have a similar behavior.

I think I get what's wrong here, I'll submit a PR fixing that shortly.

On Sat, Apr 14, 2018, at 13:31, obadz wrote:
> For instance:


> $ nix repl Welcome to Nix version 2.0. Type :? for help.  nix-repl>
> let a = 1  ^C ^C ^C
>> I have to know to end this expression with "; in a" in order to get a
> prompt again.> (Possibly related to #1757[1])


> Using Nix 2.0.


> — You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub[2], or mute the
> thread[3].> 


Links:

  1. https://github.com/NixOS/nix/issues/1757
  2. https://github.com/NixOS/nix/issues/2076

PR #2080, opened 2018-04-16
Nix repl does not interrupt partially typed in expressions on Ctrl-C

NinjaTrappeur:
This is a one liner fixing #2076 .

```
nix-repl> let
          a = "oops"
          ^C

nix-repl> 
```


edolstra:
Thanks!

Issue #2077, opened 2018-04-15
Nix allows mixing nested attrpaths and top-level sets... In an order dependent way


shlevy:
> nix-repl> { x.q = 3; x = { y = 3; z = 3; }; }
> error: attribute 'x' at (string):1:12 already defined at (string):1:3
> 
> nix-repl> { x = { y = 3; z = 3; }; x.q = 3; } 
> { x = { ... }; }

shlevy:
Thanks to @ryantrinkle for finding this.

NinjaTrappeur:
Hey,

I investigated a bit on this problem. It turns out being a bit more complicated than it seems.

Everything boils down to how we determine an attrPaths. As it is, the parser makes the assumption that the path is defined by the left-side part of a binding. So basically, it does not support mixed nested attrpaths.

This problem can lead to the example found by @ryantrinkle, but it can also lead to some more nasty behaviors not throwing any error:

```
nix-repl> { x = { y.y = 1; }; x = { y.z = 2; };  }.x.y   
{ z = 2; }

```

I think this issue is more of a bug than an improvement.

I see two options here:

1. Forbidding altogether mixed nested attrpaths and throwing an error every time we encounter that case.
2. Rewriting the addAttr function in parser.y: we need to extract the paths from e (the Expr) when it's an ExprAttrs and inject them in a AttrPath set.

I am more in favor of solution 2, but it wont come for free performance-wise.

Solution 1 may break some existing nix expression and make some people angry.

What do you think?

PR #1152, opened 2016-12-10
Add Travis-CI

zimbatm:
This is a work-in-progress so PRs can be tested for compilation success. Once that's working we can start thinking of adding static analysis tools like coverity to the mix.

Right now the issue is that Travis only provides GCC 4.8.4 and Clang 3.5.0. Both fail on the `std::regex_replace` usage as it's probably too new[1]. I think it's important for nix to be compatible with older C++ implementations as it's necessary to bootstrap on older linux distributions.

[1]: https://travis-ci.org/zimbatm/nix/builds/182907114

PR #2089, opened 2018-04-18
Parse mixed nested attrpaths and top-level sets in an order independant way

NinjaTrappeur:
Hi,

This PR implements #2077. I also added some test cases.

I am a bit confused about `tests/lang/parse-fail-dup-attrs-6.nix`.

```
{
  services.ssh.port = 23;
  services.ssh = { enable = true; };
}
```

To me, it clearly looks like the case shown in issue #2077.

Did we simply forget we should not allow mixed attrPaths and forgot to throw an error for the other way around? I am maybe missing something obvious here :).

On a side-note, from an outsider perspective, the addAttr function was confusing. I feel like it could be divided into two functions. I was unsure about this, I just added some comments.


shlevy:
@NinjaTrappeur Thanks so much for this! I actually think we should *fail* in both directions, not succeed... That is, we should have a copy of `parse-fail-dup-attrs-6.nix` in the opposite direction and then get that one to fail to parse as well :smile:

shlevy:
(I think the existence of that test case is further evidence that #2077 should be a bug :grinning: )

NinjaTrappeur:
Mmh, I am a bit confused here.

I though that `a.b=1; <=> a={b=1;};` [1]. Is this a false assumption?

shlevy:
No, you're correct. But mixing declaration styles like the test suite example does is confusing and unnecessary, it's better not to support it at all. Either use one or the other.

This becomes especially apparent in larger sets. If I see `a = { b = 1; };` in a set, I shouldn't have to look at the entire rest of the set to ensure that `set.a == { b = 1; }`. If we allow mixed declarations, then this becomes possible.

shlevy:
@edolstra Can you make a call here? IMO mixing declarations is just begging for confusion.

edolstra:
Probably it would have been better to disallow both cases from the start, but we probably don't want to break backwards compatibility. So from that perspective, it's cleaner to accept both.

Does this PR merge nested attrsets at arbitrary depths, e.g.
```
{
  services.ssh.enable = true;
  services.ssh = { port = 123; };
  services = {
    httpd.enable = true;
  };
```
?

NinjaTrappeur:
Yes. I just added this test case to this PR.

```
src/nix/nix eval -f tests/lang/parse-okay-mixed-nested-attrs-3.nix services
{ httpd = { enable = true; }; ssh = { enable = true; port = 123; }; }
```

I have a loosely-related question: is there any way to run a CI (hydra?) on a PR?

copumpkin:
> I have a loosely-related question: is there any way to run a CI (hydra?) on a PR?

Unfortunately not unless you have powers over Hydra. There's a thread on getting Travis support (#1152) but it's stalled.

flokli:
What's the status of this? Is this good to merge?

NinjaTrappeur:
It is.

The feature is tested. I can rebase the branch if needed.

I've been bitten by this particular issue several times since I pushed this PR.

edolstra:
Thanks, merged!

PR #2090, opened 2018-04-18
Allow multiple search experssions in nix search

poelzi:
The common use case is to search for packages containing multiple words
like a "git" "frontend". Having only one expressions makes this simple regular
use case very complicated. Instead, search accepts multiple regular epressions
which all need to match.

nix search git 'gui|frontend'

returns a list of all git uis for example


edolstra:
Thanks!

PR #2099, opened 2018-04-20
download: make hash mismatch error consistent with fetchurl

LnL7:
Currently the hash mismatch from nix shows the expected/actual hashes in a different order than eg. fetchurl from nixpkgs.

---

error: hash mismatch in file downloaded from 'http://ftpmirror.gnu.org/hello/hello-2.10.tar.gz': expected hash 'sha256:0000000000000000000000000000000000000000000000000000', got 'sha256:0ssi1wpaf7plaswqqjwigppsg5fyh99vdlb9kzl7c9lng89ndq1i'

vs

fixed-output derivation produced path '/nix/store/3x7dwzq014bblazs7kq20p9hyzz0qh8g-hello-2.10.tar.gz' with sha256 hash '0ssi1wpaf7plaswqqjwigppsg5fyh99vdlb9kzl7c9lng89ndq1i' instead of the expected hash '0000000000000000000000000000000000000000000000000000'

### DIFF
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -730,8 +730,8 @@ Path Downloader::downloadCached(ref<Store> store, const string & url_, bool unpa
         Hash gotHash = unpack
             ? hashPath(expectedHash.type, store->toRealPath(storePath)).first
             : hashFile(expectedHash.type, store->toRealPath(storePath));
-        throw nix::Error("hash mismatch in file downloaded from '%s': expected hash '%s', got '%s'",
-            url, expectedHash.to_string(), gotHash.to_string());
+        throw nix::Error("hash mismatch in file downloaded from '%s': got hash '%s' instead of the expected hash '%s'",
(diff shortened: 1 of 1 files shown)


dtzWill:
So this is a bit more complicated--and the ordering is intentional .  That said I also have found it confusing since I'm used to the other order as well, and ended up copying wrong hash from it... haha. So I am in favor of improving things :).

Turns out there are a number of places Nix reports wrong hash--and many (all) of them use the order "expected hash X, got Y"--so I wrote it this way to "fit in".

It seems strange to be inconsistent within a single codebase about the ordering.

Perhaps it would make sense to replace all instances of this message?

LnL7:
We could also change nixpkgs. I only realised this because of the recent change, if there are other cases where the order is different it probably explains why I always feel confused about what the correct hash is. 😃

shlevy:
Yeah, I think this ordering is better too.

Issue #1238, opened 2017-02-18
Expose `/etc/nsswitch.conf` and `/etc/protocols` etc to sandboxed nix builds?


bennofs:
Right now, the following derivation fails in the nix sandbox:

```
with (import <nixpkgs> { config = {}; });

stdenv.mkDerivation {
  name = "networking";
  buildCommand = ''
    ${strace}/bin/strace getent protocols tcp > $out
  '';
}
```

This means that even packages that run their own server on localhost for testing fail, because `getprotobyname` fails.

bennofs:
See alephcloud/hs-configuration-tools#40 for a package affected by this

copumpkin:
Seems reasonable to me. Would need to switch on host system since these don't exist on Darwin.

My inclination would be to have them contain super boring predefined values that Nix defines somewhere, rather than passing through the host. That improves determinism, and we likely don't need anything fancy someone might have added to `nsswitch.conf` (which even supports loadable libraries!) inside a test that wants to hit localhost.

domenkozar:
+1, just being able to do reverse dns on `/etc/hosts` would be a good start for some tests.

domenkozar:
If I understand correctly this needs `hosts: files` entry for `/etc/nsswitch.conf` besides https://github.com/NixOS/nix/blob/master/src/libstore/build.cc#L1990

bennofs:
@domenkozar yes, and we should probably also use `files` for the other databases, like `protocols` and `services`.

bennofs:
(And add `/etc/protocols` + `/etc/services`)

edolstra:
I think this can already be done using `--sandbox-paths /etc/protocols:...`.

copumpkin:
What I was proposing was for them to be dummy files though, much like how `/etc/passwd` doesn't actually contain real `passwd` entries but is populated in `build.cc`.

edolstra:
Yes, but especially for /etc/services it's not clear what the dummy contents should be.

bennofs:
@edolstra is there no standard `/etc/services`? Just use the IANA standard?

jonringer:
Just ran into another use case where this would have been incredibly helpful to have an `/etc/protocols` for the pythonPackages.eventlet package, which references it for the `socket.getprotobyname('tcp')` python command. (similar to @Mic92's example)

nobodyinperson:
[...]

PR #2104, opened 2018-04-24
build: add nsswitch.conf chroot

disassembler:
This adds nsswitch to chroot with only hosts mapping to files.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1998,8 +1998,10 @@ void DerivationGoal::startBuilder()
                 "nogroup:x:65534:\n") % sandboxGid).str());
 
         /* Create /etc/hosts with localhost entry. */
-        if (!fixedOutput)
+        if (!fixedOutput) {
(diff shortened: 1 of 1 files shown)


edolstra:
Seems fine, but what is the motivation/effect? Will looking up `localhost` still work?

cleverca22:
also, other tests fail due to a missing `/etc/protocols` leading to them not knowing what tcp is, so they cant create a tcp socket

cleverca22:
i also just had a thought, nix.conf can inject paths like /bin/sh into the sandbox, maybe it can get these files also?

cleverca22:
i can confirm that `nix.sandboxPaths = [ "/etc/nsswitch.conf" "/etc/protocols" ];` fixes both issues, the PR is no longer required

disassembler:
Should I close this since `nix.sandboxPaths` solves the issue? The initial reason was running a multi-node cluster integration test, in which the application was dependent on nsswitch.conf to lookup localhost properly so the nodes would trust each other.

domenkozar:
see https://github.com/NixOS/nix/issues/1238

PR #2106, opened 2018-04-24
nix: Add support for --keep-going and --keep-failed

### LINKED ISSUE #2105: Support '-j', '-k' and '-K' in the 'nix' command?
I find that I often end up having to fall back to `nix-build` after using `nix build` as the latter lacks support for the `--keep-going` and `--keep-failed` flags.

These are useful flags, that should be supported in one form or another in `nix build`.

bgamari:
Adds the same flags as were previously used in `nix-build`.

Fixes #2105.

### DIFF
--- a/src/nix/build.cc
+++ b/src/nix/build.cc
@@ -22,6 +22,19 @@ struct CmdBuild : MixDryRun, InstallablesCommand
             .longName("no-link")
             .description("do not create a symlink to the build result")
             .set(&outLink, Path(""));
+
+        mkFlag()
+            .longName("keep-failed")
+            .shortName('K')
+            .description("keep temporary directories of failed builds")
+            .set(&(bool&) settings.keepFailed, true);
+
+        mkFlag()
+            .longName("keep-going")
+            .shortName('k')
+            .description("keep going after a build fails")
+            .set(&(bool&) settings.keepGoing, true);
+
     }
 
     std::string name() override


dtzWill:
So, honestly this would satisfy all of the uses I have for `-k` and `-K`, but even so-- should these shortcuts be specific to `nix build`? Or maybe pushed up to work for all nix commands that build things?

dtzWill:
I suppose `--keep-going` doesn't really make sense, I guess I just wanted to avoid the property we currently enjoy which is that you can often just replace `nix log` with `nix build` and such without worrying about which subcommands take which arguments.

And I suppose this should probably be added to documentation?

Also TIL that I can still set those values with `--option` haha, I always just switched to using `nix-build` :/.

bgamari:
Hmm, so apparently `--keep-going` is already supported. I'm a bit unsure of whether it would be better to:
 a. add the short form to the mechanism that currently handles it (which I believe is the `Setting` class; see `globals.hh` line 97) and drop this patch, or
 b. just drop the long form from this patch

Ma27:
@edolstra any chance to get this merged? Would be really useful to have this for flakes as well.

edolstra:
`--keep-going` and `--keep-failed` already work because all nix.conf options are also supported as command line flags. So this only adds the `-k` and `-K` shortcuts, but I'm not sure whether they're sufficiently commonly used to justify a shortcut.

asymmetric:
@edolstra what's the point of having `--keep-going` as an option for the `nix` super-command, rather than only for `nix build`? Is it used in any other sub-command

I had a hard time finding this option, as I was only searching for it in `nix help build`.

regnat:
> @edolstra what's the point of having --keep-going as an option for the nix super-command, rather than only for nix build? Is it used in any other sub-command

I guess it also makes sense (at least) for `nix shell`, `nix develop` or `nix run`. And also probably for `nix eval --allow-import-from-derivation`

edolstra:
> what's the point of having --keep-going as an option for the nix super-command

There is no particular point, but `keep-going` and `keep-failed` are `nix.conf` settings, and *all* settings have a corresponding command-line option.

Ericson2314:
Perhaps this should be closed, and instead if we want to think about which command should have which options with which shortcuts, #5638 is a better venue.

Ericson2314:
Closing for the reason above. We have the flags already, and we need better infrastructure if we want to limit the commands that that support these flags in a precise way.

PR #2126, opened 2018-04-30
nix-collect-garbage: fix documentation to not mention options like --print-dead

DavidEGrayson:
I installed Nix 2.0.1 the other day and I saw that the man page for `nix-collect-garbage` list some options that aren't actually recognized, like `--print-dead`.

Pull request #1648 was supposed to fix the documentation by removing unsupported options, but I think it got merged into the  wrong branch.  I cherry-picked the one commit from there.  

I've confirmed that all four options being removed from the documentation in this pull request are indeed not recognized.


edolstra:
Thanks!

PR #2139, opened 2018-05-07
support S3 multipart uploads for large NARs

AmineChikhaoui:


src/libstore/s3-binary-cache-store.cc
```diff
@@ -271,34 +277,76 @@ struct S3BinaryCacheStoreImpl : public S3BinaryCacheStore
...
-            .WithKey(path);
+        auto stream = std::make_shared<istringstream_nocopy>(data);
 
-        request.SetContentType(mimeType);
+        auto maxThreads = std::thread::hardware_concurrency();
 
-        if (contentEncoding != "")
-            request.SetContentEncoding(contentEncoding);
+        auto executor =
+            std::make_shared<Aws::Utils::Threading::PooledThreadExecutor>(maxThreads);
```
edolstra: This thread pool should be shared between invocations, otherwise e.g. `nix copy` will start cores^2 threads (since it will do multiple copies in parallel). E.g.
```
static std::make_shared<Aws::Utils::Threading::PooledThreadExecutor> = std::make_shared<...>(maxThreads);
```
```

File: src/libstore/s3-binary-cache-store.cc [hunk 2]
```diff
@@ -271,34 +277,76 @@ struct S3BinaryCacheStoreImpl : public S3BinaryCacheStore
...
+        transferConfig.uploadProgressCallback =
+            [&](const TransferManager *transferManager,
+                const std::shared_ptr<const TransferHandle>
+                    &transferHandle) {
+              //FIXME: find a way to properly abort the multipart upload.
+              checkInterrupt();
+              printTalkative("upload progress ('%s'): '%d' of '%d' bytes",
+                             path,
+                             transferHandle->GetBytesTransferred(),
+                             transferHandle->GetBytesTotalSize());
```
edolstra: This looks like it could really spam the console. It should use Nix's progress API.

edolstra: I guess it's fine since it requires -v.

AmineChikhaoui: yeah it's only when you specify a higher debug level, not sure how many v's though :D

src/libstore/s3-binary-cache-store.cc
AmineChikhaoui: @edolstra  I pushed a fix in case you missed it :)

Issue #1964, opened 2018-03-12
Nix 2.0: nix-channel --update won't update if the user has updated that channel recently


ledettwy:
`nix-channel` as of 2.0 started using `downloadCached` to fetch channel tarballs.

https://github.com/NixOS/nix/blob/88cd2d41acb994684a3e4ead1b1676019f43b4b6/src/nix-channel/nix-channel.cc#L89

`downloadCached` uses `tarball-ttl` as a TTL for the tarball it's downloading, and is apparently set to 1 hour by default.

https://github.com/NixOS/nix/blob/88cd2d41acb994684a3e4ead1b1676019f43b4b6/src/libstore/download.hh#L52-L57

This means that running `nix-channel --update` will not pull new changes if you have already pulled changes within the last hour, which I do not think is good UX, especially for custom channels where code might be deployed more frequently than in nixpkgs.

I propose that something like an option be added to `downloadCached` which overrides the TTL, and that `nix-channel` should set this option to `0`. If a user is bothering to manually invoke `nix-channel --update`, that means that they would like the most up-to-date version of the channel, not a cached version sitting on their computer.

**Workaround:** set `tarball-ttl` to 0 in your `nix.conf` file.

PR #2145, opened 2018-05-09
make sure not to use cached channels for nix-channel --update

yorickvP:
https://github.com/NixOS/nix/issues/1964

### DIFF
--- a/src/libstore/download.hh
+++ b/src/libstore/download.hh
@@ -2,6 +2,7 @@
 
 #include "types.hh"
 #include "hash.hh"
+#include "globals.hh"
 
 #include <string>
 #include <future>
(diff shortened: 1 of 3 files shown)


src/libstore/download.hh
```diff
@@ -55,6 +55,8 @@ struct Downloader
        recent version, and if so, download it to the Nix store. */
     Path downloadCached(ref<Store> store, const string & uri, bool unpack, string name = "",
         const Hash & expectedHash = Hash(), string * effectiveUri = nullptr);
+    Path downloadCached(ref<Store> store, const string & uri, bool unpack, string name,
+        const Hash & expectedHash, string * effectiveUri, int ttl);
```
edolstra: Why a separate method? Wouldn't adding a parameter `int ttl = settings.tarballTtl` work?

yorickvP:
@edolstra updated

edolstra:
Thanks!

Issue #1532, opened 2017-08-25
Nix lack remainder and bitwise operators.


volth:
Although operators like ```%``` ```&``` ```|``` ```^``` ```<<``` ```>>``` ```~``` are less useful in a configuration language such as nix, than in a general-purpose programming language, their absence makes some tasks difficult (for example ip/mask calculations for the firewall and the route table)
Instead of just ```1<<prefixLength```, a recursive function to be written, etc
And ```ip & ~((1<<prefixLength)-1)``` becomes very tricky

copumpkin:
Yeah, I've also wanted IP/subnet manipulation functions before

danbst:
If people want to move more of validation into Nix, @edolstra maybe reconsider Nix isn't a configuration-only language anymore?

copumpkin:
IP addresses still seem pretty firmly in the realm of "configuration" to me. Although "configuration" is a pretty ill-defined term 😄

volth:
I think this ticket should be keep open.
It was fixed only partially.
There is still a open question about the shifts and increasing precision of integers

offlinehacker:
I need bitshift and char to ascii operators for base64 encoding, but that's probably another issue

PR #2157, opened 2018-05-12
add builtins: __bitAnd,  __bitOr,  __bitXor

volth:
Partially fixes https://github.com/NixOS/nix/issues/1532


src/libexpr/primops.cc
```diff
@@ -1676,6 +1676,52 @@ static void prim_div(EvalState & state, const Pos & pos, Value * * args, Value &
...
+}
+
+static void prim_bin_shl(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    mkInt(v, state.forceInt(*args[0], pos) << (state.forceInt(*args[1], pos) & 63));
+}
+
+static void prim_bin_unsigned_shr(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    mkInt(v, (unsigned long)state.forceInt(*args[0], pos) >> (state.forceInt(*args[1], pos) & 63));
```
veprbl: ```(123 >> 64)``` evaluates to ```123```?

src/libexpr/primops.cc
```diff
@@ -1676,6 +1676,52 @@ static void prim_div(EvalState & state, const Pos & pos, Value * * args, Value &
...
+}
+
+static void prim_bin_signed_shr(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    mkInt(v, (signed long)state.forceInt(*args[0], pos) >> (state.forceInt(*args[1], pos) & 63));
```
veprbl: AFAIK shifting signed things is implementation-defined. This could use a test to check that it behaves as we expect.

src/libexpr/primops.cc
volth: That's Java's behaviour:

veprbl: Interesting, I did not know about Java. In C/C++ this expression causes undefined behaviour, which you chose to overcome with ```& 63```. This enables the definition where ```123 >> 64 = 123``` which is the one closer to our hardware, and it is also the counterintuitive one. I don't see a use case for such wrapping behaviour in a scripting language such as nix. At the same time going with the "natural" definition would be a bit more user-friendly.

edolstra: IMHO the only sensible thing is for `123 >> 64` to evaluate to 0.

src/libexpr/parser.y
```diff
@@ -338,6 +338,13 @@ expr_op
...
   | expr_op '*' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__mul")), $1), $3); }
   | expr_op '/' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__div")), $1), $3); }
+  | expr_op '%' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__mod")), $1), $3); }
+  | expr_op '&' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__bin_and")), $1), $3); }
+  | expr_op '|' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__bin_or")), $1), $3); }
+  | expr_op '^' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__bin_xor")), $1), $3); }
+  | expr_op '<<' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__bin_shl")), $1), $3); }
+  | expr_op '>>>' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__bin_unsigned_shr")), $1), $3); }
+  | expr_op '>>' expr_op { $$ = new ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__bin_signed_shr")), $1), $3); }
```
edolstra: I feel these should be builtin functions, not operators, in order not to reserve these characters for a fairly uncommon use case (in Nix expressions). I.e. `builtins.bitwiseAnd 123 456` rather than `123 && 456`.

For example, we might want to use `|` as a pipe operator one day (i.e. `(f | g) x` == `g (f x)`).

volth: That might be usual ```|>```

src/libexpr/primops.cc
layus: 0 is the sensible way to go. Otherwise you need to define the size of the integers. `123 >> 64 = 123` only works with 32 or 64 bits. What would be `123 >> 32` ?

I think we should leave these aspects out of nix, and stick with `123 >> 7 = 0`

volth: Size of integer should be defined anyway for ```(1<<32)``` == ```0``` or not.
Or well, even now, for ```65536*65536``` == ```0``` or not.

Nix currently supports 64-bit ints only.

layus: > Nix currently supports 64-bit ints only.

Oh, I expected nix to support arbitrary large integers like haskell or python. I think that many users would be surprised to see their integers overflow in nix. It looks like nobody cared to implement large integers, but I do not think that overflows are the intended behavior.
With arbitrary large integers, (1<<whatever) would never be null.

@edolstra Would it be acceptable to switch to arbitrary-large integers ?

edolstra: @layus I'm not fundamentally opposed to that but it would mean a dependency on a library like gmp.

volth: That would simplify working with IPv6, and perhaps ```sha256``` will be integers too

layus:
Boost has a multiprecision library that can bind to gmp or rely on its 
own header-only implementation.
If you worry about dependencies, this would be a compile-time only 
dependency.

src/libexpr/primops.cc
layus: @edolstra  Well, boost has a mutliprecision, header-only library that comes with a fallback implementation and the ability to plug different backends if performance is needed. This would only require boost as a build dependency, and possibly gmp as a runtime dependency if we really need that level of performance, which should not be the case.

grahamc:
Can you show code demonstrating using this to accomplish:

> for example ip/mask calculations for the firewall and the route table

?

volth:
For example https://github.com/LumiGuide/lumi-example/blob/master/nix/lib.nix#L50

It is full of unreadable hacks working around impossibility to write ```x & 255``` or ```x % 256```

src/libexpr/primops.cc
```diff
@@ -2221,6 +2235,9 @@ void EvalState::createBaseEnv()
     addPrimOp("__sub", 2, prim_sub);
     addPrimOp("__mul", 2, prim_mul);
     addPrimOp("__div", 2, prim_div);
+    addPrimOp("__bin_and", 2, prim_bin_and);
+    addPrimOp("__bin_or", 2, prim_bin_or);
+    addPrimOp("__bin_xor", 2, prim_bin_xor);
```
edolstra: The only remaining issue is that the naming is inconsistent with all other builtin functions, which use camelCase. Maybe `bitAnd` etc.?

volth:
any objections left?

edolstra:
Looks good to me. Could you add some tests?

volth:
ok

Issue #1634, opened 2017-10-28
Improve search feature on Nix 1.12


Ma27:
I just installed Nix 1.12 (`1.12pre5663_c7af84ce`) during your talk on NixCon today :-D
I'm really impressed, however there's one thing I'd like to improve before the stable release.

Before Nix 1.12 I used to run `nox <search-term>` to find a package name. When I compare the results, I currently think that `nox` does its job better:

- the order of `nix search` is a bit confusing. When I run `nix search jq`, the first package I get is `nixpkgs.jquery_ui` although there's a derivation in the nixpkgs package set named `jq` which is the first result in `nox jq` as well.

- `nox` does a better job at displaying the actual search result. It uses bold texts for derivation names and outputs the results in a more compact way. I'm not sure how to change the output of `nix search` yet, but I'm open for a discussion here.

Due to personal limitations I won't be able to join the hackathon, however I'd give it a try in the next weeks :-)

davidak:
A little more intelligent search algorithm would be nice. I'm sure that already exists and only has to be used...

Ma27:
see #1649

Ma27:
@vcunat @shlevy as you stated in #1649 you'd like to "think about better sorting algorithms separately".
Could you elaborate what you have in mind? I'd love to file another patch to improve the stuff I started, however I'm not entirely sure what you'd like to see in the end %)

vcunat:
I suspect it won't be easy to get a good sorting criterion.  Many also wouldn't be implementable simply as a different comparator for the map (if that's what was meant by the separation), as ATM it's just `string -> string`.

Ma27:
understood, so ATM it's the best option to keep the search algorithm as-is, right?

The second feature I requested is improved colorization as I think that results from `nox` are way easier to read. I'm not exactly sure when, but if you're fine with it I'd just file a patch in the next time so we can actually discuss about this based on a simple proposal :-)

Ma27:
@vcunat after #2158 is merged, are you happy with the current state of `nix search` or is there anything I've forgotten? Otherwise I'd close thi sissue %)

PR #2158, opened 2018-05-12
search.cc: improve UX for `nix search`

Ma27:
As proposed in #1634 the `nix search` command could use some
improvements. Initially 0413aeb35d6ee869a98b6565781c1cf47dc80080 added
some basic sorting behavior using `std::map`, a next step would be an
improvement of the output.

This patch includes the following changes:

* Use `$PAGER` for outputs with `RunPager` from `shared.hh`:
  The same behavior is defined for `nix-env --query`, furthermore it
  makes searching huge results way easier.

* Simplified result blocks:
  The new output is heavily inspired by the output from `nox`, the first
  line shows the attribute path and the derivaiton name
  (`attribute path (derivation name)`) and the description in the second
  line.

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -7,19 +7,25 @@
 #include "common-args.hh"
 #include "json.hh"
 #include "json-to-value.hh"
+#include "shared.hh"
 
 #include <regex>
 #include <fstream>
 
 using namespace nix;
 
-std::string hilite(const std::string & s, const std::smatch & m)
+std::string wrap(std::string prefix, std::string s)
+{
+    return prefix + s + ANSI_NORMAL;
+}
+
+std::string hilite(const std::string & s, const std::smatch & m, std::string postfix)
 {
     return
         m.empty()
         ? s
         : std::string(m.prefix())
-          + ANSI_RED + std::string(m.str()) + ANSI_NORMAL
+          + ANSI_RED + std::string(m.str()) + postfix
           + std::string(m.suffix());
 }
 
@@ -75,6 +81,10 @@ struct CmdSearch : SourceExprCommand, MixJSON
                 "To search for git and frontend or gui:",
                 "nix search git 'frontend|gui'"
             },
+            Example{
+                "To display the description of the found packages:",
+                "nix search git --verbose"
+            }
         };
     }
 
@@ -164,14 +174,10 @@ struct CmdSearch : SourceExprCommand, MixJSON
 
                         } else {
                             results[attrPath] = fmt(
-                                "Attribute name: %s\n"
-                                "Package name: %s\n"
-                                "Version: %s\n"
-                                "Description: %s\n",
-                                hilite(attrPath, attrPathMatch),
-                                hilite(name, nameMatch),
-                                parsed.version,
-                                hilite(description, descriptionMatch));
+                                "* %s (%s)\n  %s\n",
+                                wrap("\e[0;1m", hilite(attrPath, attrPathMatch, "\e[0;1m")),
+                                wrap("\e[0;2m", hilite(parsed.fullName, nameMatch, "\e[0;2m")),
+                                hilite(description, descriptionMatch, ANSI_NORMAL));
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


Ma27:
Please keep in mind that this is quite opinionated as I'd like to improve the UX for `nix search`.
In case anybody disagrees with some of the changes (or has better ideas) I'm happy to discuss / try it out :)

edolstra:
:+1: on using the pager (in fact it should be enabled for all `nix` commands that produce output, but it might interact badly with the progress indicator).

However I'm not sure if hiding most of the package information is a good idea. I also don't know if it's actually a UX improvement to make the user guess what the meaning of the output columns is.

Ma27:
First of all, thanks a lot for your feedback!

> However I'm not sure if hiding most of the package information is a good idea. I also don't know if it's actually a UX improvement to make the user guess what the meaning of the output columns is.

you might be right. The main reason why I filed this patch was because I think that the current `nix search` interface is way too bloated and too hard to read. Although it's just a proposal, what do you think of the `nix search --verbose` output (which is inspired by `nox <searchterm>`)?

In the end we could simply drop the `--verbose` argument for `nix search`, however it'd be extremely cool to get some feedback for the new output I built from your for the side as first %)

Ma27:
@edolstra before I continue working on this patch, are you happy with the `verbose` output?
I could remove the `--verbose` behavior in `nix search` for now (we can talk about this in a separate ticket, right?), but I'd really love to improve the output of `nix search` as I consider the current output as too hard to read.

Furthermore I'd be happy to discuss about the verbose output (which I'd really love to see improved) %)

jcrben:
Agreed that the current output is far too verbose. I tend to avoid using `nix search` (and use `nix-env` instead) for that reason. @Ma27 did you consider including a screenshot of what your new output looks like?

Ma27:
@jcrben 

Screenshot for nix search on imgur

edolstra:
That looks pretty okay to me, except that I'd prefer swapping the attribute name and package/version name, since we're de-emphasizing the latter. (I.e. it should be `* nixpkgs.sblim-sfcc (sblim-sfcc-2.2.9)`.)

Ma27:
@edolstra you're right, fixed it.
The last issue (I guess) before we could merge this is whether or not to use `--verbose`: do you think it would make sense to provide a non-verbose version (with at least the attribute path) or should be drop this entirely and show the view I posted above by default?

Ma27:
ping @edolstra can you please elaborate what's missing to finish this PR? I'd love to see this in one of the next releases :blush:

Ma27:
@edolstra any chance to get this merged or is something else missing?

edolstra:
The only remaining problem is that it overloads the meaning of `--verbose`. That flag has always only affected log output (i.e. stderr), now it also has an effect on the primary output (stdout). So `nix search -v` will not only show more details about packages, but it may also print more log messages. Or, if we ever added a `nix.conf` option to set the default verbosity level, it would also affect the output of `nix search`.

Does anybody have a suggestion for an alternative flag name?

Ma27:
> The only remaining problem is that it overloads the meaning of --verbose. That flag has always only affected log output (i.e. stderr), now it also has an effect on the primary output (stdout). So nix search -v will not only show more details about packages, but it may also print more log messages. Or, if we ever added a nix.conf option to set the default verbosity level, it would also affect the output of nix search.

You're absolutely right, I've just seen that Nix supports `--verbose` by default and used it, but I absolutely agree with you, it doesn't make any sense to use a single flag for extended logging and enhanced outputs.

> Does anybody have a suggestion for an alternative flag name?

Something like `--detailed`?

But as you stated previously do you think we should keep such a flag or would it be better to print the enhanced output (as shown in the screenshot above) by default? When I filed this patch I thought that it might be more convenient to have a minimalistic output  by default (which is why I (ab)used `--verbose`), but now I'm not sure if we actually need this :sweat_smile:

Ma27:
@edolstra I dropped the `--verbose` usage, I guess that two lines per search result are fairly reasonable, right?

edolstra:
Thanks!

PR #2163, opened 2018-05-15
builtins.path: Don't force the path argument if it's in the store.

shlevy:
This allows us to skip potentially expensive steps, like fetchGit or
import-from-derivation, when we specify a hash up-front.


shlevy:
@edolstra ping? Worth bringing up to date?

PR #2168, opened 2018-05-18
manual: document why ~/.netrc doesn't work

grahamc:
Maybe there is a better place to document this, which is more generic?

Should probably backport to the stable release.


jcrben:
With https://github.com/NixOS/nix/pull/2170 how about just updating to expand in the conf file? Consistency is good UX.

PR #2170, opened 2018-05-20
Expand tilde to home directory in NIX_PATH

dezgeg:
Given that Nix path expressions can start with a tilde, allow the same for NIX_PATH entries.

This is needed for fixing https://github.com/NixOS/nixpkgs/issues/40165 (or well, my suggestion in https://github.com/NixOS/nixpkgs/issues/40165#issuecomment-387512180 depends on that) so this should be picked to the 2.0.x branch as well.

### DIFF
--- a/src/libexpr/parser.y
+++ b/src/libexpr/parser.y
@@ -663,7 +663,7 @@ std::pair<bool, std::string> EvalState::resolveSearchPathElem(const SearchPathEl
             res = { false, "" };
         }
     } else {
-        auto path = absPath(elem.second);
+        auto path = absPath(expandTilde(elem.second));
         if (pathExists(path))
             res = { true, path };
         else {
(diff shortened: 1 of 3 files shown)


shlevy:
@grahamc just hit this for paths in nix.conf too, might be nice there.

shlevy:
@dezgeg Can Nix path expressions actually start with a tilde, or is that just shell expansion?

dezgeg:
Yes they do: https://github.com/NixOS/nix/blob/1ad19232c4bbecb06e6acbb2a3a538544a28e1f2/src/libexpr/parser.y#L377

dtzWill:
This might not always be clear who's home should be used depending on evaluation. I hit this sometimes when I lazily point src to local repo with tilde and it works until I run with sudo :).

FWIW "~user" syntax is not supported.

PR #2171, opened 2018-05-21
libstore: Verify and repair path metadata on `nix-store --verify --check-contents'

dezgeg:
Currently, if you somehow end up with store paths that contain an
invalid permission (i.e. not 444 or 555) or mtime (i.e. not 1),
figuring out the problem and fixing it is quite difficult. (See e.g.
NixOS/nixpkgs#32242).

Fortunately, this is quite simple to fix by calling
canonicalisePathMetaData() in --check-contents handling, just need to
add some prints to show when problems are getting repaired.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -3080,7 +3080,7 @@ void DerivationGoal::registerOutputs()
             /* Canonicalise first.  This ensures that the path we're
                rewriting doesn't contain a hard link to /etc/shadow or
                something like that. */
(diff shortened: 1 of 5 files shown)


teto:
that's really cool. I have no idea why but it sometimes happens with iproute package that files have wrong permissions. My first attempt at nix-store --delete nuked my config, As a fallback I now mounted the store as rw and change the permissions manually whenever I need to. This PR would obviosuly been better.

Issue #1896, opened 2018-02-23
Nix cannot be installed through SSL intercepting proxy


makefu:
Installing nix fails when an intercepting SSL proxy is in between the host and the internet.

Running curl manually works as the respective root certificates are added to the ca certificate trust store.

```console
[root@dev ~]# mkdir /etc/nix
[root@dev ~]# echo "build-users-group =" > /etc/nix/nix.conf
[root@dev ~]# curl https://nixos.org/nix/install | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2455  100  2455    0     0   5265      0 --:--:-- --:--:-- --:--:--  5268
downloading Nix 2.0 binary tarball for x86_64-linux from 'https://nixos.org/releases/nix/nix-2.0/nix-2.0-x86_64-linux.tar.bz2' to '/tmp/nix-binary-tarball-unpack.Ml95QP5hQK'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 21.0M  100 21.0M    0     0  1901k      0  0:00:11  0:00:11 --:--:-- 3204k
warning: installing Nix as root is not supported by this script!
performing a single-user installation of Nix...
copying Nix to /nix/store...............................
initialising Nix database...
installing 'nix-2.0'
building '/nix/store/ni5hj5bcgqlfpvlyl283jj0x9jmiy5kk-user-environment.drv'...
created 6 symlinks in user environment
warning: unable to download 'https://nixos.org/channels/nixpkgs-unstable': Peer certificate cannot be authenticated with given CA certificates (60); retrying in 275 ms
warning: unable to download 'https://nixos.org/channels/nixpkgs-unstable': Peer certificate cannot be authenticated with given CA certificates (60); retrying in 608 ms
warning: unable to download 'https://nixos.org/channels/nixpkgs-unstable': Peer certificate cannot be authenticated with given CA certificates (60); retrying in 1185 ms
warning: unable to download 'https://nixos.org/channels/nixpkgs-unstable': Peer certificate cannot be authenticated with given CA certificates (60); retrying in 2551 ms
error: unable to download 'https://nixos.org/channels/nixpkgs-unstable': Peer certificate cannot be authenticated with given CA certificates (60)
[root@dev ~]# curl https://nixos.org/channels/nixpkgs-unstable
[...]

Issue #1211, opened 2017-01-31
`nix-channel --update` masks failures


grahamc:
This was an issue I ran in to, installing Nix in an environment with a proxy. The version of curl already installed masked a configuration mistake in the `https_proxy` variable. The (much) newer version of Curl packaged with Nix exposed this issue. From the perspective of installing Nix, it would get to the `nix-channel --update` stage and fail like this:

```
nix-channel: unable to check ‘https://nixos.org/channels/nixpkgs-unstable’
```

This is because the underlying curl call in the `nix-channel` perl script was failing. However, the nix-channel code masks the issue and doesn't expose any information as to why or how it failed.

Being familiar enough with Nix, I read the source to `nix-channel` and found it essentially calls:

```
/nix/store/y2dzvy1ji14cfya2rblnmyhcbfjfc75r-curl-7.52.1-bin/bin/curl https://nixos.org/channels/nixpkgs-unstable
```

which yielded the following error:

```
curl: (35) error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol
```

leading me on to discover the environment's misconfiguration. Having our tools expose more detailed information would more readily allow novice Nix users the ability to start debugging their own problems.

(FWIW the problem is with `https_proxy="https://foobar:1111"` old curl will ignore `https` and substitute in `http`, new curl can handle HTTPS proxies, and will try to negotiate TLS with the proxy server.)

mickours:
I have the same kind of error message but for an other proxy error:
```
sudo nixos-rebuild switch --upgrade --verbose
/run/current-system/sw/bin/nix-channel: unable to check ‘https://nixos.org/channels/nixos-17.03-small’
```
The real problem is the error is mask, and the internal curl error don't show up even with the verbose flag.

I tried manually and I get this:
```
/run/current-system/sw/bin/curl https://nixos.org/channels/nixos-17.03-small
curl: (56) Received HTTP code 407 from proxy after CONNECT
```
Just adding some meaningful message in the ``nix-channel`` Perl script at line 110 can do the trick. I'd like to push a PR but I'm really not familiar with Perl, any idea how to get this message and show it to the user?

For convinience, the code snippet that need to be updated:
```perl
        ($url, $filename) = `cd $tmpdir && $Nix::Config::curl $Nix::Config::curlCaFlag --netrc-file $netrcFile --netrc-optional --silent --write-out '%{url_effective}\n%{filename_effective}' -L '$url' -O`;
[...]

PR #2181, opened 2018-05-25
Install Nix behind MITM proxy

### LINKED ISSUE #2180: Installing Nix behind a HTTPS-intercepting firewall fails on macOS
Long term, the ideal would probably be to have Nix talk to Keychain to for certificate trust, however shorter term we should probably be able to override the trusted certificate bundle at installation time.

grahamc:
Sometimes corporations will have MITM proxies re-signing all their HTTPS traffic with their own certs. Our daemon-based installer currently fails under this scenario, because it forces its own set of certificates.

The daemonless installer, however, allowed installers to override the cert environment variable.

This PR brings the daemon-based installer in line feature-wise, and documents the feature.

I would like to backport this.

Closes #2180, #1896, #1211 (I think?)

### DIFF
--- a/doc/manual/installation/env-variables.xml
+++ b/doc/manual/installation/env-variables.xml
@@ -21,4 +21,51 @@ in your <filename>~/.profile</filename> (or similar), like this:</para>
 <screen>
 source <replaceable>prefix</replaceable>/etc/profile.d/nix.sh</screen>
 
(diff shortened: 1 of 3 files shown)


makefu:
This is great! I will try this PR once i am at work again on monday

PR #2194, opened 2018-05-30
Nix profile & installation: handle set -u installations

grahamc:
I've verified the profile changes work properly, and the install-multi-user changes should work properly: in the block where we check the value, we export a value for it.


grahamc:
Here is how I verified the profile fix:

```
#!/bin/sh

set -eu

[ "${HOME:-}" != "" ] && exec -c $0 "$(dirname "$(which env)")" "$USER" "$HOME"

PATH="$1"
USER="$2"
HOME="$3"

(
    set -u
    __ETC_PROFILE_NIX_SOURCED=""
    NIX_SSL_CERT_FILE=""
    . ./scripts/nix-profile-daemon.sh.in
)
```

I'm building the installer again to validate it on macOS.

PR #2197, opened 2018-05-30
Explicitly describe the effects of filterSource excluding a directory.

shlevy:


mightybyte:
Very helpful.  It took me awhile to figure this out because at first glance I assumed it worked like a purely functional list filter where each element is analyzed independently.

PR #2201, opened 2018-06-01
Add a delay before killing the builder process

### LINKED ISSUE #2176: Builder is sometimes unexpectedly killed
Nix sometimes fails to build the expression
```nix
with import <nix/config.nix>;

{ simple = builtins.derivation {
    name = "simple";
    system = "x86_64-linux";
    PATH = coreutils;
    builder = shell;
    args = ["-c" "touch $out"];
  };
}
```

Generally, the build works but for ~1% of attempts, it fails with
```
builder for '/nix/store/94yfdqk4r6vmja7qhqkyy0hl25bvvfjj-simple.drv' failed due to signal 9 (Killed)
```

I can reproduce on my laptop (nix 2.0) and in a vm (nix 2.0.2) with:
```sh
$ nix-build -E 'with import <nix/config.nix>; { simple = builtins.derivation {name="simple";system="x86_64-linux";PATH=coreutils;builder=shell;args=["-c" "touch $out"];};}'

$ i=0; while nix-build -E 'with import <nix/config.nix>; { simple = builtins.derivation {name="simple";system="x86_64-linux";PATH=coreutils;builder=shell;args=["-c" "touch $out"];};}' --check; do echo $i; i=$(($i+1)); done
```

Note I'm not able to reproduce if a delay is introduced. The following expression has been successfully built 6000 times:
```nix
with import <nix/config.nix>;

{ simple = builtins.derivation {
    name = "simple";
    system = "x86_64-linux";
    PATH = coreutils;
    builder = shell;
    args = ["-c" "sleep 0.5; touch $out"];
  };
}
```

Are you also able to reproduce?

nlewo:
Sometimes, the builder process is killed while the build succeeded. It
seems this is because the signal is sent after file descriptors are
closed (in `do_exit`) and before the process reaches the terminated
state. This leads to a build failure (failed due to signal 9).

To mitigate this issue, a delay (~10s) is introduced before sending the kill
signal to the build process if it doesn't reach the terminated state.

Fixes #2176

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1489,9 +1489,11 @@ void DerivationGoal::buildDone()
     /* Since we got an EOF on the logger pipe, the builder is presumed
        to have terminated.  In fact, the builder could also have
        simply have closed its end of the pipe, so just to be sure,
-       kill it. */
-    int status = hook ? hook->pid.kill() : pid.kill();
-
+       kill it after a delay (~10s) if the process is still alive.*/
+    int status = hook ? hook->pid.waitWithTimeout() : pid.waitWithTimeout();
+    if (status == -1) {
+        status = hook ? hook->pid.kill() : pid.kill();
+    }
     debug(format("builder process for '%1%' finished") % drvPath);
 
     result.timesBuilt++;
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -787,6 +787,25 @@ int Pid::kill()
 }
 
 
+int Pid::waitWithTimeout()
+{
+    assert(pid != -1);
+    int timeout = 10000;
+    int i = 9;
+    while (i--) {
+        int status;
+        int res = waitpid(pid, &status, WNOHANG);
+        if (res == pid) {
+            pid = -1;
+            return status;
+        }
+        checkInterrupt();
+        usleep(timeout);
+        timeout *= 2;
+    }
+    return -1;
+}
+
 int Pid::wait()
 {
     assert(pid != -1);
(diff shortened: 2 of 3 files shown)


nlewo:
I'm not sure about the root cause of the issue #2176 but with this patch, I'm no longer able to reproduce it (30000 iterations).
Moreover, this is my first C++/nix patch... so please consider this PR as a poc:/

edolstra:
Adding arbitrary timeouts doesn't seem like a good idea. For instance, why is 10 seconds an appropriate value? Doesn't this mean that the build can still fail randomly under high load? It would probably be better to fix the shell not to close stdout/stderr before exiting.

nlewo:
I agree arbitrary timeouts are not a good idea but I don't know how to "fix" this bug otherwise. Moreover we currently have this timeout but we don't know its value!

I think the problem comes from the way the worker knows the builder finished: we wait for closed file descriptors. But to terminate a process, the Linux kernel seems to first close file descriptors and then, it changes the status of the process.
It is possible the worker process sends the kill signal before these two steps. With this patch, the kernel have 10 seconds to mark the process as terminated.

It is still possible the build fails under high load. But from my basic experiments, I was no longer able to hit this issue with 10sec.

Do you suggest to prevent user build code to close stdout/stderr in order to remove the kill expression? I'll investigate this way.

tomberek:
Ran into this problem when running Nix on AWS Lambda. (https://github.com/tomberek/lambdanix). For what it's worth, this patch fixed it. Not sure why. I was also completely abusing the pseud-terminal interface (AWS does not give you access to one), so this could be other issues interacting.

PR #2205, opened 2018-06-02
straightforward implementation of 'nix copy'

### LINKED ISSUE #1639: nix copy doesn't work for drv files
With nix-copy-closure it's possible to build a drv on a remote machine without using the build hook. This is useful to avoid substituting dependencies over a slow connection.

```
$ nix copy --to ssh://desktop /path/to/build.drv
$ ssh desktop -- nix build /path/to/build.drv
cannot build missing derivation '/path/to/build.drv'
```

volth:
fixes #1639

### DIFF
--- a/src/nix/copy.cc
+++ b/src/nix/copy.cc
@@ -8,16 +8,17 @@
 
 using namespace nix;
 
-struct CmdCopy : StorePathsCommand
+struct CmdCopy : StoreCommand
 {
     std::string srcUri, dstUri;
+    bool recursive = false;
+    std::vector<std::string> srcPaths;
 
     CheckSigsFlag checkSigs = CheckSigs;
 
     SubstituteFlag substitute = NoSubstitute;
 
-    CmdCopy()
-        : StorePathsCommand(true)
+    CmdCopy() : StoreCommand()
     {
         mkFlag()
             .longName("from")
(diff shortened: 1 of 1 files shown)


shlevy:
I think sign-paths has the same problem

volth:
```nix copy``` also lacks ```--include-outputs``` feature of ```nix-copy-closure```

volth:
And... it there any meaningful use case for ```nix copy``` without ```--recursive``` ?

shlevy:
IMO recursive should be the default, but if you know that the rest of the closure is already there it can be more efficient to avoid several checks for the validity of the inputs.

volth:
```--recursive``` leaded here from ```StorePathsCommand```, ```nix-copy-closure``` has it always on.

> IMO recursive should be the default, but if you know that the rest of the closure is already there it can be more efficient to avoid several checks for the validity of the inputs.

Well, if the assumption were false, and a derivation which should be copied with ```--recursive``` were copied without, would it result in dangling references in the target nix store?

shlevy:
No, it would just fail.

shlevy:
Anyway, we can add it later if need be.

edolstra:
:-1: since this makes the `nix copy` interface inconsistent with other `nix` commands. For instance, you can't do something like `nix copy nixpkgs.hello` anymore.

volth:
> For instance, you can't do something like ```nix copy nixpkgs.hello``` anymore.

Never tried to do something like this.

I am worrying about impossibility to do ```nix copy /nix/store/something.drv```.
It results with success but copies nothing (or only .patch and .sh files)

edolstra:
I've pushed 85530bcc0b4c61499f90fa6125c19beeb0f4485e which allows copying .drv files while keeping `nix copy` consistent with other `nix` commands.

Issue #1988, opened 2018-03-19
Nix 2.0: nix copy stalls


chris-martin:
I've been deploying a server with `nix copy` for a while now, but suddenly I find it's no longer working.

```
$ nix copy /nix/store/cg47i25qrhp3lcd3hirgwc28pd7yxw0v-nixos-system-tc-webserver1-18.03pre-git --to ssh://example.com --no-check-sigs --verbose
copying 11 paths...
copying path '/nix/store/lkr614grgjfm7dghyw8zj9x5i4dwhmcy-qemu-2.11.1' to 'ssh://example.com'...
warning: dumping very large path (> 256 MiB); this may run out of memory
[1/0/11 copied (406.0/410.7 MiB)] copying path '/nix/store/lkr614grgjfm7dghyw8zj9x5i4dwhmcy-qemu-2.11.1' to 'ssh://example.com'
```

The command pauses (apparently indefinitely) with no clue as to why. If I interrupt it and run it again, I get exactly the same behavior.

Watching `top` on the server, I see that a `nix-store --serve --write` process is running.

I've tried restarting sshd and nix-daemon on the server, and I've rebooted my laptop.

---

Update: After about an hour, the `nix copy` command printed this and halted:

```
error: Nix daemon out of memory
reaping 3 worker threads
killing process 9152
[0 copied (408.5 MiB)]
error: unexpected end-of-file
```

AmineChikhaoui:
It would probably be nice if you could test with the latest master branch as Eelco did some tweaks to reduce memory consumption afaik (https://github.com/NixOS/nix/commit/48662d151bdf4a38670897beacea9d1bd750376a).

chris-martin:
Hi @AmineChikhaoui, I'm trying to do that. I think the [manual](https://nixos.org/nix/manual/#sec-building-source) is out of date - it says the first build step is to run `./configure`, but there's no such script.

AmineChikhaoui:
I generally override the nix package in my configuration.nix e.g of an old setup (you need to pull the latest revision/checksum):
```
nixpkgs.config.packageOverrides = pkgs: {
    nixUnstable = pkgs.nixUnstable.overrideAttrs (oldAttrs: rec{ 
      src = pkgs.fetchFromGitHub {
        owner = "NixOS";
        repo = "nix";
        rev = "179b896acb6deb8fea9614dfbddeaf3b23797bf5";
[...]

Issue #1969, opened 2018-03-12
Nix install run out of memory and fail when binary package size > RAM size


adevress:
In case of binary installation where size of the NAR > size of the RAM, I triggered an issue where Nix 2.0 simply run out of memory and interrupt.

> unexpected error in download thread: std::bad_alloc
download of 'http://********.cache.somewhere/cache/nar/yxnmlrrdng8ff8xk5lrxahvzjyc1i01p.nar.bz2' was interrupted
cannot build derivation '/nix/store/0m6yzz1x57zr81pvvlx6yqcbhnifz74q-all-modules.drv': 1 dependencies couldn't be built
[31;1merror:[0m build of '/nix/store/0m6yzz1x57zr81pvvlx6yqcbhnifz74q-all-modules.drv', '/nix/store/xfr1cbmd8j46dh2dgv20vcqa15fnspd0-all-benchs.drv' failed
Build step 'Execute shell' marked build as failure

This was triggered by a simple "nix-build" and was not existing before nix 2.0

adevress:
This is a major issue because installation something like CUDA 9.0 ( > 2GB ) on a medium machine with 4GB RAM fails...

edolstra:
This should partially be addressed by #619 / #1754. However another reason for the memory use regression is that evaluation and building are now done in a single process (i.e. nix-build is no longer a wrapper around nix-instantiate and nix-store -r). A solution might be to force a Boehm GC run after evaluation.

adevress:
Thx @edolstra

adevress:
Indeed: 

This is on something more "lightweight" when compiling GCC5

There is a memory leaks somewhere

c-family/.deps/c-gimplify.TPo ../../gcc-5.5.0/gcc/c-family/c-gimplify.c

[31;1merror:[0m out of memory
Build step 'Execute shell' marked build as failure
Finished: FAILURE

dtzWill:
Boehm only governs eval memory which is peanuts compared to memory used by copying around std::string's-- profiling shows copying paths is by far the largest contributor to peak memory usage, requiring strings with both the compressed (via binary cache) and decompressed to in memory at the same time (briefly, but nevertheless).

Our usage of std::string::append appears particularly painful, FWIW.  Decompression adds to a string two pages at a time (on glibc)... :(.

This isn't addressed by the linked issues, although is similar in spirit so perhaps can be handled similarly.

[...]

Issue #1681, opened 2017-11-15
nix copy uses too much memory


ledettwy:
I'm running nix copy in runInLinuxVM, and notice that for any nontrivial closures, the VM will run out of memory during the copying process.  I left it set at the default 512 megabytes.  I could obviously increase the amount of memory the VM is given, but that doesn't scale for copying complex derivations with many dependencies.

I suggest adding an option to only load and copy the contents of the paths one at a time, or even better, a way to specify an upper bound on the memory to be used while copying.

copumpkin:
Intuitively it feels that it should be possible for it to run in constant memory. What am I missing?

lheckemann:
I'm encountering this issue with a single path — `nix copy`, `nix-store --import`, and a number of other commands I've tried all fail to import the path. Would be great to know if there's *any* way at all I can import it…

ledettwy:
Possibly related to https://github.com/NixOS/nix/issues/1969 ?  Looks like some patches have gone in recently that might improve things here.

Ralith:
I see commits purporting to address this for a number of different cases, but none concerning uploading to a S3 bucket. Trying to copy a 2.8GB store path to a S3 bucket took nearly 4GB of memory and more than twenty minutes of 100% CPU. Has that been fixed?

andrewchambers:
Hitting this issue trying to do something like -  nixos-rebuild build ; nix copy ./result --to ssh://low_ram_machine

@dtzWill will those experimental changes help with ssh copy?

copumpkin:
FWIW I too am another big-upload-to-S3 guy using `nix copy` 😄(in my case it's to build NixOS AMIs)

It would surprise me if aws-sdk-cpp didn't support it, given that S3 supports almost arbitrarily large objects and multi-part uploads. If someone figured out how to implement it, would you accept the PR?

coretemp:
This is what I am seeing too:
```
a...........> copying path '/nix/store/fl3mcaqqk2vg0dmk01dfbs6nbm5skpzc-systemd-237' from 'https://cache.nixos.org'...
a...........> error: out of memory
```

The main problem I see is that it merely says "out of memory", instead of saying how much it tried to allocate, and how much was available before the allocation in the error message. Copying data should run in constant space as others have already mentioned. 

If the compression is causing higher memory requirements than needed, this is a problem too, because it raises the hosting costs for no reason other than the initial deployment.

Before the deployment at least 300MB was available on host `a`.

[...]

PR #2206, opened 2018-06-03
importPaths: Don't copy imported NAR into memory.

nh2:
Fixes `error: out of memory` of `nix-store --serve --write`
when receiving packages via SSH (and perhaps other sources).

See #1681 #1969 #1988 NixOS/nixpkgs#38808.

Performance improvement on `nix-store --import` of a 2.2 GB cudatoolkit closure:

```
When the store path already exists:
  Before:
    10.82user 2.66system 0:20.14elapsed 66%CPU (0avgtext+0avgdata   12556maxresident)k
  After:
    11.43user 2.94system 0:16.71elapsed 86%CPU (0avgtext+0avgdata 4204664maxresident)k
When the store path doesn't yet exist (after `nix-store --delete`):
  Before:
    11.15user 2.09system 0:13.26elapsed 99%CPU (0avgtext+0avgdata 4204732maxresident)k
  After:
     5.27user 1.48system 0:06.80elapsed 99%CPU (0avgtext+0avgdata   12032maxresident)k
```

The reduction is 4200 MB -> 12 MB RAM usage, and it also takes less time.

### DIFF
--- a/src/libstore/export-import.cc
+++ b/src/libstore/export-import.cc
@@ -4,6 +4,8 @@
 
 #include <algorithm>
 
+#include <fcntl.h>
+
 namespace nix {
 
 struct HashAndWriteSink : Sink
@@ -69,9 +71,22 @@ Paths Store::importPaths(Source & source, std::shared_ptr<FSAccessor> accessor, 
         if (n == 0) break;
         if (n != 1) throw Error("input doesn't look like something created by 'nix-store --export'");
 
+        /* Extract the NAR from the source, writing it to a temporary file. */
+
+        auto tempDir = createTempDir();
+        AutoDelete delTempDir(tempDir, true);
+        Path tempPath = tempDir + "/x";
+
+        AutoCloseFD fd = open(tempPath.c_str(), O_RDWR | O_CREAT | O_CLOEXEC, 0600);
+        if (!fd)
+            throw SysError(format("opening file '%1%'") % tempPath);
+
         /* Extract the NAR from the source. */
-        TeeSink tee(source);
-        parseDump(tee, tee.source);
+        FdTeeSink fdTeeSink(source, fd.get());
+        parseDump(fdTeeSink, fdTeeSink.source);
+
+        fdTeeSink.source.fdSink.flush();
+        size_t readSize = fdTeeSink.source.fdSink.written;
 
         uint32_t magic = readInt(source);
         if (magic != exportMagic)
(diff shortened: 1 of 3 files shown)


nh2:
Convenience override to try this patch out (stick it into the machine config of the target machine, e.g. I put it in `nixops`'s `libvirtd-image.nix`):

```nix
{
      # Patch nix to avoid problem running out of RAM when sending closures to the machine.
      # See
      #   https://github.com/NixOS/nix/issues/1681
      #   https://github.com/NixOS/nix/issues/1969
      #   https://github.com/NixOS/nixpkgs/issues/38808
      nixpkgs.config.packageOverrides = pkgs: {
        nix = pkgs.nixUnstable.overrideAttrs (oldAttrs: {
          src = pkgs.fetchFromGitHub {
            owner = "NixOS";
            repo = "nix";
            rev = "54b1c596435b0aaf3a2557652ad4bf74d5756514";
            sha256 = "0g7knsfj445r50rk0d9hm5n1pv20k542bz6xf5c47qmkgvfa40x4";
          };
          patches = [
            (pkgs.fetchpatch {
              url = "https://github.com/nh2/nix/commit/d31a4410d92790e2c27110154896445d99d7abfe.patch";
              sha256 = "08gcw2xw8yc61zz2nr1j3cnd6wagp5qs02mjfazrd9wa045y26hg";
            })
          ];
          # Changes cherry-picked from upstream nix `release-common.nix` that
          # aren't in `pkgs.nixUnstable` yet:
          buildInputs = oldAttrs.buildInputs ++ [
            (pkgs.aws-sdk-cpp.override {
              apis = ["s3" "transfer"];
              customMemoryManagement = false;
            })
          ];
        });
      };
}
```

nh2:
When we got this fixed, ideally somebody would also update the 18.03 EC2 images so that people can actually use them as a base on EC2 machines with <= 4 GB ram.

edolstra:
This doubles the amount of I/O and disk space needed on the target machine.

bgamari:
@edolstra, in my opinion doubling necessary disk storage in exchange for constant memory operations is absolutely the correct trade-off. Disk space is generally more plentiful than memory and is *far* easier to provision. 

Even in the short nine months using `nix` I have run into memory issues more times than I care to recall. Just today I was unable to rebuild my mail server as `nix` started OOMing after a `nixpkgs` update. While thankfully today's event occurred during non-essential maintenance, this failure could just as easily happen during a more time-critical event. Events like this has made me think twice before using `nix` on smaller machines.

joepie91:
It's worth considering that while reducing disk space by stripping down installed packages is possible as an end user, this isn't really true for memory usage; it's very easy to hit the limit even when installing relatively few things. This suggests that the disk-space-instead-of-RAM tradeoff is worth it.

samueldr:
FWIW, this allowed a build to complete when it otherwise couldn't here. The disk-space-for-RAM tradeoff is important when dealing with big (gigabytes) products on machines with a smaller amount of RAM like SBCs (raspberry pi, etc.).

thoughtpolice:
I'm going to chime in here and say this is a very annoying problem for us at work, where we tend to have very large objects we push into the store -- we have one binary object that's over 30GB of raw data, and right now we have to manually register it into the store with mount/remount hacks. This is really nasty and troublesome because we have to do it on nearly a dozen machines individually.

The disk/memory tradeoff here is *absolutely* the correct one to make, as others have mentioned here. There will always be hierarchies in the storage layer -- slower, larger storage that's further away, and faster, less plentiful storage (DRAM) that's available locally. The fact that this increases I/O on the lower tier while *decreasing* the requirements on the less plentiful, far more expensive tier, is the only argument you need, because it's far more sustainable and scalable than expecting DRAM to magically increase in density or $/GB over night.

You could probably even make an argument based on the ratio of DRAM prices to storage and compute prices have skewed so much over the past few years that trading off disk space isn't just *faster* and actually makes Nix more usable -- it's actually a *superior* architecture for the way modern workloads and system capacity have progressed.

angerman:
What's the holdup with this?

edolstra:
I've gone for a different solution (2825e05d21ecabc8b8524836baf0b9b05da993c6) that doesn't require writing the NAR to disk.

thoughtpolice:
Thank you @edolstra! I'm excited to try this since we actually ran into NAR import issues again just this week (copying large local files into the store). I'll see how this works for our use case of extremely large objects.

PR #2216, opened 2018-06-09
download: if there are active requests, never sleep for 10s

dtzWill:
In some versions/configurations libcurl doesn't handle timeouts
(especially DNS timeouts) in a way that wakes curl_multi_wait.

This doesn't appear to be a problem if using c-ares, FWIW.

See
https://github.com/NixOS/nixpkgs/pull/41619

(and the issues/PR's it refers to)


dtzWill:
Hopefully this can make it to a release, so Nixpkgs doesn't need to special-case the curl used by Nix.

PR #2223, opened 2018-06-11
libstore/gc.cc: ignore ESRCH when reading /proc

symphorien:
When running `nix-store --gc --print-roots`, nix walks `/proc/` which is inherently racy. When unlucky (this happened to me on travis) `nix-store` fails with `error: reading from file: No such process`. It seems the reason is that if a process disappears between the time `/proc/[pid]/maps` is opened and
the time it is read, the `read()` syscall will return `ESRCH`. This should be ignored.

I tried to run the testsuite, but some test failed before my commit. So I can only claim that no new failing test are introduced. Also, nix-du's test suite seems to pass more reliably.


edolstra:
Thanks!

Issue #1756, opened 2017-12-22
"nix repl" cannot tab-complete names that contain a dash ('-')


dtzWill:
# The Problem

Consider default.nix:

```
{
  aaa-bbb = 1;
  aaa-ccc = 2;
}
```

Pointing nix-repl at this, completions are as follows:

## a
```
nix-repl> a
aaa-bbb  aaa-ccc  abort
```

## aaa-
```
nix-repl> aaa-
__add                            __findFile                       __langVersion                    __stringLength                   builtins
__addErrorContext                __foldl'                         __length                         __sub                            derivation
__all                            __fromJSON                       __lessThan                       __substring                      derivationStrict
__any                            __functionArgs                   __listToAttrs                    __tail                           dirOf
__attrNames                      __genList                        __match                          __toFile                         false
__attrValues                     __genericClosure                 __mul                            __toJSON                         fetchGit
__catAttrs                       __getAttr                        __nixPath                        __toPath                         fetchMercurial
__compareVersions                __getEnv                         __nixVersion                     __toXML                          fetchTarball
__concatLists                    __hasAttr                        __parseDrvName                   __trace                          import
__concatStringsSep               __hashString                     __partition                      __tryEval                        isNull
__currentSystem                  __head                           __pathExists                     __typeOf                         map
__currentTime                    __intersectAttrs                 __readDir                        __unsafeDiscardOutputDependency  null
__deepSeq                        __isAttrs                        __readFile                       __unsafeDiscardStringContext     placeholder
__div                            __isBool                         __replaceStrings                 __unsafeGetAttrPos               removeAttrs
__elem                           __isFloat                        __seq                            __valueSize                      scopedImport
__elemAt                         __isFunction                     __sort                           aaa-bbb                          throw
__fetchurl                       __isInt                          __split                          aaa-ccc                          toString
__filter                         __isList                         __storeDir                       abort                            true
__filterSource                   __isString                       __storePath                      baseNameOf
```

Essentially completion here is same as if no input was entered (since linenoise completion callback breaks at dashes on our behalf.)

This is unfortunate (too many suggestions) but not incorrect.

## aaa-b

```
nix-repl> aaa-b
baseNameOf  builtins
```

This is just wrong :(.

# Linenoise, linenoise-ng, readline?

AFAICT this behavior is the same when using linenoise or linenoise-ng, but does not occur in readline-based `nix-repl`.

# Solution?

[...]

PR #2228, opened 2018-06-13
use editline instead of linenoise(-ng)

dtzWill:
I marked the commit as "WIP" but I don't remember
what problems or concerns I had with this.

Poking at it, things seem to work nicely?

Not sure if utf8 is supported, don't think this is needed
for anything we have in nixpkgs.

Maybe I was just unhappy about adding another dependency?


dtzWill:
Observations so far:
* Avoids the worst part of #1756 (although the fix I suggest there I've used daily and seems to work great, sorry for not pushing upstream)
* Doesn't auto-fill shared prefix of completion choices (although does complete when only 1 option available)
* control-c during completion seems to cause some sort of queued exit (??)
* Doesn't do anything silly if you try piping to nix repl (`:|nix repl`)-- this was an issue previously.

dtzWill:
Apparently editline is not already in upstream Nixpkgs so perhaps this is unwise? (apparently no other programs we package use it?)

dtzWill:
General dependency "health", I suppose.  If we are the only user it's ours to maintain/update. Sometimes there is a reason no one else uses a component, but I think we're safe here if it's what Debian uses.

Lassulus:
any progress here? The commit message still says WIP

dtzWill:
I don't remember what is "wip" about it, sorry (shame on me for not making a better note of this at at the time!).
I might have just been defensive since I grafted most of this, as I recall, from the "original" readline-based repl code and am not confident in the API and our usage of it personally.
I'll try using it as my daily-driver for a bit to see how that goes...
FWIW just pushed editline to 18.03 as well so it's available if we want it without too much trouble.

I feel we'd be better off using libedit but I think that'll take non-trivial effort since it uses a different API, but presumably nothing entirely too radically different :).

dtzWill:
FWIW I don't think working on this (either testing/improving what is started here, or exploring using libedit or something) requires much knowledge of anything Nix-internal, so it may be a good project for a wider audience of developers :).

dtzWill:
Not inserting a shared "prefix" among completions was making me sad, so put together a commit to provide that behavior.  Pushing here since I think it makes editline more usable but I'm okay if would rather merge without.

dtzWill:
rebased (not quite to master due to #2249) and replaced the prefix completion with a slightly better approach.

dtzWill:
Might act a bit differently when editing long lines that wrap, especially as interacting with terminal resizing?

KaiHa:
> I might have just been defensive since I grafted most of this, as I recall, from the "original" readline-based repl code and am not confident in the API and our usage of it personally.

[SNIP]

> I feel we'd be better off using libedit but I think that'll take non-trivial effort since it uses a different API, but presumably nothing entirely too radically different :).

Hi @dtzWill, may I ask what is the reason for not using _readline_? From the perspective of packaging Nix for Debian it would be a huge step forward, because _readline_ is already available as a package in Debian. 

_libedit_ (http://thrysoee.dk/editline/) would also be available in Debian. 

_editline_ (http://troglobit.com/projects/editline/) would be slightly more complicated, because it is not available in Debian (although there is a package by the name editline but it seems to be not related to the one you are using).

edolstra:
We used to use readline, but got rid of it because it doesn't depend on ncurses (c5f23f10a84f568874321c04984b1a14d2dce978).

KaiHa:
I have creates a [commit](https://github.com/KaiHa/nix/commit/731f4da3f426dc5788d86eed109e4a2a93ee59df) on [this branch](https://github.com/KaiHa/nix/commits/wip/readline-pr) that replaces editline by readline. As far as I can see this does not pull in ncurses as a dependency. TAB completion (with a list of all matches) is working, and it seems to work as expected.

Would it make sense to create a pull request from it or do you see other problems when using readline?

src/nix/repl.cc
```diff
@@ -173,9 +233,8 @@ void NixRepl::mainLoop(const std::vector<std::string> & files)
             printMsg(lvlError, format(error + "%1%%2%") % (settings.showTrace ? e.prefix() : "") % e.msg());
         }
 
-        // We handled the current input fully, so we should clear it
-        // and read brand new input.
-        linenoiseHistoryAdd(input.c_str());
+        // We handled the current input fully, so we should clear it and read brand new input.
+        // TODO: history_add ?
```
edolstra: I suppose this TODO still needs to be done to get proper history.

dtzWill:
On Mon, 29 Oct 2018 05:09:08 -0700, Eelco Dolstra <notifications@github.com> wrote:
> edolstra commented on this pull request.
> 
> 
> 
> > @@ -173,9 +233,8 @@ void NixRepl::mainLoop(const std::vector<std::string> & files)
>              printMsg(lvlError, format(error + "%1%%2%") % (settings.showTrace ? e.prefix() : "") % e.msg());
>          }
>  
> -        // We handled the current input fully, so we should clear it
> -        // and read brand new input.
> -        linenoiseHistoryAdd(input.c_str());
> +        // We handled the current input fully, so we should clear it and read brand new input.
> +        // TODO: history_add ?
> 
> I suppose this TODO still needs to be done to get proper history.

I believe editline will add to history automatically, or at least a
comment in the editline source suggests as much :).

Building to check presently..

Yep, history works here.  Can probably drop the TODO?

(unless the auto-history is undesired/wrong, but at least basic usage
works as I expect)

dtzWill:
Conflicts resolved (merge, not rebase eep): https://github.com/dtzWill/nix/tree/feature/editline-pr-merge 

Which also contains #2485 so it builds against 18.09 for testing.  Will tackle a rebase and drop the TODO in a few...

edolstra:
Thanks, merged!

Issue #1209, opened 2017-01-26
Don't modify dotfiles in $HOME without asking in the installer


hdgarrood:
I noticed that the installer modifies `~/.profile` without asking, which (IMO) is not good etiquette. For example, I prefer to keep my dotfiles under version control, and I've now added a bit of code to my dotfiles that adds Nix to the path when I want that to be the case. But now, whenever I install Nix on a new system, or even if I reinstall it on my current system, the installer will add the redundant extra line anyway.

In general I think that it's best to have people make any necessary changes to `~/.profile` manually, because then they're more aware of what has actually happened, and so it's less likely that the system will behave in a surprising way because of modifications to these dotfiles, and also it's more likely that they will be able to fix it if it does start behaving in a surprising way.

Would you consider changing this behaviour so that it instead just prints the line that you need to add to .profile and lets you do it yourself?

puffnfresh:
Reinstalling shouldn't add an extra line, the installer does a check to make sure the line isn't present before adding it.

hdgarrood:
I guess the installer didn't notice in my case because I changed it to use $HOME rather than the full path to my home directory (I want my dotfiles to work across machines regardless of my username).

hdgarrood:
See also #555 (although I'm suggesting Nix basically do the exact opposite of what that suggests).

issmirnov:
+1.

I would strongly appreciate a message at the end of the installation that explains where the various files live. I had to dig around on several issues before I learned about all the relevant details. Here's my suggested text to append to the install script, after `Try it! Open a new terminal, and type` and before `Thank you for using this installer. If you have any feedback, don't hesitate`. See https://github.com/NixOS/nix/blob/master/scripts/install-multi-user.sh#L255.

```
Important Information:
- Nix is installed to $NIX_ROOT
- Your user profile is at $HOME/.nix-profile
- The installer has updated $PROFILE_TARGETS with a call to load the nix daemon.
- If you are getting command not found errors, make sure that your '$PATH' contains $HOME/.nix-profile/bin

```

tomberek:
Closing as stale. Re-open if needed.

Issue #2175, opened 2018-05-22
nix upgrade-nix borks user profiles containing nix


dtzWill:
See https://github.com/NixOS/nix/issues/2070#issuecomment-390772945 .

Hopefully can be addressed with a bit more sanity checking, some of which may benefit other situations (including user invoking commands "incorrectly").

Ideas:

Check "profile" paths (following may need tweaking):
  * Must be a symlink (can be directory today, haha)
  * Always point to a "x-N-link" in the same directory.
  * x-N-link's directly point to store path
  * (?) Should be gc roots, warn if not

On a related note, it may be worth clarifying that the link "~/.nix-profile" is in fact not at all a "profile" (of the kind mentioned in docs, accepted as arguments by commands) but rather a link to a profile.

Anyway the main issue is that upgrade-nix's autodection of profile can mistakenly pick up a "profile link"... which appears to work but silently causes problems such as profile generations no longer being gc roots, as well as other strangeness re:missing generations and whatnot.

LMK if solving the "larger" issue as described above should become a separate issue... hopefully we can fix these together!

PR #2247, opened 2018-06-22
Add helpful pointers in post install script.

issmirnov:
There are a ton of issues with new users struggling on basic setup. Updated install script final blurb to point people to relevant folders.

Inspired by #1209

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -253,6 +253,12 @@ Try it! Open a new terminal, and type:
 
   $ nix-shell -p nix-info --run "nix-info -m"
 
+Important Information:
+- Nix is installed to $NIX_ROOT
+- Your user profile is at $HOME/.nix-profile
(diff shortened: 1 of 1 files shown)


dtzWill:
FWIW I'm not sure we want to tell folks their profile is $HOME/.nix-profile -- see #2175 for some discussion how this is actually a "link to the profile" which matters because confusing the two causes Nix commands to behave in somewhat subtly incorrect ways causing later commands to fail in ways that are hard to diagnose even for those not new.

Misc suggestions how to address this: find vocabulary/language to describe these things, improve tool's handling of confusion between the two, and/or avoid telling users about the details of where their "profile" is since most usage is done w/o pointing tools at a particular profile...?

Otherwise looks good, although I too would like @grahamc to take a look :).

issmirnov:
Thanks, happy to change any phrasing as needed. @grahamc awaiting your word :)

PR #2250, opened 2018-06-23
repl: use `nix build` for building instead of `nix-store -r`

dtzWill:
progress bar!

----

Recommended to test using/on-top-of fix from #2249.

Seems to work for me? :D


dtzWill:
Using "nix 2.0" command for this operation but not the others is slightly "inconsistent" but I don't think in a confusing or problematic way.  The right answer is to get nix-2.0 commands that perform the other operations (mostly nix-env)! :P.

coretemp:
If Nix 1.X really is history, then :+1: , otherwise perhaps adding a fallback would be a good idea.

PR #2258, opened 2018-06-27
Make installer allow preexisting by default

matthewbauer:
This seems to be working okay for me. What is currently holding this up?

/cc @grahamc

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -399,7 +399,7 @@ setup_report() {
     row "     Build Users" "$NIX_USER_COUNT"
     row "  Build Group ID" "$NIX_BUILD_GROUP_ID"
     row "Build Group Name" "$NIX_BUILD_GROUP_NAME"
-    if [ "${ALLOW_PREEXISTING_INSTALLATION:-}" != "" ]; then
+    if [ "${ALLOW_PREEXISTING_INSTALLATION:-1}" != "" ]; then
         row "Preexisting Install" "Allowed"
     fi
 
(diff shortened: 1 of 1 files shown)


shlevy:
I think the main problem is if the preexisting installation is somehow broken, continuing with an automated process is likely to make fixing it even harder. If you have a working preexisting solution, why aren't you just upgrading Nix with Nix?

matthewbauer:
Well there are lots of times when the Nix commands can be broken & there is no way to upgrade. This will happen to users who upgrade to macOS Mojave with old Nix versions.

edolstra:
We just need to make sure that all actions done by the installer are idempotent.

grahamc:
Well, the actions need to be more than idempotent -- the installer needs to handle every possible installation which already exists and know the right thing to do: fixing permissions to make it multi-user compatible, or fall back to the single user, or fix permissions on a badly configured multi-user setup, or fail ... etc. I much prefer the safety as is, and I think that expanding the tool to support handling weird installs potentially makes this a _much_ more complicated process. This complexity is why I don't support it now.

grahamc:
Note, I do agree we should allow the script to be re-run in happy paths, however it is all the scary dark paths that I am worried about.

shlevy:
@matthewbauer Yes, it's in exactly those cases where automatically going ahead is something to worry about though.

matthewbauer:
I don't see how it's going to mess up your system though. The only critical part is installing the daemon - everything else shouldn't break anything. At least on macOS, evan a misconfigured daemon won't break anything AFAICT.

matthewbauer:
@grahamc If the new installer will not support upgrades, could you update the 3rd bullet in default.nix:

https://github.com/NixOS/nixpkgs/blob/master/default.nix

I am not sure what to say there, exactly. But it definitely will need to be updated in time for when we want to move minver.nix to 2.0.

Issue #2050, opened 2018-04-06
`nix search` should have flag to only search package names


lilyball:
The new `nix search` command is great, but there's often a lot of noise coming from packages that include the search string in their description. For example, `nix search git` shows a lot of irrelevant stuff. It would be great if there was a flag I could provide to only search package names.

hraban:
vote to reopen: this is still relevant. And there seems to be a patch floating around somewhere.

lilyball:
I thought the stale bot didn't actually close issues? The "More info" link does not say anything about the possibility of closing issues. Did someone muck with this? Stale bots that close issues automatically are a terrible idea.

nandoflorestan:
This is essential functionality that is missing!

PR #2266, opened 2018-07-03
nix search: allow `-A` for attribute-only searches

Ma27:
* Added optional flag `-A` for this purpose (see #2050)
* ef0e228 fixes a regression when using `nix search` with `--json`

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -185,14 +185,14 @@ struct CmdSearch : SourceExprCommand, MixJSON
                             auto name = hilite(parsed.name, nameMatch, "\e[0;2m")
                                 + std::string(parsed.fullName, parsed.name.length());
 
-                            if (description.empty()) {
-                                description = "\e[3mNo description\e[23m";
-                            }
+                            auto display_description =
+                                description.empty() ? "\e[3mNo description\e[23m" : description;
+
                             results[attrPath] = fmt(
                                 "* %s (%s)\n  %s\n",
                                 wrap("\e[0;1m", hilite(attrPath, attrPathMatch, "\e[0;1m")),
                                 wrap("\e[0;2m", hilite(name, nameMatch, "\e[0;2m")),
-                                hilite(description, descriptionMatch, ANSI_NORMAL));
+                                hilite(display_description, descriptionMatch, ANSI_NORMAL));
                         }
                     }
 
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -184,6 +184,10 @@ struct CmdSearch : SourceExprCommand, MixJSON
                         } else {
                             auto name = hilite(parsed.name, nameMatch, "\e[0;2m")
                                 + std::string(parsed.fullName, parsed.name.length());
+
+                            if (description.empty()) {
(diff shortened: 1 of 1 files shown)


dtzWill:
Unrelated but consider rebasing onto latest Nix master -- in particular without cb9ef858529348df8dded2f0bface7d9612cafc6 this branch fails tests (which means it doesn't build without disabling them, making it harder to test the resulting binaries and makes it less clear that tests would pass "if" they passed on commit this was based on...).

Anyway no big deal just thought I'd mention it after running into such failures trying to give this a go :).

grahamc:
Maybe -A would
Be more consistent with existing tooling?

Ma27:
@dtzWill good point, rebased onto master
@grahamc I originally had `nix-env -qaP` in mind, but I didn't know what the actual arguments were actually used for. After actually reading the man pages I think as well that `-A` is the better option  :sweat_smile:

dtzWill:
Consider adding this fix: https://github.com/dtzWill/nix/commit/80d8ca6d7b41a8eacbcf73653fa5e92dff490674

To fix reading/writing cache in presence of "No description" packages.

Ma27:
seems legit, is it okay if I just cherry-pick it onto my branch?

dtzWill:
> seems legit, is it okay if I just cherry-pick it onto my branch?

(of course! Thanks! <3)

Ma27:
@edolstra any chance to get this merged?:)

Ma27:
rebased onto latest master and resolved conflicts.

Ma27:
anything missing here?

Ma27:
@grahamc anything missing or any chance to get this merged?

lilyball:
I'm not familiar with the C++ sources but I took a look at this PR and it looks pretty good to me.

lilyball:
Commit message and PR title reference `-a` but the implementation uses `-A`.

src/nix/search.cc
```diff
@@ -263,11 +282,13 @@ struct CmdSearch : SourceExprCommand, MixJSON
...
-        if (results.size() == 0)
-            throw Error("no results for the given search term(s)!");
+        if (!json) {
+            if (results.size() == 0)
+                throw Error("no results for the given search term(s)!");
 
-        RunPager pager;
-        for (auto el : results) std::cout << el.second << "\n";
+            RunPager pager;
+            for (auto el : results) std::cout << el.second << "\n";
```
lilyball: Don't we need to run this line when `json == true` as well?

Evils-Devils: when using json, the user can be expected to understand what an empty result means

lilyball: After taking another look at the code, it appears the real answer is "when `json == true`, `results` is always empty".

Ma27:
@edolstra anything missing here to get this merged? :)

Ma27:
Rebased onto latest master :)

Ma27:
Will close this now due to inactivity. It has been stated that `nix search` will go through a lot of changes for `flakes`, so if anyone is interested, this can be brought up later on.

PR #2268, opened 2018-07-03
tests/search.sh: minor fix to unbreak tests after search ux merge

dtzWill:


dtzWill:
(as seen on Hydra or anywhere someone tries to build Nix from master since: https://hydra.nixos.org/eval/1466850 )

Ma27:
ouch, thanks a lot for fixing it! :)

dtzWill:
No worries!


Unrelated but this is the sort of thing it'd be nice to have Travis or hydra-based bots/badges tracking :).

And tests being overly sensitive to output formatting (that's not a fixed format) is unfortunate anyway, so of all the tests to "fail" / need adjusting... this one's just asking t,o be broken by perfectly reasonable changes :).

It's possible I wrote it :P.

PR #2269, opened 2018-07-03
Manual: Updates install notes for one-liner

samueldr:
The previous instructions did not account for the fact that:
  * macOS now defaults to *multi-user*.
  * it is possible to override installation type.

Furthermore, older commented-out deb/rpm instructions were removed as they are code-rotting.

I think this should be safe to also merge in maintenance (hoping this would also reach the website).

* * *

@grahamc please review for the details about the single/multi user installation.

* * *

~~There is one issue which I'm still looking into: will the `<segmentedlist>` cause issues on the website. I will update this PR once determined. If it does, I will either change to whatever a docbook-knowing contributor suggests, and/or fix the CSS for the website manual accordingly.~~ See my following comment.

### DIFF
--- a/doc/manual/installation/installing-binary.xml
+++ b/doc/manual/installation/installing-binary.xml
@@ -6,11 +6,47 @@
 
 <title>Installing a Binary Distribution</title>
 
+<section>
+<title>Installing using the one-liner</title>
+
 <para>If you are using Linux or macOS, the easiest way to install
 Nix is to run the following command:
 
 <screen>
-$ bash &lt;(curl https://nixos.org/nix/install)
+$ curl https://nixos.org/nix/install | sh
+</screen>
+
+This will perform a default installation of Nix for your platform.
+
+<segmentedlist>
+	<?dbhtml list-presentation="table"?>
+	<?dbfo   list-presentation="table"?>
+	<title>System differences</title>
+	<segtitle>System</segtitle>
+	<segtitle>Default installation</segtitle>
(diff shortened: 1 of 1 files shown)


grahamc:
From the outset lgtm, I'll test it against the website soon.

samueldr:
image

Looks fine, the table outputs as expected from a bootstrapified doc.

For the record, I generated it this way:

```
diff --git a/Makefile b/Makefile
index 5e83ef9..40a560d 100644
--- a/Makefile
+++ b/Makefile
@@ -52,8 +52,9 @@ $(NIX_PILLS_MANUAL_IN):
 ### Prettify the Nix manual.
-NIX_MANUAL_IN = nix/manual-raw
+#NIX_MANUAL_IN = nix/manual-raw
 NIX_MANUAL_OUT = nix/manual
+NIX_MANUAL_IN = /Users/samuel/tmp/nix/doc/manual
 ifneq ($(wildcard $(NIX_MANUAL_IN)),)

```

It may be needed to `rm -rf nix/manual` in your nixos-homepage checkout if it doesn't want to generate.

grahamc:
Nice! LGTM! +1 on backport.

domenkozar:
This needs rebasing.

abathur:
I think this can be closed. 

A quick spot-check suggests:
- most of these changes are represented in some form in the current document
- I see a few factual things that have changed
- It would need to track docbook -> markdown

PR #2273, opened 2018-07-05
[wip] lib.concatMap and lib.mapAttrs to be builtins

### LINKED ISSUE #1776: lib.concatMap and lib.mapAttrs to be builtins
```concatMap``` and ```mapAttrs``` are the most often called library functions.
They could be straightforward implemented in C++ to speed up execution and to avoid generating temporary thunks.
With my nix expressions it saves about 20% of memory.

### PR DESCRIPTION
**Pull request by volth**
to reduce memory consumption when evaluating big ```nixpkgs```

Fixes #1776

### CODE REVIEW

File: src/libexpr/primops.cc [hunk 1]
```diff
@@ -1356,6 +1356,24 @@ static void prim_functionArgs(EvalState & state, const Pos & pos, Value * * args
 }
 
 
+/* Apply a function to every element of a list. */
```
edolstra: Should be 'of an attribute set'.

src/libexpr/primops.cc
```diff
@@ -1627,6 +1645,28 @@ static void prim_partition(EvalState & state, const Pos & pos, Value * * args, V
...
+/* concatMap = f: list: concatLists (map f list); */
+/* C++-version is to avoid allocating `mkApp', call `f' eagerly */
+static void prim_concatMap(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    state.forceFunction(*args[0], pos);
+    state.forceList(*args[1], pos);
+    auto len = args[1]->listSize();
+
+    Value vList;
+    state.mkList(vList, len);
```
edolstra: Can't the allocation of the list and the values be avoided by using a local array of values, i.e. `Value vs[len]`? The values can't escape so it should be fine to put them on the stack.

volth: I was afraid of a possible stack overflow

edolstra: Alternatively you could use something like `std::vector<Value, gc_allocator<Value>> list(len)` to allocate all temporary values at once.

dtzWill: I am concerned about stack-allocating this, since the length scales with number of items in list mapped over, which seems like something that could easily be thousands if ever used as part of something that maps over all packages or something.

Quick (probably incomplete) grep suggests only other variable-length stack allocation of Values is based on function arity which is by comparison much safer -- generally rather small and isn't usually extended "programatically" (such that we'd end up with functions with thousands-of-arguments).

The vector approach seems better-- I'm guessing the problem with `mkList` is it's rather expensive by comparison (I'd expect so but haven't actually tried measuring it so wanted to check)?

edolstra:
Thanks!

volth:
please undo, there are bugs

volth:
or add this fix https://github.com/volth/nix/commit/b783a00ca14b59a7971151726100c8f926f46cf4

PR #2278, opened 2018-07-05
column internal representation of attrsets

volth:
Internally attrsets are arrays like
```
[ {key:key1,value: value1, pos:pos1}
, {key:key2, value:value2, pos:pos2}
, ...
]
```

This PR changes the attrset internals to
```
{ keys:[key1, key2, ...]
, values:[value1, value2, ...]
, positions:[pos1, pos2, ...]
}
```

and deduplicate ```keys``` and ```pos``` arrays

1. it reduces memory footprint by approx 18% 
2. it speeds up comparision of attrsets (```keys``` arrays can be compared by pointers, effectively comparing arrays size and all its keys)
3. it makes convenient caching of functions performed on attrset's keys. Currently only result of lexicographic sort is memoized. Results of ```attrNames``` or ```intersectAttrs``` might be memoized in the future.

The code passes all the tests and builds my closures, but I have not tested all the Nix programs (for example I am not using ```nix-env```, etc).


volth:
| Before | After |
| --- | --- |
| evaluation statistics:                            | evaluation statistics:                             | 
|   time elapsed: 15.9774                           |   time elapsed: 15.2598                            | 
|   size of a value: 24                             |   size of a value: 24                              | 
|   size of an attr: 24                             |                                                    | 
|   environments allocated count: 17032924          |   environments allocated count: 17032924           | 
|   environments allocated bytes: 432380864         |   environments allocated bytes: 432380864          | 
|   list elements count: 8386238                    |   list elements count: 8386238                     | 
|   list elements bytes: 67089904                   |   list elements bytes: 67089904                    | 
|   list concatenations: 94157                      |   list concatenations: 94157                       | 
|   values allocated count: 28922901                |   values allocated count: 28922902                 | 
|   values allocated bytes: 694149624               |   values allocated bytes: 694149648                | 
|   sets allocated: 4660329 (579292584 bytes)       |   sets allocated: 4501875 (263913616 bytes)        | 
|                                                   |   sets name lists: 10700 (5203424 bytes)           | 
|                                                   |   sets pos lists: 31392 (13465416 bytes)           | 
|   right-biased unions: 418406                     |   right-biased unions: 418406                      | 
|   values copied in right-biased unions: 11143782  |   values copied in right-biased unions: 11143782   | 
|   symbols in symbol table: 86551                  |   symbols in symbol table: 86552                   | 
|   size of symbol table: 2922673                   |   size of symbol table: 2922603                    | 
|   number of thunks: 17677859                      |   number of thunks: 17677859                       | 
|   number of thunks avoided: 12235602              |   number of thunks avoided: 12235602               | 
|   number of attr lookups: 12113297                |   number of attr lookups: 12113297                 | 
|   number of primop calls: 6346625                 |   number of primop calls: 6346625                  | 
|   number of function calls: 15802804              |   number of function calls: 15802804               | 
|   total allocations: 1772912976 bytes             |   total allocations: 1457534032 bytes              | 
|   current Boehm heap size: 1393033216 bytes       |   current Boehm heap size: 1174929408 bytes        | 
|   total Boehm heap allocations: 2212184976 bytes  |   total Boehm heap allocations: 1852121744 bytes   |

volth:
```
nix-instantiate --eval --strict --show-trace -vv \
 '<testpkgs/maintainers/scripts/eval-release.nix>' \
 -I testpkgs=https://github.com/NixOS/nixpkgs/archive/master.tar.gz
```

| Before | After |
| --- | --- |
|  time elapsed: 151.372                            | time elapsed: 147.24                            |
|  size of a value: 24                              | size of a value: 24                             |
|  size of an attr: 24                              |                                                 |
|  environments allocated count: 74562291           | environments allocated count: 74562291          |
|  environments allocated bytes: 2092789312         | environments allocated bytes: 2092789312        |
|  list elements count: 91337130                    | list elements count: 91337130                   |
|  list elements bytes: 730697040                   | list elements bytes: 730697040                  |
|  list concatenations: 4888232                     | list concatenations: 4888232                    |
|  values allocated count: 158792462                | values allocated count: 158783487               |
|  values allocated bytes: 3811019088               | values allocated bytes: 3810803688              |
|  sets allocated: 15194579 (**4681609576** bytes)      | sets allocated: 15193824 (**1896861480** bytes)     |
|                                                   | sets name lists: 87037 (32232984 bytes)         |
|                                                   | sets pos lists: 352448 (138549424 bytes)        |
|  right-biased unions: 6836346                     | right-biased unions: 6836346                    |
|  values copied in right-biased unions: 143089131  | values copied in right-biased unions: 143089131 |
|  symbols in symbol table: 294062                  | symbols in symbol table: 294062                 |
|  size of symbol table: 12006205                   | size of symbol table: 12006205                  |
|  number of thunks: 109039595                      | number of thunks: 109039553                     |
|  number of thunks avoided: 107085312              | number of thunks avoided: 107085354             |
|  number of attr lookups: 41655745                 | number of attr lookups: 41655745                |
|  number of primop calls: 52966076                 | number of primop calls: 52966076                |
|  number of function calls: 65860454               | number of function calls: 65860454              |
|  total allocations: **11316115016** bytes             | total allocations: **8531151520** bytes             |
|  current Boehm heap size: **8449425408** bytes        | current Boehm heap size: **5423235072** bytes       |
|  total Boehm heap allocations: 14603866816 bytes  | total Boehm heap allocations: 11026657968 bytes |

volth:
```
nix-instantiate --eval --strict --show-trace -vv \
 '<testpkgs/maintainers/scripts/eval-release.nix>' \
 -I testpkgs=https://github.com/NixOS/nixpkgs/archive/master.tar.gz
```

| Before | After |
| --- | --- |
|  time elapsed: **151.372**                                | time elapsed: **132.661**                           |
|  size of a value: 24                                      | size of a value: 24                                 |
|  size of an attr: 24                                      |                                                     |
|  environments allocated count: 74562291                   | environments allocated count: 74566834              |
|  environments allocated bytes: 2092789312                 | environments allocated bytes: 2092915032            |
|  list elements count: 91337130                            | list elements count: 91339637                       |
|  list elements bytes: 730697040                           | list elements bytes: 730717096                      |
|  list concatenations: 4888232                             | list concatenations: 4888506                        |
|  values allocated count: 158792462                        | values allocated count: 158794013                   |
|  values allocated bytes: 3811019088                       | values allocated bytes: 3811056312                  |
|  sets allocated: 15194579 (**4,681,609,576** bytes)       | sets allocated: 15194795 (**1,897,084,928** bytes)  |
|                                                           | sets name lists: 87053 (32249904 bytes)             |
|                                                           | sets pos lists: 352466 (138577616 bytes)            |
|  right-biased unions: 6836346                             | right-biased unions: 6836708                        |
|  values copied in right-biased unions: 143089131          | values copied in right-biased unions: 143106250     |
|  symbols in symbol table: 294062                          | symbols in symbol table: 294719                     |
|  size of symbol table: 12006205                           | size of symbol table: 12026234                      |
|  number of thunks: 109039595                              | number of thunks: 109047427                         |
|  number of thunks avoided: 107085312                      | number of thunks avoided: 107092240                 |
|  number of attr lookups: 41655745                         | number of attr lookups: 41658468                    |
|  number of primop calls: 52966076                         | number of primop calls: 52969799                    |
|  number of function calls: 65860454                       | number of function calls: 65864335                  |
|  total allocations: **11,316,115,016** bytes              | total allocations: **8,531,773,368** bytes          |
|  current Boehm heap size: **8,449,425,408** bytes         | current Boehm heap size: **5,414,846,464** bytes    |
|  total Boehm heap allocations: 14603866816 bytes          | total Boehm heap allocations: 11027366944 bytes     |

src/libexpr/attr-set.cc
```diff
@@ -2,51 +2,83 @@
...
-/* Create a new attribute named 'name' on an existing attribute set stored
-   in 'vAttrs' and return the newly allocated Value which is associated with
-   this attribute. */
-Value * EvalState::allocAttr(Value & vAttrs, const Symbol & name)
-{
-    Value * v = allocValue();
-    vAttrs.attrs->push_back(Attr(name, v));
-    return v;
+// own sort function because std::sort has no custom `swap' function :(
+void BindingsBuilder::quicksort(iterator a, iterator b) {
```
danbst: wow. Stock `std::sort` should be more efficient, so I'm surprised there is performance gain in the end.

volth: Copy-pasting code from ```std::sort``` could improve performance even more.

volth:
Any feedback? @shlevy ? @edolstra ?

src/libexpr/attr-path.cc
```diff
@@ -68,10 +68,10 @@ Value * findAlongAttrPath(EvalState & state, const string & attrPath,
             if (attr.empty())
                 throw Error(format("empty attribute name in selection path '%1%'") % attrPath);
 
-            Bindings::iterator a = v->attrs->find(state.symbols.create(attr));
-            if (a == v->attrs->end())
+            Bindings::find_iterator a = v->attrs->find(state.symbols.create(attr));
```
shlevy: A lot of this code predates C++11, but if we're changing this let's just go to `auto` here.

src/libexpr/attr-set.cc
```diff
@@ -2,51 +2,83 @@
 #include "eval-inline.hh"
 
 #include <algorithm>
-
+#include <string.h>
 
 namespace nix {
 
+std::unordered_set<Bindings::Names, Bindings::NamesHasher>       * Bindings::pnamesTable = new std::unordered_set<Bindings::Names, Bindings::NamesHasher>      (0x4000, Bindings::NamesHasher());
```
shlevy: Should we give a name to this type? Also, `auto` on the LHS at least.

shlevy: (And a bunch more places for `auto` throughout the PR)

src/libexpr/attr-set.cc
```diff
@@ -2,51 +2,83 @@
...
 
-/* Create a new attribute named 'name' on an existing attribute set stored
-   in 'vAttrs' and return the newly allocated Value which is associated with
-   this attribute. */
-Value * EvalState::allocAttr(Value & vAttrs, const Symbol & name)
-{
-    Value * v = allocValue();
-    vAttrs.attrs->push_back(Attr(name, v));
-    return v;
+// own sort function because std::sort has no custom `swap' function :(
```
shlevy: https://en.cppreference.com/w/cpp/named_req/Swappable ?

shlevy: (yes, ADL is horrible, but such is life in the STL :tongue: )

src/libexpr/attr-set.hh
```diff
@@ -7,88 +7,193 @@
...
+
+        ~Names() {
+            assert(_lexicographicOrder == nullptr); // interned `Names' are never destructed
+        }                                           // `Names' in BindingsBuilder are not complete and should not use `lexicographicOrder()'
+
+        // exclude mutables
+        bool operator==(const Names & that) const { return symOrder == that.symOrder; }
+
+        const unsigned int * lexicographicOrder() const {
+            if (_lexicographicOrder == nullptr) {
```
shlevy: This isn't a blocker, since the evaluator is so heavily single-threaded everywhere anyway, but const member functions mutating data in a non-threadsafe way is a concern.

shlevy:
Also, if you're not opposed to crafting history, if you can break this up into logical commits it may make this easier to review overall.

src/libexpr/attr-set.hh
volth: This is thread-safe, at the worst thread race ```std::sort``` will be called twice producing the same result.
But a lot of other code (for example ```std::vectors::push_back()``` is not thread-safe)

src/libexpr/attr-set.cc
volth: Yes, I just prefer to avoid ```auto``` during development to avoid surprises (```auto``` could deduce another type, not what I think of when read the code)
But it might have sense to shrink well-tested code.

edolstra:
My main concern is that this adds a *significant* amount of complexity to the evaluator. I'm not sure whether that's worth the memory savings. An 18% reduction is nice, but probably only represents a few weeks worth of Nixpkgs/NixOS memory bloat growth...

src/libexpr/attr-set.hh
```diff
@@ -7,88 +7,193 @@
...
+                _lexicographicOrder = new unsigned int[symOrder.size()];
```
edolstra: Shouldn't this be allocated using Boehm?

src/libexpr/attr-set.cc
volth: Hm, it seems not what needed here.
This function sorts the name in ```symOrder``` array but also swaps elements in two other arrays so they are "sorted by names"

src/libexpr/attr-set.hh
edolstra: Since there actually is a proof-of-concept concurrent evaluator (https://github.com/edolstra/nix/commit/3ad4128f67b16a797e074576c0458d118cf443f6), adding new sources of races should be avoided.

I don't think this code is thread-safe, since e.g. one thread could see that `_lexicographicOrder` exists and assume that it's in sorted order, while another thread is still sorting it. This could be avoided by assigning the new array to `_lexicographicOrder` only after it has been sorted. (Also maybe mark it as std::atomic, though that probably doesn't matter on sane architectures.)

src/libexpr/attr-set.hh
volth: This will never be GC'ed.
Attrset names array are stored in a singleton hash-table (akin to the symbol table) and this ```_lexicographicOrder``` is stored inside the attrset names array to avoid sorting each time when lexicographic order is required.

volth:
> My main concern is that this adds a significant amount of complexity to the evaluator. I'm not sure whether that's worth the memory savings. An 18% reduction is nice, but probably only represents a few weeks worth of Nixpkgs/NixOS memory bloat growth...

I'd say it is a relatively low-hanging fruit for optimization with minimal changes in the evaluator (no new value types, no AST transformations, ...).
But, yes, it does not solve the problem of Nix requiring 16Gb RAM on a workstation to work with, and soon will not fit into laptops as their RAM limits grow slower than nixpkgs bloats.

volth:
This also opens a room for further optimizations:

Instead of **list of attrs** we get **objects with RTTI** ("4,501,875 allocated attrsets share 10,700 keys arrays" can be read as "there are 4,501,875 objects of 10,700 types") and very cheap ```classOf``` (one pointer comparison).

So, the operations on some most popular "objecs" (like ```{ name = ...; value = ... }```)) might be specialized and performed faster than on generic attrsets

volth:
> Can you put together a representative performance benchmark that we can make part of CI, to catch regressions at least but also to demonstrate in the code and repo history the benefits of this change?

There is already https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/metrics.nix which looks like it was made exactly for tracking Nix performance (or nixpkgs bloat rate) on CI.

Not sure if it is actually used, I have never seen its reports

edolstra:
@volth It allows Hydra to produce charts like https://hydra.nixos.org/job/nixpkgs/trunk/metrics/metric/nix-env.qa.allocations

src/libexpr/attr-set.hh
dezgeg: aarch64 will definitely need memory barriers here (which I guess default assignment and derefence operations of std::atomic do).

src/libexpr/attr-set.cc
dezgeg: It can be done by implementing an iterator which returns a proxy object whose swap() and/or assignment operators affect the two arrays. But the code ends up quite ugly: https://gist.github.com/dezgeg/30c82c82a0c60f69ab9ed104bfc56ce3

volth:
https://github.com/NixOS/nixpkgs/issues/38635

volth:
This ```resources.machines``` - https://twitter.com/nixflower/status/1055831102441562112 - is exactly the way which leaded me to enormous memory consunption as the number of simulteniously evaluated closures grows.

PR #2275, opened 2018-07-05
[bugfix] lib.concatMap and lib.mapAttrs to be builtins

volth:
#2273 has been accepted too quickly even before 2 bugs have been found
This is the fix

PR #2281, opened 2018-07-06
[bugfix] prim_mapAttrs: `f' must be evaluated lazily to avoid infinite recursion

volth:
Fix for a regression found by @dtzWill https://github.com/NixOS/nix/pull/2275#issuecomment-403136364

The problem was in Nixpkgs's  ```idrisPackages``` which expects that the ```mapAttr```'s function argument won't be evaluated eagerly.
It is true for ```lib.mapAttr``` written in Nix, but ```builtins.mapAttrs``` did the eager evaluation of its arguments like all other builtins in ```primop.c```

The actual fix is only [this line ](https://github.com/NixOS/nix/pull/2281/files#diff-169b8f6a99e8c50e9e5468f9f144015eL1362), all the rest is just to make evaluation of nested ```App``` possible

----

The fix also backported to https://github.com/NixOS/nix/pull/2278

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1076,6 +1076,8 @@ void EvalState::callPrimOp(Value & fun, Value & arg, Value & v, const Pos & pos)
 
 void EvalState::callFunction(Value & fun, Value & arg, Value & v, const Pos & pos)
 {
+    forceValue(fun, pos);
+
     if (fun.type == tPrimOp || fun.type == tPrimOpApp) {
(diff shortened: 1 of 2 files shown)


copumpkin:
👍 from me. In general I think it's good for these HOF primops to be lazy if the rest of the language is. I have written a few (unpublished) chunks of Nix that "tie the knot" using things like `mapAttrs`.

volth:
@dtzWill @edolstra It is a regression fix, current Nix ```master``` cannot evaluate ```nixpkgs``` without it

PR #2282, opened 2018-07-07
fix build without `--enable-gc'

volth:


dtzWill:
Since this is (currently) the "default", this is important to fix and also addresses deb and rpm builds as briefly mentioned in the thread mentioned above.

edolstra:
I've removed `printStats2()` instead, since it was only a temporary function for some GC debugging.

PR #2312, opened 2018-07-27
search: include version for matches too!

dtzWill:
Since the matching isn't done against fullName
the match suffix doesn't include version
for "good" derivations that are parsed properly.

Searching without a match string always shows
the full name, with this the full name
is always shown.

Really helps when you just wanna see what version
is available currently! :)

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -173,10 +173,12 @@ struct CmdSearch : SourceExprCommand, MixJSON
                             jsonElem.attr("description", description);
 
                         } else {
+                            auto name = hilite(parsed.name, nameMatch, "\e[0;2m")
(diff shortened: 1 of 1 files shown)


edolstra:
I don't understand what this does. What is the output after this change?

dtzWill:
Currently if a search matches the attribute name, the version portion of the attribute name is not shown.

This is most easily seen by looking at the output of "nix search".  The portion in parenthesis contains just the name not the version as well--indeed the version will only be present for derivations that `parseDrvName` can't handle.

dtzWill:
After this change the portion in parens contains the full name--including version-- which is what is currently shown when for entries that don't match via attribute name (try `nix search shell` for a bit of both).

PR #2319, opened 2018-08-01
Document s3 substitutions

graham-at-target:
Ideally this will be backported to 2.0.x.

Ideally we could get an authoritative, minimal list of privileges required but it seems most people testing grant `s3:*`.

cc @shlevy

### DIFF
--- a/doc/manual/packages/s3-substituter.xml
+++ b/doc/manual/packages/s3-substituter.xml
@@ -1,0 +1,142 @@
+<?xml version="1.0" encoding="utf-8"?>
+<section xmlns="http://docbook.org/ns/docbook"
+         xmlns:xlink="http://www.w3.org/1999/xlink"
+         xmlns:xi="http://www.w3.org/2001/XInclude"
+         version="5.0"
+         xml:id="ssec-s3-substituter">
+
+<title>Serving a Nix store via AWS S3 or S3-compatible Service</title>
+
+<para>Nix has built-in support for storing and fetching store paths
+from Amazon S3 and S3 compatible services. This uses the same
+<emphasis>binary</emphasis> cache mechanism that Nix usually uses to
+fetch prebuilt binaries from <uri>cache.nixos.org</uri>.</para>
+
+<para>In this example we will use the bucket named
+<literal>example-bucket</literal>.</para>
+
+<section xml:id="ssec-s3-substituter-anonymous-reads">
+  <title>Anonymous Reads to your S3-compatible binary cache</title>
+
+  <para>If your binary cache is publicly accessible and does not
+  require authentication, the simplest and easiest way to use Nix with
+  your S3 compatible binary cache is to use the HTTP URL for that
+  cache.</para>
+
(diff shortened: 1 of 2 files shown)


doc/manual/packages/s3-substituter.xml
```diff
@@ -0,0 +1,142 @@
...
+  Additionally, the following limitations are in place for
+  authenticated reads:</para>
+
+  <itemizedlist>
+    <listitem><para>The bucket must actually be hosted by Amazon S3 and
+    <emphasis>not</emphasis> an S3 compatible
+    service.</para></listitem>
+
+    <listitem><para>The bucket must be within the
+    <literal>us-east-1</literal> region.</para></listitem>
```
copumpkin: I suppose this is why it doesn't need `s3:GetBucketLocation` on reads? Seems like a problem 😦

doc/manual/packages/s3-substituter.xml
```diff
@@ -0,0 +1,142 @@
...
+  <para>Your bucket will need a bucket policy allowing the desired
+  users to perform the <literal>s3:GetObject</literal> action on all
+  objects in the bucket.</para>
+</section>
+
+
+<section xml:id="ssec-s3-substituter-authenticated-writes">
+  <title>Authenticated Writes to your S3-compatible binary cache</title>
+
+  <para>Nix support fully supports writing to Amazon S3 and S3
```
copumpkin: Why is the behavior so different between reads and writes?

graham-at-target: In short...

Download code: https://github.com/NixOS/nix/blob/45bcf5416a0ce53361fd37c6b27ba4ef6a34ce96/src/libstore/download.cc#L600-L606

Upload code: https://github.com/NixOS/nix/blob/45bcf5416a0ce53361fd37c6b27ba4ef6a34ce96/src/libstore/s3-binary-cache-store.cc#L189

The Upload code does a fancier arg parsing, which happens in a separate function (https://github.com/NixOS/nix/blob/45bcf5416a0ce53361fd37c6b27ba4ef6a34ce96/src/libstore/store-api.cc#L838-L851) and my C++-foo isn't strong enough with the time I have to rectify it :(.

doc/manual/packages/s3-substituter.xml
```diff
@@ -0,0 +1,142 @@
...
+<section xml:id="ssec-s3-substituter-anonymous-reads">
+  <title>Anonymous Reads to your S3-compatible binary cache</title>
+
+  <para>If your binary cache is publicly accessible and does not
+  require authentication, the simplest and easiest way to use Nix with
+  your S3 compatible binary cache is to use the HTTP URL for that
+  cache.</para>
+
+  <para>For AWS S3 the binary cache URL for example bucket will be
+  exactly <uri>https://example-bucket.s3.amazonaws.com</uri>. For S3
```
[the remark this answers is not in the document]
AmineChikhaoui: oh good to know :)

PR #2320, opened 2018-08-01
Support specifying custom S3 endpoints for non-Amazon S3-Compatible APIs

graham-at-target:
Based upon #2319 for the docs fix. See just 
https://github.com/NixOS/nix/commit/2951cef5eb6de8f88131d047ea8b60de7f7a139d for the meat of this PR.

cc @shlevy


graham-at-target:
Ok, this is rebased after the merge of the docs section and now only contains the endpoint code.

graham-at-target:
The thing that makes this cool is we could now do some automated testing here with minio or other light-weight S3-like APIs. On top of private cloud people being able to use private clouds :)

PR #2323, opened 2018-08-02
Allows selectively adding environment variables to pure shells.

### LINKED ISSUE #2321: Reducing purity atomically with `#! nix-shell`
`nix-shell --pure` is awesome. It's so right it should almost be the default for interpreter use.

There is one big drawback to using `--pure` you cannot easily bring in impurities at your own convenience.

[While answering on #nixos](https://logs.nix.samueldr.com/nixos/2018-08-01#1533160407-1533160077;) I checked whether it was feasible

https://github.com/NixOS/nix/blob/b27431b7cb918e6866f4bb4fa7ec6e23240cfeab/src/nix-build/nix-build.cc#L370-L378

I think it's mildly trivial. Every hard issues seems to be handled:

 * ✔️ arguments parsing
 * ✔️ passing multiple time an argument
 * ✔️ keeping environment variables around

* * *

### Pros

 * Allows using `--pure` in situations where some parts of the environment are needed. (A deploy script using `SSH_AUTH_SOCK`).

### Cons

 * Introducing a new option. (More moving parts.)

* * * 
<sub>This can be labeled improvement</sub>

samueldr:
Inlcudes documentation and test.

This implements and fixes #2321

 * * *

I am unsure about moving `keepVars` higher up; it will now also affect `nix-build`. I am requesting your suggestions for this small bit of cleanliness.

### DIFF
--- a/doc/manual/command-ref/nix-shell.xml
+++ b/doc/manual/command-ref/nix-shell.xml
@@ -32,6 +32,7 @@
     <arg><option>--run</option> <replaceable>cmd</replaceable></arg>
     <arg><option>--exclude</option> <replaceable>regexp</replaceable></arg>
     <arg><option>--pure</option></arg>
+    <arg><option>--keep</option> <replaceable>name</replaceable></arg>
     <group choice='req'>
       <arg choice='plain'>
         <group choice='req'>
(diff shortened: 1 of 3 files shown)


doc/manual/command-ref/nix-shell.xml
```diff
@@ -32,6 +32,7 @@
     <arg><option>--run</option> <replaceable>cmd</replaceable></arg>
     <arg><option>--exclude</option> <replaceable>regexp</replaceable></arg>
     <arg><option>--pure</option></arg>
+    <arg><option>--keep-var</option> <replaceable>name</replaceable></arg>
```
edolstra: This should be `--keep` for consistency with `nix run`.

PR #2335, opened 2018-08-07
repl: don't add trailing spaces to history lines

### LINKED ISSUE #2314: nix repl leaves trailing space in history
```
$ nix repl --version   
nix (Nix) 2.0.4
```

Whenever hitting return in nix repl, the line is appended to the history of commands. Now, when pressing the arrow up key, the command is brought back to the prompt. But it has a space appended to it, which shouldn't be there.

ivan:
This fixes #2314.

I tested this manually and confirmed that a trailing space is no longer present when hitting `Up` to get a previous line, and that multi-line input still works.

Sidenote: I had to revert 7de3e00ad905bba85abadd86b83973fdba8d0dfd to build nix with nixpkgs master because the aws-sdk-cpp patch failed to apply.

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -173,9 +173,12 @@ void NixRepl::mainLoop(const std::vector<std::string> & files)
             printMsg(lvlError, format(error + "%1%%2%") % (settings.showTrace ? e.prefix() : "") % e.msg());
         }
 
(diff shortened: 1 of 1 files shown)


ivan:
Will submit a fixed PR that doesn't error on 0-byte lines

PR #2046, opened 2018-04-05
Tell user why a given path is live when it can't be deleted 

NinjaTrappeur:
Hello,

This PR changes the error message displayed while trying to remove a path from the store still referenced by a GC root.

```
$ nix-store --delete /nix/store/73h7ybrxngp9vd22jgixksy8rcidxqp0-firefox-59.0.1/
  error: cannot delete path /nix/store/73h7ybrxngp9vd22jgixksy8rcidxqp0-firefox-59.0.1' since it is still alive
```

To:
```
$ nix-store --delete /nix/store/73h7ybrxngp9vd22jgixksy8rcidxqp0-firefox-59.0.1/
  error: cannot delete path '/nix/store/73h7ybrxngp9vd22jgixksy8rcidxqp0-firefox-59.0.1': it is still needed by '/nix/store/xqkhb8hcjj7hns6qphz7zgmnvpk2rlqm-system-path'
```

This pull request fixes #1810.

Note: this is my first contribution to Nix, I'm not familiar with the code-base yet, I guess you should be extra-cautious while reviewing this.

PR #2338, opened 2018-08-08
mention `nix-store --query --roots` when a path cannot be deleted

bobvanderlinden:
As a follow-up on the discussions on PR #2046, this PR will mention the usage of
`nix-store --query --roots` whenever `nix-store` cannot delete a path due to it
still being alive.

This was suggested by @edolstra, because it is not straightforward
whether a path is alive due to it being referred to by a GC root, it being a GC root or
it being referred to by GC in-memory.

### DIFF
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -791,7 +791,11 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
             assertStorePath(i);
             tryToDelete(state, i);
             if (state.dead.find(i) == state.dead.end())
-                throw Error(format("cannot delete path '%1%' since it is still alive") % i);
+                throw Error(format(
+                    "cannot delete path '%1%' since it is still alive. "
+                    "To find out why use: "
+                    "nix-store --query --roots"
+                    ) % i);
         }
 
     } else if (options.maxFreed > 0) {


bobvanderlinden:
This PR is still relevant. Could someone have a look? @NixOS/nix-core

d-goldin:
I think this would be a pragmatic and useful quickfix, considering that https://github.com/NixOS/nix/pull/2046 has stalled. How could this PR be helped to get it merged?

bobvanderlinden:
Make it known to potential reviewers. Maybe mention it on irc/matrix? I don't want to come over as spammy with the mentions I used in this PR already.

src/libstore/gc.cc
```diff
@@ -791,7 +791,11 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
             assertStorePath(i);
             tryToDelete(state, i);
             if (state.dead.find(i) == state.dead.end())
-                throw Error(format("cannot delete path '%1%' since it is still alive") % i);
+                throw Error(format(
+                    "cannot delete path '%1%' since it is still alive. "
+                    "To find out why use: "
```
Infinisil: I think this should be
```suggestion
                    "To find out why, use: "
```

bobvanderlinden:
Thank you!

Issue #2332, opened 2018-08-05
`nix log` cannot show log without `<INSTALLABLE>`


bobvanderlinden:
This issue is related to Nix 2.0 UX.

Nix 2.0 introduced the command `nix build` where it is possible to call it without attributes. For instance:

```
nix build-f <FILE>
```

To retrieve the log of such a build I presumed the same command-line could be used to refer to the same derivation:

```
nix log -f <FILE>
```

However, this results in the following error:

```
error: more arguments are required
Try 'nix --help' for more information.
```

I tried to add an attribute that referred to the root of the file, but without succes:

```
nix log -f <FILE> self
nix log -f <FILE> .
nix log -f <FILE> ""
```
I think the `nix log` is too strict on its command-line arguments. It should operate exactly the same as `nix build`.

bobvanderlinden:
I saw this is caused because `nix log` requires `<INSTALLABLE>`, which is required. Whereas `nix build` requires `<INSTALLABLES>`, which is also required, but may be empty.

I want to try to make `nix log` act the same way as `nix build` in this regard.

lheckemann:
duplicate of #2066

Issue #2078, opened 2018-04-15
nix eval is unintuitively consistent


Profpatsch:
```
> cat hi.nix                                                  
4
```

Even though it is probably internally consistent, it is highly confusing to the user (and the error messages non-descriptive)

```
> nix eval --json --show-trace ./hi.nix                       
error: path '/home/philip/tmp/hi.nix' is not in the Nix store                         
```

Hrm, I just wanted to eval a file. Maybe I need to use `-f`

```
> nix eval --json --show-trace -f ./hi.nix                    
error: more arguments are required        
```

Huh, that’s strange. I gave it a file, what does it want from me? `--help` doesn’t exactly help, either.

```
> nix eval '4'                                                
error: don't know what to do with argument '4'           
```

What?

```     
> nix eval '(4)'                                              
4                                                                  
```

wtf, seriously?

Btw, the “correct” invocation for the file is:

```
> nix eval -f ./hi.nix ""
4
```

which follows from `nix(1)` always requiring an argument and being consistent about it.

cc @LnL7 @grahamc

lheckemann:
Basically the same issue as #2066.
The biggeste problem with this, IMHO, is that besides being difficult to use, it's *inconsistent* with `nix build`.

dtzWill:
FWIW, cc #1763 re:consistency problems even within `nix eval` itself.

wpcarro:
So if I'm understand this correctly, the best way to evaluate some `file.nix` from the command line is to run:

```
nix eval --file ./file.nix ""
```

Is that right?

Profpatsch:
@wpcarro No, the best way to do it is to use `nix-instantiate --eval`.

The `nix` tool is not yet stable and its interface probably (hopefully) going to change.

edolstra:
On the flakes branch, you can do `nix eval --file ./file.nix`.

Issue #2066, opened 2018-04-10
nix log requires an attr path


lheckemann:
nix log shouldn't complain if it doesn't get an attr path argument — it should just try and get the log of the derivation at the top level of the file so that `nix build -f foo.nix` can just be changed to `nix log -f foo.nix`.

dtzWill:
Note that `nix log` accepts things like a `./result` (when `./result` is a symlink such as produced by `nix-build` and friends), which `nix build` does not.  It would be "nice" if these all worked:

* `nix log /nix/store/hhaasshh-name`
* `nix build $ARGS` <--> `nix log $ARGS` worked (as you suggest)
* `nix log ./result` worked (currently does)
* `nix log result` worked (maybe) when `./result` exists, but I think this might not be what everyone wants

lheckemann:
FWIW `nix log -f foo.nix ''` works.

bobvanderlinden:
Please see whether #2340 is the right solution for this problem.

edolstra:
Closing since the empty attribute `""` is no longer needed since the flakes branch was merged.

PR #2340, opened 2018-08-09
make single installable argument optional, allowing `nix log`

bobvanderlinden:
Previously in Nix commands like log always required an argument. This was confusing, because it is often used in combination with `nix build`, but does not expect the same arguments:

```
$ nix build
$ nix log
error: more arguments are required
```

A workaround was to use `""` as argument, since that same value is used by default for the build command:

```
$ nix build
$ nix log ""
```

This is because `nix log` has the following usage:

```
Usage: nix log <FLAGS>... <INSTALLABLE>
```

whereas `nix build` has:

```
Usage: nix build <FLAGS>... <INSTALLABLES>...
```

This PR will make the INSTALLABLE of `nix log`, `nix eval` and `nix edit` optional. That makes the usage as follows:

```
Usage: nix log <FLAGS>... <INSTALLABLE>?
```

Solves #2332 #2066
Partially solves #2078

### DIFF
--- a/src/nix/command.hh
+++ b/src/nix/command.hh
@@ -121,10 +121,12 @@ struct InstallableCommand : virtual Args, SourceExprCommand
 
     InstallableCommand()
     {
-        expectArg("installable", &_installable);
+        expectArg("installable", &_installable, useDefaultInstallable());
     }
 
     void prepare() override;
+
+    virtual bool useDefaultInstallable() { return true; }
 
 private:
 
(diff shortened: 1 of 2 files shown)


bobvanderlinden:
Apparently this PR is still relevant. Could someone have a look? @NixOS/nix-core

bobvanderlinden:
This was fixed according to https://github.com/NixOS/nix/issues/2066#issuecomment-668026174

Issue #2208, opened 2018-06-04
nix-shell dependencies can be garbage collected any time now / persistent nix-shell envs


goodwillcoding:
There does not seem to be a way to prevent nix-shell dependencies from being garbage collected at any time. Related to this, for a while now there been possible to create persistent nix-shell environments that would be spared garbage collection. Neither of the two methods (see below) are now working.

*shell.nix*
```
with import <nixpkgs> {};
stdenv.mkDerivation {
    src = "./.";
    name = "test-0.1";
    buildInputs = [ hello ];
}
```

Previous Method 1 (https://nixos.org/nix-dev/2016-November/022173.html):
```
nix-instantiate shell.nix --show-trace  --indirect --add-root "$(pwd)/"shell.drv
nix-shell $(readlink $(pwd)/shell.drv)
``` 
While this does create a gcroot to the derivatation, it does not work (my guess cause it is not a gcroom to the output)
```
$ tree /nix/var/nix/gcroots/auto/
/nix/var/nix/gcroots/auto/
└── rz3awpwi7lxpwfjmhqbab0rdfc3nj94m -> /home/user/tmp/persist-nix-shell/shell.drv
```
If nix-collect-gc is run while the shell is active it cleans up everything this derivation depends on

Previous Method 2:
```
mkdir gcroot
nix-shell shell.nix --indirect --add-root "gcroot/gc"
```
This used to create symlinks to dependencies, how ever since `--indirect` is now ignored as of https://github.com/NixOS/nix/commit/19477e8815a7c9b91c82278ec0e22957c97bec16 this also no longer works as gcroot to the output is no longer created.

Please advise on this

goodwillcoding:
@domenkozar any thoughts? ^

goodwillcoding:
@edolstra any thoughts? suggestion on this ^

srghma:
Method 1 makes gcroot for shell derivation, this indeed doesnt prevent it dependencies to be garbage collected
Method 2 should make gcroot for each dependency of shell derivation, but it's not working indeed

I have made script that makes same as method 2

```
#! /bin/sh -eu

# make gcroot for each dependency of shell and run nix-shell

mkdir -p .gcroots
nix-instantiate shell.nix --indirect --add-root $PWD/.gcroots/shell.drv
nix-store --indirect --add-root $PWD/.gcroots/shell.dep --realise $(nix-store --query --references $PWD/.gcroots/shell.drv)
[...]

Issue #1982, opened 2018-03-16
nix-build: --add-root ignored


lheckemann:
What I did: `nix-build '<nixpkgs>' -A hello --add-root test`
Expected behaviour: test symlink to exist afterwards, or perhaps an error message complaining that `--add-root` is not a recognised option, or pointing me to `-o`
Actual behaviour: result symlink, no test symlink, no error message

ElvishJerricco:
Same thing with `nix-shell`. `-o test` doesn't create any symlink either.

bdesham:
Is there any kind of workaround for this? Is it possible to create a GC root manually?

(I’m running into this issue in the context of https://github.com/commercialhaskell/stack/issues/4673. I have to run `nix-store --gc` every so often, because I don’t have that much drive space left, but then the next build of my Stack-based Haskell project takes 10 minutes of maxing out all of my CPU cores.)

lheckemann:
@bdesham yes, `nix-build` will create a `result` symlink and make it a GC root by default. You can change the name of the symlink with the `-o` option. The issue I'm describing here is that the `--add-root` option is silently ignored instead of erroring or similar.

You have multiple ways available for creating GC roots:
 - Installing paths in your user profile, system profile, or home profile using `nix-env` or `configuration.nix` or `home-manager` respectively
 - using `nix-build`'s default behaviour, with or without `-o`
 - using `nix-store -r /nix/store/$hash-$name --add-root ./root-name --indirect`
 - manually by making symlinks in `/nix/var/nix/gcroots` — I used this in a terribly ugly script at one point to root an entire build closure: https://github.com/lheckemann/dotfiles/blob/master/gcroots.sh

jakubgs:
I just wasted like an hour trying to figure out what I was doing wrong. Turns out `nix-build` just silently ignores `--add-root`. Not cool.

zimbatm:
lol, @jakubgs. I guess the thousands of hours that other people spent on the project don't count in the balance.

jakubgs:
How did you read that from my comment?

bdesham:
This is still a problem.

jakubgs:
No.

burdiyan:
[...]

PR #2346, opened 2018-08-12
[WIP] +RFC fixing arguments parsing and documentation for nix-build and nix-shell

samueldr:
`nix-build` and `nix-shell` are both implemented in the same multi-call binary.

The current implementation for arguments parsing makes them share arguments. This can lead to surprising behaviour (or lack thereof) when accidentally using a non-shared parameter.

Abridged list of nonsensical invocations:

 * `nix-build --command "foo" ./release.nix`
 * `nix-build --impure ./release.nix -A build.x86_64-linux`
 * `nix-shell --out-link "foo" ./shell.nix`

There are a couple more that could be added.

This PR intends to fix the behaviour by segregating the arguments parsing using an `if` conditional. That is, unless a better method is suggested.

Additionally, this PR adds the missing documentation for a couple of options, since while checking for the "owner" of the parameters, I realized that a couple options weren't documented.

#### Expectations

I am expecting a bit of push-back from "simply removing" the options. That is, it technically breaks backwards-compatibility with (wrong) usage of the 2.x series for nix-build and nix-shell. This has to be kept in mind.

If it is judged better to keep the non-sensical options, it may be preferred to keep them segregated in their respective binaries, turned into no-ops in the other and documented as such.

#### RFC stdin

I find the `-` (stdin) behaviour surprising with regards to `nix-shell`.

 * `nix-shell ./shell.nix` != `echo 'import ./shell.nix' | nix-shell -`

As for `nix-build` it seems the behaviour is fine.

I am probably overlooking something with the semantics of it. Though, the way it's implemented it may be better to drop this flag in `nix-shell`? The only documentation for `-` as standard input is [for nix-instantiate](https://nixos.org/nix/manual/#description-39), and there are no mention of it in neither the `nix-build` nor the `nix-shell` manuals. (They mention `nix-instantiate.` only for `--arg | --attr | -A`.

* * *

I may split this PR into a documentation-only one that can be fast-tracked. I may also split this PR furthermore while it's WIP.

* * *

Furthermore, this *will* (still a WIP) try to address #1982 and #2208 by actually implementing the flag.

### DIFF
--- a/doc/manual/command-ref/nix-build.xml
+++ b/doc/manual/command-ref/nix-build.xml
@@ -22,6 +22,12 @@
     <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="opt-common-syn.xml#xmlns(db=http://docbook.org/ns/docbook)xpointer(/db:nop/*)" />
     <arg><option>--arg</option> <replaceable>name</replaceable> <replaceable>value</replaceable></arg>
     <arg><option>--argstr</option> <replaceable>name</replaceable> <replaceable>value</replaceable></arg>
+    <arg><option>--check</option></arg>
+    <arg><option>--dry-run</option></arg>
+    <group>
+      <arg choice='plain'><option>--expr</option></arg>
+      <arg choice='plain'><option>-E</option></arg>
+    </group>
     <arg>
       <group choice='req'>
         <arg choice='plain'><option>--attr</option></arg>
--- a/doc/manual/command-ref/nix-shell.xml
+++ b/doc/manual/command-ref/nix-shell.xml
@@ -19,6 +19,7 @@
 <refsynopsisdiv>
   <cmdsynopsis>
     <command>nix-shell</command>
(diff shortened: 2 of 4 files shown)


dtzWill:
:heart:

lheckemann:
Any reason to attempt to address #2208 in this PR rather than only separating the options and implementing the shell-gc-rooting stuff in another one? With that considered, is this still WIP?

samueldr:
The only reason was that it was possible to do it while passing there. It wasn't the initial goal of the PR, and yeah, probably should be handled in another PR.

The actual PR, "fixing arguments parsing and documentation" isn't much WIP, but more "tell me what I'm doing wrong". See the "Expectations" and "RFC stdin" sections for other parts where I'm not entirely confident.

asymmetric:
I think it would make sense to split out the documentation part from the PR and fast-track it.

domenkozar:
I think such PRs would be eaiser to merge if they come with a bunch of tests :)

PR #2350, opened 2018-08-13
ignore when listxattr fails with ENODATA

symphorien:
This happens on CIFS and means the remote filesystem has no extended
attributes (if I understand correctly).

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -450,7 +450,7 @@ static void canonicalisePathMetaData_(const Path & path, uid_t fromUid, InodesSe
     ssize_t eaSize = llistxattr(path.c_str(), nullptr, 0);
 
     if (eaSize < 0) {
-        if (errno != ENOTSUP)
+        if (errno != ENOTSUP && errno != ENODATA)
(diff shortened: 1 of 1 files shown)


symphorien:
I looked superficially in the code, the ENODATA seems to come from here https://elixir.bootlin.com/linux/latest/source/fs/cifs/smb2maperror.c#L380
So the error means "the remote fs has xattrs disabled" not that this particuliar file has no xattr. So this is approximately ENOTSUP.

dezgeg:
Here they explicitly try to map `-ENODATA` for `listxattr()` to returning 0: https://elixir.bootlin.com/linux/latest/source/fs/cifs/smb2ops.c#L794

````
	/*
	 * If ea_name is NULL (listxattr) and there are no EAs, return 0 as it's
	 * not an error. Otherwise, the specified ea_name was not found.
	 */
	if (!rc)
		rc = move_smb2_ea_to_cifs(ea_data, buf_size, smb2_data,
					  SMB2_MAX_EA_BUF, ea_name);
	else if (!ea_name && rc == -ENODATA)
		rc = 0;
````

But something is clearly going wrong.

PR #2352, opened 2018-08-14
docs: Update installation location on the hacking section

carlosdagos:
I followed the docs and noticed that `make install` will install nix in under `$(pwd)/inst` and not `$(pwd)/nix` as the docs currently state 😄


edolstra:
Ah, sorry, didn't notice this before merging #2365.

carlosdagos:
Cheers @edolstra 😄

PR #2358, opened 2018-08-17
fetchGit: use a better caching scheme

graham-at-target:
The current usage technically works by putting multiple different
repos in to the same git directory. However, it is very slow as
Git tries very hard to find common commits between the two
repositories. If the two repositories are large (like Nixpkgs and
another long-running project,) it is maddeningly slow.

This change busts the cache for existing deployments, but users
will be promptly repaid in per-repository performance.

---

I ran the tests and the fetchGit test passed. There is an unrelated failure in brotli.sh, which is already failing on Darwin: https://hydra.nixos.org/build/79518003/nixlog/1 which shouldn't block this merging.

### DIFF
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -3,6 +3,7 @@
 #include "download.hh"
 #include "store-api.hh"
 #include "pathlocks.hh"
+#include "hash.hh"
 
 #include <sys/time.h>
 
@@ -84,9 +85,10 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
     if (rev != "" && !std::regex_match(rev, revRegex))
         throw Error("invalid Git revision '%s'", rev);
 
-    Path cacheDir = getCacheDir() + "/nix/git";
+    Path cacheDir = getCacheDir() + "/nix/gitv2/" + hashString(htSHA256, uri).to_string(Base32, false);
 
     if (!pathExists(cacheDir)) {
+        createDirs(dirOf(cacheDir));
         runProgram("git", true, { "init", "--bare", cacheDir });
     }
 
--- a/tests/fetchGit.sh
+++ b/tests/fetchGit.sh
@@ -9,7 +9,7 @@ clearStore
 
 repo=$TEST_ROOT/git
 
-rm -rf $repo ${repo}-tmp $TEST_HOME/.cache/nix/git
+rm -rf $repo ${repo}-tmp $TEST_HOME/.cache/nix/gitv2
 
 git init $repo
 git -C $repo config user.email "foobar@example.com"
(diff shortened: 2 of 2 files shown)


graham-at-target:
(doing some real-life testing now, outside of the automated tests.)

graham-at-target:
🥗 seems to work okay from here on Darwin!

edolstra:
I'm not really in favor of this because it multiplies the storage required for Nixpkgs (e.g. if you have a Nixpkgs clone from several repositories, like `NixOS/nixpkgs` and `NixOS/nixos-channels`).

LnL7:
The problem is that this becomes unusable when used in combination with other projects. I've seen a warning like `WARNING: no common commits` when using a new repository for the first time, maybe that could be used somehow?

Alternatively using eg. commit 0 as the identity instead of the url might be an option.

grahamc:
Right. Fetchgit can take multiple minutes per repository, just scanning for mutual histories.  This makes —pure quite painful. 

Unfortunately there is no true commit 0 because a a repository is a full dag and can support multiple initial commits. 

An ugly alternative idea might be allow fetchgit to accept an optional cache key param to override the default.

As is though fetchgit is only really usable to fetch a single repo.

grahamc:
I did a fetch of the Linux kernel in to a repo already containing nixpkgs, and the fetch took an extra 100s.

w.r.t. multiple parented repos: they can have something in common though :$ For an example, here is the beautiful future I want, Nixpkgs merged with Linux... Linix:

```
$ git clone git@github.com:nixos/nixpkgs.git
$ cd nixpkgs
$ git remote add linux git@github.com:torvalds/linux.git
$ git fetch linux
$ git merge --allow-unrelated-histories linux/master
```

AmineChikhaoui:
@edolstra Note that this is useful as we want to switch the hydra mercurial/git inputs plugins to the builtins.fetch*.

Mic92:
If storage requirements for fetching from `NixOS/nixos-channels` and `NixOS/nixpkgs` are really the only blocker for this pull request I would rather also mirror the branches from nixos-channels to nixpkgs.

lheckemann:
How about using the `name` in the attrset passed to fetchGit when available? This would not only allow sharing between nixpkgs and nixpkgs-channels (by always setting the name to nixpkgs even when fetching from channels), but also:
 - Allow sharing between other repos which may be fetched via different URLs (e.g. a local clone in addition to an HTTP one, to save downloading the whole repo)
 - encourage people to set `name`s on their `fetchGit` calls, which has the side benefit of providing a more useful store path than `{hash}-source`.

edolstra:
I've implemented an alternative to `fetchGit` that (at least for GitHub repos) eliminates the disk space issue, at the cost of not returning a `revCount` attribute: https://github.com/edolstra/nix/commit/7d33eb83455bb94add9ecdf224a0c38eb2b55575. Since this could be used for fetching Nixpkgs, it makes the disk space usage of `fetchGit` less of an issue.

Mic92:
mhm. I am a bit worried about running into API limits. When people have a few repositories 500 requests per IP can be reached easily.

Mic92:
It is even worth 60 requests per hour: https://developer.github.com/v3/#rate-limiting

Mic92:
Also downloading a tarball does not really require the API,
since fetchurl can be used directly to fetch a certain tag.

PR #2366, opened 2018-08-21
fetchGit: Remove unneeded call to git

erikarvstedt:
Originally, I wrote this to fix a nix performance issue, which turned out to have other causes specific to my setup.

This patch might still be useful as it saves a needless call to git in the common case where a rev is given that is already in `/nix/store`.
In this case, now only `~/.cache/nix/git` and `/nix/store` are queried and no subprocesses are started.

### DIFF
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -24,6 +24,20 @@ struct GitInfo
 
 std::regex revRegex("^[0-9a-fA-F]{40}$");
 
+static bool isRevInCache(const std::string & rev, const Path & cacheDir)
+{
+    try {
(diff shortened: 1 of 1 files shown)


erikarvstedt:
Your issue has a different scope, but it's still very relevant.
It could be solved with [this simple fix](https://github.com/erikarvstedt/nix/commit/f237866fa845be572d90c7e539ff0e2757600b6b), where links to fetched revisions are just identified by their SHA-1 hash (and not by name + SHA-1).
It wouldn't increase the attack surface for SHA-1 collisions in any meaningful way, so I don't see any downsides.

tomberek:
review?

erikarvstedt:
The current version of the git fetcher (`libfetchers/git.cc`) doesn't call git on a cache hit, so this PR is obsolete.

PR #2383, opened 2018-08-30
fix `error: unknown serve command 9`

cleverca22:
this appears to be a typo when checking the remote store version, and the end-result is sending serve commands the remote end cant support


edolstra:
Good catch, thanks!

PR #2384, opened 2018-08-31
docs: Add some examples to fetchGit

graham-at-target:
I found a bug - #2385 as part of writing this, but it shouldn't block these docs.

Ideally these would be backported.

Rendered:

image

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -402,17 +402,10 @@ stdenv.mkDerivation { … }
       <example>
         <title>Fetching a private repository over SSH</title>
         <programlisting>builtins.fetchGit {
-  url = "ssh://git@github.com/my-secret/repository.git";
+  url = "git@github.com:my-secret/repository.git";
   ref = "master";
   rev = "adab8b916a45068c044658c4158d81878f9ed1c3";
 }</programlisting>
-
-      <note><para>
-        Note the URL format is not the same as <command>git
-        clone</command>. <function>builtins.fetchGit</function> uses
-        a <literal>/</literal> instead of a <literal>:</literal>
-        between <literal>github.com</literal> and
-        <literal>my-secret</literal>.</para></note>
       </example>
 
       <example>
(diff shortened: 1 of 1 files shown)


doc/manual/expressions/builtins.xml
```diff
@@ -398,6 +398,91 @@ stdenv.mkDerivation { … }
...
+  ref = "master";
+  rev = "adab8b916a45068c044658c4158d81878f9ed1c3";
+}</programlisting>
+
+      <note><para>
+        Note the URL format is not the same as <command>git
+        clone</command>. <function>builtins.fetchGit</function> uses
+        a <literal>/</literal> instead of a <literal>:</literal>
+        between <literal>github.com</literal> and
+        <literal>my-secret</literal>.</para></note>
```
edolstra: Why is that? I'm not aware that we rewrite Git URLs anywhere.

grahamc: Maybe the explanation is not totally correct, but:


detected as a directory:

```
nix-repl> builtins.fetchGit { url = "git@github.com:grahamc/nixos-config.git"; ref = "master"; }    
fetching Git repository '/home/grahamc/git@github.com:grahamc/nixos-config.git'fatal: '/home/grahamc/git@github.com:grahamc/nixos-config.git' does not appear to be a git repository
fatal: Could not read from remote repository.
```

confused hostname:

```
nix-repl> builtins.fetchGit {   url = "ssh://git@github.com:grahamc/nixos-config.git";   ref = "master";  }     
fetching Git repository 'ssh://git@github.com:grahamc/nixos-config.git'ssh: Could not resolve hostname github.com:grahamc: Name or service not known
fatal: Could not read from remote repository.
```

just right:

```
nix-repl> builtins.fetchGit {   url = "ssh://git@github.com/grahamc/nixos-config.git";   ref = "master";  }      
{ outPath = "/nix/store/bf5xjh16pcgxk2lj4frmby26740q10wg-source"; rev = "0524294c41343b7c400ddc1d7442dc6bde308f95"; revCount = 66; shortRev = "0524294"; }
```

LnL7: I think you are confusing `ssh://git@github.com/NixOS/nix.git` with `git@github.com:NixOS/nix.git`, the second format isn't supported in nix since that's not a uri.

grahamc: Updated!

PR #2386, opened 2018-08-31
Manual: Passing an env var to a pure nix shell with getEnv

Profpatsch:
Describe a practical use of `builtins.getEnv` with `nix-shell`
that is not directly obvious.

Rendered: 
screenshot

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -558,7 +558,20 @@ builtins.genList (x: x * x) 5
     locate the file <filename>~/.nixpkgs/config.nix</filename>, which
     contains user-local settings for Nix Packages.  (That is, it does
     a <literal>getEnv "HOME"</literal> to locate the user’s home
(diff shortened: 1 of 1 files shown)


grahamc:
Would it be better to use nix-shell's `--pure` option?

edolstra:
Hm, the `SSH_AUTH_SOCK` thing could be considered an anti-pattern so I'm not sure it should be described here. I mean, presumably you only want `SSH_AUTH_SOCK` to be passed to the shell, not to a build, but Nix doesn't distinguish between those cases.

Profpatsch:
> I mean, presumably you only want SSH_AUTH_SOCK to be passed to the shell, not to a build

Yes, that’s why I added the “(exclusively)”. It beats `--keep` for usage of e.g. `pkgs.mkShell` in that the `shell.nix` file can contain the whole “source of truth” for developers, no complex invocation or extra wrapper script required.

PR #2392, opened 2018-09-01
Improve stats

cleverca22:
this adds percentages to the bytes in the stats
`  values allocated: 119 (2856 bytes 45%)`
and also adds json:
```
[clever@amd-nixos:~/apps/nix]$ NIX_SHOW_STATS=json NIX_SHOW_STATS_PATH=foo.json ./result/bin/nix-instantiate --eval -E '5*5'
warning: unknown setting 'build-extra-platforms'
25
evaluation statistics:

[clever@amd-nixos:~/apps/nix]$ jq .values.bytes < foo.json 
2856
```

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1730,9 +1730,7 @@ bool EvalState::eqValues(Value & v1, Value & v2)
 void EvalState::printStats()
 {
     bool showStats = getEnv("NIX_SHOW_STATS", "0") != "0";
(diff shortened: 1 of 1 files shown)


edolstra:
Looks good. Could you remove the old stats format? I don't think we need to have both, JSON output is strictly better.

vcunat:
Cross-link: https://github.com/NixOS/nixpkgs/issues/53858

PR #2394, opened 2018-09-02
nix doctor: add command

LnL7:
This command performs a bunch of sanity checks on the nix installation, most of these are not necceceraly a problem but can be very useful when debugging an issue. Most notably a way to detect worker/serve protocol mismatches in various conditions. This also works in combination with `--store ssh://host` to help debug eg. nix-copy-closure issues like https://github.com/NixOS/nix/pull/2383.

TODO: also include the details `nix-info` generates?

The code probably also needs a bit of cleanup.

---

```
$ nix doctor
Store uri: daemon

Warning: multiple versions of nix found in PATH.

  /nix/store/1wgsprdh6lafnqf8iddnm01k9kxvx5xv-nix-2.0.4/bin
  /src/nix/inst/bin

Warning: found profiles outside of /nix/var/nix/profiles.
The generation this profile points to might not have a gcroot and could be
garbage collected, resulting in broken symlinks.

  /tmp/foo/bin

Warning: protocol version of this client does not match the store.
While this is not necessarily a problem it's recommended to keep the client in
sync with the daemon.

Client protocol: 1.21
Store protocol: 1.20
```

### DIFF
--- a/src/nix/doctor.cc
+++ b/src/nix/doctor.cc
@@ -33,7 +33,24 @@ struct CmdDoctor : StoreCommand
         std::cout << "Store uri: " << store->getUri() << std::endl;
         std::cout << std::endl;
 
+        checkNixInPath();
         checkStoreProtocol(store->getProtocol());
+    }
+
+    void checkNixInPath() {
+        PathSet dirs;
+
+        for (auto & dir : tokenizeString<Strings>(getEnv("PATH"), ":"))
+            if (pathExists(dir + "/nix-env"))
+                dirs.insert(dirOf(canonPath(dir + "/nix-env", true)));
+
+        if (dirs.size() != 1) {
+            std::cout << "Warning: multiple versions of nix found in PATH." << std::endl;
+            std::cout << std::endl;
+            for (auto & dir : dirs)
+                std::cout << "  " << dir << std::endl;
+            std::cout << std::endl;
+        }
     }
 
     void checkStoreProtocol(unsigned int storeProto) {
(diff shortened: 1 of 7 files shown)


LnL7:
I didn't spend a lot of time on the error messages, they can definitively be improved. But I'm not sure if moving the all the `<profile>` with all of it's `<profile>-<gen>-link` symlinks and creating a pointer to those (like ~/.nix-profile) is always correct.

src/nix/doctor.cc
```diff
@@ -0,0 +1,111 @@
...
+                auto profileDir = canonPath(dirOf(dir), true);
+                if (hasSuffix(profileDir, "user-environment") &&
+                    store->isValidPath(profileDir)) {
+                    PathSet referrers;
+                    store->computeFSClosure({profileDir}, referrers, true,
+                            settings.gcKeepOutputs, settings.gcKeepDerivations);
+                    bool found = false;
+                    for (auto & i : roots)
+                        if (referrers.find(i.second) != referrers.end())
+                            found = true;
```
LnL7: Is there a better way to do this? It's a slow and can trigger gc unrelated messages.

copumpkin:
This is awesome. As the go-to person for "dude my Nix is broken halp" at my company, this will be a godsend for me 😄

copumpkin:
Not sure how best to implement it, but some way to determine if `nix-daemon` was started before your changed nix.conf would be pretty neat. I've seen a lot of people get confused because they change a setting in nix.conf and don't realize they need to restart the daemon.

timokau:
Would it make sense to merge this with `nix-info`? That way you could just tell somebody to execute `nix doctor` and post the results (if they can't fix it themselves based on the report).

tomberek:
Merge?

PR #2406, opened 2018-09-07
Set environment variable NIX_SHELL_NAME to derivation name

kampka:
When in a nix-shell environment, I like to have the derivation name exposed on my prompt.
At the moment this is a bit tricky to do, thus this change introduces an environment variable that exports the derivation name.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -446,6 +446,7 @@ void mainWrapped(int argc, char * * argv)
 
         restoreSignals();
 
+        setenv("NIX_SHELL_NAME", drvInfo.queryName().c_str(), 1);
         execvp(shell.c_str(), argPtrs.data());
 
         throw SysError("executing shell '%s'", shell);


copumpkin:
You should get this automatically today from the `name` environment variable. It's not uppercase or prefixed so might not feel like your average environment variable, but on the plus side behaves like all other derivation fields so is less special-casey. Does that suit your needs?

kampka:
That is actually what I am currently doing, but it feels very implicit as `name` is a very ambiguous token.
The main purpose of this request is to make the use very explicit.

kampka:
ping

Would be nice to have a decision (or discussion) either way.

edolstra:
IMHO the combination of `$IN_NIX_SHELL` and `$name` already allows doing what you want.

kampka:
True, but as I already said, this PR is not about making it possible, it's about making it explicitly clear what's what instead of relying on a generic token name `$name`.

PR #2409, opened 2018-09-07
Make fetchGit do full checkout if ref = "*"

yegortimoshenko:
Sometimes one might want to fetch a rev without knowing the ref. In this case, full repo clone is usually required. This commit adds a special case to ref handling, so that when `ref = "*"`, `fetchGit` fetches all refs prior to fetching the rev.

Usage example:

```
let
 repo = fetchGit {
   url = "https://github.com/serokell/nixage.git";
   ref = "*";
   rev = "4b83ae894ab99b55c67b6629058b3a4fdd8ba0c2";
 };
in

builtins.trace repo repo
```

### DIFF
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -120,6 +120,13 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
     {
         Activity act(*logger, lvlTalkative, actUnknown, fmt("fetching Git repository '%s'", uri));
 
+        if (ref == "*"s) {
+          /* Do full checkout. */
+          runProgram("git", true, { "-C", cacheDir, "fetch", "--quiet", "--force", "--", uri, "+refs/heads/*:refs/heads/*" });
+
+          ref = "HEAD"s;
+        }
+
         // FIXME: git stderr messes up our progress indicator, so
         // we're using --quiet for now. Should process its stderr.
         runProgram("git", true, { "-C", cacheDir, "fetch", "--quiet", "--force", "--", uri, *ref + ":" + localRef });


yegortimoshenko:
@edolstra Can I get some feedback on this?

This feature is used extensively as part of a fully sandboxable [stack-to-nix](https://github.com/serokell/stack-to-nix) fetcher. Currently, all users have to install patched Nix to their user profile (which works because patch is eval time only).

grahamc:
I'm not sure it is a great idea, because even '*' isn't all the refs. For example, github has lots of refs that wouldn't be fetched but are valid refs, and this list could change over time. For that reason, I feel the best way to go is specifying the correct ref from the start.

I understand this is complicated because other tools are making invalid assumptions about Git as well. The problem and question is do we propagate this invalid assumption to Nix? In particular this question is hard because Nix's goal is for very long-term reproducibility, which is at odds with this request.

yegortimoshenko:
Without this feature, sandboxable Nix generators for Stack and any other lock file format that doesn't include refs are not possible (even in upcoming Stack freeze file). So my motivation is primarily to improve on that front, allowing to build projects via Nix without any intermediary non-sandboxed step (i.e. what `stack2nix` does).

There are also other affected lock file formats:
- `cabal.project`: https://github.com/int-index/foundry/blob/master/cabal.project#L7
- Mix: https://github.com/serokell/hermetic/blob/master/mix.lock#L14
- Rebar3: https://github.com/serokell/derivery/blob/master/rebar.lock#L10-L11

I will check other tools and add them to the list. I think NPM and Yarn too have the same quality, and it's not limited to just that in particular.

> I understand this is complicated because other tools are making invalid assumptions about Git as well. The problem and question is do we propagate this invalid assumption to Nix? In particular this question is hard because Nix's goal is for very long-term reproducibility, which is at odds with this request.

I think this feature leads to higher, not lower reproducibility, because ref is incidental to rev. Meaning, it won't matter on which ref is the rev that we're interested in, and so if we were trying to fetch a rev from an ephemeral feature branch, `ref = "*";` won't break, while `ref = "feature-foo";` will.

And fundamentally, any remote location fetch is not long-term reproducible, it always depends on other parties.

I think the correct question (in Nix context) to ask here is: will it always produce the exact same result? The answer here seems to be yes. Having this available will make a whole class of new convenient sandboxable Nix generators possible.

grahamc:
> I think the correct question (in Nix context) to ask here is: will it always produce the exact same result? The answer here seems to be yes. Having this available will make a whole class of new convenient sandboxable Nix generators possible.

It won't because the remote list of branches isn't certain to be stable, but your point about it remaining true for any branch name seems legitimate.

Perhaps the default should be instead of having to specify `*`, an empty ref fetches all I refs.

I think some input from @NixOS/nix-core would be helpful.

grahamc:
ping @copumpkin @edolstra @peti @shlevy @vcunat

shlevy:
@yegortimoshenko I think having this functionality make sense, but I'd like to ensure we're matching the same logic these other tools use and nothing more. Does stack do a git clone, or fetch all refs, or what? Same question for mix, npm, etc. Once we're settled on that we can figure out the interface and everything.

yorickvP:
[Mix is doing](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/scm/git.ex#L90):
```
git -c core.hooksPath='' init --quiet
git remote add origin <url>
git fetch --force --quiet
git checkout --quiet rev
if <submodules> 
  git -c core.hooksPath='' submodule update --init --recursive
```

kirelagin:
Just to clarify, `git fetch` without any refspecs consults the `remote.origin.fetch` configuration option, which is set by default (after `git init && git remote add origin <url>` or `git clone <url>`) to `+refs/heads/*:refs/remotes/origin/*`, so, in other words, it fetches all _heads_.

shlevy:
OK, so let's switch this to fetch all heads?

As for interface... I don't really like `"*"`. I'm leaning toward just having ref omitted following the behavior of fetch and clone.

yegortimoshenko:
It already does fetch all heads:

https://github.com/NixOS/nix/blob/4b279a099f19bd13a98fa2ae519c4aa263bb6b6e/src/libexpr/primops/fetchGit.cc#L125

> As for interface... I don't really like `"*"`. I'm leaning toward just having ref omitted following the behavior of fetch and clone.

You mean:

```
fetchGit {
  url = https://github.com/foo/bar;
}
```

Or rather:

```
fetchGit {
  url = https://github.com/foo/bar;
  ref = "";
}
```
```

**Comment by shlevy** [comment]
Ah! Your description said all `refs`, that's why I was confused. OK, then I think the current behavior is fine.

And I meant the former example there, though I think with the way the primop is implemented both would work the same.

**Comment by yegortimoshenko** [comment]
> And I meant the former example there, though I think with the way the primop is implemented both would work the same.

OK! 

Unlike `url` or `rev`, `ref` is wrapped into `std::experimental::optional`, so those two would be different as far as I understand.

**Comment by mkaito** [comment]
I don't mean to be rude, but I'd love to see this merged. Could I help it along somehow?

**Comment by edolstra** [comment]
I stated my opinion here: https://github.com/NixOS/nix/issues/2431#issuecomment-423489879

**Comment by yorickvP** [comment]
We're using this in [serokell/stack-to-nix](https://github.com/serokell/stack-to-nix) (and [serokell/mix-to-nix](https://github.com/serokell/mix-to-nix/)), translating git repo fetches from language-specific lock files. These files includes the hashes and url, but nothing more. External repo's are not guaranteed to even have a ref for a commit. Furthermore, for internal repos, we may need ssh keys.

In the public github case, we can use `fetchTarball`, but it does not generalize to arbitrary git repos and does not cache as much as it could.

The only alternative here is probably `extraBuiltins`, which is less than optimal for portability (but better than a patched nix).

**Comment by yegortimoshenko** [comment]
I will update this patch following @shlevy's feedback until Friday, sorry for the delay.

**Comment by yorickvP** [comment]
This seems to be less needed with nix 2.2. It's possible github enabled fetching unadvertised refs, and a recent nix change allows the `"*"` through.

**Comment by kirelagin** [comment]
To be more precise, since [this commit](https://github.com/NixOS/nix/commit/#diff-7e39b3dccfe1a07ed3f4af7f4d8da628) the example from this PR is translated to `git fetch ... '*:*'`. I don’t quite understand what this refspec means, but I would guess it says “match all refs that the remote advertises”, but I don’t know how remotes determine what refs to advertise.

On a separate note, since maintaining a patched Nix is a little annoying, here is a Nix plugin that replaces the `fetchGit` primop with the one from this PR: [kirelagin/nix-fetchgit](https://github.com/kirelagin/nix-fetchgit).

**Comment by edolstra** [comment]
What change? It's certainly not intended to allow fetching `*`. If it does it's a bug.

**Comment by edolstra** [comment]
Thanks. I guess we should really validate `ref` since it's potentially untrusted user input that we're passing to another program on the command line. The use of `--` removes most risks but it still seems better to validate it.

**Comment by codygman** [comment]
I was bit again by this today, but it seems with slight modification the perfect solution is already in this thread.

> Perhaps the default should be instead of having to specify *, an empty ref fetches all I refs.

To be explicit, that means:

```
fetchGit {
  url = https://github.com/foo/bar;
}
```

Then provide a warning:

```
Warning: No ref specified for git://repo.git, FULL clone being performed, this could be very slow. Please specify the 'ref' so a shallow clone can be performed
```
Then if someone forgets it, the problem is "it's slow" rather than the inscrutable `fatal: not a tree object error`. By default we could output "ref not specified, fully cloning git://repo.git" and give users a warning to silence that.

That lets us:

1) push people towards the correct solution:

> To me not specifying a ref is just bad practice. - @edolstra 

2) We allow what most new nix users will do by default since as many external tools @yorickvP listed make this admittedly invalid assumption.

As someone who has evangelized Haskell quite a lot, and now nix I have to say...

For many an error like this where compatibility is broken in the name of technical correctness, and to top it all off a bad error message is given, would be reason to drop nix completely.

I think this solution lets us meet the software world and it's invalid assumptions where they are, but smooths their transition and nudges them towards more correctness and it's benefits.

**Comment by Ma27** [comment]
@edolstra @grahamc is there anything to be done to get this merged? Just got bitten by this as well.

**Comment by yorickvP** [comment]
@Ma27 `ref = "*";` should work in recent nix versions already. Also, it seems github might have enabled `allow-reachable-sha1-in-want` recently, but reports vary.

**Comment by Ma27** [comment]
@yorickvP on `nix (Nix) 2.4pre20200622_334e26b` (`pkgs.nixFlakes`) it doesn't (for public GitHub repositories) and breaks with the following error message:

```
invalid Git branch/tag name '*'
```

yorickvP:
For public github repo's, fetching with the https .tar.gz archive download should always work. You can try this as a workaround. It's sad to hear that nix 2.4 breaks ref="*" again.

Ma27:
> For public github repo's, fetching with the https .tar.gz archive download should always work.

I'm well-aware of that. I described my use-case (which *requires* `fetchGit` or something similar) here: https://github.com/NixOS/nix/issues/2431#issuecomment-657198373

> It's sad to hear that nix 2.4 breaks ref="*" again.

Ahh, I assumed that you meant the unstable versions by mentioning "recent nix versions". It actually works on `nix 2.3.7`, thanks for pointing that out!

On Nix 2.4 (`master` and `flakes` variant IIRC) the fetchers have been completely rewritten (see `src/libfetchers`) to provide a lot more inputs easily. I'll try to restore the `ref = "*";` feature later (but hopefully today).

PR #2422, opened 2018-09-14
Place global lock on fetchGit

yegortimoshenko:
We use `fetchGit` in CI, and lack of a lock causes irrecoverable issues once in a while: https://gist.github.com/serokell-bot/081bc82276e227fe0f34fdb3cfcf3361

There are many things that can go wrong prior to `storeLinkLock`, so this places global lock on `fetchGit`.


yegortimoshenko:
That doesn't seem to resolve the issue completely, so closing until the root cause is found.

Issue #1930, opened 2018-02-28
Nix 2.0: Print out path to stdout with nix build


ElvishJerricco:
This served as an extremely useful tool with `nix-build` in conjunction with `--no-out-link`. Two extremely common use cases:

- `$(nix-build ...)/bin/foo`
- `nix-store -qR $(nix-build ...)`

It's just nice to interpolate the path without having to copy paste it, and to make sure that the command you're running definitely has the latest version of the nix build output. It would be nice to have this back with `nix build`.

ElvishJerricco:
Trying to script around this doesn't seem viable because of GC roots. You can't isolate what the outputs of `nix build` were without putting them in an isolated temp dir, or else you can't know if `result-100` was from this build or a previous one. And if you put them there, then moving them where the user wanted will leave dangling symlinks in the GC roots. This is *technically* fine, but writing a script that inherently leaves behind unwanted dangling symlinks sounds bad.

dtzWill:
Technically speaking I think the pattern `nix-store -qR $(nix-build ...)` was always slightly broken in that once the `nix-build` command prints out the path and exits the path could be GC'd before `nix-store` parses its arguments and locks the path or whatever it does.

It's not a problem if you can be sure a GC isn't running concurrently or something, but just a thought.

Couldn't the "moving them where the user wanted" be done with `nix-store --add-root $newlocation --indirect /nix/store/path`?  Not sure I understand what you're doing re:users though :).

ElvishJerricco:
> Couldn't the "moving them where the user wanted" be done with nix-store --add-root $newlocation --indirect /nix/store/path?

@dtzWill This would still leave dangling symlinks in your gc roots. Again, not *technically* a problem, but definitely undesirable to have a command be guaranteed to leave behind unwanted symlinks.

viric:
I'm with @ElvishJerricco; I used the stdout a lot.     Even "nix build --no-link" does not output anything, so no reference at all to the outcome.

If someone wants to make the $(nix-build ) thing GC-safe, that can use a mktemp as symlink and then delete it. Annoying to impose this, though.

ElvishJerricco:
@viric I did figure out how to make such a thing, but it uses `nix-store --delete` (no argument) to wipe *all* hanging symlinks. I'd prefer to only wipe the ones created by the command, but this is somewhat reasonable.

https://gist.github.com/ElvishJerricco/e8193a5b4734877bc492e870625bc92f

[...]

PR #2423, opened 2018-09-14
nix build: Print result paths to stdout with --no-link

ElvishJerricco:
Helps with #1930

### DIFF
--- a/src/nix/build.cc
+++ b/src/nix/build.cc
@@ -20,7 +20,7 @@ struct CmdBuild : MixDryRun, InstallablesCommand
 
         mkFlag()
             .longName("no-link")
-            .description("do not create a symlink to the build result")
+            .description("print result paths to stdout instead of creating a symlink")
             .set(&outLink, Path(""));
     }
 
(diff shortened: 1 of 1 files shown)


ElvishJerricco:
I've been using this PR on my machine for a while now by applying the patch on the system's Nix package. It's been really useful so far, so I'd love to see this merged.

But I've had one problem with it, at least with Nix 2.1.1. When Nix outputs `querying info about missing paths`, it ends up looking like:

```
$ nix build --no-link nixpkgs.ghc
querying info about missing paths/nix/store/n5i1zdpmk2b1s3z96649xh8f9kr3g96s-ghc-8.4.3
```

Now, that's not on stdout, so it has no functional issue.

```
$ nix build --no-link nixpkgs.ghc | cat
/nix/store/n5i1zdpmk2b1s3z96649xh8f9kr3g96s-ghc-8.4.3
```

But it is rather annoying. Anyone know what I should do about this?

ElvishJerricco:
Alright, fixed it by clearing stderr first.

grahamc:
This is unsafe, as it leaves a race condition of the build finishing, GC, and the path being used.

7c6f434c:
I do think this race condition is manageable in most cases, as auto-GC is probably used only on a minority of deployments, but of course it would be even nicer to have some orthogonality, and maybe just the approach of always printing as in #2622 is the best.

ElvishJerricco:
@grahamc I don't think that's solvable. Once our process exits, we have no way to guarantee to the user that the paths will remain in the store without a link. Personally I consider this an edge case. Since I find this feature to be mostly a convenience for when I'm messing around interactively, I'd much rather see that listed as a warning in the `--help` text than used as a reason not to merge something like this.

xaverdh:
> This is unsafe, as it leaves a race condition of the build finishing, GC, and the path being used.

Using `symlinks` in the current directory (the default behaviour) is also racy. If two builds are running with the same `cwd`, one might overwrite the `symlink` from the other, before the caller reads it. Using temporary directories slightly improves the situation, assuming these are sufficiently "protected" (e.g. by a private mount namespace or such).

xaverdh:
> (e.g. by a private mount namespace or such)

which will not work actually, now that I think of it.. the daemon needs to be able to see the `symlink`

Issue #2421, opened 2018-09-14
Remove value for failed tryEval?


lheckemann:
Currently, `tryEval` will insert the fairly arbitrary value `false` for failed evaluations:
```
nix-repl> builtins.tryEval 5
{ success = true; value = 5; }

nix-repl> builtins.tryEval (throw "nope")
{ success = false; value = false; }
```

The most common use case (I'm guessing) for tryEval is to provide an alternative value should evaluation fail. This is rather cumbersome with its current behaviour:

```
nix-repl> let result = builtins.tryEval 5; in if result.success then result.value else "failed"
5

nix-repl> let result = builtins.tryEval (throw "nope"); in if result.success then result.value else "failed"
"failed"
```

It would be a lot more convenient thanks to the `or` keyword (and less arbitrary) not to have the value key at all for unsuccessful evaluation:

```
nix-repl> newTryEval = x: let r = builtins.tryEval x; in if r.success then r else { success = false; }

nix-repl> (newTryEval 5).value or "failed"
5

nix-repl> (newTryEval (throw "nope")).value or "failed"
"failed"
```

Some further thoughts:
- with the suggested change, the success key isn't really necessary any more and could be replaced by `(builtins.tryEval foo) ? value`
- an improved tryEval could potentially include the error message or other details to make it more versatile

Infinisil:
> an improved tryEval could potentially include the error message or other details to make it more versatile

+1 to that

But how about adding a more proper error handling machanism:

```nix
builtins.catch (throw "foo") (e: "Got error: ${e}") -> "Got error: foo"
builtins.catch "foo" (e: "Got error: ${e}") -> "foo"
```

Would be much more ergonomic than `tryEval`.

copumpkin:
I think a big part of the awkward design is attempting to not break purity too badly. Agree that `false` should probably go away in favor of `or` but a more generic `catch` will need to be looked at very carefully and weighed against the benefits of purity.

7c6f434c:
For the record continuity: https://github.com/NixOS/nix/pull/1000#issuecomment-235872787

I do agree that with `or` and `?` it is a good idea to drop `value` from failed attempts, the after the old Nix version falls out of support we can decide on dropping `success`.

lheckemann:
Since it's a fairly trivial change, I've opened #2424 implementing just the "remove value on failure" change.

7c6f434c:
[...]

PR #2424, opened 2018-09-14
tryEval: only provide value on success

lheckemann:
See #2421

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -453,7 +453,6 @@ static void prim_tryEval(EvalState & state, const Pos & pos, Value * * args, Val
         v.attrs->push_back(Attr(state.sValue, args[0]));
         mkBool(*state.allocAttr(v, state.symbols.create("success")), true);
     } catch (AssertionError & e) {
-        mkBool(*state.allocAttr(v, state.sValue), false);
         mkBool(*state.allocAttr(v, state.symbols.create("success")), false);
     }
     v.attrs->sort();
--- a/tests/lang/eval-okay-tryeval.exp
+++ b/tests/lang/eval-okay-tryeval.exp
@@ -1,1 +1,1 @@
-{ x = { success = true; value = "x"; }; y = { success = false; value = false; }; z = { success = false; value = false; }; }
+{ x = { success = true; value = "x"; }; y = { success = false; }; z = { success = false; }; }


lheckemann:
@edolstra what do you think of this?

Synthetica9 (approved):
Makes sense! Enables behaviour like

```
(builtins.tryEval (assert (1 == 2); 3)).value or "foobar"
```

(currently this gives `false`, after this change it would be `"foobar"`)

Artturin:
> ```
> (builtins.tryEval (assert (1 == 2); 3)).value or "foobar"
> ```
> 
can be done like `let x = builtins.tryEval (assert (1 == 2); 3); in if x.success then x.value else "foobar"` rn but after this pr it would be better

before this is merged we'd have to change nixpkgs

```
$ rg "tryEval.+value"
nixos/maintainers/option-usages.nix
106:          { name = showOption opt.loc; } // builtins.tryEval (strict opt.value));

lib/tests/misc.nix
345:    ( builtins.tryEval (toInt "") == { success = false; value = false; } )
346:    ( builtins.tryEval (toInt "123 123") == { success = false; value = false; } )
347:    ( builtins.tryEval (toInt "0 123") == { success = false; value = false; } )
348:    ( builtins.tryEval (toInt " 0d ") == { success = false; value = false; } )
349:    ( builtins.tryEval (toInt " 1d ") == { success = false; value = false; } )
350:    ( builtins.tryEval (toInt " d0 ") == { success = false; value = false; } )
351:    ( builtins.tryEval (toInt "00") == { success = false; value = false; } )
352:    ( builtins.tryEval (toInt "01") == { success = false; value = false; } )
353:    ( builtins.tryEval (toInt "002") == { success = false; value = false; } )
354:    ( builtins.tryEval (toInt " 002 ") == { success = false; value = false; } )
355:    ( builtins.tryEval (toInt " foo ") == { success = false; value = false; } )
356:    ( builtins.tryEval (toInt " foo 123 ") == { success = false; value = false; } )
357:    ( builtins.tryEval (toInt " foo123 ") == { success = false; value = false; } )
389:    ( builtins.tryEval (toIntBase10 "") == { success = false; value = false; } )
390:    ( builtins.tryEval (toIntBase10 "123 123") == { success = false; value = false; } )
391:    ( builtins.tryEval (toIntBase10 "0 123") == { success = false; value = false; } )
392:    ( builtins.tryEval (toIntBase10 " 0d ") == { success = false; value = false; } )
393:    ( builtins.tryEval (toIntBase10 " 1d ") == { success = false; value = false; } )
394:    ( builtins.tryEval (toIntBase10 " d0 ") == { success = false; value = false; } )
395:    ( builtins.tryEval (toIntBase10 " foo ") == { success = false; value = false; } )
396:    ( builtins.tryEval (toIntBase10 " foo 123 ") == { success = false; value = false; } )
397:    ( builtins.tryEval (toIntBase10 " foo 00123 ") == { success = false; value = false; } )
398:    ( builtins.tryEval (toIntBase10 " foo00123 ") == { success = false; value = false; } )

nixos/modules/config/terminfo.nix
19:    # can be generated with: filter (drv: (builtins.tryEval (drv ? terminfo)).value) (attrValues pkgs)

pkgs/top-level/splice.nix
59:          inherit (builtins.tryEval value0) success value;

pkgs/top-level/impure.nix
10:  try = x: def: let res = builtins.tryEval x; in if res.success then res.value else def;
```

it'd also be difficult to support old nix versions too so i think this can't be merged

Ericson2314:
I think this would be a good idea *if* we had a language versioning story.

roberth:
This would make for a good addition to Nixpkgs lib.

Small but breaking changes like this are not worth the cost of breaking old expressions.



> a good idea _if_ we had a language versioning story.

Nixpkgs lib is a decent story for "replacing" builtins.
We'd only really need a language version for changing the syntax, or the meaning of operators. Individual builtins, while built in, are not something you'd have to use, so we have little reason to break those.

PR #2428, opened 2018-09-17
Document Aliases for keep-outputs, keep-derivations

grahamc:
Please also backport to 2.1.


edolstra:
I'm not sure we *should* document these, being deprecated and all. There are many more aliases BTW.

grahamc:
I think they should be documented as being aliased now, so people can find the relevant docs from their configuration. I've seen several people confused about undocumented (because they're aliased) options. Maybe not as a second `<term>` in the listing, but as a paragraph at the end of the entry?

grahamc:
Another option could be documenting them as separate options, with the body of the varlistentry being "Deprecated: now an alias to newthing"?

grahamc:
I've now documented all aliases which exist, where the option the alias is for was already documented.

Issue #2410, opened 2018-09-09
Bad darwin reinstallation experience


infinisil:
Having a mac I thought of using it to fix some darwin failures, so I went to install Nix. Turns out that I had an older version of Nix installed already, but this should work just fine, optimally. Here I describe all the problems that I happen to encounter.

1. I go to https://nixos.org/nix/ and copy the command to install Nix `curl https://nixos.org/nix/install | sh`
2. Paste it into my terminal and run it (some line breaks inserted for readability):
```bash
$ curl https://nixos.org/nix/install | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2472  100  2472    0     0  11644      0 --:--:-- --:--:-- --:--:-- 11660
downloading Nix 2.1.1 binary tarball for x86_64-darwin from 'https://nixos.org/releases/nix/nix-2.1.1/nix-2.1.1-x86_64-darwin.tar.bz2' to '/var/folders/pv/g5wjzg293v50gg59wmjz0fdw0000gn/T/nix-binary-tarball-unpack.XXXXXXXXXX.014RNLvf'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 28.8M  100 28.8M    0     0  1644k      0  0:00:17  0:00:17 --:--:-- 1553k
Note: a multi-user installation is possible. See https://nixos.org/nix/manual/#sect-multi-user-installation
performing a single-user installation of Nix...
/var/folders/pv/g5wjzg293v50gg59wmjz0fdw0000gn/T/nix-binary-tarball-unpack.XXXXXXXXXX.014RNLvf/unpack/nix-2.1.1-x86_64-darwin/install:
directory /nix exists, but is not writable by you.
This could indicate that another user has already performed a single-user installation of Nix on this system.
If you wish to enable multi-user support see http://nixos.org/nix/manual/#ssec-multi-user.
If you wish to continue with a single-user install for infinisil please run 'chown -R infinisil /nix' as root.
```
[...]

PR #2429, opened 2018-09-17
Upgrading docs: easier to copy + link to issue

ariutta:
With `command`, it's easy for users to accidentally copy the period after the command.

### DIFF
--- a/doc/manual/installation/upgrading.xml
+++ b/doc/manual/installation/upgrading.xml
@@ -8,7 +8,7 @@
 
   <para>
     Multi-user Nix users on macOS can upgrade Nix by running this:
-    <command>$ sudo -i sh -c 'nix-channel --update &amp;&amp; nix-env
+    <command>sudo -i sh -c 'nix-channel --update &amp;&amp; nix-env
       -iA nixpkgs.nix'; sudo launchctl stop org.nixos.nix-daemon; sudo
       launchctl start org.nixos.nix-daemon</command>
   </para>
--- a/doc/manual/installation/upgrading.xml
+++ b/doc/manual/installation/upgrading.xml
@@ -7,15 +7,21 @@
   <title>Upgrading Nix</title>
 
   <para>
-    Multi-user Nix users on macOS can upgrade Nix by running
-    <command>sudo -i sh -c 'nix-channel --update &amp;&amp; nix-env
-    -iA nixpkgs.nix'; sudo launchctl stop org.nixos.nix-daemon; sudo
-    launchctl start org.nixos.nix-daemon</command>.
+    Multi-user Nix users on macOS can upgrade Nix by running this:
+    <screen>$ sudo -i sh -c 'nix-channel --update &amp;&amp; nix-env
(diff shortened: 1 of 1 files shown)


ariutta:
This is relevant to #2410.

doc/manual/installation/upgrading.xml
```diff
@@ -7,15 +7,21 @@
...
 
   <para>
-    Multi-user Nix users on macOS can upgrade Nix by running
-    <command>sudo -i sh -c 'nix-channel --update &amp;&amp; nix-env
-    -iA nixpkgs.nix'; sudo launchctl stop org.nixos.nix-daemon; sudo
-    launchctl start org.nixos.nix-daemon</command>.
+    Multi-user Nix users on macOS can upgrade Nix by running this:
+    <screen>$ sudo -i sh -c 'nix-channel --update &amp;&amp; nix-env
+      -iA nixpkgs.nix'; sudo launchctl stop org.nixos.nix-daemon; sudo
+      launchctl start org.nixos.nix-daemon</screen>
```
grahamc: We should keep `<command>` but I don't mind dropping the period, and switching to the `running this:` part.

doc/manual/installation/upgrading.xml
```diff
@@ -7,15 +7,21 @@
...
+  <para>
+    Single-user installations of Nix should run this:
+    <screen>$ nix-channel
+      --update; nix-env -iA nixpkgs.nix</screen>
   </para>
 
   <para>
-    Single-user installations of Nix should run <command>nix-channel
-    --update; nix-env -iA nixpkgs.nix</command>.
+    For more information, see <link xlink:href="https://github.com/NixOS/nix/issues/2410">Issue #2410</link>.
```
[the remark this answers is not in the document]
ariutta: Git mangled the merge here. I agree they should be separate.

doc/manual/installation/upgrading.xml
```diff
@@ -7,15 +7,21 @@
   <title>Upgrading Nix</title>
 
   <para>
-    Multi-user Nix users on macOS can upgrade Nix by running
+    Multi-user Nix users on macOS can upgrade Nix by running this:
     <command>sudo -i sh -c 'nix-channel --update &amp;&amp; nix-env
-    -iA nixpkgs.nix'; sudo launchctl stop org.nixos.nix-daemon; sudo
-    launchctl start org.nixos.nix-daemon</command>.
+      -iA nixpkgs.nix'; sudo launchctl stop org.nixos.nix-daemon; sudo
+      launchctl start org.nixos.nix-daemon</command>
```
LnL7: Does this actually reload the service and start the new version? I think this might explain an upgrade issue I noticed yesterday.

ariutta: Yes, that's what it's supposed to do. I've heard of two issues:
1) If you have a package installed that depends on `nix` version 1.x, the updated version 2.x might not take over.
2) For a multi-user installation, you shouldn't install `nix` as the current user.

grahamc:
I included much of this in PR #2434, thank you

PR #2432, opened 2018-09-19
SSL certificate search failed to find user profile certificates.

luke-clifton:


luke-clifton:
@grahamc you seem best positioned to review this. There was a slight bug in the SSL CA search path.

edolstra:
Looks good to me, but maybe it would be even better to search `$NIX_PROFILES`? E.g.
```
for i in $NIX_PROFILES; do
  if [ -e $i/etc/ssl/certs/ca-bundle.crt ]; then
    export NIX_SSL_CERT_FILE=$i/etc/ssl/certs/ca-bundle.crt
    break
  fi
done
```

luke-clifton:
@edolstra I did originally consider that, but didn't want to introduce extra semantics on the order of the profiles because

1. I was unsure if there was an existing semantics
2. Without changing the order I thought it would be counter-intuitive (last one wins, unlike `$PATH`)
3. I really didn't want to change the order because I didn't know what relied on it.

I just did a quick survey of nixpkgs, and there are numerous locations where the order is important, and oddly enough, in many cases, the first thing they do is reverse the list (see the listing at the end).

I've just pushed a commit that searches the path, preferring the last found profile (this agrees with existing use cases, but is counter to your snippet above, and, in my opinion, counter intuitive).

#### Obvious cases of reversing the `NIX_PROFILE`

```
nixpkgs/nixos/modules/programs/environment.nix:         export NIX_PROFILES="${concatStringsSep " " (reverseList cfg.profiles)}"
nixpkgs/pkgs/applications/editors/emacs/site-start.el:  (reverse (split-string (or (getenv "NIX_PROFILES") ""))))
nixpkgs/pkgs/applications/editors/vim/configurable.nix:      for p in reverse(split($NIX_PROFILES))
nixpkgs/pkgs/shells/fish/default.nix:      set -l __nix_profile_paths (echo $NIX_PROFILES | ${coreutils}/bin/tr ' ' '\n')[-1..1]
```

There are more cases that reference the variable (though, only ~30), I didn't audit them all for order dependencies.

edolstra:
Thanks!

PR #2434, opened 2018-09-19
Upgrade docs: improve the upgrade command

grahamc:
and make it more copy-pastable.

Includes the changes from #2429

### DIFF
--- a/doc/manual/installation/upgrading.xml
+++ b/doc/manual/installation/upgrading.xml
@@ -7,15 +7,16 @@
   <title>Upgrading Nix</title>
 
   <para>
-    Multi-user Nix users on macOS can upgrade Nix by running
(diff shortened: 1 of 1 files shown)


grahamc:
image

grahamc:
I've asked @LnL7 to review this before merging.

LnL7:
Another thing that might be worth adding are some instructions on how to recover a daemon installation after performing a system upgrade. Apple will reset `/etc/bashrc` and friends as well as clear out `/var/root`, removing the .nix-channels entries for root.

matthewbauer:
Great work!

One note:

After https://github.com/NixOS/nix/commit/74f6d8767de13980b0e83a249633ee4a7e6fc547, it should no longer be necessary to add ```-i``` to sudo for each command. However, since this concerns upgrading it's probably safest to leave it in.

Issue #1990, opened 2018-03-19
offline binary cache entirely breaks nixos-rebuild


cleverca22:
```
[root@system76:~]# nixos-rebuild switch --fallback
building Nix...
building the system configuration...
warning: unable to download 'http://amd.localnet:5000/mz0ga4ad618mbiwk3k2029ja1anilma6.narinfo': Couldn't connect to server (7); retrying in 306 ms
...
warning: unable to download 'http://amd.localnet:5000/ychm97sxx69zj9llm76yi386a5g9p4sm.narinfo': Couldn't connect to server (7); retrying in 19324 ms
error: unexpected end-of-file

[root@system76:~]# 
```
the only way to rebuild the machine now, is to manually change `nix.conf` to exclude that binary cache

Taeradan:
"the only way to rebuild the machine now, is to manually change nix.conf to exclude that binary cache"
How do you do that ? It's read-only, even for root, and it's getting me mad not being able to change anything on my computer anymore.....

A really blocking issue.

Taeradan:
Note that `nixos-rebuild switch --option binary-caches ""`  has no effect, it is still trying to use the binary caches in nix.conf.

jpierre03:
here nixos-rebuild switch --fallback  --show-trace --option binary-caches https://cache.nixos.org/  forced official cache and workded

cleverca22:
when i manually change `nix.conf`, i replace the symlink with a file, so the store is never modified directly
`setup-etc.pl` will revert it back to a symlink at the next switch

mpickering:
@jpierre03's answer worked for me but this is still a bad issue.

matthewbauer:
This is definitely annoying. It definitely isn't intentional behavior is it? The down host should be a warning (as the message indicates) not an error.

/cc @edolstra @shlevy

domenkozar:
It hasn't happened yet that https://cachix.org was offline (higher chance than high-availability cloudfront), so this one will be a pain one day if we don't get it solved.

dtzWill:
> It hasn't happened yet that https://cachix.org was offline (higher chance than highly-availabile cloudfront), so this one will be a pain one day if we don't get it solved.

https://github.com/cachix/cachix/issues/117

Not that you need another ping on the subject, and thanks as always for cachix :grin: but as predicted this issue means any machines using cachix are unable to do anything :(

Hmm...

chris-martin:
[...]

PR #2441, opened 2018-09-22
Always try next substituter

dasJ:
This is related to #1990.
The current behaviour is counter-intuitive and conflicts with the
documentation of --fallback which states that the option is used to
build derivations locally if they fail to download.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -3683,18 +3683,12 @@ void SubstitutionGoal::tryNext()
         tryNext();
         return;
     } catch (SubstituterDisabled &) {
-        if (settings.tryFallback) {
-            tryNext();
-            return;
-        }
-        throw;
+        tryNext();
+        return;
     } catch (Error & e) {
-        if (settings.tryFallback) {
-            printError(e.what());
-            tryNext();
-            return;
-        }
-        throw;
+        printError(e.what());
+        tryNext();
+        return;
     }
 
     /* Update the total expected download size. */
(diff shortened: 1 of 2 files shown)


dasJ:
@edolstra Any update on this? The current behaviour is kind of annoying.

dasJ:
@edolstra Ping again

grahamc:
This PR seems to gut the functionality of `--fallback`, as it seems to delete all the code where that option would do anything. Am I right about that?

dasJ:
Not sure if the option does nothing after that. From `--help`, this it what it's supposed to do:

```
       --fallback
           Whenever Nix attempts to build a derivation for which substitutes are known for each output path, but realising the output paths through the substitutes fails, fall back on building the derivation.

           The most common scenario in which this is useful is when we have registered substitutes in order to perform binary distribution from, say, a network repository. If the repository is down, the realisation of the derivation
           will fail. When this option is specified, Nix will build the derivation instead. Thus, installation from binaries falls back on installation from source. This option is not the default since it is generally not desirable
           for a transient failure in obtaining the substitutes to lead to a full build from source (with the related consumption of resources).
```

And building != using the next cache IMO

grahamc:
Can you look in to answering this question a bit more conclusively, to ensure the expected behavior of this flag is retained.

dasJ:
It seems, it's used to [return early](/nixos/nix/tree/master/src/libstore/build.cc#L1209). Otherwise, the code (my C++ isn't too good) seems to [retry](/nixos/nix/tree/master/src/libstore/build.cc#L1256)

dasJ:
I'm sick of this

PR #2443, opened 2018-09-23
Add XDG_DATA_DIRS to nix-profile.sh

matthewbauer:
This is helpful for locating resources with Nix, especially in
unmanaged situations like on macOS or non-NixOS Linux. ~Like MANPATH,
it should be harmless to have these set when the directory doesn’t
exist. If anyone knows of other variables like this, please mention them
here! We should include as many as possible so things work out-of-the-box.~

### DIFF
--- a/scripts/nix-profile.sh.in
+++ b/scripts/nix-profile.sh.in
@@ -78,6 +78,7 @@ if [ -n "$HOME" ] && [ -n "$USER" ]; then
     if [ -n "${MANPATH}" ]; then
         export MANPATH="$NIX_LINK/share/man:$MANPATH"
     fi
+    export XDG_DATA_DIRS="$NIX_LINK/share:$XDG_DATA_DIRS"
 
     export PATH="$NIX_LINK/bin:$__savedpath"
     unset __savedpath NIX_LINK NIX_USER_PROFILE_DIR NIX_PROFILES


edolstra:
No, I feel that Nix on non-NixOS systems should pollute the user's environment as little as necessary, to minimize the probability that installing a Nix package breaks the host system. (E.g. setting `MANPATH` is *not* harmless: it overrides `man`'s builtin default of searching directories relative to `PATH`.)

matthewbauer:
The most common need for this is with qt5 apps where qtbase is in propagated-user-env-pkgs. The qt5 runtime can't find the plugins provided by qt5 base, even when qtbase is installed. It needs to be able to find these plugins in XDG_DATA_DIRS. This leads to a lot of apps in Nixpkgs being nonfunctional to new users. See issue https://github.com/NixOS/nixpkgs/issues/42893.

jackTaw88:
When this path will b merged to release?

rnhmjoj:
> The qt5 runtime can't find the plugins provided by qt5 base, even when qtbase is installed.

Is this still an issue after the wrapQtAppsHook changes? (Assuming every app has been wrapped)

matthewbauer:
> Is this still an issue after the wrapQtAppsHook changes? (Assuming every app has been wrapped)

Not so much, although you can still run into situations when setting XDG_DATA_DIRS helps. Qt is not so important though.

PR #2448, opened 2018-09-25
nix-daemon: allow setting builders to "" by any user (untrusted)

dtzWill:
It'd be even better to accept a subset of builders, but this is still
a big win IMHO for making it easy to disable distributed builds from normal accounts.

### DIFF
--- a/src/nix-daemon/nix-daemon.cc
+++ b/src/nix-daemon/nix-daemon.cc
@@ -557,7 +557,8 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
                     ;
                 else if (trusted
                     || name == settings.buildTimeout.name
-                    || name == "connect-timeout")
+                    || name == "connect-timeout"
+                    || (name == "builders" && value == ""))
                     settings.set(name, value);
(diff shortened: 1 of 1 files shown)


src/nix-daemon/nix-daemon.cc
```diff
@@ -557,7 +557,8 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
                     ;
                 else if (trusted
                     || name == settings.buildTimeout.name
-                    || name == "connect-timeout")
+                    || name == "connect-timeout"
+                    || (name == "builders" && value == ""))
```
grahamc: @dtzWill would you mind sending a PR adding docs for this?

dtzWill: Don't mind at all, but not sure how to edit the docs.  Will take a look at that later today, probably should bite the bullet! If you find yourself wondering when I'm going to do it please bug me about it ;).

But sure thing, and thanks for all your documentation work! (including ensuring documentation exists for new things, like here!)

PR #2452, opened 2018-09-26
Fix overflow when verifying signatures of content addressable paths

### LINKED ISSUE #2451: Content addressable paths cannot be verified if you sign them with a trusted key
This is especially problematic as it means `nix sign-paths --all` will *always* break `nix verify --all`, which seems counter to the point of `sign-paths --all`!

```bash
$ nix-store --generate-binary-cache-key foo.bar.org-1 foo.bar.org-1.pem foo.bar.org-1.pub
$ touch foo.txt
$ path=$(nix add-to-store ./foo.txt)
$ nix sign-paths $path -k ./foo.bar.org-1.pem
$ nix path-info --json $path | jq .
[
  {
    "path": "/nix/store/kvf1a222z1fdlwl2s5gyk7biz6cfb7n7-foo.txt",
    "narHash": "sha256:0ip26j2h11n1kgkz36rl4akv694yz65hr72q4kv4b3lxcbi65b3p",
    "narSize": 112,
    "references": [],
    "ca": "fixed:r:sha256:0ip26j2h11n1kgkz36rl4akv694yz65hr72q4kv4b3lxcbi65b3p",
    "registrationTime": 1537916813,
    "signatures": [
      "foo.bar.org-1:2fnxoUUrs5u4UOL3ymNvcMlRX39MySZsIjMQLT4r9nxigqkbrzaORZi5WohTxG3zRIeoSlWxUg9+w5dvyBFNCw=="
    ]
  }
]
$ nix verify -n 1 $path
[1 paths verified]
$ nix verify -n 1 --trusted-public-keys $(cat foo.bar.org-1.pub) $path
path '/nix/store/kvf1a222z1fdlwl2s5gyk7biz6cfb7n7-foo.txt' is untrusted
[1 paths verified, 1 untrusted]
```

Note that the path is perfectly trustworthy if you *don't* add the public key to `trusted-public-keys`, because it's a content addressable path. Adding the key but not signing the path obviously works too.

ElvishJerricco:
Fixes #2451

### DIFF
--- a/src/nix/verify.cc
+++ b/src/nix/verify.cc
@@ -120,7 +120,7 @@ struct CmdVerify : StorePathsCommand
                             for (auto sig : sigs) {
                                 if (sigsSeen.count(sig)) continue;
                                 sigsSeen.insert(sig);
-                                if (info->checkSignature(publicKeys, sig))
+                                if (validSigs < ValidPathInfo::maxSigs && info->checkSignature(publicKeys, sig))
                                     validSigs++;
                             }
                         };
--- a/tests/signing.sh
+++ b/tests/signing.sh
@@ -62,6 +62,10 @@ outPathCA=$(IMPURE_VAR1=foo IMPURE_VAR2=bar nix-build ./fixed.nix -A good.0 --no
 nix verify $outPathCA
 nix verify $outPathCA --sigs-needed 1000
 
+# Check that signing a content-addressed path doesn't overflow validSigs
+nix sign-paths --key-file $TEST_ROOT/sk1 $outPathCA
+nix verify -r $outPathCA --sigs-needed 1000 --trusted-public-keys $pk1
+
 # Copy to a binary cache.
 nix copy --to file://$cacheDir $outPath2
 


ElvishJerricco:
I suppose I should add a test for this.

edolstra:
Actually we can just stop verifying sigs when we know that it's a content-addressable path, i.e. change
```
if (info2->isContentAddressed(*store)) validSigs = ValidPathInfo::maxSigs;
doSigs(info2->sigs);
```
to
```
if (info2->isContentAddressed(*store)) 
  validSigs = ValidPathInfo::maxSigs;
else
  doSigs(info2->sigs);
```

ElvishJerricco:
@edolstra I considered adding `info->isContentAddressed(*store)` to this if statement:

https://github.com/NixOS/nix/blob/b7091ce41e0073b147678f05b9a620b9bb2903c1/src/nix/verify.cc#L110

But I figured there might be a reason that we also ask substituters whether their info is content addressed. Like maybe somehow someone's metadata is screwed up and dropped the fact that the path is content addressed.

ElvishJerricco:
@edolstra Is that reason to keep the changes as is, or would you rather just call the path good if it's content-addressed?

ElvishJerricco:
@edolstra ping.

PR #2454, opened 2018-09-26
bump base nixpkgs used by default 18.03 -> 18.09

dtzWill:
Not sure when the switch should happen,
but with this it's ready when the time comes! :)


edolstra:
Weren't there some issues with glibc locales changing in an incompatible way?

dtzWill:
Oh right, hrmm.  Well not sure what to do about that-- is there an issue tracking a fix or is this just the way of things?

Issue #2034, opened 2018-04-01
Vim in nix-shell starts non-interactive bash on :shell and :terminal


pandaman64:
# How to reproduce

Run `nix-shell` with this `shell.nix`.
```
with import <nixpkgs> {};
stdenv.mkDerivation {
  name = "vim_shell_in_nix-shell";
  buildInputs = [
    vim
    bashInteractive
  ];
}
```
Then start `vim` and type `:shell`.

# What I get

Vim starts non-interactive bash (shell without arrow-keys support, tab completion, a weird prompt like `\[\][pan@pan-nixos:~/vim-shell-nix]$\[\]   `). 

You can type `:terminal` instead of `:shell`. This command opens a new pane with non-interactive bash.

# Other information

In vim, `set shell` shows vim is using this shell: `shell=/nix/store/q1g0rl8zfmz7r371fp5p42p4acmv297d-bash-4.4-p19/bin/bash`, which is a non-interactive bash.
```
[pan@pan-nixos:~/vim-shell-nix]$ /nix/store/q1g0rl8zfmz7r371fp5p42p4acmv297d-bash-4.4-p19/bin/bash

\[\][pan@pan-nixos:~/vim-shell-nix]$\[\] 
```

This issue looks like #730 though `bash` in `nix-shell` works well thanks to `bashInteractive`.

Here is my `nix-info -m`:
```
 - system: `"x86_64-linux"`                                                                                                                                                                                                                  
 - host os: `Linux 4.14.29, NixOS, 18.09pre133621.56fb68dcef4 (Jellyfish)`                                                                                                                                                                   
 - multi-user?: `yes`                                                                                                                                                                                                                        
 - sandbox: `no`                                                                                                                                                                                                                             
 - version: `nix-env (Nix) 2.0`                                                                                                                                                                                                              
 - channels(pan): `"nixpkgs-18.09pre133932.ee28e35ba37"`                                                                                                                                                                                     
 - channels(root): `"nixos-18.09pre133621.56fb68dcef4"`                                                                                                                                                                                      
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos/nixpkgs`
```

pandaman64:
It turns out that `$SHELL` is set to non-interactive bash and Vim uses `$SHELL` for its shell.

This command gave me an interactive bash:
``` 
$ SHELL=bash vim
```

I'm not sure it's desired behavior that `$SHELL` is non-interactive.

[...]

Issue #644, opened 2015-09-14
nix-shell escaping broken


Profpatsch:
(current `unstable` branch of NixOS)

In each nix-shell I start, the escaping is broken:

```
\[\][nix-shell:~/code/openlab/labapi]$\[\]  
```

Arrow up and down consequently give:

```
 ^[[A^[[B
```

and backspacing destroys the cursor position.

lethalman:
Are you sure you aren't already in a nix-shell?

Profpatsch:
Yes, from normal shell.
I created the minimal derivation,

```
with import <nixpkgs> {};

stdenv.mkDerivation {
    name = "foobar";
}
```

switched to `sh` and called `nix-shell`, and it still happens.

Profpatsch:
Is it normal that starting a shell inside another shell breaks the arrow keys?

Profpatsch:
I tried another terminal emulator, same result.

Profpatsch:
Happens on tty, too.

lethalman:
Yes because the first nix-shell uses the bash from your profile, which is interactive (with readline). The nix-shell however inside will probably have a non-interactive bash from stdenv. Hence the second nix-shell will spawn the non-interactive bash taken from the PATH.

Profpatsch:
I’m not spawning to nix-shell inside each other, though.

lethalman:
Then you have a non-interactive bash in PATH somewhere.

Profpatsch:
htop shows the following:

```
lilyterm
|- /run/current-system/sw/bin/bash
  |- bash --rc-file /run/user/1000/nix-shell.x9URzx/rc

wmertens:
Check that you didn't write bash anywhere in your config, you should write
bashInteractive there.

On Mon, Sep 14, 2015, 12:46 Profpatsch <notifications@github.com> wrote:

> htop shows the following:
>
> lilyterm
> |- /run/current-system/sw/bin/bash
>   |- bash --rc-file /run/user/1000/nix-shell.x9URzx/rc
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NixOS/nix/issues/644#issuecomment-140035279>.
>
-- 

Wout.
(typed on mobile, excuse terseness)

Profpatsch:
@wmertens You mean `users.extraUsers.user.shell` should not be `bash`?

What’s up with the file missing?

wmertens:
[...]

PR #2455, opened 2018-09-27
nix-shell: add bashInteractive to the start of the PATH, set SHELL

grahamc:
Tools which re-exec `$SHELL` or `$0` or `basename $SHELL` or even just
`bash` will otherwise  get the non-interactive bash, providing a
broken shell for the same reasons described in
https://github.com/NixOS/nixpkgs/issues/27493.

Extends c94f3d5575d7af5403274d1e9e2f3c9d72989751

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -417,16 +417,20 @@ void mainWrapped(int argc, char * * argv)
                 "dontAddDisableDepTrack=1; "
                 "[ -e $stdenv/setup ] && source $stdenv/setup; "
                 "%3%"
+                "PATH=\"%4%:$PATH\"; "
+                "SHELL=%5%; "
                 "set +e; "
                 R"s([ -n "$PS1" ] && PS1='\n\[\033[1;32m\][nix-shell:\w]\$\[\033[0m\] '; )s"
                 "if [ \"$(type -t runHook)\" = function ]; then runHook shellHook; fi; "
                 "unset NIX_ENFORCE_PURITY; "
                 "shopt -u nullglob; "
-                "unset TZ; %4%"
-                "%5%",
+                "unset TZ; %6%"
+                "%7%",
                 (Path) tmpDir,
                 (pure ? "" : "p=$PATH; "),
                 (pure ? "" : "PATH=$PATH:$p; unset p; "),
+                dirOf(shell),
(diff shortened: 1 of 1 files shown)


grahamc:
cc @scoates who found the issue

grahamc:
Closes #2034 and #644

src/nix-build/nix-build.cc
```diff
@@ -417,16 +417,19 @@ void mainWrapped(int argc, char * * argv)
                 "dontAddDisableDepTrack=1; "
                 "[ -e $stdenv/setup ] && source $stdenv/setup; "
                 "%3%"
+                "PATH=\"$(dirname %4%):$PATH\"; "
```
edolstra: This can be `PATH=%N%` where N = `dirOf(shell)`. I.e. compute the directory in C++, saving an external call.

edolstra:
Arguably this reduces `nix-shell` purity, since you don't get the shell specified by the build environment. Probably not a big deal...

src/nix-build/nix-build.cc
grahamc: Ah, nice! Fixed!

grahamc:
It does, and it is a bit annoying that it makes it harder to override the shell. I'm not aware of a nice alternative which also keeps things consistently working.

src/nix-build/nix-build.cc
```diff
@@ -417,16 +417,19 @@ void mainWrapped(int argc, char * * argv)
                 "dontAddDisableDepTrack=1; "
                 "[ -e $stdenv/setup ] && source $stdenv/setup; "
                 "%3%"
+                "PATH=\"%4%:$PATH\"; "
+                "SHELL=%4%; "
```
edolstra: Probably PATH and SHELL shouldn't both refer to `%4%`?

grahamc: :man_facepalming:

grahamc:
```
SHELL=/nix/store/pd3xx55jjik53gvbyrs4h0him03l3wah-bash-interactive-4.4-p23/bin/bash
PATH=/nix/store/pd3xx55jjik53gvbyrs4h0him03l3wah-bash-interactive-4.4-p23/bin:...
```

PR #2459, opened 2018-09-30
add a sublist primop

cleverca22:
for one of the expressions ive been optimizing, it reduces:
calls to elemAt from 108 million to 59 thousand
adds 106k calls to builtins.sublist
values created down from 400m to 5m
total eval time down from 149sec to 15sec

for a related expression in the same set, it reduces the eval time from 45 minutes to 2 minutes, while still producing the same `.drv` path at the end

a 3rd related expression took over 147 minutes and 66gig of ram (and never finished, i had to kill it)
after this change, it took 19 minutes

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1490,6 +1490,18 @@ static void prim_concatLists(EvalState & state, const Pos & pos, Value * * args,
     state.concatLists(v, args[0]->listSize(), args[0]->listElems(), pos);
 }
 
+/* return a subsection of a list */
+static void prim_sublist(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    int start = state.forceInt(*args[0], pos);
+    size_t len = state.forceInt(*args[1], pos);
+    state.forceList(*args[2], pos);
+    size_t count = std::min(args[2]->listSize() - start, len);
+
+    state.mkList(v, count);
+    auto out = v.listElems();
+    memcpy(out, args[2]->listElems() + start, count * sizeof(Value*));
+}
 
 /* Return the length of a list.  This is an O(1) time operation. */
 static void prim_length(EvalState & state, const Pos & pos, Value * * args, Value & v)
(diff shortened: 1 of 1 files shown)


disassembler:
LGTM! :+1:

src/libexpr/primops.cc
```diff
@@ -1490,6 +1490,18 @@ static void prim_concatLists(EvalState & state, const Pos & pos, Value * * args,
     state.concatLists(v, args[0]->listSize(), args[0]->listElems(), pos);
 }
 
+/* return a subsection of a list */
+static void prim_sublist(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    int start = state.forceInt(*args[0], pos);
+    size_t len = state.forceInt(*args[1], pos);
+    state.forceList(*args[2], pos);
+    size_t count = std::min(args[2]->listSize() - start, len);
```
edolstra: I think this underflows if `start` > `listSize()`.

cleverca22: i can confirm it malfunctions:
```
nix-repl> builtins.sublist 11 3 [ 1 2 3 4 5 6 7 8 9 10 ]
start is 11
len is 3
listSize is 10
count is 3
[ Segmentation fault (core dumped)
```
will update the PR soon

grahamc:
If you would, please also add an under flow test to the mix tests.

dtzWill:
Friendly ping! No rush on my account, just want to help ensure doesn't fall through the cracks :).

cleverca22:
still need to get around to this, not completely stale

Ericson2314:
Marked as draft because:

1. Needs tests
2. Needs fixed behavior with out-of-bound indices

I think the idea is good, however. @cleverca22's numbers at the top are, well, shocking! And we have `builtins.substring` so this matches.

fricklerhandwerk:
Triaged in Nix team meeting:

- @edolstra: wonder if we could be lazy, essentially implementing slices
    - @thufschmitt: thought that, too, but can't promise this will end up in the implementation
- idea approved, @thufschmitt will take over

thufschmitt:
Closing since https://github.com/NixOS/nix/pull/8290 took over (and is also closed by lack of evidence of the performance gain).
@cleverca22 , feel free to reopen if you want to work on this again

PR #2462, opened 2018-10-01
nix-profile-daemon: remove cruft

matthewbauer:
This removes part of the PATH that were being added automatically in multi-user installs:

- $HOME/.nix-profile/lib/kde4/libexec - shouldn't be needed anymore, we are now using kde5
- @localstatedir@/nix/profiles/default/lib/kde4/libexec - same as above
- @localstatedir@/nix/profiles/default - shouldn't ever contain binaries


grahamc:
Good catch, thank you!

PR #2479, opened 2018-10-16
Add --graphml option to the nix-store --query command

nlewo:
This prints the references graph of the store paths in the graphML
format [1]. The graphML format is supported by several graph tools
such as the Python Networkx library or the Apache Thinkerpop project.

I also wonder if this could replace the `--xml` output since the GraphML format is well known and easy to manually parse.

[1] http://graphml.graphdrawing.org


src/nix-store/graphml.cc
```diff
@@ -0,0 +1,91 @@
...
+static string symbolicName(const string & path)
+{
+    string p = baseNameOf(path);
+    return string(p, p.find('-') + 1);
+}
+
+
+static string makeEdge(const string & src, const string & dst)
+{
+    format f = format("  <edge source=\"%1%\" target=\"%2%\"/>\n")
```
dtzWill: Maybe `return fmt(..., xmlQuote(src), xmlQuote(dst));`? Saves the need for calling `f.str()`.

src/nix-store/graphml.cc
```diff
@@ -0,0 +1,91 @@
...
+         << "<key id='name' for='node' attr.name='name' attr.type='string'/>"
+         << "<key id='type' for='node' attr.name='type' attr.type='string'/>"
+         << "<graph id='G' edgedefault='directed'>\n";
+
+    while (!workList.empty()) {
+        Path path = *(workList.begin());
+        workList.erase(path);
+
+        if (doneSet.find(path) != doneSet.end()) continue;
+        doneSet.insert(path);
```
dtzWill: Maybe check the return value of insert instead (`.second`)? Shorter and saves searching again for the insert.

src/nix-store/graphml.cc
```diff
@@ -0,0 +1,91 @@
...
+
+        ValidPathInfo info = *(store->queryPathInfo(path));
+        cout << makeNode(info);
+
+        for (auto & p : store->queryPathInfo(path)->references) {
+            if (p != path) {
+                workList.insert(p);
```
dtzWill: Using a std::set for the worklist might not be the best, and seems like would have an unusual ordering, niether BFS nor DFS but ... alphabetically (of the current worklist)? If so, I imagine it doesn't matter to tools reading it but might annoy or confuse someone inspecting it by hand.

edolstra:
Yeah I think it's fine to replace `--xml`, since I doubt anybody is using it at the moment.

nlewo:
I'll add docs and remove `--xml` in the next days.

nlewo:
I've added docs and removed `--xml` option. So, it should be good for reviews:/

@dtzWill FYI, I'm using this option with https://github.com/nlewo/gremnix :)

PR #2482, opened 2018-10-20
Add a --keep-successful flag to nix-build.

noteed:
The new flag mirrors the --keep-failed flag; it keeps the temporary build directory when the build is successful instead of removing it.


cstrahan:
I've used the old trick of adding a failing line to the builder shell script countless times, just so I could inspect the build directory. Having this would be very nice.

PR #2483, opened 2018-10-22
Promote verbose-build and log-lines

lheckemann:
These are now fully-qualified options, allowing commands like

```
nix build --verbose-build nixpkgs.hello
nix build --log-lines 30 nixpkgs.hello
```

In order to obtain more information in case of a failure.

cc @colemickens who was asking about this on IRC, @gilligan who was asking about this on twitter :)

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -82,9 +82,9 @@ public:
     /* Whether to show build log output in real time. */
     bool verboseBuild = true;
 
-    /* If verboseBuild is false, the number of lines of the tail of
-       the log to show if a build fails. */
-    size_t logLines = 10;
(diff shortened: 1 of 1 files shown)


lheckemann:
That's no different from `nix-build` though, right?

lheckemann:
In any case, I've pushed a new version that only promotes log-lines, and will have a look at what we could do for verbose build.

lheckemann:
(original version of this PR consisted of c82db2af735bd3a680225cd12b50051b46cc9e33 for future reference)

edolstra:
Yes, but `nix build` is supposed to improve upon `nix-build` UI rather than reproduce all its warts.

PR #2485, opened 2018-10-26
Fix aws-sdk-cpp patch to work with aws-sdk-cpp on 18.09+

dtzWill:
Current patch works with 18.03, not sure if this does--
it's a simple rebase as I recall.

I've started pinning aws-sdk-cpp to a specific version
to ensure the patch applies (and nix can be built),
but I didn't put much thought into it beyond getting
things to work :).

### DIFF
--- a/release-common.nix
+++ b/release-common.nix
@@ -64,10 +64,7 @@ rec {
         apis = ["s3" "transfer"];
         customMemoryManagement = false;
       }).overrideDerivation (args: {
-        patches = args.patches or [] ++ [ (fetchpatch {
-          url = https://github.com/edolstra/aws-sdk-cpp/commit/3e07e1f1aae41b4c8b340735ff9e8c735f0c063f.patch;
-          sha256 = "1pij0v449p166f9l29x7ppzk8j7g9k9mp15ilh5qxp29c7fnvxy2";
-        }) ];
+        patches = args.patches or [] ++ [ ./transfermanager-content-encoding.patch ];
       }));
 
   perlDeps =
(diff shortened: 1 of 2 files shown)


release-common.nix
```diff
@@ -64,10 +64,7 @@ rec {
         apis = ["s3" "transfer"];
         customMemoryManagement = false;
       }).overrideDerivation (args: {
-        patches = args.patches or [] ++ [ (fetchpatch {
-          url = https://github.com/edolstra/aws-sdk-cpp/commit/3e07e1f1aae41b4c8b340735ff9e8c735f0c063f.patch;
-          sha256 = "1pij0v449p166f9l29x7ppzk8j7g9k9mp15ilh5qxp29c7fnvxy2";
-        }) ];
+        patches = args.patches or [] ++ [ ./transfermanager-content-encoding.patch ];
```
shlevy: I think you need `(arg.patches or []) ++ ...` :sob:

dtzWill: Apparently not! Seems to work as expected poking at it in `nix repl`:

```
nix-repl> a = { x = [ "bar" ]; }

nix-repl> a.x or [] ++ [ "foo" ]
[ "bar" "foo" ]

nix-repl> a = { y = [ "bar" ]; }

nix-repl> a.x or [] ++ [ "foo" ]
[ "foo" ]
```

edolstra:
I've dropped the patch instead since we don't need it anymore (6323b0729a27f312ba8cce6279de68181644823e).

PR #2489, opened 2018-10-26
Fix typo in local.mk

layus:


dtzWill:
Dup of #2847, which also fixes nix-prefetch-url.

dtzWill:
#2487

layus:
Oh, right ! Thanks.

PR #580, opened 2015-07-09
Feature: Indent multiline antiquotations (v2).

layus:
Antiquotations in indented strings will have their last empty line removed
(when applicable) and an indented antiquotation will see all its lines
correctly re-indented.

:exclamation: This features changes the semantics of indented strings and therefore triggers a mass rebuild of (nearly) all the derivations.
I still believe it is worth it, but we should obviously gather all such changes in one step.

As a small example is better than a long description, the following expression

``` nix
let 
  abc = ''
    A
      ${""}
    B
  '';
  context = body: ''
    BEGIN
      ${body}
    END
  '';
in context (context abc)
```

results in 

``` nix
''
BEGIN
  BEGIN
    A
    B
  END
END
''
```

instead of the old

```
BEGIN
  BEGIN
  A

B

END

END
```

Re-indent only applies to antiquotations alone on one line, surrounded only by spaces.

This pull requests is a rewrite of #549 on top of #573.
It fixes #543.

PR #2490, opened 2018-10-26
Feature: Indent multiline antiquotations (v3)

layus:
After a long discussion in #580, it appears that there are two safe paths to introduce this feature. Either use a custom operator for antiquotations (like `${> ... }` for example) or introduce pragma-like statements to change the semantics on a per-file basis, and in a backward compatible manner. Again, see #580 for details.

In this PR, I implemented the pragma idea introduced by @copumpkin [[1]]. This seems better than introducing a new operator. I am proposing a semantics change, not a new operator. Oh, and also because I agree with @vcunat's [[2]].

To get this new semantics within a file, one needs to add `/*@ LANGUAGE reindent @*/` at the top of the file. That's it. This implementation is not a mass rebuild, and does not break anything.

"Smart antiquotations", tersely renamed to the "reindent" language feature in this PR are antiquotations alone on a line of a multiline nix string (a.k.a. `''`-strings). When that occurs, the whole inserted block gets reindented to match the indentation level of the antiquotation itself.

This means that
```nix
/*@ LANGUAGE reindent @*/
let
  inside = "a\nb\nc";
in ''
  BEGIN
    ${inside}
    BEGIN
      ${inside}
    END
  END
''
```

ends up looking like

```nix
''
BEGIN
  a
  b
  c
  BEGIN
    a
    b
    c
  END
END
''
```

instead of the old

```nix
''
BEGIN
  a
b
c
  BEGIN
    a
b
c
  END
END
''
```

and we can get rid of code like https://github.com/NixOS/nixpkgs/blob/8070a6333f3fc41ef93c2b0e07f999459615cc8d/nixos/modules/services/x11/xserver.nix#L76-L86

### Things that need discussion

The actual syntax for pragmas can still be changed, but will be introduced once and for all here.

----

This PR needs rebasing onto smaller changes yet to merge and included here for simplicity.

The code is available for testing on layus.cachix.org as /nix/store/73fi8m9zylxc4g72r4h55wcyr99p2skh-nix-2.2pre0_0000000

[1]: https://github.com/NixOS/nix/pull/580#issuecomment-229400896
[2]: https://github.com/NixOS/nix/pull/580#issuecomment-120693002


shlevy:
@layus I think we should separate this into independent discussions

1. Should we have conditional evaluation modes of some kind? If so, what should the interface be?
2. Should we fix indentation somehow? If so, how should we expose that?

IMO for number 1 I think the answer is yes. I think pragmas are a decent approach (they work well enough for haskell), and if/when we have package specifications for nix libraries we can set project-wide settings there as well.

For number 2, I think we should fix it somehow but I'm completely ambivalent between having it as a pragma vs a different quotation syntax.

@NixOS/nix-core can y'all chime in on these questions and we can try to make a decision soon here?

domenkozar:
I'm not in Nix core team, but then we can really forever say goodbye for smoothless upgrades. Even today upgrading from NixOS version to another, Nix fails due to some compatibility features. I guess we should fix that before making more of them :)

domenkozar:
Thinking about it more, the extensions would make the errors better (eg. https://github.com/NixOS/nixpkgs/issues/36268#issue-302037547). But this should have an RFC since it's introducing a big change to the language.

edolstra:
@domenkozar This feature should not be a problem for NixOS upgrades because it only affects evaluation, and `nixos-rebuild` already fetches the latest stable Nix for evaluation. The problem is when new build features are required (as is the case for structured derivation attributes).

BTW, instead of using magic comments, which are always an ugly hack, it seems better to use something like
```
require feature reindent;
```
since then the file will fail to parse with older versions of Nix. However I really don't fancy adding annotations like that to thousands of Nix expressions. (See also https://github.com/golang/proposal/blob/master/design/28221-go2-transitions.md#build-tags for some discussion of this style.) If we're going to implement flakes, then it's probably best to specify a language version in the flake file, similar to Rust's epoch mechanism.

shlevy:
I don't think it's mutually exclusive to do it per-file or per-flake (GHC has both, for example). IMO though it's nicer to specify a set of extensions (possibly with some bundles of common extensions for convenience) than a version.

PR #2491, opened 2018-10-26
Simplify handling of extra '}'

layus:
Add a marker state at the bottom of the state stack to detect when we
close too many antiquotations.
This way, the DEFAULT state can be used as the only state for nix
expressions.


Mic92:
cool. this makes porting the grammar to other languages much easier.

PR #2494, opened 2018-10-27
copyStorePath: Fix hash errors when copying from older store

FPtje:
This commit partially reverts 48662d151bdf4a38670897beacea9d1bd750376a. When
copying from an older store (in my case a store running Nix 1.11.7), nix would
throw errors about there being no hash. This is fixed by recalculating the hash.

From the Nixcon 2018 hackathon

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -588,15 +588,19 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
 
     uint64_t total = 0;
 
-    // FIXME
-#if 0
     if (!info->narHash) {
+        StringSink sink;
+        srcStore->narFromPath({storePath}, sink);
         auto info2 = make_ref<ValidPathInfo>(*info);
         info2->narHash = hashString(htSHA256, *sink.s);
         if (!info->narSize) info2->narSize = sink.s->size();
+        if (info->ultimate) info2->ultimate = false;
         info = info2;
+
+        StringSource source(*sink.s);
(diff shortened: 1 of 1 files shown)


src/libstore/store-api.cc
```diff
@@ -589,14 +589,29 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
...
+            Callback callback;
+            MyStringSink(Callback callback) : callback(callback) { }
+            void operator () (const unsigned char * data, size_t len) override
+            {
+                StringSink::operator ()(data, len);
+                callback(len);
+            };
+        };
+        MyStringSink sink(progress);
+        srcStore->narFromPath({storePath}, sink);
```
edolstra: Note that this will cause the NAR to be copied twice (due to the later call to `narFromPath`). You could create a `StringSource` for `sink.s` and pass it to `dstStore->addToStore()`.

BTW I think the whole progress stuff can be removed here since this is just a backward compatibility path.

FPtje:
I've force pushed the following changes to the code:

- Fixed NAR being copied twice by indeed creating a `StringSource` for `sink.s` and passing it to `dstStore->addToStore()`. Thanks for helping me along the way there. I'm not familiar with the Nix code base, and your hint sent me along the right path.
- Removed the progress stuff.
- Removed `FIXME` comment, as it now seems to be properly fixed

I've tested by copying things to and from a Nix 1.11.7 store and to and from a nix 2.0.4 store.

edolstra:
Thanks!

PR #2495, opened 2018-10-29
nix doctor: return nonzero exitcode if a check fails

LnL7:
This makes it easier to use this when testing the installer or when
running the checks with other automated tooling.

### DIFF
--- a/src/nix/doctor.cc
+++ b/src/nix/doctor.cc
@@ -18,6 +18,8 @@ std::string formatProtocol(unsigned int proto)
 
 struct CmdDoctor : StoreCommand
 {
+    bool success = true;
+
     std::string name() override
     {
         return "doctor";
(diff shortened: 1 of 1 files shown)


src/nix/doctor.cc
```diff
@@ -36,13 +38,16 @@ struct CmdDoctor : StoreCommand
...
-            checkNixInPath();
-            checkProfileRoots(store);
+            success &= checkNixInPath();
+            success &= checkProfileRoots(store);
         }
-        checkStoreProtocol(store->getProtocol());
+        success &= checkStoreProtocol(store->getProtocol());
+
+        if (!success)
+            throw Exit(2);
```
LnL7: Is there something other than the description I can use to document this exit code?

edolstra: No, the `nix` command does not have a man page at the moment.

LnL7: It would be nice to have something to provide a longer description that only shows up with `nix <command> --help` similar to the examples.

PR #2501, opened 2018-10-31
config: use all of XDG_CONFIG_DIRS

lheckemann:
Previously, config would only be read from XDG_CONFIG_HOME. This change
allows reading config from additional directories, which enables e.g.
per-project binary caches or chroot stores with the help of direnv.

cc @shlevy who gave me the idea


edolstra:
Thanks!

shlevy:
Nice!!!

shlevy:
@grahamc FYI :smile:

grahamc:
:tada:

PR #2504, opened 2018-10-31
Use nativeBuildInputs for nix-shell

matthewbauer:
Now the nix-shell will generate a nix expression that looks like this:

```
with import <nixpkgs> {};
runCommand "shell" {
  nativeBuildInputs = [
    (...)
    (...)
    (...)
  ];
}
```

from

$ nix-shell --packages ... ... ...

Fixes the issue described in https://github.com/NixOS/nixpkgs/issues/38657.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -249,7 +249,7 @@ static void _main(int argc, char * * argv)
 
     if (packages) {
         std::ostringstream joined;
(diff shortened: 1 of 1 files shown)


MrChico:
Any chance of this getting some love?

matthewbauer:
Some cases could also screw this up, however. For instance in some cases you want to treat your dependencies as libraries (for zlib, libsystemd, etc.). For now it is probably safest to leave the behavior as is.

PR #2506, opened 2018-11-01
WIP:  Fix aws-sdk-cpp patch can not be applied

janpath:


janpath:
@edolstra Is this OK?

janpath:
Actually this breaks `nix-shell`, as `release.nix` and `shell.nix` use different `nixos` versions (`18.09` and `18.03`). Is this intended?

edolstra:
Thanks, indeed it should both be 18.09.

Since we don't really need this patch anymore, I've dropped it altogether (6323b0729a27f312ba8cce6279de68181644823e).

PR #2508, opened 2018-11-01
Create version.nix containing used nixos version

janpath:
This ensures, that `release.nix` and `shell.nix` always use the same version
of nixos. See https://github.com/NixOS/nix/pull/2506#issuecomment-435047869.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -1,5 +1,5 @@
 { nix ? builtins.fetchGit ./.
-, nixpkgs ? builtins.fetchGit { url = https://github.com/NixOS/nixpkgs-channels.git; ref = "nixos-18.09"; }
+, nixpkgs ? ./version.nix
 , officialRelease ? false
 , systems ? [ "x86_64-linux" "i686-linux" "x86_64-darwin" "aarch64-linux" ]
 }:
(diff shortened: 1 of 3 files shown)


edolstra:
Or even better, moved into `release-common.nix` which exists exactly for this purpose.

janpath:
@edolstra But `release-common.nix` takes a `pkgs` argument itself and is called with different `pkgs` arguments in `release.nix`. Something like this would be possible for `release-common.nix`:

```
{ nixpkgs     ? null
, nixpkgsArgs ? {}
}:

with if nixpkgs != null
  then import nixpkgs nixpkgsArgs
  else import (builtins.fetchGit ...) nixpkgsArgs;

rec {
  inherit pkgs,
  ...
}
```

Then `release.nix` could look like this

```
{ nixpkgs ? null
, ...
}:
...
  build = pkgs.lib.genAttrs systems (system:
    with import ./release-common.nix { inherit nixpkgs; nixpkgsArgs = { inherit system; }; };
...
```

Alternatively `release-common` could look like this:
```
{ nixpkgs ? null }: nixpkgsArgs:
...
```
and `release.nix` like this:
```
{ nixpkgs ? null
, ...
}:
releaseCommon = import ./release-common.nix {inherit nixpkgs;}
...
  build = pkgs.lib.genAttrs systems (system:
    with releaseCommon { inherit system; };
```

I am very new to nix and do not know which would be the most idiomatic. To me a separate `version.nix` or `nixpkgs.nix` seemed the most straight-forward.

roberth:
This has been resolved by moving to flakes and flakes-compat.

PR #2516, opened 2018-11-05
Update multiuser install remove instruction with correct path and users + group removal

goodwillcoding:
@grahamc whenever you got time :)


matthewbauer:
Looks good to me!

On a related note though, it would be nice to move these from documentation into the script itself. So you could run something like:

```
  sh <(curl https://nixos.org/nix/install) uninstall
```

and have the script run these steps for you.

Issue #2522, opened 2018-11-07
Setuid root does not work in sandbox


lukego:
Running a setuid-root binary inside a Nix sandbox does not actually set the uid to 0. For example `whoami` will report `nixbld` and `security.wappers` programs will crash with an assertion failure when detecting that the effective uid does not match the file uid.

Can reproduce either by calling `/run/wrappers/bin/sudo` in the sandbox (have to make it visible) or by running `/run/as/root $(which id)` with my little [asroot](https://gist.github.com/a26e31edf5cf9fac4e2b8ab91169ae93) NixOS module.

Seems to me like the `allow-new-privileges` option was added to make this work (https://github.com/NixOS/nix/issues/1429) but that this is not working with Nix from NixOS 18.09.

Notes:

- NixOS 18.09.
- Setuid binary on a filesystem that does not have `nosetuid` flag.
- Nix.conf includes `allow-new-privileges = true`.

Wild guess: Could this be due to userns sandboxing not providing a usable root user?

edolstra:
> Wild guess: Could this be due to userns sandboxing not providing a usable root user?

Yes, that's sounds likely. From the user_namespaces manpage:

> However, if either the user or the group  ID of  the  file has no mapping inside the namespace, the set-user-ID (set-group-ID) bit is silently ignored: the new program is executed, but the process's effective user (group) ID is left unchanged.

So for this to work, there would have to be a uid mapping to the root user in the parent namespace.

copumpkin:
Curious what the use case is here, can you elaborate? It feels slightly wrong to be doing setuid things inside a build sandbox but I'm probably missing something.

lukego:
> Curious what the use case is here, can you elaborate?

I should write a FAQ on this :-)

We have a Hydra cluster running CI benchmarks on some open source telecom network stack called [Snabb](https://github.com/snabbco/snabb). The Nix job builds a benchmark report with Rmarkdown and that involves running many (thousands) benchmark jobs. The benchmark jobs need root privileges due to the nature of the software e.g. running userspace 10G/100G ethernet device drivers, resolving physical memory, performing DMA, etc. This cannot be wrapped in virtualization due to the performance-sensitive nature e.g. enabling the IOMMU could impact performance.

[...]

PR #2525, opened 2018-11-08
Allow root uid/gid (0) in sandbox userns when allowNewPrivileges

lukego:
If Nix is run with the `--allow-new-privileges` option then it should be possible to run setuid binaries to gain root access. However, this did not work in practice because the root user/group did not exist in the sandbox due to exclusion from the kernel user-namespace. The uid would always be nixbld (1000) even when executing a setuid-root binary.

This change adds uid/gid 0 to the sandbox userns only when `--allow-new-privileges` is enabled and makes setuid work as expected.

Note: This makes it effective to *run* setuid executables but it does not make it possible to create them. This means that in practice to gain root in the sandbox you must both provide `--allow-new-privileges` and also add a suitable binary to the Nix sandbox path.

Resolves nixos/nix#2522.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2243,11 +2243,13 @@ void DerivationGoal::startBuilder()
         uid_t hostGid = buildUser ? buildUser->getGID() : getgid();
 
         writeFile("/proc/" + std::to_string(pid) + "/uid_map",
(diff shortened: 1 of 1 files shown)


edolstra:
Maybe there should also be an entry added to `/etc/passwd` for uid 1? Something like `host-root`.

PR #2552, opened 2018-11-20
Fix issues in systemv init script

KaiHa:
This commit fixes some issues reported by [lintian][1]:

- [init.d-script-missing-lsb-keyword][2]
- [init.d-script-missing-start][3]
- [init.d-script-missing-dependency-on-local_fs][4]
- [init.d-script-does-not-implement-required-option][5]

[1]: https://lintian.debian.org/
[2]: https://lintian.debian.org/tags/init.d-script-missing-lsb-keyword.html
[3]: https://lintian.debian.org/tags/init.d-script-missing-start.html
[4]: https://lintian.debian.org/tags/init.d-script-missing-dependency-on-local_fs.html
[5]: https://lintian.debian.org/tags/init.d-script-does-not-implement-required-option.html

### DIFF
--- a/misc/systemv/nix-daemon
+++ b/misc/systemv/nix-daemon
@@ -9,12 +9,13 @@
 # pidfile:     /var/run/nix/nix-daemon.pid
 
 ### BEGIN INIT INFO
-# Required-Start:    
-# Required-Stop:     
+# Provides:          nix-daemon
+# Required-Start:    $local_fs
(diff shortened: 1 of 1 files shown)


KaiHa:
This pull request can be abandoned. For Debian I will provide an own init.d script since those scripts are distribution dependent and will likely not work on other Linux distributions.

PR #2562, opened 2018-12-04
Solve hg "abandoned transaction" issue

chawki008:
When a `fetchMercurial` operation is launched, Nix starts cloning the specified repo by executing a  ``` hg clone``` command and stores the repo in the `cacheDir`. 
In case there is an interruption during the cloning process, hg is left with an "abandoned transaction" in the cached repo directory. 
When re-fetching  the repo again, Nix will find that this repo already exists in the `cacheDir` and so it will execute a ```hg pull``` command this time. And since we are left with an "abandoned transaction", hg will throw an "abandoned transaction" error and the `fetchMercurial` operation will fail.

**So now we catch that error and we run a ```hg recover ``` to fix the issue and finally we pull again by running ```hg pull```.**

The hg command always returns an 255 exit code when there is an error, **any error**, so we can't detect the "abandoned transaction" error accurately based only on hg exit code.
However, reading hg source code, it turned out that hg throws that error only if a certain file named "journal" doesn't exist. Thus, in order to detect this error accurately, i check whether that file exists or not and if it is the case i run the recovering commands.

### DIFF
--- a/src/libexpr/primops/fetchMercurial.cc
+++ b/src/libexpr/primops/fetchMercurial.cc
@@ -93,7 +93,22 @@ HgInfo exportMercurial(ref<Store> store, const std::string & uri,
             Activity act(*logger, lvlTalkative, actUnknown, fmt("fetching Mercurial repository '%s'", uri));
 
             if (pathExists(cacheDir)) {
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks!

PR #2565, opened 2018-12-06
Saner defaults for MANPATH

jvassev:
Very few systems explicitly define the manpath. When installing on
arbitrary linux, the nix-installed manapages are not discovered. To do
so always add NIX_LINK, then /usr/share/man if it exists.

### DIFF
--- a/scripts/nix-profile.sh.in
+++ b/scripts/nix-profile.sh.in
@@ -75,8 +75,18 @@ if [ -n "$HOME" ] && [ -n "$USER" ]; then
         export NIX_SSL_CERT_FILE="$NIX_LINK/etc/ca-bundle.crt"
     fi
 
+    # Prefer MANPATH if configured
+    # Otherwise build a MANPATH which would work on the LHS
     if [ -n "${MANPATH}" ]; then
-        export MANPATH="$NIX_LINK/share/man:$MANPATH"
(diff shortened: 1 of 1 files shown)


vcunat:
I can't see how `/usr/share/man` is relevant to nix-installed man pages.

edolstra:
The problem with setting `MANPATH` is that it overrides man's default search path, which might include more entries than /usr/share/man (e.g. /usr/local/share/man).

vcunat:
Hmm, I see the problem (or one at least).  The behavior of man searching "relative" to `$PATH` still isn't common, I guess?   For example, with the default man from nixpkgs we don't need any `$MANPATH` at all and it just works thanks to `$PATH`.

jvassev:
This is how I got here:
1. install nix on centos using the documented way to do so
2. `nix-env -i tree`
3. `man tree`
4. "No manual entry for tree"

If I just define MANPATH then tree's man page is found.

Also it makes sense to include  /usr/local/share/man in the default case as well.

jvassev:
@vcunat

Yes, the man on centos is version `1.6f`. If I install `man` from nixpkgs I get version `1.6g` which find all man pages.

vcunat:
Right, that seems the best work-around to me.  We might still improve compatibility with legacy man versions, but honestly _I_ don't feel like putting energy into that.

PR #2566, opened 2018-12-07
s3: make scheme configurable

LnL7:
This enables using for http for S3 request for debugging or
implementations that don't have https configured.  This is not a problem
for binary caches since they should not contain sensitive information.
Both package signatures and AWS auth already protect against tampering.

### DIFF
--- a/doc/manual/packages/s3-substituter.xml
+++ b/doc/manual/packages/s3-substituter.xml
@@ -51,6 +51,18 @@ the S3 URL:</para>
     addressing.</para></note>
   </listitem>
   </varlistentry>
+
+  <varlistentry><term><literal>scheme</literal></term>
+  <listitem>
+    <para>
+      The scheme used for S3 requests, <literal>https</literal>
+      (default) or <literal>http</literal>.  This option allows you to
+      disable HTTPS for binary caches which don't support it.
+    </para>
+    <note><para>HTTPS should be used if the cache might contain
+    sensitive information.</para></note>
+  </listitem>
+  </varlistentry>
 </variablelist>
 
 <para>In this example we will use the bucket named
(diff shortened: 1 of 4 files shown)


src/libstore/s3-binary-cache-store.cc
```diff
@@ -169,6 +172,7 @@ struct S3BinaryCacheStoreImpl : public S3BinaryCacheStore
 {
     const Setting<std::string> profile{this, "", "profile", "The name of the AWS configuration profile to use."};
     const Setting<std::string> region{this, Aws::Region::US_EAST_1, "region", {"aws-region"}};
+    const Setting<std::string> scheme{this, "", "scheme", "The scheme to use for S3, http can generally be used since caches should not contain sensitive information."};
```
grahamc: I'd rather not generally advertise `http` as being a good choice.

grahamc:
Please also extend this list: https://github.com/NixOS/nix/blob/master/doc/manual/packages/s3-substituter.xml#L28

src/libstore/s3-binary-cache-store.cc
LnL7: It's not enabled by default, it's just there for debugging, etc.

doc/manual/packages/s3-substituter.xml
```diff
@@ -51,6 +51,16 @@ the S3 URL:</para>
     addressing.</para></note>
   </listitem>
   </varlistentry>
+
+  <varlistentry><term><literal>scheme</literal></term>
+  <listitem>
+    <para>
+      This enables using for http for S3 request for debugging or
+      implementations that don't have https configured.  Should not
+      be used if the cache might contain sensitive information.
```
grahamc: Let's put `<example>https</example> (default) or <example>http</example>` at the start, and then something like:

> This option allows you to disable HTTPS for binary caches which don't support it. HTTPS should be used if the cache might contain sensitive information.

Issue #2568, opened 2018-12-09
Manual: missing prerequisites: brotli, libseccomp


endgame:
The manual section at https://nixos.org/nix/manual/#sec-prerequisites-source does not list brotli or seccomp as dependencies, but if you configure on a system without them, `./configure` will fail.

PR #2569, opened 2018-12-10
docs: add missing prerequisites: brotli, boost, libseccomp

veprbl:
The list of prerequisites is incomplete. This should add the missing ones. The minimal supported version of "boost" is determined by trying different versions until nix compiled successfully.

This would need a backport to the maintenance branch.

Fixes: #2568

### DIFF
--- a/doc/manual/installation/prerequisites-source.xml
+++ b/doc/manual/installation/prerequisites-source.xml
@@ -59,7 +59,7 @@
   <command>configure</command>.</para></listitem>
 
   <listitem><para>The <literal>boost</literal> library of version
-  1.61.0 or higher. It can be obtained from the official web site
+  1.66.0 or higher. It can be obtained from the official web site
   <link xlink:href="https://www.boost.org/" />.</para></listitem>
 
   <listitem><para>The <command>xmllint</command> and
--- a/doc/manual/installation/prerequisites-source.xml
+++ b/doc/manual/installation/prerequisites-source.xml
@@ -25,6 +25,12 @@
   If your distribution does not provide it, you can get it from <link
   xlink:href="https://www.openssl.org"/>.</para></listitem>
 
+  <listitem><para>The <literal>libbrotlienc</literal> and
+  <literal>libbrotlidec</literal> libraries to provide implementation
+  of the Brotli compression algorithm. They are available for download
+  from the official repository <link
+  xlink:href="https://github.com/google/brotli" />.</para></listitem>
+
   <listitem><para>The bzip2 compressor program and the
   <literal>libbz2</literal> library.  Thus you must have bzip2
   installed, including development headers and libraries.  If your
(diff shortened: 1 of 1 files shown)


endgame:
Great work. If you want to be specific about which parts of boost nix needs to build, it's only the `context` library, everything else is header-only. (build boost with `--wtih-libraries=context`).

doc/manual/installation/prerequisites-source.xml
```diff
@@ -77,6 +87,15 @@
   modify the parser or when you are building from the Git
   repository.</para></listitem>
 
+  <listitem><para>The <literal>libseccomp</literal> is used to provide
```
endgame: I didn't realise this was optional. I just assumed it was required because configure bombs out if you don't have it.

veprbl:
@endgame 
For me, it builds with boost that was built `--with-libraries=""`, so it seems like `context2` library is header-only. I don't think it's worth to go to such details and describe this in the manual because this might get outdated because of the changes either in nix or in boost.

endgame:
Strange. AFAICT it's unconditionally linking against `-lboost_context` at https://github.com/NixOS/nix/blob/6924bdf2bf8f50f2e6ec5d490571594450aba13a/src/libutil/local.mk#L9

Do you want that script run on amd64 or aarch64?

veprbl:
@endgame 
I found this line too, but somehow that didn't look suspicious to me at the time. Turns out `--with-libraries=""` builds all boost libraries... So this is why my build didn't fail. I poked it a bit more and it seems like the minimal working boost configuration currently is `--with-libraries=context`.

doc/manual/installation/prerequisites-source.xml
```diff
@@ -52,6 +58,10 @@
   pass the flag <option>--enable-gc</option> to
   <command>configure</command>.</para></listitem>
 
+  <listitem><para>The <literal>boost</literal> library of version
+  1.61.0 or higher. It can be obtained from the official web site
```
edolstra: I believe it should be higher due to some boost::coroutine2 bug in previous releases. Maybe 1.65 but I'm not sure.

endgame:
@veprbl Well, that explains it. Do you still need me to run that script, and if so, on which arch?

doc/manual/installation/prerequisites-source.xml
veprbl: @edolstra
I guess you are referring to
https://github.com/NixOS/nix/blob/03ce0c3a9e1c07043445579c619eed4895647b36/src/libutil/serialise.cc#L160-L162
It singles out 1.63.x, 1.64.x, 1.65.x to exclude specifically. A test can be done a following way:
```
with import <nixpkgs> {};

map (boost: nix.override { inherit boost; })
[
  #boost155
  #boost159
  #boost160 ld: symbol(s) not found for architecture x86_64
  #boost161 not in nixpkgs, but works if you add it
  boost162
  #boost163 assert fails in serialise.cc
  #boost164 assert fails in serialise.cc
  #boost165 assert fails in serialise.cc
  boost166
  boost167
  boost168
]
```
All four build for me (including `installCheckPhase`) just fine. I can change the minimum version to 1.66, if you think that is needed.

veprbl:
@endgame I think we are good. You were right about boost_context being a dependency, thanks for pointing that out. Also I learn that `coroutine2` used to be a shared library in older versions of boost.

doc/manual/installation/prerequisites-source.xml
edolstra: Thanks for investigating :-)

PR #2572, opened 2018-12-12
nar-info-disk-cache: include ca in the cache entries

LnL7:
Without this information the content addressable state and hashes are lost after the first request, this causes signatures to be required for everything even tho the path could be verified without signing.

```
$ rm -r ~/.cache/nix
$ nix path-info --store https://cache /nix/store/6j01qy6rj13q5nibm4gpmlv6x7sfgim8-foo --sigs
[0.0 MiB DL]
/nix/store/6j01qy6rj13q5nibm4gpmlv6x7sfgim8-foo ca:text:sha256:0schdl901lnx98l7gmm33x5jvz2xsshli0f15nwm7z7igxjv30vz

$ nix path-info --store https://cache /nix/store/6j01qy6rj13q5nibm4gpmlv6x7sfgim8-foo --sigs
/nix/store/6j01qy6rj13q5nibm4gpmlv6x7sfgim8-foo
```

Requiring signatures for everything might be nice as an optional setting, but the current behaviour looks like a bug.

### DIFF
--- a/src/libstore/nar-info-disk-cache.cc
+++ b/src/libstore/nar-info-disk-cache.cc
@@ -31,6 +31,7 @@ create table if not exists NARs (
     refs             text,
     deriver          text,
     sigs             text,
+    ca               text,
     timestamp        integer not null,
     present          integer not null,
     primary key (cache, hashPart),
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks, good catch :-)

Issue #2407, opened 2018-09-07
Multi-user installer: don't fail if /nix exists


Mic92:
On some space constraint systems, where I sneak in nix I have /nix on a different partition.
The current installer however does not let me install nix, when `/nix` is mounted and present:

```
---- oh no! --------------------------------------------------------------------
There are some relics of a previous installation of Nix at /nix, and
this scripts assumes Nix is _not_ yet installed. Please delete the old
Nix installation and start again.
Uninstalling nix:
1. Delete the files Nix added to your system:
  sudo rm -rf /etc/nix /nix /root/.nix-profile /root/.nix-defexpr /root/.nix-channels /home/mbailleu/.nix-profile /home/mbailleu/.nix-defexpr /home/mbailleu/.nix-channels
and that is it.
We'd love to help if you need it.
If you can, open an issue at https://github.com/nixos/nix/issues
```

Proposed solution: Check if /nix/ is not empty instead.

Mic92:
cc @grahamc what do you think about that?

Mic92:
Oh, weird I think systemd mounts my /nix partition during the installation.

abathur:
This can be closed; the core issue was fixed in #3054.

abathur:
For the record, I noticed after commenting on this one that despite #3054 getting merged, the commit that fixed this issue hasn't made it into 2.3-maintenance.

So, this is "fixed", but you'll still see this on current installers until that is either cherry picked or there's a new release from master.

PR #2573, opened 2018-12-12
install-multi-user: fail only if /nix is not empty

### LINKED ISSUE #2042: Multi-user install: allow /nix to exist
Currently multi user installer requires that `/nix` is absent.
However there are many case where one would want to `/nix` to be a mount point, so the installer should just check/ensure it has correct access rights (and modify them if needed).

Cleanup instruction also need to be updated accordingly.

uvNikita:
This will allow having /nix on a separate partition. Fixes #2042 and #2407.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -324,7 +324,7 @@ EOF
         fi
     done
 
-    if [ -d /nix ]; then
+    if [ -e /nix/store -o -e /nix/var ]; then
         failure <<EOF
 There are some relics of a previous installation of Nix at /nix, and
 this scripts assumes Nix is _not_ yet installed. Please delete the old


uvNikita:
@grahamc Is there anything that's blocking this PR from been merged? We are planning to put /nix on a separate partition and it would be very convenient if the upstream installer would support this case :)

orivej:
I am replacing `if [ -d /nix ]` with `if [ -e /nix/store -o -e /nix/var ]` for the same purpose.

uvNikita:
@orivej I replaced implementation with your version since it seems to be more reliable, thanks for the input :)

uvNikita:
For someone who will have the same issue in the future [[1]](https://xkcd.com/979/), we ended up using these packages kindly provided by @Mic92 and @dezgeg: https://github.com/Mic92/nix-fpm-multiuser/releases.

abathur:
This can be closed; a similar fix was merged in #3054.

PR #2575, opened 2018-12-12
binary-cache-store: add setting to disable local narinfo caching

LnL7:
This is nice for debugging since the cache can have some unintended side
effects when, eg. paths are removed but not invalidated.

```
$ nix path-info --store 's3://test?cache=0' /nix/store/6j01qy6rj13q5nibm4gpmlv6x7sfgim8-foo
skipping local narinfo cache for 's3://test'
querying info about '/nix/store/6j01qy6rj13q5nibm4gpmlv6x7sfgim8-foo' on 's3://test'...
downloaded 's3://nix/6j01qy6rj13q5nibm4gpmlv6x7sfgim8.narinfo' (392 bytes) in 21 ms
querying foo on s3://nix
/nix/store/6j01qy6rj13q5nibm4gpmlv6x7sfgim8-foo
```

### DIFF
--- a/src/libstore/http-binary-cache-store.cc
+++ b/src/libstore/http-binary-cache-store.cc
@@ -23,6 +23,8 @@ private:
 
 public:
 
+    const Setting<bool> cache{this, true, "cache", "Whether to cache narinfo metadata for this binary cache."};
+
     HttpBinaryCacheStore(
         const Params & params, const Path & _cacheUri)
         : BinaryCacheStore(params)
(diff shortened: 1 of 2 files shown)


src/libstore/http-binary-cache-store.cc
```diff
@@ -23,6 +23,8 @@ class HttpBinaryCacheStore : public BinaryCacheStore
 
 public:
 
+    const Setting<bool> cache{this, true, "cache", "Whether to cache narinfo metadata for this binary cache."};
```
AmineChikhaoui: should probably move to `src/libstore/binary-cache-store.hh` ?

Also I guess this is useful per binary cache but it's possible to set something globally through `--option narinfo-cache-{positive/negative}-ttl 0` that disables cache as well.

LnL7: narinfo caching doesn't make sense for file binary caches, but I could move it

LnL7:
Guess this is not particularly useful, for debugging `narinfo-cache-positive-ttl` can be used.

PR #2577, opened 2018-12-12
installables: resolve derivation outputs

LnL7:
When converting a drv path to an installable the buildables would
contain `{"out", "/nix/store/<path>.drv"}` as outputs instead of the
actual outputs from the derivation.

This makes sure the result symblinks created `nix build` point to the
outputs instead of the drv file that was built.

### DIFF
--- a/src/nix/installables.cc
+++ b/src/nix/installables.cc
@@ -96,7 +96,15 @@ struct InstallableStorePath : Installable
 
     Buildables toBuildables() override
     {
-        return {{isDerivation(storePath) ? storePath : "", {{"out", storePath}}}};
+        if (isDerivation(storePath)) {
+            std::map<std::string, Path> outputs;
+            auto drv = readDerivation(storePath);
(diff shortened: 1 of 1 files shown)


src/nix/installables.cc
```diff
@@ -96,7 +96,15 @@ struct InstallableStorePath : Installable
 
     Buildables toBuildables() override
     {
-        return {{isDerivation(storePath) ? storePath : "", {{"out", storePath}}}};
+        if (isDerivation(storePath)) {
+            std::map<std::string, Path> outputs;
+            auto drv = readDerivation(storePath);
```
LnL7: I'm not sure this is the most optimal way to query the outputs, but it works for diverted/remote stores from what I can tell.

LnL7:
This probably has unintended side effects. I just noticed `toStorePaths` also uses this, so every command that takes installables will go through this.

PR #2578, opened 2018-12-13
probably typo

volth:
...at least MSVC unable to compile this


edolstra:
It's a digraph. I must have been in a Python mode when I wrote that :-)

PR #2579, opened 2018-12-13
nix-store: make --dump-db take a list of paths to dump

catern:
Inside a derivation, exportReferencesGraph already provides a way to
dump the Nix database for a specific closure. On the command line,
--dump-db gave us the same information, but only for the entire Nix
database at once.

With this change, one can now pass a list of paths to --dump-db to get
the Nix database dumped for just those paths. (The user is responsible
for ensuring this is a closure, like for --export).

Among other things, this is useful for deploying a closure to a new
host without using --import/--export; one can use tar to transfer the
store paths, and --dump-db/--load-db to transfer the validity
information. This is useful if the new host doesn't actually have Nix
yet, and the closure that is being deployed itself contains Nix.

### DIFF
--- a/doc/manual/command-ref/nix-store.xml
+++ b/doc/manual/command-ref/nix-store.xml
@@ -1282,6 +1282,7 @@ ktorrent-2.2.1/NEWS
   <cmdsynopsis>
     <command>nix-store</command>
     <arg choice='plain'><option>--dump-db</option></arg>
+    <arg rep='repeat'><replaceable>paths</replaceable></arg>
   </cmdsynopsis>
 </refsection>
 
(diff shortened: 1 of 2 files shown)


lheckemann:
:+1: on the change in general!

This will only print the registration for the paths themselves, not their closure, right?

catern:
Correct, it's just the registration for the paths themselves, not their closure. Like with --export, the user is responsible for passing all the paths in a closure.

lheckemann:
Actually, this is missing documentation.

catern:
Ping? @lheckemann

lheckemann:
LGTM still, but I can't merge it.

Issue #2385, opened 2018-08-31
fetchGit only works on non-annotated tags


graham-at-target:
The following works:

```
(import <nixpkgs> {}).stdenv.mkDerivation {
  name = "hi";
  src = builtins.fetchGit {
    url = "https://github.com/nixos/nix.git";
    ref = "tags/1.9";
  };
}
```

But this doesn't:

```
(import <nixpkgs> {}).stdenv.mkDerivation {
  name = "hi";
  src = builtins.fetchGit {
    url = "https://github.com/nixos/nix.git";
    ref = "tags/2.0";
  };
}
```

1.9 isn't an annotated tag:

```
[nix-shell:~/projects/nix]$ git show 1.9
commit 39b3d9698e8c0972d7d54c4007c4a2e6e680b3d3 (tag: 1.9)
Author: Eelco Dolstra <eelco.dolstra@logicblox.com>
Date:   Fri Jun 12 11:02:40 2015 +0200

    Typos
```

but 2.0 is:

```
$ git show 2.0
tag 2.0
Tagger: Eelco Dolstra <edolstra@gmail.com>
Date:   Thu Feb 22 16:40:21 2018 +0100

Tagging release 2.0
-----BEGIN PGP SIGNATURE-----

iQFHBAABCAAxFiEEtUHVUwEnDgvPFcpdgXC0cm1xmN4FAlqO5GUTHGVkb2xzdHJh
QGdtYWlsLmNvbQAKCRCBcLRybXGY3kFnB/9+SXZx0DWt1iDXtp2sVH08OC5097fg
zOE1MJXIMpn1EVPmFjrTUiTThlO0GVZTXK1abxTq2MLZ2e5CpXZHzMqmIiR4axWa
jwYwG3/beIFF1RBwBma3Yc7M7vN95Eyp1YderMFZSY9qk9/0OGyXc9LZHvyaQYUX
NbNkuoHHr3qM2jYrebehDxIUNN0VfA43s4Jrci5MR8OXM8/xFk82BzVyeTmocwut
b46NBLobRXwlRktTZNXrx2wJeY7JluASHpcTAdmBlntpKTUpvkGN/9hOQ1lZUMjj
+jUB9PIOHUx0gE1QhiUplpmYCPF4B0Fx43Ek0Le3C1pDUC37+BsUX1mt
=AbBN
-----END PGP SIGNATURE-----

commit 179b896acb6deb8fea9614dfbddeaf3b23797bf5 (tag: 2.0)
[...]

PR #2582, opened 2018-12-14
fetchGit: allow fetching explicit refs

LnL7:
Trying to fetch refs that are not in refs/heads currently fails because
it looks for refs/heads/refs/foo instead of refs/foo.

This enables some cool stuff like fetching the magic pull-request refs from github.

```nix
builtins.fetchGit {
  url = https://github.com/NixOS/nixpkgs.git;
  ref = "refs/pull/1024/head;
}
```

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -419,6 +419,13 @@ stdenv.mkDerivation { … }
               This is often a branch or tag name. Defaults to
               <literal>HEAD</literal>.
             </para>
+
+            <para>
+              By default, the <varname>ref</varname> value is prefixed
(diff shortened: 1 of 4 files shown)


lionello:
I wonder if this would address #2385 as well.

FRidh:
Related issue on `fetchGit`.

lheckemann:
+1 @edolstra anything blocking merging this?

dasJ:
@edolstra Ping again

grahamc:
I pushed a commit documenting this.

edolstra:
BTW I just noticed we're not validating `ref`, so it's only thanks to Git validating it that you can't do things like
```
fetchGit { url = https://github.com/NixOS/patchelf.git; ref = "../../../../../../etc/passwd"; }
```

PR #2583, opened 2018-12-14
nix: allow using --file - to read from stdin

LnL7:
Brings the `nix-instantiate -` functionality to all the new nix commands that take the --file argument.

	$ echo '{ foo = 1 + 1; }' | nix eval -f - foo
	2

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -728,41 +728,51 @@ Value * ExprPath::maybeThunk(EvalState & state, Env & env)
 
 void EvalState::evalFile(const Path & path_, Value & v)
 {
-    auto path = checkSourcePath(path_);
+    if (path_ == "-") {
+        Expr * e = parseStdin();
+        try {
+            eval(e, v);
+        } catch (Error & e) {
+            addErrorPrefix(e, "while evaluating %1%:\n", "<stdin>");
+            throw;
+        }
+    } else {
+        auto path = checkSourcePath(path_);
 
-    FileEvalCache::iterator i;
-    if ((i = fileEvalCache.find(path)) != fileEvalCache.end()) {
-        v = i->second;
-        return;
+        FileEvalCache::iterator i;
+        if ((i = fileEvalCache.find(path)) != fileEvalCache.end()) {
+            v = i->second;
+            return;
+        }
+
+        Path path2 = resolveExprPath(path);
+        if ((i = fileEvalCache.find(path2)) != fileEvalCache.end()) {
+            v = i->second;
+            return;
+        }
+
+        printTalkative("evaluating file '%1%'", path2);
+        Expr * e = nullptr;
+
+        auto j = fileParseCache.find(path2);
+        if (j != fileParseCache.end())
+            e = j->second;
+
+        if (!e)
+            e = parseExprFromFile(checkSourcePath(path2));
+
+        fileParseCache[path2] = e;
+
+        try {
+            eval(e, v);
+        } catch (Error & e) {
+            addErrorPrefix(e, "while evaluating the file '%1%':\n", path2);
+            throw;
+        }
+
+        fileEvalCache[path2] = v;
+        if (path != path2) fileEvalCache[path] = v;
     }
-
-    Path path2 = resolveExprPath(path);
-    if ((i = fileEvalCache.find(path2)) != fileEvalCache.end()) {
-        v = i->second;
-        return;
-    }
-
-    printTalkative("evaluating file '%1%'", path2);
-    Expr * e = nullptr;
-
-    auto j = fileParseCache.find(path2);
-    if (j != fileParseCache.end())
-        e = j->second;
-
-    if (!e)
-        e = parseExprFromFile(checkSourcePath(path2));
-
-    fileParseCache[path2] = e;
-
-    try {
-        eval(e, v);
-    } catch (Error & e) {
-        addErrorPrefix(e, "while evaluating the file '%1%':\n", path2);
-        throw;
-    }
-
-    fileEvalCache[path2] = v;
-    if (path != path2) fileEvalCache[path] = v;
 }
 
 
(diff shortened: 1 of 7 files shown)


src/libexpr/eval.cc
```diff
@@ -716,41 +716,51 @@ Value * ExprPath::maybeThunk(EvalState & state, Env & env)
...
+        try {
+            eval(e, v);
+        } catch (Error & e) {
+            addErrorPrefix(e, "while evaluating %1%:\n", "<stdin>");
+            throw;
+        }
+    } else {
+        auto path = checkSourcePath(path_);
 
-    FileEvalCache::iterator i;
```
dtzWill: Pre-existing problem, but while looking through this... declaring these temporaries in the conditional seems to solve obvious scoping problems currently resolved by using variable names like `path2` for temporaries later in the sequence... depending on `auto` preferences it'd look like this:

(EDIT: just kidding, looks like a clean solution for this requires c++17 maybe? (`if (auto i = fileEvalCache.find(path); i != fileEvalCache.end())`))

src/libexpr/eval.cc
```diff
@@ -716,41 +716,51 @@ Value * ExprPath::maybeThunk(EvalState & state, Env & env)
...
-    fileParseCache[path2] = e;
+        fileParseCache[path2] = e;
 
-    try {
-        eval(e, v);
-    } catch (Error & e) {
-        addErrorPrefix(e, "while evaluating the file '%1%':\n", path2);
-        throw;
-    }
+        try {
```
dtzWill: oh, was the bit inserted at the top cloned from this? Does the code lend itself towards handling parsing/caching as appropriate, followed by single block for eval/error handling? Meh, I see some of my nits from earlier were from pre-existing code, sorry.  Ignore if they seem tedious, cleanup of existing code can and probably should be handled orthogonally...

src/libexpr/eval.cc
```diff
@@ -716,41 +716,51 @@ Value * ExprPath::maybeThunk(EvalState & state, Env & env)
...
+            eval(e, v);
+        } catch (Error & e) {
+            addErrorPrefix(e, "while evaluating the file '%1%':\n", path2);
+            throw;
+        }
 
-    fileEvalCache[path2] = v;
-    if (path != path2) fileEvalCache[path] = v;
+        fileEvalCache[path2] = v;
+        if (path != path2) fileEvalCache[path] = v;
```
dtzWill: Eep.  Okay this is an unexpected turn, nevermind.  May wish to early-return to avoid some churn, see earlier comment.

dtzWill:
(I can't mark them resolved since they're comments! D'oh.  Sorry about that :))

src/libexpr/eval.cc
```diff
@@ -716,41 +716,51 @@ Value * ExprPath::maybeThunk(EvalState & state, Env & env)
...
-    auto path = checkSourcePath(path_);
+    if (path_ == "-") {
+        Expr * e = parseStdin();
+        try {
+            eval(e, v);
+        } catch (Error & e) {
+            addErrorPrefix(e, "while evaluating %1%:\n", "<stdin>");
+            throw;
+        }
+    } else {
```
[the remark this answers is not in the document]
LnL7: not sure what style is preferred, happy to change it if desired

src/libexpr/eval.cc
```diff
@@ -716,41 +716,51 @@ Value * ExprPath::maybeThunk(EvalState & state, Env & env)
 
 void EvalState::evalFile(const Path & path_, Value & v)
 {
-    auto path = checkSourcePath(path_);
+    if (path_ == "-") {
+        Expr * e = parseStdin();
+        try {
+            eval(e, v);
+        } catch (Error & e) {
+            addErrorPrefix(e, "while evaluating %1%:\n", "<stdin>");
```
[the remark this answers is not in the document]
LnL7: the third parameter is required, I think the assumption is that a format string with 1 parameter is passed.

LnL7:
@dtzWill I dismissed some of the comments, most of them where about existing code that I just indented.

src/libexpr/eval.cc
```diff
@@ -728,41 +728,51 @@ Value * ExprPath::maybeThunk(EvalState & state, Env & env)
 
 void EvalState::evalFile(const Path & path_, Value & v)
 {
-    auto path = checkSourcePath(path_);
+    if (path_ == "-") {
+        Expr * e = parseStdin();
```
edolstra: This looks very dangerous. It means that a certain value of `path` will be interpreted as something completely different (i.e. as a blocking read of standard input). For example, doesn't this mean that `import "-"` will now read from stdin even in pure evaluation mode?

IMHO `-` should only be handled at top-level while processing the command line (and even there `-` should only have special meaning if it doesn't occur after `--`).

grahamc: I agree that evalFile should not handle this specially, and something else should be done.

However, exploiting this would not be so trivial:

```
[nix-shell:~/projects/github.com/NixOS/nix]$ ./inst/bin/nix-instantiate -E 'import "-"'
error: string '-' doesn't represent an absolute path, at (string):1:1

[nix-shell:~/projects/github.com/NixOS/nix]$ ./inst/bin/nix-instantiate -E 'import -'
error: syntax error, unexpected $end, at (string):1:8

[nix-shell:~/projects/github.com/NixOS/nix]$ ./inst/bin/nix-instantiate -E 'import "/-"'
error: getting status of '/-': No such file or directory

[nix-shell:~/projects/github.com/NixOS/nix]$ ./inst/bin/nix-instantiate -E 'import "./-"'
error: string './-' doesn't represent an absolute path, at (string):1:1

[nix-shell:~/projects/github.com/NixOS/nix]$ ./inst/bin/nix-instantiate -E 'import -'
error: syntax error, unexpected $end, at (string):1:8

[nix-shell:~/projects/github.com/NixOS/nix]$ ./inst/bin/nix-instantiate -E '(import -)'
error: syntax error, unexpected ')', at (string):1:10
```

LnL7: I'd have to look at the code again, but I think the condition can probably be moved to the --file argument without much of a problem. Which is indeed a better place to handle it.

LnL7: done

Artturin:
https://github.com/NixOS/nix/pull/6270

Artturin:
can be closed

PR #2586, opened 2018-12-15
nix-store: remove debugging print

dtzWill:


dtzWill:
Err ping @NixOS/nix-core ? Definitely silly as-is and causes every single invocation of `nix-store` to print that it "GOT HASH ASDFFOOBAR" (only formatted with the hardcoded hash string)... :laughing:

PR #2594, opened 2018-12-20
installer: update macOS version check to 10.12.2

LnL7:
Nixpkgs will drop support for <10.12 soon and thus a nix release built
using the 19.03 channel will also require a newer version of macOS.

See https://github.com/NixOS/nixpkgs/pull/47678

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -22,10 +22,12 @@ if [ -z "$HOME" ]; then
     exit 1
 fi
 
-# macOS support for 10.10 or higher
+# macOS support for 10.12.6 or higher
 if [ "$(uname -s)" = "Darwin" ]; then
-    if [ $(($(sw_vers -productVersion | cut -d '.' -f 2))) -lt 10 ]; then
-        echo "$0: macOS $(sw_vers -productVersion) is not supported, upgrade to 10.10 or higher"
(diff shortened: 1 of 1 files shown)


matthewbauer:
I think we should probably hold off on this until the 19.03 release. There were a lot more failures than originally expected.

scripts/install-nix-from-closure.sh
```diff
@@ -22,10 +22,12 @@ if [ -z "$HOME" ]; then
...
 fi
 
-# macOS support for 10.10 or higher
+# macOS support for 10.12.6 or higher
 if [ "$(uname -s)" = "Darwin" ]; then
-    if [ $(($(sw_vers -productVersion | cut -d '.' -f 2))) -lt 10 ]; then
-        echo "$0: macOS $(sw_vers -productVersion) is not supported, upgrade to 10.10 or higher"
+    macos_major=$(sw_vers -productVersion | cut -d '.' -f 2)
+    macos_minor=$(sw_vers -productVersion | cut -d '.' -f 3)
+    if [ "$macos_major" -lt 12 ] || ([ "$macos_major" -eq 12 ] && [ "$macos_minor" -lt 6 ]); then
```
matthewbauer: @lnl7 Do we care about the minor version? I would think these would be compatible, although I haven't tested it.

LnL7: I don't know what version exactly, but the previous attempt definitively needed `>10.12.2`.

Issue #2597, opened 2018-12-22
Cannot RegisterCommand from a plugin


kirelagin:
I am trying to implement a plugin that provides a command. The problem is that `RegisterCommand` and all related types, such as `Command` itself, are declared in `src/nix/command.hh`, which does not get installed.

Is this by design? What is the recommended way of building plugins that register a command?

kirelagin:
Related: the static variable that contains commands is defined in `src/nix/command.cc`, so it doesn’t exist in binaries other than `nix`, even though they still load plugins.

kirelagin:
> the static variable that contains commands is defined in src/nix/command.cc, so it doesn’t exist in binaries other than nix, even though they still load plugins
This doesn’t make any sense. I have realised that legacy commands all go through `nix` as well, so the variable should have been there. I’m not sure why they crash then 😕.

edolstra:
I'm surprised the docs say that `RegisterCommand` is supported, since that would not have worked at the time, I think. (The merge of the legacy commands into `nix` was done only very recently.)

kirelagin:
I almost got it working, the problem for now is that we need to parse options first, then init plugins, and only after that see if the command is known. I am not quite sure what would be the best way to do this, still thinking.

kirelagin:
> (The merge of the legacy commands into nix was done only very recently.)

Ohhh, now it makes sense. I was jumping back and forth between 2.1.3 and master when testing, and now I realise that `nix-shell` was crashing on 2.1.3, where it hasn’t been merged yet.

Ericson2314:
Is the new `libnixcmd` and command plugins fixing this?

mikenrafter:
Still reproduces on CppNix 2.36 (`2.36pre20260912_203f85b2`): `RegisterCommand` plugins show up in `nix --help` / `NIX_GET_COMPLETIONS`, but `nix <plugin-cmd>` fails because `parseCmdline` resolves the top-level `MultiCommand` subcommand before `initialFlagsProcessed()` → `initPlugins()`.

Filed a current repro with details: https://github.com/NixOS/nix/issues/16498

PR #2600, opened 2018-12-22
Add support for plugins that provide a new subcommand

kirelagin:
This feature has been documented for quite some time but never really worked. Now it does.

The only visible change is that previously plugins were loaded after all of the command line was parsed (so, obviously, subcommands provided by plugins were ignored), now they are loaded after the subcommand name has been parsed and right before the subcommand is resolved. Therefore, the `plugin-files` option has no effect if it comes _after_ the subcommand name. Ideally, a warning should be issues in this case, but I have no idea how to do this.

See also #2597.


kirelagin:
(cc @edolstra)

edolstra:
Looks good to me, could you resolve the merge conflict?

BTW we could also get rid of libmain (i.e. merge it into src/nix) since it's not really useful anymore now that we have a single binary.

7c6f434c:
Do we already truly have a single binary? Isn't `nix-env` still separate (and without full replacement)? Even `nix-build` has functionality not conveniently available in `nix build`…

kirelagin:
@7c6f434c Yep, all those are symlinks to `nix`.

kirelagin:
Rebased.

kirelagin:
@edolstra ping!

kirelagin:
👋

kirelagin:
:(

edolstra:
Could this be done without moving `installables.{cc,hh}` to `libmain`?

BTW the flakes branch introduces a couple of nested subcommands (`nix flake <command>`, `nix profile <command>`). Maybe we want to make those extensible as well.

kirelagin:
I think, it can be done, but why though? Many plugins will want to parse commands and, in particular, work with installables, so it makes sense to have this code in the library. `command.{cc,hh}` will definitely have to go there, and if you prefer to live `installables.cc` in `src/nix/` then it might be possible to extract the relevant declarations from `command.hh` into `installables.hh` (which doesn’t exist). But why exactly don’t you want to move them?

edolstra:
`installables.{cc,hh}` and `command.{cc,hh}` contain a lot of stuff specific to the `nix` command (like the `Installable` class) which isn't relevant to other programs that use `libmain` (like Hydra). It might make sense to move `MultiCommand` and `RegisterCommand` to `libutil`.

Issue #2604, opened 2018-12-28
configure.ac contains Bash-specific syntax


0mp:
I'm trying to port Nix to FreeBSD and I've come across a problem when running `./configure`.

It turns out that the configure file that is generated from configure.ac has Bash-specific syntax although the shebang line specifies /bin/sh.

The offending line is here: https://github.com/NixOS/nix/blob/3d974d31facefe8eaf59af56b0187e6a63fdd0cc/configure.ac#L217

Is this a bug or the shebang line should refer to Bash (if so then how can I tell autoconf the path to my Bash binary)?

PR #2607, opened 2018-12-30
Remove some bashisms from configure.ac

0mp:
"configure.ac" is used to generate "configure", which is supposed to be
run with /bin/sh (as suggested by the shebang line of "configure"). As a
result it is a good idea to remove any /bin/sh-incompatible syntax from
configure.ac. Otherwise, systems that do not use Bash as their /bin/sh
are unable to run "configure" due to syntax errors.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -214,9 +214,11 @@ AC_SUBST(ENABLE_S3, [$enable_s3])
 AC_LANG_POP(C++)
 
 if test -n "$enable_s3"; then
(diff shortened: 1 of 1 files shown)


0mp:
Related to #2604.

kirelagin:
This line from `configure.ac`

```
    if printf '%s' "${name}" | egrep '^[_[:alpha:]][_[:alnum:]]*$'
```

ends up being

```
    if printf '%s' "${name}" | egrep '^_[:alpha:]_[:alnum:]*$'
```

in `configure` after running `./bootstrap.sh`, which results in

* `grep: character class syntax is [[:space:]], not [:space:]`
* and, hm, I guess, in the end, it just does not what it was meant to.

I’m not sure how this even works 😕.

0mp:
Hmm, it looks like we have to escape `[` and `]` here (https://stackoverflow.com/a/2309394/4694621).

Issue #2503, opened 2018-10-31
Rewrite build system in Meson


Ericson2314:
**Checklist in this comment: https://github.com/NixOS/nix/issues/2503#issuecomment-2207393570**

The specific features I am thinking that Autoconf lacks are 
 - Windows support (with and without mingw)
 - Outputs Ninja or something similar

Both of these are non essential, but good for various experiments:
 - Windows support obviously allows experimenting with Nix supporting windows. GCC is *inadequate* because it doesn't offer C++ interopt with MSVC, so Windows's many C++-only APIs are unusable.
 - Outputing a dependency graph allows fun experiments with Recursive Nix (https://github.com/edolstra/nix/commit/1cfdfd728aa7faa78b55b97a2c2f2b5922cd079f) and ninja2nix (https://github.com/awakesecurity/ninja2nix)

This lowers the list down to [Meson](http://mesonbuild.com/) and [CMake](https://cmake.org/), AFIAK. I prefer Meson because:
 - It doesn't use compiler makefile generation, which avoids a regrettable ninja impurity
 - It's less crufty. CMake became better, but old cmake was bad, leaving a lot of legacy cruft. Meson is a clean slate so there's no minefield of deprecated bad ways of doing things.
 - Full disclosure, I've been working on improving Meson's cross compilation support, so I'm partial to it.

CC @taktoa

taktoa:
Sadly `ninja2nix` doesn't really exist -- I tried my best to convert Ninja features into Nix ones using IFD where necessary, but it's just not feasible. If you want to take another crack at it using Recursive Nix, `language-ninja` is still quite a good library for Ninja-related stuff in Haskell. I think the feature that I couldn't figure out how to compile to IFD was `depfile`, but I might be mistaken.

Ericson2314:
@taktoa But meson doesn't use depfile out of an ideological stance against dynamic dependencies / crawling the filesystem. Problem avoided!

Ericson2314:
So the only Recursive Nix / IFD would be converting the ninja file. Everything inside should be "static Nix".

taktoa:
> This lowers the list down to Meson and CMake, AFIAK. I prefer CMake because:

`s/I prefer CMake/I prefer Meson/` presumably?

Ericson2314:
haha yeah

7c6f434c:
[...]

PR #2613, opened 2019-01-05
WIP: line-for-line rewrite of build system in Meson

Ericson2314:
Includes https://github.com/NixOS/nix/pull/2612/files which includes dead code which was hard to produce.

Fixes https://github.com/NixOS/nix/issues/2503

CC @volth


volth:
currently MSVC version is built using
https://github.com/volth/nix-windows/blob/c0166319ed8e07bc8011fa6009540e0800498431/Makefile.win
https://github.com/volth/nix-windows/blob/c0166319ed8e07bc8011fa6009540e0800498431/build.cmd
which is obviously bad - it is too verbose and lacks the expected features such as tracking .hh-files

I will try Meson soon.

Ericson2314:
@volth I can now at least build all the C++

meson.build
```diff
@@ -0,0 +1,304 @@
...
+conf_data = configuration_data()
+
+system = get_option('with-system')
+if system == ''
+  system = host_machine.cpu() + '-' + host_machine.system()
+endif
+conf_data.set_quoted('SYSTEM', system, description : 'platform identifier (`cpu-os\')')
+
+# State should be stored in /nix/var, unless the user overrides it explicitly.
+if get_option('normal-var')
```
matthewbauer: Can we determine if an option has been manually set by the user in Meson? Adding an option like this is confusing.

matthewbauer:
There are a few install rules missing still:

https://github.com/NixOS/nix/search?q=install-data-in&unscoped_q=install-data-in

meson.build
Ericson2314: It is confusing; I forget where I got stuck before. This is the autoconf original https://github.com/NixOS/nix/blob/master/configure.ac#L49-L50 . Nix hijack's the variable, in a way. I think the issue was you cannot force-set such a thing within `meson.build`, but you should still be able to compare the user-given value to `${prefix}/var` like the autoconf. That prevents the user from forcing it to be `${prefix}/var`, but hey that's no worse than today.

Ericson2314:
Closing because https://github.com/NixOS/nix/pull/3160 is further along

PR #2618, opened 2019-01-09
Escape square brackets in configure.ac

0mp:
This patch attempts to solve the regression introduced in e8b0efdc
(#2607).

It wasn't detected by me before the previous commit because GNU grep behaviour for `egrep '^_[:alpha:]_[:alnum:]*$'` varies across versions:
- grep (GNU grep) 2.5.1-FreeBSD does not report any issues
- grep (GNU grep) 3.3 prints the following message:

    >     /bin/grep: character class syntax is [[:space:]], not [:space:]

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -275,7 +275,7 @@ test "$prefix" = NONE && prefix=$ac_default_prefix
 test "$exec_prefix" = NONE && exec_prefix='${prefix}'
 for name in $ac_subst_vars; do
     for _ in 1 2 3; do
-    if printf '%s' "${name}" | egrep '^[_[:alpha:]][_[:alnum:]]*$'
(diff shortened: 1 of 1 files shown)


kirelagin:
Now it gets turned into `'^[_[[:alpha:]]][_[[:alnum:]]]*$'`, which still doesn’t look right. Looks like the inner pair of brackets does not need to be escaped.

kirelagin:
I’ve got another regression. It looks a lot like it was introduced by the same commit.

```
src/libstore/s3-binary-cache-store.cc:95:23: error: invalid token at start of a preprocessor expression
#if AWS_VERSION_MAJOR == 1 && AWS_VERSION_MINOR < 3
```

in `config.status`:

```
D["AWS_VERSION_MAJOR"]=" "
D["AWS_VERSION_MINOR"]=" "
```

edolstra:
I've reverted this and the previous commit for now.

PR #2622, opened 2019-01-10
nix build: print built derivation paths (cachix support/nix-build compatibility)

Mic92:
Getting the build path name after a build is a common use case
not supported by nix build yet.
This is for example used in shell scripts like this: $(nix-build)/foo
Cachix even built a business model around it: nix-build | cachix

I have tested that locally and run the tests.
Some tests seem to fail without a sandbox but not due to my change as far as I can tell.

### DIFF
--- a/src/nix/build.cc
+++ b/src/nix/build.cc
@@ -2,6 +2,7 @@
 #include "common-args.hh"
 #include "shared.hh"
 #include "store-api.hh"
+#include "progress-bar.hh"
 
 using namespace nix;
 
@@ -54,17 +55,31 @@ struct CmdBuild : MixDryRun, InstallablesCommand
 
         if (dryRun) return;
 
+        std::vector<Path> buildPaths;
+
         for (size_t i = 0; i < buildables.size(); ++i) {
             auto & b(buildables[i]);
 
-            if (outLink != "")
-                for (auto & output : b.outputs)
-                    if (auto store2 = store.dynamic_pointer_cast<LocalFSStore>()) {
-                        std::string symlink = outLink;
-                        if (i) symlink += fmt("-%d", i);
-                        if (output.first != "out") symlink += fmt("-%s", output.first);
-                        store2->addPermRoot(output.second, absPath(symlink), true);
-                    }
+            for (auto & output : b.outputs) {
+              if (outLink != "") {
+                if (auto store2 = store.dynamic_pointer_cast<LocalFSStore>()) {
+                  std::string symlink = outLink;
+                  if (i)
+                    symlink += fmt("-%d", i);
+                  if (output.first != "out")
+                    symlink += fmt("-%s", output.first);
+                  store2->addPermRoot(output.second, absPath(symlink), true);
+                }
+              }
+
+              buildPaths.push_back(output.second);
+            }
+        }
+
+        stopProgressBar();
+
+        for (auto &p : buildPaths) {
+          std::cout << p << std::endl;
         }
     }
 };


mickours:
Ping! Any chance this will be merged at some point?

Mic92:
I try to remind @edolstra next Thursday.

Mic92:
ping @edolstra

edolstra:
I think it would be better to add a `--json` flag to `nix build` for output intended to be consumed by other tools.

Mic92:
I think a `--json` flag is orthogonal to this feature and also a lot harder to consume in shell scripts than having the path.

domenkozar:
Would be really nice to get this one in.

I think getting output of what one wanted to build is useful and it's already consumable as it's newline separated.

A somewhat different approach with similar goal is at https://github.com/NixOS/nix/pull/2423/files

domenkozar:
Oh there's three PRs, also https://github.com/NixOS/nix/pull/1647

edolstra:
I don't think the new CLI should copy every ad hoc misfeature of the old user interface.

dasJ:
How is getting the result of a build a misfeature?
Using `nix build --no-link` means I'll have to query the drv I just built for the out path instead of just getting the result.

domenkozar:
@edolstra it would be interesting to hear why it's a misfeature.

I feel `--json` is orthogonal as it's about choosing what output format is picked.

I'm guessing that it's because it prints store paths by default?

Maybe there would be a flag to print final store paths (newline delimited) with something like `nix build --report` (and then `nix-build --report --json` for the format).

domenkozar:
Quite sad to see this closed.

Mic92:
Indeed. It would have been nice to have this to build flakes with cachix.

adrian-gierakowski:
May I ask what was the reason for closing this?

Mic92:
It was reject by upstream.

adrian-gierakowski:
Right, but what was the reason behind the rejection? Is there an alternative proposal? Thanks!

Mic92:
Apparently json output. Which does not solve my use case though. I cannot do `$(nix build --json)`.

adrian-gierakowski:
That’s a shame.

domenkozar:
@edolstra is there any reason not to add this under a flag? This would be complimentary to the build hook, except that people can work with outputs via shell piping instead of the complicated setup.

fzakaria:
Throwing this here from #1930 as a stop-gap although also disappointed to see the option of having the OUTPUT printed is gone.
It was very useful for scripting.

```
nbuild() {
  nix build --no-link "$@"
  nix path-info "$@"
}
```

mvnetbiz:
It looks like `jq` is about 900kb on top of Nix, and its obviously more complicated for the simple case, but this doesn't seem too bad.
```console
$ nix build nixos#hello --json | jq -r 'map(.outputs.out) | first'
/nix/store/jmmw0d3nmklwafcwylvrjb9v69wrbcxf-hello-2.10
```
I think cachix or other tools that commonly use this feature should handle json output automatically or with a flag, which would make it simpler to handle multiple-output derivations as well, although currently not all outputs are printed in --json? This looks like separate bug to me.
```console
$ nix build nixos#jq --json | jq
[
  {
    "drvPath": "/nix/store/v8yam49crkivirdr3na81p8zr8mv27p9-jq-1.6.drv",
    "outputs": {
      "bin": "/nix/store/w6n9d6wvkhr68xi47n7i2dw1hkhfwhx0-jq-1.6-bin"
    }
  }
]
```
This produces the links "result  result-bin  result-dev  result-doc  result-lib  result-man" but none are printed in the output.

samuela:
Bummed to not see this being implemented, and not get to understand the reasoning behind the rejection. Here's my workaround:
```
❯ nix build .# --json | jq --raw-output ".[0].outputs.out"                            
/nix/store/dnjsv3xrhb9gajhq7vr55zikcla8q97a-sshenanigans-0.1.0
```

PR #2628, opened 2019-01-14
Context introspection

shlevy:
Add primops to inspect and set context arbitrarily from the Nix language.


shlevy:
On top of #2626 for convenience, can rebase if desired.

edolstra:
`getContext` and `appendContext` are not actually unsafe, are they?

BTW could you put these in a separate file (e.g. `primops/context.cc`)? I'd like to move away from having a giant primops.cc.

src/libexpr/primops.cc
```diff
@@ -1823,11 +1809,144 @@ static void prim_unsafeDiscardOutputDependency(EvalState & state, const Pos & po
...
+                iter->second.allOutputs = true;
+            else
+                iter->second.outputs.emplace_back(std::move(output));
+        }
+    }
+
+    state.mkAttrs(v, contextInfos.size());
+
+    auto sPath = state.symbols.create("path");
+    auto sAllOutputs = state.symbols.create("all-outputs");
```
edolstra: `camelCase` please.

shlevy: Done

shlevy:
> `getContext` and `appendContext` are not actually unsafe, are they?

I guess without one of the context discarding functions they're not, though I do think there's value in most of the time considering string context opaque and quasi-unobservable. Changed the names anyway.

> 
> BTW could you put these in a separate file (e.g. `primops/context.cc`)? I'd like to move away from having a giant primops.cc.

Done

shlevy:
@edolstra Any concerns?

Ericson2314:
Yes it would be really nice to merge this soon. We use this for some testing.

edolstra:
Running `make installcheck` I get:
```
    error: Context key '"/run/user/1000/nix-test/store/drr5fflabsdgyv90sbgc6mrpzswzgh15-eval-okay-context-introspection.nix"' is not a valid store path, at 0x1a5d74d0
    FAIL: eval-okay-context-introspection should evaluate
```

shlevy:
Ah I guess this is because isValidPath doesn't know about paths "added" during the current eval in read-only mode? Is there an alternative?

shlevy:
@edolstra Fixed, followed the logic in builtins.storePath

shlevy:
Ping?

src/libexpr/primops/context.cc
```diff
@@ -0,0 +1,187 @@
...
+           path (i.e. the kind of context you get when interpolating
+           a Nix path (e.g. ./.) into a string). False if missing.
+     allOutputs: True if the relevant path is a derivation and it is
+                  in the context as a drv file with all of its outputs
+                  (i.e. the kind of context you get when referencing
+                  .drvPath of some derivation). False if missing.
+     outputs: If a non-empty list, the relevant path is a derivation
+              and the provided outputs are referenced in the context
+              (i.e. the kind of context you get when referencing
+              .outPath of some derivation). Empty list if missing.
```
edolstra: I have trouble interpreting "False / empty list if missing". This paragraph should presumably read "If the path is a derivation, then this attribute is a list ... Otherwise this attribute is omitted."

However, I wonder if it might be easier for the caller to always include this attributes with values `false` and `[]`?

edolstra: I guess it makes sense with respect to `appendContext` to *not* include them.

PR #2635, opened 2019-01-16
manual: "sandbox" option default is "true" on Linux

veprbl:
Manual needs to be updated after https://github.com/NixOS/nix/commit/812e39313c2bcf8909b83e1e8bc548a85dcd626c


edolstra:
It's only enabled by default on Linux though.

veprbl:
Fixed

Issue #1560, opened 2017-09-07
Install script should set a desired umask


dezgeg:
Problem report noticed in IRC: macOS installation failed if `umask` is 077: https://pastebin.com/p1ppRyqM

edolstra:
CC @grahamc

Anton-Latukha:
This is just what I thought about today.
This is what scripts should check or set.
That is why I have a lot of requirement checks.
http://teaching.idallen.com/cst8177/13w/notes/000_script_style.html#umask

Anton-Latukha:
@edolstra 
Until grahamc gets here.

I think Nix doesn't have any requirements of `umask` in single-user, it is understandable. If we get to make installed working for Linux.

Of course `umask 077` breaks multi-user capability.

But it needs to be described by developer, what are file-permission requirements for Nix multi-user.
What is done from users inside `/nix` , and what workers do inside `/nix`.

To do things right `/nix` probably should be owned by a certain group, and access to `/nix` must be regulated by including user/workers there. And that goes right into installer.

That also means: checks and messages must be present, - if installed finds `/nix` already exists.
From my point of view.

I am in Linux installer see, of course that `/nix` is created just for current user. And result of deployment is default group of user.

Information about what happens in `/nix` must be passed from you to the people who make installers.
I am also interested to know.

Regarding permissions, what I found in manual is a piece about daemon:
```shell
$ chgrp nix-users /nix/var/nix/daemon-socket
$ chmod ug=rwx,o= /nix/var/nix/daemon-socket
```

But other requirements are obscure.
Probably, if users that can get access to binaries, they get to use nix-daemon, that runs workers inside `/nix`. Workers and Nix symlink binaries to users HOME. But since links are invoked with checks on permissions of objects they point - to we have question about user access to `/nix` folder to those files.
It needs more clear description or requirements.
Maybe I read less than needed.

cahlbin:
I ran into a similar issue on _macOS_, having system `umask 0007`, multi-user mode installation. The issue was that my non-privileged user could not source `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh`. Resolved by manually running `sudo chmod -R o+rX /nix/store` post-installation.

PR #2643, opened 2019-01-22
install: Check for a valid umask before running multi-user installer.

### LINKED ISSUE #2377: Installer assumes liberal umask
I attempted to install-multi-user with 2.0.4 on Ubuntu 18.04.

I got the error below.

After scratching my head for a moment, I believe the reason this simple command failed is because of file/directory permissions. By default I believe the default Ubuntu umask is 022. I have set mine, system wide, to be 0077.

The NIX installer should set the umask/file permissions it needs.

Good luck!

---- sudo execution ------------------------------------------------------------
I am executing:

    $ sudo HOME=/root /nix/store/0d60i73mcv8z1m8d2m74yfn84980gfsa-nix-2.0.4/bin/nix-env -i /nix/store/0d60i73mcv8z1m8d2m74yfn84980gfsa-nix-2.0.4

to installing a bootstrapping Nix in to the default Profile

installing 'nix-2.0.4'
building '/nix/store/9v9rz64kdblx991qw8q7472yhisz0rs2-user-environment.drv'...
while setting up the build environment: executing '/nix/store/0d60i73mcv8z1m8d2m74yfn84980gfsa-nix-2.0.4/libexec/nix/buildenv': Permission denied
builder for '/nix/store/9v9rz64kdblx991qw8q7472yhisz0rs2-user-environment.drv' failed with exit code 1
error: build of '/nix/store/9v9rz64kdblx991qw8q7472yhisz0rs2-user-environment.drv' failed

---- oh no! --------------------------------------------------------------------

buecking:
Fixes #2377 and #1560.  I tested this on Linux (Ubuntu 16.04).  I'm not sure if the arguments for `sed` on darwin are different, and I don't have a mac/bsd machine to test on.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -271,6 +271,14 @@ when we need to.
 EOF
     fi
 
+    perms=$(umask -S | sed -e 's/[wx]//g')
+    if [ "u=r,g=r,o=r" != "$perms" ]; then
+        failure <<EOF
+The installation will fail with your current umask $(umask).
+Please set your umask to the system default i.e. umask 002
+EOF
+    fi
+
     if type nix-env 2> /dev/null >&2; then
         failure <<EOF
 Nix already appears to be installed, and this tool assumes it is
@@ -753,21 +761,6 @@ EOF
           install -m 0664 "$SCRATCH/nix.conf" /etc/nix/nix.conf
 }
 
-check_umask() {
-    perms=$(umask -S | sed -e 's/[wx]//g')
-    if [ "u=r,g=r,o=r" != "$perms" ]; then
-        failure <<EOF
-The installation will fail with your current umask $(umask).
-Please set your umask to the system default i.e. umask 002
-EOF
-    fi
-}
-
-preflight_checks() {
-    echo "Running Preflight Checks"
-    check_umask
-}
-
 main() {
     if [ "$(uname -s)" = "Darwin" ]; then
         # shellcheck source=./install-darwin-multi-user.sh
@@ -778,8 +771,6 @@ main() {
     else
         failure "Sorry, I don't know what to do on $(uname)"
     fi
-
-    preflight_checks
 
     welcome_to_nix
     chat_about_sudo
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -753,6 +753,21 @@ EOF
           install -m 0664 "$SCRATCH/nix.conf" /etc/nix/nix.conf
 }
 
+check_umask() {
+    perms=$(umask -S | sed -e 's/[wx]//g')
+    if [ "u=r,g=r,o=r" != "$perms" ]; then
+        failure <<EOF
+The installation will fail with your current umask $(umask).
+Please set your umask to the system default i.e. umask 002
+EOF
+    fi
+}
+
+preflight_checks() {
+    echo "Running Preflight Checks"
+    check_umask
+}
+
 main() {
     if [ "$(uname -s)" = "Darwin" ]; then
         # shellcheck source=./install-darwin-multi-user.sh
@@ -763,6 +778,8 @@ main() {
     else
         failure "Sorry, I don't know what to do on $(uname)"
     fi
+
+    preflight_checks
 
     welcome_to_nix
     chat_about_sudo


scripts/install-multi-user.sh
```diff
@@ -753,6 +753,21 @@ EOF
...
+    if [ "u=r,g=r,o=r" != "$perms" ]; then
+        failure <<EOF
+The installation will fail with your current umask $(umask).
+Please set your umask to the system default i.e. umask 002
+EOF
+    fi
+}
+
+preflight_checks() {
+    echo "Running Preflight Checks"
```
grahamc: Why not just call this in `validate_starting_assumptions` instead of making a new "stage"?

buecking: Sure. Done

edolstra:
Instead of checking for a correct umask, isn't it better to just *set* the correct umask (e.g. `umask 0022`)?

buecking:
> Instead of checking for a correct umask, isn't it better to just _set_ the correct umask (e.g. `umask 0022`)?

There are potential consequences, side-affects, when setting your umask which was why I failed with an error.

I would be more inclined to just set the correct directory permissions on `/nix` rather than set the umask to 0022 and hope that those permissions are acceptable for all files/dirs touched by the installer.

buecking:
I made changes to `nix-install-matrix` to check for regressions, but I wasn't able to run the tests on my laptop or devbox. 

I attempted to extend the README to include my use case:

```
+## Pre-release
+
+In this example we will clone a pending PR, build a pre-release, then
+check this pre-release for regressions.
+
+We need to build a release tarball in order to run our tests. The
+release tarball contains both an installer and the required binary
+packages to install and run NixOS.
+
+```
+git clone git@github.com:contributor/nix.git --branch XXX
+cd nix && nix-build release.nix -A binaryTarball.x86_64-linux
+```
+
+
+If the build finishes successfully it should print the location of the
+tarball on the last line. Unpack the tarball and start a simple
+webserver to serve these data during our tests.
+
+```
+tar axf nix-2.3pre6621_92d08c0-x86_64-linux.tar.bz2 -C /tmp/
+cd /tmp/nix-2.3pre6621_92d08c0-x86_64-linux && python -m SimpleHTTPServer
+# Confirm the URL is working
+curl --output /dev/null --silent --head --fail \
+    "http://localhost:8000/install" || echo No Page Found!
+```
+
+Update `installUrl` in matrix.nix to point to our URL
+
+```
+  installScripts = let
+    installUrls = {
+      pre =  "http://localhost:8000/install";
+      stable = "https://nixos.org/nix/install";
+      "2.0.4" = "https://nixos.org/releases/nix/nix-2.0.4/install";
+    };
+    installUrl = installUrls.pre;
+```
+
+ XXX Write Me
+
+  1. Tweak test script to run preflight checks before creating a release tarball,
+    this step is very long so we want to get it right the first time. Suggest alternate methods
+   for saving time when building the tarball.
+ 
+ 2. Have the test script install the following packages (assume there is no bootstrap for using nixpkgs?) in which case these need to be installed with your distro's package manger.

+wget -q https://www.virtualbox.org/download/oracle_vbox_2016.asc -O- | sudo apt-key add -
+wget -q https://www.virtualbox.org/download/oracle_vbox.asc -O- | sudo apt-key add -
+sudo add-apt-repository "deb http://download.virtualbox.org/virtualbox/debian `lsb_release -cs` contrib"
+sudo apt-get update
+sudo apt-get install virtualbox-5.2
+    sudo apt install qemu-kvm libvirt-bin -- why can't we use nixos?
+    sudo apt install vagrant
+    sudo usermod -a -G libvirtd $USER
+
+  3. System set-up requirements
+    i. $USER belongs to libvirtd group
+    ii.  umask is 022
+
```

Here was the additional test case I added.  I wasn't sure if I should check the umask change on every distro or. 

I thought I would be able to use `file://` protocol instead of `http://` in maxtix.nix, but the individual tests didn't access, and for good reason. So there's a requirement that one must have access to a webserver in order to host the release tarball.

```
diff --git a/matrix.nix b/matrix.nix
index b5cf920..d028da3 100644
--- a/matrix.nix
+++ b/matrix.nix
@@ -1,7 +1,7 @@
 {
   installScripts = let
     installUrls = {
-      pre = "https://nixos.org/releases/nix/nix-2.1pre6385_d16ff76c/install";
+      pre = "http://127.0.0.1:8000/install";
       stable = "https://nixos.org/nix/install";
       "2.0.4" = "https://nixos.org/releases/nix/nix-2.0.4/install";
     };
@@ -194,6 +194,15 @@
       '';
     };
 
+    "ubuntu-16-04-non-standard-umask" = {
+      image = "ubuntu/xenial64";
+      preInstall = ''
+        apt-get update
+        apt-get install -y curl
+        echo "umask 027 >> /etc/profile"
+      '';
+    };
+
     "ubuntu-14-04" = {
       image = "ubuntu/trusty64";
       preInstall = ''
```

Issue #1261, opened 2017-03-05
Feature (removal) request: do not pipe through $PAGER by deafult


alexeymuranov:
I find it rather inconvenient that i have to set up `$NIX_PAGER` just to *disable* piping trough pager for commands like `nix-env -q`, whose equivalents in other OS do not use pager by default (`apt`, `aptitude`, home`brew`, etc.).  For me, passing the output of commands like `ls` through pager is unexpected and inconvenient.

What do others think?

Does any of the contributors actually enjoy using a pager with `nix-env -q` and such, or was this feature introduced "for the others"?

A side note (as i am arguing against using pager by default, this is not very relevant): with the pager used on my system for `man` output, at least i can scroll with the mouse wheel, but it does not work with the default pager of `nix-env`.

vcunat:
There has already been a lengthy discussion on this very topic, I don't remember where exactly.  It's only done when connected to a terminal; I personally find it a reasonable compromise.  You can very easily set up that variable globally.  I think even `ls` changes behavior when run in a terminal...

alexeymuranov:
`ls` changes behaviour??  I've never seen this.

I only run them in terminal, i do not know where else i can run them.

I think that environment variables should not be needed for just reverting to usual (/expected?) POSIX-like behaviour.  Do there exist any POSIX `ls`-like tools that use pager by default?

The argument that `$NIX_PAGER` variable is easy to set up applies both ways :).

I personally find paging `nix-env -q` plain inconvenient.

I had searched for a previous discussion before opening this issue, but didn't find any.

vcunat:
The point of contention here is that the "usual/expected" way is thought to be to use a pager in this case.  I don't really care about the default personally, exactly because the behavior is so easy to change globally.

As for `ls`, you can e.g. try this command sequence with recent coreutils:
```
$ touch "fragile file name"
$ ls
$ ls | grep fragile
```
Columnizing is also automatically changed in the default setting, I think.  (But with `ls` we're getting off-topic IMO.)

alexeymuranov:
Ok, I see what you mean about `ls`.  Personally, i am in favour of columnizing but not in favour of paging.

alexeymuranov:
I have slightly edited the issue description.  As to why i mentioned `ls` -- because `nix-env -q` reminds me more of `ls` than of `man`, for example.

veprbl:
[...]

PR #2656, opened 2019-01-26
Search: no longer output to $PAGER

evils:
Output search results to stdout, let the user see the results for install command.
As requested in #1261

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -7,7 +7,6 @@
 #include "common-args.hh"
 #include "json.hh"
 #include "json-to-value.hh"
-#include "shared.hh"
 
 #include <regex>
 #include <fstream>
@@ -274,7 +273,6 @@ struct CmdSearch : SourceExprCommand, MixJSON
         if (results.size() == 0)
             throw Error("no results for the given search term(s)!");
 
-        RunPager pager;
         for (auto el : results) std::cout << el.second << "\n";
 
     }


zimbatm:
Users who don't want a pager can `export NIX_PAGER=cat` but the reverse is not true.

Evils-Devils:
I wasn't aware of the `NIX_PAGER` variable when i made this PR
nor do i recall ever seeing documentation for it.
If this variable can be supported while the default behaviour can be changed to `cat`, that would be preferable to me, but i'm not familiar enough with how these things are implemented to do that.
I think the current behaviour is less useful as one would usually want to type a `nix-env -i` command right after, and can't view the result while typing if `less` is used.
That was my use case when i was trying out nix on debian, and i think a bias in favour of usability for newcomers is not detrimental here. (an advanced user should know how to use `NIX_PAGER`, a beginner shouldn't)
As `nix search` seems like a natural first point of contact, i think it sends a bad message to potential users to break this convention.

zimbatm:
looks like NIX_PAGER is only in the release notes and should be documented in the manual:
```
$ git grep NIX_PAGER
doc/manual/release-notes/rl-1.9.xml:  <listitem><para>The environment variable <envar>NIX_PAGER</envar>
src/libmain/shared.cc:    char * pager = getenv("NIX_PAGER");
```

alexeymuranov:
IMHO, if `NIX_PAGER` is not set, there should be no paging.

zimbatm:
The pager behaviour is consistent with other tools such as `journalctl` and `git log`

evils:
> The pager behaviour is consistent with other tools such as `journalctl` and `git log`

i believe it is inconsistent with other package search tools such as `apt` and `pkg`

edolstra:
Inconsistent in what way?

evils:
> Inconsistent in what way?

Other package managers don't use a pager to display their results.
At least, the others i've used don't do so by default.

Issue #2474, opened 2018-10-11
nix search --json failure


guibou:
`nix search --json` return a correct result, but with an error code and message:

```
$ nix search --json blender && echo "ok" || echo "\nerror: $?"                              ~
warning: using cached results; pass '-u' to update the cache
error: no results for the given search term(s)!
{"nixpkgs.blender":{"pkgName":"blender","version":"2.79b","description":"3D Creation/Animation/Publishing System"}}
error: 1
```

Morover, adding the `--json` flag returns nothing:

```

Thra11:
I just encountered this error in nix version 2.3.5.

tomberek:
Closing as stale. Re-open if needed.

PR #2657, opened 2019-01-26
Search: remove no output error on json

evils:
Don't warn about no results when using json output.
Addresses #2474


Ma27:
AFAICS this would've been fixed in #2266 (which is pending as well).

Evils-Devils:
yea but this is neater, and not dependent on the -A option working and being accepted

PR #2669, opened 2019-02-07
fix "nix build -f /dev/stdin"

volth:
The idea is to allow Nix code in heredoc

Example:

```
nix build -f /dev/stdin <<'EOF'
let
  inherit (import <nixpkgs> {}) lib buildGoPackage;
in buildGoPackage {
  ...
}
EOF
```

Discussion:

https://discourse.nixos.org/t/nix-shell-nix-build-and-heredoc/2042

### DIFF
--- a/src/libexpr/parser.y
+++ b/src/libexpr/parser.y
@@ -551,6 +551,10 @@ Path resolveExprPath(Path path)
 {
     assert(path[0] == '/');
 
+    // Do not follow symlink chain in /dev, it will be "/dev/stdin" -> "/proc/self/fd/0" -> "/tmp/sh-thd.PUK62Q (deleted)"
+    if (path.compare(0, 5, "/dev/") == 0)
+      return path;
+
     /* If `path' is a symlink, follow it.  This is so that relative
        path references work. */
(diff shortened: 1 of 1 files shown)


src/libexpr/parser.y
```diff
@@ -577,7 +581,10 @@ Expr * EvalState::parseExprFromFile(const Path & path)
 
 Expr * EvalState::parseExprFromFile(const Path & path, StaticEnv & staticEnv)
 {
-    return parse(readFile(path).c_str(), path, dirOf(path), staticEnv);
+    if (path == "/dev/stdin")
```
aszlig: This is already handled by `nix-build -`, so we don't need to special-case this, see `EvalState::parseStdin`.

volth: wow, nice find
```nix build``` is less usable anyway, it does not print the results (https://github.com/NixOS/nix/pull/2622)

LnL7: https://github.com/NixOS/nix/pull/2583 does this for all commands that take the `-f` flag.

Issue #2523, opened 2018-11-08
darwin: nix-daemon crashes due to OBJC_DISABLE_INITIALIZE_FORK_SAFETY


periklis:
On macOS 10.14 i observe lately in my system.log that `nix-daemon` crashes due to the apple changes on `fork` since 10.13. The following message comes in regurarly:
```
nix-daemon[38220]: objc[38220]: +[__NSPlaceholderArray initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```
Related discussion in nixpkgs for other packages: https://github.com/NixOS/nixpkgs/issues/44160
Related info: http://sealiesoftware.com/blog/archive/2017/6/5/Objective-C_and_fork_in_macOS_1013.html

Do we need to update the SDK dependency for our buildsystem?

copumpkin:
Why is Objective C even relevant to `nix-daemon`?

LnL7:
Yeah, I don't understand this either. I'm running 10.14 on my new machine and don't have any problems with builds or the daemon.

copumpkin:
@periklis can you run your daemon with `DYLD_PRINT_LIBRARIES=1` in the environment?

periklis:
Here is the log: https://gist.github.com/periklis/a4357cfd21cd67dafa9122f9ca119592

Is it maybe because nix-daemon is run by launchctl that in turn can be an objc software?

peel:
I've hit the same thing. However... as @copumpkin noticed the problem is deep blow the nix stack and in fact is related to some awkward security features introduced somewhere in High Sierra, setting `export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` was enough to avoid crashing the daemon. However, that's not a proper solution (as per: https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/)

LnL7:
We got a bit closer to figuring out what's going on. The released nix binaries work fine, however using a build from nixpkgs master/unstable (on 10.14.1?) causes this problem.

[...]

PR #2674, opened 2019-02-09
nix-daemon: add variable to disable fork safety

LnL7:
Since macOS 10.14 this has become an error, causing problems if the
nix-daemon loads nix during substitution (this is a forked process).

Workaround for #2523.


matthewbauer:
This seems okay for now, but I wonder if Nix is doing something unsafe? It would be nice to track this down.

LnL7:
Yeah, we should try to fix it but I think this is safe since nothing uses the objective-c before forking happens.

matthewbauer:
pinging @edolstra

Issue #2650, opened 2019-01-24
installation without internet connection


zoranbosnjak:
Nix installation (via `install-multi-user`) fails if the server is not connected to the Internet.
It seems like the only problem is `/nix-channel --update` step. Please consider making this step optional or some other method, so that the installation is also possible without the Internet.

bjornfor:
Fixed on master: https://github.com/NixOS/nix/commit/989cb3777753024072160776ea37348689b89d44

bjornfor:
(I have don't have rights in this repo to close issues.)

zoranbosnjak:
It works as expected. 
Tested with `nix-2.3.10`, using both `install` and `install-multi-user` scripts.

PR #2679, opened 2019-02-13
install script: don't abort when "nix-channel --update" fails

bjornfor:
Instead, print a message about what happened and tell the user what can
be done (run "[sudo -i] nix-channel --update nixpkgs" again at a later
time). This change allows installing Nix when you're offline.

Since the multi-user installer is so verbose, I don't print the message
until the end.

Fixes issue #2650 ("installation without internet connection").

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -240,10 +240,16 @@ EOF
 }
 trap finish_fail EXIT
 
+channel_update_failed=0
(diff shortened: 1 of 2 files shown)


matthewbauer:
This looks reasonable to me.

/cc @grahamc @edolstra

bjornfor:
Thanks!

Issue #419, opened 2014-12-14
nix-shell ignoring SHELL enviroment variable


felix-eku:
I use `fish` as shell, but if I run `nix-shell` it drops me in a `bash`.

```
[fish]> echo $SHELL
/nix/store/a64lxj7f7c0hilynpjcys45x2m515s4y-fish-2.1.1/bin/fish
[fish]> nix-shell
[nix-shell]$ echo $SHELL
/nix/store/kly624h55gf91frdhn8fjzi6xricgkzk-bash-4.2-p53/bin/bash
[nix-shell]$ exit
[fish]> echo $SHELL
/nix/store/a64lxj7f7c0hilynpjcys45x2m515s4y-fish-2.1.1/bin/fish
```

This is the `shell.nix` I am using:

```
let packages = import <nixpkgs> {}; in
{
  stdenv ? packages.stdenv,
}:

stdenv.mkDerivation {
  name = "test";
  version = "0.1.0.0";
  src = builtins.toString ./.;
}
```

PR #2681, opened 2019-02-15
Support custom shell in nix-shell/nix run

haslersn:
_EDIT: Any changes to `nix-shell` have been removed from the PR. Only `nix run` is now affected._

* ~Use `$NIX_SHELL_INTERACTIVE_SHELL` (if present) as interactive shell for `nix-shell`~
* Use `$NIX_RUN_INTERACTIVE_SHELL` (if present) as interactive shell for `nix run`

If ~`--run`~, `--command` or `-c` are given, then bash is still used.

One could also consider those environment variables to be arbitrary default commands that are executed when ~`nix-shell`~/`nix run` are ran interactive and without a command. However, I think the most common use case (by far) is to set them to a shell, hence the name `[...]_INTERACTIVE_SHELL`.

~Tests:~

~I was unsuccessful adding tests to *tests/nix-shell.sh*. Any non-interactive test doesn't trigger the feature since it's only triggered in interactive shells.~

Previous work:

Issue #419 (nix-shell ignoring SHELL enviroment variable) proposed to use `$SHELL` as interactive shell for `nix-shell`. This is of course not desirable if people depend on `nix-shell` always using bash. ~With this change however, the feature is optional since it uses previously unused environment variables.~

### DIFF
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -21,7 +21,7 @@ std::string chrootHelperName = "__run_in_chroot";
 
 struct CmdRun : InstallablesCommand
 {
-    std::vector<std::string> command = { "bash" };
+    std::vector<std::string> command;
     StringSet keep, unset;
     bool ignoreEnvironment = false;
 
@@ -30,7 +30,8 @@ struct CmdRun : InstallablesCommand
         mkFlag()
             .longName("command")
             .shortName('c')
-            .description("command and arguments to be executed; defaults to 'bash'")
+            .description("command and arguments to be executed; defaults to "
+                "\"$NIX_RUN_INTERACTIVE_SHELL\" if set, otherwise to 'bash'")
             .labels({"command", "args"})
             .arity(ArityAny)
             .handler([&](std::vector<std::string> ss) {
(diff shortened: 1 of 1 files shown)


matthewbauer:
I think this makes sense with ```nix run```. But we use a lot of Bash-only stuff in Nixpkgs, so nix-shell seems like it would be broken.

haslersn:
> But we use a lot of Bash-only stuff in Nixpkgs, so nix-shell seems like it would be broken.

It's not broken since the environment setup still happens in bash. The shell set in `$NIX_SHELL_INTERACTIVE_SHELL` is ran from that bash afterwards.

LnL7:
A nix-shell is mainly intended for reproducing builds interactively, setting that variable completely breaks that which is a bad user experience IMHO. There are already tools, like [direnv](https://direnv.net), that handle the usecase you are looking for.

haslersn:
> A nix-shell is mainly intended for reproducing builds interactively, setting that variable completely breaks that which is a bad user experience IMHO. There are already tools, like [direnv](https://direnv.net), that handle the usecase you are looking for.

direnv looks appropriate for the job.

Since that's the common oppinion here, I'm no going to remove the `nix-shell` changes from the PR.

hyperfekt:
Given that, to my understanding, `nix run` is not used for building, and there is no analogous variable for `nix-shell`anymore, could it use `$SHELL` instead, and provide a good experience for users by default?

haslersn:
Certainly, but there's some possibility that it would break scripts that somehow use `nix run` fake-interactively. Not sure though whether such things exist.

hyperfekt:
Because of Github's inaccurate search it's very hard to ensure this but every single use of nix run in a script in a cursory glance is with `--command`.

hyperfekt:
Requesting this to be closed, seems to be made obsolete by #3477.

PR #2688, opened 2019-02-19
Fix nix-build --check -K in sandbox w/o root

tollb:
Temporarily add user-write permission to build directory so that it
can be moved out of the sandbox to the store with a .check suffix.

This is necessary because the build directory has already had its
permissions set read-only, but write permission is required
to update the directory's parent link to move it out of the sandbox.

Updated the related --check "derivation may not be deterministic"
messages to consistently use the real store paths.

Added test for non-root sandbox nix-build --check -K to demonstrate
issue and help prevent regressions.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -3537,6 +3537,29 @@ StorePathSet parseReferenceSpecifiers(Store & store, const BasicDerivation & drv
 }
 
 
+static void moveCheckToStore(const Path & src, const Path & dst)
+{
+    /* For the rename of directory to succeed, we must be running as root or
+       the directory must be made temporarily writable (to update the
+       directory's parent link ".."). */
+    struct stat st;
+    if (lstat(src.c_str(), &st) == -1) {
+        throw SysError(format("getting attributes of path '%1%'") % src);
+    }
+
+    bool changePerm = (geteuid() && S_ISDIR(st.st_mode) && !(st.st_mode & S_IWUSR));
+
+    if (changePerm)
+        chmod_(src, st.st_mode | S_IWUSR);
+
+    if (rename(src.c_str(), dst.c_str()))
+        throw SysError(format("renaming '%1%' to '%2%'") % src % dst);
+
+    if (changePerm)
+        chmod_(dst, st.st_mode);
+}
+
+
 void DerivationGoal::registerOutputs()
 {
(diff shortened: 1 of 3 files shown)


src/libstore/build.cc
```diff
@@ -3004,6 +3004,29 @@ PathSet parseReferenceSpecifiers(Store & store, const BasicDerivation & drv, con
...
+
+    bool changePerm = (geteuid() && S_ISDIR(st.st_mode) && !(st.st_mode & S_IWUSR));
+
+    if (changePerm)
+        chmod_(src, st.st_mode | S_IWUSR);
+
+    if (rename(src.c_str(), dst.c_str()))
+        throw SysError(format("renaming '%1%' to '%2%'") % src % dst);
+
+    if (changePerm)
```
LnL7, line 3025: I'm not sure if it's important to restore permissions here, the check path is for debugging only and isn't registered. Outputs of failed builds, etc. that have not been garbage collected yet generally also leave writable paths behind.

tollb: I appreciate your comment. I wasn't sure it was necessary to restore permissions but figured it was conservative and couldn't hurt. I'll probably leave it as is for now since it has been tested and seems slightly cleaner.

tollb:
@LnL7 Thanks for the quick follow-up. That's good news. I'll update and force push the rebased version after local testing.

src/libstore/build.cc
LnL7: Yeah, doesn't hurt. Don't forget to re-enable that test after rebasing. 😄

tollb: Thanks for all of your help. Will do.

tollb:
@LnL I force-pushed a rebased update. I will open a trivial PR to re-enable the disabled check for Darwin in PR #2689, once this gets merged. Let me know if there is anything else I should do.

domenkozar:
@tollb you could re-enable the disabled check here to be sure it's now working.

tollb:
@domenkozar Thanks for following-up. I re-enabled the disabled Darwin check (in a second commit) and the automated tests now pass. Please let me know if I need to do anything else.
CC @LnL7

domenkozar:
cc @edolstra I'm going to merge this, feel free to revert if you have objections.

PR #2689, opened 2019-02-19
Delete temporary directory on successful build

tollb:
With --check and the --keep-failed (-K) flag, the temporary directory
was being retained regardless of whether the build was successful and
reproducible.  This removes the temporary directory, as expected, on
a reproducible check build.

Added tests to verify that temporary build directories are not
retained unnecessarily, particularly when using --check with
--keep-failed.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1680,6 +1680,7 @@ void DerivationGoal::buildDone()
         }
 
         if (buildMode == bmCheck) {
+            deleteTmpDir(true);
             done(BuildResult::Built);
             return;
         }
--- a/tests/check.nix
+++ b/tests/check.nix
@@ -1,12 +1,45 @@
+{checkBuildId ? 0}:
+
 with import ./config.nix;
 
 {
   nondeterministic = mkDerivation {
+    inherit checkBuildId;
     name = "nondeterministic";
     buildCommand =
       ''
         mkdir $out
         date +%s.%N > $out/date
+        echo "CHECK_TMPDIR=$TMPDIR"
+        echo "checkBuildId=$checkBuildId"
+        echo "$checkBuildId" > $TMPDIR/checkBuildId
+      '';
+  };
+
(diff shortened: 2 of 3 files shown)


domenkozar:
Could you rebase?

tollb:
@domenkozar: I rebased the commit, adding status code checks to be consistent with @LnL7 updates that were committed after this PR was filed. I noticed one inconsistency with a return status for Darwin vs NixOS. Since I do not have access to a Darwin system, I disabled the one status code check for Darwin and left a comment in the code (check.sh).

LnL7:
@tollb That test shouldn't be just skipped, exit status 1 means it's a genuine error.

```
error: renaming '/private/var/folders/jf/2khhhlb53_jd54vf1wk60j2w0000gn/T/nix-test/store/vxvgpsip84469dl4mfjpa34ks33s85q7-nondeterministic' to '/private/var/folders/jf/2khhhlb53_jd54vf1wk60j2w0000gn/T/nix-test/store/jfwiq0apzyiq4hcb33zv33pa3qi114gg-nondeterministic.check': Permission denied
```

tollb:
@LnL7 Thanks for following-up! The message that you reported sounds like it might be related to a second PR #2688 that I filed around the same time (February 2019). I don't have the ability to test with Darwin and that PR no longer applies cleanly. I'm also not sure why the issue only affects Darwin at this point. However, I can rebase PR #2688 if you'd be willing to test it.

Assuming the rebased #2688 works and gets merged, I could open a new PR to re-enable the check for Darwin. Please let me know if you are willing to test a rebased PR #2688 or if you have other suggestions on how to proceed.
CC @domenkozar

LnL7:
@tollb Indeed, however from what I can tell renaming a readonly dir seems to work fine on linux so I'm surprised you ran into that.

As for why, sandboxing works quite differently on darwin. We don't have bind mounts so even when the sandbox is enabled hash rewriting is still used.

PR #2692, opened 2019-02-21
nix-daemon: emit a warning if untrusted user tries to set build parameters

Ma27:
It's documented in the NixOS wiki that untrusted users can't set build
machines using `--builders` or `--option builders`[1]. This should cause
a warning at least to avoid confusion why the options won't take any
effect.

This applies to all options except `connect-timeout`, `timeout` and
`builders` if the value is empty.

[1] https://nixos.wiki/wiki/Distributed_build

---

I tested the behavior with the following VM declaration built from the nixpkgs master:

``` nix
{
  nixpatch = { pkgs, ... }: {
    nix.package = pkgs.nix.overrideAttrs (old: {
      patches = [ ../nix/0001-nix-daemon-emit-a-warning-if-untrusted-user-tries-to.patch ];
    });

    users.extraUsers.vm = {
      password = "vm";
      isNormalUser = true;
    };
  };
}
```

### DIFF
--- a/src/nix-daemon/nix-daemon.cc
+++ b/src/nix-daemon/nix-daemon.cc
@@ -566,7 +566,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
                 else if (setSubstituters(settings.extraSubstituters))
                     ;
                 else
-                    debug("ignoring untrusted setting '%s'", name);
+                    warn("ignoring untrusted setting '%s'", name);
             } catch (UsageError & e) {
                 warn(e.what());
             }


Ma27:
Is there anything that keeps this from getting merged? :)

Ma27:
Ping @shlevy

Ma27:
After reviewing the diff, I figured that it's actually sufficient to use a `warn` rather than a `debug` log entry.

yorickvP:
This change is in master now, PR can be closed :)

PR #2693, opened 2019-02-22
scripts: remove default 'sandbox = false' from multi-user installer

thoughtpolice:
The multi-user installer script set `sandbox = false` unilaterally, overriding the new default configuration of `true` for Linux. Removing it allows the platform default to kick in.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -747,7 +747,6 @@ build-users-group = $NIX_BUILD_GROUP_NAME
 
 max-jobs = $NIX_USER_COUNT
 cores = 1
-sandbox = false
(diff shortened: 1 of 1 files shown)


edolstra:
Maybe it's better to not emit `sandbox = ...` at all? That way it will use the builtin default value (false on macOS, true on Linux).

LnL7:
I could be wrong, but I think only `build-users-group` is required for a multi-user install.

thoughtpolice:
@edolstra Fixed, and confirmed to work (defaulting to `sandbox = true`) on a brand new Ubuntu 18.04 machine with the multi-user installer.

Issue #1751, opened 2017-12-20
nix-collect-garbage --dry-run does not list deleted garbage


jtojnar:
Compare

```sh-session
# nix-collect-garbage --delete-older-than 10d --dry-run
removing old generations of profile /nix/var/nix/profiles/system
removing old generations of profile /nix/var/nix/profiles/per-user/root/channels
removing old generations of profile /nix/var/nix/profiles/per-user/jtojnar/profile
would remove generation 425
removing old generations of profile /nix/var/nix/profiles/per-user/jtojnar/channels
removing old generations of profile /nix/var/nix/profiles/default
```

with

```sh-session
# nix-collect-garbage --delete-older-than 10d
removing old generations of profile /nix/var/nix/profiles/system
removing old generations of profile /nix/var/nix/profiles/per-user/root/channels
removing old generations of profile /nix/var/nix/profiles/per-user/jtojnar/profile
removing generation 425
removing old generations of profile /nix/var/nix/profiles/per-user/jtojnar/channels
removing old generations of profile /nix/var/nix/profiles/default
finding garbage collector roots...
removing stale link from ‘/nix/var/nix/gcroots/auto/60vijjwplhs2ryrc4dajd3rjkm53mzxf’ to ‘/run/user/1000/nix-build.SXkyLF/derivation-50’
removing stale link from ‘/nix/var/nix/gcroots/auto/659v5p9mli2yz9kwmnqn0frk59jczajm’ to ‘/run/user/1000/nix-build.SXkyLF/derivation-49’
[...]
deleting garbage...
deleting ‘/nix/store/ps6w4s329lyz093r90pszqbb9ybnvzs4-evolution-data-server-3.26.3-dev.lock’
deleting ‘/nix/store/4f3vd9c4r5x43iw4rrd8jdfxbyak3zzh-evolution-data-server-3.26.3.lock’
deleting ‘/nix/store/yss2s8zxxqkcdipfysn1g6p9ymblxm7j-gnome-desktop-3.26.2.lock’
deleting ‘/nix/store/m4vq8y0q1y34mpp6dvj5yhsjif5kr86g-anjuta-3.26.0.lock’
[...]
deleting ‘/nix/store/trash’
deleting unused links...
note: currently hard linking saves -0.00 MiB
992 store paths deleted, 2725.86 MiB freed
```

[...]

PR #2695, opened 2019-02-23
nix-collect-garbage: Make the --dry-run option work, as long as no generations are being deleted.

DavidEGrayson:
This is the first step of fixing issue #1751: it makes `nix-collect-garbage --dry-run` work.  It also prints an error if `--dry-run` is used in conjunction with the options for deleting generations, because we don't have the machinery needed to print which outputs reachable from those old generations are going to be deleted.

The code in this commit is similar to [lines 599-604 of nix-store.cc](https://github.com/NixOS/nix/blob/5f1891b/src/nix-store/nix-store.cc#L599-L604).

CC: @shlevy

### DIFF
--- a/src/nix-collect-garbage/nix-collect-garbage.cc
+++ b/src/nix-collect-garbage/nix-collect-garbage.cc
@@ -76,19 +76,30 @@ static int _main(int argc, char * * argv)
             return true;
         });
 
+        if (removeOld && dryRun)
+        {
+            // removeOldGenerations does not record any information
+            // about which generations would be deleted if dryRun is true, so
+            // we cannot get an accurate list of store paths to be deleted.
+            throw UsageError("Sorry, the --dry-run option is not yet "
+                "compatible with deleting generations.");
+        }
+
         initPlugins();
 
         auto profilesDir = settings.nixStateDir + "/profiles";
         if (removeOld) removeOldGenerations(profilesDir);
 
         // Run the actual garbage collector.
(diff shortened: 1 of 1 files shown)


src/nix-collect-garbage/nix-collect-garbage.cc
```diff
@@ -76,19 +76,31 @@ static int _main(int argc, char * * argv)
             return true;
         });
 
+        if (removeOld && dryRun)
+        {
+            // removeOldGenerations does not record any information
+            // about which generations would be deleted if dryRun is true, so
+            // we cannot get an accurate list of store paths to be deleted.
+            std::cerr << "error: Sorry, the --dry-run option is not yet "
+                "compatible with deleting generations." << std::endl;
```
edolstra: This should be `throw UsageError("bla bla");`.

DavidEGrayson: Thanks for taking a look!  I amended my commit so it uses `UsageError` now.

bew:
Hello!

I've juste been hit by #1751 today (as a new nix user), I see there has not been any work on this since Feb 2019.

@DavidEGrayson I hope you're doing well in the current world situation, are you going to continue working on this subject?

Thanks!

DavidEGrayson:
This pull request has been ignored for over a year. I'm not really interested in doing more work if my previous work is ignored.   I'm glad you at least see the value in it, @bew.

PR #2698, opened 2019-02-24
Add --whole-archive to include all store implementations (when BUILD_SHARED_LIBS=false)

matthewbauer:
When statically linking, symbols that are not referenced are not
included in the final executable. Usually, this is an okay
optimization, but sometimes can cause unexpected issues.

In this case, statically linked Nix was not finding the
HttpBinaryCacheStore class and the RegisterStoreImplementation that
comes with it. This is because the resulting binary would not run all
of the store registration at initialization time. Some of the store
registrations were still heppning like in the S3 case, but only
because another symbol in s3-binary-cache-store.o was being used
elsewhere in Nix. This linking behavior caused static Nix to have
errors like:

  don't know how to open Nix store ‘https://...’

Adding --whole-archive forces the linker to include these archives.
This is just applied to the local Nix libraries, and not external
libraries. As a side effect we might get some symbols that we don’t
actually need, but that is the same case as with dynamically linked
Nix.

You can try out my statically built pr here:

https://github.com/NixOS/nixpkgs/pull/56281

### DIFF
--- a/mk/libraries.mk
+++ b/mk/libraries.mk
@@ -127,7 +127,7 @@ define build-library
     $$($(1)_PATH): $$($(1)_OBJS) | $$(_d)/
 	$(trace-ar) $(AR) crs $$@ $$?
 
-    $(1)_LDFLAGS_USE += $$($(1)_PATH) $$($(1)_LDFLAGS)
+    $(1)_LDFLAGS_USE += -Wl,--whole-archive $$($(1)_PATH) -Wl,--no-whole-archive $$($(1)_LDFLAGS)
 
     $(1)_INSTALL_PATH := $$(libdir)/$$($(1)_NAME).a
 
--- a/mk/libraries.mk
+++ b/mk/libraries.mk
@@ -50,6 +50,11 @@ endif
 #   target andwill not be listed in the make help output. This is
 #   useful for libraries built solely for testing, for example.
 #
+# - $(1)_WHOLE_ARCHIVE: if defined, the whole archive of the static
+#   library will be used. This is necessary when using global
+#   constructors such as setting up a RegisterStoreImplementation.
+#   Only applies when BUILD_SHARED_LIBS=0.
+#
 # - BUILD_SHARED_LIBS: if equal to ‘1’, a dynamic library will be
 #   built, otherwise a static library.
 define build-library
@@ -136,16 +141,20 @@ define build-library
     # -force_load that seems to work similarly, but takes one argument
     # instead of a group of arguments.
 
-    ifeq ($(OS), Linux)
-      $(1)_LDFLAGS_USE += -Wl,--whole-archive
-    else ifeq ($(OS), Darwin)
-      $(1)_LDFLAGS_USE += -Wl,-force_load
+    ifdef $(1)_WHOLE_ARCHIVE
+      ifeq ($(OS), Linux)
+        $(1)_LDFLAGS_USE += -Wl,--whole-archive
+      else ifeq ($(OS), Darwin)
+        $(1)_LDFLAGS_USE += -Wl,-force_load
+      endif
     endif
 
     $(1)_LDFLAGS_USE += $$($(1)_PATH)
 
-    ifeq ($(OS), Linux)
-      $(1)_LDFLAGS_USE += -Wl,--no-whole-archive
+    ifdef $(1)_WHOLE_ARCHIVE
+      ifeq ($(OS), Linux)
+        $(1)_LDFLAGS_USE += -Wl,--no-whole-archive
+      endif
     endif
 
     $(1)_LDFLAGS_USE += $$($(1)_LDFLAGS)
--- a/mk/libraries.mk
+++ b/mk/libraries.mk
@@ -127,7 +127,28 @@ define build-library
     $$($(1)_PATH): $$($(1)_OBJS) | $$(_d)/
 	$(trace-ar) $(AR) crs $$@ $$?
 
-    $(1)_LDFLAGS_USE += -Wl,--whole-archive $$($(1)_PATH) -Wl,--no-whole-archive $$($(1)_LDFLAGS)
+    # Special care is needed to support C++ global initializers, used
+    # in RegisterStoreImplementation. The .init section needs to be in
+    # the linked executable for it find the registered stores. On
+    # Linux systems systems, wrap the linked object archive with whole
+    # archive flags. Note that --whole-archive is only supported with
+    # the bfd, gold, and lld linkers. The macOS ld64 linker has a
+    # -force_load that seems to work similarly, but takes one argument
+    # instead of a group of arguments.
+
+    ifeq ($(OS), Linux)
+      $(1)_LDFLAGS_USE += -Wl,--whole-archive
+    else ifeq ($(OS), Darwin)
+      $(1)_LDFLAGS_USE += -Wl,-force_load
+    endif
+
+    $(1)_LDFLAGS_USE += $$($(1)_PATH)
+
+    ifeq ($(OS), Linux)
+      $(1)_LDFLAGS_USE += -Wl,--no-whole-archive
+    endif
+
+    $(1)_LDFLAGS_USE += $$($(1)_LDFLAGS)
 
     $(1)_INSTALL_PATH := $$(libdir)/$$($(1)_NAME).a
 
--- a/src/libexpr/local.mk
+++ b/src/libexpr/local.mk
@@ -15,6 +15,8 @@ ifneq ($(OS), FreeBSD)
  libexpr_LDFLAGS += -ldl
 endif
 
+libexpr_WHOLE_ARCHIVE = 1
+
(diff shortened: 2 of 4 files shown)


mk/libraries.mk
```diff
@@ -127,7 +127,7 @@ define build-library
     $$($(1)_PATH): $$($(1)_OBJS) | $$(_d)/
 	$(trace-ar) $(AR) crs $$@ $$?
 
-    $(1)_LDFLAGS_USE += $$($(1)_PATH) $$($(1)_LDFLAGS)
+    $(1)_LDFLAGS_USE += -Wl,--whole-archive $$($(1)_PATH) -Wl,--no-whole-archive $$($(1)_LDFLAGS)
```
grahamc: why first one then the other?

matthewbauer: We only want the whole archive of the local Nix libraries like libnixstore.a. So this enables ```--whole-archive``` for one .a file then immediately disables it. For most dependencies, we only want to include the symbols actually used by the binary. If we enabled ```--whole-archive``` for all libraries, we would end up with tons of unused symbols from libcurl, aws, boehm-gc, etc.

The one case where we need ```--whole-archive``` is for libnixstore.a which has an ```.init``` elf section we need to keep. This comes from the reliance on C++ Global Constructors feature. Some references on this:

 - http://l4u-00.jinr.ru/usoft/WWW/www_debian.org/Documentation/elf/node4.html#SECTION00031000000000000000
- https://github.com/bazelbuild/rules_go/issues/1486
- https://www.bfilipek.com/2018/02/static-vars-static-lib.html

edolstra:
Doesn't this cause the `nix` executable to be potentially much larger than it should be?

matthewbauer:
> Doesn't this cause the `nix` executable to be potentially much larger than it should be?

Yeah, but mostly only because those symbols are missing.

```
$ du /nix/store/gnyc6ywn4zsc61w3xql1b8vr5l55qp0m-nix-2.2.2/bin/nix
1668	/nix/store/gnyc6ywn4zsc61w3xql1b8vr5l55qp0m-nix-2.2.2/bin/nix
$ du /nix/store/7hygziif4mx6bi33ija2iicwjrn1l1dv-nix-2.3pre6629_6587ffd/bin/nix
4548	/nix/store/7hygziif4mx6bi33ija2iicwjrn1l1dv-nix-2.3pre6629_6587ffd/bin/nix
```
To make things smallers, I've added a flag that just applies it for libstore.a. It doesn't help too much though:

```
$ du /nix/store/x2xlbs41grl4j60qyb4v3yyki8mc8mhq-nix-2.3pre0_0000000/bin/nix
4312	/nix/store/x2xlbs41grl4j60qyb4v3yyki8mc8mhq-nix-2.3pre0_0000000/bin/nix
```



Try:

```
(import ./release.nix {}).build.x86_64-linux.overrideAttrs (old: {
  configureFlags = old.configureFlags or [] ++ [ "--disable-shared" ];
})
```

```
$ nix build -f test.nix
$ ./result/bin/nix run nixpkgs.hello
```

Notice that the binary cache isn't working, because the regStore symbols aren't linked into the executable!

matthewbauer:
Pinging @edolstra @grahamc once more. This does increase static executable sizes, but is also necessary to use any of the store implementations. Without it, static Nix is unusable for anything like the binary cache (because HttpBinaryCacheStore class is missing).

matthewbauer:
I also should note this just effects ```--disable-shared``` builds.

edolstra:
`--whole-archive` seems like a sledgehammer approach. Surely there must be a better way to force variables with constructors to be included. https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes mentions a `used` attribute: "This attribute, attached to a variable with static storage, means that the variable must be emitted even if it appears that the variable is not referenced." Maybe that does the trick?

matthewbauer:
> `--whole-archive` seems like a sledgehammer approach. Surely there must be a better way to force variables with constructors to be included. https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes mentions a `used` attribute: "This attribute, attached to a variable with static storage, means that the variable must be emitted even if it appears that the variable is not referenced." Maybe that does the trick?

I don't think there is. ```__attribute__((used))``` is a compiler concept and is never passed to the linker from what I can tell. I have a test branch to try this here:

https://github.com/matthewbauer/nix/tree/fix-static-linking-used

It is still missing the necessary symbols. Here is a more concise sample that shows what happens:

https://stackoverflow.com/questions/1202494/why-doesnt-attribute-constructor-work-in-a-static-library

Two alternatives are possible:

- Rework the build scripts to get rid of "libnixstore.a" and compile libnixstore sources directly with the nix binary.
- Use ```ifdef``` to manually run the regStore constructors when we know we are linking to static libnixstore.a.

Both of these end up special casing static compilation in awkward ways. The second one is especially bad because we need to update it every time we add a new instance of RegisterStoreImplementation. I still think ```--whole-archive``` is the best solution here.

saschagrunert:
Do you think that we can go forward with this PR somehow? I would love to see a recent static nix version. :)

matthewbauer:
At NixCon, @edolstra wanted to see if https://github.com/NixOS/nix/pull/3154 would fix this. Building it right now.

matthewbauer:
@edolstra Unfortunately #3154 doesn't fix this. We still need ```--whole-archive``` here.

tomberek:
Would it make sense to proceed for now with `--whole-archive` as a sledgehammer for the static build, then adopt whatever might be better in the future?

matthewbauer:
> Would it make sense to proceed for now with `--whole-archive` as a sledgehammer for the static build, then adopt whatever might be better in the future?

This would put in quite a few libraries we don't need like the whole aws-cpp-sdk, openssl, sqlite3, etc.

PkmX:
I also hit this issue with a statically-linked `nix`. This also affects `libexpr` where some primops (e.g. `fetchGit`) would not get pulled unless `--whole-archive` is used.

PkmX:
For reference, `libfetchers` in the laster master also needs the `--whole-archive` flag.

Mic92:
Is store-api.hh supposed to support store implementations that are not part of the nix code base? Otherwise just having a list of initialiser would be simple and straightforward.

Mic92:
All classes that make use of static initializers:

`RegisterPrimOp`, `RegisterLegacyCommand`, `RegisterStoreImplementation` and `GlobalConfig::Register::Register`

Mic92:
I opened https://github.com/NixOS/nix/pull/3511 as an alternative fix.

matthewbauer:
> Is store-api.hh supposed to support store implementations that are not part of the nix code base? 

I think plugins are allowed to provide their own implementation.

Mic92:
Since upstream is not interested in a fix for this problem, should we have public fork that also provides static binaries for download?

edolstra:
It's not that I don't care about a fix, but it seems that the real bug here is not running static initializers. I'm not sure but I would think that C++ requires them to be executed.

Mic92:
This issue is not fixable on a language level without refactorings like https://github.com/NixOS/nix/pull/3511.
The linker who does not care about the language won't include object files into the binary unless at least one symbol from it is used unless `--whole-archive` is used. I also don't see any downside from using this options. All object files build are eventually used in the application anyway so it does not make the build less efficient or bigger.

shlevy:
@Mic92 If a compiler has an output mode that doesn't respect the semantics of the language, then it's a bug in the compiler.

matthewbauer:
> It's not that I don't care about a fix, but it seems that the real bug here is not running static initializers. I'm not sure but I would think that C++ requires them to be executed.

I think the big question is whether C++ allows unused static initializers to be subject to dead code elimination. Someone should try to build Nix dynamically with `--gc-sections` set. If you don't get the observed bug with `--gc-sections`, I would say this is a linker bug. Otherwise, I think it's a Nix bug.

Another thing to try is: instead of linking the executable from compiled sources, directly compile and link the source in one go (don't set `-c`) - gcc might be smart enough to know how to handle this edge case. This kind of screws up the separation between the Nix libraries but may be useful if it actually avoid the problem.

Kha:
This behavior is specified, or at least allowed, by the C++ standard (if I'm reading it correctly):
> If no variable or function is odr-used from a given translation unit, the non-local variables defined in that translation unit may never be initialized (this models the behavior of an on-demand dynamic library).
https://en.cppreference.com/w/cpp/language/initialization#Non-local_variables

That is, if you do not use a source file at all, the compiler is free to forego initialization.

matthewbauer:
Closing in favor of https://github.com/NixOS/nix/pull/2698 since i think that is a better approach. It makes static and dynamic linking closer to each other.

xaverdh:
> Closing in favor of #2698 since i think that is a better approach. It makes static and dynamic linking closer to each other.

I think you meant to reference pr 3677 ?

PR #2700, opened 2019-02-24
Get relative path of NIX_DATA_DIR

matthewbauer:
This is useful when you want Nix to look in a relative path at
runtime. For instance, you can make Nix work without a Nix store dir
by building like this:

$ make datadir=./share

and you just need “share” to be in the same directory as the Nix
binary. If statically linked properly, no references to $prefix should
remain.


matthewbauer:
Nevermind, I can just set NIX_DATA_DIR to ```$PWD/NIX_DATA_DIR``` at run time.

PR #2703, opened 2019-02-25
Support parameters in S3 URLs

pradd:
This pull request adds proper support for S3 URL parameters.

Currently, an attempt to download artifacts with fetchTarball or fetchUrl fails for buckets residing in regions other than us-east-1. Nix manual states explicitly that, when using another region, the "region"
parameter should be passed. Which doesn't work since us-east-1 is hardcoded.

### DIFF
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -614,6 +614,22 @@ struct CurlDownloader : public Downloader
         writeFull(wakeupPipe.writeSide.get(), " ");
     }
 
+#ifdef ENABLE_S3
+    std::tuple<std::string, std::string, Store::Params> parseS3Uri(std::string uri)
+    {
+        auto [path, params] = splitUriAndParams(uri);
+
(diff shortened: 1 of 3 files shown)


src/libstore/download.cc
```diff
@@ -614,6 +614,22 @@ struct CurlDownloader : public Downloader
...
     }
 
+#ifdef ENABLE_S3
+    std::tuple<std::string, std::string, Store::Params> parseS3Uri(std::string uri)
+    {
+        auto [path, params] = splitUriAndParams(uri);
+
+        auto slash = path.find('/', 5); // 5 is the length of "s3://" prefix
+           if (slash == std::string::npos)
+               throw nix::Error("bad S3 URI '%s'", path);
```
edolstra: Indentation looks slightly off here.

edolstra:
Looks good to me!

src/libstore/download.cc
pradd: fixed

copumpkin:
Nice, thanks!

Issue #625, opened 2015-08-27
Why not use CLONE_NEWUSER for builders?


copumpkin:
Seems like use of that mostly patches up most issues people have with escaping from namespaces, and allows the namespace creator to be run with lower power. Unless I'm misunderstanding!

lethalman:
Why should you use it? It's useful for entering a chroot without root privileges, hence I don't see much of a use for builders.

Instead, it would be very interesting for nix-shell to enforce more purity. Like nix-shell --chroot will enter a nix-build like chroot without root privileges.

copumpkin:
@lethalman I thought that most of the residual concerns around "containers are insecure" were obviated by CLONE_NEWUSER from an unprivileged account, since it allows the nix daemon to create the builder namespaces without being root. It might still need root for other things, of course. May be unnecessary, but also doesn't seem particularly harmful.

lethalman:
@copumpkin yes, if nix-daemon setuids before doing the chroot and uses CLONE_NEWUSER with proper uid/gid mappings, that's a some more security.

Ericson2314:
If this allows the nix daemon to be run without root or sacrificing (key) features, that's great!

I want to run nix at school (multi user mode so people besides myself can benefit), and its essential that nothing uses root so the sysadmins don't have reason to be worried.

Ericson2314:
The nix-shell stuff would also be nice. IIUC since nix-shell essentially sets one up as if they were a builder, implementing this for builds should give it to us for shells, and vice versa?

lethalman:
@Ericson2314 yes, if nix-build is able to use CLONE_NEWUSER, then nix-shell can take advantage of the same (or almost) isolation level.

orenbenkiki:
I would dearly love to see #324 supported out of the box, so :+1: - are there any downsides?

fhaust:
I second that. Especially in a distributed environment I think it is unreasonable to ask the admins to run the nix-daemons on every single machine.

aszlig:
I have worked on this a [while ago](https://gist.github.com/aszlig/1dccb5dc09e6235843fa), but can't find the WIP code right now. Will dig it up and push it once I'm home by the end of the week, if that helps.

copumpkin:
Awesome! Please do

edolstra:
[...]

PR #2717, opened 2019-03-10
set up user namespace in builder, not in parent

catern:
This is a minor simplification; there's now slightly less handshaking
required between the builder and the parent, and we create one less
pipe when starting the builder.

As a niche benefit, this also replaces our access to /proc/builder_pid
in the parent with an access to /proc/self in the builder. The former
access could fail if /proc was mounted in a different pid namespace
from the one we are running in; now we will work even if there is such
a misconfiguration.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -766,6 +766,12 @@ private:
     /* User selected for running the builder. */
     std::unique_ptr<UserLock> buildUser;
 
+    /* The uid we will use for the build, outside any sandboxing. */
+    uid_t hostUid;
+
+    /* The gid we will use for the build, outside any sandboxing. */
+    gid_t hostGid;
+
     /* The process ID of the builder. */
     Pid pid;
 
@@ -792,9 +798,6 @@ private:
 
     /* Pipe for the builder's standard output/error. */
     Pipe builderOut;
-
-    /* Pipe for synchronising updates to the builder user namespace. */
-    Pipe userNamespaceSync;
 
     /* The build hook. */
     std::unique_ptr<HookInstance> hook;
@@ -1846,6 +1849,8 @@ void DerivationGoal::startBuilder()
     if (settings.buildUsersGroup != "" && getuid() == 0) {
 #if defined(__linux__) || defined(__APPLE__)
         buildUser = std::make_unique<UserLock>();
+        hostUid = buildUser->getUID();
+        hostGid = buildUser->getGID();
 
         /* Make sure that no other processes are executing under this
            uid. */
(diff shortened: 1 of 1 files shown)


catern:
@edolstra ping?

catern:
Still important to me

tomberek:
Does this potentially allow the child to make additional changes to its own namespace? I'm not super familiar with the origin of this, but I'd like to understand and it looks like it is intentional.

Ericson2314:
@catern Are you still interested in this?

I am assigned to get it shepherded over the finish line if so, or closed otherwise.

catern:
I regret to say that I'm no longer interested in this

tomberek:
> I regret to say that I'm no longer interested in this

Can we ask for an explanation of the intent and intended benefit? As above, I’m not quite sure of what all the consequences would be.

Ericson2314:
Yeah @catern if it is "just what it says on the tin" that is fine, but if there was a sharable use-case separate from the technical details that it isn't too much effort to remember on your part that would be good to know.

I might take a look at rebasing this anyways :).

catern:
It's just what it says on the tin. I made this change after writing some container code which made a new pid namespace but didn't remount /proc (which is buggy behavior) and then running Nix inside it. This change made Nix work but also my container code was buggy in the first place.

There's no actual use case for this change, except that it slightly simplifies the code.

Ericson2314:
After like 10 rebases :) I have https://github.com/NixOS/nix/compare/master...obsidiansystems:userns

Ericson2314:
@edolstra Do you know why you had the parent not child set the user namespace originally? As @tomberek says, this is a good case of "Chesterton's Fence" --- we should ask that question before undoing this.

(Also, In the meeting I recall you saying you thought you had done this since, but the rebase I linked above demonstrates this is not the case.)

I had done a `git log -G userNamespaceSync` which led me to c68e5913c71badc89ff346d1c6948517ba720c93, which says:

> Run builds in a user namespace
>
> This way, all builds appear to have a uid/gid of 0 inside the
> chroot. In the future, this may allow using programs like
> systemd-nspawn inside builds, but that will require assigning a larger
> UID/GID map to the build.
>
> Issue #625.

And neither that nor issue #625 seem to shed any light on this implementation detail.

edolstra:
I don't remember why it's done this way. Maybe the kernel required it at the time?

Ericson2314:
Thanks @edolstra. Yes I too vaguely recall user namespace used to be very finicky. That is also my best guess.

Ericson2314:
Rebased as #8023

PR #2720, opened 2019-03-11
primops: disallow readFile /dev/stdin

### LINKED ISSUE #2716: builtins.getFile /dev/stdin doesn't work on linux
This might be another inconsistency between libstdc++ and libc++.

```bash
darwin $ echo foo | nix-instantiate --eval -E 'builtins.readFile /dev/stdin'
"foo"
```
```bash
linux $ echo foo | nix-instantiate --eval -E 'builtins.readFile /dev/stdin'
""
```

LnL7:
Fixes #2716

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -865,8 +865,12 @@ static void prim_readFile(EvalState & state, const Pos & pos, Value * * args, Va
 {
     PathSet context;
     Path path = state.coerceToPath(pos, *args[0], context);
-    if (path == "/dev/stdin")
-        throw EvalError(format("cannot read from stdin, at %1%") % pos);
+
+    struct stat st;
+    stat(path.c_str(), &st);
+    if (S_ISCHR(st.st_mode) || S_ISBLK(st.st_mode) || S_ISFIFO(st.st_mode) || S_ISSOCK(st.st_mode))
+        throw EvalError(format("cannot read from character special files, at %1%") % pos);
+
     try {
         state.realiseContext(context);
     } catch (InvalidPathError & e) {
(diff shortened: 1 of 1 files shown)


grahamc:
I wonder if this should be restricted somehow so it doesn’t end up in the middle of nixpkgs. For example what happens if it is used twice?

LnL7:
If stdin is read more then once subsequent calls will return an empty string since it's consumed already.

shlevy:
@LnL7 The kosher thing to do here would be to have some state in the eval state to determine if stdin has been read and if so what its contents were.

edolstra:
IMHO this is a really bad idea, reading stdin is very impure and very non-compositional (for example, consider programs like `nixos-rebuild` or NixOps where an expression might be evaluated multiple times).

edolstra:
If the goal is to pass large strings, it might be better to add some flag for this, e.g. `nix-instantiate --arg-from-file foobar /dev/stdin`.

LnL7:
I like the idea of `--arg-from-file`, do I change this to throw an error so it fails everywhere then?

edolstra:
@LnL7 Hm, I don't understand the question, what should throw an error?

LnL7:
I made this to make the behaviour between darwin and linux consistent, so if we don't want this I would prefer to also disable the current behaviour on darwin.

lheckemann:
I'm not sure this would be sufficient, e.g. `builtins.readFile /dev/fd/0` will still behave the same I presume. Perhaps checking the file type using `fstat(0, &st)` and only allowing regular files would be more appropriate.

LnL7:
Hmm guess I didn't push my second commit, that's almost what I did :smile.

lheckemann:
Maybe it should use `S_ISREG` instead — I don't think we want to allow FIFOs either?

LnL7:
Using `S_ISREG` doesn't work for symlinks without resolving them first so I extended the blacklist instead.

lheckemann:
Yes it does, as long as you use `stat` and not `lstat`.

PR #2723, opened 2019-03-14
Fix build with C++17 and C++20, fix with libc++

dtzWill:


edolstra:
Thanks, rebased and applied.

PR #2724, opened 2019-03-14
nix-store: document --add-fixed

LnL7:
```
OPERATION --ADD-FIXED
   Synopsis
       nix-store [--recursive] --add-fixed algorithm paths...

   Description
       The operation --add-fixed adds the specified paths to the Nix store. Unlike --add
       paths are registred using the specified hashing algorithm, resulting in the same
       output path as a fixed output derivation. This can be used for sources that are not
       available from a public url or broke since the download expression was written.

       This operation has the following options:

       --recursive
           Use recursive instead of flat hashing mode, used when adding directories
	   to the store.

   Example
           $ nix-store --add-fixed sha256 ./hello-2.10.tar.gz
           /nix/store/3x7dwzq014bblazs7kq20p9hyzz0qh8g-hello-2.10.tar.gz
```

### DIFF
--- a/doc/manual/command-ref/nix-store.xml
+++ b/doc/manual/command-ref/nix-store.xml
@@ -883,6 +883,60 @@ $ nix-store --add ./foo.c
 
 </refsection>
 
+<!--######################################################################-->
+
+<refsection><title>Operation <option>--add-fixed</option></title>
+
+<refsection><title>Synopsis</title>
+
+<cmdsynopsis>
+  <command>nix-store</command>
+  <arg><option>--recursive</option></arg>
+  <arg choice='plain'><option>--add-fixed</option></arg>
+  <arg choice='plain'><replaceable>algorithm</replaceable></arg>
+  <arg choice='plain' rep='repeat'><replaceable>paths</replaceable></arg>
+</cmdsynopsis>
+
+</refsection>
+
+<refsection><title>Description</title>
+
+<para>The operation <option>--add-fixed</option> adds the specified paths to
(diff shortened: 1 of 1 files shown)


doc/manual/command-ref/nix-store.xml
```diff
@@ -883,6 +883,57 @@ $ nix-store --add ./foo.c
...
+url or broke since the download expression was written.
+</para>
+
+<para>This operation has the following options:
+
+<variablelist>
+
+  <varlistentry><term><option>--recursive</option></term>
+
+    <listitem><para>Use recursive instead of flat hashing mode.</para></listitem>
```
lheckemann: > , in order to add directories to the store.

?

LnL7:
ping

doc/manual/command-ref/nix-store.xml
```diff
@@ -883,6 +883,60 @@ $ nix-store --add ./foo.c
...
+  <arg choice='plain'><replaceable>algorithm</replaceable></arg>
+  <arg choice='plain' rep='repeat'><replaceable>paths</replaceable></arg>
+</cmdsynopsis>
+
+</refsection>
+
+<refsection><title>Description</title>
+
+<para>The operation <option>--add-fixed</option> adds the specified paths to
+the Nix store.  Unlike <option>--add</option> paths are registred using the
```
grahamc:
```suggestion
the Nix store.  Unlike <option>--add</option> paths are registered using the
```

PR #2737, opened 2019-03-19
build: add setting to extend the default darwin sandbox profile

LnL7:
This makes the default profile configurable with extra capabilities
using nix.conf without having to resort to relaxed mode and expressions
using `__sandboxProfile`. eg.

	(allow ipc-sysv-shm ipc-sysv-sem)


LnL7:
ping

SuperSandro2000:
Ping

PR #2741, opened 2019-03-21
eval: improve type description for primops and applied primops

lheckemann:
This can make type errors a little easier to understand.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -130,6 +130,16 @@ std::ostream & operator << (std::ostream & str, const Value & v)
 }
 
 
+const Value *getPrimOp(const Value &v) {
+    const Value * primOp = &v;
+    while (primOp->type == tPrimOpApp) {
+        primOp = primOp->primOpApp.left;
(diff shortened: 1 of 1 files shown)


src/libexpr/eval.cc
```diff
@@ -144,8 +154,10 @@ string showType(const Value & v)
         case tApp: return "a function application";
         case tLambda: return "a function";
         case tBlackhole: return "a black hole";
-        case tPrimOp: return "a built-in function";
-        case tPrimOpApp: return "a partially applied built-in function";
+        case tPrimOp:
+            return "the built-in function " + string(v.primOp->name);
+        case tPrimOpApp:
+            return "the partially applied built-in function " + string(getPrimOp(v)->primOp->name);
```
edolstra: Can you add single quotes around the function name, e.g.
```
return fmt("the built-in function '%s'", string(v.primOp->name));
```

lheckemann: Done, thanks for the feedback!

PR #2744, opened 2019-03-26
manual: mention the "channel:" shorthand for NIX_PATH

veprbl:
This is a small step towards modernizing manual for recent nix. The cool new features should be mentioned not only in the release notes.

### DIFF
--- a/doc/manual/command-ref/env-common.xml
+++ b/doc/manual/command-ref/env-common.xml
@@ -52,10 +52,15 @@ nixpkgs=/home/eelco/Dev/nixpkgs-branch:/etc/nixos</screen>
     <envar>NIX_PATH</envar> to
 
     <screen>
-nixpkgs=https://github.com/NixOS/nixpkgs-channels/archive/nixos-14.12.tar.gz</screen>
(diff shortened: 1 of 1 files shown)


doc/manual/command-ref/env-common.xml
```diff
@@ -52,10 +52,15 @@ nixpkgs=/home/eelco/Dev/nixpkgs-branch:/etc/nixos</screen>
...
-nixpkgs=https://github.com/NixOS/nixpkgs-channels/archive/nixos-14.12.tar.gz</screen>
+nixpkgs=https://github.com/NixOS/nixpkgs-channels/archive/nixos-15.09.tar.gz</screen>
 
     tells Nix to download the latest revision in the Nixpkgs/NixOS
-    14.12 channel.</para>
+    15.09 channel.</para>
+
+    <para>A following shorthand can be used to refer to the official channels:
+    
+    <screen>nixpkgs=channel:nixos-15.09</screen>
```
grahamc: Mind using a more recent channel? :) like nixos-unstable for example.

veprbl: But... isn't it more amazing to be able to install firefox from 2015? I think it is.

PR #2745, opened 2019-03-27
install-multi-user: Detect and fail lack of systemd separately

samueldr:
Otherwise, the user is shown:

```
Sorry, I don't know what to do on Linux
```

Which is... not exactly right.

* * *

See

 * https://logs.nix.samueldr.com/nixos/2019-03-27#1553647070-1553646943;

cc @grahamc install script afficionado.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -753,9 +753,13 @@ main() {
     if [ "$(uname -s)" = "Darwin" ]; then
         # shellcheck source=./install-darwin-multi-user.sh
         . "$EXTRACTED_NIX_PATH/install-darwin-multi-user.sh"
-    elif [ "$(uname -s)" = "Linux" ] && [ -e /run/systemd/system ]; then
-        # shellcheck source=./install-systemd-multi-user.sh
-        . "$EXTRACTED_NIX_PATH/install-systemd-multi-user.sh"
+    elif [ "$(uname -s)" = "Linux" ]; then
+        if [ -e /run/systemd/system ]; then
+            # shellcheck source=./install-systemd-multi-user.sh
+            . "$EXTRACTED_NIX_PATH/install-systemd-multi-user.sh"
+        else
+            failure "Sorry, the multi-user installation requires systemd on Linux (detected using /run/systemd/system)"
+        fi
     else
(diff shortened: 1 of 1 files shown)


scripts/install-multi-user.sh
```diff
@@ -753,9 +753,13 @@ main() {
...
         . "$EXTRACTED_NIX_PATH/install-darwin-multi-user.sh"
-    elif [ "$(uname -s)" = "Linux" ] && [ -e /run/systemd/system ]; then
-        # shellcheck source=./install-systemd-multi-user.sh
-        . "$EXTRACTED_NIX_PATH/install-systemd-multi-user.sh"
+    elif [ "$(uname -s)" = "Linux" ]; then
+        if [ -e /run/systemd/system ]; then
+            # shellcheck source=./install-systemd-multi-user.sh
+            . "$EXTRACTED_NIX_PATH/install-systemd-multi-user.sh"
+        else
+            failure "Sorry, the multi-user installation requires systemd on Linux (detected using /run/systemd/system)"
```
7c6f434c: I would prefer saying «requires systemd on Linux, or manual installation of the Nix daemon» — multi-user Nix with daemon and sandboxing works perfectly without systemd after initial setup, and this message can be misunderstood to imply otherwise.

Actually, the part that needs systemd — starting nix-daemon — is the easy part, and build user creation is the more complicated and annoying part, so it would be nice to have an option to do the entire multi-user install _except_ the systemd part, but that's definitely out of scope.

samueldr: I agree. @grahamc any other opinion to make the best error message? It needs to

 * Explain why it failed (systemd missing) which is almost fine.
 * What alternatives there are.

```
Sorry, the multi-user installation requires systemd on Linux, or manual installation of the Nix daemon.
The lack of systemd was detected by missing the /run/systemd/system directory.
Nix can be configured manually to work in multi-user mode with other init systems. [???]
```

Though, I'm unsure we have the required documentation to refer the user to.

 * * *

Additionally, if we want to start breaking out the other "annoying parts" as you describe, it might be a good starting point, `install.sh --daemon --without-init` could use everything not related to systemd, allowing a bit more agnosticism in the installation, declaring that "yes, continue without init configuration support". (Though, all of this into another feature request.)

PR #2746, opened 2019-03-27
install-multi-user: reduce max-jobs from 32 to 1

bjornfor:
And follow up by removing settings that are set to the same value as Nix' built-in defaults.


bjornfor:
A colleague of mine had to reboot their machine due to the multi-user installer having this high default max-jobs value; nix-build was thrashing the machine. This does not give a good impression of Nix. Please review and merge :-)

bjornfor:
Bump and xref https://github.com/NixOS/nix/pull/2749.

vmandela:
I would recommend this as well. The number of build users (32) is too large. If we dont want to hardcode it, `$(nproc)` might be a reasonable alternative.

PR #2748, opened 2019-03-27
Make nix/unpack-channel.nix a builtin builder

edolstra:
This rewrites the builder of `<nix/unpack-channel.nix>` in C++ and Rust. Previously the builder was a shell script. Since this was the only builder that was still a shell script, this allows us to get rid of the dependency on bash, coreutils, tar, xz, bzip2 etc., reducing Nix's closure size.

This is mostly an experiment in using Rust in Nix. It uses the `tar` crate to do the parsing of tar files.


src/libstore/builtins/unpack-channel.cc
```diff
@@ -0,0 +1,39 @@
...
+    Path out = getAttr("out");
+    auto channelName = getAttr("channelName");
+    auto src = getAttr("src");
+
+    createDirs(out);
+
+    auto source = sinkToSource([&](Sink & sink) {
+        auto decompressor =
+            hasSuffix(src, ".bz2") ? makeDecompressionSink("bzip2", sink) :
+            hasSuffix(src, ".xz") ? makeDecompressionSink("xz", sink) :
```
matthewbauer: No gzip?

configure.ac
```diff
@@ -117,26 +117,15 @@ fi
...
 AC_PATH_PROG(lsof, lsof, lsof)
 
 
-NEED_PROG(cat, cat)
-NEED_PROG(tr, tr)
-AC_ARG_WITH(coreutils-bin, AC_HELP_STRING([--with-coreutils-bin=PATH],
-  [path of cat, mkdir, etc.]),
-  coreutils=$withval, coreutils=$(dirname $cat))
-AC_SUBST(coreutils)
+AC_SUBST(coreutils, [$(dirname $(type -p cat))])
```
matthewbauer: For cross compilation purposes, it is nice to let this be a flag. These should just be runtime deps AFAICT. See https://github.com/NixOS/nixpkgs/pull/58104

src/libstore/builtins/unpack-channel.cc
edolstra: Yeah, we currently lack gzip support in `compression.cc`. Probably not a big deal since our channels don't use it, but it would be nice to add.

configure.ac
edolstra: Well, with this PR, it's now only used during the tests, which I assume cannot be run in a cross-compiled build anyway?

matthewbauer: That's true right now

configure.ac
```diff
@@ -255,8 +255,8 @@ AC_ARG_WITH(sandbox-shell, AC_HELP_STRING([--with-sandbox-shell=PATH],
 AC_SUBST(sandbox_shell)
 
 AC_ARG_ENABLE(shared, AC_HELP_STRING([--enable-shared],
-  [Build shared libraries for Nix [default=yes]]),
-  shared=$enableval, shared=yes)
+  [Build shared libraries for Nix [default=no]]),
```
matthewbauer: Interesting! @edolstra Two things though:

- In Nixpkgs this will be overriden unless you set ```dontDisableStatic```
- You'll run into a nasty bug where lots of the store implementations aren't included in the final binary because no symbol references them. This is fixed in https://github.com/NixOS/nix/pull/2698 though.

nix-rust/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn read(&mut self, buf: &mut [u8]) -> std::result::Result<usize, std::io::Error> {
+        let n = (self.fun)(self.this, buf);
+        assert!(n <= buf.len());
+        Ok(n)
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn unpack_tarfile(source: Source, dest_dir: &str) -> bool {
+    // FIXME: handle errors.
```
Ericson2314: Probably best to use a separate normal abi `Result`-returning function and use idiomatic `?` inside it, and then then deal with FFI interopt just once in the `#[no_mangle] extern "C" fn` wrapper.

lovesegfault:
This is blocking on getting Nix working well on a number of Linux distros. As an employee of a Nix-based company, it'd be really nice if this landed so I can go back to doing work on my machine, which I haven't been able to for a while now.

Janiczek:
@lovesegfault (I'm unaffiliated with Nix, but...) If your company is that dependent on Nix, maybe it could fund some dedicated time from the Nix developers on this issue?

lovesegfault:
@Janiczek We definitely should, I'll see what I can do.

illegalprime:
I think this fixes https://github.com/NixOS/nixpkgs/pull/58104 as I understand it. Thanks for this!

nix-rust/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub ptr: *mut libc::c_void,
+    phantom: std::marker::PhantomData<T>,
+}
+
+impl<T> CBox<T> {
+    fn new(t: T) -> Self {
+        unsafe {
+            let size = std::mem::size_of::<T>();
+            let ptr = libc::malloc(size);
+            *(ptr as *mut T) = t; // FIXME: probably UB
```
puckipedia: This is indeed UB, as you're dereferencing an arbitrary value (which could be illegal for this `T`); this could cause double-frees, arbitrary UAFs, segfaults, etc.

nix-rust/src/error.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct CppException(*const libc::c_void); // == std::exception_ptr*
+
+extern "C" {
+    #[allow(improper_ctypes)] // YOLO
+    fn make_error(s: &str) -> CppException;
```
puckipedia: Fat pointers are, afaik, not FFI-safe at all, and may just do the entire wrong thing.

src/libutil/rust-ffi.hh
```diff
@@ -0,0 +1,84 @@
+#include "serialise.hh"
+
+namespace rust {
+
+// Depending on the internal representation of Rust slices is slightly
+// evil...
```
puckipedia: It is. There's zero guarantee that the internal representation of a slice will look like this.

src/libutil/rust-ffi.hh
```diff
@@ -0,0 +1,84 @@
...
+
+    // FIXME: how to propagate exceptions?
+    static size_t sourceWrapper(void * _this, rust::Slice<uint8_t> data)
+    {
+        auto n = ((nix::Source *) _this)->read(data.ptr, data.size);
+        return n;
+    }
+};
+
+/* C++ representation of Rust's Result<T, CppException>. */
```
puckipedia: Note that there's no guarantee that the enum actually is represented like this; Rust does a lot of niche filling and depending on the type of `T` the tag might just disappear.

nix-rust/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn unpack_tarfile(
+    source: foreign::Source,
+    dest_dir: &str,
+) -> CBox<Result<(), error::CppException>> {
+    CBox::new(tarfile::unpack_tarfile(source, dest_dir).map_err(|err| err.into()))
```
puckipedia: Returning a `CBox<Result<(), error::CppException>>` seems kind of useless, why not just return a pointer to a `CppException` directly? This also saves allocations in the best case, and is actually FFI-safe.

kreisys:
Is this merged to the flakes branch? I can't seem to be able to fetch flakes that aren't already in the store. Getting the suspicious-looking:
```

Ericson2314:
CC @edef1c

PR #2750, opened 2019-03-31
docs: Mention `--max-jobs 0` to build remotely only

nh2:
This answers an often-asked Nix question in the docs.

### DIFF
--- a/doc/manual/command-ref/opt-common.xml
+++ b/doc/manual/command-ref/opt-common.xml
@@ -107,14 +107,22 @@
 <varlistentry xml:id="opt-max-jobs"><term><option>--max-jobs</option> / <option>-j</option>
 <replaceable>number</replaceable></term>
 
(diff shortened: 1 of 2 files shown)


doc/manual/command-ref/opt-common.xml
```diff
@@ -114,7 +114,10 @@
   linkend='conf-max-jobs'><literal>max-jobs</literal></link>
   configuration setting, which itself defaults to
   <literal>1</literal>.  A higher value is useful on SMP systems or to
-  exploit I/O latency.</para></listitem>
+  exploit I/O latency.
+  Setting it to <literal>0</literal> disallows building on the local machine,
```
grahamc: Maybe make a new <para> here?

nh2: Done

PR #2756, opened 2019-04-04
manual: fix the hacking section

kalbasit:
The hacking section of the manual says to run `configurePhase` before running make, but configurePhase does not exist. Running `./configure` without any arguments worked for me.

### DIFF
--- a/doc/manual/hacking.xml
+++ b/doc/manual/hacking.xml
@@ -27,7 +27,7 @@ $ nix-shell
 To build Nix itself in this shell:
 <screen>
 [nix-shell]$ ./bootstrap.sh
-[nix-shell]$ configurePhase
+[nix-shell]$ ./configure
 [nix-shell]$ make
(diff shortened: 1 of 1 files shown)


edolstra:
That's very strange, since `configurePhase` absolutely should exist (it's defined by stdenv). Note that it does a bit more than just call `./configure`. In particular it also passed the right flags to the configure script.

kalbasit:
Interesting! It works if I run `command nix-shell` and I do get a `configurePhase`. I guess this is an issue in [zsh-nix-shell](https://github.com/chisui/zsh-nix-shell), which I use to get a native zsh nix-shell.

edolstra:
Ah yes, shell functions like `configurePhase` only work if you're in the same (bash) shell that sourced the stdenv setup script.

kalbasit:
I see. It makes sense now. Thank you!

PR #2765, opened 2019-04-14
manual: "Nix Package collection" -> "Nixpkgs package collection".

nh2:
Makes difference between Nix and Nixpkgs clearer to avoid some common confusion this sentence on IRC.


mtn:
was confused can confirm :)

PR #2767, opened 2019-04-15
repl: Restore CTRL-C behaviour

### LINKED ISSUE #2743: [regression?] nix repl: ctrl-c behaviour?
It seems that when the nix repl went from using readline -> linenoise -> editline it lost the ctrl-c behaviour of aborting partially typed expressions. Example:

```
Welcome to Nix version 2.3. Type :? for help.

nix-repl> let foo = 1;
          in
          ^C
          ^C
```
 I guess this was not intentional since more or less identical issues have been addressed previously (c.f. #1757, #2076). Currently it looks like the ctrl-c behaviour was last lost in PR #2228.

I have prepared a fix here https://github.com/xbreak/nix/tree/repl/ctrlc which sort of cherry picks from the previous solution when readline was used. I can create a PR if the approach look ok.

globin:
Install signal handler during `readline` to handle SIGINT to abort
partially typed expressions.

Creating PR to make it easier to merge. Fixes #2743, commit by @xbreak.

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -192,6 +192,14 @@ static int listPossibleCallback(char *s, char ***avp) {
   return ac;
 }
 
+namespace {
+    // Used to communicate to NixRepl::getLine whether a signal occurred in ::readline.
+    volatile sig_atomic_t g_signal_received = 0;
+
+    void sigintHandler(int signo) {
+        g_signal_received = signo;
+    }
+}
 
 void NixRepl::mainLoop(const std::vector<std::string> & files)
 {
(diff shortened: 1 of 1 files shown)


dtzWill:
FWIW for the editline upstream has been polishing things recently and in particular right now they're discussing behavior re:signals and API pros/cons-- if there's something that's problematic or could be better now's the time to join in! Or otherwise work with like-minded folks to make a solid experience for your users ^_^.

Linky: https://github.com/troglobit/editline/issues/27#issuecomment-490272968 (and the threads linked from there...)

PR #2779, opened 2019-04-19
build: add exit code for hash and check mismatches

LnL7:
Makes it easier to identify the failure reason in other tooling, eg.
differentiate between a non-deterministic --check vs a failed build.

```
$ nix-build '<nix/fetchurl.nix>' --argstr url http://example.org --argstr sha256 0000000000000000000000000000000000000000000000000000
hash mismatch in fixed-output derivation '/nix/store/nzi9ck45rwlxzcwr25is7qlf3hs5xl83-example.org':
  wanted: sha256:0000000000000000000000000000000000000000000000000000
  got:    sha256:08y4734bm2zahw75b16bcmcg587vvyvh0n11gwiyir70divwp1rm
$ echo $?
102
```
```
$ nix-build -E 'with import <nixpkgs> {}; runCommand "foo" {} "date +%s > $out"' --check
warning: rewriting hashes in '/nix/store/g3k47g0399fvjmbm0p0mnad74k4w8vkz-foo'; cross fingers
error: derivation '/nix/store/mggc8dz13ackb49qca6m23zq4fpq132q-foo.drv' may not be deterministic: output '/nix/store/g3k47g0399fvjmbm0p0mnad74k4w8vkz-foo' differs
$ echo $?
104
```

/cc @grahamc

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -4471,15 +4471,29 @@ void Worker::waitForInput()
 
 unsigned int Worker::exitStatus()
 {
+    /*
+     * 1100100
+     *    ^^^^
+     *    |||`- timeout
+     *    ||`-- output hash mismatch
+     *    |`--- build failure
+     *    `---- not deterministic
+     */
     unsigned int mask = 0;
+    bool buildFailure = permanentFailure || timedOut || hashMismatch;
+    if (buildFailure)
+        mask |= 0x04;  // 100
     if (timedOut)
-        mask |= 1;
+        mask |= 0x01;  // 101
     if (hashMismatch)
-        mask |= 2;
-    if (checkMismatch)
-        mask |= 4;
+        mask |= 0x02;  // 102
+    if (checkMismatch) {
+        mask |= 0x08;  // 104
+    }
 
-    return mask ? 100 + mask : 1;
+    if (mask)
+        mask |= 0x60;
+    return mask ? mask : 1;
 }
 
 
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -266,6 +266,12 @@ public:
     /* Set if at least one derivation had a timeout. */
     bool timedOut;
 
+    /* Set if at least one derivation fails with a hash mismatch. */
+    bool hashMismatch;
+
+    /* Set if at least one derivation is not deterministic in check mode. */
+    bool checkMismatch;
+
     LocalStore & store;
 
     std::unique_ptr<HookInstance> hook;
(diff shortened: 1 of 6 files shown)


src/libstore/build.cc
```diff
@@ -4374,7 +4384,7 @@ void Worker::waitForInput()
 
 unsigned int Worker::exitStatus()
 {
-    return timedOut ? 101 : (permanentFailure ? 100 : 1);
+    return checkMismatch ? 103 : (hashMismatch ? 102 : (timedOut ? 101 : (permanentFailure ? 100 : 1)));
```
grahamc: It might be time to make this a proper if chain :)

LnL7: Moving the status codes to exception types is also an option, not sure if there's a reason why it's stored in flags separately instead.

edolstra: Maybe we should make this an OR of bits (e.g. 4 == check mismatch, 8 == hash mismatch etc.).

LnL7: Updated. What's the best place to document this, the `nix-store --realise` section?

src/libstore/build.cc
```diff
@@ -4415,7 +4425,15 @@ void Worker::waitForInput()
...
-    return timedOut ? 101 : (permanentFailure ? 100 : 1);
+    unsigned int mask = 0;
+    if (timedOut)
+        mask |= 1;
+    if (hashMismatch)
+        mask |= 2;
+    if (checkMismatch)
+        mask |= 4;
+
+    return mask ? 100 + mask : 1;
```
edolstra: Hm, using 100 as a base seems wrong in combination with the ORs, since `100 | 4 == 100`.

Maybe it's better to drop adding 100 and just use masks 2, 4, 8, ... since that's unambiguous with the default exit code of 1.

LnL7: The offset is a bit awkward, but changing the existing exit codes would break some things that depend on it. Like some of the ofborg functionality. I didn't really look into it but perhaps the current value can be encoded in a reasonable way?

src/libstore/build.cc
```diff
@@ -4461,7 +4471,29 @@ void Worker::waitForInput()
...
+     *    ||`-- output hash mismatch
+     *    |`--- build failure
+     *    `---- not deterministic
+     */
+    unsigned int mask = 0;
+    bool buildFailure = permanentFailure || timedOut || hashMismatch;
+    if (buildFailure)
+        mask |= 0x04;  // 100
+    if (timedOut)
+        mask |= 0x01;  // 101
```
edolstra: This produces a value of `0x61 == 97` right?

grahamc: Would be nice to make this tested explicitly.

grahamc: 0x04 + 0x01 + 0x60 (below) produces 101

Also I added a test ensuring this behavior remains true. Should be good to go?

LnL7:
Thanks!

PR #2782, opened 2019-04-24
Track function start and end

grahamc:
Track function start and ends for flame graphs

With this patch, and this file I called `log.py`:

    #!/usr/bin/env nix-shell
    #!nix-shell -i python3 -p python3 --pure

    import sys
    from pprint import pprint

    stack = []
    timestack = []

    for line in open(sys.argv[1]):
        components = line.strip().split(" ", 2)
        if components[0] != "function-trace":
            continue

        direction = components[1]
        components = components[2].rsplit(" ", 2)

        loc = components[0]
        _at = components[1]
        time = int(components[2])

        if direction == "entered":
            stack.append(loc)
            timestack.append(time)
        elif direction == "exited":
            dur = time - timestack.pop()
            vst = ";".join(stack)
            print(f"{vst} {dur}")
            stack.pop()

and:

    nix-instantiate -vvvvvvv ../nixpkgs/pkgs/top-level/release.nix -A unstable > log.matthewbauer 2>&1
    ./log.py ./log.matthewbauer > log.matthewbauer.folded
    flamegraph.pl --title matthewbauer-post-pr log.matthewbauer.folded > log.matthewbauer.folded.svg

I can make flame graphs like: http://gsc.io/log.matthewbauer.folded.svg

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -9,6 +9,7 @@
 #include "json.hh"
 
 #include <algorithm>
+#include <chrono>
 #include <cstring>
 #include <unistd.h>
 #include <sys/time.h>
@@ -16,7 +17,6 @@
 #include <iostream>
 #include <fstream>
 
-#include <sys/time.h>
 #include <sys/resource.h>
 
 #if HAVE_BOEHMGC
@@ -1094,9 +1094,13 @@ void EvalState::callPrimOp(Value & fun, Value & arg, Value & v, const Pos & pos)
     }
 }
 
-
 void EvalState::callFunction(Value & fun, Value & arg, Value & v, const Pos & pos)
 {
+    std::optional<FunctionCallTrace> trace;
+    if (evalSettings.traceFunctionCalls) {
+        trace.emplace(pos);
+    }
+
     forceValue(fun, pos);
 
     if (fun.type == tPrimOp || fun.type == tPrimOpApp) {
--- a/contrib/stack-collapse.py
+++ b/contrib/stack-collapse.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i python3 -p python3 --pure
+
+# To be used with `--trace-function-calls` and `-vvvv` and
+# `flamegraph.pl`.
+#
+# For example:
+#
+# nix-instantiate --trace-function-calls -vvvv '<nixpkgs>' -A hello 2> nix-function-calls.trace
+# ./contrib/stack-collapse.py nix-function-calls.trace > nix-function-calls.folded
+# nix-shell -p flamegraph --run "flamegraph.pl nix-function-calls.folded > nix-function-calls.svg"
+
+import sys
+from pprint import pprint
+import fileinput
+
+stack = []
+timestack = []
+
+for line in fileinput.input():
+    components = line.strip().split(" ", 2)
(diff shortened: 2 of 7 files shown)


edolstra:
Isn't this potentially very expensive since it may do a syscall on the hot path? (I think on Linux the current time is obtained from the vDSO, but on other platforms it might require an actual syscall.)

src/libexpr/eval.cc
```diff
@@ -1094,13 +1095,29 @@ void EvalState::callPrimOp(Value & fun, Value & arg, Value & v, const Pos & pos)
     }
 }
 
+void fnStart(const Pos & pos) {
+    auto now = std::chrono::high_resolution_clock::now();
+    auto duration = now.time_since_epoch();
+    auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(duration);
```
edolstra: This should all be wrapped in `if (verbosity >= lvlDebug) { ... }`.

grahamc: Without this change, it'll be crazy expensive. I was hoping you'd be able to point me to exactly this! :)

grahamc:
I updated it to be conditional on debug logs, changed the log.py to accept a parameter, and replaced `:fire:` with `function-trace`.

src/libexpr/eval.cc
```diff
@@ -1094,13 +1095,32 @@ void EvalState::callPrimOp(Value & fun, Value & arg, Value & v, const Pos & pos)
     }
 }
 
+void fnStart(const Pos & pos) {
```
edolstra: Maybe mark these functions as `static inline` just to be sure.

src/libexpr/eval.cc
```diff
@@ -1094,13 +1095,32 @@ void EvalState::callPrimOp(Value & fun, Value & arg, Value & v, const Pos & pos)
...
+    if (verbosity >= lvlDebug) {
```
edolstra: Hm, actually it might be better to use `lvlVomit` given the amount of output that this is likely to produce...

src/libexpr/eval.cc
grahamc: Done!

src/libexpr/eval.cc
grahamc: Done!

edolstra: Or activate it via a specific setting (e.g. `--trace-function-calls`).

grahamc: I would be happy to do that, but would need some pointers to how to do that.

edolstra: Add a flag to `EvalSettings` like
```
    Setting<bool> traceFunctionCalls{this, false, "trace-function-calls",
        "Bla bla..."};
```
And then do `if (evalSettings.traceFunctionCalls) ...`.

grahamc:
I've updated this PR to use the option like you suggested, and also switched to RAII for the trace. I also added the stack collapse script in to `./contrib`. I haven't checked how it impacts performance when the option is disabled.

grahamc:
I'm testing this PR against some particularly complex Nix expressions which the original implementation did not properly trace.

grahamc:
Yep. It worked great on those test cases. I think this is good to go!

hamishmack:
The option to use is now `-vvvvvvv` not `--trace-function-calls`

PR #2784, opened 2019-04-25
Sort directory entries after they are returned by readdir

c00w:
The readdir syscall does not guarantee a consistent order between calls and on some
file systems this will result in non deterministic store paths for the
same collections of files.

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -5,6 +5,7 @@
 #include "finally.hh"
 #include "serialise.hh"
 
+#include <algorithm>
 #include <cctype>
 #include <cerrno>
 #include <cstdio>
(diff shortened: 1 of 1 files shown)


edolstra:
Do you have a specific case where this caused a problem? For example we already sort directory entries in NAR files to ensure a canonical representation.

c00w:
Hmmm - Where is that code? I didn't spot any ordering when I was tracing the code paths from nix-store --add, but I definitely may have missed something.

A while ago I was using nix to generate flat os disk images with programs embedded and was frustrated by reruns appearing to not be cached correctly. I was eyeballing the nix code for unrelated reasons and realized that this may have been the root cause (I was on a btrfs file system then). I couldn't reproduce changing hashes via nix-store on ext4, but ext4 actually has consistent readdir semantics due to it's index per folder.

c00w:
Tracing the write path though, I definitely see the code running into 
https://sourcegraph.com/github.com/NixOS/nix@f22540464fc2e7bbfd441f275b7581f2d4f14eb5/-/blob/src/libutil/archive.cc#L251

so this PR may be redundant with something else that I'm missing.

c00w:
Ah - the map iteration at 
https://sourcegraph.com/github.com/NixOS/nix@f22540464fc2e7bbfd441f275b7581f2d4f14eb5/-/blob/src/libutil/archive.cc#L87

is guaranteed to be in sorted order per the c++ standard.

Issue #1835, opened 2018-02-02
builtins.hashFile


chris-martin:
If I want a hash of a file, I can compose the `readFile` and `hashString` builtins like so:

```
nix-repl> builtins.hashString "sha256" (builtins.readFile /tmp/a)
"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```

However, this trick only works for text files. Could we add a `hashFile` builtin for hashing any file? Or is there already a way to do this that I'm missing?

deliciouslytyped:
Is there a workaround for this until something shows up in nix?

chris-martin:
My workaround has been to hash the file's nix store path instead, like so:

```
nix-repl> builtins.hashString "sha256" "${/tmp/a}"
"84cadb2ba2c0dc4e1c0242fd0f380a79b736b0d2cb7cecc7bfe3bf66ddcef17c"
```

Since the store path itself includes a hash of the file, this achieves a similar result.

deliciouslytyped:
Yeah, not really what I was going for, but if I just want a file-unique ID I used `baseNameOf hello.outPath`, not sure how bad of an idea that is. It relies on the outpath being in the form of `/a/b/c/d/.../hash-name`.

PR #2792, opened 2019-05-03
Add builtins.hashFile

JohnAZoidberg:
For text files it is possible to do it like so:
`builtins.hashString "sha256" (builtins.readFile /tmp/a)`
but that doesn't work for binary files.
With builtins.hashFile any kind of file can be conveniently hashed.

Fixes: #1835

I added some tests and also tested it manually in the repl:
```
nix-repl> builtins.readFile ./foo
"blah"

nix-repl> builtins.hashString "sha256" "blah"
"8b7df143d91c716ecfa5fc1730022f6b421b05cedee8fd52b1fc65a96030ad52"

nix-repl> builtins.hashFile "sha256" ./foo
"8b7df143d91c716ecfa5fc1730022f6b421b05cedee8fd52b1fc65a96030ad52"
```

Also works on binary files:

```
nix-repl> builtins.hashFile "sha256" ./inst/bin/nix
"bad8124620c1c5538b3a3317fb10cbe7d1bd1b65b692bc3717ae98933022f992"

$ sha256sum ./inst/bin/nix
bad8124620c1c5538b3a3317fb10cbe7d1bd1b65b692bc3717ae98933022f992  ./inst/bin/nix
```

I'm not sure why my test in `tests/lang/eval-fail-hashfile-missing.nix` isn't working. It evaluates but it shouldn't. In the repl it doesn't evaluate.

This is my first edit in the Nix source so please be kind if I forgot something ;)

---

By the way: I found it really weird that `EvalState::coerceToPath` and `EvalState::forceString` take the same arguments but in a different order. Is that intentional or is that something which we should fix?


src/libexpr/primops.cc
```diff
@@ -923,6 +923,20 @@ static void prim_findFile(EvalState & state, const Pos & pos, Value * * args, Va
...
+{
+    string type = state.forceStringNoCtx(*args[0], pos);
+    HashType ht = parseHashType(type);
+    if (ht == htUnknown)
+      throw Error(format("unknown hash type '%1%', at %2%") % type % pos);
+
+    PathSet context; // discarded
+    Path p = state.coerceToPath(pos, *args[1], context);
+
+    mkString(v, hashFile(ht, p).to_string(Base16, false), context);
```
edolstra: This needs a call to `checkSourcePath` to verify that access to `p` is allowed.

tests/lang/eval-fail-hashfile-missing.nix
```diff
@@ -0,0 +1,5 @@
+let
+  paths = [ ./this-file-is-definitely-not-there-7392097 "/and/neither/is/this/37293620" ];
+in
+  builtins.concatLists (map (hash: map (builtins.hashFile hash) paths) ["md5" "sha1" "sha256" "sha512"])
```
edolstra: Having more than one test here is probably pointless since evaluation will fail on the first one, if I'm not mistaken.

src/libexpr/primops.cc
JohnAZoidberg: Thanks!

tests/lang/eval-fail-hashfile-missing.nix
JohnAZoidberg: You're right. But if you run `make installcheck` the test will fail i.e. the evaluation  succeeds. But I think something else is wrong which makes the test fail.

Ekleog:
I have no objection to this change, but just wanted to suggest that we might want to have changes to the nix language (including adding builtins) go through an RFC in the future, so that more eyes could view the changes and think of the overall consistency of the nix language than there are people watching the nix repo :)

JohnAZoidberg:
> I have no objection to this change, but just wanted to suggest that we might want to have changes to the nix language (including adding builtins) go through an RFC in the future, so that more eyes could view the changes and think of the overall consistency of the nix language than there are people watching the nix repo :)

Yes, good idea. Didn't think about it for long because it's not a big change.

@edolstra wait! Sorry, I should've probably called the PR WIP. The test that should fail, succeeds. At least on my system. Can you confirm?

PR #2796, opened 2019-05-09
docs: describe $IN_NIX_SHELL values

vcunat:
I plan to merge this in a few days if noone reacts (assuming I'll still have the permissions :).  Perhaps backport to `2.*-maintenance` branches, too.

### DIFF
--- a/doc/manual/command-ref/env-common.xml
+++ b/doc/manual/command-ref/env-common.xml
@@ -14,7 +14,8 @@
 <varlistentry><term><envar>IN_NIX_SHELL</envar></term>
 
   <listitem><para>Indicator that tells if the current environment was set up by
-  <command>nix-shell</command>.</para></listitem>
+  <command>nix-shell</command>.  Since Nix 2.0 the values are
+  <literal>"pure"</literal> and <literal>"impure"</literal></para></listitem>
 
 </varlistentry>
 


doc/manual/command-ref/env-common.xml
```diff
@@ -14,7 +14,8 @@
 <varlistentry><term><envar>IN_NIX_SHELL</envar></term>
 
   <listitem><para>Indicator that tells if the current environment was set up by
-  <command>nix-shell</command>.</para></listitem>
+  <command>nix-shell</command>.  Since nix 2.0 the values are
```
edolstra: `nix` -> `Nix`.

doc/manual/command-ref/env-common.xml
```diff
@@ -14,7 +14,8 @@
 <varlistentry><term><envar>IN_NIX_SHELL</envar></term>
 
   <listitem><para>Indicator that tells if the current environment was set up by
-  <command>nix-shell</command>.</para></listitem>
+  <command>nix-shell</command>.  Since Nix 2.0 the values are
+  <literal>"pure"</literal> and <literal>"impure"</literal></para></listitem>
```
grahamc: (nit) I think the quotes belong outside of the `<literal>` tags, as the value does not contain the `"`.

vcunat: I originally had it without the quotes, but man pages render these as regular words... and that didn't seem sufficient to me.

matthewbauer:
There is an issue with IN_NIX_SHELL that should be documented. Basically if you are in a nix-shell and call nix-build, it will still think you are in the nix-shell:

```
with import <nixpkgs> {};

if lib.inNixShell then builtins.trace "IN_NIX_SHELL" hello else builtins.trace "not IN_NIX_SHELL" hello
```

This effects haskell a lot.

LnL7:
That function is a misuse of `IN_NIX_SHELL` IMHO. It should probably be replaced by a builtin that actually returns true based on what nix command is invoked or otherwise deprecated.

vcunat:
Regardless of being a misuse (or not), is there any argument against *fixing* that behavior?  (i.e. make nix remove this variable from its env)  Some fix of this should be trivial IMO.

LnL7:
Hmm, indeed. I thought that wouldn't be possible without changing the semantics, but it looks like that would work.

PR #2797, opened 2019-05-10
nix-shell: unset IN_NIX_SHELL for nested nix commands

LnL7:
This fixes the unexpected behaviour of `lib.inNixShell` from nixpkgs. With this it only evaluates to true if the current command is nix-shell.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -264,6 +264,8 @@ static void _main(int argc, char * * argv)
 
     if (runEnv)
         setenv("IN_NIX_SHELL", pure ? "pure" : "impure", 1);
+    else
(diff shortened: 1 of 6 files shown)


edolstra:
Maybe it's simpler to make `builtins.getEnv` return an empty string for `IN_NIX_SHELL`?

LnL7:
That would break the lib function completely, it assumes builtins.getEnv does return something when nix-shell is called.

LnL7:
Is this usage something that should be fixed or do you prefer providing an alternative like a builtin instead (https://github.com/NixOS/nix/pull/2796#issuecomment-491097533)?

LnL7:
I prefer the idea of #3168 over using a magical environment variable.

Issue #2619, opened 2019-01-09
Nix 2.0 deterministic build options don't do anything


arianvp:
From the docs:

```
repeat                                      The number of times to repeat a build in order to verify determinism.
enforce-determinism                         Whether to fail if repeated builds produce different output
run-diff-hook                               Whether to run the program specified by the diff-hook setting repeated builds produce a different result.
```

However, all these options seem to be ignored.  Given the following clearly non-deterministic build:

```nix
{ pkgs ? import <nixpkgs> {}}:
pkgs.runCommand "yop" {} 
''
echo "Spawning 100 processes"
for i in {1..100} ; do ( echo $RANDOM >> $out & ) ; done
''
```

A `nix build` just builds it once, instead of 20 times, whilst checking determinism

```
[arian@ryzen:~/Projects/nondet]$  nix build -f . --repeat 20 --enforce-determinism --diff-hook $(nix-build '<nixpkgs>' -A diffoscope) --run-diff-hook
[1 built]
```

arianvp:
This seems sort of related, but not fully: https://github.com/NixOS/nix/issues/1911

arianvp:
This feature was added in `2.0` so I guess it's a regression that it's broken now?
https://github.com/NixOS/nix/blob/master/doc/manual/release-notes/rl-2.0.xml#L531-L541

the original commits, per release notes:
https://github.com/NixOS/nix/commit/8bdf83f936adae6f2c907a6d2541e80d4120f051
https://github.com/NixOS/nix/commit/9a313469a4bdea2d1e8df24d16289dc2a172a169

Ping @edolstra since you introduced it in 2.0.

arianvp:
Theory:  Maybe there's a bug where not all CLI arguments are correctly passed as nix config?

fzakaria:
@arianvp i don't think you can customize nix config as command line.
somewhat related https://github.com/NixOS/nix/issues/3824

lheckemann:
> Theory: Maybe there's a bug where not all CLI arguments are correctly passed as nix config?

Are you a trusted user?

@fzakaria yes you can generally speaking, the problem is that nix will silently ignore settings under some circumstances, the most common being that the invoking user isn't listed in trusted-users. In some cases owning the store doesn't make you a trusted user (#3247); in others, settings passed on the command line are claimed to be ignored due to not being a trusted user, but still take effect (#3536). It's a bit of a mess.

Artturin:
```bash
[...]

PR #2798, opened 2019-05-10
build: run diff-hook under --check and document diff-hook

grahamc:
for r13y.com

### DIFF
--- a/doc/manual/advanced-topics/diff-hook.xml
+++ b/doc/manual/advanced-topics/diff-hook.xml
@@ -46,16 +46,14 @@ file containing:
 #!/bin/sh
 exec &gt;&amp;2
 echo "For derivation $3:"
-/run/current-system/sw/bin/runuser -u nobody -- /run/current-system/sw/bin/diff -r "$1" "$2"
+/run/current-system/sw/bin/diff -r "$1" "$2"
 </programlisting>
 
-<warning>
-  <para>The diff hook can be run as root. Take care to run as little
-  as possible as root, for this example we use <command>runuser</command>
-  to drop privileges.
-  </para>
-</warning>
 </para>
+
+<para>The diff hook is executed by the same user and group who ran the
+build. However, the diff hook does not have write access to the store
+path just built.</para>
 
 <section>
   <title>
--- a/doc/manual/advanced-topics/diff-hook.xml
+++ b/doc/manual/advanced-topics/diff-hook.xml
@@ -1,0 +1,207 @@
+<chapter xmlns="http://docbook.org/ns/docbook"
+      xmlns:xlink="http://www.w3.org/1999/xlink"
+      xmlns:xi="http://www.w3.org/2001/XInclude"
+      xml:id="chap-diff-hook"
+      version="5.0"
+      >
+
+<title>Verifying Build Reproducibility with <option linkend="conf-diff-hook">diff-hook</option></title>
+
+<subtitle>Check build reproducibility by running builds multiple times
+and comparing their results.</subtitle>
+
+<para>Specify a program with Nix's <xref linkend="conf-diff-hook" /> to
+compare build results when two builds produce different results. Note:
+this hook is only executed if the results are not the same, this hook
+is not used for determining if the results are the same.</para>
+
+<para>For purposes of demonstration, we'll use the following Nix file,
+<filename>deterministic.nix</filename> for testing:</para>
+
+<programlisting>
+let
+  inherit (import &lt;nixpkgs&gt; {}) runCommand;
+in {
+  stable = runCommand "stable" {} ''
+    touch $out
+  '';
+
+  unstable = runCommand "unstable" {} ''
+    echo $RANDOM > $out
+  '';
+}
+</programlisting>
+
+<para>Additionally, <filename>nix.conf</filename> contains:
+
+<programlisting>
+diff-hook = /etc/nix/my-diff-hook
+run-diff-hook = true
+</programlisting>
+
+where <filename>/etc/nix/my-diff-hook</filename> is an executable
+file containing:
+
+<programlisting>
+#!/bin/sh
+exec &gt;&amp;2
+echo "For derivation $3:"
+/run/current-system/sw/bin/runuser -u nobody -- /run/current-system/sw/bin/diff -r "$1" "$2"
+</programlisting>
+
+<warning>
+  <para>The diff hook can be run as root. Take care to run as little
+  as possible as root, for this example we use <command>runuser</command>
+  to drop privileges.
+  </para>
+</warning>
+</para>
+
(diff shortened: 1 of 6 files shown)


doc/manual/advanced-topics/diff-hook.xml
```diff
@@ -0,0 +1,207 @@
...
+exec &gt;&amp;2
+echo "For derivation $3:"
+/run/current-system/sw/bin/runuser -u nobody -- /run/current-system/sw/bin/diff -r "$1" "$2"
+</programlisting>
+
+<warning>
+  <para>The diff hook can be run as root. Take care to run as little
+  as possible as root, for this example we use <command>runuser</command>
+  to drop privileges.
+  </para>
```
edolstra: Hm, rather than document the lack of security, we should just fix? E.g. run the hook under the build user ID?

grahamc: Good idea, I updated the docs and it now runs as the build user / build group.

grahamc:
I want to emphasize that I don't write much C++, so please do carefully review this :P That said, it does work locally! :)

src/libstore/build.cc
```diff
@@ -461,6 +461,28 @@ static void commonChildInit(Pipe & logPipe)
...
+                throw SysError("setgid failed");
+            if (setuid(uid) == -1)
+                throw SysError("setuid failed");
+
+            try {
+                auto diff = runProgram(diffHook, true, {tryA, tryB, drvPath, tmpDir});
+                if (diff != "")
+                    printError(chomp(diff));
+            } catch (Error & error) {
+                printError("diff hook execution failed: %s", error.what());
```
LnL7: Returning the exit code from the diff-hook might be a nice alternative for https://github.com/NixOS/nix/pull/2779

arianvp:
Related: https://github.com/NixOS/nix/issues/2619

I never got `diff-hook` to work so far, as Nix seems to ignore the option for me.  Does this PR fix that?

src/libstore/build.cc
grahamc: mmmm I'd like #2779 better :) as returning this exit code means the diff hooks have to be very smart about their code to avoid being confusing with other Nix exit codes.

grahamc:
@arianvp this PR does change the handling of the options, although it does try to fix the mystery around why your options weren't working:

 - https://github.com/NixOS/nix/pull/2798/files#diff-575bfdde5e7f6c6dfccf0d2be2d872cbR703
 - https://github.com/NixOS/nix/pull/2798/files#diff-575bfdde5e7f6c6dfccf0d2be2d872cbR296

grahamc:
Also, if you follow the "tutorial" bit it should work!

src/libstore/build.cc
```diff
@@ -461,6 +461,28 @@ static void commonChildInit(Pipe & logPipe)
...
 
+void handleDiffHook(bool allowVfork, uid_t uid, uid_t gid, Path tryA, Path tryB, Path drvPath, Path tmpDir)
+{
+    auto diffHook = settings.diffHook;
+    if (diffHook != "" && settings.runDiffHook) {
+        auto wrapper = [&]() {
+            if (setgid(gid) == -1)
+                throw SysError("setgid failed");
+            if (setuid(uid) == -1)
+                throw SysError("setuid failed");
```
edolstra: I think you also need to do `setgroups(0, 0)`.

src/libstore/build.cc
```diff
@@ -461,6 +461,28 @@ static void commonChildInit(Pipe & logPipe)
...
+
+            try {
+                auto diff = runProgram(diffHook, true, {tryA, tryB, drvPath, tmpDir});
```
edolstra: To prevent a double fork, it would be nice to move uid-setting into `runProgram`, i.e. by adding uid/gid options to `RunOptions`.

src/libstore/build.cc
edolstra: And maybe also a `chdir("/")` to make sure we're not in a directory that the diff-hook user shouldn't have access to.

src/libstore/build.cc
grahamc: That seems good, should I pass allowVfork in as a RunOption? How should it default?

doc/manual/command-ref/conf-file.xml
```diff
@@ -240,6 +242,70 @@ false</literal>.</para>
...
+      <listitem>
+        <para>
+          The path to the build's scratch directory. This directory
+          will exist only if the build was run with
+          <option>--keep-failed</option>.
+        </para>
+      </listitem>
+    </orderedlist>
+
+    <para>The diff hook should not print data to stderr or stdout, as
```
arianvp: What exactly do you mean here? I guess it _should_ print to `stderr` right? Otherwise how does the user figure out the output of the diff hook?

grahamc: No. None of the output printed by the diff hook is shown to the user, it all goes to the nix-daemon log.

arianvp: But that doesn't mean it is 'bad' for a diff-hook to output something right?  The sentence now seems to imply that diff hooks shouldn't print results. But we do want them to print results, such that we can actually see what the diff-hook found out, when looking at the nix-daemon logs.

grahamc: Sure, though writing the diff report elsewhere might be a better bet. Do you have a suggestion on exactly how to reword this?

arianvp: > Note that if the diff hook prints data to stderr or stdout, and <command>nix-daemon</command> is enabled,  this is not displayed directly to the user. Instead,
it will be printed in the <command>nix-daemon</command> log

LnL7: I don't remember how this is implemented exactly, but I think it might be pretty easy to forward the output of the hook to the client. Similar to how this is done for build logs.

src/libutil/util.cc
```diff
@@ -1025,6 +1026,16 @@ void runProgram2(const RunOptions & options)
         if (source && dup2(in.readSide.get(), STDIN_FILENO) == -1)
             throw SysError("dupping stdin");
 
+        //if (options.chdir && chdir((*options.chdir).c_str()) == -1)
+        //    throw SysError("chdir failed");
```
edolstra: Was this intended to be commented out?

grahamc: Oops ... Commented that out thinking it was causing #2801. I'll uncomment and re-test. Good catch, thank you.

grahamc:
Uncommented, tested it, and pushed a fixup commit. I'm thinking it'd be best to merge this as a squashed commit -- no need for individual commits.

PR #2802, opened 2019-05-12
build: move needsHashRewrite initialization to startBuilder

### LINKED ISSUE #2801: --check is broken on Linux with Sandboxing enabled after #2775
#2775 broke `--check` on Linux with Sandboxing:

Interestingly this has broken `--check` on Linux with sandboxing:

```
[nix-shell:~/projects/github.com/NixOS/nix]$ nix-build ./test.nix -A stable ; nix-build ./test.nix -A stable --check; nix-build ./test.nix -A unstable; nix-build ./test.nix -A unstable --check
  /nix/store/z98fasz2jqy9gs0xbvdj939p27jwda38-stable.drv
building '/nix/store/z98fasz2jqy9gs0xbvdj939p27jwda38-stable.drv'...
/nix/store/yyxlzw3vqaas7wfp04g0b1xg51f2czgq-stable
checking outputs of '/nix/store/z98fasz2jqy9gs0xbvdj939p27jwda38-stable.drv'...
builder for '/nix/store/z98fasz2jqy9gs0xbvdj939p27jwda38-stable.drv' failed to produce output path '/nix/store/yyxlzw3vqaas7wfp04g0b1xg51f2czgq-stable'
error: build of '/nix/store/z98fasz2jqy9gs0xbvdj939p27jwda38-stable.drv' failed
/nix/store/a61slian4dhzajxviv1gfix29gadif3b-unstable
checking outputs of '/nix/store/dysl2dkqa8fw1f7p5b4dah6nyzr3pk9n-unstable.drv'...
builder for '/nix/store/dysl2dkqa8fw1f7p5b4dah6nyzr3pk9n-unstable.drv' failed to produce output path '/nix/store/a61slian4dhzajxviv1gfix29gadif3b-unstable'
error: build of '/nix/store/dysl2dkqa8fw1f7p5b4dah6nyzr3pk9n-unstable.drv' failed
```

with:

```nix
let
  inherit (import <nixpkgs> {}) runCommand;
in {
  stable = runCommand "stable" {} ''

    touch $out
  '';

  unstable = runCommand "unstable" {} ''
    echo $RANDOM > $out
  '';
}
```

However, with the following change, `--check`'s behavior is restored:

```
[nix-shell:~/projects/github.com/NixOS/nixp]$ git diff
diff --git a/src/libstore/build.cc b/src/libstore/build.cc
index dbadfacd..109e1468 100644
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1837,11 +1837,13 @@ void DerivationGoal::startBuilder()
                     "but this is only allowed when 'sandbox' is 'relaxed'") % drvPath);
 #endif
             useChroot = true;
+            needsHashRewrite = !useChroot;
         }
         else if (settings.sandboxMode == smDisabled)
             useChroot = false;
         else if (settings.sandboxMode == smRelaxed)
             useChroot = !fixedOutput && !noChroot;
+        needsHashRewrite = !useChroot;
     }

     if (worker.store.storeDir != worker.store.realStoreDir) {
@@ -1850,6 +1852,7 @@ void DerivationGoal::startBuilder()
         #else
             throw Error("building using a diverted store is not supported on this platform");
         #endif
+            needsHashRewrite = !useChroot;
     }

     /* If `build-users-group' is not empty, then we have to build as
```

This is probably not the patch we want to commit, but, just as a hint :)

LnL7:
The value of useChroot is not set yet in the constructor, resulting in
hash rewriting being enabled in certain cases where it should not be.

Fixes #2801

/cc @grahamc

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -803,9 +803,6 @@ private:
     /* Whether we're currently doing a chroot build. */
     bool useChroot = false;
 
-    /* Whether we need to perform hash rewriting if there are valid output paths. */
-    bool needsHashRewrite;
-
     Path chrootRootDir;
 
     /* RAII object to delete the chroot directory. */
@@ -884,6 +881,9 @@ public:
     DerivationGoal(const Path & drvPath, const BasicDerivation & drv,
         Worker & worker, BuildMode buildMode = bmNormal);
     ~DerivationGoal();
+
+    /* Whether we need to perform hash rewriting if there are valid output paths. */
+    bool needsHashRewrite();
 
     void timedOut() override;
 
@@ -1034,6 +1034,17 @@ DerivationGoal::~DerivationGoal()
     try { killChild(); } catch (...) { ignoreException(); }
     try { deleteTmpDir(false); } catch (...) { ignoreException(); }
     try { closeLogFile(); } catch (...) { ignoreException(); }
+}
+
+
+inline bool DerivationGoal::needsHashRewrite()
+{
+#if __linux__
+    return !useChroot;
+#else
+    /* Darwin requires hash rewriting even when sandboxing is enabled. */
+    return true;
+#endif
 }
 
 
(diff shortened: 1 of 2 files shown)


grahamc:
I applied the second commit (d75bdb5793e5ebf9e480f5a0012d141347725801) with the newly added test to master and confirmed it fails. I then applied the second commit, and indeed the test fails. Nice work! Nice to see the test, too!

Also, rebasing my diff-hook patches on top allowed them to work, too. Nice. Thank you!

edolstra:
Maybe it's better to turn this into a method, so we don't have to worry about the initialization order? I.e.
```c++
bool needsHashRewrite() {
  #if __linux__
    return !useChroot;
  #else
    return true;
  #endif
}
```

grahamc:
Yay! Thank you both!

PR #2808, opened 2019-05-15
install-multi-user.sh: don't break on Mojave default $HOME

domenkozar:
- [ ] needs testing
- [ ] backport to 2.2-maintenance


domenkozar:
doesn't work

PR #2864, opened 2019-05-16
Fix lock file

### LINKED ISSUE #2815: Give warning if changes shouldn't be made to the lockfile

### LINKED ISSUE #2816: Building when the lockfile exists corrupts the lockfile
For example, running 'nix build' in the tweag/nix/flakes repo:
```
$ cat flake.lock 
{
    "nonFlakeRequires": {},
    "requires": {
        "nixpkgs": {
            "uri": "github:edolstra/nixpkgs/a4d896e89932e873c4117908d558db6210fa3b56"
        }
    },
    "version": 1
}

$ nix build
error: indirect flake reference 'nixpkgs' is not allowed
```
At this point the lockfile has been changed to:
```
{
    "nonFlakeRequires": {},
    "requires": {
        "github:edolstra/nixpkgs/a4d896e89932e873c4117908d558db6210fa3b56": {
            "uri": "github:edolstra/nixpkgs/a4d896e89932e873c4117908d558db6210fa3b56"
        }
    },
    "version": 1
}
```
which is incorrect. Running `nix build` again fixes the lockfile.

### LINKED ISSUE #2817: Fix lockfile usage and updating
`resolveFlake` must be able to handle 6 use cases for what to do with the lockfile:
1. `prim_getFlake` while pure: In this case, everything is handled purely, so no registries, no updating `LockFile`, no writing `LockFile` to file.
2. `nix build dwarffs`: Update `LockFile` before usage
3. `nix build ~/nix/dwarffs`: Update `LockFile` before usage, and save it to file. This is only if the installable is a `FlakeRef` with `IsPath`.
4. `nix build ~/nix/dwarffs --no-save-lock-file`: Update `LockFile` before usage, and don't save it to file
5. `nix build X --recreate-lock-file`: Same as above (depending on what X is), but recreate the `LockFile` from scratch
6. `nix flake update`: Same as case 5

In order to implement this, I will create an enum
```
enum HandleLockFile { AllPure, UpdateLockFile, UseUpdatedLockFile, RecreateLockFile, UseNewLockFile};
```
with two helper functions
```
bool allowedToWrite (HandleLockFile handle);
bool recreateLockFile (HandleLockFile handle);
```
`resolveFlake` now gets an argument `HandleLockFile`.

### LINKED ISSUE #2818: `nix build` without updating lockfile
There should be a flag `--dont-update-lock-file` for things like nix build to just use the given lockfile, for reproducibility and security reasons.

CSVdB:
The lockfile handling in flakes is now fixed:
- Fixes #2815
- Fixes #2817 
- Fixes #2818 
- Fixes #2816


src/libexpr/primops/flake.cc
```diff
@@ -368,73 +364,120 @@ NonFlake getNonFlake(EvalState & state, const FlakeRef & flakeRef, FlakeAlias al
...
+    lockFile = entryToLockFile(dependenciesToFlakeEntry(resFlake));
+
+    if (allowedToWrite(handleLockFile)) {
+        if (auto refData = std::get_if<FlakeRef::IsPath>(&topRef.data)) {
+            writeLockFile(lockFile, refData->path + (topRef.subdir == "" ? "" : "/" + topRef.subdir) + "/flake.lock");
+
+            // Hack: Make sure that flake.lock is visible to Git, so it ends up in the Nix store.
+            runProgram("git", true, { "-C", refData->path, "add",
+                      (topRef.subdir == "" ? "" : topRef.subdir + "/") + "flake.lock" });
+        else std::cout << "Cannot write lockfile because the FlakeRef isn't of the form IsPath." << std::endl;
```
edolstra: This doesn't compile:
```
src/libexpr/primops/flake.cc: In function ‘nix::ResolvedFlake nix::resolveFlake(nix::EvalState&, const nix::FlakeRef&, nix::HandleLockFile)’:
src/libexpr/primops/flake.cc:470:9: error: expected ‘}’ before ‘else’
         else std::cout << "Cannot write lockfile because the FlakeRef isn't of the form IsPath." << std::endl;
         ^~~~
```

src/libexpr/primops/flake.cc
```diff
@@ -368,73 +364,120 @@ NonFlake getNonFlake(EvalState & state, const FlakeRef & flakeRef, FlakeAlias al
...
-        entry.flakeEntries.insert_or_assign(newResFlake.flake.originalRef, dependenciesToFlakeEntry(newResFlake));
-
-    for (auto & nonFlake : resolvedFlake.nonFlakeDeps)
-        entry.nonFlakeEntries.insert_or_assign(nonFlake.alias, LockFile::NonFlakeEntry(nonFlake.resolvedRef, nonFlake.hash));
+    if (handle == AllPure) return false;
+    else if (handle == UpdateLockFile) return true;
+    else if (handle == UseUpdatedLockFile) return false;
+    else if (handle == RecreateLockFile) return true;
+    else if (handle == UseNewLockFile) return false;
+    else throw Error("There is a new HandleLockFile element!");
```
edolstra: Better to do `assert(false)` here since this is an unreachable condition. (Maybe even `static_assert(false)` but I'm not sure the compiler is smart enough to see it's unreachable.)

src/libexpr/primops/flake.hh
```diff
@@ -36,16 +36,16 @@ struct LockFile
...
-    std::map<FlakeId, NonFlakeEntry> nonFlakeEntries;
+    std::map<FlakeAlias, NonFlakeEntry> nonFlakeEntries;
 };
 
 typedef std::vector<std::shared_ptr<FlakeRegistry>> Registries;
 
 Path getUserRegistryPath();
 
-enum RegistryAccess { DisallowRegistry, AllowRegistry, AllowRegistryAtTop };
+enum HandleLockFile { AllPure, UpdateLockFile, UseUpdatedLockFile, RecreateLockFile, UseNewLockFile};
```
edolstra: Could you add some comments explaining the semantics of each mode? It's not really clear from the name what the difference between `UpdateLockFile` and `UseUpdateLockFile` is, or between `RecreateLockFile` and `UseNewLockFile`.

edolstra:
Some confusing behaviour:
```
$ nix build dwarffs:dwarffs
Cannot write lockfile because the FlakeRef isn't of the form IsPath.
```
This gives the impression that it's an error, but the command actually succeeded. Something like `warning: cannot update lockfile of remote flake 'dwarffs:dwarffs'`. (It's shouldn't refer to `IsPath` since that's an internal.)

It's also not clear why it would want to write a lockfile, given that `dwarffs` already contains a lockfile.

PR #2865, opened 2019-05-16
Add default argument to `nix flake info`

### LINKED ISSUE #2822: `nix flake info` doesn't have a default argument
The default argument should be the current directory as a `FlakeRef::IsPath`.

CSVdB:
Fixes #2822


edolstra:
Thanks. However, while I was working on `nix flake test` I've done a refactoring of `FlakeCommand` that also fixes this in a more general way. So I'll close this one.

PR #2877, opened 2019-05-16
FlakeCommand improvements

### LINKED ISSUE #2822: `nix flake info` doesn't have a default argument

edolstra:
It now handles commonality like calling `getFlake()`, resolving relative local flake refs, and creating/updating a lockfile.

Also fixes #2822. All `nix flake` subcommands now work on relative directories and use `.` if no argument is given.

### DIFF
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -70,7 +70,7 @@ struct CmdFlakeList : EvalCommand
     }
 };
 
-void printFlakeInfo(Flake & flake, bool json) {
+void printFlakeInfo(const Flake & flake, bool json) {
     if (json) {
         nlohmann::json j;
         j["id"] = flake.id;
@@ -98,7 +98,7 @@ void printFlakeInfo(Flake & flake, bool json) {
     }
 }
 
-void printNonFlakeInfo(NonFlake & nonFlake, bool json) {
+void printNonFlakeInfo(const NonFlake & nonFlake, bool json) {
     if (json) {
         nlohmann::json j;
         j["id"] = nonFlake.alias;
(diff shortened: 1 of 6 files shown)


tests/flakes.sh
```diff
@@ -104,6 +104,13 @@ EOF
 # Test 'nix flake info'.
 nix flake info --flake-registry $registry flake1 | grep -q 'ID: *flake1'
 
+# Test 'nix flake info' on a local flake. This generates a lockfile as
```
CSVdB: I really don't think `nix flake info` should have this kind of side effect. Can you explain why you do?
Same for every other `nix flake ..` where `.. != update`.

src/nix/flake.cc
```diff
@@ -146,14 +174,11 @@ struct CmdFlakeUpdate : StoreCommand, GitRepoCommand, MixEvalArgs
 
     void run(nix::ref<nix::Store> store) override
     {
-        auto evalState = std::make_shared<EvalState>(searchPath, store);
-
-        if (gitPath == "") gitPath = absPath(".");
-        updateLockFile(*evalState, gitPath);
+        getFlake();
```
CSVdB: 1. We now have a third `getFlake` function, which has different behavior from the other two. This will make it very hard to hack on the Nix codebase.
2. I think `nix flake update` should show an error if it is called on a FlakeRef which isn't of the form `IsPath`.

edolstra: I've restored the old behaviour for `nix flake update`.

tests/flakes.sh
edolstra: It no longer does since `getFlake` doesn't need a lockfile. However, `nix flake deps` does build a lockfile.

PR #2878, opened 2019-05-17
Run builds in a pseudo-terminal

edolstra:
This allows many programs (e.g. gcc, clang, cmake) to print colorized log output (assuming `$TERM` is set to a value like `xterm`).

There are other ways to get colors, in particular setting `CLICOLOR_FORCE`, but they're less widely supported and can break programs that parse tool output.

Example output:

nix-colors

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -38,6 +38,7 @@
 #include <unistd.h>
 #include <errno.h>
 #include <cstring>
+#include <termios.h>
 
 #include <pwd.h>
 #include <grp.h>
@@ -1558,8 +1559,8 @@ void DerivationGoal::buildDone()
     if (hook) {
         hook->builderOut.readSide = -1;
         hook->fromHook.readSide = -1;
-    }
-    else builderOut.readSide = -1;
+    } else
+        builderOut.readSide = -1;
 
     /* Close the log file. */
     closeLogFile();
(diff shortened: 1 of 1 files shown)


dtzWill:
:fireworks: :+1: 

After the `--print-build-logs` option I was really hoping for something like this, being somewhat disappointed my logs weren't already flush with color goodness. :grin:

Apologies for not focusing this comment further, but hopefully the following is useful even so:

There are a few packages (if I dig up a good example I'll report back) that don't neatly handle not having a pty either at all or as stdin.  This has caused strange behavior between sandbox/non-sandboxed builds in the past.  Anyway this is all my way of saying "this is great and might improve some other pain points a bit too" :grin:.

Misc:

* generating prettier diagnostics has significant performance implications on builds... surely colors are only a small part of that.
* capturing colored logs and stripping for less-capable viewers seems like a good approach, is that indeed the plan?
* how confident are we in our escape code handling logic? Guess we'll see!
* can the pty handling be exported to a library or some such? That would be useful for a number of reasons (security, correctness, performance,convenience, portability)

Color codes have a lot of value relative to "delta" in the log stream, but "pretty" build output often does things like using '\r' to "update" a progress bar, which may not be the best use of log storage space.  Although I suppose bz2 probably mitigates the worst of that (and is how logs are stored generally, with brotli in some other places I think).


... I don't suppose while we're doing this we could move to something like bubblewrap (or whatever fits best) for managing the containment, cleanup, and myriad details juggled in the dance that is how we build (in sandbox mode)?  This could be a stepping stone towards something more like systemd-nspawn in approach and generally help define the builder's environment explicitly and easily (independently invokable maybe? for better build debugging? idk).  Nix manages an awesome range of systems-y concerns and responsibilities, but  ... well something to keep in mind, perhaps? :).
------
Ahem, colors would be neat! :)

edolstra:
Yeah, it will be interesting to see how many packages fail to build because of this. The stdenv bootstrap worked fine BTW.

Regarding terminfo, anything in Nixpkgs that uses ncurses should work because it contains terminfo. There might be a better choice than `xterm-256color` though...

nh2:
FYI this change made some Python

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe0' in position 84: ordinal not in range(128)
```

errors go away in a way that I don't quite understand, see:

https://github.com/NixOS/nixpkgs/pull/75676#issuecomment-569284177

nh2:
> errors go away in a way that I don't quite understand

I understand it now, Python 2 detects the TTY to determine the default `stdout` encoding. Written up on https://github.com/NixOS/nixpkgs/pull/75676#issuecomment-579008837.

PR #2880, opened 2019-05-20
doc: clarify that optional attrs in a function argument will be ignored unless specified

Ma27:
In `args@{ a ? 1 }: /* ... */` the value `a` won't be a part of `args`
unless it's specified when calling the function, the default value will
be ignored in this case.

My personal point of view is that this behavior is a matter of taste, at
least I was pretty sure that unmatched arguments will be a part of
`args@` while debugging some Nix code last week.

I decided to add a warning to the docs which hopefully reduces the
confusion of further Nix developers who thought the same about `args@`.

### DIFF
--- a/doc/manual/expressions/language-constructs.xml
+++ b/doc/manual/expressions/language-constructs.xml
@@ -217,7 +217,25 @@ but can also be written as:
   ellipsis(<literal>...</literal>) as you can access attribute names as 
   <literal>a</literal>, using <literal>args.a</literal>, which was given as an
   additional attribute to the function.
-  </para></listitem>
+  </para>
+
+  <warning>
+   <para>
+    The <literal>args@</literal> expression is bound to the argument passed to the function which
+    means that attributes with defaults that aren't explicitly specified in the function call
+    won't cause an evaluation error, but won't exist in <literal>args</literal>.
+   </para>
+   <para>
+    For instance
+<programlisting>
+let
+  function = args@{ a ? 23, ... }: args;
+in
+ function {}
+</programlisting>
+    will evaluate to an empty attribute set.
+   </para>
+  </warning></listitem>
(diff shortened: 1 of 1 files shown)


doc/manual/expressions/language-constructs.xml
```diff
@@ -217,7 +217,25 @@ but can also be written as:
   ellipsis(<literal>...</literal>) as you can access attribute names as 
   <literal>a</literal>, using <literal>args.a</literal>, which was given as an
   additional attribute to the function.
-  </para></listitem>
+  </para>
+
+  <note>
+   <title>Caution with optional arguments</title>
```
edolstra: Please change this to just `<warning>` (no title).

doc/manual/expressions/language-constructs.xml
```diff
@@ -217,7 +217,25 @@ but can also be written as:
...
+   <para>
+    As discussed in <link xlink:href="https://github.com/NixOS/nix/issues/1185">Nix#1185</link>,
```
edolstra: Maybe drop the "As discussed in Nix#1185", since the manual should generally be self-contained and not refer to the issue tracker.

doc/manual/expressions/language-constructs.xml
```diff
@@ -217,7 +217,25 @@ but can also be written as:
...
+    the <literal>args@</literal> expression is bound to the argument passed to the function which means that
+    unspecified defaults won't exist in <literal>args</literal>:
```
edolstra: "unspecified defaults" is a bit vague, since the defaults *are* specified, it's the actual values that aren't.

Ma27:
@edolstra thanks a lot for the review! I've just addressed all comments. Anything else? :)

PR #2882, opened 2019-05-22
tarball-ttl: document

### LINKED ISSUE #1115: Please document "tarball-ttl" option
https://github.com/NixOS/nix/blob/master/src/libstore/download.cc#L522 uses a config option called `tarball-ttl`, but I cannot find any documentation for that option in the man pages. I suppose this should be documented in the `nix.conf` man page?

grahamc:
Closes #1115

### DIFF
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -864,6 +864,31 @@ requiredSystemFeatures = [ "kvm" ];
 
   </varlistentry>
 
+  <varlistentry xml:id="conf-tarball-ttl"><term><literal>tarball-ttl</literal></term>
+
+    <listitem>
+      <para>Default: <literal>3600</literal> seconds.</para>
+
+      <para>The number of seconds a downloaded tarball is considered
+      fresh. If the cached tarball is stale, Nix will check whether
+      it is still up to date using the ETag header. Nix will download
+      a new version if the ETag header is unsupported, or the
+      cached ETag doesn't match.
+      </para>
+
+      <para>Setting the TTL to <literal>0</literal> forces Nix to always
+      check if the tarball is up to date.</para>
+
+      <para>Nix caches tarballs in
+      <filename>$XDG_CACHE_HOME/nix/tarballs</filename>.</para>
+
+      <para>Files fetched via <envar>NIX_PATH</envar>,
+      <function>fetchGit</function>, <function>fetchMercurial</function>,
+      <function>fetchTarball</function>, and <function>fetchurl</function>
+      respect this TTL.
+      </para>
+    </listitem>
+  </varlistentry>
 
   <varlistentry xml:id="conf-timeout"><term><literal>timeout</literal></term>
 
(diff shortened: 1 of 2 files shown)


teto:
would a mention to fetchGit / fetchurl be too detailed ?

grahamc:
No I don't think so, can you help me out by writing a few sentences about that?

doc/manual/command-ref/conf-file.xml
```diff
@@ -864,6 +864,17 @@ requiredSystemFeatures = [ "kvm" ];
 
   </varlistentry>
 
+  <varlistentry xml:id="conf-tarball-ttl"><term><literal>tarball-ttl</literal></term>
+
+    <listitem>
+      <para>The number of seconds tarballs in
+      <filename>~/.cache/nix/tarballs</filename> are considered fresh.
```
teto: $XDG_CACHE_HOME ?

teto:
as I essentially use it to refresh freshly pushed fixup commits, I would advance sthg like:
`Timeout that triggers builtins.fetchGit to refetch data.`

doc/manual/command-ref/conf-file.xml
```diff
@@ -864,6 +864,26 @@ requiredSystemFeatures = [ "kvm" ];
...
+      are considered fresh.
+      </para>
+
+      <para>This is most relevant for fetching tarballs that change
+      remotely via <envar>NIX_PATH</envar> or <function>fetchGit</function>.
+      </para>
+
+      <para>Setting the TTL to <literal>0</literal> forces Nix to check
+      if the tarball is up to date by ETag if the remote server supports
+      it.</para>
```
edolstra: This suggests that `ETag` is only used if TTL=0, but it's always used if supplied by the server.

grahamc: This is complicated :)

doc/manual/command-ref/conf-file.xml
```diff
@@ -864,6 +864,26 @@ requiredSystemFeatures = [ "kvm" ];
...
 
+  <varlistentry xml:id="conf-tarball-ttl"><term><literal>tarball-ttl</literal></term>
+
+    <listitem>
+
+      <para>Default: <literal>3600</literal> seconds.</para>
+
+      <para>The number of seconds a download is cached in
+      <filename>$XDG_CACHE_HOME/nix/tarballs</filename>
+      are considered fresh.
```
edolstra: `is cached in ... are considered fresh` -> `... is considered fresh`. And maybe add something like "If a cached tarball is older, Nix will check whether it is still up to date using its ETag header, and if not, download the latest version."

grahamc:
Updated!

doc/manual/expressions/builtins.xml
```diff
@@ -347,7 +347,7 @@ stdenv.mkDerivation { … }
     You can change the cache timeout either on the command line with
     <option>--option tarball-ttl <replaceable>number of seconds</replaceable></option> or
     in the Nix configuration file with this option:
-    <literal>tarball-ttl <replaceable>number of seconds to cache</replaceable></literal>.
+    <literal><xref linkend="conf-tarball-ttl" /> <replaceable>number of seconds to cache</replaceable></literal>.
```
edolstra: Hm, what does this render as?

grahamc: Screenshot_2019-07-05 Nix Package Manager Guide

It becomes a link, but (at least, locally) doesn't show a special indication that it is clickable. The term is automatically selected as the link text.

teto:
thanks ! happy to see one of the few settings I use documented.

PR #2887, opened 2019-05-23
shared: improve the help message

zimbatm:
For subcommands, the default message was confusing. It directs the user to the top-level help which is not what the user wants.

Ideally the message would be `Try 'nix <sub-command> --help' for more information.` but that would require some more refactoring.

Before:

    $ nix repl --wrong --help
    error: unrecognised flag '--wrong'

After:

    $ nix repl --wrong
    error: unrecognised flag '--wrong'
    Add '--help' to the command for usage information.

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -288,8 +288,8 @@ int handleExceptions(const string & programName, std::function<void()> fun)
         return e.status;
     } catch (UsageError & e) {
         printError(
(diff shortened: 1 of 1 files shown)


edolstra:
The new message is a bit misleading though, since just adding `--help` doesn't work if the parser fails earlier:
```
$ nix repl --wrong --help
error: unrecognised flag '--wrong'
```

PR #2909, opened 2019-05-31
Add 'nix app' command

edolstra:
This is like `nix run`, except that the command to execute is defined in a flake output, e.g.

```nix
  defaultApp = {
    type = "app";
    program = "${packages.blender_2_80}/bin/blender";
  };
```

Thus you can do

```
  $ nix app blender-bin
```

to start Blender from the `blender-bin` flake.

In the future, we can extend this with sandboxing. (For example we would want to be able to specify that Blender should not have network access by default and should only have access to certain paths in the user's home directory.)

### DIFF
--- a/src/libexpr/attr-set.hh
+++ b/src/libexpr/attr-set.hh
@@ -72,6 +72,14 @@ public:
         return {};
     }
 
+    Attr & need(const Symbol & name, const Pos & pos = noPos)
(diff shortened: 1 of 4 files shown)


JohnAZoidberg:
No manual/doc update? Can this be used without flakes?

PR #2911, opened 2019-06-03
Allow tab usage in indented strings

schrislambert:
There are a fair number of derivations in nixpkgs that seem to unintentionally use tabs in their expressions instead of spaces (evilwm, gitit, animbar, ocaml-expat), and right now spaces are the only indentation recognized.  This change allows tabs to be used by having whatever whitespace character appears first at the start of lines as the character used for indentation.  No assumptions need to be made about tab configuration because this does not support mixed tabs and spaces.  This would also allow people who prefer tabs in their local derivations to use that instead.

The only place where mixing is acceptable is with the final ignored line, as some of the packages specified above have spaces indenting their final lines, which seem to be intended to be ignored.

This is my first PR, so feel free to modify or critique it however you want.


edolstra:
I think it's better to be opinionated and discourage/disallow the use of tabs. If there are places in Nixpkgs that use tabs, they should be fixed.

7c6f434c:
>I think it's better to be opinionated and discourage/disallow the use of tabs. If there are places in Nixpkgs that use tabs, they should be fixed.

What's the benefit of discouraging tabs?

edolstra:
To ensure consistency.

schrislambert:
I know a number of languages that recommend using spaces to format, but none that outright require it.

As far as I'm aware, this is the only place where the nix language cares about whitespace.  I personally like to (ab)use elastic tabstops, so having to switch only when writing indented strings in personal derivations is a bit annoying.  It isn't breaking anything, but it does make writing .nix feel a little awkward compared to other languages.  Maybe some sort of opinionated `nix fmt` instead would work to ensure consistency across nixpkgs?  It seems like some maintainers are already using a nix formatter, since that would explain the packages mentioned above using tabs in indented strings but spaces everywhere else (most tellingly in the ignored last line of indented strings).

schrislambert:
>

schrislambert:
Somehow my cursor his the close button when I closed my laptop.  Reopened

7c6f434c:
@edolstra could you please expand a bit? Because in another place, you opposed the simplest available way to ensure syntactic consistency…

edolstra:
@7c6f434c On the contrary, I'm all in favor of consistent use of URLs.

7c6f434c:
Well, at least with spaces and tabs we hopefully won't get a limitation that makes the encouraged option inevitably suboptimal for the most frequent use case.

scintill:
If tabs aren't stripped, I suggest updating https://github.com/NixOS/nix/blob/6924bdf2bf8f50f2e6ec5d490571594450aba13a/doc/manual/expressions/language-values.xml#L113-L115

, since tabs are indentation, but multiline literals in tab-formatted file don't work as I would expect from reading that.

Maybe even make tabs a syntax error, if using them is going to subtly change the meaning of expressions. (I just spent a few minutes figuring out why two identical-looking pieces of code had different outputs...)

wmertens:
Nix is also used in user configurations, that don't need to adhere to nixpkgs guidelines. So for a better user experience, I would kindly request that this gets merged.

For consistency in nixpkgs, the CI tools can check that edited files contain only spaces for indentation.

volth:
Although I am against nix files containing `\t` (and `\r`) because we get different behaviour of visually identically code, which differ only in `\t`-vs-`space` or `\n`-vs-`\r\n`. that leads to difficult to find long lived bugs.

But.
In rare occurencies it is needed to produce file with `\t`, and there is no nice way to do it
```nix
''
  blabla
  ${"\t"}blabla
''
```

It would be better if indented string support `\t` or, more general `\x09` or `\u0009`.

kevincox:
This has been confusing me for years because I was wondering why the docs said that indentation would be stripped however I was still seeing indentation in my output. This should definitely fixed. Enforcing an indentation style on nixpkgs can be done with a simple linter (and be more effective) however there is no reason to prevent tabs for personal nix configs especially when it is a source of confusion.

wmertens:
How about outputting a warning that tabs should not be used in nix definition files, for consistency?

scintill:
still relevant AFAIK

JojOatXGME:
For reference, there is also the issue #3759 about this topic. Note that other line terminators (i.e. CRLF) also seem to “break” indented strings.

frogamic:
Is this discussion still active? I do not understand why only spaces would be stripped when there is a well understood concept of "Whitespace" (see every regex dialect), I don't want nixpkgs strict style guidelines to be forced onto me in my own personal repos where I choose to use tabs for my own reasons, I should be able to do that. Furthermore if it is to be arbitrarily limited to spaces only, the doco should be updated because stating that "whitespace" is removed is misleading.

scintill:
Not stale AFAIK

m911t:
its 2023 now

tests/lang/eval-okay-ind-string.nix
```diff
@@ -41,8 +41,9 @@ let
   '';
 
   s6 = ''  
-    Tabs are not interpreted as whitespace (since we can't guess
-    what tab settings are intended), so don't use them.
+    Tabs are only interpreted as whitespace  when they are
```
toastal, line 44: `whitespace  when` should be `whitespace when`. No need for two spaces.

toastal (changes requested):

kevincox:
If we do add a warning we need to find a way to make it relatively subtle. Probably would need to be dismissable. The last thing I want is some log noise during every Nix evaluation.

JojOatXGME:
> We can however do a warning on tabs

And please also add a warning on CRLF (and CR), since they cause the same problem. While it is probably less common to use CRLF than tabs in the Linux ecosystem, I guess adding this check in addition to the check for tabs shouldn't be that complicated.

FYI, I don't know why CRLF breaks indented string since the CR (\r) is at the end of the line, not at the beginning, but it does.

buckley310:
It's probably a good thing to eliminate "surprising behavior" where possible, rather than warn about it IMO. If RFC 137 succeeds, that sounds like the better route to me.

I don't think this will threaten ecosystem consistency, given that nixpkgs and every nix formatter I have seen uses spaces.

roberth:
Marked as draft because this is either blocked on RFC 137 for supporting breaking language changes, or this could be converted to a change that warn appropriately when such tabs are encountered.

buckley310:
RFC 137 did not pass.

PR #2920, opened 2019-06-04
Lazy flake input fetching

edolstra:
Flake inputs are now fetched lazily. That is, as long as a flake is locked by the top-level lockfile, it won't be fetched unless it's actually needed (i.e. `inputs.<foo>` is evaluated).

Note that this required a change to the lockfile format: we now store the flake ID of inputs, because this is needed *before* the flake is fetched to construct the `inputs` attrset.

Other changes:

* Moved some lockfile-related stuff to `lockfile.{cc,hh}`. There's also some funky multiple inheritance going on now to factor out some common code between flake and non-flake inputs and between top-level and non-top-level flakes.

* `resolveFlake()` now avoids fetching the top-level flake twice.


edolstra:
Just a heads up that I moved flake-related files from `src/libexpr/primops` to `src/libexpr/flake`.

PR #2921, opened 2019-06-05
Handle SIGWINCH in main thread

matthewbauer:
For the SIGWINCH signal to be caught, it needs to be set in sigaction
on the main thread. Previously, this was broken, and updateWindowSize
was never being called. Tested on macOS 10.14.

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -125,8 +125,14 @@ void initNix()
     act.sa_handler = sigHandler;
     if (sigaction(SIGUSR1, &act, 0)) throw SysError("handling SIGUSR1");
 
-    /* Make sure SIGWINCH is handled as well. */
-    if (sigaction(SIGWINCH, &act, 0)) throw SysError("handling SIGWINCH");
+#if __APPLE__
+    /* HACK: on darwin, we need can’t use sigprocmask with SIGWINCH.
+     * Instead, add a dummy sigaction handler, and signalHandlerThread
+     * can handle the rest. */
+    struct sigaction sa;
+    sa.sa_handler = sigHandler;
+    if (sigaction(SIGWINCH, &sa, 0)) throw SysError("handling SIGWINCH");
+#endif
 
     /* Register a SIGSEGV handler to detect stack overflows. */
     detectStackOverflow();
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -125,6 +125,9 @@ void initNix()
     act.sa_handler = sigHandler;
     if (sigaction(SIGUSR1, &act, 0)) throw SysError("handling SIGUSR1");
 
+    /* Make sure SIGWINCH is handled as well. */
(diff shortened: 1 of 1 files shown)


edolstra:
How does this interact with the signal handler thread, which is supposed to handle SIGWINCH? It's also not clear to me how installing a signal handler that does nothing causes `updateWindowSize` to be called.

matthewbauer:
Ah it looks like this only matter in macOS. On Linux the behavior is correct. I'll need to investigate what is actually going on.

matthewbauer:
This example program works on Linux but not macOS:

```
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <signal.h>
#include <string.h>
#include <errno.h>

static sigset_t signal_mask;

void *signal_thread (void *arg) {
  int sig_caught;

  while (1) {
    sigwait(&signal_mask, &sig_caught);
    if (sig_caught == SIGWINCH)
      fprintf(stdout, "SIGWINCH\n");
  }

  return NULL;
}

int main (int argc, char *argv[]) {
  pthread_t sig_thr_id;

  sigemptyset(&signal_mask);
  sigaddset(&signal_mask, SIGWINCH);
  sigprocmask(SIG_BLOCK, &signal_mask, NULL);

  pthread_create(&sig_thr_id, NULL, signal_thread, NULL);

  while(1);

  return 0;
}
```

while this works on macOS:

```
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <signal.h>
#include <string.h>
#include <errno.h>

static sigset_t signal_mask;

void *signal_thread (void *arg) {
  int sig_caught;

  while (1) {
    sigwait(&signal_mask, &sig_caught);
    if (sig_caught == SIGWINCH)
      fprintf(stdout, "SIGWINCH\n");
  }

  return NULL;
}

static void sigHandler(int signo) { }

int main (int argc, char *argv[]) {
  pthread_t sig_thr_id;

  struct sigaction sa;
  sa.sa_handler = sigHandler;
  sigaction(SIGWINCH, &sa, 0);

  sigemptyset(&signal_mask);
  sigaddset(&signal_mask, SIGWINCH);
  sigprocmask(SIG_BLOCK, &signal_mask, NULL);

  pthread_create(&sig_thr_id, NULL, signal_thread, NULL);

  while(1);

  return 0;
}
```

matthewbauer:
I've put an ifdef to just do it on macOS. it looks like it has nothing to do with pthread. This program also doesn't work:

```
#include <stdio.h>
#include <signal.h>

static sigset_t signal_mask;

int main (int argc, char *argv[]) {
  sigemptyset(&signal_mask);
  sigaddset(&signal_mask, SIGWINCH);
  sigprocmask(SIG_BLOCK, &signal_mask, NULL);

  int sig_caught;
  while (1) {
    sigwait(&signal_mask, &sig_caught);
    if (sig_caught == SIGWINCH)
      fprintf(stdout, "SIGWINCH\n");
  }

  return 0;
}
```

edolstra:
Thanks!

PR #2930, opened 2019-06-07
Flake evaluation cache

edolstra:
This exploits the hermetic nature of flake evaluation to speed up repeated evaluations of a flake output attribute.

For example (doing `nix build` on an already present package):

```
$ time nix build nixpkgs:firefox

real    0m1.497s
user    0m1.160s
sys     0m0.139s

$ time nix build nixpkgs:firefox

real    0m0.052s
user    0m0.038s
sys     0m0.007s
```

The cache is `~/.cache/nix/eval-cache-v1.sqlite`, which has entries like

```
INSERT INTO Attributes VALUES(
  X'92a907d4efe933af2a46959b082cdff176aa5bfeb47a98fabd234809a67ab195',
  'packages.firefox',
  1,
  '/nix/store/pbalzf8x19hckr8cwdv62rd6g0lqgc38-firefox-67.0.drv /nix/store/g6q0gx0v6xvdnizp8lrcw7c4gdkzana0-firefox-67.0 out');
```

where the hash `92a9...` is a fingerprint over the flake store path and the contents of the lockfile. Because flakes are evaluated in pure mode, this uniquely identifies the evaluation result.

### DIFF
--- a/src/libstore/sqlite.hh
+++ b/src/libstore/sqlite.hh
@@ -15,12 +15,15 @@ struct SQLite
 {
     sqlite3 * db = 0;
     SQLite() { }
-    SQLite(const Path & path);
+    SQLite(const Path & path, bool create = true);
     SQLite(const SQLite & from) = delete;
     SQLite& operator = (const SQLite & from) = delete;
     SQLite& operator = (SQLite && from) { db = from.db; from.db = 0; return *this; }
     ~SQLite();
     operator sqlite3 * () { return db; }
+
+    /* Disable synchronous mode, set truncate journal mode. */
+    void isCache();
 
     void exec(const std::string & stmt);
 };
@@ -52,6 +55,7 @@ struct SQLiteStmt
 
         /* Bind the next parameter. */
         Use & operator () (const std::string & value, bool notNull = true);
+        Use & operator () (const unsigned char * data, size_t len, bool notNull = true);
         Use & operator () (int64_t value, bool notNull = true);
         Use & bind(); // null
 
--- a/src/libexpr/flake/eval-cache.cc
+++ b/src/libexpr/flake/eval-cache.cc
@@ -1,0 +1,111 @@
+#include "eval-cache.hh"
+#include "sqlite.hh"
+
+#include <set>
+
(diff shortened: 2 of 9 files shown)


src/libstore/sqlite.hh
```diff
@@ -15,13 +15,16 @@ struct SQLite
...
-    SQLite(const Path & path);
+    SQLite(const Path & path, bool create = true);
     SQLite(const SQLite & from) = delete;
     SQLite& operator = (const SQLite & from) = delete;
     SQLite& operator = (SQLite && from) { db = from.db; from.db = 0; return *this; }
     ~SQLite();
     operator sqlite3 * () { return db; }
 
+    /* Disable synchronous mode, set truncate journal mode. */
+    void isCache();
```
andir, line 26: Just came across this in the source code. `isCache` really doesn't speak well for what it is doing. Maybe just calling it `setTruncateJournalMode` would have been better?

edolstra: No, that's an implementation detail. "isCache" is a more declarative statement - i.e. "this database is a cache so do whatever is possible to make it fast at the expensive of reliability".

gilligan: At first glance I would agree that `setTruncateJournalMode` seems to expose an implementation details and is thus maybe not ideal (On the other hand you added a comment saying "Disable synchronous mode, set truncate journal mode" because it would be very hard to derive that otherwise i guess)

From my point of view `isCache` is equally bad: My expectation is that any function adhering to a `isSomething` naming scheme falls into the category of predicate functions: 
```
if (Something is satisfied) return true else return false
```
But `isCache` returns `void` and does some stuff that you cannot guess without reading the comment so i am not sure this is in fact very "_declarative_".

It's not entirely clear to me what the level of abstraction is here, and if `isCache` is part of some API or if it is only ever going to exist for `SQLite`. 

- If this will only ever exist in the context of SQLite then this *is* implementation specific already and then i'm not sure why this shouldn't actually say what it does.
- if this is something that different storage classes will implement then I think a name like `useAsCache` would be more appropriate.

WDYT?

Issue #1761, opened 2017-12-25
`nix copy` silently ignores --no-check-sigs unless user is trusted


bgamari:
Running 
```
$ nix copy --from ssh://ben@192.168.2.102 /nix/store/drc0i99pbfymnmi8zsiqnwhji04z3m84-pandoc-1.19.2.4-armv7l-unknown-linux-gnueabihf 
```
where none of the source derivations have been unsigned fails with "error: cannot add path '/nix/store/402qhax6xjfqgglfcrccz2abkwgbv5pg-armv7l-unknown-linux-gnueabihf-ghc-8.2.1-doc' because it lacks a valid signature".

Looking at `nix copy --help`,  one might be tempted by the `--no-check-sigs` flag to work around the signature check. However, confusingly, the above command fails in the same way even with that flag if run as an untrusted user.

It would be helpful if `nix copy` warned that it was ignoring the `--no-check-sigs` flag if run as an untrusted user.

Ralith:
This is especially annoying when combined with the lack of progress feedback and the fact that the check isn't made until after the full file has been transferred, so you find yourself waiting an unpredictable long time, then see the error, then try the option and wait again, then see the error *again*.

lheckemann:
This applies to many options that are restricted to trusted users — IMO all of these should raise warnings or even errors, when passed on the command line.

vcunat:
Related: to some users it might be confusing that `nix-copy-closure` doesn't need you to pass this flag, i.e. it's as if it were specified.

_I'm not trying to say we should change the requirement to pass it explicitly; I haven't thought about it too much yet._

fzakaria:
I think this is related to [#7812 4894](https://github.com/NixOS/nix/issues/4894)
I got around it honestly by uninstalling multi-user mode and going single-user.

I had added myself as a trusted-user, restarted the daemon but still no bueno.

The log was not helpful either
```
fmzakari@cte-prod-us-west1-2-optical-office-49206:~$ journalctl -f -u nix-daemon
Jul 07 02:40:01 cte-prod-us-west1-2-optical-office-49206.c.googlers.com nix-daemon[31682]: accepted connection from pid 31791, user 780412
Jul 07 02:41:03 cte-prod-us-west1-2-optical-office-49206.c.googlers.com nix-daemon[31682]: accepted connection from pid 31835, user 780412
[...]

PR #2931, opened 2019-06-07
Daemon: warn when an untrusted user cannot override a setting

NinjaTrappeur:
In a daemon-based nix setup, some options cannot be overridden by a
client unless the client's user is considered as trusted.

Currently, if an untrusted user tries to override one of those
options, we are silently ignoring it.

This can be pretty confusing in certain situations.

e.g. a user thinks he disabled the sandbox when in reality he did not.

We are now sending a warning message letting know the user some options
have been ignored.

Related to #1761.

### DIFF
--- a/src/nix-daemon/nix-daemon.cc
+++ b/src/nix-daemon/nix-daemon.cc
@@ -574,7 +574,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
                 else if (setSubstituters(settings.extraSubstituters))
                     ;
                 else
-                    debug("ignoring untrusted setting '%s'", name);
+                    warn("ignoring the user-specified setting '%s', because it is a restricted setting and you are not a trusted user.", name);
             } catch (UsageError & e) {
                 warn(e.what());
             }


NinjaTrappeur:
As a side note, I personally have been hit by this and witnessed several other people being equally confused when confronted to the situation.

src/nix-daemon/nix-daemon.cc
```diff
@@ -573,8 +573,13 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
                     ;
                 else if (setSubstituters(settings.extraSubstituters))
                     ;
-                else
+                else {
+                    logger->log(lvlInfo,
+                        "WARNING: ignoring the " + name + " setting. " +
+                        "You need to declare your unix user as a nix.trustedUser " +
+                        "to override this setting.");
```
grahamc: What would you think about:

> Warning:  «name» is a restricted setting and you are not a trusted user.

A few things related:

lvlInfo is not really a Warning :p 
and `nix.trustedUser` is a NixOS setting, not a nix.conf setting.

NinjaTrappeur: Better indeed :)

> lvlInfo is not really a Warning :p

Totally agree! That said, I did not find anything better:

```
typedef enum {
    lvlError = 0,
    lvlInfo,
    lvlTalkative,
    lvlChatty,
    lvlDebug,
    lvlVomit
} Verbosity;
```

Do you think lvlError would be better suited here?

grahamc:
This is a very trivial diff, but a very cool UX improvement -- merging!

PR #2934, opened 2019-06-10
libstore: consider a resolution error as a NotFound

grahamc:
> My biggest problem with flights: figuring out how to make nix stop trying to fetch from cache when there is no network.  
>
> Would love it to not even try if no network interfaces are up, instead of figuring out what flag to use. 
>
> Maybe hack on that?
-- https://twitter.com/ProgrammerDude/status/1138090058727264256

### DIFF
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -376,6 +376,26 @@ struct CurlDownloader : public Downloader
                 } else {
                     // Don't bother retrying on certain cURL errors either
                     switch (code) {
+                        case CURLE_COULDNT_RESOLVE_HOST:
+                        case CURLE_COULDNT_RESOLVE_PROXY:
+                            // If Nix can't resolve the binary cache,
+                            // the network is probably down and we have
+                            // no reason to retry.
+                            //
+                            // Consider the user on an airplane hacking
+                            // on their favorite project who doesn't need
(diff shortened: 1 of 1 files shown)


roberth:
This may not always be a desirable behavior, for use cases where detecting errors early is desirable, such as in build farms. The expected symptom is a slowdown, the size of which is hard to estimate without real world testing.
Yet, lacking this information, I do expect that this PR is a net improvement.

grahamc:
For build farms: do they substitute? I'm not sure builders in a build farm do substitute.

grahamc:
ie: I think the controlling system does the substitution on their behalf.

cleverca22:
https://github.com/NixOS/hydra/blob/5dd4736aedc1d1356c54bd737db813e25192a885/src/hydra-queue-runner/build-remote.cc#L92-L98
https://github.com/NixOS/nix/blob/7cc1a2593ebece4df28abc86c8b83f5d14dc23a0/src/libstore/legacy-ssh-store.cc#L258-L271

when asking a build slave what it has, you give it a boolean, saying if you will allow it to check the cache

so the `queryValidPaths` function may infact trigger large amounts of network traffic

grahamc:
Nix has a `--no-net` flag in `master`, so closing.

PR #2935, opened 2019-06-10
Document the <nixpkgs> notation for nix-shell.

athas:
I was very confused about the examples in the documentation that use `<nixpkgs>` paths, because that is not documented in the manpage.  After asking on #nixos, I was pointed at [this Nix pill](https://nixos.org/nixos/nix-pills/nix-search-paths.html) that seems to explain what is going on.  I'm still not sure I get all the subtleties, though.

### DIFF
--- a/doc/manual/command-ref/nix-shell.xml
+++ b/doc/manual/command-ref/nix-shell.xml
@@ -69,6 +69,12 @@ unpacked to a temporary location. The tarball must include a single
 top-level directory containing at least a file named
 <filename>default.nix</filename>.</para>
 
+<para>If <replaceable>path</replaceable> is of the form
+<literal>&lt;<replaceable>foo</replaceable>&gt;</literal>, it is
+replaced with the definition of <replaceable>foo</replaceable> as set
+in the environment variable <envar>NIX_PATH</envar> or by the
+<option>-I</option> option.</para>
+
 <para>If the derivation defines the variable
 <varname>shellHook</varname>, it will be evaluated after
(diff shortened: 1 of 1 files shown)


toraritte:
It is documented in the Nix manual, but is indeed missing from the man page.

From [15.1. Values, subsection "Simple Values"](https://nixos.org/nix/manual/#idm140737317975776): "*Paths can also be specified between angle brackets, e.g. `<nixpkgs>`. This means that the directories listed in the environment variable `NIX_PATH` will be searched for the given file or directory name.*"

doc/manual/command-ref/nix-shell.xml
```diff
@@ -69,6 +69,11 @@ unpacked to a temporary location. The tarball must include a single
 top-level directory containing at least a file named
 <filename>default.nix</filename>.</para>
 
+<para>If <replaceable>path</replaceable> is of the form
+<literal>&lt;nixpkgs&gt;</literal>, it is replaced with the definition
+of <literal>nixpkgs</literal> in the environment variable
+<envar>NIX_PATH</envar>.</para>
```
edolstra: This is not a good description because it suggests that `<nixpkgs>` is special, while in fact it can be any path resolved via the Nix search path (e.g. `<foo/bar/xyzzy.nix>`). Also, the search path is not just determined by `NIX_PATH` but also by `-I` flags.

athas:
I have updated the description to mention -I and not refer to `nixpkgs` specifically.  Should the other manpages that take a *path* also be similarly updated?  It's probably less important, because they are not used as prominently or as early in the Nix documentation.

PR #2940, opened 2019-06-11
Make etc/profile.d/nix.sh idempotent

burke:
Repeatedly loading this file currently creates duplicate entries in `PATH`, `NIX_PATH`, and sometimes `MANPATH`. Testing before inserting prevents this.


edolstra:
Arguably this file isn't intended to be idempotent since it should only sourced once (by a login shell).

burke:
(I've rebased on the current master now)

If you don't want to merge this, I can just continue with my workaround, but it strikes me as good practice for scripts like this to act idempotently even when it's not really intended for them to be sourced multiple times (tooling developers in my org end up evaluating it when we switch branches and reload various shell integrations).

One other thing that strikes me, looking at this file, is that it might be beneficial to unconditionally set `MANPATH` but leave a trailing `:`, which `man` interprets as fallthrough to the system default path(s).

PR #2943, opened 2019-06-12
Added flags to override the lockfile

### LINKED ISSUE #2837: Add command line flag to override lockfiles
For issue #13 , flags were implemented to overwrite registries using flags.
Here we want to create flags which can overwrite entries in the lockfile.

Sample use case: I wrote some code, and everything works, but I want to see whether it also works with some old version of nixpkgs.

CSVdB:
Fixes #2837

### DIFF
--- a/src/libexpr/common-eval-args.cc
+++ b/src/libexpr/common-eval-args.cc
@@ -3,6 +3,7 @@
 #include "download.hh"
 #include "util.hh"
 #include "eval.hh"
+#include "flake/flakeref.hh"
 
 namespace nix {
 
@@ -33,15 +34,6 @@ MixEvalArgs::MixEvalArgs()
         .handler([&](std::vector<std::string> ss) {
             evalSettings.pureEval = false;
         });
-
-    mkFlag()
-      .longName("override-flake")
-      .labels({"original-ref", "resolved-ref"})
-      .description("override a flake registry value")
-      .arity(2)
-      .handler([&](std::vector<std::string> ss) {
-          registryOverrides.push_back(std::make_pair(ss[0], ss[1]));
-      });
 }
 
 Bindings * MixEvalArgs::getAutoArgs(EvalState & state)
--- a/src/libexpr/common-eval-args.hh
+++ b/src/libexpr/common-eval-args.hh
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "args.hh"
+#include "flake/flakeref.hh"
 
 namespace nix {
 
@@ -15,8 +16,6 @@ struct MixEvalArgs : virtual Args
     Bindings * getAutoArgs(EvalState & state);
 
     Strings searchPath;
-
-    std::vector<std::pair<std::string, std::string>> registryOverrides;
 
 private:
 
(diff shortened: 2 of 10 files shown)


src/libexpr/common-eval-args.cc
```diff
@@ -40,7 +41,16 @@ MixEvalArgs::MixEvalArgs()
...
+          registryOverrides.insert_or_assign(FlakeRef(ss[0]), FlakeRef(ss[1]));
+      });
+
+    mkFlag()
+      .longName("override-lockfile")
+      .labels({"original-ref", "resolved-ref"})
+      .description("override a lockfile value")
+      .arity(2)
+      .handler([&](std::vector<std::string> ss) {
+          lockFileOverrides.insert_or_assign(FlakeRef(ss[0]), FlakeRef(ss[1]));
```
edolstra: This probably should be added to `EvalCommand` rather than `MixEvalArgs`. The latter is also used by legacy commands (`nix-env` etc) which don't implement flake-related stuff.

CSVdB:
Fixed

tests/config.nix
```diff
@@ -0,0 +1,20 @@
...
+
+  mkDerivation = args:
+    derivation ({
+      inherit system;
+      builder = shell;
+      args = ["-e" args.builder or (builtins.toFile "builder.sh" "if [ -e .attrs.sh ]; then source .attrs.sh; fi; eval \"$buildCommand\"")];
+      PATH = path;
+    } // removeAttrs args ["builder" "meta"])
+    // { meta = args.meta or {}; };
+}
```
edolstra: `tests/config.nix` added accidentally again. Feel free to add it to `.gitignore`.

src/libexpr/common-eval-args.hh
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "args.hh"
+#include "flake/flakeref.hh"
```
edolstra: This `#include` is unnecessary here.

src/libexpr/eval.hh
```diff
@@ -6,6 +6,7 @@
 #include "symbol-table.hh"
 #include "hash.hh"
 #include "config.hh"
+#include "flake/flakeref.hh"
```
edolstra: Same.

src/nix/installables.cc
```diff
@@ -58,9 +58,32 @@ ref<EvalState> EvalCommand::getEvalState()
...
+      .longName("override-flake")
+      .labels({"original-ref", "resolved-ref"})
+      .description("override a flake registry value")
+      .arity(2)
+      .handler([&](std::vector<std::string> ss) {
+          registryOverrides.insert_or_assign(FlakeRef(ss[0]), FlakeRef(ss[1]));
+      });
+
+    mkFlag()
+      .longName("override-lockfile")
```
edolstra: I think `--override-lock` is a better name, since we're overriding a specific lock, not an entire lockfile.

src/libexpr/eval.hh
CSVdB: This import is necessary here, the `FlakeRef` concept is used now because `registryOverrides` are stored in the `EvalState`.

src/libexpr/common-eval-args.hh
CSVdB: I can take the import away here and add it to `nix/command.hh`, where it's needed to define the registry and lockfile overrides in `EvalCommand`.

PR #2974, opened 2019-06-27
checkStoreName: give more precise/verbose error information

### LINKED ISSUE #2973: request to clarify "illegal name" error
Scenario:
While playing in the REPL with the following snippet:
```Nix
mapDir = f: path:
  let
    files = builtins.readDir path;
  in
    builtins.mapAttrs (n: v: f (path + ("/" + n)) v) files;
```
I was confused by the following:
```
l> mapDir (n: v: "${n}") ./.
illegal name: '.git'
```
I didn't understand why it was complaining about illegal names, and besides, I wasn't renaming any of the attributes!

So the specific problem here was that interpolating a path doesn't just interpolate the string, it imports the file into the store, and interpolates the store path. This all makes sense after having it explained to me, but it would have been nice if the error message clarified that this wasn't an evaluation error or somesuch.

I don't have any specific fixes in mind, but @grahamc has suggested something like the following as a starting point:
```
Trying to import «path» in to the Nix store with the name «.foo» is invalid: it must begin with a letter or a number
Note: builtins.path can be used to give it an alternative name, if you cannot rename the file no disk.
```
In my case this would have told me that I need to figure out why its trying to import a path, and I would have eventually figured out the interpolation behavior.

grahamc:
```
$ sudo ./inst/bin/nix-instantiate -E '"${./.git}"'
error: The path name '.git' is invalid: it is illegal to start the
name with a period. Path names are alphanumeric and can include the
symbols +-._?= and must not begin with a period. Note: If you cannot
rename '.git', builtins.path { name = ... } can be used to give it an
alternative name.
```

Close #2973

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -86,18 +86,25 @@ string storePathToHash(const Path & path)
 void checkStoreName(const string & name)
 {
     string validChars = "+-._?=";
+
+    auto baseError = format("The path name '%2%' is invalid: %3%. "
+        "Path names are alphanumeric and can include the symbols %1% "
+        "and must not begin with a period. "
+        "Note: If '%2%' is a source file and you cannot rename it on "
+        "disk, builtins.path { name = ... } can be used to give it an "
+        "alternative name.") % validChars % name;
+
     /* Disallow names starting with a dot for possible security
        reasons (e.g., "." and ".."). */
     if (string(name, 0, 1) == ".")
-        throw Error(format("illegal name: '%1%'") % name);
+        throw Error(baseError % "it is illegal to start the name with a period");
     for (auto & i : name)
         if (!((i >= 'A' && i <= 'Z') ||
               (i >= 'a' && i <= 'z') ||
               (i >= '0' && i <= '9') ||
               validChars.find(i) != string::npos))
         {
-            throw Error(format("invalid character '%1%' in name '%2%'")
-                % i % name);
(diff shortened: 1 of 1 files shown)


src/libstore/store-api.cc
```diff
@@ -86,18 +86,24 @@ string storePathToHash(const Path & path)
 void checkStoreName(const string & name)
 {
     string validChars = "+-._?=";
+
+    auto baseError = format("The path name '%2%' is invalid: %3%. "
+        "Path names are alphanumeric and can include the symbols %1% "
+        "and must not begin with a period. "
+        "Note: If you cannot rename '%2%', builtins.path { name = ... } "
+        "can be used to give it an alternative name.") % validChars % name;
```
edolstra: That means that you *can* rename it :-) Also, the `builtins.path` advice is only applicable for sources.

src/libstore/store-api.cc
```diff
@@ -86,18 +86,25 @@ string storePathToHash(const Path & path)
 void checkStoreName(const string & name)
 {
     string validChars = "+-._?=";
+
+    auto baseError = format("The path name '%2%' is invalid: %3%. "
```
edolstra: Please use `fmt("<format-string>", args...)` in new code.

grahamc: Right now I depend on the final substitution being "curried", is there a way to support that, or should I not do the partial application?

edolstra: Thanks, didn't notice that.

PR #2975, opened 2019-06-27
Don’t use entire /etc/nsswitch.conf file

matthewbauer:
The default nsswitch.conf(5) file in most distros can handle many
different things including host name, user names, groups, etc. In Nix,
we want to limit the amount of impurities that come from these things.
As a result, we should only allow nss to be used for gethostbyname(3)
and getservent(3).

/cc @Ericson2314

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2728,7 +2728,13 @@ void DerivationGoal::runChild()
                on. */
             if (fixedOutput) {
                 ss.push_back("/etc/resolv.conf");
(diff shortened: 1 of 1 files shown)


Ericson2314:
For reference, on NixOS that line today is:
```
hosts:     files mymachines dns myhostname
```

Ericson2314:
We do need to rebuild the daemon to test this, so don't merge before that!

matthewbauer:
mymachines and myhostname should not be necessary.

PR #2977, opened 2019-06-28
Fix `http2 = false` having no effect.

### LINKED ISSUE #2971: Cannot disable http2
`nix 2.2.2` on Manjaro

We observe that `http2 = false` in nix.config has no effect

Using `nix-* -vvvvv` still shows that `HTTP/2` is used.

`nix show-config` shows that nix understands that the setting is set to `false`.

I have the suspicion that the code https://github.com/NixOS/nix/blob/64ec087f582cec33733f4102ab42a3e8f291758a/src/libstore/download.cc#L245-L246 only turns it on, but that it's already on by default in libcurl if the server supports it.

We observe that the `curl` seems to default to HTTP/2 also when downloading .nars from cache.nixos.org. Passing `--http1.1` forces it to not do that.

Maybe the docs at https://curl.haxx.se/libcurl/c/libcurl-env.html are wrong when they say

> You have to set this option if you want to use libcurl's HTTP/2 support.

aniketd:
Fixes #2971.

Setting `http2 = false` in nix config (e.g. /etc/nix/nix.conf)
had no effect, and `nix-env -vvvvv -i hello` still downloaded .nar
packages using HTTP/2.

In `src/libstore/download.cc`, the `CURL_HTTP_VERSION_2TLS` option was
being explicitly set when `downloadSettings.enableHttp2` was `true`,
but, `CURL_HTTP_VERSION_1_1` option was not being explicitly set when
`downloadSettings.enableHttp2` was `false`.

This may be because `https://curl.haxx.se/libcurl/c/libcurl-env.html` states:
"You have to set this option if you want to use libcurl's HTTP/2 support."
but, also, in the changelog, states:
"DEFAULT
Since curl 7.62.0: CURL_HTTP_VERSION_2TLS
Before that: CURL_HTTP_VERSION_1_1"

So, the default setting for `libcurl` is HTTP/2 for version >= 7.62.0.

In this commit, option `CURLOPT_HTTP_VERSION` is explicitly set to
`CURL_HTTP_VERSION_1_1` when `downloadSettings.enableHttp2` nix config
setting is `false`.

This can be tested by running `nix-env -vvvvv -i hello | grep HTTP`

cc with @nh2

### DIFF
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -244,6 +244,8 @@ struct CurlDownloader : public Downloader
             #if LIBCURL_VERSION_NUM >= 0x072f00
             if (downloadSettings.enableHttp2)
                 curl_easy_setopt(req, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2TLS);
+            else
+                curl_easy_setopt(req, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);
             #endif
             curl_easy_setopt(req, CURLOPT_WRITEFUNCTION, DownloadItem::writeCallbackWrapper);
             curl_easy_setopt(req, CURLOPT_WRITEDATA, this);


edolstra:
Thanks!

Issue #967, opened 2016-07-07
Can't build nix on Ubuntu 16.04


popey:
Tried form tip of git and from a 1.11.2 release tarball. Both end the same way:-

$ make
  GEN    Makefile.config
  GEN    doc/manual/version.txt
  GEN    doc/manual/manual.xmli
  GEN    doc/manual/manual.is-valid
I/O error : Attempt to load network entity http://docbook.org/xml/5.0/rng/docbook.rng
warning: failed to load external entity "http://docbook.org/xml/5.0/rng/docbook.rng"
Relax-NG parser error : xmlRelaxNGParse: could not load http://docbook.org/xml/5.0/rng/docbook.rng
Relax-NG schema http://docbook.org/xml/5.0/rng/docbook.rng failed to compile
-:1: parser error : Document is empty
doc/manual/local.mk:29: recipe for target 'doc/manual/manual.is-valid' failed
make: **\* [doc/manual/manual.is-valid] Error 1

Full output:-

alan@gort:~/Development/nix/nix-1.11.2$ ./bootstrap.sh 
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force 
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/bin/autoconf --force
autoreconf: running: /usr/bin/autoheader --force
autoreconf: configure.ac: not using Automake
autoreconf: Leaving directory`.'
alan@gort:~/Development/nix/nix-1.11.2$ ./configure --prefix=/usr/local
checking for a sed that does not truncate output... /bin/sed
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking for the canonical Nix system name... x86_64-linux
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for dirent.h that defines DIR... yes
checking for library containing opendir... none required
checking for struct dirent.d_type... yes
checking for gcc... (cached) gcc
checking whether we are using the GNU C compiler... (cached) yes
checking whether gcc accepts -g... (cached) yes
checking for gcc option to accept ISO C89... (cached) none needed
checking for g++... g++
[...]

Issue #506, opened 2015-03-21
Building nix fails on ubuntu 14.10 because xsltproc is passed --nonet


krafczyk:
When building the target doc/manual/manual.is-valid,

xsltproc is passed the option --nonet. Since the document which is fetched for this target is http://docbook.org/xml/5.0/rng/docbook.rng,

xsltproc fails with the following output:

  GEN    Makefile.config
  GEN    doc/manual/manual.xmli
  GEN    doc/manual/manual.is-valid
I/O error : Attempt to load network entity http://docbook.org/xml/5.0/rng/docbook.rng
warning: failed to load external entity "http://docbook.org/xml/5.0/rng/docbook.rng"
Relax-NG parser error : xmlRelaxNGParse: could not load http://docbook.org/xml/5.0/rng/docbook.rng
Relax-NG schema http://docbook.org/xml/5.0/rng/docbook.rng failed to compile
-:1: parser error : Document is empty

^
-:1: parser error : Start tag expected, '<' not found

^
doc/manual/local.mk:29: recipe for target 'doc/manual/manual.is-valid' failed
make: **\* [doc/manual/manual.is-valid] Error 1

Removing --nonet from the xsltproc lines fixes this error.

edolstra:
Try building from a tarball (http://nixos.org/releases/nix/nix-1.8/nix-1.8.tar.xz). It already contains the generated docs. If you do build from GitHub, you need to ensure that the Docbook schemas / stylesheets are installed in the appropriate locations. The configure script doesn't check for this because it's not a requirement when building from the tarball.

krafczyk:
I have failed to find where to look to ensure that the docbook schemas are installed. Can you point me to the right location?

dkubb:
> The configure script doesn't check for this because it's not a requirement when building from the tarball.

@edolstra I am building from the 1.9 tarball and am seeing the same error. I'm testing the build process on Alpine Linux, which has very few ootb dependencies.

Is it possible that you aren't seeing this error because you're building the tarball on a more capable system (NixOS?) that already has the dependencies preinstalled?

RomanHargrave:
Can confirm. I am encountering this error on Debian 8.1. I was able to resolve it by (indiscriminately) removing `--nonet` from the make scripts.

hasufell:
Building from the git repository should just work without manual fiddling.

Hodapp87:
I just ran into this too on Arch.

lambdafu:
[...]

Issue #257, opened 2014-05-16
Fails to build if xmllint is unavailable


ip1981:
Nix fails to build if xmllint is unavailable, but `configure` does not complain about missing xmllint. It just sets it to `false`:

```
configure.ac:154:AC_PATH_PROG(xmllint, xmllint, false)
```

```
# make V=1
false --nonet --xinclude doc/manual/manual.xml -o doc/manual/manual.xmli.tmp
doc/manual/local.mk:16: recipe for target 'doc/manual/manual.xmli' failed
```

nh2:
PR to fix it: #2979

PR #2979, opened 2019-07-03
autoconf improvements, fixing Ubuntu 16.04 build and probably many more

nh2:
I've put in a lot of effort in the last 2 days to make nix build well from the source repo with a bootstrap build (not depending on nix).

This should greatly improve the situation for Linux distribution packagers, as well as of people who want to quickly contribute to nix (e.g. by improving the manual) without having to have a full nix setup first.

This should fix at least the following long-standing issues, and probably many more:

* #257
* #506
* #967

Highlights:

* Many possible **build failures now result in clear upfront error messages** by `./configure` (e.g. when C++17 isn't supported or `boost` or `editline` libraries are missing or too old).
* The **docbook-without-Internet-and-missing-xsl-files problem is fixed**; it now complains upfront what the user can do about it.
* **`CXXFLAGS`, `LDFLAGS` etc can now be correctly overridden** from outside, as is expected from autoconf projects.
* It's now detected **whether we're in a release tarball**, allowing for clearer autoconf logic on what steps we can skip in that case.

Includes the commit from (and thus subsumes) PR #1201.

I recommend to review the changes commit-by-commit, as I've taken care to write useful commit messages.

Thanks!


edolstra:
Thanks!

edolstra:
I had to revert a few commits because they broke the tarball job (e486d8d40e626a20e06d792db8cc5ac5aba9a5b4), looks like there is something going wrong with detecting whether we're building from Git.

PR #2994, opened 2019-07-17
libstore: ssh-ng: forward the default setOptions

zimbatm:
Without these options it's not possible to forward build timeout options
for example. The TODO is probably applicable to all the remote stores.

### DIFF
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -172,6 +172,12 @@ void RemoteStore::initConnection(Connection & conn)
 }
 
 
+/* TODO Add a way to explicitly ask for some options to be
+        forwarded. One option: A way to query the daemon for its
+        settings, and then a series of params to SSHStore like
+        forward-cores or forward-overridden-cores that only
+        override the requested settings.
+*/
 void RemoteStore::setOptions(Connection & conn)
 {
     conn.to << wopSetOptions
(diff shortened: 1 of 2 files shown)


edolstra:
But this also causes options like the sandbox or remote build configuration to be forwarded to the SSH host. In particular, forwarding the `builders` option would be quite bad because it could cause infinite build forwarding.

zimbatm:
Isn't that argument true for all the remote stores? I think it would be best if the `RemoteStore::setOptions` would filter those options then so we get unified capabilities.

In practice I used `--builders ssh-ng://` and didn't notice the infinite build forwarding issue so maybe the option isn't forwarded already.

LnL7:
See https://github.com/NixOS/nix/pull/1946.

Sending everything makes ssh-ng unusable except if the source and target machines have the exact same `sandbox-paths` since that includes a store path. Especially problematic if the target is another platform.

edolstra:
Note that `RemoteStore` is a misnomer. It was actually originally not remote at all: it's the local store accessed via the daemon. So in that case having the same configuration makes sense.

zimbatm:
Thanks, I will be loading all of the context tomorrow

zimbatm:
Yeah that approach is not going to work.

1) the protocol behind setOptions needs to be re-thought so only a subset of the options can be passed
2) the options need to be classified so only build options are forwarded (eg: the various timeout options, sandbox-enable, ...). Or pass them to the remote builder URI as query string options like eelco suggested.

PR #2995, opened 2019-07-18
Add a post build hook

regnat:
Add a new `post-build-hook` mechanism such that passing `--option post-build-hook /foo/bar` to a nix-* command will cause `/foo/bar /nix/store/drv-that/has-been-built.drv -- out1 out2 ... outn` to be executed after each build.

One specially interesting use-case for this is to upload all the builded artifacts to a binary cache (including the ones that don't appear in the runtime closure of the final derivation or are built because of IFD)

cc @grahamc

### DIFF
--- a/doc/manual/advanced-topics/post-build-hook.xml
+++ b/doc/manual/advanced-topics/post-build-hook.xml
@@ -74,6 +74,8 @@ trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDS
 #!/bin/sh
 
 set -eu
+set -f # disable globbing
+export IFS=' '
 
 echo "Signing paths" $OUT_PATHS
 nix sign-paths --key-file /etc/nix/key.private $OUT_PATHS
@@ -88,8 +90,9 @@ exec nix copy --to 's3://example-nix-cache' $OUT_PATHS
       list of Nix store paths. In this case, we expect and want the
       shell to perform word splitting to make each output path its
       own argument to <command>nix sign-paths</command>. Nix guarantees
-      the paths will only contain characters which are safe for word
-      splitting, and free of any globs.
+      the paths will not contain any spaces, however a store path
+      might contain glob characters. The <command>set -f</command>
+      disables globbing in the shell.
     </para>
   </note>
   <para>
--- a/doc/manual/advanced-topics/post-build-hook.xml
+++ b/doc/manual/advanced-topics/post-build-hook.xml
@@ -1,0 +1,157 @@
+<chapter xmlns="http://docbook.org/ns/docbook"
+      xmlns:xlink="http://www.w3.org/1999/xlink"
+      xmlns:xi="http://www.w3.org/2001/XInclude"
+      xml:id="chap-post-build-hook"
+      version="5.0"
+      >
+
+<title>Using the <xref linkend="conf-post-build-hook" /></title>
+<subtitle>Uploading to an S3-compatible binary cache after each build</subtitle>
+
+
+<section xml:id="chap-post-build-hook-caveats">
+  <title>Implementation Caveats</title>
+  <para>Here we use the post-build hook to upload to a binary cache.
+  This is a simple and working example, but it is not suitable for all
+  use cases.</para>
+
+  <para>The post build hook program runs after each executed build,
+  and blocks the build loop. The build loop exits if the hook program
+  fails.</para>
+
+  <para>Concretely, this implementation will make Nix slow or unusable
+  when the internet is slow or unreliable.</para>
+
+  <para>A more advanced implementation might pass the store paths to a
+  user-supplied daemon or queue for processing the store paths outside
+  of the build loop.</para>
+</section>
+
+<section>
+  <title>Prerequisites</title>
+
+  <para>
+    This tutorial assumes you have configured an S3-compatible binary cache
+    according to the instructions at
+    <xref linkend="ssec-s3-substituter-authenticated-writes" />, and
+    that the <literal>root</literal> user's default AWS profile can
+    upload to the bucket.
+  </para>
+</section>
+
+<section>
+  <title>Set up a Signing Key</title>
+  <para>Use <command>nix-store --generate-binary-cache-key</command> to
+  create our public and private signing keys. We will sign paths
+  with the private key, and distribute the public key for verifying
+  the authenticity of the paths.</para>
+
+  <screen>
+# nix-store --generate-binary-cache-key example-nix-cache-1 /etc/nix/key.private /etc/nix/key.public
+# cat /etc/nix/key.public
+example-nix-cache-1:1/cKDz3QCCOmwcztD2eV6Coggp6rqc9DGjWv7C0G+rM=
+</screen>
+
+<para>Then, add the public key and the cache URL to your
+<filename>nix.conf</filename>'s <xref linkend="conf-trusted-public-keys" />
+and <xref linkend="conf-substituters" /> like:</para>
+
+<programlisting>
+substituters = https://cache.nixos.org/ s3://example-nix-cache
+trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= example-nix-cache-1:1/cKDz3QCCOmwcztD2eV6Coggp6rqc9DGjWv7C0G+rM=
+</programlisting>
+
+<para>we will restart the Nix daemon a later step.</para>
+</section>
+
+<section>
+  <title>Implementing the build hook</title>
+  <para>Write the following script to
+  <filename>/etc/nix/upload-to-cache.sh</filename>:
+  </para>
+
+  <programlisting>
+#!/bin/sh
+
+set -eu
+
+echo "Signing paths" $OUT_PATHS
+nix sign-paths --key-file /etc/nix/key.private $OUT_PATHS
+echo "Uploading paths" $OUT_PATHS
+exec nix copy --to 's3://example-nix-cache' $OUT_PATHS
+</programlisting>
+
+  <note>
+    <title>Should <literal>$OUT_PATHS</literal> be quoted?</title>
+    <para>
+      The <literal>$OUT_PATHS</literal> variable is a space-separated
+      list of Nix store paths. In this case, we expect and want the
+      shell to perform word splitting to make each output path its
+      own argument to <command>nix sign-paths</command>. Nix guarantees
+      the paths will only contain characters which are safe for word
+      splitting, and free of any globs.
+    </para>
+  </note>
+  <para>
+    Then make sure the hook program is executable by the <literal>root</literal> user:
+    <screen>
+# chmod +x /etc/nix/upload-to-cache.sh
+</screen></para>
+</section>
+
+<section>
+  <title>Updating Nix Configuration</title>
+
+  <para>Edit <filename>/etc/nix/nix.conf</filename> to run our hook,
+  by adding the following configuration snippet at the end:</para>
+
+  <programlisting>
+post-build-hook = /etc/nix/upload-to-cache.sh
+</programlisting>
+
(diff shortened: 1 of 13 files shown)


src/libstore/globals.hh
```diff
@@ -315,6 +315,8 @@ public:
         "pre-build-hook",
         "A program to run just before a build to set derivation-specific build settings."};
 
+    Setting<std::string> postBuildHook{this, "", "post-build-hook", "TODO"};
```
edolstra: Fill in TODO?

doc/manual/command-ref/conf-file.xml
```diff
@@ -656,6 +656,72 @@ password <replaceable>my-password</replaceable>
 
   </varlistentry>
 
+  <varlistentry xml:id="conf-post-build-hook">
+    <term><literal>post-build-hook</literal></term>
+    <listitem>
+      <para>Optional. The path to a program which is executed after
+      each build.</para>
+
+      <para>This option can only be set in the nix.conf, or by trusted
```
edolstra: `the nix.conf` -> `<filename>nix.conf</filename>`.

src/libstore/build.cc
```diff
@@ -1629,6 +1629,19 @@ void DerivationGoal::buildDone()
            being valid. */
         registerOutputs();
 
+        if (settings.postBuildHook != "") {
+            printMsg(lvlChatty, format("executing post-build hook '%1%'")
```
grahamc: We should probably bump this up out of chatty, in to something higher so `nix-build` clients typically see it. Otherwise, there is a long pause between the build finishing and the step being considereh completed.

src/libstore/globals.hh
regnat: Woops, fixed

doc/manual/command-ref/conf-file.xml
regnat: Fixed, thanks

src/libstore/build.cc
```diff
@@ -1629,6 +1629,19 @@ void DerivationGoal::buildDone()
...
+                % settings.postBuildHook);
+            auto outputPaths = drv->outputPaths();
+            Strings args;
+            for (auto outputPath: outputPaths)
+                args.push_front(outputPath);
+            args.push_front("--");
+            args.push_front(drvPath);
```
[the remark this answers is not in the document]
regnat: I used cli arguments because that was the simplest to implement (because the `runProgram2` interface doesn't allow to pass extra environment variables, so it requires extending it a bit) and to be more consistent with the `pre-build-hook`. But environment variables would make the hooks simpler indeed. I'll change that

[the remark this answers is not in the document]
regnat: (that being said the hook would then have to parse the `OUTPUT_PATHS` variable to get back a list. But that's probably already better)

regnat:
It looks like the test I've added doesn't work inside a `nix-build`. If I run `make installcheck` locally the test pass, but if I try `nix-build release.nix -A build.x86_64-linux` then the test fail with `error: executing '/build/nix-2.3pre0_0000000/tests/push_to_store.sh': No such file or directory`.

@edolstra @grahamc any idea of what could cause that? The file seems to be present in the sandbox but maybe there's something special about the way nix runs in here which prevents it from seeing the file

regnat:
> It looks like the test I've added doesn't work inside a `nix-build`. If I run `make installcheck` locally the test pass, but if I try `nix-build release.nix -A build.x86_64-linux` then the test fail with `error: executing '/build/nix-2.3pre0_0000000/tests/push_to_store.sh': No such file or directory`.
> 
> @edolstra @grahamc any idea of what could cause that? The file seems to be present in the sandbox but maybe there's something special about the way nix runs in here which prevents it from seeing the file

Nvm, I was using `/usr/bin/env bash` (which is present on my machine but not in the build sandbox) as the shebang for that file, hence the error.

src/libstore/build.cc
```diff
@@ -1629,6 +1629,19 @@ void DerivationGoal::buildDone()
            being valid. */
         registerOutputs();
 
+        if (settings.postBuildHook != "") {
+            printMsg(lvlChatty, format("executing post-build hook '%1%'")
+                % settings.postBuildHook);
+            auto outputPaths = drv->outputPaths();
+            std::map<std::string, std::string> hookEnvironment = getEnv();
+            hookEnvironment.emplace("DRV_PATH", drvPath);
+            hookEnvironment.emplace("OUT_PATHS", chomp(concatStringsSep(":", outputPaths)));
```
grahamc: I am thinking maybe a middle ground would be better: all arguments to the program are output paths, every other parameter passed over the environment. What do you think @edolstra? That way, truly nothing would need to be parsed.

doc/manual/advanced-topics/post-build-hook.xml
```diff
@@ -72,17 +72,12 @@ trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDS
...
-while [ "${1:---}" != "--" ]; do
-  shift
-done
-shift # pop off the "--"
-
-echo "Signing paths" "$@"
-nix sign-paths --key-file /etc/nix/key.secret "$@"
-echo "Uploading paths" "$@"
-exec nix copy --to 's3://example-nix-cache' "$@"
+ALL_OUT_PATHS=$(echo $OUT_PATHS | tr ':' ' ')
```
edolstra: Please use a space (or maybe a newline) as separator in `$OUT_PATHS`. That way you don't need to do any special processing in shell scripts.

src/libstore/build.cc
edolstra: You can use `Activity` for that.

doc/manual/advanced-topics/post-build-hook.xml
regnat: I'd rather implement [@grahamc's proposal](https://github.com/NixOS/nix/pull/2995#discussion_r305346219), is that OK for you?

src/libstore/build.cc
zimbatm: It seems more complicated to explain if the parameters are passed both in the env and the args.

doc/manual/advanced-topics/post-build-hook.xml
zimbatm: In the face of bikeshed, listen to @edolstra

doc/manual/command-ref/conf-file.xml
```diff
@@ -656,6 +656,55 @@ password <replaceable>my-password</replaceable>
...
+        <listitem><para>The hook is not called if a build is substituted.</para></listitem>
+        <listitem><para>If using the Nix daemon, the hook's output goes to the daemon's log.</para></listitem>
+        <listitem><para>If the hook fails, the build succeeds but no further builds execute.</para></listitem>
+      </itemizedlist>
+
+      <warning><para>If the nix-daemon performs the build, the
+      <option>post-build-hook</option> program executes as root.
+      </para></warning>
+
+      <para>The program is executed with no argument, but with the following
```
flokli:
```suggestion
      <para>The program is executed with no arguments, but with the following
```

doc/manual/command-ref/conf-file.xml
```diff
@@ -656,6 +656,55 @@ password <replaceable>my-password</replaceable>
...
+          <term><envar>DRV_PATH</envar></term>
+          <listitem>
+            <para>The derivation which was built.</para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><envar>OUT_PATHS</envar></term>
+          <listitem>
+            <para>Output paths of the built derivation.</para>
```
flokli: This paragraph should include how they're separated.

grahamc:
I have some additional patches here around logging the output from the post build hook. @edolstra can we setup a time to do some pair programming to finish it off?

flokli:
I gave this a try inside a CI environment, to sign and upload derivations after they're built, and it works just beautifully :heart_eyes:!

We should add a note that this will only run for derivations being actually built, not those that are substituted from a binary cache.

In my CI environment, this shouldn't be a problem (as the `nix copy` command makes sure the full closure is uploaded), but it should probably still be documented.

Do we intend to add a similar hook for substituted ones?

doc/manual/command-ref/conf-file.xml
```diff
@@ -656,6 +656,66 @@ password <replaceable>my-password</replaceable>
...
+      <filename>nix.conf</filename>, or on the command line trusted
+      users.</para>
+
+      <para>When using the nix-daemon, the daemon executes the hook as
+      <literal>root</literal>. If the nix-daemon is not involved, the
+      hook runs as the user executing the nix-build.</para>
+
+      <itemizedlist>
+        <listitem><para>The hook executes after an evaluation-time build.</para></listitem>
+        <listitem><para>The hook does not execute on substituted paths.</para></listitem>
```
grahamc: > We should add a note that this will only run for derivations being actually built, not those that are substituted from a binary cache.
- @flokli

grahamc:
I think this is ready to go!

grahamc:
Here is a demo: https://asciinema.org/a/JGe4UHRYMfj554e34gE97h45v

doc/manual/advanced-topics/post-build-hook.xml
```diff
@@ -0,0 +1,157 @@
...
+
+  <note>
+    <title>Should <literal>$OUT_PATHS</literal> be quoted?</title>
+    <para>
+      The <literal>$OUT_PATHS</literal> variable is a space-separated
+      list of Nix store paths. In this case, we expect and want the
+      shell to perform word splitting to make each output path its
+      own argument to <command>nix sign-paths</command>. Nix guarantees
+      the paths will only contain characters which are safe for word
+      splitting, and free of any globs.
```
edolstra: This is not quite true. In particular paths can contain `?` and `+`. So doing `for i in $OUT_PATHS` may cause globbing (and this has caused performance issues in the past, because globbing a path like `...-gtk+-...` requires the entire Nix store to be read.

doc/manual/command-ref/conf-file.xml
```diff
@@ -656,6 +656,61 @@ password <replaceable>my-password</replaceable>
...
+            <para>Example:
+            <literal>/nix/store/5nihn1a7pa8b25l9zafqaqibznlvvp3f-bash-4.4-p23.drv</literal>
+            </para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><envar>OUT_PATHS</envar></term>
+          <listitem>
+            <para>Output paths of the built derivation, separated by a space (<literal> </literal>) character.</para>
```
edolstra: Drop `<literal> </literal>` since it doesn't show anything useful.

doc/manual/command-ref/conf-file.xml
```diff
@@ -656,6 +656,61 @@ password <replaceable>my-password</replaceable>
...
+
+      <para>When using the nix-daemon, the daemon executes the hook as
+      <literal>root</literal>. If the nix-daemon is not involved, the
+      hook runs as the user executing the nix-build.</para>
+
+      <itemizedlist>
+        <listitem><para>The hook executes after an evaluation-time build.</para></listitem>
+        <listitem><para>The hook does not execute on substituted paths.</para></listitem>
+        <listitem><para>The hook's output always goes to the user's terminal.</para></listitem>
+        <listitem><para>If the hook fails, the build succeeds but no further builds execute.</para></listitem>
```
edolstra: It's also worth noting that the hook is executed synchronously and so blocks other builds from progressing.

grahamc: Addressed.

doc/manual/command-ref/conf-file.xml
grahamc: Addressed.

doc/manual/advanced-topics/post-build-hook.xml
grahamc: Addressed.

Issue #971, opened 2016-07-15
Installing Nix in Docker on non-NixOS Linux fails: $USER is not set


rzetterberg:
## Issue description

Installation of Nix 1.11.2 on non-NixOS Linuxes inside Docker fails with:

``` bash
/tmp/nix-binary-tarball-unpack.XXXXajokPC/unpack/nix-1.11.2-x86_64-linux/install: $USER is not set
The command '/bin/sh -c curl https://nixos.org/nix/install | sh' returned a non-zero code: 1
```

According to [this](http://unix.stackexchange.com/a/76356) **login** sets the `$USER` environment variable.

The most obvious workaround is to manually set the `$USER` var in your `Dockerfile` using: `ENV USER tester`.

Using `id -u -n` instead of the environment variable seems like a good idea at first glance. I'm willing to take a crack at it, if it seems reasonable by the maintainers.
### Steps to reproduce
1. Use fresh install
2. Install dependencies bzip2 gnupg perl curl bash sudo
3. Add user to sudo-group and sudoers
4. Run `curl https://nixos.org/nix/install | sh`
## Technical details

I have tested this on `Debian Jessie` and `Alpine 3.3` using `Docker 1.10.3` on `NixOS 16.03`. See files below for a minimal example:

**builld.sh**

``` bash
#!/usr/bin/env sh

set -e

echo "Building Alpine container"
docker build -f ./Dockerfile_alpine -t alpine-nix .

echo "Building Debian container"
docker build -f ./Dockerfile_debian -t debian-nix .
```

**Dockerfile_alpine**

``` docker
FROM alpine:3.3

RUN apk add --update bzip2 gnupg perl curl bash sudo && \
    rm -rf /var/cache/apk/*

ENV LANG en_US.utf8
RUN adduser -h /home/tester -D -s /bin/bash tester
RUN addgroup sudo
RUN addgroup tester sudo
RUN echo " tester      ALL=(ALL:ALL) NOPASSWD: ALL" >> /etc/sudoers

WORKDIR /home/tester
USER tester

RUN curl https://nixos.org/nix/install | sh
```

**Dockerfile_debian**

``` docker
FROM debian:jessie

RUN apt-get update --fix-missing && apt-get install \
    curl perl sudo locales bzip2 -y

RUN apt-get clean && apt-get purge && apt-get autoremove --purge -y && \
    rm -rf /var/lib/apt/lists/*

RUN useradd -m -d /home/tester -s /bin/bash tester
[...]

PR #3005, opened 2019-07-23
Mention that Nix isn't currently working in Docker

kenahoo:
Per issue https://github.com/NixOS/nix/issues/971


edolstra:
Is this still needed after #3007 or are there other Docker issues?

zimbatm:
I have only seen the sandbox issue

Issue #3000, opened 2019-07-19
Nix 2.2 enabled sandboxing on Linux by default, breaking a bunch of installations by default


grahamc:
I think it was the right idea to enable it by default, but probably having a way to detect its support first might be good.

The change broke installations for:
 - centos6
 - centos7
 - debian8
 - debian9
 - gentoo
 - ubuntu12.04

Before:

![2 1 3](https://user-images.githubusercontent.com/76716/61501961-dc27c980-a99f-11e9-8dfa-b369352c7f64.png)

After:

![2 2 2](https://user-images.githubusercontent.com/76716/61501962-dcc06000-a99f-11e9-9a9a-869a166ecc8a.png)

 Open these up in to new tabs and swap between them to see the differences.

Full reports:

 - 2.1.3: https://buildkite.com/organizations/grahamc/pipelines/nix-install-matrix/builds/33/jobs/b20b2d91-a52a-41be-b210-ca8b6a5709eb/artifacts/bdfccc3f-8bd0-4917-b266-96a5a0ce0b23
 - 2.2.2: https://buildkite.com/organizations/grahamc/pipelines/nix-install-matrix/builds/32/jobs/1d2deb87-dd94-427e-b761-4cffa01462b4/artifacts/60d8117e-d878-450f-9af8-685bef65a229

FruitieX:
Possibly related, installation under WSL 2 Ubuntu is also broken:

```
error: while setting up the build environment: mounting /proc: Operation not permitted                                                                                                                                                                                                                                                                                     ./install: unable to install Nix into your default profile
```

Disabling sandboxing seems to help:

```
mkdir ~/.config/nix
echo "sandbox = false" > ~/.config/nix/nix.conf
```

With this workaround in place I was able to complete installation of Nix under WSL 2.

valignatev:
This is still relevant mr bot, thank you

klarkc:
> Ah, I had this problem yesterday on archlinux with very fresh updates, and I went to this issue from the archwiki where it was listed together with the workaound

The workaround there tells to disable sandbox change nix config file, this is a important thing, should that be disabled?

MagicRB:
@klarkc extremely, its akin to disabling sandboxing in docker, a malicious build script could read all your files send them off to a server and you'd never notice

valignatev:
bot begone

taranarmo:
This seems to be still present: I faced it on Nix installation on Arch Linux (well, okay, Manjaro)

[...]

Issue #2662, opened 2019-01-30
Nix failing to install on Debian Stretch Linux 4.10 x86_64


tritoncompute:
Running on Debian Stretch
Linux 4.10 x86_64

Installing Nix in Multi user mode

Fails at
---- sudo execution ------------------------------------------------------------
I am executing:

    $ sudo HOME=/root /nix/store/5hdmx9yk7gr71b98j4vh9271k0zg5jis-nix-2.2.1/bin/nix-env -i /nix/store/5hdmx9yk7gr71b98j4vh9271k0zg5jis-nix-2.2.1

to installing a bootstrapping Nix in to the default Profile

installing 'nix-2.2.1'
error: cloning builder process: Invalid argument
error: unable to start build process

grahamc:
Try putting this in /etc/nix/nix.conf:

```
sandbox = false
```

and try again.

tritoncompute:
Same results, do you know if 'Cgroup namespace' is needed for Nix?

Thank You

tomberek:
Closing as stale. Re-open if needed.

Issue #2636, opened 2019-01-17
Can't install Nix 2.2.1 on clean Ubuntu 18.04 (error: cloning builder process: Operation not permitted)


thomasjm:
I'm trying to install Nix using the latest install script in Docker, via the simple Dockerfile below. This works fine with NIx 2.1.3 but now it doesn't work with Nix 2.2.1. I saw https://github.com/NixOS/nix/issues/1625 but for me the failure is even earlier than in that issue--the install script doesn't even succeed.

```
FROM ubuntu:18.04

RUN apt-get update
RUN apt-get install -y curl

RUN mkdir /nix
RUN chmod a+rwx /nix

RUN adduser user --home /home/user --disabled-password --gecos "" --shell /bin/bash
USER user
ENV USER user
WORKDIR /home/user

RUN curl https://nixos.org/nix/install | sh
```

Here's the full error:

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 22.5M  100 22.5M    0     0  4152k      0  0:00:05  0:00:05 --:--:-- 4854k
performing a single-user installation of Nix...
copying Nix to /nix/store.................................
initialising Nix database...
Nix: creating /home/user/.nix-profile
installing 'nix-2.2.1'
error: cloning builder process: Operation not permitted
error: unable to start build process
/tmp/nix-binary-tarball-unpack.hqawN4uSPr/unpack/nix-2.2.1-x86_64-linux/install: unable to install Nix into your default profile
```

bbarker:
I'm getting the same results on Ubuntu 18.04; last working nix version is 2.1.3 for me as well.

LnL7:
Sandboxing doesn't work in unprivileged containers.

LnL7:
If sandboxing is something you don't care about it can be disabled (like the default on <2.2).

```Dockerfile
FROM ubuntu:18.04

RUN apt-get update \
 && apt-get install -y curl \
 && mkdir -p /nix /etc/nix \
 && chmod a+rwx /nix \
 && echo 'sandbox = false' > /etc/nix/nix.conf

RUN adduser user --home /home/user --disabled-password --gecos "" --shell /bin/bash
CMD /bin/bash -l
USER user
ENV USER user
WORKDIR /home/user

RUN touch .bash_profile \
 && curl https://nixos.org/releases/nix/nix-2.2.1/install | sh
```

[...]

Issue #2632, opened 2019-01-16
Nix 2.2.x fails install on CentOS 7


joelthompson:
Nix fails to install on a fresh CentOS 7 machine. This worked properly in 2.1.3 but fails in 2.2 and 2.2.1. The error message is similar to #1521

Steps to reproduce:
1. Launch ami-4bf3d731 (the [official CentOS 7 AMI](https://wiki.centos.org/Cloud/AWS)) in us-east-1
2. Run `sudo yum install -y bzip2`
3. Run `curl https://nixos.org/nix/install | sh`

The result is this:

```
$ curl https://nixos.org/nix/install | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2476  100  2476    0     0   4012      0 --:--:-- --:--:-- --:--:--  4012
downloading Nix 2.2.1 binary tarball for x86_64-linux from 'https://nixos.org/releases/nix/nix-2.2.1/nix-2.2.1-x86_64-linux.tar.bz2' to '/tmp/nix-binary-tarball-unpack.la1m6G1nYt'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 22.5M  100 22.5M    0     0  2724k      0  0:00:08  0:00:08 --:--:-- 2601k
Note: a multi-user installation is possible. See https://nixos.org/nix/manual/#sect-multi-user-installation
performing a single-user installation of Nix...
copying Nix to /nix/store.................................
initialising Nix database...
installing 'nix-2.2.1'
error: cloning builder process: Invalid argument
error: unable to start build process
/tmp/nix-binary-tarball-unpack.la1m6G1nYt/unpack/nix-2.2.1-x86_64-linux/install: unable to install Nix into your default profile
```

I also get the same issue when trying to install using the multi-user mode (`sh <(curl https://nixos.org/nix/install) --daemon`) on a fresh machine. I also do a `sudo yum update -y` and reboot and retry the install and get the same result.

joelthompson:
Disabling sandboxing by adding `sandbox = false` to /etc/nix/nix.conf seems to fix this issue.

DougBurke:
Is there a workaround or fix for users who can not write to `/etc/nix/nix.conf`? Thanks.

chreekat:
[...]

PR #3006, opened 2019-07-25
Run test to determine if sandboxing is available

matthewbauer:
adds smEnabledIfAvailable option that will run a check where possible
to determine if user namespaces are supported.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2320,7 +2320,7 @@ void DerivationGoal::startBuilder()
                parent. This is only done when sandbox-fallback is set
                to true (the default). */
             if (child == -1 && (errno == EPERM || errno == EINVAL) && settings.sandboxFallback)
-                _exit(EPERM);
+                _exit(1);
             if (child == -1) throw SysError("cloning builder process");
 
             writeFull(builderOut.writeSide.get(), std::to_string(child) + "\n");
@@ -2328,7 +2328,7 @@ void DerivationGoal::startBuilder()
         }, options);
 
         int res = helper.wait();
-        if (res == EPERM && settings.sandboxFallback) {
+        if (res != 0 && settings.sandboxFallback) {
             useChroot = false;
             goto fallback;
         } else if (res != 0)
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2330,6 +2330,7 @@ void DerivationGoal::startBuilder()
         int res = helper.wait();
         if (res != 0 && settings.sandboxFallback) {
             useChroot = false;
+            tmpDirInSandbox = tmpDir;
             goto fallback;
         } else if (res != 0)
             throw Error("unable to start build process");
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2302,10 +2302,20 @@ void DerivationGoal::startBuilder()
                 flags |= CLONE_NEWNET;
 
             pid_t child = clone(childEntry, stack + stackSize, flags, this);
-            if (child == -1 && errno == EINVAL)
+            if (child == -1 && errno == EINVAL) {
                 /* Fallback for Linux < 2.13 where CLONE_NEWPID and
                    CLONE_PARENT are not allowed together. */
-                child = clone(childEntry, stack + stackSize, flags & ~CLONE_NEWPID, this);
+                flags &= ~CLONE_NEWPID;
+                child = clone(childEntry, stack + stackSize, flags, this);
+            }
+            if (child == -1 && (errno == EPERM || errno == EINVAL)) {
+                /* Some distros patch Linux to not allow unpriveleged
+                 * user namespaces. If we get EPERM or EINVAL, try
+                 * without CLONE_NEWUSER and see if that works.
+                 */
+                flags &= ~CLONE_NEWUSER;
+                child = clone(childEntry, stack + stackSize, flags, this);
+            }
             if (child == -1) throw SysError("cloning builder process");
 
             writeFull(builderOut.writeSide.get(), std::to_string(child) + "\n");
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2316,13 +2316,22 @@ void DerivationGoal::startBuilder()
                 flags &= ~CLONE_NEWUSER;
                 child = clone(childEntry, stack + stackSize, flags, this);
             }
+            /* Otherwise exit with EPERM so we can handle this in the
+               parent. This is only done when sandbox-fallback is set
+               to true (the default). */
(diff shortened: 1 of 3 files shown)


src/libstore/build.cc
```diff
@@ -1875,6 +1878,28 @@ void DerivationGoal::startBuilder()
...
+            useChroot = true;
+#if __linux__
+            // Enable sandboxing only if the system supports it. A
+            // clone is tried to determine if it works. Success
+            // indicates sandboxing can be enabled, otherwise disable
+            // sandboxing
+
+            int flags = CLONE_NEWUSER | CLONE_NEWNS | CLONE_NEWIPC | CLONE_NEWUTS | CLONE_PARENT | SIGCHLD;
+            size_t stackSize = 1;
+            char* stack = malloc(stackSize);
```
edolstra: Probably better to use `std::make_unique<unsigned char[]>(stackSize);`. Leaked mallocs pollute valgrind runs.

src/libstore/build.cc
```diff
@@ -1875,6 +1878,28 @@ void DerivationGoal::startBuilder()
...
+            pid_t child = clone(testChildFn, stack + stackSize, flags, NULL);
+
+            if (child > -1)
+                kill(child, SIGKILL);
+
+            if (child == -1) {
+                printError("warning: test cloning failed, sandboxing is disabled");
+                useChroot = false;
+            }
```
edolstra: Could this whole test be done inside an `std::once`? Now we're forking an extra child process for every build.

src/libstore/globals.hh
```diff
@@ -11,7 +11,7 @@
 
 namespace nix {
 
-typedef enum { smEnabled, smRelaxed, smDisabled } SandboxMode;
+typedef enum { smEnabled, smEnabledIfAvailable, smRelaxed, smDisabled } SandboxMode;
```
edolstra: Maybe this should be a separate flag, because you may want to use relaxed mode *and* disable the sandbox if unavailable.

src/libstore/build.cc
```diff
@@ -1875,6 +1878,28 @@ void DerivationGoal::startBuilder()
...
+            int flags = CLONE_NEWUSER | CLONE_NEWNS | CLONE_NEWIPC | CLONE_NEWUTS | CLONE_PARENT | SIGCHLD;
+            size_t stackSize = 1;
+            char* stack = malloc(stackSize);
+            pid_t child = clone(testChildFn, stack + stackSize, flags, NULL);
+
+            if (child > -1)
+                kill(child, SIGKILL);
+
+            if (child == -1) {
+                printError("warning: test cloning failed, sandboxing is disabled");
```
edolstra: There is a `warn` function since recently.

matthewbauer:
> What's the exact use case for this? If it's systems that lack `CLONE_NEWUSER`, it's probably better to disable only `CLONE_NEWUSER`, rather than all sandboxing. So you could check whether `clone` succeeds with and without `CLONE_NEWUSER`.

That sounds like a better idea, if it works right. As a plus that would avoid having an extra "sandbox" setting as sandboxing would still be enabled (just without CLONE_NEWUSER). There are at least 3 patches for this in the wild that I can find:

- Debian & more: https://salsa.debian.org/kernel-team/linux/raw/master/debian/patches/debian/add-sysctl-to-disallow-unprivileged-CLONE_NEWUSER-by-default.patch
- Alt Linux & more: https://lore.kernel.org/patchwork/patch/640897/
- RedHat / CentOS: https://lore.kernel.org/patchwork/patch/700143/

It looks like they all only effect CLONE_NEWUSER.

edolstra:
Looks good but I would be more comfortable with `sandbox-fallback` defaulting to `false`. The special handling for missing user namespaces is more important anyway.

matthewbauer:
> Looks good but I would be more comfortable with `sandbox-fallback` defaulting to `false`. The special handling for missing user namespaces is more important anyway.

My main concern with ```sandbox-fallback = false``` is that we get a lot of users not able to install Nix because the default sandbox doesn't work on their computers. See #2662, #2636, #2632, #2318

grahamc:
I also reported https://github.com/NixOS/nix/issues/3000 which is along the same lines. I'll run this PR through my [nix install matrix](https://buildkite.com/grahamc/nix-install-matrix).

matthewbauer:
This looks like it should be handled, but I'm not sure how:

```
error: while setting up the build environment: mounting /proc: Operation not permitted
./install: unable to install Nix into your default profile
```

grahamc:
Note that #3000 actually isn't related to WSL.

grahamc:
2.2.2's default to enable the sandbox broke installation for:

- centos6
- centos7
- debian8
- debian9
- gentoo
- ubuntu12.04

each of these failed with one of two messages:

```
centos-6-install    installing 'nix-2.2.2'
centos-6-install    error: cloning builder process: Invalid argument
centos-6-install    error: unable to start build process
centos-6-install    /tmp/nix-binary-tarball-unpack.0e2oQ1HL9U/unpack/nix-2.2.2-x86_64-linux/install: unable to install Nix into your default profile
```

or:

```
debian-8-install    installing 'nix-2.2.2'
debian-8-install    error: cloning builder process: Operation not permitted
debian-8-install    error: unable to start build process
debian-8-install    /tmp/nix-binary-tarball-unpack.N6AiK3XbCd/unpack/nix-2.2.2-x86_64-linux/install: unable to install Nix into your default profile
```

After this PR, neither of these errors come up, but all six now fail with another error:

```
centos-7-install    installing 'nix-2.3pre6828_9a0855b'
centos-7-install    warning: Nix search path entry '/home/vagrant/.nix-defexpr/channels' does not exist, ignoring
centos-7-install    error: while setting up the build environment: changing into '/tmp/nix-build-user-environment.drv-0': No such file or directory
centos-7-install    ./nix/install: unable to install Nix into your default profile
```

Here are the install matrix's reports:

- 2.1.3: https://buildkite.com/organizations/grahamc/pipelines/nix-install-matrix/builds/33/jobs/b20b2d91-a52a-41be-b210-ca8b6a5709eb/artifacts/bdfccc3f-8bd0-4917-b266-96a5a0ce0b23
- 2.2.2: https://buildkite.com/organizations/grahamc/pipelines/nix-install-matrix/builds/32/jobs/1d2deb87-dd94-427e-b761-4cffa01462b4/artifacts/60d8117e-d878-450f-9af8-685bef65a229
- this PR at 9a0855b https://buildkite.com/organizations/grahamc/pipelines/nix-install-matrix/builds/44/jobs/4d13373b-73ae-45cb-875e-48170427d659/artifacts/8e25da16-8279-4127-87e3-67c873a10c3c

PR #2291, opened 2018-07-16
nix-channel documentation: don't suggest deprecated function

Taneb:
Running `nix-instantiate --eval -E '(import <nixpkgs> {}).lib.nixpkgsVersion` emits a warning
```
trace: `lib.nixpkgsVersion` is deprecated, use `lib.version` instead!
```

PR #3015, opened 2019-07-30
nix-channel man: lib.nixpkgsVersion -> lib.version

steshaw:
The former is deprecated.


edolstra:
Well, I'm not really convinced that this rename was a good idea in the first place: https://github.com/NixOS/nixpkgs/pull/64693#commitcomment-29951594. Using `lib.version` doesn't work if you have an older version of Nixpkgs.

And also see https://github.com/NixOS/nixpkgs/pull/64693 (not saying yet another rename is a good idea, but `lib.nixpkgsVersion` doesn't have the risk of being used accidentally).

edolstra:
Closing as a duplicate of https://github.com/NixOS/nix/pull/2291.

PR #3036, opened 2019-08-09
nix-shell: provide fake $out

volth:
`nix-shell` is usually run under user account which has no write permissions to nix store.
It results in `installPhase` failure and impossibility to run `fuxupPhase` and `installCheckPhase`

This overrides $out with the path provided as -o command line switch (which defaults to `./result`), so `nix-shell` can "install" packages to a user-controlled directory

---

For example this compiles `htop` to `./result` without touching nix store
```
TMPDIR=$(mktemp -d) \
 nix-shell -E '(import <nixpkgs> {}).htop' --run 'cd $NIX_BUILD_TOP; genericBuild'
```

---

A developer workflow using stock .nix files, without writing special ones for `nix-shell`
1
```bash
mkdir $(pwd)/working-dir
TMPDIR=$(pwd)/working-dir nix-shell -E '(import <nixpkgs> {}).htop' \
  --run 'cd $NIX_BUILD_TOP; unpackPhase; cd htop*; patchPhase && configurePhase'
```

2.
```bash
TMPDIR=$(pwd)/working-dir nix-shell -E '(import <nixpkgs> {}).htop' \
  --run 'cd $NIX_BUILD_TOP/htop*; buildPhase && installPhase && fixupPhase'
```

4. run `result/bin/htop` to see how it works and what to change

5. change sources in `$(pwd)/working-dir` and goto `2`, where only changed sources will be recompiled.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -400,7 +400,9 @@ static void _main(int argc, char * * argv)
                 Path p = (Path) tmpDir + "/" + fn;
                 writeFile(p, var.second);
                 env[var.first + "Path"] = p;
-            } else
+            } else if (var.first == "out")          // 'nix-shell' usualy has no permission to write to nix store, so 'installPhase' fails.
+                env[var.first] = absPath(outLink);  // this overrides derivation's $out with path passed in -o switch (or ./result by default).
+              else                                  // while in 'nix-build' ./result is a symlink to nix store path, in 'nix-shell' it is a real output path
                 env[var.first] = var.second;
 
         restoreAffinity();


volth:
I would add `cd $NIX_BUILD_TOP` to be always executed, so the builder environment will be more like the one of casual `nix-build`, but it would be a breaking change

volth:
I see there is a similar approach described by @danbst in https://stackoverflow.com/questions/31559303/nix-shell-error-mkdir-cannot-create-directory-nix-store-read-only-fil
with `nix-shell --command "export out=/tmp/foo; return"`

But it is too late to do `export out=/tmp/foo` after
https://github.com/NixOS/nix/blob/95b9c6243e2a6c581e1e68928df8319caae07834/src/nix-build/nix-build.cc#L420
where `$out` is propagated (at least) to `$prefix` and `$NIX_LDFLAGS`.

volth:
multi-outputs derivations are to be handled too...

edolstra:
I don't think `./result` is a good default, because it's likely to clash with symlinks created by `nix-build`.

src/nix-build/nix-build.cc
```diff
@@ -400,7 +400,12 @@ static void _main(int argc, char * * argv)
                 Path p = (Path) tmpDir + "/" + fn;
                 writeFile(p, var.second);
                 env[var.first + "Path"] = p;
-            } else
+            } else if (drv.outputs.find(var.first) != drv.outputs.end()) {  // 'nix-shell' usualy has no permission to write to nix store, so 'installPhase' fails.
```
edolstra: This can be written more concisely as `if (drv.outputs.count(var.first))`.

volth:
> I don't think `./result` is a good default, because it's likely to clash with symlinks created by `nix-build`.

Yes, I used it only with explicit `-o`.
The default might be the old behavior, with `$out` pointing to nix store location (It might have sense for `root` user)

edolstra:
No, we definitely don't want people manually installing into the Nix store, even as root.

acowley:
This was exceptionally helpful for me recently in addressing a problematic install phase with a very slow build. I hope it can be merged.

volth:
In addition to @edolstra's concerns, it does not play nice with `pkgsi686Linux`
while
`nix-shell -E '(import <nixpkgs> {}).htop'` does work
`nix-shell -E '(import <nixpkgs> {}).pkgsi686Linux.htop'` will likely fail
Probably, unpatched `nix-shell` has the problem too, I just do not have unpatched `nix-shell` around (and the aarch64 community builder obviously has to `pkgsi686Linux`)

PR #3037, opened 2019-08-09
nix-daemon: allow more safe settings by "untrusted" users.

dtzWill:
These options are mostly selected based on my usage
and judging each to only change "superficial" properties
of the build or output printed.

I particularly find it useful to ask builds to only
use remote builders (when on laptop, esp battery),
hence the `--max-jobs 0` whitelisting.

TTL settings can be useful when debugging either
a build cache (mostly forcing re-fetching)
or fetching latest copy of a URL such as
https://github.com/NixOS/nixpkgs/archive/master.tar.gz
(most commonly I used this to ensure my NixOps deployments
use latest version of branch they're pointed at, for
faster/sane change/deploy cycles).

I think setting `show-trace` is safe,
assuming the eval is performed by the user
and so any increased memory usage is theirs.

I don't think we are concerned about tracing
exposing (or being disabled to "hide") information
from untrusted users, but LMK if that's not right.

In the future it might be nice to add alias options
`--remote-only` or `--local-only` or something,
but for now don't reject the current method for
requesting the behavior.

### DIFF
--- a/src/nix-daemon/nix-daemon.cc
+++ b/src/nix-daemon/nix-daemon.cc
@@ -567,7 +567,17 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
                 else if (trusted
                     || name == settings.buildTimeout.name
                     || name == "connect-timeout"
-                    || (name == "builders" && value == ""))
+                    || (name == "builders" && value == "") // "build locally, no remote builders"
+                    || (name == "max-jobs" && value == "0") // "build remotely, no local builders"
+                    || name == "show-trace" // eval-time tracing doesn't require trust
+                    || name == "log-lines" // log output size doesn't require trust
+                    || name == "print-missing" // behavioral, no trust required
+                    // TTL-related settings are useful, shouldn't require trust and are useful if you just updated cache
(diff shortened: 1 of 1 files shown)


edolstra:
This is largely redundant due to 7298a38a07d62d2077ca30a3ff619e32edb4eb1e.

dtzWill:
Okay, will close this then, thanks! :)

arcnmx:
Is `show-trace` meant to be omitted from the list in 7298a38? 2.3 generates spurious warnings with that flag despite functioning as expected (it shows the trace when evaluating as the user).

zimbatm:
`--show-trace` is a client-side option. If it's not, it should be filtered out as well. The daemon will not make use of it.

dtzWill:
In case it's useful, here's the same change after moving nix-daemon code into libstore:

https://github.com/dtzWill/nix/commit/50162849e1041d05830cee17b0038e2ff1303eff

Regretted reverting this change, both because of squelched warnings and because I use `tarball-ttl 0` in places to ensure I'm building what I think i am :).  Anyway, enjoy if useful.

Issue #2673, opened 2019-02-09
'nix-channel --update' stopped working on nix-2.2.1: error: while setting up the build environment: getting attributes of path '': No such file or directory


trofi:
The install is a `nix` install on `Gentoo` (not a NixOS).

```
$ nix-channel --update --verbose
downloading 'https://nixos.org/channels/nixpkgs-unstable'...
downloading 'https://releases.nixos.org/nixpkgs/nixpkgs-19.03pre168678.54dcb0b1376/binary-cache-url'...
unpacking channels...
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
error: while setting up the build environment: getting attributes of path '': No such file or directory
error: program '/usr/bin/nix-env' failed with exit code 1
```

I tried `strace`-ing `nix-env` and running `nix-daemon` but I did not find where that `''` path comes from. Can `nix` be improved to dump more information on where that path comes from?

Some version info:
```
$ nix --version
nix (Nix) 2.2.1
$ nix-channel --list
nixpkgs https://nixos.org/channels/nixpkgs-unstable
```

ottidmes on #nixos linked me to a thread about /bin/sh out of nix: https://logs.nix.samueldr.com/nixos/2019-02-06#1938291

Working workaround:
```
$ nix-build '<nixpkgs>' -A bash
$ sudo systemctl restart nix-daemon
```

trofi:
Found out today it was not completely working workaround. I got to the failure:

I built nix from source and **did not pass** `--with-sandbox-shell=...` option to `./configure`. Unfortunately instead of bailing out nix built the following sandbox path:

```
$ nix show-config | fgrep sandbox
...
sandbox-paths = /bin/sh=
```

I worked around it locally by passing `--with-sandbox-shell=/bin/busybox` to `./configure`.

To make problem harder to encounter I suggest to:
1. fail `./configure` if `--with-sandbox-shell=` is not passed or default to `/bin/sh`
2. mention 'source' part fo `sandbox-path` when mounting onto target fails

Thanks!

dtzWill:
> fail `./configure` if `--with-sandbox-shell=` is not passed or default to `/bin/sh`

This certainly seems in the territory of `nix doctor` if build-time check isn't preferred for whatever reason.
Truth is that in common usage Nix builders kinda really expect a `/bin/sh` to exist and probably have sane behavior in basic cases :).

[...]

PR #3038, opened 2019-08-09
handle empty SANDBOX_SHELL

volth:
`nix-build --sandbox ....` sometimes fails with
`error: while setting up the build environment: getting attributes of path '': No such file or directory`

Fixes https://github.com/NixOS/nix/issues/2673

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -64,7 +64,8 @@ Settings::Settings()
     }
 
 #if defined(__linux__) && defined(SANDBOX_SHELL)
-    sandboxPaths = tokenizeString<StringSet>("/bin/sh=" SANDBOX_SHELL);
+    if (SANDBOX_SHELL[0] != '\0')
+        sandboxPaths = tokenizeString<StringSet>("/bin/sh=" SANDBOX_SHELL);
 #endif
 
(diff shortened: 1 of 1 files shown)


edolstra:
Shouldn't this be handled at compile time? I mean, that's what the `if defined(SANDBOX_SHELL)` is for. `SANDBOX_SHELL` shouldn't be set to an empty string.

volth:
> Shouldn't this be handled at compile time?

one cannot just write
```cpp
#if defined(SANDBOX_SHELL) && (SANDBOX_SHELL != "")
```
It is doable but via a unreadable hack

matthewbauer:
Perhaps changing this line here:

https://github.com/NixOS/nix/blob/85488a93ec3b07210339f2b05aa93e970f9ac3be/src/libstore/local.mk#L42

to something like:

```
ifneq ($(sandbox_shell),)
libstore_CXXFLAGS += -DSANDBOX_SHELL="\"$(sandbox_shell)\""
endif
```

would work better?

volth:
> would work better?

Yes, if C++ compiler is not smart enough to optimize `""[0] == 0` out

edolstra:
Closing in favor of #3132.

PR #3052, opened 2019-08-22
docs: Fix formatting of aws commands

TheSirC:
Fixing the formatting of the examples 13.1 and 13.2 to send back better formatted commands akin to the one in the sixth point of Example 14.1 (`perl = perl;`).

### DIFF
--- a/doc/manual/packages/s3-substituter.xml
+++ b/doc/manual/packages/s3-substituter.xml
@@ -173,11 +173,11 @@ the S3 URL:</para>
 
 
   <example><title>Uploading with a specific credential profile for Amazon S3</title>
-    <para><command>nix copy --to 's3://example-nix-cache?profile=cache-upload&amp;region=eu-west-2' nixpkgs.hello</command></para>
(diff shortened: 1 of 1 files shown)


grahamc:
Hmm... I think `<command>` is a better choice than `<programlisting>` but you're right that the display of the command is terrible! I'm going  to take a minute to see if there is a simple improvement on the display side before merging this.

TheSirC:
I thought it was a better match formatting-wise but I do agree that `<command>` should semantically be the choice here.

grahamc:
I'm not certain this is "enough". What do you think (before & after in the same picture)?

image

This improves other `<command>`s too:

image

vs:

image

TheSirC:
Can't see properly because the crops hides the commands :disappointed:

grahamc:
oh, sorry, hehe the example is the bold `nix-serve` bit in the center of the two last pictures :)

TheSirC:
> oh, sorry, hehe the example is the bold `nix-serve` bit in the center of the two last pictures :)

Yeah sorry I just got it, silly me.

grahamc:
I opened a PR to make this nicer: https://github.com/NixOS/nixos-homepage/pull/298

TheSirC:
Thanks a lot. I am closing this then !

PR #3053, opened 2019-08-22
Drop Store::addToStore taking a string.

gleber:
This removes all uses of this function, in most cases wrapping
previously-provided string with a StringSource.

@edolstra There are couple of places where, I am moving data from a `StringSink` to a `Source` via `StringSource source(sink.s)`. It is not clear to me if this is the best way. This is my first contribution, so please bear with me. My long-term goal is to investigate the Nix code base enough to look into implementing the intentional store capability.

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -111,10 +111,11 @@ void BinaryCacheStore::writeNarInfo(ref<NarInfo> narInfo)
         diskCache->upsertNarInfo(getUri(), hashPart, std::shared_ptr<NarInfo>(narInfo));
 }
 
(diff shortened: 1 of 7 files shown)


zimbatm:
nice! I was touching this code recently and it's nice to see that extra complexity go.

edolstra:
I get a test failure in `export.sh`:
```
    + nix-store --import
    error: bad archive: input doesn't look like a Nix archive
```

The reason may be that draining a `Source` to read a NAR will cause any subsequent data from the same `Source` to be discarded. This happens with Nix exports and with serialization of NARs in the worker/serve protocol.

gleber:
Superseded by #3639

Ericson2314:
Oh i totally messed this. Bummer.

PR #3055, opened 2019-08-23
search: Don't omit version info when package name is matched

### LINKED ISSUE #2778: `nix search` drops version when search term matches package name
When the `nix search` search term matches on the package name, the `nix search` output drops the version from the name. If the term doesn't match on the name but instead matches on the attribute path or description, it displays the version.

For example:

```
> nix search libgnurl
warning: using cached results; pass '-u' to update the cache
* nixpkgs.libgnurl (libgnurl)
  A fork of libcurl used by GNUnet
> nix search nixpkgs.libgnurl
warning: using cached results; pass '-u' to update the cache
* nixpkgs.libgnurl (libgnurl-7.64.0)
  A fork of libcurl used by GNUnet
```

I'm assuming this is a consequence of using a different code path when emitting highlighting for the match in the package name.

nix (Nix) 2.2.2

lilyball:
The `name` variable already includes highlighting in addition to the version info, so with the old code it would print the version info only if `nameMatch` is empty.

Fixes #2778.

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -178,7 +178,7 @@ struct CmdSearch : SourceExprCommand, MixJSON
                             results[attrPath] = fmt(
                                 "* %s (%s)\n  %s\n",
                                 wrap("\e[0;1m", hilite(attrPath, attrPathMatch, "\e[0;1m")),
-                                wrap("\e[0;2m", hilite(name, nameMatch, "\e[0;2m")),
+                                wrap("\e[0;2m", name),
                                 hilite(description, descriptionMatch, ANSI_NORMAL));
                         }
                     }


lilyball:
CC @dtzWill

lilyball:
I tested locally and confirmed that it still highlights the match in the name.

Screen Shot 2019-08-23 at 12 28 44 PM

unode:
Anything holding this merge back? Just bumped into this issue too.

blinry:
Me too! I think that merging this could avoid a lot of confusion for beginners.

lilyball:
I still want this. Didn’t realize there was a merge conflict though. I’ll have to look at that when I get a chance.

lilyball:
I'm not really sure what the state of `nix search` is on master right now. I especially don't know how to get it to reproduce the Nix 2.3 behavior of searching my channel (experimentally I can do something like `nix search -f '<nixpkgs>' foo` but this seems to just look up `nixpkgs.foo` instead of searching). Also, testing this right now, it's printing the version but not the name!

Looking at the code right now, it looks like the version info is indeed fixed, in that it now adds ` ($version)` if the version is non-empty. But the code that's supposed to print the attribute is broken (at least, in the `nix search -f '<nixpkgs>' foo` invocation, probably because it's not actually searching but just looking up a specific attribute). And it's not printing the name (it calculates the name highlight and then throws it away).

https://github.com/NixOS/nix/blob/6678e98411cd3bfe8143a4ba1b35d1451babce32/src/nix/search.cc#L134-L142

It does appear that "don't omit version info" is indeed fixed in master (but not in 2.3-maintenance), but it's been traded for other breakage. Either way I'm going to close out this issue as the other breakage is orthogonal to the version.

PR #3056, opened 2019-08-23
operators: document exact precedence, split up similar operators

grahamc:
Incorporating the additional data from https://gist.github.com/joepie91/c3c047f3406aea9ec65eebce2ffd449d

### DIFF
--- a/doc/manual/expressions/language-operators.xml
+++ b/doc/manual/expressions/language-operators.xml
@@ -190,7 +190,7 @@ weakest binding).</para>
         <entry>11</entry>
       </row>
       <row>
-        <entry>Logical And</entry>
+        <entry>Logical AND</entry>
         <entry><replaceable>e1</replaceable> <literal>&amp;&amp;</literal>
         <replaceable>e2</replaceable></entry>
         <entry>left</entry>
@@ -198,7 +198,7 @@ weakest binding).</para>
         <entry>12</entry>
       </row>
       <row>
-        <entry>Logical Or</entry>
+        <entry>Logical OR</entry>
         <entry><replaceable>e1</replaceable> <literal>||</literal>
         <replaceable>e2</replaceable></entry>
         <entry>left</entry>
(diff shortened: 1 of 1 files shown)


joepie91:
Hmm, I think this would benefit from clarifying the order of the precedence (lowest = strongest) given that it wouldn't be obviously one way or the other, and possibly also how the associativity plays into it. I don't know whether the phrase from the original gist would fit in here, though.

zimbatm:
There is also `[]` which is just between SELECT and APP.

```
nix-repl> a = { b = 4; }
nix-repl> [ a.b ]
[ 4 ]
nix-repl> foo = x: x*x
nix-repl> [ foo 4 ]
[ «lambda @ (string):1:2» 4 ]
nix-repl> [ -4 ]
error: syntax error, unexpected '-', at (string):1:3
```

grahamc:
The paragraph right before this table already says which is weak and which is strong :)

Is [..] an operator?

grahamc:
https://github.com/grahamc/nix/blob/92ddce4f463dc3cf2327c832d45f62d985d5e09f/doc/manual/expressions/language-operators.xml#L10

joepie91:
Oh, right, that bit wasn't visible in the diff. My bad :)

doc/manual/expressions/language-operators.xml
```diff
@@ -89,47 +133,90 @@ weakest binding).</para>
...
-          <replaceable>e1</replaceable> <literal>==</literal> <replaceable>e2</replaceable>,
           <replaceable>e1</replaceable> <literal>!=</literal> <replaceable>e2</replaceable>
         </entry>
         <entry>none</entry>
-        <entry>Equality and inequality.</entry>
+        <entry>Inequality.</entry>
+        <entry>11</entry>
       </row>
       <row>
+        <entry>Logical And</entry>
```
edolstra: A bit further the table talks about "Logical AND", so for consistency it should be capitalized here as well.

PR #3057, opened 2019-08-24
Pass --no-net to improve race condition in gc-concurrent

grahamc:
It looks like the no-net option broke the gc-concurrent tests inside a nix build. 
this ... seems ... to fix it.

```

++ nix add-to-store --name garbage1 ./tarball.sh
--

```


grahamc:
No good:

```

gc-B> 1...
--


```

PR #3064, opened 2019-08-27
nix-daemon.service: add install section.

pszubiak:
The systemd software suite requires the install section to enable a service. Without this section, it's not possible to enable nix-daemon service. The multi-user seems to be a good target for nix-daemon.

Signed-off-by: Piotr Szubiakowski <pszubiak@eso.org>


edolstra:
Thanks!

PR #3071, opened 2019-08-29
builtins.currentSystem: document system/--system

### LINKED ISSUE #3029: Document the `--system` command line option
Various `nix` commands like `nix-build` accept the `--system` option for setting the value of `builtins.currentSystem`.

However this option is not documented in any man page.

grahamc:
Add a note to the builtins.currentSystem entry that `--system` and the `system` params change the detected value.

Closes #3029.

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -208,7 +208,10 @@ if builtins ? getEnv then builtins.getEnv "PATH" else ""</programlisting>
     evaluates to the Nix platform identifier for the Nix installation
     on which the expression is being evaluated, such as
     <literal>"i686-linux"</literal> or
-    <literal>"x86_64-darwin"</literal>.</para></listitem>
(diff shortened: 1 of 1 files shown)


grahamc:
cc @basvandijk

edolstra:
I'm not sure we *should* document these since messing with `system` is kind of a bad thing to do (since it confused `build.cc`).

edolstra:
Also, `--system` clashes with the option with the same name in `nix-env -q`.

grahamc:
Hmm interesting... I'm not sure what to do. I somewhat often need to use `--system`, when doing macos builds from a linux box, or linux builds from a mac box. Is the preferred option to make the system argument a `--argstr`?

edolstra:
Yes, that's better because otherwise Nix may try to build Linux packages on macOS.

Ideally `--system` would only control `builtins.currentSystem` and the build system type wouldn't be configurable at all.

matthewbauer:
Yeah `--extra-platforms` is now able to override the build system, I think.

PR #3073, opened 2019-08-30
Add an option to print the logs in a machine-readable format

regnat:
Add a `--log-format` option which can be used to print the logs in json (pretty similar to the json logger used internally) so that external tools can easily consume it.

This required moving things around a bit because nix 2.0-style commands where overriding the logging mechanism to display the status bar (which made the `--log-format` option unusable). As a nice side effect, nix 1.0-style commands can now use `--log-format bar` to get the status bar


src/nix/main.cc
```diff
@@ -79,8 +78,8 @@ struct NixArgs : virtual MultiCommand, virtual MixCommonArgs
         mkFlag()
             .longName("print-build-logs")
             .shortName('L')
-            .description("print full build logs on stderr")
-            .set(&printBuildLogs, true);
+            .description("print full build logs on stderr. DEPRECATED, use '--log-format bar-with-logs' instead")
```
edolstra: This message implies that `-L` is also deprecated, which should not be the case.

src/libutil/logging.cc
```diff
@@ -68,6 +69,34 @@ class SimpleLogger : public Logger
 
 Verbosity verbosity = lvlInfo;
 
+typedef enum {
+    logFormatRaw,
+    logFormatJSON,
+    logFormatBar,
+    logFormatBarWithLogs,
+} LogFormat;
```
edolstra: Can you move most of this to `src/nix`? That way `src/libutil` only contains the abstract Logger class and the implementations are in `src/nix`.

src/libutil/logging.cc
```diff
@@ -162,17 +210,67 @@ struct JSONLogger : Logger
...
+ * A json logger intended for external consumption
+ * (contrary to 'JSONLogger' which is an internal thing
+ */
+struct ExternalJSONLogger : JSONLogger
+{
+    ExternalJSONLogger(Logger & prevLogger) : JSONLogger(prevLogger) { }
+
+    nlohmann::json jsonActivityType(ActivityType type) override
+    {
+        switch (type) {
```
edolstra: I think it's more maintainable to log the activity type and result type as a numerical value.

src/libutil/logging.cc
edolstra: Or alternatively to `src/libmain`.

src/nix/main.cc
regnat: Indeed. I've removed it altogether. Tell me if you still want a depreciation warning on `--print-build-logs`

src/libutil/logging.cc
regnat: I've moved most of it to `libmain`. The implementation of `SimpleLogger` is still in that file, and `JSONLogger` is also still in `libutil` (because it's strongly tied to `handleJSONLogMessage` and that one is used in libstore). Does that look good?

src/libutil/logging.cc
regnat: It's an annoying tradeoff to make because magic numbers are definitely not user-friendly, but I agree these custom-print functions will probably cause some issue at some point (unless there's a way to auto-generate it?).

I've removed the custom json logger in 1e893d4b94981dae1aa1a0f69ae8864c7a3c47ae, so we're back to numerical values.

regnat:
I'm having some second-though on that PR: The internal json isn't really nice to work with from outside of Nix, so it would be nicer to first settle on a nicer format before being tied to a non-so-nice representation

zimbatm:
What makes it not so nice to use?

regnat:
There's a number of small things which either make it non-intuitive or require some non-trivial logic to get some useful information from it. In particular I noted:

- The format of the messages isn't really standardized between the different kind of messages (for instance most messages have a `text` field for storing some free-form message, except the `msg` message type which has a `msg` field for that. Also, most messages have a `level` field, but not all)
- Some information is packed into the `fields` field which contains a list of free-form elements, and it's up to the user to guess what each element corresponds to
- Amongst the most useful messages types are `activityStart` and `activityStop` which indicate when each phase of the internal pipeline (instantiation, substitution, hooks) start and stop. But the `activityStop` messages only contain an id which maps them to a previous `activityStart`, meaning that to make sense of it a client must keep internally a map of all the started activities to know which activity has ended, and retrieve from that the type of activity, its various parameters, etc…
- The activity types are magic numerical constants (so quite hard to interpret), although this has been discussed in https://github.com/NixOS/nix/pull/3073#discussion_r319624567 and is likely to stay that way

(note that I'm not at all saying that it's a bad format per se, it seems to work quite well inside of Nix, but it's probably not a good format for an external consumption by a client which doesn't know the internals of Nix)

regnat:
I'm lacking of time to work on this right now. @edolstra  What do you think of renaming the `json` log-format to `internal-json` merge this as it-is? That way we already have an (explicitely unstable and not user-friendly) machine-readable log, and we keep the possibility to add a nicer `json` log-format later.

edolstra:
@regnat Yes, I think this looks good as-is.

src/libmain/loggers.cc
```diff
@@ -0,0 +1,54 @@
...
+#include <atomic>
+#include <nlohmann/json.hpp>
+
+namespace nix {
+
+typedef enum {
+    logFormatRaw,
+    logFormatJSON,
+    logFormatBar,
+    logFormatBarWithLogs,
```
edolstra: What is `bar-with-logs`? Is it the same as `nix --print-build-logs / -L`?

src/libmain/loggers.cc
```diff
@@ -0,0 +1,54 @@
...
+    case logFormatRaw:
+        return makeSimpleLogger();
+    case logFormatJSON:
+        return new JSONLogger();
+    case logFormatBar:
+        return createProgressBar();
+    case logFormatBarWithLogs:
+        return createProgressBar(true);
+    default:
+        throw Error(format("Invalid log format '%i'") % logFormat);
```
edolstra: This should be written as
```
        throw Error("invalid log format '%i'", logFormat);
```
Also note that exception messages are not capitalized.

src/libmain/loggers.cc
zimbatm: yes: https://github.com/tweag/nix/blob/5bc27445e686fdc1be849f4183a1ff6ada9a921c/src/nix/main.cc#L78-L82

src/libmain/loggers.cc
zimbatm: fixed in 5bc2744

zimbatm:
and now with documentation. @edolstra is there anything else you would like to see happening for this PR?

regnat:
Just noticed that this somehow cause `builtins:buildEnv` to fail with "Bad file descriptor" when using the daemon unless we pass at least four `-v` flags to nix. I'll give it a look

zimbatm:
@regnat I don't think the sandboxed builder should emit any logs, this is the real issue IMO.

zimbatm:
stracing the nix-daemon:
```
[pid 12635] write(4, "gmlo\0\0\0\0P\0\0\0\0\0\0\0@nix {\"action\":\""..., 96) = -1 EBADF (Bad file descriptor)
[pid 12635] write(2, "error: writing to file: Bad file"..., 44) = 44
[pid 12630] read(10, "error: writing to file: Bad file"..., 4096) = 44
[pid 12630] write(4, "gmlo\0\0\0\0,\0\0\0\0\0\0\0error: writing t"..., 64) = 64
[pid 12630] write(4, "TLSR\0\0\0\0\t\0\0\0\37)\0\0e\0\0\0\0\0\0\0\1\0\0\0\0\0\0\0"..., 96) = 96
[pid 12630] select(11, [10], NULL, NULL, NULL <unfinished ...>
[pid 12635] +++ exited with 1 +++
(edited)
```
pid 12635 is a build sandbox

zimbatm:
when the sandbox starts, it closes all the existing file descriptors except 1 and 2 (and 12 twice):
```
[pid 12635] close(12)                   = 0
[pid 12635] close(3)                    = 0
[pid 12635] close(4)                    = 0
[pid 12635] close(5)                    = 0
[pid 12635] close(6)                    = 0
[pid 12635] close(7)                    = 0
[pid 12635] close(8)                    = 0
[pid 12635] close(9)                    = 0
[pid 12635] close(10)                   = 0
[pid 12635] close(11)                   = 0
[pid 12635] close(12)                   = -1 EBADF (Bad file descriptor)
```

edolstra:
This is probably caused by `logger` not being reset in the child process. `startProcess()` does this:
```
        if (!options.allowVfork)
            logger = makeDefaultLogger();
```
which means that the child has the wrong logger if `allowVfork = true`.

regnat:
@edolstra I've fixed the file descriptor issue (was apparently due to `vfork` indeed) and rebased on top of master − I've also found why the `nix-env` failure hadn't been caught in the tests.

src/libmain/progress-bar.cc
```diff
@@ -153,7 +164,7 @@ class ProgressBar : public Logger
         state->activitiesByType[type].its.emplace(act, i);
 
         if (type == actBuild) {
-            auto name = storePathToName(getS(fields, 0));
+            auto name = getName(fields, 4);
```
edolstra: This no longer abbreviates the name, i.e. the progress bar now looks like this:
```
[1/0/2 built, 16.7 MiB DL] building /nix/store/g29xgw3fy960ghskj45x4x1mjv6jv3l2-nix-tarball-2.4pre7196_d94cc07 (autoconfPhase): autoreconf: running: ...
```

regnat: Indeed, this was happening when using an older daemon (because the messages sent back by the daemon have been extended to include the (abbreviated) derivation name, but there's a fallback code for old daemons that don't do it, but it was missing a call to `storePathToName`).
Should be good now

src/libmain/progress-bar.cc
```diff
@@ -17,20 +16,32 @@ static std::string getS(const std::vector<Logger::Field> & fields, size_t n)
     return fields[n].s;
 }
 
+static std::string storePathToName(std::string path)
```
edolstra: Maybe use the original version? `std::string_view` is cheaper than `std::string`.

Profpatsch:
Might I ping this? :)

regnat:
I've been reminded of this last week, so I rebased this on top of the latest master and simplified it as much as I could (well, nearly rewritten it from scratch actually :thinking: ), let me know what you think

src/libmain/loggers.cc
```diff
@@ -0,0 +1,51 @@
+#include "loggers.hh"
+#include "../nix/progress-bar.hh"
```
edolstra, line 2: This makes `libmain` depend on `nix`, which is a layering violation. It means you can't link against `libmain` anymore.

regnat: Mh that's annoying :thinking: 

I moved `progress-bar` to `libmain` to fix that, but it also required moving `paths` to `libutil` to prevent `libmain` from depending on `libexpr`. Not sure if that's a great idea

edolstra:
Merged, thanks!

Issue #719, opened 2015-11-24
Nix garbage collects too much


wizeman:
I have a NixOS configuration with the following contents:
```nix
{ config, pkgs
, ... }:

let

  kodiOff_src = pkgs.srcOnly {
    name = "kodiOff-src";
    src = /home/tarballs/autogen/kodi-off.tar.gz;
  };

  kodiOff_pkg = import "${kodiOff_src}/default.nix" {};

in {
# (...)
  systemd.services.kodi-off = {
#   (...)
    serviceConfig = {
      Type = "simple";
      ExecStart = "${kodiOff_pkg}/bin/kodi-off";
    };
  };
# (...)
}
```
Now, even though I have the following options in the `/etc/nix/nix.conf` of my private Hydra server:
```
gc-keep-outputs = true
gc-keep-derivations = true
```

... and even though the configuration root hasn't been gc'ed, a Nix garbage collection still appears to delete the output of the `pkgs.srcOnly` derivation.

The above configuration is built by my private Hydra server, but if a garbage collection runs after the configuration is built, then my laptop (which is configured to use my Hydra server as a binary cache) fails to download the package during `nixops deploy`, reporting a `410 Gone` error instead:

```
building all machine configurations...
fetching path ‘/nix/store/cvz1fpbsvqqq36xwyzc33cp7jy4r7lg3-kodiOff-src’...

*** Downloading ‘http://homeserv:3000/nar/cvz1fpbsvqqq36xwyzc33cp7jy4r7lg3-kodiOff-src’ to ‘/nix/store/cvz1fpbsvqqq36xwyzc33cp7jy4r7lg3-kodiOff-src’...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (22) The requested URL returned error: 410 Gone

bzip2: Compressed file ends unexpectedly;
        perhaps it is corrupted?  *Possible* reason follows.
bzip2: Inappropriate ioctl for device
        Input file = (stdin), output file = (stdout)

It is possible that the compressed file(s) have become corrupted.
You can use the -tvv option to test integrity of such files.

You can use the `bzip2recover' program to attempt to recover
data from undamaged sections of corrupted files.

error: unexpected end-of-file
[...]

PR #3086, opened 2019-09-05
[WIP] add --root-realisations to nix-instantiate

hyperfekt:
Currently using IFD (or similar mechanisms such as readFile from a derivation) together with garbage collection is a bad experience, since all derivations required for evaluation will have to be completely rebuilt after each run.

This PR adds an option `--root-realisations` to `nix-instantiate` which mirrors `--add-root` but creates a GC root for derivations used for evaluation instead of those that are the result of the evaluation.

There is some duplication of  code with `--add-root`, but abstracting over both is complicated by `--add--root` using successive numbering for results instead of putting them in a folder and it outputting paths to the GC roots instead of the store paths.

At this point, especially since this is my first PR to Nix, and I'm not really familiar with C++, I'm mostly looking for general feedback on chance of this feature in the abstract being merged, whether to change `--add-root` behavior (numbered outputs have always seemed strange to me since there's potentially garbage left over with successive invocations), and general direction of implementation.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1742,6 +1742,11 @@ bool EvalState::eqValues(Value & v1, Value & v2)
     }
 }
 
+PathSet const & EvalState::realisedDerivations() const
+{
+    return realisedPaths;
+}
+
 void EvalState::printStats()
 {
     bool showStats = getEnv("NIX_SHOW_STATS", "0") != "0";
--- a/src/libexpr/eval.hh
+++ b/src/libexpr/eval.hh
@@ -280,7 +280,11 @@ public:
 
     void realiseContext(const PathSet & context);
 
+    PathSet const & realisedDerivations() const;
+
 private:
+
+    PathSet realisedPaths;
 
     unsigned long nrEnvs = 0;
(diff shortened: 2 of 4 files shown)


Infinisil:
Neat! I think this would fix https://github.com/NixOS/nix/issues/719

hyperfekt:
> Neat! I think this would fix #719

That's the intent, thanks for linking! I was planning to make appropriate changes to nixos-rebuild if this gets merged, potentially also rooting build time dependencies of the derivations evaluation depends on to improve the situation even further.

Infinisil:
For things downloaded with the builtin fetchers (`builtins.fetchTarball` and `builtins.fetchurl` should work too) I just found this workaround: By using indirect GC roots you can gcroot the whole ~/.cache/nix/tarballs directory:
```
find ~/.cache/nix/tarballs -type l -execdir ln -svf ~/.cache/nix/tarballs/{} /nix/var/nix/gcroots/auto/{} \;
```

Ninlives:
Any progress for this pr? I too think gc conflicts with IFD, especially for regions withouts good internet connections :/

PR #3093, opened 2019-09-16
lexer: restrict escape sequences to only meaningful ones

volth:
Nix parser is very relaxing and allows any character after `\` in strings
When the character following `\` is not one of the six which require escaping (`"` `r` `n` `$` `\` `t`) then `\` is silently skipped

so `"\1"` is the same as just`"1"`

This makes writing string constants which have `\` character (regular expressions, etc) challenging and error-prone.

There are few places in `nixpkgs` (https://github.com/NixOS/nixpkgs/pull/68946) where people write "\1" expecting "\1" , etc

### DIFF
--- a/src/libexpr/lexer.l
+++ b/src/libexpr/lexer.l
@@ -64,7 +64,7 @@ static Expr * unescapeStr(SymbolTable & symbols, const char * s, size_t length)
             else if (c == 'r') t += '\r';
             else if (c == 't') t += '\t';
             else {
-                assert(c == '"' || c == '\\' || c == '$');
+                assert(c == '"' || c == '\\' || c == '$' || c == '\n');
                 t += c;
             }
         }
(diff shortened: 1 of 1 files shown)


src/libexpr/lexer.l
```diff
@@ -64,8 +65,9 @@ static Expr * unescapeStr(SymbolTable & symbols, const char * s, size_t length)
             else if (c == 'r') t += '\r';
             else if (c == 't') t += '\t';
             else {
-                assert(c == '"' || c == '\\' || c == '$' || c == '\n');
-                t += c;
+              if (!(c == '"' || c == '\\' || c == '$' || c == '\n'))
+                std::cerr << "meaningful escape sequence \"\\" << c << "\" at " << loc->last_line << ":" << loc->last_column << std::endl;
```
dtzWill: Not sure if writing to `std::cerr` is appropriate here, but doesn't look like a logger is in scope either.  Not sure it's warranted but perhaps warnings can be pushed into a deferred list and processed by the caller at the end?

Message seems inverted-- it says "meaningful escape sequence" but only is triggered when a character is not one of the supported ("meaningful"?) ones.

volth: writer/collector for warnings might be needed, yes.
there could be more warnings, on unused variables or needless `rec` keyword or empty `let in`.

PR #3097, opened 2019-09-18
function-trace: always show the trace

zimbatm:
If the user invokes nix with --trace-function-calls it means that they
want to see the trace. This removes the need to also add `-vvvv` to the command.


src/libexpr/function-trace.hh
```diff
@@ -7,18 +7,22 @@ namespace nix {
 
 struct FunctionCallTrace
 {
-    const Pos & pos;
+    string label;
 
-    FunctionCallTrace(const Pos & pos) : pos(pos) {
+    FunctionCallTrace(const string type, const Pos & pos) {
```
edolstra: This should be `const string & type`.

zimbatm: oops, that was part of an ongoing experiment where I try to qualify the traces a bit better when the pos is undefined (like all the primop calls). Removed it from this PR for now.

PR #3098, opened 2019-09-20
Split build.cc

regnat:
All the build logic of nix is bundled into one gigantic cc file.
From my experience of discovering the nix codebase, this makes
understanding and navigating the code quite difficult, and also cause
the rebuilds to be awfully long once we touch this file.

This MR splits this file to

- Make the code easier to navigate
- Make builds faster (more parallelism and incrementality)

Barring any error on my side, this shouldn't have any functional impact
since it only moves functions around and doesn't change any code.

I couldn't test this outside of my own machine, so I'm not sure that I didn't
break the build on other platforms because of the changes in system
imports. Is there an easy way to test this somewhere?


edolstra:
I'm not really in favor of this since I'd rather 1) rewrite in Rust or 2) rewrite in C++ using threads or fibers instead of the current per-goal state machine. That's more useful than having a lot of code churn that doesn't change functionality.

regnat:
All right, I'll close it then

Ericson2314:
Let's give this another shot!

edolstra:
Yes although I did an experiment today about the overhead of the C++ compilation model and it turns out you get a 7x compilation speedup by *concatenating* all source files into a single compilation unit. In other words, splitting `build.cc` will make non-incremental builds slower unless you have a lot of cores...

Ericson2314:
I've start resurrecting it. Thanks to @regnat splitting the commits vary well, it's less hard to recreate than I feared!

Ericson2314:
@edolstra I prefer to use things like https://mesonbuild.com/Unity-builds.html for that :D. Then downstream builders can choose to either to maximize parallelism or minimize total work without upstream developers having to do extra work to support both.

I personally am building Nix with `-j50` so you know which I prefer :).

domenkozar:
Would be great then to use meson and have both incremental and unity builds :)

Ericson2314:
Reopened as https://github.com/NixOS/nix/pull/4114

PR #3102, opened 2019-09-21
nix doctor: add more logging output to checks

### LINKED ISSUE #3084: nix doctor: actually answer if there are problems or not
```
[davidak@X230:~]$ nix doctor
Store uri: daemon


```

This is the output on my system. I don't understand it.
Please answer the question if there are issues.

I have seen the PR that there can be warnings, but here are none. So i guess i have no problems?

In that case, please output: "No problems found." Maybe add a nice emoji when we support that by default in NixOS. (maybe you can detect if the terminal supports it?)

When problems are found, please output for example "5 problems found." and list them...

Also, please remove the second newline in this output.

cc @LnL7 

---

nix (Nix) 2.2.2
NixOS 19.03.173221.9ca57dc9171 (Koi)

bhipple:
When running `nix doctor` on a healthy system, it just prints the store URI and
nothing else. This makes it unclear whether the system is in a good state and
what check(s) it actually ran, since some of the checks are optional depending
on the store type.

This commit updates `nix doctor` to print an `INFO` log for every check that it
does, and to print a final message saying whether all checks passed or not. This
makes it clear to the user whether they've passed the checkup or not with the doctor.

Fixes #3084

### DIFF
--- a/src/nix/doctor.cc
+++ b/src/nix/doctor.cc
@@ -1,10 +1,16 @@
+#include <sstream>
+
 #include "command.hh"
+#include "logging.hh"
 #include "serve-protocol.hh"
 #include "shared.hh"
 #include "store-api.hh"
+#include "util.hh"
 #include "worker-protocol.hh"
 
 using namespace nix;
+
+namespace {
 
 std::string formatProtocol(unsigned int proto)
 {
@@ -14,6 +20,18 @@ std::string formatProtocol(unsigned int proto)
         return (format("%1%.%2%") % major % minor).str();
     }
     return "unknown";
+}
+
+bool checkPass(const std::string & msg) {
+    logger->log(ANSI_GREEN "[PASS] " ANSI_NORMAL + msg);
+    return true;
+}
+
+bool checkFail(const std::string & msg) {
+    logger->log(ANSI_RED "[FAIL] " ANSI_NORMAL + msg);
+    return false;
+}
+
 }
 
 struct CmdDoctor : StoreCommand
@@ -27,13 +45,12 @@ struct CmdDoctor : StoreCommand
 
     std::string description() override
     {
-        return "check your system for potential problems";
+        return "check your system for potential problems and print a PASS or FAIL for each check.";
     }
 
     void run(ref<Store> store) override
     {
-        std::cout << "Store uri: " << store->getUri() << std::endl;
-        std::cout << std::endl;
+        logger->log("Running checks against store uri: " + store->getUri());
 
         auto type = getStoreType();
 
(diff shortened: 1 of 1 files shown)


bhipple:
New output looks like this:

```
λ brh nix →  ./inst/bin/nix doctor
INFO: Store uri = daemon
INFO: Checking that there is only one nix on PATH.
INFO: Checking that all profiles are gcroots.
INFO: Checking that client and daemon store protocol version match.
INFO: All checks passed!
```
And when in the `nix-shell`, with two `nix` binaries in my `$PATH`:
```
[nix-shell:~/src/nix]$ ./inst/bin/nix doctor
INFO: Store uri = daemon
INFO: Checking that there is only one nix on PATH.
WARNING: multiple versions of nix found in PATH.

  /home/bhipple/src/nix/inst/bin
  /nix/store/j18b1ryh2mkrp0k9b1vq1hyq3289qsgr-nix-2.2.2/bin

INFO: Checking that all profiles are gcroots.
INFO: Checking that client and daemon store protocol version match.
WARNING: Not all checks passed!
```

src/nix/doctor.cc
```diff
@@ -43,20 +42,26 @@ struct CmdDoctor : StoreCommand
...
+            std::cout << "INFO: All checks passed!" << std::endl;
+        else {
+            std::cout << "WARNING: Not all checks passed!" << std::endl;
             throw Exit(2);
+        }
     }
 
     bool checkNixInPath()
     {
+        std::cout << "INFO: Checking that there is only one nix on PATH." << std::endl;
```
LnL7: While this is useful information it might be better to only log the steps on debug level instead.

edolstra:
Maybe we can colorize the output? Like in the `warn` function:
```
void Logger::warn(const std::string & msg)
{
    log(lvlWarn, ANSI_RED "warning:" ANSI_NORMAL " " + msg);
}
```

bhipple:
Thanks for the tip @edolstra!  I've rewritten the output to use the standard `Logger` and emit consistent, colorized output like so:

out

In the first run I'm in the `nix-shell`, so I fail the check that I only have a single `nix` binary in `$PATH`; in the second run, I pass all checks.

Looks nice with the colors now!

src/nix/doctor.cc
bhipple: Logically this feels like a test suite, which generally print all the tests run with their status. Particularly with the colorized output now, there's something to be said for the warm fuzzy feeling of seeing a bunch of green PASSes scroll by :) 

Since the tests that we run are conditional on the system config, as well, I think it's reasonable to print them all.  Maybe if someday we have dozens of checks it might be worth reconsidering, but as of now we only have 3.

bhipple:
@edolstra any thoughts on how it looks now?  We could also use the `systemd` output format, which would be:

```
[  OK  ] Foo
[FAILED] Bar
[  OK  ] Baz

vs.

[PASS] Foo
[FAIL] Bar
[PASS] Baz
```

edolstra:
Thanks!

PR #3107, opened 2019-09-26
Always use "bash" for dev-shell

### LINKED ISSUE #2965: nix dev-shell / develop fails when SHELL=zsh
Using the ```dev-shell``` subcommand available in the nixFlakes branch, I get this:

```
$ nix dev-shell nixpkgs:hello
zsh: no such option: rcfile
```

I believe this is from having set ```SHELL=zsh``` and this:

https://github.com/NixOS/nix/blob/d0a769cb061a13ad880c76e5ea69a76150439853/src/nix/shell.cc#L222

rcfile is not available on zsh.

matthewbauer:
Unfortunately zshell does not support the --rcfile arg. Best to
hardcode bash here. Fixes #2965

### DIFF
--- a/src/nix/shell.cc
+++ b/src/nix/shell.cc
@@ -245,7 +245,7 @@ struct CmdDevShell : Common
 
         stopProgressBar();
 
-        auto shell = getEnv("SHELL", "bash");
+        auto shell = string("bash");
 
         auto args = Strings{baseNameOf(shell), "--rcfile", rcFilePath};
 


grahamc:
It would be nice to support more than just bash. I wonder if zsh has something like rcfile?

matthewbauer:
Unfortunately not:

https://stackoverflow.com/questions/2727172/how-to-load-different-zshrc-file-via-commandline-option

zimbatm:
oops.

I wonder if we should not use the `builder` attribute so that if your builder is python, it would start a python shell.

We would need a new interface that says that the builder can accept a `--shell` argument I think, so that it can load all the libraries (or stdenv)

edolstra:
We can't use the builder attribute because that would give us a non-interactive bash.

edolstra:
Maybe we should copy the `nix-shell` behavior, which is to use `bashInteractive` from `nixpkgs`. The modern variant would be to fetch `nixpkgs#bashInteractive`. This ensures that things work on macOS (where the system bash is ancient) and that nested shells work properly.

src/nix/shell.cc
```diff
@@ -245,7 +245,7 @@ struct CmdDevShell : Common
 
         stopProgressBar();
 
-        auto shell = getEnv("SHELL", "bash");
+        auto shell = "bash";
```
edolstra: This doesn't compile because it infers the wrong type for `shell` (`char *` rather than `std::string`).

zimbatm:
> Maybe we should copy the nix-shell behavior, which is to use bashInteractive from nixpkgs. The modern variant would be to fetch nixpkgs#bashInteractive. This ensures that things work on macOS (where the system bash is ancient) and that nested shells work properly.

It's quite common to pin nixpkgs using `builtins.fetchTarball` in the code and completely ignore NIX_PATH. So this might risk of loading a version of bash coming from another channel. I guess once flakes is finished it will be easier to unify everything under one roof.

edolstra:
@zimbatm Ideally we would use the `nixpkgs` flake from the inputs of the top-level flake.

BTW, commit 9b9de3a5e321c764ec018b32c1f949a49b0c69ef makes it feasible to support other shells since it separately extracts environment variables and bash functions.

balsoft:
@zimbatm 

> So this might risk of loading a version of bash coming from another channel

Why is this an issue for an interactive shell? I would imagine that it's better to do this than to use `bash` from path in most cases. Am I missing something?

matthewbauer:
Closing now that https://github.com/NixOS/nix/pull/3565 is merged

PR #3110, opened 2019-09-29
repl: readline completion

arcnmx:
Just because not having completions was bothersome.


arcnmx:
Cruel bot, I still use this daily and would really like to see either PR land in some capacity :(

PR #3111, opened 2019-09-29
fix docs because nix.conf is a store file on Nixos (read only)

chrissound:
https://unix.stackexchange.com/questions/544340/nixos-unable-to-chmod-file-changing-permissions-of-etc-nix-nix-conf-read-o

### DIFF
--- a/doc/manual/packages/binary-cache-substituter.xml
+++ b/doc/manual/packages/binary-cache-substituter.xml
@@ -57,13 +57,15 @@ of Firefox, Nix will first check if the path is available on the
 server <literal>avalon</literal> or another binary caches. If not, it
 will fall back to building from source.</para>
 
-<para>You can also tell Nix to always use your binary cache by adding
+<para>You can also tell Nix to always use your binary cache, if using Nixos by setting the `nix.extraOptions` in `configuration`, otherwise by adding
 a line to the <filename linkend="sec-conf-file">nix.conf</filename>
 configuration file like this:
 
 <programlisting>
 binary-caches = http://avalon:8080/ https://cache.nixos.org/
 </programlisting>
+         
+<para>More information is also available here: https://nixos.wiki/wiki/Binary_Cache</para>
 
 </para>
 


doc/manual/packages/binary-cache-substituter.xml
```diff
@@ -57,13 +57,15 @@ of Firefox, Nix will first check if the path is available on the
...
-<para>You can also tell Nix to always use your binary cache by adding
+<para>You can also tell Nix to always use your binary cache, if using Nixos by setting the `nix.extraOptions` in `configuration`, otherwise by adding
 a line to the <filename linkend="sec-conf-file">nix.conf</filename>
 configuration file like this:
 
 <programlisting>
 binary-caches = http://avalon:8080/ https://cache.nixos.org/
 </programlisting>
+         
+<para>More information is also available here: https://nixos.wiki/wiki/Binary_Cache</para>
```
zimbatm: If there are relevant information there, they should be copied into the manual. The NixOS wiki is great but not part of the official documentation.

doc/manual/packages/binary-cache-substituter.xml
```diff
@@ -57,13 +57,15 @@ of Firefox, Nix will first check if the path is available on the
 server <literal>avalon</literal> or another binary caches. If not, it
 will fall back to building from source.</para>
 
-<para>You can also tell Nix to always use your binary cache by adding
+<para>You can also tell Nix to always use your binary cache, if using Nixos by setting the `nix.extraOptions` in `configuration`, otherwise by adding
```
zimbatm:
```suggestion
<para>You can also tell Nix to always use your binary cache. When using Nixos by setting the `nix.extraOptions` in `configuration`, otherwise by adding
```

edolstra: Well, changing it to ". When" creates a sentence without a main clause. But the ", if" isn't great stylistically either.

Also "NixOS" rather than "Nixos".

PR #3114, opened 2019-10-02
Add libatomic for 32-bit ARM

### LINKED ISSUE #3113: ARMv7 Build Failure: undefined reference to __atomic_fetch_add_8
I am following the instructions documented here: 
https://nixos.org/nix/manual/#ch-installing-source

Building of nix-2.3 on armv7 (Raspbian) fails:

```
# make
  LD     src/libutil/libnixutil.so
/usr/bin/ld: src/libutil/logging.o: in function `nix::Activity::Activity(nix::Logger&, nix::Verbosity, nix::ActivityType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<nix::Logger::Field, std::allocator<nix::Logger::Field> > const&, unsigned long long)':
/usr/include/c++/8/bits/atomic_base.h:514: undefined reference to `__atomic_fetch_add_8'
make: *** [mk/lib.mk:104: src/libutil/libnixutil.so] Error 1
```

The `configure` script was run as
```
./configure --disable-seccomp-sandboxing
```

## System

This is a Raspbian distribution, specifically
```
Linux 0b8f15da08f8 4.9.184-linuxkit #1 SMP Tue Jul 2 22:58:16 UTC 2019 armv7l GNU/Linux
```

The following packages have been installed:
```
build-essential
m4
libtool
libboost-all-dev
libssl-dev
libbz2-dev
libsqlite3-dev
libbrotli-dev
liblzma-dev
libcurl4-openssl-dev
libgc-dev
```

Furthermore libeditline has been installed manually from `https://github.com/troglobit/editline/releases/download/1.16.1/editline-1.16.1.tar.xz`

matthewbauer:
Fixes #3113

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -157,6 +157,11 @@ AX_BOOST_BASE([1.66], [CXXFLAGS="$BOOST_CPPFLAGS $CXXFLAGS"], [AC_MSG_ERROR([Nix
 # ends up with LDFLAGS being empty, so we set it afterwards.
 LDFLAGS="$BOOST_LDFLAGS $LDFLAGS"
 
+# Boost atomic needs GCC libatomic on 32-bit ARM
+case "$host_cpu" in
+  armv5*|armv6*|armv7*) LIBS="-latomic $LIBS"
+esac
+
 
 # Look for OpenSSL, a required dependency.
 PKG_CHECK_MODULES([OPENSSL], [libcrypto], [CXXFLAGS="$OPENSSL_CFLAGS $CXXFLAGS"])
--- a/configure.ac
+++ b/configure.ac
@@ -157,11 +157,18 @@ AX_BOOST_BASE([1.66], [CXXFLAGS="$BOOST_CPPFLAGS $CXXFLAGS"], [AC_MSG_ERROR([Nix
 # ends up with LDFLAGS being empty, so we set it afterwards.
 LDFLAGS="$BOOST_LDFLAGS $LDFLAGS"
 
-# Boost atomic needs GCC libatomic on 32-bit ARM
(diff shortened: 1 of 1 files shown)


edolstra:
This is weird. Shouldn't this be something handled by gcc / libstdc++?

matthewbauer:
> This is weird. Shouldn't this be something handled by gcc / libstdc++?

Yeah I would expect that to happen, but it doesn't look like it has. At least boost's atomic, which uses this symbol should have this check. Most platforms have hardware atomics, but arm32 does not so it needs to use the software version. It is probably necessary for PPC too.

Other projects like:

- KDE (libatomic, provides a software implementation of atomic built-ins, when needed)[https://phabricator.kde.org/D16184)
- Mesa (disk cache: Link with -latomic if necessary)[https://gitlab.freedesktop.org/mesa/mesa/commit/498faea103aa7966b435f21d8ff5e36172389b1e]
- Protobuf (Link to libatomic when necessary (eg. on armv6l))[https://github.com/protocolbuffers/protobuf/commit/56b40a8f055cfd05633054aed5d3466a6e058b72]

do this too.

puckipedia:
Came across this issue while trying to package 2.3.10 Nix, and after backporting the commit, turns out, it never worked! `$LIBS` has not been looked at since Automake was removed in 2014, in commit 0c6d62cf27b3b22fa60bddad16ea8e8d062e4a99! Adding `@LIBS@` to the LDFLAGS in the `Makefile.config.in` seems to have made it work, but feels a bit hacky (I don't entirely know what the interaction of the custom build system is with how autoconf does things by default..)

PR #3116, opened 2019-10-03
Add NIX_FORCE_COLOR env variable to skip removal of ANSI escapes

Ma27:
Currently the build log from a remote builder isn't colorful as usually
no TTY is allocated when accessing the remote machine with SSH. However
using a TTY (e.g. by setting `RequestTTY force`) doesn't work either:

```
cannot build on 'ssh://builder': serialised integer 6791531838262124011 is too large for type 'j'
```

This patch allows to skip the removal of escape sequences by adding the env var
`NIX_FORCE_COLOR=1` (e.g. in `nix.envVars` in the NixOS configuration).

### DIFF
--- a/doc/manual/command-ref/env-common.xml
+++ b/doc/manual/command-ref/env-common.xml
@@ -196,6 +196,15 @@ $ mount -o bind /mnt/otherdisk/nix /nix</screen>
 </varlistentry>
 
 
+<varlistentry><term><envar>NIX_FORCE_COLOR</envar></term>
+
+ <listitem><para>Forces the Nix process to not strip any ANSI sequences
+ from the output to <literal>stderr</literal>. Especially helpful for a
+ <command>nix-daemon</command> on a remote builder.</para></listitem>
+
+</varlistentry>
+
+
 </variablelist>
 
 
--- a/src/libutil/logging.cc
+++ b/src/libutil/logging.cc
@@ -54,7 +54,7 @@ public:
             prefix = std::string("<") + c + ">";
         }
 
-        writeToStderr(prefix + filterANSIEscapes(fs.s, !tty) + "\n");
+        writeToStderr(prefix + filterANSIEscapes(fs.s, !tty && (getEnv("NIX_FORCE_COLOR") != "1")) + "\n");
     }
 
     void startActivity(ActivityId act, Verbosity lvl, ActivityType type,


zimbatm:
env vars are easy to add but also suffer from discoverability. It would be best to introduce a new command-line option for `--colors auto|yes|no` like on most GNU tools.

edolstra:
I don't think this is the right solution. Could you try `ssh-ng://`? It tunnels log messages in the same way as the Nix daemon, so it should work better.

conferno:
> It tunnels log messages in the same way as the Nix daemon, so it should work better.

But.. nix-`daemon` strips ANSI-sequences too.

Relate problem: remote builder (and `nix-daemon`) buffers the builder's output so if there is `fetchurl` downloading a big file with `curl`drawing a progresbar it is not visible to the user, it looks like a hanged build or a network problem.

Ma27:
@zimbatm agreed! I mainly wanted to propose an approach how this could be fixedbas I'm not even sure if that's the best approach.

@edolstra thanks for the hint, didn't know about this and will try this out :)

edolstra:
> But.. nix-daemon strips ANSI-sequences too.

AFAIK the stripping is done in the UI. I think the line
```
act->result(resBuildLogLine, currentLogLine);
```
in `DerivationGoal::flushLine()` passes the original log line verbatim. I'm not sure whether `build-remote` properly tunnels those messages to the caller though. (That's part of the reason for getting rid of the separate build hook, #1221.)

Ma27:
@edolstra sorry for asking, but could you give me an example how to use a remote-machine with `ssh-ng`? As far as I understand this, the feature is used for remote stores, but I'm building remotely by default.

zimbatm:
@Ma27 Eg: `nix-build --builders ssh-ng://foobar` if your user is a trusted user.

Ma27:
This works indeed perfectly fine. Closing then :+1:

ruuda:
It would still be useful to have a way to force colors. For example, want to build something on CI with `nix build --print-build-logs`. The build logs do contain colors, but they are not visible on CI, because there is no TTY on CI, even though it is capable of interpreting ANSI escape sequences.

If I understand correctly, https://github.com/NixOS/nix/blob/73fde9eed06dfdef5d37b3d798cfc98a542a4d73/src/libutil/util.cc#L1409-L1414 is what prevents colors from showing up without a TTY. It would be nice to have a way to override this.

7c6f434c:
Presumably if you run inside expect and set TERM to what you prefer, things should work without changes to Nix?

ruuda:
> Presumably if you run inside expect and set TERM to what you prefer, things should work without changes to Nix?

It’s, not as easy as setting an environment variable, Nix checks [`isatty(3)`](https://linux.die.net/man/3/isatty). One workaround is to wrap the entire thing in [`script --command`](https://linux.die.net/man/1/script).

7c6f434c:
>It’s, not as easy as setting an environment variable, Nix checks [`isatty(3)`](https://linux.die.net/man/3/isatty).

That's why I suggested expect (but nothing against script, either)

PR #3128, opened 2019-10-09
Don't symlink org.nixos.nix-daemon.plist in installer

matthewbauer:
On Catalina, the /nix filesystem might not be mounted at start time.
To avoid this service not starting, we need to keep the launch agent
outside of the Nix store. A wait4pid will hold for our /nix dir to be
mounted.

In the multi-user install script, we originally made sure no previous
references to Nix existed. This prevented any previous installs from
contaminating the new install. However, some users need the ability to
repair their existing Nix installation without uninstalling all
references to Nix. This change allows users with existing Nix
installations to use the installer, while still outputing a warning
message on the dangers of this. As a result, the multi-user install
script work much more like the single-user install script has worked
in the past.

This is a requirement for macOS Catalina users now that
/Library/LaunchDaemons/org.nixos.nix-daemon.plisg is not managed by
the Nix store. If there is ever a change to the .plist, all users will
need to rerun this install script to get the new changes. Otherwise,
changes to the launch daemon will require manual interventions.

/cc @grahamc @edolstra

### DIFF
--- a/misc/launchd/org.nixos.nix-daemon.plist.in
+++ b/misc/launchd/org.nixos.nix-daemon.plist.in
@@ -17,7 +17,7 @@
     <array>
       <string>/bin/sh</string>
       <string>-c</string>
-      <string>/bin/wait4path @bindir@/nix-daemon &amp;&amp; @bindir@/nix-daemon</string>
+      <string>/bin/wait4path @bindir@/nix-daemon &amp;&amp; /nix/var/nix/profiles/default/bin/nix-daemon</string>
     </array>
     <key>StandardErrorPath</key>
     <string>/var/log/nix-daemon.log</string>
--- a/scripts/install-darwin-multi-user.sh
+++ b/scripts/install-darwin-multi-user.sh
@@ -39,7 +39,7 @@ EOF
 
 poly_configure_nix_daemon_service() {
     _sudo "to set up the nix-daemon as a LaunchDaemon" \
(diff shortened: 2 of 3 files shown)


abathur:
I'm not sure whether it is a problem/intentional, but I noticed that d4e51aa has not been released (0847f2f was cherry-picked into 2.3-maintenance in 7a011de).

I ran into it when I tested a reinstall in CI and saw that reinstalls are still blocked. For reference (and searchability), this means that instead of issuing a warning about Nix already being installed, it aborts with the error below, followed by uninstall instructions (which are incomplete on macOS Catalina+ anyways):
```
---- oh no! --------------------------------------------------------------------
There are some relics of a previous installation of Nix at /nix, and
this scripts assumes Nix is _not_ yet installed. Please delete the old
Nix installation and start again.
```

If this is being intentionally left out of 2.3-maintenance, I may need help understanding the trajectory of 2.3.x and the next eventual release from master. I'm working on a PR to improve the Catalina+ install/volume process and may need to recalibrate how it should behave during reinstalls.

abathur:
I'm also wondering how to interpret the warning `This installer may run into issues.` 

Assuming the installer correctly updates old artifacts outside of /nix, how likely is it to mess up something inside /nix? (and what, how bad?)

I had taken this as a generic CYA disclaimer and assumed it was okay since we enabled it here, but after seeing #4239 I realized I may be naive about the risks.

Issue #509, opened 2015-04-01
Remove 1777 permission on /nix/var/nix/{profiles,gcroots}/per-user


edolstra:
The 1777 permission is a potential security risk. (For instance, it allows a user to create a directory with the same name as another user.)

Alternatives:
- Create the per-user directories when a user is created, or during login using a PAM module.
- Move the per-user directories to the user's home directory. E.g. /nix/var/nix/profiles/per-user/alice would become /home/alice/.config/nix/profiles or something like that.

vcunat:
$HOME seems a suitable place for per-user data, although it may bring problems due to GC needing some access to user homes to find roots.

PR #3129, opened 2019-10-09
Don’t use stat --printf in nix-profile.sh

### LINKED ISSUE #3127: Can't bootstrap Nix on Ubuntu 18.04
I was trying to help out on https://discourse.nixos.org/t/custom-installation-of-nix-unset-nix-build-top/4285/3, and found that you can't bootstrap Nix on Ubuntu.

I would expect the following script (run as root) to install a usable Nix on Ubuntu from source, without a prior installation of Nix, or a binary install from `https://nixos.org/nix/...`. It does build and install fine, but the resulting `/usr/local/etc/profile.d/nix.sh` does not work. It seems some of the `@coreutils@` binaries that [it needs](https://github.com/NixOS/nix/blob/master/scripts/nix-profile.sh.in#L3) are in `/bin`, and some are in `/usr/bin`, but `configure` only sets `coreutils` to `/bin`. Neither single directory contains all that is needed.

**Note:** Probably run this in a VM because... system software.

```bash
#!/usr/bin/env bash

set -eu

cores=$(nproc --all)
echo $cores

apt install \
        tar \
        build-essential \
        pkg-config \
        libssl-dev \
        bzip2 \
        libbz2-dev \
        sqlite3 \
        libsqlite3-dev \
        libcurl4-openssl-dev \
        curl \
        cmake \
        libxslt-dev \
        libxml2 \
        libgc-dev \
        liblzma-dev \
        libbrotli-dev \
        libseccomp-dev

# Boost part, uncomment if boost need to be installed
curl -L https://dl.bintray.com/boostorg/release/1.71.0/source/boost_1_71_0.tar.bz2 \
  | tar x --bzip2
pushd boost_1_71_0
./bootstrap.sh
./b2 install
popd

# Editline part
curl -L https://github.com/troglobit/editline/releases/download/1.16.1/editline-1.16.1.tar.xz \
  | tar x --xz
pushd editline-1.16.1
./configure
make all -j $cores -l $cores
make install
popd 

curl -L https://nixos.org/releases/nix/nix-2.3/nix-2.3.tar.xz \
  | tar x --xz

pushd nix-2.3
./configure --enable-gc

make -j $cores -l $cores
make install
popd
```

In a shell

```
vagrant@ubuntu-bionic:~$ sudo mkdir /nix
vagrant@ubuntu-bionic:~$ sudo chown vagrant /nix
vagrant@ubuntu-bionic:~$ . /usr/local/etc/profile.d/nix.sh 
Command 'stat' is available in '/usr/bin/stat'
The command could not be located because '/usr/bin' is not included in the PATH environment variable.
stat: command not found
Command 'id' is available in '/usr/bin/id'
The command could not be located because '/usr/bin' is not included in the PATH environment variable.
id: command not found
Command 'stat' is available in '/usr/bin/stat'
The command could not be located because '/usr/bin' is not included in the PATH environment variable.
stat: command not found
Command 'id' is available in '/usr/bin/id'
The command could not be located because '/usr/bin' is not included in the PATH environment variable.
id: command not found
```

The `Command 'xyz' is available in '/usr/bin/xyz'` and `The command could not be located because ...` text are coming from Ubuntu's command not found handle.

Why does it use a custom `PATH` anyway? Why not just use what it's given?

matthewbauer:
This breaks on some systems where "coreutils" commands are installed
in multiple places. We only needed coreutils here for the ‘stat
--printf’ checks on the user ownership. This removes those checks.
This check should be done in Nix commands like “nix doctor”.

Fixes #3127

### DIFF
--- a/scripts/nix-profile.sh.in
+++ b/scripts/nix-profile.sh.in
@@ -1,7 +1,4 @@
 if [ -n "$HOME" ] && [ -n "$USER" ]; then
-    __savedpath="$PATH"
-    export PATH=@coreutils@
-
     # Set up the per-user profile.
     # This part should be kept in sync with nixpkgs:nixos/modules/programs/shell.nix
 
@@ -10,10 +7,6 @@ if [ -n "$HOME" ] && [ -n "$USER" ]; then
     NIX_USER_PROFILE_DIR=@localstatedir@/nix/profiles/per-user/$USER
 
     mkdir -m 0755 -p "$NIX_USER_PROFILE_DIR"
-
-    if [ "$(stat --printf '%u' "$NIX_USER_PROFILE_DIR")" != "$(id -u)" ]; then
-        echo "Nix: WARNING: bad ownership on "$NIX_USER_PROFILE_DIR", should be $(id -u)" >&2
-    fi
 
     if [ -w "$HOME" ]; then
         if ! [ -L "$NIX_LINK" ]; then
@@ -36,9 +29,6 @@ if [ -n "$HOME" ] && [ -n "$USER" ]; then
         # Create the per-user garbage collector roots directory.
         __user_gcroots=@localstatedir@/nix/gcroots/per-user/"$USER"
         mkdir -m 0755 -p "$__user_gcroots"
-        if [ "$(stat --printf '%u' "$__user_gcroots")" != "$(id -u)" ]; then
-            echo "Nix: WARNING: bad ownership on $__user_gcroots, should be $(id -u)" >&2
-        fi
         unset __user_gcroots
 
         # Set up a default Nix expression from which to install stuff.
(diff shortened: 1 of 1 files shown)


edolstra:
I'd like to get rid of these checks, but currently they're the only thing guarding against https://github.com/NixOS/nix/issues/509.

ElvishJerricco:
Is there a way to get these checks without causing  #3127?

edolstra:
1) Add a helper command to `nix` to check permissions.
2) Move profiles/gcroots to home directories so we don't need a world-writable `/nix/var/nix/{profiles,gcroots}`.

matthewbauer:
Opened https://github.com/NixOS/nix/pull/3135 as an alternative. Still not ideal, but at least will fix this particular issue.

PR #3141, opened 2019-10-12
Downloader: Log configured CA file

xbreak:
*Update*: Rescoped PR to only debug print configured CA file to address concerns by @edolstra.

It seems like a good idea to warn if no trusted CA's has been configured, to help narrow down issues like https://github.com/NixOS/nixpkgs/issues/70939.

*Update*: CA files may come from different sources if e.g. Curl or the TLS library is used from the system and not Nix. Therefore it might be misleading to warn.

In addition I added a `debug` print of the configured CA file for the same reason.

Tested on CentOS and looks like this:
```sh
$ nix-env -i firefox -vvv
...
downloading 'https://cache.nixos.org/nar/0wpvc57l0sv8jxfylfyfwrkxzij6fxvakk4k3zd11iiaanwx8viy.nar.xz'...
starting download of https://cache.nixos.org/nar/0wpvc57l0sv8jxfylfyfwrkxzij6fxvakk4k3zd11iiaanwx8viy.nar.xz
verify TLS: Nix CA file = ''
finished download of 'https://cache.nixos.org/nar/0wpvc57l0sv8jxfylfyfwrkxzij6fxvakk4k3zd11iiaanwx8viy.nar.xz'; curl status = 60, HTTP status = 0, body = 0 bytes
warning: unable to download 'https://cache.nixos.org/nar/0wpvc57l0sv8jxfylfyfwrkxzij6fxvakk4k3zd11iiaanwx8viy.nar.xz': SSL peer certificate or SSH remote key was not OK (60); retrying in 274 ms

### DIFF
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -285,6 +285,7 @@ struct CurlDownloader : public Downloader
             }
 
             if (request.verifyTLS) {
+                debug("verify TLS: Nix CA file = '%s'", settings.caFile);
                 if (settings.caFile != "")
                     curl_easy_setopt(req, CURLOPT_CAINFO, settings.caFile.c_str());
             } else {

### CODE REVIEW

File: src/libstore/download.cc [hunk 1]
```diff
@@ -285,8 +285,12 @@ struct CurlDownloader : public Downloader
             }
 
             if (request.verifyTLS) {
-                if (settings.caFile != "")
+                debug("verify TLS enabled; configured CA file = '%s'", settings.caFile);
```
zimbatm: the debug line is redundant when caFile == "", you could move it inside of the if condition

zimbatm: In that case I would tend to use a similar prefix on both log lines. Eg:
* `verify TLS: CA file = '%s'`
* `WARNING: verify TLS: no CA file provided`

xbreak: Yes, that's better. I've updated the PR.

edolstra:
I'm not sure about this. If libcurl is a native (non-Nix) package or if it's using some native TLS mechanism (I think this is the case on macOS), then it may be able to find certificates even if `settings.caFile` is unset.

xbreak:
> I'm not sure about this. If libcurl is a native (non-Nix) package or if it's using some native TLS mechanism (I think this is the case on macOS), then it may be able to find certificates even if `settings.caFile` is unset.

Any suggestions on how to improve the PR? It e.g. sounds like it wouldn't be useful to read out paths from curl either, since other paths that are unknown to curl might be used as well, depending on the TLS library.

Maybe just keep the debug log then (and rephrase to indicate that the CA file is provided by Nix: "verify TLS: **Nix** CA file = '%s'")?

xbreak:
I updated PR to only debug log the Nix configured CA file. Hope this addresses your concern @edolstra .

PR #3144, opened 2019-10-12
Fix sandbox fallback settings

matthewbauer:
The tmpDirInSandbox is different when in sandboxed vs. non-sandboxed.
Since we don’t know ahead of time here whether sandboxing is enabled,
we need to reset all of the env vars we’ve set previously. This fixes
the issue encountered in https://github.com/NixOS/nixpkgs/issues/70856.


siscia:
Should this be merged?

siscia:
ping

PR #3146, opened 2019-10-13
Fallback unwritable uid_map

### LINKED ISSUE #3145: nix-channel --update fails inside nixos-enter
I've installed NixOS into a chroot using nixos-install. Then I've entered the installation chroot with ``sudo `which nixos-enter` --root ...``. I've got one channel:
```
[root@nixos:/]# nix-channel --list  
nixos https://nixos.org/channels/nixos-19.09
```

When I try to update it, I get:
```
[root@nixos:/]# nix-channel --update
unpacking channels...
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
error: writing to file: Operation not permitted
error: program '/nix/store/l4zaz0z1x9w549xj0m0y9a2p8fik2w8a-nix-2.3/bin/nix-env' failed with exit code 1
```

Please fix the error message to contain more information (at least the file name on which the write failed).

Running the above command under strace, here's the relevant part, more or less:
```
[pid   240] openat(AT_FDCWD, "/proc/247/uid_map", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 16
[pid   240] write(16, "1000 30001 1", 12) = -1 EPERM (Operation not permitted)
[pid   240] close(16)                   = 0
[pid   240] kill(-247, SIGKILL <unfinished ...>
```
(Let me know if you need more)
I've no idea what's going on there. Not sure it matters, but I didn't have user 30001 on the host system (only 30000 as the manual said), but creating it didn't change anything.

matthewbauer:
In some sandboxes, this is unavailable. Let sandboxFallback handle it
too. Fixes #3145.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2395,13 +2395,21 @@ void DerivationGoal::startBuilder()
         uid_t hostUid = buildUser ? buildUser->getUID() : getuid();
         uid_t hostGid = buildUser ? buildUser->getGID() : getgid();
 
-        writeFile("/proc/" + std::to_string(pid) + "/uid_map",
-            (format("%d %d 1") % sandboxUid % hostUid).str());
+        try {
+            writeFile("/proc/" + std::to_string(pid) + "/uid_map",
+                (format("%d %d 1") % sandboxUid % hostUid).str());
 
-        writeFile("/proc/" + std::to_string(pid) + "/setgroups", "deny");
+            writeFile("/proc/" + std::to_string(pid) + "/setgroups", "deny");
 
-        writeFile("/proc/" + std::to_string(pid) + "/gid_map",
-            (format("%d %d 1") % sandboxGid % hostGid).str());
+            writeFile("/proc/" + std::to_string(pid) + "/gid_map",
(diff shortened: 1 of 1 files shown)


src/libstore/build.cc
```diff
@@ -2399,13 +2395,21 @@ void DerivationGoal::startBuilder()
...
+        try {
+            writeFile("/proc/" + std::to_string(pid) + "/uid_map",
+                (format("%d %d 1") % sandboxUid % hostUid).str());
+
+            writeFile("/proc/" + std::to_string(pid) + "/setgroups", "deny");
+
+            writeFile("/proc/" + std::to_string(pid) + "/gid_map",
+                (format("%d %d 1") % sandboxGid % hostGid).str());
+        } catch (SysError & e) {
+            if (settings.sandboxFallback) {
```
zimbatm: Maybe add a debug-level message here? Now that there are multiple locations, it's going to be useful if one wants to determine why a build is not happening inside of the sandbox.

matthewbauer: Yeah that would be good!

PR #3152, opened 2019-10-16
Use `lib.version` instead of the deprecated `lib.nixpkgsVersion`

adamse:


edolstra:
Duplicate of #2291.

adamse:
I see! Thanks for taking a look.

Issue #338, opened 2014-09-16
Document `import <...>` syntax in "Writing Nix Expressions" chapter of manual


cantora:
I'm just starting to learn nix and I found it confusing that `import <nixpkgs>` appears in many examples, yet the angle bracket syntax is not mentioned in the nix expression syntax documentation. I did end up finding an explanation for it in the "Common environment variables" section, but I think it the syntax should be documented with the rest of the syntax documentation.

EDIT:
I would add the documentation myself, but I actually don't feel that I fully understand the angle bracket syntax from reading the explanation in "Common environment variables".

npostavs:
+1, I thought at first this it was some kind of meta-syntax that should be substituted by the user with some other value. It's used all over the docs, in the [FAQ], but never explained...

[FAQ]: https://nixos.org/wiki/FAQ

earksiinni:
I'm also learning Nix and found this frustrating. Just created this PR to fix it: https://github.com/NixOS/nix/pull/3159

@tobiasBora, I noticed your PR (https://github.com/NixOS/nixpkgs/pull/63033) for the Nixpkgs repo, but I assume that yours fixes the issue in Nixpkgs' docs. My PR is for the Nix manual. Perhaps fixing in both places is the best policy.

earksiinni:
Can we close this now that my PR got merged?

PR #3159, opened 2019-10-21
Document import <path> syntax

earksiinni:
Fixes https://github.com/NixOS/nix/issues/338 by changing the start of the manual section on `builtins.import` to the following:

**import _path_, import _&lt;path&gt;_, builtins.import _path_, builtins.import _&lt;path&gt;_**

Load, parse and return the Nix expression in the file _path_. If path is a directory, the file default.nix in that directory is loaded. If the _&lt;path&gt;_ syntax is used, the path will be resolved relative to those listed in the NIX_PATH environment variable (see the section on NIX_PATH for details on how the resolution works). Evaluation aborts if the file doesn’t exist or contains an incorrect Nix expression. import implements Nix’s module system: you can put any Nix expression (such as a set or a function) in a separate file, and use it from Nix expressions in other files.

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -733,21 +733,13 @@ builtins.genList (x: x * x) 5
   <varlistentry xml:id='builtin-import'>
     <term><function>import</function>
     <replaceable>path</replaceable></term>
-    <term><function>import</function>
-    <replaceable>&lt;path&gt;</replaceable></term>
     <term><function>builtins.import</function>
     <replaceable>path</replaceable></term>
-    <term><function>builtins.import</function>
-    <replaceable>&lt;path&gt;</replaceable></term>
 
     <listitem><para>Load, parse and return the Nix expression in the
     file <replaceable>path</replaceable>.  If <replaceable>path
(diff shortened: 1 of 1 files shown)


edolstra:
There is no `import <path>` syntax. `import` is a function that takes a value and `<path>` is a value. So `import <path>` is equivalent to e.g. `let p = <path>; in import p`. The `<path>` syntax is already documented:
```xml
  <para>Paths can also be specified between angle brackets, e.g.
  <literal>&lt;nixpkgs&gt;</literal>. This means that the directories
  listed in the environment variable
  <envar linkend="env-NIX_PATH">NIX_PATH</envar> will be searched
  for the given file or directory name.
  </para>
```

earksiinni:
@edolstra, that makes perfect sense. I didn't realize that paths can be written with angle brackets, which is implied by the fact that `import` is a function, not a special keyword.

I do think that for beginners coming from other languages where `import` often has special syntax, this can be a point of confusion, even though `<path>` is documented. I've updated my PR accordingly to provide a "hint" in the section on `builtins.import`.

What do you think?

edolstra:
Thanks!

PR #3161, opened 2019-10-22
Remove superfluous IAM action for S3 cache

schlarpc:
`s3:ListObjects` isn't a real IAM action, but _is_ the name of an S3 API method. `s3:ListBucket` is the relevant action for that method.

https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazons3.html


zimbatm:
I can confirm that `s3:ListObjects` is an invalid entry.

PR #3163, opened 2019-10-23
nix repl: add :edit command

zimbatm:
This allows to have a repl-centric workflow to working on nixpkgs.

Usage:

    :edit <package> - heuristic that find the package file path

    :edit <path> - just open the editor on the file path

Once invoked, `nix repl` will open $EDITOR on that file path. Once the
editor exits, `nix repl` will automatically reload itself.

### DIFF
--- a/src/nix/edit.cc
+++ b/src/nix/edit.cc
@@ -59,22 +59,15 @@ struct CmdEdit : InstallableCommand
             throw Error("cannot parse line number '%s'", pos);
         }
 
-        auto editor = getEnv("EDITOR", "cat");
-
-        auto args = tokenizeString<Strings>(editor);
-
-        if (editor.find("emacs") != std::string::npos ||
-            editor.find("nano") != std::string::npos ||
-            editor.find("vim") != std::string::npos)
-            args.push_back(fmt("+%d", lineno));
-
-        args.push_back(filename);
-
         stopProgressBar();
+
+        auto args = editorFor(filename, lineno);
 
         execvp(args.front().c_str(), stringsToCharPtrs(args).data());
 
-        throw SysError("cannot run editor '%s'", editor);
+        std::string command;
+        for (const auto &arg : args) command += " '" + arg + "'";
+        throw SysError("cannot run command%s", command);
     }
 };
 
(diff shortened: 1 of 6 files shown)


zimbatm:
This PR borrows some code from `nix edit` which I would like to factor out but don't know where to place it.

Currently this is a feature preview. The doc also needs to be amended with the new command.

bhipple:
This looks nice!  It'd be good to have some test cases though.  Something basic like this:

```
[nix-shell:~/src/nixpkgs]$ echo ":edit boost" | EDITOR=ls ../nix/inst/bin/nix repl .

Loading '.'...
Added 11053 variables.

/home/bhipple/src/nixpkgs/pkgs/development/libraries/boost/generic.nix
Loading '.'...
Added 11053 variables.
```
That can assert that it uses `$EDITOR`, finds the package, reloads `nix repl`, etc.

src/nix/edit.cc
```diff
@@ -36,45 +36,17 @@ struct CmdEdit : InstallableCommand
...
 
         stopProgressBar();
 
+        auto args = editorFor(pos);
+
         execvp(args.front().c_str(), stringsToCharPtrs(args).data());
 
-        throw SysError("cannot run editor '%s'", editor);
+        std::string command;
+        for (const auto &arg : args) command += " '" + arg + "'";
```
bhipple: Note a huge deal, but it'd be more idiomatic (and performant) to use a `std::stringstream` here.

src/nix/repl.cc
```diff
@@ -440,6 +441,7 @@ bool NixRepl::processLine(string line)
              << "  <x> = <expr>  Bind expression to variable\n"
              << "  :a <expr>     Add attributes from resulting set to scope\n"
              << "  :b <expr>     Build derivation\n"
+             << "  :e <expr>     Open the derivation in $EDITOR\n"
```
bhipple: Unrelated to this PR, but it probably makes sense to curate an (unreleased) `doc/manual/release-notes/rl-2.4.xml` that gets updated with upcoming changes as we add them.  That way @edolstra doesn't have to go through the git log and write them all himself when he wants to do a release, and users can see what unreleased changes have accumulated on master.

As an extra aside, it looks like we don't have man pages or manual documentation for the nix 2.0 CLI, so I guess this is the only place it needs to be documented?

PR #3166, opened 2019-10-25
Add fetchSubmodules to builtins.fetchGit

### LINKED ISSUE #2151: builtins.fetchGit doesn't fetch submodules

blitz:
This pull request adds the ability to fetch submodules with `builtins.fetchGit`. Use it like: `builtins.fetchGit { fetchSubmodules = true; ... }`

There are some downsides to this feature:

 - Submodules are not cached (unlike the root repo),
 - Full checkouts are created in a temporary directory.

These could be fixed, but require far more code and would require re-implementing some of the submodule cloning feature. This PR does not change any of the `fetchGit` behavior when submodules are not used (the default), so only people using this feature are paying these costs.

Other missing things:
- [x] Tests
- [x] Documentation

**Note:** This is my first contribution to Nix, so please expect stupid mistakes. :)

Fixes #2151 

@edolstra @tfc

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -422,6 +422,16 @@ stdenv.mkDerivation { … }
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>fetchSubmodules</term>
+          <listitem>
+            <para>
+              A boolean parameter that specifies whether submodules
+              should be checked out. Defaults to
+              <literal>false</literal>.
+            </para>
+          </listitem>
+        </varlistentry>
       </variablelist>
 
       <example>
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -22,13 +22,14 @@ struct GitInfo
     std::string rev;
     std::string shortRev;
     uint64_t revCount = 0;
+    bool submodules = false;
 };
 
 std::regex revRegex("^[0-9a-fA-F]{40}$");
 
 GitInfo exportGit(ref<Store> store, const std::string & uri,
     std::optional<std::string> ref, std::string rev,
-    const std::string & name)
+    const std::string & name, bool fetchSubmodules)
 {
     if (evalSettings.pureEval && rev == "")
         throw Error("in pure evaluation mode, 'fetchGit' requires a Git revision");
@@ -145,7 +146,8 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
 
     printTalkative("using revision %s of repo '%s'", gitInfo.rev, uri);
 
-    std::string storeLinkName = hashString(htSHA512, name + std::string("\0"s) + gitInfo.rev).to_string(Base32, false);
+    std::string storeLinkName = hashString(htSHA512, name + std::string("\0"s) + gitInfo.rev
+        + (fetchSubmodules ? "submodules" : "")).to_string(Base32, false);
     Path storeLink = cacheDir + "/" + storeLinkName + ".link";
     PathLocks storeLinkLock({storeLink}, fmt("waiting for lock on '%1%'...", storeLink)); // FIXME: broken
 
@@ -165,16 +167,35 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
         if (e.errNo != ENOENT) throw;
     }
 
-    auto source = sinkToSource([&](Sink & sink) {
-        RunOptions gitOptions("git", { "-C", cacheDir, "archive", gitInfo.rev });
-        gitOptions.standardOut = &sink;
-        runProgram2(gitOptions);
-    });
-
     Path tmpDir = createTempDir();
     AutoDelete delTmpDir(tmpDir, true);
 
-    unpackTarfile(*source, tmpDir);
+    // Submodule support can be improved by adding caching to the submodules themselves. At the moment, only the root
+    // repo is cached.
+    if (fetchSubmodules) {
+        Path tmpGitDir = createTempDir();
+        AutoDelete delTmpGitDir(tmpGitDir, true);
+
+        runProgram("git", true, { "init", tmpDir, "--separate-git-dir", tmpGitDir });
+        runProgram("git", true, { "-C", tmpDir, "fetch", "--quiet", "--force",
+                                  "--", cacheDir, fmt("%s:%s", *ref, *ref) });
+
+        runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
+        runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
+        runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init" });
+
+        deletePath(tmpDir + "/.git");
+
+        gitInfo.submodules = true;
+    } else {
+        auto source = sinkToSource([&](Sink & sink) {
+            RunOptions gitOptions("git", { "-C", cacheDir, "archive", gitInfo.rev });
+            gitOptions.standardOut = &sink;
+            runProgram2(gitOptions);
+        });
+
+        unpackTarfile(*source, tmpDir);
+    }
 
     gitInfo.storePath = store->printStorePath(store->addToStore(name, tmpDir));
 
@@ -198,6 +219,7 @@ static void prim_fetchGit(EvalState & state, const Pos & pos, Value * * args, Va
     std::optional<std::string> ref;
     std::string rev;
     std::string name = "source";
+    bool fetchSubmodules = false;
     PathSet context;
 
     state.forceValue(*args[0]);
@@ -216,6 +238,8 @@ static void prim_fetchGit(EvalState & state, const Pos & pos, Value * * args, Va
                 rev = state.forceStringNoCtx(*attr.value, *attr.pos);
             else if (n == "name")
                 name = state.forceStringNoCtx(*attr.value, *attr.pos);
+            else if (n == "fetchSubmodules")
+                fetchSubmodules = state.forceBool(*attr.value, *attr.pos);
             else
                 throw EvalError("unsupported argument '%s' to 'fetchGit', at %s", attr.name, *attr.pos);
         }
@@ -230,13 +254,14 @@ static void prim_fetchGit(EvalState & state, const Pos & pos, Value * * args, Va
     // whitelist. Ah well.
     state.checkURI(url);
 
-    auto gitInfo = exportGit(state.store, url, ref, rev, name);
+    auto gitInfo = exportGit(state.store, url, ref, rev, name, fetchSubmodules);
 
     state.mkAttrs(v, 8);
     mkString(*state.allocAttr(v, state.sOutPath), gitInfo.storePath, PathSet({gitInfo.storePath}));
     mkString(*state.allocAttr(v, state.symbols.create("rev")), gitInfo.rev);
     mkString(*state.allocAttr(v, state.symbols.create("shortRev")), gitInfo.shortRev);
     mkInt(*state.allocAttr(v, state.symbols.create("revCount")), gitInfo.revCount);
+    mkBool(*state.allocAttr(v, state.symbols.create("submodules")), gitInfo.submodules);
     v.attrs->sort();
 
     if (state.allowedPaths)
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -182,7 +182,7 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
         runProgram("git", true, { "-C", tmpDir, "fetch", "--quiet", "--force",
                                   "--", cacheDir, fmt("%s", *ref) });
 
-        runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
+        runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", gitInfo.rev });
         runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
         runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init", "--recursive" });
 
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -179,7 +179,7 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
 
         runProgram("git", true, { "init", tmpDir, "--separate-git-dir", tmpGitDir });
         runProgram("git", true, { "-C", tmpDir, "fetch", "--quiet", "--force",
-                                  "--", cacheDir, fmt("%s:%s", *ref, *ref) });
+                                  "--", cacheDir, fmt("%s", *ref) });
 
         runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
         runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -32,6 +32,9 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
     std::optional<std::string> ref, std::string rev,
     const std::string & name, bool fetchSubmodules)
 {
+    GitInfo gitInfo;
+    gitInfo.submodules = fetchSubmodules;
+
     if (evalSettings.pureEval && rev == "")
         throw Error("in pure evaluation mode, 'fetchGit' requires a Git revision");
 
@@ -49,7 +52,6 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
         if (!clean) {
 
             /* This is an unclean working tree. So copy all tracked files. */
-            GitInfo gitInfo;
             gitInfo.rev = "0000000000000000000000000000000000000000";
             gitInfo.shortRev = std::string(gitInfo.rev, 0, 7);
 
@@ -141,7 +143,6 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
     }
 
     // FIXME: check whether rev is an ancestor of ref.
-    GitInfo gitInfo;
     gitInfo.rev = rev != "" ? rev : chomp(readFile(localRefFile));
     gitInfo.shortRev = std::string(gitInfo.rev, 0, 7);
 
@@ -190,8 +191,6 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
                 std::filesystem::remove_all(p.path());
             }
         }
-
-        gitInfo.submodules = true;
     } else {
         auto source = sinkToSource([&](Sink & sink) {
             RunOptions gitOptions("git", { "-C", cacheDir, "archive", gitInfo.rev });
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -211,6 +211,7 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
     json["name"] = name;
     json["rev"] = gitInfo.rev;
     json["revCount"] = gitInfo.revCount;
+    json["submodules"] = gitInfo.submodules;
 
     writeFile(storeLink, json.dump());
 
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -6,6 +6,7 @@
 #include "hash.hh"
 #include "tarfile.hh"
 
+#include <filesystem>
 #include <sys/time.h>
 
 #include <regex>
@@ -182,9 +183,13 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
 
         runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
         runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
-        runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init" });
+        runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init", "--recursive" });
 
-        deletePath(tmpDir + "/.git");
+        for (const auto& p : std::filesystem::recursive_directory_iterator(tmpDir)) {
+            if (p.path().filename() == ".git") {
+                std::filesystem::remove_all(p.path());
+            }
+        }
 
         gitInfo.submodules = true;
     } else {
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -243,7 +243,7 @@ static void prim_fetchGit(EvalState & state, const Pos & pos, Value * * args, Va
                 rev = state.forceStringNoCtx(*attr.value, *attr.pos);
             else if (n == "name")
                 name = state.forceStringNoCtx(*attr.value, *attr.pos);
-            else if (n == "fetchSubmodules")
+            else if (n == "submodules")
                 fetchSubmodules = state.forceBool(*attr.value, *attr.pos);
             else
                 throw EvalError("unsupported argument '%s' to 'fetchGit', at %s", attr.name, *attr.pos);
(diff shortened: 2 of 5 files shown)


src/libexpr/primops/fetchGit.cc
```diff
@@ -166,14 +168,33 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
...
     AutoDelete delTmpDir(tmpDir, true);
 
-    runProgram("tar", true, { "x", "-C", tmpDir }, tar);
+    // Submodule support can be improved by adding caching to the submodules themselves. At the moment, only the root
+    // repo is cached.
+    if (fetchSubmodules) {
+        Path tmpGitDir = createTempDir();
+        AutoDelete delTmpGitDir(tmpGitDir, true);
+
+        runProgram("git", true, { "init", tmpDir, "--separate-git-dir", tmpGitDir });
```
qolii: `--separate-git-dir`! Awesome. Git has so many options...

blitz: I thought it's easier to confidently cleanup git metadata, if I put it in a different directory to begin with.

blitz:
I've added documentation and a test.

nmattia:
@edolstra WDYT? this would be super helpful!

blitz:
@edolstra @tfc I've resolved merge conflicts.

blitz:
@edolstra It would be really nice to find a solution here to get submodule support working with `builtins.fetchGit`, because I'm currently building workarounds _again_. :)

src/libexpr/primops/fetchGit.cc
```diff
@@ -165,16 +167,35 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
...
+        runProgram("git", true, { "-C", tmpDir, "fetch", "--quiet", "--force",
+                                  "--", cacheDir, fmt("%s:%s", *ref, *ref) });
+
+        runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
+        runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
+        runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init" });
```
bjornfor: IMHO, it'd be a good idea to add `--recursive` flag above.

src/libexpr/primops/fetchGit.cc
```diff
@@ -165,16 +167,35 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
...
+
+        deletePath(tmpDir + "/.git");
```
bjornfor: Submodules also leave .git _files_ in the tree, looking like this (extracted from the end of a run of ./tests/fetchGitSubmodules.sh):

```
gitdir: ../../nix-26667-1/modules/sub
```

I think the best way to handle that is to recursively delete `.git` directories and files.

tests/fetchGitSubmodules.sh
```diff
@@ -0,0 +1,50 @@
...
+pathWithoutSubmodules=$(nix eval --raw "(builtins.fetchGit { url = file://$rootRepo; rev = \"$rev\"; }).outPath")
+pathWithSubmodules=$(nix eval --raw "(builtins.fetchGit { url = file://$rootRepo; rev = \"$rev\"; fetchSubmodules = true; }).outPath")
+
+# The resulting store path cannot be the same.
+[[ $pathWithoutSubmodules != $pathWithSubmodules ]]
+
+[[ ! -e $pathWithoutSubmodules/sub/content ]]
+[[ -e $pathWithSubmodules/sub/content ]]
+
+
```
bjornfor: I suggest to include `test $(find "$pathWithSubmodules" -name .git) == ""` here at the end, to catch impure .git submodule reference files.

bjornfor: Alternatively, clone the repo twice from different locations and verify that the repo content hash stays the same. (Currently it doesn't.)

bjornfor:
Here is a patch with my proposed changes:

```patch
diff --git a/src/libexpr/local.mk b/src/libexpr/local.mk
index 26b9f14ba..e8c019365 100644
--- a/src/libexpr/local.mk
+++ b/src/libexpr/local.mk
@@ -8,7 +8,7 @@ libexpr_SOURCES := $(wildcard $(d)/*.cc) $(wildcard $(d)/primops/*.cc) $(d)/lexe
 
 libexpr_LIBS = libutil libstore libnixrust
 
-libexpr_LDFLAGS =
+libexpr_LDFLAGS = -lstdc++fs
 ifneq ($(OS), FreeBSD)
  libexpr_LDFLAGS += -ldl
 endif
diff --git a/src/libexpr/primops/fetchGit.cc b/src/libexpr/primops/fetchGit.cc
index b08e48404..c58d28bbc 100644
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -6,6 +6,7 @@
 #include "hash.hh"
 #include "tarfile.hh"
 
+#include <filesystem>
 #include <sys/time.h>
 
 #include <regex>
@@ -182,9 +183,13 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
 
         runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
         runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
-        runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init" });
+        runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init", "--recursive" });
 
-        deletePath(tmpDir + "/.git");
+        for (const auto& p : std::filesystem::recursive_directory_iterator(tmpDir)) {
+            if (p.path().filename() == ".git") {
+                std::filesystem::remove_all(p.path());
+            }
+        }
 
         gitInfo.submodules = true;
     } else {
diff --git a/tests/fetchGitSubmodules.sh b/tests/fetchGitSubmodules.sh
index c821d2caa..987bfd69e 100644
--- a/tests/fetchGitSubmodules.sh
+++ b/tests/fetchGitSubmodules.sh
@@ -47,4 +47,5 @@ pathWithSubmodules=$(nix eval --raw "(builtins.fetchGit { url = file://$rootRepo
 [[ ! -e $pathWithoutSubmodules/sub/content ]]
 [[ -e $pathWithSubmodules/sub/content ]]
 
-
+# No .git directory or submodule reference files must be left
+test "$(find "$pathWithSubmodules" -name .git)" = ""
```

I used std::filesystem for removing the .git paths. Nix already depends on C++17, but apparently one needs -lstdc++fs to get this library. The internet says clang needs a different flag, but the current flag worked in my testing (`./shell.nix` has a `useClang` flag which I enabled). So perhaps I was reading outdated info.

nmattia:
By the way, I had a chat with @edolstra . He's refactoring the fetcher, which will need to land before this. Sharing this to clarify, and ideally to save people some time in case some of the changes don't apply anymore!

blitz:
@bjornfor Thanks for the review and fixes! As @nmattia pointed out this will likely conflict with upcoming changes, so I'd rather wait before spending more time here. Or do you think there is a chance of sneaking this in before the refactoring?

@nmattia @edolstra It would be nice if there is some issue to watch regarding the fetcher changes.

bjornfor:
I'm eagerly awaiting a status update here :-)

bjornfor:
I did some real life testing with this patch and found two issues.

1. The result of `builtins.fetchGit { fetchSubmodules = true; ... }` changes from the first run to the next. The first run results in `{ outPath = ; ...; submodules = true; }`, but the next calls result in `{ outPath = ; ...; submodules = false; }`. I think the output submodules attribute would follow the input fetchSubmodules parameter exactly and not change on the 2nd (and following) runs.

2. Using `fetchSubmodules = true` on repos that have no submodules is a fatal error:
```
nix --experimental-features nix-command eval '(builtins.fetchGit { url = "https://github.com/nixos/nix.git"; ref = "refs/heads/master"; fetchSubmodules = true; } )'
fatal: Refusing to fetch into current branch refs/heads/master of non-bare repository
fatal: The remote end hung up unexpectedly
```

Allowing `fetchSubmodules = true` on non-submodule using repos is a requirement for my use case, and is supported by git itself: `git clone --recursive https://github.com/nixos/nix.git` is no error even if nix.git uses no submodules.

@edolstra: Are you refactoring the fetcher(s)? I would _really_ like for nix to have git submodule support and can probably spend a bit more time on this if there is interest. But no sign of interest kills my motivation.

edolstra:
Yes, the new fetcher infrastructure is in the flakes branch: https://github.com/NixOS/nix/tree/flakes/src/libstore/fetchers. It replaces `fetchGit` / `fetchMercurial` with a generic `fetchTree` builtin that uses the same attributes as flake inputs (e.g. `fetchTree { type = "github"; owner = "NixOS"; repo = "nix"; rev = ...; }`). So `submodules` would be an attribute supported by the `git` fetcher. And then flakes would automatically support Git repos with submodules as an input.

src/libexpr/primops/fetchGit.cc
edolstra: Does that mean that the tree in the Nix store contains `.git` directories? If so it's important to remove them, since we must have a canonical representation for reproducibility.

src/libexpr/primops/fetchGit.cc
```diff
@@ -216,6 +238,8 @@ static void prim_fetchGit(EvalState & state, const Pos & pos, Value * * args, Va
                 rev = state.forceStringNoCtx(*attr.value, *attr.pos);
             else if (n == "name")
                 name = state.forceStringNoCtx(*attr.value, *attr.pos);
+            else if (n == "fetchSubmodules")
```
edolstra: Maybe rename this to `submodules`? "fetch" is superfluous since the function is already called `fetchGit`.

blitz: Sounds good.

tests/fetchGitSubmodules.sh
blitz: Sounds like we need a test for this.

src/libexpr/primops/fetchGit.cc
bjornfor: I agree. Yes, the PR as-is has .git files in the nix store. My follow up patch [here](https://github.com/NixOS/nix/pull/3166#issuecomment-573544529) fixes that.

tests/fetchGitSubmodules.sh
bjornfor: My follow-up patch [here](https://github.com/NixOS/nix/pull/3166#issuecomment-573544529) implements the first alternative, and it does catch the issue.

blitz:
@edolstra @tfc @bjornfor  I was wondering how to implement submodule support while keeping all the nice caching in place. I'm thinking about something along the lines of:

1. Fetch a git repo
2. Export it into the Nix store
3. If submodule fetching is enabled:
  - enumerate all submodules, their URLs and commit hashes,
 - recursively use the normal git fetching logic to fetch them into the Nix store
 - join all submodules together using hard/symlinks

This would keep the normal caching of git repositories in place and submodules wouldn't be treated any differently from normal git checkouts.

The downside is that the implementation is somewhat complex compared to just using `git clone --recursive` and we are partially reimplementing the git submodule command.

To limit the complexity in Nix itself, most of this could be implemented as Nix expression, given that a builtin is available that returns a list of submodules and their properties.

Opinions?

edolstra:
I think we should ignore the caching issue for now. I still want to add a generic caching mechanism to the fetcher infrastructure that generalizes the Git-specific caching (in `~/.cache/nix/git-revs`). So Nix would keep a mapping `fetcher input attributes -> store path` which would also work for git-with-submodules.

bjornfor:
> Yes, the new fetcher infrastructure is in the flakes branch: https://github.com/NixOS/nix/tree/flakes/src/libstore/fetchers.

@blitz @edolstra: I'm confused about what the next action is. Ok, so flakes branch has new fetcher infrastructure, and this is where the future git fetcher with submodule support will live. But flakes is an experimental branch and it would be nice to have submodules support in _standard_ Nix. (I don't know how long until flakes get stabilized and land in master.)

So, if this PR is refreshed (fixing the outstanding issues as commented in this thread), would it be merged?

blitz:
@bjornfor I'll post an updated version soon. If I should rebuild this on top of flakes, this would be good to know.

edolstra:
@bjornfor Yes, I'll take care of porting it to the new fetcher infrastructure.

blitz:
I've incorporated all fixes. I just have to investigate the test failures.

src/libexpr/primops/fetchGit.cc
blitz: Done.

tests/fetchGitSubmodules.sh
blitz: I've added your changes.

blitz:
@edolstra @bjornfor @nmattia  @tfc I'd appreciate a second look here. I think I have added tests for all issues that were pointed out and these tests are passing now.

Two tests are failing (`tests/multiple-outputs.sh` and `tests/nix-build.sh), but these are not regressions and are already failing on the master branch.

bjornfor:
@blitz: Did you build it? I'm getting this error:

```
/nix/store/m8ih8gw7qnd2c7p0qgv3cn4b6abkbanb-binutils-2.31.1/bin/ld: /build/nix-2.4pre7279_7fee022/src/libexpr/primops/fetchGit.cc:188: undefined reference to `std::filesystem::__cxx11::recursive_directory_iterator::operator++()'
```

In my patch I added `libexpr_LDFLAGS = -lstdc++fs`, but I don't see that in your PR.

blitz:
@bjornfor 

> @blitz: Did you build it? I'm getting this error:
> 
> ```
> /nix/store/m8ih8gw7qnd2c7p0qgv3cn4b6abkbanb-binutils-2.31.1/bin/ld: /build/nix-2.4pre7279_7fee022/src/libexpr/primops/fetchGit.cc:188: undefined reference to `std::filesystem::__cxx11::recursive_directory_iterator::operator++()'
> ```
> 
> In my patch I added `libexpr_LDFLAGS = -lstdc++fs`, but I don't see that in your PR.

My understanding was that `-lstdc++fs` is only required for the experimental filesystem support and is not required anymore since the filesystem support moved out of its experimental state. That may be wrong for some platforms. I'm using:

```
% nix-build release.nix -A build.x86_64-linux
```

blitz:
@bjornfor I've added `-lstdc++fs` again. It will still be good to know why this makes a difference for you.

edolstra:
Maybe use the `deletePath()` function instead?

src/libexpr/primops/fetchGit.cc
```diff
@@ -165,16 +168,39 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
...
+
+        runProgram("git", true, { "init", tmpDir, "--separate-git-dir", tmpGitDir });
+        runProgram("git", true, { "-C", tmpDir, "fetch", "--quiet", "--force",
+                                  "--", cacheDir, fmt("%s:%s", *ref, *ref) });
+
+        runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
+        runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
+        runProgram("git", true, { "-C", tmpDir, "submodule", "--quiet", "update", "--init", "--recursive" });
+
+        for (const auto& p : std::filesystem::recursive_directory_iterator(tmpDir)) {
```
edolstra: This can also be done using `readDirectory()` and `deletePath()`.

bjornfor:
> @bjornfor I've added `-lstdc++fs` again. It will still be good to know why this makes a difference for you.

Turns out I was building nix with nixpkgs @ 20.03pre207249.7e8454fb856. Using latest unstable (`20.09pre216190.6b6f9d769a5`) works without `-lstdc++fs`.

Now getting blocked by the test failures on nix master branch.

bjornfor:
> Two tests are failing (`tests/multiple-outputs.sh` and `tests/nix-build.sh), but these are not regressions and are already failing on the master branch.

@blitz: I just learned the broken tests were fixed 4th of march: https://github.com/NixOS/nix/commit/d37dc71e3cf077fa5d24a9bf8395deae21cc4410. Can you update the PR (rebase or merge master)?

IMHO, it'd be nice to keep `-lstdc++fs` around for a little while, to support building with not-bleeding-edge stdenv. Eelco suggest to use readDirectory() and deletePath(), not sure that's a hard requirement or not, but I think there is one issue left that is more important to fix first, and I'd like to get this PR building soon so I can debug it.

doc/manual/expressions/builtins.xml
```diff
@@ -422,6 +422,16 @@ stdenv.mkDerivation { … }
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>fetchSubmodules</term>
```
bjornfor: fetchSubmodules was renamed to submodules.

tests/fetchGitSubmodules.sh
```diff
@@ -0,0 +1,63 @@
...
+git -C $rootRepo submodule init
+git -C $rootRepo submodule add $subRepo sub
+git -C $rootRepo add sub
+git -C $rootRepo commit -m "Add submodule"
+
+rev=$(git -C $rootRepo rev-parse HEAD)
+
+pathWithoutSubmodules=$(nix eval --raw "(builtins.fetchGit { url = file://$rootRepo; rev = \"$rev\"; }).outPath")
+pathWithSubmodules=$(nix eval --raw "(builtins.fetchGit { url = file://$rootRepo; rev = \"$rev\"; submodules = true; }).outPath")
```
bjornfor: The test blows up if passing `ref = "refs/heads/master"` to fetchGit with `submodules = true`:

```
++ nix eval --raw '(builtins.fetchGit { url = file:///run/user/1000/nix-test/gitSubmodulesRoot; ref = "refs/heads/master"; rev = "96b77ed66ad52090eebe12ff631e79d3c30d6ba1"; submodules = true; }).outPath'
fatal: couldn't find remote ref refs/heads/master
fatal: the remote end hung up unexpectedly
```

If using `url = /path/to/repo` (instead of `url = file:///path/to/repo`) the error becomes what I see in real life:

```
++ nix eval --raw '(builtins.fetchGit { url = /run/user/1000/nix-test/gitSubmodulesRoot; ref = "refs/heads/master"; rev = "90ccb6a7d2a59893f0c7283dee98396d3149ea62"; submodules = true; }).outPath'
fatal: Refusing to fetch into current branch refs/heads/master of non-bare repository
fatal: the remote end hung up unexpectedly
``` 

I wonder why `file://...` makes a difference. But anyway, I experience the error also with git ssh url, so it has nothing to do with local file access.

bjornfor: I think this fixes it:

```patch
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -179,7 +179,7 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
 
         runProgram("git", true, { "init", tmpDir, "--separate-git-dir", tmpGitDir });
         runProgram("git", true, { "-C", tmpDir, "fetch", "--quiet", "--force",
-                                  "--", cacheDir, fmt("%s:%s", *ref, *ref) });
+                                  "--", cacheDir, fmt("%s", *ref) });
 
         runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
         runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
```

bjornfor:
@blitz: One more issue, the result of fetchGit has the "submodules" attribute set to "false" even when the input is "submodules = true":

```
[nix-shell:~/f/nix]$ ./inst/bin/nix --experimental-features nix-command eval '(builtins.fetchGit { url = ./.; submodules = true; })'
{ outPath = "/nix/store/c37mybzmffksp129bmav6918w6bc1jpq-source"; rev = "0000000000000000000000000000000000000000"; revCount = 0; shortRev = "0000000"; submodules = false; }
```

Just looking at the C++ code, I don't understand how that's possible. (The code looks good to me.)

bjornfor:
Actually, the first time something is fetched, the result of fetchGit has correct `submodules = true`, but on the second evaluation, it becomes false:

```
nix-repl> builtins.fetchGit { url = "https://github.com/NixOS/nix.git"; ref = "refs/heads/master"; submodules = true; }                                                                   
{ outPath = "/nix/store/6x8a7rqi5cgxg2xp11xgsfcldr3clzdl-source"; rev = "b79b81dd2d62d2363105c5c7a1d91be2df338ddf"; revCount = 7304; shortRev = "b79b81d"; submodules = true; }

nix-repl> builtins.fetchGit { url = "https://github.com/NixOS/nix.git"; ref = "refs/heads/master"; submodules = true; }
{ outPath = "/nix/store/6x8a7rqi5cgxg2xp11xgsfcldr3clzdl-source"; rev = "b79b81dd2d62d2363105c5c7a1d91be2df338ddf"; revCount = 7304; shortRev = "b79b81d"; submodules = false; }
```

bjornfor:
I have fixed it here: https://github.com/bjornfor/nix/tree/fetchgit-with-fetchsubmodules-v2, specifically commit https://github.com/bjornfor/nix/commit/cac9dbe428ebcbcc70dc9021ea58ebddcff46268. I also added https://github.com/bjornfor/nix/commit/90aea848af4112ce7832dcd0bbee0564c66183f3 which I think is a good idea.

src/libexpr/primops/fetchGit.cc
```diff
@@ -165,16 +168,39 @@ GitInfo exportGit(ref<Store> store, const std::string & uri,
...
+    if (fetchSubmodules) {
+        Path tmpGitDir = createTempDir();
+        AutoDelete delTmpGitDir(tmpGitDir, true);
+
+        runProgram("git", true, { "init", tmpDir, "--separate-git-dir", tmpGitDir });
+        runProgram("git", true, { "-C", tmpDir, "fetch", "--quiet", "--force",
+                                  "--", cacheDir, fmt("%s:%s", *ref, *ref) });
+
+        runProgram("git", true, { "-C", tmpDir, "checkout", "--quiet", "FETCH_HEAD" });
+        runProgram("git", true, { "-C", tmpDir, "remote", "add", "origin", uri });
```
bjornfor: Why add a remote? Is it to resolve submodule urls that are relative paths? Add a comment?

bjornfor:
@blitz: Important bugfixes: https://github.com/bjornfor/nix/commit/58a601b0a8e2afacbd875180320b745614d7b0ab and https://github.com/bjornfor/nix/commit/cd20da51adffea683709e89b4d153e60ecdbc668.

bjornfor:
Another fix here: https://github.com/bjornfor/nix/commit/2c2dbb9e163dc72d4675150d6f515d4f643be376

bjornfor:
I found a bug in my latest bugfix... here is the fix: https://github.com/bjornfor/nix/commit/503feb93df4416cb7070b58e15c19b4e1e1b4b63.

blitz:
@bjornfor Many thanks for all the fixes! I've incorporated them into the PR (with some light squashing). Regarding https://github.com/NixOS/nix/commit/2c2dbb9e163dc72d4675150d6f515d4f643be376, do you have an idea how to test this?

blitz:
@bjornfor I've given the current state a shot on our internal project and it works. The only thing that remains is that I get this warning:

```
warning: refname 'HEAD' is ambiguous.
``

bjornfor:
Re. testing 2c2dbb9, I was thinking we could create the test repo with two branches pointing to the same commit, then `fetchGit { ref = "ref-1"; rev = "the-rev"; submodules = false; }` once to get the rev in place, then `fetchGit { ref = "ref-2"; rev = "the-rev"; submodules = true; }` and see that it still works.

We should also have a test that we fetch the rev (not HEAD) in the submodules = true case. (That issue confused me a lot when the submodule repo silently fetched the HEAD of the cache repo instead of the rev!)

When does `warning: refname 'HEAD' is ambiguous` happen?

blitz:
> When does `warning: refname 'HEAD' is ambiguous` happen?

I haven't pinpointed it yet.

blitz:
@bjornfor The `std::filesystem` code is still not happy on Mac. `-lstdc++fs` is not available to link with and without it there are symbols missing. Let me see whether I find a Mac user to debug this.

edolstra:
Just avoid `std::filesystem` for now. As I commented, we already have functions that do the same thing (e.g. `readDirectory()` and `deletePath()`).

blitz:
> Just avoid `std::filesystem` for now. As I commented, we already have functions that do the same thing (e.g. `readDirectory()` and `deletePath()`).

Will do. It's somewhat unfortunate, because `std::filesystem` takes care of recursion very elegantly.

blitz:
@edolstra @bjornfor 

I've removed `std::filesystem` use in favor of using the filter support in `addToStore`. This seemed simpler than using `readDirectory` and `deletePath`.

tfc:
@edolstra @bjornfor anything else that needs to be done? i am very interested in seeing this patch merged.

edolstra:
Merged, thanks!

PR #3169, opened 2019-10-27
nix: allow top-level installables

zimbatm:
This brings a missing feature from the v1 CLI.

Old syntax: `nix-build default.nix`

New syntax: `nix build -f default.nix .`

Currently the v2 CLI can only build installables if they are part of a returnd attribute set, like nixpkgs.

### DIFF
--- a/src/nix/build.cc
+++ b/src/nix/build.cc
@@ -45,6 +45,10 @@ struct CmdBuild : MixDryRun, InstallablesCommand
                 "To build the build.x86_64-linux attribute from release.nix:",
                 "nix build -f release.nix build.x86_64-linux"
             },
+            Example{
+                "To build a top-level derivation:"
+                "nix build -f default.nix ."
+            },
         };
     }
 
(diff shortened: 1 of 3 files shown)


edolstra:
`nix build -f . ''` would also work right?

zimbatm:
Yes technically :)

Or what do you think of making the top-level installable as the default if `-f` is being used?

zimbatm:
The issue right now is that the capability is not discoverable, so my main goal is to improve on that.

Most times where `-f` is being used, it makes sense to default to building the top-level. It's a different context than building nixpkgs.

Another solution is to just document it properly, with examples added to the CLI.

edolstra:
Hm, turns out that `nix build -f default.nix` (without any `.` or empty argument) already works. (Or equivalently, `nix build -f .`). Isn't that what you want?

zimbatm:
How embarrassing. I remember now that this used to be the cased and then you fixed it.

`nix eval -f .` hasn't been fixed yet though.

edolstra:
`nix eval -f .` now also works (ca8caaec5e7581c37b46f79622c81adf52f06314).

Issue #1610, opened 2017-10-17
Feature request: Print backtrace in `trace` output


bgamari:
The `builtins.trace` function is quite handy for probing state buried deep within a derivation. However, while trying to understand a cross-compiling bug I've seen my trace output printed multiple times. Without a stack backtrace (e.g. what `throw` would produce with `--show-trace`) this is quite difficult to interpret.

copumpkin:
This seems cool and possibly not too hard to add. I'd probably make a new builtin for it though, like `builtins.dumpStack` which behaves like the identity function except dumps a stack trace when forced. I don't necessarily want a stack dump in every trace output, and we can't really change the API much. What you're asking then would be a composition of `builtins.trace` and `builtins.dumpStack`

milahu:
push : )

milahu:
> I'd probably make a new builtin for it though, like `builtins.dumpStack` which behaves like the identity function except dumps a stack trace when forced.

plus: something like `builtins.callStack` to get a list of the current callstack,
which can be manually reduced to, for example, the last 5 entries, to avoid printing "too long" stacktraces

9999years:
Got most of the way through filing a new issue before noticing this. Here's my experience and my idea for a design:

**Is your feature request related to a problem? Please describe.**

Right now, my project is showing some warnings:

```
$ nix develop
trace: warning: `pythonForBuild` (from `python*`) has been renamed to `pythonOnBuildForHost`
trace: warning: `pythonForBuild` (from `python*`) has been renamed to `pythonOnBuildForHost`
trace: warning: `pythonForBuild` (from `python*`) has been renamed to `pythonOnBuildForHost`
```

So I run `rg pythonForBuild`, but it doesn't turn up any results. The code that uses `pythonForBuild` must be in some other repository we're importing. When Nix is evaluating the code that produces this warning, it knows what the call stack is, but it's not exposed to the user.

**Describe the solution you'd like**

It would be nice if we could have an option to make `builtins.trace` show its call location, like this:

```
$ nix develop --builtins-trace-location
trace: warning: `pythonForBuild` (from `python*`) has been renamed to `pythonOnBuildForHost`
         at /nix/store/qccbdqrhwaaic3kwv9d4zak9fqh7yhlw-source/pkgs/development/interpreters/python/mk-python-derivation.nix:130:22:

          129|   validatePythonMatches = attrName: let
          130|     isPythonModule = drv:
             |                      ^
[...]

PR #3170, opened 2019-10-28
libexpr: add position to builtins.trace

zimbatm:
Let the user know where the trace was produced by adding the filename
and lineno from where that function was called.

/cc happy birthday @domenkozar :p

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -498,9 +498,9 @@ static void prim_trace(EvalState & state, const Pos & pos, Value * * args, Value
 {
     state.forceValue(*args[0]);
     if (args[0]->type == tString)
-        printError(format("trace: %1%") % args[0]->string.s);
+        printError(format("trace: %1% at %2%") % args[0]->string.s % pos);
     else
-        printError(format("trace: %1%") % *args[0]);
+        printError(format("trace: %1% at %2%") % *args[0] % pos);
     state.forceValue(*args[1]);
     v = *args[1];
 }


edolstra:
Don't think this is a good idea because it changes the meaning of the trace message, possibly in a confusing way. E.g.
```
trace: warning: types.string is deprecated because it quietly concatenates strings
```
becomes
```
trace: warning: types.string is deprecated because it quietly concatenates strings at /some/file/name.nix:1234
```
but that's not where `types.string` is called. Or what about
```
trace: warning: `stdenv.isArm` is deprecated after 18.03. Please use `stdenv.isAarch32` instead at /some/file/name.nix:1234
```
which tells you to use something in the wrong file.

domenkozar:
What about `trace:/some/file/name.nix:1234: warning: types.string is deprecated because it quietly concatenates strings`

Ideally, there would be an optional way to print the traceback for traces, but at least this makes tracing more useful.

zimbatm:
Fundamentally the issue with `builtins.trace` not reporting the location is a valid usability issue. Currently the way to find the call location can be done but is cumbersome as it involves using `grep` on the nix code.

There are two issues with the current implementation:
* It changes the output. Some tools (like lorri) have started depending on the output.
* The reported position is not always useful as it reports calls to `builtins.trace`, which might be wrapped in a function, like in `lib.traceVal`

What if there was a new trace builtin to which the caller position could be passed?

domenkozar:
@edolstra what if `--show-trace` would show full stack of `trace`? 

We are importing like 10 different github repositories with Nix, and going around each and grep for usage is quite cumbersome.

zimbatm:
I don't know if nix can produce a readable stacktrace in the imperative sense. When the thunk gets evaluated it already lost most of its context.

edolstra:
@domenkozar What do you mean with "full stack of trace"?

domenkozar:
Seems like I'm asking for https://github.com/NixOS/nix/issues/1610

zimbatm:
Closing since a lot of improvements have been added to Nix's error reporting since then.

Issue #2944, opened 2019-06-13
builtins.gitLsFiles


nomeata:
I started to use this file here:
```
# The function call
#
#   gitSource ./toplevel subpath
#
# creates a Nix store path of ./toplevel/subpath that includes only those files
# tracked by git. More precisely: mentioned in the git index (i.e. git add is enough
# to get them to be included, you do not have to commit).
#
# This is a whitelist-based alternative to manually listing files or using
# nix-gitignore.

# Internally, it works by calling git ls-files at evaluation time. To
# avoid copying all of `.git` to the git store, it only copies the least amount
# of files necessary for `git ls-files` to work; this is a bit fragile, but
# very fast.

with builtins;

# We read the git index once, before getting the subdir parameter, so that it
# is shared among multiple invocations of gitSource:

let
  filter_from_list = root: files:
    let
      all_paren_dirs = p:
        if p == "." || p == "/"
        then []
        else [ p ] ++ all_paren_dirs (dirOf p);

      whitelist_set = listToAttrs (
        concatMap (p:
          let full_path = toString (root + "/${p}"); in
          map (p': { name = p'; value = true; }) (all_paren_dirs full_path)
        ) files
      );
    in
    p: t: hasAttr (toString p) whitelist_set;

  has_prefix = prefix: s:
    prefix == builtins.substring 0 (builtins.stringLength prefix) s;
  remove_prefix = prefix: s:
    builtins.substring
      (builtins.stringLength prefix)
      (builtins.stringLength s - builtins.stringLength prefix)
      s;

  lines = s: filter (x : x != [] && x != "") (split "\n" s);
in

if builtins.pathExists ../.git
then
  let
    nixpkgs = (import ./nixpkgs.nix).nixpkgs {};

    git_dir =
      if builtins.pathExists ../.git/index
      then ../.git
      else # likely a git worktree, so follow the indirection
        let
          git_content = lines (readFile ./../.git);
          first_line = head git_content;
          prefix = "gitdir: ";
          ok = length git_content == 1 && has_prefix prefix first_line;
        in
          if ok
          then /. + remove_prefix prefix first_line
          else abort "gitSource.nix: Cannot parse ${toString ./../.git}";

    whitelist_file =
      nixpkgs.runCommand "git-ls-files" {envVariable = true;} ''
        cp ${git_dir + "/index"} index
        echo "ref: refs/heads/master" > HEAD
        mkdir objects refs
[...]

PR #3171, opened 2019-10-28
add a filter function argument to builtins.fetchGit

basvandijk:
Fixes: https://github.com/NixOS/nix/issues/2944.

When working with monorepos containing projects in sub directories
each sub project typically has a derivation where the `src` attribute
is set to the path of the sub directory.

That setup is not ideal since all files below that sub directory,
including build artefacts, temporary files or even secrets, will be
copied to the store. Ideally only files added to the git repository
should be copied to the store to ensure local nix builds are
reproducible on CI.

Note that `builtins.fetchGit` can be used to copy a local git
repository to the store. This commit extends `builtins.fetchGit` with
a filter function argument which can be used to only include a sub
directory.

The filter function has the same type as expected by
`builtins.filterSource` and has the same semantics except that the
path is relative to the git checkout.

@edolstra at NixCon you mentioned we should possibly think about a more compositional API since most builtin fetch functions now take a filter argument. However this addition provides in a common need and prevents users from IFD. Let me know what you think.

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -385,6 +385,16 @@ stdenv.mkDerivation { … }
             <para>
               The URL of the repo.
             </para>
+            <para>
+              Note that the <literal>url</literal> attribute can also be a path to a
+              local git repository in which case the files in the git index
+              (files which are either committed or staged) will be copied to the
+              store.
+            </para>
+            <para>
+              Note that only files which satisfy the optional
+              <literal>filter</literal> function are copied to the store.
+            </para>
           </listitem>
         </varlistentry>
         <varlistentry>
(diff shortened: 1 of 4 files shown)


roberth:
@Ericson2314 I see you've reopened this PR, but we should consider the other implementation strategy, which is to compose a filtering lazy tree onto the git lazy tree. This solution has better composability and separation of concerns, with the only "downside" being that lazy trees are still in development. However, I am confident that lazy trees are a good _idea_, so considering the scarcity of maintainer time, I think we should work towards a better lazy trees implementation.

thufschmitt:
_Notes from the maintainers team meeting_

- For everything but secrets, adding a `builtins.filterSource` on top should do most of what's needed (modulo the fact that it's doing more IO)
- For secrets, could something like [one of the nix-gitignore implementations](https://github.com/hercules-ci/gitignore.nix#comparison) do the trick? It would do a git-like filtering, plus allowing to exclude arbitrary other files

Otherwise, we'd rather go the lazy-trees (#6530) route as it would provide a more generic solution to this problem.

infinisil:
> @edolstra at NixCon you mentioned we should possibly think about a more compositional API since most builtin fetch functions now take a filter argument. However this addition provides in a common need and prevents users from IFD. Let me know what you think.

I recently worked on a very compositional API that's easy and safe to use for this purpose! It's implemented as a draft PR to Nixpkgs and the interface needs some feedback to figure out whether it satisfies all necessary use cases. Please take a look, try it out and give feedback! Here's a Discourse post giving a brief tour :smiley: https://discourse.nixos.org/t/easy-source-filtering-with-file-sets/29117

roberth:
> easy and safe to use for this purpose!

Not if you're looking for a fetchGit that doesn't but certain files in the store ever. With the Nix language, you can currently only achieve this by either manually parsing the files of a local repository, or by fetching through a derivation (IFD / Read From Realisation (is that what we're going for now?)).

Anyway, most repositories don't have a performance or confidentiality problem, so do check out the link to learn how to filter better and help validate the design! [Let's go](https://discourse.nixos.org/t/easy-source-filtering-with-file-sets/29117)

Issue #3183, opened 2019-10-31
Remove nix-store --dump and nix-store --restore?


bhipple:
Currently, nix has two pairs of commands for interacting with NAR files:

- nix-store --dump  --> nix-store --restore
- nix-store --export --> nix-store --import

These differ in that the former can only take one path, and does not store the
requisites of a path, so it can't be imported into a nix store the way you'd
expect. It requires that all paths not already exist on the --restore end, while
the --import one can incrementally update paths and properly sync the sqlite db.

Perhaps we should delete `nix-store --dump` and `nix-store --restore` entirely,
since they appear to be less useful subsets of the export and import commands.
This would simplify the cmdline / API surface area for newcomers and reduce
ongoing maintenance.

Here's an example of a newcomer getting tripped up and using the "inferior"
dump/restore. I also spent a while staring at this at NixCon, and would like to
help remove it if indeed there's no legitimate use case for it.
https://github.com/NixOS/nix/issues/951#issuecomment-229190493

edolstra:
Well, the entire `nix-store` command will hopefully become deprecated in the future. The `nix` command has some subcommands for dealing with NARs (like `nix dump-path`, `nix cat-nar`, `nix ls-nar`).

But it's really the `nix-store --export / --import` commands that should be removed. They have a very wacky, non-extensible format that does not handle things like signatures and has other misfeatures (e.g. the metadata comes *after* the NAR, so on import we have to read the NAR into memory even though the store path may already be in the Nix store). So we should come up with a new format for storing closures.

edolstra:
Closing since we don't want to break compatibility.

AleXoundOS:
I couldn't unpack/extract NAR with `nix cat-nar`. Only single files from it.
But `nix-store --restore` works right.

PR #3184, opened 2019-10-31
Manual: update appendix on nar production/consumption

bhipple:
The NAR produced by `--export` seems to work juts as well as the NAR produced by
`--dump`, except that it can *also* be safely imported into an existing nix
store and update the sqlite metadata. This commit changes the manual to
reference export/import by default.

See #3183 for a potential followup pending discussion.

### DIFF
--- a/doc/manual/glossary/glossary.xml
+++ b/doc/manual/glossary/glossary.xml
@@ -186,8 +186,8 @@
   <emphasis>AR</emphasis>chive.  This is a serialisation of a path in
   the Nix store.  It can contain regular files, directories and
   symbolic links.  NARs are generated and unpacked using
-  <command>nix-store --dump</command> and <command>nix-store
-  --restore</command>.</para></glossdef>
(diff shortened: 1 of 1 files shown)


bhipple:
Best I can tell, the `--dump` NAR is just an inferior version of the `--export` NAR, since the registration files for sqlite are plain-text and don't take up much space, e.g.:

```
$ nix-store --dump /nix/store/00crivbbz9skpjrjixaf7bczj02ihfza-sqlite-3.28.0/ > dump.nar
$ nix-store --export /nix/store/00crivbbz9skpjrjixaf7bczj02ihfza-sqlite-3.28.0/ > export.nar
$ ls -lh *.nar
-rw-r--r-- 1 bhipple users 1.2M Oct 31 07:15 dump.nar
-rw-r--r-- 1 bhipple users 1.2M Oct 31 07:15 export.nar

```

zimbatm:
related: https://github.com/NixOS/nix/issues/3183#issuecomment-548367347

Issue #2634, opened 2019-01-16
Remove absolute paths from .drv ?


volth:
Nix stores not at "/nix/store" are very rare and not well supported.
But on Windows path "/nix/store" is not valid.
An easy solution is to change it to "C:/nix/store".
It does work well but prevents interop with NixOS/Linux stores.

The main scenarios of the interop of machines with different locations of nix stores are:
 * to have a single NixOps deployment with NixOS and Windows.
 * alreasy existing nix function ```runLinuxVM``` to work on Windows. Also ```runInWindowsVM``` might be created and run on Linux.
 * interop between Windows/Native and Windows/WSL: Linux and Windows containers on Windows server.
 * ```nix-copy-closure``` between Linux and Windows machines

Well, everything above is not Windows-specific
Two Linux'es, one with nix store at "/nix/store" and another at "/home/user/nix/store" would hit the same limitation of the current idea of "store .drv files is at the same location as the derivations"

So my idea is not to put "/nix/store" in the .drv files. And to use instead either the relative pathes  or ```$NIX_STORE_DIR``` or some symbol with the same meaning (as ```~``` means ```$HOME```).

For the tree of linked .drvs-files it seems ok, but the compiled derivation usually can be run only being placed on the particular path as it might have hardcoded paths. So it means there are "derivations in correct store " and "derivations in wrong store" (which must not run in place but copied to a proper machine first; akin to having ```aarch64-linux``` derivations on ```x86_64-linux``` machine), and it must be tracked somehow (in sqlite? or in extra eight field in .drv?)

volth:
@aszlig for his work on ```runInWindowsVM```

Issue #2603, opened 2018-12-27
parsing of x:/path


volth:
Currently ```x:/path``` parsed as URL with scheme "x"
I propose to stop parsing it that way (as one-letter schemes are of very little use and anyway can be expressed using quoted string literals) and reserve them for Windows absolute paths, as follow:

| OS  | ```x:/path``` | ```/path``` | ```~/path``` |
| ------------- | ------------- | ------------- | ------------- |
| current | URL | absolute path | $HOME-related |
| proposed Linux  | syntax error | absolute path | $HOME-related |
| proposed Darwin  | syntax error | absolute path | $HOME-related |
| proposed Cygwin  | absolute<br/>(```cygpath -m``` form) | absolute<br/>(```cygpath -u``` form) | $HOME-related |
| proposed Msys/MinGW  | absolute<br/>(```cygpath -m``` form) | absolute<br/>(```cygpath -u``` form) | $HOME-related |
| proposed Windows<br/>(MSVC-compiled)  | absolute | syntax error | %USERPROFILE%-related |

grahamc:
It seems to me x:/... and /... should always be valid, since you can do file existence checks but not recover from syntax errors.

volth:
Allowing ```/path``` on Windows would intoduce non-determenism and subtle bugs: WinAPI accepts them adding disk letter of the current working directory.
Also ```x:/path``` is accepted by POSIX as a valid path relative to the current working directory.

volth:
As a disadvantage, it will break ```/. + localFile``` hack is someone is still using it to copy a file to nix store on the eval phase.

volth:
Does it break new flakes syntax?

zimbatm:
So any nix code that includes an absolute path is not cross-platform anymore? It might not be a bad thing, maybe absolute paths are an anti-feature and should be avoided anyways (except absolute store paths)

volth:
There are still `<nixpkgs/somepath/filename>` form of pathes where an user is able to create own entries like  `nixpkgs` in `$NIX_PATH`

PlumpMath:
I was installing nixos at the end of last year.
While using Windows or another Linux.
the early summer months of this year
We used nixos in earnest.
The reason for the initial use is that regardless of the company's development date, the duration of the re-recovery when formatted and the integration of internal complex developments is simple, so I became a full fan.
[...]

Issue #2591, opened 2018-12-18
fmt() is GCC-only


volth:
code of ```fmt()``` is GCC-only after https://github.com/NixOS/nix/commit/2b761d5f50b7dc17dc55c31980c2253c37b3c920

```fmt("%s/%d", "ab", 42)``` is ```ab/42``` on GCC but ```42/gc``` on Microsoft C++

There is also a similar pattern with possible the same problem in https://github.com/NixOS/nix/blob/be54f4a0b6f160f8a2b0b9e9c988fb5251f70a97/src/libutil/logging.hh#L107-L112

volth:
There is also a similar pattern with possible the same problem in https://github.com/NixOS/nix/blob/be54f4a0b6f160f8a2b0b9e9c988fb5251f70a97/src/libutil/logging.hh#L107-L112

volth:
Maybe.

But it is (was) the only incompatibility with MSVC, and pinning nix to
clang or gcc would require big changes in how windows nixpkgs is
bootstrapped

PR #3185, opened 2019-10-31
WIP: Windows support  -- depends on Meson build system (#3160)

Ericson2314:
Credits to @volth and @p01arst0rm for doing the real work. I just fixed some merge conflicts, fixed linux regressions, and unified the build systems. [Will delete @volth's temporary build system once I can get things building with https://github.com/NixOS/nixpkgs/pull/72366.]

The history is....wild, but if someone want's that fixed, I don't think it's worth fixing until this is closer to being ready.


volth:
wow

volth:
My idea for the merge was to slowly converge nix-linux and nix-windows, first by small PRs https://github.com/NixOS/nix/issues/2591 https://github.com/NixOS/nix/issues/2603 https://github.com/NixOS/nix/issues/2634 then replacing `Path` with a platform-independent version(probably from `boost`), and then, voila, the diff is trivial.
But that got stuck.

Ericson2314:
@volth that sounds great. I'd add the meson change as another such prereq, but I think we should have this big monster open too as we converge so we know what we are converging and trakc progress.

Ericson2314:
> Can you summarize what this does?

That's for @volth. I hardly know!

> A quick comment after having a glance at the diff: I think for the long-term maintainability of Nix it's important not to have the code littered with `#ifndef _WIN32` / `_MSC_VER`. So we should abstract over these incompatibilities in some way.

Agreed! That relates to what @volth said too. My plan is lets keep this open as a proof of concept, but only merge once everything is better abstracted and the diff small due to newer since-merged PRs. 

Also, I think the abstractions we come up alone with will be half the benefit of this work. It should inform the architectural questions of Nix we had on the flakes RFC for example.

[Also BTW check out http://ckaestne.github.io/TypeChef/, maybe someday we can do that nixpkgs wide and CPP will be a *lot* more maintainable! That's another long-term goal of mine.]

edolstra:
BTW, on the subject of abstracting over (store) paths, the Rust branch has a `StorePath` type which is basically a tuple `(hash, name)` and is enforced to be correct by its constructor (so you can only pass "correct" store paths). It also doesn't store the `/nix/store` prefix.

volth:
Yes, c:\nix\store for a while

Abstraction is a good idea, someone may want it on d:\nix\store or
\\server\nix\store

p01arst0rm:
> Yes, c:\nix\store for a while Abstraction is a good idea, someone may want it on d:\nix\store or \\server\nix\store
> […](#)
> On 10/31/19, Eelco Dolstra ***@***.***> wrote: BTW, on the subject of abstracting over (store) paths, the Rust branch has a `StorePath` type which is basically a tuple `(hash, name)` and is enforced to be correct by its constructor (so you can only pass "correct" store paths). It also doesn't store the `/nix/store` prefix. -- You are receiving this because you were mentioned. Reply to this email directly or view it on GitHub: [#3185 (comment)](https://github.com/NixOS/nix/pull/3185#issuecomment-548449035)

surely this would be easy to do as  a cxxflag passed into the build(s)? iirc thats how its done for a bunch of stuff. that way you can determine the default location in the config options

volth:
> > Yes, c:\nix\store for a while Abstraction is a good idea, someone may want it on d:\nix\store or \server\nix\store
> > […](#)
> > On 10/31/19, Eelco Dolstra _**@**_.***> wrote: BTW, on the subject of abstracting over (store) paths, the Rust branch has a `StorePath` type which is basically a tuple `(hash, name)` and is enforced to be correct by its constructor (so you can only pass "correct" store paths). It also doesn't store the `/nix/store` prefix. -- You are receiving this because you were mentioned. Reply to this email directly or view it on GitHub: [#3185 (comment)](https://github.com/NixOS/nix/pull/3185#issuecomment-548449035)
> 
> surely this would be easy to do as a cxxflag passed into the build(s)? iirc thats how its done for a bunch of stuff. that way you can determine the default location in the config options

[It is configured in `Makefile`](https://github.com/NixOS/nix/pull/3185/files#diff-a2f2a3db29e468ef4d80b60ae2ee1b4eR5-R10), not hardcoded in C++
But that `Makefile` was made only for developing purposes, it will be replaced with meson, right?
When Nix build using Nix there is a configurable option passed down:

https://github.com/NixOS/nixpkgs/blob/fc2592b2d6348f73bfc7150f71c3df7cec0bdcd2/pkgs/tools/package-management/nix/default.nix#L2

p01arst0rm:
> > > Yes, c:\nix\store for a while Abstraction is a good idea, someone may want it on d:\nix\store or \server\nix\store
> > > […](#)
> > > On 10/31/19, Eelco Dolstra _**@**_.***> wrote: BTW, on the subject of abstracting over (store) paths, the Rust branch has a `StorePath` type which is basically a tuple `(hash, name)` and is enforced to be correct by its constructor (so you can only pass "correct" store paths). It also doesn't store the `/nix/store` prefix. -- You are receiving this because you were mentioned. Reply to this email directly or view it on GitHub: [#3185 (comment)](https://github.com/NixOS/nix/pull/3185#issuecomment-548449035)
> > 
> > 
> > surely this would be easy to do as a cxxflag passed into the build(s)? iirc thats how its done for a bunch of stuff. that way you can determine the default location in the config options
> 
> [It is configured in `Makefile`](https://github.com/NixOS/nix/pull/3185/files#diff-a2f2a3db29e468ef4d80b60ae2ee1b4eR5-R10), not hardcoded in C++
> But that `Makefile` was made only for developing purposes, it will be replaced with meson, right?
> 
> When Nix build using Nix there is a configurable option passed down:
> 
> https://github.com/NixOS/nixpkgs/blob/fc2592b2d6348f73bfc7150f71c3df7cec0bdcd2/pkgs/tools/package-management/nix/default.nix#L2-L3

how are people going to go about building meson projects in nix? is that decided yet?

volth:
> how are people going to go about building meson projects in nix? is that decided yet?

No, there are even more fundamental decisions have not made yet, for example:
* how to escalate privileges for symlink creation (default Windows user has no permissions to do it)
* which language to use for imperative code (currently it is perl with patched Win32::LongPath, which might be inappropriate for "people", but currently there is no other working solution)

Ericson2314:
> How are people going to go about building meson projects in nix? is that decided yet?

@p01arst0rm I am not sure what you are asking, but we do build plenty of software using Meson in Nixpkgs today (but on Unix).

@volth my coworker proposed Lua and I must say I like that idea a lot for a new stdenv. That all said I don't think this is the right place to discus how to make Nixpkgs work on Windows. I want to emphasize that there's still tons of value we can get from Nix on Windows before that (e.g. cross compile from linux then test on Windows in separate DRV).

conferno:
> my coworker proposed Lua and I must say I like that idea a lot for a new stdenv. That all said I don't think this is the right place to discus how to make Nixpkgs work on Windows.

While `Lua vs. other` may be offtopic here, I must explain the constraints which makes difficult using of many imperative languages originated from Unix world as `stdenv.shell`.

1. **Path is UTF-16 string** and this is difficult to avoid, because
1.1. we hit 256-byte length limit (I won't go into details now, there are some ways to workarond)
1.2. files with names which cannot be represented as UTF-8 could be on disk, after `fetchzip` or in check phase. `stdenv.shell` must be able at least `cp -r` and `rm -rf` directories which have such files inside.
Thus, internal representation of paths has to be UTF-16, so neither `bash`+`coreutils` nor stock `perl` nor stock `lua` would work out of the box :(

Ericson2314:
I'll make a separate issue for figuring out the path types, as it will be a lot of work. Just as @volth made #2591 #2603 #2634 we can have threads about each concrete issue, and this can be the big "integration" thread.

volth:
> @volth my coworker proposed Lua and I must say I like that idea a lot for a new stdenv. That all said I don't think this is the right place to discus how to make Nixpkgs work on Windows.

I just notice perl->python transition of `test-driver` which might be a good first step in unifying the imperative code in `nixpkgs`.
Although there are few little problems with `python` on Windows (it might need patches to support long paths and filenames with invalid UTF-8), it has a strong fanatic lobby among the contributors and replacing `bash` with `python` in `stdenv` might be a doable task in terms of volunteering workforce which would do the rewrite. It can hardly be done with Lua or something else (cc @domenkozar @FRidh)

p01arst0rm:
> i

+1 on python for cross platform support

p01arst0rm:
> > @volth my coworker proposed Lua and I must say I like that idea a lot for a new stdenv. That all said I don't think this is the right place to discus how to make Nixpkgs work on Windows.
> 
> I just notice perl->python transition of `test-driver` which might be a good first step in unifying the imperative code in `nixpkgs`.
> Although there are few little problems with `python` on Windows (it might need patches to support long paths and filenames with invalid UTF-8), it has a strong fanatic lobby among the contributors and replacing `bash` with `python` in `stdenv` might be a doable task in terms of volunteering workforce which would do the rewrite. It can hardly be done with Lua or something else (cc @domenkozar @FRidh)

+1 on python for cross platform support

Ericson2314:
@volth What did you get for libbacktrace with MinGW, https://github.com/cloudwu/backtrace-mingw?

volth:
> @volth What did you get for libbacktrace with MinGW

I do not remember.
Well, I remember the word `libbacktrace`, but `grep` founds nothing.
Where is it used in Nix?

Ericson2314:
`BOOST_STACKTRACE_USE_BACKTRACE` in src/libutil/util.cc. I just commented that out for now.

In other news, it builds!! Will push very shortly. https://github.com/NixOS/nixpkgs/pull/73265 is the real secret sauce.

Ericson2314:
OK build instructions

1. https://github.com/NixOS/nixpkgs/pull/73265 wait for that to hit `nixos-19.09` or switch to `release-19.09`.
2. Do a `nix-shell` for now, I haven't tested other things.
3. Add `-Ddefault_library=static` to `mesonFlags` so you do static builds, since we have no `dllimport`/`dllexport` and putting those in too soon will just cause merge hell.

Of course, I hope to automate all of this away.

p01arst0rm:
nice work :)

volth:
I'd suggest `with_gc` -> `with_boehm` (other `gc` too), because in Nix discourse, "GC" is mainly about garbage files in Nix Store

Ericson2314:
Yes, that's a good idea. Also, I think I should close this one until something more up to date is ready :).

p01arst0rm:
> Yes, that's a good idea. Also, I think I should close this one until something more up to date is ready :).

is this in regards to the meson port? are you waiting for it to be brought up to latest ?

PR #3172, opened 2019-10-28
add --pure option to dev-shell

mkenigs:
Fixes #2813 

If --pure is passed to dev-shell, call execvpe with only the variables in keepVars

Issue #2813, opened 2019-05-16
nix-shell --pure


CSVdB:
Right now, `nix dev-shell` forms a modern version of `nix-shell --impure`. There is no modernization of `nix-shell --pure` yet though.

matthewbauer:
Nice to see ```dev-shell``` being added!

But I always thought ```--pure``` was a weird feature. Why would you want a pure shell? Doesn't ```nix build``` already provide a pure environment for you? I don't see the use in providing a halfway between ```nix build``` and ```nix dev-shell```.

AmineChikhaoui:
@matthewbauer You can't do development inside `nix build`, so `--pure` is actually useful to make sure you're not inheriting anything from your local environment.

matthewbauer:
I just don't think anyone really wants purity when doing development. Some might think they do, which is why things like Docker are popular. But, there are many good reasons to inherit from the local environment. Things like localizations, timezones, dev tools, and other user configurations need to come from the parent environment to work correctly. Breaking these things in pursuit of "purity" seems like a bad idea. Otherwise, you can always do:

```
env -i nix dev-shell
```

to get the pure environment

mkenigs:
Was looking at this as well as #3107. Instead of passing env vars through an rcfile, would it work to just set all of them before the call to exec or maybe use execvpe instead? Then if pure is desired just clear everything other than exceptions beforehand.

Think this would potentially help #3107 - environment would be set independent of what shell is used, but bashFunctions could be run only in the case of bash.

Would that work/be helpful for me try to PR?

CMCDragonkai:
Is there documentation on dev-shell?

edolstra:
`nix dev-shell` has an option `--ignore-environment` now, so this can be closed.

CMCDragonkai:
Is nix dev-shell deprecating nix-shell?

On 2 February 2020 22:50:53 GMT+11:00, Eelco Dolstra <notifications@github.com> wrote:
>`nix dev-shell` has an option `--ignore-environment` now, so this can
>be closed.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/NixOS/nix/issues/2813#issuecomment-581127760

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

zimbatm:
@CMCDragonkai this is only touching the flakes branch of nix. Everybody should keep using `nix-shell` until flakes is ready.

PR #3191, opened 2019-11-01
environment fixes in run

mkenigs:
Code in #3172 which adds --ignore-environment to dev-shell is redundant because --ignore-environment is already implemented for run. I moved the related code in run to a separate function. If doing it this way works, I was thinking I could move it to a common class that both run and dev-shell can inherit from.

I also changed the way ignore environment is handled, creating a new char* array if ignoreEnvironment is set rather than calling clearEnv. The previous way of doing it had to getenv, clearenv, and setenv for every kept variable, and clearenv for all variables, so it seems like creating a new array might be a better way of doing it that just does one getenv for kept variables. Not sure if it makes the call to runProgram and handling of PATH too complicated.

### DIFF
--- a/src/nix/command.cc
+++ b/src/nix/command.cc
@@ -159,7 +159,7 @@ void MixEnvironment::setEnviron() {
 
         for (const auto & var : keep) {
             auto val = getenv(var.c_str());
-            if (val) stringEnv.emplace_back(fmt("%s=%s", var.c_str(), val));
+            if (val) stringsEnv.emplace_back(fmt("%s=%s", var.c_str(), val));
         }
         vectorEnv = stringsToCharPtrs(stringsEnv);
         environ = vectorEnv.data();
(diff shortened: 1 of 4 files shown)


edolstra:
Maybe we can factor out the environment handling in `nix run` into a `MIxEnvironment` class that could be used by `nix dev-shell`? That way the latter would also get the `--keep` and `--unset` flags.

mkenigs:
Factored code out into MixEnvironment and used it in both run and dev-shell. Changes would close #3172 and #2813

src/nix/command.cc
```diff
@@ -128,4 +128,48 @@ MixDefaultProfile::MixDefaultProfile()
...
+}
+
+void MixEnvironment::setEnviron() {
+    if (ignoreEnvironment) {
+        if (!unset.empty())
+            throw UsageError("--unset does not make sense with --ignore-environment");
+
+        for (const auto & var : keep) {
+            auto val = getenv(var.c_str());
+            if (val) stringEnv.emplace_back(fmt("%s=%s", var.c_str(), val));
```
edolstra: This doesn't compile since `stringEnv` doesn't exist.

edolstra:
Thanks, merged!

PR #3196, opened 2019-11-01
Kaiha/pr/readline

KaiHa:
Implement completion and history for `nix repl` when compiled against _libreadline_ instead of _libeditline_.

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -329,6 +329,11 @@ bool NixRepl::getLine(string & input, const std::string &prompt)
 
     if (!s)
       return false;
+#ifdef READLINE
+    if (*s) {
+        add_history(s);
+    }
+#endif
     input += s;
     input += '\n';
     return true;
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -137,6 +137,30 @@ NixRepl::~NixRepl()
 
 static NixRepl * curRepl; // ugly
 
+#ifdef READLINE
+static char * completionFuncReadline(const char * text, int state)
+{
+    static StringSet matches = curRepl->completePrefix(text);
+
+    if (state == 0) {
+        matches = curRepl->completePrefix(text);
+    }
+
+    auto it = matches.begin();
+    while (state > 0 && it != matches.end()) {
+        it++;
+        state--;
+    }
+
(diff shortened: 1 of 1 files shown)


src/nix/repl.cc
```diff
@@ -302,16 +336,25 @@ bool NixRepl::getLine(string & input, const std::string &prompt)
...
     input += s;
     input += '\n';
     return true;
 }
 
-
+#ifdef READLINE
+StringSet * NixRepl::completePrefix(string prefix)
+{
+    StringSet * completions = new StringSet;
```
zimbatm: What is the reason for the function to not to return the same type in both cases?

KaiHa: TBH I have no clue why had done it this way. I had made the changes a few months back. I have now cleaned up, the signature of the `completePrefix` function is now unchanged.

edolstra:
Hm, I would prefer not to support more than one library due to the code duplication and the increased maintenance/testing burden. (Note that we got rid of readline in the past (c5f23f10a84f568874321c04984b1a14d2dce978)).

KaiHa:
Eelco Dolstra <notifications@github.com> writes:

> Hm, I would prefer not to support more than one library due to the code duplication and the increased maintenance/testing burden. (Note that we got rid of readline in the past (c5f23f10a84f568874321c04984b1a14d2dce978)).

Yes, I know. I added it back as an option in [this PR][1], because the
libeditline used by Nix is not available in some Linux distribution.

And since the [packaging][2] [efforts][3] for Debian is taking up speed
again I thought it might be a good time to add these missing features.

[1]: https://github.com/NixOS/nix/pull/2551
[2]: https://salsa.debian.org/debian/nix
[3]: https://ftp-master.debian.org/new/nix_2.3.1-1.html

arcnmx:
With #3110 murdered, this is left as the remaining issue tracking readline in nix repl.

> I would prefer not to support more than one library due to the code duplication and the increased maintenance/testing burden

Is there any way these concerns could be addressed and we could move forward in some way here? The maintenance burden seems like it should be reasonable enough - the changes are relatively small and isolated to a single file that changes pretty infrequently (and afaik the patches haven't even had a conflict in years).

PR #3199, opened 2019-11-02
docs: xref doesn't render in title

kevinastock:
The `post-build-hook` text currently appears in the index, but not on the actual title line of the section, this follows the pattern used in a previous section to get a reference into a title for `diff-hook`.

Unrelated, the subtitle on the next line is one of two in the doc besides the version line at the top of the documentation. The other is also in the advanced features section and it renders strangely in the html, it's as large as the title itself, but italic and below the title underline. Is there any general or project specific guidance of how to use docbook for this project?

### DIFF
--- a/doc/manual/advanced-topics/post-build-hook.xml
+++ b/doc/manual/advanced-topics/post-build-hook.xml
@@ -5,7 +5,7 @@
       version="5.0"
       >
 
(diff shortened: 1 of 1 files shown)


edolstra:
I think it would be better to get rid of the subtitle. Maybe change the title to something like "Uploading to S3 using a post-build hook", or making it a subsection. @grahamc

PR #3203, opened 2019-11-03
Fix progress bar when nix-prefetch-url is piped.

hvdijk:
The intent of the code was that if the window size cannot be determined,
it would be treated as having the maximum possible size. Because of a
missing assignment, it was actually treated as having a width of 0.

The reason the width could not be determined was because it was obtained
from stdout, not stderr, even though the printing was done to stderr.

This commit addresses both issues.


edolstra:
Thanks!

Issue #13, opened 2012-05-07
Recursive Nix


edolstra:
Nix builders should be able to call Nix to build things. This is essential if we want to use Nix as a "low-level" build tool (i.e. as a Make replacement), since then we need to support derivations that unpack a source distribution containing a Nix expression to do the rest of the build.

Ericson2314:
Phase separation, such as quoting the `import "${some-drv}/build.nix"` sub expr like scheme can solve the query problem, and I believe is good model for recursive nix in general. (To continue phases analogy, manual invoking nix in a build script seems like eval, which, while strictly more powerful than multi-phase has all the usual issues). In particular, nondeterminism relating to dynamic dependencies in this light seems like a problem of macro hygiene.

taktoa:
## Introduction

I spent most of my internship this past summer at Awake Security on making Haskell builds incremental via IFD, and came to the conclusion that any solution to the incremental build problem that involves the build system is going to be brittle and is going to involve essentially reimplementing the logic of your build system in Nix. The right way to go is recursive Nix.

## My experience with incrementalization via IFD

I went into this project leaning heavily towards IFD as a solution to the incremental builds via Nix problem. I thought that it was a much more elegant solution when compared to recursive Nix.

The approach I had in mind would have been fairly reusable across different build systems as I was using Ninja as an intermediate representation (i.e.: splitting the task into two tools, `cabal2ninja` and `ninja2nix`). Unfortunately Ninja turned out to be a bad fit for the problem domain due to the semantics of its `depfile` feature; Ninja claims to have "perfect dependencies", but its dependency graph is only ever perfect after the project has been built once, which prevents a tool like `ninja2nix` from working.

In any case, I still think that, if we are dead-set on not using recursive Nix, the `ninja2nix` approach is the most practical IFD-based path towards a (partially) incrementalized nixpkgs. Yet it turned out to be very hairy even just for Haskell, let alone CMake or other build systems. This is because there is a fundamental problem with using IFD for incremental builds: it necessarily involves computing a completely static representation of the build graph. To get this data, I was only able to come up with three methods:

1. Modify the build system to compute a static build graph.
2. Instrument the build system, run it in some kind of dry-run mode, and try to compute the build graph from the log output. This log could also be computed in other ways, e.g.: by intercepting the glibc `execve` wrapper via LD_PRELOAD, but the point is that this method involves actually _running_ the build system and recovering the build graph from data produced via instrumentation.
[...]

PR #3205, opened 2019-11-04
Recursive Nix support

edolstra:
This allows Nix builders to call Nix to build derivations, with some limitations.
    
Example:

```nix    
let nixpkgs = fetchTarball channel:nixos-18.03; in
    
with import <nixpkgs> {};
    
runCommand "foo"
  {
    buildInputs = [ nix jq ];
    NIX_PATH = "nixpkgs=${nixpkgs}";
  }
  ''
    hello=$(nix-build -E '(import <nixpkgs> {}).hello.overrideDerivation (args: { name = "hello-3.5"; })')
    
    $hello/bin/hello
    
    mkdir -p $out/bin
    ln -s $hello/bin/hello $out/bin/hello
    
    nix path-info -r --json $hello | jq .
  ''
```
    
This derivation makes a recursive Nix call to build GNU Hello and symlinks it from its `$out`, i.e.
    
```sh
# ll ./result/bin/
lrwxrwxrwx 1 root root 63 Jan  1  1970 hello -> /nix/store/s0awxrs71gickhaqdwxl506hzccb30y5-hello-3.5/bin/hello
    
# nix-store -qR ./result
/nix/store/hwwqshlmazzjzj7yhrkyjydxamvvkfd3-glibc-2.26-131
/nix/store/s0awxrs71gickhaqdwxl506hzccb30y5-hello-3.5
/nix/store/sgmvvyw8vhfqdqb619bxkcpfn9lvd8ss-foo
```
    
This is implemented as follows:
    
* Before running the outer builder, Nix creates a Unix domain socket `.nix-socket` in the builder's temporary directory and sets `$NIX_REMOTE` to point to it. It starts a thread to process connections to this socket. (Thus you don't need to have `nix-daemon` running.)
    
* The daemon thread uses a wrapper store (`RestrictedStore`) to keep track of paths added through recursive Nix calls, to implement some restrictions (see below), and to do some censorship (e.g. for
  purity, `queryPathInfo()` won't return impure information such as signatures and timestamps).
    
* After the build finishes, the output paths are scanned for references to the paths added through recursive Nix calls (in addition to the inputs closure). Thus, in the example above, $out has a reference to `$hello`.
    
The main restriction on recursive Nix calls is that they cannot do arbitrary substitutions. For example, doing

```sh    
nix-store -r /nix/store/kmwd1hq55akdb9sc7l3finr175dajlby-hello-2.10
```
    
is forbidden unless `/nix/store/kmwd...` is in the inputs closure or previously built by a recursive Nix call. This is to prevent irreproducible derivations that have hidden dependencies on substituters or the current store contents. Building a derivation is fine, however, and Nix will use substitutes if available. In other words, the builder has to present proof that it knows how to build a desired store path from scratch by constructing a derivation graph for that path.
    
Probably we should also disallow instantiating/building fixed-output derivations (specifically, those that access the network, but currently we have no way to mark fixed-output derivations that don't
access the network). Otherwise sandboxed derivations can bypass sandbox restrictions and access the network.
    
When sandboxing is enabled, we make paths appear in the sandbox of the builder by entering the mount namespace of the builder and bind-mounting each path. This is tricky because we do a `pivot_root()` in the builder to change the root directory of its mount namespace, and thus the host `/nix/store` is not visible in the mount namespace of the builder. To get around this, just before doing `pivot_root()`, we branch a second mount namespace that shares its `/nix/store` mountpoint
with the parent.
    
Recursive Nix currently doesn't work on macOS in sandboxed mode (because we can't change the sandbox policy of a running build) and on Linux in non-root mode (because `setns()` barfs).

This PR also adds some ccache-like functionality to Nix's makefiles that wraps GCC calls in Nix derivations to enable caching and remote builds. This requires recursive Nix when you want to do this inside a Nix build.

Implements #13.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -895,6 +895,9 @@ private:
     /* The daemon main thread. */
     std::thread daemonThread;
 
+    /* The daemon worker threads. */
+    std::vector<std::thread> daemonWorkerThreads;
+
     /* Paths that were added via recursive Nix calls. */
     PathSet addedPaths;
 
@@ -2875,17 +2878,19 @@ void DerivationGoal::startDaemon()
 
             debug("received daemon connection");
 
-            // FIXME: process on a separate thread.
-            FdSource from(remote.get());
-            FdSink to(remote.get());
-            try {
-                daemon::processConnection(store, from, to,
-                    daemon::NotTrusted, daemon::Recursive, "nobody", 65535);
-            } catch (SysError &) {
-                ignoreException();
-            }
+            auto workerThread = std::thread([this, store, remote{std::move(remote)}]() {
+                FdSource from(remote.get());
+                FdSink to(remote.get());
+                try {
+                    daemon::processConnection(store, from, to,
+                        daemon::NotTrusted, daemon::Recursive, "nobody", 65535);
+                    debug("terminated daemon connection");
+                } catch (SysError &) {
+                    ignoreException();
+                }
+            });
 
-            debug("terminated daemon connection");
+            daemonWorkerThreads.push_back(std::move(workerThread));
         }
 
         debug("daemon shutting down");
@@ -2900,6 +2905,12 @@ void DerivationGoal::stopDaemon()
 
     if (daemonThread.joinable())
         daemonThread.join();
+
+    // FIXME: should prune worker threads more quickly.
+    // FIXME: shutdown the client socket to speed up worker termination.
+    for (auto & thread : daemonWorkerThreads)
+        thread.join();
+    daemonWorkerThreads.clear();
 
     daemonSocket = -1;
 }
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2248,7 +2248,8 @@ void DerivationGoal::startBuilder()
 
     /* Fire up a Nix daemon to process recursive Nix calls from the
        builder. */
-    startDaemon();
+    if (parsedDrv->getRequiredSystemFeatures().count("recursive-nix"))
+        startDaemon();
 
     /* Run the builder. */
     printMsg(lvlChatty, format("executing builder '%1%'") % drv->builder);
@@ -2832,6 +2833,8 @@ struct RestrictedStore : public LocalFSStore
 
 void DerivationGoal::startDaemon()
 {
+    settings.requireExperimentalFeature("recursive-nix");
+
     Store::Params params;
     params["path-info-cache-size"] = "0";
     params["store"] = worker.store.storeDir;
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -13,6 +13,7 @@
 #include "nar-info.hh"
 #include "parsed-derivations.hh"
 #include "machines.hh"
+#include "daemon.hh"
 
 #include <algorithm>
 #include <iostream>
@@ -33,6 +34,7 @@
 #include <sys/select.h>
 #include <sys/resource.h>
 #include <sys/socket.h>
+#include <sys/un.h>
 #include <fcntl.h>
 #include <netdb.h>
 #include <unistd.h>
@@ -765,9 +767,6 @@ private:
        immediate input paths). */
     PathSet inputPaths;
 
-    /* Referenceable paths (i.e., input and output paths). */
-    PathSet allPaths;
-
     /* Outputs that are already valid.  If we're repairing, these are
        the outputs that are valid *and* not corrupt. */
     PathSet validPaths;
@@ -805,8 +804,12 @@ private:
     /* Pipe for the builder's standard output/error. */
     Pipe builderOut;
 
-    /* Pipe for synchronising updates to the builder user namespace. */
+    /* Pipe for synchronising updates to the builder namespaces. */
     Pipe userNamespaceSync;
+
+    /* The mount namespace of the builder, used to add additional
+       paths to the sandbox as a result of recursive Nix calls. */
+    AutoCloseFD sandboxMountNamespace;
 
     /* The build hook. */
     std::unique_ptr<HookInstance> hook;
@@ -886,6 +889,26 @@ private:
     /* The remote machine on which we're building. */
     std::string machineName;
 
+    /* The recursive Nix daemon socket. */
+    AutoCloseFD daemonSocket;
+
+    /* The daemon main thread. */
+    std::thread daemonThread;
+
+    /* Paths that were added via recursive Nix calls. */
+    PathSet addedPaths;
+
+    /* Recursive Nix calls are only allowed to build or realize paths
+       in the original input closure or added via a recursive Nix call
+       (so e.g. you can't do 'nix-store -r /nix/store/<bla>' where
+       /nix/store/<bla> is some arbitrary path in a binary cache). */
+    bool isAllowed(const Path & path)
+    {
+        return inputPaths.count(path) || addedPaths.count(path);
+    }
+
+    friend class RestrictedStore;
+
 public:
     DerivationGoal(const Path & drvPath, const StringSet & wantedOutputs,
         Worker & worker, BuildMode buildMode = bmNormal);
@@ -941,6 +964,14 @@ private:
 
     /* Write a JSON file containing the derivation attributes. */
     void writeStructuredAttrs();
+
+    void startDaemon();
+
+    void stopDaemon();
+
+    /* Add 'path' to the set of paths that may be referenced by the
+       outputs, and make it appear in the sandbox. */
+    void addDependency(const Path & path);
 
     /* Make a file owned by the builder. */
     void chownToBuilder(const Path & path);
@@ -1043,6 +1074,7 @@ DerivationGoal::~DerivationGoal()
     /* Careful: we should never ever throw an exception from a
        destructor. */
     try { killChild(); } catch (...) { ignoreException(); }
+    try { stopDaemon(); } catch (...) { ignoreException(); }
     try { deleteTmpDir(false); } catch (...) { ignoreException(); }
     try { closeLogFile(); } catch (...) { ignoreException(); }
 }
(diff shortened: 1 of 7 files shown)


zimbatm:
Assuming that the inner build is quite large. How would the build be distributed to the same set of remote builders as the outer build?

One thing I was wondering is, if it would make sense for the inner build to just return a drv file in $out instead, and then let the outer scheduler update its build plan accordingly. This would be a bit closer to IFD but where the evaluation happens in a builder instead of all in the client.

Ericson2314:
@edolstra I am worry about exposing `nix-build`-in-`nix-build` because we do `nix-instantiate`-in-`nix-build` (my RFC). Your commit messages do mention putting behind a feature flag, and if we ban it in Nxpkgs initially that alleviates my worries. But, let me lay out those worries.

In short, while I think `nix-build`-in-`nix-build` is a decent last to retrofit existing shody stuff, it's never the way anything should *strive* to work.  `nix-build`-in-`nix-build` is worse because:

 - Timing is observable: derivations can observe how long nested `nix-build`s take, and therefore whether the thing they want to build was built before. This weakens our purity/caching. Even if we aren't building adversarial stuff that tries to exploit this, it still could could painful, hard to repoduce bugs.

 - "By default", it's sequential: you have to wait for the last nix-build to finish in your script, or waste effort writing your own parallelization code. With "ret cont" you don't need to put in effort to get better scheduling.

 - Resource usage: Derivations that idle waiting for nested nix-build waste space. I call `nix-instantiate`-in`nix-build` "ret cont" because compared to this you are morally serializing your continuation in the "sucessor" derivation. That successor derivation will almost surely be a lot smaller.

I don't want to be in a position where people write a bunch of stuff that uses `nix-build`-in`-nix-build` because of this PR, and then no one has the energy to rewrite it for `nix-instantiate`-in-`nix-build`. Even worse than the code is everybody getting excited about this, and then learning a bunch of other stuff---I'll admit ret-cont is weirder up front and more of an "unlearning" step. I'd like to avoid that tech debt and culture whiplash.

edolstra:
I've made an initial version of a nix-ccache flake: https://github.com/edolstra/nix-ccache. It provides a wrapper around gcc/g++ that executes the compilation of the preprocessed source in a recursive nix-build call.

matthewbauer:
> I've made an initial version of a nix-ccache flake: https://github.com/edolstra/nix-ccache. It provides a wrapper around gcc/g++ that executes the compilation of the preprocessed source in a recursive nix-build call.

Very cool! This looks a lot like what @layus talks about at NixCon. How costly is it to run every compilation in nix-build, though? Perhaps we need some heuristic to determine whether a C file is big enough to look up in the cache, otherwise we impose a constant builder setup for every .c file.

edolstra:
A quick unscientific measurement suggests the overhead is ~0.15s per GCC call on my laptop. (This also depends on the size of the preprocessor output, since it needs to be copied to the Nix store.) This is enough to make configure scripts much slower, so right now there is a special check to disable building through recursive Nix when the input is called "conftest". A heuristic like you suggest might be better.

volth:
Wouldn't it be more flexible to add in form of `builtins.nixBuild` accepting string of nix code to eval and returning a list of derivations?

Your example will be just
```diff
- hello=$(nix-build -E '(import <nixpkgs> {}).hello.overrideDerivation (args: { name = "hello-3.5"; })')
+ hello=${builtins.nixBuild ''(import <nixpkgs> {}).hello.overrideDerivation (args: { name = "hello-3.5"; })''}
```

Advantages are:
1. inner `nix-build`'s settings will be consistent with the outer, in terms that the inner won't have own `--cores` or `$TMPDIR`
2. it would easier to restrict what the `nix-build` is allowed to do
3. returning the list of derivations instead of raw stdout capture of `nix-build` executable, it could be used in `buildInputs = builtins.nixBuild "some nix code"`. for example, when building java apps, the inner nix-build could look up into `pom.xml` and build a list of `fetchurl`s for maven dependencies

Ericson2314:
@volth you just reinvented import from derivation :) But a big benefit of recursive (to me at least) is trying to leverage eval less not more, i.e. nix-exprs can just be one unprivileged way to get drv files.

Other than that< I think you might prefer https://github.com/NixOS/rfcs/pull/40.

>   1. inner `nix-build`'s settings will be consistent with the outer

Ret-cont recursive also does this.

> 3. returning the list of derivations instead of raw stdout capture of `nix-build` executable

Never need to go stdout->store path either.

edolstra:
> Wouldn't it be more flexible to add it in form of builtins.nixBuild accepting string of nix code to eval and returning a list of derivations?

Well, `hello` was just a toy example. The main usefulness of recursive Nix is if you *don't* know the inner derivations in advance. For example, in the case of `nix-ccache` (which wraps gcc invocations in nix builds), it's typically a makefile driving the build process, so you don't know at the outer expression level which inner builds are going to be done, or in what order.

kolloch:
Hi, I am searching for a working example with recursive nix.

Does it work with the nix in nixpkgs-unstable? Are there docs already?

zimbatm:
@kolloch have a look at https://github.com/NixOS/nix/pull/3205/files#diff-e9794c2e4d63a50bf65a1a0ce0873a19 for an example. The feature is hidden behind a feature flag.

In terms of Nix releases, it looks like it's only available in master at the moment.

kolloch:
I assume that does that mean my nix daemon also has to be from master?

zimbatm:
Yes, in this case the daemon needs to be updated as well since it controls the build sandbox.

mlvzk:
When will this land in stable?

zimbatm:
When Nix 3.0 will be released. Or use `nixpkgs.nixUnstable`

roberth:
> Probably we should also disallow instantiating/building fixed-output derivations (specifically, those that access the network, but currently we have no way to mark fixed-output derivations that don't
access the network). Otherwise sandboxed derivations can bypass sandbox restrictions and access the network.

Fixed-output derivations (FODs) would be part of a recursive Nix solution for [fetchNodeModules](https://github.com/NixOS/nixpkgs/pull/128749). I'd like for it to leverage Nix's ability to cache downloads in the form of FOD and share the downloaded modules in the form of store paths.

Instead of prohibiting FODs, we could stop the console logs from going to the recursive Nix socket and send them directly to the calling derivation's log instead. This way, the sandboxed build can only determine whether an FOD is fetchable. That's still sufficient to extract information from the network bit by bit, so we'd also have to "prohibit" failures by killing the parent derivation whenever a recursive derivation fails.

It's worth noting that the [ret-cont](https://github.com/NixOS/rfcs/pull/40) solution does not have this problem, but can support the `fetchNodeModules` use case.

kvtb:
any chance for backporting to 2.3 ?

lukego:
The example does not build for me. I wonder why?

I'm running NixOS from nixpkgs master with

```
  nix.package = pkgs.nixUnstable;
  nix.extraOptions = ''
    experimental-features = nix-command flakes recursive-nix
    trusted-users = luke
  '';
```

and I have copied the example derivation from the PR into `example.nix` but the various `nix build` commands I've tried are all failing:

```
[luke@snowy:~]$ nix build -L -f example.nix
foo> error: creating directory '/nix/var': Permission denied
error: builder for '/nix/store/ikqxl3gpd54ww9czp0hxydjf0i15a74y-foo.drv' failed with exit code 1;
       last 1 log lines:
       > error: creating directory '/nix/var': Permission denied
       For full logs, run 'nix log /nix/store/ikqxl3gpd54ww9czp0hxydjf0i15a74y-foo.drv'.

[luke@snowy:~]$ sudo nix build -L -f example.nix
foo> error: creating directory '/nix/var': Permission denied
error: builder for '/nix/store/ikqxl3gpd54ww9czp0hxydjf0i15a74y-foo.drv' failed with exit code 1

[luke@snowy:~]$ sudo nix --experimental-features 'nix-command recursive-nix' build -L -f example.nix
foo> error: creating directory '/nix/var': Permission denied
error: builder for '/nix/store/ikqxl3gpd54ww9czp0hxydjf0i15a74y-foo.drv' failed with exit code 1

[luke@snowy:~]$ sudo nix --experimental-features 'nix-command recursive-nix' build --impure -L -f example.nix
foo> error: creating directory '/nix/var': Permission denied
error: builder for '/nix/store/ikqxl3gpd54ww9czp0hxydjf0i15a74y-foo.drv' failed with exit code 1

[luke@snowy:~]$ sudo nix --experimental-features 'nix-command recursive-nix' build --no-sandbox --impure -L -f example.nix
foo> error: cannot open connection to remote store 'daemon': error: reading from file: Connection reset by peer
error: builder for '/nix/store/ikqxl3gpd54ww9czp0hxydjf0i15a74y-foo.drv' failed with exit code 1
```

Can anyone see which requirement I'm failing to fulfil? I'd love to use recursive Nix.

jkarni:
@lukego I also have the same issue. Have you been able to find a solution?

lukego:
@jkarni Yes, I have been able to get past this problem and onto the next problem with recursive nix, that it hangs in large builds :grin:. See https://github.com/NixOS/nix/issues/7297 for more on that and a (hopefully) working small example for reference.

I am *ahem* 80% sure that the problem above was resolved by adding `requiredSystemFeatures = [ "recursive-nix" ]`.

basile-henry:
> I am ahem 80% sure that the problem above was resolved by adding `requiredSystemFeatures = [ "recursive-nix" ]`.

I can confirm that the system feature was needed for me as well as enabling the experimental feature!

Why is that the case though? Where did you find documentation about this?
I didn't think `nix` ever cared about the names of system features, I always understood system features as a way to match derivations with builders, not a way to opt into nix experimental features I have already opted into (via `/etc/nix/nix.conf` or CLI).

roberth:
When using remote builders, you wouldn't want to schedule a derivation that requires the experimental feature onto a machine where it's not enabled, so Nix must require both.

basile-henry:
Ah, so `recursive-nix` is a special feature because under the hood it uses the local machine as a remote builder?

> you wouldn't want to schedule a derivation that requires the experimental feature onto a machine where it's not enabled

I understand that, but would have expected it to be on the user to ensure the experimental nix feature was enabled (either using system features, or simply by ensuring all their remote builders have the experimental nix feature enabled).

PR #3208, opened 2019-11-07
manual: Explain `nix-env -E` without `-f`

nh2:
Also explain how to install multiple derivations.


nh2:
Please reopen.

I never received any feedback on this manual improvement, and think it is still useful.

I think stalebots that close are pretty bad, as most people won't have the time and patience to keep answering to it. For example, today I already received 10 stalebot messages.

PR #3211, opened 2019-11-08
gitignore /precompiled-headers.h.gch

zimbatm:
Introduced recently in 5ff4d77f5585f9aad068fcc473393095e0640ae4


edolstra:
Can you also add `.h.pch` (which is what clang uses)?

zimbatm:
done

PR #3214, opened 2019-11-08
Switch to std::uncaught_exceptions

singron:
This returns the number of uncaught exceptions. std::uncaught_exception
just returned whether there were any uncaught exceptions. c++17
deprecated it and c++20 will remove it. GCC 9 warns that the function is
deprecated.

std::uncaught_exceptions() can be > 1 if all exceptions are caught
before they propagate to a stack unwind. E.g. if a first exception
occurs and a destructor runs as part of the stack unwind, then that
destructor may cause additional exceptions to be thrown as long as they
are caught before that destructor returns.

I'm unsure if this can occur in nix, but I handled it by recording the
first count in the constructor and comparing it to the count in the
destructor. That way the object can determine whether the exception was
thrown during its lifetime.

### DIFF
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -218,6 +218,7 @@ struct ConnectionHandle
 {
     Pool<RemoteStore::Connection>::Handle handle;
     bool daemonException = false;
+    int uncaught_exceptions = std::uncaught_exceptions();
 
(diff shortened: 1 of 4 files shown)


edolstra:
See 93b1ce1ac522ff92626da5cdd689008063935844.

PR #3216, opened 2019-11-09
Add an optional `hash` parameter to `builtins.fetchGit`

Ma27:
cc @globin @xbreak 

----

This is particularly useful to ensure the purity e.g. of a repo which is
pinned to a git tag (tags can be force-pushed which makes builds silently irreproducible)
or when trying to evaluate a nix expression without internet connectivity (if the
cache-entry is expired, the eval will break as the git repo can't be re-fetched).

This change adds an optional `hash` parameter (with an SRI-hash[1]) to the builtin
`fetchGit` and checks if there's a content-addressable path which matches the hash
(the hash can be determined using `nix to-sri --type sha256 $(nix-prefetch-git ...)`).

If no such path exists, it will be attempted to fetch the Git repository in
order to compare the hashes. Please note that caching is still used here, so
if the repo is already fetched and the only hash in the expression changes,
the evaluation will fail pretty fast.

[1] https://www.w3.org/TR/SRI/


doc/manual/expressions/builtins.xml
```diff
@@ -405,6 +405,15 @@ stdenv.mkDerivation { … }
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>sha256</term>
```
FRidh: I think we should support SRI hashes instead.

doc/manual/expressions/builtins.xml
```diff
@@ -504,6 +513,28 @@ stdenv.mkDerivation { … }
         <programlisting>builtins.fetchGit {
   url = "ssh://git@github.com/nixos/nix.git";
   ref = "master";
+}</programlisting>
+      </example>
+
+      <example>
+        <title>Fetching a given revision from a repo and validate its sha256</title>
+        <para>
+          <function>builtins.fetchGit</function> can create a fixed-output
```
FRidh: Expressions that are fetched with `builtins.fetchGit` and `builtins.fetchTarball` are not derivations if I understand correctly.

Ma27:
> While I want to be able to fetch them, they are mutable

The `ref` argument is needed to e.g. fetch git tags (checking out a git tag as a `rev` didn't work here, as far as I understand the `fetchgit` code from `nixpkgs`, `ref`/`rev` is mixed together, so `rev = "v1.2.3"` works there, but not with `builtins.fetchGit`).

doc/manual/expressions/builtins.xml
Ma27: Ah you're right. As far as I understand this, those are actually paths, so "fixed-output path" (the needed function from the store's API is named `makeFixedOutputPath` btw) might be a better choice here?

Ericson2314:
CC @matthewbauer

matthewbauer:
> > While I want to be able to fetch them, they are mutable
> 
> The `ref` argument is needed to e.g. fetch git tags (checking out a git tag as a `rev` didn't work here, as far as I understand the `fetchgit` code from `nixpkgs`, `ref`/`rev` is mixed together, so `rev = "v1.2.3"` works there, but not with `builtins.fetchGit`).

I would consider fetchGit with "rev" to be fixed-output and fetchGit with "ref" to be not fixed-output. In Git, branches and tags are considered references, and technically are not content-addressable (although tags require force push). Perhaps another arg could be added like `tag` or `namedRev` for only the tag subset of refs.

Ma27:
@globin and I actually discussed about this last weekend :) 

The reason why `ref` is currently used is because tags can be fetched with it and we use tags (although those are not 100% content-addressable) pretty often in nixpkgs.

> Perhaps another arg could be added like tag or namedRev for only the tag subset of refs.

That's a good idea actually! Will implement this probably next weekend :)

edolstra:
Could you replace `sha256` with generic `hash` argument containing an SRI hash, as @FRidh suggests? We've been trying to move away from `sha1/sha256/...` arguments (see e.g. `<nix/fetchurl.nix>`).

doc/manual/expressions/builtins.xml
edolstra: That phrase should be avoided (since they're not outputs), "content-addressed" is more universal.

Ma27: Agreed. Avoiding the term "fixed-output" in this context might prevent even more people from confusing that just as I did :)

matthewbauer:
> Morally speaking, fetching using a `ref` and a `hash` is not worse than using `rev`.

Won't fetching use a `ref` and a `hash` inevitably lead to hash mismatch errors, though? Perhaps to Nix it's all the same, but not to Git. The `rev` commit hash is content-addressable: derived from its parent commit(s), working tree, and commit metadata. Nix could be taught one day to recognize it as such, making the `hash` argument here optional. On the other hand we could never do the same for `ref` as it is only a pointer to a commit.

globin:
Yeah but we often want to pin on a tag and guard that with the hash to notice if people force push a new version and still have a pure evaluation.

edolstra:
> Nix could be taught one day to recognize it as such, making the hash argument here optional.

That would be tricky because the `rev` is a hash over the entire history rather than the current tree. So unless we copy the entire history to the store, Nix won't be able to verify the `rev`.

BTW with flakes it's fine to use just a tag/branch because it will get locked to a specific `rev` in the lock file.

doc/manual/expressions/builtins.xml
Ma27: The patch now provides a single `hash` argument which expects an SRI hash. In case of a hash mismatch, SRI hashes will be printed and it's documented how to generate an SRI hash of a git repo (`nix to-sri --type sha256 $(nix-prefetch-git ...)`).

doc/manual/expressions/builtins.xml
Ma27: Using the term `content-addressable path` now.

Ma27:
@edolstra may I ask if there's anything else to add/fix? :)

Ma27:
@edolstra may I ask if there's anything else I'm currently missing? :)

Ericson2314:
@edolstra Matt and I were also talking about using tree hashes, which Nix could verify without storing history.

Ma27:
Just updated the patch to work with the refactorings on the store API on master :)

Ma27:
@edolstra @Ericson2314 is there anything I can address here atm? :)

edolstra:
The new fetcher infrastructure adds a generic `narHash` attribute to all input types, e.g.
```
fetchTree { type = "tarball"; url = file://foo/bar.tar.gz; narHash = "sha256-xdKv2pq/IiwLSnBBJXW8hNowI4MrdZfW+SYqDQs7Tzc="; }
```
I think that's better than adding ad hoc support to specific fetchers. Also, the name `hash` is pretty ambiguous since it could be confused with a commit hash (i.e. the `rev` attribute).

jtojnar:
Related: https://github.com/NixOS/nixpkgs/pull/79987

Ma27:
@jtojnar did you link this just for documentation purposes? This patch only supports SRI hashes as well :)

I'll take a closer look at the new fetcher code later to confirm that this can be closed now.

Ma27:
@edolstra so I took a closer look at the new fetcher-infrastructure today. I currently have two questions about this:

* Would it be possible to "backport" this to nix-master? I'd really love to use this feature on stable nix :)
* The `git`-fetcher doesn't seem to support a `narHash`-attribute atm. I guess that in favor of this PR I could file a patch that adds `narHash`-support to `fetchTree { type = "git"; }`, right?

Ma27:
Closing now. The fetcher-API has been backported to nix-master quite recently. I already have a patch locally which adds support for the `narHash`-functionality (same hash-values as in my `hash`-param).... will file a PR tonight or tomorrow. Thanks for all the feedback!

Ericson2314:
> @edolstra Matt and I were also talking about using tree hashes, which Nix could verify without storing history.

For anyone else reading this, that's now done in https://github.com/NixOS/nix/pull/3635

PR #3219, opened 2019-11-10
Fix extra semicolons warnings

Ericson2314:
Very low hanging fruit!

### DIFF
--- a/src/libstore/store-api.hh
+++ b/src/libstore/store-api.hh
@@ -21,7 +21,7 @@ namespace nix {
 
 
 MakeError(SubstError, Error);
-MakeError(BuildError, Error) /* denotes a permanent build failure */;
+MakeError(BuildError, Error); // denotes a permanent build failure
 MakeError(InvalidPath, Error);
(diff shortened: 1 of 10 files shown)


src/libstore/store-api.hh
```diff
@@ -20,12 +20,12 @@
...
 
 
-MakeError(SubstError, Error)
-MakeError(BuildError, Error) /* denotes a permanent build failure */
-MakeError(InvalidPath, Error)
-MakeError(Unsupported, Error)
-MakeError(SubstituteGone, Error)
-MakeError(SubstituterDisabled, Error)
+MakeError(SubstError, Error);
+MakeError(BuildError, Error) /* denotes a permanent build failure */;
```
bhipple: This technically works, but it'd be nicer to put the `;` next to the actual expression instead of after the comment.

Ericson2314: Hah! Too much sed

Ericson2314:
Thanks!

PR #3238, opened 2019-11-25
Ensure enough space in attrset bindings

puckipedia:
When allocating bindings for attrset when using __overrides, we need to use the capacity of the original attrs, otherwise there's no space for dynamic attributes.

```nix
(rec { "${"foo"}" = 5; __overrides = { b = 5; }; }.foo)
```

### DIFF
--- a/tests/lang/eval-okay-attrs6.nix
+++ b/tests/lang/eval-okay-attrs6.nix
@@ -1,0 +1,4 @@
+rec {
+  "${"foo"}" = "bar";
+   __overrides = { bar = "qux"; };
+}
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -875,7 +875,7 @@ void ExprAttrs::eval(EvalState & state, Env & env, Value & v)
         if (hasOverrides) {
             Value * vOverrides = (*v.attrs)[overrides->second.displ].value;
             state.forceAttrs(*vOverrides);
-            Bindings * newBnds = state.allocBindings(v.attrs->size() + vOverrides->attrs->size());
+            Bindings * newBnds = state.allocBindings(v.attrs->capacity() + vOverrides->attrs->size());
             for (auto & i : *v.attrs)
                 newBnds->push_back(i);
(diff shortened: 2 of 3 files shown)


grahamc:
Can you provide a test case?

puckipedia:
Added.

tests/lang/eval-okay-attrs6.nix
```diff
@@ -0,0 +1,4 @@
+rec {
+  "${"foo"}" = "bar";
+   __overrides = { bar = "qux"; };
+}
```
grahamc: before this PR:

```
[grahamc@Petunia:~]$ ls -la $(which nix)
lrwxrwxrwx 1 root root 59 Dec 31  1969 /run/current-system/sw/bin/nix -> /nix/store/dj2ylmvx2c288h51yx8ljg4h6bh5vyla-nix-2.3/bin/nix

[grahamc@Petunia:~]$ nix repl
Welcome to Nix version 2.3. Type :? for help.

nix-repl> rec {
            "${"foo"}" = "bar";
             __overrides = { bar = "qux"; };
          }
nix: src/libexpr/attr-set.hh:54: void nix::Bindings::push_back(const nix::Attr&): Assertion `size_ < capacity_' failed.
zsh: abort (core dumped)  nix repl
```

after this PR:

```
[nix-shell:~/projects/github.com/NixOS/nix]$ ./inst/bin/nix repl --experimental-features nix-command

nix-repl> rec {
                      "${"foo"}" = "bar";
                       __overrides = { bar = "qux"; };
                    }
{ __overrides = { ... }; bar = "qux"; foo = "bar"; }
```

edolstra:
I don't understand why the capacity of `v.attrs` matters, since the size should reflect what we're copying from `v.attrs`:
```c++
            for (auto & i : *v.attrs)
                newBnds->push_back(i);
```

puckipedia:
At the point that the reallocation is done, `v.attrs` only contains non-dynamic attrs, so using the size means that there is no space left for dynamic attrs after `__overrides` has been processed.

PR #3262, opened 2019-12-11
Content-addressed paths

regnat:
(draft) companion MR for NixOS/rfcs#62

Makes derivations with `contentAddressed = true` content-addressed in the store.

You probably don't want to try this in an existing Nix installation since it requires a change in the database schema (and there's no guaranty that it won't corrupt the store ;) )


tomberek:
Minimal testing seems promising. I'm not as strong in C++, but I can perform some testing, review, etc.

I can see aliasOf added in `src/libstore/schema.sql`, but does that also do a migration for people with a previous schema?

regnat:
Closing in favor of https://github.com/NixOS/nix/pull/3528

PR #3265, opened 2019-12-12
libstore/ssh: Improve error message on failing `execvp`

Profpatsch:
If the `throw` is reached, this means that execvp into `ssh` wasn’t
successful. We can hint at a usual problem, which is a missing `ssh`
executable.

Test with:

```
env PATH= ./result/bin/nix-copy-closure --builders '' unusedhost
```

and the bash version with

```
env PATH= ./result/bin/nix-copy-closure --builders '' localhost
```


Profpatsch:
Please inspect my C++ code carefully, I’m like a helpless child.

edolstra:
Thanks!

Issue #3153, opened 2019-10-17
nix-shell shebang with '.nix-shell' suffix executed as nix code.


nyarly:
Consider this shebang script:

```
#! /usr/bin/env nix-shell
#! nix-shell -i bash

echo "Hello!"
```

If saved has `hello.nix-shell`, when run it will output

`error: undefined variable 'echo' at /tmp/test.nix-shell:4:1`

It seems that the shebang detection code puts us into a nix language execution mode instead of honoring the `-i` flag.

https://github.com/NixOS/nix/blob/906d56a96b442d4dd8f924c1ce0d1eec0e214af3/src/nix-build/nix-build.cc#L73 was identified as potentially related.

Infinisil:
This seems to be the actual cause: https://github.com/NixOS/nix/blob/906d56a96b442d4dd8f924c1ce0d1eec0e214af3/src/nix-build/nix-build.cc#L109

That is, if the filename it was invoked with has "nix-shell" in it, it *doesn't* interpret is as a shebang.

I don't know what this check is for, I suspect it's just an oversight and can probably be removed.

I traced it back to https://github.com/nixos/nix/commit/dc7e8fae48f65da7aea86043a25a9381b678ebe6 by @edolstra

zimbatm:
It's because `nix-shell` is a symlink to `nix`. It's using the busybox-style packing of multiple programs into one by detecting the argv1. The issue is that the regex should be something like `(^|/)nix-shell$`

Infinisil:
@zimbatm The `runEnv` already includes the check on `argv[0]` (which is the executable name, not `argv[1]`).

PR #3269, opened 2019-12-14
nix-shell: don't check for "nix-shell" in shebang script name

xzfc:
Fix bug: if the path to the nix-shell shebang script contains `"nix-shell"`, it won't run properly.

```sh
$ cat a-script-that-should-be-run-in-nix-shell.sh
#! /usr/bin/env nix-shell
#! nix-shell -p hello -i sh
hello

$ ./a-script-that-should-be-run-in-nix-shell.sh
error: undefined variable 'hello' at /tmp/a-script-that-should-be-run-in-nix-shell.sh:3:1

$ mv a-script-that-should-be-run-in-nix-shell.sh a-script.sh

$ ./a-script.sh
Hello, world!
```

Probably this check was relevant when nix-build [was a perl script](https://github.com/NixOS/nix/commit/80ebc553eca19dafc64c47420cd49ddd506bc9b7) (as perl handles `argv` differently than C++).

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -106,7 +106,7 @@ static void _main(int argc, char * * argv)
     // Heuristic to see if we're invoked as a shebang script, namely,
     // if we have at least one argument, it's the name of an
     // executable file, and it starts with "#!".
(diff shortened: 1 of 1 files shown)


zimbatm:
this would fix https://github.com/NixOS/nix/issues/3153

Issue #1492, opened 2017-07-27
nix-collect-garbage -d should issue a warning when run as non-root user


chris-martin:
I've been using NixOS for about 16 months and I [just now figured out](https://github.com/NixOS/nixpkgs/issues/23926#issuecomment-318237755) that you have to run `nix-collect-garbage` as root to remove old NixOS builds. In retrospect, this probably should have been obvious to me, since it would be bad to allow non-root users to delete OS generations. Still I think it would be a good idea for `nix-collect-garbage` to print a something near the end of the output reminding the user that only their *profile* generations have been deleted, and that you need to run the command again with `sudo` if you want to delete OS generations.

cleverca22:
if ran as root, it can delete generations from all users, so no need to run it twice then

dtzWill:
Agreed a message or warning would be nice :+1:.

I like the current behavior but never hurts to make sure user is clear on what is (and isn't) happening :).

fricklerhandwerk:
This is a NixOS issue and something to be documented in the NixOS manual. @chris-martin please open an issue on [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) if you still care.

For Nix we should at least make more explicit in the manual that `nix-collect-garbage -d` only deletes the current user's profiles.

tomberek:
Closing as stale. Re-open if needed.

PR #3286, opened 2019-12-22
nix-collect-garbage: print warning when we can't remove old generations

andrew-d:
I ran into https://github.com/NixOS/nix/issues/1492 recently and figured that it'd be nice to print a warning here.

cc @dtzWill @chris-martin @cleverca22

### DIFF
--- a/src/nix-collect-garbage/nix-collect-garbage.cc
+++ b/src/nix-collect-garbage/nix-collect-garbage.cc
@@ -19,7 +19,10 @@ bool dryRun = false;
 
 void removeOldGenerations(std::string dir)
 {
-    if (access(dir.c_str(), R_OK) != 0) return;
+    if (access(dir.c_str(), R_OK) != 0) {
+        printError(format("warning: could not remove generations in: %1%") % dir);
+        return;
+    }
 
     bool canWrite = access(dir.c_str(), W_OK) == 0;
 


src/nix-collect-garbage/nix-collect-garbage.cc
```diff
@@ -19,7 +19,10 @@ bool dryRun = false;
 
 void removeOldGenerations(std::string dir)
 {
-    if (access(dir.c_str(), R_OK) != 0) return;
+    if (access(dir.c_str(), R_OK) != 0) {
+        printError(format("warning: could not remove generations in: %1%") % dir);
```
zimbatm:
```suggestion
        printError(format("warning: cannot remove generation %1%: Permission denied") % dir);
```
Be more explicit about the type of error, follow the same pattern as `rm`
```
$ rm /etc/hosts 
rm: cannot remove '/etc/hosts': Permission denied
```

edolstra: Note that `format` is unnecessary. You can also do:
```
        printError("warning: cannot remove generation '%s': Permission denied", dir);
```

edolstra:
I'm not so sure about this. It sounds like it could spam the user's terminal with a lot of fairly useless messages (you get a warning for every profile directory to which you don't have access).

Maybe it would be better to have a separate flag to clean up *all* profiles vs. the user's profiles. Then the former can print an error (not a warning) if it doesn't have access while the latter can be quiet.

PR #3289, opened 2019-12-23
Pass -J to tar for xz decompression

michaelforney:
Some tar implementations can't auto-detect compression formats, so
they must be specified explicitly.


edolstra:
Does this work on macOS?

michaelforney:
Yes, I believe macOS uses tar from libarchive, which does support `-J`.

PR #3291, opened 2019-12-26
nix/main: Set `verboseBuild` to `args.printBuildLogs`

Ma27:
When running `nix build -L`, it's slightly confusing to get the last 10
loglines again after the log-output.

### DIFF
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -143,11 +143,11 @@ void mainWrapped(int argc, char * * argv)
     }
 
     verbosity = lvlWarn;
-    settings.verboseBuild = false;
 
     NixArgs args;
 
     args.parseCmdline(argvToStrings(argc, argv));
+    settings.verboseBuild = args.printBuildLogs;
 
     settings.requireExperimentalFeature("nix-command");
 


edolstra:
Repeating the last log lines of the failing build is not necessarily a bad thing, for example when you're using `--keep-going` (since then the original error may have scrolled past a long time ago).

Ma27:
I'm afraid I have to disagree here: at least in my experience the majority of errors occurs before the last 10 log-lines, that's one of the reasons why I almost always use `-L`.

Would it be a compromise in your opinion to only enable such a behavior if `--keep-going` is not set?

edolstra:
That may be true but in that case we should just get rid of the printing of the last few lines entirely (and just say something like "derivation X failed; run `nix log X` to see the build log").

Ma27:
So you'd be fine with removing the "last N log lines"-feature? (Please note that I only care about it, if `-L` is set as I get regularly confused with having several lines duplicated when having a failed build).

edolstra:
Yes, sounds good to me.

Ma27:
Just discovered that this can be done by setting the log-lines option, so this can be closed IMHO.

@edolstra should we still remove the functionality (or set log-lines by default to 0)?

PR #3297, opened 2020-01-02
passAsFile: hash the attribute name instead of numbering sequentially

edef1c:
This makes the paths consistent without relying on ordering.


edolstra:
Thanks!

grahamc:
:joy:

PR #3305, opened 2020-01-06
Add support for unicode escape sequences in fromJSON

### LINKED ISSUE #2257: Unicode in Github description causes hydra to fail to evaluate
A PR was created that was using the github plugin for hydra that contained some unicode characters. This caused hydra to stop evaluating declarative jobsets. All existing jobsets continued to function, but the configuration could no longer be changed.

https://github.com/NixOS/nix/blob/bd56b5fe3f5374e845fd58f8ecd9e111f2e5104c/src/libexpr/json-to-value.cc#L31 is the exact line of the code where it fails

and this is a fragment of the JSON that failed to parse
```
returned exit code 1:\r\n\u001b[31;1merror:\u001b[0m access to path
```

knl:
As fromTOML supports unicode escape sequences, bring fromJSON on par. JSON defaults
to UTF-8 encoding (every JSON parser must support UTF-8), thus this change parses the
`u hex hex hex hex` sequence (escaped `u` followed by 4 hexadecimal digits) into an
UTF-8 representation.

Add a test to verify correct parsing, using all escape sequences from json.org.

Caught by @basvandijk while debugging Hydra issue, where JSON strings coming from GitHub  contained `\ u` (on purpose written with a space in between to avoid this Hydra issue), which prevented the evaluation of any jobset in a project, as JSON parsing was failing.

Fixes #2257

### DIFF
--- a/src/libexpr/json-to-value.cc
+++ b/src/libexpr/json-to-value.cc
@@ -22,7 +22,6 @@ static string parseJSONString(const char * & s)
             if (*s == '"') res += '"';
             else if (*s == '\\') res += '\\';
             else if (*s == '/') res += '/';
-            else if (*s == '/') res += '/';
             else if (*s == 'b') res += '\b';
             else if (*s == 'f') res += '\f';
             else if (*s == 'n') res += '\n';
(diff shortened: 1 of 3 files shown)


edolstra:
Thanks!

piegamesde:
As far as I can tell, this code has not been released yet. When will it be included in a release? Will/can it be back-ported to 2.3?

PR #3307, opened 2020-01-09
builtins.fromJSON: use nlohmann/json parser instead of custom parser

yorickvP:
Nix already links to [nlohmann/json](https://github.com/nlohmann/json), so we might just as well use it for JSON parsing, instead of a custom parser that is accumulating complexity.

This uses the sax backend to avoid accumulating a lot of data in memory before conversion to nix. I am not sure about the use of new/delete in combination with things that might end up GC'd, but I suspect this is fine because they are all gone before a GC ever happens.

### DIFF
--- a/src/libexpr/json-to-value.cc
+++ b/src/libexpr/json-to-value.cc
@@ -5,6 +5,7 @@
 #include <nlohmann/detail/exceptions.hpp>
 
 using json = nlohmann::json;
+using std::unique_ptr;
 
 namespace nix {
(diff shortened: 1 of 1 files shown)


edolstra:
Looks good, thanks.

Maybe you can use `std::unique_ptr` instead of new/delete?

yorickvP:
Done! It's an improvement.

edolstra:
Thanks!

PR #3313, opened 2020-01-11
libexpr: show expression in assertion errors

LnL7:
Includes the expression of the condition in the assertion message if
the assertion failed, making assertions much easier to debug. eg.

    error: assertion (withPython -> (python2Packages != null)) failed at pkgs/tools/security/nmap/default.nix:11:1


bjornfor:
That's very useful! There is a small typo in the commit message.

FRidh:
let me guess, overrides did not function because of the assert?

LnL7:
Thanks!

PR #3315, opened 2020-01-12
Unbreak build without pch

Ericson2314:
The change for WSL relied on `#include <sys/utsname.h>` being included everywhere. Also change `release.nix` so we build both builds in hydra/ofborg, preventing such regressions in the future.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -98,6 +98,7 @@ let
 
 
     build = pkgs.lib.genAttrs systems (system:
+      pkgs.lib.genAttrs [ "regular-headers" "precompiled-headers" ] (pchVariant:
 
       let pkgs = import nixpkgs { inherit system; }; in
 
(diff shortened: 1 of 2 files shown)


edolstra:
Thanks, I've cherry-picked the first commit and did the regression test in a cheaper way (c86c71c2b1feac19f3363662cf20abc88ab47e03).

Ericson2314:
Err I understand wanting to keep the job names the same, but I think we really ought to CI everything we claim to to support, be it for production or ease of development.

edolstra:
What's the problem? It *is* tested now.

Ericson2314:
The opposite regression is still possible. Say someone develops without PCH for some reason, and breaks it. There would be no warning until someone else (e.g. you) goes back in a nix shell and finds their development environment broken.

Basically in my view if we aim to support both `PRECOMPILED_HEADERS=0` and `PRECOMPILED_HEADERS=1`, then Hydra should build both.

Ericson2314:
Oh forgive me, it was just the coverage checks doing this not all the CI builds.

PR #3318, opened 2020-01-14
doc: touchup release notes for 2.3

bhipple:
- At the top of the release notes, we announce sandboxing is now enabled by default,
then at the bottom it says it's now disabled when missing kernel support. These
can be merged into one point for clarity.

- The point about `max-jobs` defaulting to 1 appears unrelated to sandboxing.

CC @edolstra


edolstra:
Thanks!

PR #3321, opened 2020-01-15
Fix #2984: attempt to copy if unlinking source directory fails

### LINKED ISSUE #2984: Can't build NixOS from source with a chroot store
This looks like a Nix bug to me, not a problem in nixpkgs. If you use a chroot store, Nix is unable to build a NixOS VM without substituters. You get a `Permission denied` error regardless of whether you use a multi-user or single-user install of Nix.

```
nix build --no-link --store $(pwd)/foo --substituters "" --cores 30 -I nixpkgs=channel:nixos-19.03 -f "<nixpkgs/nixos>" --arg configuration '{}' vm
[11 built]
error: moving build output '/nix/store/kjqdwl76vp9wclcxpw2xmh0z73rq95py-documentation-highlighter' from the sandbox to the Nix store: Permission denied
```

yorickvP:
[We were seeing an error](https://github.com/input-output-hk/haskell.nix/pull/399) on single-user nix mode, where `rename` fails if the right permissions on the source path are missing. Since the paths are removed in sandbox mode anyways, we can use `copyPath` to work around this.

Testcase: `nix-build '<nixpkgs>' -A documentation-highlighter --substituters ""`

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -3591,8 +3591,15 @@ void DerivationGoal::registerOutputs()
                 if (buildMode == bmRepair)
                     replaceValidPath(path, actualPath);
                 else
-                    if (buildMode != bmCheck && rename(actualPath.c_str(), worker.store.toRealPath(path).c_str()) == -1)
-                        throw SysError(format("moving build output '%1%' from the sandbox to the Nix store") % path);
+                    if (buildMode != bmCheck && rename(actualPath.c_str(), worker.store.toRealPath(path).c_str())) {
+                        if(errno == EACCES) {
+                            // No need to free since we are in a chroot, and that gets deleted later.
+                            // Hope it's a source problem
+                            copyPath(actualPath, worker.store.toRealPath(path));
+                        } else {
+                            throw SysError(format("moving build output '%1%' from the sandbox to the Nix store") % path);
+                        }
+                    }
             }
             if (buildMode != bmCheck) actualPath = worker.store.toRealPath(path);
         }


edolstra:
Hm, do you know *why* `EACCES` happens here?

yorickvP:
I suspect the build creates an `actualPath/..` with no write bit set. Being root will ignore that completely, but single-user mode would have a problem. That suggests we can also fix this problem with a well-placed chmod.

zhenyavinogradov:
Another trivial expression that triggers the problem: `(import <nixpkgs> {}).runCommand "test" {} "mkdir --mode u-w $out"`

domenkozar:
There's quite a few references here and on #2984 that trigger this bug on single-user installation.

nh2:
I can confirm this issue on a single-user installation on Ubuntu:

`nix-build mypackage.nix --out-link .nix-build-gc-root` gives:

```
error: moving build output '/nix/store/...mypackage-0.1.0.0' from the sandbox to the Nix store: Permission denied
```

Also including the perms for

```
% ls -lah .nix-build-gc-root
lrwxrwxrwx 1 niklas niklas 58 Jun  9 14:22 .nix-build-gc-root -> /nix/store/...-mypackage-0.1.0.0/
```

but I don't think those matter, given that `rm .nix-build-gc-root` changes nothing about the error.

thomasjm:
@yorickvP doing a `chmod u+w` or equivalent on the source path sounds better. I just experimented locally and confirmed that the problem seems to arise iff the source path is missing the write bit, like you said. Falling back from `rename` to `copyPath` seems like it could incur a performance penalty to copy a large tree, no?

Ericson2314:
BTW https://git.savannah.gnu.org/cgit/guix.git/commit/?id=d445c30ea6f9693454ca96bae9089ba8738a6f78 seems slightly related.

mkaito:
I ran into this issue again today. I suppose there aren't that many single-user nix users out there. Should I reinstall nix in multi-user mode, or can I hope for this to get fixed?

nh2:
I found a workaround for my case:

I had written a derivation that copied the outputs of another derivation like this:

```sh
cp -r ${pkgs.otherDerivation} $out
```

After that, the permissions of `$out` are `dr-xr-xr-x` and then that fails to copy from the sandbox to the nix store.

My workaround is to do this instead:

```sh
cp -r ${pkgs.otherDerivation} $out
chmod -R u+w $out
```

then it works also in my single-user install.

Ericson2314:
Perhaps #4054 fixed this?

yorickvP:
Hopefully fixed by #4312, have no way to reproduce this anymore.

PR #3322, opened 2020-01-15
WIP: Coverage per-platform

Ericson2314:
I was quite confused in https://github.com/NixOS/nix/pull/3315#issuecomment-574312784 (sorry!) but shouldn't we do the coverage test on more platforms than just Linux? Say a missing header bug is Darwin (or Windows!) only?

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -231,8 +231,9 @@ let
         '');
 
 
-    coverage =
-      with pkgs;
+    coverage = pkgs.lib.genAttrs systems (system:
+
+      let pkgs = import nixpkgs { inherit system; }; in with pkgs;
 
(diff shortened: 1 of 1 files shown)


edolstra:
No, I don't think so. Since the tests are largely the same the coverage will be largely the same.

Ericson2314:
I did get as far as finding http://logan.tw/posts/2015/04/28/check-code-coverage-with-clang-and-lcov/

If not coverage, then I think we should have a distinct attribute for building with `PRECOMPILE_HEADERS=0` for all platforms.

Ericson2314:
Reopened the original alternative https://github.com/NixOS/nix/pull/3323

PR #3323, opened 2020-01-15
CI with and without precompiled headers on multiple platforms

Ericson2314:
Coverage isn't so good as it is linux-only.

Put system at the end as its idiomatic.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -97,7 +97,8 @@ let
       };
 
 
-    build = pkgs.lib.genAttrs systems (system:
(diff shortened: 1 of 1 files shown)


release.nix
```diff
@@ -166,7 +170,7 @@ let
       with import nixpkgs { inherit system; };
 
       let
-        toplevel = builtins.getAttr system jobs.build;
+        toplevel = builtins.getAttr system jobs.build.recompiled-headers;
```
matthewbauer: this looks like a mistake

PR #3325, opened 2020-01-17
nix-shell: clean up the tmpDir and escape variables

xzfc:
The problem fixed: each nix-shell invocation creates a new temporary directory (`/tmp/nix-shell-*`) and never cleans up.

And while I'm here, shellescape all variables inlined into the rcfile. See what might happen without escaping:

```bash
$ export TZ="';echo pwned'"
$ nix-shell -p hello --run hello
pwned
Hello, world!
```

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -423,13 +423,18 @@ static void _main(int argc, char * * argv)
            lose the current $PATH directories. */
         auto rcfile = (Path) tmpDir + "/rc";
         writeFile(rcfile, fmt(
-                (keepTmp ? "" : "rm -rf '%1%'; "s) +
+                R"(_nix_shell_clean_tmpdir() { rm -rf %1%; }; )"s +
+                (keepTmp ?
+                    "trap _nix_shell_clean_tmpdir EXIT; "
+                    "exitHooks+=(_nix_shell_clean_tmpdir); "
+                    "failureHooks+=(_nix_shell_clean_tmpdir); ":
+                    "_nix_shell_clean_tmpdir; ") +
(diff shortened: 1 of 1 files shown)


src/nix-build/nix-build.cc
```diff
@@ -423,13 +436,18 @@ static void _main(int argc, char * * argv)
            lose the current $PATH directories. */
         auto rcfile = (Path) tmpDir + "/rc";
         writeFile(rcfile, fmt(
-                (keepTmp ? "" : "rm -rf '%1%'; "s) +
+                R"(_nix_shell_clean_tmpdir() { rm -rf %1%; }; )"s +
+                (keepTmp ?
+                    "trap _nix_shell_clean_tmpdir EXIT; "
+                    "exitHooks+=(_nix_shell_clean_tmpdir); "
+                    "failureHooks+=(_nix_shell_clean_tmpdir); ":
```
xzfc: [stdenv setup] overrides the trap with `exitHandler` function which calls `exitHook` or `failureHook`. I've put the trap at line 441 as a fallback.

[stdenv setup]: https://github.com/NixOS/nixpkgs/blob/39700a65ef75e473f46270e07238cabff0462631/pkgs/stdenv/generic/setup.sh#L113-L149

src/nix-build/nix-build.cc
```diff
@@ -67,6 +67,19 @@ std::vector<string> shellwords(const string & s)
     return res;
 }
 
+static string shellquote(string s) {
```
edolstra: There is a function named `shellEscape` in `util.cc` that appears to do the same thing.

xzfc: Resolved.

PR #3344, opened 2020-02-03
ssh-store: add remote-store and remote-program query params

LnL7:
Brings the functionality of ssh-ng:// in sync with the legacy ssh://
implementation.  Specifying the remote store uri enables various useful
things. eg.

    $ nix copy --to ssh-ng://cache?remote-store=file://mnt/cache --all

### DIFF
--- a/src/libstore/ssh-store.cc
+++ b/src/libstore/ssh-store.cc
@@ -16,6 +16,8 @@ public:
 
     const Setting<Path> sshKey{(Store*) this, "", "ssh-key", "path to an SSH private key"};
     const Setting<bool> compress{(Store*) this, false, "compress", "whether to compress the connection"};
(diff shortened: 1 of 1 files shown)


src/libstore/ssh-store.cc
```diff
@@ -82,7 +84,9 @@ ref<FSAccessor> SSHStore::getFSAccessor()
 ref<RemoteStore::Connection> SSHStore::openConnection()
 {
     auto conn = make_ref<Connection>();
-    conn->sshConn = master.startCommand("nix-daemon --stdio");
+    conn->sshConn = master.startCommand(
```
LnL7: Might not be worth the effort, but this could be generalised between the 2 implementations to avoid some of duplication.

PR #3346, opened 2020-02-06
Make lockToCurrentCPU re-entrant

### LINKED ISSUE #3345: nix-shell inherits single CPU affinity when derivation has import from derivation
As title says. A `nix-shell` for the following derivation will have its affinity set to a single core:

``` nix
with import <nixpkgs> {};
stdenv.mkDerivation
  {
    src = lib.cleanSourceWith
      {
        src = ./.;
        filter = _: _: builtins.hasAttr "bar" (builtins.readDir hello);
      };
    name = "foo";
  }
```

``` console
$ nix-shell --run nproc
1
```

There's probably a `restoreAffinity()` missing somewhere.

layus:
Fixes #3345

### DIFF
--- a/src/libutil/affinity.cc
+++ b/src/libutil/affinity.cc
@@ -10,6 +10,8 @@ namespace nix {
 
 
 #if __linux__
+// lockToCurrentCPU
+static bool didLockCPU = false;
 static bool didSaveAffinity = false;
 static cpu_set_t savedAffinity;
 #endif
@@ -18,6 +20,12 @@ static cpu_set_t savedAffinity;
 void setAffinityTo(int cpu)
 {
 #if __linux__
+    if (didSaveAffinity) {
+        printError("setAffinity cannot be re-entered");
+        // Just return. Failing to set affinity only impacts performance, not
+        // validity of nix operations and build results.
+        return;
+    }
     if (sched_getaffinity(0, sizeof(cpu_set_t), &savedAffinity) == -1) return;
     didSaveAffinity = true;
     debug(format("locking this thread to CPU %1%") % cpu);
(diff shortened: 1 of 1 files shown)


layus:
I guess there could be an issue if some part of the program forks. I think nix may use pthreads, but I do not think it works with multiple processes. In that setup, the above patch works correctly.

The example you describe in #3345 is not an IFD (or did I miss something ?). But you use a builtin that loads local filesystem data (filterSource & related). This works a bit differently than other derivations indeed.

nmattia:
It is unfortunately IFD related, if the argument to `readDir` is e.g. a local directory (and not a derivation) the shell gets the correct number of CPUs.

layus:
Oh, I did not look closely enough.

I did check that it works on the example you provided. So I can tell that it fixes your use case. But I do not know how nix forks and threads itself.

That being said, the comment is quite clear about this being "only" a cpu cache optimization. So the change cat at worst miss an optimization opportunity.

nmattia:
@edolstra WDYT?

src/libutil/affinity.cc
```diff
@@ -34,7 +42,13 @@ int lockToCurrentCPU()
...
 #if __linux__
     int cpu = sched_getcpu();
-    if (cpu != -1) setAffinityTo(cpu);
+    // `lockToCurrentCPU` is re-entrant, because it is idempotent.
+    // But we have to avoid calling `setAffinity` twice,
+    // as that would corrupt the `savedAffinity`.
+    if (!didLockCPU && cpu != -1) {
+        didLockCPU = true;
+        setAffinityTo(cpu);
+    }
```
regnat, line 51: I’m not sure I understand what the reason for this. Since `setAffinityTo` is pseudo-reentrant now, this code could stay unchanged (and we could avoid the need for the `didLockCPU` variable altogether), no?

regnat (changes requested):
:+1: on the idea. A couple small comments though.

Also, can you add a regression test for #3345 ? Either in `tests/nix-shell.sh` or as its own thing (unless it’s too complex to test properly, but I don’t think it should be)

PR #869, opened 2016-04-01
Enable custom nix-shell prompts

zopa:
Allow overriding the nix-shell prompt by setting NIX_SHELL_PS1.

My use-case: I often switch workspaces after firing off a long-running process, and it's nice to be notified when the process completes. So I have a bell character in my PS1, which, through a chain of triggers, highlights the workspace in my status bar. This pull request lets me do the same with and within `nix-shell`.

PR #3349, opened 2020-02-09
Allow PS1 to be customized in nix-shell

neilmayhew:
Currently there's no way to customize the `nix-shell` prompt permanently. This change allows the variable `NIX_SHELL_PS1` to be set (not exported) in `.bashrc` and it will then be used to set `PS1` in `nix-shell`.

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nix/3349)


edolstra:
Could you mention this variable in `doc/manual/command-ref/env-common.xml`?

nomeata:
This would fix https://github.com/NixOS/nix/pull/869, right?

domenkozar:
Ping @neilmayhew

PR #3350, opened 2020-02-10
Remove #[macro_use]

curiousleo:
As of Rust 2018, `#[macro_use]` is no longer required in most circumstances, we can just `use` macros the same way as functions. I think it is generally a good idea to remove `#[macro_use]` when not needed, to stop macros from polluting the crate's global namespace.

https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macro_rules-style-macros


edolstra:
Thanks!

PR #3351, opened 2020-02-10
Implement std::error::Error and document error variants

curiousleo:
Rust errors should implement `std::error::Error` for interoperability.
https://rust-lang.github.io/api-guidelines/interoperability.html#c-good-err

In addition, public modules and definitions should be documented. Since I was looking at this module anyway, I made a start here.
https://rust-lang.github.io/api-guidelines/documentation.html

**Before:**
Screenshot_2020-02-10 nixrust Error - Rust(1)

**After:**
Screenshot_2020-02-10 nixrust Error - Rust

### DIFF
--- a/nix-rust/src/error.rs
+++ b/nix-rust/src/error.rs
@@ -1,30 +1,55 @@
+//! Error types
+
 use std::fmt;
 
+/// An error that can occur in this crate.
 #[derive(Debug)]
 pub enum Error {
+    /// The store path could not be resolved.
     InvalidPath(crate::store::StorePath),
+    /// The store path is in the wrong format.
     BadStorePath(std::path::PathBuf),
+    /// The store path does not exist in the store.
     NotInStore(std::path::PathBuf),
+    /// The NAR info is in the wrong format.
     BadNarInfo,
+    /// Base32 string is invalid.
     BadBase32,
+    /// The store path name is empty.
     StorePathNameEmpty,
+    /// The store path name is longer than 211 characters.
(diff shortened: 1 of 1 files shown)


edolstra:
This conflicts with some refactoring to `Error` that I've made (but haven't finished yet) to split it into multiple `Error` types (and maybe use error_chain or failure, but I haven't decided on that yet).

nix-rust/src/error.rs
```diff
@@ -1,30 +1,55 @@
+//! Error types
+
 use std::fmt;
 
+/// An error that can occur in this crate.
 #[derive(Debug)]
 pub enum Error {
+    /// The store path could not be resolved.
```
edolstra: IMHO this isn't really desirable since it duplicates the error descriptions in the `fmt::Display` trait implementation.

curiousleo: I see what you mean. The error variants have very explicit names too, so perhaps these docs don't add much.

curiousleo:
> This conflicts with some refactoring to `Error` that I've made (but haven't finished yet) to split it into multiple `Error` types (and maybe use error_chain or failure, but I haven't decided on that yet).

Okay, good to know! I'll close this PR then.

Ekleog:
@edolstra FWIW, error-chain is long dead, and failure has been halfway merged into the stdlib -- the part that wasn't is now available in `thiserror`, which is as a consequence the crate new projects should probably go to by default.

Issue #2395, opened 2018-09-02
Use of PR_SET_PDEATHSIG results in Broken pipe errors when using SSH substituters


nh2:
## Problem

Nix 2.0.4.

Sometimes (on my desktop, often, namely in ~66% of the cases), nix (on Ubuntu 16.04) fails when fetching stuff from my binary cache (NixOS 18.03).

Message:

```
copying path '/nix/store/4nny4bm8aflicjalfnrfbkfdd5b54nv3-conduit-1.2.13.1.tar.gz' from 'ssh://nix-store-user@mymachine.example.com'...
writing to file: Broken pipe
```

It's nondetemrinistic, so trying it multiple times usually helps, but it is still very annoying.

## Investigation

`strace`'d the failure, saw:

```
6837  write(10<pipe:[7624621]>, "\3\0\0\0\0\0\0\0C\0\0\0\0\0\0\0/nix/store/4nny4bm8aflicjalfnrfbkfdd5b54nv3-conduit-1.2.13.1.tar.gz\0\0\0\0\0", 88) = -1 EPIPE (Broken pipe)
```

Searched for the pipe number `7624621` further up, found (starting from the first occurrence):

```
6830  pipe2([9<pipe:[7624621]>, 10<pipe:[7624621]>], O_CLOEXEC) = 0
6830  pipe2([11<pipe:[7624622]>, 12<pipe:[7624622]>], O_CLOEXEC) = 0
6830  vfork( <unfinished ...>
6833  prctl(PR_SET_PDEATHSIG, SIGKILL)  = 0
6833  rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
6833  close(10<pipe:[7624621]>)         = 0
6833  close(11<pipe:[7624622]>)         = 0
6833  dup2(9<pipe:[7624621]>, 0</dev/pts/12>) = 0<pipe:[7624621]>
```

Noticed the `prctl(PR_SET_PDEATHSIG, SIGKILL)  = 0` in there.

Looked up https://stackoverflow.com/questions/10761197/prctlpr-set-pdeathsig-signal-is-called-on-parent-thread-exit-not-parent-proc for details.

Grepped for `PR_SET_PDEATHSIG` in `nix`'s code, found:

https://github.com/NixOS/nix/blob/54996b51fba46e8f287b1ac6a58854e62b476985/src/libutil/util.cc#L926-L927

Commented out the lines:

```c++
[...]

PR #3353, opened 2020-02-10
Fix PR_SET_PDEATHSIG results in Broken pipe (#2395)

tbsmoest:
Fixes: #2395

The ssh client is lazily started by the first worker thread, that
requires a ssh connection. To avoid the ssh client to be killed, when
the worker process is stopped, start the ssh client in a long-lived
synchronous worker thread.

Add the class SyncWorker, that provides a single thread in which tasks
can be executed synchronously.

### DIFF
--- a/src/libstore/ssh.cc
+++ b/src/libstore/ssh.cc
@@ -33,6 +33,9 @@ std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string
     out.create();
 
     auto conn = std::make_unique<Connection>();
+    ProcessOptions options;
+    options.dieWithParent = false;
+
     conn->sshPid = startProcess([&]() {
         restoreSignals();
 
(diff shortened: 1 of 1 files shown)


edolstra:
Hm, can't we just *not* set PR_SET_PDEATHSIG?

tbsmoest:
We could and it would solve the ssh issue. However, I'm not sure whether the child cleanup will work in all cases then. As the PR_SET_PDEATHSIG was explicitly introduced to guarantee the proper cleanup, I was assuming it is still needed. I will check whether it has become redundant in the meantime.

tbsmoest:
It seems, that it is save to not set PR_SET_PDEATHSIG for the SSH fork. The signals SIGINT and SIGTERM are forwarded. SIGKILL is not a problem either, as all children are killed, too. SIGSTOP causes issues with orphaned children when it is used in combination with SIGINT, but that is not solved by PR_SET_PDEATHSIG either.

I have updated the PR accordingly.

tbsmoest:
@edolstra, I have removed my initial solution and now simply deactivated the PR_SET_PDEATHSIG for ssh. I this ok with you?

edolstra:
Thanks!

nh2:
@chpatrick made a backport of this fix for `nix-2.3.2` which is on NixOS 20.03: https://github.com/chpatrick/nix/commit/ebaba525fc879f9ee3461b07ad03cd2111d27479

PR #3355, opened 2020-02-12
remove builtins.exec

andir:
IMO that feature is very dangerous to have in Nix and probably doesn't fit the entire model at all.

It currently isn't documented so there shouldn't be many users. Those that need it can probably add this via a plugin. The plugin example (`tests/plugins/plugintest.c`) already shows how to add such a plugin and the removed code can probably just pasted into that file to get started.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -193,45 +193,6 @@ void prim_importNative(EvalState & state, const Pos & pos, Value * * args, Value
 }
 
 
-/* Execute a program and parse its output */
-void prim_exec(EvalState & state, const Pos & pos, Value * * args, Value & v)
-{
-    state.forceList(*args[0], pos);
-    auto elems = args[0]->listElems();
-    auto count = args[0]->listSize();
-    if (count == 0) {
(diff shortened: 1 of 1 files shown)


shlevy:
builtins.exec is already hidden behind a very verbose flag, and allows for a much nicer extension experience without any native code. What is the concern?

grahamc:
People see `builtins.exec` and think "ooooh!", and then nearly nothing persuades them against it.

shlevy:
Are there real examples of people shooting themselves in the foot or failing to do things the right way due to this?

flokli:
It's a massive opportunity to shoot yourself in the foot.

It's undocumented, https://github.com/NixOS/nix/commit/0bb8db257d98a32abde759f4d07d28b5178bd3bf mentioned it was meant as an intermediate thing, so I think if people want it, moving it to an optional plugin would be the right move.

zimbatm:
Would removing it solve a actual current problem? This provides an escape hatch to do funny and crazy things. It's properly protected behind a scary flag.

edolstra:
Closing since it doesn't really harm anybody as it's hidden behind a flag.

Ericson2314:
I would put this behind an experimental feature.

PR #3357, opened 2020-02-14
Pass through http proxy env vars in pure shell

carlosdagos:
Allow a pure nix-shell to inherit environment variables to signal some programs (such as `curl`) to use a proxy. This is quite common in eg corporate environments. 

Issue arises for me whilst using `stack`, I opened a PR over there as well, but maybe here it can be solved in a more general way. Reference: https://github.com/commercialhaskell/stack/pull/5171. As mentioned there, possible solutions are to add those variables to something like `~/.bashrc`, which is totally fine too of course. However, maybe these vars are prevalent enough to warrant their inclusion in a pure nix shell.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -97,7 +97,11 @@ static void _main(int argc, char * * argv)
     std::string outLink = "./result";
 
     // List of environment variables kept for --pure
-    std::set<string> keepVars{"HOME", "USER", "LOGNAME", "DISPLAY", "PATH", "TERM", "IN_NIX_SHELL", "TZ", "PAGER", "NIX_BUILD_SHELL", "SHLVL"};
+    std::set<string> keepVars{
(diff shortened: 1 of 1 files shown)


samueldr:
Added in #2323, might be the more general way:

```
       --keep name
           When a --pure shell is started, keep the listed environment
           variables.
```

```
~ $ export HI="hello there!"

~ $ nix-shell -p stdenv --pure --run 'set -eu; echo $HI'
/tmp/nix-shell-27378-0/rc: line 1: HI: unbound variable

~ 1 $ nix-shell -p stdenv --pure --run 'set -eu; echo $HI' --keep HI
hello there!

```

carlosdagos:
Hey @samueldr, thanks for your feedback. I didn't know about #2323, that one is interesting. However I think that's not what I'm after, as using the `--keep` flag would still require different software to support http proxy env vars, whereas this change would just by default support them.

PR #3364, opened 2020-02-19
[do not merge] tarball: Workaround bad branch names

samueldr:
Currently, from a git checkout, using `nix-build` fails like this for
me:

```
[...]
  GEN    nix-2.4pre0_dirty.tar.bz2
tar: .git/logs/refs/remotes/origin/mojave-/homeless-shelter: Cannot stat: No such file or directory
tar: .git/refs/remotes/origin/mojave-/homeless-shelter: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
make: *** [mk/dist.mk:8: nix-2.4pre0_dirty.tar.bz2] Error 2
```

As can be deduced from the two `No such file or directory`, the files
listing in `.dist-files` gets expanded once used by `tar`.

The branch name is `mojave-$HOME`.

This is a *dumb workaround* rather than a proper fix.

Fixing this by escaping using `sed` to double-up the `$` fails like this:

```
make: *** No rule to make target '.git/logs/refs/remotes/origin/mojave-$$HOME', needed by 'nix-2.4pre0_dirty.tar.bz2'.  Stop.
```

And trying to escape it as `\$` fails like this:

```
make: *** No rule to make target '.git/logs/refs/remotes/origin/mojave-\$HOME', needed by 'nix-2.4pre0_dirty.tar.bz2'.  Stop.
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -228,7 +228,14 @@
             buildInputs = tarballDeps ++ buildDeps ++ propagatedDeps;
 
             postUnpack = ''
-              (cd $sourceRoot && find . -type f) | cut -c3- > $sourceRoot/.dist-files
+              (cd $sourceRoot && find . -type f) | grep -v '\$' | cut -c3- > $sourceRoot/.dist-files
+              #                                    ------------   --------
+              #                                        \              \
+              #                                         \       Strips the ./ prefix
+              #                                          \
+              #               Ensures we don't get files that would get wrongly expanded by make
+              #                   this is the case if there is e.g. the mojave-$HOME branch.
+
               cat $sourceRoot/.dist-files
             '';
 


edolstra:
I guess this is because you're using `nix-build`. A flake build (i.e. `nix build`) ensures a clean source tree (no .git or "dirty" files).

samueldr:
Oh, weird, I tried a simpler "fix" at first, where I deleted `.git` to get a clean slate, but I must have done it in a wrong way is it failed. Though now I try again and it works. I'm confused.

I see, so flake knows about sources of dirt? What kind of sources of dirt are recognized by `flake`, and where is the source/documentation about those cleanups?

I'm fine with this not being merged in that case, feel free to close. I'll add filtering to the compatibility shim, but would like to know what kind of "dirt" is handled by flakes.

samueldr:
I won't edit the comment, just in case, but I'm re-reading myself and I sure sound annoyed. I'm merely surprised, and curious to read about the change proper.

edolstra:
It's actually the same behaviour as `fetchGit` on a dirty git tree: it copies only the files tracked by Git to the Nix store.

edolstra:
So you should be able to simulate the flake behaviour by doing something like `fetchGit ./.` for the top-level source tree. That will also give you the correct `rev` and `revCount` attributes (at least for a clean tree).

PR #3365, opened 2020-02-19
Re-add release-common.nix to flakes branch

Ma27:
Needed to build `nix` from the flakes branch when not using flakes
globally. Otherwise, `shell.nix` doesn't eval.

### DIFF
--- a/release-common.nix
+++ b/release-common.nix
@@ -1,0 +1,82 @@
+{ pkgs }:
+
+with pkgs;
+
+rec {
+  # Use "busybox-sandbox-shell" if present,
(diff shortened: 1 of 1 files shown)


edolstra:
Note that `shell.nix` is gone (it had in fact reappeared inadvertently). 50cf77cecd6ed2c98fab1651e54f7a8dcc162db6

If you need to bootstrap the flakes branch, you can use `nixFlakes` from Nixpkgs or get it from the binary cache.

Ma27:
Thanks a lot for the clarification! I guess this PR is obsolete then -> closing :)

PR #3366, opened 2020-02-19
[WIP] Add a compatibility shim for non-flake nix

samueldr:
This compatibility shim uses an extremely naïve parsing of `flake.lock`
to get the dependencies.

It then uses a naïve implementation of flakes to produce the build.

* * *

## Intent

I am sharing the current state of the shim, half-expecting that there are things to fix.

With a flakes-free nix, this handles:

 * using `nix build` or `nix-build` to build.
 * using `nix-shell`.

It is my opinion that this, or another implementation of such a shim, should be present in the flakes branch, and *then* in the master branch, up until a full release cycle of nix has had flakes enabled by default.

Not doing so makes for a confusing experience for getting started in looking at the flakes branch, at first. Then, allowing this for one full release might not be as useful at that point, but will allow laggards to at least be able to use this repository with `nix v"flakes-1"`.

I would recommend, once flakes get into master, assuming there is such a shim, that the warning documents that this is going to be dropped in `nix v"next release"`, also linking to a proper location for upgrades instructions.

## Implementation details

This implementation is probably wrong in some ways. It does work for nix, but I wouldn't assume it works for all flakes. Surely, it would be helpful to include such a shim in all "core" flakes around the NixOS organization, if it can be fixed to be good enough for all of our flakes.

I am open to critique, rework, everything to make this better for inclusion in nix. Tell me what I did wrong, I'll be glad to get it sorted out.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,58 @@
+let
+  # Fields we care about when parsing, in order they'll be merged.
+  flakeFields = [
+    "info"
+    "inputs"
+    "locked"
+    "original"
+  ];
+
+  # The lock information.
+  flakeLockInfos = builtins.fromJSON (builtins.readFile ./flake.lock);
+
+  # This is a watered-down version of flake.lock parsing.
+  # This is probably somewhat wrong.
+  inputsInfo = builtins.mapAttrs (input: data:
+    let
+      # Ensure we have at least empty attrsets for each fields.
+      data' = (
+        (builtins.listToAttrs (builtins.map (name: { inherit name; value = {}; }) flakeFields))
+        // data
+      );
+    in
+    builtins.foldl' (coll: next: coll // next) {}
+      (builtins.map (attrname: data'.${attrname}) flakeFields)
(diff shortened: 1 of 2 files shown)


edolstra:
Maybe this can be moved into a separate repo so it can be used by other projects? I.e. `shell.nix` would look something like
```
(import (fetchTarball https://github.com/NixOS/flake-compat-shim.git) {
  flake = ./flake.nix;
  lockFile = ./flake.lock;
}).devShell.${builtins.currentSystem}
```

zimbatm:
Related to this topic, I also have this [nix/sources.nix](https://gist.github.com/zimbatm/32fdffd634c75d621120a30380f38a7d) which can be used if you want to switch from [niv](https://github.com/nmattia/niv) to flakes.

edolstra:
I've added a default.nix and shell.nix that calls https://github.com/edolstra/flake-compat. It's similar to @zimbatm's sources.nix but uses the latest lock file format (which is simpler since it doesn't require parsing of flake URLs).

PR #3368, opened 2020-02-20
.dir-locals.el: Set additional lambda indentation to zero

tbsmoest:
To allow emacs to handle the currently used lambda indentation correctly.


edolstra:
Thanks, this has annoyed me for a long time :-)

PR #3375, opened 2020-02-26
install-multi-user: allow overriding user count

domenkozar:
Motivation at https://github.com/cachix/install-nix-action/issues/13

Open question is if it should have been a flag like `--daemon` and if it needs to be documented somewhere.


edolstra:
A flag seems nicer to me. Environment variables don't have good discoverability.

domenkozar:
I've made it into a flag

Issue #3288, opened 2019-12-22
nix-env --query --attr-path shows unrelated number instead of attribute path


bcc32:
Running `nix-env --query --attr-path` shows a number next to each package rather than the attribute path I would have expected:

```
$ nix-env -qP
5  binutils-wrapper-2.30
7  direnv-2.15.1
8  fasd-unstable-2016-08-11
3  nethogs-0.8.5
0  perl5.30.0-local-lib-2.000024
2  pv-1.6.6
6  rhash-2018-02-05
1  rxvt-unicode-with-perl-with-unicode3-9.22
4  speedtest-cli-2.0.0
```

The numbers appear to be sequential, decreasing in order of most recently installed packages.  It's not clear why these numbers are displayed, nor how to get it to show the package attribute path instead.

By the way, I'm running:

```
$ nix-env --version
nix-env (Nix) 2.3
```

zimbatm:
the numbering seems to be the position in the list that is defined in `manifest.nix` . See `/nix/var/nix/profiles/per-user/$USER/profile/manifest.nix` where $USER is your username.

PR #3380, opened 2020-03-01
display attr-path only when queried available

contrun:
Currently the option `--attr-path` is not relevant when queried with `--installed`. We have no way to know what the original attribute path is. This pull request remove the display of the misleading `attr-path` obtained from `manifest.nix`.

Closes https://github.com/NixOS/nix/issues/3288


edolstra:
It's probably better to throw an error if `-P` is used with `-q`.

contrun:
Thanks. I've updated the commit to throw an error.

rben01:
Is it possible to get the attribute path that generated an installed package? E.g., `python3-3.8.7` → `nixpkgs.python38Full`

contrun:
It would be a huge improvement in usability, but I don't think this information is stored.

Issue #2980, opened 2019-07-04
Make nix-channel --list show inherited channels


matthiasbeyer:
It would be nice to list "inherited" channels (as in channels the root user has set) when running `nix-channel --list` as user.

tomberek:
Closing as stale. Re-open if needed.

PR #3381, opened 2020-03-01
list also channels created by root

contrun:
Closes https://github.com/NixOS/nix/issues/2980

### DIFF
--- a/src/nix-channel/nix-channel.cc
+++ b/src/nix-channel/nix-channel.cc
@@ -14,12 +14,12 @@ typedef std::map<string,string> Channels;
 
 static Channels channels;
 static Path channelsList;
+static Path rootChannelsList;
 
-// Reads the list of channels.
-static void readChannels()
+static void doReadChannels(Path path)
 {
-    if (!pathExists(channelsList)) return;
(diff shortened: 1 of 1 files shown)


edolstra:
I'm not sure we should do this, for a couple of reasons:

* Channels are basically deprecated, so it's probably better not to change their behaviour at this point.
* It could be confusing that nix-channel silently merges different channel lists. (Cf `nix flake list`, which shows you whether a flake comes from the global, local or user registries.)
* If I read this correctly, it causes commands like `nix-channel --add` to merge the contents of the user and root channels *and* rewrite both to have the same contents. That doesn't seem right.

contrun:
Thanks for the comments. My method is indeed ad hoc.

Only when running by the root user, it writes the channel information of the root user to `/nix/store/profiles/per-user/root/nix-channels`. This way ordinary users can obtain the channel information of root. The channel information of ordinary users are not written to `/nix/store/profiles/per-user/root/nix-channels`.

Now that channels are being deprecated, feel free to close this PR. Or if improvement on this commit is desirable, I can add a `--root` flag to the list operation.

PR #3388, opened 2020-03-07
Assertion syntax in error in the manual?

mayl:
I'm a complete neophyte to Nix, but based on the example in the manual and looking in `nixpkgs` a bit it seems like the actual `assert` syntax is `assert e1 -> e2`  not `assert e1; e2`.  If I've got this garbled, please ignore of course, but even if so I think it's still worth pointing out that for a new learner this explanation and example _seem_ contradictory as written and could probably be clarified.

### DIFF
--- a/doc/manual/expressions/language-constructs.xml
+++ b/doc/manual/expressions/language-constructs.xml
@@ -271,7 +271,7 @@ evaluate to a Boolean value (<literal>true</literal> or
 on or between features and dependencies hold.  They look like this:
 
 <programlisting>
-assert <replaceable>e1</replaceable>; <replaceable>e2</replaceable></programlisting>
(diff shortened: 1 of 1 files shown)


LnL7:
While very often used together, `assert` and `->` (implication) are separate things.

```
nix-repl> assert true; 42
42

nix-repl> assert false; 42
error: assertion 'false' failed at (string):1:1
```

```
nix-repl> true -> 1 > 2
false
nix-repl> false -> 1 > 2
true
```

mayl:
Ok, I see.  Thanks for clarifying.  I guess part of what made this confusing for me is that the implication operator isn't introduced until after `assert` is described, and it's every assert example uses it and without further explanation.

Issue #3398, opened 2020-03-09
pathInfoCache does not respect narinfo-cache-negative-ttl


roberth:
According to the manual:

**narinfo-cache-negative-ttl**
The TTL in seconds for negative lookups. If a store path is queried from a substituter but was not found, there will be a negative lookup cached in the local disk cache database for the specified duration.

### The problem

While this isn't false, there's also an in-memory LRU cache, `Store.pathInfoCache` that doesn't expire until many other paths are used.
This is problematic when `nix-daemon` is caching missing binary cache entries for distributed build processes (not managed by the daemon), because we don't have a sane way to clear that cache.

### Solution

Turn the LRU cache values into tuples of time and `shared_ptr<ValidPathInfo>`. This should make things work.

Ideally, in order to be able to use a ttl > 0, add a method to clear a closure from the caches in all stores, including the default substituters, accessible via the daemon protocol.

### Relevant code

Every store, including binary caches have an in-memory cache for path validity

https://github.com/NixOS/nix/blob/d37dc71e3cf077fa5d24a9bf8395deae21cc4410/src/libstore/store-api.hh#L267

Once queried, the result sticks around as long as no huge number of other paths is queried (which is the case). [LruCache](https://github.com/NixOS/nix/blob/1f64f4c7c83d8efd67051a399c738c71815e0594/src/libutil/lru-cache.hh) has no concept of expiry other than exceeding capacity. 

Store::isValidPath returns false if the `shared_ptr` from the LRU cache is null.

https://github.com/NixOS/nix/blob/f8abbdd4565542464f31f4dc203a9c3e091b3536/src/libstore/store-api.cc#L247

[queryPathInfo](https://github.com/NixOS/nix/blob/f8abbdd4565542464f31f4dc203a9c3e091b3536/src/libstore/store-api.cc#L315) behaves the same.

edolstra:
This is why hydra-queue-runner periodically calls `localStore->clearPathInfoCache()`.

roberth:
Hercules CI agent does that too, but we're still dealing with stale negative cache entries. Does it clear the `nix-daemon` cache too?

edolstra:
No, it doesn't clear the daemon's cache, I think.

roberth:
Ok, do you think putting timestamps in the lru cache is a good idea?

edolstra:
Yes, that would be a good idea.

edolstra:
Fixed in #3403.

PR #3403, opened 2020-03-11
pathInfoCache: Respect disk cache TTLs #3398

roberth:


src/libstore/store-api.hh
```diff
@@ -261,10 +262,40 @@ public:
...
+        // Null if missing
+        std::shared_ptr<const ValidPathInfo> value;
+
+        // Whether the value is valid as a cache entry. The path may not exist.
+        bool isKnownNow();
+
+        // Past tense, because a path can only be assumed to exists when
+        // isKnownNow() && didExist()
+        inline bool didExist() {
+          return value != 0;
```
edolstra: => `nullptr`.

src/libstore/store-api.hh
```diff
@@ -261,10 +262,40 @@ public:
 
 protected:
 
+    struct PathInfoCacheValue {
+
+        PathInfoCacheValue(std::shared_ptr<const ValidPathInfo> value) :
+            time_point(std::chrono::steady_clock::now()),
+            value(value) {}
```
edolstra: Since we're using C++20 now, constructors for "trivial" data classes can be omitted using designated initializers, e.g.
```
PathInfoCacheValue { .value = some_value }
```
and `time_point` could have a default initializer:
```
         std::chrono::time_point<std::chrono::steady_clock> time_point = std::chrono::steady_clock::now();
```

src/libstore/store-api.cc
```diff
@@ -301,11 +309,11 @@ void Store::queryPathInfo(const StorePath & storePath,
...
         {
             auto res = state.lock()->pathInfoCache.get(hashPart);
-            if (res) {
+            if (res && res->isKnownNow()) {
                 stats.narInfoReadAverted++;
-                if (!*res)
+                if (!res->didExist())
                     throw InvalidPath("path '%s' is not valid", printStorePath(storePath));
-                return callback(ref<const ValidPathInfo>(*res));
+                callback(ref<const ValidPathInfo>(res->value));
```
edolstra: Hm, removing the return looks potentially wrong...

roberth: Yes it was. The compiler was complaining about `return` in a `void` function. I've re-added the return and then it works.

roberth:
Ready for review. I'll make a PR with the previous variation for a 2.3 backport.

PR #3405, opened 2020-03-12
pathInfoCache: Respect disk cache TTLs #3398 (2.3 backport)

roberth:
Backport of #3403 without C++20

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -104,7 +104,7 @@ void BinaryCacheStore::writeNarInfo(ref<NarInfo> narInfo)
 
     {
         auto state_(state.lock());
-        state_->pathInfoCache.upsert(hashPart, std::shared_ptr<NarInfo>(narInfo));
+        state_->pathInfoCache.upsert(hashPart, PathInfoCacheValue(std::shared_ptr<NarInfo>(narInfo)));
     }
 
     if (diskCache)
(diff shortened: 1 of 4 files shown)


edolstra:
Hm, this might be a bit too big of a chance for the stable branch... It's also more of a feature rather than a bug fix.

roberth:
Caching more would be risky, but this caches less.

> narinfo-cache-negative-ttl
>
>    The TTL in seconds for negative lookups. If a store path is queried from a substituter but was not found, there will be a negative lookup cached in the local disk cache database for the specified duration.

It does not mention an additional long-lived cache in `nix-daemon` that can't be cleared.

If the implementation doesn't conform to the docs, isn't that a bug?

edolstra:
Nix doesn't have any long-lived processes itself though, so it's not really a bug.

roberth:
Isn't `nix-daemon` a long-lived process that is part of Nix?

edolstra:
All the real work is done by the `nix-daemon` child processes which exist only as long as the connection.

roberth:
Will Nix 2.4 be released for NixOS 20.09?

Ericson2314:
@roberth you mean 3.0? I think we could make that.

roberth:
@Ericson2314 branch-off for 20.09 is in about a week. Is that too soon or is the goal to stabilize it during the release month?

roberth:
I still have to ask users to patch their system Nix, while this patch has run in production for over half a year. Can it be merged?

roberth:
Thanks!

PR #3408, opened 2020-03-13
fetchGit: allow fetching revs which are not ancestors of HEAD without specifying a ref

basvandijk:
Note that `builtins.fetchGit` defaults `ref` to `HEAD` when not specified. This is problematic when fetching a `rev` which is not an ancestor of `HEAD` when no `ref` is available:

```
nix-instantiate -v --eval -E \
'builtins.fetchGit { 
  url = "https://github.com/.../..."; 
  rev = "9ec3a6973f6a17dc4b2566520253fc6203f75ac6"; 
}'
evaluating file '/nix/store/lrnvapsqmf0ja6zfyx4cpxr7ahdr7f9b-nix-2.3.3/share/nix/corepkgs/derivation.nix'
fetching Git repository 'https://github.com/.../...'...
using revision 9ec3a6973f6a17dc4b2566520253fc6203f75ac6 of repo 'https://github.com/.../...'
fatal: not a tree object: 9ec3a6973f6a17dc4b2566520253fc6203f75ac6
```

This error occurs because `rev` is never fetched since it's not an ancestor of the ref `HEAD` which is fetched.

This commit introduces the `fetchRevInsteadOfRef` argument to `builtins.fetchGit` which when `true` fetches the `rev` directly instead of fetching `HEAD`.

`fetchRevInsteadOfRef` defaults to `false` to keep backwards compatibility and since git servers by default [don't allow fetching revisions directly](https://www.git-scm.com/docs/git-config/2.25.1#Documentation/git-config.txt-uploadpackallowAnySHA1InWant) (Note that GitHub does allow this).

This commit now also checks whether `rev` is an ancestor of `ref` when the latter is specified.

### DIFF
--- a/doc/manual/expressions/builtins.xml
+++ b/doc/manual/expressions/builtins.xml
@@ -422,6 +422,28 @@ stdenv.mkDerivation { … }
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>fetchRevInsteadOfRef</term>
+          <listitem>
+            <para>
+              Defaults to <literal>false</literal>.
+            </para>
+
+            <para>
+              When set to <literal>true</literal> the specified
+              <varname>rev</varname> is fetched <emphasis>directly</emphasis>
+              instead of fetching the <varname>ref</varname> (which defaults to
+              <literal>HEAD</literal> when not specified). This is useful in
+              case a <varname>ref</varname> is not available.
+            </para>
+
+            <para>
+              Do note that git servers by default <link
+              xlink:href="https://www.git-scm.com/docs/git-config/2.25.1#Documentation/git-config.txt-uploadpackallowAnySHA1InWant">don't allow fetching revisions directly</link>
+              (GitHub does allow it).
+            </para>
+          </listitem>
+        </varlistentry>
       </variablelist>
 
       <example>
--- a/src/libexpr/primops/fetchGit.cc
+++ b/src/libexpr/primops/fetchGit.cc
@@ -26,9 +26,80 @@ struct GitInfo
 
 std::regex revRegex("^[0-9a-fA-F]{40}$");
 
+void fetchRev(const Path cacheDir, const std::string & uri, std::string rev) {
+    Activity act(*logger, lvlTalkative, actUnknown, fmt("fetching '%s' of Git repository '%s'", rev, uri));
+
+    // FIXME: git stderr messes up our progress indicator, so
+    // we're using --quiet for now. Should process its stderr.
+    runProgram("git", true, { "-C", cacheDir, "fetch", "--quiet", "--", uri, rev });
+}
+
+void fetchRef(const Path cacheDir, const std::string & uri, std::string ref, time_t now, Path localRefFile) {
+    Activity act(*logger, lvlTalkative, actUnknown, fmt("fetching '%s' of Git repository '%s'", ref, uri));
+
+    // FIXME: git stderr messes up our progress indicator, so
+    // we're using --quiet for now. Should process its stderr.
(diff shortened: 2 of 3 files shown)


edolstra:
In the past we've decided that `fetchGit` should require a `ref` if you're pulling from a non-master branch. I don't really want to change that since it will requiring cloning an entire repository rather than a single branch.

basvandijk:
>  ... it will requiring cloning an entire repository rather than a single branch.

Why does it require cloning an entire repository?

Only fetching `rev` like I do [here](https://github.com/NixOS/nix/pull/3408/files#diff-7e39b3dccfe1a07ed3f4af7f4d8da628R175) should be enough no?

basvandijk:
Also when both specifying a `rev` and a `ref` I now check if `rev` is an ancestor of `ref`. [I'm careful](https://github.com/NixOS/nix/blob/abb4ffe055199d1d21a7ab4dd6c32304bf1d142c/src/libexpr/primops/fetchGit.cc#L141:L166) to only fetch `ref` when `rev` is not an ancestor of the already cached `ref`.

basvandijk:
BTW the reason I need this is that in `naersk` we need to [fetch crates with a git source](https://github.com/nmattia/naersk/blob/ee15d1214a8fb58967a24d629062e4ccdd9a925a/lib.nix#L91:L94). However since crates only specify a `rev` we don't have a `ref` available.

I currently work around this by [setting `ref = rev`](https://github.com/nmattia/naersk/blob/7d23ad7e33a92434b52129b16f1e2df2a9b8515e/lib.nix#L94:L102) but that results in ugly git warnings like:

```
warning: refname '9ec3a6973f6a17dc4b2566520253fc6203f75ac6' is ambiguous.
Git normally never creates a ref that ends with 40 hex characters
because it will be ignored when you just specify 40-hex. These refs
may be created by mistake. For example,

  git switch -c $br $(git rev-parse ...)

where "$br" is somehow empty and a 40-hex ref is created. Please
examine these refs and maybe delete them. Turn this message off by
running "git config advice.objectNameWarning false"
```

edolstra:
Git in general does not allow fetching an arbitrary revision from a remote for security reasons. (IIRC there is a server setting to disable this check.) You have to specify a ref that contains the revision.

basvandijk:
TIL that git servers indeed [don't allow fetching revisions directly by default](https://www.git-scm.com/docs/git-config/2.25.1#Documentation/git-config.txt-uploadpackallowAnySHA1InWant) (GitHub however does).

So if this is parameterizable on the server, what about making it parameterizable on the client as well?

I forced-pushed a proposal for this where I introduce the `fetchRevInsteadOfRef` argument to `builtins.fetchGit` (I'm open for other names). It defaults to `false` to keep backwards compatibility and to mimic the default on the server. When `fetchRevInsteadOfRef == true` `builtins.fetchGit` fetches the specified `rev` directly instead of fetching `ref`.

This keeps the old semantics, although with the additional ancestor checks, but allows fetching revisions directly.

@edolstra WDYT?

yorickvP:
@basvandijk GitHub does not allow fetching 'unadvertised refs' (one of the heads or tags, basically), which is the git default. We work around this by specifying `ref = "*";` where needed and setting `rev` to the revision, does that work for you? Alternatively, `fetchTarball` will work on non-private github repos and usually download less data.

nmattia:
@yorickvP what's the `ref="*"` about? Is that an instruction to nix to perform a full clone first? Or is that a git thing?

```
$ git fetch origin '*'
fatal: invalid refspec '*'
```

yorickvP:
It's a git thing, `git fetch origin 'refs/heads/*'`. Also see https://github.com/NixOS/nix/pull/2409

Issue #1559, opened 2017-09-06
Nix single user install on Linux wants 'nixbld' group, else - it fails


Anton-Latukha:
@edolstra
Setting-up with installer Nix in single-user mode on Linux crashes into error:
```
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
creating /root/.nix-profile
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
installing ‘nix-1.11.14’
building path(s) ‘/nix/store/gbvx4dcw6ibps16dx4wy4dxp3vf4i204-user-environment’
error: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
```
Reason probably in this block:
https://github.com/NixOS/nix/blob/0b606aad46e1d96da36d4831df63ad90f11d21c3/src/libstore/build.cc#L491-L499

Please, look into it.

There are 4 bug reports regarding this error:
2016-01-13. #936 Install Nix as root 
2016-03-13. #879 install.sh homepage instructions only work without build-user-group
2015-11-13. #697 how to correctly install Nix on Arch Linux
2015-11-10. #692 Installation in alternate directory fails

Workarounds people use are straight-forward.

Create 'nixbld' by hand.

Or as @mboes mentioned: 
```shell
echo "build-users-group =" > /etc/nix/nix.conf
```

As workarounds are simple, and there are 4 bugreports already - most people find that bugrepots, so not file new, and if person decide to proceed official install - they need to use mentioned workarounds at the moment.

It happens on any clean Linux distro. Myself I reproduce it with 100% on Ubuntu and Arch Linux and Alpine Linux.

This is why I came, I saw clear as day, that community is using NixOS, or they existing Nix setups, macOS or official Docker image. And general installation to Linux distributions is not polished.

As I am a Linux maniac and DevOps, I know Linux systems on level of all nuts&bolts, and how to make most universal install. I want to take that task on myself, to take a hand in release of flood on Linux distributions of the Nix beast.

Official install code is the best documentation for IT hacker guys and Linux distribution maintainers. It is easy to replicate then, if universal one is made. From that, ridiculously, packages of Nix easily can be made for all distributions, to infect their ecosystem. Users are much more happy to install package with package manager they trust, than run unknown script install. And with distributions are much easier to talk, when you have legit package for their distro.

[...]

Issue #936, opened 2016-06-13
Install Nix as root


davidak:
How do i install Nix (single-user mode) as root on Linux?

I don't want to do a full manual install (compile it).

I tried the default way, but it fails:

```
[root@ipfire ~]# curl https://nixos.org/nix/install | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1383  100  1383    0     0   4150      0 --:--:-- --:--:-- --:--:--  4190
unpacking Nix binary tarball for i686-linux from `https://nixos.org/releases/nix/nix-1.11.2/nix-1.11.2-i686-linux.tar.bz2'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 64.8M  100 64.8M    0     0  1438k      0  0:00:46  0:00:46 --:--:-- 1377k
warning: installing Nix as root is not supported by this script!
performing a single-user installation of Nix...
directory /nix does not exist; creating it by running ‘mkdir -m 0755 /nix && chown root /nix’ using sudo
copying Nix to /nix/store...........................
initialising Nix database...
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
creating /root/.nix-profile
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
installing ‘nix-1.11.2’
building path(s) ‘/nix/store/mk06d1iiyscqnr9zl23bf42jb2vds1gh-user-environment’
error: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
nix-binary-tarball-unpack.HoMEIazdBT/nix-1.11.2-i686-linux/install: unable to install Nix into your default profile
```

> warning: installing Nix as root is not supported by this script!

what is missing here? can i do the last steps manually?
could you support installing as root?

mboes:
I found a fairly simple workaround:

```
$ echo "build-users-group =" > /etc/nix/nix.conf
```

before running the install script.

this suggests that the `install` script could be modified so that it creates this configuration file to override the defaults in case the install user is root.

davidak:
`sh: you do not have 'shasum' installed, which I need to verify the binary tarball`

this must be a new dependency. i don't have it on that system :/

mboes:
[...]

Issue #697, opened 2015-11-13
install.sh homepage instructions only work without build-user-group


vizanto:
```sh
➭ sudo ./install
Password:
warning: installing Nix as root is not supported by this script!
performing a single-user installation of Nix...
copying Nix to /nix/store................................
initialising Nix database...
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
warning: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
installing ‘nix-1.10’
download-from-binary-cache.pl: could not download ‘https://cache.nixos.org/1p6vvyx1n5kgvhwv066qx0x0a358q69f.narinfo’ (Curl error 60)
building path(s) ‘/nix/store/1p6vvyx1n5kgvhwv066qx0x0a358q69f-user-environment’
error: the group ‘nixbld’ specified in ‘build-users-group’ does not exist
./install: unable to install Nix into your default profile
```

So, to fix this I do:
```
{18:06}[]~/nix-1.10-x86_64-darwin ➭ sudo mkdir /etc/nix/
{18:08}[]~/nix-1.10-x86_64-darwin ➭ sudo nano /etc/nix/nix.conf
```

```
➭ cat /etc/nix/nix.conf
build-users-group =
```

And now, install works:
```
{18:08}[]~/nix-1.10-x86_64-darwin ➭ sudo ./install
warning: installing Nix as root is not supported by this script!
performing a single-user installation of Nix...
copying Nix to /nix/store................................
initialising Nix database...
installing ‘nix-1.10’
download-from-binary-cache.pl: could not download ‘https://cache.nixos.org/1p6vvyx1n5kgvhwv066qx0x0a358q69f.narinfo’ (Curl error 60)
building path(s) ‘/nix/store/1p6vvyx1n5kgvhwv066qx0x0a358q69f-user-environment’
created 7 symlinks in user environment
installing ‘nss-cacert-3.20’
download-from-binary-cache.pl: could not download ‘https://cache.nixos.org/x3n7q3svl82n8kx4dn9qvx127gbj5xqf.narinfo’ (Curl error 60)
building path(s) ‘/nix/store/x3n7q3svl82n8kx4dn9qvx127gbj5xqf-user-environment’
created 9 symlinks in user environment
[...]

PR #3415, opened 2020-03-14
build: disallow root without build users

LnL7:
Currently nix enforces build user separation for root by setting `build-users-group = nixbld` [by default](https://github.com/NixOS/nix/commit/ada3e3fa15bc14aebb2bafd1240c15cf1fd99351) in this case. However there's no validation allowing the user to disable it. For nix-daemon specifically there used to be a check that ensured builds would not run as root, however that was also removed in https://github.com/NixOS/nix/commit/98968fbb63a1a049b2439bfc2a7d53e5b51471e3.

A bunch of issues mention this bug as a "workaround" for root installations which is pretty 
bad IMHO. And it looks like there are also things in nix that don't account for this behaviour, making it only partially functional.

```
suspicious ownership or permission on '...'; rejecting this build output
```

- https://github.com/NixOS/nix/issues/697
- https://github.com/NixOS/nix/issues/1559
- https://github.com/NixOS/nix/issues/936#issuecomment-475795730
- https://github.com/NixOS/nixpkgs/issues/82357

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -528,7 +528,8 @@ public:
 
 UserLock::UserLock()
 {
-    assert(settings.buildUsersGroup != "");
+    if(settings.buildUsersGroup == "")
+        throw Error("the 'build-users-group' option is required for security reasons");
 
     /* Get the members of the build-users-group. */
     struct group * gr = getgrnam(settings.buildUsersGroup.get().c_str());
@@ -1943,7 +1944,7 @@ void DerivationGoal::startBuilder()
 
     /* If `build-users-group' is not empty, then we have to build as
        one of the members of that group. */
-    if (settings.buildUsersGroup != "" && getuid() == 0) {
+    if (getuid() == 0) {
 #if defined(__linux__) || defined(__APPLE__)
         buildUser = std::make_unique<UserLock>();
 
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -3334,6 +3334,10 @@ void DerivationGoal::runChild()
                 throw SysError("setuid failed");
         }
 
+        /* Prevent users from doing something very dangerous. */
+        if (geteuid() == 0 || getegid() == 0)
+            throw Error("build seems to be running as root, refusing");
+
         /* Fill in the arguments. */
         Strings args;
 


zimbatm:
related: https://github.com/nix-community/nix/pull/1 which makes root also use the daemon protocol when the nix-daemon is started

edolstra:
> However there's no validation allowing the user to disable it. 

What do you mean with validation?

Allowing the user to disable build isolation is a legitimate use case, especially in containers.

LnL7:
@edolstra I disagree with the premise that user separation is not important for containers. I would argue the opposite, being able to setup an environment needed to run a project including global files, users, etc. is exactly what containers are intended for.

Another example of a project that completely refuses to run as root or setuid is postgresql, preventing users from shooting themselves in the foot with a setup that has horrible security implications. Regardless of whether it might be slightly less bad in certain contexts.

edolstra:
@LnL7 The issues you linked to suggest that people really need this (or think they do...). I'm not saying user separation cannot be useful for containers. But generally a container should run a single service (and shouldn't have a writable Nix store).

BTW, we also allow user separation to be disabled for non-root users (e.g. when you do `--store ~/my-nix`). For most people, that's just as bad for security as disabling user separation for root. (I mean, my user account contains all my actual data - I don't really care about root.) So should we disallow that as well?

LnL7:
@edolstra Yeah, I think this mainly stems from confusion about build user separation and the nix-daemon are separate things that don't depend on eachother.

As for a regular single user install, yes it would be good to make that more clear. However unlike with root there are cases where not enough permissions are available. Sandboxing also plays a role in this ofcourse, if you feel strongly about it I could also include that in the condition.

InLaw:
any update?

Ericson2314:
CC @yorickvP because #8017, after/during which is a good chance to document what exactly are the semantics of the build environment. Even if we don't want to make this an error, we could make it a warning.

Ericson2314:
In any case, I think a good preperatory step would be making `build-users-group` `std::optional<std::string>`, not `std::string`. @LnL7 would you be interested in doing that?

fricklerhandwerk:
Triaged in Nix team meeting:

- @Ericson2314 not clear if this is desirable at all
  - special-casing root adds complexity
  - there may be interactions with automatic UID allocation
  - OTOH this is a common gotcha
  - the assert should be dead code to begin with
- @edolstra if you run Nix in a container it may not even matter

PR #3416, opened 2020-03-14
WIP: more rust

Ericson2314:
I think I've become a bit accelerationist about this. If we have some Rust, we might as well have all the Rust. Polyglot is bad.

### DIFF
--- a/nix-rust/src/util/base64.rs
+++ b/nix-rust/src/util/base64.rs
@@ -1,25 +1,26 @@
 use crate::error::Error;
 use lazy_static::lazy_static;
 
-// Returns the length of a base-32 representation of this hash.
+// Returns the length of a base-64 representation of this hash.
 pub fn encoded_len(input_len: usize) -> usize {
     if input_len == 0 {
         0
     } else {
-        (input_len * 8 - 1) / 5 + 1
+        ((4 * input_len / 3) + 3) & !3
     }
 }
 
 pub fn decoded_len(input_len: usize) -> usize {
-    input_len * 5 / 8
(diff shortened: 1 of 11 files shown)


nix-rust/src/util/base64.rs
```diff
@@ -0,0 +1,163 @@
+use crate::error::Error;
```
edolstra: We don't need our own base64 implementation, since there is a crate for that.

nix-rust/src/util/base32.rs
```diff
@@ -64,7 +65,7 @@ pub fn decode(input: &str) -> Result<Vec<u8>, crate::Error> {
     let mut nr_bits_left: usize = 0;
     let mut bits_left: u16 = 0;
 
-    for c in input.chars().rev() {
+    for c in input.chars() {
```
edolstra: Hm, is this correct? IIRC we really do need to process it in reverse order (our base32 encoding is pretty quirky).

nix-rust/src/util/base64.rs
edolstra: Likewise for base16 - we can use the `hex` crate.

edolstra:
I suppose we have to decide whether we really want to go forward with Rust. Currently we only use it for one small type (`StorePath`) which could be easily rewritten in C++.

Ericson2314:
Yeah I thought this would help me do the CAS stuff, but then C++ wasn't so bad so I'll just leave this as is.

Ericson2314:
Closing if/until there is a big Rust push

PR #3418, opened 2020-03-15
Generalize `isFixedOutput` in preparation for CA drvs

Ericson2314:
Today's fixed output derivations and regular derivations differ in a few
ways which are largely orthogonal. This replaces `isFixedOutput` with a
`type` that returns an enum of possible combinations.

I think this is basically a tech debt PR --- our plans could change a lot but it's still nice to decompose `isFixedOutput` to make the code, especially `build.cc` easier to read.

CC @regnat @edolstra


src/libstore/derivations.hh
```diff
@@ -33,6 +34,21 @@ typedef std::map<StorePath, StringSet> DerivationInputs;
...
+enum DerivationTypeAxis : uint8_t {
+    DtAxisCA = 0b10000000,
+    DtAxisFixed = 0b01000000,
+    DtAxisImpure = 0b00100000,
+};
+enum DerivationType : uint8_t {
+    DtRegular = 0b0000000,
+    DtCAFixed = 0b11100000,
+};
+
```
edolstra: Why do we need such an encoding? Why not just
```
struct DerivationType {
    bool isImpure = false;
    bool isFixedOutput = false;
    bool isCA = false;
    ...
};
```

Ericson2314: Well I wanted to emphasize that not all combinations are valid. But we could turn a derivation type into one of those rather than doing `&` tests.

src/libstore/build.cc
```diff
@@ -3658,19 +3658,30 @@ void DerivationGoal::registerOutputs()
 
             /* Check the hash. In hash mode, move the path produced by
                the derivation to its content-addressed location. */
-            Hash h2 = recursive ? hashPath(h.type, actualPath).first : hashFile(h.type, actualPath);
+            Hash h2 = recursive ? hashPath(ht, actualPath).first : hashFile(ht, actualPath);
 
             auto dest = worker.store.makeFixedOutputPath(recursive, h2, i.second.path.name());
 
-            if (h != h2) {
+            // true if ither floating CA, or incorrect fixed hash.
```
asymmetric:
```suggestion
            // true if either floating CA, or incorrect fixed hash.
```

src/libstore/local-store.cc
```diff
@@ -559,21 +559,29 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
...
-        out->second.parseHashInfo(recursive, h);
-
-        check(makeFixedOutputPath(recursive, h, drvName), out->second.path, "out");
-    }
-
-    else {
-        Hash h = hashDerivationModulo(*this, drv, true);
-        for (auto & i : drv.outputs)
-            check(makeOutputPath(i.first, h, drvName), i.second.path, i.first);
+    // Don't need the answer, but do this anways to assert is proper
```
asymmetric:
```suggestion
    // Don't need the answer, but do this anyways to assert is proper
```

src/libstore/local-store.cc
```diff
@@ -559,21 +559,29 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
...
+    // combination. The code below is more general and naturally allows
+    // combinations that currently prohibited.
```
asymmetric:
```suggestion
    // combinations that are currently prohibited.
```

src/libstore/derivations.hh
regnat: I'm not thrilled about this encoding either (well I don't mind having it internally, but it doesn't feel right to have it leaking outside).

Maybe just instead have a family of functions `bool is{CA,Fixed,Impure}(DerivationType derivationType)` so that the actual encoding stays local but we can still ensure that only a few combinations are valid

src/libstore/local-store.cc
```diff
@@ -559,21 +559,29 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
...
-        Hash h = hashDerivationModulo(*this, drv, true);
-        for (auto & i : drv.outputs)
-            check(makeOutputPath(i.first, h, drvName), i.second.path, i.first);
+    // Don't need the answer, but do this anyways to assert is proper
+    // combination. The code below is more general and naturally allows
+    // combinations that are currently prohibited.
+    drv.type();
+
+    std::optional<Hash> h;
+    for (auto & i : drv.outputs) {
```
regnat, line 568: I'm not sure what the point of this change is. Looks like it allows mixing FO and non-FO outputs in the same derivation, but that doesn't seem related to the rest of the changes :thinking: Or did I miss anything?

src/libstore/build.cc
```diff
@@ -3644,10 +3644,10 @@ void DerivationGoal::registerOutputs()
            hash). */
         std::string ca;
 
-        if (fixedOutput) {
+        if (i.second.hashAlgo != "") {
```
regnat, line 3647: Can't this be replaced by `DtAxisCA & derivationType` ?

src/libstore/local-store.cc
Ericson2314: Good catch, this is useful later for CA stuff, but doesn't need to be part of this PR.

Ericson2314: (I split it out of a branch that did more stuff for ca derivations, and I guess I forgot to remove this.)

src/libstore/derivations.hh
Ericson2314: Yes, that is just what I am thinking :)

src/libstore/build.cc
Ericson2314: Ah, this too was a remnant getting ahead to encoding CA derivations as having outputs with a hash algo but not a hash.

Ericson2314: We'll remove for now.

Ericson2314:
@regnat Thanks for the review, should all be fixed now.

Ericson2314:
OK conflicts removed. @edolstra Have fixed all your concerns, and @regnat is happy with it too.

Ericson2314:
@edolstra might we get this merged soon?

src/libstore/derivations.cc
```diff
@@ -8,6 +8,32 @@
...
 namespace nix {
 
+bool derivationIsCA(DerivationType dt) {
+    switch (dt) {
+    case DerivationType::Regular: return false;
+    case DerivationType::CAFixed: return true;
+    };
+    // Since enums can have non-variant values, but making a `default:` would
+    // disable exhaustiveness warnings.
+    abort();
```
edolstra, line 18: Could you change this to `assert(false)`? That way, if we ever hit this assertion, we get an error message with location info.

edolstra:
There's a merge conflict in derivations.cc.

Ericson2314:
@edolstra were fixing it, I recommend you just merge https://github.com/NixOS/nix/pull/3807 and afterwords https://github.com/NixOS/nix/pull/3830. #3830 includes this and #3424, and resolves the 3-way conflicts in some non-trivial ways.

src/libstore/derivations.cc
meditans: That's changed now, thanks

PR #3424, opened 2020-03-19
hashDerivationModulo: Generalize for multiple fixed ouputs per drv

Ericson2314:
See documentation in header and comments in implementation for details.

This is actually done in preparation for floating ca derivations, not
multi-output fixed ca derivations, but the distinction doesn't yet
matter.

CC @regnat @edolstra @kmicklas


src/libstore/derivations.hh
```diff
@@ -87,10 +88,42 @@ Derivation readDerivation(const Store & store, const Path & drvPath);
...
+   (after all, (the hash of) the file being downloaded is unchanged).
+   So the *output paths* should not change. On the other hand, the
+   *derivation paths* should change to reflect the new dependency
+   graph.
+
+   For fixed output derivations, this returns map from the names of
+   each output to hash unique up to the output's contents.
+
+   For regular derivations, it returns a single hash of the derivation
+   ATerm, after subderivations have been likewise expunged from hat
```
cole-h:
```suggestion
   ATerm, after subderivations have been likewise expunged from that
```

src/libstore/local-store.cc
```diff
@@ -571,7 +571,9 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
     }
 
     else {
-        Hash h = hashDerivationModulo(*this, drv, true);
+        // Regular non-CA derivation should always return a single hash and not
```
cole-h:
```suggestion
        // Regular, non-CA derivation should always return a single hash and not
```

src/libexpr/primops.cc
```diff
@@ -742,7 +742,9 @@ static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * *
                 DerivationOutput(StorePath::dummy.clone(), "", ""));
         }
 
-        Hash h = hashDerivationModulo(*state.store, Derivation(drv), true);
+        // Regular non-CA derivation should always return a single hash and not
```
cole-h:
```suggestion
        // Regular, non-CA derivation should always return a single hash and not
```

src/libstore/derivations.hh
```diff
@@ -87,10 +88,42 @@ Derivation readDerivation(const Store & store, const Path & drvPath);
 // FIXME: remove
 bool isDerivation(const string & fileName);
 
-Hash hashDerivationModulo(Store & store, const Derivation & drv, bool maskOutputs);
+/* Returns hashes with the details of fixed-output subderivations
+   expunged.
+
+   A fixed-output derivation is a derivation whose outputs have an a
+   specified content hash hash algorithm. (Currently they must have
```
cole-h:
```suggestion
   A fixed-output derivation is a derivation whose outputs have a
   specific content hash hash algorithm. (Currently they must have
```

(or `the specified`)

Would this still make sense if one  of the `hash`es was dropped -- `specific content hash algorithm`?

Ericson2314:
@cole-h Thanks for catching all my types. I guess this is what happens when I bang out some prose just before going to bed.

src/libstore/derivations.cc
```diff
@@ -338,49 +338,82 @@ bool BasicDerivation::isFixedOutput() const
...
-                readDerivation(store, store.toRealPath(store.printStorePath(i.first))), false)).first;
+        const auto res = pathDerivationModulo(store, i.first);
+        if (const Hash *pval = std::get_if<0>(&res)) {
+            // regular non-CA derivation, replace derivation
+            inputs2.insert_or_assign(pval->to_string(Base16, false), i.second);
+        } else if (const std::map<std::string, Hash> *pval = std::get_if<1>(&res)) {
+            // CA derivation's output hashes
+            std::set justOut = { std::string("out") };
+            for (auto & output : i.second) {
+                /* Put each one in with a single "out" output.. */
```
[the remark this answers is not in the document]
Ericson2314: Just because it was used elsewhere a lot. I would personally always use `//` too.

src/libstore/derivations.cc
```diff
@@ -338,49 +338,82 @@ bool BasicDerivation::isFixedOutput() const
...
+/* See the header for interface details. There are the implemen
+
+   For fixed ouput derivations, each hash the map is not the output's
+   content hashes, but a hash of that hash along with other constant
+   data. The key point is that value is a pure function of the output's
+   contents, and there are no preimage attacks spoofing an output's
+   contents for a derivation or derivation for output's contents.
+
+   For regular derivations, it looks up each subderivation from its hash
+   and recurs. If the subderivation is also regular, it simply
```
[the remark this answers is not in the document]
Ericson2314: :) To me "recurses" reads weird.

src/libstore/derivations.cc
```diff
@@ -338,49 +338,82 @@ bool BasicDerivation::isFixedOutput() const
...
+   substitutes the derivation path with its hash. If the subderivation
+   is fixed-output, however, it takes each output hash and pretends it
+   is a derivation hash producing a single "out" output. This is so we
+   don't leak the provenance of fixed outputs, pointless reducing cache
```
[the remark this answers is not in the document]
Ericson2314: reducing pointless cache misses, actually. I messed up word order and antonym!

src/libstore/derivations.hh
Ericson2314: It was supposed to be "hash and hash algorithm", roughly like the old version of the comment.

src/libstore/derivations.cc
```diff
@@ -338,49 +338,82 @@ bool BasicDerivation::isFixedOutput() const
...
-   function, and that for fixed-output derivations we return a hash of
-   its output path. */
-Hash hashDerivationModulo(Store & store, const Derivation & drv, bool maskOutputs)
+/* See the header for interface details. There are the implemen
+
+   For fixed ouput derivations, each hash the map is not the output's
+   content hashes, but a hash of that hash along with other constant
+   data. The key point is that value is a pure function of the output's
+   contents, and there are no preimage attacks spoofing an output's
+   contents for a derivation or derivation for output's contents.
```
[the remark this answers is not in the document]
Ericson2314: It was missing some articles and commas, fixed now, thanks.

Ericson2314:
OK fixed the typos (thanks again @cole-h!) and also added a `typedef` for all the identical new `std::variant`s.

src/libstore/derivations.cc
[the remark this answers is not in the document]
cole-h: A "me-problem" it is, then!

/me wishes he could resolve his own comments

src/libstore/derivations.cc
```diff
@@ -338,49 +338,81 @@ bool BasicDerivation::isFixedOutput() const
...
+                    store,
+                    store.toRealPath(store.printStorePath(drvPath))),
+                false)).first;
+    }
+    return h->second;
+}
+
+/* See the header for interface details. These are the implementation details.
+
+   For fixed ouput derivations, each hash in the map is not the
```
cole-h: Just to be conformant with how it's referred to elsewhere.

```suggestion
   For fixed-output derivations, each hash in the map is not the
```

src/libstore/derivations.cc
```diff
@@ -338,49 +338,81 @@ bool BasicDerivation::isFixedOutput() const
...
+   corresponding output's content hash, but a hash of that hash along
+   with other constant data. The key point is that the value is a pure
+   function of the output's contents, and there are no preimage attacks
+   spoofing an either an output's contents for a derivation, or
+   derivation for an output's contents.
```
cole-h: IMHO, this makes the sentence more clear and readable.

```suggestion
   either spoofing an output's contents for a derivation, or
   spoofing a derivation for an output's contents.
```

src/libstore/derivations.cc
```diff
@@ -338,49 +338,81 @@ bool BasicDerivation::isFixedOutput() const
...
+
+   For regular derivations, it looks up each subderivation from its hash
+   and recurs. If the subderivation is also regular, it simply
+   substitutes the derivation path with its hash. If the subderivation
+   is fixed-output, however, it takes each output hash and pretends it
+   is a derivation hash producing a single "out" output. This is so we
+   don't leak the provenance of fixed outputs, reducing pointless cache
+   misses as the build itself won't know this.
```
cole-h: This remark seems more like an aside.

```suggestion
   misses (as the build itself won't know this).
```

or

```suggestion
   misses, because the build itself won't know this.
```

src/libstore/derivations.hh
```diff
@@ -87,10 +88,38 @@ Derivation readDerivation(const Store & store, const Path & drvPath);
...
+> DrvHashModulo;
+
+/* Returns hashes with the details of fixed-output subderivations
+   expunged.
+
+   A fixed-output derivation is a derivation whose outputs have a
+   specified content hash and hash algorithm. (Currently they must have
+   exactly one output (`out'), which is specified using the `outputHash'
+   and `outputHashAlgo' attributes, but the algorithm doesn't assume
+   this). We don't want changes to such derivations to propagate upwards
```
cole-h: Since this whole sentence is in parentheses, it makes sense to keep the punctuation inside the parentheses as well.

```suggestion
   this.) We don't want changes to such derivations to propagate upwards
```

src/libstore/derivations.hh
```diff
@@ -87,10 +88,38 @@ Derivation readDerivation(const Store & store, const Path & drvPath);
...
+   through the dependency graph, changing output paths everywhere.
+
+   For instance, if we change the url in a call to the `fetchurl'
+   function, we do not want to rebuild everything depending on it (after
+   all, (the hash of) the file being downloaded is unchanged).  So the
+ *output paths* should not change. On the other hand, the *derivation
+ paths* should change to reflect the new dependency graph.
```
cole-h: Small reflow issue ;)

```suggestion
   *output paths* should not change. On the other hand, the *derivation
   paths* should change to reflect the new dependency graph.
```

src/libstore/derivations.hh
```diff
@@ -87,10 +88,38 @@ Derivation readDerivation(const Store & store, const Path & drvPath);
...
+   A fixed-output derivation is a derivation whose outputs have a
+   specified content hash and hash algorithm. (Currently they must have
+   exactly one output (`out'), which is specified using the `outputHash'
+   and `outputHashAlgo' attributes, but the algorithm doesn't assume
+   this). We don't want changes to such derivations to propagate upwards
+   through the dependency graph, changing output paths everywhere.
+
+   For instance, if we change the url in a call to the `fetchurl'
+   function, we do not want to rebuild everything depending on it (after
+   all, (the hash of) the file being downloaded is unchanged).  So the
```
cole-h:
```suggestion
   function, we do not want to rebuild everything depending on it -- after
   all, (the hash of) the file being downloaded is unchanged.  So the
```

src/libstore/derivations.hh
```diff
@@ -87,10 +88,38 @@ Derivation readDerivation(const Store & store, const Path & drvPath);
...
+ *output paths* should not change. On the other hand, the *derivation
+ paths* should change to reflect the new dependency graph.
+
+   For fixed output derivations, this returns a map from the names of
+   each output to hashes unique up to the outputs' contents.
```
cole-h: As above, so below. (Conform with its references elsewhere.)

```suggestion
   For fixed-output derivations, this returns a map from the names of
   each output to their hashes, unique up to the outputs' contents.
```

Does this wording make sense? It just needs a pause (read: comma) in there somewhere, so I hope this facilitates this satisfactorily. Maybe we could also slice off the plural from `hashes` since a single output cannot have multiple hashes (I believe -- maybe my understanding is incorrect):

```suggestion
   For fixed-output derivations, this returns a map from the names of
   each output to its hash, unique up to the outputs' contents.
```

src/libstore/derivations.hh
```diff
@@ -87,10 +88,38 @@ Derivation readDerivation(const Store & store, const Path & drvPath);
...
+
+   For regular derivations, it returns a single hash of the derivation
+   ATerm, after subderivations have been likewise expunged from that
```
cole-h: Nothing needs to be changed here, but what does `ATerm` mean in this context? If somebody could give a brief description, I would really appreciate that :)

(Or pointing to its documentation or something and telling me to RTFM works too.)

src/libstore/derivations.hh
Ericson2314: Indeed it does. I like to violate the rules and do both `".` and `."` for the same reasons. Nesting matters!

src/libstore/derivations.hh
Ericson2314:
```suggestion
   function, we do not want to rebuild everything depending on it---after
   all, (the hash of) the file being downloaded is unchanged.  So the
```

Ericson2314: LaTeX taught me 3 for me, 2 for en :)

src/libstore/derivations.hh
Ericson2314: So the purity agrees
```suggestion
   For fixed-output derivations, this returns a map from the name of
   each output to its hash, unique up to the output's contents.
```

src/libstore/derivations.hh
Ericson2314: hahaha we are in the bowel's of Nix history now. It come from https://strategoxt.org/Stratego/ATermLibrary, which was something Eelco's adviser was working on IIRC. I think the thesis mentions this.

src/libstore/derivations.hh
cole-h: Honestly, the rule for `".` and `."` is super strange. When quoting, I don't want punctuation that wasn't originally there to be inside the quote... but leaving it outside the quote just looks strange. It's why I try to pad it with extra words before the end of a sentence. English is weird.

src/libstore/derivations.cc
Ericson2314: To me it's a bit more mandatory/essential than an aside, so I think I want to keep it as it is.

src/libstore/derivations.cc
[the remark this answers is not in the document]
Mic92: It seemed to be used in multi-line statements mostly.

src/libstore/derivations.hh
zimbatm: If you look in the git history you can find references to that library. Since Nix is using just a small subset, it has been replaced by its own implementation since.

Issue #1994, opened 2018-03-20
support for custom ssh port


volth:
There is ```NIX_SSHOPTS="-p10022"``` to specify custom ssh and it works.

The only problem is the custom port specified that way is the port is not visible in ssh:// urls:

```
copying path '/nix/store/...-qemu-2.11.1' from 'http://10.0.0.1:44444'...
```

but

```
copying path '/nix/store/...-python2.7-Markdown-2.6.10' to 'ssh://root@x3.wan'
````

It would be nice to add first class support for ```ssh``` ports, as it is done for ```http```

volth:
```git``` supports port in ```ssh://``` scheme, for example "ssh://git@github.com:22/asdf/asdf.git"

shlevy:
@volth I think there are enough `ssh` options that we probably just want to rely on the (daemon) user's ssh config... Is that not adequate?

volth:
One difficult case (besides just ugliness of passing ```$NIX_SSHOPTS=-p10022``` here and there) is many remote builders sharing the same external IP from behind NAT.
When the builders have the same IP but different SSH ports.

shlevy:
@volth You can easily cover this case in your ssh config though... give them different `Host` keys but put the same IP in the `Hostname`

volth:
I see, there are many hacky workarounds.
Tinc network would allow to give them unique IPs, etc

grahamc:
Using an ssh config file isn't a very hacky solution though, is it?

On Apr 7, 2018 8:44 PM, "volth" <notifications@github.com> wrote:

I see, there are many hacky workarounds.
Tinc network would allow to give them unique IPs, etc

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/NixOS/nix/issues/1994#issuecomment-379509744>, or mute
the thread

volth:
> Using an ssh config file isn't a very hacky solution though, is it?

it requires an extra piece of software to be generated from nix or be manually maintained to be in sync with ```nix.buildMachines```.

3noch:
I would also really like to have this. I'm building a tool that runs `nix-build` for you and adds the `--builders` flag, but I can't set the port with this. Supporting a non-standard port is a ton of extra automation work.

shlevy:
@3noch Can you explain why this can't work with ssh config?

shlevy:
[...]

Issue #1572, opened 2017-09-21
nix no longer tells me why I can't build remotely


pikajude:
Derivation in question is a `fetchgit` derivation. In nix 1.11, when a remote builder refused to build a derivation, or nix didn't bother asking the remote builder, there would be a debug message along the lines of `hook reply is "decline"`. No such message is printed anymore, making it a lot harder to figure out what's going wrong:

```
$ nix-store --realise /nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv -vvv
querying info about missing paths...
starting pool of 8 threads
starting substituter program ‘/nix/store/4j9jacx8mjd4jlj53wvymyhxq7dqyj5d-nix-1.11.14/libexec/nix/substituters/download-using-manifests.pl’
starting substituter program ‘/nix/store/4j9jacx8mjd4jlj53wvymyhxq7dqyj5d-nix-1.11.14/libexec/nix/substituters/download-from-binary-cache.pl’
reaping 4 worker threads
these derivations will be built:
  /nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv
building ‘/nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv’
|   building of ‘/nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv’: created
|   building of ‘/nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv’: woken up
|   entered goal loop
|   |   building of ‘/nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv’: init
|   |   building of ‘/nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv’: loading derivation
|   |   acquiring global GC lock ‘/nix/var/nix/gc.lock’
|   |   acquiring read lock on ‘/nix/var/nix/temproots/87548’
|   |   acquiring write lock on ‘/nix/var/nix/temproots/87548’
|   |   downgrading to read lock on ‘/nix/var/nix/temproots/87548’
|   |   building of ‘/nix/store/glilggh7hx78jg2nx3y1j7rpvdzch6mf-devin-fcfc839.drv’: have derivation
|   |   acquiring write lock on ‘/nix/var/nix/temproots/87548’
[...]

PR #3425, opened 2020-03-19
Add user@address:port support

mkg20001:
Adds support for https://github.com/NixOS/nix/issues/1994

Also fixes some issues with https://github.com/NixOS/nix/issues/1572 by adding debug logs that say why a machine was skipped

Note: There's some incomplete stuff and I also have no (actual) clue about C++ so I'd appreciate help

### DIFF
--- a/src/libstore/ssh.cc
+++ b/src/libstore/ssh.cc
@@ -2,14 +2,18 @@
 
 namespace nix {
 
-SSHMaster::SSHMaster(const std::string & host, const std::string & keyFile, bool useMaster, bool compress, int logFD)
-    : host(host)
-    , fakeSSH(host == "localhost")
-    , keyFile(keyFile)
+SSHMaster::SSHMaster(const std::string & hostNport, const std::string & keyFile, bool useMaster, bool compress, int logFD)
+    : keyFile(keyFile)
     , useMaster(useMaster && !fakeSSH)
     , compress(compress)
     , logFD(logFD)
 {
+    auto res = tokenizeString<std::vector<string>>(hostNport, ":");
+    port = res.size() > 1 ? atoi(res[1].c_str()) : 0; // if we have a port, parse that
+    host = res[0];
+
+    fakeSSH = host == "localhost" && !port;
+
     if (host == "" || hasPrefix(host, "-"))
         throw Error("invalid SSH host name '%s'", host);
 }
@@ -22,6 +26,8 @@ void SSHMaster::addCommonSSHOpts(Strings & args)
         args.insert(args.end(), {"-i", keyFile});
     if (compress)
         args.push_back("-C");
+    if (port != 0)
+        args.insert(args.end(), {"-p", std::to_string(port)});
 }
 
 std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command)
--- a/src/libstore/ssh.hh
+++ b/src/libstore/ssh.hh
@@ -9,12 +9,13 @@ class SSHMaster
 {
 private:
 
-    const std::string host;
+    std::string host;
     bool fakeSSH;
     const std::string keyFile;
     const bool useMaster;
     const bool compress;
     const int logFD;
+    int port;
 
     struct State
     {


src/libstore/ssh.cc
```diff
@@ -2,14 +2,20 @@
...
-    : host(host)
-    , fakeSSH(host == "localhost")
+SSHMaster::SSHMaster(const std::string & hostNport, const std::string & keyFile, bool useMaster, bool compress, int logFD)
+    : fakeSSH(host == "localhost")
     , keyFile(keyFile)
     , useMaster(useMaster && !fakeSSH)
     , compress(compress)
     , logFD(logFD)
 {
+    auto res = split(hostNport, *":");
```
volth: `*":"` is `':'`, isn't it?

src/libutil/util.cc
```diff
@@ -1308,6 +1308,32 @@ bool hasSuffix(std::string_view s, std::string_view suffix)
         && s.substr(s.size() - suffix.size()) == suffix;
 }
 
+std::vector<std::string> split(const std::string& value,
```
edolstra: We already have a function `tokenizeString` to split strings.

src/libutil/util.cc
```diff
@@ -1308,6 +1308,32 @@ bool hasSuffix(std::string_view s, std::string_view suffix)
...
+    while ((q = value.find(separator, p)) != std::string::npos)
+    {
+        result.emplace_back(value, p, q - p);
+        p = q + 1;
+    }
+    result.emplace_back(value, p);
+    return result;
+}
+
+std::string join(const std::vector<std::string> strings,
```
edolstra: This looks a lot like the function `concatStringsSep`.

edolstra:
Not really opposed to this, but it's a lot of code to support something that can already be achieved via `ssh_config`.

mkg20001:
I'll move the additional debug logs to another PR

Right now just IPv6 support is missing

Meanwhile I learned how to C++, so that's why that took so long

tomberek:
Still interested

vincentbernat:
There is already code for compression and SSH key, both of them can also be set in `ssh_config`. One can also use `NIX_SSHOPTS`. I am also interested into this to make `darwin.builder` work on another port than 22. I think the idea to support it in the URI of the builder is to avoid messing with too many user files.

Ericson2314:
Note if we do this, we should also consider doing it with

> Use the same notion of a Machine config for remote builders.

from https://github.com/NixOS/hydra/issues/1164

This syntax would be used in the build machines files too, and we don't want that to drift further apart between Nix and Hydra.

mkg20001:
Currently I've got no intention of updating this pr, so feel free to force push / make a new one

Ericson2314:
OK, very understandable after 3 years! I converted it to a draft.

Ericson2314:
Discussed last Nix team meeting.

Verdict: Idea sounds good. Port parsing however should be pat of the URL parser in general, not extracted in this special case. Store that don't support ports can just fail on that; stores that do like this one can set it very easily.

Ericson2314:
OK this PR is on the right track now, but it is not yet ready:

1. Restore passing the user part of the URL to stores, get rid of `authHack` tricks.
2. Tests if possible
3. Release notes

Mic92 (approved):

src/libutil/include/nix/util/url.hh
```diff
@@ -5,10 +5,75 @@
...
+     * Parsed representation of a URL authority.
+     *
+     * It consists of user information, hostname and an optional port number.
+     * Note that passwords in the userinfo are not yet supported and are ignored.
+     *
+     * @todo Maybe support passwords in userinfo part of the url for auth.
+     */
+    struct Authority
+    {
+    public:
```
Ericson2314:
```suggestion
```

src/libutil/include/nix/util/url.hh
```diff
@@ -5,10 +5,75 @@
...
+            IPv6,
+            IPvFuture
+        };
+
+        static Authority parse(std::string_view encodedAuthority);
+        bool operator==(const Authority & other) const = default;
+        std::string to_string() const;
+        friend std::ostream & operator<<(std::ostream & os, const Authority & self);
+
+        /** Type of the host subcomponent, as specified by rfc3986 3.2.2. Host. */
```
Ericson2314:
```suggestion
        /**
         * Type of the host subcomponent, as specified by rfc3986 3.2.2. Host.
         */
```

cole-h:
FYI: this appears to have broken `NIX_SSHOPTS`, c.f. https://github.com/NixOS/nix/issues/14148

EDIT: only when using a host like `root@localhost`, it seems

src/libstore/ssh.cc
```diff
@@ -18,24 +18,62 @@ static std::string parsePublicHostKey(std::string_view host, std::string_view ss
...
-    , fakeSSH(host == "localhost")
+    : authority(authority)
+    , hostnameAndUser([authority]() {
+        std::ostringstream oss;
+        if (authority.user)
+            oss << *authority.user << "@";
+        oss << authority.host;
+        return std::move(oss).str();
+    }())
+    , fakeSSH(authority.host == "localhost")
```
xokdvium, line 69: @cole-h, this will make it behave as it used to:

```suggestion
    , fakeSSH(authority.to_string() == "localhost")
```

Issue #3223, opened 2019-11-12
darwin: DEFAULT_ALLOWED_IMPURE_PREFIXES needs to include /bin/bash on macOS 10.15


lilyball:
On macOS 10.15 and later, executing `/bin/sh` read `/private/var/select/sh`, and if that fails it executes `/bin/bash`. Our current `DEFAULT_ALLOWED_IMPURE_PREFIXES` allows `/bin/sh` but does not allow `/bin/bash`.

We also need to update the sandbox profile itself to include `/bin/bash` for standard packages. Currently nixpkgs [adds `/bin/sh` to `impureHostDeps`](https://github.com/NixOS/nixpkgs/blob/fbfc258f1e60656821999771aa5fc8d175c784cb/pkgs/stdenv/darwin/default.nix#L25). Either we need to extend this to include `/bin/bash`, or we need to tweak Nix to detect when `/bin/sh` [is included](https://github.com/NixOS/nix/blob/aabf5c86c9df1b4e1616a4cf06ee14a6edf2e5e1/src/libstore/build.cc#L3037-L3045) and add `/bin/bash` as well. The former option will rebuild the entire nixpkgs world on darwin because every single stdenv derivation will be modified, whereas the latter will just hack the profile for a macOS 10.15-specific behavior (and we can specialize it to only do this when the current system is macOS 10.15 or later).

lilyball:
Technically we don't need to extend `DEFAULT_ALLOWED_IMPURE_PREFIXES` if we put the hack into Nix to insert `/bin/bash` into the profile instead, but we probably should do it anyway because it's weird to reject a derivation for specifying that they want `/bin/bash` when they have access to it anyway.

Also, we should probably only allow `/bin/bash` as an impure host prefix on macOS 10.15+.

lilyball:
We should also take steps to explicitly deny access to `/private/var/select/sh` in case the user adds it to their `extra-sandbox-paths`; keeping this denied will ensure `/bin/sh` is always `/bin/bash` and can't be pointed at `dash` or `zsh`. I don't know how the sandbox behaves if the profile tries to both deny and allow access to the same literal path, but we'll want to ensure deny takes precedence (if allow always takes precedence, we'll instead want to filter out `/private/var/select/sh` from any attempt to add it to the profile).

lilyball:
For the moment I'm leaning towards patching Nix to insert `/bin/bash` when it sees `/bin/sh` on Catalina, without making any changes to nixpkgs.

[...]

PR #3429, opened 2020-03-20
darwin sandbox

LnL7:
This should make the darwin sandbox functional by default, resolving the issues with dependency tracking of impurities like frameworks in favour of opening up the default sandbox a bit more.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -20,13 +20,6 @@ namespace nix {
    must be deleted and recreated on startup.) */
 #define DEFAULT_SOCKET_PATH "/daemon-socket/socket"
 
-/* chroot-like behavior from Apple's sandbox */
-#if __APPLE__
-    #define DEFAULT_ALLOWED_IMPURE_PREFIXES "/System/Library /usr/lib /dev /bin/sh"
-#else
-    #define DEFAULT_ALLOWED_IMPURE_PREFIXES ""
-#endif
-
 Settings settings;
 
 static GlobalConfig::Register r1(&settings);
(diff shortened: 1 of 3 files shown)


src/libstore/globals.cc
```diff
@@ -68,7 +61,12 @@ Settings::Settings()
     sandboxPaths = tokenizeString<StringSet>("/bin/sh=" SANDBOX_SHELL);
 #endif
 
-    allowedImpureHostPrefixes = tokenizeString<StringSet>(DEFAULT_ALLOWED_IMPURE_PREFIXES);
+
+/* chroot-like behavior from Apple's sandbox */
+#if __APPLE__
+    sandboxPaths = tokenizeString<StringSet>("/System/Library/Frameworks /System/Library/PrivateFrameworks /bin/sh /bin/bash /private/tmp /private/var/tmp /usr/lib");
```
LnL7: I would _really_ prefer not to need bash here (https://github.com/NixOS/nix/issues/3223) but other than either disallowing `/bin/sh` completely or having nix depend on something like fuse bindfs I don't think we can work around it. 😕

src/libstore/sandbox-defaults.sb
```diff
@@ -85,3 +91,7 @@
        (literal "/etc")
        (literal "/var")
        (literal "/private/var/tmp"))
+
+; This is used by /bin/sh on macOS 10.15 and later.
+(allow file*
+       (literal "/private/var/select/sh"))
```
lilyball: We actually want to explicitly *disallow* access to this, because this file is what’s used to control whether `/bin/sh` is `bash` or `zsh`. If we disallow access it will always resolve to bash. If we allow access then the user can configure it to be `zsh` instead.

LnL7: Thought I tested that and it failed, but looks like it's indeed only an error message.

```
$ sandbox-exec -f sandbox.sb /bin/sh -c "echo ok"
Error opening /private/var/select/sh: Operation not permitted
ok
```

Luckily it doesn't really matter and not having to spew out errors out by default is nice.
```
Failed to exec /bin/zsh as variant for /bin/sh (1: Operation not permitted). Falling back to /bin/bash.
ok
```

src/libstore/globals.cc
lilyball: I really wish there was some way to say "allow /bin/bash but only when executed by /bin/sh" 🤷‍♀️

Issue #3430, opened 2020-03-21
install script should check for xz before downloading


juhp:
I tried to run `curl -L --proto '=https' --tlsv1.2 https://nixos.org/nix/install` in a fedora container:

```
[root@60d24278c208 /]# curl -L --proto '=https' --tlsv1.2 https://nixos.org/nix/install | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2399  100  2399    0     0   3978      0 --:--:-- --:--:-- --:--:--  3985
downloading Nix 2.3.3 binary tarball for x86_64-linux from 'https://nixos.org/releases/nix/nix-2.3.3/nix-2.3.3-x86_64-linux.tar.xz' to '/tmp/nix-binary-tarball-unpack.E1mquB8Q3K'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    86  100    86    0     0   2866      0 --:--:-- --:--:-- --:--:--  2866
100 16.4M  100 16.4M    0     0  2451k      0  0:00:06  0:00:06 --:--:-- 3626k
tar (child): xz: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
sh: failed to unpack 'https://nixos.org/releases/nix/nix-2.3.3/nix-2.3.3-x86_64-linux.tar.xz'
```

So it would be better if the install script complains about missing xz before downloading.

pmiddend:
The install script seems to test for `tar` and then uses `tar -J`, but doesn't test for `xz`. I'll make a PR about that.

PR #3431, opened 2020-03-21
installer: also test for xz to unpack

pmiddend:
As mentioned in #3430, we are checking for `tar`, and then use `-J` to unpack the `xz` file. But this needs `xz`, too!

### DIFF
--- a/scripts/install.in
+++ b/scripts/install.in
@@ -36,6 +36,7 @@ tarball="$tmpDir/$(basename "$tmpDir/nix-@nixVersion@-$system.tar.xz")"
 
 require_util curl "download the binary tarball"
 require_util tar "unpack the binary tarball"
+require_util xz "unpack the binary tarball"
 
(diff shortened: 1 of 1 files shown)


LnL7:
On some platforms, like darwin, tar supports lzma commpression without having xz available as a command. This change makes the command mandatory which means people would have to install another package manager to make that available first.

```
$ xz --help
The program 'xz' is currently not installed. You can install it
by typing:
  nix-env -iA nixpkgs.lzma.bin

$ tar -v -xJf nix-2.3.5-x86_64-darwin.tar.xz -C /tmp
x nix-2.3.5-x86_64-darwin/install
x nix-2.3.5-x86_64-darwin/create-darwin-volume.sh
x nix-2.3.5-x86_64-darwin/install-darwin-multi-user.sh
x nix-2.3.5-x86_64-darwin/install-systemd-multi-user.sh
x nix-2.3.5-x86_64-darwin/install-multi-user
...
```

PR #3432, opened 2020-03-21
Fix #3369: verify output hash in nix-build --check

### LINKED ISSUE #3369: nix-build --check doesn't validate hash for fixed-output derivations
Hi,

Looks like nix-build --check (same for nix-store -r --check) doesn't validate hashes correctly for fixed output derivations if outputhash didn't change but the output did. Here is my minimal example:
./test1.nix
```
let
  pkgs = import <nixpkgs> {};
in
pkgs.stdenv.mkDerivation rec {
  name = "testtest";

  outputHashAlgo = "sha256";
  outputHash = "0bllx0p5ppdbhcq3ifddrjmbvdlw53f3brjxdcv2f65n1zira4y2";
  outputHashMode = "recursive";

  phases = "buildPhase";

  buildPhase = ''
    mkdir -p $out

    VERSION=1

    echo "Building $VERSION"

    echo "$VERSION" > text.txt

    cp text.txt $out/
  '';
}
```

The outputHash is correct and if I run `nix-build --check ./test1.nix` it works as expected:
```
nix-build --check  ./test1.nix
checking outputs of '/nix/store/i3hmsf14h0z8fl6b1mdsp8b10cc1vmxf-testtest.drv'...
building
Building 1
warning: rewriting hashes in '/nix/store/jazacw0l5zc0gsfbvqan3q5dhr10xb1n-testtest'; cross fingers
/nix/store/jazacw0l5zc0gsfbvqan3q5dhr10xb1n-testtest
```

However if I change the buildPhase script, like set VERSION=2, `nix-build --check` does build it as expected but succeeds even though the output has changed:
```
nix-build --check  ./test1.nix
checking outputs of '/nix/store/170hzq7hm31gzwswkqpxy070j8ww31dq-testtest.drv'...
building
Building 2
warning: rewriting hashes in '/nix/store/jazacw0l5zc0gsfbvqan3q5dhr10xb1n-testtest'; cross fingers
/nix/store/jazacw0l5zc0gsfbvqan3q5dhr10xb1n-testtest
```

So I think that's by itself is an issue since I believe nix should have failed with hash mismatch since it clearly rebuilt the derivation (see it printed "Bulding 2").

But here even more weird behavior which hopefully can help to investigate the issue. If I spoil the outputHash:
```
let
  pkgs = import <nixpkgs> {};
in
pkgs.stdenv.mkDerivation rec {
  name = "testtest";

  outputHashAlgo = "sha256";
  outputHash = "0bllx0p5ppdbhcq3ifddrjmbvdlw53f3brjxdcv2f65n1zira000";  //  I replaced last 3 characters with 000
  outputHashMode = "recursive";

  phases = "buildPhase";

  buildPhase = ''
    mkdir -p $out

    VERSION=2

    echo "Building $VERSION"

    echo "$VERSION" > text.txt

    cp text.txt $out/
  '';
}
```

`nix-build` fails as expected (can't run it with --check right away):
```
nix-build  ./test1.nix        
  /nix/store/27kwjxw2z7v8k92af5lil428wgsz7y1s-testtest.drv
building '/nix/store/27kwjxw2z7v8k92af5lil428wgsz7y1s-testtest.drv'...
building
Building 2
hash mismatch in fixed-output derivation '/nix/store/76cf68ph3077xndriql3py1yggb3mh0x-testtest':
  wanted: sha256:0bllx0p5ppdbhcq3ifddrjmbvdlw53f3brjxdcv2f65n1zira000
  got:    sha256:1j7i9iipr2b07dp4dd3pq1paf137sywcsz8zqkklf04k92hgrzmi
error: build of '/nix/store/27kwjxw2z7v8k92af5lil428wgsz7y1s-testtest.drv' failed
```
However now, if I change the outputHash back to the original 0bllx0p5ppdbhcq3ifddrjmbvdlw53f3brjxdcv2f65n1zira4y2 (which is still wrong, since I changed VERSION=1 to VERSION=2), `nix-build --check` starts to fail as expected:
```
nix-build --check  ./test1.nix
checking outputs of '/nix/store/170hzq7hm31gzwswkqpxy070j8ww31dq-testtest.drv'...
building
Building 2
warning: rewriting hashes in '/nix/store/jazacw0l5zc0gsfbvqan3q5dhr10xb1n-testtest'; cross fingers
hash mismatch in fixed-output derivation '/nix/store/76cf68ph3077xndriql3py1yggb3mh0x-testtest':
  wanted: sha256:0bllx0p5ppdbhcq3ifddrjmbvdlw53f3brjxdcv2f65n1zira4y2
  got:    sha256:1j7i9iipr2b07dp4dd3pq1paf137sywcsz8zqkklf04k92hgrzmi
error: build of '/nix/store/170hzq7hm31gzwswkqpxy070j8ww31dq-testtest.drv' failed
```

I ran it on macos with nix 2.3.1 in multi-user mode, but also reproduced on linux (multiuser) and mac single user.

dshkyra:
`nix-build --check` (or `nix-store --realise --check`) behaves differently
depending on whether expected output of fixed-output derivation was
built (and registered) before or not.

Throw an exception in case of hash mismatch when `--check` flag is
specified even if the expected output wasn't registered before.

Help to check and detect cases when output of fixed-output derivation is
changed however outputHash doesn't change and so matches the output
built before.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -3673,7 +3673,7 @@ void DerivationGoal::registerOutputs()
 
                 Path actualDest = worker.store.toRealPath(worker.store.printStorePath(dest));
 
-                if (worker.store.isValidPath(dest))
+                if (worker.store.isValidPath(dest) || buildMode == bmCheck)
                     std::rethrow_exception(delayedException);
 
                 if (actualPath != actualDest) {


src/libstore/build.cc
```diff
@@ -3673,7 +3673,7 @@ void DerivationGoal::registerOutputs()
 
                 Path actualDest = worker.store.toRealPath(worker.store.printStorePath(dest));
 
-                if (worker.store.isValidPath(dest))
+                if (worker.store.isValidPath(dest) || buildMode == bmCheck)
```
DzmitrySudnik: Why that condition is here at all? Feels like if the hashes don't match, the builder should error. What am I missing?

dshkyra: I believe the point here is just delaying the exception -- the exception is re-thrown at [the end of the function](https://github.com/NixOS/nix/pull/3432/files#diff-267a6391980cbd9a743958945ff44ef9R3842). Unfortunately not in the case when `buildMode == bmCheck`.

thufschmitt:
Thanks for reopening that @Artturin .

@dshkyra (if you're still around as it's a fairly old one), do you happen to have an example of such a non-failure? I failed to reproduce it on a recent Nix, and I can't tell whether it's because my repro isn't correct or whether it got accidentally fixed in the meantime

dshkyra:
> Thanks for reopening that @Artturin .
> 
> @dshkyra (if you're still around as it's a fairly old one), do you happen to have an example of such a non-failure? I failed to reproduce it on a recent Nix, and I can't tell whether it's because my repro isn't correct or whether it got accidentally fixed in the meantime

@thufschmitt thanks for looking into my PR. I don't have chance to repro it right now, but the steps to repro are
1. Create a test derivation that references another fixed output derivation
2. Run `nix-build --check`
3. Change the fixed output derivation, so it generates a different `outputHash`. Keep the previous `outputHash` in the test derivation
4. Run `nix-build --check`. Expected: it fails showing hash mismatch. Actual: success.

I looked on the code. Seems like the code was significantly reworked since I opened this PR. So the issue might be fixed by now.

tomberek:
Is this an example?
Where you switch "b" to "c" in a's script:
```
let
  a = derivation {
    name = "original";
    system = "x86_64-linux";
    builder = "/bin/sh";
    args = ["-c" ''
      echo ${c}
      echo "hi" >  $out
    ''];
  };
  b = derivation {
    name = "first";
    system = "x86_64-linux";
    builder = "/bin/sh";
    outputHash = "sha256-mOpuTyFvL7S2n/+bOkSELDhobKaF8/VdxIxdP7EQe+Q=";
    args = ["-c" ''
      echo "hi" >  $out
    ''];
  };
  c = derivation {
    name = "first-modified";
    system = "x86_64-linux";
    builder = "/bin/sh";
    outputHash = "sha256-mOpuTyFvL7S2n/+bOkSELDhobKaF8/VdxIxdP7EQe+Q=";
    args = ["-c" ''
      echo "hi" >  $out
    ''];
  };
in
a
```

```
[tom@tframe:~/nix/main/t2]$ nix-build ./default.nix --check
  /nix/store/prz5ay9p3q9l2jnlwqfhxndw78218jll-first-modified.drv
  /nix/store/h4bpkqpkahk28fmk264rk2y68yjhmbkf-original.drv
error: some outputs of '/nix/store/h4bpkqpkahk28fmk264rk2y68yjhmbkf-original.drv' are not valid, so checking is not possible
```

dshkyra:
That's not an issue anymore with the latest nix (2.18). I'm closing this.

The problem was that if you change `echo "hi" >  $out` to `echo "hi!" >  $out` in the script below, `nix-build --check` didn't fail if the derivation was already registered. It fails with nix 2.18, i.e. works as expected.
```
derivation {
    name = "hello";
    system = "x86_64-linux";
    builder = "/bin/sh";
    outputHash = "sha256-mOpuTyFvL7S2n/+bOkSELDhobKaF8/VdxIxdP7EQe+Q=";
    args = ["-c" ''
      echo "hi" >  $out
    ''];
```
```
$ nix-build --check
checking outputs of '/nix/store/03cgs6q576bf26l9kq38rrr8zhl919zr-hello.drv'...
error: hash mismatch in fixed-output derivation '/nix/store/03cgs6q576bf26l9kq38rrr8zhl919zr-hello.drv':
         specified: sha256-mOpuTyFvL7S2n/+bOkSELDhobKaF8/VdxIxdP7EQe+Q=
            got:    sha256-GttBz476DDdb9k0IvA/gJ6cg/vDXrAUUDCof4SABVaI=
error: building '/nix/store/03cgs6q576bf26l9kq38rrr8zhl919zr-hello.drv' failed
```

PR #3433, opened 2020-03-21
make hash format configurable

LnL7:


edolstra:
Hm, I'm not sure we should do this. If we want to nudge everybody towards a consistent hash format (e.g. SRI) then it seems better to *not* make this configurable.

LnL7:
Is there a timeline for this, making nix >=2.2 required? I recently almost committed a SRI hash in nixpkgs because the default changed.

The base16 and base32 formats are also already used interchangeably in nixpkgs depending on whether a hash from upstream is used or it's calculated by nix.

jtojnar:
Can we have this as hidden setting until nixpkgs support SRI hashes?

Ideally, we would temporarily add `.envrc` containing `_NIX_HASH_FORMAT=base32` to nixpkgs until minver is bumped, so that commands use compatible output in nixpkgs and modern style everywhere else.

I do not really want to have to choose between playing with flakes and not having to manually shape the mounds of TOFU I encounter when contributing to Nixpkgs.

zimbatm:
I think a better option would be to record the hash encoding alongside the hash in src/libutil/hash.hh . Then when the "expected ... got ..." message is printed, it would allow displaying the original hash encoding. This would allow users and tools to search for that actual hash message in the source. And the "got" encoding can be overriden using the encoding from the original hash.

zimbatm:
I resurrected one of my old experimental branch in #3655. I think that printing the original encoding of a hash would be a better way forward as it allows mixed hashes environments.

PR #3434, opened 2020-03-21
Revise division of labor in deserialization of derivations

Ericson2314:
Now:

 - No `Store` methods are defined in `derivations.cc`

 - `parseDerivation` on the underlying string is exposed, since that is the root functionality. (It might be better to make this take an arbitrary `std::istringstream` at some point.)

 - `Store::derivationFromPath` and `Store::readDerivation` are defined less redundantly.

~~I think it makes more sense to define the data model (derivations), before the operations (store api).~~ This has already been done by other already-merged PRs.


Ericson2314:
@edolstra This little PR look OK to you? I have a lot of changes I'd like to land so don't want to accumulate too much of a backlog.

Ericson2314:
Renamed because this is now a bit more than merely flipping the dependency.

PR #3439, opened 2020-03-23
Store parsed hashes in `DerivationOutput`

Ericson2314:
It's best to detect invalid data as soon as possible, with data types
that make storing it impossible.


Ericson2314:
I had to add a bit more machinery to the C++ side Rust FFI to make this tick. Hopefully that makes it...no less safe than before :D.

puckipedia:
I think the Rust code will cause double frees, as writing into a pointer will also drop the value contained in it. You probably want to replace it with a `ptr::write` or something similar?

Ericson2314:
@puckipedia good catch!

Ericson2314:
Oh, I guess whatever compiler was being used on Darwin couldn't `constexpr`as hard. Removing that which I only put in based on compilers errors, not actual need, anyways.

Ericson2314:
OK! No more spooky FFI changes needed.

src/libstore/derivations.hh
```diff
@@ -12,11 +12,24 @@ namespace nix {
 
 /* Abstract syntax of derivations. */
 
+/// Pair of a hash, and how the file system was ingested
+struct FileSystemHash {
```
edolstra, line 16: Let's call this `DerivationOutputHash` or something like that, since `FileSystemHash` is a bit vague.

src/libstore/derivations.hh
```diff
@@ -12,11 +12,24 @@ namespace nix {
...
+    FileIngestionMethod method;
+    Hash hash;
+    FileSystemHash(FileIngestionMethod method, Hash hash)
+        : method(std::move(method))
+        , hash(std::move(hash))
+    { }
```
edolstra, line 22: We can remove this constructor (and the others) and use designated initializers (`DerivationOutputHash { .method = ..., .hash = ... }`).

src/libstore/store-api.hh
```diff
@@ -838,6 +839,9 @@ std::optional<ValidPathInfo> decodeValidPathInfo(
     std::istream & str,
     bool hashGiven = false);
 
+/* Compute the prefix to the hash algorithm which indicates how the files were
+   ingested. */
+std::string makeFileIngestionPrefix(const FileIngestionMethod m);
```
edolstra, line 844: Wrong indentation?

src/libutil/hash.hh
```diff
@@ -44,6 +44,12 @@ struct Hash
        string. */
     Hash(std::string_view s, HashType type = htUnknown);
 
+    Hash(const Hash &) = default;
+
+    Hash(Hash &&) = default;
+
+    Hash & operator = (const Hash &) = default;
```
edolstra, line 51: Do we need these? They're already defined by default right?

Ericson2314: I will try without, we might have only needed them for `StorePath` when I was debugging STL coppying issues.

src/libstore/store-api.hh
Ericson2314: Oh, actually this is just a freestanding function and the closing brace is the namespace (so no indent). I guess in the later PR it gets a better home in `content-address.h`.

src/libutil/hash.hh
Ericson2314: Not needed!

Ericson2314:
OK! I think I got all of the comments, and CI has turned green.

PR #3440, opened 2020-03-23
installer: Set files read-only when copying into store

gnprice:
After installing Nix, I found that all the files and directories
initially copied into the store were writable, with mode 644 or 755:

  drwxr-xr-x 9 root root 4096 Dec 31  1969 /nix/store/ddmmzn4ggz1f66lwxjy64n89864yj9w9-nix-2.3.3

The reason is that that's how they were in the unpacked tarball, and
the install-multi-user script used `rsync -p` without doing anything
else to affect the permissions.

The plain `install` script for a single-user install takes care to
do a `chmod -R a-w` on each store path copied.  We could do the same
here with one more command; or we can pass `--chmod` to rsync, to
have it write the files with the desired modes in the first place.

Tested the new `rsync` command on both a Linux machine with a
reasonably-modern rsync (3.1.3) and a Mac with its default, ancient,
rsync 2.6.9, and it works as expected on both.  Thankfully the latter
is just new enough to have `--chmod`, which dates to rsync 2.6.7.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -567,7 +567,7 @@ install_from_extracted_nix() {
         cd "$EXTRACTED_NIX_PATH"
 
         _sudo "to copy the basic Nix files to the new store at $NIX_ROOT/store" \
-              rsync -rlpt ./store/* "$NIX_ROOT/store/"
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks!

PR #3443, opened 2020-03-24
Update hacking.xml

bqv:
Configure step is no longer accurate.


edolstra:
Hm, how so? I use `configurePhase` all the time.

bqv:
Didn't work for me on master! I suppose I was doing something wrong. Feel free to close

PR #3446, opened 2020-03-25
Generalize `Derivation` type to better support unresolved derivations

Ericson2314:
This is basically an extended meditation on the `Derivation` types and `hashDerivationModulo`. I wanted to break things into smaller steps (resolving drvs vs hashing them vs making output paths), in preparation for things like the intensional store and better remote building.

Each commit can be reviewed separately; they all should build, and, except for the second to last, each should leave the code in a working state.

~~Not sure whether this or #3439 is more likely to get merged first, but the combination should work well together.~~ Indeed it does, and that is now this PR.

Depends on #3434
Depends on #3883


tomberek:
Still needed after the large amount of existing re-work of derivation representation?

Ericson2314:
I would still like to do this. It is a different sort of change. The history might be an unsalvagable mess, but I soon as well reuse this PR even if I just remake it.

tomberek:

Ericson2314:
I'm happy to report that, 6 years later, the derivation input type is templated and hash modulo is reworked!

PR #3449, opened 2020-03-25
Smaller derivation refactor

Ericson2314:
This is the first few commits of #3446. Trying to bisect the error with something I hope we can agree is a worth-to-merge step in its own right.


Ericson2314:
The main PR now works, and this doesn't have much value on it's own, so I'm just going to close it.

PR #3450, opened 2020-03-25
Extend Rust FFI

Ericson2314:
Do idiomatic C++ copy and move constructors for a few things, so
wrapping structs' defaults can work.


Ericson2314:
Also serves as bisect for #3439

src/libutil/rust-ffi.hh
```diff
@@ -30,14 +30,29 @@ protected:
...
+    }
 
     Value(Value && other)
         : raw(other.raw)
     {
         other.evacuate();
     }
 
+    // Not all Rust types are Clone / Copy, but our base "value" class needs to
+    // have a copy constructor so the ones that are can be.
```
cole-h: "so the ones that are can be" sounds super duper omega wacky. Maybe something like:

```
    // Our base Value class needs to have a copy constructor so that Rust
    // types that implement Copy/Clone can be copied/cloned.
```

Ericson2314:
```suggestion
    // Not all Rust types are Clone / Copy, but our base Value class needs to
    // have a copy constructor so that ones which do implement Copy/Clone
    // can be copied/cloned.
```

cole-h: "ones" is a bit too vague for my tastes.

"copy constructor, so types that do implement Copy/Clone can be copied/cloned." maybe?

---

Slightly more nit-picky: I'd smush the `Clone / Copy` to `Copy/Clone` (so it matches the second instance on the next line).

src/libutil/rust-ffi.hh
```diff
@@ -30,14 +30,30 @@ protected:
...
+    }
 
     Value(Value && other)
         : raw(other.raw)
     {
         other.evacuate();
     }
 
+    // Not all Rust types are Clone / Copy, but our base Value class needs to
+    // have a copy constructor so that ones which do implement Copy/Clone
```
cole-h:
```suggestion
    // have a copy constructor so that types which do implement Copy/Clone
```

:)

src/libstore/path.hh
```diff
@@ -43,6 +44,19 @@ struct StorePath : rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>
...
+    StorePath(StorePath && that) = default;
+
+    StorePath(const StorePath & that)
+    {
+        ffi_StorePath_clone_to(that, *this);
+    }
+
+    void operator = (const StorePath & that)
+    {
+        (rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>::operator = (that));
```
andir: Where does that magic size come from? I see that we start to duplicate that throughout (at least) this file a few times. Shouldn't there be a comment where the magic 24 comes from? For the `3 * sizeof(void *)` part one can probably guess that the struct hold 3 pointers. The 24 bytes could be 3 integers? Can't we use a single `sizeof(SomeStruct)` instead? Does this take into account alignment changes if a new member is added to whatever the underlying struct is?

I know you didn't introduce it but this might go out of sync with the other definition thus I am commenting.

Ericson2314: Nevermind the numbers, I am repeating the whole superclass. Maybe I should `#ifdef` and then `#undef` the superclass; that would help without leaking anything more to the files the include the header.

Ericson2314:
@edolstra It has come up again in https://github.com/NixOS/nix/pull/3673 and https://github.com/NixOS/nix/pull/3523 that the lack of copy constructors this PR fixes prevents us from using idiomatic C++ STL functions, which is a bummer. Is there anything I can do to make this acceptable for you? For example, after #3671 is merged, I plan on writing additional unit tests to ensure `clone` is appropriately called.

Ericson2314:
To someday return to if a Rust rewrite happens.

maisiliym:
Relevant: https://github.com/dtolnay/cxx

PR #3455, opened 2020-03-29
Replace some `bool recursive` with a new `FileIngestionMethod` enum

Ericson2314:
`recursive` is a rather overloaded identifier in nix. There is the recursive demon enum, various one-off meanings like with traversing requisites or printing derivations, and the way we hash file paths. The daemon one already has an enum, and the one-off `bool recursive`s are fine, but I think the "file ingestion method" ones are common enough that we ought to use a dedicated type for it. Furthermore---unlike the others which really are something juxtaposed with its negation---we just so happen to have two file ingestion methods today, and could easily have more.


src/libstore/store-api.cc
```diff
@@ -200,9 +200,9 @@ StorePath Store::makeTextPath(std::string_view name, const Hash & hash,
 
 
 std::pair<StorePath, Hash> Store::computeStorePathForPath(std::string_view name,
-    const Path & srcPath, bool recursive, HashType hashAlgo, PathFilter & filter) const
+    const Path & srcPath, FileIngestionMethod recursive, HashType hashAlgo, PathFilter & filter) const
 {
-    Hash h = recursive ? hashPath(hashAlgo, srcPath, filter).first : hashFile(hashAlgo, srcPath);
+    Hash h = static_cast<bool>(recursive) ? hashPath(hashAlgo, srcPath, filter).first : hashFile(hashAlgo, srcPath);
```
edolstra: `static_cast<bool>()` seems to negate the advantage of using an enum, since it requires knowing that `Recursive` casts to `true`. Maybe better to be explicit (`method == Recursive`).

Also the variable name `recursive` should be something like `ingestionMethod`.

src/libexpr/primops.cc
```diff
@@ -563,7 +563,7 @@ static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * *
 
     std::optional<std::string> outputHash;
     std::string outputHashAlgo;
-    bool outputHashRecursive = false;
+    FileIngestionMethod outputHashRecursive = FileIngestionMethod::Flat;
```
edolstra: Maybe change to `auto ingestionMethod = FileIngestionMethod::Flat;`.

src/libstore/path.hh
```diff
@@ -73,6 +73,11 @@ const size_t storePathHashLen = 32; // i.e. 160 bits
 /* Extension of derivations in the Nix store. */
 const std::string drvExtension = ".drv";
 
+enum struct FileIngestionMethod : bool {
```
edolstra: Maybe `FileIngestionMethod` should `PathIngestionMethod` since it's more than just files. Or maybe `PathCopyMode` to be a bit less wordy.

src/libstore/store-api.cc
Ericson2314: I thought changing the enum "superclass" would break this, making it easier to fix later, but it doesn't. I'll definitely fix now.

src/libstore/path.hh
Ericson2314: I wanted something shorter too, but naming it's hard:

 - `PathIngestionMethod` but we are ingesting the file contents, not just the paths

 - `PathCopyMode` sometimes there are no nars involved and we aren't copying anything.

Ericson2314: One thing that will help is I hope to make a `FileHash` struct for the pair of the ingestion method and the hash; that is a shorter name, and will make a lot of functions take 1 arg less.

Ericson2314: @edolstra Do you have a name in mind? If not, we've now addressed your other points so I think this ready to go?

src/nix/command.cc
```diff
@@ -31,21 +31,21 @@ void StoreCommand::run()
     run(getStore());
 }
 
-StorePathsCommand::StorePathsCommand(bool recursive)
+StorePathsCommand::StorePathsCommand(FileIngestionMethod recursive)
```
matthewbauer, line 34: This has nothing to do with file ingestion

matthewbauer: See how nix copy uses it to default to recursive copy operations.

edolstra:
Thanks, merged!

Ericson2314:
@edolstra thank you!!!

Issue #3278, opened 2019-12-19
504 was not retried while fetching nar


roberth:
Current behavior:

```
[...]
unable to download 'https://[redacted].cachix.org/nar/[redacted].nar.xz': HTTP error 504
[...]
```

Expected behavior:
 - perform retries
 - log the retries as warnings as usual

<details><summary>nix show-config</summary>

version: nix (Nix) 2.3

```
nix show-config 
allow-import-from-derivation = true
allow-new-privileges = false
allow-unsafe-native-code-during-evaluation = false
allowed-impure-host-deps = 
allowed-uris = 
allowed-users = *
auto-optimise-store = false
build-hook = /nix/store/...-nix-2.3/libexec/nix/build-remote
build-poll-interval = 5
build-users-group = nixbld
builders = 
builders-use-substitutes = false
compress-build-log = true
connect-timeout = 0
cores = 0
diff-hook = 
download-attempts = 5
enforce-determinism = true
extra-platforms = i686-linux
extra-sandbox-paths = 
extra-substituters = 
fallback = false
filter-syscalls = true
fsync-metadata = true
gc-check-reachability = false
gc-reserved-space = 8388608
hashed-mirrors = http://tarballs.nixos.org/
http-connections = 25
http2 = true
impersonate-linux-26 = false
keep-build-log = true
keep-derivations = true
keep-env-derivations = false
keep-failed = false
keep-going = false
keep-outputs = false
log-lines = 10
max-build-log-size = 0
max-free = 18446744073709551615
max-jobs = 4
max-silent-time = 0
min-free = 0
min-free-check-interval = 5
narinfo-cache-negative-ttl = 3600
narinfo-cache-positive-ttl = 2592000
netrc-file = /etc/nix/netrc
plugin-files = 
post-build-hook = 
pre-build-hook = 
print-missing = true
pure-eval = false
repeat = 0
require-sigs = true
restrict-eval = false
run-diff-hook = false
sandbox = true
sandbox-build-dir = /build
sandbox-dev-shm-size = 50%
sandbox-fallback = false
sandbox-paths = /bin/sh=/nix/store...-busybox-1.30.1/bin/busybox
secret-key-files = 
show-trace = false
stalled-download-timeout = 300
store = auto
substitute = true
substituters = https://cache.nixos.org/
sync-before-registering = false
system = x86_64-linux
system-features = benchmark big-parallel kvm nixos-test
tarball-ttl = 3600
timeout = 0
trace-function-calls = false
[...]

PR #3457, opened 2020-03-30
downloader: when retry but can't resume, retry the whole file

domenkozar:
I'm not sure I fully understand how code works, but I assume `init` is called for each retry and this is the correct way to sync up `warn` and the actual behavior of resuming.

Fixes https://github.com/NixOS/nix/issues/3278

I'd appreciate a maintenance release since a lot of users of Cachix are hitting this occasionally.

### DIFF
--- a/src/libstore/download.cc
+++ b/src/libstore/download.cc
@@ -308,7 +308,7 @@ struct CurlDownloader : public Downloader
             curl_easy_setopt(req, CURLOPT_NETRC_FILE, settings.netrcFile.get().c_str());
             curl_easy_setopt(req, CURLOPT_NETRC, CURL_NETRC_OPTIONAL);
 
-            if (writtenToSink)
+            if (writtenToSink && acceptRanges && encoding.empty())
                 curl_easy_setopt(req, CURLOPT_RESUME_FROM_LARGE, writtenToSink);
 
             result.data = std::make_shared<std::string>();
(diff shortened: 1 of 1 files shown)


edolstra:
I'm not sure this is correct. If we've already written to a sink, then we *can't* do a retry because we would end up writing the same data twice. Otherwise we get errors like "error 9 while decompressing xz file". (See 53247d6b116905e7233b1efd6c14845e20d27442.)

domenkozar:
Seems like CI fails due to https://github.com/NixOS/nix/commit/e322a16523c3c0644a900453abbe32bd0fd982e3

domenkozar:
@edolstra is there any way to reset the sink?

edolstra:
No. But if ranged requests are unavailable, we could discard the first `n` bytes (where `n` is the number of bytes previously written to the sink).

domenkozar:
Couldn't `download` raise an exception so that the caller would setup the sink again?

edolstra:
That was what 78fa47a7f08a4cb6ee7061bf0bd86a40e1d6dc91 tried to do but it was a lot more complicated.

domenkozar:
Actually, we should prevent from connecting the sink until we assert response status code is 2xx

edolstra:
@domenkozar But the status code already *is* 2xx. The problem is that the transmission of the body gets interrupted.

domenkozar:
I'm not so sure, https://github.com/NixOS/nix/issues/3278 is 504 and I've seen reports with `526 Invalid SSL Certificate`, so I'm a bit puzzled why those are not retried if the body wasn't consumed (and if it was 200, status code wouldn't be able to change?).

edolstra:
Hm, it's possible `writeCallback` is called with the 5xx body.

domenkozar:
Going to continue in https://github.com/NixOS/nix/issues/3278

Issue #2149, opened 2018-05-10
~/.config/nix/nix.conf overrides values of /etc/nix/nix.conf


domenkozar:
I'd expect `substituters` in `~/.config/nix/nix.conf` to be appended on top of the system values, but currently Nix overrides them.

LnL7:
There's `extra-substituters`, same for certain other options.

edolstra:
I'm not sure if append is desirable behaviour in general. What if you *want* to override? Also, should it prepend or append?

Maybe appending could be a special syntax, e.g.
```
substituters += https://...
```
or `--substituters +https://...`.

dezgeg:
For the command line syntax, the GNU command line guidelines for long options generally supports `--foo=bar` to be equivalent to `--foo bar`. So if we supported that, then `--substituters+=https://foo.bar` would be a quite logical extension.

ElvishJerricco:
Additionally, it'd be nice if there were ways to specify many config files other than the two default ones.

grahamc:
Configuration files can use `import` to load additional config files. Does that suit your needs, @ElvishJerricco?

ElvishJerricco:
@grahamc Had no idea.That's awesome. Is that documented? And I assume it still causes overwriting of options?

grahamc:
Oops, I meant `include` -- yes it is documented: https://nixos.org/nix/manual/#description-41

domenkozar:
Using `+=` seems fine to me as long as there's an equivalent for `--options substituters ...`

NorfairKing:
I just ran into a very confusing issue where I had a shell alias in a `shellHook` with multiple `--option trusted-public-keys` options.
Suddenly my machine was recompiling the linux kernel after a nixpkgs upgrade. I couldn't figure out what was wrong until @domenkozar pointed out that `--option trusted-public-keys` overrides instead of appending. So :+1: on this issue!

domenkozar:
This seems to be addressed with https://github.com/NixOS/nix/commit/ff4dea63c9403880500f82ce273713ecf793d2d9 (on master), but I still need to confirm that.

NorfairKing:
Not stale.

NorfairKing:
Not stale bot marked as not stale due to activity of the stale bot.

PR #3458, opened 2020-03-30
NIX_USER_CONF_FILES

zimbatm:
add NIX_USER_CONF_FILES
    
Motivation: maintain project-level configuration files.
    
Document the whole situation a bit better so that it corresponds to the
implementation, and add NIX_USER_CONF_FILES that allows overriding
which user files Nix will load during startup.

### DIFF
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -19,26 +19,30 @@
 
 <refsection><title>Description</title>
 
-<para>Nix reads settings from two configuration files:</para>
+<para>By default Nix reads settings from the following places:</para>
 
-<itemizedlist>
+<para>The system-wide configuration file
+<filename><replaceable>sysconfdir</replaceable>/nix/nix.conf</filename>
+(i.e. <filename>/etc/nix/nix.conf</filename> on most systems), or
+<filename>$NIX_CONF_DIR/nix.conf</filename> if
+<envar>NIX_CONF_DIR</envar> is set. Values loaded in this file are not forwarded to the Nix daemon. The
+client assumes that the daemon has already loaded them.
+</para>
 
-  <listitem>
-    <para>The system-wide configuration file
-    <filename><replaceable>sysconfdir</replaceable>/nix/nix.conf</filename>
-    (i.e. <filename>/etc/nix/nix.conf</filename> on most systems), or
-    <filename>$NIX_CONF_DIR/nix.conf</filename> if
-    <envar>NIX_CONF_DIR</envar> is set.</para>
-  </listitem>
+<para>User-specific configuration files:</para>
 
-  <listitem>
-    <para>The user configuration file
-    <filename>$XDG_CONFIG_HOME/nix/nix.conf</filename>, or
-    <filename>~/.config/nix/nix.conf</filename> if
-    <envar>XDG_CONFIG_HOME</envar> is not set.</para>
-  </listitem>
+<para>
+  If <envar>NIX_USER_CONF_FILES</envar> is set, then each path separated by
+  <literal>:</literal> will be loaded in reverse order.
+</para>
 
-</itemizedlist>
+<para>
+  Otherwise it will look for <filename>nix/nix.conf</filename> files in
+  <envar>XDG_CONFIG_DIRS</envar> and <envar>XDG_CONFIG_HOME</envar>.
+
+  The default location is <filename>$HOME/.config/nix.conf</filename> if
+  those environment variables are unset.
+</para>
 
 <para>The configuration files consist of
 <literal><replaceable>name</replaceable> =
--- a/doc/manual/command-ref/env-common.xml
+++ b/doc/manual/command-ref/env-common.xml
@@ -33,7 +33,7 @@
 
     will cause Nix to look for paths relative to
     <filename>/home/eelco/Dev</filename> and
-    <filename>/etc/nixos</filename>, in that order.  It is also
+    <filename>/etc/nixos</filename>, in this order.  It is also
     possible to match paths against a prefix.  For example, the value
 
     <screen>
@@ -59,7 +59,7 @@ nixpkgs=https://github.com/NixOS/nixpkgs-channels/archive/nixos-15.09.tar.gz</sc
     15.09 channel.</para>
 
     <para>A following shorthand can be used to refer to the official channels:
-    
+
     <screen>nixpkgs=channel:nixos-15.09</screen>
     </para>
 
@@ -137,12 +137,19 @@ $ mount -o bind /mnt/otherdisk/nix /nix</screen>
 
 <varlistentry><term><envar>NIX_CONF_DIR</envar></term>
 
-  <listitem><para>Overrides the location of the Nix configuration
+  <listitem><para>Overrides the location of the system Nix configuration
   directory (default
   <filename><replaceable>prefix</replaceable>/etc/nix</filename>).</para></listitem>
 
 </varlistentry>
 
+<varlistentry><term><envar>NIX_USER_CONF_FILES</envar></term>
+
+  <listitem><para>Overrides the location of the user Nix configuration files
+  to load from (defaults to the XDG spec locations). The variable is treated
+  as a list separated by the <literal>:</literal> token.</para></listitem>
+
+</varlistentry>
 
 <varlistentry><term><envar>TMPDIR</envar></term>
 
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -31,6 +31,7 @@ Settings::Settings()
     , nixLogDir(canonPath(getEnv("NIX_LOG_DIR").value_or(NIX_LOG_DIR)))
     , nixStateDir(canonPath(getEnv("NIX_STATE_DIR").value_or(NIX_STATE_DIR)))
     , nixConfDir(canonPath(getEnv("NIX_CONF_DIR").value_or(NIX_CONF_DIR)))
+    , nixUserConfFiles(getUserConfigFiles())
     , nixLibexecDir(canonPath(getEnv("NIX_LIBEXEC_DIR").value_or(NIX_LIBEXEC_DIR)))
     , nixBinDir(canonPath(getEnv("NIX_BIN_DIR").value_or(NIX_BIN_DIR)))
     , nixManDir(canonPath(NIX_MAN_DIR))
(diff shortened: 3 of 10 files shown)


domenkozar:
> The manual generation is currently failing. I tried understanding the error and bisecting the documentation and wasn't able to reduce the error to something actionable.

If this isn't a proof for https://github.com/NixOS/rfcs/pull/64 then I don't know what is :)

edolstra:
It's actually the opposite: DocBook at least has a concept of an "error". Markdown is garbage in, garbage out.

zimbatm:
I can fix the doc error with some more elbow grease. Can I get feedback on the PR instead? :D 

@edolstra does the PR make sense to you overall?

edolstra:
@zimbatm Yes, makes sense. However it may make more sense to add an environment variable `NIX_CONF_FILE` so you don't need to create a directory just for your configuration file. (E.g. a per-project config file could be in `<project-dir>/.nix.conf`).

edolstra:
Or generalize this to a *list* of configuration files (e.g. `NIX_CONF_FILES = my-project.conf /etc/nix/nix.conf ~/.config/nix/nix.conf `). After all, it may not be desirable to have the per-project config file completely override the user config file.

lheckemann:
What advantage does this have over using the existing XDG_CONF_DIRS support?

edolstra:
@lheckemann Good point, that might be the way to go.

roberth:
@lheckemann

> This is useful because sometimes it is desirable to be more selective than changing the config home of all XDG-respecting programs.

For example, you'll want your editor to read the usual configuration files.

lheckemann:
You don't need to change the config home (where programs write their config to), only add a dir to the config path (where they read their config from) — and if the extra entry in the config path only contains relevant config (i.e. nix config) it'll only affect nix.

andir:
On 06:52 01.04.20, Linus Heckemann wrote:
> You don't need to change the config home (where programs write their
> config to), only add a dir to the config path (where they read their
> config from) — and if the extra entry in the config path only contains
> relevant config (i.e. nix config) it'll only affect nix.

Yeah, I am just setting that as part of my .envrc:

> export XDG_CONFIG_DIRS="$(pwd)/.config:$XDG_CONFIG_DIRS"

It has served me well for a while now. If there was another variable on
the nix end it should probably just allow a list of files (as eelco
suggested).

roberth:
I was focused on `NIX_USER_CONF_DIR` so I didn't notice the plural in `XDG_CONFIG_DIRS`. This does seem adequate for most cases. If you don't want an extra ".config" directory in your project, it seems like you could generate it in a `shell.nix` dependency.

zimbatm:
@lheckemann `XDG_CONFIG_DIRS` is loaded before `XDG_CONFIG_HOME` as per the XDG spec. If the user has set any configuration they will take precedence over XDG_CONFIG_DIRS.

zimbatm:
A variation on the XDG idea would be to push `XDG_CONFIG_DIRS=$XDG_CACHE_HOME:$XDG_CONFIG_DIRS` and then override `XDG_CACHE_HOME=$PWD/config` where `$PWD` is the project root. Unfortunately, the first thing I noticed is that `git` only implements the XDG spec partially and doesn't look at `XDG_CONFIG_DIRS`. So I think this PR is really needed.

doc/manual/command-ref/conf-file.xml
```diff
@@ -29,13 +29,35 @@
...
+    </para>
+  </listitem>
+
+  <listitem>
+    <para>
+      Then finally, the user configuration file in
+      <filename>$NIX_USER_CONF_DIR/nix.conf</filename>.
+      where <envvar>NIX_USER_CONF_DIR</envvar> defaults to
+      <filename>$XDG_CONFIG_HOME/nix</filename>
+      and <envvar>XDG_CONFIG_HOME</envvar> defaults to
```
zimbatm, line 53: I found the XML error:
```suggestion
      and <envar>XDG_CONFIG_HOME</envar> defaults to
```

zimbatm:
I fixed the doc generation and added a minimal test. The last thing to address is @edolstra's comment on [`NIX_CONF_FILE` and `NIX_CONF_FILES`.](https://github.com/NixOS/nix/pull/3458#issuecomment-607181594).

zimbatm:
@edolstra the env has been renamed to NIX_CONF_FILES and now appends to the existing config paths.

matthewbauer:
Does this override the system configuration or just append to it? For instance, what happens when a local config has `substituters = ...`. Is https://cache.nixos.org/ still loaded?

zimbatm:
Keys get overridden and the last writer wins. For example, if `/etc/nix/nix.conf` sets the `substituters` key and `~/.config/nix/nix.conf` as well, then the user config will take precedence. It's a valid concern but I think that should be handled in a separate PR as it extends the scope of the work and is not a new issue introduced by this PR.

domenkozar:
That's https://github.com/NixOS/nix/issues/2149

zimbatm:
Who is able to give this PR another substantial review? Does it make sense how I changed the code structure to load all the config files?

doc/manual/command-ref/conf-file.xml
```diff
@@ -29,13 +29,36 @@
...
+      <filename>$XDG_CONFIG_HOME/nix/nix.conf</filename> where
+      <envar>XDG_CONFIG_HOME</envar> defaults to
+      <filename>$HOME/.config</filename>.
+    </para>
+  </listitem>
+
+  <listitem>
+    <para>
+      And finally, every files pointed to by
+      <envar>NIX_EXTRA_CONF_FILES</envar>, separated by <literal>:</literal>
```
PkmX, line 59: Is this implemented?

PkmX:
It also seems like the `NIX_CONF_FILES` as implemented are actually extra conf files that are read after `$XDG_CONFIG_DIRS` and `$XDG_CONFIG_HOME`, but I think @edolstra's comment about `NIX_CONF_FILES` is that it specifies the absolute list of conf files to be loaded ignoring `$NIX_CONF_DIR/nix.conf` and all users config files. This would fit my use case where my build system is essentially a wrapper over Nix and needs exact control over the config, and I don't want any system-wide or user-wide `nix.conf` to be loaded.

doc/manual/command-ref/conf-file.xml
zimbatm: yes? You can see it in action in the admittedly minimal test.

PkmX: It only uses `NIX_CONF_FILES`, not `NIX_EXTRA_CONF_FILES`.

zimbatm: oh ok. This is a typo. Fixed in 976ead5229d776a9559629b279bd7a27f9945111

doc/manual/command-ref/conf-file.xml
```diff
@@ -19,7 +19,7 @@
 
 <refsection><title>Description</title>
 
-<para>Nix reads settings from two configuration files:</para>
+<para>Nix reads settings from the following places, in that order:</para>
```
edolstra, line 22: 'in this order'

doc/manual/command-ref/conf-file.xml
```diff
@@ -29,13 +29,36 @@
     (i.e. <filename>/etc/nix/nix.conf</filename> on most systems), or
     <filename>$NIX_CONF_DIR/nix.conf</filename> if
     <envar>NIX_CONF_DIR</envar> is set.</para>
+
+    <para>Values loaded in that file are not forwarded to the nix-daemon. The
```
edolstra, line 33: 'in this file'

'Nix daemon'

doc/manual/command-ref/conf-file.xml
```diff
@@ -29,13 +29,36 @@
...
+    client assumes that the daemon has already loaded them.
+    </para>
+  </listitem>
+
+  <listitem>
+    <para>
+      For each folder in <envar>XDG_CONFIG_DIRS</envar>, separated by
```
edolstra, line 40: 'each directory'

doc/manual/command-ref/conf-file.xml
```diff
@@ -29,13 +29,36 @@
...
+      <literal>:</literal>.
+
+      Load <filename><replaceable>folder</replaceable>/nix/nix.conf</filename>.
```
edolstra, line 43: 'directory'

doc/manual/command-ref/conf-file.xml
```diff
@@ -29,13 +29,36 @@
...
+      The user configuration file in
+      <filename>$XDG_CONFIG_HOME/nix/nix.conf</filename> where
+      <envar>XDG_CONFIG_HOME</envar> defaults to
+      <filename>$HOME/.config</filename>.
+    </para>
+  </listitem>
+
+  <listitem>
+    <para>
+      And finally, every files pointed to by
```
edolstra, line 58: 'every file'

doc/manual/command-ref/conf-file.xml
```diff
@@ -29,13 +29,36 @@
...
+      <envar>NIX_CONF_FILES</envar>, separated by <literal>:</literal>
+      if that environment variable exists.
```
edolstra, line 60: This is phrased awkwardly (it's the *contents* of NIX_CONF_FILE that's separated by `:`).

doc/manual/command-ref/env-common.xml
```diff
@@ -137,12 +137,19 @@ $ mount -o bind /mnt/otherdisk/nix /nix</screen>
...
-  <listitem><para>Overrides the location of the Nix configuration
+  <listitem><para>Overrides the location of the system Nix configuration
   directory (default
   <filename><replaceable>prefix</replaceable>/etc/nix</filename>).</para></listitem>
 
 </varlistentry>
 
+<varlistentry><term><envar>NIX_CONF_FILES</envar></term>
+
+  <listitem><para>Specifies another set of nix configuration files to load
```
edolstra, line 148: `Nix`

doc/manual/command-ref/env-common.xml
```diff
@@ -137,12 +137,19 @@ $ mount -o bind /mnt/otherdisk/nix /nix</screen>
...
+  that override the default locations. Each file is seprated by
```
edolstra, line 149: 'separated'

src/libstore/globals.cc
```diff
@@ -77,11 +78,26 @@ void loadConfFile()
...
+
+    // Get all the extra user-specified configuration files
+    std::vector<Path> files = tokenizeString<std::vector<string>>(configFiles, ":");
+
+    // Append the expanded XDG config files
     auto dirs = getConfigDirs();
-    // Iterate over them in reverse so that the ones appearing first in the path take priority
-    for (auto dir = dirs.rbegin(); dir != dirs.rend(); dir++) {
-        globalConfig.applyConfigFile(*dir + "/nix/nix.conf");
+    for (auto dir = dirs.begin(); dir != dirs.end(); dir++) {
```
edolstra, line 96: `for (auto & dir : dirs)`

src/libutil/util.hh
```diff
@@ -387,6 +388,24 @@ string replaceStrings(const std::string & s,
 std::string rewriteStrings(const std::string & s, const StringMap & rewrites);
 
 
+/* Create a string from a Range (like std::vector), separated by "delimited" */
+template <typename Range, typename Value = typename Range::value_type>
+std::string join(Range const& elements, const char *const delimiter) {
```
edolstra, line 393: This looks a lot like the `concatStringsSep` function in util.cc.

doc/manual/command-ref/conf-file.xml
zimbatm: fixed!

src/libutil/util.hh
zimbatm: Indeed! Removed

src/libstore/globals.cc
zimbatm: much nicer

zimbatm:
This is ready for another round of review.

* I changed `NIX_CONF_FILES` to `NIX_USER_CONF_FILES` to be more specific and also indicate that it overrides the user configuration. 
* I changed the logic so that `NIX_USER_CONF_FILES` replaces the default locations where user config is being loaded. The system `/etc/nix/nix.conf` is still being loaded in any case as it plays a special role.
* Fixed the docs, probably introduced some more language issues

PR #3459, opened 2020-03-30
Backport libfetchers from the flakes branch

edolstra:
This provides a pluggable mechanism for defining new fetchers. It adds a builtin function `fetchTree` that generalizes existing fetchers like `fetchGit`, `fetchMercurial` and `fetchTarball`. `fetchTree` takes a
set of attributes, e.g.
```
fetchTree {
  type = "git";
  url = "https://example.org/repo.git";
  ref = "some-branch";
  rev = "abcdef...";
}
```
The existing fetchers are just wrappers around this. Note that the input attributes to `fetchTree` are the same as flake input specifications and flake lock file entries.

All fetchers share a common cache stored in `~/.cache/nix/fetcher-cache-v1.sqlite`. This replaces the ad hoc caching mechanisms in `fetchGit` and `download.cc` (e.g. `~/.cache/nix/{tarballs,git-revs*}`).

This also adds support for Git worktrees (c169ea59049f861aaba429f48b828d0820b74d1d).


src/libfetchers/git.cc
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    std::optional<Path> getSourcePath() const override
+    {
+        if (url.scheme == "file" && !ref && !rev)
+            return url.path;
+        return {};
+    }
+
+    void markChangedFile(std::string_view file, std::optional<std::string> commitMsg) const override
```
andir: It isn't really obvious why Nix now needs the ability to create commits / stage files. These aren't being used anywhere yet (in this PR at least).

src/libfetchers/github.cc
```diff
@@ -0,0 +1,216 @@
...
+        if (ref)
+            attrs.emplace("ref", *ref);
+        if (rev)
+            attrs.emplace("rev", rev->gitRev());
+        return attrs;
+    }
+
+    void clone(const Path & destDir) const override
+    {
+        std::shared_ptr<const Input> input = inputFromURL(fmt("git+ssh://git@github.com/%s/%s.git", owner, repo));
```
andir: This assumes the user has an SSH key setup for usage with github.com. Shouldn't it default to cloning public repositories via https? Also the clone methods do not appear to be used anywhere. Did I miss the usages?

Ericson2314:
I like the idea of a `libfetchers`; I thought I saw some stuff going in `libtstore` and this is definitely better. Are the new builtins behind an experimental feature?

edolstra:
Yes, `fetchTree` (the only new builtin) is behind an experimental feature.

Ericson2314:
Thanks for confirming!

edolstra:
@andir There were a couple of methods that are only used in the flakes branch. I've removed them.

PR #3472, opened 2020-04-04
Uses `StorePathHash` typedef to always remember the length

Ericson2314:


Ericson2314:
This isn't needed since #3702

PR #3476, opened 2020-04-06
Rename download to filetransfer

knl:
This change renames all instances of Download to FileTransfer. I initially was taken by surprise that `getDownloader->download()` would also do uploads, depending on whether `Download` struct has data field or not. Thus, in order to ease the cognitive burden of the future readers of the code, I went in and did a rename.

In addition, this change introduces a new method called `upload`, which is exactly the same as `download`, which in turns just enqueues the request and synchronously waits. This is to make it more readable at the call site what is happening.

All the changes will be squashed into a single one prior to (hopefully) approval and merging.


edolstra:
Could you rebase this?

knl:
> Could you rebase this?

@edolstra Done

domenkozar:
Once again :)

knl:
@domenkozar @edolstra Done, again :)

PR #3477, opened 2020-04-07
`nix run` using $SHELL as default command

Ninlives:
Since `nix run` just sets the `PATH`, I think it makes sense to use `$SHELL` as the default command for a more fluent workflow.

Additionally, the `nix dev-shell` uses option `--rcfile`, which is not compatible with all mainstream shells(I think zsh is one of the mainstream shell, is it?), and the rcfile usually contains bash-specific content, I think it is better to just use "bash" in dev-shell instead of "$SHELL". What's your opinion? Just discussion though, this is not part of this pr.

### DIFF
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -59,14 +59,14 @@ struct RunCommon : virtual Command
 
 struct CmdRun : InstallablesCommand, RunCommon, MixEnvironment
 {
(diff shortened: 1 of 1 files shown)


edolstra:
Yes, `nix dev-shell` should use bash. Other shells could use `nix print-dev-env` to get the environment variables from the builder.

PR #3480, opened 2020-04-09
Print float values with full precision.

phile314:
Instead of printing at most 6 digits, print as many digits as are
necessary to preserve float values exactly.

I did run into this when building a config file using nix which has float values requiring at least seven significant digits. This makes the values sometimes less pretty, but is less surprising and makes the float values more versatile.

See also http://www2.open-std.org/JTC1/SC22/WG21/docs/papers/2005/n1822.pdf 
> This max_digits10 number is  also the number of decimal digits required to avoid loss of accuracy when converting to a string and back to a floating-point representation, for example during serialization.


edolstra:
It's important to note that this change causes evaluation results to diverge between different versions of Nix. But maybe we shouldn't care because few people use floating point anyway.

phile314:
Is there anything I can do to help getting this merged?

phile314:
Ping

phile314:
Any news?

infinisil:
There is a PR related to this https://github.com/NixOS/nix/pull/6238, though it's kind of blocked on a solution like https://github.com/NixOS/rfcs/pull/137

Ericson2314:
See also https://github.com/NixOS/nix/issues/8259: we accidentally made float printing in JSON more precise.

Ericson2314:
Marking draft because this is old and there are conflicts.

Issue #3475, opened 2020-04-06
nix build occasionally crashes on identical invocations


andir:
I have an internal setup with Nix 2.3.1 (and 2.3.3) where running `nix build -f some.nix packagesA packagesB -o output` randomly fails. It takes some ~3 minutes to trigger in a `while true; nix build …` loop. During that time no changes to the inputs are done. 

The error manifests in either a log message about the current value not being of type a `tAttrs` or as a segmentation fault (since a random integer can't be converted to a proper type string).

When disabling the garbage collector by using `GC_DONT_GC` the error doesn't occur.

I am trying to build two attribute sets of packages. Each of these attribute sets has a few derivation. The error always occurs when nix tries to evaluate  the second attribute. I tried to reproduce it with simple shareable files but so far no success. We have two (larger) sets of packages where one of them (`packagesB`) depends on some expressions that were already evaluated as part of the first attribute (`packagesA`).

When I replace `some.nix` in the original invocation with a `proxy.nix` that just contains `import some.nix` the error also doesn't reproduce anymore. Maybe the GC behaves differently in this case?
When I swap the order on the cli the error is also gone. This probably means the references between the two sets are relevant?

The crash (or the erroneous log message) occurs in the `nix::findAlongAttrPath` function just after the type comparison:
https://github.com/NixOS/nix/blob/2097983218094fca194c64cb2275cc625f41a018/src/libexpr/attr-path.cc#L63-L67
```
src/libexpr/attr-path.cc:
59                 according to what is specified in the attrPath. */
60
61              if (apType == apAttr) {
62
63                  if (v->type != tAttrs) <---
64                      throw TypeError(
65                          format("the expression selected by the selection path '%1%' should be a set but is %2%")
66                          % attrPath % showType(*v));
67
68                  if (attr.empty())
> p v->type
$2 = nix::tBool   # should have been tAttrs
```

Here is the stack trace of a crash when `v->type` was something unexpected (e.g. random garbage, some pointer value, …):
```
#0  0x00007f3e763a215a in raise () from /nix/store/1ncwrl8bplq3xhmj8pxfkx4y0i90vmnx-glibc-2.30/lib/libc.so.6
#1  0x00007f3e7638c548 in abort () from /nix/store/1ncwrl8bplq3xhmj8pxfkx4y0i90vmnx-glibc-2.30/lib/libc.so.6
#2  0x00007f3e76e92be4 in nix::showType[abi:cxx11](nix::Value const&) [clone .cold] ()
[...]

PR #3492, opened 2020-04-13
SourceExprCommand: allocate the vSourceExpr via uncollectable memory

andir:
Previously the memory would occasionally be collected during eval since
the GC doesn't consider the member variable as alive / doesn't scan the
region of memory where the pointer lives.

By using the traceable_allocator<T> allocator provided by Boehm GC we
can ensure the memory isn't collected. It should be properly freed when
SourceExprCommand goes out of scope.

Related to #3475 

I've been running my reproducer for  >20min. Usually it would trigger the error within a few minutes. So far it looks very stable. Having spent some time on reading through the code  and the various GC docs I believe this might be the correct fix (or approach).

The `vSourceExpr` will be released when the current command (`CmdBulid`) goes out of scope. That is probably fine as that is basically the same as the lifetime of the entire program.

I would appreciate if we could get this into a stable (v2.3-branch) release.

### DIFF
--- a/src/nix/command.hh
+++ b/src/nix/command.hh
@@ -41,7 +41,7 @@ private:
 
     std::shared_ptr<EvalState> evalState;
 
-    Value * vSourceExpr = 0;
+    std::shared_ptr<Value> vSourceExpr;
 };
 
 enum RealiseMode { Build, NoBuild, DryRun };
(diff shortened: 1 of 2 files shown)


nagisa:
Can this be included into next point release?

PR #3494, opened 2020-04-14
log all ifd

cleverca22:

PR #3506, opened 2020-04-16
Add --include-ifd option to nix-instantiate

matthewbauer:
This allows use you to get all drvs that your Nix expression
references. Output looks like this:

```
  $ nix-instantiate --include-ifd
  /nix/store/xvplw4i56ys3j1lk4bxnr8r4725rb6c8-bauer-1.5.0.drv
  /nix/store/nk7j41ldlq39lmmsg5v1mlwwwpnqnbgk-README.drv
  /nix/store/pp84r1z7j7aqqarbwkq55h5asy35rs15-package-list.drv
```

This can be put into nix-store to get all paths that will be needed to build the Nix expression:

```
  $ nix-store -qR --include-outputs $(nix-instantiate --include-ifd) | grep -v \.drv$
  /nix/store/00crwk6z71l36av0gspzi0z410h8fq1r-libXtst-1.2.3.tar.bz2
  /nix/store/9krlzvny65gdc8s7kpb6lkx8cd02c25b-default-builder.sh
  /nix/store/c0sr4qdy8halrdrh5dpm7hj05c6hyssa-unpack-bootstrap-tools.sh
  ...
```

Comes from discussion in https://github.com/NixOS/nix/pull/3494.

/cc @domenkozar @Ericson2314

### DIFF
--- a/doc/manual/command-ref/nix-instantiate.xml
+++ b/doc/manual/command-ref/nix-instantiate.xml
@@ -162,6 +162,13 @@ input.</para>
 
   </varlistentry>
 
+  <varlistentry><term><option>--include-ifd</option></term>
+
+    <listitem><para>Print all imported derivations along with the
+    final derivation.</para></listitem>
+
+  </varlistentry>
+
 </variablelist>
 
 <variablelist condition="manpage">
--- a/src/libexpr/eval.hh
+++ b/src/libexpr/eval.hh
@@ -88,6 +88,9 @@ public:
 
     const ref<Store> store;
 
+    /* List of derivations that got imported. */
+    std::list<std::pair<string, string>> importedDrvs;
+
 private:
     SrcToStore srcToStore;
 
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -55,6 +55,9 @@ void EvalState::realiseContext(const PathSet & context)
         auto ctx = store->parseStorePath(decoded.first);
         if (!store->isValidPath(ctx))
             throw InvalidPathError(store->printStorePath(ctx));
+        if (ctx.isDerivation()) {
+            importedDrvs.push_back(decoded);
+        }
         if (!decoded.second.empty() && ctx.isDerivation()) {
             drvs.push_back(StorePathWithOutputs{ctx.clone(), {decoded.second}});
 
--- a/src/nix-instantiate/nix-instantiate.cc
+++ b/src/nix-instantiate/nix-instantiate.cc
@@ -28,7 +28,7 @@ enum OutputKind { okPlain, okXML, okJSON };
 
 void processExpr(EvalState & state, const Strings & attrPaths,
     bool parseOnly, bool strict, Bindings & autoArgs,
-    bool evalOnly, OutputKind output, bool location, Expr * e)
+    bool evalOnly, OutputKind output, bool location, bool includeIfd, Expr * e)
 {
     if (parseOnly) {
         std::cout << format("%1%\n") % *e;
@@ -79,6 +79,11 @@ void processExpr(EvalState & state, const Strings & attrPaths,
                 }
                 std::cout << fmt("%s%s\n", drvPath, (outputName != "out" ? "!" + outputName : ""));
             }
+            if (includeIfd) {
+                for (auto & i : state.importedDrvs) {
+                    std::cout << fmt("%s%s\n", i.first, (i.second != "out" ? "!" + i.second : ""));
+                }
+            }
         }
     }
 }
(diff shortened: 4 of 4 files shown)


edolstra:
Hm, I don't really want to add flags to the legacy commands (`nix-instantiate`, etc.). Especially when the output is hard to parse (e.g. if I understand this correctly, the output on stdout doesn't distinguish between the regular nix-instantiate output and the IFD derivations). The `nix` command is intended to have a `--json` flag whenever appropriate for non-trivial output.

domenkozar:
>  the output on stdout doesn't distinguish between the regular nix-instantiate output and the IFD derivations

I'm not sure why that's a requirement? To me it seems like a bonus :)

matthewbauer:
> Hm, I don't really want to add flags to the legacy commands (`nix-instantiate`, etc.). Especially when the output is hard to parse (e.g. if I understand this correctly, the output on stdout doesn't distinguish between the regular nix-instantiate output and the IFD derivations). The `nix` command is intended to have a `--json` flag whenever appropriate for non-trivial output.

Is there a `nix` command equivalent to `nix-instantiate` and `nix-store`? We could make this an option to `nix eval`, but we'd still need to get the final .drv through `nix-instantiate`

This keeps track of "importedDrvs" regardless of `--include-ifds`, so we can always add support for a new command if needed. For instance, maybe something like:

```
$ nix closure --dev ./default.nix
```

would be the equivalent of:

```
$ nix-store -qR --include-outputs $(nix-instantiate --include-ifd ./default.nix) | grep -v \.drv$
```

Infinisil:
Could this be used to work around https://github.com/NixOS/nix/issues/719?

Ericson2314:
Perhaps `libnixexpr` should store a lambda in `EvalState` that could do either the logging or the accumulation for #3494 or #3506. That adds the core missing feature to the libraries while punting on the user interface the executable should have.

matthewbauer:
> Could this be used to work around #719?

Yeah, you can register the .drv (or its output) as a gcroot I think, with something like `nix-store --add-result --indirect -r $(nix-instantiate --include-ifd)`. gc'ing ifd actually kind of makes sense - we have no live references to the imported derivation. Perhaps we could have Nix create a "super-drv" which records what IFD happened to produce a given derivation, and an accompanying `gc-keep-super-derivations = true`

Ericson2314:
Yeah I would just make a little file with all the drv paths and `addToStore` it with the proper references.

matthewbauer:
> Perhaps `libnixexpr` should store a lambda in `EvalState` that could do either the logging or the accumulation for #3494 or #3506. That adds the core missing feature to the libraries while punting on the user interface the executable should have.

I don't think there's any reason you can't do both #3494 and #3506.

Ericson2314:
@matthewbauer absolutely one could do both. Why my lambda thing you would write a lambda that both accumulated and logged.

matthewbauer:
> @matthewbauer absolutely one could do both. Why my lambda thing you would write a lambda that both accumulated and logged.
> 
> I just don't want to see nothing happen because we went back and forth on the exact user interface: clearly there is a strong desire for _some_ way to get this information from Nix.

Well it's not so useful because no one uses Nix as a library. I think it would be better to just get something working through one of the Nix commands. Adding this feature to `nix` is definitely doable, if only there was an equivalent to `nix-instantiate`!

hyperfekt:
Given that this is the second instance (compare #3086) someone has taken the time to implement something like this, there seems to be a considerable need.

alyssais:
> Hm, I don't really want to add flags to the legacy commands (nix-instantiate, etc.).

So no new Nix features can be stabilised until the experimental CLI has been?

domenkozar:
@edol :pray:

edolstra:
The new CLI has no equivalent for `nix-instantiate` because it tries to hide the concept of `.drv` files. (This would allow switching away from having to materialize the dependency graph in the filesystem as `.drv` files, and instead sending an in-memory representation of the graph to the daemon, for instance.)

The old CLI is full of flags that seem "bolted on" with little structure or organization (e.g. `nix-instantiate --eval` does something completely different from `nix-instantiate`). The new CLI should avoid those mistakes. Maybe we can come up with higher-level subcommands that directly address the use cases for which `--include-ifd` is intended.

There is also the issue of whether we should even *want* to enshrine the concept of IFD in the user interface. IFD has serious issues as you probably know.

matthewbauer:
> The new CLI has no equivalent for `nix-instantiate` because it tries to hide the concept of `.drv` files. (This would allow switching away from having to materialize the dependency graph in the filesystem as `.drv` files, and instead sending an in-memory representation of the graph to the daemon, for instance.)
> 
> The old CLI is full of flags that seem "bolted on" with little structure or organization (e.g. `nix-instantiate --eval` does something completely different from `nix-instantiate`). The new CLI should avoid those mistakes. Maybe we can come up with higher-level subcommands that directly address the use cases for which `--include-ifd` is intended.
> 
> There is also the issue of whether we should even _want_ to enshrine the concept of IFD in the user interface. IFD has serious issues as you probably know.

Yeah - avoiding .drv's would be nice. Can you take a look at https://github.com/NixOS/nix/pull/3523 which adds a `nix refs` command.

matthewbauer:
Closing in favor of #3523

PR #3511, opened 2020-04-17
[RFC] libstore: remove static initialisation

Mic92:
This is is an alternative approach to https://github.com/NixOS/nix/pull/2698
If you prefer this approach than I would go ahead and remove the other static intializers as well

### DIFF
--- a/src/libstore/http-binary-cache-store.cc
+++ b/src/libstore/http-binary-cache-store.cc
@@ -159,7 +159,7 @@ protected:
 
 };
 
-static RegisterStoreImplementation regStore([](
+OpenStore openHttpBinaryCacheStore = [](
     const std::string & uri, const Store::Params & params)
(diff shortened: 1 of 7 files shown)


matthewbauer:
@edolstra can answer for sure, but I think this is how plugins are registered in Nix.

edolstra:
Right, this breaks plugins.

Mic92:
@edolstra Could you please elobrate how this break plugins? If you take a look at my version, I kept the old API in place and only made in-tree stores non-static.

edolstra:
It's also much uglier. The current approach is much cleaner since it doesn't require all store implementations to be enumerated in a global header file.

Mic92:
The implementation can be moved to its own compilation unit if necessary.

Mic92:
https://github.com/NixOS/nix/pull/2698 is less intrusive.

PR #3517, opened 2020-04-19
Add '--save' option to nix-shell

raboof:
In my first few months of using Nix, I used nix-shell extensively
(it is great!), but had a hard time remembering the shell.nix
syntax.

I'd often find myself starting out by using nix-shell with '-p', making the
command line grow longer and longer as I added additional '-p' parameters
for that project.

I think it would be helpful to have a '--save' option that, instead of starting
a nix shell, produces a shell.nix corresponding to the current set of parameters.
When passed with no parameters, it'd still give me a nice shell.nix to get
started with without having to hunt for an example.

This PR is of course a straw man, very incomplete and perhaps not the right
approach in the first place. I just wanted to provide something more than a
idea.

Would this be interesting to explore further? Are there any particular
requirements or desired behaviour that I should take into account?

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -76,6 +76,7 @@ static void _main(int argc, char * * argv)
     auto packages = false;
     // Same condition as bash uses for interactive shells
     auto interactive = isatty(STDIN_FILENO) && isatty(STDERR_FILENO);
+    auto save = false;
     Strings attrPaths;
     Strings left;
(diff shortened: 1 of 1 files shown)


edolstra:
In general I'd prefer not to add new features to the old UI. New features should probably go into the `nix` command (e.g. `nix dev-shell`).

raboof:
OK, I haven't used `nix dev-shell` yet, will look into it :+1:

PR #3520, opened 2020-04-20
Try to substitute tree for all uses of fetchTree instead of just in flakes

mkenigs:
Trying to fix the FIXME in fetchTree.cc:86

Wasn't quite sure how having a nar hash interacts with resolved refs. Since narHash is optional, a resolved ref could be immutable, but then possibly not locked? But then it doesn't make sense to assume that fetchOrSubstituteTree returns a lockedRef:
https://github.com/NixOS/nix/blob/8c75621da6260d92459e105892751457e1659741/src/libexpr/flake/flake.cc#L210

Should that be treated as an immutable ref but not locked? Or should substitute only occur for locked refs?

### DIFF
--- a/src/libfetchers/fetchers.cc
+++ b/src/libfetchers/fetchers.cc
@@ -52,8 +52,38 @@ Attrs Input::toAttrs() const
     return attrs;
 }
 
+Tree Input::substituteTree(ref<Store> store) const
+{
+    assert(narHash);
+    auto storePath = store->makeFixedOutputPath(true, narHash.value(), "source");
+
+    store->ensurePath(storePath);
+
+    debug("using substituted/cached input '%s' in '%s'",
+        to_string(), store->printStorePath(storePath));
+
+    auto actualPath = store->toRealPath(storePath);
+
+    return {
+        Tree {
+            .actualPath = actualPath,
+            .storePath = std::move(storePath),
+            .info = { narHash.value() }
+        }
+    };
+}
+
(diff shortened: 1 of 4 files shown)


src/libfetchers/fetchers.cc
```diff
@@ -52,8 +52,38 @@ Attrs Input::toAttrs() const
...
+        }
+    };
+}
+
 std::pair<Tree, std::shared_ptr<const Input>> Input::fetchTree(ref<Store> store) const
 {
+    if (narHash) {
+        try {
+            auto tree = substituteTree(store);
+            return {std::move(tree), shared_from_this()};
```
edolstra, line 81: Unfortunately, this is not correct because this will return a different `Tree` and `Input` than `fetchTreeInternal()`. For instance, for a Git input, `fetchTreeInternal()` will return an `Input` that has `rev` set, and a `Tree` that has `revCount` and `lastModified` set.

So we can only substitute the store path if we can get those `Input` / `Tree` attributes in some other way. It works for flake inputs because `flake.lock` provides the missing attributes.

mkenigs: Ok makes sense. So basically a locked ref always needs to be returned? Not that you couldn't substitute an unlocked ref with a narHash, but that wouldn't allow things like lockfile creation? If so was wondering why LockedRef isn't a subclass, which fetchTree could then return.

PR #3522, opened 2020-04-21
Replace select() with poll()

HackerFoo:
This change fixes fixes a bug causing `nix-build` to abort with `reached FD_SETSIZE limit` when distributing ~400 jobs across 5 build machines.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -33,7 +33,6 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/utsname.h>
-#include <sys/select.h>
 #include <sys/resource.h>
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


src/libstore/build.cc
```diff
@@ -4789,8 +4789,7 @@ void Worker::waitForInput()
        terminated. */
 
     bool useTimeout = false;
-    struct timeval timeout;
-    timeout.tv_usec = 0;
+    int timeout = 0;
```
edolstra, line 4792: This should be `long` to be consistent with `std::max(1L, (long) ...)`.

HackerFoo: Fixed, thanks.

edolstra:
Thanks!

PR #3523, opened 2020-04-22
Add --include-build-refs to `nix path-info' command

matthewbauer:
This print out all of the store paths referenced in the Nix
expression.

This is useful with tools like cachix, and replace the previous usages
of:

```
  $ nix-store -qR --include-outputs $(nix-instantiate '<nixpkgs>' -A hello)
```

which now become:

```
  $ nix path-info --recursive --include-build-refs nixpkgs.hello
```

which can be piped into cachix like:

```
  $ nix path-info --include-build-refs nixpkgs.hello | cachix push my-cache
```

It provides a flag for ` --include-build-refs`, meant to parallel dependencies needed for `nix build`.

Related to cachix/cachix#52
and #3506

/cc @domenkozar @edolstra


edolstra:
Maybe these could be flags to `nix path-info`? We already have `nix path-info -r` to show the runtime dependencies.

matthewbauer:
> Maybe these could be flags to `nix path-info`? We already have `nix path-info -r` to show the runtime dependencies.

That might make sense, but right now `nix path-info` wants store paths. To calculate, build and eval-time dependencies, we need an expression.

For build-time dependencies, we actually could use the "deriver" field to get that derivation, but it's frequently missing. Also eval-time will always need an expression (unless we add something like a "deriver-expression" field)

matthewbauer:
Perhaps we could rename it to `nix expr-info`?

gleber:
Maybe call it `nix info` and make it take both expressions and nix store
paths?

edolstra:
`nix path-info` (like most `nix` commands) works on a lot more than store paths:
```
$ nix path-info nixpkgs#hello
/nix/store/y26qxcq1gg2hrqpxdc58b2fghv2bhxjg-hello-2.10

$ nix path-info -f ~/Dev/nixpkgs hello
/nix/store/7hz2ms6mxdbfwj8bl4fcd171lly4l9wx-hello-2.10

$ nix path-info --impure --expr '(import <nixpkgs> {}).hello'
/nix/store/rr3y0c6zyk7kjjl8y19s4lsrhn4aiq1z-hello-2.10
```

matthewbauer:
Updated to use `nix path-info`, instead of adding a new command.

src/nix/command.cc
```diff
@@ -61,15 +66,48 @@ void StorePathsCommand::run(ref<Store> store)
...
+            if (includeEvalDeps) {
+                for (auto & i : installables) {
+                    auto state = getEvalState();
+
+                    for (auto & b : i->toBuildables()) {
+                        if (!b.drvPath)
+                            throw UsageError("Cannot find eval references without a derivation path");
+                    }
+
+                    // force evaluation of package argument
```
matthewbauer, line 85: @edolstra Is there a way to force evaluation, even when caching is possible? I need importedDrvs to be re-evaluated to get the right value.

src/nix/command.cc
```diff
@@ -61,15 +66,48 @@ void StorePathsCommand::run(ref<Store> store)
     }
 
     else {
+        if (!recursive && (includeBuildDeps || includeEvalDeps))
+            throw UsageError("--build and --eval require --recursive");
```
domenkozar, line 70: Would it be better to just assume `--recursive`?

src/nix/command.cc
```diff
@@ -61,15 +66,48 @@ void StorePathsCommand::run(ref<Store> store)
...
             storePaths.push_back(p.clone());
 
         if (recursive) {
+            if (includeEvalDeps) {
+                for (auto & i : installables) {
+                    auto state = getEvalState();
+
+                    for (auto & b : i->toBuildables()) {
+                        if (!b.drvPath)
+                            throw UsageError("Cannot find eval references without a derivation path");
```
domenkozar, line 82: Print what's `b`

src/nix/command.cc
```diff
@@ -61,15 +66,48 @@ void StorePathsCommand::run(ref<Store> store)
...
+                    for (auto & d : (*state).importedDrvs)
+                        storePaths.push_back(store->parseStorePath(d.first));
+                }
+            }
+
+            if (includeBuildDeps) {
+                for (auto & i : installables) {
+                    for (auto & b : i->toBuildables()) {
+                        if (!b.drvPath) // Note we could lookup deriver from the DB to get drvPath
+                            throw UsageError("Cannot find build references without a derivation path");
```
domenkozar, line 97: same here

src/nix/command.cc
matthewbauer: Perhaps, although I'm not sure how to handle `--no-recursive --build` or `--no-recursive --eval`. /That/ should be an error, but I don't think we have a way to tell whether that flag is set. Adding the assertion here is easier.

src/nix/command.cc
matthewbauer: There's not a good way to print what b is without the drv path. We could show just "out" path, I guess:

https://github.com/NixOS/nix/blob/c9d0cf7e02d4b3af3c027d7d74d5caa3c8963d26/src/nix/installables.hh#L13-L14

matthewbauer: (but we don't have the store here so we can't know location nix store is at

domenkozar: Printing at least the outputs would be desired otherwise there's no way to recover from this error.

src/nix/command.cc
domenkozar: There's no `--no-recursive` flag so I'm a bit confused :)

matthewbauer: `nix copy` will set recursive=true so that https://github.com/NixOS/nix/blob/c9d0cf7e02d4b3af3c027d7d74d5caa3c8963d26/src/nix/command.cc#L39 is active

src/nix/command.cc
matthewbauer: Added so the error message is now:

```
error: Cannot find build references for 'hello-2.10' without a derivation path
```

matthewbauer:
@edolstra Let me know if any concerns remain. It should be ready for re-review.

matthewbauer:
Updated to support `--build` with plain output paths. Also refactored data types a bit.

src/libexpr/primops.cc
```diff
@@ -56,6 +56,7 @@ void EvalState::realiseContext(const PathSet & context)
             throw InvalidPathError(store->printStorePath(ctx));
         if (!decoded.second.empty() && ctx.isDerivation()) {
             drvs.push_back(StorePathWithOutputs{ctx.clone(), {decoded.second}});
+            importedDrvs.push_back(StorePathWithOutputs{ctx.clone(), {decoded.second}});
```
Ericson2314, line 59: minor nit: Wait maybe instead of `push_back`ing each only, we can insert the entirety of `drv` at once?

matthewbauer: originally i had:

```
importedDrvs.insert(importedDrvs.end(), drvs.begin(), drvs.end());
```

but that gave me this:

```
In file included from src/libexpr/primops.cc:1:
In file included from /Users/mbauer/Code/nix/precompiled-headers.h:1:
/nix/store/4i88rry2ckc928by1dzg0q9w98c9n5dr-libc++-7.1.0/include/c++/v1/algorithm:1919:21: error: 
      object of type 'nix::StorePathWithOutputs' cannot be assigned because its copy assignment
      operator is implicitly deleted
        *--__result = _VSTD::move(*--__last);
                    ^
/nix/store/4i88rry2ckc928by1dzg0q9w98c9n5dr-libc++-7.1.0/include/c++/v1/algorithm:1948:19: note: 
      in instantiation of function template specialization
      'std::__1::__move_backward<nix::StorePathWithOutputs *, nix::StorePathWithOutputs *>'
      requested here
    return _VSTD::__move_backward(__unwrap_iter(__first), __unwrap_iter(__last), __unwrap_...
                  ^
/nix/store/4i88rry2ckc928by1dzg0q9w98c9n5dr-libc++-7.1.0/include/c++/v1/vector:1757:12: note: in
      instantiation of function template specialization
      'std::__1::move_backward<nix::StorePathWithOutputs *, nix::StorePathWithOutputs *>'
      requested here
    _VSTD::move_backward(__from_s, __from_s + __n, __old_last);
           ^
/nix/store/4i88rry2ckc928by1dzg0q9w98c9n5dr-libc++-7.1.0/include/c++/v1/vector:2019:17: note: in
      instantiation of member function 'std::__1::vector<nix::StorePathWithOutputs,
      std::__1::allocator<nix::StorePathWithOutputs> >::__move_range' requested here
                __move_range(__p, __old_last, __p + __old_n);
                ^
src/libexpr/primops.cc:74:18: note: in instantiation of function template specialization
      'std::__1::vector<nix::StorePathWithOutputs, std::__1::allocator<nix::StorePathWithOutputs>
      >::insert<std::__1::__wrap_iter<nix::StorePathWithOutputs *> >' requested here
    importedDrvs.insert(importedDrvs.end(), drvs.begin(), drvs.end());
                 ^
src/libstore/path.hh:78:15: note: copy assignment operator of 'StorePathWithOutputs' is implicitly
      deleted because field 'path' has a deleted copy assignment operator
    StorePath path;
              ^
src/libstore/path.hh:19:20: note: copy assignment operator of 'StorePath' is implicitly deleted
      because base class 'rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>' has a deleted
      copy assignment operator
struct StorePath : rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>
                   ^
src/libutil/rust-ffi.hh:35:5: note: copy assignment operator is implicitly deleted because
      'Value<48, &nix::ffi_StorePath_drop>' has a user-declared move constructor
    Value(Value && other)
    ^
In file included from src/libexpr/primops.cc:1:
In file included from /Users/mbauer/Code/nix/precompiled-headers.h:1:
/nix/store/4i88rry2ckc928by1dzg0q9w98c9n5dr-libc++-7.1.0/include/c++/v1/algorithm:1744:19: error: 
      object of type 'nix::StorePathWithOutputs' cannot be assigned because its copy assignment
      operator is implicitly deleted
        *__result = *__first;
                  ^
/nix/store/4i88rry2ckc928by1dzg0q9w98c9n5dr-libc++-7.1.0/include/c++/v1/algorithm:1769:19: note: 
      in instantiation of function template specialization
      'std::__1::__copy<std::__1::__wrap_iter<nix::StorePathWithOutputs *>,
      nix::StorePathWithOutputs *>' requested here
    return _VSTD::__copy(__unwrap_iter(__first), __unwrap_iter(__last), __unwrap_iter(__result));
                  ^
/nix/store/4i88rry2ckc928by1dzg0q9w98c9n5dr-libc++-7.1.0/include/c++/v1/vector:2021:24: note: in
      instantiation of function template specialization
      'std::__1::copy<std::__1::__wrap_iter<nix::StorePathWithOutputs *>,
      nix::StorePathWithOutputs *>' requested here
                _VSTD::copy(__first, __m, __p);
                       ^
src/libexpr/primops.cc:74:18: note: in instantiation of function template specialization
      'std::__1::vector<nix::StorePathWithOutputs, std::__1::allocator<nix::StorePathWithOutputs>
      >::insert<std::__1::__wrap_iter<nix::StorePathWithOutputs *> >' requested here
    importedDrvs.insert(importedDrvs.end(), drvs.begin(), drvs.end());
                 ^
src/libstore/path.hh:78:15: note: copy assignment operator of 'StorePathWithOutputs' is implicitly
      deleted because field 'path' has a deleted copy assignment operator
    StorePath path;
              ^
src/libstore/path.hh:19:20: note: copy assignment operator of 'StorePath' is implicitly deleted
      because base class 'rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>' has a deleted
      copy assignment operator
struct StorePath : rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>
                   ^
src/libutil/rust-ffi.hh:35:5: note: copy assignment operator is implicitly deleted because
      'Value<48, &nix::ffi_StorePath_drop>' has a user-declared move constructor
    Value(Value && other)
    ^
2 errors generated.
make: *** [mk/patterns.mk:3: src/libexpr/primops.o] Error 1
```

src/nix/command.cc
```diff
@@ -61,15 +66,49 @@ void StorePathsCommand::run(ref<Store> store)
...
+                    auto state = getEvalState();
+
+                    for (auto & b : i->toBuildables())
+                        if (!b.drvPath)
+                            throw UsageError("Cannot find eval references for '%s' without a derivation path", b.what());
+
+                    // force evaluation of package argument
+                    i->toValue(*state);
+
+                    for (auto & d : (*state).importedDrvs)
```
Ericson2314, line 87:
```suggestion
                    for (auto & d : state->importedDrvs)
```

src/libexpr/primops.cc
Ericson2314: Ughhhh once again we need #3450

Ericson2314:
Hmm i actually think `--eval` alone should get all imported paths, so e.g. `import (builtins.fetchurl ...)` doesn't slip under the radar. The combination of `--eval` and `--build` can turn those paths into derivations where possible.

matthewbauer:
> Hmm i actually think `--eval` alone should get all imported paths, so e.g. `import (builtins.fetchurl ...)` doesn't slip under the radar. The combination of `--eval` and `--build` can turn those paths into derivations where possible.

Good idea! Those paths are needed for evaluation even if they aren't actually derivations.

src/libexpr/primops.cc
```diff
@@ -54,6 +54,7 @@ void EvalState::realiseContext(const PathSet & context)
         auto ctx = store->parseStorePath(decoded.first);
         if (!store->isValidPath(ctx))
             throw InvalidPathError(store->printStorePath(ctx));
+        importedPaths.push_back(ctx.clone());
```
Ericson2314, line 57: I think we would want to push the path to the specific output rather than drv file in this case? for `--build --eval `would we get the original drv file again.

src/libexpr/primops.cc
```diff
@@ -54,6 +54,7 @@ void EvalState::realiseContext(const PathSet & context)
         auto ctx = store->parseStorePath(decoded.first);
         if (!store->isValidPath(ctx))
             throw InvalidPathError(store->printStorePath(ctx));
+        realisedPaths.push_back(ctx.clone());
         if (!decoded.second.empty() && ctx.isDerivation()) {
             drvs.push_back(StorePathWithOutputs{ctx.clone(), {decoded.second}});
```
Ericson2314, line 60: Err... don't we need to do something like this?
```suggestion
        if (decoded.second.empty()) {
            realisedPaths.push_back(ctx.clone());
        } else {
            assert (ctx.isDerivation());
            drvs.push_back(ctx.clone(), {decoded.second}})
            // ... (can't include in suggestion)
            realisedPaths.push_back(i->second.path);
```

matthewbauer: i'm unsure what you mean, but I think in this case just the part before the `!` is fine. computeFSClosure doesn't care about outputs.

Ericson2314: Ah! I forgot we were requiring `recursive`

Ericson2314: Ah! I forgot we were requiring `recursive`

src/libexpr/primops.cc
matthewbauer: It may be useful to have more information, but this is all that's needed for the computeFSClosure. The original drv needs to be included to get dependencies for `--build`. It may end up getting in multiple times, but I think that's okay.

From counting the number included, I am fairly sure it is the right thing. For instance, here's the deps I get for each:

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-right" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">recursive</th>
<th scope="col" class="org-left">evaltime</th>
<th scope="col" class="org-left">buildtime</th>
<th scope="col" class="org-left">runtime</th>
<th scope="col" class="org-right">total</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">Y</td>
<td class="org-left">N</td>
<td class="org-left">N</td>
<td class="org-left">N</td>
<td class="org-right">0</td>
</tr>


<tr>
<td class="org-left">Y</td>
<td class="org-left">Y</td>
<td class="org-left">N</td>
<td class="org-left">N</td>
<td class="org-right">218</td>
</tr>


<tr>
<td class="org-left">Y</td>
<td class="org-left">N</td>
<td class="org-left">N</td>
<td class="org-left">Y</td>
<td class="org-right">293</td>
</tr>


<tr>
<td class="org-left">Y</td>
<td class="org-left">Y</td>
<td class="org-left">N</td>
<td class="org-left">N</td>
<td class="org-right">511</td>
</tr>


<tr>
<td class="org-left">Y</td>
<td class="org-left">N</td>
<td class="org-left">Y</td>
<td class="org-left">X</td>
<td class="org-right">1003</td>
</tr>


<tr>
<td class="org-left">Y</td>
<td class="org-left">Y</td>
<td class="org-left">Y</td>
<td class="org-left">X</td>
<td class="org-right">1008</td>
</tr>
</tbody>
</table>

src/libexpr/primops.cc
matthewbauer: or it doesn't compute dependencies of outputs at least

Ericson2314: I can't marked resolved but I'm satisfied as to all these.

matthewbauer:
Does anyone have objections to the naming of the options here? `--eval`, `--build`, and `--no-run` make sense to me but I realized it is a little confusing because you're not actually doing those things - just recursing into them. Perhaps `--recursive-eval`, `--recursive-build` are better... I also thought about something like `--closure-type`.

edolstra:
I'm hesitant about adding these flags:

* Build dependencies expose the concept of store derivations, which the `nix` command tries to move away from (so that in the future, we could stop materializing them to disk, and instead just send an in-memory graph of build actions to the daemon).

* Evaluation-time dependencies are unpredictable: if evaluation caching is enabled (the default), then `nix path-info` could output different results when you run it a second time.

Ericson2314:
1. Sorry I must be missing something. Build-time dependencies still exists as a concept, whether or not we have the drv files. Is the issue using drv file paths to refer to derivations? Or just that the daemon might not remember how something was built?

2. Could we put this info in the evaluation cache so it can be retrieved later? At the end of the day, we (and I assume others) will be happy with any approve that allows one to be sure they are saving enough dependencies to make importing derivations reproducible----this is a feature we really want. If there is a different approach you would prefer, we'd be happy to try that instead.

matthewbauer:
Updated to just provide `--build` since `--eval` could have some bad interactions with the eval cache.

domenkozar:
@edolstra there's no way to do this currently with flakes, which really limits ways to inspect closures for build dependencies.

If we stop materializing to disk, we can still compute the build closure it just can't be cached?

hyperfekt:
Can `--eval` not be made to run without the evaluation cache? I'm thinking it's still going to be very useful.

matthewbauer:
@domenkozar @edolstra Any chance this can get some more review? I've updated on to master and now renamed it to --include-build-refs. I'm hoping the longer, more specific name means it's a little less objectionable and more obvious how it fits into the --recursive flag flow.

domenkozar:
@matthewbauer does it address concerns raised by @edolstra? I'm +1 for merging it otherwise.

src/libexpr/primops.cc
tomberek: What standard commands or flags would enumerate all of these, but it can be unclear how to get a report/listing of each row. How was this generated?

PR #3528, opened 2020-04-23
Implementation of a CAS store

regnat:
New implementation of NixOS/rfcs#62 (replaces the old prototype from https://github.com/NixOS/nix/pull/3262)

Currently at a very early stage:

- Building a cas derivation locally works
- Derivations can depend (transitively) on CAS derivations and the cutoff will be correctly done when possible
- Most tests fail, probably just because the output path returned by most commands isn't the right one anymore
- Nothing is implemented on the remote {building,caching} side

/cc @edolstra @Ericson2314 @layus

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1226,6 +1226,9 @@ void DerivationGoal::outputsSubstituted()
 {
     trace("all outputs substituted (maybe)");
 
+    if (useDerivation)
+        worker.store.resolveDerivation(*(dynamic_cast<Derivation *>(drv.get())));
+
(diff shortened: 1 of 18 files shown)


src/libstore/build.cc
```diff
@@ -799,6 +799,9 @@ class DerivationGoal : public Goal
     /* Whether this is a fixed-output derivation. */
     bool fixedOutput;
 
+    /* Whether this is a content adressed derivation */
+    bool contentAddressed = false;
```
Ericson2314, line 803: @regnat Sorry I have not responded to this PR more thoroughly yet.

First thing I wanted to point out is I made https://github.com/NixOS/nix/pull/3418 so that we might talk about the various independent-ish aspects of derivations, rather than keeping around multiple bools of which only 1 should be true.

PR #3537, opened 2020-04-25
nix-env: refactor uninstallDerivations

### LINKED ISSUE #3531: nix-env -e silently ignores errors
**Problem**

If an invalid package name is given to `nix-env -i` or `nix-env -u`, an error-message is printed and nix-env exits with an exit-code !=0.

But `nix-env -e` silently ignores invalid package names, and always exits with exit-code 0.
But silently ignoring errors is a bad idea.

**Solution**

- Print an error-message for each invalid/not installed package, given to `nix-env -e`.
- Exit with exit code != 0, if there was any error.

lheckemann:
Reduces the number of store queries it performs. Also prints a warning
if any of the selectors did not match any installed derivations.

UX Caveats:
- Will print a warning that nothing matched if a previous selector
  already removed the path
- Will not do anything if no selectors were provided (no change from
  before).

Fixes #3531

### DIFF
--- a/src/nix-env/nix-env.cc
+++ b/src/nix-env/nix-env.cc
@@ -718,28 +718,39 @@ static void uninstallDerivations(Globals & globals, Strings & selectors,
     while (true) {
         string lockToken = optimisticLockProfile(profile);
 
-        DrvInfos installedElems = queryInstalled(*globals.state, profile);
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks!

PR #3541, opened 2020-04-27
Fix long paths permanently breaking GC

alyssais:
Suppose I have a path /nix/store/[hash]-[name]/a/a/a/a/a/[...]/a,
long enough that everything after "/nix/store/" is longer than 4096
(MAX_PATH) bytes.

Nix will happily allow such a path to be inserted into the store,
because it doesn't look at all the nested structure.  It just cares
about the /nix/store/[hash]-[name] part.  But, when the path is deleted,
we encounter a problem.  Nix will move the path to /nix/store/trash, but
then when it's trying to recursively delete the trash directory, it will
at some point try to unlink /nix/store/[hash]-[name]/a/a/a/a/a/[...]/a.
This will fail, because the path is too long.  After this has failed,
any store deletion operation will never work again, because Nix needs to
delete the trash directory before recreating it to move new things to
it.  (I assume this is because otherwise a path being deleted could
already exist in the trash, and then moving it would fail.)

This means that if I can trick somebody into just fetching a tarball
containing a path of the right length, they won't be able to delete
store paths or garbage collect ever again, until the offending path is
manually removed from /nix/store/trash.  (And even fixing this manually
is quite difficult if you don't understand the issue, because the
absolute path that Nix says it failed to remove is also too long for
rm(1).)

This patch fixes the issue by making Nix's recursive delete operation
use unlinkat(2).  This function takes a relative path and a directory
file descriptor.  We ensure that the relative path is always just the
name of the directory entry, and therefore its length will never exceed
255 bytes.  This means that it will never even come close to AX_PATH,
and Nix will therefore be able to handle removing arbitrarily deep
directory hierachies.

Since directory file descriptors are being used outside of just
readDirectory, I made a variant of readDirectory that takes an already
open directory stream, to avoid the directory being opened multiple
times.  As we have seen from this issue, the less we have to interact
with paths, the better, and so it's good to reuse file descriptors where
possible.

---

Thanks to @puckipedia.

Checked on IRC whether I should report here or not given security implications.
[Tarball THAT WILL BREAK YOUR GARBAGE COLLECTION](https://github.com/NixOS/nix/files/4540543/longpath.tar.gz)

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -268,16 +268,13 @@ bool isLink(const Path & path)
 }
 
 
-DirEntries readDirectory(const Path & path)
+DirEntries readDirectory(DIR *dir, const Path & path)
 {
     DirEntries entries;
     entries.reserve(64);
 
-    AutoCloseDir dir(opendir(path.c_str()));
-    if (!dir) throw SysError(format("opening directory '%1%'") % path);
-
     struct dirent * dirent;
-    while (errno = 0, dirent = readdir(dir.get())) { /* sic */
+    while (errno = 0, dirent = readdir(dir)) { /* sic */
         checkInterrupt();
         string name = dirent->d_name;
         if (name == "." || name == "..") continue;
@@ -292,6 +289,14 @@ DirEntries readDirectory(const Path & path)
     if (errno) throw SysError(format("reading directory '%1%'") % path);
 
     return entries;
+}
+
+DirEntries readDirectory(const Path & path)
+{
+    AutoCloseDir dir(opendir(path.c_str()));
+    if (!dir) throw SysError(format("opening directory '%1%'") % path);
+
+    return readDirectory(dir.get(), path);
 }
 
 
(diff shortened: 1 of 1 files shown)


alyssais:
Looking into CI failures

src/libutil/util.cc
```diff
@@ -389,12 +394,20 @@ void writeLine(int fd, string s)
...
 
 
-static void _deletePath(const Path & path, unsigned long long & bytesFreed)
+static void _deletePath(int parentfd, const Path & path, unsigned long long & bytesFreed)
 {
     checkInterrupt();
 
+    string name(baseNameOf(path));
+
+    AutoCloseFD fd = openat(parentfd, name.c_str(), 0);
```
edolstra, line 403: May be a good idea to use `O_NOFOLLOW` here.

edolstra:
Looks good to me, thanks!

There are probably a bunch of other places where we should use the `*at()` APIs to avoid hitting the path size limit, e.g. in NAR (de)serialization.

src/libutil/util.cc
alyssais: Hit this while I was trying to fix the installcheck :D

alyssais:
installcheck passes locally for me now

PR #3542, opened 2020-04-27
Set GCROOT to store path to prevent garbage collection

mkenigs:
Running `nix-collect-garbage` messes up a `nix dev-shell`. For example in `nix dev-shell nixpkgs.hello`, patchelf is gone after garbage collection.

nix-shell leaves buildInputs in an environment variable, which prevents garbage collection. `nix-store --gc --print-roots` shows `/proc/*/environ`

This PR adds something similar for dev-shell

### DIFF
--- a/src/nix/shell.cc
+++ b/src/nix/shell.cc
@@ -279,7 +279,7 @@ struct CmdDevShell : Common, MixEnvironment
 
         setEnviron();
         // prevent garbage collection until shell exits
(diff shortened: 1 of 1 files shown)


edolstra:
Can you rename this to `NIX_GCROOT`? Otherwise looks good.

mkenigs:
Done

PR #3544, opened 2020-04-28
Allow empty hashes

matthewbauer:
This allows hash="" so that it can be used for debugging purposes. For
instance, this gives you an error message like:

warning: found empty hash, assuming you wanted 'sha256:0000000000000000000000000000000000000000000000000000'
hash mismatch in fixed-output derivation '/nix/store/jx3gikmipizpk181cgfa1l4wwcamy6p0-nixpkgs':
  wanted: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
  got:    sha256-Wm7FDDnzNteClYOn+LRCtPNcK9wjEtpj+k9IIfygD7o=

### DIFF
--- a/src/libutil/hash.cc
+++ b/src/libutil/hash.cc
@@ -201,6 +201,10 @@ Hash::Hash(const std::string & s, HashType type)
         memcpy(hash, d.data(), hashSize);
     }
 
+    else if (s == "") {
+        printError("warning: found empty hash, assuming you wanted '%s'", this->to_string());
+    }
+
     else
         throw BadHash("hash '%s' has wrong length for hash type '%s'", s, printHashType(type));
 }


NorfairKing:
@matthewbauer I think the only argument against this is that 0 may be a valid hash (I don't think it's likely, but still). 
Could we instead just always print the gotten hash, even if the wanted hash was of incorrect length?

That being said, I think this is already an improvement and my suggestion is about a case that practically should not happen. (Unless the hash on an empty string is 0).

edolstra:
I think we have to be very careful about this not having unintended side effects. It seems safer to handle this at a higher level than in the `Hash` type (which is used all over the place).

edolstra:
An example of "higher level" would be `fetchurl`, which could set a dummy hash if none is specified, or `prim_derivationStrict` (but it would have to be careful not to turn normal derivations into fixed-output derivations with a dummy hash).

matthewbauer:
Right - a more explicit `try` / `catch` here would be better.

edolstra:
Will close this PR for the reason mentioned baove.

matthewbauer:
updated with changes here:

https://github.com/NixOS/nix/pull/3674

please take a look when you get a chance to.

PR #3546, opened 2020-04-29
builtins.readFile: do not truncate content

### LINKED ISSUE #3026: Nix repl - fails to read /proc files
Please see very small shell session for understanding: 
```
$nix repl                                                                                                                                                             Welcome to Nix version 2.1.3. Type :? for help.

nix-repl> builtins.readFile /etc/os-release
"NAME=NixOS\nID=nixos\nVERSION=\"18.09.2466.884b0a4417e (Jellyfish)\"\nVERSION_CODENAME=jellyfish\nVERSION_ID=\"18.09.2466.884b0a4417e\"\nPRETTY_NAME=\"NixOS 18.09.2466.884b0a4417e (Jellyfish)\"\nHOME_URL=\"https://nixos.org/\"\nSUPPORT_URL=\"https://nixos.org/nixos/support.html\"\nBUG_REPORT_URL=\"https://github.com/NixOS/nixpkgs/issues\"\n"

nix-repl> builtins.readFile /proc/cpuinfo
""
```

guibou:
This closes #3026 by allowing `builtins.readFile` to read a file with a
wrongly reported file size, for example, files in `/proc` may report a
file size of 0. Reading file in `/proc` is not a good enough motivation,
however I do think it just makes nix more robust by allowing more file
to be read.  Especially, I do considerer the previous behavior to be
dangerous because nix was previously reading truncated files. Examples
of file system which incorrectly report file size may be network file
system or dynamic file system (for performance reason, a dynamic file
system such as FUSE may generate the content of the file on demand).

```
nix-repl> builtins.readFile "/proc/version"
""
```

With this commit:

```
nix-repl> builtins.readFile "/proc/version"
"Linux version 5.6.7 (nixbld@localhost) (gcc version 9.3.0 (GCC)) #1-NixOS SMP Thu Apr 23 08:38:27 UTC 2020\n"
```

Here is a summary of the behavior changes:

- If the reported size is smaller, previous implementation
was silently returning a truncated file content. The new implementation
is returning the correct file content.

- If a file had a bigger reported file size, previous implementation was
failing with an exception, but the new implementation is returning the
correct file content. This change of behavior is coherent with this pull
request.

Open questions

- The behavior is unchanged for correctly reported file size, however
performances may vary because it uses the more complex sink interface.
Considering that sink is used a lot, I don't think this impacts the
performance a lot.
- `builtins.readFile` on an infinite file, such as `/dev/random` may
fill the memory.
- it does not support adding file to store, such as `${/proc/version}`.

### DIFF
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -419,7 +419,7 @@ void LocalStore::findRuntimeRoots(Roots & roots, bool censor)
 
                 try {
                     auto mapFile = fmt("/proc/%s/maps", ent->d_name);
-                    auto mapLines = tokenizeString<std::vector<string>>(readFile(mapFile, true), "\n");
+                    auto mapLines = tokenizeString<std::vector<string>>(readFile(mapFile), "\n");
                     for (const auto & line : mapLines) {
                         auto match = std::smatch{};
                         if (std::regex_match(line, match, mapRegex))
@@ -427,7 +427,7 @@ void LocalStore::findRuntimeRoots(Roots & roots, bool censor)
                     }
 
                     auto envFile = fmt("/proc/%s/environ", ent->d_name);
-                    auto envString = readFile(envFile, true);
+                    auto envString = readFile(envFile);
                     auto env_end = std::sregex_iterator{};
                     for (auto i = std::sregex_iterator{envString.begin(), envString.end(), storePathRegex}; i != env_end; ++i)
                         unchecked[i->str()].emplace(envFile);
--- a/src/libutil/serialise.hh
+++ b/src/libutil/serialise.hh
@@ -148,6 +148,9 @@ struct StringSink : Sink
 {
     ref<std::string> s;
     StringSink() : s(make_ref<std::string>()) { };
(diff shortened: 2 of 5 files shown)


edolstra:
Note that this makes the `drain` parameter to `readFile()` rather pointless, since both cases do the same thing now:
```
return drain ? drainFD(fd.get()) : readFile(fd.get());
```
where `readFile(fd) == return drainFD(fd, true);`.

There does appear to be a small but measurable speed penalty btw, e.g. running `nix-env -qa`:
```
elapsed time:       median =      4.7344  mean =      4.7326  stddev =      0.0114  min =      4.7089  max =      4.7584  [rejected, p=0.00000, Δ=0.01442±0.00663]
```
Don't know if it's because of the increased number of system calls or having to reallocate the std::string contents.

guibou:
@edolstra how do you run the performance benchmark? I'll try to see if the `std::string` size can be `reserve`d using the `st_size` field.

I'll remove the `drain` parameter to `readFile`, thank for seeing that.

guibou:
@edolstra `drain` argument removed, and now the string size is correctly reserved. I hope it fixs the performance regression.

PR #3549, opened 2020-04-29
Merge legacy `fetchGit`-builtin with the generic `fetchTree`-function

Ma27:
The original idea was to implement a git-fetcher in Nix's core that
supports content hashes[1]. In #3549[2] it has been suggested to
actually use `fetchTree` for this since it's a fairly generic wrapper
over the new fetcher-API[3] and already supports content-hashes.

This patch implements a new git-fetcher based on `fetchTree` by
incorporating the following changes:

* Removed the original `fetchGit`-implementation and replaced it with an
  alias on the `fetchTree` implementation.

* Ensured that the `git`-fetcher from `libfetchers` always computes a
  content-hash and returns an "empty" revision on dirty trees (the
  latter one is needed to retain backwards-compatibility).

* The hash-mismatch error in the fetcher-API exits with code 102 as it
  usually happens whenever a hash-mismatch is detected by Nix.

* Removed the `flakes`-feature-flag: I didn't see a reason why this API
  is so tightly coupled to the flakes-API and at least `fetchGit` should
  remain usable without any feature-flags.

* It's only possible to specify a `narHash` for a `git`-tree if either a
  `ref` or a `rev` is given[4].

* It's now possible to specify an URL without a protocol. If it's missing,
  `file://` is automatically added as it was the case in the original
  `fetchGit`-implementation.

[1] https://github.com/NixOS/nix/pull/3216
[2] https://github.com/NixOS/nix/pull/3549#issuecomment-625194383
[3] https://github.com/NixOS/nix/pull/3459
[4] https://github.com/NixOS/nix/pull/3216#issuecomment-553956703

----

cc @edolstra @globin @xbreak @FRidh


Ma27:
@domenkozar would you mind taking a look at this? :)

edolstra:
The reason for not adding a `narHash` flag to `fetchGit` is that we now have a more general builtin function `fetchTree`, which should already support `narHash`. So `fetchGit` / `fetchMercurial` are kind of legacy function that we keep for backward compatibility.

src/libexpr/primops/fetchGit.cc
```diff
@@ -34,12 +36,18 @@ static void prim_fetchGit(EvalState & state, const Pos & pos, Value * * args, Va
...
+                narHash = state.forceStringNoCtx(*attr.value, *attr.pos);
             else
                 throw EvalError("unsupported argument '%s' to 'fetchGit', at %s", attr.name, *attr.pos);
         }
 
         if (url.empty())
             throw EvalError(format("'url' argument required, at %1%") % pos);
+        if (narHash && !rev)
+            throw EvalError(
+                format("'narHash' can only be used for 'fetchGit' if 'rev' is set, at %1%") % pos
```
edolstra, line 49: I think this is too strict. If you use a tag ref, it's fine to specify a `narHash` but not a `rev`.

Ma27: Agreed. Just realized that I allowed both revs and refs in the original PR.

Ma27:
> The reason for not adding a narHash flag to fetchGit is that we now have a more general builtin function fetchTree, which should already support narHash. So fetchGit / fetchMercurial are kind of legacy function that we keep for backward compatibility.

I see. When I filed this patch, I wasn't' sure whether it's intended that `fetchGit` exists on its own. Should I modify this to use `fetchTree` entirely?

This would still require some modifications in the `git`-fetcher and the wrapper in `fetchTree` that passes arguments to the fetcher API.

edolstra:
Yes, please. All the other changes (like returning 102) look good to me.

Ma27:
Rebased onto latest master and merged `fetchTree` and `fetchGit` into a single C++-function using the fetcher-API. More details can be found in the new commit-message.

@globin since you were quite interested in this and I had to change a lot, you may want to take another look at this :)

Ma27:
ping @domenkozar @edolstra

domenkozar:
I'll leave it up to Eelco for final review.

edolstra:
Would you mind rebasing this on the flakes branch? I made some changes to the fetcher API there. (In particular `Input` is now basically just a set of attributes.)

Ma27:
Yes, already on it :)

Ma27:
@edolstra rebased onto `flakes` :)

src/libexpr/primops/fetchTree.cc
```diff
@@ -64,8 +97,15 @@ static void prim_fetchTree(EvalState & state, const Pos & pos, Value * * args, V
 
         for (auto & attr : *args[0]->attrs) {
             state.forceValue(*attr.value);
-            if (attr.value->type == tString)
-                attrs.emplace(attr.name, attr.value->string.s);
+            if (attr.value->type == tPath)
```
edolstra, line 100: => `if (attr.value->type == tPath || attr.value->type == tString)` since `coerceToString` handles strings as well.

src/libexpr/primops/fetchTree.cc
```diff
@@ -48,10 +51,40 @@ void emitTreeAttrs(
...
+
+void addURI(EvalState &state, fetchers::Attrs &attrs, Symbol name, std::string v)
+{
+    string n(name);
+    attrs.emplace(name, n == "url" ? fixURI(v, state) : v);
+}
+
+void validateFetcherAttrs(fetchers::Attrs attrs, const Pos &pos)
 {
-    settings.requireExperimentalFeature("flakes");
```
edolstra, line 53: We need to keep this in `fetchTree`.

src/libexpr/primops/fetchTree.cc
```diff
@@ -48,10 +51,40 @@ void emitTreeAttrs(
     v.attrs->sort();
 }
 
-static void prim_fetchTree(EvalState & state, const Pos & pos, Value * * args, Value & v)
+std::string fixURI(std::string uri, EvalState &state)
+{
+    state.checkURI(uri);
```
edolstra, line 56: We can remove `checkURI` since the check can easily be by-passed.

src/libexpr/primops/fetchTree.cc
```diff
@@ -36,6 +36,9 @@ void emitTreeAttrs(
         mkString(*state.allocAttr(v, state.symbols.create("shortRev")), rev->gitShortRev());
     }
 
+    if (input.getType() == "git")
+        mkBool(*state.allocAttr(v, state.symbols.create("submodules")), maybeGetBoolAttr(input.attrs, "submodules").value_or(false));
```
edolstra, line 40: This should only be done for `fetchGit`.

src/libexpr/primops/fetchTree.cc
```diff
@@ -75,12 +115,24 @@ static void prim_fetchTree(EvalState & state, const Pos & pos, Value * * args, V
...
+
+        validateFetcherAttrs(attrs, pos);
 
         input = fetchers::Input::fromAttrs(std::move(attrs));
-    } else
-        input = fetchers::Input::fromURL(state.coerceToString(pos, *args[0], context, false, false));
+    } else {
+        auto url = fixURI(state.coerceToString(pos, *args[0], context, false, false), state);
+
+        if (type && type.value() == "git") {
```
edolstra, line 127: => `type == "git"` should also work.

src/libfetchers/git.cc
```diff
@@ -269,6 +269,10 @@ struct GitInputScheme : InputScheme
                     "lastModified",
                     haveCommits ? std::stoull(runProgram("git", true, { "-C", actualUrl, "log", "-1", "--format=%ct", "HEAD" })) : 0);
 
+                // Backward compatibility: set 'rev' to
+                // 0000000000000000000000000000000000000000 for a dirty tree.
+                input.attrs.insert_or_assign("rev", "0000000000000000000000000000000000000000");
```
edolstra, line 274: This should only be done for `fetchGit`.

src/libexpr/primops/fetchTree.cc
```diff
@@ -48,10 +51,40 @@ void emitTreeAttrs(
...
-    settings.requireExperimentalFeature("flakes");
+    if (!attrs.count("type"))
+        throw Error("attribute 'type' is missing in call to 'fetchTree', at %s", pos);
+
+    if (getStrAttr(attrs, "type") == "git") {
+        auto narHash = maybeGetStrAttr(attrs, "narHash");
+        auto ref = maybeGetStrAttr(attrs, "ref");
+        auto rev = maybeGetStrAttr(attrs, "rev");
+
+        if (narHash && !ref && !rev)
```
edolstra, line 76: This validation should be done in `git.cc`, not `fetchTree.c`. Also, I think this test isn't really necessary, since people should be free to specify a narHash without a rev/ref. It could be fine in flake.nix, since the lock file will contain the rev...

src/libexpr/primops/fetchTree.cc
Ma27: Why though? At least `fetchGit` works (until now) without flakes.

src/libexpr/primops/fetchTree.cc
Ma27: Without that most of the tests fail like this:

```
    + nix eval --raw --expr 'builtins.fetchTarball https://github.com/NixOS/patchelf/archive/master.tar.gz' --impure --restrict-eval
    error: --- RestrictedPathError --- nix
    access to URI 'https://github.com/NixOS/patchelf/archive/master.tar.gz' is forbidden in restricted mode
```

src/libexpr/primops/fetchTree.cc
Ma27: Can you explain to me why this shouldn't happen for e.g. `fetchTree { type = "git"; }`?

src/libfetchers/git.cc
Ma27: Do you want me to move this to the primop implementation? This is already the case for `git`-sources only.

Ma27:
@edolstra rebased onto latest `flakes`-branch and fixed most of your comments (I added an answer/comment to the review-comments I didn't address (yet)).

Ericson2314:
@edolstra Is there a rough ETA for merging the flakes branch? This change would be desirable for us too: if flakes will be merged soon that's fine, but if not, then maybe we could back-port something like the earlier `master`-targetted version of this one?

edolstra:
@Ericson2314 I can merge it after the RFC is merged.

Ma27:
@edolstra may I ask why though? Those fetchers are even on non-flakes master atm.

Ma27:
@edolstra rebased conflicts against master. Is there anything else tbd to get this merged?

Ma27:
@edolstra the RFC is closed now - anything else we'd have to wait for until you can merge this? :)

src/libexpr/primops/fetchTree.cc
edolstra: Because `fetchTree` (not `fetchGit`) needs to be behind an experimental feature flag.

src/libfetchers/git.cc
edolstra: What I mean is that it shouldn't be done for `fetchTree`, even for Git inputs.

src/libexpr/primops/fetchTree.cc
Ma27: Oh, you're absolutely right, fixed. Sorry.

src/libfetchers/git.cc
Ma27: Fixed. However there's now no `rev` field if only a url for fetchGit is specified.

Ma27:
@edolstra resolved the merge conflict. Is there anything to discuss about the pending two comments you've left?

src/libexpr/primops/fetchTree.cc
edolstra: We're trying to avoid returning unnecessary information from `fetchTree` since that might influence the evaluation result. But it probably doesn't matter much here.

tests/fetchGit.sh
```diff
@@ -123,7 +129,7 @@ path2=$(nix eval --impure --raw --expr "(builtins.fetchGit file://$repo).outPath
 # Using local path with branch other than 'master' should work when clean or dirty
 path3=$(nix eval --impure --raw --expr "(builtins.fetchGit $repo).outPath")
 # (check dirty-tree handling was used)
-[[ $(nix eval --impure --raw --expr "(builtins.fetchGit $repo).rev") = 0000000000000000000000000000000000000000 ]]
+[[ $(nix eval --impure --expr "(builtins.fetchGit $repo).rev or null") = null ]]
```
edolstra, line 132: This breaks compatibility since there might be projects that expect that `rev` is always set. `fetchGit` (but not `fetchTree`) on dirty trees should continue to return `000...`.

Ma27: so, partly revert my change to ensure fetchGit returns an empty rev here?

edolstra: Yes please.

Ma27: Fixed. However I'm well-aware that my current fix is fairly ugly since I didn't manage to allocate a new attr to `v` in `prim_fetchGit`. If you have a better solution, please tell me :)

edolstra:
@Ma27 Merged, thanks for all your work on this!

Ma27:
You're welcome :) 

I'm surprised though that there's now a failingtest although I ran `make installcheck` several times locally. Tonight I have time to look into this.

Ma27:
For the macos issue I asked in `#nixos-dev` whether someone with the needed hardware can take a look.
`nix-build -A checks.x86_64-linux` repeatedly builds localle.

Issue #3554, opened 2020-05-01
Warn about using cache only after some time


aepsil0n:
**Is your feature request related to a problem? Please describe.**
`nix search` warns about using the cache for searches every time it is used without `-u`. Warnings usually should be interpreted as calls to action. However, when one uses `nix search` frequently, most of the time there is not much value in updating the cache.

**Describe the solution you'd like**
Only warn about using the cache and remind the user to update via `-u` after a certain period of time, long enough for relevant changes to have occurred. My suggestion would be **3 days** to get updated frequently enough on potential changes on the unstable branch without continuously being distracted by warnings. But please bikeshed, what time frame makes sense.

**Describe alternatives you've considered**
- Add a command-line flag to suppress the warning. This would also allow users to ignore the warning or build their own logic around when to use or not to use the cache.
- Go further and automatically update the cache after a certain time has passed since the last update. This would make the warning unnecessary altogether.

edolstra:
The cache used by `nix search` has been completely rewritten and is now completely transparent (i.e. it never gives outdated results). So I'll close this issue.

jpotier:
hey @edolstra could you link a MR / commit so that curious readers such as myself can jump and read on that change?

aepsil0n:
Oh, sorry, I was not aware of that. Thanks for pointing it out though.

edolstra:
b69323f8c93d5b067b6baaa0acbb93f800fcd9bd is the commit that makes `nix search` use the new evaluation cache.

PR #3555, opened 2020-05-01
Warn about using the cache only when its older

aepsil0n:
The age of the cache is determined by the date it was last modified from the filesystem.

---

This PR addresses #3554. See that issue for motivation.


edolstra:
The new evaluation cache in the flakes branch doesn't need this so I'll close this PR.

Issue #1216, opened 2017-02-03
Wait for build users instead of failing


expipiplus1:
It would be nice to have the option of waiting for a build user to become free instead of failing to build.

I'm talking about this error: `error: all build users are currently in use; consider creating additional users and adding them to the ‘nixbld’ group`

Is there ever a case where failing is the preferred option to waiting? Hopefully waiting can become the default behaviour.

waynr:
For anyone else who encounters this problem, the way I solved it was by bumping my `nixbld<N>` users from 10 to 100. My multi user nix build is happily chugging along now.

jonringer:
For NixOS users who may stumble upon this thread, the setting is `nix.nrBuildUsers = <int>;` in your /etc/nixos/configuration.nix

https://nixos.org/nixos/options.html#nrbuildusers

Gabriel439:
One thing that triggers this problem is enabling distributed builds when dealing with a large number of cores (32+)

The reason why is that,  by default, the setting for `nix.nrBuildUsers` is essentially `max 32 nix.maxJobs`, which is fine when all builds are local because `nix.maxJobs` then never exceeds `nix.nrBuildUsers`.  However, things break down if the number of build slots across all available remote builders plus the number of local build slots exceeds `nix.nrBuildUsers`, because then you run the risk of exhausting the available Nix users.

The fix is to increase the `nix.nrBuildUsers` option like @waynr suggested, but if you don't want to guess what the value should be then set it to be the sum of all available build slots across all machines that builds can be delegated to (including the local machine).

For example, if your NixOS configuration were:

```nix
{ nix.maxJobs = 32;

  nix.distributedBuilds = true;

  nix.buildMachines = [
    { hostName = "example.com";
      sshUser = "example";
      sshKey = "/root/.ssh/id_rsa";
      system = "x86_64-linux";
      maxJobs = 32;
      speedFactor = 1;
      supportedFeatures = [ "kvm" ];
      mandatoryFeatures = [ "perf" ];
    }
  ];
}
```

Then the correct setting for `nix.nrBuildUsers` would be `64` (32 for the local machine and 32 for the remote builder).

This could potentially be mitigated at the NixOS level by changing the default value for `nix.maxJobs` to factor in the `nix.buildMachines` option when calculating how many builders to create.  However, that wouldn't fix people using Nix the package manager on a non-NixOS system.

balsoft:
Is there any work on this?

balsoft:
[...]

Issue #2069, opened 2018-04-11
"waiting for locks or build slots" is not a very helpful message


LisannaAtHome:
I think the message that nix prints when something is blocking the build could be a bit more descriptive. "waiting for locks or build slots" does not indicate whether there is a resource problem that the user needs to go fix (let alone what the resource problem is), or whether the same build is already underway by another process. Even turning the logging up to vomit doens't yield anything particularly insightful.

7c6f434c:
Possibly related: it _is_ possible to run multiple Nix instances in parallel as root and get an error (instead of blocking to wait for build slots) after running out of build users.

shlevy:
@ledettwy Any interest in providing a patch, or suggested output at least?

domenkozar:
See #1216

balsoft:
See #3577

balsoft:
This should be resolved in 5ed5d7acbd00298df7b4e3a638d731143084da2e

PR #3564, opened 2020-05-05
Wait for build users when none are available

balsoft:
Solves #1216  

Descriptions from there:
>It would be nice to have the option of waiting for a build user to become free instead of failing to build.
>I'm talking about this error: error: all build users are currently in use; consider creating additional users and adding them to the ‘nixbld’ group
>Is there ever a case where failing is the preferred option to waiting? Hopefully waiting can become the default behaviour.

Additional implementation details: when build users are available, the behavior is exactly the same as before (choose the first available user, take a lock on it, use it). However, when no build users are available at the moment, repeat the search every two seconds. Whenever a build user becomes available, choose it and continue as usual. Previous behavior was to error out.

Tested on:
- [x] NixOS
- [ ] Non-NixOS Linux
- [ ] Darwin
- [ ] other platforms.

_Note:_
I have tried implementing a more "proper" solution with `libfswatch` (to only check the userpool when it changes), but for some reason the callback didn't trigger when lock files were released. If you believe that my work on implementing this with `libfswatch` is of interest, I can publish it.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -507,9 +507,6 @@ private:
     Path fnUserLock;
     AutoCloseFD fdUserLock;
 
-    bool findFreeUser();
-
-
     string user;
     uid_t uid;
     gid_t gid;
@@ -525,20 +522,19 @@ public:
     uid_t getGID() { assert(gid); return gid; }
     std::vector<gid_t> getSupplementaryGIDs() { return supplementaryGIDs; }
 
+    bool findFreeUser();
+
     bool enabled() { return uid != 0; }
 
 };
+
 
 UserLock::UserLock()
 {
     assert(settings.buildUsersGroup != "");
     createDirs(settings.nixStateDir + "/userpool");
-
-    if (findFreeUser()) return;
-
-    printError("waiting for build users");
-
-    do std::this_thread::sleep_for(std::chrono::seconds(2)); while (! findFreeUser());
+    /* Mark that user is not enabled by default */
+    uid = 0;
 }
 
 bool UserLock::findFreeUser() {
@@ -1398,6 +1394,30 @@ void DerivationGoal::tryToBuild()
 {
     trace("trying to build");
 
+    /* If `build-users-group' is not empty, then we have to build as
+       one of the members of that group. */
+    if (settings.buildUsersGroup != "" && getuid() == 0) {
+#if defined(__linux__) || defined(__APPLE__)
+        if (!buildUser) buildUser = std::make_unique<UserLock>();
+
+        if (!buildUser->enabled()) {
+            if (!buildUser->findFreeUser()) {
+                debug("waiting for build users");
+                worker.waitForAWhile(shared_from_this());
+                return;
+            }
+
+            /* Make sure that no other processes are executing under this
+               uid. */
+            buildUser->kill();
+        }
+#else
+        /* Don't know how to block the creation of setuid/setgid
+           binaries on this platform. */
+        throw Error("build users are not supported on this platform for security reasons");
+#endif
+    }
+
     /* Obtain locks on all output paths.  The locks are automatically
        released when we exit this function or Nix crashes.  If we
        can't acquire the lock, then continue; hopefully some other
@@ -1948,22 +1968,6 @@ void DerivationGoal::startBuilder()
         #else
             throw Error("building using a diverted store is not supported on this platform");
         #endif
-    }
-
-    /* If `build-users-group' is not empty, then we have to build as
-       one of the members of that group. */
-    if (settings.buildUsersGroup != "" && getuid() == 0) {
-#if defined(__linux__) || defined(__APPLE__)
-        buildUser = std::make_unique<UserLock>();
-
-        /* Make sure that no other processes are executing under this
-           uid. */
-        buildUser->kill();
-#else
-        /* Don't know how to block the creation of setuid/setgid
-           binaries on this platform. */
-        throw Error("build users are not supported on this platform for security reasons");
-#endif
     }
 
     /* Create a temporary directory where the build will take
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -507,6 +507,7 @@ private:
     Path fnUserLock;
     AutoCloseFD fdUserLock;
 
+    bool isEnabled;
     string user;
     uid_t uid;
     gid_t gid;
@@ -524,7 +525,7 @@ public:
 
     bool findFreeUser();
 
-    bool enabled() { return uid != 0; }
+    bool enabled() { return isEnabled; }
 
 };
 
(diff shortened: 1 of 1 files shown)


domenkozar:
Thank you a thousand times!

LnL7:
I think this also resolves the fact that we make 32 build users in the installer, this was mostly the reason for it.

balsoft:
> I think this also resolves the fact that we make 32 build users in the installer, this was mostly the reason for it.

This is my next nixpkgs PR, yes ;)

src/libstore/build.cc
```diff
@@ -526,10 +529,19 @@ class UserLock
...
 
-
 UserLock::UserLock()
 {
     assert(settings.buildUsersGroup != "");
+    createDirs(settings.nixStateDir + "/userpool");
+
+    if (findFreeUser()) return;
+
+    printError("waiting for build users");
```
cole-h, line 539: Is this really an error? Maybe `warn` would be a better fit? (I found this after a quick grep -- there might be a more suitable function.)

Also (maybe it's just me), but I think it would be nice to see the waiting message every time we check, e.g. put it inside the do-while loop. It would at least give feedback that "Yes, we're still waiting. Sorry this is taking so long."

balsoft: I'm exactly replicating the behavior of `waiting for locks or build slots` message. It's a `printError` too and it only shows once.

cole-h: Ah, OK, that makes sense. (I don't typically run into the "waiting for locks or build slots" message, so it didn't even occur to me.)

Nevermind, then :)

balsoft: I think this merits further discussion. Maybe both of them should be refactored (see https://github.com/NixOS/nix/issues/2069), but I don't think this discussion belongs in this PR :)

edolstra:
This is not a good solution because it puts the entire build loop to sleep (since it's single-threaded). So no other builds/substitutions will make any progress until another process releases a build user. In fact, it seems that this will deadlock if this process itself has acquired all build users (since then it can't make progress to release any build users).

For comparison, if we can't acquire a lock on an output path, the `DerivationGoal` is put to sleep so that other goals can make progress:
```
    if (!outputLocks.lockPaths(lockFiles, "", false)) {
        worker.waitForAWhile(shared_from_this());
        return;
    }
```

balsoft:
@edolstra Thanks for the suggestion, I'll implement the same logic.

balsoft:
Please, re-evaluate if possible.

I have changed the behaviour to match that of output locks. However, now https://github.com/NixOS/nix/issues/2069 is even more of an issue.

balsoft:
My testing methodology, for anyone interested:

1. Run nix-daemon from this version of nix
2. Create a file
_test-nix.nix_
```
with import <nixpkgs> {};
let newPackage = n: hello.overrideAttrs (_: { name = "hello-${toString builtins.currentTime}-${toString n}"; });
in buildEnv { name = "env"; ignoreCollisions = true; paths = builtins.genList newPackage 100; }
```
3.  `nix build -f test-nix.nix --max-jobs $BUILDUSERS+3`
4.  `nix build -f test-nix.nix` as many times as needed to run out of build users.

If the last two commands succeed and use up all available build users while not crashing and not using users two times over, my patch is correct.

Please test and report.

balsoft:
@domenkozar I had a temptation to do so here, but I think it merits a separate PR. I'll create one now.

balsoft:
I'll come back to https://github.com/NixOS/nix/issues/2069 when I have time.

balsoft:
@domenkozar fixed in #3557 (will rebase if/when this PR is merged)

src/libstore/build.cc
```diff
@@ -530,6 +532,12 @@ class UserLock
 UserLock::UserLock()
 {
     assert(settings.buildUsersGroup != "");
+    createDirs(settings.nixStateDir + "/userpool");
+    /* Mark that user is not enabled by default */
+    uid = 0;
```
edolstra, line 537: `0` seems a risky choice as a sentinel value for UIDs given that 0 == root. So if there's ever any confusion, we might end up doing an operation as/to root. Might be better to use `-1` or `std::optional<uid_t>`.

src/libstore/build.cc
```diff
@@ -1391,6 +1394,30 @@ void DerivationGoal::tryToBuild()
 {
     trace("trying to build");
 
+    /* If `build-users-group' is not empty, then we have to build as
+       one of the members of that group. */
```
edolstra, line 1398: If I understand this correctly, it will try to acquire a build user even when we're doing a remote build, which is unnecessary. Maybe it's better to add a new state to the state machine, e.g. `startBuild`, with a transition from `tryToBuild` to `startBuild` if we're doing a local build.

src/libstore/build.cc
edolstra: Alternatively, to ensure that `UserLock` always represents a valid user, have a builder function that returns `std::unique_ptr<UserLock>` so it can represent failure to acquire a user.

src/libstore/build.cc
balsoft: Oh yes, this sounds very much like something I didn't even think about.

I'll try to wrap my head around this and implement it next weekend.

Thank you again!

src/libstore/build.cc
balsoft: I've thought about adding a function that can represent a failure, and it does appear to be a better solution from some perspectives, but I didn't want to make the diff too big (since this is my first PR). Now that it's inevitably going to be massive, I'll wrap the result in optional and make it represent failure.

src/libstore/build.cc
balsoft: Added a state and (hopefully) fixed this.

balsoft:
@edolstra I've fixed both of issues you've found, please look at it again when you have time.

The CI seems to be slightly broken (it can't find a builder for macOS).

I've tested this change on my NixOS desktop as explained above.

edolstra:
Thanks!

osa1:
Sorry for reviving and old thread and n00b question, but, what does it mean if I'm still getting "all build users are currently in use" errors today? Does that mean my nix setup is too old and does not have this patch? Or could it be a configuration issue?

balsoft:
This patch never became part of the 2.3 (stable) branch, so it's only available on master/2.4pre right now. @edolstra is there any chance to backport this to 2.3?

osa1:
Thanks for the quick response. Do we have a public URL for 2.4pre that I can use? I'm using https://github.com/cachix/install-nix-action which allows specifying install URL (default is https://nixos.org/nix/install).

balsoft:
Please see https://github.com/cachix/install-nix-action#usage-with-flakes (except you don't need all the `experimental-features` bits, just copy `install_url` and `install_options`

PR #3565, opened 2020-05-05
Use nixpkgs#bashInteractive for dev-shell

mkenigs:
Instead of using system bash for dev-shell, uses nixpkgs#bashInteractive as discussed in https://github.com/NixOS/nix/pull/3107#issuecomment-535981016. If dev-shell is used for a flake with a toplevel nixpkgs input, it uses that nixpkgs per https://github.com/NixOS/nix/pull/3107#issuecomment-536957104.

Reused a lot of code from InstallableFlake - seems like that should maybe get factored out?

@matthewbauer

### DIFF
--- a/src/nix/dev-shell.cc
+++ b/src/nix/dev-shell.cc
@@ -6,9 +6,6 @@
 #include "derivations.hh"
 #include "affinity.hh"
 #include "progress-bar.hh"
-#include "attr-path.hh"
-#include "eval-cache.hh"
-#include "flake/flake.hh"
 
 #include <regex>
 
--- a/src/nix/dev-shell.cc
+++ b/src/nix/dev-shell.cc
@@ -286,9 +286,8 @@ struct CmdDevShell : Common, MixEnvironment
 
     std::string getBashPath(ref<Store> store)
     {
-        auto flakeRef = FlakeRef::fromAttrs({{"type","indirect"}, {"id", "nixpkgs"}});
         auto state = getEvalState();
-        auto lockedFlake = std::make_shared<flake::LockedFlake>(lockFlake(*state, flakeRef, lockFlags));
+        auto lockedFlake = std::make_shared<flake::LockedFlake>(lockFlake(*state, installable->nixpkgsFlakeRef(), lockFlags));
         auto cache = openEvalCache(*state, lockedFlake, true);
         auto root = cache->getRoot();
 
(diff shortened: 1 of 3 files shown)


src/nix/dev-shell.cc
```diff
@@ -281,6 +284,40 @@ struct CmdDevShell : Common, MixEnvironment
         };
     }
 
+    std::string getBashPath(ref<Store> store)
```
edolstra, line 287: Maybe this function can be generalized to fetch any package. For example, it could be used by the Git fetcher to fetch `nixpkgs#git` it `git` is not installed. Although that's a bit tricky right now because libfetcher doesn't have access to libexpr.

src/nix/installables.cc
```diff
@@ -409,6 +409,18 @@ InstallableFlake::getCursor(EvalState & state, bool useEvalCache)
     return res;
 }
 
+FlakeRef InstallableFlake::nixpkgsFlakeRef() const
+{
+    auto lockedFlake = lockFlake(*(cmd.getEvalState()), flakeRef, cmd.lockFlags);
```
edolstra, line 414: Calling `lockFlake` again is potentially expensive (and can print some warnings/messages again). So it would be better if we store `lockedFlake` somewhere in `InstallableFlake` so we only lock it once.

src/nix/dev-shell.cc
mkenigs: Could I start by taking out the SourceExprCommand from InstallableValue? Looks like all it's used for is state, getAutoArgs, and lockFlags, and it might be simpler for an Installable to be separate from a command. Then Installables could be used by InstallableCommands/fetchGit/dev-shell.

src/nix/installables.cc
mkenigs: Store it in _lockedFlake now

mkenigs:
Updated to just create a bashInteractive InstallableFlake. InstallableFlake will only call lockFlake once, but this will ultimately still call lockFlake twice, because the bashInteractive installable will call it a second time.

Seems like preventing that would require being able to have a LockedFlake return a LockedFlake for one of its inputs, but not sure if that's possible right now. And then everything used by toStorePath might need to get moved to LockedFlake or somewhere in libexpr.

edolstra:
Thanks, merged! I changed `nixpkgsFlakeRef()` to use nixpkgs from `lockedFlake->lockFile`, since otherwise it will not necessarily use the same nixpkgs as what the top-level flake use.

PR #3568, opened 2020-05-06
libstore/build.cc: more explicit error about form of output

kolloch:
Be more explicit about why we expect a regular file as output
when outputHashMode=flat for a fixed output derivation.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -3679,7 +3679,7 @@ void DerivationGoal::registerOutputs()
                 if (!S_ISREG(st.st_mode) || (st.st_mode & S_IXUSR) != 0)
                     throw BuildError(
                         format("output path '%1%' should be a non-executable regular file "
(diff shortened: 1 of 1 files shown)


src/libstore/build.cc
```diff
@@ -3678,7 +3678,8 @@ void DerivationGoal::registerOutputs()
                 /* The output path should be a regular file without execute permission. */
                 if (!S_ISREG(st.st_mode) || (st.st_mode & S_IXUSR) != 0)
                     throw BuildError(
-                        format("output path '%1%' should be a non-executable regular file") % path);
+                        format("output path '%1%' should be a non-executable regular file "
+                               "since recursive hashing is not enabled (outputHashMode=flat).") % path);
```
domenkozar, line 3682:
```suggestion
                               "since recursive hashing is not enabled (outputHashMode=flat)") % path);
```

PR #3570, opened 2020-05-06
nix/search: no error for empty search results if json is enabled

Mic92:
- result list will be always empty if --json is passed
- for scripts an empty search result is not really an error,
  we rather want to distinguish between evaluation errors and empty results

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -265,7 +265,7 @@ struct CmdSearch : SourceExprCommand, MixJSON
                 throw SysError("cannot rename '%s' to '%s'", tmpFile, jsonCacheFileName);
         }
 
-        if (results.size() == 0)
+        if (!json && results.size() == 0)
             throw Error("no results for the given search term(s)!");
 
         RunPager pager;


Mic92:
Snippet for quick testing:

```nix
{}: {
  nix.package = pkgs.nix.overrideAttrs (old: {
    patches = [
      (pkgs.fetchpatch {
        url = "https://github.com/Mic92/nix-1/commit/e60dc42caa659985aa689958c656f26dc51eae56.patch";
        sha256 = "19cg2q34sir0q83ni6jqvxlglqsz1xld57gizqzvs8g5m5ckb806";
      })
    ];
  });
}
```

Before:
```
[nix-shell:~/git/nixpkgs]$ nix search --json > /dev/null
warning: using cached results; pass '-u' to update the cache
error: no results for the given search term(s)!
```

After
```
$ nix search --json >/dev/null
warning: using cached results; pass '-u' to update the cache
```

edolstra:
Note that it's not really worth fixing bugs in the `nix search` cache because it has been completely rewritten on the flakes branch.

domenkozar:
@edolstra sorry for that, do you want to revert?

Mic92:
No need to revert. The bug is present in both branches. I fixed the flakes branch in https://github.com/NixOS/nix/pull/3572

edolstra:
Ah sorry, I was confused with the JSON cache in the old `nix search`. I forgot it also has a `--json` flag for output :-)

PR #3571, opened 2020-05-06
Add unit tests

gilligan:
**Overview**
This PR adds unit tests for libutil using https://github.com/google/googletest - The idea was to start somewhere and see how well writing tests works out and then open a PR to get some feedback and also some help.

**Motivation**
For now Nix has only ever had functional tests. They are super useful and can catch a lot of problems but unit tests can provide added benefits like documentation and making it easier to refactor or port code.

**Notes**
I picked some simple module (libutil) for testing first. There are going to be other parts of the code base which are more tricky to test and in various cases it just won't make any sense to try to do so. That being said it would be nice to gradually improve the amount of tests and to encourage contributors to provide tests when they change code/add features.
 
**TODOs**

- [x] Properly embed this into the build rules (`tests/unit-tests/local.mk` is a total hack)
- [x] Execute during nix build and/or during CI
- [x] Address the "XXX" comments in the tests


tests/unit-tests/test-util.cc
```diff
@@ -0,0 +1,594 @@
...
+        ASSERT_ANY_THROW(canonPath(".."));
+        ASSERT_ANY_THROW(canonPath("../"));
+        ASSERT_DEATH({ canonPath(""); }, "path != \"\"");
+    }
+
+    /* ----------------------------------------------------------------------------
+     * dirOf
+     * --------------------------------------------------------------------------*/
+
+    // XXX: according to the doc of `dirOf`, dirOf("/") and dirOf("/foo")
```
gilligan, line 89: Maybe just change the docs to describe the behavior correctly?

tests/unit-tests/test-util.cc
```diff
@@ -0,0 +1,594 @@
...
+        auto p1 = baseNameOf("dir/foo");
+        ASSERT_STREQ(std::string(p1).c_str(), "foo");
+    }
+
+    TEST(baseNameOf, pathWithTrailingSlashRoot) {
+        auto p1 = baseNameOf("/");
+        ASSERT_STREQ(std::string(p1).c_str(), "");
+    }
+
+    // XXX: according to the doc of `baseNameOf`, baseNameOf("/dir/") should return
```
gilligan, line 132: The behavior is consistent with what `basename` does in the shell so i guess we can just change the description of the function?

tests/unit-tests/test-util.cc
```diff
@@ -0,0 +1,594 @@
...
+    }
+
+    TEST(isInDir, notInDir) {
+        auto p1 = isInDir("/zes/foo/bar", "/foo");
+        ASSERT_EQ(p1, false);
+    }
+
+    // XXX: hm, bug or feature? :) Looking at the implementation
+    // this might be problematic.
+    TEST(isInDir, emptyDir) {
```
gilligan, line 155: What about this one?

tests/unit-tests/test-util.cc
```diff
@@ -0,0 +1,594 @@
...
+        ASSERT_EQ(isDirOrInDir("/foo", "/bar"), false);
+    }
+
+    TEST(isDirOrInDir, relativePaths) {
+        ASSERT_EQ(isDirOrInDir("/foo/..", "/foo"), true);
+    }
+
+    // XXX: while it is possible to use "." or ".." in the
+    // first argument this doesn't seem to work in the second.
+    TEST(isDirOrInDir, DISABLED_shouldWork) {
```
gilligan, line 183: What should be done about this?

tests/unit-tests/test-util.cc
```diff
@@ -0,0 +1,594 @@
...
+    TEST(toLower, emptyString) {
+        ASSERT_STREQ(toLower("").c_str(), "");
+    }
+
+    TEST(toLower, nonLetters) {
+        auto s = "!@(*$#)(@#=\\234_";
+        ASSERT_STREQ(toLower(s).c_str(), s);
+    }
+
+    // XXX: std::tolower() doesn't cover this. This probably doesn't matter
```
gilligan, line 308: Like i say in the comment this probably is not relevant because it will only ever used in the context of storepaths which cannot cotnain funky  characters like that anyway?

tests/unit-tests/test-util.cc
```diff
@@ -0,0 +1,594 @@
...
+     * statusOk
+     * --------------------------------------------------------------------------*/
+
+    TEST(statusOk, zeroIsOk) {
+        ASSERT_EQ(statusOk(0), true);
+        ASSERT_EQ(statusOk(1), false);
+    }
+
+    /* ----------------------------------------------------------------------------
+     * replaceInSet : XXX: this function isn't called anywhere!
```
gilligan, line 368: unless i am mistaken replaceInSet is not called anywhere so it doesn't only not need tests it could in fact be removed?

edolstra: Yes, looks like it.

tests/unit-tests/test-util.cc
```diff
@@ -0,0 +1,594 @@
...
+namespace nix {
+
+    /* ----------------------------------------------------------------------------
+     * absPath
+     * --------------------------------------------------------------------------*/
+
+    TEST(absPath, doesntChangeRoot) {
+        auto p = absPath("/");
+
+        ASSERT_STREQ(p.c_str(), "/");
```
edolstra, line 15: Maybe there is some C++-based function that knows how to use `==` to test equality? Then we don't have to use `c_str()` everywhere.

gilligan: That is slightly annoying but on the other hand if we were to use some variation of ASSERT_TRUE(a == b) we would only get output like true being false instead of both strings and a diff.

Maybe there is some way to make this work with ASSERT_STREQ though? I am not sure.

Kha: The googletest primer says that `ASSERT_EQ` does work on `std::string`s (and in fact anything that defines `==`, and `<<` for output), so `ASSERT_EQ(p, "/")` should just work.

gilligan: Just goes to show that i don't know what i'm doing when dealing with C++ :sweat_smile: Both of you were right and I just replaced all uses with `ASSERT_EQ` and dropped the `c_str()` everywhere. 

Thanks!

gilligan:
@Kha thanks, fixed!

edolstra:
@gilligan I've integrated the tests into `make check`. I've also moved them to `src/libutil/tests`.

gilligan:
@edolstra From my point of view this would now be ready to be merged.

There are 2 remaining `XXX` in the test and i'm happy to address them in whatever way you prefer. I would however also be Okay with merging this first and addressing those separately.

Anything else you would like to see done to get this over the finishing line?

bhipple:
This is phenomenal!  Thanks guys!

PR #3572, opened 2020-05-06
nix/search: no error for empty search results if json is enabled

Mic92:
- result list will be always empty if --json is passed
- for scripts an empty search result is not really an error,
  we rather want to distinguish between evaluation errors and empty results

(cherry picked from commit e60dc42caa659985aa689958c656f26dc51eae56)


Mic92:
follow up of https://github.com/NixOS/nix/pull/3570

Mic92:
Fixed in https://github.com/NixOS/nix/commit/14a3a62bfca6c572b9a415cfa80cdbd7ad4326b3

PR #3573, opened 2020-05-06
Flake support

edolstra:
This PR adds experimental support for flakes. The flake file format is defined by [RFC 49](https://github.com/NixOS/rfcs/pull/49). To use it, you have to add
```
```
to `nix.conf`.

Other changes:

* The `nix` command is flake-based now. For instance, `nix build nixpkgs#hello` builds the `hello` package from the `nixpkgs` flake. Non-flake-based operation is still supported, e.g. `nix build -f /path/to/expr.nix`. However, `nix` no longer uses `NIX_PATH` so it's no longer affected by channels.

* New subcommand `nix flake` for working with flakes.

* New subcommand `nix profile` for imperative package management (intended to replace `nix-env`). (This is not completely finished. Note that once a profile has been modified using `nix profile`, you can't use `nix-env` on it anymore.)

* New subcommand `nix run` (previously named `nix app`) to run an application defined by an `app` or `defaultApp` output of a flake.

* The `nix` command has an evaluation cache to speed up querying flake attributes. For instance, `nix search` uses this instead of the old JSON cache.


domenkozar:
Nice!

> New subcommand nix profile for imperative package management (intended to replace nix-env). (This is not completely finished. Note that once a profile has been modified using nix profile, you can't use nix-env on it anymore.)

I really really wish we'd abandon imperative package management. I know many would disagree that it fills a gap of ease of use, but I think that can be filled with:

- `nix dev-shell` providing ad-hoc environments
- `nix flake --init pkg1, pkg2` or some command to start a flake scaffold

And in general conventions to use a flake to manage a profile and tooling around to make it really easy to start with one. Then it's just a matter of something like issuing a command to edit a profile flake.

One of the major pain points of previous interfaces was that people start using and get used to imperative package management and then there's no way to convert back and forth.

I hope it's not too late to reconsider.

src/nix/search.cc
```diff
@@ -63,214 +48,140 @@ struct CmdSearch : SourceExprCommand, MixJSON
...
-                    throw Error("error writing to %s", tmpFile);
-                throw;
-            }
-
-            if (writeCache && rename(tmpFile.c_str(), jsonCacheFileName.c_str()) == -1)
-                throw SysError("cannot rename '%s' to '%s'", tmpFile, jsonCacheFileName);
-        }
-
-        if (!json && results.size() == 0)
+        if (!results)
```
Mic92, line 183: Please don't re-introduce the bug fixed in master: https://github.com/NixOS/nix/pull/3570

```suggestion
        if (!json && !results)
```

edolstra:
Imperative package management serves a different use case from declarative package management: it doesn't require packages to be updated in lockstep. The ability *not* to upgrade everything at the same time is in fact one of the main selling points of Nix, IMHO.

It would be great to have a command line interface for modifying declarative specifications (i.e. flakes) though.

Gaelan:
I'd argue that non-lockstep updates can and should be handled declaratively, either by keeping old versions around in nixpkgs or by explicitly depending on multiple versions of nixpkgs. Having a bunch of imperatively managed dependencies with different versions seems like exactly the sort of irreproducible dependency hell Nix was designed to avoid.

edolstra:
Quite the opposite: that was *exactly* what Nix was designed for (i.e. the ability to update packages independently without running into dependency hell). Declarative specifications of your environment came later.

Keeping old versions around in Nixpkgs is infeasible and there is no guarantee that they'll evaluate/build in the same way. (For instance, if I have an old version of LibreOffice in my profile, I don't want to be forced to rebuild it and possibly have it break just because I'm installing a completely unrelated package.)

Explicitly depending on multiple older versions of Nixpkgs is fine and that's what flakes allow. However, rebuilding such a profile is a lot more expensive: Nix needs to fetch and evaluate the old Nixpkgs versions just to determine that nothing has changed. By contrast, a profile just records the store paths of installed packages so it doesn't have to re-evaluate anything.

bjornfor:
@edolstra : Won't the evaluation cache help a lot here?

Mic92:
Let's move the discussion about imperative package management into this issue and focus on code review in this [issue](https://github.com/NixOS/nix/issues/3579)

src/libexpr/flake/flake.cc
```diff
@@ -0,0 +1,660 @@
...
+        v);
+}
+
+static RegisterPrimOp r2("getFlake", 1, prim_getFlake);
+
+}
+
+Fingerprint LockedFlake::getFingerprint() const
+{
+    // FIXME: as an optimization, if the flake contains a lock file
```
Mic92, line 647: Is this `FIXME` supposed to be fixed before merging?

src/libexpr/flake/flakeref.cc
```diff
@@ -0,0 +1,196 @@
...
+
+        auto flakeRoot = path;
+        std::string subdir;
+
+        while (flakeRoot != "/") {
+            if (pathExists(flakeRoot + "/.git")) {
+                auto base = std::string("git+file://") + flakeRoot;
+
+                auto parsedURL = ParsedURL{
+                    .url = base, // FIXME
```
Mic92, line 131: FIXME for what reason?

src/libexpr/flake/flakeref.hh
```diff
@@ -0,0 +1,55 @@
...
+
+    bool operator==(const FlakeRef & other) const;
+
+    FlakeRef(const std::shared_ptr<const fetchers::Input> & input, const Path & subdir)
+        : input(input), subdir(subdir)
+    {
+        assert(input);
+    }
+
+    // FIXME: change to operator <<.
```
Mic92, line 29: FIXME.

src/libexpr/flake/lockfile.cc
```diff
@@ -0,0 +1,256 @@
...
+        auto lockedNode = std::dynamic_pointer_cast<const LockedNode>(i);
+        if (lockedNode && !lockedNode->lockedRef.input->isImmutable()) return false;
+    }
+
+    return true;
+}
+
+bool LockFile::operator ==(const LockFile & other) const
+{
+    // FIXME: slow
```
Mic92, line 187: FIXME?

src/libexpr/symbol-table.hh
```diff
@@ -28,6 +28,12 @@ public:
         return s == s2.s;
     }
 
+    // FIXME: remove
```
Mic92, line 31: FIXME

src/libexpr/symbol-table.hh
```diff
@@ -68,9 +74,10 @@ private:
     Symbols symbols;
 
 public:
-    Symbol create(const string & s)
+    Symbol create(std::string_view s)
     {
-        std::pair<Symbols::iterator, bool> res = symbols.insert(s);
+        // FIXME: avoid allocation if 's' already exists in the symbol table.
```
Mic92, line 79: FIXME.

src/libfetchers/git.cc
```diff
@@ -79,6 +79,63 @@ struct GitInput : Input
...
+        return {};
+    }
+
+    void markChangedFile(std::string_view file, std::optional<std::string> commitMsg) const override
+    {
+        auto sourcePath = getSourcePath();
+        assert(sourcePath);
+
+        runProgram("git", true,
+            { "-C", *sourcePath, "add", "--force", "--intent-to-add", std::string(file) });
```
Mic92, line 132: To avoid ambiguity:

```suggestion
            { "-C", *sourcePath, "add", "--force", "--intent-to-add", "--", std::string(file) });
```

misc/bash/completion.sh
```diff
@@ -0,0 +1,19 @@
+function _complete_nix {
```
Mic92, line 1: Would it make sense to do the same for zsh? https://github.com/spwhitt/nix-zsh-completions

src/libexpr/eval-cache.cc
```diff
@@ -0,0 +1,503 @@
...
+        if (!queryAttribute.next()) return {};
+
+        auto rowId = (AttrType) queryAttribute.getInt(0);
+        auto type = (AttrType) queryAttribute.getInt(1);
+
+        switch (type) {
+            case AttrType::Placeholder:
+                return {{rowId, placeholder_t()}};
+            case AttrType::FullAttrs: {
+                // FIXME: expensive, should separate this out.
```
Mic92, line 191: FIXME?

src/libexpr/eval-cache.cc
```diff
@@ -0,0 +1,503 @@
...
+        throw;
+    }
+
+    if (root->db && (!cachedValue || std::get_if<placeholder_t>(&cachedValue->second))) {
+        if (v.type == tString)
+            cachedValue = {root->db->setString(getKey(), v.string.s), v.string.s};
+        else if (v.type == tBool)
+            cachedValue = {root->db->setBool(getKey(), v.boolean), v.boolean};
+        else if (v.type == tAttrs)
+            ; // FIXME: do something?
```
Mic92, line 326: FIXME?

src/libexpr/eval-cache.cc
```diff
@@ -0,0 +1,503 @@
...
+                        throw EvalError("cached failure of attribute '%s'", getAttrPathStr(name));
+                    else
+                        return std::make_shared<AttrCursor>(root,
+                            std::make_pair(shared_from_this(), name), nullptr, std::move(attr));
+                }
+                // Incomplete attrset, so need to fall thru and
+                // evaluate to see whether 'name' exists
+            } else
+                return nullptr;
+                //throw TypeError("'%s' is not an attribute set", getAttrPathStr());
```
Mic92, line 361: Remove?

src/libexpr/eval-cache.cc
```diff
@@ -0,0 +1,503 @@
...
+        }
+    }
+
+    auto & v = forceValue();
+
+    if (v.type != tAttrs)
+        return nullptr;
+        //throw TypeError("'%s' is not an attribute set", getAttrPathStr());
```
Mic92, line 369: remove?

src/libfetchers/indirect.cc
```diff
@@ -0,0 +1,140 @@
...
+            if (!std::regex_match(path[1], refRegex))
+                throw BadURL("in flake URL '%s', '%s' is not a branch/tag name", url.url, path[1]);
+            input->ref = path[1];
+            if (!std::regex_match(path[2], revRegex))
+                throw BadURL("in flake URL '%s', '%s' is not a commit hash", url.url, path[2]);
+            input->rev = Hash(path[2], htSHA1);
+        } else
+            throw BadURL("GitHub URL '%s' is invalid", url.url);
+
+        // FIXME: forbid query params?
```
Mic92, line 112: FIXME?

src/libfetchers/mercurial.cc
```diff
@@ -60,6 +60,41 @@ struct MercurialInput : Input
...
+    }
+
+    void markChangedFile(std::string_view file, std::optional<std::string> commitMsg) const override
+    {
+        auto sourcePath = getSourcePath();
+        assert(sourcePath);
+
+        // FIXME: shut up if file is already tracked.
+        runProgram("hg", true,
+            { "add", *sourcePath + "/" + std::string(file) });
```
Mic92, line 91: I am unsure here, test if this actually work:

```suggestion
            { "add", "--", *sourcePath + "/" + std::string(file) });
```

src/libfetchers/mercurial.cc
```diff
@@ -60,6 +60,41 @@ struct MercurialInput : Input
...
+
+        if (commitMsg)
+            runProgram("hg", true,
+                { "commit", *sourcePath + "/" + std::string(file), "-m", *commitMsg });
```
Mic92, line 95: Same as above, check if `hg` accepts `--`.

```suggestion
                { "commit", "-m", *commitMessage, "--"  *sourcePath + "/" + std::string(file) });
```

src/libfetchers/path.cc
```diff
@@ -114,6 +130,9 @@ struct PathInputScheme : InputScheme
                     throw Error("path URL '%s' has invalid parameter '%s'", url.to_string(), name);
                 input->lastModified = lastModified;
             }
+            else if (name == "narHash")
+                // FIXME: require SRI hash.
```
Mic92, line 134: FIXME?

src/libfetchers/path.cc
```diff
@@ -134,6 +153,9 @@ struct PathInputScheme : InputScheme
                 input->revCount = getIntAttr(attrs, "revCount");
             else if (name == "lastModified")
                 input->lastModified = getIntAttr(attrs, "lastModified");
+            else if (name == "narHash")
+                // FIXME: require SRI hash.
```
Mic92, line 157: FIXME?

src/libfetchers/registry.cc
```diff
@@ -0,0 +1,222 @@
...
+    if (!pathExists(path))
+        return std::make_shared<Registry>(type);
+
+    try {
+
+        auto json = nlohmann::json::parse(readFile(path));
+
+        auto version = json.value("version", 0);
+
+        // FIXME: remove soon
```
Mic92, line 25: FIXME?

misc/bash/completion.sh
edolstra: If somebody wants to contribute that code, sure.

Mic92: cc @spwhitt

spwhitt: If we have something to share I support doing so, however I believe Eelco wants completions that are generated by the Nix command, and our current completions are not. Also @olejorgenb and @hedning are in charge of the project as far as I'm concerned, whatever they want to do is fine with me.

dasJ:
One thing I found when I played around with flakes: Is there any particular reason for the inputs to be non-lazy?
I evaluate ~30 systems with a single flake.nix and having to download every flake on every system is a bit unneeded (we have some large auto-generated flakes).
Lazy evaluation would have no downside I can see and would give potentially faster eval times for a lot of unused inputs (when evaluating a single system).

Valodim:
> nix build nixpkgs#hello

worth considering, the # is part of zsh globbing syntax (if extendedglob is enabled, which it always is in practice), where it works as a kleene-star. users of zsh will forever have to quote this syntax as  `nixpkgs\#hello` or `'nixpkgs#hello'`.

perhaps we could use some other character, like  `:`?

edolstra:
@dasJ Inputs *are* lazy (as long as they're locked). Please create an issue if this is not working correctly for you.

@Valodim Other characters were considered but `#` is the least ambiguous and it fits in the URL syntax.

Mic92:
To relief the pain for zsh users: I think it is possible to disable extend glob patterns for certain commands:

I use for example in my zshrc:

```
alias curl="noglob curl"
```

There is probably a more elegant way to just disable extended globs.

Kloenk:
`nix flake ckeck` evaluate every packages and legacyPackages attribute. This breaks most of the time, if your flake is supporting other systems besides the one you are currently running.

edolstra:
@Kloenk Hm, how so? Evaluation of `packages` should always succeed, even with packages for different system types. `nix flake check` doesn't check `legacyPackages` at the moment BTW (except that it needs to be a set of system types).

Kloenk:
> @Kloenk Hm, how so? Evaluation of `packages` should always succeed, even with packages for different system types. `nix flake check` doesn't check `legacyPackages` at the moment BTW (except that it needs to be a set of system types).

Ok. The flakes version of home-manager breaks, because it cannot evaluate some of the system types, but I guess this should be fixed in home-manager then.

Kloenk:
> @Kloenk Hm, how so? Evaluation of `packages` should always succeed, even with packages for different system types. `nix flake check` doesn't check `legacyPackages` at the moment BTW (except that it needs to be a set of system types).

If I understand this output correctly, it tries to build the `check`.* attributes for every system:
~~~
$ nix flake check --show-trace
warning: Git tree '/home/kloenk/proj/home-manager' is dirty
warning: unknown flake output 'lib'
error: a 'x86_64-darwin' with features {} is required to build '/nix/store/7xjyrlvjqdipyicjaxp7bwh1vxx716ga-home-manager.drv', but I am a 'x86_64-linux' with features {kvm, nixos-test, recursive-nix}
~~~

Kloenk:
> > @Kloenk Hm, how so? Evaluation of `packages` should always succeed, even with packages for different system types. `nix flake check` doesn't check `legacyPackages` at the moment BTW (except that it needs to be a set of system types).
> 
> If I understand this output correctly, it tries to build the `check`.* attributes for every system:
> 
> ```
> $ nix flake check --show-trace
> warning: Git tree '/home/kloenk/proj/home-manager' is dirty
> warning: unknown flake output 'lib'
> error: a 'x86_64-darwin' with features {} is required to build '/nix/store/7xjyrlvjqdipyicjaxp7bwh1vxx716ga-home-manager.drv', but I am a 'x86_64-linux' with features {kvm, nixos-test, recursive-nix}
> ```

Oh, its not the `check` attribute, but the `package` attribute. If I comment out the checks completely I still get this error.

src/libfetchers/github.cc
```diff
@@ -8,21 +8,18 @@
 
 namespace nix::fetchers {
 
-std::regex ownerRegex("[a-zA-Z][a-zA-Z0-9_-]*", std::regex::ECMAScript);
-std::regex repoRegex("[a-zA-Z][a-zA-Z0-9_-]*", std::regex::ECMAScript);
-
-struct GitHubInput : Input
+struct GitArchiveInput : Input
```
cole-h, line 11: Since this file now contains both GitHub and GitLab functionality, maybe it should be renamed to something like `git-forge(s).cc`? Just a thought to make it somewhat more obvious what the file encompasses, when skimming by filename. (Despite "git forge" being a somewhat recent phrase, I feel it's even less likely someone will know GitLab is also implemented in this file just by reading the filename.)

Valodim: "git forge" seems like a term specific from the emacs community. it certainly sounds more like a brand than a generic term to me. `git platform`?

(we should probably resist bikeshedding the issue beyond this)

src/libstore/globals.hh
```diff
@@ -369,6 +369,9 @@ public:
 
     Setting<bool> warnDirty{this, true, "warn-dirty",
         "Whether to warn about dirty Git/Mercurial trees."};
+
+    Setting<std::string> flakeRegistry{this, "https://github.com/NixOS/flake-registry/raw/master/flake-registry.json", "flake-registry",
+        "Path or URI of the global flake registry."};
```
Ericson2314, line 374: Let's move non-store-related settings out of `libstore`.

src/libfetchers/github.cc
cole-h: I don't think it's Emacs-specific, seeing as e.g. https://forgeperf.org/ exists and is in no way related to Emacs (not to mention SourceForge). However, I agree bikeshedding isn't fun. I just thought it would be nice to disambiguate (or inspire the idea of disambiguation).

kisik21: Git hosting is also a nice term, and a more widely-used one AFAIK.

dasJ:
Not sure if this is a regression on the flakes branch or if it's a regular nix bug (if so, I'll open another issue), but `nix copy -s` seems to be faulty **when copying .drv files**. It builds them locally, copies the .drv files to the target machine, leaving the build outputs on the local machine. Imo, No build should happen on `nix copy`.

andir:
I just tried to test something out with flakes and noticed that a bunch of the commands are working without the actual flakes feature being enabled. Most noticeably `flake init`, `flake info` and `flake clone`

```
$ nix --version
nix (Nix) 2.4pre20200521_00b562c

# nix show-config | grep experi
experimental-features = nix-command

$ nix flake --help
Usage: nix flake COMMAND FLAGS... ARGS...

Common flags:

Available commands:
  archive      copy a flake and all its inputs to a store
  check        check whether the flake evaluates and run its tests
  clone        clone flake repository
  info         list info about a given flake
  init         create a skeleton 'flake.nix' file in the current directory
  list-inputs  list flake inputs
  show         show the outputs provided by a flake
  update       update flake lock file

Note: this program is EXPERIMENTAL and subject to change.

$ nix flake update
error: experimental Nix feature 'flakes' is disabled

$ nix flake check
error: experimental Nix feature 'flakes' is disabled

$ nix flake archive
error: experimental Nix feature 'flakes' is disabled

$ nix flake clone --dest .
error: do not know how to clone input 'path:/home/andi/dev/nixos/flakes-test/my-flake'

$ nix flake init

$ nix flake info
Resolved URL:  path:/home/andi/dev/nixos/flakes-test/my-flake
Locked URL:    path:/home/andi/dev/nixos/flakes-test/my-flake?narHash=sha256-f6i+EAwvH806wJyR9XHAbqKpuaxJbbZzaiaDcmkI9fY=
Description:   A flake for building Hello World
Path:          /nix/store/i7igvwwd14wbk4q9f59647p9aad1ir08-source

$ nix flake list-inputs
error: experimental Nix feature 'flakes' is disabled

$ nix flake show
error: experimental Nix feature 'flakes' is disabled

$ nix flake update
error: experimental Nix feature 'flakes' is disabled
```

edolstra:
@andir Thanks, fixed.

andir:
I did a bit more thinking about the new `getFlake` builtin.

1. `builtins.getFlake` and `getFlake` are always present even if flakes is disabled. This makes it rather hard to implement a fallback / runtime-detection of flakes. For previous builtins (most commonly seen in the nixpkgs lib) we used the pattern `builtins.primOp or fallbackImplInNix`. With `getFlake` that does not currently work. It is throwing an error instance that can't be catched via `tryEval`. In case flake are (not yet) enabled the builtin should be missing. The observable behaviour should be indistinguishable from a Nix installation without the flake code paths. This allows users to either implement a fallback code path or to emit a sane warning that is sensible for their environment.

2. IMHO It shouldn't be in the default prelude. Having it within `builtins` should be enough. This way we do not clutter that global namespace even further.

3. With the flakes feature disable the builtin actually does attempt to do some work / input sanitation:

```
$ nix show-config | grep exp
experimental-features = nix-command
$ nix repl
nix-repl> builtins.getFlake
«primop»

nix-repl> builtins.getFlake ""
error: '' is not a valid URL

nix-repl> builtins.getFlake "github:nixos/nixpkgs"
error: experimental Nix feature 'flakes' is disabled
```
The builtin shouldn't even attempt to parse the flakref if it isn't enabled.

src/libexpr/eval.cc
```diff
@@ -161,12 +162,12 @@ const Value *getPrimOp(const Value &v) {
...
 
-string showType(const Value & v)
+string showType(ValueType type)
 {
-    switch (v.type) {
+    switch (type) {
         case tInt: return "an integer";
-        case tBool: return "a boolean";
-        case tString: return v.string.context ? "a string with context" : "a string";
+        case tBool: return "a Boolean";
```
andir, line 169: pardon me if I don't get it but why does the `b` have to be capitalized in this case?

src/libexpr/eval.cc
```diff
@@ -471,14 +501,21 @@ Value * EvalState::addConstant(const string & name, Value & v)
 Value * EvalState::addPrimOp(const string & name,
     size_t arity, PrimOpFun primOp)
 {
+    auto name2 = string(name, 0, 2) == "__" ? string(name, 2) : name;
+    Symbol sym = symbols.create(name2);
+
+    /* Hack to make constants lazy: turn them into a application of
```
andir, line 507: Why is this required? I think the comment (or commit message) should motivate that action and not just describe it.

src/libexpr/eval.cc
edolstra: Because "Boolean" comes from a name (George Boole).

edolstra:
@andir Thanks for the suggestion. `builtins.getFlake` now only exists if the `flakes` feature is enabled (and `getFlake` is gone).

src/nix/installables.cc
```diff
@@ -440,6 +440,16 @@ ref<eval_cache::EvalCache> openEvalCache(
         }));
 }
 
+static std::string showAttrPaths(const std::vector<std::string> & paths)
+{
+    std::string s;
```
andir, line 445: This might be a bit nit-picky but I think we can improve readability and reduce duplicate code (e.g. the concat and quote code) by just implementing it like this:

```cpp
static std::string showAttrPaths(const std::vector<std::string> & paths)
{
        std::string s;
        auto quotedPaths = quoteStrings(paths);
        auto it = quotedPaths.begin();

        if (it != quotedPaths.end())
                s = *it++;

        if (it != quotedPaths.end()) {
                auto rem = concatStringsSep(", ", std::vector<std::string>(it, quotedPaths.end()));
                s += " or " + rem;
        }

        return s;
}
```

This removes the ternary operator and the line that contains three different statements that mostly consists of quote & operator characters. Worst case this will create a copy of the remaining elements of the vector but IMHO that shouldn't hurt given the benefits.

gilligan: I sort of got nerd-sniped reading this and tried to write something that i can parse quite easily:

```c++
/**
 * Takes a list of paths and turns them into a comma separated list
 * of the format shown below:
 *
 * { }               -> ""
 * { "x" }           -> "'x'"
 * { "x", "y" }      -> "'x' or 'y'
 * { "x", "y", "z" } -> "'x' or 'y', 'z'"
 */
static std::string showAttrPaths(const std::vector<std::string> & paths)
{
    if (paths.size() == 0)
        return "";

    if (paths.size() == 1)
        return "'" + paths[0] + "'";

    auto quotedPaths = quoteStrings(paths);
    auto it = std::next(quotedPaths.begin());

    return "'"
        + paths[0]
        + "' or "
        + concatStringsSep(", ", std::vector<std::string>(it, quotedPaths.end()));
}
```

By the way: In the case of an empty `paths` vector, should the result really just be an empty string?

B4dM4n:
I recently started experimenting with flakes and hit a similar issue as @Kloenk with `nix flake check`.

When using the `forAllSystems` idiom to support multiple systems and adding an `apps` or `defaultApp` output `nix flake check` fails with something like this on a `x86_64-linux` system:
```log
a 'aarch64-linux' with features {} is required to build '/nix/store/9wpnn4am8jc1vyp40ks6hizk3m53d4r0-hello.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
```

[example flake gist](https://gist.github.com/B4dM4n/7dec32408bd8355f6cf86381243cfd90)

I fixed this error with bf1deda1d58536ce299eab38ffcbd6e3816c34b0 (will open a PR later).

Working on the fix made me realize that checks can currently only be build for the host system, which limits their usability a bit. For CI environments this is not a big issue as the check can be run on many hosts, but for personal usage I would like to run `nix flake check` on my local machine and build the checks for all supported systems (`settings.thisSystem`, `settings.extraPlatforms` and remote builds).

A possible solution I came up with is 5fb4cc370412b868200a5b8634b5edc850fdf954, which replaces the `--no-build` flag with a `--build-systems` flag. The new flag defaults to the local system to retain the current behavior, but allows to specify the exact set of systems that will be build (empty list is equivalent to `--no-build`).

Example usage:
```shell
nix flake check --build-systems ""
nix flake check --build-systems "x86_64-linux x86_64-darwin i686-linux"
```

I can open a PR for this proposal for further discussion if desired.

Ericson2314:
I am not up date with flakes, so perhaps I misunderstand what @B4dM4n but it sounds like more evidence that we really need to move away from building and running tests in the same derivation and towards 3 derivations:

1. Build regular thing
2. Build tests
3. Run tests

It's mainly on Nixpkgs to just do this, but any way we can get the tooling to nudge us in that direction is a plus.

B4dM4n:
@Ericson2314 The issue I described has nothing to do with regular package tests. `nix flake check` just helps to ensure that a `flake.nix` follows the basic structure for Nix flakes and also builds some of the outputs (currently `apps`, `defaultApp` and `checks`).

Ericson2314:
At least `checks` is it's own derivation. I suppose 1 and 3 needs names, but 2 doesn't, in my list above, at a bare minimum.

mkenigs:
I've been thinking about usability, and I guess some of this might be more of my thoughts related to nix in general. But if I was trying to to convince someone to use nix, I would want a flake to look something more like this:
```
{
  description = "optional description";
  inputs.build = [ boost ];
  inputs.test = [ testframework ];
  outputs = { self, inputs }: {
    mkFlake {
      name = "name";
      build =
        ''
          make
        '';
      install =
        ''
          cp name $out/bin/
        '';
      test =
        ''
          test
        '';
    }
  }
}
```

I don't know how things like nixpkgs, system, mkDerivation, and other outputs would be handled, so I guess this is more just my thoughts rather than something concrete. But if it was possible, I think something along these lines would be way easier to digest as a newcomer to nix. It's pretty easy to understand what's going on: you say what dependencies you need, nix does it's nix magic, and you get something you can build/run/develop/test anywhere with `nix build`/ `run`/`develop`/`test (flake check)`. I think it's hard to realize that's what nix can do when it's hidden behind complexity, but it's actually incredibly appealing once you see it.

edolstra:
@mkenigs Agree completely but that's out of scope for the first iteration of flakes. In the future we might have a `flake.toml` containing a high-level description of a package, without any extraneous cruft like `mkDerivation`, function calls etc.

PR #3576, opened 2020-05-07
build: don't log raw stderr

LnL7:
Second attempt, this time removing raw stderr from the build side instead of trying to determine what to show on the client.

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -4132,13 +4132,9 @@ void DerivationGoal::flushLine()
         ;
 
     else {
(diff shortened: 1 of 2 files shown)


LnL7:
Right, ssh doesn't implement RemoteStore so none of the logging is handled there.

I'm not sure removing raw stderr from the build side is the way to go then, until legacy ssh is removed. However that means there's no way to distinguish build logs from other output (like hash mismatches) unless those are all converted to a structured format.

@edolstra Thoughts?

PR #3577, opened 2020-05-08
Improve 'waiting for' message by supplying a reason for the message

### LINKED ISSUE #2069: "waiting for locks or build slots" is not a very helpful message

balsoft:
Resolves #2069 by allowing to pass the reason for which we are waiting to `Worker::wait*` functions.

Implementation details: stores the latest reason for which some `DerivationGoal` is sleeping in `Worker`, and if a `wait*` function is called with a reason different from the previous one, that reason is printed to stderr. Example output:

```
$ nix build -f test-nix.nix --max-jobs 8 --no-substitute
waiting for build slots...
```

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -1469,7 +1469,7 @@ void DerivationGoal::tryToBuild()
             case rpPostpone:
                 /* Not now; wait until at least one child finishes or
                    the wake-up timeout expires. */
-                worker.waitForAWhile(shared_from_this(), "build hook");
+                worker.waitForAWhile(shared_from_this(), "a build hook");
                 outputLocks.unlock();
                 return;
             case rpDecline:
@@ -4456,7 +4456,7 @@ void SubstitutionGoal::tryToRun()
        a substituter to run.  This is because substitutions cannot be
        distributed to another machine via the build hook. */
     if (worker.getNrLocalBuilds() >= std::max(1U, (unsigned int) settings.maxBuildJobs)) {
-        worker.waitForBuildSlot(shared_from_this(), "substitution slot");
+        worker.waitForBuildSlot(shared_from_this(), "a substitution slot");
         return;
     }
 
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -257,6 +257,11 @@ private:
     /* Cache for pathContentsGood(). */
     std::map<StorePath, bool> pathContentsGoodCache;
 
+    /* Name of resource for which we are currently waiting (e.g locks, build slots) */
+    string blockingResourceName;
+
+    void updateBlockingResourceName(string resourceName);
+
 public:
 
     const Activity act;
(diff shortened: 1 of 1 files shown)


src/libstore/build.cc
```diff
@@ -1464,7 +1469,7 @@ void DerivationGoal::tryToBuild()
             case rpPostpone:
                 /* Not now; wait until at least one child finishes or
                    the wake-up timeout expires. */
-                worker.waitForAWhile(shared_from_this());
+                worker.waitForAWhile(shared_from_this(), "build hook");
```
kirelagin, line 1472:
```suggestion
                worker.waitForAWhile(shared_from_this(), "a build hook");
```

src/libstore/build.cc
```diff
@@ -4451,7 +4456,7 @@ void SubstitutionGoal::tryToRun()
        a substituter to run.  This is because substitutions cannot be
        distributed to another machine via the build hook. */
     if (worker.getNrLocalBuilds() >= std::max(1U, (unsigned int) settings.maxBuildJobs)) {
-        worker.waitForBuildSlot(shared_from_this());
+        worker.waitForBuildSlot(shared_from_this(), "substitution slot");
```
kirelagin, line 4459:
```suggestion
        worker.waitForBuildSlot(shared_from_this(), "a substitution slot");
```

src/libstore/build.cc
```diff
@@ -4706,27 +4711,36 @@ void Worker::childTerminated(Goal * goal, bool wakeSleepers)
...
+        printError(format("waiting for %1%...") % resourceName);
+    }
+}
 
-void Worker::waitForBuildSlot(GoalPtr goal)
+void Worker::waitForBuildSlot(GoalPtr goal, string resourceName)
 {
-    debug("wait for build slot");
+    debug("wait for build slots");
+    updateBlockingResourceName(resourceName);
```
kirelagin, line 4724: I think this line will print the message even if we are not actually waiting, i.e. if we take the first branch in the `if` below. I’m not sure if that is the desired behaviour or not.

balsoft: Yes, thanks for noticing that. I think the reason I did that is that `debug("wait for build slot")` was here...

balsoft:
No, not exactly. It prints those messages slightly earlier in the code -- right at the time when the waiting is scheduled, whereas previously it was printed before the thread goes to sleep. However, it shouldn't be much of a difference (if I understand the execution path correctly).

edolstra:
Thanks, I've gone for a slightly different approach in 5ed5d7acbd00298df7b4e3a638d731143084da2e (using `Activity` to make "waiting for lock" messages show in the progress bar).

balsoft:
Thanks!

PR #3583, opened 2020-05-09
Installables refactor

mkenigs:
Simplifies the inheritance structure for Installables, so that parent classes only contain functionality used by all children. Makes Installables less interdependent with commands, and helps #3565


edolstra:
Thanks!

PR #3584, opened 2020-05-10
nix-env: validate json outputsToInstall

LnL7:
This isn't particularly useful since the json output already includes
metadata.  However the evaluation behaves differently, in this case
`meta.outputsToInstall` is also validated.  The same is done by the
channel index generation but it's currently not exposed in a way other
tools like ofborg could use it.

https://github.com/NixOS/nixos-channel-scripts/blob/7a681103b2b3ce150f7f96394f19aa0ad4797ca1/generate-programs-index.cc#L110

If this flag is added to the ofborg evaluation checks it should catch problems like https://github.com/NixOS/nixpkgs/pull/87138 in the future.

/cc @grahamc @cole-h

### DIFF
--- a/src/libexpr/get-drvs.cc
+++ b/src/libexpr/get-drvs.cc
@@ -117,7 +117,7 @@ DrvInfo::Outputs DrvInfo::queryOutputs(bool onlyOutputsToInstall)
     /* Check for `meta.outputsToInstall` and return `outputs` reduced to that. */
     const Value * outTI = queryMeta("outputsToInstall");
     if (!outTI) return outputs;
-    const auto errMsg = Error("this derivation has bad 'meta.outputsToInstall'");
+    const auto errMsg = Error("derivation '%s' has bad 'meta.outputsToInstall'", name);
         /* ^ this shows during `nix-env -i` right under the bad derivation */
     if (!outTI->isList()) throw errMsg;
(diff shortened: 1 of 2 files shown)


doc/manual/command-ref/nix-env.xml
```diff
@@ -1144,6 +1145,14 @@ user environment elements, etc. -->
 
   </varlistentry>
 
+  <varlistentry><term><option>--outputs</option></term>
+
+    <listitem><para>Print the outputs of the derivation.
+    Unlike <option>--out-path</option> only the outputs that would be installed are
```
cole-h, line 1151: Nit:

```suggestion
    Unlike <option>--out-path</option>, only the outputs that would be installed are
```

LnL7:
Alternatively I could add this to `--json` somewhere, for example by replacing the plain `outputsToInstall` value with a lookup.

edolstra:
It seems misleading that `--outputs` *doesn't* print the outputs but only the outputs that would be installed.

I agree it would be better to add it to `--json`, since this info isn't really useful for human consumption.

LnL7:
I wasn't sure what the behaviour of `--json` was with regard to errors, but since ofborg uses it to check evaluation I assume it's expected to stop just like the other commands.

PR #3586, opened 2020-05-11
WIP: add fancy debug logs

mkg20001:


src/build-remote/build-remote.cc
```diff
@@ -162,7 +163,15 @@ static int _main(int argc, char * * argv)
                             bestLoad = load;
                             bestSlotLock = std::move(free);
                             bestMachine = &m;
+                        } else {
+                            debug("skipping '%s': there is a better machine", m.storeUri);
```
domenkozar, line 167: What does better mean?

mkg20001: actually I'm not sure either, but I'm guessing it's based on the parameters above a machine that has a higher speed factor and is available right now

src/build-remote/build-remote.cc
```diff
@@ -162,7 +163,15 @@ static int _main(int argc, char * * argv)
...
                         }
+                    } else {
+                            debug("skipping '%s': does not meet all condiditions\n - enabled = %s\n - system: type is matching = %s / required %s / got %s\n - features: all supported = %s / mandatory supported = %s / required %s / got %s", m.storeUri,
+                                m.enabled,
+                                std::find(m.systemTypes.begin(), m.systemTypes.end(), neededSystem) != m.systemTypes.end(), neededSystem, join(m.systemTypes, ","),
```
mkg20001, line 172: the `join()` function call here should be replaced by the `concatStringsSep()` one

Issue #3575, opened 2020-05-06
install-nix script installs nixpkgs-unstable by default


prusnak:
While experimenting with `nixos/nix` docker container I found out that `install-nix-from-closure.sh` automatically adds nixpkgs-unstable channel by default, because of the following:

https://github.com/NixOS/nix/blob/672985531c6b9481f8eb8210a17328dfc2515c9b/scripts/install-nix-from-closure.sh#L132-L135

This channel automatically becomes part of the docker image. This can be fixed easily (see my patch in https://github.com/NixOS/docker/issues/18), but I am raising a question whether this is really desired to have this automatically installed.

I personally think the addition of the channel(s) should happen manually and not automatically.

prusnak:
If adding a nixpkgs-channel is really desired by default, we could at least provide a way how to skip this step via an environment variable.

domenkozar:
@prusnak you're welcome to open a PR :)

prusnak:
PR in https://github.com/NixOS/nix/pull/3588

PR #3588, opened 2020-05-12
Introduce NIX_INSTALLER_NO_CHANNEL_ADD which skips nix-channel --add

prusnak:
Fixes https://github.com/NixOS/nix/issues/3575

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -40,30 +40,43 @@ elif [ "$(uname -s)" = "Linux" ] && [ -e /run/systemd/system ]; then
 fi
 
 INSTALL_MODE=no-daemon
-# Trivially handle the --daemon / --no-daemon options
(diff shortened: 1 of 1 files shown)


edolstra:
Why an environment variable rather than a command line flag? Such variables are pretty undiscoverable.

prusnak:
In order to add a new command-line flag I'd need to rework the current naive flag handling. Is that desired within the same PR?

domenkozar:
@prusnak I think it's fine to do it as part of the same PR.

prusnak:
Added commit 9e12b2f5b86963bc6d3a4ee37cc759357fb62263 which exposes installer config env vars via command line flags.

prusnak:
Thanks!

LnL7:
FYI this doesn't seem to work in combination with `--daemon`, that's still updating a channel for me.

domenkozar:
Weird, the code seems to be correct to me.

PR #3590, opened 2020-05-13
nix errors phase 2

bburdette:
This PR is for phase 2 of the [nix-error-proposal](https://github.com/bburdette/nix-error-proposal/blob/master/proposal.md) project.

Overall the idea is to enhance the format of error messages, and their content.  This phase is just for the formatting part, the content of error messages is largely left alone.

To use the new format there are some new macros in logging.hh:
```
    logError
    logWarning
```
These require an ErrorInfo struct as an argument instead of a format string.  In the default logger, the ErrorInfo is used to print errors to cerr.  This late conversion of ErrorInfo to string should help with future language server support, or other similar efforts.

These macros prevent creating unnecessary ErrorInfo structs, similar to the printInfo and printError macros.  That is, filtering takes place before the struct is created, not after.

ErrorInfo uses a special fmt - 'hintfmt' - that highlights templated values in yellow.  For consistency, I changed the printError/warn/printInfo macros to use the same format style.  That also required removing the remaining calls to boost::format.

ErrorInfo usage fits into two big categories:  BaseError descendants, and converted printError messages.  

**Exceptions**  

BaseError descendants - Error, SysError, and so forth, all use the new format now.  The MakeError macro uses the class name as the error name, so SysError will print:

```
  error: --- SysError ------------------------------------------ error-demo
  statting file : Bad file descriptor
```

Pos goes into ErrorInfo as a special field, and gets printed at a standardized place in the message.  It got removed from the string templates so as not to duplicate information.

**printError**

For printError messages, there were more decisions to make.  The ErrorInfo format is pretty heavy and not always appropriate.  I categorized printError calls as follows:
  1) *Problems.*  reporting a problem that may require user intervention.
    - converted to ErrorInfo format.
  2) *Status.*  printing status messages that inform the user of progress.
    - converted to printInfo.
  3) *Warnings.*  printing warnings with error verbosity, as in printError("warning: ...
    - converted to logWarning()
  4) *Other.*  technical uses, like forwarding remote build messages to cerr.
    - left alone.

warn() messages.
  these I left alone for now; reasoning that the `printError("warning..."` ones are more important, so those are distiguished by logWarning calls.  warn() calls are intact other than converting from boost::format.  

**What's missing**

This is already a huge PR, so I'm leaving some things out that belong in phase 2.

- show-trace is not included in ErrorInfo.  It still works as it has in the past, but its printed to cerr and doesn't make it into the logger except as strings.

- lines of code.  This is a big part of the new format!  But I think there's enough to look at in this PR.  Its next on my list.

**Moving code around.**

BaseError and descendants got moved out of types.hh (and util.cc) and into error.hh.

ErrorInfo now uses Verbosity instead of its own ErrLevel type.  Consequently, Verbosity was moved from logging.hh to error.hh.

Theres a new header, fmt.hh, where fmt, FormatOrString, hintfmt, and so forth are defined.

**Misc**

In eval.cc and eval-inline are some functions like throwEvalError, that used pos in a string template.  With pos being passed to ErrorInfo instead, I found it confusing to leave pos as the last argument in the function:

```
// old version: pos used as a value in the string template
throwTypeError("value is %1% while a list was expected, at %2%", v, pos);

// new version: pos is the first argument, indicating that its not used 
// in the template.
throwTypeError(pos, "value is %1% while a list was expected", v);
```

It seems like cpu affinity was meant to be printed in affinity.cc, but wasn't.  I added an ostream operator<< for affinity, so I could print it:
```
{
        std::ostringstream oss;
        oss << savedAffinity;
        printError("failed to restore CPU affinity %1%", oss.str());
}
```


domenkozar:
@edolstra could you take a look?

bburdette:
So I've been working on the lines-of-code feature, but I've been cautioned about including yet more changes in this PR!  Still, a preview:

Selection_042

Thankfully this will be a much smaller PR than the current one.

edolstra:
Looks great, thanks!

Style note: In `logError` calls, we can omit the `ErrorInfo` constructor since it can be inferred. This makes these calls less verbose and less deeply nested. For example:
```
diff --git a/src/build-remote/build-remote.cc b/src/build-remote/build-remote.cc
index 0caa97fa..d0d2c1b1 100644
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -200,12 +200,11 @@ static int _main(int argc, char * * argv)
 
                 } catch (std::exception & e) {
                     auto msg = chomp(drainFD(5, false));
-                    logError(
-                        ErrorInfo { 
-                            .name = "Remote build",
-                            .hint = hintfmt("cannot build on '%s': %s%s",
-                        bestMachine->storeUri, e.what(),
-                        (msg.empty() ? "" : ": " + msg))
+                    logError({
+                        .name = "Remote build",
+                        .hint = hintfmt("cannot build on '%s': %s%s",
+                            bestMachine->storeUri, e.what(),
+                            (msg.empty() ? "" : ": " + msg))
                     });
                     bestMachine->enabled = false;
                     continue;
```
Given that there are so many of these calls I think that's worthwhile.

Some nitpicks about the output: there are multiple trailing empty lines for some error messages, e.g.
```
$ nix build /foo
error: --- SysError ------------------------------------------------- nix
getting status of '/foo' : No such file or directory



```
and there shouldn't be a space in front of the colon (`'/foo' :`).

src/error-demo/error-demo.cc
```diff
@@ -1,66 +1,157 @@
...
+
 int main()
 {
-    using namespace nix;
+    verbosity = lvlVomit;
 
     // In each program where errors occur, this has to be set.
     ErrorInfo::programName = std::optional("error-demo");
 
+    // 'DemoError' appears as error name.
```
gilligan, line 19: I wonder if it would be possible to turn this `demo` file into unit tests instead? Would that be feasible? WDYT @bburdette ?

bburdette: I am not opposed.  What do you envision as tests?  String comparison of predicted vs actual output?

gilligan: Yes i think that is what I would envision. I have to admit i didn’t yet find the time to take a closer look so maybe some parts might be hard(er) to test when log functions just print to stderr bur whenever an Exception is thrown with a message that encodes information in a certain format that could be tested.

src/libutil/error.hh
```diff
@@ -1,121 +1,164 @@
...
-    string errLineOfCode;
+    std::optional<string> errLineOfCode;
     std::optional<string> nextLineOfCode;
 };
 
-// ----------------------------------------------------------------
-// format function for hints.  same as fmt, except templated values
-// are always in yellow.
+// -------------------------------------------------
+// ErrorInfo.
```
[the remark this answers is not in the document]
bburdette: Ok added some comments at the top and removed the "ErrorInfo."

src/libutil/error.cc
```diff
@@ -15,80 +41,117 @@ std::ostream& operator<<(std::ostream &os, const hintformat &hf)
...
+            return fmt("(%1%)", errPos.line);
+        }
+    }
+    else {
+        return "";
+    }
 }
 
-void printCodeLines(const string &prefix, const NixCode &nixCode)
+void printCodeLines(std::ostream &out, const string &prefix, const NixCode &nixCode)
```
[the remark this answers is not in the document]
bburdette: comment added!

src/libutil/error.cc
```diff
@@ -2,9 +2,35 @@
...
+const std::string nativeSystem = SYSTEM;
+
+
+BaseError & BaseError::addPrefix(const FormatOrString & fs)
 {
+    prefix_ = fs.s + prefix_;
+    return *this;
+}
+
+const string& BaseError::calcWhat() const
```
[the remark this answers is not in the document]
bburdette: done

src/libutil/error.cc
```diff
@@ -2,9 +2,35 @@
...
+#include <sstream>
 
-namespace nix
+namespace nix {
+
+
+const std::string nativeSystem = SYSTEM;
+
+
+BaseError & BaseError::addPrefix(const FormatOrString & fs)
```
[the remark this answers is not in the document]
bburdette: comment added!

src/error-demo/error-demo.cc
gilligan: I actually just spent some time on this and am almost done converting everything from your `error-demo.cc` into a unit test. I'll probably have it done later this evening for you to check out.

bburdette: Wow just saw this!  I've been working on it too, but u

gilligan: Please have a look at:
https://github.com/gilligan/nix/commit/952e72c804ce79ffeee8f58261d4cbdeb9d56ac6

I would suggest to remove `error-demo` and replace it with that. Some notes:

- Shortly after i started i realized that this would really be tests for `logging.h` 
- I am using the google test to capture output and assert that. That seemed like the best approach to me
- You will see notes with `XXX` where i don't quite understand why `description` is not set.
- The asserts do look a bit ugly. It might be nicer to use c++ raw strings so the expected output is actually human readable. Using those in the `json.hh` tests:

```c++
        ASSERT_EQ(out.str(),
R"#({
  "list": [
    "element"
  ]
})#");
```

I was using those at first but then realized that because of the escape codes for coloring they wouldn't work. I guess ideally we'd still use those and add the escape codes (?). Or we leave that as an improvement for later. I'll leave it up to you.

bburdette: Nice!  The testing::internal::CaptureStderr() is helpful.  My approach was doing a test in bash; this is better I think.  

I do wonder about the empty strings on Talkative and etc.  On the plus side, its only the logError and logWarning that are used in actual code at this point.  Seems like a problem with the stderr capture somehow?  "Chatty" still has a "talkative" description; I made a note.  

Anyway overall LGTM.  Thanks for expediting this!

gilligan:
@bburdette thanks for incorporating the tests!

Ericson2314:
Hope we can get this merged soon! The status quo with the demo merge but not integrated is perfectly understandable as an intermediate step, but not something we should leave around as it makes best practices quite confusing!

Ericson2314:
Also, I was just ad-hoc adding some `strerr(errno)` to help myself debug something, which is a real bummer this PR systematically does that.

gilligan:
@edolstra is there anything that in your view makes this PR not ready to be merged?

edolstra:
@bburdette Thanks, merged!

bburdette:
Awesome, thanks for the merge!!

Issue #2531, opened 2018-11-09
manual: unfinished sentence


obfusk:
This sentence seems to be unfinished: "... in parallel, or set." Set what?
https://github.com/NixOS/nix/blob/master/doc/manual/expressions/simple-building-testing.xml#L81

PR #3592, opened 2020-05-16
Remove -j option from simple-build-testing

Mic92:
By default Nix/NixOS already sets a reasonable default `max-jobs = auto`
so we don't need to mention it in this tutorial.
The option is still documented in other parts of the documentation
if users ever stumble over this.

Fixes https://github.com/NixOS/nix/issues/2531


zimbatm:
added graham since he wrote that paragraph

PR #3600, opened 2020-05-20
Automatic UID allocation

edolstra:
This PR does the following:

* It adds an experimental feature and option `auto-uid-allocation` which provides an alternative to having a `nixbld` group of pre-created build users. When enabled, Nix allocates UIDs/GIDs in the range 872415232+ on Linux and 56930 on macOS.

* It adds an experimental feature `cgroups` that causes builds to be executed in a cgroup. This allows getting some statistics from a build (such as CPU time) and in the future may allow setting resource limits. But it mainly exists because the `uid-range` feature requires it.

* It adds a system feature `uid-range` that causes a build to be executed as root in a UID namespace with 65,536 UIDs available. This allows things like `systemd-nspawn` and NixOS containers to run inside a Nix build.


domenkozar:
A few questions/comments:

1. Does it work on macOS?
2. Why does it need to be experimental feature?
3. Missing documentation for the nix.conf option

edolstra:
No, this will never work on macOS since it requires cgroups.

It's an experimental feature so we can remove it if it turns out not to be a good idea, or change the interface (e.g. the names for the `requiredSystemFeatures`).

7c6f434c:
This needs the `systemd` cgroup to already exist on host, and moreover this needs the daemon to be inside this cgroup, right? (Unsure whether it could just create the necessary structure for the sandbox even if it is missing on the host)

edolstra:
Yes, it needs the `systemd` hierarchy to be mounted on `/sys/fs/cgroup/systemd`. We could create it if it doesn't exist (it's just `mount -t cgroup -o none,name=systemd none /sys/fs/cgroup/systemd`). I don't think it's possible for a process *not* to be in a cgroup hierarchy. (If you create a named hierarchy it will contain all processes in the system in the root cgroup.)

7c6f434c:
I think I have heard of people trying to run systemd on NixOS using only cgroup2; and also, of course, I wonder what will be eventually needed to run lighter-weight NixOS tests on non-NixOS Linux. I guess creating a `none,name=systemd` cgroup hierarchy could be just mentioned in the error message. Is there a simple test I should run on my non-systemd Nixpkgs-kernel system to see if systemd-nspawn indeed works in such setups?

edolstra:
I initially used the unified hierarchy instead of the systemd hierarchy. The unified hierarchy works but it creates an impurity: it depends on the host system configuration what controllers are available in the unified hierarchy. So it seems better to use the systemd hierarchy since it only allows process tracking and not resource control.

If you want to test it on a non-NixOS system, mount the systemd hierarchy and try to build this expression: https://gist.github.com/edolstra/5cb2ec5c79ac8faf208058fd9375b448

7c6f434c:
`auto-allocate-uids` needs both to be enabled and added to experimental features; there is a message saying `auto-allocate-uids = true` is needed, but a more generic message is actually printed.

Once things are configured: I do get «Container nixos exited successfully.», and the target path is built with reasonable output (and there is a running `systemd` inside the build but no apparent attempts at escaping happen).

Logs: firewall failing seems expected, `system-getty.slice` failing due to it is slightly surprising, the following might mean I am doing something wrong on the host system or maybe you also get these:
```
Failed to create symlink /sys/fs/cgroup/cpuacct: Read-only file system
Failed to create symlink /sys/fs/cgroup/cpu: Read-only file system
Failed to create symlink /sys/fs/cgroup/net_prio: Read-only file system
Failed to create symlink /sys/fs/cgroup/net_cls: Read-only file system
```

Overall `nix-store -r` time is <3s wall-clock time, which is just great, thanks for that feature.

7c6f434c:
Looked at the code and got an idea to test… `auto-allocate-uids` requires `systemd` cgroup hierarchy to exist even if `systemd-cgroup` feature is not enabled.

edolstra:
> auto-allocate-uids requires systemd cgroup hierarchy to exist even if systemd-cgroup feature is not enabled.

Yeah that's currently true, didn't think about that. In principle however it could use any existing hierarchy since it's only used for tracking processes.

7c6f434c:
… or even mount «nix» hierarchy if `systemd` one is not mounted, I guess…

I am OK with mounting `systemd` hierarchy on boot, I guess a slightly more detailed message is enough. I just have _no_ cgroup hierarchies mounted by default so it was very cheap for me to check this combination of options.

edolstra:
> firewall failing seems expected

Actually that succeeds for me. Which makes me realize a big problem with this approach: whether certain networking features (firewall, NAT, ...) work depend on what kernel modules are loaded in the host system. I don't think there is any way to restrict access...

7c6f434c:
Erm. Does that require structured required features (and, in this instance, assertions about host kernel modules) for a proper-ish solution? I guess whoever cares could write a very minimal `sinit` VM to demonstrate whether some dependency is not listed…

edolstra:
Or maybe a seccomp filter could be used to restrict access to undeclared features.

7c6f434c:
Seccomp filter sounds like something that requires ahead-of-time enumeration of all possible things that could go wrong

edolstra:
I just realized that the situation isn't that bad. Or rather, it was already bad and this doesn't make it worse. It was already possible to create network devices, firewall tables etc. depending on the host kernel configuration. For example:
```
with import <nixpkgs> {};

runCommand "foo"
  {
    buildInputs = [ pkgs.utillinux pkgs.iproute pkgs.iptables ];
  }
  ''
    unshare -m -n -U -r -- bash -c "
      set -e
      mkdir -p foo/run foo/nix/store
      mount --rbind /nix/store foo/nix/store
      ip link add foo-h type veth peer name foo-c
      chroot foo iptables -t nat -A POSTROUTING -p tcp -s 192.168.1.2
      chroot foo iptables -t nat -L
    "
    mkdir $out
  ''
```
This will even cause kernel modules like `veth` and `iptable_nat` to be loaded if they're not already.

7c6f434c:
Ah, it might be that the host dependency got exposed by the fact that I did not enable module autoloading. I guess this impurity was «mitigated» by there being little incentive to do such things in public expressions.

Ericson2314:
https://github.com/NixOS/nix/pull/4155 this fixed the conflicts. Just linking it before for some reason GitHub didn't.

knedlsepp:
What's needed to get this over the finish line? It looks like having this could enable some more close integration between container technologies and nix.

lukegb:
Re: `systemd-cgroup`: my understanding was that interacting with systemd's cgroup hierarchy is [very strongly discouraged](https://www.freedesktop.org/wiki/Software/systemd/ControlGroupInterface/), unless [delegation](https://systemd.io/CGROUP_DELEGATION/) is used (which sounds like what we should be doing if we want to run sub-systemd?)

Ericson2314:
I updated https://github.com/NixOS/nix/pull/4155 to fix more conflicts. Regardless of what meaningful changes are needed for this, I think that should be merged in (or everything squashed together or whatever) so we can stay in a no-conflict state.

davidscherer:
In addition to removing the limited number of build users and being useful for testing, this could make Nix much more efficient and capable at using and transforming arbitrary containers. Although that is rarely the first choice way to do anything it would be very useful.  What stands in the way of landing this?

src/libstore/lock.cc
```diff
@@ -1,93 +1,228 @@
...
+
+    std::optional<Path> getCgroup() override
+    {
+        if (!cgroup) {
+            /* Create a systemd cgroup since that's the minimum
+               required by systemd-nspawn. */
+            auto ourCgroups = getCgroups("/proc/self/cgroup");
+            auto systemdCgroup = ourCgroups["systemd"];
+            if (systemdCgroup == "")
+                throw Error("'systemd' cgroup does not exist");
```
roberth, line 178: I've tried this with this PR at 79152e3 as `nix.package` on NixOS inside a podman container using arion. With `auto-allocate-uids = true` I always get:

```console
# nix-build -I nixpkgs=channel:nixos-unstable --expr 'with import <nixpkgs> {}; hello.overrideAttrs(o: { n = 3; })'
warning: warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
  /nix/store/2pw20yqb6iy7f33kyn5pssxg7mvzmyaf-hello-2.10.drv
error: 'systemd' cgroup does not exist
```

However, I don't think this is unique to podman. Could it be caused by cgroupsv2?

Here's what it looks like on my host

```console
$ cat /proc/$(pidof nix-daemon)/cgroup 
0::/system.slice/nix-daemon.service
```

and it's pretty much the same in my container:

```console
# cat /proc/$(pidof nix-daemon)/cgroup 
0::/system.slice/nix-daemon.service
```

src/libstore/build/local-derivation-goal.cc
```diff
@@ -645,12 +646,32 @@ void LocalDerivationGoal::startBuilder()
                 dirsInChroot.erase(worker.store.printStorePath(*i.second.second));
         }
 
-#elif __APPLE__
-        /* We don't really have any parent prep work to do (yet?)
-           All work happens in the child, instead. */
+        if (useSystemdCgroup) {
+            settings.requireExperimentalFeature("systemd-cgroup");
```
Mic92, line 650: Is there anything that speaks against just using cgroup namespaces and create a new hierarchy?
Than this would not depend on what cgroups are available outside and leak less of information/build impurities.

wwilson:
What can we do to get this merged? It would be very useful for improving Nix’s ability to transform arbitrary (non-Nix) containers.

edolstra:
@wwilson Yeah I'd like to have a look at it soon since it would be great to have this merged.

mkenigs:
@wwilson a next step for testing this could be moving a nixos test to a container instead of a VM. That would require extending the test framework to support containers, which I took the first step towards in https://github.com/NixOS/nixpkgs/pull/139788 (just refactoring so far). @tfc wanted to wait to merge my PR until we actually sketch out what using containers will look like

Ericson2314:
https://github.com/NixOS/nix/pull/6181 fixed conflicts again.

src/libutil/experimental-features.cc
```diff
@@ -11,6 +11,8 @@ std::map<ExperimentalFeature, std::string> stringifiedXpFeatures = {
     { Xp::NixCommand, "nix-command" },
     { Xp::RecursiveNix, "recursive-nix" },
     { Xp::NoUrlLiterals, "no-url-literals" },
+	{ Xp::AutoAllocateUids, "auto-allocate-uids" },
+	{ Xp::SystemdCgroup, "systemd-cgroup" },
```
Ericson2314, line 15: Aye, a tab snuck in, my bad. I hope we can get a linter for this.

wwilson:
At Antithesis we've developed a pure-Nix container construction and management tool, sort of like a much more powerful version of pkgs.dockertools. We would love to upstream it so the rest of the community can benefit, but it currently depends on this PR, fyi.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -502,6 +495,9 @@ void LocalDerivationGoal::startBuilder()
         }
     }
 
+    useUidRange = parsedDrv->getRequiredSystemFeatures().count("uid-range");
+    useSystemdCgroup = parsedDrv->getRequiredSystemFeatures().count("Systemd-cgroup");
```
cole-h, line 499: I have the funny feeling this shouldn't be capitalized?

```suggestion
    useSystemdCgroup = parsedDrv->getRequiredSystemFeatures().count("systemd-cgroup");
```

src/libstore/globals.hh
```diff
@@ -275,6 +275,19 @@ public:
           multi-user settings with untrusted users.
         )"};
 
+    #if __linux__
+    Setting<bool> autoAllocateUids{this, false, "auto-allocate-uids",
+        "Whether to allocate UIDs for builders automatically."};
+
+    const uint32_t idsPerBuild = 1 << 16;
+
+    Setting<uint32_t> startId{this, 872415232, "start-id",
```
cole-h, line 284: Where does this number come from, out of curiosity?

src/libstore/local-store.cc
```diff
@@ -890,48 +897,48 @@ void LocalStore::queryPathInfoUncached(const StorePath & path,
 
 std::shared_ptr<const ValidPathInfo> LocalStore::queryPathInfoInternal(State & state, const StorePath & path)
 {
-    /* Get the path info. */
```
cole-h, line 893: Nit: Most of the diff in this file from this point on seems weirdly spaced to me.

src/libstore/globals.hh
```diff
@@ -275,6 +283,27 @@ public:
           multi-user settings with untrusted users.
         )"};
 
+    Setting<bool> autoAllocateUids{this, false, "auto-allocate-uids",
+        "Whether to allocate UIDs for builders automatically."};
+
+    Setting<uint32_t> startId{this,
+        #if __linux__
+        872415232,
```
lheckemann, line 291:
```suggestion
        0x34000000,
```
to make it clear that it's a "round-ish" number?

Ericson2314:
#7330 FYI for the test you were trying to write, @edolstra.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -10,3 +10,43 @@
   This avoids a lot of spurious errors where some benign strings end-up having
   a context just because they are read from a store path
   ([#7260](https://github.com/NixOS/nix/pull/7260)).
+
+* Nix can now automatically pick UIDs for builds, removing the need to
+  create `nixbld*` user accounts. these UIDs are allocated starting at
+  872415232 on Linux and 56930 on macOS.
```
cole-h, line 16: Would it be possible to add / change these to the hex values? Somebody grepping the codebase for this magic number won't find it since you changed it to hex.

src/libstore/globals.hh
```diff
@@ -275,6 +283,27 @@ public:
...
         )"};
 
+    Setting<bool> autoAllocateUids{this, false, "auto-allocate-uids",
+        "Whether to allocate UIDs for builders automatically."};
+
+    Setting<uint32_t> startId{this,
+        #if __linux__
+        0x34000000,
+        #else
+        56930,
```
cole-h, line 293: Also, maybe change this to hex as well for symmetry?

edolstra: This one is pretty random (0xde62) so not very interesting in hex.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -363,6 +400,60 @@ static void linkOrCopy(const Path & from, const Path & to)
...
+        throw Error("cgroups are not supported on this platform");
+        #endif
+    }
+
+    /* Make sure that no other processes are executing under the
+       sandbox uids. This must be done before any chownToBuilder()
+       calls. */
+    killSandbox(false);
+
     /* Right platform? */
```
Ericson2314, line 457: Any reason this code is no longer at the top? It makes sense to me to check the drv itself before doing IO like checking the filesystem --- cheap checks first.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -2330,6 +2434,10 @@ DrvOutputs LocalDerivationGoal::registerOutputs()
                 sink.s = rewriteStrings(sink.s, outputRewrites);
                 StringSource source(sink.s);
                 restorePath(actualPath, source);
+
+                /* FIXME: set proper permissions in restorePath() so
+                   we don't have to do another traversal. */
+                canonicalisePathMetaData(actualPath, {}, inodesSeen);
```
Ericson2314, line 2440: This just become needed?

src/libstore/build/local-derivation-goal.cc
```diff
@@ -580,10 +642,11 @@ void LocalDerivationGoal::startBuilder()
 
         printMsg(lvlChatty, format("setting up chroot environment in '%1%'") % chrootRootDir);
 
-        if (mkdir(chrootRootDir.c_str(), 0750) == -1)
+        // FIXME: make this 0700
```
Ericson2314, line 645: This FIXME is fairly short so I doubt I would know how to interpret it a long time from know. Should we file an issue and link it here?

ncfavier:
This is nice, but it breaks nix-top which [gets build users from the `nixbld` group](https://github.com/samueldr/nix-top/blob/6cb88ff9870998156ac6311df1a3446d7fb69769/nix-top#L38-L40). Could we have a Nix command that lists the UIDs for the currently active builds? Or does this PR make nix-top obsolete?

PR #3607, opened 2020-05-21
Handle circular flake dependencies in list-inputs

mkenigs:
`nix flake list-inputs` currently segfaults for circular flake dependencies. This PR keeps track of visited nodes and only visits them once. Second commit adds a test that currently segfaults.


edolstra:
Thanks!

Issue #3456, opened 2020-03-30
Use /etc/zshenv instead of /etc/zshrc on macOS Catalina


surajbarkale:
By default, nix adds following lines to `/etc/zshrc`
```shell
# Nix
if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
  . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
fi
# End Nix
```
However, this file can get overwritten during OS update. This happened for me when upgrading to macos 10.15.4 recently.

I suggest writing these lines to `/etc/zshenv` instead which is not present by default and is loaded by zsh.

lilyball:
Looks like this issue can be closed now.

grahamc:
Thanks!

YorikSar:
It seems fix for this issue triggered #4169.

PR #3608, opened 2020-05-22
Use /etc/zshenv instead of /etc/zshrc for profile

surajbarkale:
As noted in https://github.com/NixOS/nix/issues/3456 the `/etc/zshenv` file provides a better place for sourcing the nix environment.


domenkozar:
https://stackoverflow.com/questions/21038903/path-variable-in-zshenv-or-zshrc suggests this won't work on macOS.

surajbarkale:
That thread is about `$HOME/.zshenv` file. I have been using `/etc/zshenv` since start of this year without any issues.

PR #3626, opened 2020-05-25
Make `functionArgs` primitive accept primops (fix #3624)

### LINKED ISSUE #3624: `functionArgs` does not work with `primop` or `primop-app`
**Describe the bug**
`functionArgs` is not considering primps or primops applications as functions, and throws an uncatchable error.

**Steps To Reproduce**
For instance, `lib.functionArgs map` reduces to an error.

**Expected behavior**
`functionArgs` should return `{}`

`nix-env (Nix) 2.3.4`

**Additional context**
I guess this is just about adding an if statement there to handle primops:
https://github.com/NixOS/nix/blob/25ed842725864f1fa9330d3c2ea6f741c63fca07/src/libexpr/primops.cc#L1346-L1350

I'll post a PR, then.

W95Psp:
Hi!

I ran into the bug I documented in #3624, that is `functionArgs` not accepting `primop`s or partial applications or `primop`s.
This PR just makes `functionArgs` returns `{}` when a primitive operation is given.

It compiles properly for me, and as expected, evaluating `(import <nixpkgs>).lib.functionArgs map` indeed yields `{}`.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1346,6 +1346,10 @@ static void prim_catAttrs(EvalState & state, const Pos & pos, Value * * args, Va
 static void prim_functionArgs(EvalState & state, const Pos & pos, Value * * args, Value & v)
 {
     state.forceValue(*args[0], pos);
+    if (args[0]->type == tPrimOpApp || args[0]->type == tPrimOp) {
+        state.mkAttrs(v, 0);
+        return;
+    }
     if (args[0]->type != tLambda)
         throw TypeError(format("'functionArgs' requires a function, at %1%") % pos);
 


puckipedia:
This is not correct behavior, as there are primops (like `builtins.fetchGit`) that take an attrset of a certain shape, which can't be verified with this.

W95Psp:
Oh, right, I did not think about such primops.
However, it seems that most of primops that accept attrset as one of their formal argumentrs (from what I saw, this includes `prim_fetchMercurial`, `prim_fetchTree`, `prim_fetchurl`, `prim_fetchTarball`, `prim_fetchGit`, and others) are polymorphic and accepts other types (such as strings). Such a "polymorphic" function `f` in nix would looks like `x: if builtins.typeOf x = "set" then ... else ...`, and then `functionArgs f` yields `{}`.

Therefore, what should be the correct behavior for such "polymorphic" primops?

I found `prim_genericClosure` that takes an attrset and nothing else, I guess there's a couple more. But then, what is the good strategy? Decide such primops are rare and add manually cases in `functionArgs` would be rather ugly. Another option would I guess be to change `addPrimOp` so that instead of simply taking an arity, it takes a list of maybe attr names (say `list (maybe (list name))`), and reflect that change in `PrimOp` struct.

This would allow `functionArgs` to have the proper behaviour for both primops and primops partial applications.

Not too sure it's a good idea though.

puckipedia:
oh huh, i was not aware you could `builtins.fetch*` strings. huh

Infinisil:
I think this change is correct and should be merged because:
- The semantics of `builtins.functionArgs` are that any arguments it returns *have* to be accepted by the function, but that doesn't mean that other arguments can't be accepted. So returning `{}` for primops is an entirely consistent and safe behavior
- In fact, primops are implemented very much like `args: ... args.foo ...` Nix functions, taking attributes from `args` whenever they need. And `builtins.functionArgs` for such Nix functions also obviously return `{}`
- In addition, the current behavior of `builtins.isFunction` returning `true` for primops, yet `builtins.functionArgs` failing is very inconsistent

Ericson2314:
All this stuff is wack, but @Infinisil does make sense. And if `functionArgs` is a conservative approximation, I guess we can give it more info about primops later?

edolstra:
Thanks, makes sense!

PR #3629, opened 2020-05-26
ci: upload binaryTarball as an artifact

domenkozar:


domenkozar:
Sigh - seems like github will compress the file and there's no way to turn it off

PR #3632, opened 2020-05-27
installer: don't require xz on darwin

LnL7:
On macOS the system tar has builtin support for lzma while xz isn't
available as a separate binary.  There's no builtin package manager
there available either so having to install lzma (without nix) would be
rather painful.

Alternative to adding this condition would be to revert #3431, in which case potentially more works out of the box but with the downside of worse error handling.

/cc @domenkozar

### DIFF
--- a/scripts/install.in
+++ b/scripts/install.in
@@ -36,7 +36,9 @@ tarball="$tmpDir/$(basename "$tmpDir/nix-@nixVersion@-$system.tar.xz")"
 
 require_util curl "download the binary tarball"
 require_util tar "unpack the binary tarball"
-require_util xz "unpack the binary tarball"
+if [ "$(uname -s)" != "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


domenkozar:
Weird, I tested it on github actions, but it seems like that has `xz` installed.

LnL7:
We didn't check this part, unless you also built and tested the full installer script (pre tarball).

domenkozar:
The code path in the PR is part of nixos.org/nix/install/ script, so I'm not sure what you mean.

Mic92:
@domenkozar I also came across this. This does not make a good first impression. Could we have a patch release for this? Otherwise it works perfect on Catalina!

domenkozar:
@Mic92 it seems like Eelco already prepared 2.3.6 release so it should be done soon :)

PR #3635, opened 2020-05-28
Git hashing as a new file ingestion method --- contains #3754

Ericson2314:
This is an umbrella PR for all the work we are doing towards git object support in Nix, as part of https://discourse.nixos.org/t/obsidian-systems-is-excited-to-bring-ipfs-support-to-nix . We will probably break up into smaller PRs as it nears completion. More details forthcoming.

-----------------

CC @nlewo

depends on #3754


grahamc:
Interesting! I wonder if we could teach IPFS about NARs, in the same way that IPFS knows about Git?

Ericson2314:
@grahamc We could. But there are reason to want to try git objects even if it is not the only way to talk to IPFS:
 - "native" format for much source. If we want to collaborate with other distros, archive.org, software heritage foundation, etc., I think it's nice to keep data in its original form.
 - deterministic `fetchgit` with just one hash.
 - low latency CI (thanks @Profpatsch for the link!) https://docs.google.com/presentation/d/14dxac2omYI5Feaoiw-u09qB1fQgJU7ASNJag7YMg3TI/edit#slide=id.g61b36395ff_1_0 (thanks @Profpatsch!)

We'll write an RFC that lays out a fully-fleshed motivation. This PR is supposed to be the sneak preview / if anyone has any specific implementation concerns we ought to know about sooner rather than later.

zimbatm:
@Ericson2314 are there limitations around large files similar to Git? Is it possible to cut them up in smaller pieces in a [casync](https://github.com/systemd/casync) fashion?

Ericson2314:
@zimbatm git works with arbitrary sized files, but poorly. I went with it because it is popular, especially among the data we have as input (git repos, of course)---it's important for network effects that the addresses themselves, not merely the data, be well known.

I think that's a fine choice for now, and in a world where these p2p technologies are much more widely used we'll have ample opportunity to change it. https://github.com/NixOS/nix/issues/3640 for example doesn't say we should put git and nar on equal footing, but anything that might go in the `ca` field. Git itself will need to migrate off sha-1, and while there is a plan already (https://github.com/git/git/blob/master/Documentation/technical/hash-function-transition.txt), perhaps there is still time to raise large file / chunking issues.

Profpatsch:
This branch has 99 commits now.

I wonder how you ever intend it to be reviewed and merged?

Ericson2314:
Well, I don't just yet. Most of the commits are merge comments and many of the other ones have already been merged, so the actual diff is shrinking even as the commit count is growing.

At some point if all dependent PRs are merged we might rebase it.

balodja:
So much useful work just lays here in dust :)

Is there any chance at least part of that code to be merged?

Ericson2314:
I haven't forgotten this stuff and plan on coming back to. At the very least, the implementation of now-accepted https://github.com/NixOS/rfcs/pull/92/files shares some early refactor/cleanup commits with this, and I hope those can be kept. For further steps see what I wrote in https://github.com/NixOS/nix/issues/859#issuecomment-970415183

That said, before I do any of this I want to improve the testing in Nix --- see some of my more recent open PRs. We've seen some big regressions in 2.4 that need to be fixed and have taken quite some time, and I really want to make sure that there will never be as difficult version upgrade again. My putting in that testing first I make sure that all the new things I have in mind for Nix --- and refactors that go with them -- won't just destabilize the codebase further.

Ericson2314:
Closing in favor of https://github.com/NixOS/nix/pull/8918. What basically happened is that think the other order, Git hashing first, then content-addresed store lookup, makes more sense. So if that gets merged this, this will become the same as https://github.com/NixOS/nix/pull/3754

PR #3639, opened 2020-05-29
Remove `addToStore` variant as requested by `FIXME`

Ericson2314:
The idea is it's always more flexible to consumer a `Source` than a
plain string, and it might even reduce memory consumption.

I also looked at `addToStoreFromDump` with its `// FIXME: remove?`, but
the worked needed for that is far more up for interpretation, so I
punted for now.


Ericson2314:
I did a `git blame`, and found the fixme is written by @edolstra in https://github.com/NixOS/nix/commit/48662d151bdf4a38670897beacea9d1bd750376a. I hope this means it is utterly uncontroversial :).

Issue #3612, opened 2020-05-23
flakes: incorrectly rejects valid Git branch/tag names


emilazy:
**Describe the bug**

Nix on the `flakes` branch incorrectly rejects Git ref names with `/` in them.

**Steps to reproduce**

```
emily@renko ~> nix build --no-link 'github:NixOS/nixpkgs?ref=pull/88692/head'#pkgsMusl.libusb1
error: GitHub URL 'github:NixOS/nixpkgs?ref=pull/88692/head#pkgsMusl.libusb1' contains an invalid branch/tag name
```

**Expected behaviour**

Nix should accept everything the GitHub API/`git fetch` does; https://api.github.com/repos/NixOS/nixpkgs/commits/pull/88692/head works, so this should too.

bqv:
It's fairly obvious why this bug came to be, is the intention to fix it or leave it as a known limitation? As I see it, solving it may end up overcomplicating some grammars but it is slightly unacceptable.

jbaum98:
I think there's been a regression here:
```sh
$ ./result/bin/nix build --no-link 'github:NixOS/nixpkgs?ref=pull/88692/head'#pkgsMusl.libusb1
error: --- BadURL ------------------------------------------------ nix
URL 'github:NixOS/nixpkgs?ref=pull/88692/head#pkgsMusl.libusb1' contains an invalid branch/tag name

$ ./result/bin/nix --version
nix (Nix) 3.0pre20200806_6146447
```

PR #3642, opened 2020-05-30
Improve ref validity checking in fetchgit

knl:
The previous regex was too strict and did not match what git was allowing. It
could lead to `fetchGit` not accepting valid branch names, even though they
exist in a repository (for example, branch names containing `/`, which are
pretty standard, like `release/1.0` branches).

The new regex defines what a branch name should **NOT** contain. It takes the
definitions from `refs.c` in https://github.com/git/git and `git help
check-ref-format` pages.

This change also introduces a test for ref name validity checking, which
compares the result from Nix with the result of `git check-ref-format --branch`.

Should fix https://github.com/NixOS/nix/issues/3612


edolstra:
Thanks!

PR #3645, opened 2020-06-01
Cache tree in fetchOrSubstituteTree

mkenigs:
Caches tree in addition to lockedRef, and explicitly writes out the logic for different combinations of cached/uncached flakes and indirect/resolved/locked flakes. This eliminates uneccessary calls to lookupInFlakeCache, fetchTree, maybeLookupFlake, and flakeCache.push_back


edolstra:
Thanks!

PR #3649, opened 2020-06-02
ValidPathInfo: make ca field a proper datatype

meditans:


Ericson2314:
Oh, I thought I could edit this because it is from a repo I have access too, but I guess that is not the case. This contains a number of other PRs:

 - #3434
 - #3450
 - #3439
 - #3453

And the culmination of our work to tighten up the API replacing strings with proper data type. We're doing this to help root out bugs with the git objects work, but we hope it can be merged separately as #3455 as it cleans up the code regardless of whether git objects ever makes it in.

Ericson2314:
This is still WIP because we need to actually write the `ca` field parser, but the more important thing is the C++ datatypes so I figured it was worth getting the WIP PR up.

src/nix/verify.cc
```diff
@@ -87,7 +87,7 @@ struct CmdVerify : StorePathsCommand
                 if (!noContents) {
 
                     std::unique_ptr<AbstractHashSink> hashSink;
-                    if (info->ca == "")
+                    if (info->ca)
```
matthewbauer, line 90: I think the logic should flip here. `info->ca == ""` should now be `!info->ca`

src/nix-store/nix-store.cc
```diff
@@ -940,11 +940,15 @@ static void opServe(Strings opFlags, Strings opArgs)
...
-                info.narHash = Hash(readString(in), htSHA256);
+                info.narHash = Hash(readString(in), HashType::SHA256);
                 info.references = readStorePaths<StorePathSet>(*store, in);
                 in >> info.registrationTime >> info.narSize >> info.ultimate;
                 info.sigs = readStrings<StringSet>(in);
-                in >> info.ca;
+                {
+                    std::string rawCA;
+                    in >> rawCA;
+                    info.ca = parseContentAddressOpt(rawCA);
```
matthewbauer, line 950:
```suggestion
                    info.ca = parseContentAddressOpt(readString(in));
```

src/libutil/rust-ffi.hh
```diff
@@ -171,7 +197,7 @@ struct Result
     }
 
     /* Rethrow the wrapped exception or return the wrapped value. */
-    T unwrap()
+    T unwrap() &&
```
matthewbauer, line 200: what does `&&` do here?

src/libstore/store-api.cc
```diff
@@ -778,24 +760,23 @@ bool ValidPathInfo::isContentAddressed(const Store & store) const
         printError("warning: path '%s' claims to be content-addressed but isn't", store.printStorePath(path));
     };
 
-    if (hasPrefix(ca, "text:")) {
-        Hash hash(std::string(ca, 5));
-        if (store.makeTextPath(path.name(), hash, references) == path)
+    if (! ca) {}
```
matthewbauer, line 763: this should jut return false here to make that explicit

src/libstore/store-api.cc
```diff
@@ -470,8 +467,8 @@ void Store::pathInfoToJSON(JSONPlaceholder & jsonOut, const StorePathSet & store
                     jsonRefs.elem(printStorePath(ref));
             }
 
-            if (info->ca != "")
-                jsonPath.attr("ca", info->ca);
+            if (info->ca)
+                jsonPath.attr("ca", renderContentAddress(info->ca));
```
matthewbauer, line 471: I think this will automatically happen if you create a to_string() instance for ContentAddress

src/libstore/remote-store.cc
```diff
@@ -375,13 +375,15 @@ void RemoteStore::queryPathInfoUncached(const StorePath & path,
...
+            info->narHash = Hash(readString(conn->from), HashType::SHA256);
             info->references = readStorePaths<StorePathSet>(*this, conn->from);
             conn->from >> info->registrationTime >> info->narSize;
             if (GET_PROTOCOL_MINOR(conn->daemonVersion) >= 16) {
                 conn->from >> info->ultimate;
                 info->sigs = readStrings<StringSet>(conn->from);
-                conn->from >> info->ca;
+                string caOptRaw;
+                conn->from >> caOptRaw;
+                info->ca = parseContentAddressOpt(caOptRaw);
```
matthewbauer, line 386:
```suggestion
                info->ca = parseContentAddressOpt(readString(conn->from));
```

src/libstore/local-store.cc
```diff
@@ -997,18 +999,16 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
 
             deletePath(realPath);
 
-            if (info.ca != "" &&
-                !((hasPrefix(info.ca, "text:") && !info.references.count(info.path))
-                    || info.references.empty()))
+            if (info.ca && !info.references.empty() && !std::holds_alternative<TextHash>(*info.ca))
```
matthewbauer, line 1002: I think this is missing the `&& !info.references.count(info.path)` from the original condition when it's a text hash. That is, text hashes are allow but only when it has no self references.

src/libstore/legacy-ssh-store.cc
```diff
@@ -114,7 +114,11 @@ struct LegacySSHStore : public Store
             if (GET_PROTOCOL_MINOR(conn->remoteVersion) >= 4) {
                 auto s = readString(conn->from);
                 info->narHash = s.empty() ? Hash() : Hash(s);
-                conn->from >> info->ca;
+                {
+                    std::string rawCaOpt;
+                    conn->from >> rawCaOpt;
+                    info->ca = parseContentAddressOpt(rawCaOpt);
```
matthewbauer, line 120:
```suggestion
                    info->ca = parseContentAddressOpt(readString(conn->from));
```

src/libstore/file-hash.cc
```diff
@@ -0,0 +1,53 @@
+#include "file-hash.hh"
```
matthewbauer, line 1: might want to call this "content-address.cc/.hh" since that's what it mostly provides. otherwise it's a bit confusing what the difference is between this and all of the other hash functions in hash.cc.

src/libstore/derivations.hh
```diff
@@ -15,13 +16,14 @@ namespace nix {
 struct DerivationOutput
 {
     StorePath path;
-    std::string hashAlgo; /* hash used for expected hash computation */
-    std::string hash; /* expected hash, may be null */
-    DerivationOutput(StorePath && path, std::string && hashAlgo, std::string && hash)
+    std::optional<FileSystemHash> hash; /* hash used for expected hash computation */
```
matthewbauer, line 19: Does this make sense here?

matthewbauer: That is, a derivation might not have any file ingestion at all. If it's not fixed output, I think it will only have file ingestion through its inputs.

matthewbauer: I would recommend just using the `Hash` type here.

src/libstore/file-hash.hh
```diff
@@ -0,0 +1,66 @@
...
+struct TextHash {
+    Hash hash;
+    TextHash(const TextHash &) = default;
+    TextHash(TextHash &&) = default;
+    TextHash & operator = (const TextHash &) = default;
+};
+
+/// Pair of a hash, and how the file system was ingested
+struct FileSystemHash {
+    FileIngestionMethod method;
```
matthewbauer, line 22: this might need to be optional

src/libstore/build.cc
```diff
@@ -1457,6 +1463,20 @@ void DerivationGoal::tryToBuild()
        supported for local builds. */
     bool buildLocally = buildMode != bmNormal || parsedDrv->willBuildLocally();
 
+    auto started = [&]() {
```
matthewbauer, line 1466: Nice! This is definitely useful

src/libstore/derivations.hh
Ericson2314: Very confusingly, the `hashAlgo` string also contained the `r:`. See https://github.com/NixOS/nix/pull/3439, where this comes from.

src/nix-store/nix-store.cc
Ericson2314: GitHub has multi-line suggestions now, can you use those @matthewbauer?

src/nix/verify.cc
Ericson2314: Good catch
```suggestion
                    if (!info->ca)
```

src/libstore/file-hash.hh
Ericson2314: I don't think so?

src/libutil/rust-ffi.hh
Ericson2314: From earlier PR too, it makes something like taking `this` by rvalue reference

src/nix-store/nix-store.cc
```diff
@@ -940,11 +940,15 @@ static void opServe(Strings opFlags, Strings opArgs)
...
+                }
```
matthewbauer, line 951:
```suggestion:-5+1
                info.ca = parseContentAddressOpt(readString(in));
```

src/nix-store/nix-store.cc
```diff
@@ -940,11 +940,15 @@ static void opServe(Strings opFlags, Strings opArgs)
                 auto deriver = readString(in);
                 if (deriver != "")
                     info.deriver = store->parseStorePath(deriver);
-                info.narHash = Hash(readString(in), htSHA256);
+                info.narHash = Hash(readString(in), HashType::SHA256);
                 info.references = readStorePaths<StorePathSet>(*store, in);
                 in >> info.registrationTime >> info.narSize >> info.ultimate;
                 info.sigs = readStrings<StringSet>(in);
-                in >> info.ca;
+                {
```
matthewbauer, line 947:
```suggestion
```

src/nix-store/nix-store.cc
```diff
@@ -940,11 +940,15 @@ static void opServe(Strings opFlags, Strings opArgs)
...
+                    std::string rawCA;
```
matthewbauer, line 948:
```suggestion
```

src/nix-store/nix-store.cc
```diff
@@ -940,11 +940,15 @@ static void opServe(Strings opFlags, Strings opArgs)
...
+                    in >> rawCA;
```
matthewbauer, line 949:
```suggestion
```

src/nix-store/nix-store.cc
matthewbauer, line 951:
```suggestion
```

src/libstore/remote-store.cc
```diff
@@ -375,13 +375,15 @@ void RemoteStore::queryPathInfoUncached(const StorePath & path,
...
             if (deriver != "") info->deriver = parseStorePath(deriver);
-            info->narHash = Hash(readString(conn->from), htSHA256);
+            info->narHash = Hash(readString(conn->from), HashType::SHA256);
             info->references = readStorePaths<StorePathSet>(*this, conn->from);
             conn->from >> info->registrationTime >> info->narSize;
             if (GET_PROTOCOL_MINOR(conn->daemonVersion) >= 16) {
                 conn->from >> info->ultimate;
                 info->sigs = readStrings<StringSet>(conn->from);
-                conn->from >> info->ca;
+                string caOptRaw;
```
matthewbauer, line 384:
```suggestion
```

src/libstore/remote-store.cc
```diff
@@ -375,13 +375,15 @@ void RemoteStore::queryPathInfoUncached(const StorePath & path,
...
+                conn->from >> caOptRaw;
```
matthewbauer, line 385:
```suggestion
```

src/libstore/legacy-ssh-store.cc
```diff
@@ -114,7 +114,11 @@ struct LegacySSHStore : public Store
             if (GET_PROTOCOL_MINOR(conn->remoteVersion) >= 4) {
                 auto s = readString(conn->from);
                 info->narHash = s.empty() ? Hash() : Hash(s);
-                conn->from >> info->ca;
+                {
```
matthewbauer, line 117:
```suggestion
```

src/libstore/legacy-ssh-store.cc
```diff
@@ -114,7 +114,11 @@ struct LegacySSHStore : public Store
...
+                    std::string rawCaOpt;
```
matthewbauer, line 118:
```suggestion
```

src/libstore/legacy-ssh-store.cc
```diff
@@ -114,7 +114,11 @@ struct LegacySSHStore : public Store
...
+                    conn->from >> rawCaOpt;
+                    info->ca = parseContentAddressOpt(rawCaOpt);
+                }
```
matthewbauer, line 121:
```suggestion
```

src/libstore/legacy-ssh-store.cc
```diff
@@ -114,7 +114,11 @@ struct LegacySSHStore : public Store
             if (GET_PROTOCOL_MINOR(conn->remoteVersion) >= 4) {
                 auto s = readString(conn->from);
                 info->narHash = s.empty() ? Hash() : Hash(s);
-                conn->from >> info->ca;
+                {
+                    std::string rawCaOpt;
+                    conn->from >> rawCaOpt;
```
matthewbauer, line 119:
```suggestion
```

src/libstore/file-hash.hh
matthewbauer: Mainly because of the above comment. Not all uses of FileSystemHash will have a real "FileIngestionMethod" defined. Consider derivations.

src/libutil/rust-ffi.hh
Ericson2314: The point is that since we are moving `data` out, the original is consumed/invalidated. Making it take it by rvalue-refrence means the original can't be used anyways (it was an rvalue or `std::move`d), which implements this idea and roughly matches the Rust semantics.

src/libstore/local-store.cc
Ericson2314: Good catch
```suggestion
            bool refersToSelf = info.references.count(info.path) > 0;
            if (info.ca && !info.references.empty() && !(std::holds_alternative<TextHash>(*info.ca) && !refersToSelf))
```

src/libstore/legacy-ssh-store.cc
matthewbauer, line 121:
```suggestion
                info->ca = parseContentAddressOpt(readString(conn->from));
```

src/nix-store/nix-store.cc
matthewbauer, line 951:
```suggestion
                info.ca = parseContentAddressOpt(readString(in));
```

src/libstore/remote-store.cc
matthewbauer, line 386:
```suggestion
                info->ca = parseContentAddressOpt(readString(conn->from));
```

src/libstore/build.cc
Ericson2314: I actually don't know where this came from!

src/libstore/store-api.cc
Ericson2314: I cleaned this one up a good bit :)

src/libstore/file-hash.cc
meditans: Moved, thanks

src/libstore/local-store.cc
```diff
@@ -1253,20 +1255,20 @@ bool LocalStore::verifyStore(bool checkContents, RepairFlag repair)
...
                 auto info = std::const_pointer_cast<ValidPathInfo>(std::shared_ptr<const ValidPathInfo>(queryPathInfo(i)));
 
                 /* Check the content hash (optionally - slow). */
-                printMsg(lvlTalkative, "checking contents of '%s'", printStorePath(i));
+                printMsg(Verbosity::Talkative, "checking contents of '%s'", printStorePath(i));
 
                 std::unique_ptr<AbstractHashSink> hashSink;
-                if (info->ca == "" || !info->references.count(info->path))
-                    hashSink = std::make_unique<HashSink>(info->narHash.type);
+                if (info->ca || !info->references.count(info->path))
```
matthewbauer, line 1268: should be `!info->ca`

src/libstore/local-store.cc
```diff
@@ -997,18 +999,18 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
...
+            bool refersToSelf = info.references.count(info.path) > 0;
+            if (info.ca && !info.references.empty() && !(std::holds_alternative<TextHash>(*info.ca) && !refersToSelf))
                 settings.requireExperimentalFeature("ca-references");
 
             /* While restoring the path from the NAR, compute the hash
                of the NAR. */
             std::unique_ptr<AbstractHashSink> hashSink;
-            if (info.ca == "" || !info.references.count(info.path))
-                hashSink = std::make_unique<HashSink>(htSHA256);
+            if (info.ca || !info.references.count(info.path))
```
matthewbauer, line 1010: should be `!info.ca`

meditans: That's fixed now, thanks!

src/libstore/local-store.cc
meditans: That's fixed now, thanks!

src/libstore/file-hash.hh
Ericson2314: Oh, we should rename the data type then. This is strictly for content-addressed things.

Ericson2314:
OK this now just contains #3439 and #3650. The PRs it doesn't contain were precisely the ones @edolstra found dodgy or unnecessary, so :crossed_fingers: this is now good.

src/libstore/content-address.hh
```diff
@@ -0,0 +1,56 @@
...
+    Flat = false,
+    Recursive = true
+};
+
+struct TextHash {
+    Hash hash;
+};
+
+/// Pair of a hash, and how the file system was ingested
+struct FixedOutputHash {
```
Ericson2314, line 18: I went with `FixedOutputHash` for this PR, now that it is not just used in derivations.

PR #3650, opened 2020-06-02
Remove `HashType::htUnknown`

Ericson2314:
Instead, `Hash` uses `std::optional<HashType>`. In the future, we may
also make `Hash` itself require a known hash type, encouraging people to
use `std::optional<Hash>` instead.

~~#3453 makes most of the noise.~~


Ericson2314:
Whew! got it working.

src/libutil/hash.cc
```diff
@@ -202,14 +214,16 @@ Hash::Hash(std::string_view s, HashType type)
...
 }
 
-Hash newHashAllowEmpty(std::string hashStr, HashType ht)
+Hash newHashAllowEmpty(std::string hashStr, std::optional<HashType> ht)
 {
     if (hashStr.empty()) {
-        Hash h(ht);
-        warn("found empty hash, assuming '%s'", h.to_string(SRI, true));
+        if (!ht)
+            throw BadHash("empty hash requires explicit hash type");
```
Ericson2314, line 224: This is a new validation, actually. I think we want it.

CC @matthewbauer

Ericson2314:
@edolstra OK this not longer contains the `enum struct` change, and so much more value per diff.

CC @zimbatm

src/libutil/hash.cc
matthewbauer: it's okay if hash type is unset - you just have to have sri in your string

matthewbauer:
How is std::optional better than HashType::Unknown? They seem like two ways to represent the same thing to me.

Ericson2314:
@matthewbauer Well the `std::optional` isn't used on every hash type. I am purposefully being cautious in this PR, but with the idea that if this is OK it becomes much easier and more incremental to start chipping away at the rest of the `std::optional`. In particular I mention in the OP about removing it from the definition of `Hash` and using `std::optional<Hash>` instead.

src/libutil/hash.cc
Ericson2314: Yeah this is just in the empty string case, when the SRI or legacy prefix can't help.

PR #3653, opened 2020-06-02
Error on more hash mismatch cases

matthewbauer:
instead of asserting, show the user what was expected vs. what was
handled.

This condition is very close to the one right after it, but this one
happens when the tree narHash is not equal to input narHash.

### DIFF
--- a/src/libfetchers/fetchers.cc
+++ b/src/libfetchers/fetchers.cc
@@ -62,8 +62,9 @@ std::pair<Tree, std::shared_ptr<const Input>> Input::fetchTree(ref<Store> store)
     if (!tree.info.narHash)
         tree.info.narHash = store->queryPathInfo(tree.storePath)->narHash;
 
-    if (input->narHash)
-        assert(input->narHash == tree.info.narHash);
+    if (input->narHash && input->narHash != tree.info.narHash)
+        throw Error("NAR hash mismatch in input '%s' (%s), expected '%s', got '%s'",
+            to_string(), tree.actualPath, input->narHash->to_string(SRI), tree.info.narHash.to_string(SRI));
 
     if (narHash && narHash != input->narHash)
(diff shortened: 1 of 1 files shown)


src/libfetchers/fetchers.cc
```diff
@@ -62,8 +62,9 @@ std::pair<Tree, std::shared_ptr<const Input>> Input::fetchTree(ref<Store> store)
     if (!tree.info.narHash)
         tree.info.narHash = store->queryPathInfo(tree.storePath)->narHash;
 
-    if (input->narHash)
-        assert(input->narHash == tree.info.narHash);
+    if (input->narHash && input->narHash != tree.info.narHash)
+        throw Error("NAR hash mismatch in input '%s' (%s), expected '%s', got '%s'",
+            to_string(), tree.actualPath, input->narHash->to_string(SRI), tree.info.narHash.to_string(SRI));
```
Ericson2314, line 67: Maybe this should go in `fetchTreeInternal`?

matthewbauer: narHash is in a valid attr for all inputs, so I think all of them need to handle this case. it's replacing an assert that was there previously.

matthewbauer:
Example of what happens without this:

```
builtins.fetchTree {
  type = "git";
  url = "https://github.com/nixos/nix.git";
  rev = "bfa1acd85c4d15c5ea95337138f47672659e2a9e";
  narHash = "sha256-0000000000000000000000000000000000000000000=";
}
```

Ericson2314:
@edolstra I am not sure what to make of PR this, as I am stuck trying to understand why `fetchTree{,Internal}` returns another input. That seems dubious to me, and changing that would avoid the issue @matthewbauer found.

Ericson2314:
@edolstra nevermind, I think this is just a good change, and my "why two inputs" questions are orthogonal.

matthewbauer:
This is fixed on flakes branch.

PR #3655, opened 2020-06-03
libutils/hash: remove default encoding

zimbatm:
This is the first step in a series of patches.

The idea, in the end, is that it will be possible to display the hash using the original encoding. This is to solve the issue where "expect <hash> got <hash>" is printed in a different encoding than what is in the source code. This makes it harder to scan the source to replace the hash.

1. Remove all the defaults (this PR). This is done separately so that it can be merged in the flakes branch and all the merge conflicts resolved there.
2. Rename Base to HashEncoding. "Base" doesn't make sense anymore since SRI. Maybe introduce "PrefixedBase16" and "PrefixedBase32" to remove the (encoding, includeType) tuple since it doesn't make sense with SRI either?
3. Change the parser to store the original encoding. Restore the `hash.to_string()` function but that prints using the original encoding. Remove the encoding from places where we want to show the original hash.

/cc @Ericson2314 @matthewbauer who have conflicting PRs open


Ericson2314:
For step 2, I would rather have https://github.com/NixOS/nix/pull/3453 merged first, as I think that opens the door to more renaming choices, but I still like the idea (and am fine about breakage with any other PR @matthewbauer and I have open, we can fix those).

For step 3...part of me just doesn't like storing extra information that is only used for diagnostics, but I certainly am sympathetic to the use case. Maybe we could have separate a `Hash` and `HashWithEncoding` to just hit the points where it matters. (e.g. drv files all use the same encoding.)

gilligan:
```c++
to_string(Base32, false);
to_string(SRI, true);
```
I lack the insight to make any comments on broader design choices here but I think it would be great if if `to_string` could be changed to avoid the boolean blindness and instead use something like:

```c++
to_string(Base32, HashFormat::Plain);
to_string(SRI, HashFormat::TypePrefix);
```

zimbatm:
With the introduction of SRI,  (SRI, false) doesn't make any sense. That's why I am proposing to change the API to be `to_string(HashEncoding)` and give each variant a name. (this is point 2 of the plan)

gilligan:
@zimbatm okay cool. I don’t know what SRI stands for by the way 😅

zimbatm:
It stands for Subresource Integrity and is a standard from the w3c: https://www.w3.org/TR/SRI/
The idea was to re-use an existing standard instead of inventing our own.

PR #3656, opened 2020-06-03
Add error message when FileIngestionMethod is out of bounds

matthewbauer:
bool coerces anything >0 to true, but in the future we may have other
file ingestion methods. This shows a better error message when the
“recursive” byte isn’t 1.


src/libstore/daemon.cc
```diff
@@ -358,8 +358,10 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         std::string s, baseName;
         FileIngestionMethod method;
         {
-            bool fixed, recursive;
+            bool fixed; uint8_t recursive;
             from >> baseName >> fixed /* obsolete */ >> recursive >> s;
+            if (recursive > (uint8_t) FileIngestionMethod::Recursive)
+                throw Error(format("Unsupported FileIngestionMethod with value of %i; you may need to upgrade nix-daemon.") % recursive);
```
edolstra, line 364: `format` is deprecated, you can do `throw Error("unsupported ... %i ...", recursive)`. Also for consistency, error messages are not capitalized and don't end in a period.

PR #3657, opened 2020-06-03
Clarify the description of StorePath inputs

meditans:


Ericson2314:
Does this look correct / worth documenting?

Issue #3658, opened 2020-06-04
Can't autoreconf when building from source


darlo-me:
**Describe the bug**

When running ./bootstrap.sh (or autoreconf -f), I get this error message:
```
autoreconf-2.69: Entering directory `.'
autoreconf-2.69: configure.ac: not using Gettext
autoreconf-2.69: running: aclocal --force 
autoreconf-2.69: configure.ac: tracing
autoreconf-2.69: configure.ac: not using Libtool
autoreconf-2.69: running: /usr/bin/autoconf-2.69 --force
configure.ac:118: error: possibly undefined macro: AC_MSG_ERROR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf-2.69: /usr/bin/autoconf-2.69 failed with exit status: 1
```

It seems installing autoconf-archive fixes the issue

**Steps To Reproduce**

> curl https://nixos.org/releases/nix/nix-2.3.6/nix-2.3.6.tar.xz > nix-2.3.6.tar.xz
> tar -xvf nix-2.3.6.tar.xz
> cd nix-2.3.6
> ./bootstrap.sh

**Expected behavior**

Clearer error (install autoconf-archive) or documentation (mention it in the manual)

**Additional context**

I encountered this issue while trying to install the gentoo ebuild (https://github.com/trofi/nix-guix-gentoo)

Similar bug mentioned here: https://github.com/NixOS/nix/pull/3154#issuecomment-545207139/

trofi:
Looking at `aclocal` run I think it means release tarball lacks 3 `m4` files:
```
$ aclocal -I m4/ --install
aclocal-1.16: installing 'm4//ax_boost_base.m4' from '/usr/share/aclocal/ax_boost_base.m4'
aclocal-1.16: installing 'm4//ax_require_defined.m4' from '/usr/share/aclocal/ax_require_defined.m4'
aclocal-1.16: installing 'm4//pkg.m4' from '/usr/share/aclocal/pkg.m4'
```

edolstra:
`./bootstrap.sh` is only for building from a Git tree. When building from a tarball, `./configure` should be sufficient. BTW we'll stop doing tarball releases for Nix >= 2.4.

trofi:
Sometimes `configure.ac` has to be patched downstream. To regenerate `configure` from patched `configure.ac` all `.m4` files has to already present in tarball. Be it used by `./bootstrap.sh` or by manual `autoconf` run.

[...]

PR #3659, opened 2020-06-04
bootstrap.sh: always copy external .m4 files in

trofi:
`autoreconf -vif` does not copy in new `.m4` files by default.
The change adds `aclocal --install` call and adds missing files
for pkg-config and boost macros.

Reported-by: Daryl Anthony Chouinard
Bug: https://github.com/NixOS/nix/issues/3658


trofi:
Closing as obsolete as there are no plans to provide working tarballs: #3658

PR #3664, opened 2020-06-04
Add `src/libutil/tests/libutil-tests` to `.gitignore`

Ericson2314:
I gather this comes from the new unit tests.


gilligan:
Yes indeed it does. Sorry, didn’t think of adding it to .gitignore

Ericson2314:
Eh, it's well worth having unit tests at all! :)

PR #3666, opened 2020-06-05
add progress bar with multiple output

Kloenk:
This adds a progress bar which expands down to show all current build processes and the last line of the output.

### Example
[asciicast](https://asciinema.org/a/j6LpB39aFc7WVlhT4HZnsOOuB)

Initial Idea from @petabyteboy

### DIFF
--- a/src/nix/progress-bar.cc
+++ b/src/nix/progress-bar.cc
@@ -3,6 +3,7 @@
 #include "sync.hh"
 #include "store-api.hh"
 #include "names.hh"
+#include "globals.hh"
 
 #include <atomic>
 #include <map>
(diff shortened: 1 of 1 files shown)


gleber:
A better recording: https://asciinema.org/a/336924

I think this is an awesome UI feature! I am not sure thought that it should become the default.

@edolstra What do you think?

Kloenk:
@gleber I put it behind the experimental feature `progress-bar`. So it's not the default, but you can activate it.

davidak:
Where is the progress bar? :grimacing: 

I would expect something like this and that would be great to see the overall progress of the command.

https://asciinema.org/a/348804?t=58

PR #3670, opened 2020-06-08
Add tests for pool.hh

gilligan:
Adds tests for
- constructing a pool
- increasing/decreasing capacity
- flushing resources
- reuse of resources

### DIFF
--- a/src/libutil/tests/pool.cc
+++ b/src/libutil/tests/pool.cc
@@ -1,0 +1,127 @@
+#include "pool.hh"
+#include <gtest/gtest.h>
+
+namespace nix {
+
+    struct TestResource
+    {
+
+        TestResource() {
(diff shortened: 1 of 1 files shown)


src/libutil/tests/pool.cc
```diff
@@ -0,0 +1,127 @@
...
+
+        TestResource r = *(pool.get());
+
+        ASSERT_EQ(pool.count(), 1);
+        ASSERT_EQ(pool.capacity(), 1);
+        ASSERT_EQ(r.dummyValue, 1);
+        ASSERT_EQ(r.good, true);
+    }
+
+    TEST(Pool, capcaityCanBeIncremented) {
```
edolstra, line 48: "capcaity" -> "capacity"

Issue #3622, opened 2020-05-25
Extend Unit Tests


gilligan:
## Nix Unit Testing
Tooling for Unit Testing has been added via #3571 - Unit Tests will help to discover bugs and inconsistencies but also work as documentation and create a reference that makes refactoring easier down the road. This issue is meant to organize the progress on extending unit test coverage on the Nix code base. 

**Note**: Not _all_ code can reasonably be tested or even needs to. Yet a certain coverage is desirable and we are starting from zero.

### How To Test

Nix uses [Google Test](https://github.com/google/googletest). Look at the [Google Test Primer](https://github.com/google/googletest/blob/master/googletest/docs/primer.md) and refer to the existing tests for some guidance on how to write tests.

**Note**: We are writing tests for existing implementations. However you should still avoid looking at the implementation whenever possible as this often leads to tests that (even if unconsciously) re-implement the functions under test. Look instead at the header files and base tests on your assumptions about how the functions should be behaving and encode this in assertions.

### What To Test

The list below just lists the top-level directories from `src/` and is meant as a simple starting point to iterate on.

- [ ] build-remote
- [ ] cpptoml
- [ ] error-demo
- [ ] libexpr
- [ ] libfetchers
- [ ] libmain
- [ ] libstore
- [ ] libutil
  - ~url.hh~  #3603 
  - ~util.hh~ #3571 
  - hash.hh #3620 
  - json.hh #3621 
- [ ] nix
- [ ] nix-build
- [ ] nix-channel
- [ ] nix-collect-garbage
- [ ] nix-copy-closure
- [ ] nix-daemon
- [ ] nix-env
- [ ] nix-instantiate
- [ ] nix-prefetch-url
- [ ] nix-store
- [ ] resolve-system-dependencies

### How to participate

Mostly **just do it**... 

But if you are going to write some more involved tests you might want to mention your intention to do so. If you do write a PR that adds tests, please mention this issue ;)

gilligan:
So as I continue to work on unit tests (and towards covering everything that is reasonably testable in `libutil`) I added a FAQ addressing some topics that came up recently.

I would like to continue to work on tests but I also don't want to do so on my own indefinitely so I would like to move into a direction of recognizing unit tests as mandatory for future contributions/PRs instead of a "nice thing to have".

When the Rust code was added that came with unit tests right away because that's just more common place and easily added to Rust code anyway. Yet now that we have the setup in place there is no reason to not test C++ code.

[...]

PR #3671, opened 2020-06-08
Add unit tests for rust-ffi

gilligan:
Adds very simple tests for String and Result from rust-ffi.h

**Note**: The `Result` test case for Err values should be adjusted such
that it carries an actual exception. I failed to get that done however.

@edolstra Maybe you know how to do this.

### DIFF
--- a/src/libutil/tests/local.mk
+++ b/src/libutil/tests/local.mk
@@ -10,6 +10,6 @@ libutil-tests_SOURCES := $(wildcard $(d)/*.cc)
 
 libutil-tests_CXXFLAGS += -I src/libutil
 
-libutil-tests_LIBS = libutil
+libutil-tests_LIBS = libutil libnixrust
 
 libutil-tests_LDFLAGS := $(GTEST_LIBS)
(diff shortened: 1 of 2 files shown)


Ericson2314:
Looking forward to this :). I'll add some more in https://github.com/NixOS/nix/pull/3450 once this is merged.

edolstra:
I feel that because the Rust API is still very much a work in progress, it's premature to add tests at this point. (E.g. if we switch to `cxx` it will invalidate all these tests.)

gilligan:
@edolstra i would say that is maybe even more reason to do so. The idea behind unit tests is not to forever encode the way something is supposed to work without ever changing. Instead it’s supposed to make it easier to do the changes i’d say.

If there are changes that will land sometime soon that will break this tests that’s great. That means the tests will be fixed and adjusted and also that they cover relevant code.

gilligan:
@edolstra I added a short _FAQ_ over at https://github.com/NixOS/nix/issues/3622 which also touches on "unit tests versus functional tests" i.e the topic you are raising here.

gilligan:
Considering #3702 i am going to close this.

PR #3673, opened 2020-06-09
WIP: Add Store API-based HashedMirrors

matthewbauer:
This is a replacement for “hashed-mirrors” which provide a way to
supply fixed output, flat store objects. An advantage of
hashed-mirrors over other substituters is that we can use them in
multiple store directories. HashedMirrorStore succeeds hashed mirror,
but now uses the store api.

url format for HashedMirrorStore looks like:

  - hashed-mirror+file://
  - hashed-mirror+http://

The hashed-mirrors option still works, it just rewrites them as
substituters for the above.

Currently only file:// is implemented, http:// still needs to be
added.

@edolstra Does this make sense to you? Specifically, I'm wondering if adding `ca` here breaks any assumptions we have in the store api. Leaving this partially unfinished (no http support) to wait for some feedback.

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -263,7 +263,7 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
     stats.narInfoWrite++;
 }
 
-bool BinaryCacheStore::isValidPathUncached(const StorePath & storePath)
+bool BinaryCacheStore::isValidPathUncached(const StorePath & storePath, const std::string ca)
 {
     // FIXME: this only checks whether a .narinfo with a matching hash
     // part exists. So ‘f4kb...-foo’ matches ‘f4kb...-bar’, even
@@ -271,7 +271,7 @@ bool BinaryCacheStore::isValidPathUncached(const StorePath & storePath)
     return fileExists(narInfoFileFor(storePath));
 }
 
-void BinaryCacheStore::narFromPath(const StorePath & storePath, Sink & sink)
+void BinaryCacheStore::narFromPath(const StorePath & storePath, Sink & sink, const std::string ca)
 {
     auto info = queryPathInfo(storePath).cast<const NarInfo>();
 
@@ -298,7 +298,7 @@ void BinaryCacheStore::narFromPath(const StorePath & storePath, Sink & sink)
 }
 
 void BinaryCacheStore::queryPathInfoUncached(const StorePath & storePath,
-    Callback<std::shared_ptr<const ValidPathInfo>> callback) noexcept
+    Callback<std::shared_ptr<const ValidPathInfo>> callback, const std::string ca) noexcept
 {
     auto uri = getUri();
     auto storePathS = printStorePath(storePath);
--- a/src/libstore/hashed-mirror-store.cc
+++ b/src/libstore/hashed-mirror-store.cc
@@ -182,6 +182,10 @@ public:
         }
     }
 
+    bool supportsOtherStoreDir() {
+        return true;
+    }
+
 };
 
 static RegisterStoreImplementation regStore([](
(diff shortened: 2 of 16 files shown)


src/libstore/binary-cache-store.cc
```diff
@@ -263,15 +263,15 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
     stats.narInfoWrite++;
 }
 
-bool BinaryCacheStore::isValidPathUncached(const StorePath & storePath)
+bool BinaryCacheStore::isValidPathUncached(const StorePath & storePath, const std::string ca)
```
Ericson2314, line 266: nit for later: but `std::string_view` would be better

src/libstore/hashed-mirror-store.cc
```diff
@@ -0,0 +1,202 @@
...
+
+    StorePath addTextToStore(const string & name, const string & s,
+        const StorePathSet & references, RepairFlag repair = NoRepair) override
+    { unsupported("addTextToStore"); }
+
+    BuildResult buildDerivation(const StorePath & drvPath, const BasicDerivation & drv,
+        BuildMode buildMode = bmNormal) override
+    { unsupported("buildDerivation"); }
+
+    StorePathSet queryValidPaths(const StorePathSet & paths, SubstituteFlag maybeSubstitute, std::map<std::string, std::string> pathsInfo) override
```
Ericson2314, line 140: We can use `StorePath` keys if https://github.com/NixOS/nix/pull/3450 is merged

src/libstore/binary-cache-store.cc
edolstra: Yes, `const std::string` isn't very useful. Maybe you intended `const std::string &`.

edolstra:
It's not entirely clear to me why we need a separate `HashedMirror`. It seems that you should be able to use any `BinaryCacheStore` so long as it only contains CA paths without references. Or to be precise, you can substitute a store path from another store, even if it has another prefix, if you can compute what the corresponding store path in that store would be.

src/libstore/binary-cache-store.cc
matthewbauer: actually i just meant the callee shouldn't modify ca - `const std::string &` would let you modify it. but it's confusing right next to `const StorePath &`.

matthewbauer:
Current BinaryCacheStore requires that .narinfo's exist for each valid path. That hasn't been the case with hashed-mirrors, and I think it would break existing hashed-mirror workflows.

edolstra:
I would just convert hashed mirrors into binary caches by copying the tarballs to a binary cache, and then get rid of the hashed mirror support in `fetchurl`.

edolstra:
In fact `cache.nixos.org` should already contain all the tarballs in `tarballs.nixos.org` (and a lot more) because of the way Hydra works.

Ericson2314:
@edolstra So you mean we can keep the extra CA arguments, and use them to look up data in any store regardless of it's prefix, by using the CA argument to compute the path for that store's prefix on the fly?

src/libstore/store-api.hh
```diff
@@ -445,7 +448,7 @@ public:
        sizes) of a set of paths.  If a path does not have substitute
        info, it's omitted from the resulting ‘infos’ map. */
     virtual void querySubstitutablePathInfos(const StorePathSet & paths,
-        SubstitutablePathInfos & infos) { return; };
+        SubstitutablePathInfos & infos, std::map<std::string, Derivation> drvs = {}) { return; };
```
Ericson2314, line 451: Ought to be `std::map<std::string /* StorePath */, std::map<std::string /* output name */, std::string /* ContentAddress */>>`

Ericson2314: For back-compat, we can overload the path, and turn the set into a map with all empty strings.

matthewbauer:
closing - and will open a PR with a slightly different approach

matthewbauer:
opened https://github.com/NixOS/nix/pull/3689 as an alternative

PR #3674, opened 2020-06-09
Allow empty hash in derivations

matthewbauer:
follow up of https://github.com/NixOS/nix/pull/3544

This allows hash="" so that it can be used for debugging purposes. For
instance, this gives you an error message like:

  warning: found empty hash, assuming you wanted 'sha256:0000000000000000000000000000000000000000000000000000'
  hash mismatch in fixed-output derivation '/nix/store/asx6qw1r1xk6iak6y6jph4n58h4hdmbm-nix':
    wanted: sha256:0000000000000000000000000000000000000000000000000000
    got:    sha256:0fpfhipl9v1mfzw2ffmxiyyzqwlkvww22bh9wcy4qrfslb4jm429

this only effects "derivations" with outputHash set.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1131,9 +1131,14 @@ static void prim_path(EvalState & state, const Pos & pos, Value * * args, Value 
             filterFun = attr.value;
         } else if (n == "recursive")
             method = FileIngestionMethod { state.forceBool(*attr.value, *attr.pos) };
-        else if (n == "sha256")
-            expectedHash = Hash(state.forceStringNoCtx(*attr.value, *attr.pos), htSHA256);
-        else
+        else if (n == "sha256") {
+            auto hashStr = state.forceStringNoCtx(*attr.value, *attr.pos);
+            if (hashStr == "") {
+                expectedHash = Hash(htSHA256);
+                printError("warning: found empty hash, assuming you wanted '%s'", expectedHash.to_string());
+            } else
+                expectedHash = Hash(hashStr, htSHA256);
+        } else
             throw EvalError(format("unsupported argument '%1%' to 'addPath', at %2%") % attr.name % *attr.pos);
     }
     if (path.empty())
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -719,7 +719,13 @@ static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * * 
             throw Error(format("multiple outputs are not supported in fixed-output derivations, at %1%") % posDrvName);
 
         HashType ht = outputHashAlgo.empty() ? htUnknown : parseHashType(outputHashAlgo);
-        Hash h(*outputHash, ht);
+
+        Hash h;
+        if (outputHash->empty()) {
+            h = Hash(ht);
+            printError("warning: found empty hash, assuming you wanted '%s'", h.to_string());
+        } else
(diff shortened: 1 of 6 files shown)


Ericson2314:
Yay, no more coppying and modifying a hash from elsewhere when I am making a new fixed output derivation!

vcunat:
> Yay, no more copying and modifying a hash from elsewhere [...]

That's not a good way of obtaining the hashes, regardless of this PR.  (because `fetchurl` -> `curl --insecure`)

matthewbauer:
> EDIT: well, perhaps we should now drop the `--insecure` iff hash is empty 🤔

Why is --insecure set in fetchurl builder at all?

Ericson2314:
Woah, I did not know that. Yikes.

matthewbauer:
> > Yay, no more copying and modifying a hash from elsewhere [...]
> 
> That's not a good way to obtain hashes, regardless of this PR. (because `fetchurl` -> `curl --insecure`)
> 
> EDIT: well, perhaps we should now drop the `--insecure` iff hash is empty 🤔

I think this should be addressed in https://github.com/NixOS/nixpkgs/pull/89885/commits/0046802ab6d3389b45f3e9fce6f5e2746e2e7f80. Note that this is actually a Nixpkgs issue, not a Nix one.

vcunat:
Well... I thought the main argument is that when you know the output hash, there's not that much to gain from TLS (well, some privacy, though meta-data will probably leak part of that anyway).

matthewbauer:
Yeah - I noticed fetchgit just sets it unconditionally already (https://github.com/NixOS/nixpkgs/blob/a1aecffc971202acaca5882e87510d2a15a000f7/pkgs/build-support/fetchgit/default.nix#L64).

src/libexpr/primops.cc
```diff
@@ -719,7 +719,13 @@ static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * *
             throw Error(format("multiple outputs are not supported in fixed-output derivations, at %1%") % posDrvName);
 
         HashType ht = outputHashAlgo.empty() ? htUnknown : parseHashType(outputHashAlgo);
-        Hash h(*outputHash, ht);
+
+        Hash h;
+        if (outputHash->empty()) {
+            h = Hash(ht);
+            printError("warning: found empty hash, assuming you wanted '%s'", h.to_string());
```
edolstra, line 726: => `warn()`.

src/libexpr/primops.cc
```diff
@@ -1125,9 +1131,14 @@ static void prim_path(EvalState & state, const Pos & pos, Value * * args, Value
...
-        else if (n == "sha256")
-            expectedHash = Hash(state.forceStringNoCtx(*attr.value, *attr.pos), htSHA256);
-        else
+        else if (n == "sha256") {
+            auto hashStr = state.forceStringNoCtx(*attr.value, *attr.pos);
+            if (hashStr == "") {
+                expectedHash = Hash(htSHA256);
+                printError("warning: found empty hash, assuming you wanted '%s'", expectedHash.to_string());
+            } else
+                expectedHash = Hash(hashStr, htSHA256);
```
edolstra, line 1140: There is a lot of cut&paste here, maybe it's better to add a function that parses a hash and prints a warning it if's empty?

edolstra:
Thanks!

PR #3675, opened 2020-06-09
Fetch commits from github private repos using Authorization header

imalsogreg:
`nix flake info` calls a github API that requires authorization when the repository in question is private. Currently this request fails with a 404.

This commit adds a `headers` parameter to the `FileTransferRequest`, and the github fetchers now call it with `Authorization: token <github-access-token>`.

This differs from the current convention in nix of passing the access token as a query parameter, because query parameter tokens are deprecated and will be disallowed in November 2020. Using them today triggers a warning email.

This PR grew to include the work done in #3688. I've now tested it with private github and gitlab repositories.

```
[greghale@cedric:~/code/nix]$ result/bin/nix --option gitlab-access-token $GITLAB_TOKEN flake info gitlab:secret_org/secret_repo/tmp
warning: flake 'gitlab:secret_org/secret_repo/622bee9dc5f8723fa3e8aaaf4359946277e441e8' has deprecated attribute 'edition'
Resolved URL:  gitlab:secret_org/secret_repo/tmp
Locked URL:    gitlab:secret_org/secret_repo/622bee9dc5f8723fa3e8aaaf4359946277e441e8
Description:   Secret repo's description
Path:          /nix/store/11z2mfkba3kfkcb18l6aab4k1gk9b6h5-source
Revision:      622bee9dc5f8723fa3e8aaaf4359946277e441e8
Last modified: 2020-06-10 22:20:05

[greghale@cedric:~/code/nix]$ result/bin/nix --option github-access-token $GITHUB_TOKEN flake info github:secret_org/secret_repo
warning: flake 'github:secret_ord/secret_repo/622bee9dc5f8723fa3e8aaaf4359946277e441e8' has deprecated attribute 'edition'
Resolved URL:  github:secret_org/secret_repo
Locked URL:    github:secret_org/secret_repo/622bee9dc5f8723fa3e8aaaf4359946277e441e8
Description:   Secret repo's description
Path:          /nix/store/aaa111kba4kfkcb18l6i234k1gk9b6h5-source
Revision:      622bee9dc5f8723fa3e8aaaf4359946277e441e8
Last modified: 2020-06-10 22:20:05
```


Ericson2314:
Link to deprecation notice: https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters

imalsogreg:
@Ericson2314 I only changed the one API endpoint that didn't yet use an API token (thus resulting in a 404). Think I should update the other github fetchers with the same logic, or keep the scope of this PR small (fixing the 404)?

Kloenk:
Just saw this. I did a similar approach for gitlab. but used a map instead of a vector. 
See #3688

Ericson2314:
"Keys" can be repeated in headers, so a vector is more appropriate. But good to see this is being worked on for gitlab and github :).

Kloenk:
@imalsogreg 
> "Keys" can be repeated in headers, so a vector is more appropriate. But good to see this is being worked on for gitlab and github :).

That's a good argument. Will you implement my gitlab headers? Then I would close my pr.

imalsogreg:
@Kloenk Sure! I'll take a stab at it now.

src/libfetchers/github.cc
```diff
@@ -226,34 +258,46 @@ struct GitLabInputScheme : GitArchiveInputScheme
...
             host_url, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
             input.getRev()->to_string(Base16, false));
 
-        /* # FIXME: add privat token auth (`curl --header "PRIVATE-TOKEN: <your_access_token>"`)
-        std::string accessToken = settings.githubAccessToken.get();
-        if (accessToken != "")
-            url += "?access_token=" + accessToken;*/
+        std::string accessToken = settings.gitlabAccessToken.get();
+        if (accessToken != "") {
+            auto auth_header = accessHeaderFromToken(accessToken);
```
Kloenk, line 295: This [api documentation](https://docs.gitlab.com/ee/api/README.html#personalproject-access-tokens) of gitlab says If we use the `Authorization` header then we have to prefix the token with `Bearer`. We either should prefix it here or state that Bearer is required in front of the token.

imalsogreg: Thanks, good catch! I updated the way access token gets translated into headers here:  https://github.com/NixOS/nix/pull/3675/files#diff-597785e0a47d7f86a1982d22662296afR262

Kloenk:
Looks got to meet so far.

Maybe squash the commits? I think 8 is kinda a lot for this small change.

imalsogreg:
@Kloenk squashed down to 1, good point :) I'm ready for a merge when it's ready, and am happy to address any other feedback.

Ericson2314:
I wouldn't bother squashing things until it passes CI, unless one is find rebasing and force pushing every single time.

Kloenk:
@Ericson2314 the checks are failing, because flake-combat is not updated yet, and the new flake.lock uses version 6 (which is not supported yet).

So ci is not working on the flake branch at the current moment.

Ericson2314:
Ah right, forgot this was to that branch.

cole-h:
FYI, flake-compat [just updated](https://github.com/edolstra/flake-compat/commit/c75e76f80c57784a6734356315b306140646ee84) to support lockfile versions 6 and 7 -- if you re-trigger CI, it should turn green (hopefully)!

imalsogreg:
@edolstra or @Ericson2314 , would either of you mind giving this another look and/or a merge? I've brought it up to date with `master`.

Thanks for the flakes blog posts! They're super helpful.

edolstra:
Thanks, will have a look. I'm getting deprecation notices from GitHub every few weeks :-)

imalsogreg:
@edolstra @Ericson2314 @Kloenk Thanks! Also @kquick has opened a PR against my PR adding the ability to authenticate with on-premisis github/gitlab instances (https://github.com/imalsogreg/nix/pull/1). It involves changing the format of a nix config string for tokens, to accommodate the different types of tokens that github and gitlab accept, and to accommodate the fact that users have different tokens on multiple instances of github.

Should we merge that in to my PR and evaluate both at once? https://github.com/imalsogreg/nix/pull/1

Kloenk:
It's a rather big pr for that functionality. So I would keep it as 2 separate prs

PR #3678, opened 2020-06-10
Add a way to get all the outputs of a derivation with their label

regnat:
Generalize `queryDerivationOutputNames` and `queryDerivationOutputs` by adding a `queryDerivationOutputMap` that returns the map `outputName=>outputPath`.

This is needed by #3528 as it needs a way to get the path of a specific derivation output, which (afaik) isn't possible with the current API

(not that this is not equivalent to merging the results of `queryDerivationOutputs` and `queryDerivationOutputNames` as sets don't preserve the order, so we would end up with an incorrect mapping).

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -347,6 +347,15 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         break;
     }
 
+    case wopQueryDerivationOutputMap: {
+        auto path = store->parseStorePath(readString(from));
+        logger->startWork();
+        OutputPathMap outputs = store->queryDerivationOutputMap(path);
+        logger->stopWork();
+        writeOutputPathMap(*store, to, outputs);
+        break;
+    }
+
     case wopQueryDeriver: {
         auto path = store->parseStorePath(readString(from));
         logger->startWork();
--- a/src/libstore/path.hh
+++ b/src/libstore/path.hh
@@ -62,6 +62,7 @@ public:
 
 typedef std::set<StorePath> StorePathSet;
 typedef std::vector<StorePath> StorePaths;
+typedef std::map<string, StorePath> OutputPathMap;
 
 /* Extension of derivations in the Nix store. */
 const std::string drvExtension = ".drv";
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -242,6 +242,16 @@ bool Store::PathInfoCacheValue::isKnownNow()
     return std::chrono::steady_clock::now() < time_point + ttl;
 }
 
+StorePathSet Store::queryDerivationOutputs(const StorePath & path)
+{
+    auto outputMap = this->queryDerivationOutputMap(path);
+    StorePathSet outputPaths;
+    for (auto & i: outputMap) {
+        outputPaths.emplace(std::move(i.second));
+    }
+    return outputPaths;
+}
+
 bool Store::isValidPath(const StorePath & storePath)
 {
     std::string hashPart(storePath.hashPart());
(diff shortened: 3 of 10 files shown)


src/libstore/store-api.cc
```diff
@@ -241,6 +241,26 @@ bool Store::PathInfoCacheValue::isKnownNow()
...
+{
+    auto outputMap = this->queryDerivationOutputMap(path);
+    StorePathSet outputPaths;
+    for (auto & i: outputMap) {
+        outputPaths.emplace(std::move(i.second));
+    }
+    return outputPaths;
+}
+
+StringSet Store::queryDerivationOutputNames(const StorePath & path)
```
Ericson2314, line 254: Maybe add a comment saying these are obsolete?

src/libstore/path.hh
```diff
@@ -61,6 +61,7 @@ struct StorePath : rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>
 
 typedef std::set<StorePath> StorePathSet;
 typedef std::vector<StorePath> StorePaths;
+typedef std::map<string, StorePath> StorePathMap;
```
Ericson2314, line 64: I might call this `OutputPathMap` instead, as `StorePathMap` sounds to me like `std::map<StorePath, ???>`.

src/libstore/daemon.cc
```diff
@@ -308,8 +308,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
 
     case wopQueryReferences:
     case wopQueryReferrers:
-    case wopQueryValidDerivers:
-    case wopQueryDerivationOutputs: {
```
Ericson2314, line 312: Do you know what the policy is on supporting on protocol versions? I don't!

edolstra: The daemon should support older clients. So we can't remove APIs like this.

edolstra:
Do we actually need these functions? The outputs of a derivation are also available by doing `readDerivation(drvPath)`. IIRC, the only reason why we have `LocalStore::queryDerivationOutputs()`  was to make the garbage collector more efficient (since getting this stuff from SQLite is faster than reading a lot of .drvs). `queryDerivationOutputNames()` doesn't look like it was ever necessary from a performance perspective...

edolstra:
As a cleanup I've removed `queryDerivationOutputNames()` (045b07200c77bf1fe19c0a986aafb531e7e1ba54).

Ericson2314:
I agree having 3 of them is unnecessary, but this new on seams good to me? In https://github.com/NixOS/nix/pull/3523#issuecomment-643176537 you talk about not wanting to rely on derivations existing on disk as much. Avoiding `readDerivation` with these functions would seem to me in service of that goal.

src/libstore/daemon.cc
regnat: Woops sorry, I'll restore them

regnat:
> Do we actually need these functions? The outputs of a derivation are also available by doing `readDerivation(drvPath)`. IIRC, the only reason why we have `LocalStore::queryDerivationOutputs()` was to make the garbage collector more efficient (since getting this stuff from SQLite is faster than reading a lot of .drvs). `queryDerivationOutputNames()` doesn't look like it was ever necessary from a performance perspective...

Atm yes, but that won't be true anymore in a CA setting where the output paths don't match what's in the drv (which is the reason why I needed this in the first place)

src/libstore/path.hh
regnat: Indeed, that makes sense. Changed in 53edff29

regnat:
I've rebased it on top of master. No idea what the OSX failure is. It seems totally unrelated

Ericson2314:
Yes, we currently fetch the latest `nixos-20.03-small`, but I think that might be a bit too fast non-deterministism, as it opens the door to all these issues on Darwin deps.

edolstra:
Thanks, merged!

PR #3680, opened 2020-06-10
libutils/hash: rename Base to HashEncoding

zimbatm:
Compress the (Base, usePrefix) tuple to a single field that represents
the hash encoding. This makes more sense because SRI always has a
prefix.

This is a continuation of the work started at #3655

If this PR gets merged, the next PR will be introducing a ParsedHash subclass which retains the information on the original hash and allows `hash.to_string(hash.original_encoding)`


src/nix/hash.cc
```diff
@@ -10,18 +10,21 @@ using namespace nix;
...
-        mkFlag(0, "sri", "print hash in SRI format", &base, SRI);
-        mkFlag(0, "base64", "print hash in base-64", &base, Base64);
-        mkFlag(0, "base32", "print hash in base-32 (Nix-specific)", &base, Base32);
-        mkFlag(0, "base16", "print hash in base-16", &base, Base16);
+        mkFlag(0, "base16", "print hash in base16", &encoding, Base16);
+        mkFlag(0, "base32", "print hash in base32 (Nix-specific)", &encoding, Base32);
+        mkFlag(0, "base64", "print hash in base64", &encoding, Base64);
+        mkFlag(0, "prefixed-base16", "print hash in base16", &encoding, PrefixedBase16);
+        mkFlag(0, "prefixed-base32", "print hash in base32 (Nix-specific)", &encoding, PrefixedBase32);
+        mkFlag(0, "prefixed-base64", "print hash in base64", &encoding, PrefixedBase64);
```
edolstra, line 26: Let's not add these `prefixed-*` options. This is a legacy format that we don't need to expose in the new UI.

edolstra:
It's not entirely clear to me what the practical advantage of these changes is. OTOH the disadvantage (a lot of code churn) is clear.

zimbatm:
Agreed, without the next change, the PR doesn't bring a lot of value. Let me then bolt on the next PR so things become clearer.

src/nix/hash.cc
zimbatm: Ok!

Ericson2314:
@edolstra I would have thought one of the benefits of not having a stable interfaces is we need not shun churn, and can iteratively refactor things until they are nicer. (Last I checked, everyone agreed Nix was less than ideal when used as a library.) Even this PR without the follow-up I value for removing some redundancy from the printing options.

I would be more than happy to contribute and review dozens of these PRs if you change your mind about churn, so that their value would add up into something quite significant.

zimbatm:
Generally, I think the idea would be to move more things into Rust and have more of the churn go there. But that's really a discussion we need to have separately.

Ericson2314:
Sure, the first thing I tried was https://github.com/NixOS/nix/pull/3416, but by now I've gotten more familiar with C++ and am happy to refactor things in either language.

zimbatm:
I pushed 2 more commits to finalize the design. I didn't know how to sub-class the Hash struct so the original encoding is embedded in there.

Ericson2314:
I think this will work.
```cpp
struct HashWithEncoding : Hash
{
    HashEncoding encoding;
    HashWithEncoding(std::string_view s, HashEncoding encoding HashType type = htUnknown)
        : Hash(s, type), encoding(encoding);
    std::string to_string() const {
        return to_string(encoding);
    }
}
```

zimbatm:
related to #3650

Ericson2314:
Hmm? I think #3650 is a good change, but I better made it not contain the `enum struct` churn PR if it's going to be merged.

src/libutil/hash.hh
```diff
@@ -30,6 +41,7 @@ struct Hash
     unsigned char hash[maxHashSize] = {};
 
     HashType type = htUnknown;
+    HashEncoding encoding = heUnknown;
```
edolstra, line 44: Please remove this field. The `Hash` type should not store an encoding. It creates all sorts of issues like whether `operator ==` should compare the encoding.

zimbatm:
Closing as this PR has completely rotted. I am not sure the change makes sense in the new code.

PR #3685, opened 2020-06-11
documentation: mention substituter precedence

asymmetric:
Mentions the order in which substituters are queried.

### DIFF
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -888,7 +888,7 @@ password <replaceable>my-password</replaceable>
   <varlistentry xml:id="conf-substituters"><term><literal>substituters</literal></term>
 
     <listitem><para>A list of URLs of substituters, separated by
-    whitespace.  The default is
+    whitespace.  The substituters are queried from left to right.  The default is
     <literal>https://cache.nixos.org</literal>.</para></listitem>
 
   </varlistentry>


edolstra:
That's actually not true. Binary caches have a priority field that determines the order (see the `Priority` field in https://cache.nixos.org/nix-cache-info).

xwvvvvwx:
@edolstra what happens if two caches have the same priority?

edolstra:
Then they're tried in the order in which they're defined in nix.conf.

xwvvvvwx:
@edolstra thanks!

I'm curious about why caches are allowed to define their own priority? 

I have a few cacix substituters defined and would like to be able to tell nix to prefer binaries from `cache.nixos.org` over those from my cachix sustituters, but that doesn't seem to be possible if any cache can set itself to prio 1 without my consent?

asymmetric:
@edolstra Would this be better?

> Substituters with the same `Priority` value are queried from left to right.

grahamc:
btw it isnt a good idea to depend on priority as a security feature, since it doesn't provide meaningful control over what you get from what cache.

edolstra:
The original use case was to ensure that `cache.nixos.org` (which is fast) is prioritized over `hydra.nixos.org` (which is slow). Nowadays `hydra.nixos.org` no longer provides a binary cache so it doesn't matter anymore.

BTW I think you can also set the priority in the substituter URL (e.g. `http://bla?priority=123`).

xwvvvvwx:
> btw it isnt a good idea to depend on priority as a security feature, since it doesn't provide meaningful control over what you get from what cache.

@grahamc Could you elaborate a bit? I'm not sure I follow.

Fuuzetsu:
Is higher Priority better? The source says otherwise https://github.com/NixOS/nix/blob/master/src/libstore/store-api.hh#L195 but common usage and also experimental findings like in https://discourse.nixos.org/t/how-does-nix-choose-a-substituter-when-there-is-more-than-one-choice/3658 suggest otherwise.

asymmetric:
@edolstra does specifying the priority in the substituter URL have precedence over the `Priority` field in the cache itself?

@Fuuzetsu where do you see disagreement on the sorting? There seems to be agreement on `lower value = higher priority`.

Fuuzetsu:
@asymmetric I set priority 50 in my s3 cache amd it gets picked over priority 40 official nixos cache.

In linked issue, nix-serve with priority 30 gets checked only after the official priority 40 cache.

asymmetric:
@Fuuzetsu in the linked issue, priority 30 gets checked **before** 40:

> My local nix.conf specifies both the internal CI server as well as cache.nixos.org as substituters.
> 
> I’ve noticed that the internal CI server is picked for paths that exist on BOTH caches.
> I wondered how does it pick a CI server so I tried changing the order of the CI servers in nix.conf but it kept picking the internal CI server.

> The CI server cache’s priority is set to 30.

Fuuzetsu:
> in the linked issue, priority 30 gets checked before 40

Oh, you're right. I have a priority to change. Thank you for the doc update.

PR #3689, opened 2020-06-12
Substitutions from different store dirs

matthewbauer:
Substituters can substitute from one store dir to another with a
little bit of help. The store api just needs to have a CA so it can
recompute the store path based on the new store dir. We can only do
this for fixed output derivations with no references, though.


src/libstore/local-store.cc
```diff
@@ -855,16 +855,24 @@ StorePathSet LocalStore::querySubstitutablePaths(const StorePathSet & paths)
...
     if (!settings.useSubstitutes) return;
     for (auto & sub : getDefaultSubstituters()) {
-        if (sub->storeDir != storeDir) continue;
-        for (auto & path : paths) {
-            if (infos.count(path)) continue;
+        for (auto & path_ : paths) {
+            auto path(path_.clone());
             debug("checking substituter '%s' for path '%s'", sub->getUri(), printStorePath(path));
             try {
                 auto info = sub->queryPathInfo(path);
```
Ericson2314, line 866: Need to use the `dstPath`

src/libstore/local-store.cc
```diff
@@ -855,16 +855,24 @@ StorePathSet LocalStore::querySubstitutablePaths(const StorePathSet & paths)
 
 
 void LocalStore::querySubstitutablePathInfos(const StorePathSet & paths,
-    SubstitutablePathInfos & infos)
+    SubstitutablePathInfos & infos, std::map<std::string, std::string> pathsCA)
 {
     if (!settings.useSubstitutes) return;
     for (auto & sub : getDefaultSubstituters()) {
-        if (sub->storeDir != storeDir) continue;
```
Ericson2314, line 862: probably better to call this `srcPath`

src/libstore/store-api.cc
```diff
@@ -580,6 +580,13 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
```
Ericson2314, line 580: On previous line, need to use `srcStorePath`

src/libstore/store-api.cc
```diff
@@ -580,6 +580,13 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
 
     uint64_t total = 0;
 
+    // recompute store path on the chance dstStore does it differently
+    if (hasPrefix(info->ca, "fixed:") && info->references.empty()) {
+        auto info2 = make_ref<ValidPathInfo>(*info);
+        info2->path = dstStore->makeStorePath("output:out", hashString(htSHA256, info->ca), storePath.name());
+        info = info2;
+    }
```
Ericson2314, line 588: Move path calculation up to define `srcStorePath` to use with `srcStore->queryPathInfo`.

src/libstore/store-api.cc
```diff
@@ -580,6 +580,13 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
...
+
     if (!info->narHash) {
         StringSink sink;
         srcStore->narFromPath({storePath}, sink);
```
Ericson2314, line 592: Down below at 
```
dstStore->addToStore(*info, *source, repair, checkSigs);
```
update info to contain `dstStorePath` then

src/libstore/store-api.cc
```diff
@@ -580,6 +580,13 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
 
     uint64_t total = 0;
 
+    // recompute store path on the chance dstStore does it differently
+    if (hasPrefix(info->ca, "fixed:") && info->references.empty()) {
+        auto info2 = make_ref<ValidPathInfo>(*info);
+        info2->path = dstStore->makeStorePath("output:out", hashString(htSHA256, info->ca), storePath.name());
```
Ericson2314, line 586: See `ValidPathInfo::isContentAddressed` for a different-looking way to calculate a store path from a CA.

Ericson2314:
@edolstra We really like this as a new future. One downside of *replacing* hashed mirrors with it, however, is that we still use the same "name" when rewriting the path even though it is immaterial for fixed output derivations. This means one has to be quite cautious about renaming fixed output derivations, which isn't really the case today.

I might still consider the old https://github.com/NixOS/nix/pull/3673 in addition to this as a stop-gap until we add another mitigation, such as making a `<ca>.cainfo` file to allow looking up content addresses rather than store paths directly.

matthewbauer:
There are some legitimate issues with #3673 I think though. We don’t have any good disk caching because we don’t have narinfo and we also will end up requesting a lot more things from tarballs.nixos.org than we currently do. It might make sense to keep hashed-mirrors for people that still use them.

Ericson2314:
I am fine with either transition plan (this + #3673 hash mirror store or this without removing hashed mirrors), but for the record I do think we could fix the issues with #3673 with enough elbow grease (for example, perhaps `CURLOPT_NOBODY` to just see whether something exists).

src/libstore/build.cc
```diff
@@ -4408,6 +4411,19 @@ void SubstitutionGoal::tryNext()
         throw;
     }
 
+    if (info->path != storePath) {
+        if (info->isContentAddressed(*sub)) {
+            auto info2 = std::const_pointer_cast<ValidPathInfo>(std::shared_ptr<const ValidPathInfo>(info));
+            info2->path = storePath.clone();
```
edolstra, line 4417: This looks dangerous. Isn't this changing the `ValidPathInfo` object that might point to the store's `ValidPathInfo` cache? Maybe just make a copy (`auto info2 = std::make_shared<ValidPathInfo>(*info)`).

src/nix/add-to-store.cc
```diff
@@ -21,6 +22,13 @@ struct CmdAddToStore : MixDryRun, StoreCommand
             .labels = {"name"},
             .handler = {&namePart},
         });
+
+        addFlag({
+            .longName = "flat",
+            .shortName = 0,
+            .description = "use flat file ingestion",
+            .handler = {&ingestionMethod, FileIngestionMethod::Flat},
+        });
```
edolstra, line 31: Maybe we should split `add-to-store` into `add-path` and `add-file`, analogous to `nix hash-file` and `nix hash-path`.

src/libstore/build.cc
edolstra: We should check here that the path has no references, since we can't rewrite inter-store references.

src/nix/add-to-store.cc
edolstra: Also we shouldn't take about "file ingestion" in the UI, since nobody will know what that means.

src/libstore/store-api.hh
```diff
@@ -445,7 +449,7 @@ public:
        sizes) of a set of paths.  If a path does not have substitute
        info, it's omitted from the resulting ‘infos’ map. */
     virtual void querySubstitutablePathInfos(const StorePathSet & paths,
-        SubstitutablePathInfos & infos) { return; };
+        SubstitutablePathInfos & infos, std::map<std::string, std::string> pathsCA = {}) { return; };
```
edolstra, line 452: Need a comment here what `pathsCA` is for.

edolstra: Also, if I understand it correctly, `paths` and `pathsCA` should be combined into a type like `std::map<StorePath, std::optional<std::string>>`.

src/libstore/store-api.cc
```diff
@@ -660,9 +688,17 @@ void copyPaths(ref<Store> srcStore, ref<Store> dstStore, const StorePathSet & st
...
-            auto storePath = dstStore->parseStorePath(storePathS);
+            auto storePath = srcStore->parseStorePath(storePathS);
+            auto info = srcStore->queryPathInfo(storePath);
+
+            auto storePathForDst = storePath.clone();
+            if (info->isContentAddressed(*srcStore)) {
+                storePathForDst = dstStore->makeFixedOutputPathFromCA(storePath.name(), info->ca);
+                if (storePathForDst != storePath)
+                    debug("rewriting path '%s' to '%s' for substituter '%s'", srcStore->printStorePath(storePath), dstStore->printStorePath(storePathForDst), dstStore->getUri());
+            }
```
edolstra, line 699: This requires a change to the `copyPaths()` interface, since it can now add a *different* path to the destination store than what the caller asked for, with no way to know what path got added. So maybe it should return a `std::map<StorePath, StorePath>`.

src/libstore/store-api.cc
```diff
@@ -642,15 +662,23 @@ void copyPaths(ref<Store> srcStore, ref<Store> dstStore, const StorePathSet & st
...
-            if (dstStore->isValidPath(dstStore->parseStorePath(storePath))) {
+        [&](const Path & storePathS) {
+            auto storePath = srcStore->parseStorePath(storePathS);
+
+            auto info = srcStore->queryPathInfo(storePath);
+            auto storePathForDst = storePath.clone();
+            if (info->isContentAddressed(*srcStore)) {
+                storePathForDst = dstStore->makeFixedOutputPathFromCA(storePath.name(), info->ca);
+                if (storePathForDst != storePath)
+                    debug("rewriting path '%s' to '%s' for substituter '%s'", srcStore->printStorePath(storePath), dstStore->printStorePath(storePathForDst), dstStore->getUri());
```
edolstra, line 673: We're not rewriting, right?

src/libstore/store-api.hh
matthewbauer: Yeah that would be better. Is it okay to make changes to store-api.hh like that? We'd have to replace old querySubstitutablePathInfos that may not care about ca.

src/libstore/build.cc
Ericson2314: Yes. Though I do look forward to later recursively rewriting an all-CA closure provided the store path is the same length!

src/libstore/store-api.cc
Ericson2314: Indeed we aren't.

src/libstore/store-api.hh
```diff
@@ -739,11 +744,13 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
...
-   (i.e. if A is a reference of B, then A is copied before B), but
-   the set of store paths is not automatically closed; use
-   copyClosure() for that. */
-void copyPaths(ref<Store> srcStore, ref<Store> dstStore, const StorePathSet & storePaths,
+   in parallel. They are copied in a topologically sorted order (i.e.
+   if A is a reference of B, then A is copied before B), but the set
+   of store paths is not automatically closed; use copyClosure() for
+   that. Returns a map of what each path was copied to the dstStore
+   as. */
+std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore,
```
Ericson2314, line 752: you can use `std::pair` or `std::tuple` to return multiple values.

matthewbauer:
One issue I noticed is that wopQuerySubstitutablePathInfo and wopQuerySubstitutablePathInfos aren't getting the ca value in the remote store. I'm not sure how to change this without breaking compatibility with older protocols.

src/libstore/build.cc
```diff
@@ -4412,14 +4416,16 @@ void SubstitutionGoal::tryNext()
     sub = subs.front();
     subs.pop_front();
 
-    if (sub->storeDir != worker.store.storeDir) {
-        tryNext();
-        return;
+    auto subPath = storePath;
+    if (ca) {
```
edolstra, line 4420: Shouldn't we do `tryNext()` if `sub->storeDir != worker.store.storeDir && !ca`?

src/libstore/build.cc
```diff
@@ -4526,8 +4545,15 @@ void SubstitutionGoal::tryToRun()
             Activity act(*logger, actSubstitute, Logger::Fields{worker.store.printStorePath(storePath), sub->getUri()});
             PushActivity pact(act.id);
 
+            auto subPath = storePath;
+            if (ca) {
+                subPath = sub->makeFixedOutputPathFromCA(storePath.name(), *ca);
```
edolstra, line 4550: Since we already computed `subPath` in `SubstitutionGoal::tryNext()`, we could avoid computing it again by storing `subPath` in `SubstitutionGoal`.

Ericson2314:
I think error in CI has been fixed on master since.

src/libstore/build.cc
Ericson2314: @matthewbauer I think @edolstra is right? It will eventually try next after `queryPathInfo` fails, but we shouldn't wait for that guaranteed failure, right?

Mic92:
Wow this actually really cool. This could solve installing nix without root into the user's home right?

Ericson2314:
@Mic92 So it is my goal to handle more obscure situations like:
 - rewriting content addresses with store paths with different store dirs if the store dir lengths align.
 - rewriting names if the lengths align
 - *opting in* to rewriting if the lengths don't align, but the destination is shorter so we can pad with `////`

This basic approach scales to those sorts of things, and combined with CA derivations will get us decently close to making arbitrary store dirs work in practice. I think that is the right foundation on the Nix sides to then pursue the various tricks discussed in https://github.com/NixOS/rfcs/pull/17 on the Nixpkgs side and get all the way there for practical purposes.

Mic92:
> @Mic92 So it is my goal to handle more obscure situations like:
> 
>     * rewriting content addresses with store paths with different store dirs if the store dir lengths align.
> 
>     * rewriting names if the lengths align
> 
>     * _opting in_ to rewriting if the lengths don't align, but the destination is shorter so we can pad with `////`
> 
> 
> This basic approach scales to those sorts of things, and combined with CA derivations will get us decently close to making arbitrary store dirs work in practice. I think that is the right foundation on the Nix sides to then pursue the various tricks discussed in [NixOS/rfcs#17](https://github.com/NixOS/rfcs/pull/17) on the Nixpkgs side and get all the way there for practical purposes.

I thought a bit about changing the length of the store path i.e. exporting to different directories. The hard part are string references elf binaries, but it may be possible to solve this as well: https://github.com/yalue/elf32_string_replace There might be other binary formats that could also break but I think with plain text substitutions + elf binaries most programs would already run.

Ericson2314:
@mic92 Cool. Basically it's my view that we need the basic Nix support to get us motivated, but once we are motivated we can peruse all the tricks in the book to beat back mass rebuilds:

 - Rewrite strings tables, or even delay linking of data sections (c.f. how I think gettext internationalization works)

 - Can't actually stat SOs or dylibs during build (trying to observe file or even whether it is mounted at location would fail build), and just get dummy path and symbols via TAPI file or linker script (cc @eternaleye)

 - Using LTO and other tricks to share work between optimization levels---it's great to get a quick smoke test with global -O0 and if all is well wait for bigger builds

 - https://github.com/NixOS/nix/issues/3819

 - Moving stuff from bash to eval time

I think we can really pull all stops and increase our productivity by *orders of magnitude*. Just need to get these initial steps out of the way so the "crowd" is motivated enough that the rest can be crowdsourced.

edolstra:
Sorry, but that's crazy talk. There is no way we're going to teach Nix how rewrite ELF binaries, or any file format. Nix is a generic package manager so it doesn't (and shouldn't) have any knowledge of particular file formats.

And I don't think it's even possible for ELF binaries. A tool like `patchelf` can rewrite some things because we know the relevant structure. But .data and .text (code) sections don't really have any defined structure, so if we move a string like `/nix/store/abc` in a data section, we wouldn't know how to update the references in the text section. (ELF binaries don't have to be relocatable after all, so a machine instruction in the .text section can refer directly to a .data address like 0x58a000.) 

Also, how would you distinguish between `/nix/store` occurring as part of a reference, versus (say) in an HTML doc that shouldn't be rewritten (e.g. a string like "Nix by default uses /nix/store")?

Mic92:
> Sorry, but that's crazy talk. There is no way we're going to teach Nix how to rewrite ELF binaries, or any file format. Nix is a generic package manager so it doesn't (and shouldn't) have any knowledge of particular file formats.
> 
> And I don't think it's even possible for ELF binaries. A tool like `patchelf` can rewrite some things because we know the relevant structure. But .data and .text (code) sections don't really have any defined structure, so if we move a string like `/nix/store/abc` in a data section, we wouldn't know how to update the references in the text section. (ELF binaries don't have to be relocatable after all, so a machine instruction in the .text section can refer directly to a .data address like 0x58a000.)
> 
> Also, how would you distinguish between `/nix/store` occurring as part of a reference, versus (say) in an HTML doc that shouldn't be rewritten (e.g. a string like "Nix by default uses /nix/store")?

That would be external tools, not part of nix, like patchelf is not part of nix itself.

Ericson2314:
> > Sorry, but that's crazy talk. There is no way we're going to teach Nix how to rewrite ELF binaries, or any file format. Nix is a generic package manager so it doesn't (and shouldn't) have any knowledge of particular file formats.
> > And I don't think it's even possible for ELF binaries. A tool like `patchelf` can rewrite some things because we know the relevant structure. But .data and .text (code) sections don't really have any defined structure, so if we move a string like `/nix/store/abc` in a data section, we wouldn't know how to update the references in the text section. (ELF binaries don't have to be relocatable after all, so a machine instruction in the .text section can refer directly to a .data address like 0x58a000.)
> > Also, how would you distinguish between `/nix/store` occurring as part of a reference, versus (say) in an HTML doc that shouldn't be rewritten (e.g. a string like "Nix by default uses /nix/store")?
> 
> That would be external tools, not part of nix, like patchelf is not part of nix itself. It's possible to compile binaries to be relocatable.

Sorry!! Yes I absolutely did mean for any "tricks" to be in Nixpkgs/user code and not Nix, which I agree should stay a generic tool (c.f. CMake vs Ninja). Edited my post to make that clear----keep the kitchen sink out of Nix :)!

src/libstore/build.cc
matthewbauer: this was addressed in 1f9cb06db23f357f06404969820be1d3d4118a18

bhipple:
It looks like you're deleting hashed mirrors, but is there a way to have a dead simple FTP server serving source tarballs for mirroring into enterprise networks?  We rely really heavily on the hashed mirrors in our production setup.  Is it possible to make a simple nix binary cache that can *only* hold fixed-output derivations?

Ericson2314:
>  Is it possible to make a simple nix binary cache that can only hold fixed-output derivations?

@bhipple A previous version of this made a substituter out of the hashed mirrors code, so existing hashed mirrors still work. @edolstra didn't want that since regular binaries caches subsume it, but I think it would be great to resurrect that as a store plugin or something.

Ericson2314:
@bhipple https://github.com/NixOS/nix/pull/3673/files is the earlier version with the `HashedMirrorStore`.

PR #3698, opened 2020-06-15
WIP: Use std::string_view everywhere

Ericson2314:
As an optimization I couldn't care less about `const std::string &` vs `std:string_view`. However, `std:string_view` matches Rust's `&str`, which is used far more than `&String`, so I figure this is good for FFI, which is more likely to be worth it.

Still, this is a real slog so I'll not keep working on it unless told this is wanted.


edolstra:
Looks good to me!

Ericson2314:
OK everything builds (!), but there are some bugs I need to work out. I will also find myself some extra feedback on style/idioms.

Ericson2314:
I think this should wait for Rust haha :)

PR #3702, opened 2020-06-16
Rewrite StorePath class in C++

edolstra:
This PR rewrites the Rust implementation of `StorePath` in C++. `StorePath` was an experiment in Rust/C++ interop. However, it's currently the *only* bit of Rust in Nix, and depending on a Rust compiler for a fairly trivial class is rather wasteful. (Originally we also used Rust for tarfile processing, but that has been replaced by libarchive.)

We also get a minor speed up and maximum RSS reduction from using a different internal representation of `StorePath`.

I also replaced `storePathToHash()` by the more efficient `StorePath::hashPart()`.

### DIFF
--- a/src/libstore/path.hh
+++ b/src/libstore/path.hh
@@ -1,59 +1,78 @@
 #pragma once
 
-#include "rust-ffi.hh"
+#include "types.hh"
 
 namespace nix {
 
-/* See path.rs. */
-struct StorePath;
-
 class Store;
+struct Hash;
 
-extern "C" {
-    void ffi_StorePath_drop(void *);
-    bool ffi_StorePath_less_than(const StorePath & a, const StorePath & b);
-    bool ffi_StorePath_eq(const StorePath & a, const StorePath & b);
-    unsigned char * ffi_StorePath_hash_data(const StorePath & p);
-}
-
-struct StorePath : rust::Value<3 * sizeof(void *) + 24, ffi_StorePath_drop>
+class StorePath
 {
+    std::string baseName;
+
+    StorePath(const StorePath & path)
+        : baseName(path.baseName)
+    { }
+
+public:
+
+    /* Size of the hash part of store paths, in base-32 characters. */
+    constexpr static size_t HashLen = 32; // i.e. 160 bits
+
     StorePath() = delete;
 
-    static StorePath make(std::string_view path, std::string_view storeDir);
+    StorePath(std::string_view baseName);
 
-    static StorePath make(unsigned char hash[20], std::string_view name);
+    StorePath(const Hash & hash, std::string_view name);
 
-    static StorePath fromBaseName(std::string_view baseName);
+    StorePath(StorePath && path)
+        : baseName(std::move(path.baseName))
+    { }
 
-    rust::String to_string() const;
+    StorePath & operator = (StorePath && path)
+    {
+        baseName = std::move(path.baseName);
+        return *this;
+    }
+
+    std::string_view to_string() const
+    {
+        return baseName;
+    }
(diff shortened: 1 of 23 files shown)


grahamc:
I'm sad to see going back to c++ and removing rust from the build tooling altogether. To me, it seemed like a good foot-hold to build on top of, even if it is small and currently wasteful. I suppose next steps on future Rust work would require a more extensive project plan?

edolstra:
The problem with a Rust migration is that it would take years (if it's ever completed), and in the meantime only people who know Rust *and* C++ can contribute to Nix. So without a viable migration plan, it's better to stick to C++ for now.

Ericson2314:
I think with a dedicated sprint we could have done it in, say, 3 months. But I do agree it's better to stick to one language if we are not so sprinting----let's not pay the polyglot tax until will commit to making it temporary.

Ericson2314:
Re @grahamc the plan I would like to see is:

1. Refactor C++ to make more Rust-like with C++17 ideoms (e.g. `std::optional`, `std::variant`)
2. Quickly and dumbly rewrite line-by-line, initial rewrite should have no interesting changes, but `std::variant` can become Rust `enum`
3. Continue cleaning up once everything is in Rust.

I am not sure much more *planning* is needed than that. I would put greater emphasis on *staffing* than project planning, so step 2 gets done quickly.

src/libstore/path.hh
```diff
@@ -1,59 +1,78 @@
...
     {
-        return !(*this == other);
+        return baseName != other.baseName;
     }
 
-    StorePath clone() const;
+    StorePath clone() const
+    {
+        return StorePath(*this);
+    }
```
Ericson2314, line 62: I hope we can get rid of this and go back to just using idiomatic copy constructors, and STL copying in particular?

Ericson2314:
Maybe we can keep a `rust` branch in this repo without this, and with https://github.com/NixOS/nix/pull/3671 and  https://github.com/NixOS/nix/compare/master...Ericson2314:more-rust-ffi, to remind/inspire people to someday do such a thing.

src/libstore/path.hh
edolstra: Yes, but I didn't want to clutter this PR to much. Will do that afterwards.

Ericson2314: Sorry, I meant to say "after" and didn't. Great plan, happy to do some of that for you if you like :).

Kloenk:
Maybe the [minimal daemon](https://github.com/NixOS/rfcs/pull/68) could be the start of changing to rust?

Ericson2314:
@Kloenk :) Well, `libnixstore` is still the biggest library in Nix I think. I imagine working bottom-up with `libnixutils` and top-down with `libexpr` could maybe work, switching tactics whenever one gets stuck.

maisiliymBridj:
> I'm sad to see going back to c++ and removing rust from the build tooling altogether. To me, it seemed like a good foot-hold to build on top of, even if it is small and currently wasteful. I suppose next steps on future Rust work would require a more extensive project plan?

Im currently working on a project I call `uniks` (it has also been named `myx` previously) which is a subset of a bigger project called sajban, in rust, and `uniks` is essentially a superman version of nix. I should have a draft release within weeks. I mostly hang out in #sajban-dev:matrix.org on matrix. I wont be binding to nix with C++, as you have already discovered for yourselves it is not a very good approach. Im currently heavily relying on flakes, adding my own type data in each, along with `toJSON` and `fromJSON` ... just wish `derivation` had a `stdin` argument to pass data to the builder through stdin.

PR #3703, opened 2020-06-16
Add tests for libfetchers/attrs.hh

gilligan:
Add tests for the json/attr functions

### DIFF
--- a/src/libfetchers/tests/attrs.cc
+++ b/src/libfetchers/tests/attrs.cc
@@ -1,0 +1,286 @@
+#include "attrs.hh"
+#include "error.hh"
+
+#include <nlohmann/json.hpp>
+#include <gtest/gtest.h>
+
+using json = nlohmann::json;
+
+namespace nix::fetchers {
+
+    /* ----------------------------------------------------------------------------
+     *jsonToAttrs 
+     * --------------------------------------------------------------------------*/
+
+    TEST(jsonToAttrs, simpleJson) {
+        json j = {
+            {"num", 42 },
+            {"string", "this is a string" },
+            {"bool", true }
+        };
+
+        Attrs attrs = jsonToAttrs(j);
+        auto it = attrs.find("num");
+        ASSERT_NE(it, attrs.end());
+        ASSERT_EQ(std::get<int64_t>(it->second), 42);
+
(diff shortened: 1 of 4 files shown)


src/libfetchers/tests/attrs.cc
```diff
@@ -0,0 +1,284 @@
...
+        ASSERT_NE(it, attrs.end());
+        ASSERT_EQ(std::get<int64_t>(it->second), 42);
+
+        it = attrs.find("string");
+        ASSERT_NE(it, attrs.end());
+        ASSERT_EQ(std::get<std::string>(it->second), "this is a string");
+
+        it = attrs.find("bool");
+        ASSERT_NE(it, attrs.end());
+        //ASSERT_EQ(std::get<Explicit<bool>>(it->second).t, true); // unexpected index exception
```
gilligan, line 31: I could use some help with this. I really don't understand what's going on here and why i am failing to get the bool value out of there

src/libfetchers/tests/attrs.cc
```diff
@@ -0,0 +1,284 @@
...
+        Attrs attrs = {
+          { "num", 42 },
+          { "string", "this is a string" },
+          { "bool", b }
+        };
+
+        std::map<std::string, std::string> expected = {
+            { "num", "42" },
+            { "string", "this is a string" },
+            { "bool", "1" }
```
gilligan, line 279: I do wonder why the boolean is being converted to `"1"` or `"0"` here instead of `true` or `false`..

edolstra: IIRC this is fixed on the flakes branch, which rewrites most of libfetchers. Please rebase on that branch. BTW, that's an example of the cost of unit tests: you end up writing a lot of code that has to be discarded when the API changes.

gilligan: > BTW, that's an example of the cost of unit tests: you end up writing a lot of code that has to be discarded when the API changes.

I think Nix might be the only current project that I know where unit tests are regarded a controversial topic. 

The way everyone else seems to look at it is that unit tests are a net win because they provide confidence when making changes and thus effectively making API changes easier and safer.

PR #3704, opened 2020-06-16
Add another missing #include

Ericson2314:
Waiting for CI to make sure I catch everything.

### DIFF
--- a/src/libutil/config.hh
+++ b/src/libutil/config.hh
@@ -1,3 +1,4 @@
+#include <cassert>
 #include <map>
 #include <set>
 


edolstra:
Note that the Hydra `coverage` job builds with precompiled headers disabled, so it should catch this.

Ericson2314:
I think the GitHub CI does too, in which case the issue must be something like master changing from when the CI was run.

Ericson2314:
I removed the WIP from this in any case, now that I verified the coverage job myself.

edolstra:
Maybe it would be better to just disable precompiled headers in the regular job. Running coverage just for that would be a bit expensive.

Ericson2314:
I'll look into that, then.

PR #3707, opened 2020-06-17
replaced uncaught_exception with uncaught_exceptions

p01arst0rm:
small fix, uncaught_exception is deprecated in favour of uncaught_exceptions.
nothing huge.

### DIFF
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -228,7 +228,7 @@ struct ConnectionHandle
 
     ~ConnectionHandle()
     {
-        if (!daemonException && std::uncaught_exception()) {
+        if (!daemonException && std::uncaught_exceptions()) {
             handle.markBad();
             debug("closing daemon connection because of an exception");
         }
--- a/src/libutil/json.cc
+++ b/src/libutil/json.cc
@@ -173,7 +173,7 @@ JSONObject JSONPlaceholder::object()
 
 JSONPlaceholder::~JSONPlaceholder()
 {
-    assert(!first || std::uncaught_exception());
+    assert(!first || std::uncaught_exceptions());
 }
 
 }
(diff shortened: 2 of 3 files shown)


Ericson2314:
https://en.cppreference.com/w/cpp/error/uncaught_exception It's removed in C++20, in fact.

edolstra:
We had to revert this a couple of months ago because it didn't work on macOS yet: 93b1ce1ac522ff92626da5cdd689008063935844. Maybe it works now.

Ericson2314:
Looks like it is passing CI? Maybe switch to 20.03 did the trick.

edolstra:
Could be, but the Hydra builders are not running Catalina yet.

Ericson2314:
The error message says "introduced in macOS 10.12", so I think just sierra is sufficient? I have no idea what our support window is, however.

edolstra:
I think the problem is that the headers are too old, or something.

Ericson2314:
In that case, since we are using a nixpkg-provided headers, the builder OS version shouldn't matter so much?

edolstra:
Probably, but I don't know how pure the stdenv on darwin is these days.

Ericson2314:
I'll go check.

matthewbauer:
It might help to add `-mmacosx-version-min=10.12` to CFLAGS. I didn't think it was necessary since we set MACOSX_DEPLOYMENT_TARGET elsewhere.

Ericson2314:
Hmm, we have been setting `MACOSX_DEPLOYMENT_TARGET` to something new enough since github.com/nixos/nixpkgs/commit/f1070eec27e889566d441d90d54dc3619ce6c91e . Maybe it was the headers version.

In any event, I think we should set up a hydra job for just this branch, or just merge it and one of us can monitor hydra to see what happens.

edolstra:
Okay, we'll see what happens :-)

p01arst0rm:
#WorksForMe

PR #3709, opened 2020-06-17
Mention number of derivations to be build/fetched in output

expipiplus1:
Also correct grammar for the case of a single derivation.

### DIFF
--- a/doc/manual/advanced-topics/post-build-hook.xml
+++ b/doc/manual/advanced-topics/post-build-hook.xml
@@ -122,7 +122,7 @@ post-build-hook = /etc/nix/upload-to-cache.sh
 
   <screen>
 $ nix-build -E '(import &lt;nixpkgs&gt; {}).writeText "example" (builtins.toString builtins.currentTime)'
-these derivations will be built:
+this derivation will be built:
   /nix/store/s4pnfbkalzy5qz57qs6yybna8wylkig6-example.drv
 building '/nix/store/s4pnfbkalzy5qz57qs6yybna8wylkig6-example.drv'...
 running post-build-hook '/home/grahamc/projects/github.com/NixOS/nix/post-hook.sh'...
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -48,7 +48,10 @@ void printMissing(ref<Store> store, const StorePathSet & willBuild,
     unsigned long long downloadSize, unsigned long long narSize, Verbosity lvl)
 {
     if (!willBuild.empty()) {
-        printMsg(lvl, "these derivations will be built:");
+        if (willBuild.size() == 1)
+            printMsg(lvl, fmt("this derivation will be built:"));
+        else
+            printMsg(lvl, fmt("these %d derivations will be built:", willBuild.size()));
         auto sorted = store->topoSortPaths(willBuild);
         reverse(sorted.begin(), sorted.end());
         for (auto & i : sorted)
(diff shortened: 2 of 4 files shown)


expipiplus1:
Produces output like

```
$ nix-shell -p kakoune
  /nix/store/a48gid3azj821qvxa88xz821w144zas2-kakoune-2020.01.16.drv
these 2 paths will be fetched (0.98 MiB download, 4.19 MiB unpacked):
  /nix/store/b2q8105b1ba1zyw3isl6vfl23xr3v26n-bash-4.4-p23-dev
  /nix/store/f6sg1j79w45rs2wa0acgqfqg7ralx5jb-kakoune-unwrapped-2020.01.16
copying path '/nix/store/b2q8105b1ba1zyw3isl6vfl23xr3v26n-bash-4.4-p23-dev' from 'https://cache.nixos.org'...
copying path '/nix/store/f6sg1j79w45rs2wa0acgqfqg7ralx5jb-kakoune-unwrapped-2020.01.16' from 'https://cache.nixos.org'...
building '/nix/store/a48gid3azj821qvxa88xz821w144zas2-kakoune-2020.01.16.drv'...
```

I've sometimes wanted to know the number of derivations/paths in big lists.

The `these -> this` change was to avoid writing `these 1 derivations...`.

src/libmain/shared.cc
```diff
@@ -48,15 +48,21 @@ void printMissing(ref<Store> store, const StorePathSet & willBuild,
...
         for (auto & i : sorted)
             printMsg(lvl, fmt("  %s", store->printStorePath(i)));
     }
 
     if (!willSubstitute.empty()) {
-        printMsg(lvl, fmt("these paths will be fetched (%.2f MiB download, %.2f MiB unpacked):",
+        printMsg(lvl, fmt("%s will be fetched (%.2f MiB download, %.2f MiB unpacked):",
+                willSubstitute.size() == 1
+                    ? "this path"
+                    : fmt("these %d paths",willSubstitute.size()),
```
edolstra, line 65: BTW if we ever want to add translations to Nix, then splitting strings in this way is annoying, because in many languages the translation of "will be fetched" depends on the number of the subject. So it's better to write
```
if (willSubstitute.size() == 1)
    printMsg(lvl, fmt("this path will be fetched...", ...);
else
    printMsg(lvl, fmt("these %d paths will be fetched...", ...);
```
This is also easier to read IMHO.

doc/manual/advanced-topics/post-build-hook.xml
```diff
@@ -122,7 +122,7 @@ post-build-hook = /etc/nix/upload-to-cache.sh
 
   <screen>
 $ nix-build -E '(import &lt;nixpkgs&gt; {}).writeText "example" (builtins.toString builtins.currentTime)'
-these derivations will be built:
+this derivation will be built:
```
edolstra, line 125: It seems to me that in this case we should just omit the entire `this derivation will be built` message because it's superfluous. After all the next line already says `building '/nix/store/s4pnfbkalzy5qz57qs6yybna8wylkig6-example.drv'...`.

expipiplus1: In the case where there is a derivation being built and some being fetched, it distinguishes the build expression from the one being fetched.

src/libmain/shared.cc
expipiplus1: That is indeed how I had it initially, but I was unhappy with the duplication this incurred.

expipiplus1: Fixed

doc/manual/advanced-topics/post-build-hook.xml
expipiplus1: Also, the derivation may not be displayed promptly if one has to be "waiting for locks or build slots..."

expipiplus1:
Thanks!

PR #3710, opened 2020-06-17
Reserve the `__contentAddressed` derivation parameter

regnat:
Not implementing anything here, just throwing an error if a derivation sets `__contentAddressed = true` without `--experimental-features content-addressed-paths` (and also with it as there's nothing implemented yet)

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -809,9 +809,6 @@ private:
     /* Whether this is a fixed-output derivation. */
     bool fixedOutput;
 
-    /* Whether this is a content adressed derivation */
-    bool contentAddressed = false;
-
     /* Whether to run the build in a private network namespace. */
     bool privateNetwork = false;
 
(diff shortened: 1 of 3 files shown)


src/libstore/build.cc
```diff
@@ -1195,6 +1198,14 @@ void DerivationGoal::haveDerivation()
 
     parsedDrv = std::make_unique<ParsedDerivation>(drvPath, *drv);
 
+    contentAddressed = parsedDrv->contentAddressed();
+
+    if (this->contentAddressed) {
```
edolstra, line 1203: Don't need `this->` here.

src/libstore/build.cc
```diff
@@ -809,6 +809,9 @@ class DerivationGoal : public Goal
     /* Whether this is a fixed-output derivation. */
     bool fixedOutput;
 
+    /* Whether this is a content adressed derivation */
+    bool contentAddressed = false;
```
Ericson2314, line 813: Maybe not add this and use a temporary below in hopes that #3418 is merged soon?

regnat: Mh right. I've removed it in favor of directly calling `parsedDrv->contentAddressed()`

src/libstore/build.cc
regnat: Thanks, fixed

regnat:
> BTW shouldn't `content-addressed-paths` be `content-addressed-derivations` (or `ca-derivations` for consistency)? Since content-addressed paths are already partially supported, and there is a `ca-references` feature for full support.

You're definitely right. I didn't know about `ca-references`. I'll align with it

PR #3711, opened 2020-06-17
Use `ansicolor.hh` in `nix repl` rather than duplicates

Ericson2314:


Ma27:
The new ANSI color codes have a small syntax error and make the `;1m` directly printed to the console. Fixed in #3805.

PR #3712, opened 2020-06-17
Make successful states coherent

meditans:
The successful states used in these two places in the code were slightly
different. Should they be the same list?

### DIFF
--- a/src/libstore/filetransfer.cc
+++ b/src/libstore/filetransfer.cc
@@ -72,8 +72,7 @@ struct curlFileTransfer : public FileTransfer
 
         curl_off_t writtenToSink = 0;
 
-        inline static const std::set<long> successfulStatuses;
-
+        inline static const std::set<long> successfulStatuses {200, 201, 204, 206, 304, 0 /* other protocol */};
         /* Get the HTTP status code, or 0 for other protocols. */
         long getHTTPStatus()
         {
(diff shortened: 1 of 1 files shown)


src/libstore/filetransfer.cc
```diff
@@ -352,8 +354,7 @@ struct curlFileTransfer : public FileTransfer
             if (writeException)
                 failEx(writeException);
 
-            else if (code == CURLE_OK &&
-                (httpStatus == 200 || httpStatus == 201 || httpStatus == 204 || httpStatus == 206 || httpStatus == 304 || httpStatus == 0 /* other protocol */))
+            else if (code == CURLE_OK && successfulStatuses.find(httpStatus) != successfulStatuses.end())
```
edolstra, line 357: => `successfulStatuses.count(httpStatus)`

src/libstore/filetransfer.cc
```diff
@@ -72,6 +72,8 @@ struct curlFileTransfer : public FileTransfer
 
         curl_off_t writtenToSink = 0;
 
+        inline static const std::set<long> successfulStatuses;
```
edolstra, line 75: This looks like it lacks an initializer?

meditans: Thanks, fixed

src/libstore/filetransfer.cc
meditans: Thanks, fixed!

Ericson2314:
Thanks!

PR #3714, opened 2020-06-17
Add response body to network errors

Ericson2314:
@bburdette So we've gotten the new field into the error message after much wrangling, but I wasn't quite sure what to do next, and I thought it might be a good time to solicit your advice anyways.


bburdette:
Ok, so the response is a member of the FileTransferError exception.  Do you plan on the response showing up as part of the message the user sees when the FileTransferError is handled?  Or will it be used for something else, like the `error` member?   If its only for the message, you could put it in as a regular value in the string template, something like:

```
throw FileTransferError(Interrupted, "%s of '%s' was interrupted\nresponse was: %s", 
                                     request.verb(), request.uri, response));
```

But I'm guessing the purpose off the response field is something else, in which case LGTM.

Ericson2314:
I was thinking maybe just show it on higher verbosity levels. I was also reminded of you making `SysError` store the `errno`, rendering it on demand. I like it when structured errors and nicer messages are complementary.

bburdette:
I don't really have a plan as yet for varying amounts of error detail depending on verbosity, but you could do that in the FileTransferError constructor if you're building the hint message there.  SysError takes this both-ways approach of saving the errno and embedding it into the hint in its constructor, though without any verbosity checks.

Ericson2314:
Included #3720 as I needed that to be able to verify this. (My best source of 500 errors used escaped `/` in URLs.)

Ericson2314:
OK I did it just like `SysError`, storing the hint and the original value, and not bothering to look at verbosity as this sort of error is very rare.

Ericson2314:
I added a bunch more stuff to #3720, so I don't think it makes sense to block this on that just cause they contain a commit in common, hence the rename.

src/libstore/filetransfer.cc
```diff
@@ -824,6 +840,21 @@ void FileTransfer::download(FileTransferRequest && request, Sink & sink)
...
 }
 
+template<typename... Args>
+FileTransferError::FileTransferError(FileTransfer::Error error, std::shared_ptr<string> response, const Args & ... args)
+    : Error(args...), error(error), response(response)
+{
+    const auto hf = hintfmt(args...);
+    // FIXME: Due to https://github.com/NixOS/nix/issues/3841 we don't know how
+    // to print different messages for different verbosity levels. For now
+    // we add some heuristics for detecting when we want to show the response.
```
Ericson2314, line 850: @edolstra We got stuck trying to make it respect `-vvv`, and opened up https://github.com/NixOS/nix/issues/3841 in response. Perhaps the heuristic with no way to override is good enough for now.

Ericson2314:
OK CI is fixed.

PR #3718, opened 2020-06-18
Optimize function/primop calls

edolstra:
Functions can now take more than one argument and function applications can apply more than one argument at a time. For non-partially applied functions/primops, this avoids allocation of application thunks. Also, it allows merging of environments (e.g. in a function like `x: y: ...`, we only allocate one environment).

On `nix-env -qa -f '<nixpkgs>'`, this gives a ~8.5% speedup, reduces the number of environment allocations by 29.5% and the number of value allocations by 4.8%.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -118,6 +118,7 @@ static void printValue(std::ostream & str, std::set<const Value *> & active, con
         break;
     case tThunk:
     case tApp:
+    case tPartialApp:
         str << "<CODE>";
         break;
     case tLambda:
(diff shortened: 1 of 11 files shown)


andir:
This looks like an amazing improvement. Is there something that could be done to get this merged?

edolstra:
IIRC, there was some segfault. I can have a look at cherry-picking at least the commits that don't break anything.

andir:
5253cb4b68ad248f37b27849c0ebf3614e4f2777 seems to be fine applied onto the current git master branch. Could probably be extracted just as with those changes that were in #5501.

edolstra:
@andir Thanks, done in #5581.

PR #3720, opened 2020-06-18
Avoid `fmt` when constructor already does it

Ericson2314:
~~We have a larger problem that passing computed strings to the first
variable argument of many exception constructors is unsafe because that
first variable argument is interpreted not as a plain string, but format
string, and if it contains '%' boost::format will abort, since there are
no arguments to the format string.~~

~~In this particular instance '%' was used as part of an escape code in a
URL, which, when the download failed, caused Nix to abort displaying the
`FileTransferError`.~~

https://github.com/NixOS/nix/pull/3724 will fix the underlying issue. This is just a cleanup for brevity's sake. 

Most of the changes are very straight forward, but I did refactor a bit more aggressively with the SQLite errors, taking `SysError` my guide.

CC @bburdette


matthewbauer:
There are a few similar cases in:

- https://github.com/NixOS/nix/blob/6c000eed80565d83d596da800ca0db92e248342e/src/libutil/util.cc#L1020
- https://github.com/NixOS/nix/blob/6c000eed80565d83d596da800ca0db92e248342e/src/libutil/util.cc#L1147
- https://github.com/NixOS/nix/blob/6c000eed80565d83d596da800ca0db92e248342e/src/libfetchers/mercurial.cc#L200
- https://github.com/NixOS/nix/blob/6c000eed80565d83d596da800ca0db92e248342e/src/libmain/shared.cc#L52
- https://github.com/NixOS/nix/blob/6c000eed80565d83d596da800ca0db92e248342e/src/libstore/sqlite.cc#L21
- https://github.com/NixOS/nix/blob/6c000eed80565d83d596da800ca0db92e248342e/src/libstore/filetransfer.cc#L678

(most of those won't have a percent though, but best to be safe)

edolstra:
This is a regression compared to 2.3, where both `throw Error("foo %s")` and `throw Error(fmt("foo %s"))` printed `error: foo %s`. This is because we had a specialisation for `fmt`:
```c++
/* A helper for formatting strings. ‘fmt(format, a_0, ..., a_n)’ is
   equivalent to ‘boost::format(format) % a_0 % ... %
   ... a_n’. However, ‘fmt(s)’ is equivalent to ‘s’ (so no %-expansion
   takes place). */

inline std::string fmt(const std::string & s)
{
    return s;
}

...
```

We should restore that specialisation.

Ericson2314:
#3724 is better to fix the underlying issue, so I'll make this just a duplicate formatting cleanup.

Ericson2314:
Guess `clang` has different semantics around `protected` here.

Ericson2314:
Huh, this is the second PR of mine where there are weird perl issues.

edolstra:
What do the SQLite changes do?

Ericson2314:
@edolstra they shouldn't have any observable effect.

The idea is the helper function takes var args like the error constructors, formatiting the users format string. Also, the errors themselves store the original information (error number and extended error number) should somebody want to display the information differently.

With @bburdette's approval, I would complete the process of separating structure and presentation by making `SysError` and thse SQLite errors just store the user's original `hintformat` value, and then render the boilerplate stuff (with the `strerr` and similiar for the error numbers, and shortened message for the SQLite is busy) on demand when the exception is caught and rendered near `main`.

Ericson2314:
I think this is till a good cleanup. Rebased it.

src/libstore/sqlite.hh
```diff
@@ -96,10 +96,20 @@ struct SQLiteTxn
...
 
-MakeError(SQLiteError, Error);
-MakeError(SQLiteBusy, SQLiteError);
+struct SQLiteError : Error
+{
+    const char *path;
+    int errNo, extendedErrNo;
+
+    template<typename... Args>
+    [[noreturn]] static void throw_(sqlite3 * db, const std::string & fs, const Args & ... args);
```
thufschmitt, line 105: Unless I’m missing something, this is very fragile since the definition of the templated function isn’t in the header − so only the instances of the template that happen to be used in `sqlite.cc` will be available outside (which happens to be the case because the only usages outside of `sqlite.cc` seem to be whith `args` being a single string)

thufschmitt (changes requested):
> I think this is till a good cleanup

Indeed, thanks for the rebase :)

There’s just an issue with the sqlite stuff where you introduce a nasty little footgun. Other than that :+1:

src/libstore/sqlite.hh
Ericson2314: Oh good catch!

Ericson2314: Fixed this.

Ericson2314: Fixed it an even better way!

thufschmitt (approved):
Looks good now :)

Set to merge once the CI is green :heavy_check_mark:

orbekk:
I wonder if this may have broken the build ([run link](https://github.com/NixOS/nix/runs/6100064244?check_suite_focus=true):

```
error: builder for '/nix/store/diz7s7lkp2an9aa8vrwb225mbghgksfn-nix-2.9.0pre20220420_9345b4e.drv' failed with exit code 2;
       last 10 log lines:
       >   CXX    src/libstore/remote-store.o
       > src/libstore/sqlite.cc:28:20: error: calling a protected constructor of class 'nix::SQLiteBusy'
       >         auto exp = SQLiteBusy(path, err, exterr, std::move(hf));
       >                    ^
       > src/libstore/sqlite.hh:122:23: note: declared protected here
       > MakeError(SQLiteBusy, SQLiteError);
       >                       ^
       > 1 error generated.
       > make: *** [mk/patterns.mk:3: src/libstore/sqlite.o] Error 1
       > make: *** Waiting for unfinished jobs....
       For full logs, run 'nix log /nix/store/diz7s7lkp2an9aa8vrwb225mbghgksfn-nix-2.9.0pre20220420_9345b4e.drv'.
```

thufschmitt:
> I wonder if this may have broken the build

It did indeed :/ Looking at it right now

Ericson2314:
Thanks for fixing!

PR #3721, opened 2020-06-19
Try to substitute builtins.fetch*

matthewbauer:
builtins.fetch* aren’t technically derivations, so they never got
substituted. To work around this, we can try to substitute this store
path during evaluation. This can happen any time a hash is provided.

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -134,6 +134,19 @@ static void fetch(EvalState & state, const Pos & pos, Value * * args, Value & v,
     if (evalSettings.pureEval && !expectedHash)
         throw Error("in pure evaluation mode, '%s' requires a 'sha256' argument", who);
 
+    // try to substitute if we can
+    if (settings.useSubstitutes && expectedHash) {
+        auto substitutableStorePath = fetchers::trySubstitute(state.store, unpack, *expectedHash, name);
+        if (substitutableStorePath) {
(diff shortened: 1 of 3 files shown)


edolstra:
Please check the flakes branch, it already does this:
```c++
td::pair<Tree, Input> Input::fetch(ref<Store> store) const
{
    if (!scheme)
        throw Error("cannot fetch unsupported input '%s'", attrsToJson(toAttrs()));

    /* The tree may already be in the Nix store, or it could be
       substituted (which is often faster than fetching from the
       original source). So check that. */
    if (hasAllInfo()) {
        try {
            auto storePath = computeStorePath(*store);

            store->ensurePath(storePath);

            debug("using substituted/cached input '%s' in '%s'",
                to_string(), store->printStorePath(storePath));

            auto actualPath = store->toRealPath(storePath);

            return {fetchers::Tree(std::move(actualPath), std::move(storePath)), *this};
        } catch (Error & e) {
            debug("substitution of input '%s' failed: %s", to_string(), e.what());
        }
    }
```

matthewbauer:
Actually this just applies to fetchTree right? I think it would still be useful for builtins.fetchurl & builtins.fetchTarball. I can make it match what fetchTree does though.

Ericson2314:
I get the other `builtins.fetch*` in https://github.com/NixOS/nix/pull/3754/files, but this might be still useful as an intermediate step.

matthewbauer:
Since fetchTree can do substitution on tarballs, I think we are mostly covered just doing fetchTree everywhere.

PR #3724, opened 2020-06-19
Hintfmt percent test, and fix

bburdette:
In light of the issue from https://github.com/NixOS/nix/pull/3720, I thought adding some tests would be a good idea.  

Also there's a tentative solution here... its maybe a little inelegant to search-and-replace on percent, but its simpler than some other things I thought of.  Still thinking though.

### DIFF
--- a/src/libutil/fmt.hh
+++ b/src/libutil/fmt.hh
@@ -120,6 +120,13 @@ public:
         return *this;
     }
 
+    template<class T>
+    hintformat& operator%(const normaltxt<T> &value)
+    {
+        fmt % value.value;
+        return *this;
+    }
+
     std::string str() const
     {
         return fmt.str();
@@ -142,6 +149,6 @@ inline hintformat hintfmt(const std::string & fs, const Args & ... args)
 inline hintformat hintfmt(std::string plain_string)
 {
     // we won't be receiving any args in this case, so just print the original string
-    return hintfmt("%s", plain_string);
+    return hintfmt("%s", normaltxt(plain_string));
 }
 }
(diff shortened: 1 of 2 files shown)


src/libutil/fmt.hh
```diff
@@ -136,4 +139,11 @@ inline hintformat hintfmt(const std::string & fs, const Args & ... args)
     return f;
 }
 
+inline hintformat hintfmt(std::string fs)
+{
+    // we won't be receiving any args in this case, so escape all percents.
+    boost::replace_all(fs, "%", "%%");
+    hintformat f(fs);
+    return f;
+}
```
Ericson2314, line 148: Nit, but I think I rather do
```suggestion
inline hintformat hintfmt(std::string plain_string)
{
    // we won't be receiving any args in this case, so just print the original string
    return hintformat { "%s", plain_string };
}
```
which has the slight advantage of not relying on the format string format, even though that format is unlikely to change.

src/libutil/fmt.hh
```diff
@@ -136,4 +139,11 @@ inline hintformat hintfmt(const std::string & fs, const Args & ... args)
     return f;
 }
 
+inline hintformat hintfmt(std::string fs)
+{
+    // we won't be receiving any args in this case, so escape all percents.
+    boost::replace_all(fs, "%", "%%");
```
edolstra, line 145: Hm, why is this needed? Can't we just do `return hintfmt("%s", fs)` here?

src/libutil/fmt.hh
bburdette: nice!  I like that better too.

PR #3726, opened 2020-06-21
flakes: Only check apps for the current system

B4dM4n:
When running `nix flake check` against a flake containing apps for unsupported systems, it will fail with an error like this:

```log
a 'aarch64-linux' with features {} is required to build '/nix/store/9wpnn4am8jc1vyp40ks6hizk3m53d4r0-hello.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
```

[example flake gist](https://gist.github.com/B4dM4n/7dec32408bd8355f6cf86381243cfd90)

This change will only check apps for the current system as done by the `checks` output.

CC @edolstra

### DIFF
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -241,13 +241,13 @@ struct CmdFlakeCheck : FlakeCommand
 
         std::vector<StorePathWithOutputs> drvPaths;
 
-        auto checkApp = [&](const std::string & attrPath, Value & v, const Pos & pos) {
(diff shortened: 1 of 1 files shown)


edolstra:
I realized that it's actually inconsistent to have `nix flake check` build `apps` at all since it doesn't build `packages` either. So I've removed building of `apps` in 334e26bfc2ce82912602e8a0f9f9c7e0fb5c3221. To test an app, it should be included in `checks.${system}`.

PR #3730, opened 2020-06-22
Improve hash parsing and errors

Ericson2314:
- Ensure hash is in form `<algo>-<prefix>` and not SRI.

 - Better errors if something goes wrong

 - `std::string_view` for no coppying


Ericson2314:
WIP because will also fix some `libfetchers` FIXMEs.

Ericson2314:
Hmm i cannot reproduce this test failure. It seems like the `echo -n '' > dummy` did not work.

Ericson2314:
CC @domenkozar tests failed saying there's no `nix-build`. No what that's about?

src/libutil/parser.hh
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
edolstra, line 1: Maybe rename this file to `split.hh`? Since "parser" is pretty generic.

src/libutil/hash.cc
```diff
@@ -136,63 +136,101 @@ std::string Hash::to_string(Base base, bool includeType) const
...
+Hash Hash::parseAnyPrefixed(std::string_view original)
+{
+    auto rest = original;
+    auto [optParsedType, isSRI] = getParsedTypeAndSRI(rest);
 
-    size_t size = s.size() - pos;
+    // Either the string or user must provide the type, if they both do they
+    // must agree.
+    if (!optParsedType)
+        throw BadHash("hash '%s' does not include a type.", rest);
```
edolstra, line 180: Style note: For consistency, no periods at the end of exception messages please.

Ericson2314:
OK, fixed both.

Ericson2314:
Fixed some more conflicts

Ericson2314:
Ah, it is already fixed on this branch.

Ericson2314:
Conflicts fixed.

Ericson2314:
@edolstra wiggled the metadata so GitHub would recompute the diff, and this PR is actually readable.

Ericson2314:
OK this should be ready.

Issue #3734, opened 2020-06-23
`--store ./relative/path` doesn't work


Ericson2314:
**Describe the bug**

Only absolute paths are recognized.

**Steps To Reproduce**

1. Try any nix command with `--store $(readlink -e ./relative/path)` that succeeds
2. Try same command with `--store ./relative/path` it fails not knowing what store to use

**Expected behavior**

it canonicalizes the path and makes a chroot store with that.

**`nix-env --version` output**

master

Ericson2314:
CC @matthewbauer

PR #3736, opened 2020-06-23
Allow relative paths in --store option

meditans:
In nix commands which accept --store options, we can now specify a
relative path, which will be canonicalized. The relative path must start
with `./` like in the `nix` language.

Fixes https://github.com/NixOS/nix/issues/3734


meditans:
It works, but it's marked as WIP as some tests should be added for this.

src/libstore/store-api.cc
```diff
@@ -864,7 +864,7 @@ StoreType getStoreType(const std::string & uri, const std::string & stateDir)
 {
     if (uri == "daemon") {
         return tDaemon;
-    } else if (uri == "local" || hasPrefix(uri, "/")) {
+    } else if (uri == "local" || hasPrefix(uri, "/") || hasPrefix(uri, "./")) {
```
regnat, line 867: Might make sense to also allow `.` as a path

meditans: The cost of that would be divergence from the nix expression language though, right?

matthewbauer: this already happens with cli args. for instance:

```
nix build -f . hello
```

Ericson2314: but does `-f <url>` work? Because we are overloading URL and relative path syntax, and URL syntax already is infamous for having parser mistakes/ambiguities/vulnerabilities, I rather move slowly. We can always add `.` later.

Of course, if @edolstra wants / there's concensus to go ahead and do `.`, we'll happily add it, but I rather advocate for the smallest change that gets the job done.

matthewbauer: url's do work (although they have to be in tar format):

```
nix build -f https://github.com/NixOS/nixpkgs/archive/master.tar.gz hello
```

matthewbauer: rules are:

https://github.com/NixOS/nix/blob/3f01fa1c9c1b5ce0ffd701f8eeb1b16e57aa86d5/src/libstore/filetransfer.cc#L827-L834

Ericson2314: OK fair, but it as wrong of me to compare `-f <url>` in the first place, because this sort of "each store parses the thing" means having `else` cases is no good either---everything should be `if (...) { ... } else return nullptr`.

Now we could still do `.`, yes. The above does not work argument against that.

Ericson2314: @edolstra's weak preference is if it doesn't contain `://` and does contain `/` then it is deemed a path and is turned into a LocalStore.

Ericson2314: In meeting we got @edolstra's preference, and I just updated the PR accordingly.

src/libstore/store-api.cc
```diff
@@ -914,12 +914,20 @@ ref<Store> openStore(const std::string & uri_,
     throw Error("don't know how to open Nix store '%s'", uri);
 }
 
+static bool isNonUriPath(const std::string & spec) {
```
matthewbauer, line 917: Can you just use !isUri from file-transfer?

Issue #3640, opened 2020-05-29
Make ValidPath (nar) `hash` optional if `ca` field is present


Ericson2314:
I figure as long as we have one way to verify a path's contents, we should be fine. Conferred with @LnL7, who thought not doing this was just a holdover from before the `ca` field existed.

This would be good to give other NAR alternatives slightly more equal footing, which generally sounds nice, but which I'd specifically be interested in because of the git object work for IPFS. It's by no means a blocker of that, of course.

matthewbauer:
note that one is base16, while the other is base32:

```
1|/nix/store/9krlzvny65gdc8s7kpb6lkx8cd02c25b-default-builder.sh|sha256:51858393c5698aba2e7c4e2832054a6985ab3eeda5db7373db5627d0acceefa8|1547485019||152|||fixed:r:sha256:1a7grsnd09snvdrp7nx5xlzap1b9982k4a2fghpbm2k9qn9q71ai
```

Ericson2314:
CC @zimbatm who was interested in multiple different hashing schemes.

PR #3738, opened 2020-06-23
Hash always has a valid type

Ericson2314:
@edolstra based on your passing remark on our call, we made some of the ValidPathInfo/NarInfo hashes `std::optional`, but we're not really sure what the correct "bussiness logic" is, so that will be important to review.

~~These two remaining test failures may also turn up interesting things.~~ **edit** nope, they weren't.

The parser changes use some of the same techniques I used in #3730 -- lots of `std::string_view` including a `rest` I mutate as I go----in general trying to write the code as if I had proper parser combinators at my disposal.

The rest is rote.


src/libstore/store-api.hh
```diff
@@ -115,7 +115,8 @@ struct ValidPathInfo
 {
     StorePath path;
     std::optional<StorePath> deriver;
-    Hash narHash;
+    // TODO document this
+    std::optional<Hash> narHash;
```
Ericson2314, line 119: Here is one of the now-optional hashes

src/libfetchers/tree-info.hh
```diff
@@ -11,7 +11,7 @@ namespace nix::fetchers {
 
 struct TreeInfo
 {
-    Hash narHash;
+    std::optional<Hash> narHash;
```
Ericson2314, line 14: Another

src/libstore/nar-info.hh
```diff
@@ -10,7 +10,7 @@ struct NarInfo : ValidPathInfo
 {
     std::string url;
     std::string compression;
-    Hash fileHash;
+    std::optional<Hash> fileHash;
```
Ericson2314, line 13: another

src/libfetchers/tree-info.hh
Ericson2314: Though I think this one is fine and just means the hash is calculted---obviously these builtins aren't going to be making non-CA paths.

src/libstore/nar-info.hh
Ericson2314: Per meeting this one is OK

src/libstore/store-api.hh
Ericson2314: @edolstra So I checked and while this one is optional, the constructor for NarInfo (derived class) requires that it is present with assertions. Is that good enough for now, or do you also want the refactor so that it is required everywhere? (I do recall you were saying the nix installer trick where the nar hash would be missing is long gone.)

(I must admit due to #3640 I am a bit partial to keeping it optional, but I'm happy to do whatever you like.)

src/libfetchers/tree-info.hh
edolstra: I think this one should not be optional. Anyway it's better to leave libfetchers unchanged since it's all rewritten in the flakes branch.

src/libstore/store-api.hh
edolstra: I think this one should *not* be optional since a store path should always have a NAR hash.

Ericson2314: This is actually turns out to be a fairly substantial change, because it looks like the `BinaryCacheStore::addToStore` overload that doesn't take the `ValidPathInfo` was uploading flat files as-is and calling them nars, with the `narHash` unset.

I think that a bug not intended feature, but either way, fixing it is the sort of major behavior change I rather do separately from this refactor which has no behavior changes?

src/libfetchers/tree-info.hh
Ericson2314: We thought it might be easy to fix this, but it isn't either: the caching stuff doesn't store the `narHash` and since the cache also doesn't distinguish between stores, doing a `queryPathInfo` to get the Nar hash in the case of a cache-hit doesn't work.

Perhaps something could be made to work, but like you said `libfetchers` is all rewritten anyways, and I think the 2 lines we changed in `libfetchers` with the `std::optional` is much more minimal than whatever it would take to make it work without the `std::optional`.

Ericson2314: This change did go away with the flakes branch being merged, yay!

edolstra:
There are some conflicts, could you have a look at those? Otherwise LGTM.

Ericson2314:
OK, fixed! Will get the PRs that include this too.

Ericson2314:
Again?! :D

Ericson2314:
Fixed. Guess I shouldn't complain as it's my own other PR which caused the conflict :)

B4dM4n:
This PR introduced a evaluation error for me when building the `niv` package:

```log
% nix run 'nix/86805a2c0a25f5ceefac0d64e64ba57ace73b7f5' -- build -f https://github.com/nmattia/niv/archive/c0a61e6283933fa6eed6d3451ffd86c9df5209ec.tar.gz -v --show-trace
error: --- Error ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix
store path mismatch in (possibly filtered) path added from '/nix/store/p46a908qp9pg5qdvm0v20kwisqixjdcq-source'
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- show-trace ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
trace: while evaluating 'sourceByRegex'
at: (8:30) in file: /nix/store/p46a908qp9pg5qdvm0v20kwisqixjdcq-source/default.nix

     7|
     8|   sourceByRegex = name: src: regexes:
      |                              ^
     9|     builtins.path {

trace: from call site
at: (22:16) in file: /nix/store/p46a908qp9pg5qdvm0v20kwisqixjdcq-source/default.nix

    21|
    22|   niv-source = sourceByRegex "niv" ./. [
      |                ^
    23|     "^package.yaml$"
```

The same build was successful with a5f7d310dd10fe86b6f6aa1c2771c30f113741d4 (one commit before):
```shell
nix run 'nix/a5f7d310dd10fe86b6f6aa1c2771c30f113741d4' -- build -f https://github.com/nmattia/niv/archive/c0a61e6283933fa6eed6d3451ffd86c9df5209ec.tar.gz -v --show-trace
```

Not sure if `niv` is using `builtins.path` in a wrong way ([default.nix#L9](https://github.com/nmattia/niv/blob/c0a61e6283933fa6eed6d3451ffd86c9df5209ec/default.nix#L9)) and it needs to be fixed on their side.

Ericson2314:
Thanks for reporting this, I do think it's a bug on the Nix side I'm sorry to have introduced. https://github.com/NixOS/nix/issues/3845 is the next step of this process, and I think will root out any renaming bugs, but if that can't be done fast enough we can separately investigate this one

B4dM4n:
The build failure is fixed in #3881. Thanks for your effort.

Ericson2314:
Actually was totally coincidental! (As far as I know.) Thanks, Matt!

PR #3740, opened 2020-06-24
Minimal CA support

regnat:
First very limited support:

- Only `nix-build` works properly with CA derivations (`nix-store`, `nix build`, etc.. will print and create a symlink to a non-existent output path)
- It isn't possible to depend on a content-addressed derivation

Mostly taken out of #3528 with some changes to reuse the logic already in `nix/make-content-addressable.cc`.
Builds on top of #3678 

/cc @edolstra @Ericson2314


Ericson2314:
It will probably be another day or two before I can properly read this, sorry.

PR #3744, opened 2020-06-24
Warn when experimental features are used

edolstra:
If you use an experimental feature not listed in the `experimental-features` setting, you now get a warning rather than a fatal error. This makes things like the 'nix' command a bit less obnoxious.

However, if you set `allow-experimental-features = false`, it's still a fatal error. This is primarily for CI systems where use of experimental features in the daemon could go unnoticed.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -2,6 +2,7 @@
 #include "util.hh"
 #include "archive.hh"
 #include "args.hh"
+#include "sync.hh"
 
 #include <algorithm>
 #include <map>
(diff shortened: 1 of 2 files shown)


Kloenk:
I don't think all experimental features should be enabled automatically. Maybe we should include `nix-command` in the default settings for experimental-features?

src/libstore/globals.cc
```diff
@@ -129,8 +130,14 @@ bool Settings::isExperimentalFeatureEnabled(const std::string & name)
 
 void Settings::requireExperimentalFeature(const std::string & name)
 {
-    if (!isExperimentalFeatureEnabled(name))
-        throw Error("experimental Nix feature '%1%' is disabled; use '--experimental-features %1%' to override", name);
+    if (!isExperimentalFeatureEnabled(name)) {
+        if (allowExperimentalFeatures) {
+            static Sync<std::unordered_set<std::string>> warned;
```
matthewbauer, line 135: this requires `#include <unordered_set>` for PRECOMPILE_HEADERS=0

colemickens:
Are there other experimental features now besides `nix-command` and `flakes`? This seems to undo most of the (friction|awareness) that was added by `--experimental-features` in the first place.

Is "`nix` gave warnings in  >=2.4" going to be enough cover for making breaking changes, as needed and desired? I was actually excited to see the required explicit opt-in, despite the hassle, as it gave me hope that Nix devs would feel free to improve upon the v2 `nix` cli.

edolstra:
Let's keep this unchanged for now.

Ericson2314:
Yes, I can imagine people complaining "but the warning was burried in a huge log, and I had no idea I was using an unstable feature". Thanks for not doing this!

domenkozar:
On the other hand, maintaining tutorials of any kind is going to be really painful, because now Nix 2.3 rejects `--experimental-features` flag while Nix 2.4 requires it.

edolstra:
We could backport the experimental-features flag to 2.3.

PR #3746, opened 2020-06-25
Introduce `StoreReferences` and `ContentAddressWithReferences`

Ericson2314:
# Change itself

## `StoreReferences`

This formalizes references where we don't necessarily know what we are called out, so self references are tracked with a `bool` not a `StorePath` in the set.

This is used with `FixedOutputInfo` (part of `StorePathDescriptor`) and internally the building logic.

## `ContentAddressWithReferences`

This is useful to make any sort of content-addressed store path, whether it has references or not. It also makes initializing `ValidPathInfo` easier, by making sure the `path` and `ca` fields are in sync.

This is a sum type, not a pair of `ContentAddress` and `References`, because different types of content-addressing as different rules about what sorts of references are allowed.

# Motivation

## Constructing `ValidPathInfo`s

The chief motivation today is constructing `ValidPathInfo`. Today, for content-addressed paths, we have to:

1. Compute the path from some CA info.
2. Set that path in the `ValidPathInfo`
3. Set a slightly different CA in the `ValidPathInfo`
4. Set references in the `ValidPathInfo`

It is easy for those steps where the same information must be provided again and again to drift out of sync.

With this change there is *one* data type for all content-addressed info (hashes and reference) that one can create with nice designated initializers (AKA explicit field names) and then use that to construct the `ValidPathInfo` all at once. This keeps everything in sync.

## Unblocking future feature work

 - #3959 leverages this to support text-hashed derivations. This is a key part of accepted RFC 92 https://github.com/NixOS/nix/issues/6316. 

 - #3754 leverages this to allow fetching from stores which only support content addressing. This was done as part of the IPFS / git hashing work, and something that we would hopefully return to with https://github.com/NixOS/rfcs/pull/133.

## Deduplicating the `addToStore` methods

We currently have many method variations to add store objects to a store. In https://github.com/NixOS/nix/pull/4030 @roberth deduplicated this some in a new revision of the remote store protocol. With this + #3959, we can do something similar to the `Store` interface itself, cutting down on boilerplate.

# Context

More philosophically, the idea we are inching towards is that `ValidPathInfo` and `NarInfo` are better as interfaces types (for talking to the SQLite DB and binary cache stores, respectively) than core data model types. For the CA world, `StorePathDescriptor` + `Realisations` is much better, as we are separating the trust free parts (what the data is) from the trustful parts (realisations, signatures etc.)

I *think* that even the older input-addressed code can be reformulated along these lines, but rest assured I would not attempt to relegate `ValidPathInfo` and `NarInfo` to the boundaries until I am confident I know how not make the input-addressed case worse.

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests

-----

Further refinements to the content addressing data types are made in #3959, so it might be good to check that for reference.


roberth (commented):
I suppose this might be a little bit nicer representation in cases where you're special-casing the self reference, but we now have on the order of 40 call sites that are copying the references set, regardless of whether a self-reference is present. I've underestimated the weight of references before, so perhaps I'm a bit too wary. There's generally an order of magnitude more of those than store paths though.
If you could find more cases where the self-reference can/should be omitted when reading, that would help a lot. If this PR helps a lot with clarity it's also a win.

Found some surface level things, but this is not a proper review :)

Ericson2314:
> I suppose this might be a little bit nicer representation in cases where you're special-casing the self reference, but we now have on the order of 40 call sites that are copying the references set, regardless of whether a self-reference is present. I've underestimated the weight of references before, so perhaps I'm a bit too wary. There's generally an order of magnitude more of those than store paths though.

That makes sense. I'm happy to just `std::shared_ptr` everything.

src/libstore/content-address.hh
```diff
@@ -2,14 +2,20 @@
 
 #include <variant>
 #include "hash.hh"
+#include "path.hh"
 
 namespace nix {
 
+/*
+ * Mini content address
```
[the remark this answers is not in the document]
Ericson2314: This was supposed to be a section heading of sorts. I should find the other commit that makes that clearer.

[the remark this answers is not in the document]
Ericson2314: Found it (actually in another PR and rightly so, but I borrowed it's organization. This is not cleaned up.

Ericson2314:
@roberth upon further thought, I think all the `referencesToSelf` stuff is just used at interface boundaries where there is already some `O(n)` traversal, so this at last no worse than a constant factor slowdown. Does that sound right to you?

Also (as I've amended the PR description to say), consider checking out #3959 where further change the content address types. If those changes look good to you, I might back-port them to this PR so that one can just be about the derivations.

I'm happy to make the constructor change and rename fields, but I think I want to confirm with @edolstra before I spend the time making those sorts of final touches.

roberth:
> ... constant factor slowdown. Does that sound right to you?

It seems to improve code quality, so I don't think it has to be an issue. It's not I/O so I think we have some constant factor budget to spend. I'll assume most copies are short-lived as well, so memory footprint shouldn't be an issue, and if I assume wrongly, they were already copied before this PR, because that's the only way to retain them.

src/libstore/content-address.hh
```diff
@@ -74,4 +76,96 @@ ContentAddressMethod parseContentAddressMethod(std::string_view rawCaMethod);
...
 std::string renderContentAddressMethod(ContentAddressMethod caMethod);
 
+/*
+ * References set
+ */
+
+template<typename Ref>
+struct PathReferences
+{
+    std::set<Ref> references;
```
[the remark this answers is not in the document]
Ericson2314: Yeah me original names sucked. Fixed now.

src/libstore/path-info.hh
```diff
@@ -104,6 +103,9 @@ struct ValidPathInfo
     ValidPathInfo(StorePath && path, Hash narHash) : path(std::move(path)), narHash(narHash) { };
     ValidPathInfo(const StorePath & path, Hash narHash) : path(path), narHash(narHash) { };
 
+    ValidPathInfo(const Store & store,
+        StorePathDescriptor && ca, Hash narHash);
+
```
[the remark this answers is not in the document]
Ericson2314: So `Store` is here only used to get the store dir to compute a path from this info. "Valid" should probably just be dropped from the name, there are no special invariants here beyond what meets the eye (e.g. if something says "nar hash", it should be a hash of some nar, etc.).

https://github.com/NixOS/nix/pull/6236 would factor out a data type for "just" these store path functions, precisely so we can make clearer in the types for situations like this when some powerful IO might be done vs just a little more configure information is needed.

Ericson2314:
I started making a custom iterator to avoid some copies, but it was too much work, so I relent :). The `ValidPathInfo` info changes are reverted; I can always return to them later.

src/libstore/path-info.hh
```diff
@@ -78,15 +77,11 @@ struct ValidPathInfo
...
     /* Return true iff the path is verifiably content-addressed. */
     bool isContentAddressed(const Store & store) const;
 
-    /* Functions to view references + hasSelfReference as one set, mainly for
-       compatibility's sake. */
-    StorePathSet referencesPossiblyToSelf() const;
-    void insertReferencePossiblyToSelf(StorePath && ref);
-    void setReferencesPossiblyToSelf(StorePathSet && refs);
-
```
Ericson2314, line 89: These don't exist, and made it into this header by mistake (my fault long ago! As it so happens due to an earlier version of this PR.)

src/libstore/store-api.hh
```diff
@@ -87,6 +87,8 @@ enum BuildMode { bmNormal, bmRepair, bmCheck };
 struct BuildResult;
 
 
+typedef std::map<StorePath, std::optional<ContentAddress>> StorePathCAMap;
```
Ericson2314, line 90: Was moved from `path.hh` where it is unused and doesn't belong.

src/libstore/legacy-ssh-store.cc
```diff
@@ -134,7 +134,6 @@ struct LegacySSHStore : public virtual LegacySSHStoreConfig, public virtual Stor
             /* Hash will be set below. FIXME construct ValidPathInfo at end. */
             auto info = std::make_shared<ValidPathInfo>(path, Hash::dummy);
 
-            PathSet references;
```
Ericson2314, line 137: Dead code (I think)

src/nix/verify.cc
```diff
@@ -81,14 +81,14 @@ struct CmdVerify : StorePathsCommand
 
         ThreadPool pool;
 
-        auto doPath = [&](const Path & storePath) {
+        auto doPath = [&](const StorePath & storePath) {
```
Ericson2314, line 84: A cleanup to avoid print-parse round trips that would be good regardless

src/libstore/store-api.cc
```diff
@@ -742,13 +760,13 @@ StorePathSet Store::queryValidPaths(const StorePathSet & paths, SubstituteFlag m
     std::condition_variable wakeup;
     ThreadPool pool;
 
-    auto doQuery = [&](const Path & path) {
+    auto doQuery = [&](const StorePath & path) {
```
Ericson2314, line 763: More misc improvement to do either way.

Ericson2314:
#7717 and #7718 shrink the diff

Ericson2314:
Thanks @edolstra and @thufschmitt for reviewing all the little PRs that shrink the diff. I've merged master back into this branch and touched up a few more things --- I think the diff is now as minimal as its going to get.

Ericson2314:
Oh hmm the recent switch to C++20 is messing up my designated initializers, will take a look.

src/libstore/content-address.hh
```diff
@@ -55,23 +89,47 @@ std::optional<ContentAddress> parseContentAddressOpt(std::string_view rawCaOpt);
...
-  > ContentAddressMethod;
+/*
+ * Full content address
+ *
+ * See the schema for store paths in store-api.cc
+ */
 
-ContentAddressMethod parseContentAddressMethod(std::string_view rawCaMethod);
+// This matches the additional info that we need for makeTextPath
+struct TextInfo : TextHash {
```
thufschmitt, line 114: I think using inheritance here isn't really nice. It's easy-enough to confuse one for another without adding some implicit conversion between them. Can't we have a proper subfield instead?

```suggestion
struct TextInfo {
    TextHash hash;
```

(same goes for [FixedOutputInfo](https://github.com/NixOS/nix/pull/3746/files#diff-42e44a338fd5db8e9cd1f43e4522641e0fe8300bff8d58a8c7edd6ac82f827e8R121) ofc)

thufschmitt: Also, that would allow properly using aggregate initialization, which would be a huge win for readability in the rest of the code

src/libstore/path-info.cc
```diff
@@ -21,25 +21,45 @@ void ValidPathInfo::sign(const Store & store, const SecretKey & secretKey)
     sigs.insert(secretKey.signDetached(fingerprint(store)));
 }
 
-
-bool ValidPathInfo::isContentAddressed(const Store & store) const
+std::optional<ContentAddressWithReferences> ValidPathInfo::contentAddressWithReferenences() const
```
thufschmitt, line 24:
```suggestion
std::optional<ContentAddressWithReferences> ValidPathInfo::contentAddressWithReferences() const
```

thufschmitt (commented):
I left a few comments, but that seems correct overall.

The thing I'm not clear on, however, is the point of the change. It's adding a lot of accidental complexity, and I don't see the advantages. Care to elaborate a bit on that?

src/libstore/store-api.cc
```diff
@@ -162,73 +163,80 @@ StorePath Store::makeOutputPath(std::string_view id,
 }
 
 
+/* Stuff the references (if any) into the type.  This is a bit
+   hacky, but we can't put them in `s' since that would be
+   ambiguous. */
```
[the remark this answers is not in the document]
Ericson2314: Heh, I had to remember too what I was thinking 2 years ago!

Hopefully it is starting to make more sense now, but happy to improve it further.

src/libstore/content-address.hh
Ericson2314: Yeah when I first did this, I could use flat designated initializers in GCC (in a non-standard way which I didn't realize) which was nice. Now absolutely the inheritance is not pulling its weight.

Ericson2314: Done. I wonder also if it might make sense to just inline the type to avoid `hash.hash` and make the designated initializers nicer.

Kind of an evil thought (flattening and duplicating data definitions for sake of nice syntax), but very tempting :)

Ericson2314:
> The thing I'm not clear on, however, is the point of the change. It's adding a lot of accidental complexity, and I don't see the advantages. Care to elaborate a bit on that?

Sure.

First of all, does the part about trying to initialize a `ValidPathInfo` in one go make sense? That is the most "self-contained" motivation.

As to the other part, a problem I see today is that the fixed output hashing and text hashing uses very separate code paths despite being basically two instances of the same idea:
 - separate functions to construct store paths
 - separate functions to add to store

Meanwhile, we a good bit of duplicated code in `make-content-addressed.cc` and `local-derivation-goal.cc`, as you observed two years ago.

Finally, my RFCs, without further cleanups, would make this problem worse:
- RFC 92 means the `local-derivation-goal.cc` stuff must also support text hashing.
- RFC 133 adds git hashing, which means instead of 2 similar things we now have 3, and even more duplication if we stay the current course.

This RFC doesn't fix these latter problems by itself --- it is just added more data structures. But i think in doing so it sets us up to.
 - https://github.com/NixOS/nix/pull/3959 allowing text hashing in more places becomes quite simple.
 - I would continue what @roberth started in https://github.com/NixOS/nix/pull/4030, getting rid of `addTextToStore` (or at least making it a trivial non-virtual wrapper).

-----

The simplest thing to do might be just looking at #3959 to see the extant instance of this stuff being put to use.

src/libstore/binary-cache-store.cc
```diff
@@ -306,11 +306,21 @@ StorePath BinaryCacheStore::addToStoreFromDump(Source & dump, std::string_view n
...
+            *this,
+            name,
+            FixedOutputInfo {
+                .hash = {
+                    .method = method,
+                    .hash = nar.first,
+                },
+                .references = {
+                    .others = references,
+                    .self = false,
```
roberth, line 318:
```suggestion
                    // caller is not capable of creating a self-reference, because this is content-addressed without modulus
                    .self = false,
```

Is this correct?

src/libstore/binary-cache-store.cc
```diff
@@ -414,15 +424,21 @@ StorePath BinaryCacheStore::addToStore(
...
+            *this,
+            name,
+            FixedOutputInfo {
+                .hash = {
+                    .method = method,
+                    .hash = h,
+                },
+                .references = {
+                    .others = references,
+                    .self = false,
```
roberth, line 436:
```suggestion
                     // caller is not capable of creating a self-reference, because this is content-addressed without modulus
                    .self = false,
```

src/libstore/local-store.cc
```diff
@@ -1320,7 +1321,18 @@ StorePath LocalStore::addToStoreFromDump(Source & source0, std::string_view name
...
 
-    auto dstPath = makeFixedOutputPath(method, hash, name, references);
+    ContentAddressWithReferences desc = FixedOutputInfo {
+        .hash = {
+            .method = method,
+            .hash = hash,
+        },
+        .references = {
+            .others = references,
+            .self = false,
```
Ericson2314, line 1331:
```suggestion
            // caller is not capable of creating a self-reference, because this is content-addressed without modulus
            .self = false,
```

src/libstore/store-api.hh
roberth, line 1: :heart:

roberth (approved):
Independently glad that you make these code improvements and that we're making a good step towards dynamic derivations, which is crucial for the sustainability of Nixpkgs.

PR #3754, opened 2020-06-26
WIP: FullContentAddress --- contains #3721 and #3746

Ericson2314:
This adds a new type `FullContentAddress`, which differs from `ContentAddress` (now renamed `MiniContentAddress`) in that it contains everything needed to create a store path. As such, it is like a combination of the ValidPathInfo "CA" and "References" fields, but without the other extraneous metadata.

The motivation is multi-fold:

1. Be more cautious with references. When we just pass around `ContentAddress`, is easier to forget about CA data with references. With the ca-derivations work, we hope those become commonplace, and we want to support them right.

2. Have `FullContentAddress` as an alternative to `StorePath` in internal interfaces. The idea is just a small step beyond #3721 and similar PRs, which have lots of `StorePath, std::optional<ContentAddress>` in arguments, and instead do `std::variant<StorePath, FullContentAddress>`.

3. Eventually, do cross-store-dir substitutions of ca-references data, as when the store paths have the same length this should be possible.


src/libstore/local-store.cc
```diff
@@ -1035,7 +1057,11 @@ StorePath LocalStore::addToStoreFromDump(const string & dump, const string & nam
 {
     Hash h = hashString(hashAlgo, dump);
 
-    auto dstPath = makeFixedOutputPath(method, h, name);
+    auto dstPath = makeFixedOutputPath(name, FixedOutputInfo {
+        method,
+        h,
+        {},
```
matthewbauer, line 1063: Can you not have references with addToStoreFromDump?

Ericson2314: @matthewbauer I would say all those new `{}` are things that probably deserve auditing :)

src/libfetchers/git.cc
```diff
@@ -192,7 +192,10 @@ struct GitInputScheme : InputScheme
                 input.attrs.insert_or_assign("revCount", getIntAttr(infoAttrs, "revCount"));
             input.attrs.insert_or_assign("lastModified", getIntAttr(infoAttrs, "lastModified"));
             return {
-                Tree(store->toRealPath(storePath), std::move(storePath)),
+                Tree {
+                    store->toRealPath(store->makeFixedOutputPathFromCA(storePathDesc)),
```
andir, line 196: Could we add the field names to assignments like these? I am always a bit careful when most of the types are essentially strings and could be constructed from strings. You do not want the reordering of some fields in a struct to cause any side-effects on the actual code. Here at least `toRealPath` returns a string which is a valid input for the `StorePathDesc` type afaics.

Ericson2314: I would like to too!

Issue #3750, opened 2020-06-25
Add "nix challenge" command that is equivalent to "guix challenge"


matthewbauer:
A feature added to guix that is missing in Nix is `guix challenge`. It's documented here:

https://guix.gnu.org/manual/en/html_node/Invoking-guix-challenge.html

see also https://news.ycombinator.com/item?id=23543986

PR #3761, opened 2020-06-29
nix/build: add `--rebuild` option

Ma27:
Occasionally, `nix-build --check` is fairly helpful and I'd like to be
able to use this feature for flakes that need to be built with `nix
build` as well.

cc @edolstra

### DIFF
--- a/src/nix/build.cc
+++ b/src/nix/build.cc
@@ -9,6 +9,7 @@ using namespace nix;
 struct CmdBuild : InstallablesCommand, MixDryRun, MixProfile
 {
     Path outLink = "result";
+    BuildMode buildMode = bmNormal;
 
     CmdBuild()
     {
@@ -25,6 +26,12 @@ struct CmdBuild : InstallablesCommand, MixDryRun, MixProfile
             .longName = "no-link",
             .description = "do not create a symlink to the build result",
             .handler = {&outLink, Path("")},
+        });
+
+        addFlag({
+            .longName = "rebuild",
+            .description = "Rebuild an already built package locally and compare the result to the existing store-paths.",
+            .handler = {&buildMode, bmCheck},
         });
     }
 
@@ -53,7 +60,7 @@ struct CmdBuild : InstallablesCommand, MixDryRun, MixProfile
 
     void run(ref<Store> store) override
     {
-        auto buildables = build(store, dryRun ? Realise::Nothing : Realise::Outputs, installables);
+        auto buildables = build(store, dryRun ? Realise::Nothing : Realise::Outputs, installables, buildMode);
 
         if (dryRun) return;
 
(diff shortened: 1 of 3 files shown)


FRidh:
I think `nix build --check` (the new UI) would be confusing, since flakes have tests denoted as `checks`.

bhipple:
We could perhaps adopt the `--challenge` name from `guix`, which does the same thing.

Ma27:
That's a nice suggestion! I just wanted to ask for a better idea since I couldn't come up with a non-verbose alternative to `--check`.

@FRidh @edolstra do you agree with `--challenge`? If yes, I'd change my PR accordingly.

FRidh:
In https://github.com/NixOS/nix/issues/3750 it is suggested to have a separate command for it, `nix challenge`.

Ma27:
I'd be fine with this as well.

Ma27:
I should note though that if we add a `nix challenge`-command we should try to reach feature-parity with `guix challenge` which also supports e.g. comparing build products to the results from binary caches. Also it's directly possible to introspect the differences of store-paths with tools such as diffoscope (https://guix.gnu.org/manual/en/html_node/Invoking-guix-challenge.html).

Thus I also have to revoke my :+1: for `--challenge`, using this name would IMHO create too much confusion for people who also know `guix`.

Implementing such a feature would cost fairly much additional time I cannot invest atm. However I'd like to have this feature for the `nix`-command as well (to test flakes) which is why I'd suggest to either merge this PR as-is or to rename `--check` to something more suitable first (e.g. to `--rebuild`, better suggestions are welcome).

@FRidh @edolstra thoughts?

Ma27:
Rebased onto master and tested again. @edolstra anything missing to get this merged? :)

Ma27:
@edolstra I think that `nix build --check` is the best solution for now (as it's consistent with `nix-build --check`) and as I've pointed out, we'd need to do way more stuff to implement `nix challenge` (I'm happy to do so, but in another PR), so can we merge this now?

edolstra:
@Ma27 Yeah, I think it's better to rename it to `--rebuild`, since "check" is rather ambiguous (check what? re-run the `checkPhase`?). It could even be a top-level command, i.e. `nix rebuild`.

Ma27:
@edolstra I like `--rebuild`, will change it now.

> It could even be a top-level command, i.e. nix rebuild.

I don't think that's a good idea. I'm fairly sure that people will start to confuse this with `nixos-rebuild` :-)

Also, rather than `nix rebuild` we probably want to implement sth. like `nix challenge` (see https://github.com/NixOS/nix/pull/3761#issuecomment-657080090 for that) with a few more features.

domenkozar:
Note that `challenge` is really a bad name, I'm sure most people can't guess what it would do.

`check` is a bit vague, but at least it gives some constraints to what it could do.

Ma27:
@domenkozar well, the `challenge` thing is something to be discussed when we're about to implement sth like #3750. Right now I only want something like `nix-build --check` for the new `nix` command.

Ma27:
@edolstra updated accordingly.

edolstra:
Thanks!

Issue #3696, opened 2020-06-15
nix copy builds drv files


dasJ:
**Describe the bug**

`nix copy` is supposed to copy full closures and not build anything.
However, it does build stuff when copying drv files.

**Steps To Reproduce**

1. Create `broken.nix` with
```nix
derivation {
  name = "test";
  system = "x86_64-linux";
  builder = "this will error";
}
```
2. `nix copy $(nix-instantiate broken.nix) --to ssh://a-string-that-does-not-matter`

**Expected behavior**

The file should just be copied.

**`nix-env --version` output**

nix-env (Nix) 2.4pre20200521_00b562c

**Additional context**

cc @ajs124

dasJ:
Doubt anyone will ever change their decision on this "design choice".

edolstra:
FYI, `nix copy` has a flag `--derivation` to copy the derivation rather than its outputs.

ajs124:
```bash
        drvFlag=""
        if [[ "${storePath}" == *\.drv ]]; then
            drvFlag="--derivation"
        fi
        nix store sign --key-file "${key}" $drvFlag "${storePath}" || continue
        nix copy --to "${url}?compression=none" $drvFlag "${storePath}"
```
This is what's needed because of this design decision

PR #3762, opened 2020-06-29
nix/copy: Add `--build` option

Ma27:
In 9570036146b9bdbd66ce0b9f71479d0f56f3bf35 `nix copy` was changed to
actually build derivations that are supposed to be copied.

This is not desirable though if you want to copy derivations for a
different platform architecture to a remote store.

This patch only builds drvs if `nix copy --build` is specified,
otherwise it falls back to the previous behavior where drvs were only
copied.

cc @edolstra

### DIFF
--- a/src/nix/copy.cc
+++ b/src/nix/copy.cc
@@ -13,7 +13,6 @@ struct CmdCopy : StorePathsCommand
     std::string srcUri, dstUri;
 
     CheckSigsFlag checkSigs = CheckSigs;
-
     SubstituteFlag substitute = NoSubstitute;
 
     CmdCopy()
@@ -46,7 +45,11 @@ struct CmdCopy : StorePathsCommand
             .handler = {&substitute, Substitute},
         });
 
-        realiseMode = Build;
+        addFlag({
+            .longName = "build",
+            .description = "Whether to build derivations",
+            .handler = {&realiseMode, Build},
+        });
     }
 
     std::string description() override


ajs124:
Fixes #3696

Ma27:
@edolstra anything missing to get this merged?

edolstra:
Not sure this works since `RealiseMode` only affects what gets build, not what gets copied. In any case, if you just want to copy the `.drv`, then the new `--derivation` flag can be used for that (3624c042ace05db88794e87ee37f3296bab19bc8).

Ma27:
> Not sure this works since RealiseMode only affects what gets build, not what gets copied

Well, I don't want to get anything built when using `nix copy`. When I do `nix copy /nix/store/foobar.drv --to ssh://remote-host`, I don't expect a build process. Especially since the `.drv` file is for the architecture on `ssh://remote-host` (which is not necessarily the same I have locally).

Ericson2314:
I kinda of wish `/nix/store/foo-bar.drv` always meant just the drv, and one instead used `/nix/store/foo-bar.drv!name` or `/nix/store/foo-bar.drv!*` (for all of them, if we want) to mean the outputs.

I wanted this not with `nix copy`, but `nix why-depends`, when I want to do all the same logic but on the drvs themselves.

Ma27:
While `nix copy --derivation` works as expected, we probably want to implement a patch which automatically sets the proper `OperateOn`-value depending on the input (e.g. `OperateOn::Derivation` for a `.drv`-file by default).

Apart from that this is just a discussion about which default should be preferred and I don't think that we'll reach an actual agreement there. Another idea would be to allow user-defined aliases for certain [commands such like `git`](https://git-scm.com/book/en/v2/Git-Basics-Git-Aliases) does.

Will close this for now, as mentioned, a better solution would be to depend on the `.drv` file (@edolstra what do you think about this?).

PR #3764, opened 2020-06-29
Fix segfault if a file from an invalid searchpath-entry is attempted to get built

Ma27:
When I try to build an expression from a searchpath that happens to be
wrong, `nix-build` segfaults:

```
λ ma27 [~] → nix-build '<foo/bar>' -I foo=$(mktemp -d)
[1]    16252 segmentation fault (core dumped)  nix-build '<foo/bar>' -I foo=$(mktemp -d)
```

This happens if the `file`-Symbol in the `Pos` struct doesn't have a
value (due to the missing searchpath-entry). This causes a segfault
when the `Error`-implementation tries to cast `Pos` to an `ErrPos`.

This was most likely caused by 55eb71714854b262b5e1079ff250a13cc0bbf644,
previously it was explicitly checked if `pos` has any value. This patch
basically switches back to this behavior by setting `nixCode` to
`std::nullopt` if `pos` is incomplete.

Also added a testcase to make sure that this doesn't appear again in the
future.

### DIFF
--- a/src/libexpr/parser.y
+++ b/src/libexpr/parser.y
@@ -688,12 +688,13 @@ Path EvalState::findFile(SearchPath & searchPath, const string & path, const Pos
         Path res = r.second + suffix;
         if (pathExists(res)) return canonPath(res);
     }
+    auto nixCode = pos ? std::optional<NixCode>{NixCode { .errPos = pos }} : std::nullopt;
     throw ThrownError({
         .hint = hintfmt(evalSettings.pureEval
             ? "cannot look up '<%s>' in pure evaluation mode (use '--impure' to override)"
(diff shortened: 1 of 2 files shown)


Ma27:
cc @edolstra @domenkozar @bburdette 
We should also port this patch to `flakes` where I originally discovered this bug.

bburdette:
Good find.  I came across something similar in my show-trace branch, in a different context.  I've got a PR out for the general case of pos->errpos conversion, here:  https://github.com/NixOS/nix/pull/3767

edolstra:
Is this PR still necessary after #3767?

edolstra:
Appears to work now.

Issue #3507, opened 2020-04-16
Static initialization order between evalSettings and getHome2


PkmX:
I ran into another issue while building a statically-linked Nix and this seems to be the case of the infamous static initialization order fiasco.

In my case `evalSettings` is constructed before `getHome2`. `evalSettings` calls `getDefaultNixPath` while initializing `nixPath` member, which in turns calls `getHome()`, which then calls `operator()` on the uninitialized `getHome2` variable. Eventually this throws `std::bad_function_call` since its `init` member isn't initialized. With dynamic linking it is not a problem since `libnixutil.so` will be loaded first and thus its constructors will be run before others.

Adding `__attribute__((init_priority(101)))` to `getHome2` does workaround the issue for now, but seeing #2698 I think we need to consider moving away from using static initialization for side effects.

tomberek:
Closing as stale. Re-open if needed.

PR #3765, opened 2020-06-29
Fix and CI static builds

Ericson2314:
We like static builds because we can use to make demos more self-contained and less spooky on systems without a regular installation of Nix.

There number of things that didn't work out of the box also makes us think CI would be a good idea. Static linking shouldn't be this invasive for most programs, but a few things have accumulated. Hopefully we can eventually fix them more elegantly by using a non-bespoke build system (that e.g. could use pkg-config private deps to figure out the right link order).


mk/programs.mk
```diff
@@ -32,7 +32,7 @@ define build-program
   $$(eval $$(call create-dir, $$(_d)))
 
   $$($(1)_PATH): $$($(1)_OBJS) $$(_libs) | $$(_d)/
-	$$(trace-ld) $(CXX) -o $$@ $$(LDFLAGS) $$(GLOBAL_LDFLAGS) $$($(1)_OBJS) $$($(1)_LDFLAGS) $$(foreach lib, $$($(1)_LIBS), $$($$(lib)_LDFLAGS_USE))
+	$$(trace-ld) $(CXX) -o $$@ $$(LDFLAGS) $$(GLOBAL_LDFLAGS) $$($(1)_OBJS) $$($(1)_LDFLAGS) $$(foreach lib, $$($(1)_LIBS), $$($$(lib)_LDFLAGS_USE)) -lz
```
matthewbauer, line 35: this probably shouldn't be in upstream

matthewbauer: maybe make it conditional on static? by setting GLOBAL_LDFLAGS_AFTER

Ericson2314: Well I think it's harmless in non-static ones? All this blah blah_after stuff is still hacky.

matthewbauer: Yeah i guess. Maybe just needs a big fixme or something. Libz is in everything so maybe no one will mind.

src/libutil/util.cc
```diff
@@ -511,7 +510,8 @@ std::string getUserName()
 }
 
 
-static Lazy<Path> getHome2([]() {
+static Path getHome2()
```
matthewbauer, line 513: This is also not equivalent, but i had no idea how to fix lazy for static musl

Ericson2314:
I guess I'll call it semi-wip, hah.

domenkozar:
Also needs `-A build-static` in `.github/nix/test.yml`

matthewbauer:
https://github.com/NixOS/nix/issues/3507 has a good explanation of why the getHome* changes are needed.

mk/programs.mk
edolstra: Maybe we can put this in `Makefile` or `local.mk` at toplevel (e.g. `GLOBAL_LDFLAGS = -lz`). But it would be better to fix whatever package needs libz implicitly :-)

src/nix/why-depends.cc
```diff
@@ -156,7 +156,7 @@ struct CmdWhyDepends : SourceExprCommand
             auto pathS = store->printStorePath(node.path);
 
             assert(node.dist != inf);
-            logger->stdout("%s%s%s%s" ANSI_NORMAL,
+            logger->stdout_("%s%s%s%s" ANSI_NORMAL,
```
edolstra, line 159: Maybe `cout` instead of `stdout_`? (Like `std::cout`.)

flake.nix
```diff
@@ -427,10 +442,47 @@
       checks = forAllSystems (system: {
         binaryTarball = self.hydraJobs.binaryTarball.${system};
         perlBindings = self.hydraJobs.perlBindings.${system};
+      } // nixpkgs.lib.optionalAttrs (builtins.elem system linuxSystems) {
+        build-static = self.hydraJobs.build-static.${system};
```
edolstra, line 446: Maybe use `buildStatic` since we're using camelCase for these jobs.

mk/programs.mk
Ericson2314: https://github.com/libarchive/libarchive/issues/1446 I finally found the underlying issue! Will make a workaround in `configure.ac`.

Ericson2314:
@edolstra OK think this is finally cleaned up and ready!

Ericson2314:
Ugh, tests failed because evidently the channel doesn't include https://github.com/nixos/nixpkgs/commit/9f5ba92b945d38f624513c5ae81c0181595cb901 yet?

Ericson2314:
Well, and now there are different problems.

edolstra:
BTW I noticed that the `nix` binary has some references to openssl and nlohmann_json:
```
$ nix-store -qR ./result/
/nix/store/4nayz4kkm15s5xxl9q3x4amz63wnffrg-openssl-1.1.1g-x86_64-unknown-linux-musl
/nix/store/akmnk90lr4snlh52i90fk4h9j4k9bhxh-nlohmann_json-3.7.3-x86_64-unknown-linux-musl
/nix/store/nkgg7yaaijh1f06a8zck2diz3014fm13-nix-3.0pre20201017_e6247a5-x86_64-unknown-linux-musl
```
which would be nice to get rid of to get a really minimal closure.

PR #3773, opened 2020-07-01
Add lines-of-code and show-trace features to new error format.

bburdette:
This PR brings in show-trace to the new error format, and adds lines of code too.   

I renamed addPrefix in BaseError to addTrace, since it was always being used for trace anyway.  addTrace takes an optional ErrPos and stores the traces in a list now.

The Pos struct gets a new 'origin' enum, which can be string, stdin, or file.  if Pos.origin is stdin or string, then Pos.file points to the actual code with the error (stored in the symbol table).  If Pos.origin is file, then Pos.file is the *filename* of the code with the error, as before.

At error print time, I get the code either from the symbol table for stdin or string, or by reading in the file.  This is a little redundant in the case of files, since we just read those in for parsing - now we have to read them again for error reporting.  But in the 'happy path' case its a good thing - we're not caching the nix file text for error reporting that may never happen.

Since the nix code is being read from the file or string, I removed the lines of code part from ErrorInfo.  So now ErrorInfo only has errPos, and not nixCode.  The downside here is if we wanted to 'hydrate' errors with the LOC and store them for later reference, that no longer is possible.  But that's only a theoretical use case at this point, which we could restore when needed.  For now it makes things simpler at the call site to have just errPos.

show-trace is currently a Setting<bool> in settings, which is in libstore.  Since ErrorInfo and the loggers are in libutil they don't have access to this flag.  So, I put a showTrace flag into the loggers, and I set that flag in handleExceptions, which is where traces were being printed before.

Some examples of lines-of-code and show-trace:

Simple LOC with highlighted error location:
Selection_046

Hash error followed by trace.  Error is first, with call sites in reverse order, which is different from before.
Selection_047

addErrorContext doesn't include an errPos, just the context message.
Selection_048

'double' trace with some lambdas.  See eval.cc line 1293.
Selection_049


src/libexpr/parser.y
```diff
@@ -569,13 +569,24 @@ formal
...
+    switch (origin) {
+        case foFile: 
+            data.file = data.symbols.create(path);
+            break;
+        case foStdin:
+        case foString:
+            data.file = data.symbols.create(text);
+            break;
+        default:
+            throw Error("invalid FileOrigin in parse");
```
edolstra, line 587: Since this cannot happen it should be an `assert(false)` or `static_assert(false)`.

bburdette: done.

bburdette:
> We could move `show-trace` to libutil. See `ArchiveSettings` in libutil/archive.cc as an example.

Great - I moved show-trace to loggerSettings and ripped out the getter and setter code.  Much cleaner.

PR #3776, opened 2020-07-02
Allow depending on content-addressed derivations

regnat:
Builds on top of #3740, and extends it to allow depending on ca-derivations, with a proper cutoff in case two different derivations in the build chain produce the same output (see [the added test](https://github.com/NixOS/nix/blob/d32357a9c73d4cb6c20ff9f76b4e9304c51a2a81/tests/content-addressed.sh))


Ericson2314:
This does seem like a bunch code doubling down on the "real and fake outputs" approach, which I do worry will be a source of bugs.

Now that derivation outputs are fully parsed, I'd like to exploit that, going from
```c++
struct DerivationOutput
{
    StorePath path;
    std::optional<FixedOutputHash> hash; /* hash used for expected hash computation */
};
```
to
```c++
struct DerivationOutputLegacy {
    StorePath path;
};
struct DerivationOutputCAFixed {
    StorePath path;
    /* hash used for expected hash computation */
    FixedOutputHash hash;
};
struct DerivationOutputCAFloating {
    /* hash used for ca path computation */
    HashType hashType;
};
typedef std::variant<
    DerivationOutputLegacy,
    DerivationOutputCAFixed,
    DerivationOutputCAFloating
> DerivationOutput;

/* accessors to make it easier to handle multiple cases at once. */
std::optional<StorePath> outputPath(DerivationOutput) { ... }
std::optional<HashType> outputHashAlgo(DerivationOutput) { ... }
```

I really do think we can make that change, fix all the type errors "on autopilot', and be quite close to a working version.

Ericson2314:
https://github.com/NixOS/nix/pull/3754/files should also help. Every time in that PR I do something like 
```c++
FixedOutputInfo {
    { .method = ..., .hash = ... },
    {} // empty references
}
```
the `{}` is practically a Mad Lib beginning to be filed in with richer reference data.

src/libstore/rewrite-derivation.cc
```diff
@@ -0,0 +1,100 @@
...
+
+    for (auto & i : drv.outputs) {
+        // XXX: There's certainly a better and less error-prone way
+        // of getting the name than to look it up in the drv environment
+        string name = ParsedDerivation(StorePath::dummy, drv).getStringAttr("name").value_or("");
+        StorePath outPath = store.makeOutputPath(i.first, h, name);
+        auto outputEnvVar = drv.env.find(i.first);
+        if (outputEnvVar != drv.env.end())
+            outputEnvVar->second = store.printStorePath(outPath);
+        debug(format("Rewrote output %1% to %2%")
```
edolstra, line 32: `format` is obsolete, please do `debug("rewrite output '%s' to '%s'", arg1, arg2)`.

src/libstore/rewrite-derivation.cc
```diff
@@ -0,0 +1,100 @@
...
+        };
+    }
+
+    /* Use the masked derivation expression to compute the output path. */
+    Hash h = hashDerivationModulo(store, drv, true);
+
+    for (auto & i : drv.outputs) {
+        // XXX: There's certainly a better and less error-prone way
+        // of getting the name than to look it up in the drv environment
+        string name = ParsedDerivation(StorePath::dummy, drv).getStringAttr("name").value_or("");
```
edolstra, line 27: This can be lifted out of the for loop.

src/libstore/store-api.cc
```diff
@@ -742,6 +743,39 @@ std::string Store::showPaths(const StorePathSet & paths)
     return s;
 }
 
+ValidPathInfo Store::makeContentAddressed(
+        const ValidPathInfo & info,
+        const StringMap & extraRewrites
+    )
+{
+    bool hasSelfReference = info.references.find(info.path) != info.references.end();
```
edolstra, line 751: => `bool hasSelfReference = info.references.count(info.path)`.

src/libstore/store-api.hh
```diff
@@ -673,6 +674,14 @@ public:
     virtual void createUser(const std::string & userName, uid_t userId)
     { }
 
+    /* Create a new path similar to the one referred to by `info`, but that's
+     * content-addressed.
+     *
+     * This doesn't alter the original path
```
edolstra, line 680: This should document whether `info` needs to denote a valid path.

src/libstore/local-fs-store.cc
```diff
@@ -79,8 +79,6 @@ ref<FSAccessor> LocalFSStore::getFSAccessor()
 
 void LocalFSStore::narFromPath(const StorePath & path, Sink & sink)
 {
-    if (!isValidPath(path))
-        throw Error("path '%s' is not valid", printStorePath(path));
```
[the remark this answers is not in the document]
edolstra: If the issue is that build.cc is calling `Store::makeContentAddressed()` on an invalid path, then maybe that function should take a separate argument for the path (so it can use `dumpPath()`).

src/libstore/store-api.hh
```diff
@@ -115,6 +115,7 @@ struct ValidPathInfo
 {
     StorePath path;
     std::optional<StorePath> deriver;
+    std::optional<string> outputname;
```
Ericson2314, line 118: Is the intent that we can have `outputname` whenever we have `deriver`, except for backwards compat with old infos?

src/libstore/build.cc
```diff
@@ -4248,6 +4286,7 @@ void DerivationGoal::done(BuildResult::Status status, std::optional<Error> ex)
     mcRunningBuilds.reset();
 
     if (result.success()) {
+        registerOutputMappings();
```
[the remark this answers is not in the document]
regnat: There might be a better place for this indeed.
I didn't put it in `registerOutputs` because we want it to be called even when no build took place (either because we substituted the output or because we've short-cutted the build after resolving the derivation).

src/libstore/local-store.cc
```diff
@@ -576,6 +576,17 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
     }
 }
 
+void LocalStore::linkDeriverToPath(const StorePath deriver, const string outputName, const StorePath output)
```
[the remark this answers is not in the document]
regnat: Oh forgot these, thanks

src/libstore/rewrite-derivation.hh
```diff
@@ -0,0 +1,23 @@
...
+#include "types.hh"
+#include "path.hh"
+
+namespace nix {
+
+/**
+ * Replace all occurences of a path in `keys(pathRewrites)` in the derivation
+ * by its associated value.
+ */
+void rewriteDerivation(Store & store, Derivation & drv, const StringMap & pathRewrites);
```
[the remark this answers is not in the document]
regnat: I wasn't sure what the best choice was so I left it outside. But sure, I can add it to the class if that seems better to you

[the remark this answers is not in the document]
regnat: Mh actually that might be much better because we can make them methods of `BasicDerivation` that are overriden by `Derivation` and remove a fishy `dynamic_cast` in `build.cc`

src/libstore/store-api.hh
edolstra, line 118: It seems incorrect to have an `outputName` field in `ValidPathInfo`. A CA store path can be produced by differently named output of different derivations (e.g. the "foo" output of derivation A could produce the same contents as the "bar" output of derivation B).

edolstra: BTW it looks like this field is not actually used anywhere.

src/libstore/local-store.cc
```diff
@@ -576,6 +576,17 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
...
+{
+    retrySQLite<void>([&]() {
+        auto state(_state.lock());
+        state->stmtAddDerivationOutput.use()
+            (queryValidPathId(*state, deriver))
+            (outputName)
+            (printStorePath(output))
```
edolstra, line 586: `output` is an "actual" output here, right? In that case, doesn't this cause confusion wrt the meaning of the `DerivationOutputs` table?

src/libstore/rewrite-derivation.hh
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Replace all the input derivations of the derivation by their output path
+ * (as given by `queryDerivationOutputMap`)
+ * **if this one differs from the one written in the derivation**
+ */
+bool resolveDerivation(Store & store, Derivation & drv);
```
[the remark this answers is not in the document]
regnat: Oh my bad I forgot to document it.

It serves to indicate whether the resolving actually modified the derivation or not (to know whether to retry the substitution). But I'm not very happy with it, so I'd be happy if you have any suggestion on improving this

src/libstore/build.cc
```diff
@@ -1349,6 +1356,18 @@ void DerivationGoal::inputsRealised()
 {
     trace("all inputs realised");
 
+    if (useDerivation) {
+        bool isModified = resolveDerivation(worker.store, *(dynamic_cast<Derivation *>(drv.get())));
+        // Define the actual outputs of the build
+        for (auto & i: drv->outputs)
+            actualOutputs.emplace(i.first, i.second.path);
```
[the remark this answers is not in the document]
Ericson2314: I think this is intentional, but this is the sort of foot-gun I want to avoid by making `DerivationOutput` use variants in my comment above.

src/libstore/store-api.hh
regnat: Indeed, I've added it for an experiment and forgot to remove it. This field doesn't really make any sense (even the `deriver` field should probably be deprecated as a store path can now have several derivers)

Ericson2314: Yeah once we upload the trust mapping, we'll probably want to key it by the derivation not the output path, which also helps different signatures sign for different derivers.

PR #3777, opened 2020-07-02
Run the tests in parallel

regnat:
Change the way the testsuite runs to make each test a separate make target, allowing to run them in parallel rather than sequentially as we do now.
This causes `make installcheck -j$(nproc)` to drop from ~4m30 to ~40s on my machine.

To allow that, each test now runs under a different `TEST_ROOT` (which also ensures that they are truly independant), and runs `tests/init.sh` first to setup everything

PR #3778, opened 2020-07-02
Parallel tests fixes

regnat:
Fix the issues introduced by #3777 

- Restore `tests/plugins/libplugintest.*` as a dependency of the testsuite
- Shorten the path to the test root to fix a `socket path is too long` error on the OSX builders

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -8,7 +8,7 @@ define run-install-test
 
   .PHONY: $1.test
   $1.test: $1 $(test-deps)
-	@env TEST_NAME=$1 TESTS_ENVIRONMENT="$(tests-environment)" mk/run_test.sh $1
+	@env TEST_NAME=$(notdir $(basename $1)) TESTS_ENVIRONMENT="$(tests-environment)" mk/run_test.sh $1
 
 endef
 
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -1,10 +1,13 @@
 # Run program $1 as part of ‘make installcheck’.
+
+test-deps =
+
 define run-install-test
 
   installcheck: $1.test
 
   .PHONY: $1.test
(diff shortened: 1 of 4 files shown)


regnat:
(waiting for the CI to confirm the OSX fix as I can't test it locally)

Ericson2314:
Maybe we can use relative paths when we open create socket to get around this?

regnat:
> Maybe we can use relative paths when we create the socket to get around this?

I tried that but it's not allowed

I managed to work around the path length by directly putting the daemon socket under `$TEST_ROOT` in the tests. In theory the error can still happen if `$TMPDIR` is too long, but that gives us roughly as much margin as what we used to have before.

Now I'm stuck again with a non-sensical error on the OSX builder :angry:  (the `nix-shell` test fails on a call to `nix develop`, except that I couldn't find any call to `nix develop` in the whole codebase, so I have no idea where this can come from)

regnat:
> Now I'm stuck again with a non-sensical error on the OSX builder angry

Mh, the same pipeline passed on Tweag's repo (https://github.com/tweag/nix/runs/833213722). I'll try force-pushing to restart the pipeline and see if the failure persists

edolstra:
Looks like there is also a dependency on `tests/config.nix` missing: https://nix-cache.s3.amazonaws.com/log/r2qnl4d2k9ix8i2gbi816x6lszmjgf6z-nix-2.4pre7875_5596f879.drv

regnat:
> Looks like there is also a dependency on tests/config.nix missing

Yup', I've also restored it in this branch

The OSX failure now seems to be a real one introduced by  5596f87 (I've tried rebasing it on top of yesterday's master removing the "parallel tests" MR, and it still fails: https://github.com/tweag/nix/runs/833623967)

PR #3782, opened 2020-07-03
Don't use read-only mode for nix build --dry-run

### LINKED ISSUE #1795: 'nix build --dry-run' fails on new derivations, acts like readOnlyMode
Using nixpkgs master (shouldn't matter which, but):
```console
$ git rev-parse HEAD
6887a0fc9a802331a2d77b499f9692548899c84c
```

Using `nix build --dry-run` on an installable we don't have a `.drv` file for, fails:
```console
$ nix build -f . --dry-run android-studio
don't know how to build these paths (may be caused by read-only store access):
  /nix/store/0fp7f41a6s1zfha98qf6180qghc3858y-android-studio-3.0.1.0.drv!out
```

Which honestly is an error I've only seen when doing unusual things previously.

`nix-build` on the same works as expected:
```console
$ nix-build --dry-run -A android-studio
  /nix/store/0fp7f41a6s1zfha98qf6180qghc3858y-android-studio-3.0.1.0.drv
these paths will be fetched (645.11 MiB download, 1151.45 MiB unpacked):
  /nix/store/5h6zhfg68lzih2r8v4bj3dqi7bkwlfml-android-studio-fhs-env-fhs
  /nix/store/8ai9s4h4x1hwszaf8mp1dl1pbnx1cjxb-android-studio
  /nix/store/955pgx40p0iz4vss7yk0b7ahc7hi1k5n-android-studio-fhs-env-init
  /nix/store/d1q9aypz215rg59bb1fgymjr0qicnymf-android-studio-fhs-env-usr-target
  /nix/store/ihdlxc3gqvj1ak9pjg521yiv0a5niqpp-android-studio-fhs-env-chrootenv-etc
  /nix/store/jdpkx4dwr7sr95dfy93vvcl3fx59sfc4-profile
  /nix/store/lrxd00h7cq8ffqrrgx3c81bfyj2gzwyr-android-studio-fhs-env-usr-multi
  /nix/store/nq7bid5yn9gia84g43aiznrs2z8jxcl6-android-studio-fhs-env
  /nix/store/xwxcr2gjv23dy8svjpv2mrqnfynrdq69-fonts.conf
```

And since it instantiates the `.drv` file, `nix build` works again:
```console
$ nix build -f . --dry-run android-studio
  /nix/store/0fp7f41a6s1zfha98qf6180qghc3858y-android-studio-3.0.1.0.drv
these paths will be fetched (645.11 MiB download, 1151.45 MiB unpacked):
  /nix/store/5h6zhfg68lzih2r8v4bj3dqi7bkwlfml-android-studio-fhs-env-fhs
  /nix/store/8ai9s4h4x1hwszaf8mp1dl1pbnx1cjxb-android-studio
  /nix/store/955pgx40p0iz4vss7yk0b7ahc7hi1k5n-android-studio-fhs-env-init
  /nix/store/d1q9aypz215rg59bb1fgymjr0qicnymf-android-studio-fhs-env-usr-target
  /nix/store/ihdlxc3gqvj1ak9pjg521yiv0a5niqpp-android-studio-fhs-env-chrootenv-etc
  /nix/store/jdpkx4dwr7sr95dfy93vvcl3fx59sfc4-profile
  /nix/store/lrxd00h7cq8ffqrrgx3c81bfyj2gzwyr-android-studio-fhs-env-usr-multi
  /nix/store/nq7bid5yn9gia84g43aiznrs2z8jxcl6-android-studio-fhs-env
  /nix/store/xwxcr2gjv23dy8svjpv2mrqnfynrdq69-fonts.conf
```

This is breaks one of my primary uses for `--dry-build`: using it after modifying expressions to see what the fallout is.  Anyway I don't think it's quite the intention of `--dry-run` anyway.

--------------

FWIW this seems to have been introduced in https://github.com/NixOS/nix/commit/c36467ad2e2e27d04e681144e0e10417c53c10c1#diff-fc4f446fc21b1365be7f6339ceb8efe8L232 although I'm not sure why, haven't looked too closely.

madjar:
In dry run mode, new derivations can't be create, so running the command on anything that has not been evaluated before results in an error message of the form `don't know how to build these paths (may be caused by read-only store access)`.

For comparison, the classical `nix-build --dry-run` doesn't use read-only mode.

Closes #1795

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -716,7 +716,7 @@ Buildables build(ref<Store> store, Realise mode,
         }
     }
 
-    if (mode == Realise::Nothing)
+    if (mode == Realise::Nothing || mode == Realise::Derivation)
         printMissing(store, pathsToBuild, lvlError);
     else if (mode == Realise::Outputs)
         store->buildPaths(pathsToBuild, bMode);
--- a/src/nix/build.cc
+++ b/src/nix/build.cc
@@ -52,7 +52,7 @@ struct CmdBuild : InstallablesCommand, MixDryRun, MixJSON, MixProfile
 
     void run(ref<Store> store) override
     {
-        auto buildables = build(store, dryRun ? Realise::Nothing : Realise::Outputs, installables, buildMode);
+        auto buildables = build(store, dryRun ? Realise::Derivation : Realise::Outputs, installables, buildMode);
 
         if (dryRun) return;
 


dtzWill:
cc 94eb5fad76cd086e3f49329532f81083726f89b3, which does not mention the impact as including this but (untested) looks like maybe it was motivated by this sort of issue? Regardless maybe this can be fixed while visiting that logic, if it hasn't already been addressed? :innocent: 

(if there are known issues in the way, please share as maybe someone can fix them ^_^)

madjar:
94eb5fa doesn't fix the issue, so I rebased on top of it and updated the patch.

I'm guessing this is ready for review.

madjar:
@edolstra Would you be interested in merging that? It would enable a few workflows, including https://github.com/Mic92/nix-build-uncached for flakes.

zimbatm:
Note that the current version works fine if the repo doesn't use IFD.

madjar:
@zimbatm I'm not sure that's true. I have here a repo without IFD, and I still get the issue (because `nix build --dry-run` won't create the `.drv` files).

You can reproduce with `nix build --dry-run nixpkgs#hello`.

madjar:
Hey folks!

I just rebased this PR on top of master.

## Reminder of what this PR does

So, right now, running `nix build --dry-run` on a derivation that is not already in the store will say "don't know how to build these paths (may be caused by read-only store access):", which is not super helpful.

This PR changes the realisation mode for `--dry-run` from `Nothing` to `Derivation`. As a result, the derivation will get instantiated.

I believe it's good to change the behaviour of `--dry-run` instead of introducing a new flag, because that is the behaviour of the classical `nix-build` command, and because I'm not sure that a command that only returns a usable result if some other command has been run before is a use case.

regnat:
I think you can remove the `Realise::Nothing` enum member altogether as it's never used anymore (It was used in only two places, #4541 removed one and you remove the other)

anka-213:
Could we also introduce a new flag that never builds anything for those who want to preserve the current behaviour (if anyone does)?
Or even better, one that says what `--dry-run` would build, but that is definitely out of scope for this pull request.

Ma27:
In case anybody needs a version compatible with the latest master: https://gist.github.com/Ma27/4337efe6be1e3858b0209ad1c9cbdcfe

(cc @edolstra would be awesome if this could be merged, though)

edolstra:
@madjar @Ma27 Thanks, applied!

PR #3783, opened 2020-07-05
address failing addTrace test

bburdette:
this PR is just so I can prompt macos rebuilds and try to fix my failing addTrace.showTracesWithShowTrace test!

### DIFF
--- a/src/libutil/error.cc
+++ b/src/libutil/error.cc
@@ -110,7 +110,6 @@ std::optional<LinesOfCode> getCodeLines(const ErrPos &errPos)
         int pl = errPos.line - 1;
 
         LinesOfCode loc;
-
         do
         {
             std::getline(iss, line);
--- a/src/libutil/error.cc
+++ b/src/libutil/error.cc
@@ -110,6 +110,7 @@ std::optional<LinesOfCode> getCodeLines(const ErrPos &errPos)
         int pl = errPos.line - 1;
 
         LinesOfCode loc;
+
         do
         {
             std::getline(iss, line);
--- a/src/libutil/error.cc
+++ b/src/libutil/error.cc
@@ -354,26 +354,21 @@ std::ostream& showErrorInfo(std::ostream &out, const ErrorInfo &einfo, bool show
 
         for (auto iter = einfo.traces.rbegin(); iter != einfo.traces.rend(); ++iter)
         {
-            try {
+            out << std::endl << prefix;
+            out << ANSI_BLUE << "trace: " << ANSI_NORMAL << iter->hint.str();
+
+            if (iter->pos.has_value() && (*iter->pos)) {
+                auto pos = iter->pos.value();
                 out << std::endl << prefix;
-                out << ANSI_BLUE << "trace: " << ANSI_NORMAL << iter->hint.str();
+                printAtPos(prefix, pos, out);
 
-                nl = true;
-                if (*iter->pos) {
-                    auto pos = iter->pos.value();
+                auto loc = getCodeLines(pos);
+                if (loc.has_value())
+                {
                     out << std::endl << prefix;
-
-                    printAtPos(prefix, pos, out);
-                    auto loc = getCodeLines(pos);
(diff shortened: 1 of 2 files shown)


Ericson2314:
I tried to fix this, but only got so far as reproducing issues on linux and making some cleanups:
```diff
diff --git a/src/libutil/error.cc b/src/libutil/error.cc
index a4ee7afc2..89256b667 100644
--- a/src/libutil/error.cc
+++ b/src/libutil/error.cc
@@ -353,15 +353,16 @@ std::ostream& showErrorInfo(std::ostream &out, const ErrorInfo &einfo, bool show

         out << ANSI_BLUE << std::string(lw, '-') << tracetitle << std::string(rw, '-') << ANSI_NORMAL;

-        for (auto iter = einfo.traces.rbegin(); iter != einfo.traces.rend(); ++iter)
+        for (auto trace : einfo.traces)
         {
-            try {
-                out << std::endl << prefix;
-                out << ANSI_BLUE << "trace: " << ANSI_NORMAL << iter->hint.str();
+            out << std::endl << prefix;
+            out << ANSI_BLUE << "trace: " << ANSI_NORMAL << trace.hint.str();
+
+            nl = true;

-                nl = true;
-                if (*iter->pos) {
-                    auto pos = iter->pos.value();
+            if (trace.pos) {
+                auto pos = *trace.pos;
+                if (pos) {
                     out << std::endl << prefix;

                     printAtPos(prefix, pos, out);
@@ -373,8 +374,8 @@ std::ostream& showErrorInfo(std::ostream &out, const ErrorInfo &einfo, bool show
                         out << std::endl << prefix;
                     }
                 }
-            } catch(const std::bad_optional_access& e) {
-                out << iter->hint.str() << std::endl;
+            } else {
+                out << trace.hint.str() << std::endl;
             }
         }
     }
```

I couldn't figure out how to change the code to make the test pass, and frankly I am confused by a few things:

 - all the `out << std::endl` i.e. putting the newline at the *beginning* and not the end of the `<<` chain. I usually think of a text file as a list of lines, and each line ends with a `\n`. So it makes the most sense to me if each `<<` chain is either pushing part of a line (no `std::endl`), or multiple lines (multiple `std:endl`, and ending with `std:endl`). So your style reads to me as pushing a blank line (or is it ending the previous line? and then part of the next line (or is it the entirety of the next line except for the `\n`) so I am confused.

 - If `bad_optional_access` happens the hint will be printed twice. That seems wrong.

edolstra:
I'm not seeing macOS issues, but I do see a failure on i686-linux: https://hydra.nixos.org/build/123569347

Ericson2314:
@edolstra That is the same error. I think it's somewhat non-deterministic but architecture agnostic.

bburdette:
>     * all the `out << std::endl` i.e. putting the newline at the _beginning_ and not the end of the `<<` chain. I usually think of a text file as a list of lines, and each line ends with a `\n`. So it makes the most sense to me if each `<<` chain is either pushing part of a line (no `std::endl`), or multiple lines (multiple `std:endl`, and ending with `std:endl`). So your style reads to me as pushing a blank line (or is it ending the previous line? and then part of the next line (or is it the entirety of the next line except for the `\n`) so I am confused.

Re the newline at the beginning, initially the code was written with newlines at the end, but the messages ultimately get printed with the assumption that they won't end with a newline, like other messages in the code base.  My initial try ended up with extra blank lines as a result.  The choice was either make this code without a terminating newline, or change everything else to have them.

>     * If `bad_optional_access` happens the hint will be printed twice. That seems wrong.

You're right, that's an oversight on my part.  I removed extra print and got rid of the exception based approach for good measure.

bburdette:
Well its 2 for 2.  Maybe this will fix it?  I don't know how to know for sure.

Ericson2314:
@bburdette Ah OK, that makes sense. Perhaps we can fix it be separating the thrown vs final rendered format strings a bit (something I was thinking of doing anyways) but that is definitely a good bit extra work.

PR #3788, opened 2020-07-07
Add a script to install nix on non-systemd systems.

crinklywrappr:
I cleaned up and moved my changes from [this post](https://discourse.nixos.org/t/install-nix-daemon-on-non-systemd-init/7911/3?u=daf) into the repo.  The new install script has very few changes from the original `install-multi-user.sh`.

I didn't test to make sure this ran, and I'm not sure if additional changes are required, eg changes to a makefile to get the script to the right location in the tarball.  However, this is almost a direct copy of what I ran on Artix.

I will keep up with the pull request and respond to CRs, and anything else required to get this merged.


edolstra:
We really shouldn't have another install script that is nearly identical to `scripts/install-multi-user.sh`. That's a maintainability mess. Isn't it possible to add support for non-systemd systems to `scripts/install-multi-user.sh`?

crinklywrappr:
Sure. I'll change it.

crinklywrappr:
Someone may need to enlighten me as to what went wrong here.  It appears the installation worked, but one of the tests broke.

edolstra:
Hm, I don't think that failure has anything to do with your changes...

virusdave:
Wow, i *JUST* started working on something similar today.  Thanks for doing the work for me!  :)

edolstra:
Thanks, merged!

PR #3795, opened 2020-07-08
Only store hash in DerivationOutput for fixed output derivations

matthewbauer:
We just need the hash here because we can calculate the store path from its fixed output hash (with some extra parameters that is). This leads to less data duplication.

To calculate the path, we need a name, a storeDir, a hash, and a hash method. We get the storeDir from the store and the drvName from the BasicDerivation. Previously, BasicDerivation didn't have a name, so had to add this in as well.


Ericson2314:
@regnat This is the preparatory step for https://github.com/NixOS/nix/pull/3776#issuecomment-653262754 . I thought this preparatory change would just be "getting the churn out of the way", switching from the `path` field to accessor but doing nothing further. Being able to deduplicate `BasicDerivation` by calculating the store path was just icing on the cake!

src/libstore/derivations.hh
```diff
@@ -13,10 +13,20 @@ namespace nix {
...
 /* Abstract syntax of derivations. */
 
-struct DerivationOutput
+struct DerivationOutputIntensional
 {
     StorePath path;
-    std::optional<FixedOutputHash> hash; /* hash used for expected hash computation */
+};
+
+struct DerivationOutputFixed
```
Ericson2314, line 21: Oh, it just occurred to me I guess we could get rid of this and just use `FixedOutputHash` directly.

src/libstore/derivations.hh
```diff
@@ -13,10 +13,20 @@ namespace nix {
 
 /* Abstract syntax of derivations. */
 
-struct DerivationOutput
+struct DerivationOutputIntensional
```
Ericson2314, line 16: Should be `DerivationOutputExtensional` (though perhaps we might want to drop the "extensional" vs "intensional" jargon)

matthewbauer: oh right

matthewbauer: do we have a better name than intensional / extensional?

src/libstore/derivations.hh
matthewbauer: yeah, but it would be less type safe right? we could do the same with DerivationOutputExtensional and StorePath

src/libstore/derivations.cc
```diff
@@ -129,9 +129,11 @@ static DerivationOutput parseDerivationOutput(const Store & store, istringstream
 }
 
 
-static Derivation parseDerivation(const Store & store, const string & s)
+static Derivation parseDerivation(const Store & store, const string & s, string name)
```
regnat, line 152: Shouldn't we rather read the name from `drv.env["name"]` to make them more self-contained ? Atm it's technically possible to realize a derivation that doesn't have a `name` field (or has a `name` field that's different from the name of the store path), but there's no way to build such a derivation (even manually because it won't be possible to import it into the store)

src/libstore/derivations.hh
Ericson2314: I wouldn't say "less type safe", just a matter of preference. I don't have a preference, just pointing out the possibility for others that might.

src/libstore/derivations.hh
regnat: I tend to use content-addressed/input-addressed. That's more verbose but doesn't require 10s of brain power to remember which one is which

src/libstore/derivations.cc
Ericson2314: @regnat I would consider magic things in env a libexpr convention rather than it's nice for libstore to not be aware of.

Matt and I talked about getting the name from the derivation output paths during parsing, which would use libstore-aware conventions. We didn't do that in this PR cause we could get away with it, but if we need to get rid of this new argument we can do that.

regnat: > I would consider magic things in env a libexpr convention rather than it's nice for libstore to not be aware of.

I agree in principle, but given that the drv format is not extensible that's the best way we have to pass meta information about a derivation. And there's quite a bit of precedent about reading information from the `env` field (`preferLocalBuild`, `__json`, `requiredSystemFeatures`…), so it seems legit to also do that for the name

src/libstore/derivations.hh
Ericson2314: "input addressed" is great!

edolstra:
Confused by the title - what is `OutputDerivationOutput`?

Ericson2314:
Title is now fixed. The thing we were talking about is `DerivationOutputFixed`, but that is also a newly created type, so better to not refer to it in the title at all.

edolstra:
Could you look at the merge conficts? Otherwise LGTM.

One possible improvement: it's a bit ugly to have to pass `drvName` to `DerivationOutput::path()`, since you could pass the wrong `drvName`. Since a derivation knows its own name, maybe we could have a method `Derivation::path(store, outputName)` that avoids this.

Ericson2314:
> Could you look at the merge conficts? Otherwise LGTM.

Done

> One possible improvement: it's a bit ugly to have to pass drvName to DerivationOutput::path(), since you could pass the wrong drvName. Since a derivation knows its own name, maybe we could have a method Derivation::path(store, outputName) that avoids this.

Yes that we bugging me too. Happy to do that, would you mind if it was a separate PR? If I do it in this I will just get a bunch of conflicts for #3830, so I rather do it on top of #3830.

edolstra:
@Ericson2314 Sure!

Issue #3751, opened 2020-06-25
Add "nix processes" command that is equivalent to "guix processes"


matthewbauer:
Guix has a feature that is missing in Nix:

https://guix.gnu.org/manual/en/html_node/Invoking-guix-processes.html

`guix processes` is used to list derivations that are running by the daemon. It's useful to system administrators who want to know what is using up so much of their resources.

Kloenk:
There is as an adhoc solution [nix-top](https://github.com/samueldr/nix-top). But yeah, a real nix subcommand could be usefully (or at least protocol definition to query this info from the daemon)

zimbatm:
Related to that, it would be great if nix would report back *what* is blocking a build when it is locked or waiting on a runner to become available.

Kloenk:
One simple idea which would not be too complicated in implementing is writing the current build step into `/nix/var/nix/userpool/${build-user}`.

Is that a good idea?

Atemu:
I like the idea of it being discoverable via files. Though perhaps they should live in `/run` instead, this kind of information doesn't need to be held permanently.

Kloenk:
> I like the idea of it being discoverable via files. Though perhaps they should live in `/run` instead, this kind of information doesn't need to be held permanently.

Of course. My idea behind that was that it would make it easier to see if the look on the user is already held without calling flock. As the locks live under /nix/var/nix/userpool

matthewbauer:
I made a quick proof of concept here:

https://github.com/NixOS/nix/pull/3800

I think handling build steps would be useful too. Right now you can check /proc for child processes, but it only works on systems with a procfs.

Ericson2314:
I assume we limit to trusted users?

matthewbauer:
userpool is only readable by root (or whoever owns NIX_STORE_DIR) so any new information we expose would have that restriction. I don’t think it’s worth letting trusted-users see it. I don’t think there’s anything stopping them from doing `pstree $(pgrep nix-daemon)` though.

Atemu:
Still important to me.

Kloenk:
Still important to me

Atemu:
Still important to me.

PR #3800, opened 2020-07-11
Add nix processes command

matthewbauer:
adds a processes command for #3751

some thing still missing:

- mostly relies on procfs for cmdline info
- no phase reporting

output looks like:

```
Build User: nixbld1
Build Process: nix-daemon 3535     
Child Process: /nix/store/b3zsk4ihlpiimv3vff86bb5bxghgdzb9-gcc-9.2.0/libexec/gcc/x86_64-unknown-linux-gnu/9.2.0/cc1 -quiet -D _FORTIFY_SOURCE=2 -idirafter /nix/store/fwpn2f7a4iqszyydw7ag61zlnp6xk5d3-glibc-2.30-dev/include -idirafter /nix/store/b3zsk4ihlpiimv3vff86bb5bxghgdzb9-gcc-9.2.0/lib/gcc/x86_64-unknown-linux-gnu/9.2.0/include-fixed conftest.c -quiet -dumpbase conftest.c -mtune=generic -march=x86-64 -auxbase conftest -g -O2 -O2 -Wformat=1 -Wformat-security -Werror=format-security -fstack-protector-strong -fno-strict-overflow -fPIC --param ssp-buffer-size=4 -o /build/cceFfp2A.s 
File Lock: /nix/store/l285fr6qxcf0v0iaphs38lcghv206zar-hello-2.10.lock
```

### DIFF
--- a/src/nix/processes.cc
+++ b/src/nix/processes.cc
@@ -2,6 +2,7 @@
 #include "store-api.hh"
 #include "pathlocks.hh"
 
+#include <fstream>
 #include <fcntl.h>
 
 using namespace nix;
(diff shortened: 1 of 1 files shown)


matthewbauer:
Ok - this actually doesn't quite work. The userpool locks are always owned by the same process in the nix-daemon. We'll need to filter by uid to determine what the actual pid is.

matthewbauer:
Updated to show the correct pid based on uid of the owning process.

edolstra:
Not sure this is the best approach. 1) It's Linux specific, so at least there should be a big `#if __linux__` around it.  2) It only works when using build users. 3) It make the Nix CLI depend on the internals of `LocalStore`. For instance, it will break once #3600 gets merged. 4) The name `nix processes` is a bit vague. (What processes?) Maybe something like `nix debug show-daemon-status` is better.

Atemu:
Status makes me think more of "how's the daemon doing, is it still up?" (though that also sounds useful), how about `nix daemon monitor`?

You'd use the command to monitor what the daemon is up to atm afterall.

matthewbauer:
> Not sure this is the best approach. 1) It's Linux specific, so at least there should be a big `#if __linux__` around it.

That sounds alright. Some non-Linux systems have procfs, but I have no idea whether it looks the same.

> 3) It make the Nix CLI depend on the internals of `LocalStore`. For instance, it will break once #3600 gets merged.

Right - this is just using the locks in the filesystem. These don't end up being as clear as I originally hoped.

---

What approach would be better? We could keep track of it in nix-daemon - which right now doesn't really know about what builds are happening.
What

Atemu:
Still important to me.

PR #3801, opened 2020-07-11
Constant space `addToStoreFromDump` and deduplicate code

Ericson2314:
`addToStoreFromDump` now takes a stream just like `addToStore` from path, and their `LocalStore` implementations are deduplicated.

I suggest reviewing commit by commit.

(There is a `FIXME` about removing `addToStoreFromDump` altogether, but we can't do that just yet. I don't believe this PR will make doing so harder.)


Ericson2314:
CC @matthewbauer

edolstra:
What's the purpose of `demux` and the `wanted` argument?

Ericson2314:
Demux was the way I found to share the code. I would have rather not made `addToStoreCommon` and just make `addToStore` use `addToStoreFromDump`, but I couldn't figure out how to do that except by using two `sinkToSource` calls and coroutines.

wanted is so for `addToStoreFromDump` in the `sinkToSource` call I know how data is being requested from the source. This I can take advantage because the whole `addToStoreFromDump` is "source to sink to source".

Ericson2314:
@edolstra In short, if the `LambdaSink` in local store can become a `LambdaSource`, then this callback hell can go away.

Ericson2314:
@edolstra Kept on pushing on it, and `wanted` and `demux` are both gone! The trick was making `ChainSource`, after which I could greatly simplify many things.

src/libstore/local-store.cc
```diff
@@ -1033,138 +1033,76 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
...
+    bool inMemory = false;
+
+    std::string dump;
+
+    /* Fill out buffer, and decide whether we are working strictly in
+       memory based on whether we break out because the buffer is full
+       or the original source is empty */
+    while (dump.size() < settings.narBufferSize) {
+        auto oldSize = dump.size();
+        constexpr size_t chunkSize = 1024;
```
edolstra, line 1072:
```suggestion
        constexpr size_t chunkSize = 65536;
```

edolstra:
There's a merge conflict in daemon.cc, otherwise LGTM.

edolstra:
Hm, there's a failure in tests/add.sh.

Ericson2314:
https://github.com/NixOS/nix/runs/890820640 oh great, it's master.

edolstra:
Looks like ac2fc7ba1fe6b64ec535e4ce63d13fcadf7fdba7 caused it.

src/libstore/local-store.cc
meditans: Thanks, this is now corrected in https://github.com/NixOS/nix/pull/3844

PR #3807, opened 2020-07-13
Change types to prepare the way for CA derivations --- contains #3795

Ericson2314:
This is a small changes on top of #3795. The idea is that PR does some other fixes and "boring things' such that the difference between it and this should be very readable.

--------------------

Main commit:

We've added the variant to `DerivationOutput` to support them, but made
`DerivationOutput::path` partial to avoid actually implementing them.

With this chage, we can all collaborate on "just" removing
`DerivationOutput::path` calls to implement CA derivations.


Ericson2314:
CC @regnat  --- Sorry I forgot to ping!

src/libstore/derivations.hh
```diff
@@ -13,10 +13,40 @@ namespace nix {
 
 /* Abstract syntax of derivations. */
 
-struct DerivationOutput
+struct DerivationOutputInputAddressed
 {
+    /* Will need to become `std::optional<StorePath>` once input-addressed
+       derivations are allowed to depend on cont-addressed derivations */
```
Ericson2314, line 19: I figure this is fine for now.

Ericson2314:
Conflicts fixed.

Issue #2431, opened 2018-09-18
fetchGit fails with a not very helpful error message when fetching a revision not in the remote's HEAD


sgraf812:
Rather than trying to hijack #1923 any further, I instead open a new issue here.

I was trying to use the [new (as of Nix 2) way of pinning nixpkgs](https://nixos.wiki/wiki/FAQ/Pinning_Nixpkgs), but it doesn't work for the at the time of this writing most recent `nixos-18.03` commit 01f5e794913a18494642b5f237bd76c054339d61:

```
$ cat << EOF | nix repl
builtins.fetchGit {
  url = https://github.com/nixos/nixpkgs-channels;
  rev = "01f5e794913a18494642b5f237bd76c054339d61";
}
EOF
Welcome to Nix version 2.1.1. Type :? for help.

fatal: not a tree object
error: program 'git' failed with exit code 128

error: Unexpected linenoise keytype: 0
```

The problem here is mostly the `fatal: not a tree object` error. I suspect this is due to `fetchGit` cloning with `--depth 1` or taking some other shortcut, which only considers `master` (in this case `nixos-unstable`) commits. 01f5e794913a18494642b5f237bd76c054339d61 isn't an ancestor of master, so it will not be fetched, hence the error.

Or maybe it's some other reason altogether.

~~Anyway, my current workaround is to use `fetchTarball` instead.~~ Or even by specifying `ref = "nixos-18.03";` as in #1923. I think this would be worth adding to the manuel entry.

grahamc:
This is a problem with Git in general with or without `--depth 1` depending upon the circumstances of the repository and the remote. See https://nixos.org/nix/manual/#idm140737317570688

sgraf812:
Right, I was a bit thrown off by the error message and thought that the cached git repo was corrupted, so I tried to delete it twice, to no avail. A hint in the error message of `fetchGit` would be great.

sgraf812:
Or a warning saying 'you have provided no `ref` field, making a shallow clone of the default branch'.

roberth:
Users run into this problem. [Example.](https://stackoverflow.com/questions/52421987/how-to-find-nixpkgs-revision-where-macos-darwin-binary-cache-is-built/52426587#52426587)

Why not fetch the whole thing by default? Fetching nixpkgs was super slow on my machine despite this optimization, so my guess is that it's even better to just fetch the whole repo as a default, or otherwise as a fallback.

[...]

PR #3814, opened 2020-07-14
Add `allRefs = true` argument to `fetchGit` to explicitly perform a full checkout

Ma27:
Sometimes it's needed to fetch the revision of a git repository without knowing the specific `ref`. In those cases it's needed to perform a full checkout which is now possible by declaring `builtins.fetchGit { allRefs = true;  /* url, rev, etc */ }`.

Refs #2431 #2409 

For further details, please read the descriptions of the two commits.

cc @edolstra

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -324,6 +324,11 @@ static RegisterPrimOp primop_fetchGit({
           A Boolean parameter that specifies whether submodules should be
           checked out. Defaults to `false`.
 
+        - allRefs  
+          Whether to fetch all refs of the repository. With this argument being
+          true, it's possible to load a `rev` from *any* `ref` (by default only
+          `rev`s from the specified `ref` are supported).
+
       Here are some examples of how to use `fetchGit`.
 
         - To fetch a private repository over SSH:
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -392,6 +392,28 @@ struct GitInputScheme : InputScheme
         AutoDelete delTmpDir(tmpDir, true);
         PathFilter filter = defaultPathFilter;
 
+        RunOptions checkCommitOpts(
+            "git",
+            { "-C", repoDir, "cat-file", "commit", input.getRev()->gitRev() }
+        );
+        checkCommitOpts.searchPath = true;
+        checkCommitOpts.mergeStderrToStdout = true;
+
+        auto result = runProgram(checkCommitOpts);
+        if (WEXITSTATUS(result.first) == 128
+            && result.second.find("bad file") != std::string::npos
+        ) {
+            throw Error(
+                "Cannot find Git revision '%s' in ref '%s' of repository '%s'! "
+                    "Please make sure that the " ANSI_BOLD "rev" ANSI_NORMAL " exists on the "
+                    ANSI_BOLD "ref" ANSI_NORMAL " you've specified or add " ANSI_BOLD
+                    "allRefs = true;" ANSI_NORMAL " to " ANSI_BOLD "fetchGit" ANSI_NORMAL ".",
+                input.getRev()->gitRev(),
+                *input.getRef(),
+                actualUrl
+            );
+        }
+
         if (submodules) {
             Path tmpGitDir = createTempDir();
             AutoDelete delTmpGitDir(tmpGitDir, true);
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -59,12 +59,13 @@ struct GitInputScheme : InputScheme
         if (maybeGetStrAttr(attrs, "type") != "git") return {};
 
         for (auto & [name, value] : attrs)
-            if (name != "type" && name != "url" && name != "ref" && name != "rev" && name != "shallow" && name != "submodules" && name != "lastModified" && name != "revCount" && name != "narHash")
+            if (name != "type" && name != "url" && name != "ref" && name != "rev" && name != "shallow" && name != "submodules" && name != "lastModified" && name != "revCount" && name != "narHash" && name != "allRefs")
                 throw Error("unsupported Git input attribute '%s'", name);
 
         parseURL(getStrAttr(attrs, "url"));
         maybeGetBoolAttr(attrs, "shallow");
         maybeGetBoolAttr(attrs, "submodules");
+        maybeGetBoolAttr(attrs, "allRefs");
 
         if (auto ref = maybeGetStrAttr(attrs, "ref")) {
             if (std::regex_search(*ref, badGitRefRegex))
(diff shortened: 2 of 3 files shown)


bjornfor:
Git clone fetches all branches by default, that's why I think fetchGit should do the same, by simply omitting the ref argument.

Ma27:
Well, we're not cloning a git repo though, only the specified ref (or `master` as fallback) is fetched using `git fetch`. That's the entire problem behind this :)

Ma27:
Also, I think it's fine to fetch master only by default: unless you need it, you shouldn't have to fetch a full repo.

src/libfetchers/git.cc
```diff
@@ -322,6 +330,22 @@ struct GitInput : Input
...
+            "git",
+            { "-C", repoDir, "cat-file", "commit", input->rev->gitRev() }
+        );
+        checkCommitOpts.searchPath = true;
+        checkCommitOpts.mergeStderrToStdout = true;
+
+        auto result = runProgram(checkCommitOpts);
+        if (WEXITSTATUS(result.first) == 128 && result.second.find("bad file") != std::string::npos)
+            throw Error(
+                "Cannot find git-revision in '%s' of repo '%s' on ref '%s'!",
```
domenkozar, line 343: This should recommend what to do instead.

bjornfor:
> Well, we're not cloning a git repo though, only the specified ref (or master as fallback) is fetched using git fetch. That's the entire problem behind this :)

I know, but instead of falling back to master we _could_ fall back to "all". IMHO that's more natural, because `git clone URL` can be translated more directly to`fetchGit { URL }` (less new things to learn).

Ma27:
@bjornfor I agree with @edolstra's reasoning in https://github.com/NixOS/nix/issues/2431#issuecomment-423340947. This should just be an escape-hatch if you really need it.

src/libfetchers/git.cc
Ma27: Fixed. It seems as the `fmt` function doesn't do the highlighting of parameters, so I re-added this manually.

Ma27:
Rebased onto latest master.

Ma27:
@edolstra would you be fine with the overall approach taken here to work around the usability-issues of fetchGit?

Ma27:
Rebased onto master now.

Ma27:
cc @edolstra would you mind taking a look at this? :)

Ma27:
@edolstra @domenkozar rebased onto latest master. Would you mind taking a look? :)

Ma27:
@edolstra is there anything missing in this PR? :)

src/libfetchers/git.cc
```diff
@@ -391,6 +400,30 @@ struct GitInputScheme : InputScheme
...
+        auto result = runProgram(checkCommitOpts);
+        if (WEXITSTATUS(result.first) == 128
+            && result.second.find("bad file") != std::string::npos
+        ) {
+            std::string msg = fmt("Cannot find git-revision in '" ANSI_YELLOW "%s" ANSI_NORMAL
+                    "' of repo '" ANSI_YELLOW "%s" ANSI_NORMAL "' on ref '" ANSI_YELLOW
+                    "%s" ANSI_NORMAL "!\n\n"
+                    "Please make sure that the " ANSI_BOLD "rev" ANSI_NORMAL " exists on the "
+                    ANSI_BOLD "ref" ANSI_NORMAL " you've specified or add " ANSI_BOLD
+                    "allRefs = true;" ANSI_NORMAL " to " ANSI_BOLD "fetchGit" ANSI_NORMAL ".",
```
edolstra, line 419: This message seems wrong: `Cannot find git-revision in '<rev>' ...` should be `Cannot find Git revision '<rev>' in ref '<ref>' of repository '<repo>'`.

BTW the ANSI_YELLOW stuff should be unnecessary, it's added automatically if you do `throw Error("... %s ...", bla)`.

edolstra:
For backwards compatibility, instead of adding an `allRefs` argument, we could also bring back `ref = "*"`.

BTW, when using this, what `ref` gets stored in lock files?

src/libfetchers/git.cc
Ma27: Fixed, thanks!

Ma27:
> For backwards compatibility, instead of adding an allRefs argument, we could also bring back ref = "*".

Unfortunately I don't remember the exact details why I decided to create a new attribute, but I *think* that this was to avoid doing even more magic with `ref` (we already check if it starts with `refs/` etc.).

However I just realized that we'd have to use a string anyways to get this working as flake input (unless we alter the logic in `libexpr`, see below in this comment). So @edolstra it's up to you what you'd prefer, I guess I'd be fine with both variants.

> BTW, when using this, what ref gets stored in lock files?

None as I just realized that using this as flake-input doesn't even evaluate (only tested with `fetchGit` here, sorry!).

The reason is that flake-attrs can't seem to be a `boolean`: https://github.com/NixOS/nix/blob/5999978a053b3ec16f448c40b54a1a62ceb82c90/src/libexpr/flake/flake.cc#L118-L122

Ma27:
ping @edolstra how shall we proceed here? :)

Ma27:
ping @edolstra

edolstra:
> The reason is that flake-attrs can't seem to be a boolean: 

We can just handle booleans there. The only reason they're not implemented is because there was no need at the time.

Ma27:
...which means that the current approach is fine?

Ma27:
cc @edolstra anything else tbd here now? :)

Ma27:
@edolstra is the current change fine now?

Ma27:
cc @edolstra anything tbd here? :)

Ma27:
cc @edolstra can we merge this?

src/libexpr/primops/fetchTree.cc
```diff
@@ -323,6 +323,11 @@ static RegisterPrimOp primop_fetchGit({
           A Boolean parameter that specifies whether submodules should be
           checked out. Defaults to `false`.
 
+        - allRefs  
+          Whether to fetch all refs of the repository. With this argument being
+          true, it's possible to load a `rev` from *any* `ref` (by default only
+          `rev`s from `ref = "refs/heads/master"` are supported).
```
edolstra, line 329: Shouldn't this read "only `rev`s from `ref` are supported"?

src/libfetchers/git.cc
```diff
@@ -392,6 +401,28 @@ struct GitInputScheme : InputScheme
...
+        RunOptions checkCommitOpts(
+            "git",
+            { "-C", repoDir, "cat-file", "commit", input.getRev()->gitRev() }
+        );
+        checkCommitOpts.searchPath = true;
+        checkCommitOpts.mergeStderrToStdout = true;
+
+        auto result = runProgram(checkCommitOpts);
+        if (WEXITSTATUS(result.first) == 128
+            && result.second.find("bad file") != std::string::npos
```
edolstra, line 413: Checking for strings like "bad file" in stderr seems pretty fragile... Do we actually need this test? Presumably a wrong `rev` will fail in the subsequent call to `git archive`.

Ma27: I'm afraid no: if `git cat-file` errors, the exit code is always 128, the reason for the error is on stderr.

Ma27:
@edolstra rebased, fixed the docs and also added a test :)

edolstra:
Thanks, merged!

AleXoundOS:
> I know, but instead of falling back to master we _could_ fall back to "all".

Originally I thought so too. But in practice this could leave missing commits in master branch unnoticed, until a feature branch gets deleted on remote.

PR #3816, opened 2020-07-14
Fix localhost::addToStore(...Path...)

Ericson2314:
We were calculating the nar hash wrong when the file ingestion method
was flat. I don't think there's anything we can do in that case but dump
the file again, so that's what I do.

As an optimization, we again could reuse the original dump for just the
recursive and non-sha256 case, but I rather do that after this fix, and
after my other PRs which deduplicate this code.


edolstra:
Thanks. To keep things simple I've just reverted a2c27022e9afc394e08d34d349587c8903fc1a97 since we don't actually use the non-recursive case anymore.

PR #3822, opened 2020-07-16
Optimize `addToStoreSlow` and remove `TeeParseSink`

Ericson2314:
`TeeParseSink` already didn't have any methods. I think it was mainly a vestige left when 1cf480110879ffc8aee94b4b75999da405b71d7c did not revert the entire original commit. It's easy to rewrite with `TeeSource`, so that's what I did.

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -234,7 +234,7 @@ ValidPathInfo Store::addToStoreSlow(std::string_view name, const Path & srcPath,
 
     TeeSink sinkIfNar { narHashSink, caHashSink };
 
-    /* We use the tee sink if we need to hash he nar twice */
+    /* We use the tee sink if we need to hash the nar twice */
     auto & sink = method == FileIngestionMethod::Recursive && hashAlgo != htSHA256
         ? static_cast<Sink &>(sinkIfNar)
         : narHashSink;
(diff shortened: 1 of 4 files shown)


src/libstore/store-api.cc
```diff
@@ -226,16 +226,37 @@ ValidPathInfo Store::addToStoreSlow(std::string_view name, const Path & srcPath,
...
 
-    auto [narHash, narSize] = hashPath(htSHA256, srcPath);
+    TeeSink sinkIfNar { narHashSink, caHashSink };
 
-    auto hash = method == FileIngestionMethod::Recursive
-        ? hashAlgo == htSHA256
-          ? narHash
-          : hashPath(hashAlgo, srcPath).first
-        : hashFile(hashAlgo, srcPath);
+    /* We use the tee sink if we need to hash he nar twice */
```
edolstra, line 237:
```suggestion
    /* We use the tee sink if we need to hash the nar twice */
```

src/libstore/store-api.cc
```diff
@@ -226,16 +226,37 @@ ValidPathInfo Store::addToStoreSlow(std::string_view name, const Path & srcPath,
...
+    });
+
+    TeeSource tapped { *fileSource, sink };
+
+    ParseSink blank;
+    auto & parseSink = method == FileIngestionMethod::Flat
+        ? fileSink
+        : blank;
+
+    parseDump(parseSink, tapped);
```
Ericson2314, line 253: If sha256 and recursive, don't need to use `tapped` can just use the original `*fileSource`

Ericson2314:
```suggestion
    parseDump(
        parseSink,
        method == FileIngestionMethod::Recursive && hashAlgo == htSHA256
            ? *fileSource // don't need to hash twice if we just can use the `narHash` twice
            : tapped);
```

Issue #3804, opened 2020-07-12
builtins.match stackoverflows for nixpkgs's nixos.revision


Mic92:
**Describe the bug**

When having enough branches in git than `builtins.match` will stackoverflow.
This leads infinite recursion errors when build nixos from git.

**Steps To Reproduce**

The source is `lib.commitIdFromGitRepo` which causes stack overflows on my machine with the following input by calling `builtins.match` with the following input:

```nix
builtins.match ''.*
([^
 ]*) refs/heads/master
.*'' ''# pack-refs with: peeled fully-peeled sorted
1245688ae717951615a676cc60882422e3ddb230 refs/heads/master
d7ff6ab94aae4deadbe26d20900e8a0d19ebc860 refs/remotes/origin/acme
7225ce8f4bf78167f32ef5370b0673abff66c757 refs/remotes/origin/acpi-call
2571c87a6c97b69995d6cbe8fde5c65264477db1 refs/remotes/origin/aerc
4358dfff8d66106319985e5cbd7f9d379d42b823 refs/remotes/origin/appdaemon
4c26d0ec5f6ccc32ebf1b39b69d243d28bd4e9a0 refs/remotes/origin/arm-flakes
777b80ff26ee0fbeec1736654daab8e3c5476ec7 refs/remotes/origin/backport-nix-build-uncached
f03d1058d904440822297eac77be5f0c6ca8a8e9 refs/remotes/origin/backport/release-18.09/pr-46034
0bfada7dda710e1a1a99ab67639eb1deb981cbed refs/remotes/origin/bcc
f3b68a724edcddc4cf62b614a1a3b03fce357da8 refs/remotes/origin/bitwarden-cli
4f5e0efb3c66cd1dff03ca1916295bf1d44c7754 refs/remotes/origin/blsd
eae491cabb17855eceb11f13415f2da58449d85c refs/remotes/origin/breakpoint-hook
97ac11f6a3563c1b50034d1079a208411803ef0b refs/remotes/origin/bsld
59387b1f24160b3ce2192cf3579a3ac29abab7ac refs/remotes/origin/build-go-package
9a1b767a550be8b044b5db3f32dbb270973204c5 refs/remotes/origin/build-inputs
8b7f4fa8a64b141cd103689789909a30a7b972d1 refs/remotes/origin/buildkite
1021fe523207fb17f91940e7484a932d7ffb05e1 refs/remotes/origin/bundler-env
[...]

Issue #1537, opened 2017-08-29
Regex behavior differs across platforms


copumpkin:
I seem to remember seeing some regex-related changes in 1.12 so this might no longer be an issue, but I want to capture it in case it is. If I fire up `nix-repl` on 1.11.x on Linux and Darwin and run

```nix
builtins.match "^((|\..*)\.sw[a-z]|.*~)$" "foo"
```

Darwin will complain that `error: compiling pattern ‘^((|..*).sw[a-z]|.*~)$’: empty (sub)expression`, whereas Linux will return `null`.

copumpkin:
This is slightly painful because that happens to be the regex we use to filter junk from `nixpkgs` in the standard NixOS image builder, and I need to be able to _evaluate_ the image build process on Darwin 😦 

cc @edolstra

copumpkin:
Yeah, so I guess we moved from 1.11's calls to `regcomp` which I guess varies from platform to platform, to `std::regex` in 1.12, which I assume is consistent across platforms.

copumpkin:
@edolstra do you have some sort of plan for the cutover to 1.12 with differing behavior in `builtins.match`? Do we need to bump the minimum Nix version on nixpkgs to make it work? Otherwise, silently changing behavior could cause a lot of woe.

edolstra:
I don't have a plan. I would suggest rewriting that regexp into something that doesn't rely on Linux-specific behavior. 

Also, `builtins.match` was undocumented in 1.11 so using it was at your own risk anyway ;-)

edolstra:
Anyway, the behavior no longer differs across platforms, so I'll close this.

clacke:
Still an issue on 2.0pre:

Darwin:

```
$ nix-instantiate --version
nix-instantiate (Nix) 2.0pre5968_a6c0b773
$ nix-instantiate --eval -E 'builtins.match "(|.*/)([^/]*)" "path/to/blah"'
error: invalid regular expression '(|.*/)([^/]*)', at (string):1:1
```

Linux:

```
$ nix-instantiate --version
nix-instantiate (Nix) 2.0pre5914_48c192ca
$ nix-instantiate --eval -E 'builtins.match "(|.*/)([^/]*)" "path/to/blah"'
[ "path/to/" "blah" ]
```

clacke:
Verified that Darwin nix 2.0 also has this issue.

copumpkin:
That's surprising! The full source to `match` is this:

[...]

Issue #1331, opened 2017-04-14
builtins.match: \s and \S do not work in 1.12


volth:
```builtins.match ''\s+(\S+)\s+'' " foo "``` works in ```nix-1.11.8``` but fails with ```error: regex_error``` in ```nix-1.12pre5122_c60715e```

edolstra:
This is probably a consequence of the switch to `std::regex` in b05b98df7544d02387f583ca5434f33f3e9cb471.

edolstra:
`builtins.match` supports POSIX extended regular expressions, which AFAIK don't support `\s` and `\S`. Apparently the libc RE functions (which we used prior to b05b98d) do support them, but this is undocumented (see https://linux.die.net/man/7/regex) and not portable. For instance, on OS X, your example returns `null`.

I've added a note to the manual documenting that `builtins.match` support POSIX extended REs.

volth:
There was also difference in interpreting ``` +``` (space, plus).
It worked in 1.11 but for 1.12 I had to change it to ```[ ]+``` (openbracket, space, closebracket, plus).
Introducing this limited version of regex is very confusing.

PR #3826, opened 2020-07-17
switch from std::regex to boost::regex

Mic92:
1. std::regex is not consistent across different platforms (libcxx vs libstdc++)
2. libstdc++ implementation is braindead:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93502
   and overflows the stack if the input string gets too long.

If this change is accepted I will also replace all other instances.

### DIFF
--- a/src/libexpr/eval.hh
+++ b/src/libexpr/eval.hh
@@ -11,6 +11,7 @@
 #include <optional>
 #include <unordered_map>
 #include <mutex>
+#include <boost/regex.hpp>
 
 
 namespace nix {
@@ -119,7 +120,7 @@ private:
     std::unordered_map<Path, Path> resolvedPaths;
 
     /* Cache used by prim_match(). */
-    std::unordered_map<std::string, std::regex> regexCache;
+    std::unordered_map<std::string, boost::regex> regexCache;
 
 public:
 
--- a/src/libexpr/local.mk
+++ b/src/libexpr/local.mk
@@ -15,7 +15,7 @@ libexpr_CXXFLAGS += -I src/libutil -I src/libstore -I src/libfetchers -I src/lib
 
 libexpr_LIBS = libutil libstore libfetchers
 
-libexpr_LDFLAGS =
+libexpr_LDFLAGS = -lboost_regex
 ifneq ($(OS), FreeBSD)
  libexpr_LDFLAGS += -ldl
 endif
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -18,7 +18,7 @@
 
 #include <algorithm>
 #include <cstring>
-#include <regex>
+#include <boost/regex.hpp>
 #include <dlfcn.h>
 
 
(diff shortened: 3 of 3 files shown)


edolstra:
I'm not really in favor of this. `boost::regex` is just the predecessor of `std::regex` so going back to `boost::regex` seems like a step back. We'd be pulling in an external dependency for something that's already provided by the standard library.

BTW what string sizes are we talking about here?

Mic92:
Depends on the platform's default stacksize limit but on x86_64-linux this leads to reliable crashes: https://github.com/NixOS/nix/issues/3804

edolstra:
Looks like this may get fixed upstream: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86164#c8

Mic92:
> I'm not really in favor of this. `boost::regex` is just the predecessor of `std::regex` so going back to `boost::regex` seems like a step back. We'd be pulling in an external dependency for something that's already provided by the standard library.
> 
> BTW what string sizes are we talking about here?

It does not really increase the closure size though in Nix. There might be some distributions such as Debian that package liboost_regex separately: https://packages.debian.org/sid/amd64/libboost-regex1.71-dev/filelist

What is your proposed fix instead? fixing libstdc++'s regex is beyond my capability.

Mic92:
> Looks like this may get fixed upstream: [gcc.gnu.org/bugzilla/show_bug.cgi?id=86164#c8](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86164#c8)

Not really the commenter gave up.

Mic92:
> I'm not really in favor of this. `boost::regex` is just the predecessor of `std::regex` so going back to `boost::regex` seems like a step back. We'd be pulling in an external dependency for something that's already provided by the standard library.
> 
> BTW what string sizes are we talking about here?

28k in my case.

edolstra:
Note that libboost_regex.so is about 1.3 MB, but worse, it pulls in libicu4c which is 34 MB.

conferno:
maybe PCRE?
it is small, fast and consistent

Mic92:
I think it used `regcomp` before: https://github.com/NixOS/nix/commit/b05b98df7544d02387f583ca5434f33f3e9cb471#diff-52d20285486e90d564ca2535d4d0fe55L11

It could be reverted to that. However its too much guessing from my side what the maintainers actually want as a acceptable solution so I just leave it as it is.

zimbatm:
+1 for PCRE, oniguruma or any other regexp engine that has a consistent evaluation across platforms.

adisbladis:
@zimbatm While I would absolutely _love_ PCRE the problem is that Nix is already using POSIX ERE and migrating to something else would break `builtins.match`.

conferno:
buildins.match was already broken about 2 years ago as part of @shlevy
refactoring.
There are few abandoned tickets about that.
POSIX is recent introduction, it broke few things of existing usage of
buildings.match, but there was not a tragedy.

conferno:
#1331 #1537 and more linked from there

PR #3829, opened 2020-07-17
Remove storetype delegate reg store -- contains #3736

meditans:
to each Store implementation. The generic regStore implementation will
only be for the ambiguous shorthands, like "" and "auto".

This also could get us close to simplifying the daemon command.


src/nix-daemon/nix-daemon.cc
```diff
@@ -277,7 +278,9 @@ static int _main(int argc, char * * argv)
         initPlugins();
 
         if (stdio) {
-            if (getStoreType() == tDaemon) {
+            if (openUncachedStore().dynamic_pointer_cast<UDSRemoteStore>()) {
```
edolstra, line 281: Doesn't this cause the store to be opened twice in the non-UDSRemoteStore case? Maybe it's better to do `store = openUncachedStore(); if (store.dynamic_pointer_cast ...) ... `.

Ericson2314: @edolstra Ah I remember now: I wanted to do that, but I wasn't sure how to destroy the opened store in the `UDSRemoteStore` before reopening it, since `nix::ref` is always non-null.

Ericson2314: The connections are opened on demand, so this was never a problem, but the new way also uses the store that was opened/downcast so there is no duplicate work.

roberth:
It seems like the diff still contains changes from #3736 (merged). Rebasing on master will make the review easier.

Ericson2314:
Thanks for catching @roberth! I can't edit the commit message but I did fix the diff.

Ericson2314:
@edolstra OK fixed up post-merge so this is ready again.

Ericson2314:
Thanks!!

PR #3830, opened 2020-07-17
Mischellaneous changes to prepare for CA drvs -- contains #3807, #3424, and #3418

Ericson2314:
This doesn't actually implement any new features, :astonished:. What it does do is adjusting existing code to better take advantage of the new variants. I hope by essentially "pushing down" the `throw Error("not implemented")` closer to where the algorithsm actually need to be generalized, it becomes easier for people to tackle them independently.


src/libstore/build.cc
```diff
@@ -3721,9 +3721,24 @@ void DerivationGoal::registerOutputs()
            hash). */
         std::optional<ContentAddress> ca;
 
-        if (fixedOutput) {
+        if (! std::holds_alternative<DerivationOutputInputAddressed>(i.second.output)) {
+            DerivationOutputFloating outputHash;
+            std::visit(overloaded {
+                [&](DerivationOutputInputAddressed doi) {
+                    throw Error("No.");
```
edolstra, line 3728: ?

src/libstore/local-store.cc
```diff
@@ -544,35 +544,43 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
...
+                    throw Error("derivation '%s' has incorrect output '%s', should be '%s'",
+                        printStorePath(drvPath), printStorePath(doia.path), printStorePath(recomputed));
+                envHasRightPath(doia.path, i.first);
+            },
+            [&](DerivationOutputFixed dof) {
+                StorePath path = makeFixedOutputPath(dof.hash.method, dof.hash.hash, drvName);
+                envHasRightPath(path, i.first);
+            },
+            [&](DerivationOutputFloating _) {
+                throw Error("Floating CA output derivations are not yet implemented");
```
edolstra, line 581: Maybe we should have an `Unimplemented` exception type. This would make it easier to grep for these.

src/libstore/derivations.hh
```diff
@@ -6,17 +6,48 @@
...
-struct DerivationOutput
+struct DerivationOutputInputAddressed
 {
+    /* Will need to become `std::optional<StorePath>` once input-addressed
+       derivations are allowed to depend on cont-addressed derivations */
     StorePath path;
-    std::optional<FixedOutputHash> hash; /* hash used for expected hash computation */
+};
+
+struct DerivationOutputFixed
```
Ericson2314, line 24: Should the variant be only of new types, or should it include old ones if they are 1-1?

src/libstore/derivations.hh
```diff
@@ -6,17 +6,48 @@
...
+{
+    FixedOutputHash hash; /* hash used for expected hash computation */
+};
+
+struct DerivationOutputFloating
+{
+    /* information used for expected hash computation */
+    FileIngestionMethod method;
+    HashType hashType;
```
Ericson2314, line 33: Should we actually have this flexibility, or should we just enforce NAR-sha256?

Ericson2314: If we do remove these fields, we can still encode drv files in a way that allows for adding them later

src/libstore/build.cc
Ericson2314:
```suggestion
                    assert(false); // Enclosing `if` handles this case in other branch
```

src/libstore/local-store.cc
meditans: `UnimplementedError` is now available, thanks!

Ericson2314:
Conflicts fixed.

Ericson2314:
Conflicts fixed

PR #3833, opened 2020-07-18
flakes: rename devShell to developPackages

Mic92:
The cli was changed but not the flake format.
It's not clear to a user that both names belong together.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -414,7 +414,7 @@
 
       defaultPackage = forAllSystems (system: self.packages.${system}.nix);
 
-      devShell = forAllSystems (system:
+      developmentEnv = forAllSystems (system:
         with nixpkgsFor.${system};
         with commonDeps pkgs;
 
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -422,7 +422,11 @@ struct CmdFlakeCheck : FlakeCommand
                             }
                         }
 
-                        else if (name == "defaultPackage" || name == "devShell") {
+                        else if (name == "defaultPackage" || name == "developmentEnv" || name == "devShell") {
+                            if (name == "devShell") {
+                                warn("flake output 'devShell' was replaced by 'developmentEnv'");
+                            }
+
                             state->forceAttrs(vOutput, pos);
                             for (auto & attr : *vOutput.attrs) {
                                 checkSystemName(attr.name, *attr.pos);
@@ -821,7 +825,7 @@ struct CmdFlakeShow : FlakeCommand
 
                     logger->stdout("%s: %s '%s'",
                         headerPrefix,
-                        attrPath.size() == 2 && attrPath[0] == "devShell" ? "development environment" :
+                        attrPath.size() == 2 && (attrPath[0] == "developmentEnv" && attrPath[0] == "devShell") ? "development environment" :
                         attrPath.size() == 3 && attrPath[0] == "checks" ? "derivation" :
                         attrPath.size() >= 1 && attrPath[0] == "hydraJobs" ? "derivation" :
                         "package",
@@ -831,6 +835,7 @@ struct CmdFlakeShow : FlakeCommand
                 if (attrPath.size() == 0
                     || (attrPath.size() == 1 && (
                             attrPath[0] == "defaultPackage"
+                            || attrPath[0] == "developmentEnv"
                             || attrPath[0] == "devShell"
                             || attrPath[0] == "nixosConfigurations"
                             || attrPath[0] == "nixosModules"
@@ -846,7 +851,7 @@ struct CmdFlakeShow : FlakeCommand
                 }
 
                 else if (
-                    (attrPath.size() == 2 && (attrPath[0] == "defaultPackage" || attrPath[0] == "devShell"))
+                    (attrPath.size() == 2 && (attrPath[0] == "defaultPackage" || attrPath[0] == "developmentEnv" || attrPath[0] == "devShell"))
                     || (attrPath.size() == 3 && (attrPath[0] == "checks" || attrPath[0] == "packages"))
                     )
                 {
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -205,7 +205,7 @@ struct Common : InstallableCommand, MixProfile
 
     Strings getDefaultFlakeAttrPaths() override
     {
-        return {"devShell." + settings.thisSystem.get(), "defaultPackage." + settings.thisSystem.get()};
+        return {"developmentEnv." + settings.thisSystem.get(), "devShell." + settings.thisSystem.get(), "defaultPackage." + settings.thisSystem.get()};
     }
 
     StorePath getShellOutPath(ref<Store> store)


flake.nix
```diff
@@ -414,7 +414,7 @@
 
       defaultPackage = forAllSystems (system: self.packages.${system}.nix);
 
-      devShell = forAllSystems (system:
+      developPackages = forAllSystems (system:
```
zimbatm, line 417:
```suggestion
      developPackage = forAllSystems (system:
```
Logically this should be a singular as there is only one value per system available.

src/nix/flake.cc
```diff
@@ -422,7 +422,11 @@ struct CmdFlakeCheck : FlakeCommand
                             }
                         }
 
-                        else if (name == "defaultPackage" || name == "devShell") {
+                        else if (name == "defaultPackage" || name == "developPackages" || name == "devShell") {
+                            if (name == "devShell") {
+                                warn("flake output 'devShell' is deprecated. Use by 'developPackages' instead");
```
zimbatm, line 427:
```suggestion
                                warn("flake output 'devShell' is deprecated. Use 'developPackages' instead");
```

edolstra:
While `devShell` is not a great name, I'm not convinced `developPackage` is better:

* It's not a package, it's a derivation that configures a development environment. In fact there is no requirement that `devShell` actually builds, let alone produces a package.

* I have trouble parsing `developPackage`. Is `develop` short for development, or is it a verb in the imperative tense?

Maybe we want something like `developmentEnvironment`?

Kloenk:
> Maybe we want something like `developmentEnvironment`?

I think that's a rather long name. Non native English speaker could have a problem spelling that.

Mic92:
> > Maybe we want something like `developmentEnvironment`?
> 
> I think that's a rather long name. Non native English speaker could have a problem spelling that.

I don't think so, both words are quite common in IT/English. Someone who has trouble spelling this out will have problems understanding the nix manual in the first place.

zimbatm:
Actually, we are making the same mistake again and conflating the various use-cases, just like `nix-shell`.

`nix develop` right now is similar to `nix-shell` in that it loads the various inputs of a derivation amd creates an environment to debug that package. In that regard, `devShell` should be dropped entirely and use the `defaultPackage` only.

Then there is another use-case, which is to create a development environment for a repository. This is related but not entirely the same thing. Usually, in my experience, the development doesn't care about making build phases available to the user. The main facility is to bring the compilers, interpreters, linting tools, ... and add them to the PATH. Then there might be some other environment variable being set, but those are development specific (instead of package-specific).
That repository might be a monorepo with many languages and packages, so `nix develop` is not good enough.

Mic92:
> Actually, we are making the same mistake again and conflating the various use-cases, just like `nix-shell`.
> 
> `nix develop` right now is similar to `nix-shell` in that it loads the various inputs of a derivation amd creates an environment to debug that package. In that regard, `devShell` should be dropped entirely and use the `defaultPackage` only.
> 
> Then there is another use-case, which is to create a development environment for a repository. This is related but not entirely the same thing. Usually, in my experience, the development doesn't care about making build phases available to the user. The main facility is to bring the compilers, interpreters, linting tools, ... and add them to the PATH. Then there might be some other environment variable being set, but those are development specific (instead of package-specific).
> That repository might be a monorepo with many languages and packages, so `nix develop` is not good enough.
> 
> For that reason, I would propose to deprecate `devShell` simply, and then come up with a better solution for the second case.

What would this look like? Is `mkShell` not the appropriate replacement for it?

blaggacao:
I totally agree with @zimbatm the help text of `nix develop` tells tales. -> `nix hack`? It's not a descriptive name but aligned with intention.

There is also a ambiguity on the `nix develop` side of the cli.

- `nix hack ./myproject` -> I want to hack on "myproject"
- `nix develop ./myproject` -> I want to develop on "myproject".

What's the difference?

7c6f434c:
>> > Maybe we want something like `developmentEnvironment`?
>> 
>> I think that's a rather long name. Non native English speaker could have a problem spelling that.
>
>I don't think so, both words are quite common in IT/English. Someone who has trouble spelling this out will have problems understanding the nix manual in the first place.

Well, being comfortable with typing an un-spell-checkable combination of two long words and being comfortable reading long-form texts might be correlated but are distinct things

edolstra:
>  That repository might be a monorepo with many languages and packages, so nix develop is not good enough.

I'm not following, why would `nix develop` have a problem with having many languages and packages?

> For that reason, I would propose to deprecate devShell simply, and then come up with a better solution for the second case.

Let's not deprecate things until we have something beter. Hopefully `flake.toml` will give us a more declarative way to specify developer environments without all the cruft like derivations and build phases.

> So nix develop might be just plain confusing.

There is no perfect terminology but `nix hack` is just as confusing if not more so. `nix develop` conveys that it's intended to set up a development environment (unlike say `nix shell` or `nix run`).

blaggacao:
> `nix develop` conveys that it's intended to set up a development environment

Totally agree! I just was giving the code priority when interpreting, see:
https://github.com/NixOS/nix/blob/master/src/nix/develop.cc#L261

That description ought to be fixed then so it's crystal clear to the world:

- `nix develop ./myproject` sets up a development environment without the cruft.

Mic92:
So should I just rename it to `developmentEnvironment` now?

blaggacao:
Maybe the `develop` command needs to be fixed first to concisely clarify it's semantics. Then this one can comply. Right now, it's not clear with what ultimate intepretation to comply with.

And as @zimbatim said conflating the use cases is a recipe for problems.

Mic92:
> Maybe the `develop` command needs to be fixed first to concisely clarify it's semantics. Then this one can comply. Right now, it's not clear with what ultimate intepretation to comply with.

This sounds more like a different PR to add documentation to `nix develop`.

Mic92:
I renamed it to `developmentEnv`. It's not as horrible long as `developmentEnvironment` and `env` should be a pretty common abbreviation of `environment` (i.e. [direnv](https://github.com/direnv/direnv) or [setenv(3)](https://man7.org/linux/man-pages/man3/setenv.3.html)

Mic92:
> Actually, we are making the same mistake again and conflating the various use-cases, just like `nix-shell`.
> 
> `nix develop` right now is similar to `nix-shell` in that it loads the various inputs of a derivation amd creates an environment to debug that package. In that regard, `devShell` should be dropped entirely and use the `defaultPackage` only.
> 
> Then there is another use-case, which is to create a development environment for a repository. This is related but not entirely the same thing. Usually, in my experience, the development doesn't care about making build phases available to the user. The main facility is to bring the compilers, interpreters, linting tools, ... and add them to the PATH. Then there might be some other environment variable being set, but those are development specific (instead of package-specific).
> That repository might be a monorepo with many languages and packages, so `nix develop` is not good enough.
> 
> For that reason, I would propose to deprecate `devShell` simply, and then come up with a better solution for the second case.

I think the problem here is what people put in `developmentEnv`. I think it's still possible to put a new abstraction behind this attributes that not just loads stdenv but instead export saner environment variables. We don't have this abstraction yet but I don't see how it's appearance would conflict with the current flake format in a backwards-incompatible way.

blaggacao:
I was pondering on the two use cases a little during the last days, with no constraints or holy cows.

This is where my mind drifted towards: https://discourse.nixos.org/t/entice-the-developer/8310

zimbatm:
I think I'd rather keep `devShell` for now. `developmentEnv` doesn't map 1:1 to `nix develop` so it's not a marked improvement.

PR #3840, opened 2020-07-20
README: improve development docs

Mic92:


edolstra:
Maybe we can link to the development section in the manual instead? Duplicating the same info in multiple places doesn't seem ideal.

Mic92:
The manual that I found is for the released version of nix. We cannot put it there because is no flakes yet. Is there a link to the latest build of the manual? Otherwise I would suggest to remove the hacking guide from the manual and bundle with the source instead. It does not make sense to point users to out-of-date guides.

edolstra:
There is https://hydra.nixos.org/job/nix/master/build.x86_64-linux/latest/download-by-type/doc/manual#chap-hacking. Once we convert to Markdown we can link to the manual directly on GitHub...

doc/manual/hacking.xml
```diff
@@ -4,18 +4,31 @@
...
 the latest version of Nix from GitHub:
 <screen>
-$ git clone git://github.com/NixOS/nix.git
+$ git clone https://github.com/NixOS/nix.git
 $ cd nix
 </screen>
 </para>
 
-<para>To build it and its dependencies:
+<para>To build it Nix using one of the targets provided by <literal>default.nix</literal>:
```
edolstra, line 15:
```suggestion
<para>To build Nix using one of the targets provided by <literal>default.nix</literal>:
```

doc/manual/hacking.xml
```diff
@@ -4,18 +4,31 @@
...
+$ nix-build -A outputs.hydraJobs.build.x86_64-linux
+$ nix-build -A outputs.hydraJobs.build.aarch64-linux
+$ nix-build -A outputs.hydraJobs.build.x86_64-darwin
+$ nix-build -A outputs.hydraJobs.build.i686-linux
 </screen>
+
+or if you have a flakes-enabled nix:
+
+<screen>
+$ nix build ".#hydraJobs.build.x86_64-linux"
```
edolstra, line 26: Worth mentioning that you can do `nix build` without any arguments to get a build for the current platform.

Mic92: Simplified.

doc/manual/hacking.xml
```diff
@@ -4,18 +4,37 @@
...
 the latest version of Nix from GitHub:
 <screen>
-$ git clone git://github.com/NixOS/nix.git
+$ git clone https://github.com/NixOS/nix.git
 $ cd nix
 </screen>
 </para>
 
-<para>To build it and its dependencies:
+<para>To build Nix for the current operating system/architecture use</para>
```
edolstra, line 15:
```suggestion
<para>To build Nix for the current operating system/architecture use
```

Kloenk:
Great improvement. But when I try to hack something, I don't want to open my web browser. So I would prefer it in the readme, even if then it is duplicated.

This enables me to just cat README.md, and then start hacking.

Mic92:
Well, it seems like we change to some other document format at some point. Also docbook is supposed to be a plaintext format too.

Kloenk:
> Well, it seems like we change to some other document format at some point. Also docbook is supposed to be a plaintext format too.

Fair enough. Then maybe add a path to the file in the readme? So you can find it directly with cat.

Mic92:
> > Well, it seems like we change to some other document format at some point. Also docbook is supposed to be a plaintext format too.
> 
> Fair enough. Then maybe add a path to the file in the readme? So you can find it directly with cat.

It will be replaced once the conversion happens.

Issue #3841, opened 2020-07-20
Small error/logging refactor


Ericson2314:
Currently There is an `ErrorInfo` struct which

1. Contains a verbosity level
2. Is a field of `BaseError`

I think this causes problems:

1. It doesn't make sense to associate exceptions with verbosity levels: an exception is a failure that aborts the computation. (It's the *handler* / catch block that decide whether something can keep going and isn't fatal.)

2. There is no way to render an error different based on the verbosity level: @edolstra and I on a call talked bout making HTTP error response bodies show up with higher log levels, this doesn't appear very easy.

Here is what I tentatively  propose:

1. `ErrorInfo` is really about logging messages, which may or may not have anything to do with exceptions/control flow. It should be renamed to `LogEntry`, and put in `logging,hh`.
    - `error.hh` would import `logging.hh` instead of the other way around
    - `enum Verbosity` can also move back

2. Rather than storing an `ErrorInfo`, `BaseError` should have virtual method:
    ```c++
    // Render the exception into a log entry containing more or less information bes
    virtual LogEntry render(Verbosity v) = 0;
    ```
    which will allow exceptions to be rendered with different amount of information based on the `-vvv`.

Part 1 I definitely like, part 2 I am not sure about because the `render` should always return a `lvlError`, right? so the `Verbosity` parameter (how much information is in the error) is used rather incongruously.

bburdette:
I don't object to renaming ErrorInfo, or moving it around.  It IS about the logging.  To me the main advantage with ErrorInfo is we're getting structured data to the logger rather than just strings.  

Right now, though, the ErrorInfo is not extensible.  Extensibility is just an extra complication if its not really needed, but maybe its necessary.  The ErrorInfo has structured data in it for Nix language errors, but it doesn't have special fields for anything else - http in this instance.  

Having a render function in exceptions would mean that you would need to store all the information that goes into an ErrorInfo (or 'LogEntry') in the exception, so that you could build an ErrorInfo later and pass it to the logger.  Just storing the ErrorInfo in the first place seems like a clean way to go to me, assuming it can contain the data you care about.

If you only care about rendering http response codes into a hint string in the log, you can do it by checking verbosity when creating the ErrorInfo.  If verbosity is high, put the code into the hint, if its low leave it out.  That doesn't require code changes in the ErrorInfo/Logger.

[...]

PR #3847, opened 2020-07-21
 libutil/logging: extend `internal-json` logger to make it more machine-readable

Ma27:
The new error-format is pretty nice from a UX point-of-view, however
it's fairly hard to parse the output e.g. for editor plugins such as
vim-ale[1] that use `nix-instantiate --parse` to determine syntax errors in
Nix expression files.

This patch extends the `internal-json` logger by adding the fields
`line`, `column` and `file` to easily locate an error in a file and the
field `raw_msg` which contains the error-message itself without
code-lines and additional helpers.

An exemplary output may look like this:

```
[nix-shell]$ ./inst/bin/nix-instantiate ~/test.nix --log-format minimal
{"action":"msg","column":1,"file":"/home/ma27/test.nix","level":0,"line":4,"raw_msg":"syntax error, unexpected IF, expecting $end","msg":"<full error-msg with code-lines etc>"}
```

[1] https://github.com/dense-analysis/ale

### DIFF
--- a/src/libutil/logging.cc
+++ b/src/libutil/logging.cc
@@ -184,6 +184,33 @@ struct JSONLogger : Logger {
         json["action"] = "msg";
         json["level"] = ei.level;
         json["msg"] = oss.str();
+        json["raw_msg"] = ei.hint->str();
+
+        if (ei.errPos.has_value() && (*ei.errPos)) {
+            json["line"] = ei.errPos->line;
+            json["column"] = ei.errPos->column;
+            json["file"] = ei.errPos->file;
+        } else {
+            json["line"] = nullptr;
+            json["column"] = nullptr;
+            json["file"] = nullptr;
+        }
+
+        if (loggerSettings.showTrace.get() && !ei.traces.empty()) {
+            nlohmann::json traces = nlohmann::json::array();
+            for (auto iter = ei.traces.rbegin(); iter != ei.traces.rend(); ++iter) {
+                nlohmann::json stackFrame;
+                stackFrame["raw_msg"] = iter->hint.str();
+                if (iter->pos.has_value() && (*iter->pos)) {
+                    stackFrame["line"] = iter->pos->line;
+                    stackFrame["column"] = iter->pos->column;
+                    stackFrame["file"] = iter->pos->file;
+                }
+                traces.push_back(stackFrame);
+            }
+
+            json["trace"] = traces;
+        }
 
         write(json);
     }
--- a/doc/manual/command-ref/opt-common.xml
+++ b/doc/manual/command-ref/opt-common.xml
@@ -106,7 +106,14 @@
     </varlistentry>
 
     <varlistentry><term>internal-json</term>
-    <listitem><para>Outputs the logs in a structured manner. NOTE: the json schema is not guarantees to be stable between releases.</para></listitem>
+     <listitem>
+      <para>Outputs the logs in a structured manner.</para>
+      <warning>
+       <para>
+        While the schema itself is relatively stable, the format of the error-messages (namely of the <literal>msg</literal>-field) can change between several releases.
+       </para>
+      </warning>
+     </listitem>
     </varlistentry>
 
     <varlistentry><term>bar</term>
--- a/src/libutil/tests/logging.cc
+++ b/src/libutil/tests/logging.cc
@@ -34,6 +34,24 @@ namespace nix {
         }
     }
 
+    TEST(logEI, jsonOutput) {
+        SymbolTable testTable;
+        auto problem_file = testTable.create("random.nix");
+        testing::internal::CaptureStderr();
+
+        makeJSONLogger(*logger)->logEI({
+                .name = "error name",
+                .description = "error without any code lines.",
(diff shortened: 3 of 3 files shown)


Kloenk:
Is Jason a better name, because it's in a Jason format?

Ma27:
Fair point. Can we agree on `minimal-json` though? That's IMHO better to make it distinguishable from `internal-json` (which I intentionally didn't choose as it basically puts the entire, pretty-printed error into a JSON).

cc @edolstra WDYT?

edolstra:
I don't really understand why we need another JSON output format when we already have one. What information can't you get from the existing format?

Ma27:
@edolstra the current json format returns a fully detailed message containing code-lines etc. This is IMHO fairly hard to parse for editor-plugins where I only need the line-number, line-column and message (which is provided in single JSON-fields in my new logger).

Also, we probably *want* to use the full-blown human-readable message in the `internal-json` logger (AFAIU) since that's how log-messages are received from a remote builder.

Finally, the `internal-json` logger is subject to change between releases (according to the man-pages) which would mean that plugins have to be modified for each new Nix release (in the worst case).

I see two options to take care of it:

* provide another logger for the purpose I described (which is currently the case in this PR)
* modify the `internal-json` logger to define (1) a somewhat stable structure and (2) expose error-msg (without code-lines), line-nr and line-column as additional json fields.

I'm fine with either of those options, so feel free to decide which you prefer.

Ma27:
cc @bburdette who also might be interested in reviewing this as they implemented the beautiful new error-messages.

edolstra:
I think it's better to improve the existing JSON logger than to add another one. The whole point of the JSON logger was that it's the last logger we'll ever need, since it should provide all information in a format that other tools can easily process.

> Finally, the internal-json logger is subject to change between releases

I don't think that should be the case. Having a log format that can change in arbitrary ways is not very useful. I think it's more that clients should be prepared to handle message types or fields being added or removed. (Conversely though, who says that the a new logger wouldn't be subject to change?)

> Also, we probably want to use the full-blown human-readable message in the internal-json logger (AFAIU) since that's how log-messages are received from a remote builder.

No, the daemon should send structured error messages to the client. Currently errors from the daemon don't get formatted very well. For instance, the daemon doesn't know the terminal width of the client. And it's hard for the client to add context to an error received from the daemon.

> modify the internal-json logger to define (1) a somewhat stable structure and (2) expose error-msg (without code-lines), line-nr and line-column as additional json fields.

I think this is the best approach. The code lines can probably be put in a separate JSON field which tools like IDEs can ignore.

Ma27:
> Conversely though, who says that the a new logger wouldn't be subject to change?

Well, first of all it was explicitly stated in the man-pages. Also, error-information is only provided in the `msg` field which can be an arbitrary string (which is in *this* case a problem as we'd currently have to parse e.g. line-numbers etc. from it).

> I think this is the best approach. The code lines can probably be put in a separate JSON field which tools like IDEs can ignore.

That's also perfectly fine for me. I'll update the PR later today, thanks for your feedback!

Ma27:
@edolstra updated accordingly %)

Ericson2314:
Maybe take a look at what i wrote in https://github.com/NixOS/nix/issues/3841 ? I think this separation of rendering and data would also be good for machine-readable logging; currently we bake a lot of data in the hint which is conceptually separate, and should be actually separate for machine-readable logging.

bburdette:
I like it!  I could see adding the trace info into the json format if at some point there's a tool that needs it.

Ma27:
@bburdette fair point, added traces as well.
@Ericson2314 AFAIU this is something where we'd have to refactor the error-API first, right? After reading through the issue I'd generally agree with you, but this seems slightly out of scope for this PR (unless I misunderstood your point).
@edolstra anything else tbd? :)

Ericson2314:
Yes I suppose this changes are independent of this PR.

src/libutil/logging.cc
```diff
@@ -184,6 +184,33 @@ struct JSONLogger : Logger {
...
+
+        if (ei.errPos.has_value() && (*ei.errPos)) {
+            json["line"] = ei.errPos->line;
+            json["column"] = ei.errPos->column;
+            json["file"] = ei.errPos->file;
+        } else {
+            json["line"] = -1;
+            json["column"] = -1;
+            json["file"] = nullptr;
+        }
```
Ericson2314, line 197: I don't think `-` is a great sentinel value. I rather do one of:

 1. Make suboject, stored at `pos`, which is nullable.
 2. Use `null` for each as the sentinel value
 3. Skip fields altogether rather than using sentinel value

in decreasing order of preference.

Ma27: @Ericson2314 I decided to set `null` by default here since AFAIU the `pos` struct *can* be undeclared, so with this we have a fair sentinel value and are on the safe side IMHO :)

Ma27:
afaics all comments are resolved. Anything else to add @edolstra?

Ma27:
ping @edolstra anything else tbd to get this merged?

Ma27:
ping @edolstra @Ericson2314

PR #3851, opened 2020-07-23
Parse CA derivations using new output variants

Ericson2314:
We no longer need `ParsedDerivation` for this because everything libstore needs to know about is in the `BasicDerivation` proper.


src/libexpr/primops.cc
```diff
@@ -639,6 +640,11 @@ static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * *
...
             }
 
+            if (i->name == state.sContentAddressed) {
+                settings.requireExperimentalFeature("ca-derivations");
+                contentAddressed = state.forceBool(*i->value, pos);
+            }
+
             /* The `args' attribute is special: it supplies the
                command-line arguments to the builder. */
             if (i->name == state.sArgs) {
```
edolstra, line 650:
```suggestion
            else if (i->name == state.sArgs) {
```

PR #3854, opened 2020-07-23
Save changes made by "nix registry pin" to user registry

B4dM4n:
Changes made by `nix registry pin` are currently not written to the user registry and therefore lost when nix exits.


edolstra:
Thanks!

PR #3856, opened 2020-07-23
Make `Buildable` a `std::variant`

Ericson2314:
I think this better captures the intent of what's going on: we either
have an opaque store path, or a drv path with some outputs.

Having this structure will also help us support CA derivations: we'll
have to allow the outpath paths to be optional, so the structure we gain
now makes up for the structure we loose then.


src/nix/command.cc
```diff
@@ -131,11 +131,18 @@ void MixProfile::updateProfile(const Buildables & buildables)
...
 
     for (auto & buildable : buildables) {
-        for (auto & output : buildable.outputs) {
-            if (result)
-                throw Error("'--profile' requires that the arguments produce a single store path, but there are multiple");
-            result = output.second;
-        }
+        std::visit(overloaded {
+            [&](BuildableOpaque bo) {
+                result = bo.path;
```
edolstra, line 136: Need to check that there is only one `result` here.

Ericson2314: Fixed this, and with less duplication, by accumulating everything in a vector, and the throwing an error if the vector's length is not exactly 1.

PR #3857, opened 2020-07-24
Convert manual to Markdown

edolstra:
This PR converts the Nix manual from DocBook to Markdown.

The manual is currently rendered using [`mdbook`](https://github.com/rust-lang/mdBook), which is a Rust tool used to format books like [The Rust Programming Language](https://doc.rust-lang.org/book/). You can see the output here: https://hydra.nixos.org/job/nix/markdown/build.x86_64-linux/latest/download-by-type/doc/manual

Manpages are generated from Markdown using [`lowdown`](https://kristaps.bsd.lv/lowdown/). The output isn't great but it's good enough.

Most of the conversion was done using Pandoc with some manual cleanups afterwards. Please let me know if you spot any conversion errors.


Ma27:
While I acknowledge that this will lower the barrier to contribute to our manuals, shouldn't we wait until people came to an agreement in NixOS/rfcs#72 a majority is fine with?

Profpatsch:
The lowdown page states (in a footnote):

> You may be tempted to write manpages in Markdown, but please don't: use mdoc(7), instead — it's built for that purpose! The man output is for technical documentation only (section 7).

edolstra:
I know, but if the goal is to make it easier to contribute to the docs, then switching from DocBook to mdoc/troff would be insane.

dtzWill:
In case an expression for lowdown is needed, I've been sitting on this: https://github.com/dtzWill/nixpkgs/blob/nixos-dtz/pkgs/tools/text/lowdown/default.nix :innocent:.

edolstra:
@dtzWill Thanks, do you want to add it to nixpkgs? If so you may want to copy the multiple outputs that the expression in flake.nix has.

sternenseemann:
Added a PR for a lowdown package with multiple outputs in #94225.

doc/manual/src/command-ref/nix-build.md
```diff
@@ -0,0 +1,105 @@
...
+    $ nix-build -E 'with import <nixpkgs> { }; runCommand "foo" { } "echo bar > $out"'
+    $ cat ./result
+    bar
+
+Build the GNU Hello package from the latest revision of the master
+branch of Nixpkgs:
+
+    $ nix-build https://github.com/NixOS/nixpkgs/archive/master.tar.gz -A hello
+
+# Environment variables
```
Profpatsch, line 105: Variable list include is missing.

Probably no includes where used?

edolstra: Yes, Markdown doesn't have a standard include mechanism so I dropped it. If we really want this, we can hack something.

sternenseemann: I guess using some kind of preprocessor is the simplest since it wouldn't be necessary to parse any custom markdown. [mdsh](https://github.com/zimbatm/mdsh#link-includes) has such a feature for example.

edolstra:
Some more improvements:

* The `nix` command now has a manpage (though most of the subcommands don't have a description yet, just an automatically generated list of flags).

* The `nix.conf` manpage / manual section is now generated from the C++ sources. So we no longer have two sets of option descriptions.

* Similarly, the manual section that lists the builtin functions is generated from the C++ sources. You can now also do `:doc builtins.<name>` to get documentation about a builtin function (rendered using the `lowdown` library).

edolstra:
Unless there are major objections I'd like to merge this (without prejudice to whatever markup language / tools people end up using for Nixpkgs/NixOS). Having a manpage for `nix` and the `:doc` command in `nix repl` alone is worth merging this.

src/libexpr/primops.cc
```diff
@@ -897,6 +1156,9 @@ static void prim_toPath(EvalState & state, const Pos & pos, Value * * args, Valu
    corner cases. */
 static void prim_storePath(EvalState & state, const Pos & pos, Value * * args, Value & v)
 {
+    if (evalSettings.pureEval)
+        throw EvalError("builtins.storePath' is not allowed in pure evaluation mode");
+
```
infinisil, line 1161: Discovered by @9999years..

fricklerhandwerk: It's been there before, but [in a slightly different form](https://github.com/NixOS/nix/commit/d4dcffd64349bb52ad5f1b184bee5cc7c2be73b4#diff-0f59bb6f197822ef9f19ceae9624989499d170c84dfdc1f486a8959bb4588cafR2061-R2062).

9999years: Committed directly to `master` lol 👨‍🍳💋

PR #3858, opened 2020-07-24
Group 'nix' subcommands

edolstra:
As suggested in https://gist.github.com/gilligan/a0942a490c5ab5b854cc5ad668374529, this PR puts various `nix` subcommands underneath `nix store`, `nix hash` and `nix nar`. This unclutters the top-level interface.

I didn't move `nix path-info` since I don't know the best name (e.g. `nix store path-info`? `nix store info`?).

### DIFF
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -86,6 +86,7 @@ void Args::parseCmdline(const Strings & _cmdline)
                 throw UsageError("unrecognised flag '%1%'", arg);
         }
         else {
+            pos = rewriteArgs(cmdline, pos);
             pendingArgs.push_back(*pos++);
             if (processArgs(pendingArgs, false))
                 pendingArgs.clear();
@@ -390,10 +391,6 @@ MultiCommand::MultiCommand(const Commands & commands)
         .optional = true,
         .handler = {[=](std::string s) {
             assert(!command);
-            if (auto alias = get(deprecatedAliases, s)) {
-                warn("'%s' is a deprecated alias for '%s'", s, *alias);
-                s = *alias;
-            }
             if (auto prefix = needsCompletion(s)) {
                 for (auto & [name, command] : commands)
                     if (hasPrefix(name, *prefix))
--- a/src/libutil/args.hh
+++ b/src/libutil/args.hh
@@ -115,6 +115,9 @@ protected:
 
     virtual bool processArgs(const Strings & args, bool finish);
 
+    virtual Strings::iterator rewriteArgs(Strings & args, Strings::iterator pos)
+    { return pos; }
+
     std::set<std::string> hiddenCategories;
 
 public:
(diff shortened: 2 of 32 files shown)


Kloenk:
Sounds grate. But I think this creates more a need for things like `alias` as supported by git.

Ericson2314:
I was thinking `hash to-*` should be like `hash convert --to *`. Once we merge the PR with the parsing changes we can have a `--from` too.

davidak:
>I didn't move nix path-info since I don't know the best name (e.g. nix store path-info? nix store info?).

>Summary: query information about store paths

I guess `nix store path-info` would be intuitive then. `nix store info` sounds like general information about the store, like total size, number of packages...

matthewbauer:
I think this would be confusing since we already have `nix-store` and `nix-hash`. If you wanted to add man pages for any of those commands it would run into collisions since `man nix store` and `man nix-store` are equivalent (see `--no-subpages` in https://man7.org/linux/man-pages/man1/man.1.html).

Also, `nix doctor`, `nix upgrade-nix`, and `nix show-config` seem pretty fundamental commands, hiding them behind "system" seems bad.

Ma27:
Apart from @matthewbauer's comment I think that while this might look pretty, it may be fairly hard to remember which command is under which two(!) parents which is something at least I'd be rather unhappy with during daily use.

Also I'd like to mention that `git` - which was IIRC one of the idols for the introduction of the three-letter `nix`-command - has most of the comments directly at "top-level".

edolstra:
> Also, nix doctor, nix upgrade-nix, and nix show-config seem pretty fundamental commands, hiding them behind "system" seems bad.

I agree, which is why I didn't add a `nix system` command.

> Also I'd like to mention that git - which was IIRC one of the idols for the introduction of the three-letter nix-command - has most of the comments directly at "top-level".

Sure, but the git CLI is also widely considered to be a bit of a mess. And it does have second-level commands like `git remote` and `git submodules`. But it's pretty inconsistent about it, e.g. `git branch` has various suboperations, only they're identified using flags (e.g. `git branch -d`) rather than subcommands.

> If you wanted to add man pages for any of those commands it would run into collisions since man nix store and man nix-store are equivalent 

Hm yeah, that's pretty annoying...

Ma27:
> And it does have second-level commands like git remote and git submodules

Yes that's the case for special features of `git` (like it's the case for `nix flake` in Nix for instance). To me it seems like this change would hide several features under one "category".

The great thing about tools like git is that I can press `git <TAB>` and I get a completeable list of available subcommands. With this change, I have to found out again under which "category" those are (similar to the `nix-*` cmds at the moment).

FRidh:
> I didn't move nix path-info since I don't know the best name (e.g. nix store path-info? nix store info?).

Maybe `nix store query-path`? It says query in the `--help`.

edolstra:
Need to update this to be in line with #4302.

PR #3859, opened 2020-07-24
`queryDerivationOutputMap` no longer assumes all outputs have a mapping

Ericson2314:
This assumption is broken by CA derivations. Making a PR now to do the
breaking daemon change as soon as possible (if it is already too late,
we can bump protocol instead).

CC @regnat


src/libstore/build.cc
```diff
@@ -2756,8 +2756,12 @@ struct RestrictedStore : public LocalFSStore
     void queryReferrers(const StorePath & path, StorePathSet & referrers) override
     { }
 
-    OutputPathMap queryDerivationOutputMap(const StorePath & path) override
-    { throw Error("queryDerivationOutputMap"); }
+    std::map<std::string, std::optional<StorePath>> queryDerivationOutputMap(const StorePath & path) override
+    {
+        if (!goal.isAllowed(path))
+            throw InvalidPath("cannot query output map for unknown path '%s' in recursive Nix", printStorePath(path));
+        return next->queryDerivationOutputMap(path);
```
Ericson2314, line 2763: Doesn't matter until we actually implement CA derivations, but we should censor returned output paths too.

src/libstore/worker-protocol.hh
```diff
@@ -70,6 +70,56 @@ template<class T> T readStorePaths(const Store & store, Source & from);
...
+void write(const Store & store, Sink & out, const std::optional<T> & optVal)
+{
+    out << (optVal ? 1 : 0);
+    if (optVal)
+        write(store, out, *optVal);
+}
+
+StorePath read(const Store & store, Source & from, Proxy<StorePath> _);
+
+void write(const Store & store, Sink & out, const StorePath & storePath);
```
Ericson2314, line 123: While the generic machinery is a bit overkill for just one problem, I would like to keep this stuff to make the daemon protocol less annoying to work with in the future.

Ericson2314:
OK, this is now backwards compatible, and thus is ready to be merged.

src/libstore/remote-store.cc
```diff
@@ -57,30 +57,36 @@ void writeStorePathCAMap(const Store & store, Sink & out, const StorePathCAMap &
...
-        out << store.printStorePath(i.second);
-    }
+    out << store.printStorePath(storePath);
+}
+
+
+template<>
+std::optional<StorePath> read(const Store & store, Source & from, Phantom<std::optional<StorePath>> _)
+{
+	auto s = readString(from);
```
edolstra, line 77:
```suggestion
    auto s = readString(from);
```

tab -> spaces

src/libstore/store-api.hh
```diff
@@ -340,10 +340,16 @@ public:
...
-    virtual OutputPathMap queryDerivationOutputMap(const StorePath & path)
+    /* Query the mapping outputName => outputPath for the given derivation. All
+       outputs are mentioned so ones mising the mapping are mapped to
+       `std::nullopt`.  */
+    virtual std::map<std::string, std::optional<StorePath>> queryDerivationOutputMap(const StorePath & path)
     { unsupported("queryDerivationOutputMap"); }
 
+    /* Query the mapping outputName=>outputPath for the given derivation.
+       Assume every output has a mapping and throw an exception otherwise. */
+    OutputPathMap queryDerivationOutputMapAssumeTotal(const StorePath & path);
```
edolstra, line 351:
```suggestion
    OutputPathMap queryDerivationOutputMap(const StorePath & path);
```

src/libstore/store-api.hh
```diff
@@ -340,10 +340,16 @@ public:
     /* Query the outputs of the derivation denoted by `path'. */
     virtual StorePathSet queryDerivationOutputs(const StorePath & path);
 
-    /* Query the mapping outputName=>outputPath for the given derivation */
-    virtual OutputPathMap queryDerivationOutputMap(const StorePath & path)
+    /* Query the mapping outputName => outputPath for the given derivation. All
+       outputs are mentioned so ones mising the mapping are mapped to
+       `std::nullopt`.  */
+    virtual std::map<std::string, std::optional<StorePath>> queryDerivationOutputMap(const StorePath & path)
```
edolstra, line 346:
```suggestion
    virtual std::map<std::string, std::optional<StorePath>> queryPartialDerivationOutputMap(const StorePath & path)
```

PR #3880, opened 2020-07-30
Add "bundle" command to Nix

### LINKED ISSUE #3705: Integrate nix-bundle into new Nix command
With the new "app" installables in the flakes branch, we have a way to keep track of program entry points. This seems like it would be a great fit for a "bundle" or "pack" command in Nix.

I think lots of inspiration can come from how guix pack works: https://hpc.guix.info/blog/2020/05/faster-relocatable-packs-with-fakechroot/.

guix has three "execution engines":

- unprivileged user namespaces (like in nix-bundle)
- PRoot (ptrace-based)
- Fakechroot (LD_PRELOAD-based)

We may also want to support different bundle formats, like:

- Docker / OCI
- Arx (used for nix-bundle)
- AppImage (uses squashfs)
- VirtualBox image

Most of these would have to link in Nixpkgs somehow, to get the right tools, which may be a little tricky.

matthewbauer:
This adds a ‘nix bundle’ command which hooks into [nix-bundle](https://github.com/matthewbauer/nix-bundle). It can
be used in a similar way as nix-bundle, with the benefit of hooking
into the new “app” functionality. For instance,

```
$ nix bundle nixpkgs#jq
$ ./jq --help
jq - commandline JSON processor [version 1.6]
...
$ scp jq machine-without-nix:
$ ssh machine-without-nix ./jq
jq - commandline JSON processor [version 1.6]
...
```

Note that nix-bundle currently requires Linux to run. Other bundlers
might not have that requirement.

“bundlers” are meant to be reusable, so that, other repos can
implement their own bundling.

Fixes #3705

### DIFF
--- a/src/nix/bundle.cc
+++ b/src/nix/bundle.cc
@@ -6,18 +6,18 @@
 
 using namespace nix;
 
-struct CmdExport : InstallableCommand
+struct CmdBundle : InstallableCommand
 {
-    std::string exporter = "github:matthewbauer/nix-bundle";
+    std::string bundler = "github:matthewbauer/nix-bundle";
     Path outLink;
 
-    CmdExport()
+    CmdBundle()
     {
         addFlag({
-            .longName = "exporter",
-            .description = "use custom exporter",
+            .longName = "bundler",
+            .description = "use custom bundler",
             .labels = {"flake-url"},
-            .handler = {&exporter},
+            .handler = {&bundler},
             .completer = {[&](size_t, std::string_view prefix) {
                 completeFlakeRef(getStore(), prefix);
             }}
@@ -35,15 +35,15 @@ struct CmdExport : InstallableCommand
 
     std::string description() override
     {
-        return "export an application out of the Nix store";
+        return "bundle an application so that it works outside of the Nix store";
     }
 
     Examples examples() override
     {
         return {
             Example{
-                "To export Hello:",
-                "nix export hello"
+                "To bundle Hello:",
+                "nix bundle hello"
             },
         };
     }
@@ -71,12 +71,12 @@ struct CmdExport : InstallableCommand
         auto app = installable->toApp(*evalState);
         store->buildPaths(app.context);
 
-        auto [exporterFlakeRef, exporterName] = parseFlakeRefWithFragment(exporter, absPath("."));
+        auto [bundlerFlakeRef, bundlerName] = parseFlakeRefWithFragment(bundler, absPath("."));
         const flake::LockFlags lockFlags{ .writeLockFile = false };
-        auto exporter = InstallableFlake(
-            evalState, std::move(exporterFlakeRef),
-            Strings{exporterName == "" ? ("defaultExporter." + settings.thisSystem.get()) : exporterName},
-            Strings({"exporters." + settings.thisSystem.get() + "."}), lockFlags);
+        auto bundler = InstallableFlake(
+            evalState, std::move(bundlerFlakeRef),
+            Strings{bundlerName == "" ? ("defaultBundler." + settings.thisSystem.get()) : bundlerName},
+            Strings({"bundlers." + settings.thisSystem.get() + "."}), lockFlags);
 
         Value * arg = evalState->allocValue();
         evalState->mkAttrs(*arg, 1);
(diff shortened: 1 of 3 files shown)


grahamc:
Would this be applicable to something like the nixos-generators project? This looks like something that would require fairly substantial docs.

grahamc:
Oops, that sounds pretty down on the idea. I'm super in to it. I'd love to be able to easily export closures in some fashion, especially being pluggable.

matthewbauer:
> Would this be applicable to something like the nixos-generators project? This looks like something that would require fairly substantial docs.

So this doesn't 100% fit into nixos-generators, but I think we might be able to tweak the exporters format to handle this. Right now, we are expecting the flake to be an "App", like is expected in `nix run`. nixos-generators wants a NixOS configuration. Flakes do know about nixos configurations, so it is definitely possible to add support. We'd probably want to separate them so there's no ambiguity. So we could have "appExporters" and "nixosExporters".

zimbatm:
I like that the exporter is itself a derivation so we can keep this composable.

src/nix/export.cc
```diff
@@ -0,0 +1,126 @@
...
+
+        Bindings::iterator i = vRes->attrs->find(evalState->sDrvPath);
+        if (i == vRes->attrs->end())
+            throw Error("the exporter '%s' does not produce a derivation", exporter.what());
+
+        PathSet context2;
+        StorePath drvPath = store->parseStorePath(evalState->coerceToPath(*i->pos, *i->value, context2));
+
+        i = vRes->attrs->find(evalState->sOutPath);
+        if (i == vRes->attrs->end())
```
edolstra, line 103: We have a nicer idiom for this: `auto attr = vRes->attrs->get(evalState->sOutPath); if (!attr) throw ...`.

src/nix/export.cc
```diff
@@ -0,0 +1,126 @@
...
+#include "shared.hh"
+#include "store-api.hh"
+#include "fs-accessor.hh"
+
+using namespace nix;
+
+struct CmdExport : InstallableCommand
+{
+    std::string exporter = "github:matthewbauer/nix-bundle";
+    Path outLink;
```
edolstra, line 12: `std::optional<Path>`

src/nix/export.cc
```diff
@@ -0,0 +1,126 @@
+#include "command.hh"
+#include "common-args.hh"
+#include "shared.hh"
+#include "store-api.hh"
+#include "fs-accessor.hh"
+
+using namespace nix;
+
+struct CmdExport : InstallableCommand
```
edolstra, line 9: The name `export` is pretty vague and overloaded (e.g. we already have a `nix-store --export`). It's worth noting that this command does not actually export anything by itself since the build result is in the Nix store.

Maybe `bundle` (though that's not really correct either)? 

Or an "apply" flag to `nix build` analogous to `nix eval --apply`, e.g. `nix build --apply-generator github:matthewbauer/nix-bundle hello`.

src/nix/export.cc
```diff
@@ -0,0 +1,126 @@
...
+
+        auto app = installable->toApp(*evalState);
+        store->buildPaths(app.context);
+
+        auto [exporterFlakeRef, exporterName] = parseFlakeRefWithFragment(exporter, absPath("."));
+        const flake::LockFlags lockFlags{ .writeLockFile = false };
+        auto exporter = InstallableFlake(
+            evalState, std::move(exporterFlakeRef),
+            Strings{exporterName == "" ? ("defaultExporter." + settings.thisSystem.get()) : exporterName},
+            Strings({"exporters." + settings.thisSystem.get() + "."}), lockFlags);
```
edolstra, line 79: Given that this is a function anyway, we don't need to use the `.${system}` attribute. We can just pass `system` as a function argument. So the interface of the exporter would be `{ system, program, ... }: ...`.

src/nix/export.cc
```diff
@@ -0,0 +1,126 @@
...
+            throw Error("the exporter '%s' does not produce a derivation", exporter.what());
+
+        StorePath outPath = store->parseStorePath(evalState->coerceToPath(*i->pos, *i->value, context2));
+
+        store->buildPaths({{drvPath}});
+
+        auto accessor = store->getFSAccessor();
+        auto outPathS = store->printStorePath(outPath);
+        if (accessor->stat(outPathS).type != FSAccessor::tRegular)
+            throw Error("'%s' is not a file; an exporter must only create a single file", outPathS);
```
edolstra, line 113: Not sure if this restriction is really useful. I can imagine some exporters producing multiple files (e.g. I seem to remember that instance-store AMIs are a directory of files).

edolstra:
BTW `nix run` now also works on regular packages, e.g. `nix run nixpkgs#gimp` works. Maybe the exporter should be applied to an store path rather than an app. Whether the exporter can handle a package containing multiple programs is up to the exporter.

matthewbauer:
Renamed from "export" to "bundle" for now. Open to other names (or consolidation with another command). I like having it be top-level as it's quite similar in function to "build" / "run".

src/nix/export.cc
matthewbauer: Yeah, I thought export would imply it doesn't just have to be nix-bundle proper, but it is a little confusing. I suppose you could make an exporter / bundler that just produces a .nar though.

matthewbauer:
> BTW `nix run` now also works on regular packages, e.g. `nix run nixpkgs#gimp` works. Maybe the exporter should be applied to an store path rather than an app. Whether the exporter can handle a package containing multiple programs is up to the exporter.

I like having a single entry point for these things since it means you don't have to guess what to run. At the very least, something like this would need hints of which store path to run. Right now "program" is just a path with context to an executable.

src/nix/bundle.cc
```diff
@@ -0,0 +1,127 @@
...
+#include "common-args.hh"
+#include "shared.hh"
+#include "store-api.hh"
+#include "fs-accessor.hh"
+
+using namespace nix;
+
+struct CmdBundle : InstallableCommand
+{
+    std::string bundler = "github:matthewbauer/nix-bundle";
```
zimbatm, line 11: is the plan to integrate this with nixpkgs or another shared github org?

PR #3883, opened 2020-07-31
Can build single CA derivations

Ericson2314:
Featurewise, we're testing little more than what https://github.com/NixOS/nix/pull/3740 can do, but a behind the scenes there is a lot of infra changes in place to make this robust:

 - Via #3830 the legacy style build hashes are thoroughly excised from the floating CA derivation case--almost everywhere Nix now either handles the case where output paths are not known a priori, or throws an `UnimplementedError` saying it cannont. (I say "almost" because there is one more issue to sort out with `src/libexpr/get-drvs.hh:getOutPath`.)

-  `DerivationGoal::registerOutputs` has been almost completely rewritten. Before it would first calculate content hashes and then references, because we never had references in the CA case. Now, it calculates references and then hashes, so we can do both for CA derivations. This is also integrated with the existing rewriting/relocating we did for `--check` and `--repair`.

The new functionality is ready to go. WIP because of conflicts and because it depends on PRs which still have rough edges.


Ericson2314:
OK I now rebased away some PRs this PR previously contained that it doesn't actually needed. That means I can also remove the WIP!

src/libstore/build.cc
```diff
@@ -3689,162 +3835,314 @@ void DerivationGoal::registerOutputs()
...
+                            wanted.to_string(SRI, true),
+                            got.to_string(SRI, true)));
+                }
+            }
 
-                if (worker.store.isValidPath(dest))
-                    std::rethrow_exception(delayedException);
+            assert(newInfo0.ca);
+            return newInfo0;
+        }}();
```
regnat, line 4062: Possibly rather subjective, but I'd find the control flow easier to follow if it was structured more like:

```cpp
auto newInfoFromCA = [&](const DerivationOutputCAFloating outputHash) -> ValidPathInfo {
// everything from L3999 to L4061
};
ValidPathInfo newInfo = [&]() std::visit(overloaded {
  [&](DerivationOutputInputAddressed outputP) { ... };
  [&](DerivationOutputCAFixed dof) { return newInfoFromCA(DerivationOutputCAFloating { ... }); };
  [&](DerivationOutputCAFloating dof) { return newInfoFromCA(DerivationOutputCAFloating dof; };
})
```

src/libstore/build.cc
```diff
@@ -3689,162 +3835,314 @@ void DerivationGoal::registerOutputs()
...
-                    BuildError("hash mismatch in fixed-output derivation '%s':\n  wanted: %s\n  got:    %s",
-                        worker.store.printStorePath(dest),
-                        h.to_string(SRI, true),
-                        h2.to_string(SRI, true)));
-              } else {
-                  // matched the fixed hash, so no move needed.
-                  needsMove = false;
-              }
+            rewriteOutput();
+            /* FIXME optimize and deduplicate with addToStore */
```
regnat, line 4009: More than just `addToStore`, this is morally a duplicate of `CmdMakeContentAddressable::run` (modulo a couple of details like the fact that the original path isn't a valid path)

src/libstore/build.cc
```diff
@@ -3950,7 +4246,13 @@ void DerivationGoal::registerOutputs()
        outputs, this will fail. */
     {
         ValidPathInfos infos2;
-        for (auto & i : infos) infos2.push_back(i.second);
+        for (auto & [outputName, newInfo] : infos) {
+            /* FIXME: we will want to track this mapping in the DB whether or
```
regnat, line 4250: I might have forgotten something obvious, but is there anything preventing us from calling `linkDeriverToPath` without a derivation? (obviously we don't have the `drvPath`, but afaik we can recompute it since derivations are content-addressed)

src/libstore/derivations.cc
```diff
@@ -15,7 +15,8 @@ std::optional<StorePath> DerivationOutput::pathOpt(const Store & store, std::str
         },
         [&](DerivationOutputCAFixed dof) -> std::optional<StorePath> {
             return {
-                store.makeFixedOutputPath(dof.hash.method, dof.hash.hash, drvName)
+                // FIXME if we intend to support multiple CA outputs.
+                dof.path(store, drvName, "out")
```
regnat, line 19: Doesn't this invalidate the hash of existing CA derivations?

Ericson2314: I think this does exactly what the old code did, just factored out?

src/libstore/build.cc
Ericson2314: We have a `BasicDerivation` not `Derivation` in the general case. We don't currently have an `unparse` for those, so it's just a matter of bike-shdding whether we want to take the opportunity to switch to JSON or anything else while we're at it.

src/libstore/build.cc
```diff
@@ -3693,162 +3835,314 @@ void DerivationGoal::registerOutputs()
...
+                    /* FIXME build inverted map up front so no quadratic waste here */
+                    for (auto & r : refs)
+                        for (auto & [o, p] : scratchOutputs)
+                            if (r == p)
+                                referencedOutputs.insert(o);
+                    return referencedOutputs;
+                },
+            }, x);
+        }},
+        {[&](const std::string & path, const std::string & parent) {
```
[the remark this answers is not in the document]
Ericson2314: This prohibition already exists, or at least my intent was not to add any new ones.

src/libstore/build.cc
```diff
@@ -1426,14 +1510,20 @@ void DerivationGoal::tryToBuild()
...
+    /* Obtain locks on all output paths, if the paths are known a priori.
+
+       The locks are automatically released when we exit this function or Nix
+       crashes.  If we can't acquire the lock, then continue; hopefully some
+       other goal can start a build, and if not, the main loop will sleep a few
+       seconds and then retry this goal. */
     PathSet lockFiles;
-    for (auto & outPath : drv->outputPaths(worker.store))
-        lockFiles.insert(worker.store.Store::toRealPath(outPath));
+    /* FIXME: Should lock something like the drv itself so we don't build same
```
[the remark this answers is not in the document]
Ericson2314: When it goes to move the output to it's CA path, it takes a file lock and checks whether it already is registered, so I think we're good.

src/libstore/build.cc
```diff
@@ -3642,47 +3784,47 @@ void DerivationGoal::registerOutputs()
...
+            : worker.store.toRealPath(p);
+    };
+
+    /* Check whether the output paths were created, and make all
+       output paths read-only.  Then get the references of each output (that we
+       might need to register), so we can topologically sort them. For the ones
+       that are most definitely already installed, we just store their final
+       name so we can also use it in rewrites. */
+    StringSet outputsToSort;
+    std::map<std::string, std::variant<StorePath, StorePathSet>> outputReferences;
```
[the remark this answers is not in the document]
Ericson2314: Good idea! Did it.

src/libstore/build.cc
Ericson2314: Agreed. I just think those details, while conceptually insignificant, will make abstracting over this and `CmdMakeContentAddressable::run` harder. I am thinking start with making `addToStoreFromDump` references-aware, maybe then merge with `addTextToStore`, and only then re-compare with `CmdMakeContentAddressable::run`.

src/libstore/build.cc
```diff
@@ -2199,8 +2370,14 @@ void DerivationGoal::startBuilder()
            rebuilding a path that is in settings.dirsInChroot
            (typically the dependencies of /bin/sh).  Throw them
            out. */
-        for (auto & i : drv->outputs)
-            dirsInChroot.erase(worker.store.printStorePath(i.second.path(worker.store, drv->name)));
+        for (auto & i : drv->outputs) {
+            /* If the name isn't known a prior (i.e. floating content-addressed
+               derivation), the temporary location we use should be fresh and
```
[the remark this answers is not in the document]
Ericson2314: Yes this was very confusing! Hopefully it is better now.

src/libstore/remote-store.cc
```diff
@@ -70,6 +70,20 @@ void write(const Store & store, Sink & out, const StorePath & storePath)
     out << store.printStorePath(storePath);
 }
 
+
+template<>
+std::optional<StorePath> read(const Store & store, Source & from, Phantom<std::optional<StorePath>> _)
+{
+	auto s = readString(from);
+    return s == "" ? std::optional<StorePath> {} : store.parseStorePath(s);
```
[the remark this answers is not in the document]
Ericson2314: Maybe....? The comparability considerations are numerous and it's hard to keep track of them all. I hope when we discuss in person we can try to get a grasp on those.

> that's a pretty big and dense diff

If we can get a grasp on those, maybe we can merge #3895 first, and then this PR should be easier to review.

src/libstore/build.cc
```diff
@@ -1452,32 +1542,35 @@ void DerivationGoal::tryToBuild()
        omitted, but that would be less efficient.)  Note that since we
        now hold the locks on the output paths, no other process can
        build this derivation, so no further checks are necessary. */
-    validPaths = checkPathValidity(true, buildMode == bmRepair);
-    if (buildMode != bmCheck && validPaths.size() == drv->outputs.size()) {
+    bool allValid = true;
+    for (auto & [_, status] : initialOutputs) {
```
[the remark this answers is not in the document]
Ericson2314: Good point. I was thinking we already had the information but didn't consider that it could change. We just acquired some locks above this, so this a good time to check (again).

src/libstore/build.cc
Ericson2314: Ah great idea! I agree this is better. And furthermore, the code that checks the wanted vs got can be moved out of the `newInfoFromCA` and into the `DerivationOutputCAFixed` case, avoiding yet another `std::get_if`.

src/libstore/build.cc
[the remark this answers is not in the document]
regnat: Oh indeed it's already forbidden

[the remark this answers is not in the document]
regnat: Oh indeed it's already forbidden

src/libstore/derivations.cc
regnat: Ah yes, didn't notice that "out" was special-cased so that `outputPathName(foo, "out") == foo`

src/libstore/build.cc
regnat: Yep', makes sense. That's roughly what I'd done in b959baee6d51b44e51c8896c302ccd11fd0a458a (except that I don't use `addToStoreFromDump` because I wanted to keep the original pathInfo)

src/libstore/derivations.cc
Ericson2314: We since landed https://github.com/NixOS/nix/pull/3875 which should allow doing this better, and including working hypothetical multi-output fixed-output CA derivations by not assuming `"out"`.

Ericson2314: FIXME is now gone.

src/libstore/build.cc
Ericson2314: #3958 shows we don't need a new `unparse`, but there still is the question of whether to either:

 - relax foreign key constraint, so we can link without file
 - write drv file to store so we can link with current foreign key constraint

I have no string preference. My only concern is we might want to keep around any mapping in the *build* closure of other mappings, which changes GC semantics.

src/libstore/build.cc
Ericson2314: Yeah, I have found the C++ non-trivial to deduplicate too. I think it's better to land the features and *then* choose the abstractions to make sure we don't abstract the wrong way----the plethora of `addToStore`-like functions kindof implies we've already been abstracting suboptimally, so I think a rethink should be very broad and systematic.

Ericson2314:
@edolstra @regnat Is there anything more we need to do before landing this?

regnat:
> @edolstra @regnat Is there anything more we need to do before landing this?

Might be worth a larger-scale test just to make sure that we don't screw anything (though I don't know what the expectations of stability are on the master branch), but other than that nope.

I tried to simplify https://github.com/NixOS/nix/pull/3883#discussion_r468606846, but the naive version ended-up being more complex than the current thing (mostly because atm `make-content-addressed` only handle recursive-sha256 hashing which simplifies things a lot), and simplifying it ended in a rabbit-hole of refactorings that don't seem to be leading anywhere. So I think we can keep things as they are for the moment

regnat:
This is causing issues with an older daemon : 

With my system nix daemon (2.3.6), I get

```console
$ ./inst/bin/nix-build '<nixpkgs>' -A hello
error: --- Error ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ nix-build
invalid operation 41
```

41 corresponds to `wopQueryDerivationOutputMap`.
(I'll double-check that this wasn't already the case on master though)

Ericson2314:
@regnat my guess is the underlying issue is on master, but we might only trigger with this branch, as it uses it a lot more.

regnat:
> @regnat my guess is the underlying issue is on master, but we might only trigger with this branch, as it uses it a lot more.

Digged this a bit further:

The underlying issue is kind-of on master because using `RemoteStore::queryDerivationOutputMap` will systematically fail on an old daemon. 
This was somehow intended because :
1. we don't use it on master, except as part of `queryDerivationOutputPaths` which fallbacks if the daemon is too old
2. my intent when introducing it was to only use it for floating-ca derivations (which require a recent daemon anyways), which isn't the case here (for example in `nix-build.cc` it's called unconditionally).

That being said, using it only for floating-ca drvs is a bit of a pain (because at that point we shouldn't care whether the derivation is CA or not).

I was hoping that we could introduce a fallback directly inside `RemoteStore::queryDerivationOutputMap` that would fetch the drv from the remote store and try to parse it (and fail if it's floating-ca), but `getDerivation` isn't implemented for `RemoteStore` so that wouldn't work.

We could add a workaround that would call `queryDerivationOutputNames` and `queryDerivationOutputs` and zip them (which might work in practice because at least in 2.3.6 the implementation seems to guaranty that they will be in the same order), but that's quite ugly.

Ericson2314:
> I was hoping that we could introduce a fallback directly inside `RemoteStore::queryDerivationOutputMap` that would fetch the drv from the remote store and try to parse it (and fail if it's floating-ca), but `getDerivation` isn't implemented for `RemoteStore` so that wouldn't work.

I like your plan. What `getDerivation` is this? I think the protocol should allow fetching the underlying drv file like any other store path, right?

regnat:
> I like your plan. What `getDerivation` is this? I think the protocol should allow fetching the underlying drv file like any other store path, right?

Not right out of the box : 

- There's a store method called `getDerivation`, but that's not implemented for `RemoteStore`
- There's also `Store::narFromPath` that's also not implented for `RemoteStore`, **but** the worker protocol has already the logic to handle it (because it's used by `SSHStore`), so we just need to upstream `SSHStore::narFromPath` into `RemoteStore` and it should be good. That's a lot of hops for such a workaround, but hopefully that should work

regnat:
I tried to bootstrap gcc with this branch (without marking anything as CA, just to ensure backwards compat) and uncovered a couple issues:

1. Fetching the patch of `attr` failed with a hash mismatch
2. The build of `binutils` and `gcc` (the wrappers around `binutils-unwrapped` and `gcc-unwrapped`) fail with 
    ```
    changing mode of '/tmp/store/nix/store/d1rj0n6y0bysiddjrbsf9niap94vdxmz-binutils-wrapper-2.31.1.drv.chroot/nix/store/qd6nbwmx0hsp8k6z4p8vf55z80f8sf1s-binutils-wrapper-2.31.1-info' to 120777: No such file or directory
    ```

The first error is a bit fishy because https://github.com/NixOS/nixpkgs/commit/f36899e11826b18fdd60496b28af692afbea8b85 seems to indicate that the hash error is legit, but running it with my system `nix-build` doesn't show the error. So I'm not sure what to think about it.

The tests were ran with `NIX_PATH` set to `nixpkgs=https://github.com/nixos/nixpkgs/archive/e9bd19d4d352f45e871cfc3e91f854cb76ff48e1.tar.gz`, running `./inst/bin/nix-build '<nixpkgs>' -A gcc --store /tmp/store --option substitute false` (where `/tmp/store` was empty at the begining). To bypass the failures and get the build going I built the failing drvs with my system nix (also with substitution disabled so that I could make sure that the build was expected to suceed) and restarted from that state.

regnat:
>     2. The build of `binutils` and `gcc` (the wrappers around `binutils-unwrapped` and `gcc-unwrapped`) fail with
>        ```
>        changing mode of '/tmp/store/nix/store/d1rj0n6y0bysiddjrbsf9niap94vdxmz-binutils-wrapper-2.31.1.drv.chroot/nix/store/qd6nbwmx0hsp8k6z4p8vf55z80f8sf1s-binutils-wrapper-2.31.1-info' to 120777: No such file or directory
>        ```

This is raised at https://github.com/NixOS/nix/blob/c224a5e5c1e0a55c8e8e7c3c1c0e50ac831f9964/src/libstore/build.cc#L4105

It probably mostly happens in a chroot store : The reason is that the file in question is a symlink (to `/nix/store/blahblah-binutils-2.31.1-info`), and the symlink isn't valid outside of the chroot (but `chmod` tries to follow it).

The following patch fixes it:

```diff
diff --git a/src/libstore/build.cc b/src/libstore/build.cc
index ba28e78c8..e01ce9d62 100644
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -2065,7 +2065,7 @@ void linkOrCopy(const Path & from, const Path & to)
            file (e.g. 32000 of ext3), which is quite possible after a
            'nix-store --optimise'. FIXME: actually, why don't we just
            bind-mount in this case?
-           
+
            It can also fail with EPERM in BeegFS v7 and earlier versions
            which don't allow hard-links to other directories */
         if (errno != EMLINK && errno != EPERM)
@@ -4101,7 +4101,11 @@ void DerivationGoal::registerOutputs()
                     if (lstat(actualPath.c_str(), &st))
                         throw SysError("getting attributes of path '%1%'", actualPath);
                     mode |= 0200;
-                    if (chmod(actualPath.c_str(), mode) == -1)
+                    // Try to change the perms, but only if the file isn't a
+                    // symlink as symlinks permissions are mostly ignored
+                    // and calling `chmod` on it will just forward the call to
+                    // the target of the link
+                    if (!S_ISLNK(st.st_mode) && chmod(actualPath.c_str(), mode) == -1)
                         throw SysError("changing mode of '%1%' to %2$o", actualPath, mode);
                 }
                 if (rename(
```

tests/content-addressed.nix
```diff
@@ -0,0 +1,32 @@
...
+    buildCommand = ''
+      set -x
+      echo "Building a legacy derivation"
+      mkdir -p $out
+      echo "Hello World" > $out/hello
+    '';
+    __contentAddressed = true;
+    outputHashMode = "recursive";
+    outputHashAlgo = "sha256";
+  };
```
regnat, line 19: This derivation isn't input-addressed :upside_down_face: 
I guess the `__contentAddressed =true;` line is an overlook

```suggestion
  rootLegacy = mkDerivation {
    name = "simple-input-addressed";
    buildCommand = ''
      set -x
      echo "Building a legacy derivation"
      mkdir -p $out
      echo "Hello World" > $out/hello
    '';
  };
```

regnat:
> Fetching the patch of attr failed with a hash mismatch

This is probably due to 3e3eaa90dd1aaf4684697f27726ec72aec75206c (which has been partially reverted in 641c95070162595c71a1f775a2364cd8533e1c4b but without adding back `tarballs.nixos.org`as the default of `hashed-mirrors`), so nothing to do with this PR

edolstra (approved):

Ericson2314:
@regnat 

>     * There's a store method called `getDerivation`, but that's not implemented for `RemoteStore`

Sorry, I must be doing something silly, but I really don't see it. Just ones defined elsewhere.

>     * There's also `Store::narFromPath` that's also not implemented for `RemoteStore`, **but** the worker protocol has already the logic to handle it (because it's used by `SSHStore`), so we just need to upstream `SSHStore::narFromPath` into `RemoteStore` and it should be good. That's a lot of hops for such a workaround, but hopefully that should work

I agree it should be part of `RemoteStore`, but this isn't a problem in practice: in the only other extant case, using a Unix domain socket, `LocalFSStore` is also mixed in as a superclass, and it provides `narFromPath`.

So I think we fix this right away!

regnat:
> > * There's a store method called `getDerivation`, but that's not implemented for `RemoteStore`
> Sorry, I must be doing something silly, but I really don't see it. Just ones defined elsewhere.

My double-bad, it's `readDerivation` and not `getDerivation`.

> I agree it should be part of `RemoteStore`, but this isn't a problem in practice: in the only other extant case, using a Unix domain socket, `LocalFSStore` is also mixed in as a superclass, and it provides `narFromPath`.
> 
> So I think we fix this right away!

That's a fair point. Actually both `SSHStore` and `LocalFSStore` implement `readDerivation` (using the default implem that calls out to `getFSAccessor`), so I think we can directly resort to `readDerivation`
(and after a second look it's actually trivial to implement it for `RemoteStore` too without changing the protocol if we want)

regnat:
> So I think we fix this right away!

I've implemented it in 0b5f3397915775c11428394675261d77f182b53a if you want to cherry-pick it

Ericson2314:
@regnat What branch is that commit on? I'd love to cherry-pick it, but I'm having trouble fetching it!

Ericson2314:
@regnat see my previous comment?

regnat:
> @regnat see my previous comment?

Uh sorry, totally missed it.

Here you are: https://github.com/NixOS/nix/pull/4014

Ericson2314:
@edolstra it looks like this was accidentally closed because GitHub got excited about the word "fix". Can you repopen it?

edolstra:
Thanks, merged!

PR #3893, opened 2020-08-04
Get rid of json.hpp and use nlohmann/json everywhere

### LINKED ISSUE #3884: Get rid of json.hh and just use nlohmann/json
~~I suspect the answer is that we don't need both.~~ *edit* confirmed below. I would like to get rid of json.hh and just use nlohmann. In particular, nlohmnan has infrastructure for marshaling in both directions, while json.hh only has the `fromJSON` derivation, and I think we should get in the habit of always implementing both directions even if we just need one, to sanity check our schemas make sense.

CC @matthewbauer

Ericson2314:
Fixes #3884


src/nix-env/nix-env.cc
```diff
@@ -857,34 +859,34 @@ static VersionDiff compareVersionAgainstSet(
...
-                placeholder.write(nullptr);
+                metaObj[j] = nullptr;
             } else {
                 PathSet context;
-                printValueAsJSON(*globals.state, true, *v, placeholder, context);
+                metaObj[j] = printValueAsJSON(*globals.state, true, *v, context);
             }
         }
     }
+    std::cout << topObj;
```
edolstra, line 889: So there is a big difference between json.hpp and nlohmann/json: the former outputs JSON on the fly while the latter builds the entire JSON datastructure in memory. So the former is O(1) memory (more or less) while the latter is O(n).

edolstra: It also means that if you're piping the output into another tool, that tool won't receive any data until everything has been processed.

Ericson2314: Yes that was bugging me to. Some thing we do aren't too big so it doesn't really matter, but others like nar listings it might. https://github.com/nlohmann/json/discussions/2275 They are talking about streaming interfaces here, maybe I'll go pitch in.

Ericson2314:
This is not good yet, but nlohman json has been improving the sort of "direct printing" one would need to fix the perf issues here. I would rather keep this open as a reminder me to check back in on nlohman json and once that is done redo this.

Ericson2314:
It looks like this is no longer needed because https://github.com/NixOS/nix/pull/7313 did the the same same thing?

Ericson2314:
Ah yes, it is the same. Huzzah!

PR #3895, opened 2020-08-04
More templated STL support for the daemon protocol

Ericson2314:
It's a little boilerplate up front, but I think it makes things much easier, later.

I'm currently getting an invalid tag for a `std::optional`, which is stumping me as it's on a code path that I thought was operationally identical to the old one. Even weirder that the failure is Linux-specific.

This only intentionally changes the protocol for the `StorePathCAMap` serialization, which I think is OK as #3754 changes all that anyways.


src/libstore/worker-protocol.hh
```diff
@@ -66,14 +66,104 @@ typedef enum {
...
+            auto k = WorkerProto<K>::read(store, from);
+            auto v = WorkerProto<V>::read(store, from);
+            resMap.insert_or_assign(std::move(k), std::move(v));
+        }
+        return resMap;
+    }
+
+    static void write(const Store & store, Sink & out, const std::map<K, V> & resMap)
+    {
+        out << resMap.size();
```
regnat, line 126: If I'm not mistaken, this breaks an implicit assumption in the protocol that the first line of every non-scalar “packet” indicates the size (in number of lines) of the payload.

I'm not sure how important this is in practice − it's conceptually nicer as it means that the outer syntax of the protocol isn't context dependent and might prevent some ugly failures when things go bad − but I think it's worth mentioning (ftr the choice of using `2*map.size()` in the first iteration was guided by that)

src/libstore/worker-protocol.hh
```diff
@@ -66,14 +66,104 @@ typedef enum {
...
-
-void writeStorePaths(const Store & store, Sink & out, const StorePathSet & paths);
-
-StorePathCAMap readStorePathCAMap(const Store & store, Source & from);
-
-void writeStorePathCAMap(const Store & store, Sink & out, const StorePathCAMap & paths);
-
-void writeOutputPathMap(const Store & store, Sink & out, const OutputPathMap & paths);
+template<typename T>
+struct WorkerProto {
```
regnat, line 70: What's the motivation for using a templated struct here rather than a namespace like it was done in #3859 (with templated functions inside). Nothing fundamental, but I feel like `worker_proto::read<StorePath>()` is more natural than `WorkerProto<StorePath>::read()`

Ericson2314: We were worried about implicit conversation so we wanted to "recur" with an explicit `<T>` template instantiation.

However implicit conversations ended up not being a problem, and I've made it so the most important PRs no longer depend on this one so we can go either way.

src/libstore/worker-protocol.hh
Ericson2314: Ah OK. I didn't realize your intent there. I'm fine either way.

regnat: Either am I tbh. I just wanted to point that out in case that was a deliberate design choice by @edolstra

src/libstore/worker-protocol.hh
Ericson2314: @edolstra What do you think about switching from the workspace to `WorkerProto`? Happy to revert that part if you don't like it. As I wrote above I introduced it because i thought there was a bug. I'm 50-50 on whether to keep it or not.

edolstra: I think I would prefer keeping `worker_proto`. Seeing `WorkerProto` objects constructed everywhere is kind of confusing, since it requires you to read the `WorkerProto` definition to see that they're just a zero-cost wrapper around static methods...

Ericson2314: Changing it back, but for the record I believe `Class::method`, which is what I'm doing here, syntax never constructs an object of that class?

src/libstore/worker-protocol.hh
Ericson2314: I think this ended up being done in an earlier PR?

src/libstore/worker-protocol.hh
Ericson2314: Changed it back! Should be ready now.

Issue #3810, opened 2020-07-14
preferLocalBuild should not override max-jobs=0


grahamc:
**Is your feature request related to a problem? Please describe.**

The documented behavior of `max-jobs` defines that even with a value of 0, builds with `preferLocalBuild = true` still build locally. This is a fairly major policy decision that the daemon makes on behalf of the user, and is a decision the user cannot easily override.

Sometimes a machine should perform _absolutely_ no builds, in which case their options to fully disabling builds is to either delete all the build users (not sure nix-daemon would be happy with this) or set the `system` to a bogus value and assume/hope nobody will use it.

**Describe the solution you'd like**

I propose we extend the `features` feature of remote building to local building, too, and use preferLocalBuild as a listed feature.

Concretely, I want to:

1. extend the `nix.conf` to accept a `mandatory-features` option
2. treat derivations where `preferLocalBuild == true` as if they had the "local-preferred" feature
3. add `local-preferred` to the default value of `system-features`

**Backwards Compatibility**

* The current behavior of `max-jobs=0` can be achieved by setting `mandatory-features = local-preferred` and `max-jobs=auto`. 
* Other users may want to set `max-jobs=0` and assign the `local-preferred` feature to a specific builder which is nearby.

Ericson2314:
I would be fine just changing the behavior of `max-jobs = 0`, but the desugaring to `local-preferred` is neat and I like it :)

> This is a fairly major policy decision that the daemon makes on behalf of the user, and is a decision the user cannot easily override.

To be clear, is this a daemon thing in particular? What happens if one is not using the daemon, but has remove builders?

grahamc:
> To be clear, is this a daemon thing in particular? What happens if one is not using the daemon, but has remove builders?

Same behavior :).

Ericson2314:
OK whew. That would have made back-compat much more tricky.

domenkozar:
The reason the user might be using remote builder / binary cache is to avoid HDD space pollution and there's no way to disable that.

zupo:
I'm one of those Nix beginners that was just bitten hard by `preferLocalBuild` overriding max-jobs=0. My colleague and I started the day with a simple task: add https://nixbuild.net/ to our 500+ tiny VPSes hosting 70k websites so that when we run `nixos-rebuild switch`, their memory, CPU and disk don't get completely overused. 

[...]

Issue #2238, opened 2018-06-17
Improve error message when missing remote building feature forbids build


nh2:
I tried today to build a big package (chromium) remotely on a powerful machine with

```
nix-store --realise --builders 'ssh://root@1.2.3.4 x86_64-linux' /nix/store/i0kwyxpihg1gcp9jg4qwp7qcrpagj818-chromium-67.0.3396.87.drv /nix/store/bmigs53iryqpqjsy5w4qjfndlh6hxbms-chromium-67.0.3396.87.drv
```

but it would build only locally. Trying to force the remote build with `-j0`:

```
nix-store --realise --builders 'ssh://root@1.2.3.4 x86_64-linux' -j0 /nix/store/i0kwyxpihg1gcp9jg4qwp7qcrpagj818-chromium-67.0.3396.87.drv /nix/store/bmigs53iryqpqjsy5w4qjfndlh6hxbms-chromium-67.0.3396.87.drv
```

got me:

`unable to start any build; either increase '--max-jobs' or enable remote builds`

But I have remote build enabled -- `--builders` is given right there!

On IRC I learned that it's because chromium has `requiredSystemFeatures = [ "big-parallel" ];`, and I need to declare that my remote machine is `big-parallel`.

Solution:

```
nix-store --realise --builders 'ssh://root@1.2.3.4 x86_64-linux - - - big-parallel' -j0 /nix/store/i0kwyxpihg1gcp9jg4qwp7qcrpagj818-chromium-67.0.3396.87.drv /nix/store/bmigs53iryqpqjsy5w4qjfndlh6hxbms-chromium-67.0.3396.87.drv
```

---

But ideally we would improve the error message, perhaps

* not suggesting `or enable remote builds` when it's enabled
* printing a helpful warning/info message when when a builder with fitting architecture isn't selected just because it doesn't have a special flag like `big-parallel`

shonfeder-da:
I would vote for the second option. 

An error along these lines would have been the most helpful for me:

```
considering building on remote machine 'ssh://vm'
hook reply is 'decline'
this build requires system features [foo,...]
the remote machine is only declared as supporting [...]
if the remote machine supports it, add foo to the requiredSystemsFeatures of the `builders` configuration
```

domenkozar:
Similarly for local builds: https://github.com/NixOS/nix/issues/2238

colemickens:
[...]

PR #3897, opened 2020-08-05
error messages for issue 2238

bburdette:
Added a logError to build-remote to show the failure together with system info; added a check to the existing error site and reworded the error there.  Added a link to docs.

Here's the tail end of the result from 
```
sudo nix-store -j0 --realise --option substitute false --builders 'ssh://root@10.0.0.237 x86_64-linux - - - foo foo' /nix/store/crv8z01l69c3y0scq4cvdgyxrk2ldd1f-rustc-1.45.0.drv 
```

Selection_052

### DIFF
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -172,7 +172,7 @@ static int _main(int argc, char * * argv)
                     else
                     {
                         // build the hint template.
-                        string hintstring =  "required (system, features): (%s, %s)";
+                        string hintstring =  "derivation: %s\nrequired (system, features): (%s, %s)";
                         hintstring += "\n%s available machines:";
                         hintstring += "\n(systems, maxjobs, supportedFeatures, mandatoryFeatures)";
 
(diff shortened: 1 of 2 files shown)


bburdette:
#2238

edolstra:
Thanks!

B4dM4n:
This PR adds the following error to all `nix build`s for me, but the builds still succeed:
```shell
% nix build --impure --expr 'with import <nixpkgs> { }; runCommand "test" { } "touch $out"'
error: --- Remote build --- build-remote
Failed to find a machine for remote build!

required (system, features): (x86_64-linux, )
2 available machines:
(systems, maxjobs, supportedFeatures, mandatoryFeatures)
(x86_64-darwin, 2, , )
(x86_64-darwin, 2, , )
% realpath result
/nix/store/ml206rqibmzwffbbqssq879fj8jpivkb-test
```

bburdette:
> This PR adds the following error to all `nix build`s for me, but the builds still succeed:

Yeah, its only in retrospect that we can say it was a critical error - when neither remote nor local build is possible.  Maybe this is better as an warning, or even just an Info level message.

nh2:
> This PR adds the following error to all `nix build`s for me, but the builds still succeed:

You can use `--max-jobs 0` to force building on remotes only, see https://github.com/NixOS/nix/issues/1572#issuecomment-605385149 (with the exception of #3810).

nh2:
I made a follow-up PR to add more details in #3927.

PR #3899, opened 2020-08-05
Make narHash in ValidPathInfo not optional

meditans:


Ericson2314:
We realized we need to remove `Hash::operator bool ()`, and then we will catch a bunch more stuff.

src/libstore/path-info.hh
```diff
@@ -100,8 +100,8 @@ struct ValidPathInfo
 
     ValidPathInfo(const ValidPathInfo & other) = default;
 
-    ValidPathInfo(StorePath && path) : path(std::move(path)) { };
-    ValidPathInfo(const StorePath & path) : path(path) { };
+    ValidPathInfo(StorePath && path) : path(std::move(path)), narHash(Hash::dummy) { };
+    ValidPathInfo(const StorePath & path) : path(path), narHash(Hash::dummy) { };
```
edolstra, line 104: Ideally we'd require a `narHash` at construction time. Do you know if there are any places where we construct a `ValidPathInfo` and we don't know the `narHash`?

Ericson2314: Agreed. We should make this an argument, and at the very least callers would opt into using `Hash::dummy` and then initialize later, though better yet `Hash::dummy` can go away altogether.

Ericson2314:
@edolstra OK the `Hash::dummy` usage is pretty minimal now.

src/libstore/legacy-ssh-store.cc
```diff
@@ -93,15 +93,20 @@ struct LegacySSHStore : public Store
         try {
             auto conn(connections->get());
 
+            /* No longer support missing NAR hash */
+            assert(GET_PROTOCOL_MINOR(conn->remoteVersion) >= 4);
```
Ericson2314, line 97: One of the more interesting changes

src/libstore/legacy-ssh-store.cc
```diff
@@ -111,12 +116,14 @@ struct LegacySSHStore : public Store
...
             info->narSize = readLongLong(conn->from);
 
-            if (GET_PROTOCOL_MINOR(conn->remoteVersion) >= 4) {
+            {
                 auto s = readString(conn->from);
-                info->narHash = s.empty() ? std::optional<Hash>{} : Hash::parseAnyPrefixed(s);
-                info->ca = parseContentAddressOpt(readString(conn->from));
-                info->sigs = readStrings<StringSet>(conn->from);
+                if (s == "")
+                    throw Error("NAR hash is now mandatory");
```
Ericson2314, line 122: Another notable change

src/libstore/binary-cache-store.cc
```diff
@@ -385,7 +385,10 @@ StorePath BinaryCacheStore::addToStore(const string & name, const Path & srcPath
         h = hashString(hashAlgo, s);
     }
 
-    ValidPathInfo info(makeFixedOutputPath(method, *h, name));
+    ValidPathInfo info {
+        makeFixedOutputPath(method, *h, name),
+        Hash::dummy, // Will be fixed in addToStore, which recomputes nar hash
```
Ericson2314, line 390: This is rather tricky. Because the other one calculates the Nar hash, I think we should eventually make an `addToStoreFromDump` out of it, and then this tricky bit can go away.

src/libstore/binary-cache-store.cc
```diff
@@ -396,7 +399,10 @@ StorePath BinaryCacheStore::addToStore(const string & name, const Path & srcPath
 StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s,
     const StorePathSet & references, RepairFlag repair)
 {
-    ValidPathInfo info(computeStorePathForText(name, s, references));
+    ValidPathInfo info {
+        computeStorePathForText(name, s, references),
+        Hash::dummy, // Will be fixed in addToStore, which recomputes nar hash
```
Ericson2314, line 404: To get rid of this like we do above, we would need `addToStoreFromDump` to also support arbitrary reference, but I think that's a good idea anyways.

PR #3901, opened 2020-08-05
Add a flag to start the REPL on evaluation errors

edolstra:
This allows interactively inspecting the state of the evaluator at the point of failure.

Example:

```
$ nix eval path:///home/eelco/Dev/nix/flake2#modules.hello-closure._final --start-repl-on-eval-errors
error: --- TypeError -------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix
at: (20:53) in file: /nix/store/4264z41dxfdiqr95svmpnxxxwhfplhy0-source/flake.nix

    19|
    20|           _final = builtins.foldl' (xs: mod: xs // (mod._module.config { config = _final; })) _defaults _allModules;
      |                                                     ^
    21|         };

attempt to call something which is not a function but a set

Starting REPL to allow you to inspect the current state of the evaluator.

The following extra variables are in scope: arg, fun

nix-repl> fun
error: --- EvalError -------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix
at: (150:28) in file: /nix/store/4264z41dxfdiqr95svmpnxxxwhfplhy0-source/flake.nix

   149|
   150|           tarballClosure = (module {
   151|             extends = [ self.modules.derivation ];

attribute 'derivation' missing

nix-repl> :t fun
a set

nix-repl> builtins.attrNames fun
[ "tarballClosure" ]

nix-repl>
```

TODO: 
* Add the lexical environment at the failure site to the scope of the repl.
* Add many more places where the repl can be invoked.
* Add a breakpoint builtin?

### DIFF
--- a/src/nix/command.cc
+++ b/src/nix/command.cc
@@ -31,6 +31,30 @@ void StoreCommand::run()
     run(getStore());
 }
 
+EvalCommand::EvalCommand()
+{
+    addFlag({
+        .longName = "start-repl-on-eval-errors",
+        .description = "start an interactive environment if evaluation fails",
+        .handler = {&startReplOnEvalErrors, true},
+    });
+}
+
+extern std::function<void(const Error & error, const std::map<std::string, Value *> & env)> debuggerHook;
+
+ref<EvalState> EvalCommand::getEvalState()
+{
+    if (!evalState) {
+        evalState = std::make_shared<EvalState>(searchPath, getStore());
+        if (startReplOnEvalErrors)
+            debuggerHook = [evalState{ref<EvalState>(evalState)}](const Error & error, const std::map<std::string, Value *> & env) {
+                printError("%s\n\n" ANSI_BOLD "Starting REPL to allow you to inspect the current state of the evaluator.\n" ANSI_NORMAL, error.what());
+                runRepl(evalState, env);
+            };
+    }
+    return ref<EvalState>(evalState);
+}
+
 StorePathsCommand::StorePathsCommand(bool recursive)
     : recursive(recursive)
 {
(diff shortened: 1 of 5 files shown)


lf-:
> It's a bit long. Maybe something like?
> 
> ```
>          .longName = "repl-on-error",
> ```

--debugger?

domenkozar:
This would be the easiest fix for https://github.com/NixOS/nix/pull/3901

src/nix/command.cc
```diff
@@ -31,6 +31,30 @@ void StoreCommand::run()
     run(getStore());
 }
 
+EvalCommand::EvalCommand()
+{
+    addFlag({
+        .longName = "start-repl-on-eval-errors",
```
domenkozar, line 37:
```suggestion
        .longName = "debugger",
```

This is going to be used quite often so having a short long name is nice :)

domenkozar:
It would be useful to show for such errors at the end `To interacitvely inspect the scope pass --debugger`.

Something like `TypeErrorDebugger` that also takes fun & args.

lf-:
What's the status on this PR? I'd really like to be able to use this feature.

domenkozar:
cc @bburdette

bburdette:
@lf- I have been [hacking](https://www.zknotes.com/page/2021-06-10%20helpful%20repl%20error) on this problem a bit.  i have a branch that builds and kicks you into the repl on most eval exceptions, and doesn't appear to significantly impact stack size.

I'm still grappling with how best to get an appropriate set of symbols in the repl though.  Usually its easy enough to access an Env at exception time, but Envs contain only values, not symbols, and those values are stored in segfault-terminated arrays.  Envs are mainly for looking up symbol values where the location of the value within the env is known already, not introspecting to discover what symbols are available.  

My current thinking is maybe adding a value count to the Env would be inexpensive and allow iterating over the values to find Attr sets to add to the repl namespace.

jakubgs:
This is really cool and could be extremely useful.

PR #3909, opened 2020-08-06
Add hashed-mirrors back

matthewbauer:
Some users have their own hashed-mirrors setup, that is used to mirror
things in addition to what’s available on tarballs.nixos.org. Although
this should be feasible migrate to a Binary Cache, it’s not always
easy, since you have to remember what "name" each of the tarballs has.
Continuing to support hashed-mirrors is cheap, so it’s best to leave
support in Nix. Note that NIX_HASHED_MIRRORS is also supported in
Nixpkgs through fetchurl.nix.

This excludes tarballs.nixos.org from the default, as in
\#3689. All of these are available on cache.nixos.org.

### DIFF
--- a/doc/manual/command-ref/conf-file.xml
+++ b/doc/manual/command-ref/conf-file.xml
@@ -370,6 +370,33 @@ false</literal>.</para>
 
   </varlistentry>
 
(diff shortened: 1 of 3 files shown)


Ericson2314:
Can we make it a store implementation though?

Issue #3904, opened 2020-08-06
Feature: add `:doc` command to nix repl


lf-:
**Is your feature request related to a problem? Please describe.**

I am always having to look up function documentation in the nixpkgs repository with either `ripgrep` or the tool I wrote because of this problem, [`nix-doc`](https://github.com/lf-/nix-doc), but neither is perfect: `ripgrep` finds a lot of overridden copies of the function, and you still often have to jump between several files to decipher what's required to call a function.

By comparison, `nix-doc` uses a heuristic of looking in the AST for function definitions with comments above them to eliminate overrides, but this heuristic misses functions that are defined in their own file such as `fetchFromGitHub`, which is a problem I don't think I can practically solve in a tool that doesn't intend to understand the Nix language.

The Nix repl currently *has* the information to provide a better documentation experience as it knows where the *actual* function is defined through `builtins.unsafeGetAttrPos`, but it is not immediately obvious how it can nicely be extended to call some external tool to parse the nix file and find the comments above the function and output them.

**Describe the solution you'd like**

I want to have the following session in `nix repl` (also imagine it has colour):

```
nix-repl> nixpkgs = import <nixpkgs> {}
nix-repl> :doc nixpkgs.lib.composeExtensions
    Compose two extending functions of the type expected by 'extends'
    into one where changes made in the first are available in the
    'super' of the second
composeExtensions =  # /nix/store/...-nixpkgs/nixpkgs/lib/fixed-points.nix:74
    f: g: self: super: ...
```

**Describe alternatives you've considered**

It is /maybe/ possible to do this in nixpkgs, but you'd need a nix parser in nixpkgs OR make nix expose the comments in some attribute (discussed on IRC; plausible). Ergonomically it is probably superior to have this built into `nix repl` or to make `nix repl` support some kind of profile mechanism or other trick to add more `:command`s.

**Additional context**

cc @Infinisil we discussed this on IRC in #nix-lang a couple of weeks ago

Ericson2314:
Maybe let's start with `:locate` before we enshrine a inline documentation format in Nix itself?

lf-:
[...]

PR #3912, opened 2020-08-08
Add builtins.unsafeGetLambdaPos

lf-:
This is useful for a potential pure-Nix implementation of #3904 as `unsafeGetAttrPos` does not appear to be usable for finding lambdas that are defined by import, for example.

One thing I noticed while implementing this is that the way that Nix exposes Pos for things in the REPL as a set is surprising:

```
nix-repl> s = {a = 2;}                             

nix-repl> builtins.unsafeGetAttrPos "a" s          
{ column = 3; file = " {a = 2;}\n"; line = 1; }

nix-repl> f = a: "2"

nix-repl> builtins.unsafeGetLambdaPos f             
{ column = 2; file = " a: \"2\"\n"; line = 1; }
```

in that it doesn't report whether the `file` attribute is actually a file or arbitrary Nix source. It is unclear whether it's possible to rely on there being a leading space to determine whether a given `file` is actually Nix code.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -444,6 +444,17 @@ static void prim_genericClosure(EvalState & state, const Pos & pos, Value * * ar
         v.listElems()[n++] = i;
 }
 
+/* Return position information of the given lambda. */
+void prim_unsafeGetLambdaPos(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    /* ensure the argument is a function */
+    state.forceValue(*args[0], pos);
+    if (args[0]->type != tLambda) {
+        throwTypeError(pos, "value is %1% while a lambda was expected", *args[0]);
+    }
+
+    state.mkPos(v, &args[0]->lambda.fun->pos);
+}
 
 static void prim_abort(EvalState & state, const Pos & pos, Value * * args, Value & v)
 {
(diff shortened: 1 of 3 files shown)


Ericson2314:
I would rather we had `:doc` without this. I'm not eager to have more `builtin.unsafe*`, though pure eval mode does help mitigate.

matthewbauer:
> I would rather we had `:doc` without this. I'm not eager to have more `builtin.unsafe*`, though pure eval mode does help mitigate.

I don't think pure eval mode disallows any of the unsafe builtins.

---

Do we need to restrict this to lambdas? We should have position information for most Nix expressions.

lf-:
I believe that the reason it's called unsafe is because it basically guarantees that your function is impure as it then starts depending on the exact source making the derivation. Perhaps it should be named differently, but then it's inconsistent. I want this functionality for debugging mostly, anyway, and I can see it being potentially useful outside of the REPL in tracing or other places. 

I do not believe it is a risk for restricted evaluation mode because you'd have to first evaluate the file that's not in the Nix store before you get any information on it.

lf-:
@matthewbauer I looked into it. If we *just* have the value, it is not possible to get the position of things other than lambdas. See `src/libexpr/value.hh` in `struct Value`: the only value types that appear to have any metadata to speak of in there are lambdas.

It might be possible to take advantage of lazy evaluation in some way to achieve this, since perhaps we could do something prior to the forceValue call, but I don't know enough about Nix internals to implement this idea.

I have also made it so `:doc` can be implemented directly as a Nix plugin with #3934.

If there isn't any other proposed use of `unsafeGetLambdaPos` besides hypothetically building an inferior version of the now-existing `nix-doc` nix plugin that just got into `nixpkgs` (that uses essentially this code but natively from C++ and does the docs stuff in-process), I think we should abandon this PR.

PR #3915, opened 2020-08-10
WIP: Remote building for untrusted clients -- contains #3446

Ericson2314:


Ericson2314:
Ah, I'm afraid this approach is dead due to `drv.inputSrcs` being overridden in `build-remote.cc` :(. When I send over the normalized original derivation, we have the right output paths but wrong inputs, as the hashed input drvs cannot contribute anything. If I extend that derivation with the closure of the original inputs, we have the right inputs, bu wrong output paths.

PR #3917, opened 2020-08-11
Simplify code as output env vars are unconditional

Ericson2314:
Since the jsonObject unique ptr is reset to flush the string to make`__json`, all these `!jsonObject` conditions will always be true.


edolstra:
So that's a bug then. The intent was not to add environment variables for the outputs in structured attrs mode. But maybe it's too late to fix this since it would change derivations.

Ericson2314:
Yeah I was unsure about that. Doing some git-blame it looks like this was the case from the beginning of structured attrs. Pity there were no experimental features then!

PR #3918, opened 2020-08-11
Allow trustless building of CA derivations --- contains #3922

Ericson2314:
Include a long comment explaining the policy. Perhaps this can be moved
to the manual at some point in the future.

Also bump the daemon protocol minor version, so clients can tell whether
`wopBuildDerivation` supports trustless CA derivation building. I hope
to take advantage of this in a follow-up PR to support trustless remote
building with the minimal sending of derivation closures.

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -688,7 +688,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         auto path = store->parseStorePath(readString(from));
         logger->startWork();
         logger->stopWork();
-        dumpPath(store->printStorePath(path), to);
+        dumpPath(store->toRealPath(store->printStorePath(path)), to);
         break;
     }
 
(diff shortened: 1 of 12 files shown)


src/libstore/daemon.cc
```diff
@@ -454,8 +454,46 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
...
+           Lastly, do note that when we parse fixed-output content-addressed
+           derivations, we throw out the precomputed output paths and just
+           store the hashes, so there aren't two competing sources of truth an
+           attacker could exploit. */
+        if (drv.type() == DerivationType::InputAddressed && !trusted)
+            throw Error("you are not privileged to build input-addressed derivations");
+
+        /* Make sure that the non-input-addressed derivations that got this far
+           are in fact content-addressed. */
+        assert(derivationIsCA(drv.type()));
```
Kha, line 495: Does this assertion not break on building *trusted* input-addressed derivations?

Ericson2314: Hahaha yes it does. Thanks.

Ericson2314: I've begun https://github.com/NixOS/nix/pull/3922 so that we can properly test this PR.

Ericson2314: Fixed, and tests confirm now.

PR #3919, opened 2020-08-11
demote remote build message to Info

bburdette:
Since this message can be a normal part of the build, it might be better off as an Info level message.  

So now we get the same message, but Info level:
Selection_055

And with --quiet they are filtered out:
Selection_053


bburdette:
@B4dM4n

PR #3921, opened 2020-08-12
Trustless remote building for input-addressed drvs

### LINKED ISSUE #2789: distributed builds require a trusted remote user
Attempting to initiate a remote build results in the following: `error: you are not privileged to build derivations` (nix (Nix) 2.2.2)
 
Adding the remote user to the remote machine's `nix.conf` `trusted-users` allows the build to continue as normal. This seems to be intended behaviour, but I had a few problems with it:
1. The [nix manual](https://nixos.org/nix/manual/#chap-distributed-builds) and [wiki page](https://nixos.wiki/wiki/Distributed_build) do not seem to mention this requirement at all, so this was surprising. This should probably be mentioned in instructions and documented with a warning.
    - My assumption was that since untrusted/allowed local users can initiate builds through the daemon, doing so over ssh with the distributed build functionality would work in the same way without requiring a change to a less secure system configuration. Considering that it's recommended to store the ssh key for a remote build account without a passphrase, and that "trusted-users is essentially equivalent to giving that user root access", it seems reasonable to expect a warning here before setting up such a machine.
    - There is a small troubleshooting footnote on the wiki page that mentions `nix.trustedUsers`, but it wasn't clear whether it was referring to the local (like when needed for use with `--builders`) or the remote machine.
2. The above error message was also unclear to me, since the user does indeed have permission to build derivations under normal circumstances. Very few google results exist for the message, so I assumed I was dealing with some sort of misconfiguration or environment issue. Something more explicit like `remote builds require a trusted user` might be more obvious?
    - `nix copy --to ssh://builder /nix/store/whatever.drv && ssh builder nix build /nix/store/whatever.drv && nix copy --no-check-sigs --from ssh://builder /nix/store/whatever` works fine, so I was confused why remote builds with `--builder` and nix.conf `builders` weren't able to do the same.

I understood there would be drawbacks to setting it up this way (not trusting the remote user will prevent them from supplying binary substitutes for example, and with multiple builders could even result in duplicated work if the build scheduler isn't aware), but did not expect or find mention that it would be disallowed and prevented entirely. Is there a reason why this check exists? Attempting to send an unsigned build input to the remote machine would've already failed out the build before hitting this wopBuildDerivation error, so it's not immediately clear to me why this particular operation is gated to trusted users to begin with? Sorry if I'm misunderstanding what a "build derivation" operation actually is/does.

Ericson2314:
# Motivation

Fixes #2789.

Input-addressed outputs cannot be safely pushed to a remote builder without trusting the client. This makes trustless remote building without experimental features completely useless in practice. Making them work is widely requested.

The solution is simple: when not trusted, instead of using `buildDerivation` (which sends a mere `BasicDerivation` and requires trust, use `buildPaths` and send the entire `Derivation` and its closure).

# Context

## Known failure mode

We still cannot copy input-addressed paths to a remote builder safely. In that case, this still fails (see the should-fail test). The manual work around is to tell the remote builder to build that path first, so it already has it. This can be done with `nix-build --store <them> --eval-store <us>`.

That is manual and tedious, but in may real world cases can be avoided in practice. (e.g. when remote builders are always used, and the remote builders trust one another, this won't arise.)

Seems good enough to start with.

~~Depends on #8230~~ for testing with untrusted remote store.

~~Depends on #7515~~ for knowing whether the remote store trusts us 

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
 - [ ] documentation in the internal API docs

# Priorities


src/libstore/store-api.hh
```diff
@@ -158,7 +158,9 @@ public:
 
     const Setting<int> pathInfoCacheSize{this, 65536, "path-info-cache-size", "size of the in-memory store path information cache"};
 
-    const Setting<bool> isTrusted{this, false, "trusted", "whether paths from this store can be used as substitutes even when they lack trusted signatures"};
+    const Setting<bool> isTrusted{this, false, "trusted", "whether paths from this store can be used as substitutes even when they lack trusted signatures. The dual of \"trusting\""};
+
+    Setting<bool> isTrusting{this, true, "trusting", "whether (we think) paths an be be added to this store even when they lack trusted signatures. The dual of \"trusted\""};
```
matthewbauer, line 163:
```suggestion
    Setting<bool> isTrusting{this, true, "trusting", "whether (we think) paths can be added to this store even when they lack trusted signatures. The dual of \"trusted\""};
```

src/libstore/store-api.hh
```diff
@@ -158,7 +158,9 @@ public:
 
     const Setting<int> pathInfoCacheSize{this, 65536, "path-info-cache-size", "size of the in-memory store path information cache"};
 
-    const Setting<bool> isTrusted{this, false, "trusted", "whether paths from this store can be used as substitutes even when they lack trusted signatures"};
+    const Setting<bool> isTrusted{this, false, "trusted", "whether paths from this store can be used as substitutes even when they lack trusted signatures. The dual of \"trusting\""};
```
matthewbauer, line 161: I think dual is a little too mathematical for user-facing text. How about just `see also`?

src/libstore/store-api.hh
matthewbauer: I think we need some example of when you want to set this true or false. Could we just ask the builder whether they trust us, eliminating the need for a setting?

src/libstore/store-api.cc
```diff
@@ -834,7 +834,21 @@ std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStor
...
+                        auto outputMap = srcStore->queryDerivationOutputMap(drvPath);
+                        auto p = std::find_if(outputMap.begin(), outputMap.end(), [&](auto & i) {
+                            return i.second == storePath;
+                        });
+                        dstStore->buildPaths({{
+                            *info->deriver,
+                            p != outputMap.end() ? StringSet { p->first } : StringSet {},
+                        }});
+                    } else {
+                        dstStore->ensurePath(storePath);
```
matthewbauer, line 850: What happens if storePath can't be ensured by dstStore? Do we have some way to ask if it can build it, and if not, say they need to be trusted to build this path?

src/build-remote/build-remote.cc
```diff
@@ -282,10 +282,21 @@ static int _main(int argc, char * * argv)
...
 
-        auto result = sshStore->buildDerivation(*drvPath, drv);
+        BasicDerivation drv = store->readDerivation(*drvPath);
+
+        if (sshStore->isTrusting || derivationIsCA(drv.type())) {
+            drv.inputSrcs = store->parseStorePathSet(inputs);
+            result = sshStore->buildDerivation(*drvPath, drv);
+        } else {
+            try {
+                sshStore->buildPaths({{*drvPath}});
```
matthewbauer, line 294: Can we ensure drvPath exists on the remote ssh store?

src/libstore/store-api.hh
Ericson2314: Will do. FWIW, It isn't quite the true dual anyways, since it's the client's guess rather than server's response.

src/libstore/store-api.hh
```diff
@@ -158,12 +158,18 @@ public:
 
     const Setting<int> pathInfoCacheSize{this, 65536, "path-info-cache-size", "size of the in-memory store path information cache"};
 
-    const Setting<bool> isTrusted{this, false, "trusted", "whether paths from this store can be used as substitutes even when they lack trusted signatures"};
+    const Setting<bool> isTrusted{this, false, "trusted", "whether paths from this store can be used as substitutes even when they lack trusted signatures. Compare \"trusting\""};
+
+    Setting<bool> isTrusting{this, true, "trusting", "whether (we think) paths can be added to this store even when they lack trusted signatures. Compare \"trusted\""};
```
matthewbauer, line 163: Could we just ask the builder whether they trust us, eliminating the need for a setting?

Ericson2314: We could but this a good bit more work. Think of the complicated case where one sshes and then the `nix-daemon --stdio` talks to another deaemon. Need a new `Store` interface that all the stores implement. Also, we should really keep it signature based---never mind who "you" are, what is the public key you have signatures for (whether you downloaded or made yourself)? That is the "right way" to do it, and I wouldn't want to do anything but the right away in the protocols where evolution is costly, but that's more work in addition.

My own reservation is we don't have anything like "experimental settings" yet, which would be ideal so we can switch to the stuff above at a later point.

matthewbauer: Maybe I'm misunderstanding this then. When you set this to true & you're not a trusted-user on that remote store, you get an error right? I just want to ask the daemon whether it will give us an error. It's a new op, but doesn't need an experimental flag or anything. It's the daemon that decides which mode you can or cannot use.

matthewbauer: You actually don't need a new op necessarily, but could wrap it in try catch statements.

matthewbauer: You're always going to prefer "true" if you can do it right? So maybe it should be opt-in, but just like `allow-trustless-remote-builds`.

src/build-remote/build-remote.cc
Ericson2314: Yes; done.

src/libstore/store-api.cc
Ericson2314: This is the final fallback if we don't know how to build it. If this fails, we are out of options.

I suppose we could suggest trying without `trustless=true`, but since that's opt-in I wouldn't really worry.

src/libstore/store-api.cc
```diff
@@ -822,7 +822,23 @@ std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStor
...
+                        auto outputMap = srcStore->queryDerivationOutputMap(drvPath);
+                        auto p = std::find_if(outputMap.begin(), outputMap.end(), [&](auto & i) {
+                            return i.second == storePath;
+                        });
+                        // drv file is always CA
+                        copyStorePath(srcStore, dstStore, drvPath, repair, checkSigs);
+                        dstStore->buildPaths({{
+                            drvPath,
+                            p != outputMap.end() ? StringSet { p->first } : StringSet {},
+                        }});
```
edolstra, line 838: It seems very unexpected to me that `copyPaths` would *build* something on the destination store...

What's the use case for this?

src/libstore/store-api.hh
```diff
@@ -157,7 +157,9 @@ public:
...
```
edolstra, line 162: We should be careful about adding another signature-related flag since it's getting very hard to understand (we already have `trusted` and `--no-check-sigs`, and it's already not really clear how they interact...).

Why do we actually need "trusting"? Either the remote store accepts the paths we send to it or it doesn't...

src/nix-daemon/nix-daemon.cc
```diff
@@ -278,14 +279,26 @@ static int _main(int argc, char * * argv)
                 printVersion("nix-daemon");
             else if (*arg == "--stdio")
                 stdio = true;
-            else return false;
+            else if (*arg == "--trust") {
```
edolstra, line 282: What does this flag do? What are we trusting?

src/libstore/store-api.cc
Ericson2314: admittedly this isn't the ideal place to put it, but the idea is that when we have an input-addressed path and the remote builder doesn't trust us, we ask the remote builder to build the path itself.

src/libstore/store-api.hh
Ericson2314: > We should be careful about adding another signature-related flag since it's getting very hard to understand (we already have `trusted` and `--no-check-sigs`

Agreed, this isn't a new axis, and I am not sure about the differences between `trusted` and `--no-check-sigs` either.

> Either the remote store accepts the paths we send to it or it doesn't...

 Agreed, this isn't a new axis but rather a way to guess whether the remote store would accept non-CA paths or wouldn't. As @matthewbauer says above it would be slicker to not have the user guess but instead try and fallback, but I wasn't sure where to put the state from trying. Maybe I'm overthinking it and it can just be local vars in `copyPaths` and `build-remote`.

src/nix-daemon/nix-daemon.cc
Ericson2314: `stdin`. When we have a Unix domain socket we can do `getPeerInfo`, but we can't do that with stdin so currently we just always trust `stdin` (that is, if we aren't just proxying stdin another daemon, as is the common case with `ssh-ng`).

This provides a way to *not* trust stdin, which is useful for testing. I could make these flag `--trust-stdin` /  `--no-trust-stdin` for clarity.

src/libstore/store-api.cc
Ericson2314: @edolstra I was thinking it would be very nifty if the instead of client trying to guess what the remote builder needs and copying it over, the client could open a fresh connection *acting as the daemon* with a new sort of `RestrictedStore` so the server (acting as the client) could just directly request what it needs, as managed with regular substitution goals, etc.

It's very fancy, but this does mean we can avoid any modifications to `copyPaths`.

Ericson2314: This is now fixed in the last commit. Only the build hook will fall back on building things which cannot be trusted. All the other uses of `copyPaths` will never build anything, as today.

osvein:
What's the status on this? Will it make it into nix 2.4?

Ericson2314:
That's for @edolstra to say :). I'm happy to continue fixing merge conflicts.

davean:
Things seem pretty broken without this @edolstra any hope? Whats holding this up now?

Ericson2314:
mac failure seems unreleated?

zimbatm:
This would be useful for the nix-community builders. So we can give semi-trusted members build capacity without giving them essentially root access.

winterqt:
I'd like to see this merged to investigate installing Nix on some of the [GCC compile farm](https://cfarm.tetaneutral.net/) machines, with remote building capabilities -- it would, like the nix-community builders, be a net good for contributors.

chasecaleb:
I would also love to see this, because I'm trying to get some Nix infrastructure setup for CI/CD at work. Remote builds would be a great solution if this was implemented.

fricklerhandwerk:
@Ericson2314 please add a PR description so we can see at a glance what it does and why.

Ericson2314:
https://github.com/NixOS/nix/pull/7515 relates to this.

virusdave:
> I would also love to see this, because I'm trying to get some Nix infrastructure setup for CI/CD at work. Remote builds would be a great solution if this was implemented.

Piling in to the "Please let's get this landed; it's very helpful" chorus here.  Is there anything specifically blocking this change currently?

Ericson2314:
@virusdave I would say getting https://github.com/NixOS/nix/pull/7515 finished first would help.

src/libstore/store-api.hh
Ericson2314: Using https://github.com/NixOS/nix/pull/7515 we have gotten rid of it!

src/libstore/store-api.hh
Ericson2314: Using https://github.com/NixOS/nix/pull/7515 we have gotten rid of it!

src/nix-daemon/nix-daemon.cc
Ericson2314: I have generalized the flag so it also works in the Unix domain socket case; it forces the trust one way or the other instead of checking the user via`authPeer`.

Ericson2314:
I am happy to say that with https://github.com/NixOS/nix/pull/7515 done (and a few small last-minute things cleaned up), I think this PR is finally ready! Hope to get it merged next week!

Ericson2314:
Making this draft not because it is not ready, but because I pulled out #8230 as preparatory PR we should merge first for better reviewing.

Ericson2314:
OK, this is now quite small and easy to review!

src/build-remote/build-remote.cc
```diff
@@ -286,23 +288,58 @@ static int main_build_remote(int argc, char * * argv)
...
+        // ### Content-addressing part
+        //
+        // ```
+        // ...trustCond... || drv.type().isCA()
+        // ```
+        //
+        // See the very large comment in `case wopBuildDerivation:` in
+        // `src/libstore/daemon.cc` that explains the trust model here.
+        //
+        // This condition mirrors that: that code enforces the "rules";
```
fricklerhandwerk, line 315:
```suggestion
        // This condition mirrors that: that code enforces the "rules" outlined there;
```

src/build-remote/build-remote.cc
```diff
@@ -286,23 +288,58 @@ static int main_build_remote(int argc, char * * argv)
...
+            if (!drv.inputDrvs.empty())
+                drv.inputSrcs = store->parseStorePathSet(inputs);
+            optResult = sshStore->buildDerivation(*drvPath, (const BasicDerivation &) drv);
+            auto & result = *optResult;
+            if (!result.success())
+                throw Error("build of '%s' on '%s' failed: %s", store->printStorePath(*drvPath), storeUri, result.errorMsg);
+        } else {
+            copyPaths(*store, *sshStore, StorePathSet {*drvPath}, NoRepair, NoCheckSigs, substitute);
+            auto res = sshStore->buildPathsWithResults({ DerivedPath::Built { *drvPath, OutputsSpec::All {} } });
+            // One path to build should mean one result back
```
fricklerhandwerk, line 336:
```suggestion
            // One path to build should produce exactly one build result
```

src/build-remote/build-remote.cc
```diff
@@ -286,23 +288,58 @@ static int main_build_remote(int argc, char * * argv)
...
+            // 2. Changing the `inputSrcs` set changes the associated
+            //    output ids, which break CA derivations
+            if (!drv.inputDrvs.empty())
+                drv.inputSrcs = store->parseStorePathSet(inputs);
+            optResult = sshStore->buildDerivation(*drvPath, (const BasicDerivation &) drv);
+            auto & result = *optResult;
+            if (!result.success())
+                throw Error("build of '%s' on '%s' failed: %s", store->printStorePath(*drvPath), storeUri, result.errorMsg);
+        } else {
+            copyPaths(*store, *sshStore, StorePathSet {*drvPath}, NoRepair, NoCheckSigs, substitute);
```
fricklerhandwerk, line 334: Closure of direct inputs was copied already as part of setting up the build, but here we need all transitive inputs of dependent derivations.
```suggestion
            copyClosure(*store, *sshStore, StorePathSet {*drvPath}, NoRepair, NoCheckSigs, substitute);
```

Ericson2314, line 334:
```suggestion
            copyClosure(*store, *sshStore, StorePathSet {*drvPath}, NoRepair, NoCheckSigs, substitute);
```

src/build-remote/build-remote.cc
```diff
@@ -286,23 +288,58 @@ static int main_build_remote(int argc, char * * argv)
...
+        // ```
+        // ...trustCond... || drv.type().isCA()
+        // ```
+        //
+        // See the very large comment in `case wopBuildDerivation:` in
+        // `src/libstore/daemon.cc` that explains the trust model here.
+        //
+        // This condition mirrors that: that code enforces the "rules";
+        // we do the best we can given those "rules".
+        if (std::optional trust = sshStore->isTrustedClient(); (!trust || *trust) || drv.type().isCA()) {
```
fricklerhandwerk, line 317: Put the explanatory comments around the helper variables.
```suggestion
        std::optional trusted = sshStore->isTrustedClient();
        // for backward compatibility (use existing comments here)
        bool trustedOrLegacy = !trusted || *trusted;
        if (trustedOrLegacy || drv.type().isCA())  {
```

fricklerhandwerk (approved):
Reviewed in Nix team meeting.

arcnmx:
:tada: :tada: :tada:

PR #3922, opened 2020-08-12
Test `RemoteStore::buildDerivation`

Ericson2314:
Fix `wopNarFromPath` which needed a `toRealPath`.

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -688,7 +688,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         auto path = store->parseStorePath(readString(from));
         logger->startWork();
         logger->stopWork();
-        dumpPath(store->toRealPath(path)), to);
+        dumpPath(store->toRealPath(path), to);
         break;
     }
 
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -688,7 +688,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         auto path = store->parseStorePath(readString(from));
         logger->startWork();
         logger->stopWork();
-        dumpPath(store->printStorePath(path), to);
+        dumpPath(store->toRealPath(store->printStorePath(path)), to);
         break;
     }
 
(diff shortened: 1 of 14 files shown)


tests/build-remote.sh
```diff
@@ -5,26 +5,29 @@ clearStore
 if ! canUseSandbox; then exit; fi
 if ! [[ $busybox =~ busybox ]]; then exit; fi
 
-chmod -R u+w $TEST_ROOT/machine0 || true
-chmod -R u+w $TEST_ROOT/machine1 || true
-chmod -R u+w $TEST_ROOT/machine2 || true
-rm -rf $TEST_ROOT/machine0 $TEST_ROOT/machine1 $TEST_ROOT/machine2
-rm -f $TEST_ROOT/result
```
Ericson2314, line 12: We now run `init.sh` before every test so we don't need this.

tests/build-remote.sh
```diff
@@ -5,26 +5,29 @@ clearStore
...
-
 unset NIX_STORE_DIR
 unset NIX_STATE_DIR
 
+function join_by { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d}"; }
+
+builders=(
+  "ssh://localhost?remote-store=$TEST_ROOT/machine1"
+  "$TEST_ROOT/machine2 - - 1 1 foo"
+  "ssh-ng://localhost?remote-store=$TEST_ROOT/machine3?systemFeatures=baz - - 1 1 baz"
```
Ericson2314, line 16: `?systemFeatures=baz` won't work, but it's failing before it even gets that far.

src/libstore/daemon.cc
```diff
@@ -688,7 +688,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         auto path = store->parseStorePath(readString(from));
         logger->startWork();
         logger->stopWork();
-        dumpPath(store->printStorePath(path), to);
+        dumpPath(store->toRealPath(store->printStorePath(path)), to);
```
edolstra, line 691:
```suggestion
        dumpPath(store->toRealPath(pat)), to);
```

Ericson2314:
Oh good, still works.

PR #3924, opened 2020-08-12
Make `system-features` a store setting

Ericson2314:
This seems more correct. It also means one can specify the features a store should support with `--store` and `remote-store=...`, which is useful. I use this to clean up the build remotes test.

### DIFF
--- a/src/libstore/store-api.hh
+++ b/src/libstore/store-api.hh
@@ -164,6 +164,10 @@ public:
 
     Setting<bool> wantMassQuery{this, false, "want-mass-query", "whether this substituter can be queried efficiently for path validity"};
 
+    Setting<StringSet> systemFeatures{this, settings.systemFeatures,
+        "system-features",
+        "Optional features that the system this store builds on implements (like \"kvm\")."};
+
 protected:
 
     struct PathInfoCacheValue {
--- a/tests/build-remote.sh
+++ b/tests/build-remote.sh
@@ -1,31 +1,36 @@
 source common.sh
 
-clearStore
-
 if ! canUseSandbox; then exit; fi
 if ! [[ $busybox =~ busybox ]]; then exit; fi
-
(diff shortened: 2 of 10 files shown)


tests/build-remote.sh
```diff
@@ -1,31 +1,36 @@
...
+)
+
 # Note: ssh://localhost bypasses ssh, directly invoking nix-store as a
 # child process. This allows us to test LegacySSHStore::buildDerivation().
+# ssh-ng://... likewise allows us to test RemoteStore::buildDerivation().
 nix build -L -v -f build-hook.nix -o $TEST_ROOT/result --max-jobs 0 \
   --arg busybox $busybox \
   --store $TEST_ROOT/machine0 \
-  --builders "ssh://localhost?remote-store=$TEST_ROOT/machine1; $TEST_ROOT/machine2 - - 1 1 foo" \
-  --system-features foo
```
Ericson2314, line 23: No longer need this, because it's just on the chroot store we're using as a (not so remote) builder.

src/libstore/store-api.hh
```diff
@@ -164,6 +164,10 @@ public:
 
     Setting<bool> wantMassQuery{this, false, "want-mass-query", "whether this substituter can be queried efficiently for path validity"};
 
+    Setting<StringSet> systemFeatures{this, settings.systemFeatures,
```
Ericson2314, line 167: The old setting is still there for backwards compat. It just is used to default the store setting.

tests/build-remote.sh
Kloenk: Not sure how testing works in nix. But can we keep it to have the remote build test? Or would it not call the remote build code?

Ericson2314: Hmm? Now both builders are associated with their own unique feature, so it's more deterministic what would be used where. Do you mean you want to test that `--system-features foo` will still default the per-store stetting as intended?

PR #3927, opened 2020-08-13
Tell in detail why remote building fails on -v

nh2:
See #1572. Further improves upon #3897 (CC @bburdette).

Also further improve the final error message, giving a practical example of
how to set machine features because that's what most people need to succeed
in building large packages remotely.


src/build-remote/build-remote.cc
```diff
@@ -123,11 +130,22 @@ static int _main(int argc, char * * argv)
...
+                        printInfo(format("declined remote machine '%1%' for building '%2%' because it does not have the needed system type: '%3%") % m.storeUri % drvstr % neededSystem);
+                    } else if (!m.allSupported(requiredFeatures)) {
+                        std::ostringstream joinedFeatures;
+                        for (const auto & i : requiredFeatures)
+                            joinedFeatures << i << ' '; // includes trailing space
+                        printInfo(format("declined remote machine '%1%' for building '%2%' because it does not have all required features: [ %3%]") % m.storeUri % drvstr % joinedFeatures.str());
+                    } else if (!m.mandatoryMet(requiredFeatures)) {
+                        std::ostringstream joinedFeatures;
+                        for (const auto & i : requiredFeatures)
+                            joinedFeatures << i << ' '; // includes trailing space
```
Ericson2314, line 146: Unless I'm missing something, there is a `concatStringsSep` function in `utils.hh` you could use for this?

nh2: Done.

nh2: I also noticed `concatStringsSep` does O(n²) string append, and `ostringstream` is no better; fixing it in PR #3941.

src/build-remote/build-remote.cc
```diff
@@ -123,11 +130,18 @@ static int _main(int argc, char * * argv)
                 for (auto & m : machines) {
                     debug("considering building on remote machine '%s'", m.storeUri);
 
-                    if (m.enabled && std::find(m.systemTypes.begin(),
-                            m.systemTypes.end(),
-                            neededSystem) != m.systemTypes.end() &&
-                        m.allSupported(requiredFeatures) &&
-                        m.mandatoryMet(requiredFeatures)) {
+                    if (!m.enabled) {
+                        debug(format("declined remote machine '%1%' because it is disabled") % m.storeUri);
```
edolstra, line 134: The `format` function is deprecated, please do:

```suggestion
                        debug("declined remote machine '%s' because it is disabled", storeUri);
```

src/libstore/build.cc
```diff
@@ -4882,9 +4882,14 @@ void Worker::run(const Goals & _topGoals)
                             "or enable remote builds."
                             "\nhttps://nixos.org/nix/manual/#chap-distributed-builds");
                 else
-                   throw Error("unable to start any build; remote machines may not have "
-                            "all required system features."
-                            "\nhttps://nixos.org/nix/manual/#chap-distributed-builds");
+                   throw Error("unable to start any build; either increase '--max-jobs' "
+                            "to also allow building locally, or ensure that the configred"
```
edolstra, line 4886:
```suggestion
                            "to also allow building locally, or ensure that the configured"
```

bburdette:
Got a screenshot of output from this?

src/build-remote/build-remote.cc
```diff
@@ -123,11 +130,18 @@ static int _main(int argc, char * * argv)
...
-                    if (m.enabled && std::find(m.systemTypes.begin(),
-                            m.systemTypes.end(),
-                            neededSystem) != m.systemTypes.end() &&
-                        m.allSupported(requiredFeatures) &&
-                        m.mandatoryMet(requiredFeatures)) {
+                    if (!m.enabled) {
+                        debug("declined remote machine '%s' because it is disabled", storeUri);
+                    } else if (std::find(m.systemTypes.begin(), m.systemTypes.end(), neededSystem)
+                        == m.systemTypes.end()) {
+                        printInfo(format("declined remote machine '%1%' for building '%2%' because it does not have the needed system type: '%3%") % m.storeUri % drvstr % neededSystem);
```
edolstra, line 137:
```suggestion
                        printInfo("declined remote machine '%s' for building '%s' because it does not have the needed system type '%s", m.storeUri, drvstr, neededSystem);
```

`format` is deprecated.

nh2: > `format` is deprecated.

Ah OK, I somehow thought it was the other way around.

nh2:
I've squashed the commit suggestions.

nh2:
> Got a screenshot of output from this?

@bburdette I didn't make a screenshot but it's litearlly just another of the many lines in the `-v` output, like

```
...
declined remote machine 'static-haskell-nix-ci' for building '/nix/store/package-hash1234' because the derivation does not have all mandatory features to build on that machine: [ big-parallel ]
...
```

domenkozar:
@nh2 could you rebase and we'll merge it?

nh2:
@Ericson2314 Can you help me rebase it?

You deleted the message I enhanced in https://github.com/NixOS/nix/pull/3927/files#diff-19f999107b609d37cfb22c58e7f0bc1cf76edf1180e238dd6389e03cc279b604R4885 as part of #4114 here:

https://github.com/NixOS/nix/commit/3bab1c5bb0a56f850a7bc1bacc9f974b108cf601#diff-5784d1ad7c22601c75b86872534049bea5173377de5d61b7446452beabc16d3fL5324

Thanks!

nh2:
@Ericson2314 short ping :)

domenkozar:
:pray: this one is a huge UX improvement, it would be a shame to let it stale.

nh2:
@Ericson2314 Short ping 2 years later ^ :)

bitonic:
Just to chime in, I lost half an hour on this today. It'd be great to get this merged.

Ericson2314:
Ah! I am seeing and all its pings this now :(

Ericson2314:
As penance, I will just rebase this myself.

Ericson2314:
@nh2 I have rebased, and applied the PR description template.

Ericson2314:
As to the content of the new error messages, I hope @fricklerhandwerk can review them.

src/build-remote/build-remote.cc
```diff
@@ -137,14 +144,19 @@ static int main_build_remote(int argc, char * * argv)
...
-                    {
+                    if (!m.enabled) {
+                        debug("declined remote machine '%s' because it is disabled", storeUri);
+                    } else if (neededSystem != "builtin" &&
+                        std::find(m.systemTypes.begin(), m.systemTypes.end(), neededSystem)
+                        == m.systemTypes.end()) {
+                        debug("declined remote machine '%s' for building '%s' because it does not have the needed system type '%s", m.storeUri, drvstr, neededSystem);
+                    } else if (!m.allSupported(requiredFeatures)) {
+                        std::string joinedFeatures = concatStringsSep(" ", requiredFeatures); // includes leading space
+                        debug("declined remote machine '%s' for building '%s' because it does not have all required features: [ %s ]", m.storeUri, drvstr, joinedFeatures);
```
fricklerhandwerk, line 155: This should be explicit on what was expected and what is available.

src/build-remote/build-remote.cc
```diff
@@ -137,14 +144,19 @@ static int main_build_remote(int argc, char * * argv)
...
+                    } else if (!m.mandatoryMet(requiredFeatures)) {
+                        std::string joinedFeatures = concatStringsSep(" ", requiredFeatures); // includes leading space
+                        debug("declined remote machine '%s' for building '%s' because the derivation does not have all mandatory features to build on that machine: [ %s ]", m.storeUri, drvstr, joinedFeatures);
```
fricklerhandwerk, line 158: I don't understand what this means. I thought a machine may be lacking features to build a drv, how would it work the other way round?

src/libstore/build/worker.cc
```diff
@@ -283,9 +283,14 @@ void Worker::run(const Goals & _topGoals)
...
                 else
-                   throw Error("unable to start any build; remote machines may not have "
-                            "all required system features."
-                            "\nhttps://nixos.org/manual/nix/stable/advanced-topics/distributed-builds.html");
+                   throw Error("unable to start any build; either increase '--max-jobs' "
+                            "to also allow building locally, or ensure that the configured"
+                            "remote builder machines have all required system features."
+                            "For example, to enable the 'big-parallel' feature, use:"
+                            "\n    --builders 'yourbuilderhostname - - - - big-parallel'"
+                            "\nhttps://nixos.org/nix/manual/stable/advanced-topics/distributed-builds"
```
fricklerhandwerk, line 291: This should instead refer to the features section in `man nix.conf` as we did for other hints in error messages recently.

src/libstore/build/worker.cc
```diff
@@ -283,9 +283,14 @@ void Worker::run(const Goals & _topGoals)
...
+                            "\nYou can enable -v verbosity to see why any machine "
```
fricklerhandwerk, line 292:
```suggestion
                            "\nEnable -v verbosity to show why any machine "
```

src/build-remote/build-remote.cc
```diff
@@ -137,14 +144,19 @@ static int main_build_remote(int argc, char * * argv)
...
-                                neededSystem) != m.systemTypes.end()) &&
-                        m.allSupported(requiredFeatures) &&
-                        m.mandatoryMet(requiredFeatures))
-                    {
+                    if (!m.enabled) {
+                        debug("declined remote machine '%s' because it is disabled", storeUri);
+                    } else if (neededSystem != "builtin" &&
+                        std::find(m.systemTypes.begin(), m.systemTypes.end(), neededSystem)
+                        == m.systemTypes.end()) {
+                        debug("declined remote machine '%s' for building '%s' because it does not have the needed system type '%s", m.storeUri, drvstr, neededSystem);
```
fricklerhandwerk, line 152: Which one does it have then?

fricklerhandwerk (changes requested):
I think we should be more explicit with the expectations that were failed as represented by the error message. I'd be in favor of having a very structured presentation similar to what we had in the profile conflict PR, e.g

```
error condition

  expected: ...

  actual: ...

hint on what to do
```

@roberth we may want to write a guideline :)

roberth:
> I think we should be more explicit with the expectations that were failed as represented by the error message. I'd be in favor of having a very structured presentation similar to what we had in the profile conflict PR, e.g
> 
> ```
> error condition
> 
>   expected: ...
> 
>   actual: ...
> 
> hint on what to do
> ```

Explaining why nothing doesn't satisfy a set of constraints is always a recipe for long and complicated error messages. That's just the nature of the problem: there's a lot of data involved and we can't guess what should match up with what.
In this case we'd have to track exactly why each of the remote builders does not satisfy the requirements and pass that information to the exception constructor.

Realistically the error message would have to be

```
error condition

expected:

not builder 1
  reasons

not builder 2
  reasons

not builder 3
  reasons

...
```

You could sort the build machines by relevance and truncate after some limit (say, 3) to reduce the length of the message.

To make sure that the algorithm for checking and the algorithm for giving reasons are in sync, we'll want to use some abstractions that we don't have yet, or perhaps we should accept that we're computing and remembering too much in the happy flow.

fricklerhandwerk:
If I read the code correctly, the debug outputs are per machine, do all we need at this point are slightly more verbose texts that simply also contain the expected properties. And I only propose to make them multiline and possibly indented to make sure that the additional visual noise doesn't preclude noticing the relevant bits.

I don't see why this would require additional infrastructure.

roberth:
I assumed that you wanted to incorporate the details into the exception message instead, so that extra log verbosity isn't needed. If it remains a `lvlDebug` thing, the performance cost isn't too important.

fricklerhandwerk:
@roberth I have no opinion on whether it should be in the exception. If setting up the infra to lift it to the exception level I'd rather merge it as debug messages and allow for following up in a more principled manner when time allows.

Ericson2314:
#7865 structured exceptions would help with this.

src/build-remote/build-remote.cc
tomberek: likely the message is backwards

roberth: It's accurate as is, but "decline" might not be the best word perhaps (for all of these)

src/build-remote/build-remote.cc
```diff
@@ -137,14 +144,19 @@ static int main_build_remote(int argc, char * * argv)
...
-                    if (m.enabled
-                        && (neededSystem == "builtin"
-                            || std::find(m.systemTypes.begin(),
-                                m.systemTypes.end(),
-                                neededSystem) != m.systemTypes.end()) &&
-                        m.allSupported(requiredFeatures) &&
-                        m.mandatoryMet(requiredFeatures))
-                    {
+                    if (!m.enabled) {
+                        debug("declined remote machine '%s' because it is disabled", storeUri);
```
roberth, line 148:
```suggestion
                        debug("avoiding remote machine '%s' because it is disabled", storeUri);
```

How about avoid for all of these?

The problem with decline is that it seems like the machine is initiating the remote build or something.

PR #3934, opened 2020-08-15
Add plugin command support for `nix repl`

lf-:
Motivation: I want to make https://github.com/lf-/nix-doc a repl command, but it is useful to be able to extend the REPL otherwise.

Side note: In case my documentation changes look weird, there was an error in the documentation, which previously claimed that `RegisterCommand` was possible from a plugin, which it is not, because it is in `src/nix` and thus cannot be linked to by a library since that directory only links into an executable. I encountered this myself when initially implementing this feature and that is why I had to refactor the REPL logic out into libexpr.

- Refactor the repl core into libexpr
- Use dependency injection via std::function to provide the completion
  functions from editline so we don't introduce extra dependencies for
  libexpr
- Add a RegisterReplCmd analogous to RegisterPrimOp for repl commands
- Refactor: get rid of the "ugly" global curRepl and replace it with a
  trick with closures on the nix side (that are effectively globals
  [static vars in a templated function per-closure], but
  unique per repl user and thus not as ugly ;p)
- Rip out readline support since there appears to be no build system
  support for it according to `rg READLINE` and it is thus dead code
- Integration test this new plugin functionality
- Document it (see above)

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -558,6 +558,18 @@ Path getDataDir()
     return dataDir ? *dataDir : getHome() + "/.local/share";
 }
 
+Strings editorFor(const Path & file, unsigned int line)
+{
+    auto editor = getEnv("EDITOR").value_or("cat");
+    auto args = tokenizeString<Strings>(editor);
+    if (line > 0 && (
+        editor.find("emacs") != std::string::npos ||
+        editor.find("nano") != std::string::npos ||
+        editor.find("vim") != std::string::npos))
+        args.push_back(fmt("+%d", line));
+    args.push_back(file);
+    return args;
+}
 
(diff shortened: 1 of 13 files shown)


src/libutil/util.cc
```diff
@@ -558,6 +558,18 @@ Path getDataDir()
     return dataDir ? *dataDir : getHome() + "/.local/share";
 }
 
+Strings editorFor(const Path & file, unsigned int line)
```
lf-, line 561: extracted because it needed to be available from libexpr and didn't inherently depend on any nix expression stuff

lf-:
I am not interested in pursuing these changes any further due to lack of feedback and potential future integration of a documentation system.

PR #3935, opened 2020-08-15
Get rid of Hash::dummy from BinaryCacheStore

Ericson2314:
CC @roberth. I think between this `BinaryCacheStore::addToStoreCommon`  and your `RemoteStore::addCAToStore` we are narrowing in on "the one `addToStore` to rule them all".

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -401,19 +401,30 @@ StorePath BinaryCacheStore::addToStore(const string & name, const Path & srcPath
     /* FIXME: Make BinaryCacheStore::addToStoreCommon support
        non-recursive+sha256 so we can just use the default
        implementation of this method in terms of addToStoreFromDump. */
-    StringSink sink;
-    std::optional<Hash> h;
+
+    HashSink sink { hashAlgo };
     if (method == FileIngestionMethod::Recursive) {
         dumpPath(srcPath, sink, filter);
-        h = hashString(hashAlgo, *sink.s);
     } else {
-        auto s = readFile(srcPath);
-        dumpString(s, sink);
-        h = hashString(hashAlgo, s);
+        readFile(srcPath, sink);
     }
+    auto h = sink.finish().first;
 
-    auto source = StringSource { *sink.s };
-    return addToStoreFromDump(source, name, FileIngestionMethod::Recursive, htSHA256, repair);
+    auto source = sinkToSource([&](Sink & sink) {
+        dumpPath(srcPath, sink, filter);
+    });
+    return addToStoreCommon(*source, repair, CheckSigs, [&](HashResult nar) {
+        ValidPathInfo info {
+            makeFixedOutputPath(method, h, name),
+            nar.first,
+        };
+        info.narSize = nar.second;
+        info.ca = FixedOutputHash {
+            .method = method,
+            .hash = h,
+        };
+        return info;
+    })->path;
 }
 
 StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s,
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -311,7 +311,7 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
 
     (void) addToStoreCommon(narSource, repair, checkSigs, {[&](HashResult nar) {
         /* FIXME reinstate these, once we can correctly do hash modulo sink as
-           needed. */
+           needed. We need to throw here in case we uploaded a corrupted store path. */
         // assert(info.narHash == nar.first);
         // assert(info.narSize == nar.second);
         return info;
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -142,7 +142,7 @@ struct FileSource : FdSource
     }
 };
 
-StorePath BinaryCacheStore::addToStoreCommon(
+ref<const ValidPathInfo> BinaryCacheStore::addToStoreCommon(
     Source & narSource, RepairFlag repair, CheckSigsFlag checkSigs,
     std::function<ValidPathInfo(HashResult)> mkInfo)
 {
@@ -297,7 +297,7 @@ StorePath BinaryCacheStore::addToStoreCommon(
 
     stats.narInfoWrite++;
 
-    return narInfo->path;
+    return narInfo;
 }
 
 void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource,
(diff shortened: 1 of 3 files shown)


roberth (commented):
Implementing `addCAToStore` *might* make sense if it's easy to implement, but it will need extra tests because it's not part of "normal" Nix use afaict.

src/libstore/binary-cache-store.hh
```diff
@@ -72,6 +72,10 @@ private:
 
     void writeNarInfo(ref<NarInfo> narInfo);
 
+    StorePath addToStoreCommon(
+        Source & narSource, RepairFlag repair, CheckSigsFlag checkSigs,
+        std::function<ValidPathInfo(HashResult)> mkInfo);
```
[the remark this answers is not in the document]
Ericson2314: Sounds good! To be clear I I think it's better in a future PR not this one to actually start unifying the `addToStore` interfaces. But we can do that change no for free to prepare for that :).

src/libstore/binary-cache-store.cc
```diff
@@ -300,6 +296,41 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
...
     stats.narInfoWrite++;
+
+    return narInfo->path;
+}
+
+void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource,
+    RepairFlag repair, CheckSigsFlag checkSigs)
+{
+    if (!repair && isValidPath(info.path)) {
+        // FIXME: copyNAR -> null sink
```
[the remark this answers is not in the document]
Ericson2314: I think agree it's better to make the caller responsible, but I'm a bit wary on changing the direction of this FIXME as it and this code already existed, I just moved it here.

I'll let @edolstra decide :).

src/libstore/binary-cache-store.cc
```diff
@@ -162,23 +155,26 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
     /* Read the NAR simultaneously into a CompressionSink+FileSink (to
        write the compressed NAR to disk), into a HashSink (to get the
        NAR hash), and into a NarAccessor (to get the NAR listing). */
-    HashSink fileHashSink(htSHA256);
+    HashSink fileHashSink { htSHA256 };
     std::shared_ptr<FSAccessor> narAccessor;
+    HashSink narHashSink { htSHA256 };
     {
     FdSink fileSink(fdTemp.get());
```
roberth, line 162: This is beyond the scope of this PR, but since we're revisiting these store API methods it's worth noting that we could optimize away the tmpfile if we know a little bit more in advance. Although optimizing away a tmpfile seems unimpressive, it changes the time taken from `O(sum(steps))` to `O(max(steps))`, which is significant when compression and upload take similar amounts of time.

***When the file is known to be small (low hanging fruit)***

Add a size parameter to `addToStoreCommon` or use a fancy sink that only writes to file when it crosses a limit. This is similar to what `LocalStore::addToStoreFromDump` does.

***When we know the nar hash in advance***

For http binary caches this does require us to change the binary cache filenames to match *uncompressed* hashes, which seems to be equivalent and can only result in one-time duplication in existing caches when new paths are uploaded.
I don't know yet how IPFS caches fit into this picture, but if those can compress after hashing, this would be beneficial.
Another reason to do this is so we don't need to compress before we can decide to reuse an available nar file.

In this case it does make sense to have both `addToStore(const ValidPathInfo & info, .....)` and `addToStore(....., std::function<ValidPathInfo(HashResult))` where the prior can have a default implementation in terms of the latter.

roberth (commented):
1× fixme
1× store api and binary cache idea. I'll make an issue for that

src/libstore/binary-cache-store.cc
```diff
@@ -162,23 +155,26 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
...
     }
 
     auto now2 = std::chrono::steady_clock::now();
 
+    auto info = mkInfo(narHashSink.finish());
     auto narInfo = make_ref<NarInfo>(info);
-    narInfo->narSize = info.narSize;
-    narInfo->narHash = info.narHash;
+    narInfo->narSize = info.narSize; // FIXME needed?
+    narInfo->narHash = info.narHash; // FIXME needed?
```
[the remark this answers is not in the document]
Ericson2314: Yeah I guess I was feeling spooked when it wasn't working, but now I have no issue removing those.

src/libstore/binary-cache-store.cc
edolstra: I decided not to try to optimize away the temporary file for small NARs because the overhead is likely to be insignificant compared to stuff like HTTP requests.

src/libstore/binary-cache-store.cc
```diff
@@ -142,17 +142,10 @@ struct FileSource : FdSource
...
-    RepairFlag repair, CheckSigsFlag checkSigs)
+ref<const ValidPathInfo> BinaryCacheStore::addToStoreCommon(
+    Source & narSource, RepairFlag repair, CheckSigsFlag checkSigs,
+    std::function<ValidPathInfo(HashResult)> mkInfo)
 {
-    assert(info.narSize);
-
-    if (!repair && isValidPath(info.path)) {
-        // FIXME: copyNAR -> null sink
-        narSource.drain();
```
edolstra, line 152: Are we sure we don't need this anymore? Previously the drain was necessary to ensure the postcondition that the NAR is completely read, to avoid leaving the connection in an unknown state. `FramedSource` should address that, but maybe there are still places where it could happen.

edolstra: Oh I see it's still drained elsewhere. Sorry :-)

src/libstore/binary-cache-store.cc
```diff
@@ -300,6 +294,41 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
...
+void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource,
+    RepairFlag repair, CheckSigsFlag checkSigs)
+{
+    if (!repair && isValidPath(info.path)) {
+        // FIXME: copyNAR -> null sink
+        narSource.drain();
+        return;
+    }
+
+    (void) addToStoreCommon(narSource, repair, checkSigs, {[&](HashResult nar) {
```
edolstra, line 310:
```suggestion
    addToStoreCommon(narSource, repair, checkSigs, {[&](HashResult nar) {
```

edolstra (approved):

Ericson2314:
@edolstra OK this is all ready.

PR #3940, opened 2020-08-16
Add commented-out test for remote building with fixed output derivations

Ericson2314:
Since there is a preexisting bug, this serves as a reminder to fix it.


Ericson2314:
Great! This confirms it's an existing bug.

edolstra:
What's the bug exactly? Are remote fixed-output derivations broken?

PR #3941, opened 2020-08-18
util: concatStringsSep(): Avoid quadratic string append

nh2:
Repeated `+=` makes no complexity guarantees; this does.

I also renamed the variables for some more clarity (to be aligned with the docs of the function).

### DIFF
--- a/src/libutil/util.hh
+++ b/src/libutil/util.hh
@@ -355,16 +355,25 @@ template<class C> C tokenizeString(std::string_view s, const string & separators
 
 
 /* Concatenate the given strings with a separator between the
-   elements. */
+   elements. The elements must be assignable to `std::string`. */
 template<class C>
-string concatStringsSep(const string & sep, const C & ss)
+string concatStringsSep(const string & sep, const C & elems)
 {
-    string s;
-    for (auto & i : ss) {
-        if (s.size() != 0) s += sep;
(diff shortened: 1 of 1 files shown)


nh2:
I also noticed that the `prim_concatStringSep` is very optimistic, assuming that each list element is 32 chars long:

https://github.com/NixOS/nix/blob/e849b198720c60c186c8f9486c43c495ad436e1b/src/libexpr/primops.cc#L2113-L2120

If a nix language user uses `builtins.concatStringsSep` on a long list of long strings (e.g. to construct some big file), this will also be O(n²).

Could we use the same approach there, computing `state.coerceToString(pos, *args[1]->listElems()[n], context);` for each element ahead of time, of alternatively `res.reserve()`ing the buffer size exponentially to avoid that from happening?

edolstra:
Doesn't std::string grow the capacity by a constant factor when it exceeds the current capacity, thereby amortizing the cost to O(1) per character?

nh2:
> Doesn't std::string grow the capacity by a constant factor when it exceeds the current capacity, thereby amortizing the cost to O(1) per character?

Many implementations do, but the standard doesn't guarantee it, see full answer in https://github.com/NixOS/nix/pull/3942#issuecomment-675378630.

PR #3942, opened 2020-08-18
builtins.concatStringsSep: Guarantee amortised O(n)

nh2:
Fixes https://github.com/NixOS/nix/pull/3941#issuecomment-675198161.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -2111,12 +2111,18 @@ static void prim_concatStringSep(EvalState & state, const Pos & pos, Value * * a
     state.forceList(*args[1], pos);
 
     string res;
-    res.reserve((args[1]->listSize() + 32) * sep.size());
+    res.reserve((args[1]->listSize() + 32) * sep.size()); // guess
     bool first = true;
 
     for (unsigned int n = 0; n < args[1]->listSize(); ++n) {
         if (first) first = false; else res += sep;
-        res += state.coerceToString(pos, *args[1]->listElems()[n], context);
+        string s = state.coerceToString(pos, *args[1]->listElems()[n], context);
+        // Exponential growth for amortised overall linear append.
+        if (res.capacity() < res.size() + s.length()) {
+            size_t add = std::max(s.length(), (size_t) (0.5 * res.capacity()));
+            res.reserve(res.capacity() + add);
+        }
+        res += s;
     }
 
     mkString(v, res, context);


Ericson2314:
What if we made another C++ `concatStringsSep` that took an iterator (something I wanted anyways, FWIW) and then wrote them both interms of that?

nh2:
Answering https://github.com/NixOS/nix/pull/3941#issuecomment-675294833:

> Doesn't std::string grow the capacity by a constant factor when it exceeds the current capacity, thereby amortizing the cost to O(1) per character?

There's no C++ standard guarantee for it: https://en.cppreference.com/w/cpp/string/basic_string/operator%2B%3D:

> **Complexity**
> There are no standard complexity guarantees

The amortised-constant guarantee that the standard gives for `vector`/`string` is on [`push_back()`](https://en.cppreference.com/w/cpp/string/basic_string/push_back).

GCC's `libstdc++` and LLVM's `libc++` currently implement `+=` with exponential growth (see e.g. [this, section Growth strategy](https://shaharmike.com/cpp/std-string/)), but they may change it and other standard library implementations may not do it this way.

So it's more of a "do you want to code against API guarantees or implementation details" choice.

This PR guarantees the complexity relying only on what the standard says. It would also be OK to rely on the implementation but I think that should have a comment then.

edolstra:
I don't think we need to worry too much about theoretical implementations not doing the sane thing. Having said that, it would be nice to avoid doing *any* reallocations, but then we should evaluate each string, sum the sizes and `reserve()` the entire result string.

BTW another optimization that might be worthwhile is to avoid allocating the temporary strings returned by `coerceToString()` (which is wasteful if the value is a plain string). Ideally it would return something like Rust's `Cow` type, i.e. either a string or a string_view.

nh2:
> Having said that, it would be nice to avoid doing _any_ reallocations, but then we should evaluate each string, sum the sizes and `reserve()` the entire result string.

@edolstra How should that best be done? There seem to be 2 straightforward ways:

* Computing `state.coerceToString(pos, *args[1]->listElems()[n], context);` twice for each string (the first time just to compute its length, the second time for copying it over). I'm a bit scared of that because inside `coerceToString()` the `v.type == tPath` looks like it'll have to `copyPathToStore()`, which requires a `typedef std::map<Path, StorePath> SrcToStore` lookup, which _might_ be slower than what the optimisation provides.
* Computing once it once and storing the returned `string`s (but this would require ~50% more maximum memory, assuming that at the worst point we'd have to keep around the original nix strings, the output of the corercion, and the final string buffer), unless your `Cow` idea is also implemented.

edolstra:
The second option should be fine since they're temporary strings anyway and probably not very large.

Ericson2314:
Hold up, fine if stale bot is marking things stale, but it should be closing them. We need to actually triage old stuff. We can't just let a bot due our duties for us.

I assume this was just a misconfiguration, as the Nixpkgs stale bot doesn't close things to my knowledge.

nh2:
It looks like `nix`'s stale bot was configured to close already 14 months ago:

https://github.com/NixOS/nix/pull/4526#pullrequestreview-585069597

It also closed some other PRs that I think really shouldn't be closed: https://github.com/NixOS/nix/pull/3208#issuecomment-1100630134

PR #3955, opened 2020-08-24
improve help text for tarball-ttl

DavHau:
Add unit for tarball-ttl to help text


domenkozar:
Much better help text was added since.

PR #3958, opened 2020-08-24
CA derivations that depend on other CA derivations

Ericson2314:
This is tested and works, but some things should be cleaned wrt the new TODOs.

Feature-wise, the good news is that we no longer rebuild resolution-equivalent drvs for CA derivations. We resolve the derivation if it isn't already resolved, and then build that instead.

The bad news is input-addressed derivations still cannot depend on CA derivations, that will wait for later.

Finally, thanks to @regnat, as I stole a few things of stuff from #3776 :).


regnat (commented):
> The bad news is input-addressed derivations still cannot depend on CA derivations, that will wait for later.

What's preventing this except the need to recompute the output paths?

src/libstore/local-store.cc
```diff
@@ -803,13 +803,36 @@ StorePathSet LocalStore::queryValidDerivers(const StorePath & path)
...
+        auto resolvedPathOptIter = resolutions->find(path);
+        if (resolvedPathOptIter != resolutions->end()) {
+            auto & [_, resolvedPathOpt] = *resolvedPathOptIter;
+            if (resolvedPathOpt)
+                path = *resolvedPathOpt;
+            haveCached = true;
+        }
+    }
+    /* can't just use else-if instead of `!haveCached` because we need to unlock
+       `drvPathResolutions` before it is locked in `Derivation::resolve`. */
```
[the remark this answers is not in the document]
Ericson2314: This function and `queryPartialDerivationOutputMap` are mutually recursive, so it would deadlock.

tests/content-addressed.sh
```diff
@@ -2,15 +2,17 @@
...
+    local commonArgs=("--experimental-features" "ca-derivations" "./content-addressed.nix" "-A" "$derivationPath" "--no-out-link")
+    local out1=$(nix-build "${commonArgs[@]}" --arg seed 1)
+    local out2=$(nix-build "${commonArgs[@]}" --arg seed 2)
+    test $out1 == $out2
+}
 
-test $out1 == $out2
+testDerivation root
+testDerivation dependent
+testDerivation transitivelyDependent
```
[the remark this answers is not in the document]
Ericson2314: Good idea, let's through some `-j0` in there.

src/libstore/build.cc
```diff
@@ -4248,10 +4285,14 @@ void DerivationGoal::registerOutputs()
...
         ValidPathInfos infos2;
         for (auto & [outputName, newInfo] : infos) {
-            /* FIXME: we will want to track this mapping in the DB whether or
-               not we have a drv file. */
             if (useDerivation)
                 worker.store.linkDeriverToPath(drvPath, outputName, newInfo.path);
+            else {
+                /* Once a floating CA derivations reaches this point, it must
+                   already be resolved, drvPath the basic derivation path, and
+                   a file existsing at that path for sake of the DB's foreign key. */
```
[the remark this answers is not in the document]
Ericson2314: Yes good catch, it does not hold, and with the remote builder that is the expected behavior! :O I guess we write down the derivation in that case?

src/libstore/build.cc
```diff
@@ -1369,23 +1453,59 @@ void DerivationGoal::inputsRealised()
...
+                worker.store.printStorePath(pathResolved));
+            act = std::make_unique<Activity>(*logger, lvlInfo, actBuildWaiting, msg,
+                Logger::Fields {
+                       worker.store.printStorePath(drvPath),
+                       worker.store.printStorePath(pathResolved),
+                   });
+
+            auto resolvedGoal = worker.makeDerivationGoal(
+                pathResolved, wantedOutputs,
+                buildMode == bmRepair ? bmRepair : bmNormal);
```
Ericson2314, line 1481: Good catch, that was a bad cargo could on my end from line 1347. There it makes sence that we don't also want to check dependencies, here I do think we would want to check the resolved derivation.
```suggestion
                pathResolved, wantedOutputs, buildMode);
```

Ericson2314:
> What's preventing this except the need to recompute the output paths?

Nothing major. I'd need to change the types and drv file format at least somewhat to accommodate the lack of a known output path, and I just didn't want to get into that yet until this was all good.

tests/content-addressed.sh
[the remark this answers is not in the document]
Ericson2314: Done

src/libstore/local-store.cc
```diff
@@ -785,13 +803,36 @@ StorePathSet LocalStore::queryValidDerivers(const StorePath & path)
...
+    if (!haveCached && drv.type() == DerivationType::CAFloating) {
+        /* Resolve drv and use that path instead. */
+        auto pathResolved = writeDerivation(*this, drv.resolve(*this));
```
Ericson2314, line 829: Ah! so this can fail if one of the new `inputSrcs` from derivation resolution isn't a path we've already realized. This has convinced me we should relax the foreign key on the `DerivationOutputs` table after all, so that we can just use `makeTextPath` here and not care that those references are dangling.

Ericson2314: After the last round of improvements to the test suite, this issue is reproduced and the test suite fails. PR marked WIP accordingly.

Ericson2314:
OK, I think this is ready. Everything should work except for remote building, which will require changes to the build hook protocol to fix. I think that's best dealt with in another PR.

regnat:
I tried bootsraping firefox with this branch (as I had done for #3883), but I ran into a lot of (apparently random) failures.
These seem to appear randomly (and non-deterministically) so I haven't been able to put a proper repro yet (nor confirm that it's specifically due to this branch and not something occuring on master).

The errors are
- `opening file '/proc/24534/uid_map': Permission denied` (I didn't double-check, but I guess that the pid changes between two errors)
- `creating pipe: Too many open files`
- `opening directory '/tmp/nix-build-foo-0.0.drv-0': Bad file descriptor` followed by `opening directory '/home/regnat/test-store/nix/store/xxx-foo-0.0.drv.chroot/bin': Bad file descriptor` followed by `creating pipe: Too many open files`

regnat:
> nor confirm that it's specifically due to this branch and not something occuring on master

It's still building the world, but nothing crashed so far on master, so it's likely to be specific to this branch

Ericson2314:
@regnat huh, weird. Are those happening when building input-addressed derivations too?

regnat:
I tried again tonight and this morning, and I actually hit twice the same kind of error on master.

```
error (ignored): error: --- SysError --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix-build
opening directory '/home/regnat/test-store-master/nix/store/iyfs2mp14qq1rzpcm5xvn101af5ih17p-libcanberra-0.30.tar.xz.drv.chroot/var/run': Bad file descriptor
error: --- SysError --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix-build
creating pipe: Too many open files
```

It seems to be less frequent than what I had yesterday, but might just be an accident − I'm using a shared machine for that, maybe it had a higher load yesterday or something like that which caused it to trigger this more often.

Ericson2314:
@regnat Just looking at the code, I had no idea what could go wrong/differently in the non-floating-CA case, so I'm relieved you could reproduce without this branch!

Anything else need changing here then? With the latest pushes I'm writing down resolved trees for sake of caching (resolving means no super-linear storage overhead, and no new network overhead, thankfully) but we can always revisit these things later.

src/libstore/derivations.hh
```diff
@@ -137,7 +147,9 @@ enum RepairFlag : bool { NoRepair = false, Repair = true };
 
 /* Write a derivation to the Nix store, and return its path. */
 StorePath writeDerivation(Store & store,
-    const Derivation & drv, RepairFlag repair = NoRepair);
+    const Derivation & drv,
+    RepairFlag repair = NoRepair,
+    bool readOnly = false);
```
regnat, line 152: It looks like this flag isn't used anywhere

src/libstore/local-store.cc
```diff
@@ -805,18 +805,58 @@ StorePathSet LocalStore::queryValidDerivers(const StorePath & path)
...
+       `drvPathResolutions` before it is locked in `Derivation::resolve`. */
+    if (!haveCached && drv.type() == DerivationType::CAFloating) {
+        /* Try resolve drv and use that path instead. */
+        auto attempt = drv.tryResolve(*this);
+        if (!attempt)
+            /* If we cannot resolve the derivation, we cannot have any path
+               assigned so we return the map of all std::nullopts. */
+            return outputs;
+        /* Just compute store path */
+        auto pathResolved = writeDerivation(*this, *std::move(attempt), NoRepair, true);
```
Ericson2314, line 837: @regnat I thought the same thing! But it is still used here.

regnat: Oh thanks, missed that one ;)

Ericson2314:
https://github.com/NixOS/nix/pull/4056#issuecomment-698188488 I take it this means we agree this is ready.

src/libstore/derivations.cc
```diff
@@ -644,4 +644,57 @@ std::string downstreamPlaceholder(const Store & store, const StorePath & drvPath
     return "/" + hashString(htSHA256, clearText).to_string(Base32, false);
 }
 
+
+// N.B. Outputs are left unchanged
+static void rewriteDerivation(Store & store, BasicDerivation & drv, const StringMap & rewrites) {
```
edolstra, line 649:
```suggestion
static void rewriteDerivation(Store & store, BasicDerivation & drv, const StringMap & rewrites) 
{
```

Style nitpick.

src/libstore/build.cc
```diff
@@ -1464,8 +1466,40 @@ void DerivationGoal::inputsRealised()
...
+
+        if (!fullDrv.inputDrvs.empty() && fullDrv.type() == DerivationType::CAFloating) {
+            /* We are be able to resolve this derivation based on the
+               now-known results of dependencies. If so, we become a stub goal
+               aliasing that resolved derivation goal */
+            std::optional attempt = fullDrv.tryResolve(worker.store);
+            assert(attempt);
+            Derivation drvResolved { *std::move(attempt) };
+
+            auto pathResolved = writeDerivation(worker.store, drvResolved);
```
Ericson2314, line 1480: Note we are writing down the resolved derivation so we can link the deriver to the CA output in the sqlite db.

edolstra: Do we need that? It doubles the number of .drv files on disk, while we'd like to get rid of them eventually...

src/libstore/build.cc
```diff
@@ -1906,6 +1941,9 @@ void DerivationGoal::buildDone()
     done(BuildResult::Built);
 }
 
+void DerivationGoal::resolvedFinished() {
```
edolstra, line 1944: This function seems a bit unnecessary.

src/libstore/build.cc
```diff
@@ -4274,12 +4314,21 @@ void DerivationGoal::registerOutputs()
...
+    bool isCaFloating = drv->type() == DerivationType::CAFloating;
+
+    auto drvPath2 = drvPath;
+    if (!useDerivation && isCaFloating) {
+        /* Once a floating CA derivations reaches this point, it
+           must already be resolved, so we don't bother trying to
+           downcast drv to get would would just be an empty
+           inputDrvs field. */
+        Derivation drv2 { *drv };
+        drvPath2 = writeDerivation(worker.store, drv2);
```
edolstra, line 4326: `drvPath2` looks unused?

src/libstore/derivations.hh
```diff
@@ -127,7 +128,16 @@ struct Derivation : BasicDerivation
     std::string unparse(const Store & store, bool maskOutputs,
         std::map<std::string, StringSet> * actualInputs = nullptr) const;
 
-    Derivation() { }
+    /* Return the underlying basic derivation but with
+
+       1. input drv outputs moved to input sources.
```
edolstra, line 133: What does this mean?

src/libstore/local-store.cc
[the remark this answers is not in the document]
edolstra: Maybe there should be internal variants that take a reference to the locked `drvPathResolutions`? Then you only need to lock it once.

edolstra (approved):

src/libstore/build.cc
Ericson2314: It's now used below in the `linkDeriverToPath` call as originally intended, and given a better name.

src/libstore/build.cc
Ericson2314: Well many derivation goals could resolve to the same derivation, Trying to make the resolve derivations track all those to call `done` on seems wrong to me.

src/libstore/build.cc
Ericson2314: Well it only doubles them for floating CA derivations. Not doing this means changing the SQL schema, and I think there's a few such change's will end up wanting to make, so I rather save that for later to think about them all at once.

PR #3959, opened 2020-08-25
Derivations can output "text-hashed" data

Ericson2314:
# Motivation

We're finally getting to RFC 92 features proper!

Derivations are currently serialized to store objects (and also thus given drv paths, even when we just stick with in-memory derivations) using "text hashing" / "text file ingestion method", a type of content addressing. That means, if we want to compute derivations we need to use

 - content-addressed derivations (already have!)
 - with derivations outputs using text ingestion method (don't yet have!)

The latter part is what we are adding.

# Context

A few things needed to make this go.

1. A little cleanup of `ContentAddressMethod` building on #3746. The two variants are now `TextIngestionMethod` and `FileIngestionMethod`, which together are tantamount to a 3-way enum:
   - flat ingestion
   - Nix archive ingestion ("recursive")
   - text ingestion
  
   This data type no longer species any hash type, but is *just* the way of ingestion data. A nice single purpose now.

2. `TextHashMethod` -> `TextIngestionMethod`, since it no longer implies SHA-256 or any other hash type (or hashing at all!), though out of an abundance of caution we disallow trying to use any other hash type with it. (We're separating concerns but not supporting all the options yet!)

3. `FixedOutputHashMethod` is deleted. It was a type for a (hash type, flat | recursive) pair, but we simply don't need that anymore. Everything supports text hashing too now.

4. The new stuff is gated under a new `dynamic-derivations` feature. I imagine I'll just put all the constituent features that make up dynamic derivations under this, as they are meant to be used together.

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
 - [ ] documentation in the internal API docs

# Priorities

In particular, this means that derivations can output derivations. But that ramification isn't (yet!) useful as we would want, since there is no way to have a dependent derivation that is itself a dependent derivation.

~~depends on #3746~~


kisik21:
I'd like to learn what are the possible use-cases once this feature becomes truly useful and complete (meaning that the drawback you mentioned earlier is solved). This feels very confusing to me as a casual Nix user

Ericson2314:
@kisik21 the use-case is instead of using IPFD, we can build derivations produced by derivations. For example, we could do `nix-instantiate $(cabal2nix ..) > $out` in a derivation-producing derivation.

kisik21:
Oh, I see! So the extra step of running node2nix or whatever gets used to package up node packages, or something else, is not needed with derivation-ception that's being made possible here.

Or should we call it Nix-ception? 😂

Ericson2314:
This should pass, and is rebased to now only contain one other PR.

L-as:
I still think we should remove text hashing entirely.

anka-213:
> the use-case is instead of using IPFD, we can build derivations produced by derivations. For example, we could do `nix-instantiate $(cabal2nix ..) > $out` in a derivation-producing derivation.

So how is this different from IFD? It sounds very similar to my untrained ears. (I'm assuming you meant IFD and had just been  writing "IPFS" so many times in the context of this PR that we got a hybrid between the two ;)

Ericson2314:
@anka-213 IFD is about the evaluator kicking off builds and waiting for the results. This is pure store layer, merely derivations that produce derivations.

amjoseph-nixpkgs:
> I'd like to learn what are the possible use-cases once this feature becomes truly useful and complete

Fun fact: you can use hash-scanning instead of "string contexts" to calculate `inputDrvs`.  I wrote a tool (using [libnixstore](https://github.com/helsinki-systems/harmonia/tree/master/libnixstore)) to run both calculations and compare them; in my full-workstation build of nixpkgs this is the only one place where they disagreed:

https://github.com/NixOS/nixpkgs/blob/4102f23e2d602e81cef029e236c503c56ee085ad/pkgs/stdenv/linux/make-bootstrap-tools.nix#L275

Ericson2314:
(Just marked it draft because the PR it depends on is unmerged, and the dependency bot isn't really good enough to track this on its own.)

Ericson2314:
Once https://github.com/NixOS/nix/pull/8231 is merged, I can finish adding feature gates, and then shit should be ready!

src/libstore/content-address.hh
```diff
@@ -22,7 +22,7 @@ namespace nix {
  * Somewhat obscure, used by \ref Derivation derivations and
  * `builtins.toFile` currently.
  */
-struct TextHashMethod : std::monostate { };
+struct TextIngestionMethod : std::monostate { };
```
amjoseph-nixpkgs, line 25: Maybe there should be an explanation somewhere of exactly what the difference is between `TextIngestionMethod` and `FileIngestionMethod::Flat`. 

Here's the what I could come up with:

- they are hashed identically but tagged differently: `text:${algo}:${hash}` instead of `fixed:${algo}:${hash}`
- text-ingested store paths may not have self-references

There must be something else that I'm missing though, right?  It seems like it's just one extra bit on `FileIngestionMethod::Flat` and a no-self-references assertion.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -2426,37 +2426,43 @@ SingleDrvOutputs LocalDerivationGoal::registerOutputs()
...
+                    switch (m2) {
+                    case FileIngestionMethod::Recursive:
+                        dumpPath(actualPath, caSink);
+                        break;
+                    case FileIngestionMethod::Flat:
+                        readFile(actualPath, caSink);
+                        break;
+                    }
+                },
+            }, outputHash.method.raw);
```
amjoseph-nixpkgs, line 2457: There are a lot of `switch`es where `case TextIngestionMethod` is just a copy-paste duplicate of `case FileIngestionMethod::Flat`.  Can't C++ represent disjunctive deep pattern-matching in these cases like ML/Rust/etc can?  If not, it seems like the code would be considerably simplified if `Text` were a third variant of the `FileIngestionMethod` enum.

I know that decision (to have `TextHashMethod` be separate from `FileHashMethod` predates this PR), but that technical debt seems to be getting amplified.  I wanted to figure out what the difference was between `TextIngestionMethod` and `FileIngestionMethod::Flat`, so I figured I would just look for the one or two places where those codepaths diverged... and there turn out to be closer to a dozen places, yet a lot of them diverge only to do the same thing on both branches (!).

Or maybe I've totally misunderstood what's going on here and there really is some fundamental difference between `text:` and `fixed:`-without-self-references.

src/libstore/content-address.hh
Ericson2314: Yup that's correct. The emphasis for me is on the first one: ultimately `ContentAddressWithReferences` is supposed to represent every of content-addressed store path. `Flat` and `Text` may actually be the same thing, but so long as `text:` and `fixed:` are not, then they are different *by fiat*.

src/libstore/build/local-derivation-goal.cc
Ericson2314: Yeah per my other comment, there is no fundamental difference, `text:` vs `fixed:` store paths.

> Can't C++ represent disjunctive deep pattern-matching in these cases like ML/Rust/etc can? I

Nope. RIP.

>  If not, it seems like the code would be considerably simplified if Text were a third variant of the `FileIngestionMethod` enum.

I would be fine doing that if it works; it is after all a similar change to getting rid of `FixedOutputHashMethod`.

I didn't do it yet however because I was not yet sure that everywhere a `FixedOutputHashMethod` is taken today, a `ContentAddressingMethod` could be passed instead. In fact. I know that is not the case: we still have redundancies like `addTextToStore` vs `addToStoreFromDump` that should be combined into @roberth's `addCAToStore` (currently just a `RemoteStore` method).

So temporary I rather keep this as is: verbose C++ is annoying, but it is *much* less bad than partiality. However as mentioned in https://github.com/NixOS/nix/pull/3746 I indeed do want to clean up the tech debt of text having pointless bespoke code paths, after which yes let's gloriously flatten the enum! :)

src/libstore/content-address.hh
```diff
@@ -22,7 +22,7 @@ namespace nix {
  * Somewhat obscure, used by \ref Derivation derivations and
  * `builtins.toFile` currently.
  */
```
amjoseph-nixpkgs, line 24:
```suggestion
 * Somewhat obscure, used by \ref Derivation derivations and
 * `builtins.toFile` currently.
 *
 * TextIngestionMethod is identical to FileIngestionMethod::Fixed except that
 * the former may not have self-references and is tagged `text:${algo}:${hash}`
 * rather than `fixed:${algo}:${hash}`.  The contents of the store path are
 * ingested and hashed identically, aside from the slightly different tag and
 * restriction on self-references.
 */
```

src/libstore/build/local-derivation-goal.cc
amjoseph-nixpkgs: Please mark this resolved.

src/libstore/content-address.hh
amjoseph-nixpkgs: Please mark this resolved.

src/libstore/content-address.hh
Ericson2314: Thanks!

src/libstore/misc.cc
```diff
@@ -83,14 +83,16 @@ void Store::computeFSClosure(const StorePath & startPath,
...
+const ContentAddress * getDerivationCA(const BasicDerivation & drv)
 {
     auto out = drv.outputs.find("out");
-    if (out != drv.outputs.end()) {
-        if (const auto * v = std::get_if<DerivationOutput::CAFixed>(&out->second.raw()))
-            return v->hash;
+    if (out == drv.outputs.end())
+        return nullptr;
+    if (auto dof = std::get_if<DerivationOutput::CAFixed>(&out->second)) {
+
```
roberth, line 92:
```suggestion
```

src/libstore/misc.cc
```diff
@@ -83,14 +83,16 @@ void Store::computeFSClosure(const StorePath & startPath,
 }
 
 
-std::optional<ContentAddress> getDerivationCA(const BasicDerivation & drv)
+const ContentAddress * getDerivationCA(const BasicDerivation & drv)
```
roberth, line 86:
```suggestion
const ContentAddress * getFixedOutputAddress(const BasicDerivation & drv)
```

 - "CA" doesn't convey anything not already in the type name.
 - I think "derivation" is ok to be implied.

tests/dyn-drv/text-hashed-output.nix
```diff
@@ -0,0 +1,29 @@
+with import ./config.nix;
+
+# A simple content-addressed derivation.
+# The derivation can be arbitrarily modified by passing a different `seed`,
+# but the output will always be the same
+rec {
+  root = mkDerivation {
```
roberth, line 7:
```suggestion
  hello = mkDerivation {
```

Graph connotations aren't helpful here. It's only a root of the reverse dependency graph.

tests/dyn-drv/text-hashed-output.nix
```diff
@@ -0,0 +1,29 @@
...
+      set -x
+      echo "Building a CA derivation"
+      mkdir -p $out
+      echo "Hello World" > $out/hello
+    '';
+    __contentAddressed = true;
+    outputHashMode = "recursive";
+    outputHashAlgo = "sha256";
+  };
+  dependent = mkDerivation {
```
roberth, line 19:
```suggestion
  producingDrv = mkDerivation {
```

tests/dyn-drv/text-hashed-output.sh
```diff
@@ -0,0 +1,26 @@
...
+# we only need to:
+#
+# - instantiate the root derivation
+# - build the dependent derivation
+# - check that the path of the output coincides with that of the original derivation
+
+drv=$(nix-instantiate ./text-hashed-output.nix -A root)
+nix show-derivation "$drv"
+
+drvDep=$(nix-instantiate ./text-hashed-output.nix -A dependent)
```
roberth, line 18:
```suggestion
drvProducingDrv=$(nix-instantiate ./text-hashed-output.nix -A dependent)
```

tests/dyn-drv/text-hashed-output.sh
```diff
@@ -0,0 +1,26 @@
...
+nix show-derivation "$drvDep"
+
+out1=$(nix-build ./text-hashed-output.nix -A dependent --no-out-link)
+
+nix path-info $drv --derivation --json | jq
+nix path-info $out1 --derivation --json | jq
+
+test $out1 == $drv
```
roberth, line 26: I don't suppose we can do `nix build $drvProducingDrv^out^out` yet?

tests/dyn-drv/text-hashed-output.nix
```diff
@@ -0,0 +1,29 @@
...
+    name = "text-hashed-root.drv";
+    buildCommand = ''
+      echo "Copying the derivation"
+      cp ${builtins.unsafeDiscardOutputDependency root.drvPath} $out
```
Ericson2314, line 23:
```suggestion
      cp ${builtins.unsafeDiscardOutputDependency hello.drvPath} $out
```

tests/dyn-drv/text-hashed-output.sh
```diff
@@ -0,0 +1,26 @@
...
+# is a normal recursive derivation, while the second, named dependent, has the
+# new outputHashMode "text". Note that in "dependent", we don't refer to the
+# build output of root, but only to the path of the drv file. For this reason,
+# we only need to:
+#
+# - instantiate the root derivation
+# - build the dependent derivation
+# - check that the path of the output coincides with that of the original derivation
+
+drv=$(nix-instantiate ./text-hashed-output.nix -A root)
```
Ericson2314, line 15:
```suggestion
drv=$(nix-instantiate ./text-hashed-output.nix -A hello)
```

tests/dyn-drv/text-hashed-output.sh
Ericson2314:
```suggestion
drvProducingDrv=$(nix-instantiate ./text-hashed-output.nix -A producingDrv)
```

tests/dyn-drv/text-hashed-output.sh
Ericson2314: :) not yet, but soon!

src/libstore/content-address.hh
```diff
@@ -244,10 +272,28 @@ struct ContentAddressWithReferences
...
+     * @param method Way ingesting the file system data.
+     *
+     * @param hash Hash of ingested file system data.
+     *
+     * @param refs References to other store objects or oneself.
+     *
+     * Do note that not all combinations are supported.
+     */
+    static ContentAddressWithReferences fromParts(
+        ContentAddressMethod method, Hash hash, StoreReferences refs);
```
roberth, line 292: `noexcept` seems valuable and feasible here too.

src/libstore/daemon.cc
```diff
@@ -401,18 +401,22 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
...
-                ContentAddressMethod contentAddressMethod = ContentAddressMethod::parse(camStr);
+                auto [contentAddressMethod, hashType_] = ContentAddressMethod::parse(camStr);
+                auto hashType = hashType_; // work around clang bug
                 FramedSource source(from);
                 // TODO this is essentially RemoteStore::addCAToStore. Move it up to Store.
                 return std::visit(overloaded {
-                    [&](const TextHashMethod &) {
+                    [&](const TextIngestionMethod &) {
+                        if (hashType != htSHA256)
+                            throw UnimplementedError("Only SHA-256 is supported for adding text-hashed data, but '%1' was given",
```
roberth, line 411: `name` seems like essential context for this error message.

src/libstore/remote-store.cc
```diff
@@ -628,26 +629,29 @@ ref<const ValidPathInfo> RemoteStore::addCAToStore(
         if (repair) throw Error("repairing is not supported when building through the Nix daemon protocol < 1.25");
 
         std::visit(overloaded {
-            [&](const TextHashMethod & thm) -> void {
+            [&](const TextIngestionMethod & thm) -> void {
+                if (hashType != htSHA256)
+                    throw UnimplementedError("Only SHA-256 is supported for adding text-hashed data, but '%1' was given",
```
roberth, line 634: Also add the `name` to the message here?

roberth (commented):
Hardly anything significant it seems. :rocket:

src/libstore/build/local-derivation-goal.cc
```diff
@@ -2522,6 +2529,11 @@ SingleDrvOutputs LocalDerivationGoal::registerOutputs()
                             wanted.to_string(SRI, true),
                             got.to_string(SRI, true)));
                 }
+                if (!newInfo.references.empty())
+                    delayedException = std::make_exception_ptr(
+                        BuildError("illegal path references in fixed-output derivation '%s'",
+                            worker.store.printStorePath(drvPath)));
```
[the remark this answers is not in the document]
Ericson2314: Hmm... I think this might be fixing a big? Or at least making a more explicit error message?

CA paths depend on the references, and fixed-output derivations have fixed CA output paths: therefore the references must be fixed ahead of time.

A previous version of this PR flirted with extending fixed-output derivations to support fixed non-trivial references, but I ripped that out to avoid any mission creep.

src/libstore/content-address.cc
```diff
@@ -173,4 +228,52 @@ ContentAddressWithReferences ContentAddressWithReferences::withoutRefs(const Con
     }, ca.raw);
 }
 
+ContentAddressWithReferences ContentAddressWithReferences::fromParts(
+    ContentAddressMethod method, Hash hash, StoreReferences refs)
+{
+    return std::visit(overloaded {
+        [&](TextIngestionMethod _) -> ContentAddressWithReferences {
+            if (refs.self)
+                throw UsageError("Cannot have a self reference with text hashing scheme");
```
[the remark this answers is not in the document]
Ericson2314: It looks like it can happen with a malformed computed derivation containing a self-reference.

How would you like me to extend the message?

src/libstore/content-address.hh
```diff
@@ -244,10 +272,28 @@ struct ContentAddressWithReferences
     { }
 
     /**
-     * Create a ContentAddressWithReferences from a mere ContentAddress, by
-     * assuming no references in all cases.
+     * Create a `ContentAddressWithReferences` from a mere
+     * `ContentAddress`, by assuming no references in all cases.
      */
     static ContentAddressWithReferences withoutRefs(const ContentAddress &);
```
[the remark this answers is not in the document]
Ericson2314: "Claim" sounds good to me

src/libstore/content-address.hh
roberth: Ok, makes sense.

src/libstore/content-address.cc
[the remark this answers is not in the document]
Ericson2314: Rethrowing when defining variables sucks, so I just used `std::optional`. Ahhh, much better.

src/libstore/misc.cc
Ericson2314: The method name predated this PR, so I am a bit wary of changing it.

However, I did change its type, so maybe I am just being silly.

roberth: Just a suggestion. Resolving.

src/libstore/build/local-derivation-goal.cc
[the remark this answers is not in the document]
roberth: This is a sensible restriction. If someone needs it removed, we can rely on them to start a conversation and we'll discover the best end state. Resolving.

roberth (approved):

edolstra:
This causes a warning:
```
src/libstore/content-address.cc: In static member function ‘static nix::ContentAddressMethod nix::ContentAddressMethod::parsePrefix(std::string_view&)’:
src/libstore/content-address.cc:41:16: warning: ‘method’ may be used uninitialized [-Wmaybe-uninitialized]
   41 |         method = TextIngestionMethod {};
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
src/libstore/content-address.cc:37:26: note: ‘method’ declared here
   37 |     ContentAddressMethod method = FileIngestionMethod::Flat;
      |                          ^~~~~~
```
Any idea what causes that?

PR #3961, opened 2020-08-26
Add a settings option for log-format

bqv:
This removes the need to pass --log-format to every nix invocation for your log-format of choice


src/libmain/loggers.cc
```diff
@@ -21,7 +22,17 @@ LogFormat parseLogFormat(const std::string & logFormatStr) {
 }
 
 Logger * makeDefaultLogger() {
-    switch (defaultLogFormat) {
+    auto logFormat = defaultLogFormat;
+    std::string settingsLogFormat = settings.logFormat.get();
+    if (settingsLogFormat != "")
```
regnat, line 27: If I'm not mistaken, this means that the `log-format` option takes precedence over the `log-format`.

You can probably get around that by changing the handler of the [log-format](https://github.com/NixOS/nix/blob/master/src/libmain/common-args.cc#L52-L58) option to `settings.set("log-format", format)` like it's done for example for [max-jobs](https://github.com/NixOS/nix/blob/master/src/libmain/common-args.cc#L60-L68)

bqv: Quite right! I've done that.

edolstra:
This appears to break the `--log-format` command line option, for example
```
nix eval --expr '1 + 2'  --json --log-format internal-json -vvvv
```
no longer prints any JSON messages.

src/libmain/loggers.cc
```diff
@@ -21,7 +22,17 @@ LogFormat parseLogFormat(const std::string & logFormatStr) {
...
+    {
+        try {
+            logFormat = parseLogFormat(settingsLogFormat);
+        } catch (Error) {
```
edolstra, line 31: This should be `Error &`. But I don't think we should catch this error at all. We shouldn't silently ignore bad values.

src/libmain/common-args.cc
```diff
@@ -54,7 +54,9 @@ MixCommonArgs::MixCommonArgs(const string & programName)
         .description = "format of log output; \"raw\", \"internal-json\", \"bar\" "
                         "or \"bar-with-logs\"",
         .labels = {"format"},
-        .handler = {[](std::string format) { setLogFormat(format); }},
+        .handler = {[=](std::string format) {
+            settings.set("log-format", format);
+        }},
```
edolstra, line 59: It should be possible to remove the `--log-format` option from `MixCommonArgs`, since any `nix.conf` option is automatically also a command line option.

src/libmain/loggers.cc
bqv: My thinking was that this would at least allow nix to run in a situation where the config is invalid, so user's systems don't get "stuck". I suppose there are easier ways of solving that though!

bqv:
> This appears to break the `--log-format` command line option, for example

I'm unsure why this is, as of now. I'll get back to this a little later

src/libmain/common-args.cc
Kloenk: But then it is hidden, and only seeable by --help-config

edolstra: That's a separate issue because it's true for all config options.

Kloenk: Not true for max-jobs?

src/libmain/loggers.cc
Kloenk: CMD line option overrides config file options. So --log-format should be able to fix that.

cole-h:
I tried debugging the issue where `--log-format` no longer works, but wasn't able to get anywhere due to my lack of familiarity with the `nix` codebase. All I figured out was that `settings.logFormat.get()` always returns the default value (an empty string).

Seeing that flags like `--extra-substituters` work, I imagine `settings.<name>.get()` is the proper way to get the value, but I don't know why the value of `--log-format` is not being set in the first place...

(BTW: `log-format` can be safely removed from `MixCommonArgs`, as it doesn't actually appear to do anything -- I put an `abort();` in the handler and nothing happened when passing `--log-format asdf`.)

src/libmain/common-args.cc
bqv: I kept it because I specialised Settings and removed the --option option instead

bqv:
It works!

cole-h:
Applying the following patch allows both `--option log-format` and `--log-format` to work (by moving the `createDefaultLogger()` call into the setter), as well as removes `--log-format` from `MixCommonArgs` and fixes the build (by fixing the `toJSON()` function): https://gist.github.com/cole-h/ec8e4334e6595c2848252e9dda54941d.

Kloenk:
So its ready to be merged? Would need this for one of my prs.
CC @edolstra

src/libstore/globals.cc
```diff
@@ -189,6 +190,52 @@ template<> void BaseSetting<SandboxMode>::convertToArg(Args & args, const std::s
...
+{
+    if (value == LogFormat::raw) return "raw";
+    else if (value == LogFormat::rawWithLogs) return "raw-with-logs";
+    else if (value == LogFormat::internalJson) return "internal-json";
+    else if (value == LogFormat::bar) return "bar";
+    else if (value == LogFormat::barWithLogs) return "bar-with-logs";
+    else abort();
+}
+
+template<> nlohmann::json BaseSetting<LogFormat>::toJSON()
```
regnat, line 220: This isn't correct anymore in master (`toJSON()` isn't expected to be equal to `toString()` anymore). You can get the templated definition by including `abstractsettingtojson.hh`

src/libstore/globals.cc
```diff
@@ -189,6 +190,52 @@ template<> void BaseSetting<SandboxMode>::convertToArg(Args & args, const std::s
...
+{
+    return AbstractSetting::toJSON();
+}
+
+template<> void BaseSetting<LogFormat>::convertToArg(Args & args, const std::string & category)
+{
+    args.addFlag({
```
regnat, line 227: There's a way to define custom completers for the cli, you could use that so that the autocompletion suggest the different log formats

regnat (changes requested):
> EDIT: Do note, however, that this means libstore depends on libmain (for access to createDefaultLogger()). I don't know if this is desirable
 or not (or even extremely undesirable), but I can't see any other way.                                                                       
                                                                                                                                              
That's a bit annoying imho. IIRC there was a good reason to keep the progress bar in libmain, but a good (?) way to break this would be to mak
e the logger pluggable like we do for the stores. Something like                                                                              
                                                                                                                                              
```cpp                                                                                                                                        
// Somewhere in libutil                                                                                                                       
struct LoggerBuilder {                                                                                                                        
  std::string name;                                                                                                                           
  std::function<(Logger*) void*> builder;                                                                                                     
};                                                                                                                                            
                                                                                                                                              
static std::vector<std::shared_ptr<LoggerBuilder>> registeredLoggers;                                                                         
                                                                                                                                              
void registerLogger(std::string name, std::function<(Logger*) void> builder) {                                                                
  LoggerBuilder lBuilder{ .name = name, .builder = builder };                                                                                 
  registeredLoggers.push_back(std::shared_ptr<LoggerBuilder>(lBuilder));
}

Logger* makeDefaultLogger() {
  for (auto & builder : registeredLoggers) {
    if (builder.name == settings.logFormat) {
      return builder.builder();
    }
  }
  throw Error("Unknown logger %s", settings.logFormat);
}

// Can be called in the file defining the logger
registerLogger("bar", []() -> Logger* { makeProgressBar(false); });
registerLogger("bar-with-logs", []() -> Logger* { makeProgressBar(true); });
registerLogger("json", []() -> Logger* { makeJSONLogger(); });
```

And then we can recover the list of values for the logger option by introspecting the value of `registeredLoggers`:

```cpp
std::string listLogFormats()
{
  std::string res;
  for (auto & builder : registeredLoggers) {
    if (res) res += ", ";
    res += builder.name;
  }
  return res;
}

    Setting<LogFormat> logFormat{this, LogFormat::bar, "log-format",
        format("Default build output logging format; %s.", listLogFormats()};
```

src/libstore/globals.cc
cole-h: FWIW, `abstractsettingtojson.hh` is already included by `globals.hh`, so to need to include it in `globals.cc` as well.

cole-h:
Thanks @regnat for the help! Following your advice (both here and on IRC), I refactored the loggers stuff to be pluggable (letting us drop the libmain <- libstore dependency).

There's one thing I'm still unhappy with, however: the fact that `man nix.conf` shows the default value as "3" (because it's an enum variant), instead of "bar". I don't know if there's a way to make whatever generates these defaults be run through a `to_string()` of some sort, but that would be swell. I've left a `FIXME` indicating this lack of satisfaction (input welcome).

One idea I just had would be to create a new log format type that's really just a string (as done with `Path` in libutil/types.hh), in order to override the default logic and also show the logger's string name in the manpage. Thoughts, anyone?

src/libstore/globals.cc
cole-h: I've done so (based on the completion function for `hashTypes`), but I don't know how to test if it really works (nor can I, AFAIK, as I use fish rather than bash or zsh).

src/libstore/globals.hh
```diff
@@ -12,6 +13,8 @@
 
 namespace nix {
 
+std::string listLogFormats();
```
cole-h, line 16: I don't really like that this is outside the `Settings` class, but putting it anywhere inside gives undefined references errors (due to its use in the `logFormat` description). Happy to hear alternatives.

regnat:
Thanks @cole-h , I like that a lot!

> There's one thing I'm still unhappy with, however: the fact that `man nix.conf` shows the default value as "3" (because it's an enum variant), instead of "bar". I don't know if there's a way to make whatever generates these defaults be run through a `to_string()` of some sort, but that would be swell. I've left a `FIXME` indicating this lack of satisfaction (input welcome).
> 
> One idea I just had would be to create a new log format type that's really just a string (as done with `Path` in libutil/types.hh), in order to override the default logic and also show the logger's string name in the manpage. Thoughts, anyone?

That sounds definitely a good thing (The static enum doesn't really make sense now that the thing is pluggable imho).

Regarding your IRC message:

> in the generated nix.conf.5 manpage, bar and bar-with-logs won't show up because they're declared outside of libutil

I think I found the reason why: both `settings` and `registeredLoggers` (and all the `registerLoggers` calls) are static variables. And in C++ the order in which they will be initialized isn't defined. This means that it's totally possible that `settings` will be instantiated while only part of the loggers have been registered, which in turn means that only this subset of loggers will show-up in the help (and the manual).
A way out of that is to replace the toplevel `Settings settings;` (in globals.hh) by a function `Settings* settings()` that will lazily initialize a `Settings` object and return a pointer to it (that way we're guaranteed that it will be instantiated *after* all the loggers have been registered). I've tried it in https://github.com/tweag/nix/commit/b52f6bee042c087668b812da2ce9462f20671ba3, and it definitely works, except that it's quite invasive because it requires rewriting all the accesses to `settings.foo` into `settings()->foo` (and there's a lot of these). So it might not be such a good idea.
Another option would be to allow making the description of a setting lazy (so that we only evaluate `listLogFormats()` once everything is registered), but I don't see how to do that nicely.

src/libstore/globals.hh
regnat: I think you're allowed to make it a static member of the `Settings` class if you want to

cole-h: With @LnL7's help, I created a new type for `LogFormatSetting`, which lets me remove `listLogFormats` from publicity and instead just call it in the constructor for that setting.

(Your `static` idea didn't seem to work.)

cole-h:
Because @LnL7 showed me how to get the json library to specialize enum conversion, the string wrapper is no longer necessary.

I also moved progress-bar.{hh,cc} and names.{hh,cc} to libutil, which allows all the loggers to be registered in libutil (so they all show up in the manpage).

I think this is ready for merge (or at least "final review"). Thanks everyone for your help (especially @regnat)!

regnat (approved):
Nice :+1:  I don't have the commit rights for the repo, but it looks all good on my side

(I'm not utterly fond of having both a static list of known loggers and a registration mechanism, but it's definitely not a blocker for me).

cole-h:
Ping @edolstra -- would appreciate another review (or, even better, a merge) on this, when you get a chance.

Ericson2314:
This is really cool! I'm especially excited because I think it might make it easier for hydra alternatives to stream logs per derivation?

Hope to see this land soon :).

bqv:
@edolstra ping, have you had a chance to look over this?

bqv:
Seems eelco was ignoring this because he has other plans https://github.com/NixOS/nix/commit/490861eae077851f15047addcac501846ec07503

bqv:
Close in favor of #4296

PR #3965, opened 2020-08-27
Pass all args when auto-calling a function with an ellipsis

### LINKED ISSUE #598: `--arg/--argstr` gets ignored when variables don't occur explicitly in the argument to the top-level function
```
[dfranke@pathfinder:~]$ nix-repl
Welcome to Nix version 1.9. Type :? for help.

nix-repl> #This behaves as expected

nix-repl> let f = { x ? "foo" }: x; g = y@{...}: f y; in g { x = "bar"; }
"bar"

nix-repl> :quit

[dfranke@pathfinder:~]$ cat test1.nix
let f = { x ? "foo", ... }: x; in y@{...}: f y

[dfranke@pathfinder:~]$ #This too should evaluate to "bar", not "foo"

[dfranke@pathfinder:~]$ nix-instantiate --eval --argstr x bar test1.nix
"foo"

[dfranke@pathfinder:~]$ #Now x is explicitly mentioned in the pattern

[dfranke@pathfinder:~]$ cat test2.nix
let f = { x ? "foo", ... }: x; in y@{x,...}: f y

[dfranke@pathfinder:~]$ #...and that's sufficient to produce a different result"

[dfranke@pathfinder:~]$ nix-instantiate --eval --argstr x bar test2.nix
"bar"

[dfranke@pathfinder:~]$ #Something's definitely wrong here!
```

glittershark:
The command line options --arg and --argstr that are used by a bunch of
CLI commands to pass arguments to top-level functions in files go
through the same code-path as auto-calling top-level functions with
their default arguments - this, however, was only passing the arguments
that were *explicitly* mentioned in the formals of the function - in the
case of an as-pattern with an ellipsis (eg args @ { ... }) extra passed
arguments would get omitted. This fixes that to instead pass *all*
specified auto args in the case that our function has an ellipsis.

Fixes #598

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1299,12 +1299,23 @@ void EvalState::autoCallFunction(Bindings & args, Value & fun, Value & res)
     Value * actualArgs = allocValue();
     mkAttrs(*actualArgs, fun.lambda.fun->formals->formals.size());
 
-    for (auto & i : fun.lambda.fun->formals->formals) {
-        Bindings::iterator j = args.find(i.name);
-        if (j != args.end())
-            actualArgs->attrs->push_back(*j);
-        else if (!i.def)
-            throwTypeError("cannot auto-call a function that has an argument without a default value ('%1%')", i.name);
+    if (fun.lambda.fun->formals->ellipsis) {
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks!

PR #3969, opened 2020-08-30
fallback to refreshing narinfo when nar is missing

### LINKED ISSUE #1885: hash mismatch in downloaded path
TLDR; keep all of your binary cache history or don't use cache for ookupOn one of the CI slaves I've seen:

```
hash mismatch in downloaded path '/nix/store/nmhvs4pihm56qnhpb6j09ajcwxjx9h07-digest-0.0.1.2': expected 8afe6105f80feb8f8b9e176bace053da40714598125b1201f91775fc9b1c22f4, got 772e334daa6beff675bee87adaaa2d38e45c0ccd15f065dad32ac88c22cd6637
```

Turns out what happened:

1. Hydra built `/nix/store/nmhvs4pihm56qnhpb6j09ajcwxjx9h07-digest-0.0.1.2`
2. Build slave substituted it and wrote the hash to `binary-cache-v3.sqlite`
3. Hydra GC-ed `/nix/store/nmhvs4pihm56qnhpb6j09ajcwxjx9h07-digest-0.0.1.2`
4. Months later someone wanted the same old package and Hydra rebuilt `/nix/store/nmhvs4pihm56qnhpb6j09ajcwxjx9h07-digest-0.0.1.2`
5. Build slave downloaded the new narinfo, but `binary-cache-v3.sqlite` had the old hash 

I presume this is due to binary non-determinism?

domenkozar:
Not sure this is the best way, but it does work :-D

Fixes #1885

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -320,7 +320,21 @@ void BinaryCacheStore::narFromPath(const StorePath & storePath, Sink & sink)
     try {
         getFile(info->url, *decompressor);
     } catch (NoSuchBinaryCacheFile & e) {
-        throw SubstituteGone(e.info());
+        // Nar may be missing because new narinfo is available
+        {
+            auto state_(state.lock());
+            state_->pathInfoCache.erase(std::string(info->path.hashPart()));
+        }
+        diskCache = getNarInfoDiskCache();
+        diskCache->deleteNarInfo(std::string(info->path.hashPart()));
+
+        auto info = queryPathInfo(storePath).cast<const NarInfo>();
+
+        try {
+            getFile(info->url, *decompressor);
+        } catch (NoSuchBinaryCacheFile & e) {
+            throw SubstituteGone(e.info());
+        }
     }
 
     decompressor->finish();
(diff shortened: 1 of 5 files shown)


edolstra:
Wouldn't this still fail in `copyStorePath()`, if `BinaryCacheStore::narFromPath()` returns a NAR with an unexpected hash? Deleting it from the database will only make it work on the next invocation.

domenkozar:
The use case here is that there's a GC on the binary cache meanwhile. So there won't be nar nor narinfo, while Nix will cache narinfo.

That's why we fallback to deleting the cache entry on 404 and retry.

The case where nar is corrupted is already being handled and if it's not corrupted it should just work?

domenkozar:
@edolstra does that clear up the motivation?

edolstra:
The motivation is clear. I was asking about how this works. How is it that `copyStorePath()` and `SubstitutionGoal` don't fail if `narFromPath()` returns a NAR with an unexpected hash?

domenkozar:
`NoSuchBinaryCacheFile` exception is raised inside `narFromPath` because the nar is missing in the binary cache.

That's when we refresh the narinfo and retry fetching the new nar (the optimization could be to check if nar url differs and not re-fetch it)

domenkozar:
Note that this would allow deleting nar files that are younger than 30 days, currently that results into Nix not recovering without deleting binary cache sqlite file.

domenkozar:
@edolstra I'm stuck here as I don't understand your questions.

PR #3972, opened 2020-08-31
progress-bar: multiple output lines

kloenk:
#3666 rebased onto master

CC: @edolstra @petabyteboy

### DIFF
--- a/src/libmain/progress-bar.cc
+++ b/src/libmain/progress-bar.cc
@@ -3,6 +3,7 @@
 #include "sync.hh"
 #include "store-api.hh"
 #include "names.hh"
+#include "globals.hh"
 
 #include <atomic>
 #include <map>
@@ -65,9 +66,12 @@ private:
 
         std::map<ActivityType, ActivitiesByType> activitiesByType;
 
+        int lastLines = 1;
+
         uint64_t filesLinked = 0, bytesLinked = 0;
 
         uint64_t corruptedPaths = 0, untrustedPaths = 0;
+
 
         bool active = true;
         bool haveUpdate = true;
@@ -81,12 +85,14 @@ private:
 
     bool printBuildLogs;
     bool isTTY;
+    bool printMultiline;
 
 public:
 
-    ProgressBar(bool printBuildLogs, bool isTTY)
+    ProgressBar(bool printBuildLogs, bool isTTY, bool printMultiline)
         : printBuildLogs(printBuildLogs)
         , isTTY(isTTY)
+        , printMultiline(printMultiline)
     {
         state_.lock()->active = isTTY;
         updateThread = std::thread([&]() {
@@ -94,7 +100,7 @@ public:
             while (state->active) {
                 if (!state->haveUpdate)
                     state.wait(updateCV);
-                draw(*state);
+                draw(*state, std::nullopt);
                 state.wait_for(quitCV, std::chrono::milliseconds(50));
             }
         });
@@ -140,8 +146,7 @@ public:
     void log(State & state, Verbosity lvl, const std::string & s)
     {
         if (state.active) {
-            writeToStderr("\r\e[K" + filterANSIEscapes(s, !isTTY) + ANSI_NORMAL "\n");
-            draw(state);
+            draw(state, std::optional(s));
         } else {
             auto s2 = s + ANSI_NORMAL "\n";
             if (!isTTY) s2 = filterANSIEscapes(s2, true);
@@ -325,12 +330,14 @@ public:
         updateCV.notify_one();
     }
 
-    void draw(State & state)
+    void draw(State & state, const std::optional<std::string> & s)
     {
         state.haveUpdate = false;
         if (!state.active) return;
 
         std::string line;
+        auto width = getWindowSize().second;
+        if (width <= 0) width = std::numeric_limits<decltype(width)>::max();
 
         std::string status = getStatus(state);
         if (!status.empty()) {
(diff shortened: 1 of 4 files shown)


edolstra:
This should probably be a different `--log-format` type (e.g. `fullscreen` or something). That way users can still use the single-line progress bar.

BTW what happens if there are more activities than the height of the terminal?

Kloenk:
> This should probably be a different `--log-format` type (e.g. `fullscreen` or something). That way users can still use the single-line progress bar.

Thas is currently addressed with the experimental feature `progress-bar`.

> BTW what happens if there are more activities than the height of the terminal?

Good Question

Kloenk:
> This should probably be a different `--log-format` type (e.g. `fullscreen` or something). That way users can still use the single-line progress bar.

If we make the option `log-format` should this be an option for nix.conf?

Kloenk:
@edolstra oh we already have log format. Didn't see that. The problem I have with that, It doesn't seems like I can set it via nix.conf. so I cannot default onto it in my own config

edolstra:
@Kloenk There is a PR for that: https://github.com/NixOS/nix/pull/3961

Kloenk:
> @Kloenk There is a PR for that: https://github.com/NixOS/nix/pull/3961

Ok, then I will wait for that, as it's makes it easier

src/libmain/progress-bar.cc
```diff
@@ -365,6 +368,62 @@ class ProgressBar : public Logger
...
+			if (!status.empty()) {
+				line += '[';
+				line += status;
+				line += "]";
+			}
+
+      #if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
+			  system("cls");
+			#else
+        writeToStderr("\r\033[K\r");
```
aszlig, line 390:
```suggestion
        writeToStderr("\r\033[K");
```

src/libmain/progress-bar.cc
```diff
@@ -365,6 +368,62 @@ class ProgressBar : public Logger
...
+	      for (auto i = 1; i < state.lastLines; i++)
+	        writeToStderr("\033[1A\r\033[K\r");
```
aszlig, line 392:
```suggestion
        writeToStderr(fmt("\033[%dA\r\033[K", state.lastLines - 1));
```

What about using [DECSTBM](https://www.vt100.net/docs/vt510-rm/DECSTBM.html) instead? This would also get rid of line tracking since all you need to do is set the scrolling region and push new lines.

For example like this (given that `lineStart` and `lineEnd` point to the absolute top and bottom margins):

```suggestion
	      writeToStderr(fmt("\033[s\033[%d;%dr\033[u", lineStart, lineEnd));
```

The `\e[s` and `\e[u` CSI are to save and restore cursor positions.

[Here is a small demo](https://gist.github.com/aszlig/f48026a6ec1f58efad84b41e336521d7) on how this could be even improved further, which also uses [DECSLRM](https://www.vt100.net/docs/vt510-rm/DECSLRM.html) if the terminal supports it (currently it's just checks if `TERM` starts with `xterm`) to set left margins. The latter could also be useful for constraining build outputs (eg. in conjunction with [DECOM](https://vt100.net/docs/vt510-rm/DECOM)) to the output of prefixed status like `building somedrv: some build output` and should make a lot of the CSI handling in ProgressBar obsolete.

Running this looks something like this. Note that I deliberately filled the terminal with stuff, so that behaviour is clear when we're reaching the bottom row of the terminal:

decstbm

The demo currently doesn't handle `SIGWINCH` and apart for the mentioned XTerm check it doesn't query terminal features. This is btw. something which I think we should probably do in Nix in general, since for example right now we just assume that SGR is *always* working on all terminals.

src/libmain/progress-bar.cc
```diff
@@ -365,6 +368,62 @@ class ProgressBar : public Logger
...
+			std::string status = getStatus(state);
+			if (!status.empty()) {
+				line += '[';
+				line += status;
+				line += "]";
+			}
+
+      #if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
+			  system("cls");
+			#else
```
aszlig, line 389: Since Nix uses spaces for indentation, please use spaces (instead of tabs) here as well to stay consistent.

src/libmain/progress-bar.cc
```diff
@@ -365,6 +368,62 @@ class ProgressBar : public Logger
...
+					}
+					state.lastLines += 1;
+					writeToStderr("\n" + filterANSIEscapes(line, false, width));
+				}
+			}
+		}
+
+		void draw(State & state)
+		{
+			if (settings.isExperimentalFeatureEnabled("progress-bar")) {
```
aszlig, line 422: What's the reason to hide this behind such a feature flag?

aszlig (approved):
LGTM :-) See the comments for my pesky nitpicks. Especially the DECSTBM/DECSLRM part is something that certainly could be done later and I might even give it a try to revamp the TUI in general for the [NixCon Hackday](https://2020.nixcon.org/#hackday) if you want to join.

src/libmain/progress-bar.cc
Kloenk: This should be changed once #3961 lands

src/libmain/progress-bar.cc
Kloenk: @aszlig  I don't know enough about that to implement it, so feel free to take this pr over. I also would be available on https://jitsi.nixcon.net/progress-bar at the moment

bqv:
c.f. #4296

Kloenk:
I updated it to a more generic version. @edolstra do you mind merging this, until your new log format is ready? I know some guys would like to use it.

Ericson2314:
The original was closed in lieu of #4296, which was never finished. Not sure what to do about this.

kloenk:
https://gerrit.lix.systems/c/lix/+/1369

PR #3981, opened 2020-09-01
Allow HTTP binary cache to request absolute uris

domenkozar:
@edolstra I'm not sure if this opens up any security concerns, but I'd like narinfo to point to nar url that doesn't match binary cache uri prefix.

My thinking is if you trust the binary cache you can also trust uri requests it returns.

### DIFF
--- a/src/libstore/http-binary-cache-store.cc
+++ b/src/libstore/http-binary-cache-store.cc
@@ -85,7 +85,7 @@ protected:
         checkEnabled();
 
         try {
-            FileTransferRequest request(cacheUri + "/" + path);
+            FileTransferRequest request(makeRequest(path));
             request.head = true;
             getFileTransfer()->download(request);
             return true;
@@ -103,7 +103,7 @@ protected:
         std::shared_ptr<std::basic_iostream<char>> istream,
         const std::string & mimeType) override
     {
-        auto req = FileTransferRequest(cacheUri + "/" + path);
+        auto req = makeRequest(path);
         req.data = std::make_shared<string>(StreamToSourceAdapter(istream).drain());
         req.mimeType = mimeType;
         try {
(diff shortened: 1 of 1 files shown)


src/libstore/http-binary-cache-store.cc
```diff
@@ -115,7 +115,11 @@ class HttpBinaryCacheStore : public BinaryCacheStore
 
     FileTransferRequest makeRequest(const std::string & path)
     {
-        FileTransferRequest request(cacheUri + "/" + path);
+        auto uri = cacheUri + "/" + path;
+        if (hasPrefix(path, "https://") || hasPrefix(path, "http://") || hasPrefix(path, "file://")) {
+          auto uri = path;
+        }
```
domenkozar, line 121: I've tried `if .. else` here but declaring `auto uri` in each block resulted in a compiler error.

edolstra:
Yeah I don't think this is a major security concern since we're checking the NAR hash.

Kloenk:
I wouldn't say I trust my binary store. I only trust the signature from it.

src/libstore/http-binary-cache-store.cc
edolstra: This doesn't do anything since `auto uri` is scoped to the if-block. Do you mean `uri = path`?

src/libstore/http-binary-cache-store.cc
```diff
@@ -115,7 +115,11 @@ class HttpBinaryCacheStore : public BinaryCacheStore
...
+        FileTransferRequest request(uri);
```
edolstra, line 122: You can also do
```
return FileTransferRequest(
  hasPrefix(path, "https://") || hasPrefix(path, "http://") || hasPrefix(path, "file://")
  ? path
  :  cacheUri + "/" + path);

domenkozar:
@Kloenk if you use the signing mechanism then it doesn't matter where the files are coming from, you trust the person that  signed the contents. If you don't use signing, then you trust the binary cache. Either way it doesn't matter much where the content is coming from.

@edolstra ready for re-review.

PR #3982, opened 2020-09-01
Minor docs updates

cole-h:
After the new markdown docs were merged, the `Hacking` link broke.

Additionally, when I was trying to hack on Nix yesterday, I was confused as to why `make install` was trying to access `/usr`, whereas the Hacking page noted Nix would be installed to `./inst`. I had to look at https://hydra.nixos.org/build/126292991/download/1/manual/installation/building-source.html to realize I needed to pass `--prefix=$(pwd)/inst` when configuring for the Hacking page to be accurate. Thus, I've added the flag to the page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Information on additional installation methods is available on the [Nix download
 
 ## Building And Developing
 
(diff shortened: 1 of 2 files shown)


vcunat:
Well, I don't really feel entitled for push access anymore, but I think it I can fairly safely say this PR is an improvement.

Issue #3960, opened 2020-08-25
Installer updates .bash_profile instead of .bashrc


mPiotrek:
**Describe the bug**

After installation, the nix commands (like `nix-env`) do not work even after opening a new shell.
Since `.bash_profile` is sourced by login shells, anything added there won't be available in shells started from graphical environment. On the other hand, if the `.bashrc` was being modified the commands would be sourced appropriately.
Here is an example of a confused user.
https://github.com/NixOS/nix/issues/555#issuecomment-510480029

**Steps To Reproduce**

1. Install Nix
2. Open a new `bash` (in a graphical environment)
3. Try tu use Nix

**Expected behavior**

3. Commands should be available

**`nix-env --version` output**

`nix-env (Nix) 2.3.7`

**Additional context**

I've done all of it with `bash` but my default shell is `fish` so maybe I'm missing something here.

Jiehong:
I confirm this issue (discovered when trying to install nix from fish as well (and met https://github.com/NixOS/nix/issues/3980)).

mPiotrek:
It's been over a year and so now I know more about Linux, POSIX, general conventions and what these various dotfiles are supposed to be.

`.bashrc` isn't really the right place for this. The proper solution would be to source it from both `.bash_profile`, as well as `.xprofile` (and possibly `.zsh_profile`, etc.). To not duplicate the code, one would usually put it in `.profile` and source that from any other of these files.

tomberek:
Closing as stale. Re-open if needed.

PR #3983, opened 2020-09-02
bugfix: add nix profile to bashrc for users installing from non login shells (#3960)

Jiehong:


scripts/install-nix-from-closure.sh
```diff
@@ -202,18 +202,7 @@ added=
...
-            if ! grep -q "$p" "$fn"; then
-                echo "modifying $fn..." >&2
-                echo -e "\nif [ -e $p ]; then . $p; fi # added by Nix installer" >> "$fn"
-            fi
-            added=1
-            break
-        fi
-    done
-    for i in .zshenv .zshrc; do
+    for i in .bash_profile .bashrc .bash_login .profile .zshenv .zshrc; do
```
matthewbauer, line 205: Note that we only add the value to one of these (see break below).

Jiehong: ah, thanks!

Issue #3980, opened 2020-09-01
Installation works in fish shell, but then it's not usable


Jiehong:
Hi,

Installing nix from a fish shell (Archlinux), seems to work at first, but nix is then not usable.

**Steps To Reproduce**

1. Start an OS and use a fish shell (in my case: Archlinux)
2. Install nix with the official installation process: `curl -L https://nixos.org/nix/install | sh`
3. This will work and show the following message:

```
installing 'nix-2.3.7'
unpacking channels...

Installation finished!  To ensure that the necessary environment
variables are set, either log in again, or type

  . /home/vagrant/.nix-profile/etc/profile.d/nix.sh

in your shell.
```

4. Running `. /home/vagrant/.nix-profile/etc/profile.d/nix.sh` reveals it's not working, because fish is not sh. The following error message is given:

```
~/.nix-profile/etc/profile.d/nix.sh (line 6): Unsupported use of '='. In fish, please use 'set NIX_LINK $HOME/.nix-profile'.
from sourcing file ~/.nix-profile/etc/profile.d/nix.sh
        called on line 185 of file /usr/share/fish/config.fish
in function '.' with arguments '/home/user/.nix-profile/etc/profile.d/nix.sh'
source: Error while reading file “/home/user/.nix-profile/etc/profile.d/nix.sh”
```

**Expected behaviour**

The installation process should detect the default shell, and/or provide a profile for other shells (aka `nix.fish` inside `/home/user/.nix-profile/etc/profile.d/`)

domenkozar:
Reported in https://github.com/digitallyinduced/ihp/issues/819

Kineolyan:
Hello, I just wrote a translation that seems to work for fish. Do you want me to open a PR with this?

domenkozar:
Yup.

Kineolyan:
Hello everyone, I submitted PR #4689 for this topic. The code has been approved so far, but I was lacking experience on nix build system to properly write tests. If you want to chime in with your expertise, that would be great. Otherwise, I let the good people from nix project decide what to do with the PR.

danielb2:
I'd like to be able to use nix with fish. how's this coming ?

samlaf:
Same here. Would be nice to support fish.

danielb2:
There really isn't much to initialize to make it usable in fish, so I don't see why this ticket is open so long.

to boot, @Jiehong  has already provided the code. C'mon nix people ...

jordigarcl:
[...]

Issue #440, opened 2015-01-11
adding support for a nix.fish script along side $HOME/.nix-profile/etc/profile.d/nix.sh


cartazio:
hey all, 
to complement the nix.sh script, because I use fish as my root shell,
i had to write a nix.fish

https://gist.github.com/cartazio/21523fb72db0227611ac is that file

and inline here too

``` fish
#! /usr/local/bin/fish
if test  -n "$HOME" ;
    set -xg NIX_LINK "$HOME/.nix-profile"

    # Set the default profile.
    if not test -L "$NIX_LINK" ;
        echo "creating $NIX_LINK" >&2
        set -l _NIX_DEF_LINK /nix/var/nix/profiles/default
        /nix/store/cdybb3hbbxf6k84c165075y7vkv24vm2-coreutils-8.23/bin/ln -s "$_NIX_DEF_LINK" "$NIX_LINK"
    end

    set -xg PATH $NIX_LINK/bin $NIX_LINK/sbin $PATH

    # Subscribe the user to the Nixpkgs channel by default.
    if not test -e $HOME/.nix-channels ;
        echo "https://nixos.org/channels/nixpkgs-unstable nixpkgs" > $HOME/.nix-channels
    end

    # Append ~/.nix-defexpr/channels/nixpkgs to $NIX_PATH so that
    # <nixpkgs> paths work when the user has fetched the Nixpkgs
    # channel.
    # set -xg  NIX_PATH ${NIX_PATH:+$NIX_PATH:}nixpkgs=$HOME/.nix-defexpr/channels/nixpkgs
    set -xg  NIX_PATH $NIX_PATH $HOME/.nix-defexpr/channels/nixpkgs

    # Set $SSL_CERT_FILE so that Nixpkgs applications like curl work.
    if test  -e /etc/ssl/certs/ca-bundle.crt ;  # Fedora, NixOS
        set -xg SSL_CERT_FILE /etc/ssl/certs/ca-bundle.crt ;
    else if test -e /etc/ssl/certs/ca-certificates.crt ;  # Ubuntu, Debian
        set -xg SSL_CERT_FILE /etc/ssl/certs/ca-certificates.crt
    else if test -e "$NIX_LINK/etc/ca-bundle.crt" ;  # fall back to Nix profile
        set -xg SSL_CERT_FILE "$NIX_LINK/etc/ca-bundle.crt"
    end
end
```

How would I go about upstreaming this and (future) fish scripting support?
[...]

PR #3985, opened 2020-09-02
bugfix: add a fish shell file to correctly source vars (#3980)

Jiehong:
I don't understand how nix-profile.sh.in gets turned into nix-profile.sh, so this PR might be missing something.

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -247,3 +247,42 @@ variables are set, either log in again, or type
 in your shell.
 EOF
 fi
+
+added=
+p=$HOME/.nix-profile/etc/profile.d/nix.fish
+if [ -z "$NIX_INSTALLER_NO_MODIFY_PROFILE" ] && [ -f "$HOME/.config/fish/fish.config"]; then
+    # Make the fish shell source nix.fish during login.
+    for i in fish.config; do
+        fn="$HOME/.config/fish/$i"
+        if [ -w "$fn" ]; then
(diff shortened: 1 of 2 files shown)


scripts/install-nix-from-closure.sh
```diff
@@ -247,3 +247,42 @@ variables are set, either log in again, or type
 in your shell.
 EOF
 fi
+
+added=
+p=$HOME/.nix-profile/etc/profile.d/nix.fish
```
Jiehong, line 252: I'm assuming `nix-profile.fish.in` is converted to `nix.fish` somehow, but that's unclear.

edolstra:
See #440. The main problem with adding fish support is that it will quickly get out of sync with the bash profile script, since the maintainers (e.g. me) don't use fish.

Jiehong:
> See #440. The main problem with adding fish support is that it will quickly get out of sync with the bash profile script, since the maintainers (e.g. me) don't use fish.

Ah, I didn't see that one. I just read it, and saw what people tried to do there.

I understand your concern here.

Perhaps an alternative would be to display a message if the user has fish installed, saying that nix will only work in bash?

cole-h:
Maybe we could suggest https://github.com/lilyball/nix-env.fish/. When I was using Nix on Arch, that plugin saved my sanity.

Jiehong:
This won't be merged, as the team won't maintain non-bash scripts.

Instead, I made a new PR to display a message instead: https://github.com/NixOS/nix/pull/4033

Issue #2033, opened 2018-04-01
Default Nix installations don't include user channels in NIX_PATH, causing inconsistent tool behavior


grahamc:
Should Nix installations include `$HOME/.nix-defexpr/channels` in the NIX_PATH? Should it appear before or after root channels?

```$ echo $NIX_PATH
nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixos/nixpkgs:nixos-config=/etc/nixos/configuration.nix:/nix/var/nix/profiles/per-user/root/channels

$ nix-channel --list             
nixpkgsunstable https://nixos.org/channels/nixpkgs-unstable

$ nix-env -iA nixpkgsunstable.git
installing 'git-minimal-2.16.2'

$ nix-shell -p nixpkgsunstable.git
error: undefined variable 'nixpkgsunstable' at (string):1:94
(use '--show-trace' to show detailed location information)

$ nix-shell '<nixpkgsunstable>' -p git
error: file 'nixpkgsunstable' was not found in the Nix search path (add it using $NIX_PATH or -I), at (string):1:94
(use '--show-trace' to show detailed location information)

$ nix-build '<nixpkgsunstable>' -A git
error: file 'nixpkgsunstable' was not found in the Nix search path (add it using $NIX_PATH or -I)

```

vs:

```
$ export NIX_PATH=$NIX_PATH:$HOME/.nix-defexpr/channels

$ nix-env -iA nixpkgsunstable.git    
replacing old 'git-minimal-2.16.2'
installing 'git-minimal-2.16.2'

$ nix-shell -p nixpkgsunstable.git    
error: undefined variable 'nixpkgsunstable' at (string):1:94
(use '--show-trace' to show detailed location information)

$ nix-shell '<nixpkgsunstable>' -p git

[nix-shell:~]$ exit

$ nix-build '<nixpkgsunstable>' -A git
/nix/store/jaani1kx4a4kvz11d0myfkjjiiklfl7w-git-2.16.2
```

With this configuration, and the explicitly specified nixpkgs, is it possible for users to override `nixpkgs`?

shlevy:
Yes, before root IMO.

zimbatm:
Do you know why `nix-env` and `nix-shell` are not using the same lookup method?

For everything to be perfectly aligned, `nix-channel --list` should probably report all the channels on the NIX_PATH instead of using that fixed location.

shlevy:
[...]

PR #3996, opened 2020-09-08
Macos big sur installer fixes

### LINKED ISSUE #3852: macOS version check broken in Big Sur beta 3.
Big Sur beta 3 changes the version from 10.15.6 (or 10.16? Whatever it was yesterday. :)) to 11.0.

looks like it should be something like:
```
# macOS support for 10.12.6 or higher
if [ "$(uname -s)" = "Darwin" ]; then
    macos_major=$(sw_vers -productVersion | cut -d '.' -f 1)
    macos_minor=$(sw_vers -productVersion | cut -d '.' -f 2)
    macos_patch=$(sw_vers -productVersion | cut -d '.' -f 3)
    if [ "$macos_major" -lt 10 ] || { [ "$macos_major" -eq 10 ] && [ "$macos_minor" -lt 12 ]; } || { [ "$macos_major" -eq 10 ] && [ "$macos_minor" -eq 12 ] && [ "$macos_patch" -lt 6 ]; }; then
        echo "$0: macOS $(sw_vers -productVersion) is not supported, upgrade to 10.12.6 or higher"
        exit 1
    fi
fi
```

### LINKED ISSUE #3957: investigate fixes for macOS Big Sur installer if `apfs.util -B` is indeed removed
I want to go ahead and put a pin in a [Big Sur installation issue I saw on Discourse](https://discourse.nixos.org/t/installing-on-macos-11-big-sur/8739):

> I can’t get the firmlinks thing working. Because (1) apfs.util doesn’t have the `-B` anymore, which I resolved by just doing a reboot.

IIRC, someone discovered `apfs.util -B` (`apfs.util --help` describes the purpose of this option as "bootstrap the root filesystem") as a way to skirt what had previously appeared to be a required reboot between adding /nix to synthetic.conf and being able to actually make the /nix directory. Here it is in the installer:

https://github.com/NixOS/nix/blob/c129e7c8f4846fcc146cfb3e408a75dd73794794/scripts/create-darwin-volume.sh#L135

If this is indeed true and the functionality hasn't moved elsewhere (or isn't a bug; or isn't something we can get restored by griping), I think our install process is going to necessarily become a clunky two-phase thing?

abathur:
Goal: make sure the installer scripts will run across macOS versions

Caveats, in case this ends up like the last one... 😬 
- I have tested installs with multiple macOS versions on a single system, but not on Linux.
- This work is just focused on procedural differences necessary to install on Big Sur as seen on existing Intel hardware; it has nothing directly to do with the ARM/DTK effort. (I gather the install will fail on a DTK without a release tarball available for that architecture.)

This will also fix #3852 and fix #3957.

A few notes (repeated from a later comment):
1. I had to fix a few macOS install breaks introduced by other commits already in master. If a new release needs to be cut from master before these changes land, those fixes will need to be cherry-picked/ported to keep macOS installs (for _all_ versions) working correctly.
2. After daemon installs, commands throw `file 'nixpkgs' was not found in the Nix search path...` errors on invocation. I don't think I caused this (maybe ec9dd9a did)? If that's correct, it'll need attention outside the scope of this PR.

### DIFF
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -26,6 +26,20 @@ test_synthetic_conf() {
     grep -q "^nix$" /etc/synthetic.conf 2>/dev/null
 }
 
+# Create the paths defined in synthetic.conf, saving us a reboot.
+create_synthetic_objects(){
+    # Big Sur takes away the -B flag we were using and replaces it
+    # with a -t flag that appears to do the same thing (but they
+    # don't behave exactly the same way in terms of return values).
+    # This feels a little dirty, but as far as I can tell the
+    # simplest way to get the right one is to just throw away stderr
+    # and call both... :]
+    {
+        /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t || true # Big Sur
+        /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B || true # Catalina
+    } >/dev/null 2>&1
+}
+
 test_nix() {
     test -d "/nix"
 }
@@ -101,7 +115,7 @@ main() {
 
     if ! test_nix; then
         echo "Creating mountpoint for /nix..." >&2
-        /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B || true
+        create_synthetic_objects # the ones we defined in synthetic.conf
         if ! test_nix; then
             sudo mkdir -p /nix 2>/dev/null || true
         fi
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -5,42 +5,13 @@ root_disk() {
     diskutil info -plist /
 }
 
-apfs_volumes_for() {
-    disk=$1
-    diskutil apfs list -plist "$disk"
-}
-
-disk_identifier() {
-    xpath "/plist/dict/key[text()='ParentWholeDisk']/following-sibling::string[1]/text()" 2>/dev/null
-}
-
-volume_list_true() {
-    key=$1
-    xpath "/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='$key']/following-sibling::true[1]" 2> /dev/null
-}
-
-volume_get_string() {
-    key=$1 i=$2
-    xpath "/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict[$i]/key[text()='$key']/following-sibling::string[1]/text()" 2> /dev/null
+# i.e., "disk1"
+root_disk_identifier() {
+    diskutil info -plist / | xmllint --xpath "/plist/dict/key[text()='ParentWholeDisk']/following-sibling::string[1]/text()" -
 }
 
 find_nix_volume() {
-    disk=$1
-    i=1
-    volumes=$(apfs_volumes_for "$disk")
-    while true; do
-        name=$(echo "$volumes" | volume_get_string "Name" "$i")
-        if [ -z "$name" ]; then
-            break
-        fi
-        case "$name" in
-            [Nn]ix*)
-                echo "$name"
-                break
-                ;;
-        esac
-        i=$((i+1))
-    done
+    diskutil apfs list -plist "$1" | xmllint --xpath "(/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[starts-with(translate(text(),'N','n'),'nix')]/text())[1]" - 2>/dev/null || true
 }
 
 test_fstab() {
@@ -89,9 +60,7 @@ test_t2_chip_present(){
 }
 
 test_filevault_in_use() {
-    disk=$1
-    #      list vols on disk | get value of Filevault key | value is true
-    apfs_volumes_for "$disk" | volume_list_true FileVault | grep -q true
+    fdesetup isactive >/dev/null
 }
 
 # use after error msg for conditions we don't understand
@@ -143,12 +112,12 @@ main() {
         fi
     fi
 
-    disk=$(root_disk | disk_identifier)
+    disk="$(root_disk_identifier)"
     volume=$(find_nix_volume "$disk")
     if [ -z "$volume" ]; then
         echo "Creating a Nix Store volume..." >&2
 
-        if test_filevault_in_use "$disk"; then
+        if test_filevault_in_use; then
             # TODO: Not sure if it's in-scope now, but `diskutil apfs list`
             # shows both filevault and encrypted at rest status, and it
             # may be the more semantic way to test for this? It'll show
@@ -178,6 +147,7 @@ main() {
     if ! test_fstab; then
         echo "Configuring /etc/fstab..." >&2
         label=$(echo "$volume" | sed 's/ /\\040/g')
+        # shellcheck disable=SC2209
         printf "\$a\nLABEL=%s /nix apfs rw,nobrowse\n.\nwq\n" "$label" | EDITOR=ed sudo vifs
     fi
 }
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -110,8 +110,8 @@ if [ "$(uname -s)" = "Darwin" ]; then
         "$self/create-darwin-volume.sh"
     fi
 
-    info=$(diskutil info -plist / | xpath "/plist/dict/key[text()='Writable']/following-sibling::true[1]" 2> /dev/null)
-    if ! [ -e $dest ] && [ -n "$info" ] && { [ "$macos_major" -gt 10 ] || { [ "$macos_major" -eq 10 ] && [ "$macos_minor" -gt 14 ]; }; }; then
+    writable="$(diskutil info -plist / | xmllint --xpath "name(/plist/dict/key[text()='Writable']/following-sibling::*[1])" -)"
+    if ! [ -e $dest ] && [ "$writable" = "false" ]; then
         (
             echo ""
             echo "Installing on macOS >=10.15 requires relocating the store to an apfs volume."
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -27,6 +27,15 @@ if [ "$(uname -s)" = "Darwin" ]; then
     IFS='.' read macos_major macos_minor macos_patch << EOF
 $(sw_vers -productVersion)
 EOF
+    # TODO: this is a temporary speed-bump to keep people from naively installing Nix
+    # on macOS Big Sur (11.0+, 10.16+) until nixpkgs updates are ready for them.
+    # *Ideally* this is gone before next Nix release. If you're intentionally working on
+    # Nix + Big Sur, just comment out this block and be on your way :)
+    if [ "$macos_major" -gt 10 ] || { [ "$macos_major" -eq 10 ] && [ "$macos_minor" -gt 15 ]; }; then
+        echo "$0: nixpkgs isn't quite ready to support macOS $(sw_vers -productVersion) yet"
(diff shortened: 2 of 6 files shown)


domenkozar:
cc @LnL7

scripts/create-darwin-volume.sh
```diff
@@ -89,9 +73,7 @@ test_t2_chip_present(){
 }
 
 test_filevault_in_use() {
-    disk=$1
-    #      list vols on disk | get value of Filevault key | value is true
-    apfs_volumes_for "$disk" | volume_list_true FileVault | grep -q true
+    fdesetup isactive
```
matthewbauer, line 76: This command doesn't exist on macOS catalina. Could we keep `volume_list_true` to continue supporting it?

abathur: We'll obviously keep whatever works, but can you double-check? I have it on Catalina (and Mojave, and Big Sur) at `/usr/bin/fdesetup`:
```
$ sw_vers -productVersion && fdesetup isactive
10.15.5
false
```

abathur: I also triggered a travis-ci build to check availability for this on a span of available versions from 10.12.6 through 10.15 at https://travis-ci.com/github/nix-community/nix-travis-ci/builds/183399921. These are "failing" because fdesetup isactive returns 1, but they all appear to have the command so far.

matthewbauer: Ahh sorry! Yeah I have it. I had been trying fdsetup :)

abathur: No worries. Happy to have you giving it a shake :)

scripts/install-nix-from-closure.sh
```diff
@@ -108,8 +109,13 @@ if [ "$(uname -s)" = "Darwin" ]; then
         "$self/create-darwin-volume.sh"
     fi
 
-    info=$(diskutil info -plist / | xpath "/plist/dict/key[text()='Writable']/following-sibling::true[1]" 2> /dev/null)
-    if ! [ -e $dest ] && [ -n "$info" ] && [ "$macos_major" -gt 14 ]; then
+    writable="$(diskutil info -plist / | xmllint --xpath "/plist/dict/key[text()='Writable']/following-sibling::*[1]/text()" -)"
+    # TODO: I wonder a bit if we need the version test, here. This may seem odd since I
```
LnL7, line 113: I think I suggested removing it, just checking if it's writable should be enough.

scripts/install-nix-from-closure.sh
```diff
@@ -24,10 +24,11 @@ fi
 
 # macOS support for 10.12.6 or higher
 if [ "$(uname -s)" = "Darwin" ]; then
-    macos_major=$(sw_vers -productVersion | cut -d '.' -f 2)
-    macos_minor=$(sw_vers -productVersion | cut -d '.' -f 3)
-    if [ "$macos_major" -lt 12 ] || { [ "$macos_major" -eq 12 ] && [ "$macos_minor" -lt 6 ]; }; then
-        echo "$0: macOS $(sw_vers -productVersion) is not supported, upgrade to 10.12.6 or higher"
+    verinfo="$(mktemp)"
```
LnL7, line 27: Should be pretty easy to avoid creating and leaving behind unnecessary tempfiles.

scripts/create-darwin-volume.sh
```diff
@@ -55,6 +26,19 @@ test_synthetic_conf() {
     grep -q "^nix$" /etc/synthetic.conf 2>/dev/null
 }
 
+# Create the paths defined in synthetic.conf, saving us a reboot.
+create_synthetic_objects(){
+    # Big Sur takes away the -B flag we were using and replaces it
```
LnL7, line 31: I had a feeling this might not be very stable, that's why I included the true fallback. I'd look at it as best effort so this seems fine. Perhaps even just always try both.

scripts/create-darwin-volume.sh
```diff
@@ -5,42 +5,13 @@ root_disk() {
...
-        fi
-        case "$name" in
-            [Nn]ix*)
-                echo "$name"
-                break
-                ;;
-        esac
-        i=$((i+1))
-    done
+    diskutil apfs list -plist "$1" | xmllint --xpath "/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[contains(translate(text(),'N','n'),'nix')]/text()" -
```
LnL7, line 14: Nice, did you check this with a few conditions?

scripts/create-darwin-volume.sh
LnL7: Does this need testing for T1/no T1 conditions or did you already do that?

LnL7 (commented):
Looks pretty good to me.

scripts/create-darwin-volume.sh
```diff
@@ -11,7 +11,7 @@ root_disk_identifier() {
 }
 
 find_nix_volume() {
-    diskutil apfs list -plist "$1" | xmllint --xpath "/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[contains(translate(text(),'N','n'),'nix')]/text()" -
+    diskutil apfs list -plist "$1" | xmllint --xpath "/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[contains(translate(text(),'N','n'),'nix')]/text()" - || true 2>/dev/null
```
cole-h, line 14: Should this be `- 2>/dev/null || true`, not `- || true 2>/dev/null`?

```suggestion
    diskutil apfs list -plist "$1" | xmllint --xpath "/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[contains(translate(text(),'N','n'),'nix')]/text()" - 2>/dev/null || true
```

abathur: Thanks :)

scripts/create-darwin-volume.sh
abathur: I've pushed a commit to do this. :)

scripts/install-nix-from-closure.sh
abathur: Addressed in 7ccb5b0

scripts/install-nix-from-closure.sh
abathur: Tentatively addressed in 30bd194, but leaving this open as a reminder to remove the comment once I've confirmed it works.

scripts/create-darwin-volume.sh
abathur: While thinking through this again I did notice and fix a pair of issues:
- after 16d6181 it should break after the first match like the existing loop
- after 4b42097 it should match at the beginning of the volume name only

Does that cover what you have in mind?

abathur:
I think this now achieves the main goal--making sure the installer scripts will run across macOS versions.

A few notes:
1. I had to fix a few macOS install breaks introduced by other commits already in master. If a new release needs to be cut from master before these changes land, those fixes will need to be cherry-picked/ported to keep macOS installs (for _all_ versions) working correctly.
2. I've only tested this on macOS. 
3. After daemon installs, commands throw `file 'nixpkgs' was not found in the Nix search path...` errors on invocation. I don't think I caused this (maybe ec9dd9a did)? If that's correct, it'll need attention outside the scope of this PR.

abathur:
@domenkozar I think this is roughly ready. Notes:

- Despite adding support, 0baa6e7 also adds a temporary check to limit Big Sur installs in case Apple releases it before some issues related to supporting Big Sur in nixpkgs can be ironed out (requested by @LnL7).
- Changes here probably merit a broad installer test before the next release
- When I tested this on daemon installs, nix-commands throw `file 'nixpkgs' was not found in the Nix search path...` errors on invocation. I'm planning to open a separate issue for that unless you see something I've done that you think could cause it (I suspect ec9dd9a may be the source).

Atemu:
> When I tested this on daemon installs, nix-commands throw file `'nixpkgs' was not found in the Nix search path...` errors on invocation.

This already happened on my recent Catalina installation. Doing step 3. from [here](https://github.com/nix-community/home-manager#installation) fixes the issue.

See https://github.com/NixOS/nix/issues/2033 for more info.

domenkozar:
@abathur I'm a bit short on time to test this. However, if you generate the installer script, it should be trivial to test with github actions once 11.0 is available for testing.

abathur:
@domenkozar I have them for the last two commits. I'll ping you on IRC.

abathur:
If anyone wants to test this, I've hosted two darwin-x86_64 tarballs with matching stub install scripts (these scripts don't check the tarball hash).

1. To test the actual Big Sur install, target https://files.t-ravis.com/install-allowed (which will download nix-3.0pre20200914_4b42097-x86_64-darwin.tar.xz, built from 4b42097).
2. To test the install guards that exclude Big Sur systems until Nixpkgs support is ready, target https://files.t-ravis.com/install-blocked (which will download nix-3.0pre20200917_0baa6e7-x86_64-darwin.tar.xz, built from 0baa6e7).

For reference, I also re-ran some trivial CI jobs with these URLs. 
- "allowed" run: https://travis-ci.com/github/nix-community/nix-travis-ci/builds/188871929
- "blocked" run: https://travis-ci.com/github/nix-community/nix-travis-ci/builds/188873113

Notes:
- You shouldn't see any difference between the allowed/blocked installers here; the "blocked" installer should only affect Big Sur+, and travis-ci hasn't added Big Sur runners yet.
- The macOS jobs "fail" after successfully installing because there's no nixpkgs pre-specified on the daemon installs (single-user should be OK). I'm not sure if that behavior is correct, but I'm pretty sure it's out-of-scope here.
- All of the Linux jobs are canceled--there's no Linux installer.

abathur:
@domenkozar GitHub Actions added Big Sur preview runners today, so I've done comparison runs with the installers above:
- Normal run (default installer) with Big Sur in the OS matrix to document existing condition https://github.com/abathur/resholve/actions/runs/307640557
- Run with "install-allowed" https://github.com/abathur/resholve/actions/runs/307647998
   - install-nix-action runs clean
   - but my build does fail with a c compiler issue; this is roughly why @LnL7 asked me to block installs
- Run with "install-blocked" https://github.com/abathur/resholve/actions/runs/307655319
   - install fails with the message `nixpkgs isn't quite ready to support macOS 11.0 yet`

domenkozar:
Could we also run with the tbd branch?

abathur:
I've force-pushed a squashed history which collapses all of the incremental fix commits back down into:
- 4 commits fixing issues directly necessary for Big Sur, which can be cherry-picked to `2.3-maintenance`: 9c3dc9d 1f02b65 e736f8f fe80790
- 3 commits fixing issues problems already present in master: 3a8699a b719f68 c40bad4
- 1 commit which blocks install on Big Sur install, which we can eventually drop/revert depending on the status of the tbd/stdenv work: f289bdb

domenkozar:
Thanks!

domenkozar:
Cherry-picked 4 commits to 2.3-maintenance, we'll hopefully release 2.3.8 once nixpkgs is ready.

abathur:
@domenkozar I didn't notice that 2.3.8 has already been cut or I would've done this sooner, but I ran an install test just now on my laptop with Big Sur:

```
 - system: `"x86_64-darwin"`
 - host os: `Darwin 20.1.0, macOS 10.16`
 - multi-user?: `no`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.3.8`
 - channels(abathur): `"nixpkgs-21.03pre246624.cfed29bfcb2"`
 - nixpkgs: `/Users/abathur/.nix-defexpr/channels/nixpkgs`
```

domenkozar:
Nice!!

PR #4005, opened 2020-09-11
Fix auto argument passing for more auto arguments than formals

Infinisil:
The change in #3965 didn't account for when the number of auto arguments is bigger than the number of formal arguments. This causes the following:

    $ nix-instantiate --eval -E '{ ... }@args: args.foo' --argstr foo foo
    nix-instantiate: src/libexpr/attr-set.hh:55: void nix::Bindings::push_back(const nix::Attr&): Assertion `size_ < capacity_' failed.
    Aborted (core dumped)

This fixes it

Ping @glittershark

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1348,7 +1348,7 @@ void EvalState::autoCallFunction(Bindings & args, Value & fun, Value & res)
     }
 
     Value * actualArgs = allocValue();
-    mkAttrs(*actualArgs, fun.lambda.fun->formals->formals.size());
+    mkAttrs(*actualArgs, std::max(static_cast<uint32_t>(fun.lambda.fun->formals->formals.size()), args.size()));
 
     if (fun.lambda.fun->formals->ellipsis) {
         // If the formals have an ellipsis (eg the function accepts extra args) pass


src/libexpr/eval.cc
```diff
@@ -1348,7 +1348,7 @@ void EvalState::autoCallFunction(Bindings & args, Value & fun, Value & res)
     }
 
     Value * actualArgs = allocValue();
-    mkAttrs(*actualArgs, fun.lambda.fun->formals->formals.size());
+    mkAttrs(*actualArgs, std::max((uint32_t) fun.lambda.fun->formals->formals.size(), args.size()));
```
glittershark, line 1351: generally I'd push for a `static_cast` here over a C-style cast.

glittershark (changes requested):
LGTM, one code style nit.

src/libexpr/eval.cc
Infinisil: Thanks, did that, I literally have no clue about C++ :)

glittershark (approved):

ajs124 (approved):
As far as I can tell, this makes my hydra-evaluator not segfault, so that's nice.

PR #4006, opened 2020-09-11
docs+test: fix remaining installer downloads without -L

lf-:
There were a couple of instances in old release notes that still end up on the docs page, and in my view it would be useful to have only working commands on the page. I got confused by these when I first installed Nix.

Also, for some reason the Darwin test had the wrong command, so we should probably check if it is being run and checked regularly :)


zimbatm:
thanks!

PR #4009, opened 2020-09-13
Implement RemoteStore.addToStoreFromDump

roberth:
This implements `RemoteStore.addToStoreFromDump`, allowing `nix-daemon` to serve as a proxy for another `nix-daemon`. A possible use case for this is distrusting a container that runs as an otherwise trusted uid. The implementation was fairly straightforward.

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -393,6 +393,34 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         break;
     }
 
+    case wopAddToStoreCANar: {
+        HashType hashAlgo;
+        std::string baseName;
+        FileIngestionMethod method;
+        {
+            uint8_t recursive;
+            std::string hashAlgoRaw;
+            from >> baseName >> recursive >> hashAlgoRaw;
+            if (recursive > (uint8_t) FileIngestionMethod::Recursive)
+                throw Error("unsupported FileIngestionMethod with value of %i; you may need to upgrade nix-daemon", recursive);
+            method = FileIngestionMethod { recursive };
+            hashAlgo = parseHashType(hashAlgoRaw);
+        }
+
+        logger->startWork();
+        std::optional<StorePath> path;
+        {
+            FramedSource narSource(from);
+            path = store->addToStoreFromDump(narSource, baseName, method, hashAlgo);
+        }
+        logger->stopWork();
+
+        if (path /* always */) {
+            to << store->printStorePath(*path);
+        }
+        break;
+    }
+
     case wopAddTextToStore: {
         string suffix = readString(from);
         string s = readString(from);
@@ -748,59 +776,12 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
             info.ultimate = false;
 
         if (GET_PROTOCOL_MINOR(clientVersion) >= 23) {
-
-            struct FramedSource : Source
-            {
-                Source & from;
-                bool eof = false;
-                std::vector<unsigned char> pending;
-                size_t pos = 0;
-
-                FramedSource(Source & from) : from(from)
-                { }
-
-                ~FramedSource()
-                {
-                    if (!eof) {
-                        while (true) {
-                            auto n = readInt(from);
-                            if (!n) break;
-                            std::vector<unsigned char> data(n);
-                            from(data.data(), n);
-                        }
-                    }
-                }
-
-                size_t read(unsigned char * data, size_t len) override
-                {
-                    if (eof) throw EndOfFile("reached end of FramedSource");
-
-                    if (pos >= pending.size()) {
-                        size_t len = readInt(from);
-                        if (!len) {
-                            eof = true;
-                            return 0;
-                        }
-                        pending = std::vector<unsigned char>(len);
-                        pos = 0;
-                        from(pending.data(), len);
-                    }
-
-                    auto n = std::min(len, pending.size() - pos);
-                    memcpy(data, pending.data() + pos, n);
-                    pos += n;
-                    return n;
-                }
-            };
-
             logger->startWork();
-
             {
                 FramedSource source(from);
                 store->addToStore(info, source, (RepairFlag) repair,
                     dontCheckSigs ? NoCheckSigs : CheckSigs);
             }
-
             logger->stopWork();
         }
 
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -408,16 +408,15 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         }
 
         logger->startWork();
-        std::optional<StorePath> path;
-        {
+        // Destructor of FramedSource must run before stopWork.
+        StorePath path = [&](){
             FramedSource narSource(from);
-            path = store->addToStoreFromDump(narSource, baseName, method, hashAlgo);
-        }
+            return store->addToStoreFromDump(narSource, baseName, method, hashAlgo);
+        }();
         logger->stopWork();
 
-        if (path /* always */) {
-            to << store->printStorePath(*path);
-        }
+        to << store->printStorePath(path);
+
         break;
     }
 
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -2,7 +2,6 @@
 #include "monitor-fd.hh"
 #include "worker-protocol.hh"
 #include "store-api.hh"
-#include "local-store.hh"
 #include "finally.hh"
 #include "affinity.hh"
 #include "archive.hh"
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -572,46 +572,6 @@ void RemoteStore::addToStore(const ValidPathInfo & info, Source & source,
 }
 
 
-StorePath RemoteStore::addToStore(const string & name, const Path & _srcPath,
-    FileIngestionMethod method, HashType hashAlgo, PathFilter & filter, RepairFlag repair)
-{
-    if (repair) throw Error("repairing is not supported when building through the Nix daemon");
-
-    auto conn(getConnection());
-
-    Path srcPath(absPath(_srcPath));
-
-    conn->to
-        << wopAddToStore
-        << name
-        << ((hashAlgo == htSHA256 && method == FileIngestionMethod::Recursive) ? 0 : 1) /* backwards compatibility hack */
-        << (method == FileIngestionMethod::Recursive ? 1 : 0)
-        << printHashType(hashAlgo);
-
-    try {
-        conn->to.written = 0;
-        conn->to.warn = true;
-        connections->incCapacity();
-        {
-            Finally cleanup([&]() { connections->decCapacity(); });
-            dumpPath(srcPath, conn->to, filter);
-        }
-        conn->to.warn = false;
-        conn.processStderr();
-    } catch (SysError & e) {
-        /* Daemon closed while we were sending the path. Probably OOM
-           or I/O error. */
-        if (e.errNo == EPIPE)
-            try {
-                conn.processStderr();
-            } catch (EndOfFile & e) { }
-        throw;
-    }
-
-    return parseStorePath(readString(conn->from));
-}
-
-
 StorePath RemoteStore::addToStoreFromDump(Source & narSource, const string & name,
     FileIngestionMethod method, HashType hashAlgo, RepairFlag repair)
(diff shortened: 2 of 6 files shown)


edolstra:
Not sure we should add this. `addToStoreFromDump()` is really superfluous since we already have an `addToStore()` that takes a `Source`. And the `RemoteStore` implementation of the latter implements it using `FramedSink` / `FramedSource` to ensure that the connection is left in a known state if an exception occurs.

roberth:
> Not sure we should add this. `addToStoreFromDump()` is really superfluous since we already have an `addToStore()` that takes a `Source`. 

This is the only method in the `Store` interface that can take any nar source and return it as a content-addressable store path without knowing the hash in advance, which is good to keep around for performance.

This is more efficient than `virtual void addToStore(const ValidPathInfo & info, Source & narSource, RepairFlag repair = NoRepair, CheckSigsFlag checkSigs = CheckSigs)`, because in order to use that method we'd have to

 - drain the nar stream to compute the hash for the `ValidPathInfo`
 - restart and drain it again while sending it to the store with `addToStore`

Whereas with `addToStoreFromDump` we don't need to precompute the hash and therefore only need to produce the nar once.


> And the `RemoteStore` implementation of the latter implements it using `FramedSink` / `FramedSource` to ensure that the connection is left in a known state if an exception occurs.

That seems quite important. I'll check if there's more commonality that can be factored out.

Ericson2314 (approved):
Baring the `FramedSink`/`FramedSource` changes, looks good to me. Agreed we should not get rid of `addToStoreFromDump` because it's the only one that calculates the content address for you.

Ericson2314:
For similar reasons I made a `addToStoreFromDump` implementation for `BinaryCacheStore` in https://github.com/NixOS/nix/pull/3935.

Ericson2314:
Also, https://github.com/NixOS/nix/pull/3829/files too helps with chaining stores.

roberth:
I've added the new `woAddToStoreCANar`, which is a framed and cleaned up version of `wopAddToStore`.

TODO: I've overlooked its last usage in `StorePath RemoteStore::addToStore(const string & name, const Path & _srcPath,
    FileIngestionMethod method, HashType hashAlgo, PathFilter & filter, RepairFlag repair)` which might be possible to implement in terms of fromdump.

Ericson2314:
If we are making a new version of `wopAddToStore` we might consider some other changes.

The other end is s bit cumbersome because the protocol always dumps a NAR, but the `addToStoreFromDump` expects the format to correspond to the file ingestion method. We could either plain dump the file for flat, or we could add a second `FileIngestionMethod` parameter to `addToStoreFromDump` to indicate what what the dump is, as opposed to what sort of content-address we should use for the remote-side-computed narinfo.

roberth:
@Ericson2314 I've spent a long time reasoning the implications of using a plain dump and trying to implement it, but it doesn't seem to pay off. We'll introduce a lot of complexity in the form of compatibility logic around the daemon protocol and all it might ever save is a single `Source` wrapper that turns a NAR into a file (not too hard). That's an upper bound in the cost of leaving the new `wopAddToStoreCANar` as is. I'll save you the long rant, but it boils down to

 - NAR is a good common format that supports all store paths. This allows us to simplify the `Store` api further than has been done; a possibility we'd throw out by supporting multiple dump formats. (Specifically this is about `virtual` methods. The rest are essentially static utility functions despite their syntax)
 - I've had a thorough look at `addToStoreSlow` specifically and its problems are not caused by having a single dump format. It can be clarified by refactoring in various ways, including merging all `FileIngestionMethod` conditionals into one and using previously mentioned `Source` wrapper instead of its `Sink` counterpart.

I have to conclude that the refactoring isn't worth it. I suggest someone could do it if they really want to, and they can do it even after this PR. I don't think it's worth their time and effort though.

src/libstore/remote-store.cc
```diff
@@ -554,78 +556,9 @@ void RemoteStore::addToStore(const ValidPathInfo & info, Source & source,
...
-                    if (ex) {
-                        try {
-                            std::rethrow_exception(ex);
-                        } catch (...) {
-                            ignoreException();
-                        }
-                    }
-                }
+            conn.withFramedSink([&](Sink & sink) {
+              copyNAR(source, sink);
```
roberth, line 560: Refactored, behavior unchanged.

src/libstore/remote-store.cc
```diff
@@ -982,6 +931,83 @@ std::exception_ptr RemoteStore::Connection::processStderr(Sink * sink, Source *
     return nullptr;
 }
 
+
+struct FramedSink : nix::BufferedSink
```
roberth, line 935: Moved for reuse, unchanged.

src/libstore/remote-store.cc
```diff
@@ -982,6 +931,83 @@ std::exception_ptr RemoteStore::Connection::processStderr(Sink * sink, Source *
...
+            ex = nullptr;
+            std::rethrow_exception(ex2);
+        }
+        conn->to << len;
+        conn->to(data, len);
+    };
+};
+
+void
+ConnectionHandle::withFramedSink(std::function<void(Sink &sink)> fun) {
```
roberth, line 968: Extracted from `void RemoteStore::addToStore(const ValidPathInfo & info, Source & source, .....`, unchanged.

src/libstore/remote-store.hh
```diff
@@ -62,9 +62,9 @@ public:
     void addToStore(const ValidPathInfo & info, Source & nar,
         RepairFlag repair, CheckSigsFlag checkSigs) override;
 
-    StorePath addToStore(const string & name, const Path & srcPath,
+    StorePath addToStoreFromDump(Source & dump, const string & name,
         FileIngestionMethod method = FileIngestionMethod::Recursive, HashType hashAlgo = htSHA256,
-        PathFilter & filter = defaultPathFilter, RepairFlag repair = NoRepair) override;
+        RepairFlag repair = NoRepair) override;
```
roberth, line 67: GitHub may render this as a "change" but it's really a deletion and an addition.

Addition:
 - Providing a real implementation instead of throwing an exception realizes the goal of this PR, allowing daemons to be chained

Deletion:
 - `addToStore(... Path ...)` has a default implementation that delegates to `addToStoreFromDump` which is all we need

src/libutil/serialise.hh
```diff
@@ -398,4 +400,51 @@ struct StreamToSourceAdapter : Source
 };
 
 
+/* Like SizedSource, but guarantees that the whole frame is consumed after
+   destruction. This ensures that the original stream is in a known state. */
+struct FramedSource : Source
```
roberth, line 405: Moved, unchanged.

roberth (commented):
This is ready for review now.
I've added some comments to clarify the changes.

Just in case, I've done additional testing by building a fresh package including a `filterSource` source and running `nix-store --verify-path` on the derivation closure and output closure, using a chained nix-daemon where the final daemon is Nix 2.4 before this change, the intermediate daemon is from this PR and the client is either from this PR or from before.

src/libstore/daemon.cc
```diff
@@ -394,6 +393,34 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
...
+
+        logger->startWork();
+        std::optional<StorePath> path;
+        {
+            FramedSource narSource(from);
+            path = store->addToStoreFromDump(narSource, baseName, method, hashAlgo);
+        }
+        logger->stopWork();
+
+        if (path /* always */) {
```
edolstra, line 418: This test can be removed or replaced by `assert(path)`.

src/libstore/worker-protocol.hh
```diff
@@ -50,6 +50,7 @@ typedef enum {
     wopAddToStoreNar = 39,
     wopQueryMissing = 40,
     wopQueryDerivationOutputMap = 41,
+    wopAddToStoreCANar = 42, // since Nix 3.0
```
edolstra, line 53: Shouldn't this be called `wopAddToStoreFromDump`?

edolstra (approved):

src/libstore/daemon.cc
roberth: I've changed the scope to a lambda to allow returning path without the `optional` contraption.

src/libstore/worker-protocol.hh
roberth: I figured it's a sibling of `wopAddToStoreNar`, so I've added `CA` to distinguish it, being only capable of producing that kind of store path. `Dump` sounds a bit vague to me, whereas `Nar` is specific.

Ericson2314: Yeah `wopAddToStoreFromDump` sounds like it isn't always expected nar-formatted objects, which would be confusing.

Ericson2314:
@roberth 

> @Ericson2314 I've spent a long time reasoning the implications of using a plain dump and trying to implement it, but it doesn't seem to pay off.

That's fine, we've had trouble with it too. But remember the second alternative I proposed is fixing up `addToStoreFromDump` itself.

Your current deamon code is wrong because it doesn't yet jump through the hoops of using `RetrieveRegularNARSink` in the `Flat` case.

Could you either change the interface of `addToStoreFromDump` or use `RetrieveRegularNARSink` in the daemon code?

Ericson2314:
If you go with the `RetrieveRegularNARSink` route, perhaps my never-finished #3802 could be of some use?

roberth:
> Your current deamon code is wrong because it doesn't yet jump through the hoops of using RetrieveRegularNARSink in the Flat case.

I don't think that's the case (I don't think I've made a mistake with my testing..) because the protocol now uses the exact same format as `addToStoreFromDump`, which is always a NAR.

> That's fine, we've had trouble with it too. But remember the second alternative I proposed is fixing up addToStoreFromDump itself.

The only downside of having NAR as the format for `Flat` is to do with hashing, but this code doesn't need to do anything with hashing. The only reason I mentioned hashing is because I was also looking at the impact of changing the protocol not to use NAR for `Flat`. That impact seems rather small.

It's quite nice having to deal with only one dump format. Only `LocalStore` doesn't have that luxury, but that's solved with `RetrieveRegularNARSink`.

I'd like to keep a reasonable scope for this PR and not overhaul `LocalStore::addToStore*`.

roberth:
Ok, rebased and formatted, so should be mergeable.

@Ericson2314 Thanks for the example. I think I'm going to leave it as is because the new daemon implementation is already constant memory with `FramedSource`.

Ericson2314:
> format as `addToStoreFromDump`, which is always a NAR.

This is not true. `addToStoreFromDump` expects the "dump" source to match the file ingestion method, unless someone changed it in the past week or two.

Ericson2314:
Yes, that is still the case. See `LocalStore::addToStoreFromDump` and how it uses the `method` parameter. This is why `wopAddToStore` need both with `RetrieveRegularNARSink` at all. And remeber that while nar -> flat is constant space, flat -> nar isn't because the nar needs to know the length up front which the flat won't provide.

Maybe let's just get the preperatory change pulling out `FrameSource` landed first? Because that is a great thing we can get out of the way, but I imagine what I just wrote will unfortunately mean a bunch more time needs to be spent on this.

src/libstore/worker-protocol.hh
Ericson2314: I actually I wonder if we shouldn't just reuse `wopAddToStore` (by checking the daemon protocol) since this fully replaces it.

roberth (commented):
@Ericson2314 Thanks for pointing out a problem I wasn't going to find with testing.

I'll rename it to `wopAddToStoreCA` because it's not always a NAR.

I didn't assume depending on the server/client version was an option because it's not how compatibility was handled earlier, but it seems quite sensible to just replace `wopAddToStore`.

@edolstra I propose that we simplify the handling of these dumps by not parsing in "intermediate" methods unless strictly necessary. The only reason we had to do so is to find the end of the NAR, which can now be done with `FramedSource`, which separates the end of the NAR from the end of the stream that transports the frames. The legacy store protocol code and `nix-store --import` can still perform the NAR parsing that's required for their correctness, but it doesn't have to be part of the core.

Ericson2314:
@roberth Yeah it's very annoying it's not tested, but I think `nix add-to-store  --store ...` can reproduce it. We should get a test case with that.

roberth:
@Ericson2314 I've used `builtins.path { recursive = false; }` to test it. Also I've reimplemented it, so have a look at #4030.

PR #3727, opened 2020-06-21
WIP: IPFS Store --- contains #3714

Ericson2314:
This is a rewritten https://github.com/NixIPFS/nix/tree/ipfs_binary_cache_legacy (thanks for pioneering!) that

 - Adds support for upload
 - Represents the store graph natively with IPLD
 - At @edolstra's request, there's no using git hashes or extra content addressing beyond what Nix already does. That is on another branch for consideration after.

PR #4011, opened 2020-09-14
Document store options

### LINKED ISSUE #3963: Document all store parameters
We're now (on the markdown branch) generating the nix.conf documentation from the source. We should do the same for store parameters. Currently they already have description strings in the source, but they're not queryable. We should generate a manual section out of them and add nix subcommand to query them.

regnat:
Fix #3963 

This turned out to be much more convoluted than what I though, because the obvious solution (having a loop that instantiate one instance of each store to grab the available options) isn't really feasible − instantiating some stores requires doing some actual effectful (and potentially slow) things and it's hard to work around it. So instead I splitted the store hierarchy into two parallel class hierarchies, something like

```
+-------------------+     +----------------+     +-------+
|    SubSubStore    | --> |    SubStore    | --> | Store |
+-------------------+     +----------------+     +-------+
  |                         |                      |
  |                         |                      |
  v                         v                      v
+-------------------+     +----------------+     +-------------+     +--------+
| SubSubStoreConfig | --> | SubStoreConfig | --> | StoreConfig | --> | Config |
+-------------------+     +----------------+     +-------------+     +--------+
```

(where `A --> B` means that A is a subclass of B).

This complexifies things quite a bit (esp. because the diamond inheritance crazyness means that we have to use virtual inheritance everywhere) but allows to instantiate the config for a store independently of the store itself, allowing for a side-effect free and fast `nix describe-stores` command (being fast is I think really important as it's intended to run as part of the build)


Ericson2314:
I like this a lot! But should the vertical arrows in the diagram point in the other direction?

I think https://github.com/NixOS/nix/pull/3829/files might also help ever so slightly with this. As it removes some stuff which awkwardly overlapped with the store registration mechanism.

edolstra:
> Include the full nlohmann/json header in config.hh 

Doesn't this include compile times a lot?

regnat:
> > Include the full nlohmann/json header in config.hh
> 
> Doesn't this increase compile times a lot?

My intuition (based on the fact that Nix itself was building fine without that change) is that every .cc file that includes `config.hh` also includes `nlohmann/json`. I might be wrong (not sure how to check that) but if that's the case then compile times should remain exactly the same (except for the perl bindings, but I don't think that the increase here is significant)

regnat:
> I like this a lot! But should the vertical arrows in the diagram point in the other direction?

Nah I don't think so. `Store` is a subclass of `StoreConfig` (and likewise for the others)

> I think https://github.com/NixOS/nix/pull/3829/files might also help ever so slightly with this. As it removes some stuff which awkwardly overlapped with the store registration mechanism.

#3829 seems mostly orthogonal, or did I miss something?

regnat:
> > > Include the full nlohmann/json header in config.hh
> > 
> > 
> > Doesn't this increase compile times a lot?
> 
> My intuition (based on the fact that Nix itself was building fine without that change) is that every .cc file that includes `config.hh` also includes `nlohmann/json`. I might be wrong (not sure how to check that) but if that's the case then compile times should remain exactly the same (except for the perl bindings, but I don't think that the increase here is significant)

Dunno how reliable it is, but the CI times between [this branch](https://github.com/NixOS/nix/pull/4011/checks) and [master](https://github.com/NixOS/nix/runs/1117421767) are sensibly the same. I also tried to `nix-build` nix before and after 
f1730e0  and the duration was significantly the same

edolstra:
> My intuition (based on the fact that Nix itself was building fine without that change) is that every .cc file that includes config.hh also includes nlohmann/json. 

Maybe because `precompiled-headers.h` includes it. Try building with `make PRECOMPILE_HEADERS=0`.

Ericson2314:
> #3829 seems mostly orthogonal, or did I miss something?

It is mostly orthogonal. But the `getStoreType` it removed was a very incomplete way of also figuring out the sort of store without actually initializing anything. (It was very incomplete because it didn't interact with the store registration mechanism at all.)

So your PR effectively restores the functionality I removed, while my PR removes the inferior alternate way of doing things. Together, they mean the API is strictly better :) --- you can figure out what store it is without duplicate code or doing effects.

src/libstore/store-api.cc
```diff
@@ -1091,8 +1066,41 @@ static RegisterStoreImplementation regStore([](
...
+                auto store = implem.create(parsedUri.scheme, baseURI, params);
+                if (store) {
+                    store->init();
+                    store->warnUnknownSettings();
+                    return ref<Store>(store);
+                }
+            }
+        }
+    }
+    catch (BadURL) {
```
edolstra, line 1092:
```suggestion
    catch (BadURL &) {
```

edolstra:
The `describe-stores` command crashes:

```
$ nix describe-stores
nix: src/libutil/config.hh:202: virtual nix::AbstractSetting::~AbstractSetting(): Assertion `created == 123' failed.
Aborted (core dumped)
```

regnat:
> Maybe because precompiled-headers.h includes it. Try building with make PRECOMPILE_HEADERS=0.

That was it indeed!

I've moved the implementation of `BaseSetting<T>::toJSONObj()` to a separate header file that's only included where needed so that `config.hh` can depend only on `json_fwd` rather than on the full `json` header.

> The describe-stores command crashes:

Ekk. Fixed, I had forgotten to restore some casts in the s3 store (assumed that all the stores were somehow tested in `installcheck`, but obviously the s3 store is trickier to test…). I've also added a test that runs `nix describe-stores` to ensure that this won't happen again

regnat:
Rebased on top of master to fix a small conflict

Ericson2314:
@regnat how come we have `MyStore(Params & params)` but not `MyStore(std::string & url, Params & params)` as the default? Say there is no default for the rest of the URL after the schema?

regnat:
> @regnat how come we have `MyStore(Params & params)` and not just `MyStore(std::string & scheme, std::string & uri, Params & params)` as the default? Say there is no default for the rest of the URL?

There might be some not-so-sensible constructors here and there (I introduced a couple of these at some point and might have forgotten to clean them all up), but the interface expected by `RegisterStoreImplementation` is

```cpp
std::function<std::shared_ptr<Store> (const std::string & scheme, const std::string & uri, const Store::Params & params)> create;
```

so we always need to provide the full url when opening a store

Ericson2314:
OK. For the record, https://github.com/NixOS/nix/pull/3727/commits/cde9c15d17ee5e38f5fcd55e97573ae7e6a4ebb4 was my attempt. I might see if I can remove some constructors from `master` to rule that issue out.

Ericson2314:
Ah, I  just forgot to upcast `this` for the config constructor. Nevermind!

Issue #374, opened 2014-10-17
nix-store does not obey `NIX_IGNORE_SYMLINK_STORE`


rrnewton:
When `/nix` is a symlink into a network file system, the current installer fails.  Down inside its `install` script, this is what happens:

``` bash
+ /nix/store/clnpynyac3hx3a6z5lsy893p7b4rwnyf-nix-1.7/bin/nix-store --init
error: `/nix' is not a directory
```

It would be nice if `nix-store` would also obey the `NIX_IGNORE_SYMLINK_STORE` flag, and tolerate this symlink.

Or..... perhaps this is unintended, and the current behavior is just a bug.  Maybe it just should have followed the symlink when checking the file type.  For example:

``` bash
$ file /nix
/nix: symbolic link to `/nfs/wherever'
$ file /nix/
/nix/: directory
```

knedlsepp:
I have a symlink `/nix` to a different filesystem. This works fine if I install things in the binary cache. Building things locally always fails because of this impurity even though i set `NIX_IGNORE_SYMLINK_STORE`, as described in the docs.

rrnewton:
Ok, so does that amount o a failure to reproduce?  I'll give it a try again.

gilligan:
@rrnewton is this still a problem / still relevant or can this issue be closed?

tomberek:
Closing as stale. Re-open if needed.

Issue #3513, opened 2020-04-19
error: the path '/nix' is a symlink; this is not allowed for the Nix store and its parent directories


ebadiere:
I am running Mac OS Catalina 10.15.4.

I followed the instructions in this issue to get around the /nix issue in Catalina:
https://github.com/NixOS/nix/issues/2925#issuecomment-539570232

Now I am getting the error: the path '/nix' is a symlink; message.

```---- sudo execution ------------------------------------------------------------
I am executing:

    $ sudo /nix/store/aizhr07dljmlbf17wfrj40x3s0b5iv3d-nix-2.3.4/bin/nix-store --load-db

to load data for the first time in to the Nix Database

error: the path '/nix' is a symlink; this is not allowed for the Nix store and its parent directories

---- oh no! --------------------------------------------------------------------
Jeeze, something went wrong. If you can take all the output and open
an issue, we'd love to fix the problem so nobody else has this issue.

:(

We'd love to help if you need it.

If you can, open an issue at https://github.com/nixos/nix/issues

Or feel free to contact the team,
 - on IRC #nixos on irc.freenode.net
 - on twitter @nixos_org```

domenkozar:
Best to follow instructions later on, which use the create volume script. Symlink approach is not recommended anymore.

Issue #2926, opened 2019-06-06
Turn NIX_IGNORE_SYMLINK_STORE=1 into a nix.conf option


jwiegley:
It would be nice to permanently enable this through the system configuration.

matthewbauer:
Could we just allow symlinked stores everywhere? Not sure what the rationale for that warning is. On macOS at least, symlinks are used for things like this all the time (/tmp is symlinked to /private/tmp, /etc is symlinked to /private/etc).

edolstra:
The reason is that builders could canonicalize a path (e.g. do `realpath $dir`) and end up storing the symlink target in derivation outputs. Which is bad is the output is shared with other machines via a binary cache.

pseudometric:
That's a valid concern — but the "firmlinks" Apple uses for this don't have that problem.

pseudometric:
Oops! :/ Never mind, it does. I thought it wouldn't because `cd /nix; cd ..` does the right thing – but `realpath` still shows a different path, and in fact builds fail with `hash mismatch in fixed-output derivation`.

It should be possible to instead mount a new APFS volume at `/nix`, and avoid this problem.

pseudometric:
Links created with `synthetic.conf(5)` are just regular symlinks and have this problem; if I do that for Nix, I get the expected build problem: `hash mismatch in fixed-output derivation`. OS X now has ”firmlinks” which don’t have this problem — but there appears to be no way for users to add more. The system comes with a fixed set of them, listed in `/usr/share/firmlinks`; annoyingly, that includes `/sw` for Fink but not `/nix`. I tried booting to single-user mode and adding an entry for `/nix` to that file, but that did nothing; it appears to be just a conveniently accessible list of the firmlinks maintained by the system, initialized once at build time and not actually used for anything.

Perhaps making a synthetic directory `/nix` and mounting a new APFS filesystem there will work, though I haven’t found a way to mount somewhere outside `/Volumes`.

pseudometric:
Found a way:

- Create new APFS volume with Disk Utility (click the plus-sign above "Volume" at upper left).
- Find the volume UUID: `diskutil list` / `diskutil info`.
- Add a line reading `/nix` to `/etc/synthetic.conf`; this will result in an empty `/nix` directory on reboot, to be the mount point.
- Reboot to single-user ("recovery") mode: hold down command-R while rebooting.
[...]

PR #4022, opened 2020-09-16
Fix multi-user installation with a symlinked store

maljub01:
Installing Nix with a symlinked store is not recommended, but is nevertheless
supported according to the documentation.

However, even when setting `NIX_IGNORE_SYMLINK_STORE=1`, installation still fails
with the following error:

```
I am executing:

    $ sudo /nix/store/aizhr07dljmlbf17wfrj40x3s0b5iv3d-nix-2.3.4/bin/nix-store --load-db

to load data for the first time in to the Nix Database

error: the path '/nix' is a symlink; this is not allowed for the Nix store and its parent directories
```

This is because sudo does not preserve environment variables and so the environment variable
`NIX_IGNORE_SYMLINK_STORE` will always be undefined when evaluated by `nix-store`.

This change makes sudo preserve the relevant environment variable in order to make installation
work again.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -209,7 +209,7 @@ __sudo() {
 
     echo "I am executing:"
     echo ""
-    printf "    $ sudo %s\\n" "$cmd"
+    printf "    $ sudo --preserve-env=NIX_IGNORE_SYMLINK_STORE %s\\n" "$cmd"
     echo ""
     echo "$expl"
     echo ""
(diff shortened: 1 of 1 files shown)


maljub01:
This potentially fixes #374 and #3513.

Ericson2314 (approved):

maljub01:
Thank you for the review. However, although this does allow the installation to succeed, I've run into several issues allover the place where things failed because of the missing environment variable.

Given how many other things are broken and the brittleness of trying to pipe through the environment variable everywhere, I now prefer to close this in favor of implementing #2926.

PR #4035, opened 2020-09-18
libfetchers/github: allow `url` attribute

Ma27:
Since 108debef6f703aa3ca1f6d6c45428880624ec6f8 we allow a
`url`-attribute for the `github`-fetcher to fetch tarballs from
self-hosted `gitlab`/`github` instances.

However it's not used when defining e.g. a flake-input

    foobar = {
        type = "github";
        url = "gitlab.myserver";
        /* ... */
    }

and breaks with an evaluation-error:

    error: --- Error --------------------------------------nix
    unsupported input attribute 'url'

This patch allows flake-inputs to be fetched from self-hosted instances
as well.

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -8,9 +8,9 @@
 
 namespace nix::fetchers {
 
-// A github or gitlab url
-const static std::string urlRegexS = "[a-zA-Z0-9.]*"; // FIXME: check
-std::regex urlRegex(urlRegexS, std::regex::ECMAScript);
+// A github or gitlab host
+const static std::string hostRegexS = "[a-zA-Z0-9.]*"; // FIXME: check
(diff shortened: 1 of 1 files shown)


edolstra:
`url` seems wrong since `gitlab.server` is not a valid URL. Maybe it should be `server` or `host` or something like that?

Ma27:
@edolstra you're obviously right and I'm fine with a rename. I'd just like to point out that I used the name-decision that was originally taken in 108debef6f703aa3ca1f6d6c45428880624ec6f8.

I think `host` is better. If you agree with this I'll update my PR accordingly.

edolstra:
Ah, sorry about that. Yes please do!

Ma27:
@edolstra fixed. Please note that this needs to be fixed as well now in https://github.com/edolstra/flake-compat/blob/master/default.nix :)

Issue #2919, opened 2019-06-04
add --no-store-add flag to nix-prefetch-url


cdepillabout:
It would be nice if there was a command that output the hash of a URL passed on the command line.  This would be similar to `nix-prefetch-url`, but it wouldn't add the URL to the nix store.  This useful for times when the `/nix/store` can't be modified.  An example of this would be when running within `nix-build`.

I can think of two ways this could be implemented:

1. Add a `--no-store-add` flag to `nix-prefetch-url`.

    `nix-prefetch-url` could be called like the following:
    
    ```console
    $ nix-prefetch-url --no-store-add 'ftp://ftp.gnu.org/pub/gnu/hello/hello-2.10.tar.gz'
    0ssi1wpaf7plaswqqjwigppsg5fyh99vdlb9kzl7c9lng89ndq1i
    ```

    This would be exactly like running without `--no-store-add`, except the downloaded file is not added to `/nix/store`.

    The other `nix-prefetch-*` commands would then also have a similar `--no-store-add` flag added.

    I think this would be a relatively easy change to make to `nix-prefetch-url`.  The only worry I have is that the name "nix-*prefetch*-url" doesn't really make any sense when using the `--no-store-add` flag, since the URL isn't really being pre-fetched (and inserted into the nix store).

1. Add a new `nix-hash-url` command.

    `nix-hash-url` would be somewhere in the middle of `nix-hash` and `nix-prefetch-url`.  It would fetch the specified URL and hash it, but not actually put it in the nix store.

    It could be called like the following:

    ```console
    $ nix-hash-url 'ftp://ftp.gnu.org/pub/gnu/hello/hello-2.10.tar.gz'
    0ssi1wpaf7plaswqqjwigppsg5fyh99vdlb9kzl7c9lng89ndq1i
    ```

    This would be more of a substantial change, since it is adding a new command, but it could probably share a lot of code with `nix-prefetch-url`.

    There could also be other `nix-hash-*` commands, like `nix-hash-git`, `nix-hash-hg`.

--------------------------------------------------------------------------------

My opinion would be to go with 1, since it would be quicker to implement, and it more-or-less makes sense given what `nix-prefetch-url` currently does.

I'd be willing to send a PR adding this flag if it would be approved.

cdepillabout:
The motivation for this is https://github.com/NixOS/nixpkgs/pull/61067.

[...]

PR #4036, opened 2020-09-18
nix-prefetch-url: Add --executable flag

chreekat:
pkgs.fetchurl supports an executable argument, which is especially nice
when downloading a large executable. This patch adds the same option to
nix-prefetch-url.

I have tested this to work on the simple case of prefetching a little
executable:

1. nix-prefetch-url --executable https://my/little/script
2. Paste the hash into a pkgs.fetchurl-based package, script-pkg.nix
3. Delete the output from the store to avoid any misidentified artifacts
4. Realise the package script-pkg.nix

I repeated the above while using --name, as well.

I suspect --executable would have no meaningful effect if combined with
--unpack, but I have not tried it.

### DIFF
--- a/doc/manual/src/command-ref/nix-prefetch-url.md
+++ b/doc/manual/src/command-ref/nix-prefetch-url.md
@@ -51,6 +51,9 @@ Nix store is also printed.
     result to the Nix store. The resulting hash can be used with
     functions such as Nixpkgs’s `fetchzip` or `fetchFromGitHub`.
 
+  - `--executable`  
+    Set the executable bit on the downloaded file.
+
   - `--name` *name*  
     Override the name of the file in the Nix store. By default, this is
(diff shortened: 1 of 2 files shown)


chreekat:
Although I forgot to mention it in the commit, I also tested that I could execute the file. :)

chreekat:
Rebased onto latest master because I don't think the CI failure was related to my changes. Let's see.

chreekat:
According to [a comment by @edolstra](https://github.com/NixOS/nix/issues/2919#issuecomment-498667575), "We shouldn't add new flags to `nix-prefetch-url`. Any new functionality should be added to the nix command (e.g. as a nix prefetch subcommand)."

Unfortunately, adding an entire new subcommand to `nix`, which is still marked as experimental so I don't want to use it anyway, is probably not something I have time for right now.

edolstra:
Thanks!

PR #4038, opened 2020-09-18
Add a nix.conf option for allowing a symlinked store

### LINKED ISSUE #2926: Turn NIX_IGNORE_SYMLINK_STORE=1 into a nix.conf option

maljub01:
Using an environment variable for allowing a symlinked store is currently broken, and is extremely brittle (See #4022). This is because nix invocations that happen in a different environment than the user's will still return errors complaining about the symlink (ex. through `sudo` or the `nix-daemon`).

Fixes #2926.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -883,7 +883,7 @@ public:
         "Path or URI of the global flake registry."};
 
     Setting<bool> ignoreSymlinkStore{
-        this, false, "ignore-symlink-store",
+        this, false, "allow-symlinked-store",
         R"(
           If set to `true`, Nix will stop complaining if the store directory
           (typically /nix/store) contains symlink components.
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -882,7 +882,7 @@ public:
     Setting<std::string> flakeRegistry{this, "https://github.com/NixOS/flake-registry/raw/master/flake-registry.json", "flake-registry",
         "Path or URI of the global flake registry."};
 
-    Setting<bool> ignoreSymlinkStore{
+    Setting<bool> allowSymlinkedStore{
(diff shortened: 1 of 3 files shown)


src/libstore/globals.hh
```diff
@@ -881,6 +881,19 @@ public:
 
     Setting<std::string> flakeRegistry{this, "https://github.com/NixOS/flake-registry/raw/master/flake-registry.json", "flake-registry",
         "Path or URI of the global flake registry."};
+
+    Setting<bool> ignoreSymlinkStore{
+        this, false, "ignore-symlink-store",
```
maljub01, line 886: I'm using `ignore-symlink-store` here to be consistent with the environment variable. However, there are arguably better names for this config option. I'm happy to change it if you have any suggestions.

maljub01: For example, one possible option could be `allow-symlink-store`.

Ericson2314 (approved):
Looks good. In fact, i would like to have a policy that for every env var there exists a more proper configuration option (be it global, per store, etc.)

src/libstore/globals.hh
edolstra: :+1: on `allow-symlink-store` (or maybe `allow-symlinked-store`).

maljub01: Done.

maljub01:
> Looks good. In fact, i would like to have a policy that for every env var there exists a more proper configuration option (be it global, per store, etc.)

Agreed! What would the process be to try to make that happen?

edolstra:
In fact we should just get rid of the environment variable. The only reason it exists is that it was a hack (which probably shouldn't have been documented...) and we didn't have a config file back then.

Issue #4019, opened 2020-09-15
When updating to the newest JSON library version (beginning 3.8.0), nix fails to build


OmnipotentEntity:
**Describe the bug**

Nix fails to build with the following error message when using a version of nlohmann's json library that includes the binary type feature.

```
builder for '/nix/store/cd3a1sxaq9idl129bvgbgrazpfnpp792-nix-3.0pre20200829_f156513.drv' failed with exit code 2; last 10 log lines:
                   from src/libexpr/json-to-value.cc:4:
  /nix/store/a15crh4aa98283pmx0gxzchgi39547qf-nlohmann_json-3.9.1/include/nlohmann/detail/input/binary_reader.hpp:289:113: error: 'nlohmann::json_sax<nlohmann::basic_json<> >' is not an accessible base of 'nlohmann::detail::binary_reader<nlohmann::basic_json<>, nlohmann::detail::iterator_input_adapter<__gnu_cxx::__normal_iterator<const char*, std::__cxx11::basic_string<char> > >, nix::JSONSax>::json_sax_t' {aka 'nix::JSONSax'}
    289 |                 return get_number<std::int32_t, true>(input_format_t::bson, len) && get_bson_binary(len, value) && sax->binary(value);
    CXX    src/nix/repl.o
    CXX    src/libexpr/nixexpr.o
    CXX    src/nix/registry.o
    CXX    src/nix/profile.o
    CXX    src/nix/ping-store.o
  make: *** [mk/patterns.mk:3: src/libexpr/json-to-value.o] Error 1
  make: *** Waiting for unfinished jobs....
```

**Steps To Reproduce**

1. `nix run -I nixpkgs=channel:nixos-unstable nixpkgs.nixpkgs-review -c nixpkgs-review pr 97266`

**Expected behavior**

Builds

**`nix-env --version` output**

nix-env (Nix) 2.3.6

**Additional context**

This was caused by changes to the Sax class in version 3.8.0 to accommodate a new internal type called "binary" which handles binary input/output from binary filetypes (such as CBOR and BSON).  You'll need to implement a `binary` method in the `nix::JSONSax` class.  I do not think you are currently using binary types, so you can just implement a "do nothing" version safely.  There are examples of this in the nlohmann JSON tests.

matthewbauer:
I just had to add `bool binary(binary_t& val)` for this to work with 3.8.0. I bet more stuff changed with 3.9.0 though.

[...]

PR #4040, opened 2020-09-19
Fix compatibility with nlohmann-json 3.9.1

OmnipotentEntity:
This is intended to fix https://github.com/NixOS/nixpkgs/pull/97266 and https://github.com/NixOS/nix/issues/4019

However, this is my first time submitting a PR to this repo, and in this manner, so please be patient with me.  I'm willing to work at it until it becomes right.  The upside is it is a very small PR.

Because this issue is between an incompatibility between nixpkgs and nix, I overrode the json version here to demonstrate functionality.  However, simply having the `binary` function ought not cause any issues for json version 3.7.3.  Once this PR becomes ready to accept I can and expect to remove them.

### DIFF
--- a/src/libexpr/json-to-value.cc
+++ b/src/libexpr/json-to-value.cc
@@ -115,6 +115,14 @@ public:
     {
         return handle_value<void(Value&, const char*)>(mkString, val.c_str());
     }
+#if NLOHMANN_JSON_VERSION_MAJOR >= 3 && NLOHMANN_JSON_VERSION_MINOR >= 8
+    bool binary(binary_t&)
+    {
+        // This function ought to be unreachable
+        assert(false);
+        return true;
+    }
+#endif
 
     bool start_object(std::size_t len)
     {


Ericson2314:
This looks good, except we should find a way to reduplicate the override in the flake.nix.

cole-h:
@Ericson2314 I don't think the overrides are intended to be permanent -- they only exist to show that this fixes the build with nlohmann::json 3.9.1, since the 3.9.1 update has not been merged in Nixpkgs yet ("I overrode the json version here to demonstrate functionality").

cc @matthewbauer -- You suggested using a `throw` here; should this be done, or is a `return true;` fine?

src/libexpr/json-to-value.cc
```diff
@@ -116,6 +116,12 @@ class JSONSax : nlohmann::json_sax<json> {
         return handle_value<void(Value&, const char*)>(mkString, val.c_str());
     }
 
+    bool binary(binary_t &)
```
edolstra, line 119: An alternative to overriding nlohmann_json is to make it build with both old and new versions, e.g.
```
#if NLOHMANN_JSON_VERSION_MAJOR >= 3 && NLOHMANN_JSON_VERSION_MINOR >= 8
     bool binary(binary_t &) override
#endif
...
```

OmnipotentEntity: I'm reasonably sure that this code will compile under 3.7.x as well, because the existence of an additional, unused function in the class won't cause a compile error.

OmnipotentEntity: Strike that, of course it won't, `binary_t` doesn't exist.  I'll make this adjustment.  Thanks for the eyes.

Ericson2314:
Also, I'm thinking @matthewbauer is right we should through, otherwise we could silently drop a field or something.

OmnipotentEntity:
A binary value can only be generated from reading specific binary types,
such as BSON or CBOR. If you are only reading plain JSON then this is
properly unreachable code. I'm not familiar with how nix uses json though.

If this is the case would it be better to throw or assert(false)?

edolstra:
Yes, `assert(false)` sounds good.

OmnipotentEntity:
Alright, that's changed.  Again, the purpose of overriding nlohmann_json is merely to prove that it works under the newest json version.  I will be reverting that change, because nix shouldn't be tied to an explicit json version.

OmnipotentEntity:
Now that CI has had a chance to chew through it, I've pushed the version without the changes to flake.nix.

Ericson2314 (approved):

PR #4041, opened 2020-09-19
Serialize SandboxMode enum to string for JSON

cole-h:
Rather than showing an integer as the default, instead show the string
referenced in the description.

The nix.conf.5 manpage used to show "default: 0", which is unnecessarily
opaque and confusing (doesn't 0 mean false, even though the default is
true?); now it properly shows that the default is true.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -147,6 +147,12 @@ bool Settings::isWSL1()
 
 const string nixVersion = PACKAGE_VERSION;
 
+NLOHMANN_JSON_SERIALIZE_ENUM(SandboxMode, {
+    {SandboxMode::smEnabled, true},
+    {SandboxMode::smRelaxed, "relaxed"},
+    {SandboxMode::smDisabled, false},
+});
+
 template<> void BaseSetting<SandboxMode>::set(const std::string & str)
 {
     if (str == "true") value = smEnabled;


Ericson2314 (approved):
Good idea. Are there any back compat issues we need to worry about?

cole-h:
I wouldn't think so. AFAICT this only affects the output of `nix show-config --json` (which displays the description and is what's used to generate the manpage) -- `./inst/bin/nix show-config --json --experimental-features nix-command | jq '.["sandbox"].defaultValue'` now shows `"true"` rather than `0`.

I also thought about making `"true"` an actual boolean `true` (same with `"false"`) -- thoughts?

Ericson2314:
Ah great, so this is all `--experimental-features nix-command` stuff. Issue averted!

> I also thought about making "true" an actual boolean true (same with "false") -- thoughts?

Sounds good too!

cole-h:
OK, done :)

(Maybe this will help people who parse the output from having to convert from a string to a boolean.)

src/libstore/globals.hh
```diff
@@ -14,6 +14,12 @@ namespace nix {
 
 typedef enum { smEnabled, smRelaxed, smDisabled } SandboxMode;
 
+NLOHMANN_JSON_SERIALIZE_ENUM(SandboxMode, {
```
edolstra, line 17: Doesn't this require globals.hh to include nlohmann/json.hpp? So it wouldn't work with `PRECOMPILE_HEADERS=0`.

Ericson2314: I think it does because it passed tests? That header is included via `abstractsettingtojson.hh`

edolstra: Ouch, we need to revert that. It causes most compilation units to pull in nlohmann/json.hpp.

edolstra: Background: including that header adds about half a second per compilation unit.

cole-h: I've rebased this on master to incorporate your removal of `abstract-setting-to-json.hh` from `globals.hh` and moved the macro call to `globals.cc`, which I hope is satisfactory.

I tested locally that this still works for manpage-generation purposes.

PR #4055, opened 2020-09-23
Use gold as the linker on Linux

regnat:
Saves ~7s in the linking phase


Ericson2314:
I suppose this is fine for now, but if we ever try to use this flake from Nixpkgs we should probably have a way to make it less "opinionated".

PR #4056, opened 2020-09-23
Allow non-CA derivations to depend on CA ones

regnat:
Builds on top of #3958

There's a few things that I'm not really happy with wrt. the clarity of the code, but it would require a bit of refactoring so I'll probably address that in a subsequent PR in order not to delay this one

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -493,7 +493,8 @@ void DerivationGoal::inputsRealised()
     if (useDerivation) {
         auto & fullDrv = *dynamic_cast<Derivation *>(drv.get());
 
-        if (!fullDrv.inputDrvs.empty() && fullDrv.type() == DerivationType::CAFloating) {
+        if ((!fullDrv.inputDrvs.empty() &&
+             fullDrv.type() == DerivationType::CAFloating) || fullDrv.type() == DerivationType::DeferredInputAddressed) {
             /* We are be able to resolve this derivation based on the
                now-known results of dependencies. If so, we become a stub goal
                aliasing that resolved derivation goal */
@@ -3165,6 +3166,15 @@ void DerivationGoal::registerOutputs()
             },
             [&](DerivationOutputCAFloating dof) {
                 return newInfoFromCA(dof);
+            },
+                [&](DerivationOutputDeferred) {
+                // No derivation should reach that point without having been
+                // rewritten first
+                assert(false);
+                // Ugly, but the compiler insists on having this return a value
+                // of type `ValidPathInfo` despite the `assert(false)`, so
+                // let's provide it
+                return *(ValidPathInfo*)0;
             },
         }, output.output);
 
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -21,6 +21,9 @@ std::optional<StorePath> DerivationOutput::path(const Store & store, std::string
         [](DerivationOutputCAFloating dof) -> std::optional<StorePath> {
             return std::nullopt;
         },
+        [](DerivationOutputDeferred) -> std::optional<StorePath> {
+            return std::nullopt;
+        },
     }, output);
 }
 
@@ -37,6 +40,7 @@ bool derivationIsCA(DerivationType dt) {
     case DerivationType::InputAddressed: return false;
     case DerivationType::CAFixed: return true;
     case DerivationType::CAFloating: return true;
+    case DerivationType::DeferredInputAddressed: return false;
     };
     // Since enums can have non-variant values, but making a `default:` would
     // disable exhaustiveness warnings.
@@ -48,6 +52,7 @@ bool derivationIsFixed(DerivationType dt) {
     case DerivationType::InputAddressed: return false;
     case DerivationType::CAFixed: return true;
     case DerivationType::CAFloating: return false;
+    case DerivationType::DeferredInputAddressed: return false;
     };
     assert(false);
 }
@@ -57,6 +62,7 @@ bool derivationIsImpure(DerivationType dt) {
     case DerivationType::InputAddressed: return false;
     case DerivationType::CAFixed: return true;
     case DerivationType::CAFloating: return false;
+    case DerivationType::DeferredInputAddressed: return false;
     };
     assert(false);
 }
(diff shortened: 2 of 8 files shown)


Ericson2314:
@regnat So everything is the same but the commit at the end? Are you opposed to merging #3958 so it's easier to just review those? (Correct me if I am wrong, but this is mainly about adding the missing feature and doesn't change the way #3958 worked for the CA -> CA case too much?)

src/libstore/derivations.cc
```diff
@@ -491,12 +514,15 @@ DrvHashModulo hashDerivationModulo(Store & store, const Derivation & drv, bool m
     }
     case DerivationType::InputAddressed:
         break;
+    case DerivationType::DeferredInputAddressed:
+        break;
```
Ericson2314, line 518: I think we need to set `hasUnknownHash = true` here or something so that `DeferredInputAddressed` cascades to downstream input-addressed derivations?

tests/content-addressed.nix
```diff
@@ -15,7 +15,7 @@ rec {
     '';
   };
   rootCA = mkDerivation {
-    name = "dependent";
+    name = "rootCA";
```
Ericson2314, line 18: Thanks :blush:

src/libstore/derivations.cc
```diff
@@ -514,7 +540,13 @@ DrvHashModulo hashDerivationModulo(Store & store, const Derivation & drv, bool m
                         justOut);
                 }
             },
+            [&](UnknownHashes) {
+                hasUnknownHash = true;
+            },
         }, res);
+        if (hasUnknownHash) {
+            return unknownHashes;
```
Ericson2314, line 548:
```suggestion
            return UnknownHashes {};
```

src/libstore/derivations.hh
```diff
@@ -168,9 +175,12 @@ std::string outputPathName(std::string_view drvName, std::string_view outputName
 // whose output hashes are always known since they are fixed up-front.
 typedef std::map<std::string, Hash> CaOutputHashes;
 
+struct UnknownHashes {};
+
 typedef std::variant<
     Hash, // regular DRV normalized hash
-    CaOutputHashes
+    CaOutputHashes, // Fixed-output derivation hashes
+    UnknownHashes // Deferred hashes for floating outputs drvs and their dependencies
```
Ericson2314, line 183: Is it safe to put `UnknownHashes` in the memo table? I long though it wasn't but I think since you represent them differently it is!

Ericson2314 (approved):
Nice to see this implemented! :)

regnat:
> Are you opposed to merging #3958 so it's easier to just review those?

Certainly not! This builds on top of #3958 but doesn't change anything to the way it works, so it definitely makes sense not to make it a blocker.

src/libstore/derivations.cc
```diff
@@ -454,6 +475,8 @@ static const DrvHashModulo & pathDerivationModulo(Store & store, const StorePath
     return h->second;
 }
 
+UnknownHashes unknownHashes;
```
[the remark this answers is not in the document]
regnat: Indeed we don't. I added it because I had forgotten that it should be `UnknownHashes {}` and not `UnknownHashes` if I wanted to inline it so it was failing with some obscure error :facepalm: 

Fixed in 3c723d3be

src/libstore/derivations.cc
regnat: I'm not sure. The reasoning behind the current code is that that way we can call `hashDerivationModulo` on a derivation that's advertising as `DeferredIA` but actually has a statically known hash. It's used in `rewriteDerivation` to convert the `DeferredIA` into an `IA`.
And if the derivation is indeed `Deferred`, then at least one recursive call to `hashDerivationModulo` will return `UnknownHash{}` so we'll get the same result

Ericson2314: Ah OK. That makes total sense. Then I'd say just one thing we might do is add a flag to `hashDerivationModulo` about whether it should try to un-defer `DeferredIA` derivations: in `build.cc` it definitely should just like this, but I think in `primops.cc` it shouldn't because that be impure, leaking what derivations we've already built into eval itself.

regnat: > that be impure, leaking what derivations we've already built and which CAs we got from them, into eval itself.

I don't think it does, `hashDerivationModulo` only looks at the derivations themselves. So if we depend on a floating-ca derivation it will always return `UnknownHash{}`, regardless of whether it has been built or not. The reason why we have a different return in `build.cc` is that we call it on the resolved derivation which is still marked as `Deferred` (because we haven't rebuilt its inputs yet) but doesn't depend on any floating-ca derivation anymore.

Ericson2314: Ah right, and it's only `drvPathResolutions` that we insert into by hand in `build.cc`, whereas `drvHashes` is just manually inserted into in `primops.cc`, where as you say there is nothing intrinsically impure going on. Sorry for the noise then! It's tricky but ends up working out great :).

Ericson2314: Wait, one thing is that input-addressed derivations conventionally keep their their input-addressed input drvs. But if we rewrite then `hashDerivationModulo`, they won't have any input drvs, even if those input drvs were input-addressed.

This isn't too bad, in fact this is how we "wish" input-addressed derivations worked all along, but it does mean that if one replaced a fixed output derivation with a floating CA derivation that produces the same outputs, one will get a different input-addressed path.

edolstra:
Please rebase now that #3958 is merged, it will make it a bit more convenient to review.

Ericson2314:
(rebasing is fine, but it should be enough merely to do something that will get GitHub to recompute the diff, such as pushing any new commit or even just setting the PR target branch to something else and then setting it back.)

regnat:
I've just rebased on top of master, should make the diff smaller

regnat:
I've just added a test case to ensure that non-ca derivations that depend on a CA one are properly cached (this is kind-of obvious given the flow of the code, but I think it's worth testing if only to prove that it actually works)

regnat:
Rebased on top of master to fix a conflict.

@edolstra is this good to go? Should be mostly uncontroversial after #3958

Ericson2314:
Well there was the thread up to https://github.com/NixOS/nix/pull/4056#discussion_r494511509 with a smidgen of controversy, but that's just something for you to be aware of @edolstra. I'm with this for now; and we can either change that later or just not care.

src/libstore/build/derivation-goal.cc
```diff
@@ -3166,6 +3167,15 @@ void DerivationGoal::registerOutputs()
             [&](DerivationOutputCAFloating dof) {
                 return newInfoFromCA(dof);
             },
+                [&](DerivationOutputDeferred) {
```
Ericson2314, line 3170: Indentation issue?

Issue #3550, opened 2020-04-30
btrfs compression ignored


vcunat:
Observed: btrfs compression settings is not honored for `/nix/store` paths substituted from binary cache (files are uncompressed), but it works for locally built files.

Source: [forum](https://discourse.nixos.org/t/good-filesystem-for-the-nix-store/3566/5?u=vcunat), confirmed locally (nix-2.3.4).

zimbatm:
How is the compression enabled? Is it not done transparently on the filesystem level?

vcunat:
A mount option.  That's how most people use it, I believe.

Normally it's transparent, so Nix most likely does something unusual, possibly around the read-only remounts and mount namespaces.  I'm affected even though `nix.readOnlyStore = false;`, but I suppose the inner read-write remount was a more likely culprit anyway.

asymmetric:
How can Nix disable the compression, if it happens at the filesystem level?

asymmetric:
Seems like that only triggers when the store is read-only though, which wasn't my case (and seems like it wasn't @vcunat's case either).

Anyway, I wonder if the call to `mount` should preserve the mount options the filesystem is mounted with, but I don't see a way to do that with [`mount(2)`](http://man7.org/linux/man-pages/man2/mount.2.html), so I must be missing something.

AleXoundOS:
> I suppose the inner read-write remount was a more likely culprit anyway

With `nix.readOnlyStore = false` there is no additional inner mount point for `/nix/store`. And newly substituted store paths still have no compression. Proof:
```console
$ mount | grep "on / type btrfs"
/dev/mapper/vg-fs_tree on / type btrfs (rw,noatime,compress-force=zstd,ssd,space_cache,subvolid=274,subvol=/fs_tree)
$ mount | grep -c /nix/store
0
$ nix-shell -p adbfs-rootless
these paths will be fetched (3.78 MiB download, 21.67 MiB unpacked):
  /nix/store/mkhq81nll904clsx020p42mqvsdawixd-platform-tools-28.0.1
  /nix/store/v4y99snaxs97aahmqzw6048557g536xk-adbfs-rootless-2016-10-02
copying path '/nix/store/mkhq81nll904clsx020p42mqvsdawixd-platform-tools-28.0.1' from 'https://cache.nixos.org'...
copying path '/nix/store/v4y99snaxs97aahmqzw6048557g536xk-adbfs-rootless-2016-10-02' from 'https://cache.nixos.org'...
[...]

PR #4057, opened 2020-09-23
nix/store on btrfs compression: add workaround

martinetd:
See #3550.

As suggested for now I've added a "preallocate-contents" option, and tested by restarting `nix-daemon` with `/path/to/output/derivation/bin/nix-daemon --option preallocate-contents false`

This works as a local fix but I definitely want to make this automatic for btrfs.

It's easy to call `statfs(realStoreDir.c_str(), &statfsbuf)` and check if `statfsbuf.f_type == BTRFS_SUPER_MAGIC` ; the question is how to make the setting "auto" ?

I'm thinking of:
 - changing from a Bool to a tristate true/false/auto enum; but would that work with the command-line `--option` switch? (I'm not too comfortable about introducing such a new setting type...)
 - if option is set to auto, check in `localStore` next to `makeStoreWritable` (e.g. some new `checkFsQuirks` method?), and if btrfs force the option somehow.. `globalConfig.set`? not really good with code compartimentization :/

Needless to say I don't like what I just suggested very much, so suggestions or comments welcome.
This is at least a starting point and folks who care can set the option in their nix.conf until then (depending on timing on the 3.0 release I'd be greedy and ask if it'd be possible to backport in 2.3 maintenance branch, but let's discuss that after we agree on the auto part I guess!)

Thanks!

### DIFF
--- a/src/libutil/archive.cc
+++ b/src/libutil/archive.cc
@@ -27,6 +27,8 @@ struct ArchiveSettings : Config
         #endif
         "use-case-hack",
         "Whether to enable a Darwin-specific hack for dealing with file name collisions."};
+    Setting<bool> preallocateContents{this, true, "preallocate-contents",
+        "Whether to preallocate files when writing objects with known size."};
 };
 
 static ArchiveSettings archiveSettings;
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks, I've merged it. We could add auto-detect later if there is interest.

PR #4059, opened 2020-09-24
Make ValidPath (nar) `hash` optional if `ca` field is present

### LINKED ISSUE #3640: Make ValidPath (nar) `hash` optional if `ca` field is present

Ericson2314:
We port the [`These` data type](https://hackage.haskell.org/package/these-1.1.1.1/docs/Data-These.html#t:These) from Haskell to C++ to maintain the invariant that we must have at least one of a CA or nar hash. This ensures that we can always validate the store entry.

Additionally, we also insure that the `LocalStore` and `BinaryCacheStore` always have NAR hashes, which makes this feature easier to undo without breaking data in the wild.

`LocalStore` and `BinaryCacheStore` sound like all the "base" stores we have today, so what's the prupose? The intended use-case for this feature would be novel stores that serve data by CA, such a `HashedMirrorStore` to replace the hashed mirrors mechanism, IPFS store, or software archive stores.

CC @nlewo

This PR is pretty noisy. Most of it is just replacing `.field` with various accessor methods.

Fix #3640


src/libstore/content-address.hh
```diff
@@ -12,12 +12,24 @@ enum struct FileIngestionMethod : uint8_t {
...
 
 /// Pair of a hash, and how the file system was ingested
 struct FixedOutputHash {
     FileIngestionMethod method;
     Hash hash;
+    bool operator ==(FixedOutputHash otherHash) const noexcept {
+        return method == otherHash.method && hash == otherHash.hash;
+    };
+    bool operator !=(FixedOutputHash otherHash) const noexcept {
+        return method != otherHash.method && hash == otherHash.hash;
```
cole-h, line 31:
```suggestion
        return method != otherHash.method || hash != otherHash.hash;
```

edolstra:
I'm not convinced this is a good idea. It seems that it makes everything way messier. I mean, it's basically a regression: NAR hashes used to be optional which meant that there lots of places where you could end up with a zero hash in your database.

Ericson2314:
There might be other ways of accomplishing my goal, such as an alternative to `queryPathInfo`, or letting stores choose what sort of a `ValidPathInfo` subtype they want with an associated type ("nested type name" I gather C++ calls them?).

To reiterate, I want to keep the requirement the database are narinfo files have this info. I suppose even having remote store / legacy ssh store continue to require NAR hashes is workable too. It just means one would have to talk to these new CA-only store directly and not via a remote store protocol.

tomberek:
OBE?

PR #4060, opened 2020-09-24
Make the tests depend on the nix executable

regnat:
That way just running `make installcheck` is enough to run them against the actual nix code in the repo.

(I don't know whether that's actually desirable for everyone. It seems to make sense to me to have this, so opening the PR anyways but feel free to close it if that breaks anything)

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -1,13 +1,11 @@
 # Run program $1 as part of ‘make installcheck’.
 
-test-deps =
-
 define run-install-test
 
   installcheck: $1.test
 
   .PHONY: $1.test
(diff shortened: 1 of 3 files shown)


edolstra:
I don't think this is enough. The tests also require corepkgs and the nix-* symlinks.

regnat:
Indeed. I've added a bunch of missing deps, `make clean && make installcheck` now works fine

regnat:
Rebased on top of master, and checked that `make clean && make installcheck` still works

regnat:
Closing as this isn’t compatible with https://github.com/NixOS/nix/pull/4239

PR #4063, opened 2020-09-24
Fixes fall-through to report correct description of hash-file command.

kquick:


Ericson2314 (approved):
Oops, that looks like my mistake! Thanks for fixing!

Issue #4050, opened 2020-09-22
flake registry could record dir path as well


kquick:
Currently a `nix registry add flake:foo github:team/repo?dir=subdir/in/repo` ignores the dir specification and just registers the repo itself.  This means that when used, the current syntax should be `inputs.foo.url = flake:foo?dir=subdir/in/repo` to properly reference the desired location.

The suggested enhancement is to allow the registry to record the dir specification as well.  This would allow the registry to be used to absorb any future changes in the corresponding repos.  For example, if `subdir/in/repo/flake.nix` became `nix/flake.nix`, this could be updated by the single change updating the registry for the `flake:foo` rather than editing every flake that imports `flake.foo` and changing the directory path in those individual flakes.

edolstra:
The registry does support subdirectories, see for instance [`blender-bin`](https://github.com/NixOS/flake-registry/blob/090d2e959af4b35d35d7b0717c5669e77ba9678b/flake-registry.json#L4-L13). Maybe `nix registry add` is dropping the `dir`?

cole-h:
What happens if you quote the URI (e.g. `nix registry add flake:foo 'github:team/repo?dir=subdir/in/repo'`)? Maybe your shell is doing something funky with the `?` that causes it to ignore everything after.

kquick:
Sorry for the confusion, there were multiple effects occurring here:
1. The directory path isn't currently visible in the `registry list` output.
2. I had some automation around this that was choosing  between a github: or path: target for the registry entry and it was specifying `$ nix registry add flake:foo /path/to/root?dir=sub/path`, which is accepted, but the attribute portion is silently dropped.
3. The result was invalid, so a `$ nix flake info flake:foo` displayed only an error.

I've added a pull-request to display the attribute info on the `registry list` command to help with the first item and hopefully avoid others getting confused by this.

P.S. Thanks for the suggestion @cole-h... I had tried that as well, and it didn't help.  It turns out that `nix` is the one ignoring the `?` and everything following it.  :-)

kquick:
Thanks for the merges and updates.  This should hopefully prevent anyone from getting confused like I was.

PR #4069, opened 2020-09-25
Add attrs like directory path to registry list output.

kquick:
The registry targets can have associated attributes (e.g. `github.com/owner/repo?dir=subdir/in/repo`) but the attributes are not displayed on a registry listing. They are currently only visible via the `$ nix flake info flake:foo` command, but if there is a problem with the registry entry, an error is displayed without any local information.  It is sometimes possible to extract the attribute from the error message, but since the attributes are local information that doesn't depend on the remote object, they should be visible independently of success in accessing the remote object.

This can also cause confusion about the registry support for these attributes (see issue #4050).

This patch adds the display of the attributes to the `registry list` command.  At present, the only attribute I'm aware of is the "dir" attribute; in the future it may be appropriate to add filtering if there are attributes that should not be visible.

### DIFF
--- a/src/nix/registry.cc
+++ b/src/nix/registry.cc
@@ -6,6 +6,7 @@
 #include "store-api.hh"
 #include "fetchers.hh"
 #include "registry.hh"
+#include <iterator>
 
 using namespace nix;
 using namespace nix::flake;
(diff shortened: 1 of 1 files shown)


edolstra:
Thanks! I've fixed this in a slightly different way (namely making `nix registry list` do basically the same thing as `FlakeRef::to_string()`). (64e9b3c83b7cf7f3c7348426666ccca2ca395d28)

PR #4072, opened 2020-09-25
Generate a warning on an implicit input specification.

kquick:
The ability to implicitly specify inputs as referencing flakes of the
same name is not specifically identified by the documentation (and
it's not clear if this behavior is intended to remain in the future).

If there is a typographical mismatch between an explicit input name
and the argument used for the `output`, this will result in a more
obscure error "cannot find flake 'flake:foo' in the flake registries",
which leads the user in the direction of managing the registries
instead of noticing the mis-spelling.

By adding a warning when an implicit input is created, it can make it
easier for the user to identify whether this was intended or whether
they have made an erroneous typo resulting in mismatched names.

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -218,11 +218,14 @@ static Flake getFlake(
 
         if (outputs->value->lambda.fun->matchAttrs) {
             for (auto & formal : outputs->value->lambda.fun->formals->formals) {
-                if (formal.name != state.sSelf)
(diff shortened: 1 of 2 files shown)


kquick:
As an update, I now see where these are documented in RFC0049, but I still think  that the warning is useful for the reason above.

kquick:
This is ready for merge.

tomberek:
stale: re-open as needed

PR #4074, opened 2020-09-26
Fix missing create-darwin-volume.sh in installer

stephank:
This appears to have gotten lost in the flakes merge.

I've made these changes by manually comparing with: https://github.com/NixOS/nix/blob/926c3a6664a9dfb288ca35af8aae40c4e4a2badb/release.nix

That commit number is based on the last successful pre-flakes Hydra build I could find of master: https://hydra.nixos.org/eval/1600231

Downstream issues:
- https://github.com/numtide/nix-flakes-installer/issues/4
- https://github.com/cachix/install-nix-action/issues/48

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -236,10 +236,10 @@
             }
             ''
               cp ${installerClosureInfo}/registration $TMPDIR/reginfo
+              cp ${./scripts/create-darwin-volume.sh} $TMPDIR/create-darwin-volume.sh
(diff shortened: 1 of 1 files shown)


domenkozar:
I'm not sure this is needed as master has been tested to work with Big Sur.

stephank:
Looks like a similar fix was part of #3996, so that's correct. Thanks!

PR #4078, opened 2020-09-26
doc/manual: update hacking docs

Ma27:
* By default, build artifacts should be installed into `outputs/` rather
  than `inst/`[1].
* Add instructions on how to run unit-tests.

[1] 733d2e9402807e54d503c3113e854bfddb3d44e0


Ericson2314:
Isn't it `outputs/out` by default?

Ma27:
You're absolutely right, sorry!

Ma27:
Oops forgot to push the update last night, just did that now :)

Issue #3779, opened 2020-07-02
flakes: nix build --override-input should NOT update flake.lock


colemickens:
**Describe the bug**

Oh this was quite surprising to see:

```
sudo nix --experimental-features 'nix-command flakes' --pure-eval \
  build --profile /nix/var/nix/profiles/system \
  --override-input master /home/cole/code/nixpkgs/master \
  --override-input stable /home/cole/code/nixpkgs/stable \
  --override-input cmpkgs /home/cole/code/nixpkgs/cmpkgs \
  --override-input pipkgs /home/cole/code/nixpkgs/pipkgs \
  --override-input home /home/cole/code/home-manager/cmhm-flakes   
    .#nixosConfigurations.xeep.config.system.build.toplevel


warning: Git tree '/home/cole/code/nixcfg' is dirty
warning: updating lock file '/home/cole/code/nixcfg/flake.lock':
* Updated 'home': 'git+file:///home/cole/code/home-manager/cmhm-flakes?ref=cmhm-flakes&rev=7ff1ecdecee6de0924804b1b7fac629d17580c4c' -> 'git+file:///home/cole/code/home-manager/cmhm-flakes?ref=cmhm-flakes&rev=b10c561c1f343f4268ab4d115977bfe1a93608ee'
warning: Git tree '/home/cole/code/nixcfg' is dirty
```

**Expected behavior**

In my opinion, a call to `nix-build` should not touch flake.lock ever. This is very unexpected to me. In fact, I virtually **never** expect `flake.lock` to be touched *unless* I'm making an explicit call to `nix flake ...`.

```
$ nix-shell --command "nix --version"
nix (Nix) 2.4pre20200622_334e26b
```

edolstra:
> In my opinion, a call to nix-build should not touch flake.lock ever.

Strongly disagree. Our behaviour is consistent with commands like `cargo build` that also update the lock file if you've made a change to `Cargo.toml` that requires the lock file to be updated.

colemickens:
But I'm explicitly *not* changing any files. I'm just specifying some command line args?

It's a `build` command. I would understand if I wrote `nix flake update --change-input`, but I  wrote `nix build --override-input` which I feel safe saying naively reads as "do a nix build but temporarily override this input".

What happened here strongly violates my expectations and I'm at least somewhat accustomed to using `cargo`.

colemickens:
[...]

PR #4086, opened 2020-09-28
Allow non-flake local path for --override-input

kquick:
This PR makes it possible to use --override-input for non-flake inputs as well as flake inputs.  See https://github.com/NixOS/nix/commit/40968dc679245c6b6a56fc8a5935fb22e5e2f21b for a full description of the main change in this PR.


kquick:
This is ready for review @edolstra and would be good get merged to fix flake support for `--override-input`.

colemickens:
Does this fix https://github.com/NixOS/nix/issues/3779? I don't know the Nix codebase well enough to understand the code, but it looks like the test case indicates it might? (Context: Currently there is no way to use `--override-input` and `--no-update-lock-file` (and `--no-write-lock-file` doesn't help).

kquick:
This does not fix #3779  as I understand it: this PR does not change the stance of flake.lock updates based on command-line `--override-input` .  The main thrust of this is to allow `--override-input` to reference a target that does not contain a `flake.nix`.  Non-flake inputs supported in the flake file via `input.foo = { ...; flake = false; }`, but prior to this PR, specifying `nix build --override-input foo other-location` would fail if `other-location/flake.nix` didn't exist.

@edolstra , I updated this from master to resolve the merge conflict, so it's in mergeable state again.

bqv:
I also desire this.

@edolstra by self assigning this, do you mean you're going to implement it yourself elsewhere?

kquick:
Re-synchronized changes with master again, ready for merging, @edolstra .

bqv:
AIUI when he blanks you like this it's cause he either doesn't care or he's going to do it better himself

jtojnar:
Or just busy with other work.

L-as:
This is still needed AFAICT?

cstml:
> This is still needed AFAICT?

Very much so.

thufschmitt:
So I’ve spent a bit of time trying to merge this with master to review it (there’s quite a bunch of conflicts :( ), only to realize that as far as I can tell, the underlying issue is already [fixed in master](https://github.com/NixOS/nix/pull/4959) (and apparently by me :man_facepalming: ). So I’ll close this.

That being said,
1. the split of `getFlake` in this branch might still be valuable
2. (to my shame) the fix didn’t include any test for it, while this one does

@kquick , do you think you could open a couple of PRs for just these?

thufschmitt:
(just ftr, the manual test I ran to make sure that it was indeed working on master was:
```console
$ cat <<EOF > flake.nix
{
  inputs.foo = {
    url = "path:./a";
    flake = false;
  };

  outputs = { self, foo }: {
    foo = builtins.readFile "\${foo}/foo";
  };
}
EOF
$ mkdir -p a && echo a > a/foo
$ mkdir -p b && echo b > b/foo
$ nix eval .#foo
"a\n"
$ nix eval .#foo --override-input foo $PWD/b
warning: not writing modified lock file of flake 'path:/tmp/tmp.HB372FSu5F':
• Updated input 'foo':
    'path:./a?lastModified=1&narHash=sha256-Nep3gAa7Eq4+Dv9plpoqx%2f+AJ+gWeV7U9lEiGCPRrR4=' (1970-01-01)
  → 'path:/tmp/tmp.HB372FSu5F/b?lastModified=1652775844&narHash=sha256-lGQM58%2frdQGqiOUqjGIYvo4nwaoVuGFnFx813YbaA5Y=' (2022-05-17)
"b\n"
```
)

L-as:
FWIW this still doesn't work for me

thufschmitt:
> FWIW this still doesn't work for me

Mh strange :thinking: 

Is the exact repro I sent broken or are you doing something slightly different? I just noticed that @colemickens [mentioned some time ago that my fix was broken](https://github.com/NixOS/nix/issues/5401), so maybe it just works under some very specific conditions

PR #4091, opened 2020-09-29
Access tokens for flakes

kquick:
This patch updates the ability to specify access tokens in the `nix.conf` file for use in the github fetcher.  This PR builds upon the previous functionality provided by @imalsogreg to allow the specification of more than two access tokens because:

 * Gitlab has two different kinds of access tokens that can be used (4 authorization methods, but just two that are relevant here).

 * Both gitlab and github are available for use by organizations and other groups, so they will exist at locations other than https://github.com and https://gitlab.com, both publicly and privately.

 * There are also non-github and non-gitlab sites hosting git that will have access controls.

This patch provides the ability to specify the access token to use for a specific remote host by matching against the `"host"` attribute in the corresponding flake input specification.

This adds a new entry to the `nix.conf` file called `access-tokens`, which is a space separated list of host=token specifications.  It removes the `gitlab-access-token` provided in PR #3675 because that token is new enough that it is not likely to be broadly used.  The `github-access-token` is still supported, but generates deprecation warnings when used.


src/libstore/globals.hh
```diff
@@ -860,10 +860,56 @@ public:
         )"};
 
     Setting<std::string> githubAccessToken{this, "", "github-access-token",
-        "GitHub access token to get access to GitHub data through the GitHub API for `github:<..>` flakes."};
+        "GitHub access token to get access to GitHub data through the GitHub API for `github:<..>` flakes (deprecated, please use 'access-tokens' instead)."};
```
edolstra, line 863: We can just remove this option.

kquick: Sounds good, I've removed this config option, @edolstra .

imalsogreg:
@kquick Have you been using this? I'm wondering if I'm doing it wrong, or if we have a bug. Demonstrating below:

I have the `access-tokens` field set in my nix.conf, and nix picks it up.
```
[root@nixos:~]# nix show-config | grep token
warning: unknown setting 'gc-check-reachability'
access-tokens = "gitlab.com=PAT:REDACTED github.com=REDACTED"
```

`nix flake info github.com:imalsogreg/some_private_repo` fails, and it does so not with a 404, but a 401, suggesting that my token is reaching the `curl` command within `nix`. (If I run this with no token, it's a 404).
```
[root@nixos:~]# nix flake info github:imalsogreg/work_queue
warning: unknown setting 'gc-check-reachability'
error: --- FileTransferError --------------------------------------------------------------------------------- nix
unable to download 'https://api.github.com/repos/imalsogreg/work_queue/commits/HEAD': HTTP error 401 ('Unauthorized')

response body:

{
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest"
}
```

Using the same github token, I am able to hit the URL that `nix` is trying to hit.
```
curl -H "Authorization: Bearer REDACTED" "https://api.github.com/repos/imalsogreg/work_queue/commits/HEAD"
[[succeeds]]
```

stephank:
The example (in the code comment) shows quotes, but it appears that's not really a syntax construct in nix.conf. It's just parsing it as a literal string including quotes.

@kquick In your case, it's reading your GitHub token as `github.com=REDACTED"` (notice the ending quote), which is an invalid token. It's also reading your GitLab token as `"gitlab.com=PAT:REDACTED`, which will never match the `gitlab.com` host. (I ran into the latter.)

Simply removing the quotes in nix.conf works.

imalsogreg:
Thanks @stephank , that was it! Opened a small documentation fix PR here: https://github.com/NixOS/nix/pull/4301

kquick:
Oops, yes, that's a bug in the docs: my local usage doesn't have the quotes.  Thanks @stephank and @imalsogreg for the investigation and PR to fix!

nrdxp:
Would it be possible to use this as is, or possibly extend it a bit to provide aws credentials (primarily for uploading to s3 cache). As it is, because a `post-build-hook` seems to clear the callers environment, it is impossible to set s3 credentials without root access for Nix daemon users. If not, I may need to open a proper issue for that.

kquick:
The credentials from the `access-tokens` are only used in the nix libfetcher code for git urls.  The nix libstore s3 code uses the aws-sdk-cpp library, so I suspect that will need some explicit support (viz. https://sdk.amazonaws.com/cpp/api/LATEST/md__docs__credentials__providers.html) to either use those credentials or some other mechanism (unless something is already providing this: I haven't used the S3 store myself).

PR #4093, opened 2020-09-29
Add eval-system option

matthewbauer:
Add a new `eval-system` option.
Unlike `system`, it just overrides the value of `builtins.currentSystem`.
This is more useful than overriding `system`, because you can build these derivations on remote builders which can work on the given system.
In contrast, `system` also effects scheduling which will cause Nix to build those derivations locally even if that doesn't make sense.

`eval-system` only takes effect if it is non-empty. If empty (the default) `system` is used as before, so there is no breakage.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -214,7 +214,11 @@ public:
           In general, you do not have to modify this setting.
           While you can force Nix to run a Darwin-specific `builder` executable on a Linux machine, the result would obviously be wrong.
 
-          This value is available in the Nix language as [`builtins.currentSystem`](@docroot@/language/builtin-constants.md#builtins-currentSystem).
+          This value is available in the Nix language as
+          [`builtins.currentSystem`](@docroot@/language/builtin-constants.md#builtins-currentSystem)
+          if the
+          [`eval-system`](#conf-eval-system)
+          configuration option is set as the empty string.
         )"};
 
     Setting<time_t> maxSilentTime{
--- a/doc/manual/rl-next/eval-system.md
+++ b/doc/manual/rl-next/eval-system.md
@@ -1,0 +1,12 @@
+---
+synopsis: Add new `eval-system` setting
+prs: 4093
+---
+
+Add a new `eval-system` option.
+Unlike `system`, it just overrides the value of `builtins.currentSystem`.
+This is more useful than overriding `system`, because you can build these derivations on remote builders which can work on the given system.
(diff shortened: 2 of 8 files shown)


Ericson2314:
Oh this is good I think. We can put this in `EvalSettings`, default it based on `settings.system`, and only set `builtins.currentSystem` based on this.

This removes the incentive to start talking about `builtins.currentSystem` in the libstore settings, which would be a layer violation. It also makes it easier to get rid of `settings` layer as we should not have global variables that only make sense for certain stores not others.

src/libstore/globals.hh
```diff
@@ -144,6 +144,19 @@ public:
           at build time.
         )"};
 
+    Setting<std::string> evalSystem{
+        this, "", "eval-system",
+        R"(
+          This option specifies the canonical Nix system name to
+          report when evaluating Nix expressions. This can be accessed
+          via builtins.currentSystem. Unlike `system`, this setting
```
fricklerhandwerk, line 152:
```suggestion
          via [`builtins.currentSystem`](@docroot@/language/builtins-constants.md#builtin-constants-currentSystem). Unlike [`system`](#conf-system), this setting
```

src/libstore/globals.hh
```diff
@@ -144,6 +144,19 @@ public:
           at build time.
         )"};
 
+    Setting<std::string> evalSystem{
+        this, "", "eval-system",
+        R"(
+          This option specifies the canonical Nix system name to
```
fricklerhandwerk, line 150:
```suggestion
          This option specifies the system type string to
```

https://github.com/NixOS/nix/pull/8570

src/libstore/globals.hh
```diff
@@ -144,6 +144,19 @@ public:
...
+          report when evaluating Nix expressions. This can be accessed
+          via builtins.currentSystem. Unlike `system`, this setting
+          does not change what kind of derivations can be built
+          locally. This is useful for evaluating Nix on your system
```
edolstra, line 154: "evaluating Nix": should be evaluating Nix code or something like that.

src/libstore/globals.hh
```diff
@@ -144,6 +144,19 @@ public:
...
+          but building the derivation on remote builders.
+
+          It defaults to `system` when unset.
```
fricklerhandwerk, line 157:
```suggestion
          It defaults to the value of the `system` configuration setting when unset.
```

Ericson2314:
I forgot tests and release note. Now they added, and this is ready to merge.

PR #4096, opened 2020-10-01
Remove title for nix-env

pickfire:


mweinelt:
Hi @pickfire, can you take the time to explain why this change makes sense?

edolstra:
The title field is metadata for lowdown so we can't remove it. It would be nice if it didn't appear in the HTML manual but there is no easy way to do that.

Ericson2314:
Isn't something like
```markdown
---
title: nix-env
---
```
a thing people do in markdown?

pickfire:
Huh? Oh, I forgot yesterday when I send this hactoberfest already started (I do want to do hacktoberfest but I bet 4 pulls are easy). No, it wasn't because of hacktoberfest, I was just browsing nix docs and I noticed this weird thing in nix-env https://nixos.org/manual/nix/unstable/command-ref/nix-env.html, looks like it is also there for nix-build but I don't see this in all pages.

PR #4101, opened 2020-10-02
Fix profile update in nix command

B4dM4n:
This fixes a wrong assertion when using `nix --profile ...` to update a profile:

```
$ nix build --profile my-profile -v nix
nix: src/nix/command.cc:155: nix::MixProfile::updateProfile(const Buildables&)::<lambda(nix::BuildableFromDrv)>: Assertion `!output.second' failed.
[1]    9634 abort (core dumped)  nix build --profile my-profile -v nix
```

This typo was intorduced by e9fad3006b0b6f3a6430fdbfc61392cac6834502.

CC @Ericson2314


grahamc:
It would be really nice to have some (unit and/or integration) tests on stuff like this. :(

PR #4105, opened 2020-10-03
Fix missing includes causing build failure

stephank:
I was trying to get Nix master (57b935a) to build on Darwin using `nix-shell`, but ran into missing string & string_view includes in several places. Curiously, it does build with `nix build --rebuild`, but I don't understand the difference.

### DIFF
--- a/src/libexpr/symbol-table.hh
+++ b/src/libexpr/symbol-table.hh
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <map>
+#include <string>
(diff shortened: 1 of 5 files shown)


edolstra:
I don't think these should be necessary. For instance, including `<string>` before `types.hh` shouldn't do anything since the latter already includes the former.

stephank:
Whatever the issue was, it seems to have disappeared. Strange! Thanks for taking the time to look.

Issue #4061, opened 2020-09-24
flake input url doesn't support branches with slashes


felschr:
**Describe the bug**

It seems that branches with slashes in them (e.g. `feat/flake`) aren't supported in input urls.
Neither of these work:
`"github:Myhlamaeus/pre-commit-hooks.nix/feat/flake"`
`"github:Myhlamaeus/pre-commit-hooks.nix?ref=feat/flake"`

**Steps To Reproduce**

Reference a git branch with a `/` in a flake.nix, for example:
```nix
inputs.pre-commit-hooks.url = "github:Myhlamaeus/pre-commit-hooks.nix/feat/flake";
```

**Expected behavior**
Flake should be downloaded.

**`nix-env --version` output**
nix-env (Nix) 3.0pre20200829_f156513

felschr:
Related: https://github.com/nixos/nix/issues/3612

Ma27:
@felschr would you mind testing #4115 :)

edolstra:
I don't think the first one should be supported since it's pretty ambiguous.

It's better in this case to use the attrset notation:
```
inputs.pre-commit-hooks = {
  type = "github";
  owner = "Myhlamaeus";
  repo = "pre-commit-hooks.nix";
  ref = "feat/flake";
};
```

AluisioASG:
It could be supported for the `?ref=` case though, no?

wucke13:
So I just got:

````
$ nix run 'github:petabyteboy/nixpkgs?ref=feature/seshat2#element-desktop'
URL 'github:petabyteboy/nixpkgs?ref=feature/seshat2#element-desktop' contains an invalid branch/tag name
````

using `nix (Nix) 3.0pre20200829_f156513`. Is this a bug or is it just not supposed to work? I'm not entirely sure, reading the comments in this issue...

Ma27:
Shouldn't this be solved with #4115 merged?

wucke13:
Yes totally, I had an old nix3 laying in my users profile. Sorry for the noise. As far as I'm concerned, we should close this issue, it's definitely working.

Silver-Golden:
@Ma27 I know this is an old issue but do you happen remember the logic behind parsing the ref from both the url and ``?ref=``?

Parsing it from the url causes issues for Gitlab, which has folders, there is a ticket over here: #8433

PR #4115, opened 2020-10-06
libfetchers/github: allow slashes in refs

Ma27:
Refs #4061

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -37,15 +37,29 @@ struct GitArchiveInputScheme : InputScheme
         std::optional<std::string> ref;
         std::optional<std::string> host_url;
 
-        if (path.size() == 2) {
-        } else if (path.size() == 3) {
+        auto size = path.size();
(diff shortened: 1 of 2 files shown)


Ma27:
@edolstra in #4061 you stated that this is probably not a good idea due to its ambiguity. As an alternative to this patch, would it make sense to improve the error to explicitly mention the workaround? Just having an `invalid URL`-error in that case is probably kinda confusing.

Ma27:
@edolstra may I ask what's your opinion re my previous comment?

AluisioASG:
If you keep only the regex change, does it work for the `?ref=` case? That doesn't seem as ambiguous to me.

Ma27:
Seems fine as well. @edolstra wdyt?

Ma27:
cc @edolstra how shall we proceed here?

edolstra:
Thanks. I was worried that it might introduce ambiguities in the other fetchers but that appears not to be the case.

PR #4116, opened 2020-10-07
support name attribute in fetchGit

AmineChikhaoui:
I think this was accidentally removed during the refactoring of
fetch{Tree,Git} which introduced issues such as:

```
unsupported Git input attribute 'name'
```


AmineChikhaoui:
I'm not sure the failures in the checks are false positives or not.

AmineChikhaoui:
@edolstra does this look good to you ?

AmineChikhaoui:
@edolstra would you mind just adding the simple change `name != "name"` so that fetchGit would support `name` as attribute ?

PR #4125, opened 2020-10-08
rust: small patches

Horki:
Small linter fixes for rust:
- redundant static lifetimes
-  '-' as u8: changed to b'-'; byte literal
- len() == 0; into is_empty()
- needles lifetimes
- etc..


edolstra:
Thanks!

PR #4128, opened 2020-10-09
Add a zsh completion script

regnat:
Extends the completion mechanism to also output the descriptions, and add a zsh completion script taking advantage of this.

This is a breaking change for the existing bash completion script (which can cause some minor issues if one updates the Nix client but not the script), but afaik it hasn't made it to any release yet, so it shouldn't be a major issue.

The current implementation also has an issue if the word to complete has a tab in it, but that sounds unlikely-enough to happen in practice

### DIFF
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -17,8 +17,19 @@ void Args::addFlag(Flag && flag_)
     if (flag->shortName) shortFlags[flag->shortName] = flag;
 }
 
+void Completions::add(std::string completion, std::string description)
+{
+    insert(Completion{
+            .completion = completion,
+            .description = description
+            });
+}
+
+bool Completion::operator<(const Completion & other) const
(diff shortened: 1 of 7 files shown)


src/libutil/args.cc
```diff
@@ -17,8 +17,19 @@ void Args::addFlag(Flag && flag_)
     if (flag->shortName) shortFlags[flag->shortName] = flag;
 }
 
+void Completions::add(std::string completion, std::string description)
+{
+    insert(Completion{
```
edolstra, line 22: Maybe add an assert here to check that `description` doesn't contain a newline.

matthewbauer:
Could you add a local.mk file like:

```
$(eval $(call install-file-as, $(d)/completion.zsh, $(datarootdir)/zsh/site-functions/_nix, 0644))
```

so it's picked up by zsh fpath?

PR #4132, opened 2020-10-10
ZSH completion should be written like this

doronbehar:
The previous script was a script one needs to `source`. This script is a script one adds to their ZSH fpath. See:
https://unix.stackexchange.com/a/33898/135796

cc @regnat

More over, Nixpkgs should chose this script here if `nixUnstable` is used:

https://github.com/NixOS/nixpkgs/blob/763bbfde7c3f3aa7f84b443b0c3fbbbd2c35f53e/nixos/modules/programs/zsh/zsh.nix#L281-L282


doronbehar:
Haven't seen https://github.com/spwhitt/nix-zsh-completions/pull/34 . Closing.

regnat:
Given the size of this completion script, I think it's nice to have it in the Nix codebase itself so that we have at least a simple completion works out-of-the box, so it's totally fine to reopen this imho

doronbehar:
Yes but it's not even installed, so it seems:

https://github.com/NixOS/nix/blob/20d2140e450b066a521933dd322d089fd6c248fa/Makefile#L14-L17

If this PR would also make it get installed, we'll have collisions in NixOS systems:

https://github.com/NixOS/nixpkgs/blob/763bbfde7c3f3aa7f84b443b0c3fbbbd2c35f53e/nixos/modules/programs/zsh/zsh.nix#L281-L282

non-NixOS users might benefit from this change (without making it get installed) but I would recommend them anyway to use nix-zsh-completions directly, so I don't care enough to maintain this patch.

Issue #4133, opened 2020-10-11
nix-build to replace CR->LF in builder output


volth:
Some builders (`ninja` to name one) use CR at the end of line, effectively printing

```
[1/990] compiling uno.cpp\r
[2/990] compiling dos.cpp\r
[3/990] compiling tres.cpp\r
```
supposing to update the last line without scrolling the lines upward

As this `ninja` output passes through `nix-build`, `nix-build` buffers it and there is no output at all for long time.
It would be nice to replace CR->LF to see the progress

bjornfor:
This is a regression since Nix started running builds in a pseudo terminal. Current workarounds: `ninja | tee dummy` or `ninja -v`.

volth:
It looks intentional:
https://github.com/NixOS/nix/blob/54f4500457f1849163d3f7d82228a50e5c372316/src/libstore/build.cc#L4605-L4609

PR #4134, opened 2020-10-11
do not skip CR-ended lines emitted by builder

volth:
Some builders (`ninja` to name one) use CR at the end of lines, effectively printing

```
[1/990] compiling uno.cpp\r
[2/990] compiling dos.cpp\r
[3/990] compiling tres.cpp\r
```
supposing to update the last line without scrolling the lines upward

As this `ninja` output passes through `nix-build`, `nix-build` buffers it and there is no output at all for long time.
It would be nice to replace CR->LF to see the progress

Fixes: https://github.com/NixOS/nix/issues/4133

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -4601,16 +4601,18 @@ void DerivationGoal::handleChildOutput(int fd, const string & data)
             return;
         }
 
-        for (auto c : data)
-            if (c == '\r')
-                currentLogLinePos = 0;
-            else if (c == '\n')
-                flushLine();
-            else {
+        int lastc = -1;
+        for (auto c : data) {
+            if (c == '\r' || c == '\n') {
+                if (lastc != '\r')
+                    flushLine();
+            } else {
                 if (currentLogLinePos >= currentLogLine.size())
                     currentLogLine.resize(currentLogLinePos + 1);
                 currentLogLine[currentLogLinePos++] = c;
             }
+            lastc = c;
+        }
 
         if (logSink) (*logSink)(data);
     }


Ericson2314:
Is this for logging or also for terminal output? IMO buffering pty -> pty doesn't really make sense at all, and we should preserve all the `\r` tricks by virtue of doing little-to-nothing. OTOH the log should be append only.

volth:
> Is this for logging or also for terminal output? IMO buffering pty -> pty doesn't really make sense at all, and we should preserve all the `\r` tricks by virtue of doing little-to-nothing. OTOH the log should be append only.

It is only for terminal output, logging preserves `\r`

volth:
My decision to let `nix-build` emit builders' `\r` as `\n` might look too opinionated, it makes look-and-feel different from running a builder:




But there is a reason for that: `nix-build` can combine outputs of many builders, and if it would emit `\r` it might interfere badly with output of another builder.

Ericson2314:
Oh, I forgot about the interleaving. Yeah then we can't really preserve the pty-ness meaningfully, as 2D text displays don't interleave.

Unless we could somehow change the behavior based on whether we were interleaving or displaying the raw terminal output, +1 from me.

I suppose the ideal thing to do would be tell the builder programs that "your output is text stream but it does support colors", but sadly Unix doesn't offer that midway between pty and pipe.

volth:
> I suppose the ideal thing to do would be tell the builder programs that "your output is text stream but it does support colors"

It would not prevent them to print `\r`-ended lines

Ericson2314:
@volth Sure, but if they didn't think they could "return the carriage with feeding the a new line", they might not try --- c.f. what ninja does when given a pipe. And if they did try anyways I would feel all the more justified normalizing it to a `\n`.

I guess I didn't mean "ideal" w.r.t to this PR, but ideal w.r.t the switch from pipe to pty that Nix did.

volth:
There are too many builders like `ninja`: maven, npm, git, wget, curl, ...
I bet no one of them do any checks before emitting \r.
And.. even if they would do, the check will succeed, because the there
is a pseudo TTY in sandbox.

Ericson2314:
We're basically in agreement. I'm saying this is the best fix we got. :)

> And.. even if they would do, the check will succeed, because the there
is a pseudo TTY in sandbox.

I was just saying think that switch to the pty got us more funny things than we bargained for. (c.f. https://github.com/NixOS/nix/issues/4133#issuecomment-706671966 from the issue), but I don't want to go back to pipes because I like the colors.

volth:
`nix-shell` seems not affected: there was no issue with lost `\r`-ended lines

PR #4135, opened 2020-10-11
Split out `local-fs-store.hh`

Ericson2314:
This matches the already-existing `local-fs-store.cc`.

CC @regnat


Ericson2314:
Maybe this is can be a less controversial split :).

Ericson2314:
Thanks!

PR #4136, opened 2020-10-11
Split out uds-remote-store.{cc.hh}

Ericson2314:
The Raymond Chen method doesn't work because the original filenames are still used at the end, but I think this copying is the next best thing.

CC @regnat


Ericson2314:
@edolstra I promise I'm not binary searching what level of splitting you'll put up with :smile:.

Issue #4138, opened 2020-10-12
The drvOutputId->OutputPath link is lost for CA derivations when the drv file is deleted


regnat:
**Describe the bug**

The sqlite database keeps a mapping between derivations output − a tuple (derivationFile, outputName) − and their actual output path as this one can't be inferred from the derivation file itself when CA derivations come into play.
However, this mapping is actually `(pathIdx, outputName) -> outputPath` where `pathIdx` is the index of the entry representing the derivation file in the db. In practice, this means that once the drv file is garbage-collected this index doesn't have any meaning anymore, so this mapping if effectively lost (there's even an sql rule saying that it should be deleted when the drv file is garbage-collected). So Nix won't be aware anymore of the fact that the derivation has been built, and the next `nix-build` call will have to rebuild it.

**Steps To Reproduce**

From a recent-enough Nix, apply the following patch and run `make install && make tests/content-addressed.sh.test`

```diff
diff --git a/tests/content-addressed.sh b/tests/content-addressed.sh
index 61ec03fe3..82a979b2a 100644
--- a/tests/content-addressed.sh
+++ b/tests/content-addressed.sh
@@ -5,6 +5,15 @@ source common.sh
 drv=$(nix-instantiate --experimental-features ca-derivations ./content-addressed.nix -A rootCA --arg seed 1)
 nix --experimental-features 'nix-command ca-derivations' show-derivation --derivation "$drv" --arg seed 1
 
+# Build the derivation a first time and add a gc root for the output path
+nix-build --experimental-features ca-derivations ./content-addressed.nix -A rootCA --arg seed 1 -o $TEST_ROOT/var/nix/gcroots/caRoot
+# Run the gc to remove the drv file
+nix-store --gc
+# Try buinding it again with `-j0`. This should succeed as the output path is
+# still here, but will fail because we lost that knowledge when removing the
+# drv file
+nix-build --experimental-features ca-derivations ./content-addressed.nix -A rootCA --arg seed 1 -o $TEST_ROOT/var/nix/gcroots/caRoot -j0
+
 testDerivation () {
     local derivationPath=$1
     local commonArgs=("--experimental-features" "ca-derivations" "./content-addressed.nix" "-A" "$derivationPath" "--no-out-link")
```

**Expected behavior**

The above should work.
More generally, one should be able to register a `drvOutputId -> outputPath` mapping without having the drv file around

cc @edolstra @Ericson2314

regnat:
[...]

PR #4139, opened 2020-10-12
Always keep the CA derivations around

regnat:
Workaround for https://github.com/NixOS/nix/issues/4138

### DIFF
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -574,13 +574,16 @@ bool LocalStore::canReachRoot(GCState & state, StorePathSet & visited, const Sto
 
     /* If keep-derivations is set and this is a derivation, then
        don't delete the derivation if any of the outputs are alive. */
-    if (state.gcKeepDerivations && path.isDerivation()) {
-        for (auto & [name, maybeOutPath] : queryPartialDerivationOutputMap(path))
-            if (maybeOutPath &&
-                isValidPath(*maybeOutPath) &&
-                queryPathInfo(*maybeOutPath)->deriver == path
-                )
-                incoming.insert(*maybeOutPath);
+    if (path.isDerivation()) {
+        auto drv = readDerivation(path);
+        if (state.gcKeepDerivations || derivationIsCA(drv.type())) {
+            for (auto & [name, maybeOutPath] : queryPartialDerivationOutputMap(path))
+                if (maybeOutPath &&
+                    isValidPath(*maybeOutPath) &&
+                    queryPathInfo(*maybeOutPath)->deriver == path
+                    )
+                    incoming.insert(*maybeOutPath);
+        }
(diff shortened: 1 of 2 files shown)


edolstra:
Does this mean that CA derivations are never garbage-collected?

regnat:
They aren't until their output-path is (like if `keep-derivations` is set to `true` in the config)

Ericson2314:
In fact, perhaps they should be kept around even if their output isn't.

Suppose we have:

> ( resolved-gcc.drv , "out" ) -> `foo`
> ( resolved-bash.drv [ `foo` / "gcc" ] , "out" ) -> `bar`

If we get rid of the row for `foo`, then if someone tries to rebuild (unresolved) bash.drv they'll end up rebuilding GCC!

We're going to need a pretty big migration I'm afraid to have the `bar` row preserve the `foo` row without also preserving the `foo` path itself.

Ericson2314 (approved):
All that said, this PR makes the situation wholly better, and it is an easy enough stop-gap to back out of whenever we do do the sqlite overhaul.

regnat:
> In fact, perhaps they should be kept around even if their output isn't.
> 
> Suppose we have:
> 
> > ( resolved-gcc.drv , "out" ) -> `foo`
> > ( resolved-bash.drv [ `foo` / "gcc" ] , "out" ) -> `bar`
> 
> If we get rid of the row for `foo`, then if someone tries to rebuild (unresolved) bash.drv they'll end up rebuilding GCC!
> 
> We're going to need a pretty big migration I'm afraid to have the `bar` row preserve the `foo` row without also preserving the `foo` path itself.

I'm not 100% sure, but I think that the `DerivationOutputs` entries stay there as long as the corresponding derivation is there (even if their output is gc-ed), so in that case the first mapping would stay alive because `resolved-gcc.drv` would (as it's a dependency of `resolve-bash.drv`)

Ericson2314:
Ah, but `resolved-gcc.drv` is *not* a dependency of `resolve-bash.drv`, because resolved floating CA derivations always have empty `inputDrvs`. That's the difficulty.

Ericson2314:
Ah CI failed because of that annoying spurious failure. Do we have an issue for it?

regnat:
> Ah, but `resolved-gcc.drv` is _not_ a dependency of `resolve-bash.drv`, because resolved floating CA derivations always have empty `inputDrvs`. That's the difficulty.

Oh yes because we only have this mapping for resolved drvs :thinking: 
I think we need to dig all that a little to properly fix this (but the workaround is worth keeping in the meantime)

edolstra:
> In fact, perhaps they should be kept around even if their output isn't.

But then derivations will accumulate forever... Or they could be treated like a cache and have some expiration policy.

PR #4143, opened 2020-10-13
Properly type the derivation and substitution goal maps

Ericson2314:
As a bonus, `Worker::removeGoal` is less inefficient, although not yet what I'd call efficient. :)


src/libstore/build/worker.cc
```diff
@@ -89,14 +86,14 @@ GoalPtr Worker::makeSubstitutionGoal(const StorePath & path, RepairFlag repair,
...
 }
 
-
-static void removeGoal(GoalPtr goal, WeakGoalMap & goalMap)
+template<typename G>
+static void removeGoal(std::shared_ptr<G> goal, std::map<StorePath, std::weak_ptr<G>> & goalMap)
 {
     /* !!! inefficient */
-    for (WeakGoalMap::iterator i = goalMap.begin();
+    for (typename std::map<StorePath, std::weak_ptr<G>>::iterator i = goalMap.begin();
```
edolstra, line 93: Doesn't `auto i = ...` work here?

edolstra (approved):

src/libstore/build/worker.cc
Ericson2314: It does!

Issue #2114, opened 2018-04-27
builtins.fetchurl with fixed sha256 doesn't attempt to substitute at evaluation time


copumpkin:
It seems like it should check and use the configured binary caches, right? I have a scenario:

1. My binary cache contains a fixed-output derivation whose hash matches the sha256 in a `builtins.fetchurl` call
2. The machine evaluating `builtins.fetchurl` for various reasons has no access to the underlying URL (think proprietary software or authentication)
3. If I just evaluate the file, it fails because it's unable to download the URL in question
4. If I `nix-store -r` the path, then evaluating the file works fine

I'm using `builtins.fetchurl` rather than a nixpkgs flavor here because it speaks native authenticated S3. I could do it inside a conventional builder, but I figured I'd report this strange behavior.

evanrelf:
I still care about this

PR #4149, opened 2020-10-15
WIP: fetchTree: substitute from binary caches before trying to fetch

andir:
This allows fetching the referred sources from a binary cache in
addition to the original source Uri. In cases where accessing the
upstream is expensive, slow or unreliable and the soures are within on
of the trusted binary caches we can now subsitute them from there.

TODOs:
 - [ ] mostly a copy of what substitution-goal.cc does, this might mean we should refactor the entire substitution code to be useable for cases like this?
 - [ ] figure out which of the headers clangd pulled in on the fly are
       actually required for this
 - [ ] what are the downsides of having getPublicKeys public? It
       appears that at least one of the friend class cases only existed
       because it was not public.

Relates to #2114

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -1,3 +1,8 @@
+#include "content-address.hh"
+#include "crypto.hh"
+#include "local-store.hh"
+#include "local-fs-store.hh"
+#include "path-info.hh"
 #include "primops.hh"
 #include "eval-inline.hh"
 #include "store-api.hh"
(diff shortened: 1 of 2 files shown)


Ericson2314:
> mostly a copy of what substitution-goal.cc does, this might mean we should refactor the entire substitution code to be useable for cases like this?

Check out #3754, which among many other changes does this without copying that stuff.

Ericson2314:
In fact https://github.com/NixOS/nix/blob/11882d7c7ce3b6dc51dd7c0536f9662dc254ac0a/src/libfetchers/fetchers.cc#L122 is supposed to try substitutors already. We should figure out why it's not working for you.

andir:
> In fact
> 
> https://github.com/NixOS/nix/blob/11882d7c7ce3b6dc51dd7c0536f9662dc254ac0a/src/libfetchers/fetchers.cc#L122
> is supposed to try substitutors already. We should figure out why it's not working for you.

The reason might be that I'm using a release version of Nix (2.3.7) and didn't expect this to have been changed.

andir:
> > mostly a copy of what substitution-goal.cc does, this might mean we should refactor the entire substitution code to be useable for cases like this?
> 
> Check out #3754, which among many other changes does this without copying that stuff.

That indeed looks much nicer! Looking forward to play with that once it's merged/released :+1:

PR #4151, opened 2020-10-15
Bump Nixpkgs to hopefully fix linkrot

Ericson2314:
There was an issue fetching a patch for a dependency for musl builds in #3765. Bumping Nixpkgs should fix it, but I'll do that separately in this PR to isolate any problems that might arise in either.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,11 +18,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1591633336,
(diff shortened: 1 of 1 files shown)


Kloenk (approved):

edolstra:
I'm thinking we should switch to nixpkgs-20.09, but I just realized the removal of the Perl VM test infrastructure makes that a bit more work...

Ericson2314:
Oh interesting. I thought 20.09 wasn't out yet so I didn't switch for that reason, but I'll keep that on my radar.

PR #4153, opened 2020-10-15
Allow a "file" libfetcher type for plain files

### LINKED ISSUE #3785: flakes: support downloading files (as opposed to github, git and tarball types)
I've sometimes depended on binary packages fetched from a remote. With flakes, this isn't possible to do within the inputs, eg. I can't do something like:

```nix
inputs.rust-analyzer = {
  type = "file";
  url = "https://github.com/rust-analyzer/rust-analyzer/releases/download/2020-06-15/rust-analyzer-linux";
  flake = false;
};
```

I would like it if there was a way to specify an input of type file.

The alternative I'm currently using is to specify the hash, url etc directly within the derivation. However, this means I cannot update it using the `nix flake` commands and I have to deal with any such packages differently from everything else.

**Additional context**
I believe this is useful in both the context above with rust-analyzer when you don't want to build it yourself, but there may be situations where you can't build the software yourself (eg. proprietary software for example, or maybe downloading rpm:s) but would still like to manage them the same as all other dependencies.

bqv:
Resolves #3785

Hopefully isn't awful


johnae:
I'm the original issue author. Would really like to see this feature in Nix flakes. Thanks for working on this @bqv!

bqv:
@edolstra any thoughts? Do you have something else in the works?

junyi-hou:
Second this. I think this is a great feature that allows flakes gets used in more places.

Yajo:
Really needed! Could somebody from the core team review it? 😃

tomberek:
duplicate? https://github.com/NixOS/nix/issues/5979

thufschmitt (changes requested):
Sorry for the very long delay @bqv . I’ve taken the liberty of merging `master` into your branch to fix the conflicts, but feel free to force-push over it if you don’t like it.

Feature-wise, that looks good. It’s not feature-complete − you can’t for example specify a `file` input as a url − but it’s already a strict improvement over the existing.

I’m a bit annoyed by the fact that this causes the “tarball” input scheme to fetch more than just tarballs. Maybe you could add the “file” support as its own scheme (by writing a `FileInputScheme` similar to `TarballInputScheme` and registering it the same way).
That shouldn’t cause a lot of duplication since most of the logic is already factored out in `downloadFile` and should make the whole thing more logical.

Also, can you add some tests? Both testing direct calls to `builtins.fetchTree { type = "file"; ... }` and flake inputs of type `file`.

bqv:
Not remotely ;) Left open because I honestly just forgot. Someone else can takeover this, I'm closing it but I'll leave the branch open.

thufschmitt:
> Not remotely ;)

Fair enough :smile: Let’s hope that someone will then :pray: (happy to provide guidance to any volunteer)

tomberek:
still interested

Issue #3534, opened 2020-04-24
nix-build loops while building


domenkozar:
Nix 2.3.4

Nix seems to loop between the same derivations, see a sample of the log at https://gist.github.com/anamarijas/c5a1ea49ca1e1cbdd3f70f5a8d4aa969

To reproduce:

- `git clone https://github.com/cachix/ghcide-nix`
- `cd ghcide-nix`
- `cachix use ghcide-nix`
- `git checkout ghc-8.8`
- `nix-build`

domenkozar:
I've tested on 2.3.3 and master, same loop.

andir:
I saw a similar issue a while ago when a binary cache reported a 404 on a file that should be there. Nix did also complain that it didn't know how to build some paths. Can you post the full output that you get from Nix?

domenkozar:
Full output is 300MB, do you have something to grep for?

I can confirm removing binary caches makes it build.

Was 404 on nar file or narinfo?

andir:
On 10:42 26.04.20, Domen Kožar wrote:
> Was 404 on nar file or narinfo?

I honestly do not recall anymore. Could have been either. My guess is on
a NAR file but I can't provide evidence.

By now I can't reproduce it anymore. I just noticed that the temproots
file would eventually fill up the entire disk and only then the build
would abort.

It was on some cachix bucket shortly after cloudflare was enabled there,
if that helps.

domenkozar:
Thanks @andir I'll take a look at this.

domenkozar:
My initial hypothesis is that this is an instance of https://github.com/NixOS/nix/issues/3373 but during IFD.

andir:
On 02:39 27.04.20, Domen Kožar wrote:
> My initial hypothesis is that this is an instance of
> https://github.com/NixOS/nix/issues/3373 but during IFD.

In my case there was no IFD.

domenkozar:
Refs https://github.com/NixOS/nix/issues/1217

domenkozar:
I can confirm this is due to missing references, so referenced issues are valid.

I managed to reproduce this again and in the logs it can be observed that Nix is looping between:

```
some references of path '/nix/store/kk48ba50zs7dj2n30nx4wxxan6273rnq-basement-0.0.11-lib-basement' could not be realised
some references of path '/nix/store/10rgwgpv65fvwajap619pn7i1xj00zky-blaze-builder-0.4.1.0-lib-blaze-builder' could not be         realised
```

[...]

PR #4158, opened 2020-10-18
Fix substitution loop #3964, #3534

### LINKED ISSUE #3964: Build loops trying to substitute when output is missing

roberth:
Closes #3964, #3534 

It did not require refactoring or extra parameters to terminate the loop.

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -330,8 +330,13 @@ void DerivationGoal::outputsSubstitutionTried()
 
     /*  If the substitutes form an incomplete closure, then we should
         build the dependencies of this derivation, but after that, we
-        can still use the substitutes for this derivation itself. */
-    if (nrIncompleteClosure > 0) retrySubstitution = true;
+        can still use the substitutes for this derivation itself.
+
+        If the nrIncompleteClosure != nrFailed, we have another issue as well.
+        In particular, it may be the case that the hole in the closure is
+        an output of the current derivation, which causes a loop if retried.
+     */
+    if (nrIncompleteClosure > 0 && nrIncompleteClosure == nrFailed) retrySubstitution = true;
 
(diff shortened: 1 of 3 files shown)


Ericson2314 (approved):

ajs124:
I think I'm seeing this (or maybe some other substitution loop) on 2.3.9, was this backported?

PR #4161, opened 2020-10-18
nix develop: Add --redirect flag to redirect dependencies

edolstra:
This is primarily useful if you're hacking simultaneously on a package and one of its dependencies. E.g. if you're hacking on Hydra and Nix, you would start a dev shell for Nix (which installs in e.g. `~/Dev/nix/outputs`), and then a dev shell for Hydra as follows:

```
$ nix develop \
  --redirect .#hydraJobs.build.x86_64-linux.nix ~/Dev/nix/outputs/out \
  --redirect .#hydraJobs.build.x86_64-linux.nix.dev ~/Dev/nix/outputs/dev
```

(This assumes `hydraJobs.build.x86_64-linux` has a `passthru.nix` attribute. You can also use a store path.)

This causes all references in the environment to those store paths to be rewritten to `~/Dev/nix/outputs/{out,dev}`. Note: don't forget to run `fixupPhase` in the first shell, otherwise pkgconfig files are not in the correct location. Note 2: unfortunately, you may need to set `LD_LIBRARY_PATH=~/Dev/nix/outputs/out/lib` because Nixpkgs' ld-wrapper only adds `-rpath` entries for `-L` flags that point to the Nix store.

@viric

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -185,7 +185,22 @@ struct Common : InstallableCommand, MixProfile
         "UID",
     };
 
+    std::vector<std::pair<std::string, std::string>> redirects;
+
+    Common()
+    {
+        addFlag({
+            .longName = "redirect",
+            .description = "redirect a store path to a mutable location",
+            .labels = {"installable", "outputs-dir"},
+            .handler = {[&](std::string installable, std::string outputsDir) {
+                redirects.push_back({installable, outputsDir});
+            }}
+        });
+    }
+
     std::string makeRcScript(
+        ref<Store> store,
         const BuildEnvironment & buildEnvironment,
         const Path & outputsDir = absPath(".") + "/outputs")
     {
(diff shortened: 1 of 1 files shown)


viric:
Oh I'll have to give a try to this! I have yet to grasp flakes fully. I'll go through your blogposts for that.

viric:
Do you think this will work for a common c++ scenario of, let's say, qpdfview and poppler?

edolstra:
@viric It should also work without flakes though (e.g. `nix develop -f /path/to/nixpkgs poppler`). Though I just ran into a bug starting `nix develop` on qpdfview...

tomberek:
Is this supposed to work when the input is not a flake? Or is that a desired feature?

SeTSeR:
Can this be used to build some package against host libc? I wanted to build latest librsvg from Nix against libc-2.31 from Ubuntu 20.04. I used this command to start shell:
```sh
nix develop nixpkgs/nixos-unstable#librsvg --redirect nixpkgs/nixos-unstable#glibc.dev /usr
```
However, I got the warning, that this flake is not used by the build environment. Then I tried to write simple program to ensure I build against old libc:
```c
#include <unistd.h>

int main(void) {
    closefrom(3);
    return 0;
}
```
I expected that this program should not compile as `closefrom` was introduced in `glibc-2.34`, but it compiles successfully. Am I missing something?

PR #4171, opened 2020-10-20
Fix iterating over $NIX_PROFILES in Zsh

### LINKED ISSUE #4167: nix-daemon.sh doesn't set NIX_SSL_CERT_FILE on macOS in zsh
**Describe the bug**

When `nix-daemon.sh` is imported from `/etc/zshrc` (or new `/etc/zshenv`) on macOS, it tries to guess a proper value for `NIX_SSL_CERT_FILE`, but fails, because there is no special case for macOS, and catch all clause doesn't work as expected.

**Steps To Reproduce**

1. Install Nix with daemon on macOS
2. Start zsh shell
3. Run `echo $NIX_SSL_CERT_FILE`
4. See nothing

**Expected behavior**

When `nss-cacert` package is installed in root profile, it should be `/nix/var/nix/profiles/default/etc/ssl/certs/ca-bundle.crt`.

```
nix-env (Nix) 2.3.7
```

**Additional context**

Issue is in relying on shell splitting words in a string variable in `for` loop. It's [not the case](http://zsh.sourceforge.net/FAQ/zshfaq03.html#l18) for Zsh.

YorikSar:
NIX_PROFILES is space separated list of directories, and passing it into
for as is is considered to be 1-element list with the whole string. With
shwordsplit option Zsh emulates other shells in this regard ans
implicitely splits unquoted strings into words.

Fixes #4167.

### DIFF
--- a/scripts/nix-profile-daemon.sh.in
+++ b/scripts/nix-profile-daemon.sh.in
@@ -17,11 +17,21 @@ elif [ -e /etc/pki/tls/certs/ca-bundle.crt ]; then # Fedora, CentOS
     export NIX_SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt
 else
   # Fall back to what is in the nix profiles, favouring whatever is defined last.
-  for i in $NIX_PROFILES; do
-    if [ -e $i/etc/ssl/certs/ca-bundle.crt ]; then
-      export NIX_SSL_CERT_FILE=$i/etc/ssl/certs/ca-bundle.crt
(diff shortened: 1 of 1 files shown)


jdheyburn:
I was experiencing the error in the linked issue, and only on manually applying this change did it work, so thank you for the PR 🎉 . 

For anyone stumbling on this before it is released, I manually altered this file for the fix: `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh`

regnat (approved):

PR #4174, opened 2020-10-21
Allow the copy of ca derivation outputs

regnat:
Big ugly PR adding support for moving CA derivation outputs around.
Not to be merged as-it-is, but can work as a POC.

It's obviously already possible to move the output paths around, but there's no way to copy the mapping `(drvPath, outputName) -> outputPath` which is essential for ca derivations. That's what this PR enables. More precisely, it

- Define (in code) the notions of:
  - **derivation output**: the triplet `(drvPath, outputName, outputPath)` along with some meta information (currently a set of dependencies, should probably be extended at least with a signature),
  - **derivation output id**: The tuple `drvPath!outputName` that uniquely (modulo non-determinism) identifies a derivation output,
  - **derivation input**: either a plain store path or a derivation output
    (this is very badly named because it's a general identifier for the elements of the store and not just the thing that's fed to derivations).
- Add a `Store::registerDrvOutput(DrvOutputId&, DrvOutputInfo)` method which generalizes `LocalStore::linkDeriverToPath`
  1. By making it a virtual method of the `Store` class (so that every store can implement it rather than just `LocalStore`)
  2. By allowing to not just register the `DrvOutputId->outPath` mapping, but also all the meta-information that might be needed
- Allow `copyPaths(...)` and the `nix copy` command to work on derivation outputs
  - `nix copy nixpkgs#hello` will now copy the derivation outputs of `nixpkgs#hello` rather than just the output paths.
  - It's now possible to run `nix copy /nix/store/....-hello.drv!out` to manually specify a derivation output
  - `nix copy /nix/store/....-hello` will just copy the output path like it used to do
- Fix https://github.com/NixOS/nix/issues/4138 (because otherwise copying drv outputs to a remote store would require copying the whole drv closure too which is painful).

Still to do:

- [ ] Allow signing drv outputs (which will be needed as it's the bit that has to be trusted now)
- [ ] Implement transparent substitution (needs hooking into the `SubstitutionGoal` which I haven't done yet
- [ ] Properly register the dependencies of a drv output in the database.
  To allow properly copying derivation outputs we need to compute their dependency closure (which is different from the closure of the output path).
  Currently there's a (fragile and probably not exact) logic to recompute it on-the-fly for the local store, but we should rather store all the dependencies of a drv output in a new table in the db

/cc @edolstra @Ericson2314

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -292,22 +292,26 @@ void DerivationGoal::haveDerivation()
 
     parsedDrv = std::make_unique<ParsedDerivation>(drvPath, *drv);
 
-
     /* We are first going to try to create the invalid output paths
        through substitutes.  If that doesn't work, we'll build
        them. */
     if (settings.useSubstitutes && parsedDrv->substitutesAllowed())
-        for (auto & [_, status] : initialOutputs) {
-            if (!status.wanted) continue;
-            if (!status.known) {
-                warn("do not know how to query for unknown floating content-addressed derivation output yet");
-                /* Nothing to wait for; tail call */
-                return DerivationGoal::gaveUpOnSubstitution();
-            }
+        for (auto& [outputName, status] : initialOutputs) {
+            if (!status.wanted)
+                continue;
             addWaitee(upcast_goal(worker.makeSubstitutionGoal(
-                status.known->path,
+                DrvOutputId{drvPath, outputName},
                 buildMode == bmRepair ? Repair : NoRepair,
                 getDerivationCA(*drv))));
+            if (status.known) {
+                // If we know the output path, also depend on directly fetching
+                // it as the remote cache might have the output path but not
+                // know of the output id
+                addWaitee(upcast_goal(worker.makeSubstitutionGoal(
+                    status.known->path,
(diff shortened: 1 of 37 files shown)


src/libstore/build/derivation-goal.cc
```diff
@@ -292,22 +292,28 @@ void DerivationGoal::haveDerivation()
...
-        for (auto & [_, status] : initialOutputs) {
-            if (!status.wanted) continue;
-            if (!status.known) {
-                warn("do not know how to query for unknown floating content-addressed derivation output yet");
-                /* Nothing to wait for; tail call */
-                return DerivationGoal::gaveUpOnSubstitution();
+        for (auto& [outputName, status] : initialOutputs) {
+            if (!status.wanted)
+                continue;
+            auto drvSubstitution = getEnv("USE_DRV_SUBSTITUTION");
```
zimbatm, line 302: Is it possible to use a setting instead? The proliferation of environment variables that changes the behaviour of Nix can be problematic.

regnat:
Oops, that was just some debugging code that wasn't supposed to be pushed 😇

There shouldn't be a need for such an option once it works properly because the new behavior fallbacks on the old one as needed

Issue #4170, opened 2020-10-20
Clarify relationship between nix develop and build


mkenigs:
Was thinking about the difference between `develop` and `build` because running `nix develop --phase build` feels like it should be part of the `build` command, eg `nix build --phase build`. The fact that `--phase` causes `develop` to exit the bash shell makes it feel like it isn't really doing what the help text describes: "run a bash shell." I know currently there's differences between `develop`, `build`, and `run`, but as a user it kinda feels like `build` should be an alias for `develop && run every stdenv phase` and `run` should be an alias for `build && ./result/bin/binary`. Personally, I'd also find it more convenient for development if `nix build` preserved state in my source directory somehow rather than rebuilding everything, so that I could just run `nix build` when I change one file instead of `nix develop`, `make`, and `make install`.

Maybe the current complexity is necessary, but would it be possible to simplify the differences between `develop` and `build`? And then could `--phase` be moved to the `build` command?

Seems like that could also simplify some of the differences in implementation of `develop`, `build`, and `run`.

Ericson2314:
`nix build` is a core part of nix, `nix develop` is a command that tries to do various helpful things based on Nixpkgs and which I think would be better as a separate tool so it and Nixpkgs's `stdenv` can evolve in tandem.

mkenigs:
I think that's what I find confusing. It seems like thinking of `build` as a development tool is appropriate (https://nixos.org/manual/nix/unstable/hacking.html), and for a new user I think the most obvious assumption is that `develop` is a tool to drop you into the environment that `build` uses. The help text seems to indicate that. But instead, to understand the difference between the two I need to understand core nix functionality as well as `stdenv`. `build` is one of the 8 "Main commands" for `nix` and I think it would be more beginner friendly if understanding those commands didn't require understanding of more complex nix concepts.

Maybe I'm wanting to abuse `build` or overlooking the importance of those differences, but I don't think it's beginner friendly.

Ericson2314:
@mkenigs To be clear, what I'm saying is that `nix develop` is (imo) a gross layer violation as `nix-shell` was before it, and having something like `nix build --phase build` would make it worse. But i think are we in agreement that if `nixpkgs build` was some separate tool, then there wouldn't be the same expectation for `nix develop` to work like it.

mkenigs:
[...]

PR #4176, opened 2020-10-21
nix develop --phase: chdir to flake directory

### LINKED ISSUE #3976: nix develop --phase: Chdir to flake directory
E.g. `nix develop --build <flake-dir>` should be executed in `<flake-dir>` because that's the directory that contains the Makefile.

mkenigs:
For git+file and path flakes, chdir to flake directory so that phases
that expect to be in the flake directory can run

Fixes #3976

Personally I feel like this is kind of hacky to make exceptions for just git+file and path flakes, which is part of why I think what develop does should be clarified (#4170) But it is convenient for this to work

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -9,6 +9,7 @@
 #include "progress-bar.hh"
 #include "run.hh"
 
+#include <memory>
 #include <nlohmann/json.hpp>
 
 using namespace nix;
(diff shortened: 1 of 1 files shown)


src/nix/develop.cc
```diff
@@ -408,9 +409,34 @@ struct CmdDevelop : Common, MixEnvironment
...
+                auto type = getStrAttr(attrs, "type");
+                std::string flakeDir;
+                if (type == "git") {
+                    auto url = parseURL(getStrAttr(attrs, "url"));
+                    if (url.scheme == "file") {
+                        flakeDir = url.path;
+                    }
+                } else if (type == "path") {
+                    flakeDir = getStrAttr(attrs, "path");
+                }
```
regnat, line 430: I think this can be replaced by `installableFlake.getLockedFlake()->flake.resolvedRef.input.getSourcePath()` (or something of that taste)

mkenigs: Haven't tested it but changed to use `getSourcePath`

mkenigs:
@regnat this look good now?

edolstra:
@mkenigs Thanks!

PR #4180, opened 2020-10-22
Allow substituting paths when building remotely using `ssh-ng://`

### LINKED ISSUE #2770: builders-use-substitutes only implemented for LegacySSHStore?
I have been noticing that our remote builds never seem to pull from the central artifact cache, preferring instead to have the local builder upload the required store paths, which is much slower.  The relevant config option is `builders-use-substitutes`, but as far as I can tell, this is only implemented for LegacySSHStore:

- The flag winds up in buildersUseSubstitutes.
- This is only used in build-remote.cc _main to pass Substitute to copyPaths
- copyPaths passes it to Store::queryValidPaths

That is implemented by:

- LegacySSHStore::queryValidPaths - passes it to a Connection::get()
- LocalStore::queryValidPaths - ignores it
- RemoteStore::queryValidPaths - ignores it
- Store::queryValidPaths - ignores it
- SSHStore doesn't implement queryValidPaths, so it must come from RemoteStore::queryValidPaths, which ignores it.

The implication is that if you use ssh:// in the `builders` option, then the option should work, but if you use ssh-ng://, then it will be ignored.  I did an experiment, and that seems to be the case: I always get a `copying path ... to 'ssh-ng://...'`, but if I clear out the built store paths, then use ssh:// for same operation, I skip all the `copying ... to`.  The remote nix-daemon doesn't seem to log about its copying so as far as I can tell there's no indication about what it did, but the missing "copying to" suggests that it's working as intended.

Is this a known omission?  Is it better to use ssh:// for now?

Ma27:
Until now, it was not possible to substitute missing paths from e.g.
`https://cache.nixos.org` on a remote server when building on it using
the new `ssh-ng` protocol.

This is because every store implementation except legacy `ssh://`
ignores the substitution flag passed to `Store::queryValidPaths` while
the `legacy-ssh-store` substitutes the remote store using
`cmdQueryValidPaths` when the remote store is opened with `nix-store
--serve`.

This patch slightly modifies the daemon protocol to allow passing an
integer value suggesting whether to substitute missing paths during
`wopQueryValidPaths`. To implement this on the daemon-side, the
substitution logic from `nix-store --serve` has been moved into a
protected method named `Store::substitutePaths` which gets currently
called from `LocalStore::queryValidPaths` and `Store::queryValidPaths`
if `maybeSubstitute` is `true`.

Fixes #2770

cc @elaforge @edolstra @Ericson2314 @lheckemann

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -274,8 +274,17 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
 
     case wopQueryValidPaths: {
         auto paths = worker_proto::read(*store, from, Phantom<StorePathSet> {});
+
+        SubstituteFlag substitute = NoSubstitute;
+        if (GET_PROTOCOL_MINOR(clientVersion) >= 27) {
+            substitute = readInt(from) ? Substitute : NoSubstitute;
+        }
+
         logger->startWork();
-        auto res = store->queryValidPaths(paths);
+        if (substitute) {
+            store->substitutePaths(paths);
+        }
+        auto res = store->queryValidPaths(paths, substitute);
         logger->stopWork();
         worker_proto::write(*store, to, res);
         break;
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,7 @@
       versionSuffix =
         if officialRelease
         then ""
-        else "pre${builtins.substring 0 8 (self.lastModifiedDate or self.lastModified)}_${self.shortRev or "dirty"}";
+        else "pre${builtins.substring 0 8 (self.lastModifiedDate or self.lastModified or "19700101")}_${self.shortRev or "dirty"}";
 
       officialRelease = false;
 
@@ -117,6 +117,7 @@
 
         nix = with final; with commonDeps pkgs; (stdenv.mkDerivation {
           name = "nix-${version}";
+          inherit version;
 
           src = self;
 
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -259,6 +259,9 @@ StorePathSet RemoteStore::queryValidPaths(const StorePathSet & paths, Substitute
     } else {
         conn->to << wopQueryValidPaths;
         worker_proto::write(*this, conn->to, paths);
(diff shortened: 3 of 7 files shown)


Ericson2314 (approved):
Excellent change!

src/libstore/daemon.cc
```diff
@@ -276,8 +276,13 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
 
     case wopQueryValidPaths: {
         auto paths = worker_proto::read(*store, from, Phantom<StorePathSet> {});
+        SubstituteFlag substitute = NoSubstitute;
+        if (GET_PROTOCOL_MINOR(clientVersion) >= 27) {
+            substitute = readInt(from) ? Substitute : NoSubstitute;
+        }
         logger->startWork();
-        auto res = store->queryValidPaths(paths);
+
```
Ericson2314, line 284: Not to be a nit-pick but I would rather not put a new blank line here because the `startWork`, `queryValidPaths`, and `stopWork` very much go together. Maybe if you want to add blank lines they could go before and after that group?

Ma27: Fixed :)

Ma27:
@edolstra does this look OK to you as well? :)

lheckemann (approved):

Ma27:
ping @edolstra anything missing to get this merged? %)

edolstra:
`maybeSubstitute` is a pretty ugly hack and it would probably be better to get rid of it somehow rather than add it to other stores. For instance, `nix copy` could call `destStore->substitutePaths()` (or `destStore->buildPaths()`) directly.

Ma27:
@edolstra so you mean that the daemon (and store) workers should invoke `destStore->substitutePaths()` directly if needed?

Ma27:
@edolstra the `Store::substitutePaths()` method is now called from `nix-daemon` directly. Anything else to fix to get this merged? :)

Ma27:
ping @edolstra anything else tbd here to get this merged? :)

Ma27:
@edolstra any updates here? :)

edolstra:
Thanks!

PR #4181, opened 2020-10-23
darwin: encrypt nix volume if filevault is enabled

abathur:
# Motive/Goal 
Get single-invocation `curl -L https://nixos.org/nix/install | sh` installs working on macOS Catalina+

---
Recap in case it helps anyone:
- To support read-only / in Catalina, we updated the installer to move /nix to its own volume. 
- We wanted to encrypt this volume when FileVault FDE was enabled on the system to meet the user's expectation that their data was encrypted, but didn't find an approach we could confidently recommend.
- To sate the Nix-starved mob, we sliced off all of the cases we felt like we could handle, hid them behind the `--darwin-use-unencrypted-nix-store-volume` scare-flag to make sure people with Opinions and/or compliance concerns would pay attention, and left the remainder (non-T2 devices with FileVault enabled) to manually prepare their volumes.
---
We've recently confirmed that we can skirt the race condition by using a RunAtLoad LaunchDaemon to unlock and mount the volume. (I've tested straightforward cases like restoring editors with open files on the /nix volume, and restoring .apps managed by Nix. It may still be possible to force race conditions with multiple RunAtLoad daemons, but I think it's OK to work around those as-needed with `wait4path`.)

## Status
- PR currently works
    - tested locally with FileVault enabled on Big Sur beta using both single & multi-user installs
    - install [working in CI](https://github.com/abathur/syspolicyd_assessments/actions/runs/326476454) on matrix of FileVault yes/no, Catalina/Big Sur, --daemon ~--no-daemon~
- ~To finish the code out, we need to haggle over how this should affect single-user mode and "default" macOS installs. (see upcoming 2nd post for more on this):~ blocked single-user macOS installs
- To finish the code out, we need to haggle over the curing/uninstall concept?
- Need doc edits once code is baked.

## Try it out?
I have a temporary installer up at https://files.t-ravis.com/install-volume, so `curl https://files.t-ravis.com/install-volume | sh` should work. If you can't try it out, you can see output from the [CI run (same one listed earlier)](https://github.com/abathur/syspolicyd_assessments/actions/runs/326476454). I refresh it to test larger functionality changes, but it may be a few commits behind if they don't merit the chore of regenerating. Here's a history of the versions it has pointed at, with the current at the bottom:
- ~4e3a355~ 
- ~1668c39 (broken)~ 
- ~b9657d5~ 
- ~c95ce48~ 
- ~366cb2b (install OK but service broken)~ 
- ~eab5cb2 (install OK but service broken)~
- ~97d946a (install OK but service broken)~
- ~0c565a2~
- ~6541c0d~
- ~385994d~
- 0f73713

@LnL7 @lilyball @matthewbauer @dhess @mroi @domenkozar (others I'm forgetting are inevitably also interested; feel free to add people)

### DIFF
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -100,14 +100,12 @@ prepare_darwin_volume_password(){
     # can tell, the file with this password (/var/db/SystemKey) is
     # inside the FileVault envelope. If that isn't true, it may make
     # sense to store the password inside the envelope?
-    sudo /usr/bin/security add-generic-password -a "$1" -s "$2" -l "$1 encryption password" -D "Encrypted volume password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" -T /System/Library/CoreServices/APFSUserAgent -T /System/Library/CoreServices/CSUserAgent "/Library/Keychains/System.keychain" &>/dev/null
-    # TODO: decide if we should add `-T /System/Library/CoreServices/APFSUserAgent`
-    # This should let the system seamlessly supply the password for this volume
-    # which in turn means the fstab entry is enough for the system to (eventually)
-    # decrypt and mount the volume we're adding, but I hesitate because I'm not
-    # certain the system _should_ transparently failover if the LaunchDaemon is
-    # broken for some reason? Without supplying this flag, the system will instead
-    # start prompting them to allow APFSUserAgent to access this credential.
+    sudo /usr/bin/security add-generic-password -a "$1" -s "$2" -l "$1 encryption password" -D "Encrypted volume password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" -T /System/Library/CoreServices/APFSUserAgent -T /System/Library/CoreServices/CSUserAgent -T /usr/bin/security "/Library/Keychains/System.keychain" &>/dev/null
+    # TODO: /usr/bin/security could be replaced with our own binary at some point?
+    # *UserAgent exemptions should let the system seamlessly supply the password
+    # if noauto is removed from the fstab entry. This is intentional, so that
+    # the user will hopefully look for help if the volume stops mounting,
+    # rather than failing over into subtle race-condition problems.
 
     # 2. add a password with the -U (update) flag and -w (prompt if last)
     #    flags, but specify no keychain; security will use the first it finds
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -e
 
+# make it easy to play w/ 'Case-sensitive APFS'
+readonly NIX_VOLUME_FS="${NIX_VOLUME_FS:-APFS}"
+
 # i.e., "disk1"
 root_disk_identifier() {
     diskutil info -plist / | xmllint --xpath "/plist/dict/key[text()='ParentWholeDisk']/following-sibling::string[1]/text()" -
@@ -187,7 +190,7 @@ main() {
     if [ -n "$create_volume" ]; then
         echo "Creating a Nix volume..." >&2
 
-        sudo diskutil apfs addVolume "$disk" APFS "$volume" -mountpoint /nix
+        sudo diskutil apfs addVolume "$disk" "$NIX_VOLUME_FS" "$volume" -mountpoint /nix
         new_uuid="$(volume_uuid "$volume")"
 
         if [ "$INSTALL_MODE" = "no-daemon" ]; then # exported by caller
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -6,6 +6,10 @@ root_disk_identifier() {
     diskutil info -plist / | xmllint --xpath "/plist/dict/key[text()='ParentWholeDisk']/following-sibling::string[1]/text()" -
 }
 
+volume_uuid(){
+    diskutil info -plist "$1" | xmllint --xpath "(/plist/dict/key[text()='VolumeUUID']/following-sibling::string/text())" -
+}
+
 find_nix_volume() {
     diskutil apfs list -plist "$1" | xmllint --xpath "(/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[starts-with(translate(text(),'N','n'),'nix')]/text())[1]" - 2>/dev/null || true
 }
@@ -180,6 +184,13 @@ main() {
     if [ -n "$create_volume" ]; then
         echo "Creating a Nix volume..." >&2
 
+        sudo diskutil apfs addVolume "$disk" APFS "$volume" -mountpoint /nix
+
+        if [ "$INSTALL_MODE" = "no-daemon" ]; then # exported by caller
+            # TODO: is there a better way to do this?
+            sudo chown $USER:admin /nix
+        fi
+
         if test_filevault_in_use; then
             # security program's flags won't let us both specify a keychain
             # and be prompted for a pw to add; two step workaround:
@@ -191,18 +202,18 @@ main() {
             # can tell, the file with this password (/var/db/SystemKey) is
             # inside the FileVault envelope. If that isn't true, it may make
             # sense to store the password inside the envelope?
-            sudo /usr/bin/security add-generic-password -a "Nix Volume" -s "Nix Volume" -D "Nix Volume password" "/Library/Keychains/System.keychain"
+            sudo /usr/bin/security add-generic-password -a "$volume" -s "$(volume_uuid "$volume")" -D "$volume encryption password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" "/Library/Keychains/System.keychain"
+            # TODO: decide if we should add `-T /System/Library/CoreServices/APFSUserAgent`
+            # This should let the system seamlessly supply the password for this volume
+            # which in turn means the fstab entry is enough for the system to (eventually)
+            # decrypt and mount the volume we're adding, but I hesitate because I'm not
+            # certain the system _should_ transparently failover if the LaunchDaemon is
+            # broken for some reason? Without supplying this flag, the system will instead
+            # start prompting them to allow APFSUserAgent to access this credential.
 
             # 2. add a password with the -U (update) flag and -w (prompt if last)
             #    flags, but specify no keychain; security will use the first it finds
-            prepare_darwin_volume_password | sudo diskutil apfs addVolume "$disk" APFS 'Nix Volume' -mountpoint /nix -stdinpassphrase
-        else
-            sudo diskutil apfs addVolume "$disk" APFS 'Nix Volume' -mountpoint /nix
-        fi
-
-        if [ "$INSTALL_MODE" = "no-daemon" ]; then # exported by caller
-            # TODO: is there a better way to do this?
-            sudo chown $USER:admin /nix
+            prepare_darwin_volume_password | sudo diskutil apfs encryptVolume "$volume" -user disk -stdinpassphrase
         fi
     else
         echo "Using existing '$volume' volume" >&2
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -120,9 +120,10 @@ main() {
         echo "     ------------------------------------------------------------------ "
         echo ""
         echo "  1. Remove the entry from fstab using 'sudo vifs'"
-        echo "  2. Remove /Library/LaunchDaemons/org.nixos.darwin-store.plist"
-        echo "  3. Destroy the data volume using 'diskutil apfs deleteVolume'"
-        echo "  4. Remove the 'nix' line from /etc/synthetic.conf (or the file)"
+        echo "  2. Run `sudo launchctl bootout system/org.nixos.darwin-store`
+        echo "  3. Remove /Library/LaunchDaemons/org.nixos.darwin-store.plist"
+        echo "  4. Destroy the data volume using 'diskutil apfs deleteVolume'"
+        echo "  5. Remove the 'nix' line from /etc/synthetic.conf (or the file)"
         echo ""
     ) >&2
 
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -120,7 +120,7 @@ main() {
         echo "     ------------------------------------------------------------------ "
         echo ""
         echo "  1. Remove the entry from fstab using 'sudo vifs'"
-        echo "  2. Run `sudo launchctl bootout system/org.nixos.darwin-store`
+        echo "  2. Run `sudo launchctl bootout system/org.nixos.darwin-store`"
         echo "  3. Remove /Library/LaunchDaemons/org.nixos.darwin-store.plist"
         echo "  4. Destroy the data volume using 'diskutil apfs deleteVolume'"
         echo "  5. Remove the 'nix' line from /etc/synthetic.conf (or the file)"
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -62,7 +62,7 @@ generate_mount_command(){
     if test_filevault_in_use; then
         printf "    <string>%s</string>\n" /bin/sh -c '/usr/bin/security find-generic-password -l "Nix Volume" -a "Nix Volume" -s "Nix Volume" -w | /usr/sbin/diskutil apfs unlockVolume "Nix Volume" -mountpoint /nix -stdinpassphrase'
     else
-        printf "    <string>%s</string>\n" /usr/sbin/diskutil mount nobrowse -mountPoint /nix "Nix Volume"
+        printf "    <string>%s</string>\n" /usr/sbin/diskutil mount -mountPoint /nix "Nix Volume"
     fi
 }
 
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -97,7 +97,7 @@ set PASSPHRASE [exec /usr/bin/xxd -l 32 -p -c 256 /dev/random]
 
 # Cargo culting: people recommend this; not sure how necessary
 set send_slow {1 .1}
-spawn /usr/bin/sudo /usr/bin/security add-generic-password -a "$VOLUME" -l "$volume encryption password" -s "$UUID" -D "Encrypted volume password" -U -w
+spawn /usr/bin/sudo /usr/bin/security add-generic-password -a "$VOLUME" -l "$VOLUME encryption password" -s "$UUID" -D "Encrypted volume password" -U -w
 expect {
     "password data for new item: " {
         send -s -- "$PASSPHRASE\r"
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -201,7 +201,7 @@ main() {
             # can tell, the file with this password (/var/db/SystemKey) is
             # inside the FileVault envelope. If that isn't true, it may make
             # sense to store the password inside the envelope?
-            sudo /usr/bin/security add-generic-password -a "$volume" -s "$new_uuid" -D "$volume encryption password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" "/Library/Keychains/System.keychain"
+            sudo /usr/bin/security add-generic-password -a "$volume" -s "$new_uuid" -l "$volume encryption password" -D "Encrypted volume password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" "/Library/Keychains/System.keychain"
             # TODO: decide if we should add `-T /System/Library/CoreServices/APFSUserAgent`
             # This should let the system seamlessly supply the password for this volume
             # which in turn means the fstab entry is enough for the system to (eventually)
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -204,6 +204,11 @@ main() {
         else
             sudo diskutil apfs addVolume "$disk" APFS 'Nix Volume' -mountpoint /nix
         fi
+
+        if [ "$INSTALL_MODE" = "no-daemon" ]; then
+            # TODO: is there a better way to do this?
+            sudo chown $USER:admin /nix
+        fi
     else
         echo "Using existing '$volume' volume" >&2
     fi
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -86,7 +86,7 @@ generate_mount_daemon(){
   <string>org.nixos.darwin-store</string>
   <key>ProgramArguments</key>
   <array>
-    $(generate_mount_command)
+$(generate_mount_command)
   </array>
 </dict>
 </plist>
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -1,34 +1,128 @@
-#!/bin/sh
-set -e
+#!/usr/bin/env bash
+set -eu
+
+# support import-only `. create-darwin-volume.sh no-main[ ...]`
+if [ "${1-}" = "no-main" ]; then
+    shift
+    readonly _CREATE_VOLUME_NO_MAIN=1
+else
+    readonly _CREATE_VOLUME_NO_MAIN=0
+    # declare some things we expect to inherit from install-multi-user
+    # I don't love this (because it's a bit of a kludge)
+    readonly NIX_ROOT="${NIX_ROOT:-/nix}"
+    readonly ESC='\033[0m'
+    readonly GREEN='\033[32m'
+    readonly RED='\033[31m'
+    _sudo() {
+        shift # throw away the 'explanation'
+        /usr/bin/sudo "$@"
+    }
+    failure() {
+        if [ "$*" = "" ]; then
+            cat
+        else
+            echo "$@"
+        fi
+        exit 1
+    }
+    task() {
+        echo "$@"
+    }
+fi
 
 # make it easy to play w/ 'Case-sensitive APFS'
 readonly NIX_VOLUME_FS="${NIX_VOLUME_FS:-APFS}"
-
-readonly NIX_VOLUME_MOUNTD="${NIX_VOLUME_MOUNTD:-/Library/LaunchDaemons/org.nixos.darwin-store.plist}"
+readonly NIX_VOLUME_LABEL="${NIX_VOLUME_LABEL:-Nix}"
+# shellcheck disable=SC1003,SC2026
+readonly NIX_VOLUME_FOR_FSTAB="${NIX_VOLUME_LABEL/ /'\\\'040}"
+readonly NIX_VOLUME_MOUNTD_DEST="${NIX_VOLUME_MOUNTD_DEST:-/Library/LaunchDaemons/org.nixos.darwin-store.plist}"
 
 # i.e., "disk1"
 root_disk_identifier() {
     /usr/sbin/diskutil info -plist / | xmllint --xpath "/plist/dict/key[text()='ParentWholeDisk']/following-sibling::string[1]/text()" -
 }
 
-volume_uuid(){
-    /usr/sbin/diskutil info -plist "$1" | xmllint --xpath "(/plist/dict/key[text()='VolumeUUID']/following-sibling::string/text())" -
+readonly ROOT_SPECIAL_DEVICE="$(root_disk_identifier)" # usually 'disk1'
+# Strongly assuming we'll make a volume on the device root is on
+# But you can override NIX_VOLUME_USE_DISK to create it on some other device
+readonly NIX_VOLUME_USE_DISK="${NIX_VOLUME_USE_DISK:-$ROOT_SPECIAL_DEVICE}"
+
+substep(){
+    # shellcheck disable=SC2068
+    printf "   %s\n" "" "- $1" "" ${@:2}
 }
 
+printf -v _UNCHANGED_GRP_FMT "%b" $'\033[2m%='"$ESC" # "dim"
+# bold+invert+red and bold+invert+green just for the +/- below
+# red/green foreground for rest of the line
+printf -v _OLD_LINE_FMT "%b" $'\033[1;7;31m-'"$ESC ${RED}%L${ESC}"
+printf -v _NEW_LINE_FMT "%b" $'\033[1;7;32m+'"$ESC ${GREEN}%L${ESC}"
+_diff() {
+    # shellcheck disable=SC2068
+    /usr/bin/diff --unchanged-group-format="$_UNCHANGED_GRP_FMT" --old-line-format="$_OLD_LINE_FMT" --new-line-format="$_NEW_LINE_FMT" --unchanged-line-format="  %L" $@
+}
+
+confirm_rm() {
+    if ui_confirm "Can I remove $1?"; then
+        # ok "Yay! Thanks! Let's get going!"
+        _sudo "to remove $1" rm "$1"
+    fi
+}
+confirm_edit() {
+    cat <<EOF
+It looks like Nix isn't the only thing here, but I think I know how to edit it
+out. Here's the diff:
+EOF
+
+    # could technically test the diff, but caller should do it
+    _diff "$1" "$2"
+    if ui_confirm "Does the change above look right?"; then
+        # ok "Yay! Thanks! Let's get going!"
+        _sudo "remove nix from $1" cp "$2" "$1"
+    fi
+}
+
+volume_uuid(){
+    /usr/sbin/diskutil info -plist "$1" | xmllint --xpath "(/plist/dict/key[text()='VolumeUUID']/following-sibling::string[1]/text())" - 2>/dev/null
+}
+
+volume_special_device(){
+    /usr/sbin/diskutil info -plist "$1" | xmllint --xpath "(/plist/dict/key[text()='DeviceIdentifier']/following-sibling::string[1]/text())" - 2>/dev/null
+}
 find_nix_volume() {
+    /usr/sbin/diskutil apfs list -plist "$1" | xmllint --xpath "(/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[text()='$NIX_VOLUME_LABEL'])[1]" - &>/dev/null
+}
+find_nix_volume_label() {
     /usr/sbin/diskutil apfs list -plist "$1" | xmllint --xpath "(/plist/dict/array/dict/key[text()='Volumes']/following-sibling::array/dict/key[text()='Name']/following-sibling::string[starts-with(translate(text(),'N','n'),'nix')]/text())[1]" - 2>/dev/null || true
 }
 
 test_fstab() {
-    /usr/bin/grep -q "/nix apfs rw" /etc/fstab 2>/dev/null
+    /usr/bin/grep -q "$NIX_ROOT apfs rw" /etc/fstab 2>/dev/null
 }
 
 test_nix_symlink() {
-    [ -L "/nix" ] || /usr/bin/grep -q "^nix." /etc/synthetic.conf 2>/dev/null
+    [ -L "$NIX_ROOT" ] || /usr/bin/grep -q "^nix." /etc/synthetic.conf 2>/dev/null
+}
+test_synthetic_conf_mountable() {
+    /usr/bin/grep -q "^${NIX_ROOT:1}$" /etc/synthetic.conf 2>/dev/null
+}
+test_synthetic_conf_symlinked() {
+    /usr/bin/grep -qE "^${NIX_ROOT:1}\s+\S{3,}" /etc/synthetic.conf 2>/dev/null
 }
 
-test_synthetic_conf() {
-    /usr/bin/grep -q "^nix$" /etc/synthetic.conf 2>/dev/null
+test_nix_volume_mountd_installed() {
+    test -e "$NIX_VOLUME_MOUNTD_DEST"
+}
+
+# true here is a cruft smell when ! test_keychain_by_uuid
+test_keychain_by_label() {
+    # Note: doesn't need sudo just to check; doesn't output pw
+    security find-generic-password -a "$NIX_VOLUME_LABEL" &>/dev/null
+}
+# current volume password
+test_keychain_by_uuid() {
+    # Note: doesn't need sudo just to check; doesn't output pw
+    security find-generic-password -s "$(volume_uuid "$NIX_VOLUME_LABEL")" &>/dev/null
 }
 
 # Create the paths defined in synthetic.conf, saving us a reboot.
@@ -46,33 +140,27 @@ create_synthetic_objects(){
 }
 
 test_nix() {
-    test -d "/nix"
+    test -d "$NIX_ROOT"
 }
 
 test_voldaemon() {
-    test -f "$NIX_VOLUME_MOUNTD"
+    test -f "$NIX_VOLUME_MOUNTD_DEST"
 }
 
 test_filevault_in_use() {
     /usr/bin/fdesetup isactive >/dev/null
 }
 
-# use after error msg for conditions we don't understand
-suggest_report_error(){
-    # ex "error: something sad happened :(" >&2
-    echo "       please report this @ https://github.com/nixos/nix/issues" >&2
-}
-
 generate_mount_command(){
-    if test_filevault_in_use; then
-        printf "    <string>%s</string>\n" /bin/sh -c '/usr/bin/security find-generic-password -a "Nix" -w | /usr/sbin/diskutil apfs unlockVolume "Nix" -mountpoint /nix -stdinpassphrase'
+    if [ "${1-}" = "" ]; then
+        printf "    <string>%s</string>\n" /bin/sh -c "/usr/bin/security find-generic-password -s '$1' -w | /usr/sbin/diskutil apfs unlockVolume '$NIX_VOLUME_LABEL' -mountpoint $NIX_ROOT -stdinpassphrase"
     else
-        printf "    <string>%s</string>\n" /usr/sbin/diskutil mount -mountPoint /nix "Nix"
+        printf "    <string>%s</string>\n" /usr/sbin/diskutil mount -mountPoint "$NIX_ROOT" "$NIX_VOLUME_LABEL"
     fi
 }
 
 generate_mount_daemon(){
-    cat << EOF
+    cat <<EOF
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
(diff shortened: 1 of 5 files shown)


abathur:
I could use help hammering out an issue with as many nix+macOS stakeholders as we can round up on short notice.

I have opinions here, but my overriding interest is finding the quickest path to single-invocation macOS Nix installs for Catalina and Big Sur that Just Work without making prospective users crawl through glass anymore. I'm going to try to sketch out the facts and and opinions I'm already aware of without staking anyone down by naming names. Please point out anything I'm missing!

## Question
Should this approach apply to single-user installs, or not? (and, in either case, how)

## Context
My rough understanding is that single-user installs are intended to be simpler, work even if Nix doesn't support the service manager, install without root, (and be simple to uninstall?)

Starting with Catalina, I think it's fair to characterize both single and multi-user macOS installs as already violating:
- the "install without root" goal (AFAIR, we currently need sudo to edit /etc/synthetic.conf and /etc/fstab, figure out if the device has a T2 chip, and create/mount the volume.)
- the "simple to uninstall" goal (the intuitive `rm -rf /` isn't enough because the system root is read-only; you really need to do something like remove nix from fstab and synthetic.conf, remove the Nix volume, and reboot)

The PR as it stands:
- adds a few additional sudo invocations to interact with the `security` and `launchctl` commands, write the LaunchDaemon, and a few more
- adds an "active" component (the LaunchDaemon); note: the current .plist "KeepAlive" setting attempts to run until a path we expect to be on the volume is available; this can and should be refined to limit the resources it can waste if the volume isn't mountable for some reason
- the user might fail to uninstall this active component (it may be possible to make it self-disable or self-destruct to avoid wasting resources if forgotten?)
- when FileVault is enabled, generate a random password, use it to encrypt the new volume, and insert this credential in the system keychain (where it could be read by other users with admin access)

There are also some more general concerns here, like minimizing the number of possible install variants.

## Solutions?
- include the LaunchDaemon mounter with multi-user installs; block single-user installs when FileVault is enabled (i.e., only support FileVault/FDE in daemon mode) and forward users who try it to the docs; change the default macOS install to multi-user mode
- include the LaunchDaemon mounter with both single and multi-user installs
- prompt the user (for both type of install?) to accept the LaunchDaemon (explain the purpose of the daemon, show them the full plist, clarify the consequences of not accepting it)
- entirely disable single-user installs for macOS

I'll edit the above to add any suggestions that don't boil down to one of the above.

domenkozar:
As per my NixCon talk `entirely disable single-user installs for macOS` is the way forward and you lay down the complexity of such division by just a small tiny portion of the installation logic.

edolstra:
@domenkozar I thought you were in favor of getting rid of multi-user installs?

> you lay down the complexity of such division by just a small tiny portion of the installation logic.

Hm, I have trouble parsing this...

domenkozar:
I am in favor of reducing complexity across whole Nix code base.

One of the factors is branching of logic per installation method. I'm in favor of merging them, if that's the pragmatic solution for macOS by defaulting to multi-user installations.

dhess:
> As per my NixCon talk `entirely disable single-user installs for macOS` is the way forward and you lay down the complexity of such division by just a small tiny portion of the installation logic.

I completely agree. I've been solely using multi-user installs on all of my Macs for the past few years. It works reliably and is more secure than single-user, to boot.

dhess:
@abathur Thank you for all of your work on this. It's a (mostly) thankless task and full of corner cases, but greatly appreciated.

lilyball:
I was actually planning on looking into this problem this very week, as I'm currently working on figuring out how to deploy Nix to my team 😅 I have not yet looked at what this PR does but here are my immediate thoughts:

1. Last time I looked into this, it appeared that putting the volume passphrase into the system keychain and granting access to a few specific processes allowed the system to mount it automatically upon boot, just as it mounts the volume when unencrypted. This is mentioned [in this PR comment](https://github.com/NixOS/nix/pull/3692#issuecomment-648886321) and the reference for this is [this third-party install script](https://gist.github.com/burke/1e750ef40bd251b72213e4f4a4818668#file-apfs_volume-rb-L176-L193). According to that script, the keychain entry needs to be accessible by `APFSUserAgent` and `CSUserAgent`. If this approach just works, that's fantastic. My worry here is just that the need to grant access to 2 specific daemons is undocumented and it's not clear if this will be reliable going forward.
2. I like the convenience of single-user as I don't have to use `sudo` to update my channels, and the conceptual simplicity of only having one profile instead of having default + per-user is nice. I worry that when introducing Nix to colleagues, a multi-user install might lead them to accidentally installing some stuff in the default profile and some in the per-user profile (though my intended usage of Nix does not have them explicitly installing anything at all). Having said that, the fact that there are two different ways of installing Nix, and the fact that the shell setup is different for these, and the usage patterns are different, is annoying. Ultimately, it's probably a good idea to remove single-user install, given that we need `sudo` to install it anyway. Single-user install is probably only worth having if we have a sudoless installation (which is never going to happen as that means the installation path would be relative to the user's home folder and would break all binary caches).

lilyball:
Regarding the keychain thing, I wonder if someone can expend an Apple developer account incident to see if we can get in touch with an Apple filesystem engineer in order to confirm whether the "grant keychain access to these 2 processes" is reliable going forward, or whether there's any alternative supported mechanism.

lilyball:
> adds an "active" component (the LaunchDaemon); note: the current .plist "KeepAlive" setting attempts to run until a path we expect to be on the volume is available; this can and should be refined to limit the resources it can waste if the volume isn't mountable for some reason

What is the reason for using `KeepAlive` here? Do we actually expect it to fail to mount a few times before it successfully mounts? With it set up this way, if I manually unmount the volume it will be expected to immediately remount it, which seems rather odd.

scripts/create-darwin-volume.sh
```diff
@@ -83,6 +61,64 @@ suggest_report_error(){
...
+  <array>
+$(generate_mount_command)
+  </array>
+</dict>
+</plist>
+EOF
+}
+
+prepare_darwin_volume_password(){
+    sudo /usr/bin/expect -f - "$1" "$2" << 'EOF'
```
lilyball, line 100: `tclsh` is deprecated on macOS. Is `/usr/bin/expect` reliable going forward? I hadn't thought of this before, `expect` requires `tclsh`, but it doesn't print the deprecation warning `tclsh` does upon entering the REPL, which could just be an oversight on Apple's part, or it could indicate that `expect` will stay even as `tclsh` is deprecated or removed from `PATH`.

scripts/create-darwin-volume.sh
```diff
@@ -83,6 +61,64 @@ suggest_report_error(){
...
+log_user 0
+set VOLUME [lindex $argv 0];
+set UUID [lindex $argv 1];
+set PASSPHRASE [exec /usr/bin/ruby -rsecurerandom -e "puts SecureRandom.hex(32)"]
```
lilyball, line 104: `/usr/bin/ruby` is deprecated on macOS (all scripting languages besides bash are; python3 is available if and only if Xcode is installed).

I think bash version of this hex generation could look like

```sh
cat /dev/random | head -c 32 | xxd -p -c 256
```

scripts/create-darwin-volume.sh
```diff
@@ -83,6 +61,64 @@ suggest_report_error(){
...
+set send_slow {1 .1}
+spawn /usr/bin/sudo /usr/bin/security add-generic-password -a "$VOLUME" -s "$UUID" -D "$VOLUME encryption password" -U -w
+expect {
+    "password data for new item: " {
+        send -s -- "$PASSPHRASE\r"
+        expect "retype password for new item: " {
+            send -s -- "$PASSPHRASE\r"
+        }
+    }
+}
```
lilyball, line 116: I assume we're using `expect` here purely to avoid having the password show up in the command-line arguments in case anyone is sniffing it?

I don't think we need to worry about that. The volume is going to be mounted the entire time the computer is running, so any process that can sniff the command-line arguments can also just look at the volume while it's mounted anyway. We should ditch the complexity of `expect` (and the risk that it might be deprecated too since `tclsh` is) and just provide the passphrase on the command-line.

scripts/create-darwin-volume.sh
```diff
@@ -92,8 +128,9 @@ main() {
         echo "     ------------------------------------------------------------------ "
         echo ""
         echo "  1. Remove the entry from fstab using 'sudo vifs'"
-        echo "  2. Destroy the data volume using 'diskutil apfs deleteVolume'"
-        echo "  3. Remove the 'nix' line from /etc/synthetic.conf or the file"
+        echo "  2. Remove /Library/LaunchDaemons/org.nixos.darwin-store.plist"
+        echo "  3. Destroy the data volume using 'diskutil apfs deleteVolume'"
+        echo "  4. Remove the 'nix' line from /etc/synthetic.conf (or the file)"
```
lilyball, line 133:
```suggestion
        echo "  2. Run `sudo launchctl bootout system/org.nixos.darwin-store`
        echo "  3. Remove /Library/LaunchDaemons/org.nixos.darwin-store.plist"
        echo "  4. Destroy the data volume using 'diskutil apfs deleteVolume'"
        echo "  5. Remove the 'nix' line from /etc/synthetic.conf (or the file)"
```

scripts/create-darwin-volume.sh
```diff
@@ -127,42 +164,75 @@ main() {
     fi
 
     disk="$(root_disk_identifier)"
-    volume=$(find_nix_volume "$disk")
+    volume=$(find_nix_volume "$disk") # existing volname starting w/ "nix"?
     if [ -z "$volume" ]; then
-        echo "Creating a Nix Store volume..." >&2
+        volume="Nix Volume"    # otherwise use default
```
lilyball, line 169: Previously this was `Nix Store`. I'm partial to just calling it `Nix`, but I think `Nix Store` makes more sense than `Nix Volume` anyway; it's already a volume, so putting "volume" in the name is redundant.

scripts/create-darwin-volume.sh
```diff
@@ -83,6 +61,64 @@ suggest_report_error(){
     echo "       please report this @ https://github.com/nixos/nix/issues" >&2
 }
 
+generate_mount_command(){
+    if test_filevault_in_use; then
+        printf "    <string>%s</string>\n" /bin/sh -c '/usr/bin/security find-generic-password -a "Nix Volume" -w | /usr/sbin/diskutil apfs unlockVolume "Nix Volume" -mountpoint /nix -stdinpassphrase'
+    else
+        printf "    <string>%s</string>\n" /usr/sbin/diskutil mount -mountPoint /nix "Nix Volume"
```
lilyball, line 68: If a volume exists starting with "Nix" already, we use that volume. This mount command needs to be parameterized accordingly.

scripts/create-darwin-volume.sh
```diff
@@ -127,42 +164,75 @@ main() {
...
+            # 1. add a blank pw to system keychain
+
+            # system is in some sense less secure than user keychain... (it's
+            # possible to read the password for decrypting the keychain) but
+            # the user keychain appears to be available too late. As far as I
+            # can tell, the file with this password (/var/db/SystemKey) is
+            # inside the FileVault envelope. If that isn't true, it may make
+            # sense to store the password inside the envelope?
+            sudo /usr/bin/security add-generic-password -a "$volume" -s "$new_uuid" -D "$volume encryption password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" "/Library/Keychains/System.keychain"
+            # TODO: decide if we should add `-T /System/Library/CoreServices/APFSUserAgent`
```
lilyball, line 213: According to the comment I left earlier, we should include `/System/Library/CoreServices/CSUserAgent` as well. I don't know under what conditions (if any) CoreStorage would be used on Catalina+, but maybe this is important for backwards-compatibility?

scripts/create-darwin-volume.sh
```diff
@@ -127,42 +164,75 @@ main() {
...
+            # and be prompted for a pw to add; two step workaround:
+            # 1. add a blank pw to system keychain
+
+            # system is in some sense less secure than user keychain... (it's
+            # possible to read the password for decrypting the keychain) but
+            # the user keychain appears to be available too late. As far as I
+            # can tell, the file with this password (/var/db/SystemKey) is
+            # inside the FileVault envelope. If that isn't true, it may make
+            # sense to store the password inside the envelope?
+            sudo /usr/bin/security add-generic-password -a "$volume" -s "$new_uuid" -D "$volume encryption password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" "/Library/Keychains/System.keychain"
```
lilyball, line 212: The label will default to the service name. We should provide an explicit label instead.

And the kind should probably be something generic instead of including the volume name.

Here's what I'd recommend:

```suggestion
            sudo /usr/bin/security add-generic-password -a "$volume" -s "$new_uuid" -l "$volume encryption password" -D "Encrypted volume password" -j "Added automatically by the Nix installer for use by /Library/LaunchDaemons/org.nixos.darwin-store.plist" "/Library/Keychains/System.keychain"
```

scripts/create-darwin-volume.sh
```diff
@@ -127,42 +164,75 @@ main() {
...
+            # This should let the system seamlessly supply the password for this volume
+            # which in turn means the fstab entry is enough for the system to (eventually)
+            # decrypt and mount the volume we're adding, but I hesitate because I'm not
+            # certain the system _should_ transparently failover if the LaunchDaemon is
+            # broken for some reason? Without supplying this flag, the system will instead
+            # start prompting them to allow APFSUserAgent to access this credential.
+
+            # 2. add a password with the -U (update) flag and -w (prompt if last)
+            #    flags, but specify no keychain; security will use the first it finds
+            prepare_darwin_volume_password "$volume" "$new_uuid" | sudo diskutil apfs encryptVolume "$volume" -user disk -stdinpassphrase
```
lilyball, line 223: Note that earlier when I recommended we just pass the password in the command-line arguments, that means we can get rid of this two-step keychain setup.

lilyball: Also if we end up keeping the two-step setup we should probably move the first step into `prepare_darwin_volume_password`, it's already responsible for updating the keychain item so it'll be cleaner if we can remove its assumption that the keychain item already exists.

scripts/create-darwin-volume.sh
```diff
@@ -127,42 +164,75 @@ main() {
...
-    if ! test_fstab; then
-        echo "Configuring /etc/fstab..." >&2
-        label=$(echo "$volume" | sed 's/ /\\040/g')
-        # shellcheck disable=SC2209
-        printf "\$a\nLABEL=%s /nix apfs rw,nobrowse\n.\nwq\n" "$label" | EDITOR=ed sudo vifs
+    if ! test_voldaemon; then
+        echo "Configuring LaunchDaemon to mount '$volume'..." >&2
+        generate_mount_daemon | sudo tee /Library/LaunchDaemons/org.nixos.darwin-store.plist >/dev/null
+
+        sudo launchctl load /Library/LaunchDaemons/org.nixos.darwin-store.plist
```
lilyball, line 233: `load` is a deprecated legacy command.

```suggestion
        sudo launchctl bootstrap system /Library/LaunchDaemons/org.nixos.darwin-store.plist
```

scripts/create-darwin-volume.sh
```diff
@@ -127,42 +164,75 @@ main() {
...
+
+        sudo launchctl start org.nixos.darwin-store
```
lilyball, line 235: I don't think this is necessary, bootstrapping the daemon should be enough as it's configured to run automatically. If we did need to run it manually then we should use `kickstart` as `start` is a legacy command, but we really shouldn't need this at all.

```suggestion
```

scripts/create-darwin-volume.sh
```diff
@@ -127,42 +164,75 @@ main() {
...
     fi
 
-    if ! test_fstab; then
-        echo "Configuring /etc/fstab..." >&2
-        label=$(echo "$volume" | sed 's/ /\\040/g')
-        # shellcheck disable=SC2209
-        printf "\$a\nLABEL=%s /nix apfs rw,nobrowse\n.\nwq\n" "$label" | EDITOR=ed sudo vifs
+    if ! test_voldaemon; then
+        echo "Configuring LaunchDaemon to mount '$volume'..." >&2
+        generate_mount_daemon | sudo tee /Library/LaunchDaemons/org.nixos.darwin-store.plist >/dev/null
```
lilyball, line 231: As previously mentioned, we need to pipe the volume name here so it can write the correct plist.

This does raise the worry of "what if we detect the plist already exists, but it has the wrong volume name" though.

scripts/create-darwin-volume.sh
lilyball: Heck, for that matter we should probably embed the service too, so that way if I delete the volume and clean up everything except for my keychain password, when it creates a brand new password this won't accidentally read the old one.

scripts/create-darwin-volume.sh
lilyball: I'm inclined to say if we detect that the plist already exists, we should unload it (`sudo launchctl bootout system/org.nixos.darwin-store`, note that this may return a non-zero exit code if the process is currently running but it will boot it out anyway according to my own experience), delete it, and then recreate it. This way we can be sure to embed the correct info.

Or at least, we should do this if `create_volume` is set. If it's not set, then we're re-using an existing volume and presumably the existing plist (if there is one) is therefore still valid.

scripts/install-nix-from-closure.sh
```diff
@@ -52,8 +43,8 @@ elif [ "$(uname -s)" = "Linux" ]; then
     echo "Note: a multi-user installation is possible. See https://nixos.org/nix/manual/#sect-multi-user-installation" >&2
 fi
 
-INSTALL_MODE=no-daemon
-CREATE_DARWIN_VOLUME=0
+export INSTALL_MODE=no-daemon
+CREATE_DARWIN_VOLUME=${CREATE_DARWIN_VOLUME:-1} # now default
```
lilyball, line 47: We can reduce name duplication using `:=` instead

```suggestion
: ${CREATE_DARWIN_VOLUME:=1} # now default
```

scripts/install-nix-from-closure.sh
```diff
@@ -69,13 +60,17 @@ while [ $# -gt 0 ]; do
         --no-modify-profile)
             NIX_INSTALLER_NO_MODIFY_PROFILE=1;;
         --darwin-use-unencrypted-nix-store-volume)
-            CREATE_DARWIN_VOLUME=1;;
+            (
+                echo "Warning: the flag --darwin-use-unencrypted-nix-store-volume"
+                echo "         is no longer needed and will be removed in the future."
+                echo ""
```
lilyball, line 66: It may not be necessary but I'm not sure if that means we should remove support for it.

scripts/create-darwin-volume.sh
lilyball, line 231: We need to check if the volume is encrypted first. If this is using a pre-existing unencrypted Nix volume, then the LaunchDaemon is superfluous (and the keychain entry won't exist).

For that matter, if it's using a pre-existing Nix volume at all, then there's no guarantee about the keychain entry's status. We should probably actually just skip this if we're not creating the volume. Anyone who creates their own volume is then on the hook for managing its mounting, and we can print a warning to this effect in the "Using existing volume" path.

scripts/install-nix-from-closure.sh
```diff
@@ -69,13 +60,17 @@ while [ $# -gt 0 ]; do
...
+            ) >&2;;
```
lilyball, line 67: The subshell can be replaced with a braced list instead.

```suggestion
            {
                echo "Warning: the flag --darwin-use-unencrypted-nix-store-volume"
                echo "         is no longer needed and will be removed in the future."
                echo ""
            } >&2;;
```

Though looking at the subsequent lines we're using subshells unnecessarily there too.

scripts/install-nix-from-closure.sh
abathur: clarity/consistency w/ other installer code on this one

scripts/create-darwin-volume.sh
abathur: I'll treat the "expect" comments here as contingent on the answer to this one, since they're moot if we throw it overboard.

1. I roughly agree with respect to my own system. 
2. My animating _concern_ is that passing the password will ward off prospective Nixers with policy/compliance/superstitious limitations.
3. I admit my own concern is basically superstitious, so I'm eager to defer to any consensus among senior Nix community members.

scripts/create-darwin-volume.sh
abathur: Removing the KeepAlive setting nullifies this though, yeah?

scripts/create-darwin-volume.sh
abathur: Good point. I was reacting to a comment about "Nix Store" not making sense because it holds more than just /nix/store, but didn't stop to consider the ATM machine I was setting loose.

scripts/create-darwin-volume.sh
abathur: > Heck, for that matter we should probably embed the service too

Can you clarify what you mean by "embed the service"? Passing the `-s` arg to find-generic-password?

scripts/create-darwin-volume.sh
abathur: I'll tentatively try this out and generate a new installer from it, but I'm leaving the comment open for now. 

The label was giving me a lot of grief at some point in development (I think it had something to do with what fields security was willing to recognize for the -U/update op?), but I would prefer a distinct label to fiddling with the kind, and I already have to hoop-jump to test other changes, so it's cheap to give it a roll.

scripts/create-darwin-volume.sh
abathur: tentatively added for testing

scripts/create-darwin-volume.sh
abathur: tentatively added for testing

scripts/create-darwin-volume.sh
abathur: This is something I left ~unhandled for discussion once I realized I'd need to thread it through everything (though I'm focused on the big-picture question about single-user mode before some of these narrower implementation questions).

My preference is to support a single fixed name (and maybe support pulling overriding it via the environment, if someone's really determined to do their own thing.) There are already a lot of moving parts here, and I actually think it's a kindness to tell the user "If you want to prepare your own volume, just name/label it 'Nix'"?

scripts/install-nix-from-closure.sh
abathur: Can you clarify?

scripts/install-nix-from-closure.sh
abathur: tentatively added both for testing

scripts/create-darwin-volume.sh
lilyball: It makes not doing this less harmful, but it's still a good idea to unload the job before deleting the plist. Especially if the user is going to reinstall before rebooting, I think if the job is still loaded `launchctl` will just error on the `bootstrap`.

scripts/create-darwin-volume.sh
lilyball: Yeah, the `-s` arg is what I meant.

scripts/create-darwin-volume.sh
lilyball: If we pass the `-s` flag to `find-generic-password` (and I think we should) then that's not a fixed string, that's the volume UUID. I think this is important because, as previously stated, if I uninstall Nix but forget to delete the keychain entry, reinstalling it will give me two keychain entries with the same account name but different services, and we need to pick the right one.

Since we have to embed that UUID in the launch daemon, that means we do need to delete and recreate the plist rather than saying "oh it already exists, leave it alone".

In the case where the user prepares their own volume, we should just not touch the launch daemon at all. Either the user is re-reusing a previous volume (at which point the previous launch daemon applies), or they made their own custom volume and the launch daemon shouldn't exist at all. Or to put it another way, we should only touch the launch daemon when we add an entry to the keychain.

scripts/install-nix-from-closure.sh
lilyball: I'm just thinking someone who doesn't care about the volume being encrypted might prefer not having a launch daemon to deal with. But it's probably better to have only a single installation configuration, which means removing support for this flag is correct.

scripts/create-darwin-volume.sh
abathur: Good point on reinstalling. I was just thinking it made sense to save them a step, but it'd definitely undercut people just trying to get the system clean to reinstall.

scripts/install-nix-from-closure.sh
abathur: Gotcha. 

By this point, it's fairly compatible with the existing install practice (in early drafts I had removed the fstab step to save us an install step and the user an uninstall step, but emily pointed out in #nix-darwin IRC that diskutil mount will pull mount options from it). Removing the scare-flag is an indication that I feel like solving the restore race-condition gets us to the point where a random Mac user can run the installer and it'll do something I could personally defend with a straight face if it ended up tending on HN.

I'm open to providing a no-daemon path here (though I feel like it's best to make an angry mob demand it, for the ~simplicity reason you mention). One option I've mentioned is just y/n prompting users to accept the daemon (but installing it by default if headless), which gives us a place (outside of the docs) to make it clear what problems it might cause, and that they'll have to manually fix their install if they regret it later. 

After subtracting the daemon, there's little technical difference between this PR and what we currently have, but I see having it as the default and having it as an opt-in fallback as a big qualitative difference! I think it's akin to Nix giving you *options* to do something radical like move "/nix", but putting it in a well-supported location by default.

abathur:
Want to address a few keychain comments from @lilyball that I see as leading to a question I'm on the fence about. I'm mixing related statements from a main-thread comment and another inline/review comment.

> 1. Last time I looked into this, it appeared that putting the volume passphrase into the system keychain and granting access to a few specific processes allowed the system to mount it automatically upon boot, just as it mounts the volume when unencrypted. 

When I wrote the linked summary, we had the impression that using the system keychain would be enough to skirt the race condition, and were lacking someone with the spare hardware/time/energy to try it and report/confirm one way or another. A few of us (myself included) have recently verified that it doesn't decrypt early enough to keep restored apps from failing if they depend on something on the Nix volume. 

> According to that script, the keychain entry needs to be accessible by `APFSUserAgent` and `CSUserAgent`

I suspect the Shopify script took a straightforward approach to the general question of how to set up a credential: let the OS prompt for credentials, tick the box to store the password in keychain, and then mimic the records it creates. 

I see how they got here, but I'm on the fence about how closely we should follow:
- If we add these two exemptions and the launchdaemon fails to mount the volume, the system should be able to seamlessly decrypt and mount the volume.
- If we don't add these exemptions and the launchdaemon fails to mount the volume, the system will prompt the user to allow APFSUserAgent to access the credential the next time it tries to auto-mount the volume (IIRC it's an allow/deny/always-allow prompt).
- If it fails over seamlessly, I worry that it'll mask problems with the launchdaemon and lead to people just living with race-condition restore problems for as long as it takes them to notice, realize Nix is to blame, and get annoyed enough to turn up.
- If it fails over squeakily (i.e., triggers the APFSUserAgent credential prompt) I have a *little* hope that they'll notice and look into the issue. (I suppose many people will still shrug, click through, and end up with race-condition problems?)

In earlier drafts, I intentionally made a nonstandard credential so that the system would prompt the user for a password (which they won't know, forcing them to look into the issue) if the daemon failed to mount it. This failure would also be near the problem, making it easier for them to find help (relative to trying to sort out quirky restore behavior which the user might blame on the app itself).

scripts/install-nix-from-closure.sh
lilyball: I think we should definitely investigate the reliability of simply granting access to `APFSUserAgent` (and `CSUserAgent`) for the keychain item and skipping the daemon. If that works, and if we believe that's reliable going forward, then that would greatly simplify things. I just don't know how to prove that this is future-compatible without getting feedback from an Apple engineer on this topic.

lilyball: Oops I just saw your top-level comment on this.

lilyball:
> When I wrote the linked summary, we had the impression that using the system keychain would be enough to skirt the race condition, and were lacking someone with the spare hardware/time/energy to try it and report/confirm one way or another. A few of us (myself included) have recently verified that it doesn't decrypt early enough to keep restored apps from failing if they depend on something on the Nix volume.

Oh that's a real bummer. Is it just a problem with auto-login not waiting for it to mount first?

> If it fails over squeakily (i.e., triggers the APFSUserAgent credential prompt) I have a *little* hope that they'll notice and look into the issue. (I suppose many people will still shrug, click through, and end up with race-condition problems?)

I'm not convinced that seeing the dialog will cause anyone to look into it, they'll just be confused, hit allow, and then have the race condition. In this scenario I'd rather the seamless behavior. My only real concern with that is just that we're relying on undocumented behavior in terms of which processes to grant access to, but I suppose as long as the processes exist (or as long as passing a missing binary to the `-T` flag isn't an error) then granting access is at worst harmless.

---

I do worry slightly about the scenario where the system decides to automount it before the daemon has had a chance to run. launchd does not let you specify dependencies (e.g. "gui depends on me") as it believes any such dependencies should be handled via IPC, and I don't know if the system makes any effort to allow all configured LaunchDaemons to run before going into the GUI session. So I wonder if the LaunchDaemon only works because it wins a race condition against login. Hopefully any such race has a very comfortable margin though.

I'm also thinking maybe we should specify `noauto` in the `fstab` entry. This way if something goes wrong with the daemon the volume will simply not mount, which will be a very clear signal to the user that something went wrong. That doesn't fix the theoretical potential race though as such a race would still exist against GUI app restoration.

abathur:
> Oh that's a real bummer. Is it just a problem with auto-login not waiting for it to mount first?

I'm not certain how you mean auto-login here (automatic login to the user account? I don't use this, but I think FV disables it?) or what exactly it comes down to. I suspect it's something like mount failing/timing out and retrying until the login keychain is available, and that dragging out past the point at which restoring apps have tried and failed. I found the basic behavior easier to reproduce than I expected: make an encrypted volume, create a file on it, open that file in textedit, and reboot.

> I'm not convinced that seeing the dialog will cause anyone to look into it, they'll just be confused, hit allow, and then have the race condition... I'm also thinking maybe we should specify `noauto` in the `fstab` entry. This way if something goes wrong with the daemon the volume will simply not mount, which will be a very clear signal to the user that something went wrong.

I agree, and noauto is probably a much better approach. Good point. I'll implement and test tonight.

> That doesn't fix the theoretical potential race though as such a race would still exist against GUI app restoration... I do worry slightly about the scenario where the system decides to automount it before the daemon has had a chance to run. launchd does not let you specify dependencies (e.g. "gui depends on me") as it believes any such dependencies should be handled via IPC, and I don't know if the system makes any effort to allow all configured LaunchDaemons to run before going into the GUI session. So I wonder if the LaunchDaemon only works because it wins a race condition against login. Hopefully any such race has a very comfortable margin though.

I am not sure how big the risk is. I haven't seen any problems yet. It wouldn't shock me if it's possible, but given what I've seen so far I'm more worried about another launchdaemon beating it than a GUI app restore. It will be nice to see how well it holds up under more robust use and/or devious test cases.

lilyball:
> I'm not certain how you mean auto-login here (automatic login to the user account? I don't use this, but I think FV disables it?) or what exactly it comes down to. I suspect it's something like mount failing/timing out and retrying until the login keychain is available

I meant automatic login of the user, yes, and FV is completely orthogonal.

That said, I thought we were talking about the system keychain, not the login keychain. The idea was the password would be in the system keychain and marked as accessible to `APFSUserAgent` and that this would apparently allow it to auto-mount the volume prior to GUI login (and the login keychain being unlocked). But you're talking about the login keychain here, which makes me wonder what you actually tested.

I'll have to see if I can find the time to test this myself.

> given what I've seen so far I'm more worried about another launchdaemon beating it than a GUI app restore

Yeah having another LaunchDaemon that tries to invoke something from `/nix` without using `wait4path` first would definitely be an issue.

scripts/create-darwin-volume.sh
abathur: Resolving since I've tentatively folded this into `prepare_darwin_volume_password`; two-step handled elsewhere.

abathur:
> I meant automatic login of the user, yes, and FV is completely orthogonal.

Not according to https://support.apple.com/en-us/HT201476. But I have no direct experience to comment from.

> That said, I thought we were talking about the system keychain, not the login keychain.

We are. I'm just describing part of the full timeline without re-reading carefully enough to realize the description sucked and I didn't connect the dots.
- System keychain becomes available at some point
- launchd begins running LaunchDaemons
- User initiates login process
- System begins trying to mount from fstab 
   - these succeed early if unencrypted 
   - but I think the system puts them in some sort of limbo if they are encrypted, perhaps because its authors know the system adds decryption credentials to the login keychain
- System begins restoring GUI apps
- Some time later encrypted mounts get mounted (regardless of whether their credential was stored in the system or login keychain). I've only been able to come up with 3 explanations:
    - they're waiting to check keychain at all until they know the login keychain is available because they expect encryption passwords to be there
    - they're explicitly checking the login keychain first, fall into a loop where they fail/timeout until it finally becomes available, and then finally check the system keychain
    - encrypted and unencrypted volumes all get handled at the same time, but the unlock process takes some significant amount of time, causing encrypted mounts to consistently lose the race with restoring apps

lilyball:
> Not according to https://support.apple.com/en-us/HT201476. But I have no direct experience to comment from.

Oh huh you're right, I thought it just required you to unlock the disk at boot but then would auto-login, but it does not actually allow for setting an auto-login user. I don't use auto-login so I didn't realize.

> but I think the system puts them in some sort of limbo if they are encrypted, perhaps because its authors know the system adds decryption credentials to the login keychain

So you're saying if the volume is encrypted with the password stored in the system keychain, and I sit at the login screen without logging in, it will never mount?

I definitely need to test this.

---

I'm also wondering if there's any way we can set up automounting for the volume. The info I'm seeing about `/etc/auto_master` is all about how to automount a remote volume, but if there's a way for us to mount a local volume, I'm wondering if that would then mean anything that tries to access `/nix` would block on the volume mounting (assuming of course that doing this would work with the system keychain approach pre-login).

scripts/create-darwin-volume.sh
lilyball: It occurs to me that since `/usr/bin/security` has access to the password, keeping it out of the command-line does literally nothing at all because any process that cared could just immediately invoke `/usr/bin/security` to read the password.

abathur:
> So you're saying if the volume is encrypted with the password stored in the system keychain, and I sit at the login screen without logging in, it will never mount?

Yes (with FileVault enabled, at least).

> I'm also wondering if there's any way we can set up automounting for the volume. The info I'm seeing about `/etc/auto_master` is all about how to automount a remote volume, but if there's a way for us to mount a local volume, I'm wondering if that would then mean anything that tries to access `/nix` would block on the volume mounting (assuming of course that doing this would work with the system keychain approach pre-login).

I tried automount a month or so back, but it was also too late to solve the problem.

mroi:
> I tried automount a month or so back, but it was also too late to solve the problem.

It may be late, but it should block any process trying to access a file there until the volume is mounted? And I am pretty sure you can automount local volumes as well.

abathur:
> I am pretty sure you can automount local volumes as well.

Yes.

> It may be late, but it should block any process trying to access a file there until the volume is mounted? 

Perhaps, but it didn't help apps restore and doesn't solve the problem.

abathur:
@mroi @lilyball I think I caught myself lying about `automount` working with local volumes. Are either of you familiar enough with it to verify?

I'm fairly sure it's a dead end by this point (after banging my head on it again today), but I revisited it to verify for sure whether or not `automount` was capable of unlocking a volume with keychain credentials (because that might mean we could force a call to automount in the daemon...) and ran into trouble. 

I thought I made this work a month ago with a line in /etc/auto_master like `/- auto_nix` and a direct map like `<mountpath> -fstype=apfs :/dev/disk1s7`, but I didn't make a note of the mountpath I used, and I haven't been able to it to work again today. 

I hadn't used `automount` previously, so I suspect that I misinterpreted the output from `automount -vc` indicating it successfully mapped paths (and those showing up in `mount`), and didn't look closely enough at `diskutil apfs list` to notice that the volume wasn't actually mounting.

If you have any success--I also recall having trouble getting it to mount at any root path.

scripts/create-darwin-volume.sh
abathur: > I assume we're using `expect` here purely to avoid having the password show up in the command-line arguments in case anyone is sniffing it?
> 
> I don't think we need to worry about that. The volume is going to be mounted the entire time the computer is running, so any process that can sniff the command-line arguments can also just look at the volume while it's mounted anyway. We should ditch the complexity of `expect` (and the risk that it might be deprecated too since `tclsh` is) and just provide the passphrase on the command-line.

I finally found a way to have and eat our cake, here (dump the expect dependency, without the password ending up in the process table). Apparently `security -i` will happily run commands from a heredoc. See b38e85f 

> It occurs to me that since `/usr/bin/security` has access to the password, keeping it out of the command-line does literally nothing at all because any process that cared could just immediately invoke `/usr/bin/security` to read the password.

The process still needs root to read the system keychain, but this is generally still a point of concern. The daemon component probably needs to get replaced with a binary at some point.

I've tried a few things here:
1. Find some imperative command that uses whatever auth mechanism is used when entries from fstab are mounted. 
    - ~`mount`~
    - ~`diskutil mount`~ 
    - ~`diskutil apfs unlockVolume` with no passphrase~
    - `automount`? Jury is still out, but I'm not optimistic.
2. Scripts can be added to the exemption dialog, but aren't actually allowed access.

lilyball:
I have not ever actually configured automount myself. I did look through the manpages the other day and was disappointed that it didn't have any direct examples of automounting a local volume.

abathur:
@mroi @lilyball I've been searching through some of the underlying components here to see if I can find any way we could trigger, on-demand, the system mechanism that unlocks from keychain. I stumbled on an interesting issue (https://github.com/openwall-com-au/BootUnlock/issues/12) on a project (https://github.com/openwall-com-au/BootUnlock) that is picking at something similar. 

I haven't had time to read much of it yet, but a quick skim makes me think it may help answer some timeline questions.

The repo also includes an approach to any process that can use security having access. I've seen this elsewhere but forgot about it. They're making/signing a copy of the security program, so that they can just permit that one.

> I have not ever actually configured automount myself. I did look through the manpages the other day and was disappointed that it didn't have any direct examples of automounting a local volume.

Yes. Tricky to search, too. I assume I found an example somewhere, but I haven't been able to re-find it.

lilyball:
> The repo also includes an approach to any process that can use security having access. I've seen this elsewhere but forgot about it. They're making/signing a copy of the security program, so that they can just permit that one.

We really should not go down that rabbit hole. We should refrain from putting anything else on the system outside of the normal Nix locations that's not strictly necessary (which is to say, removing the launch daemon plist should be sufficient to clean up the daemon behavior). And since the volume is expected to be mounted always while the system is running I really would not worry about trying to keep the password secure. Any attacker who has access to the running system can already inspect the contents of the volume, we're encrypting it to protect against offline access, and I don't think "attacker has access to running system but can only exfiltrate a few dozen bytes of data, wants to record the password so they can swipe the physical machine later and decrypt the Nix volume" is a practical threat model to care about.

abathur:
> We really should not go down that rabbit hole.

I agree. Mostly just find it interesting/telling that they haven't found anything better. 

Manually invoking diskarbitrationd, or killing it and letting launchd restart it, will both trigger a failed attempt to mount the volume (visible in the logs), so I'm not sure it's the originator of the requests that actually unlock it. If it is, there's just something different about the two contexts...

I spent a while stepping through verbose logs, and tried restarting most of the other components I could spot as involved, but nothing triggered it to automatically mount.

The log entries closest to the actual auth process appear to be `APFSUserAgent: [com.apple.diskunlock:Unlock]` and  `APFSUserAgent: [com.apple.diskunlock:Keychain]` if anyone wants a starting point.

galaxy4public:
> I agree. Mostly just find it interesting/telling that they haven't found anything better.

We did :), but it requires a creation of a tool that is quite complex.  The reverse engineering of Apple's SecurityAgentPlugins shows that the `HomeDirMechanism`, a plugin that prepares user's home directory right after the system confirmed that the supplied credentials are valid, can mount different filesystems including network (AFS, Samba, etc) and *encrypted APFS* (if the latter has attached "Cryptousers" in its container keybag.

Now, the challenge is that all Apple tools (`fdesetup`, `diskutil`, etc.) are explicitly configured to allow adding this information to the APFS container where the System volume resides.  We are working on a tool that would allow an administrator to add the keybag with "Cryptousers" to any APFS container, but the work progresses quite slowly.

scripts/create-darwin-volume.sh
```diff
@@ -83,6 +61,58 @@ suggest_report_error(){
     echo "       please report this @ https://github.com/nixos/nix/issues" >&2
 }
 
+generate_mount_command(){
+    if test_filevault_in_use; then
+        printf "    <string>%s</string>\n" /bin/sh -c '/usr/bin/security find-generic-password -a "Nix" -w | /usr/sbin/diskutil apfs unlockVolume "Nix" -mountpoint /nix -stdinpassphrase'
```
lilyball, line 66: Since we're passing the `-s` flag with the UUID when creating the password, we need to be passing it here too, as mentioned [in this comment](https://github.com/NixOS/nix/pull/4181/files#r512356331).

abathur: This is still on my radar. I set out trying to iron out some of these issues, and found myself feeling like it would be a big leg up to have access to the UI/X affordances of the multi-user installer before tackling them. 

The draft refactor to get that functionality also put me really close (I feel like) to laying some groundwork for directly curing more issues instead of just throwing instructions at the user. In turn, that put me really close to laying some groundwork for uninstalling bits of this.

It isn't a full solution yet (i.e., not covering components I didn't add yet; not covering linux), nor am I sure if it'll end up in scope here or if parts will need to get broken off...

abathur:
Apologies for the radio silence. I've pushed a large ~WIP refactor (fair warning: I hope to obsolete & remove chunks of this, pending discussion) of this PR (and the existing installer).

I think the most-important high-level changes are:
1. closes the single-user macOS install path
2. sources create-darwin-volume from the darwin multi-user installer
3. supports "curing" darwin-volume cruft if any is left from a previous install (removing it if Nix is the only thing in it, editing Nix out otherwise)
4. extends the "curing" concept one step further into an "uninstaller" proof-of-concept
    - *not* a full uninstaller--that's out-of-scope here
    - but I want to demonstrate the case because there's naturally a lot of code and UI/X overlap between the tasks, and I suspect that discussion will clarify whether the functionality belongs here or elsewhere

I have updated the installer in the first post if you'd like to try it. Otherwise, you can see it in action:
- with user interaction: https://asciinema.org/a/zzFgKW03XFA89aBLFa29ZGAYa
- headless: https://github.com/abathur/syspolicyd_assessments/runs/1376970745?check_suite_focus=true

I'll also briefly outline ~motives in case it helps connect any dots. If the above makes (narrative?) sense to you, you can probably ignore the list below.
- I hoped to skirt some state/ergonomic problems people might run into around uninstall/reinstall (and be a little lazy :) by keeping the volume-mounter generic, but @lilyball poked a few holes in this approach, making it clear that we need to embed the volume UUID. But, this makes detecting existing state more critical.
- Working through detecting/objecting to these conditions and walking users through fixing them made me suspect it wouldn't be much harder (and better from a dev/community-relations perspective) to just "cure" cruft (with user input).
- I took the tentative agreement on removing single-user mode for macOS as a green light to partially integrate the volume script and multi-user installer to take advantage of the latter's better UI/X affordances. I'm a little leery of blowback, here, so I've taken some pains to ensure it's still possible to run this script standalone for now.
- It seemed like the logic, code, and UI idioms I needed to prompt users about curing each component would readily compose into an uninstaller, so I wanted to stretch a little to test out the concepts and see if anything interesting shakes out of it.

scripts/create-darwin-volume.sh
abathur: I've wrestled with this a bit. The latest update may resolve it, though that's contingent on what people think of the "curing" approach (and it likely needs logical refinement).

Basically, I've cut attempts to infer user intent or do anything magic based on what exists. If the root is read-only and `NIX_VOLUME_CREATE=1` (the default), the installer will want to create a volume (encrypted if filevault is on), and the "curing" version currently enabled asks the user for permission to remove anything that conflicts (volume, keychain credentials, fstab, synthetic.conf, daemon) before it'll install.

I've gone to a little pain to set up create-darwin-volume.sh so that it'll still behave roughly as it currently does if invoked with no arguments, but to also support passing a function to call when you source or invoke it. I won't say there's a clear path to it yet since I haven't thought through the logistics of scripting this exact case, but I'm inclined to close off some options for now and focus on getting the golden path right--but make the installer scripts flexible enough that people can leverage them rather than just copy and customize them?

abathur:
I've done some additional refactoring of the work here. I've changed enough that I think it merits a fresh look, so I've opened a new PR #4289 to take over the work here and am closing this one. If you're subscribed here, you'll probably want to subscribe to that one.

PR #4182, opened 2020-10-23
Print built derivations as json for build 

### LINKED ISSUE #1930: Nix 2.0: Print out path to stdout with nix build

mkenigs:
Add --json option to nix build to allow machine readable output on
stdout with all built derivations

Fixes #1930

I think closes #2423 and #1647

Wasn't quite sure is there a better place to put a test?

### DIFF
--- a/src/nix/build.cc
+++ b/src/nix/build.cc
@@ -5,9 +5,11 @@
 #include "store-api.hh"
 #include "local-fs-store.hh"
 
+#include <nlohmann/json.hpp>
+
 using namespace nix;
 
-struct CmdBuild : InstallablesCommand, MixDryRun, MixProfile
+struct CmdBuild : InstallablesCommand, MixDryRun, MixJSON, MixProfile
 {
     Path outLink = "result";
     BuildMode buildMode = bmNormal;
@@ -86,6 +88,8 @@ struct CmdBuild : InstallablesCommand, MixDryRun, MixProfile
                     }, buildables[i]);
 
         updateProfile(buildables);
+
+        if (json) logger->cout("%s", buildablesToJSON(buildables, store).dump());
     }
 };
 
--- a/src/nix/installables.cc
+++ b/src/nix/installables.cc
@@ -16,7 +16,34 @@
 #include <regex>
 #include <queue>
 
+#include <nlohmann/json.hpp>
+
 namespace nix {
+
+nlohmann::json BuildableOpaque::toJSON(ref<Store> store) const {
+    nlohmann::json res;
+    res["path"] = store->printStorePath(path);
+    return res;
+}
+
+nlohmann::json BuildableFromDrv::toJSON(ref<Store> store) const {
+    nlohmann::json res;
+    res["drvPath"] = store->printStorePath(drvPath);
+    for (const auto& [output, path] : outputs) {
+        res["outputs"][output] = path ? store->printStorePath(*path) : "";
+    }
+    return res;
+}
+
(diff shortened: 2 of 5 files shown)


regnat (changes requested):
That looks good overall, except that it brings back the intentionally removed behavior of printing the built store paths 🤔

(Ftr, and unless I misunderstood the matter, the issue with printing these paths is that they don't come with any guaranty of not being gc-ed if `--no-link` is passed or the symlinks are removed, bringing back some scary memories of use-after-free)

dasJ:
> That looks good overall, except that it brings back the intentionally removed behavior of printing the built store paths thinking

@regnat There was still nobody whould could explain to me why `$(nix build)/bin/whatever` is an antipattern. The only "explanation" is the gc which is completely irrelevant in cases where I just want to execute a tool and I can just retry when it fails.
There is no alternative to this workflow (except this PR which requires me to ADDITIONALLY use jq (with its annoying syntax) instead of just `$()`

regnat:
> @regnat There was still nobody whould could explain to me why `$(nix build)/bin/whatever` is an antipattern. The only "explanation" is the gc which is completely irrelevant in cases where I just want to execute a tool and I can just retry when it fails.

I don't want to enter this debate. I didn't state any position, just noted that it might be controversial. But fwiw:

> There is no alternative to this workflow (except this PR which requires me to ADDITIONALLY use jq (with its annoying syntax) instead of just `$()`

There are cases where it wouldn't work, but `nix shell --run whatever` actually does what you want (and is arguably nicer).

mkenigs:
Thanks @regnat , applied those changes

mkenigs:
@regnat I think including `json_fwd` requires passing PRECOMPILE_HEADERS=1 to make. Is that okay? And would that need to be added to https://nixos.org/manual/nix/unstable/hacking.html?

regnat:
> @regnat I think including `json_fwd` requires passing PRECOMPILE_HEADERS=1 to make. Is that okay? And would that need to be added to the nix-shell instructions for `make` at https://nixos.org/manual/nix/unstable/hacking.html?

No that shouldn't. I guess what's happening is that `installables.cc` was transitively including `json.hpp` via `installables.hh`.
So you need to explicitely include `json.hpp` in `installables.cc`.

The reason why it works with `PRECOMPILE_HEADERS=1` is that this flag cause the compiler to first bundle all the external headers (including `json` and `json_fwd`) into one big precompiled thing `precompiled-headers.h.gch` which is systematically included. This is faster than all the individual includes because of the precompilation step, but the consequence is that when that flag is set all the external includes are redundant.

mkenigs:
> There are cases where it wouldn't work, but `nix shell --run whatever` actually does what you want (and is arguably nicer).

What are cases where you really do need `nix build`? To me it seems like even out links are pretty inconsistent with nix philosophy, since they can be left anywhere, rather than managed declaratively. `nix shell --run whatever` seems a lot nicer to me.

dasJ:
@mkenix Mainly for projects which are independent from nixpkgs. Not everything is fitted for nixpkgs, for example [NixDroid](https://github.com/ajs124/nixdroid).

edolstra:
Thanks!

PR #4189, opened 2020-10-26
Allow nix.conf options to be set in flake.nix

edolstra:
This makes it possible to have per-project configuration in `flake.nix`, e.g. binary caches and other stuff:

```
nixConfig.bash-prompt-suffix = "[1;35mngi# [0m";
nixConfig.substituters = [ "https://cache.ngi0.nixos.org/" ];
```

@garbas

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -71,11 +71,17 @@ static std::tuple<fetchers::Tree, FlakeRef, FlakeRef> fetchOrSubstituteTree(
     return {std::move(tree), resolvedRef, lockedRef};
 }
 
-static void expectType(EvalState & state, ValueType type,
-    Value & value, const Pos & pos)
+static void forceTrivialValue(EvalState & state, Value & value, const Pos & pos)
 {
     if (value.type == tThunk && value.isTrivial())
         state.forceValue(value, pos);
+}
+
+
+static void expectType(EvalState & state, ValueType type,
+    Value & value, const Pos & pos)
+{
+    forceTrivialValue(state, value, pos);
     if (value.type != type)
         throw Error("expected %s but got %s at %s",
             showType(type), showType(value.type), pos);
@@ -114,7 +120,6 @@ static FlakeInput parseFlakeInput(EvalState & state,
                 expectType(state, tString, *attr.value, *attr.pos);
                 input.follows = parseInputPath(attr.value->string.s);
             } else {
-                state.forceValue(*attr.value);
                 if (attr.value->type == tString)
                     attrs.emplace(attr.name, attr.value->string.s);
                 else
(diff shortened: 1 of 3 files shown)


regnat:
How would that work with dependencies? Like if I depend on a flake `foo` that sets some Nix options. Would these be ignored or inherited in the top-level flake?

edolstra:
Currently it only uses options from the top-level flake. Since dependencies are fetched lazily that's really the only option, unless we propagate options into the lock file.

matthewbauer:
How useful is this without the "trusted-public-keys" setting? Without setting that, you just get an error when some project you are working on introduces a new substituter (unless you happen to have it yourself). Of course allowing trusted-public-keys is probably a security concern.

It would be nice UX to prompt users if no trusted key matches, asking if they trust the cache URL. If not, it would get on a list of "untrusted-public-keys" so we would avoid trying it in the future.

src/libexpr/flake/flake.cc
```diff
@@ -223,10 +228,41 @@ static Flake getFlake(
     } else
         throw Error("flake '%s' lacks attribute 'outputs'", lockedRef);
 
+    auto sNixConfig = state.symbols.create("nixConfig");
```
garbas, line 231: I think we can safely rename this to only `config`.

edolstra:
`binary-caches` exists as a deprecated alias of `substituters`. It's not a good name because there are substituters that are not binary caches and that you might want to configure in flake.nix, e.g. `ssh://some-server`.

andir:
I like the idea of having per-project configuration. In the past I had to use some weird hacks with direnv to get my nix CLI use a slightly different configuration for some customer project. Not being able to set a public key is reasonable but I wonder what the value in this is if we can't use our own caches? We would still need system-wide (or if trusted, user local) configuration of trusted signing keys. How many binary caches are there that someone would be using in a project/company to warrant that not being configured the same time you configure the public keys?

The given example of an SSH substituters might be the only one I really see as "complete" here. At least with SSH it is common to configure authenticaiton outside of what Nix can do (right now).


As long as we keep the number of supported settings as restrictive as proposed in this PR this is probably fine. If we ever expand it further (e.g. allowing users to set `sandbox = false`) we should think about requiring an explicit "ack" from the user (the first time the flake, in that version, is being built). I really like the direnv model for confirmation from users.

src/libexpr/flake/flake.cc
```diff
@@ -599,4 +635,30 @@ Fingerprint LockedFlake::getFingerprint() const
...
 Flake::~Flake() { }
 
+void ConfigFile::apply()
+{
+    for (auto & [name, value] : options) {
+        // FIXME: support 'trusted-public-keys' (and other options), but make it TOFU.
+        if (name != "bash-prompt-suffix" &&
+            name != "bash-prompt" &&
+            name != "substituters" &&
+            name != "extra-substituters")
```
domenkozar, line 645: netrc-file?

edolstra: Maybe if it's relative to the flake directory, otherwise it's probably not very useful.

adrian-gierakowski: I guess to make netrc-file useful we’d also need extra-sandbox-paths?

Btw. I wonder why the path of netrc-file is not automatically added to sandbox? Is there any use of setting this option without adding the path to sandbox?

matthewbauer: netrc-file should only be read by the Nix builtin fetchers which aren't subject to sandboxing rules.

src/libexpr/flake/flake.cc
edolstra: No, that would be ambiguous with `config` used in the NixOS module system.

domenkozar:
@edolstra no netrc-file or public keys support? That really limits the use of binary caches for flakes.

blaggacao:
I want to link to a nascent community project with certain scope overlap. I furthermore suggest and encourage, as far as I'm conceded to do so, to explore any so opportunity for collusion and thought entanglement https://github.com/numtide/devshell

jmatsushita:
For those who like me, wonder where to add the `nixConfig` attribute, I found [here](https://zimbatm.com/NixFlakes/#flake-schema) that it is a top level attribute in a `flake.nix`. So it's used like so:
```nix
{
  description = "...";
  nixConfig.bash-prompt = "\[nix-develop\]$ ";
  ...
}
```

Issue #4200, opened 2020-10-29
Memory corruption


kirelagin:
**Describe the bug**

When trying to build a certain expression, I am getting:

```
malloc(): invalid size (unsorted)
```

or, sometimes, much less often either:

```
malloc(): invalid next size (unsorted)
```

or

```
malloc(): invalid next->prev_inuse (unsorted)
```

or some other `malloc` error, which I failed to write down.

There seems to be something interesting going on with string allocation.

**Steps To Reproduce**

I managed to minimise our case to a simple build with `haskell.nix` of an empty pseudo-Haskell project with a long enough `stack.yaml` file. By varying the size of `stack.yaml` you can get different errors and have fun (🥳).

Unfortunately, since it uses `haskell.nix`, you might need to setup the [IOHK binary cache](https://github.com/input-output-hk/plutus#iohk-binary-cache). Or maybe not, I’m not sure... maybe we can share our cache...

* `git clone https://github.com/serokell/nix-malloc-repro.git`
* `cd nix-malloc-repro`
* `nix build -f .`

Note that _both_ `nix-build` and `nix-daemon` dump core.

**Expected behavior**

Some error related to the Haskell project being completely empty.

**Additional context**

The repository has its own pinned dependencies and I am using `nixUnstable` from the `nixos-20.09` branch, although I don’t think any of this matters.

* It reproduces with current `master` (e0ca98c2071b815578470e280df8fdb750c7e23b).
* @notgne2, who has older nixpkgs with Nix f15651303f8596bf34c67fc8d536b1e9e7843a87 cannot reproduce it.

kirelagin:
I guess I’ll just leave it bisecting overnight. So far, if I did not mess up, I know that:

```
git bisect start
# bad: [ed02d20e1d437ea2c3f99b13f1dd2ff31c48e0c8] Merge pull request #4085 from 0mp/patch-2
git bisect bad ed02d20e1d437ea2c3f99b13f1dd2ff31c48e0c8
# good: [6387550d58884ac09204c339ed3a3cd700780a75] Get rid of confusing `std::optional<bool>` for validity
git bisect good 6387550d58884ac09204c339ed3a3cd700780a75
```

kirelagin:
Current status:

```
# bad: [dac8a1a5e178f3a000bb6995c41b6c36d3a0de6c] Merge pull request #4052 from ujjwaljainnn/fix-typo
[...]

PR #4030, opened 2020-09-17
Overhaul wopAddToStore

roberth:
- Implement `RemoteStore::addToStoreFromDump`. Closes #4009 
 - Constant memory `addToStore` in daemon. Closes #3802
 - Fixes #358 (confirmed now), possibly #1117
 - Introduce `ContentAddressMethod` which represents the labels that precede CA hashes but not the hash
 - Fold `wopAddTextToStore` into `wopAddToStore`. Not super necessary but was easy with the above in place and fits well with the new `ContentAddress`* stuff
 - `FramedSource` and `FramedSink` are now reusable
 - Affects #1184

PR #4206, opened 2020-10-30
Fix memory corruption caused by GC-invisible coroutine stacks

roberth:
Crucially this introduces BoehmGCStackAllocator, but it also
adds a bunch of wiring to avoid making libutil depend on bdw-gc.
The extra wiring, more than half of the PR, can be removed if it's ok to depend on bdw-gc.

Part of the solutions for #4178, #4200

Ping @edolstra

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -27,6 +27,10 @@
 #include <gc/gc.h>
 #include <gc/gc_cpp.h>
 
+#include <boost/coroutine2/coroutine.hpp>
+#include <boost/coroutine2/protected_fixedsize_stack.hpp>
+#include <boost/context/stack_context.hpp>
+
 #endif
 
 namespace nix {
@@ -220,6 +224,26 @@ static void * oomHandler(size_t requested)
     /* Convert this to a proper C++ exception. */
     throw std::bad_alloc();
 }
+
+class BoehmGCStackAllocator : public StackAllocator {
+  boost::coroutines2::protected_fixedsize_stack stack;
+
+  public:
+    boost::context::stack_context allocate() override {
+        auto sctx = stack.allocate();
+        GC_add_roots(static_cast<char *>(sctx.sp) - sctx.size, sctx.sp);
+        return sctx;
+    }
+
+    void deallocate(boost::context::stack_context sctx) override {
+        GC_remove_roots(static_cast<char *>(sctx.sp) - sctx.size, sctx.sp);
+        stack.deallocate(sctx);
+    }
+
+};
+
+static BoehmGCStackAllocator boehmGCStackAllocator;
+
 #endif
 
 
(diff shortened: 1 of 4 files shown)


edolstra:
Yes, it's okay to make libutil depend on bdw-gc. The only downside is that it makes the footprint of nix-daemon a bit bigger, if we ever want to have a separate daemon executable (https://github.com/NixOS/rfcs/pull/68).

Another solution would be to not have GC roots on coroutine stacks. Where are we doing that?

8 MiB sounds like quite a lot for coroutines. So long as we're not evaluating in coroutines (which we shouldn't, since the stack overflow detection probably doesn't work there), then a much smaller stack should be sufficient. How much does Boost allocate by default?

roberth:
> have GC roots on coroutine stacks. Where are we doing that?

We've started evaluating in coroutines with #4030. It's not so obvious because of the indirections, but the source filter in eg `filterSource` needed to run while creating the dump, which is now interleaved with sending the dump.

> 8 MiB sounds like quite a lot for coroutines

I think I agree, but for evaluation it seems like a safe number. It's only virtual memory, so a bigger value shouldn't hurt.

> which we shouldn't, since the stack overflow detection probably doesn't work there

This works. Boost provides `protected_fixedsize_stack` for this purpose.
The default stack size, 128K iirc, is indeed too small for evaluation. That had the benefit of immediately confirming that overflow detection worked :) 

> Yes, it's okay to make libutil depend on bdw-gc.

Does that mean you want to remove the `DefaultStackAllocator` etc boilerplate? Keeping it helps with the NixOS/rfcs#68 goal.

mkaito:
Figured I'd drop by to report that this fixed the `malloc` issues that had been plaguing me for months on nix unstable.

roberth:
@edolstra Not sure if you saw my response. I think I've addressed your concerns.

edolstra:
@roberth Thanks!

wizeman:
I suspect that this PR has resulted in `hydra-queue-runner` locking up and consuming 100% CPU usage for me (possibly due to an unrelated bug, or maybe because this PR is incomplete?).

Could you take a look at the stack trace in https://github.com/NixOS/hydra/issues/816#issuecomment-724140905 and see whether it's related?

wizeman:
Do you think the following patch would fix the 100% CPU usage / infinite stack overflow? I don't understand the code (way too many C++isms for me) so I'm not sure if it's correct...

```diff
diff --git a/src/libutil/serialise.cc b/src/libutil/serialise.cc
index 28f6968d0..038ede049 100644
--- a/src/libutil/serialise.cc
+++ b/src/libutil/serialise.cc
@@ -195,7 +195,7 @@ class DefaultStackAllocator : public StackAllocator {
     }
 
     void deallocate(boost::context::stack_context sctx) {
-        deallocate(sctx);
+        stack.deallocate(sctx);
     }
 };
```

wizeman:
I've submitted the above patch as PR #4242.

PR #4220, opened 2020-11-04
Copy drv outputs between stores

regnat:
Partial (cleaner) rewrite of #4174. This PR doesn't try to get substitution to work, just to be able to manually copy drv outputs closures between stores (which is already a big deal).

The exact schema of the local db and the remote caches isn't really set yet, but we're hopefully approaching it. The rational behind most of the changes is (lengthily) explained in https://hackmd.io/@Xs46ZZwNQtOhEkXnSehjeA/rkDU1BTOv


regnat:
Closing in favor of #4225

PR #4225, opened 2020-11-06
Minimal implementation of a drv outputs copy

regnat:
Depends on #4227 #4282 #4284 

This PR tries to provide the simplest possible way of copying drv outputs.
In particular it doesn't try to tackle the issue of copying the drv output closure at all, all it does is allow to do `nix copy /nix/store/...-foo.drv!out` which will copy the corresponding outpath and register the out path mapping for `...-foo.drv!out`.

For coherence (and also because it's at least as simple to do things that way), `nix copy nixpgs#foo` has the same semantics, meaning that it will both copy the output path(s) and register the derivation output(s).

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -444,6 +444,33 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
     })->path;
 }
 
+std::optional<const DrvOutputInfo> BinaryCacheStore::queryDrvOutputInfo(const DrvOutputId & id)
+{
+    auto outputInfoFilePath =
+        "/drvOutputs/" + std::string(id.drvPath.hashPart()) + "!" + id.outputName + ".doi";
+    auto rawOutputInfo = getFile(outputInfoFilePath);
+
+    if (rawOutputInfo) {
+        return { DrvOutputInfo::parse(*rawOutputInfo, outputInfoFilePath) };
+    } else {
+        return std::nullopt;
+    }
+}
+
+std::optional<StorePath> BinaryCacheStore::queryOutputPathOf(
+    const StorePath& drvPath,
+    const std::string& outputName) {
+    if(auto outputInfo = queryDrvOutputInfo(DrvOutputId{drvPath, outputName}))
+        return {outputInfo->outPath};
+    return std::nullopt;
+}
+
+void BinaryCacheStore::registerDrvOutput(const DrvOutputId & id, const DrvOutputInfo & info)
+{
+    auto filePath = "/drvOutputs/" + std::string(id.drvPath.hashPart()) + "!" + id.outputName + ".doi";
+    upsertFile(filePath, info.to_string(), "text/x-nix-derivertopath");
+}
+
 ref<FSAccessor> BinaryCacheStore::getFSAccessor()
 {
     return make_ref<RemoteFSAccessor>(ref<Store>(shared_from_this()), localNarCache);
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -434,7 +434,9 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
     if (!repair && isValidPath(path))
         return path;
 
-    auto source = StringSource { s };
+    StringSink sink;
+    dumpString(s, sink);
+    auto source = StringSource { *sink.s };
     return addToStoreCommon(source, repair, CheckSigs, [&](HashResult nar) {
         ValidPathInfo info { path, nar.first };
         info.narSize = nar.second;
--- a/src/libstore/drv-output-info.hh
+++ b/src/libstore/drv-output-info.hh
@@ -28,6 +28,9 @@ struct DrvInput : RawDrvInput {
 
     std::string to_string() const;
     static DrvInput parse(const std::string & strRep);
+
+    const RawDrvInput variant() const
+    { return static_cast<RawDrvInput>(*this); }
 };
 
 struct DrvOutputInfo {
--- a/src/libstore/drv-output-info.hh
+++ b/src/libstore/drv-output-info.hh
@@ -1,0 +1,42 @@
+#pragma once
+
+#include "path.hh"
+
+namespace nix {
+
+struct DrvOutputId {
+    StorePath drvPath;
+    std::string outputName;
+
+    std::string to_string() const;
+
+    static DrvOutputId parse(const std::string &);
+
+    bool operator<(const DrvOutputId& other) const { return to_pair() < other.to_pair(); }
+    bool operator==(const DrvOutputId& other) const { return to_pair() == other.to_pair(); }
+
+private:
+    // Just to make comparison operators easier to write
+    std::pair<StorePath, std::string> to_pair() const
+    { return std::make_pair(drvPath, outputName); }
+};
+
+typedef std::variant<StorePath, DrvOutputId> RawDrvInput;
+
+struct DrvInput : RawDrvInput {
+    using RawDrvInput::RawDrvInput;
+
+    std::string to_string() const;
+    static DrvInput parse(const std::string & strRep);
+};
+
+struct DrvOutputInfo {
+    StorePath outPath;
+
+    std::string to_string() const;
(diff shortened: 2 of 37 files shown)


regnat:
I've updated the PR to make the `outputPath` of derivation outputs a foreign key pointing to a `ValidPaths` entry.
This should make things cleaner as it removes the possibility of having a dangling reference (which would cause issues if this drv output gets rebuild and yields a different output path − bc of non-determinism).

@edolstra the `DerivationOutputs` table currently doesn't have a `user` field, but we'll probably want it at some point in the future. Should we add it right now (at the cost of keeping some possibly never used cruft) or defer it (at the cost of another db schema chang)

Ericson2314:
> @edolstra the DerivationOutputs table currently doesn't have a user field, but we'll probably want it at some point in the future. Should we add it right now (at the cost of keeping some possibly never used cruft) or defer it (at the cost of another db schema chang)

I would defer it, as the migration is not so complex that I'm worried about doing it later, but the reverse migration is lossy.

src/libstore/binary-cache-store.cc
```diff
@@ -434,7 +434,9 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
     if (!repair && isValidPath(path))
         return path;
 
-    auto source = StringSource { s };
+    StringSink sink;
+    dumpString(s, sink);
+    auto source = StringSource { *sink.s };
```
Ericson2314, line 439: This looks like a bugfix we should PR sooner?

src/libstore/drv-output-info.hh
```diff
@@ -0,0 +1,47 @@
...
+    using RawDrvInput::RawDrvInput;
+
+    std::string to_string() const;
+    static DrvInput parse(const std::string & strRep);
+
+    const RawDrvInput variant() const
+    { return static_cast<RawDrvInput>(*this); }
+};
+
+std::list<std::string> stringify_refs(const std::set<DrvInput> & refs);
```
Ericson2314, line 36: unused

src/libstore/local-store.cc
```diff
@@ -581,17 +615,20 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
...
 
 
-void LocalStore::linkDeriverToPath(const StorePath & deriver, const string & outputName, const StorePath & output)
+void LocalStore::registerDrvOutput(const DrvOutputId& id, const DrvOutputInfo & info)
 {
     auto state(_state.lock());
-    return linkDeriverToPath(*state, queryValidPathId(*state, deriver), outputName, output);
+    // XXX: This ignores the references of the output because we can
+    // recompute them later from the drv and the references of the associated
+    // store path, but doing so is both inefficient and fragile.
```
Ericson2314, line 623: This doesn't apply to this version, right?

Ericson2314:
I think I would feel most comfortable if the first minimal version didn't store mappings for unresolved derivations to paths, because doing that while making sure we don't have accidental mapping collisions is very hard.

src/libstore/binary-cache-store.cc
regnat: We could yes. Though I don't think this method was used anywhere before this PR so probably not an emergency

src/libstore/local-store.cc
regnat: Indeed it doesn't. Removed in 58ea93c3ba8ff647105500bd2219c66ae1791308

src/libstore/drv-output-info.hh
regnat: Thanks, removed in a2803d672817d2dc1cb3a9992f88232456feb1b8

regnat:
> I think I would feel most comfortable if the first minimal version didn't store mappings for unresolved derivations to paths, because doing that while making sure we don't have accidental mapping collisions is very hard.

I fail to see which collision you think of here. It's possible that the drv would resolve to something else (and possibly transitively produce a different output path) if you were to rebuild it, but it's not fundamentally different from having the resolved derivation producing a different output path.

Ericson2314:
> I fail to see which collision you think of here.

The only where the collected build-time-only deps could be built again different.

> but it's not fundamentally different from having the resolved derivation producing a different output path.

Sure, but if we just start with resolved derivation mappings, we are sure to be correct: rather than stuff being dubiously cached, it will just fail to be cached if some resolved derivation along the way is different. 

Because that is is a "definitely correct" starting point, I think it's where I'd like to begin.

------

Another idea i was thinking was leaving the current DerivationOutputs as-is, and having a new table for CA derv mappings. Queries can union the two tables together, and this means we can continue having the foreign key and simple cleanup for input-addressed derivations. Thoughts?

Ericson2314:
So picking up where I left off with https://github.com/NixOS/nix/pull/4225#issuecomment-725603841

I think the most minimal thing to do is:

1. Just to "atomic" resolved dependency mappings for now:
2. Allow them to be individually copied between stores with a separate command
3. Don't worry about making it easy to copy all the resolved derivation mappings needed to re-resolve and unresolved derivation

Yes, we don't want users to have to actually do this tedious stuff in practice, but I think there are good building blocks to have in the end too, and by starting with that, we avoid anything with multiple competing designs.

regnat:
@Ericson2314 
 
> I think the most minimal thing to do is:
> 
>     1. Just to "atomic" resolved dependency mappings for now:

What do you mean exactly by “atomic”? The way I understand it is “without bothering with the dependencies”, but that's already what this PR does.

>     2. Allow them to be individually copied between stores with a separate command

That's an interesting point. I overloaded nix-copy for that because
1. It's not a lot more complicated than having a different command
2. It's actually quite easy for the tests
3. It's quite likely that this is the (low-level) interface we'll want eventually, so it makes sense to do it right now rather than adding a new command that we'll have to deprecate (or remove, after all it's just experimental) in a few months.

But there's arguments to be made in the other directions, so…

>     3. Don't worry about making it easy to copy all the resolved derivation mappings needed to re-resolve and unresolved derivation
 
I feel like it would make things noticeably more complex. Even if just for the tests, it's quite nice to be able to expect `nix copy --to $REMOTE ./foo.nix && clearStore && nix copy --from $REMOTE ./foo.nix` to work, that gives a nice and easy to check property (I'm secretly dreaming of adding property-based testing to Nix)

> Another idea i was thinking was leaving the current DerivationOutputs as-is, and having a new table for CA derv mappings. Queries can union the two tables together, and this means we can continue having the foreign key and simple cleanup for input-addressed derivations, and we separate the "normative" CA mappings from the purely-for-caching IA ones. Thoughts?

As a matter of fact I've been discussing with Eelco yesterday and we reached the same conclusion (though we might not need to do any union as it's totally fine for the two tables to overlap). The nice side of this is that it makes it easier to iterate on the ca-specific schema without touching the rest

regnat:
I've had a chat with @Eelco yesterday, which led to a few suggestions for this:

1. We could (as @Ericson2314 suggested above) add a separate table in the db to track the output mappings of the CA derivations (to make it easier to evolve that schema independently of the rest). That way we can even have a second version scheme for the ca-specific db aspects, allowing people to easily go back-and-forth between ca and mainline Nix
2. `queryDrvOutputInfo` and friends should be
    1. Async (like `queryPathInfo`)
    2. Cached in memory (and possibly on-disk, though that's less urgent)

Ericson2314:
Glad we're all on the same page for multiple tables!

> What do you mean exactly by “atomic”? The way I understand it is “without bothering with the dependencies”, but that's already what this PR does.

Right, but my big schick is that that it's sketchy to have mappings of *un*resolved derivations without dependencies.

> What do you mean exactly by “atomic”? The way I understand it is “without bothering with the dependencies”, but that's already what this PR does.

Well the main thing is if we don't have dependencies and restrict ourselves to resolved derivations, we would have a hard time overloading it.

(I also think it is good to allow just copying outputs, derivations, and mappings independent, but this is less important; I'm not disagreeing that copping the data and mappings is a good default once we can support it.)

> Even if just for the tests

It does make testing harder, because it's such a cumberson low-level interface. I think that's just the (IMO worth it) price of making a minimal thing with as uncontroversial design as we can.

---------

Basically taking a step back, because I'm very keen on keeping the full provenance of resolved derivation mappings around, I think it's just going to be more work before we get to good usable interfaces. But I'm OK with that.

Issue #4228, opened 2020-11-06
nix develop does not print output in GitLab CI


mschwaig:
**Describe the bug**

I'm trying to use the new nix command in CI.

If I run `nix develop nixpkgs#tree --command tree` locally, it produces the following output
```sh
warning: unknown setting 'extra-sandbox-paths'
.

0 directories, 0 files
```

If i run the same thing in a GitLab CI Runner, it produces the following output

```sh
Running with gitlab-runner 13.5.0 (v13.5.0)
  on shell_nixos-gitlab-runner_6a9b50e2300a 9YssniSh
Preparing the "shell" executor
00:00
Using Shell executor...
Preparing environment
00:00
Running on nixos-gitlab-runner...
Getting source from Git repository
00:01
Fetching changes with git depth set to 50...
Initialized empty Git repository in /var/lib/private/gitlab-runner/builds/9YssniSh/0/mschwaig/gitlab_ci_output_test_flake/.git/
Created fresh repository.
Checking out f477a4b2 as try_nix-host_runner...
Skipping Git submodules setup
Executing "step_script" stage of the job script
00:10
$ nix develop nixpkgs#tree --command tree
warning: unknown setting 'extra-sandbox-paths'
Cleaning up file based variables
00:00
Job succeeded
```

So there is no output from the actual `nix develop command`.

I ran into this when running actual cargo and python tests via `nix develop` so it's not specific to the tree command. I just could not think of a better minimal example.

**Steps To Reproduce**

I have published a repo with an assigned CI Runner that produces this error. You can [take a look at that repo](https://gitlab.com/mschwaig/gitlab_ci_output_test_flake) and [the output CI pipline in GitLab](https://gitlab.com/mschwaig/gitlab_ci_output_test_flake/-/jobs/834663532).

The runner I'm using is a NixOS based GitLab Runner that I have set up in the cloud. The builds I originally ran were run on a completely different host, but I could reproduce it with my own Runner, so I think you should be able to too.

I am happy to try out a few suggested modifications or provide access to my setup. [Here is the config for the runner](https://gist.github.com/mschwaig/e65984fff6940f68aea7806c2e7b38a5).

**Expected behavior**

Same output to the terminal in CI as well as I got when running locally.

**Additional context**

I have created a few different kinds of Runners wit the GitLab Runner module in nixpkgs. Each branch in my repo uses a different kind of runner, but the all run into the same issue.

I never got any output from cargo. Python would at least give me some output when the a builder failed, like so:
[...]

PR #4233, opened 2020-11-07
nix develop: Preserve stdin with `-c`

Kha:
For use in editor integration, I would like to run a derivation taking input from stdin instead of $src, basically
```
echo "print('hi')" | nix develop .#foo -c python -
```
However, this currently breaks because of `nix develop`'s use of `--rcfile`: the stdin is executed by the shell instead of being passed to the `-c` command. Not using `--rcfile` with `-c`, just like it's already being done with `--phase`, fixes this.

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -474,9 +474,9 @@ struct CmdDevelop : Common, MixEnvironment
             ignoreException();
         }
 
-        // If running a phase, don't want an interactive shell running after
+        // If running a phase or single command, don't want an interactive shell running after
         // Ctrl-C, so don't pass --rcfile
-        auto args = phase ? Strings{std::string(baseNameOf(shell)), rcFilePath}
+        auto args = phase || !command.empty() ? Strings{std::string(baseNameOf(shell)), rcFilePath}
             : Strings{std::string(baseNameOf(shell)), "--rcfile", rcFilePath};
 
         restoreAffinity();
--- a/tests/nix-shell.sh
+++ b/tests/nix-shell.sh
@@ -58,8 +58,12 @@ output=$($TEST_ROOT/shell.shebang.rb abc ruby)
 
 # Test 'nix develop'.
 nix develop -f shell.nix shellDrv -c bash -c '[[ -n $stdenv ]]'
-# Preserve stdin with `-c`
+
+# Ensure `nix develop -c` preserves stdin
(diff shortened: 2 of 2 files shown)


src/nix/develop.cc
```diff
@@ -437,7 +437,7 @@ struct CmdDevelop : Common, MixEnvironment
             std::vector<std::string> args;
             for (auto s : command)
                 args.push_back(shellEscape(s));
-            script += fmt("exec %s\n", concatStringsSep(" ", args));
+            script += concatStringsSep(" ", args);
```
regnat, line 440: Why is this needed?

tests/nix-shell.sh
```diff
@@ -58,6 +58,7 @@ output=$($TEST_ROOT/shell.shebang.rb abc ruby)
 
 # Test 'nix develop'.
 nix develop -f shell.nix shellDrv -c bash -c '[[ -n $stdenv ]]'
+echo foo | nix develop -f shell.nix shellDrv -c cat  # preserve stdin with `-c`
```
regnat, line 61:
```suggestion
echo foo | nix develop -f shell.nix shellDrv -c cat  # preserve stdin with `-c` | grep -q foo
```

(to make sure that the command is actually ran)

Also, this might also #4228 . If it does, do you mind also adding a test-case for it? Something like

```sh
# Ensure that `nix develop --command` actually executes the command if stdout isn't a terminal
echo "" | nix develop nixpkgs#hello --command echo foo |& grep -q foo
```

regnat (changes requested):
Nice :) 

Left a couple of comments, but looks good otherwise

src/nix/develop.cc
Kha: `exec` isn't needed *anymore*, so I thought I may as well simplify the code (since it's not being used for `--phase` either). Before, it was necessary for exiting the then-interactive shell.

tests/nix-shell.sh
Kha: Hah, indeed it does! Added.

src/nix/develop.cc
regnat: I don't have a strong opinion on this, but it seems cleaner to keep it, if only to avoid having a useless bash process lying around while the command is running

Kha: Yeah, why not. Reverted.

regnat (approved):

Issue #4227, opened 2020-11-06
Test the db migration


regnat:
**Is your feature request related to a problem? Please describe.**

Nix can transparently update the local db if its internal schema is newer than the existing one. But there's no test for it.

**Describe the solution you'd like**

If #4226 gets implemented, we could also add a test that creates and fill a db with the old Nix and tries accessing it with the new one to ensure that everything has been correctly migrated.

regnat:
Fixed by #4239

regnat:
Fixed by #4239

PR #4239, opened 2020-11-09
Add a CI check to ensure compatibility with an old daemon

### LINKED ISSUE #4226: Test talking to an older daemon

There's no way atm to test that Nix master will behave correctly if the daemon is older (or vice-versa, but that's probably less an issue). Yet a failure due to a bad interaction between different versions pops-up every so often.

Add a test (at least a CI job if it's too complicated to do that locally) that runs (a subset of) the testsuite using older daemons (at least the latest released version, dunno if it's worth adding more).

regnat:
Run the `remote-store` test on the CI where `nix-daemon` is replaced by the nix provided in the global environment (assumed to be the latest stable release)

Fix #4226

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -150,6 +150,11 @@
       # 'nix.perl-bindings' packages.
       overlay = final: prev: {
 
+        # An older version of Nix to test against when using the daemon.
+        # Currently using `nixUnstable` as the stable one doesn't respect
+        # `NIX_DAEMON_SOCKET_PATH` which is needed for the tests.
+        mainstream-nix = prev.nixUnstable;
+
         nix = with final; with commonDeps pkgs; stdenv.mkDerivation {
           name = "nix-${version}";
           inherit version;
@@ -157,6 +162,8 @@
           src = self;
 
           VERSION_SUFFIX = versionSuffix;
+
+          OUTER_NIX = mainstream-nix;
 
           outputs = [ "out" "dev" "doc" ];
 
@@ -485,6 +492,8 @@
 
         stdenv.mkDerivation {
           name = "nix";
+
+          OUTER_NIX = mainstream-nix;
 
           outputs = [ "out" "dev" "doc" ];
 
--- a/flake.nix
+++ b/flake.nix
@@ -144,6 +144,32 @@
             echo "file installer $out/install" >> $out/nix-support/hydra-build-products
           '';
 
+      testNixVersions = pkgs: client: daemon: with commonDeps pkgs; pkgs.stdenv.mkDerivation {
+        NIX_DAEMON_PACKAGE = daemon;
+        NIX_CLIENT_PACKAGE = client;
+        name = "nix-tests-${client.version}-against-${daemon.version}";
+        inherit version;
+
+        src = self;
+
+        VERSION_SUFFIX = versionSuffix;
+
+        nativeBuildInputs = nativeBuildDeps;
+        buildInputs = buildDeps ++ awsDeps;
+        propagatedBuildInputs = propagatedDeps;
+
+        enableParallelBuilding = true;
+
+        dontBuild = true;
+        doInstallCheck = true;
+
+        installPhase = ''
+          mkdir -p $out
+        '';
+        installCheckPhase = "make installcheck";
+
+      };
+
     in {
 
       # A Nixpkgs overlay that overrides the 'nix' and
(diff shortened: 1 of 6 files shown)


regnat:
This might of course cause some issues if the `remote-store` test starts testing stuff that's not available with the latest stable release. But it's not the case atm and it seems fair to keep things that way

edolstra:
Can you add this as a Hydra job? That (rather than Github actions) is the canonical test suite for Nix. It's also easier for developers since you can run a test using nix-build.

regnat:
> Can you add this as a Hydra job? That (rather than Github actions) is the canonical test suite for Nix. It's also easier for developers since you can run a test using nix-build.

I've actually just rewritten it to run as part of the standard testsuite (peeking the reference nix daemon from an environment variable that's set in `flake.nix`). Having it as a hydra job would be even nicer, but I don't think I could easily reuse the existing tests in that case (or could I?), while here I just have a test that sources `remote-store.sh` with an overriden nix-daemon which means that we have a reasonably good coverage at a low price

roberth:
I like the approach Eelco suggested. It allows for more combinations to be tested, and to reuse the whole test suite.

> reasonably good coverage at a low price

The price is always low when tests run on CI. You don't need to test compatibility when iterating on a feature, but you can still do it with `nix-build` or a flake command when you're working on the protocol.

It'd look like this:

###### old daemon

 - Allow specifying the old nix package. This is your `OUTER_NIX`. Perhaps you could call it call it `NIX_DAEMON_PACKAGE`
   - The normal package does not specify this variable
   - In the `checks` we add the normal package with this variable as an override
 - The test suite always uses `NIX_DAEMON_PACKAGE`'s daemon when set, so the whole test suite will be run against the old daemon
 - Test cases that require a new daemon are in a conditional based on the daemon version or protocol version

###### old client

 - Similar to "old daemon", but instead of overriding based on the current normal package, we import an older Nix revision using flakes and then set the daemon to be the current package

###### version combinations

 - Derivations per old version approach allows multiple version combinations to be tested
 - Most of the value is in having one older version to run against. It doesn't even have to be old, just older. We can retain relevant versions to run against as development progresses
 - We should probably check against Nix 2.3 too but that can be a separate issue. Backporting test suite changes seems perfectly acceptable to me
 - Mistakes in the test suite can be corrected on the maintenance branches
 - Irrelevant failures can be patched away if necessary (still beats not testing anything at all)

regnat:
> The price is always low when tests run on CI. You don't need to test compatibility when iterating on a feature, but you can still do it with `nix-build` or a flake command when you're working on the protocol.

Yes sorry, I didn't mean “low price” as “doesn't require too much computation” but rather as “easy to implement and maintain”.

> I like the approach Eelco suggested. It allows for more combinations to be tested, and to reuse the whole test suite.

Well, my point is that I don't think we can (easily) reuse the test-suite because it's currently tied to the actual build process − the test setup points `PATH` to the directory that has been defined by the `configure` script, `make installcheck` depends on building a Nix plugin with hardcoded include paths to `src/` etc…

In practice it means that we can't take an arbitrary `nix` and run the test suite against it − hence the solution of just adding one extra test with a hack to get an external `nix-daemon`.

I agree that I'd be better in the long run to indeed have it as a matrix of hydra checks, but that would be a longer term goal as it would first require un-coupling the install-testsuite from the build process (and do so without breaking the development cycle which − at least for me − relies a lot on being able to run `make install && make installcheck` in a row).

regnat:
I've updated this a little to

1. Make it actually work in case there's a schema change (in the previous version, some commands bypassing the daemon were sneaking in the middle, triggering a db migration so that the daemon was unable to read the database for the rest of the test)
2. Add a test for the db migration (fixing #4227 )

regnat:
@edolstra is there anything you'd like to see on this?

Ericson2314:
Yeah @edolstra very much looking forward to daemon protocol code changes being less scary after this!

Ericson2314:
Catching and fixing some things in https://github.com/NixOS/nix/pull/4588 only makes me want this more!

regnat:
I've rebased this on top of master and reworked a bit the way it works:

1. It is now possible to run the `installcheck` independently from the rest of the build, by reusing an external Nix
2. It's also possible to use a different Nix version for the client and the daemon in `installcheck`. In that case, the db migration between the two versions will also be checked
3. There's a new `installTests` flake test that runs the install tests with
    1. The current Nix version both as client and daemon
    2. The current Nix version as client and an older one as daemon

Issue #2259, opened 2018-06-27
Manual's build example doesn't build


pushcx:
[Chapter 13. A Simple Nix Expression](https://nixos.org/nix/manual/#ch-simple-expression) (from `doc/manual/expressions/simple-expression.xml`) introduces building packages with a worked example of building GNU hello. I attempted to use this to learn to build nix packages and was unable to build the package. I don't know where I went wrong, so this bug report is something of a step-by-step. (For context, I'm a longtime professional linux sysadmin and developer who's totally new to nix. I was able to set up a NixOS box from its manual to play around with.)

The chapter is structured around three steps: writing a nix expression (13.1), writing a build script (13.2), and adding the package (13.3).

13.1 doesn't explain where to create `default.nix`, but from the note "It is customary to place each package in a separate directory" I figured I should create a `test` directory with this `default.nix` in it. In 13.2 I also put `builder.sh` in this directory.

Then instead of the section heading for 13.3 ("Arguments and Variables") matching the third bullet point in the intro, it's kind of a random name. It talks about editing `all-packages.nix` but the syntax is invalid. As a newbie, I don't have the experience to fill in `...` so the example is unusable. Additionally it talks about editing `pkgs/top-level/all-packages.nix` like that directory already exists, and the only file by that name that `find` turned up is `/nix/store/ilagh0i1vxhp68hd9r73lh0f2mjqbj5x-nixos-18.09pre143771.a8c71037e04/nixos/pkgs/top-level/all-packages.nix`. The description of the nix store in 5.3 has a wishy-washy warning against editing files in `/nix/store` and I don't see any way to regenerate that hash (or maybe I don't need to? The introduction to the hash at the top of chapter 1 says it's only based on dependencies, not contents). Still, it isn't clear at all what I'm supposed to be doing to what file, or why this is required for building hello.

Then section 13.4 starts "You can now try to build Hello." and `nix-build -A hello` looks like it's assuming I put those files in a directory named `hello`, though after renaming my directory I got the error `error: getting status of '/home/pushcx/default.nix': No such file or directory` from outside the directory and from inside I got `error: cannot auto-call a function that has an argument without a default value ('stdenv')` (with no file name or line-number). The error message is useless because I'm trying to build a package, not auto-call a function, and "auto-call" does not appear in the manual. I don't see anything more to experiment with to try to fix this.

After chatting with infinisil in #nixos, I think the two causes of my confusion are:

1. The manual does not provide a complete, worked example of building a package.
[...]

PR #4240, opened 2020-11-10
2259 error message - "auto-call" error

bburdette:
revised this "auto-call" error which I've found to be somewhat inscrutable in the past.  This is intended to address (in part) issue #2259.

- UndefinedVarError seemed more appropriate than TypeError
- added Pos
- big error message with examples.
- link to a nix pill for further explanation.

Current output is like this:

image

Revised error output is like so:
 
image

This reflect my perhaps imperfect understanding of autoCall, feedback on that is welcome.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1375,7 +1375,7 @@ void EvalState::autoCallFunction(Bindings & args, Value & fun, Value & res)
             } else if (!i.def) {
                 throwMissingArgumentError(i.pos, R"(cannot evaluate a function that has an argument without a value ('%1%')
 
-Nix attempted to evaluate a function as a top level expression; in this case it must have all its 
+nix attempted to evaluate a function as a top level expression; in this case it must have its
 arguments supplied either by default values, or passed explicitly with --arg or --argstr.
 
 https://nixos.org/manual/nix/stable/#ss-functions)", i.name);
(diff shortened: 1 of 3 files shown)


bburdette:
Some slight changes:

image

edolstra:
I think very long error messages should be avoided since then we just end up duplicating the manual. The error message can also get out of date, for instance it talks about `nix-build` and a specific (non-flake) style of importing Nixpkgs. It may also lead the user astray that we're talking about `callPackage` but the error message might have nothing to do with `callPackage`.

 It's probably better to have a shorter message that links to the manual for more information (*not* Nix pills which are not an authoritative reference for Nix).

src/libexpr/eval.cc
```diff
@@ -1370,7 +1365,31 @@ void EvalState::autoCallFunction(Bindings & args, Value & fun, Value & res)
             if (j != args.end()) {
                 actualArgs->attrs->push_back(*j);
             } else if (!i.def) {
-                throwTypeError("cannot auto-call a function that has an argument without a default value ('%1%')", i.name);
+                throwUndefinedVarError(i.pos, R"(cannot auto-call a function that has an argument without a default value ('%1%')
+
+An 'auto-call' is when a nix expression is evaluated without any external arguments.
```
edolstra, line 1370: This is not correct. An auto-call is when a top-level function is called *with* external arguments *or* with default values if no matching argument is provided.

It's probably better to say something like `you need to pass "--arg %1% ...'  or '--argstr %1% ...'".

edolstra: `UndefinedVarError` is probably misleading. There is no undefined variable, just a missing argument.

bburdette:
Thanks for the feedback!

- pared down the error message
- error is more generic rather than callPackage centric
- docs link goes to function reference in the manual.
- added a MissingArgumentError rather than using TypeError or UndefinedVarError.

Here's the result:

image

bburdette:
Thanks for the merge!

PR #4263, opened 2020-11-16
libstore: always use the socket if it exists

zimbatm:
This change modifies the behaviour of Nix to always communicate with the
daemon, even if the store is writable to that user.

Bypassing the daemon had a number of issues. Mainly, the client
behaviour is not always consistent between users leading to confusion.
The max-jobs counter is not shared between various instances of the Nix
client. The clients fight between each-other to hold the lock on the
sqlite database.

By making the access consistent, it opens the road for the daemon to
permanently hold the lock on the sqlite database, and get rid of the
"sqlite is busy" type of errors.

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -1042,9 +1042,7 @@ std::shared_ptr<Store> openFromNonUri(const std::string & uri, const Store::Para
 {
     if (uri == "" || uri == "auto") {
         auto stateDir = get(params, "state").value_or(settings.nixStateDir);
-        if (access(stateDir.c_str(), R_OK | W_OK) == 0)
-            return std::make_shared<LocalStore>(params);
-        else if (pathExists(settings.nixDaemonSocketFile))
+        if (pathExists(settings.nixDaemonSocketFile))
             return std::make_shared<UDSRemoteStore>(params);
         else
             return std::make_shared<LocalStore>(params);


edolstra:
This makes Nix less efficient when running as root, by requiring an unnecessary redirection through the daemon.

> By making the access consistent, it opens the road for the daemon to permanently hold the lock on the sqlite database, and get rid of the "sqlite is busy" type of errors.

Not until we have a multi-threaded, non-forking daemon...

zimbatm:
> This makes Nix less efficient when running as root

Do you use `sudo` to improve performance during a build? On my systems, building as a user is the common case so it's not really an optimization I am benefiting from. It's important to remove as many branches as possible in the code.

edolstra:
No, but Hydra connects to most build slaves as root.

zimbatm:
This doesn't feel right. Maybe the daemon could be disabled on those machines but that doesn't strike me as a good solution either.

Would you be willing to re-consider that decision if the daemon overhead was reduced sufficiently?

andir:
What is the overhead that you are talking about? I assume it is all the
store data going through the Unix socket before going to the filesystem?

zimbatm:
I guess the next step would be to write some sort of benchmarks then.

7c6f434c:
Having an option to run something one time with a different NIX_CONF_DIR and nix.conf without messing with the main daemon is what I find a feature, not a bug…

I'd assume that for `nix-instantiate` the context-switching of the daemon use is more important than how the data is copied?

edolstra:
The overhead is having to start another process and send all operations to it via RPC.

What's the actual problem here? I'm not really aware of issues caused by *not* going through the daemon. In particular

> The max-jobs counter is not shared between various instances of the Nix
client. The clients fight between each-other to hold the lock on the
sqlite database.

is not solved by this patch because the daemon worker processes don't share max-jobs or SQLite locks either.

zimbatm:
I'd like to re-open this for consideration.

Over the years I have seen many cases of users confused about this behavior. Especially on non-NixOS distros where the nix-daemon doesn't auto-reload when the nix.conf file is changed. They configure remote builders or change settings, try as root, and it works, and then get confused why it doesn't work as a user.

Regarding Hydra, we would have to ask the infra team, but i don't think that's the current performance bottleneck.

Issue #3931, opened 2020-08-13
nixFlakes's nix-build segfaults in boehm-gc


kalbasit:
**Describe the bug**

I'm using Nix version 2.4pre20200721_ff314f1, and I'm trying to build my host (does not use Flakes). It crashes randomly with over 90% reproducibility. I can only reproduce with nix-build running as root.

**`nix-env --version` output**

```
nix (Nix) 2.4pre20200721_ff314f1
```

**Steps to reproduce**

- Clone https://github.com/kalbasit/shabka at kalbasit/shabka@b314d3a91a7d9c7b34fe153040cf3f4ec14aad6d
- Clone https://github.com/kalbasit/dotshabka at kalbasit/dotshabka@210a10a00132f8a6ec5e5155ec44fa99be1f0a8c
- In shabka, run the following command: `nix-build --no-out-link -A external.nixpkgs.release-20-03.path` and record its output.
- Run `sudo --preserve-env=RELEASE gdb --args nix-build {path_from_above}/nixos -A system -k -I nixos-config={path_to_dotshabka}/hosts/achilles/configuration.nix --show-trace`

**Additional context**

The following is only a part of the trace. The entire trace is available [here](https://gist.github.com/b29028edaf14d8761f2872c99cc78d67).

<details><summary>gdb backtrace</summary>

```
(gdb) run
Starting program: /nix/store/q3lp9d5s9c0jlkw3x84w029zmgnlqx29-nix-2.4pre20200721_ff314f1/bin/nix-build /nix/store/8nssa19jsrw2x1h9fryqwsqac07005ii-nixpkgs-release-20-03-4364ff933ebec0ef856912b182f4f9272aa7f98f/nixos -A system -k -I nixos-config=/yl/code/repositories/github.com/kalbasit/dotshabka/hosts/achilles/configuration.nix --show-trace
(gdb) bt

Thread 1 "nix-build" received signal SIGSEGV, Segmentation fault.
0x00007ffff7ad1d40 in GC_push_all_eager () from /nix/store/krbh984z3lpvj614gfp2ss59b8ss2nq1-boehm-gc-8.0.4/lib/libgc.so.1
(gdb) bt
#0  0x00007ffff7ad1d40 in GC_push_all_eager () from /nix/store/krbh984z3lpvj614gfp2ss59b8ss2nq1-boehm-gc-8.0.4/lib/libgc.so.1
[...]

PR #4264, opened 2020-11-16
BoehmGCStackAllocator: ignore stack protection page

roberth:
Fix potential cause of https://github.com/NixOS/nix/issues/4178#issuecomment-727750030

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -233,15 +233,25 @@ class BoehmGCStackAllocator : public StackAllocator {
     std::max(boost::context::stack_traits::default_size(), static_cast<std::size_t>(8 * 1024 * 1024))
     };
 
+    // This is specific to boost::coroutines2::protected_fixedsize_stack.
+    // The stack protection page is included in sctx.size, so we have to
+    // subtract one page size from the stack size.
+    std::size_t pfss_usable_stack_size(boost::context::stack_context &sctx) {
+      return sctx.size - boost::context::stack_traits::page_size();
+    }
+
   public:
     boost::context::stack_context allocate() override {
         auto sctx = stack.allocate();
-        GC_add_roots(static_cast<char *>(sctx.sp) - sctx.size, sctx.sp);
+
+        // Despite what the docs warn, the only *implemented* stack direction
+        // is "down". This is why we subtract the stack size.
+        GC_add_roots(static_cast<char *>(sctx.sp) - pfss_usable_stack_size(sctx), sctx.sp);
         return sctx;
     }
 
     void deallocate(boost::context::stack_context sctx) override {
-        GC_remove_roots(static_cast<char *>(sctx.sp) - sctx.size, sctx.sp);
+        GC_remove_roots(static_cast<char *>(sctx.sp) - pfss_usable_stack_size(sctx), sctx.sp);
         stack.deallocate(sctx);
     }
 
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -244,14 +244,41 @@ class BoehmGCStackAllocator : public StackAllocator {
     boost::context::stack_context allocate() override {
         auto sctx = stack.allocate();
 
-        // Despite what the docs warn, the only *implemented* stack direction
-        // is "down". This is why we subtract the stack size.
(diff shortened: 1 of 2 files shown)


roberth:
Update: Boehm GC seems to be incompatible with coroutines. I haven't gotten to the bottom of this, but disabling GC during source addition seems like an acceptable workaround.
For details, see the inline comments.

edolstra:
Eh. That seems super ugly. It's probably better to go back to the old situation (i.e. not having GC roots on coroutine stacks).

roberth:
> Eh. That seems super ugly.

I agree. I wish it was simple.

> It's probably better to go back to the old situation (i.e. not having GC roots on coroutine stacks).

What would that old situation be?

I know #4030 introduced a case of evaluation in coroutines but #3931 predates that.
For #4030 this can be accomplished by enumerating the files before dumping to the daemon, but that doesn't attempt to fix the problem causing #3931.

It seems likely that the mere existence of a coroutine can crash the collector. The gdb stack traces all point to the collector code that finds roots in the stacks of all threads.

These seem to be the options to move forward:
 1. use threads instead of coroutines. This could be a drop-in replacement, but probably has a slight performance cost.
 2. use callbacks instead of coroutines. A lot of work, less readable, better performance.
 3. investigate precisely why the collector crashes and fix the real issue.
 4. figure out what else might cause #3931. Might lead to another case of eval on coroutine or some new insight.
 5. try to avoid GC allocations while any coroutines are running, which we won't be able to guarantee; crash will resurface.
 6. try to avoid GC allocations while any coroutines are running and disable the collector while we know a coroutine exists. Like previous but avoids crash. Instead causes a space leak while any coroutine exists. (this PR + enumerate files before dumping)
 7. something else?

I hope this can be fixed soon. 1 and 6 seem to be the shortest paths to a Nix that doesn't crash.

Ericson2314:
Honestly, we might consider replacing coroutines with plain old threads?

Ma27:
Yesterday I debugged a nixops deployment where the evaluation resulted in a segfault (that was hidden by `nixops` not showing the output). Finally, a minimal NixOS configuration to reproduce the problem with `nix (Nix) 2.4pre20201118_79aa7d9` was this:

``` nix
{
  minimal_segfault_example = { pkgs, lib, config, ... }: {
    environment.etc."foo".text = "${builtins.path {
      path = pkgs.path;
      filter = pkgs.lib.cleanSourceFilter;
      name = "nixpkgs-${pkgs.lib.version}";
    }}";
  };
}
```

Long story short, this PR appears to solve the problem. In case anybody else stumbles upon this issue as well, the workaround is quite easy, just do `nix-shell -p nix` and re-run `nixops deploy`.

Unfortunately I didn't closely follow the changes related to that, so I don't have an actual opinion on how to proceed here.

cc @lheckemann

jonringer:
For others who land on this page.

Until there's a proper fix, you may try adding `GC_DONT_GC=1` to the command.

```
GC_DONT_GC=1 nix build ...
```

this seems to allow IFD to complete successfully for me

dhess:
I suspect this is still a problem.

roberth:
We might have learned something more if we'd merged the first commit. We can still do that.

At least that's a proper fix for the problem where GC causes an apparent "stack overflow".

-----

Crashes should be top priority, right? For me they are, but the lack of action from the maintainers of the project on this issue has made me follow their example. I regret that, but I'm glad I could avoid an alternative that would have been worse.

If only someone could make a decision, that would be nice.

Ma27:
My example from [above](https://github.com/NixOS/nix/pull/4264#issuecomment-733189235) still segfaults on 2.4 and evaluates fine on 2.3, so this is still a problem.

That said I'm starting to wonder if it was a good idea to "disband the Nix core team" (RFC0044) in favor of the steering committee. The first team was solely "responsible" for the development of Nix itself, the second one is responsible for the development of the entire ecosystem. Having an unfixed crash on a - well - unstable, but *heavily & officially* used version of Nix for months - even though there's a working fix - is IMHO a symptom of that, just as a rather non-transparent roadmap is.

Don't get me wrong - I use nixUnstable myself and I'm fairly grateful for the work that has been done here, but I'm starting to ask myself if such a team that used to bring in a certain kind of structure is what's missing in the ongoing Nix 2.4 development.

But anways, cc @edolstra for the actual problem.

roberth:
Done in #4944

PR #4267, opened 2020-11-17
Track the dependencies of derivation outputs

### LINKED ISSUE #4249: Track the dependencies of derivation outputs
Currently Nix knows the runtime dependencies of a path because each entry in the database keeps a list of its references.
This is enough for input-addressed derivations as having their output path is enough for most intent and purposes.

However, this doesn't take us all the way in a ca world. For example, `nixpkgs#hello` depends on `nixpkgs#glibc`. Currently if one copies `nixpkgs#hello` to another machine, this also copies `nixpkgs#glibc`, meaning that `nix build nixpkgs#glibc` on the remote machine will be a no-op.
But if `glibc` was content-addressed, this wouldn't work because the store on the remote machine would only have the output path of `glibc`, but not the mapping `glibc!out -> /nix/store/...-glibc`.

To fix that, we need to track the dependencies of each derivation output (which will be other derivation outputs or store paths).

regnat:
Builds on top of #4238 

Fix #4249 by keeping track of all the dependencies of each derivation output.

The dependencies of a drv output `foo!out` is defined as the subset of its build inputs that are references of its output path − or said otherwise the minimal set of drv outputs that need to be realised to get the runtime closure of the output path (the motivation for this is given in #4238)


regnat:
Closing in favor of #4836 as this one is severely outdated

PR #4269, opened 2020-11-19
Make drv hash modulo memo table thread-safe

Ericson2314:
Let's get one step closer to the daemon not needing to fork.

CC @roberth


Ericson2314:
Oh neat I'll take a look at that branch.

edolstra:
@Ericson2314 It's this one: https://github.com/edolstra/nix/commits/parallel-eval-2. It has some other thread-safety fixes, but they only affect the evaluator, not the daemon.

Ericson2314:
Thanks!

PR #4271, opened 2020-11-21
Skip Reference Check on bundler

wiltaylor:
The main reason for this change is a lot of bundlers are going to copy paths from the nix store into a bundle and this check will cause them to fail.

My specific use case is to create a .nix folder inside the derivation that can be used with nix-user-chroot from an app image.


edolstra:
I think it's better to remove the check altogether. Bundlers can use `allowedReferences = []` if they want to disallow references.

wiltaylor:
Thanks, that will actually make my life a lot easier too :). Made the changes.

PR #4275, opened 2020-11-23
fetchMercurial: set HGPLAIN when invoking hg

lukegb:
Without setting HGPLAIN, the user's environment leaks into
hg invocations, which means that the output may not be in the
expected format.

HGPLAIN is the Mercurial-recommended solution for this in that
it's intended for uses by scripts and programs which are looking
to parse Mercurial's output in a consistent manner.


edolstra:
Thanks!

PR #4281, opened 2020-11-24
Escape filename given to nix-shell in shebang mode

### LINKED ISSUE #4229: nix-shell in shebang fails if script name has a space or other shell metacharacters
**Describe the bug**

If my script name has a space in it, using `nix-shell` as a shebang fails as it doesn't escape the space. This also happens if the name contains other shell metacharacters, such as `foo{1,2}.sh`, but a space is the most common way to reproduce this.

**Steps To Reproduce**

Create a new file named `foo bar.sh` with the contents:

```bash
#!/usr/bin/env nix-shell
#!nix-shell -i bash -p bash
echo test
```

Mark it as executable and then execute it (or just pass it to `nix-shell` directly).

**Expected behavior**

It should execute and echo `test`.

**Actual behavior**

```
> ./foo\ bar.sh
bash: ./foo: No such file or directory
```

nix-env (Nix) 2.3.7

lilyball:
This prevents spaces or other metacharacters from causing nix-shell to execute the wrong path.

Fixes #4229.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -217,9 +217,9 @@ static void main_nix_build(int argc, char * * argv)
                 // read the shebang to understand which packages to read from. Since
                 // this is handled via nix-shell -p, we wrap our ruby script execution
                 // in ruby -e 'load' which ignores the shebangs.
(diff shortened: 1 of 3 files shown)


lilyball:
@edolstra This has been sitting open for a while and is rather trivial, could I get this reviewed please?

edolstra:
Thanks!

Issue #3697, opened 2020-06-15
Automatic code formatting for nixos/nix


bburdette:
It would be great if there was a recommended formatting tool and config file for the nix project.  Working on the last nix-errors PR I spent a few days trying to find a tool that would match the existing formatting, but I wasn't able to get anything to work perfectly.  That PR was already huge, so a bunch of noise from reformatting would not have helped matters.  

I tried astyle, uncrustify, and clang-format.  Uncrustify was probably closest - its has a ton of configuration options.  Astyle was also decent, arguably because it mostly leaves the code alone compared to the others.  clang-format was in third place for me, but perhaps it could have been made to work by spending more time on it.  I didn't save my config attempts with that one.  

Here's where I left off with uncrustify:
[uncrustify.cfg.txt](https://github.com/NixOS/nix/files/4781475/uncrustify.cfg.txt)

And here's a diff with the default uncrustify settings:
[uncrustify.diff.txt](https://github.com/NixOS/nix/files/4781489/uncrustify.diff.txt)

Unfortunately I don't appear to have saved my astyle settings, but its way less involved than the above.  I'd like to hear if anyone has had more luck than me with formatting!

gilligan:
This is something i would be very interested in as well. I actually looked into using [clang-format](http://clang.llvm.org/docs/ClangFormat.html) together with @andir just to get an impression of how feasible this is and what options it provides.

That being said the question of which tool to pick is probably secondary so I would love to hear what @edolstra thinks about this **in general** and after that we could talk about what approaches to take and which tools could qualify as candidates.

I believe the value lies in making it
(a) easier to contribute because nobody has to wonder how to format their code
(b) less effort for eelco and other maintainers to review PRs because the formatting is dealt with.

_PS_:
I understand that because of the currently ongoing maintenance of the `flake` branch the introduction of automatic formatting may have to be delayed somewhat but it needs to be discussed anyway.

grahamc:
btw here is a program which lets you reformat code in one branch (master), and cleanly rebase another branch (flakes) on top of it, applying the formatting on each commit, without dealing with any merge conflicts:

https://gsc.io/content-addressed/c508bcd89346cca9f30c40c47251a6674d8a2b2f97e8990d5eb8f62117ec1d12.sh

edolstra:
`uncrustify` looks pretty light-weight. `clang-format` is less ideal because it would add half a gigabyte of clang stuff to the closure when you're not using clang...

[...]

PR #4282, opened 2020-11-25
fix the hash rewriting for ca-derivations

thufschmitt:
Add self-hash rewriting for ca-derivations.
This had apparently been forgotten in #3883, and contrary to what I though there was no test for it (and as it's mostly orthogonal to all the rest I didn't notice it otherwise until now).

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -6,7 +6,7 @@
 #include "globals.hh"
 #include "json-to-value.hh"
 #include "names.hh"
-#include "references.hh"
+#include "path-references.hh"
 #include "store-api.hh"
 #include "util.hh"
 #include "value-to-json.hh"
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -4,7 +4,7 @@
 #include "worker.hh"
 #include "builtins.hh"
 #include "builtins/buildenv.hh"
-#include "references.hh"
+#include "path-references.hh"
 #include "finally.hh"
 #include "util.hh"
 #include "archive.hh"
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -2384,13 +2384,16 @@ SingleDrvOutputs LocalDerivationGoal::registerOutputs()
             if (!outputRewrites.empty()) {
                 debug("rewriting hashes in '%1%'; cross fingers", actualPath);
 
-                /* FIXME: this is in-memory. */
-                StringSink sink;
-                dumpPath(actualPath, sink);
+                /* FIXME: Is this actually streaming? */
+                auto source = sinkToSource([&](Sink & nextSink) {
+                    RewritingSink rsink(outputRewrites, nextSink);
+                    dumpPath(actualPath, rsink);
+                    rsink.flush();
+                });
+                Path tmpPath = actualPath + ".tmp";
+                restorePath(tmpPath, *source);
                 deletePath(actualPath);
-                sink.s = rewriteStrings(sink.s, outputRewrites);
-                StringSource source(sink.s);
-                restorePath(actualPath, source);
+                movePath(tmpPath, actualPath);
 
                 /* FIXME: set proper permissions in restorePath() so
                    we don't have to do another traversal. */
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -2379,14 +2379,14 @@ SingleDrvOutputs LocalDerivationGoal::registerOutputs()
             continue;
         auto references = *referencesOpt;
 
-        auto rewriteOutput = [&]() {
+        auto rewriteOutput = [&](const StringMap & rewrites) {
             /* Apply hash rewriting if necessary. */
-            if (!outputRewrites.empty()) {
+            if (!rewrites.empty()) {
                 debug("rewriting hashes in '%1%'; cross fingers", actualPath);
 
                 /* FIXME: Is this actually streaming? */
                 auto source = sinkToSource([&](Sink & nextSink) {
-                    RewritingSink rsink(outputRewrites, nextSink);
+                    RewritingSink rsink(rewrites, nextSink);
                     dumpPath(actualPath, rsink);
                     rsink.flush();
                 });
@@ -2442,7 +2442,7 @@ SingleDrvOutputs LocalDerivationGoal::registerOutputs()
                         "since recursive hashing is not enabled (one of outputHashMode={flat,text} is true)",
                         actualPath);
             }
-            rewriteOutput();
+            rewriteOutput(outputRewrites);
             /* FIXME optimize and deduplicate with addToStore */
             std::string oldHashPart { scratchPath->hashPart() };
             HashModuloSink caSink { outputHash.hashType, oldHashPart };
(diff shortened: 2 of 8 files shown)


src/libstore/build/derivation-goal.cc
```diff
@@ -3121,6 +3121,20 @@ void DerivationGoal::registerOutputs()
...
+                    StringSink sink;
+                    dumpPath(actualPath, sink);
+                    RewritingSink rsink2(oldHashPart, std::string(newInfo0.path.hashPart()), nextSink);
+                    rsink2((unsigned char *) sink.s->data(), sink.s->size());
+                    rsink2.flush();
+                });
+                Path tmpPath = actualPath + ".tmp";
+                restorePath(tmpPath, *source);
+                deletePath(actualPath);
+                movePath(tmpPath, actualPath);
```
edolstra, line 3136: Maybe a call to `canonicalisePathMetaData()` is needed here. (See `rewriteOutputs()` which does so after doing a rewrite.)

edolstra:
Maybe this can be integrated with `rewriteOutputs()` so we don't have to do multiple rewrites?

Ericson2314:
Wait isn't 
https://github.com/NixOS/nix/blob/13c557fe823549db75c3dc24c99c46e1c4e1378e/src/libstore/build/derivation-goal.cc#L1270-L1272
supposed to do it?

regnat:
> Wait isn't
> 
> https://github.com/NixOS/nix/blob/13c557fe823549db75c3dc24c99c46e1c4e1378e/src/libstore/build/derivation-goal.cc#L1270-L1272
> 
> supposed to do it?

This line is a rewriting of the **input** of the build. What's missing is the final rewrite of the self-references after the build

src/libstore/build/derivation-goal.cc
regnat: Indeed, thanks

Ericson2314:
@regnat sorry, here's the output one
https://github.com/NixOS/nix/blob/05d9442f68ba906ae50c12deab63fc1b9836b149/src/libstore/build/derivation-goal.cc#L3015
But I don't think the `finish` one is being called in enough places?

regnat:
> @regnat sorry, here's the output one
> 
> https://github.com/NixOS/nix/blob/05d9442f68ba906ae50c12deab63fc1b9836b149/src/libstore/build/derivation-goal.cc#L3015
> 
> But I don't think the `finish` one is being called in enough places?

I think it is, but it's called _after_ `rewriteOutput()` (and needs to be, because getting the final hash can only happen after we've rewritten the hashes of the dependencies). We could just call `rewriteOutput()` (once for rewriting the dependencies and once for rewriting the self-references) that should be equivalent − except that we'd need to make sure that we only warn once about the hash rewriting.

src/libstore/build/derivation-goal.cc
Ericson2314: If we give `rewriteOutput` a references parameter above, I think we can deduplciate this?

1. `rewriteOutput(outputRewrites);`
2. calc hahes
3. ```
   if (scratchPath != newInfo0.path)
       rewriteOutput({scratchPath, newInfo0.path});
   ```

regnat: Duh indeed. Thanks, should be fixed now

src/libstore/build/derivation-goal.cc
```diff
@@ -3121,6 +3121,13 @@ void DerivationGoal::registerOutputs()
             newInfo0.references = refs.second;
             if (refs.first)
                 newInfo0.references.insert(newInfo0.path);
+            if (scratchPath != newInfo0.path) {
+                // Also rewrite the output path
+                rewriteOutput({{
+                    std::string(scratchPath.hashPart()),
+                    std::string(newInfo0.path.hashPart())
+                }});
```
Ericson2314, line 3129: Sorry to contradict my own prior suggestion here, but I forgot about the two sets of `{..}`, maybe this would be better to visually distinguish input from output?
```suggestion
                rewriteOutput({
                    { std::string(scratchPath.hashPart()), std::string(newInfo0.path.hashPart()) }
                });
```

src/libstore/build/derivation-goal.cc
```diff
@@ -3121,6 +3121,13 @@ void DerivationGoal::registerOutputs()
             newInfo0.references = refs.second;
             if (refs.first)
                 newInfo0.references.insert(newInfo0.path);
+            if (scratchPath != newInfo0.path) {
+                // Also rewrite the output path
```
Ericson2314, line 3125: Ah it just occurred to me, `refs.first` is whether we have a self reference, so we can use this to optimize slightly.

I also threw in a comment explaining. Feel free to reword.
```suggestion
            if (refs.first) {
                // patch has self-references, so should also rewrite them.
                // Note that this doesn't change the end-goal CA we ca calculated above, since that calculation skips the contents of self-references.
                // But rather, if we have self-references, relocate from `scratchPath`, and then *don't* do this,
                // our self-references to `scratchPath` become non-self references, and then the CA is wrong!
                // Instead we "corrupt" the path at `scratchPath` so when we later relocate it it's thereby uncorrupted, the opposite of the above problem.
```

src/libstore/build/derivation-goal.cc
regnat: I've actually reformated it in yet-another-way :stuck_out_tongue: (which makes me wish for https://github.com/NixOS/nix/issues/3697 to be a think), with an explicit call to the `StringMap` constructor because it's indeed to easy to forget what's going on otherwise

src/libstore/build/derivation-goal.cc
regnat: Indeed, `res.first` is a nicer option (I kept the `schatchPath != newInfo0.path` condition too because I think it's useful to prevent rewriting CA derivations).

src/libstore/build/derivation-goal.cc
Ericson2314: Wait, how does that work, isn't `StringMap` the outer set of braces?

(Fine with the idea though!)

src/libstore/build/derivation-goal.cc
Ericson2314: Looks good to me!

Ericson2314 (approved):
Thanks for catching this mistake of mine @regnat :)

src/libstore/build/derivation-goal.cc
regnat: > Wait, how does that work, isn't StringMap the outer set of braces?

I guess that's why people say that one shouldn't write code too late in the evening 🥴

Ericson2314:
See https://github.com/NixOS/nix/pull/4284#issuecomment-736841994, I guess this one can just be the code de-dup.

Ericson2314:
Still keen on this one, lest the rewrite logic drift apart.

Ericson2314 (approved):

edolstra:
@regnat I guess this is still needed? If so, could you resolve the merge conflict please?

Ericson2314:
It would be good to get this in before 2.4 too, I think.

Ericson2314:
@thufschmitt I rebased this https://github.com/Ericson2314/nix/commits/fix-ca-hash-rewriting it would be good to get it merged!

thufschmitt:
Thanks a lot @Ericson2314 ! I've re-rebased them and force-pushed

Ericson2314:
Weird this failed, but I rebased it yet again.

thufschmitt:
The darwin CI failure looks quite legit and related:

> nix-tests>     error: ca hash mismatch importing path '/build/nix-test/tests/fetchClosure/store/y0af26shqb7ag6jplsjffw3lkbq4391k-dependencies-top';
> nix-tests>              specified: sha256:1k6givbgikzbx4a399vn6y5lvnysa2iw2gfm71fmihqkaqdspdd6
> nix-tests>              got:       sha256:0jdn54c8rcjpv1y5wn1rwqfrfrxli4m2lgiba8qidahmi6mr3slv

Can you have another look?

Ericson2314:
Agreed; but I was thinking it might be easier to fix post rebase?

If you like I can also just open a fresh PR and take this one over if you are busy.

thufschmitt:
Mh sorry, my brain broke down, ignore my last comment. Thanks for the rebase, I'll look at the failing test :)

thufschmitt:
CI is green now, but I had to disable some test or older daemons because this changes the hash of CA paths with self-references (which is a shame, but also very much needed)

Ericson2314 (approved):
Sorry for the belated review!

PR #4284, opened 2020-11-25
Allow fixed-output derivations to depend on (floating) content-addressed ones

regnat:
Fix an overlook of https://github.com/NixOS/nix/pull/4056

### DIFF
--- a/tests/content-addressed.nix
+++ b/tests/content-addressed.nix
@@ -16,14 +16,16 @@ rec {
   };
   rootCA = mkDerivation {
     name = "rootCA";
-    outputs = [ "out" "dev" ];
+    outputs = [ "out" "dev" "foo"];
     buildCommand = ''
       echo "building a CA derivation"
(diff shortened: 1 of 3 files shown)


tests/content-addressed.nix
```diff
@@ -60,4 +60,15 @@ rec {
       echo ${rootCA}/non-ca-hello > $out/dep
     '';
   };
+  dependentFixedOutput = mkDerivation {
```
edolstra, line 63: Shouldn't there be a build of this attribute in `tests/content-addressed.sh`?

regnat: Uh of course :man_facepalming: 

Fixed in 9bd8184f1fb2e91ac4fb7207abe56f2a30a81d97

Ericson2314:
Oh it looks like this contained an older version of #4282 when that one was still being reviewed?

PR #4291, opened 2020-11-29
Add dir attribute to output argument if input specified dir.

kquick:
If the input url specifies a dir component, this PR ensures that dir
specification is passed as an attribute on the argument provided to
the output function for that input:

    inputs.foo.url = "https://github.com/group/proj?dir=sub/dir";
    outputs = { self, foo }:
        assert foo.dir == "sub/dir"; ...

This is useful for various library functions and other operations that
need to know if the input specified a subdirectory relative to the
fetch.

### DIFF
--- a/src/libexpr/flake/call-flake.nix
+++ b/src/libexpr/flake/call-flake.nix
@@ -10,9 +10,13 @@ let
         let
 
           sourceInfo =
-            if key == lockFile.root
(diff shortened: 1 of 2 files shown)


kquick:
@edolstra This is ready for review and merge.

tomberek:
stale: re-open as needed

PR #4296, opened 2020-12-01
Progress indicator improvements

edolstra:
This replaces the current one-line progress bar with a progress indicator with the following features:

* Interactive: you can hit the `L` key to enable display of build output, `v` to increase verbosity, `r` to show which paths are remaining to be built, etc.

* Multi-line: rather than cramming a lot of info into a single line, it has separate progress bars for substitutions, builds, etc.

* More logging: separately from the progress indicator (which only shows transient info), we log to stderr how long successful builds / substitutions took.

* More use of colors.

[asciicast](https://asciinema.org/a/379797)


SuperSandro2000:
Does this preserve color of log messages?

Can we customize the colors and used characters?

edolstra:
If you mean color in build output, then yes ([example](https://asciinema.org/a/euEjvDJxg9ZME3F1Rh2sCt2IZ)).

There is no customization yet.

src/libmain/progress-bar.cc
```diff
@@ -58,6 +81,56 @@ class ProgressBar : public Logger
...
+            stats.left += j.second->expected > j.second->done ? j.second->expected - j.second->done : 0;
+            stats.active++;
+        }
+
+        stats.expected = std::max(stats.expected, act.expected);
+
+        return stats;
+    }
+
+    enum StatusLineGroup {
```
andir, line 118: Wouldn't it make sense to use a `enum class` here for some "free" type-safety?

mweinelt:
I hope this can take some more inspiration from https://github.com/maralorn/nix-output-monitor. I particularly like the list of packages that are currently being built and on which machines that happens.

src/libmain/progress-bar.cc
```diff
@@ -71,53 +144,220 @@ class ProgressBar : public Logger
...
+                    {
+                        auto state(state_.lock());
+                        if (state->prompt) {
+                            state->prompt->set_value(c != '\n' ? c : std::optional<char>());
+                            state->prompt.reset();
+                            continue;
+                        }
+                    }
+
+                    if (c == 'l') {
```
andir, line 251: These `c == 'x'` switches should probably have an explicit `else if` for all the other cases. That makes it explicit that only one of them will be active at a time. Making reading this code easier and more explicit. Another option would be a switch on `c`.

src/libmain/progress-bar.cc
```diff
@@ -71,53 +144,220 @@ class ProgressBar : public Logger
...
     std::thread updateThread;
+    std::thread inputThread;
 
     std::condition_variable quitCV, updateCV;
 
-    bool printBuildLogs;
-    bool isTTY;
+    std::optional<struct termios> savedTermAttrs;
+
+    Pipe inputPipe;
```
andir, line 169: `inputPipe` is only used to cancel the `inputThread`. Is that right? It didn't occur to me in the first passage of reading the thread that was the intend of it. It is a bit unfortunate to have the whole indirection through the pollfd.

src/libmain/progress-bar.cc
```diff
@@ -184,8 +438,8 @@ class ProgressBar : public Logger
             auto sub = getS(fields, 1);
             i->s = fmt(
                 hasPrefix(sub, "local")
-                ? "copying " ANSI_BOLD "%s" ANSI_NORMAL " from %s"
-                : "fetching " ANSI_BOLD "%s" ANSI_NORMAL " from %s",
+                ? ANSI_BOLD "%s" ANSI_NORMAL " from %s"
```
andir, line 441: These two cases appear to be identical.

src/libmain/progress-bar.cc
```diff
@@ -324,174 +603,273 @@ class ProgressBar : public Logger
...
+        state.haveUpdate = false;
+        if (!state.active) return;
+
+        auto width = getWindowSize().second;
+        if (width <= 0) width = std::numeric_limits<decltype(width)>::max();
+
+        std::string s;
+
+        for (size_t i = 1; i < state.prevStatusLines; ++i)
+            s += "\r\e[K\e[A";
```
andir, line 812: What are these escape sequences? Maybe drop a comment what they are doing and why they are here.

andir (commented):
Just did a quick read through some of the code and had a few remarks. I really like the prospect of having this (or something like it) in Nix. Having a better progress indicator for what Nix is doing is great. :+1:

AmineChikhaoui:
Looks interesting. Btw I tried this a bit and seems `nix repl` is broken with this branch.

ryantm:
Looks cool!

One concern I've had with the new nix shell commands is how they work with a `dumb` terminal like an Emacs inferior shell. They just look like they are hanging without any output until they finish. 

It would be nice if we detected dumb terminals and reverted to basic output.

bqv:
Fyi dumb terms are detected, but only by TERM it seems?

E.g. Try running your nix commands with `env TERM=dumb`, which is what I've been doing to workaround the emacs issue

src/libmain/progress-bar.cc
madjar: I was curious so I looked it up. Posting the answer in case anybody else wonders.

Anything of the form `\e[A` is an ansi escape, that is interpreted by the terminal. It's the same kind that is used to output color. '\r' is a carriage return, which puts the cursor back at the beginning of the line.

All put together we have:
- `\r` return the cursor at the beginning of the line
- `\e[K` erases the line from the cursor
- `\e[A` moves the cursor one line up

So, for each line we had in the previous status line, we erase a full line and move the cursor up. Ie, we delete all we printed previously.

bqv:
How's this going?

tomberek:
Interface is nice, but seems to have interfered with `flake show`
```
[root@tomberek:/etc/nixos]# nix flake show
warning: Git tree '/etc/nixos' is dirty
git+file:///etc/nixos
└───nixosConfigurations
    └───tomberek: NixOS configuration

[root@tomberek:/etc/nixos]# /root/nix/result/bin/nix flake show
warning: Git tree '/etc/nixos' is dirty

Type 'h' for help.
```

SuperSandro2000:
I am using this PR for the last days. The download progress is broken for remote builders and it also can't display what it still needs to build.

garbas:
As we are improving the progress indicator we must also solve the problem when progress indication is not desired/needed.

An example of such cases was already reported in https://github.com/NixOS/nix/issues/4402 and https://github.com/NixOS/nix/issues/4506.

In case no TTY is attached to output (eg. when piping to another command) or when running in CI and using `--no-interactive` cli flag, the command should error out in case we reach any interactive prompt. Error should show how to accept/reject/select via cli flags.

collares:
Just so it shows up on search engines: Using this branch with nixos-unstable leads to  "UsageError:
'metadata' is not a recognised command" errors since https://github.com/NixOS/nixpkgs/pull/116807.

lovesegfault:
@edolstra is this getting revived for 2.10?

nrolland:
It might be nice to give this PR some love at nixconf 2022 (?)

tomberek:
Anyone have a rebase of this?

nrolland:
it might be easier to rebase on some commit before [bb411e4  Fix progress bar flicker with -L](https://github.com/NixOS/nix/commit/bb411e4ae16d6a5c61ea595c0c12e2ecee081ff9) which changed quite a bit `src/libmain/progress-bar.cc`

Alizter:
Will 2023 be the year for this?

dev-null-undefined:
> Will 2023 be the year for this?

To be honest isn't https://github.com/maralorn/nix-output-monitor better?
I have been using it for rebuilding my system for quite some time and it seems to work pretty reliably for me.
With really nice output.
I feel like instead of this it would be nicer to create option for using the nix-output-monitor(nom) in the nixos-rebuild script.

Ericson2314:
I started merging newer commits into this.

Ericson2314:
Triaged in the Nix team meeting

- @ericson2314 enqueued this old stalled PR to the board because it seemed to come up in a number of other PRs, some of which were closed in lieu of this one.
  - https://github.com/NixOS/nix/pull/3961
  - https://github.com/NixOS/nix/pull/3972
  - https://github.com/NixOS/nix/issues/5561
  - https://github.com/NixOS/nix/pull/5917

- alternative design we should consider: https://github.com/maralorn/nix-output-monitor

Will discuss further

imincik:
I very much appreciate any effort to improve nix command user interface, but I watched asciinema demo and think I like the output of [nom (Nix output monitor)](https://github.com/maralorn/nix-output-monitor) much more.

Since very long time has passed since this PR was submitted, did you consider taking some experience from this tool ?

infinisil:
@imincik Expand the detailed discussion in the [meeting notes](https://github.com/NixOS/nix/pull/4296#issuecomment-1808221801), this was discussed. In short, nom being written in Haskell makes this very tricky, but if somebody had the resources to rewrite it in C++ and PR it, that would change the game.

szlend:
Imo the default progress indicator should be simple (max 1-3 lines) and most importantly have a fixed output height so the output doesn't bounce around. This is a major usability issue with nom imo ([example](https://asciinema.org/a/7hJXH2iFLEkKxG1lL25lspqNn)). It's fine to have a more advanced progress indicator, but please put it behind a flag.

Alizter:
@szlend I think your concern is a valid one that I believe can be solved with the appropriate configuration, on top of some sane defaults. With UI (especially TUI) you can never satisfy every user. I believe a similar point appears in the meeting notes.

itslychee:
Any update on this PR, would be amazing to have!

jordan-bravo:
It's been ten months since this PR had any activity. Any updates?

Gordon01:
In my opinion, the progress bar in Nix could have a better user experience. It’s informative, clean, and displays most of the useful information — but only after you figure out what those numbers actually mean, since it’s quite hard to guess from the behavior alone. I’m referring to the start of the line, for example: `[6/66/98 built, ...]`.

I believe there’s a lot of room for improvement in how Nix indicates build progress, and I doubt an ideal solution will appear soon. Introducing a “new” progress bar as an alternative to the current one seems like a reasonable step forward.

bahrom04:
2026

Alizter:
> In my opinion, the progress bar in Nix could have a better user experience. It’s informative, clean, and displays most of the useful information — but only after you figure out what those numbers actually mean, since it’s quite hard to guess from the behavior alone. I’m referring to the start of the line, for example: `[6/66/98 built, ...]`.

I would prefer the numbers be explained in the man pages or documentation so that we don't clutter the status for everybody.

PR #4297, opened 2020-12-01
shut up clang warnings

regnat:
- Fix some class/struct discrepancies
- Explicit the overloading of `run` in the `Cmd*` classes
- Ignore a warning in the generated lexer

### DIFF
--- a/src/libexpr/lexer.l
+++ b/src/libexpr/lexer.l
@@ -12,6 +12,8 @@
 
 
 %{
+#pragma clang diagnostic ignored "-Wunneeded-internal-declaration"
+
 #include <boost/lexical_cast.hpp>
 
 #include "nixexpr.hh"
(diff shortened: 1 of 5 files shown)


src/libexpr/lexer.l
```diff
@@ -12,6 +12,8 @@
 
 
 %{
+#pragma clang diagnostic ignored "-Wunneeded-internal-declaration"
```
edolstra, line 15: Note that this causes a gcc warning:
```
src/libexpr/lexer.l:15: warning: ignoring #pragma clang diagnostic [-Wunknown-pragmas]
   15 | #pragma clang diagnostic ignored "-Wunneeded-internal-declaration"
      | 
```

regnat: Obviously :unamused: 

I've opened https://github.com/NixOS/nix/pull/4300 to fix that one

PR #4302, opened 2020-12-02
Adds Nix CLI Guideline to docs

garbas:
As we are working towards Nix 3.0 we want to make sure that we make a
huge step forward in Nix's user experience. And once 3.0 is out of the
door we need to make sure that all future commands and features keep up
the standard of user experience.

This PR adds a CLI guideline document to the Nix documentation. Consider
this document a good starting point and a checklist when somebody will
be (re)implementing commands.

Clearly this guideline does nothing to improve user experience on its
own and can only be useful as long as it is going to be read and
cared for. But it is a first step into that direction.


Ericson2314:
Hmm this takes a sort of "behind the curtains" tone addressing Nix developers not users, though it's in the manual not some `CONTRIBUTING.md`. Might it be rephrased in a more "hi user, you should be able to expect these conventions to be respected across all Nix commands" way?

doc/manual/src/command-ref/cli-guideline.md
```diff
@@ -0,0 +1,571 @@
...
+      and [aligning of text](#text-alignment).
+    - [Autocomplete](#shell-completion) of options.
+    - Show [next possible steps](#next-steps).
+    - Showing some [“tips”](#educate-the-user) when running logs running tasks
+      (eg. building / downloading) in order to teach users interesting bits of
+      Nix ecosystem.
+    - [Help pages](#help-is-essential) to be as good as we can write them
+      pointing to external documentation and tutorials for more.
+- **Tier2**: Commands that are somewhere between Tier1 and Tier2, not really
+  exposing some implementation detail, but not something that we expect a user.
```
andir, line 77: Did you mean `Tier3` instead of `Tier2` here? The sentence doesn't make much sense to me.

edolstra:
@Ericson2314 Good point. It would be better to put this either in a "developer info" appendix (we do already have "hacking" appendix), or in a separate book (which could also include #4280).

doc/manual/src/command-ref/cli-guideline.md
```diff
@@ -0,0 +1,571 @@
...
+
+Rules are there to guide you by limiting your options. But not everything can
+fit the rules all the time. In those cases document the exceptions in [Appendix
+1: Commands naming exceptions](#appendix-1-commands-naming-exceptions) and
+provide reason. The rules want to force a Nix developer to look, not just at
+the command at hand, but also the command in a full context alongside other
+`nix` commands.
+
+```shell
+$ nix [<GROUP>] <COMMAND> [<ARGUMENTS>] [<OPTIONS>]
```
edolstra, line 44: So I tried to apply this today and realized it's confusing. If `nix foo` is a command, then what is `nix`?

Maybe the current terminology ("subcommand") is better.

doc/manual/src/command-ref/cli-guideline.md
```diff
@@ -0,0 +1,571 @@
...
+      pointing to external documentation and tutorials for more.
+- **Tier2**: Commands that are somewhere between Tier1 and Tier2, not really
+  exposing some implementation detail, but not something that we expect a user.
+  From Tier2 command we expect less attention to details, but still some:
+    - Proper use of [colors](#colors), [emojis](#special-unicode-characters)
+      and [aligning of text](#text-alignment).
+    - [Autocomplete](#shell-completion) of options.
+- **Tier3**: Commands that expose certain internal functionality of `nix`,
+  mostly used by other scripts.
+    - [Autocomplete](#shell-completion) of options.
```
edolstra, line 84: Instead of numbered tiers, we currently have the following categories (shown in `nix --help`):

* Main commands
* Infrequently used commands
* Utility/scripting commands
* Commands for upgrading or troubleshooting your Nix installation

We should say here how this corresponds to the tiers.

doc/manual/src/command-ref/cli-guideline.md
```diff
@@ -0,0 +1,571 @@
...
+
+This classification tries to separate commands in 3 categories in terms of
+their importance in regards to the new users. Users who are likely to be
+impacted the most by bad user experience. This does not mean that we will not
+take care of Tier3 commands, it means we will only put more helpful details for
+beginners into Tier1 commands.
+
+- **Tier1**: Commands used for our main use cases and most likely used by new
+  users. From Tier1 commands we expect attention to details, such as:
+    - Proper use of [colors](#colors), [emojis](#special-unicode-characters)
```
Profpatsch, line 67: I did not know there was a proper use of emoji already :slightly_smiling_face:

doc/manual/src/command-ref/cli-guideline.md
```diff
@@ -0,0 +1,571 @@
...
+
+# Help is essential
+
+Help should be built into your command line so that new users can gradually
+discover new features when they need them. 
+
+## Looking for help
+
+Since there is no standard way how user will look for help we rely on ways help
+is provided by commonly used tools. As a guide for this we took `git` and
```
Profpatsch, line 94: Isn’t `git` notoriously famous for having a badly designed command line?

Ericson2314: I think the overall style of nesting and `--help` vs `-h` was popular enough, it's just the specific commands, flags, and groups that are atrocious.

7c6f434c: Or maybe find another reference point… By the way, Monotone supported abbreviations from command much more widely than Git (including, i.e., any unambiguous prefix which is quite convenient)

On the other hand any better reference point would lead to providing plain text output (entry per line) wheneve possible which is apparently too shell-friendly to be allowed in new Nix interface which prefers only JSON to be usable.

doc/manual/src/command-ref/cli-guideline.md
7c6f434c: Separately, while emojis are described for conveying emotions which is not the first thing I expect from a command-line tool, and also outside the range of many nice fonts in case of single-font configuration of a terminal emulator, there are pictograms that are in BMP, have better support, and a lot of precedent of use in technical writing!

garbas:
> @Ericson2314 Good point. It would be better to put this either in a "developer info" appendix (we do already have "hacking" appendix), or in a separate book (which could also include #4280).

This guideline should only be for the experimental `nix` command. Maybe it would be worthwhile to create a separate section Experimental (or something similar) and move all `nix` related documentation there? But, `developer info` would be definitely be better then current location.

doc/manual/src/command-ref/cli-guideline.md
garbas: `nix foo` can either be a `COMMAND` or a `GROUP`. For example `nix store` is a `GROUP` and `nix init` is a `COMMAND` and also `nix store copy` is a `COMMAND`.

I would call `nix` top level command, but I don't really care how we call things as long as it is consistent :)

edolstra:
It's not really about whether this is experimental but about whether it's user-facing, which it isn't - it's only for people who want to contribute to Nix. Let's make an appendix named "Contributing" which can include this and the existing "Hacking" section.

doc/manual/src/command-ref/cli-guideline.md
zimbatm: Let's not forget mercurial. `hg` is known to be more user-friendly than `git` to use.

zimbatm:
Would it make sense to also introduce the notion of ["plumbing" vs "porcelain"](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain)? The old-school nix-* CLI is very much plumbing and can be used in all sorts of conditions, which is great. The Nix 2.0 CLI seems to be doing a bit of both.

edolstra:
> Would it make sense to also introduce the notion of "plumbing" vs "porcelain"? 

Roughly the "utility/scripting commands" (tier 3 in the current doc) are the plumbing.

doc/manual/src/command-ref/cli-guideline.md
garbas: As with everything the balance is the key. I guess you were alerted by the word emojis  :) and seeing them all over your terminal. I'm also annoyed by command line tools that abuse this. But I have also experience command line tools that know how to use emojis.

Like it or not we are all emotional beings. Most of our decisions are based on emotions. And this is why things that we like to use everyday also convey certain emotions. Why wouldn't command line tools do the same.

Just to make sure you understand for what I am advocating I'm going to repeat it in the best ever (:wink:) format to explain things, single bullet points:

- We should use emojis in only few places: (1) this is because some find them annoying (2) with every emoji the rest of the emojis _"lose on power"_.  And when I say few places I really mean only two, maybe maybe tree places.

So key is balance and rather to not use emojis then to use emojis, but there might be few places where usage of emojis makes Nix better. Of course there is a way to disable this altogether and we should even disable it automatically (eg. when tty is not attached to stdout/stderr).

doc/manual/src/command-ref/cli-guideline.md
garbas: Oh yes, this should be `Tier 3`, but I'm going to change the naming here on the request of Eelco.

toonn:
There's a piece of similar documentation which I perceive to be missing which is a transition plan from the old to the new cli. I'm looking forward to a better UX but I hope the plan is not "Nix 2.0 has the old cli and the new experimental stuff is discouraged, Nix 3.0 has the new cli and only the new cli." Will there be an acclimatization period where the new cli is no longer marked as experimental and users are encouraged to attempt transitioning without being forced give up on the familiar cli all of a sudden?

edolstra:
The plan is roughly:

- Nix 2.4 has the new CLI, but marked as experimental.
- Nix 3.0 deprecates the old CLI and makes the new one non-experimental.
- Nix 4.0 removes the old CLI. But this step may never happen :-)

toonn:
Thanks for the quick response. One more question. Apparently `man` looks for `command-subcommand` when you ask for `man nix build` for example. How will this interact with having a man page for both `nix build` and `nix-build`, will `nix help build` be the only way to get a man page (or is it only brief help output?) on the build subcommand? ([Reference for the behavior](https://stackoverflow.com/questions/32747089/how-does-man-find-gits-subcommand-manpages))

edolstra:
Currently the new man pages are called `nix3-*` (e.g. `nix3-build`) to avoid ambiguity.

PR #4305, opened 2020-12-02
Check for rosetta 2 support before installing x86_64-darwin Nix

matthewbauer:
Check for "rosetta 2" on ARM64 macOS first. Otherwise, installation will fail with:

  Bad CPU type in executable

when running nix-store during later installation. This should be a safe check, but will need someone without rosetta 2 to verify this works as expected first (I've now installed it after setting up my new system).

### DIFF
--- a/scripts/install.in
+++ b/scripts/install.in
@@ -45,7 +45,7 @@ case "$(uname -s).$(uname -m)" in
         path=@tarballPath_x86_64-darwin@
         system=x86_64-darwin
         ;;
-    Darwin.arm64)
+    Darwin.arm64|Darwin.aarch64)
         # check for Rosetta 2 support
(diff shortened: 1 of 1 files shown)


Kloenk:
Apple put my m1 mbp in the Post today. Will test before installing.

Kloenk:
Yes, the com.apple.oahd.plist does not exists on a new mbp m1. 

I just did ls on that path, but either ways, LGTM

edolstra:
Thanks, merged!

abathur:
I've opened a PR on my fork at abathur/nix#1 to hopefully bait some discussion about whether or not the installer should install rosetta automatically (and how).

domenkozar:
@matthewbauer would be great if you open a PR against `2.3-maintenance` so that we get this out in Nix 2.3.10

PR #4306, opened 2020-12-03
Add cache.nixos.org to trusted substituters by default

matthewbauer:
When a user is untrusted, Nix will only trust substituters that are in
trusted-substituters. But users don’t always know if they are trusted
or not by the Nix daemon, especially when sharing configs accross
machines. So for example, a user has a nix.conf in $HOME/.config/nix/
that looks like this:

  substituters = https://cache.nixos.org https://my-custom-cache.cachix.org

Nix will override the default substituters, but skip both with:

  warning: ignoring untrusted substituter ‘https://cache.nixos.org’
  warning: ignoring untrusted substituter ‘https://my-custom-cache.cachix.org’

So the user now gets no substituters, and would have to remove the
substituters config setting to reenable cache.nixos.org. To fix this,
just include https://cache.nixos.org by default. This can still be
overridden by the user.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -583,7 +583,7 @@ public:
 
     Setting<Strings> substituters{
         this,
-        nixStore == "/nix/store" ? Strings{"https://cache.nixos.org/"} : Strings(),
+        Strings{"https://cache.nixos.org/"},
         "substituters",
         R"(
           A list of URLs of substituters, separated by whitespace. The default
@@ -592,7 +592,7 @@ public:
         {"binary-caches"}};
 
     Setting<StringSet> trustedSubstituters{
-        this, nixStore == "/nix/store" ? StringSet{"https://cache.nixos.org/"} : {}, "trusted-substituters",
+        this, {"https://cache.nixos.org/"}, "trusted-substituters",
         R"(
           A list of URLs of substituters, separated by whitespace. These are
           not used by default, but can be enabled by users of the Nix daemon
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -592,7 +592,7 @@ public:
         {"binary-caches"}};
 
     Setting<StringSet> trustedSubstituters{
-        this, {}, "trusted-substituters",
+        this, nixStore == "/nix/store" ? StringSet{"https://cache.nixos.org/"} : {}, "trusted-substituters",
         R"(
           A list of URLs of substituters, separated by whitespace. These are
           not used by default, but can be enabled by users of the Nix daemon


Ericson2314:
Because of relocatable source code / deprecation hassjed mirrors, maybe this should be unconditional?

matthewbauer:
That sounds like a good idea. Should both `substituters` and `trusted-substituters` be unconditional?

grahamc:
Does this set a default, or does it force users to trust c.n.o?

grahamc:
To say more, there are Nix users who don't use our cache and don't want to trust our cache.

matthewbauer:
> Does this set a default, or does it force users to trust c.n.o?
> […](#)

So it adds another option they need to set if they don't trust our cache.nixos.org. Currently, they just need to set `substituters = ...` to remove cache.nixos.org. With this, they also need to set `trusted-substituters = ....` to do that. If we're concerned that's too much configuration, we could make trusted-substituters default to the value of substituters if unset I think.

Ericson2314:
Two questions:

1. Does `~/.config/nix/nix.conf` vs CLI flags make a difference?
2. What about if Nix made every `substituter` also a `trusted-substituter`?

domenkozar:
Since 99% of users will use this I support this being opt-out via some flag.

matthewbauer:
> 1. Does `~/.config/nix/nix.conf` vs CLI flags make a difference?

It shouldn't make any difference - CLI flags should always have precedence over any nix.conf though.

> 2. What about if Nix made every `substituter` also a `trusted-substituter`?

I'm not sure how to do this in the way the current configuration works. We'd want the default for one setting to depend on the default of the other setting, which I don't think? we have a way to do right now.

Ericson2314:
> I'm not sure how to do this in the way the current configuration works. We'd want the default for one setting to depend on the default of the other setting, which I don't think? we have a way to do right now.

I would change every access of `trusted-substitutors` to instead check `substitutors ++ trusted-substitutors`.

edolstra:
> What about if Nix made every substituter also a trusted-substituter?

That's already the case, see `daemon.cc`:
```c++
                StringSet trusted = settings.trustedSubstituters;
                for (auto & s : settings.substituters.get())
                    trusted.insert(s);
```

matthewbauer:
Oh this was my bad! This was just a case of a missing slash in my substituters not being picked up (`https://cache.nixos.org/` was in `trusted-substituters` but not `https://cache.nixos.org/`). I had assumed trusted-substituters was completely empty.

See # for a real fix for that problem.

domenkozar:
I wish that uris would be normalized, that would save lots of frustration.

Ericson2314:
@matthewbauer we may still want to do the change of not caring about the store path?

matthewbauer:
> @matthewbauer we may still want to do the change of not caring about the store path?

Yes - https://github.com/NixOS/nix/pull/4321

PR #4309, opened 2020-12-03
Update flake to match removal of utillinuxminimal dependency

shadowrylander:
As per the change in https://github.com/NixOS/nixpkgs/pull/97294

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -90,7 +90,7 @@
             lowdown
             gmock
           ]
-          ++ lib.optionals stdenv.isLinux [libseccomp utillinuxMinimal]
+          ++ lib.optionals stdenv.isLinux [libseccomp]
           ++ lib.optional (stdenv.isLinux || stdenv.isDarwin) libsodium;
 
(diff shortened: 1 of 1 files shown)


edolstra:
I may be missing something here but what does the kernel's util-linux dependency have to do with Nix? Nix depends on util-linux because it needs `unshare` in the tests  (https://github.com/NixOS/nix/pull/2211).

shadowrylander:
Hmm... I think I'm missing something here as well... With `utillinuxminimal` in the flake I'm getting an error stating that said variable is undefined.

edolstra:
IIRC `utillinux` was renamed to `util-linux` on nixpkgs master. However, Nix is pinned to 20.09 so it shouldn't be affected by that rename.

shadowrylander:
Ah; my fault, then. I pinned its `nixpkgs` input to the master branch. Whoops. Sorry for wasting your time; I'll go ahead and close this! 😅

PR #4310, opened 2020-12-03
Add x86_64-darwin and aarch64 to "extra-platforms" automatically when Rosetta2 is detected

matthewbauer:
First:

**Add extraPlatforms for Rosetta 2 macOS**

macOS systems with ARM64 can utilize a translation layer at /Library/Apple/usr/libexec/oah to run x86_64 binaries. This change makes Nix recognize that and it to "extra-platforms". Note that there are two cases here since Nix could be built for either x86_64 or aarch64. In either case, we can switch to the other architecture. Unfortunately there is not a good way to prevent aarch64 binaries from being run in x86_64 contexts or vice versa - programs can always execute programs for the other architecture.

~Question: Can someone confirm that /Library/Apple/usr/libexec/oah **does not** exist when Rosetta 2 is not installed? There could be some stubs there even if Rosetta 2 is not setup yet.~

Second:

**Use posix_spawn_setbinpref_np to advise which architecture to run**

When running universal binaries like /bin/bash, Darwin XNU will choose which architecture of the binary to use based on "binary preferences". This change sets that to the current platform for aarch64 and x86_64 builds. In addition it now uses posix_spawn instead of the usual execve. Note, that this does not prevent the other architecture from being run, just advises which to use.

/cc @abathur @thefloweringash @edolstra

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -50,6 +50,10 @@
 #define pivot_root(new_root, put_old) (syscall(SYS_pivot_root, new_root, put_old))
 #endif
 
+#if __APPLE__
+#include <spawn.h>
+#endif
+
 #include <pwd.h>
 #include <grp.h>
 
@@ -2844,7 +2848,27 @@ void DerivationGoal::runChild()
             }
         }
 
+#if __APPLE__
+        posix_spawnattr_t attrp;
+
+        if (posix_spawnattr_init(&attrp))
+            throw SysError("failed to initialize builder");
+
+        if (posix_spawnattr_setflags(&attrp, POSIX_SPAWN_SETEXEC))
+            throw SysError("failed to initialize builder");
+
+        if (drv->platform == "aarch64-darwin") {
+            cpu_type_t cpu = CPU_TYPE_ARM64;
+            posix_spawnattr_setbinpref_np(&attrp, 1, &cpu, NULL);
+        } else if (drv->platform == "x86_64-darwin") {
+            cpu_type_t cpu = CPU_TYPE_X86_64;
+            posix_spawnattr_setbinpref_np(&attrp, 1, &cpu, NULL);
+        }
+
+        posix_spawn(NULL, builder, NULL, &attrp, stringsToCharPtrs(args).data(), stringsToCharPtrs(envStrs).data());
+#else
         execve(builder, stringsToCharPtrs(args).data(), stringsToCharPtrs(envStrs).data());
+#endif
 
         throw SysError("executing '%1%'", drv->builder);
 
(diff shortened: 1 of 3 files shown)


Ericson2314 (approved):

abathur:
Not sure which bits you're hoping I'll have feedback on; @LnL7 is almost certainly a better mention if he has bandwidth!

It's not clear to me exactly what you mean or under what circumstances Rosetta 2 isn't installed or set up (perhaps it would be obvious if I had access to an ARM mac?), but I do still have an intel Big Sur install, so I did run `find / -name oah` to avoid being useless:
- Catalina: no results
- Intel Big Sur: (I actually see 6 different paths, but I think they're all either the libexec or db version of the path)
    - /System/Library/ProxySigningStubsLibrary/Apple/usr/libexec/oah (file says: directory)
        - This has files: debugserver, oahd, oahd-helper, oahd-root-helper, runtime, runtime_t8027, translate_tool
    - /System/Library/Templates/Data/private/var/db/oah (file says: directory)
        - ls says no perm, I assume it needs full disk

thefloweringash:
I see an unofficial [install script][] for Rosetta 2 that uses the presence of ` /Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist` to detect if it has been installed. On my arm64 mac with Rosetta installed I can confirm this file exists.

[install script]: https://github.com/rtrouton/rtrouton_scripts/blob/45375866b681eac03703ce1edc9a68ac68f13e8f/rtrouton_scripts/install_rosetta_on_apple_silicon/install_rosetta_on_apple_silicon.sh#L33

abathur:
`sudo find / -name com.apple.oahd.plist` came up empty on both catalina and intel big sur

abathur:
Also, /usr/sbin/softwareupdate does list the --install-rosetta option ("install Rosetta 2"), and running it on Intel yields `Installing Rosetta 2 on this system is not supported.`

Kloenk:
> I see an unofficial [install script][] for Rosetta 2 that uses the presence of ` /Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist` to detect if it has been installed. On my arm64 mac with Rosetta installed I can confirm this file exists.
> 
> 
> 
> [install script]: https://github.com/rtrouton/rtrouton_scripts/blob/45375866b681eac03703ce1edc9a68ac68f13e8f/rtrouton_scripts/install_rosetta_on_apple_silicon/install_rosetta_on_apple_silicon.sh#L33

Hopefully around Tuesday I can look if that path exists, before installing Rosetta 2 if that helps.

Kloenk:
What's the state of this one? The path was correct. (Only forgot to confirm it here though)

PR #4315, opened 2020-12-04
Properly store the outputs of CA derivations

regnat:
For each “known” derivation output, store:
- its output id
- its resolved derivation
- its output path

(morally what we want is two families of mappings: One mapping
unresolved derivations to their resolved version and one mapping
resolved derivations to their output path, but it's both simpler and
more efficient to combine them)

This comes with a set of needed changes:

- New `drv-output-info` module declaring the types needed for describing
  these mappings
- New `Store::registerDrvOutput` method registering all the needed informations
  about a derivation output (also replaces `LocalStore::linkDeriverToPath`)
- new `Store::queryDrvOutputInfo` method to retrieve the informations for a
  derivations

This introcudes some redundancy on the remote-store side between
`wopQueryDerivationOutputMap` and `wopQueryDrvOutputInfo`.
However we might need to keep both (regardless of backwards compat)
because we sometimes need to get some infos for all the outputs of a
derivation (where `wopQueryDerivationOutputMap` is handy), but all the
stores can't implement it − because listing all the outputs of a
derivation isn't really possible for binary caches where the server
doesn't allow to list a directory.

For the local store, this uses a different db tables that will only be added to the schema if the `ca-derivations` experimental flag is on, meaning that people not using it won't be affected by the schema change

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -443,6 +443,25 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
     })->path;
 }
 
+std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOutput & id)
+{
+    auto outputInfoFilePath =
+        "/drvOutputs/" + std::string(id.drvPath.hashPart()) + "!" + id.outputName + ".doi";
+    auto rawOutputInfo = getFile(outputInfoFilePath);
+
+    if (rawOutputInfo) {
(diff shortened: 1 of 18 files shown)


regnat:
I've reworked this a little to

1. Remove the `resolvedDrv` field as it might not give much
2. Remove some unused code
3. Rename the added table into `Realisations` as `OutputMappings` was a rather ugly and imprecise term

regnat:
Rebased on top of latest master to fix a conflict

regnat:
Closing in favor of #4330

PR #4317, opened 2020-12-04
rust: base32; decode_len, add case for 0

Horki:
@edolstra Hi, added a small update, 
- same check as in `encode_len`


zimbatm:
Doesn't `decoded_len` already return 0 for `0 * 5 / 8`? It's different for `encoded_len` because of the padding.

Horki:
@zimbatm yes you are right

PR #4319, opened 2020-12-05
libstore/openStore: fix stores with IPv6 addresses

Ma27:
In `nixStable` (2.3.7 to be precise) it's possible to connect to stores
using an IPv6 address:

    nix ping-store --store ssh://root@2001:db8::1

This is also useful for `nixops(1)` where you could specify an IPv6
address in `deployment.targetHost`.

However, this behavior is broken on `nixUnstable` and fails with the
following error:

    $ nix store ping --store ssh://root@2001:db8::1
    don't know how to open Nix store 'ssh://root@2001:db8::1'

This happened because `openStore` from `libstore` uses the `parseURL`
function from `libfetchers` which expects a valid URL as defined in
RFC2732. However, this is unsupported by `ssh(1)`:

    $ nix store ping --store 'ssh://root@[2001:db8::1]'
    cannot connect to 'root@[2001:db8::1]'

This patch now allows both ways of specifying a store (`root@2001:db8::1`) and
also `root@[2001:db8::1]` since the latter one is useful to pass query
parameters to the remote store.

In order to achieve this, the following changes were made:

* The URL regex from `url-parts.hh` now allows an IPv6 address in the
  form `2001:db8::1` and also `[2001:db8::1]`.

* In `libstore`, a new function named `extractConnStr` ensures that a
  proper URL is passed to e.g. `ssh(1)`:

  * If a URL looks like either `[2001:db8::1]` or `root@[2001:db8::1]`,
    the brackets will be removed using a regex. No additional validation
    is done here as only strings parsed by `parseURL` are expected.

  * In any other case, the string will be left untouched.

Unresolved questions:

* I'm not really sure whether we want to allow both variants of IPv6
  addresses in the URL parser. However it should be noted that both seem
  to be possible according to RFC2732:

  > This document incudes an update to the generic syntax for Uniform
  > Resource Identifiers defined in RFC 2396 [URL].  It defines a syntax
  > for IPv6 addresses and allows the use of "[" and "]" within a URI
  > explicitly for this reserved purpose.

* Currently, it's not supported to specify a port number behind the
  hostname, however it seems as this is not really supported by the URL
  parser. Hence, this is probably out of scope here.

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -10,6 +10,8 @@
 #include "archive.hh"
 #include "callback.hh"
 
+#include <regex>
+
 namespace nix {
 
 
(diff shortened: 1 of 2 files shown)


Ericson2314 (approved):

Ma27:
cc @edolstra would you mind taking a look? :)

src/libstore/store-api.cc
```diff
@@ -1099,7 +1126,7 @@ ref<Store> openStore(const std::string & uri_,
         auto parsedUri = parseURL(uri_);
         params.insert(parsedUri.query.begin(), parsedUri.query.end());
 
-        auto baseURI = parsedUri.authority.value_or("") + parsedUri.path;
+        auto baseURI = extractConnStr(parsedUri.authority.value_or("") + parsedUri.path);
```
edolstra, line 1129: Shouldn't `extractConnStr` be applied in `SSHStore` / `LegacySSHStore`? Doing it here seems wrong (e.g. a binary cache URL like `http://[2001:db8::1]/` would be mangled to `http://2001:db8::1/`, which doesn't work with curl).

Ma27: Fair point, will take care of this.

Ma27:
@edolstra fixed :)

blaggacao:
@Ma27 what do you think of #4490 ?
It seems that the `%ifidentifier` is not properly recognized. Would that be fixable, in principle?

PR #4320, opened 2020-12-06
Support statically built Nix on Darwin

matthewbauer:
This adds support for static Darwin Nix via https://github.com/NixOS/nixpkgs/pull/106115.

- Need to skip the `-Ur` flag for darwin cctools
- Update nixpkgs to my branch (should probably wait until merged into Nixpkgs & maybe released)
- Enable x86_64-darwin builds.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -476,7 +476,7 @@
         # Binary package for various platforms.
         build = forAllSystems (system: self.packages.${system}.nix);
 
-        buildStatic = lib.genAttrs linux64BitSystems (system: self.packages.${system}.nix-static);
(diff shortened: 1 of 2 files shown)


edolstra:
Yes, we should stick to the Nixpkgs 20.09 branch for now. Otherwise looks good to me.

domenkozar:
@matthewbauer care to sync with nixpkgs changes?

domenkozar:
@matthewbauer ping :)

matthewbauer:
Static compilation is broken by https://github.com/NixOS/nixpkgs/pull/107238 - will need to address it in nixpkgs.

PR #4325, opened 2020-12-08
Hide the sqlite statements declarations for the local store

regnat:
Hide the sqlite statements declarations for the local store
    
These have no need to be in the public interface and it causes spurious rebuilds each time one wants to add or remove a new statement.

### DIFF
--- a/src/libstore/local-store.hh
+++ b/src/libstore/local-store.hh
@@ -55,19 +55,8 @@ private:
         /* The SQLite database object. */
         SQLite db;
 
-        /* Some precompiled SQLite statements. */
-        SQLiteStmt stmtRegisterValidPath;
-        SQLiteStmt stmtUpdatePathInfo;
-        SQLiteStmt stmtAddReference;
-        SQLiteStmt stmtQueryPathInfo;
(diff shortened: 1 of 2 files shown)


src/libstore/local-store.hh
```diff
@@ -55,19 +55,8 @@ private:
...
-        SQLiteStmt stmtQueryReferences;
-        SQLiteStmt stmtQueryReferrers;
-        SQLiteStmt stmtInvalidatePath;
-        SQLiteStmt stmtAddDerivationOutput;
-        SQLiteStmt stmtQueryValidDerivers;
-        SQLiteStmt stmtQueryDerivationOutputs;
-        SQLiteStmt stmtQueryPathFromHashPart;
-        SQLiteStmt stmtQueryValidPaths;
+        struct Stmts;
+        std::shared_ptr<Stmts> stmts;
```
edolstra, line 59: This can probably be a `std::unique_ptr`?

edolstra (approved):

src/libstore/local-store.hh
regnat: Indeed, thanks.

Fixed in c0f21f0

PR #4330, opened 2020-12-08
Properly store the outputs of CA derivations − take 2

### LINKED ISSUE #4138: The drvOutputId->OutputPath link is lost for CA derivations when the drv file is deleted

regnat:
Reopening of #4315 with a branch from the nixos/nix repo just to allow dpulls to work properly

Fix #4138

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -449,7 +449,8 @@ std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOut
     auto rawOutputInfo = getFile(outputInfoFilePath);
 
     if (rawOutputInfo) {
-        return { Realisation::parse(*rawOutputInfo, outputInfoFilePath) };
+        return {Realisation::fromJSON(
+            nlohmann::json::parse(*rawOutputInfo), outputInfoFilePath)};
     } else {
         return std::nullopt;
     }
@@ -457,7 +458,7 @@ std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOut
 
 void BinaryCacheStore::registerDrvOutput(const Realisation& info) {
     auto filePath = realisationsPrefix + "/" + info.id.to_string() + ".doi";
-    upsertFile(filePath, info.to_string(), "text/x-nix-derivertopath");
+    upsertFile(filePath, info.toJSON(), "application/json");
 }
 
 ref<FSAccessor> BinaryCacheStore::getFSAccessor()
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -443,6 +443,23 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
     })->path;
 }
 
+std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOutput & id)
+{
+    auto outputInfoFilePath = realisationsPrefix + "/" + id.to_string() + ".doi";
+    auto rawOutputInfo = getFile(outputInfoFilePath);
+
+    if (rawOutputInfo) {
+        return { Realisation::parse(*rawOutputInfo, outputInfoFilePath) };
+    } else {
+        return std::nullopt;
+    }
+}
+
+void BinaryCacheStore::registerDrvOutput(const Realisation& info) {
+    auto filePath = realisationsPrefix + "/" + info.id.to_string() + ".doi";
+    upsertFile(filePath, info.to_string(), "text/x-nix-derivertopath");
+}
+
 ref<FSAccessor> BinaryCacheStore::getFSAccessor()
 {
     return make_ref<RemoteFSAccessor>(ref<Store>(shared_from_this()), localNarCache);
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -493,8 +493,9 @@ void DerivationGoal::inputsRealised()
     if (useDerivation) {
         auto & fullDrv = *dynamic_cast<Derivation *>(drv.get());
 
-        if ((!fullDrv.inputDrvs.empty() && derivationIsCA(fullDrv.type()))
-            || fullDrv.type() == DerivationType::DeferredInputAddressed) {
+        if (settings.isExperimentalFeatureEnabled("ca-derivations") &&
+            ((!fullDrv.inputDrvs.empty() && derivationIsCA(fullDrv.type()))
+            || fullDrv.type() == DerivationType::DeferredInputAddressed)) {
             /* We are be able to resolve this derivation based on the
                now-known results of dependencies. If so, we become a stub goal
                aliasing that resolved derivation goal */
@@ -3405,12 +3406,11 @@ void DerivationGoal::registerOutputs()
         drvPathResolved = writeDerivation(worker.store, drv2);
     }
 
-    if (useDerivation || isCaFloating)
-        for (auto & [outputName, newInfo] : infos)
-            worker.store.registerDrvOutput(
-                DrvOutputId{drvPathResolved, outputName},
-                DrvOutputInfo{.outPath = newInfo.path,
-                              .resolvedDrv = drvPathResolved});
+    if (settings.isExperimentalFeatureEnabled("ca-derivations"))
+        for (auto& [outputName, newInfo] : infos)
+            worker.store.registerDrvOutput(Realisation{
+                .id = DrvOutput{drvPathResolved, outputName},
+                .outPath = newInfo.path});
 }
 
 
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -2094,6 +2094,20 @@ struct RestrictedStore : public LocalFSStore, public virtual RestrictedStoreConf
         /* Nothing to be done; 'path' must already be valid. */
     }
 
+    void registerDrvOutput(const Realisation & info) override
+    {
+        if (!goal.isAllowed(info.id.drvPath))
+            throw InvalidPath("cannot register unknown drv output '%s' in recursive Nix", printStorePath(info.id.drvPath));
+        next->registerDrvOutput(info);
+    }
+
+    std::optional<const Realisation> queryRealisation(const DrvOutput & id) override
+    {
+        if (!goal.isAllowed(id.drvPath))
+            throw InvalidPath("cannot query the output info for unknown derivation '%s' in recursive Nix", printStorePath(id.drvPath));
+        return next->queryRealisation(id);
+    }
+
     void buildPaths(const std::vector<StorePathWithOutputs> & paths, BuildMode buildMode) override
     {
         if (buildMode != bmNormal) throw Error("unsupported build mode");
@@ -3393,7 +3407,10 @@ void DerivationGoal::registerOutputs()
 
     if (useDerivation || isCaFloating)
         for (auto & [outputName, newInfo] : infos)
-            worker.store.linkDeriverToPath(drvPathResolved, outputName, newInfo.path);
+            worker.store.registerDrvOutput(
+                DrvOutputId{drvPathResolved, outputName},
+                DrvOutputInfo{.outPath = newInfo.path,
+                              .resolvedDrv = drvPathResolved});
 }
 
 
--- a/src/libstore/ca-specific-schema.sql
+++ b/src/libstore/ca-specific-schema.sql
@@ -1,0 +1,11 @@
+-- Extension of the sql schema for content-addressed derivations.
+-- Won't be loaded unless the experimental feature `ca-derivations`
+-- is enabled
+
+create table if not exists Realisations (
+    drvPath text not null,
+    outputName text not null, -- symbolic output id, usually "out"
+    outputPath integer not null,
+    primary key (drvPath, outputName),
+    foreign key (outputPath) references ValidPaths(id) on delete cascade
+);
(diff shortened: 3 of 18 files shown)


Ericson2314:
Well this is a boring thing to bring up, but what do all you say we use JSON for the new binary cache files? It would be nice to avoid these bespoke parsers a bit.

regnat:
> Well this is a boring thing to bring up, but what do all you say we use JSON for the new binary cache files? It would be nice to avoid these bespoke parsers a bit.

Good point… I kept the same syntax as the narinfo files for consistency, but I don't care about changing that − might even be nice to have a more structured format to store lists of things or the like.

I'd rather not have it block this PR though ;)

edolstra:
Yes, I'm in favor of using JSON for new files (in fact we're using it for NAR listings in the binary cache).

src/libstore/binary-cache-store.cc
```diff
@@ -443,6 +443,25 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
...
+    if (rawOutputInfo) {
+        return { Realisation::parse(*rawOutputInfo, outputInfoFilePath) };
+    } else {
+        return std::nullopt;
+    }
+}
+
+void BinaryCacheStore::registerDrvOutput(const Realisation& info) {
+    auto filePath = "/drvOutputs/" + std::string(info.id.drvPath.hashPart()) +
+        "!" + info.id.outputName + ".doi";
```
Ericson2314, line 461: what do you all think about nesting further like this?
```suggestion
    auto filePath = "/drvOutputs/" + std::string(info.id.drvPath.hashPart()) +
        "/" + info.id.outputName + ".doi";
```
rather than using `!`

regnat: I used `!` because that's the already standard delimiter for derivation outputs, so it makes sense to use the same on binary caches − I've actually pushed 391f753 to make this use `id.to_string()` directly which makes this more natural.

That being said, I don't have a very strong opinion, so a `/` could be good to

regnat:
> Yes, I'm in favor of using JSON for new types of files (in fact we're using it for NAR listings in the binary cache).

Ok, I've opened 4332 to track this

src/libstore/binary-cache-store.cc
regnat: We could even use a `/` for `id.to_string()`, but I like the fact that `!` gives a clear message of *this isn't a store path, don't try to use it as such*

src/libstore/build/derivation-goal.cc
```diff
@@ -3391,9 +3406,12 @@ void DerivationGoal::registerOutputs()
         drvPathResolved = writeDerivation(worker.store, drv2);
     }
 
-    if (useDerivation || isCaFloating)
-        for (auto & [outputName, newInfo] : infos)
-            worker.store.linkDeriverToPath(drvPathResolved, outputName, newInfo.path);
+    if (settings.isExperimentalFeatureEnabled("ca-derivations") &&
+        (useDerivation || isCaFloating))
```
Ericson2314, line 3410: Do we want to `registerDrvOutput` for non-CA derivations?

If so, we can do
```suggestion
    if (settings.isExperimentalFeatureEnabled("ca-derivations")
```
and make it unconditional if we like, since there no longer is a derivation foreign key.

If not, we could do
```suggestion
    if (isCaFloating)  {
        assert(settings.isExperimentalFeatureEnabled("ca-derivations"));
```

regnat: Oh you're right, we can get rid of this

Ericson2314:
@regnat does this make any changes w.r.t. storing the unresolved derivations? I think it doesn't but I am not yet sure. Of course, I'll be very happy of it doesn't :).

regnat:
> @regnat does this make any changes w.r.t. storing the unresolved derivations? I think it doesn't but I am not yet sure. Of course, I'll be very happy of it doesn't :).

Nah that one doesn't

Ericson2314:
> Nah that one doesn't

Great!

> This introduces some redundancy on the remote-store side between
`wopQueryDerivationOutputMap` and `wopQueryDrvOutputInfo`.
However we might need to keep both (regardless of backwards compat)
because we sometimes need to get some infos for all the outputs of a
derivation (where `wopQueryDerivationOutputMap` is handy), but all the
stores can't implement it − because listing all the outputs of a
derivation isn't really possible for binary caches where the server
doesn't allow to list a directory.

What if we stored one map per derivation with all the outputs in the binary caches?

(Also, I think that was renamed `wopQueryRealisation`?)

regnat:
> > This introduces some redundancy on the remote-store side between
> > `wopQueryDerivationOutputMap` and `wopQueryDrvOutputInfo`.
> > However we might need to keep both (regardless of backwards compat)
> > because we sometimes need to get some infos for all the outputs of a
> > derivation (where `wopQueryDerivationOutputMap` is handy), but all the
> > stores can't implement it − because listing all the outputs of a
> > derivation isn't really possible for binary caches where the server
> > doesn't allow to list a directory.
> 
> What if we stored one map per derivation with all the outputs in the binary caches?

That was my initial idea, but it's possible to only push a subset of the outputs to the binary cache, meaning that if you add some more afterwards you'd have to update the entry, and that would break the − really nice property imho − that everything in the binary cache is immutable (we could also use a storing scheme like `drvHash/outputName` like you suggest above and list the directories, but not every cache permits listing directories − and I think not requiring that is also a very cool property that would be sad to loose).

> 
> (Also, I think that was renamed `wopQueryRealisation`?)

Someone should make a variant of `sed` that also updates the commit messages to make sure that they accurately reflect your changes :unamused:

Ericson2314:
@regnat well what if you had to push the mapping for all outputs whenever you pushed any of them? Fundamentally, the mappings all trace back to builds, and it's impossible to just build some of the derivations, so I think this could actually work.

I think unlike the local store, we might very much want to have a relation not function for the binary cache. This would allow pushing conflicting mappings when we have some outputs and other outputs from another, in a particularly honest fashion. I guess we could just do `<drv-hash>/<mapping-file-checksum>` or something to avoid collisions or the pain of dealing with incremented numbers.

regnat:
> @regnat well what if you had to push the mapping for all outputs whenever you pushed any of them? Fundamentally, the mappings all trace back to builds, and it's impossible to just build some of the derivations, so I think this could actually work.

Up to a certain point, yes, except that you can `nix-build foo.out && nix-collect-garbage && nix copy --to binary-cache foo.out` (or just `nix-build foo.out && nix copy --to binary-cache foo.out` where `foo.out` is already in a different binary cache so you don't have to build `foo.dev`).

> I think unlike the local store, we might very much want to have a relation not function for the binary cache. This would allow pushing conflicting mappings when we have some outputs and other outputs from another, in a particularly honest fashion. I guess we could just do `<drv-hash>/<mapping-file-checksum>` or something to avoid collisions or the pain of dealing with incremented numbers.

I'm not sure I get exactly how that would work… do you mind expanding a bit further how you consider things? (If possible in a new issue as this PR doesn't do much for binary caches except for a basic and probably not definitive implementation of `queryRealisation`)

Ericson2314:
@regnat Sure I can expand elsewhere, maybe let's just merge this.

regnat:
> @regnat Sure I can expand elsewhere, maybe let's just merge this, as long as we are happy of all manner of breaking changes elsewhere. `queryRealisation` is guarded by the experimental issue so I think we're good.

I'm definitely happy with any breaking change that migh (and will) happen!

I've rebased to fix the mention of `wopQueryDrvOutputInfo`  commit message.

Ericson2314:
I made the multiple outputs issue: #4344. [Side note, I wonder if I could get enough perms to at label the issue and whatnot?]

src/libstore/binary-cache-store.cc
```diff
@@ -443,6 +443,23 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
     })->path;
 }
 
+std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOutput & id)
+{
+    auto outputInfoFilePath = realisationsPrefix + "/" + id.to_string() + ".doi";
```
edolstra, line 448: What does `doi` mean?

BTW this URL format means that there can be only one mapping per realisation, which is probably fine for a binary cache, but it does mean you can't take the union of two binary caches (e.g. a mirror of multiple upstream binary caches) - or at least not of their `/realisations` directory.

src/libstore/ca-specific-schema.sql
```diff
@@ -0,0 +1,13 @@
+-- Extension of the sql schema for content-addressed derivations.
+-- Won't be loaded unless the experimental feature `ca-derivations`
+-- is enabled
+
+create table if not exists Realisations (
```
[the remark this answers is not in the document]
regnat: There's indeed a terminology mismatch and we need to settle on one thing.

I really like `Realisation` tbh because these are precisely witnesses that a derivation has been realised (for the definition you give above). But it's indeed a quite overloaded term so if you don't like it we can find something else.

I tend to use `DerivationOutput` only for the “symbolic” `foo.drv!out` because it's also a very overloaded term and it's easy to get lost if we abuse of it.

(also, they aren't really `Derivers` (though they could be used as such) because the intent is to associate derivation outputs (`foo.drv!out`) to their output paths and not the other way around)

TL;DR, I'm fine changing the naming, but for my own sanity (and other's too) I think we need to find clear and distinct names for

1. `foo.drv!out`
2. The relation `(foo.drv!out, /nix/store/123-foo-out)`

src/libstore/ca-specific-schema.sql
```diff
@@ -0,0 +1,13 @@
...
+    id integer primary key autoincrement not null,
+    drvPath text not null,
+    outputName text not null, -- symbolic output id, usually "out"
```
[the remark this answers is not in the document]
Ericson2314: Oh thanks I should have caught this. Yes let's start as stringent as possible and only relax later.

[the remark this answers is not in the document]
regnat: Indeed. I'll drop the `id` for now

src/libstore/ca-specific-schema.sql
```diff
@@ -0,0 +1,13 @@
...
+    outputPath integer not null,
+    foreign key (outputPath) references ValidPaths(id) on delete cascade
+);
+
+create index if not exists IndexRealisations on Realisations(outputPath);
```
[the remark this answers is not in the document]
regnat: Of course. Actually the index on `outputPath` is quite useless

src/libstore/local-store.hh
```diff
@@ -21,7 +21,7 @@ namespace nix {
    0.7.  Version 2 was Nix 0.8 and 0.9.  Version 3 is Nix 0.10.
    Version 4 is Nix 0.11.  Version 5 is Nix 0.12-0.16.  Version 6 is
    Nix 1.0.  Version 7 is Nix 1.3. Version 10 is 2.0. */
-const int nixSchemaVersion = 10;
+const int nixSchemaVersion = 11;
```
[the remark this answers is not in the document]
regnat: Nope indeed

regnat (commented):
@edolstra Just gave a pass after your comments, I'll let you have a second look

Ericson2314:
> as it's the only place in which we can't break backwards compat for free

How come? Can we change all implementations to assert the experimental feature before doing anything else?

regnat:
> > as it's [`queryRealization` in the daemon protocol is] the only place in which we can't break backwards compat for free
> 
> How come? Can we change all implementations to assert the experimental feature before doing anything else?

Oh yes, you're right. Well, now it's implemented and the code is actually cleaner that way, so might as well keep it

Ericson2314:
Yeah I'm not against this new version (though I haven't closely looked), just think if we did https://github.com/NixOS/nix/issues/4344 we definitely would change it again.

PR #4333, opened 2020-12-08
Store the realisations as JSON in the binary cache

### LINKED ISSUE #4332: Use json for storing the realisations in the binary cache
Follow-up of https://github.com/NixOS/nix/pull/4330#issuecomment-740696586.

The realisations in the binary cache are currently stored in the same format as narinfo files, but we could make them json instead which would lead to a simpler, more robust and more standard parsing.

regnat:
Fix #4332

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -449,7 +449,8 @@ std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOut
     auto rawOutputInfo = getFile(outputInfoFilePath);
 
     if (rawOutputInfo) {
-        return { Realisation::parse(*rawOutputInfo, outputInfoFilePath) };
+        return {Realisation::fromJSON(
+            nlohmann::json::parse(*rawOutputInfo), outputInfoFilePath)};
     } else {
         return std::nullopt;
     }
@@ -457,7 +458,7 @@ std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOut
 
 void BinaryCacheStore::registerDrvOutput(const Realisation& info) {
     auto filePath = realisationsPrefix + "/" + info.id.to_string() + ".doi";
-    upsertFile(filePath, info.to_string(), "text/x-nix-derivertopath");
+    upsertFile(filePath, info.toJSON(), "application/json");
 }
 
 ref<FSAccessor> BinaryCacheStore::getFSAccessor()
--- a/src/libstore/realisation.cc
+++ b/src/libstore/realisation.cc
@@ -1,5 +1,6 @@
 #include "realisation.hh"
 #include "store-api.hh"
+#include <nlohmann/json.hpp>
 
 namespace nix {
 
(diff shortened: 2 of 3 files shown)


Ericson2314 (approved):

Ericson2314:
@regnat I think the depends and the target branch clash, which is why `dpulls` is complaining?

regnat:
> @regnat I think the depends and the target branch clash, which is why `dpulls` is complaining?

Nah, unless I misunderstood something dpulls only complains because the dependency isn't merged. Should be green once we merge it

Ericson2314:
@regnat but the PR is into `ca/properly-store-outputs`, not `master`. It shouldn't matter whether or not `ca/properly-store-outputs` is merged into anything else because that doesn't effect whether things can be merged into it, and this pull request doesn't depend on anything?

regnat:
> @regnat but the PR is into `ca/properly-store-outputs`, not `master`. It shouldn't matter whether or not `ca/properly-store-outputs` is merged into anything else because that doesn't effect whether things can be merged into it, and this pull request doesn't depend on anything?

Oh yes, I see what you mean. Well the thing is that *something* (either dpulls or github itself) automatically changes the target branch to `master` if the current one gets deleted.

edolstra (approved):

Ericson2314:
Oh wow, how odd.

Well, I feel like we might as well actually merge this into the other one now that it's finished? (and it doesn't add more complexity.)

src/libstore/binary-cache-store.cc
```diff
@@ -449,15 +449,16 @@ std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOut
...
+            nlohmann::json::parse(*rawOutputInfo), outputInfoFilePath)};
     } else {
         return std::nullopt;
     }
 }
 
 void BinaryCacheStore::registerDrvOutput(const Realisation& info) {
     auto filePath = realisationsPrefix + "/" + info.id.to_string() + ".doi";
-    upsertFile(filePath, info.to_string(), "text/x-nix-derivertopath");
+    upsertFile(filePath, info.toJSON(), "application/json");
```
zimbatm, line 461: is the out of toJSON reproducible?

regnat: Good question… I guess the answer is “no, but we don't really care here”:

- No because:
    1. It probably won't be reproducible across Nix versions because we might want to add new fields or extend existing fields (like we already occasionally do for the narinfo files)
    2. `nlohmann::json` (the library we're using behind the scenes to convert to json) doesn't explicitely guarantees that its output is reproducible between versions − although it seems to hold in practice. I've [asked about it](https://github.com/nlohmann/json/discussions/2526) though because it could be needed in other places
- But we don't care because these only have to be semantically equivalent (we never try to content-address them or whatever, and we want to give us the possibility of extending their format in the future anyways)

zimbatm: If we start using more JSON outputs, it would be nice if the fields were alphabetically sorted by default.

regnat: If it's just that, it's already the case − and even if there's no guaranty that the default won't change, that bit of the library is pluggable − so I'm not too worried.

The bits that might change is stuff like the representation of numbers (`1000` vs `1E3` for example). That bit is hardcoded and the author didn't give any formal guaranty that it wouldn't change in the future.

PR #4340, opened 2020-12-09
Store the realisations of unresolved derivations

regnat:
When building a (non fully-input-addressed) derivation, we first resolve it to a `BasicDerivation` that depends directly on the output paths of its dependencies (rather than on the derivation outputs) − this is what enables early cutoff.

Currently we register the realisation of the resolved derivation, so that subsequent calls to `nix-build` can directly reuse this without rebuilding the derivation.  However, we don't register the unresolved derivation, meaning that we have to re-do the resolution step each time.
This in turn means that we must either keep all the build inputs or keep the knowledge of their output path (without the content of the path itself).
The former is very costly in disk-space (it amounts to making every build-input a runtime dependency), and the latter comes with a big set of challenges in presence of non-determinism that makes this a too heavy-handed solution for the time being (though we might switch to it eventually if we can find a way to solve its issues as it's a rather tempting design).

To avoid these issues, we choose a third way which is to also register the realisation of the unresolved derivation − so we can completely sidestep the resolving phase, which makes these behave as classical input-addressed derivations.

Depends on #4330

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -124,6 +124,17 @@ DerivationGoal::DerivationGoal(const StorePath & drvPath, const BasicDerivation 
     , buildMode(buildMode)
 {
     this->drv = std::make_unique<BasicDerivation>(BasicDerivation(drv));
+
+    auto outputHashes = staticOutputHashes(worker.store, drv);
+    for (auto &[outputName, outputHash] : outputHashes)
+      initialOutputs.insert({
+            outputName,
+            InitialOutput{
+                .wanted = true, // Will be refined later
+                .outputHash = outputHash
+            }
+          });
+
     state = &DerivationGoal::haveDerivation;
     name = fmt(
         "building of '%s' from in-memory derivation",
@@ -258,8 +269,20 @@ void DerivationGoal::loadDerivation()
 
     assert(worker.store.isValidPath(drvPath));
 
+    auto fullDrv = new Derivation(worker.store.derivationFromPath(drvPath));
+
+    auto outputHashes = staticOutputHashes(worker.store, *fullDrv);
+    for (auto &[outputName, outputHash] : outputHashes)
+      initialOutputs.insert({
+            outputName,
+            InitialOutput{
+                .wanted = true, // Will be refined later
+                .outputHash = outputHash
+            }
+          });
+
     /* Get the derivation. */
-    drv = std::unique_ptr<BasicDerivation>(new Derivation(worker.store.derivationFromPath(drvPath)));
+    drv = std::unique_ptr<BasicDerivation>(fullDrv);
 
     haveDerivation();
 }
@@ -1022,14 +1045,6 @@ void DerivationGoal::buildDone()
 void DerivationGoal::resolvedFinished() {
     assert(resolvedDrv);
 
-    // If the derivation was originally a full `Derivation` (and not just
-    // a `BasicDerivation`, we must retrieve it because the `staticOutputHashes`
-    // will be wrong otherwise
-    Derivation fullDrv = *drv;
-    if (auto upcasted = dynamic_cast<Derivation *>(drv.get()))
-        fullDrv = *upcasted;
-
-    auto originalHashes = staticOutputHashes(worker.store, fullDrv);
     auto resolvedHashes = staticOutputHashes(worker.store, *resolvedDrv);
 
     // `wantedOutputs` might be empty, which means “all the outputs”
@@ -1038,7 +1053,7 @@ void DerivationGoal::resolvedFinished() {
         realWantedOutputs = resolvedDrv->outputNames();
 
     for (auto & wantedOutput : realWantedOutputs) {
-        assert(originalHashes.count(wantedOutput) != 0);
+        assert(initialOutputs.count(wantedOutput) != 0);
         assert(resolvedHashes.count(wantedOutput) != 0);
         auto realisation = worker.store.queryRealisation(
                 DrvOutput{resolvedHashes.at(wantedOutput), wantedOutput}
@@ -1047,7 +1062,7 @@ void DerivationGoal::resolvedFinished() {
         // realisation won't be there
         if (realisation) {
             auto newRealisation = *realisation;
-            newRealisation.id = DrvOutput{originalHashes.at(wantedOutput), wantedOutput};
+            newRealisation.id = DrvOutput{initialOutputs.at(wantedOutput).outputHash, wantedOutput};
             worker.store.registerDrvOutput(newRealisation);
         } else {
             // If we don't have a realisation, then it must mean that something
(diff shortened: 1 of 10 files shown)


regnat:
I've updated it on top of latest master. @edolstra @Ericson2314 mind having a look?

src/libstore/derivations.cc
```diff
@@ -756,8 +756,13 @@ std::optional<BasicDerivation> Derivation::tryResolveUncached(Store & store) {
         StringSet newOutputNames;
         for (auto & outputName : input.second) {
             auto actualPathOpt = inputDrvOutputs.at(outputName);
-            if (!actualPathOpt)
+            if (!actualPathOpt) {
+                warn("Input %s!%s missing, aborting the resolving",
+                    store.printStorePath(input.first),
+                    outputName
```
edolstra, line 762:
```suggestion
                warn("output %s of input %s missing, aborting the resolving",
                    outputName
                    store.printStorePath(input.first),
```

We probably shouldn't expose the `!output` syntax in the UI.

src/libstore/derivations.cc
```diff
@@ -782,6 +787,8 @@ std::optional<BasicDerivation> Derivation::tryResolve(Store& store, const StoreP
     // This is quite dirty and leaky, but will disappear once #4340 is merged
     static Sync<std::map<StorePath, std::optional<Derivation>>> resolutionsCache;
 
+    debug("Trying to resolve %s", store.printStorePath(drvPath));
```
edolstra, line 790:
```suggestion
    debug("trying to resolve %s", store.printStorePath(drvPath));
```

Messages are currently lower-case.

src/libstore/local-store.cc
```diff
@@ -883,7 +883,7 @@ StorePathSet LocalStore::queryValidDerivers(const StorePath & path)
 
 
 std::map<std::string, std::optional<StorePath>>
-LocalStore::queryDerivationOutputMapNoResolve(const StorePath& path_)
+LocalStore::queryPartialDerivationOutputMap(const StorePath& path_)
```
edolstra, line 886:
```suggestion
LocalStore::queryPartialDerivationOutputMap(const StorePath & path_)
```

src/libstore/build/derivation-goal.cc
```diff
@@ -1019,7 +1020,45 @@ void DerivationGoal::buildDone()
...
 void DerivationGoal::resolvedFinished() {
-    done(BuildResult::Built);
+    assert(resolvedDrv);
+
+    // If the derivation was originally a full `Derivation` (and not just
+    // a `BasicDerivation`, we must retrieve it because the `staticOutputHashes`
+    // will be wrong otherwise
+    Derivation fullDrv = *drv;
+    if (auto upcasted = dynamic_cast<Derivation *>(drv.get()))
+        fullDrv = *upcasted;
```
edolstra, line 1030: This seems a bit ugly... It makes it hard to reason about `DerivationGoal` if it sneakily does something different if the `drv` (which is a `BasicDerivation`) is not *really* a `BasicDerivation`. E.g. what does this mean for the remote build case (which is the case where we receive a `BasicDerivation`)? 

Maybe it's cleaner to handle this in the `DerivationGoal` constructors and `getDerivation` method. I.e. add a field `originalHashes` to `DerivationGoal`, which gets initialized in `DerivationGoal::DerivationGoal(... BasicDerivation ...)` from the `BasicDerivation`, and in `DerivationGoal::getDerivation` from the full derivation.

src/libstore/build/derivation-goal.cc
```diff
@@ -1019,7 +1020,45 @@ void DerivationGoal::buildDone()
...
+
+    auto originalHashes = staticOutputHashes(worker.store, fullDrv);
+    auto resolvedHashes = staticOutputHashes(worker.store, *resolvedDrv);
```
edolstra, line 1033: As an aside I find the "hash" terminology confusing since it's not really clear what it is a hash of... We should find a better name.

edolstra (approved):

src/libstore/derivations.cc
regnat: Indeed (I mean we might eventually… but even if we do this might not be the proper syntax to expose)

src/libstore/build/derivation-goal.cc
regnat: I agree that the terminology isn't really ideal. I couldn't think of anything better though…

Maybe we could just manipulate the `DrvOutput` `(hash, outputName)`, so that we don't have to name this precise bit which doesn't make much sense by itself

src/libstore/build/derivation-goal.cc
regnat: Fixed in 3c8b629c03f4e55835508119b8c4fca6bb428ef9

edolstra:
@regnat Looks like the build is currently failing:

```
src/libstore/derivations.cc:762:21: error: expected ')'
                    store.printStorePath(input.first),
```

regnat:
> @regnat Looks like the build is currently failing:
> 
> ```
> src/libstore/derivations.cc:762:21: error: expected ')'
>                     store.printStorePath(input.first),
> ```

Oh thanks, I ignored the CI warning thinking that it was only the transient OSX failure, but it was more than that indeed

src/libstore/build/derivation-goal.cc
```diff
@@ -124,6 +124,17 @@ DerivationGoal::DerivationGoal(const StorePath & drvPath, const BasicDerivation
     , buildMode(buildMode)
 {
     this->drv = std::make_unique<BasicDerivation>(BasicDerivation(drv));
+
+    auto outputHashes = staticOutputHashes(worker.store, drv);
```
Ericson2314, line 128: I'm a little  confused, is there any way we can abstract over this stuff in `haveDerivation` where it usual set? It looks like that is called just after?

src/libstore/derivations.cc
```diff
@@ -782,6 +787,8 @@ std::optional<BasicDerivation> Derivation::tryResolve(Store& store, const StoreP
     // This is quite dirty and leaky, but will disappear once #4340 is merged
```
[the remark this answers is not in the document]
regnat: Removed in a7829c70716bf9eb622da24e6675136a85c8cacb

regnat:
@Ericson2314 I've removed the now useless resolving cache.

@edolstra any blocker left for this PR?

src/libstore/build/derivation-goal.cc
Ericson2314: @regnat and I talked about this a rather meandering conversation (my fault!). @regnat mentioned the awkwardness of downcasting, but we do that a number of times already. `checkPathValidity` is the other place too.

Also, we should double check this is OK in the case of input-addressed remote buildings. Since the sender will strip away the `inputDrvs` but keep the output hashes, the remote side will calculate a different drvHash. Is this OK? I'm not sure.

regnat: > @regnat mentioned the awkwardness of downcasting, but we do that a number of times already. checkPathValidity is the other place too.

I honestly don't have much of an opinion on this (except that we should eventually fix the root cause that forces us to downcast… but that's work for another day).

> Also, we should double check this is OK in the case of input-addressed remote buildings. Since the sender will strip away the inputDrvs but keep the output hashes, the remote side will calculate a different drvHash. Is this OK? I'm not sure.

I might obviously miss something, but it seems OK for me as this new `drvHash` is only used for registering the realisation remotely, meaning that
1. It doesn't change the output path of the derivation
2. It's never sent back to the sender

(and it's a correct output hash from the builder's pov, so at the very least it doesn't hurt to have it registered, although it's not strictly required for input-addressed drvs)

PR #4346, opened 2020-12-10
WIP: Avoid using integer identifiers in install script

matthewbauer:
A common issue in multi-user installation is that a group with id
30000 or a user with 30001+ already exists. We don’t have either of
these reserved so any other package could also use these values,
causing conflicts.

- We don’t need to know up front what UID or GID to use for Linux’s
adduser. Leaving this blank will result in adduser finding an open id
in SYS_UID_MIN, SYS_UID_MAX for uid and SYS_GID_MIN, SYS_GID_MAX for
gid. In addition, --gid accepts the group name, no need for an
integer.

- For macOS’s dscl, we do need unique uids and gids up front. We can
figure these out by finding the greatest [ug]id allocated & adding 1.
Nothing else requires us to know what the id is.

Nix doesn’t appear to have any requirements that identifiers be
consecutive. Each nixbldN can be any value in UID_MIN, UID_MAX.

### DIFF
--- a/scripts/install-darwin-multi-user.sh
+++ b/scripts/install-darwin-multi-user.sh
@@ -60,24 +60,23 @@ poly_group_exists() {
     /usr/bin/dscl . -read "/Groups/$1" > /dev/null 2>&1
 }
 
-poly_group_id_get() {
-    dsclattr "/Groups/$1" "PrimaryGroupID"
-}
-
 poly_create_build_group() {
+    last_gid=$(/usr/bin/dscl . -list /Groups PrimaryGroupID | awk '{print $2}' | sort -ug | tail -1)
+
+    if [ -z "$last_gid" ]; then
+        # Something bad has happened; try to recover
+        last_gid=29999
+    fi
+
     _sudo "Create the Nix build group, $NIX_BUILD_GROUP_NAME" \
           /usr/sbin/dseditgroup -o create \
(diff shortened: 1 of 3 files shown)


abathur:
Some unknown subset of Big Sur versions might have complicated this. 

You can see more in #4531 and #4532, (and in #nix-darwin logs this past week) but TL;DR: it looks like we may need to stay between UID 200-400 on macOS, so it isn't sufficient to just start at the max current. 

I wrote a migration script for existing users which has at least at first swing at this (though it's probably a bit time-wasteful to use as-is

21stce:
The right way to fix this is to maintain a small database (for example in the form of a JSON or text file) of what the script has changed successfully. That would then also allow uninstalling nix.

Ericson2314:
@matthewbauer What exactly was WIP about this? Were you waiting on review of the concept to finish?

Ericson2314:
I am not too familiar with it, but yeah there is some sort of test.

abathur:
There are at least 2 types:
- some installer tests in the flake (I guess they just run on hydra?)
- install tests in GHA that can use cachix to run in your fork https://nixos.org/manual/nix/stable/contributing/hacking.html#installer-tests before merge

Ericson2314:
With the testing situation pretty good, you interested in finishing this @matthewbauer?

Issue #4347, opened 2020-12-11
Use the hash modulo for the derivation outputs


regnat:
The derivation outputs are currently indexed by the derivation hash (not modulo).
This means that a in a fixed-output will be propagated until late in the build loop, leading to possibly useless builds/fetches.

For example, assume that `hello` depends on `gcc` and `hello.src`, with `hello.src` fixed-output and the rest floating-ca.
Assume also that `hello!out` exists locally, but not `gcc`.
If we try to rebuild `hello!out`, nothing will happen as it's already there.

Now if we change `hello.src` (but in such a way that the output hash stays the same), what will happen is that `hello!out` will be a new derivation output (because the hash of `hello` will change), so we'll have to re-resolve it, which involves re-downloading `gcc` (or at least its resolution info if we end-up allowing them to be dangling).

If instead `hello!out` uses the hash modulo of `hello`, then it will stay the same regardless of the changes to `hello.src`, so there's no need to re-resolve it, and we save time and energy.

Issue #4344, opened 2020-12-10
Should we allow spitting the derivation output map per output?


Ericson2314:
The SQL database necessary stores rows individually, but we have a choice for other types of stores. https://github.com/NixOS/nix/pull/4330 currently stores a file per `derivation × output` in the binary cache, but we might also consider storing one file for the entire derivation and all its outputs.

The big idea in doing so is that:
- Ultimately, after all the trusting and sharing of mappings is "unwound" a mapping comes from a build
- All outputs must be built at once, so the "original" mappings should always be available for all outputs if they exist for any outputs at all
- Mixing outputs from different non-deterministic builds might cause issues.

Now, this does clash a bit with the important ability we have to GC individual outputs, and exchange them individually as needed between stores. We certainly don't want to loose that ability, nor have to mutate binary cache to add or remove entries from an existing file.

The crux of the solution is to ensure that a mapping for *any* output must come with, and keep alive, the mappings for *all* outputs. The slight challenge is to implement this with the local store and SQLite DB. Perhaps we'll need a single-column `DrvHash` table, with foreign keys from the build mappings to that table, and a restriction that one cannot delete an individual mapping unless one deletes all the mappings for the derivation?

PR #4348, opened 2020-12-11
Use the hash modulo in the derivation outputs

regnat:
Rather than storing the derivation outputs as `drvPath!outputName` internally,
store them as `drvHashModulo!outputName` (or `outputHash!outputName` for
fixed-output derivations).

This makes the storage slightly more opaque, but enables an earlier
cutoff in cases where a fixed-output dependency changes (but keeps the
same output hash) − same as what we already do for input-addressed
derivations.

Depends on: #4330
Fix: #4347

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -504,9 +504,6 @@ void DerivationGoal::inputsRealised()
             Derivation drvResolved { *std::move(attempt) };
 
             auto pathResolved = writeDerivation(worker.store, drvResolved);
-            /* Add to memotable to speed up downstream goal's queries with the
-               original derivation. */
-            drvPathResolutions.lock()->insert_or_assign(drvPath, pathResolved);
 
             auto msg = fmt("Resolved derivation: '%s' -> '%s'",
                 worker.store.printStorePath(drvPath),
@@ -2097,15 +2094,15 @@ struct RestrictedStore : public LocalFSStore, public virtual RestrictedStoreConf
 
     void registerDrvOutput(const Realisation & info) override
     {
-        if (!goal.isAllowed(info.id.drvPath))
-            throw InvalidPath("cannot register unknown drv output '%s' in recursive Nix", printStorePath(info.id.drvPath));
+        // XXX: Should we check for something here? Probably, but I'm not sure
+        // how
         next->registerDrvOutput(info);
     }
 
     std::optional<const Realisation> queryRealisation(const DrvOutput & id) override
     {
-        if (!goal.isAllowed(id.drvPath))
-            throw InvalidPath("cannot query the output info for unknown derivation '%s' in recursive Nix", printStorePath(id.drvPath));
+        // XXX: Should we check for something here? Probably, but I'm not sure
+        // how
         return next->queryRealisation(id);
     }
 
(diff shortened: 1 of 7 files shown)


regnat:
A few notes on that:

1. We could go a bit further for the sake of consistency, and use the exact same hash that input-addressed derivations use for computing the output path (in which case we could get rid of the output name as it's included in the path, or just keep it for information as input-addressed drv outputs do). But I'm not sure the added opacity is worth it − and it would kind-of conflict with #4344 as the different outptus would get a different hash
2. I couldn't find a way of storing this hash in the drv files themselves without breaking backwards compatibility, so we have to recompute it whenever we need it
3. As a consequence of 2., this relies a lot on the in-memory caching of drv hashes modulo to not be absurdly inefficient. Not a big issue I think, but not utterly clean either

src/libstore/build/derivation-goal.cc
```diff
@@ -2097,15 +2094,15 @@ struct RestrictedStore : public LocalFSStore, public virtual RestrictedStoreConf
 
     void registerDrvOutput(const Realisation & info) override
     {
-        if (!goal.isAllowed(info.id.drvPath))
-            throw InvalidPath("cannot register unknown drv output '%s' in recursive Nix", printStorePath(info.id.drvPath));
+        // XXX: Should we check for something here? Probably, but I'm not sure
+        // how
```
edolstra, line 2098: Probably `registerDrvOutput` shouldn't be allowed in `RestrictedStore` at all, since the caller is completely untrusted and supposed to be sandboxed.

src/libstore/build/derivation-goal.cc
```diff
@@ -2097,15 +2094,15 @@ struct RestrictedStore : public LocalFSStore, public virtual RestrictedStoreConf
...
         next->registerDrvOutput(info);
     }
 
     std::optional<const Realisation> queryRealisation(const DrvOutput & id) override
     {
-        if (!goal.isAllowed(id.drvPath))
-            throw InvalidPath("cannot query the output info for unknown derivation '%s' in recursive Nix", printStorePath(id.drvPath));
+        // XXX: Should we check for something here? Probably, but I'm not sure
+        // how
```
edolstra, line 2105: `queryRealisation()` should only succeed in `RestrictedStore` if `DrvOutput` refers to a derivation that the client added (via `addToStore`) *and* where it has built the output (i.e. has done a `buildDerivation()`). Otherwise it can leak info to which the sandboxed build isn't supposed to have access.

regnat:
Needs to double-check that this doesn't change the daemon protocol for `queryDrvOutputMap`

regnat:
> Needs to double-check that this doesn't change the daemon protocol for `queryDrvOutputMap`

This isn't the case, `queryDrvOutputMap` only manipulates `StorePath`s, so we're good on that front

src/libstore/build/derivation-goal.cc
regnat: I guess once we take the time to properly implement this, we should allow it if the drv output matches an allowed path, but that's complicated to check as we only know its hash modulo and not the drv path. So I'll just forbid everything for now

src/libstore/build/derivation-goal.cc
regnat: Indeed… like https://github.com/NixOS/nix/pull/4348#discussion_r541031277 I'll just restrict everything for now as it's slightly tricky to implement and we might change it eventually anyways.

I've opened #4353 to track it

Issue #3938, opened 2020-08-16
show path in JSON/TOML error reporting


zimbatm:
**Is your feature request related to a problem? Please describe.**

When importing JSON or TOML files in Nix code, it's not obvious where the data is coming from whenever an error is happening.

```
$ nix-instantiate --eval demo.nix 
error: --- JSONParseError ---------------------------------------------------- nix-instantiate
[json.exception.parse_error.101] parse error at line 5, column 13: syntax error while parsing array - unexpected string literal; expected ']'
```

It's great that the error shows in which line of the JSON the error is happening, but the user doesn't know from which file it's happening. To find that out, they have to dig into the Nix code.

A general principle in error reporting is to measure how much effort the user has to go through to find a resolution to the problem. How many hoops they have to jump through. Right now it is not completely straight-forward.

`./demo.nix`
```nix
let
  importJSON = file: builtins.fromJSON (builtins.readFile file);
in
  importJSON ./demo.json
```
`./demo.json`
```json
{
  "users": [
    "eelco",
    "grahamc"
    "zimbatm"
  ]
}
```

**Describe the solution you'd like**

As we see in the example, the `readFile` path has long been forgotten when `fromJSON` is being called.

I can see multiple potential solutions:

1. Add more builtins. Implement `importJSON` and `importTOML` (and `importYAML`?). That makes it trivial to report the path since the function has access to both data. The downside is that it adds more builtins.

2. Extend the string context to contain the source of the data. And change `fromJSON` to take advantage of the extra data during error reporting. The advantage is that this change could be done transparently without changing any of the APIs. Users that upgrade Nix will automatically get the improved error messages.

3. Re-implement `importJSON` with `tryEval` to extend the error message. I haven't really thought of the implications of this one as it seems like the clunkiest approach (it sort of breaks the lazy evaluation).

Solution 2 is my preferred one as it allows to disconnect the read from the eval.

domenkozar:
This is fixed:

```
$ /nix/store/hbwdvww1nqdav95iggp4zpgnpcwk037b-nix-2.6.0pre19700101_59a5f35/bin/nix-instantiate foo.nix --show-trace
error: [json.exception.parse_error.101] parse error at line 5, column 13: syntax error while parsing array - unexpected string literal; expected ']'

       … while decoding a JSON string

       at /home/domen/dev/nixpkgs/foo.nix:2:22:

            1| let
            2|   importJSON = file: builtins.fromJSON (builtins.readFile file);
[...]

PR #4351, opened 2020-12-11
primops/fromJSON: add error position in case of parse error

Ma27:
This makes it easier to track down where invalid JSON was passed to
`builtins.fromJSON`.

Related to #3938

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1621,7 +1621,12 @@ static RegisterPrimOp primop_toJSON({
 static void prim_fromJSON(EvalState & state, const Pos & pos, Value * * args, Value & v)
 {
     string s = state.forceStringNoCtx(*args[0], pos);
-    parseJSON(state, s, v);
+    try {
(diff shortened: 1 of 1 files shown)


src/libexpr/primops.cc
```diff
@@ -1621,7 +1621,13 @@ static RegisterPrimOp primop_toJSON({
 static void prim_fromJSON(EvalState & state, const Pos & pos, Value * * args, Value & v)
 {
     string s = state.forceStringNoCtx(*args[0], pos);
-    parseJSON(state, s, v);
+    try {
+        parseJSON(state, s, v);
+    } catch (JSONParseError &e) {
+        auto info = e.info();
+        info.errPos = pos;
+        throw EvalError(info);
```
edolstra, line 1629: Isn't it better to add the error position to `e` (maybe using `addErrorTrace`) and then rethrow `e`? That way it doesn't change the type.

Ma27:
@edolstra fixed, thanks.

PR #4355, opened 2020-12-12
Refactoring for private Value type

Infinisil:
This is a relatively simple but large refactoring for making `Value::type` private. This is done because that field exposes implementation details such as:
- Different kinds of lists: `tList1`, `tList2`, `tListN`
- Different kinds of thunks: `tThunk`, `tApp` and `tBlackhole`
- Different kinds of functions: `tLambda`, `tPrimOp` and `tPrimOpApp`

While lists already have a generic interface (`Value::isList()`, `Value::listElems()` and `Value::listSize()`), the same isn't true for thunks and functions. So in order to minimize the necessary changes, new functions were introduced to be able to distinguish between kinds of thunks and functions:
- `Value::isThunk()`, `Value::isApp()` and `Value::isBlackhole()`
- `Value::isLambda()`, `Value::isPrimOp()` and `Value::isPrimOpApp()`

In addition, a new type for the normal form type of a value was introduced, `NormalType`, which can be determined using `Value::normalType()`. This returns the same types as `builtins.typeOf` does, plus `nThunk` for thunk values. With this refactoring, this type is the canonical way to check the result type of an evaluation (plus `nThunk` for when it hasn't been evaluated yet).

And finally, in order to set the underlying `type` of a `Value`, setters for all the `ValueType`'s are introduced. In the future this should probably be handled only by `Value::mk<Type>` functions, such that these setters can be removed.

This PR will be useful for #4090, as such a feature will introduce a new attribute set type and will need a generic interface for accessing them.

Best review commit-by-commit

### DIFF
--- a/src/libexpr/attr-path.cc
+++ b/src/libexpr/attr-path.cc
@@ -67,7 +67,7 @@ std::pair<Value *, Pos> findAlongAttrPath(EvalState & state, const string & attr
 
         if (apType == apAttr) {
 
-            if (v->type != tAttrs)
+            if (v->normalType() != nAttrs)
                 throw TypeError(
                     "the expression selected by the selection path '%1%' should be a set but is %2%",
                     attrPath,
--- a/src/libexpr/attr-path.cc
+++ b/src/libexpr/attr-path.cc
@@ -67,7 +67,7 @@ std::pair<Value *, Pos> findAlongAttrPath(EvalState & state, const string & attr
 
         if (apType == apAttr) {
 
-            if (v->normalType() != nAttrs)
+            if (v->type() != nAttrs)
                 throw TypeError(
                     "the expression selected by the selection path '%1%' should be a set but is %2%",
                     attrPath,
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -36,13 +36,11 @@ void EvalState::forceValue(Value & v, const Pos & pos)
         Env * env = v.thunk.env;
         Expr * expr = v.thunk.expr;
         try {
-            v.setBlackhole();
+            v.mkBlackhole();
             //checkInterrupt();
             expr->eval(*this, *env, v);
         } catch (...) {
-            v.setThunk();
-            v.thunk.env = env;
-            v.thunk.expr = expr;
+            v.mkThunk(env, expr);
             throw;
         }
     }
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -56,7 +56,7 @@ void EvalState::forceValue(Value & v, const Pos & pos)
 inline void EvalState::forceAttrs(Value & v)
 {
     forceValue(v);
-    if (v.normalType() != nAttrs)
+    if (v.type() != nAttrs)
         throwTypeError("value is %1% while a set was expected", v);
 }
 
@@ -64,7 +64,7 @@ inline void EvalState::forceAttrs(Value & v)
 inline void EvalState::forceAttrs(Value & v, const Pos & pos)
 {
     forceValue(v, pos);
-    if (v.normalType() != nAttrs)
+    if (v.type() != nAttrs)
         throwTypeError(pos, "value is %1% while a set was expected", v);
 }
 
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -56,7 +56,7 @@ void EvalState::forceValue(Value & v, const Pos & pos)
 inline void EvalState::forceAttrs(Value & v)
 {
     forceValue(v);
-    if (v.type != tAttrs)
+    if (v.normalType() != nAttrs)
         throwTypeError("value is %1% while a set was expected", v);
 }
 
@@ -64,7 +64,7 @@ inline void EvalState::forceAttrs(Value & v)
 inline void EvalState::forceAttrs(Value & v, const Pos & pos)
 {
     forceValue(v, pos);
-    if (v.type != tAttrs)
+    if (v.normalType() != nAttrs)
         throwTypeError(pos, "value is %1% while a set was expected", v);
 }
 
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -36,11 +36,11 @@ void EvalState::forceValue(Value & v, const Pos & pos)
         Env * env = v.thunk.env;
         Expr * expr = v.thunk.expr;
         try {
-            v.type = tBlackhole;
+            v.setBlackhole();
             //checkInterrupt();
             expr->eval(*this, *env, v);
         } catch (...) {
-            v.type = tThunk;
+            v.setThunk();
             v.thunk.env = env;
             v.thunk.expr = expr;
             throw;
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -32,7 +32,7 @@ LocalNoInlineNoReturn(void throwTypeError(const Pos & pos, const char * s, const
 
 void EvalState::forceValue(Value & v, const Pos & pos)
 {
-    if (v.type == tThunk) {
+    if (v.isThunk()) {
         Env * env = v.thunk.env;
         Expr * expr = v.thunk.expr;
         try {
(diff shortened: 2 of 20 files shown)


src/libexpr/eval.cc
```diff
@@ -198,8 +193,11 @@ string showType(const Value & v)
         case tPrimOpApp:
             return fmt("the partially applied built-in function '%s'", string(getPrimOp(v)->primOp->name));
         case tExternal: return v.external->showType();
+        case tThunk: return "a thunk";
+        case tApp: return "a function application";
+        case tBlackhole: return "a black hole";
     default:
-        return showType(v.type);
+        return showType(v.normalType());
```
Infinisil, line 200: Regarding `showType`, the function `showType(ValueType)` was removed and replaced with `showType(NormalType)`, because it's not really possible to get a `ValueType` anymore. The new `showType(NormalType)` also aborts for thunks, and it's only used on forced values. The `showValue(Value)` function however does need to be able to print thunks, since it's used on non-forced values. So this now handles them before passing the normal type to `showType(NormalType)`

src/libexpr/eval.cc
```diff
@@ -1575,36 +1573,36 @@ void ExprConcatStrings::eval(EvalState & state, Env & env, Value & v)
            since paths are copied when they are used in a derivation),
            and none of the strings are allowed to have contexts. */
         if (first) {
-            firstType = vTmp.type;
+            firstType = vTmp.normalType();
```
Infinisil, line 1576: The `ExprConcatStrings::eval` function is an interesting place where the new normal form type has to be used to ensure it keeps working the same way

src/libexpr/value-to-json.cc
```diff
@@ -70,14 +70,17 @@ void printValueAsJSON(EvalState & state, bool strict,
...
             v.external->printValueAsJSON(state, strict, out, context);
             break;
 
-        case tFloat:
+        case nFloat:
             out.write(v.fpoint);
             break;
 
+        case nNonNormal:
+            throw TypeError("cannot convert %1% to JSON without --strict", showType(v));
```
Infinisil, line 82: Took the liberty to improve the error message slightly here. This is the only behavioral change afaik

src/libexpr/value-to-xml.cc
```diff
@@ -118,14 +118,19 @@ static void printValueAsXML(EvalState & state, bool strict, bool location,
...
             for (unsigned int n = 0; n < v.listSize(); ++n)
                 printValueAsXML(state, strict, location, *v.listElems()[n], doc, context, drvsSeen);
             break;
         }
 
-        case tLambda: {
+        case nFunction: {
+            if (!v.isLambda()) {
+                // TODO: Serialize primops and other functions
+                doc.writeEmptyElement("unevaluated");
```
Infinisil, line 131: This `NormalType` change actually discovered a small bug in the xml conversion: it prints `unevaluated` for `tPrimOp` and `tPrimOpApp`'s. I didn't want to bother changing the format of that so I left it.

src/libexpr/value.hh
```diff
@@ -90,7 +106,46 @@ std::ostream & operator << (std::ostream & str, const ExternalValueBase & v);
 
 struct Value
 {
+private:
     ValueType type;
+
+friend std::string showType(const Value & v);
+friend void printValue(std::ostream & str, std::set<const Value *> & active, const Value & v);
```
Infinisil, line 113: These function seemed best to make friends with for simplicity

src/libexpr/eval-inline.hh
```diff
@@ -32,39 +32,39 @@ LocalNoInlineNoReturn(void throwTypeError(const Pos & pos, const char * s, const
...
         try {
-            v.type = tBlackhole;
+            v.setBlackhole();
             //checkInterrupt();
             expr->eval(*this, *env, v);
         } catch (...) {
-            v.type = tThunk;
+            v.setThunk();
             v.thunk.env = env;
             v.thunk.expr = expr;
```
edolstra, line 45: If we're refactoring this stuff anyway, then it's better to have `v.setThunk()` take arguments to initialize the value properly, i.e.
```
v.mkThunk(env, expr);
```

(Also `mkThunk` instead of `setThunk` for consistency with all the existing `mk...` functions.)

src/libexpr/value.hh
```diff
@@ -90,7 +106,46 @@ std::ostream & operator << (std::ostream & str, const ExternalValueBase & v);
...
+
+    inline void setInt() { type = tInt; };
+    inline void setBool() { type = tBool; };
+    inline void setString() { type = tString; };
+    inline void setPath() { type = tPath; };
+    inline void setNull() { type = tNull; };
+    inline void setAttrs() { type = tAttrs; };
+    inline void setList1() { type = tList1; };
+    inline void setList2() { type = tList2; };
+    inline void setListN() { type = tListN; };
```
edolstra, line 125: These setters break the abstraction of having a single list type. They should be unnecessary if `mkList` is a friend or member of `Value`.

src/libexpr/eval-inline.hh
edolstra: Hiding the initialisation of thunks and other types is a good thing anyway if we ever want to revive the parallel evaluator, since (IIRC) it required fields to be set in a particular order. E.g. when transitioning from tBlackhole -> tString, you have to set the `s` field *before* setting type to `tString`, since otherwise another thread might think the value is initialized.

src/libexpr/attr-path.cc
```diff
@@ -67,7 +67,7 @@ std::pair<Value *, Pos> findAlongAttrPath(EvalState & state, const string & attr
 
         if (apType == apAttr) {
 
-            if (v->type != tAttrs)
+            if (v->normalType() != nAttrs)
```
edolstra, line 70: Instead of `normalType()` it would be nicer to write `v->isAttrs()`.

Do we actually need `normalType()`? I guess it's useful in `switch` statements...

src/libexpr/value.hh
```diff
@@ -29,6 +29,22 @@ typedef enum {
     tFloat
 } ValueType;
 
+// This type abstracts over all actual value types in the language,
+// grouping together implementation details like tList*, different function
+// types, and types in non-normal form (so thunks and co.)
+typedef enum {
+    nThunk,
```
edolstra, line 36: `NormalType` might be a bit of a confusing name since thunks are not normal forms. Maybe `AbstractType`?

src/libexpr/eval-inline.hh
Infinisil: Sounds good, I'll create `v.mk<Type>` methods. I can also remove the global `mkInt(Value & v, NixInt n)` functions (though e.g. `EvalState::mkAttrs` should stay since it initializes memory)

src/libexpr/attr-path.cc
Infinisil: Other than it being useful in switch statements, it's also necessary in `ExprConcatStrings::eval`, which uses `firstType = vTmp.normalType()` to store the type, later comparing it to the types of other expressions. Similarly, `expectType` in `src/libexpr/flake/flake.cc` does a comparison between types. This wouldn't be possible with `is<Type>()` functions.

src/libexpr/value.hh
Infinisil: How about this: We make `ValueType type` be an unnamed union with
```c++
union {
  tInt = 1,
  // ...
} type;
```

And then let's use the `ValueType` name for the current `NormalType`.

Infinisil: Then we should probably also use `Value::type()` instead of `Value::normalType()`

edolstra: Yeah that sounds good.

Infinisil: Wasn't able to easily use an anonymous enum because the `showType` and `printValue` friends wouldn't have access to the variants like that. Instead I renamed the original `ValueType` to `InternalType`, so that `NormalType` can be `ValueType` instead. Also renamed the field to `Value::internalType` which is needed for `Value::type()` to be a function.

src/libexpr/value.hh
Infinisil: These are now replaced with `Value::mkList(size_t)`

src/libexpr/eval-inline.hh
Infinisil: I left most static `mk*` functions around since I don't have motivation to refactor this as well. But this could easily be done later.

Infinisil:
Addressed all comments.

Note that this will be a backwards incompatible change for Nix plugins, which now instead of setting `type = tExternal` or another type have to use the `Value::mk*` functions

PR #4358, opened 2020-12-12
Change builder argv[0] to full path, instead of only basename

puckipedia:
See commit message.
I have a few projects that use Nix, but cannot resolve their own location without knowing the full path they were called with (e.g. GCC). It seems, either way, that this decision was arbitrarily made (see 822794001cb4260b8c04a7bd2d50d890edae709a), and calling the builder by its full path (so `/nix/store/...-foo/bar` instead of `bar`) follows expected behavior more closely.

Running a similar patch, but against 2.3.5, locally.

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -2804,13 +2804,13 @@ void DerivationGoal::runChild()
                 args.push_back(drv->builder);
             } else {
                 builder = drv->builder.c_str();
-                args.push_back(std::string(baseNameOf(drv->builder)));
+                args.push_back(std::string(drv->builder));
             }
         }
 #else
         else {
             builder = drv->builder.c_str();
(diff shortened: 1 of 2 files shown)


grahamc:
I wonder if there is a problem w.r.t. initial bootstrapping using /bin/sh in the build, which points to something else (like ash) on disk? I have no reason to think this is a problem, just curious.

puckipedia:
I haven't actually seen anything use `/bin/sh`, outside a single NixOS hydra test. the `/bin/sh` in the Nix sandbox is just a static busybox binary, and being called as `/bin/sh` is fine for that binary anyways.

edolstra:
I don't think we should change this. It's been this way for a long time without causing problems, it's an observable change to the behaviour of the sandbox, and it's not really clear why it's necessary (e.g. `gcc` is never invoked as the top-level builder). If a top-level program really needs an absolute `argv[0]`, you can always re-exec with an absolute path.

PR #4366, opened 2020-12-15
Use the fs accessor for readInvalidDerivation

regnat:
Extend `FSAccessor::readFile` to allow not checking that the path is a
valid one, and rewrite `readInvalidDerivation` using this extended
`readFile`.

Several places in the code use `readInvalidDerivation`, either because
they need to read a derivation that has been written in the store but
not registered yet, or more generally to prevent a deadlock because
`readDerivation` tries to lock the state, so can't be called from a
place where the lock is already held.
However, `readInvalidDerivation` implicitely assumes that the store is a
`LocalFSStore`, which isn't always the case.

The concrete motivation for this is that it's required for `nix copy
--from someBinaryCache` to work, which is tremendously useful for the
tests.

### DIFF
--- a/src/libstore/local-fs-store.cc
+++ b/src/libstore/local-fs-store.cc
@@ -19,10 +19,10 @@ struct LocalStoreAccessor : public FSAccessor
 
     LocalStoreAccessor(ref<LocalFSStore> store) : store(store) { }
 
-    Path toRealPath(const Path & path)
+    Path toRealPath(const Path & path, bool requireValidPath = true)
     {
         auto storePath = store->toStorePath(path).first;
-        if (!store->isValidPath(storePath))
+        if (requireValidPath && !store->isValidPath(storePath))
             throw InvalidPath("path '%1%' is not a valid store path", store->printStorePath(storePath));
         return store->getRealStoreDir() + std::string(path, store->storeDir.size());
     }
@@ -61,9 +61,9 @@ struct LocalStoreAccessor : public FSAccessor
         return res;
     }
 
-    std::string readFile(const Path & path) override
+    std::string readFile(const Path & path, bool requireValidPath = true) override
     {
-        return nix::readFile(toRealPath(path));
+        return nix::readFile(toRealPath(path, requireValidPath));
     }
 
     std::string readLink(const Path & path) override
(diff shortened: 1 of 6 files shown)


src/libstore/local-fs-store.cc
```diff
@@ -61,9 +61,9 @@ struct LocalStoreAccessor : public FSAccessor
         return res;
     }
 
-    std::string readFile(const Path & path) override
+    std::string readFile(const Path & path, bool requireValidPath = true) override
```
Ericson2314, line 64: I think the overrides shouldn't have the default args? Elsewhere in the codebase they at least are only part of the declarations, IIRC.

Ericson2314 (approved):
Great idea!

src/libstore/local-fs-store.cc
regnat: That's a good question. Triggered me to read https://stackoverflow.com/questions/6464404/virtual-function-default-arguments-behaviour which is… interesting.

The gist of it seems to that if we only call `accessor.readFile(path)` where `accessor` is declared of type `FSAccessor` (which is probably the case) then we don't need the default arguments for the overrides. But that looks quite fragile. Maybe I could just split the interface as:

```cpp
virtual bool isValidPath(const Path& path);
virtual std::string readFileUnchecked(const Path& path);
std::string readFile(const Path& path) {
  if (!isValidPath(path)) throw InvalidPath(...);
  return readFileUnchecked(path);
}
```

Ericson2314: Hmm I read it the other way around, where up casting certainly gives you the outer default arg, but no upcasting would also give you the outer default arg if there is no inner default arg.

regnat: Nah, just checked:

```cpp
struct Foo {
  virtual void foo(int n = 1) { }
};

struct Bar : Foo {
  void foo(int n) override { }
};

int main() {
  Foo x;
  x.foo();
  Bar y;
  y.foo();
}
```

`x.foo()` typechecks fine, but `y.foo()` doesn't

Ericson2314: Sorry to have misled you with that wrong interpretation then.

PR #4372, opened 2020-12-16
Add a new Cmd type working on RealisedPaths

regnat:
Where a `RealisedPath` is a store path with its history, meaning either
an opaque path for stuff that has been directly added to the store, or a
`Realisation` for stuff that has been built by a derivation

This is a low-level refactoring that doesn't bring anything by itself
(except a few dozen extra lines of code :/ ), but raising the
abstraction level a bit is important on a number of levels:

- Commands like `nix build` have to query for the realisations after the
  build is finished which is fragile (see
  27905f12e4a7207450abe37c9ed78e31603b67e1 for example). Having them
  oprate directly at the realisation level would avoid that
- Others like `nix copy` currently operate directly on (built) store
  paths, but need a bit more information as they will need to register
  the realisations on the remote side

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -704,23 +704,43 @@ Buildables build(ref<Store> store, Realise mode,
     return buildables;
 }
 
-StorePathSet toStorePaths(ref<Store> store,
-    Realise mode, OperateOn operateOn,
+std::set<RealisedPath> toRealisedPaths(
+    ref<Store> store,
+    Realise mode,
+    OperateOn operateOn,
     std::vector<std::shared_ptr<Installable>> installables)
 {
-    StorePathSet outPaths;
-
+    std::set<RealisedPath> res;
     if (operateOn == OperateOn::Output) {
         for (auto & b : build(store, mode, installables))
             std::visit(overloaded {
                 [&](BuildableOpaque bo) {
-                    outPaths.insert(bo.path);
+                    res.insert(bo.path);
                 },
                 [&](BuildableFromDrv bfd) {
+                    auto drv = store->readDerivation(bfd.drvPath);
+                    auto outputHashes = staticOutputHashes(*store, drv);
                     for (auto & output : bfd.outputs) {
-                        if (!output.second)
-                            throw Error("Cannot operate on output of unbuilt CA drv");
-                        outPaths.insert(*output.second);
+                        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+                            if (!outputHashes.count(output.first))
+                                throw Error(
+                                    "The derivation %s doesn't have an output "
+                                    "named %s",
+                                    store->printStorePath(bfd.drvPath),
+                                    output.first);
+                            auto outputId = DrvOutput{outputHashes.at(output.first), output.first};
+                            auto realisation = store->queryRealisation(outputId);
+                            if (!realisation)
+                                throw Error("Cannot operate on output of unbuilt CA drv %s", outputId.to_string());
+                            res.insert(RealisedPath{*realisation});
+                        }
+                        else {
+                            // If ca-derivations isn't enabled, behave as if
+                            // all the paths are opaque to keep the default
+                            // behavior
+                            assert(output.second);
+                            res.insert(*output.second);
+                        }
                     }
                 },
             }, b);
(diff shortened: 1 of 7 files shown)


src/libstore/realisation.hh
```diff
@@ -28,8 +39,60 @@ struct Realisation {
...
+
+    GENERATE_CMP(OpaquePath, me->path);
 };
 
-typedef std::map<DrvOutput, Realisation> DrvOutputs;
+
+/**
+ * A store path with all the history of how it went into the store
+ */
+struct RealisedPath {
```
Ericson2314, line 60: Is all this C++ worth it? In derivation.hh we just wrapped a variant and called it a day. Other places we `typedef`ed the variant. It would be good to be consistent whatever route we choose, I suppose.

Ericson2314, line 60: On a separate note, I think it would be good to make a plan for reducing into fewer types this, `StorePathWithOutputs`, `Buildable`, and perhaps even the separation of `inputsDrv` vs `inputSrcs`. We've had too many similar data types for too long, but only recently have we been bringing them closer together.

src/libcmd/installables.cc
```diff
@@ -704,23 +704,43 @@ Buildables build(ref<Store> store, Realise mode,
...
+                    auto outputHashes = staticOutputHashes(*store, drv);
                     for (auto & output : bfd.outputs) {
-                        if (!output.second)
-                            throw Error("Cannot operate on output of unbuilt CA drv");
-                        outPaths.insert(*output.second);
+                        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+                            if (!outputHashes.count(output.first))
+                                throw Error(
+                                    "The derivation %s doesn't have an output "
+                                    "named %s",
```
edolstra, line 728:
```suggestion
                                    "the derivation '%s' doesn't have an output named '%s'",
```

src/libcmd/installables.cc
```diff
@@ -704,23 +704,43 @@ Buildables build(ref<Store> store, Realise mode,
...
+                                    store->printStorePath(bfd.drvPath),
+                                    output.first);
+                            auto outputId = DrvOutput{outputHashes.at(output.first), output.first};
+                            auto realisation = store->queryRealisation(outputId);
+                            if (!realisation)
+                                throw Error("Cannot operate on output of unbuilt CA drv %s", outputId.to_string());
```
edolstra, line 734:
```suggestion
                                throw Error("cannot operate on an output of unbuilt content-addresed derivation '%s'", outputId.to_string());
```

src/libstore/realisation.cc
```diff
@@ -46,4 +46,35 @@ Realisation Realisation::fromJSON(
...
+    // closure
+    StorePathSet initialStorePaths, pathsClosure;
+    for (auto& path : startPaths)
+        initialStorePaths.insert(path.path());
+    store.computeFSClosure(initialStorePaths, pathsClosure);
+    ret.insert(startPaths.begin(), startPaths.end());
+    ret.insert(pathsClosure.begin(), pathsClosure.end());
+}
+
+void RealisedPath::closure(Store& store, RealisedPath::Set& ret) const
```
edolstra, line 68:
```suggestion
void RealisedPath::closure(Store& store, RealisedPath::Set & ret) const
```

src/libstore/realisation.hh
```diff
@@ -3,6 +3,34 @@
...
+ * ```
+ *
+ * will generate comparison operators semantically equivalent to:
+ * ```
+ * bool operator<(const ClassName& other) {
+ *   return field1 < other.field1 && field2 < other.field2 && ...;
+ * }
+ * ```
+ */
+#define GENERATE_ONE_CMP(COMPARATOR, MY_TYPE, FIELDS...) \
```
edolstra, line 20: Maybe move this into a separate file in libutil?

src/libstore/realisation.hh
```diff
@@ -32,8 +54,60 @@ struct Realisation {
...
+    Raw raw;
+
+    using Set = std::set<RealisedPath>;
+
+    RealisedPath(StorePath path) : raw(OpaquePath{path}) {}
+    RealisedPath(Realisation r) : raw(r) {}
+
+    /**
+     * Syntactic sugar to run `std::visit` on the raw value:
+     * path.visit(blah) == std::visit(blah, path.raw)
```
edolstra, line 90: TBH I wouldn't mind getting rid of the use of `std::visit`, which often seems unnecessarily verbose/complicated because of the use of closures compared to
```
if (auto realisation = variant.get_if<Realisation>)
  do something 
else if (auto realisation = variant.get_if<OpaquePath>)
  do something else
```

edolstra (approved):

src/libstore/realisation.hh
regnat: I don't know in the general case (it's indeed very verbose and easily hard to read, but at the same time the exhaustiveness check is a killer feature…). But in this precise case, there was only one single call to `visit`, so it's definitely not worth it indeed.

Fixed in 4f84b2c4d638ace8e745fb78860a549217dbc00d

src/libstore/realisation.hh
regnat: Yeah I eventually removed because it was a lot of boilerplate for nearly nothing

src/libstore/realisation.hh
regnat: OK, I guess that can be useful elsewhere.

Done in 343687405d6de18c0b1b7aab814e3ea6a0a61420

src/libstore/realisation.hh
edolstra: @regnat For some reason that commit isn't showing up in this PR, did you push it?

PR #4378, opened 2020-12-17
Fix the detection of already built drv outputs

regnat:
PRs #4370 and #4348 had a bad interaction in that the second broke the fist
one in a not trivial way.

The issue was that since #4348 the logic for detecting whether a
derivation output is already built requires some logic that was specific
to the `LocalStore`.

It happens though that most of this logic could be upstreamed to any `Store`,
which is what this commit does.


regnat:
As a nice side effect, this also makes this detection logic also work with non-ca derivations

Issue #4375, opened 2020-12-16
Add more documentation to nix-bundle


laikq:
**Describe the bug**

Hi, I would like to request some more documentation for nix-bundle. I'm really hyped for this feature! But I think missing documentation is one large problem of Nix/NixOS/nixpkgs, and I would like to give some perspective.

From my perspective as a user who has used and written flakes, but is not fluent in C++ and does not really understand the structure of nix (the program), here are my questions (going through the manual entry for `nix bundle`):
- "bundle an application so that it works outside of the Nix store": Hmm, I see that the `nix bundle` command is supposed to be very flexible with the `--bundler` flag. However, I think it would be useful to describe the default behavior in short here (what *is* the generated file?, what does it require? -- kernel namespaces?)
- synopsis: what does "installable" mean? The description of [nix-bundle](https://github.com/matthewbauer/nix-bundle) mentions an entrypoint to the derivation, but it seems `nix bundle` uses something like the `defaultApp` attribute of a nix flake, but I couldn't find documentation for that anywhere (except for a [usage example on NixOS Wiki](https://nixos.wiki/wiki/Flakes)).
- `--bundler`: What is a bundler? (Is it a flake with a `defaultApp` which in turn accepts a store path as an argument?) Are there examples for different bundlers than the default one? (is there a docker bundler, for example?)

**Nix version**

2.4pre20201201_5a6ddb3

zimbatm:
Also: reminder that the default bundler is still `github:matthewbauer/nix-bundle`. I don't have anything against the author, I think this should be controlled by the NixOS org.

laikq:
Hmm, it seems that a `--bundler` is a Nix flake with a `defaultBundler` attribute, and the `defaultBundler` attribute must evaluate to a function accepting `{ program-path, system }` as arguments and return a derivation, which is the `result`.

PR #4379, opened 2020-12-17
Add 'nix run' manpage

laikq:
See #4375. I'm new to Nix (the program), so expect beginner errors :upside_down_face:


edolstra:
Note that there already is a PR that adds a manpage for `nix run` and other commands, see https://github.com/NixOS/nix/pull/4336.

laikq:
Nice, thank you!

Issue #4364, opened 2020-12-15
Make `nix-build-if-changed` a first class feature of nix


Profpatsch:
**Is your feature request related to a problem? Please describe.**

When using nix on CI, I want every `nix-build`/`nix-store --realize` to be a no-op if it’s already built. This means already built `OR` available on a substituter. If you run a naïve `nix-build`, it will download the full run time closure, which can be many megabytes of useless networking if all you want to do is verify whether the build succeeds.

One step further, if I want to generate a script with nix (e.g. a test suite), and then run it on CI (outside of the nix sandbox), it obviously needs to download the runtime closure. However, it *does not* have to download it if it didn’t change, because that means we ran the test before and the result won’t change (as long as all test dependencies are tracked by nix). Compare to how [`bazel`](bazel.build/) handles test suites in a similar fashion.

**Describe the solution you'd like**

This requirement is so common that multiple people have written wrappers around nix independently of each other. The ones that I know of:

* [Mic92/nix-build-uncached](https://github.com/Mic92/nix-build-uncached/)
* [Profpatsch/nix-build-if-changed.py](https://gist.github.com/Profpatsch/fbbc9a0006e246076f11efca0387d293)

I think at this point it would make sense for nix to get one or more flags to make those scripts unnecessary. (They effectively have to parse `nix-build --dry-run` output to figure out what nix would do and then evaluate a second time.)

A simple `--fetch=no`/`--fetch=if-changed`/`--fetch=always (default)` flag to `nix-build` might be enough to get most of the advantages; it would only download the resulting path of it had to do a build, and it would signal (via exit code?) whether it actually did the build and download. This might confuse multiple separate issues though, so I don’t know if that’s the best semantics.

* `--fetch=always`: current behavior`
* `--fetch=if-changed`: the result will only be fetched if it had to be built (aka wasn’t either in the local store or completely substitutable). This is a good way to e.g. build test scripts if you want to run them as well.
* `--fetch=no`: the result should never be fetched. You might think “but if I have to build I will have it in my store by design”, but that’s not true if you have a `--builder` configured because you don’t want to actually build things on CI. In that case, even though the builder builds the resulting path, all CI wants to know is if the store path can be successfully built, it does not need the result.

**Describe alternatives you've considered**

* Keep the scripts and add some kind of primitive check to nix to expose whether it has already built the result, so that we don’t need to parse stderr. This would still duplicate evaluation time however!
[...]

PR #4387, opened 2020-12-20
Make `nix-build --store whatever` work

Ericson2314:
Much of the logic for building is really about scheduling builds, not actually performing build steps. We really ought to separate `build-derivation.cc` to make that concept distinct, but for now we just hack it up with downcasts.

Why? Well consider #3946 and #4364, an feature that's clearly desired. The crux of the matter is that with today's work flow of doing a local `nix-build` and then a copy to the cache, `nix` doesn't know what the *end goal* is until it's too late. First schedules to have everything local, and then it copies it all back to the cache.

With this workflow, the schedule knows which store is the end destination. And just as only things you don't have are sent to the remote builder, now only things that are not in the cache are sent to the local store to build.

There might still be some extra downloading: I wanted to add a "corrupt nars, still no-op cause nar info" test, but that didn't work. Even if there was, however, that's now just a matter of small tweaks to fix, but the architecture is fundamentally rectified to support this feature.

Progress on #3946 and #4364.

~~Depends on #4366~~


colemickens:
Does this support a scenario where you need to check against multiple stores? For me/#3964, it's important that I'm able to check against `cache.nixos.org` and `my-own.cachix.org` at the same time.

Ericson2314:
If you mean the thing I said would need a "layered store", no this PR doesn't, but that's a separate feature in my mind. We can always do it too.

This PR is purely about "removing the weaknesses and restrictions that make additional features appear necessary", but "layered store" is probably necessary either way :).

src/libstore/build/derivation-goal.cc
```diff
@@ -3269,7 +3289,13 @@ void DerivationGoal::registerOutputs()
             }
         }
 
+        auto localStoreP = dynamic_cast<LocalStore *>(&worker.store);
+        if (!localStoreP)
+            Unsupported("Can only register outputs with local store");
```
edolstra, line 3294:
```suggestion
            throw Unsupported("Can only register outputs with local store");
```

src/libstore/build/derivation-goal.cc
```diff
@@ -3398,11 +3424,16 @@ void DerivationGoal::registerOutputs()
        paths referenced by each of them.  If there are cycles in the
        outputs, this will fail. */
     {
+        auto localStoreP = dynamic_cast<LocalStore *>(&worker.store);
+        if (!localStoreP)
+            Unsupported("Can only register outputs with local store");
```
edolstra, line 3429:
```suggestion
            throw Unsupported("can only register outputs with local store");
```

src/libstore/store-api.hh
```diff
@@ -372,6 +372,21 @@ public:
     void queryPathInfo(const StorePath & path,
         Callback<ref<const ValidPathInfo>> callback) noexcept;
 
+    /* Check whether the given valid path info is sufficiently well-formed
+       (e.g. hash content-address or signature) in order to be included in the
```
edolstra, line 376: "Well-formed" is a bit vague, since to me that just suggests it's syntactically valid. Maybe say "is signed by a trusted public key or is content-addressed"?

tests/binary-cache-build-remote.sh
```diff
@@ -0,0 +1,13 @@
+source common.sh
+
+clearStore
+clearCacheCache
+
+# Fails without remote builders
+(! nix-build --store "file://$cacheDir" dependencies.nix)
+
+# Succeeds with default store as build remote.
+nix-build --store "file://$cacheDir" --builders 'auto - - 1 1' -j0 dependencies.nix
```
edolstra, line 10: Maybe check here that the result actually exists in `$cacheDir`? E.g. `nix path-info --store "file://$cacheDir" $outPath` or something like that.

src/build-remote/build-remote.cc
```diff
@@ -71,11 +71,15 @@ static int main_build_remote(int argc, char * * argv)
...
+        auto store = openStore();
 
         /* It would be more appropriate to use $XDG_RUNTIME_DIR, since
            that gets cleared on reboot, but it wouldn't work on macOS. */
-        currentLoad = store->stateDir + "/current-load";
+        currentLoad = "/current-load";
+        if (auto localStore = store.dynamic_pointer_cast<LocalFSStore>())
+            currentLoad = std::string { localStore->stateDir } + currentLoad;
+        else
+            currentLoad = settings.nixStateDir + currentLoad;
```
edolstra, line 82:
```suggestion
        auto currentLoadName = "/current-load";
        if (auto localStore = store.dynamic_pointer_cast<LocalFSStore>())
            currentLoad = std::string { localStore->stateDir } + currentLoadName;
        else
            currentLoad = settings.nixStateDir + currentLoadName;
```

Ericson2314:
OK comments are fixed.

Ericson2314:
Thanks!

PR #4389, opened 2020-12-22
Backport issue 3964 substitution loop

roberth:
Backport #4158

I had to revert the test, because it appears to rely on >2.3 functionality. See commit message.

@ajs124 could you try this?


ajs124:
Thank!

I don't really have a good reproducer, so I'm not sure if this really helps, but I'll test it, once I run into this again.

edolstra (approved):

Issue #3803, opened 2020-07-12
Nix repl does not work with flakes


Mic92:
**Describe the bug**

Before it was possible to use to use `nix repl <nixpkgs/nixos>` to get nixos in a repl.

This seems not possible anymore with flakes `nix repl ".#"`:

```
Welcome to Nix version 2.4pre20200622_334e26b. Type :? for help.

Loading '.#'...
error: --- SysError ----------------------------------------------------------- nix
getting status of '/home/joerg/.homesick/repos/dotfiles/.#': No such file or directory
```

How do I load flakes in a nix repl?

Mic92:
Current workaround:

```
cat > ./nixos-config.nix <<EOF
(builtins.getFlake (toString ./.)).nixosConfigurations.turingmachine
EOF
nix repl ./nixos-config.nix
```

Ma27:
FWIW, if a snippet like this in your `flake.nix` you can also do `nix run .#repl`:

``` nix
{
  outputs = { flake-utils, ... }: {
      apps.repl = flake-utils.lib.mkApp {
        drv = pkgs.writeShellScriptBin "repl" ''
          confnix=$(mktemp)
          echo "builtins.getFlake (toString $(git rev-parse --show-toplevel))" >$confnix
          trap "rm $confnix" EXIT
          nix repl $confnix
        '';
      };
  };
}
```

FlorianFranzen:
Another issue I have in this regard is that `getFlakes` does not seem to ever provide `self.rev` or `self.lastModified`.

edrex:
How can `:lf` be used to load a particular output in a flake-based NixOS config?

```
> nix repl
Welcome to Nix 2.6.1. Type :? for help.

nix-repl> :lf /etc/nixos#myhost
error: unexpected fragment 'myhost' in flake reference '/etc/nixos#myhost'
```

srid:
@edrex 
<img width="558" alt="image" src="https://user-images.githubusercontent.com/3998/157729463-e2d6e285-9981-4e37-b989-fb38312f6d36.png">

edrex:
thanks @srid, I swear I tried that but it's working now.

RuRo:
For anyone coming to this issue from google, I found that the most "elegant" solution is to set
```nix
nix.nixPath = [ "nixpkgs=${inputs.nixpkgs}" ];
```
in your system config, where `inputs` are the system flake inputs passed via `specialArgs`.

[...]

PR #4392, opened 2020-12-23
repl: add :lh command

Synthetica9:
`:lh <hash>` is a convenience shortcut, it does the same as
`:l https://github.com/nixos/nixpkgs/archive/<hash>.tar.gz`.

I'm not sure if this sort of thing is within scope, but it is something I would use very often myself, so I can imagine this would also be useful to other people.

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -415,6 +415,7 @@ bool NixRepl::processLine(string line)
              << "  :e <expr>     Open the derivation in $EDITOR\n"
              << "  :i <expr>     Build derivation, then install result into current profile\n"
              << "  :l <path>     Load Nix expression and add it to scope\n"
+             << "  :lh <hash>    Download and load Nixpkgs commit from GitHub\n"
              << "  :p <expr>     Evaluate and print expression recursively\n"
              << "  :q            Exit nix-repl\n"
              << "  :r            Reload all files\n"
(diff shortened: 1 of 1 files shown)


edolstra:
It would probably be better to add flake support to the repl (#3803), e.g. a command like `:lf <flakeref>` would allow `:lf nixpkgs/<hash>`.

Synthetica9:
> It would probably be better to add flake support to the repl (#3803), e.g. a command like :lf <flakeref> would allow :lf nixpkgs/<hash>.

I agree, it was just that I saw this would be a pretty trivial change to make, so I wanted to have the option out there. Full flake support seems to be a significantly bigger undertaking.

Mic92:
> > It would probably be better to add flake support to the repl (#3803), e.g. a command like :lf  would allow :lf nixpkgs/.
> 
> I agree, it was just that I saw this would be a pretty trivial change to make, so I wanted to have the option out there. Full flake support seems to be a significantly bigger undertaking.

Is there no easy to access function that loads flakes? There is a builtin `builtin.getFlake` - I would expect this one to be also reachable from c++.

terlar:
Still a good feature request

Ericson2314:
We now have the `:lf` tha @edolstra suggested. Furthermore, one can also do `:a fetch-however-one-likes`. Can we close this?

Ericson2314:
Closing per the above.

PR #4397, opened 2020-12-24
Cast variants fully for libc++10

matthewbauer:
libc++10 seems to be stricter on what it allows in variant conversion.
I'm not sure what the rules are here, but this is the minimal change
needed to get through the compilation errors.


r-burns:
This fixes nixUnstable under gcc10Stdenv for me :+1:

edolstra:
Looks like this breaks the build with GCC 9:
```
src/libexpr/eval-cache.cc: In member function 'nix::Value& nix::eval_cache::AttrCursor::forceValue()':
src/libexpr/eval-cache.cc:396:25: error: invalid use of non-static member function 'nix::ValueType nix::Value::type() const'
  396 |         else if (v.type == nPath)
      |                  ~~~~~~~^~~~~~~~
```

src/libexpr/eval-cache.cc
```diff
@@ -393,9 +393,9 @@ Value & AttrCursor::forceValue()
         if (v.type() == nString)
             cachedValue = {root->db->setString(getKey(), v.string.s, v.string.context),
                            string_t{v.string.s, {}}};
-        else if (v.type() == nPath)
-            cachedValue = {root->db->setString(getKey(), v.path), v.path};
-        else if (v.type() == nBool)
+        else if (v.type == nPath)
```
edolstra, line 396: Bad merge? This definitely needs to stay `v.type()`.

PR #4418, opened 2021-01-02
Fix #4234, send build logs to Nix 2.3

### LINKED ISSUE #4234: nixUnstable doesn't send build output to nix 2.3.8 clients
**Describe the bug**

Building derivations on a system with nixUnstable using the stable nix client doesn't include any build output.

**Steps To Reproduce**

Building a derivation.

```nix

pkgs.runCommandNoCC "hello" {} ''
  echo Hello World!
  echo ${toString builtins.currentTime} > $out
''
```

```sh
$ nix-build hello.nix
  /nix/store/r3m7jjkzqqrjkw94w31jh73qcsrddl9h-hello.drv
building '/nix/store/r3m7jjkzqqrjkw94w31jh73qcsrddl9h-hello.drv'...
/nix/store/br1dc1nzfn9pxn94sdl4q6cmgczyd0xi-hello
```

**Expected behavior**

```sh
$ nix-build hello.nix
  /nix/store/fc2cxlwl80ldsln2zaf28vyqbf9w1ncx-hello.drv
building '/nix/store/fc2cxlwl80ldsln2zaf28vyqbf9w1ncx-hello.drv'...
Hello World!
/nix/store/wxcl8ndlddvk76ash8r8q00ha6wl76nh-hello
```

- nix-build 2.3.8
- nix-daemon (master a3a22bd80478dd2827e8617e682084523684200b)

roberth:
Fixes #4234 introduced in #3073

Sends build logs to clients in the extra way that Nix 2.3 clients expects to receive.

The conditional on clientVersion isn't accurate for a certain
range of master commits, but those are unsupported anyway.

It's not accurate for repeated builds either, because that
would complicate the code for too little gain.

~@ regnat this is one possible solution. If you have another solution that is perhaps more in line with #3073, we could go with that instead.~ It changed the way the build logs are treated internally, affecting the log messages on the socket. This PR supplements the log to be compatible with 2.3. I don't think there's much else to it.

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -142,6 +142,14 @@ struct TunnelLogger : public Logger
 
     void result(ActivityId act, ResultType type, const Fields & fields) override
     {
+        if (GET_PROTOCOL_MINOR(clientVersion) < 22 && type == resBuildLogLine && settings.verboseBuild) {
+            // Nix < 2.4 expects build logs to be logged as general logs at lvlError.
+            // We probably shouldn't do this for repeated builds, but that's not
+            // something we can know here without complicating the implementation
+            // unnecessarily.
+            auto lastLine = fields[0].s;
+            log(lvlError, lastLine);
+        }
         if (GET_PROTOCOL_MINOR(clientVersion) < 20) return;
         StringSink buf;
         buf << STDERR_RESULT << act << type << fields;


roberth:
@edolstra could you have a look?

Rahul75-dev (commented):
To make all issyes to migrate linux to citrix command to form language c,c++ html modules changes are done

roberth:
> To make all issyes to migrate linux to citrix command to form language c,c++ html modules changes are done

@Rahul75-dev this makes no sense at all. Please delete it or I'll have to report this as spam.

regnat (approved):

roberth:
We should increment the protocol version after this is merged, for 2.3-based clients that want to support a wide range of daemon versions correctly.

I would do it myself in this PR, but it's already been three months, so it will probably just cause merge conflicts. :(

roberth:
Personally, I would actually prefer for the behavior of all nix-daemon versions to remain as simple and predictable as possible. You could focus on other important things that remain relevant in the long term / after the upgrade.

PR #4434, opened 2021-01-07
Add 'nix store prefetch-{file,tarball}'

edolstra:
These replace `nix-prefetch-url` and `nix-prefetch-url --unpack`, respectively.


bjornfor:
I always thought "archive" would be a better name than "tarball" in this context, given that it's meant to work on zip files too. Thoughts?

edolstra:
Yes, or "tree", to make it consistent with `fetchTree`.

edolstra:
... which makes me realize it should just take a flake ref as argument (so it could be used to fetch stuff like Git repos).

Issue #4423, opened 2021-01-04
nix flakes: add support for git submodules


DanilaFe:
I recently learned about Nix Flakes, and I tried out making my blog a perfectly reproducible build. The trouble is, my blog uses a Git submodule to reference some code that I then present in my blog posts. This is fine and all, but I run into an issue - although I can fetch my blog source using Flakes' `input` attribute, it doesn't fetch submodules! This means that the code that I reference in my build process is simply not there.

This seems to be _almost_ possible; from a brief glance around Nix's source code, I discovered that a flake input gets turned into a "regular" input, which can point to a git repo and have a `submodules` attribute that, I think, would do what I want. It feels as though it was intended that this would work, but it doesn't. 

It doesn't work because, for a reason unknown to me, a flake input's attributes must all be strings (except for `flake`). Here's the code that rejects non-string arguments:

https://github.com/NixOS/nix/blob/724b7f4fb660212a97ba6482208c299158720c5b/src/libexpr/flake/flake.cc#L123-L127

If non-string arguments were not rejected, a `submodules` field could be forwarded to the Git input scheme, which I suspect would make this work. Is there a reason that Flakes rejects all non-string "third party" attributes? It puts them into an attribute set anyway. Right now, I'm falling into a rather amusing loop:

![duality of nix](https://user-images.githubusercontent.com/4361282/103520843-a93ff280-4e2c-11eb-8cae-16b4c5ca9e46.png)

If `submodules` is a string, it's rejected by the Git input scheme. But if it's a `bool`, it's rejected by the Flake code.

knedlsepp:
I think this is a duplicate of https://github.com/NixOS/nix/issues/4357

blaggacao:
Apparently closed by #4435

knedlsepp:
I think there is still one piece missing: the flake itself is not being fetched using submodules.

DanilaFe:
Yes, perhaps it's worth closing this issue, and opening the original one (since this is a duplicate)? 

Also, I am not sure how to go about requiring a flake itself to pull its own submodules. If a flake only has inputs and outputs, and an external definition determines the method for downloading it (a flake can be a local folder, too, or a tarball, can't it?), then where's the compromise? 

A flake is assumed to be a git repo, so it would make sense for a universal "fetch submodules" option, perhaps as an `output` attribute. This would be orthogonal to the `type = "git"; submodules = "true"` attributes inside an `input`, since it wouldn't work for non-flake inputs.

[...]

PR #4435, opened 2021-01-08
Allow Flake inputs to accept boolean and integer attributes

### LINKED ISSUE #4357: Flake inputs: git submodules/recursive checkout

Can't just add git url to inputs to fetch sources with submodules. 

A `recursive: bool` field in flake inputs available when url refers to a git repo

Not using flake.lock or flake inputs, but instead using `fetchFromGitHub`

**Additional context**

DanilaFe:
Looking through the code base for Nix, I found that it creates `fetchers::Input` from the attributes given to Flakes' input refs (see also lines 108 through 133 in `flake.cc`): 

https://github.com/NixOS/nix/blob/920e6a6920fa0ae82150bb2b0c210a03ccf5919b/src/libexpr/flake/flakeref.cc#L191-L198

The attributes accepted by the fetches can be strings, integers, or booleans. However, currently, the code in `flake.cc` rejects all attributes that are not strings. It does, however, forward all extra string arguments, no matter what they are, to the `fetchers::Input`. This seems inconsistent, and prevents input types like Git from accepting certain arguments, including `submodules`.

This PR expands the code in `flake.cc` to accept non-string arguments. Combined with this, I was able to make non-flake inputs with recursive git modules work as follows:

```Nix
inputs = {
    blog-source = {
      flake = false;
      url = "https://my.example.com/repo.git";
      type = "git";
      submodules = true;
    }
};
```

Note that `type = "git"` is already supported by Nix; this PR allows the `submodules = true` argument to be passed in alongside `type`. I believe this would close #4357 (and its duplicate, #4423).

So far, the only non-integer attributes are  `submodules` (bool), `revCount` (int), and `lastModified` (int). While the latter two are not particularly useful, I think the benefits for git submodules and consistency make this a worthwhile change.

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -120,15 +120,19 @@ static FlakeInput parseFlakeInput(EvalState & state,
                 expectType(state, nString, *attr.value, *attr.pos);
                 input.follows = parseInputPath(attr.value->string.s);
             } else {
-                if (attr.value->type() == nString) {
-                    attrs.emplace(attr.name, attr.value->string.s);
-                } else if (attr.value->type() == nBool) {
-                    attrs.emplace(attr.name, Explicit<bool>{ attr.value->boolean });
-                } else if (attr.value->type() == nInt) {
-                    attrs.emplace(attr.name, attr.value->integer);
-                } else {
-                    throw TypeError("flake input attribute '%s' is %s while a string, boolean, or integer is expected",
-                        attr.name, showType(*attr.value));
+                switch (attr.value->type()) {
+                    case nString:
+                        attrs.emplace(attr.name, attr.value->string.s);
+                        break;
+                    case nBool:
+                        attrs.emplace(attr.name, Explicit<bool> { attr.value->boolean });
+                        break;
+                    case nInt:
+                        attrs.emplace(attr.name, attr.value->integer);
+                        break;
+                    default:
+                        throw TypeError("flake input attribute '%s' is %s while a string, boolean, or integer is expected",
+                            attr.name, showType(*attr.value));
                 }
             }
         } catch (Error & e) {
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -131,7 +131,7 @@ static FlakeInput parseFlakeInput(EvalState & state,
                         attrs.emplace(attr.name, attr.value->integer);
                         break;
                     default:
-                        throw TypeError("flake input attribute '%s' is %s while a string, boolean, or integer is expected",
+                        throw TypeError("flake input attribute '%s' is %s while a string, Boolean, or integer is expected",
                             attr.name, showType(*attr.value));
                 }
             }
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -120,11 +120,16 @@ static FlakeInput parseFlakeInput(EvalState & state,
                 expectType(state, nString, *attr.value, *attr.pos);
                 input.follows = parseInputPath(attr.value->string.s);
             } else {
-                if (attr.value->type() == nString)
+                if (attr.value->type() == nString) {
                     attrs.emplace(attr.name, attr.value->string.s);
-                else
-                    throw TypeError("flake input attribute '%s' is %s while a string is expected",
+                } else if (attr.value->type() == nBool) {
+                    attrs.emplace(attr.name, Explicit<bool>{ attr.value->boolean });
+                } else if (attr.value->type() == nInt) {
+                    attrs.emplace(attr.name, attr.value->integer);
+                } else {
+                    throw TypeError("flake input attribute '%s' is %s while a string, boolean, or integer is expected",
                         attr.name, showType(*attr.value));
+                }
             }
         } catch (Error & e) {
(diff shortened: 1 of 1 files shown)


src/libexpr/flake/flake.cc
```diff
@@ -120,11 +120,16 @@ static FlakeInput parseFlakeInput(EvalState & state,
...
                 input.follows = parseInputPath(attr.value->string.s);
             } else {
-                if (attr.value->type() == nString)
+                if (attr.value->type() == nString) {
                     attrs.emplace(attr.name, attr.value->string.s);
-                else
-                    throw TypeError("flake input attribute '%s' is %s while a string is expected",
+                } else if (attr.value->type() == nBool) {
+                    attrs.emplace(attr.name, Explicit<bool>{ attr.value->boolean });
+                } else if (attr.value->type() == nInt) {
```
Infinisil, line 127: I'd probably go for a `switch` here

DanilaFe: I didn't realize it was an enum! Yes, `switch` works much better here.

DanilaFe: I pushed a commit that switches this to a switch.

knedlsepp:
Do you know how one would make sure that nix fetches submodules for `self` with this?

src/libexpr/flake/flake.cc
```diff
@@ -120,11 +120,20 @@ static FlakeInput parseFlakeInput(EvalState & state,
...
+                        attrs.emplace(attr.name, attr.value->string.s);
+                        break;
+                    case nBool:
+                        attrs.emplace(attr.name, Explicit<bool> { attr.value->boolean });
+                        break;
+                    case nInt:
+                        attrs.emplace(attr.name, attr.value->integer);
+                        break;
+                    default:
+                        throw TypeError("flake input attribute '%s' is %s while a string, boolean, or integer is expected",
```
edolstra, line 134:
```suggestion
                        throw TypeError("flake input attribute '%s' is %s while a string, Boolean, or integer is expected",
```

DanilaFe:
@knedlsepp unfortunately, I'm not sure. This change affects only the `inputs`, which `self` is not part of. So, to make sure a flake has its submodules initialized and cloned, it must be that whatever is invoking the flake has submodules enabled. With this change, it's impossible to force the _current_ flake to have its _own_ submodules resolved, only its dependencies. I think this is still consistent, though - a flake's own `flake.nix` does not have control over the fetcher used to retrieve it, as far as I know.

L-as:
> @knedlsepp unfortunately, I'm not sure. This change affects only the `inputs`, which `self` is not part of. So, to make sure a flake has its submodules initialized and cloned, it must be that whatever is invoking the flake has submodules enabled. With this change, it's impossible to force the _current_ flake to have its _own_ submodules resolved, only its dependencies. I think this is still consistent, though - a flake's own `flake.nix` does not have control over the fetcher used to retrieve it, as far as I know.

What you say makes sense, but when you consider how it's actually used, it's very inconvenient. A flake will either always want its submodules initialized, or never want to. A flake that has submodules will not work if you don't enable them, so there is no point in allowing a user to not fetch the submodules for a flake (if they are required to build it).

Since whether submodules should be fetched depends on the flake, I propose that the flake set for itself whether its submodules should be fetched. You could also just unconditionally fetch them for flakes, since I can't see any scenario where you wouldn't want them. It's not foreign repositories being fetched for nixpkgs; they're flakes fit for nix. If the owner has added a submodule, it's likely required for using the flake.

Would a PR enabling fetching of submodules unconditionally for all flakes be accepted?

nrdxp:
I'm getting a very strange error when trying to use this. _After_ successfully fetching the module and all submodules, nix gave an error that disallowed usage saying:
```console
submodule is not a Boolean
```
when it fact it is. I was able to work around by manually removing the `submodule` keys from the flake.lock and it works without it, since the repo has already been fetched.

hurricanehrndz:
I am also getting submodule is not a Boolean.

mhuesch:
I am running into the same issue that @L-as nicely described, where my `self` flake has submodules and I am stuck in the inconvenient situation of not being able to fetch them.

I would be interested in contributing to a flake-submodule-fetcher PR if such a thing would be accepted.

kenranunderscore:
I'm also in the same situation as @mhuesch and @L-as.
[Here](https://discourse.nixos.org/t/nix-flakes-and-submodules/7904/4) is a post by someone showing how to correct this when invoking from the command line. Is there an update on how to fetch submodules for `self`? As prefixing every `nix build` target with `.?submodules=1` seems quite inconvenient for me.

Ofc I could try converting the submodule to a flake input, but that creates different inconveniences for us.

con-f-use:
By the way, to add insult to injury, nix doen't even tell that it's because you're using something from a submodule. It just says something like:

```
error: getting status of '/nix/store/qjj8lgnp70m9842n2qr4bljb3hrd04dh-source/mysubmodule': No such file or directory
```

nrdxp:
> I could try converting the submodule to a flake input, but that creates different inconveniences for us.

It would be inconvenient to do manually, but if Nix itself could gain the power of locking the submodules as flake inputs so that Nix can handle the fetching of them lazily, that might be the best solution in the long run.

Issue #4442, opened 2021-01-11
allowSubstitutes is an anti-feature


domenkozar:
I can't describe how many hours I and people I have helped have lost to save a few miliseconds of machine time.

For example `preferLocalBuild` ignores builders and `--max-jobs` per https://github.com/NixOS/nix/issues/3810 giving the user the impression that flags are being ignored.

Having such exceptions completely change the way Nix works makes it also really hard to debug binary cache issues - I've gotten lots of reports of why is binary cache not substituting.

Refs #3686

domenkozar:
Oh and another point: the reason the user might be using remote builder / binary cache is to avoid HDD space pollution and there's no way to disable that.

grahamc:
I think I would rather keep `preferLocalBuild` and have it respect `--max-jobs` instead. I don't relish the idea of uploading 10's of GB of data to a remote server over DSL to write a short text file, just to download it again.

Ninlives:
the `preferLocalBuild` overwrites the `--max-jobs` may be an anti-feature, but allowSubstitutes/preferLocalBuild themselves are useful. I do not want to waste time on querying binary cache when I just want to write a text file.

Ma27:
The problem I have with removing `preferLocalBuild` is that I regularly use it to configure on a per-derivation level whether or not I want remote builds. As @grahamc stated, this is helpful to avoid uploading tons of data to a remote server just to build e.g. a proprietary package where only an unpack and a `patchelf` happens. This is particularly helpful for me since I'm regularly (not during the pandemic though) at places where I have rather bad internet.

domenkozar:
It's a great power feature that hits Nix beginners really had and given that it's used in nixpkgs there no opt-out.

Of course all expirienced Nix users will say they have no problem with it, but it makes behavior with substituters and remove builders hard to debug and reason about.

bgamari:
I use `preferLocalBuild` quite heavily as it can mean the difference between waiting a few seconds and a few minutes in the case of derivations with large inputs and outputs.

domenkozar:
I often (need to start tracking how often) get Cachix support request why is something not being substituted and I have templated an answer, but the response from Nix beginners is usually: wow

This probably can be fixed with better logging/messaging and fixing all the bugs, but I feel like that's one of those things that will never materialize.

Ma27:
Maybe it would help to make this clear in at least the new progress bar that's currently under development: https://github.com/NixOS/nix/pull/4296

[...]

PR #4443, opened 2021-01-12
Don't let 'preferLocalBuild' override 'max-jobs=0'

### LINKED ISSUE #3810: preferLocalBuild should not override max-jobs=0

rickynils:
This resolves #3810 by changing the behavior of `max-jobs = 0`, so
that specifying the option also avoids local building of derivations
with the attribute `preferLocalBuild = true`.

A more complicated resolution was suggested by @grahamc in #3810, that also made it possible to retain the original behavior of `max-jobs=0`. However, it has been suggested ([here](https://github.com/NixOS/nix/issues/3810#issuecomment-658257637) and [here](https://github.com/NixOS/nix/issues/4442#issuecomment-758003203)) that just letting `preferLocalbuild` respect `max-jobs=0` is an acceptable solution too.

### DIFF
--- a/src/libstore/parsed-derivations.cc
+++ b/src/libstore/parsed-derivations.cc
@@ -101,6 +101,10 @@ bool ParsedDerivation::canBuildLocally(Store & localStore) const
         && !drv.isBuiltin())
         return false;
 
+    if (settings.maxBuildJobs.get() == 0
+        && !drv.isBuiltin())
+        return false;
+
     for (auto & feature : getRequiredSystemFeatures())
         if (!localStore.systemFeatures.get().count(feature)) return false;
 
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -679,13 +679,9 @@ void DerivationGoal::tryToBuild()
 }
 
 void DerivationGoal::tryLocalBuild() {
-    bool buildLocally = buildMode != bmNormal || parsedDrv->willBuildLocally(worker.store);
-
-    /* Make sure that we are allowed to start a build.  If this
-       derivation prefers to be done locally, do it even if
-       maxBuildJobs is 0. */
+    /* Make sure that we are allowed to start a build. */
     unsigned int curBuilds = worker.getNrLocalBuilds();
-    if (curBuilds >= settings.maxBuildJobs && !(buildLocally && curBuilds == 0)) {
+    if (curBuilds >= settings.maxBuildJobs) {
         worker.waitForBuildSlot(shared_from_this());
         outputLocks.unlock();
         return;


edolstra:
Can you also update this code in `DerivationGoal::tryLocalBuild()`:
```c++
    /* Make sure that we are allowed to start a build.  If this
       derivation prefers to be done locally, do it even if
       maxBuildJobs is 0. */
    unsigned int curBuilds = worker.getNrLocalBuilds();
    if (curBuilds >= settings.maxBuildJobs && !(buildLocally && curBuilds == 0)) {
        worker.waitForBuildSlot(shared_from_this());
        outputLocks.unlock();
        return;
    }
```

rickynils:
@edolstra Is there anything else than the comment that actually needs to be changed in that code? My commit changes the behavior of `willBuildLocally` so that it always is false if `max-jobs = 0`. That means `buildLocally` will be false in the `DerivationGoal::tryLocalBuild()` as long as `buildMode == bmNormal`. I might have missed something though?

domenkozar:
Probably the comment needs changing too.

edolstra:
The condition `&& !(buildLocally && curBuilds == 0)` can be dropped.

rickynils:
@edolstra Thanks, I've removed it now.

edolstra:
Thanks!

src/libstore/parsed-derivations.cc
```diff
@@ -101,6 +101,10 @@ bool ParsedDerivation::canBuildLocally(Store & localStore) const
         && !drv.isBuiltin())
         return false;
 
+    if (settings.maxBuildJobs.get() == 0
+        && !drv.isBuiltin())
```
Ericson2314, line 105: Why do we have this isn't builtin check here? do derivations running builtin builders not count towards max jobs?! I wouldn't think so

PR #4449, opened 2021-01-14
Fix gcc10 build

jonringer:
Wasn't able to update to the latest nixUnstable in https://github.com/NixOS/nixpkgs/pull/109309 because it failed to build with gcc10.

Updated the flake so that it tracks unstable which has gcc10 now.

Also fixed a compiler error for gcc where there was an implicit cast from `int64_t` to `long unsigned int`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,16 +18,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1602702596,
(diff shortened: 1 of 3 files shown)


edolstra:
Thanks, I've cherry-picked the gcc 10 fix, but I'd like to stay on 20.09 for now so I left out the change to nixpkgs-unstable.

Issue #4230, opened 2020-11-06
nix-shell should support mixing -p with a filename


lilyball:
**Describe the bug**

`nix-shell` can take a filename to source its expression from, or it can take a `-p` flag to declare packages to use, but it cannot mix the two. This is really unfortunate because it means I can't use a file that declares a customized/pinned version of nixpkgs and pull individual packages from that with `-p`. I have to instead construct an expression using `-E`, which is annoying and not something I want to teach my coworkers how to do. I can't use the `-I nixpkgs=somePath` option either as I have a file that vends nixpkgs rather than a nixpkgs checkout.

What I'd really like to do is to be able to write

```bash
#!/usr/bin/env nix-shell
#!nix-shell pinned.nix -i bash -p bash jq
echo script goes here
```

But today, if I do this, it tries to interpret `deps.nix` as a package.

There's two issues here. The first is that the `-p` flag declares "all arguments not otherwise associated with a flag are to be interpreted as packages, regardless of their location on the command line", which is rather confusing as it makes `nix-shell foo -p bar` behave like `nix-shell -p foo bar`, and the second is that the package form hard-codes `import <nixpkgs>`.

**Suggested resolution**

I think two things should be done:

1. Stop treating `nix-shell foo -p bar` like `nix-shell -p foo bar`. I realize this is technically a backwards-incompatible change, although the `nix-shell` documentation indicates that the usage is `-p packages...`.
2. If I pass a single filename to `nix-shell` prior to the `-p` flag, replace `import <nixpkgs> {}` with something like `(let expr = import $path; in if builtins.isFunction expr then expr {} else expr)`. Ideally any `--arg` or `--argstr` flags would be supported here as well. If multiple filenames are given to `nix-shell` along with the `-p` flag then they could either all be imported and merged together, or just throw an error.

A less invasive change would be to define a new flag I can use to replace the `import <nixpkgs> {}` with, such that `nix-shell foo -p bar` still retains the current behavior (though as I stated before that behavior is confusing). Or this could be hacked further by replacing the `import <nixpkgs> {}` with something like `(let customPath = builtins.tryEval (toString <nix-shell-pkgs>); expr = import (if customPath.success then customPath.value else <nixpkgs>); in if builtins.isFunction expr then expr {} else expr)`. This way I could run `nix-shell -I nix-shell-pkgs=pinned.nix -p bash jq` and it would pull from my custom file instead. I don't see any clear benefit to this over defining a new flag though, and the new flag would be easier to document.

[...]

PR #4452, opened 2021-01-14
nix-shell: Add strictDeps-proof --tools, --inputs

### LINKED ISSUE #4254: Bash completion in nix-shell -p (also solving support for strictDeps by default)

`nix-shell` doesn't currently provide bash completion for new commands that provide their own completion scripts.
https://github.com/NixOS/nixpkgs/pull/103501 aims to change that, but can't support completions in `nix-shell -p` because that command adds to the wrong attribute, `buildInputs` instead of `nativeBuildInputs`.

`nix-shell` should set `nativeBuildInputs` (one way or another).

This aligns with having `strictDeps` by default (which is not a requirement though). StrictDeps enforces desirable properties like outputs not retaining references to build tools. It also prevents mistakes that break cross-compilation.

We can pick from these designs or think of something better.

###### principled scheme

`-p` sets both and emits a warning
`--tool` adds to `nativeBuildInputs`
`--input` adds to `buildInputs`

###### minimal scheme

`-p` sets `nativeBuildInputs`
`--input` adds to `buildInputs`

With the minimal scheme we don't have to update all documentation to change `-p`s to `--tool`s, but it will silently break some `nix-shell` invocations.

###### compatible scheme

Like the principled scheme but without the warning. Does nothing to help towards strictDeps.

Also add `buildInputs` to `XDG_DATA_DIRS`, but this "non-strictDeps" behavior is risky, potentially affecting any build in Nixpkgs.

**Additional context**

roberth:
Unlike `-p`, which will be phased out, these are future proof with `strictDeps` in mind.

A more immediate benefit is that `--tools` is capable of providing shell autocompletions.

Closes #2504, closes #4254 

@Ericson2314  perhaps we could duplicate `-p` arguments in `buildInputs` and `nativeBuildInputs`, emulating non-strictDeps. Seems rather crude though. What do you think?

Preserves the odd `packages -p` behavior also mentioned in #4230 and assumes intent of `-p` aka `--inputs` for those packages.

### DIFF
--- a/doc/manual/src/command-ref/nix-shell.md
+++ b/doc/manual/src/command-ref/nix-shell.md
@@ -13,7 +13,7 @@
   [`--exclude` *regexp*]
   [--pure]
   [--keep *name*]
-  {{`--packages` | `-p`} {*packages* | *expressions*} … | [*path*]}
+  {[*path*] | {`--tools` | `--inputs`} {*packages* | *expressions*} … | }
 
 # Description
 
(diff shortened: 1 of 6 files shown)


edolstra:
Not sure whether we should add more features to the (legacy) `nix-shell` command. Everything we add to it creates technical debt since we have to port those features to the `nix` command (or not, but that would make migrating to the `nix` command harder).

roberth:
I don't think `nix-shell` can be considered legacy, because we need it for shebangs. If we migrated it to `#!/usr/bin/env nix shell` we'd run into the issue that only a single argument is widely supported in shebangs. `env` will receive `nix shell` as a single argument, which doesn't work. The [workaround](https://unix.stackexchange.com/questions/399690/multiple-arguments-in-shebang/477651#477651) for that isn't portable.

We seem to have these use cases here
 1. entering a stdenv-based shell: `nix develop`
 2. running some commands without installing: `nix shell`
 3. shebangs: `nix-shell`

I don't think there's a real need for interactively invoked nix-shells based on stdenv (`-p`/`--tools`/`--inputs`) because if what you're doing is complicated enough that you depend on more than `$PATH`, you should probably write a file for it anyway.

I understand your caution with the legacy options, but in this case I don't think it creates technical debt, because we don't need to support this in the `nix` command.

Also, if for whatever reason we do want stdenv-based ad-hoc shells in the `nix` command, it'd be nice to port a complete solution rather than one with known problems.

roberth:
Still valid. Also, since the nix command can't replace `nix-shell` due to unix shebang legacy, `nix command` isn't an excuse.

I'll happily rebase this after a preliminary approval from a maintainer.

roberth:
:notes:

fricklerhandwerk:
Triaged in Nix team meeting:

- @ericson2314: a `nixpkgs` CLI can solve this without current or new layer violations
    - Nixpkgs has various typos of dependencies, but `nix-shell` has a very naive view of this, and can not practically be kept up to date
        - for example `nix-shell` is not aware of cross-compilation as it predates Nixpkgs support for that
    - A more robust solution would be having a dedicated `nixpkgs` CLI and get `nix-shell` out of the way
        - it would in fact be worth a great project, as it is not blocked on Nix maintainers
- @thufschmit: don't necessarily think we should have a `nixpkgs` CLI, but in any case Nix should not have to bother with these issues
    - this also means that we may have to reconsider `nix develop`

PR #4461, opened 2021-01-18
Fix some logging with ca derivations

regnat:
- Make the error message for missing outputs more useful
- Fix the error message when a dep is missing

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -394,7 +394,7 @@ OutputPathMap Store::queryDerivationOutputMap(const StorePath & path) {
     OutputPathMap result;
     for (auto & [outName, optOutPath] : resp) {
         if (!optOutPath)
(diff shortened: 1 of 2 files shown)


src/libstore/store-api.cc
```diff
@@ -394,7 +394,7 @@ OutputPathMap Store::queryDerivationOutputMap(const StorePath & path) {
     OutputPathMap result;
     for (auto & [outName, optOutPath] : resp) {
         if (!optOutPath)
-            throw Error("output '%s' has no store path mapped to it", outName);
+            throw Error("output '%s' of derivation '%s' has no store path mapped to it", outName, path.to_string());
```
edolstra, line 397: Shouldn't this be `printStorePath(path)`?

regnat: Oh indeed. Fixed

Ericson2314 (approved):

Issue #3868, opened 2020-07-28
Flag for `nix search` to include other meta attributes such as homepage


lilyball:
**Is your feature request related to a problem? Please describe.**
I often have to look up the homepages of software when searching, either because the description is too generic, or missing (e.g. with autogenerated packages), or because I want more info.

**Describe the solution you'd like**
`nix search` should have flags that include additional meta properties, including `homepage`, `longDescription`, and `licenses`.

lilyball:
I still want this

lilyball:
I still want this.

Also, having to periodically comment "I still want this" is very tedious.

06kellyjac:
Still important

superherointj:
This could be useful as a replacement for this command line:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/make-tarball.nix#L67

Which could fix a repology batch issue that is causing imprecise stats and queries.

Related post: https://github.com/NixOS/nix/pull/4463#issuecomment-762382017

Aleksanaa:
Still important

debtquity:
would be nice to implement. In it's current state, `nixpkgs` maintainers would need to stuff the short description (`meta.description`) with keywords in order to appear in relevant searches with `nix search`.

PR #4463, opened 2021-01-18
nix-env: Use pname and version attrs in -q

jtojnar:
Previously[1], we started including `pname` and `version` keys in `nix-env -qa --json` output (and thus in packages.json).

But those values were obtained using parseDrvName, which does not work in many cases, for example when a project does not follow Nix’s arbitrary requirement that version names have to start with a digit.

The keys are also misleading as many modern Nixpkgs packages pass identically named attributes to mkDerivation but these keys do not use those values.

Let’s fix this discrepancy by officially recognizing `pname` and `version` attributes as special and using them in the JSON output.

This will be useful for Repology as well.

[1]: https://github.com/NixOS/nix/commit/cd933b22d2041b7efc348dcc09ff255967ffc663

### DIFF
--- a/src/libexpr/get-drvs.cc
+++ b/src/libexpr/get-drvs.cc
@@ -57,6 +57,36 @@ std::string DrvInfo::queryName() const
 }
 
 
+std::optional<std::string> DrvInfo::queryPname() const
+{
+    if (!attrs) {
+        pname = std::optional<std::string>{std::nullopt};
+    } else if (!pname.has_value()) {
+        auto i = attrs->find(state->symbols.create("pname"));
+        if (i != attrs->end() && i->value->type() == nString)
+            pname = state->forceStringNoCtx(*i->value);
+        else
+            pname = std::optional<std::string>{std::nullopt};
+    }
+    return *pname;
+}
+
+
+std::optional<std::string> DrvInfo::queryVersion() const
+{
+    if (!attrs) {
+        version = std::optional<std::string>{std::nullopt};
+    } else if (!version.has_value()) {
+        auto i = attrs->find(state->symbols.create("version"));
+        if (i != attrs->end() && i->value->type() == nString)
+            version = state->forceStringNoCtx(*i->value);
+        else
+            version = std::optional<std::string>{std::nullopt};
+    }
+    return *version;
+}
+
+
 std::string DrvInfo::querySystem() const
 {
     if (system == "" && attrs) {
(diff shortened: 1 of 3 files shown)


jtojnar:
Verified that this patch prints correct values in `nix-env -f channel:nixos-unstable-small -qa --json`:

Unstable Nix:

```
1901858:  "vapoursynth": {
1901859:    "name": "vapoursynth-R52",
1901860:    "pname": "vapoursynth-R52",
1901861-    "version": "",
```

This patch:

```
1904437:  "vapoursynth": {
1904438:    "name": "vapoursynth-R49",
1904439:    "pname": "vapoursynth",
1904440-    "version": "R49",
```

jtojnar:
Wanted to mark this as closing https://github.com/NixOS/nixos-homepage/issues/306 but that issue disappeared.

edolstra:
I'm hesitant to modify the semantics of `nix-env` since it's kinda legacy at this point. The replacement (`nix profile`) does not use the (p)name/version attributes at all.

jtojnar:
These keys were only added [recently](https://github.com/NixOS/nix/commit/cd933b22d2041b7efc348dcc09ff255967ffc663) and we still use `nix-env` output for `packages.json` which ends up being picked up by Repology.

r-burns:
Still relevant

superherointj:
> I'm hesitant to modify the semantics of `nix-env` since it's kinda legacy at this point. The replacement (`nix profile`) does not use the (p)name/version attributes at all.

@edolstra As this is still needed to correct data for Repology. Couldn't this be merged?
If not, we need some other solution because problem still persists. 
`nix search --json` also doesn't work for this case. Do you have any suggestion of solution?

superherointj:
In case this PR is not accepted (which seems so?).

I'm seeking a replacement for this:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/make-tarball.nix#L67

How to use `nix profile` to replace `nix-env` for this use case?

With `nix-env` I'd list packages this way:
`nix-env --query --available --meta --json`

What would be the equivalent using `nix profile`?
`nix profile list` is not listing for me (wonder why)

Ericson2314:
It might be prudent to generate the repology info in a way that is agnostic to `nix-env` *and* `nix-profile`.

tomberek:

superherointj:
Likely the issue is still valid and no one bothered to look into.
There was a [repology issue](https://github.com/NixOS/nix/issues/3868#issuecomment-1109683184), but don't ask me the details, it's been so long and so many PRs/issues. :D

Issue #4075, opened 2020-09-26
Binary cache upload optimization


roberth:
**Is your feature request related to a problem? Please describe.**

Binary cache uploads can take longer than necessary.

**Describe the solution you'd like**

I'd like to avoid compressing the NAR before determining that it's already available in the cache.
This can be done by naming the NAR file after the *uncompressed* hash.
It has further benefits for the uploading implementation, which is that when the ValidPathInfo is already known (currently always the case), we can compress and upload simultaneously rather than sequentially, saving time even in the case where we do actually need to compress and upload.

**Describe alternatives you've considered**

Do we have a good reason to name the NAR file after its compressed hash? All I can think of is that it simplifies a possible integrity check, but that check can still be done by reading the narinfo files. In fact you probably want to read those anyway so you can garbage collect any incomplete uploads that didn't make it to the narinfo upload step.

**Additional context**

 - Somewhat off-topic [comment](https://github.com/NixOS/nix/pull/3935/files#r495441052) for https://github.com/NixOS/nix/pull/3935 discussing how it relates to a possible `Store` interface refactoring

zimbatm:
Assuming a compromised cache, uncompressing before verification also has the potential of including zip-bomb equivalents that explode disk usage. I just wanted to reply to the challenge and don't think that it's a significant-enough threat to prevent this change :D

roberth:
@zimbatm that's an interesting thought though!

> Assuming a compromised cache

Normally I'd say you have bigger problems in that case, but perhaps it's useful to have an _untrusted_ cache for content-addressable stuff like sources. So that would be a cache where you don't configure a public key and rely on CA hashes for verification.

But even then, before unpacking we will have queried the narinfo, telling us the size, that we can then use to abort decompression if necessary. Or of course even earlier if we decide that the recorded size is too big in the first place.

Currently Nix doesn't seem to have any protection against large NARs, so it's kind of an orthogonal issue.

Ericson2314:
Hehe I hadn't heard about zip bombs, how fun! I think the risk is less for us because we are compressing whole archives rather than individual files? When we decompress, we can just bail out if it exceeds the nar size also specified in the narinfo file.

zimbatm:
Oops, I didn't mean to start a thread, please ignore my comment :D

I think that this issue is a good idea.

roberth:
Perhaps of interest to @zhaofengli, author of [attic](https://github.com/zhaofengli/attic).

PR #4464, opened 2021-01-18
narinfo: Change NAR URLs to be addressed on the NAR hash instead of the compressed hash

adisbladis:
This change is to simplify [Trustix](https://github.com/tweag/trustix) indexing and makes it possible to reconstruct this URL regardless of the compression used.

In particular this means that https://github.com/tweag/trustix/blob/7c2e9ca597de233846e0b265fb081626ca6c59d8/contrib/nix/nar/nar.go#L61-L71 can be removed and only the bits that are required to establish trust needs to be published in the Trustix build logs.

A notable change is that the compressed files no longer comes with a file extension indicating their compression. This should be fine as the associated narinfo file has the compression metadata.
For applications where this is not feasible (like Trustix) the application could check the compression by the magic numbers.

I have manually tested if my local version of Nix (2.3.10) can still substitute from a binary cache created after this change and everything still works.

### Related issues
- https://github.com/NixOS/nix/issues/4075

### Acknowledgements
This work is done within the context of [Trustix](https://nlnet.nl/project/Trustix/) and has been sponsored by a [NLNet Foundation NGI-0 PET](https://nlnet.nl/thema/NGIZeroPET.html) grant.

cc @Ericson2314 @domenkozar

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -176,11 +176,7 @@ ref<const ValidPathInfo> BinaryCacheStore::addToStoreCommon(
     auto [fileHash, fileSize] = fileHashSink.finish();
     narInfo->fileHash = fileHash;
     narInfo->fileSize = fileSize;
-    narInfo->url = "nar/" + narInfo->fileHash->to_string(Base32, false) + ".nar"
-        + (compression == "xz" ? ".xz" :
-           compression == "bzip2" ? ".bz2" :
-           compression == "br" ? ".br" :
-           "");
+    narInfo->url = "nar/" + info.narHash.to_string(Base32, false) + ".nar";
 
     auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(now2 - now1).count();
     printMsg(lvlTalkative, "copying path '%1%' (%2% bytes, compressed %3$.1f%% in %4% ms) to binary cache",
--- a/tests/binary-cache.sh
+++ b/tests/binary-cache.sh
@@ -55,7 +55,7 @@ basicTests
 # Test whether Nix notices if the NAR doesn't match the hash in the NAR info.
 clearStore
 
-nar=$(ls $cacheDir/nar/*.nar.xz | head -n1)
+nar=$(ls $cacheDir/nar/*.nar | head -n1)
 mv $nar $nar.good
 mkdir -p $TEST_ROOT/empty
 nix-store --dump $TEST_ROOT/empty | xz > $nar


zimbatm:
Nice. Did you check what content-types were being used on upload to S3?

adisbladis:
> Nice. Did you check what content-types were being used on upload to S3?

The content type is always set to `application/x-nix-nar` regardless of the compression used.
The compression can either be fetched from the narinfo or inferred by looking at the magic numbers.

edolstra:
The problem with this scheme is that it makes the files in `nar/` no longer content-addressed (i.e. they're addressed by the uncompressed contents rather than their actual contents). One result is that if you re-upload a NAR with a different compression algorithm or settings, it can overwrite a previously existing NAR. Since `.narinfo` files have a `FileHash` field that specifies the hash of the compressed file, this invalidates the old `.narinfo` files. So we would have to get rid of the `FileHash`, `FileSize` and `Compression` fields. (It could be that the current implementation no longer checks `FileHash`, but `Compression` would still be a problem.)

adisbladis:
> The problem with this scheme is that it makes the files in nar/ no longer content-addressed (i.e. they're addressed by the uncompressed contents rather than their actual contents). 

I don't think this is semantically meaningful?

> One result is that if you re-upload a NAR with a different compression algorithm or settings, it can overwrite a previously existing NAR.
> Since .narinfo files have a FileHash field that specifies the hash of the compressed file, this invalidates the old .narinfo files.

Afaik Nix checks for the remote store narinfo before copying. If so this should be a no-op on existing narinfos.

> So we would have to get rid of the FileHash, FileSize and Compression fields. (It could be that the current implementation no longer checks FileHash, but Compression would still be a problem.)

I'm not proposing getting rid of these entirely, at least not in this iteration.
What I want to do is to be able to ignore them and uncompress NARs in a binary cache proxy, setting these fields:
```
FileHash: NarHash
FileSize: FileHash
Compression: none
```

grahamc:
I'm not keen on this, the file being named for what it is makes it easy to tell if the file is valid from an operational point of view. Additionally, I'm not sure it is a good idea: the structure presented in Nix is just one possible layout of NARs in the cache. It is very easy to imagine a binary cache which segments its data in a `nar/68/sm/67/lc/d4/68sm67lcd4pnmyhijpyh134a7ykgyjhq...` structure. The narinfo file gives cache operators this flexibility, and I'm not keen on Trustix solidifying it in stone.

edolstra:
I just had a discussion with @adisbladis about this and decided to merge. Since Trustix stores trust mappings from store paths to NAR hashes, there is no way to get the URL of the compressed file in arbitrary binary caches if it's not named after the NAR hash. 

The main remaining issue is that we need to start ignoring the Compression field but we can do that in a future PR.

grahamc:
I feel this is a big loss, but okay. Does this mean binary caches which don't store this way can't participate in trustix?

adisbladis:
> Does this mean binary caches which don't store this way can't participate in trustix

You could participate purely to track reproducibility.

Without a canonical address to access the NARs it's not possible for a binary cache proxy to serve NARs from multiple caches as compression schemes may differ, and therefore URLs are unpredictable, even though the actual unpacked contents are identical.

This isn't just an issue for Trustix, but for anyone who wants to build a highly available binary cache proxy substituting from multiple upstreams.

grahamc:
My main concerns here are how quickly the Nix codebase adopted this change, and how little discussion there was in how this could impact other use cases. This could easily spread as an assumption in other parts of the codebase, breaking use cases that are important to me. I'm not interested in building on sand, and I feel like my world has become sand.

domenkozar:
I want to raise a few problems with this:

1) backwards compatibility becomes harder, since all code now needs to handle both cases

2) it's not possible to have two compressions for the same nar inside the binary cache anymore (because they are now pre-compression hash addressed  and contain no extension)

3) it's now expensive (decompress) to verify hash of nar file

Not saying that those are blockers, but they do affect our future decisions.

adisbladis:
To elaborate a bit further:

When a request for a NAR comes in, and that URL is not strictly reproducible our cache proxy would have to store a substantial amount of state to be able to go from the non-reproducible URL back to the store path prefix so we can actively poll known binary caches for this content.
I'm not even sure this is practically feasible.

grahamc:
Additionally:

* it is very convenient to have builders upload directly to the cache using pre-signed S3 tokens which authorize writes to a single content-addressed URL using `s3:x-amz-content-sha256`
* this puts another road block in to having NARs located on separate domains, which is something I've wanted to have support for
* it implies cache content can never move, and that narinfos are never updated

I would much prefer binary cache proxies follow what I feel is the spec,: query the upstream cache for the narinfo to locate the nar, instead of trying to play games reconstructing URLs.

grahamc:
> substantial amount of state to be able to go from the non-reproducible URL back to the store path prefix so we can actively poll known binary caches

You only need to know the derivation's hash, since that is how you look up the narinfo to find the nar... which is how binary caches work.

adisbladis:
> I would much prefer binary cache proxies follow what I feel is the spec,: query the upstream cache for the narinfo to locate the nar, instead of trying to play games reconstructing URLs.

That's not practical at all as I've already explained because of the massive amounts of state required.
Once you've served up the narinfo and the client has cached that NAR URL you cannot do high availability across caches.

> this puts another road block in to having NARs located on separate domains, which is something I've wanted to have support for

You could implement this with redirects, additionally this doesn't need to invalidate any client caches.

grahamc:
I don't understand why you need a lot of state to look up the narinfo file on a binary cache.

adisbladis:
> You only need to know the derivation's hash, since that is how you look up the narinfo to find the nar... which is how binary caches work.

There is an unbounded time period between when a client requests (and caches) a narinfo and when it (possibly) requests a NAR.
Caches may have uploaded the package in question in the mean time and uploaded them using a variety of compression schemes, resulting in different NAR URLs.

When the request for the NAR finally comes in we have lost the context of which narinfo this NAR is associated with, so it's not possible to poll other binary caches for the same NAR but with a different compression.
This is where we'd be required to store state mapping NAR URLs back into store path prefixes, otherwise we're lacking enough metadata.

rbvermaa:
It looks to me that this change might be a bit more controversial was orginally thought, given the activity post-merge on this PR. 

Given that @grahamc and @domenkozar have both expressed (some of) their hesitations/problems with the change, would it be perhaps a good idea to revert this change for now and start an RFC for this?

grahamc:
I would like for that to happen, @rbvermaa -- changing the semantics and expectations of the binary cache like this seems significant enough to warrant an RFC. To that end, I've opened a revert PR: https://github.com/NixOS/nix/pull/4535

rbvermaa:
> It looks to me that this change might be a bit more controversial was orginally thought, given the activity post-merge on this PR.
> 
> Given that @grahamc and @domenkozar have both expressed (some of) their hesitations/problems with the change, would it be perhaps a good idea to revert this change for now and start an RFC for this?

@adisbladis I noticed you reacted with a thumbs down, which means you do not agree with what I suggest. Could you please elaborate on why you do not agree? Do you think the change is not controversial, or perhaps you have another reason?

@edolstra Would you mind commenting on your view?

adisbladis:
> @adisbladis I noticed you reacted with a thumbs down, which means you do not agree with what I suggest. Could you please elaborate on why you do not agree? Do you think the change is not controversial, or perhaps you have another reason?

Going straight from "needs more discussion" to "write an RFC" is extreme and an excellent way to quash dissenting view points as it's puts an unreasonable burden on the author to push their work forward.

I think reverting with the reason "needs more discussion" is annoying but within the reasonable realm of responses, going straight for an RFC isn't.

At this moment (this is just a quick summary, any values should be seen as rough estimates) the average time to get an RFC merged is 113 days and a median of 48 days.
This time is obviously not including the time to actually write an RFC.
That's a significant hurdle only a handful of people are willing to go through, out of the currently 22 merged RFCs there were only 13 authors.

grahamc:
I don't follow the reasoning. RFCs are to explore the design space and come up with a great solution. I don't see it being extreme, but I suppose I also don't see the RFC process as punishment. I think big changes should get proper consideration and time for the large community of users to contribute to it. I would hope that we as a project can follow the process we have agreed to for exploring significant changes.

rbvermaa:
> > @adisbladis I noticed you reacted with a thumbs down, which means you do not agree with what I suggest. Could you please elaborate on why you do not agree? Do you think the change is not controversial, or perhaps you have another reason?
> 
> Going straight from "needs more discussion" to "write an RFC" is extreme and an excellent way to quash dissenting view points as it's puts an unreasonable burden on the author to push their work forward.
> 
> At this moment (this is just a quick summary, any values should be seen as rough estimates) the average time to get an RFC merged is 113 days and a median of 48 days.
> This time is obviously not including the time to actually write an RFC.
> That's a significant hurdle only a handful of people are willing to go through, out of the currently 22 merged RFCs there were only 13 authors.
> 
> I think reverting with the reason "needs more discussion" and report an issue is annoying but within the reasonable realm of responses, going straight for an RFC isn't.
> If it turns out an RFC is needed after some discussion, then so be it.

Perhaps I should have suggested two things in stead:

* revert the change so that more discussion can happen, as the change was more controversial than originally anticipated
* ask for a discussion whether an RFC is needed for the change during the previously mentioned discussion

@adisbladis Does that make sense?

adisbladis:
> Perhaps I should have suggested two things in stead: …

This sounds like a much more reasonable path forwards.

edolstra:
Okay, let's revert until after the 2.4 branch-off for now. We can continue to discuss it here.

PR #4467, opened 2021-01-21
Improve error formatting

edolstra:
Changes:

* The divider lines are gone. These were in practice a bit confusing, in particular with `--show-trace` or `--keep-going`, since then there were multiple lines, suggesting a start/end which wasn't the case.

* Instead, multi-line error messages are now indented to align with the prefix (e.g. `error: `).

* The `description` field is gone since we weren't really using it.

* `hint` is renamed to `msg` since it really wasn't a hint.

* The error is now printed *before* the location info.

* The `name` field is no longer printed since most of the time it wasn't very useful since it was just the name of the exception (like `EvalError`). Ideally in the future this would be a unique, easily googleable error ID (like rustc).

* `trace:` is now just `…`. This assumes error contexts start with something like "while doing X".

* Error locations are printed in a more standard format (`filename:line:column`).

Example before:

```
error: --- AssertionError ---------------------------------------------------------------------------------------- nix
at: (7:7) in file: /home/eelco/Dev/nixpkgs/pkgs/applications/misc/hello/default.nix

     6|
     7|   x = assert false; 1;
      |       ^
     8|

assertion 'false' failed
----------------------------------------------------- show-trace -----------------------------------------------------
trace: while evaluating the attribute 'x' of the derivation 'hello-2.10'
at: (192:11) in file: /home/eelco/Dev/nixpkgs/pkgs/stdenv/generic/make-derivation.nix

   191|         // (lib.optionalAttrs (!(attrs ? name) && attrs ? pname && attrs ? version)) {
   192|           name = "${attrs.pname}-${attrs.version}";
      |           ^
   193|         } // (lib.optionalAttrs (stdenv.hostPlatform != stdenv.buildPlatform && !dontAddHostSuffix && (attrs ? name || (attrs ? pname && attrs ? version)))) {
```

Example after:

```
error: assertion 'false' failed

       at /home/eelco/Dev/nixpkgs/pkgs/applications/misc/hello/default.nix:7:7:

            6|
            7|   x = assert false; 1;
             |       ^
            8|

       … while evaluating the attribute 'x' of the derivation 'hello-2.10'

       at /home/eelco/Dev/nixpkgs/pkgs/stdenv/generic/make-derivation.nix:192:11:

          191|         // (lib.optionalAttrs (!(attrs ? name) && attrs ? pname && attrs ? version)) {
          192|           name = "${attrs.pname}-${attrs.version}";
             |           ^
          193|         } // (lib.optionalAttrs (stdenv.hostPlatform != stdenv.buildPlatform && !dontAddHostSuffix && (attrs ? name || (attrs ? pname && attrs ? version)))) {
```

@garbas

### DIFF
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -172,13 +172,14 @@ static int main_build_remote(int argc, char * * argv)
                     else
                     {
                         // build the hint template.
-                        string hintstring =  "derivation: %s\nrequired (system, features): (%s, %s)";
-                        hintstring += "\n%s available machines:";
-                        hintstring += "\n(systems, maxjobs, supportedFeatures, mandatoryFeatures)";
+                        string errorText =
+                            "Failed to find a machine for remote build!\n"
+                            "derivation: %s\nrequired (system, features): (%s, %s)";
+                        errorText += "\n%s available machines:";
+                        errorText += "\n(systems, maxjobs, supportedFeatures, mandatoryFeatures)";
 
-                        for (unsigned int i = 0; i < machines.size(); ++i) {
-                          hintstring += "\n(%s, %s, %s, %s)";
-                        }
+                        for (unsigned int i = 0; i < machines.size(); ++i)
+                            errorText += "\n(%s, %s, %s, %s)";
 
                         // add the template values.
                         string drvstr;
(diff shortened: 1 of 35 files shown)


domenkozar:
cc @bburdette

bburdette:
I for one will miss the divider, although the show-trace one was maybe a bit much.  What about program name?  I like seeing that in gists of errors and whatnot, takes one question off the table right away.  In a future time where the only command is nix it won't be relevant, but I feel like its nice to have for now.

Probably hintfmt should be renamed too?  But not sure what the name should be.  msgfmt is pretty generic.  Something like colorfmt would be more accurate.

edolstra:
Yeah the program name is useful (especially to see whether the error came from the daemon) but it won't work anymore once everything is called "nix". We probably should add something to the trace to indicate that the error came from the daemon.

`colorfmt` or `colorize` would be a good name.

PR #4468, opened 2021-01-21
scripts/install-nix-from-closure: only show progress if a terminal is used

Ma27:
While the progress dots during the copying of the store work fine on a
normal terminal, those look pretty off if the script is run inside a
provisioning script of e.g. `vagrant` or `packer` where `stderr` and
`stdout` are captured:

    default: .
    default: ..
    default: .
    default: .
    default: .

To work around this, the script checks with `-t 0` if it's
running on an actual terminal and doesn't show the progress if that's not
the case.

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -166,9 +166,15 @@ fi
 mkdir -p $dest/store
 
 printf "copying Nix to %s..." "${dest}/store" >&2
+# Insert a newline if no progress is shown.
+if [ ! -t 0 ]; then
+  echo ""
+fi
 
 for i in $(cd "$self/store" >/dev/null && echo ./*); do
-    printf "." >&2
+    if [ -t 0 ]; then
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Would it be reasonable to detect if the command is not run in an interactive shell and then activate this automatically?

edolstra:
I agree it would probably be better to just check for a terminal (`if [ -t 0 ]; ...`) and disable the progress indicator automatically.

Ma27:
@edolstra @SuperSandro2000 updated the PR :)

Ma27:
cc @edolstra anything else to do ? :)

Ma27:
@edolstra is there any chance to include this in the next 2.3.x if another one is planned? This is for at least one project I'm working on an improvement as this would make the provisioning output readable :)

edolstra:
Yes, I've just cherry-picked it to 2.3 branch.

PR #4477, opened 2021-01-26
Build ca derivations remotely

regnat:
Allow using the build hook to build ca derivations remotely.

Required extending the remote build protocol(s) a bit to send back the newly built outputs.

### DIFF
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -311,6 +311,13 @@ connected:
                     localStore->locksHeld.insert(store->printStorePath(path)); /* FIXME: ugly */
             copyPaths(ref<Store>(sshStore), store, missingPaths, NoRepair, NoCheckSigs, NoSubstitute);
         }
+        // XXX: Should e done as part of `copyPaths`
+        for (auto & realisation : missingRealisations) {
+            // Should hold, because if the feature isn't enabled the set
+            // of missing realisations should be empty
+            settings.requireExperimentalFeature("ca-derivations");
+            store->registerDrvOutput(realisation);
+        }
 
         return 0;
     }
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -248,7 +248,7 @@ connected:
         std::cerr << "# accept\n" << storeUri << "\n";
 
         auto inputs = readStrings<PathSet>(source);
-        auto outputs = readStrings<PathSet>(source);
+        auto wantedOutputs = readStrings<StringSet>(source);
 
         AutoCloseFD uploadLock = openLockFile(currentLoad + "/" + escapeUri(storeUri) + ".upload-lock", true);
 
@@ -273,6 +273,7 @@ connected:
         uploadLock = -1;
 
         auto drv = store->readDerivation(*drvPath);
+        auto outputHashes = staticOutputHashes(*store, drv);
         drv.inputSrcs = store->parseStorePathSet(inputs);
 
         auto result = sshStore->buildDerivation(*drvPath, drv);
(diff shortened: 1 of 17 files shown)


regnat:
Rebased on top of master to fix some conflicts

src/libstore/build/entry-points.cc
```diff
@@ -58,6 +58,27 @@ BuildResult Store::buildDerivation(const StorePath & drvPath, const BasicDerivat
         result.status = BuildResult::MiscFailure;
         result.errorMsg = e.msg();
     }
+    // XXX: Should use something like
+    // `queryPartialDerivationOutputMap`, but we can't do that
+    // as it assumes a registered derivation
```
Ericson2314, line 63: There is a `DerivationGoal::queryPartialDerivationOutputMap` for this purpose that you might be able to somehow use?

regnat: Mh I might indeed. Will need to update it a bit to handle non-static output paths, but I think it's a good thing anyways. Thanks :)

regnat: At a second though it can't, because `queryPartialDerivationOutputMap` only returns the output paths and not the full realisations. So we'll have to update it first.

I've updated the comment to reflect this

Ericson2314: OK sounds good.

tests/build-remote-content-addressed-fixed.sh
```diff
@@ -1,5 +0,0 @@
-source common.sh
```
regnat, line 1: This shouldn't be deleted (but fixed in a separate PR)

regnat:
Rebased on top of latest master

src/build-remote/build-remote.cc
```diff
@@ -273,23 +273,50 @@ static int main_build_remote(int argc, char * * argv)
...
 
         if (!result.success())
             throw Error("build of '%s' on '%s' failed: %s", store->printStorePath(*drvPath), storeUri, result.errorMsg);
 
-        StorePathSet missing;
-        for (auto & path : outputs)
-            if (!store->isValidPath(store->parseStorePath(path))) missing.insert(store->parseStorePath(path));
+        std::set<Realisation> missingRealisations;
+        StorePathSet missingPaths;
+        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
```
edolstra, line 286: Maybe this should check whether `drv` is a CA derivation? Since we don't have to do the CA-related stuff (missingRealisations etc.) if it isn't.

src/build-remote/build-remote.cc
```diff
@@ -273,23 +273,50 @@ static int main_build_remote(int argc, char * * argv)
...
+            for (auto & outputName : wantedOutputs) {
+                auto thisOutputHash = outputHashes.at(outputName);
+                auto thisOutputId = DrvOutput{ thisOutputHash, outputName };
+                if (!store->queryRealisation(thisOutputId)) {
+                    notice("Missing output %s", outputName);
```
edolstra, line 291:
```suggestion
                    debug("missing output %s", outputName);
```

src/build-remote/build-remote.cc
```diff
@@ -273,23 +273,50 @@ static int main_build_remote(int argc, char * * argv)
...
             Activity act(*logger, lvlTalkative, actUnknown, fmt("copying outputs from '%s'", storeUri));
             if (auto localStore = store.dynamic_pointer_cast<LocalStore>())
-                for (auto & i : missing)
-                    localStore->locksHeld.insert(store->printStorePath(i)); /* FIXME: ugly */
-            copyPaths(ref<Store>(sshStore), store, missing, NoRepair, NoCheckSigs, NoSubstitute);
+                for (auto & path : missingPaths)
+                    localStore->locksHeld.insert(store->printStorePath(path)); /* FIXME: ugly */
+            copyPaths(ref<Store>(sshStore), store, missingPaths, NoRepair, NoCheckSigs, NoSubstitute);
+        }
+        // XXX: Should e done as part of `copyPaths`
```
edolstra, line 314:
```suggestion
        // XXX: Should be done as part of `copyPaths`
```

src/libstore/build/derivation-goal.cc
```diff
@@ -1159,13 +1159,14 @@ HookReply DerivationGoal::tryBuildHook()
...
-        for (auto & [_, status] : initialOutputs) {
-            if (!status.known) continue;
-            if (buildMode != bmCheck && status.known->isValid()) continue;
-            missingPaths.insert(status.known->path);
+        StringSet missingOutputs;
+        for (auto & [outputName, status] : initialOutputs) {
+            // XXX: Does this include known CA outputs?
+            if (buildMode != bmCheck && status.known && status.known->isValid()) continue;
+            missingOutputs.insert(outputName);
+            /* missingPaths.insert(status.known->path); */
```
edolstra, line 1167:
```suggestion
```

src/libstore/build/entry-points.cc
```diff
@@ -58,6 +58,26 @@ BuildResult Store::buildDerivation(const StorePath & drvPath, const BasicDerivat
...
+    auto staticDrvOutputs = drv.outputsAndOptPaths(*this);
+    auto outputHashes = staticOutputHashes(*this, drv);
+    for (auto & [outputName, staticOutput] : staticDrvOutputs) {
+        auto outputId = DrvOutput{outputHashes.at(outputName), outputName};
+        if (staticOutput.second)
+            result.builtOutputs.insert_or_assign(
+                    outputId,
+                    Realisation{ outputId, *staticOutput.second}
+                    );
+        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
```
edolstra, line 72: Likewise this should probably be conditional on drv being a CA derivation instead.

PR #4479, opened 2021-01-26
Command plugins

shlevy:
This makes it possible to define and use external plugins that add new `nix` subcommands.


src/libstore/globals.cc
```diff
@@ -237,8 +237,24 @@ void MaxBuildJobsSetting::set(const std::string & str, bool append)
...
+       command foo, with flag bar that takes one argument. How could
+       we know that nix foo --bar --plugin-files xyz needs to pass
+       --plugin-files as an argument to --bar, not as a top-level
+       flag, before plugins are loaded? */
+    for (auto pos = cmdline.begin(); pos != cmdline.end(); ++pos) {
+        auto arg = *pos;
+        if (arg == "--")
+            break;
+        else if (arg == "--option") {
+            if (*(++pos) == "plugin-files")
```
shlevy, line 253: Oops, segfault here, fixing.

shlevy: Fixed

src/libstore/globals.cc
```diff
@@ -237,8 +237,32 @@ void MaxBuildJobsSetting::set(const std::string & str, bool append)
...
-void initPlugins()
+void initPlugins(const Strings & cmdline)
 {
+    /* !!! This is not quite correct, but while it could be better
+       some error is unavoidable. Consider a plugin that adds a
+       command foo, with flag bar that takes one argument. How could
+       we know that nix foo --bar --plugin-files xyz needs to pass
+       --plugin-files as an argument to --bar, not as a top-level
+       flag, before plugins are loaded? */
+    for (auto pos = cmdline.begin(); pos != cmdline.end(); ++pos) {
```
edolstra, line 248: Can we avoid having multiple command line parsers? It sounds like a recipe for subtle bugs where the two parsers interpret the command line in different ways.

zimbatm: Git solves this by adding two levels of option parsing. `git [global options] <command> <args...>`

So for example, if you want to change directory you would do `git -C ../other_dir commit -a`. `-a` is not parsed by the `git` dispatcher and it just forwarded to the `commit` command. It's then up to the command to do its own option parsing.

shlevy: I'm not sure how we can without breaking compatibility. If nix worked the way git does, then yes this would be fine, but currently after the subcommand is given flags are first passed to the subcommand and then to the higher level parser if unknown.

The alternatives I can think of off-hand:

1. Simply ignore `--plugin-files`/`--option plugin-files` altogether on the command line
2. Require `--plugin-files`/`--option plugin-files` to come before the first non-flag argument, and maybe deprecate the subcommand unknown flag bubble-up so we can eventually reach the two-level behavior git has
3. Special-case missing subcommand detection: when we have an unknown subcommand, fall back on the behavior here for the rest of the args list and then re-parse after plugin init.

1 and 2 are technically breaking changes to the CLI, but I doubt there's enough usage nor enough expectation of stability around the plugin interface to be a big enough deal.

Currently I'm leaning toward 2.

shlevy:
Moved the libnixcmd part to #4483 since that seems more clearly an appropriate implementation.

edolstra:
Closed via #4486.

PR #4480, opened 2021-01-26
Command plugins 2.3

shlevy:
This makes it possible to define and use external plugins that add new `nix` subcommands.

Backport of #4479


edolstra:
I think we generally shouldn't add features to the stable branch, especially potentially intrusive/risky features like this one (which introduces a new library and changes command line processing).

shlevy:
Is there a timeline for a 2.4 release?

PR #4483, opened 2021-01-27
Move command plugin interface to libnixcmd

shlevy:


Mic92:
By the way I see this error also quite often with nixUnstable: in my own CI: https://github.com/NixOS/nix/runs/1778184963#step:4:1501

PR #4486, opened 2021-01-28
Command plugins

shlevy:
This makes it possible to define and use external plugins that add new nix subcommands.

### DIFF
--- a/doc/manual/src/release-notes/rl-2.4.md
+++ b/doc/manual/src/release-notes/rl-2.4.md
@@ -1,0 +1,7 @@
+# Release 2.4 (202X-XX-XX)
+
+  - It is now an error to modify the `plugin-files` setting via a
+    command-line flag that appears after the first non-flag argument
+    to any command, including a subcommand to `nix`. For example,
+    `nix-instantiate default.nix --plugin-files ""` must now become
+    `nix-instantiate --plugin-files "" default.nix`.
--- a/doc/manual/src/release-notes/rl-2.4.md
+++ b/doc/manual/src/release-notes/rl-2.4.md
@@ -5,3 +5,4 @@
     to any command, including a subcommand to `nix`. For example,
     `nix-instantiate default.nix --plugin-files ""` must now become
     `nix-instantiate --plugin-files "" default.nix`.
+  - Plugins that add new `nix` subcommands are now actually respected.
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -60,6 +60,7 @@ void Args::parseCmdline(const Strings & _cmdline)
         verbosity = lvlError;
     }
 
+    bool argsSeen = false;
     for (auto pos = cmdline.begin(); pos != cmdline.end(); ) {
 
         auto arg = *pos;
@@ -88,6 +89,10 @@ void Args::parseCmdline(const Strings & _cmdline)
                 throw UsageError("unrecognised flag '%1%'", arg);
         }
         else {
+            if (!argsSeen) {
+                argsSeen = true;
+                initialFlagsProcessed();
+            }
             pos = rewriteArgs(cmdline, pos);
             pendingArgs.push_back(*pos++);
             if (processArgs(pendingArgs, false))
@@ -96,6 +101,9 @@ void Args::parseCmdline(const Strings & _cmdline)
     }
 
     processArgs(pendingArgs, true);
+
+    if (!argsSeen)
+        initialFlagsProcessed();
 }
 
 bool Args::processFlag(Strings::iterator & pos, Strings::iterator end)
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -306,8 +306,8 @@ Strings argvToStrings(int argc, char * * argv)
     return args;
 }
 
-MultiCommand::MultiCommand(const Commands & commands)
-    : commands(commands)
+MultiCommand::MultiCommand(const Commands & commands_)
+    : commands(commands_)
 {
     expectArgs({
         .label = "subcommand",
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -53,6 +53,9 @@ static int main_build_remote(int argc, char * * argv)
         unsetenv("DISPLAY");
         unsetenv("SSH_ASKPASS");
 
+        /* If we ever use the common args framework, make sure to
+           remove initPlugins below and initialize settings first.
+        */
         if (argc != 2)
             throw UsageError("called without required arguments");
 
--- a/src/libmain/common-args.cc
+++ b/src/libmain/common-args.cc
@@ -79,4 +79,11 @@ MixCommonArgs::MixCommonArgs(const string & programName)
     hiddenCategories.insert(cat);
 }
 
+void MixCommonArgs::initialFlagsProcessed()
+{
+    initPlugins();
+    pluginsInited();
+}
+
+
 }
--- a/src/libmain/common-args.hh
+++ b/src/libmain/common-args.hh
@@ -7,10 +7,14 @@ namespace nix {
 //static constexpr auto commonArgsCategory = "Miscellaneous common options";
 static constexpr auto loggingCategory = "Logging-related options";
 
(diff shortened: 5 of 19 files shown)


shlevy:
Alternative implementation of #4479

shlevy:
@edolstra Thoughts on this one?

shlevy:
@edolstra Ping.

shlevy:
@edolstra if there are no objections I'll merge this soon.

edolstra:
Please not, I'm not so happy about making it impossible to pass `--option` after the subcommand. That's suboptimal UX.

shlevy:
@edolstra It's not impossible, it's just a warning. The only option that can't be passed after the subcommand is plugin-files. We can remove the deprecation warning too.

shlevy:
@edolstra Can we set up some time next week to talk through next steps? Maybe Monday in the EST morning?

flokli:
Is there any further read about this feature in general, about what plugins are supposed to do/access, and what not?

I'm somewhat worried exposing the somewhat in flux internal Nix API, which was never stable or formalized to plugins either will lead to not being able to change internal API (cause people are afraid to break plugins), or changes in Nix suddenly causing breakages (cause there is no contract about what's stable).

shlevy:
@flokli plugins predate this work, and in fact we thought command plugins already worked: https://nixos.org/manual/nix/stable/#conf-plugin-files

There is no expectation of stability around these APIs, or there shouldn't be.

This PR simply fixes when command plugins are loaded so they can take effect.

flokli:
> There is no expectation of stability around these APIs, or there shouldn't be.

@shlevy thanks for the clarification. The section in the manual isn't clear enough for my taste regarding stability guarantees, but that's probably out of scope for this PR.

shlevy:
@flokli You're right, fixed https://github.com/NixOS/nix/commit/35205e2e922952fc0654260a07fc3191c5afc2cc

flokli:
Thanks!

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -0,0 +1,13 @@
+# Release 2.4 (202X-XX-XX)
+
+  - When using the `nix` command, it is now deprecated to pass
+    top-level flags (e.g. `--option`) after the subcommand. For
+    example, `nix show-config --option foo bar` should instead be
+    written `nix --option foo bar show-config`.
```
edolstra, line 6:
```suggestion
```

If this only affects plugin loading (which is undocumented anyway) then there is no need to call this deprecated since it works fine for every other setting.

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -0,0 +1,13 @@
...
+  - It is now an error to modify the `plugin-files` setting via a
+    command-line flag that appears after the first non-flag argument
+    to any command, including a subcommand to `nix`. For example,
+    `nix-instantiate default.nix --plugin-files ""` must now become
+    `nix-instantiate --plugin-files "" default.nix`.
+  - Plugins calling `registerCommand` in a static initializer will
+    now have their new commands respected.
```
edolstra, line 13:
```suggestion
  - Plugins can now add new `nix` subcommands.
```

src/libutil/args.cc
```diff
@@ -320,8 +328,13 @@ MultiCommand::MultiCommand(const Commands & commands)
 
 bool MultiCommand::processFlag(Strings::iterator & pos, Strings::iterator end)
 {
-    if (Args::processFlag(pos, end)) return true;
     if (command && command->second->processFlag(pos, end)) return true;
+    auto flag = *pos;
+    if (Args::processFlag(pos, end)) {
+        if (command)
+            warn("top-level flag '%s' passed after subcommand '%s'; this is deprecated", flag, command->first);
```
edolstra, line 335: This should not be deprecated. Probably it should only be a warning/error for `plugin-files`.

doc/manual/local.mk
```diff
@@ -20,7 +20,7 @@ dummy-env = env -i \
 	NIX_SSL_CERT_FILE=/dummy/no-ca-bundle.crt \
 	NIX_STATE_DIR=/dummy
 
-nix-eval = $(dummy-env) $(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
+nix-eval = $(dummy-env) $(bindir)/nix --experimental-features nix-command --store dummy:// eval -I nix/corepkgs=corepkgs --impure --raw
```
edolstra, line 23:
```suggestion
nix-eval = $(dummy-env) $(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
```

doc/manual/local.mk
```diff
@@ -58,7 +58,7 @@ $(d)/nix.json: $(bindir)/nix
 	@mv $@.tmp $@
 
 $(d)/conf-file.json: $(bindir)/nix
-	$(trace-gen) $(dummy-env) $(bindir)/nix show-config --json --experimental-features nix-command > $@.tmp
+	$(trace-gen) $(dummy-env) $(bindir)/nix --experimental-features nix-command show-config --json > $@.tmp
```
edolstra, line 61:
```suggestion
	$(trace-gen) $(dummy-env) $(bindir)/nix show-config --json --experimental-features nix-command > $@.tmp
```

shlevy:
@edolstra Updated with removal of the deprecation.

doc/manual/src/release-notes/rl-2.4.md
shlevy: Deprecation removed

doc/manual/src/release-notes/rl-2.4.md
shlevy: Entry adjusted in latest push.

src/libutil/args.cc
shlevy: Deprecation removed

doc/manual/local.mk
shlevy: This change was removed when removing the deprecation.

doc/manual/local.mk
shlevy: This change was removed when removing the deprecation.

edolstra:
Thanks!

aakropotkin:
What is the correct way to load plugins to avoid the error message?
I have tried everything I can think of, and while `nix` doesn't actually crash, I'm going to have to wrap the executable in something to swallow this error message or something because I cannot get this message to go away.

These all emit the warning:
```
nix --plugin-files "$PWD/libfoo.so" CMD ...;
nix --option plugin-files "$PWD/libfoo.so" CMD ...;
nix CMD --plugin-files "$PWD/libfoo.so" ...;
nix CMD --option plugin-files "$PWD/libfoo.so" ...;
NIX_CONFIG="plugin-files = $PWD/libfoo.so" nix CMD ...;
```

This one convinced the error message was just completely bogus:
```
{
  export XDG_CONFIG_HOME="$( mktemp -d; )";
  cp -r -T -- "$HOME/.config/nix" "$XDG_CONFIG_HOME/nix";
  echo "plugin-files = $PWD/libfoo.so" >> "$XDG_CONFIG_HOME/nix/nix.conf";
  nix CMD ...;
}
```

I can file a bug report but wanted to check to make sure I wasn't just using the flags incorrectly or something.

shlevy:
@aakropotkin fixed in https://github.com/NixOS/nix/pull/7736

shlevy:
@aakropotkin Note that even without this fix nothing actually fails to work, you just get this erroneous warning message. The client does in fact load the plugin-files.

aakropotkin:
Yeah luckily the plugins still work which is the most important thing.

Out of curiosity @shlevy have you encountered crashes on exit using `allocValue` from plugins? I know you use them quite a bit and thought you might know how to deal with them. Notably these only seem to appear on Darwin. I'm guessing it's a static initialization issue, but haven't had time to do a deep debug session.

shlevy:
@aakropotkin I haven't, but my usage has been very sporadic, sorry.

PR #4487, opened 2021-01-28
Copy ca derivation outputs

regnat:
Yet another take on https://github.com/NixOS/nix/pull/4220

Depends on #4340 and #4372

(the diff includes the one of #4372 because I can only target the PR at one branch)

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -45,6 +45,16 @@ struct PluginFilesSetting : public BaseSetting<Paths>
     void set(const std::string & str, bool append = false) override;
 };
 
+/* MakeError(MissingExperimentalFeature, Error); */
+class MissingExperimentalFeature: public Error
+{
+public:
+    std::string missingFeature;
+
+    MissingExperimentalFeature(std::string feature);
+    virtual const char* sname() const override { return "MissingExperimentalFeature"; }
+};
+
 class Settings : public Config {
 
     unsigned int getDefaultCores();
@@ -632,7 +642,7 @@ public:
           is `root`.
 
           > **Warning**
-          > 
+          >
           > Adding a user to `trusted-users` is essentially equivalent to
           > giving that user root access to the system. For example, the user
           > can set `sandbox-paths` and thereby obtain read access to
@@ -722,13 +732,13 @@ public:
           The program executes with no arguments. The program's environment
           contains the following environment variables:
 
-            - `DRV_PATH`  
+            - `DRV_PATH`
               The derivation for the built paths.
 
               Example:
               `/nix/store/5nihn1a7pa8b25l9zafqaqibznlvvp3f-bash-4.4-p23.drv`
 
-            - `OUT_PATHS`  
+            - `OUT_PATHS`
               Output paths of the built derivation, separated by a space
               character.
 
@@ -759,7 +769,7 @@ public:
           documentation](https://ec.haxx.se/usingcurl-netrc.html).
 
           > **Note**
-          > 
+          >
           > This must be an absolute path, and `~` is not resolved. For
           > example, `~/.netrc` won't resolve to your home directory's
           > `.netrc`.
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -45,7 +45,6 @@ struct PluginFilesSetting : public BaseSetting<Paths>
     void set(const std::string & str, bool append = false) override;
 };
 
-/* MakeError(MissingExperimentalFeature, Error); */
 class MissingExperimentalFeature: public Error
 {
 public:
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -783,6 +783,24 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
 }
 
 
+std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore, const RealisedPath::Set & paths,
+    RepairFlag repair, CheckSigsFlag checkSigs, SubstituteFlag substitute)
+{
+    StorePathSet storePaths;
+    std::set<Realisation> realisations;
+    for (auto path : paths) {
+        storePaths.insert(path.path());
+        if (auto realisation = std::get_if<Realisation>(&path.raw))
+            realisations.insert(*realisation);
+    }
+    auto pathsMap = copyPaths(srcStore, dstStore, storePaths, repair, checkSigs, substitute);
+    for (auto& realisation : realisations) {
+        dstStore->registerDrvOutput(realisation);
+    }
+
+    return pathsMap;
+}
+
 std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore, const StorePathSet & storePaths,
     RepairFlag repair, CheckSigsFlag checkSigs, SubstituteFlag substitute)
 {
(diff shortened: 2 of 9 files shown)


src/libstore/store-api.cc
```diff
@@ -783,9 +783,16 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
 }
 
 
-std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore, const StorePathSet & storePaths,
+std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore, const RealisedPath::Set & paths,
```
Ericson2314, line 786: I think the original function is basically intact within this? I would prefer the original function stay as in, then, and a new `copyPaths` for the `RealizedPath` stuff parts call the original. Then we maintain a separation of concerns.

regnat: Oh that's very true. Esp. given that the `StorePathSet` overload is still defined just above. Fixed in b7ce59e

src/libstore/store-api.cc
```diff
@@ -783,6 +783,24 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
 }
 
 
+std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore, const RealisedPath::Set & paths,
+    RepairFlag repair, CheckSigsFlag checkSigs, SubstituteFlag substitute)
+{
```
Ericson2314, line 788: `settings.isExperimentalFeatureEnabled("ca-derivations")` somewhere in here?

Ericson2314: For the local side

src/libstore/store-api.cc
```diff
@@ -783,6 +783,24 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
...
+    StorePathSet storePaths;
+    std::set<Realisation> realisations;
+    for (auto path : paths) {
+        storePaths.insert(path.path());
+        if (auto realisation = std::get_if<Realisation>(&path.raw))
+            realisations.insert(*realisation);
+    }
+    auto pathsMap = copyPaths(srcStore, dstStore, storePaths, repair, checkSigs, substitute);
+    for (auto& realisation : realisations) {
+        dstStore->registerDrvOutput(realisation);
```
Ericson2314, line 798: separately arrange for this to fail nicely if it isn't enabled remotely

regnat: Fixed in e1c0bf9

(I made it just log the error because as we can't control the remote side, it might make sense to just copy the output paths without failing at the end)

src/libstore/store-api.cc
regnat: I think in practice this always hold (because the `RealisedPathsCommand` only uses `OpaquePath`s if the flag isn't set. But I've added a check here too in 00a30ea just for security

src/libstore/globals.hh
```diff
@@ -672,7 +682,7 @@ public:
           send a series of commands to modify various settings to stdout. The
           currently recognized commands are:
 
-            - `extra-sandbox-paths`  
+            - `extra-sandbox-paths`
```
edolstra, line 685: IIRC this trailing whitespace is significant, it causes it to be rendered as a definition list.

src/libstore/store-api.cc
```diff
@@ -783,6 +783,35 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
 }
 
 
+std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore, const RealisedPath::Set & paths,
+    RepairFlag repair, CheckSigsFlag checkSigs, SubstituteFlag substitute)
+{
+    settings.requireExperimentalFeature("ca-derivations");
+    StorePathSet storePaths;
+    std::set<Realisation> realisations;
+    for (auto path : paths) {
```
edolstra, line 792:
```suggestion
    for (auto & path : paths) {
```

src/libstore/store-api.cc
```diff
@@ -783,6 +783,35 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
...
+        storePaths.insert(path.path());
+        if (auto realisation = std::get_if<Realisation>(&path.raw))
+            realisations.insert(*realisation);
+    }
+    auto pathsMap = copyPaths(srcStore, dstStore, storePaths, repair, checkSigs, substitute);
+    try {
+        for (auto& realisation : realisations) {
```
edolstra, line 799:
```suggestion
        for (auto & realisation : realisations) {
```
Style nitpick.

src/libstore/store-api.cc
```diff
@@ -783,6 +783,35 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
...
+            dstStore->registerDrvOutput(realisation);
+        }
+    } catch (MissingExperimentalFeature & e) {
+        // Don't fail if the remote doesn't support CA derivations is it might
+        // not be whithin our control to change that, and we might still want
```
edolstra, line 804:
```suggestion
        // not be within our control to change that, and we might still want
```

src/libstore/store-api.hh
```diff
@@ -752,15 +752,12 @@ void copyStorePath(ref<Store> srcStore, ref<Store> dstStore,
    that. Returns a map of what each path was copied to the dstStore
    as. */
 std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStore,
-    const StorePathSet & storePaths,
+    const RealisedPath::Set&,
```
edolstra, line 755:
```suggestion
    const RealisedPath::Set &,
```

src/libstore/globals.hh
```diff
@@ -28,6 +28,16 @@ struct MaxBuildJobsSetting : public BaseSetting<unsigned int>
     void set(const std::string & str, bool append = false) override;
 };
 
+/* MakeError(MissingExperimentalFeature, Error); */
```
edolstra, line 31:
```suggestion
```

edolstra (approved):
LGTM apart from some comments.

I wonder though whether copyPaths should be overloaded like this. Maybe it makes more sense to separate copying paths from copying realisations. There might be cases where you just want to copy a store path closure without registering realisation mappings.

src/libstore/globals.hh
regnat: Aaaaaargh thanks. Fixed in a21de5181a7a5e6377e63c749a1d7f0758c57eb8

regnat:
> I wonder though whether copyPaths should be overloaded like this. Maybe it makes more sense to separate copying paths from copying realisations. There might be cases where you just want to copy a store path closure without registering realisation mappings.

That's a fair point.
- For `copyPaths` itself, it's an overload as you say, `copyPaths(..., StorePathsSet)` still exists (and is used in a number of places). Maybe we could rename the overload as `copyRealisations` or something like that.
- For `nix copy`, currently the only way to only copy a store path is to run `nix copy /nix/store/foobar` (as all the other forms of installable will yield a `Realisation`). Maybe we could add a new `OperateOn` enum member to operate on realisations rather than just store paths (along with a flag similar to `--derivations` to change the `OperateOn` we want to use). If that sounds good I can do that in a follow-up PR

regnat:
Rebased on top of master

PR #4494, opened 2021-01-29
Allow StorePathWithOutputs in the cli

regnat:
Allow things like `nix build /nix/store/...-foo.drv!dev`.

This is especially useful for ca derivations (because this is the
canonical way of refering to the outputs of ca derivations), but is also
generally useful for consitency − for example it means that
`nix build $(nix-instantiate ./foobar.nix)` will always work − barring
any GC interaction.

Depends on #4372

### DIFF
--- a/src/nix/installables.cc
+++ b/src/nix/installables.cc
@@ -332,30 +332,30 @@ Installable::getCursor(EvalState & state)
 struct InstallableStorePath : Installable
 {
     ref<Store> store;
-    StorePath storePath;
+    StorePathWithOutputs storePath;
 
-    InstallableStorePath(ref<Store> store, StorePath && storePath)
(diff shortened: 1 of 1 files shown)


edolstra:
We should document this syntax somewhere. Currently I don't think it is (since it was mostly a hack to get multiple outputs to work between nix-instantiate and nix-build).

edolstra:
@regnat There is a merge conflict now, can you look into that?

regnat:
Superseded by #4541

Ericson2314:
I think you meant https://github.com/NixOS/nix/pull/4543, but either way, thanks.

regnat:
> I think you meant #4543, but either way, thanks.

Yes thanks. Edited the comment to rectify

PR #4470, opened 2021-01-22
Handle missing etag in 304 Not Modified response

matthewbauer:
GitHub now omits the etag, but 304 implies it matches the one we
provided. Just use that one to avoid having an etag-less resource.

Fixes #4469

PR #4495, opened 2021-01-29
Use passthru for perl-bindings, allows Nix patching for Hydra

Infinisil:
This allows patching Nix for Hydra with additional overlays, because `.overrideAttrs` and co. will persist the passthru's.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -115,7 +115,7 @@
       # 'nix.perl-bindings' packages.
       overlay = final: prev: {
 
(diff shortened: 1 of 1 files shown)


Infinisil:
Notably this PR is based on the Nix version that Hydra has pinned in its `flake.lock`, meaning this is a version compatible with hydra (in contrast with master Nix which isn't compatible). The motivation for this is because it allows patching Nix to include https://github.com/NixOS/nix/pull/4470.

PR #4500, opened 2021-02-01
--no-net -> --offline

domenkozar:


jonringer:
This probably should have been an addition with a deprecation. The current `nixos-rebuild --flake` invocation still supplies `--no-net` to nix when executing commands.

domenkozar:
@jonringer Probably. 

`nixpkgs` shouldn't be using experimental command or it shouldn't be experimental.

jonringer:
The only way to have the `--flake` option be usable is to install the nixUnstable version

jonringer:
there is no "stable" release with flakes

domenkozar:
The solution then is to pin Nix using resholve when building nixos-rebuild, that will make sure it's clear what flag should be used.

jonringer:
I would say, that's a solution, but less than ideal.

The use case was using `nixos-rebuild --flake` to bootstrap a machine, and so introducing less imperative steps is a significant improvement to maintainability.

domenkozar:
@jonringer wouldn't that make it work (unless you override nixUnstable)?

jonringer:
we are using `nixos-rebuild --target ssh://<ip> --flake <url>`. So it's targeting a remote machine.

not sure how many people use `--target` in practice. But I'm assuming nixops would as well.

domenkozar:
https://github.com/NixOS/nix/commit/37352aa7e19e0bfebbd0c32985cbf79a83508538

Issue #356, opened 2014-09-27
Error handling (tryEval?)


jameshfisher:
What documentation is there for error handling in the Nix expression language? I think that evaluation can _abort_ with an error message for many different reasons. I also see a concept of _throwing_ an error message, and this appears to be a separate concept to _abort_.  I also see brief mentions of a builtin called `tryEval`, but its behavior is not documented anywhere. 

Is the error/exception/throw/abort/tryEval semantics defined more precisely anywhere?

sjmackenzie:
@jameshfisher great work on this thread, very considerate of you!

jb55:
(triage) some good stuff in here, what's the status of this issue? can it be closed?

deliciouslytyped:
I'm trying to use `tryEval` to run tests for some complicated nix expressions I was dealing with. I ran into it not being able to catch a lot of thing things.

FaustXVI:
I am having the same issue as @deliciouslytyped . The behavior described by @jameshfisher would solve it so :+1: :smile: 
Looking at the code it seems to me it could be solved like this (please let me know what I am missing) : 
```diff
- mkBool(*state.allocAttr(v, state.sValue), false);
+ mkBool(*state.allocAttr(v, state.sValue), e);
```

lilyball:
I too would really like for the message to be exposed in the `tryEval` results.

lilyball:
I still want this

deliciouslytyped:
Apparently https://github.com/NixOS/nix/issues/5455 https://github.com/NixOS/nix/pull/5564 were closed because "wontfix", though I disagree with the sentiment.

drupol:
Hi!

I'm looking for a way to have the `reason` of the failure in the attribute returned by `tryEval`.

Would it be hard to implement/add such an attribute?

PR #4514, opened 2021-02-02
Include note about type of catched errors in tryEval documentation

sternenseemann:
Reference #356.


src/libexpr/primops.cc
```diff
@@ -696,10 +696,14 @@ static RegisterPrimOp primop_tryEval({
...
-      *e* deeply, so ` let e = { x = throw ""; }; in (builtins.tryEval
-      e).success ` will be `true`. Using ` builtins.deepSeq ` one can
-      get the expected result: `let e = { x = throw ""; }; in
+      successful and `false` otherwise. `tryEval` will only prevent
+      errors created by `throw` or `assert` from being thrown. Errors
+      that are thrown in spite of `tryEval` are, e. g. errors created
+      by `abort` or type errors generated by builtins. Also note that
+      this doesn't evaluate *e* deeply, so ` let e = { x = throw ""; };
+      in (builtins.tryEval e).success ` will be `true`. Using
+      ` builtins.deepSeq ` one can get the expected result:
```
edolstra, line 705:
```suggestion
      this doesn't evaluate *e* deeply, so `let e = { x = throw ""; };
      in (builtins.tryEval e).success` will be `true`. Using
      `builtins.deepSeq` one can get the expected result:
```

sternenseemann: Fixed.

PR #4524, opened 2021-02-05
Check for existing store path in nix profile

matthewbauer:
Previously, nix profile allowed you to install if they had the same
store path.

For instance, you could do:

$ nix profile install github:nixos/nixpkgs#hello
$ nix profile install nixpkgs#hello

This leads to duplication since multiple elements are the same path.
It could also cause issues when upgrading when two elements have the
same store path at one point, but then diverge later, leading to
conflicts.

The new error message looks like:

  error: Cannot install 'github:nixos/nixpkgs#hello' when its store path (/nix/store/lwl497mzqp0k7cfhh51c948nkagc3bbr-hello-2.10) is already installed through 'github:nixos/nixpkgs#legacyPackages.x86_64-linux.hello'

### DIFF
--- a/src/nix/profile.cc
+++ b/src/nix/profile.cc
@@ -235,6 +235,16 @@ struct CmdProfileInstall : InstallablesCommand, MixDefaultProfile
 
         std::vector<StorePathWithOutputs> pathsToBuild;
 
+        // make sure we don’t have this store path already in the profile
+        auto checkForExistingPath = [&](Installable & installable, StorePath & outPath)
+        {
+            for (auto & element : manifest.elements) {
+                if (std::find(element.storePaths.begin(), element.storePaths.end(), outPath) != element.storePaths.end())
+                    throw Error("Cannot install '%s' when its store path (%s) is already installed through '%s'",
+                                installable.what(), store->printStorePath(outPath), element.describe());
+            }
+        };
+
         for (auto & installable : installables) {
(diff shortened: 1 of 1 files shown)


src/nix/profile.cc
```diff
@@ -260,11 +273,14 @@ struct CmdProfileInstall : InstallablesCommand, MixDefaultProfile
 
                     std::visit(overloaded {
                         [&](BuildableOpaque bo) {
+                            checkForExistingPath(*installable, bo.path);
+
```
cole-h, line 277: Nit, but why a blank line only here and not around 283 as well?

edolstra:
Maybe it's better to check this in `ProfileManifest::build()`. For instance, `nix profile upgrade` can also cause multiple profile elements to resolve to the same store path.

matthewbauer:
> multiple

Yeah that sounds fine. It's a little bit less clear what the user should do in that case, but at least it's better than the original error message.

It's kind of an unusual scenario to have two paths installed in a profile that point to different store paths then converge to one store path in a `nix profile upgrade` run. I think it can happen, but every case I can think of would have collisions preventing you from installing them in the first place.

Ericson2314:
I recall some recent changes had something to do with this?

bobvanderlinden:
This is covered by https://github.com/NixOS/nix/pull/10090

PR #4526, opened 2021-02-06
Add Stale bot

### LINKED ISSUE #4522: Enable stalebot
This repository currently has ~1300 open issues and ~200 open PRs, but many of them are very old and may not be applicable anymore. We should enable stalebot to ping contributors and auto-close very old issues/PRs if there is no response (where very old is more than 2 years or so?).

garbas:
The configuration was taken from nixpkgs repository and adjusted to
`NixOS/nix`.

A `stale` label was added to the labels (with gray color).

Issues and PRs with `critical` label are excluded from interacting with the
stale bot.

Fixes #4522

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,9 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+daysUntilStale: 180
+daysUntilClose: false
+exemptLabels:
+  - "critical"
+staleLabel: "2.status: stale"
+markComment: |
+  I marked this as stale due to inactivity. &rarr; [More info](https://github.com/NixOS/nix/blob/master/.github/STALE-BOT.md)
+closeComment: false
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,9 +1,10 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 daysUntilStale: 180
-daysUntilClose: false
+daysUntilClose: 365
 exemptLabels:
   - "critical"
-staleLabel: "2.status: stale"
+staleLabel: "stale"
 markComment: |
   I marked this as stale due to inactivity. &rarr; [More info](https://github.com/NixOS/nix/blob/master/.github/STALE-BOT.md)
-closeComment: false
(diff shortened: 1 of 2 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,9 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+daysUntilStale: 180
+daysUntilClose: false
```
edolstra, line 3: Let's also auto-close issues older than (say) 1 year.

ajs124: That would mean closing a lot of issues, which are definitely still valid, e.g. the two oldest open issues #8 and #13.

edolstra: Sure, but only if nobody comments on them to keep them open.

garbas: Done.

I think what was not said is that even when an issue gets closed, it can be reopened if there is enough interest. Currently it is obvious that there is a management problem, since at the time of writing this we have 1.3k issues and 203 PRs open.

At the same time I'm not trying to say that by closing the issues those issues go away. Far from it. Being closed (and with a stale label) it means there was not enough interest to solve them.

And what happens is that this stale bot in a lot of cases serve as a reminder to get back to the issues/PRs. I know this happened to me on quite few occasions.

nh2: The auto-closing is starting to kill contributions that I think shouldn't be killed:

* https://github.com/NixOS/nix/pull/3208#issuecomment-1100630134
* https://github.com/NixOS/nix/pull/3942#issuecomment-1100514524

abathur: Iirc the open PR count had been hovering between 291 and 296 recently

Ericson2314: Closing was a terrible idea, and done in contra to the community's decision for Nixpkgs in https://github.com/NixOS/rfcs/blob/master/rfcs/0051-mark-stale-issues.md

Ericson2314: I have opened an RFC for this at https://github.com/NixOS/rfcs/pull/124

PR #4528, opened 2021-02-06
nix-daemon: deprecate trusted substituters

### LINKED ISSUE #4520: normal allowed-users should be able to use  --substituters
**Describe the bug**

When an not-trusted allowed-user trys to pass `--substituters` to a command it will bounce it with an error message of ```warning: ignoring untrusted substituter 'file:///tmp/nixstore'```

I believe this is an incorrect error and  should be allowed because
 - Nix will not accept unsigned packages, so even an untrusted substituter would now allow invalid or malicious subtitutions to be entered into the nix store.
 - Both `nix copy` and `nix-copy-closure` works with untrusted users. This means a user can still use any untrusted subsituter they want, but now have to do it run two command instead of.

Given that `nix copy` and `nix-copy-closure` show that a user can load signed packages from any substituter they want, there is no reason why passing `--substituters` to a command should reject untrusted substituters. The packages will either be signed and accepted or unsigned and rejected regardless of the subsituter.

**Steps To Reproduce**

```
# Nix 2.3
% nix run  --substituters "file:///tmp/nixstore https://cache.nixos.org/" /nix/store/ryihq763zni75rwg0xpn469mcak97rqv-nix-2.4pre20201205_a5d85d0 -c nix
warning: ignoring untrusted substituter 'file:///tmp/nixstore'
...
```

```
# Nix 2.4/3.x
% nix shell  --substituters "file:///tmp/nixstore https://cache.nixos.org/" /nix/store/ryihq763zni75rwg0xpn469mcak97rqv-nix-2.4pre20201205_a5d85d0 -c nix
warning: ignoring untrusted substituter 'file:///tmp/nixstore'
...
```

```
# All versions of nix
% nix-env --substituters "file:///tmp/nixstore https://cache.nixos.org/" -i /nix/store/ryihq763zni75rwg0xpn469mcak97rqv-nix-2.4pre20201205_a5d85d0
installing 'nix-2.4pre20201205_a5d85d0'
warning: ignoring untrusted substituter 'file:///tmp/nixstore'
...
```

```
# All versions of nix
nix copy --to file:///tmp/nixstore /nix/store/ryihq763zni75rwg0xpn469mcak97rqv-nix-2.4pre20201205_a5d85d0; echo $?
0
# This works even if the store path has already been GCed and needs to be re-imported.
nix copy --from file:///tmp/nixstore /nix/store/ryihq763zni75rwg0xpn469mcak97rqv-nix-2.4pre20201205_a5d85d0; echo $?
0
```

georgyo:
Having a trusted list of substituters does not actually provde much
security value. A normal not-trusted allowed user has several
methods to add signed binaries to the nix store, such as `nix copy`
and `nix-copy-closure`.

All packages from any substituter must be signed, even from a
"trusted" substituter. This provides much more security that the
package was built in a trusted environment that a list of URLs.
Similarly, if a user could generate signed malicious packages they
have other methods to accomplish their goals.

Closes #4520


grahamc:
A substituter doesn’t need to be signed, it can have ?trusted=1 in the
url to get around it. 

On February 6, 2021, GitHub <notifications@github.com> wrote:
> Having a trusted list of substituters does not actually provde much
>  security value. A normal not-trusted allowed user has several
>  methods to add signed binaries to the nix store, such as nix copy
>  and nix-copy-closure.
>
> All packages from any substituter must be signed, even from a
>  "trusted" substituter. This provides much more security that the

>  package was built in a trusted environment that a list of URLs.

>  Similarly, if a user could generate signed malicious packages they
>  have other methods to accomplish their goals.
>
> Closes #4520 <https://github.com/NixOS/nix/issues/4520>
>

> You can view, comment on, or merge this pull request online at: 

> <https://github.com/NixOS/nix/pull/4528>
>
> Commit Summary
>  *  nix-daemon: depricate trusted substituters
>
> File Changes
>  *  M
>     src/libstore/daemon.cc
>     <https://github.com/NixOS/nix/pull/4528/files#diff-
>     254e9e2297983b49e1135886f091ce73461d7facacb5b7ea78660ba3084f629a>

>     (10) 
>  *  M
>     src/libstore/globals.hh
>     <https://github.com/NixOS/nix/pull/4528/files#diff-
>     9412152b7224a524f972b70909301342567c69c9f391625677764859063c48eb>

>     (9) 
>
> Patch Links:
>  *  <https://github.com/NixOS/nix/pull/4528.patch>
>  *  <https://github.com/NixOS/nix/pull/4528.diff>
>

—
<https://github.com/NixOS/nix/pull/4528>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAASXLCUHUHP6EQEZU6YVSLS5V4CNANCNFSM4XGPRI7Q>.

georgyo:
> A substituter doesn’t need to be signed, it can have ?trusted=1 in the
url to get around it. 

That is both unfortunate and undocumented. The commit that added this gives a bit of background, bb1d046f5c37a01ee85fc30d4602e8be8213eb84.

It makes it unclear what "trusted" means when combined with the configuration options `trusted-substituters`,  `trusted-binary-caches`, and `trusted-users`. You can have a `trusted-substituters` that is not trusted and requires signatures. And you have have a `extra-substituters` that is trusted and doesn't require signatures. And a trusted user can do anything of those things on the fly.

I can see the desire for a `trusted-user` to add a substituter on the fly without requiring that specific one to have signatures, but that makes it much more difficult do distribution via air gaped methods such as USB sticks and CDs; ideally for restrictive and low bandwidth networks.

I'm not sure what the right answer here is.

SuperSandro2000:
You have a typo in the title: deprecate

georgyo:
@grahamc There seems to be only one reference to `?trusted=1` anywhere, and that is in nixpkgs install media; the original use case of adding that flag.

My proposal is that we:
- Add an `unsigned-substituters` and  `extra-unsigned-substituters` option to both nix.conf and the nix tooling. 
 - Only `trusted-users` can pass these arguments.
- Remove the store-api `?trusted=1` method entirely.
- Fully depreciate `trusted-substituters` entirely as the name is confusing and not descriptive of what it allows.

This enables all use cases except for having any method non-trusted user being able to declare an unsigned substituters via the command line. That use case _already_ required modifications to nix.conf to enable that work flow, and we should encourage users to either sign their packages or add the unsigned substituter as an option in nix.conf.

Does this seem like a reasonable path forward?

Ericson2314:
I think I would prefer to aim towards the original: we should always strive to trust public keys, and not care who gives us the data as long as it is signed. I would think the installer can just have signed store entries too so that use-case doesn't matter.

Conversely as long as we have a the non-signed escape hatch, we should also have the "query param" way of doing things as it exists internally like it or not, and allows finer grained control.

osvein:
I would like to see this implemented, but with this patch the substituters will still be accessed by the daemon, i.e. using root's SSH identity and file system permissions. Unlike nix copy, which accesses the source store safely using the client's identity.

thufschmitt:
Discussed during the Nix team meeting. There's a confusion (inducted by the naming) about trusted substituters: The motivation for this “trust” mechanism isn't to prevent adding arbitrary paths to the store (that's indeed the role of signatures), but to prevent users to leak private information. Otherwise, since the substitution is done by the daemon (which is root), someone could run a build with `--substituters file:///super-secret-directory` and use that to get information about it.

That's something that should be reflected in the docs for `trusted-substituters` (PR welcome :) )

- Signing is indeed sufficient to make sure we don't trust data we shouldn't.

- The fundamental issue here is actually separate from signing. When using a daemon, we don't want the user to specify to the daemon substituter which lead it to make publically accessible information which would not otherwise be.

  - @thufschmitt left a good comment on this elsewhere, but we are having trouble finding it.

- Might want to rename and tweak this feature to make it more flexible.
</details>

AriFordsham:
Ultimately, maybe the tidiest solution would be to have the frontend do the downloading from the substituter, ander the user's permissions, and have the daemon expose a mechanism for adding signed paths to the store.

thufschmitt:
> Ultimately, maybe the tidiest solution would be to have the frontend do the downloading from the substituter, under the user's permissions, and have the daemon expose a mechanism for adding signed paths to the store.

Yes, absolutely. That would be quite a long-term thing though, because all the scheduling is currently done by the daemon and it would have to be moved client-side for something like this to work

georgyo:
> Ultimately, maybe the tidiest solution would be to have the frontend do the downloading from the substituter, under the user's permissions, and have the daemon expose a mechanism for adding signed paths to the store.

This is actually exactly what `nix copy` and `nix-copy-closure` does today. All the downloading happens client side and that daemon only accepts packages if they are signed.

The problem is that you also need a way to express unsigned substituter as well. You cannot have a normal, non-trusted, user download content from an unsigned substituter in a way the daemon could accept. 

In the context of this ticket, I still think that `?trusted=1` should be removed as a URL parameter. Or at least configurable in nix.conf or ignored by non-trusted users. That flag is the singular thing that prevents normal users from specifying different subsituters today. It is also remains undocumented.

tomberek:

Issue #4531, opened 2021-02-07
macOS Big Sur updates w/ multi-user Nix present "booted to Recovery" because "an error occurred migrating user data"


abathur:
_No need to triage this. Opening it for visibility/support. To keep this focused, I'll save detail/discussion for the related PR._

---

We've had several people *with multi-user/daemon nix installs* report trouble with system updates on IRC. (Single-user installs have been the default for a while; if you don't *know* you have a multi-user install, you probably don't). 

During the multiple-reboot phase of a system update, the system will boot into recovery and show the message:
> **Your computer booted to Recovery because a failure occurred during installation.**
> An error occurred migrating user data during an install.
> Reinstall macOS to resolve the issue.

### Workarounds
1. **If you run into this issue during an update:** Multiple users have reported getting around this problem **without having to reinstall macOS** by simply clicking OK and rebooting again, and/or by opening Disk Utility in recovery, running First Aid, and rebooting.
2. **If you have not updated yet:** I've prepared a nixbld-user migration script that should let you skip the heartburn. I'll add a link here once it's available (I'm waiting for a test run to complete).

emilazy:
This was fixed by #4532, right? Or do we still need automation to migrate existing accounts?

abathur:
Fixed will be squishy, here. 

4532 should start fixing new installs whenever a release is cut, but makes no attempt to automatically fix existing installs.

tomberek:
Closing as stale. Re-open if needed.

PR #4532, opened 2021-02-07
fix nixbld user name/uid for macOS/darwin

abathur:
I've been able to confirm the update problem reported in #4531 on a spare system. This PR will:
1. Add a migration script people can run before taking the update to avoid the problem. (We probably don't need to carry this in the source for long, but it seems good to have a canonical source to link.)
2. Update the installer to adopt the same values (just for macOS--a hedge against this being an ongoing problem with every update).

## Background
I'm enough of a masochist to run clean-install + upgrade cycles until I have a vague sense of the problem:

1. It looks like the upgrade is doing some sort of user/account migration, and faceplants over something about our nixbld users (i.e., it runs clean if you remove them first).
2. ~I get the impression, because of the report timing, that this is something new after Big Sur 11.0 and probably new in 11.2.~ 
    - Reports in this thread have clarified that this has been an ongoing pain point on Big Sur. 
    - I've since confirmed that I see this with the update to 11.3 beta--it definitely seems like this will be an ongoing problem.
4. It doesn't seem to be a critical problem, as long as you ignore the error message that tells you to reinstall (i.e., users who jump the turnstile haven't reported ongoing trouble.)
5. I eventually found a sane way around when I stumbled on some new `sysadminctl -addUser` flags (`[-GID <group ID>]`, `[-roleAccount]`) and this usage note `*Role accounts require name starting with _ and UID in 200-400 range`. 
    - I confirmed that the update runs clean if we change `nixbld#` usernames to `_nixbld#` and use UIDs in the 200-400 range. 
    - The GID doesn't appear to matter, here.

## Open questions
1. I chose 301 as the new starting UID, but it's just a guess. Objections?
    - There are many macOS system role accounts in the 200-300 range (but none from 300-400), and at least on Catalina the only free contiguous block is from 271-299. I read this as 200-300 being Apple's territory, but if they go make a few dozen new role users I assume they'll just keep rolling into the 300s? 
    - I don't know if these are hard-set or dynamically allocated at update/install? In the latter case it'd be fine if they keep rolling into the 300s; in the former it might cause trouble.
2. Taking up 32 UIDs from 301-332 by default feels a bit greedy (but it may be wasted energy to sweat this?) Is it feasible to make a more-judicious choice (at least on macOS, perhaps based on `sysctl -n hw.ncpu` or `sysctl -n hw.physicalcpu`?)

### DIFF
--- a/scripts/bigsur-nixbld-user-migration.sh
+++ b/scripts/bigsur-nixbld-user-migration.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+((NEW_NIX_FIRST_BUILD_UID=301))
+
+id_available(){
+	dscl . list /Users UniqueID | grep -E '\b'$1'\b' >/dev/null
+}
+
+change_nixbld_names_and_ids(){
+	local name uid next_id
+	((next_id=NEW_NIX_FIRST_BUILD_UID))
+	echo "Attempting to migrate nixbld users."
+	echo "Each user should change from nixbld# to _nixbld#"
+	echo "and their IDs relocated to $next_id+"
+	while read -r name uid; do
+		echo "   Checking $name (uid: $uid)"
+		# iterate for a clean ID
+		while id_available "$next_id"; do
+			((next_id++))
+			if ((next_id >= 400)); then
+				echo "We've hit UID 400 without placing all of your users :("
+				echo "You should use the commands in this script as a starting"
+				echo "point to review your UID-space and manually move the"
+				echo "remaining users (or delete them, if you don't need them)."
+				exit 1
+			fi
+		done
+
+		if [[ $name == _* ]]; then
+			echo "      It looks like $name has already been renamed--skipping."
+		else
+			# first 3 are cleanup, it's OK if they aren't here
+			sudo dscl . delete /Users/$name dsAttrTypeNative:_writers_passwd &>/dev/null || true
+			sudo dscl . change /Users/$name NFSHomeDirectory "/private/var/empty 1" "/var/empty" &>/dev/null || true
+			# remove existing user from group
+			sudo dseditgroup -o edit -t user -d $name nixbld || true
+			sudo dscl . change /Users/$name UniqueID $uid $next_id
+			sudo dscl . change /Users/$name RecordName $name _$name
+			# add renamed user to group
+			sudo dseditgroup -o edit -t user -a _$name nixbld
+			echo "      $name migrated to _$name (uid: $next_id)"
+		fi
+	done < <(dscl . list /Users UniqueID | grep nixbld | sort -n -k2)
+}
+
+change_nixbld_names_and_ids
--- a/scripts/install-darwin-multi-user.sh
+++ b/scripts/install-darwin-multi-user.sh
@@ -4,6 +4,8 @@ set -eu
 set -o pipefail
 
 readonly PLIST_DEST=/Library/LaunchDaemons/org.nixos.nix-daemon.plist
+NIX_FIRST_BUILD_UID="301"
+NIX_BUILD_USER_NAME_TEMPLATE="_nixbld%d"
 
 dsclattr() {
     /usr/bin/dscl . -read "$1" \
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -25,7 +25,9 @@ readonly RED='\033[31m'
 readonly NIX_USER_COUNT=${NIX_USER_COUNT:-32}
 readonly NIX_BUILD_GROUP_ID="30000"
 readonly NIX_BUILD_GROUP_NAME="nixbld"
-readonly NIX_FIRST_BUILD_UID="30001"
+# darwin installer needs to override these
+NIX_FIRST_BUILD_UID="30001"
+NIX_BUILD_USER_NAME_TEMPLATE="nixbld%d"
 # Please don't change this. We don't support it, because the
 # default shell profile that comes with Nix doesn't support it.
 readonly NIX_ROOT="/nix"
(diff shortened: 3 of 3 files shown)


emilazy:
Maybe we could count down from 400 to minimize the risk of an update collision disaster down the line? The number of nixbld users limits how high a `-j` you can use, and there's frequently reason to want to build more jobs than you have cores (especially with hyperthreading, but even without if most of your builds are short/IO-heavy and the compute-heavy stuff doesn't overload your CPU; e.g. Rust does its own concurrency-limiting logic anyway), so I'd be a bit hesitant about reducing it too far.

kevingriffin:
I'd like to add that I've actually seen this behavior in Big Sur ever since the betas, on both Intel and ARM. Our entire organization uses nix for development machines, and we've seen it consistently across updates.

If what we saw was indeed this issue, it seems pretty likely that this isn't limited to 11.2, but is just part of something new in Big Sur. I have a feedback ticket open with Apple about this issue, and I'll link to this PR as part of the dialogue there.

dhess:
> 2. I get the impression, because of the report timing, that this is something new after Big Sur 11.0 and probably new in 11.2.

It's not super important, but we run Nix on several different Macs of various makes and models, and we saw this issue on both the 11.0->11.1 update, and the 11.1->11.2 update. I didn't really think anything of it, nor mention it in IRC, during the 11.0->11.1 update because I figured it was just a typical Apple one-off bug. But now I'm certain that it's been around since the 11.1 update.

(We may have also seen it with the original 11.0 install, but I don't recall anymore.)

abathur:
@kevingriffin @dhess Thanks for clarifying this; updated the summary.

matthewbauer:
Does the group id not matter? It's still 30000 here.

The migrate script doesn't appear to be working since the `nixbld` group still has references to nixbld1, nixbld2, nixbld3, nixbldN:

```
$ dscacheutil -q group -a name nixbld
name: nixbld
password: *
gid: 30000
users: nixbld1 nixbld2 nixbld3 nixbld4 nixbld5 nixbld6 nixbld7 nixbld8 nixbld9 nixbld10 nixbld11 nixbld12 nixbld13 nixbld14 nixbld15 nixbld16 nixbld17 nixbld18 nixbld19 nixbld20 nixbld21 nixbld22 nixbld23 nixbld24 nixbld25 nixbld26 nixbld27 nixbld28 nixbld29 nixbld30 nixbld31 nixbld32
```

I'm not sure how to fix this. I ran:

```
for i in $(seq 1 32); do sudo dscl . change /Users/_nixbld$i RecordName _nixbld$i nixbld$i; done
```

to reverse this.

abathur:
> Does the group id not matter? It's still 30000 here.

It doesn't appear to, so I didn't touch it.

> The migrate script doesn't appear to be working since the `nixbld` group still has references to nixbld1, nixbld2, nixbld3, nixbldN

Good catch, it looks like I got caught up in optimizing for the smallest change that didn't drop us into recovery. I'll disable the recommendation in the other issue and take a look at it.

abathur:
@matthewbauer I've updated it to remove each account with dseditgroup before the change and re-add it after. I have force-pushed this change, but won't have a chance to test the whole script again until later today.

<details><summary>If you've already run this script (and have not reverted it) you should be able to fix group membership with these commands</summary>
```
sudo dscl . delete /Groups/nixbld dsAttrTypeStandard:GroupMembership
sudo dscl . append /Groups/nixbld dsAttrTypeStandard:GroupMembership _nixbld{1..32}
```
</details>

abathur:
I'll also update the main post, but just wanted to mention here that I've added some extra evidence that this will be an ongoing problem:
1. Install macOS 11.0 (beta), and Nix
2. Update to 11.2, observe boot to recovery as already described, and reboot into the OS.
3. Update to 11.3 (beta), and get booted into a fresh hell. This time, after making me pass an activation lock with my AppleID, it booted me to some sort of recovery/reset screen I wasn't already familiar with that was trying to goad me into resetting the password for all of my accounts (the list shows all of the nixbld accounts, with my real one mixed in). It never prompted me for an admin login previous to this. Despite this, I was still able to simply boot into the OS.

I'm starting to think it'll make sense to pin #4531?

abathur:
If you've been affected by this and want to report it to Apple, you can refer to the feedback I filed last night: FB8997501

dhess:
I ran the `nixbld -> _nixbld` conversion script in this PR on an 11.2 system that had previously had problems with System Update, then updated to 11.2.1, and everything went fine!

However, this system uses `nix-darwin` and when I ran `darwin-rebuild switch` after the update, `nix-darwin` created these users again, so there are more moving parts here, I'm afraid.

abathur:
I've rebased to re-run tests now that #4577 is merged. macOS install: https://github.com/abathur/nix/runs/2088031573

robertoschwald:
> However, this system uses `nix-darwin` and when I ran `darwin-rebuild switch` after the update, `nix-darwin` created these users again, so there are more moving parts here, I'm afraid. Anyway, the System Update issue appears to root-caused!

Does not happen here. Users stay _nixbldXX after darwin-rebuild switch on my systems.

dhess:
> > However, this system uses `nix-darwin` and when I ran `darwin-rebuild switch` after the update, `nix-darwin` created these users again, so there are more moving parts here, I'm afraid. Anyway, the System Update issue appears to root-caused!
> 
> Does not happen here. Users stay _nixbldXX after darwin-rebuild switch on my systems.

It's been fixed in nix-darwin since I wrote that comment on Feb 9. :)

domenkozar:
@abathur should we backport this to 2.3?

abathur:
@domenkozar I think so.

domenkozar:
Pushed to `2.3-maintenance` branch.

PR #4543, opened 2021-02-12
Low level `<drvPath>^<outputName>` installable syntax to match existing `<highLevelInstallable>^<outputNames>` syntax 

Ericson2314:
https://github.com/NixOS/nix/pull/6426 added `^<output-names>` to high level installables (flakes ones and attribute paths). This adds it to our low level installable --- a plain store path --- to match.

I think this is important because it greatly helps "complete" the expressive power of Nix's low-level "plumbing" commands, and doing that (and eventually stabilizing them) is a good cost / benefit win. 

In addition:

One ingredient for #7261, but should be good even if we reject that issue.
Also important for RFC 92 (tracking issue #6316).


matthewbauer:
I think CmdProfileInstall also needs to change to reflect this:

https://github.com/NixOS/nix/blob/master/src/nix/profile.cc#L267-L270

so only outputs listed in BuildableFromDrv "outputs" are installed.

regnat:
I'm not sure what this brings implementation-wise compared to https://github.com/NixOS/nix/pull/4494/ (except obviously for the doc and test which are obviously missing in the latter). In particular it seems to me that having an extra `Installable` type just for that is a lot of boilerplate for not much benefit. Or do I miss something?

Ericson2314:
@regnat I'm afraid I didn't even realize you had a PR this did this!

Ericson2314:
@regnat I think the simpler and pure (no DB access) `toBuildables` is notable. (To be clear, yours made the problem no worse, just inherited the complexity) I am very skeptical of all the complexity we have today (along with `--derivation`) and so hope this can be the beginning of getting rid of it.

zseri:
This probably doesn't solve the shell 'escape needed' problems. It shows probably unwanted e.g. in bash:
```
$ nix profile install nixpkgs!lol
bash: !lol: event not found
```

Ericson2314:
Yeah we didn't settle on a surface syntax yet.

zseri:
Currently, this syntax is (already) used with CA-drvs, which leads to problems with `nix-collect-garbage`, because it terminates when it finds any store which uses that syntax. Maybe this should be fixed in a smaller PR earlier if possible, because this one seems to take another while until merged (and I'm not sure if this PR does really fix that annoyance).

regnat:
> Currently, this syntax is (already) used with CA-drvs, which leads to problems with nix-collect-garbage, because it terminates when it finds any store which uses that syntax

Unless I missed something, these have only been created because of a now-closed bug (https://github.com/NixOS/nix/issues/5176), and it only concerns lockfiles. Is there more of these?

Ericson2314:
Yeah this has nothing to do with that one, which @regnat fixed. This is just making an internal syntax exposed to the user.

In this case, we in fact *rely* on `!` being disallowed in store paths, so there is no ambiguity.

tomberek:
To clarify, this is existing syntax that happens to not function with installables, but would otherwise be allowed. Would this leak into things like profiles or machine-readable output?

Ericson2314:
This is preexing syntax for some legacy commands (e.g. `nix-build`) and used internally (though I modernized it). Basically, see `DerivedPath` and `StorePathWithOutputs` in the code base, I have made the former replace the latter where possible.

This specific change won't leak anywhere, this is just changing the CLI syntax for installables. In way, the leaking behind the scenes is already done! --- thanks to those `DerivedPath` refactors underneath the hood.

This, or something like this, is part of RFC 92, so we should figure out what the experimental feature is called and get this merged. I had been waiting to hopefully add some more tests firsts for the status quo first, but if that is controversial we should just go ahead with this and the other parts of RFC 92.

Rahul75-dev (approved):

Ericson2314:
OK, experimental feature added as discussed in the UI call today!

Ericson2314:
I fixed conflicts in the change log. This needs to be reviewed. We cannot drag on the implementation of an accepted RFC indefinitely.

tomberek (commented):
- perhaps squash?
- should we expect something like: `nixpkgs#some-attr!out!out` to work in the flakes paradigm?

otherwise looks like step one of rfc92

Ericson2314:
> This does change behavior for performing a build referencing a drv file. 

It only changes the behvior in that `^` (I changed it from `!`) is no longer valid in a CLI store path. if you pass `/nix/store/<hash>-<name>.drv` where there is no `^` in that store path, nothing is changed.

Ericson2314:
@edolstra I combined the structs as you requested. I also realized more functionality was supported than I originally thought, so I updated the tests accordingly.

src/libcmd/installables.cc
```diff
@@ -796,7 +808,23 @@ std::vector<std::shared_ptr<Installable>> SourceExprCommand::parseInstallables(
...
             std::exception_ptr ex;
 
-            if (s.find('/') != std::string::npos) {
+            auto found = s.rfind('^');
+            if (found != std::string::npos) {
+                try {
+                    result.push_back(std::make_shared<InstallableStorePath>(
+                        store,
+                        DerivedPath::Built::parse(*store, s.substr(0, found), s.substr(found + 1))));
+                    settings.requireExperimentalFeature(Xp::ComputedDerivations);
```
edolstra, line 817:
```suggestion
```

src/nix/nix.md
```diff
@@ -130,6 +130,19 @@ the Nix store. Here are the recognised types of installables:
   If you want to operate on the store derivation itself, pass the
   `--derivation` flag.
 
+* **Indexed store derivations**: `/nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv^out`
+
+  *(Experimental, part of by the `computed-derivations` experimental feature.)*
+
+  Store derivations can be indexed with a non-empty comma-separated list
+  of specific output names, or `*` meaning all ouptuts. This allows
```
edolstra, line 138:
```suggestion
  of specific output names, or `*` meaning all outputs. This allows
```

src/nix/nix.md
```diff
@@ -130,6 +130,19 @@ the Nix store. Here are the recognised types of installables:
   If you want to operate on the store derivation itself, pass the
   `--derivation` flag.
 
+* **Indexed store derivations**: `/nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv^out`
+
+  *(Experimental, part of by the `computed-derivations` experimental feature.)*
```
edolstra, line 135:
```suggestion
```

src/nix/nix.md
```diff
@@ -130,6 +130,19 @@ the Nix store. Here are the recognised types of installables:
...
+
+  Store derivations can be indexed with a non-empty comma-separated list
+  of specific output names, or `*` meaning all ouptuts. This allows
+  finer control versus just specifying a derivation (without
+  `--derivation`) and getting all the outputs.
+
+  This is especially useful for (currently unstable) floating content
+  addressed derivations, which do not have precomputed output paths that
+  can be used instead.
+
```
edolstra, line 145: I think this section can be generalized, since the `^` syntax is supported for other types of installables (e.g. `nixpkgs#hello^bin`). We can start with this description from the 2.9 release notes:
```
* You can now specify which outputs of a derivation `nix` should
  operate on using the syntax `installable^outputs`,
  e.g. `nixpkgs#glibc^dev,static` or `nixpkgs#glibc^*`. By default,
  `nix` will use the outputs specified by the derivation's
  `meta.outputsToInstall` attribute if it exists, or all outputs
  otherwise.
```

Ericson2314: Indeed it can, in fact you already added even more complete docs below for that I will just fold the new stuff into.

src/libstore/derived-path.cc
```diff
@@ -93,24 +93,25 @@ DerivedPath::Opaque DerivedPath::Opaque::parse(const Store & store, std::string_
     return {store.parseStorePath(s)};
 }
 
-DerivedPath::Built DerivedPath::Built::parse(const Store & store, std::string_view s)
+DerivedPath::Built DerivedPath::Built::parse(const Store & store, std::string_view drvS, std::string_view outputsS)
```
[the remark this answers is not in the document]
Ericson2314: See https://github.com/NixOS/nix/pull/6815 which I made so this deduplication can happen. Note that part of the complication is `!` is still used in the remote store protocol for `DerivedPath`, so some trickiness is needed so `^` and `!` are used for different use-cases, but we can do better de-duplicating than we do currently.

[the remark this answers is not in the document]
Ericson2314: @edolstra can we return to this after this PR? #6815 to deduplicate things has prooved to be much bigger than this. I would really appreciate being able to make this change and then do the clean up afterwards.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -1,3 +1,8 @@
 # Release X.Y (202?-??-??)
 
-* Allow explicitly selecting outputs with *store derivations* installable syntax too.
+* `<nix/fetchurl.nix>` now accepts an additional argument `impure` which
+  defaults to `false`.  If it is set to `true`, the `hash` and `sha256`
+  arguments will be ignored and the resulting derivation will have
```
zseri, line 5: wouldn't it make more sense to make it an error to specify these arguments, or would that break when combined with indirect usage in nixpkgs or such?

Ericson2314: That is unrelated to this PR. I happen to agree but please bring up elsewhere.

roberth:
We already have a high-level selector `.` that works at the package attrset level instead of the derivation level. Most packaging users won't need to know about `^`.
`builtins.outputOf` will be used in package outputs so that users of the package don't need to know that a dynamic derivation was involved.
The high level selector is simpler. It implicitly supports `^*`, by not using it. Multiple outputs are not supported, but this is not much of a problem because the shell can take care of that: `nix build .#pandoc.{data,out}`.
For the same reason I'd say multi-output selection can be removed, but I guess it's too late for that?

Ericson2314:
@roberth So keep in mind that today:

1. `^` is added for high level things too, see https://github.com/NixOS/nix/pull/6426. `nix build .#pandoc^*` is all outputs, `nix build .#pandoc` is just "outputs to install"
2. `nix build .#pandoc.data` and `nix build .#pandoc.out` will do the same thing ("outputs to install")!

I think (2) is terrible, but hopefully we can fix it. It would be nice to use `.` everywhere but it is already a valid store path so it is ambiguous there.

src/libstore/derived-path.cc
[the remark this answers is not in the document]
Ericson2314: A simpler patch is not available because we still need to parse with with both `!` and `^`, because the former is still used internally e.g. in the protocols.

roberth:
> @roberth So keep in mind that today:
> 
> 1. `^` is added for high level things too, see [nix: Respect meta.outputsToInstall, and use all outputs by default #6426](https://github.com/NixOS/nix/pull/6426). `nix build .#pandoc^*` is all outputs, `nix build .#pandoc` is just "outputs to install"

So this relies on the derivation rather than the package attributes, right? I'd consider that low level, because while it starts at the expression level, it then goes down to the derivation level for the outputs. It does not evaluate the `.out` attribute.

> 2. `nix build .#pandoc.data` and `nix build .#pandoc.out` will do the same thing ("outputs to install")!
> 
> I think (2) is terrible, but hopefully we can fix it.

This looks alright to me.

```console
$ nix build .#pandoc.out --print-out-paths 
/nix/store/zdhywqp4iaav96mw1nal6px9bwk51jz2-pandoc-2.17.1.1
$ nix build .#pandoc.data --print-out-paths 
/nix/store/v6wmag6m1dfmkslq2zdya7m8lwhlycww-pandoc-2.17.1.1-data
```

Maybe the problem is with another command?

>  It would be nice to use `.` everywhere but it is already a valid store path so it is ambiguous there.

Alternatively, it would also be nice for `.` to always mean attribute selection. Maybe we should see it that way?

thufschmitt:
>  > 2.  nix build .#pandoc.data and nix build .#pandoc.out will do the same thing ("outputs to install")!
> >
> > I think (2) is terrible, but hopefully we can fix it.
>
> This looks alright to me.

IIRC this was the case for a short amount of time, but then Eelco found a trick to make it work properly again

Ericson2314:
Oh great!

Ericson2314:
@roberth All three of your points sound good to me. This is indeed low level; I just meant it is no longer "out of nowhere" but in sync with the higher level syntax.

src/libcmd/installables.cc
```diff
@@ -395,44 +395,56 @@ static StorePath getDeriver(
 struct InstallableStorePath : Installable
 {
     ref<Store> store;
-    StorePath storePath;
+    DerivedPath req;
```
roberth, line 398: This is not actually a single path, despite the type name.
I'm ok to leave that as tech debt for now.
See https://github.com/NixOS/nix/issues/6507#issuecomment-1295822346

doc/manual/src/release-notes/rl-next.md
zseri: this stems from the merging of `master` right?

src/libcmd/installables.cc
Ericson2314: It does actually need to be multiple paths now because of `^foo,bar` and `^*`.

src/libstore/derived-path.cc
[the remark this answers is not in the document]
Ericson2314: https://github.com/NixOS/nix/pull/6815 now works! But I'll still keep it separately because that is much bigger than this.

fricklerhandwerk:
Discussed in Nix team meeting 2022-11-28:


- @fricklerhandwerk: what are installables anyway?
    - @Ericson2314: when you do `nix build foo bar baz`, each of the arguments could be
        - store derivation path
        - store path
            - which could be not built but substitutable
        - path to Nix expression
        - flake reference
    - @Ericson2314: we can already do `nix build nixpkgs#foo^man` (i.e., flake references), but not for store paths such as `nix build /nix/store/<hash>-foo^man`.
    - Are there any objections that we want to have this or the implementation strategy?
        - agreement on both

-> to review

doc/manual/src/release-notes/rl-next.md
```diff
@@ -2,3 +2,5 @@
 
 * The `repeat` and `enforce-determinism` options have been removed
   since they had been broken under many circumstances for a long time.
+
+* Allow explicitly selecting outputs with *store derivations* installable syntax too.
```
edolstra, line 6: This is a bit vague, maybe just give an example (`foo.drv^out`).

src/libcmd/installables.cc
```diff
@@ -399,44 +399,56 @@ static StorePath getDeriver(
...
 
     InstallableStorePath(ref<Store> store, StorePath && storePath)
-        : store(store), storePath(std::move(storePath)) { }
+        : store(store),
+          req(storePath.isDerivation()
+              ? (DerivedPath) DerivedPath::Built {
+                  .drvPath = std::move(storePath),
+                  .outputs = {},
```
Ericson2314, line 409: Note it previously read the derivation and returned all its output names explicitly here, but that didn't work e.g. when the drv file itself needs to be downloaded and so we cannot yet read it.

src/libstore/derived-path.cc
```diff
@@ -78,24 +78,25 @@ DerivedPath::Opaque DerivedPath::Opaque::parse(const Store & store, std::string_
...
+            throw Error(
+                 "Explicit list of wanted outputs '%s' must not be empty. Consider using '*' as a wildcard meaning all outputs if no output in particular is wanted.", outputsS);
+	}
     return {drvPath, outputs};
 }
 
 DerivedPath DerivedPath::parse(const Store & store, std::string_view s)
 {
-    size_t n = s.find("!");
+    size_t n = s.rfind("!");
```
Ericson2314, line 96:
```suggestion
    size_t n = s.find("!");
```

src/nix/nix.md
```diff
@@ -177,9 +184,17 @@ operate are determined as follows:
   /nix/store/q6580lr01jpcsqs4r5arlh4ki2c1m9rv-glibc-2.33-123-dev             44200560
   ```
 
-* If you didn't specify the desired outputs, but the derivation has an
-  attribute `meta.outputsToInstall`, Nix will use those outputs. For
-  example, since the package `nixpkgs#libxml2` has this attribute:
+  and likewise, again using a store path to a "drv" file to specify the derivation:
+
+  ```console
+  # nix path-info -S --eval-store auto --store https://cache.nixos.org '/nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv^*'
```
edolstra, line 190:
```suggestion
  # nix path-info -S '/nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv^*'
```

src/nix/nix.md
```diff
@@ -177,9 +184,17 @@ operate are determined as follows:
...
+  …
+  ```
+
+* If you didn't specify the desired outputs, but the derivation comes
+  from an expression which has an attribute `meta.outputsToInstall`, Nix
```
edolstra, line 195: Probably best to change back to the original wording since a store .drv does not have `meta.outputsToInstall`.

edolstra (approved):

tests/build-explicit-output.sh
```diff
@@ -0,0 +1,43 @@
+source common.sh
```
Ericson2314, line 1: @edolstra says just have one `build.sh`

src/nix/nix.md
```diff
@@ -190,11 +190,9 @@ operate are determined as follows:
   # nix path-info -S '/nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv^*'
   …
   ```
-
-* If you didn't specify the desired outputs, but the derivation comes
-  from an expression which has an attribute `meta.outputsToInstall`, Nix
-  will use those outputs. For example, since the package
-  `nixpkgs#libxml2` has this attribute:
+* If you didn't specify the desired outputs, but the derivation hs an
```
fricklerhandwerk, line 193:
```suggestion
* If you didn't specify the desired outputs, but the derivation has an
```

Ericson2314:
OK @edolstra, I think all review from today is now addressed!

amjoseph-nixpkgs:
Hey I know this is the bikesheddiest of bikesheds, but I just realized we could have used `/` instead of `^`:

```
nix build /nix/store/09wshq4g5mc2xjx24wmxlw018ly5mxgl-bash-5.2-p15.drv/man
```

... since Nix strictly forbids a top-level store path which isn't a derivation from having the suffix `.drv`.  Since derivations are files, not directories, the string above can never be a valid filesystem path.

Who cares, you ask?  Maybe it's just me, but the `6` key on a QWERTY keyboard is farther from the home row position than any other key, meaning that `6` and `^` are mistyped more often than any other character.

Like I said, bikesheds.  Glad I got that out of my system.  We now return you to your regularly-scheduled software development...

Ericson2314:
@amjoseph-nixpkgs I think some command allow you specify a store path with trailing path relative to it? It could get a git confusing in that case. On a conceptual level I want `/nix/store/09wshq4g5mc2xjx24wmxlw018ly5mxgl-bash-5.2-p15.drv^man/share/man` to work too, but I realize that can also be confusing and maybe it's better to stick to some sort of  `${...}` notation.

PR #4549, opened 2021-02-15
Test macos/linux installer script for each push

### LINKED ISSUE #4047: Create a working installer script for each commit/PR
Using https://github.com/actions/upload-release-asset we could create a `prerelease` release for each master commit
and upload all artifacts needed for the installer script to work.

The downside is that this wouldn't work for PRs (not that big of a deal) and that release page would quickly get polluted while there's no way to filter out prereleases.

domenkozar:
This works by using Cachix feature of serving a file from
a store path.

Fixes https://github.com/NixOS/nix/pull/4564 
Fixes #4047

cc @edolstra

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,10 +8,52 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    env:
+      CACHIX_NAME: nix-ci
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v2.3.4
       with:
         fetch-depth: 0
     - uses: cachix/install-nix-action@v12
+    - uses: cachix/cachix-action@v8
+      with:
+        name: '${{ env.CACHIX_NAME }}'
+        signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
     #- run: nix flake check
     - run: nix-build -A checks.$(if [[ `uname` = Linux ]]; then echo x86_64-linux; else echo x86_64-darwin; fi)
+  installer:
+    if: github.event_name == 'push'
+    needs: tests
+    runs-on: ubuntu-latest
+    env:
+      CACHIX_NAME: nix-ci
+    outputs:
+      installerURL: ${{ steps.prepare-installer.outputs.installerURL }}
+    steps:
+    - uses: actions/checkout@v2.3.4
+      with:
+        fetch-depth: 0
+    - uses: cachix/install-nix-action@v12
+    - uses: cachix/cachix-action@v8
+      with:
+        name: '${{ env.CACHIX_NAME }}'
+        signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+    - id: prepare-installer
+      run: scripts/prepare-installer-for-github-actions
+  installer_test:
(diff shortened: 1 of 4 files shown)


edolstra:
Wouldn't it be better to serve the tarball from GitHub as an uploaded artifact?

> It doesn't yet works because it seems like that flake shim sets self.shortRev for checks attribute but not for outputs?

This is probably because changing `flake.nix` dirties the tree, so there is no `shortRev`.

domenkozar:
> Wouldn't it be better to serve the tarball from GitHub as an uploaded artifact?

Yes! However that's really tricky to implement as the installer script wants urls hardcoded for each platform.

> This is probably because changing flake.nix dirties the tree, so there is no shortRev.

What's the workaround here? A new flake that changes input for `systems`? Some way to pass in `--arg`?

SuperSandro2000 (commented):
You are missing final new lines.

domenkozar:
@abathur this PR creates an artifact, that is a zip file containing a tar that has the executable to install Nix.

That's a huge hack due to limitations of GHA that it always creates a zip out of artifacts and that those artifacts are never executables.

Maybe it would be easier to just print the url for the installer instead, that would allow us to use it with cachix-action.

WDYT?

domenkozar:
This PR now adds automated testing of installer script via GHA.

- builds installer script for linux/macos and pushes it to Cachix
- uses install-nix-action by pointing to Cachix (it has a non-advertised feature to serve files from nars)

domenkozar:
@edolstra this is ready to be merged and it already caught a bug :)

regnat (approved):
Really cool!

Is there a way to print some instructions for installing this locally? Afaict the actuall installer URL is never shown in the logs nor exported in any way (maybe this could even be exported in an artifact)

domenkozar:
I tried an artifact before, but it doesn't preserve execution bit.

install-nix-action inputs are printed in beginning of the log

regnat:
> I tried an artifact before, but it doesn't preserve execution bit.

Yeah I saw that, I was just thinking of a plain text file containing the actual URL or something like that. But it's not really important, and certainly not a blocker for this PR.

> install-nix-action inputs are printed in beginning of the log

Uh. So today I learned that the little triangle at the begining of the `Run` line isn't some kind of shell prompt but a button to display more infos about what's being run :man_facepalming:

.github/workflows/test.yml
```diff
@@ -8,10 +8,52 @@ jobs:
...
         fetch-depth: 0
     - uses: cachix/install-nix-action@v12
+    - uses: cachix/cachix-action@v8
+      with:
+        name: '${{ env.CACHIX_NAME }}'
+        signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
     #- run: nix flake check
     - run: nix-build -A checks.$(if [[ `uname` = Linux ]]; then echo x86_64-linux; else echo x86_64-darwin; fi)
+  installer:
+    if: github.event_name == 'push'
```
Artturin, line 25: The commit message says `PR/push` while this is only on push

would this test have caught the typo introduced in https://github.com/NixOS/nix/pull/6285 if the test was run?

abathur: Probably, though iirc they still won't run if you don't have your own cachix key (and, as a glance at the master history shows, these are not always green).

PR #4551, opened 2021-02-16
Add x86_64 compute levels as system features

danieldk:
When performing distributed builds of machine learning packages, it
would be nice if builders without the required SIMD instructions can
be excluded as build nodes using `requiredSystemFeatures`.

Since x86_64 has accumulated a large number of different instruction
set extensions, listing all possible extensions would be unwieldy.
AMD, Intel, Red Hat, and SUSE have recently defined four different
microarchitecture levels that are now part of the x86-64 psABI
supplement and will be used in glibc 2.33:

https://gitlab.com/x86-psABIs/x86-64-ABI
https://lwn.net/Articles/844831/

This change uses libcpuid to detect CPU features and then uses them
to add to system-features the supported x86_64 levels. For example
on a Ryzen 3700X:

```
$ src/nix/nix show-config nix-command  | grep system-feature
system-features = benchmark big-parallel kvm nixos-test recursive-nix x86-64-v1 x86-64-v2 x86-64-v3
```


edolstra:
Maybe these should be system types instead of features. Because if a package is compiled for x86-64-v3, you don't want to install it on a machine that doesn't support those instructions.

danieldk:
> Maybe these should be system types instead of features. Because if a package is compiled for x86-64-v3, you don't want to install it on a machine that doesn't support those instructions.

That probably makes more sense. I have updated the PR. Should the types be `x86-64-v[1234]-linux` like they are used by glibc/gcc or would it be better to use `x86_64-v[1234]-linux` for consistency with `x86_64-linux`?

danieldk:
I have updated the PR to use `x86_64-v[1234]-linux` (with an underscore) to be in line with `x86_64-linux`.

sternenseemann:
> Maybe these should be system types instead of features.

I don't think that's a good idea, as the resulting system “doubles” are incompatible with any established target string system: Neither autotools nor LLVM track the x86_64 microarchitecture level in their triples (even though the latter even has a sub arch type concept), leading to all sorts of problems:

* Autotools will parse a string of the form `x86_64-v2-linux` effectively the same as `x86_64-linux`, since `v2` will be considered the vendor part.
* Some preliminary testing (with `--print-effective-triple`) seems to suggest that LLVM can also not be motivated to parse it correctly, I think you are supposed to pass this info via `-march=`. Seemingly the `v2` ends up either in the `vendor` or in the `environment` part. In any case, their triple class [can't represent](https://llvm.org/doxygen/classllvm_1_1Triple.html#a9ffca842bbaefcf99484f59a83b618d4) the microarchitecture.

Effectively, this means that in order for nixpkgs to support these microarchitecture levels, we consciously diverge from autotools (which is our source of truth at the moment, although we diverge in some cases, but accidentally). This makes `system` and `config` strings even more ambiguous which is problematic for parsing. Additionally, there would not really be any benefit, since we couldn't pass these triples on to `configure` scripts and expect them to do the right thing, rather we'd need to specifically persuade or dissuade them from using the respective features. Autotools for examples seems to ship [macros for detecting certain x86 features](https://www.gnu.org/software/autoconf-archive/ax_check_x86_features.html).

I think reverting these to `features` would be better. nixpkgs could use the platform attribute sets to represent the relevant information and add a convenient way of passing in microarchitecture level. We already have support for tracking some instruction set features (e.g. `sse3`, `avx2` etc.).

>  Because if a package is compiled for x86-64-v3, you don't want to install it on a machine that doesn't support those instructions.

I don't think we check the derivation's `system` anywhere when *installing*? Setting `system` in the derivation just prevents you from building it on a machine which would also be achieved by features. Checking `system` on install would also make cross-compiled derivations uninstallable.

Ericson2314:
Agree with everything @sternenseemann wrote.

amjoseph-nixpkgs:
Likewise agree with all of @sternenseemann's comments.

One thing that is helpful to keep in mind: the purpose of gnu triples (and llvm triples) is to tell you *when linking will succeed*:

- If two binaries were built for the same host triple, it should always be possible to link them.
- (In theory) new canonical triples should be introduced only when binaries built for one cannot be linked with binaries built for the other.

This is why cpu features like MMX, SSE, AVX, etc don't affect the host triple.  32-bit ARM is unusual: they changed the ABI when hardware floating point was added; this is why you see things like arm-*-gnueabi**hf**.

PR #4566, opened 2021-02-21
Add support for bare git repositories when using git+file

orbekk:
Local git repositories must be cloned if they are bare repositories. This pull request adds a check to treat bare repos as non-local to force a clone to happen.

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -153,12 +153,14 @@ struct GitInputScheme : InputScheme
 
     std::pair<bool, std::string> getActualUrl(const Input & input) const
     {
-        // Don't clone file:// URIs (but otherwise treat them the
-        // same as remote URIs, i.e. don't use the working tree or
-        // HEAD).
+        // file:// URIs are normally not cloned (but otherwise treated the
+        // same as remote URIs, i.e. we don't use the working tree or
+        // HEAD). Exception: If _NIX_FORCE_HTTP is set, or the repo is a bare git
+        // repo, treat as a remote URI to force a clone.
         static bool forceHttp = getEnv("_NIX_FORCE_HTTP") == "1"; // for testing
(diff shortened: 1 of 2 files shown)


src/libfetchers/git.cc
```diff
@@ -153,12 +159,13 @@ struct GitInputScheme : InputScheme
 
     std::pair<bool, std::string> getActualUrl(const Input & input) const
     {
-        // Don't clone file:// URIs (but otherwise treat them the
+        // file:// URIs are normally not cloned (but otherwise treat them the
         // same as remote URIs, i.e. don't use the working tree or
```
regnat, line 163:
```suggestion
        // file:// URIs are normally not cloned (but otherwise treated the
        // same as remote URIs, i.e. we don't use the working tree or
```

regnat (approved):

src/libfetchers/git.cc
orbekk: Thanks, suggestion applied.

src/libfetchers/git.cc
```diff
@@ -23,6 +23,12 @@ static bool isNotDotGitDirectory(const Path & path)
     return not std::regex_match(path, gitDirRegex);
 }
 
+static bool isBareGitRepository(const Path & path)
+{
+    auto output = runProgram("git", true, { "-C", path, "rev-parse", "--is-bare-repository" });
```
edolstra, line 28: Calling an external program seems a bit expensive. We can also just do `return !pathExists(path + "/.git")` right?

orbekk: That seems reasonable, patch updated. Just note that this will have false positives on paths that are not git repos.

orbekk:
@edolstra Just a reminder in case you missed it. I updated the commits to avoid spawning git.

edolstra:
Thanks!

Issue #4568, opened 2021-02-22
Cannot sign store path that contains a `flake.nix`


stephank:
**Describe the bug**

It appears `nix store sign` (and probably other similar commands) treats paths in `/nix/store` that contain a `flake.nix` as a flake, and try to evaluate it. This also happens when these commands visit such a path while traversing dependencies of some other path.

I've found specific cases where this breaks signing for us, but it may also have undesirable influence on general behaviour.

**Steps To Reproduce**

We have a flake that roughly does this:

```nix
{
  outputs = { self, nixpkgs }:
  with nixpkgs.lib;
  let
    # Subject is an example, in practice we use an attribute set with a bunch of parameters here.
    mkEnv = subject: {
      # Simple build to simulate `flake.nix` in the output.
      defaultPackage = mapAttrs (_: pkgs:
        pkgs.runCommand "greet" { } ''
          cp -r '${self}' $out
          chmod 0755 $out
          echo 'Hello ${subject}' > $out/greeting
        ''
      ) nixpkgs.legacyPackages;
    };
  in
    (mkEnv "world") // {
      inherit mkEnv;
    };
}
```

The purpose of this is that the flake exports both packages and modules, which are instantiated per environment (staging / production). We do a 'default' set of exports to make stuff like `nix build` work for testing, which is an instance with `subject == "world"` in the example above, but this default is never actually deployed. The actual deployments might do something like `subflake.mkEnv "staging"`.

Important in this example is that the store output contains the source tree with its `flake.nix`, and that the `defaultPackage` for the flake will not exist in the store at all, because the dependant will only ever call `mkEnv` with some other parameter.

To replicate the issue, create a directory structure:

- `a/flake.nix`, where the contents is the above example
- `b/flake.nix`, which will simulate our deployment flake that uses `a`. Contents are: (make sure to tweak the path)

```nix
{
  inputs.subflake.url = "/path/to/a";
  outputs = { self, subflake }: {
    defaultPackage = (subflake.mkEnv "foobar").defaultPackage;
  };
}
```

Also generate a signing key:

```sh
# nix-store --generate-binary-cache-key dummy priv pub
```

Now build and try to sign:

```
# nix build ./b
warning: creating lock file '/Users/stephank/foo/b/flake.lock'
# ls -l result
lrwxr-xr-x 1 stephank staff 49 Feb 22 16:20 result -> /nix/store/wzgkz3jpiy4jalb46dzzp4xmgn99xq4v-greet
[...]

PR #4569, opened 2021-02-22
Try parse installables as store paths first

stephank:
Potential fix for #4568.

This simply flips the order of tests for flake / store path in `SourceExprCommand::parseInstallables`. There was some extra logic to prioritise the parseFlake exception, but I don't believe we need it anymore.

I was a little surprised this passed all tests. I'm not sure how much confidence that gives us, but figured I'd open a PR to discuss. Please consider I'm not very familiar with this code or the impact of this change.

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -647,41 +647,22 @@ std::vector<std::shared_ptr<Installable>> SourceExprCommand::parseInstallables(
     } else {
 
         for (auto & s : ss) {
-            std::exception_ptr ex;
-
-            try {
-                auto [flakeRef, fragment] = parseFlakeRefWithFragment(s, absPath("."));
-                result.push_back(std::make_shared<InstallableFlake>(
-                        this,
-                        getEvalState(),
-                        std::move(flakeRef),
-                        fragment == "" ? getDefaultFlakeAttrPaths() : Strings{fragment},
-                        getDefaultFlakeAttrPathPrefixes(),
-                        lockFlags));
-                continue;
-            } catch (...) {
-                ex = std::current_exception();
-            }
-
             if (s.find('/') != std::string::npos) {
                 try {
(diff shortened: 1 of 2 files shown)


stephank:
For what it's worth, we do flake builds exclusively, and have a moderately complex Node.js / PHP application that we build with Nix. I deployed `stephank:master` to our build server, which is a little behind `NixOS:master` but includes this patch, and was able to successfully build and deploy just now.

src/libcmd/installables.cc
```diff
@@ -647,41 +647,22 @@ std::vector<std::shared_ptr<Installable>> SourceExprCommand::parseInstallables(
...
-            } catch (...) {
-                ex = std::current_exception();
-            }
-
             if (s.find('/') != std::string::npos) {
                 try {
                     result.push_back(std::make_shared<InstallableStorePath>(store, store->followLinksToStorePath(s)));
                     continue;
-                } catch (BadStorePath &) {
                 } catch (...) {
```
regnat, line 654: I think we shouldn’t drop every exception like that, but rather keep the old logic of storing the exception (unless it’s a `BadStorePath`) and eventually throwing it if both parsing methods failed

tests/flakes.sh
```diff
@@ -722,3 +723,8 @@ git -C $flakeB commit -a -m 'Foo'
 
 # Test list-inputs with circular dependencies
 nix flake metadata $flakeA
+
+# Test flake in store does not evaluate
```
regnat, line 727: The comment probably isn’t really clear out-of-context. Something like below might be clearer

```suggestion
# A store path containing a `flake.nix` should be treated as a store path, and not a flake
```

Also it might be worth adding a test to ensure that for example `file:/nix/store/foobar` will be treated as a flake url and not a store path

regnat (changes requested):
Thanks for the PR. That makes a lot of sense IMHO. Just left a couple suggestions to make things cleaner

PR #4570, opened 2021-02-23
Split {,local-}derivation-goal.{cc,hh} 

Ericson2314:
This separates the scheduling logic (including simple hook pathway) from the local-store needing code.

This should be the final split for now. I'm reasonably happy with how it's turning out, even before I'm done moving code into `local-derivation-goal`. Benefits:

1. This will help "witness" that the hook case is indeed a lot simpler, and also compensate for the increased complexity that comes from content-addressed derivation outputs.
2. It also moves us ever so slightly towards a world where we could use off-the-shelf storage or sandboxing, since `local-derivation-goal` would be gutted in those cases, but `derivation-goal` should remain nearly the same.

N.B.
```sh
git diff HEAD^^^ --color-moved --find-copies-harder --patience --stat
```
makes nicer output.

CC @regnat


src/libstore/build/local-derivation-goal.cc
```diff
@@ -182,7 +184,15 @@ inline bool DerivationGoal::needsHashRewrite()
 }
 
 
-void DerivationGoal::killChild()
+LocalStore & LocalDerivationGoal::getLocalStore()
+{
+    auto p = dynamic_cast<LocalStore *>(&worker.store);
```
regnat, line 122: Rather than this specialized method, maybe we could have a method `Store & DerivationGoal::getStore()` that would be overriden as `LocalStore & DerivationGoal::getStore()`

src/libstore/build/derivation-goal.hh
```diff
@@ -280,57 +166,37 @@ struct DerivationGoal : public Goal
...
-    void deleteTmpDir(bool force);
+    /* Close the read side of the logger pipe. */
+    virtual void closeReadPipes();
+
+    /* Cleanup hooks for buildDone() */
+    virtual void cleanupHook0();
+    virtual void cleanupHook1();
+    virtual bool cleanupHook2();
+    virtual void cleanupHook3();
+    virtual void cleanupHook4();
```
regnat, line 197: That's quite an ugly set of names :wink:  Maybe rename them to something slightly more telling like:

```cpp
    virtual void clearMountNamespace();
    virtual void stopRecursiveNix();
    virtual bool cleanFailedBuild();
    virtual void cleanBuildDirectory();
    virtual void cleanSuccessfulBuild();
```

regnat (changes requested):

src/libstore/build/derivation-goal.hh
Ericson2314: Yes those are ugly indeed :). I did choose some different names though, so the `LocalDerivationGoal`-specific details didn't leak into the interface so much.

src/libstore/build/local-derivation-goal.cc
Ericson2314: Errr I'm bit a confused, you wan to replace all the `worker.store` with `getStore()`? This is method right now is just a regular one on the subclass, and not an override of anything on the base class.

src/libstore/build/worker.cc
```diff
@@ -59,17 +59,21 @@ std::shared_ptr<DerivationGoal> Worker::makeDerivationGoalCommon(
 std::shared_ptr<DerivationGoal> Worker::makeDerivationGoal(const StorePath & drvPath,
     const StringSet & wantedOutputs, BuildMode buildMode)
 {
-    return makeDerivationGoalCommon(drvPath, wantedOutputs, [&]() {
-        return std::make_shared<DerivationGoal>(drvPath, wantedOutputs, *this, buildMode);
+    return makeDerivationGoalCommon(drvPath, wantedOutputs, [&]() -> std::shared_ptr<DerivationGoal> {
+        return !dynamic_cast<LocalStore *>(&store)
```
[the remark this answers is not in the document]
Ericson2314: Hmm. Yes downcasting has a bad rep, but:

1. The casting here is not partial, at least.
2. Keeping it as is keeps any of the `Goal` stuff from leaking into the `Store` interface. It is just an implementation detail today, and the remote stores avoid it entirely.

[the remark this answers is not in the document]
regnat: Mh you're right, the interface creep argument is definitely a good one, that's just some dirty internal cuisine that shouldn't leak.

Ericson2314:
Is this test failure related?

regnat:
> Is this test failure related?

Doesn't look like it. I've restarted the job

PR #4571, opened 2021-02-23
Make `DerivationGoal::drv` a full Derivation

### LINKED ISSUE #4559: De-duplicate the computation of the static output hashes in the drv goal
_See https://github.com/NixOS/nix/pull/4340#discussion_r578999812_

regnat:
This field used to be a `BasicDerivation`, but this `BasicDerivation` was downcasted to a `Derivation` when needed (implicitely or not), so we might as well make it a full `Derivation` and upcast it when needed.

This also allows getting rid of a weird duplication in the way we compute the static output hashes for the derivation. We had to do it differently and in a different place depending on whether the derivation was a full derivation or just a basic drv, but we can now do it unconditionally on the full derivation.

Fix #4559


Ericson2314 (approved):
Looks good!

PR #4573, opened 2021-02-24
Explicitly set default branch name for git to suppress "git hint"

oxalica:
This fixes the "git hint" printed when running `nix flake update` to fetch a git flake input.
```
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint:   git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint:   git branch -m <name>
```

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -13,6 +13,12 @@ using namespace std::string_literals;
 
 namespace nix::fetchers {
 
+// Explicit initial branch of our bare repo to suppress warnings from new version of git.
+// The value itself does not matter, since we always fetch a specific revision or branch.
+// It is set with `-c init.defaultBranch=` instead of `--initial-branch=` to stay compatible with
+// old version of git, which will ignore unrecognized `-c` options.
+const std::string gitInitialBranch = "__nix_dummy_branch";
+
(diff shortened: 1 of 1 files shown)


oxalica:
Renamed to `__nix_dummy_branch` since the initial branch of our bare repo actually does not matter.
The reason why using `-c init.defaultBranch=<name>` instead of `--initial-branch=<name>` is to make it compatible with old version of  git.

Artturin (approved):

roberth:
> Renamed to `__nix_dummy_branch` since the initial branch of our bare repo actually does not matter.
> The reason why using `-c init.defaultBranch=<name>` instead of `--initial-branch=<name>` is to make it compatible with old version of git.

Could you add something like this as an inline comment?

oxalica:
Comments are added. Also rebased.

roberth (approved):

edolstra:
Thanks!

PR #4574, opened 2021-02-25
libstore: support passing a builder's public SSH host key

grahamc:
This is already used by Hydra, and is very useful when materializing
a remote builder list from service discovery. This allows the service
discovery tool to only sync one file instead of two.

### DIFF
--- a/src/libstore/machines.cc
+++ b/src/libstore/machines.cc
@@ -54,9 +54,15 @@ ref<Store> Machine::openStore() const {
     if (hasPrefix(storeUri, "ssh://")) {
         storeParams["max-connections"] = "1";
         storeParams["log-fd"] = "4";
+    }
+
+    if (hasPrefix(storeUri, "ssh://") || hasPrefix(storeUri, "ssh-ng://")) {
         if (sshKey != "")
             storeParams["ssh-key"] = sshKey;
+        if (sshPublicHostKey != "")
+            storeParams["base64-ssh-public-host-key"] = sshPublicHostKey;
     }
+
     {
         auto & fs = storeParams["system-features"];
         auto append = [&](auto feats) {
--- a/src/libstore/ssh.cc
+++ b/src/libstore/ssh.cc
@@ -2,24 +2,37 @@
 
 namespace nix {
 
-SSHMaster::SSHMaster(const std::string & host, const std::string & keyFile, bool useMaster, bool compress, int logFD)
+SSHMaster::SSHMaster(const std::string & host, const std::string & keyFile, const std::string & sshPublicHostKey, bool useMaster, bool compress, int logFD)
     : host(host)
     , fakeSSH(host == "localhost")
     , keyFile(keyFile)
+    , sshPublicHostKey(sshPublicHostKey)
     , useMaster(useMaster && !fakeSSH)
     , compress(compress)
     , logFD(logFD)
 {
     if (host == "" || hasPrefix(host, "-"))
         throw Error("invalid SSH host name '%s'", host);
+
+    auto state(state_.lock());
+    state->tmpDir = std::make_unique<AutoDelete>(createTempDir("", "nix", true, true, 0700));
 }
 
 void SSHMaster::addCommonSSHOpts(Strings & args)
 {
+    auto state(state_.lock());
+
     for (auto & i : tokenizeString<Strings>(getEnv("NIX_SSHOPTS").value_or("")))
         args.push_back(i);
     if (!keyFile.empty())
         args.insert(args.end(), {"-i", keyFile});
+    if (!sshPublicHostKey.empty()) {
+        Path fileName = (Path) *state->tmpDir + "/host-key";
+        auto p = host.rfind("@");
+        string thost = p != string::npos ? string(host, p + 1) : host;
+        writeFile(fileName, thost + " " + base64Decode(sshPublicHostKey) + "\n");
+        args.insert(args.end(), {"-oUserKnownHostsFile=" + fileName});
+    }
     if (compress)
         args.push_back("-C");
 }
@@ -87,7 +100,6 @@ Path SSHMaster::startMaster()
 
     if (state->sshMaster != -1) return state->socketPath;
 
-    state->tmpDir = std::make_unique<AutoDelete>(createTempDir("", "nix", true, true, 0700));
 
     state->socketPath = (Path) *state->tmpDir + "/ssh.sock";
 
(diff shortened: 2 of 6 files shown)


src/libstore/machines.cc
```diff
@@ -56,6 +56,8 @@ ref<Store> Machine::openStore() const {
         storeParams["log-fd"] = "4";
         if (sshKey != "")
             storeParams["ssh-key"] = sshKey;
+        if (sshPublicHostKey != "")
+            storeParams["base64-ssh-public-host-key"] = sshPublicHostKey;
```
grahamc, line 60: Maybe both of these should be passed if the storeUri is ssh-ng as well?

src/libstore/ssh.cc
```diff
@@ -16,10 +17,21 @@ SSHMaster::SSHMaster(const std::string & host, const std::string & keyFile, bool
...
+    auto state(state_.lock());
+    if (!state->tmpDir)
+        state->tmpDir = std::make_unique<AutoDelete>(createTempDir("", "nix", true, true, 0700));
+
     for (auto & i : tokenizeString<Strings>(getEnv("NIX_SSHOPTS").value_or("")))
         args.push_back(i);
     if (!keyFile.empty())
         args.insert(args.end(), {"-i", keyFile});
+    if (!sshPublicHostKey.empty()) {
+        Path fileName = (Path) *state->tmpDir + "/host-key";
```
Infinisil, line 29: Small thing: I think `static_cast<Path>(...)` is preferred in C++ over the C-style `(Path) ...` casts. Though for `Path` specifically, all other parts of the codebase use the C-style one already.

Also TIL that you can define castings in C++ like https://github.com/NixOS/nix/blob/199081ad00e6ee4c704eaac34211b454fe0f310c/src/libutil/util.hh#L184

regnat: Unless I'm missing something, this cast is just a nasty way of accessing the (otherwise private) `state->tmpDir->path` field − by abusing the fact that as it's the first member of the class it has the same address as the class itself (so a `static_cast` wouldn't work because you're not supposed to be safely able to do that. And a `dynamic_cast` would fail too).

> Also TIL that you can define castings in C++ like
> ```
> operator Path() const { return path; }
> ```

Technically speaking that's not casting, it's just the definition of the `operator()` method which is the one called if you do `(*state->tmpDir)()`. But it looks indeed like it's the properly intended way of accessing the underlying path.

To bottom-line:

```suggestion
        Path fileName = (*state->tmpDir)() + "/host-key";
```

src/libstore/machines.cc
regnat: I think that would make sens indeed (though I guess it would be nice to move towards systematically bundling these directly in the store uri)

src/libstore/ssh.cc
```diff
@@ -16,10 +17,21 @@ SSHMaster::SSHMaster(const std::string & host, const std::string & keyFile, bool
...
+        auto p = host.find("@");
```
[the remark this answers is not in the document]
regnat: (It actually shouldn't matter too much in practice, because the regex that Nix uses to parse the url doesn't allow `@` in user names)

regnat (changes requested):
Really cool to see some hydra features backported into Nix.

If I'm not mistaken, the `sshPublicHostKey` field of the machines file isn't documented in the manual, probably worth adding an entry for it.

src/libstore/ssh.cc
Ambroisie: @regnat actually there *is* a way to define conversion operators, see [here](https://en.cppreference.com/w/cpp/language/cast_operator). Those can be used with the `*_cast<T>` familly.

regnat: I stand corrected then, looks like I was wrong all the way down.
So this is indeed a conversion operator, and although it isn't obvious from the cppreference page, `(Path)foo` actually calls `foo.operator Path()` (as shown [here](https://godbolt.org/z/qozse5)).

*(I knew I shouldn't try to pretend that I know C++ :upside_down_face: )*

src/libstore/machines.cc
grahamc: Done!

src/libstore/ssh.cc
```diff
@@ -16,10 +17,21 @@ SSHMaster::SSHMaster(const std::string & host, const std::string & keyFile, bool
 
 void SSHMaster::addCommonSSHOpts(Strings & args)
 {
+    auto state(state_.lock());
+    if (!state->tmpDir)
+        state->tmpDir = std::make_unique<AutoDelete>(createTempDir("", "nix", true, true, 0700));
```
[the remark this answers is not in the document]
grahamc: Done!

src/libstore/ssh.cc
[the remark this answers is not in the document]
grahamc: Done!

grahamc:
All set, @regnat. Want to take another look?

regnat (commented):
Looks good! Can you just add a small § to the [distributed builds docs](https://github.com/NixOS/nix/blob/master/doc/manual/src/advanced-topics/distributed-builds.md#L113)? I guess something along the lines of:

```markdown
8. The (base64-encoded) public host key of the remote machine. If omitted, SSH
    will use its regular known-hosts file.
```

grahamc:
D'oh, I knew I was forgetting something. Thanks.

grahamc:
Docs included.

regnat (approved):
:+1: 

Deferring to @edolstra

src/libstore/ssh.cc
```diff
@@ -2,24 +2,37 @@
...
     for (auto & i : tokenizeString<Strings>(getEnv("NIX_SSHOPTS").value_or("")))
         args.push_back(i);
     if (!keyFile.empty())
         args.insert(args.end(), {"-i", keyFile});
+    if (!sshPublicHostKey.empty()) {
+        Path fileName = (Path) *state->tmpDir + "/host-key";
+        auto p = host.rfind("@");
+        string thost = p != string::npos ? string(host, p + 1) : host;
+        writeFile(fileName, thost + " " + base64Decode(sshPublicHostKey) + "\n");
+        args.insert(args.end(), {"-oUserKnownHostsFile=" + fileName});
```
grahamc, line 34: Maybe these should be created at the start if it is specified, just like the tmpdir, to avoid a race condition of writing and rewriting this file a lot of times?

grahamc:
I believe the macOS failure is unrelated.

regnat:
> I believe the macOS failure is unrelated.

Yeah, it's a strange flakiness that comes up from time to time

grahamc:
I need this on 2.3 personally, so I've backported it. I could open that PR if that would be interesting: https://github.com/NixOS/nix/compare/master...grahamc:builder-host-key-stable

grahamc:
Oh, great! Thank you a lot!

PR #4576, opened 2021-02-25
Only cache successfully extracted tarballs

### LINKED ISSUE #4533: Nix cached truncated files downloaded
**Describe the bug**

In cases of some network error or proxy server error, downloading is ended too early and gives a truncated file. But the truncated file is still cached by nix, leading to immediate `truncated gzip input` every next time without refetching.

This issue happens because:
1. The response of github tarball url does **NOT** contain `Content-Length`, which makes `curl` not possible to validate the output length.
2. Some network error like proxy server being killed, will not RESET the connection. This makes `curl` returns zero but gives truncated file.
2. Nix caches the output after `curl` returns zero, but before extracting. It results in cached truncated files.

**Steps To Reproduce**

1. Setup a proxy server.
2. Run `nix-build -A osu-lazer https://github.com/r-ryantm/nixpkgs/archive/cc77f910fc1a06b7cb7eb43639c5904540483c70.tar.gz` or with other github archive URL.
3. During the download, kill the proxy server. `nix` will fail with `truncated gzip input`
4. Run the command again, it immediately fails with `truncated gzip input` again without network access.

image

**Expected behavior**
Nix caches the downloaded file only if it can be unpacked successfully.
So it can re-download the file and re-build in the next time.

`nix-env (Nix) 2.4pre20201205_a5d85d0`

oxalica:
Fixes #4533

### DIFF
--- a/src/libfetchers/fetchers.hh
+++ b/src/libfetchers/fetchers.hh
@@ -143,7 +143,8 @@ DownloadFileResult downloadFile(
     const std::string & url,
     const std::string & name,
     bool immutable,
-    const Headers & headers = {});
+    const Headers & headers = {},
+    bool addToCache = true);
 
 std::pair<Tree, time_t> downloadTarball(
     ref<Store> store,
--- a/src/libfetchers/tarball.cc
+++ b/src/libfetchers/tarball.cc
@@ -9,12 +9,53 @@
 
 namespace nix::fetchers {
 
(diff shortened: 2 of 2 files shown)


oxalica:
Close since it's fixed in curl side https://github.com/curl/curl/commit/d1f40078c13e85c56332dcb7f908fe2a7b65eb22

PR #4577, opened 2021-02-25
simplify changing cachix cache for install tests

abathur:
@domenkozar  In the process of getting set up to [try out](https://github.com/abathur/nix/runs/1982035429) the new installer tests that landed in #4549 I noticed a couple of small workflow improvements:

- Make cachix cache name deterministic to avoid workflow friction (having to re-make this change every time, remember to take it back out before merge or test it on a separate branch, etc.)
    - I wanted to make this a secret to make it easily configurable, but GA apparently won't allow an output (i.e., the installerURL) to have a secret in it.
    - Instead I'm using lowercase`<username>-<repo>-install-tests` (i.e., `abathur-nix-install-tests`)
- Set CACHIX_AUTH_TOKEN in addition to CACHIX_SIGNING_KEY so that both methods are supported. It looks like cachix will try signing key first, then auth token.

Test runs in my fork for this change: 
- [successful install tests](https://github.com/abathur/nix/runs/1991061752?check_suite_focus=true)
- [skip install tests if neither secret is declared](https://github.com/abathur/nix/runs/1991311247?check_suite_focus=true) (implying the user isn't working on the installer?)

If adopted, Nix would need a new `nixos-nix-install-tests` cache.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,52 +8,62 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
-    env:
-      CACHIX_NAME: nix-ci
+
     steps:
     - uses: actions/checkout@v2.3.4
       with:
         fetch-depth: 0
     - uses: cachix/install-nix-action@v12
+    - run: echo CACHIX_NAME="$(echo $GITHUB_REPOSITORY-install-tests | tr "[A-Z]/" "[a-z]-")" >> $GITHUB_ENV
     - uses: cachix/cachix-action@v8
       with:
         name: '${{ env.CACHIX_NAME }}'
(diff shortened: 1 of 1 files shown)


abathur:
@domenkozar I've modified this after the initial plan didn't quite work out.

I wanted to just move the CACHIX_NAME into the repo secrets so that it'd be easy to set once per fork, but I didn't realize GA would literally prevent it from appearing in the output of a step (which breaks the installerURL). 

I still think there's ergonomic value in something lower-friction than having to modify the workflow, so I've fallen back on a deterministic CACHIX_NAME (`<username>-<repo>-install-tests`) so the user can just create the matching cache and add a token/key for it to the repo's secrets. I also figured out how to keep these from trying to run if the secret isn't set.

domenkozar:
Seems to work! https://github.com/NixOS/nix/runs/2087652478

PR #4583, opened 2021-02-26
Revert "Add support for building JARs from Java sources"

puckipedia:
The code has been here since 2013, but seems to have gone unused all this time.


Ericson2314 (approved):

Ericson2314:
The makefile support used to be shared with other projects. I suspect that's how it got in here.

puckipedia:
Interesting. I looked around a bit and all I could find were forks of Nix sharing the same approximate makefile structure.

Ericson2314:
Perhaps this was just internally at LogicBox. Any any event I was just repeating @edolstra told me.

edolstra:
Yes, it was used internally at LogicBlox. The `mk` directory was included (with history) from https://github.com/edolstra/make-rules/. But since that's archived we can get rid of the unnecessary stuff.

PR #4587, opened 2021-02-28
Throw error for derivation goal with bogus wanted output

Ericson2314:
I think without #4543 it's hard to hit this, but we should have this check either way.

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1274,7 +1274,9 @@ void DerivationGoal::checkPathValidity()
     // If we requested specific elements, the loop above removes all the valid
     // ones, so any that are left must be invalid.
     if (!wantedOutputsLeft.empty())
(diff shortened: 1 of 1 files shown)


regnat (approved):

src/libstore/build/derivation-goal.cc
```diff
@@ -1267,6 +1270,11 @@ void DerivationGoal::checkPathValidity()
             }
         }
     }
+    // If we requested all the outputs via the empty set, we are always fine.
+    // If we requested specific elements, the loop above removes all the valid
+    // ones, so any that are left must be invalid.
+    if (!wantedOutputsLeft.empty())
+        throw UsageError("some wanted outputs are not provided by the derivation: %s", concatStringsSep(", ", wantedOutputsLeft));
```
edolstra, line 1277: This shouldn't be a `UsageError` but a regular error. `UsageError` is for command-line arguments that are never correct. 

We should probably also show the `drvPath` here.

Ericson2314: Fixed both.

PR #4588, opened 2021-02-28
Factor out serialization for `BuildResult`

Ericson2314:
With the differences between the two protocols reconciled, I also factor out canonical serializers for `BuildResult` to be used in the latest versions of each protocol.

Depends on #6223
~~Depends on #6312~~

There is currently some bug, which is hard to debug because of the forking. Once of the dependent PRs is merged, I can rebase the other, and that will help with bisecting.


src/libstore/remote-store.cc
```diff
@@ -62,9 +62,41 @@ void write(const Store & store, Sink & out, const Realisation & realisation)
...
 void write(const Store & store, Sink & out, const DrvOutput & drvOutput)
-{ out << drvOutput.to_string(); }
+{
+    out << drvOutput.to_string();
+}
+
+
+// BuildResult marshalliig is valid for daemon protocol >= 29 && legacy ssh >= 6
+
+BuildResult read(const Store & store, Source & from, Phantom<BuildResult> _)
```
regnat, line 77: These serialization methods seem a bit too rigid. What if we need to extend the `BuildResult` type? We would have to change this “instance”, but we can't because that would break backwards-compat with older stores.

Maybe we could start embedding some json (or bson or cbor or whatever) like I've done for `Realisation` so that we have more flexibility wrt backwards-compat?
Or maybe we could change these `read` functions to also take as parameter the version of the remote store so that they can adjust how they serialize accordingly (that would just be pushing the complexity of all the `if (GET_PROTOCOL_MINOR())` into the `worker_proto` namespace, but that's still better than having these inlined I think)?

regnat (changes requested):
Thanks for catching my mistake before it gets too annoying :)

> or 28 since the other conditions used decimal.

I have to say that having the conditions in decimal but the definition of the protocol number in hexa was causing way too much computations in my brain, which is why I wrote that one in hexa (but apparently that still wasn't enough to prevent me from messing-up the numbers… )

src/libstore/remote-store.cc
Ericson2314: > Maybe we could start embedding some json (or bson or cbor or whatever) like I've done for Realisation so that we have more flexibility wrt backwards-compat?

Eh, I don't like that as a solution for back compat. It just kicks the can down the road for whatever decodes that. (I do like it for reasons of having a standardized format, though :).)

> Or maybe we could change these `read` functions to also take as parameter the version of the remote store so that they can adjust how they serialize accordingly (that would just be pushing the complexity of all the `if (GET_PROTOCOL_MINOR())` into the `worker_proto` namespace, but that's still better than having these inlined I think)?

I made branch with that. The annoying thing is that the daemon protocol and serve protocol have separate versions, and it's unclear how to deal with that.

------

In https://github.com/NixOS/nix/pull/4588 I have `Map<std::string, BuildResult>`, so using the worker protocol stuff in particular, and not just factoring out into unrelated functions with extra version params, is needed for the `Map` generics to work.

I think my preference is just to do this for now. But then change the worker proto to allow for *non*cannonical serialization with like an optional extra parameter. Then the protocol version for each can be used to choose the serialization, without given the map set etc version parameters they don't use.

regnat: > Eh, I don't like that as a solution for back compat. It just kicks the can down the road for whatever decodes that

Kind-of, yeah. But 99% of the time “whatever decodes that” is just `BuildResult read(...)` (or even one layer down a `BuildResult::fromJSON()` static method), so it stays confined. And because we can introspect what we get, we don't have to pass the protocol number everywhere anymore (I'm no expert, but I think that's what protobuf does for example, and that seem to scale pretty well).

> I do like it for reasons of having a standardized format, though :)

You're giving me second though here, because I wouldn't really like an hypothetical standardized format to be specified directly in the C++ code :thinking:. But maybe I'm just over-thinking it.

> The annoying thing is that the daemon protocol and serve protocol have separate versions, and it's unclear how to deal with that.

Oh yes :unamused:. The simpler I guess would be to align both versions, but it's not utterly pretty either…

Ericson2314: On the other hand, since `buildPaths` throws and exception unless it succeeds, maybe I should just be returning the realizations themselves, and then this problem goes away. I think maybe https://github.com/NixOS/nix/pull/4594 which I should soon finish is the first thing to consider.

Ma27:
:+1: Applying this patch fixes `got unknown message type 5 from Nix daemon`-errors for me when building code remotely using `ssh-ng`.

src/libstore/remote-store.cc
```diff
@@ -62,9 +62,41 @@ void write(const Store & store, Sink & out, const Realisation & realisation)
...
+}
+
 void write(const Store & store, Sink & out, const DrvOutput & drvOutput)
-{ out << drvOutput.to_string(); }
+{
+    out << drvOutput.to_string();
+}
+
+
+// BuildResult marshalliig is valid for daemon protocol >= 29 && legacy ssh >= 6
```
cole-h, line 75: Nit (typo):

```suggestion
// BuildResult marshalling is valid for daemon protocol >= 29 && legacy ssh >= 6
```

Ericson2314:
OK! Relying on the new #6223, this no longer constrains the evolution of the code.

Ericson2314:
Rebased.

Ericson2314:
Because the other dependent PR is merged, #6223 now just contain this while being linear. This PR is no longer needed.

PR #4589, opened 2021-03-01
Make `nix build --json` return output paths in floating CA case

Ericson2314:
The libnixcmd build function now will fill in the missing output paths
in the buildables it works with. The JSON code, and possibly other code,
then just does the right thing.

depends on #4588

### DIFF
--- a/tests/build-floating-ca-output.sh
+++ b/tests/build-floating-ca-output.sh
@@ -1,6 +1,10 @@
 source common.sh
 
-nix build -f multiple-outputs.nix --json a.all b.all | jq --exit-status '
+sed -i 's/experimental-features .*/& ca-derivations/' "$NIX_CONF_DIR"/nix.conf
+
+nix build -f multiple-outputs.nix --arg floatingCA true --json a.all b.all | jq
+
+nix build -f multiple-outputs.nix --arg floatingCA true --json a.all b.all | jq --exit-status '
   (.[0] |
     (.drvPath | match(".*multiple-outputs-a.drv")) and
     (.outputs |
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -716,10 +716,40 @@ Buildables build(ref<Store> store, Realise mode,
         }
     }
 
+    std::map<StorePath, BuildResult> buildRes;
+
     if (mode == Realise::Nothing)
         printMissing(store, pathsToBuild, lvlError);
     else if (mode == Realise::Outputs)
-        store->buildPaths(pathsToBuild, bMode);
+        buildRes = store->buildPaths(pathsToBuild, bMode);
+
+    if (settings.isExperimentalFeatureEnabled("ca-derivations"))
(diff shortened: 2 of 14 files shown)


tests/build-floating-ca-output.sh
```diff
@@ -0,0 +1,16 @@
+source common.sh
+
+sed -i 's/experimental-features .*/& ca-derivations/' "$NIX_CONF_DIR"/nix.conf
+
+nix build -f multiple-outputs.nix --arg floatingCA true --json a.all b.all | jq
```
regnat, line 5: Is this first call intended?

regnat (commented):
To make the types more explicit (and also prevent a couple of unneeded back-and-forths), could it make sense to make `build` return a vector of `RealisedPath`?

tests/build-floating-ca-output.sh
Ericson2314: Oops good catch, that was just while debugging.

Ericson2314:
This is now just a test, and ready to go.

edolstra:
Regarding `new-cli`, there are already a lot of tests of the `nix` command in the top-level tests directory, so this could get confusing.

tests/new-cli/build/common.sh
```diff
@@ -0,0 +1 @@
+source $(dirname "${BASH_SOURCE[0]}")/../common.sh
```
SuperSandro2000, line 1: Wouldn't it be easier to directly source this file?

Ericson2314:
@edolstra would you prefer I flatten this, or put those in the subdir?

edolstra:
@Ericson2314 Yes please. `nix build` tests can go in `tests/build.sh` (which already has a `nix build` test), and CA tests can go in `tests/ca`.

tests/build-delete.sh
```diff
@@ -0,0 +1,56 @@
+source common.sh
+
+clearStore
+
+set -o pipefail
+
+# https://github.com/NixOS/nix/issues/6572
```
Ericson2314, line 7: These tests do not yet work with CA derivations, so I pulled them out into their own file.

tests/output-normalization.sh
```diff
@@ -0,0 +1,9 @@
+source common.sh
+
+testNormalization () {
```
Ericson2314, line 3: This test also do not yet work with CA derivations, so I pulled it out into its own file.

tests/ca/new-build-cmd.sh
```diff
@@ -0,0 +1,5 @@
+source common.sh
+
+export NIX_TESTS_CA_BY_DEFAULT=1
+cd ..
+source ./build.sh
```
Ericson2314, line 5: This is us running the (remaining) `tests/build.sh` tests, except with CA derivations enabled by default. This shows that this functionality works with or with or without CA derivations.

tomberek (approved):

PR #4590, opened 2021-03-01
Move the CA tests to a sub-directory

regnat:
Requires a slight update to the test infra to work properly, but
having the possibility to group tests that way makes the whole thing
quite cleaner imho


regnat:
I guess the fact that the CI doesn't run is related to https://www.githubstatus.com/incidents/xn0sd2x4nd7f

PR #4593, opened 2021-03-01
(Backport #4574) distributed builds: load remote builder host key from the machines file

grahamc:
Opening the backport in case you think it is a good idea.

From the commit:

This is already used by Hydra, and is very useful when materializing
a remote builder list from service discovery. This allows the service
discovery tool to only sync one file instead of two.


tomberek:
still needed? okay to close?

Ericson2314:
I wouldn't mind just merging it, either is fine I'd think.

Ericson2314:
There is some interest in a new 2.3 point release (#9244) so I think this is good.

Issue #4598, opened 2021-03-03
nix-shell -i exits with wrong code when interpreter is invalid


lovesegfault:
**Describe the bug**
When using `nix-shell` in a shebang, and specifying an interpreter which does not exist in `PATH`, the script will halt with an error, but will exit with code `0`. This leads callers to interpret it as successful, even though that is not the case.

**Steps To Reproduce**
1. Write an `example.py` file:
```
#!/usr/bin/env nix-shell
#!nix-shell --pure -p perl -i python

print("foo")
```
2. `./example.py`
3. `echo "$?"`
4. Verify that it says `0`, meaning it succeeded, despite having failed

**Expected behavior**

For `example.py` to exit with non-zero code.

**`nix-env --version` output**
`nix-env (Nix) 2.4pre20201205_a5d85d0`

**Additional context**

PR #4601, opened 2021-03-03
nix-build: check that envCommand exists

lovesegfault:
When starting a nix-shell with `-i` it was previously possible for it to
silently fail in the scenario where the specified interpreter didn't
exist. This happened due to the `exec` call masking the issue.

With this change we check that the interpreter that will be exec'd
exists before attempting to `exec` it, and erroring if it doesn't.

Fixes: #4598

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -447,6 +447,7 @@ static void main_nix_build(int argc, char * * argv)
                 "unset NIX_ENFORCE_PURITY; "
                 "shopt -u nullglob; "
                 "unset TZ; %6%"
+                "shopt -s execfail;"
                 "%7%",
                 shellEscape(tmpDir),
                 (pure ? "" : "p=$PATH; "),


src/nix-build/nix-build.cc
```diff
@@ -447,7 +447,9 @@ static void main_nix_build(int argc, char * * argv)
                 "unset NIX_ENFORCE_PURITY; "
                 "shopt -u nullglob; "
                 "unset TZ; %6%"
-                "%7%",
+                "envCommand=(%7%);"
+                "[ ! -x \"${envCommand[1]}\" ] && exit 1;"
```
samueldr, line 451:
```suggestion
                "[ ! -x \"${envCommand[1]}\" ] && exit 127;"
```

127 is commonly used in these situations.

lovesegfault: Fixed, thanks!

cpcloud:
Should this have a test that asserts failure on a bogus interpreter?

matthewbauer:
Shouldn't exec give the 127 exit code?

```
$ bash -c 'exec abc'; echo $?
bash: line 0: exec: abc: not found
127
```

I think this problem is related to the "_nix_shell_clean_tmpdir" trap. It doesn't propagate error codes at all. We would need something like https://stackoverflow.com/a/25064409.

cpcloud:
Adapting the SO answer to this case IIUC:

`test.sh`:

```sh
#!/usr/bin/env bash

set -eu

die() {
    echo "Dying!!"
    exit "$1"

}

trap 'die $?' EXIT

exec abc
```

Run `test.sh`:

```sh
❯ ./test.sh; echo $?
./test.sh: line 13: exec: abc: not found
Dying!!
0
```

cpcloud:
Wrapping the final `exec abc` in `bash -c` (`bash -c 'exec abc'`) will of course propagate the `exec abc` code because `bash` is invoked as a child process

matthewbauer:
Hmm... something with trap & exec seems to get converted to exit 0. If you remove the trap it works as you would expect.

It looks like you can set `execfail` to make this work (without the $? too):

```
#!/usr/bin/env bash

shopt -s execfail

die() {
    echo "Dying!!"
}

trap die EXIT

exec abc
```

lovesegfault:
@matthewbauer That's a way better way of doing this! Updated the PR.

lovesegfault:
Pinging @edolstra for review

PR #4603, opened 2021-03-04
improve man page for nix.conf (builders)

DavHau:
#### Motivation
How to exactly specify builders in nix.conf was unclear to me. The existing docs mention the format of `nix.machines`, but there is no such option in nixos. Probably the `nix.buildMachines` option is meant here, but this option is an attribute set, vs. the nix.conf expects a string. How to map the attribute set to the string format is unclear as well.

#### Done
Improve the help text for the builders options. Specify the exact format and provide two examples, a minimal one and one that uses the full spec.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -207,15 +207,8 @@ public:
     Setting<std::string> builders{
         this, "@" + nixConfDir + "/machines", "builders",
         R"(
-          A semicolon-separated list of build machines, where each machine follows this format:
-
-          {protocol}://{user}@{host} [{comma sep. systems} - {maxJobs} {speedFactor} {comma sep. features}]
-
-          Examples:
-          
-          ssh://root@builder1.com
-          
-          ssh://root@builder2.com x86_64-linux,aarch64-linux - 40 20 nixos-test,benchmark,big-parallel,kvm
+          A semicolon-separated list of build machines.
+          For the exact format and examples, see [the manual chapter on remote builds](../advanced-topics/distributed-builds.md)
         )"};
 
     Setting<bool> buildersUseSubstitutes{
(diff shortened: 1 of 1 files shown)


src/libstore/globals.hh
```diff
@@ -206,7 +206,17 @@ public:
 
     Setting<std::string> builders{
         this, "@" + nixConfDir + "/machines", "builders",
-        "A semicolon-separated list of build machines, in the format of `nix.machines`."};
+        R"(
+          A semicolon-separated list of build machines, where each machine follows this format:
+
+          {protocol}://{user}@{host} [{comma sep. systems} - {maxJobs} {speedFactor} {comma sep. features}]
```
domenkozar, line 212: there's also ssh public key as of recently

src/libstore/globals.hh
```diff
@@ -206,7 +206,17 @@ public:
...
+
+          Examples:
+          
+          ssh://root@builder1.com
+          
+          ssh://root@builder2.com x86_64-linux,aarch64-linux - 40 20 nixos-test,benchmark,big-parallel,kvm
```
edolstra, line 218: This should be formatted as markdown. Also it might be better to refer to the chapter on remote building in the manual rather than duplicate that information here.

DavHau: Oh, I wasn't aware that there is a manual chapter about this. Referencing it sounds like a good Idea. Is there any way to ensure that the exact version of the manual is referenced which matches the current nix version?

edolstra: Something like `[the manual chapter on remote builds](../advanced-topics/distributed-builds.md)` should work in the mdbook manual.

DavHau: Thanks. I did that now. Afterwards I checked the resulting mdbook and the link works.

Issue #1184, opened 2017-01-11
"Dumping very large path" message isn't helpful


copumpkin:
It would be helpful to know which path is being dumped and ideally what led to the dumpage. Given [where the warning is emitted right now in the code](https://github.com/NixOS/nix/blob/master/src/libutil/serialise.cc#L62), it's not obvious to me how to add that detail to the message, but it's pretty unfriendly right now. I guess we could add a constructor parameter to an `FdSink` that tells it some human-meaningful representation of the thing being written to it?

cc @edolstra @shlevy

shlevy:
Yeah, a message in FdSink constructor seems the way to go.

shlevy:
(well, actually, using constant space for dumping is obviously the way to go :stuck_out_tongue: )

copumpkin:
https://github.com/NixOS/nix/pull/619

domenkozar:
@edolstra how come 01d33eab235b0f7ef770c0168eba9c795772be4d never landed in master?

domenkozar:
@edolstra would you mind explaining what's keeping those two commits back?

Those are two of the top annoying bugs I encounter using Nix and even being experienced, they are annoying to overcome.

I think it's better to make it work&correct and we can fix the performance one day if needed.

zimbatm:
This is still an issue today. I just got that error message and don't know what path is causing the issue.

domenkozar:
Same here today, nix-daemon ate all my 32G of memory and I have no idea what path caused it.

Ericson2314:
I think the prereq patch has landed on master in some form? This might also be straight up fixed on master too.

aleb:
https://github.com/NixOS/nix/commit/cf4a7ae26de6d8869856d3beddfe8eb2440578a3 landed as https://github.com/NixOS/nix/commit/34f25124ba2ab32b8a95d6b37cd68d7bb85ff2d4

Any chance it can be backported to the 2.3-maintenance branch used for making the 2.3.* releases https://github.com/NixOS/nix/blob/2.3-maintenance/src/libstore/local-store.cc ?

The https://github.com/NixOS/nix/pull/2981 can be closed as it's forgotten and this different fix has been merged instead.

pingiun:
this isn't stale, as the message is still shown and still not helpful

PR #4605, opened 2021-03-04
1184 large path error [WIP]

bburdette:
This is an exploratory PR which touches the Sink and Source objects and their descendants.  

The idea is to address issue #1184, the vague "large path" error.  The error we get now is this:

```
[nix-shell:~/code/nix-error-project/1184/test-case]$ ./build.sh 
warning: unknown setting 'experimental-features'
^Cerror: interrupted by the user
```

Whereas with these changes its this:

```
[nix-shell:~/code/nix-error-project/1184/test-case]$ ./newbuild.sh 
path: /home/bburdette/code/zknotes/zknotes/server
```

The main idea is pretty simple, add a `source_identifier` parameter to the Sink::operator(), and add a corresponding source_identifier string to Source:

```
struct Sink
{
    virtual ~Sink() { }                                 // NEW PARAMETER  ↓
    virtual void operator () (std::string_view data, std::string_view source_identifier) = 0;   
    virtual bool good() { return true; }
};
```

```
/* Abstract source of binary data. */
struct Source
{
    virtual ~Source() { }

    /* Store exactly ‘len’ bytes in the buffer pointed to by ‘data’.
       It blocks until all the requested data is available, or throws
       an error if it is not going to be available.   */
    void operator () (char * data, size_t len);

    /* Store up to ‘len’ in the buffer pointed to by ‘data’, and
       return the number of bytes stored.  It blocks until at least
       one byte is available. */
    virtual size_t read(char * data, size_t len) = 0;

    virtual bool good() { return true; }

    std::string source_identifier;                     // NEW STRING 

    void drainInto(Sink & sink);

    std::string drain();
};
```

With these elements in place, its possible to issue a more specific error message:

```
static void warnLargeDump(std::string_view source_identifier)
{
    warn("dumping very large path (> 256 MiB); this may run out of memory\npath: %s", source_identifier);
}

void FdSink::write(std::string_view data, std::string_view source_identifier)
{
    written += data.size();
    static bool warned = false;
    if (warn && !warned) {
        if (written > threshold) {
            warnLargeDump(source_identifier);
            warned = true;
        }
    }
    try {
        writeFull(fd, data);
    } catch (SysError & e) {
        _good = false;
        throw;
    }
}
```

So that's it.  But as you'd expect, there's a lot of code that gets disturbed as a result here.  And, there are are fair number of decisions to be made about whether to put something in these source_identifier strings, and if so what to put there.  

Furthermore, there's a certain reduction in api elegance, for instance with flush you might have something like this:

```
    // are we adding the string 'pathname' to the output here?   no, but looks kind of like we are
   outstream.flush(pathname);    
```

At any rate, before sinking a whole lot more hours into putting reasonable `source_identifier` strings everywhere, I'd gauge the acceptability of this idea in general.  Thanks for any feedback!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,3 +32,5 @@ endif
 include mk/lib.mk
 
 GLOBAL_CXXFLAGS += -g -Wall -include config.h -std=c++17
+
+LDFLAGS += -ldl
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -165,7 +165,6 @@ ref<const ValidPathInfo> BinaryCacheStore::addToStoreCommon(
     TeeSource teeSource { narSource, teeSinkUncompressed };
     narAccessor = makeNarAccessor(teeSource);
     compressionSink->finish();
-    // fileSink.flush();
     fileSink.flush(narSource.source_identifier);
     }
 
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -165,7 +165,8 @@ ref<const ValidPathInfo> BinaryCacheStore::addToStoreCommon(
     TeeSource teeSource { narSource, teeSinkUncompressed };
     narAccessor = makeNarAccessor(teeSource);
     compressionSink->finish();
-    fileSink.flush();
+    // fileSink.flush();
+    fileSink.flush(narSource.source_identifier);
     }
 
     auto now2 = std::chrono::steady_clock::now();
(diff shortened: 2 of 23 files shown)


edolstra:
Instead of passing down a source_identifier (which is a bit ad hoc), can't we use `Error.addTrace` for this? Similar to what I did in 0187838e2e7ff01f1b480e3e85d9e96da0b4b78e to provide more info about the origin of "unexpected EOF reading a line" errors.

edolstra:
Maybe a nicer solution: create an `Activity` when we copy a path to the store. That way it shows up in the progress bar and the user can see if a path is taking a long time to copy. The progress bar can even show how many bytes have been copied.

bburdette:
That sounds much easier!  I'll definitely check in to that.

Ericson2314:
It was my understanding we are almost always doing things in constant space now? I would just like to get rid of this failure mode altogether!

bburdette:
going with the Activity method, way less intrusive!  

new PR:  https://github.com/NixOS/nix/pull/4608

PR #4608, opened 2021-03-05
add path to 'very large path' error

bburdette:
This is to address issue #1184, the vague 'dumping very large path' warning.  Rather than changing the warning message itself, we're creating an Activity when we copy a path, similar to what happens in [store-api.cc](https://github.com/NixOS/nix/blob/52b6e0f83746997f7223a69482ee0eda4ebbc2ff/src/libstore/store-api.cc#L743).

Here's the output with my test case:

```
[nix-shell:~/code/nix-error-project/1184/test-case]$ ./newbuild.sh 
copying path '/home/bburdette/code/zknotes/zknotes/server'...
^Cerror: interrupted by the user
```

I put in an `isInStore` check before adding the Activity, otherwise there's quite a bit of extra output:

```
[nix-shell:~/code/nix-error-project/1184/test-case]$ ./newbuild.sh 
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/stdenv/generic/default-builder.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/stdenv/linux/bootstrap-tools/scripts/unpack-bootstrap-tools.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/stdenv/generic/builder.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/build-support/setup-hooks/move-docs.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/build-support/setup-hooks/make-symlinks-relative.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/build-support/setup-hooks/compress-man-pages.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/build-support/setup-hooks/strip.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/build-support/setup-hooks/patch-shebangs.sh'...

                      <a few hundred more lines of these>

copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/build-support/setup-d
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/development/compilers/rust/setup-hook.sh'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/build-support/rust/patch-registry-deps'...
copying path '/nix/store/rrhid0i04x3v4majkx9sjzg6jrpxl3hz-nixos-20.09.2787.6e7f25001fe/nixos/pkgs/development/compilers/llvm/10/llvm-outputs.patch'...
copying path '/home/bburdette/code/zknotes/zknotes/server'...
```

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1842,7 +1842,7 @@ string EvalState::copyPathToStore(PathSet & context, const Path & path)
               // show paths we're copying, unless they're already in the store.
               std::unique_ptr<PushActivity> pact;
               auto path_to_add = checkSourcePath(path);
-              if (!store->isInStore(path)) {
+              if (!store->isInStore(path) || repair) {
                 Activity act(*logger, lvlInfo, actCopyPath, fmt("copying path '%s'", path_to_add));
                 pact = std::unique_ptr<PushActivity>(new PushActivity(act.id));
               }
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1838,7 +1838,18 @@ string EvalState::copyPathToStore(PathSet & context, const Path & path)
     else {
         auto p = settings.readOnlyMode
             ? store->computeStorePathForPath(std::string(baseNameOf(path)), checkSourcePath(path)).first
-            : store->addToStore(std::string(baseNameOf(path)), checkSourcePath(path), FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
+            : [&] {
+              // show paths we're copying, unless they're already in the store.
+              if (!store->isInStore(path)) {
+                auto path_to_add = checkSourcePath(path);
+                Activity act(*logger, lvlInfo, actCopyPath, fmt("copying path '%s'", path_to_add));
+                PushActivity pact(act.id);
+                return store->addToStore(std::string(baseNameOf(path)), path_to_add, FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
+              }
+              else
+                return store->addToStore(std::string(baseNameOf(path)), checkSourcePath(path), FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
+            } ();
+
         dstPath = store->printStorePath(p);
         srcToStore.insert_or_assign(path, std::move(p));
         printMsg(lvlChatty, "copied source '%1%' -> '%2%'", path, dstPath);
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1840,14 +1840,13 @@ string EvalState::copyPathToStore(PathSet & context, const Path & path)
             ? store->computeStorePathForPath(std::string(baseNameOf(path)), checkSourcePath(path)).first
             : [&] {
               // show paths we're copying, unless they're already in the store.
+              std::unique_ptr<PushActivity> pact;
+              auto path_to_add = checkSourcePath(path);
               if (!store->isInStore(path)) {
-                auto path_to_add = checkSourcePath(path);
                 Activity act(*logger, lvlInfo, actCopyPath, fmt("copying path '%s'", path_to_add));
-                PushActivity pact(act.id);
-                return store->addToStore(std::string(baseNameOf(path)), path_to_add, FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
+                pact = std::unique_ptr<PushActivity>(new PushActivity(act.id));
               }
-              else
-                return store->addToStore(std::string(baseNameOf(path)), checkSourcePath(path), FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
+              return store->addToStore(std::string(baseNameOf(path)), path_to_add, FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
             } ();
 
         dstPath = store->printStorePath(p);


src/libexpr/eval.cc
```diff
@@ -1838,7 +1838,18 @@ string EvalState::copyPathToStore(PathSet & context, const Path & path)
...
+              // show paths we're copying, unless they're already in the store.
+              if (!store->isInStore(path)) {
+                auto path_to_add = checkSourcePath(path);
+                Activity act(*logger, lvlInfo, actCopyPath, fmt("copying path '%s'", path_to_add));
+                PushActivity pact(act.id);
+                return store->addToStore(std::string(baseNameOf(path)), path_to_add, FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
+              }
+              else
+                return store->addToStore(std::string(baseNameOf(path)), checkSourcePath(path), FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
+            } ();
```
Ericson2314, line 1851: why not avoid duplication with?
```suggestion
            : [&] {
              // show paths we're copying, unless they're already in the store.
              auto path_to_add = checkSourcePath(path);
              if (!store->isInStore(path)) {
                Activity act(*logger, lvlInfo, actCopyPath, fmt("copying path '%s'", path_to_add));
                PushActivity pact(act.id);
              }
              return store->addToStore(std::string(baseNameOf(path)), path_to_add, FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
            } ();
```
Also, maybe we need to include `repair` in the condition?

bburdette: That would be good, except as I understand it the Activity is supposed to exist while the addToStore is happening.  With this variation, the PushActivity goes out of scope before the addToStore call, and when it does it terminates the activity in its destructor.  This means the 'Activity' is over before the addToStore starts.

bburdette: Another possibility would be with a unique_ptr, like this:

```
        auto p = settings.readOnlyMode
            ? store->computeStorePathForPath(std::string(baseNameOf(path)), checkSourcePath(path)).first
            : [&] {
              // show paths we're copying, unless they're already in the store.
              std::unique_ptr<PushActivity> pact;
              auto path_to_add = checkSourcePath(path);
              if (!store->isInStore(path)) {
                Activity act(*logger, lvlInfo, actCopyPath, fmt("copying path '%s'", path_to_add));
                pact = std::unique_ptr<PushActivity>(new PushActivity(act.id));
              }
              return store->addToStore(std::string(baseNameOf(path)), path_to_add, FileIngestionMethod::Recursive, htSHA256, defaultPathFilter, repair);
            } ();
```

re 'repair' that makes sense to me.  I'd want to see repair paths being copied.  At present I don't know how to tell the difference between these and the fusillade of copyPath messages I got in the above example.

Ekleog:
Interesting, I hadn't been aware of this PR when opening https://github.com/NixOS/nix/pull/4625! Guess that's an alternative, and both ways make reasonable sense to me — I kind of like #4625 just because it adds metadata, thus hopefully opening the way for other improved diagnostics, but OTOH it's quite a bit more code so I guess some balance will have to be found :)

The only thing I think is missing from here is the fact that currently the logic is to display the warning only once. I changed said logic in [this part](https://github.com/NixOS/nix/pull/4625/files#diff-48d402eeed8594649798f9f983f89d98d79cbf661f0ab23802544db45c7d8ca6L62), [this part](https://github.com/NixOS/nix/pull/4625/files#diff-48d402eeed8594649798f9f983f89d98d79cbf661f0ab23802544db45c7d8ca6L397), [this part](https://github.com/NixOS/nix/pull/4625/files#diff-fa972d8bbb8a4a7d6426b3733e72a808bd903d9482bd5b69866189af16d31145R106) and [this part](https://github.com/NixOS/nix/pull/4625/files#diff-fa972d8bbb8a4a7d6426b3733e72a808bd903d9482bd5b69866189af16d31145R171) of my PR if I'm not missing any.

domenkozar:
@Ekleog Have you also seen the discussion at https://github.com/NixOS/nix/pull/4605

src/libexpr/eval.cc
bburdette: took the unique_ptr approach to reduce duplication; added the repair flag.

domenkozar:
@edolstra what's missing here?

domenkozar:
@edolstra would be great to know if this is a good way forward to debug these warnings.

longregen:
It would be really great to have some way to debug this. I'm not a seasoned NixOS user, and I was trying to install it on a 16GB machine, and I can't tell what is making the `nixos-install` procedure to fail.

jceb:
I an into the same error message - `fatal flex scanner internal error--end of buffer missed` shortly after I started the installation of NixOS via `nixos-install`. As a workaround, I massively increased swap space. Currently, the installation is using ~40 GB of RAM in total. A bit excessive :exploding_head: 

image

jceb:
Passing the option `--no-channel-copy` to `nixos-install` solved the issue for me.

bburdette:
@Ericson2314 I'm happy to get this branch working again, if it looks like its something that's desired.

Ericson2314:
@bburdette Yeah maybe. But perhaps it would be better to put the activities in the underlying implementations?

Ericson2314:
I think this is no longer needed because `InputAccessor::fetchToStore` does this: https://github.com/NixOS/nix/blob/02bd821f2e71372d31bbe6700bd68086cc2ee70a/src/libfetchers/input-accessor.cc#L36

PR #4618, opened 2021-03-08
Sign the derivation outputs

### LINKED ISSUE #4248: Add a signature to the output mappings for ca derivations
For input-addressed derivation outputs, the critical thing that has to be trusted is the content of the output path, so it is the part that has to be signed when fetching from a binary cache.

However, for content-addressed derivation outputs, the content can be checked using the store path (as the store path is essentially a hash of its content).
However the link between the symbolic derivation output `(drvPath, outputName)` and the output store path has to be trusted, so we should sign it on binary caches and allow checking this signature locally like we do for the narinfos.

regnat:
Add a `signature` field to the derivation outputs, and sign them (the same way we sign path-infos/nars).

Fix #4248


Ericson2314 (approved):
Did just a quick read through, but looks good.

PR #4619, opened 2021-03-08
Fix gcc10 build again

jonringer:
lowdown_term_rndr returns void, yet it was being assigned.
```
src/libcmd/markdown.cc: In function 'std::string nix::renderMarkdownToTerminal(std::string_view)':
src/libcmd/markdown.cc:43:37: error: void value not ignored as it ought to be
   43 |     int rndr_res = lowdown_term_rndr(buf, nullptr, renderer, node);
```

### DIFF
--- a/src/libcmd/markdown.cc
+++ b/src/libcmd/markdown.cc
@@ -40,9 +40,7 @@ std::string renderMarkdownToTerminal(std::string_view markdown)
         throw Error("cannot allocate Markdown output buffer");
     Finally freeBuffer([&]() { lowdown_buf_free(buf); });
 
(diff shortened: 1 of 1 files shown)


Mic92 (approved):

edolstra:
Maybe you're compiling against an older version of lowdown? On 0.8.0 / master, [it returns an int](https://github.com/kristapsdz/lowdown/blob/148f9b2f586c41b7e36e73009db43ea68c7a1a4d/lowdown.h#L401-L403).

jonringer:
I think I was inheriting what ever was in nixpkgs. I'll close as this doesn't seem to be an issue any longer

Issue #4034, opened 2020-09-18
Warn when adding large local paths


roberth:
**Is your feature request related to a problem? Please describe.**

The `dumping large paths` message has been slightly helpful by providing a hint that something big is getting added to the store each time.
This message is disappearing with #4030 (unless you use an older daemon), so you'll be filling up your store with slight variations of large paths if you aren't careful. Virtually nobody checks the size of their derivation closures.

**Describe the solution you'd like**

Provide a warning when adding a large local path to the store, when it is
 - not the output of an evaluation time fetcher with a content hash, because these tend to be unavoidable and tend to change less often
 - larger than a threshold that defaults to, say, 250 MB
 - is not a single file (which would seem rather intentional)

Example warning

```
You're copying to the Nix store from path `/home/user/workspace` which seems to be rather large (> 100MB). In most cases this happens by accident because the path wasn't filtered correctly. Alternatively, you may want to split your build into smaller derivations if possible, or set `warn-large-path-threshold = 250 #MB` to a larger value.
```

**Describe alternatives you've considered**

Nix flakes do alleviate the problem by only including (potentially unstaged!) files that are listed in git's staging area.
However, user can still reference paths that aren't in a repo.

**Additional context**

cole-h:
The "you'll be filling up your store with slight variations of large paths if you aren't careful" makes me think the overhaul should re-implement this warning before it could be considered merge-worthy. Since we currently have a warning for this, I feel it would be a step back to have it removed with no alternative.

I also don't think the "is not a single file" criterion is a great idea to not warn on -- it might be intentional, but it also might not. Maybe there are files `a1` and `A1`: `A1` is 100G, while `a1` is 100K. Holding shift slightly too long, and now Nix is trying to add 100G to the store. A contrived example, but I hope you get my point.

Sorry for being so negative 😬

PR #4625, opened 2021-03-09
Make the “dumping large path” warning include information on what path is being large

Ekleog:
Disclaimer: this is my first time seriously modifying nix's source code, so I'm far from sure I did everything right. However, I do think that the added debug information is an incremental improvement, and can then be refined as we go, eg. if it turns out some of the debug labels are not precise enough.

Example output with the change:
```
[nix-shell:~/prog/nix]$ outputs/out/bin/nix-instantiate -E '"${./.}"'
  while dumping from source 'path:/home/ekleog/prog/nix'
```

What do you think? :)

### DIFF
--- a/perl/lib/Nix/Store.xs
+++ b/perl/lib/Nix/Store.xs
@@ -183,7 +183,7 @@ void exportPaths(int fd, ...)
 void importPaths(int fd, int dontCheckSigs)
     PPCODE:
         try {
-            FdSource source(fd);
+            FdSource source(fd, "");
             store()->importPaths(source, dontCheckSigs ? NoCheckSigs : CheckSigs);
         } catch (Error & e) {
             croak("%s", e.what());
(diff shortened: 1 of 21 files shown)


cole-h:
See also https://github.com/NixOS/nix/pull/4608.

Ekleog:
Thank you! I'm going to comment on this other PR, both ways look reasonable to me. About the failed test, I _think_ it's due to a timeout that led to the test failing and thus a spurious failure, but not 100% sure about that

roberth:
The occurrence of this message should have been greatly reduced by #4030. In which situation does the message still crop up? It'd be good to make that stream too.

That said, this PR does have merit of its own when it comes to `StringSink`. The warning should probably be removed from `FdSink` though, because that sink does not have a memory problem.
#4034 does describe another use case for logic _like_ this, except it's a linter-like warning instead of "we're about to crash". That can probably be implemented instead as a pass-through sink or source that takes the original path as a constructor argument.

Ekleog:
Seeing how this is merge-conflicting, I'm going to close this as rebasing would probably be just as hard as reimplementing.

PR #4628, opened 2021-03-10
Dynamic derivations RFC 92

Ericson2314:
See the tests in https://github.com/obsidiansystems/nix/blob/dynamic-drvs/tests/dyn-drv/ for what is being implemented.

Needs:

- [x] Make cooler tests
- [x] RFC draft
- [x] Unstable feature gate to prevent accidental use
- [x] Fix Perl bindings

~~Depends on #4594~~
~~Depends on #3959~~
~~Depends on #4543~~
~~Depends on #5364~~
~~Depends on #7601~~
~~Depends on #8353~~
~~Depends on #8369~~
~~Depends on #8813~~
~~Depends on #8829~~
~~Depends on #8927~~
~~Depends on #8938~~

### DIFF
--- a/doc/manual/src/protocols/derivation-aterm.md
+++ b/doc/manual/src/protocols/derivation-aterm.md
@@ -2,8 +2,18 @@
 
 For historical reasons, [derivations](@docroot@/glossary.md#gloss-store-derivation) are stored on-disk in [ATerm](https://homepages.cwi.nl/~daybuild/daily-books/technology/aterm-guide/aterm-guide.html) format.
 
-Derivations are serialised in the following format:
+Derivations are serialised in one of the following formats:
 
-```
-Derive(...)
-```
+- ```
+  Derive(...)
+  ```
+
+  For all stable derivations.
+
+- ```
+  DrvWithVersion(<version-string>, ...)
+  ```
+
+  The only `version-string`s that are in use today are for [experimental features](@docroot@/contributing/experimental-features.md):
+
+  - `"xp-dyn-drv"` for the [`dynamic-derivations`](@docroot@/contributing/experimental-features.md#xp-feature-dynamic-derivations) experimental feature.
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -350,25 +350,37 @@ void DerivationGoal::gaveUpOnSubstitution()
 
     /* The inputs must be built before we can build this goal. */
     inputDrvOutputs.clear();
-    if (useDerivation)
-        for (auto & i : dynamic_cast<Derivation *>(drv.get())->inputDrvs) {
+    if (useDerivation) {
+        std::function<void(ref<SingleDerivedPath>, const DerivedPathMap<StringSet>::ChildNode &)> addWaiteeDerivedPath;
+
+        addWaiteeDerivedPath = [&](ref<SingleDerivedPath> inputDrv, const DerivedPathMap<StringSet>::ChildNode & inputNode) {
+            if (!inputNode.value.empty())
+                addWaitee(worker.makeGoal(
+                    DerivedPath::Built {
+                        .drvPath = inputDrv,
+                        .outputs = inputNode.value,
+                    },
+                    buildMode == bmRepair ? bmRepair : bmNormal));
+            for (const auto & [outputName, childNode] : inputNode.childMap)
+                addWaiteeDerivedPath(
+                    make_ref<SingleDerivedPath>(SingleDerivedPath::Built { inputDrv, outputName }),
+                    childNode);
+        };
+
+        for (const auto & [inputDrvPath, inputNode] : dynamic_cast<Derivation *>(drv.get())->inputDrvs.map) {
             /* Ensure that pure, non-fixed-output derivations don't
                depend on impure derivations. */
             if (experimentalFeatureSettings.isEnabled(Xp::ImpureDerivations) && drv->type().isPure() && !drv->type().isFixed()) {
-                auto inputDrv = worker.evalStore.readDerivation(i.first);
+                auto inputDrv = worker.evalStore.readDerivation(inputDrvPath);
                 if (!inputDrv.type().isPure())
                     throw Error("pure derivation '%s' depends on impure derivation '%s'",
                         worker.store.printStorePath(drvPath),
-                        worker.store.printStorePath(i.first));
+                        worker.store.printStorePath(inputDrvPath));
             }
 
-            addWaitee(worker.makeGoal(
-                DerivedPath::Built {
-                    .drvPath = makeConstantStorePathRef(i.first),
-                    .outputs = i.second,
-                },
-                buildMode == bmRepair ? bmRepair : bmNormal));
+            addWaiteeDerivedPath(makeConstantStorePathRef(inputDrvPath), inputNode);
         }
+    }
 
     /* Copy the input sources from the eval store to the build
        store. */
@@ -501,7 +513,7 @@ void DerivationGoal::inputsRealised()
                 return ia.deferred;
             },
             [&](const DerivationType::ContentAddressed & ca) {
-                return !fullDrv.inputDrvs.empty() && (
+                return !fullDrv.inputDrvs.map.empty() && (
                     ca.fixed
                     /* Can optionally resolve if fixed, which is good
                        for avoiding unnecessary rebuilds. */
@@ -515,7 +527,7 @@ void DerivationGoal::inputsRealised()
             }
         }, drvType.raw);
 
-        if (resolveDrv && !fullDrv.inputDrvs.empty()) {
+        if (resolveDrv && !fullDrv.inputDrvs.map.empty()) {
             experimentalFeatureSettings.require(Xp::CaDerivations);
 
             /* We are be able to resolve this derivation based on the
@@ -552,11 +564,13 @@ void DerivationGoal::inputsRealised()
             return;
         }
 
-        for (auto & [depDrvPath, wantedDepOutputs] : fullDrv.inputDrvs) {
+        std::function<void(const StorePath &, const DerivedPathMap<StringSet>::ChildNode &)> accumInputPaths;
+
+        accumInputPaths = [&](const StorePath & depDrvPath, const DerivedPathMap<StringSet>::ChildNode & inputNode) {
             /* Add the relevant output closures of the input derivation
                `i' as input paths.  Only add the closures of output paths
                that are specified as inputs. */
-            for (auto & j : wantedDepOutputs) {
+            auto getOutput = [&](const std::string & outputName) {
                 /* TODO (impure derivations-induced tech debt):
                    Tracking input derivation outputs statefully through the
                    goals is error prone and has led to bugs.
@@ -568,21 +582,30 @@ void DerivationGoal::inputsRealised()
                    a representation in the store, which is a usability problem
                    in itself. When implementing this logic entirely with lookups
                    make sure that they're cached. */
-                if (auto outPath = get(inputDrvOutputs, { depDrvPath, j })) {
-                    worker.store.computeFSClosure(*outPath, inputPaths);
+                if (auto outPath = get(inputDrvOutputs, { depDrvPath, outputName })) {
+                    return *outPath;
                 }
                 else {
                     auto outMap = worker.evalStore.queryDerivationOutputMap(depDrvPath);
-                    auto outMapPath = outMap.find(j);
+                    auto outMapPath = outMap.find(outputName);
                     if (outMapPath == outMap.end()) {
                         throw Error(
                             "derivation '%s' requires non-existent output '%s' from input derivation '%s'",
-                            worker.store.printStorePath(drvPath), j, worker.store.printStorePath(depDrvPath));
+                            worker.store.printStorePath(drvPath), outputName, worker.store.printStorePath(depDrvPath));
                     }
-                    worker.store.computeFSClosure(outMapPath->second, inputPaths);
+                    return outMapPath->second;
                 }
-            }
-        }
+            };
+
+            for (auto & outputName : inputNode.value)
+                worker.store.computeFSClosure(getOutput(outputName), inputPaths);
+
+            for (auto & [outputName, childNode] : inputNode.childMap)
+                accumInputPaths(getOutput(outputName), childNode);
+        };
+
+        for (auto & [depDrvPath, depNode] : fullDrv.inputDrvs.map)
+            accumInputPaths(depDrvPath, depNode);
     }
 
     /* Second, the input sources. */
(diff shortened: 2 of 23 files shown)


L-as:
How come you cancelled CI?

Ericson2314:
I didn't. If some jobs fail others are cancelled. There is one new bug relating to the eval store (I need to get crafty about which drvs are to be found where). Existing bugs were in the JSON output changing for nix build because the derivation keys might not be strings, and perhaps something else also boring.

Ericson2314:
I have rebased this. It is no failing tests, but i don't think that will be too bad to fix.

Ericson2314:
The last commit is finally ready for review!

src/libstore/derivations.cc
```diff
@@ -267,6 +267,28 @@ static DerivationOutput parseDerivationOutput(const Store & store, std::istrings
...
+static DerivedPathMap<StringSet>::ChildNode parseDerivedPathMapNode(const Store & store, std::istringstream & str)
+{
+    DerivedPathMap<StringSet>::ChildNode node;
+    node.value = parseStrings(str, false);
+    char c = str.get();
+    switch (c) {
+    case ')':
+        return node;
+    case ',':
+        expect(str, "[");
```
roberth, line 279: This would be the right thing to do in a vacuum, but wouldn't we get a slightly better error if we repurposed the `^` character here? Then old Nix would report something like `output not found 'foo^bar'`, which is a slightly better hint than something like `couldn't parse deadbeef.drv at char 123`, I think.

(And *maybe* we get an even better error thanks to `requiredFeatures` if we do that. May well be wishful thinking though; seems like a later stage...)

src/nix/app.cc
```diff
@@ -11,6 +11,25 @@
 
 namespace nix {
 
+static std::variant<StorePath, DownstreamPlaceholder> makePlaceHolder(const SingleBuiltPath & req)
```
roberth, line 14:
```suggestion
static std::variant<StorePath, DownstreamPlaceholder> makePlaceholder(const SingleBuiltPath & req)
```

roberth (commented):
First quick look. Idea for better error.
I'll need to understand more of the code.

src/libstore/derivations.cc
Ericson2314: The new error should cause it to fail with

> `expected "\""`

which is not ideal, but better than it "getting off" and trying to parse the next bit after `inputDrvs` as `inputDrvs`.

roberth: Wouldn't a format like this give an even better error, or is that not feasible?

```aterm
[
  "cat",
  "dog",
  "cat^kitten",
  "goose^gosling"
]
```

roberth: ie avoiding the `outputs` × `deeper`, but rather make use of the unused space within the individual output syntax.

Ericson2314: (recapping live discussion for others) I made the format self-versioning to try to make the error better while keeping the deduplication we currently have that would go away with the flat list of `^`

doc/manual/src/protocols/derivation-aterm.md
```diff
@@ -0,0 +1,17 @@
+For historical reasons, [derivations](@docroot@/glossary.md#gloss-store-derivation) are stored on-disk in [ATerm](https://homepages.cwi.nl/~daybuild/daily-books/technology/aterm-guide/aterm-guide.html) format.
+
+Derivations either serialized in one of the following formats:
```
fricklerhandwerk, line 3:
```suggestion
Derivations are serialised in one of the following formats:
```

We try to use BE consistently.

doc/manual/src/protocols/derivation-aterm.md
```diff
@@ -0,0 +1,17 @@
+For historical reasons, [derivations](@docroot@/glossary.md#gloss-store-derivation) are stored on-disk in [ATerm](https://homepages.cwi.nl/~daybuild/daily-books/technology/aterm-guide/aterm-guide.html) format.
```
fricklerhandwerk, line 1:
```suggestion
# ATerm Derivations

For historical reasons, [derivations](@docroot@/glossary.md#gloss-store-derivation) are stored on-disk in [ATerm](https://homepages.cwi.nl/~daybuild/daily-books/technology/aterm-guide/aterm-guide.html) format.
```

doc/manual/src/protocols/derivation-aterm.md
```diff
@@ -0,0 +1,17 @@
...
+  Derive(...)
+  ```
+
+  For all stable derivations.
+
+- ```
+  Derive(<version-string>, ...)
+  ```
+
+  The only `version-string`s that are in use today are for [exerimental features]((@docroot@/contributing/experimental-features.md):
```
fricklerhandwerk, line 15:
```suggestion
  The only `version-string`s that are in use today are for [exerimental features](@docroot@/contributing/experimental-features.md):
```

doc/manual/src/protocols/derivation-aterm.md
Ericson2314: Ah and what I wrote was just pain ungrammatical too whoops.

src/libstore/build/derivation-goal.cc
```diff
@@ -1475,22 +1499,24 @@ void DerivationGoal::waiteeDone(GoalPtr waitee, ExitCode result)
...
-    if (!dg) return;
+    std::optional info = tryGetConcreteDrvGoal(waitee);
+    if (!info) return;
+    const auto & [dg, drvReq] = *info;
 
-    auto outputs = fullDrv.inputDrvs.find(dg->drvPath);
-    if (outputs == fullDrv.inputDrvs.end()) return;
+    auto & outputs = fullDrv.inputDrvs.ensureSlot(drvReq.get()).value;
+    // FIXME: need a `findSlot`
+    //if (outputs == fullDrv.inputDrvs.end()) return;
```
roberth, line 1508: Any reason not to implement this?

src/libstore/derivations.cc
```diff
@@ -256,29 +260,90 @@ static DerivationOutput parseDerivationOutput(const Store & store,
...
+{
+    DerivedPathMap<StringSet>::ChildNode node;
+    switch (str.peek()) {
+    case '[':
+        node.value = parseStrings(str, false);
+        break;
+    case '(':
+        // Older derivation should never use new form, but newer
+        // derivaiton can use old form.
+        assert(version == DerivationATermVersion::DynamicDerivations);
```
roberth, line 300: This is not an internal invariant, because we're not guaranteed to get `.drv`s that this Nix version wrote.
Hence this should be a proper parsing exception that briefly explains the issue.

src/libstore/derivations.cc
```diff
@@ -256,29 +260,90 @@ static DerivationOutput parseDerivationOutput(const Store & store,
...
+    switch (str.peek()) {
+    case '(':
+        version = DerivationATermVersion::Traditional;
+        expect(str, "(");
+        break;
+    case 'V':
+        version = DerivationATermVersion::DynamicDerivations;
+        xpSettings.require(Xp::DynamicDerivations);
+        expect(str, "V(");
+        expect(str, "xp-dyn-drv"); // Only verison we have so far
```
roberth, line 337:
```suggestion
        expect(str, "xp-dyn-drv"); // Only version we have so far
```

I use T-mobile

src/libstore/tests/derivation.cc
```diff
@@ -235,11 +244,105 @@ TEST_JSON(simple,
...
+      ],
+      "env": {
+        "BIG_BAD": "WOLF"
+      },
+      "outputs": {}
+    })",
+    makeDynDepDerivation(*store))
+
+TEST_ATERM(DynDerivationTest, dynDerivationDeps,
+    R"(DeriveV(xp-dyn-drv,[],[("/nix/store/c015dhfh5l0lp6wxyvdn7bmwhbbr6hr9-dep2.drv",(["cat","dog"],[("cat",["kitten"]),("goose",["gosling"])]))],["/nix/store/c015dhfh5l0lp6wxyvdn7bmwhbbr6hr9-dep1"],"wasm-sel4","foo",["bar","baz"],[("BIG_BAD","WOLF")]))",
```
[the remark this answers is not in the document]
Ericson2314: Ah I think I know what to do (Make sure the prefix is not `Derive`), and I'll manually test it that way too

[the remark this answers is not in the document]
Ericson2314:
```
 $ nix --extra-experimental-features nix-command store add-file /tmp/nix-shell.k33Z5K/nix-test/tests/dyn-drv/dep-built-drv/store/p8m8wa846ym3k6qzjr97lcfzm4psbmgm-use-dynamic-drv-in-non-dynamic-drv.drv --store $PWD/foo
error: error parsing derivation '/nix/store/9hn00i9kgvx3fxa0hdmbqiha0g7qad86-p8m8wa846ym3k6qzjr97lcfzm4psbmgm-use-dynamic-drv-in-non-dynamic-drv.drv': error: expected string 'Derive(['
```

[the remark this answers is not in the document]
Ericson2314:
```
 $ ./outputs/out/bin/nix --extra-experimental-features nix-command store add-file /tmp/nix-
shell.k33Z5K/nix-test/tests/dyn-drv/dep-built-drv/store/p8m8wa846ym3k6qzjr97lcfzm4psbmgm-use-dynamic-drv-in-non-dynamic-drv.drv --store $PWD/foo
error: error parsing derivation '/nix/store/9hn00i9kgvx3fxa0hdmbqiha0g7qad86-p8m8wa846ym3k6qzjr97lcfzm4psbmgm-use-dynamic-drv-in-non-dynamic-drv.drv': error: expected string 'Derive([', got 'DrvVrs("'
```
Better thing I'll PR in a moment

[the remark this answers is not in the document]
Ericson2314: https://github.com/NixOS/nix/pull/8938

doc/manual/src/protocols/derivation-aterm.md
```diff
@@ -2,8 +2,18 @@
...
-Derive(...)
-```
+- ```
+  Derive(...)
+  ```
+
+  For all stable derivations.
+
+- ```
+  Derive(<version-string>, ...)
```
roberth, line 14:
```suggestion
  DrvWithVersion(<version-string>, ...)
```

src/libstore/derivations.cc
```diff
@@ -376,14 +469,52 @@ static void printUnquotedStrings(std::string & res, ForwardIterator i, ForwardIt
...
     s.reserve(65536);
-    s += "Derive([";
+
+    /* Use older unversioned form if possible, for wider compat. Use
+       newer form only if we need it, which we do for
+       `Xp::DynamicDerivations`. */
+    if (std::find_if(
+            inputDrvs.map.begin(),
+            inputDrvs.map.end(),
+            [](auto & kv) { return !kv.second.childMap.empty(); })
```
roberth, line 505: Could you add a Derivation method for this?

src/libstore/misc.cc
```diff
@@ -322,24 +334,43 @@ std::map<DrvOutput, StorePath> drvOutputReferences(
...
         }
-    }
+        if (!inputNode.value.empty()) {
+            auto d = makeConstantStorePathRef(inputDrv);
+            for (const auto & [outputName, childNode] : inputNode.childMap) {
+                SingleDerivedPath next = SingleDerivedPath::Built { d, outputName };
+                accumRealisations(
+                    // FIXME: Someday have "deep" realizations analogous to
+                    // deep placeholders, rather than just straight-up
+                    // resolving the computed drv to a store path.
```
roberth, line 365:
```suggestion
```

This is not urgent, so best to put this in an issue and reference it from the tracking issue.

src/libstore/remote-store.cc
```diff
@@ -172,7 +172,21 @@ void RemoteStore::ConnectionHandle::processStderr(Sink * sink, Source * source,
     auto ex = handle->processStderr(sink, source, flush);
     if (ex) {
         daemonException = true;
-        std::rethrow_exception(ex);
+        try {
+            std::rethrow_exception(ex);
+        } catch (const Error & e) {
+            // Ugly backwards compatibility hack.
```
roberth, line 178:
```suggestion
            // Nix versions before #4628 did not have an adequate behavior for reporting that the derivation format was upgraded.
            // To avoid having to add compatibility logic in many places, we expect to catch almost all occurrences of the
            // old incomprehensible error here, so that we can explain to users what's going on when their daemon is
            // older than #4628 (2023).
```

src/libstore/misc.cc
```diff
@@ -125,14 +125,26 @@ void Store::queryMissing(const std::vector<DerivedPath> & targets,
 
     std::function<void(DerivedPath)> doPath;
 
+    std::function<void(ref<SingleDerivedPath>, const DerivedPathMap<StringSet>::ChildNode &)> accumDerivedPath;
+
+    accumDerivedPath = [&](ref<SingleDerivedPath> inputDrv, const DerivedPathMap<StringSet>::ChildNode & inputNode) {
```
roberth, line 130:
```suggestion
    enqueueDerivedPaths = [&](ref<SingleDerivedPath> inputDrv, const DerivedPathMap<StringSet>::ChildNode & inputNode) {
```

src/libstore/build/derivation-goal.cc
```diff
@@ -552,11 +565,13 @@ void DerivationGoal::inputsRealised()
             return;
         }
 
-        for (auto & [depDrvPath, wantedDepOutputs] : fullDrv.inputDrvs) {
+        std::function<void(const StorePath &, const DerivedPathMap<StringSet>::ChildNode &)> accumDerivedPath;
+
+        accumDerivedPath = [&](const StorePath & depDrvPath, const DerivedPathMap<StringSet>::ChildNode & inputNode) {
```
roberth, line 570:
```suggestion
        accumInputPaths = [&](const StorePath & depDrvPath, const DerivedPathMap<StringSet>::ChildNode & inputNode) {
```

roberth (commented):
Minor suggestions :rocket:

src/libstore/misc.cc
Ericson2314:
```suggestion
                    // TODO deep resolutions for dynamic derivations, issue #8947, would go here.
```

Ericson2314: #8947

Ericson2314: I left a reference, just to make this all less cryptic :). Let me know if that is OK.

src/libstore/derivations.cc
Ericson2314: I made a static function, since nothing else needs it yet. Did you just want it pulled out for legibility, or did you want something publicly in the header in particular?

roberth (approved):

roberth:
That's an MVP for ya!

You'll want to use this with `recursive-nix` for trying this out with non-trivial builds (writing not a singular drv), but be aware that the interface may change. We want to offer a safer, more limited recursive nix functionality by default, so that a graph of drvs can be constructed, sources can be generated, etc, but that's not been built yet and may or may not use the existing daemon protocol. Again this is experimental. ymmv.

:rocket:

domenkozar:
Is there any documentation how to use dynamic derivations?

Ericson2314:
Not enough, that is definitely next on the agenda.

Some parts like the [`outputOf` builtin](https://hydra.nixos.org/build/234737970/download/1/manual/language/builtins.html#builtins-outputOf) are documented, but other parts are not (like "text" content addressing derivation outputs) because the original concept was already undocumented! And there is definitely no tutorial yet on putting all the pieces together.

For now, I would look at the [tests](https://github.com/NixOS/nix/tree/master/tests/dyn-drv) and imitate what they do. Hopefully that is not too cumbersome.

Ericson2314:
Yes. Because derivations themselves are always content addressed, any derivation that produces derivations must be derivation which opts to content-address its outputs.

PR #4646, opened 2021-03-16
Fix Nix to properly work with stores using a scoped IPv6 address

### LINKED ISSUE #4490: Link-local address store broke recently
Link-local used to work in storl uri's. Ex
`nix copy --to ssh://root@fe80::818c:da4d:8975:415c%enp0s25 ./result`
Which is a quick way to force-push nix drvs in-extremis.

Recently (sometime in the last 2-3 months, which was the last time nix was updated), this feature broke, and now returns the error 
`don't know how to open Nix store 'ssh://root@fe80::818c:da4d:8975:415c%enp0s25'`

Variants were tried with square brackets, to no avail.

Edit: Nix built from rev 1acbb61696c118712417bcd1c59021cc84650e16 (11 days old rev)

Ma27:
According to RFC4007[1], IPv6 addresses can have a so-called zone_id
separated from the actual address with `%` as delimiter. In contrast to
Nix 2.3, the version on `master` doesn't recognize it as such:

    $ nix ping-store --store ssh://root@fe80::1%18 --experimental-features nix-command
    warning: 'ping-store' is a deprecated alias for 'store ping'
    error: --- Error ----------------------------------------------------------------- nix
    don't know how to open Nix store 'ssh://root@fe80::1%18'

I modified the IPv6 match-regex accordingly to optionally detect this
part of the address. As we don't seem to do anything special with it, I
decided to leave it as part of the URL for now.

Fixes #4490

[1] https://tools.ietf.org/html/rfc4007

-----

@maisiliym  can you check whether this commit fixes the problem for you?

cc @edolstra @blaggacao

### DIFF
--- a/src/libutil/tests/url.cc
+++ b/src/libutil/tests/url.cc
@@ -117,6 +117,24 @@ namespace nix {
         ASSERT_EQ(parsed, expected);
     }
 
+    TEST(parseURL, parseScopedRFC4007IPv6Address) {
+        auto s = "http://[fe80::818c:da4d:8975:415c\%enp0s25]:8080";
+        auto parsed = parseURL(s);
+
+        ParsedURL expected {
+            .url = "http://[fe80::818c:da4d:8975:415c\%enp0s25]:8080",
+            .base = "http://[fe80::818c:da4d:8975:415c\%enp0s25]:8080",
+            .scheme = "http",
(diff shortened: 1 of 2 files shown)


roberth (approved):
Looks like it conforms to the rfc.

(urls are such a gift, mumble mumble)

PR #4650, opened 2021-03-17
Use the appropriate config dir for the registry

samueldr:
It was reported on IRC that Nix would create a `.config` dir even though  `XDG_CONFIG_HOME` was set.

This is the only occurence of `.config` I could grep for that wasn't in `getConfigDir()`.

The test is a bit... ugly... And might be missing some other actions that could create the config dir...

But otherwise, it **does** test this particular case. Reverting the fix (tip of this branch) makes it fail as expected.

I guess we want to test other commands that can write to the config directory, are there others?

We probably also want to test the equivalent for the different caches.

### DIFF
--- a/tests/config.sh
+++ b/tests/config.sh
@@ -1,5 +1,31 @@
 source common.sh
 
+# Isolate the home for this test.
+# Other tests (e.g. flake registry tests) could be writing to $HOME in parallel.
+export HOME=$TEST_ROOT/userhome
+
+# Test that using XDG_CONFIG_HOME works
+# Assert the config folder didn't exist initially.
+[ ! -e "$HOME/.config" ]
+# Without XDG_CONFIG_HOME, creates $HOME/.config
+unset XDG_CONFIG_HOME
+# Run against the nix registry to create the config dir
+# (Tip: this relies on removing non-existent entries being a no-op!)
+nix registry remove userhome-without-xdg
+# Verifies it created it
+[ -e "$HOME/.config" ]
+# Remove the directory it created
+rm -rf "$HOME/.config"
+# Run the same test, but with XDG_CONFIG_HOME
+export XDG_CONFIG_HOME=$TEST_ROOT/confighome
+# Assert the XDG_CONFIG_HOME/nix path does not exist yet.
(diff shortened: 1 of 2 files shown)


Ericson2314 (approved):

tests/config.sh
```diff
@@ -1,10 +1,36 @@
...
+[ ! -e "/tmp/confighome/nix" ]
+nix registry remove userhome-with-xdg
+# Verifies the confighome path has been created
+[ -e "/tmp/confighome/nix" ]
+# Assert the .config folder hasn't been created.
+[ ! -e "$HOME/.config" ]
+
 # Test that files are loaded from XDG by default
-export XDG_CONFIG_HOME=/tmp/home
+export XDG_CONFIG_HOME=/tmp/confighome
```
samueldr, line 30: Yes, this export is now redundant considering the previous export.

The following tests, though, are not redundant. None are.

edolstra:
We shouldn't use fixed paths like `/tmp/home` in the tests since that's racy and breaks in multi-user environments. I see that recently got introduced in 895516cadf0739dd8359a38d503e0aaf5068dc4b but it's a mistake.

samueldr:
Totally agree that it may be racy. I even changed it to reduce raciness assuming (verified through grepping) that the path is unique between all tests.

Though I see it's not foolproof. What strategy should be used to isolate each test files?

samueldr:
Are we blocking this on the fixed paths?

If so, I'll need a suggestion to make this go forward. I'm not sure what approach could be used in these tests.

If we're not blocking on the fixed paths, anything else needs to be changed?

regnat:
> What strategy should be used to isolate each test files?

I guess using `$TEST_ROOT` is good-enough (it's doesn't really bring much more guaranty if the test is run twice concurrently, but it's what everything else uses, so at least there's only one place to change if must be)

samueldr:
Applied the fix to use `$TEST_ROOT`.

 - [x] Tested the new test still fails without the fix (the last commit)
 - [x] Tested the tests work with the fix

edolstra:
Thanks @samueldr!

PR #4654, opened 2021-03-19
Make `warn` method respect logging verbosity

CitadelCore:
Currently, the `warn` method does not respect the logging verbosity as set by the `--verbose` and `--quiet` parameters. This pull request fixes this.


edolstra:
Thanks! Turns out that the underlying problem is that `ProgressBar::log()` didn't respect the verbosity level (3e0e443181997c52b0db19ae781948c573a634dd).

Issue #2587, opened 2018-12-15
Nix multi-user install doesn't support non-login non-interactive shells on Ubuntu (which breaks distributed builds)


djahandarie:
Using the multi-user install on an Ubuntu machine `nixhost` followed by running commands of the form `ssh nixhost nix-<whatever>` result in the following error:

```
cannot build on 'ssh://XXXX': cannot connect to 'XXXX': bash: nix-store: command not found
```

This means that the host cannot be used as a `buildMachine` in a distributed build setup.

Investigation
---------------

A basic multi-user-install on Ubuntu modifies `/etc/profile.d/nix.sh`, because `/etc/bashrc` doesn't exist, only `/etc/bash.bashrc` does.

`/etc/profile`, which calls `/etc/profile.d/nix.sh`, is only called for full login shells.

This means the code is only called for full login shells (interactive or not).

Thoughts
---------------

If we also appended the nix sourcing to `/etc/bash.bashrc` (by modifying [PROFILE_TARGETS in  install-multi-user.sh](https://github.com/NixOS/nix/blob/6ba1726eeb58cd8e7d69e2a090346f26fdf634a5/scripts/install-multi-user.sh)), that would at least result in the code running on non-login *interactive* shells, but that still would not result in the code being run on non-login *non-interactive* shells, due to a check at the top of `/etc/bash.bashrc`

[An askubuntu answer](https://askubuntu.com/a/247769) suggests *prepending* to `/etc/bash.bashrc` if you want to execute commands on non-login non-interactive shells.

Questions
--------------

1. Do we want nix to be usable from non-login non-interactive shells by default?

2. Do we want to accomplish the above by prepending to `/etc/bash.bashrc`?

3. If we don't, should we consider changing the distributed builder to use full login shells (like [this superuser Q&A documents](https://superuser.com/questions/306530/run-remote-ssh-command-with-full-login-shell))?

cc @grahamc

knedlsepp:
Having the same issue on CentOS. It did however work for the root user out of the box. The issue was that `~root/.bashrc` contained the line `source /etc/bashrc`. So my personal solution is just to put the following in my personal bashrc on that server:
```bash
if [ -f /etc/bashrc ]; then
        . /etc/bashrc
fi
```

mschwaig:
I ran into this on an Ubuntu 18.04 box with a Multi-User Install of Nix when I tried to use `nix-copy-closure` or `nix copy` to copy  some store paths over there.

[...]

PR #4663, opened 2021-03-23
Added Debian-based OS's profiles

### LINKED ISSUE #4662: Fix profile sourcing on Debian-based OS's
https://github.com/NixOS/nix/blob/6749d9e0574c53d6fa94713801f6b80c224dfe31/scripts/install-multi-user.sh#L34

This line will do nothing on Debian-based OS's as the global `bash` and `zsh`-`rc`/`env` (`/etc/bashrc` and `/etc/zshrc`) are replaced respectively by `/etc/bash.bashrc` and `/etc/zsh/zshrc`, so you need to manually source `/etc/profile.d/nix.sh` in order to make Nix work

Stig124:
Fixes #4662 

Added `/etc/bash.bashrc` and `/etc/zsh/zshenv` as they replace `/etc/bashrc` and `/etc/zshenv` on Debian OS's

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -31,7 +31,7 @@ readonly NIX_FIRST_BUILD_UID="30001"
 readonly NIX_ROOT="/nix"
 readonly NIX_EXTRA_CONF=${NIX_EXTRA_CONF:-}
 
-readonly PROFILE_TARGETS=("/etc/bashrc" "/etc/profile.d/nix.sh" "/etc/zshenv")
+readonly PROFILE_TARGETS=("/etc/bashrc" "/etc/profile.d/nix.sh" "/etc/zshenv" "/etc/bash.bashrc" "/etc/zsh/zshenv")
 readonly PROFILE_BACKUP_SUFFIX=".backup-before-nix"
 readonly PROFILE_NIX_FILE="$NIX_ROOT/var/nix/profiles/default/etc/profile.d/nix-daemon.sh"
 


bjornfor:
+1.

Note that on e.g Ubuntu there is code in `/etc/bash.bashrc` to return early unless the shell is interactive. That means appending the nix code at the _end_ of the file makes `nix-copy-closure --to ubuntu-machine ...` fail out-of-the-box.

Stig124:
> +1.
> 
> Note that on e.g Ubuntu there is code in `/etc/bash.bashrc` to return early unless the shell is interactive. That means appending the nix code at the _end_ of the file makes `nix-copy-closure --to ubuntu-machine ...` fail out-of-the-box.

Then, you can use `sed` to always put that at the top of the file

Something like :

```bash
sudo sed -i '1i source /etc/profile.d/nix.sh' /etc/bash.bashrc # Would work just fine with /etc/bashrc too
```

bjornfor:
Yes. Maybe something for a follow-up PR? (Btw, I don't have commit rights to this repo.)

Stig124:
> Yes. Maybe something for a follow-up PR? (Btw, I don't have commit rights to this repo.)

I'm working on something to solve that issue

bjornfor:
> [...] makes nix-copy-closure --to ubuntu-machine ... fail out-of-the-box.

There's already an issue about that: https://github.com/NixOS/nix/issues/2587

abathur:
Hmm. 

In https://github.com/Stig124/nix/commit/c40bad415104398450866d32682011db0acb0310 I changed the installer to start touching each profile target in order to fix up #3608, which seemed to be correct but only worked if /etc/zshenv already existed.

I don't really know if there's a non-obnoxious way to address it, but the installer will create and populate all of the profile targets for all users. I'm not quite sure if this will just amount to a little littering, or if creating all of the files will mess up anyone's start-up order.

Issue #2109, opened 2018-04-25
imports don't resolve symlinks absolutely, leading to unexpected behaviour


Infinisil:
Have a look at the following directory structure:

```
.
├── 1
│   ├── content.nix                       content: "In 1"
│   └── default.nix -> ../2/default.nix
└── 2
    ├── content.nix                       content: "In 2"
    └── default.nix                       content: import ./content.nix
```

Now this is what happens with evaluation:

```
$ nix-instantiate --eval 1
"In 1"
$ nix-instantiate --eval 2
"In 2"
```

However what I **expect** to happen is
```
$ nix-instantiate --eval 1
"In 2"
$ nix-instantiate --eval 2
"In 2"
```

because the symlink should have influenced all imports as well.

This can lead to unexpected behaviour. A good example of this is if I were to have a directory like this (with `/path/to/nixpkgs` being a nixpkgs checkout):
```
.
└── nixos -> /path/to/nixpkgs/nixos
```

Then trying to use this folder will fail:

```
$ nix-instantiate nixos/release-combined.nix -A tested
error: opening file '/home/infinisil/Test/nixsym/default.nix': No such file or directory
```

dezgeg:
Another case to consider is the same directory layout as in the first example but with an explicit `/default.nix` suffix:
````
$ nix-instantiate --eval 1/default.nix
"In 2"
$ nix-instantiate --eval 2/default.nix 
"In 2"
````
i.e. the expected result. So I agree that there is a bug at least in the way that the result should be consistent.

**Comment by chreekat** [comment]
Interesting. For me, the existing behavior *is* the expected behavior. Symlinks represent real files in real directories, and shouldn't serve as portals to elsewhere. I opened #2150 because nix-shell *is* tunneling through the portal, and that surprised me.

I guess now someone gets to figure out which of us has the "wrong" intuition. :)

**Comment by kirelagin** [comment]
I actually came here to report the discrepancy that @dezgeg pointed out and before reading this issue I was on the same side as @chreekat.

But I have to admit that the nixos example is very convincing and it made me think about the standard practices of handling symlinks in terms of context. From my experiments if I do:

```
$ ln -s /bin .
$ ls bin/..
```

the output is the listing of the root directory, not of the current one, as I would expect, which means that my and @chreekat’s intuition is actually wrong and the standard practice is for symlinks to actually serve as portals 🤷‍♂️.

**Comment by ony** [comment]
it looks even weirder with

```
.
[...]

PR #4678, opened 2021-03-31
Traverse symlinked path components in import

ony:
Fix for #2109 

*Note that this will not fix case where symlink is followed by `..` (traverse level up).*


Ericson2314:
This is probably better, but also a big breaking change.

tomberek:

PR #4688, opened 2021-04-03
libcmd/installables: force re-evaluation of cached failures

Ma27:
I think that it's not very helpful to get "cached failures" in a wrong
`flake.nix`. This can be very confusing when debugging a Nix expression.
See for instance NixOS/nixpkgs#118115.

In fact, the eval cache allows a forced reevaluation which is used for
e.g. `nix eval`.

This change makes sure that this is the case for `nix build` as well. So
rather than

    λ ma27 [~/Projects/exp] → ../nix/outputs/out/bin/nix build -L --rebuild --experimental-features 'nix-command flakes'
    error: cached failure of attribute 'defaultPackage.x86_64-linux'

the evaluation of already-evaluated (and failed) attributes looks like
this now:

    λ ma27 [~/Projects/exp] → ../nix/outputs/out/bin/nix build -L --rebuild --experimental-features 'nix-command flakes'
    error: attribute 'hell' missing

           at /nix/store/mrnvi9ss8zn5wj6gpn4bcd68vbh42mfh-source/flake.nix:6:35:

                5|
                6|     packages.x86_64-linux.hello = nixpkgs.legacyPackages.x86_64-linux.hell;
                 |                                   ^
                7|

cc @edolstra

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -503,7 +503,11 @@ std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableF
     auto root = cache->getRoot();
 
     for (auto & attrPath : getActualAttrPaths()) {
-        auto attr = root->findAlongAttrPath(parseAttrPath(*state, attrPath));
+        auto attr = root->findAlongAttrPath(
+            parseAttrPath(*state, attrPath),
+            true
+        );
+
         if (!attr) continue;
 
         if (!attr->isDerivation())
--- a/src/libexpr/eval-cache.cc
+++ b/src/libexpr/eval-cache.cc
@@ -486,11 +486,11 @@ std::shared_ptr<AttrCursor> AttrCursor::getAttr(std::string_view name)
     return getAttr(root->state.symbols.create(name));
 }
 
-std::shared_ptr<AttrCursor> AttrCursor::findAlongAttrPath(const std::vector<Symbol> & attrPath)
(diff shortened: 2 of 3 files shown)


Kha:
Should `--show-trace` imply `-F`?

kisik21:
I think it should, since sometimes you use `--show-trace` to debug why the Nix expression you just launched failed, and it wouldn't be helpful if it just says "cached failure". `--no-force-reevaluation` could be used to invert that behavior (though I doubt anyone will need it)

Ma27:
> Should --show-trace imply -F?

Good idea. Would wait with any further changes though until I get feedback from @edolstra  in terms of whether or not this approach is the way to go.

> --no-force-reevaluation could be used to invert that behavior (though I doubt anyone will need it)

Yeah, I'd personally prefer the `-F` behavior to be the default, though I don't know if there were some good reasons this isn't the case yet that I'm not aware of.

The "cached eval failure" thing might be useful for e.g. CI systems if you have a relatively complex expression and want to get fast feedback (and thus no re-evaluation if it's already "known" that it won't evaluate).

edolstra:
Re-evaluation in case of a cached failure is supposed to be automatic, see 4dcb183af31d5cb33b6ef8e581e77d1c892a58b9. We shouldn't add a flag for that.

Ma27:
@edolstra as shown above, this is not the case by default for `nix build`. Should we also force errors there, then?

Ma27:
@edolstra so always forcing eval errors would be the way to go?

PR #4691, opened 2021-04-05
[configure]: Detect the JSON Library

cgohla:
This extends the `configure.ac` to check for the presence of the [JSON library](https://github.com/nlohmann/json), and inject an optional extra path into `CXXFGLAGS`.

Follow up to #4684 .

**Note**
This is a draft because it's my first try at _autoconf_ scripting, so I'm eager to get feedback about it. There might be a more idiomatic way to achieve the same outcome, in particular `AC_ARG_VAR` actually creates an output variable, which we are not using.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -311,6 +311,20 @@ fi
 AC_SUBST(tarFlags)
 
 
+# Do we have the JSON library?
+AC_ARG_VAR([nlohmann_json_include], [Path to the JSON includes])
+AC_LANG_PUSH(C++)
+    [if test -n "$nlohmann_json_include" ; then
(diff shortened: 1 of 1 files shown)


edolstra:
We should probably just include a copy of nlohmann/json again (i.e. revert 4145cd2da002e1bd8affa0392c80118eabe58e3c).

cgohla:
@edolstra I tried to solve to implement your suggestion in #4693, which seems to work, but there is a test failure on macOS that doesn't mean anything to me.

edolstra:
Closing this for #4693.

PR #4697, opened 2021-04-06
nix-rust: various optimizations

zseri:
This PR changes some internal API parts of `nix-rust`, but the observable behavoir (except perf) shouldn't change.

* omit unnecessary reallocations in Store::compute_path_closure
* use match in PathInfo::parse_nar_info
* util::base32: omit references; use u32 for nr_bits_left
* merge store/store.rs into store/mod.rs
* dependencies: use futures-util instead of futures, optional, to prevent bit-rot


Ericson2314:
To be honest, I think `nix-rust` is basically dead code?

zseri:
It is afaik basically completely dead code (`nix build -v -L` didn't mention it). It might be usable otherwise... currently, I'm trying to get the marked `#[cfg(unused)]` (basically commented out) portions of the code to work again.

asymmetric:
Does anyone what the story is behind this code?

/cc @edolstra

edolstra:
A while back we had an experiment to rewrite parts of Nix in Rust, but it was abandoned. The `nix-rust` directory is a remnant of that.

zseri:
I would be interested, too, as I may try in the future to reimplement parts of the Nix backend (e.g. store + daemon, etc.) in Rust, to adapt Nix to use cases where dealing with e.g. the Nix language itself would be unnecessary complex and high-level; others have been tried experiments with the frontend (e.g. nix-dram); and then there is interest for Nix language tooling incl. partial interpretation (for e.g. flake-generator); I think it would be a good idea to try to modularize Nix more (and maybe provide a C API for the most often needed stuff), so that shallow wrappers around parts of Nix can be build without needing to build everything from scratch.

griff:
@zseri you might be interested in [Nix.rs](https://github.com/griff/Nix.rs). A project I have been working for the last 3 weeks to port the legacy worker protocol of `nix-store --serve` and `ssh://<host>` store to Rust.

It is not on crates.io yet but it is complete enough that I have made my own toy version of nixbuild.net that stores build results in a local file store and builds in memory in a docker container with no network.

PR #4700, opened 2021-04-09
Revert #4595: libfetchers/tarball: Lock on effectiveUrl

### LINKED ISSUE #4672: flakes: Locking on effective urls breaks fetching github releases (and probably other things)
**Describe the bug**

Before https://github.com/NixOS/nix/pull/4595 was merged we used flakes to lock and fetch github releases. Github uses redirects to so-called signed urls which expire after some time. This means that any flake.lock will, after the merge of #4595, contain unusable urls.

The intention of the merged pull request seems to have been to enable locking of nixos channels which ALSO uses redirects but for a different purpose. I see how that might be useful but I still question the point of enabling the use of nixos channels when flakes are supposed to replace them in the end.

In any case - that merge broke many of of our projects so that we cannot update flake inputs anymore unless we use an older version of Nix.

**Steps To Reproduce**

1. Use a Nix version after pull request 4595 was merged
2. Add an input to a github release, for example:
```nix
{
   inputs.fluxcd-x86_64-linux = {
    url = "https://github.com/fluxcd/flux2/releases/download/v0.10.0/flux_0.10.0_linux_amd64.tar.gz";
    flake = false;
  };
}
```
3. Lock that input by running `nix flake update`.
4. See how the locked url is actually a signed expiring url

**Expected behavior**

The old behavior of NOT following the redirect.

**`nix-env --version` nix-env (Nix) 2.4pre20210317_8a5203d**

dramforever:
This reverts #4595.

Locking on the redirect destination proved to be problematic for some
URLs. For example, GitHub releases downloads redirect to URLs that
expire after some time, which would be broken by this behavior.

Fixes #4672.

We might be able to make a new scheme e.g. `channel` just for the use case mentioned in #4595, but I feel like given the relative minor use case, it would not be worth it.

### DIFF
--- a/src/libfetchers/fetchers.hh
+++ b/src/libfetchers/fetchers.hh
@@ -145,13 +145,7 @@ DownloadFileResult downloadFile(
     bool immutable,
     const Headers & headers = {});
 
-struct DownloadTarballMeta
-{
-    time_t lastModified;
-    std::string effectiveUrl;
-};
-
-std::pair<Tree, DownloadTarballMeta> downloadTarball(
+std::pair<Tree, time_t> downloadTarball(
     ref<Store> store,
     const std::string & url,
     const std::string & name,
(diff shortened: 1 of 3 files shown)


dramforever:
no longer relevant

see 42f0246698cff491f084547f7aaa9e778f629065

Issue #4609, opened 2021-03-06
"Lighter" and "cleaner" shell for `nix develop` and `nix shell`


garbas:
## Proposal

Final goal is to provide **better integration with other shell environments** - eg. zsh, fish, ...

Depending how this will be implemented a very nice to have would be to not spawn a new process when entering shell environment. This would mean the current user shell environment would be updated *in-place* (like direnv does it).

## Reasoning

For those who don't use bash using `nix-shell`, `nix develop` or `nix shell` usually means you need to run `nix-shell --run $SHELL` or creating an alias.

I understand there are good reasons for using bash. Using bash ensures that development environments are very predictably and running commands from `nix-shell` would act as close as it can within a sandbox.

But on the other side we are forcing developers to use bash or use workarounds. As all debates around editors also debates around shell environments are a *hot* topic among developers (which is our audience in this case). I wouldn't expect developers switching to bash just so they can use Nix. I would actually expect the opposite, potential users being turned away from Nix due to the lack of good integration with their shell environment.

As with most of the UX decisions also this comes down to compromising one thing over the other.

## Bonus

Another bonus to consider is also which environment variables we should expose inside the shell environments.

From the perspective of a developer who just wants to enter a development environments to be able to develop their project, all the extra variables that get created confuse. This "type" of developers are not interested in building a project with Nix, but rather only interested in using Nix to provide needed development environment.

From the perspective of a developer who is working on a tool that will be built later with `nix build` they want to see - especially when packaging - 

We already have options to ignore environment variables (via `--ignore-environment` flag). We can make this a default and maybe even improve further and provide even "cleaner" development environments.

Ericson2314:
`nix shell` and `nix develop` are geared for debugging Nix builds. We should have a different command for just using things that doesn't even bother with `stdenv` / `genericBuild` (which use/do bashism and are slow).

abathur:
I was initially going to post in agreement with @Ericson2314, but after reviewing the docs I'm unsure. Clearing up messaging and managing expectations seem like preconditions for thinking this through. I have tried to keep my head in the sand on both flakes and new-style commands, so maybe I'm a good rubber duck here...

`nix --help` says (with irrelevant items removed):
```console
Main commands:
  build    build a derivation or fetch a store path
  develop  run a bash shell that provides the build environment of a derivation
  repl     start an interactive environment for evaluating Nix expressions
  run      run a Nix application
  shell    run a shell in which the specified packages are available
```

I read this as:
[...]

PR #4702, opened 2021-04-09
nix shell: set MANPATH for installables that have a man dir

sternenseemann:
The major change introduced in this commit is that nix shell will extend
`MANPATH` with the `/share/man` directory of all installables present in the
shell. This means that the support for man pages in nix shell becomes
more portable. Currently man pages are discovered by an undocumented
(at least not in its man page) feature of man-db's `man(1)` which also
checks `PATH` to heuristically find prefixes to search for man pages.
This is not supported by mandoc's `man(1)`, but `MANPATH` is supported by
both.

This allows us to partially revert the changes made in
c87f4b9324b87a059cf760a477177f322bb8dc26 as well: Since we don't need to
rely on man-db's `PATH` behavior for discovering man pages, we can only
add a `bin` directory to `PATH` if it exists.

Tested with both mandoc and man-db:

* `nix shell -f '<nixpkgs>' mblaze -c man mshow`
* `nix shell -f '<nixpkgs>' libunwind.devman -c man libunwind`

### DIFF
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -104,16 +104,22 @@ struct CmdShell : InstallablesCommand, RunCommon, MixEnvironment
         setEnviron();
 
         auto unixPath = tokenizeString<Strings>(getEnv("PATH").value_or(""), ":");
+        auto manPath = tokenizeString<Strings>(getEnv("MANPATH").value_or(""), ":");
 
         while (!todo.empty()) {
             auto path = todo.front();
             todo.pop();
             if (!done.insert(path).second) continue;
 
-            if (true)
-                unixPath.push_front(store->printStorePath(path) + "/bin");
+            auto pathString = store->printStorePath(path);
 
-            auto propPath = store->printStorePath(path) + "/nix-support/propagated-user-env-packages";
+            if (accessor->stat(pathString + "/bin").type != FSAccessor::tMissing)
+                unixPath.push_front(pathString + "/bin");
+
+            if (accessor->stat(pathString + "/share/man").type != FSAccessor::tMissing)
+                manPath.push_front(pathString + "/share/man");
+
+            auto propPath = pathString + "/nix-support/propagated-user-env-packages";
             if (accessor->stat(propPath).type == FSAccessor::tRegular) {
                 for (auto & p : tokenizeString<Paths>(readFile(propPath)))
                     todo.push(store->parseStorePath(p));
@@ -121,6 +127,7 @@ struct CmdShell : InstallablesCommand, RunCommon, MixEnvironment
         }
 
         setenv("PATH", concatStringsSep(":", unixPath).c_str(), 1);
+        setenv("MANPATH", concatStringsSep(":", manPath).c_str(), 1);
 
         Strings args;
         for (auto & arg : command) args.push_back(arg);


edolstra:
The problem with this is that there are numerous such environment variables (e.g. `XDG_DATA_DIRS`, `PYTHONPATH`, `PKG_CONFIG_PATH`) and it wouldn't be very elegant to add ad hoc support for each of them to `nix shell`.

grahamc:
Maybe this is something mkShell or stdenv.mkDerivation should do. 

On April 15, 2021, GitHub ***@***.***> wrote:
> The problem with this is that there are numerous such environment
> variables (e.g. XDG_DATA_DIRS, PYTHONPATH, PKG_CONFIG_PATH) and it
> wouldn't be very elegant to add ad hoc support for each of them to nix
> shell.
>

—
<https://github.com/NixOS/nix/pull/4702#issuecomment-820371270>, or
unsubscribe <https://github.com/notifications/unsubscribe-
auth/AAASXLGCRY7KY5CQBTOIJITTI3JEDANCNFSM42VHFLVA>.

sternenseemann:
> The problem with this is that there are numerous such environment variables (e.g. `XDG_DATA_DIRS`, `PYTHONPATH`, `PKG_CONFIG_PATH`) and it wouldn't be very elegant to add ad hoc support for each of them to `nix shell`.

From what I understand `nix shell` is intended for temporary shells where a certain program is available, not for development environment. This is why we don't reuse the derivation environment anymore, right? We need `PATH` for execution and to me `MANPATH` is in scope since it is required for user documentation. Also we already try to support this by adding directories to `PATH` which don't exist, but this relies on `man-db` specific behavior.

I think setting up `MANPATH` and possibly `XDG_DATA_DIRS` or whatever GNU info uses to discover pages is different from setting up things for development environments like `PYTHONPATH`.

> Maybe this is something mkShell or stdenv.mkDerivation should do. 

That would work for the old `nix-shell`, yes, but all `nix shell` has is ad hoc support for certain environment variables, with variables really meaning just `PATH`.

edolstra:
The problem is that we haven't really decided what the semantics or use case of `nix shell` is supposed to be. Some more discussion here: https://github.com/NixOS/nix/issues/4609

> I think setting up MANPATH and possibly XDG_DATA_DIRS or whatever GNU info uses to discover pages is different from setting up things for development environments like PYTHONPATH.

This is already a pretty open-ended set of environment variables, especially if you consider packages that provide plugins. E.g. should `nix shell` update `QT_PLUGIN_PATH`?

SuperSandro2000:
bump

SuperSandro2000:
@thufschmitt can you help to bring this forward?

Ericson2314:
I enqueued it on project board.

fricklerhandwerk:

- There are a lot of variables we would have to bless independently, this doesn't scale.
- @thufschmitt: if someone wants to write a consistent proposal, we'll gladly review it
- @ericson2314: this opens up many architectural questions, especially how we can delegate authority away from Nix to provide solutions for particular use cases where Nix is too generic
    - declarative single-origin profiles are an entirely different problem from ad hoc combining packages, they are different use cases
        - none of them should be a concern for core Nix
    - propose closing this
    - doesn't make sense to keep the problem within our domain of responsibility and at the same time delegating it away
        - @fricklerhandwerk: agree, we've been dodging `nix shell` subjects lately but still changes would be bottlenecked on maintainers
- have to discuss this further

sternenseemann:
If you are taking that stance, I think `nix shell` is already to far—how is `MANPATH` different from `PATH`?

I am actually very open to explore this hardline view, as I believe it would hold the key to solve one problem we have, namely the dubious status of (development) environments. We currently inherit development environments from derivations' build environments which is arguably wrong. The symptoms of this are the variety of `withPackages` facilities that are *only* intended for interactive use and not (usually) utilised by derivations (to be built) and the fact that `mkShell` yields a derivation that can never be built.

`nix run` and `nix shell` are steps away from this that are to be applauded, but ultimately they are not flexible enough (or rather not flexible at all) which is why it is very difficult to  use them as basis for a development-environment-first alternative to `nix develop`. `nix develop` is a misnomer, as it is a derivation debugging facility, strictly speaking.

I feel like the solution may be to have a second type of special Nix value in addition to derivation that does not declare a build environment and instructions, but an environment modification (how to encode that is another question, you could probably look to `direnv`, [shadowenv](https://shopify.github.io/shadowenv/) or invent something). That would mean that user (nixpkgs) code can prescribe what variables are set and Nix could ignore this topic completely. (An alternative to a new value type would be to support special `nix-support` files, e.g. a shadowlisp script that is called to do the env setup.)

In any case, if we are serious about “delegating authority away from Nix”, a lot of things need to be re-evaluated. By the same logic, it feels questionable that the following things are Nix commands:

* `nix develop`—is okay of course in the sense of `debug-derivation`, but is development environments *really* Nix's authority?
* `nix profile`
* `nix shell` (`nix run` seems okay in comparison)
* `nix bundle`
* `nix fmt`

sternenseemann:
To be even more provocative: Shouldn't shells, any kind of environments just be `nix run .#shell` or `$(nix-build -A shell --no-out-link)/bin/shell` where it is totally opaque to Nix how it works?

thufschmitt:
Discussed during the Nix team meeting:

@sternenseemann since you're already climbing down that rabbit hole, would you be willing to write down in more detail / discuss the idea of splitting away `nix develop` from derivations you expose in https://github.com/NixOS/nix/pull/4702#issuecomment-1614861395 ? That would be a much better long-term solution than adding environment variables in an ad-hoc fashion

<details><summary>Discussion log</summary>


- Suggestion: Since Sterni already started sketching out a more principled proposal, we could try to sneak him into expanding it into something more concrete
    - @tomberek: That’s expanding the scope of the discussion quite a lot, shouldn’t we first focus on the small-scope problem at hand?
     - @regnat: I’d rather know where we’re going before expanding it in any way. Possibly we can allow MANPATH once we know we have an escape route

</details>

SuperSandro2000:
I tried rebasing this PR on nix 2.19 and tested it successful with the example commands from above.

```diff
diff --git a/src/nix/run.cc b/src/nix/run.cc
index ea0a17897..5491f4a7b 100644
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -111,16 +111,22 @@ struct CmdShell : InstallablesCommand, MixEnvironment
         setEnviron();

         auto unixPath = tokenizeString<Strings>(getEnv("PATH").value_or(""), ":");
+        auto manPath = tokenizeString<Strings>(getEnv("MANPATH").value_or(""), ":");

         while (!todo.empty()) {
             auto path = todo.front();
             todo.pop();
             if (!done.insert(path).second) continue;

-            if (true)
-                unixPath.push_front(store->printStorePath(path) + "/bin");
+            auto pathString = store->printStorePath(path);

-            auto propPath = CanonPath(store->printStorePath(path)) + "nix-support" + "propagated-user-env-packages";
+            if (accessor->maybeLstat(CanonPath(pathString) + "bin"))
+                unixPath.push_front(pathString + "/bin");
+
+            if (accessor->maybeLstat(CanonPath(pathString) + "share" + "man"))
+                manPath.push_front(pathString + "/share/man");
+
+            auto propPath = CanonPath(pathString) + "nix-support" + "propagated-user-env-packages";
             if (auto st = accessor->maybeLstat(propPath); st && st->type == SourceAccessor::tRegular) {
                 for (auto & p : tokenizeString<Paths>(accessor->readFile(propPath)))
                     todo.push(store->parseStorePath(p));
@@ -128,6 +134,7 @@ struct CmdShell : InstallablesCommand, MixEnvironment
         }

         setenv("PATH", concatStringsSep(":", unixPath).c_str(), 1);
+        setenv("MANPATH", concatStringsSep(":", manPath).c_str(), 1);

         Strings args;
         for (auto & arg : command) args.push_back(arg);
```

fricklerhandwerk:
> a second type of special Nix value in addition to derivation that does not declare a build environment and instructions, but an environment modification

A few spontaneous ideas:

- Encode it as an attribute set and add functionality to the command line that translates it to something the current shell can eat
  - Advantages:
    - Declaration can be agnostic to the shell
    - Will support additional shells once Nix supports it
  - Disadvantages:
    - Requires Nix to know a lot of things about this
    - Imposes policy from below in the stack (I think this is a disadvantage, because it binds Nix to every policy decision forever)
- Let package authors produce environments for every shell they care about and run like any other executable through `nix run`
  - In the simplest form, everything is manual: `nix run foo#shells.fish` which can (but doesn't have to) be aliased to `nix shell foo#fish`
  - Optionally add convenience to detect the current shell
  - Advantages:
    - Very simple and transparent on the Nix side
    - Moves domain knowledge up the stack where it belongs
      - Up the stack you can provide any formalism or abstraction you need to make it shine
    - Allows for really cool things out of the box, such as non-trivial default configurations
      - Those could even be overridden on the fly if we keep the `nix-shell -p` semantics where the argument is a literal Nix expression.
       - Then you could do **crazy** stuff like `nix run nginx#'shells.fish.override { rootDir = "/home/user/test"; }'
  - Disadvantages:
   - No obvious fallback if a shell isn't specified
     - Except the regular build output put into `$PATH`, which is what's being discussed here 
    - Requires package authors to think about yet another thing in their package
      - Can be alleviated with support tooling in Nixpkgs

thufschmitt:
@fricklerhandwerk your second proposal seems to be more or less https://github.com/NixOS/nix/issues/7501, which everyone seems to broadly agree with. So :100: to it.

fricklerhandwerk:
My ramblings https://github.com/NixOS/nix/pull/4702#issuecomment-1824737185 were missing the fact that only `nix run` and `nix develop` take exactly one thing to run an executable. We recently discussed [how to refer to attributes from different expression sources](https://discourse.nixos.org/t/installables-cli-design/38601), and `nix shell` [or whatever you want to call it](https://github.com/NixOS/nix/issues/4715) naturally needs to support multiple such references.   

It's not clear how one would compose an environment from entire shell runs in a simple and obvious way. In any case it would need some sort of contract to implement against, and I'd prefer it to be very small, both in terms of API surface as well as amount of custom code needed to implement it.
- One weird and probably stupid idea would be always adding Nix to the shell's environment, and expecting any given shell environment to take parameters in order to run `nix run <tail of environments>.
- Another, possibly more sensible option, would be sticking to POSIX shell (i.e. requiring environments to be POSIX shell scripts), and leave room for wrappers to translate that to other shells. Then environment composition would mean running their scripts in sequence. The problem arises what will run them (does Nix ship with its own `sh`?), and who's responsible for making the result available to arbitrary shells.
- The sanest I currently can come up with is, for `nix shell` to indeed just expect an attrset of environment variables in the [package attrset](https://github.com/NixOS/nix/pull/9762/files#diff-8327b62197c4dd0503828bcdbbea4eca2bd37528c89e52378e77536ed8140b89R284) and concatenate those literally. But then again, why does Nix have to specify what that should look like? It's exactly such an arbitrary, inflexible forever-policy that IMO we should avoid in Nix, and instead maintain such encodings in Nixpkgs. A practical issue with having this in Nix is that it would ship with a convenient but strictly less powerful means of composition (which by itself is not wrong, but) that works **very differently** from what you'd do in Nix expressions, *the primay way of doing real things*.

In short, I suggest to leave [`buildEnv`](https://github.com/NixOS/nixpkgs/blob/43e4d9e2eb5124b9b4920e0ed6372d5c66c66670/pkgs/build-support/buildenv/default.nix) to Nixpkgs, and reduce the Nix command line interface to essentially `nix run`, `nix build`, `nix eval`, and `nix store` (roughly in descending order of abstraction), explicitly encouraging the community to build and organise useful tooling around that.

Aleksanaa:
@fricklerhandwerk Should we draft an RFC to see what can be developed from this idea and what's the attitude of the community?

iFreilicht:
> In short, I suggest to leave [`buildEnv`](https://github.com/NixOS/nixpkgs/blob/43e4d9e2eb5124b9b4920e0ed6372d5c66c66670/pkgs/build-support/buildenv/default.nix) to Nixpkgs, and reduce the Nix command line interface to essentially `nix run`, `nix build`, `nix eval`, and `nix store` (roughly in descending order of abstraction), explicitly encouraging the community to build and organise useful tooling around that.

I agree with the general architectural idea, but I would hate to see some of these very useful commands go. I use `nix shell` and direnv's integration with nix daily, they're the main reasons why I use Nix in the first place. I would be fine with these being maintained "elsewhere", but I'm wondering how to bring them to users conveniently.

The main challenge I see is to provide an easy way to add functionality to one's installation. There must be some sort of **canonical default entrypoint** to start from. I concur that Nix shouldn't

> [...] ship with a convenient but strictly less powerful means of composition [...] that works very differently from what you'd do in Nix expressions, the primay way of doing real things.

so this default should be declarative and work exactly like composition works everywhere else in Nix, so no `nix profile`, no additional scripts or other shenanigans.

Maybe the solution can simply be to source Nix differently. In addition to adding 

```
. /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh
```

to `/etc/bashrc` or `/etc/zshenv`, we could consider automatically creating a default `user` flake (or regular nix file) in `.config/nix/`, (similar to how `configuration.nix` is created by default in NixOS), and sourcing the resulting environment of that something like 

```bash
source "$(nix build --no-link --print-out-paths path:$HOME/.config/nix/user#userEnv)"
# OR
source "$(nix build --no-link --print-out-paths --file $HOME/.config/nix/userEnv.nix)"
```

after or during the shell startup by default.

This would replace `nix profile` for user-wide installations of programs in a 100% declarative way, it would be completely replaceable (or extensible) by other tools like home-manager, and how the environment is actually built would be the responsibility of nixpkgs' `buildEnv`.

If we find a good mechanism for this entrypoint, I believe this architectural change would solve all the issues that motivate it without negatively impacting the user experience at all.

(We still might need `nix profile`, though just as a barebones tool for managing generations and rollbacks. Installing or upgrading anything would be handled by additional tools that could be part of `userEnv` by default)

fricklerhandwerk:
> (@iFreilicht) I'm wondering how to bring [some of these very useful commands] to users conveniently.

My general idea on this is to essentially port all the porcelain into Nixpkgs and, from there, provide installers for various targets that embed Nix. Then you'd install a Nixpkgs release rather than a Nix release which implicitly sets you up with some Nixpkgs. That could also solve the compatibility struggles between Nix and Nixpkgs to some degree. But I haven't put much thought into how exactly all of that would work. (Another potentially beneficial side effect of doing it that way would be that we could then formally eject the installation business from the Nix code base and let distributors take care of that, according to their needs. Then all we'd have to do is provide documentation on the moving parts that need to be considered. My impression is that the Nixpkgs community would be much more capable of carrying that load than Nix maintainers are at the moment.)

fricklerhandwerk:
I gave the whole thing another series of thoughts, and studied the [`direnv` source code](https://github.com/direnv/direnv) to see what we can learn from it. Here's another proposal, narrowing down https://github.com/NixOS/nix/pull/4702#issuecomment-1824737185 and addressing considerations from https://github.com/NixOS/nix/pull/4702#issuecomment-1902631369:

- Add rules in Nixpkgs:
  - derivations can expose a flat attribute set `env` with environment variables and their values, following some restrictions:
    - no function values
    - only flat lists
  - derivations can expose `shellHook.<shell>` scripts in the respective language
- Add Nixpkgs library functions to
  - compose lists of `env` attribute sets
    - plain values are overwritten in order
    - lists are concatenated as configured per attribute name
      - e.g. `:` by default, and `null` could mean "overwrite in order"
  - convert attribute sets to shell-specific instructions to set up that environment
  - create shell-specific scripts from a list of derivations based on `env` and `shellHook.<shell>` that will
    - stack the environments
    - convert the result to instructions for setting the new values
          - there probably needs to be a flag specifying whether to leave the original value for list attributes, as in `PATH=value:$PATH`
    - make instructions to preserve values of changed environment variables
    - make instructions to revert the environment to the original state 
    - concatenate the hooks
  - wrap the shell-specific scripts in convenience executables
    - observing a `--pure` flag that will discard surrounding the environment
  
That way we could produce arbitrary shells from Nixpkgs packages. One could also reproduce build environments based on https://github.com/NixOS/nixpkgs/pull/324789. Maybe with some trickery we could build a wrapper that takes some information from a failed build and starts the shell in the temporary build directory.

Most of the complexity of `direnv` could be skipped entirely. A Nixpkgs-specific variant would merely have to manage allowed directories and handling asynchronous rebuilds. A project-specific configuration file such as `.nixenv` would only have to specify which environment source (Nix file and attribute) to use for the given directory.

What do you think?

PR #4720, opened 2021-04-19
Include sys/wait.h everywhere WIFEXITED etc is used

alyssais:
This is required on NetBSD, and I think FreeBSD too.


Ericson2314:
CC @p01arst0rm.

PR #4724, opened 2021-04-21
Make `nix shell` fallback to static outputs when needed

### LINKED ISSUE #4721: `nix shell` fails with the `ca-derivations` experimental feature
**Describe the bug**

When the `ca-derivations` feature is enabled, running `nix shell <anything>` results in an error of the form:

```
error: cannot operate on an output of unbuilt content-addresed derivation 'sha256:2e735b3cdc44d7317001c9e6ca8b4a1f3bbed7b405d33994b2a90e3b5e1b619c!out'
```

**Steps To Reproduce**

```console
$ nix shell nixpkgs#hello --experimental-features 'nix-command flakes ca-derivations'
```

(*maybe* the experimental feature also needs to be enabled on the daemon. I didn't check and I don't think it's the case, but it's possible)

regnat:
In case of pure input-addressed derivations, the build loop doesn't guaranty that the realisations are stored in the db (if the output paths are already there or can be substituted, the realisations won't be registered). This caused `nix shell` to fail in some cases because it was assuming that the realisations were always existing.

A better (but more involved) fix would probably to ensure that we always register the realisations, but in the meantime this patches the surface issue.

Fix #4721

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -736,9 +736,18 @@ std::set<RealisedPath> toRealisedPaths(
                                     output.first);
                             auto outputId = DrvOutput{outputHashes.at(output.first), output.first};
                             auto realisation = store->queryRealisation(outputId);
-                            if (!realisation)
-                                throw Error("cannot operate on an output of unbuilt content-addresed derivation '%s'", outputId.to_string());
-                            res.insert(RealisedPath{*realisation});
+                            if (!realisation) {
+                                // TODO: remove this check once #4725 is fixed
+                                // as we’ll have the guaranty that if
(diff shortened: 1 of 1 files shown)


Ericson2314:
Can we have an issue for the underlying issue, and a TODO in the code to change back once it is fixed?

regnat:
> Can we have an issue for the underlying issue, and a TODO in the code to change back once it is fixed?

I've just created the issue: https://github.com/NixOS/nix/issues/4725

The TODO is a good idea too, indeed. I'll add it

regnat:
Just added to TODO

Ericson2314 (approved):

Ericson2314:
Thanks!

PR #4730, opened 2021-04-22
libutil: allow decompression with none/empty method

lukegb:
The S3 store relies on the ability to be able to decompress things with
an empty method, because it just passes the value of the Content-Encoding
directly to decompress.

If the file is not compressed, then this will cause the compression
routine to get confused.

This caused NixOS/nixpkgs#120120.

### DIFF
--- a/src/libutil/compression.cc
+++ b/src/libutil/compression.cc
@@ -186,7 +186,9 @@ struct BrotliDecompressionSink : ChunkedCompressionSink
 
 ref<std::string> decompress(const std::string & method, const std::string & in)
 {
-    if (method == "br") {
+    if (method == "none" || method == "")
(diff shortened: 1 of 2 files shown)


regnat (approved):

edolstra:
Not sure I understand. What has changed that's causing https://github.com/NixOS/nixpkgs/issues/120120?

edolstra:
Looks like 8a0c00b85600991cdb9aa05902defec6ac44b777 made a bit of a mess of what compression methods are supported by `decompress`, `makeDecompressionSink` and the new `makeDecompressionSource`. We should make `decompress` a simple wrapper around `makeDecompressionSink` again, and get rid of `makeDecompressionSource`.

PR #4731, opened 2021-04-22
nix-instantiate: parse to json ast

### LINKED ISSUE #4726: feature: parse nix expression to json ast

i need to parse `nixpkgs/pkgs/top-level/all-packages.nix`
to map package names to file paths

for example, parse this nix
```nix
{ colorz = callPackage ../tools/misc/colorz { }; }
```
to json
```json
{ "colorz": "../tools/misc/colorz" }
```

my program is in javascript, so json would be the best format (easy + fast)
currently `nix-instantiate --parse` can only produce "ATerm format"

ideally i could call `nix-instantiate --parse --json /path/to/file.nix`
the get the parsed ast in json format

[nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt) is terribly slow ... 40 to 120 times slower, probably a bug in the parser library [rnix-parser](https://github.com/nix-community/rnix-parser)

```
f=/nix/store/v0xwj556c69yppjzylz2diqk66vliswb-nixos-20.09.3857.b2a189a8618/nixos/pkgs/top-level/all-packages.nix

time /tmp/nixpkgs-fmt/target/debug/nixpkgs-fmt --parse $f >/dev/null 
real    0m7.111s

time /tmp/nixpkgs-fmt/target/debug/nixpkgs-fmt --parse --output-format json $f >/dev/null 
real    0m20.632s

time nix-instantiate --parse $f >/dev/null 
real    0m0.176s

time node simple-regex-parser.js >/dev/null 
real    0m0.133s
```

a simple regex parser is ... too simple
```js
const path = "/nix/store/v0xwj556c69yppjzylz2diqk66vliswb-nixos-20.09.3857.b2a189a8618/nixos/pkgs/top-level/all-packages.nix";
const fs = require('fs');
const src = fs.readFileSync(path, 'utf8');
console.log(src.match(/[^ \t]*? = callPackage .*?;/g).join('\n'));
```

writing an ATerm parser for javascript also sounds like a bad idea (slow)

**Additional context**

https://github.com/NixOS/nix/pull/291 (add `--json` option for `--eval`)
https://github.com/NixOS/nix/issues/1102#issuecomment-259659097 "According to c5baaaf, there was a 80x speedup in switching to Bison/Yacc."

milahu:
resolve #4726 

work in progress ... edit: almost done

basically there are two possible solutions:

1\. recursive: on `struct Expr` and its derived structs, add a method `showAsJson`.
the root expression is printed with `expr->showAsJson(std::cout);`
and `showAsJson` prints child expressions with `childExpr->showAsJson(std::cout);`.
a similar solution is used in [nlohmann/json](https://github.com/nlohmann/json/blob/6f551930e5c7ef397056de121c0da82f77573cca/include/nlohmann/detail/output/serializer.hpp#L96)

2\. use an iterative visitor function to print the AST nodes
to make this work, we would have to expose the node type in `struct Expr`
-> see `internalType` in `struct Value`
this could be faster than solution 1, but its harder to implement (manage the stack manually)

1a\. Alipha from `##c++` on freenode suggested a more complex solution
https://wandbox.org/permlink/luDOhrLzloiSME6h
but imho we dont need that complexity (keep it simple)


src/libexpr/nixexpr.hh
```diff
@@ -79,16 +79,28 @@ struct Expr
...
+struct ExprAsJson : Expr
+{
+    void show(std::ostream & str) const;
+};
+
 std::ostream & operator << (std::ostream & str, const Expr & e);
 
 #define COMMON_METHODS \
-    void show(std::ostream & str) const; \
+    void showAsAterm(std::ostream & str) const; \
```
edolstra, line 103: Let's not rename this to minimize code churn.

src/libexpr/nixexpr.hh
```diff
@@ -79,16 +79,28 @@ struct Expr
...
     virtual Value * maybeThunk(EvalState & state, Env & env);
     virtual void setName(Symbol & name);
 };
 
+struct ExprAsAterm : Expr
+{
+    void show(std::ostream & str) const;
+};
+
+struct ExprAsJson : Expr
```
edolstra, line 95: What's the purpose of these classes?

src/libexpr/nixexpr.hh
```diff
@@ -79,16 +79,28 @@ struct Expr
 {
     virtual ~Expr() { };
     virtual void show(std::ostream & str) const;
+    virtual void showAsAterm(std::ostream & str) const;
+    virtual void showAsJson(std::ostream & str) const;
```
edolstra, line 83: This should be
```
virtual nlohmann::json to_json() const;
```
or something like that.

edolstra (changes requested):

src/nix-instantiate/nix-instantiate.cc
```diff
@@ -31,7 +31,12 @@ void processExpr(EvalState & state, const Strings & attrPaths,
     bool evalOnly, OutputKind output, bool location, Expr * e)
 {
     if (parseOnly) {
-        std::cout << format("%1%\n") % *e;
+        if (output == okJSON) {
+            std::cout << format("%1%\n") % (ExprAsJson *) *e;
```
edolstra, line 35: We shouldn't add this functionality to `nix-instantiate` since it's legacy. Better to add a new command like `nix parse`.

src/libexpr/nixexpr.hh
milahu: `nlohmann::json` sounds like overhead .. intermediary data structure + unnecessary pretty printing

src/libexpr/nixexpr.hh
milahu: my attempt to solve this in the most lazy way:
by type-casting `Expr` to `ExprAsJson`, it should use the `showAsJson` methods

src/nix-instantiate/nix-instantiate.cc
milahu: yes, thanks for remind. also the CLI of `nix-instantiate` is not ideal (`--json`, `--xml`)

src/libexpr/nixexpr.hh
milahu: done

milahu:
status: working prototype

`echo '{ foo = "bar"; }' >test.nix; ./nix-instantiate --parse --json test.nix | jq`

<details><summary>json ast result</summary>

```json
{
  "type": 6,
  "recursive": false,
  "attrs": [
    {
      "type": "attr",
      "inherited": false,
      "name": "foo",
      "value": {
        "type": 10,
        "value": "bar"
      }
    }
  ],
  "dynamicAttrs": []
}
```
type 6 = ExprAttrs, type 10 = ExprString, ... see `enum class NodeTypeId` in `nixexpr.hh`
`"type": "attr",` should be numeric type

</details>

src/nix-instantiate/nix-instantiate.cc
milahu: first draft in 4c35710

src/libexpr/nixexpr.hh
milahu: that [dump](https://github.com/nlohmann/json/blob/6f551930e5c7ef397056de121c0da82f77573cca/include/nlohmann/detail/output/serializer.hpp#L96) function is similar to my implementation of `showAsJson`, also using recursion, not iteration

milahu:
i will remove the other formats like `json-arrays-fmt`, only needed for benchmarks

<details><summary>benchmark results</summary>

my `json-arrays` format needs 50% of the size
and 70% of the parse-time of the `json` format
80 ms vs 120 ms for a 900 KByte nix file (all-packages.nix from nixpkgs)

my `json-numtypes` format (with types encoded as numbers) can be parsed 10% faster
than my `json` format (with types encoded as strings)

generating the json and walking the tree has the same speed
in javascript, consuming array-trees is not faster than consuming object-trees
(but javascript is highly optimized for objects)

<details><summary>benchmark graphs</summary>

three graphs: generate, parse, walk
three formats: json, json-numtypes, json-arrays = purple, green, blue

benchmark-parse-json-vs-json-arrays 1620024276

</details>

</details>

milahu:
the CLI of `nix parse` needs some more work
should we just use the same interface as `nix eval`?

```
$ nix parse --output-format json '(1 + 2)'
{"type":"ExprConcatStrings","line":1,"column":2,"strings":[{"type":"ExprInt","value":1},{"type":"ExprInt","value":2}]}

$ nix parse --output-format json --file some-file.nix
...
```

note: concat strings and arithmetic sum have the same node type

milahu:
push

needed for nix-gui, to avoid stupid workarounds like [this](https://github.com/nix-gui/nix-gui/blob/master/nix_dump_syntax_tree_json/src/main.rs)

milahu:
closing in favor of #5512

PR #4732, opened 2021-04-22
Aways register the realisations

### LINKED ISSUE #4725: Ensure that a built path always has an associated realisation
If a pure input-addressed path is built without the `ca-derivations` experimental feature, then its associated realisations will never be stored (because next time Nix is asked to build it, it will bypass most of the build loop, and in particular skip the output registration step).
The same thing happens if the path is substituted from a remote store.

This is annoying, because it would be much more convenient in higher-level code to just be able to assume that the realisations are here − and more than being inconvenient, this is a call for bugs (like https://github.com/NixOS/nix/issues/4721 which is precisely an instance of this)

regnat:
Fix #4725

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -736,18 +736,9 @@ std::set<RealisedPath> toRealisedPaths(
                                     output.first);
                             auto outputId = DrvOutput{outputHashes.at(output.first), output.first};
                             auto realisation = store->queryRealisation(outputId);
-                            if (!realisation) {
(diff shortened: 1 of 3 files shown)


Ericson2314:
@regnat want to undo the nix-shell workaround then?

regnat:
> @regnat want to undo the nix-shell workaround then?

Oh yes, of course. I've reverted it in https://github.com/NixOS/nix/pull/4732/commits/d3df747cb6813fc5f9d0fcc63d6fd16292c92e04

PR #4733, opened 2021-04-22
Add nix query-substituters command

### LINKED ISSUE #3752: Add "nix weather" command that is equivalent to "guix weather"
Guix has a feature that is missing in Nix:

https://guix.gnu.org/manual/en/html_node/Invoking-guix-weather.html

guix weather is used to determine whether substitutes are available for your build at all. It can give some idea of how much resources you'll have to spend when you build a derivation.

matthewbauer:
nix query-substituters is used to figure out how much of a flake's packages are
in a substituter. This allows us to get an idea for how much we'll
have to build locally. It can be used with Nixpkgs, but also
individual projects to determine objectively how much is missing from
the cache.

Example usage is:

$ nix query-substituters nixpkgs --substituters https://cache.nixos.org

Note that this is using derivations instead of output paths to
determine what depends on each path. This is necessary since we don't
know what dependencies a given store path will have unless it's built.

/cc @edolstra @domenkozar @vcunat @masaeedu

Fixes #3752

### DIFF
--- a/src/nix/weather.cc
+++ b/src/nix/weather.cc
@@ -5,12 +5,13 @@
 #include "eval-cache.hh"
 #include "attr-path.hh"
 #include "nar-info-disk-cache.hh"
+#include "json.hh"
 
 #include <queue>
 
 using namespace nix;
 
-struct CmdWeather : InstallablesCommand
+struct CmdWeather : InstallablesCommand, MixJSON
 {
     bool noClosure = false;
 
@@ -141,6 +142,8 @@ struct CmdWeather : InstallablesCommand
         uint64_t totalNarSize = 0;
         std::optional<uint64_t> totalDownloadSize = 0;
 
+        auto jsonOut = json ? std::make_unique<JSONObject>(std::cout) : nullptr;
+
         for (auto & sub : subs) {
             StorePathSet validPaths = sub->queryValidPaths(outPaths);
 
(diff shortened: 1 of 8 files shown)


grahamc:
I wonder if this should respect `WantMassQuery`?

matthewbauer:
> I wonder if this should respect `WantMassQuery`?

Yeah that sounds good. I've updated with a check for that.

edolstra:
What does "weather" mean in this context? I have trouble connecting it with what the command does.

src/nix/weather.cc
```diff
@@ -0,0 +1,191 @@
...
+        std::optional<uint64_t> totalDownloadSize = 0;
+
+        for (auto & sub : subs) {
+            StorePathSet validPaths = sub->queryValidPaths(outPaths);
+
+            ssize_t pathsFound = 0;
+            uint64_t narSize = 0;
+            std::optional<uint64_t> downloadSize = 0;
+
+            for (auto & path : validPaths) {
```
edolstra, line 151: This looks very sequential so it will be latency-bound. Maybe you can use `queryMissing` instead? It should be a lot faster. Or use the async version of `queryPathInfo`.

src/nix/weather.cc
```diff
@@ -0,0 +1,191 @@
...
+                        totalDownloadSize = std::nullopt;
+                    totalNarSize += info->narSize;
+                }
+            }
+
+            logger->cout("Substituter %s", sub->getUri());
+            logger->cout("  %6.1f%% of paths have substitutes available (%s of %s)", (float) 100 * pathsFound / outPaths.size(), pathsFound, outPaths.size());
+            if (downloadSize)
+                logger->cout("  %s downloaded (compressed)", formatSize(*downloadSize));
+            logger->cout("  %s downloaded (uncompressed)", formatSize(narSize));
```
edolstra, line 177: "downloaded (compressed)" and `"downloaded` (uncompressed)" is confusing to me. Maybe just "compressed size" (or "download `size)` and "uncompressed size"?

domenkozar:
The naming `weather` is clever, but not clear to the user.

How about `nix query-substituters` or something similar?

andir:
Could this also get a machine readable output so we can script around it?

src/nix/weather.cc
matthewbauer: queryValidPaths should populate the cache up front. These queryPathInfo should be just cache lookups. I supposed some caches may not have cached backing them up? In that case, it would be very slow.

PR #4736, opened 2021-04-23
doc: fix section in nix3 man page metadata

alyssais:
These man pages said they were in section 7, even though we were
installing them to section 1 (which is the right place for them).

### DIFF
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -80,7 +80,7 @@ install: $(d)/src/command-ref/new-cli
 	  if [[ $$name = SUMMARY ]]; then continue; fi; \
 	  printf "Title: %s\n\n" "$$name" > $$i.tmp; \
 	  cat $$i >> $$i.tmp; \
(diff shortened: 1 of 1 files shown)


edolstra:
This gives:
```
lowdown: invalid option -- 'M'
```

Is this a new option? When I added manpage generation via lowdown, it didn't have any way to set the section IIRC.

AmandaCameron:
Was this meant to be merged? https://github.com/NixOS/nix/runs/2418966042 seems to be spitting out the same error that @edolstra  mentioned and my hydra instance is.

edolstra:
Yes, because I updated to lowdown 0.8.4 first (788008385ef5bf7edb799977525d6f73f02c76bc).

AmandaCameron:
Ah, I see now, for whatever reason my flake is pulling in lowdown-0.8.0
instead of anything newer, despite it meant to be pulling from
nixos-unstable, which has lowdown at 0.8.4 as well.

PR #4759, opened 2021-05-01
libstore/derivation-goal: don't raise an error if the only remaining output is a wildcard

Ma27:
Since 255d145ba7ac907d1cba8d088da556b591627756, Nix can use wildcards to
query a derivation's output internally, for instance

    /nix/store/zipc1j00q8i30xaylg0fz64lf2ypx752-bash-interactive-4.4-p23.drv!*

This however seems to break the check in `derivation-goal.cc` which
errors if requested outputs aren't present. This can be reproduced on
Nix without this patch by opening a `nix-shell` like this:

    $ nix-shell -A hello
    error: derivation '/nix/store/zipc1j00q8i30xaylg0fz64lf2ypx752-bash-interactive-4.4-p23.drv' does not have wanted outputs '*'

As `*` as a wildcard accepts anything, the error won't be thrown if the
wildcard is the only remainig, "unmatched" output.

cc @Ericson2314 @regnat @edolstra @grahamc

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1292,10 +1292,13 @@ void DerivationGoal::checkPathValidity()
     // If we requested all the outputs via the empty set, we are always fine.
     // If we requested specific elements, the loop above removes all the valid
     // ones, so any that are left must be invalid.
-    if (!wantedOutputsLeft.empty())
-        throw Error("derivation '%s' does not have wanted outputs %s",
-            worker.store.printStorePath(drvPath),
-            concatStringsSep(", ", quoteStrings(wantedOutputsLeft)));
+    if (!wantedOutputsLeft.empty()) {
+        if (!(wantedOutputsLeft.size() == 1 && wantedOutputs.find("*") != wantedOutputs.end())) {
+            throw Error("derivation '%s' does not have wanted outputs %s",
+                worker.store.printStorePath(drvPath),
+                concatStringsSep(", ", quoteStrings(wantedOutputsLeft)));
+        }
+    }
 }
 
 


Ericson2314 (changes requested):
Thanks for catching this, but the problem is actually earlier when a `DerivedPath` is parsed. The `*` should always become and empty outputs set.

Ma27:
@Ericson2314 OK in fact your check was sufficient, the only mistake I made was that I assumed the daemon to be forward-compatible (as it used to be the case). Since latest Nix master for CLI and stable `nix-daemon` and latest Nix master for CLI and daemon both work fine, this patch isn't needed and can be closed. Thanks for your clarification! :+1:

Ericson2314:
@Ma27 that is a bug, because the version negotiation is supposed to make things fall back properly when the client is newer.

Ma27:
@Ericson2314 ok I'm not sure what the policy is for different unstable daemons. But, how would you suggest to fix that? Unfortunately my knowledge about Nix internals is a bit limited :D

Ericson2314:
@Ma27 I think I must have forgotten to bump the version number at some point :/. I think I saw two different branches independently bump it and then no conflict, rather than +2.

> OK I'm not sure what the policy is for different unstable daemons. 

Currently we bump the version number every time. That does quickly go through version numbers between stable releases, but is supposed to avoid the problem you had.

Ma27:
> Currently we bump the version number every time. That does quickly go through version numbers between stable releases, but is supposed to avoid the problem you had.

Right, now that you say it :) 

Would it be possible for you to fix this? I'm not sure if I understand the entire patch well-enough to know where exactly a versioning check is needed here.

Ma27:
@Ericson2314 in fact, I was too curious and decided to take another look. AFAIU `DerivedPath::Built::to_string` should only add the wildcard if the daemon supports that, so this behavior should be behind an `if` checking for the daemon version, correct?

Ma27:
I'll check if that makes sense so far. If that sounds correct to you, I'd reopen this PR wiht a new patch :)

Ma27:
The current diff looks like this:

``` diff
diff --git a/src/libstore/derived-path.cc b/src/libstore/derived-path.cc
index 13833c58e..abcfe68e3 100644
--- a/src/libstore/derived-path.cc
+++ b/src/libstore/derived-path.cc
@@ -1,5 +1,6 @@
 #include "derived-path.hh"
 #include "store-api.hh"
+#include "worker-protocol.hh"
 
 #include <nlohmann/json.hpp>
 
@@ -35,13 +36,19 @@ std::string DerivedPath::Opaque::to_string(const Store & store) const {
     return store.printStorePath(path);
 }
 
-std::string DerivedPath::Built::to_string(const Store & store) const {
-    return store.printStorePath(drvPath)
-        + "!"
-        + (outputs.empty() ? std::string { "*" } : concatStringsSep(",", outputs));
+std::string DerivedPath::Built::to_string(Store & store) const {
+    auto base = store.printStorePath(drvPath);
+    auto daemonProtocolVersion = store.getProtocol();
+    if (!outputs.empty()) {
+        return base + "!" + concatStringsSep(",", outputs);
+    }
+    if (GET_PROTOCOL_MINOR(daemonProtocolVersion) >= 30) {
+        return base + "!*"
+    }
+    return base;
 }
 
-std::string DerivedPath::to_string(const Store & store) const
+std::string DerivedPath::to_string(Store & store) const
 {
     return std::visit(
         [&](const auto & req) { return req.to_string(store); },
diff --git a/src/libstore/derived-path.hh b/src/libstore/derived-path.hh
index 7a2fe59de..60b028bda 100644
--- a/src/libstore/derived-path.hh
+++ b/src/libstore/derived-path.hh
@@ -42,7 +42,7 @@ struct DerivedPathBuilt {
     StorePath drvPath;
     std::set<std::string> outputs;
 
-    std::string to_string(const Store & store) const;
+    std::string to_string(Store & store) const;
     static DerivedPathBuilt parse(const Store & store, std::string_view);
 };
 
@@ -72,7 +72,7 @@ struct DerivedPath : _DerivedPathRaw {
         return static_cast<const Raw &>(*this);
     }
 
-    std::string to_string(const Store & store) const;
+    std::string to_string(Store & store) const;
     static DerivedPath parse(const Store & store, std::string_view);
 };
 
diff --git a/src/libstore/remote-store.cc b/src/libstore/remote-store.cc
index 761b4a087..553587f55 100644
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -35,7 +35,7 @@ StorePath read(const Store & store, Source & from, Phantom<StorePath> _)
     return store.parseStorePath(readString(from));
 }
 
-void write(const Store & store, Sink & out, const StorePath & storePath)
+void write(Store & store, Sink & out, const StorePath & storePath)
 {
     out << store.printStorePath(storePath);
 }
diff --git a/src/libstore/worker-protocol.hh b/src/libstore/worker-protocol.hh
index 001ed25e3..fdd692cf0 100644
--- a/src/libstore/worker-protocol.hh
+++ b/src/libstore/worker-protocol.hh
@@ -9,7 +9,7 @@ namespace nix {
 #define WORKER_MAGIC_1 0x6e697863
 #define WORKER_MAGIC_2 0x6478696f
 
-#define PROTOCOL_VERSION (1 << 8 | 29)
+#define PROTOCOL_VERSION (1 << 8 | 30)
 #define GET_PROTOCOL_MAJOR(x) ((x) & 0xff00)
 #define GET_PROTOCOL_MINOR(x) ((x) & 0x00ff)

```

AFAIU if the remote side is old enough it doesn't have a problem with building opaque derivations (i.e. derivations with no outputs specified) as it was the case with `StorePathWithOutputs` without outputs defined. Please note that this doesn't compile yet, I gotta figure out a better way to test for the daemon protocol version (as `getProtocol()` is non-const, but the `store` parameter is).

Ma27:
Continued in #4763. In fact I misunderstood your explanation about what exactly went wrong. The new PR does a bump of the daemon protocol version and fixes the checks where necessary :)

PR #4761, opened 2021-05-02
Mark `__impureHostDeps` paths as optional

### LINKED ISSUE #4658: Build sandbox seems broken on macOS 11 Big Sur
**Describe the bug**

I get the error `error: getting attributes of path '/usr/lib/libSystem.B.dylib: No such file or directory` when attempting to run builds with the sandbox enabled on Big Sur.

**Steps to reproduce**
Set `sandbox` to `true` or `relaxed` and try to build anything.

emilazy:
Starting in macOS 11, the on-disk dylib bundles are no longer available, but nixpkgs needs to be able to keep compatibility with older versions that require `/usr/lib/libSystem.B.dylib` in `__impureHostDeps`. Allow it to keep backwards compatibility with these versions by marking these dependencies as optional.

Fixes #4658.

If you’d like to test out this fix, the following settings should with with nix-darwin:

```nix
{ pkgs, ... }:

{
  nix.package = pkgs.nixFlakes.override (prev: {
    patches = prev.patches or [ ] ++ [
      (pkgs.fetchpatch {
        url = "https://github.com/NixOS/nix/pull/4761.patch";
        sha256 = "1z7s2dkbsd3fk5688s0870k942r8nsnmr140sn5vgjafywihbihb";
      })
    ];
  });

  # Then enable:
  #nix.useSandbox = "relaxed";
}
```

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -581,7 +581,9 @@ void LocalDerivationGoal::startBuilder()
                 throw Error("derivation '%s' requested impure path '%s', but it was not in allowed-impure-host-deps",
                     worker.store.printStorePath(drvPath), i);
 
-            dirsInChroot[i] = i;
+            /* Allow files in __impureHostDeps to be missing; e.g.
+               macOS 11+ has no /usr/lib/libSystem*.dylib */
+            dirsInChroot[i] = {i, true};
         }
 
 #if __linux__


emilazy:
Note that I don’t know much about Darwin sandboxing or if `/usr/lib/libSystem.B.dylib` necessarily has to stay in the default nixpkgs sandbox paths; see https://github.com/NixOS/nix/issues/4658#issuecomment-830740007.

emilazy:
I don’t understand the CI failure and I’m also seeing some build failures with the sandbox on (e.g. `pkgs.fish` during the tests); not sure whether these are general macOS infelicities or caused by the patch.

PR #4763, opened 2021-05-02
Bump version number for `DerivedPath` changes

Ma27:
I guess I misunderstood John's initial explanation about why wildcards
for outputs are sent to older stores[1]. My `nix-daemon` from 2021-03-26
also has version 1.29, but misses the wildcard[2]. So bumping seems to
be the right call.

However this now properly (I guess?) fixes errors like this:

```
$ nix-shell -A hello
error: derivation '/nix/store/zipc1j00q8i30xaylg0fz64lf2ypx752-bash-interactive-4.4-p23.drv' does not have wanted outputs '*'
```

[1] https://github.com/NixOS/nix/pull/4759#issuecomment-830812464
[2] 255d145ba7ac907d1cba8d088da556b591627756

cc @Ericson2314 @edolstra

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -263,7 +263,7 @@ static void writeValidPathInfo(
 static std::vector<DerivedPath> readDerivedPaths(Store & store, unsigned int clientVersion, Source & from)
 {
     std::vector<DerivedPath> reqs;
-    if (GET_PROTOCOL_MINOR(clientVersion) >= 29) {
+    if (GET_PROTOCOL_MINOR(clientVersion) >= 30) {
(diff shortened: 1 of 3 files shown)


Ericson2314 (approved):
LGTM! Thanks.

Sort of a funny thing to do after the fact, but I suppose it does help. (This is what I recommended, to be clear.)

bjornfor:
Thanks! I was seeing the same issue with `nix-env`, but I haven't found the time to make an issue about it.

regnat (approved):
It’s a bit unfortunate that we couldn’t catch this earlier, but nice to have it fixed now :)

PR #4765, opened 2021-05-03
Add ignored_acls setting

MagicRB:
This PR adds a `ignored-acls` setting to `nix.conf`, which makes it possible to ignore certain ACLs, like the ones added by selinux or nfs.

Fixes and closes https://github.com/NixOS/nixpkgs/issues/29778

Signed-off-by: Magic_RB <magic_rb@redalder.org>

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -583,9 +583,7 @@ static void canonicalisePathMetaData_(const Path & path, uid_t fromUid, InodesSe
             throw SysError("querying extended attributes of '%s'", path);
 
         for (auto & eaName: tokenizeString<Strings>(std::string(eaBuf.data(), eaSize), std::string("\000", 1))) {
-            /* Ignore SELinux security labels since these cannot be
-               removed even by root. */
-            if (eaName == "security.selinux") continue;
+            if (settings.ignoredAcls.get().count(eaName)) continue;
             if (lremovexattr(path.c_str(), eaName.c_str()) == -1)
                 throw SysError("removing extended attribute '%s' from '%s'", eaName, path);
         }
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -805,6 +805,15 @@ public:
           may be useful in certain scenarios (e.g. to spin up containers or
           set up userspace network interfaces in tests).
         )"};
+
+    Setting<StringSet> ignoredAcls{
+        this, {"security.selinux"}, "ignored-acls",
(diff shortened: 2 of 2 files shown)


bbqsrc:
Just hit this issue when trying to install NixOS to an NFSv4 drive. This approach seemed entirely noninvasive and user opt-in, so hopefully can be merged. 😄

ethancedwards8:
This pr actually solves an issue for me too! If this could get merged it would be great. I'm confused as to why the pr was been sitting here for a month with no reviews or comments except from people who's issues are solved by this.

src/libstore/local-store.cc
```diff
@@ -529,9 +529,16 @@ static void canonicalisePathMetaData_(const Path & path, uid_t fromUid, InodesSe
...
+            bool ignore =  false;
+            std::list<std::string>::iterator it;
+            std::list<std::string> ignoredAcls = settings.ignoredAcls.get();
+            for (it = ignoredAcls.begin(); it != ignoredAcls.end(); ++it) {
+                if (eaName == *it) {
+                    ignore = true;
+                    break;
+                }
+            }
+            if (ignore) continue;
```
edolstra, line 541: If you change the type of `settings.ignoredAcls` to `StringSet`, then it should be possible to write this as
```
if (settings.ignoredAcls.get().count(eaName)) continue;
```

edolstra:
Ideally it shouldn't be necessary for users to configure ignored ACLs. Maybe it's better to just ignore `system.nfs4_acl`?

MagicRB:
> Ideally it shouldn't be necessary for users to configure ignored ACLs. Maybe it's better to just ignore `system.nfs4_acl`?

My thinking was that random ACLs may crop up in the future and this is a future proofed way to handle that case. Though it may be desirable to ignore `system.nfs4_acl` by default

MagicRB:
I pushed your suggestion, I haven't tested it yet in my Hydra, will do so soon.

MagicRB:
I'm not 100% sure, but I think that keeping the NFS ACL around means that FOD will fail, https://hydra.redalder.org/build/102/nixlog/9. I don't suppose the NFS ACL could be excluded from the hash somehow?

junjihashimoto:
@edolstra 
I need this feature.
I look at xattr codes and I don't think it has anything to do with the hash mismatch.
image

It works in my environment.
What do we need to move forward?

junjihashimoto:
Thx!

PR #4769, opened 2021-05-04
install: Fix hash variable for Darwin.arm64

andersk:
When commit 233b61d3d6482544c35b9d340240bf3260acff13 (#4224) renamed `@binaryTarball_${system}@` to `@tarballHash_${system}@`, it updated this reference for every platform except Darwin.arm64.

Cc @zimbatm.


abathur (approved):
Good catch.

PR #4770, opened 2021-05-04
Source complete env in `nix-shell` with `__structuredAttrs = true;`

Ma27:
###### Notes

* This is intended to help getting structured attrs in nixpkgs forward (https://github.com/NixOS/nixpkgs/pull/72074)
* Missing things:
  * [x] Move logic to generate .attrs.json & .attrs.sh into `ParsedDerivation`
  * [x] Don't create `.attrs.json` in PWD
  * [x] Address review comments
  * [x] nix develop

CCing @regnat (because I had to touch input rewrites inside `{Local,}DerivationGoal` a bit), @Ericson2314 (because he touched a lot of code down there), @globin (because of the nixpkgs PR), @edolstra 

###### Commit message

This is needed to push the adoption of structured attrs[1] forward. It's
now checked if a `__json` exists in the environment-map of the derivation
to be openend in a `nix-shell`.

Derivations with structured attributes enabled also make use of a file
named `.attrs.json` containing every environment variable represented as
JSON which is useful for e.g. `exportReferencesGraph`[2]. To
provide an environment similar to the build sandbox, `nix-shell` now
adds a `.attrs.json` to `cwd` (which is mostly equal to the one in the
build sandbox) and removes it using an exit hook when closing the shell.

To avoid leaking internals of the build-process to the `nix-shell`, the
entire logic to generate JSON and shell code for structured attrs was
moved into the `ParsedDerivation` class.

[1] https://nixos.mayflower.consulting/blog/2020/01/20/structured-attrs/
[2] https://nixos.org/manual/nix/unstable/expressions/advanced-attributes.html#advanced-attributes


Ma27:
Also, anybody with sufficient Darwin & C++ knowledge to tell me what to use instead of `std::filesystem::current_path()`? :sweat_smile:

src/nix-build/nix-build.cc
```diff
@@ -422,12 +430,38 @@ static void main_nix_build(int argc, char * * argv)
...
+            StringSet wantedOutputs;
+            for (auto const & output : drvInfo.queryOutputs()) {
+                wantedOutputs.insert(output.first);
+            }
+
+            // FIXME this means that a drv will be re-evaluated just to get all inputs.
+            auto drvGoal = worker.makeDerivationGoal(storePath, wantedOutputs, bmNormal);
+            drvGoal->getDerivation();
+            drvGoal->inputsRealised();
+            if (auto structAttrs = drvGoal->generateStructuredAttrs(std::nullopt)) {
```
regnat, line 448: Can we get rid of that by moving `generateStructuredAttrs` out of `drvGoal`?

regnat (changes requested):
Did a first review.

- A big :+1: on the end-goal
- Implementation-wise, I’m not really fond of inlining the `DerivationGoal` in `nix-build.cc`. That’s quite a big abstraction breach (and time has already brought enough of these…)
- I’m also not thrilled by the `.attrs.json` in PWD. It’d be nice if there were an alternative way of doing that − if only because I’d like to start two `nix-shell`s from the same directory without them meddling with each other.

(this also made me realise that `nix-shell` doesn’t really support CA derivations right now… but that’s another topic)

Ericson2314:
@regnat wrote

> Also, does this really need to be a member of `DerivationGoal` or could it be a standalone function just taking `parsedDrv` as input? (or maybe even a method of `ParsedDerivation`?)

Yes i would strongly prefer that too.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -1083,113 +1083,18 @@ void LocalDerivationGoal::initEnv()
...
-                    storePaths.insert(worker.store.parseStorePath(p.get<std::string>()));
-                worker.store.pathInfoToJSON(jsonRoot,
-                    exportReferences(storePaths), false, true);
-            }
-            json[i.key()] = nlohmann::json::parse(str.str()); // urgh
-        }
+    if (auto structAttrs = generateStructuredAttrs(inputRewrites)) {
+        auto value = structAttrs.value();
+        auto jsonSh = value.first;
+        auto json = value.second;
```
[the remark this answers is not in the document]
Ma27: Didn't know about this, so thanks anyway!

Ma27:
@regnat @Ericson2314  I'm not sure about it and this is in fact my primary pain point as well. The problem is that from my understanding `exportReferences` (from derivation goal) is needed to be able to create the value for `exportReferencesGraph` and IMHO this variable should also appear in a `nix-shell`. Do you have a hint how to do that without using DerivationGoal itself? As I said in the OP, I'm not too experienced with Nix internals :) 

Other than that, it's totally possible to move this to ParsedDerivation (in fact I even have something for it in my git stash lying around, but I didn't use it in the end because of my problem with `exportReferencesGraph` :D).

Regarding `.attrs.json` being in cwd: I've seen a few things such as `pkgs.closureInfo` making use of `.attrs.json` by assuming in the `buildCommand` that it's in the cwd. As `nix-shell` should emulate the build env as good as possible, this isn't such a big problem from my PoV. **However** I just realized that I have to make sure to **only** remove `.attrs.json` if it was added by `nix-shell` and didn't exist there already.

globin:
> Regarding `.attrs.json` being in cwd: I've seen a few things such as `pkgs.closureInfo` making use of `.attrs.json` by assuming in the `buildCommand` that it's in the cwd. As `nix-shell` should emulate the build env as good as possible, this isn't such a big problem from my PoV. **However** I just realized that I have to make sure to **only** remove `.attrs.json` if it was added by `nix-shell` and didn't exist there already.

There are quite a number of uses of .attrs.sh with structured-attrs, e.g. `passAsFile` is not supported and needs to be replaced by jq parsing .attrs.json also if one wants to create a builder not relying on bash it will most probably use .attrs.json, although that will create other problems with nix-shell.

Maybe calling nix-shell could create a subdir in the cwd to mitigate the cleanup problem? Although that creates other problems with nix-shell, when wanting to run only selected phases and not the whole `genericBuild`.
With Linux 5.11 overlayfs mounts are supported for non-privileged users, maybe that might be a viable route in the longer run.

regnat:
> from my understanding `exportReferences` (from derivation goal) is needed to be able to create the value for `exportReferencesGraph` and IMHO this variable should also appear in a `nix-shell`. Do you have a hint how to do that without using DerivationGoal itself? As I said in the OP, I'm not too experienced with Nix internals :)

Ahah, I missed that ;)
Looking at `exportReferences`, I think it could also be removed from `DerivationGoal` (except maybe the part that checks that we don’t try to access something outside of the closure, but that can probably be pulled out, and we can ignore it for nix-shell).

> Regarding `.attrs.json` being in cwd: I've seen a few things such as `pkgs.closureInfo` making use of `.attrs.json` by assuming in the `buildCommand` that it's in the cwd. As `nix-shell` should emulate the build env as good as possible, this isn't such a big problem from my PoV. **However** I just realized that I have to make sure to **only** remove `.attrs.json` if it was added by `nix-shell` and didn't exist there already.

You’d need to be a bit more careful − somehow keep a reference count of all the `nix-shell` that still point to it (for example I can open two `nix-shell` in a row, and then close the first one but keep the second open)… that sounds quite complex.

I wonder whether we could strengthen the semantics of `__structuredAttrs` to say that the `.attrs.json` file isn’t necessarily in `CWD` should be accessed via some environment variable `ATTRS_JSON_FILE` (with some magic to somehow stay backwards compatible, like have `ATTRS_JSON_FILE=$PWD/.attrs.json` for a normal build) so that we have more flexibility here.

Ma27:
>  Looking at exportReferences, I think it could also be removed from DerivationGoal (except maybe the part that checks that we don’t try to access something outside of the closure, but that can probably be pulled out, and we can ignore it for nix-shell).

Ok, that's good to know. I'll check if that's doable without having to fix too much stuff after that. @regnat do you think it makes sense to add it to the `Store` itself, the `{Parsed,}Derivation` (if yes, which? :)). Or do you have a more suitable place in mind?

> You’d need to be a bit more careful − somehow keep a reference count of all the nix-shell that still point to it (for example I can open two nix-shell in a row, and then close the first one but keep the second open)… that sounds quite complex.

Ohhhh, you're absolutely right, I didn't think of that, sorry!!

I actually think that `ATTRS_JSON_FILE` may be the way to go: for the builder we'd only need to set it to `$PWD/.attrs.json`, for `nix-shell` we could use a temporary directory. I'm happy to change that accordingly if @globin and @edolstra would be fine with that.

Ma27:
@regnat changed your primary two issues (moved code to generate `.attrs.{sh,json}` into `ParsedDerivation`, use env vars to reference these files in both the `nix-shell` & the build sandbox). Does the approach look okay to you now? If yes, I can resolve the remaining review comments to get this completely ready :) 

(would also appreciate a review from @Ericson2314 & @edolstra btw)

src/libstore/store-api.hh
```diff
@@ -695,6 +695,8 @@ public:
 
     const Stats & getStats();
 
+    StorePathSet exportReferences(const StorePathSet & storePaths, const StorePathSet & inputPaths);
```
Ericson2314, line 698: To be honest, I'm a bit way of putting more odd and ends (i.e. non-virtual methods) in the Store class (I prefer to keep "interfaces" slim like type classes.

But then again, this method doesn't have any other types it might be associated with (Store paths are "upstream" of stores), so maybe it's fine.

Ericson2314 (commented):
I'm on my phone so can't review so well, but very much liking the look of the split. That's good even if we weren't fixing functionality with it.

edolstra:
> I actually think that ATTRS_JSON_FILE may be the way to go: for the builder we'd only need to set it to $PWD/.attrs.json, for nix-shell we could use a temporary directory. I'm happy to change that accordingly if @globin and @edolstra would be fine with that.

Yes, I think that's a good idea. Writing to `./.attrs.json` breaks if I have two nix-shells with a different `-A` attribute.

BTW given the legacy status of `nix-shell`, it's more important to consider how structured attrs should work with `nix develop`. I think it mostly already works since `nix develop` builds a pseudo-derivation to extract the stdenv environment, and the structured attrs are present in that build.

regnat (commented):
> @regnat changed your primary two issues (moved code to generate .attrs.{sh,json} into ParsedDerivation, use env vars to reference these files in both the nix-shell & the build sandbox). Does the approach look okay to you now?

Yes, looks great. And I have to say that I’m utterly pleased by
image
:smile:

Ma27:
> Yes, I think that's a good idea. Writing to ./.attrs.json breaks if I have two nix-shells with a different -A attribute.

After realising that implementing this isn't as hard as I thought, I added it in 2a247b8

> I think it mostly already works since nix develop

I'm currently on `nix (Nix) 2.4pre20210503_6d2553a` (both CLI & daemon) and it doesn't work for me on @globin's [structured-attrs branch](https://github.com/nixos/nixpkgs/tree/structured-attrs):

```
λ ma27 [~/Projects/nixpkgs-structured-attrs] at  structured-attrs ?
→ nix develop .#hello --builders ''                                                                                                                                                                    [7c3a400378b]
error: builder for '/nix/store/s3d3hljq64g46riqv7y2hni62q3cy42w-hello-2.10-env.drv' failed to produce output path for output 'out' at '/nix/store/s3d3hljq64g46riqv7y2hni62q3cy42w-hello-2.10-env.drv.chroot/nix/store/pr9771b3i92n8k2lk98aibwgfl0crnq5-hello-2.10-env'
```

> BTW given the legacy status of nix-shell, it's more important to consider how structured attrs should work with nix develop

I'm happy to take a look at this as well at some point :) 

Btw given that 2.4 doesn't seem to have a roadmap and @globin told me that he'd like to try getting this into one of the upcoming NixOS releases (not 21.05 though), I'd like to raise the question whether it'd be okay for you to backport the `nix-shell` part. I know that the diff to 2.3 is well, kinda large and a lot of the code would probably have to be written a second time, but I'd like now know in the first place if this would be okay for you.

Ma27:
Btw @edolstra @regnat I also fixed the integration with `nix develop` and structured attrs :)

src/libstore/build/local-derivation-goal.cc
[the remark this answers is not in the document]
Ericson2314: Don't forget about this one!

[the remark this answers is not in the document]
Ma27: Yup, that one isn't checked in the todolist I added to the OP :)

src/libstore/build/derivation-goal.cc
```diff
@@ -143,6 +143,115 @@ void DerivationGoal::work()
...
+            /* This case is only relevant for the nix-shell */
+            outputs[i.first] = hashPlaceholder(i.first);
+        }
+    }
+    json["outputs"] = outputs;
+
+    /* Handle exportReferencesGraph. */
+    auto e = json.find("exportReferencesGraph");
+    if (e != json.end() && e->is_object()) {
+        for (auto i = e->begin(); i != e->end(); ++i) {
```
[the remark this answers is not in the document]
Ma27: AFAIU this won't work in both cases in the structured attrs stuff as we need both `*i` (i.e. the value) and `i.key()`.

src/libstore/build/derivation-goal.cc
```diff
@@ -143,6 +143,115 @@ void DerivationGoal::work()
...
+            std::ostringstream str;
+            {
+                JSONPlaceholder jsonRoot(str, true);
+                StorePathSet storePaths;
+                for (auto & p : *i)
+                    storePaths.insert(worker.store.parseStorePath(p.get<std::string>()));
+                worker.store.pathInfoToJSON(jsonRoot,
+                    exportReferences(storePaths), false, true);
+            }
+            json[i.key()] = nlohmann::json::parse(str.str()); // urgh
```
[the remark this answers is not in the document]
Ma27: Hmm, should this also happen in *this* PR?

Ma27:
Since I'm not sure if we should add even more changes in this already non-trivial (but kinda ready) PR and AFAIU the [other suggestion](https://github.com/NixOS/nix/pull/4770#discussion_r628320656) isn't applicable. I split the JSON and bash generation into two methods of `ParsedDerivation` which resolves the other two comments, so I'd re-request a review now :)

src/libstore/build/derivation-goal.cc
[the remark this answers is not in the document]
regnat: Nah, I just didn’t notice that this was only appearing in the diff because it was code that moved around rather than new code. So can be left as a follow-up :)

src/libstore/build/derivation-goal.cc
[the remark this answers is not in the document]
regnat: Oh yes, then you can do something like 

```c++
for (auto & [key, value] : e)
  …
```

(There’s an example of that in somewhere in the big code snippet under https://github.com/nlohmann/json#stl-like-access).

But it’s not like it’s new code either, so it can probably also be left for later

src/nix/get-env.sh
```diff
@@ -8,7 +8,13 @@ if [[ -n $stdenv ]]; then
...
 fi
 
-for __output in $outputs; do
+if [ -e .attrs.sh ]; then
+    __olist="${!outputs[@]}"
+else
+    __olist=$outputs
+fi
+
+for __output in $__olist; do
```
regnat, line 17: I wouldn’t mind seeing a test for this, because that’s way too many bash-isms for my brain to parse, so I’ve no idea how (and whether) it works

src/nix/develop.cc
```diff
@@ -255,6 +255,11 @@ struct Common : InstallableCommand, MixProfile
         // FIXME: properly unquote 'outputs'.
         StringMap rewrites;
         for (auto & outputName : tokenizeString<std::vector<std::string>>(replaceStrings(outputs->second.quoted, "'", ""))) {
+            std::regex ptrn(R"re(\[([A-z0-9]+)\]=.*)re");
+            std::smatch match;
+            if (std::regex_match(outputName, match, ptrn)) {
+                outputName = match[1];
+            }
```
regnat, line 262: That’s probably worth a comment, because right out of the blue this doesn’t make any sense to me 🤔

regnat (changes requested):
LGTM, except the `nix develop` part which looks like black magic to me, so is probably worth some explanations and tests

(or alternatively you can drop it for the time being and add it back as a follow-up)

Ma27:
@regnat Added a testcase and a few comments :)

regnat (approved):
Looks good, thanks!

It would probably be nicer if the `--redirect` feature of `nix develop` wasn’t depending on `stdenv`, but as `nix develop` is already partially tied to it I guess it’s not a really big issue

Ma27:
@edolstra anything missing here to get this merged? :)

src/libstore/build/local-derivation-goal.cc
```diff
@@ -1083,113 +1083,19 @@ void LocalDerivationGoal::initEnv()
...
-            }
-            json[i.key()] = nlohmann::json::parse(str.str()); // urgh
-        }
+    if (auto structAttrsJson = parsedDrv->prepareStructuredAttrs(inputRewrites, worker.store, inputPaths)) {
+        auto json = structAttrsJson.value();
+        auto jsonSh = parsedDrv->writeStructuredAttrsShell(json);
+
+        writeFile(tmpDir + "/.attrs.sh", rewriteStrings(jsonSh, inputRewrites));
+        chownToBuilder(tmpDir + "/.attrs.sh");
+        env["ATTRS_SH_FILE"] = tmpDir + "/.attrs.sh";
```
edolstra, line 1094: Maybe prefix this environment variable with `NIX_` (like `NIX_LOG_FD`, `NIX_BUILD_CORES` etc).

src/libstore/build/local-derivation-goal.cc
```diff
@@ -1083,113 +1083,19 @@ void LocalDerivationGoal::initEnv()
...
+        writeFile(tmpDir + "/.attrs.json", rewriteStrings(json.dump(), inputRewrites));
+        chownToBuilder(tmpDir + "/.attrs.json");
+        env["ATTRS_JSON_FILE"] = tmpDir + "/.attrs.json";
```
edolstra, line 1097: Idem

src/libstore/build/local-derivation-goal.cc
```diff
@@ -1083,113 +1083,19 @@ void LocalDerivationGoal::initEnv()
...
-                JSONPlaceholder jsonRoot(str, true);
-                StorePathSet storePaths;
-                for (auto & p : *i)
-                    storePaths.insert(worker.store.parseStorePath(p.get<std::string>()));
-                worker.store.pathInfoToJSON(jsonRoot,
-                    exportReferences(storePaths), false, true);
-            }
-            json[i.key()] = nlohmann::json::parse(str.str()); // urgh
-        }
+    if (auto structAttrsJson = parsedDrv->prepareStructuredAttrs(inputRewrites, worker.store, inputPaths)) {
```
edolstra, line 1088: Does `prepareStructuredAttrs` need `inputRewrites`? It should be possible to apply the rewrites to the `structAttrsJson.value()` right?

src/libstore/parsed-derivations.cc
```diff
@@ -121,4 +123,116 @@ bool ParsedDerivation::substitutesAllowed() const
...
+            json[i.key()] = nlohmann::json::parse(str.str()); // urgh
+        }
+    }
+
+    return json;
+}
+
+std::string ParsedDerivation::writeStructuredAttrsShell(nlohmann::json & json)
+{
+    /* As a convenience to bash scripts, write a shell file that
```
edolstra, line 170: This comment should be moved to the declaration in the header file. Then if we ever generate API docs it will show up.

src/libstore/parsed-derivations.hh
```diff
@@ -36,6 +36,9 @@ public:
     bool willBuildLocally(Store & localStore) const;
 
     bool substitutesAllowed() const;
+
+    std::optional<nlohmann::json> prepareStructuredAttrs(std::optional<StringMap> inputRewrites, Store & store, const StorePathSet & inputPaths);
+    std::string writeStructuredAttrsShell(nlohmann::json & json);
```
edolstra, line 41: Does this method actually rely on anything other than `json`? In that case it should be a free function.

src/libstore/store-api.hh
edolstra, line 698: This needs a comment about what `exportReferences` does. Also the name is not ideal, since it doesn't really export anything... (Especially confusing since there is a `nix-store --export` command.)

src/libstore/build/local-derivation-goal.cc
Ma27: @edolstra Yeah I kept it in this codepath as it used to be there previously and seemed kinda correct to do the rewrite when building the datastructure. But if you'd prefer to move it to LocalDerivationGoal, that'd work for me as well :)

Ma27: @edolstra OK actually I think you're right, given that `nix-shell` uses `std::nullopt` for input rewrites it's not a strict part of `prepareStructuredAttrs`, so I moved it to `LocalDerivationGoal`. Gonna re-request a review from @regnat to make sure I didn't do anything wrong with CA integrations there :)

src/libstore/parsed-derivations.cc
```diff
@@ -134,14 +134,8 @@ std::optional<nlohmann::json> ParsedDerivation::prepareStructuredAttrs(std::opti
...
     for (auto & i : drv.outputs) {
-        if (inputRewrites) {
-            /* The placeholder must have a rewrite, so we use it to cover both the
-               cases where we know or don't know the output path ahead of time. */
-            outputs[i.first] = rewriteStrings(hashPlaceholder(i.first), inputRewrites.value());
-        } else {
-            /* This case is only relevant for the nix-shell */
-            outputs[i.first] = hashPlaceholder(i.first);
-        }
+        /* This case is only relevant for the nix-shell */
```
regnat, line 137: I suppose this comment doesn’t make sense anymore, dose it?

regnat (commented):
Just had a look at the last commit (9952e4edd47b421571ee4305275d45d21dc320d4), but it looks OK to me.

The thing with the rewrites of the json is a bit weird, but it was already weird before the PR, so it’s not really a problem for now

src/libstore/parsed-derivations.cc
Ma27: Good catch, fixed!

Ma27:
cc @edolstra anything else tbd to get this merged? :)

Ma27:
@edolstra @regnat merge master into this branch. Another review would be appreciated :) 

If nothing else is to be done, we can also merge this btw :)

PR #4773, opened 2021-05-05
doc: builtins: use a definition list

alyssais:
This looks a _lot_ better (and is also more semantically meaningful).

Since this list is generated in a Nix expression, I don't think using HTML here is going to be the thing that puts people off modifying this part of the documentation!


regnat (commented):
That’s nicer indeed, but that’s also not in line with the rest of the documentation which uses this encoding a lot. So not sure that much it’s a net win overall 🤷🏻‍♂️

(That being said, I’m quite fan of the bits of the PR that put the arguments in the code block, I find that much easier to parse visually)

fricklerhandwerk:
LGTM!

alyssais:
This should only affect the manual, not the :doc output generated by lowdown.

PR #4787, opened 2021-05-10
Implement `builtins.floor` and `builtins.ceil` using the C library functions internally

### LINKED ISSUE #4782: Add builtins.floor and builtins.ceiling
There's not a great way to do a `float -> int` conversion in nix.

Use case:
```
some_hard_constraint = 500;
some_soft_constraint = builtins.floor (some_hard_constraint * .9);
```

Using `toString` then string manipulation to achieve the same goal, but this not ergonomic, and much more expensive than a raw number -> number conversion.

should be relatively easy to implement http://www.cplusplus.com/reference/cmath/floor/

Ma27:
Closes #4782

Note: even though the type is internally called `NixFloat`, it's
actually a `double`.

cc @edolstra @jonringer


edolstra (approved):

jonringer (approved):
That was quick :). Awesome!

PR #4790, opened 2021-05-10
buildenv: Create dangling symlinks in environments

Infinisil:
Previously the buildenv builder skipped symlinks completely if they couldn't be resolved. And since the builder is running in the Nix sandbox, it can't resolve any links that aren't accessible in the sandbox, even if they would be resolveable outside the sandbox.

This commit changes this behavior: If symlinks can't be resolved, they are still created in the environment, on the assumption that they are resolveable for the end user. A warning during the build is still issued.

- [x] Added a test ensuring that this new behavior can be used
- [x] Tested it manually with
  ```bash
  $ sudo outputs/out/bin/nix-env -p $PWD/profile/main -ri -E \
    '_: (import <nixpkgs> {}).runCommand "l" {} "mkdir $out && ln -s ${toString builtins.currentTime} $out/foo"'
    /nix/store/gy0rbk15n187irrpfvc4mqrnnbiy2rdh-l.drv
  building '/nix/store/gy0rbk15n187irrpfvc4mqrnnbiy2rdh-l.drv'...
  building '/nix/store/a9cm6361zlz0hd3zlhbrm6n7vsxz705k-user-environment.drv'...
  warning: creating dangling symlink '/nix/store/4avrk1xyp0bk57347pg3sc8qyzi7f9l5-user-environment/foo'
  $ realpath profile/main/foo/
  /nix/store/q3kvm0gk4w8z8mah9vgh2myf61abmak9-l/1620672396
  ```
  (`sudo` so that not my older nix-daemon version is used)

### DIFF
--- a/tests/user-envs.sh
+++ b/tests/user-envs.sh
@@ -13,7 +13,7 @@ mkdir -p $TEST_HOME
 nix-env --switch-profile $profiles/test
 
 # Query available: should contain several.
-test "$(nix-env -f ./user-envs.nix -qa '*' | wc -l)" -eq 6
+test "$(nix-env -f ./user-envs.nix -qa '*' | wc -l)" -eq 7
 outPath10=$(nix-env -f ./user-envs.nix -qa --out-path --no-name '*' | grep foo-1.0)
 drvPath10=$(nix-env -f ./user-envs.nix -qa --drv-path --no-name '*' | grep foo-1.0)
 [ -n "$outPath10" -a -n "$drvPath10" ]
@@ -158,12 +158,13 @@ nix-env -q '*' | grep -q foo-2.0
 nix-env -e '*'
 (! nix-env -i foo-1.0 foo-2.0)
 
-# Installing "*" should install one foo and one bar.
+# Installing "*" should install one foo and one bar and the dangling-link
 nix-env -e '*'
 nix-env -i '*'
-test "$(nix-env -q '*' | wc -l)" -eq 2
+test "$(nix-env -q '*' | wc -l)" -eq 3
 nix-env -q '*' | grep -q foo-2.0
 nix-env -q '*' | grep -q bar-0.1.1
+nix-env -q '*' | grep -q dangling-link
 
 # Test priorities: foo-0.1 has a lower priority than foo-1.0, so it
 # should be possible to install both without a collision.  Also test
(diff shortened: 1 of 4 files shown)


Infinisil:
@lheckemann That issue doesn't appear to be a problem anymore even before this PR, since Nix only issues a warning for dangling symlinks in any case.

roberth:
> I'm not sure how to "fix" that, can we just ignore CI? Checking against a different daemon version in CI only makes sense if the change to test is in the client.

You can disable a test selectively. Compatibility tests specify `NIX_DAEMON_PACKAGE`. You could parse its version. These test are fairly new, so I don't know of an established pattern for this. I sure am glad we have these tests new though!

tests/user-envs.sh
```diff
@@ -179,3 +180,15 @@ nix-env --set $outPath10
 [ "$(nix-store -q --resolve $profiles/test)" = $outPath10 ]
 nix-env --set $drvPath10
 [ "$(nix-store -q --resolve $profiles/test)" = $outPath10 ]
+
+# This tests behavior of the daemon, and as such only needs to run
+# when the daemon is the self version
+if [[ -n "${NIX_CLIENT_PACKAGE:-}" && -n "${NIX_DAEMON_PACKAGE:-}"
+    && "$NIX_CLIENT_PACKAGE" == "$NIX_DAEMON_PACKAGE" ]]; then
```
Infinisil, line 187: Let's see if this works. This should only make the test run when client version matches daemon version.

Infinisil: It does :)

tests/user-envs.sh
```diff
@@ -179,3 +180,15 @@ nix-env --set $outPath10
 [ "$(nix-store -q --resolve $profiles/test)" = $outPath10 ]
 nix-env --set $drvPath10
 [ "$(nix-store -q --resolve $profiles/test)" = $outPath10 ]
+
+# This tests behavior of the daemon, and as such only needs to run
+# when the daemon is the self version
```
roberth, line 185:
```suggestion
# This tests behavior of the daemon that has not been backported to
# previous versions, so for now we don't run it when we test with
# other versions.
```

Infinisil:
I think the work here could be useful, but I personally don't have an interest in it anymore. Whoever does is free to reuse my work from this PR.

Issue #4256, opened 2020-11-13
nix::DerivationGoal::registerOutputs(): Assertion `newInfo.ca' failed.


teto:
**Describe the bug**

nix crashes when rebuilding my flake based configuration. This is the second time this specific errors to me but it is not deterministic as rerunning the same command succeeds afterwards.

some context
```
❯ sudo coredumpctl debug                                                                                                                                                              ~/nixpkgs
           PID: 2085 (nix)
           UID: 0 (root)
           GID: 0 (root)
        Signal: 6 (ABRT)
     Timestamp: Fri 2020-11-13 14:03:08 CET (1min 12s ago)
  Command Line: nix --experimental-features nix-command flakes build git+file:///home/teto/home#nixosConfigurations."jedha".config.system.build.toplevel --impure --no-write-lock-file --override-input nixpkgs-teto /home/teto/nixpkgs --override-input hm /home/teto/hm --override-input nova /home/teto/nova/nova-nix --out-link /tmp/nixos-rebuild.stYtk8/result
    Executable: /nix/store/fb8rfml6sr1d92zyfxzz1ax75a8djjcl-nix-2.4pre20201102_550e11f/bin/nix
 Control Group: /user.slice/user-1000.slice/session-4.scope
          Unit: session-4.scope
         Slice: user-1000.slice
       Session: 4
     Owner UID: 1000 (teto)
       Boot ID: 098bc23ed9fe47e6b5fbcd2e39e68680
    Machine ID: 6ba303eb0232411fbc60369b9affc9e7
      Hostname: jedha
       Storage: /var/lib/systemd/coredump/core.nix.0.098bc23ed9fe47e6b5fbcd2e39e68680.2085.1605272588000000.lz4
       Message: Process 2085 (nix) of user 0 dumped core.
```
and the coredump 
```
#0  0x00007fd6cb5c641a in raise () from /nix/store/kah5n342wz4i0s9lz9ka4bgz91xa2i94-glibc-2.32/lib/libc.so.6
#1  0x00007fd6cb5b0523 in abort () from /nix/store/kah5n342wz4i0s9lz9ka4bgz91xa2i94-glibc-2.32/lib/libc.so.6
#2  0x00007fd6cb5b041f in __assert_fail_base.cold.0 () from /nix/store/kah5n342wz4i0s9lz9ka4bgz91xa2i94-glibc-2.32/lib/libc.so.6
[...]

PR #4798, opened 2021-05-12
Relock wait for build slot goals

matthewbauer:
When we don’t have enough free job slots to run a goal, we put it in
the waitForBuildSlot list & unlock its output locks. This will
continue from where we left off (tryLocalBuild). However, we need the
locks to get reacquired when/if the goal ever restarts. So, we need to
send it back through tryToBuild to get reqacquire those locks.

I think this bug was introduced in
https://github.com/NixOS/nix/pull/4570. It leads to some builds
starting without proper locks.

/cc @edolstra @ericson2314 @regnat

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -153,6 +153,7 @@ void LocalDerivationGoal::killChild()
 void LocalDerivationGoal::tryLocalBuild() {
     unsigned int curBuilds = worker.getNrLocalBuilds();
     if (curBuilds >= settings.maxBuildJobs) {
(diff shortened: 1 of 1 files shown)


Ericson2314:
Woah! I am not quite sure how I introduced that, but this might also factor into the #4256 stuff in that I am assuming we would have lock already in the input addressed case.

matthewbauer:
> I am not quite sure how I introduced that, but this might also factor into the #4256 stuff in that I am assuming we would have lock already in the input addressed case.

That issue predates #4570, so I wouldn't think so, but might be worth a shot.

PR #4801, opened 2021-05-13
Fix tokenize output names in DerivedPath

matthewbauer:
This should fix the issue described in
https://discourse.nixos.org/t/derivation-does-not-have-wanted-outputs-dev-out/12905.

Specifically, we get an error of

  error: derivation '/nix/store/_.drv' does not have wanted outputs 'dev,out'

when a path like /nix/store/_.drv!dev,out is sent to the daemon.

/cc @ericson2314 @edolstra

### DIFF
--- a/src/libstore/derived-path.cc
+++ b/src/libstore/derived-path.cc
@@ -62,7 +62,7 @@ DerivedPath::Built DerivedPath::Built::parse(const Store & store, std::string_vi
     auto outputsS = s.substr(n + 1);
     std::set<string> outputs;
     if (outputsS != "*")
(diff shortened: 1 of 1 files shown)


Ericson2314 (approved):
Wow....just wow.... thank you so much @matthewbauer!!!

Ericson2314:
I wonder if we should try to run all the tests with the daemon or something.

Mic92 (approved):
Tested and works for me.

edolstra:
Thanks!

grahamc:
> I wonder if we should try to run all the tests with the daemon or something, that would have caught this when I wrote it.

What do you think it would take to do this?

PR #4804, opened 2021-05-14
Filter ANSI codes when piping nix commands

matthewbauer:
This removes \e[... codes from sequences so that commands like

  nix search > ~/out.list

Don’t retain the ANSI formating.

### DIFF
--- a/src/libmain/progress-bar.cc
+++ b/src/libmain/progress-bar.cc
@@ -464,7 +464,10 @@ public:
             Logger::writeToStdout(s);
             draw(*state);
         } else {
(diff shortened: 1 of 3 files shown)


regnat:
Is there also a way to force this one way or another? For example I quite frequently run `nix search nixpkgs something | less` and I wouldn’t like to loose the formatting in that case

emilazy:
Many programs support controlling this with environment variables like `CLICOLOR`/`CLICOLOR_FORCE`. It’d be nice to respect those rather than adding yet another program-specific convention to learn.

Issue #4739, opened 2021-04-24
nix search without a term should error


SuperSandro2000:
**Is your feature request related to a problem? Please describe.**
When you run ``nix search`` without any query it lists you all packages and freezes your terminal for multiple seconds until the Ctrl + C comes through.

**Describe the solution you'd like**
It should error instead and force the user to provide a query.

**Describe alternatives you've considered**
Writing shell script around the problem myself.

**Additional context**
If that functionality is desired it could be moved to a flag.

regnat:
Related: https://github.com/NixOS/nix/issues/3553

Issue #3553, opened 2020-05-01
Make `nix search` without search terms show the help message


milibopp:
**Is your feature request related to a problem? Please describe.**
I noticed that new users are somewhat confused by `nix search` behaviour, when called without search terms. A common expectation of modern command line tools is that you call them with no arguments to learn more about how to use them. This problem is compounded by the fact, that listing all packages can take some time to evaluate, especially when used with `-u`, which you are warned about. So a new user notices that they have unintentionally triggered an unknown long-running process, but they only wanted to get a quick information.

**Describe the solution you'd like**
When there are no search terms, display the help message (like `nix search --help`) instead of listing all packages.

**Describe alternatives you've considered**
I see two primary alternatives:
- Leave as is.
- Change the behaviour as described and add an option to display all packages (e.g. `--all`). This could make sense, if people actually use this feature on purpose.

**Additional context**
n/a

bew:
@edolstra Hello, could you elaborate on why this is being pushed to a later release time and time again?

Do you (Nix team) agree with this change? would you take a PR implementing this change?

PR #4809, opened 2021-05-14
Add nix search --all

sophrosyne97:
Since calling nix search without a search string is likely to occur by mistake it should not try to match for all packages.
Given how long particular full searches can take a full match should occur from the '--all' flag.
Instead, empty searches now throw an error to the user.

Closes issues #4739 and #3553

(Note: for #3553 I opted to make nix-search throw an error on an empty search instead of show the help message due to it being an InstallableCommand, which meant that it would only show the help message if there was a flake in the current directory and throw an error otherwise. )

### DIFF
--- a/src/nix/search.md
+++ b/src/nix/search.md
@@ -5,7 +5,7 @@ R""(
 * Show all packages in the `nixpkgs` flake:
 
   ```console
-  # nix search nixpkgs
+  # nix search -a nixpkgs
   * legacyPackages.x86_64-linux.AMB-plugins (0.8.1)
     A set of ambisonics ladspa plugins
 
@@ -34,7 +34,7 @@ R""(
 * Show all packages in the flake in the current directory:
 
   ```console
-  # nix search
+  # nix search -a
   ```
 
 * Search for Firefox or Chromium:
@@ -56,8 +56,7 @@ flake) for packages whose name or description matches all of the
 regular expressions *regex*.  For each matching package, It prints the
 full attribute name (from the root of the installable), the version
 and the `meta.description` field, highlighting the substrings that
-were matched by the regular expressions. If no regular expressions are
-specified, all packages are shown.
+were matched by the regular expressions.
 
 # Flake output attributes
 
--- a/tests/search.sh
+++ b/tests/search.sh
@@ -19,10 +19,10 @@ clearCache
 
 ## Search expressions
 
-# Check that empty search string matches all
-nix search -f search.nix '' |grep -q foo
-nix search -f search.nix '' |grep -q bar
-nix search -f search.nix '' |grep -q hello
+# Check that '--all' flag matches all
+nix search -af search.nix |grep -q foo
+nix search --all -f search.nix |grep -q bar
+nix search -af search.nix |grep -q hello
 
 ## Tests for multiple regex/match highlighting
 
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -25,8 +25,17 @@ struct CmdSearch : InstallableCommand, MixJSON
 {
     std::vector<std::string> res;
 
+    bool all = false;
+
     CmdSearch()
     {
+        addFlag({
+            .longName = "all",
+            .shortName = 'a',
+            .description = "Show all packages for a flake.",
+            .handler = {&all, true},
+        });
+
         expectArgs("regex", &res);
     }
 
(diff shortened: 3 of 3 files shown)


ShamrockLee:
Tested using `nix shell github:sophrosyne97/nix/master`
LGTM

 - system: `"x86_64-linux"`
 - host os: `Linux 5.4.112, NixOS, 20.09.20210430.be58d0f (Nightingale)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.4pre20210326_dd77f71`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

src/nix/search.md
```diff
@@ -56,8 +56,7 @@ flake) for packages whose name or description matches all of the
 regular expressions *regex*.  For each matching package, It prints the
 full attribute name (from the root of the installable), the version
 and the `meta.description` field, highlighting the substrings that
-were matched by the regular expressions. If no regular expressions are
-specified, all packages are shown.
+were matched by the regular expressions.
```
regnat, line 59:
```suggestion
were matched by the regular expressions. If `-a` or `--all` is passed,
all packages are shown
```

tests/search.sh
```diff
@@ -19,7 +19,7 @@ clearCache
...
 ## Search expressions
 
-# Check that empty search string matches all
-nix search -f search.nix '' |grep -q foo
-nix search -f search.nix '' |grep -q bar
-nix search -f search.nix '' |grep -q hello
+# Check that '--all' flag matches all
+nix search -af search.nix |grep -q foo
+nix search --all -f search.nix |grep -q bar
+nix search -af search.nix |grep -q hello
```
regnat, line 25:
```suggestion
nix search -af search.nix |grep -q hello
# Check that an empty search string raises an error
nix search -f search.nix || fail "Running 'nix search' with an empty search string should fail"
```

regnat (changes requested):
A couple of small suggestions to extend the documentation and the tests, but otherwise looks good

tests/search.sh
regnat: Mh I wrote that a bit too fast, should probably rather be something like

```bash
(! nix search -f search.nix ) || fail "Running 'nix search' with an empty search string should fail"
```

edolstra:
I'm not sure I agree with the premise that not specifying a regex is likely an error. `nix search <flake>` is a convenient way to show all the packages in a flake.

>  Given how long particular full searches can take a full match should occur from the '--all' flag.

Every search is a full search. The regex just restricts what gets printed.

sophrosyne97:
> I'm not sure I agree with the premise that not specifying a regex is likely an error.

Perhaps not likely for someone familiar with the command, but I would say it is likely an error for new users.
I made the mistake #3553 mentions myself when I was first trying out the experimental features for nix and couldn't really tell if the full search was a bug or a feature as it was happening. This flag would save new users time and from confusion in the future.

Although an empty search is quick, it is inconsistent with most common search commands which assume that every search begins as an empty one. This idiosyncrasy can make even those familiar with the command error prone as shown with #4739, which in the long run can make the feature not so quick. I thought that the short flag '-a' might bridge the gap between being convenient and quick while also demanding explicit user intent for a full search. That way we can stop the unpleasant experience of having the terminal lock up and make the command a little more predictable.

tests/search.sh
sophrosyne97: added and tested, builds successfully

src/nix/search.md
sophrosyne97: added the changes

SuperSandro2000:
@sophrosyne97 can you rebase?

sophrosyne97:
> @sophrosyne97 can you rebase?

Rebased and tested.

SuperSandro2000:
Can you rebase on master and fix the failing test?

roberth:
This PR was discussed in the Nix team meeting today:

- @edolstra, @roberth prefer the math simplicity of current implementation
- @tomberek: probably bigger issue is forgetting to specify the flakeref
- @edolstra: let's close or postpone
- @ericson2314: It seems what is what is *actually* bugging people is that control-C is not working well (a separate issue). So they accidentally spam their terminally with a huge list of packages and then cannot easily get out of it. So if we solve that problem (fix C-c, page output) then this won't be so bad.

Decision: close

Issue #4661, opened 2021-03-23
error: cannot operate on an output of unbuilt content-addresed derivation


Ninlives:
**Describe the bug**

I keep getting this weired error when using nix shell:

![image](https://user-images.githubusercontent.com/17873203/112155283-1be88f00-8c20-11eb-917e-2f2a07d0e34d.png)

**`nix --version` output**
nix (Nix) 2.4pre20210308_1c0e3e4

oxalica:
Cannot reproduce on 2.4pre20210317_8a5203d

Ninlives:
This seems to be caused by the 'ca-derivation' feature, the problem disappear after remove that feature.

regnat:
Should be fixed by https://github.com/NixOS/nix/pull/4724

tomberek:
I still get this with things like path-info or build with `--dry-run` on 2.7.0pre. @thufschmitt are there some code paths that don't utilize the fallback?

SuperSandro2000:
I am getting this often when building things with a remote builder.

thufschmitt:
Closing since #6275 and #6276 are closed

mikepurvis:
This is definitely still an issue with `nix realisation info`. Trivial reproduction in a container:

```
$ buildah from docker.io/nixos/nix:2.8.0pre20220419_1cdad10
$ buildah run nix-working-container bash
# echo "experimental-features = nix-command flakes ca-derivations" >> /etc/nix/nix.conf
# nix realisation info nixpkgs#hello
these 4 paths will be fetched (6.60 MiB download, 31.35 MiB unpacked):
  /nix/store/6w8g7njm4mck5dmjxws0z1xnrxvl81xa-glibc-2.34-115
  /nix/store/dgpfbg2pykka81v94wka6ck9vi6dffa5-libidn2-2.3.2
  /nix/store/gs5yll0jabliajg4ds0bfvy73mjsirxd-libunistring-0.9.10
  /nix/store/js9xgkjs8qhjv6r48g1dk9ian8iszg8w-hello-2.12
error: cannot operate on an output of unbuilt content-addressed derivation 'sha256:ae2424d11cb50c81329dd1ecea5ca690090e45c2c3d6bd0e27eafcb10c128cfa!out'
# echo $?
1
```

thufschmitt:
> This is definitely still an issue with nix realisation info.

[...]

PR #4818, opened 2021-05-17
Enforce the use of properly built paths in libcmd

regnat:
Replace `DerivedPathWithHints` by a new `BuiltPath` type that serves as a proof that the corresponding path has been built.

This is mostly an internal refactoring at that point, but it should enforce that we can’t hit errors like https://github.com/NixOS/nix/issues/4661 by (generally) forbidding the CLI from accessing the output path of an unbuilt derivation.

### DIFF
--- a/src/libcmd/command.hh
+++ b/src/libcmd/command.hh
@@ -143,7 +143,7 @@ private:
 };
 
 /* A command that operates on zero or more store paths. */
-struct RealisedPathsCommand : public InstallablesCommand
+struct BuiltPathsCommand : public InstallablesCommand
 {
 private:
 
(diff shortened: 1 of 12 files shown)


src/libcmd/command.hh
```diff
@@ -216,7 +216,7 @@ static RegisterCommand registerCommand2(std::vector<std::string> && name)
     return RegisterCommand(std::move(name), [](){ return make_ref<T>(); });
 }
 
-DerivedPathsWithHints build(ref<Store> store, Realise mode,
+BuiltPaths build(ref<Store> store, Realise mode,
     std::vector<std::shared_ptr<Installable>> installables, BuildMode bMode = bmNormal);
```
Ericson2314, line 220: Just the thing I was hoping to see at some point :smiley:

Issue #3605, opened 2020-05-21
macOS: unexpected EOF reading a line


roberth:
**Describe the bug**

Nix has failed to build on darwin with this log.
https://hydra.nixos.org/build/119296096/nixlog/1

It did succeed afterwards, so unless someone recognizes what's the problem in the log, it's ok to close this and move on.

The failure happens in test case tests/user-envs.sh, with the last couple of lines being:

```
    29 store paths deleted, 0.01 MiB freed
    + test -e /private/tmp/nix-build-nix-2.3.4.drv-0/nix-test/store/8fid272lbh1milx38cni6wi6dx7v8z7b-foo-1.0
    + '[' -e /private/tmp/nix-build-nix-2.3.4.drv-0/nix-test/store/kzm50sz0sp6gsvav10kx8xasppw0yrql-foo-2.0 ']'
    + nix-env -e '*'
    uninstalling 'foo-1.0'
    uninstalling 'bar-0.1'
    building '/private/tmp/nix-build-nix-2.3.4.drv-0/nix-test/store/prvd5da3s52k2gr2pvx510wvn7shyjqh-user-environment.drv'...
    created 0 symlinks in user environment
    ++ nix-env -q '*'
    ++ wc -l
    + test 0 -eq 0
    + nix-env -i foo
    installing 'foo-2.0'
    these derivations will be built:
      /private/tmp/nix-build-nix-2.3.4.drv-0/nix-test/store/wsrlj1hjzncx3g2i1jp65lbb1v3py2wp-foo-2.0.drv
    building '/private/tmp/nix-build-nix-2.3.4.drv-0/nix-test/store/wsrlj1hjzncx3g2i1jp65lbb1v3py2wp-foo-2.0.drv'...
    error: unexpected EOF reading a line
```

**Steps To Reproduce**

This may be a race condition or other intermittent bug.

The entire log of the build failure is available here:
https://hydra.nixos.org/build/119296096/nixlog/1

**Expected behavior**

Nix just builds on darwin without a test failure.

**`nix-env --version` output**

N/A

**Additional context**

edolstra:
This test has been randomly failing for a while.

domenkozar:
In case it's relevant, it also fails on bar:

```
    + nix-env -i bar-0.1
    installing 'bar-0.1'
    these derivations will be built:
[...]

PR #4828, opened 2021-05-18
Restore an accidentally suppressed negation

regnat:
Accidentally removed in ca96f5219489c1002499bfe2c580fdd458219144. This caused `nix run` to systematically fail with

```
error: app program '/nix/store/…' is not in the Nix store
```


andir:
Perhaps time to add a simple run invocation to the tests?

regnat:
> Perhaps time to add a simple run invocation to the tests?

As a matter of fact, there is one, and it caught this all right. The problem (at least on my side) is https://github.com/NixOS/nix/issues/3605 which tends to make me ignore unexpected CI failures as “oh nvm, that’s the flaky darwin test striking again” (which is pretty stupid I guess, but…)

Issue #4822, opened 2021-05-17
Recent nix version broke build on Darwin


jgeerds:
**Describe the bug**

Since `2.4pre20210503_6d2553a` I can no longer build a (modified) Version of `python3Packages.pillow` on Darwin:

```
[…]
clang -I/nix/store/5569680f49bm0m4gnzck4wikp04nk3wr-libwebp-1.1.0/include -I/nix/store/cvkb1yxprk3k9gbgdqcr7s5jmg3ajfk2-libxcb-1.14-dev/include -I/nix/store/afnbyi0965q2fc218vx3mjay1wffpzip-freetype-2.10.4-dev/include -I/nix/store/mi97zxhpwllxdi6k6jq2w6qk58fi6kwj-openjpeg-2.4.0-dev/include/openjpeg-2.4 -I/private/tmp/nix-build-python3.8-Pillow-8.2.0debug.drv-1/Pillow-6.1.0/src/libImaging -I/nix/store/fn854v53z5696mz5l5d4pjxn13w8r9jz-libjpeg-turbo-2.0.6-dev/include -I/nix/store/mi97zxhpwllxdi6k6jq2w6qk58fi6kwj-openjpeg-2.4.0-dev/include -I/nix/store/lhs3fpqmmxsx4xnvzsjxl8n6hrc8hnnz-libtiff-4.2.0-dev/include -I/nix/store/rxfh4xff3v5kgrjvx6ix78jww6fndysv-zlib-1.2.11-dev/include -I/nix/store/f5q0jcnjnw498azlkd02xak18cxz1gs6-lcms2-2.12-dev/include -I/nix/store/d818466c3fi2yiscaa748c4nfx6f75jc-libimagequant-2.14.1/include -I/nix/store/5569680f49bm0m4gnzck4wikp04nk3wr-libwebp-1.1.0/include -I/nix/store/cvkb1yxprk3k9gbgdqcr7s5jmg3ajfk2-libxcb-1.14-dev/include -I/nix/store/l4q9nfrq3wz4qgk5ikbzfcs7iqdb32y1-python3-3.8.9/include -I/nix/store/l4q9nfrq3wz4qgk5ikbzfcs7iqdb32y1-python3-3.8.9/include/python3.8 -c src/_imagingcms.c -o build/temp.macosx-10.6-x86_64-3.8/src/_imagingcms.o
[...]

PR #4831, opened 2021-05-18
Fix extra slash in canonPath output

matthewbauer:
When you have a symlink like:

  /tmp -> ./private/tmp

you need to resolve ./private/tmp relative to /tmp’s dir: ‘/’. Unlike
any other path output by dirOf, / ends with a slash. We don’t want
trailing slashes here since we will append another slash in the next
comoponent, so clear s like we would if it was a symlink to an absoute
path.

This should fix at least part of the issue in
https://github.com/NixOS/nix/issues/4822, will need confirmation that
it actually fixes the problem to close though.

Introduced in f3f228700a52857fe6e8632df4e935551ea219ff.

/cc @ony @edolstra

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -155,6 +155,9 @@ Path canonPath(const Path & path, bool resolveSymlinks)
                     s.clear();  /* restart for symlinks pointing to absolute path */
                 } else {
                     s = dirOf(s);
+                    if (s == "/") {  // we don’t want trailing slashes here, which dirOf only produces if s = /
+                        s.clear();
+                    }
                 }
             }
         }


ony (approved):
In case of `/tmp` points to `./private/tmp` or to `private/tmp` :

1. After reading symlink we end up with `s = "/"; temp = "./private/tmp/abc";` as noted already `dirOf("/tmp")` => `"/"`.
2. We go to next cylce with `i` poiting to `.` in `./private/tmp/abc`
3. We ignore `.` as per  second `if` up until `/` which in its turn is skipped as per first `if`.
4. We end up in `/* Normal compoennt... */` where we add `'/'`.

Indeed leading extra-slashes. Thank you for spotting this.

ony:
I suspect that "little surprise" with double-slashes might also cause hash invalidation.

Checked [build output](https://github.com/NixOS/nix/pull/4679/checks?check_run_id=2233473703#step:6:470), but only `nix` was built and without second pass to use newly built `nix`.

And at that time I only looked at green checkbox. So I guess other contributors might not look into output as well without any explicit marker like "Warning: with this change part of the cache will be invalidated".

PR #4833, opened 2021-05-19
Always send the realisations as JSON

regnat:
Align all the worker protocol with `buildDerivation` which inlines the
realisations as one opaque json blob.
That way we don’t have to bother changing the remote store protocol
when the definition of `Realisation` changes, as long as we keep the
json backwards-compatible


Ericson2314 (approved):
While I had some qualms about the JSON (vs say bson / cbor) at some point. This change *strictly speaking* is just making the cannonical `Realization` serialization be used rather than manually breaking down / building up the realization into pieces to (de-)serialize separately. That I can always get behind :).

PR #4834, opened 2021-05-19
Extract a generic `computeClosure` function

regnat:
Move the `closure` logic of `computeFSClosure` to its own (templated) function.

This doesn’t bring much by itself (except for the ability to properly
test the “closure” functionality independently from the rest), but it
allows reusing it (in particular for the realisations which will require
a very similar closure computation)


Ericson2314:
:) It's already merged, but wanted to say I really like this.

Issue #4835, opened 2021-05-19
“Two Glibc” issue


regnat:
**Describe the bug**

The current design of ca derivations make it possible to have a closure in which the same derivation is realised twice with two different outputs paths (what Eelco’s PhD thesis calls [equivalence class colision](https://edolstra.github.io/pubs/phd-thesis.pdf#page=155)). This can cause some ugly breakages and whatnot.

**Steps To Reproduce**

~~Didn’t put a nice repro yet, but in the meantime the idea to get there is explained in the paper linked above.~~

See https://github.com/NixOS/nix/blob/6481155a60de00386ede63340013b5234a58102c/tests/ca/duplicate-realisation-in-closure.sh

**Expected behavior**

Some better heuristics might be tried later, but the current expected behavior is to just reject the substitution if it leads to a duplication in the closure.

regnat:
Added a link to a Nix test that reproduces the issue

PR #4836, opened 2021-05-19
Track the dependencies of CA realisations

regnat:
Fix https://github.com/NixOS/nix/issues/4249 , and more importantly is a required (though not sufficient) step towards fixing https://github.com/NixOS/nix/issues/4835 .

This changes the semantics of a `Realisation` to make it a closure: A realisation now depends on all the inputs of the derivation whose outputs end-up in the runtime closure.

This is important for #4835 because without this we don’t have a way to know whether substituting a realisation will lead to a duplicated dependency or not.

Depends on #4833  and #4834

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -59,6 +59,8 @@ struct LocalStore::State::Stmts {
     SQLiteStmt QueryAllRealisedOutputs;
     SQLiteStmt QueryPathFromHashPart;
     SQLiteStmt QueryValidPaths;
+    SQLiteStmt QueryRealisationRealisationReferences;
+    SQLiteStmt AddRealisationRealisationReference;
 };
 
 int getSchema(Path schemaPath)
@@ -316,7 +318,7 @@ LocalStore::LocalStore(const Params & params)
             )");
         state->stmts->QueryRealisedOutput.create(state->db,
             R"(
-                select Output.path, Realisations.signatures from Realisations
+                select Realisations.id, Output.path, Realisations.signatures from Realisations
                     inner join ValidPaths as Output on Output.id = Realisations.outputPath
                     where drvPath = ? and outputName = ?
                     ;
@@ -327,6 +329,19 @@ LocalStore::LocalStore(const Params & params)
                     inner join ValidPaths as Output on Output.id = Realisations.outputPath
                     where drvPath = ?
                     ;
+            )");
+        state->stmts->QueryRealisationRealisationReferences.create(state->db,
+            R"(
+                select drvPath, outputName from Realisations
+                    join RealisationsRefs on realisationReference = Realisations.id
+                    where referrer = ?;
+            )");
+        state->stmts->AddRealisationRealisationReference.create(state->db,
+            R"(
+                insert or replace into RealisationsRefs (referrer, realisationReference)
+                values (
+                    ?,
+                    (select id from Realisations where drvPath = ? and outputName = ?));
             )");
     }
 }
@@ -666,13 +681,17 @@ void LocalStore::registerDrvOutput(const Realisation & info)
     settings.requireExperimentalFeature("ca-derivations");
     auto state(_state.lock());
     retrySQLite<void>([&]() {
-        state->stmts->RegisterRealisedOutput.use()
-            (info.id.strHash())
-            (info.id.outputName)
-            (printStorePath(info.outPath))
-            (concatStringsSep(" ", info.signatures))
+        state->stmts->RegisterRealisedOutput
+            .use()(info.id.strHash())(info.id.outputName)(printStorePath(
+                info.outPath))(concatStringsSep(" ", info.signatures))
             .exec();
     });
+    uint64_t myId = state->db.getLastInsertedRowId();
+    for (auto& outputId : info.drvOutputDeps) {
+        state->stmts->AddRealisationRealisationReference
+            .use()(myId)(outputId.strHash())(outputId.outputName)
+            .exec();
+    }
 }
 
 void LocalStore::cacheDrvOutputMapping(State & state, const uint64_t deriver, const string & outputName, const StorePath & output)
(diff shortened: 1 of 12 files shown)


regnat:
Mh, I forgot to add a migration for the SQL db

regnat:
> Mh, I forgot to add a migration for the SQL db

Fixed

src/libstore/realisation.hh
```diff
@@ -28,6 +28,14 @@ struct Realisation {
 
     StringSet signatures;
 
+    /**
+     * The realisations that are required for the current one to be valid.
+     *
+     * When importing this realisation, the store will first check that all its
+     * dependencies exist, and map to the correct output path
+     */
+    std::map<DrvOutput, StorePath> dependentRealisations;
```
Ericson2314, line 37: I'm not sure I am comfortable referring to just the store path and not the entire realization. The whole point of this is coherence, and so we don't want just *any* realization connecting those two, but a coherent one. Similarly, the signatures ought to form some sort of Merkel dag.

Ericson2314: If we do that, I would change `Realization` to not have the `DrvOutput`, and call the original `RealizationWithKey` or similar. This avoids denormalization here and in `DrvOutputs`. (Actually, I think this should just reuse `DrvOutputs`.)

Ericson2314: In further thought, it should just be a regular content addressed (excluding signatures and other non-normative metadata) Merkle DAG.

We would store content addresses for each realization in the db. The map values can be (keyed or unkeyed, doesn't matter) realization *hashes*, not store paths. That keeps fetching from being any deeper than it is today, but still gives us the nice Merkel DAG guarantees.

It's content addressing all the way down!

regnat: > I'm not sure I am comfortable referring to just the store path and not the entire realization. The whole point of this is coherence, and so we don't want just any realization connecting those two, but a coherent one.

Maybe I’m missing something, but for me “coherence” is just about having the same store paths everywhere. The rest of the realisations is either metadata that shouldn’t come into play here (signatures for example), or stuff that’s implied by the output path (the transitive coherence, because if two CA realisations have the same output path, then they are bound to be coherent)

Also, I didn’t make it contain the full realisation, because that would have implied that each realisation has to *physically* (for a rather broad notion of physically) contain its whole dependency tree, which would make things awfully slow and redundant.

> In further thought, it should just be a regular content addressed (excluding signatures and other non-normative metadata) Merkle DAG.

Mh, not sure I follow you here

src/libstore/misc.cc
```diff
@@ -254,5 +254,44 @@ StorePaths Store::topoSortPaths(const StorePathSet & paths)
         }});
 }
 
+std::map<DrvOutput, StorePath> drvOutputReferences(
+    const std::set<Realisation> inputRealisations,
+    const StorePathSet pathReferences)
```
edolstra, line 259:
```suggestion
    const std::set<Realisation> & inputRealisations,
    const StorePathSet & pathReferences)
```

src/libstore/realisation.cc
```diff
@@ -21,11 +22,52 @@ std::string DrvOutput::to_string() const {
     return strHash() + "!" + outputName;
 }
 
+std::set<Realisation> Realisation::closure(Store & store, std::set<Realisation> startOutputs)
```
edolstra, line 25:
```suggestion
std::set<Realisation> Realisation::closure(Store & store, const std::set<Realisation> & startOutputs)
```

src/libstore/realisation.cc
```diff
@@ -21,11 +22,52 @@ std::string DrvOutput::to_string() const {
...
+{
+    std::set<Realisation> res;
+    Realisation::closure(store, startOutputs, res);
+    return res;
+}
+
+void Realisation::closure(Store & store, std::set<Realisation> startOutputs, std::set<Realisation> & res)
```
edolstra, line 32: Idem

src/libstore/store-api.cc
```diff
@@ -780,20 +780,40 @@ std::map<StorePath, StorePath> copyPaths(ref<Store> srcStore, ref<Store> dstStor
...
+            pool, Realisation::closure(*srcStore, toplevelRealisations),
+            [&](const Realisation& current) -> std::set<Realisation> {
+                std::set<Realisation> children;
+                for (const auto& [drvOutput, _] : current.dependentRealisations) {
+                    auto currentChild = srcStore->queryRealisation(drvOutput);
+                    if (!currentChild)
+                        throw Error(
+                            "Incomplete realisation closure: '%s' is a "
+                            "dependency "
+                            "of '%s' but isn’t registered",
```
edolstra, line 807:
```suggestion
                            "dependency of '%s' but isn’t registered",
```

src/libstore/local-store.cc
```diff
@@ -59,6 +59,8 @@ struct LocalStore::State::Stmts {
     SQLiteStmt QueryAllRealisedOutputs;
     SQLiteStmt QueryPathFromHashPart;
     SQLiteStmt QueryValidPaths;
+    SQLiteStmt QueryRealisationRealisationReferences;
+    SQLiteStmt AddRealisationRealisationReference;
```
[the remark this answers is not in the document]
regnat: It is indeed. That was because there used to be a `RealisationPathReference`, but it’s actually not needed. Fixed in dcabb461242e5c38e0a2780dc5869b91c13241d6

src/libstore/realisation.cc
regnat:
```suggestion
void Realisation::closure(Store & store, const std::set<Realisation> & startOutputs, std::set<Realisation> & res)
```

src/libstore/realisation.hh
Ericson2314: For reference, we discused this and decided it should be fine.

src/libstore/local-store.cc
```diff
@@ -664,14 +711,18 @@ void LocalStore::registerDrvOutput(const Realisation & info, CheckSigsFlag check
...
     retrySQLite<void>([&]() {
-        state->stmts->RegisterRealisedOutput.use()
-            (info.id.strHash())
-            (info.id.outputName)
-            (printStorePath(info.outPath))
-            (concatStringsSep(" ", info.signatures))
+        auto state(_state.lock());
+        state->stmts->RegisterRealisedOutput
+            .use()(info.id.strHash())(info.id.outputName)(printStorePath(
+                info.outPath))(concatStringsSep(" ", info.signatures))
```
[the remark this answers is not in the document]
regnat: Yup’ indeed. Fixed in c93a141698c763f135c33304733e078251513161

regnat (commented):
@Ericson2314 @edolstra I’ve fixed the stylistic issues we discussed. PTAL

src/libstore/local-store.cc
```diff
@@ -664,14 +711,22 @@ void LocalStore::registerDrvOutput(const Realisation & info, CheckSigsFlag check
...
+        auto state(_state.lock());
         state->stmts->RegisterRealisedOutput.use()
-            (info.id.strHash())
-            (info.id.outputName)
-            (printStorePath(info.outPath))
-            (concatStringsSep(" ", info.signatures))
+                (info.id.strHash())
+                (info.id.outputName)
+                (printStorePath(info.outPath))
+                (concatStringsSep(" ", info.signatures))
```
Ericson2314, line 720: did you want these indented more?

Ericson2314 (approved):

src/libstore/local-store.cc
regnat: Nope

PR #4842, opened 2021-05-20
Make `nix-shell` support content-addressed derivations

### LINKED ISSUE #4821: `nix-shell` doesn’t work with CA derivations
**Describe the bug**

Running `nix-shell` on a content-addressed derivation (or a dependency thereof) won’t add anything to the environment

**Steps To Reproduce**

```console
$ nix-shell '<nixpkgs>' --arg config '{ contentAddressedByDefault = true; }' -A hello
# Possibly wait for the world to build
nix-shell$ gcc
gcc: command not found
```

**Expected behavior**

`gcc` (and all the build inputs of hello) should be in `PATH`

**Additional context**

This is likely due to the fact that `nix-shell` loads the inputs of the **unresolved** derivation, but these are just placeholders, so they obviously don’t contain anything

regnat:
Resolve the derivation before trying to load its environment −
essentially reproducing what the build loop does − so that we can
effectively access our dependencies (and not just their placeholders).

Fix #4821

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -387,6 +387,13 @@ static void main_nix_build(int argc, char * * argv)
 
         if (dryRun) return;
 
+        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+            auto resolvedDrv = drv.tryResolve(*store);
+            if (!resolvedDrv)
+                throw Error("unable to resolve the derivation '%s'. nix-shell can’t continue", drvInfo.queryDrvPath());
+            drv = *resolvedDrv;
+        }
+
         // Set the environment.
         auto env = getEnv();
 
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -389,8 +389,7 @@ static void main_nix_build(int argc, char * * argv)
 
         if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
             auto resolvedDrv = drv.tryResolve(*store);
-            if (!resolvedDrv)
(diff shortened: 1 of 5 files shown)


src/nix-build/nix-build.cc
```diff
@@ -387,6 +387,13 @@ static void main_nix_build(int argc, char * * argv)
 
         if (dryRun) return;
 
+        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+            auto resolvedDrv = drv.tryResolve(*store);
+            if (!resolvedDrv)
+                throw Error("unable to resolve the derivation '%s'. nix-shell can’t continue", drvInfo.queryDrvPath());
```
Ericson2314, line 393: Should we do
```suggestion
            assert(resolvedDrv);
```
?

I think it should always resolve because the previous step builds all the deps, right?

regnat: It should, yes. But given that this isn’t in any way performance sensitive and that `assert` makes for some pretty ugly errors, I’d rather keep the nice exception

Ericson2314: Fair. Maybe some sort of "internal error" and/or "this should be impossible because we just resolved" then?

Ericson2314 (approved):
Other than the error message looks goods to me!

src/nix-build/nix-build.cc
regnat: I eventually put an assert, just with a dirty hack to add some slightly more meaningful message to the error

(Makes me think that it would be nice to have a standard way of having assert with error messages)

PR #4849, opened 2021-05-25
Make the Nar hash non modulo

### LINKED ISSUE #4792: [CA-drvs] nix-store --export fails with "error: hash of path ... has changed"
**Describe the bug**

`nix-store --export /nix/store/q5dfzj5jm03asw3lzq6l0m2ivr0miixs-libunistring-0.9.10` fails on my machine with:
```
error: hash of path '/nix/store/q5dfzj5jm03asw3lzq6l0m2ivr0miixs-libunistring-0.9.10' has changed from 'sha256:1fl38plbsahjcnmpnrd6yqk8wzpw0wr6wxghyzm3wzg3ww571s0w' to 'sha256:1k4iikm5r2zsl5g3sf9wd02f4qf61qn022cnp82j4dc67b1zxbpa'!
```

**Steps To Reproduce**

unknown

**Expected behavior**

`nix-store --export` should succeed or at least suggest a possible solution.

nix-env (Nix) 2.4pre20210510_7f9759b

**Additional context**

My `nix.conf`:
```
max-jobs = auto
experimental-features = nix-command ca-derivations ca-references flakes
keep-outputs = true
keep-derivations = true
substituters = https://cache.ytrizja.de https://cache.nixos.org https://nix-community.cachix.org https://cache.ngi0.nixos.org/
trusted-public-keys = cache.ytrizja.de:iZzXw6+kwvY/6CZMufK+dul5i44nJeJ3rsYg2HNpLQc= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs= cache.ngi0.nixos.org-1:KqH5CBLNSyX184S9BKZJo1LxrxJ9ltnY2uAs5c/f1MA=
```

regnat:
It makes much more sense to have the Nar hash be a plain straight hash rather than a hash modulo

Fix #4792

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -2354,32 +2354,19 @@ void LocalDerivationGoal::registerOutputs()
             }
             auto got = caSink.finish().first;
             auto refs = rewriteRefs();
-            HashModuloSink narSink { htSHA256, oldHashPart };
-            dumpPath(actualPath, narSink);
-            auto narHashAndSize = narSink.finish();
-            ValidPathInfo newInfo0 {
-                worker.store.makeFixedOutputPath(
+
+            auto finalPath = worker.store.makeFixedOutputPath(
                     outputHash.method,
                     got,
                     outputPathName(drv->name, outputName),
                     refs.second,
-                    refs.first),
-                narHashAndSize.first,
-            };
-            newInfo0.narSize = narHashAndSize.second;
-            newInfo0.ca = FixedOutputHash {
-                .method = outputHash.method,
-                .hash = got,
-            };
-            newInfo0.references = refs.second;
-            if (refs.first)
-                newInfo0.references.insert(newInfo0.path);
-            if (scratchPath != newInfo0.path) {
+                    refs.first);
+            if (scratchPath != finalPath) {
                 // Also rewrite the output path
                 auto source = sinkToSource([&](Sink & nextSink) {
                     StringSink sink;
                     dumpPath(actualPath, sink);
-                    RewritingSink rsink2(oldHashPart, std::string(newInfo0.path.hashPart()), nextSink);
+                    RewritingSink rsink2(oldHashPart, std::string(finalPath.hashPart()), nextSink);
                     rsink2(*sink.s);
                     rsink2.flush();
                 });
@@ -2388,6 +2375,21 @@ void LocalDerivationGoal::registerOutputs()
                 deletePath(actualPath);
                 movePath(tmpPath, actualPath);
             }
+
+            HashResult narHashAndSize = hashPath(htSHA256, actualPath);
+            ValidPathInfo newInfo0 {
+                finalPath,
+                narHashAndSize.first,
+            };
+
+            newInfo0.narSize = narHashAndSize.second;
+            newInfo0.ca = FixedOutputHash {
+                .method = outputHash.method,
+                .hash = got,
+            };
+            newInfo0.references = refs.second;
+            if (refs.first)
+                newInfo0.references.insert(newInfo0.path);
 
             assert(newInfo0.ca);
             return newInfo0;
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1436,14 +1436,10 @@ bool LocalStore::verifyStore(bool checkContents, RepairFlag repair)
                 /* Check the content hash (optionally - slow). */
                 printMsg(lvlTalkative, "checking contents of '%s'", printStorePath(i));
 
-                std::unique_ptr<AbstractHashSink> hashSink;
-                if (!info->ca || !info->references.count(info->path))
-                    hashSink = std::make_unique<HashSink>(info->narHash.type);
-                else
-                    hashSink = std::make_unique<HashModuloSink>(info->narHash.type, std::string(info->path.hashPart()));
+                auto hashSink = HashSink(info->narHash.type);
 
-                dumpPath(Store::toRealPath(i), *hashSink);
-                auto current = hashSink->finish();
+                dumpPath(Store::toRealPath(i), hashSink);
+                auto current = hashSink.finish();
 
                 if (info->narHash != nullHash && info->narHash != current.first) {
                     printError("path '%s' was modified! expected hash '%s', got '%s'",
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1152,17 +1152,13 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
 
             /* While restoring the path from the NAR, compute the hash
                of the NAR. */
-            std::unique_ptr<AbstractHashSink> hashSink;
-            if (!info.ca.has_value() || !info.references.count(info.path))
-                hashSink = std::make_unique<HashSink>(htSHA256);
-            else
-                hashSink = std::make_unique<HashModuloSink>(htSHA256, std::string(info.path.hashPart()));
+            HashSink hashSink(htSHA256);
 
-            TeeSource wrapperSource { source, *hashSink };
+            TeeSource wrapperSource { source, hashSink };
 
             restorePath(realPath, wrapperSource);
 
-            auto hashResult = hashSink->finish();
+            auto hashResult = hashSink.finish();
 
             if (hashResult.first != info.narHash)
                 throw Error("hash mismatch importing path '%s';\n  specified: %s\n  got:       %s",
--- a/src/nix/verify.cc
+++ b/src/nix/verify.cc
@@ -97,15 +97,11 @@ struct CmdVerify : StorePathsCommand
 
                 if (!noContents) {
 
-                    std::unique_ptr<AbstractHashSink> hashSink;
-                    if (!info->ca)
-                        hashSink = std::make_unique<HashSink>(info->narHash.type);
-                    else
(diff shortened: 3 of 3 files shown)


regnat:
Note that this is a big breaking change for CA paths with self-references (because it changes their `narHash`).

But it seemed very weird to have the `narHash` be modulo, and now is the last occasion we have to change it

edolstra:
Hm, I don't understand. It has to be modulo to be able to deal with self-references right?

regnat:
> Hm, I don't understand. It has to be modulo to be able to deal with self-references right?

Maybe I misunderstand the exact respective semantics of the `ca` and `nar` hashes, but my understanding is that the nar hash is essentially acting at the “transport layer” (as “the hash of the blob that you’ll download from a binary cache). So there’s no reason for it to even know the underlying path is content-addressed or not (and self-referential or not).

edolstra:
Hm yes, you're right. Only the hash used to construct the store path needs to be modulo self-references.

Ericson2314:
This makes sense to me, but I think you are going to need to change way more code:

 - other store impls
 - `addToStoreFromDump`
 - `addTextToStore`?

regnat:
> This makes sense to me, but I think you are going to need to change way more code:

Looking a bit closer at the code, I indeed missed `LocalStore::verifyStore` and `CmdVerify::run`. But all the others seem to assume a non-modulo hash (which is what prompted me to this change: The “modulo” part wasn’t done consistently).

There’s actually surprisingly little occurences of `HashModuloSink` in the codebase, I think we must be missing a few when checking the CA field

Ma27:
@regnat @Ericson2314 @edolstra I'm afraid that this has introduced a regression causing some hash-mismatches when using existing tooling such as bundix:

* Tested with Nix `nix-2.4pre20210612_79674c6`/`nix-2.4pre20210707_02dd6bb` (both versions client & daemon side).
* I only have `experimental-features = nix-command flakes` set in my `/etc/nix/nix.conf`.
* Used [`bundix`](https://github.com/nix-community/bundix) at version 2.5.0 (which uses Nix 2.3 for `nix-prefetch-url`).

When I generate a `gemset.nix` (in my case for [`vagrant`](https://github.com/hashicorp/vagrant/) at 2.2.17), I get - apart from a few other gems - the following declaration:

```
  childprocess = {
    groups = ["default"];
    platforms = [];
    source = {
      remotes = ["https://rubygems.org"];
      sha256 = "1lvcp8bsd35g57f7wz4jigcw2sryzzwrpcgjwwf3chmjrjcww5in";
      type = "gem";
    };
    version = "4.1.0";
  };
```

When building the corresponding derivation with Nix 2.3 it works perfectly fine, however I get the following error on the Nix versions I mentioned above:

```
error: ca hash mismatch importing path '/nix/store/hvwszfxaaj4r07yki6plhyfvccs5khpf-childprocess-4.1.0.gem';
         specified: sha256:1lvcp8bsd35g57f7wz4jigcw2sryzzwrpcgjwwf3chmjrjcww5in
         got:       sha256:1rg8yrf0dph30cd14z4kdir9v1k9yaizy83ai7byq55pq27x0hb8
error: builder for '/nix/store/flvaxlird6ki2hlmw5rgf7s3c29wmran-ruby2.7.3-childprocess-4.1.0.drv' failed with exit code 1
error: 1 dependencies of derivation '/nix/store/5k3xfpzm1j82693kg6r0f9kgbd8ghx5h-vagrant-2.2.17.drv' failed to build
error: 1 dependencies of derivation '/nix/store/hmg8shjlpd67v82xx1s1rgbyxzqcp7f2-vagrant-2.2.17.drv' failed to build
```

When reverting this PR, the issue is gone.

I currently have two notes:

* I don't really understand how big the impact of said "breaking change" was supposed to be, can you elaborate?
* To me, this seems like a major problem though: Nix 2.4 now breaks with hashes of existing tooling (and AFAIU I don't have any experimental CA stuff enabled). One of the killer-features of Nix is that you can instantiate & realize expressions that are several years old. When having potential hash mismatches for hashes from older Nix versions this doesn't seem to be the case anymore.

Please tell me if I'm missing something, but currently I'd strongly prefer to revert this change temporarily (which is also helpful so that we can update `nixUnstable` in `nixpkgs` to incorporate the fixes for the GC issues there).

Ma27:
OK while the problem is still there, I'm starting to doubt my initial assumption that this PR is the culprit:

* When I build things locally all the time, I don't seem have any issues.
* When I try to build things remotely (using `ssh-ng`, tested with `nix-2.4pre20210707_02dd6bb` with and without this PR reverted) I always get the hash mismatch.

But still, any ideas?

regnat:
I’d be surprised if this specific PR was the root of the problem as it’s only supposed to change the behavior of CA derivations with references, which aren’t supposed to happen without the `ca-references` or `ca-derivations` experimental feature (and your derivation doesn’t look like something that should contain a reference)

> When I try to build things remotely (using ssh-ng, tested with nix-2.4pre20210707_02dd6bb with and without this PR reverted) I always get the hash mismatch.

Is the remote using the same Nix version? Maybe there’s been a mismatch in the way CA hashes are computed that somehow only manifests here :thinking:
Also, does this only happen for this specific derivation?

If you have an easy repro, I can have a look at it

Ma27:
@regnat, I guess I managed to isolate the problem:

* Obtain a Nix 2.3 via `nix-shell -p nix` and run inside the shell `nix-prefetch-url https://rubygems.org/gems/childprocess-4.1.0.gem` (which should give you the hash `1lvcp8bsd35g57f7wz4jigcw2sryzzwrpcgjwwf3chmjrjcww5in`)
* Install `nix (Nix) 2.4pre20210601_5985b8b` on both your local machine and a remote builder and enable the experimental features `nix-command flakes`. (the remote builder in question has the configuration `ssh-ng://builder-ng x86_64-linux - 10 4 big-parallel,kvm,nixos-test`).
* Check out `nixpkgs` at `44188d64727f056e1596b89019cda5b92eb0b368`
* Add the following Nix expression (let's call it `test.nix`):
```nix
with import <nixpkgs> { };

let
  childprocess ={
    platforms = [];
    source = {
      remotes = ["https://rubygems.org"];
      sha256 = "1lvcp8bsd35g57f7wz4jigcw2sryzzwrpcgjwwf3chmjrjcww5in";
      type = "gem";
    };
    version = "4.1.0";
  };
in bundlerEnv {
  name = "tmp";
  version = "1.2.4";
  inherit ruby;
  gemfile = writeText "Gemfile" "";
  lockfile = writeText "Gemfile.lock" "";
  gemset = { inherit childprocess; };
}
 ```
* Run `nix build -f test.nix -L  -j0` (to force a remote build). You should now see the following error:
```
error: ca hash mismatch importing path '/nix/store/hvwszfxaaj4r07yki6plhyfvccs5khpf-childprocess-4.1.0.gem';
         specified: sha256:1lvcp8bsd35g57f7wz4jigcw2sryzzwrpcgjwwf3chmjrjcww5in
         got:       sha256:1rg8yrf0dph30cd14z4kdir9v1k9yaizy83ai7byq55pq27x0hb8
error: builder for '/nix/store/kcwb7qsca8i0nbwjmbqic888kz542p95-ruby2.7.3-childprocess-4.1.0.drv' failed with exit code 1
error: 1 dependencies of derivation '/nix/store/8rkpjmzv9v5zhjysaxkim3qm5qvf1lkh-tmp.drv' failed to build
```
* Run `nix build -f test.nix -L  --option builders ''` and everything should work fine.

**Comment by regnat** [comment]
@Ma27 I couldn’t reproduce it. Here’s what I tried:

```console
$ rm -rf /tmp/store* # Just to be sure
$ nix --version
nix (Nix) 2.4pre20210707_02dd6bb
$ export NIX_REMOTE=/tmp/store                                     
$ nix-prefetch-url https://rubygems.org/gems/childprocess-4.1.0.gem
path is '/nix/store/hvwszfxaaj4r07yki6plhyfvccs5khpf-childprocess-4.1.0.gem'
1lvcp8bsd35g57f7wz4jigcw2sryzzwrpcgjwwf3chmjrjcww5in
$ nix -L -j0 --option builders 'ssh-ng://localhost?remote-store=/tmp/store2 x86_64-linux - 10 4 big-parallel,kvm,nixos-test' --option experimental-features 'nix-command flakes' build -f test.nix
```

With `test.nix` being

```nix
let
 nixpkgs = builtins.getFlake "nixpkgs/44188d64727f056e1596b89019cda5b92eb0b368";
in

with nixpkgs.legacyPackages.${builtins.currentSystem};

let
  childprocess ={
    platforms = [];
    source = {
      remotes = ["https://rubygems.org"];
      sha256 = "1lvcp8bsd35g57f7wz4jigcw2sryzzwrpcgjwwf3chmjrjcww5in";
      type = "gem";
    };
    version = "4.1.0";
  };
in bundlerEnv {
  name = "tmp";
  version = "1.2.4";
  inherit ruby;
  gemfile = writeText "Gemfile" "";
  lockfile = writeText "Gemfile.lock" "";
  gemset = { inherit childprocess; };
}
```

I tried different combinations (building it locally first or not, pre-fetching the source or not), but none of them caused the error (I had a few very strange plugins-related warnings, but that seems totally unrelated)

Ma27:
@regnat well, first of all thanks a lot for taking a look! I guess I'll try to isolate the entire issue into a VM expression to rule out any kind of differences between our setups I might've missed :)

PR #4853, opened 2021-05-27
Throw freenode down the memory hole

Synthetica9:
More accurately reflect the current situation, with Freenode's implosion and all


adisbladis:
I'm not sure that a channel on Libera should be mentioned as it's not official?

PR #4863, opened 2021-06-01
Fix pure build of 2.3

jonringer:
`utillinuxMinimal` is now an alias on 21.05.

### DIFF
--- a/release-common.nix
+++ b/release-common.nix
@@ -57,7 +57,7 @@ rec {
       git
       mercurial
     ]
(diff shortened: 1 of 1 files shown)


Kloenk:
Should this also update the nixpkgs tarball in release.nix? Like I did here https://github.com/NixOS/nix/pull/4867/files#diff-d5f7a586debedc022115cf3432285b544ed70a8d9f2c99634b4cb376ecb939d2R2

PR #4876, opened 2021-06-02
Add a pointer from "realising" to `nix log`.

jfly:
While working my way through the nix-pills, I was surprised to not see
the stdout of a build when realising a derivation. I eventually
discovered `nix log` and https://github.com/NixOS/nix-pills/pull/146,
but the *first* place I looked was this man page expecting to see a
note about stdout/stderr getting suppressed during builds.
Unfortunately, I didn't find anything. I think this addition to the man
page would help future learners like myself.

### DIFF
--- a/doc/manual/src/command-ref/nix-store.md
+++ b/doc/manual/src/command-ref/nix-store.md
@@ -156,13 +156,11 @@ To test whether a previously-built derivation is deterministic:
 $ nix-build '<nixpkgs>' -A hello --check -K
 ```
 
-To see the stderr/stdout of a build:
+Use [`--read-log`](#operation---read-log) to show the stderr and stdout of a build:
 
 ```console
 $ nix-store --read-log $(nix-instantiate ./test.nix)
 ```
-
-Use `nix log` to see the logs of a build.
 
 # Operation `--serve`
 
--- a/doc/manual/src/command-ref/nix-store.md
+++ b/doc/manual/src/command-ref/nix-store.md
@@ -156,6 +156,14 @@ To test whether a previously-built derivation is deterministic:
 $ nix-build '<nixpkgs>' -A hello --check -K
 ```
 
+To see the stderr/stdout of a build:
+
+```console
+$ nix-store --read-log $(nix-instantiate ./test.nix)
+```
+
+Use `nix log` to see the logs of a build.
+
 # Operation `--serve`
 
 ## Synopsis


edolstra:
A few comments:

* `nix log` is an experimental command so it shouldn't be used here. You could use `nix-store --read-log` instead.
* The example should be more concrete, since users may not know what the argument to `nix log` or `nix-store --read-log` should be.
* It's not really obvious whether users will see this line in the manual. (There are *many* places where we could conceivably refer users to `nix-store --read-log`...)

jfly:
@edolstra I've updated the man page with a concrete example of how to use `nix-store --read-log`.

I totally understand if you want to reject this PR. I can't speak for other people, but I can say that one of the first bits of debugging I did was:

1. `man nix-store`.
2. Search for `--realise`.
3. Search for stdout, hoping to find a note about what happens to stdout/stderr when doing a `nix-store --realise`.

So this man change would have helped me in this specific case. That all said, I probably only ran into this confusion because of how the pill 7.2 is written, so it might be more helpful to get https://github.com/NixOS/nix-pills/pull/146 merged up =)

jfly:
@edolstra ping on this?

jfly:
I think this is still useful, I don't think this should be stale.

doc/manual/src/command-ref/nix-store.md
```diff
@@ -156,6 +156,14 @@ To test whether a previously-built derivation is deterministic:
 $ nix-build '<nixpkgs>' -A hello --check -K
 ```
 
+To see the stderr/stdout of a build:
```
fricklerhandwerk, line 159:
```suggestion
Use [`--read-log`](#operation---read-log) to show the stderr and stdout of a build:
```

doc/manual/src/command-ref/nix-store.md
```diff
@@ -156,6 +156,14 @@ To test whether a previously-built derivation is deterministic:
...
+
+```console
+$ nix-store --read-log $(nix-instantiate ./test.nix)
+```
+
+Use `nix log` to see the logs of a build.
+
```
fricklerhandwerk, line 166:
````suggestion
Alternatively, use [`nix log`](./new-cli/nix3-nix-log.md) (experimental) to show the build logs:

```console
nix log $(nix-instantiate ./test.nix)
```
````

To be honest, I find it a bit hard to integrate it smoothly here, so I'd also be fine with removing the line completely: 

```suggestion
```

fricklerhandwerk (approved):
We should move this forward. @jfly please use links to commands and options such that their documentation is easier to find, I added suggestions.

jfly:
Thanks, @fricklerhandwerk! I've updated the PR.

fricklerhandwerk:
@jfly Thanks for following up and keeping at it!

PR #4878, opened 2021-06-02
flake check keep going

regnat:
- Check the CA hash when importing stuff in the local store
- Let `nix flake check` keep going when keep-going is set


src/nix/flake.cc
```diff
@@ -272,25 +272,40 @@ struct CmdFlakeCheck : FlakeCommand
         auto state = getEvalState();
         auto flake = lockFlake();
 
+        bool hasErrors = false;
+        auto throw_ = [&](const Error & e) {
```
edolstra, line 276: Maybe call this `reportError`, "throw" suggests that it won't return.

src/nix/flake-check.md
```diff
@@ -22,6 +22,8 @@ This command verifies that the flake specified by flake reference
 that the derivations specified by the flake's `checks` output can be
 built successfully.
 
+If the `keep-going` option is set to `true`, Nix will keep evaluating as much as it can and report the errors as it encounters them. Otherise it will stop at the first error.
```
edolstra, line 25:
```suggestion
If the `keep-going` option is set to `true`, Nix will keep evaluating as much as it can and report the errors as it encounters them. Otherwise it will stop at the first error.
```

Also could you break lines at ~80 characters?

edolstra (approved):

src/nix/flake.cc
regnat: Oh yes, you’re right. Fixed in https://github.com/NixOS/nix/pull/4878/commits/7d651f5c3fc721b8957a2266e5e14e99bdb77077

PR #4885, opened 2021-06-03
Improve explanation of NIX_PATH prefix syntax

kini:
The previous wording seemed to imply that the "channel:" syntax would resolve to a github archive URL, which is not the case.


kini:
Sorry, didn't realize that #3611 already updated the URL. I'd still like to change the wording so that it's clear what exactly the `channel:<channel-name>` abbreviated syntax expands to. I'll push a different change in a minute.

kini:
OK, updated. Not sure how to test the change; I tried running `make man` but the resulting `.1` files in `doc/manual/` don't seem to contain any of this text.

edolstra:
`env-common` is no longer included in the manpages, I think this got lost in the conversion to Markdown.

Issue #4893, opened 2021-06-05
Nix segfault at evalutation. GC_DONT_GC=1 doesn't help


rissson:
**Describe the bug**

Nix segfaults when when evaluating.

**Steps To Reproduce**

1. Run `nix flake check git+https://gitlab.com/lama-corp/infra/dotshabka\?rev=c69be3f98dfa8b943f83bf2f9a8cb17ab403ae0b`
2. `[1]    1690602 segmentation fault (core dumped)`

**Expected behavior**

Nix doesn't segfault.

**`nix-env --version` output**

`nix-env (Nix) 2.4pre20210601_5985b8b`
Also happens on `nix-env (Nix) 2.4pre20210503_6d2553a`
Also happens on nix's `8e6ee1b9e924fbbbeb5594eb89e7a570f36ab6e1` (latest commit on master at the time of writing)

**Additional context**

Adding `GC_DONT_GC=1` to the environment doesn't help.

Here are the latest lines output when ran with `-vvv`:
```
evaluating flake...
using substituted/cached input 'github:SoxinOS/soxin/587ae3306c4c1ce130f217b06c07ea8b3b43d400' in '/nix/store/sr31h4gm1gp8clk3y7rqsk6gry3c7i0a-source'
checking access to '/nix/store/sr31h4gm1gp8clk3y7rqsk6gry3c7i0a-source/flake.nix'
evaluating file '/nix/store/sr31h4gm1gp8clk3y7rqsk6gry3c7i0a-source/flake.nix'
using substituted/cached input 'github:NixOS/nixpkgs/4714dcf148dfb528f4107ceb5dc86529ca71294a' in '/nix/store/ghknj4cnjsjajx88zji9wlfr9rrz4x8y-source'
checking access to '/nix/store/ghknj4cnjsjajx88zji9wlfr9rrz4x8y-source/flake.nix'
evaluating file '/nix/store/ghknj4cnjsjajx88zji9wlfr9rrz4x8y-source/flake.nix'
checking access to '/nix/store/ghknj4cnjsjajx88zji9wlfr9rrz4x8y-source/lib'
evaluating file '/nix/store/ghknj4cnjsjajx88zji9wlfr9rrz4x8y-source/lib/default.nix'
checking access to '/nix/store/ghknj4cnjsjajx88zji9wlfr9rrz4x8y-source/lib/default.nix'
[...]

PR #4895, opened 2021-06-08
avoid null pointer dereference when attrSet->pos is null

manveru:
This may be a fix for https://github.com/NixOS/nix/issues/4893

I added a small test that just checks for the segfault, given that `builtins.tryEval` cannot catch this kind of error.

To recap, here's the behavior before:

```
❯ nix eval --expr 'builtins.getAttr "bar" (builtins.listToAttrs [ { name = "foo"; value = "bar"; } ])'
[1]    1731608 segmentation fault (core dumped)  nix eval --expr 
```

And after the fix:

```
❯ ./result/bin/nix eval --expr 'builtins.getAttr "bar" (builtins.listToAttrs [ { name = "foo"; value = "bar"; } ])'
error: attribute 'bar' missing for call to 'getAttr'

       at «string»:1:1:

            1| builtins.getAttr "bar" (builtins.listToAttrs [ { name = "foo"; value = "bar"; } ])
             | ^
```

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -564,22 +564,21 @@ static Bindings::iterator getAttr(
             funcName
         );
 
-        Pos aPos = *attrSet->pos;
-        if (aPos == noPos) {
-            throw TypeError({
-                .msg = errorMsg,
-                .errPos = pos,
-            });
-        } else {
+        if (attrSet->pos && *attrSet->pos != noPos) {
             auto e = TypeError({
                 .msg = errorMsg,
-                .errPos = aPos,
+                .errPos = *attrSet->pos,
             });
 
             // Adding another trace for the function name to make it clear
             // which call received wrong arguments.
             e.addTrace(pos, hintfmt("while invoking '%s'", funcName));
             throw e;
+        } else {
+            throw TypeError({
+                .msg = errorMsg,
+                .errPos = pos,
+            });
         }
     }
 
--- a/tests/lang.sh
+++ b/tests/lang.sh
@@ -32,7 +32,7 @@ done
 for i in lang/eval-fail-*.nix; do
     echo "evaluating $i (should fail)";
     i=$(basename $i .nix)
-    if nix-instantiate --eval lang/$i.nix; then
+    if ! expect 1 nix-instantiate --eval lang/$i.nix; then
         echo "FAIL: $i shouldn't evaluate"
         fail=1
     fi
--- a/tests/lang/eval-fail-list-to-attrs.nix
+++ b/tests/lang/eval-fail-list-to-attrs.nix
@@ -1,0 +1,4 @@
+# https://github.com/NixOS/nix/issues/4893
+builtins.getAttr "nope" (
+  builtins.listToAttrs [ { name = "foo"; value = "bar"; } ]
+)


src/libexpr/primops.cc
```diff
@@ -564,22 +564,29 @@ static Bindings::iterator getAttr(
             funcName
         );
 
-        Pos aPos = *attrSet->pos;
-        if (aPos == noPos) {
+        if (attrSet->pos == nullptr) {
```
regnat, line 567: This might be factored out a slight bit I guess. Something along the lines of:

```cpp
if (attrSet->pos && *attrSet->pos != noPos) {
  // Throw with the right location and the stack trace
} else {
  // Throw with the parent location
}
```

tests/list-to-attrs-nullptr.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
regnat, line 1: Rather than adding a brand new toplevel test (which does cost a bit as initializing a test is slightly expensive), I think you can

- Add this to `tests/lang/eval-fail-list-to-attrs.nix`
- Change https://github.com/NixOS/nix/blob/master/tests/lang.sh#L35 to `if ! expect 1 nix-instantiate --eval lang/$i.nix; then` (I checked, all the other tests do return `1` as they should)

Also: Can you add a reference to the issue in the test file, so that we know why it’s here?

regnat (changes requested):
Thanks for the fix!

I have a couple of comments to clean this a slight bit, but overall 👍🏻

src/libexpr/primops.cc
manveru: Does this look OK to you now?

regnat (approved):
Looks good now!

manveru:
Thanks for all the guidance :)

roberth (approved):

roberth:
I've run `nixUnstable` with this patch for some days now and it caused a significant improvement in stability.

cc @edolstra

domenkozar:
I plan to merge this in a few days unless someone says raises a concern.

edolstra:
@domenkozar I want to have a look at it first.

roberth:
The majority of other sites where `pos` is dereferenced do not check for null either.
So while this PR does make `pos` handling a bit more robust, it doesn't seem to be a complete fix.
Part of the problem is that the codebase seems to have two representations of "no position": null or `&noPos`. The latter is more robust; more than `std::optional<Pos>` even, because it doesn't ever need to throw or crash. It would seem best to remove all `null` initializations of `pos` with an `assert(pos)` in the appropriate `Attr` constructor. However, I didn't spot any such initializations in a quick sample. Perhaps the problem is that some `Bindings` aren't initialized correctly?

Meanwhile, it doesn't hurt to merge this.

regnat:
> Part of the problem is that the codebase seems to have two representations of "no position": null or `&noPos`. The latter is more robust; more than `std::optional<Pos>` even, because it doesn't ever need to throw or crash.

Couldn’t we make `pos` a `ref` (assuming that `ref` has no runtime overhead)? That way we can statically enforce that it can’t be null and we sidestep this issue

> It would seem best to remove all null initializations of pos with an assert(pos) in the appropriate Attr constructor. However, I didn't spot any such initializations in a quick sample.

IIRC in the current case, we explicitly pass a `pos` argument to the `Attr` constructor, but this argument happens to be null

Ma27 (approved):
This fixes a fairly weird problem I had when forgetting to pass `version` (along with `pname`) to `buildRustPackage` in an overlay evaluated by a flake.

roberth:
@edolstra This change is an improvement. Could you merge it?
It may not be the final fix, but the improved stability in the meanwhile will be greatly appreciated.

roberth:
There's #5192 now, which builds on the ideas from this PR and is more thorough.

manveru:
Awesome, that looks way better!
I'll close this PR then.

PR #4899, opened 2021-06-09
[2.3] backport: Make `functionArgs` primitive accept primops

sternenseemann:
(cherry picked from commit b2748c6e99239ff6803ba0da76c362790c8be192)

This fixes user observable evaluation errors which shouldn't happen when using `lib.generators.toPretty`, so it seems sensible to fix this in 2.3 as well.


sternenseemann:
Thanks!

PR #4904, opened 2021-06-10
flake.nix: Propagate nlohmann_json

roberth:
Without it, the nix headers are unusable.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -102,7 +102,6 @@
 
         propagatedDeps =
           [ (boehmgc.override { enableLargeConfig = true; })
-            nlohmann_json
           ];
 
         perlDeps =
@@ -408,7 +407,7 @@
 
             doInstallCheck = true;
 
-            lcovFilter = [ "*/boost/*" "*-tab.*" ];
+            lcovFilter = [ "*/boost/*" "*-tab.*" "*/nlohmann/*" ];
 
             # We call `dot', and even though we just use it to
             # syntax-check generated dot files, it still requires some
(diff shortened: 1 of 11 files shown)


flake.nix
```diff
@@ -103,6 +102,7 @@
 
         propagatedDeps =
           [ (boehmgc.override { enableLargeConfig = true; })
+            nlohmann_json
```
regnat, line 105: Wouldn’t that also propagate it to the non-dev outputs? (Not that it’s a big dependency ofc, but still)

edolstra:
Hm, but we're not using external nlohmann_json anymore (9f28dd97ae6afc68f0574a251325336c12d60c6f). I guess the problem is that we should install the headers in $dev/include/nix.

roberth:
@edolstra Yes we are using external nlohmann_json. That commit didn't do much, but I think I've completed the removal of the external one in my second commit here.

src/libexpr/flake/lockfile.hh
```diff
@@ -2,7 +2,7 @@
 
 #include "flakeref.hh"
 
-#include <nlohmann/json_fwd.hpp>
+#include <nlohmann/json.hpp>
```
edolstra, line 5: No no, we should not include json.hpp in header files! IIRC it adds a huge amount of compilation time to every compilation unit.

roberth:
Ok, the code base is in limbo between using an external nlohmann_json and a vendored nlohmann_json.
Clearly I don't know how it should be done, so feel free to take my commits and finish it. It shouldn't take much effort for someone who knows how it's best integrated.

PR #4906, opened 2021-06-11
Make `computeFSClosure` ca-aware

### LINKED ISSUE #4820: `nix-store --query --roots` breaks with CA derivations and `keep-derivations = true`
**Describe the bug**

Running `nix-store --query --roots` on a path whose reverse closure contains unbuilt CA derivations, when the `keep-derivations` option is set to true fails with

```
error: output 'foo' of derivation '/nix/store/123-abc.drv' has no store path mapped to it
```

**Steps To Reproduce**

*I haven’t built a nice self-contained example yet. Probably will have to for testing anyways*

**Expected behavior**

Should work

**Additional context**

At a glance this seems to be caused by the call to `queryDerivationOutputs` [here](https://github.com/NixOS/nix/blob/master/src/libstore/misc.cc#L56). If that’s true then replacing it by `queryPartialDerivationOutputMap` should do the trick

regnat:
Fix #4820 by preventing nix-collect garbage from crashing if
`keep-outputs` or `keep-derivations` is true

### DIFF
--- a/src/libstore/misc.cc
+++ b/src/libstore/misc.cc
@@ -30,7 +30,7 @@ void Store::computeFSClosure(const StorePathSet & startPaths,
 
             if (includeDerivers && path.isDerivation())
                 for (auto& [_, maybeOutPath] : queryPartialDerivationOutputMap(path))
-                    if (maybeOutPath && isValidPath(*maybeOutPath) && queryPathInfo(*maybeOutPath)->deriver == path)
+                    if (maybeOutPath && isValidPath(*maybeOutPath))
                         res.insert(*maybeOutPath);
             return res;
         };
--- a/src/libstore/misc.cc
+++ b/src/libstore/misc.cc
@@ -29,9 +29,9 @@ void Store::computeFSClosure(const StorePathSet & startPaths,
                     res.insert(i);
 
             if (includeDerivers && path.isDerivation())
-                for (auto& i : queryDerivationOutputs(path))
-                    if (isValidPath(i) && queryPathInfo(i)->deriver == path)
-                        res.insert(i);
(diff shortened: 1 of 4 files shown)


src/libstore/misc.cc
```diff
@@ -29,9 +29,9 @@ void Store::computeFSClosure(const StorePathSet & startPaths,
                     res.insert(i);
 
             if (includeDerivers && path.isDerivation())
-                for (auto& i : queryDerivationOutputs(path))
-                    if (isValidPath(i) && queryPathInfo(i)->deriver == path)
-                        res.insert(i);
+                for (auto& [_, maybeOutPath] : queryPartialDerivationOutputMap(path))
+                    if (maybeOutPath && isValidPath(*maybeOutPath) && queryPathInfo(*maybeOutPath)->deriver == path)
```
Ericson2314, line 33: I don't think `deriver == path` will work if multiple fixed-output derivations build the same thing?

src/libstore/misc.cc
```diff
@@ -29,9 +29,9 @@ void Store::computeFSClosure(const StorePathSet & startPaths,
                     res.insert(i);
```
Ericson2314, line 29: Does `queryValidDerivers` on the line above take CA derivations into account?

Ericson2314 (approved):
I think this is strictly better than before, but there might be more to do.

regnat:
> I think this is strictly better than before, but there might be more to do.

Indeed there was, thanks for the review.

The branch now runs the `gc` test with CA derivations, so hopefully everything is properly caught now 🤞🏻 (at least everything that used to be tested still works 😉)

PR #4908, opened 2021-06-11
Make `nix develop` work with CA derivations

### LINKED ISSUE #4823: `nix develop` crashes with CA derivations
**Describe the bug**

Running `nix develop` on a CA derivation fails with `error: std::get: wrong index for variant`

**Steps To Reproduce**

```console
# If someone knows a less convoluted way of passing a custom nixpkgs argument with flakes, please let me know 😉
$ nix develop --file $(nix eval --raw nixpkgs#path) --arg config '{ contentAddressedByDefault = true; }' hello
error: std::get: wrong index for variant
```

**Expected behavior**

Like always, should work™

**Additional context**

I guess the error comes from [here](https://github.com/NixOS/nix/blob/master/src/nix/develop.cc#L152). I’m not sure what the shortest way to fix is − short of inlining more of the `derivation` builtin in `nix develop`, but that doesn’t look awesome

regnat:
Fix #4823 

Depends on #4842 , but just to avoid a conflict as they because they both modify the same test. Both PRs are otherwise unrelated

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -144,17 +144,26 @@ StorePath getDerivationEnvironment(ref<Store> store, const StorePath & drvPath)
     /* Rehash and write the derivation. FIXME: would be nice to use
        'buildDerivation', but that's privileged. */
     drv.name += "-env";
-    for (auto & output : drv.outputs) {
-        output.second = { .output = DerivationOutputInputAddressed { .path = StorePath::dummy } };
-        drv.env[output.first] = "";
-    }
     drv.inputSrcs.insert(std::move(getEnvShPath));
-    Hash h = std::get<0>(hashDerivationModulo(*store, drv, true));
+    if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+        for (auto & output : drv.outputs) {
+            output.second = {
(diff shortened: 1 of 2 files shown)


src/nix/develop.cc
```diff
@@ -144,26 +144,34 @@ StorePath getDerivationEnvironment(ref<Store> store, const StorePath & drvPath)
...
-        drv.env[output.first] = "";
-    }
     drv.inputSrcs.insert(std::move(getEnvShPath));
-    Hash h = std::get<0>(hashDerivationModulo(*store, drv, true));
+    if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+        for (auto & output : drv.outputs) {
+            output.second = {
+                .output = DerivationOutputDeferred{},
+            };
+            drv.env[output.first] = "";
```
Ericson2314, line 153: Per https://github.com/NixOS/nix/blob/e6150de90d8db101209fc6363f5f7696ee8192c4/src/libexpr/primops.cc#L1146 I think it should be 
```suggestion
            drv.env[output.first] = hashPlaceholder(output.first);
```

Ericson2314 (approved):

Issue #749, opened 2015-12-24
Semantically meaningful ways to communicate our intent


copumpkin:
I realize we can use `builtins.trace` to implement warnings, but that doesn't tell Nix that we want them to be warnings. We can also use a wrapper function around `assert`/`throw` to emit a custom error message when an assertion fails, but Nix still won't know that we're attaching the message to the failed assertion.

This means that if Nix ever evolved a slightly more user-friendly front-end, it would be hard for us to collect all "warnings" during evaluation (i.e., deprecated aliases, other potential gotchas) and print them out in an orderly fashion (e.g., in Hydra), and all the assorted uses of `assert`/`throw`.

I'm thinking it would be nice to add a couple of simple (for now) builtins that provide semantically meaningful feedback of that sort. Something like `builtins.warn`, which would behave like trace at first, but could eventually be made to collect warnings (and potentially allow us to suppress certain types of them) in a structured fashion. `builtins.expect`/`builtins.assume` might allow us to give slightly more color (along with string messages) to assertions we make, while also giving us potential structure going forward.

cc @garbas, eternal advocate for user-friendliness, and @edolstra for obvious reasons

copumpkin:
Doubly cool might be if  `builtins.warn` and friends accepted an attrset and `toJSON`'d it before running, pulling out some sort of canonical human-readable string from it by default. This would allow us to come up with structured messages (deprecations might have a field telling you what to use instead, etc.)

vcunat:
> Something like builtins.warn, which would behave like trace at first, but could eventually be made to collect warnings

That sounds like it should be implement(ed/able) purely in nixpkgs in the start (`lib.warn`).

A quick thought: IMO great flexibility could be easily provided if we allowed `throw`-ing any type and allowed catching as well. Then we could use attrmaps to encode any kinds of errors/warnings and also handle them within nix-written code...

copumpkin:
#1000

Profpatsch:
I think it’s a very important feature to have.
https://github.com/NixOS/nixpkgs/pull/19315 also depends on it.

nmattia:
Hi, I would find this very useful. Is there a plan, and is there anything we can do to help?

garbas:
@nmattia As far as I'm aware there is no plan. You can use helpers in `nixpkgs`: `lib.warn` and `lib.info`

Profpatsch:
I’d still like to see this. Using `trace` for warnings is not possible due to how hydra works.

[...]

PR #4914, opened 2021-06-15
Add builtins.debug

gytis-ivaskevicius:
When working on tests I thought that it would be useful to have some sort of logger which would be enabled by increasing verbosity level. This PR adds `builtins.debug` which evaluates the first argument when `-vv` is specified.

Exact use case:
I and @blaggacao started working on these check utils https://github.com/numtide/flake-utils/blob/master/check-utils.nix and plan is to add this kind of logging to simplify the debugging process for the end-users

PS: Please be gentle, I don't do CPP :D

### DIFF
--- a/src/libexpr/eval.hh
+++ b/src/libexpr/eval.hh
@@ -646,6 +646,9 @@ struct EvalSettings : Config
 
     Setting<bool> useEvalCache{this, true, "eval-cache",
         "Whether to use the flake evaluation cache."};
+
+    Setting<bool> traceVerbose{this, false, "trace-verbose",
+        "Whether `builtins.traceVerbose` should trace its first argument when evaluated."};
 };
 
 extern EvalSettings evalSettings;
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -970,6 +970,15 @@ static RegisterPrimOp primop_trace({
 });
 
 
+/* Takes two arguments and evaluates to the second one. Used as the
+ * builtins.traceVerbose implementation when --trace-verbose is not enabled
+ */
+static void prim_second(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    state.forceValue(*args[1], pos);
+    v = *args[1];
+}
+
 /*************************************************************
  * Derivations
  *************************************************************/
@@ -3925,6 +3934,18 @@ void EvalState::createBaseEnv()
         addPrimOp("__importNative", 2, prim_importNative);
         addPrimOp("__exec", 1, prim_exec);
     }
+
+    addPrimOp({
+        .fun = evalSettings.traceVerbose ? prim_trace : prim_second,
+        .arity = 2,
+        .name = symbols.create("__traceVerbose"),
+        .args = { "e1", "e2" },
+        .doc = R"(
+          Evaluate *e1* and print its abstract syntax representation on standard
+          error if `--trace-verbose` is enabled. Then return *e2*. This function
+          is useful for debugging.
+        )",
+    });
 
     /* Add a value containing the current Nix expression search path. */
     mkList(v, searchPath.size());
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -973,7 +973,7 @@ static RegisterPrimOp primop_trace({
 /* Takes two arguments and evaluates to the second one. Used as the
  * builtins.traceVerbose implementation when --trace-verbose is not enabled
  */
-static void prim_second(EvalState & state, const Pos & pos, Value * * args, Value & v)
+static void prim_second(EvalState & state, const PosIdx pos, Value * * args, Value & v)
 {
     state.forceValue(*args[1], pos);
     v = *args[1];
@@ -3938,7 +3938,7 @@ void EvalState::createBaseEnv()
     addPrimOp({
         .fun = evalSettings.traceVerbose ? prim_trace : prim_second,
         .arity = 2,
-        .name = symbols.create("__traceVerbose"),
+        .name = "__traceVerbose",
         .args = { "e1", "e2" },
         .doc = R"(
           Evaluate *e1* and print its abstract syntax representation on standard
--- a/tests/lang.sh
+++ b/tests/lang.sh
@@ -5,6 +5,8 @@ export NIX_REMOTE=dummy://
 
 nix-instantiate --eval -E 'builtins.trace "Hello" 123' 2>&1 | grep -q Hello
 nix-instantiate --eval -E 'builtins.addErrorContext "Hello" 123' 2>&1
+nix-instantiate --trace-verbose --eval -E 'builtins.traceVerbose "Hello" 123' 2>&1 | grep -q Hello
+(! nix-instantiate --eval -E 'builtins.traceVerbose "Hello" 123' 2>&1 | grep -q Hello)
 (! nix-instantiate --show-trace --eval -E 'builtins.addErrorContext "Hello" 123' 2>&1 | grep -q Hello)
 nix-instantiate --show-trace --eval -E 'builtins.addErrorContext "Hello" (throw "Foo")' 2>&1 | grep -q Hello
 


blaggacao (approved):
LGTM (no CPP)

src/libexpr/primops.cc
```diff
@@ -878,6 +878,33 @@ static RegisterPrimOp primop_trace({
...
+            printError("debug: %1%", args[0]->string.s);
+        else
+            printError("debug: %1%", *args[0]);
+    }
+    state.forceValue(*args[1], pos);
+    v = *args[1];
+}
+
+static RegisterPrimOp primop_debug({
+    .name = "__debug",
```
Mic92, line 897: How is this different from `builtins.trace`?

gytis-ivaskevicius: It works only if `-vv` is passed. Basically, Nix lang level debug logging.

gytis-ivaskevicius:
I'd appreciate a review. 
\CC: @domenkozar @edolstra @regnat

src/libexpr/primops.cc
Mic92: I see.

src/libexpr/primops.cc
```diff
@@ -878,6 +878,33 @@ static RegisterPrimOp primop_trace({
...
+    .args = {"e1", "e2"},
+    .doc = R"(
+      Evaluate *e1* and print its abstract syntax representation on standard
+      error if talkative verbosity enabled. Then return *e2*. This function
```
edolstra, line 901:
```suggestion
      error if talkative verbosity is enabled. Then return *e2*. This function
```

edolstra:
LGTM. Only problem is that the name `debug` is a bit generic...

Could you add a test?

gytis-ivaskevicius:
Does anyone have any naming suggestions?

HaoZeke:
`pesticide` --> `bugnixer` ?

Infinisil (changes requested):
I think @L-as's suggestion of `builtins.traceVerbose` is great. In addition to the below suggestions, I'm not entirely sure whether using `-v` for this is a good idea, though I don't have any good reasons against it.

Here's a branch where I implemented the below two suggestions (and also having a separate flag instead of `-v`, but not sure if we want that): https://github.com/Infinisil/nix/commit/3cfde4389156bbfa58270ff72d2189b8009fec89

src/libexpr/primops.cc
```diff
@@ -878,6 +878,33 @@ static RegisterPrimOp primop_trace({
 });
 
 
+/* Evaluate the first expression and print it on standard error if talkative verbosity
+ * is enabled. Then return the second expression.  Useful for debugging. */
+static void prim_debug(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    state.forceValue(*args[0], pos);
```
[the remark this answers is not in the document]
Infinisil: Notably this causes Nix evaluation to slightly differ if verbose mode is enabled, but I think it's not a problem because the result here doesn't affect the evaluation result, other than if the expression fails to evaluate. I think this is a worthwhile tradeoff for having an essentially no-cost `builtins.traceVerbose` we can spam in the codebase for debugging purposes.

balsoft:
I like the idea, as long as you can't "lift" the verbosity level to the Nix language itself.

gytis-ivaskevicius:
Hey guys, finally got around to it.
Can someone advise me on how to test this? `nix-instantiate` seems to be ignoring verbosity flags :/

gytis-ivaskevicius:
Finally got around to this. I think its ready for review

src/libexpr/primops.cc
```diff
@@ -892,6 +892,15 @@ static RegisterPrimOp primop_trace({
 });
 
 
+/* Takes two arguments and evaluates to the second one. Used as the
+ * builtins.traceVerbose implementation when --trace-verbose is not enabled
+ */
+static void prim_second(EvalState & state, const Pos & pos, Value * * args, Value & v)
```
regnat, line 898: I find it a bit annoying that this doesn’t evaluate its first argument, as it means that the same Nix expression can run correctly without `--trace-verbose` but fail with it (`builtins.traceVerbose (rec { x = x; }.x) "foo"` for example).

Could we instead have something that does a `coerceToString` on its first argument (and just discards it) to be sure that both evaluation modes have the exact same semantics?

(that would obviously mean that `traceVerbose` is never free, so there’s a tradeoff to make here)

gytis-ivaskevicius: From one point of view - yes you are right. But in general, if something is unused - it should not be evaluated but I can definitely see it being misleading.

I am down for both solutions, would anyone else like to share their thoughts on this?

Infinisil: I already commented on this in https://github.com/NixOS/nix/pull/4914#discussion_r759708467

edolstra: You could use `prim_seq` here, which does evaluate its first argument. But I think being lazy is better. Otherwise you have to worry about the runtime cost of a debug message (e.g. `builtins.traceVerbose "${big expensive expression}" ...`) even when they're disabled.

src/libexpr/eval.hh
```diff
@@ -477,6 +477,9 @@ struct EvalSettings : Config
 
     Setting<bool> useEvalCache{this, true, "eval-cache",
         "Whether to use the flake evaluation cache."};
+
+    Setting<bool> traceVerbose{this, false, "trace-verbose",
+        "Whether `builtins.traceVerbose` should trace its first argument when evaluated."};
```
Infinisil, line 482: While I did add this in my branch, I'm not sure if we want this. On one hand, having `-v` also have an influence on evaluation is a bit weird, but on the other hand it probably won't be a problem.

src/libexpr/primops.cc
gytis-ivaskevicius: I guess we are sticking with it being lazy. Resolving this conversation

src/libexpr/eval.hh
gytis-ivaskevicius: Yeah, I think it's fine. Does not really matter.

Infinisil:
> Yeah, I think it's fine. Does not really matter.

I think this _does_ matter, why shouldn't it? This isn't an implementation detail, this is the user interface. It's the difference of requiring the user to learn an extra flag, or being able to reuse `-v`. On one hand, I don't see any problems to reuse `-v`, but on the other hand it seems cleaner to have this separate `--trace-verbose` flag, especially because the `-v` flag currently doesn't involve evaluation at all.

So I guess I'm leaning towards having the new `--trace-verbose`, but I'd like someone else to think about this.

gytis-ivaskevicius:
Alright, I would prefer `-v`. Definitely not a deal-breaker for me. Does anyone else have any thoughts on this?

roberth:
This also reminds me of https://github.com/NixOS/nixpkgs/pull/140763, which makes `lib.warn` abort. It'd be great to have this behavior as a `builtins.warn`, which then also works in pure mode.

(A `--show-trace-on-warn` would also be great but is just not feasible without having traces as first class objects. Stackless evaluator? Would be cool but not a priority.)

Infinisil:
@roberth `--show-trace-on-warn` could be feasible if there was a `builtins.warn`

roberth:
> @roberth `--show-trace-on-warn` could be feasible if there was a `builtins.warn`

Do we have first-class traces yet? Afaik, they're hidden behind the implicit C++ exception stack. Otherwise, realistically, we can only do `--error-on-warn`.

Infinisil:
This is related to https://github.com/NixOS/nix/issues/749, ping @Profpatsch

Infinisil:
I think it's related because of different verbosities of messages:
- Error (maybe `throw`, though that stops execution)
- Warning (The thing discussed in https://github.com/NixOS/nix/issues/749)
- Info (`builtins.trace`)
- Debug (`builtins.debug`, this PR)

edolstra:
@Profpatsch What misdesign are you referring to?

thufschmitt:
@gytis-ivaskevicius I’ve merged the branch with master, but now the build is failing. Can you have a look?

thufschmitt:
We discussed that a bit during the last [UX meeting](https://hackmd.io/U21zZ1bwQiW-EnORBaNU1Q?both#builtinsdebug). There's a broad consensus on this being great, just needs to fix the build.

@gytis-ivaskevicius think you can take a bit of time to do that?

gytis-ivaskevicius:
Alright, will do! (Expect a push tomorrow/day after tomorrow)

gytis-ivaskevicius:
Sorry, a little late. Took me 2 attempts to resolve the issue :sweat_smile:

thufschmitt (approved):
Thanks, let's roll this!

thufschmitt:
> Thanks, let's roll this!

Once the CI agrees, that is

thufschmitt:
Duh, forgot about the release notes. I'll add them in another PR

Issue #3121, opened 2019-10-07
Copy local flakes to the store lazily


edolstra:
Currently flakes are evaluated from the Nix store, so when using a local flake, it's first copied to the store. This means that
```
$ cd /path/to/nixpkgs
$ nix build .#hello
```
is a lot slower than the non-flake alternative
```
$ nix build -f . hello
```

Ideally, we would copy the flake to the store only when its `outPath` attribute is evaluated. However, we also need to ensure that it's not possible to access untracked files (i.e. we need to check every file against `git ls-files`).

hmenke:
Still important.

L-as:
It would be nice if Nix could take advantage of the filesystem's native CoW functionality (if present) in order to speed up copying.
We discussed this briefly in #offtopic:nixos.org.

lilyball:
I just hit this. My first attempt at a workaround was to remove the `self` arg from `outputs` (as without `self` I can't access the source tree at all). It turns out that makes it copy the tree and then throw an error about how the `outputs` function doesn't take a `self` arg.

Lazily copying the flake only when `outPath` is evaluated would be ideal, but being able to just drop the `self` arg to suppress the copy would be a great first step.

@L-as Taking advantage of CoW would be nice but it's not doable on macOS where the Nix store lives on a separate volume (separate volume group even).

lilyball:
Also for context, in my case the flake was not in a git repo, it was just in a folder. Copying to the nix store is unacceptable because the folder contains multiple git repos along with all their build artifacts. Copying a git repo to the Nix store at least would avoid copying untracked files, but in my case it had hundreds of thousands of files and multiple gigabytes of data to look at and copy.

Atemu:
@lilyball @L-as Taking advantage of CoW doesn't work on Linux either due to a VFS limitation: https://github.com/NixOS/nix/issues/5513

One thing I'd like to understand in this issue is why a local flake can't be evaluated "directly" just like the old `default.nix`-style file evaluation.  
I know copying has benefits for hermetic evaluation and such but I don't need that, like, at all.  
Sure, remote flakes should be copied to the Nix store and that's really great functionality but I see no point whatsoever in doing the same for local flakes that are already in the FS and not expected to change without the user's knowledge.

TLATER:
@Atemu from what I've read, it's to help enforce hermetic evaluation and avoid impurities. Presumably it also has advantages for code simplicity, because you don't need to write something separate for local flakes.

[...]

PR #4922, opened 2021-06-17
libfetchers/git: fetch submodules by default

### LINKED ISSUE #4423: nix flakes: add support for git submodules

nrdxp:
This is the desired behavior according to @edolstra's comment: https://github.com/NixOS/nix/issues/4423#issuecomment-799270569

In addition to making submodules opt-out instead of opt-in, this allows submodules to be pulled for a top-level flake's `self` attribute automatically, which was previously impossible.

#### Note:
This doesn't work with the "github:owner/repo" syntax, but neither do submodules in general. I can open a separate ticket for that if desired

Resolves #4423

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -44,7 +44,7 @@ void emitTreeAttrs(
 
     if (input.getType() == "git")
         mkBool(*state.allocAttr(v, state.symbols.create("submodules")),
-            fetchers::maybeGetBoolAttr(input.attrs, "submodules").value_or(false));
+            fetchers::maybeGetBoolAttr(input.attrs, "submodules").value_or(true));
 
     if (auto revCount = input.getRevCount())
         mkInt(*state.allocAttr(v, state.symbols.create("revCount")), *revCount);
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -172,7 +172,7 @@ struct GitInputScheme : InputScheme
         Input input(_input);
 
         bool shallow = maybeGetBoolAttr(input.attrs, "shallow").value_or(false);
-        bool submodules = maybeGetBoolAttr(input.attrs, "submodules").value_or(false);
+        bool submodules = maybeGetBoolAttr(input.attrs, "submodules").value_or(true);
         bool allRefs = maybeGetBoolAttr(input.attrs, "allRefs").value_or(false);
 
         std::string cacheType = "git";
--- a/tests/fetchGitSubmodules.sh
+++ b/tests/fetchGitSubmodules.sh
@@ -42,8 +42,8 @@ r1=$(nix eval --raw --expr "(builtins.fetchGit { url = file://$rootRepo; rev = \
 r2=$(nix eval --raw --expr "(builtins.fetchGit { url = file://$rootRepo; rev = \"$rev\"; submodules = false; }).outPath")
 r3=$(nix eval --raw --expr "(builtins.fetchGit { url = file://$rootRepo; rev = \"$rev\"; submodules = true; }).outPath")
 
-[[ $r1 == $r2 ]]
-[[ $r2 != $r3 ]]
+[[ $r1 == $r3 ]]
+[[ $r2 != $r1 ]]
 
 r4=$(nix eval --raw --expr "(builtins.fetchGit { url = file://$rootRepo; ref = \"master\"; rev = \"$rev\"; }).outPath")
 r5=$(nix eval --raw --expr "(builtins.fetchGit { url = file://$rootRepo; ref = \"master\"; rev = \"$rev\"; submodules = false; }).outPath")
@@ -52,13 +52,13 @@ r7=$(nix eval --raw --expr "(builtins.fetchGit { url = $rootRepo; ref = \"master
 r8=$(nix eval --raw --expr "(builtins.fetchGit { url = $rootRepo; rev = \"$rev\"; submodules = true; }).outPath")
 
 [[ $r1 == $r4 ]]
-[[ $r4 == $r5 ]]
+[[ $r4 == $r6 ]]
 [[ $r3 == $r6 ]]
 [[ $r6 == $r7 ]]
 [[ $r7 == $r8 ]]
 
 have_submodules=$(nix eval --expr "(builtins.fetchGit { url = $rootRepo; rev = \"$rev\"; }).submodules")
-[[ $have_submodules == false ]]
+[[ $have_submodules == true ]]
 
 have_submodules=$(nix eval --expr "(builtins.fetchGit { url = $rootRepo; rev = \"$rev\"; submodules = false; }).submodules")
 [[ $have_submodules == false ]]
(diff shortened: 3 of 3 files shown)


nrdxp:
can someone rerun the test suite? The failure appears transient and didn't fail on a local run of the test in a mac vm. It's also not related or caused by my changes.

Kha:
I agree this is the correct default, but there should still be a way to opt out for `self`. Consider, for example, writing a `flake.nix` for a project that includes LLVM as a submodule. If the Nix build uses Nixpkgs LLVM anyway, copying the submodule to the store on every change would be a great waste of work and disk space.

nrdxp:
I agree, and thought of that as well. I'll take a look at it later and see if I can figure out a simple way to do just that, but setting attributes on `self` might just be more suitably handled in a separate PR. Thoughts?

jonringer:
> I agree this is the correct default, but there should still be a way to opt out for `self`. Consider, for example, writing a `flake.nix` for a project that includes LLVM as a submodule. If the Nix build uses Nixpkgs LLVM anyway, copying the submodule to the store on every change would be a great waste of work and disk space.

That sounds more like git submodule abuse.  I know that submodules are an easy solution to hard problems.... maybe nix has just changed me, idk.

Kha:
Obviously with Nix no-one would do that, but outside of our bubble that seems like a somewhat popular approach
https://github.com/rust-lang/rust/tree/master/src
https://github.com/ghc/ghc
(counterpoint: https://github.com/ziglang/zig-bootstrap/issues/17)

nrdxp:
Anyone know where I could start on implementing @Kha's suggestion? It'd be great to set attributes on self, but after grepping around a while, I'm not really sure where in the codebase I would start from?

Kha:
It looks like the `inputs` attributes are parsed here: https://github.com/NixOS/nix/blob/223e0569ff2c93d8a907188b40d1928d6967e91f/src/libexpr/flake/flake.cc#L93-L94. You'll probably need to special-case `inputName == "self"` (or `sSelf`?) there since most other input attributes don't make sense on it, and then check for the value of the new attribute wherever `self` is actually instantiated.

SuperSandro2000:
We should definitely disable this by default in nixpkgs as it will break hashes and we don't want to fetch submodules by default anyway.

andir:
> We should definitely disable this by default in nixpkgs as it will break hashes and we don't want to fetch submodules by default anyway.

This is not the same git fetcher that is being used in nixpkgs. The change here is likely only affecting the builtin fetcher.

kreisys:
> > We should definitely disable this by default in nixpkgs as it will break hashes and we don't want to fetch submodules by default anyway.
> 
> This is not the same git fetcher that is being used in nixpkgs. The change here is likely only affecting the builtin fetcher.

☝️that's true. go ahead and do a `git grep fetchGit` (with a capital G) in a `nixpkgs` tree.
Can anyone counter this point? If not then PLEASE, I implore you, let's move forward with this.This solves a very real problem that I'm once again finding myself having to contort around. If I had one loonie (that is one Canadian dollar 😌) for every hour lost on git submodules in the past year I could've bought y'all a dozen rounds of very fancy drinks

shlevy:
@edolstra Any objection? If not will merge next week.

DieracDelta:
Is there a way to disable submodules being pulled in for `self`?

nrdxp:
@DieracDelta, if you skim the history of this very PR you will see that idea was requested already, but I didn't have time or the knowledge of this codebase to yet implement. I've talked with a few colleagues who have more experience with the Nix codebase, but they are similarly time constrained atm unfortunately.

 I'd be happy to have such a feature though. Conceptually, I see no reason to limit `self`; we should be able to set many of the same options we use on other flakes. I felt that the request was a bit out of scope for this PR, and so hopefully another PR will emerge when someone can get around to it to allow for setting relevant options on `self`.

rraval:
Obligatory ["this broke my workflow"](https://xkcd.com/1172/).

We develop software in a large monorepo with submodules that point to extremely large third party projects. We use a `flake.nix` to fully specify a `devShell` environment with all developer dependencies specified. We then use [`nix-direnv`](https://github.com/nix-community/nix-direnv) to conveniently expose the `devShell` as a facade you can simply `cd` into.

To be clear, we're using nix purely for a developer environment that can operate on a folder. That folder happens to contain source code as a git clone, but nix is not used to actually build any part of the project.

As a result of this change, `nix print-dev-env path:/some/repo` is now trying to clone all submodules for no particularly good reason (in this usecase! I acknowledge that cloning submodules seems reasonable when nix is actually building something from `self`).

I have tried the following to no avail:
- `nix print-dev-env git+file:///some/repo?submodules=false`
- `nix print-dev-env git+file:///some/repo?submodules=0`

For now, I'm looking to revert my own install of the `nix` binary to something right before this was merged. However, that's a stopgap and real support for avoiding submodules would be appreciated.

jonringer:
We should probably make it so that people can opt-in or opt-out of using submodules. something like `{ inputs.self.enableSubmodules = true; ... }`

DieracDelta:
@nrdxp   I am not super experienced with c++ (or the nix codebase for that matter), but I'm happy to attempt this once we are certain of the behavior. Was proposed behavior already worked out?

@jonringer I quite like your proposal here. The follow up question I have is: what should the default behavior be? My vote would be for:

- no submodules are cloned by default
- submodules added to `inputs.self.enabledSubmodules` are cloned
- submodules added to `inputs.self.disabledSubmodules` are *not* cloned
- if a submodule is in both `inputs.self.enabledSubmodules` and `inputs.self.disabledSubmodules`, the flake will refuse to evaluate.
- if a submodule is listed in either `inputs.self.enabledSubmodules` or `inputs.self.disabledSubmodules` but does not exist, the flake will refuse to evaluate.
- if a submodule is not listed in either `inputs.self.enabledSubmodules` or `inputs.self.disabledSubmodules`, the flake will evaluate but not clone the submodule. This preserves backwards compatability.

edolstra:
The case where Nix is used just for dev-shells would be addressed by https://github.com/NixOS/nix/issues/3121 since then the top-level flake wouldn't have to be copied to the Nix store.

nrdxp:
@jonringer, if you would agree, instead of having special attributes on self, why not just allow setting the same options as on any other git flake input type?

jonringer:
> @jonringer, if you would agree, instead of having special attributes on self, why not just allow setting the same options as on any other git flake input type?

if you're referring to just having a `inputs.self.submodules = <bool>;`, then I'm okay with it, is it should be at least a solution for people who really don't or really do want submodules. Individual submodules could be considered a "nice to have", and will likely be a significantly more complex feature to implement.

PR #4927, opened 2021-06-21
include additional supported compression formats into binary-cache-store

tomberek:
Append extensions for additional libarchive supported formats. This specifically came up when looking at utilizing gzip compression and expecting to see `.gz` or `.gzip` in the binary cache.

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -183,6 +183,7 @@ ref<const ValidPathInfo> BinaryCacheStore::addToStoreCommon(
            compression == "lzip" ? ".lzip" :
            compression == "lz4" ? ".lz4" :
            compression == "br" ? ".br" :
+           compression == "gzip" ? ".gz" :
            "");
 
     auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(now2 - now1).count();
--- a/tests/gz.sh
+++ b/tests/gz.sh
@@ -1,0 +1,28 @@
+source common.sh
+
+clearStore
+clearCache
+
+cacheURI="file://$cacheDir?compression=gzip"
+
+outPath=$(nix-build dependencies.nix --no-out-link)
+
+nix copy --to $cacheURI $outPath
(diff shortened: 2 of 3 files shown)


src/libstore/binary-cache-store.cc
```diff
@@ -183,6 +183,9 @@ ref<const ValidPathInfo> BinaryCacheStore::addToStoreCommon(
            compression == "lzip" ? ".lzip" :
            compression == "lz4" ? ".lz4" :
            compression == "br" ? ".br" :
+           compression == "gzip" ? ".gz" :
+           compression == "lzma" ? ".lzma" :
+           compression == "compress" ? ".compress" :
```
edolstra, line 188: What format is "compress"?

tomberek: I don't have a need for it, but did found that it works. Seems to produce archives that can be manipulated with: https://linux.die.net/man/1/uncompress

I came across the possibility of that option here: https://github.com/NixOS/nix/blob/53857552278173641c0489a3196c51aab769bac8/src/libutil/compression.cc#L265

tomberek: Maybe that one should be “.Z” to be standard.

edolstra:
I'm not sure if there is a use case for these formats. In particular, it seems to me that brotli and zstd can do anything that .Z and .gz do but better/faster. And enabling them does increase the attack surface a bit, and adds complexity for anybody who wants to write a compatible implementation of the binary cache client.

tomberek:
> I'm not sure if there is a use case for these formats. In particular, it seems to me that brotli and zstd can do anything that .Z and .gz do but better/faster. And enabling them does increase the attack surface a bit, and adds complexity for anybody who wants to write a compatible implementation of the binary cache client.

That’s fair. My interest was is .gz due to easy integration with other technology (zsync for differential downloads of similar NARs without needing a server), the rest was just about noticing it worked.

tomberek:

PR #4939, opened 2021-06-24
Make CA derivations compatible with recursive Nix

### LINKED ISSUE #4353: Make CA derivations work with recursive Nix
Currently running `nix` with `--experimental-features` ca-derivations in recursive-nix will most probably error out because the `queryRealisation` and `registerDrvOutput` methods are forbidden. We should implement a proper access-control to allow it when it's safe to do so

regnat:
Add an access-control list to the realisations in recursive-nix (similar
to the already existing one for store paths), so that we can build
content-addressed derivations in the restricted store.

Fix #4353

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1333,13 +1333,18 @@ struct RestrictedStore : public virtual RestrictedStoreConfig, public virtual Lo
     std::optional<const Realisation> queryRealisation(const DrvOutput & id) override
     // XXX: This should probably be allowed if the realisation corresponds to
     // an allowed derivation
-    { throw Error("queryRealisation"); }
+    {
+        if (!goal.isAllowed(id))
+            throw InvalidPath("cannot query an unknown output id '%s' in recursive Nix", id.to_string());
+        return next->queryRealisation(id);
+    }
 
     void buildPaths(const std::vector<DerivedPath> & paths, BuildMode buildMode) override
     {
         if (buildMode != bmNormal) throw Error("unsupported build mode");
 
         StorePathSet newPaths;
+        std::set<Realisation> newRealisations;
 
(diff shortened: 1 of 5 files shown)


src/libstore/build/local-derivation-goal.cc
```diff
@@ -1333,13 +1333,18 @@ struct RestrictedStore : public virtual RestrictedStoreConfig, public virtual Lo
     std::optional<const Realisation> queryRealisation(const DrvOutput & id) override
     // XXX: This should probably be allowed if the realisation corresponds to
     // an allowed derivation
-    { throw Error("queryRealisation"); }
+    {
+        if (!goal.isAllowed(id))
+            throw InvalidPath("cannot query an unknown output id '%s' in recursive Nix", id.to_string());
+        return next->queryRealisation(id);
```
edolstra, line 1339: Maybe this should be restricted further, to check whether `id` has been built *by this RestrictedStore* (i.e. whether we've seen a `buildPaths`  call to realise this `id`). Otherwise, you could use this to observe whether `id` has previously been built, which would be impure.

regnat: Maybe the implementation isn’t correct, but my intent is that `isAllowed(DrvOutput & id)` returns true iff
- `id` is a dependency of the current derivation, or
- `id` has been built by the current `RestrictedStore`

(which is the same thing as what’s currently done for `StorePath`s).

Is that wrong?

regnat: Oh wait, the first condition doesn’t seem to make any sense. I’ll remove that bit

regnat: OK, I’ve rewritten things so that we don’t track the inputs of the derivation, and `isAllowed(DrvOutput & id)` only returns true if `id` has been built by the current store

PR #4944, opened 2021-06-24
Fix gc crash

roberth:
I've figured out another cause of gc crashes, related to coroutines.

This now solves
 - a crash that occurs whenever a gc-registered thread is in a coroutine (the boehmgc patch)
 - an inappropriate "stack overflow" when gc runs while a coroutine exists (the BoehmGCStackAllocator commit)

I'll run `nixUnstable` with this PR and #4895 to see if any other crashes occur.

### DIFF
--- a/boehmgc-coroutine-sp-fallback.diff
+++ b/boehmgc-coroutine-sp-fallback.diff
@@ -1,5 +1,5 @@
 diff --git a/pthread_stop_world.c b/pthread_stop_world.c
-index 1cee6a0b..8977b0dd 100644
+index 1cee6a0b..46c3acd9 100644
 --- a/pthread_stop_world.c
 +++ b/pthread_stop_world.c
 @@ -674,6 +674,8 @@ GC_INNER void GC_push_all_stacks(void)
@@ -11,7 +11,7 @@ index 1cee6a0b..8977b0dd 100644
  
      if (!EXPECT(GC_thr_initialized, TRUE))
        GC_thr_init();
-@@ -723,6 +725,30 @@ GC_INNER void GC_push_all_stacks(void)
+@@ -723,6 +725,28 @@ GC_INNER void GC_push_all_stacks(void)
            hi = p->altstack + p->altstack_size;
            /* FIXME: Need to scan the normal stack too, but how ? */
            /* FIXME: Assume stack grows down */
@@ -35,9 +35,7 @@ index 1cee6a0b..8977b0dd 100644
 +              lo = hi - stack_limit;
 +            }
 +          #else
-+            if (lo < hi || lo >= hi + stack_limit) { // sp outside stack
-+              lo = hi + stack_limit;
-+            }
++          #error "STACK_GROWS_UP not supported in boost_coroutine2 (as of june 2021), so we don't support it in Nix."
 +          #endif
          }
          GC_push_all_stack_sections(lo, hi, traced_stack_sect);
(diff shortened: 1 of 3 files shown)


jonringer:
should the boehm-gc patch be upstreamed? looks like it addresses some of the `FIXME` comments above

roberth:
> should the boehm-gc patch be upstreamed? looks like it addresses some of the `FIXME` comments above

The existing fixmes in boehm are for an unrelated feature on very rare architectures where the stack grows in the other direction. I did not implement those fixmes because we don't need it. Otoh, I couldn't resist writing the code to support these rare archs in my own part. I'll remove it, so the next person won't think it matters to Nix.

roberth:
> should the boehm-gc patch be upstreamed?

As it is, it is part of a very specific solution that works for Nix, which isn't ideal because it scans garbage stack sections sometimes. That's ok because it's a conservative garbage collector and in the case of Nix those garbage sections will not be scanned anymore when the coroutine has ended. This is the main reason I have doubts about upstreaming.
A more accurate implementation involves more changes as outlined in the inline comments in the patch. These affect both boehmgc and boost_coroutine2.
So the PR is a local optimum. With significant effort, it can be made accurate and upstream-worthy, but it improves Nix only marginally, so it's not something I would enjoy doing in my free time.

manveru:
Can confirm that this fixes a segfault we experience when building `github:input-output-hk/daedalus/4.1.0#daedalus` with GC enabled.

domenkozar:
> > should the boehm-gc patch be upstreamed?
> 
> As it is, it is part of a very specific solution that works for Nix, which isn't ideal because it scans garbage stack sections sometimes. That's ok because it's a conservative garbage collector and in the case of Nix those garbage sections will not be scanned anymore when the coroutine has ended. This is the main reason I have doubts about upstreaming.
> A more accurate implementation involves more changes as outlined in the inline comments in the patch. These affect both boehmgc and boost_coroutine2.
> So the PR is a local optimum. With significant effort, it can be made accurate and upstream-worthy, but it improves Nix only marginally, so it's not something I would enjoy doing in my free time.

That's fair. Could we open an issue and explain what we did, so they're aware of our use case?

edolstra:
Really great work, thanks!

Yeah it would be good to get upstream's ideas about this.

roberth:
While formulating my question I've found issues I hadn't seen before, with some recommendations. I'll summarize my findings here.
https://github.com/ivmai/bdwgc/issues/150 is an interaction with someone who seems to have found a similar solution. Since then, functions have been [added](https://github.com/ivmai/bdwgc/commit/5668de71107022a316ee967162bc16c10754b9ce) to manipulate the recorded stack base per thread. Maintaining these stack base pointers [requires a gc lock](https://github.com/ivmai/bdwgc/issues/274#issuecomment-492423287) around the coroutine context switch. While the current approach to scanning the coroutine stacks works, it'd be better to provide our own stack "pushing" routine as mentioned [here](https://github.com/ivmai/bdwgc/issues/274#issuecomment-492426364), which runs as part of gc and does not need to worry about synchronization as much. It also may help with potential mark stack overflows, although I don't think we've experienced those.
So most of this will be helpful when writing a more accurate solution.
For now our solution is similar to https://github.com/ivmai/bdwgc/issues/150 and we err on the side of overly conservative gc when it comes to stacks during coroutine execution.

lilyball:
The boehmgc patch only applies to Linux. Boehm GC has a separate "stop the world and push thread stacks" implementation for Darwin (see `darwin_stop_world.c`) and this still crashes when GC is invoked from a coroutine.

roberth:
> The boehmgc patch only applies to Linux.

This was an oversight. See https://github.com/NixOS/nix/issues/4919

> crashes when GC is invoked from a coroutine.

It's more frequent than that. The main cause is that an OS thread has its stack pointer in a coroutine, which is not supported by an unpatched GC. Which thread (or stack) invokes the GC doesn't really matter beyond that.

PR #4947, opened 2021-06-25
Allow to compile after `./configure --enable-gc=no`

Pamplemousse:
As of now, configuring with `--enable-gc=no`, and trying to compile triggers a
```
  CXX    src/libexpr/eval.o
src/libexpr/eval.cc:67:42: error: use of undeclared identifier 'traceable_allocator'
    return std::allocate_shared<Value *>(traceable_allocator<Value *>(), v);
```

This PR seems to fix it, but I have very low confidence in my C++.

cc @regnat 

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -64,7 +64,11 @@ static char * dupStringWithLen(const char * s, size_t size)
 
 RootValue allocRootValue(Value * v)
 {
+#if HAVE_BOEHMGC
     return std::allocate_shared<Value *>(traceable_allocator<Value *>(), v);
+#else
+    return std::make_shared<Value *>(v);
+#endif
 }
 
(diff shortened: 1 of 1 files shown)


regnat (approved):

src/libexpr/eval.cc
```diff
@@ -64,7 +64,11 @@ static char * dupStringWithLen(const char * s, size_t size)
 
 RootValue allocRootValue(Value * v)
 {
+#if HAVE_BOEHMGC
     return std::allocate_shared<Value *>(traceable_allocator<Value *>(), v);
+#else
+    return std::make_shared<Value *>(std::move(v));
```
edolstra, line 70: This std::move looks unnecessary since there is not much point in moving a pointer.

PR #4949, opened 2021-06-26
Doc macos install

toonn:
I went through the macOS installation docs recently and found out `apfs.util` no longer has a `-B` flag. I believe the `-t` flag accomplishes the same (more or less) behavior.

### DIFF
--- a/doc/manual/installation/installing-binary.xml
+++ b/doc/manual/installation/installing-binary.xml
@@ -378,7 +378,7 @@ sudo rm /Library/LaunchDaemons/org.nixos.nix-daemon.plist
           of new entries without a reboot:
         </para>
 
-        <screen>alice$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B</screen>
+        <screen>alice$ /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t</screen>
       </listitem>
 
       <listitem>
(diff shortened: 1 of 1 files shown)


edolstra:
Maybe it's better to do what master does, which is:
```
    # Big Sur takes away the -B flag we were using and replaces it
    # with a -t flag that appears to do the same thing (but they
    # don't behave exactly the same way in terms of return values).
    # This feels a little dirty, but as far as I can tell the
    # simplest way to get the right one is to just throw away stderr
    # and call both... :]
    {
        /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -t || true # Big Sur
        /System/Library/Filesystems/apfs.fs/Contents/Resources/apfs.util -B || true # Catalina
    } >/dev/null 2>&1
```

toonn:
For docs it doesn't feel quite right. I can mention "On older macOS you should use -B." or something?

Wouldn't that be enough? Let me know what you think.

toonn:
Btw, might `apfs.util -t || apfs.util -B` not be a bit more future-proof? They might introduce a new `-B` flag in a future version.

abathur:
> Btw, might `apfs.util -t || apfs.util -B` not be a bit more future-proof? They might introduce a new `-B` flag in a future version.

They might, and there is certainly a more principled way to handle it. IIRC the Big Sur version of this can exit non-zero even with the correct flags (though I don't recall under what circumstances/system-state I discovered this), while I _think_ the Catalina version may exit 0 as long as you use the correct flag (regardless of whether it actually did anything helpful for you).

I hoped to make it clear-ish in the code code that: 
- the first draft is a dumb brute-force approach that throws away exit codes without considering whether or not they contain any actionable information
- any smarter approach will need to wrestle with the differing exit semantics

toonn:
I'm closing this, since 2.4 has been released anyway. I still have some improvements to suggest but I'll create a fresh PR for those.

PR #4953, opened 2021-06-28
wip: fixes for cygwin

corngood:
I'm just working on building nix on cygwin.  In addition to these changes, I needed to build libeditline locally.

Boost coroutines are also currently broken on cygwin, so I've been using the changes described here:

https://cygwin.com/pipermail/cygwin-developers/2020-September/011970.html

to work around the problem.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -271,6 +271,16 @@ AC_ARG_ENABLE(doc-gen, AC_HELP_STRING([--disable-doc-gen],
   doc_generate=$enableval, doc_generate=yes)
 AC_SUBST(doc_generate)
 
+# Look for lowdown
+PKG_CHECK_MODULES([LOWDOWN], [lowdown], [
+  CXXFLAGS="$LOWDOWN_CFLAGS $CXXFLAGS"
+  AC_DEFINE([HAVE_LOWDOWN], [1], [Use lowdown]
+  have_lowdown=1)
+], [
+  have_lowdown=0
+  doc_generate=no
+])
+
 
 # Setuid installations.
 AC_CHECK_FUNCS([setresuid setreuid lchown])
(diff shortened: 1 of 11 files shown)


corngood:
@edolstra are you able to take a look or recommend a reviewer?

Ericson2314:
It would be very good to see this revived! The Nix team is more active now and so we can ensure it won't get neglected.

What is the state of Cygwin/Msys2 cross in Nixpkgs? That one issue would be merging things we cannot CI. But it shouldn't be too hard to fix that.

corngood:
This is a better place for what I posted on #8901:

I've never had much trouble building nix on cygwin. What I did have a lot of trouble with is fibers/coroutines, which are used by nix via boost. They are sort of fundamentally broken in cygwin, which causes nix to crash (e.g. on network operations).

I actually proposed a patch to cygwin which allows them to work, but it hasn't been merged:

https://cygwin.com/pipermail/cygwin-developers/2020-September/011970.html

Last time I did any work on this, it was mostly on the nixpkgs side. The idea was to be able to cross compile a patched cygwin1.dll, and use that to bootstrap, and build up from there. Unfortunately I haven't had time to work on it for a while.

I gave up on bootstrapping using unmodified cygwin.

If anyone's interested in seeing where I was going:

https://github.com/corngood/nix/tree/cygwin

It's is a mess of WIP commits, but this will show you the changes from the base:

https://github.com/corngood/nixpkgs/compare/66de24bef8351fb6ee4bd8e75717e99e1ada5794..cygwin

There are some other things in there that I think are important for cygwin, such as stateless rebasing of dlls, and using native symlinks for dlls.

I'd love to get back to this at some point, but performance on cygwin (forking especially) is so bad that it makes it extremely time consuming to build a stdenv.

Ericson2314:
OK Awesome! Yes let's take it to the Nixpkgs, and start merging PRs there for your stuff, and then come back to this. :)

Artturin:
@corngood @Ericson2314 FYI a improvement to win-dll-links was merged https://github.com/NixOS/nixpkgs/pull/252459

Ericson2314:
This can be closed now, right? :)

corngood:
Yeah, I'll close this.  I may have more changes when I get to cygwin-hosted builds, but that'll be using nixpkgs instead of cygwin packages.

PR #4955, opened 2021-06-28
Only show "available" packages in nix search in legacyPackages

matthewbauer:
legacyPackages can export meta.available to show whether they are
available or not. If it’s unset, we assume it’s true. If available is
false, we don’t need to show it in nix search.

/cc @edolstra

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -109,6 +109,12 @@ struct CmdSearch : InstallableCommand, MixJSON
                     std::replace(description.begin(), description.end(), '\n', ' ');
                     auto attrPath2 = concatStringsSep(".", attrPath);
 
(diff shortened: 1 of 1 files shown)


kvtb:
Should `nix` be so aware about `nixpkgs` structure and treat some names magically?

vcunat:
If you mean `legacyPackages`, that name is already defined by the (experimental) `flake` feature, not nixpkgs.

Ericson2314:
> Should `nix` be so aware about `nixpkgs` structure and treat some names magically?

Yes the layering makes this dubious to me. I want there to be a "Nixpkgs command" that can do all sorts of nice things without causing compat headaches.

PR #4956, opened 2021-06-28
set up builder pipe

domenkozar:
@Ericson2314 any idea why this was commented?

An attempt to debug https://github.com/NixOS/nix/issues/3605


domenkozar:
Seems like it was @edolstra in e84c265645

edolstra:
I think this line should be removed.

PR #4957, opened 2021-06-29
nix-gc: ignore non-existing links

Mic92:
Before I would get:

finding garbage collector roots...
deleting garbage...
deleting '/nix/store/trash'
deleting unused links...
error: getting status of '/nix/store/.links/1gwb16smc3b53iaa5nzfxjypys4pl32liw59aw01zvfjpaz53wg': No such file or directory

### DIFF
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -664,7 +664,11 @@ void LocalStore::removeUnusedLinks(const GCState & state)
         if (name == "." || name == "..") continue;
         Path path = linksDir + "/" + name;
 
-        auto st = lstat(path);
+        struct stat st;
+        if (lstat(path.c_str(), &st)) {
+            if (errno == ENOENT) continue;
+            throw SysError("getting status of %1%", path);
+        }
 
         if (st.st_nlink != 1) {
             actualSize += st.st_size;


edolstra:
Hm, how can we get ENOENT if the file is returned by readDir?

domenkozar:
link destination doesn't exist?

Mic92:
> Hm, how can we get ENOENT if the file is returned by readDir?

It seems that something in the process is deleting the link while the directory is read.
I was able to reproduce it even when I created a link to some location.

Mic92:
> link destination doesn't exist?

`lstat` does not resolve link targets

Mic92:
Not sure if it is relevant but I have enabled `auto-optimise-store = true` and the comments above suggest that there might be some race conditions regarding that.

edolstra:
In principle, `auto-optimise-store` only adds links so it shouldn't interfere with the garbage collector. And `removeUnusedLinks` is called while holding the global GC lock, so there should be only one instance of it running at any time.

Mic92:
> In principle, `auto-optimise-store` only adds links so it shouldn't interfere with the garbage collector. And `removeUnusedLinks` is called while holding the global GC lock, so there should be only one instance of it running at any time.

It does so. I can still reproduce this issue without this patch:

```
 sudo /nix/store/2k292jfgxi65vn799salvwl1rmr796di-unit-script-nix-gc-start/bin/nix-gc-start
finding garbage collector roots...
deleting garbage...
deleting '/nix/store/trash'
deleting unused links...
error: getting status of '/nix/store/.links/1gwb16smc3b53iaa5nzfxjypys4pl32liw59aw01zvfjpaz53wg': No such file or directory
sudo   5,76s user 5,02s system 99% cpu 10,782 total
```

Mic92:
I disabled the auto-optimize-store option for now. I will probably add a warning in the nixos module to not able it with nixUnstable as it will break things in the current state.

Mic92:
Ok. I just realized a weird file in there:

```
/nix/store/.links/1gwb16smc3b53iaa5nzfxjypys4pl32liw59aw01zvfjpa$'\366'z53wg
```

I was not able to delete this manually on my filesystem but I was able to move it to `/.trash`

edolstra:
Interesting, some kind of filesystem corruption?

Mic92:
> Interesting, some kind of filesystem corruption?

Not so sure. It is zfs so filesystem corruption are a lot less likely than bugs somewhere in the stack. However I have seen weird characters in nix store paths before, for example in nix sqlite database, which I had to remove manually.

PR #4961, opened 2021-06-29
nix profile: add aliases

ethancedwards8:
This PR adds aliases to the `nix profile` subcommand, which I think are helpful and I was surprised aren't there already.

A case can be made for aliases for `install` or `upgrade`, but I think for right now only having them for `list` and `remove` is fine.

### DIFF
--- a/src/nix/profile.cc
+++ b/src/nix/profile.cc
@@ -549,8 +549,10 @@ struct CmdProfile : NixMultiCommand
         : MultiCommand({
               {"install", []() { return make_ref<CmdProfileInstall>(); }},
               {"remove", []() { return make_ref<CmdProfileRemove>(); }},
+              {"rm", []() { return make_ref<CmdProfileRemove>(); }},
               {"upgrade", []() { return make_ref<CmdProfileUpgrade>(); }},
               {"list", []() { return make_ref<CmdProfileList>(); }},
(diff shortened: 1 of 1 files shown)


ethancedwards8:
Builds on MacOS M1 using Rosetta.

edolstra:
I'm not convinced this is a good idea. If the goal is to reduce typing, I think tab completion is good enough for that. And adding aliases pollutes the search space for tab completion.

If we do decide to add aliases, we should have some guidance in doc/manual/src/contributing/cli-guideline.md on when to add an alias.

BTW, I think implemented like this, the manual pages will be duplicated for each alias.

ethancedwards8:
> I'm not convinced this is a good idea. If the goal is to reduce typing, I think tab completion is good enough for that. And adding aliases pollutes the search space for tab completion.

IMO the goal was less to "reduce typing" and more "conform to traditional standards".
> 
> If we do decide to add aliases, we should have some guidance in doc/manual/src/contributing/cli-guideline.md on when to add an alias.
> 
> BTW, I think implemented like this, the manual pages will be duplicated for each alias.

Yes I believe that is so.

One option I have thought about is creating an option for aliases in nix.conf, that way man pages will not be cluttered, and it gives the end user more fine grain control.

ethancedwards8:
Is this still open for discussion?

PR #4967, opened 2021-06-30
Prefer to throw specific errors

Pamplemousse:
Some errors could be made less generic to allow finer catching and adaptation from calling code.


src/libexpr/eval.cc
```diff
@@ -908,7 +908,7 @@ void EvalState::evalFile(const Path & path_, Value & v, bool mustBeTrivial)
         // computation.
         if (mustBeTrivial &&
             !(dynamic_cast<ExprAttrs *>(e)))
-            throw Error("file '%s' must be an attribute set", path);
+            throw FormatError("file '%s' must be an attribute set", path);
```
regnat, line 911: I’d rather make this an `EvalError` as it’s not about the file being syntactically invalid, but more about the evaluation semantics

regnat (approved):

PR #4968, opened 2021-06-30
Make `nix registry` suitable for maintaining custom registries

balsoft:
> nix registry: add --registry flag

Add a `--registry` flag to `nix registry add`, `nix registry pin` and `nix registry remove`. Useful to maintain custom registries.

> nix registry pin: add a way to pin to a custom locked

Add an optional second argument to `nix registry pin`, that specifies a flake reference to be resolved as locked. Makes it more consistent with `nix registry add`.


balsoft:
I have used this version to create https://github.com/serokell/flake-registry, appears to work just fine.

PR #4969, opened 2021-06-30
flake.nixConfig: fix flake-registry config settings

balsoft:
Before this commit, nixConfig.flake-registry didn't have any real effect
on the evaluation, since config was applied after inputs were evaluated.
Change this behavior: apply the config in the beginning of flake::lockFile.

Tested on single-user and multi-user systems.

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -298,8 +298,10 @@ LockedFlake lockFlake(
 
     auto flake = getFlake(state, topRef, lockFlags.useRegistries, flakeCache);
 
-    flake.config.apply();
-    // FIXME: send new config to the daemon.
+    if (lockFlags.applyNixConfig) {
+        flake.config.apply();
+        // FIXME: send new config to the daemon.
+    }
 
     try {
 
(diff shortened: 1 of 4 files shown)


mkaito (approved):
> +3 −2

My favourite kind of PR

src/libexpr/flake/flake.cc
```diff
@@ -298,6 +298,9 @@ LockedFlake lockFlake(
 
     auto flake = getFlake(state, topRef, lockFlags.useRegistries, flakeCache);
 
+    flake.config.apply();
```
regnat, line 301: Unless I miss something, this will also apply for the `getFlake` builtin, which would be quite impure.

Maybe you can add a new field to `LockFlags` so that it’s only triggerd conditionally

regnat (changes requested):

src/libexpr/flake/flake.cc
balsoft: Good point!

Added a flag and set it true when:

- used through `getLockedFlake` in installable
- `nix flake {update,lock}`

PR #4971, opened 2021-07-02
Respect TERM=dumb more consistently

ylh:
This is a huge quality-of-life improvement for my niche use case:
current
patched


edolstra:
Looks good, but maybe instead of `TERM=dumb` we should respect `NO_COLOR` instead or additionally (see https://no-color.org/).

ylh:
First time hearing of that convention, but by the looks of it, checking for `NO_COLOR` in addition seems like the most sensible option. I've tacked on a commit accordingly.

edolstra:
Thanks!

PR #4972, opened 2021-07-02
Redirect some Nix commands output to a pager 

### LINKED ISSUE #4548: Revive piping output into the pager
Some of the old CLI commands (like `nix-env -qa`) automatically pipe results into the pager. The new CLI should do this as well (in particular `nix search`).

thufschmitt:
Make `nix search`, `nix flake show` and `nix flake metadata` pipe their
output into a pager to make their result easier to visualise.

Fix #4548

### DIFF
--- a/src/libcmd/repl.cc
+++ b/src/libcmd/repl.cc
@@ -29,6 +29,7 @@ extern "C" {
 #include "attr-path.hh"
 #include "store-api.hh"
 #include "log-store.hh"
+#include "loggers.hh"
 #include "common-eval-args.hh"
 #include "get-drvs.hh"
 #include "derivations.hh"
--- a/src/libmain/loggers.cc
+++ b/src/libmain/loggers.cc
@@ -1,5 +1,6 @@
 #include "loggers.hh"
 #include "progress-bar.hh"
+#include "paged-logger.hh"
 #include "util.hh"
 
 namespace nix {
@@ -51,6 +52,16 @@ void setLogFormat(const LogFormat & logFormat) {
 
 void createDefaultLogger() {
     logger = makeDefaultLogger();
+}
+
+RunPager::RunPager()
+    : previousLogger(logger)
+{
+    logger = new PagedLogger(previousLogger);
+}
+
+RunPager::~RunPager() {
+    logger = previousLogger;
 }
 
 }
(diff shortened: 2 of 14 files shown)


edolstra:
So the reason I didn't enable paging yet is that it can interact badly with the progress indicator. Maybe what we should do is show the progress indicator until the first output is written, then disable the progress indicator and start the pager.

regnat:
> So the reason I didn't enable paging yet is that it can interact badly with the progress indicator. Maybe what we should do is show the progress indicator until the first output is written, then disable the progress indicator and start the pager.

I’ve actually disabled the progress indicator altogether when using the pager ([here](https://github.com/NixOS/nix/pull/4972/files#diff-aca69dfa28aaf6fff9e8629ac7240576bd8d5a3eab1c52ca30aea44d5e6da1bdR389)). But waiting for the first output might be nicer (I’ve no idea how to do that though :) ).

edolstra:
Ideally we would use `logger->writeToStdout` everywhere (instead of `std::cout`), and then the progress bar logger can disable itself and start the pager the first time `writeToStdout` is called.

regnat:
@edolstra I’ve rewritten the proxy mechanism in 8ad1ed5fb66b9443997aac761c1baf6771b30160 to wrap a logger rather than being its own different thing. That way we can nicely intercept `writeToStdout` to start the pager as needed and tell the underlying pager not to dump arbitrary stuff on stdeer anymore

src/nix/flake.cc.orig
```diff
@@ -0,0 +1,1180 @@
+#include "command.hh"
```
SuperSandro2000, line 1: You included the file by accident, right?

SuperSandro2000:
Does less -R support the progress bar?

src/nix/flake.cc.orig
regnat: Ooops, yes. Thanks, fixed

regnat:
> Does less -R support the progress bar?

`-R` wouldn’t. `-r` might, but it’s a bit hacky (and explicitly not recommended), so I’d rather not rely on that

thufschmitt:
@edolstra friendly ping :)

edolstra:
Marking it as draft since it has a bunch of conflicts. Anybody want to pick this up?

tomberek:

PR #4973, opened 2021-07-02
Add ignore-git option

balsoft:
Allows to ignore .git directories when interpreting local flake references,
thus interpreting all local flakes as path:// and not git+file://. This
has multiple effects, notably:

- (pro) Allows to use flakes without adding flake.nix to the index (useful for
  repositories where flake.nix cannot be commited for one reason or another,
  see https://discourse.nixos.org/t/my-painpoints-with-flakes/9750 for example);
- (con) Disables the gitignore behavior;
- (con) Puts the entire .git into nix store.

### DIFF
--- a/src/libexpr/flake/flakeref.cc
+++ b/src/libexpr/flake/flakeref.cc
@@ -120,38 +120,40 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
             if (!allowMissing && !pathExists(path + "/flake.nix"))
                 throw BadURL("path '%s' is not a flake (because it doesn't contain a 'flake.nix' file)", path);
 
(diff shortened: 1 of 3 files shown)


blaggacao:
For completeness sake, another con is &mdash; I guess: doesn't provide `self.rev`

balsoft:
Note: this may not be needed. I completely forgot about the already-existing behavior of `path:`. E.g. `nix build --ignore-git .` is equivalent to `nix build path:.`

xaverdh:
Well it would be nice to have a flag which prevents adding the lock file to the index though?

L-as:
I don't think this is needed to be quite honest.

PR #4975, opened 2021-07-02
upload-release.pl: add aarch64-darwin

domenkozar:


edolstra:
Actually this PR isn't needed because releases are made using upload-release.pl from master.

PR #4976, opened 2021-07-02
Add a use-registries config option (and deprecate `--no-registries` )

balsoft:
Some people want to avoid using registries at all on their system; Instead
of having to add --no-registries to every command, this commit allows to
set registries = false in the config. --no-registries is still allowed
everywhere it was allowed previously, but now as a shorthand to --option
registries false rather than a separate flag.

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -1,4 +1,5 @@
 #include "flake.hh"
+#include "eval.hh"
 #include "lockfile.hh"
 #include "primops.hh"
 #include "eval-inline.hh"
@@ -296,7 +297,9 @@ LockedFlake lockFlake(
 
     FlakeCache flakeCache;
 
-    auto flake = getFlake(state, topRef, lockFlags.useRegistries, flakeCache);
+    auto useRegistries = lockFlags.useRegistries.value_or(settings.useRegistries);
+
+    auto flake = getFlake(state, topRef, useRegistries, flakeCache);
 
     if (lockFlags.applyNixConfig) {
         flake.config.apply();
@@ -464,7 +467,7 @@ LockedFlake lockFlake(
                             throw Error("cannot update flake input '%s' in pure mode", inputPathS);
 
                         if (input.isFlake) {
-                            auto inputFlake = getFlake(state, *input.ref, lockFlags.useRegistries, flakeCache);
+                            auto inputFlake = getFlake(state, *input.ref, useRegistries, flakeCache);
 
                             /* Note: in case of an --override-input, we use
                                the *original* ref (input2.ref) for the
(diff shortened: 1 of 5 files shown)


regnat:
> Some people want to avoid using registries at all on their system

Isn’t it simpler to just set `flake-registry` to an empty registry in that case? (possibly making the empty file or empty json object a valid registry to make this even simpler)

balsoft:
> Isn’t it simpler to just set flake-registry to an empty registry in that case?

That was my first thought as well, but this still leaves a system registry (`/etc/nix/registry.json`) and a user registry (`$XDG_CONFIG_HOME/nix/registry.json`). This option disables registries completely.

nrdxp:
Until this is merged, it is also possible to simply specify an empty registry, to simulate no registry:
```json
# registry.json
{
  "version": 2
}
```

balsoft:
> Until this is merged, it is also possible to simply specify an empty registry, to simulate no registry:

Once again, this doesn't quite work since there are also user and system registries.

src/libstore/globals.hh
```diff
@@ -956,6 +956,10 @@ public:
           resolves to a different location from that of the build machine. You
           can enable this setting if you are sure you're not going to do that.
         )"};
+
+    // The setting is named "registries" to keep compability with the old --no-registries flag
+    Setting<bool> useRegistries{this, true, "registries",
```
edolstra, line 961: This should be `use-registries` or something like that, since most of our boolean options are a verb (e.g. `keep-failed`). `registries` suggests that this is the list of registries.

balsoft: As noted in the comment, this is to keep compatibility with `--no-registries` option used in flake commands (such as `nix build --no-registries`). If you think it's better to keep both the option and a config setting, I can do that.

src/libexpr/flake/flake.hh
```diff
@@ -102,7 +102,7 @@ struct LockFlags
 
     /* Whether to use the registries to lookup indirect flake
        references like 'nixpkgs'. */
-    bool useRegistries = true;
+    std::optional<bool> useRegistries = std::nullopt;
```
edolstra, line 105: A tristate is confusing to me, maybe it should just be initialized to `settings.registries`?

src/libexpr/flake/flake.cc
```diff
@@ -296,7 +297,12 @@ LockedFlake lockFlake(
 
     FlakeCache flakeCache;
 
-    auto flake = getFlake(state, topRef, lockFlags.useRegistries, flakeCache);
+    bool useRegistries = settings.useRegistries;
+
+    if (lockFlags.useRegistries.has_value())
+      useRegistries = *lockFlags.useRegistries;
```
edolstra, line 303:
```suggestion
    auto useRegistries = lockFlags.useRegistries.value_or(settings.useRegistries);
```

src/libexpr/flake/flake.hh
balsoft: I have thought about this a bit, I don't like initializing fields with global state. It may be confusing if the state changes after the structure has been initialized, since then `lockFlags.useRegistries` may go out of sync with `settings.useRegistries`.

src/libstore/globals.hh
edolstra: Yeah, let's rename it to `use-registries` and provide `--no-registries` as a deprecated option that will be removed eventually.

balsoft: @edolstra done

PR #4980, opened 2021-07-03
libcmd/installables: implement completion for Nix expressions passed via `-f`

Ma27:
Note: this isn't extensively tested yet, but I'm using it from now on for the Nix installation on my workstation.

cc @edolstra @domenkozar

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -171,14 +171,44 @@ Strings SourceExprCommand::getDefaultFlakeAttrPathPrefixes()
 
 void SourceExprCommand::completeInstallable(std::string_view prefix)
 {
-    if (file) return; // FIXME
+    if (file) {
+        evalSettings.pureEval = false;
+        auto state = getEvalState();
+        Expr *e = state->parseExprFromFile(
+            resolveExprPath(state->checkSourcePath(lookupFileArg(*state, *file)))
+        );
 
-    completeFlakeRefWithFragment(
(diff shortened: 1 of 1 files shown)


src/libcmd/installables.cc
```diff
@@ -171,14 +171,44 @@ Strings SourceExprCommand::getDefaultFlakeAttrPathPrefixes()
...
+        }
+
+        Value &v1(*findAlongAttrPath(*state, prefix_, *autoArgs, root).first);
+        state->forceValue(v1);
+        Value v2;
+        state->autoCallFunction(*autoArgs, v1, v2);
+
+        if (v2.type() == nAttrs) {
+            for (auto & i : *v2.attrs) {
+                completions->add(i.name);
```
edolstra, line 201: This needs to filter out names that don't start with `prefix`. Otherwise I get:
```
# nix build -f ../nixpkgs hel<TAB>
Display all 14279 possibilities? (y or n)
```

It may be possible to factor out some commonality with `NixRepl::completePrefix`.

Ma27: Ahhh, `bash` doesn't filter it out on its own? I have to admit I only tested it with `zsh` which does this :) 
Thanks for pointing it out, gotta fix this!

Ma27: OK, fixed!

nrdxp:
Tested this and it works with no issues so far.

PR #4983, opened 2021-07-05
Fix devShell handling of env values including @ and %

manveru:
Given the following `devShell`, both `nix develop` and `nix print-dev-env` would fail parsing the environment file:

```shell
❮ nix --experimental-features 'nix-command flakes' -L print-dev-env
warning: Git tree '/home/manveru/tmp/quoting' is dirty
error: shell environment '/nix/store/xwc73vzp1lcsw1if8vs76r9z4xkgpfyl-test-env' has unexpected line 'BAR=@
       BASH=/noshell
       BASHOPTS=cmdhist:extquote:interactive_co'
```

here's the flake i used to test this:

```nix
{
  outputs = { nixpkgs, ... }: {
    devShell.x86_64-linux = derivation {
      name = "test";
      builder = "${nixpkgs.legacyPackages.x86_64-linux.bash}/bin/bash";
      args = [
        (builtins.toFile "builder" ''
          set -exuo pipefail
          echo hier
          echo $FOO > $out
        '')
      ];
      outputs = [ "out" ];
      system = "x86_64-linux";
      FOO = "%";
      BAR = "@";
    };
  };
}
```

I included a simple test that is now part of the normal `shell.nix` tests, and should prevent regressions.

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -54,7 +54,7 @@ BuildEnvironment readEnvironment(const Path & path)
         R"re((?:[a-zA-Z_][a-zA-Z0-9_]*))re";
 
     static std::string simpleStringRegex =
-        R"re((?:[a-zA-Z0-9_/:\.\-\+=]*))re";
+        R"re((?:[a-zA-Z0-9_/:\.\-\+=@%]*))re";
 
     static std::string dquotedStringRegex =
         R"re((?:\$?"(?:[^"\\]|\\[$`"\\\n])*"))re";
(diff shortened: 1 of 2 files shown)


regnat (approved):
Thanks, that looks good.

That’s not a complete fix, as it’s still possible to break it (picking stuff from https://github.com/minimaxir/big-list-of-naughty-strings ), but it’s already strictly better

manveru:
Yeah, the easiest way to break it atm is to use a backslash, but that complicates the regex too much and seems uncommon enough that I didn't want to tackle it here.

PR #4986, opened 2021-07-05
Fix wrong hash var for aarch64-darwin binary

jkeifer:
I noticed the nix:master build in hydra does not properly add the tarball hash for the aarch64-darwin build. From the build logs it appears the proper template var name is `@tarballHash_aarch64-darwin@`, not `@binaryTarball_aarch64-darwin@`.


domenkozar:
Does this need backport to 2.3?

jkeifer:
I don't believe it needs to be backported. It appears the hash vars are named differently between the two builds and that the 2.3 maintenance build is templating the hash correctly already.

PR #4988, opened 2021-07-06
Add a name attribute to the fetchers

### LINKED ISSUE #3888: builtins.fetchGit doesn't support the name attribute anymore
With `nix-2.3.7` adding the `name` attribute to a call to `builtins.fetchGit` correctly results in a store path with the specified name:

```
nix-instantiate --eval -E 'builtins.fetchGit { name = "foo"; rev = "..."; url = "..."; }'
{ outPath = "/nix/store/...-foo"; rev = "..."; revCount = ...; shortRev = "..."; }
```

However in nix `master` (a2842588ec86a0f488a385d453eda86e1f52f05a) this is not supported anymore:

```
nix-instantiate --eval -E 'builtins.fetchGit { name = "foo"; rev = "..."; url = "..."; }'
error: --- Error ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix-instantiate
unsupported Git input attribute 'name'
```

Is this as intended?

@edolstra this looks to be related to this [FIXME](https://github.com/NixOS/nix/pull/3459/files#diff-7e39b3dccfe1a07ed3f4af7f4d8da628R57).

regnat:
Allow `fetchGit`, `fetchTarball`, `fetchMercurial` and `fetchFromGitHub` to take a `name` argument

Fix #3888

### DIFF
--- a/src/libexpr/primops/fetchMercurial.cc
+++ b/src/libexpr/primops/fetchMercurial.cc
@@ -62,6 +62,7 @@ static void prim_fetchMercurial(EvalState & state, const Pos & pos, Value * * ar
     fetchers::Attrs attrs;
     attrs.insert_or_assign("type", "hg");
     attrs.insert_or_assign("url", url.find("://") != std::string::npos ? url : "file://" + url);
+    attrs.insert_or_assign("name", name);
     if (ref) attrs.insert_or_assign("ref", *ref);
     if (rev) attrs.insert_or_assign("rev", rev->gitRev());
     auto input = fetchers::Input::fromAttrs(std::move(attrs));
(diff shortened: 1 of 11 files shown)


edolstra:
I'm not sure whether `fetchTree` should support `name`. We need it for the other builtins for backward compatibility, but `fetchTree` is new so we don't have that issue.

Does this allow a name attribute in flake URL? As with `fetchTree`, that may be undesirable since it reduces CA effectiveness.

Otherwise LGTM.

regnat:
> I'm not sure whether fetchTree should support name. We need it for the other builtins for backward compatibility, but fetchTree is new so we don't have that issue.

Oh you’re right. I’ve explicitly forbidden it in 7e5c79a2d210e7b986803d5fe5c7ebb504b3e74e

> Does this allow a name attribute in flake URL? As with fetchTree, that may be undesirable since it reduces CA effectiveness.

No, trying to add a `name` parameter to a flake url will fail with `error: unsupported input attribute 'name'`

PR #4989, opened 2021-07-06
 Move `nix flake {new,init}` at toplevel

### LINKED ISSUE #4502: Rename common flake commands
Common flake commands shouldn't be inside `nix flake` subcommand.

I suggest renaming the following:
 - `nix flake init` -> `nix init`
 - `nix flake new` -> `nix new`

/cc @edolstra

thufschmitt:
Fix #4502

Requires reworking a bit the existing alias mechanism to allow aliases to be
composite commands themselves (like `flake init` and `flake new` are)

Depends on #5881
This only should happen once flakes are no longer marked as experimental


tomberek:
Should this be delayed until flakes are no longer experimental?

tomberek:

PR #4993, opened 2021-07-07
Be more helpful when a flake eval is missing some files

### LINKED ISSUE #4507: Be smart and helpful when some files are missing during the evaluation
A change that is (already) confusing some Nix developers (including me) is that in the case of project repository being tracked under git only files tracked by git are visible by `nix build`. It happens that a user will forget to add a file to git and `nix build` will complain that it can not find a file during `nix build`.

In contrast "old" `nix-build` will usually copy everything under the project folder (well depends what your `src = ...` looks like).

Now there is a good reason that this behavior changed. With this a user will catch a problem earlier locally instead of pushing to upstream and only then realizing that CI failed due to a missing file.

But we need to help the user understand what is happening and that this is a feature that is saving him time and not just 

I propose to do the following:

- In case of missing nix files we can already suggest what the problem can be. This error happens during the evaluation. The edge case we must make sure it works is a missing `flake.nix` (this is possible if you first create a nix project and only then git repository). With this error we can point directly to which file is missing

- In case of non-nix missing files we can not directly point which file(s) is missing and causing `nix build` to fail. What we can show is that there is a number of untracked files in Git and that Nix when used with git will only see files tracked by git. Please run `git status` and add missing files.

/cc @edolstra

thufschmitt:
When building a local git flake, if a file a missing at eval-time but
present in the source directory, tell the user that he probably forgot
to check it out

Fix #4507

The current approach is far from complete as is only works for `nix build`, but having a more generic thing is actually quite complex or hacky, and I’d like to have some feedback on my plan before implementing it.

My idea is to

- Add a `void enhanceMessage(void(void))` method to `Installable` that’s semantically just calling its argument, but can also add some informations to the error message if one is thrown.
    (In practice, this will be a no-op, except for `InstallableFlake` when the flake input is a local git repo − or probably mercurial too)
- Override `run()` in `InstallableCommand` to run `installable.enhanceMessage(run(store))`

I think this would work, but it isn’t utterly pretty, so if you have any concern or better idea, feel free to chime in.

Depends on #3121 
We depends on above ticket because a ticket provides needed information without guessing it.

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -528,26 +528,45 @@ std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableF
     auto cache = openEvalCache(*state, lockedFlake);
     auto root = cache->getRoot();
 
-    for (auto & attrPath : getActualAttrPaths()) {
-        auto attr = root->findAlongAttrPath(
-            parseAttrPath(*state, attrPath),
-            true
-        );
+    try {
+        for (auto & attrPath : getActualAttrPaths()) {
+            auto attr = root->findAlongAttrPath(
+                parseAttrPath(*state, attrPath),
+                true
+            );
 
-        if (!attr) continue;
+            if (!attr) continue;
 
-        if (!attr->isDerivation())
-            throw Error("flake output attribute '%s' is not a derivation", attrPath);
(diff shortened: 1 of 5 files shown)


thufschmitt:
Superseded by #3121

PR #4996, opened 2021-07-08
Add a matrix of stdenvs to the flake

### LINKED ISSUE #4129: Restore the ability to build with clang on linux

Before the migration to a flake, the `shell.nix` file took a `useClang` argument that allowed to easily use `clangStdenv` rather than the default one. This argument doesn't exist anymore, and to the best of my knowledge, the only way to use clang is to patch the `flake.nix` file

Restore a simple way to use clang rather than gcc for the build

**Additional context**

I guess adding back use `useClang` argument isn't possible (https://github.com/NixOS/nix/issues/3843#issuecomment-661710951). Maybe it's possible to just add a new set of platforms (something like `x86_64-linux-clang` and friends) to the flake, but I'm not aware of anything on the nixpkgs side to handle that

regnat:
For a (currently hardcoded and limited) list of stdenvs,
make `.#$nix-${stdenvName}` correspond to a Nix built with the
corresponding stdenv.

For example, `.#nix-${clang11Stdenv}` is Nix built with clang11.
    
Likewise, `devShell.x86_64-linux.clang11StdenvPackages` is a development shell for Nix with clang11 

Fix #4129

/cc @pamplemousse

### DIFF
--- a/doc/manual/src/contributing/hacking.md
+++ b/doc/manual/src/contributing/hacking.md
@@ -35,6 +35,25 @@ variables are set up so that those dependencies can be found:
 $ nix-shell
 ```
 
+or if you have a flake-enabled nix:
+
+```console
+$ nix develop
+```
+
+To get a shell with a different compilation environment (e.g. stdenv,
+gccStdenv, clangStdenv, clang11Stdenv):
+
+```console
+$ nix-shell -A devShells.x86_64-linux.clang11StdenvPackages
+```
+
+or if you have a flake-enabled nix:
+
+```console
+$ nix develop .#clang11StdenvPackages
+```
+
 To build Nix itself in this shell:
 
 ```console
--- a/flake.nix
+++ b/flake.nix
@@ -22,15 +22,36 @@
 
       crossSystems = [ "armv6l-linux" "armv7l-linux" ];
 
+      stdenvs = [ "gccStdenv" "clangStdenv" "clang11Stdenv" "stdenv" ];
+
       forAllSystems = f: nixpkgs.lib.genAttrs systems (system: f system);
+      forAllSystemsAndStdenvs = f: forAllSystems (system:
+        nixpkgs.lib.listToAttrs
+          (map
+            (n:
+            nixpkgs.lib.nameValuePair "${n}Packages" (
+              f system n
+            )) stdenvs
+          )
+      );
+
+      forAllStdenvs = stdenvs: f: nixpkgs.lib.genAttrs stdenvs (stdenv: f stdenv);
 
       # Memoize nixpkgs for different platforms for efficiency.
-      nixpkgsFor = forAllSystems (system:
-        import nixpkgs {
-          inherit system;
-          overlays = [ self.overlay ];
-        }
-      );
+      nixpkgsFor =
+        let stdenvsPackages = forAllSystemsAndStdenvs
+          (system: stdenv:
+            import nixpkgs {
+              inherit system;
+              overlays = [
+                (overlayFor (p: p.${stdenv}))
+              ];
+            }
+          );
+        in
+        # Add the `stdenvPackages` at toplevel, both because these are the ones
+        # we want most of the time and for backwards compatibility
+        forAllSystems (system: stdenvsPackages.${system} // stdenvsPackages.${system}.stdenvPackages);
 
       commonDeps = pkgs: with pkgs; rec {
         # Use "busybox-sandbox-shell" if present,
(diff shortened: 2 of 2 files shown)


flake.nix
```diff
@@ -591,15 +612,20 @@
...
+      nixpkgsFor.${system}.lib.nameValuePair
+        "nix-${stdenvName}"
+        nixpkgsFor.${system}."${stdenvName}Packages".nix
+      ) stdenvs))
+      );
 
       defaultPackage = forAllSystems (system: self.packages.${system}.nix);
 
-      devShell = forAllSystems (system:
+      devShell = forAllSystemsAndStdenvs (system: stdenv:
```
edolstra, line 624: I don't think we should make `devShell` an attrset of derivations/shells. It's assumed to be a single derivation, like `defaultPackage`. So e.g. `nix flake check` won't check these. Maybe to enforce uniformity we should disallow attributes deeper than the expected level in the CLI...

Instead we probably should add a `devShells` flake output (and maybe rename `devShell` to `defaultDevShell`). That would also be less verbose since it can be searched by default, so instead of
```
nix develop .#devShell.x86_64-linux.clang11StdenvPackages
```
you could do
```
nix develop .#clang11StdenvPackages
```

Ericson2314 (approved):
I do like this. I will build on this to duplicate the cross / static stuff.

(It was only ever the is only the way it use because there used to be an issue with `nix` being overridden in https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/static.nix, but it no longer is overridden there.)

flake.nix
regnat: Oh indeed, `devShells` is much nicer. I’ve changed that in 6378f362e, and I’ll probably open another PR shortly to make `nix develop` and the various `nix flake` commands aware of it

doc/manual/src/contributing/hacking.md
```diff
@@ -35,6 +35,25 @@ variables are set up so that those dependencies can be found:
...
+gccStdenv, clangStdenv, clang11Stdenv):
+
+```console
+$ nix-shell -A devShell.x86_64-linux.clang11StdenvPackages
+```
+
+or if you have a flake-enabled nix:
+
+```console
+$ nix develop .#devShell.x86_64-linux.clang11StdenvPackages
```
edolstra, line 54: This should be `.#clang11StdenvPackages` now right?

regnat: Indeed, thanks. And actually the flake search path was broken (was missing a trailing `.`), so I fixed it.

I’ve also fixed the command above to use `devShells` and not `devShell`

flake.nix
Pamplemousse: I tried to use `nix develop .#clang11StdenvPackages` with this PR, and I get an error:
```shell
nix --experimental-features "nix-command flakes" develop .#clang11StdenvPackages
warning: Git tree '/src/nix' is dirty
error: flake 'git+file:///src/nix?shallow=1' does not provide attribute 'devShells.x86_64-linuxclang11StdenvPackages', 'packages.x86_64-linux.clang11StdenvPackages', 'legacyPackages.x86_64-linux.clang11StdenvPackages' or 'clang11StdenvPackages'

nix --version
nix (Nix) 2.4pre20210805_d64f967
```

Also, completion gives other attributes:
```
.#defaultPackage     .#checks            .#                                                                               
.#hydraJobs          .#devShells         .#devShell                                                                       
.#nix-armv7l-linux   .#nix-armv6l-linux  .#nix                                                                            
.#nix-clang11Stdenv  .#nix-clangStdenv   .#nix-gccStdenv                                                                  
.#nix-stdenv         .#nix-static        .#overlay
```

It should be `nix develop .#nix-clang11Stdenv` .

Pamplemousse:
ALso, it seems that the `devShell` 's `shellHook` (https://github.com/NixOS/nix/blob/master/flake.nix#L614-L619) is ran OK when doing `nix develop`, but not when doing `nix develop .#nix-clang11Stdenv` .

Example:
```shell
$ nix develop .#nix-clang11Stdenv
warning: Git tree '/home/pamplemousse/Workspace/tools/nix' is dirty
$ echo $MANPATH

$ nix develop                    
warning: Git tree '/home/pamplemousse/Workspace/tools/nix' is dirty
$ echo $MANPATH
/home/pamplemousse/Workspace/tools/nix/outputs/out/share/man:
```

Pamplemousse:
Ah, nevermind!
I got confused because `nix develop .#clang11StdenvPackages` did not work (because of the missing `.`), so I started using `.#nix-clang11Stdenv`, which does *NOT* use `devShells`)...

Pamplemousse:
This is blocking for google/oss-fuzz#6317, as they require good upstream integration (they won't merge as long as I use my fork of `nix` over there).

Is there anything blocking the merge?

Ericson2314:
Can we get this in before 2.4? I think it is decently important, and I would like to delete the duplicate `mkDerivation` for static and cross nix too.

regnat:
Rebased on top of latest master.

@edolstra care to have a look?

PR #5009, opened 2021-07-13
Add 'auto' support to machines files

matthewbauer:
This lets the remote machine fill in extra details for system types and supported feature. For instance, you
can do this:

```
ssh-ng://me@nixos1.local auto /home/me/.ssh/id_rsa 4 0 auto
ssh-ng://me@nixos2.local auto /home/me/.ssh/id_rsa 4 1 auto
ssh-ng://me@nixos3.local auto /home/me/.ssh/id_rsa 8 2 auto
```

and each machine can determine what it supports.

### DIFF
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -130,6 +130,38 @@ static int main_build_remote(int argc, char * * argv)
                 for (auto & m : machines) {
                     debug("considering building on remote machine '%s'", m.storeUri);
 
+                    if (std::find(m.systemTypes.begin(), m.systemTypes.end(), "*") != m.systemTypes.end() ||
+                        std::find(m.supportedFeatures.begin(), m.supportedFeatures.end(), "*") != m.supportedFeatures.end()) {
+                        // wildcard is used, so we need to ask the machine what it supports
+                        try {
+                            Activity act(*logger, lvlTalkative, actUnknown, fmt("connecting to '%s'", m.storeUri));
+
+                            auto store = m.openStore();
+
+                            if (hasPrefix(m.storeUri, "ssh://"))
+                                throw Error("Cannot use wildcards with legacy ssh store. Instead use the ssh-ng:// protocol.");
+
+                            store->connect();
+
+                            if (std::find(m.supportedFeatures.begin(), m.supportedFeatures.end(), "*") != m.supportedFeatures.end()) {
+                                m.supportedFeatures.erase("*");
+                                m.supportedFeatures.insert(store->systemFeatures.get().begin(), store->systemFeatures.get().end());
+                            }
+
+                            if (std::find(m.systemTypes.begin(), m.systemTypes.end(), "*") != m.systemTypes.end()) {
+                                m.systemTypes.erase(std::find(m.systemTypes.begin(), m.systemTypes.end(), "*"));
+                                m.systemTypes.insert(m.systemTypes.end(), store->systemTypes.get().begin(), store->systemTypes.get().end());
+                            }
+                        } catch (std::exception & e) {
+                            auto msg = chomp(drainFD(5, false));
+                            printError("cannot connect to '%s': %s%s",
+                                bestMachine->storeUri, e.what(),
+                                msg.empty() ? "" : ": " + msg);
+                            m.enabled = false;
+                            continue;
+                        }
+                    }
+
                     if (m.enabled && std::find(m.systemTypes.begin(),
                             m.systemTypes.end(),
                             neededSystem) != m.systemTypes.end() &&
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -130,8 +130,8 @@ static int main_build_remote(int argc, char * * argv)
                 for (auto & m : machines) {
                     debug("considering building on remote machine '%s'", m.storeUri);
 
-                    if (std::find(m.systemTypes.begin(), m.systemTypes.end(), "*") != m.systemTypes.end() ||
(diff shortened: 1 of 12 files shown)


src/build-remote/build-remote.cc
```diff
@@ -130,6 +130,38 @@ static int main_build_remote(int argc, char * * argv)
...
+                    if (std::find(m.systemTypes.begin(), m.systemTypes.end(), "*") != m.systemTypes.end() ||
+                        std::find(m.supportedFeatures.begin(), m.supportedFeatures.end(), "*") != m.supportedFeatures.end()) {
+                        // wildcard is used, so we need to ask the machine what it supports
+                        try {
+                            Activity act(*logger, lvlTalkative, actUnknown, fmt("connecting to '%s'", m.storeUri));
+
+                            auto store = m.openStore();
+
+                            if (hasPrefix(m.storeUri, "ssh://"))
+                                throw Error("Cannot use wildcards with legacy ssh store. Instead use the ssh-ng:// protocol.");
```
regnat: I think this will not just fail for `ssh` stores, but also for any store other than `ssh-ng` (`/foo/bar`, `unix:/other/daemon/socket`, etc..) as their `systemTypes` field will never be set

src/build-remote/build-remote.cc
```diff
@@ -130,6 +130,38 @@ static int main_build_remote(int argc, char * * argv)
...
+
+                            store->connect();
+
+                            if (std::find(m.supportedFeatures.begin(), m.supportedFeatures.end(), "*") != m.supportedFeatures.end()) {
+                                m.supportedFeatures.erase("*");
+                                m.supportedFeatures.insert(store->systemFeatures.get().begin(), store->systemFeatures.get().end());
+                            }
```
regnat, line 150: I find it a bit nicer to do that in one pass by keeping an iterator to the found value (also, we can preserve the order when expanding the wildcard rather than adding the new elements at the begining):

```suggestion
                            if (auto wildcardIter = m.supportedFeatures.find("*"); wildcardIter != m.supportedFeatures.end()) {
                                m.supportedFeatures.insert(wildcardIter, store->systemFeatures.get().begin(), store->systemFeatures.get().end());
                                m.supportedFeatures.erase(wildcardIter);
                            }
```

src/build-remote/build-remote.cc
```diff
@@ -130,6 +130,38 @@ static int main_build_remote(int argc, char * * argv)
...
+
+                            if (std::find(m.systemTypes.begin(), m.systemTypes.end(), "*") != m.systemTypes.end()) {
```
regnat, line 152: Like above, keeping the iterator is probably more efficient and easier to read

src/build-remote/build-remote.cc
```diff
@@ -130,6 +130,38 @@ static int main_build_remote(int argc, char * * argv)
...
+                                m.systemTypes.erase(std::find(m.systemTypes.begin(), m.systemTypes.end(), "*"));
+                                m.systemTypes.insert(m.systemTypes.end(), store->systemTypes.get().begin(), store->systemTypes.get().end());
```
regnat, line 154: Any reason why you’re putting things at the end here while you add them at the begining above?

regnat (changes requested):
I like this idea a lot!

Can you add some tests for it?

Also, I’m not entirely convinced by the choice of `*` as for me it really conveys the meaning of “catch-all” (“this store can build everything”) rather than the current inference. Maybe reusing `auto` would be nicer? (Yes I know, this is pretty much bike shedding)

src/build-remote/build-remote.cc
regnat: Maybe the constructor of `LocalFSStore` could initialize these with the same values that are used in the new daemon code

matthewbauer: We can cover this in a test, but I think unix: (UDSRemoteStore) will work fine because it uses RemoteStore. We should be able to support anything that uses the worker protocol. We could support the ssh protocol too, but I got the impression it's meant to be gradually phased out.

src/build-remote/build-remote.cc
matthewbauer: no not particularly

matthewbauer: probably systemTypes should be a set because we don't really care about order here.

src/build-remote/build-remote.cc
regnat: Oh yes of course, `UDSRemoteStore` will work just fine.

I think the point still stand about the `LocalStore` though, esp. now that it’s possible to do something like `nix build --store https://my.binary.cache --builders 'local …'`

regnat (approved):

edolstra:
Not sure this is a good idea, since it makes every invocation of `build-remote` open an SSH connection to every `auto` machine, which has pretty high overhead.

Mic92:
Could this be fixed with better caching?

matthewbauer:
:+1: to a cache (or maybe just a builders db) - I'll try to take a look at it this weekend.

It actually might be useful for another project I have been looking at - making it possible to agree to new ssh host keys from within Nix. This would make trusted-users able to accept new keys or changed keys, without polluting /etc/ssh/known_hosts.

Ericson2314:
N.B. `auto` is also a valid store URL. Not sure if this makes this more confusing or, the opposite, nicer.

Ericson2314:
For https://github.com/NixOS/nix/pull/3921 there was already interested in @edolstra for better caching to remove the need for an additional store option. If we can do caching for that, we should be able to do caching for this!

(But we must do caching in Nix itself rather than the build hook, which is an ephemeral process. So this is blocked on a few other cleanups.)

Issue #4665, opened 2021-03-23
Deprecate legacy ssh store / nix-store --serve


Ericson2314:
Issues like #3372 and #2770 show that things slip through the cracks when we have two sorts of remote stores.

I think after the next release they will finally be at feature parity, we should then deprecate the legacy one so we only have one thing to maintain. 

Finally, we can then add the protocol version parameter to our serialization infra. This resolves the tension between structure / code reuse and and being able to evolve the protocol that has hamstrung https://github.com/NixOS/nix/pull/4588, for example.

rickynils:
@Ericson2314 Currently, no command line Nix options are used when using `ssh-ng://` (see https://github.com/NixOS/nix/issues/5600#issuecomment-984694263). I think that resolving it should be on the list.

Ericson2314:
Thanks, added.

asymmetric:
@Ericson2314 https://github.com/NixOS/nix/issues/2770 ticks the first box

nrdxp:
> @Ericson2314 Currently, no command line Nix options are used when using `ssh-ng://` (see [#5600 (comment)](https://github.com/NixOS/nix/issues/5600#issuecomment-984694263)). I think that resolving it should be on the list.

I'd like to ask specifically about signing. Right now it is already possible to capture an entire remote build and upload cycle in a single command: `nix copy --eval-store auto --store ssh-ng://my.builder.com --to s3://myBucket`. This is pretty great for building a package on some remote builder and then copying it to your intended cache all in one go.

The only issue I have run into is signing. Because the nix.conf setting `nix-secret-key-files` doesn't propagate, I have to run a `nix store sign` manually on the artifact, which is fine for s3 backed caches, but this doesn't work for ssh stores. It'd be nice if there was either a flag to `nix copy` to specify a singing key explicitly, or if `nix-secret-key-files` was respected by the remote builder.

jmgilman:
> > @Ericson2314 Currently, no command line Nix options are used when using `ssh-ng://` (see [#5600 (comment)](https://github.com/NixOS/nix/issues/5600#issuecomment-984694263)). I think that resolving it should be on the list.
> 
[...]

PR #5010, opened 2021-07-13
Use ssh-ng by default

matthewbauer:
I think this is now safe to do. I haven’t had any issues using ssh-ng, but I suspect most users don’t know
about it.

Relates to #4665

### DIFF
--- a/src/libstore/machines.cc
+++ b/src/libstore/machines.cc
@@ -24,7 +24,7 @@ Machine::Machine(decltype(storeUri) storeUri,
         || hasPrefix(storeUri, "auto")
         || hasPrefix(storeUri, "/")
         ? storeUri
-        : "ssh://" + storeUri),
+        : "ssh-ng://" + storeUri),
     systemTypes(systemTypes),
     sshKey(sshKey),
     maxJobs(maxJobs),


Mic92:
Is there any writeup about what the difference between the two is? Do we get a changelog entry for this?

Mic92:
cc @rickynils since this is a breaking change for nixbuild.net I believe.

rickynils:
@Mic92 This shouldn't be breaking for nixbuild.net, since most of documentation for nixbuild.net explicitly states that you should use `ssh://eu.nixbuild.net`. So as long as the legacy remote builders still are supported, things should be fine for nixbuild.net. I do however intend to support `ssh-ng` too, since I want nixbuild.net to be completely interchangeable with a "normal" remote builder.

I would also be interested in some comparison between `ssh://` and `ssh-ng://`, although I guess I'll notice the differences when implementing this for nixbuild.net.

regnat:
One thing that I think `ssh://` does but not `ssh-ng://` is respect the `builders-use-substitutes` flag (I didn’t try it  recently, but https://github.com/NixOS/nix/blob/master/src/libstore/remote-store.cc#L291-L308 is ignoring the flag)

Ericson2314:
I tried to beef up #4665 to account for this, and potential blockers like what the @regnat mentioned.

rickynils:
nixbuild.net now supports ssh-ng:// in addition to ssh://, but it is not documented yet since we want to test it more.

nyabinary:
Any reason why this hasn't been set to default yet?

kjeremy:
I've been using this for a few years now. The only thing that I have run into is that ssh-ng stores do not work with `nix log`.

Mic92:
Basically this issue: https://github.com/NixOS/nix/issues/7359

Enzime:
Not sure if this has been fixed yet https://github.com/NixOS/nix/issues/6950

Ericson2314:
I would like to try doing this. (Another alternative is simply forcing the user to be explicit.)

Ericson2314:
I think I am going with forcing the user to be explicit

PR #5011, opened 2021-07-13
libexpr: Fix read out-of-bound on the heap

Pamplemousse:
* When `s` contains `\r\n`, `len` does not get decremented enough, and so more looping than necessary happens;
  * I replaced the `while` loop with a `for` because `len` is a `size_t`, thus decrementing it if its value is `0` would not stop the loop.
  
cc @regnat

### DIFF
--- a/src/libexpr/lexer.l
+++ b/src/libexpr/lexer.l
@@ -38,11 +38,13 @@ static void adjustLoc(YYLTYPE * loc, const char * s, size_t len)
     loc->first_line = loc->last_line;
     loc->first_column = loc->last_column;
 
-    while (len--) {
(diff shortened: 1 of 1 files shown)


edolstra:
This breaks some tests because it changes column/line numbers:
```
    evaluating lang/eval-okay-getattrpos-functionargs.nix (should succeed)
    1c1
    < { column = 1; file = "eval-okay-getattrpos-functionargs.nix"; line = 1; }
    ---
    > { column = 11; file = "eval-okay-getattrpos-functionargs.nix"; line = 2; }
```

Pamplemousse:
Good thing there are tests, my the conditional in the loop was broken :sweat_smile:

PR #5014, opened 2021-07-14
Log ssh output for ssh-ng too

matthewbauer:
I think this was just an oversight. Right now, ssh:// gets ssh errors but not ssh-ng://.

/cc @edolstra

### DIFF
--- a/src/libstore/ssh-store.cc
+++ b/src/libstore/ssh-store.cc
@@ -24,6 +24,10 @@ struct SSHStoreConfig : virtual RemoteStoreConfig
 class SSHStore : public virtual SSHStoreConfig, public virtual RemoteStore
 {
 public:
+    // Hack for getting remote build log output.
+    // Intentionally not in `SSHStoreConfig` so that it doesn't appear in
+    // the documentation
+    const Setting<int> logFD{(StoreConfig*) this, -1, "log-fd", "file descriptor to which SSH's stderr is connected"};
 
     SSHStore(const std::string & scheme, const std::string & host, const Params & params)
         : StoreConfig(params)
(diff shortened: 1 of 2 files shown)


regnat (approved):

src/libstore/ssh-store.cc
```diff
@@ -24,6 +24,10 @@ struct SSHStoreConfig : virtual RemoteStoreConfig
 class SSHStore : public virtual SSHStoreConfig, public virtual RemoteStore
 {
 public:
+    // Hack for getting remote build log output.
+    // Intentionally not in `SSHStoreConfig` so that it doesn't appear in
+    // the documentation
+    const Setting<int> logFD{(StoreConfig*) this, -1, "log-fd", "file descriptor to which SSH's stderr is connected"};
```
Ericson2314, line 30: N.B. this seemed sketch, but is actually fine because this is exactly how `LegacySSHStore` does it.

Ericson2314 (approved):

edolstra:
I think the reason why `SSHStore` doesn't have this is because it's mostly unnecessary: remote build log output is received via the daemon protocol (as `STDERR_*` messages). So this will only affect non-daemon error messages like connection issues.

Ericson2314:
https://github.com/NixOS/nix/pull/6029 will fix this using logs conveyed the way @edolstra described. When that is merged, this can be closed.

FWIW, I think that way is indeed better, because for e.g. a TcpRemoteStore (https://github.com/NixOS/nix/pull/5265) we don't easily get an additional stream "for free" anyways.

Ericson2314:
Closing as described above.

Issue #4677, opened 2021-03-30
More context in context string


guibou:
**Is your feature request related to a problem? Please describe.**

I'm gathering informations about all the dependencies of a project, such as licences, maintainers, kind (i.e. haskell deps, python deps, ...) and count of the dependencies, ...

It is straightforward to walk the nix structures and gather some information. `builtins.genericClosure` helps. However I'm missing the dependencies which are implicitly attached to string as string context.

I can recover the context using `builtins.getContext`, however it does returns a path in the nix store to the derivation, not the original attribute set, so the information I care about is lost.

For example, the expression `pkgs.runCommand "foo" {} ''${pkgs.blender}''` implicitly depends on `pkgs.blender`. However I cannot find a reference to the derivation `blender` other than using `builtins.getContext ((pkgs.runCommand "foo" {} ''${pkgs.blender}'').buildCommand)` which returns an attribute set where the `drv` path is the key: `{ "/nix/store/8vwn856hd9q6413qqnjiv3l4x7wxq31r-blender-2.91.0.drv" = { ... }; }`, unfortunately, the derivation attribute set is lost, including the `meta` information I care about.

Unfortunately, the `.drv` does not contain the `meta` information either.

**Describe the solution you'd like**

The string context could include a reference to the original derivation which was interpolated as a path?

**Describe alternatives you've considered**

- I considered rebuilding the information based on the `.drv` path, by looking in all the derivation of my projects (including `nixpkgs`), but it is slow (i.e. I need to generate the attribute set of ALL the possible outputPath) and it may not be complete, if a derivation does not exists as an explicit nix reference somewhere (e.g. `"${pkgs.blender}"` will be found when scanning `pkgs`, but `"${pkgs.f x y}"` won't, because only `f` will be found when scanning `pkgs`.)
- I can just accept that it is not complete
- Adding meta informations in .drv

**Additional context**

The issue with this approach is that it will cost memory for the new entry in the context as well as keeping a reference to the associated attribute set. I did not estimated the memory overhead introduced.

roberth:
Related
 - https://github.com/NixOS/nix/issues/7662

roberth:
Possible solution: improve output hashing so that `meta` can be written to derivations without causing rebuilds.
Traditional input addressed derivations were already in a many derivations to one output relation; this leans into that.
- https://github.com/NixOS/nix/pull/14686

PR #5036, opened 2021-07-21
let derivations expose .inputDrvs and .inputSrcs

### LINKED ISSUE #5032: derivation to expose all the fields
instantiated `derivation`s expose `.drvName` and output as calculated fields:

https://github.com/NixOS/nix/blob/099df07e1ee1663a9176384071f61a19b0f5686c/src/libexpr/primops.cc#L1206-L1227

Is it possible to expose all the 7 fields of computed derivation, for example as
```
.final.outputs
.final.inputDrvs
.final.inputSrcs
.final.platform
.final.builder
.final.args
.final.env
```
?

`.inputDrvs` `.inputSrcs` are especially interested to have at eval time to deal with the dependency graph ( `.drv`-files are not always written to disk, so eval time is the only time when the graph exists) and to fix nixpkg's features like `checkMetaRecursively`, which rely on explicit dependencies mentioned in `buildInputs`, `propagatedBuildInputs`, ... and miss the deps brought via `stdenv` or as `"${coreutils}/bin/xxx"`:

https://github.com/NixOS/nixpkgs/blob/40bb731ec71bdd7e85ae8c00208d90042e98ce21/pkgs/stdenv/generic/make-derivation.nix#L137-L138

vobloeb:
example on how derivations could expose .inputDrvs and .inputSrcs

partially fixes #5032

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -519,6 +519,18 @@ void prim_valueSize(EvalState & state, const Pos & pos, Value * * args, Value & 
  *************************************************************/
 
 
+/* Add a wrapper around the derivation primop that computes the
+   `drvPath' and `outPath' attributes lazily. */
+static void prim_derivation(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    Value fun;
+    state.evalFile(/*settings.nixDataDir + "/nix/corepkgs/derivation.nix"*/state.sDerivationNix, fun);
+    state.forceFunction(fun, pos);
+    mkApp(v, fun, *args[0]);
+    state.forceAttrs(v, pos);
+}
+
+
 /* Construct (as a unobservable side effect) a Nix derivation
    expression that performs the derivation described by the argument
    set.  Returns the original set extended with the following
@@ -526,7 +538,7 @@ void prim_valueSize(EvalState & state, const Pos & pos, Value * * args, Value & 
    derivation; `drvPath' containing the path of the Nix expression;
    and `type' set to `derivation' to indicate that this is a
    derivation. */
-static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * * args, Value & v)
+static void prim_derivationStrict2(EvalState & state, const Pos & pos, Value * * args, Value & v)
 {
     state.forceAttrs(*args[0], pos);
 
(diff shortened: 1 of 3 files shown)


src/libexpr/primops.cc
```diff
@@ -1387,6 +1435,28 @@ static void prim_mapAttrs(EvalState & state, const Pos & pos, Value * * args, Va
 }
 
 
+static void prim_genAttrs(EvalState & state, const Pos & pos, Value * * args, Value & v)
```
kvtb, line 1438: it is just `lib.genAttr` implemented in C++

minijackson:
I'm interested in this PR, what can we do to push it forward?

Ericson2314:
Derivations should by default be treated as block boxes; @roberth has done good work trying to make Nixpkgs hide this stuff *more* not less.

At the same time though, I see that it is easier for Nix code to throw information away than recover it, so we might consider making some primop that exposes the entire derivation.

minijackson:
@Ericson2314 to be fair, I'm not interested in the derivation internals per se, but only want the derivation inputs at eval time. This information is already exposed by `nix path-info -r --derivation`, but cannot be used easily in Nix code.

This leads for example the report bundler to [manually scan the attributes](https://github.com/NixOS/bundlers/blob/master/report/default.nix#L142-L148) for derivation, which is error prone (current implementation misses strings with context), despite the info being available in the Nix evaluator.

I agree that exposing the internals is unsafe / constraining. What would you think of a primop that only exposes build inputs?

src/libexpr/primops.cc
```diff
@@ -1387,6 +1436,31 @@ static void prim_mapAttrs(EvalState & state, const Pos & pos, Value * * args, Va
 }
 
 
+/* Given a set of attribute names, return the set of the corresponding
+   attributes from the given set. Former `lib.genAttrs'
```
roberth, line 1440: Let's not get ahead of ourselves ;)

```suggestion
   attributes from the given set. Implements Nixpkgs `lib.genAttrs'
```

-----


`genAttrs` might be a good candidate for a primop regardless of your original goal. As currently implemented it allocates an unnecessary list, and 2-attr attrsets for each attribute name. I don't know what would be the impact of this optimization though. (TODO: register it with a documentation string like most of the other primops, add tests)

roberth: I would also argue that `builtins` is explicitly not a replacement for `lib`, because the builtins can't be changed. The analogy between kernel syscalls and a standard library (like libc or go) is fitting.

src/libexpr/primops.cc
```diff
@@ -519,14 +519,27 @@ void prim_valueSize(EvalState & state, const Pos & pos, Value * * args, Value &
...
+/* Add a wrapper around the derivation primop that computes the
+   `drvPath', `outPath', `inputSrcs' and `inputDrvs' attributes lazily. */
+static void prim_derivation(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    Value fun;
+    state.evalFile(state.sDerivationNix, fun);
+    state.forceFunction(fun, pos);
+    mkApp(v, fun, *args[0]);
+    state.forceAttrs(v, pos);
+}
```
roberth, line 531: Although this wrapper is a little silly from a mechanical perspective, if we register it with a docstring that would solve a little UX problem that we have. (Could be an independent PR as well)

roberth (commented):
I probably shouldn't have reviewed this before asking about the use case, although there's some interesting things in here that I think we should consider regardless of the original (sub)goal.

I suspect that your problem could be better solved by improving the string context
 - https://github.com/NixOS/nix/issues/4677

... as usually the dependency information can already be obtained in a derivation using [exportReferencesGraph](https://nixos.org/manual/nix/stable/language/advanced-attributes.html?highlight=closure#adv-attr-exportReferencesGraph), and so usually the use case for doing it outside the derivation is to access things that aren't just paths.

With the string context, you could add attach the information you might need, such as package `meta` attrs, or the whole package attrset for that matter, directly to the string context (with a bit of help from an updated Nixpkgs of course; it doesn't want to tell `derivation` anything yet that doesn't belong in a derivation).

src/libexpr/primops.cc
```diff
@@ -764,13 +777,49 @@ static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * *
...
             i.second.path, {"!" + i.first + "!" + drvPath});
     }
+
+    {
+        Value * inputSrcsVal = state.allocAttr(v, state.symbols.create("inputSrcs"));
+        state.mkList(*inputSrcsVal, drv.inputSrcs.size());
+        int i=0;
+        for (auto & p : drv.inputSrcs) {
+            inputSrcsVal->listElems()[i] = state.allocValue();
+            mkString(*(inputSrcsVal->listElems()[i]), p, /*context=*/{"=" + p});
```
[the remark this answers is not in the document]
Ericson2314: Oh don't worry, this would be a type error once rebased, anyways :)

tomberek:

PR #5037, opened 2021-07-21
Avoid global counters

Pamplemousse:
1. I wondered if `nrThunks` could not be incremented in `EvalState::mkThunk_`, given that the stats are for the `EvalState` anyway (VS having it in the inlined `mkThunk` as of now, which is called from other places)
2. `nrValuesFreed` is also a global, but I don't see it used anywhere (anymore); Could it be removed?
Seems to have been introduced by https://github.com/NixOS/nix/commit/30964103dc31c32ccb69912fdd73474cd65447e9 (cc @edolstra).

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -854,39 +854,37 @@ Value * Expr::maybeThunk(EvalState & state, Env & env)
 }
 
 
-unsigned long nrAvoided = 0;
-
 Value * ExprVar::maybeThunk(EvalState & state, Env & env)
 {
     Value * v = state.lookupVar(&env, *this, true);
     /* The value might not be initialised in the environment yet.
        In that case, ignore it. */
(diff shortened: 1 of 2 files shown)


edolstra:
Looks good. There may be a slight performance penalty though, since this requires a pointer indirection through `state`. It probably doesn't matter for dynamically linked Nix since I think PIC requires an indirection for globals anyway, but it might for static linking.

Mic92:
> Looks good. There may be a slight performance penalty though, since this requires a pointer indirection through `state`. It probably doesn't matter for dynamically linked Nix since I think PIC requires an indirection for globals anyway, but it might for static linking.

We likely get PIE even for static linking because of security: https://github.com/NixOS/nixpkgs/pull/124187#issuecomment-851057077. My hypothesis is also that the state struct is in the cache anyway because of the other fields in there.

PR #5043, opened 2021-07-24
libutil: use uniform initialization in _deletePath

alyssais:
Otherwise I get a compiler error when building for NetBSD:

```
src/libutil/util.cc: In function 'void nix::_deletePath(const Path&, uint64_t&)':
src/libutil/util.cc:438:17: error: base operand of '->' is not a pointer
  438 |     AutoCloseFD dirfd(open(dir.c_str(), O_RDONLY));
      |                 ^~~~~
src/libutil/util.cc:439:10: error: 'dirfd' was not declared in this scope
  439 |     if (!dirfd) {
      |          ^~~~~
src/libutil/util.cc:444:17: error: 'dirfd' was not declared in this scope
  444 |     _deletePath(dirfd.get(), path, bytesFreed);
      |                 ^~~~~
```

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -435,7 +435,7 @@ static void _deletePath(const Path & path, uint64_t & bytesFreed)
     if (dir == "")
         dir = "/";
 
-    AutoCloseFD dirfd(open(dir.c_str(), O_RDONLY));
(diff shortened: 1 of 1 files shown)


Ericson2314 (approved):
I like this style better anyways.

PR #5048, opened 2021-07-26
--eval-store and faster closure copying

### LINKED ISSUE #5025: Separate stores for evaluation, building and storing the result
For several use cases it would be useful to distinguish between the stores used to store .drv files, to build derivations, and to store the resulting closure. 

For example:

* `nix path-info --store https://cache.nixos.org nixpkgs#hello` doesn't work because the evaluation of `nixpkgs#hello` tries to create .drv files in `cache.nixos.org`.
* For NixOps, we want to do evaluation on the local machine, do builds on some arbitrary machines (could be the local machine, the destination machine, or some other builder) and store the result in the destination machine.
* If `/nix/store` is a high-latency remote store (as with flox), then copying lots of little .drv files to the remote is slow. So it's better to evaluate locally and only copy .drv closures to the remote as needed and in a single round-trip.

### LINKED ISSUE #5026: Optimize closure copying for high-latency remotes
Currently copying closures with lots of small files (like .drv closures or NixOS system closures) is slow because every store path is copied separately and requires a round-trip to the remote. Sending the entire set of missing paths in a single call (like with the old `nix-store --export`) is much faster.

edolstra:
When the Nix daemon is running remotely and is accessed via a high-latency connection, an operation like `nix build nixpkgs#hello` can be very slow because the client first needs to send a lot of .drv files one-by-one to the remote, each requiring a round trip. To make this faster, this PR does the following:

* It adds an option `--eval-store` that allows the evaluation to be done in a local Nix store. This is also useful for store that can't build, like binary caches. For instance, you can now do: 
    ```
    # nix path-info --json --store https://cache.nixos.org --eval-store auto nixpkgs#hello
    ```

* It adds a new worker protocol interface to add multiple store paths in the same call. On a link with a 50 ms delay, this speeds up the copying of the `hello` .drv closure from 57.8s to 0.7s. Note that this disables parallel copying (the use of `processGraph()` with a thread pool in `copyPaths()` is commented out).

Fixes #5026.
Partially fixes #5025.

### DIFF
--- a/src/libstore/path-info.hh
+++ b/src/libstore/path-info.hh
@@ -105,6 +105,11 @@ struct ValidPathInfo
     ValidPathInfo(const StorePath & path, Hash narHash) : path(path), narHash(narHash) { };
 
     virtual ~ValidPathInfo() { }
+
+    static ValidPathInfo read(Source & source, const Store & store, unsigned int format);
+    static ValidPathInfo read(Source & source, const Store & store, unsigned int format, StorePath && path);
+
+    void write(Sink & sink, const Store & store, unsigned int format, bool includePath = true) const;
 };
 
 typedef std::map<StorePath, ValidPathInfo> ValidPathInfos;
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -270,7 +270,7 @@ connected:
 
         {
             Activity act(*logger, lvlTalkative, actUnknown, fmt("copying dependencies to '%s'", storeUri));
-            copyPaths(store, ref<Store>(sshStore), store->parseStorePathSet(inputs), NoRepair, NoCheckSigs, substitute);
+            copyPaths(*store, *sshStore, store->parseStorePathSet(inputs), NoRepair, NoCheckSigs, substitute);
         }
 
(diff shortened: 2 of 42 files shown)


Mic92:
How can I enable `--eval-store auto` globally?

edolstra:
You can't since it's not a config option and I'm not sure if it makes sense to turn it into one, since it could have unexpected interactions with scripts that use `--store`.

src/libstore/path-info.hh
```diff
@@ -105,6 +105,11 @@ struct ValidPathInfo
     ValidPathInfo(const StorePath & path, Hash narHash) : path(path), narHash(narHash) { };
 
     virtual ~ValidPathInfo() { }
+
+    static ValidPathInfo read(Source & source, const Store & store, unsigned int format);
+    static ValidPathInfo read(Source & source, const Store & store, unsigned int format, StorePath && path);
+
+    void write(Sink & sink, const Store & store, unsigned int format, bool includePath = true) const;
```
Ericson2314, line 112: This shouldn't be part of the core data type, but instead defined like and with the other serializers in `src/libstore/worker-protocol.hh`.

Issue #4337, opened 2020-12-09
/nix/store is writable in a nix-shell started by root


zhaofengli:
**Describe the bug**

`/nix/store` is read-write in a nix-shell started by root.

**Steps To Reproduce**

```
$ mount | grep store
rpool/nix on /nix/store type zfs (ro,relatime,xattr,posixacl) <-- no active nix-shell: ok
$ nix-shell

[nix-shell:~]$ mount | grep store
rpool/nix on /nix/store type zfs (ro,relatime,xattr,posixacl) <-- non-root nix-shell: ok

[nix-shell:~]$ exit
$ sudo nix-shell -p bash

[nix-shell:/home/zhaofeng]# mount | grep store
rpool/nix on /nix/store type zfs (rw,relatime,xattr,posixacl) <-- root nix-shell: why?
```

**Expected behavior**

`/nix/store` should stay read-only.

**`nix-env --version` output**

The machine above is running unstable with `nix-env (Nix) 2.4pre20201201_5a6ddb3`.
Also reproduced on a 20.09 machine with `nix-env (Nix) 2.3.9`.

**Additional context**

The behavior is not related to ZFS, and I was able to reproduce it on a machine with single-partition ext4.

_Edit: Somehow I was under the impression that this would affect the mount namespace outside the nix-shell, but this is not the case with more testing. I have updated the description._

regnat:
Note that it only happens because nix bypasses the daemon by default when running as root (see #4263). A workaround is to set `NIX_REMOTE=daemon` to force it to use the daemon − in which case the store will stay read-only:

```console
$ sudo nix-shell -p mount --run mount | grep store
znvme/nix on /nix/store type zfs (rw,relatime,xattr,noacl)
$ sudo NIX_REMOTE=daemon nix-shell -p mount --run mount | grep store
znvme/nix on /nix/store type zfs (ro,relatime,xattr,noacl)
```

robinbb:
This is clearly not desirable, as it is astonishing. The root user should be able to launch a Nix shell without unknowingly winding up with a writable Nix store.

edolstra:
We used to have code for restoring the original mount namespace before executing the shell (a0ef21262f4d5652bfb65cfacaec01d89c475a93), but I had to revert it (01d07b1e92c298f729a73705907b2987da9a4d0c).

zhaofengli:
[...]

PR #5057, opened 2021-07-27
nix develop: Support chroot stores

### LINKED ISSUE #5024: nix develop: Support chroot stores
`nix develop --store /path nixpkgs#hello` currently succeeds, but the subshell isn't run in a mount namespace (like `nix shell`) so it doesn't actually work.

edolstra:
Fixes #5024.

### DIFF
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -1,3 +1,4 @@
+#include "run.hh"
 #include "command.hh"
 #include "common-args.hh"
 #include "shared.hh"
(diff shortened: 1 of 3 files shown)


src/nix/run.cc
```diff
@@ -20,45 +21,43 @@ using namespace nix;
...
 
-        restoreProcessContext();
+    /* If this is a diverted store (i.e. its "logical" location
+       (typically /nix/store) differs from its "physical" location
+       (e.g. /home/eelco/nix/store), then run the command in a
+       chroot. For non-root users, this requires running it in new
+       mount and user namespaces. Unfortunately,
+       unshare(CLONE_NEWUSER) doesn't work in a multithreaded program
+       (which "nix" is), so we exec() a single-threaded helper program
+       (chrootHelper() below) to do the work. */
```
regnat, line 41: This sounds like it would also fix #4337 for `nix develop`. Is it the case?

regnat (approved):
That’s great :)

src/nix/run.cc
edolstra: I think so.

PR #5058, opened 2021-07-27
Allow running all the tests with the daemon 

### LINKED ISSUE #4824: Force the use of the daemon in all the tests

Nix frequently has some bugs living in the communication between the client and the daemon. These are generally not spotted in the testsuite because most tests run in single-user mode.

As a continuation of https://github.com/NixOS/nix/pull/4239, also run the tests in a mode where all the tests actually use the Nix daemon (if possible)

**Additional context**

Related (and somewhat prompted by) https://github.com/NixOS/nix/pull/4801#issuecomment-840260498

regnat:
When `NIX_DAEMON_PACKAGE` is set, make all the tests use the Nix daemon.
That way we can test every piece of Nix functionality both with and without the daemon.

Tests for which using the daemon isn’t possible or doesn’t make sense can selectively be disabled with `needLocalStore`.
Similarely, tests can enforce a constraint on a new-enough daemon by using `requireDaemonNewerThan 2.4pre_20210714`

(There’s a few tests that break with the daemon for bad reasons. I didn’t fix these, but they are marked with `buggyNeedLocalStore`)

Fix #4824

### DIFF
--- a/tests/binary-cache.sh
+++ b/tests/binary-cache.sh
@@ -1,5 +1,7 @@
 source common.sh
 
+needLocalStore "“--no-require-sigs” can’t be used with the daemon"
+
 # We can produce drvs directly into the binary cache
 clearStore
 clearCacheCache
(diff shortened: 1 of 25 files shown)


Ericson2314 (approved):
This is great. Simply tracking what fails is super valuable even.

PR #5059, opened 2021-07-27
Add a github action step to run the tests against current master

thufschmitt:
Make github actions also run the testsuite using a `nix-daemon` executable that comes from the master branch.

This allows ensuring that every commit on master is backwards-compatible with the previous one.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,10 @@ jobs:
     - run: nix-build -A checks.$(nix-instantiate --eval -E '(builtins.currentSystem)')
     - run: |
         NIX_BUILD=$(nix-build -A defaultPackage.$(nix-instantiate --eval -E '(builtins.currentSystem)'))/bin/nix-build
-        $NIX_BUILD --option experimental-features 'flakes' --impure --expr '(builtins.getFlake (builtins.toPath ./.)).lib.testAgainst.${builtins.currentSystem} (builtins.getFlake "github:nixos/nix/master").defaultPackage.${builtins.currentSystem}'
+        $NIX_BUILD \
+          --option experimental-features 'flakes' \
+          --option access-tokens github.com=${{ secrets.GITHUB_TOKEN }} \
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -21,6 +21,9 @@ jobs:
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
     - run: nix-build -A checks.$(nix-instantiate --eval -E '(builtins.currentSystem)')
+    - run: |
+      NIX_BUILD=$(nix-build -A defaultPackage.$(nix-instantiate --eval -E '(builtins.currentSystem)'))/bin/nix-build
+      $NIX_BUILD --option experimental-features 'flakes' --impure --expr 'builtins.getFlake (builtins.toPath ./.).lib.testAgainst.x86_64-linux (builtins.getFlake (builtins.toPath ./.)).defaultPackage.x86_64-linux'
```
edolstra, line 26: Why is `--impure` needed?

regnat: Both flake inputs (`./.` and `github:nixos/nix`) are mutable refs, so they aren’t allowed in pure mode

tomberek, line 26: needs extra parens?
```suggestion
      $NIX_BUILD --option experimental-features 'flakes' --impure --expr '(builtins.getFlake (builtins.toPath ./.)).lib.testAgainst.x86_64-linux (builtins.getFlake (builtins.toPath ./.)).defaultPackage.x86_64-linux'
```

regnat:
I’ve rewritten the test to make sure that we use exactly the same derivation that has been built previously (both for the current Nix daemon and master), so that we can just fetch it from the cache rather than re-building everything again. This brings the extra CI time to a bit less than 5mins (so not trivial, but acceptable)

Ericson2314:
Oh this is tricky, but the goals is indeed a good one.

tomberek:
stale and OBE, reopen if needed

Issue #5061, opened 2021-07-28
SCP-style URLs in `builtins.fetchGit` still broken


considerate:
**Describe the bug**

Evaluating builtins.fetchGit with a git SCP-like URL causes the error:
`error: 'file://git@github.com:NixOS/nix.git' is not a valid URL`

This was originally reported in #3503 and fixed in https://github.com/NixOS/nix/pull/4978 but the version 02dd6bb610e55a009cd7a4c83639698d3a7acaa2 of nix which contains this PR still results in the broken behavior.

**Steps To Reproduce**

create test.nix with this content:
```
builtins.fetchGit {
  url = "git@github.com:NixOS/nix.git";
  ref = "master";
  rev = "02dd6bb610e55a009cd7a4c83639698d3a7acaa2";
}
```

Run `nix-build test.nix`

**Expected behavior**

The nix repo should be downloaded and a symlink should be created.

**`nix-env --version` output**

```
nix-env (Nix) 2.4pre20210707_02dd6bb
```

regnat:
Indeed, it looks like #4978 only fixed `builtins.fetchGit "git@…"` and not `builtins.fetchGit { url = "git@…"; }`:

```
nix-repl> builtins.fetchGit "git@github.com:NixOS/nix.git"
{ lastModified = 1627468261; lastModifiedDate = "20210728103101"; narHash = "sha256-rpPXcGhgCUo9fCF6aqez8ZaU4vlhvwM3NaDhSO9XWLg="; outPath = "/nix/store/fgnma6ppv68lchgim8wp1l8mlczjdmzm-source"; rev = "7816ef6c5123ef0f11b8b86439dd6b934c1e5426"; revCount = 10533; shortRev = "7816ef6"; submodules = false; }

nix-repl> builtins.fetchGit { url = "git@github.com:NixOS/nix.git"; }
error: 'file://git@github.com:NixOS/nix.git' is not a valid URL
```

considerate:
I think it might be due to this call: https://github.com/NixOS/nix/blob/ec2c6bd47041bda1f30e66ee48de662ddcd07378/src/libexpr/primops/fetchTree.cc#L106

`addURI` calls `fixURI` regardless of the type of the tree.

considerate:
On a somewhat unrelated note, this case seems to be redundant/inaccessible:

[...]

Issue #3503, opened 2020-04-16
URL parsing when using fetchGit does not accept the SSH URL that Github suggests anymore


rbvermaa:
See the following example.

With 2.3.4:

```
nix-repl> builtins.fetchGit { url = "git@github.com:RelationalAI/myrepo.git"; }
{ outPath = "/nix/store/ls6m...-source"; rev = "..."; revCount = ...; shortRev = "..."; }
```

With latest master:

```
nix-repl> builtins.fetchGit { url = "git@github.com:RelationalAI/myrepo.git"; }
error: 'file://git@github.com:RelationalAI/myrepo.git' is not a valid URL
```

rbvermaa:
Btw, using `ssh://git@github.com/RelationalAI/myrepo.git` does work both in 2.3.4 and latest master.

bjornfor:
Support for `git@...` URLs is even documented:

```
$ nix repl
warning: unknown setting 'extra-sandbox-paths'
Welcome to Nix version 2.4pre20201102_550e11f. Type :? for help.

nix-repl> :doc fetchGit
[...]
    Here are some examples of how to use fetchGit.

  - To fetch a private repository over SSH:

          builtins.fetchGit {
            url = "git@github.com:my-secret/repository.git";
            ref = "master";
            rev = "adab8b916a45068c044658c4158d81878f9ed1c3";
          }
[...]
nix-repl> builtins.fetchGit {
                      url = "git@github.com:my-secret/repository.git";
                      ref = "master";
                      rev = "adab8b916a45068c044658c4158d81878f9ed1c3";
                    }
error: --- BadURL ------------------------------------------------------------------------------------ nix
'file://git@github.com:my-secret/repository.git' is not a valid URL
```

I'm trying to use Nix at $dayjob and currently have to work around this :-(

bjornfor:
I have looked at the code and while removing the `file://` prefix is doable (for me), I see that Nix is really wants that string to be an _URL_ (it calls `parseURL()` from more than one place). But what we want here is the scp semi-URL format `[user@]host:path`.

It feels a bit ugly to make `parseURL()` accept the scp format, but I don't know what else to do. (Nix also wants to check the inputs for the allowed-uris functionality, so we cannot simply _drop_ checking the input string either.)

bjornfor:
@edolstra Can you add this issue to the [nix-2.4 milestone](https://github.com/NixOS/nix/milestone/11)? (I'd be sad if 2.4 gets released with this regression.)

PR #5062, opened 2021-07-28
Support SCP-like URLs in builtins.fetchGit attrs

considerate:
This extends https://github.com/NixOS/nix/pull/4978 with
supporting the SCP-like urls in expressions like

```nix
builtins.fetchGit {
  url = "git@github.com:NixOS/nix.git";
  ref = "master";
}
```


VerTiGoEtrex:
Thanks for writing up this fix — hopefully merged soon. This is a pretty annoying bug.

considerate:
Thank you @VerTiGoEtrex. The main credits lies with @regnat though, since he's the one who wrote the original fix for SCP-like URLs. I just applied the same fix to URLs in attribute sets.

considerate:
@edolstra @regnat Is this something that should be prioritized for the nix 2.4 release? The bug https://github.com/NixOS/nix/issues/3503 was listed as critical and since https://github.com/NixOS/nix/issues/5061 is about the same functionality I think that this ought to be granted the same level of importance.

edolstra:
#3503 was critical because it was a regression. But `fetchTree` is new so it doesn't necessarily have to support this syntax.

regnat:
> #3503 was critical because it was a regression. But `fetchTree` is new so it doesn't necessarily have to support this syntax.

Note that this isn’t just about `fetchTree`, but also `fetchGit { url = …; }`, which already existed in 2.3

considerate:
> #3503 was critical because it was a regression. But `fetchTree` is new so it doesn't necessarily have to support this syntax.

I'm not sure if I agree that this is not a regression. This bug affects the following syntax:
```
builtins.fetchGit {
  url = "git@github.com:NixOS/nix.git";
  ref = "master";
  rev = "02dd6bb610e55a009cd7a4c83639698d3a7acaa2";
}
```

The above works fine on nix-2.3 but not on nix-2.4-pre.

regnat:
It was, but my fix was incomplete and only fixed `fetchGit "git@foo"`

considerate:
@edolstra @regnat Thank you for prioritizing this issue!

PR #5071, opened 2021-07-30
Don’t send the experimental-features to the daemon

### LINKED ISSUE #5017: The Nix daemon crashes when started without `--experimental-features ca-derivations`, but a client tries to force it
**Describe the bug**

(privileged) clients can enable some experimental features on the daemon they talk to. This is generally fine, except for the `ca-derivations` which will cause the daemon to crash with something like `src/libstore/sqlite.cc:92: nix::SQLiteStmt::Use::Use(nix::SQLiteStmt&): Assertion `stmt.stmt' failed`.

**Steps To Reproduce**

- Start the Nix daemon without the `ca-derivations` experimental feature
- Run a (privileged) build with `--extra-experimental-features ca-derivations`

**Expected behavior**

Ideally this should transparently work as expected.
This might be a bit tricky because of the DB schema constraints, but we should at least yield a proper error in that case.

**Additional context**

The underlying issue is that enabling `ca-derivations` changes the underlying db schema (which can only happen at the begining of the process when we open the DB).
Then some precompiled SQL queries in `LocalStore` depend on the ca-specific schema, so they are left undefined if the experimental-feature isn’t set ([here](https://github.com/NixOS/nix/blob/master/src/libstore/local-store.cc#L342-L385)).
In a “normal” setting, this is fine because we’ll never use these as `ca-derivations` isn’t set. But since the client can enable the feature after-the-fact, we end-up using it anyways, leading to the assertion failure (which is mostly a glorified null pointer exception)

regnat:
The experimental features are, well, experimental, and shouldn’t be
carelessly and transparently enabled.
Besides, some (`ca-derivations` at least) need to be enabled at startup
in order to work properly.
So it’s better to just require that daemon be started with the right
`experimental-features` option.

Fix #5017

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -227,6 +227,12 @@ struct ClientSettings
             try {
                 if (name == "ssh-auth-sock") // obsolete
                     ;
+                else if (name == settings.experimentalFeatures.name) {
+                    // We don’t want to forward the experimental features to
+                    // the daemon, as that could cause some pretty weird stuff
+                    if (tokenizeString<Strings>(value) != settings.experimentalFeatures.get())
+                        debug("Ignoring the client-specified experimental features");
+                }
                 else if (trusted
                     || name == settings.buildTimeout.name
                     || name == "connect-timeout"
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -222,6 +222,7 @@ void RemoteStore::setOptions(Connection & conn)
         overrides.erase(settings.buildCores.name);
         overrides.erase(settings.useSubstitutes.name);
         overrides.erase(loggerSettings.showTrace.name);
+        overrides.erase(settings.experimentalFeatures.name);
         conn.to << overrides.size();
         for (auto & i : overrides)
             conn.to << i.first << i.second.value;


edolstra:
For security it's better to ignore this option in the daemon, rather than not send it in the client.

Ericson2314:
I think it's better still that the daemon fail operations with experimental features it didn't opt into, and so the client should send them to confirm access

Ericson2314:
I'd the setting doesn't make sense daemon side it shouldn't be sent, but if it is required it should be sent to get better error messages

Ericson2314:
I think an underlying issue is that the single global settings is A layer violation since it had libnixstore and higher level concerns. I would like to see that broken up. Then we wouldn't send eval-layer stuff anyways.

Experimental features maybe should also be broken up by layer.

edolstra:
@Ericson2314 I agree it would be good to only send store-level settings, but having multiple experimental features settings would be annoying for users.

Ericson2314:
@regnat if we have enums of features rather than arbitrary strings we can categorize automatically.

regnat:
> For security it's better to ignore this option in the daemon, rather than not send it in the client

Yes, that’s fair. I’ve changed that in b2bce915adc6b1290494388b8c73b89c253629fb (also logging if the client-specified experimental-features differ from the daemon one to prevent some impossible-to-debug issues).

edolstra:
@regnat Looks like a test is now failing: https://github.com/NixOS/nix/pull/5066/checks?check_run_id=3477082632

```
    error: experimental Nix feature 'recursive-nix' is disabled; use '--experimental-features recursive-nix' to override
```

I suppose we need to enable the recursive-nix feature in the daemon?

PR #5075, opened 2021-07-30
libutil: initialize the base64 decode array only once

andir:
Previously, despite having a boolean that tracked initialization, the
decode characters have been "calculated" every single time a base64
string was being decoded.

With this change we only initialize the decode array once in a
thread-safe manner.

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -6,14 +6,15 @@
 
 #include <cctype>
 #include <cerrno>
+#include <climits>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <climits>
+#include <future>
 #include <iostream>
+#include <mutex>
 #include <sstream>
 #include <thread>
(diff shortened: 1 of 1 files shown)


regnat (approved):

src/libutil/util.cc
```diff
@@ -1472,15 +1473,12 @@ string base64Encode(std::string_view s)
...
 string base64Decode(std::string_view s)
 {
-    bool init = false;
-    char decode[256];
-    if (!init) {
-        // FIXME: not thread-safe.
-        memset(decode, -1, sizeof(decode));
+    static std::once_flag flag;
+    std::call_once(flag, [](){
+        base64DecodeChars = { (char)-1 };
```
kvtb, line 1478: If you expect that fills `base64DecodeChars` with `-1`'s, that is wrong

the code effectively does `base64DecodeChars[0] = -1`

PR #5088, opened 2021-08-03
Evaluate nix-shell -i args relative to script

### LINKED ISSUE #4232: nix-shell shebang does not interpret -E relative to script dir
**Describe the bug**

When using `nix-shell` in a shebang, it interprets filenames given to it relative to the script dir, but it interprets the `-E` flag relative to the cwd. This is rather problematic. I can't think of any reason to want the `-E` flag interpreted relative to the current dir and doing so means I'm forced to write a deps.nix file to disk next to my script if I need to import some other file at a known path.

More generally, I should be able to replace `#!nix-shell deps.nix -i bash` with `#!nix-shell -E 'import ./deps.nix' -i bash`.

**Steps To Reproduce**

Write a script that looks something like

```bash
#!/usr/bin/env nix-shell
#!nix-shell -i bash -E "with import ./deps.nix {}; mkShell { buildInputs = [bash]; }"
echo it works
```

**Expected behavior**

It shouldn't matter what my cwd is when I execute this, it should behave the same and import the `deps.nix` file that lives right next to the script.

**Actual behavior**

It tries to import `./deps.nix` from whatever my cwd is.

nix-env (Nix) 2.3.7

matthewbauer:
When writing a shebang script, you expect your path to be relative to
the script, not the cwd. We previously handled this correctly for
relative file paths, but not for expressions.

This handles both -p & -E args. My understanding is this should be
what we want in any cases I can think of - people run scripts from
many different working directories. @edolstra is there any reason to
handle -p args differently in this case?

Fixes #4232

### DIFF
--- a/doc/manual/rl-next/shebang-relative.md
+++ b/doc/manual/rl-next/shebang-relative.md
@@ -1,0 +1,8 @@
+synopsis: ensure nix-shell shebang uses relative path
+prs: #5088
+description: {
+
+`nix-shell` shebangs use the script file's relative location to resolve relative paths to files passed as command line arguments, but expression arguments were still evaluated using the current working directory as a base path.
+The new behavior is that evalutations are performed relative to the script.
+
+}
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -298,7 +298,9 @@ static void main_nix_build(int argc, char * * argv)
     else
         for (auto i : left) {
             if (fromArgs)
-                exprs.push_back(state->parseExprFromString(std::move(i), state->rootPath(CanonPath::fromCwd())));
+                exprs.push_back(state->parseExprFromString(
+                    std::move(i),
+                    state->rootPath(CanonPath::fromCwd(inShebang ? dirOf(script) : "."))));
             else {
                 auto absolute = i;
                 try {
@@ -311,7 +313,7 @@ static void main_nix_build(int argc, char * * argv)
                     /* If we're in a #! script, interpret filenames
                        relative to the script. */
                     exprs.push_back(state->parseExprFromFile(resolveExprPath(state->checkSourcePath(lookupFileArg(*state,
-                                        inShebang && !packages ? absPath(i, absPath(dirOf(script))) : i)))));
+                                        inShebang ? absPath(i, absPath(dirOf(script))) : i)))));
             }
         }
 
--- a/tests/functional/nix-shell.sh
+++ b/tests/functional/nix-shell.sh
@@ -59,6 +59,16 @@ chmod a+rx $TEST_ROOT/shell.shebang.sh
 output=$($TEST_ROOT/shell.shebang.sh abc def)
 [ "$output" = "foo bar abc def" ]
 
+# Test nix-shell shebang mode with an alternate working directory
+sed -e "s|@ENV_PROG@|$(type -P env)|" shell.shebang.expr > $TEST_ROOT/shell.shebang.expr
+chmod a+rx $TEST_ROOT/shell.shebang.expr
+# Should fail due to expressions using relative path
+! $TEST_ROOT/shell.shebang.expr bar
+cp shell.nix config.nix $TEST_ROOT
+# Should succeed
+output=$($TEST_ROOT/shell.shebang.expr bar)
+[ "$output" = '-e load(ARGV.shift) -- '"$TEST_ROOT"'/shell.shebang.expr bar' ]
+
 # Test nix-shell shebang mode again with metacharacters in the filename.
 # First word of filename is chosen to not match any file in the test root.
 sed -e "s|@ENV_PROG@|$(type -P env)|" shell.shebang.sh > $TEST_ROOT/spaced\ \\\'\"shell.shebang.sh


arobertn:
What about the `-I` arg (also mentioned in #4232)? E.g. in case:

```
#!/usr/bin/env nix-shell
#! nix-shell -I nixpkgs=../nix/nixpkgs.nix --pure -p <pkgs...>
```

Or is there a preferred alternative way to set the search path?

lf-:
Can this get reviewed? This still is an issue in Nix 2.9.1.

tomberek (changes requested):
This is a net-positive change. Please add release notes as this is a behavior change.

Ericson2314:
Fixed conflicts, but this also needs tests.

fricklerhandwerk:
Please un-draft when ready, this is easier for maintainers to keep overview.

tomberek:
@matthewbauer I added tests and notes, let me know if thit is what you intended
(also pinging @lf- for visibility )

tomberek:
The worry is someone may have started to depend on these being relative to the call-site. Unfortunate, but might need a flag to control the behavior for the legacy CLI, "--relative"?

The new nix-command should not have this design oddity.

lf-:
> The worry is someone may have started to depend on these being relative to the call-site. Unfortunate, but might need a flag to control the behavior for the legacy CLI, "--relative"?

I think this is improbable and a bit of a bike shed. The workarounds (haskell plus bash polyglot! horrible) that people have had to apply would not be broken by this, as they start nix-shell in the same directory every time. I cannot see any conceivable reason someone would want the old behaviour: somehow you would be evaluating a script with respect to the current directory's default.nix?

That feels like a use case so contrived I doubt anyone has actually tried it, compared to the 99% use case that doesn't work today and I don't believe the new CLI supports at all.

fricklerhandwerk:
Strongly in favor of merging and deferring backwards compatibility hacks to the improbable case of a spacebar heating outcry.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -0,0 +1,8 @@
+# Release X.Y (202?-??-??)
+
+- Fixed a bug where `nix-env --query` ignored `--drv-path` when `--json` was set.
+
+- Introduced the store [`mounted-ssh-ng://`](@docroot@/command-ref/new-cli/nix3-help-stores.md).
+  This store allows full access to a Nix store on a remote machine and additionally requires that the store be mounted in the local filesystem.
+
+- [`nix-shell`](../command-ref/nix-shell.md) shebangs now interpret "-E" with relative paths
```
fricklerhandwerk, line 8:
```suggestion
- [`nix-shell`](../command-ref/nix-shell.md) shebangs now interpret relative paths in expressions passed to `-E` with the script file's location as the base directory rather than the caller's current working directory.
```

doc/manual/rl-next/shebang-relative.md
```diff
@@ -0,0 +1,7 @@
+synopsis: ensure nix-shell shebang uses relative path
+prs: #5088
+description: {
+
+`nix-shell` shebangs use relative paths for files, but expressions were still evaluated using the current working directory. The new behavior is that evalutations are performed relative to the script.
```
fricklerhandwerk, line 5:
```suggestion
`nix-shell` shebangs use the script file's location to resolve relative paths to files passed as command line arguments, but expression arguments were evaluated using the current working directory as a base path.
The new behavior is that Nix language expressions passed to `nix-shell` when used in a `#!` context are now evaluated relative to the script's file system location.
```

fricklerhandwerk (changes requested):
Some clarifications, because I had a hard time untangling what the release notes were trying to convey.

@roberth are we holding it correctly? Seems weird that there are two places for release notes now.

fricklerhandwerk (approved):

doc/manual/src/release-notes/rl-next.md
roberth: > Seems weird that there are two places for release notes now.

Just follow https://github.com/NixOS/nix/blob/master/doc/manual/src/contributing/hacking.md#add-a-release-note
`rl-next.md` should not be committed and has been added to `.gitignore`.
Please remove `rl-next.md` from the commits.


For context, we changed the way we write release notes today, but I believe the above is all you need to know.
 - #9393

fricklerhandwerk: @matthewbauer in that case, pick any of my suggestions for the one true release note.

doc/manual/rl-next/shebang-relative.md
roberth: This is a breaking change for example if you have a repo with a `scripts` directory and you invoke them as

```console
$ ./scripts/foo
```

in which case the `foo` script likely has a workaround for this issue,

```
#!/usr/bin/env nix-shell
#!nix-shell --expr 'import ./scripts/shell.nix { withBar = true; }'
```

 which will cause it to break when Nix is updated to a version after this PR.
 
```
error: getting status of '/home/user/src/unicorn/scripts/scripts/shell.nix': No such file or directory
```

Note the double `scripts/`.

I don't know if we should break an established command like this.
We should take this seriously and add `--relative` to opt in to the new behavior.
For what it's worth, it's also a way to "declare" that a recent Nix is needed by the script.

roberth (changes requested):

roberth:
I think it's great that we don't dogmatically only work on the new CLI, but we should take the old CLI more seriously when we change it.

fricklerhandwerk:
Agreed on being especially careful about stable commands, but we also quite clearly said that what's arguably a bug should be just fixed.

roberth:
I'm not opposed to the change, only to the fact that it's currently not opt-in.

Do you want more #9412? Because that's how you get more #9412.

doc/manual/src/release-notes/rl-next.md
tomberek: I was trying to update it to the new style. But then Github wasn't registering my pushes to the branch. I'll fixup and merge later tonight if still needed.

tomberek: fixed the git problems, rl-next issues and incorporated @fricklerhandwerk's suggestions

doc/manual/rl-next/shebang-relative.md
tomberek: @roberth here's an attempt at adding a new "--relative" flag: https://github.com/NixOS/nix/commit/2fd7ee21d11bf7b06aff674663fe618e41eff5e2

PR #5090, opened 2021-08-04
More support for `NO_COLOR` environment variable

nagy:
This builds on top of #4971. Without this, color is still shown in the
REPL.

See https://no-color.org/ for more information about the variable.

### DIFF
--- a/src/libutil/ansicolor.hh
+++ b/src/libutil/ansicolor.hh
@@ -2,16 +2,19 @@
 
 namespace nix {
 
+// A forward declaration from `util.hh`
+bool shouldANSI();
+
 /* Some ANSI escape sequences. */
-#define ANSI_NORMAL "\e[0m"
-#define ANSI_BOLD "\e[1m"
-#define ANSI_FAINT "\e[2m"
-#define ANSI_ITALIC "\e[3m"
-#define ANSI_RED "\e[31;1m"
-#define ANSI_GREEN "\e[32;1m"
-#define ANSI_YELLOW "\e[33;1m"
-#define ANSI_BLUE "\e[34;1m"
-#define ANSI_MAGENTA "\e[35;1m"
-#define ANSI_CYAN "\e[36;1m"
+#define ANSI_NORMAL  "\e[0m"
+#define ANSI_BOLD    "\e[1m"
+#define ANSI_FAINT   "\e[2m"
+#define ANSI_ITALIC  "\e[3m"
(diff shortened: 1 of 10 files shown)


nagy:
Since with these changes the `shouldANSI` functions gets called more often, maybe it deserves to introduce a static variable to cache the return value like so:

```patch
 bool shouldANSI()
 {
-    return isatty(STDERR_FILENO)
+    static bool should_ansi = isatty(STDERR_FILENO)
         && getEnv("TERM").value_or("dumb") != "dumb"
         && !getEnv("NO_COLOR").has_value();
+    return should_ansi;
 }
```

teoljungberg:
Maybe this is out of scope, but this does still colorize the nix-shell prompt that I don't believe can be customized anywhere.

image

---

Commands that used to have colors (such as the keyword `error` in the screenshot below was colorized red prior)
image

nagy:
> but this does still colorize the nix-shell prompt 

I think this should also be addressed. I will try to implement this and then undraft this PR.

src/libutil/ansicolor.hh
```diff
@@ -2,16 +2,19 @@
...
+#define ANSI_NORMAL  "\e[0m"
+#define ANSI_BOLD    "\e[1m"
+#define ANSI_FAINT   "\e[2m"
+#define ANSI_ITALIC  "\e[3m"
+#define ANSI_RED     (::std::string( ::nix::shouldANSI() ? "\e[31;1m" : ""))
+#define ANSI_GREEN   (::std::string( ::nix::shouldANSI() ? "\e[32;1m" : ""))
+#define ANSI_YELLOW  (::std::string( ::nix::shouldANSI() ? "\e[33;1m" : ""))
+#define ANSI_BLUE    (::std::string( ::nix::shouldANSI() ? "\e[34;1m" : ""))
+#define ANSI_MAGENTA (::std::string( ::nix::shouldANSI() ? "\e[35;1m" : ""))
+#define ANSI_CYAN    (::std::string( ::nix::shouldANSI() ? "\e[36;1m" : ""))
```
edolstra, line 18: I don't think this should be dynamic. The current approach is to use ANSI color unconditionally when constructing messages, and then strip ANSI codes when printing to a non-terminal. That way, messages from the daemon (whose stderr isn't connected to a terminal) get colorized properly.

nagy: In that case, I am closing the PR since I dont know how to implement it.

nagy:
Closing this PR, since I cannot implement the desired filtering approach.

PR #5095, opened 2021-08-05
docs: add troubleshooting session in remote builds

happysalada:
This PR just adds two sections to the remote builds docs.

It adds a trouble shooting section. @tomberek went with me through an ssh problem, where the remote nix store was reachable with `nix store ping` but the remote build failed an unrelated error

It adds a warnings section. This is just to say that there is a surprising behavior on darwin where the closures are copied back on the local machine. Even if the closure is a linux closure. I'm not sure here if there is a flag to disable that or if it's me.

One piece of feedback that I have is that running a remote build is much slower that just pushing a nix expression and building it on remote. Remote build actually tries to send lots of things over ssh other than the derivation. Somehow the build ends up waiting a lot. I tried increasing the build jobs count, but it didn't do much.
It makes me think that it would be amazing if there was a way to just ask to evaluate a nix expression on a remote machine.
The current solution that I basically use at the moment is to have a git hook and a remote repo on a machine and there is an automated build everytime I push a change to the git repo. It has the inconvenience of having to push to a repo, but it's very fast.

Last but not least, thank you for the awesome package manager!


domenkozar (changes requested):

doc/manual/src/advanced-topics/distributed-builds.md
```diff
@@ -156,3 +156,19 @@ option `builders-use-substitutes` in your local `nix.conf`.
...
+
+## Troubleshooting
+
+### ssh
+
+even if the store is reachable with `nix store ping --store your_store_url`, the nix-daemon might not be able to reach it. nix-daemon operates as root so you need to make sure that the machine is reachable via ssh as root user.
+
+- add the `-vvv` flags to show more information on the potential problem.
+- make sure your store is reachable by the root user. `sudo ssh -i "your_ssh_key_path" your_user@your_hostname`
+- make sure your builders configuration contains the ssh_key, user, and hostName
```
SuperSandro2000, line 168: Why not configure the user and key via ssh_config?

doc/manual/src/advanced-topics/distributed-builds.md
```diff
@@ -156,3 +156,19 @@ option `builders-use-substitutes` in your local `nix.conf`.
...
+
+## Warnings
+
+### darwin
+
+When using nix-darwin resulting closure will be copied back to your local machine.
```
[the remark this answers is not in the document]
happysalada: it was unintuitive to me that using a remote builder would copy things back onto your machine.
If you are on darwin and you are building something on nixos, the closures are not valid for your system. I don't understand why it would try to copy back into your machine. Since the behavior was surprising to me, I tried to add a warning.

doc/manual/src/advanced-topics/distributed-builds.md
happysalada: those need to be valid for the root user. ssh_config is for your user, if I'm not wrong.

doc/manual/src/advanced-topics/distributed-builds.md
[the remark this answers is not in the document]
domenkozar: That could be described in the general introduction of remote builds, as it's not specific to nix-darwin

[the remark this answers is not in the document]
happysalada: Agreed, would you think the fix is just removing this section here then ?
I'm happy to, just want to make sure I'm not misunderstanding.

doc/manual/src/advanced-topics/distributed-builds.md
SuperSandro2000: ssh_config is for every user, you just need to set it for the root user which is way more convenient than fiddling with ssh options around.

happysalada: how about if I changed that sentence to
`- make sure your ssh_config is set for the root user to access that server or that your builders configuration contains the ssh_key, user and hostname`

happysalada:
Took care of both of the comments.

happysalada:
Understood, I'll check for spelling and capitalization.

happysalada:
I re-opened in https://github.com/NixOS/nix/pull/5345
I am happy to modify or close the PR entirely if you think it's not valuable enough.
I tried to fix the language, punctuation and capitalization.

Issue #5065, opened 2021-07-29
import (builtins.fetchurl …) fails on chrooted store


regnat:
**Describe the bug**

When using a chrooted (or “diverted”) store, `import`-ing the result of a `fetchurl` fails with `error: path '…' is not in the Nix store`

**Steps To Reproduce**

```console
$ nix eval --impure --store /tmp/store --expr 'import (builtins.fetchurl "https://raw.githubusercontent.com/NixOS/nix/master/flake.nix")'
error: path '/tmp/store/nix/store/3pmpcm1gf44nrvfr5nh3kqh722czdifz-flake.nix' is not in the Nix store
(use '--show-trace' to show detailed location information)
```

**Expected behavior**

Should work

**`nix-env --version` output**

```console
$ nix-env --version                                                                                                   
nix-env (Nix) 2.4pre20210705_e700ecb
```

madeline-os:
This bug has reappeared in `master`. I have not yet bisected to see where exactly it broke.

Issue #3956, opened 2020-08-24
builtins.readDir does not honor --store


bluescreen303:
The following snippet works fine using normal `nix-build`, but fails when using `nix-build --store /mnt`. 

```
{pkgs ? import <nixpkgs> {}, lib ? pkgs.lib}:
with lib;

let foo = pkgs.runCommand "dir" {} ''
              mkdir -p $out/foo
              mkdir -p $out/bar
          '';

in pkgs.writeText "dir.json" (builtins.toJSON (builtins.readDir foo))
```

**Error**

```
error: opening directory '/nix/store/8g66y9n449pq8n46xrqc178nbqhjnrbf-dir': No such file or directory
```

**Expected behavior**

```
building '/nix/store/7l4a0ajxsx1pvazc5l3238db9p7cq0gp-dir.drv'...
these derivations will be built:
  /nix/store/61hz4ikmhghmbx8cx8savs8rxx4ddlj9-dir.json.drv
building '/nix/store/61hz4ikmhghmbx8cx8savs8rxx4ddlj9-dir.json.drv'...
/nix/store/2dpvnqprj8yym5rpm154bj9fh3vd60wz-dir.json
```

**Warning**
When running without `--store` once, the directory gets created in `/nix/store`, which makes any following tries **with** `--store` succeed. So watch out when trying to reproduce and always delete the "dir" derivation from the store.

**`nix-env --version` output**

nix-env (Nix) 2.3.7

Ericson2314:
Hmm, this is an interesting problem. `--store`, by default, create a store which, while stored elsewhere, is still mounted (in a chroot) at `/nix/store`. The problem is that building, and not eval, occurs in that chroot. `builtins.readFile` is also meant to pick up various impurities, so it is normal for it to not use a chroot-root-relative path.

Internal to Nix, there is a function "store.toRealPath" which will modify a path to include the path to the dir mounted as root, but I am not sure how it could get used here automatically. Maybe there is a builtin for it?

bluescreen303:
I searched for it, but didn't find it. `builtins.storeDir` is always "/nix/store" and builtins.storePath also does not seem to return a real path.

Yes, I understand `readFile` and `readDir` are meant to pick up impurities, that's what I use it for. But since /nix/store has been shadowed when running with `--store` I don't think there's a case for wanting to read the original during eval. But even when such a case does exist, passing it a derivation should be a clear indication a path from the `--store` version is requested.

Ericson2314:
[...]

PR #5099, opened 2021-08-06
Return the store path instead of the real path in fetchTree.cc::fetch

madeline-os:
This resolves issue https://github.com/NixOS/nix/issues/5065

With this change the real path is added to `allowedPaths` in the evaluation state, but the result of the fetch operation is the store path. At first I tried just referencing the store path in the string context, but that just pushed the error further down into `realiseContext`.

I'm not confident that this is quite the correct fix but i managed to build a substantial derivation with it. In any case I should add a test before this is considered ready. Each primop that uses this fetch implementation should be exercised.

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -229,20 +229,21 @@ static void fetch(EvalState & state, const Pos & pos, Value * * args, Value & v,
         ? fetchers::downloadTarball(state.store, *url, name, (bool) expectedHash).first.storePath
         : fetchers::downloadFile(state.store, *url, name, (bool) expectedHash).storePath;
 
-    auto path = state.store->toRealPath(storePath);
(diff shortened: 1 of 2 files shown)


Ericson2314:
Per https://github.com/NixOS/nix/issues/3956 we probably want a `builtins.toRealPath` for anyone relying on the old behavior.

Shados:
Did some light testing, this does appear to resolve the issue for `builtins.fetchTarball` at least, but I'm still seeing the same problem when using nixpkgs' `pkgs.fetchurl` & co.

madeline-os:
Oh that's strange. The example given in the reported issue worked fine for me on this patch. I will look into it further. thanks for trying it out!

Ericson2314:
`pkgs.fetchurl` just makes regular derivations, and so ought to be completely unaffected by this.

PR #5101, opened 2021-08-06
nix-shell --pure: Let it work for any derivation

### LINKED ISSUE #5092: `nix-shell --pure` with a custom derivation keeps the existing $PATH
**Describe the bug**
 I have been investigating trying to create a minimal `mkShell` via https://github.com/NixOS/nixpkgs/pull/132617
 I am running into the problem however that when I call `nix-shell --pure` the resulting _$PATH_ still has the prior items.
(it's still impure)

**Steps To Reproduce**
You can reproduce this with any simple derivation.
```nix
let
  nixpkgs = import (builtins.fetchTarball {
    name = "nixos-unstable-2020-09-24";
    url =
      "https://github.com/nixos/nixpkgs/archive/5aba0fe9766a7201a336249fd6cb76e0d7ba2faf.tar.gz";
    sha256 = "05gawlhizp85agdpw3kpjn41vggdiywbabsbmk76r2dr513188jz";
  }) { };
in
with nixpkgs;
derivation {
    name = "test-purity-path";
    system = builtins.currentSystem;
    builder = writeScript "builder.sh" ''
        export > $out
    '';
}
```

```bash
❯ nix-shell purity-derivation.nix --pure

[nix-shell:~/code/nix/playground]$ echo $PATH | tr : '\n'
/nix/store/fgbzvd4c6nly9m4dpczrxybdpkm8mnk3-bash-interactive-4.4-p23/bin
/home/fmzakari/.nix-profile/bin
/home/fmzakari/.nix-profile/bin
/home/fmzakari/.local/bin
/home/fmzakari/bin
/usr/local/sbin
/usr/local/bin
/usr/sbin
/usr/bin
/sbin
/bin
/home/fmzakari/.zsh/plugins/zsh-syntax-highlighting
/home/fmzakari/.zsh/plugins/powerlevel10k

[nix-shell:~/code/nix/playground]$ echo $IN_NIX_SHELL
pure
```

**Expected behavior**

I expect $PATH to be empty.

```
❯ nix-env --version
nix-env (Nix) 2.3.12
```

Is this something specific happening to _stdenv_ ?
I can't seem to find the offending code but it should be generic and not specific to _stdenv_.

fzakaria:
`nix-shell --pure` when applied to a non stdenv derivation doesn't seem
to clear the PATH. It expects the stdenv/setup file to do so.

This adds an explicit `unset PATH` by nix-build.cc (nix-shell) itself so
that it's not reliant on stdenv/setup anymore.

This does not break impure nix-shell since the PATH is persisted as the
variable `p` prior in the bash rcfile

This was tested with the following derivation
```nix
let
  nixpkgs = import (builtins.fetchTarball {
    name = "nixos-unstable-2020-09-24";
    url =
      "https://github.com/nixos/nixpkgs/archive/5aba0fe9766a7201a336249fd6cb76e0d7ba2faf.tar.gz";
    sha256 = "05gawlhizp85agdpw3kpjn41vggdiywbabsbmk76r2dr513188jz";
  }) { };
in
with nixpkgs;
derivation {
    name = "test-purity-path";
    system = builtins.currentSystem;
    builder = writeScript "builder.sh" ''
        export > $out
    '';
    FOO = "bar";
    passAsFile = ["FOO"];
}
```

```bash
❯  /home/fmzakari/code/github.com/NixOS/nix/result/bin/nix-shell purity-derivation.nix --pur
[nix-shell:~/code/nix/playground]$ echo $PATH
/nix/store/fgbzvd4c6nly9m4dpczrxybdpkm8mnk3-bash-interactive-4.4-p23/bin:

[nix-shell:~/code/nix/playground]$ exit
exit
bash: rm: command not found

❯ /home/fmzakari/code/github.com/NixOS/nix/result/bin/nix-shell purity-derivation.nix

[nix-shell:~/code/nix/playground]$ echo $PATH
/nix/store/fgbzvd4c6nly9m4dpczrxybdpkm8mnk3-bash-interactive-4.4-p23/bin::/home/fmzakari/.nix-profile/bin:/home/fmzakari/.nix-profile/bin:/home/fmzakari/.local/bin:/home/fmzakari/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/fmzakari/.zsh/plugins/zsh-syntax-highlighting:/home/fmzakari/.zsh/plugins/powerlevel10k
```

Interestingly it looks like the rcfile has a dependency on **rm** during a trap if we do _passAsFile_
I'd love if we could remove this dependency as well.
https://github.com/NixOS/nix/blob/94ec9e47030c2a7280503d338f0dca7ad92811f5/src/nix-build/nix-build.cc#L484

fixes #5092

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -489,6 +489,9 @@ static void main_nix_build(int argc, char * * argv)
                     "_nix_shell_clean_tmpdir; ") +
                 (pure ? "" : "[ -n \"$PS1\" ] && [ -e ~/.bashrc ] && source ~/.bashrc;") +
                 "%2%"
+                // always clear PATH.
+                // when nix-shell is run impure, we rehydrate it with the `p=$PATH` above
+                "unset PATH;"
                 "dontAddDisableDepTrack=1;\n"
                 + structuredAttrsRC +
                 "\n[ -e $stdenv/setup ] && source $stdenv/setup; "


fzakaria:
@roberth ^^ 
Relevant to some of the chatting we were doing.
Small fix but kinda relevant.

I might open an issue to see if the community is open to supporting a special attribute "__SHELL" that you can set on a derivation that would pin the version of the shell used by nix-shell rather than relying on the channel _<nixpkgs>_
https://github.com/NixOS/nix/blob/94ec9e47030c2a7280503d338f0dca7ad92811f5/src/nix-build/nix-build.cc#L365

roberth:
`nix-shell` is very tightly coupled with bash, stdenv, nixpkgs. `nix develop` is very similar, except it replaces `<nixpkgs>` by new undocumented code that gives special treatment to `nixpkgs` flake inputs.

We should move on from the many things that `nix-shell` and `nix develop` currently do, because it's architecturally flawed and limits the things Nix can do.

How about one command that runs the actual sandbox, but with a stdin. That's already enough for stdenv to give you an interactive shell on failure, or provide a "breakpoint" function that does the same where you invoke it.

With the use case of troubleshooting builds covered, we can simplify how the "developer shell" feature is implemented. No need to worry about cleaning up build-related files etc. It could be as simple as a string with context, placed in the `devShell` attribute of `default.nix` or in the `devShell.${system}` in a flake. Just build the context and exec the string. This way, Nix does what the expression author wants and no more.

So, I guess the issue is deeper than `__SHELL`. Did I mention bash-independence? That's possible with something like I described. It's definitely something that people will expect when you give them `__SHELL`.

roberth:
The change itself makes sense given the circumstances.

fzakaria:
I love that idea.

There's a lot of similarities to what Zimbatm is doing with his project.

I think you hit the nail on the head of the separation between debugging a
failing derivation and the developer environment.

Separating them explicitly is key to a simple solution.

We should collaborate on this.
(I also am searching for a PhD project 😂)

PR #5106, opened 2021-08-08
libstore: use set instead of list for waiter list

andir:
This replaces the O(n) search complexity in our insert code with a
lookup of O(log n). It also makes removing waitees easier as we can use
the extract method provided by the set class.


edolstra:
Thanks, didn't know about `std::owner_less`!

Issue #3019, opened 2019-08-01
Race substitutors against each other


michaelpj:
From my reading of `SubstitutionGoal`, it tries each substitutor in sequence to try and find one that has the goal.

However, if you have a number of substitutors this can take a little while, especially since the 404 case can be particularly slow. Not a long time, but noticeable.

I think this is more relevant than it used to be. Historically most people would have just had the nixos cache. With tools like `cachix` that encourage users to add new binary caches for specific use cases, it's easy to have quite a few binary caches.

We could instead race the substitutors against each other, i.e. try them in parallel, take the first one that returns a success and kill the others.

The really radical version would be to also race the *build* against the substitution. The gain here would be that adding substitutors adds no latency to completing a build that we need to do locally, since we would always start it immediately. However, the build might trigger more substitutions, all of which we'd want to kill if we eventually got a substitution for the original goal.

domenkozar:
@edolstra does this sound good to you if it was implemented?

NorfairKing:
https://github.com/NixOS/nix/issues/3514 Relevant.

@edolstra It sounds like @domenkozar would be interested in implementing this. What do you think?

domenkozar:
My C++ is not sufficient enough to fix this, but I can fix it on cachix side with https://github.com/cachix/cachix/issues/221

NorfairKing:
Still relevant.

NorfairKing:
Still not stale

ch1bo:
> 404 case can be particularly slow. Not a long time, but noticeable.
Stumbled upon this when our substituter was down and it was stalling completely for a long time. The connection was not able to be established at all in this case AFAIS, e.g. 
```
curl -v https://hydra.iohk.io/ql8jd1n7l07b43fnr0wxyikhxafhr4am.narinfo
*   Trying 147.75.83.98:443...
```

Dietr1ch:
I kind of need this feature to set up network local nix store sharing (slow internet :cry:)


So far I'm using nix-serve-ng and priority 40 (same as cache.nixos.org/nix-cache-info) in hopes that I can download from any machine. ([Setup details](https://github.com/Dietr1ch/systemn/tree/master/server/nix-serve.nix))
This somehow works fine when all machines are online and I get parallel downloads, but if one of them is offline and happens to be queried first, then operations just hang on `Could not resolve hostname intermitent-peer.local`.

PR #5109, opened 2021-08-08
Short-circuit querying substituters on success

Kha:
This PR ensures that `querySubstitutablePathInfos` returns the `SubstitutablePathInfo`s of the *first* succeeding substituter, i.e. the one that will eventually be used for the actual substitution. Semantically this doesn't matter since no caller of this function checks for anything other than success or failure anyway, but it does mean that we can skip querying the other substituters.

On a project that makes heavy use of small derivations with configured substituters (in order)
* a file:// binary cache restored from a GitHub Actions cache,
* Cachix,
* and cache.nixos.org

this change speeds up the delay between evaluation and substitution (that is, the "querying info about missing paths" part) from 25s to 2.5s if all 842 derivations can be fetched from the local binary cache.

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1076,6 +1076,10 @@ void LocalStore::querySubstitutablePathInfos(const StorePathCAMap & paths, Subst
     if (!settings.useSubstitutes) return;
     for (auto & sub : getDefaultSubstituters()) {
         for (auto & path : paths) {
+            if (infos.find(path.first) != infos.end())
+                // choose first succeeding substituter
+                continue;
+
             auto subPath(path.first);
 
             // recompute store path so that we can use a different store root
--- a/src/libstore/store-api.hh
+++ b/src/libstore/store-api.hh
@@ -430,9 +430,10 @@ public:
     virtual StorePathSet querySubstitutablePaths(const StorePathSet & paths) { return {}; };
 
     /* Query substitute info (i.e. references, derivers and download
-       sizes) of a map of paths to their optional ca values. If a path
-       does not have substitute info, it's omitted from the resulting
-       ‘infos’ map. */
+       sizes) of a map of paths to their optional ca values. The info
+       of the first succeeding substituter for each path will be
+       returned. If a path does not have substitute info, it's omitted
+       from the resulting ‘infos’ map. */
     virtual void querySubstitutablePathInfos(const StorePathCAMap & paths,
         SubstitutablePathInfos & infos) { return; };
 


tomberek:
How does this interact with https://github.com/NixOS/nix/blob/master/src/libstore/daemon.cc#L726-L746 ?

Kha:
Yes, this is the one caller that actually inspects the returned info - but at least within Nix, there is no user of this daemon operation that does so, so the result is again irrelevant. And as described above the info returned after this change is arguably "more correct", anyway.

src/libstore/local-store.cc
```diff
@@ -1076,6 +1076,10 @@ void LocalStore::querySubstitutablePathInfos(const StorePathCAMap & paths, Subst
     if (!settings.useSubstitutes) return;
     for (auto & sub : getDefaultSubstituters()) {
         for (auto & path : paths) {
+            if (infos.find(path.first) != infos.end())
+                // choose first succeeding substituter
+                continue;
```
andir, line 1081: How will this affect signature verification? If I have a substituter that I do not trust in the list of substituters it will only ever yield the path info from that substituter?

Kha: Again, there is no semantic change here because the info returned from this function is never actually inspected, and returning the first succeeding substituter is at the very least no less, and arguably more, correct than returning the last one, which is is the current behavior. Signature validation is part of the actual substitution, not the initial querying of substituters that uses this function.

Kha:
So I tried to better understand why the ["query substituters without actually substituting" phase](https://github.com/NixOS/nix/blob/223e0569ff2c93d8a907188b40d1928d6967e91f/src/libstore/build/worker.cc#L241-L244) between evaluation and building/substitution exists at all. The comment opaquely talks about "efficiency", and a [prior commit message](https://github.com/NixOS/nix/commit/bbdf08bc0facb5157a10c794712dae7e5902be03) implies that the goal is to prime the substituters/nar info cache *in parallel*. I'm not quite sure why the actual build loop can't do this, at least when `max-jobs` is sufficiently high.

In any case, this should mean that *if* a substituter claims to know a set of paths, but then *repeatedly* fails to actually provide those paths, this PR could slightly slow down substitution because the fallback substituters miaght not be queried with the same level of parallelism. But that sounds like a very fringe case that doesn't justify slowing down the common case.

src/libstore/local-store.cc
andir: Thanks for clarifying. I didn't realize that.

tomberek:
Perhaps update comment at https://github.com/NixOS/nix/blob/master/src/libstore/store-api.hh#L436 to clarify what is returned.

Mic92:
@Kha I am curios how you are using the github action cache for nix. mind sharing?

Kha:
@Mic92 Sure, it's all open source:
https://github.com/leanprover/lean4/blob/9686910c72aaf1d39d28792265950b8f9f44880e/.github/workflows/nix-ci.yml#L60
https://github.com/leanprover/lean4/blob/9686910c72aaf1d39d28792265950b8f9f44880e/.github/workflows/nix-ci.yml#L103-L106

Here are two substituted runs before and after adding the cache:
https://github.com/leanprover/lean4/runs/3279221584?check_suite_focus=true#step:9:458 (this one does a bit of building afterwards, didn't find a completely clean one)
https://github.com/leanprover/lean4/runs/3233727788?check_suite_focus=true#step:8:475

As you can see, substitution of the many drvs is sped up from 46s to 5s with the local cache, in return for 5s + 6s + 8s for restoring, rebuilding, and saving the cache. Not bad, but with this patch it wouldn't spend those 81s just above the linked line uselessly querying lean4.cachix.org and cache.nixos.org. Oh, and the Actions cache is also nice for PRs that cannot push to Cachix.

domenkozar:
@edolstra what do you think about this improvement?

nrdxp:
from my understanding, the caches are currently checked in a synchronous fashion. Would there be any simple way to mimic the tokio [select](https://docs.rs/tokio/0.2.13/tokio/macro.select.html) macro from rust to simply check all substituters simultaneously, returning on the first success, and cancelling the others?

This seems like the ideal behavior here (for performance anyway).

domenkozar:
That's https://github.com/NixOS/nix/issues/3019 - not really a low hanging fruit compared to what @Kha is suggesting.

edolstra:
> Would there be any simple way to mimic the tokio select macro from rust to simply check all substituters simultaneously, returning on the first success, and cancelling the others?

Not in C++ (especially not the cancellation part). In async Rust it's a lot easier (see e.g. https://github.com/NixOS/nix/blob/7ee639f9db987a442fdf3820d3681a77b2ee1688/nix-rust/src/store/store.rs#L22-L52 for an implementation of `computeFSClosure()` that has a fraction of the complexity of the C++ implementation).

PR #5121, opened 2021-08-10
Backport 2.3: Set the NIX_PATH a bit more like master does it

Ericson2314:
1. Per-user profile before system-wide

2. Already-defined NIX_PATH (including empty) should not be modified.

However, I did *not* take master's change of only adding directories
which exist, because I think that would cause a regression where the
directories didn't yet exist, but then one ran a `nix-env` or
`nix-channel` command which creates them. Nix master would correctly
re-default the command each page load, but with Nix 2.3 the shell
profile / rc would not be automatically reloaded.

The commit on master we're imitating is
ec9dd9a5aeb9b7fca170d86b8906ffac7e5ef057.

### DIFF
--- a/scripts/nix-profile-daemon.sh.in
+++ b/scripts/nix-profile-daemon.sh.in
@@ -34,5 +34,15 @@ else
   unset -f check_nix_profiles
 fi
 
-export NIX_PATH="nixpkgs=@localstatedir@/nix/profiles/per-user/root/channels/nixpkgs:@localstatedir@/nix/profiles/per-user/root/channels"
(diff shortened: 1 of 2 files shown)


edolstra:
Is this a bug fix? Because otherwise I'd prefer not to mess with `NIX_PATH` on the stable branch.

Ericson2314:
Not really, unless one counts the two scripts acting rather differently (and different from NixOS's too) as a bug.

Ericson2314:
2.4 is close enough

PR #5145, opened 2021-08-17
Docs build: depend on locally built nix executable and not installed one

### LINKED ISSUE #5140: Some of the targets are installed into `$(prefix)` during the build phase
**Describe the bug**

Some targets are installed into `$(prefix)` during the build phase

While building Nix from the master branch, some targets like `libnixutil.so` are installed into `$(prefix)` during the build phase. If the build user doesn't have permission to write into `$(prefix)` the build fails.

It doesn't happen on the 2.3-maintenance branch, or if the build is configured with the `--disable-doc-gen` option on the master branch.

**Steps To Reproduce**

1. Checkout Nix master branch
2. `./bootstrap.sh`
3. `./configure --prefix=/usr`
4. `make`
```
  GEN    Makefile.config
  GEN    src/libexpr/parser-tab.cc
  GEN    src/libexpr/lexer-tab.cc
  CXX    src/build-remote/build-remote.o
  CXX    src/nix-build/nix-build.o
  CXX    src/nix-channel/nix-channel.o
  CXX    src/nix-collect-garbage/nix-collect-garbage.o
  CXX    src/nix-copy-closure/nix-copy-closure.o
  ...
  CXX    src/libutil/util.o
  CXX    src/libutil/xml-writer.o
  LD     /usr/lib64/libnixutil.so
/opt/rh/gcc-toolset-9/root/usr/bin/ld: cannot open output file /usr/lib64/libnixutil.so: Permission denied
make: *** [mk/lib.mk:117: /usr/lib64/libnixutil.so] Error 1
```
**Expected behavior**

Targets aren't installed `$(prefix)` during the build

fedepell:
Previously many of the documentation targets were depending on
$(bindir)/nix which is the installed version. This meant that its
install rules would be triggered (which in chain would also trigger
the install of libraries, as reported in #5140). Therefore a build
of the documentation without an installation would not be possible
(which apart from doing unwanted operations it may also generate
permission problems for example).
The fix makes the rules depend on $(nix_DIR)/nix instead, which is
the executable in the build tree.

Fixes #5140.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,7 @@ makefiles = \
   misc/zsh/local.mk \
   misc/systemd/local.mk \
   misc/launchd/local.mk \
-  misc/upstart/local.mk \
-  doc/manual/local.mk \
-  doc/internal-api/local.mk
+  misc/upstart/local.mk
 endif
 
 ifeq ($(ENABLE_BUILD)_$(ENABLE_TESTS), yes_yes)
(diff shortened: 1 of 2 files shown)


fedepell:
This still seems to be present in current master. Would it make sense to merge this? (I didn't re-check if it still applies cleanly right now, but can do it if the MR would be considered worth still)

Thanks!
F.

Ericson2314 (approved):

Issue #5146, opened 2021-08-17
Naming overlays arguments something other than `final` and `prev` should not be a hard failure


deliciouslytyped:
**Describe the bug**
I'm not familiar with the internals and intentions of `nix flake check`, however I was helping @bryanhonof figure out what was wrong when he got `overlay does not take an argument named 'final'`.

There are several issues here:
1. lambdas should not semantically care about what their arguments are called
2. the error message is extremely misleading and suggests some manner of evaluation error
3. stylistic/linting issues should not cause a hard failure
4. linting issues causing hard failure prevents the checks from proceeding further

**Steps To Reproduce**
@bryanhonof can you give a repro for this?

**Expected behavior**

We will try to alleviate 2. a bit, by PR-ing a better error message.

We also took a short look at fixing the hard-failure problem but the changes looked not-quite-trivial - or at least we didn't offhand see how we should deal with the test suite, etc. I also don't know `nix flake check` how nix flake check is expected to behave, but standard things should be included, like being able to override checks if you decide to maintain everything as a hard failure. - otherwise, the solution is to separate linter issues, and give them as warnings.

**`nix-env --version` output**
`nix-env (Nix) 2.4pre20210802_47e96bb`

bryanhonof:
> **Steps To Reproduce**
> @bryanhonof can you give a repro for this?

Any flake where the `overlay` attribute is a lambda that takes anything else than `final: prev:` should reproduce the error. E.g. [the flake template from poetry2nix](https://github.com/nix-community/poetry2nix/blob/master/templates/app/flake.nix) is able to reproduce this error. Since the `nixpkgs.lib.composeManyExtensions` returns a lambda with `self: super:`.

deliciouslytyped:
The responsible code is at https://github.com/NixOS/nix/blob/c000cec27fcb16548606830410be265eb082f777/src/nix/flake.cc#L333

deliciouslytyped:
You can use `hack = f: final: prev: f final prev` as a workaround.

bryanhonof:
[Someone has fixed the `lib.compose{Many,}Extensions` in NixPkgs](https://github.com/NixOS/nixpkgs/commit/d09222e48f0a7124522a4d8c484d27a5125efef3). So this is no longer reproducible with the `poetry2nix` flake template. Instead try the following flake: 

```nix
#flake.nix
{
  description = "Overlay error message flake";

  inputs.flake-utils.url = "github:numtide/flake-utils";
[...]

PR #5147, opened 2021-08-17
Change Error message @checkOverlay

bryanhonof:
The current error message makes it look like something went wrong during parsing of the flake. Whilst it's just a new convention to use `final: prev: ...` instead of `self: super: ...`. This change makes it more clear what the user did wrong.

See #5146.

### DIFF
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -333,7 +333,11 @@ struct CmdFlakeCheck : FlakeCommand
         auto checkOverlay = [&](const std::string & attrPath, Value & v, const Pos & pos) {
             try {
                 state->forceValue(v, pos);
-                if (!v.isLambda() || v.lambda.fun->matchAttrs || std::string(v.lambda.fun->arg) != "final")
+                if (!v.isLambda())
+                   throw Error("Expected a lambda for the overlay attribute, but got something else");
+                if (v.lambda.fun->matchAttrs)
+                    throw Error("By convention the attribute 'overlay' is expected to be a simple lambda starting with 'final: ', not an attribute set matching lambda");
+                if (std::string(v.lambda.fun->arg) != "final")
                     throw Error("By convention the attribute 'overlay' expects the argument 'final' not '" + std::string(v.lambda.fun->arg) + "'");
                 auto body = dynamic_cast<ExprLambda *>(v.lambda.fun->body);
                 if (!body || body->matchAttrs || std::string(body->arg) != "prev")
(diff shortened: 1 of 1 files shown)


src/nix/flake.cc
```diff
@@ -334,10 +334,10 @@ struct CmdFlakeCheck : FlakeCommand
             try {
                 state->forceValue(v, pos);
                 if (!v.isLambda() || v.lambda.fun->matchAttrs || std::string(v.lambda.fun->arg) != "final")
-                    throw Error("overlay does not take an argument named 'final'");
+                    throw Error("By convention the attribute 'overlay' expects the argument 'final' not '" + std::string(v.lambda.fun->arg) + "'");
```
roberth, line 337: This message isn't right for "matchAttrs" type lambdas, such as `{ foo, bar }: x`. That case should have its own message.
Same below.

bryanhonof: @roberth Do you have a proposal for what error it should throw?

roberth: By convention the attribute 'overlay' is expected to be a simple lambda starting with 'final: ', not an attribute set matching lambda.

I guess you could add a representation of it too, but a source position `Pos` would be even better.

edolstra: Position info is already added to the error context by the `try { ... } catch { ... }`.

edolstra, line 337: This will crash if `v` is not a lambda.

bryanhonof: In the new commit the `!v.isLambda()` check is now isolated, and also throws a more appropriate error message, telling the user the attribute should be a lambda.

src/nix/flake.cc
```diff
@@ -333,11 +333,15 @@ struct CmdFlakeCheck : FlakeCommand
         auto checkOverlay = [&](const std::string & attrPath, Value & v, const Pos & pos) {
             try {
                 state->forceValue(v, pos);
-                if (!v.isLambda() || v.lambda.fun->matchAttrs || std::string(v.lambda.fun->arg) != "final")
-                    throw Error("overlay does not take an argument named 'final'");
+                if (!v.isLambda())
+                   throw Error("Expected a lambda for the overlay attribute, but got something else");
+                if (v.lambda.fun->matchAttrs)
+                    throw Error("By convention the attribute 'overlay' is expected to be a simple lambda starting with 'final: ', not an attribute set matching lambda");
```
edolstra, line 339:
```suggestion
                    throw Error("the attribute 'overlay' must a function that takes a single argument named 'final ', not an attribute set");
```

It's not a convention but a requirement. Also instead of "lambda" it's better to say "function".

roberth: What is responsible for this "requirement"?

Assuming we do need this, why say "function", a general term, when the value must be provided by a lambda? Or does Nix have another implementation of "function" that is acceptable here, but is not a lambda?

edolstra: This exact check is the requirement.

PR #5149, opened 2021-08-17
Non blocking gc

edolstra:


Mic92:
cc @domenkozar maybe relevant for cachix watch-store?

domenkozar:
Will Nix then consider referrers paths a valid?

Ericson2314:
I hope referrers to a path will still prevent deletions!

edolstra:
Yes, if a path has any refererrs, then deletion will fail. Previously it would recursively delete unreachable referrers (especially dangerous in combination with `--ignore-liveness`).

src/libutil/util.cc
```diff
@@ -1714,6 +1711,66 @@ AutoCloseFD createUnixDomainSocket(const Path & path, mode_t mode)
...
+            strcpy(addr.sun_path, base.c_str());
+            if (connect(fd, (struct sockaddr *) &addr, sizeof(addr)) == -1)
+                throw SysError("cannot connect to socket at '%s'", path);
+            _exit(0);
+        });
+        int status = pid.wait();
+        if (status != 0)
+            throw Error("cannot connect to socket ar '%s'", path);
+    } else {
+        strcpy(addr.sun_path, path.c_str());
```
Mic92, line 1767:
```suggestion
        memset(addr.sun_path, path.c_str(), path.size() + 1);
```

src/libutil/util.cc
```diff
@@ -1714,6 +1711,66 @@ AutoCloseFD createUnixDomainSocket(const Path & path, mode_t mode)
...
+
+    if (path.size() + 1 >= sizeof(addr.sun_path)) {
+        Pid pid = startProcess([&]() {
+            auto dir = dirOf(path);
+            if (chdir(dir.c_str()) == -1)
+                throw SysError("chdir to '%s' failed", dir);
+            std::string base(baseNameOf(path));
+            if (base.size() + 1 >= sizeof(addr.sun_path))
+                throw Error("socket path '%s' is too long", base);
+            strcpy(addr.sun_path, base.c_str());
```
Mic92, line 1758:
```suggestion
            memset(addr.sun_path, base.c_str(), base.size() + 1);
```

src/libutil/util.cc
```diff
@@ -1714,6 +1711,66 @@ AutoCloseFD createUnixDomainSocket(const Path & path, mode_t mode)
...
+            if (bind(fd, (struct sockaddr *) &addr, sizeof(addr)) == -1)
+                throw SysError("cannot bind to socket '%s'", path);
+            _exit(0);
+        });
+        int status = pid.wait();
+        if (status != 0)
+            throw Error("cannot bind to socket '%s'", path);
+    } else {
+        strcpy(addr.sun_path, path.c_str());
```
Mic92, line 1738:
```suggestion
        strcpy(addr.sun_path, path.c_str(), path.size() + 1);
```

src/libutil/util.cc
```diff
@@ -1714,6 +1711,66 @@ AutoCloseFD createUnixDomainSocket(const Path & path, mode_t mode)
...
+
+    if (path.size() + 1 >= sizeof(addr.sun_path)) {
+        Pid pid = startProcess([&]() {
+            auto dir = dirOf(path);
+            if (chdir(dir.c_str()) == -1)
+                throw SysError("chdir to '%s' failed", dir);
+            std::string base(baseNameOf(path));
+            if (base.size() + 1 >= sizeof(addr.sun_path))
+                throw Error("socket path '%s' is too long", base);
+            strcpy(addr.sun_path, base.c_str());
```
Mic92, line 1729:
```suggestion
            strcpy(addr.sun_path, base.c_str(), base.size() + 1);
```

src/libstore/gc.cc
```diff
@@ -774,44 +782,21 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
...
 
-            /* Read the store and immediately delete all paths that
-               aren't valid.  When using --max-freed etc., deleting
-               invalid paths is preferred over deleting unreachable
-               paths, since unreachable paths could become reachable
-               again.  We don't use readDirectory() here so that GCing
-               can start faster. */
+            /* Read the store and delete all paths that are invalid or
+               unreachable. We don't use readDirectory() here so that
+               GCing can start faster. */
```
roberth, line 790: What's the strategy for making sure clients can safely add a gc root for the paths they create during the gc process?

I expected the store directory to be read at the very start, to limit the deletable paths to those that existed at the start. That way only one bad interleaving remains: client/addToStore -> gc/readdir -> gc/findRoots -> client/addRoot, but this seems already far less likely than client/addToStore -> gc/readdir -> client/addRoot -> gc/findRoots.
Ideally we'd have implicit temporary roots per daemon connection or per LocalStore object.

> We don't use readDirectory() here so that GCing can start faster.

It's a synchronous interleaving, so any real speedup would have to come from kernel read-ahead cleverness. If you start it early in the gc process, you could run it concurrently.

src/libutil/util.cc
edolstra: This should also be memcpy right?

src/libstore/gc.cc
edolstra: > What's the strategy for making sure clients can safely add a gc root for the paths they create during the gc process?

Clients connect to the garbage collector to inform it about temporary roots.

src/libutil/util.cc
Mic92: right.

src/libutil/util.cc
```diff
@@ -1714,6 +1711,66 @@ AutoCloseFD createUnixDomainSocket(const Path & path, mode_t mode)
...
+            if (connect(fd, (struct sockaddr *) &addr, sizeof(addr)) == -1)
+                throw SysError("cannot connect to socket at '%s'", path);
+            _exit(0);
+        });
+        int status = pid.wait();
+        if (status != 0)
+            throw Error("cannot connect to socket ar '%s'", path);
```
dtzWill, line 1765:
```suggestion
            throw Error("cannot connect to socket at '%s'", path);
```

src/libstore/gc.cc
```diff
@@ -147,24 +112,55 @@ void LocalStore::addTempRoot(const StorePath & path)
...
+            assert(c == '1');
+            debug("got ack for GC root '%s'", printStorePath(path));
+        } catch (SysError & e) {
+            /* The garbage collector may have exited, so we need to
+               restart. */
+            if (e.errNo == EPIPE) {
+                debug("GC socket disconnected");
+                state->fdRootsSocket.close();
+                goto restart;
+            }
```
andersk, line 146: This is missing a rethrow for other error codes (#6648).

Issue #2437, opened 2018-09-20
--daemon installer assumes the presence of rsync


srhb:
When `rsync` does not exist, installation fails and the user has to go through uninstalling in order to try again with rsync installed. It would be nice to either not depend on rsync being present, or guard the whole install on its presence earlier.

masaeedu:
Could someone point me to the installer files? Just got bitten by this.

zimbatm:
see `scripts/install-multi-user.sh`

tv42:
Bugs don't disappear just because time passes.

nathan-at-least:
I just encountered this 3+ year old bug while running `sh <(curl -L https://nixos.org/nix/install) --daemon`, which is the very first recommended command on [the install guide](https://nixos.org/guides/install-nix.html).

The target system is:

```
user@penguin:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
```

domenkozar:
I've hit this one too. A good start would be `require_util` used to guard against rsync.

PR #5150, opened 2021-08-17
Remove rsync usage in the installer

domenkozar:
It's not commonly installed on systems like debian,
so avoid the bootstrapping problem by using cp and
chmod.

Not tested yet.

Fixes https://github.com/NixOS/nix/issues/2437

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -701,7 +701,10 @@ install_from_extracted_nix() {
         cd "$EXTRACTED_NIX_PATH"
 
         _sudo "to copy the basic Nix files to the new store at $NIX_ROOT/store" \
-              rsync -rlpt --chmod=-w ./store/* "$NIX_ROOT/store/"
+              cp -RLp ./store/* "$NIX_ROOT/store/"
+
+        _sudo "to make the new store non-writable at $NIX_ROOT/store" \
+              chmod -R ugo-w "$NIX_ROOT/store/"
 
         if [ -d "$NIX_INSTALLED_NIX" ]; then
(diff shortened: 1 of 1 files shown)


scripts/install-multi-user.sh
```diff
@@ -701,7 +701,10 @@ install_from_extracted_nix() {
         cd "$EXTRACTED_NIX_PATH"
 
         _sudo "to copy the basic Nix files to the new store at $NIX_ROOT/store" \
-              rsync -rlpt --chmod=-w ./store/* "$NIX_ROOT/store/"
+              cp -rdp ./store/* "$NIX_ROOT/store/"
+
+        _sudo "to make the new store non-writable at $NIX_ROOT/store" \
+              chmod ugo-w "$NIX_ROOT/store/"
```
SuperSandro2000, line 707:
```suggestion
              chmod -w "$NIX_ROOT/store/"
```

domenkozar: Good point about `-R`!

removing `ugo` breaks the installer as write permission is left for the group.

SuperSandro2000: Doesn't -w remove it for all? I would have thought it bad whatever.

domenkozar: I thought that too, but it removes it only for the user.

domenkozar:
Tested at https://github.com/domenkozar/test-cachix-installer/runs/3369467883?check_suite_focus=true using Debian self-hosted runner (that previously failed due to rsync missing).

domenkozar:
Another benefit is smaller Nix closure for self-contained statically linked Nix.

PR #5156, opened 2021-08-20
nix develop: Fix `devShells` lookup

jtojnar:
It currently fails with the following error:

    error: flake 'git+file://…' does not provide attribute 'devShells.x86_64-linuxhaskell', 'packages.x86_64-linux.haskell', 'legacyPackages.x86_64-linux.haskell' or 'haskell'


jtojnar:
Also, it looks like `packages` in


https://github.com/NixOS/nix/blob/7a54b10a8a29dadea2bb2c639a15cfec25a4d38c/src/libcmd/installables.cc#L169

PR #5162, opened 2021-08-22
Make the prompt used in normal shells configurable

tomberek:
Attempting to allow `nixConfig.bash-prompt` and `nixConfig.bash-prompt-suffix` in `nix shell` to match behavior of `nix develop`.

If this is not in line with the expected semantics of `nix shell`, we can close.

The motivating problem is that during use of `nix shell` and `nix develop` during Summer of Nix, some would end up in nested shells of either flavor and this would cause "interesting issues" because there is no feedback by default that they are in a nested environment. Adding a nixConfig to their flakes would allow a way to alleviate the problem, but it would only work for `nix develop`.

Potential problems: just as with `nix develop`, this may cause PS1 changes that a user does NOT want.

Note: If there is a better way to access the relevant developSettings instead of reaching into globalConfig, I'd love to learn.

```
  nixConfig.bash-prompt = ''\[\033]2;\[\033[1;32m\][prompt] \[\033[0m\]'';
  nixConfig.bash-prompt-suffix = "[suffix] ";
```

### DIFF
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -16,6 +16,7 @@
 #endif
 
 #include <queue>
+#include <nlohmann/json.hpp>
 
 using namespace nix;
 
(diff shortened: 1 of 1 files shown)


edolstra:
Currently `nix shell` is very minimal, it just extends PATH and doesn't depend on any particular shell. This sets it apart from `nix develop`. So I'm a bit hesitant to add shell-specific functionality... I think eventually we should add a flake output type to declare arbitrary environment variables in `nix shell`.

tomberek:
> I think eventually we should add a flake output type to declare arbitrary environment variables in `nix shell`

@edolstra Is this new force-push closer to what you had in mind? Where there is an "environment.${system}" attribute in the flake schema which is evaluated, built, and used to set the environment for `nix shell`?

Issue #3234, opened 2019-11-21
restrict-eval and readFile on builtins.path


nmattia:
There is something I do not understand about `restrict-eval`. My assumption was that it disallowed any `readFile` et al. from derivations, but that appears not to be the case. I put together this Nix snippet, of which some targets fail to evaluate, which I cannot explain:

``` nix
let
  pkgs = import <nixpkgs> {};
  drv = pkgs.runCommand "foo" {} "cp -r '${./.}' $out";
  drv2 = pkgs.runCommand "foo" {} "cp -r '${builtins.path { path = ./.; }}' $out";
  ps =
    { # works
      p1 = ./file;

      # works
      p2 = ./. + "/file" ;

      # access to path /nix/store/... is forbidden in restricted mode
      p3 = "${./. + "/file"}" ;

      # access to path /nix/store/.../file is forbidden in restricted mode
      p4 = "${./.}" + "/file" ;

      # works
      p5 = drv + "/file" ;

      # access to path /nix/store/.../file is forbidden in restricted mode
      p6 = pkgs.lib.cleanSource ./. + "/file" ;

      # access to path /nix/store/.../file is forbidden in restricted mode
      p7 = builtins.path { path =  ./.; } + "/file" ;

      # works
      p8 = drv2 + "/file" ;
    };
in
  pkgs.lib.mapAttrs (
    n: p:
      pkgs.runCommand "build-${n}" {} "echo '${builtins.readFile p}' > $out"
    ) ps
```

It is basically different ways of reading a file, which may or may not come from a derivation. Run with the following command:

> for i in `seq 1 8`; do echo -n "p$i:"; NIX_PATH=src=.:nixpkgs=/path/to/nixpkgs/ nix-build --option restrict-eval true -A p$i || echo "failed: p$i"; done

* `p1` working makes sense.
* `p2` working makes sense as well, since we create a `path` from an existing path.
* `p3` failing fits with my understanding of restrict-eval preventing reading from derivation, as well as `p4`, `p6` and `p7` failing.
* `p5` working fits my _new_ understanding that restrict-eval does _not_ prevent reading from derivation. This however contradicts the point above.
* `p8` is a complete mystery, unless `builtins.path` itself creates some issue that is then resolved by wrapping everything in a (non-`builtins.path`) derivation.

The question is: is this a bug in `builtins.path`? if not, what is the benefit of preventing access to `builtins.path` derivations (`p7`), but not to derivations importing `builtins.path` (`p8`)?

[...]

PR #5163, opened 2021-08-23
Allow use of path and filterSource in flakes

### LINKED ISSUE #2538: filterSource is disallowed in restrict-eval
`restrict-eval` does not allow access to paths that are derived from the search path by `filterSource`.

Is there a reason to disallow this?

It seems that it could be implemented by making [`addPath`](https://github.com/NixOS/nix/blob/8327a7a8fa0ce6f6b1d1e0ea1324d053a031e438/src/libexpr/primops.cc#L1031) perform the `restrict-eval` check on the `path` argument and if it was ok, add it to `evalState.allowedPaths` at the end.

This would prevent problems like https://github.com/NixOS/nixpkgs/pull/50342

-----

Documentation from the Nix manual:

> If set to true, the Nix evaluator will not allow access to any files outside of the Nix search path (as set via the NIX_PATH environment variable or the -I option), or to URIs outside of allowed-uri.

### LINKED ISSUE #3732: `builtins.filterSource` does not work in flakes
As discussed https://discourse.nixos.org/t/flakes-and-filtersource/7787
I have a flake with 2 outputs,a depends on b and b depends on ./.
In theory if I change the nix expression for a the b expression should not rebuild,but it still does as flake.nix is included in ./.
`filterSource` could remove flake from ./. and avoid rebuilding b,but evaluating `filterSource` in a flake keepgiving weird errors.

For `builtins.filterSource` to work inside flake like outside them.

Just keeping the depency on ./. is not as bad as the flake still only includes the files tracked by the git repo,but still keep doing more rebuilds than I'd like when editing `flake.nix`

**Additional context**

tomberek:
Using `builtins.filterSource` and/or `builtins.path` inside of flakes doesn't quite work as expected at the moment. Is this fundamentally a bad idea? There are a few issues about this, but nothing conclusive. This patch allows it, but i have no idea if this is breaking some invariant.

This part seems odd, but seems to be needed:
```
state.realiseContext({});
```

[Discourse](https://discourse.nixos.org/t/flakes-and-filtersource/7787/3)
[Discourse2](https://discourse.nixos.org/t/builtins-filtersource-causes-rebuilds/12564/2)
Fixes #3732
Fixes #2538

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1886,6 +1886,20 @@ static void addPath(EvalState & state, const Pos & pos, const string & name, con
         dstPath = state.store->printStorePath(*expectedStorePath);
 
     mkString(v, dstPath, {dstPath});
+    if (state.allowedPaths)
+        state.allowedPaths->insert(v.string.s);
+
+    try {
+        state.realiseContext({});
+    } catch (InvalidPathError & e) {
+        throw EvalError({
+            .msg = hintfmt("cannot import '%1%', since path '%2%' is not valid", path, e.path),
+            .errPos = pos
+        });
+    } catch (Error & e) {
+        e.addTrace(pos, "while importing '%s'", path);
+        throw e;
+    }
 }
 
 


Mazurel:
I have made a simple [repo](https://github.com/Mazurel/nix-cleaned-path-issue) to test `lib.cleanSource` issue, and it seems that this patch have fixed it.

I do not really know Nix codebase well, although the code looks good to me. The `state.realiseContext({});` should not do anything from what I understand, but based on my tests it is required indeed.

ilkecan:
I think the problems in linked/mentioned issues/topics can be categorized into three:
## 1. Usage of `builtins.filterSource` doesn't prevent a rebuild when a file that didn't pass the filter is updated
- This is because `filterSource` is inherently impure and I think it should be deprecated in favour of `builtins.path`.
- The change in this PR is unrelated to this.
- Both discourse topics belong to this category.

## 2. Using a filter (`filterSource` or `path`) on a string with a context result in `string '<string>' cannot refer to other paths`
- Main implication of this is that the flake inputs can not be filtered, which I think is a huge problem. This can currently be avoided with `builtins.unsafeDiscardStringContext` but ideally it should be fixed within Nix if possible.
- This PR doesn't have any effect on this.
- https://github.com/NixOS/nix/issues/3732 (I think) belongs to this category.

## 3. Trying to access a file in the filtered (using `filterSource` or `path`) directory when `restrict-eval` is enabled results in `access to path '<path>' is forbidden in restricted mode`
- The reason is explained [in here](https://github.com/NixOS/nix/issues/3234#issuecomment-557460549).
- This PR fixes this problem.
- https://github.com/NixOS/nix/issues/2538, https://github.com/NixOS/nix/issues/3234 and https://github.com/Mazurel/nix-cleaned-path-issue are in this category.

tomberek:
There are several motivating use-cases. One significant one is when developing an "internal flake" (ie: one where the flake.nix is in the same repo as the source code, and someone is probably using `src = ./.;` or some variant of that) and modifying the flake.nix invalidates the cached package builds. It would be convenient to filter these cases to avoid repeated rebuilds when the relevant source for the packages is not changing.

Both path and filterSource use the same underlying mechanism:
https://github.com/NixOS/nix/blob/master/src/libexpr/primops.cc#L1915
https://github.com/NixOS/nix/blob/master/src/libexpr/primops.cc#L2002

## Problem # 1
Some of the issues with filterSource is why path was introduced. But i'm not clear on exactly which impurity you (@ilkecan) are talking about.

The baseNameOf default is present in both, but path allows you to set it to remove that impurity. One possibility is to have the default for a directory to use something like "source" instead of baseNameOf?

## Problem # 2
Likely the difference is between using a path and string-with-context, common idioms being: 
- `./. + "/some-dir"`
- `"${some-src}/some-dir}"`

Both use baseNameOf(path) by default. Both path and filterSource check if context is empty:
https://github.com/NixOS/nix/blob/master/src/libexpr/primops.cc#L1900
https://github.com/NixOS/nix/blob/master/src/libexpr/primops.cc#L1971

Should this check also check if the context's paths are in state.allowedPaths rather than just checking if it is empty? (@edolstra ?)

## Problem # 3
If this fixes it: :tada:

ilkecan:
The impurity of `filterSource` I was referring to was the one you are talking about in the next sentence. I created [a PR](https://github.com/NixOS/nix/pull/5286) to add a warning about the usage of `filterSource` to avoid # 1.

\# 2 is about not being able to filter a string with contexts in general. Nix complains since the flake input contains a context. That check might normally be necessary for other cases but I think we should be able to filter a nix input. Your idea to check `state.allowedPaths` sounds interesting.

ilkecan:
Even with this PR, trying to import a file accessed through `${pkgs.path}` produces `access to path '<path>' is forbidden in restricted mode`. 

This can be tested by doing `nix develop` with the following flake:
```
{
  inputs = {
    nixpkgs.url = "nixpkgs/nixos-unstable";
  };

  outputs = { self, nixpkgs, ... }@inputs: {
    devShell.x86_64-linux =
      let
        pkgs = import nixpkgs { system = "x86_64-linux"; };
      in
      pkgs.mkShell {
        shellHook =''
          echo "${(pkgs.callPackage "${toString nixpkgs}/pkgs/development/misc/resholve" {}).resholveScript}"
        '';
      };
  };
}
```
Expected error
`error: cannot coerce a function to a string`
Actual error
`error: access to path '/nix/store/jma6rgslg0pyx3h3r8viymf4l2zqp0xx-7jc0a6qq2w4xjj4p61mspp4s54yniz6m-source/pkgs/development/misc/resholve' is forbidden in restricted mode`

Note that `${pkgs.path}` should probably not be used in this case anyway since Nix will unnecessarily move the directory to another store path. Instead, one of the following should be used
- `${toString pkgs.path}`
- `${nixpkgs}`

edolstra:
@ilkecan I've fixed the `${pkgs.path}` case in 972405edf5ff0d90185fb4021993c4a942787578. But you're right that it's better not to do it this way because it causes an unnecessary copy.

PR #5167, opened 2021-08-23
nix-store --serve: pass on `settings.keepFailed` from SSH store

Ma27:
When doing e.g.

    nix-build -A package --keep-failed --option \
      builders \
      'ssh://mfhydra?remote-store=/home/bosch/store x86_64-linux - 10 4 big-parallel'

this doesn't work properly because this build-setting is ignored.

I changed this behavior by passing the `settings.keepFailed` through the
serve-protocol to remote machines to make sure that I can introspect the
build-directory (which is particularly helpful when I have to look at a
`config.log` from a failed build for instance).

### DIFF
--- a/src/libstore/legacy-ssh-store.cc
+++ b/src/libstore/legacy-ssh-store.cc
@@ -237,6 +237,10 @@ private:
             conn.to
                 << settings.buildRepeat
                 << settings.enforceDeterminism;
+
+        if (GET_PROTOCOL_MINOR(conn.remoteVersion) >= 7) {
+            conn.to << ((int) settings.keepFailed);
+        }
     }
 
 public:
--- a/tests/failing.nix
+++ b/tests/failing.nix
@@ -1,0 +1,22 @@
+{ busybox }:
+with import ./config.nix;
+let
+
+  mkDerivation = args:
+    derivation ({
+      inherit system;
+      builder = busybox;
+      args = ["sh" "-e" args.builder or (builtins.toFile "builder-${args.name}.sh" "if [ -e .attrs.sh ]; then source .attrs.sh; fi; eval \"$buildCommand\"")];
+    } // removeAttrs args ["builder" "meta"])
+    // { meta = args.meta or {}; };
+in
+{
+
+  failing = mkDerivation {
+    name = "failing";
+    buildCommand = ''
+      echo foo > bar
+      exit 1
+    '';
+  };
+}
--- a/src/libstore/serve-protocol.hh
+++ b/src/libstore/serve-protocol.hh
@@ -5,7 +5,7 @@ namespace nix {
 #define SERVE_MAGIC_1 0x390c9deb
 #define SERVE_MAGIC_2 0x5452eecb
 
-#define SERVE_PROTOCOL_VERSION (2 << 8 | 6)
+#define SERVE_PROTOCOL_VERSION (2 << 8 | 7)
 #define GET_PROTOCOL_MAJOR(x) ((x) & 0xff00)
 #define GET_PROTOCOL_MINOR(x) ((x) & 0x00ff)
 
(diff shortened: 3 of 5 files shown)


Ma27:
The build issue on the commit seems not related and the PR's checks seem fine.
cc @edolstra

src/libstore/legacy-ssh-store.cc
```diff
@@ -253,6 +253,10 @@ struct LegacySSHStore : public virtual LegacySSHStoreConfig, public virtual Stor
 
         putBuildSettings(*conn);
 
+        if (GET_PROTOCOL_MINOR(conn->remoteVersion) >= 7) {
+            conn->to << ((int) settings.keepFailed);
+        }
+
```
regnat, line 259: Can’t this go inside `putBuildSettings` ?

tests/failing.nix
```diff
@@ -0,0 +1,22 @@
...
+with import ./config.nix;
+let
+
+  mkDerivation = args:
+    derivation ({
+      inherit system;
+      builder = busybox;
+      args = ["sh" "-e" args.builder or (builtins.toFile "builder-${args.name}.sh" "if [ -e .attrs.sh ]; then source .attrs.sh; fi; eval \"$buildCommand\"")];
+    } // removeAttrs args ["builder" "meta"])
+    // { meta = args.meta or {}; };
```
regnat, line 11: Why do you need all that in addition to the `mkDerivation` that `config.nix` already exports?

regnat (changes requested):

tests/failing.nix
Ma27: The test otherwise fails with

```
    copying 1 paths...
    copying path '\''/nix/store/szbwpiamb2znfgw6iv8c8rr7m5d2b9vz-builder-failing.sh'\'' to '\''ssh://localhost'\''...

    ErroSysError~executing '\''/nix/store/xvvgw9sb8wk6d2c0j3ybn7sll67s3s4z-bash-4.4-p23/bin/bash'\'': No such file or directory
    note: keeping build directory '\''/tmp/nix-shell.2zeviF/nix-build-failing.drv-3'\''
    error: build of '\''/nix/store/i3a2nqjbapjrspr1n1q7f9mmwivdcva6-failing.drv'\'' on '\''ssh://localhost?remote-store=/tmp/nix-shell.2zeviF/nix-test/tests/build-remote-input-addressed/machine1?system-features=foo'\'' failed: error: builder for '\''/nix/store/i3a2nqjbapjrspr1n1q7f9mmwivdcva6-failing.drv'\'' failed with exit code 1
    error: builder for '\''/nix/store/i3a2nqjbapjrspr1n1q7f9mmwivdcva6-failing.drv'\'' failed with exit code 1'
```

All of the `build-remote.sh` tests seem to get their own `$busybox` at the moment and this test uses it rather than `shell` as builder.

src/libstore/legacy-ssh-store.cc
Ma27: Good point, updated!

Ma27:
Hmm, not sure if the broken build is related btw.

Ma27:
cc @regnat @edolstra anything else tbd here? :)

edolstra:
This is a bit risky, since propagating `keep-failed` to remote builders can cause unintentionally filling up the disk of the remote builders. And Nix does not have a mechanism for cleaning up failed temporary directories.

Ma27:
@edolstra I disagree: I can also fill up a remote-store (or a remote-store at a different location to make it slightly harder), but identifying the cause of a full drive is a single `df -h` away IMHO. Also, these issues are usually caught early via either quotas (possible with e.g. ZFS snapshots or a tmpfs with a fixed size) or monitoring. What I want to say is, I don't think the risk of someone filling up a disk isn't necessarily greater than it already is when building remotely.

Ma27:
@edolstra WDYT regarding my response? I actually think, that the change isn't as problematic as you think (but I'm happy to be proven wrong ofc :))

edolstra:
Thanks!

PR #5172, opened 2021-08-25
Set enableOwnership in macOS install

matthewbauer:
For external hard disks where ownership is not enabled by default.

### DIFF
--- a/scripts/create-darwin-volume.sh
+++ b/scripts/create-darwin-volume.sh
@@ -759,6 +759,10 @@ setup_volume() {
 
     await_volume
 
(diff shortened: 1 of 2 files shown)


abathur:
I guess they [aren't behaving quite right when you haven't set up the cachix part](https://github.com/matthewbauer/nix/runs/3416896499?check_suite_focus=true#step:5:12), but domen added basic installer generation and test support earlier this year. With it enabled, you can extract the generated installer URL from each CI run for additional testing.

https://github.com/NixOS/nix/pull/4577 should have enough info to set it up.

PR #5180, opened 2021-08-26
Don’t require experimental features for nix show-config

matthewbauer:
This command is available in nix stable (2.3), so we should allow it
in nix 2.4 too.


edolstra:
> error: unrecognised flag '--experimental-features'

Try `--option experimental-features nix-command`. That should work on both 2.3 and 2.4.

matthewbauer:
That works! Thanks - had not realized that worked.

Issue #5184, opened 2021-08-28
nix from git tries to write to system root at build time: cannot open output file /usr/lib/libnixutil.so: Permission denied


trofi:
When building on a non-nixos for system-wide installation build system fails to write to /usr:

```
$ cd /tmp
$ git clone https://github.com/NixOS/nix.git
$ cd nix
$ ./bootstrap.sh
$ ./configure --prefix=/usr ac_cv_header_aws_s3_S3Client_h=no
$ make
...
$ LANG=C make
  LD     /usr/lib/libnixutil.so
/usr/lib/gcc/x86_64-pc-linux-gnu/12.0.0/../../../../x86_64-pc-linux-gnu/bin/ld: cannot open output file /usr/lib/libnixutil.so: Permission denied
collect2: error: ld returned 1 exit status
make: *** [mk/lib.mk:117: /usr/lib/libnixutil.so] Error 1
```

I think that is not expected.

trofi:
I think it comes from `doc/manual/nix.json` target:

```
$ make -r -d doc/manual/nix.json
...
Considering target file 'doc/manual/nix.json'.
 File 'doc/manual/nix.json' does not exist.
  Considering target file '/usr/bin/nix'.
    Considering target file 'src/build-remote/build-remote.o'.
     Looking for an implicit rule for 'src/build-remote/build-remote.o'.
...
```

I guess the intent was to use local `nix` binary instead of already installed (and outdated) one. Seemingly problematic rule:

```
doc/manual/local.mk:$(d)/nix.json: $(bindir)/nix
```

grawlinson:
Just run into this issue using the 2.4 tarball from GitHub.

Here's build output from `make` on Arch Linux:

```
config.status: creating config.h
  GEN    Makefile.config
  GEN    src/libexpr/parser-tab.cc
  GEN    src/libexpr/parser-tab.hh
  GEN    src/libexpr/lexer-tab.cc
  GEN    src/libexpr/lexer-tab.hh
  GEN    src/nix/get-env.sh.gen.hh
src/libexpr/lexer.l:295: warning, -s option given but default rule can be matched
src/libexpr/lexer.l:295: warning, -s option given but default rule can be matched
  GEN    src/nix-env/buildenv.nix.gen.hh
  GEN    doc/manual/generate-manpage.nix.gen.hh
  GEN    src/nix-channel/unpack-channel.nix.gen.hh
  GEN    doc/manual/utils.nix.gen.hh
[...]

PR #5185, opened 2021-08-28
doc/manual/local.mk: use local src/nix/nix instead of /usr/bin/nix

trofi:
Build failure happens when one tries to build `nix` on a linux
system as user with later plan to install as root (typical package
building separation):

```
$ cd /tmp
$ git clone https://github.com/NixOS/nix.git
$ cd nix
$ ./bootstrap.sh
$ ./configure --prefix=/usr ac_cv_header_aws_s3_S3Client_h=no
$ make
...
$ LANG=C make
  LD     /usr/lib/libnixutil.so
/usr/lib/gcc/x86_64-pc-linux-gnu/12.0.0/../../../../x86_64-pc-linux-gnu/bin/ld: cannot open output file /usr/lib/libnixutil.so: Permission denied
make: *** [mk/lib.mk:117: /usr/lib/libnixutil.so] Error 1
```

Build failure happens because `doc/manual/nix.json` depends
on installed `$(bindir)` path instead of local inplace path:

```
$ make -r -d doc/manual/nix.json
...
Considering target file 'doc/manual/nix.json'.
 File 'doc/manual/nix.json' does not exist.
  Considering target file '/usr/bin/nix'.
    Considering target file 'src/build-remote/build-remote.o'.
     Looking for an implicit rule for 'src/build-remote/build-remote.o'.
...
doc/manual/local.mk:$(d)/nix.json: $(bindir)/nix
```

The fix is to always use inplace built `src/nix/nix` binary.

Closes: https://github.com/NixOS/nix/issues/5184

### DIFF
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -18,7 +18,7 @@ dummy-env = env -i \
 	NIX_SSL_CERT_FILE=/dummy/no-ca-bundle.crt \
 	NIX_STATE_DIR=/dummy
 
-nix-eval = $(dummy-env) $(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
+nix-eval = $(dummy-env) src/nix/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
 
 $(d)/%.1: $(d)/src/command-ref/%.md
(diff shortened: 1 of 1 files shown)


edolstra:
Maybe we should get rid of the default make target, since `make install` is really the only properly supported way to build.

edolstra:
Closing this one in favor of #5145.

PR #5188, opened 2021-08-28
s3-binary-cache: fix build against aws-sdk-cpp with custom memory all…

trofi:
…ocator

When packaging `nix` in Gentoo distribution I encountered a build failure
against packages `aws-sdk-cpp` (enabled custom memory allcoator by default):

```
src/libstore/s3-binary-cache-store.cc: In member function
  'virtual nix::StorePathSet nix::S3BinaryCacheStoreImpl::queryAllValidPaths()':
src/libstore/s3-binary-cache-store.cc:431:32: error:
  no matching function for call to 'Aws::S3::Model::ListObjectsRequest::WithBucket(std::string&)'
  430 |                     Aws::S3::Model::ListObjectsRequest()
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  431 |                     .WithBucket(bucketName)
      |                     ~~~~~~~~~~~^~~~~~~~~~~~
```

The change adds two `s2a`/`a2s` helpers to convert strings from one allocator
scheme to another.

Tested on aws-sdk-cpp with enabled and disabled custom memory allocator.

### DIFF
--- a/src/libstore/s3-binary-cache-store.cc
+++ b/src/libstore/s3-binary-cache-store.cc
@@ -28,6 +28,10 @@ using namespace Aws::Transfer;
 
 namespace nix {
 
+/* Convert string types. They differ only by allocator type. */
+static std::string a2s (const Aws::String & s) { return std::string(std::cbegin(s), std::cend(s)); }
+static Aws::String s2a (const std::string & s) { return Aws::String(std::cbegin(s), std::cend(s)); }
+
 struct S3Error : public Error
 {
     Aws::S3::S3Errors err;
@@ -45,7 +49,7 @@ R && checkAws(const FormatOrString & fs, Aws::Utils::Outcome<R, E> && outcome)
     if (!outcome.IsSuccess())
         throw S3Error(
             outcome.GetError().GetErrorType(),
-            fs.s + ": " + outcome.GetError().GetMessage());
+            fs.s + ": " + a2s(outcome.GetError().GetMessage()));
     return outcome.GetResultWithOwnership();
 }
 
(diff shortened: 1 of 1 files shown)


src/libstore/s3-binary-cache-store.cc
```diff
@@ -28,6 +28,10 @@ using namespace Aws::Transfer;
 
 namespace nix {
 
+/* Convert tring types. They differ only by allocator type. */
```
Mindavi, line 31: nit: `Convert string types`.

trofi: Good catch! Updated.

edolstra:
I'd prefer not to do this since it's an unnecessary conversion and can be quite expensive (e.g. applying `a2s` to the HTTP body doubles the memory allocated). Nix requires building aws-sdk-cpp with `-DCUSTOM_MEMORY_MANAGEMENT=0`.

trofi:
I'll try to do `a2s()` a function without any overhead under `-DCUSTOM_MEMORY_MANAGEMENT=0`. Would you consider such a patch?

trofi:
Oh, I just noticed that use or not use of custom memory management depends on both user's preference and use of shared library:

```
# If building shared libraries, custom memory management enabled is the default, otherwise regular memory management is the default.
# We make custom memory management the default on shared library builds because it is safer and much more difficult to accidentally
# allocate in one DLLs heap while freeing in another (which will lead to runtime crashes)
if(CUSTOM_MEMORY_MANAGEMENT OR (BUILD_SHARED_LIBS AND NOT DEFINED CUSTOM_MEMORY_MANAGEMENT))
    set(USE_AWS_MEMORY_MANAGEMENT ON)
    message(STATUS "Custom memory management enabled; stl objects now using custom allocators")
else()
    set(USE_AWS_MEMORY_MANAGEMENT OFF)
    message(STATUS "Custom memory management disabled")
endif()
```

I'll ask downstream to provide `-DCUSTOM_MEMORY_MANAGEMENT=OFF` option and use that instead.

trofi:
Downstream request: https://bugs.gentoo.org/811108

Issue #4715, opened 2021-04-16
Rename 'nix shell'


edolstra:
The `nix shell` command replaces `nix-shell -p` to provide a shell in which the specified packages are available. However, the name `nix shell` is confusing since there are also *development shells* (created by `nix develop`), which this isn't; and the old `nix-shell` by default does start a development shell. So we should come up with a better name. Some suggestions have been made:

* `nix load nixpkgs#git` - "load" a package into the current shell. This one mainly makes sense in the context of updating the current shell, rather than starting a subshell.
* `nix push nixpkgs#git` / `nix pop` - update the current shell to add a package to the environment (similar to `pushd` / `popd`)
* `nix use nixpkgs#git` - "use" is probably too ambiguous though.
* `nix add nixpkgs#git` - probably sounds too permanent.
* `nix bonk nixpkgs#git` - random name with no connotations that forces people to read the manpage.

nixinator:
nix ininate nixpkgs#git

bit then again, I'm biased... *and* a robot from the future.

raboof:
I actually quite like `nix shell`: it nicely indicates that the command affects the 'current shell', and does not have 'side effects' (using the term loosely). I'm not sure the confusion with development shells is such a problem: it's clear for regular users that `nix shell` is the one they need, and when developers learn about `nix develop` they're likely already familiar with `nix shell`, so that might be smooth enough.

I like push/pop because it makes it clearer you can 'nest' shells, but it's not so clear "what is pushed on what" and where that change is 'visible', so I'm not sure it's a net improvement.

`nix enter` might be another option, as you're 'entering' an environment with different behavior.

andir:
Wasn't this already replaced with `nix run`?

bjornfor:
An idea: adding another command level to disambiguate: `nix add <where>`, i.e. `nix add [new-shell|this-shell|user-profile|global-system]`

I used `add` here, but other verbs also work well with the explicit 3rd argument.

edolstra:
> Wasn't this already replaced with nix run?

No, `nix run` doesn't start a shell, it directly runs the specified app or the default program from a package.

> nix enter might be another option, as you're 'entering' an environment with different behavior.

Maybe `nix enter` could be the subshell variant of `nix load`.

abathur:
I recall joking (to myself? in public? I'd have to go look...) that `nix snowglobe` would be groan-inducingly perfect.

MatthewCroughan:
[...]

Issue #4453, opened 2021-01-15
bash autocompletion for nix shell (nix command)


roberth:
**Is your feature request related to a problem? Please describe.**

It's frustrating when autocompletion doesn't work in a shell.

**Describe the solution you'd like**

Simple option: add entries to `XDG_DATA_DIRS`.

Complicated option: make `nix shell` use stdenv, which can take care of this logic, at the cost of a Nixpkgs dependency.

**Describe alternatives you've considered**

Write GUIs for everything that shouldn't have a GUI.

**Additional context**

Completions are available for derivation-based shells since https://github.com/NixOS/nixpkgs/pull/103501

`nix-shell` also needs adjustment, see #4452

roberth:
Another way to make this work is by using `passthru` to inform how a package should added to the environment variables. See also [`env` noun suggestion](https://github.com/NixOS/nix/issues/4715#issuecomment-875729369)

This is similar to what the `stdenv` machinery does in a package build, but simplified and specifically for use _outside_ a package build.
Perhaps this should be incorporated into the "app" concept or vice versa.

roberth:
Basically what I'm suggesting is to treat "environment variables" as a serious deployment target, where a package can tell Nix exactly what to do, or rely on the default (`PATH` + possibly `XDG_DATA_DIRS`)

Something like

```nix
passthru.useViaEnvironment = {
  PATH = "${self.outPath}/bin:${commonlyUsedRuntimeDep}/bin";
  XDG_DATA_DIRS = "${self.outPath}/share/foo";
  shellHook = "echo To initialize the foo package config, run foo init";
}
```

just to illustrate the concept.

I'm assuming https://github.com/NixOS/nixpkgs/pull/119942 for simplicity on the Nix side of this interface. The self-referential nature of this feature can be worked around with placeholders, but I would prefer to exploit the existing language instead.

How the exactly the environment variables should be merged isn't obvious. Nix could hardcode some of them. Nixpkgs doesn't know how to do it yet either https://github.com/NixOS/nixpkgs/issues/122244.

SuperSandro2000:
Is this similar to https://github.com/nix-community/lorri/pull/61 ?

roberth:
@SuperSandro2000 yes, it's all necessary to make completions work beyond simple nix-shell + stdenv (https://github.com/NixOS/nixpkgs/pull/103501)

roberth:
[...]

Issue #801, opened 2016-02-14
a flag to force shebang mode


aespinosa:
I've been having trouble to load my default.nix environment for my Jenkins jobs. My script is being interpreted as nix code instead of the interpreter I specified:

```
[workspace] $ /usr/bin/env nix-shell /var/folders/bh/q33bp93j2js6zv27rds0n1x00000gn/T/hudson5799651993425365822.sh
error: undefined variable ‘jupyter’ at /var/folders/bh/q33bp93j2js6zv27rds0n1x00000gn/T/hudson5799651993425365822.sh:4:1
```

Here's the Jenkins 'Execute Shell' build step:

```
#!/usr/bin/env nix-shell
#!nix-shell -i bash

jupyter --version
```

I would expect to have something like 'jupyter: command not found' if the interpreter being launched was bash.

I think having a flag to force interpretation will be good to remove some ambiguities.

tomberek:
Closing as stale. Re-open if needed.

PR #5189, opened 2021-08-28
Shebang support with flakes

tomberek:
Enables shebang usage of nix shell. All arguments with `#! nix` get
added to the nix invocation. This implementation does NOT set any
additional arguments other than placing the script path itself as the
first argument such that the interpreter can utilize it.

Example below:

```
    #!/usr/bin/env nix
    #! nix shell --quiet
    #! nix nixpkgs#bash
    #! nix nixpkgs#shellcheck
    #! nix nixpkgs#hello
    #! nix --ignore-environment --command bash
    # shellcheck shell=bash
    set -eu
    shellcheck "$0" || exit 1
    function main {
        hello
        echo 0:"$0" 1:"$1" 2:"$2"
    }
    "$@"
```

Overall it should make sense to any users of `nix-shell`.

- [x] add documentation
- [x] more tests
- [ ] decide on defaults
    - [ ] automatically inject "shell" argument? (edit: `nix run` would work if it added PATHs)
    - [ ] ~re-use the "-i interpreter" concept from nix-shell?~
    - [x] lock file options?
    - [ ] whitelist/blacklist certain options
- [x] remove duplication of heuristic logic/code with `nix-build`
- [ ] support a bootstrap mechanism detecting if nix is installed and fetching otherwise (eg nix —store or nix-portable)?
- [ ] merge

# Checklist for maintainers

 - [x] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests

### DIFF
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -105,7 +105,8 @@ void Args::parseCmdline(const std::string & programName, const Strings & _cmdlin
 
                 std::string line;
                 std::getline(stream,line);
-                while (std::getline(stream,line) && !line.empty()){
+                std::string commentChars("#/\\%@*-");
+                while (std::getline(stream,line) && !line.empty() && commentChars.find(line[0]) != std::string::npos){
                     line = chomp(line);
 
                     std::smatch match;
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -1,6 +1,8 @@
 #include "args.hh"
 #include "hash.hh"
 
+#include <fstream>
+#include <string>
 #include <regex>
 #include <glob.h>
 
@@ -93,14 +95,17 @@ void Args::parseCmdline(const std::string & programName, const Strings & _cmdlin
     if (isNixCommand && cmdline.size() > 0) {
         auto script = *cmdline.begin();
         try {
-            auto lines = tokenizeString<Strings>(readFile(script), "\n");
-            if (std::regex_search(lines.front(), std::regex("^#!"))) {
-                lines.pop_front();
+            std::ifstream stream(script);
+            char shebang[3]={0,0,0};
+            stream.get(shebang,3);
+            if (strncmp(shebang,"#!",2) == 0){
                 for (auto pos = std::next(cmdline.begin()); pos != cmdline.end();pos++)
                     savedArgs.push_back(*pos);
                 cmdline.clear();
 
-                for (auto line : lines) {
+                std::string line;
+                std::getline(stream,line);
+                while (std::getline(stream,line) && !line.empty()){
                     line = chomp(line);
 
                     std::smatch match;
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -1,6 +1,7 @@
 #include "args.hh"
 #include "hash.hh"
 
+#include <regex>
 #include <glob.h>
 
 #include <nlohmann/json.hpp>
@@ -63,6 +64,12 @@ static std::optional<std::string> needsCompletion(std::string_view s)
 
 void Args::parseCmdline(const Strings & _cmdline)
 {
+    // Default via 5.1.2.2.1 in C standard
+    Args::parseCmdline("", _cmdline);
+}
+
+void Args::parseCmdline(const std::string & programName, const Strings & _cmdline)
+{
     Strings pendingArgs;
     bool dashDash = false;
 
@@ -77,6 +84,36 @@ void Args::parseCmdline(const Strings & _cmdline)
     }
 
     bool argsSeen = false;
+
+    // Heuristic to see if we're invoked as a shebang script, namely,
+    // if we have at least one argument, it's the name of an
+    // executable file, and it starts with "#!".
+    Strings savedArgs;
+    auto isNixCommand = std::regex_search(programName, std::regex("nix$"));
+    if (isNixCommand && cmdline.size() > 0) {
+        auto script = *cmdline.begin();
+        try {
+            auto lines = tokenizeString<Strings>(readFile(script), "\n");
+            if (std::regex_search(lines.front(), std::regex("^#!"))) {
+                lines.pop_front();
+                for (auto pos = std::next(cmdline.begin()); pos != cmdline.end();pos++)
+                    savedArgs.push_back(*pos);
+                cmdline.clear();
+
+                for (auto line : lines) {
+                    line = chomp(line);
+
+                    std::smatch match;
+                    if (std::regex_match(line, match, std::regex("^#!\\s*nix\\s(.*)$")))
+                        for (const auto & word : shellwords(match[1].str()))
+                            cmdline.push_back(word);
+                }
+                cmdline.push_back(script);
+                for (auto pos = savedArgs.begin(); pos != savedArgs.end();pos++)
+                    cmdline.push_back(*pos);
+            }
+        } catch (SysError &) { }
+    }
     for (auto pos = cmdline.begin(); pos != cmdline.end(); ) {
 
         auto arg = *pos;
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -67,10 +67,10 @@ static std::optional<std::string> needsCompletion(std::string_view s)
 void Args::parseCmdline(const Strings & _cmdline)
 {
     // Default via 5.1.2.2.1 in C standard
-    Args::parseCmdline("", _cmdline);
+    Args::parseCmdline(_cmdline, false);
 }
 
-void Args::parseCmdline(const std::string & programName, const Strings & _cmdline)
+void Args::parseCmdline(const Strings & _cmdline, bool allowShebang)
 {
     Strings pendingArgs;
     bool dashDash = false;
@@ -91,8 +91,7 @@ void Args::parseCmdline(const std::string & programName, const Strings & _cmdlin
     // if we have at least one argument, it's the name of an
     // executable file, and it starts with "#!".
     Strings savedArgs;
-    auto isNixCommand = std::regex_search(programName, std::regex("nix$"));
-    if (isNixCommand && cmdline.size() > 0) {
+    if (allowShebang){
         auto script = *cmdline.begin();
         try {
             std::ifstream stream(script);
@@ -105,7 +104,7 @@ void Args::parseCmdline(const std::string & programName, const Strings & _cmdlin
 
                 std::string line;
                 std::getline(stream,line);
-                std::string commentChars("#/\\%@*-");
+                static const std::string commentChars("#/\\%@*-");
                 while (std::getline(stream,line) && !line.empty() && commentChars.find(line[0]) != std::string::npos){
                     line = chomp(line);
 
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -14,6 +14,7 @@
 #include <future>
 #include <iostream>
 #include <mutex>
+#include <regex>
 #include <sstream>
 #include <thread>
 
@@ -1459,6 +1460,49 @@ std::string shellEscape(const std::string_view s)
     return r;
 }
 
+/* Recreate the effect of the perl shellwords function, breaking up a
+ * string into arguments like a shell word, including escapes
+ */
+std::vector<std::string> shellwords(const std::string & s)
+{
+    std::regex whitespace("^(\\s+).*");
+    auto begin = s.cbegin();
+    std::vector<std::string> res;
+    std::string cur;
+    enum state {
+        sBegin,
+        sQuote
+    };
+    state st = sBegin;
+    auto it = begin;
+    for (; it != s.cend(); ++it) {
+        if (st == sBegin) {
+            std::smatch match;
+            if (regex_search(it, s.cend(), match, whitespace)) {
+                cur.append(begin, it);
+                res.push_back(cur);
+                cur.clear();
+                it = match[1].second;
+                begin = it;
+            }
+        }
+        switch (*it) {
+            case '"':
+                cur.append(begin, it);
+                begin = it + 1;
+                st = st == sBegin ? sQuote : sBegin;
+                break;
+            case '\\':
+                /* perl shellwords mostly just treats the next char as part of the string with no special processing */
+                cur.append(begin, it);
+                begin = ++it;
+                break;
+        }
+    }
+    cur.append(begin, it);
+    if (!cur.empty()) res.push_back(cur);
+    return res;
+}
 
 void ignoreException(Verbosity lvl)
 {
--- a/src/libutil/util.hh
+++ b/src/libutil/util.hh
@@ -695,7 +695,7 @@ std::string shellEscape(const std::string_view s);
 
 
 /* Recreate the effect of the perl shellwords function, breaking up a
- * string into arguments like a shell word, including escapes */
+   string into arguments like a shell word, including escapes. */
 std::vector<std::string> shellwords(const std::string & s);
 
 
--- a/src/libutil/util.hh
+++ b/src/libutil/util.hh
@@ -694,10 +694,13 @@ std::string toLower(const std::string & s);
 std::string shellEscape(const std::string_view s);
 
 
-/**
- * Exception handling in destructors: print an error message, then
- * ignore the exception.
- */
+/* Recreate the effect of the perl shellwords function, breaking up a
+ * string into arguments like a shell word, including escapes */
+std::vector<std::string> shellwords(const std::string & s);
+
+
+/* Exception handling in destructors: print an error message, then
+   ignore the exception. */
 void ignoreException(Verbosity lvl = lvlError);
 
 
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -29,50 +29,6 @@ using namespace std::string_literals;
 
 extern char * * environ __attribute__((weak));
 
-/* Recreate the effect of the perl shellwords function, breaking up a
- * string into arguments like a shell word, including escapes
- */
-static std::vector<std::string> shellwords(const std::string & s)
-{
-    std::regex whitespace("^(\\s+).*");
-    auto begin = s.cbegin();
-    std::vector<std::string> res;
-    std::string cur;
-    enum state {
-        sBegin,
-        sQuote
-    };
-    state st = sBegin;
-    auto it = begin;
-    for (; it != s.cend(); ++it) {
-        if (st == sBegin) {
-            std::smatch match;
-            if (regex_search(it, s.cend(), match, whitespace)) {
-                cur.append(begin, it);
-                res.push_back(cur);
-                cur.clear();
-                it = match[1].second;
-                begin = it;
-            }
-        }
-        switch (*it) {
-            case '"':
-                cur.append(begin, it);
-                begin = it + 1;
-                st = st == sBegin ? sQuote : sBegin;
-                break;
-            case '\\':
-                /* perl shellwords mostly just treats the next char as part of the string with no special processing */
-                cur.append(begin, it);
-                begin = ++it;
-                break;
-        }
-    }
-    cur.append(begin, it);
-    if (!cur.empty()) res.push_back(cur);
-    return res;
-}
-
 static void main_nix_build(int argc, char * * argv)
 {
     auto dryRun = false;
--- a/tests/flakes/flakes.sh
+++ b/tests/flakes/flakes.sh
@@ -69,8 +69,9 @@ cat > $nonFlakeDir/shebang.sh <<EOF
 #! nix --offline shell
 #! nix flake1#fooScript
 #! nix --no-write-lock-file --command bash
-set -e
+set -ex
 foo
+echo "\$@"
 EOF
 chmod +x $nonFlakeDir/shebang.sh
 
@@ -517,3 +518,4 @@ expectStderr 1 nix flake metadata $flake2Dir --no-allow-dirty --reference-lock-f
 
 # Test shebang
 [[ $($nonFlakeDir/shebang.sh) = "foo" ]]
+[[ $($nonFlakeDir/shebang.sh "bar") = "foo"$'\n'"bar" ]]
--- a/tests/flakes/flakes.sh
+++ b/tests/flakes/flakes.sh
@@ -78,6 +78,18 @@ chmod +x $nonFlakeDir/shebang.sh
 git -C $nonFlakeDir add README.md shebang.sh
 git -C $nonFlakeDir commit -m 'Initial'
 
+cat > $nonFlakeDir/shebang-perl.sh <<EOF
+#! $(type -P env) nix
+#! nix run nixpkgs#perl
+## Modules used
+use strict;
+use warnings;
+
+# Print function
+print("Hello World\n");
+EOF
+chmod +x $nonFlakeDir/shebang-perl.sh
+
 # Construct a custom registry, additionally test the --registry flag
 nix registry add --registry $registry flake1 git+file://$flake1Dir
 nix registry add --registry $registry flake2 git+file://$flake2Dir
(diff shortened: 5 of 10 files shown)


roberth:
I really like this. I didn't think we could use `nix` because portable `/usr/bin/env` doesn't support extra arguments, but you've proven me wrong :tada.

My only objection as far as the design is concerned, is that this isn't about shells at all, but on closer reading, this isn't specific to the [dubious `nix shell`](https://github.com/NixOS/nix/issues/4715) command at all. In fact, this seems to solve #801.

src/libutil/args.cc
```diff
@@ -69,6 +76,39 @@ void Args::parseCmdline(const Strings & _cmdline)
...
+
+    // Heuristic to see if we're invoked as a shebang script, namely,
+    // if we have at least one argument, it's the name of an
+    // executable file, and it starts with "#!".
+    Strings savedArgs;
+    auto runEnv = std::regex_search(programName, std::regex("nix$"));
+    if (runEnv && cmdline.size() > 0) {
+        auto script = *cmdline.begin();
+        try {
+            auto lines = tokenizeString<Strings>(readFile(script), "\n");
```
regnat, line 88: I don’t know how much of an issue that is in practice, but this code reads the whole file, even when it bails right after because it doesn’t start by a shebang. Maybe you could use an `FdSource` or similar instead to only read as much as is needed

src/libutil/args.cc
```diff
@@ -69,6 +76,39 @@ void Args::parseCmdline(const Strings & _cmdline)
...
+                lines.pop_front();
+                for (auto pos = std::next(cmdline.begin()); pos != cmdline.end();pos++)
+                    savedArgs.push_back(*pos);
+                cmdline.clear();
+
+                for (auto line : lines) {
+                    line = chomp(line);
+
+                    std::smatch match;
+                    if (std::regex_match(line, match, std::regex("^#!\\s*nix (.*)$")))
```
regnat, line 99:
```suggestion
                    if (std::regex_match(line, match, std::regex("^#!\\s*nix\\s(.*)$")))
```

regnat (commented):
A couple of comments, but I really like the overall idea.

>  automatically inject "shell" argument?

I have a mixed feeling on that one: On the one hand, the only other command that (I think) could make sense in a shebang is `nix develop`, and that’s really an edge-case. On the other hand, having a too different behaviour when Nix is invoked as a shebang would probably be quite confusing (“why is `./myScript` working but `nix myScript` returning some nonsense?”)

> whitelist/blacklist certain options

I don’t think that would really be useful. At least from a security pov, you’re running an arbitrary script anyways, so restricting the interpreter wouldn’t change a lot.

> support a bootstrap mechanism detecting if nix is installed and fetching otherwise (eg nix —store or nix-portable)?

How could that work? If you don’t have Nix installed, you can’t run the shebang at all afaik

roberth:
To support `nix develop`, you could override the various `basePath`/`baseDir` variables to point to the script's location. That way its functioning does not depend on the caller's working directory, which could be anywhere.

tomberek:
Another issue that has come up as I'm writing the docs (stealing most of it from nix-shell), is that we don't run the setup or shell hooks to modify PYTHONPATH or PERL5LIB and so forth. Not sure if we want to adopt exactly the legacy nix-shell behavior which led us down an convoluted path. Is there an easier way?

roberth:
> we don't run the setup or shell hooks

This is why we need to[ rename](https://github.com/NixOS/nix/issues/4715) the current `nix shell` to something like [`nix env run`](https://github.com/NixOS/nix/issues/4715#issuecomment-875729369).
If you want to run stdenv, currently your only `nix-command` option is `nix develop`.

I would like Nix to [defer to packages](https://github.com/NixOS/nix/issues/4453#issuecomment-875747554), allowing it to take "installing to environment variables" seriously. Maybe it's overkill because for regular commands `PATH`+`XDG_DATA_DIRS` will generally do, but perhaps this idea can be used for more complicated stdenv-like environment composition. Assuming Nix can somehow learn how to compose variables. (Not everything is `:`-separated)

Previously [my (and apparently Eelco's) assumption was](https://github.com/NixOS/nix/pull/4452#issuecomment-760894468) that we don't need the stdenv/setup-based ad-hoc shells like you are asking for. What's your use case? For packaging we've assumed writing an expression is best.

Anyway that's all fairly orthogonal to the shebang implementation, as long as we don't

>> automatically inject "shell" argument

I hope I was able to provide some context.

src/libutil/args.cc
tomberek: Something like this? https://github.com/tomberek/nix/commit/cf02a4f1a5024369836e28082542011554def798

Could not see how to use FdSource for this use case.

roosemberth:
```
    #!/usr/bin/env nix
    #! nix shell --quiet
    #! nix nixpkgs#bash
    #! nix nixpkgs#shellcheck
```

I find utterly confusing references to derivations are written as `#! nix f#pkg` instead of `#! nix shell f#pkg` or simply `#! f#pkg`.

I think we should also support a varadic derivation names until the newline (e.g.  `#! nix shell f#pkg1 f#pkg2`) like `nix shell` already does.

roberth:
The `#! nix` has the same role as `#!nix-shell` which seemed to have served a technical purpose for potentially mixing shebang aware interpreters. I wonder to what extend this is required. Perhaps it could be optional? For example, we could strip either  `#!` or `#!nix`, whichever is a longer prefix. That way, we only need to confuse the reader when it's necessary for technical reasons.

Note that `#! nix` with the space would then be parsed as a `#!` prefix and a `nix` argument instead of `#! nix` as a prefix. I think this is an improvement, because `#!nix` reads a bit more like a single keyword.

roosemberth:
I agree that `#!nix` would be much more idiomatic. I think making both `#!` and `#! nix` valid will just create FUD for no reason.

Do you have examples of shebang-aware interpreters that'd play nice with nix shebangs as described in this document?
I've never seen them in the wild (doesn't mean they don't exist), but the idea of supporting bot cases does seem a bit feature-creep to me.

On 2 September 2021 12:45:22 CEST, Robert Hensing ***@***.***> wrote:
>The `#! nix` has the same role as `#!nix-shell` which seemed to have served a technical purpose for potentially mixing shebang aware interpreters. I wonder to what extend this is required. Perhaps it could be optional? For example, we could strip either  `#!` or `#!nix`, whichever is a longer prefix. That way, we only need to confuse the reader when it's necessary for technical reasons.
>
>Note that `#! nix` with the space would then be parsed as a `#!` prefix and a `nix` argument instead of `#! nix` as a prefix. I think this is an improvement, because `#!nix` reads a bit more like a single keyword.
>
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/NixOS/nix/pull/5189#issuecomment-911524500
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tomberek:
The proposed message is variadic and can be put all on one line. Sometimes these lines end up very long and having them split up can be helpful. 

As to the prefix, this was meant to copy the way nix-shell does it( most of the code was simply copied or adopted over). 

There are several shebang-aware interpreters, eg perl. I’m sure there are more. 

The biggest issue is being unable to load the stdenv in the “right way”. I suspect that is actually what people will want. The examples of legacy nix-shell show bringing in python and Perl libraries and that will only work if we run shellHooks or at least do more than just modify PATH.

ymeister:
Forgive me if I am wrong, but is it really necessary considering `env -S` option:
```
-S, --split-string=S
              process and split S into separate arguments; used to pass multiple arguments on shebang lines
```

That way your example becomes:
```
#!/usr/bin/env -S nix shell --quiet nixpkgs#bash nixpkgs#shellcheck nixpkgs#hello --ignore-environment --command bash
# shellcheck shell=bash
set -eu
shellcheck "$0" || exit 1
function main {
    hello
    echo 0:"$0" 1:"$1" 2:"$2"
}
"$@"
```

Aside from multi-line support, I don't see any problem here.

roberth:
@ymeister that's a good suggestion, but it is not universally supported on the other unixes. It also doesn't allow spaces in arguments for use with `--expr`.

tomberek:
> Forgive me if I am wrong, but is it really necessary considering `env -S` option

The original `nix-shell` shebang mechanism did not rely on that feature - i presume for various compatibility reasons. If one wishes to use it for their own scripts, they are welcome to do so. It should work with this PR (if not, I'd consider that a bug).

My greater concern is that examples such as (https://github.com/NixOS/nix/blob/7ab3bc32b139e251ff8d9abf25d6cadf3c0c6945/doc/manual/src/command-ref/nix-shell.md#use-as-a--interpreter) won't work with `nix shell` as it is currently implemented. Thus this is involved with the whole `nix-shell`/`nix shell`/`nix develop`/`nix run`/`nix env do-a-thing` design and debate. My summary of the situation is that there are a few use cases:

1) Please put some packages A, B, C into my PATH. - usually `nix-shell -p somepkg` now `nix shell`
2) I want to develop on package A. - usually `nix-shell -A somepkg` now `nix develop`
3) Please make language-specific libraries available: eg: PYTHONPATH, PERL5LIB... - usually `nix-shell -p` or `nix-shell shell.nix`
4) Just run a program, don't change any env vars -  `nix run`

To make (1)(3) possible `nix-shell` would create a derivation behind the scenes on your behalf, composing it together and additionally invoking stdenv environment setup and shell hooks. This is why `nix shell nixpkgs#python3 nixpkgs#python3Packages.requests` does not work as expected. The shebang example from this PR hits this issue. The only way to express "please run the shellHooks" is to create your own flake devShell, but ad-hoc CLI mechanisms don't seem to allow it.

#### Another approach
- Can we add "multiple installables" support to `nix develop`? Then `nix develop` is always the one that brings in stdenv/shellHooks.

This would create three distinct tools:
- `nix run` - only runs an executable, no changes to env vars, multiple installables makes no sense
- `nix shell` - only adds things to PATH, multiple installables means to add them all to the PATH
- `nix develop` - add things to PATH and run shell hooks. multiple installables means to do a "buildEnv" composition.

ymeister:
@roberth that's a good notice. Albeit the wide adoption of `env -S` among other unixes, it doesn't seem to be POSIX-compliant, so that's a bummer. On another note, I haven't experienced any problem using it with `--expr`:
```
#! /usr/bin/env -S nix shell --impure --expr "with import <nixpkgs> {}; python.withPackages (pkgs: with pkgs; [ prettytable ])" -c python

import prettytable

# Print a simple table.
t = prettytable.PrettyTable(["N", "N^2"])
for n in range(1, 10): t.add_row([n, n * n])
print t
```

(`with import <nixpkgs> {}` requires `nixpkgs` flake path in `$NIX_PATH`. That can be achieved either by pinning it system-wide (NixOS), or by exporting `NIX_PATH="nixpkgs=$(nix flake metadata nixpkgs --json | jq -r .path)"`)

Above code also showcases an example of a workaround of the issue discussed by @tomberek. Although I agree that it is just a workaround and is nowhere near a proper solution like one proposed by @tomberek.

roberth:
I may have been wrong about `-S` with quotes. I only read the man page, not the `info` page (:confused:) which goes into more detail, confirming your observations for at least GNU env and FreeBSD.
Instead of `<nixpkgs>` you should be able to reference a local flake when the base path is passed correctly. Also note that `nix shell --expr` is very different from `nix-shell --expr` (without and with stdenv respectively; also use case (1) vs (2))

Artturin:
Please make sure these 2 keep working, Thanks.

```rust
#!/usr/bin/env nix-shell
//! ```cargo
//! [dependencies]
//! time = "0.1.25"
//! ```
/*
#!nix-shell -i rust-script -p rustc -p rust-script -p cargo
*/
fn main() {
    for argument in std::env::args().skip(1) {
        println!("{}", argument);
    };
    println!("{}", std::env::var("HOME").expect(""));
    println!("{}", time::now().rfc822z());
}
// vim: ft=rust
```
```rust
#!/usr/bin/env nix-shell
#![allow()] /*
#!nix-shell -i bash -p rustc
rsfile="$(readlink -f $0)"
binfile="/tmp/$(basename "$rsfile").bin"
rustc "$rsfile" -o "$binfile" --edition=2021 && exec "$binfile" $@ || exit $?
*/
fn main() {
    for argument in std::env::args().skip(1) {
        println!("{}", argument);
    };
    println!("{}", std::env::var("HOME").expect(""));
}
// vim: ft=rust
```

chvp:
For anyone stumbling onto this in the future, you can avoid having to set NIX_PATH by using the following (as roberth hinted):
```
#!/usr/bin/env -S nix shell --impure --expr "(import (builtins.getFlake \"nixpkgs\") {}).python3.withPackages (ps: [ ps.requests ])" --command python
```

src/libutil/args.cc
tomberek: Is that approach somewhat sane in a C++ world?

tomberek:
This has been in draft for too long. I'm going to accept for now that multiple `nix develop` installables is an independent problem and proceed.
@Artturin your examples look like this:
```
#!/usr/bin/env nix
#![allow()] /*
#!nix shell nixpkgs#rustc --command bash
rsfile="$(readlink -f $0)"
binfile="/tmp/$(basename "$rsfile").bin"
rustc "$rsfile" -o "$binfile" --edition=2021 && exec "$binfile" $@ || exit $?
*/
fn main() {
    for argument in std::env::args().skip(1) {
        println!("{}", argument);
    };
    println!("{}", std::env::var("HOME").expect(""));
}
// vim: ft=rust
```

```
#!/usr/bin/env nix
//! ```cargo
//! [dependencies]
//! time = "0.1.25"
//! ```
/*
#!nix shell nixpkgs#rustc nixpkgs#rust-script nixpkgs#cargo --command rust-script
*/
fn main() {
    for argument in std::env::args().skip(1) {
        println!("{}", argument);
    };
    println!("{}", std::env::var("HOME").expect(""));
    println!("{}", time::now().rfc822z());
}
// vim: ft=rust
```

src/libutil/args.cc
roberth: Looks alright.
The `#!` check is a bit complicated, but at least it won't suffer when the first "line" is very long (e.g. not a text file). When it looks like a shebang, you'll still want to ignore the whole line though.

src/nix-build/nix-build.cc
```diff
@@ -29,50 +29,6 @@ using namespace std::string_literals;
...
-            }
-        }
-        switch (*it) {
-            case '"':
-                cur.append(begin, it);
-                begin = it + 1;
-                st = st == sBegin ? sQuote : sBegin;
-                break;
-            case '\\':
-                /* perl shellwords mostly just treats the next char as part of the string with no special processing */
```
roberth, line 65: Off-topic: did we stop taking perl as an example?

tests/flakes/flakes.sh
```diff
@@ -468,3 +512,243 @@ nix store delete $(nix store add-path $badFlakeDir)
 
 [[ $(nix path-info      $(nix store add-path $flake1Dir)) =~ flake1 ]]
 [[ $(nix path-info path:$(nix store add-path $flake1Dir)) =~ simple ]]
+  outputs = { self, nixpkgs, b }: {
```
roberth, line 515: I love to see more test cases, but this doesn't look right. Did a rebase go wrong?

tomberek: fixed

src/libutil/args.cc
```diff
@@ -69,6 +76,36 @@ void Args::parseCmdline(const Strings & _cmdline)
     }
 
     bool argsSeen = false;
+
+    // Heuristic to see if we're invoked as a shebang script, namely,
```
edolstra, line 80: Since this is very similar to the code in nix-build.cc, it would be nice to factor out the commonality.

edolstra (changes requested):
Looks good, but it needs an entry in the release notes and documentation (probably in the `nix shell` manpage, see the `Use as a #!-interpreter` section in the `nix-shell` manpage as a basis).

It would be good to add a test case for `nix run`, since you may want to write
```
#! /usr/bin/env nix run nixpkgs#perl
... perl code ...
```

src/libutil/util.hh
```diff
@@ -526,6 +526,11 @@ std::string toLower(const std::string & s);
 std::string shellEscape(const std::string_view s);
 
 
+/* Recreate the effect of the perl shellwords function, breaking up a
+ * string into arguments like a shell word, including escapes */
```
edolstra, line 530:
```suggestion
   string into arguments like a shell word, including escapes. */
```

grahamc:
I wonder if we should review this a bit more carefully, since it may not work for common use cases like shells with python packages.

tests/flakes/flakes.sh
```diff
@@ -468,3 +478,7 @@ nix store delete $(nix store add-path $badFlakeDir)
 
 [[ $(nix path-info      $(nix store add-path $flake1Dir)) =~ flake1 ]]
 [[ $(nix path-info path:$(nix store add-path $flake1Dir)) =~ simple ]]
+
+# Test shebang
+[[ $($nonFlakeDir/shebang.sh) = "foo" ]]
```
roberth, line 483: Arguments must be passed to the script (and the script should echo it back to make the test work)
Something like:

```suggestion
[[ $($nonFlakeDir/shebang.sh "foo bar") = "arg[1] was foo bar" ]]
```

roberth (commented):
Are script arguments supported?

tests/flakes/flakes.sh
tomberek: Yes, args are passed along in the expected way. Updated the test. Thanks.

tomberek:
> Are script arguments supported?

Yes

src/libutil/args.cc
roberth: I like https://github.com/tomberek/nix/commit/cf02a4f1a5024369836e28082542011554def798.
It would also be nice to have a way to have an explicit final nix shebang lines, so that we can stop scanning. This would allow for huge files to start with a nix shebang, which is useful for installers. Perhaps a line with `#! nix-end-of-arguments` or something?

src/libutil/args.cc
```diff
@@ -69,6 +76,36 @@ void Args::parseCmdline(const Strings & _cmdline)
...
+                    savedArgs.push_back(*pos);
+                cmdline.clear();
+
+                for (auto line : lines) {
+                    line = chomp(line);
+
+                    std::smatch match;
+                    if (std::regex_match(line, match, std::regex("^#!\\s*nix\\s(.*)$")))
+                        for (const auto & word : shellwords(match[1].str()))
+                            cmdline.push_back(word);
```
tomberek, line 101: @roberth i considered an “else, break out of scanning” condition here.

src/libutil/args.cc
tomberek: Another issue I’ve sometime hit is the desire to have a bit of shell parsing of the command line, or access to a SOURCE_SCRIPT var. (it’s an XY problem situation, I’m still trying to support perlPackages, etc)

src/libutil/args.cc
tomberek: What would you think about doing a regex check for `^[^#\/\\\*-]` and breaking out?

roberth: That might be a bit aggressive.
The Rust ones here are already quite interesting https://nixos.wiki/wiki/Nix-shell_shebang
I think I've seen a `stack run` + `nix-shell` one once, which was pretty weird.
Your suggestion does happen to work, but I can't help feeling that it's by coincidence.
Perhaps breaking out on the first empty line would be sensible enough? I think those only tend to occur after the shebang lines. I could be wrong, and it'd be an easy mistake to add a newline between two tools' shebang stuff. I guess Nix would be the first one, so it all works out?

Worst case we can add extra instructions to let the script configure shebang scanning later.

fricklerhandwerk:
- @thufschmitt: have to discuss this, the design space is pretty broad here

fricklerhandwerk:

- @roberth: @tomberek is not here, so let's keep it short
- @edolstra: `nix shell` is not sufficient for things like perl with perl packages
- @roberth: `nix develop` is supported
- @edolstra: need a `nix-shell -p` equivalent
  - @roberth: this is mostly an orthogonal problem, not sure if I said that in the meeting
- @roberth: I don't see a `nix develop` or `nix run` test case (anymore?? not sure)

tomberek:
> Discussed in Nix team meeting 2022-12-23:
> 
> * @roberth: @tomberek is not here, so let's keep it short

Apologies.

> * @edolstra: `nix shell` is not sufficient for things like perl with perl packages
> * @roberth: `nix develop` is supported
> * @edolstra: need a `nix-shell -p` equivalent
>   
>   * @roberth: this is mostly an orthogonal problem, not sure if I said that in the meeting

The current design is doing the bare minimum and merely interprets the command line in a specific way. It therefore supports any nix command. This might be too powerful, but to help explain:
```
#!/usr/bin/env nix
#! nix eval
#! nix nixpkgs#hello --json --option dummy
# the dummy option parameter is to swallow the file arg that is passed at the end
```
or 
```
#!/usr/bin/env nix
#! nix store add-file
when run, this script will add itself into the store
```
But I don't think these are the priority.


> * @roberth: I don't see a `nix develop` or `nix run` test case (anymore?? not sure)

The issue about `develop` supporting perlPackages is that one cannot compose the environment you want without expressing a function call in Nix (buildEnv, mkDerivation, withPackages, etc). We can interpret multiple installables for the develop command to have a specific composition, or use --apply, or something else. This lack of support is not shebang specific, but is about develop, and any fix for it is easily adopted by the shebang support.

`nix run` support is largely subsumed by `nix shell --command bash`, but this brings up a good question. One might want to support this:
```
#!/usr/bin/env nix
#! nix run
#! nix nixpkgs#hello --option dummy
```
But the file parameter is a bit annoying, should we have a form of the shebang that does NOT append the own-file-name to the call?
```
#!/usr/bin/env nix
#! nix run nixpkgs#hello        # comments are ignored and will swallow the own-file-name?
```
I'm open to ideas.

src/libutil/args.cc
```diff
@@ -69,6 +78,40 @@ void Args::parseCmdline(const Strings & _cmdline)
     }
 
     bool argsSeen = false;
+
+    // Heuristic to see if we're invoked as a shebang script, namely,
+    // if we have at least one argument, it's the name of an
+    // executable file, and it starts with "#!".
+    Strings savedArgs;
+    auto isNixCommand = std::regex_search(programName, std::regex("nix$"));
+    if (isNixCommand && cmdline.size() > 0) {
```
Ericson2314, line 87: I think this should be done in the caller, `libutil` shouldn't care what the `programName` is.

fricklerhandwerk:
Reviewed in Nix team meeting 2023-05-01:

 - status unclear
 - @roberth: stuck on syntax?
 - @edolstra: not a replacement for `nix-shell -p` yet
 - @roberth: that's ok, separate scope
 - @roberth: can technically be done with `nix develop --expr '...'`
 - @tomberek: what about the syntax? Currently scans all comment-like lines
 - @tomberek will rebase, @roberth still assigned for review

src/libutil/util.cc
```diff
@@ -1459,6 +1460,49 @@ std::string shellEscape(const std::string_view s)
...
+            case '"':
+                cur.append(begin, it);
+                begin = it + 1;
+                st = st == sBegin ? sQuote : sBegin;
+                break;
+            case '\\':
+                /* perl shellwords mostly just treats the next char as part of the string with no special processing */
+                cur.append(begin, it);
+                begin = ++it;
+                break;
```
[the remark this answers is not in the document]
tomberek: @roberth This is still pending.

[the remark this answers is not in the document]
roberth: I think trying to conform with existing syntaxes and conventions is a far harder problem than coming up with a useful syntax that works well for our use case.
New proposal:

 - Use double backtick quotes as the only support type of quotation
 - Disallow any shell-like syntax outside the quoted strings, and be eager to reject syntax in general

This way we keep the code fairly simple, leaving reserved syntax for possible later extension, leaving our options open, while solving "our own use case" well: postponing good inline expression support seems like a mistake the more I think about it.

I'll have a go at implementing this.

roberth:
I've changed the quoting and interpretation of relative paths in #8327. wdyt?

tomberek:
> I've changed the quoting and interpretation of relative paths in #8327. wdyt?

Looks reasonable, but more contentious. I'm okay with either approach.

tomberek:
v2 was merged, please try it out and provide feedback

PR #5197, opened 2021-08-31
Remove dynamic initialization of base64DecodeChars

roberth:
Solves a [build error](https://github.com/NixOS/nixpkgs/pull/136271) while simplifying the code.

Static initialization is simpler and doesn't require any form of
synchronization.

This solves a build error on clang.

    src/libutil/util.cc:1451:30: error: implicit instantiation of undefined template 'std::__1::array<char, 256>'
    static std::array<char, 256> base64DecodeChars;
                             ^
    /nix/store/w6mp8b25lld3b7s0gvi709dfkw619s36-libcxx-7.1.0-dev/include/c++/v1/__tuple:223:64: note: template is declared here
    template <class _Tp, size_t _Size> struct _LIBCPP_TEMPLATE_VIS array;


edolstra:
This is rather ugly...

BTW it's possible to use `static` instead of explicit synchronisation, since C++ ensures that static variables are initialised in a thread-safe way. So something like
```
string base64Decode(std::string_view s)
{
    static std::array<char, 256> base64DecodeChars = []() {
      std::array<char, 256> res;
      ....
      return res;
    };
} 
```
should work.

Or if the problem is `std::array`, then just use a `std::string`.

roberth:
This code is super uninteresting and should just work, which is why I chose the simplest option.
Sure it's not aesthetically pleasing, but does that really matter here?

src/libutil/util.cc
```diff
@@ -1474,11 +1485,6 @@ string base64Encode(std::string_view s)
 string base64Decode(std::string_view s)
 {
     static std::once_flag flag;
```
dtzWill, line 1487:
```suggestion
```

(I think?)

andir:
I am with Eelco on this one. Lets keep the list of magic tables to a minimum. The lambda for the static is probably the most readable version. If you want to be super fancy you could try to get this done with a constexpr lambda that is "executed" during compilation.

roberth:
If you guys prefer _maintaining code_ over simply storing 256 unchanging bytes of data, that's your loss.
I suggest you fix the build (see link in PR descr) in whichever way you fancy, but I will have no part in it.

PR #4155, opened 2020-10-17
Merge master into #3600

Ericson2314:
The next few conflicts are more interesting, so I stopped here for now.

PR #5212, opened 2021-09-03
Merge master into #3600

mkenigs:
https://github.com/NixOS/nix/pull/4155 is now out of date as well


edolstra:
Thanks! There is a compilation error in `local-derivation-goal.cc`:
```
src/libstore/build/local-derivation-goal.cc:2439:9: error: no matching function for call to 'canonicalisePathMetaData'
        canonicalisePathMetaData(actualPath, -1, inodesSeen);
```

Could you have a look at that?

edolstra:
Thanks, awesome!

PR #5215, opened 2021-09-05
Fix use-registries logic in builtins.getFlake

### LINKED ISSUE #5205: `builtins.getFlake` doesn't work with `--impure` anymore
**Describe the bug**

```
$ nix eval --impure --expr '(builtins.getFlake "nixpkgs").description'
error: 'flake:nixpkgs' is an indirect flake reference, but registry lookups are not allowed
```

**Expected behavior**

It should not err. This used to work.

**`nix (Nix) 2.4pre20210802_47e96bb`**

ncfavier:
Fixes #5205.

cc @balsoft

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -686,7 +686,7 @@ static void prim_getFlake(EvalState & state, const Pos & pos, Value * * args, Va
         lockFlake(state, flakeRef,
             LockFlags {
                 .updateLockFile = false,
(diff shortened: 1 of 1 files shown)


balsoft (changes requested):
I'm pretty sure the fix is wrong. I'll look at this bug next week.

EDIT: Actually, I misread the patch, looks correct. The test failures still have to be looked into.

balsoft (approved):
Ok, actually, the patch looks good and the test failures seem unrelated. Sorry for the confusion.

Issue #5213, opened 2021-09-04
`builtins.hashFile` and `builtins.hashString` do not discard string context


kvtb:
```
nix-repl> builtins.hasContext (builtins.hashFile "sha256" (builtins.fetchurl "https://github.com/NixOS/nixpkgs/archive/release-21.05.tar.gz")) 
true

nix-repl> builtins.hasContext (builtins.hashString "sha256" (builtins.fetchurl "https://github.com/NixOS/nixpkgs/archive/release-21.05.tar.gz"))
true
```

Looking at the comments, they should discard string context, but it restored just on the next lines, in `coerceToPath` and `forceString` (should it be `forceStringNoCtx`?)

https://github.com/NixOS/nix/blob/2a8017092025a7c108ca1a829a8b4f0bcf7ee2c0/src/libexpr/primops.cc#L1363-L1364

https://github.com/NixOS/nix/blob/2a8017092025a7c108ca1a829a8b4f0bcf7ee2c0/src/libexpr/primops.cc#L3038-L3039

cc @shlevy @Ericson2314 @JohnAZoidberg

edolstra:
Fixed in c6fa7775de413a799b9a137dceced5dcf0f5e6ed.

Issue #4015, opened 2020-09-15
builtins.hashFile discards string context


exFalso:
**Describe the bug**

`builtins.hashFile` discards the string context, making `builtings.hashFile ${something}` fail.

**Steps To Reproduce**

```bash
$ nix-instantiate -E 'let pkgs = import <nixpkgs> {}; in pkgs.writeText "hello.sha256" (builtins.hashFile "sha256" (pkgs.writeText "hello" "world"))'
error: opening file '/nix/store/x4vggq06hhyxpzy168lhqyn2sy5h8j2q-hello': No such file or directory
```

**Expected behavior**

No failure

**`nix-env --version` output**

```
$ nix-env --version
nix-env (Nix) 2.3.6
```

**Additional context**

I'm fairly certain the culprit is https://github.com/NixOS/nix/blob/2a8017092025a7c108ca1a829a8b4f0bcf7ee2c0/src/libexpr/primops.cc#L1363, which seems to have been lifted verbatim from https://github.com/NixOS/nix/blob/2a8017092025a7c108ca1a829a8b4f0bcf7ee2c0/src/libexpr/primops.cc#L3038

Ericson2314:
Yeah this will need to block on the context being realized, or if we don't want to do that just fail.

kvtb:
Actually, no.

`builtins.hashFile` does not discard the string context (although I would expect it to discard and I found this issue a moment before filine my own about "builtins.hashFile DOES NOT discard string context"):

```
nix-repl> builtins.hasContext (builtins.hashFile "sha256" (builtins.fetchurl "https://github.com/NixOS/nixpkgs/archive/release-21.05.tar.gz")) 
true

nix-repl> builtins.hasContext (builtins.hashString "sha256" (builtins.fetchurl "https://github.com/NixOS/nixpkgs/archive/release-21.05.tar.gz"))
true
```

but string contexts have nothing to do with your problem.

you just want your
 `pkgs.writeText "hello.sha256" (builtins.hashFile "sha256" (pkgs.writeText "hello" "world"))`
to trigger IFD like
 `pkgs.writeText "hello.sha256" (builtins.readFile (pkgs.writeText "hello" "world"))`
does.

exFalso:
> Actually, no.
> 
> `builtins.hashFile` does not discard the string context (although I would expect it to discard and I found this issue a moment before filing my own issue about "builtins.hashFile DOES NOT discard string context"):
> 
> ```
[...]

PR #5216, opened 2021-09-05
`builtins.hashFile`, `builtins.hashString`: realize context before calculation, and discard afterwards

kvtb:
… 
Fixes https://github.com/NixOS/nix/issues/4015
Fixes https://github.com/NixOS/nix/issues/5213


regnat:
> builtins.hashString: realize context before calculation

This doesn’t realize it for `hashString`, does it? (just asking to make sure as that’s not what the description of the PR implies)

kvtb:
> > builtins.hashString: realize context before calculation
> 
> This doesn’t realize it for `hashString`, does it? (just asking to make sure as that’s not what the description of the PR implies)

yes, it seems needless for `hashString`
As far as I understand, even if `hashString` calculated hash of string with context, the string and its hash won't change on context realization

PR #5223, opened 2021-09-06
Expr::show(): avoid emiting meaningless chars

kvtb:
Expr::show(): avoid emiting meaningless chars


tomberek:
Can you show a before/after of the output?

kvtb:
`foo."${bar}"` -> `foo.${bar}`

PR #5225, opened 2021-09-08
Add missing include in util.cc

ncfavier:
The clang build error that #5197 was trying to fix seems to be caused by a missing `<array>` header in `util.cc` (cc @andir #5075).

This is completely untested but should be obvious enough.

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -4,6 +4,7 @@
 #include "finally.hh"
 #include "serialise.hh"
 
+#include <array>
 #include <cctype>
 #include <cerrno>
 #include <climits>


happysalada:
I'm starting a local darwin build, I'll update with the results when it's done (my machine is quite old, this might take some time).

ncfavier:
Test failure unrelated.

happysalada:
Yeah I was having a look and it doesn't seem to be related to this.
What gives me hope is that the compilation went ahead until the end.

happysalada:
The build is just finished and it worked like a charm!
great job on finding this fix!

```
❯ ./result/bin/nix --version
nix (Nix) 2.4pre20210908_7f0d177
```

if you want to make a PR on nixpkgs that includes this patch, I'm happy to review it and merge it.
we can also wait a couple of days for upstream if you want.

ncfavier:
I'm fine with waiting. I expect this to be merged quickly anyway

edolstra:
Thanks!

Issue #5208, opened 2021-09-02
Make it possible to run the Nix daemon as a non-root user


regnat:
# Problem

In multi-user mode, Nix currently assumes that the daemon is running as `root` (and the installer sets things up that way).
This can be (and in practice is) a blocker for some corporate environments where installing Nix could be acceptable, but running it as root is out of the question.
More generally, this is a security concern as it means running a fairly large (and not tremendously well audited) codebase as root.

Most of what Nix does however don’t require it: For the most part, everything the Nix daemon needs is to be able to write to the `/nix` directory.
So it should be possible to run the Nix daemon as its own dedicated user, which would also be the user of `/nix`.

# Solution

Make it possible (and the default) to have the Nix daemon run as a non-root user

# Additional context

## What is root used for atm? (and how can we avoid it?)

(Possibly I’m missing some uses, please let me know if you think of something that’s not already there)

- Writing to the Nix store. This can trivially be changed by changing the owner of the store (like in single-user mode)
- Using the build users. This is partially redundant with the build sandbox, except that
    1. The build sandbox doesn’t actually change the user, which might mean that different sandboxed builds might still be able to interfere with each other
    2. Fixed-output derivations aren’t sandboxed
    3. The sandbox is costly, so the default configuration (`sandbox = relaxed`) allows derivations to opt-out of it (by setting `__noChroot`)

    Point 1. can be fixed by using a [subuid](https://man7.org/linux/man-pages/man5/subuid.5.html) mapping for the `nix` user (like all container systems do).
    Point 2 and 3 can be fixed by having a light sandbox that can be used both by FO derivations and `__noChroot` derivations.
- The GC needs to follows the roots through the user home directories (for indirect roots).

    I don’t think there’s a solution to that, so I think the root discovery system will have to be factored out into a separate executable (with a much smaller attack surface) with the right capabilities.

alexshpilkin:
Not stale, dammit.

Ericson2314:
Is there a branch for this, even if it has conflicts?

thufschmitt:
@Ericson2314 oh, I forgot to backlink it, but yes, there's https://github.com/NixOS/nix/pull/5380 open for that (and it's essentially conflict-free barring some trivial things)

clamydo:
Would it be also possible to change the prefix of the nix store from `/nix/store` to something custom and potentially in the users home?

This would help to run nix as a user in a single-user scensrio without requiring root to set up a non-user-owned folder in the file system root.

[...]

PR #5226, opened 2021-09-08
Move the default profiles to the user’s home

thufschmitt:
Rather than using `/nix/var/nix/{profiles,gcroots}/per-user/`, put the user
profiles and gcroots under `$XDG_DATA_DIR/nix/{profiles,gcroots}`.

This means that the daemon no longer needs to manage these paths itself
(they are fully handled client-side). In particular, it doesn’t have to
`chown` them anymore (removing one need for root).

This does change the layout of the gc-roots created by nix-env, and is
likely to break some stuff, so I’m not sure how to properly handle that.

First step towards #5208

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1517,7 +1517,7 @@ void LocalDerivationGoal::startDaemon()
                 try {
                     daemon::processConnection(store, from, to,
                         daemon::NotTrusted, daemon::Recursive,
-                        [&](Store & store) { store.createUser("nobody", 65535); });
+                        [&](Store & store) {});
                     debug("terminated daemon connection");
                 } catch (SysError &) {
                     ignoreException();
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -201,8 +201,6 @@ LocalStore::LocalStore(const Params & params)
             throw SysError("could not set permissions on '%s' to 755", perUserDir);
     }
 
-    createUser(getUserName(), getuid());
-
     /* Optionally, create directories and set permissions for a
        multi-user install. */
     if (getuid() == 0 && settings.buildUsersGroup != "") {
@@ -1823,20 +1821,6 @@ void LocalStore::signPathInfo(ValidPathInfo & info)
     }
 }
 
-
-void LocalStore::createUser(const std::string & userName, uid_t userId)
-{
-    for (auto & dir : {
-        fmt("%s/profiles/per-user/%s", stateDir, userName),
-        fmt("%s/gcroots/per-user/%s", stateDir, userName)
-    }) {
-        createDirs(dir);
-        if (chmod(dir.c_str(), 0755) == -1)
-            throw SysError("changing permissions of directory '%s'", dir);
-        if (chown(dir.c_str(), userId, getgid()) == -1)
-            throw SysError("changing owner of directory '%s'", dir);
-    }
-}
 
 std::optional<std::pair<int64_t, Realisation>> LocalStore::queryRealisationCore_(
         LocalStore::State & state,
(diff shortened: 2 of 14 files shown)


fricklerhandwerk:
Discussed in Nix team meeting on 2022-12-19.

Decision: take out the migration into a different PR, and review this one together when it's ready.

- @thufschmitt: changing the directory layout may be breaking for existing tools, but not sure who is relying on this
    - @edolstra: migrating the profile dir precludes rolling back
        - @Ericson2314: we could prefer the old dir if it exists
            - @thufschmitt: then we're stuck with these two locations forever (which we already are)
            - @Ericson2314: at least for a few migration cycles
            - @thufschmitt: will also break third-party tools depending on when Nix first was installed, even if for the same version of Nix. this is the same problem as with NixOS `config.stateVersion`.
                - @infinisil: what about introducing a tool to query for profile paths so tools don't have to rely on actual file system paths?
    - @infinisil: can put a symlink
        - deamon has to be root to follow it
- @edolstra: the daemon should not trust user directories
    - @thufschmitt: this PR does not change anything about how GC works, there is nothing there which cannot already be done by e.g. setting `--profile`
- agreement: take out the migration into a different PR, and review it together next time

</details>

thufschmitt:
I've removed the migration code, now any existing installation will keep using the old profiles directory.

I've also changed the behaviour for `root` to keep the `/nix/var/nix/profiles/default` location as it's also used as a system-wide profile (so it should stay world-readable and stable)

src/libstore/build/local-derivation-goal.cc
```diff
@@ -1517,7 +1517,7 @@ void LocalDerivationGoal::startDaemon()
                 try {
                     daemon::processConnection(store, from, to,
                         daemon::NotTrusted, daemon::Recursive,
-                        [&](Store & store) { store.createUser("nobody", 65535); });
+                        [&](Store & store) {});
```
edolstra, line 1520: I think we can remove the `authHook` argument to `processConnection()` now, since it's only used to call `createUser()`.

edolstra:
This is probably important enough to mention in the release notes, even if it shouldn't break anything.

domenkozar:
See https://github.com/cachix/cachix-action/issues/137

ncfavier:
This makes the assumptions in https://github.com/nix-community/home-manager/blob/master/modules/lib-bash/activation-init.sh invalid, causing failures like https://github.com/nix-community/home-manager/actions/runs/4301868239/jobs/7499628919#step:8:260. How do we migrate smoothly?

fricklerhandwerk:
@ncfavier please open an issue and ping me, we'll discuss it on the team.

rycee:
I'm looking at how to adapt Home Manager to this change. Should I use some HM specific directory to keep the profile and gcroot? E.g. `$XDG_DATA_DIR/home-manager/{profiles,gcroots}`. And nix-collect-garbage will automatically handle the profile but leave the gcroot alone? Or should I put the HM profile and gcroot in `$XDG_DATA_DIR/nix/{profiles,gcroots}`?

Also, this commit doesn't seem to contain any documentation updates. Are those in a separate PR?

rycee:
Hmm, I had a look at `nix-collect-garbage.cc` and it seems to me that it will only delete old profile generations if they are present in the Nix state dir? If so, won't it also fail to expire old user profiles after this migration? I'm not very familiar with the Nix code, though, so I may be completely mistaken here.

thiagokokada:
> This makes the assumptions in https://github.com/nix-community/home-manager/blob/master/modules/lib-bash/activation-init.sh invalid, causing failures like https://github.com/nix-community/home-manager/actions/runs/4301868239/jobs/7499628919#step:8:260. How do we migrate smoothly?

Interesting. I just hit this issue when trying to use the Nix static inside a Steam Deck, with the store inside `$HOME` (introduced in https://discourse.nixos.org/t/nix-2-10-0-released/20291). While it is unfortunate that this bug, having this being fixed can be exciting.

fricklerhandwerk:
@Rycee Apparently this still needs more work on our side. Please open an issue so we can discuss it on the team.

PR #5235, opened 2021-09-11
ionice to idle during GC

kvtb:
GC can remove millions of files, especially in `/nix/store/.links`.
That makes desktop system unresponsive during GC for many minutes.

### DIFF
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -17,6 +17,26 @@
 #include <unistd.h>
 #include <climits>
 
+#if __linux__
+#include <sys/syscall.h>
+
+enum {
+    IOPRIO_WHO_PROCESS = 1,
+    IOPRIO_WHO_PGRP,
+    IOPRIO_WHO_USER
+};
+
+enum {
+    IOPRIO_CLASS_NONE,
+    IOPRIO_CLASS_RT,
+    IOPRIO_CLASS_BE,
+    IOPRIO_CLASS_IDLE
+};
+
+#define IOPRIO_CLASS_SHIFT 13
+#endif
+
+
 namespace nix {
 
 
@@ -734,6 +754,12 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
     if (state.shouldDelete)
         deletePath(reservedPath);
 
+#if __linux__
+    if (syscall(SYS_ioprio_set, IOPRIO_WHO_PROCESS, 0, IOPRIO_CLASS_IDLE << IOPRIO_CLASS_SHIFT))
+        throw SysError("collectGarbage: ionice to idle");
+
+    try {
+#endif
     /* Acquire the global GC root.  This prevents
        a) New roots from being added.
        b) Processes from creating new temporary root files. */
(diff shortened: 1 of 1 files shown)


kvtb:
didn't help much :(
Maybe also add sleep() after about every 50,000 files to be deleted

src/libstore/gc.cc
```diff
@@ -17,8 +17,39 @@
...
+        IOPRIO_CLASS_NONE,
+        IOPRIO_CLASS_RT,
+        IOPRIO_CLASS_BE,
+        IOPRIO_CLASS_IDLE
+    };
+
+    enum {
+        IOPRIO_CLASS_SHIFT = 13
+    };
+    IoNice(int cl) {
```
roberth, line 39: You may as well hardcode this parameter, because the error message expects it that way and we don't have a use case for another value than `IOPRIO_CLASS_IDLE`.

kvtb: yes

maybe whole nix (and nix-daemon) should be ionice'd.
creating derivations in /nix/store is io heavy and low priority too

roberth: Nix can be used interactively or in a 'management' role though. Starving deployment on an I/O bottlenecked server could make a bad day much worse.

kvtb: I meant exactly the case of a " I/O bottlenecked server".
It is when Nix must idle to what server normally does in order not to put the server down

roberth: Or it is when Nix must be usable in order to fix the I/O problem.
If your service must provide hard realtime guarantees, use a realtime operating system. If your service must provide soft realtime guarantees, being able to fix issues quickly is at least as important as the service itself. If you got yourself into a situation where it's I/O bottlenecked, that is a problem with your deployment, because the deployment should be designed to stay well within limits. Hence, to fix the problem with the server, you need to be able to deploy, and the sooner, the better. If you prioritize the production load, you'd have to shut down the production service, which is likely worse, because it is not obvious and not something you would typically dare to automate either.
So I don't think idle is a good default for nix-daemon. In the few cases where it does make sense, specifically when you know you won't need Nix to fix stuff, you can set IOSchedulingClass in NixOS' systemd options. However, if you don't need Nix to fix stuff, you probably don't need its daemon either.

kvtb: I observed LA > 30 on a cheap vps with spindle disk every time when Nix downloaded and unpacked packages during routing upgrade.
There was nothing realtime so I ignored that and forget.
And only now I realize that it could be avoided.

Mic92:
btw. one can also do resource control for systemd services easily: https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html
Interesting values are: `IOWeight` and `CPUWeight`.

roberth:
@kvtb see also https://github.com/NixOS/nixpkgs/pull/138741

kvtb:
> @kvtb see also [NixOS/nixpkgs#138741](https://github.com/NixOS/nixpkgs/pull/138741)

That's nice.

Although there is one use case uncovered:

`nix-collect-garbage -d` - the only way to remove old generations of system closures, and thus the heavies GC removal on workstations - does not work via `nix-daemon`

thufschmitt:
Discussed during the Nix maintainers team meeting: It's hard to ensure that this will work properly without breaking the rest, so we won't accept it. Our suggestion would be to instead manually run `nix-collect-garbage` as root (which would bypass the daemon) with a custome ionice setting (along the lines of `sudo ionice -c Idle nix-collect-garbage -d`).

- Neat idea, but no one is too familiar with this

- @edolstra: Be wary of priority inversion, if a deprioritized GC thread holds some locks we might get into trouble.

  - GC is no longer takes a big global locks so this is less of a problem
  - But other code like SQLite just takes locks
  - "Idle" seems too low

- @edolstra: Also be aware of auto-GC where other non-GC things are happening in the same process
  - @edolstra: It appears it can be set per-thread, but this is setting it per-process

- @thufschmitt: Let's close this and instead suggest running something like `sudo ionice -c Idle nix-collect-garbage -d`
</details>

PR #5241, opened 2021-09-13
Draft: add support for lfs for fetchGit

marius851000:
I wanted to add veloren, an open-source video game that have a flake expression, using crate2nix.

It use git lfs, so I needed to support it to the flake. Right now, hydra is able to fetch the game repo, at least as a flake (and check fail, but that's for another, unrelated reason I should fix). I'm unsure a few things, so I'll add comment to my code.

Also, I'm unsure of the name of the parameter. I named it largeFileSystem, but it could be renamed to another thing.

Also, this need git lfs to be present in the path (just like fetchGit require git).

additionally, is there any specific procedure when wanting to change how flake url work ? I hope I don't need an RFC.

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -47,6 +47,8 @@ struct GitInputScheme : InputScheme
                 attrs.emplace(name, value);
             else if (name == "shallow")
                 attrs.emplace(name, Explicit<bool> { value == "1" });
+            else if (name == "largeFileSystem")
+                attrs.emplace(name, Explicit<bool> { value == "1" });
             else
                 url2.query.emplace(name, value);
         }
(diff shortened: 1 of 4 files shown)


src/libfetchers/git.cc
```diff
@@ -47,6 +47,8 @@ struct GitInputScheme : InputScheme
                 attrs.emplace(name, value);
             else if (name == "shallow")
                 attrs.emplace(name, Explicit<bool> { value == "1" });
+            else if (name == "largeFileSystem")
```
marius851000, line 50: Comparing a "1" string seems pretty bad. Should creating a function that parse a string as boolean (like parse true or false as valid entry) be a good idea ? A user might think putting ?largeFileSystem=true work, then be surprised (but the shallow code did that).

src/libfetchers/git.cc
```diff
@@ -391,10 +397,20 @@ struct GitInputScheme : InputScheme
                 utimes(localRefFile.c_str(), times);
             }
 
+            // don't know how to check if the lfs was cloned up to the end, so fetch it all the time (if enabled, which should be uncommon)
+            if (largeFileSystem) {
+                if (getProgramStatus("git", true, { "-C", repoDir, "remote", "get-url", "nix_fetcher_remote_for_lfs"})) {
```
marius851000, line 402: Is there a better way to do that ? In particular, the user see an error logged on stderr if the remote doesn't exist (but I put it before it run the git lfs fetch command to prevent issue if interupted, as git fail to create a remote if one already exist).

src/libutil/util.cc
```diff
@@ -1034,6 +1034,19 @@ std::vector<char *> stringsToCharPtrs(const Strings & ss)
     return res;
 }
 
+// return true if the programmed returned ok, false otherwise
+bool getProgramStatus(Path program, bool searchPath, const Strings & args,
```
marius851000, line 1038: I didn't find a function that allow to check the output of a program, but I will be suprised if there aren't any in this code base.

marius851000 (commented):
Commented the question I had. I'm not really good at C++, not having practiced this for a while (and never having more than a begginer level to begin with). But it work. I tested to fetch local repo, as well as remote repo, including with submodule using lfs themselves. I also tested updating an already fetch repo (with veloren, in hydra. it work).

PR #5247, opened 2021-09-14
Remove the 'ca-references' feature check

### LINKED ISSUE #3422: ca-references should not be experimental
Recently I started running into issues building my aarch64 systems on the community box. The community box is on 19.09, and I'm building systems on unstable. The error I get is the following:
```
experimental Nix feature 'ca-references' is disabled
```
Which in turn comes from this
https://github.com/NixOS/nix/blob/b816515f613b7013ba0d5489841568b8f666f224/src/libstore/local-store.cc#L1010

This is new behavior as I've been doing my system builds on that remote builder for a while now and never faced any issues. 

As of right now this blocks a few people, such as myself and @cleverca22  from successfully using the community box for testing our aarch64 system builds.

cc. @grahamc

### LINKED ISSUE #4425: nix flakes: nix profile install always fails with "claims to be content-addressed but isn't" "because it lacks a valid signature"
**Describe the bug**

`nix profile install` isn't working for me

**Steps To Reproduce**

```
$ nix profile install nixpkgs#stdenv
warning: path '/nix/store/lnkmn9zqkrf3a9bk310lfb4yq95m4hml-profile' claims to be content-addressed but isn't
error: --- Error ----------------------------------------------------------------------------------------------------------------------------------------- nix
cannot add path '/nix/store/lnkmn9zqkrf3a9bk310lfb4yq95m4hml-profile' because it lacks a valid signatur
```
**Expected behavior**

the package installs in my profile

```
$ nix --version
nix (Nix) 2.4pre20201201_5a6ddb3
```

**Additional context**

edolstra:
CA paths with references seem stable enough so I don't think we need to keep it behind a feature check.

Fixes #3422.
Fixes #4425.

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1244,11 +1244,6 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
 
             deletePath(realPath);
 
-            // text hashing has long been allowed to have non-self-references because it is used for drv files.
-            bool refersToSelf = info.references.count(info.path) > 0;
-            if (info.ca.has_value() && !info.references.empty() && !(std::holds_alternative<TextHash>(*info.ca) && !refersToSelf))
-                settings.requireExperimentalFeature("ca-references");
-
             /* While restoring the path from the NAR, compute the hash
                of the NAR. */
             HashSink hashSink(htSHA256);


edolstra:
Also CC @Ericson2314.

regnat (approved):
Awesome!

Ericson2314:
I don't think CA *self* references should be stable yet.

Ericson2314:
Based on what @regnat says in https://github.com/NixOS/nix/pull/4849#issuecomment-848143468, I am not even sure all the self references in all locations are implemented right, yet.

Ericson2314:
Moreover, I don't see the practical benefit of doing this: most useful stuff would be still blocked by CA-derivations, but one can now add-to-store in various ways? Seems we are giving up a few degrees of design freedom for no one's benefit.

edolstra:
We need CA references for `nix profile`. I don't think it needs self-references though, so we could put those behind a feature check.

Ericson2314:
Will `nix profile` become stable soon? I wonder if we could also use a built-in derivation builder like corepkgs's fetchurl does to punt on that issue for now.

In general, I'm a big fan of those built-in derivations.

edolstra:
`nix profile` actually moved away from using the built-in derivation (used by `nix-env`) to simplify things. It just builds the symlink tree client-side and calls `addToStore()`.

Ericson2314:
I'm not sure what's better about that? Besides avoiding unstable features, a derivation allows downstream derivations in a nice way. That could be be useful for CI-ing a bunch of people's personal environments.

Still, just making non-self-references stable is most or the benefit. Seeing a little mini-RFC for that would be great to help build confidence in the governance mechanism.

PR #5249, opened 2021-09-14
Add missing 'nix profile' subcommands

### LINKED ISSUE #3951: Profile generations management with `nix profile` command
We can use `nix build --profile <PROFILE> <installable>` to update profiles like `nix-env --set`. But what the `nix` command can't do yet is manage profile generations - e.g. rollback, list or switch to a certain generation.

A port of `nix-env --list-generations`, `nix-env --switch-generation` and `nix-env --rollback` to the `nix profile` command as subcommands with respective names.

Continue using `nix-env` which feels not elegant when we have the `nix` command that's becoming the hub for all Nix-related tasks.

**Additional context**
The `nixos-rebuild` tool could be subsequently updated to use the new `nix profile`-based commands instead of nix-env.

edolstra:
This adds `nix profile rollback` and `nix profile wipe-history` subcommands. It also makes `nix profile history` show profile dates.

Fixes #3951.

### DIFF
--- a/src/libstore/profiles.cc
+++ b/src/libstore/profiles.cc
@@ -236,6 +236,37 @@ void switchLink(Path link, Path target)
 }
 
 
+void switchGeneration(
+    const Path & profile,
+    std::optional<GenerationNumber> dstGen,
+    bool dryRun)
+{
+    PathLocks lock;
+    lockProfile(lock, profile);
+
+    auto [gens, curGen] = findGenerations(profile);
+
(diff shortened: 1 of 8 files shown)


fluffynukeit:
I'm excited to try this out after discovering today that the 2021-08-23 release doesn't include these commands.  I couldn't figure out how to manage my `nix profile` operations, and turns out that some important pieces/conveniences were just in the works.  Thanks for the hard work.

Issue #1556, opened 2017-09-04
Document the store URI syntax


taktoa:
In particular, the parameters that can be passed after the `?` are not documented, and are difficult to find in the source code.

The following docstring includes everything I was able to figure out about the Nix store URI syntax:

```
        description = ''
          This specifies the Nix store that Hydra should use, as a store URI.

          This can take one of the following forms:

          <itemizedlist>
            <listitem>
              <literal>local</literal>:
              the Nix store in <literal>/nix/store</literal> and the database
              in <literal>/nix/var/nix/db</literal>, accessed directly.
            </listitem>
            <listitem>
              <literal>daemon</literal>:
              the Nix store accessed via a Unix domain socket connection to
              the <literal>nix-daemon</literal>.
            </listitem>
            <listitem>
              <literal>auto</literal>:
              depending on whether the user has write access to the local
              Nix store, this will be equivalent to either
              <literal>local</literal> or <literal>daemon</literal>.
            </listitem>
            <listitem>
              <literal>file://<replaceable>path</replaceable></literal>:
              a binary cache stored in <replaceable>path</replaceable>.
            </listitem>
            <listitem>
              <literal>https://<replaceable>path</replaceable></literal>:
              a binary cache accessed via HTTP over TLS.
            </listitem>
            <listitem>
              <literal>s3://<replaceable>path</replaceable></literal>:
              a writable binary cache stored on Amazon S3.
            </listitem>
            <listitem>
              <literal>ssh://[<replaceable>user</replaceable>@]<replaceable>host</replaceable></literal>:
              a remote Nix store accessed by running
              <command>nix-store</command> <option>--serve</option> via SSH.
            </listitem>
          </itemizedlist>

          Any of these store URIs can be suffixed by a parameter string, e.g.:
          <literal><![CDATA[https://example.com?foo=bar&baz=quux]]></literal>.

          In particular, the <literal>secret-key</literal> parameter is useful
          for specifying a binary cache signing key file.

          The parameters that can be passed as part of a Nix store URI are not
          currently documented, but if you want to find them it is useful to
          look at all the classes in the Nix <literal>libstore</literal> API
          that are subclasses of the <literal>Store</literal> class. These
          classes represent handlers for different kinds of Nix store URIs,
          and whenever they have a field of type <literal>Setting</literal>
          or <literal>PathSetting</literal>, it corresponds to a store URI
          parameter key.
        '';
```

dtzWill:
[...]

PR #5252, opened 2021-09-15
Document store CLI argument

### LINKED ISSUE #5171: Document `--store` option
I can't find any documentation on the `--store` option, which instructs Nix to use a store path other than `/nix/store`. The way I undestand it is that `--store` takes a string of the format `nix-root-path?store-path`, where `nix-root-path` is an actual location in the filesystem where the `nix` directory resides, and `store-path` is the store prefix used by Nix to build packages (usually `/nix/store` -- if it is something else, then cache.nixos.org will not be of much use). My understanding may or may not be correct; it would be great to have some official documentation to clear things up.

tomberek:
Fixes #5171 and #1556

Mostly took verbiage from previous issues and release notes.

### DIFF
--- a/src/nix/nix.md
+++ b/src/nix/nix.md
@@ -114,6 +114,34 @@ flake in the current directory.
 
 Most `nix` subcommands operate on a *Nix store*.
 
-TODO: list store types, options
+* **local**: The Nix store in `/nix/store` and database in `/nix/var/nix/db`,
+  accessed directly.
+
+* **daemon**: The Nix store accessed via a Unix domain socket connection to
+  nix-daemon.
+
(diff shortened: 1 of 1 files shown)


regnat:
There’s a `nix describe-stores` command that generates a documentation for all the existing stores and their options. It would be great to use it to generate this documentation (or possibly an appendix as it’s more exhaustive but also less reader-friendly)

Ericson2314:
Yeah we should generate that appendix and basically just point to it.

Ericson2314:
I think we can close this now that we have `nix --extra-experimental-features nix-command help-stores`?

fricklerhandwerk:
Yes!

PR #5265, opened 2021-09-17
TCP support in the Nix daemon

edolstra:
This PR makes it possible to connect to a remote Nix daemon via TCP, and for the Nix daemon to listen on multiple sockets (including TCP) via systemd.

Lacking encryption/authentication, this is primarily useful to testing, e.g. when locally simulating high-latency SSH store connections via qdisc/netem.

Example usage:
``` 
# nix path-info --store tcp://example.org:1234 ...
```

### DIFF
--- a/src/nix/daemon.md
+++ b/src/nix/daemon.md
@@ -8,8 +8,8 @@ R""(
   # nix daemon
   ```
 
-* The daemon does not have native support for listening on a TCP
-  socket, but you can do this using `socat`:
+* The daemon does not have native support for opening a TCP socket to
+  listen to, but you can do this using `socat`:
 
   ```console
   # socat TCP-LISTEN:3456,reuseaddr,fork EXEC:'nix daemon --stdio',nofork
--- a/src/nix/daemon.md
+++ b/src/nix/daemon.md
@@ -31,4 +31,18 @@ management framework such as `systemd`.
 
 Note that this daemon does not fork into the background.
 
+# Systemd socket activation
+
+`nix daemon` supports systemd socket-based activation using the
+`nix-daemon.socket` unit in the Nix distribution. It supports
+listening on multiple addresses; for example, the following stanza in
+`nix-daemon.socket` makes the daemon listen both on a Unix domain
+socket and on port 1234:
+
+```
+[Socket]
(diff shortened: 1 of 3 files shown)


Ericson2314:
Great!! I was hoping we'd do these things.

src/nix/daemon.md
```diff
@@ -8,6 +8,19 @@ R""(
   # nix daemon
   ```
 
+* The daemon does not have native support for listening on a TCP
+  socket, but you can do this using `socat`:
```
Ericson2314, line 12:
```suggestion
* The daemon does not have native support for opening on a TCP
  socket to listen to, but you can do this using `socat`:
```

Ericson2314: Since it does have native support to listen on an already-open file descriptor via the systemd env vars.

andir:
I wonder if we really need that in Nix itself. if it is purely about testing locally wouldn't socat on the current unix socket work just fine?

NickCao:
But for building something like nixbuild.net, this can be extremely helpful.

Ericson2314:
To me this is good enough to merge I don't care about the daemon being able to open sockets of any sort because I basically believe one should always use socket activation.

I assume @edolstra is mainly waiting to get 2.4 our before returning to new features. That is also fine with me.

haslersn:
> Lacking encryption/authentication, this is primarily useful to testing

Can you elaborate on this? What could an adversary with access to the TCP socket do, except for DoS (by consuming compute resources, bandwidth and disk space)? Is there a way for an adversary to compromise other users that use the same Nix daemon over TCP?

My potential use case: Sharing a long-living Nix daemon across short-living (and mutually distrusting) CI jobs.

Ericson2314:
I have fixed the comments in this, but before I go too far proposing it I figured it would be good to start with something like https://github.com/NixOS/nix/pull/7739

Ericson2314:
https://github.com/NixOS/nix/pull/6312 merges master into this.

Ericson2314:
(This can be rebased once https://github.com/NixOS/nix/pull/14723 is merged)

tomberek:

Issue #2595, opened 2018-12-21
extra-platforms detection should be smarter


matthewbauer:
Currently, when ```system = x86_64-linux```, we assume that we can also set ```extra-platforms = i686-linux```. That is not always true. Specifically, you cannot run i686 in these two cases:

- Windows Subsystem for Linux
- ```CONFIG_IA32_EMULATION = n``` (https://github.com/NixOS/nixpkgs/issues/51097)

In these cases we should NOT set ```extra-platforms = i686-linux```.

We also should think about the idea of "compatible" architectures. Many architectures have backwards compatible support so that:

- i787 -> i686 -> i586 -> i486 -> i386
- armv7 -> armv6 -> armv5 -> armv4 -> armv3

x86-64 only supports 32 bit modes in some cases (see above). Likewise, aarch64 can only run armv7 on processors that support 32-bit mode, which is not guaranteed. Each of these ARM architectures can run in both big endian and little endian modes. Each of these rules could automatically get added to extra-platforms as well. There are similar ideas with endianness and 64/32-bit modes in PowerPC, MIPS, SPARC, RISC-V, and WASM.

Relevant code:

https://github.com/NixOS/nix/blob/812e39313c2bcf8909b83e1e8bc548a85dcd626c/src/libstore/globals.hh#L265-L270

r-burns:
Still relevant

r-burns:
Still relevant

patryk4815:
apple silicon M1 support only 64bit:
`error: cannot set 32-bit personality: Invalid argument`

extra info: https://news.ycombinator.com/item?id=27278019

matthewbauer:
This is running Linux on the M1 right?

Is there some other way we can what personality the system supports? I think we could just do `personality(PER_LINUX32)` & if it succeeds enable arm32 & then do `personality()` to reset it.

patryk4815:
@matthewbauer Yep, it is nixos-vm on apple M1. There are many problems, but I will create a separate issue for it. 

1. For example `binfmt` is not working bad, because CPU not support 32bit, so choosing/detecting interpreter for binfmt is wrong:
<img width="584" alt="Zrzut ekranu 2022-03-30 o 21 43 38" src="https://user-images.githubusercontent.com/3074260/160935119-50d97f68-cacb-4f5e-a6ab-e29ff37f95c8.png">

2. Even I fix binfmt for `armv7l-linux`, but build are still failing with `error: cannot set 32-bit personality: Invalid argument`

PR #5276, opened 2021-09-20
set the PER_LINUX32 personality flag, when building for armv6l-linux or armv7l-linux

cleverca22:
this prevents 32bit builds from detecting a 64bit kernel and picking the wrong target

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1778,6 +1778,10 @@ void LocalDerivationGoal::runChild()
             if (personality(PER_LINUX32) == -1)
                 throw SysError("cannot set i686-linux personality");
         }
+        if (drv->platform == "armv7l-linux" || drv->platform == "armv6l-linux") {
+          if (personality(PER_LINUX32) == -1)
+            throw SysError("cannot set 32bit linux personality");
+        }
 
         /* Impersonate a Linux 2.6 machine to get some determinism in
            builds that depend on the kernel version. */


src/libstore/build/local-derivation-goal.cc
```diff
@@ -1778,6 +1778,10 @@ void LocalDerivationGoal::runChild()
             if (personality(PER_LINUX32) == -1)
                 throw SysError("cannot set i686-linux personality");
         }
+        if (drv->platform == "armv7l-linux" || drv->platform == "armv6l-linux") {
+          if (personality(PER_LINUX32) == -1)
+            throw SysError("cannot set 32bit linux personality");
+        }
```
edolstra, line 1784: This can be combined with the i686-linux case above it (i.e. `if (... || drv->platform == "armv7l-linux" || drv->platform == "armv6l-linux") { ... }`). Otherwise LGTM.

misuzu:
This is awesome! I've used this patch to successfully build minimal armv7l system with `mpd` using Oracle Cloud aarch64 VM.

samueldr:
Note that *technically speaking* it advertises as `armv8l-linux`, which can cause some issues with some packages. But this does not mean this PR is wrong.

This PR is right, and good, but can't be relied on for a `pkgsArmv7lLinux` equivalent to `pkgsi686Linux` in Nixpkgs.

32 bit ARM `armv8l-linux` is *mostly the same* as `armv7l-linux`, and except for few build systems that will barf on that unexpected input (as @cleverca22 can confirm), it should give reasonable enough results.

Annoyingly enough, it can't be *relied on* due to the ambiant impurity of the CPU being a` armv8l-linux`. Not unlike how using an `armv7-linux` system to build `armv6l-linux` "native" builds.

A bad analogy is like if on `x86_64-linux` we had revisions like `x86_64-haswell-linux`. So the issue is that the advertisement of the differences in 32 bit ARM is "more pure" than on 64 bit x86, where ¯\\\_(ツ)\_/¯ is the best answer to whether an `x86_64-linux` system supports new instructions. Results here are probably as reasonable for `armv7l-linux` as results for `x86_64-linux` are on a recent *Ryzen* chip given to a *Core 2* system.

samueldr:
(See also #2595)

PR #5279, opened 2021-09-21
Fix relative path input handling

### LINKED ISSUE #5168: Restrict "path" inputs in flakes
Currently you can do wacky stuff like declaring an input `foo.url = path:../`, which will cause Nix to start copying the entire Nix store. We need to specify what absolute/relative paths are allowed in flake inputs. Some possibilities:

* Don't allow paths at all.
* Only allow paths from flakes accessed via a `path:` (or other local) URL. This way e.g. a `github` flake can't read local paths.
* Only allow relative paths to other flakes in the same repository.

edolstra:
Previously inputs like `path:..` or `path:../foo` were accepted. In particular, `path:..` caused Nix to copy the entire Nix store.

Fixes #5168.

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -89,10 +89,12 @@ static void expectType(EvalState & state, ValueType type,
 }
 
 static std::map<FlakeId, FlakeInput> parseFlakeInputs(
-    EvalState & state, Value * value, const Pos & pos);
+    EvalState & state, Value * value, const Pos & pos,
+    const std::optional<Path> & baseDir);
 
 static FlakeInput parseFlakeInput(EvalState & state,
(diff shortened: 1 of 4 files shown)


ju1m:
I suspect (but haven't bisected to be sure) this PR to also impact inputs like `path:./foo`.
On my current `nix` `2.4pre20210922_bcd73eb`, old inputs like this are still working:
```nix
inputs.foo = { type = "path"; path = "./foo"; flake = false; };
```
But such inputs can no longer be added/updated automatically:
```
$ nix flake lock --update-input foo
error: cannot fetch input 'path:./foo' because it uses a relative path
```
This error message was introduced by this PR.
A workaround is to temporarily use an absolute path to update the `flake.lock`'s `narHash` for `foo`,
and then edit `flake.nix` and `flake.lock` to put the relative path back again.

PR #5281, opened 2021-09-21
fix install on macOS monterey

abathur:
I installed Monterey today and tried to install Nix from a very recent master commit and it failed. (I am not sure how the current release installer fares).

The error message from touch, which we're using as a cheap way to detect the read-only root, appears to have shifted. The change here just accepts both forms. I guess this is fine if we don't find some additional conditions that can cause this error message. (In that case we may have to fall back on a slower method of testing this.)

I've used the installer from [my own CI build of this](https://github.com/abathur/nix/runs/3668153655?check_suite_focus=true) to verify that the install works on Monterey.

### DIFF
--- a/scripts/install-darwin-multi-user.sh
+++ b/scripts/install-darwin-multi-user.sh
@@ -13,11 +13,22 @@ NIX_BUILD_USER_NAME_TEMPLATE="_nixbld%d"
 read_only_root() {
     # this touch command ~should~ always produce an error
     # as of this change I confirmed /usr/bin/touch emits:
+    # "touch: /: Operation not permitted" Monterey
     # "touch: /: Read-only file system" Catalina+ and Big Sur
     # "touch: /: Permission denied" Mojave
     # (not matching prefix for compat w/ coreutils touch in case using
     # an explicit path causes problems; its prefix differs)
-    [[ "$(/usr/bin/touch / 2>&1)" = *"Read-only file system" ]]
+    case "$(/usr/bin/touch / 2>&1)" in
+        *"Read-only file system") # Catalina, Big Sur
+            return 0
+            ;;
+        *"Operation not permitted") # Monterey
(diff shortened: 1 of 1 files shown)


edolstra:
Wouldn't this check also fail if `LANG` is set to something other than English?

abathur:
Good question. 

I've poked at using a few languages via preferences and LANG/LC_ALL fiddling and didn't notice any differences (the former just caused `locale` to report C or blank for most settings), but that's a far cry from _knowing_ this check or installer as a whole works when other languages are set.

domenkozar:
Should we also backport this to 2.3?

abathur:
Not directly, at least. `2.3-maintenance` doesn't carry the same code.

It's still in beta, so we have some unknown amount of time (we're roughly in the first few days of the ~plausible release window, given the last ~8 years of history; IIRC October is most common, but Big Sur didn't drop until mid November.)  

If a 2.4 release is as close as the dwindling issue list under the 2.4 milestone suggests, that may be fine?

abathur:
> Wouldn't this check also fail if `LANG` is set to something other than English?

> I've poked at using a few languages via preferences and LANG/LC_ALL fiddling and didn't notice any differences (the former just caused `locale` to report C or blank for most settings), but that's a far cry from _knowing_ this check or installer as a whole works when other languages are set.

To follow up, I've just finished a successful test of this installer on a clean install of macOS where I selected Spanish as the primary/only interface language at OS install-time.

abathur:
For further follow-up, I just successfully installed the current 2.3 release cleanly (as recommended with volume flag and single-user mode).

abathur:
@domenkozar I've also confirmed I can run the current 2.3 installer with the `--daemon` flag. At least at first blush, it doesn't look like 2.3-maintenance will need any updates to be Monterey-ready.

Issue #5280, opened 2021-09-21
Huge performance regression when evaluating clean flake


Kha:
**Describe the bug**

On a clean checkout of a sizeable repo (~20k commits), the time for the first evaluation of a commit has recently increased by more than 10x (and an unhelpful message from Git is shown):
```bash
$ nix build  # cache build

$ echo >> flake.nix; git commit -am foo
[fvis 296fea547d] foo
 1 file changed, 1 insertion(+)

$ nix shell github:NixOS/nix/90b2dd570cbd8313a8cf45b3cf66ddef2bb06e07 -c sh -c 'nix --version; time nix build'
nix (Nix) 2.4pre20210902_90b2dd5
2.10user 0.59system 0:02.60elapsed 103%CPU (0avgtext+0avgdata 287372maxresident)k
0inputs+402536outputs (0major+115786minor)pagefaults 0swaps

$ echo >> flake.nix; git commit -am foo
[fvis 07a735a0b9] foo
 1 file changed, 1 insertion(+)

$ nix shell github:NixOS/nix/6678e98411cd3bfe8143a4ba1b35d1451babce32 -c sh -c 'nix --version; time nix build'
nix (Nix) 2.4pre20210907_6678e98
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint: 	git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint: 	git branch -m <name>
94.89user 2.88system 0:35.60elapsed 274%CPU (0avgtext+0avgdata 1234872maxresident)k
264inputs+402224outputs (1major+488572minor)pagefaults 0swaps
```
This is due to https://github.com/NixOS/nix/pull/4922, which enables cloning submodules by default. But the `submodules` option is [extremely inefficient](https://github.com/nrdxp/nix/blob/769ca4e26d973660756de85b51efdb7fc36a00fd/src/libfetchers/git.cc#L436-L440), *even if the repo does not even contain any submodules*.

I know there has been some discussion on the PR itself after merging, but I felt that this regression deserved its own issue due to its severity.

**Steps To Reproduce**

```
git clone https://github.com/leanprover/lean4
cd lean4
nix flake prefetch
```

**Expected behavior**

A much faster store copy, as before

[...]

PR #5284, opened 2021-09-22
Revert "Merge pull request #4922 from nrdxp/default-submodules"

edolstra:
#4922 causes a few issues, see [here](https://github.com/NixOS/nix/pull/4922#issuecomment-921038895) and [here](https://github.com/NixOS/nix/issues/5280). So it seems best to revert for now and figure out what to do after the 2.4 release.

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -36,7 +36,7 @@ void emitTreeAttrs(
 
     if (input.getType() == "git")
         mkBool(*state.allocAttr(v, state.symbols.create("submodules")),
-            fetchers::maybeGetBoolAttr(input.attrs, "submodules").value_or(true));
+            fetchers::maybeGetBoolAttr(input.attrs, "submodules").value_or(false));
 
     if (!forceDirty) {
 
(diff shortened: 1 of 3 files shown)


knedlsepp:
😕 somewhat on the other end of the spectrum that not having submodule support is a showstopper.

nrdxp:
I don't really understand the complaints personally, and they can be easily worked around with an alternate fetcher. I don't know why you'd want to pull all of llvm as a flake input anyway :confused: 

But alas, we are using flakes in production and this broke _our_ workflow so we are probably going to maintain a patch and build our own Nix on our Hydra until a better solution emerges unfortunately. I originally opened the PR so we _wouldn't_ have to do this.

I'd really love to study up on C++ and solve this myself, but Nix is an especially intimidating codebase and my backlog just keeps growing unfortunately :facepalm:

alexbakker:
Unfortunately the commit being reverted here was already included in nixpkgs' ``nixUnstable``, so this will break my workflow as well.

knedlsepp:
@edolstra As there are plans to stabilise flakes soonish, I'd like to shed light again on this unfortunate revert. I'm really hoping that we can somehow make git submodules enabled by default before flakes is declared stable. I imagine that changing defaults for submodules is a breaking change and the current UX of `nix build  "git+file://$(pwd)?submodules=1"` is pretty sub-optimal compared to `nix build`. It would be unfortunate if we'll be stuck with it.

PR #5286, opened 2021-09-22
Warn about the usage of filterSource with Nix store paths

ilkecan:
There is a lot of confusion caused by the behaviour of `filterSource` when used on a nix store path. I guess this is becoming more common with the increased usage of flakes, where the repository is moved to the store before the evaluation. This PR adds a warning that tries to explain why `filterSource` behaves unexpectedly and points the users to `builtins.path` instead.

Non exhaustive list of people's confusion:
https://discourse.nixos.org/t/flakes-and-filtersource/7787
https://discourse.nixos.org/t/builtins-filtersource-causes-rebuilds/12564
https://discourse.nixos.org/t/how-to-make-src-in-a-flake-nix-not-change-a-lot/15129


src/libexpr/primops.cc
```diff
@@ -1919,6 +1919,20 @@ static RegisterPrimOp primop_filterSource({
     .name = "__filterSource",
     .args = {"e1", "e2"},
     .doc = R"(
+      > **Warning**
+      >
+      > `filterSource` should not be used to filter Nix output paths
+      > (directories whose parent is `/nix/store`). Since `filterSource`
```
edolstra, line 1925:
```suggestion
      > `filterSource` should not be used to filter store paths. Since `filterSource`
```

edolstra (approved):

src/libexpr/primops.cc
ilkecan: Fixed it.

PR #5295, opened 2021-09-26
build: also throw hash-mismatch errors if `buildMode == bmCheck`

Ma27:
This actually bit me quite recently in `nixpkgs` because I assumed that
`nix-build --check` would also error out if hashes don't match anymore[1]
and so I wrongly assumed that I couldn't reproduce the mismatch error.

The fix is rather simple, during the output registration a so-called
`delayedException` is instantiated e.g. if a FOD hash-mismatch occurs.
However, in case of `nix-build --check` (or `--rebuild` in case of `nix
build`), the code-path where this exception is thrown will never be
reached.

By adding that check to the if-clause that causes an early exit in case
of `bmCheck`, the issue is gone. Also added a (previously failing)
test-case to demonstrate the problem.

[1] https://github.com/NixOS/nixpkgs/pull/139238, the underlying issue
    was that `nix-prefetch-git` returns different hashes than `fetchgit`
    because the latter one fetches submodules by default.

cc @edolstra @Ericson2314

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -2474,7 +2474,13 @@ void LocalDerivationGoal::registerOutputs()
         infos.emplace(outputName, std::move(newInfo));
     }
 
-    if (buildMode == bmCheck) return;
+    if (buildMode == bmCheck) {
+        // In case of FOD mismatches on `--check` an error must be thrown as this is also
+        // a source for non-determinism.
+        if (delayedException)
+            std::rethrow_exception(delayedException);
+        return;
+    }
 
     /* Apply output checks. */
     checkOutputs(infos);
--- a/tests/fixed.nix
+++ b/tests/fixed.nix
@@ -21,6 +21,14 @@ rec {
     (f ./fixed.builder2.sh "recursive" "sha1" "vw46m23bizj4n8afrc0fj19wrp7mj3c0")
   ];
 
+  # Expression to test that `nix-build --check` also throws an error if the hash of
+  # fixed-output derivation has changed even if the hash exists in the store (in this
+  # case the hash exists because of `fixed.builder2.sh`, but building a derivation
+  # with the same hash and a different result must throw an error).
+  check = [
+    (f ./fixed.builder1.sh "recursive" "md5" "3670af73070fa14077ad74e0f5ea4e42")
+  ];
+
   good2 = [
     # Yes, this looks fscked up: builder2 doesn't have that result.
     # But Nix sees that an output with the desired hash already
--- a/tests/fixed.sh
+++ b/tests/fixed.sh
@@ -15,6 +15,11 @@ nix path-info --json $path | grep fixed:md5:2qk15sxzzjlnpjk9brn7j8ppcd
 echo 'testing good...'
 nix-build fixed.nix -A good --no-out-link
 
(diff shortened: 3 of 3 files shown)


Ma27:
Hmm, `make installcheck` actually worked for me. Will take a look tomorrow.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -2474,7 +2474,13 @@ void LocalDerivationGoal::registerOutputs()
         infos.emplace(outputName, std::move(newInfo));
     }
 
-    if (buildMode == bmCheck) return;
+    if (buildMode == bmCheck) {
+        // In case of FOD mismatches on `--check` an error must be thrown as this is also
+        // a source for non-determinism.
+        if (delayedException)
+            std::rethrow_exception(delayedException);
```
regnat, line 2481: It’d be nice in that case to register the invalid path before throwing (like we do when not `--check`-ing). But I’m not sure whether that can be done without complicating the code a lot, so I’ll leave that up to you

regnat (changes requested):

Ma27:
OK, `nix build -L` on this branch works fine for me. @regnat would you mind looking into why the GHA build fails? :)

tests/fixed.nix
```diff
@@ -21,6 +21,10 @@ rec {
     (f ./fixed.builder2.sh "recursive" "sha1" "vw46m23bizj4n8afrc0fj19wrp7mj3c0")
   ];
 
+  check = [
```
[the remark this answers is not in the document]
Ma27: Done!

src/libstore/build/local-derivation-goal.cc
Ma27: Which code-path do you refer to? I guess it makes sense since otherwise this path could be re-used in a later build, or am I misunderstanding this? :)

regnat:
> OK, `nix build -L` on this branch works fine for me. @regnat would you mind looking into why the GHA build fails? :)

I think the failure is in one of the flake checks, not in the build of Nix itself. Probably `checks.x86_64-linux.installTests.againstSelf` (which runs the testsuite but with the daemon everywhere).

Ma27:
Ah good point, then I'll take a look later at this :)

regnat:
> Probably checks.x86_64-linux.installTests.againstSelf (which runs the testsuite but with the daemon everywhere).

Gah sorry, it’s `checks.x86_64-linux.installTests.againstCurrentUnstable` which fails (because as it’s using an old daemon, the test fails). You can make it conditional on the daemon version with something like `(! isDaemonNewer "2.4pre20210927") || test …`

src/libstore/build/local-derivation-goal.cc
regnat: What I say is that when `--check` is used and there’s a hash mismatch, then the mismatched path isn’t registered at all (and as you say, it would make sense to do so). But it’s not really critical given that it already wasn’t the case

Ma27:
Thanks, fixed @regnat! :)

regnat (approved):

Ma27:
AFAICS the macOS issue doesn't seem related.

regnat:
> AFAICS the macOS issue doesn't seem related.

It doesn’t seem indeed. I’ve restarted the job

PR #5301, opened 2021-09-27
libexpr: throw a more helpful eval-error if a builtin is not available due to a missing feature-flag

Ma27:
I found it somewhat confusing to have an error like

    error: attribute 'getFlake' missing

if the required experimental-feature (`flakes`) is not enabled. Instead,
I'd expect Nix to throw an error just like it's the case when using e.g. `nix
flake` without `flakes` being enabled.

With this change, the error looks like this:

    $ nix-instantiate -E 'builtins.getFlake "nixpkgs"'
    error: Experimental Nix feature 'flakes' is missing to call builtins.getFlake! You can it via '--extra-experimental-features flakes'.

           at «string»:1:1:

                1| builtins.getFlake "nixpkgs"
                 | ^

I didn't use `settings.requireExperimentalFeature` here on purpose
because this doesn't contain a position. Also, it doesn't seem as if we
need to catch the error and check for the missing feature here since
this already happens at evaluation time.

cc @edolstra @regnat

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -465,6 +465,23 @@ EvalState::~EvalState()
 }
 
 
+void EvalState::requireExperimentalFeatureOnEvaluation(
+    const std::string & feature,
+    const std::string_view fName,
+    const Pos & pos)
+{
+    if (!settings.isExperimentalFeatureEnabled(feature)) {
+        throw EvalError({
+            .msg = hintfmt(
+                "Cannot call '%2%' because experimental Nix feature '%1%' is disabled. You can enable it via '--extra-experimental-features %1%'.",
+                feature,
+                fName
+            ),
+            .errPos = pos
+        });
+    }
+}
+
 Path EvalState::checkSourcePath(const Path & path_)
 {
     if (!allowedPaths) return path_;
--- a/src/libexpr/eval.hh
+++ b/src/libexpr/eval.hh
@@ -140,6 +140,12 @@ public:
         std::shared_ptr<Store> buildStore = nullptr);
     ~EvalState();
 
+    void requireExperimentalFeatureOnEvaluation(
+        const std::string & feature,
+        const std::string_view fName,
+        const Pos & pos
+    );
+
     void addToSearchPath(const string & s);
 
     SearchPath getSearchPath() { return searchPath; }
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -688,6 +688,8 @@ void callFlake(EvalState & state,
 
 static void prim_getFlake(EvalState & state, const Pos & pos, Value * * args, Value & v)
 {
+    state.requireExperimentalFeatureOnEvaluation("flakes", "builtins.getFlake", pos);
+
     auto flakeRefS = state.forceStringNoCtx(*args[0], pos);
     auto flakeRef = parseFlakeRef(flakeRefS, {}, true);
     if (evalSettings.pureEval && !flakeRef.input.isImmutable())
@@ -703,7 +705,7 @@ static void prim_getFlake(EvalState & state, const Pos & pos, Value * * args, Va
         v);
 }
 
-static RegisterPrimOp r2("__getFlake", 1, prim_getFlake, "flakes");
+static RegisterPrimOp r2("__getFlake", 1, prim_getFlake);
 
 }
 
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -3606,15 +3606,13 @@ static RegisterPrimOp primop_splitVersion({
 RegisterPrimOp::PrimOps * RegisterPrimOp::primOps;
 
 
-RegisterPrimOp::RegisterPrimOp(std::string name, size_t arity, PrimOpFun fun,
-    std::optional<std::string> requiredFeature)
+RegisterPrimOp::RegisterPrimOp(std::string name, size_t arity, PrimOpFun fun)
 {
     if (!primOps) primOps = new PrimOps;
     primOps->push_back({
         .name = name,
         .args = {},
         .arity = arity,
-        .requiredFeature = std::move(requiredFeature),
         .fun = fun
     });
 }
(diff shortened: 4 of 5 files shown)


regnat (commented):
The logic seems right, but I’m (very vaguely) worried about a possible performance impact on the evaluation (because the check for a missing feature needs to happen at each primop call).

Maybe you could rewrite the primop registration logic as something along the lines of:

```c++
            if (!primOp.requiredFeature || settings.isExperimentalFeatureEnabled(*primOp.requiredFeature))
                addPrimOp({
                    .fun = primOp.fun,
                    ...
                });
            else
                addPrimOp({
                    .fun = [](EvalState foo, Pos pos, Value ** args, Value & v){ throw EvalError("Builtin %s requires the experimental feature %s", primOp.name, *primOp.requiredFeature),
                     ...
                 });
                    ...
                });
```

so that we don’t need to check anything anymore when calling it

Ma27:
@regnat this is basically what I wanted to implement first, however it didn't really work out:

* In order to throw an EvalError with the missing feature and primop name, the function needs access to the `primOp` which has to be done via a capture expression (i.e. `[&]()` or something similar).
* However, a lambda with a capture expression cannot be cast into a function pointer (which is the type of `PrimOpFun`). AFAIU this is because a function pointer cannot hold any "state" such as variables from the upper scope.
* Hacking around with static variables doesn't work out either because that'll lead us to bogus results as soon as we have multiple primops that are only available if a feature-flag is enabled (AFAICS only `getFlake` is affected currently).
* I tried to convert `PrimOpFun` into a `std::function` which fixed the compilation error, however I ran into weird memory corruptions that caused a segfault in the end when trying to call  `getFlake` without `flakes` being enabled (I think this is because of `primOp` being copied into the function's closure somehow).

In the end, this is why I left `PrimOpFun` to be a function-pointer. Unless someone of you has a better idea - I mean, I'm not very seasoned in C++, so it's possible that I totally missed something - I think we could add a `isFunctionEnabled`-field to the `PrimOp` struct so that we don't have to check for feature-flags on each invocation, but only if a boolean is true/false.

src/libexpr/eval.hh
```diff
@@ -31,6 +31,7 @@ struct PrimOp
     Symbol name;
     std::vector<std::string> args;
     const char * doc = nullptr;
+    std::optional<std::string> missingFeature;
```
regnat, line 34:
```suggestion
    std::optional<std::string> requiredFeature;
```

(Unless you really want it to mean `missingFeature`, in which case I assume it should only be non-null if the feature is actually missing, but that’s not what the code is doing currently)

regnat (changes requested):
> this is basically what I wanted to implement first, however it didn't really work out

Right. I think we could make this work by ensuring that we capture by copy everything that needs to be, but we can leave it out for now.

I just have one minor naming comment then, otherwise it’s good-to-go

edolstra:
This breaks the intent of why it's implemented the way it is: namely to allow Nix expressions to check for the availability of a feature by doing e.g. `builtins ? getFlake`. If it wasn't for that, we could just call `requireExperimentalFeature("foo")` inside the primop, no need for any special handling in `callPrimOp()`.

Ma27:
@edolstra I see your point and yeah, it's probably even better to just use `requireExperimentalFeature` (perhaps with some positioning information).

However, I think it's a rather obscure way to check for feature-availability via `builtins ? getFlake` for two reasons:

* It's not really clear for the reader of such an expression why this is done that way (i.e. on each occasion a comment is needed).
* This only works for flakes. I could imagine that at least for `recursive-nix` & `ca-derivaitons` (or some features in the future) it may be helpful to have a way to check for feature-availability.

Hence my suggestion to
* use `requireExerpimentalFeature` as you suggested
* add a new builtin `builtins.isExperimentalFeatureEnabled` (or a catchier name, idk :p)

From my PoV this can be done **until** 2.4 is out, after that it's difficult because changing the outcome of `builtins ? getFlake` also affects expressions not using experimental features and is thus a breaking change in the expression language itself.

WDYT? :)

edolstra:
It just occurred to me that allowing any kind of feature check in Nix expressions (whether `builtins ? foo` or `builtins.isExperimentalFeatureEnabled`) is bad for purity, since it allows writing code like "if CA derivations are enabled then return a CA derivation else return a non-CA derivation". This breaks reproducibility and the evaluation cache (which currently doesn't take features into account).

So it's probably best to turn this into an uncatchable error using `requireExperimentalFeature`.

Ma27:
Fair. Will update the PR accordingly later today :)

Ma27:
@edolstra @regnat I implemented an alternative `requireExperimentalFeature` that throws an evaluation-error with a `Pos`. AFAICS this isn't an error that will be caught somewhere else where it's checked which feature is missing.

src/libexpr/eval.cc
```diff
@@ -465,6 +465,24 @@ EvalState::~EvalState()
...
+void EvalState::requireExperimentalFeatureOnEvaluation(
+    const std::string & feature,
+    const std::optional<std::string> & fName,
+    const Pos & pos)
+{
+    if (!settings.isExperimentalFeatureEnabled(feature)) {
+        auto f = fName.value_or("a function");
+        throw EvalError({
+            .msg = hintfmt(
+                "Experimental Nix feature '%1%' is missing to call %2%! You can it via '--extra-experimental-features %1%'.",
```
greedy, line 477: There's a missing word:

```suggestion
                "Experimental Nix feature '%1%' is missing to call %2%! You can enable it via '--extra-experimental-features %1%'.",
```

An alternative wording:

```suggestion
                "Experimental Nix feature '%1%' must be enabled to use %2%! You can enable it via '--extra-experimental-features %1%'.",
```

Ma27:
Applied, thanks!

src/libexpr/eval.cc
```diff
@@ -465,6 +465,24 @@ EvalState::~EvalState()
 }
 
 
+void EvalState::requireExperimentalFeatureOnEvaluation(
+    const std::string & feature,
+    const std::optional<std::string> & fName,
```
edolstra, line 470:
```suggestion
    std::string_view fName,
```

since I don't really see a situation where this could be unset.

src/libexpr/eval.cc
```diff
@@ -465,6 +465,24 @@ EvalState::~EvalState()
...
+    const Pos & pos)
+{
+    if (!settings.isExperimentalFeatureEnabled(feature)) {
+        auto f = fName.value_or("a function");
+        throw EvalError({
+            .msg = hintfmt(
+                "Experimental Nix feature '%1%' is missing to call %2%! You can enable it via '--extra-experimental-features %1%'.",
```
edolstra, line 477:
```suggestion
                "Cannot call '%2%' because experimental Nix feature '%1%' is disabled. You can enable it via '--extra-experimental-features %1%'.",
```

Ma27:
@edolstra applied your suggestions, thanks!

edolstra:
Thanks!

Issue #5300, opened 2021-09-27
Wrong line number shown in syntax error message


oxalica:
**Describe the bug**

Error message shows on wrong line.

**Steps To Reproduce**

1. Create `test.nix` with content:
```nix
{
  imports = [
    ./foo.nix
    ./bar.nix
    ./baz.nix
  ];
  foo = then; # Keyword `then` triggers a syntax error.
  # Some
  # More
  # Comments
  # After
}
```

2. Run `nix-instantiate --parse ./test.nix`

3. See the error message shows an error on line 10 `# Comments` instead of line 7 `foo = then;`.
```
error: syntax error, unexpected THEN

       at /home/oxa/repo/fork/nix/test.nix:10:9:

            9|   # More
           10|   # Comments
             |         ^
           11|   # After
```

4. Comment out three paths in array of `imports` gives correct error location of `then`. The issue seems related to paths.

**Expected behavior**

```
error: syntax error, unexpected THEN

       at /home/oxa/repo/fork/nix/test.nix:7:9:

            6|   ];
            7|   foo = then; # Keyword `then` triggers a syntax error.
             |         ^
            8|   # Some
```

**`nix-env --version` output**

I'm on current master.
nix (Nix) 2.4pre20210927_9c766a4

domenkozar:
Cc @Radvendii

Radvendii:
Yep, my change makes it read the character after the path twice. Working on a fix.

PR #5307, opened 2021-09-29
reset yylloc when yyless(0) is called

Radvendii:
Fixes https://github.com/NixOS/nix/issues/5300

See also [this stackoverflow answer](https://stackoverflow.com/questions/59754253/yyllocp-first-line-returns-uninitialized-value-in-second-iteration-of-a-reentra) for another solution to the same problem. This seemed like a simpler / more minimal change though.

### DIFF
--- a/src/libexpr/lexer.l
+++ b/src/libexpr/lexer.l
@@ -28,6 +28,8 @@ using namespace nix;
 
 namespace nix {
 
+// backup to recover from yyless(0)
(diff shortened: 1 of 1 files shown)


Radvendii:
I can't really make sense of the Mac error but it seems to be unrelated?

edolstra:
Thanks. Yeah you can ignore the Mac error.

Issue #5260, opened 2021-09-16
flakes: git filters broken


alarsyo:
**Describe the bug**

If you have a problem with a specific package or NixOS,
you probably want to file an issue at https://github.com/NixOS/nixpkgs/issues. 

**Steps To Reproduce**

**Expected behavior**

**`nix-env --version` output**

**Additional context**

nrdxp:
why would fetching submodules break git-crypt? Do you have your secrets in a submodule?

plabadens:
I'm able to reproduce the issue. Somehow this has changed the way the git repo gets moved to the nix store.

nrdxp:
oh I guess I can see that, since it probably copies the repo from the git history (which has the secrets encrypted by default), instead of just a plain copy. For the sake of investigation, is this broken even when your file tree is dirty? I would imagine the behavior would be different in this case since it can't simply rely on the commit history since dirty changes don't exist there yet.

plabadens:
I'm currently doing a bit of testing, but it seems to evaluate perfectly fine when the worktree is dirty, as you expected. Using `git-crypt` was always a bit of a hack and I should probably be moving to something like `sops-nix` anyway.

alarsyo:
Yes, I should also move to sops-nix or age-nix, but having a dirty worktree is indeed a great workaround for the time being, thanks for the tip!

miuirussia:
I also encountered this bug, rolled back the changes that led to an error using the overlay: https://github.com/miuirussia/nix-flake-env/blob/master/overlays/nixUnstable/default.nix

nuance:
I was using git-lfs to manage some binaries in my nix config, which this also appears to break. Reverting the commit fixes the issue.

nrdxp:
the commit is reverted in master so we can close this I think.

roberth:
Git-crypt should _not_ work, at least not with the way flakes are currently implemented. Nix Flakes will write your secrets unconditionally to the store, which is not a secure place.
Also, like arbitrary smudge filters, it is not reproducible. See https://github.com/NixOS/nix/pull/4635 for a more detailed argumentation.

I would like for git-crypt to work _somehow_, but it would take special treatment of local flakes, which I think it is too early to consider that at this point.

lunik1:
Something of a workaround for the `git-crypt` case is to ensure the encrypted file is dirty before you evaluate the flake; this way the unencrypted file is evaluated. Of course, your secrets will still be globally readable in the nix store.

[...]

PR #5308, opened 2021-09-30
Nix 2.4 release notes

### LINKED ISSUE #4521: Write Nix 2.4 release notes

edolstra:
Fixes #4521.

### DIFF
--- a/doc/manual/src/release-notes/rl-2.4.md
+++ b/doc/manual/src/release-notes/rl-2.4.md
@@ -23,7 +23,9 @@ more than 2800 commits from 195 contributors since release 2.3.
 * The **`nix` command** has seen a lot of work and is now almost at
   feature parity with the old command-line interface (the `nix-*`
   commands). It aims to be [more modern, consistent and pleasant to
-  use](../contributing/cli-guideline.md) than the old CLI.
+  use](../contributing/cli-guideline.md) than the old CLI. It is still
+  marked as experimental but its interface should not change much
+  anymore in future releases.
 
 * **Flakes** are a new format to package Nix-based projects in a more
   discoverable, composable, consistent and reproducible way. A flake
@@ -180,8 +182,9 @@ more than 2800 commits from 195 contributors since release 2.3.
 * **New built-in functions**:
 
   - `builtins.fetchTree` allows fetching a source tree using any
-    backends supported by the fetcher infrastructure. It replaces
-    `fetchGit`, `fetchMercurial` and `fetchTarball`.
+    backends supported by the fetcher infrastructure. It subsumes the
+    functionality of existing built-ins like `fetchGit`,
+    `fetchMercurial` and `fetchTarball`.
 
   - `builtins.getFlake` fetches a flake and returns its output
     attributes. This function should not be used inside flakes! Use
@@ -247,6 +250,8 @@ more than 2800 commits from 195 contributors since release 2.3.
 
   - With `-p`, it passes any `--arg`s as Nixpkgs arguments.
 
+  - Support for structured attributes.
+
 * `nix-prefetch-url` has a new `--executable` flag.
 
 * On `x86_64` systems, [`x86_64` microarchitecture
@@ -281,7 +286,8 @@ more than 2800 commits from 195 contributors since release 2.3.
   to your `nix.conf` if you want to use it, or pass
   `--extra-experimental-features nix-command` on the command line.
 
-* `nix run` has been renamed to `nix shell`.
+* The old `nix run` has been renamed to `nix shell` (and there is a
+  new `nix run` that does something else, as described above).
 
 * It is now an error to modify the `plugin-files` setting via a
   command-line flag that appears after the first non-flag argument to
@@ -290,7 +296,7 @@ more than 2800 commits from 195 contributors since release 2.3.
   `nix-instantiate --plugin-files "" default.nix`.
 
 * We no longer release source tarballs. If you want to build from
-  source, please build from the Git repository.
+  source, please build from the tags in the Git repository.
 
 ## Contributors
 
--- a/doc/manual/src/release-notes/rl-2.4.md
+++ b/doc/manual/src/release-notes/rl-2.4.md
@@ -47,7 +47,7 @@ more than 2800 commits from 195 contributors since release 2.3.
   contents. Previously Nix could only build **input-addressed** store
   paths, where the hash is computed from the derivation dependency
   graph. Content-addressing allows deduplication, early cutoff in
-  build systems, and unpriviliged closure copying. This is still [an
+  build systems, and unprivileged closure copying. This is still [an
   experimental
   feature](https://discourse.nixos.org/t/content-addressed-nix-call-for-testers/12881).
 
@@ -96,7 +96,7 @@ more than 2800 commits from 195 contributors since release 2.3.
     program from the `hello` package in `nixpkgs`.
 
   - `nix flake` is the container for flake-related operations, such as
-    creating a new flake, quering the contents of a flake or updating
+    creating a new flake, querying the contents of a flake or updating
     flake lock files.
 
   - `nix registry` allows you to query and update the flake registry,
@@ -116,7 +116,7 @@ more than 2800 commits from 195 contributors since release 2.3.
       changes.
 
     **Warning**: after a profile has been updated using `nix profile`,
-    it is no longer useable with `nix-env`.
+    it is no longer usable with `nix-env`.
 
   - `nix store diff-closures` shows the differences between the
     closures of two store paths in terms of the versions and sizes of
@@ -205,7 +205,7 @@ more than 2800 commits from 195 contributors since release 2.3.
 * Nix now uses `libarchive` to decompress and unpack tarballs and zip
   files, so `tar` is no longer required.
 
-* The priority of substituters can now be overriden using the
+* The priority of substituters can now be overridden using the
   `priority` substituter setting (e.g. `--substituters
   'http://cache.nixos.org?priority=100 daemon?priority=10'`).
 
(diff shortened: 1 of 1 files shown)


doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+             87|           [ curl
+             88|             bzip3 xz brotli editline
+               |             ^
+             89|             openssl sqlite
+  ```
+
+* The **`nix` command** has seen a lot of work and is now almost at
+  feature parity with the old command-line interface (the `nix-*`
+  commands). It aims to be [more modern, consistent and pleasant to
+  use](../contributing/cli-guideline.md) than the old CLI.
```
Ma27, line 26: IIRC the command is still considered experimental though, correct? If that's supposed to remain as-is in 2.4 then it should be mentioned here as well.

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+
+* A new setting that allows **experimental features** to be enabled
+  selectively. This allows us to merge unstable features into Nix more
+  quickly and do more frequent releases.
+
+## Other features
+
+* There are many new `nix` subcommands:
+
+  - `nix develop` is intended to replace `nix-shell`. It has a number
```
Ma27, line 64: I guess it makes sense to mention `nix shell`/`nix run` here as well (at least the first one) since these are also providing subsets of features of `nix-shell`.

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+
+* **Languages changes**: the only new language feature is that you can
+  now have antiquotations in paths, e.g. `./${foo}` instead of `./. +
+  foo`.
+
+* **New built-in functions**:
+
+  - `builtins.fetchTree` allows fetching a source tree using any
+    backends supported by the fetcher infrastructure. It replaces
+    `fetchGit`, `fetchMercurial` and `fetchTarball`.
```
Ma27, line 184: I guess it's worth mentioning that `fetchGit`/etc are still available, but considered deprecated (I guess?)

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+
+* When auto-calling a function with an ellipsis, all arguments are now
+  passed.
+
+* New `nix-shell` features:
+
+  - It preserves the `PS1` environment variable if
+    `NIX_SHELL_PRESERVE_PROMPT` is set.
+
+  - With `-p`, it passes any `--arg`s as Nixpkgs arguments.
```
Ma27, line 248: (shameless plug) I guess it's worth mentioning that nix-shell (as well as nix develop) now support derivations using `__structuredAttrs = true;` (ref https://github.com/NixOS/nix/pull/4770)

doc/manual/src/release-notes/rl-2.4.md
roberth: Removing builtins is a bad idea because old expressions won't be buildable anymore. Being able to build expressions from ages ago is a remarkable and unique capability.
In Nixpkgs we don't have this constraint, because legacy software can pin an older version but still be integrated into a system based on `master`, but when the Nix language or its semantics change, this can't be done.
Deprecation would imply removal. Otherwise, we're just bugging people for no real reason.

roberth: For the same reasons, I'm very concerned about the unreproducible behavior of `builtins.fetchGit` and/or `fetchTree`. Too little consideration is given to the smudge filters, running them only in arbitrary circumstances which [change](https://github.com/NixOS/nix/issues/5260), while the only sane default is not to run them unless specific supported ones are explicitly requested, such as git-lfs.

Ma27: Agreed. I just wanted to point out that `It replaces fetchGit, [...]` sounds exactly like it's not available anymore (to me at least, though I'm not a native speaker).

Btw, I fully agree that builtins shouldn't be removed because old Nix expressions should remain evaluatable.

abathur:
I don't know if the release note gets circulated widely enough that it's important to nail the descriptions of `nix develop`, `nix shell`, and `nix run`, but this discussion on discourse (from the linked post to the end) has some good context on how people are making sense of these: https://discourse.nixos.org/t/difference-between-nix-shell-and-nix-run/2951/14

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+* The new experimental feature `no-url-literals` disables URL
+  literals. This helps to implement [RFC
+  45](https://github.com/NixOS/rfcs/pull/45).
+
+* Nix now uses `libarchive` to decompress and unpack tarballs and zip
+  files, so `tar` is no longer required.
+
+* The priority of substituters can now be overriden using the
+  `priority` substituter setting (e.g. `--substituters
+  'http://cache.nixos.org?priority=100 daemon?priority=10'`).
```
SuperSandro2000, line 207:
```suggestion
  'http://cache.nixos.org?priority=100&daemon?priority=10'`).
```
Should this be like this?

doc/manual/src/release-notes/rl-2.4.md
samueldr: Maybe it should be worded in a manner to imply "you're invited to test it with your use cases to help snuff out unlikely issues with esoteric use cases".

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+* `nix run` has been renamed to `nix shell`.
+
+* It is now an error to modify the `plugin-files` setting via a
+  command-line flag that appears after the first non-flag argument to
+  any command, including a subcommand to `nix`. For example,
+  `nix-instantiate default.nix --plugin-files ""` must now become
+  `nix-instantiate --plugin-files "" default.nix`.
+
+* We no longer release source tarballs. If you want to build from
+  source, please build from the Git repository.
```
trofi, line 293: I was confused in https://github.com/NixOS/nix/issues/3658#issuecomment-641890996 and am still confused reading this wording. It sounds like you plan to stop doing any releases (including tagging any releases after 2.4). Is it correct?

Or you still plan to `git tag` future releases? What should users of non-NixOS distros do when they want nix from source? Use tagged releases or always use `matser` branch?

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+
+* The `nix` command is now marked as an experimental feature. This
+  means that you need to add
+
+  > experimental-features = nix-command
+
+  to your `nix.conf` if you want to use it, or pass
+  `--extra-experimental-features nix-command` on the command line.
+
+* `nix run` has been renamed to `nix shell`.
```
tomberek, line 284:
```suggestion
* Legacy `nix run` command has been renamed to `nix shell`.
```

Ericson2314:
Not sure if this means we're about to cut a branch, but in case it does I started cleaning up some things:

 - https://github.com/NixOS/nix/pull/4589 is now just testing, always good to test more things!
 - https://github.com/NixOS/nix/pull/5311 might help with performance, ensuring `std::visit` doesn't result in extra copying w.r.t. to Nix 2.3.

doc/manual/src/release-notes/rl-2.4.md
edolstra: @roberth Make an issue about it, the release notes are not the place to discuss this.

edolstra:
@Ericson2314 Do you mean you want those to be merged before or after branching?

doc/manual/src/release-notes/rl-2.4.md
edolstra: No, it means what it says: we don't release source tarballs (i.e. with autoconf-generated stuff in them). Of course we still tag releases in git...

trofi: That's clear. Thank you!

doc/manual/src/release-notes/rl-2.4.md
roberth: I agree, so I've created #5313. I commented here because it would catch your attention, as I did raise the issue before, albeit in a PR.

roberth: > Deprecation would imply removal. Otherwise, we're just bugging people for no real reason.

I was wrong about this. `fetchGit` should remain untouched, to preserve some degree of reproducibility, but it is broken so it _should_ be avoided, except when building legacy software, in which case a warning isn't too much of a nuisance.

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -1,8 +1,492 @@
...
+
+* The new `--log-format` flag selects what Nix's output looks like. It
+  defaults to a terse progress indicator. There is a new
+  `internal-json` output format for use by other programs.
+
+* `nix eval` has a new `--apply` flag that applies a function to the
+  evaluation result.
+
+* `nix eval` has a new `--write-to` flag that allows it to write a
+  nested attribute set of string leaves to a correspondingly directory
```
Kha, line 224:
```suggestion
  nested attribute set of string leaves to a corresponding directory
```

doc/manual/src/release-notes/rl-2.4.md
edolstra: Changed the wording, there is no intent to remove `fetch*`.

Ericson2314:
@edolstra Before.

Issue #5118, opened 2021-08-10
Substituter query parallelism is far too low


Kha:
TL;DR: a single-line change can make binary cache querying 50x as fast

Continuing from #5109, I decided to investigate *why* substitution (querying) is/seems to be fairly slow. Okay, maybe my 800 derivations are simply a bit much, I thought, but then I found @edolstra's [commit message](https://github.com/kha/nix/commit/90ad02bf626b885a5dd8967894e2eafc953bdf92) stating that HTTP/2 should handle this in under a second. So what's going on? The async API described in that commit never really got put to use as far as I can see, so it turns out that throughput is (*severely*) limited by the size of the [thread pool](https://github.com/kha/nix/blob/a6ba313a0aac3b6e2fef434cb42d190a0849238e/src/libstore/misc.cc#L103) used to fetch path infos in parallel via the synchronous API. The standard size of that pool is the number of hardware threads (8 on this machine), but we can improve throughput by >50x by massively scaling up that pool, at least in my use case!

#threads | "querying missing paths"
--- | ---
8 | 38s
32 | 10s
100 | 1.89s
300 | 0.89s
1000 | 0.73s

Now I can't talk about any substituters other than HTTP/2 binary caches and I don't know how exactly this should be fixed (should the pool size be a new option (with a high default or low default?), should the thread pool be replaced with true asynchrony, should doing the same for actual substitution be investigated as well, should substitution querying and actual substitution be merged after all, ...), but obviously I would be very happy about *any* solution removing this bottleneck and would be willing to implement it when there is consensus on the design.

domenkozar:
Can you benchmark more entries between 8 and 100? That would yield a good default.

Without changing the design, I'd propose to chose a better default and make the pool size configurable.

From the Cachix side it would be best if Nix implemented batched lookups, that would probably require something like returning JSON with a list of narinfo entries. That would also require adding something like `DynamicQuerying: true` to `nix-cache-info`.

michaelpj:
Possibly also relevant: https://github.com/NixOS/nix/issues/3019

Kha:
> Can you benchmark more entries between 8 and 100? That would yield a good default.

I added another entry for 32 threads, though the benchmark should probably be redone to ensure consistent conditions. I'm a bit suspicious of the 100 threads entry now, its cumulative time (threads * total time) is much lower than the other entries. I've been recompiling Nix for each change in thread count so far :grimacing: .

[...]

PR #5324, opened 2021-10-03
Make number of parallel substituter queries configurable

Kha:
This is the minimal fix for #5118. The default is taken from `http-connections`, though as stated, multiplexing can mean that a far higher number is even more effective. Still, the new default should be a sizeable improvement over the old one of `nproc` for most machines while still low enough that time and memory overhead from the additional threads should be insignificant.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -250,6 +250,17 @@ public:
         )",
         {"build-use-substitutes"}};
 
+    Setting<unsigned int> maxQuerySubstitutersJobs{
+        this, 25, "max-query-substituters-jobs",
+        R"(
+          This option defines the maximum number of parallel queries Nix will
+          use to ask substituters for the presence of substitutes (which is a
+          separate step from actually downloading substitutes). For builds of
+          many derivations backed by substituters supporting HTTP/2,
+          significant speedups of this step have been observed by increasing
+          this option thanks to request multiplexing.
+        )"};
+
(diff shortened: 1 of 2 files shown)


src/libstore/globals.hh
```diff
@@ -250,6 +250,17 @@ public:
         )",
         {"build-use-substitutes"}};
 
+    Setting<unsigned int> maxQuerySubstitutersJobs{
```
SuperSandro2000, line 253: Wouldn't it be better to just check if the substituter supports http/2 and then significantly increase this setting?

Kha: The better solution is not to use threads at all here, which would obviate the need for any parallelism guessing. This is the trivial fix hopefully in time for Nix 2.4 that at least gives people an option to max out the connections. Any more complex solution should come later.

domenkozar:
@edolstra thoughts?

@Kha could you rebase on top of master to fix CI?

domenkozar:
@edolstra any reason to hold this?

edolstra:
It's probably better to rewrite `Store::queryMissing()` to be single-threaded, similar to what we did with `Store::computeFSClosure()` in 75989bdca773eedb8b8d1cc8a7675900358acd25. This will probably require some more async versions of APIs like `querySubstitutablePathInfos()`.

domenkozar:
@edolstra that would be great. Merging this will fix query performance for low-core counts until we get the async version done.

Issue #5320, opened 2021-10-02
`nix store gc` fails to delete self-referred path: executing SQLite statement 'delete from ValidPaths where path = '...';': constraint failed


trofi:
**Describe the bug**

My nix store is in state that does not allow removing seemingly unused stored path:

```
$ nix store gc
0 store paths deleted, 0.00 MiB freed
error: executing SQLite statement 'delete from ValidPaths where path = '/nix/store/2pklamh9d8x964wbawway71rks402bxr-gcc-11.1.0-lib';': constraint failed (in '/nix/var/nix/db/db.sqlite')
```

**`nix-env --version` output**

nix-env (Nix) 2.4pre20210922_bcd73eb

Attaching full vacuumed `db.sqlite`: 
[db.sqlite.gz](https://github.com/NixOS/nix/files/7271172/db.sqlite.gz)

trofi:
This system saw a few derivations built with `config.contentAddressedByDefault = true;`. I wonder if we lack a few deletes against `Realisations` table (or similar):

```
$ sqlite3 db.sqlite

sqlite> pragma foreign_keys=1;
sqlite> delete from ValidPaths where path = '/nix/store/2pklamh9d8x964wbawway71rks402bxr-gcc-11.1.0-lib';
Error: FOREIGN KEY constraint failed

sqlite> EXPLAIN delete from ValidPaths where path = '/nix/store/2pklamh9d8x964wbawway71rks402bxr-gcc-11.1.0-lib';

...
0     Init           0     1     0                    0   Start at 1; Start: .abort (AFTER DELETE ON ValidPaths)
...
0     Init           0     1     0                    0   Start at 1; Start: .abort (AFTER DELETE ON Realisations)
```

I noticed that `ValidPaths` has -predelete trigger to handle self-references:

```
CREATE TRIGGER DeleteSelfRefs before delete on ValidPaths
  begin
    delete from Refs where referrer = old.id and reference = old.id;
  end;
```

Should `RealisationsRefs` have something similar? Or my db is really corrupted? CC @regnat

trofi:
I tracked it down to a single `RealisationsRefs` self-reference:

```
INSERT INTO ValidPaths VALUES(880618,'/nix/store/2pklamh9d8x964wbawway71rks402bxr-gcc-11.1.0-lib',...);

INSERT INTO Refs VALUES(880618,879284);
INSERT INTO Refs VALUES(880618,880618); -- direct self-ref

[...]

PR #5329, opened 2021-10-04
nix: add (failing) selfreference test for multiple realizations

trofi:
The test illustrates failure in issue #5320. Here derivation and
it's built input have identical CA sotre path. As a result we generate
extraneout reference to build input.


trofi:
Was merged as 6ada4963111908211a4bbcc2ae65f4205cffaa18

PR #5335, opened 2021-10-05
Support arbitrary-length socket paths

edolstra:
Extracted from #5149.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -126,8 +126,7 @@
           ''
             mkdir -p $out/nix-support
 
-            # Converts /nix/store/50p3qk8kka9dl6wyq40vydq945k0j3kv-nix-2.4pre20201102_550e11f/bin/nix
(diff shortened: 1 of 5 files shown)


regnat (approved):
Oh my. That’s quite hacky, but also great!

(It’s a shame though that Linux doesn’t have `bindat` and `connectat` as that would have avoided this strange `createProcess` dance)

Any idea why the CI isn’t running on this PR?

edolstra:
Yeah I was wondering that too, but it did run successfully on my branch except for the usual macOS failure (https://github.com/edolstra/nix/actions/runs/1307887287).

Ericson2314:
> (It’s a shame though that Linux doesn’t have `bindat` and `connectat` as that would have avoided this strange `createProcess` dance)

http://lkml.iu.edu/hypermail/linux/kernel/1604.0/04290.html unfortunately, no response.

PR #5342, opened 2021-10-06
Add support for sourcehut input scheme

Misterio77:
Hello!

I was wondering if there's plans (or interest in patches) for supporting other "specialized" input types besides `github` and `gitlab`.

I currently use [sourcehut](https://sourcehut.org) for a couple projects ([some cool people of our community](https://sr.ht/~sumner/) also do), so i went ahead and tried implementing a input type for it :)

Currently, to refer to my nix config repo on sourcehut, i have to use `git+https://git.sr.ht/~misterio/nix-config?ref=main`, which is a lot harder to remember and type by heart than the shorthand version my patch introduces: `sourcehut:~misterio/nix-config`

Thanks a lot!

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -8,6 +8,7 @@
 
 #include <optional>
 #include <nlohmann/json.hpp>
+#include <fstream>
 
 namespace nix::fetchers {
 
@@ -17,7 +18,7 @@ struct DownloadUrl
     Headers headers;
 };
 
-// A github or gitlab host
+// A github, gitlab, or sourcehut host
 const static std::string hostRegexS = "[a-zA-Z0-9.]*"; // FIXME: check
 std::regex hostRegex(hostRegexS, std::regex::ECMAScript);
 
@@ -348,7 +349,95 @@ struct GitLabInputScheme : GitArchiveInputScheme
     }
 };
 
+struct SourceHutInputScheme : GitArchiveInputScheme
+{
+    std::string type() override { return "sourcehut"; }
+
+    std::optional<std::pair<std::string, std::string>> accessHeaderFromToken(const std::string & token) const override
+    {
+        // SourceHut supports both PAT and OAuth2. See
+        // https://man.sr.ht/meta.sr.ht/oauth.md
+        return std::pair<std::string, std::string>("Authorization", fmt("Bearer %s", token));
+        // Note: This currently serves no purpose, as this kind of authorization
+        // does not allow for downloading tarballs on sourcehut private repos.
+        // Once it is implemented, however, should work as expected.
+    }
+
+    Hash getRevFromRef(nix::ref<Store> store, const Input & input) const override
+    {
+        // TODO: In the future, when the sourcehut graphql API is implemented for mercurial
+        // and with anonymous access, this method should use it instead.
+
+        auto ref = *input.getRef();
+
+        auto host = maybeGetStrAttr(input.attrs, "host").value_or("git.sr.ht");
+        auto base_url = fmt("https://%s/%s/%s",
+            host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"));
+
+        Headers headers = makeHeadersWithAuthTokens(host);
+
+        std::string ref_uri;
+        if (ref == "HEAD") {
+            auto file = store->toRealPath(
+                downloadFile(store, fmt("%s/HEAD", base_url), "source", false, headers).storePath);
+            std::ifstream is(file);
+            std::string line;
+            getline(is, line);
+
+            auto ref_index = line.find("ref: ");
+            if (ref_index == std::string::npos) {
+                throw BadURL("in '%d', couldn't resolve HEAD ref '%d'", input.to_string(), ref);
+            }
+
+            ref_uri = line.substr(ref_index+5, line.length()-1);
+        } else
+            ref_uri = fmt("refs/heads/%s", ref);
+
+        auto file = store->toRealPath(
+            downloadFile(store, fmt("%s/info/refs", base_url), "source", false, headers).storePath);
+        std::ifstream is(file);
+
+        std::string line;
+        std::string id;
+        while(getline(is, line)) {
+            auto index = line.find(ref_uri);
+            if (index != std::string::npos) {
+                id = line.substr(0, index-1);
+                break;
+            }
+        }
+
+        if(id.empty())
+            throw BadURL("in '%d', couldn't find ref '%d'", input.to_string(), ref);
+
+        auto rev = Hash::parseAny(id, htSHA1);
+        debug("HEAD revision for '%s' is %s", fmt("%s/%s", base_url, ref), rev.gitRev());
+        return rev;
+    }
+
+    DownloadUrl getDownloadUrl(const Input & input) const override
+    {
+        auto host = maybeGetStrAttr(input.attrs, "host").value_or("git.sr.ht");
+        auto url = fmt("https://%s/%s/%s/archive/%s.tar.gz",
+            host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
+            input.getRev()->to_string(Base16, false));
+
+        Headers headers = makeHeadersWithAuthTokens(host);
+        return DownloadUrl { url, headers };
+    }
+
+    void clone(const Input & input, const Path & destDir) override
+    {
+        auto host = maybeGetStrAttr(input.attrs, "host").value_or("git.sr.ht");
+        Input::fromURL(fmt("git+https://%s/%s/%s",
+                host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo")))
+            .applyOverrides(input.getRef(), input.getRev())
+            .clone(destDir);
+    }
+};
+
 static auto rGitHubInputScheme = OnStartup([] { registerInputScheme(std::make_unique<GitHubInputScheme>()); });
 static auto rGitLabInputScheme = OnStartup([] { registerInputScheme(std::make_unique<GitLabInputScheme>()); });
+static auto rSourceHutInputScheme = OnStartup([] { registerInputScheme(std::make_unique<SourceHutInputScheme>()); });
 
 }
--- a/flake.nix
+++ b/flake.nix
@@ -501,6 +501,12 @@
           inherit (self) overlay;
         });
 
+        tests.sourcehutFlakes = (import ./tests/sourcehut-flakes.nix rec {
+          system = "x86_64-linux";
+          inherit nixpkgs;
+          inherit (self) overlay;
+        });
+
         tests.setuid = nixpkgs.lib.genAttrs
           ["i686-linux" "x86_64-linux"]
           (system:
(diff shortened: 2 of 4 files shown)


Misterio77:
I mean, as of now, we do already support custom gitlab instances by specifying the `host` attribute, right? It could work in a similiar way (this already does work with the patch).

But about mercurial, yeah, i was wondering if an attribute or different input type would make sense.

Thanks for the input!

Misterio77:
I think i might have got it working (a minimal version of the xml-related code works outside of nix), but i can't seem to get `tinyxml-2` to successfully link:
```
src/libfetchers/github.cc:383: error: undefined reference to 'tinyxml2::XMLDocument::XMLDocument(bool, tinyxml2::Whitespace)'
src/libfetchers/github.cc:383: error: undefined reference to 'tinyxml2::XMLDocument::XMLDocument(bool, tinyxml2::Whitespace)'
src/libfetchers/github.cc:384: error: undefined reference to 'tinyxml2::XMLDocument::LoadFile(char const*)'
src/libfetchers/github.cc:384: error: undefined reference to 'tinyxml2::XMLDocument::LoadFile(char const*)'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
/nix/store/77y1wj1k4bmcrs8zpxlc5c8aqbk6i572-tinyxml-2-6.0.0/include/tinyxml2.h:765: error: undefined reference to 'tinyxml2::XMLNode::FirstChildElement(char const*) const'
src/libfetchers/github.cc:389: error: undefined reference to 'tinyxml2::XMLNode::Value() const'
src/libfetchers/github.cc:389: error: undefined reference to 'tinyxml2::XMLNode::Value() const'
src/libfetchers/github.cc:383: error: undefined reference to 'tinyxml2::XMLDocument::~XMLDocument()'
src/libfetchers/github.cc:383: error: undefined reference to 'tinyxml2::XMLDocument::~XMLDocument()'
src/libfetchers/github.cc:383: error: undefined reference to 'tinyxml2::XMLDocument::~XMLDocument()'
src/libfetchers/github.cc:383: error: undefined reference to 'tinyxml2::XMLDocument::~XMLDocument()
```

Misterio77:
Fixed! Works perfectly now :)

puckipedia:
How does this work with `hg.sr.ht`? I suspect that calling it `sourcehut:` is a bad idea if it doesn't actually support all sourcehut services :p

src/libfetchers/github.cc
```diff
@@ -348,7 +349,90 @@ struct GitLabInputScheme : GitArchiveInputScheme
...
+        if (ref == "HEAD") {
+            auto file = store->toRealPath(
+                downloadFile(store, fmt("%s/HEAD", base_url), "source", false, headers).storePath);
+            std::ifstream is(file);
+            std::string line;
+            getline(is, line);
+            ref_uri = line.substr(line.find("ref: ")+5, line.length()-1);
+        } else {
+            ref_uri = fmt("refs/heads/%s", ref);
+        }
```
Mic92, line 389: Also not my personal preference but this is how the nix code base is formatted:

```suggestion
        } else
            ref_uri = fmt("refs/heads/%s", ref);
```

src/libfetchers/github.cc
```diff
@@ -348,7 +349,90 @@ struct GitLabInputScheme : GitArchiveInputScheme
...
+        while(getline(is, line)) {
+            auto index = line.find(ref_uri);
+            if (index != std::string::npos) {
+                id = line.substr(0, index-1);
+                break;
+            }
+        }
+        if(id.empty()) {
+            throw BadURL("in '%d', couldn't find ref '%d'", input.to_string(), ref);
+        }
```
Mic92, line 406:
```suggestion
        if(id.empty())
            throw BadURL("in '%d', couldn't find ref '%d'", input.to_string(), ref);
```

src/libfetchers/github.cc
```diff
@@ -348,7 +349,90 @@ struct GitLabInputScheme : GitArchiveInputScheme
...
+        Headers headers = makeHeadersWithAuthTokens(host);
+
+        std::string ref_uri;
+        if (ref == "HEAD") {
+            auto file = store->toRealPath(
+                downloadFile(store, fmt("%s/HEAD", base_url), "source", false, headers).storePath);
+            std::ifstream is(file);
+            std::string line;
+            getline(is, line);
+            ref_uri = line.substr(line.find("ref: ")+5, line.length()-1);
```
Mic92, line 386: Should this error if `line.find("ref: ")` returns `string::npos`?

Mic92 (commented):
Only some nitpicks. Documentation looks great.

src/libfetchers/github.cc
Misterio77: Hmmm. I think when the HEAD isn't found, the refs (thus the `id`) will also end up empty, throwing an error on line 405.

It makes sense, though, to throw it here too, just to be safe.

src/libfetchers/github.cc
Misterio77: Thank you

Misterio77:
Thanks for the review!

Mic92 (approved):
Tested with `nix-colors` on this repo:

```diff
diff --git a/flake.lock b/flake.lock
index 861af1c54..b66847ffd 100644
--- a/flake.lock
+++ b/flake.lock
@@ -16,24 +16,40 @@
         "type": "github"
       }
     },
+    "nix-colors": {
+      "locked": {
+        "lastModified": 1633318162,
+        "narHash": "sha256-rr6BMT4o66KWU9yXVv9W/yB/zMCVZC+zb4OvRftc/h8=",
+        "owner": "~misterio",
+        "repo": "nix-colors",
+        "rev": "182b4b8709b8ffe4e9774a4c5d6877bf6bb9a21c",
+        "type": "sourcehut"
+      },
+      "original": {
+        "owner": "~misterio",
+        "repo": "nix-colors",
+        "type": "sourcehut"
+      }
+    },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1632864508,
-        "narHash": "sha256-d127FIvGR41XbVRDPVvozUPQ/uRHbHwvfyKHwEt5xFM=",
+        "lastModified": 1638501222,
+        "narHash": "sha256-tgOZW/rLcB4P7iffx88K3iF70YH+44PtV6kVg5YGIg0=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "82891b5e2c2359d7e58d08849e4c89511ab94234",
+        "rev": "45c527ae2de6f2f36d5790273343f06cd14bc0d1",
         "type": "github"
       },
       "original": {
         "id": "nixpkgs",
-        "ref": "nixos-21.05-small",
+        "ref": "nixos-21.11-small",
         "type": "indirect"
       }
     },
     "root": {
       "inputs": {
         "lowdown-src": "lowdown-src",
+        "nix-colors": "nix-colors",
         "nixpkgs": "nixpkgs"
       }
     }
diff --git a/flake.nix b/flake.nix
index 942388ab4..41d146c7b 100644
--- a/flake.nix
+++ b/flake.nix
@@ -2,9 +2,11 @@
   description = "The purely functional package manager";
 
   inputs.nixpkgs.url = "nixpkgs/nixos-21.11-small";
+  inputs.nix-colors.url = "sourcehut:~misterio/nix-colors";
+
   inputs.lowdown-src = { url = "github:kristapsdz/lowdown"; flake = false; };
 
-  outputs = { self, nixpkgs, lowdown-src }:
+  outputs = { self, nixpkgs, lowdown-src, nix-colors }:
 
     let

```

Mic92:
I think this is good to merge @regnat

otavio (approved):

ratsclub (approved):

nyxra (approved):

thufschmitt:
I wonder whether having a custom input type for sourcehut is worth it. Github (and to a certain extent Gitlab) make sense because they are huge, but since sourcehut is much smaller, I have the impression than just using a git input for it would be simpler and avoid some potential maintenance headaches.

Misterio77:
I understand your point.

But I do wonder what the criteria for inclusion should be? Nixpkgs does have a fetcher for it (as well as for gitea, among others), and that too could require maintenance. I don't think the tarball URL for it is changing any time soon, so it probably would only need changes if all fetchers are (and they seem nicely abstracted? so I'm not sure).

Not having the shorthand, however, makes adoption much more difficult, I'd say.

thufschmitt (changes requested):
> I'm not sure if this can accurately represent the Nix community, but taking a look at [the NUR listing](https://github.com/nix-community/NUR/blob/master/repos.json) (192 repos total): 175 are on github, 8 on gitlab (2 of these self-hosted) and 5 on sourcehut.

Thanks, that’s a strong argument. I didn’t know how to measure this kind of things, but I think it’s the best metric we could reasonably get.

So I guess that’s a :+1: from me on the principle.

Implementation-wise, it looks all right, but it would be great to have an integration test for it. Something similar to `tests/github-flakes.nix` (that you can run with `nix build .#hydraJobs.tests.githubFlakes`), except that it could be simpler and more precise by spawning a real sourcehut server.

thufschmitt:
Oh too bad. There was some issues with upgrading nixpkgs which is why we’re still using an old version (though it should obviously be done at some point). A dummy server will be enough then

Mic92:
Here is the context why nix is still stuck on an old release: https://github.com/NixOS/nix/pull/5722

Misterio77:
I've added an integration test very similar to the github one: fake nixpkgs "repo" with tarball and ref resolver, registry file with just nixpkgs (but on sourcehut instead).
Please let me know if there's anything to be improved!

thufschmitt (approved):
Thanks, looks good now 👍

(Hopefully we’ll be able to replace the smoke server by a real one at some point)

NomisIV:
It would be nice to get something similar working for gitea (and also codeberg, which uses gitea). Maybe it would even be a good idea to write a general function to generate any input scheme provided a name and an API template link, so that other input schemes can be added trivially.

Issue #5306, opened 2021-09-29
musl build is broken


edolstra:
**Describe the bug**

buildStatic.x86_64-linux and buildStatic.aarch64-linux are broken, see https://hydra.nixos.org/build/151755012. Looks like this happened somewhere between af94b54 and 50a3586.

PR #5350, opened 2021-10-07
Revert "mk: prefert inplace library paths to system ones"

tomberek:
Reverts 4993174be59adc688efad49f2f9205688bd2ee26

@trofi can you refactor?

Fixes: https://github.com/NixOS/nix/issues/5306


trofi:
Ah, static build failure is unfortunate. Revert sounds good. Will have a closer look.

PR #5364, opened 2021-10-08
Make recursive-nix work even when not privileged

### LINKED ISSUE #5360: Recursive Nix is broken with `--store`
**Describe the bug**

[`addDependency`](https://github.com/NixOS/nix/blob/01e9f046a8f8fafb4d084153d4b30dd3a8d7aef5/src/libstore/build/local-derivation-goal.cc#L1435) uses `setns` which needs privileges to enter the mount namespace to add bind mounts. With `--store` when not run as root, this fails.

When using Recursive Nix with `--store` and root, it fails [here](https://github.com/NixOS/nix/blob/01e9f046a8f8fafb4d084153d4b30dd3a8d7aef5/src/libstore/build/local-derivation-goal.cc#L1614) instead for some yet unknown reason.

In practice, this means that any non-trivial derivation built inside the sandbox will fail as it causes `addDependency` to be called.

**Steps To Reproduce**

```nix

let
  hello-nix = pkgs.writeText "hello.nix" ''
    with import ${pkgs.path} {}; hello
  '';
in pkgs.runCommand "djE" {
  nativeBuildInputs = [ pkgs.nixUnstable ];
  requiredSystemFeatures = [ "recursive-nix" ];
} ''
  nix-build ${hello-nix}
  cp -L ./result $out
''
```

**Expected behavior**

It should work.

`nix (Nix) 2.4pre20210922_bcd73eb`

L-as:
Before this, `setns` would fail when switching to the mount namespace,
since we did not have the privileges to do so when not root.

Closes #5360

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -260,6 +260,7 @@ void LocalDerivationGoal::cleanupHookFinally()
 void LocalDerivationGoal::cleanupPreChildKill()
 {
     sandboxMountNamespace = -1;
+    sandboxUserNamespace = -1;
 }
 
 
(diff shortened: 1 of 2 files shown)


L-as:
Note: The `sudo` part of the issue isn't resolved, but it's a nicher use case.

L-as:
bump

Ericson2314:
@edolstra any questions with this one? I thought it was a pretty straight-forward fix.

L-as:
I think this patch is still needed.

edolstra:
Thanks!

PR #758, opened 2016-01-02
correct the nix-shell usage examples to prevent newcomer errors.

exi:
The misconception that configurePhase worked properly in nix-shell came up a few times. This is documented poorly right now.

PR #5365, opened 2021-10-09
Clarify that not all nix packages will use the default build phases

arafangion:
As suggested by @domenkozar in https://github.com/NixOS/nix/pull/758#issuecomment-928308931 I have attempted to provide a conservative pull request to make the change.

This is my first PR for nix, so I'm not entirely sure about conventions.  Happy to adjust and amend as required.

### DIFF
--- a/doc/manual/src/command-ref/nix-shell.md
+++ b/doc/manual/src/command-ref/nix-shell.md
@@ -110,18 +110,18 @@ shell in which to build it:
 
 ```console
 $ nix-shell '<nixpkgs>' -A pan
-[nix-shell]$ unpackPhase
+[nix-shell]$ eval ${unpackPhase:-unpackPhase}
 [nix-shell]$ cd pan-*
-[nix-shell]$ configurePhase
-[nix-shell]$ buildPhase
+[nix-shell]$ eval ${configurePhase:-configurePhase}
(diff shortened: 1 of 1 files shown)


arafangion:
@domenkozar - you mean, similar to how it was done in #758 ?  Why are there potentially different environment variables in that case, then?  In #758, they used $configurePhase-configurePhase but in my own example I tried locally, I had to use $configurePhase instead as the environment variable.

Also, I'm confused as to why we would have the regular configurePhase exposed, then - are we working around (what seems to be) a bug by basically telling people not to use the exposed commands, or are we working around specific derivations that just happen to not use those default commands?

I would have thought that using the commands (configurePhase), as an example, was how it was supposed to be done and that that these other nix files override it without changing that default was a (temporary?) aberration?

domenkozar:
`$configurePhase-configurePhase`is a typo, it should be `$configurePhase:-configurePhase`

I'm not sure why it needs to fall back, it's probably one of those bash things.

That needs to be looked into, but I don't have time at the moment.

Ericson2314:
I thought we needed the braces? `${configurePhase:-configurePhase}`

tomberek:
Thoughts on something like: https://github.com/flox/nix/commit/c03252cab70379214fabcd839cf432d447a13c42

PR #5380, opened 2021-10-13
Allow gc-ing with a rootless daemon

thufschmitt:
Depends on https://github.com/NixOS/nix/pull/5226

This PR adds the possibility to defer the collection of the gc roots to an external daemon (communicating with it through a UNIX socket).

The motivation for that is that (excluding https://github.com/NixOS/nix/pull/5226), finding the GC roots is the only part of Nix that strictly requires root. So deferring that part to an external tool means that it’s possible to run the Nix daemon as a non-root user (with the obvious security gains that come with it).

Along with it is included such an external tracing deamon (under `src/nix-find-roots`), written in such a way that it doesn’t depend on anything but the c++ stl.

https://github.com/regnat/rootless-nix-daemon-test provides a sample setup in which the nix-daemon is running as a normal `nix-daemon` user.

Using this feature is currently guarded behind the `external-gc-daemon` experimental flag

Fix https://github.com/NixOS/nix/issues/5208

### DIFF
--- a/src/libutil/experimental-features.cc
+++ b/src/libutil/experimental-features.cc
@@ -275,8 +275,8 @@ constexpr std::array<ExperimentalFeatureDetails, numXpFeatures> xpFeatureDetails
             Make the garbage collector use an external daemon for the tracing.
 
             This makes it possible to run a multi-user Nix daemon as a non-root
-            user (only the tracing daemon needs to be root), reducing the attack
-            surface a lot.
+            user. Only the tracing daemon needs to be root. This reduces the attack
+            surface.
 
             This requires more infrastructure and isn't directly supported by the
             installer.


fricklerhandwerk:
Discussed in Nix team meeting on 2022-12-19:

  - @edolstra: this is only useful outside of NixOS, because if you own the store, you can mess with the entire system
      - @Ericson2314: this could be fixed in principle
          - @edolstra: yes, if we had a restricted daemon that only takes content-addressed

- @Ericson2314 *post meeting* created an [issue to discuss this](https://github.com/NixOS/nix/issues/7471).

tests/gc-external-daemon.sh
```diff
@@ -0,0 +1,39 @@
...
+    wait $pidGcDaemon || true
+    trapFunctions[killGcDaemon]=:
+}
+
+startGcDaemon
+
+source ./gc.sh
+source ./gc-concurrent.sh
+source ./gc-runtime.sh
+source ./gc-auto.sh
```
Ericson2314, line 39: For the sake of parallelism, can we have for each `gc-$test.sh`, and `gc-external-daemon/gc-$test.sh` that is the external daemon version of just that one test? (Basically like we do for CA derivations.)

src/libstore/globals.hh
```diff
@@ -122,6 +122,13 @@ public:
           for supported store types and settings.
         )"};
 
+    Setting<std::string> gcSocketPath {
+        this,
+        getEnv("NIX_GC_SOCKET_PATH").value_or("auto"),
```
Ericson2314, line 127: Now that we have `NIX_CONFIG`, are new `NIX_` env vars needed? I am not so sure.

tests/gc-external-daemon.sh
thufschmitt: Oh indeed!

src/libstore/globals.hh
thufschmitt: Fair point, it's not really needed indedd. I'll drop it

tests/functional/gc-external-daemon/common.sh
```diff
@@ -0,0 +1,6 @@
+source ../common.sh
+
+enableFeatures "external-gc-daemon"
+echo "gc-socket-path = $NIX_GC_SOCKET_PATH" >> "$NIX_CONF_DIR"/nix.cong
```
Ericson2314, line 4:
```suggestion
echo "gc-socket-path = $NIX_GC_SOCKET_PATH" >> "$NIX_CONF_DIR"/nix.conf
```

Ericson2314:
The macOS failure is very odd; if connecting to the socket (including `chdir`) fails, then the failure should be caught and in-process root-finding done instead.

thufschmitt:
> The macOS failure is very odd; if connecting to the socket (including `chdir`) fails, then the failure should be caught and in-process root-finding done instead.

The `gc-socket` here is the one of the non-blocking daemon thingy. So it's “just” the usual CI flakiness at play AFAIU :unamused:

thufschmitt:
The timeouts on Ubuntu are actually more worrisome :smile:

Ericson2314:
I saw both types of flakiness on other PRs. Hopefully it works now. If so, do you think this ready to merge @thufschmitt? I would really like to see it merged!

thufschmitt:
> do you think this ready to merge @thufschmitt?

I would still like @edolstra to confirm that the tracing algorithm is correct if he can find time, but apart from that, yes it is!

thufschmitt:
@NixOS/nix-team I'd like to get this merged after the next Nix release. @Ericson2314 already gave it a thorough pass, and it has been running in production for more than two years. Any objection?

Ericson2314:
I think with @edolstra's recent fix to our `connect` wrapper, this will finally pass.

thufschmitt:
Just rebased on top of master. Hopefully that'll fix the CI failure

Ericson2314:
I think the problem might be that the test instrumentation that @edolstra recently added is missing in the new version.

brainrake:
Looks like the remaining test failures are caused by 1h timeout.

Ericson2314:
@brainrake I think that is because something is hanging. It should finish in well under an hour.

thufschmitt:
I didn't _really_ understand why it was timing out, but I managed (I hope) to work around it any way :sweat_smile:

thufschmitt:
Discussed during the Nix maintainers meeting on 2024-01-26.

Good to merge after the 2.20 release.

Would be nice to have @edolstra take a look first if he has time, but not a blocker.

src/nix-find-roots/main.cc
```diff
@@ -0,0 +1,188 @@
...
+            currentPos += newStr.length();
+        }
+    }
+
+    return original;
+}
+
+#define SD_LISTEN_FDS_START 3 // Like in systemd
+
+int main(int argc, char * * argv)
```
edolstra, line 98: I don't think we should add another top-level command. This should be a subcommand of `nix`, similar to `build-remote`.

src/nix-find-roots/main.cc
```diff
@@ -0,0 +1,188 @@
...
+  std::cerr << msg << std::endl;
+}
+
+TracerConfig parseCmdLine(int argc, char** argv)
+{
+    std::function<void(std::string_view msg)> log = logStderr;
+    std::function<void(std::string_view msg)> debug = logNone;
+    fs::path storeDir = "/nix/store";
+    fs::path stateDir = "/nix/var/nix";
+    fs::path socketPath = "/nix/var/nix/gc-trace-socket/socket";
```
edolstra, line 23: This should use the regular `settings` from `globals.hh`.

src/nix-find-roots/lib/find-roots.cc
```diff
@@ -0,0 +1,233 @@
...
+#include <algorithm>
+#include <fstream>
+#include <optional>
+
+#include "find-roots.hh"
+
+
+namespace nix::roots_tracer {
+namespace fs = std::filesystem;
+using std::set, std::string;
```
edolstra, line 24:
```suggestion
```
Just use `std::string` etc. We got rid of the `string` alias a while ago (df552ff53e68dff8ca360adbdbea214ece1d08ee).

src/libstore/gc.cc
```diff
@@ -241,82 +243,35 @@ void LocalStore::findTempRoots(Roots & tempRoots, bool censor)
...
-        /* We only ignore permanent failures. */
-        if (e.errNo == EACCES || e.errNo == ENOENT || e.errNo == ENOTDIR)
-            printInfo("cannot read potential root '%1%'", path);
-        else
-            throw;
+    const TracerConfig opts {
+      .storeDir = fs::path(storeDir),
+      .stateDir = fs::path(stateDir.get())
+    };
+    const std::set<fs::path> standardRoots = {
```
edolstra, line 256: We don't use `fs::path` elsewhere, so for consistency it's better to use `Path` or `CanonPath`.

flake.nix
```diff
@@ -152,6 +152,30 @@
             '';
           };
 
+          nix-find-roots = prev.stdenv.mkDerivation {
```
edolstra, line 155: Does this need to be a separate package? That seems unnecessarily complicated.

src/nix-find-roots/main.cc
```diff
@@ -0,0 +1,188 @@
...
+
+        if (remoteSocket == -1) {
+            if (errno == EINTR) continue;
+            throw Error("Error accepting the connection");
+        }
+
+        opts.log("accepted connection");
+
+        auto printToSocket = [&](std::string_view s) {
+            send(remoteSocket, s.data(), s.size(), 0);
```
edolstra, line 165: Use `FdSink` which provides buffering. All these `send`s are inefficient.

src/nix-find-roots/main.cc
```diff
@@ -0,0 +1,188 @@
...
+                (struct sockaddr*) & remoteAddr,
+                &remoteAddrLen
+                );
+
+        if (remoteSocket == -1) {
+            if (errno == EINTR) continue;
+            throw Error("Error accepting the connection");
+        }
+
+        opts.log("accepted connection");
```
edolstra, line 162:
```suggestion
        printInfo("accepted connection");
```

edolstra (changes requested):

src/nix-find-roots/main.cc
```diff
@@ -0,0 +1,188 @@
...
+        { "store_dir", required_argument, 0, 's' },
+        { "state_dir", required_argument, 0, 'd' },
+        { "help", no_argument, 0, 'h' },
+        { 0, 0, 0, 0 },
+    };
+
+    int option_index = 0;
+    int opt_char;
+    while((opt_char = getopt_long(argc, argv, "vd:s:l:h",
+                    long_options, &option_index)) != -1) {
```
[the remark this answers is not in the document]
thufschmitt: Nah. The whole point for this thing is that

> [it is] written in such a way that it doesn’t depend on anything but the c++ stl.

because there's not much point in having a separate daemon if it depends on the same code as the rest.

This is also the reason for most of the oddities you mention below.

src/libutil/experimental-features.cc
```diff
@@ -268,6 +268,20 @@ constexpr std::array<ExperimentalFeatureDetails, numXpFeatures> xpFeatureDetails
...
+        .description = R"(
+            Make the garbage collector use an external daemon for the tracing.
+
+            This makes it possible to run a multi-user Nix daemon as a non-root
+            user (only the tracing daemon needs to be root), reducing the attack
+            surface a lot.
+
+            This requires more infrastructure and isn't directly supported by the
+            installer.
+        )"
```
roberth, line 283: Ok, but then how does anyone set this up?

src/libutil/experimental-features.cc
```diff
@@ -268,6 +268,20 @@ constexpr std::array<ExperimentalFeatureDetails, numXpFeatures> xpFeatureDetails
...
     },
+    {
+        .tag = Xp::ExternalGCDaemon,
+        .name = "external-gc-daemon",
+        .description = R"(
+            Make the garbage collector use an external daemon for the tracing.
+
+            This makes it possible to run a multi-user Nix daemon as a non-root
+            user (only the tracing daemon needs to be root), reducing the attack
+            surface a lot.
```
roberth, line 279:
```suggestion
            This makes it possible to run a multi-user Nix daemon as a non-root
            user. Only the tracing daemon needs to be root. This reduces the attack
            surface.
```

No need for subjective statements or parentheses.

roberth (changes requested):
Users can't use this without docs.

src/nix-find-roots/lib/find-roots.cc
```diff
@@ -0,0 +1,233 @@
...
+
+static string quoteRegexChars(const string & raw)
+{
+    static auto specialRegex = std::regex(R"([.^$\\*+?()\[\]{}|])");
+    return std::regex_replace(raw, specialRegex, R"(\$&)");
+}
+static std::regex storePathRegex(const fs::path storeDir)
+{
+    return std::regex(quoteRegexChars(storeDir) + R"(/[0-9a-z]+[0-9a-zA-Z\+\-\._\?=]*)");
```
roberth, line 33:
```suggestion
    return std::regex(quoteRegexChars(storeDir) + R"(/[0-9a-z]+-(?!\.\.?(-|$))[0-9a-zA-Z\+\-\._\?=]+)");
```

See https://github.com/NixOS/nix/pull/9867/files#diff-bdf530c7e47ccaf7f924b90d1f61c3aa7866c63a3ce5ebdcbe07960284dca223R11

Ericson2314:
Are we just waiting for a final review after re-agreeing on the "don't use other libraries approach"? We did say we wanted to do this after releasing 2.20.

doc/manual/src/installation/installing-rootless-daemon.md
```diff
@@ -0,0 +1,138 @@
...
+
+    [Socket]
+    ListenStream=/nix/var/nix/daemon-socket/socket
+    SocketUser=nix-daemon
+
+    [Install]
+    WantedBy=sockets.target
+    EOF
+    ```
+7. Install the systemd services for the tracing daemon:
```
roberth, line 82: Is it a "gc roots" or "tracing" daemon?
We should use _one_ name everywhere.

tests/nixos/rootless-daemon.nix
```diff
@@ -0,0 +1,177 @@
...
+    users.users.nix-daemon.isNormalUser = true;
+    users.users.nix-daemon.group = "nix-daemon";
+    users.groups.nix-daemon = {};
+    environment.variables.NIX_REMOTE = "daemon"; # Even for root
+    virtualisation.writableStore = true;
+
+    boot.readOnlyNixStore = false;
+
+    # No root daemon
+    nix.enable = false;
```
roberth, line 18: non-blocking:

Exposing a module for this kind of setup would be nice, but needs a few improvements.

I think you could blacklist the daemon module, at least on recent enough NixOS the daemon-specific stuff is separate from the Nix installation and config file.

I think we need a `nix.daemon.enable` in NixOS to be sure.

roberth (commented):
Have reviewed the docs.

twoolie:
Is there anything still blocking progress on this PR?

GHSA-h4vv-h3jq-v493 has shown that getting to a rootless daemon is still important.

tomberek:
> Is there anything still blocking progress on this PR?
> 
> [GHSA-h4vv-h3jq-v493](https://github.com/NixOS/nix/security/advisories/GHSA-h4vv-h3jq-v493) has shown that getting to a rootless daemon is still important.

Needs a rebase, some attention, and someone to shepherd it forward. I'm not aware of any fundamental blockers.

tests/nixos/rootless-daemon.nix
```diff
@@ -0,0 +1,177 @@
...
+    environment.etc."nix/nix.conf".source = pkgs.writeTextFile {
+      name = "nix.conf";
+      text = ''
+        experimental-features = external-gc-daemon
+      '';
+    };
+
+    systemd.user.sockets.nix-daemon = {
+    };
+    systemd.user.services.nix-daemon = {
```
artemist, line 58: You probably don't want user services, those will only start on a user's first login session. This should be a system service running as the user `nix-daemon`

Ericson2314:
Closed in favor of https://github.com/NixOS/nix/pull/15143. The rest of the Nix team (I recused myself) decided they rather first land support using `libstore` etc. and not re-implementing from scratch. Such hardening can come liter if/when a security audit thinks it is prudent to do so.

PR #5384, opened 2021-10-14
preloadNSS / dns timeout

baloo:
@tomberek was reporting dns timeout when running commands like `nix --version`.

Reading glibc again, I stumbled upon `__nss_configure_lookup` which is actually exactly what we're trying to achieve.

I think this simplifies the workaround a lot as it does not rely on a side-effect anymore, and is pretty straight-forward to understand.


tomberek:
Builds on x86_64-linux. Limited testing of both nix and nix-static. While the method utilized is simpler and more appropriate, it does have the GNU-specific requirement. Would be good to explain/document the various nix/nix-static/darwin/etc implications and test them.

baloo:
@tomberek I pushed a comment, let me know what you think or if there is anything missing.

PR #5385, opened 2021-10-14
Add `dirtyRev` and `dirtyShortRev` to `fetchGit`

### LINKED ISSUE #4682: Should a flake's `self.rev` return something useful in modified repos?
What's the thinking on why a flake's `self.rev` won't evaluate in a dirty repo? IMO, it would be a lot nicer to work with `self.rev` in Nix expressions if it returned `<rev>-dirty` (or similar) in this situation.

I do understand the argument that it's better to fail in this situation than to return `<rev>-dirty`, but then it seems a bit inconsistent that you *can* get a modified repo's last-modified via `self.lastModified` or `self.lastModifiedDate`.

Enzime:
This also adds it to `nix flake metadata --json` as `.dirtyRevision` and when running `nix flake metadata` it'll output the dirty revision as the replacement for `Revision: `

Fixes #4682

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -22,7 +22,7 @@ void emitTreeAttrs(
 {
     assert(input.isLocked());
 
-    auto attrs = state.buildBindings(8);
+    auto attrs = state.buildBindings(10);
 
 
     state.mkStorePathString(tree.storePath, attrs.alloc(state.sOutPath));
@@ -54,6 +54,11 @@ void emitTreeAttrs(
         else if (emptyRevFallback)
             attrs.alloc("revCount").mkInt(0);
 
+    }
+
+    if (auto dirtyRev = fetchers::maybeGetStrAttr(input.attrs, "dirtyRev")) {
+        attrs.alloc("dirtyRev").mkString(*dirtyRev);
+        attrs.alloc("dirtyShortRev").mkString(*fetchers::maybeGetStrAttr(input.attrs, "dirtyShortRev"));
     }
 
     if (auto lastModified = input.getLastModified()) {
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -243,6 +243,13 @@ std::pair<StorePath, Input> fetchFromWorkdir(ref<Store> store, Input & input, co
         "lastModified",
         workdirInfo.hasHead ? std::stoull(runProgram("git", true, { "-C", actualPath, "--git-dir", gitDir, "log", "-1", "--format=%ct", "--no-show-signature", "HEAD" })) : 0);
 
+    if (workdirInfo.hasHead) {
+        input.attrs.insert_or_assign("dirtyRev", chomp(
+            runProgram("git", true, { "-C", actualPath, "--git-dir", gitDir, "rev-parse", "--verify", "HEAD" })) + "-dirty");
+        input.attrs.insert_or_assign("dirtyShortRev", chomp(
+            runProgram("git", true, { "-C", actualPath, "--git-dir", gitDir, "rev-parse", "--verify", "--short", "HEAD" })) + "-dirty");
+    }
+
     return {std::move(storePath), input};
 }
 }  // end namespace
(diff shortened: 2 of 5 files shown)


maydayv7:
(Bumping...)

Enzime:
@maydayv7 if you're interested in using this PR, take a look at the way I'm using it from my dotfiles:

I use an overlay to patch Nix in `nixpkgs`: https://github.com/Enzime/dotfiles-nix/blob/7b0b6befe0bf864c83ec03c913fc06128942483d/overlays/nix/flake.nix#L19-L25

And here is where I actually use `dirtyRev`: https://github.com/Enzime/dotfiles-nix/blob/7b0b6befe0bf864c83ec03c913fc06128942483d/flake.nix#L51-L54

maydayv7:
Thanks a lot!

lilyball:
I really want this.

@Enzime Just as a tip, your second example of where you use `dirtyRev` can be simplified like `full = self.rev or self.dirtyRev or "dirty-inputs";`.

lilyball:
It looks like this always picks 7 digits for the short revision? Git defaults to 7 but will increase the length depending on the number of objects in the repository to avoid collisions in large repos. If we don't want to run a second `git describe` command to let git pick the abbreviation, then we should probably pick a longer default (I tend to go for 9 when I'm shortening revs, but I don't know what logic git uses to pick this number).

For reference, in current github:NixOS/nix git picks 9 instead of 7 for abbreviations, and in github:NixOS/nixpkgs it picks 11. So it's probably best to actually let git go ahead and run its logic here.

src/libfetchers/git.cc
```diff
@@ -244,6 +244,12 @@ std::pair<StorePath, Input> fetchFromWorkdir(ref<Store> store, Input & input, co
         "lastModified",
         workdirInfo.hasHead ? std::stoull(runProgram("git", true, { "-C", actualPath, "--git-dir", gitDir, "log", "-1", "--format=%ct", "--no-show-signature", "HEAD" })) : 0);
 
+    // QUESTION: should this just return an empty hash if there are no commits?
+    if (workdirInfo.hasHead) {
+        input.attrs.insert_or_assign("dirtyRev", chomp(
+            runProgram("git", true, { "-C", actualPath, "--git-dir", gitDir, "describe", "--always", "--match=''", "--abbrev=0", "--dirty" })));
```
lilyball, line 250: This command is surprisingly slow in github:NixOS/nixpkgs, I'm really not sure why given the `--match=''` means it always ends up outputting the full sha, and it's not the `--dirty` flag that makes it slow (it's slow without that too).

I'm inclined to say this should instead run `git rev-parse --verify HEAD` for the sha, and then some other command to compute dirty status (perhaps `git diff --quiet HEAD`).

I'm also not sure how submodules affect this. AIUI current behavior is to ignore submodules unless the flake is specified with `?submodules=1`, but I don't know if that affects the current revision calculation or if that propagates to this command somehow.

Enzime: I've chosen to switch this to use `git rev-parse --verify HEAD` as suggested :+1: 

As this code only runs if the worktree is dirty so we don't need to check that it's dirty and we can manually append "-dirty" unconditionally

Enzime:
@lilyball I've also fixed the short revisions not matching Git by adding an extra call to `git rev-parse` with `--short` :+1:

Enzime:
@edolstra @thufschmitt any chance I could get a review on this PR? the linked issue #4682 shows there is interest in being able to access the revision even in dirty worktrees

fricklerhandwerk:
Triaged in the Nix team meeting (notes for maintainers):

- @edolstra: have to be careful about adding metadata, potentially expensive to maintain
    - probably not a problem here
    - @thufschmitt: is it an issue with `libgit`?
    - @edolstra: no, but should check to make sure

src/libexpr/primops/fetchTree.cc
```diff
@@ -56,6 +56,13 @@ void emitTreeAttrs(
 
     }
 
+    // `dirtyRev` does not exist only when there are no commits
+    // OR when `builtins.getFlake` is called
```
thufschmitt, line 60: I'm not sure what that comment means: `dirtyRev` should only exist when fetching a dirty local repo, right?

thufschmitt (approved):
Sorry for the delay. This looks good to me, except for one comment which seems a bit off.

This is missing documentation, but none of this is documented any way, so it will have to wait for https://github.com/NixOS/nix/pull/8509/ (or equivalent) to land first.

src/libexpr/primops/fetchTree.cc
Enzime: Yeah, I originally added this comment just to help me when writing this PR, I'm happy to remove it if you think it'll be less confusing

thufschmitt: Yeah, I think it would be better to drop it. Just ping me once it's done and I can merge

Enzime: @thufschmitt updated :+1:

thufschmitt: Thanks :)

AleXoundOS:
Is it expected that neither `dirtyRev` nor `rev` exist when `--override-input` is used with `nix build`?

AleXoundOS:
@Enzime, was the original intention to ignore the `--override-input` case in this PR? And how do you think it should be handled?

con-f-use:
Hmm... `self.rev or self.dirtyRev or "dirty"` now gives me a mysterious `Failed assertions: system.copySystemConfiguration is not supported with flakes`

PR #5388, opened 2021-10-15
Add another path where a Rosetta 2 daemon plist file is possibly located

yvt:
This PR fixes Rosetta 2 detection on macOS 11.5.0 and later.

<https://github.com/NixOS/nix/pull/4310> updated `Settings::getDefaultExtraPlatforms` to detect the presence of Rosetta 2 and include `x86_64-darwin` automatically when eligible. It does so by checking the existence of a file at `/Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist`. However, [as of macOS 11.5.0](https://github.com/rtrouton/rtrouton_scripts/issues/87), this detection method stopped working because the file is now located in a different path `/System/Library/LaunchDaemons/com.apple.oahd.plist`. This PR introduces a check for this new path to fix this issue.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -148,7 +148,8 @@ StringSet Settings::getDefaultExtraPlatforms()
     // machines. Note that we can’t force processes from executing
     // x86_64 in aarch64 environments or vice versa since they can
     // always exec with their own binary preferences.
-    if (pathExists("/Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist")) {
+    if (pathExists("/Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist") ||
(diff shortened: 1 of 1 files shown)


roberth:
This change has added `x86_64-darwin` to my machine, but when I build an `x86_64-darwin` derivation, it fails without an error message (although the lack of a message may be due to a low level exception handling bug that I'm fixing). When I invoke an intel executable manually, I get a "bad file descriptor" error.
I did not enable Rosetta on the machine; I've only installed nix and then nix-darwin on it.

I use an M1 with

```
> sw_vers                                                                                                                                                                                                                                                                            ~
ProductName:	macOS
ProductVersion:	11.6
BuildVersion:	20G165
```

it has the a file in the "new" location but it does not have an `oahd` process.

Issue #2646, opened 2019-01-22
Regression: error: restoring parent mount namespace: Invalid argument


mpickering:
There is a regression in `2.2` due to the following two commits.

a32ff2573ba4d0df4e3360c5a96398738da953f6
a0ef21262f4d5652bfb65cfacaec01d89c475a93

Reverting these two commits causes the error not to happen.

It can be reproduced by running this script.

https://github.com/mpickering/ghc-artefact-nix/blob/master/ghc-head-from.sh

```
wget https://github.com/mpickering/ghc-artefact-nix/blob/master/ghc-head-from.sh
nix-shell -p jq
./ghc-head-from.sh
```

mpickering:
@tilpner says this can be reproduced with `nix-shell -p hello` rather than my complicated instructions.

edolstra:
On what system?

mpickering:
@edolstra We are both on NixOS.

tilpner:
@iblech Are you sure it requires 2.? Mounting/unmounting is not necessary here.

iblech:
You are right! Condition 2 is not necessary. Reliably, the following happens:

* A first invocation of `nix-shell -p foo`, for any value of `foo` I've tested, successfully fetches the required derivations and then aborts with `error: restoring parent mount namespace: Invalid argument`.
* A second invocation immediately following the first one works just fine.

This is on Kernel 4.20.6 on a virtual private server.

volth:
```
nix-shell --run git -p git \
  --option binary-caches http://cache.nixos.org \
  -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/18.09.tar.gz
```
as root
100% reproducible

iblech:
Can confirm, @volth's error-inducing approach is 100% reproducible here.

7c6f434c:
With prefetched `git` from all versions (no builds necessary):
* `NIX_REMOTE=daemon` of course helps
* `-I nixpkgs=` to an `https://` URL fails as reported
* the Nixpkgs tarball being cached doesn't change anything
*  `file:///` URL also fails
* local directory as Nixpkgs path works fine
* local tarball as a file gets interpreted as a Nix file and doesn't provide a data point

edolstra:
This is caused by
> EINVAL The caller is multithreaded and tried to join a new user namespace.
We need to get rid of all other threads before restoring the mount namespace.

arianvp:
[...]

PR #5391, opened 2021-10-15
Restore parent mount namespace in restoreProcessContext

### LINKED ISSUE #4337: /nix/store is writable in a nix-shell started by root

yorickvP:
This ensures any started processes can't write to /nix/store (except
during builds). This partially reverts 01d07b1e, which happened because
of #2646.

The problem was only happening after nix downloads anything, causing
me to suspect the download thread. The problem turns out to be:
"A  process  can't  join a new mount namespace if it is sharing
filesystem-related attributes with another process", in this case this
process is the curl thread.

Ideally, we might kill it before spawning the shell process, but it's
inside a static variable in the getFileTransfer() function. So
instead, stop it from sharing FS state using unshare(). A strategy
such as the one from #5057 (single-threaded chroot helper binary) is
also very much on the table.

Fixes #4337.

Thanks @puckipedia for helping with debugging and ideas.

### DIFF
--- a/src/libstore/filetransfer.cc
+++ b/src/libstore/filetransfer.cc
@@ -544,6 +544,14 @@ struct curlFileTransfer : public FileTransfer
             stopWorkerThread();
         });
 
+#ifdef __linux__
+        /* Cause this thread to not share any FS attributes with the main thread,
+           because this causes setns() in restoreMountNamespace() to fail.
+           Ideally, this would happen in the std::thread() constructor. */
+        if (unshare(CLONE_FS) != 0)
+            throw SysError("unsharing filesystem state in download thread");
+#endif
+
         std::map<CURL *, std::shared_ptr<TransferItem>> items;
 
         bool quit = false;
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -495,6 +495,7 @@ void LocalStore::makeStoreWritable()
         throw SysError("getting info about the Nix store mount point");
 
     if (stat.f_flag & ST_RDONLY) {
+        saveMountNamespace();
         if (unshare(CLONE_NEWNS) == -1)
             throw SysError("setting up a private mount namespace");
 
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -1629,10 +1629,34 @@ void setStackSize(size_t stackSize)
     }
     #endif
 }
+static AutoCloseFD fdSavedMountNamespace;
 
-void restoreProcessContext()
+void saveMountNamespace()
+{
+#if __linux__
+    static std::once_flag done;
+    std::call_once(done, []() {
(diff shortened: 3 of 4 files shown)


yorickvP:
Todo: check if boehm threads interfere

edolstra:
I figured out why the Boehm GC threads are no longer a problem: we call `initGC()` early in `main()` now, before `LocalStore::makeStoreWritable()` calls `unshare(CLONE_NEWNS)`. So the Boehm GC threads don't share the FS state of the main thread.

Gerschtli:
Hey, since the release of version 2,5, I get your introduced error message in my CI. Maybe my CI setup is a bit too hacky and does things you are trying to avoid, but maybe someone could give me a hint/pointer what could be causing the issue.

So my CI (https://github.com/Gerschtli/nix-config/blob/master/.github/workflows/ci.yml) installs nix and cachix and for aarch64-linux stuff creates an emulated aarch64 container where I am then also installing nix and cachix but now as aarch64 version. The outer installation is just for pushing of dependencies to cachix.

I get the error In the "Setup cachix" group in the aarch64 container.

```
Setup cachix
  unexpected error in download thread: error: unsharing filesystem state in download thread: Operation not permitted
  error: download of 'https://cachix.org/api/v1/install' was interrupted
  Error: The process '/home/runner/work/_actions/Gerschtli/run-on-arch-action/groups/src/run-on-arch.sh' failed with exit code 1
```

AFAIK to trigger this error there need to be a process running that modifies the nix store while another nix process starts which does not seem to be the case for me..

Any further information would be appreciated :)

tristanpemble:
@Gerschtli I filed an issue with the Docker image repository that might be related to your root cause: https://github.com/NixOS/docker/issues/52

PR #5394, opened 2021-10-15
Add backport action

domenkozar:
This follows what we've been using in nixpkgs for a few months and will help with automation around backporting changes to older branches.


abathur:
Seems like this would be nice :)

Should it track some of the changes in https://github.com/NixOS/nixpkgs/commit/6954a396d11feaa1f543fccd3b994ed3db258d26?

Issue #5238, opened 2021-09-12
Installer assumes no TTY when a TTY is present


winterqt:
**Describe the bug**

When running the installer with `curl -L https://nixos.org/nix/install | sh -s -- --darwin-use-unencrypted-nix-store-volume --daemon` as recommended in the manual, the script assumes that no TTY is present.

**Steps To Reproduce**

1. Run `curl -L https://nixos.org/nix/install | sh -s -- --darwin-use-unencrypted-nix-store-volume --daemon` in iTerm 2 (not sure if the terminal emulator used is important)

**Expected behavior**

For the script to detect a TTY successfully

**`nix-env --version` output**

n/a

**Additional context**

n/a

abathur:
AFAIK this is expected behavior with the pipe syntax while `sh <(curl...)` would do what you expect?

winterqt:
That's what I was thinking, yeah.

The change to the manual was introduced in #4985, I guess it wasn't thought through completely

Not sure how we'd want to proceed, though. Just revert it, or is there another shell-universal way that lets us still have access to the TTY?

winterqt:
@edolstra If I were to make a PR reverting #4985, what branch should I base it off? `2.3-maintenance` like the original PR?

tomberek:
Closing as stale. Re-open if needed.

PR #5395, opened 2021-10-15
Revert "doc: Change install syntax to be fish compatible"

winterqt:
This reverts commit ea962a84c35cb260e5feee5c5ba114f0db171055, due to #5238. In my opinion, this is the best solution until a cross shell one-liner is found that doesn't make `sh` headless.

### DIFF
--- a/doc/manual/installation/installing-binary.xml
+++ b/doc/manual/installation/installing-binary.xml
@@ -12,7 +12,7 @@
 </para>
 
 <screen>
-  $ curl -L https://nixos.org/nix/install | sh -s
+  $ sh &lt;(curl -L https://nixos.org/nix/install)
 </screen>
 
 <para>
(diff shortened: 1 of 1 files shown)


winterqt:
@edolstra Is `2.3-maintenance` the correct branch to merge this into, or should I switch it to `2.4-maintenance`?

edolstra:
It should go to master first, and then I'll cherry-pick it to the 2.4 branch.

winterqt:
> It should go to master first, and then I'll cherry-pick it to the 2.4 branch.

This file doesn't exist on master though, since master is using mdbook.

winterqt:
I'm confused: so what do I change? The file does not exist on master.

edolstra:
Sorry for the confusion. The source of the confusion is that #4985 was applied to 2.3-maintenance without being a backport, which was the wrong thing to do.

PR #5400, opened 2021-10-17
nix/flake: validate `builders` from `flake` output

Ma27:
While implementing a flake exposing `builders` — such as it's done in
`import-cargo`[1] — I realized that `nix flake check` doesn't seem to
allow such an output:

    $ nix flake check
    warning: unknown flake output 'builders'

I actually think that this is a valid field and added an actual validator
which expects lambdas being exposed inside.

Also fixed a minor typo in the `bundler`-check.

[1] https://github.com/edolstra/import-cargo

cc @edolstra

### DIFF
--- a/src/nix/flake-check.md
+++ b/src/nix/flake-check.md
@@ -54,6 +54,11 @@ The following flake output attributes must be *Nixpkgs overlays*:
 * `overlay`
 * `overlays`.`*name*
 
+The following flake output attributes must be functions for building
+derivations:
+
+* `builders`.`*name*
+
 The following flake output attributes must be *NixOS modules*:
 
 * `nixosModule`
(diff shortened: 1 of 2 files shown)


Ma27:
cc @edolstra anything else missing to get this merged? :)

tomberek:
I’m still it clear on what the proposed semantics are. What is a “builder”?

builder.<name>=   { lockFile, pkgs }: derivation; ?

edolstra:
I'm a bit reluctant to formalize the `builders` output since it's not really clear that it's useful to standardize it. Currently, the outputs that are checked by `nix flake check` are basically those that are consumed by tools (like `nix` or Hydra or `nixos-rebuild`). But that's not the case for `builders`, and it's not even clear what the semantics of `builders` is (e.g. you could also include them in `lib`).

Ma27:
Hmm, I guess I see your point now :) 
What is your opinion about formalizing `lib` then? I think it'd be quite nice if there's a standardized way of exposing functions via flakes :)

Ma27:
Closing. Personally, I think that the flake-format should be well-defined via an RFC with actual community involvement rather than having a few blessed people deciding on their own what to do.

And this PR doesn't contribute to the first approach and is also insufficiently thought-out.

Issue #5392, opened 2021-10-15
`builtins.jsonSchemaValidate attrset ./path/to/schema.json`


blaggacao:
**Is your feature request related to a problem? Please describe.**

- I have a config management problem that might for example involve nomad job configs or kubernetes manifests.
- I want to use nix as my _single_ config management language.
- The only type system light that I have currently available is the `nixpkgs` (not `nix`!) module system.
- I have to re-implement config types manually.
- Error reporting UX of the `nixpkgs` module system is not acceptable for my "internal clients".

**Describe the solution you'd like**
- Since official json schemata are ubiquitously available either as schema or as `openapi` spec or similar
- I do `builtins.jsonSchemaValidate attrset ./path/to/schema.json` to validate my my config
- _without_ IFD
- _with_ great error provenance reporting

**Describe alternatives you've considered**
- IFD with `ajv`
- Does not solve the problem of error provenance reporting.

**Additional context**
- https://github.com/fluidattacks/makes/issues/619

---

There are a few candidate json schema validate C++ libs that I've found on https://json-schema.org/

<ul>
<li><a href="https://github.com/KayEss/json-schema">f5-json-schema</a> <em>draft-07</em> (Boost Software License 1.0)</li>
<li><a href="https://github.com/pboettch/json-schema-validator">JSON schema validator for JSON for Modern C++</a> <em>draft-07</em> (MIT)</li>
<li><a href="https://github.com/tristanpenman/valijson">Valijson</a> <em>draft-07</em> header-only library, works with many JSON parser implementations (BSD-2-Clause)</li>
<li><a href="https://github.com/danielaparker/jsoncons/blob/master/doc/ref/jsonschema/jsonschema.md">jsoncons</a> <em>draft-07</em> Header-only library(Boost Software License 1.0)</li>
</ul>

/cc @kamadorueda

kamadorueda:
I'm willing to work on this one (on my free time)

I'll be familiarizing myself a little bit with the codebase in #3505

tomberek:
Closing as stale. Re-open if needed.

PR #5403, opened 2021-10-18
builtins.validateAsJSON

kamadorueda:
- Add https://github.com/pboettch/json-schema-validator
  as a dependency

  `json-schema-validator` is a library that uses `nlohmann` internally,
  which is being used in other components of Nix already

  `json-schema-validator` is MIT licensed,
  complies with the latest draft of JSON schema
  and is the second one with most stars on GitHub
- Add a new builtin for validating data with a JSON schema.

  This builtin receives two arguments: `schema` and `data`,
  both in Nix format
- Demo:

  ```nix
  let
    schema = {
      title = "A person";
      properties = {
        age = {
          description = "Age of the person";
          type = "number";
          minimum = 1;
          maximum = 200;
        };
        name = {
          description = "Complete Name for the person";
          first.type = "string";
          last.type = "string";
          required = [ "first" "last" ];
          type = "object";
        };
      };
      required = [ "name" "age" ];
      type = "object";
    };
  in
  map (validateAsJSON schema) [
    { }
    { age = 24; name.first = "Jane"; }
    { age = 24; name.first = "Jane"; name.last = "Doe"; }
  ]
  ```

  ```bash
  $ nix-instantiate --eval --strict test.nix
  [
    { success = false;
      value = "At '/', required property 'name' not found in object"; }
    { success = false;
      value = "At '/name', required property 'last' not found in object"; }
    { success = true;
      value = { age = 24; name = { first = "Jane"; last = "Doe"; }; }; }
  ]
  ```
- This demo was also added to the documentation in a simplified form:
  renderized: image

- Added language tests

Mentions #5392

### DIFF
--- a/src/libexpr/primops/validateAsJSON.cc
+++ b/src/libexpr/primops/validateAsJSON.cc
@@ -1,0 +1,114 @@
+#include "eval-inline.hh"
+#include "primops.hh"
+#include "value-to-json.hh"
+
+#include <iostream>
+#include <sstream>
+
+#include <nlohmann/json-patch.cpp>
+#include <nlohmann/json-schema.hpp>
+#include <nlohmann/json-uri.cpp>
+#include <nlohmann/json-validator.cpp>
+#include <nlohmann/json.hpp>
+
+class custom_error_handler : public error_handler
+{
+    void error(const json::json_pointer &ptr, const json &instance, const std::string &message) override
+    {
+        std::string pos = ptr.to_string();
+
+        if (pos == "")
+            pos = "/";
+
+        throw std::invalid_argument("At '" + pos + "', " + message);
+    }
+};
+
+namespace nix
+{
+
+static void prim_validateAsJSON(EvalState &state, const Pos &pos, Value **args, Value &v)
+{
+    state.forceValue(*args[0], pos);
+    state.forceValue(*args[1], pos);
+
+    PathSet context;
+    std::ostringstream dataStr;
+    std::ostringstream schemaStr;
+    printValueAsJSON(state, true, *args[0], schemaStr, context);
+    printValueAsJSON(state, true, *args[1], dataStr, context);
+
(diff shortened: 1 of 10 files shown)


CitadelCore:
Would be really nice to have this in.

Princemachiavelli:
I would really like to start using this as well. I think one of the most common use cases would be validating against an external JSON schema. I plan on doing some tests with how it interacts with `fromJSON` and how URIs are handled or could be handled.

edolstra:
Maybe this should be an experimental feature initially? This is probably the most complex builtin in the Nix language, so it's good not to commit to it too quickly.

`validateAsJSON` should not return an error string if the input fails to validate. Error strings are likely to be unstable across versions of `json-schema-validator`, which would lead to a difference in evaluation results between Nix versions.

adrian-gierakowski:
What remains to be done for this to become mergeable?

Ericson2314:
This seems quite heavyweight? What is the intended use-case? Just asserting something is valid before processing it, or doing something else with validation errors?

hsjobeki:
Although I enjoyed the demo, this is excellent work. I have one criticism point:

    ```
    $ nix-instantiate --eval --strict test.nix
    [
      { success = false;
        value = "At '/', required property 'name' not found in object"; }
      { success = false;
        value = "At '/name', required property 'last' not found in object"; }
      { success = true;
        value = { age = 24; name = { first = "Jane"; last = "Doe"; }; }; }
    ]
    ```
The 'value' attribute in the returned attrset has different types and semantic meanings depending on the success field.

Maybe it is better to have just one field that contains the JSON
And an extra field for the error message, which is empty in the case of `success=true;` This clearly separates the error message from the JSON value. 

Just an idea to think of...

thufschmitt:
We discussed it during the Nix team meeting. The agreement is that it's not something we want in Nix (if only because it would tie the language spec to the one of jsonschema). But it could be implemented as either a plugin, or in pure Nix

PR #5416, opened 2021-10-22
--debugger flag

bburdette:
This is an exploratory PR implementing a --debugger flag in the nix command.  If there's an exception, it kicks you into the repl where you can inspect values from the current Env when the exception happened.  

Its by no means complete!  But perhaps best to see if I'm on the right track now before committing more time to this effort.

### Symbols at nix evaluation time

During nix script execution, Values are stored in Env structs, and nix Exprs know the offsets needed to find these values.  Those offsets originate in StaticEnvs during the bindVars phase, before execution happens.  StaticEnvs contain symbols and their offsets.

Normally StaticEnvs are thrown away after bindVars, so if there's an exception, there's no way to know what symbols are in scope.  

### what this PR does

This PR stores each StaticEnv in a shared_ptr instead of as a stack variable.  In debug mode, the shared_ptrs are in turn stored in the Exprs.  Then when an exception occurs in an Expr, it can refer to its StaticEnv to see what symbols are in scope.  

Sample output is below.  The immediate env symbols are listed first with prefix '0', and surrounding envs follow, up to the builtins with prefix '3'.  

```
bburdette@BB-5520:~/code/nix-error-project/REPL-on-error$ ./rec-undef.sh
envtype: an integer
error: cannot add a string to an integer

       at /home/bburdette/code/nix-error-project/REPL-on-error/rec-undef.nix:21:36:

           20|                 in
           21|                 builtins.toString (q + r + ".1");
             |                                    ^
           22|

Starting REPL to allow you to inspect the current state of the evaluator.

0r
0q
1meta
1version
1pname
1nativeBuildInputs
1buildInputs
1src
1subPackages
1sourceRoot
1cargoSha256
2callPackage
2lib
2stdenv
2openssl
2fetchFromGitHub
2rustPlatform
2sqlite
2pkgconfig
3builtins
3__getFlake
3true
3false
3null
3__currentTime
3__currentSystem
3__nixVersion
3__storeDir
3__langVersion
3__nixPath
3scopedImport
3import
3__typeOf
3isNull
3__isFunction
3__isInt
3__isFloat
3__isString
3__isBool
3__isPath
3__genericClosure
3abort
3throw
3__addErrorContext
3__tryEval
3__getEnv
3__seq
3__deepSeq
3__trace
3derivationStrict
3placeholder
3__toPath
3__storePath
3__pathExists
3baseNameOf
3dirOf
3__readFile
3__findFile
3__hashFile
3__splitVersion
3__unsafeDiscardStringContext
3__hasContext
3__unsafeDiscardOutputDependency
3__getContext
3__appendContext
3__readDir
3__toXML
3__toJSON
3__fromJSON
3__toFile
3__filterSource
3__path
3__attrNames
3__attrValues
3__getAttr
3__unsafeGetAttrPos
3__hasAttr
3__isAttrs
3removeAttrs
3__listToAttrs
3__intersectAttrs
3__catAttrs
3__functionArgs
3__mapAttrs
3__isList
3__elemAt
3__head
3__tail
3map
3__filter
3__elem
3__concatLists
3__length
3__foldl'
3__any
3__all
3__genList
3__sort
3__partition
3__concatMap
3__add
3__sub
3__mul
3__div
3__bitAnd
3__bitOr
3__bitXor
3__lessThan
3toString
3__substring
3__stringLength
3__hashString
3__match
3__split
3__concatStringsSep
3__replaceStrings
3__parseDrvName
3__compareVersions
3fetchMercurial
3fetchTree
3__fetchurl
3fetchTarball
3fetchGit
3fromTOML
3derivation
expr:
((q + r) + ".1")
The following extra variables are in scope: __add, __addErrorContext, __all, __any, __appendContext, __attrNames, __attrValues, __bitAnd, __bitOr, __bitXor, __catAttrs, __compareVersions, __concatLists, __concatMap, __concatStringsSep, __currentSystem, __currentTime, __deepSeq, __div, __elem, __elemAt, __fetchurl, __filter, __filterSource, __findFile, __foldl', __fromJSON, __functionArgs, __genList, __genericClosure, __getAttr, __getContext, __getEnv, __getFlake, __hasAttr, __hasContext, __hashFile, __hashString, __head, __intersectAttrs, __isAttrs, __isBool, __isFloat, __isFunction, __isInt, __isList, __isPath, __isString, __langVersion, __length, __lessThan, __listToAttrs, __mapAttrs, __match, __mul, __nixPath, __nixVersion, __parseDrvName, __partition, __path, __pathExists, __readDir, __readFile, __replaceStrings, __seq, __sort, __split, __splitVersion, __storeDir, __storePath, __stringLength, __sub, __substring, __tail, __toFile, __toJSON, __toPath, __toXML, __trace, __tryEval, __typeOf, __unsafeDiscardOutputDependency, __unsafeDiscardStringContext, __unsafeGetAttrPos, abort, baseNameOf, buildInputs, builtins, callPackage, cargoSha256, derivation, derivationStrict, dirOf, false, fetchFromGitHub, fetchGit, fetchMercurial, fetchTarball, fetchTree, fromTOML, import, isNull, lib, map, meta, nativeBuildInputs, null, openssl, pkgconfig, placeholder, pname, q, r, removeAttrs, rustPlatform, scopedImport, sourceRoot, sqlite, src, stdenv, subPackages, throw, toString, true, version

nix-repl> q
5

nix-repl> fromTOML
«primop»

nix-repl>
```

### Next steps

So if the shared_ptr approach is deemed ok for StaticEnv, there are more tasks before this is ready:

- ensure Env and Expr are available at every exception throw.  That means adding them as parameters to functions like forceInt().
- add a builtin for entering the repl without having to artificially force an exception - builtins.break() perhaps.
- have some repl commands for displaying the StaticEnv symbols at various Env levels.
- deal with ExprWith.  Either evaluate With expressions when entering the debugger, or provide a way to force With evaluation, and indicate that there are unevaluated With expression in the Env stack.

This PR doesn't get us a call stack.  That would be nice to have, but would involve more changes, maybe to Env.  Maybe better for another PR if this lands.


domenkozar:
What would be great is to have instructions how to build this PR and give it a try if someone gets stuck. To gather some info if it helps!

bburdette:
Good call.  I'm going to do another round of code cleanup on this, and add something for presenting the in-scope vars in a more aesthetic way.

bburdette:
I've made some progress with this!  I have a little blog article about it [here](https://www.zknotes.com/page/2021-12-27%20nix%20debugger%20progress).  

What's different from last time:  I added a DebugTrace stack in the EvalState, which is updated when you're in --debugger mode.  Also, the latest error is stashed there for reference.  

There are 3 new repl commands:

`:d stack`   - shows the trace stack, similar to the output you get with --show-trace.
`:d error`    - shows the exception that prompted the debugger.
`:d env`      - shows the symbols with values in the current chain of Envs.  

## So to sum up:

- StaticEnvs are stored in shared_ptrs in the Expr objects.  That way we can look up the symbols that are in scope when the error occurs.
- DebugTrace objects are stored in EvalState.  In debug mode, they are added/removed by DebugTraceStacker objects during eval.
- New repl commands.

The DebugTrace approach is similar to what's used by the evalSettings.traceFunctionCalls feature - create a sentinel object if you're in trace mode, otherwise nullptr.  

## What isn't here:

- some exceptions in eval.cc use a 'fakeEnv' rather than a real one, and don't have an Expr either.  To make them work correctly an Env and Expr would need to be passed in.
- I haven't worked on the primops exceptions yet.  
- with expressions will likely need special treatment in `:d env`.

## trying this yourself

Clone this branch and build nix as usual.   

I have a folder of tests I've been using, [here](https://github.com/bburdette/nix-error-proposal/tree/master/repl-tests).  There's a readme.

bburdette:
For those who are interested;  now this has the debug-step behavior shown in this blog post: [full nix debugger[(https://www.zknotes.com/page/2022-01-08%20full%20nix%20debugger).  Execution will halt each time a DebugTrace is created.  To proceed to the next DebugTrace, quit the repl with ctrl-d.  To stop nix you'll have to kill the process unfortunately.  

This is by no mean the final form of the debugger - some folks wanted to try it out so here it is.  I'll be adding more sane controls to it soon.

bburdette:
debugger progress is made!  I added step and go commands to the repl.  Now the debugger specific commands are:

```
  :d <cmd>      Debug mode commands
  :d stack      Show call stack
  :d env        Show env stack
  :d error      Show current error
  :d go         Go until end of program, exception, or builtins.break().
  :d step       Go one step
  ```

So now ctrl-d or :q will actually quit the repl instead of going to the next DebugTrace.  
  
Also, there is a builtins.break now.  You call it on a value, and when the Expr is evaluated, the break happens.  Then you end up in the repl with a message like this:
  
```
info: breakpoint reached; value was "source/server"

      at /home/bburdette/op-code/nix-error-proposal/repl-tests/break.nix:45:16:

          44|
          45|   sourceRoot = break "source/server";
            |                ^
          46|   cargoSha256 = "1sx9ihzwn1vb1dx48ykzg1b9xdbjk4cbzs89labhgljhh51sr64p";

Starting REPL to allow you to inspect the current state of the evaluator.

The following extra variables are in scope: callPackageWith, callPackagesWith, commonAttrs, condition, drv, extendDerivation, hydraJob, lib, makeOverridable, makeScope, makeScopeWithSplicing, outputToAttrListElement, outputs, outputsList, overrideDerivation, passthru

Welcome to Nix 2.7.0. Type :? for help.

nix-repl>
```  

### next up

My goal is to have all the EvalError (and descendant) exceptions kick you into the repl when `--debugger` is set.  This time around I expanded EvalError coverage to the forceValue family of functions (forceInt and etc), and also to the primops.  I have about 20 or so EvalError calls to look at in the rest of nix, all the ones in eval.cc and primops.cc should be handled.  

Beyond that:
 - safety and UI implications of debug-ception, where you enter the debugger again when you eval something at the repl
 - some exceptions don't give me the local vals I'm expecting.  
 - debuggerHook is defined multiple places
 - `with`, and how it shows up in the `:d env`.
 - displaying more than 3 lines of code at a time.
 - inspecting the envs at different levels of the call stack.

**Comment by bburdette** [comment]
Latest commits address these points.
 - safety and UI implications of debug-ception, where you enter the debugger again when you eval something at the repl
   - debug-ception errors go on the stack like other errors, and you can go back with ctrl-D.
 - debuggerHook is defined multiple places
   - fixed this.
 - `with`, and how it shows up in the `:d env`.
   - `:d env` now shows both 'with' and 'static' names.
 - inspecting the envs at different levels of the call stack.  
   - `:d stack` displays numbers beside each DebugTrace.
   - new command `:d show <index>` will load that DebugTrace into the repl, allowing inspection of the relevant symbols.
 - EvalError coverage
   - All the EvalErrors should kick off the debugger now, except in cases where EvalErrors are thrown but nixexpr evaluation isn't under way.

Still outstanding:
 - displaying more than 3 lines of code at a time.
 - a repl command for showing code would be great, for instance printing the definition of a lambda expr.  
 - occasional wonky behavior.  for instance:
   - unexpected 'infinite recursion encountered' when evaling some vars.
   - break doesn't get into the DebugTrace stack correctly
 - probably will take a look at the `:d` commands and maybe do some renaming and etc.

**Comment by bburdette** [comment]
Ok this is ready for some feedback.  Current debugger commands are:

  :d <cmd>      Debug mode commands
  :d stack      Show trace stack
  :d env        Show env stack
  :d show <idx> Show current trace, or change to call stack index
  :d go         Go until end of program, exception, or builtins.break().
  :d step       Go one step

And there's a `break <expr>` primop too.  

I got rid of the `:d error` which showed the current breakpoint - instead its `:d show` now.  

I also turned off the debugger during `tryEval` since its pretty annoying to stop on those, which aren't typically what you're trying to debug.  

with `:d stack`, each DebugTrace prints with an index number next to it.  You can `:d show 3`, for instance, to load the env for a specific DebugTrace.

Re the outstanding issues from last time:


- displaying more than 3 lines of code at a time.  
   - nice-to-have.  leave for a later PR.
- a repl command for showing code would be great, for instance printing the definition of a lambda expr.
   - nice-to-have.  leave for a later PR.
- occasional wonky behavior. for instance:
   - unexpected 'infinite recursion encountered' when evaling some vars.
     - this is due to vars under construction when the exception occurs, so they're 'blackholes'.  would be nice to show these in another color, but leaving it for a later PR for now.
   - break doesn't get into the DebugTrace stack correctly 
     - fixed.

**Comment by bburdette** [comment]
finally working with macos, phew.   removed nonexisting 'libnix'.

File: src/libexpr/eval.hh [hunk 1]
```diff
@@ -35,6 +37,7 @@ struct PrimOp
     const char * doc = nullptr;
 };
 
+typedef std::map<std::string, Value *> valmap;
```
edolstra, line 40: `valmap` won't be scanned for GC roots, is that a problem? If so it should use `std::map<..., traceable_allocator<...>>`.

Also `valmap` should be `ValMap` or something like that to follow our naming conventions.

src/libexpr/eval.cc
```diff
@@ -711,128 +713,354 @@ std::optional<EvalState::Doc> EvalState::getDoc(Value & v)
...
-    throw EvalError({
+    auto error = EvalError({
         .msg = hintfmt(s, symbols[sym], positions[p2]),
         .errPos = positions[p1]
     });
+
+    if (debuggerHook)
+        debuggerHook(&error, env, expr);
+
+    throw error;
```
edolstra, line 945: There is a lot of boilerplate like this, which could be avoided with a helper function that optionally starts the debugger and then throws the exception, e.g.
```
    throwError(EvalError(...));
```

edolstra (approved):

src/libexpr/primops.cc
```diff
@@ -818,6 +902,8 @@ static RegisterPrimOp primop_floor({
 static void prim_tryEval(EvalState & state, const PosIdx pos, Value * * args, Value & v)
 {
     auto attrs = state.buildBindings(2);
+    auto saveDebuggerHook = debuggerHook;
```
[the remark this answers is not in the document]
bburdette: This wasn't my first choice, but without this you always get sidetracked into a tryEval right away, like this:

error: file 'nixpkgs-overlays' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at /nix/store/wb7y020amk3ylb3cjjv521ja8lrnnfbs-nixos-21.11.337329.fd43ce017d4/nixos/pkgs/top-level/impure.nix:42:36:

           41|       isDir = path: pathExists (path + "/.");
           42|       pathOverlays = try (toString <nixpkgs-overlays>) "";
             |                                    ^
           43|       homeOverlaysFile = homeDir + "/.config/nixpkgs/overlays.nix";

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0. Type :? for help.

nix-repl>

src/libcmd/repl.cc
```diff
@@ -434,7 +495,73 @@ bool NixRepl::processLine(std::string line)
...
              << "  :log <expr>   Show logs for a derivation\n"
-             << "  :st [bool]    Enable, disable or toggle showing traces for errors\n";
+             << "  :st [bool]    Enable, disable or toggle showing traces for errors\n"
+             << "  :d <cmd>      Debug mode commands\n"
+             << "  :d stack      Show trace stack\n"
+             << "  :d env        Show env stack\n"
+             << "  :d show       Show current trace\n"
+             << "  :d show <idx> Change to another trace in the stack\n"
+             << "  :d go         Go until end of program, exception, or builtins.break().\n"
+             << "  :d step       Go one step\n"
```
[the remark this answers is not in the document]
bburdette: Ok now I have these more terse commands:

```
        if (debuggerHook) {
             std::cout
             << "\n"
             << "        Debug mode commands\n"
             << "  :env          Show env stack\n"
             << "  :bt           Show trace stack\n"
             << "  :st           Show current trace\n"
             << "  :st <idx>     Change to another trace in the stack\n"
             << "  :c            Go until end of program, exception, or builtins.break().\n"
             << "  :s            Go one step\n"
             ;
        }
```

I renamed the existing `:st` ("show trace") to `:te` ("trace enable"), and the old `:s <expr>` for starting a shell with a derivation is now `:sh`.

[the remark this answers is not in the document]
bburdette: Also "backtrace" is a synonym for "bt", and similarly there are "step" and "continue" synonyms.

src/libexpr/primops.cc
```diff
@@ -725,6 +760,44 @@ static RegisterPrimOp primop_genericClosure(RegisterPrimOp::Info {
...
+    .args = {"v"},
+    .doc = R"(
+      In debug mode (enabled using `--debugger`), pause Nix expression evaluation and enter the REPL.
+      Otherwise, return the argument `v`.
+    )",
+    .fun = [](EvalState & state, const PosIdx pos, Value * * args, Value & v)
+    {
+        if (debuggerHook && !state.debugTraces.empty()) {
+            PathSet context;
+            auto s = state.coerceToString(pos, *args[0], context).toOwned();
```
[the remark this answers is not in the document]
bburdette: For now I went with the simple solution - don't print the argument at all.  Seems fine as you can see pretty clearly where you are in the code.

src/libexpr/eval.hh
bburdette: Ok its ValMap now and traceable_allocator

src/libexpr/eval.cc
bburdette: Ok I [tried that](https://github.com/bburdette/nix/commit/2c9fafdc9e43f6da39c289888dedbbbf0ea0b208), and remembered why I didn't do it before!  The exception that gets thrown ends up being a generic Error rather than a specific subclass, which breaks `tryEval` and probably other code that depends on the exception type.  

But I did eliminate some boilerplate with a `debugError` function.

```
void debugError(Error * e, Env & env, Expr & expr)
{
    if (debuggerHook)
        debuggerHook(e, env, expr);
}
```

bburdette: the template function approached worked out.  so now these:
```
    debugLastTrace(error);

    throw error;
```

look like this:
```
    debugThrowLastTrace(error);
```

src/libexpr/nixexpr.cc
```diff
@@ -348,73 +360,90 @@ void ExprVar::bindVars(const EvalState & es, const StaticEnv & env)
     this->level = withLevel;
 }
 
-void ExprSelect::bindVars(const EvalState & es, const StaticEnv & env)
+void ExprSelect::bindVars(const EvalState & es, const std::shared_ptr<const StaticEnv> & env)
 {
+    if (debuggerHook)
+        staticEnv = env;
```
roberth, line 366: This has a performance cost in "production" mode, especially the `shared_ptr`, I suspect.
Do these ever get released? A plain pointer and `new` may be low hanging fruit here.

It would also be nice to defer the assignment to the debugger code, so we don't have the (small) overhead here. I guess you could look up the root of the ast with `pos` and work from there? It seems a bit tricky to do this without duplicating the `bindVars` logic though.
If you could conditionally throw an exception when the old `bindVars` impl reaches the breakpoint, it seems you could recover all lost performance.

1. bindVars runs as usual
2. evaluator reaches breakpoint
3. breakpoint sets a private mutable field (`isRetrieveStaticEnvMode = true`)
4. breakpoint invokes `bindVars` (again)
5. `bindVars` reaches the breakpoint
6. `ExprBreakpoint::bindVars` sees `isRetrieveStaticEnvMode == true`, copies the `StaticEnv`, throws it
7. evaluator catches it
8. evaluator starts repl

I guess this doesn't work for stepping though, unless you alter the AST to insert more breakpoints. Yikes.

Maybe we can challenge @pennae for a better idea?

MatthewCroughan:
Attempting to perform tab completion on a name who's value is bound to the result of a `builtins.break` function call will cause a segfault. Easy to reproduce:

```
nix-repl> x = builtins.break {}

nix-repl> x.Segmentation fault (core dumped)
```  

Additionally, I have noticed that the eval cache in `~/.cache/nix` can play tricks when trying to perform debugs. Perhaps it's best if `--debugger` would ignore this eval cache, as not to confuse the person debugging.

MatthewCroughan:
When a path is available in the Nix code you're trying to debug, you have to use `--impure` in order to access that path. There may be a solution to providing these paths purely, just a suggestion.

In the following example, `../nix/haskell.nix` is not available, however if I pass `--impure` I can access it from a relative path.

```
 nix build --debugger --impure
warning: Git tree '/home/matthew/tmp/forlee' is dirty
info: breakpoint reached

      at /nix/store/6m32flw60xa76mn84pgrvrs1ivif69i2-source/onchain/default.nix:4:12:

           3| let
           4|   inherit (builtins.break (inputs.nixpkgs.legacyPackages.${system}.callPackage ../nix/haskell.nix { inherit inputs system; }))
            |            ^
           5|     pkgs plutusProjectIn ghcid;

Starting REPL to allow you to inspect the current state of the evaluator.
```

src/libexpr/nixexpr.cc
bburdette: The shared_ptrs are released when the Expr tree is released.  There isn't a one-to-one correspondence between Envs and Exprs, so with plain pointers you can't just delete when the Expr is deleted.  In production mode the SharedPtrs are not copied so there shouldn't be much cost involved in freeing them - just the check at the site where the StaticEnv is created, and the `if (debuggerHook)` checks.

bburdette:
> Attempting to perform tab completion on a name who's value is bound to the result of a `builtins.break` function call will cause a segfault.

@MatthewCroughan That's not good!  But I don't get the same result:

```
[bburdette@HOSS:~/op-code/nix-error-proposal/repl-tests]$ ./undefined-variable.sh
error: undefined variable 'description'

       at /home/bburdette/op-code/nix-error-proposal/repl-tests/undefined-variable.nix:46:17:

           45|     homepage = https://github.com/bburdette/zknotes;
           46|     license = [ description ];
             |                 ^
           47|     # license = [ licenses ];

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0. Type :? for help.

nix-repl> x = builtins.break {}

info: breakpoint reached

      at «string»:1:2:

           1|  builtins.break {}
            |  ^

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0. Type :? for help.

nix-repl>
```

It doesn't show up in the above output, but I get a "breakpoint reached" when I `nix-repl> x.<tab>`, not a segfault.  Maybe there's something else in your test setup that leads to segfault?  I'd like to address this one!

MatthewCroughan:
@bburdette This seems like a more reproducible example. Does this still not happen for you?
```
user: matthew swordfish in ~/tmp/dbug 
❯ cat default.nix 
{
  x = builtins.break {};
}

user: matthew swordfish in ~/tmp/dbug 
❯ nix eval -f ./default.nix --debugger
Segmentation fault (core dumped)

user: matthew swordfish in ~/tmp/dbug 
❯ nix eval -f ./default.nix x --debugger
{ }

user: matthew swordfish in ~/tmp/dbug 
❯ nix --version
nix (Nix) 2.9.0pre20220512_2acdb90
```

MatthewCroughan:
```
user: matthew swordfish in ~/tmp/dbug 
❯ nix eval -f ./default.nix --debugger --vvvvvvvvvvvvvvvvvvvvv
evaluating file '/home/matthew/tmp/dbug/default.nix'
Segmentation fault (core dumped)
```

MatthewCroughan:
It seems like the debugger did not kick in and drop me into a repl when I did `nix eval -f ./default.nix x --debugger`, when I expected it to.

bburdette:
Its a problem with the thunk in EvalState::forceValueDeep.  Turns out its Expr pointer is null in this case.

bburdette:
@MatthewCroughan that thunk problem should be fixed now.

MatthewCroughan:
@bburdette I can confirm that my minimal example from above is fixed. I can also confirm that tab completion of names bound to the result of `builtins.break {}` is also fixed.

```
user: matthew swordfish in ~/tmp/dbug via ❄️  impure (shell) 
❯ nix eval -f ./default.nix --debugger
info: breakpoint reached

      at /home/matthew/tmp/dbug/default.nix:2:7:

           1| {
           2|   x = builtins.break (builtins.break {});
            |       ^
           3| }

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0pre20220515_86ba0a7. Type :? for help.

error: quit the debugger

user: matthew swordfish in ~/tmp/dbug via ❄️  impure (shell) took 2s 
❯ nix eval -f ./default.nix x --debugger
<CODE>
```

bburdette:
Ok I moved the StaticEnv storage to a std::map<Expr,StaticEnv> that lives in EvalState.  Seems to work fine and should save memory in non-debug mode.

roberth:
The debugger doesn't launch when I run the following. Why could that be?

```
nix run github:bburdette/nix/debug-exploratory-PR#nix -- \
  eval --debugger github:hercules-ci/nixpkgs/getExe-writeScript-2#firefox
```

bburdette:
> The debugger doesn't launch when I run the following. Why could that be?

That's because some errors don't have Pos information available, so they rely on the last entry in the DebugTrace stack.  If there is no such entry, then the debugger doesn't start.  Solution is either get pos information to all those sites, or add more traces.

bburdette:
Probably annoying-but-consistent is better than skipping tryEval exceptions.  That behavior could go into a separate flag perhaps.  This is what you get now:

```
[bburdette@HOSS:~/op-code/nix-error-proposal/repl-tests]$ ./ffthing.sh
error: cannot look up '<nixpkgs-overlays>' in pure evaluation mode (use '--impure' to override)

       at /nix/store/8bjs9453y2bkh1pdxhf3hymvgr026x2j-source/pkgs/top-level/impure.nix:41:36:

           40|       isDir = path: builtins.pathExists (path + "/.");
           41|       pathOverlays = try (toString <nixpkgs-overlays>) "";
             |                                    ^
           42|       homeOverlaysFile = homeDir + "/.config/nixpkgs/overlays.nix";

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0. Type :? for help.

nix-repl> :c
error: value is a function while a set was expected

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0. Type :? for help.

nix-repl>
```

bburdette:
This addresses the concern from @pennae about the global debuggerHook.  Instead there is a debugRepl function pointer that lives in evalState, and the logic from the debuggerHook closure mostly moves into a EvalState::runDebugRepl member function.

An issue I ran into here was that the NixRepl takes a ref<EvalState> as an arg.  But if EvalState keeps a ref to itself, it would never get deleted!  And if it creates a ref on the spot to pass to NixRepl, then you have two 'lineages' of refs to the same EvalState pointer, which would lead to a double dealloc.  

So, I went with making the EvalState that gets passed to NixRepl a plain reference instead of a ref<>.  This counts on EvalState being managed by the ref that lives in EvalCommand, where EvalState is created. 

With this approach there's also the consequence of de-consting all the EvalState::throw* members, since staring a repl with EvalState is not a const thing to do.  This makes sense to me; the debuggerHook was kind of making an end run around constness before. 

An alternative here would be to put the EvalCommand into a debugRepl closure function, and then when its time to run the repl the EvalCommand could supply a ref<EvalState> to pass to NIxRepl.  But such a closure isn't copyable and would make creating more EvalStates more difficult.

edolstra:
> An issue I ran into here was that the NixRepl takes a ref as an arg.

This can be solved by making `EvalState` a subclass of `public std::enable_shared_from_this` and calling `ref(shared_from_this())` to get a new `ref` pointer.

edolstra:
@bburdette Could you add an entry to the release notes (`rl-next.md`)? Thanks!

pennae:
we've run this through our performance benchmarks and see a roughly 1% regression for search and parsing, about 2% for system eval. that seems well acceptable for a debugger that could be expanded to greater capabilities with no extra cost in the future, especially considering that it's still a lot faster than 2.8.1 :)

<details><summary>just spitballing possible optimizations</summary>
some of the runtime overhead may still be avoidable by using AST manipulation to generate debug trace events instead of checking `debugRepl` every time. presumably the same thing could be used to store the StaticEnvs, pushing them onto a stack when entered and popping them back off when exited (which would also obviate the need for the Expr->StativEnv map). such a modification should be relatively easy to implement as a new insertBreakpoints virtual if we were to split bindVars into createEnv and bindEnv, then having the wrapper expr just call createEnv on its child a second time and storing the result in itself :thinking: 
</details>

edolstra:
I also did some performance testing and got a small slowdown on `nix-env -qa` (p=0.00228, Δ=0.04154±0.03788, mean=6.6255) and no statistically significant slowdown on NixOS evaluation.

edolstra:
@bburdette Awesome, merged! Thanks for all your work on this!

573:
Thanks to @bburdette for this great addition to the nix toolbelt also thanks to all who contributed!

Artturin:
https://github.com/NixOS/nix/issues/6594

roberth:
This is a great feature but should not have been merged without tests.
 - can we please #8610

PR #5420, opened 2021-10-23
Allow to disable global flake-registry with ""

### LINKED ISSUE #4874: Allow empty config `flake-registry = ""`, to disable default global flakes
For my Nix install, I **don't** want any global flake registries, as these are non-deterministic global states.

To make the flake registry idempotent and not use any global state, I would like:
:point_right: The possibility to set `flake-registry = ""` in `nix.conf`, to skip the load of global flakes.

Creating a custom file containing `{}`, and pointing the config option to it.

The main downside being that the `nix.conf` file is not standalone anymore, and depends on a non-Nix-managed file.

bew:
Closes #4874

I'm not sure why the implementation of `getGlobalRegistry` was wrapped in a lambda, but returning nullptr from it made the compiler not able to find a type for both nullptr and Registry for the lambda..
So I simply removed the lambda, problem solved AFAI can see.

### DIFF
--- a/src/libfetchers/fetch-settings.hh
+++ b/src/libfetchers/fetch-settings.hh
@@ -71,7 +71,12 @@ struct FetchSettings : public Config
         "Whether to warn about dirty Git/Mercurial trees."};
 
     Setting<std::string> flakeRegistry{this, "https://channels.nixos.org/flake-registry.json", "flake-registry",
-        "Path or URI of the global flake registry."};
+        R"(
+          Path or URI of the global flake registry.
+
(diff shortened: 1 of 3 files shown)


asymmetric:
Shouldn't there be an entry in the release notes for this? Happy to create it if that's the consensus.

thufschmitt:
> Shouldn't there be an entry in the release notes for this? Happy to create it if that's the consensus.

Indeed, thanks for pointing that out. Happy to review it if you create one

PR #5424, opened 2021-10-23
Allow references in addPath

### LINKED ISSUE #5410: filterSource: error: store path <PATH> is not allowed to have references
**Describe the bug**

Before 4806f2f6b0fd2cae401b89fe19d8c528ffd88b5f, it was possible to have references in paths being filtered. I think this was actually incorrect, since the resulting path lost all references (at least that's what https://github.com/NixOS/nix/pull/5347#issue-1019953804 implies). The comment also says that 

> This also makes builtins.{path,filterSource} work on derivation outputs if IFD is enabled.

However, actually `filterSource` did work on derivation outputs before then, for some reason. This was used in `crate2nix`, among other projects. Now it doesn't work with crate2nix, since the output of the derivation we're filtering (and then importing from) contains references.

**Steps To Reproduce**

```
$ nix build --no-eval-cache github:soywod/himalaya/767eee95f52f86000b76f2f10dde951046214bff
error: store path '/nix/store/wn9mdf67b6ah1555d2cm0gl53gix8cm2-himalaya-crate2nix' is not allowed to have references
```

**Expected behavior**

The derivation is built, as on Nix before that commit.

I suppose the solution is to allow the filtered paths to have references, but then scan the new, filtered path for references afterwards. Just passing through references may result in dangling references if the filter cleans some of them out.

**Nix version**

`nix (Nix) 2.5pre20211008_4c17ebe`

balsoft:
Allows `addPath` on paths with references, therefore allowing `builtins.path` and `builtins.filterSource` to work on paths with references.

Closes #5410

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1859,12 +1859,12 @@ static void addPath(
         // be rewritten to the actual output).
         state.realiseContext(context);
 
+        StorePathSet refs = StorePathSet();
+
         if (state.store->isInStore(path)) {
             auto [storePath, subPath] = state.store->toStorePath(path);
-            auto info = state.store->queryPathInfo(storePath);
-            if (!info->references.empty())
-                throw EvalError("store path '%s' is not allowed to have references",
-                    state.store->printStorePath(storePath));
+            // FIXME: we should scanForReferences on the path before adding it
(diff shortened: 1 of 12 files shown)


balsoft:
Superceded by #5494

PR #5432, opened 2021-10-26
Make experimental-features a proper type

thufschmitt:
Rather than having them plain strings scattered through the whole codebase, create an enum containing all the known experimental features.

This means that
- Nix can now `warn` when an unkwown experimental feature is passed (making it much nicer to spot typos and spot deprecated features)
- It’s now easy to remove a feature altogether (once the feature isn’t experimental anymore or is dropped) by just removing the field for the enum and letting the compiler point us to all the now invalid usages of it.


Ericson2314:
Yay! Now we can break these up by library/layer.

PR #5434, opened 2021-10-26
Support building flakes from a Git repo url with submodules

timothyklim:
This patch can support workaround for #5284: `nix build "git+file://$(pwd)?submodules=1"`.


kreisys:
... siphoning submodules stakeholders to https://github.com/NixOS/nix/pull/5497

PR #5437, opened 2021-10-26
Fix flake relative paths

roberth:
This may be the `call-flake.nix` part of a fix for https://github.com/NixOS/nix/pull/5279#issuecomment-941931071.

When a flake references a local path `inputs.foo.url = "path:subflake";`, the path is relative to the defining flake. This relation must be tracked in order to find the source for the path.

This problem can go unnoticed as the original author has the desired path in the store; added as part of creating the lock file. Other contributors are out of luck and will see an error like:

    error: cannot fetch input 'path:subflake?narHash=sha256-BIVxJv82kw1MKcUxusLrmOcf+O%2fqnGZRH7JDlUIIhoU=' because it uses a relative path

The PR (as is) assumes that the root flake is always the parent, which is obviously not true, but it fixes the simple case and is ready to read the `parent` from the lock file when it is added.

### DIFF
--- a/src/libexpr/flake/call-flake.nix
+++ b/src/libexpr/flake/call-flake.nix
@@ -23,7 +23,8 @@ let
                   && !isAbsolutePath node.locked.path;
 
               # FIXME: put parent node reference in lock file and use it here.
-              parentPath = rootSrc + "/${rootSubdir}";
+              parentKey = lockFile.root;
+              parentPath = allNodes.${parentKey}.sourceInfo.outPath;
             in
               if isRelativePathNode
               then { path = parentPath + "/" + node.locked.path; }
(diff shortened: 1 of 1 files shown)


maisiliym:
Tested this PR, and it fails by using the wrong parent. The relative path is in one of the inputs of the flake called, but nix fails by looking for this path in the source of the flake called, instead of looking in the source of the relative input's parent.

roberth:
> so this looks like a bigger problem than the PR addresses.

:100:

Another problem is the use of a hash for local paths that are effectively pinned by the parent. This requires redundant lock file updates that are annoying in normal use and are very breaking when accidentally skipped, reverted, etc.

roberth:
I am not working on this. Someone could use this as a starting point.

tomberek:
Is there a resolution to this? The lazy-trees branch does not seem to address the the problem with relative paths and the common problem of the author easily not realizing this is happening.

roberth:
The `lazy-trees` branch does seem to address this for subflakes in https://github.com/NixOS/nix/pull/6530/commits/8a0a55fe12cd3598729400fa4e9314b4b0490fb9:

> They're no longer locked in the lock file.

I suppose the same problem exists for absolute paths. I'd like those to be caught, but I don't think we can reasonably do so before CI, as `path:` inputs are useful for local development. It'd be up to users to configure something to avoid them, I suppose, like a pre-commit hook or perhaps we could have a `nix.conf` setting to disallow `path:`.

PR #5452, opened 2021-10-30
git: extend cache dir lock over all mutating operations

### LINKED ISSUE #4051: Simultaneous builtins.fetchGit calls can fail.
This small test illustrates the problem.

```bash
#!/usr/bin/env bash

rm -rf ~/.cache/nix/gitv2

nix-instantiate --eval -E '(builtins.fetchGit {url="https://github.com/nixos/nixpkgs"; ref="master"; rev="0f5099067829194dc63f74ee1459269fa096d35d";}).outPath' &
nix-instantiate --eval -E '(builtins.fetchGit {url="https://github.com/nixos/nixpkgs"; ref="master"; rev="aabe6e23a688bbfe81685984f05706e9c945056c";}).outPath' &

for p in $(jobs -p) do
    wait $p;
done
```

This will fail non-deterministically with errors like:

```
fatal: cannot mkdir /home/travis/.cache/nix/gitv2/17blyky0ja542rww32nj04jys1r9vnkg6gcfbj83drca9a862hwp: File exists

fatal: unable to open objects/pack/tmp_pack_yFMUU3: No such file or directory
```

This is straightforwardly problematic for e.g. a build machine, where multiple processes may be evaluating `builtins.fetchGit` expressions with the same URL but different refs/revs at the same time. It seems evaluation should lock checkout paths in `~/.cache/nix/gitv2` before manipulating them.

~Strangely, I'm not able to reproduce this issue on a NixOS machine; it only seems to occur on non-NixOS Linux and MacOS machines, in both single-user and multi-user mode. I'm not sure why this should be the case; `builtins.fetchGit` is part of evaluation, not realization.~ Seems I just got extraordinarily lucky when testing this hypothesis, this is definitely an issue on NixOS too.

Nix version 2.3.6

Kha:
Presumably fixes #4051. I was not able to reproduce the race described there, but parallel fetches of the same repo seems like a fringe use case that should rarely benefit from concurrency (and potentially even suffer from redundant downloads when parallelized?).

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -324,16 +324,12 @@ struct GitInputScheme : InputScheme
             Path cacheDir = getCacheDir() + "/nix/gitv3/" + hashString(htSHA256, actualUrl).to_string(Base32, false);
             repoDir = cacheDir;
 
-            Path cacheDirLock = cacheDir + ".lock";
             createDirs(dirOf(cacheDir));
-            AutoCloseFD lock = openLockFile(cacheDirLock, true);
-            lockFile(lock.get(), ltWrite, true);
+            PathLocks cacheDirLock({cacheDir + ".lock"});
 
             if (!pathExists(cacheDir)) {
                 runProgram("git", true, { "-c", "init.defaultBranch=" + gitInitialBranch, "init", "--bare", repoDir });
             }
-
-            deleteLockFile(cacheDirLock, lock.get());
 
             Path localRefFile =
                 input.getRef()->compare(0, 5, "refs/") == 0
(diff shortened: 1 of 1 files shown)


src/libfetchers/git.cc
```diff
@@ -397,6 +393,9 @@ struct GitInputScheme : InputScheme
                 utimes(localRefFile.c_str(), times);
             }
 
+            // we will only use read-only operations on specific revisions in the remainder
+            cacheDirLock.unlock();
+
```
edolstra, line 398:
```suggestion
```

This is unnecessary since the lock is released when `cacheDirLock` goes out of scope.

Kha: Yeah, I mostly did it to have something explicit to attach the comment to. Moved it to the end of the block instead.

PR #5453, opened 2021-10-30
Docker image with Nix inside

garbas:
This work was borrowed from https://github.com/NixOS/docker/pull/31

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -1,0 +1,264 @@
+{ pkgs ? import <nixpkgs> { }
+, lib ? pkgs.lib
+, name ? "nix"
+, tag ? "latest"
+, crossSystem ? null
+, channelName ? "nixpkgs"
+, channelURL ? "https://nixos.org/channels/nixpkgs-unstable"
+}:
+let
+  buildPkgs = pkgs;
+  targetPkgs =
+    if crossSystem != null && crossSystem != pkgs.system
+    then {
+      aarch64-linux = pkgs.pkgsCross.aarch64-multiplatform;
+      armv7l-linux = pkgs.pkgsCross.armv7l-hf-multiplatform.system;
+      x86_64-linux = pkgs.pkgsCross.gnu64;
+      powerpc64le-linux = pkgs.pkgsCross.musl-power;
+      i686-linux = pkgs.pkgsCross.gnu32;
+    }.${crossSystem}
+    else pkgs;
+
+  defaultPkgs = [
+    targetPkgs.nix
+    targetPkgs.bashInteractive
+    targetPkgs.coreutils-full
+    targetPkgs.gnutar
+    targetPkgs.gzip
+    targetPkgs.gnugrep
+    targetPkgs.which
+    targetPkgs.curl
+    targetPkgs.less
+    targetPkgs.wget
+    targetPkgs.man
+    targetPkgs.cacert.out
+    targetPkgs.findutils
+  ];
+
+  users = {
+
+    root = {
+      uid = 0;
+      shell = "/bin/bash";
+      home = "/root";
+      gid = 0;
+    };
+
+  } // lib.listToAttrs (
+    map
+      (
+        n: {
+          name = "nixbld${toString n}";
+          value = {
+            uid = 30000 + n;
+            gid = 30000;
+            groups = [ "nixbld" ];
+            description = "Nix build user ${toString n}";
+          };
+        }
+      )
+      (lib.lists.range 1 32)
+  );
+
+  groups = {
+    root.gid = 0;
+    nixbld.gid = 30000;
+  };
+
+  userToPasswd = (
+    k:
+    { uid
+    , gid ? 65534
+    , home ? "/var/empty"
+    , description ? ""
+    , shell ? "/bin/false"
+    , groups ? [ ]
+    }: "${k}:x:${toString uid}:${toString gid}:${description}:${home}:${shell}"
+  );
+  passwdContents = (
+    lib.concatStringsSep "\n"
+      (lib.attrValues (lib.mapAttrs userToPasswd users))
+  );
+
+  userToShadow = k: { ... }: "${k}:!:1::::::";
+  shadowContents = (
+    lib.concatStringsSep "\n"
+      (lib.attrValues (lib.mapAttrs userToShadow users))
+  );
+
+  # Map groups to members
+  # {
+  #   group = [ "user1" "user2" ];
+  # }
+  groupMemberMap = (
+    let
+      # Create a flat list of user/group mappings
+      mappings = (
+        builtins.foldl'
+          (
+            acc: user:
+              let
+                groups = users.${user}.groups or [ ];
+              in
+              acc ++ map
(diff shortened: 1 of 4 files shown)


Ericson2314:
You shouldn't need a native band cross package set, instead take a single *host* Nixpkgs, and use `buildPaclages` to get a build time package set out of it.

make-boostrap-tools.nix in Nixpkgs shows what this idiom looks like.

docker.nix
```diff
@@ -0,0 +1,264 @@
...
+
+      mkdir $out/tmp
+
+      mkdir -p $out/etc/nix
+      cat $nixConfContentsPath > $out/etc/nix/nix.conf
+
+      mkdir -p $out/root
+      mkdir -p $out/nix/var/nix/profiles/per-user/root
+
+      ln -s ${profile} $out/nix/var/nix/profiles/default-1-link
```
edolstra, line 216: Maybe it's better to initialize the profile by running `nix-env -i ${defaultPkgs}` or something like that. That way we don't have to generate an (old-style) `manifest.nix` here.

edolstra:
We should document somewhere (e.g. in the installation section of the manual) how to get our docker image and what it provides.

docker.nix
```diff
@@ -0,0 +1,264 @@
...
+    targetPkgs.gzip
+    targetPkgs.gnugrep
+    targetPkgs.which
+    targetPkgs.curl
+    targetPkgs.less
+    targetPkgs.wget
+    targetPkgs.man
+    targetPkgs.cacert.out
+    targetPkgs.findutils
+  ];
```
SuperSandro2000, line 36:
```suggestion
  defaultPkgs = with targetPkgs; [
    nix
    bashInteractive
    coreutils-full
    gnutar
    gzip
    gnugrep
    which
    curl
    less
    wget
    man
    cacert.out
    findutils
  ];
```

docker.nix
```diff
@@ -0,0 +1,264 @@
...
+          name = "nixbld${toString n}";
+          value = {
+            uid = 30000 + n;
+            gid = 30000;
+            groups = [ "nixbld" ];
+            description = "Nix build user ${toString n}";
+          };
+        }
+      )
+      (lib.lists.range 1 32)
```
Mic92, line 60: What is a reasonable number here? I have an AMD EPYC with 64 cores and 128 Hyperthreads.
Should this be configurable?

edolstra: This is the number used by NixOS so it seems reasonable here. Also, if you have lots of cores, you probably want parallelism *inside* a derivation rather than run a lot of derivations in parallel.

Hopefully this will become a non-issue with #3600.

docker.nix
garbas: I'm not 100% but if I remember correctly I wasn't able to generate profile with `nix-env`. Maybe @adisbladis knows more about since this is version of creating profiles.

garbas:
> We should document somewhere (e.g. in the installation section of the manual) how to get our docker image and what it provides.

I'm not sure this would fall under the installation section, since you are not really installing Nix, but you would be just using it like any other docker image. I wonder if the website would be the best place to document it.

roberth:
I would prefer for this image to be defined in the nixpkgs repo, because this repo is bottlenecked by reviews and won't have the pace of development that nixpkgs has. Also, as a dockerTools maintainer, I'd like to include the official nix image in the dockerTools test suite.

cc @utdemir

garbas:
@roberth I don't object to the idea that the helper functions for Nix image to be defined in nixpkgs, but for the sake of moving forward with having a Docker image with latest Nix I would suggest to merge this PR and at your own time. You can bump nixpkgs input that will include this new function and than we can use here.

roberth:
@garbas So if I understand correctly, you're suggesting to keep the final image definition in this repo, like `image = pkgs.buildNixImage { nix = self.nix; }` (omitting `system` tomfoolery for brevity) but have almost all of the implementation of the image in Nixpkgs?

That seems like an improvement, but it does complicate the release process, compared to a release process that is steered by Nixpkgs. I suppose "gatekeeping" could be a good thing for trust. Is that what you had in mind?

edolstra:
> I would prefer for this image to be defined in the nixpkgs repo

It seems very strange to have the release process for Nix in another repo.

> That seems like an improvement, but it does complicate the release process, compared to a release process that is steered by Nixpkgs.

I don't understand this. Why should the Nix release process be steered by Nixpkgs?

roberth:
> > I would prefer for this image to be defined in the nixpkgs repo
> 
> It seems very strange to have the release process for Nix in another repo.

The docker image isn't the same as the Nix package. This repo is hard for the community to make changes, unlike Nixpkgs. If you don't consider Nixpkgs to be part of the Nix release process, I'd find that rather concerning.

> > That seems like an improvement, but it does complicate the release process, compared to a release process that is steered by Nixpkgs.
> 
> I don't understand this. Why should the Nix release process be steered by Nixpkgs?

It reduces the cyclic dependency between Nix and Nixpkgs. Again, I'm not saying Nix should be released by Nixpkgs; just the docker image.

I wouldn't be concerned if this project had more than one maintainer who merges PRs.

edolstra:
>  If you don't consider Nixpkgs to be part of the Nix release process, I'd find that rather concerning.

Ehm...? Nixpkgs is a dependency of the Nix release process, but that doesn't mean the Nix release process should be moved into the Nixpkgs repo. Might as well move it into the gcc repo...

What exactly is the use case for putting this in the Nixpkgs repo?

> It reduces the cyclic dependency between Nix and Nixpkgs.

What cyclic dependency?

roberth:
> What cyclic dependency?

Nixpkgs depending on Nix and Nix depending on dockerTools.

> Might as well move it into the gcc repo...

I figured you might take this to an extreme. That's not the point. Less cyclic is generally better. I'm not saying it is completely avoidable. That would make no sense, as you know.

> > I wouldn't be concerned if this project had more than one maintainer who merges PRs.

I'm just trying to help with the maintenance load. I feel this discussion has reached the tipping point, so I'll shut up now and hope that you'll have time to review fixes.

domenkozar:
@garbas do you plan to setup jobs to push this docker image to docker hub?

Ericson2314:
What about the comments I left in https://github.com/NixOS/nix/pull/5453#issuecomment-955603148 ? I suppose I will make a PR against this, sigh.

garbas:
@domenkozar Yes, this was the first step.
@Ericson2314 As conversation moved later on `docker.nix` will be anyway extracted into `nixpkgs` as a function (eg. `buildNixImage`) which we will than use here. I think having everything in one file will make that easier migrate.

Ericson2314:
@garbas I wasn't asking for more files? I will open a PR in a second with the changes I had in mind. I think it a fine change whether the stuff is living here temporarily or permanently, though yes my original motivation was people will look to this repo to see how things *ought* to be done, and I don't want them to come away with an unusual style.

Ericson2314:
PR's up https://github.com/NixOS/nix/pull/5573

PR #5456, opened 2021-10-31
bundle: pass drv attrset instead of path

tomberek:
@edolstra @matthewbauer @Radvendii This is a WIP and breaking API change for passing a drv's attrset rather than a path with the bundler mechanism. This looks at the bundler to be conceptually of type:

`nix bundle --bundler`:: (Drv -> Drv) -> Drv -> Drv  

Looking at it in this "applyFunc" mindset allows for a wide variety of use-cases:
- defaultNixBundle
- toDockerImage
- toSBOM
- toVulnerabilityScanResult
- toBuildTimeClosure
- toFullBuildFromSourceClosure
- overrideAttrs
- overrideSystem (currently bundle uses `settings.thisSystem.get()`)
- any other override
- toFineGrainedBuild (via recursiveNix/RFC92/nix-make)
- and so on.

This seems more natural and can recover existing behavior of toApp (src/nix/app.cc) with something like:
```
recoverProgram = drv: "${drv.outPath}/bin/${
if drv?meta?mainProgram
then drv.meta.mainProgram 
else (builtins.parseDrvName drv.name).name
}";
```

Starting to collect a set of bundlers that use this API: https://github.com/tomberek/nix-generators


tomberek:
Downside, this would make it more difficult to bundle something for which you can only obtain the runtime closure.

Radvendii:
Another downside: this does not specify the executable to be run, only the derivation. We can discover the specific executable  if there's only one in `/bin` or there is one named the same thing as the derivation, but this is not always the case.

tomberek:
> Another downside: this does not specify the executable to be run, only the derivation. We can discover the specific executable if there's only one in `/bin` or there is one named the same thing as the derivation, but this is not always the case.

I agree, but this is already an issue with the "app" concept and the existing "bundler" implementation. I think that is an issue worth addressing, but this PR should be orthogonal to that. Right now this doesn't change that behavior. Or at least isn't intended to.

tomberek:
TODO: Allow both a raw path, as well as a richer derivation attrset.

tomberek:
- [ ] docs
- [ ] release notes
- [ ] backwards compat?
- [ ] https://github.com/NixOS/bundlers

tomberek:
Updated implementation, added tests, and mangled https://github.com/tomberek/nix-generators in an attempt to support both APIs. (Should probably remove it if/when this is merged)

Some remaining issues:
1) what should be the default attribute be for the programInstallable?
  - defaultApp; this was used before because it was the source of "program".
  - defaultPackage: it is a derivation
2) Are the checks correct?
3) We retain some ability to control mainProgram with meta, but now the interface does not clearly expose this common need.
4) Move back-compat logic into Nix itself to support legacy nix-bundle? (or update nix-bundle to new API?)
  - how many people use this or the ability to extend it oneself already?
5) more docs
6) more tests
7) or does this "apply" function not belong in bundle, but somewhere else?

src/nix/bundle.md
```diff
@@ -18,19 +18,50 @@ R""(
...
+homepage](https://github.com/NixOS/bundlers) for more details.
 
 > **Note**
 >
 > This command only works on Linux.
 
 # Bundler definitions
 
-TODO
+If no flake output attribute is given, `nix bundle` tries the following
```
edolstra, line 42: This section should be renamed `# Flake output attributes` for consistency with other commands.

src/nix/bundle.md
```diff
@@ -18,19 +18,50 @@ R""(
...
+* `defaultBundler.<system>`
+
+If an attribute *name* is given, `nix run` tries the following flake
+output attributes:
+
+* `bundler.<system>.<name>`
+
+# Bundlers
+
+An bundlers is specified by a flake output attribute named
```
edolstra, line 54:
```suggestion
A bundler is specified by a flake output attribute named
```

src/nix/bundle.md
```diff
@@ -18,19 +18,50 @@ R""(
...
+
+```nix
+bundlers.x86_64-linux.identity = drv: drv;
+
+bundlers.x86_64-linux.blender_2_79 = drv: self.packages.x86_64-linux.blender_2_79;
+
+defaultBundler.x86_64-linux = drv: drv;
+```
+
+A bundler must be a function that accepts a derivation and returns a derivation.
```
edolstra, line 65:
```suggestion
A bundler must be a function that accepts an arbitrary value (typically a derivation or app definition) and returns a derivation.
```

tomberek: @edolstra if the arbitrary value is an installable from a flakeref, should the flakeref details be added in some way? Eg: should the bundler be able to inspect the attrPath or the rev or lastModified as available in "self"?

PR #5459, opened 2021-11-02
installer: Do not use echo -e in #!/bin/sh script

### LINKED ISSUE #5458: Installer adds invalid line `-e ` to ~/.profile, leading to `-bash: -e: command not found` at login
On Ubuntu 20.04:

```console
ubuntu@test:~$ curl -L https://nixos.org/nix/install | sh
modifying /home/ubuntu/.profile...

Installation finished!  To ensure that the necessary environment
variables are set, either log in again, or type

  . /home/ubuntu/.nix-profile/etc/profile.d/nix.sh

in your shell.
ubuntu@test:~$ logout
```

Then, upon logging in again:

```console
-bash: -e: command not found
ubuntu@test:~$ tail -n2 .profile
-e 
if [ -e /home/ubuntu/.nix-profile/etc/profile.d/nix.sh ]; then . /home/ubuntu/.nix-profile/etc/profile.d/nix.sh; fi # added by Nix installer
```

This is because the installer uses `echo -e`, which is not supported by Ubuntu’s `/bin/sh` (`dash`).

andersk:
ShellCheck correctly warns:

```
In scripts/install-nix-from-closure.sh line 218:
                echo -e "\nif [ -e $p ]; then . $p; fi # added by Nix installer" >> "$fn"
                     ^-- SC3037: In POSIX sh, echo flags are undefined.

In scripts/install-nix-from-closure.sh line 229:
                echo -e "\nif [ -e $p ]; then . $p; fi # added by Nix installer" >> "$fn"
                     ^-- SC3037: In POSIX sh, echo flags are undefined.
```

Indeed, this actually breaks on Ubuntu where `/bin/sh` is `dash`.

Fixes #5458.

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -215,7 +215,7 @@ if [ -z "$NIX_INSTALLER_NO_MODIFY_PROFILE" ]; then
         if [ -w "$fn" ]; then
             if ! grep -q "$p" "$fn"; then
                 echo "modifying $fn..." >&2
-                echo -e "\nif [ -e $p ]; then . $p; fi # added by Nix installer" >> "$fn"
+                printf '\nif [ -e %s ]; then . %s; fi # added by Nix installer\n' "$p" "$p" >> "$fn"
             fi
             added=1
             break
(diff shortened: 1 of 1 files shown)


regnat:
Do you mind rebasing on top of master? The OS X CI was broken until a couple hours ago

PR #5464, opened 2021-11-02
Update flake.md

bburdette:
remove errant `R""(`


edolstra:
Those are unfortunately needed because we're `#include`-ing the Markdown sources directly in C++. `R""( ... )""` makes it a multi-line string literal.

PR #5470, opened 2021-11-02
[2.4] installer: Do not use echo -e in #!/bin/sh script

### LINKED ISSUE #5458: Installer adds invalid line `-e ` to ~/.profile, leading to `-bash: -e: command not found` at login

andersk:
Cherry-pick to 2.4-maintenance:

* 0b55c8767dfa15540f47579442350ffad51fc047 “Disable the eval-store test when using the daemon” from #5390
  * for passing CI
* 8a3746de0a3a78abdab09fb7e194c58b9bd16172 “install-nix-actionv@v14.1”
  * for passing CI
* #5459
  * fixes #5458

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/checkout@v2.3.4
       with:
         fetch-depth: 0
-    - uses: cachix/install-nix-action@v14
+    - uses: cachix/install-nix-action@v14.1
     - run: echo CACHIX_NAME="$(echo $GITHUB_REPOSITORY-install-tests | tr "[A-Z]/" "[a-z]-")" >> $GITHUB_ENV
     - uses: cachix/cachix-action@v10
       if: needs.check_cachix.outputs.secret == 'true'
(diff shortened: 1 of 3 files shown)


edolstra:
Should we do a 2.4.1 release? I guess the installer being broken on Ubuntu is serious enough...

PR #5471, opened 2021-11-02
command-ref/nix-shell: fix --pure, --keep

simon04:
Currently `man nix-shell` outputs (`–pure` instead of `--pure`):

```
Synopsis
       nix-shell [--arg name value] [--argstr name value] [{--attr | -A} attrPath] [--command cmd] [--run
       cmd] [--exclude regexp] [–pure] [–keep name] {{--packages | -p} {packages | expressions} … | [path]}
```


edolstra:
Thanks!

PR #5472, opened 2021-11-03
async realisation substitution

regnat:
Make the substitution of the `Realisations` async − following what’s done for the Nar fetching.

This gives a small (around ~5%~ 10%) performance improvement when building stuff with a fresh binary-cache cache because of the increased parallelism.

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -437,39 +437,19 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
     })->path;
 }
 
-std::optional<const Realisation> BinaryCacheStore::queryRealisation(const DrvOutput & id)
+void BinaryCacheStore::queryRealisationUncached(const DrvOutput & id,
+    Callback<std::shared_ptr<const Realisation>> callback) noexcept
 {
-    if (diskCache) {
-        auto [cacheOutcome, maybeCachedRealisation] =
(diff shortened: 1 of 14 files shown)


src/libstore/binary-cache-store.cc
```diff
@@ -437,40 +437,29 @@ StorePath BinaryCacheStore::addTextToStore(const string & name, const string & s
...
-
     auto outputInfoFilePath = realisationsPrefix + "/" + id.to_string() + ".doi";
-    auto rawOutputInfo = getFile(outputInfoFilePath);
 
-    if (rawOutputInfo) {
-        auto realisation = Realisation::fromJSON(
-            nlohmann::json::parse(*rawOutputInfo), outputInfoFilePath);
+    auto callbackPtr = std::make_shared<decltype(callback)>(std::move(callback));
+
+    Callback<std::shared_ptr<std::string>> newCallback = {
```
andir, line 447: Just a curious thought: Does the return type really have to be a string in a `shared_ptr`? std::string already manages a heap allocation (outside of SSO) and if the value is moved (or copie elision kicks in) then not using a `shared_ptr` gets rid of one memory indirection.

edolstra: Yes, we should get rid of the shared_ptrs around std::string and replace with moves. This predates our use of C++11.

domenkozar:
cc @tomberek ^^

PR #5474, opened 2021-11-03
Fix some (unimportant) memory leaks

mohe2015:
I don't know if this way to fix them is fine but I saw this as the easiest way to fix them without thoroughly knowing the code.

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -16,6 +16,7 @@
 
 #include <regex>
 #include <queue>
+#include <memory>
 
 #include <sys/un.h>
 #include <fcntl.h>
@@ -1978,7 +1979,7 @@ void LocalDerivationGoal::runChild()
         /* Execute the program.  This should not return. */
         if (drv->isBuiltin()) {
             try {
-                logger = makeJSONLogger(*logger);
+                logger = std::unique_ptr<Logger>(makeJSONLogger(*logger));
 
                 BasicDerivation & drv2(*drv);
                 for (auto & e : drv2.env)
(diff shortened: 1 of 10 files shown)


src/libstore/build/local-derivation-goal.cc
```diff
@@ -1978,7 +1979,7 @@ void LocalDerivationGoal::runChild()
         /* Execute the program.  This should not return. */
         if (drv->isBuiltin()) {
             try {
-                logger = makeJSONLogger(*logger);
+                logger = std::unique_ptr<Logger>(makeJSONLogger(*logger));
```
mohe2015, line 1982: look dangerous to me

src/libstore/daemon.cc
```diff
@@ -944,10 +944,10 @@ void processConnection(
         throw Error("the Nix client version is too old");
 
     auto tunnelLogger = new TunnelLogger(to, clientVersion);
-    auto prevLogger = nix::logger;
+    auto& prevLogger = nix::logger;
```
mohe2015, line 947: looks dangerous

src/libstore/daemon.cc
```diff
@@ -944,16 +944,22 @@ void processConnection(
         throw Error("the Nix client version is too old");
 
     auto tunnelLogger = new TunnelLogger(to, clientVersion);
```
mohe2015, line 946: I hope this could be somehow simplified. Also this could potentially be wrong (to my knowledge only when logger is changed while the tunnelLogger here is still used. As I don't have the domain specific knowledge I don't know if this is the case.

mohe2015:
The logger variable has uses at other places that make this more complicated. Closing. Will create another PR for only the first change.

Issue #2357, opened 2018-08-15
Use sqlite dotfile vfs to allow use over NFS


wmertens:
See http://sqlite.1065341.n5.nabble.com/Safe-sqlite-over-remote-filesystem-tp103435p103438.html

Basically, by using the dotfile vfs, it should be safe to share a nix store over NFS.

The easiest way to achieve this is to simply add `"unix-dotfile"` as the fourth parameter at https://github.com/NixOS/nix/blob/19265ed26c5bd25b299896ed851bda6c444cf1a8/src/libstore/sqlite.cc#L31

However, to make this optional, I think the best way would be to instead add `SQLITE_OPEN_URI` as a flag in the third parameter on that same line, and allow specifying the `path` to the db file as an option. That `db-path` could then e.g. be `"file:/nix/store/shared-db?vfs=unix-dotfile"`, protecting against misconfiguration.

louisabraham:
Hi, I don't think that's true, at least with NFS version 3.

domenkozar:
This patch was mentioned in https://rgoswami.me/posts/local-nix-no-root/

wmertens:
I'm going to close this issue, because from the thread it's obvious you can use sqlite over NFS, although the macOS implementation might cause issues.

So it's basically a niche not-quite-a-problem.

veprbl:
Would be nice to introduce this as an option instead of applying patches as suggested in https://nixos.wiki/wiki/Nix_Installation_Guide#NFS .

wmertens:
Now that I see how small the patch and config is, I think it does warrant an option. Perhaps the Nix option `use-sqlite-wal = false` can also open sqlite with unix-dotfile?

Reopening.

PR #5475, opened 2021-11-03
libstore: Use unix-dotfile vfs if useSQLiteWAL is false

doronbehar:
Fixes https://github.com/NixOS/nix/issues/2357 . cc @wmertens @veprbl 

I don't know how to test this, since I don't know how to reproduce the sqlite db corruption issue mentioned in https://nixos.wiki/wiki/Nix_Installation_Guide#NFS . I also rarely write C code so this is probably wrong.

I also hope I got the logic correct.

### DIFF
--- a/src/libstore/sqlite.cc
+++ b/src/libstore/sqlite.cc
@@ -1,4 +1,5 @@
 #include "sqlite.hh"
+#include "globals.hh"
 #include "util.hh"
 
 #include <sqlite3.h>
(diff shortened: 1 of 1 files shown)


wmertens:
Well, LGTM in any case :) I think the corruption is pretty hard to do but involves 2 writers, perhaps one on NFS and the other on the filesystem.

doronbehar:
Thanks for the approval! Backports: #5490 #5491 .

tomberek:
This commit makes me think there may be some interesting situations where one can query or use [sqlite via the HTTP VFS](https://github.com/psanford/sqlite3vfshttp). This can allow interesting or bulk queries to be run efficiently over a sqlite uploaded to S3/static hosting with Range requests.

PR #5477, opened 2021-11-03
In checkOverlay, accept underscored names for final/prev args.

### LINKED ISSUE #4416: `nix flake check`: allow `_` as arguments to overlays

I am creating a flake which has some overlays. One of that overlay is like this:

```nix
overlays = {
  inputs = _: _: { inherit inputs; };
};
```

`nix flake check` complaints that this overlay has neither the `final` nor the `prev` arguments:

```plain
$ nix flake check --show-trace
warning: Git tree '/home/nmelzer/.config/nixpkgs' is dirty
error: --- Error ------------------------------------------------------------------------------ nix
overlay does not take an argument named 'final'
------------------------------------------- show-trace --------------------------------------------
trace: while checking the overlay 'overlays.inputs'
at: (26:9) in file: /nix/store/637sw06gkn06lj82rqdhj0qaz1izkc2k-source/flake.nix

    25|       overlays = {
    26|         inputs = _: _: { inherit inputs; };
      |         ^
    27|         emacs = emacs.overlay;

trace: while checking flake output 'overlays'
```

If though I name the arguments properly, `nix-linter` will complain:

```plain
$ nix-linter flake.nix
Unused argument `final` at flake.nix:26:18-50
Unused argument `prev` at flake.nix:26:25-50
```

If `check` would allow for the "unnamed" argument `_` or `_final` and `_prev` as appropriate, we could make `nix-linter` and `nix flake check` agree on each others conventions.

league:
I wonder if this would be an acceptable quick fix for a pain point I and some others have with `flake check`. I'd like to be able to run `nix-linter` within `checks`, via `pre-commit-hooks`. But `nix-linter` and `flake check` disagree with each other: the linter forbids unused arguments unless they are ‘unnamed’ (`_`) or prefixed with `_`, but check requires exactly the names `final` and `prev` for overlay arguments, even if they are unused.

This just adds the capability to use `_final`/`_prev` in addition to `final`/`prev`, respectively.

Resolves #4416, see also #5146.

### DIFF
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -346,10 +346,14 @@ struct CmdFlakeCheck : FlakeCommand
         auto checkOverlay = [&](const std::string & attrPath, Value & v, const Pos & pos) {
             try {
                 state->forceValue(v, pos);
-                if (!v.isLambda() || v.lambda.fun->hasFormals() || std::string(v.lambda.fun->arg) != "final")
+                if (!v.isLambda() || v.lambda.fun->hasFormals() ||
+                    (std::string(v.lambda.fun->arg) != "final" &&
+                        std::string(v.lambda.fun->arg) != "_final"))
                     throw Error("overlay does not take an argument named 'final'");
                 auto body = dynamic_cast<ExprLambda *>(v.lambda.fun->body);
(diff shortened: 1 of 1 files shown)


Sahabee:
🤗

PR #5366, opened 2021-10-09
ca-specific-schema.sql: add index on RealisationsRefs(referrer)

trofi:
For a typical desktop system (~2K packages) we can easily get 100K
entries in RealisationsRefs. Without indices query for RealisationsRefs
requires linear scan.

Inefficiency is seen as a 100% CPU load of nix-daemon for the following
scenario:

    $ nix edit -f . bash # add unused environment variable, like FOO="1"
    # populate RealisationsRefs, build fresh system
    $ nix build -f nixos system --arg config '{ contentAddressedByDefault = true; }'
    $ nix edit -f . bash # add unused environment variable, like FOO="2"
    $ time nix build -f nixos system --arg config '{ contentAddressedByDefault = true; }'

In this case `bash `will be rebuilt a few times and then rest of CPU
time is spent on scanning RealisationsRefs table (about 5 CPU-minutes
on my machine).

Before the change:

    $ time nix build -f nixos system ... # step 4 above
    real    34m3,613s
    user    0m5,232s
    sys     0m0,758s

Of all this time about 29.5 minutes are taken by nix-daemon's CPU time.

After the change:

    $ time nix build -f nixos system ... # step 4 above
    real    4m50,061s
    user    0m5,038s
    sys     0m0,677s

Of all this time about 1 minute is taken by nix-daemon's CPU time.
Most of the time is spent polling for non-existent realisations on
cache-nixos.org.

PR #5479, opened 2021-11-03
Fix the removal of ca-induced self-references

### LINKED ISSUE #5320: `nix store gc` fails to delete self-referred path: executing SQLite statement 'delete from ValidPaths where path = '...';': constraint failed

thufschmitt:
If the derivation `foo` depends on `bar`, and they both have the same
output path (because they are CA derivations), then this output path
will depend both on the realisation of `foo` and of `bar`, which
themselves depend on each other.
This confuses SQLite which isn’t able to automatically solve this
diamond dependency scheme.

Help it by adding a trigger to delete all the references between the
relevant realisations.

Fix #5320

### DIFF
--- a/src/libstore/ca-specific-schema.sql
+++ b/src/libstore/ca-specific-schema.sql
@@ -13,6 +13,19 @@ create table if not exists Realisations (
 
 create index if not exists IndexRealisations on Realisations(drvPath, outputName);
 
+-- We can end-up in a weird edge-case where a path depends on itself because
+-- it’s an output of a CA derivation, that happens to be the same as one of its
+-- dependencies.
+-- In that case we have a dependency loop (path -> realisation1 -> realisation2
+-- -> path) that we need to break by removing the dependencies between the
+-- realisations
+create trigger if not exists DeleteSelfRefsViaRealisations before delete on ValidPaths
+  begin
+    delete from RealisationsRefs where realisationReference = (
+      select id from Realisations where outputPath = old.id
+    );
+  end;
+
 create table if not exists RealisationsRefs (
     referrer integer not null,
     realisationReference integer,
--- a/src/libstore/ca-specific-schema.sql
+++ b/src/libstore/ca-specific-schema.sql
@@ -21,7 +21,7 @@ create index if not exists IndexRealisations on Realisations(drvPath, outputName
 -- realisations
 create trigger if not exists DeleteSelfRefsViaRealisations before delete on ValidPaths
   begin
-    delete from RealisationsRefs where realisationReference = (
+    delete from RealisationsRefs where realisationReference in (
       select id from Realisations where outputPath = old.id
     );
   end;
--- a/src/libstore/ca-specific-schema.sql
+++ b/src/libstore/ca-specific-schema.sql
@@ -32,6 +32,8 @@ create table if not exists RealisationsRefs (
     foreign key (referrer) references Realisations(id) on delete cascade,
     foreign key (realisationReference) references Realisations(id) on delete restrict
 );
+-- used by deletion trigger
+create index if not exists IndexRealisationsRefsRealisationReference on RealisationsRefs(realisationReference);
 
 -- used by QueryRealisationReferences
 create index if not exists IndexRealisationsRefs on RealisationsRefs(referrer);
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -482,18 +482,18 @@ void LocalStore::openDB(State & state, bool create)
         SQLiteStmt stmt;
         stmt.create(db, "pragma main.journal_mode;");
         if (sqlite3_step(stmt) != SQLITE_ROW)
-            throwSQLiteError(db, "querying journal mode");
+            SQLiteError::throw_(db, "querying journal mode");
         prevMode = std::string((const char *) sqlite3_column_text(stmt, 0));
     }
     if (prevMode != mode &&
         sqlite3_exec(db, ("pragma main.journal_mode = " + mode + ";").c_str(), 0, 0, 0) != SQLITE_OK)
-        throwSQLiteError(db, "setting journal mode");
+        SQLiteError::throw_(db, "setting journal mode");
 
     /* Increase the auto-checkpoint interval to 40000 pages.  This
        seems enough to ensure that instantiating the NixOS system
        derivation is done in a single fsync(). */
     if (mode == "wal" && sqlite3_exec(db, "pragma wal_autocheckpoint = 40000;", 0, 0, 0) != SQLITE_OK)
-        throwSQLiteError(db, "setting autocheckpoint interval");
+        SQLiteError::throw_(db, "setting autocheckpoint interval");
 
     /* Initialise the database schema, if necessary. */
     if (create) {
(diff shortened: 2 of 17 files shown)


src/libstore/local-store.cc
```diff
@@ -130,6 +130,19 @@ void migrateCASchema(SQLite& db, Path schemaPath, AutoCloseFD& lockFd)
             txn.commit();
         }
 
+        if (curCASchema < 3) {
```
cole-h, line 133: Not totally related to this PR, but I wonder if you could take inspiration from the Hydra db migration setup instead of bloating this file with potentially loads of migrations.

Specifically, I'd imagine there would be a directory full of `.sql` schemas alongside the `ca-specific-schema.sql` and this function would instead just sequentially run all of the transactions that haven't happened yet, according to the current schema version.

(Maybe this was talked about in a past PR / issue / the RFC; if so, sorry)

regnat: Indeed, that could be much nicer. I just blindly copied what already existed for the standard schema :D

cole-h: I created a new issue so that others can chime in with their thoughts / feelings (specifically, Eelco and other source stakeholders): https://github.com/NixOS/nix/issues/5482

Sahabee (commented):
_****_

src/libstore/ca-specific-schema.sql
```diff
@@ -13,6 +13,19 @@ create table if not exists Realisations (
...
 
+-- We can end-up in a weird edge-case where a path depends on itself because
+-- it’s an output of a CA derivation, that happens to be the same as one of its
+-- dependencies.
+-- In that case we have a dependency loop (path -> realisation1 -> realisation2
+-- -> path) that we need to break by removing the dependencies between the
+-- realisations
+create trigger if not exists DeleteSelfRefsViaRealisations before delete on ValidPaths
+  begin
+    delete from RealisationsRefs where realisationReference = (
```
trofi, line 24: I gave it a go on real system and it fails to delete entries still. I suggest the following change:

`delete from RealisationsRefs where realisationReference = ( ...` -> `delete from RealisationsRefs where realisationReference IN (...`. I think we expect multiple realizations to point to the same store path worst case. Not just one.

With `IN` I was able to collect all the garbage on my system \o/.

src/libstore/local-store.cc
```diff
@@ -130,6 +130,19 @@ void migrateCASchema(SQLite& db, Path schemaPath, AutoCloseFD& lockFd)
...
+            SQLiteTxn txn(db);
+            db.exec(R"(
+                create trigger if not exists DeleteSelfRefsViaRealisations before delete on ValidPaths
+                begin
+                    delete from RealisationsRefs where realisationReference = (
```
trofi, line 138: Same here: `=` -> `IN`.

trofi (changes requested):

src/libstore/ca-specific-schema.sql
trofi, line 24: I also suggest adding `CREATE INDEX IndexRealisationsRefsRealisationReference on RealisationsRefs(realisationReference);` index.

Before the index we linearly scan all the RealisationRefs:

```
sqlite>.eqp full
sqlite> delete from RealisationsRefs where realisationReference IN ( select id from Realisations where outputPath = 1234567890 );
QUERY PLAN
|--SCAN RealisationsRefs
`--LIST SUBQUERY 1
   `--SEARCH Realisations USING COVERING INDEX IndexRealisationsRefsOnOutputPath (outputPath=?)
```

After the index bots are index lookups:

```
sqlite> CREATE INDEX IndexRealisationsRefsRealisationReference on RealisationsRefs(realisationReference);
sqlite> delete from RealisationsRefs where realisationReference IN ( select id from Realisations where outputPath = 1234567890 );
QUERY PLAN
|--SEARCH RealisationsRefs USING INDEX IndexRealisationsRefsRealisationReference (realisationReference=?)
`--LIST SUBQUERY 1
   `--SEARCH Realisations USING COVERING INDEX IndexRealisationsRefsOnOutputPath (outputPath=?)
```

Sahabee: I m larner

Ericson2314:
I'm a bit confused, is this just temporarily deleting some edges, such that the DB "at rest" shouldn't be different? Or is this permently deleting some edges?

trofi:
> I'm a bit confused, is this just temporarily deleting some edges, such that the DB "at rest" shouldn't be different? Or is this permently deleting some edges?

I think this change does neither. The "self-reference-looking" edges are still stored in database and kept being created. Now they can be deleted when path is deleted by `nix-collect-garbage` instead of failing with foreign key constraint violation.

regnat:
> is this just temporarily deleting some edges, such that the DB "at rest" shouldn't be different?

Yes, this is just “forcing” the order in which the related references are removed when we remove a derivation output, because SQLite sees a circular dependency with some `ON DELETE RESTRICT` in the middle, so it can’t do the removal.
In essence this is the same thing as https://github.com/NixOS/nix/blob/bceda304982a34c65d5a1dab449cb5bc59f63b83/src/libstore/schema.sql#L29-L32, just slightly more involved

Ericson2314:
OK whew thanks for confirming :).

regnat:
Rebased on top of #5366 . Should be good to go as soon as #5366 is merged

zseri:
from the build log:
```
2021-12-14T08:58:34.5628351Z     deleting garbage...
2021-12-14T08:58:34.5629635Z     deleting '/build/nix-test/tests/ca/selfref-gc/store/0hmsy2dxg652l7iidg82civ38q2fmsz7-selfref-gc.drv'
2021-12-14T08:58:34.5632185Z     error: executing SQLite statement 'delete from ValidPaths where path = '/build/nix-test/tests/ca/selfref-gc/store/dwxdqfjaym0shbk7iwcnqxdx9z49acz3-selfref-gc';': constraint failed (in '/build/nix-test/tests/ca/selfref-gc/var/nix/db/db.sqlite')
2021-12-14T08:58:34.5633827Z     0 store paths deleted, 0.00 MiB freed
```
doesn't seem like it reliably works.

regnat:
> Mh, must be an issue with the rebase. I’ll have a look

Ah of course, was the `requireDaemonNewerThan` call which wasn’t right anymore. Should be good now :crossed_fingers:

thufschmitt:
@edolstra anything I should do to get this merged?

trofi:
I'd suggest adding indices: https://github.com/NixOS/nix/pull/5479#pullrequestreview-797014560

src/libstore/ca-specific-schema.sql
trofi: Proposed as https://github.com/NixOS/nix/pull/6324 into this branch.

trofi (approved):

Ericson2314:
@thufschmitt when I was working on another unrelated change I noticed we can currently have realization->realization self-references (because we don't filter out self-references when write realization deps). Would that help with this?

PR #5486, opened 2021-11-03
Fix leaking pthread_attr_t

mohe2015:
pthread_attr_destroy was not called.k

Didn't verify that the patch format is 100% valid.


boehmgc-coroutine-sp-fallback.diff
```diff
@@ -22,6 +22,7 @@ index 1cee6a0b..46c3acd9 100644
 +          if (pthread_attr_getstacksize(&pattr, &stack_limit)) {
 +            ABORT("GC_push_all_stacks: pthread_attr_getstacksize failed!");
 +          }
++          pthread_attr_destroy(&pattr);
```
mohe2015, line 25: Ohh it's C - error checking...

mohe2015:
Don't know if that patch is correctly regenerated now - checked out the git at the tag and applied and diffed.

PR #5492, opened 2021-11-04
Nix 2.3 Backport: PathSubstitutionGoal: Clean up pipe

fendor:
If there were many top-level goals (which are not destroyed until the
very end), commands like

```
  $ nix copy --to 'ssh://localhost?remote-store=/tmp/nix' \
    /run/current-system --no-check-sigs --substitute-on-destination
```
could fail with "Too many open files". So now we do some explicit
cleanup from amDone(). It would be cleaner to separate goals from
their temporary internal state, but that would be a bigger refactor.

Backport 8a29052cb2f52ef2c82c36fb3818fd0f66349729

### DIFF
--- a/src/libstore/build.cc
+++ b/src/libstore/build.cc
@@ -179,6 +179,8 @@ public:
 
     virtual string key() = 0;
 
+    virtual void cleanup() { }
+
 protected:
 
     virtual void amDone(ExitCode result);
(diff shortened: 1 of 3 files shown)


fendor:
cc @Ericson2314, @ElvishJerricco

Ericson2314 (approved):

PR #5493, opened 2021-11-04
docs: Correct fallback user config path

jtojnar:
This is in line with XDG Base Directory Specification, where ~/.config is supposed to be used when XDG_CONFIG_HOME is unset.

It also better matches the reality, where ~/.config/nix.conf does not seem to be used.

I tested it by unsetting `XDG_CONFIG_HOME` variable and adding `experimental-features = foo` to `~/.config/nix/nix.conf`. It did have the effect of overriding my global `experimental-features = nix-command`. Unlike `~/.config/nix.conf`.

### DIFF
--- a/doc/manual/src/command-ref/conf-file-prefix.md
+++ b/doc/manual/src/command-ref/conf-file-prefix.md
@@ -17,7 +17,7 @@ By default Nix reads settings from the following places:
 
     Otherwise it will look for `nix/nix.conf` files in `XDG_CONFIG_DIRS`
     and `XDG_CONFIG_HOME`. If these are unset, it will look in
-    `$HOME/.config/nix.conf`.
+    `$HOME/.config/nix/nix.conf`.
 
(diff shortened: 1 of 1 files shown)


doc/manual/src/command-ref/conf-file-prefix.md
```diff
@@ -17,7 +17,7 @@ By default Nix reads settings from the following places:
 
     Otherwise it will look for `nix/nix.conf` files in `XDG_CONFIG_DIRS`
```
jtojnar, line 18: This does not actually fully match the [basedir-spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

> If $XDG_CONFIG_DIRS is either not set or empty, a value equal to /etc/xdg should be used.

We would need `/etc/xdg` as a fallback in

https://github.com/NixOS/nix/blob/c5fd0b46ae56447d9d0a2028d464a5bbf753bf55/src/libutil/util.cc#L565

PR #5494, opened 2021-11-04
Allow references in addPath

### LINKED ISSUE #5410: filterSource: error: store path <PATH> is not allowed to have references

balsoft:
Allows `addPath` on paths with references, therefore allowing `builtins.path` and `builtins.filterSource` to work on paths with references.

Closes #5410

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1859,12 +1859,12 @@ static void addPath(
         // be rewritten to the actual output).
         state.realiseContext(context);
 
+        StorePathSet refs = StorePathSet();
+
         if (state.store->isInStore(path)) {
             auto [storePath, subPath] = state.store->toStorePath(path);
-            auto info = state.store->queryPathInfo(storePath);
-            if (!info->references.empty())
-                throw EvalError("store path '%s' is not allowed to have references",
-                    state.store->printStorePath(storePath));
+            // FIXME: we should scanForReferences on the path before adding it
+            refs = state.store->queryPathInfo(storePath)->references;
             path = state.store->toRealPath(storePath) + subPath;
         }
 
@@ -1904,7 +1904,7 @@ static void addPath(
         if (!expectedHash || !state.store->isValidPath(*expectedStorePath)) {
             dstPath = state.store->printStorePath(settings.readOnlyMode
                 ? state.store->computeStorePathForPath(name, path, method, htSHA256, filter).first
-                : state.store->addToStore(name, path, method, htSHA256, filter, state.repair));
+                : state.store->addToStore(name, path, method, htSHA256, filter, state.repair, refs));
             if (expectedHash && expectedStorePath != state.store->parseStorePath(dstPath))
                 throw Error("store path mismatch in (possibly filtered) path added from '%s'", path);
         } else
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1859,7 +1859,7 @@ static void addPath(
         // be rewritten to the actual output).
         state.realiseContext(context);
 
-        StorePathSet refs = StorePathSet();
+        StorePathSet refs;
 
         if (state.store->isInStore(path)) {
             auto [storePath, subPath] = state.store->toStorePath(path);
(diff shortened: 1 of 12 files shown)


balsoft:
CI fail appears to be unrelated

src/libexpr/primops.cc
```diff
@@ -1859,12 +1859,12 @@ static void addPath(
...
 
+        StorePathSet refs = StorePathSet();
+
         if (state.store->isInStore(path)) {
             auto [storePath, subPath] = state.store->toStorePath(path);
-            auto info = state.store->queryPathInfo(storePath);
-            if (!info->references.empty())
-                throw EvalError("store path '%s' is not allowed to have references",
-                    state.store->printStorePath(storePath));
+            // FIXME: we should scanForReferences on the path before adding it
```
Ericson2314, line 1866: This is just conservatively taking all the references, even if they happened to be filtered out? That seems decent.

balsoft: Yes. This may result in unneeded references, but I think this bug is rare enough that it doesn't matter, at least for now. FIXME is there in case someone stumbles on this in the future and wants to fix it :)

balsoft:
I've been using this for the past week or so with no issues. Builds crate2nix stuff just fine.

src/libexpr/primops.cc
```diff
@@ -1859,12 +1859,12 @@ static void addPath(
         // be rewritten to the actual output).
         state.realiseContext(context);
 
+        StorePathSet refs = StorePathSet();
```
edolstra, line 1862:
```suggestion
        StorePathSet refs;
```

src/libstore/binary-cache-store.cc
```diff
@@ -308,16 +308,17 @@ void BinaryCacheStore::addToStore(const ValidPathInfo & info, Source & narSource
 }
 
 StorePath BinaryCacheStore::addToStoreFromDump(Source & dump, const string & name,
-    FileIngestionMethod method, HashType hashAlgo, RepairFlag repair)
+    FileIngestionMethod method, HashType hashAlgo, RepairFlag repair, StorePathSet references)
```
edolstra, line 311:
```suggestion
    FileIngestionMethod method, HashType hashAlgo, RepairFlag repair, const StorePathSet & references)
```

and everywhere else.

edolstra (approved):

src/libexpr/primops.cc
edolstra: It's not ideal since it allows untrusted users to add paths with incorrect references to the store. But then that's already the case with `addTextToStore()`...

balsoft: > It's not ideal since it allows untrusted users to add paths with incorrect references to the store

Hm, I'm not sure I see how this is possible? The references here must come from a path that's already in store, and therefore has been `scanForReferences` once already, it's just that the user may filter it so some references are not actually present in the resulting path while being part of path info as far as nix is concerned.

edolstra: The issue isn't whether the references are correct here, but that `RemoteStore::addToStoreFromDump` trusts whatever the client passes in. So you could write an untrusted client that adds paths with incorrect references.

Issue #5312, opened 2021-09-30
Support self-fetching parameters in flake.nix


nrdxp:
#5284 happened, so instead I'll open a proper issue to address what I feel would be the "correct" solution since I am incapable of implementing it atm.

I see no particular reason why `self` should be _especially handicapped_ from the other inputs, but should be able to be easily handled as a regular `git` type input, or whichever active VCS Nix can detect.

you wouldn't even have to change the syntax, but simply allow `inputs.self.submodules = true`, for example.

Nix would simply have to know that `inputs.self` is the current `self`.

edolstra:
There is a very good reason why `self` is different: we have to fetch the `self` flake before we can read its `flake.nix`. So any attributes like `submodules` have to be known in advance.

nrdxp:
There is no possible mechanism to read the flake inputs before `self` is fetched? Like maybe just pulling the _lock file_ into the Nix store for reading before the rest of the flake?

mstone:
@nrdxp — I’ve been thinking about this a bit because it’s going to bite everyone who tries to flakeify upstream projects that use submodules, like, in my case, qemu.

So far the best idea I’ve come up with is to write the flake to use an auxiliary input, in addition to `self`, that fetches the same underlying repo just with submodules enabled.

Unfortunately, that still leaves several quality-of-life problems that I wonder if we can fix:

a) that the self input is fetched but is unused (other than for obtaining flake.nix/flake.lock)

b) that the self-with-submodules input is “locked” when we’d rather that it be “derived from” or “tied” to self — e.g. so that we see dirty working copy edits / whatever the flake-ref resolves to when being built, and don’t have to remember to push commits that bump flake.lock just to test changes.

(P.S. — if you’re interested in this approach, I wrote more at https://gist.github.com/mstone/4218e6fa9ef98e5153f543fb8e7b4743 with several variations in this idea, though I’m not really sure that any of them are ready for prime time. :-/)

Kha:
> There is a very good reason why `self` is different: we have to fetch the `self` flake before we can read its `flake.nix`. So any attributes like `submodules` have to be known in advance.

Would this still be the case with #3121? Afaiu, in that case we would evaluate `flake.nix` in the local checkout/git cache and only then copy `self` to the store.

edolstra:
#3121 only addresses the case of a local git tree.

[...]

PR #5399, opened 2021-10-17
Optimize submodules fetching

Kha:
Replace the current naive submodule fetching code with one that mostly does the minimal amount of necessary work/fetching
* For remote repos, we clone into `.cache/nix` as usual, then `checkout` & `submodule update` into the temp directory using `--work-tree`. It doesn't get much more direct than that.
* For local repos, we shouldn't do the same because we don't want to change the `HEAD` of that repo (and there is no direct way to check out submodules for a specific commit that is not `HEAD`). Instead, we do a cheap clone into the temp dir that shares all objects, including those of submodules, with the original repo. Note that if the submodule commit has *not* yet been fetched into the original repo, `submodule update` will then fetch it into the temp dir, so checking out another commit afterwards will fetch the upstream again. This could potentially be improved by fetching into `.cache/nix` even for local repos.

Some measurements:

Store-copying (prefetched) Nixpkgs without submodules/with submodules (old)/with submodules (new)
```
$ nix store delete /nix/store/jq0sw73dh69bgm588277jvxzg881vjmq-source && sync && time nix eval --expr 'builtins.fetchGit { url = "https://github.com/NixOS/nixpkgs"; rev = "de084e5de9a4279320f902c3b92df34c47798fdd"; }'
1 store paths deleted, 91.36 MiB freed
{ lastModified = 1633782067; lastModifiedDate = "20211009122107"; narHash = "sha256-GGWh0sqDoll1JSqRri7CAm5dWfhtTMt3OG0arGF+Lxw="; outPath = "/nix/store/jq0sw73dh69bgm588277jvxzg881vjmq-source"; rev = "de084e5de9a4279320f902c3b92df34c47798fdd"; revCount = 322195; shortRev = "de084e5"; submodules = false; }
nix eval --expr   3.35s user 1.08s system 89% cpu 4.954 total
$ /run/current-system/sw/bin/nix store delete /nix/store/jq0sw73dh69bgm588277jvxzg881vjmq-source && sync && time nix eval --expr 'builtins.fetchGit { url = "https://github.com/NixOS/nixpkgs"; rev = "de084e5de9a4279320f902c3b92df34c47798fdd"; submodules = true; }'
1 store paths deleted, 91.36 MiB freed
[...]

PR #5497, opened 2021-11-04
A different take on git submodule support for flakes

kaii-zen:
As the title implies, this is a whole 'nother take. Ça veut dire, it does not interact **AT ALL** with `submodules = true;`. It does involve more code though.

## Idea
If a tree has a `.gitmodules` in it, then we:
- Get the paths where submodules go
- Get the respective URLs that go with those paths
- Traverse the git tree to work out the `rev`s
- Construct flake-friendly URLs (`git+xxxx://host/who/what?rev=xxxx` etc)
- Build an attrset that maps paths to flake-friendly URLs
- Slap this into the attrset returned by `fetchTree`

That's it. It looks like this:

```console
❯ nix eval --impure --expr 'builtins.fetchTree git+https://github.com/runtimeverification/iele-semantics' | nixfmt
{
  lastModified = 1636045922;
  lastModifiedDate = "20211104171202";
  modules = {
    "deps/secp256k1" =
      "git+https://github.com/bitcoin-core/secp256k1?allRefs=1&rev=f532bdc9f77f7bbf7e93faabfbe9c483f0a9f75f";
    plugin =
      "git+https://github.com/runtimeverification/blockchain-k-plugin?allRefs=1&rev=cc7384e565e4c8df4d17a3330cd9951d32d4830f";
    "tests/ethereum-tests" =
      "git+https://github.com/ethereum/tests.git?allRefs=1&rev=7057d0e021f7dbee37a2ddb6acf34f4946bc157d";
    "web/k-web-theme" =
      "git+https://github.com/runtimeverification/k-web-theme?allRefs=1&rev=189fe6650488f8f04dc89d25a61cdfd888cd9dd8";
  };
  narHash = "sha256-9GDWDomnRyJ47mDbygoBHThq9Q+3JGZDYTIHB8SsMlc=";
  outPath = "/nix/store/6d1w3h7fygfwl1f42x1bx6xhsrsmh7qj-source";
  rev = "431a4a44dad39aedb1dff126475cdce3b8d91e83";
  revCount = 1399;
  shortRev = "431a4a4";
  submodules = false;
}
```

It's just plain contextless strings so unless I'm fundamentally misunderstanding things, no dragons lie therein.
From here the user is empowered to do as much or as little as she wants with it: use `builtins.getFlake` for a flake, `builtins.fetchTree` for a non-flake, or ignore it altogether.

See `tests/flakes-with-submodules.sh` for a (rudimentary) example of how this might be used within a flake.

This could potentially be used as a basis for a more "flakeful" UX, where submodules behave like `indirect` (registry) inputs and get populated with the Right Thing if and only if they get added to the `outputs` function args set... or not.

Since my C++ is rusty and my familiarity with the Nix codebase is flakey (both puns intended btw 🙃), I invite y'all to punch holes in my reasoning and point out any blind spots so that we can address them and hopefully reach a solution that everyone's happy with.

I trust that this would save the sanity of at least a handful of Nixers/Nixresses out there who are forced to interact with git submodules against their better judgement (or in accordance with their poor judgement. to each their own 🤷‍♀️), thereby empowering them to once again pursue their (title-cased) Hopes and Dreams!

kk 'nuff prose. known issues/caveats:
- this does add the submodule data to the cache, so a `rm -rf ~/.cache/nix` is required (otherwise you'd get `error: input attribute 'modules' is missing`)
- ~~dirty submodule scenario isn't handled yet~~ added, and fixed so that it also works without `__contentAddressed = true;` (thanks @shlevy!)
- does not solve the "I want `self` to have `submodules = true;`" use case (https://github.com/NixOS/nix/issues/5312)
- might segfault, murder your kittens or both

Shout out to @cleverca22 for essentially onboarding me to the codebase (... and writing most of the harder parts)

Cheers

### DIFF
--- a/src/libfetchers/fetchers.cc
+++ b/src/libfetchers/fetchers.cc
@@ -257,6 +257,13 @@ std::optional<time_t> Input::getLastModified() const
     return {};
 }
 
+std::optional<std::string> Input::getModules() const
+{
+    if (auto s = maybeGetStrAttr(attrs, "modules"))
+        return *s;
+    return {};
+}
+
 ParsedURL InputScheme::toURL(const Input & input)
 {
     throw Error("don't know how to convert input '%s' to a URL", attrsToJSON(input.attrs));
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -24,15 +24,99 @@ static std::string readHead(const Path & path)
     return chomp(runProgram("git", true, { "-C", path, "rev-parse", "--abbrev-ref", "HEAD" }));
 }
 
-static Attrs readSubmodules(const Path & path)
+static string getRootTree(const Path & path, const string & gitrev) {
+    auto commitOutput = runProgram("git", true, { "-C", path, "cat-file", "-p", gitrev });
+    auto lines = tokenizeString<std::vector<string>>(commitOutput, "\n");
+    auto words = tokenizeString<std::vector<string>>(lines[0]);
+    auto rootTree = words[1];
+    return rootTree;
+}
+
+enum gitType {
+  blob,
+  tree,
+  commit
+};
+
+string getBlob(const Path & path, const string & gitrev, const string & blobhash) {
+  return runProgram("git", true, { "-C", path, "cat-file", "-p", blobhash });
+}
+
+static std::map<string, std::pair<gitType, string>> gitTreeList(const Path & path, const string & treehash) {
+  auto treeOutput = runProgram("git", true, { "-C", path, "cat-file", "-p", treehash });
+  auto lines = tokenizeString<std::vector<string>>(treeOutput, "\n");
+  std::map<string, std::pair<gitType, string>> results;
+  for (auto line : lines) {
+    auto words = tokenizeString<std::vector<string>>(line);
+    auto hash = words[2];
+    auto name = words[3];
+    gitType type;
+    if (words[1] == "blob") type = blob;
+    else if (words[1] == "tree") type = tree;
+    else if (words[1] == "commit") type = commit;
+    else assert(0); // FIXME
+    results.insert(std::pair{name, std::pair{type, hash}});
+  }
+
+  return results;
+}
+
+std::map<string,string> parseSubmodules(const string & contents) {
+  auto lines = tokenizeString<std::vector<string>>(contents, "\n");
+  std::map<string,string> results;
+  string path;
+  for (auto line : lines) {
+    auto words = tokenizeString<std::vector<string>>(line);
+    if (words[1] != "=") continue;
+
+    if (words[0] == "path") path = words[2];
+    if (words[0] == "url") {
+      string url = words[2];
+      results.insert(std::pair{path, url});
+    }
+  }
+  return results;
+}
+
+string findCommitHash(const Path & path, const std::map<string, std::pair<gitType, string>> & _entries, const string & pathToFind) {
+  auto entries = _entries;
+
+  auto tokens = tokenizeString<std::vector<string>>(pathToFind, "/");
+  std::pair<gitType, string> x;
+  for (auto token : tokens) {
+    auto i = entries.find(token);
+    if (i == entries.end()) return "fail";
+
+    x = i->second;
+    if (x.first == tree) entries = gitTreeList(path, x.second);
+  }
+  if (x.first != commit) return "fail";
+
+  return x.second;
+}
+
+static Attrs readSubmodules(const Path & path, const string & gitrev)
 {
-    auto output = runProgram("git", true, { "-C", path, "submodule", "status" });
+    auto rootTree = getRootTree(path, gitrev);
+    printTalkative("root tree is %s", rootTree);
+    auto entries = gitTreeList(path, rootTree);
     Attrs attrs;
-    for (auto line : tokenizeString<std::vector<string>>(output, "\n;")) {
-        auto tokens = tokenizeString<std::vector<string>>(line);
-        auto url = tokens[0];
-        auto path = tokens[1];
-        attrs.emplace(path, url);
+
+    auto i = entries.find(".gitmodules");
+
+    if (i == entries.end()) return attrs;
+
+    auto submodules = getBlob(path, gitrev, i->second.second);
+
+    printTalkative("submodule file %s", submodules);
+
+    auto parsedModules = parseSubmodules(submodules);
+    for (auto & [path, url] : parsedModules) {
+      printTalkative("submodule %s fetched from %s", path, url);
+      auto commitHash = findCommitHash(path, entries, path);
+      printTalkative("found %s", commitHash);
+
+      attrs.emplace(path, url + "?rev=" + commitHash);
     }
 
     return attrs;
@@ -201,6 +285,8 @@ struct GitInputScheme : InputScheme
         if (submodules) cacheType += "-submodules";
         if (allRefs) cacheType += "-all-refs";
 
+        printTalkative("fetch is getting ran");
+
         auto getImmutableAttrs = [&]()
         {
             return Attrs({
@@ -228,7 +314,8 @@ struct GitInputScheme : InputScheme
         if (input.getRev()) {
             if (auto res = getCache()->lookup(store, getImmutableAttrs())) {
                 Path p = store->printStorePath(res->second);
-                auto modulesInfo = readSubmodules(p);
+                printTalkative("case 1");
+                auto modulesInfo = readSubmodules(p, "FIXME");
                 return makeResult(res->first, std::move(res->second), modulesInfo);
             }
         }
@@ -303,6 +390,8 @@ struct GitInputScheme : InputScheme
                     "lastModified",
                     haveCommits ? std::stoull(runProgram("git", true, { "-C", actualUrl, "log", "-1", "--format=%ct", "--no-show-signature", "HEAD" })) : 0);
 
+                auto modulesInfo = readSubmodules(actualUrl, "HEAD");
+                input.modules = modulesInfo;
                 return {
                     Tree(store->toRealPath(storePath), std::move(storePath)),
                     input
@@ -336,7 +425,8 @@ struct GitInputScheme : InputScheme
                 if (!input.getRev() || input.getRev() == rev2) {
                     input.attrs.insert_or_assign("rev", rev2.gitRev());
                     Path p = store->printStorePath(res->second);
-                    auto modulesInfo = readSubmodules(p);
+                    printTalkative("case 2");
+                    auto modulesInfo = readSubmodules(p, "FIXME");
                     return makeResult(res->first, std::move(res->second), modulesInfo);
                 }
             }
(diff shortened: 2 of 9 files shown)


src/libfetchers/git.cc
```diff
@@ -24,6 +26,110 @@ static std::string readHead(const Path & path)
...
+      auto commitHash = findCommitHash(path, entries, subPath);
+      printTalkative("found %s", commitHash);
+
+      static const std::regex barePathRegex("^/.*$");
+      std::string prefix = "git+";
+      if (std::regex_match(url, barePathRegex))
+          prefix = prefix + "file://";
+
+
+      attrs.emplace(subPath, prefix + url + "?allRefs=1&rev=" + commitHash);
```
Kha, line 127: If the submodule commit is available only in the local clone and not in the remote, we need to pass the former + submodule path as `url` here I think.

Kha (commented):
Very cool. I had something like this in mind as well, but then decided for the simpler approach in #5399 for now.

src/libfetchers/git.cc
Kha: Drop "only" from my sentence to avoid redundant fetches as well

knedlsepp:
Would this be able to handle nested submodules?

kreisys:
> Would this be able to handle nested submodules?

Not as conveniently as it would be with simply enabling submodules for `self`. In my understanding the only way to achieve that is by enabling it across the board, which we cannot because it breaks existing workflows.

The happy use case here is when each of your submodules is a flake itself, handling its own submodules without you having to care about it.

With other use cases, such as requiring the recursively cloned source tree for the build itself, you'd have to write some nix code to manipulate the URL provided into an attrset (until https://github.com/NixOS/nix/pull/5434 adds `&submodule=1` support) to which you'd `// { submodule = true; }` before passing it along to `builtins.fetchTree`. Somewhat ugly but works and can be factored out, tucked away and reused when necessary. IMHO this is still 100% better than the current state of affairs where you're simply out of luck 🤷‍♀️

src/libfetchers/git.cc
```diff
@@ -24,6 +26,110 @@ static std::string readHead(const Path & path)
...
+  if (x.first != commit) return "fail";
+
+  return x.second;
+}
+
+static Attrs readSubmodules(const Path & path, const string & gitrev)
+{
+    auto rootTree = getRootTree(path, gitrev);
+    printTalkative("root tree is %s", rootTree);
+    auto entries = gitTreeList(path, rootTree);
```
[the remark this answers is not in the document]
kreisys: Idk
/cc @cleverca22

src/libfetchers/git.cc
```diff
@@ -24,6 +26,110 @@ static std::string readHead(const Path & path)
...
+    else if (words[1] == "tree") type = tree;
+    else if (words[1] == "commit") type = commit;
+    else assert(0); // FIXME
+    results.insert(std::pair{name, std::pair{type, hash}});
+  }
+
+  return results;
+}
+
+std::map<string,string> parseSubmodules(const string & contents) {
```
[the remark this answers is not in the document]
kreisys: That's a very neat trick thank you!

src/libfetchers/git.cc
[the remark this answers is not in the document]
cleverca22: looks like you can just skip directly to `ls-tree` on a rev

but you still need to recurse manually, `git ls-tree $rev $path/` for each dir

src/libfetchers/git.cc
```diff
@@ -24,6 +26,110 @@ static std::string readHead(const Path & path)
...
+    if (words[0] == "path") path = words[2];
+    if (words[0] == "url") {
+      string url = words[2];
+      results.insert(std::pair{path, url});
+    }
+  }
+  return results;
+}
+
+string findCommitHash(const Path & path, const std::map<string, std::pair<gitType, string>> & _entries, const string & pathToFind) {
```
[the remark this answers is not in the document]
cleverca22: ahhh, and that solves the recursion i mentioned a few comments up

tests/flakes-with-submodules.sh
```diff
@@ -0,0 +1,119 @@
+source common.sh
+
+if [[ -z $(type -p git) ]]; then
+    echo "Git not installed; skipping flake tests"
+    exit 99
+fi
+
+clearStore
+rm -rf $TEST_HOME/.{cache,config}
```
shlevy, line 9: There's `clearCache`... Why do you need to nuke `config`?

src/libfetchers/fetchers.cc
```diff
@@ -257,6 +257,13 @@ std::optional<time_t> Input::getLastModified() const
     return {};
 }
 
+std::optional<std::string> Input::getModules() const
+{
+    if (auto s = maybeGetStrAttr(attrs, "modules"))
```
shlevy, line 262: I'm surprised that this works... Isn't modules an attrset? Doesn't `maybeGetStrAttr` throw if the attribute isn't a string?

kreisys: It's already stored as JSON. `fetchTree` reconstructs it into an attrset here: https://github.com/NixOS/nix/pull/5497/files#diff-e473b01a82cc31e51baa89b26b4fe5083e56df27cd834b9722789841ef9d5620R38-R50

not the prettiest thing in the world but it works 🤷‍♀️

shlevy: :scream: 

Does the process producing the JSON sort keys?

kreisys: oops, you're right. we only sort in `fetchTree` 😅

kreisys: Hmmm... looks like nlohmann takes care of it, unless I'm missing something: https://json.nlohmann.me/features/object_order/

FlorianFranzen:
I rebased this on the latest master and pushed it to my fork under [FlorianFranzen/nix/submodules](https://github.com/FlorianFranzen/nix/tree/submodules). I have started using this instead of loading submodules by default and it safes me a lot of rebuilds. Would love to have `getTree` lazyly execute when I access any modules, otherwise I think it is a great first step in submodule support.

kreisys:
> I rebased this on the latest master and pushed it to my fork under [FlorianFranzen/nix/submodules](https://github.com/FlorianFranzen/nix/tree/submodules). I have started using this instead of loading submodules by default and it safes me a lot of rebuilds. Would love to have `getTree` lazyly execute when I access any modules, otherwise I think it is a great first step in submodule support.

We actually did try to implement that using a thunk at some point but it ended up causing an infinite recursion (IIRC) so was reverted and scoped out... December kinda forced a context switch and I haven't had the time to re-wrap my head around this but I really want to see this through as it's been a recurring pain point in my workflow. Anyways I'm happy it's useful. Help me draw more attention 😎

FlorianFranzen:
@cleverca22 Any chance we could get some input on this?

I will have a look at how to implement a lazy version of this. If necessary I could also add an experimental feature flag for it.

I would also like to see this integrated with `nix flake metadata`.

tests/fetchGit.sh
```diff
@@ -37,6 +37,8 @@ path0_=$(nix eval --impure --raw --expr "(builtins.fetchTree { type = \"git\"; u
 export _NIX_FORCE_HTTP=1
 [[ $(tail -n 1 $path0/hello) = "hello" ]]
 
+ls -l $repo
+nix eval --debug --impure --raw --expr "(builtins.fetchGit file://$repo).outPath"
```
FlorianFranzen, line 41: Are these really needed?

dmadisetti (commented):
Really excited for this change. Please let us know if any additional support is needed- but this looks fine to me

src/libfetchers/git.cc
```diff
@@ -24,6 +27,212 @@ static std::string readHead(const Path & path)
...
+    return rootTree;
+}
+
+enum gitType {
+  blob,
+  tree,
+  commit
+};
+
+string getBlob(const Path & path, const string & gitrev, const string & blobhash) {
```
Smaug123, line 44: Forgive my ignorance, but is there a reason this isn't `static` (given that `getRootTree` is)?

kreisys: Probably because I suck at C++. @cleverca22 do you have a better answer? 😅

tests/flakes-with-submodules.sh
```diff
@@ -0,0 +1,149 @@
...
+
+git -C $flakeDir add .
+git -C $flakeDir commit -m'add flake.nix'
+
+[[ $(nix run $flakeDir#cat-own-readme) == "FNORD" ]]
+
+git -C $flakeWithSubmodules submodule add $nonFlakeDir
+git -C $flakeWithSubmodules submodule add $flakeDir
+git -C $flakeWithSubmodules add .
+git -C $flakeWithSubmodules commit -m'add submodules'
```
Smaug123, line 66: If I were being a Mighty Stickler I'd request a test that declares this feature's behaviour in the presence of recursive submodules. (That is, repo A has a submodule B, which itself has a submodule C.)

src/libfetchers/git.cc
```diff
@@ -24,6 +27,212 @@ static std::string readHead(const Path & path)
...
+  auto commonGitDir = chomp(runProgram(
+      "git", true, {"-C", actualUrl, "rev-parse", "--path-format=absolute", "--git-common-dir"}));
+  if (commonGitDir != ".git")
+    gitDir = commonGitDir;
+
+  return gitDir;
+}
+
+/* Check whether this repo has any commits. There are
+   probably better ways to do this. */
```
Smaug123, line 72: Strictly speaking there's a hilarious but more correct answer:
```
git cat-file --batch-check='%(objecttype)' --batch-all-objects --unordered | grep commit | head -1
```
But I'm sure any sane repo is in fact going to have an entry in `/refs/heads`.

tests/flakes-with-submodules.sh
Smaug123: (Ah, I see you commented above that this isn't really supported. And anyway I wouldn't consider this blocking!)

Smaug123:
Does anyone know any blockers for this going in? And/or who has The Power To Merge?

thegergo02:
Would love to see this implemented for secrets management.

dmadisetti:
@thegergo02  A bit of a hack now for submodule secrets management is to reference a local "fake flake" and then override it with `--override-input secret_flake submodule/path`

Fake flake example: https://github.com/dmadisetti/.dots/blob/template/nix/spoof/flake.nix (I have a bit of templating stuff around it)
Main Flake: https://github.com/dmadisetti/.dots/blob/2c8b22900aa368f285a8eef88d658c2ffc01d48b/flake.nix#L33

Smaug123:
Oh yuck, the merge with master is pretty hairy :(

FlorianFranzen:
I do not mind putting in the work to get this rebased onto master. But I will need some feedback if this is something that actually has a chance of being merged. 

And given the ongoing work on #6530, it probably also makes sense to wait till that is merged.

kreisys:
> I do not mind putting in the work to get this rebased onto master. But I will need some feedback if this is something that actually has a chance of being merged.
> 
> And given the ongoing work on #6530, it probably also makes sense to wait till that is merged.

Makes two of us 🤭

Ericson2314:

- One part might be obsolete given some changes on master. @edolstra will enquire.

- The rest (exposing the submodules at the Nix level) should be discussed

fricklerhandwerk:
Discussed in Nix team meeting 2023-10-13:

- Shouldn't be a blocker for stabilizing `fetchTree`
    - Maybe submodules should be experimental then
- The `lazy-trees` branch has another solution for that (as part of the `libgit` rewrite), could be backported
    - Similar idea overall
    - @edolstra wants to open a first PR for that today
- @edolstra: Prefer the `libgit` approach
- Decision: The tests are great and should be incorporated. The rest is very likely conflicting with the planned `libgit` rewrite.

@FlorianFranzen @kaii-zen @Ericson2314 would one of you be interested in splitting out the tests?

tomberek:

PR #5504, opened 2021-11-05
Make the flake options work when using the daemon

regnat:
When setting flake-local options (with the `nixConfig` field), forward these options to the daemon in case we’re using one.

This is necessary in particular for options like `binary-caches` or `post-build-hook` to make sense.

Fix <https://github.com/NixOS/nix/commit/343239fc8a1993f707a990c2cd54a41f1fa3de99#r44356843>

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -307,7 +307,7 @@ LockedFlake lockFlake(
 
     if (lockFlags.applyNixConfig) {
         flake.config.apply();
-        // FIXME: send new config to the daemon.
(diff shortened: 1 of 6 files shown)


domenkozar:
There's still one piece missing, how does one trust the flags in environments without a stdin?

regnat:
> There's still one piece missing, how does one trust the flags in environments without a stdin?

I was actually very frustrated because I knew there was an issue I was meaning to write, and couldn’t remember what it was. Thanks for reminding me :)

(actually there’s another piece missing which is #5505 )

PR #5512, opened 2021-11-07
nix parse: parse a nix expr or nix file to aterm or json syntax tree (AST)

### LINKED ISSUE #4726: feature: parse nix expression to json ast

milahu:
fix #4726
rebase of #4731

the `nix parse` CLI looks like this

```
nix parse input-file.nix
nix parse --expr '(if true then true else false)'
nix parse input-file.nix --output-format json
nix parse input-file.nix --output-format aterm # default format
```

stress test: parse `nixpkgs/pkgs/top-level/all-packages.nix` &rarr; success, `jq` says the json is valid

<details>
<summary>json output sample</summary>

```
nix parse --expr '(if true then true else false)' --output-format json | jq
```

```json
{
  "type": "ExprIf",
  "line": 1,
  "column": 2,
  "cond": {
    "type": "ExprVar",
    "line": 1,
    "column": 5,
    "name": "true"
  },
  "then": {
    "type": "ExprVar",
    "line": 1,
    "column": 15,
    "name": "true"
  },
  "else": {
    "type": "ExprVar",
    "line": 1,
    "column": 25,
    "name": "false"
  }
}
```

</details>


src/libexpr/nixexpr-as-json.cc
```diff
@@ -0,0 +1,326 @@
...
+    if ((std::uint8_t) *c <= 92)
+      o << String_showAsJson_replace_array[(std::uint8_t) *c];
+    else
+      o << *c;
+  }
+}
+
+void Expr::showAsJson(std::ostream & str) const
+{
+    abort();
```
andir, line 48: This should probably produce a descriptive error where someone has to look to fix the situation. I could imagine that with the next subclass of Expr it might be easily forgotten and then this will only yield an obscure runtime error.

src/libexpr/nixexpr-as-json.cc
```diff
@@ -0,0 +1,326 @@
...
+}
+
+void ExprInt::showAsJson(std::ostream & str) const
+{
+    str << "{\"type\":\"" << NodeTypeName::ExprInt << "\"";
```
andir, line 53: Any chance we could use a proper JSON serialization library instead? Nix already uses nlohmann JSON and has a home-grown JSON serialiser in `src/libutil/json.hh`.

Could the interface could be `void ExprType::showAsJson(json& list)` instead and an example implementation could look like this:
```c++
void ExprType::showAsJson(json& list) {
  json elem = {
    { "type", NodeTypeName::ExprType },
    { "value", n },
  };
  list.push_back(elem);
}
```

src/nix/parse.cc
```diff
@@ -0,0 +1,122 @@
...
+    void run() override
+    {
+        RepairFlag repair = NoRepair;
+
+        auto store = openStore();
+
+        Strings searchPath = {"."}; // TODO better?
+
+        auto state = std::make_unique<EvalState>(searchPath, store);
+        state->repair = repair;
```
andir, line 83: That is the default value of `repair`. Why do you define it explicitly?

src/libexpr/nixexpr-as-json.cc
milahu: uhm ... this is just copy-paste from `Expr::show` which i renamed to `Expr::showAsAterm`

src/libexpr/nixexpr-as-json.cc
milahu: > use a proper JSON serialization library

define proper ... this is working, and its fast
i did not use nlohmann, cos it has a different internal datastructure, iirc

src/nix/parse.cc
milahu: copy-paste from nix-instantiate, consider it removed ; )

src/libexpr/nixexpr-as-json.cc
milahu: > JSON serialiser in `src/libutil/json.hh`

would make an interesting benchmark ... someone, anyone? ; )

milahu: ... i would rather add a machine-readable json format, as i described in https://github.com/NixOS/nix/issues/4726#issuecomment-824993605
benefit: json is smaller, [can be parsed faster](https://github.com/NixOS/nix/pull/4731#issuecomment-831548502)

src/libexpr/nixexpr-as-json.cc
andir: ugh, this sin't "end of input" but instead terminates the process.

src/libexpr/nixexpr-as-json.cc
andir: >would make an interesting benchmark ... someone, anyone? ; )
My argument here isn't about performance at all. It is about correctness. It avoid programming errors that we can rule out by not doing the stringy stuff on our own. The nlohmann example also makes the code less verbose IMO.

milahu: > It is about correctness

can be verified by piping the output to `jq` which throws on json parse errors

```sh
nixpkgs=~/src/nixos/nixpkgs/

parallel --version | grep "GNU parallel" >/dev/null || {
  echo "requires gnu parallel"
  exit 1
}

test_nix_file() {
  f="$1"
  nix="./outputs/out/bin/nix --extra-experimental-features nix-command"
  # fixme: nix prints \r to stderr. workaround: 2>/dev/null
  json="$($nix parse --output-format json "$f" 2>/dev/null)" && {
      echo "$json" | jq >/dev/null && {
          #echo -e "ok $f"
          :
        } || {
          echo -e "json error in $f";
        }
    } || {
      echo -e "nix error in $f"
    }
}
export -f test_nix_file

find $nixpkgs -name '*.nix' | sort --random-sort \
  | parallel --jobs 50 -k --progress test_nix_file
```

nixpkgs has 20K nix files, so this will take some time ...

i found a small bug, fixed in 834f883
apart from that, no json errors so far : )

edgecases are non-ascii strings ... these are simply passed through

```
echo löl | iconv -f utf8 -t latin1
l�l

./outputs/out/bin/nix --extra-experimental-features nix-command parse \
  --expr '"'"$(echo löl | iconv -f utf8 -t latin1)"'"' | jq -r 
l�l
```

andir: Again it might work *now* but it requires special care whenever another Expr type is being added or when someone adds a character that you aren't escaping already. It is work that we don't have to do (and maintain). Just use a serializer that is already done.

piegamesde: If you care about performance, maybe JSON is not the best format you can aim for. And if you *really* care about performance, you should aim for a lazy event-driven API (the XML parser folks pioneered this and called it "SAX").

milahu: > Just use a serializer that is already done.

just merge my patch. : P

the whole point of my patch is to provide a FAST nix-to-json parser
and your "pretty" solution via some other high-level json printer is (probably) slower

"probably" = i did no benchmark, but its an educated guess

compare my 30 lines in [src/libexpr/nixexpr-as-json.cc](https://github.com/NixOS/nix/blob/834f883cb51acebce50107f4b9f0d50235e3fc42/src/libexpr/nixexpr-as-json.cc#L11)
(no need to print ascii, no need to validate input)

with 250 lines in nlohmann's `dump_escaped`
https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/serializer.hpp#L238
https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/serializer.hpp#L382

> it requires special care whenever another Expr type is being added

how often does that happen? once in five years?

the only "special care" i can think of is keeping backward compatibility
by appending new types at end-of-file in [src/libexpr/nixexpr-node-types.def](https://github.com/NixOS/nix/blob/834f883cb51acebce50107f4b9f0d50235e3fc42/src/libexpr/nixexpr-node-types.def)

> when someone adds a character that you aren't escaping already

what do you mean?
the json spec is constant
when bugs appear, someone will fix them

milahu: utf8 test

<details>

```sh
#!/bin/sh

nix="./outputs/out/bin/nix --extra-experimental-features nix-command"

echo 1 test null byte
printf '"a \0 z"' | $nix parse --output-format json /dev/stdin | jq
# error: syntax error, unexpected end of file, expecting '"'

echo 2 test valid utf8
inputString=$'one \1 two \2 three \3 four \4 skull \xE2\x98\xA0 lol \U0001f602'
printf '"%s"' "$inputString" | $nix parse --output-format json /dev/stdin | jq

echo 3 test invalid utf8
# invalid utf8 is ignored by nix parser, so its simply passed through
# https://stackoverflow.com/questions/1301402/example-invalid-utf8-string
inputString=$'\xc3\x28 \xa0\xa1 \xe2\x28\xa1'
printf '"%s"' "$inputString" | $nix parse --output-format json /dev/stdin | jq

echo 4 test random utf8 in 100 byte blocks ... this will loop forever
while true; do
# https://unix.stackexchange.com/questions/245623/how-do-i-create-a-text-file-1-gigabyte-containing-random-characters-with-utf-8
inputString="$(
  dd if=/dev/urandom bs=100 count=1 status=none | perl -CO -ne '
    BEGIN{$/=\4}
    no warnings "utf8";
    print chr(unpack("L>",$_) & 0x7fffffff)
  '
)"
printf '"%s"' "$inputString" | $nix parse --output-format json /dev/stdin | jq >/dev/null || {
  echo "json error in inputString:"
  echo "$inputString" | hexdump -C
}
sleep 0.1 # make the loop easier to kill
done
```

</details>

src/libexpr/nixexpr-as-json.cc
```diff
@@ -0,0 +1,326 @@
...
+
+namespace nix {
+
+// binary operators are implemented in nixexpr.hh MakeBinOp
+
+// https://stackoverflow.com/questions/7724448
+// note: we use full jump table to make this as fast as possible
+// note: we assume valid input. errors should be handled by the nix parser
+// 93 * 7 = 651 byte
+char String_showAsJson_replace_array[93][7] = {
```
andir, line 15: Any reason you aren't just using something like below?
```
const char* foo[] = { "1", "2", "3", "4" };
const size_t fooLength = sizeof(foo);
```

That would remove the need to track the byte sizes require for this escaping... Alternatively just use a JSON library ;-)

src/libexpr/nixexpr-as-json.cc
gilligan: https://en.cppreference.com/w/cpp/utility/program/abort

src/libexpr/nixexpr-as-json.cc
milahu: the "should be faster" reason. im not sure how your array would look in memory

milahu: this is actually an interesting question ...

how to find the memory layout of this program?

```c
#include <stdio.h>

char array1[3][10] = { "a", "b", "findme" };

const char* array2[] = { "1", "2", "findme" };

int main() {
    printf("%s", array1[0]);
    printf("%s", array1[1]);
    printf("%s", array2[0]);
    printf("%s", array2[1]);
}
```

with `gcc a.c -O0; pahole a.out` i dont see the `findme` value

andir: Check this out: https://godbolt.org/z/enxGvshhY

milahu: `const char a1[93][7]` is 5% faster than `const char* a2[]` based on [this benchmark](https://github.com/milahu/random/blob/master/cpp/benchmark-char-vs-char-pointer/bench-string-table-char-array.cpp)

src/libexpr/nixexpr-as-json.cc
milahu: was added in [d4f0b0fc6c](https://github.com/NixOS/nix/commit/d4f0b0fc6cdb9ae2fd5fc057d621bc80b36a7b18#diff-029ba0c36fadf8ae76fc565815d5d197104ff4b60643ca945ce145eaadf7b2d7) by @edolstra - maybe he knows : )
my second guess is: this means "empty input"

src/libexpr/nixexpr-as-json.cc
milahu: > when someone adds a character that you aren't escaping already

im starting to understand your concern

so, what im currently doing:
escape bytes 0 to 92 = ascii control chars (\n \r \b ...), doublequotes, backslash
everything else (bytes 93 to 255) is simply passed through as is

does this work with unicode?
as for the json spec, it is allowed to have raw unicode in json strings
when unicode is used in json, it must be utf16, for example \u0001

can the escaping break valid unicode?
lets look at the valid unicode byte ranges
ascii bytes are from 0 to 127 = 7 bit
so the first unicode byte must be 128 to 255

what about the following unicode bytes?

https://lemire.me/blog/2018/05/09/how-quickly-can-you-check-that-a-string-is-valid-unicode-utf-8/
converted with [hex2dec.py](https://stackoverflow.com/a/43545026/10440128)

First Byte | Second Byte | Third Byte | Fourth Byte
-- | -- | -- | --
[0,127] |   |   |  
[194,223] | [128,191] |   |  
224 | [160,191] | [128,191] |  
[225,236] | [128,191] | [128,191] |  
237 | [128,159] | [128,191] |  
[238,239] | [128,191] | [128,191] |  
240 | [144,191] | [128,191] | [128,191]
[241,243] | [128,191] | [128,191] | [128,191]
244 | [128,143] | [128,191] | [128,191]

so ... ALL unicode bytes are in the range from 128 to 255
and since i escape only bytes 0 to 92, this works : )

one rare edgecase, where this could break: non-unicode input, for example latin1 encoding.
some ascii control chars, like \1 are encoded as unicode \u0001,
so the result string can be mixed unicode and latin1. simple solution: blame the user.
either for throwing non-unicode strings at nix, or for throwing ascii control chars at nix.
(ascii control chars are worse than non-unicode strings)

gilligan:
I absolutely don’t understand who or what should benefit from re-inventing the wheel here with problems that have already been solved.

@Ericson2314 had created https://github.com/NixOS/nix/issues/3884 long ago to consolidate things and use only nlohman/json (not directly related but same principle) and this PR now is another example of fiddling with JSON and dealing with problems the nix code should not have to deal with.

Please just don’t. Optimize for ease of use, for stability and clean, fewer code instead of trying to make something faster just because you can.

milahu:
> Optimize for ease of use, for stability and clean, fewer code
> instead of trying to make something faster just because you can.

optimize for correct results and optimal performance
instead of trying to make something "pretty" just because [you](https://neilonsoftware.com/difficult-people-on-software-projects/developers/the-hostage-taker/) can

you have not one example where my code is failing
except in the unrealistic edgecases (ascii control chars + latin1) i described [here](https://github.com/NixOS/nix/pull/5512#discussion_r745358156)

> @Ericson2314 had created #3884 long ago to consolidate things and use only nlohman/json

this nlohman/json must be the greatest thing ever invented ...

did you read the code for [the nlohman/json string serializer](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/serializer.hpp#L382)?
why should i waste my cpu on this nonsense?

piegamesde:
@milahu Please take a step back and reflect on what you are writing here. If you'd like to know how to convince maintainers to merge your patch, well, this is how you *don't*. (For a start, try dropping the sarcasm and the personal attacks.)

Don't forget that somebody will have to maintain what you are writing. Nix is already a large enough pile of hacks that could fancy some cleanup, and you are just dumping more work onto that pile. *Of course* this will be met with contempt.

On another note, I still think JSON is not what you want if you really care that much about performance. It's like you're trying to make your go-kart a bit faster but actually you want a racing car.

milahu:
> convince maintainers to merge your patch

rich hearts wear poor coats
i know who i am, no need to fit your false expects

> somebody will have to maintain what you are writing

what exactly does that mean?
a new nix syntax? how often does that happen? daily? once every 1000 years?

all i hear is lazy/stupid excuses, and youre surprised im pissed?

nix is a low level tool, of course performance matters here
when "some people" dont get that
mabye they should stay away from the "dirty" backend work

> I still think JSON is not what you want if you really care that much about performance

off topic. this patch aims for a fast nix-to-json parser

milahu:
> this patch aims for a fast nix-to-json parser

better reasons to NOT use nix as parser:

`nix parse` will produce abstract syntax trees,
not concrete syntax trees as with [rnix-parser](https://github.com/nix-community/rnix-parser)

for example, `nix parse` will

* resolve relative paths to absolute paths
* parse `1 - 2` like `__sub 1 2` as [workaround for Sub/Negative conflict](https://github.com/milahu/lezer-parser-nix/commit/87607055e99b1449874b7fd313e1b1b3d7f0e6c4)
* parse `import <nixpkgs>` like `import (__findFile __nixPath "nixpkgs")`

also, no incremental parsing, as with  [tree-sitter-nix](https://github.com/cstrahan/tree-sitter-nix) or [lezer-parser-nix](https://github.com/milahu/lezer-parser-nix)

Ericson2314:
It would be indeed nice to be able to dump the AST to JSON, but it should use nlohmann. We've indeed made the other stuff all nlohmann now.

piegamesde:
OP has been banned from the organisation, I think this can safely be closed

PR #5524, opened 2021-11-09
installer: improve existing rc backup nag

abathur:
While looking into #5521 I noticed that the installer instructions don't really connect the dots about what the user is expected to do.

~This is a tentative rephrase, but I also realize that this kind of thing may well be informed by history--that we're intentionally leaving out steps so that the kinds of people confused by this will ask for help instead of screwing up their systems and trashing us on social.~ I think GC would've cautioned if this were the case.

Objections? @toonn @grahamc @domenkozar 

<details><summary>for reference, should look something like:</summary>

```
I back up shell profile/rc scripts before I add Nix to them. 
I need to back up /etc/bashrc to /etc/bashrc.backup-before-nix,
but the latter already exists.

Here's how to clean up the old backup file:

1. Back up (copy) /etc/bashrc and /etc/bashrc.backup-before-nix to
   another location, just in case.

2. Ensure /etc/bashrc.backup-before-nix does not have anything
   Nix-related in it. If it does, something is probably quite
   wrong. Please open an issue or get in touch immediately.

3. Once you confirm /etc/bashrc is backed up and doesn't already
   mention Nix, replace it with /etc/bashrc.backup-before-nix.
```
</details>

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -387,19 +387,27 @@ EOF
     fi
 
     for profile_target in "${PROFILE_TARGETS[@]}"; do
+        # TODO: I think it would be good to accumulate a list of all
+        #       of the copies so that people don't hit this 2 or 3x in
+        #       a row for different files.
         if [ -e "$profile_target$PROFILE_BACKUP_SUFFIX" ]; then
+            # this backup process first released in Nix 2.1
             failure <<EOF
-When this script runs, it backs up the current $profile_target to
-$profile_target$PROFILE_BACKUP_SUFFIX. This backup file already exists, though.
+I back up shell profile/rc scripts before I add Nix to them.
+I need to back up $profile_target to $profile_target$PROFILE_BACKUP_SUFFIX,
+but the latter already exists.
 
-Please follow these instructions to clean up the old backup file:
+Here's how to clean up the old backup file:
 
-1. Copy $profile_target and $profile_target$PROFILE_BACKUP_SUFFIX to another place, just
-in case.
+1. Back up (copy) $profile_target and $profile_target$PROFILE_BACKUP_SUFFIX
+   to another location, just in case.
 
-2. Take care to make sure that $profile_target$PROFILE_BACKUP_SUFFIX doesn't look like
-it has anything nix-related in it. If it does, something is probably
-quite wrong. Please open an issue or get in touch immediately.
+2. Ensure $profile_target$PROFILE_BACKUP_SUFFIX does not have anything
+   Nix-related in it. If it does, something is probably quite
+   wrong. Please open an issue or get in touch immediately.
+
+3. Once you confirm $profile_target is backed up and doesn't already
+   mention Nix, replace it with $profile_target$PROFILE_BACKUP_SUFFIX.
 EOF
         fi
     done


scripts/install-multi-user.sh
```diff
@@ -387,19 +387,25 @@ EOF
     fi
 
     for profile_target in "${PROFILE_TARGETS[@]}"; do
+        # TODO: I think it would be good to accumulate a list of all
+        #       of the copies so that people don't hit this 2 or 3x in
+        #       a row for different files.
```
abathur, line 392: I would, to be frank, like it to work as commented, and not as coded. 

I didn't want to let that pass completely undocumented--but I'm chasing other white whales at the moment and do not have the emotional/intellectual energy to make this happen without an installer test suite that ensures the important bits stay nailed down.

grahamc (approved):
I left some feedback with abathur in a PM on Matrix about point #3 and rewording it a bit, but this looks great to me.

abathur:
> I left some feedback with abathur in a PM on Matrix about point #3 and rewording it a bit, but this looks great to me.

I've force-pushed addressing feedback per #3 :)

scripts/install-multi-user.sh
```diff
@@ -387,19 +387,26 @@ EOF
...
         if [ -e "$profile_target$PROFILE_BACKUP_SUFFIX" ]; then
             failure <<EOF
-When this script runs, it backs up the current $profile_target to
-$profile_target$PROFILE_BACKUP_SUFFIX. This backup file already exists, though.
+Starting with Nix 2.1, I back up shell profile/rc scripts before I add
```
edolstra, line 395: Is the "Starting with Nix 2.1" bit necessary?

scripts/install-multi-user.sh
```diff
@@ -387,19 +387,26 @@ EOF
...
 
-2. Take care to make sure that $profile_target$PROFILE_BACKUP_SUFFIX doesn't look like
-it has anything nix-related in it. If it does, something is probably
-quite wrong. Please open an issue or get in touch immediately.
+2. Ensure $profile_target$PROFILE_BACKUP_SUFFIX does not have anything
+   Nix-related in it. If it does, something is probably quite
+   wrong. Please open an issue or get in touch immediately.
+
+3. Once you confirm $profile_target is backed up and doesn't already
+   mention Nix, replace it with $profile_target$PROFILE_BACKUP_SUFFIX.
```
toonn, line 409:
```suggestion
3. Once you confirm $profile_target is backed up and the backup doesn't already
   mention Nix, replace it with $profile_target$PROFILE_BACKUP_SUFFIX and
   make sure $profile_target$PROFILE_BACKUP_SUFFIX is no longer present.
```

I feel like what we're doing here is we have `mv $profile_target$PROFILE_BACKUP_SUFFIX $profile_target` in mind but since we're not outright telling the user to do so we have to be more precise. Someone could interpret "replace $target_profile" as copying the backup back to `/etc/bashrc` without removing it.

The other bit is "and doesn't already mention Nix" sounds like `/etc/bashrc` shouldn't already mention Nix but that's usually the case when `/etc/bashrc.backup-before-nix` has been created.

abathur: Good points, both. What about:
```
3. Once you confirm $profile_target is backed up and $profile_target$PROFILE_BACKUP_SUFFIX doesn't
   mention Nix, run: mv $profile_target$PROFILE_BACKUP_SUFFIX $profile_target`
```

scripts/install-multi-user.sh
abathur: I'm thinking ahead about something, but I suppose not it's not adding anything for now. 

I'll move that into a comment in a force-push, after I get feedback from toonn.

scripts/install-multi-user.sh
toonn: Yeah, suggesting the concrete command to run is probably better than a very verbose explanation.

Issue #5515, opened 2021-11-07
__functor causes infinite loop when it returns itself


andir:
**Describe the bug**
As of 81e7c40264520b387358917987d101f5f5ae4705 a functor that doesn't take two arguments but one leads to an inifite `callFunction` loop. Previously this lead to a stack exhaustion and did terminate eventually.

Example expression:

```nix
let a = { __functor = x: x; }; in a {}
```

This will cause the functor that resides within the attribute set `a` to return itself. Since `__functor` requires 2 (or more?) arguments, the language then tries to call the attribute set again and that results in the `__functor` function to be called again... This now loops until you manually terminate it.

The recent code changes seem to have encouraged the compiler to apply tail call optimization, and thus the loop never consumes the entire stack(?).

**Steps To Reproduce**

1. `nix-instantiate --eval -E '({ __functor = args@{ ... }: args; }) {}'`
2. Observe that it never terminates.
3. Attach `gdb` to the process and see that it loops in `callFunction` but never actually adds another stack frame.

**Expected behavior**

I'd expect the evaluation to realize that this is an infinite loop and terminate with an error.

Stack overflow on **2.3.16**
Infinite loop on **81e7c40264520b387358917987d101f5f5ae4705** (and newer).

**Additional context**

I encountered this while writing some additional libexpr unit tests for #5377 .

PR #5527, opened 2021-11-09
Support documentation on functors

edolstra:
For example, given a functor like

```
id = {
  __args = [ "x" ];
  __doc = ''The identity function. It takes a single argument *x* and returns it.'';
  __functor = _: x: x;
};
```

the repl can now print documentation about it:

```
nix-repl> :doc lib.id
Synopsis: lib.id x

    The identity function. It takes a single argument x and returns it.
```

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -656,13 +656,22 @@ Value & EvalState::getBuiltin(const string & name)
 
 std::optional<EvalState::Doc> EvalState::getDoc(Value & v)
 {
-    std::vector<std::string> args;
+    std::vector<Doc::Arg> args;
 
     auto getLambdaArgs = [&](Expr * expr, bool skipFirst)
     {
         while (auto lambda = dynamic_cast<ExprLambda *>(expr)) {
             // FIXME: handle attrset matches.
-            if (!skipFirst) args.push_back(lambda->arg != sEpsilon ? (std::string) lambda->arg : "_");
+            if (!skipFirst) {
+                Doc::Arg arg { .name = lambda->arg != sEpsilon ? (std::string) lambda->arg : "_" };
+                if (lambda->hasFormals()) {
+                    arg.attrs.emplace();
+                    arg.ellipsis = lambda->formals->ellipsis;
+                    for (auto & formal : lambda->formals->formals)
+                        arg.attrs->insert(formal.name);
+                }
+                args.push_back(std::move(arg));
+            }
             skipFirst = false;
             /* Note that we don't evaluate the body, so this only
                works if it's a lambda AST node. I.e. `x: ...` works,
@@ -672,16 +681,16 @@ std::optional<EvalState::Doc> EvalState::getDoc(Value & v)
         }
     };
 
-    if (v.isPrimOp()) {
-        auto v2 = &v;
-        if (v2->primOp->doc)
-            return Doc {
-                .pos = noPos,
-                .name = v2->primOp->name,
-                .arity = v2->primOp->arity,
-                .args = v2->primOp->args,
-                .doc = v2->primOp->doc,
-            };
+    if (v.isPrimOp() && v.primOp->doc) {
+        for (auto & arg : v.primOp->args)
+            args.push_back({.name = arg});
+        return Doc {
+            .pos = noPos,
+            .name = v.primOp->name,
+            .arity = v.primOp->arity,
+            .args = std::move(args),
+            .doc = v.primOp->doc,
+        };
     }
 
     else if (auto functor = isFunctor(v)) {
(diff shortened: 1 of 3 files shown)


andir:
Alternative idea:
```
nix-repl> f = builtins.addDoc "some doc string" (a: b: a + b)
nix-repl> x = builtins.addDoc "another doc string" { a = "1"; }
nix-repl> :d f
    some doc string
nix-repl> :d x
    another doc string
```

Without having spent too much time thinking about this the implementation could add a references to a String to the Value struct. This means we can document all and every single expression in the language, something that was always missing to generate docs from a set of functions (without resorting to some kind of hacks).

I don't really like the current ergonomics around the `__functor` implementation. `builtins.typeOf functor` always returns `set`. It is easy to get the functor syntax wrong and end up in an inifite loop (see #5515).

sternenseemann:
@andir's proposal would also allow for *arbitrary* values to be documented which could be quite nice.

edolstra:
@andir That was my original thought, but the problem is that Values don't have enough space to add annotations like docstrings. Functors on the other hand already are attrsets.

andir:
> @andir That was my original thought, but the problem is that Values don't have enough space to add annotations like docstrings. Functors on the other hand already are attrsets.

You mean we can't add a pointer to the Value struct that defaults to `nullptr`? If so, why is that?

edolstra:
Right, it would make every Value 8 bytes larger.

andir:
I've actually gone ahead and made that change (with another pointer value in each Value instance). It is very rough around the edges but in principle it works:
https://github.com/NixOS/nix/commit/8e252320f52426c734883709eb398d2161c3fe82

```

./benchmark/bin/benchmark ./with-docs/bin/nix-instantiate ~/dev/nixos/nixpkgs/nixos/tests/simple.nix
maximum RSS:        median = 561232.0000  mean = 561240.4444  stddev =     67.8953  min = 561180.0000  max = 561412.0000
soft page faults:   median = 138058.0000  mean = 138058.3333  stddev =      2.4495  min = 138055.0000  max = 138063.0000
system CPU time:    median =      0.2016  mean =      0.2019  stddev =      0.0077  min =      0.1907  max =      0.2136
user CPU time:      median =      1.6646  mean =      1.6743  stddev =      0.0275  min =      1.6316  max =      1.7097
elapsed time:       median =      2.0381  mean =      2.0230  stddev =      0.0382  min =      1.9672  max =      2.0772

./benchmark/bin/benchmark ./without-docs/bin/nix-instantiate ~/dev/nixos/nixpkgs/nixos/tests/simple.nix
maximum RSS:        median = 559624.0000  mean = 559636.8889  stddev =     63.0485  min = 559572.0000  max = 559744.0000
soft page faults:   median = 137672.0000  mean = 137674.1111  stddev =      5.3955  min = 137670.0000  max = 137687.0000
system CPU time:    median =      0.2137  mean =      0.2162  stddev =      0.0142  min =      0.1917  max =      0.2353
user CPU time:      median =      1.6506  mean =      1.6574  stddev =      0.0234  min =      1.6267  max =      1.6927
elapsed time:       median =      2.0295  mean =      2.0180  stddev =      0.0306  min =      1.9757  max =      2.0683
```

image


image

mightyiam:
@edolstra can we meet to discuss options regarding this, please?

roberth:
@edolstra My old PR https://github.com/NixOS/nix/pull/1652 achieves this without attrset overhead and with less syntax. Docs should be as easy as possible to write.

`:doc` wasn't a thing back then, but I'll be happy to update the code to work with the new command instead, if you agree with the source position based approach.

mightyiam:
Hey, @roberth, we are a team of three, working in mob programming format on core documentation issues in the Nix ecosystem. We'd love to coordinate efforts with you.

mightyiam:
We are @a-kenji, @101v and I.

roberth:
> Hey, @roberth, we are a team of three, working in mob programming format on core documentation issues in the Nix ecosystem. We'd love to coordinate efforts with you.

Hi @mightyiam and the mob,
You might have read in the other PR that Eelco and I disagree about the documentation format; comments vs __functor.
I don't plan to work on this controversial issue, so feel free to take over.

@edolstra How about having both? Documentation seems important enough to warrant the relative complexity of having both the added power of __functor and the ease of use of doc comments.

mightyiam:
Thank you, @roberth . We intend to treat https://github.com/NixOS/nix/issues/228 as the tracking issue for this feature in general.

We are working on a draft design that is solely from a user's perspective, user being a documentation author and consumer.
Will post an update there as soon as we have something to show.

tomberek:

PR #5529, opened 2021-11-09
rl-2.4.md: Evaluator requires registered store paths

### LINKED ISSUE #5495: Nix 2.4 regression, missing paths when evaluating inside the sandbox.
**Describe the bug**

While updating the `nix` package in Nixpkgs, https://github.com/NixOS/nixpkgs/pull/144197#issuecomment-958442489

The `nixpkgs-release-checks` derivation invokes a few `nix-*` commands inside the sandbox. This used to work.

The command that breaks inside the sandbox is:

```

        nix-env -f $src \
            --show-trace --argstr system "$platform" \
            --arg config '{ allowAliases = false; }' \
            -qa --drv-path --system-filter \* --system \
            "''${opts[@]}" 2>&1 >/dev/null | tee eval-warnings.log

```

**Steps To Reproduce**

1. Rebase nixpkgs PR 144197 onto 88d6a4f9ef8ec7aa00f6c810b3fe244f991eaf00 (base of the PR at the time of discovery)
2. `nix-build pkgs/top-level/release.nix -A tarball.nixpkgs-basic-release-checks`

**Expected behavior**

Build success.

My NixOS nix-daemon is
nix-env (Nix) 2.4pre20210707_02dd6bb

The Nix in the sandbox is the first 2.4 release.

**Additional context**

I've tried a different store location, but that did not make a difference; see diff.

```diff
--- a/pkgs/top-level/nixpkgs-basic-release-checks.nix
+++ b/pkgs/top-level/nixpkgs-basic-release-checks.nix
@@ -5,8 +5,8 @@ pkgs.runCommand "nixpkgs-release-checks" { src = nixpkgs; buildInputs = [nix]; }
 
     export NIX_STATE_DIR=$TMPDIR
     export NIX_PATH=nixpkgs=$TMPDIR/barf.nix
-    opts=(--option build-users-group "")
-    nix-store --init
+    opts=(--option build-users-group "" --option store "$TMPDIR/.physical-store-location")
+    nix-store --init "''${opts[@]}"
 
     echo 'abort "Illegal use of <nixpkgs> in Nixpkgs."' > $TMPDIR/barf.nix
```

Console log (regardless of diff above):

```
nixpkgs]$ nix-build pkgs/top-level/release.nix -A tarball.nixpkgs-basic-release-checks 
  /nix/store/bgm2hvq921fafirbqw7svjnvmnvy7kf6-nixpkgs-release-checks.drv
building '/nix/store/bgm2hvq921fafirbqw7svjnvmnvy7kf6-nixpkgs-release-checks.drv'...
checking Nixpkgs on x86_64-linux
error: path '/nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source' is not valid

       … while adding path '/nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/tools/misc/fx_cast'

       at /nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/build-support/nix-gitignore/default.nix:162:5:

          161|   gitignoreFilterSourcePure = filter: patterns: root:
          162|     filterSource (gitignoreFilterPure filter patterns root) root;
             |     ^
          163|

       … while evaluating 'gitignoreFilterSourcePure'

       at /nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/build-support/nix-gitignore/default.nix:161:49:

          160|
          161|   gitignoreFilterSourcePure = filter: patterns: root:
             |                                                 ^
          162|     filterSource (gitignoreFilterPure filter patterns root) root;

       … from call site

       at /nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/tools/misc/fx_cast/node-packages.nix:1697:13:

         1696|       name = args.name + "-package-json";
         1697|       src = nix-gitignore.gitignoreSourcePure [
             |             ^
         1698|         "*"

       … while evaluating the attribute 'src' of the derivation 'fx_cast_bridge-package-json'

       at /nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

       … while evaluating the attribute 'installPhase' of the derivation 'node-dependencies-fx_cast_bridge-0.1.2'

       at /nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

       … while evaluating the attribute 'buildPhase' of the derivation 'fx_cast_bridge-0.1.2'

       at /nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

       … while querying the derivation named 'fx_cast_bridge-0.1.2'
error: builder for '/nix/store/bgm2hvq921fafirbqw7svjnvmnvy7kf6-nixpkgs-release-checks.drv' failed with exit code 1;
       last 10 log lines:
       >        … while evaluating the attribute 'buildPhase' of the derivation 'fx_cast_bridge-0.1.2'
       >
       >        at /nix/store/hdr4yj37klan2w6dlif6hpbfazcxxb48-source/pkgs/stdenv/generic/make-derivation.nix:205:7:
       >
       >           204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
       >           205|       name =
       >              |       ^
       >           206|         let
       >
       >        … while querying the derivation named 'fx_cast_bridge-0.1.2'
       For full logs, run 'nix log /nix/store/bgm2hvq921fafirbqw7svjnvmnvy7kf6-nixpkgs-release-checks.drv'.

```

roberth:
Closes #5495 

Needs backport.

### DIFF
--- a/doc/manual/src/release-notes/rl-2.4.md
+++ b/doc/manual/src/release-notes/rl-2.4.md
@@ -339,6 +339,16 @@ more than 2800 commits from 195 contributors since release 2.3.
 * We no longer release source tarballs. If you want to build from
   source, please build from the tags in the Git repository.
 
+* The Nix evaluator refuses to read from unregistered store paths,
+  to prevent reading garbage files.
+  This breaks builds that run the Nix evaluator in the sandbox using
+  an initially empty `NIX_STATE_DIR` but still using the regular store.
+  Builds that run the evaluator must now use a fresh store dir, where
+  sources can be written, registered and imported as expected.
+  See [#5495](https://github.com/NixOS/nix/issues/5495#issuecomment-964443016)
+  and [Nixpkgs#144197](https://github.com/NixOS/nixpkgs/commit/ca5ecbed4b5775f873fa15d7bf064b00d9e2862a)
+  for an example of a fix.
+
 ## Contributors
 
 This release has contributions from


doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -339,6 +339,16 @@ more than 2800 commits from 195 contributors since release 2.3.
 * We no longer release source tarballs. If you want to build from
   source, please build from the tags in the Git repository.
 
+* The Nix evaluator refuses to read from unregistered store paths,
+  to prevent reading garbage files.
+  This breaks builds that run the Nix evaluator in the sandbox using
+  an initially empty `NIX_STATE_DIR` but still using the regular store.
+  Builds that run the evaluator must now use a fresh store dir, where
```
edolstra, line 346:
```suggestion
  Builds that run the evaluator must now use a fresh store, where
```

doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -339,6 +339,16 @@ more than 2800 commits from 195 contributors since release 2.3.
...
+  sources can be written, registered and imported as expected.
+  See [#5495](https://github.com/NixOS/nix/issues/5495#issuecomment-964443016)
+  and [Nixpkgs#144197](https://github.com/NixOS/nixpkgs/commit/ca5ecbed4b5775f873fa15d7bf064b00d9e2862a)
+  for an example of a fix.
```
edolstra, line 350: I'm not sure that those fixes are great advise. People really shouldn't use `NIX_STATE_DIR` and `NIX_STORE_DIR`. It's better to use `--store dummy://` if you don't need a writable store (which should be the case for Nixpkgs), or `--store /some/path` if you do.

roberth: I don't think `dummy` is sufficient, because it seems that the affected build needs to import from source; add a source (`builtins.path` etc) to the store and read back from it during evaluation.

edolstra: I think that points at a major bug in Nixpkgs. It should be possible to evaluate it in read-only mode. If that's not the case it has to be fixed.

roberth: `--store /some/path` is not sufficient and `--readonly-mode` does not seem to affect the outcome either. See https://github.com/NixOS/nixpkgs/pull/145501

edolstra: Strange, I tried `nix-env -qa --drv-path --store dummy://` and it works.

roberth: Well, it's still the same error.

regnat: @edolstra the problem isn’t really about having a read-only store or not.

The issue here (in the linked issue I mean) is that the evaluation happens inside a derivation (doing some kind of read-only recursive Nix), and the files being evaluated are the inputs of the derivation (so are under `/nix/store`). And when the evaluator tries to access a path that’s under `storeDir`, it will assume that it’s meant as a store path, and so check its validity. Which fails inside the sandbox because although it has a view of a subset of the store, it doesn’t have access to the host’s Nix db to know that the path is indeed valid. Hence the error.

The reason for using `NIX_STORE_DIR` is just to prevent Nix from trying to check these paths (which look like store paths, but aren’t really). Using `--store 'dummy://?store=/somewhere/else'` would probably have the same effect

roberth:
Please merge or close.

roberth:
Let's just move on.

PR #5534, opened 2021-11-10
2.4 Backport: Allow references in addPath

ElvishJerricco:
Backports #5494 to 2.4.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1859,12 +1859,12 @@ static void addPath(
         // be rewritten to the actual output).
         state.realiseContext(context);
 
+        StorePathSet refs = StorePathSet();
+
         if (state.store->isInStore(path)) {
             auto [storePath, subPath] = state.store->toStorePath(path);
-            auto info = state.store->queryPathInfo(storePath);
-            if (!info->references.empty())
-                throw EvalError("store path '%s' is not allowed to have references",
-                    state.store->printStorePath(storePath));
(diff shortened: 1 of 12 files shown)


edolstra:
Since this is a new feature and not a bug fix (AFAIK) we shouldn't backport it to 2.4.

yu-re-ka:
This feature was supported before in 2.3, so users have started relying on it. Not having it in 2.4 effectively would prevent nixpkgs from updating until 2.5.

yu-re-ka:
This feature was supported before in 2.3, so people have started relying on it.

chuchana:
Upgrading nix from v2.3.16 to v2.4 broke `nix-env --upgrade` for me:
`error: store path '/nix/store/[…]' is not allowed to have references`

Unless I'm mistaken, this should work again once the next version built off the current master is released. So while it might not technically be a bugfix, it very much feels like one.

Ericson2314:
Yes it's fixing and unintentional breaking change between 2.3 and 2.4, that's reason enough to consider it a bugfix.

yu-re-ka:
Any update on this?

jokogr:
This seems to block me from using `nix-shell` when on `nixpkgs-21.11-darwin` using nix flake and nix-darwin, #5539 could be related.

(Things were working just fine with nixpkgs-21.05-darwin and nix 2.3)

jokogr:
@edolstra would you consider merging this?

Mic92:
This version is no longer maintained.

PR #5536, opened 2021-11-11
Install nlohmann_json headers

roberth:
These headers are included by the libexpr, libfetchers, libstore
and libutil headers.
Considering that these are vendored sources, Nix should expose them,
as it is not a good idea for reverse dependencies to rely on a
potentially different source that can go out of sync.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ makefiles = \
   src/libexpr/local.mk \
   src/libcmd/local.mk \
   src/nix/local.mk \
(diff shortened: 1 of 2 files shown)


roberth:
@edolstra Could you review, to make the includes complete?

roberth:
@regnat Could you help us out?

grawlinson:
This particular commit causes conflicts with system provided package `nlohmann-json` on every Linux distribution.

PR #5538, opened 2021-11-11
src/nlohmann: Add license and readme

roberth:
We're not relicensing it.


edolstra:
json.hpp already contains the license, right? So I'm not sure that we need a separate license file.

roberth:
`json_fwd.hpp` doesn't, and I didn't read `json.hpp`.

So it wasn't obvious to me.

Issue #4003, opened 2020-09-10
`nix-shell` breaks with an eval-error in `nixpkgs` with Nix 3.0pre


Ma27:
**Describe the bug**

Running `nix-shell` on *any* `nixpkgs` checkout (checked on 1d60af77224f25a0c428b21a34463ed1719caa27) fails with the following error on Nix 3.0:

```
→ nix-shell -A hello
error: --- TypeError -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix-shell
at: (84:1) in file: /home/ma27/Projects/nixpkgs/pkgs/top-level/impure.nix

    83|
    84| import ./. (builtins.removeAttrs args [ "system" "platform" ] // {
      | ^
    85|   inherit config overlays crossSystem crossOverlays;

anonymous function at /home/ma27/Projects/nixpkgs/pkgs/top-level/default.nix:20:1 called with unexpected argument 'inNixShell'
```

**Expected behavior**

Getting an open `nix-shell`.

**`nix-env --version` output**

```
nix-env (Nix) 3.0pre20200829_f156513
```

**Additional context**

I didn't have the issue on `2.4pre20200721_ff314f1` and identified 626200713bb3cc844a9feb6af583c9b6b42c6dbc (from #3965) to be the problem. Reverting this commit on top of 77d4758cf2f45a6e999646e292ae563d6f987716 fixes the issue.

The patch 626200713bb3cc844a9feb6af583c9b6b42c6dbc fixes a bug where auto-calling a function `a@{ ... }: a` is missing all extra-arguments when being called from CLI with `nix-instantiate --argstr arg value` as the evaluator *only* added the attributes that were defined in the `{ ... }`-expression.

While this fixes an actual bug, this breaks [`pkgs/top-level/impure.nix` from `nixpkgs`](https://github.com/NixOS/nixpkgs/blob/6e72633c4202e1a22bbd0d3d7950e92e6de73f1d/pkgs/top-level/impure.nix) since this file defines a function with an *ellipsis*. And all arguments `a@{ ... }` from this function get passed to [`pkgs/top-level/default.nix`](https://github.com/NixOS/nixpkgs/blob/6e72633c4202e1a22bbd0d3d7950e92e6de73f1d/pkgs/top-level/default.nix#L20) which defines a function **without** an ellipsis.

[...]

PR #5543, opened 2021-11-12
Be mindful about the default arguments passed to nixpkgs in nix-shell

andir:
Starting with 6eeb6f9c84 Nix started injecting `inNixShell`
unconditonally into all nix-shell calls. This caused to a minor breakage
with old Nix expressions that had a thin wrapper around nixpkgs. While
we can fix those projects we should try to be backwards compatible in
this regard.

The error encountered in those cases looked like this:
```
    error: anonymous function at /home/andi/dev/nixos/nix/tests/shell-empty-attrs-args.nix:1:1 called with unexpected argument 'inNixShell'
           at «string»:1:18:

                1| {...}@args: with import <nixpkgs> args; (pkgs.runCommandCC or pkgs.runCommand) "shell" { buildInputs = [ (foo) (run) (echo "$(foo)") ]; } ""
                 |                  ^

           … while evaluating anonymous lambda

           at «string»:1:1:

                1| {...}@args: with import <nixpkgs> args; (pkgs.runCommandCC or pkgs.runCommand) "shell" { buildInputs = [ (foo) (run) (echo "$(foo)") ]; } ""
                 | ^

           … from call site
```

This should be backported to 2.4.

### DIFF
--- a/tests/nix-shell.sh
+++ b/tests/nix-shell.sh
@@ -52,6 +52,11 @@ output=$(NIX_PATH=nixpkgs="$shellDotNix" nix-shell --pure -p foo bar --run 'echo
 output=$(NIX_PATH=nixpkgs="$shellDotNix" nix-shell --pure -p foo --argstr fooContents baz --run 'echo "$(foo)"')
 [ "$output" = "baz" ]
 
+# Test nix-shell -p with older versions of Nixpkgs that do not expect
+# the (new) isNixShell attribute.
+output=$(NIX_PATH=nixpkgs=shell-empty-attrs-args.nix nix-shell --pure -p foo --run 'echo "$(foo)"')
+[ "$output" = "zes" ]
+
 # Test nix-shell shebang mode
 sed -e "s|@ENV_PROG@|$(type -P env)|" shell.shebang.sh > $TEST_ROOT/shell.shebang.sh
 chmod a+rx $TEST_ROOT/shell.shebang.sh
--- a/tests/nix-shell.sh
+++ b/tests/nix-shell.sh
@@ -52,6 +52,18 @@ output=$(NIX_PATH=nixpkgs="$shellDotNix" nix-shell --pure -p foo bar --run 'echo
 output=$(NIX_PATH=nixpkgs="$shellDotNix" nix-shell --pure -p foo --argstr fooContents baz --run 'echo "$(foo)"')
 [ "$output" = "baz" ]
 
+# Test that we are compatible with an old nixpkgs that doesn't accept the
+# inNixShell attribute
+output=$(NIX_PATH=nixpkgs="$PWD/shell-no-in-nix-shell.nix" nix-shell --pure -p foo)
+[ "$output" = "noInNixShell" ]
+
+# Test that passing arguments that aren't part of the formals (of the outer
+# shell expression) are still passed into the expression while *not* passing
+# inNixShell
+output=$(NIX_PATH=nixpkgs="$PWD/shell-no-in-nix-shell.nix" nix-shell --pure -p foo --argstr fooContents zes)
+[ "$output" = "zes" ]
+
+
 # Test nix-shell shebang mode
 sed -e "s|@ENV_PROG@|$(type -P env)|" shell.shebang.sh > $TEST_ROOT/shell.shebang.sh
 chmod a+rx $TEST_ROOT/shell.shebang.sh
--- a/tests/nix-shell.sh
+++ b/tests/nix-shell.sh
@@ -52,11 +52,6 @@ output=$(NIX_PATH=nixpkgs="$shellDotNix" nix-shell --pure -p foo bar --run 'echo
 output=$(NIX_PATH=nixpkgs="$shellDotNix" nix-shell --pure -p foo --argstr fooContents baz --run 'echo "$(foo)"')
 [ "$output" = "baz" ]
 
-# Test nix-shell -p with older versions of Nixpkgs that do not expect
-# the (new) isNixShell attribute.
-output=$(NIX_PATH=nixpkgs=shell-empty-attrs-args.nix nix-shell --pure -p foo --run 'echo "$(foo)"')
-[ "$output" = "zes" ]
-
 # Test nix-shell shebang mode
 sed -e "s|@ENV_PROG@|$(type -P env)|" shell.shebang.sh > $TEST_ROOT/shell.shebang.sh
 chmod a+rx $TEST_ROOT/shell.shebang.sh
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -270,7 +270,12 @@ static void main_nix_build(int argc, char * * argv)
 
     if (packages) {
         std::ostringstream joined;
-        joined << "{...}@args: with import <nixpkgs> args; (pkgs.runCommandCC or pkgs.runCommand) \"shell\" { buildInputs = [ ";
+        joined << "{...}@args: "
+               << "let nixpkgs = import <nixpkgs>; "
+               << "nixpkgsArgs = builtins.intersectAttrs (builtins.functionArgs nixpkgs) args; "
+               << "in "
+               << "with nixpkgs nixpkgsArgs; "
+               << "(pkgs.runCommandCC or pkgs.runCommand) \"shell\" { buildInputs = [ ";
         for (const auto & i : left)
             joined << '(' << i << ") ";
         joined << "]; } \"\"";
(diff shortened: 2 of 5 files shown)


grahamc:
Note a #!nix-shell script with -I nixpkgs= pinned to nixos-19.09 breaks without this patch.

andir:
@edolstra any chance you could have a look at this fix?

edolstra:
I'm getting a test failure:

```
    ++ [[  - foo - bar = \ \-\ \f\o\o\ \-\ \b\a\r ]]
    +++ NIX_PATH=nixpkgs=/build/source/tests/ca-shell.nix
    +++ nix-shell --pure -p foo bar --run 'echo "$(foo) $(bar)"'
    warning: unknown experimental feature 'ca-references'
      /build/nix-test/tests/ca/nix-shell/store/nc72sy0m12ya6w0rqaskrjk7x64hq58l-bar.drv
      /build/nix-test/tests/ca/nix-shell/store/ry7jj2wpkgddqjy0y1ww270hlbnplh0m-foo.drv
    Resolved derivation: '/build/nix-test/tests/ca/nix-shell/store/nc72sy0m12ya6w0rqaskrjk7x64hq58l-bar.drv' -> '/build/nix-test/tests/ca/nix-shell/store/x44b2vpcr9qwgfhi7140hjsvrirzs6wq-bar.drv'...
    Resolved derivation: '/build/nix-test/tests/ca/nix-shell/store/ry7jj2wpkgddqjy0y1ww270hlbnplh0m-foo.drv' -> '/build/nix-test/tests/ca/nix-shell/store/mbllmk66bnzhfjsvsmw0qy3kgcaqlrlk-foo.drv'...
    warning: unknown experimental feature 'ca-references'
    building '/build/nix-test/tests/ca/nix-shell/store/x44b2vpcr9qwgfhi7140hjsvrirzs6wq-bar.drv'...
    building '/build/nix-test/tests/ca/nix-shell/store/mbllmk66bnzhfjsvsmw0qy3kgcaqlrlk-foo.drv'...
    ++ output='foo bar'
    ++ '[' 'foo bar' = 'foo bar' ']'
    +++ NIX_PATH=nixpkgs=/build/source/tests/ca-shell.nix
    +++ nix-shell --pure -p foo --argstr fooContents baz --run 'echo "$(foo)"'
    warning: unknown experimental feature 'ca-references'
    ++ output=foo
    ++ '[' foo = baz ']'
make: *** [mk/lib.mk:123: tests/ca/nix-shell.sh.test] Error 1
```

tests/nix-shell.sh
```diff
@@ -58,6 +58,11 @@ output=$(NIX_PATH=nixpkgs=shell.nix nix-shell --pure -p foo bar --run 'echo "$(f
 output=$(NIX_PATH=nixpkgs=shell.nix nix-shell --pure -p foo --argstr fooContents baz --run 'echo "$(foo)"')
 [ "$output" = "baz" ]
 
+# Test nix-shell -p with custom nixpkgs expression that does not expect
+# the (new) isNixShell attribute
```
edolstra, line 62:
```suggestion
# Test nix-shell -p with older versions of Nixpkgs that do not expect
# the (new) isNixShell attribute.
```

We don't support `nixpkgs` pointing to things that are not Nixpkgs. We do want to support older versions of Nixpkgs however.

andir:
> I'm getting a test failure:
> 
> ```
>     ++ [[  - foo - bar = \ \-\ \f\o\o\ \-\ \b\a\r ]]
>     +++ NIX_PATH=nixpkgs=/build/source/tests/ca-shell.nix
>     +++ nix-shell --pure -p foo bar --run 'echo "$(foo) $(bar)"'
>     warning: unknown experimental feature 'ca-references'
>     these 2 derivations will be built:
>       /build/nix-test/tests/ca/nix-shell/store/nc72sy0m12ya6w0rqaskrjk7x64hq58l-bar.drv
>       /build/nix-test/tests/ca/nix-shell/store/ry7jj2wpkgddqjy0y1ww270hlbnplh0m-foo.drv
>     Resolved derivation: '/build/nix-test/tests/ca/nix-shell/store/nc72sy0m12ya6w0rqaskrjk7x64hq58l-bar.drv' -> '/build/nix-test/tests/ca/nix-shell/store/x44b2vpcr9qwgfhi7140hjsvrirzs6wq-bar.drv'...
>     Resolved derivation: '/build/nix-test/tests/ca/nix-shell/store/ry7jj2wpkgddqjy0y1ww270hlbnplh0m-foo.drv' -> '/build/nix-test/tests/ca/nix-shell/store/mbllmk66bnzhfjsvsmw0qy3kgcaqlrlk-foo.drv'...
>     warning: unknown experimental feature 'ca-references'
>     building '/build/nix-test/tests/ca/nix-shell/store/x44b2vpcr9qwgfhi7140hjsvrirzs6wq-bar.drv'...
>     building '/build/nix-test/tests/ca/nix-shell/store/mbllmk66bnzhfjsvsmw0qy3kgcaqlrlk-foo.drv'...
>     ++ output='foo bar'
>     ++ '[' 'foo bar' = 'foo bar' ']'
>     +++ NIX_PATH=nixpkgs=/build/source/tests/ca-shell.nix
>     +++ nix-shell --pure -p foo --argstr fooContents baz --run 'echo "$(foo)"'
>     warning: unknown experimental feature 'ca-references'
>     ++ output=foo
>     ++ '[' foo = baz ']'
> make: *** [mk/lib.mk:123: tests/ca/nix-shell.sh.test] Error 1
> make: *** Waiting for unfinished jobs....
> ```

I am not able to reproduce that locally with `make installcheck`. However, I had to revert 09471d2680292af48b2788108de56a8da755d661 as that caused a test to fail with `can't compare lists to lists` (or something along the lines).

andir:
I wonder how CI ever passed with my patch. 🤔 Anyway. The implementation is wrong and the fix doesn't appear to be as simple as I did hope.

First of, the current implementation (proposed here in  e1b9bb70b7a13707c3ec8d24aab739c7580782fb and earlier) removes all arguments that are not part of the formal function arguments. Less than ideal and probably breaking someone's code.

Then there is also the whole story that is covered in #4003. While that issue was forward fixed in Nixpkgs we never thought about backwards compatibility in that PR (or related conversations).

The flaw in my implementation is that it might work in the tests but wouldn't work with actual Nixpkgs. In nixpkgs the attribute is never part of any function signature but is just removed from the arguments if passed:
https://github.com/Ma27/nixpkgs/blob/fa6064ad86a020f52a75219bff044714f2a9ebbd/pkgs/top-level/impure.nix#L84

If we want to retain backwards compatibility, we can't inject the `inNixShell` attribute without hacks such as evaluating without it if it fails the first time around.

I've reverted my changes in this PR and added test cases that should pass if we are backwards compatible instead.


@edolstra any pointers on what we should do here? Isn't the `IN_NIX_SHELL` environment variable enough to do whatever the feature was originally designed to do?

edolstra:
@andir See #3168. The problem with `IN_NIX_SHELL` is that it has unclear semantics. It's used to influence both setup hooks (like `shellHook`) and evaluation (as in `src = if getEnv "IN_NIX_SHELL" != "" then null else ./.`). The latter means that unrelated evaluations (e.g. a `nix-build` inside a `nix-shell`) will be affected by `IN_NIX_SHELL` even when that's not intended. Also, it's better not to rely on environment variables in Nix expressions, because they don't work in pure mode.

andir:
> @andir See #3168. The problem with `IN_NIX_SHELL` is that it has unclear semantics. It's used to influence both setup hooks (like `shellHook`) and evaluation (as in `src = if getEnv "IN_NIX_SHELL" != "" then null else ./.`). The latter means that unrelated evaluations (e.g. a `nix-build` inside a `nix-shell`) will be affected by `IN_NIX_SHELL` even when that's not intended. Also, it's better not to rely on environment variables in Nix expressions, because they don't work in pure mode.

When I referred to the environment variable I meant during the shells runtime. Having it available during eval is already impure and that's bad. I don't get why the Nix argument is required at all. It sounds like a hack. We already have shell.nix and equivalent flake attributes for shell expressions. In the simplest case these are the same expression as for a nix build. And if someone wants to reuse the code they should pass that flag via one of the shell expressions.

Injecting `inNixShell` leaks the runtime environment (e.g. the Nix version used) into the expression. It is in no way better that the environment variable we had before.

shurrman:
@andir @edolstra Hi, is there a way to avoid it in a pkgs pinned to 19.x?

andir:
> @andir @edolstra Hi, is there a way to avoid it in a pkgs pinned to 19.x?

Use a stable nix release (<=2.3).

andir:
Since nobody cares about backwards compat I'll close this.

Issue #5507, opened 2021-11-05
Have a mechanism for trusting flake-local Nix options when stdin isn’t a tty


regnat:
When a flake sets a non-whitelisted Nix option, Nix trusts it on a TOFU model: The first call to `nix build` (or any other Nix command) will ask the user whether he wants to apply the option).
However that’s not possible when stdin isn’t a TTY (say a CI job), and the current behavior is to always assume `no` in that case.

Although that’s the sensible default option, there should be a way to override this, probably an `--accept` flag (or at least a way to force Nix in an interactive mode so that we can `yes | nix build`)

_Originally posted by @domenkozar in https://github.com/NixOS/nix/issues/5504#issuecomment-962026288_

PR #5544, opened 2021-11-12
Adds an accept-flake-config flag

tomberek:
Adds a warning when a saved setting is utilized.
Flattens the if logic structure
Corrects incorrect logic where if a setting was accepted, but not allowed to be written; in that case the setting would not be used for that instantiation.

## TODO
- [x] Test case
- [ ] triple-check the if logic

Note: Created during NixUX Hacking Fridays! @regnat @garbas @Radvendii 

Fixes: #5507


regnat:
image :tada:

PR #5546, opened 2021-11-12
allow paths in flake local options

### LINKED ISSUE #5505: Allow paths in flake-local options

Flakes can define some Nix options with the `nixConfig` toplevel attribute. However these can’t be paths, meaning that there’s no way to refer to a file by its relative path.

Be able to write `nixConfig.post-build-hook = ./my-post-hook.sh;`

Allow relative paths for these configs (as in "strings-that-represent-relative-paths"), like #4875 requests.
That should be mostly similar, except that the "path" syntax is able to cache the errors earlier.

regnat:
Fix #5505

Depends on #5544 because they both touch the same test, and #5544 made the testing easier

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -251,6 +251,10 @@ static Flake getFlake(
             forceTrivialValue(state, *setting.value, *setting.pos);
             if (setting.value->type() == nString)
                 flake.config.settings.insert({setting.name, state.forceStringNoCtx(*setting.value, *setting.pos)});
+            else if (setting.value->type() == nPath) {
+                PathSet emptyContext = {};
+                flake.config.settings.insert({setting.name, state.coerceToString(*setting.pos, *setting.value, emptyContext, false, true, true)});
+            }
             else if (setting.value->type() == nInt)
                 flake.config.settings.insert({setting.name, state.forceInt(*setting.value, *setting.pos)});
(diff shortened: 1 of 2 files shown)


nrdxp:
Tested this locally to attempt a workaround for #5694 and it appears to work properly. Also attempted to set a path outside the flake just to make sure it fails properly and it appears to be all good. Thanks for this :+1:

nrdxp (approved):

nrdxp:
So this is working as is, as this comment could be used for motivation for a future PR perhaps, but it would be nice if we could put a nix store shebang at the top of the script, similar to how `writeShellScript` in nixpkgs does it. The PATH of the `post-build-hook` in particular is fairly skimp and I'm getting an error when trying `#!/usr/bin/env bash`. I could do `#!/bin/sh` but I'm using bash specific syntax in my script, so if `bin/sh` is not a symlink to bash it will fail.

May not be possible though, the way flakes are currently evaled.

regnat:
> it would be nice if we could put a nix store shebang at the top of the script, similar to how `writeShellScript` in nixpkgs does it. 
> 
> May not be possible though, the way flakes are currently evaled.

Yeah, that’s a slightly tricky thing to do (would essentially be one giant IFD), so it’s probably better to keep things simple for the time being. But it should ideally be possible to do something like that indeed.

Issue #5463, opened 2021-11-02
Document builtins.fetchTree


edolstra:
It's experimental but it still should be documented.

kamadorueda:
@edolstra I started documenting fetchTree at https://github.com/NixOS/nix/pull/5556, initially for the git backend 

I would like to help documenting the remaining backends, too

elikoga:
Can rework my WIP PR to the state of #5556

thufschmitt:
Fixed by https://github.com/NixOS/nix/pull/8509

PR #5556, opened 2021-11-14
fetchTree: add documentation

kamadorueda:
- Since this section is going to be long
  due to the need of documenting the different backends
  then giving it its own section seems reasonable.
  Additionally, by introducing fetchers before derivations and
  after the basics of the language,
  it guides the lector through the mental model
  that in Nix you build your derivations from the data you fetch
  in a reproducible way in the first place
- Start by documenting the `git` backend and then I'll be adding more

Mentions #5463

### DIFF
--- a/doc/manual/src/expressions/fetchers.md
+++ b/doc/manual/src/expressions/fetchers.md
@@ -1,0 +1,147 @@
+**Warning**:
+This section is **experimental** and its interface is subject to change.
+
+# Fetchers
+
+Nix supports fetching source trees
+from different locations
+through the `fetchTree` builtin.
+
+Below we document the supported backends
+and their configuration options.
+
+## Git
+
+Allows you to fetch a Git source tree (similar to a `git clone`).
+
+> **Note**
+>
+> The `.git/` folder won't be saved for reproducibility.
+
+Options:
+
+- `type`
+
+  Must be equal to `"git"`.
+- `url`
+
+  Location of the git repository as accepted by `git`
+  (i.e. a local path, http(s) or SSH URL).
+- `ref`
+
+  The git ref to look for the requested revision under.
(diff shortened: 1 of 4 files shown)


doc/manual/src/expressions/fetchers.md
```diff
@@ -0,0 +1,147 @@
+**Warning**:
+This section is **experimental** and its interface is subject to change.
```
andir, line 2: How can this be experimental when all the old fetchers are now using this under the hood?

kamadorueda: I added the warning because of this small detail: https://github.com/NixOS/nix/blob/bceda304982a34c65d5a1dab449cb5bc59f63b83/src/libexpr/primops/fetchTree.cc#L179

edolstra:
The fetchers stuff is already mostly documented in the `nix flake` manpage (under "Flake references"). I think it's better to have the `fetchTree` doc refer to that, rather than duplicate it in another section.

bew:
> The fetchers stuff is already mostly documented in the `nix flake` manpage (under "Flake references"). I think it's better to have the `fetchTree` doc refer to that, rather than duplicate it in another section.

I agree, let's not duplicate everything, but what about the other way around?
The `nix flake` doc page is already quite big, maybe it could refer to the `Fetchers` doc page for the doc of individual fetchers? (which could also document `fetchTree`?)

doc/manual/src/expressions/fetchers.md
regnat: > How can this be experimental when all the old fetchers are now using this under the hood?

I don’t think here’s a contradiction here: This is indeed used internally, but the interface isn’t stable yet. So it makes sense to mark this as experimental

Issue #5558, opened 2021-11-14
Gitlab set free header


Kloenk:
**Describe the bug**
https://github.com/NixOS/nix/blob/8d12084707be9da7a03a33e8de95766854818fcb/src/libfetchers/github.cc#L289-L295

documents that any header can be set for gitlab libfetchers requests.

**Steps To Reproduce**

1. use `access-token` setting for a gitlab libfetcher request with something other that `Oauth2` or `PAT`
2. Warning is issues

**Expected behavior**
header should be set.

**`nix-env --version` output**
master

PR #5559, opened 2021-11-14
libfetchers: set free gitlab headers

Kloenk:
fixes: #5558

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -300,7 +300,7 @@ struct GitLabInputScheme : GitArchiveInputScheme
         if ("PAT" == token.substr(0, fldsplit))
             return std::make_pair("Private-token", token.substr(fldsplit+1));
         warn("Unrecognized GitLab token type %s",  token.substr(0, fldsplit));
-        return std::nullopt;
+        return std::make_pair(token.substr(0,fldsplit), token.substr(fldsplit+1));
     }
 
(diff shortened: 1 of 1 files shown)


src/libfetchers/github.cc
```diff
@@ -300,7 +300,7 @@ struct GitLabInputScheme : GitArchiveInputScheme
         if ("PAT" == token.substr(0, fldsplit))
             return std::make_pair("Private-token", token.substr(fldsplit+1));
         warn("Unrecognized GitLab token type %s",  token.substr(0, fldsplit));
```
edolstra, line 302: Should this warning be removed then? If these tokens are actually fine, then we shouldn't spam the user with useless warnings.

yu-re-ka:
This allows me to use `JOB-TOKEN:...` to fetch private projects from CI jobs (with an additional gitlab patch to allow these necessary endpoints with job tokens).

src/libfetchers/github.cc
yu-re-ka: While tokens other than PAT and OAuth2 are not necessarily bad, they are also not confirmed to be working out of the box. Some tokens are not allowed to call the necessary endpoints.

The warning might also be useful for detecting typos of users who wanted to write one of the known token types.

So it would be reasonable to keep the warning, while allowing users to use any token type if they know what they're doing.

PR #5571, opened 2021-11-16
[2.3-maintenance] Override socket path with `NIX_DAEMON_SOCKET_PATH` so we can test 2.4+ compat

Ericson2314:
This is the first step to testing compatibility with 2.3 in 2.4 and
master.

(adapted from commit 223fbe644a4cde45269c01e971331d9414971d46)

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -34,7 +34,7 @@ Settings::Settings()
     , nixLibexecDir(canonPath(getEnv("NIX_LIBEXEC_DIR", NIX_LIBEXEC_DIR)))
     , nixBinDir(canonPath(getEnv("NIX_BIN_DIR", NIX_BIN_DIR)))
     , nixManDir(canonPath(NIX_MAN_DIR))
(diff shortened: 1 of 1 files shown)


Ericson2314:
@regnat this is in response to your comment
https://github.com/NixOS/nix/blob/b367f1061c34f8f358ebf1f02c7bdbb7be6df645/flake.nix#L508-L510
from a0866c8ea4bc66f9aacc7ad19139d57946b3df18. #5572 can be updated to show the interesting failures we are currently trying to fix once this is merged.

regnat (approved):
> @regnat this is in response to your comment

Yup’ indeed. Dunno why I didn’t backport it at the time.

43d4d75e22ea1f7ee2936fe725c1f7ea7a5005e6 might even be a better thing to backport as it avoids the need for a short drv name, but it might be a bit more involved, so this is a good middle-ground

Ericson2314:
Yes I agree 43d4d75 a nice thing worthy of being backported, but I am not sure what the appetite is for moving things to 2.3 no matter how good, so I just did the bare minimum to unblock this testing.

Ericson2314:
Thanks @edolstra!

PR #5572, opened 2021-11-16
Test against stable versions

Ericson2314:
This will currently fail, but after #5571 and a bump to the 2.3 input, might succeed.

Helps with #12461

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,6 +16,59 @@
         "type": "github"
       }
     },
+    "lowdown-src_2": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1633514407,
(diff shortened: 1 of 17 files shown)


regnat:
I’m not sure that we want to test against both 2.3 and 2.4 in the Github CI.
That would be great of course, but these tests are also quite costly, and the CI is already pretty slow. So maybe just sticking to 2.4 (or the version in nixpkgs as we currently do) would be enough.

Hydra is another matter, ofc, and it would probably make sense to have all these checks there

Ericson2314:
@regnat That makes sense and sounds good to me. Also, I hope to backport this to 2.4, after which one could argue master testing 2.4 is enough.

Ericson2314:
@andir another such testing across versions PR.

tomberek:

PR #5577, opened 2021-11-16
Add haskell's $ operator to the nix language

### LINKED ISSUE #1845: Adding an operator like Haskell's '$'
The title says it all, I'm thinking it would be useful for eg.
```nix
builtins.trace "my debug stuff" $
# some multiline complex code that would require parenthesing otherwise
```

What do you think about it? I guess it's too late for the feature freeze, but $ being already reserved for ${} I guess it wouldn't be a breaking change anyway.

Mewp:
Since I started using nix, one thing has been constantly bothering me: lack of the `$` operator in the nix language.

One doesn't need to look far to find function calls with nested parentheses that would be better off without them, and I couldn't find a reason for the absence of this operator. So I implemented it to see if this would work.

The patch is simple, all tests pass, including two simple test cases added to test this operator.
I have also manually tried parsing my own nix derivations with added usages of this new operator.
That being said, I do not claim to have thought of every possible edge case.

The operator itself is the same as function application, i.e. `foo a`, except that `foo bar baz` is left associative, and `foo $ bar $ baz` is right associative.

Also, this PR fixes #1845 .


edolstra:
I'm not really in favor of this. It's another syntax for Nix users to learn, while the benefit is saving a small amount of typing. Nix is not a general purpose language, so convenience features like this are probably less important than keeping the learning curve manageable.

If we add something like this, I would be more in favor of a pipe syntax (`f | g | h` instead of `h $ g $ f`).

L-as:
Changes like this should probably go through an RFC FYI.

max-privatevoid:
I don't think the learning curve will be impacted much by this. Using this operator is completely optional and when a user runs into a situation where it makes sense to use it, they are likely quite proficient in Nix already and working on moderately complex code.

thufschmitt:
Closing for the same reasons as https://github.com/NixOS/nix/issues/1845. Let's get some community consensus on this first

PR #5465, opened 2021-11-02
Fixes #5376

ck3d:
scp uri can not contain a slash before @

Fixes #5376

Issue #5376, opened 2021-10-12
fetchGit failes with non default ssh port


ck3d:
Since https://github.com/NixOS/nix/pull/4978 was merged, fetchGit do not respect a ssh port.

**Steps To Reproduce**

With current Nix 2.4pre20211006_53e4794:
```
nix-repl> builtins.fetchGit { url = "ssh://git@github.com:22/NixOS/nix.git"; }
fetching Git repository 'ssh://git@github.com/22/NixOS/nix.git'fatal: remote error:
  22/NixOS/nix is not a valid repository name
  Visit https://support.github.com/ for help
error: program 'git' failed with exit code 128
```

With old Nix 2.4pre20210922_bcd73eb:
```
nix-repl> builtins.fetchGit { url = "ssh://git@github.com:22/NixOS/nix.git"; }
{ lastModified = 1634049780; lastModifiedDate = "20211012144300"; narHash = "sha256-NYjt6jTZyt35hJJfjQsVPB4w7iaN/z40Vzp0k1q1cKA="; outPath = "/nix/store/k302z97m35fl3cnycrgq814kfb490f48-source"; rev = "5176b072ed55d1528cb36e0c865e4fcffbc6a7a1"; revCount = 10824; shortRev = "5176b07"; submodules = true; }
```

**Expected behavior**

fetchGit shall respect the given ssh port as it does for Nix 2.4pre20210922_bcd73eb.

markuskowa:
it would be great if this could be fixed before the 2.4 release.

ck3d:
This issue was fixed in master, but it is still relevant for 2.4

terlar:
Yeah, it would be nice if this could be added to the official 2.4 release. We start to have first-time users of Nix run into this at work and it is making the road a bit bumpy.

ck3d:
Fix was back ported in https://github.com/NixOS/nix/pull/5655

PR #5580, opened 2021-11-17
Fix detection of scp-style URIs to support non-standard SSH ports

ksonj:
This PR attempts to fix the broken detection of SCP style URIs (e.g. `git@github.com:NixOS/nix`). Currently the regex also matches URIs with a fixed port, e.g. `ssh://git@github.com:22/NixOS/nix`. This worked with Nix 2.3 but does not currently with Nix 2.4

Since scp URIs do not have a scheme this simply adds a test for the existence of a "://" substring

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -74,8 +74,11 @@ std::string fixURI(std::string uri, EvalState & state, const std::string & defau
 
 std::string fixURIForGit(std::string uri, EvalState & state)
 {
+    /* Detects scp-style uris (e.g. git@github.com:NixOS/nix) and fixes
+     * them by removing the `:` and assuming a scheme of `ssh://`
+     * */
     static std::regex scp_uri("([^/].*)@(.*):(.*)");
-    if (uri[0] != '/' && std::regex_match(uri, scp_uri))
+    if (uri[0] != '/' && std::regex_match(uri, scp_uri) && uri.find("://") == std::string::npos)
         return fixURI(std::regex_replace(uri, scp_uri, "$1@$2/$3"), state, "ssh");
     else
         return fixURI(uri, state);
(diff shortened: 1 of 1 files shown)


ksonj:
I just realized there's already https://github.com/NixOS/nix/issues/5376 and a very similar fix in https://github.com/NixOS/nix/pull/5465. The only difference is that the other PR assumes a scheme of "ssh://" to rule out scp-style which I believe is not always accurate.

E.g. the URL `https://ksonj@github.com:443/NixOS/nix` works with Nix 2.3 and with this fix, but not with 2.4 and still not with the other PR

src/libexpr/primops/fetchTree.cc
```diff
@@ -74,8 +74,11 @@ std::string fixURI(std::string uri, EvalState & state, const std::string & defau
 
 std::string fixURIForGit(std::string uri, EvalState & state)
 {
+    /* Detects scp-style uris (e.g. git@github.com:NixOS/nix) and fixes
+     * them by removing the `:` and assuming a scheme of `ssh://`
+     * */
     static std::regex scp_uri("([^/].*)@(.*):(.*)");
```
regnat, line 80: Couldn’t this be fixed by just changing the regex to

```suggestion
    static std::regex scp_uri("([^/]*)@(.*):(.*)");
```

?

I have the feeling that this `.` was just a typo

regnat: (I think that the best specification for what’s allowed in an ssh username is https://datatracker.ietf.org/doc/html/rfc2396#section-3.2.2 , which fobids `/` in a username. So it’s not overly restrictive either)

ksonj: :bulb: that is true. Let me quickly fix that

regnat (approved):

Ericson2314:
Should this be backported if there is a regression from 2.3? Do we have a test?

regnat:
> Should this be backported if there is a regression from 2.3?

Yup’ indeed, it should

> Do we have a test?

Unfortunately not. I couldn’t find a way to write a meaningful test for this without either some non-trivial refactoring to make this pure and allow for unit-testing, or having to spawn-up an ssh server to make sure that everything works end-to-end, which might be a bit too much

regnat:
> or having to spawn-up an ssh server to make sure that everything works end-to-end, which might be a bit too much

Although now that I think of it, that could be a good fit for a hydra test

PR #5585, opened 2021-11-17
doc: De-emphasize nix-env without -A

jtojnar:
The manual uses `nix-env -i` without `-A` prominently, teaching a bad practice to newcomers.

### DIFF
--- a/doc/manual/src/command-ref/nix-env.md
+++ b/doc/manual/src/command-ref/nix-env.md
@@ -238,13 +238,25 @@ a number of possible ways:
 
 ## Examples
 
-To install a specific version of `gcc` from the active Nix expression:
+To install a package using a specific attribute path from the active Nix expression:
+
+```console
+$ nix-env -iA gcc40mips
+installing `gcc-4.0.2'
+$ nix-env -iA xorg.xorgserver
+installing `xorg-server-1.2.0'
+```
+
+To install a specific version of `gcc` using the derivation name:
 
 ```console
 $ nix-env --install gcc-3.3.2
 installing `gcc-3.3.2'
 uninstalling `gcc-3.1'
 ```
+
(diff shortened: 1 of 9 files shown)


doc/manual/src/introduction.md
```diff
@@ -76,7 +76,7 @@ there after an upgrade.  This means that you can _roll back_ to the
 old version:
 
 ```console
-$ nix-env --upgrade some-packages
+$ nix-env --upgrade -A some-packages
```
SuperSandro2000, line 79: Do we want to add nixpkgs. here, too?

doc/manual/src/command-ref/nix-env.md
```diff
@@ -238,31 +238,37 @@ a number of possible ways:
 
 ## Examples
 
-To install a specific version of `gcc` from the active Nix expression:
+To install a package using a specific attribute path from the active Nix expression:
+
+```console
+$ nix-env -i -A gcc40mips
+installing `gcc-4.0.2'
+$ nix-env -i -A xorg.xorgserver
```
SuperSandro2000, line 246:
```suggestion
$ nix-env -iA gcc40mips
installing `gcc-4.0.2'
$ nix-env -iA xorg.xorgserver
```
Everywhere else this is used. Not sure what is better for the doc, I let you decide.

SuperSandro2000 (approved):

doc/manual/src/command-ref/nix-env.md
jtojnar: Yeah, I did not change the original format when moving it around but I guess it is a good idea to unify this when touching it. Examples should show what people want to use, after all.

jtojnar:
Could we also backport this to the 2.4 branch so that it gets to the stable manual?

Issue #5487, opened 2021-11-03
`nix repl`'s `:e` command breaks stdout for editors


trofi:
The following command launches tty editor:
```
$ EDITOR=nano nix edit -f. nano
```

The following command shows no output for editor:
```
$ EDITOR=nano nix repl .
nix-repl> :e nano
# no editor output
```

I suspect `nix repl` does not restore some terminal state on `:e` and confuses editors.

**`nix-env --version` output**
nix-env (Nix) 2.5pre20211007_844dd90

/cc @edolstra

cole-h:
I've also seen this with `kakoune`, but it just causes kakoune to exit because "stdout is not a tty": https://github.com/mawww/kakoune/blob/a70dbda54c16f327d6f6e7fa5573e3751f7fda47/src/terminal_ui.cc#L458.

PR #5586, opened 2021-11-17
Fix :e in repl

mopleen:
Closes https://github.com/NixOS/nix/issues/5487

Co-authored-by: Alexander Bantyev balsoft@balsoft.ru

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -471,7 +471,10 @@ bool NixRepl::processLine(string line)
         auto args = editorFor(pos);
         auto editor = args.front();
         args.pop_front();
-        runProgram(editor, true, args);
+
+        // runProgram redirects stdout to a StringSink,
+        // using runProgram2 to allow editors to display their UI
+        runProgram2(RunOptions { .program = editor, .searchPath = true, .args = args });
 
         // Reload right after exiting the editor
         state->resetFileCache();


src/nix/repl.cc
```diff
@@ -471,7 +471,14 @@ bool NixRepl::processLine(string line)
...
         auto editor = args.front();
         args.pop_front();
-        runProgram(editor, true, args);
+
+        try {
+            // runProgram redirects stdout to a StringSink,
+            // using runProgram2 to allow editors to display their UI
+            runProgram2(RunOptions { .program = editor, .searchPath = true, .args = args });
+        } catch (ExecError & e) {
+            // Ignore errors from editor
```
regnat, line 480: You probably don’t want to catch the error here. It’s caught somewhere above in the stack so that it doesn’t crash the repl but it just printed (that can probably help debugging issues caused by a bad `$EDITOR`, and that’s what `runProgram` was emulating too):

```console
$ EDITOR=false nix repl '<nixpkgs>'
Welcome to Nix 2.5pre20211105_3d6ee22. Type :? for help.

Loading '<nixpkgs>'...
Added 15309 variables.

nix-repl> :e hello
error: program 'false' failed with exit code 1

nix-repl>
```

regnat (changes requested):

src/nix/repl.cc
mopleen: Thanks, good catch!

PR #5589, opened 2021-11-17
Push source into `nix/` subdir to avoid annoying extra rebuilds

### LINKED ISSUE #5574: Nix's build system should filterSource or similar
Right now, changing `flake.nix` etc. in any way rebuild everything. These mass rebuilds make tweaking those files really annoying. It would be nice if we could filter those files out from the `src` of building Nix, so that this no longer happens.

Ericson2314:
Currently, `flake.nix`, `docker.nix`, etc. "pollute" the most of the nix build's derivations, even though these files are not need at build time. This makes end-to-end testing of builds rather annoying, too many edits are "mass rebuilds".

Rather than do lots of complicated filtering, we just moved Nix itself to be in a sub-directory, just like `perl/` for the perl bindings. This means that we can take whole subdirs, which, after the usual git-based filtering, means we get what we want.

The churn of so many files being moved might raise some eyebrows, but I this this is the right approach. Git handles "pure" renames quite well (as opposed to file splits), so I don't expect this to negatively impact other developers.

Even more importantly, we want the Nix repo to be a good example for (potential) users of Nix of best practices. I think it's important people remember not thinking fancy filter is the best/only way to do things, and I also certainly don't want them to think that extra rebuilds are the inevitable cost of purity. This avoids making both negative impressions.

The one caveat is that without making `officialRelease = true`, `versionSuffix` will end up tainting everything with the commit hash anyways. I have just been setting `officialRelease = true` and remembering not to commit it, which works but we can do better. I am not sure what the right thing to do is, and am certainly open to suggestions.

Mostly fixes #5574


edolstra:
I don't think this is a good approach. Being forced to move almost the entire source tree into a subdirectory just to work around a Nix issue is *not* best practice and not something most projects would want to do. I mean, surely we don't want to tell users that adding a flake.nix requires reorganizing their entire repository.

Ericson2314:
@edolstra this has nothing to do with flakes I now see. Proper use of Nix with or without flakes *does* require separation components into subdirectories to avoid silly rebuilds, and that is good practice anyways even without Nix to make logical boundaries clear.

Remember, this is not a single package repo, for nix itself and the perl bindings are separate packages.

Ericson2314:
On pause until a decision is reached in https://github.com/NixOS/nix/issues/7876./

Issue #972, opened 2016-07-15
"Did you mean ..." when typing wrong package name


mpscholten:
Currently when typing the wrong package name, e.g. `nix-env -i cabal` instead of `nix-env -i cabal-install` you will get `error: selector ‘cabal’ matches no derivations`. That's not really helpful.

Maybe `nix-env` could check for possible package with a similar name, and then output something like `error: selector ‘cabal’ matches no derivations. Did you mean ‘cabal-install‘?`

copumpkin:
Let's do this on attribute paths too. Attribute names are so last year :)

PR #5594, opened 2021-11-18
Offer suggestions for nix-env -i

mopleen:
This implements the first simplest iteration of a fuzzy search - match by prefix.

Looks like this:
```
user:nix alex$ ./nix-env -i hask
error: selector 'hask' matches no derivations, maybe you meant:
       haskell-ci
       haskell-language-server
       hasklig
```

Closes https://github.com/NixOS/nix/issues/972

### DIFF
--- a/src/nix-env/nix-env.cc
+++ b/src/nix-env/nix-env.cc
@@ -224,6 +224,91 @@ static void checkSelectorUse(DrvNames & selectors)
 }
 
 
+namespace {
+
+std::set<std::string> searchByPrefix(const DrvInfos & allElems, std::string_view prefix) {
+    constexpr std::size_t maxResults = 3;
+    std::set<std::string> result;
+    for (const auto & drvInfo : allElems) {
+        const auto drvName = DrvName { drvInfo.queryName() };
+        if (hasPrefix(drvName.name, prefix)) {
+            result.emplace(drvName.name);
+
+            if (result.size() >= maxResults) {
+                break;
+            }
+        }
+    }
+    return result;
+}
+
+struct Match
+{
+    DrvInfo drvInfo;
+    std::size_t index;
+
+    Match(DrvInfo drvInfo_, std::size_t index_)
+    : drvInfo{std::move(drvInfo_)}
+    , index{index_}
+    {}
+};
+
+/* If a selector matches multiple derivations
+   with the same name, pick the one matching the current
+   system.  If there are still multiple derivations, pick the
+   one with the highest priority.  If there are still multiple
+   derivations, pick the one with the highest version.
+   Finally, if there are still multiple derivations,
+   arbitrarily pick the first one. */
+std::vector<Match> pickNewestOnly(EvalState & state, std::vector<Match> matches) {
+    /* Map from package names to derivations. */
+    std::map<std::string, Match> newest;
+    StringSet multiple;
+
+    for (auto & match : matches) {
+        auto & oneDrv = match.drvInfo;
+
+        const auto drvName = DrvName { oneDrv.queryName() };
+        long comparison = 1;
+
+        const auto itOther = newest.find(drvName.name);
+
+        if (itOther != newest.end()) {
+            auto & newestDrv = itOther->second.drvInfo;
+
+            comparison =
+                oneDrv.querySystem() == newestDrv.querySystem() ? 0 :
+                oneDrv.querySystem() == settings.thisSystem ? 1 :
+                newestDrv.querySystem() == settings.thisSystem ? -1 : 0;
+            if (comparison == 0)
+                comparison = comparePriorities(state, oneDrv, newestDrv);
+            if (comparison == 0)
+                comparison = compareVersions(drvName.version, DrvName { newestDrv.queryName() }.version);
+        }
+
+        if (comparison > 0) {
+            newest.erase(drvName.name);
+            newest.emplace(drvName.name, match);
+            multiple.erase(drvName.fullName);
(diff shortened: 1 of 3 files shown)


src/nix-env/nix-env.cc
```diff
@@ -224,6 +224,100 @@ static void checkSelectorUse(DrvNames & selectors)
 }
 
 
+namespace {
+
+template <class Container, class F>
+void forEachWithIndex(Container & container, F && func) {
```
edolstra, line 230: Check out `enumerate()` in util.hh, it allows you to do 
```
for (const auto & [i, item] : enumerate(container)) { ... }
```

src/nix-env/nix-env.cc
```diff
@@ -224,6 +224,100 @@ static void checkSelectorUse(DrvNames & selectors)
...
+    std::size_t index = 0;
+    for (auto & item : container) {
+        func(item, index);
+        ++index;
+    }
+}
+
+std::set<std::string> fuzzySearch(const DrvInfos & allElems, const DrvName & name) {
```
edolstra, line 238: Maybe "fuzzy" is not a great name since it's just prefix matching. But if we had a fuzzy search (like using Levenshtein distance) it would be great to move this function into `libutil` so it could be used for other things like attribute/variable name mismatches.

src/nix-env/nix-env.cc
```diff
@@ -224,6 +224,100 @@ static void checkSelectorUse(DrvNames & selectors)
...
+    }
+    return result;
+}
+
+struct Match
+{
+    DrvInfo drvInfo;
+    std::size_t index;
+
+    Match(DrvInfo drvInfo_, std::size_t index_)
```
edolstra, line 259: IMHO we don't need trivial constructors like this in C++20 anymore since we can use designated initializers, i.e.
```
Match { .drvInfo = ..., .index = ... }
```
which are easier to read.

mopleen: True but in this case I was aiming to enable the use of `emplace`

src/nix-env/nix-env.cc
mopleen: True. Renamed to `searchByPrefix` to better reflect the current situation. Real fuzzy search would be nicer but I guess it could be left to be improved at a later stage

src/nix-env/nix-env.cc
edolstra: Not sure if that really matters. But for efficiency you probably want to declare `DrvInfo && drvInfo` because otherwise it's passed by value which is probably not what you want.

src/nix-env/nix-env.cc
Ericson2314: Factor out, then improve, is always good! :)

src/nix-env/nix-env.cc
mopleen: This is actually the preferred way to pass for sink-like semantics, it is optimized very well. If I use && then only rvalues would be accepted, which is an unnecessary restriction for this usecase.

https://godbolt.org/z/KvsdWh13d
Flip 1 on line 20 to make sure the generated code is the same

src/nix-env/nix-env.cc
```diff
@@ -224,6 +224,91 @@ static void checkSelectorUse(DrvNames & selectors)
...
+            multiple.insert(drvName.fullName);
+        }
+    }
+
+    matches.clear();
+    for (auto & [name, match] : newest) {
+        if (multiple.find(name) != multiple.end())
+            printInfo(
+                "warning: there are multiple derivations named '%1%'; using the first one",
+                name);
```
regnat, line 303:
```suggestion
            warn(
                "there are multiple derivations named '%1%'; using the first one",
                name);
```

src/nix-env/nix-env.cc
```diff
@@ -232,80 +317,43 @@ static DrvInfos filterBySelector(EvalState & state, const DrvInfos & allElems,
...
+            if (prefixHits.empty()) {
+                throw Error("selector '%1%' matches no derivations", selector.fullName);    
+            } else {
+                std::string suggestionMessage = ", maybe you meant:";
+                for (const auto & drvName : prefixHits) {
+                    suggestionMessage += fmt("\n%s", drvName);
+                }
+                throw Error("selector '%1%' matches no derivations" + suggestionMessage, selector.fullName);
+            }
+        }
```
regnat, line 354: Could this also be used at https://github.com/NixOS/nix/blob/4b28798bfcb659bf83557a33546a03115f8bfed2/src/nix-env/nix-env.cc#L706-L711 rather than calling `checkSelectorUse` at the end?

regnat (changes requested):
Two small comments, but looks good overall!

src/nix-env/nix-env.cc
mopleen: Maybe. I'm not sure what's happening in `opSetFlag` though

regnat: I’m not 100% sure either. But it doesn’t matter that much anyways, it’s not a very widely used thing so we can live without suggestions

regnat (approved):

fzakaria:
I agree prefix matching is a bit simple especially to call it "fuzzy" --

How about using something similar to Git which does levenshtein distance (a string metric for measuring the difference between two sequences.)
https://github.com/git/git/blob/f443b226ca681d87a3a31e245a70e6bc2769123c/help.c#L579

Plenty of implementations to use (according to licenses) available.

PR #5598, opened 2021-11-18
Fix testing the other daemon

Ericson2314:
The eventual PATH entry needs the `.../bin` or we will not use the right daemon.


regnat (approved):
Oh my :man_facepalming: 

Thanks :)

PR #5602, opened 2021-11-19
Allow testing `ssh://` and `ssh-ng://` across versions too

Ericson2314:
We want to make sure both protocols don't break. Also recall `ssh-ng://` will be slightly different than `unix://` because the former transfers files over the protocol too.

I am not sure what the status of adding new configuration env vars is. I hope if `NIX_DAEMON_SOCKET` path was OK, these too are too, but if not I suppose we could try to sneak and shove some `?remote-command=...` into the tests.

~~depends on #5598~~

The dependency is mainly to avoid merge conflicts.

### DIFF
--- a/src/libstore/legacy-ssh-store.cc
+++ b/src/libstore/legacy-ssh-store.cc
@@ -17,7 +17,10 @@ struct LegacySSHStoreConfig : virtual CommonSSHStoreConfig
 {
     using CommonSSHStoreConfig::CommonSSHStoreConfig;
 
-    const Setting<Path> remoteProgram{(StoreConfig*) this, "nix-store", "remote-program",
+    const Setting<Path> remoteProgram{
+        (StoreConfig*) this,
+        getEnv("_NIX_TEST_LEGACY_SSH_REMOTE_PROGRAM").value_or("nix-store"),
+        "remote-program",
         "Path to the `nix-store` executable on the remote machine."};
 
     const Setting<int> maxConnections{(StoreConfig*) this, 1, "max-connections",
--- a/src/libstore/ssh-store.cc
+++ b/src/libstore/ssh-store.cc
@@ -16,7 +16,10 @@ struct SSHStoreConfig : virtual RemoteStoreConfig, virtual CommonSSHStoreConfig
     using RemoteStoreConfig::RemoteStoreConfig;
     using CommonSSHStoreConfig::CommonSSHStoreConfig;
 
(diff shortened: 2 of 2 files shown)


src/libstore/legacy-ssh-store.cc
```diff
@@ -18,7 +18,12 @@ struct LegacySSHStoreConfig : virtual StoreConfig
     const Setting<Path> sshKey{(StoreConfig*) this, "", "ssh-key", "path to an SSH private key"};
     const Setting<std::string> sshPublicHostKey{(StoreConfig*) this, "", "base64-ssh-public-host-key", "The public half of the host's SSH key"};
     const Setting<bool> compress{(StoreConfig*) this, false, "compress", "whether to compress the connection"};
-    const Setting<Path> remoteProgram{(StoreConfig*) this, "nix-store", "remote-program", "path to the nix-store executable on the remote system"};
+    const Setting<Path> remoteProgram{
+        (StoreConfig*) this,
+        getEnv("NIX_LEGACY_SSH_REMOTE_PROGRAM").value_or("nix-store"),
```
edolstra, line 23: There is a bit of a convention to prefix these variable with `_NIX_TEST_`. Alternatively, we can avoid the environment variable by appending `?remote-program=...` to the store URL.

Ericson2314: I switched it to `_NIX_TEST_` instead. As to `?remote-program=...` I can pursue if you want, but do note we have some fancy ssh urls e.g. doing iterative ssh with `?remote-store=` so it is a bit a non-trivial. I am thinking the env var might be better because it means no one can forget the remote program part going forward if they are adding another ssh test, especially if it is also complicated.

Ericson2314:
Is this OK now? Would be nice to merge.

thufschmitt (approved):

regnat:
Actually while this is OK, it’s a bit limited in what it can test, because when `NIX_DAEMON_PACKAGE` is set, the daemon is systematically used, meaning that the test isn’t really between versions since the daemon and remote builder versions will be the same.
Maybe we could do it the other way around by setting the remote programs to the client version.

Ericson2314:
Ah good point. Maybe let's start with this, just because it's not worse than today?

Ericson2314:
@andir an example of one of my PRs relating to cross version testing. It might be nice to do call to page it all back in and answer any questions you have.

Ericson2314:
Making this a draft until there are some actual tests to go with it.

tomberek:

PR #5603, opened 2021-11-19
"recursive-nix" system feature only with experimental feature

Ericson2314:
I worry this patch won't work as written because the order of
initialization is not what we need, or even the same in all situations.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -122,7 +122,7 @@ StringSet Settings::getDefaultSystemFeatures()
     /* For backwards compatibility, accept some "features" that are
        used in Nixpkgs to route builds to certain machines but don't
        actually require anything special on the machines. */
-    StringSet features{"nixos-test", "benchmark", "big-parallel", "recursive-nix"};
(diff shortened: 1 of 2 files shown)


Ericson2314:
I thought this was attempted before, but I couldn't find a PR/issue for it.

src/libstore/globals.cc
```diff
@@ -122,7 +122,10 @@ StringSet Settings::getDefaultSystemFeatures()
     /* For backwards compatibility, accept some "features" that are
        used in Nixpkgs to route builds to certain machines but don't
        actually require anything special on the machines. */
-    StringSet features{"nixos-test", "benchmark", "big-parallel", "recursive-nix"};
+    StringSet features{"nixos-test", "benchmark", "big-parallel"};
+
+    if (requireExperimentalFeature(Xp::CaDerivations))
```
regnat, line 127:
```suggestion
    if (isExperimentalFeatureEnabled(Xp::RecursiveNix))
```

Ericson2314:
Oh, I realized we can do this like you did the `ca-derivations` feature.

regnat (approved):

PR #5620, opened 2021-11-21
nix-shell --pure: let variables for Wayland through

alyssais:
We let `DISPLAY` (X11) through, so we should let the Wayland equivalents through as well.  Similarly, we let `HOME` through, so it should be okay to allow `XDG_RUNTIME_DIR` (which is needed for connecting to Wayland with `WAYLAND_DISPLAY`) through as well.  Otherwise graphical applications will either fall back to X11 (if they support it), or just not work (if they don't).


regnat (approved):
Looks pretty sensible indeed, thanks

Issue #5627, opened 2021-11-22
Builtin lexical list comparison


Infinisil:
**Is your feature request related to a problem? Please describe.**
Currently lists aren't comparable:
```
nix-repl> [] < []
error: cannot compare a list with a list
```

**Describe the solution you'd like**
Be able to compare lists using lexicographic ordering, where shorter lists are ordered before longer ones, and otherwise the elements are compared in order:
```
nix-repl> [] < []
false
nix-repl> [] < [ 0 ]
true
nix-repl> [ 0 ] < [ 1 ]
true
```

**Describe alternatives you've considered**
This can be implemented in Nix itself, but it's clunky and not very efficient:
```nix
{
  lexicalLessThan = a: b:
    let
      alen = lib.length a;
      compareElements = index:
        if index == alen then false
        else if lib.elemAt a index == lib.elemAt b index then compareElements (index + 1)
        else lib.elemAt a index < lib.elemAt b index;
    in
    if alen == lib.length b then compareElements 0
    else alen < lib.length b;
}
```

**Additional context**
Implementing this is backwards compatible, but probably the `builtins.langVersion` should be increased for this change to be detectable.

regnat:
I’m curious what the use-case for this is (esp. since as you say it’s implementable as a library).
The main reason I generally encounter for wanting some ordering on this kind of not-obviously-ordered data-structures is to have them as keys in (ordered) maps, but I don’t think you’ll want to implement a custom map in Nix anyways ;)

Infinisil:
The PR is WIP, but it can be used for https://github.com/NixOS/nixpkgs/pull/147077. I'll document how this is used there

Infinisil:
Now done with https://github.com/NixOS/nix/pull/5631!

PR #5631, opened 2021-11-22
Make lists be comparable

Infinisil:
Makes lists comparable using lexicographic comparison.

Increments builtins.langVersion in order for this change to be
detectable

Implements https://github.com/NixOS/nix/issues/5627


Infinisil:
Actually, there's two different ways of doing lexicographic ordering, and I'm not sure which one makes more sense:
1. Smaller lists are always earlier, with strings: "a", "b", "ab", "bb"
2. The first letter decides the order, with strings: "a", "ab", "b", "bb"

This PR currently implements 1., but I think 2. makes more sense, partly because it's also how strings are ordered in Nix, and also the second one is what I need for my use-case.

Infinisil:
Also 2 is used by Haskell

edolstra:
I think lexicographic (i.e. dictionary) ordering almost always has meaning 2. For instance, it's what `operator <` on `std::vector` does in C++. The only place I've seen meaning 1 is in the ordering of keys in canonical CBOR...

Infinisil:
I updated this PR to use number 2 instead. Apparently both versions can be called lexicographic order, see [Wikipedia](https://en.wikipedia.org/wiki/Lexicographic_order), but number 2 is the usual one.

edolstra:
Can you add this to the release notes and possibly update the docs for the `<` operator?

src/libexpr/primops.cc
```diff
@@ -3103,7 +3119,7 @@ static void prim_lessThan(EvalState & state, const Pos & pos, Value * * args, Va
 {
     state.forceValue(*args[0], pos);
     state.forceValue(*args[1], pos);
-    CompareValues comp;
+    CompareValues comp = CompareValues(state);
```
edolstra, line 3122:
```suggestion
    CompareValues comp{state};
```

Infinisil:
@edolstra Added a release note, expanded the description of the `<` to mention lexicographic comparison, and applied your suggestions

PR #5632, opened 2021-11-23
Make docker image downloadable in Hydra UI

garbas:


flake.nix
```diff
@@ -407,10 +407,18 @@
...
-          });
+            image = import ./docker.nix { inherit pkgs; tag = version; };
+          in pkgs.runCommand "docker-image-tarball-${version}"
+            { meta.description = "Docker image with Nix for ${system}";
+            }
+            ''
+              mkdir -p $out/nix-support
+              image=$out/image.tar.gz
+              cp ${image} $image
+              echo "file binary-dist $image" >> $out/nix-support/hydra-build-products
```
edolstra, line 420:
```suggestion
              echo "file binary-dist ${image}" >> $out/nix-support/hydra-build-products
```

Copying is expensive (especially for large images) since it results in two copies in the binary cache.

garbas: Tnx, totally makes sense.

I did however did it a little differently, just so I could later use this hydraJob to push the image to docker hub.

Infinisil:
This broke CI on master

PR #5599, opened 2021-11-18
Fix heap use after free in progress-bar.cc

fzakaria:
Fix some heap-use-after-free in progress-bar.cc

These are somewhat tricky failures here due to temporary variable creation and _string_view_

I've included links to the fixes with more information to inform other developers to be wary.

Here was the failure:
```
[nix-shell:~/code/github.com/NixOS/nix]$ ./src/nix/nix build -f ~/code/playground/nix/concurrent.nix 
trace: Warning: `stdenv.lib` is deprecated and will be removed in the next release. Please use `lib` instead. For more information see https://github.com/NixOS/nixpkgs/issues/108938
==2023391==WARNING: ASan is ignoring requested __asan_handle_no_return: stack top: 0x7ffd7376c000; bottom 0x7ff604901000; size: 0x00076ee6b000 (31925383168)
False positive error reports may follow
For details see https://github.com/google/sanitizers/issues/189
querying info about missing paths=================================================================
==2023391==ERROR: AddressSanitizer: heap-use-after-free on address 0x606000d3d4ad at pc 0x00000067893e bp 0x7ffd73767570 sp 0x7ffd73766d20
READ of size 6 at 0x606000d3d4ad thread T0
    #0 0x67893d in __interceptor_memcpy.part.0 (/home/fzakaria/code/github.com/NixOS/nix/src/nix/nix+0x67893d)
    #1 0x7ff62ab22057 in std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) (/home/fzakaria/code/github.com/NixOS/nix/src/libutil/libnixutil.so+0x4da057)
    #2 0x7ff62ab133b3 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) (/home/fzakaria/code/github.com/NixOS/nix/src/libutil/libnixutil.so+0x4cb3b3)
    #3 0x83bf37 in std::basic_ostream<char, std::char_traits<char> >& std::operator<<<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, std::basic_string_view<char, std::char_traits<char> >) /nix/store/sjhz1j2d1ssn59f66kqp92xj9mpsww2d-gcc-10.3.0/include/c++/10.3.0/string_view:621:14
[...]

PR #5636, opened 2021-11-23
Fix use after free in content-address.cc

mopleen:
Inspired by https://github.com/NixOS/nix/pull/5599

P.S. std::string_view is very sneaky!

### DIFF
--- a/src/libstore/content-address.cc
+++ b/src/libstore/content-address.cc
@@ -120,8 +120,10 @@ ContentAddress parseContentAddress(std::string_view rawCa) {
 
 ContentAddressMethod parseContentAddressMethod(std::string_view caMethod)
 {
-    std::string_view asPrefix {std::string{caMethod} + ":"};
-    return parseContentAddressMethodPrefix(asPrefix);
+    std::string asPrefix = std::string{caMethod} + ":";
+    // parseContentAddressMethodPrefix takes its argument by reference
+    std::string_view asPrefixView = asPrefix;
+    return parseContentAddressMethodPrefix(asPrefixView);
 }
 
 std::optional<ContentAddress> parseContentAddressOpt(std::string_view rawCaOpt)


src/libstore/content-address.cc
```diff
@@ -120,8 +120,10 @@ ContentAddress parseContentAddress(std::string_view rawCa) {
 
 ContentAddressMethod parseContentAddressMethod(std::string_view caMethod)
 {
-    std::string_view asPrefix {std::string{caMethod} + ":"};
-    return parseContentAddressMethodPrefix(asPrefix);
+    std::string asPrefix = std::string{caMethod} + ":";
+    // parseContentAddressMethodPrefix takes its argument by reference
+    std::string_view asPrefixView = asPrefix;
+    return parseContentAddressMethodPrefix(asPrefixView);
```
regnat, line 126: Does this present any advantage over

```suggestion
    return parseContentAddressMethodPrefix(asPrefix);
```

?

regnat (approved):
Thanks :)

(I do have one small question, but more for curiosity than anything else, so it doesn’t really matter)

src/libstore/content-address.cc
mopleen: It just wouldn't work, because parseContentAddressMethodPrefix takes its argument as (std::string_view &) and & won't bind to a temporary

regnat:
Ahah, missed the “&”!

Thanks for the explanation :)

fzakaria:
Nice.
Without ASan checking that the _string_view_ isn't stored anywhere when constructing the ContentAddressMethod is so tricky.

src/libstore/content-address.cc
edolstra: `std::string_view &` should be changed to `std::string_view` since there is no point in passing a `std::string_view` by reference (it's just a 2-word tuple internally). And then what @regnat suggests should work.

mopleen: Ideally, yes, would be nice to take it by value. But it seems that other code is doing something modifying the argument, I'm not very confident about what's going on there. I think, it leaves the non-consumed part in there.

roberth:
@edolstra Could this be backported? It is used by nix-daemon `wopAddToStore` in non-experimental code.

edolstra:
To 2.4? I think 2.5 already has it.

PR #5639, opened 2021-11-23
references.cc: clarify search optimization

milahu:
improve code readability

### DIFF
--- a/src/libutil/references.cc
+++ b/src/libutil/references.cc
@@ -31,6 +31,9 @@ static void search(
             isBase32[(unsigned char) base32Chars[i]] = true;
     });
 
+    /* Nested loops, but the inner loop also advances i, so
+       this is an asymptotically optimal solution for finding
+       hash-like substrings. */
     for (size_t i = 0; i + refLength <= s.size(); ) {
         /* Check whether the next `refLength` bytes are a possible
            reference (because they are all base-32 characters). */
--- a/src/libutil/references.cc
+++ b/src/libutil/references.cc
@@ -12,7 +12,10 @@
 namespace nix {
 
 
-static size_t refLength = 32; /* characters */
+/**
+ * A Nix reference is 32 characters long
+ */
+constexpr size_t refLength = 32;
 
 
 static void search(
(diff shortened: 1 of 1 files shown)


src/libstore/references.cc
```diff
@@ -32,7 +32,7 @@ static void search(
         bool match = true;
         for (j = refLength - 1; j >= 0; --j)
             if (!isBase32[(unsigned char) s[i + j]]) {
-                i += j + 1;
+                i += j + 1; // skip refLength bytes
```
milahu, line 35: more precise: `seek to next block`

Ericson2314:
The original comment didn't really help me, so I added some more.

src/libutil/references.cc
```diff
@@ -12,7 +12,10 @@
 namespace nix {
 
 
-static size_t refLength = 32; /* characters */
+/**
+ * A Nix reference is 32 characters long
```
roberth, line 16: From what I know a "Nix"(?) reference is a relation between store objects, so this still makes no sense.
Is this about the store path hash part?

OT: why isn't this in libstore if it's about store concepts?

src/libutil/references.cc
```diff
@@ -29,15 +32,21 @@ static void search(
     });
 
     for (size_t i = 0; i + refLength <= s.size(); ) {
```
roberth, line 34:
```suggestion
    /* Nested loops, but the inner loop also advances i, so
       this is an asymptotically optimal solution for finding
       hash-like substrings. */
    for (size_t i = 0; i + refLength <= s.size(); ) {
```

It screams mutual recursion, but I'll ignore that.

src/libutil/references.cc
```diff
@@ -29,15 +32,21 @@ static void search(
...
+           reference (because they are all base-32 characters). */
+        bool possibleReference = true;
+        for (size_t j = refLength; j > 0; --j)
+            if (!isBase32[(unsigned char) s[i + j - 1]]) {
+                /* We found an invalid character at index `i + j - 1`,
+                   the next possible reference (which doesn't contain
+                   it) starts at index `i + j`. */
+                i += j;
+                /* Use variable because we cannot directly continue in
+                   the outer loop in C++. */
```
roberth, line 45: Edsger Dijkstra wouldn't mind.

roberth (changes requested):

src/libutil/references.cc
Ericson2314: This happened in 3ebe1341abe1b0ad59bd4925517af18d9200f818 with my assistance. I got fooled by the types and thought this was store agnostic :(. I agree that given how the code is written today, it doesn't belong here.

Ericson2314: @roberth I say we merge this after perhaps rephrasing the above, and then I'll move it back. How does that sound?

PR #5640, opened 2021-11-24
[backport 2.3] Separate auth and logic for the daemon

Ericson2314:
Before, `processConnection` wanted to know a user name and user id, and `nix-daemon --stdio`, when it isn't proxying to an underlying daemon, would just assume "root" and 0. But `nix-daemon --stdio` (no proxying) shouldn't make guesses about who holds the other end of its standard streams.

Now `processConnection` takes an "auth hook", so `nix-daemon` can provide the appropriate policy and daemon.cc doesn't need to know or care what it is.

(cherry picked from commit 8d4162ff9e940ea9e2f97b07f3030a722695901a)

Depends on https://github.com/NixOS/nix/pull/5650
(not actually, but yes in terms of it wouldn't be useful until then.)

### DIFF
--- a/src/nix-daemon/nix-daemon.cc
+++ b/src/nix-daemon/nix-daemon.cc
@@ -743,7 +743,11 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
 
 
 static void processConnection(bool trusted,
-    const std::string & userName, uid_t userId)
+    /* Arbitrary hook to check authorization / initialize user data / whatever
+       after the protocol has been negotiated. The idea is that this function
+       and everything it calls doesn't know about this stuff, and the
+       `nix-daemon` handles that instead. */
+    std::function<void(Store &)> authHook)
 {
     MonitorFdHup monitor(from.fd);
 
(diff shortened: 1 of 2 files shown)


regnat:
Care to explain why this needs to be backported? The new test makes me believe that it’s just for the `fake ssh` feature, but I’m not sure what the link between the two is

Ericson2314:
@regnat heh a few things:

1. I backported it trying to fix cross daemon test failures, and yes it ended up being only that test that excised it. However, I found the `remote-store` and `remote-program` query params are also easy to backport (and I did that in https://github.com/NixOS/nix/pull/5650), and between all backports it will allow my newer more extensive remote building tests to work.

2. However, my comment i wrote and backported is overly narrow. The case with the `"root"` that is causing the problems is hit whiner `nix-daemon --stdio` is run, and that daemon wouldn't *itself* connect to another daemon. So it will also occurs when ssh-ing to machines that are "in single user mode".

Ericson2314:
draft until it is needed.

roberth (approved):

Ericson2314:
My comments from 4 years ago were not that clear :). The reasons to backport this are:

1. It will enable writing more tests, for cross-version testing purposes

2. For anyone still using 2.3, it fixes some genuinely weird and unexpected behavior for Nix will attempt to create profiles and GC root dirs for root with `nix-daemon --stdio`.

PR #5650, opened 2021-11-25
[backport 2.3] Allow testing with different daemons

Ericson2314:
The goal is to make something like `testNixVersions` on master which can run the install checks against two pre-built nix versions. The goal is to then use that function in a *later* version of Nix (to avoid cycle dependencies!) to check 2.3's (modernized) test suite against a 2.3 client but *newer* daemon.

Concretely what his PR is doing is cherry-picking a number of test suite improves to 2.3 from later versions that generally harden the testsuite and in particular allow for this cross-version testing.

 - [x] Cherry-pick commits 
 - [x] Revise commits to ensure test suite still runs the usual way
 - [x] Turn a "at least 2.4pre..." to "no later than 2.4pre..."
 - [x] Hand-backport the new bits in `flake.nix` to this `release.nix` so we actually run a daemon test against self.

For later:

 - [ ] Test with newer daemon marking fixing bugs and/or marking tests broken
  
~~depends on #5649~~
~~depends on #5753~~
~~depends on #5754~~
~~Depends on #9495~~

### DIFF
--- a/mk/run_test.sh
+++ b/mk/run_test.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-set -u
+set -eu -o pipefail
 
 red=""
 green=""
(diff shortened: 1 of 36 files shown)


regnat:
Couldn’t we rather use **master**’s testsuite with an old client (adding guards like the existing `requireNewerDaemonThan`)? That would remove the need for backporting anything to the testuite

Ericson2314:
@regnat the command line has changed so much, I expect that to be a big plain. Think of all those `--impure` `--expr` everywhere.

Also remember this is a 1-time cost. Going forward we already have the support we we need to e.g. test 2.4 with master daemon.

Ericson2314:
I tested macOS and Linux builds locally, but GHA is completely busted because that stuff bit-rots so fast.

Ericson2314:
OK did a few more cherry-picks so that this actually runs a daemon test. It's finally ready!

roberth (approved):

PR #5652, opened 2021-11-25
Make `nix flake show` less strict wrt what apps can be

### LINKED ISSUE #5560: Inconsistency between `nix run` and `nix flake show`
**Describe the bug**

Some inputs accepted by `nix run` are not supported by `nix flake show`

**Steps To Reproduce**

The documentation says:

> nix run builds and runs installable, which must evaluate to an app or a regular Nix derivation.
> 
> If installable evaluates to an app (see below), it executes the program specified by the app definition.
> 
> If installable evaluates to a derivation, it will try to execute the program <out>/bin/<name>, where out is the primary output store path of the derivation and name is the meta.mainProgram attribute of the derivation if it exists, and otherwise the name part of the value of the name attribute of the derivation (e.g. if name is set to hello-1.10, it will run $out/bin/hello).

Now, given the flake:

```nix
      {
        apps.x86_64-linux.helloWorld = {
          type = "app";
          program = "${nixpkgs.legacyPackages.x86_64-linux.hello}/bin/hello";
        };
        apps.x86_64-linux.helloWorld2 = nixpkgs.legacyPackages.x86_64-linux.hello;
      }
```

This works:
- `nix run .#helloWorld`
- `nix run .#helloWorld2`

However, this does not work:
- `nix flake show`

```
git+file:///example
└───apps
    └───x86_64-linux
        ├───helloWorld: app
error: not an app definition
```

**Expected behavior**

`nix flake show` is consistent with `nix run`

nix-env (Nix) 2.5pre20211007_844dd90

**Additional context**

thufschmitt:
Allow `apps.foo` and `defaultApp` to be a plain derivation like `nix run` does.

Fix #5560

cc @kamadorueda

### DIFF
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -1024,13 +1024,17 @@ struct CmdFlakeShow : FlakeCommand, MixJSON
                     (attrPath.size() == 2 && attrPath[0] == "defaultApp") ||
                     (attrPath.size() == 3 && attrPath[0] == "apps"))
                 {
-                    auto aType = visitor.maybeGetAttr("type");
-                    if (!aType || aType->getString() != "app")
-                        throw EvalError("not an app definition");
-                    if (json) {
-                        j.emplace("type", "app");
-                    } else {
-                        logger->cout("%s: app", headerPrefix);
+                    if (visitor.isDerivation())
+                        showDerivation();
+                    else {
+                        auto aType = visitor.maybeGetAttr("type");
+                        if (!aType || aType->getString() != "app")
+                            throw EvalError("not an app definition");
+                        if (json) {
+                            j.emplace("type", "app");
+                        } else {
+                            logger->cout("%s: app", headerPrefix);
+                        }
                     }
                 }
 
--- a/tests/flakes.sh
+++ b/tests/flakes.sh
@@ -380,6 +380,11 @@ cat > $templatesDir/trivial/flake.nix <<EOF
   outputs = { self, nixpkgs }: {
     packages.x86_64-linux.hello = nixpkgs.legacyPackages.x86_64-linux.hello;
     defaultPackage.x86_64-linux = self.packages.x86_64-linux.hello;
+    apps.x86_64-linux.helloWorld = {
+        type = "app";
+        program = "\${nixpkgs.legacyPackages.x86_64-linux.hello}/bin/hello";
+    };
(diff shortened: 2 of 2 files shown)


edolstra:
Maybe the fix for this should be the other way around, i.e. commands like `nix run` should enforce that apps actually are apps. It seems better not to overload the types of flake outputs like `apps`.

regnat:
> Maybe the fix for this should be the other way around, i.e. commands like `nix run` should enforce that apps actually are apps. It seems better not to overload the types of flake outputs like `apps`.

That sounds sensible indeed. I think we’d still want `nix run` to be able to fallback to running a derivation from `packages` or `legacyPackages` (`nix run nixpkgs#hello` is so insanely practical), but that doesn’t prevent making the schema of `apps` more strict

tomberek:
To clarify. The proposal is to ensure that
1) a flake output defined as an app.<system>.<name> is actually an app structure `{type="app";program=path;}`
2) a flake output defined as a package.<system>.<name> is a normal derivation, not an app

tomberek:
Possible alternative by checking using toApp. https://github.com/flox/nix/commit/b1b36d4f942d0fdea36b96d76440c5e3d6526867

It's not ideal as we're just checking to ensure the attrPath has or doesn't have an `app.` prefix. Does this meet the intent?

```
{
  outputs = {self, nixpkgs, ...}: {
    packages.x86_64-linux.PkgAsPkg = nixpkgs.legacyPackages.x86_64-linux.hello; # Should be Good
    packages.x86_64-linux.AppAsPkg = self.apps.x86_64-linux.AppAsApp;           # Should be Bad

    apps.x86_64-linux.AppAsApp = {                                              # Should be Good
      type = "app";
      program = "${self.packages.x86_64-linux.PkgAsPkg}/bin/hello";
    };
    apps.x86_64-linux.PkgAsApp = self.packages.x86_64-linux.PkgAsPkg;           # Should be Bad

  };
}
```

thufschmitt:
> Possible alternative by checking using toApp. [flox@b1b36d4](https://github.com/flox/nix/commit/b1b36d4f942d0fdea36b96d76440c5e3d6526867)

I’d say that’s good-enough :) Just need to make sure that there’s a nice error message if the user specifies a plain derivation under `apps, but other than that it’s fine

tomberek:
Suggestions from NixUX:
- [ ] better error message suggestion the correct fix
- [ ] tests
- [ ] release notes

thufschmitt:
Closing in favor of #6389

Issue #3504, opened 2020-04-16
nix repl :b - looking for a way to display full log


reactormonk:
When I'm building a derivation from the repl via `:b deriv`, the repl truncates the output, which makes debugging said derivation failure impossible. Is there a way to get access to that information, preferably from within the repl?

edolstra:
There are two things we can do:

* Add a `:log <deriv>` command that invokes `nix log` to show the build log.
* Merge the progress indicator PR so you can interactively toggle whether to show logs for in-progress builds.

nikp123:
This may be an inappropriate place to ask, but can we get a log that follows the build process when invoking ``:b`` in repl or something similar.

PR #5666, opened 2021-11-26
nix repl: add :log

balsoft:
- RunPager: restore stdout upon pager exit
- nix repl: add :log

Closes https://github.com/NixOS/nix/issues/3504

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -427,7 +427,7 @@ RunPager::RunPager()
     });
 
     pid.setKillSignal(SIGINT);
-
+    stdout = fcntl(STDOUT_FILENO, F_DUPFD_CLOEXEC, 0);
     if (dup2(toPager.writeSide.get(), STDOUT_FILENO) == -1)
         throw SysError("dupping stdout");
 }
@@ -438,7 +438,7 @@ RunPager::~RunPager()
     try {
         if (pid != -1) {
             std::cout.flush();
-            close(STDOUT_FILENO);
+            dup2(stdout, STDOUT_FILENO);
             pid.wait();
         }
     } catch (...) {
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -430,7 +430,8 @@ bool NixRepl::processLine(string line)
              << "  :s <expr>     Build dependencies of derivation, then start nix-shell\n"
              << "  :t <expr>     Describe result of evaluation\n"
              << "  :u <expr>     Build derivation, then start nix-shell\n"
-             << "  :doc <expr>   Show documentation of a builtin function\n";
+             << "  :doc <expr>   Show documentation of a builtin function\n"
+             << "  :log <expr>   Show logs for a derivation\n";
     }
 
     else if (command == ":a" || command == ":add") {
@@ -500,7 +501,7 @@ bool NixRepl::processLine(string line)
         runNix("nix-shell", {state->store->printStorePath(drvPath)});
     }
 
-    else if (command == ":b" || command == ":i" || command == ":s") {
+    else if (command == ":b" || command == ":i" || command == ":s" || command == ":log") {
         Value v;
         evalString(arg, v);
         StorePath drvPath = getDerivationPath(v);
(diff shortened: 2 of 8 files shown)


src/nix/repl.cc
```diff
@@ -514,6 +516,29 @@ bool NixRepl::processLine(string line)
...
+                auto log = sub->getBuildLog(drvPath);
+                if (!log) continue;
+                stopProgressBar();
+                printInfo("got build log for '%s' from '%s'", drvPathRaw, sub->getUri());
+                std::cout << *log;
+                foundLog = true;
+                break;
+            }
+
+            if (!foundLog) throw Error("build log of '%s' is not available", drvPathRaw);
```
Radvendii, line 541: I haven't tested this exact code, but something like:
```suggestion
            std::shared_ptr<std::string> log = null;
            for (auto & sub : subs) {
                log = sub->getBuildLog(drvPath);
                if (log) break;
            }
            
            if (log) {
                RunPager pager;
                stopProgressBar();
                printInfo("got build log for '%s' from '%s'", drvPathRaw, sub->getUri());
                std::cout << *log;
            }
            else throw Error("build log of '%s' is not available", drvPathRaw);
```

balsoft: Actually, since we're using the `sub` to print the `got build log` message, this makes it more complicated and messy. I think the original is fine for now.

Radvendii: Oh good call. A more minor tweak then maybe:
```suggestion
            bool foundLog = false;

            RunPager pager;
            for (auto & sub : subs) {
                auto log = sub->getBuildLog(drvPath);
                if (log) {
                    stopProgressBar();
                    printInfo("got build log for '%s' from '%s'", drvPathRaw, sub->getUri());
                    std::cout << *log;
                    foundLog = true;
                    break;
                }
            }

            if (!foundLog) throw Error("build log of '%s' is not available", drvPathRaw);
```

src/libmain/shared.cc
```diff
@@ -427,7 +427,7 @@ RunPager::RunPager()
     });
 
     pid.setKillSignal(SIGINT);
-
+    stdout = dup(STDOUT_FILENO);
```
edolstra, line 430: I think this will cause us to leak a file descriptor if we don't mark it as CLOEXEC, see https://github.com/NixOS/nix/pull/5683. On Linux, `dup3` can be used to atomically create this FD as CLOEXEC.

src/nix/repl.cc
```diff
@@ -514,6 +516,28 @@ bool NixRepl::processLine(string line)
...
+            auto subs = getDefaultSubstituters();
+
+            subs.push_front(state->store);
+
+            bool foundLog = false;
+            RunPager pager;
+            for (auto & sub : subs) {
+                auto log = sub->getBuildLog(drvPath);
+                if (log) {
+                    stopProgressBar();
```
edolstra, line 533: Hm, why do we need to stop the progress bar?

src/nix/repl.cc
```diff
@@ -514,6 +516,28 @@ bool NixRepl::processLine(string line)
...
+                    printInfo("got build log for '%s' from '%s'", drvPathRaw, sub->getUri());
+                    std::cout << *log;
```
edolstra, line 535: This should probably use `logger->writeToStdout()`.

src/nix/repl.cc
balsoft: This is done in `nix/log.cc` too, which is where it was copied from. Not doing so seems to leave the progress bar there after the operation is finished.

src/libmain/shared.cc
balsoft: I think the better option is the `fcntl` one I have pushed. Not sure it works on darwin or any other platform though, so maybe I need to hind it behind `#ifdef __linux__`

src/nix/repl.cc
```diff
@@ -33,6 +33,7 @@ extern "C" {
 #include "command.hh"
 #include "finally.hh"
 #include "markdown.hh"
+#include "progress-bar.hh"
```
regnat, line 36: I guess you don’t need this anymore

PR #5668, opened 2021-11-26
reproducibility: determinstic man page output for cores

tomberek:
Man page output depends on the build machine.

Fixes: https://r13y.com/diff/bceec81ab96da678ea6855d55af19840fba8d158daa4e4da618bb7896789bdfa-000c650227704ff22fee3a885db90cd5ac3ba19f68da841cbfd85065051a90a7.html

20211126_10h13m16s_grim

Note: Created during NixUX Friday Extravaganza @regnat @balsoft @Radvendii

### DIFF
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -12,11 +12,13 @@ man-pages := $(foreach n, \
 clean-files += $(d)/*.1 $(d)/*.5 $(d)/*.8
 
 # Provide a dummy environment for nix, so that it will not access files outside the macOS sandbox.
+# Set cores to 0 because otherwise nix show-config resolves the cores based on the current machine
 dummy-env = env -i \
 	HOME=/dummy \
 	NIX_CONF_DIR=/dummy \
 	NIX_SSL_CERT_FILE=/dummy/no-ca-bundle.crt \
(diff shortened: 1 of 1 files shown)


balsoft (approved):

SuperSandro2000:
Do I understand this correct that now the manual would show that the default value for cores is 0? I don't think that would be the right approach. Why not rephrase that the default value is equal to your core count or something like that?

tomberek:
@SuperSandro2000 This is the description in the text:
```
... and defaults to `1`.  The value `0` means that the builder should use all available CPU cores in the system.
```
Which doesn't seem quite right, it is `max(1,std::thread::hardware_concurrency)` in the code, which seem far closer in spirit to 0 than to say the default is 1.
And if unspecified the default used is dynamic and determined by `std::thread::hardware_concurrency` which then becomes hard-coded into the **at creation time** based on the machine running it.

Therefore the default for any nix client is not exactly the same as `0`, but that is closer to being accurate than using whatever happened on the build machine. I'm sure there is a better way to fix it, but it would require refactoring how the md/man generation is done. Suggestions are welcome.

Extra: it might be good to also update the description `and defaults to '1'` to be more accurate?

siraben (approved):

Issue #3505, opened 2020-04-16
Use position in all errors


domenkozar:
There's a lot of errors that don't include the position, which makes it hard to understand what's going on.

Lots of occurrences are easy to fix, e.g. https://github.com/NixOS/nix/pull/3502

This would also mean getting rid of `git grep noPos` and making sure all evaluation errors thrown include a `Pos`.

domenkozar:
For example, there are still calls of `forceValue` that don't include Pos:

```
$ git grep forceValue | grep -v pos
src/libexpr/attr-path.cc:        state.forceValue(*v);
src/libexpr/eval-inline.hh:    forceValue(v);
src/libexpr/eval-inline.hh:    forceValue(v);
src/libexpr/eval.cc:        state.forceValue(*vAttrs);
src/libexpr/eval.cc:    forceValue(fun);
src/libexpr/eval.cc:            forceValue(*v);
src/libexpr/eval.cc:void EvalState::forceValueDeep(Value & v)
src/libexpr/eval.cc:        forceValue(v);
src/libexpr/eval.cc:    forceValue(*i->value);
src/libexpr/eval.cc:    forceValue(v1);
src/libexpr/eval.cc:    forceValue(v2);
src/libexpr/eval.hh:    void forceValueDeep(Value & v);
src/libexpr/get-drvs.cc:    state->forceValue(v);
src/libexpr/get-drvs.cc:        state.forceValue(v);
src/libexpr/primops.cc:    state.forceValueDeep(*args[0]);
src/libexpr/primops.cc:    forceValue(v);
src/libexpr/primops/fetchGit.cc:    state.forceValue(*args[0]);
src/libexpr/primops/fetchMercurial.cc:    state.forceValue(*args[0]);
src/libexpr/primops/fetchTree.cc:    state.forceValue(*args[0]);
src/libexpr/primops/fetchTree.cc:            state.forceValue(*attr.value);
src/libexpr/primops/fetchTree.cc:    state.forceValue(*args[0]);
src/libexpr/value-to-json.cc:    if (strict) state.forceValue(v);
src/libexpr/value-to-xml.cc:    if (strict) state.forceValue(v);
src/libexpr/value-to-xml.cc:                    if (strict) state.forceValue(*a->value);
src/libexpr/value-to-xml.cc:                    if (strict) state.forceValue(*a->value);
src/nix-build/nix-build.cc:            state->forceValue(v);
src/nix-env/user-env.cc:    state.forceValue(topLevel);
[...]

PR #5676, opened 2021-11-27
forceValue: make pos mandatory

kamadorueda:
- Make passing the position to `forceValue` mandatory,
  this way we remember people that the position is
  important for better error messages
- Add pos to all `forceValue` calls

Mentions #3505

### DIFF
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -51,27 +51,11 @@ void EvalState::forceValue(Value & v, const Pos & pos)
 }
 
 
-inline void EvalState::forceAttrs(Value & v)
-{
-    forceValue(v, v.determinePos(noPos));
-    if (v.type() != nAttrs)
-        throwTypeError("value is %1% while a set was expected", v);
-}
-
-
 inline void EvalState::forceAttrs(Value & v, const Pos & pos)
 {
     forceValue(v, pos);
     if (v.type() != nAttrs)
         throwTypeError(pos, "value is %1% while a set was expected", v);
-}
-
-
-inline void EvalState::forceList(Value & v)
-{
-    forceValue(v, v.determinePos(noPos));
-    if (!v.isList())
-        throwTypeError("value is %1% while a list was expected", v);
 }
 
 
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -53,7 +53,7 @@ void EvalState::forceValue(Value & v, const Pos & pos)
 
 inline void EvalState::forceAttrs(Value & v)
 {
-    forceValue(v);
+    forceValue(v, v.determinePos(noPos));
     if (v.type() != nAttrs)
         throwTypeError("value is %1% while a set was expected", v);
 }
@@ -69,7 +69,7 @@ inline void EvalState::forceAttrs(Value & v, const Pos & pos)
 
 inline void EvalState::forceList(Value & v)
 {
-    forceValue(v);
+    forceValue(v, v.determinePos(noPos));
     if (!v.isList())
         throwTypeError("value is %1% while a list was expected", v);
 }
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -522,7 +522,7 @@ ref<eval_cache::EvalCache> openEvalCache(
             auto vFlake = state.allocValue();
             flake::callFlake(state, *lockedFlake, *vFlake);
 
-            state.forceAttrs(*vFlake);
+            state.forceAttrs(*vFlake, noPos);
 
             auto aOutputs = vFlake->attrs->get(state.symbols.create("outputs"));
             assert(aOutputs);
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -198,8 +198,9 @@ void SourceExprCommand::completeInstallable(std::string_view prefix)
             prefix_ = "";
         }
 
-        Value &v1(*findAlongAttrPath(*state, prefix_, *autoArgs, root).first);
-        state->forceValue(v1);
+        auto [v, pos] = findAlongAttrPath(*state, prefix_, *autoArgs, root);
+        Value &v1(*v);
+        state->forceValue(v1, pos);
         Value v2;
         state->autoCallFunction(*autoArgs, v1, v2);
 
@@ -446,7 +447,7 @@ struct InstallableAttrPath : InstallableValue
     std::pair<Value *, Pos> toValue(EvalState & state) override
     {
         auto [vRes, pos] = findAlongAttrPath(state, attrPath, *cmd.getAutoArgs(state), **v);
-        state.forceValue(*vRes);
+        state.forceValue(*vRes, pos);
         return {vRes, pos};
     }
 
@@ -496,7 +497,7 @@ Value * InstallableFlake::getFlakeOutputs(EvalState & state, const flake::Locked
     auto aOutputs = vFlake->attrs->get(state.symbols.create("outputs"));
     assert(aOutputs);
 
-    state.forceValue(*aOutputs->value);
+    state.forceValue(*aOutputs->value, aOutputs->value->determinePos(noPos));
 
     return aOutputs->value;
 }
@@ -608,7 +609,7 @@ std::pair<Value *, Pos> InstallableFlake::toValue(EvalState & state)
     for (auto & attrPath : getActualAttrPaths()) {
         try {
             auto [v, pos] = findAlongAttrPath(state, attrPath, *emptyArgs, *vOutputs);
-            state.forceValue(*v);
+            state.forceValue(*v, pos);
             return {v, pos};
         } catch (AttrPathNotFound & e) {
         }
--- a/src/libexpr/attr-path.cc
+++ b/src/libexpr/attr-path.cc
@@ -58,7 +58,7 @@ std::pair<Value *, Pos> findAlongAttrPath(EvalState & state, const string & attr
         Value * vNew = state.allocValue();
         state.autoCallFunction(autoArgs, *v, *vNew);
         v = vNew;
-        state.forceValue(*v, v->determinePos(vIn.determinePos(noPos)));
+        state.forceValue(*v, noPos);
 
         /* It should evaluate to either a set or an expression,
            according to what is specified in the attrPath. */
--- a/src/libexpr/attr-path.cc
+++ b/src/libexpr/attr-path.cc
@@ -58,7 +58,7 @@ std::pair<Value *, Pos> findAlongAttrPath(EvalState & state, const string & attr
         Value * vNew = state.allocValue();
         state.autoCallFunction(autoArgs, *v, *vNew);
         v = vNew;
-        state.forceValue(*v);
+        state.forceValue(*v, v->determinePos(vIn.determinePos(noPos)));
 
         /* It should evaluate to either a set or an expression,
            according to what is specified in the attrPath. */
--- a/src/libexpr/eval-cache.cc
+++ b/src/libexpr/eval-cache.cc
@@ -336,7 +336,7 @@ Value & AttrCursor::getValue()
     if (!_value) {
         if (parent) {
             auto & vParent = parent->first->getValue();
-            root->state.forceAttrs(vParent, vParent.determinePos(noPos));
+            root->state.forceAttrs(vParent, noPos);
             auto attr = vParent.attrs->get(parent->second);
             if (!attr)
                 throw Error("attribute '%s' is unexpectedly missing", getAttrPathStr());
@@ -381,7 +381,7 @@ Value & AttrCursor::forceValue()
     auto & v = getValue();
 
     try {
-        root->state.forceValue(v, v.determinePos(noPos));
+        root->state.forceValue(v, noPos);
     } catch (EvalError &) {
         debug("setting '%s' to failed", getAttrPathStr());
         if (root->db)
--- a/src/libexpr/eval-cache.cc
+++ b/src/libexpr/eval-cache.cc
@@ -381,7 +381,7 @@ Value & AttrCursor::forceValue()
     auto & v = getValue();
 
     try {
-        root->state.forceValue(v);
+        root->state.forceValue(v, v.determinePos(noPos));
     } catch (EvalError &) {
         debug("setting '%s' to failed", getAttrPathStr());
         if (root->db)
--- a/src/libexpr/eval-cache.cc
+++ b/src/libexpr/eval-cache.cc
@@ -336,7 +336,7 @@ Value & AttrCursor::getValue()
     if (!_value) {
         if (parent) {
             auto & vParent = parent->first->getValue();
-            root->state.forceAttrs(vParent);
+            root->state.forceAttrs(vParent, vParent.determinePos(noPos));
             auto attr = vParent.attrs->get(parent->second);
             if (!attr)
                 throw Error("attribute '%s' is unexpectedly missing", getAttrPathStr());
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1280,7 +1280,7 @@ void ExprOpHasAttr::eval(EvalState & state, Env & env, Value & v)
     e->eval(state, env, vTmp);
 
     for (auto & i : attrPath) {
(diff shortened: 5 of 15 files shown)


regnat (approved):
Great!

I didn’t check whether all the calls to `determinePos` were justified, but in any case that’s a big improvement and a good thing for the future

src/libexpr/eval-inline.hh
```diff
@@ -53,7 +53,7 @@ void EvalState::forceValue(Value & v, const Pos & pos)
 
 inline void EvalState::forceAttrs(Value & v)
 {
-    forceValue(v);
+    forceValue(v, v.determinePos(noPos));
```
edolstra, line 56: What's the performance impact of these non-lazy calls to `determinePos()`?

BTW `determinePos()` is currently defined in `eval.cc`, but if we're going to call it everywhere (including on some hot paths), it should probably be moved to `value.hh` so that it can be inlined.

domenkozar: @kamadorueda it would be really cool to have this merged before it gets conflicts :blush:

kamadorueda: Here are some steps done:

1. I tried to inline determinePos, however it uses struct `Pos`, class `Bindings` and others, which by the time determinePos needs to be declared in order for the compiler to inline it, are not defined, so the compiler complains that we do not know their size:

image

which means that we cannot inline it

1. So I tried benchmarking the performance of this PR (top) vs nix-master (bottom) and there is certainly a difference:

Screenshot from 2022-01-21 09-33-25

Aparently 0.05 seconds every 6 or 7 seconds

kamadorueda: @edolstra 
I added a few commits to slim down the critical path,
closed my browser, code editor, only left open the terminal, 
and increased the bench-marking rounds to 100 (~10 minutes running) to have more data points

this is the result:

image

where there is a difference in execution time (rejected) the difference is negative (so faster),

I think it should not be becoming faster even if data tells so,
so I would be happy if someone help me confirm the results on other machine

```bash
nix-master $ clear; PAGER=cat benchmark -t 100 -f ../prev -- ./outputs/out/bin/nix-env -qaf ../nixpkgs
nix-pr $ clear; PAGER=cat benchmark -t 100 -f ../new -c ../prev -- ./outputs/out/bin/nix-env -qaf ../nixpkgs
```

tomberek: https://github.com/edolstra/benchmark

ConnorBaker: @ConnorBaker reminder to replicate benchmarks on M1 macbook and AWS aarch64 instance

edolstra:
Thanks, merged it. Because I did still observe a small slowdown, I've made most calls to `determinePos()` lazy by passing them as a lambda to `forceValue()` (bd383d1b6f91c4fe7ac21c52771e92027f649fa0).

src/libexpr/eval-inline.hh
ConnorBaker: @kamadorueda how do you run the benchmark perl script? I was able to get everything set up, but I'm not sure what arguments to pass to it.

```bash
nix build github:kamadorueda/nix/50efc5499a7d924828bed654be207a846c040fa0 --out-link nix-pr
nix build github:nixos/nix/5f08db69d18ea5b868cbb58993822e7dd0307518 --out-link nix-master
nix build github:edolstra/benchmark/096970f0ac3e74405db124902c39ecc8c0394713 --out-link benchmark
nix flake clone github:nixos/nixpkgs/21.11 --dest nixpkgs
```

kamadorueda: With your setup, would be something like:

```bash
# 100 rounds, save results in ./prev, run nix-env -qaf over nixpkgs using nix with this PR
$ PAGER=cat ./benchmark -t 100 -f ./prev -- ./nix-pr/bin/nix-env -qaf ./nixpkgs

# 100 rounds, save results in ./new, compare with ./prev, run nix-env -qaf over nixpkgs using nix-master
$ PAGER=cat ./benchmark -t 100 -f ./new -c ./prev -- ./nix-master/bin/nix-env -qaf ../nixpkgs
```

I just tried to document the tool: https://github.com/edolstra/benchmark/pull/3

ConnorBaker: When I tried to run it I got errors about being unable to set affinity.

```bash
$ PAGER=cat ./benchmark/bin/benchmark -t 3 -f ./prev -- ./nix-pr/bin/nix-env -qaf ./nixpkgs
run 0...
taskset: failed to set pid 39970's affinity: Invalid argument
run 1...
taskset: failed to set pid 39972's affinity: Invalid argument
run 2...
taskset: failed to set pid 39974's affinity: Invalid argument
note: discarding 10% highest outliers
maximum RSS:        median =  10574.0000  mean =  10574.0000  stddev =      2.8284  min =  10572.0000  max =  10576.0000
soft page faults:   median =     97.0000  mean =     97.0000  stddev =      1.4142  min =     96.0000  max =     98.0000
user CPU time:      median =      0.0006  mean =      0.0006  stddev =      0.0009  min =      0.0000  max =      0.0012
elapsed time:       median =      0.0016  mean =      0.0016  stddev =      0.0001  min =      0.0016  max =      0.0016
```

This is on NixOS 21.11 on an AWS t4g.medium instance:

```bash
$ uname -a
Linux ip-10-0-0-170.ec2.internal 5.10.81 #1-NixOS SMP Sun Nov 21 12:46:37 UTC 2021 aarch64 GNU/Linux
```

On macOS I don't get any errors, but the time is also suspiciously short:

```bash
$ PAGER=cat ./benchmark/bin/benchmark -t 3 -f ./prev -- ./nix-pr/bin/nix-env -qaf ./nixpkgs
run 0...
run 1...
run 2...
note: discarding 10% highest outliers
maximum RSS:        median = 1458176.0000  mean = 1458176.0000  stddev =  23170.4750  min = 1441792.0000  max = 1474560.0000
soft page faults:   median =    107.5000  mean =    107.5000  stddev =      0.7071  min =    107.0000  max =    108.0000
system CPU time:    median =      0.0004  mean =      0.0004  stddev =      0.0000  min =      0.0004  max =      0.0004
user CPU time:      median =      0.0000  mean =      0.0000  stddev =      0.0000  min =      0.0000  max =      0.0000
elapsed time:       median =      0.0007  mean =      0.0007  stddev =      0.0001  min =      0.0007  max =      0.0008
```

What were you running it on?

kamadorueda: - `Linux nixos 5.16.3 #1-NixOS SMP PREEMPT Thu Jan 27 11:03:05 UTC 2022 x86_64 GNU/Linux`
- config: https://github.com/kamadorueda/machine
- demo: 
Peek 2022-01-20 09-27

so your commands seem to be right, that `affinity: Invalid argument` seems like a permissions/kernel problem

ConnorBaker: For what it's worth, I also tried with an x86_64 c6i.xlarge on AWS with the most recent NixOS 21.11 AMI:

```bash
$ uname -a
Linux ip-10-0-0-127.ec2.internal 5.10.81 #1-NixOS SMP Sun Nov 21 12:46:37 UTC 2021 x86_64 GNU/Linux
```

I'm seeing the same issue:

```bash
PAGER=cat ./benchmark/bin/benchmark -t 3 -f ./prev -- ./nix-pr/bin/nix-env -qaf ./nixpkgs
run 0...
taskset: failed to set pid 33112's affinity: Invalid argument
run 1...
taskset: failed to set pid 33114's affinity: Invalid argument
run 2...
taskset: failed to set pid 33116's affinity: Invalid argument
note: discarding 10% highest outliers
maximum RSS:        median =  10538.0000  mean =  10538.0000  stddev =      2.8284  min =  10536.0000  max =  10540.0000
soft page faults:   median =    107.5000  mean =    107.5000  stddev =      0.7071  min =    107.0000  max =    108.0000
system CPU time:    median =      0.0004  mean =      0.0004  stddev =      0.0006  min =      0.0000  max =      0.0008
elapsed time:       median =      0.0010  mean =      0.0010  stddev =      0.0000  min =      0.0010  max =      0.0010
```

I'm not sure what could be causing this problem.

How are you setting up / building the sources? If you use the same series of commands I posted above to get/build them (with flakes), do you see similar behavior? Or is it not reproducible on your end?

kamadorueda: I'm using exactly your same commands, but maybe a difference can be that this is my home computer and not a cloud instance

here is a demo doing the steps one by one

Peek 2022-01-20 09-27

kamadorueda: this seems to be failing command:

https://man7.org/linux/man-pages/man1/taskset.1.html

kamadorueda: Executed with this arguments:
https://github.com/edolstra/benchmark/blob/096970f0ac3e74405db124902c39ecc8c0394713/benchmark.pl#L39

kamadorueda: I think I get the problem now, it's using the taskset from your system:

taskset is in nixpkgs.util-linux, which is not specified in the flake of the benchmark tool: https://github.com/edolstra/benchmark/blob/master/flake.nix#L18

ConnorBaker: I changed the flake to include it but I'm still seeing the same behavior.

```bash
$ git diff
diff --git a/flake.lock b/flake.lock
index 4f89d76..623d5c8 100644
--- a/flake.lock
+++ b/flake.lock
@@ -17,11 +17,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1642786796,
-        "narHash": "sha256-FsKuowd1SZGNhsX+YeNPlodp3mecyVbLPxCcUMbkLU8=",
+        "lastModified": 1643948526,
+        "narHash": "sha256-XmRNI2zmR0iTX7LUJy93MG6GZoMHy1nlMNcp31XHi8w=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "1281c5be8ae76b231e181cde812b7b6cdc3028f8",
+        "rev": "d67ad28fc301305baeeb364a04f0565c5f5118c8",
         "type": "github"
       },
       "original": {
diff --git a/flake.nix b/flake.nix
index a55646b..4a0de74 100644
--- a/flake.nix
+++ b/flake.nix
@@ -22,6 +22,7 @@
               StatisticsTTest
               UnixGetrusage
               FileSlurp
+              pkgs.util-linux
             ];
 
             unpackPhase = "true";

$ ./result/bin/benchmark -t 3 -- sleep 1
run 0...
taskset: failed to set pid 6481's affinity: Invalid argument
run 1...
taskset: failed to set pid 6483's affinity: Invalid argument
run 2...
taskset: failed to set pid 6485's affinity: Invalid argument
note: discarding 10% highest outliers
maximum RSS:        median =  10546.0000  mean =  10546.0000  stddev =      2.8284  min =  10544.0000  max =  10548.0000
soft page faults:   median =    103.0000  mean =    103.0000  stddev =      2.8284  min =    101.0000  max =    105.0000
system CPU time:    median =      0.0007  mean =      0.0007  stddev =      0.0000  min =      0.0007  max =      0.0008
elapsed time:       median =      0.0009  mean =      0.0009  stddev =      0.0000  min =      0.0009  max =      0.0010
```

kamadorueda: @ConnorBaker I'm able to reproduce it now, benchmark is trying to attach to CPU 6, but you maybe don't have that much CPU on the cloud instance:

(my home computer have 16)

```
[kamadorueda@nixos:/data]$ taskset -c 16 sleep 1
taskset: failed to set pid 3509's affinity: Invalid argument

[kamadorueda@nixos:/data]$ taskset -c 15 sleep 1
(ok)

[kamadorueda@nixos:/data]$ taskset -c 0 sleep 1
(ok)

```

kamadorueda: `-c 6` attaches to the 7th cpu
https://github.com/edolstra/benchmark/blob/096970f0ac3e74405db124902c39ecc8c0394713/benchmark.pl#L39

ConnorBaker: Thank you for sorting that out for me!

I used an `m6g.metal` instance and wasn't able to reproduce the `taskset` error (like you said, I think because of the core count).

```bash
$ lscpu
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 64
  On-line CPU(s) list:  0-63
Vendor ID:              ARM
  Model name:           Neoverse-N1
    Model:              1
    Thread(s) per core: 1
    Core(s) per socket: 64
    Socket(s):          1
    Stepping:           r3p1
    BogoMIPS:           243.75
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
Caches (sum of all):    
  L1d:                  4 MiB (64 instances)
  L1i:                  4 MiB (64 instances)
  L2:                   64 MiB (64 instances)
  L3:                   32 MiB (1 instance)
NUMA:                   
  NUMA node(s):         1
  NUMA node0 CPU(s):    0-63
Vulnerabilities:        
  Itlb multihit:        Not affected
  L1tf:                 Not affected
  Mds:                  Not affected
  Meltdown:             Not affected
  Spec store bypass:    Mitigation; Speculative Store Bypass disabled via prctl
  Spectre v1:           Mitigation; __user pointer sanitization
  Spectre v2:           Not affected
  Srbds:                Not affected
  Tsx async abort:      Not affected

$ lsmem
RANGE                                  SIZE  STATE REMOVABLE   BLOCK
0x0000000000000000-0x000000007fffffff    2G online       yes     0-1
0x0000000100000000-0x0000003fffffffff  252G online       yes   4-255
0x0000008100000000-0x000000817fffffff    2G online       yes 516-517

Memory block size:         1G
Total online memory:     256G
Total offline memory:      0B
```

I think part of the issue was that when I specified the path to clone Nixpkgs to, I forgot to make it a relative path, and it cloned it to the root directory (so it was at `/nixpkgs` instead of `/root/nixpkgs`). It was able to do that since on the EC2 instance I'm using I haven't set up any users, so it defaults to root.

Anyway, probably not relevant any more, but here's what I observed on the m6g.metal instance:

```bash
$ PAGER=cat ./benchmark/bin/benchmark -t 20 -f ./prev -- ./nix-pr/bin/nix-env -qaf ./nixpkgs
# Output snipped
maximum RSS:        median = 2023162.0000  mean = 1873273.7778  stddev = 436250.4782  min = 674012.0000  max = 2023432.0000
soft page faults:   median = 506947.0000  mean = 469244.7778  stddev = 109796.7131  min = 167440.0000  max = 507034.0000
system CPU time:    median =      0.9361  mean =      0.9208  stddev =      0.0556  min =      0.8147  max =      0.9861
user CPU time:      median =     10.3891  mean =      9.5182  stddev =      2.5623  min =      2.4753  max =     10.4820
elapsed time:       median =     11.3407  mean =     10.8687  stddev =      1.3863  min =      7.0558  max =     11.4138
```

```bash
$ PAGER=cat ./benchmark/bin/benchmark -t 20 -f ./new -c ./prev -- ./nix-master/bin/nix-env -qaf ./nixpkgs
# Output snipped
maximum RSS:        median = 2023060.0000  mean = 1873229.1111  stddev = 436261.1472  min = 674060.0000  max = 2023512.0000  [not rejected, p=0.99976, Δ=-44.66667±436546.64994]
soft page faults:   median = 506948.5000  mean = 469238.2222  stddev = 109793.4156  min = 167443.0000  max = 507037.0000  [not rejected, p=0.99986, Δ=-6.55556±109868.26108]
system CPU time:    median =      0.9401  mean =      0.9402  stddev =      0.0398  min =      0.8621  max =      1.0020  [not rejected, p=0.23556, Δ=0.01946±0.04839]
user CPU time:      median =     10.3588  mean =      9.4907  stddev =      2.5674  min =      2.4120  max =     10.4415  [not rejected, p=0.97448, Δ=-0.02755±2.56658]
elapsed time:       median =     11.3278  mean =     10.8647  stddev =      1.3801  min =      7.0591  max =     11.3926  [not rejected, p=0.99318, Δ=-0.00397±1.38413]
```

kamadorueda: That is very nice, those 5 "not rejected" mean this PR did not introduced any relevant performance impact, at least from the point of view of evaluating nixpkgs

Cool! Thank you

PR #5678, opened 2021-11-27
Document libsodium, which is now mandatory, as a dependency.

t184256:
libsodium is mandatory since 0df69d96e02ce4c9e17bd33333c5d78313341dd3, yet was not even mentioned in prerequisites.

As a side-note, the whole cryptographic library footprint is quite baffling.
I was already wondering why pull in openssl just for sha256/sha512 hashing,
and now as 2.4 hard-depends on libsodium, should we, perhaps, just use hashing from sodium and drop the openssl dependency?


t184256:
Kind of answering myself, libsodium doesn't have sha1/md5, so, depends on when are we going to wave goodbye to these for real.

t184256:
I tried my hand at it in anyway, see https://github.com/NixOS/nix/pull/5681

PR #5680, opened 2021-11-28
Don't move the arguments of the primOp

andir:
Moving arguments of the primOp into the registration structure makes it
impossible to initialize a second EvalState with the correct primOp
registration. It will end up registering all those "RegisterPrimOp"'s
with an arity of zero on all but the 2nd instance of the EvalState.

Not moving the memory will add a tiny bit of memory overhead during the
eval since we need a copy of all the argument lists of all the primOp's.
The overhead shouldn't be too bad as it is static (based on the amonut
of registered operations) and only occurs once during the interpreter
startup.

This has been extracted from #5377 as it fixes a bug that should be fixed regardless of the unit tests being added.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -3732,7 +3732,7 @@ void EvalState::createBaseEnv()
                 .fun = primOp.fun,
                 .arity = std::max(primOp.args.size(), primOp.arity),
                 .name = symbols.create(primOp.name),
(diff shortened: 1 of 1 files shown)


regnat (approved):
So that’s where these weird issues were coming from!

Thanks a lot for digging into that :)

(cc @Pamplemousse who might be interested too)

Issue #2803, opened 2019-05-12
Builders inherit ignored signals


hvdijk:
When a systemd service does not contain

    [Service]
    IgnoreSIGPIPE=no

systemd will cause SIGPIPE to be ignored, and Nix does nothing to restore this in builders. As a result, builds behave subtly differently depending on whether nix-daemon is used or not, and whether nix-daemon is launched from systemd or launched manually.

This showed up when running diffutils 3.7's test suite inside a builder, which tests that in a specific situation, diff correctly fails with SIGPIPE. This test failed, because diff exited with a regular exit code instead, until I stopped the nix-daemon service and launched nix-daemon manually.

After modifying nix-daemon.service as described, diffutils's tests do pass.

Is adding `IgnoreSIGPIPE=no` the right solution, or should Nix fix this up itself, for all signals no matter how it's started? The former solves the immediate problem the simplest way, the latter would take more work but would potentially fix more cases.

hvdijk:
Had to look a bit for how to create a simple standalone testcase, but:

    nix-build --expr 'with builtins; derivation {
      name = "test";
      system = currentSystem;
      __noChroot = true;
      PATH = "/bin:/usr/bin";
      builder = "/bin/bash";
      args = ["-c" "set -o pipefail; cat /bin/bash | head -c1 >/dev/null; echo $?"];
    }'

Paths may need to be adjusted. This should print "141", but depending on how nix-daemon was started, may actually print "cat: write error: Broken pipe", followed by "1". This can be modified to conditionally create `$out`, if desired.

luke-clifton:
This causes me endless number of headaches. I think `nix` itself needs to actually fix the issues itself. I can't control how other people have started their daemon.

hvdijk:
I forgot about this when setting up Nix on a new system and spent time again digging into the same failing test in diffutils 3.8. While it would be nice to fix this in Nix itself so that it is guaranteed to work well for everybody, the systemd service file is provided by Nix itself, could at least that be updated so that a default installation will just work?

ju1m:
For the record, `public-inbox`'s author [opted for reenabling SIGPIPE](https://public-inbox.org/meta/20211208010730.f47xxgzj53nwgvja@sourcephile.fr/T/#m38685d23bd686442d91a4890bac3c018d427b96b):
> Yes, blocking SIGPIPE makes sense for most daemons so I think
> systemd is being reasonable, here.  However, this lei test is
[...]

PR #5683, opened 2021-11-29
createTempFile(): Mark file as CLOEEXEC

### LINKED ISSUE #5674: Weird file descriptors
File descriptors returned by`open`  aren't consecutive when entering `nix develop`.

```
$ nix develop nixpkgs#hello
$ cat >fds.c <<EOF
#include <fcntl.h>
#include <stdio.h>
int main() {
  for (int i = 0; i < 20; i++) printf("%d\n", open("/dev/null", O_RDWR));
  return 0;
}
EOF
$ gcc fds.c
$ ./a.out
3
4
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
21
22
23

```

I would expect the FDs to be consecutive. This results in a test failure in https://github.com/jart/cosmopolitan/blob/d6a039821f927cc81b71e11f21599ff93e1fed62/test/libc/calls/fcntl_test.c#L61

nix-env (Nix) 2.4

edolstra:
Fixes #5674.

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -512,6 +512,7 @@ std::pair<AutoCloseFD, Path> createTempFile(const Path & prefix)
     AutoCloseFD fd(mkstemp((char *) tmpl.c_str()));
     if (!fd)
         throw SysError("creating temporary file '%s'", tmpl);
+    closeOnExec(fd.get());
     return {std::move(fd), tmpl};
 }
 


ju1m:
@edolstra does this change also concern `nix build` or is it only for `nix develop`?
That question was raised while trying to debug a [test of public-inbox](https://public-inbox.org/meta/YfnoOVOG3TrbqZFs@codewreck.org/) which is hanging when run by `nix build` but is working when entering manually `nix build`'s sandbox to run it.

martinetd:
Eh, I just meant that I didn't test/look at the nix build code -- nix develop uses createTempFile to create nix-shell rc script with the build environment variables but I don't think that applies to nix build so these aren't related and would be best checked.

I just tried adding `ls -l /proc/$$/fd` and /proc/self/fd to preCheck and it doesn't look like there is any weird fd, so it's probably fine :)

ju1m:
For the record, the problem was related to something else: https://github.com/NixOS/nix/issues/2803 .

PR #5684, opened 2021-11-29
nix repl: Mark as experimental

### LINKED ISSUE #5604: `nix repl` is available without the `nix-command` experimental feature
**Describe the bug**

`nix repl` is available without the `nix-command` experimental feature.

**Steps To Reproduce**

```
[nix-shell:.../channels/nixpkgs]$ nix repl
Welcome to Nix 2.4. Type :? for help.


[nix-shell:.../channels/nixpkgs]$ nix build
```
**Expected behavior**

It is unclear whether any commands under the new `nix` command should work without the flag enabled.

```
[nix-shell:.../channels/nixpkgs]$ nix-env --version
nix-env (Nix) 2.4

[nix-shell:.../channels/nixpkgs]$ type nix-env
nix-env is /nix/store/ag01yxbp4z0bksh7qc67xc31rgk91yaa-nix-2.4/bin/nix-env
```

edolstra:
Fixes #5604.

### DIFF
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -344,8 +344,7 @@ void mainWrapped(int argc, char * * argv)
     if (!args.command)
         throw UsageError("no subcommand specified");
 
-    if (args.command->first != "repl"
-        && args.command->first != "doctor"
+    if (args.command->first != "doctor"
         && args.command->first != "upgrade-nix")
         settings.requireExperimentalFeature(Xp::NixCommand);
 


bjornfor:
Bring back 'nix-repl' then?

edolstra:
@bjornfor You mean as a separate project?

bjornfor:
> @bjornfor You mean as a separate project?

It doesn't really matter to me, but as a NixOS user I wonder about this chain of events:
1. Use `nix-repl`.
2. `nix-repl` is removed, replaced by `nix repl`.
3. `nix repl` is marked as experimental -- no way to run a Nix REPL without opt-in to unstable/experimental Nix.

It feels like we just lost the ability to run a Nix REPL.

regnat:
I agree with @bjornfor’s concerns . Contrary to the other `nix` commands, `nix repl` is the only interface to something that isn’t possible with the old cli anymore (not counting the now defunct `nix-repl`). So if it becomes experimental, there should be a replacement.

I don’t know whether it’s worth adding a `nix-repl` command just for the sake of keeping the whole `nix` command experimental or whether it’s fine to just keep it as-it-is (esp. given that there’s other non-experimental `nix` subcommands like `doctor` or `upgrade-nix`).

Ericson2314:
Has the ship sailed on this?

fricklerhandwerk:

- @ericson2314: can't make it experimental any more, it's stable and widely used
- @regnat this is the only new `nix` command that is not guarded by an experimental feature flag but where we know we want to make changes
    - @fricklerhandwerk: *sigh*

PR #5690, opened 2021-11-30
docs: document set theory symbols

efx:
These symbols confused me so I wanted to gloss them to help future users. You can see the context here: https://discourse.nixos.org/t/flakes-what-does-mean-big-bundle-of-questions

### DIFF
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -47,7 +47,7 @@
     the store object at `P` contains the path `Q` somewhere. The
     *references* of a store path are the set of store paths to which it
     has a reference.
-    
+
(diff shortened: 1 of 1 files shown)


doc/manual/src/glossary.md
```diff
@@ -98,3 +98,7 @@
     store. It can contain regular files, directories and symbolic
     links.  NARs are generated and unpacked using `nix-store --dump`
     and `nix-store --restore`.
+  - `∅` \
+    The empty set symbol. In the context of package history, this denotes a package was not present prior to installation.
```
edolstra, line 102:
```suggestion
    The empty set symbol. In the context of profile history, this denotes a package is not present in a particular version of the profile.
```

PR #5691, opened 2021-11-30
Add a github cron to check the hydra status

regnat:
Add a regular github action that will check the status of the latest hydra evaluation.

Things aren’t ideal right now because this job will only notify “the user who last modified the cron syntax in the workflow file” (so myself atm). But at least that’ll give a notification for failing hydra jobs

### DIFF
--- a/.github/workflows/hydra_status.yml
+++ b/.github/workflows/hydra_status.yml
@@ -1,0 +1,16 @@
+name: Hydra status
+on:
+  schedule:
+    - cron: "12,42 * * * *"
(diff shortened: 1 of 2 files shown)


abathur:
This is probably a no-op, but I noticed recently, after helping someone get set up to run installer tests on their fork, that their action history was full of skips for this test. 

I thought it was worth noting, but AFAIK the `if: github.repository_owner == 'NixOS'` _is_ the "right" way to keep it from running on random forks, so I'm not sure there's anything more than can be done...

Issue #4612, opened 2021-03-06
Be able to run `nix build` / `nix develop` from anywhere in the project


garbas:
## Proposal

Allow to enter development environment or even build a project by being anywhere inside the project tree.

## Reasoning

It is already true that even old commands `nix-shell` and `nix-build` by default target local project. By local I mean project within the repository from where you run the command. With new command line, `nix`, we can make this even better.

I would like to propose to extend the default functionality and allow users to enter development shell or build a project from anywhere they are in the project tree.=

When a user would not provide any argument, we would lookup the root of the project by scanning all parent directories until we find `flake.nix` in any of them. After that we would use that flake to run the command against.

Such a change in default would not confuse existing users, but adds interesting new possibilities when developing with Nix.

RCoeurjoly:
If a flake repository is always a git repository, then it is easy.
But I tried running nix build in a regular Linux directory and it builds without problems, so I don't think it is a good idea to ¨lookup the root of the project by scanning all parent directories until we find flake.nix in any of them¨

Radvendii:
@RCoeurjoly why not?

PR #5692, opened 2021-11-30
allow discovery of a git repo's flake.nix in a parent directory

lincolnauster:
This allows for the discovery of flake.nix's in parent directories of a git repository, allowing for workflows like this:

```bash
$ cd $some_source_directory
$ cat flake.nix
...
$ cd src
$ vi $some_file
$ nix build
...
# before, this would fail as the flake.nix is not strictly in the CWD. This patch,
# however, has Nix traverse up the directory and use the first flake.nix it sees
# in the git repo
```

Just one commit right now, here's the message:
```
When traversing a directory looking for the root of a repository, allow
Nix to use the earliest flake.nix that it encounters. This allows for
workflows where builds of the whole package are performed in arbitrary
subdirectories of that package in a do-what-i-mean fashion.

Errors and other outputs are unchanged.
```
---

Feedback on style/etc is very welcome - this would be my first time using C++ for anything of a decent size! Also: (where) should this change be documented? I don't see a CHANGELOG.md or similar. Thanks!


regnat (changes requested):
Thanks a lot! I guess that would solve https://github.com/NixOS/nix/issues/4612

Apart from the small inline comments, I just have one concern: The logic also works when an explicit path is passed (say `nix build /foo/bar`), in which case I think it shouldn’t (I’d expect `nix build /foo/bar` to directly fail if `/foo/bar/flake.nix` doesn’t exist, not to try finding a `flake.nix` in a parent directory).

Ideally we should only have this logic when the path isn’t specified (`nix build`). Distinguishing between `nix build` and `nix build .` might be a bit tricky (because they are resolved to the same thing pretty early on), but maybe we could as a rough approximation just make this logic work when `path` is equal to `.`.

> (where) should this change be documented? I don't see a CHANGELOG.md or similar

There’s a (somewhat hidden) release notes file under `doc/manual/src/release-notes/rl-next.md`

src/libutil/util.cc
```diff
@@ -1831,4 +1831,18 @@ void commonChildInit(Pipe & logPipe)
     close(fdDevNull);
 }
 
+/* Given two pointers to c strings, return a pointer to the start of the first
+ * non-equivalent part of the first string. */
+const char *removeStartingOverlap(const char *a, const char *b)
```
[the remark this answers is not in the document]
lincolnauster: Ah, thanks for catching that breakage, and for telling me about string_view, nice to know that exists! I'm pretty sure that function's fixed with cd5e57838 and bee39a5b7.

src/libutil/util.cc
```diff
@@ -1831,4 +1831,26 @@ void commonChildInit(Pipe & logPipe)
...
+
+    for (auto ap = a.begin(), bp = b.begin(),
+              ae = a.end(),   be = b.end();
+         ap < ae && bp < be;
+         ++ap, ++bp, ++start
+    ) {
+        if (*ap != *bp) break;
+    }
+
+    return a.substr(start, end - start);
```
regnat, line 1853: Don’t you want 
```suggestion
    return a.substr(start);
```
?

I think `a.substr(start, end-start)` will be too short if `a.length()>b.length()` (which might not happen here, but still)

src/libutil/util.hh
```diff
@@ -649,5 +649,5 @@ template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
 
 std::string showBytes(uint64_t bytes);
 
-
+const std::string_view removeStartingOverlap(std::string & a_ref, std::string & b_ref);
```
regnat, line 652:
```suggestion
const std::string_view removeStartingOverlap(const std::string & a_ref, const std::string & b_ref);
```

(or directly take `string_view`s as input, C++ will happily do the conversion implicitely)

src/libutil/util.cc
```diff
@@ -1831,4 +1831,26 @@ void commonChildInit(Pipe & logPipe)
     close(fdDevNull);
 }
 
+/* Given two pointers to c strings, return a pointer to the start of the first
```
regnat, line 1834:
```suggestion
/* Given two strings, return a pointer to the start of the first
```

src/libutil/util.cc
regnat: I’m not sure, but maybe you can also rewrite this as

```suggestion
    return std::string_view(ap, a.end());
```

in which case you won’t need `start` anymore

regnat (commented):
> Ah, thanks for catching that breakage, and for telling me about string_view, nice to know that exists! I'm pretty sure that function's fixed with cd5e578 and bee39a5.

I did nitpick a bit more on it ;)
(but apart from the possible bug on the return line, it’s really nitpicking, so feel free to ignore it)

lincolnauster:
thanks for all the suggestions - I'm slowly working through it between my classes, I'll look at the rest of it later today!

src/libexpr/flake/flakeref.cc
```diff
@@ -136,10 +142,13 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
...
                     };
 
-                    if (subdir != "") {
+                    if (flakedir) {
                         if (parsedURL.query.count("dir"))
                             throw Error("flake URL '%s' has an inconsistent 'dir' parameter", url);
-                        parsedURL.query.insert_or_assign("dir", subdir);
+                        const char *subdir = removeStartingOverlap(
+                            flakedir.value().c_str(), flakeRoot.c_str()
+                        );
```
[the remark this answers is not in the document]
lincolnauster: It looks like this completely overwrites the previous logic for assigning subdir (and you can see that in the earlier commits with the shadowed variable name), so I've just gotten rid of the assignment you're talking about here (e5d96f4b9). The function's getting a bit large though, I'll see what I can do about that.

src/libexpr/flake/flakeref.cc
```diff
@@ -136,10 +142,13 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
                         .query = decodeQuery(match[2]),
                     };
 
-                    if (subdir != "") {
+                    if (flakedir) {
```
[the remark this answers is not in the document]
lincolnauster: Yep, completely forgot about that error message, thank you! 1ad6f1022

lincolnauster:
The CI failures look to be fixed with adea817d0 and d2b227931. I everything looks correct to me, but more testing by people with more familiarity with the expected behavior would be greatly appreciated. Thanks!

regnat:
Unfortunately there’s still a failure on the testsuite, not sure why :/ (you can test it locally with `make installcheck` or `make tests/flake.sh.test`).

Speaking of test, do you mind adding one? Something like below should do it:

```diff
diff --git a/tests/flakes.sh b/tests/flakes.sh
index 9e10322b9..aa528f412 100644
--- a/tests/flakes.sh
+++ b/tests/flakes.sh
@@ -132,6 +132,16 @@ nix build -o $TEST_ROOT/result flake1#foo
 nix build -o $TEST_ROOT/result flake1
 [[ -e $TEST_ROOT/result/hello ]]
 
+# Test building from a subdir
+(
+    cd "$flake1Dir"
+    mkdir subdir
+    nix build ./subdir && fail "Building from a subdir shouldn’t work when the path is passed on the cli" || true
+    cd subdir
+    rm "$TEST_ROOT/result" && nix build -o "$TEST_ROOT/result" .
+    [[ -e $TEST_ROOT/result/hello ]]
+)
+
 nix build -o $TEST_ROOT/result $flake1Dir
 nix build -o $TEST_ROOT/result git+file://$flake1Dir
```

src/libutil/util.cc
```diff
@@ -1831,4 +1831,25 @@ void commonChildInit(Pipe & logPipe)
     close(fdDevNull);
 }
 
+/* Given two strings, return a pointer to the start of the first
+ * non-equivalent part of the first string. This is a non-owning reference to
+ * data owned by a, and must not outlive a. */
+const std::string_view removeStartingOverlap(const std::string & a_ref, const std::string & b_ref)
```
knedlsepp, line 1837: Maybe we could make use of std::mismatch?

lincolnauster:
Hi, I'm getting back to this as finals slow down; running `nix flake check` seems to be broken on both my branch and upstream (error message below). Is that expected at the moment?

```
error: attribute 'pkgconfig' missing

       at /nix/store/ijrwprm3yzljdv1hjrmnvxfg453h420f-source/flake.nix:99:13:

           98|             buildPackages.autoreconfHook
           99|             buildPackages.pkgconfig
             |             ^
          100|
```

src/libutil/util.cc
lincolnauster: Yep, works great - didn't even know that existed! 6ab209ac4

src/libexpr/flake/flakeref.cc
```diff
@@ -120,36 +120,38 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
                repo (i.e. the directory containing .git). The flake.nix
                seen earliest is used. */
 
-            auto flakeInfo = findFlakeDirs(absPath(path, baseDir, true), path);
-            if (flakeInfo.first) {
-                // the flake is contained in a git repo
+            auto [gitRepo, dir] = path == "." ?
+                  findFlakeDirs(absPath(path, baseDir, true), path)
+                : std::make_pair(std::nullopt, absPath(path, baseDir, true));
```
regnat, line 131: I think this will mean that `nix build /some/path` will never treat `/some/path` as a git repository, which is problematic.
For example running `nix build $PWD` from the Nix repository will start by taking forever copying the whole directory (including the several-hundred-MBs of locally built artifacts) to the store

lincolnauster:
Would #5720 be a better candidate for merging? Looks like tests are passing there, it better matches git's behavior of stopping at fs boundaries. I kind of forgot about this PR, tbh; I would love to see this feature in a release soon, so it might be better to go with 5720 :)

PR #5696, opened 2021-11-30
Fix #5299 and remove uncesssary unbounded buffer

### LINKED ISSUE #5299: nix-store 2.3 incompatible with nix-daemon 2.4pre
Using nix-store 2.3 on a system where nix-daemon 2.4pre is running results in unexpected behavior.

**Steps To Reproduce**

Following script
```
#!/usr/bin/env nix-shell
#!nix-shell -i sh -p nix

nix-store --version
echo hello > test.txt
STORE=$(nix-store --add-fixed sha256 test.txt)
hexdump -C "$STORE"
```

produces following output:
```
warning: unknown setting 'experimental-features'
nix-store (Nix) 2.3.15
warning: unknown setting 'experimental-features'
00000000  68 65 6c 6c 6f 0a 00 00  00 00 00 00 00 00 00 00  |hello...........|
00000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00010000  00 00 00 00 00 00                                 |......|
00010006
```

The host system runs nix version `2.4pre20210908_3c56f62`.

**Expected behavior**

nix-store shall not add 65536 zeros at the end of the file.

Ericson2314:
See each commit message for details

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -432,19 +432,26 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
             }
 
             auto dumpSource = sinkToSource([&](Sink & saved) {
-                TeeSource savedNARSource(from, saved);
-                RetrieveRegularNARSink savedRegular { saved };
-
                 if (method == FileIngestionMethod::Recursive) {
-                    /* Get the entire NAR dump from the client and save it to
-                      a string so that we can pass it to
-                      addToStoreFromDump(). */
+                    /* We parse the NAR dump through into `saved` unmodified,
+                       so why all this extra work? We still parse the NAR so
+                       that we aren't sending arbitrary data to `saved`
+                       unwittingly`, and we know when the NAR ends so we don't
+                       consume the rest of `from` and can't parse another
+                       command. (We don't trust `addToStoreFromDump` to not
+                       eagerly consume the entire stream it's given, past the
+                       length of the Nar. */
+                    TeeSource savedNARSource(from, saved);
                     ParseSink sink; /* null sink; just parse the NAR */
                     parseDump(sink, savedNARSource);
-                } else
+                } else {
+                    /* Incrementally parse the NAR file, stripping the
+                       metadata, and streaming the sole file we expect into
+                       `saved`. */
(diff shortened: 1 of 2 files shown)


regnat (approved):
Thanks!

It does indeed fix #5299 and makes things nicer to read in the end :+1:

dtzWill (commented):
See review comments for some minor touchups to the larger of introduced comments.

Also in interest of not just nitpicking and helping craft this instead here's one possible alternative (attempting to keep/use your language) which maybe is not better:

> Pass through the NAR unmodified into `saved`, but use parseDump to parse it anyway so we aren't sending arbitrary data unwittingly, and so that we only consume the NAR off of `from` and not past it.
> (We don't trust `addToStoreFromDump` to not eagerly consume the entire stream it's given, past the length of the NAR.)

(personally I have no idea why `addtoStoreFromDump` is mentioned here, but leaving it in spirit of Chesterson's fence, and I'm very much not suggesting it's not there for a good reason :smile:).

Whatever you think is best, hope this helps! :smile:.

---

Tried to not too intrusively clarify we're letting `parseDump` drive the amount consumed (TeeSource only writing to `saved` what is read from it in `parseDump`, since that seems to be the point of this)--behavior that wasn't obvious to me but I'm not super familiar with the Source/Sink bits we've got, but IDK.  Mostly "parse it" is not obviously "parse and read a limited amount (one object)" as that's not normally strictly speaking what I think of as the role of a parser.  But that is clearly what's expected, for example, in the `else` side of this and likely elsewhere so :shrug:.

Honestly maybe what you have with your selection of the suggested fixups is best! :+1:.

Ericson2314:
@dtzWill Would you want to just make another PR rewriting the comment as you fit? I would be happy to review it!

Ma27:
@edolstra @Ericson2314 as shown in https://github.com/NixOS/nix/pull/5740, this actually causes noticeable, yet weird problems.
Would you mind backporting this and perhaps even releasing a new Nix 2.4? :)

Ericson2314:
N.B. backport PR is already up at https://github.com/NixOS/nix/pull/5721

PR #5699, opened 2021-11-30
darwin-store: Ensure NIX_ROOT is mounted after reboot.

jsoo1:
When a darwin host is rebooted, /nix was not mounted.


jsoo1:
Closing in favor of #5698

PR #5702, opened 2021-12-01
reproducibility: hide non-reproducible settings from manual

baloo:
Because the manual is generated from default values which are themselves
generated from various sources (cpuid, bios settings (kvm), number of
cores). This commit hides non-reproducible settings from the manual
output.

### DIFF
--- a/doc/manual/generate-options.nix
+++ b/doc/manual/generate-options.nix
@@ -8,17 +8,19 @@ concatStrings (map
     let option = options.${name}; in
     "  - `${name}`  \n\n"
     + concatStrings (map (s: "    ${s}\n") (splitLines option.description)) + "\n\n"
-    + "    **Default:** " + (
-      if option.value == "" || option.value == []
-      then "*empty*"
-      else if isBool option.value
-      then (if option.value then "`true`" else "`false`")
-      else
-        # n.b. a StringMap value type is specified as a string, but
-        # this shows the value type.  The empty stringmap is "null" in
-        # JSON, but that converts to "{ }" here.
-        (if isAttrs option.value then "`\"\"`"
-         else "`" + toString option.value + "`")) + "\n\n"
+    + (if option.documentDefault
+       then "    **Default:** " + (
+       if option.value == "" || option.value == []
+       then "*empty*"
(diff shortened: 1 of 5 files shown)


baloo:
This is one of the issue we have over at https://r13y.com

doc/manual/generate-options.nix
```diff
@@ -8,17 +8,19 @@ concatStrings (map
...
+       then "*empty*"
+       else if isBool option.value
+       then (if option.value then "`true`" else "`false`")
+       else
+         # n.b. a StringMap value type is specified as a string, but
+         # this shows the value type.  The empty stringmap is "null" in
+         # JSON, but that converts to "{ }" here.
+         (if isAttrs option.value then "`\"\"`"
+          else "`" + toString option.value + "`")) + "\n\n"
+       else "")
```
regnat, line 23:
```suggestion
       else "    **Default:** *machine-specific*")
```

src/libutil/config.hh
```diff
@@ -232,16 +232,19 @@ protected:
...
     const T defaultValue;
+    const bool documentDefault;
 
 public:
 
     BaseSetting(const T & def,
+        const bool documentDefault,
         const std::string & name,
         const std::string & description,
         const std::set<std::string> & aliases = {})
```
regnat, line 243: Is this actually used? It looks like aggregate initialization is used everywhere, so there’s probably no need to change the constructor

regnat (changes requested):
A couple of small comments, but looks good overall!

src/libutil/config.hh
baloo: What would you suggest instead?
Put a default value (`true`) and override that in the `Setting` constructor where needed?

I feel like modifying the constructor is better, and less error-prone (no ordering problems)

regnat: Nah, I just have the impression that the constructor is unused − `BaseSetting { x, y , z }` doesn’t use the constructor, it just initializes the fields as they appear in the struct. But it doesn’t really matter anyways

regnat (approved):

PR #5705, opened 2021-12-01
nix store gc --dry-run: print paths

### LINKED ISSUE #5704: "nix store gc --dry-run" doesn't print anything: fix or remove?
**Describe the bug**

The `nix store gc` command supports `--dry-run` as described in it's `--help`:

```
      · --dry-run
        Show what this command would do without doing it.
```

However nothing is printed.

**Steps To Reproduce**

1. Have items available for gc
2. `nix store gc --dry-run`
3. Wait for command to do a bunch of work but then report nothing

**Expected behavior**

List of what would be deleted, as described by `--help`/docs.

```
nix-env (Nix) 2.5pre20211126_55275fc
```

**Additional context**

Investigating a little, the relevant code is here:

https://github.com/NixOS/nix/blob/f3f32f0c30f197af2fd5a90702f7002ae74b2dfb/src/nix/store-gc.cc#L39

Which converts "dry run" into telling gc bits to return paths in `results` but they are ignored instead of printed.

See also:  #1751.

Unlike with that issue, it seems `nix store gc --dry-run` doesn't list generations that would be removed either.

Hopefully this can be fixed with some sort of `if (dryRun) for (auto &r: results) printSomehow(r)` addition, I'll put together a PR if can populate that last  part with something reasonable :).

dtzWill:
Fixes #5704 .

Functionality is basic, and indeed code was copied from how it's already done here for non-experimental version: https://github.com/NixOS/nix/blob/f3f32f0c30f197af2fd5a90702f7002ae74b2dfb/src/nix-store/nix-store.cc#L605

---

Would add a test, but there appears no tests for `nix store gc` in-tree so wasn't sure where to put it (and adding new test/setup seems beyond scope here).  If there's a good place let me know :smile:.

I'm seeing many ways to print things, so mostly just did what we already do for this but maybe that's not right for new code-- LMK and happy to fix :+1:.

(seems sometimes with `fmt("%s\n", x)` and mix of newlines using `<< "\n"` and `<< std::endl`.  Each with trade-offs re:allocations, race against other prints (possibly separating path from newline, maybe), and flushing behavior, I think?)

Anyway short fix for issue I filed earlier, apologies if this is in wrong direction haven't been following development very closely but wanted to chip in :innocent:.

Thanks!

### DIFF
--- a/src/nix/store-gc.cc
+++ b/src/nix/store-gc.cc
@@ -37,6 +37,10 @@ struct CmdStoreGC : StoreCommand, MixDryRun
         GCResults results;
         PrintFreed freed(options.action == GCOptions::gcDeleteDead, results);
         store->collectGarbage(options, results);
+        if (dryRun)
+          for (auto & i : results.paths)
+            std::cout << i << std::endl;
+
     }
 };
 


edolstra:
IMHO writing (potentially) thousands of paths to stdout that the user didn't ask for is not really good UX. It's also not obvious whether this should print the dead or alive paths. I think it would be better to add a `--json` flag, and then the JSON output could include both dead and alive paths.

dtzWill:
Okay.  Unfortunately the similarly trivial implementation doesn't quite work as `GCResults`'s `bytesFreed `doesn't update the count in this case (although comment about the field declaration suggests it does, FWIW).  Haven't looked into how hard plumbing that through the relevant code would be, unsure at a glance :smile:.  I'm briefly concerned some stuff (like link cleanup?) might be more of a "do it and count as you go" sort of situation (delete until count is 1, or whatever) but I don't know.

I'm interested but unsure when I can commit, apologies.

---

Quick test/sample of this here: https://github.com/dtzWill/nix/commit/46f400254975378e5d1132283847145db965c298, not pushing because it doesn't work (as described above) and can't mark this as draft myself apparently (boo, PR authors should be able to surely? anyway).  And maybe poking at `PrintFreed` to handle the "would be" language, but perhaps not.  Presumably it's the way it is now so that exceptions during GC still print what was freed or something?

Anyway, this seems more than I can tackle immediately but maybe in a bit I can poke at the relevant code :+1:.

Thanks for taking a look and no worries glad for such concerns as what is good UX.  Especially for the new cli! :grin:.

A `--json` option seems neat, and straightforward enough to add.  Sounds good to me :+1:.

Ericson2314:
Draft because reviewer and author agreed on other approach.

tomberek:

Issue #5566, opened 2021-11-15
Re-installation issue on OS X


bbarker:
I was attempting to reinstall nix on OS X for[ reasons described elsewhere](https://discourse.nixos.org/t/reinstalling-nix-on-os-x-hangs/16064/2), however, this fails, as shown below. The error appears like it [may be related to active directory,](https://discussions.apple.com/thread/252518458) and indeed this machine is an AD-connected system.

```
macbookwh:~ bbarker$ sudo diskutil umount /Volumes/Nix\ Store/
Volume Nix Store on disk1s6 unmounted
macbookwh:~ bbarker$ curl -L https://nixos.org/nix/install | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  4024  100  4024    0     0   2519      0  0:00:01  0:00:01 --:--:-- 3929k
downloading Nix 2.4 binary tarball for x86_64-darwin from 'https://releases.nixos.org/nix/nix-2.4/nix-2.4-x86_64-darwin.tar.xz' to '/var/folders/_f/1dn5fzlx35n3_zc7nw290w119r5pmn/T/nix-binary-tarball-unpack.XXXXXXXXXX.nLcQoy0z'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 33.5M  100 33.5M    0     0  2374k      0  0:00:14  0:00:14 --:--:-- 2903k
Switching to the Multi-user Installer
Welcome to the Multi-User Nix Installation

This installation tool will set up your computer with the Nix package
manager. This will happen in a few stages:

1. Make sure your computer doesn't already have Nix. If it does, I
   will show you instructions on how to clean up your old install.

2. Show you what we are going to install and where. Then we will ask
   if you are ready to continue.

3. Create the system users and groups that the Nix daemon uses to run
   builds.

4. Perform the basic installation of the Nix files daemon.

5. Configure your shell to import special Nix Profile files, so you
   can use Nix.

6. Start the Nix daemon.

Would you like to see a more detailed list of what we will do?
No TTY, assuming you would say yes :)

We will:

 - make sure your computer doesn't already have Nix files
   (if it does, I will tell you how to clean them up.)
 - create local users (see the list above for the users we'll make)
[...]

PR #5711, opened 2021-12-02
darwin-install: nail down diskutil invocations

abathur:
Same purpose as de9efa3, applied to invocations added in 079bde2aef

For some unclear reason, we get occasional reports that the installer fails from people who do not have /usr/sbin on their PATH . It's a standard part of the PATH, so I have no clue what they're doing to remove it--but I guess it's fairly cheap to avoid.

### DIFF
--- a/scripts/install-darwin-multi-user.sh
+++ b/scripts/install-darwin-multi-user.sh
@@ -218,7 +218,7 @@ EOF
         setup_darwin_volume
     fi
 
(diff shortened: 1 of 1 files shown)


abathur:
I suspect this fixes the cause of recent reports in https://discourse.nixos.org/t/nix-installation-fails-on-macos/16366 and https://github.com/NixOS/nix/issues/5566#issuecomment-983576758

Issue #5697, opened 2021-11-30
flakes: flake inputs incorrectly following from the root of the current flake


nrdxp:
**Describe the bug**

I noticed recent versions of Nix in the last month or so had strange follows behavior that caused some errors I wasn't seeing before. I boiled it down to this: flake inputs now seem to want to track their follows from the root of the consuming flake, rather than from their own root as it was before. This means that if flake B has inputs C and D with `C.follows = "D/C"`, and flake A pulls B as an input, you'll now see an error when trying to generate A's lockfile like so:
```
error: input 'B/C' follows a non-existent input 'D/C'
```
Whereas before, this error did not occur and C would correctly come from B's lockfile, that is: "B/D/C". Now it seems Nix no longer knows about "B/D/C" and thinks "D" doesn't exist at all.

**Steps To Reproduce**

Just so happens this is occurring in a public work repo so you can refernce it to reproduce:

1. get or build nix on or after the offending commit: 07bffe799853ef5e2757b9892e4e3dac89bfccbb (discovered with `git bisect`)
2. run `nix flake info github:input-output-hk/bitte/b6f9b26935425a2efadb9b35201c99916ced3a33`
3. You should then see the following error:
```
error: input 'bitte-cli/nixpkgs' follows a non-existent input 'fenix/nixpkgs'
```
4. If you'd like to see the old behavior, simply get a build of Nix before the aforementioned commit and notice how there is no error

**Expected behavior**

Since follows are declared in flake B, it would make sense they would try to follow starting from _its_ root, and not the root of the consuming flake.

nrdxp:
@balsoft as the original author of the offending commit, I wonder if you may have an idea of what's going on?

balsoft:
Here we go again...

I'll look into it on the weekend

smunix:
Thanks @balsoft ! I can confirm, the fix works here

sheyll:
Hi there, might it be possible that this bug came back? I'm trying the current 2.6 release and get the same behaviour:

```
[sven@fedora lbm]$ nix --version
nix (Nix) 2.6.0
[sven@fedora lbm]$ nix flake info github:input-output-hk/bitte/b6f9b26935425a2efadb9b35201c99916ced3a33
warning: 'nix flake info' is a deprecated alias for 'nix flake metadata'
error: input 'cli/nixpkgs' follows a non-existent input 'fenix/nixpkgs'
(use '--show-trace' to show detailed location information)
```

[...]

PR #5713, opened 2021-12-02
Flakes: computeLocks: pass correct LockParent when reusing oldLock

balsoft:
Previously, when we were attempting to reuse the old lockfile
information in the computeLocks function, we have passed the parent of
the current input to the next computeLocks call. This was incorrect,
since the follows are resolved relative to the parent. This caused
issues when we tried to reuse oldLock but couldn't for some
reason (read: mustRefetch is true), in that case the follows were
resolved incorrectly.

Fix this by passing the correct parent, and adding some tests to
prevent this particular regression from happening again.

Closes https://github.com/NixOS/nix/issues/5697

P.S perhaps we should write more tests for the follows behaviour...

CC @nrdxp

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -480,11 +480,16 @@ LockedFlake lockFlake(
                             }
                         }
 
+                        LockParent newParent {
+                            .path = inputPath,
+                            .absolute = false
+                        };
+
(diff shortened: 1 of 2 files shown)


nrdxp:
thanks for the quick fix! :pray:

PR #5715, opened 2021-12-02
Introduce `builtins.groupBy` primop

Infinisil:
This function is very useful in nixpkgs, but its implementation in Nix
itself is rather slow due to it requiring a lot of attribute set and
list appends.

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -5,3 +5,6 @@
 * `nix develop` now has a flag `--unpack` to run `unpackPhase`.
 
 * Lists can now be compared lexicographically using the `<` operator.
+
+* New built-in function: `builtins.groupBy`, with the same functionality as
+  Nixpkgs' `lib.groupBy`, but faster.
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -2928,6 +2928,56 @@ static RegisterPrimOp primop_partition({
     .fun = prim_partition,
 });
 
+static void prim_groupBy(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    state.forceFunction(*args[0], pos);
+    state.forceList(*args[1], pos);
+
+    ValueVectorMap attrs;
+
+    for (auto vElem : args[1]->listItems()) {
+        Value res;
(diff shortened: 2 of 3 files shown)


edolstra:
Looks good, do you have any numbers on the performance improvement?

doc/manual/src/release-notes/rl-next.md
```diff
@@ -5,3 +5,6 @@
 * `nix develop` now has a flag `--unpack` to run `unpackPhase`.
 
 * Lists can now be compared lexicographically using the `<` operator.
+
+* New built-in function: `builtins.groupBy`, with the same functionality as
+  nixpkgs `lib.groupBy`, but faster
```
edolstra, line 10:
```suggestion
  Nixpkgs' `lib.groupBy`, but faster.
```

Infinisil:
> Looks good, do you have any numbers on the performance improvement?

Yup, see https://github.com/NixOS/nixpkgs/pull/147077#issuecomment-984909819

src/libexpr/primops.cc
```diff
@@ -2928,6 +2928,57 @@ static RegisterPrimOp primop_partition({
...
 });
 
+static void prim_groupBy(EvalState & state, const Pos & pos, Value * * args, Value & v)
+{
+    state.forceFunction(*args[0], pos);
+    state.forceList(*args[1], pos);
+
+    ValueVectorMap attrs;
+
+    for (unsigned int n = 0; n < args[1]->listSize(); ++n) {
```
edolstra, line 2938: This can be written using the list iterator now, i.e. `for (auto vElem : args[1]->listItems()) ...`.

Infinisil: Nice, did this

roberth:
@Infinisil @edolstra doesn't this need tests?

gilligan:
@Infinisil Could you add tests for this? There are various tests that have been merged already and some open PR where @andir has been working on covering most primops.

I’m sure andi (or i though andi now has more practice) would also be willing to help out in case you don’t know how.

I just really really think that whenever new builtins are being added there should be *tests*. At the very minimum the code that is presented as example in the documentation.

Infinisil:
Added one in https://github.com/NixOS/nix/pull/5786

PR #5720, opened 2021-12-03
flakes: search up to git or filesystem boundary

tomberek:
While parsing a flakeref, upon not finding a flake.nix, search upwards
until git or filesystem boundary.

Darwin?

@Radvendii @balsoft via: NixUX Hacking Extravaganza

Fixes: https://github.com/NixOS/nix/issues/4612


tests/flake-searching.sh
```diff
@@ -0,0 +1,24 @@
...
+mkdir subdir
+cd subdir
+
+for i in "" . "$PWD" .# .#test; do
+    nix build $i || fail "flake should be found by searching up directories"
+done
+
+for i in "path:$PWD"; do
+    ! nix build $i || fail "flake should not search up directories when using 'path:'"
+done
```
balsoft, line 24: Perhaps also test that it doesn't search after `.git` is found?

The filesystem boundary check might be tricky to test because of permissions required to mount things, so maybe it's ok to leave that one out

balsoft (commented):
Perhaps this should also be documented in the manual

Radvendii:
So, there were some tricky design decisions we came up against that I'm not sure what the right choice is. It essentially boils down to "under what circumstances should we search upwards". Here are some example circumstances:

1. `nix build`
2. `nix build .`
3. `nix build $PWD`
4. `nix build path:$PWD`
5. `inputs.foo.url = "./foo";` inside another flake

The issue si that (1) is pretty clearly a case where we want to search upwards, and (5) is pretty clearly a case where we don't want to search upwards. But anywhere that we draw the line feels a little contrived and is bound to cause some confusion ("why doesn't X work like X'??")

Right now we draw the line between (3) and (4), but I'm also not sure the code is robust in this. As in, I worry that a small change somewhere else in the code that isn't even thinking about this could change whether (4) and (5) get counted.

balsoft:
I'd like to note that we also considered allowing `nix build ''` and `nix build '#hello'` and making those search upwards, but decided it broke the fact that an installable is a valid URL and so wasn't worth it.

Also after a second thought to me it makes the most sense to only search upwards when no installable is provided (so, (1) from @Radvendii 's list) but it would require quite a lot of refactoring so maybe it's not worth it.

regnat:
Isn’t that a duplicate of #5692 ?

Radvendii:
@regnat It is indeed a duplicate :woman_facepalming: whoops. I think the only thing we thought of that @lincolnauster didn't is filesystem boundaries. So maybe we should just add that there and ditch this PR. I don't know if there's a strong reason to prefer one implementation or another.

@balsoft the issue with *only* including `nix build` is that the level of inconvenience between building `defaultPackage` and any other output becomes large. The problem of "I have to know where I am or where the flake is" comes back, but only for outputs besides `defaultPackage`. I think I would find this frustrating.

The solution that they settled on in the other PR is (2), including `nix build .` but nothing beyond that. I can sitll imagine this causing confusion, but I think in practice it might be the best compromise.

regnat:
> I think the only thing we thought of that @lincolnauster didn't is filesystem boundaries

What’s the reason for for searching above fs boundaries? I mean I don’t think there’s a lot of cases where crossing fs boundaries would be useful, but I don’t really see why it would be harmful either

balsoft:
I think the idea is that git does it, so we should also do it :)

doc/manual/src/release-notes/rl-next.md
```diff
@@ -1,4 +1,16 @@
...
+
+* `nix develop` now has a flag `--unpack` to run `unpackPhase`.
+
+* Lists can now be compared lexicographically using the `<` operator.
+
+* New built-in function: `builtins.groupBy`, with the same functionality as
+  Nixpkgs' `lib.groupBy`, but faster.
+
+* Nix now searches for a flake.nix up until git or filesystem boundary.
+
```
thufschmitt, line 14:
```suggestion
* The Nix cli now searches for a flake.nix up until the root of the current git repository or a filesystem boundary rather than just in the current directory
```

tests/flake-searching.sh
thufschmitt: Agreed, that’s worth a few extra tests (esp. given that these are rather cheap)

src/libexpr/flake/flakeref.cc
```diff
@@ -122,6 +122,27 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
...
+                bool found = false;
+                while (path != "/") {
+                    if (pathExists(path + "/flake.nix")) {
+                        found = true;
+                        break;
+                    } else if (pathExists(path + "/.git"))
+                        throw Error("unable to find a flake before encountering git boundary at '%s'", path);
+                    else {
+                        if (lstat(path).st_dev != device)
+                            throw Error("unable to find a flake before encountering filesystem boundary at '%s'", path);
```
thufschmitt, line 139:
```suggestion
                        throw Error("path '%s' is not part of a flake (neither it nor its parent directories contain a 'flake.nix' file)", origPath);
```

thufschmitt (changes requested):
Good overall (and very useful :) ). A couple points in inline comments.

Also, as @Radvendii [points out](https://github.com/NixOS/nix/pull/5720#issuecomment-985639200), we probably don’t want this logic in any case. A good compromise chosen by https://github.com/NixOS/nix/pull/5692 is to only do this search when the path is `.` (so covering `nix build` or `nix build .#foo` but not `nix build /foo/bar` nor `nix build ./foo`)

tomberek:
> 1. `nix build`
> 2. `nix build .`
> 3. `nix build $PWD`
> 4. `nix build path:$PWD`
> 5. `inputs.foo.url = "./foo";` inside another flake

@balsoft refactored and implemented the design such that the searching only occurs when the installable has no explicit scheme. This is more consistent, easier to describe/document, and should be least-surprise.

src/nix/nix.md
```diff
@@ -57,9 +57,33 @@ the Nix store. Here are the recognised types of installables:
...
+    repository and `/foo/bar/baz` contains `flake.nix`, then the flake
+    reference `/foo/bar/baz` is equivalent to
+    `git+file:///foo/bar?dir=baz`. Note that it will only include
+    files indexed by git. In particular, files which are matched by
+    `.gitignore` will not be available in the flake. If this is
+    undesireable, specify `path:<directory>` explicitly;
+  - If the directory is not inside a Git repository, then it is
+    equivalent to `path:<directory>` (see [the `nix flake` manual
+    page](./nix3-flake.md)), which includes the entire contents of the
+    path.
```
thufschmitt, line 86: I find this explanation a bit confused. Here’s another take at it (which might not be better tbh, but gives another data-point):

```suggestion
  When the flake reference is a raw path (a path without any URL
  scheme), it is interpreted as a `path:` or `git+file:` url in the following
  way:
  
  - If the path is within a Git repository, then the url will be of the form
    `git+file://[GIT_REPO_ROOT]?dir=[RELATIVE_FLAKE_DIR_PATH]`
    where `GIT_REPO_ROOT` is the path to the root of the git repository,
    and `RELATIVE_FLAKE_DIR_PATH` is the path (relative to the directory
    root) of the closest parent of the given path that contains a `flake.nix` within
    the git repository.
    If no such directory exists, then Nix will error-out.
    
    Note that the search will only include files indexed by git. In particular, files
    which are matched by `.gitignore` or have never been `git add`-ed will not be
    available in the flake. If this is undesireable, specify `path:<directory>` explicitly;
    
    For example, if `/foo/bar` is a git repository with the following structure:
    ```
   .
   └── baz
       ├── blah
       │   └── file.txt
       └── flake.nix
    ```

  Then `/foo/bar/baz/blah` will resolve to `git+file:///foo/bar?dir=baz`

  - If the supplied path is not a git repository, then the url will have the form
    `path:FLAKE_DIR_PATH` where `FLAKE_DIR_PATH` is the closest parent
    of the supplied path that contains a `flake.nix` file (within the same file-system).
    If no such directory exists, then Nix will error-out.
    
    For example, if `/foo/bar/flake.nix` exists, then `/foo/bar/baz/` will resolve to
   `path:/foo/bar`


```

tomberek: The new explanation is very nice. Thanks @thufschmitt

thufschmitt (approved):

src/libexpr/flake/flakeref.cc
```diff
@@ -122,6 +122,28 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
 
             if (isFlake) {
 
+                if (!allowMissing && !pathExists(path + "/flake.nix")){
+                    notice("path '%s' does not contain a 'flake.nix', searching up",path);
```
thufschmitt, line 126: I’m not entirely convinced by this message. But we’ll see at usage whether it needs to be dropped or not

thufschmitt (approved):

mschwaig:
Git is making it's searching upward behavior more strict right now due to `CVE-2022-24765` .
See https://github.blog/2022-04-12-git-security-vulnerability-announced/ for more info about that vulnerability.

Their new behavior is to stop traversal at the point where ownership of the directory changes away from the current user.
Searching upward is a tricky feature from a security POV, because as an attacker if
* you expect that some upward-searching command that basically executes code
  - will be run inside the folder `A/B` and
  - not find the file it's looking for inside `A/B` and at the same time
* you can write the file the command is looking for into `A`

That way you can turn being able to write a specific file into code execution.

This is a problem for `git` because while at its core git does not provide code execution as a feature, there is a lot of supplementary stuff like nice git-specific shell prompts that make it happen.

I would expect this issue to actually be worse for nix flakes, where code execution is the whole point of many operations.

edolstra:
Yes, stopping at a directory where ownership changes is a good idea. In particular, we don't want to read somebody else's `flake.nix` from `/tmp`.

PR #5722, opened 2021-12-03
update flake.nix to 21.11

Mic92:


regnat (approved):

flake.nix
```diff
@@ -1,7 +1,7 @@
 {
   description = "The purely functional package manager";
 
-  inputs.nixpkgs.url = "nixpkgs/nixos-21.05-small";
+  inputs.nixpkgs.url = "nixpkgs/nixos-21.11-small";
   inputs.lowdown-src = { url = "github:kristapsdz/lowdown"; flake = false; };
```
Mic92, line 5: Not sure if this is still required?

edolstra:
Unfortunately I had to revert this for now because util-linux is broken in the musl build (https://hydra.nixos.org/build/161448992). So we should fix that in nixpkgs first.

alyssais:
We already did, but the proper fix is https://github.com/NixOS/nixpkgs/commit/2ebeb02a99aab4fc6813fe401e2d375a5c00bc8a, which is a mass rebuild, so I'm not sure about backporting it.  Thoughts?

Mic92:
Mhm. Would targeting unstable for the musl build acceptable?

PR #5724, opened 2021-12-03
Allow -h in addition to --help

knedlsepp:
I'm currently trying to figure out flakes and the shiny new `nix` command.
So far I really like it, but I'm somewhat missing the very common `-h` option for help.
I swear I did run `nix flake -h` or `nix run -h` at least ten times today despite knowing better.
I guess this is just muscle-memory.

### DIFF
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -74,6 +74,7 @@ struct NixArgs : virtual MultiCommand, virtual MixCommonArgs
 
         addFlag({
             .longName = "help",
+            .shortName = 'h',
             .description = "Show usage information.",
             .handler = {[&]() { throw HelpRequested(); }},
         });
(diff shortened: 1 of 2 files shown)


regnat (approved):
Thanks!

The small breaking change should be acceptable given the experimental nature of the CLI (and the fact that `--human-redable` isn’t something I’d expect to see in a non-interactive shell script)

edolstra:
This was discussed a while ago in the UX meetings, but we decided not to have `-h` since it's far from standard (e.g. GNU and git don't have it).

edolstra:
Some previous discussion: #3760

knedlsepp:
> Some previous discussion: #3760

Thx for linking the reasons this was rejected before. I missed that issue.

PR #5726, opened 2021-12-04
launchd: Set NumberOfFiles to 4096

azuwis:
The default maxfiles on macOS 11 and macOS 12 is 256, which is too low
for nix to work:

```
$ launchctl limit maxfiles
	maxfiles    256            unlimited
```

Set NumberOfFiles of nix-daemon to 4096 to avoid `Too many open files`
error.


hexagonal-sun:
This fixed an issue that stopped me from bootstrapping my initial Darwin config.  Thanks @azuwis .

roberth:
This is the same number chosen by `nix-darwin`:

https://github.com/LnL7/nix-darwin/blob/a7492a8c76dcc702d0a65cd820a5f9baa8702684/modules/services/nix-daemon.nix#L55

:+1:

PR #5740, opened 2021-12-06
libstore: fix buffer resize in `addToStoreFromDump`

Ma27:
The original issue I'm aiming to solve were hash-mismatches on
remote-builds like this:

      /nix/store/2nwv1fm5bs0a2sgb2a014gv7hgfaj3f7-ruby2.7.4-vagrant_cloud-3.0.5.drv
    building '/nix/store/2nwv1fm5bs0a2sgb2a014gv7hgfaj3f7-ruby2.7.4-vagrant_cloud-3.0.5.drv' on 'ssh-ng://builder-ng'...
    copying 1 paths...
    copying path '/nix/store/j3wsx0in5hgnl3sr91i8h19b74w5hcz5-vagrant_cloud-3.0.5.gem' to 'ssh-ng://builder-ng'...
    error: ca hash mismatch importing path '/nix/store/j3wsx0in5hgnl3sr91i8h19b74w5hcz5-vagrant_cloud-3.0.5.gem';
             specified: sha256:0np0d8rjca130si5iaxasbqmfbbx4l3kd9mxdsa3p5mqiia7za3b
             got:       sha256:14iwaq2d9ai4x36w2vd4gv0zwpa9yj5nv8l6rxqfs5njd69jv90h
    error: builder for '/nix/store/2nwv1fm5bs0a2sgb2a014gv7hgfaj3f7-ruby2.7.4-vagrant_cloud-3.0.5.drv' failed with exit code 1

In contrast to my initial assumption this isn't related to hashing
changes in the CAS-implementation, but actually a problem when using
`nix-prefetch-url(1)` from Nix 2.3 with a daemon from Nix 2.4.

Basically the following happens:

* `nix-prefetch-url` from 2.3 transmits the file to the daemon using
  `wopAddToStore`.
* The daemon (`LocalStore::addToStoreFromDump` to be precise) checks
  whether the entire file can be kept in memory before writing it to the
  disk.

  This happens by writing into a buffer in chunks of 2^16 bytes. If an
  EOF is reached before the buffer exceeds `settings.narBufferSize`, the
  file can be kept in memory entirely.
* When the source is entirely read in, a last iteration will happen
  which results in the EOF. However, in this round the buffer is
  increased by 2^16 bytes again without downsizing it on the EOF (this
  only happens in the rounds without an EOF).
* The result is that the files in the store now have the size
  `(original_size + 2^16)` where the excess bytes are null-bytes at the
  end of the file.
* Because the file has different contents now than computed via the hash
  sink, a hash-mismatch occurs as soon as the file is copied to a remote
  system where the hash is apparently re-validated (in fact I could work
  around the issue from above by declaring `--builders ''`).

Resizing the buffer back to the original size in case of an EOF solves
the problem for me. I also added a regression-test that fails without
the change in `LocalStore::addToStoreFromDump` (in this case the size of
the imported file would be `65542` rather than `6`).

The issue is only observable with a 2.3 `nix-prefetch-url(1)` because
the 2.4 implementation uses `wopAddToStoreNar` which imports paths
slightly different.

cc @Ericson2314 @edolstra

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -181,7 +181,14 @@
           mkdir -p $out
         '';
 
-        installCheckPhase = "make installcheck -j$NIX_BUILD_CORES -l$NIX_BUILD_CORES";
+        installCheckPhase = ''
+          export PATH=${runCommand "nix-2.3-prefetch" { } ''
+            mkdir -p $out/bin
+            cp ${nixStable}/bin/nix-prefetch-url $out/bin/nix-prefetch-url-2.3
+            $out/bin/nix-prefetch-url-2.3 --version | grep -q 2.3
+          ''}/bin:$PATH
+          make installcheck -j$NIX_BUILD_CORES -l$NIX_BUILD_CORES
+        '';
       };
 
       binaryTarball = buildPackages: nix: pkgs: let
@@ -310,6 +317,13 @@
             echo "doc manual $doc/share/doc/nix/manual" >> $doc/nix-support/hydra-build-products
           '';
 
+          preInstallCheck = ''
+            export PATH=${runCommand "nix-2.3-prefetch" { } ''
+              mkdir -p $out/bin
+              cp ${nixStable}/bin/nix-prefetch-url $out/bin/nix-prefetch-url-2.3
+              $out/bin/nix-prefetch-url-2.3 --version | grep -q 2.3
+            ''}/bin:$PATH
+          '';
           doInstallCheck = true;
           installCheckFlags = "sysconfdir=$(out)/etc";
 
@@ -627,6 +641,12 @@
               PATH=$prefix/bin:$PATH
               unset PYTHONPATH
               export MANPATH=$out/share/man:$MANPATH
+
+              export PATH=${runCommand "nix-2.3-prefetch" { } ''
+                mkdir -p $out/bin
+                cp ${nixStable}/bin/nix-prefetch-url $out/bin/nix-prefetch-url-2.3
+                $out/bin/nix-prefetch-url-2.3 --version | grep -q 2.3
+              ''}/bin:$PATH
             '';
         });
 
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1337,6 +1337,7 @@ StorePath LocalStore::addToStoreFromDump(Source & source0, const string & name,
             got = source.read(dump.data() + oldSize, want);
         } catch (EndOfFile &) {
             inMemory = true;
+            dump.resize(oldSize);
             break;
         }
         dump.resize(oldSize + got);
--- a/tests/local.mk
+++ b/tests/local.mk
@@ -1,5 +1,6 @@
 nix_tests = \
   hash.sh lang.sh add.sh simple.sh dependencies.sh \
+  nix-prefetch-url-2.3.sh \
   config.sh \
   gc.sh \
   ca/gc.sh \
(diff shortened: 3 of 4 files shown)


flake.nix
```diff
@@ -310,6 +317,13 @@
             echo "doc manual $doc/share/doc/nix/manual" >> $doc/nix-support/hydra-build-products
           '';
 
+          preInstallCheck = ''
+            export PATH=${runCommand "nix-2.3-prefetch" { } ''
+              mkdir -p $out/bin
+              cp ${nixStable}/bin/nix-prefetch-url $out/bin/nix-prefetch-url-2.3
+              $out/bin/nix-prefetch-url-2.3 --version | grep -q 2.3
+            ''}/bin:$PATH
+          '';
```
Ma27, line 326: This isn't very pretty, but I wanted to hack together a regression-test now. Any suggestions on how to implement that a little bit nicer? :)

Ericson2314:
Oh oof i just fixed this the other day with https://github.com/NixOS/nix/commit/2f5c913d4a21c60799290d6efe769d188c57b906. In https://github.com/NixOS/nix/pull/5572 I hope to test against stable versions more broadly, but yes we might need a new test anyways as I didn't see this failure when working on that PR.

Ma27:
So I guess I should leave out the fix, but the tests are still worth to be included?

Ericson2314:
@Ma27 Yeah, I and if you could reproduce the problem and fix with just the "plain" test with my https://github.com/NixOS/nix/pull/5572 doing the cross-version stuff, that would be good to know, too.

flake.nix
```diff
@@ -181,7 +181,14 @@
...
         '';
 
-        installCheckPhase = "make installcheck -j$NIX_BUILD_CORES -l$NIX_BUILD_CORES";
+        installCheckPhase = ''
+          export PATH=${runCommand "nix-2.3-prefetch" { } ''
+            mkdir -p $out/bin
+            cp ${nixStable}/bin/nix-prefetch-url $out/bin/nix-prefetch-url-2.3
+            $out/bin/nix-prefetch-url-2.3 --version | grep -q 2.3
+          ''}/bin:$PATH
+          make installcheck -j$NIX_BUILD_CORES -l$NIX_BUILD_CORES
```
edolstra, line 190: This seems very complicated. `testNixVersions` already takes a `client` and `daemon` version, and now we're adding *another* version into the mix (`nixStable`), passed in a very hacky way (by adding a `nix-2.3-prefetch` derivation to `PATH`).

Shouldn't the client/daemon matrix be able to test this? (E.g. by enabling `againstLatestStable`)

Ma27: Yeah, that should certainly be possible and as mentioned above, I mostly hacked this together to get a working regression test for others :) 

Is the `testNixVersions` thing somehow easily usable for dev purposes btw? :) I'd still like to be able to test something like this `make tests/foobar.sh.test`.

Ericson2314: @Ma27 yes it is. You can define `NIX_{DAEMON,CLIENT}_PACKAGE` yourself.

Ma27:
Closing. The actual bug got fixed by @Ericson2314 and I'm not really motivated to add a test with a second Nix 2.3 running. Also, given that 22.05 doesn't have 2.3 by default anymore, so the problematic code-paths aren't even used anymore.

If we should implement/expand (don't know about the current status here) cross-version daemon tests (to check for backwards compat of the binary protocols) in general is a different topic, though.

PR #5747, opened 2021-12-08
Collect eval memory before building with nix-build

### LINKED ISSUE #5200: `nix-build` is not freeing eval memory during long builds
**Describe the bug**

On a fairly simple NixOS system `nix-build` currently allocates about 1.8 - 2GiB of RAM during system config eval. The memory usage is not great but not the topic of this issue. The problem I am seeing is that on low-end devices those 2GiB of RAM that are being allocated to eval the system configuration are a significant portion of the systems resources.

**Steps To Reproduce**

1. Run `nix-build foo.nix` where `foo.nix` defines a NixOS system.
2. Observe the memory growth of the `nix-build` process while the eval is running.
3. As soon as it starts building paths (even just some symlinks, configuration files, ...) it should have free'ed the eval memory but doesn't.
4. Depending on the nix expression and system config you might run out of memory if your system has <4GiB of total RAM available.

**Expected behavior**

The eval memory should not be allocated anymore once the actual build starts.

2.3.15 & 2.4.pre20210802_47e96bb

**Additional context**

`nix-build` doesn't require all of the eval data to stay in memory while running the build. From looking at the code it seems like it needs a list of drv paths and output mappings. None of the inputs to `buildPaths` in the `nix-build.cc` file require the actual eval state to be still around.

Since our current eval situation isn't ideal (with regards to GC, memory leaks, ...) there are generally three solutions I can see:

a) optimize the eval code such that more of the memory is actually GC tracked and release the eval state, the DrvInfos, Expressions lists and ensure that `GC_gcollect` is called once before starting the build.
b) Move all the evalulation into a child process that is forked off from `nix-build` and returns a list of target (output paths, drvs, output mappings, ...). This way the memory consumption will be tied to the lifetime of the children and we don't really have to care about it at all. By the time the build start the eval memory will be free again.
c) Do nothing and tell users to deploy (more) swap such that the memory can be swapped to disk instead of staying in RAM during the entire build. Less ideal but something that can be done today.

I've attempted a) and b) to varying degrees. While refactoring the way our eval code creates lists of expressions (and other currently untracked allocations) has resulted in a few hundred megabytes less memory consumption it introduced random memory failures. Apparently the GC is a bit too eager in some cases or I've missed a spot where we manually free memory.

With b) I've probably not spent enough time tracking down all the global state that has to be restored/closed before/after the child is spawned. It worked well for some simple `nix-shell` invocations but failed during build of an entire NixOS system.

Regarding c) I am not actually sure the memory is inactive enough to be swapped. I haven't tested that theory but I'd assume that this works in most of the cases.

andir:
```
nix-build: free eval memory before building paths

This ensures that we free some of the memory that was allocated during
the eval of the build target. In particular this will free the Boehm GC
managed memory allocations.

On my sample system I was able to reduce the memory consumption from
550MiB to about 200MiB when building the NixOS GNOME test.

The memory did initially climb to 550MiB and then shrink down to 200MiB
once the builds were kicked off.

This means I can now run more build processes (without having to resort
to swap) on resource constrained systems.

Fixes #5200 (for me)
```

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -579,6 +579,19 @@ static void main_nix_build(int argc, char * * argv)
             else
                 drvMap[drvPath] = {drvMap.size(), {outputName}};
         }
+        // Can only free the eval state after all the DrvInfo's have been freed
+        // as they carry a reference to the eval state instance (which isn't
+        // managed in any way).
+        drvs = {};
+        exprs = {};
+        state = nullptr;
+
+        // Collect garbage a couple of times as per Boehm GC docs that suggest
+        // to run it more than just once.
+        // Running it 10 times seems reasonable and doesn't appear to take too
+        // much time.
+        for (int i = 0; i < 10; i++)
+            GC_gcollect();
 
         buildPaths(store, evalStore, buildMode, pathsToBuild, dryRun);
 
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -74,6 +74,47 @@ std::vector<string> shellwords(const string & s)
     return res;
 }
 
+void buildPaths(ref<Store> store, ref<Store> evalStore, BuildMode buildMode, const std::vector<StorePathWithOutputs> & paths0, bool dryRun) {
+    auto paths = toDerivedPaths(paths0);
+    /* Note: we do this even when !printMissing to efficiently
+       fetch binary cache data. */
+    uint64_t downloadSize, narSize;
+    StorePathSet willBuild, willSubstitute, unknown;
+    store->queryMissing(paths,
+        willBuild, willSubstitute, unknown, downloadSize, narSize);
+
+    if (settings.printMissing)
+        printMissing(ref<Store>(store), willBuild, willSubstitute, unknown, downloadSize, narSize);
+
+    if (!dryRun)
+        store->buildPaths(paths, buildMode, evalStore);
+}
+
+
+std::map<std::string, std::string> getShellEnv(ref<Store> store, bool pure, std::set<string> & retainVarsInPure) {
+    auto env = getEnv();
+    auto tmp = getEnv("TMPDIR");
+    if (!tmp) tmp = getEnv("XDG_RUNTIME_DIR").value_or("/tmp");
+
+    if (pure) {
+        decltype(env) newEnv;
+        for (auto & i : env)
+            if (retainVarsInPure.count(i.first))
+                newEnv.emplace(i);
+        env = newEnv;
+        // NixOS hack: prevent /etc/bashrc from sourcing /etc/profile.
+        env["__ETC_PROFILE_SOURCED"] = "1";
+    }
+
+    env["NIX_BUILD_TOP"] = env["TMPDIR"] = env["TEMPDIR"] = env["TMP"] = env["TEMP"] = *tmp;
+    env["NIX_STORE"] = store->storeDir;
+    env["NIX_BUILD_CORES"] = std::to_string(settings.buildCores);
+
+    return env;
+}
+
+
+
 static void main_nix_build(int argc, char * * argv)
 {
     auto dryRun = false;
@@ -102,6 +143,8 @@ static void main_nix_build(int argc, char * * argv)
     AutoDelete tmpDir(createTempDir("", myName));
 
     std::string outLink = "./result";
+
+    DrvInfos drvs;
 
     // List of environment variables kept for --pure
     std::set<string> keepVars{
@@ -286,7 +329,6 @@ static void main_nix_build(int argc, char * * argv)
     if (runEnv)
         setenv("IN_NIX_SHELL", pure ? "pure" : "impure", 1);
 
-    DrvInfos drvs;
 
     /* Parse the expressions. */
     std::vector<Expr *> exprs;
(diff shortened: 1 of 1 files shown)


edolstra:
It would be nice to have some numbers on the performance cost. Doing a GC slows down `nix-build`, so ideally this is something we only do in low-memory conditions.

I'm also concerned with doing the GC 10 times, which seems strange. I wouldn't expect a mark-and-sweep collector to free more memory on subsequent runs. The main risk here is that if there is any (false) pointer that keeps the `pkgs` data structure alive, then we'll be doing 10 sweeps over a multi-gigabyte heap without freeing up a lot of memory. (False pointers can easily happen, see for instance 455d1f01d041baf58abc3c9baf5e59ee054d9c9b.) In fact, it might not free any memory, since Boehm is not a compacting collector, so due to heap fragmentation it probably wouldn't be able to return a lot of memory to the OS.

BTW, as a workaround, you can use `nix-store -r $(nix-instantiate ...)`. `nix-build` used to be implemented that way.

andir:
> It would be nice to have some numbers on the performance cost. Doing a GC slows down `nix-build`, so ideally this is something we only do in low-memory conditions.
 
> I'm also concerned with doing the GC 10 times, which seems strange. I wouldn't expect a mark-and-sweep collector to free more memory on subsequent runs.

I do agree. I just went by what was in some of the related BoehmGC discussions and documentation. IIRC I tried with just one GC run and that wasn't sufficient to see a difference when I original tested this (~6M ago).

> The main risk here is that if there is any (false) pointer that keeps the `pkgs` data structure alive, then we'll be doing 10 sweeps over a multi-gigabyte heap without freeing up a lot of memory. (False pointers can easily happen, see for instance [455d1f0](https://github.com/NixOS/nix/commit/455d1f01d041baf58abc3c9baf5e59ee054d9c9b).)

But then that is a bug that has to be fixed. It wouldn't surface right now because we don't care about the memory. We probably should care about the memory and fix the issues once we observe them.

> In fact, it might not free any memory, since Boehm is not a compacting collector, so due to heap fragmentation it probably wouldn't be able to return a lot of memory to the OS.

Apparently it does so anyway.


> BTW, as a workaround, you can use `nix-store -r $(nix-instantiate ...)`. `nix-build` used to be implemented that way.

I know. We discussed this in #5200. I dont' think that is a valid workaround for regular uses that just want to call `nixos-rebuild` to rebuild their system configuration. It would just be yet another oddity while learning how to deal with nix and the nix language.

andir:
Here are the results of some basic build testing. I've built the attribute before running the test. There were no actual builds performed during the testing.

"new" is at commit ba655d46171b9eae5d5a3c54e47a0d13c0861ac6
"old" is at commit f66923efdefa5160222b6d6ef0dd9fa822242116
"new-one-gc" is the same as new but with a signle GC round.

The nixpkgs checkout was a2e281f5770247855b85d70c43454ba5bff34613.

```
Benchmark 1: ./old/bin/nix-build ~/dev/nixos/nixpkgs/nixos/tests/simple.nix
  Time (mean ± σ):      2.275 s ±  0.029 s    [User: 1.710 s, System: 0.292 s]
  Range (min … max):    2.234 s …  2.339 s    10 runs
 
Benchmark 2: ./new/bin/nix-build ~/dev/nixos/nixpkgs/nixos/tests/simple.nix
  Time (mean ± σ):      2.544 s ±  0.127 s    [User: 2.962 s, System: 0.309 s]
  Range (min … max):    2.430 s …  2.724 s    10 runs
 
Benchmark 3: ./new_one_gc/bin/nix-build ~/dev/nixos/nixpkgs/nixos/tests/simple.nix
  Time (mean ± σ):      2.441 s ±  0.033 s    [User: 2.089 s, System: 0.303 s]
  Range (min … max):    2.395 s …  2.519 s    10 runs
 
Summary
  './old/bin/nix-build ~/dev/nixos/nixpkgs/nixos/tests/simple.nix' ran
    1.07 ± 0.02 times faster than './new_one_gc/bin/nix-build ~/dev/nixos/nixpkgs/nixos/tests/simple.nix'
    1.12 ± 0.06 times faster than './new/bin/nix-build ~/dev/nixos/nixpkgs/nixos/tests/simple.nix'
```

It is indeed slower. Personally I'd prefer correctness (no leaked memory, system can reclaim space) over a fraction of a second in interactive evals.

Mic92:
> BTW, as a workaround, you can use `nix-store -r $(nix-instantiate ...)`. `nix-build` used to be implemented that way.

Could this be racy when garbage collection is used?

src/nix-build/nix-build.cc
```diff
@@ -570,8 +579,21 @@ static void main_nix_build(int argc, char * * argv)
             else
                 drvMap[drvPath] = {drvMap.size(), {outputName}};
         }
```
Mic92, line 581: Could we do the evaluation in a child process and reclaim the memory this way?

edolstra:
> Could this be racy when garbage collection is used?

Yes, if you want to do it right you have to do `nix-store -r ... --gc-root ...`.

src/nix-build/nix-build.cc
Mic92: I think this option was discussed but dismissed.

thufschmitt:
Since these GC run aren’t (hopefully) changing anything to the semantics of the program, could it make sense to start them in a different thread while the build is starting? That way the speed impact would be minimal

roberth:
Some possible mitigations for the performance cost:
a. reduce the priority of the thread that performs the GC
b. don't GC if there's no memory pressure; `while (!memoryPressure() && !interrupted) { sleep(1s); }; if (!interrupted) gc()`

Regardless of this GC may still incur a cost by increasing utilization of the memory bus, which may very well be the bottleneck of a build process, so (a) is not a complete fix, but (b) is a "complete" if the system has plenty of memory.

> Since these GC run aren’t (hopefully) changing anything to the semantics of the program, could it make sense to start them in a different thread while the build is starting? That way the speed impact would be minimal

Some things to make sure
 - Boehm GC doesn't stop threads involved in the monitoring and scheduling of the builds
 - The threads don't reference any GC-allocated memory. If they do, provide a garbage collection root that isn't the thread that uses them; that thread can't serve as a root because the GC would have to pause that thread

tomberek:

PR #5749, opened 2021-12-09
Better diagnostics if no valid signature found

GavinRay97:
I downloaded Nix tonight, and immediately broke it by accidentally removing the default binary caching.
After figuring this out, I also failed to fix it properly, due to using the wrong key for Nix's default binary cache

If the diagnostic message would have been clearer about what/where a "signature" for a "substituter" is + comes from, it probably would have saved me a few hours.

Maybe we can save other noobs the same pain?

---

If you're wondering how on earth someone could be this dumb, the process was roughly:

- "I want to try Nix, and I have to write something in Haskell. Nix works great with Haskell right?"
- "Online, people say to use this `stack` thing, and it has Nix integration docs."
- Proceed to see this in the docs
    - image
- Neither `/etc/nix/nix.conf` exists, or `~/.config/nix/nix.conf`
- So I create `~/.config/nix/nix.conf` with the following content:

```conf
trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
substituters = https://hydra.iohk.io
```

- Now I will install `stack` with Nix:
```
user@MSI:~$ nix-env -f "<nixpkgs>" -iA stack
installing 'stack-2.7.3'

these 698 derivations will be built:
  /nix/store/dhvmh2hb7900z4ki92cfiwap7rhhnqfv-bootstrap-stage0-stdenv-linux.drv
  ...
these 587 paths will be fetched (1125.72 MiB download, 2348.57 MiB unpacked):
  /nix/store/02z0ga1d6p21zkvcwiz8s42qwdzdzl72-cmake-3.21.2.tar.gz
```
- Wait a minute, this doesn't seem right?! That, or Nix sucks.
- Bug someone on Matrix
- They tell me I've disabled the default cache by creating `nix.conf` without `https://cache.nixos.org`
- I try to put it back, but manage to screw up the signature on one of the keys
- Now I just get a bunch of this:

image

- "Wtf does this mean?!"
- Google more, eventually fix the signature
- Why didn't the warning just tell me what was wrong?!
- Proceed to file this PR

### DIFF
--- a/src/libstore/build/substitution-goal.cc
+++ b/src/libstore/build/substitution-goal.cc
@@ -138,10 +138,8 @@ void PathSubstitutionGoal::tryNext()
        only after we've downloaded the path. */
     if (!sub->isTrusted && worker.store.pathInfoIsUntrusted(*info))
     {
-        warn("substituter '%s' does not have a valid signature for path '%s'",
-            sub->getUri(), worker.store.printStorePath(storePath));
-        warn("verify that your nix.conf contains a correct signature in 'trusted-public-keys' for %s",
-            sub->getUri());
+        warn("the substitute for '%s' from '%s' is not signed by any of the keys in 'trusted-public-keys'",
+            worker.store.printStorePath(storePath), sub->getUri());
         tryNext();
         return;
     }
--- a/src/libstore/build/substitution-goal.cc
+++ b/src/libstore/build/substitution-goal.cc
@@ -140,6 +140,8 @@ void PathSubstitutionGoal::tryNext()
     {
         warn("substituter '%s' does not have a valid signature for path '%s'",
             sub->getUri(), worker.store.printStorePath(storePath));
+        warn("verify that your nix.conf contains a correct signature in 'trusted-public-keys' for %s",
+            sub->getUri());
         tryNext();
         return;
     }


GavinRay97:
A secondary PR probably ought to be an enabled-by-default warning if a user has zero binary cache sources enabled, but I'm not familiar enough with the codebase to file that PR as quickly.

Whatever (odd) scenarios in which someone would want zero binary caching, I'm sure a toggleable warning message to remind you "Hey, you are going to compile the world from source", wouldn't kill someone.

src/libstore/build/substitution-goal.cc
```diff
@@ -140,6 +140,8 @@ void PathSubstitutionGoal::tryNext()
     {
         warn("substituter '%s' does not have a valid signature for path '%s'",
             sub->getUri(), worker.store.printStorePath(storePath));
+        warn("verify that your nix.conf contains a correct signature in 'trusted-public-keys' for %s",
+            sub->getUri());
```
edolstra, line 144: This message is not quite correct: `nix.conf` does not contain signatures but public keys; and the public keys are not *for* a specific substituter (so the `for %s` is misleading). I think it's better to combine the two warnings into something like:
```
warn("the substitute for '%s' from '%s' is not signed by any of the keys in 'trusted-public-keys'",
            worker.store.printStorePath(storePath), sub->getUri());
```
Or even better, we could show the key name from the signature (e.g. `'trusted-public-keys' does not contain a key for 'cache.nixos.org-1'`).

edolstra (changes requested):

src/libstore/build/substitution-goal.cc
GavinRay97: This makes sense to me -- thanks for the swift review!

> Or even better, we could show the key name from the signature (e.g. 'trusted-public-keys' does not contain a key for 'cache.nixos.org-1').

Yeah this is great. Where are the possible places a user can/could have set `trusted-public-keys` out of curiosity? It might be good to say "This is where/how you can fix it" if you're starting from ground zero with nix.

GavinRay97: While I'm at this, I will try to fix this error too if that's alright.
Conversation from a Nix/NixOS channel in a developer slack a while ago

Swann = @evertedsphere

image

edolstra: `trusted-public-keys` can be set in several configuration files (global, user's home directory), `flake.nix`, on the command line, and in an environment variable. So it's probably best not to refer to `nix.conf` specifically in the error message.

PR #5753, opened 2021-12-09
Clean up daemon handling in the tests

Ericson2314:
# Motivation
`init.sh` is tested on its own. We used to do that. I deleted it in 4720853129b6866775edd9f90ad6f10701f98a3c but I am not sure why. Better to just restore it; at one point working on this every other test passed, so seems good to check whether `init.sh` can be run twice.

We don't *need* to run `init.sh` twice, but I want to try to make our tests as robust as possible so that manual debugging (where tests for better or worse might be run ways that we didn't expect) is less fragile.

------------

Split `common.sh` into the vars and functions definitions vs starting
the daemon (and possibly other initialization logic). This way,
`init.sh` can just `source` the former. Trying to start the daemon
before `nix.conf` is written will fail because `nix daemon` requires
`--experimental-features 'nix-command'`.

`killDaemon` is idempotent, so it's safe to call when no daemon is
running.

`startDaemon` and `killDaemon` use the PID (which is now exported to
subshells) to decide whether there is work to be done, rather than
`NIX_REMOTE`, which might conceivably be set differently even if a
daemon is running.

`startDaemon` and `killDaemon` can save/restore the old `NIX_REMOTE` as
`NIX_REMOTE_OLD`.

`init.sh` kills daemon before deleting everything (including the daemon
socket).

# Context

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests

### DIFF
--- a/tests/init.sh
+++ b/tests/init.sh
@@ -1,8 +1,14 @@
+set -e -o pipefail
+
 source common.sh
 
 test -n "$TEST_ROOT"
 if test -d "$TEST_ROOT"; then
     chmod -R u+w "$TEST_ROOT"
+    # We would delete any daemon socket, so let's stop the daemon first.
+    if [[ -n "${NIX_DAEMON_PACKAGE:-}" ]]; then
+        killDaemon
+    fi
     rm -rf "$TEST_ROOT"
 fi
 mkdir "$TEST_ROOT"
--- a/tests/init.sh
+++ b/tests/init.sh
@@ -1,14 +1,13 @@
-set -e -o pipefail
+set -eu -o pipefail
 
-source common.sh
+# Don't start the daemon
+source common/vars-and-functions.sh
 
 test -n "$TEST_ROOT"
 if test -d "$TEST_ROOT"; then
     chmod -R u+w "$TEST_ROOT"
     # We would delete any daemon socket, so let's stop the daemon first.
-    if [[ -n "${NIX_DAEMON_PACKAGE:-}" ]]; then
-        killDaemon
-    fi
+    killDaemon
     rm -rf "$TEST_ROOT"
 fi
 mkdir "$TEST_ROOT"
--- a/.gitignore
+++ b/.gitignore
@@ -75,7 +75,7 @@ perl/Makefile.config
 
 # /tests/
 /tests/test-tmp
-/tests/common.sh
+/tests/common/vars-and-functions.sh
 /tests/result*
 /tests/restricted-innocent
 /tests/shell
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -1,0 +1,12 @@
+set -e
+
+if [[ -z "${COMMON_SH_SOURCED-}" ]]; then
+
+COMMON_SH_SOURCED=1
+
+source "$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")/common/vars-and-functions.sh"
+if [[ -n "${NIX_DAEMON_PACKAGE:-}" ]]; then
+    startDaemon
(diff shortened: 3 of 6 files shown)


edolstra:
What was the issue here? I re-run `init.sh` all the time without any problems.

Ericson2314:
OK updated with lots more stuff. Curious if CI will agree with me, but seems good locally.

Ericson2314:
macOS build timed it out?

abathur:
Probably the flaky build issue in #3605

Ericson2314:
Thanks

Ericson2314:
Rebased, hoping won't get that spurious error this time.

Ericson2314:
Yay it did!

Ping @edolstra and @regnat

Ericson2314:
@edolstra Does this look good to you? I would like to start pushing on this and other testing improvements again prior to getting back to fun feature work like RFC 92, so we have a firmer foundation before any more (potentially destabilizing) new feature.

tomberek (approved):
Test failures (repair.sh.test) are unrelated. Re-based onto master and functions as expected.

Ericson2314:
@thufschmitt rebased!

tests/init.sh
```diff
@@ -1,8 +1,15 @@
-source common.sh
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+# Don't start the daemon
+source common/vars-and-functions.sh
```
Ericson2314, line 6: @thufschmitt to answer the question you asked on IRC a bit better, I didn't put the starting the daemon in init for these reasons:

1. Running tests without recalling init should still work. if one wants to examine state: the test itself starts and stops the daemon, rather than something else starting and it stopping.
2. `init.sh` now stops any rougue daemon, and never starts any. It is just to reset the state. 

The file is split because `init.sh` still needs the env vars.

thufschmitt: Oh yes of course, it’s nice to be able to manually run `init.sh` once and then the test (I actually do that all the time)

Ericson2314:
@edolstra @thufschmitt approved. Is this OK with you?

Ericson2314:
@edolstra pinging again. It really would be good to merge this. It doesn't commit us to doing any more testing, it just makes the existing testing more robust.

Ericson2314:
I was inspired to return to this because #7600 has a devilishly hard to debug issue because it only cropped up in the daemon tests, and I was reminded how flaky that stuff is today.

In the process I found that the history had become quite messy, and so I am cleaning it up. This is not just the first commit in what was a few cleanups.

tests/init.sh
```diff
@@ -1,8 +1,14 @@
+set -e -o pipefail
+
 source common.sh
 
 test -n "$TEST_ROOT"
 if test -d "$TEST_ROOT"; then
     chmod -R u+w "$TEST_ROOT"
+    # We would delete any daemon socket, so let's stop the daemon first.
+    if [[ -n "${NIX_DAEMON_PACKAGE:-}" ]]; then
```
Ericson2314, line 9: This is not the ideal condition for this, but it works for now. Another PR will improve it.

thufschmitt (approved):
This has been stalled for too long, indeed. Let's merge!

thufschmitt:
Well, now there are some conflicts :upside_down_face:

Ericson2314:
Thanks @thufschmitt. For some reason when I rebased --- no conflicts!

roberth (approved):

Issue #5302, opened 2021-09-28
Nix flakes always thinks worktree is dirty in github actions pull request builds


arianvp:
**Describe the bug**

Nix flakes keeps saying my tree is dirty; eventhough I'm in a clean detached HEAD

**Steps To Reproduce**

https://github.com/arianvp/webauthn-oidc/blob/master/.github/workflows/build.yml#L33-L36

```
Run git status
  git status
  git diff-index HEAD
  nix build --print-build-logs --log-format raw .#
  nix build --print-build-logs --log-format raw --rebuild # Build again to check reproducibility
  shell: /bin/bash -e {0}
HEAD detached at pull/3/merge
nothing to commit, working tree clean
warning: Git tree '/home/runner/work/webauthn-oidc/webauthn-oidc' is dirty
warning: Git tree '/home/runner/work/webauthn-oidc/webauthn-oidc' is dirty
```

**Expected behavior**

The work tree is considered not dirty

**`nix-env --version` output**

**Additional context**

regnat:
I’m not sure why this happens, but as a workaround you can use `nix build .\?rev=$(git rev-parse HEAD)` to force it to use the version from the index

TransLunarInjection:
Probably related to #4140

Issue #4140, opened 2020-10-12
'git diff-index HEAD' does not detect dirty->clean transitions


edolstra:
**Describe the bug**

We use `git diff-index HEAD` to detect whether a tree is dirty. However, this doesn't always work. For example:

```console
$ nix flake info --json | jq .url
"git+file:///home/eelco/Dev/dwarffs?ref=master&rev=d181d714fd36eb06f4992a1997cd5601e26db8f5"

$ nano flake.nix # make it dirty

$ nix flake info --json | jq .url
warning: Git tree '/home/eelco/Dev/dwarffs' is dirty
"git+file:///home/eelco/Dev/dwarffs"

$ nano flake.nix # remove the edits

$ nix flake info --json | jq .url
warning: Git tree '/home/eelco/Dev/dwarffs' is dirty
"git+file:///home/eelco/Dev/dwarffs"

$ git diff-index HEAD
:100644 100644 c4a857d5657f4e01a29e9c753db95ffd00675c0e 0000000000000000000000000000000000000000 M      flake.nix

$ git diff

$ git diff-index HEAD

$ nix flake info --json | jq .url
"git+file:///home/eelco/Dev/dwarffs?ref=master&rev=d181d714fd36eb06f4992a1997cd5601e26db8f5"
```

**`nix-env --version` output**

3.0pre20201002_57b935a

**Additional context**

```console
$ git --version
git version 2.28.0
```

efx:
Note you'll want to use `nix flake metadata --json` to try any of the `nix info` commands above. (`metadata` subcommand has replaced `info`).

PR #5758, opened 2021-12-09
git fetcher: fix detection of dirty git workspaces

mschwaig:
I used the steps outlined in #4140 to verify the issue was still there (using `nix flake metadata` which replaced `nix flake info`) with my local version of Nix and to verify that my first commit here fixes it.

The above did not fix the actual issue I was having similar to #5302, but in GitLab CI. There I think the problem is the code relying on the `.git/refs/heads` directory. I fixed this as well, but the error handling for `git rev-parse`is still incomplete. That's why this PR is a draft.

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -220,11 +220,17 @@ struct GitInputScheme : InputScheme
         if (!input.getRef() && !input.getRev() && isLocal) {
             bool clean = false;
 
+            auto env = getEnv();
+            // Set LC_ALL to C: because we rely on the error messages from git rev-parse to determine what went wrong
+            // that way unknown errors can lead to a failure instead of continuing through the wrong code path
+            env["LC_ALL"] = "C";
+
             /* Check whether HEAD points to something that looks like a commit,
                since that is the refrence we want to use later on. */
             auto result = runProgram(RunOptions {
                 .program = "git",
                 .args = { "-C", actualUrl, "--git-dir=.git", "rev-parse", "--verify", "--no-revs", "HEAD^{commit}" },
+                .environment = env,
                 .mergeStderrToStdout = true
             });
             auto exitCode = WEXITSTATUS(result.first);
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -222,17 +222,25 @@ struct GitInputScheme : InputScheme
 
             /* Check whether HEAD points to something that looks like a commit,
                since that is the refrence we want to use later on. */
-            bool hasHead = false;
-            try {
-                runProgram("git", true, { "-C", actualUrl, "rev-parse", "--verify", "--no-revs", "HEAD^{commit}" });
-                hasHead = true;
-            } catch (ExecError & e) {
-                // git exits with status 128 here if it does not detect a repository.
-                if (!WIFEXITED(e.status) || WEXITSTATUS(e.status) != 128) {
-                    throw Error("Git tree '%s' is broken.\n'git rev-parse --verify HEAD' failed with exit code %d.", actualUrl, WEXITSTATUS(e.status));
-                }
+            auto result = runProgram(RunOptions {
+                .program = "git",
+                .args = { "-C", actualUrl, "--git-dir=.git", "rev-parse", "--verify", "--no-revs", "HEAD^{commit}" },
+                .mergeStderrToStdout = true
+            });
+            auto exitCode = WEXITSTATUS(result.first);
+            auto errorMessage = result.second;
+
+            if (errorMessage.find("fatal: not a git repository") != std::string::npos) {
+                throw Error("'%s' is not a git repository.", actualUrl);
+            } else if (errorMessage.find("fatal: Needed a single revision") != std::string::npos) {
+                // indicates that the repo does not have any commits
+                // we want to proceed and will consider it dirty later
+            } else if (exitCode != 0) {
+                // any other errors should lead to a failure
+                throw Error("Getting the HEAD of the git tree '%s' failed with exit code %d:\n%s", actualUrl, exitCode, errorMessage);
             }
 
+            bool hasHead = exitCode == 0;
             try {
                 if (hasHead) {
                     // Using git diff is preferrable over lower-level operations here,
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -220,21 +220,21 @@ struct GitInputScheme : InputScheme
         if (!input.getRef() && !input.getRev() && isLocal) {
             bool clean = false;
 
-            /* Check whether this repo has any commits. There are
-               probably better ways to do this. */
-            auto gitDir = actualUrl + "/.git";
-            auto commonGitDir = chomp(runProgram(
-                "git",
-                true,
-                { "-C", actualUrl, "rev-parse", "--git-common-dir" }
-            ));
-            if (commonGitDir != ".git")
-                    gitDir = commonGitDir;
-
-            bool haveCommits = !readDirectory(gitDir + "/refs/heads").empty();
+            /* Check whether HEAD points to something that looks like a commit,
+               since that is the refrence we want to use later on. */
+            bool hasHead = false;
+            try {
+                runProgram("git", true, { "-C", actualUrl, "rev-parse", "--verify", "--no-revs", "HEAD^{commit}" });
+                hasHead = true;
+            } catch (ExecError & e) {
+                // git exits with status 128 here if it does not detect a repository.
+                if (!WIFEXITED(e.status) || WEXITSTATUS(e.status) != 128) {
+                    throw Error("Git tree '%s' is broken.\n'git rev-parse --verify HEAD' failed with exit code %d.", actualUrl, WEXITSTATUS(e.status));
+                }
+            }
 
             try {
-                if (haveCommits) {
+                if (hasHead) {
                     runProgram("git", true, { "-C", actualUrl, "diff-index", "--quiet", "HEAD", "--" });
                     clean = true;
                 }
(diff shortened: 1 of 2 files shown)


mschwaig:
I found a good source to consider for this PR:
https://github.com/git/git/blob/ad1260b6c994f7c0f9c259bd39f39979f7f4ecc2/git-sh-setup.sh

It has a `require_clean_work_tree` function that is meant as a building-block for automated scripts.

regnat (commented):
> but the error handling for git rev-parseis still incomplete

Is it in less complete than what was already there? It seems to me that this PR is handling the “not a git repository” case, which is already as much as what the other code was doing.

Do you mind also adding a test (well two) for that?

mschwaig:
> Is it in less complete than what was already there? It seems to me that this PR is handling the “not a git repository” case, which is already as much as what the other code was doing.

I did not have a good feeling about just throwing the error there, so I added a specific error message (daab386e9575705988dedbfe1548586da49ce7ae). The idea is that if you point nix at a broken repo it should tell you to fix it.

Next I will look into the commands and arguments called in that script I posted above, since I only found that after I had implemented my fixes and it seems like a good official source for how to do this right.

> Do you mind also adding a test (well two) for that?

I don't mind adding tests but I have to look into how to do that here and how to best test these changes. If you have suggestions for specific test cases in mind feel free to post them.

mschwaig:
I came to the conclusion that using `git diff` works better for our use case than using multiple lower-level commands like the previous version of my fix or [git-sh-setup.sh](https://github.com/git/git/blob/ad1260b6c994f7c0f9c259bd39f39979f7f4ecc2/git-sh-setup.sh) do. I also added tests for the two specific fixed issues.

src/libfetchers/git.cc
```diff
@@ -223,22 +223,32 @@ struct GitInputScheme : InputScheme
...
+            /* Check whether HEAD points to something that looks like a commit,
+               since that is the refrence we want to use later on. */
+            bool hasHead = false;
+            try {
+                runProgram("git", true, { "-C", actualUrl, "rev-parse", "--verify", "--no-revs", "HEAD^{commit}" });
+                hasHead = true;
+            } catch (ExecError & e) {
+                // git exits with status 128 here if it does not detect a repository.
+                if (!WIFEXITED(e.status) || WEXITSTATUS(e.status) != 128) {
+                    throw Error("Git tree '%s' is broken.\n'git rev-parse --verify HEAD' failed with exit code %d.", actualUrl, WEXITSTATUS(e.status));
```
regnat, line 235: Maybe use the `addTrace` mechanism so that the original exception is shown, just with more infos

```suggestion
                    e.addTrace({}, hintfmt("while trying to get the HEAD of the git tree '%s'", actualUrl);
                    throw;
```

regnat (changes requested):
Thanks, looks great now!

Just a small suggestion, but :+1: overall

mschwaig:
There were still a few issues with this PR before bc66c8316648a30b1e08e93d01a09f52333adcf4 with that git `rev-parse command`.

* It was writing some extra output when it failed.
* It was written to proceed when it does not detect a repo at all (which would still eventually lead to an error, but later).
* It also proceeded on repos without any commits, because not detecting a repo and detecting a repo without commits share the same status code. This was not done intentionally by me but seems to be expected behavior as the flake tests use lots of repos without commits.

Those cases are now distinguished by the error message `git rev-parse` returns. I added tests for a few of them. Doing this by status code is sadly not possible when repos without commits are allowed, at least not without adding more git commands.

I am not fully convinced that allowing repos without commits is a good feature. It looks like flakes kind of rely on this by `nix` moving `flake.nix` and `flake.lock` from the untracked state to the unstaged state, so that they are picked up by `git ls-files` and therefore part of a dirty tree right away without the user ever committing them even when there is no commit in the repo at all. I'm assuming this is done for convenience. I find it a bit magical. I'm not sure how much of this is intentionally part of the tests.

```
mschwaig@mutalisk ~/t/testAnotherFlakeThing6> git init
Initialized empty Git repository in /home/mschwaig/tmp/testAnotherFlakeThing6/.git/
mschwaig@mutalisk ~/t/testAnotherFlakeThing6 (main)> nix flake init
mschwaig@mutalisk ~/t/testAnotherFlakeThing6 (main)> git status
On branch main

No commits yet

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	new file:   flake.nix

no changes added to commit (use "git add" and/or "git commit -a")
mschwaig@mutalisk ~/t/testAnotherFlakeThing6 (main)> git ls-files
flake.nix
```

johnrichardrinehart:
Forgive my interjection: I'm affected by this PR via #5302. I see no problem with allowing fresh repos (no commits) to use flakes. In fact, I think that's a very valid use case for development purposes prior to an initial commit. The alternative is to `commit && echo "fix some stuff" && commit --amend` until everything is good - that UX is worse than the magical one (I agree it's kind of magical but it seems like it's origin is [here](https://github.com/NixOS/nix/blob/fcddaa4b9b22a910a24a513bc05579a12eb1db41/src/nix/flake.cc#L725-L729)).

Generally, though, in the case where someone is both:
1) starting a new repo
2) developing a flake

I think it's totally reasonable to expect that `nix` would need that `flake.nix` in the working tree and that, if it's not present, it will add it itself.

src/libfetchers/git.cc
```diff
@@ -225,17 +225,25 @@ struct GitInputScheme : InputScheme
...
+                .program = "git",
+                .args = { "-C", actualUrl, "--git-dir=.git", "rev-parse", "--verify", "--no-revs", "HEAD^{commit}" },
+                .mergeStderrToStdout = true
+            });
+            auto exitCode = WEXITSTATUS(result.first);
+            auto errorMessage = result.second;
+
+            if (errorMessage.find("fatal: not a git repository") != std::string::npos) {
+                throw Error("'%s' is not a git repository.", actualUrl);
+            } else if (errorMessage.find("fatal: Needed a single revision") != std::string::npos) {
```
regnat, line 238: That’s not gonna work if the locale isn’t english because the error message will be translated:

```console
$ git init
$ nix flake init                                                 
$ LANG=C nix run github:mschwaig/nix/fix-git-workspace-dirty-detection -- build # Works with a C (or en) locale
warning: Git tree '/tmp/tmp.hjV72WecGo' is dirty
warning: creating lock file '/tmp/tmp.hjV72WecGo/flake.lock'
warning: Git tree '/tmp/tmp.hjV72WecGo' is dirty
$ LANG=fr_FR nix run github:mschwaig/nix/fix-git-workspace-dirty-detection -- build # But fails with another
error: Getting the HEAD of the git tree '/tmp/tmp.hjV72WecGo' failed with exit code 128:
       fatal: Une seule rision attendue
       
```

Maybe you could run another command here (`git rev-parse --all` ? `git rev-parse --git-dir` ? might be a better one) to distinguish between the no-repo and repo-but-no-commit cases. Since it’s an edge-case the cost of an extra command shouldn’t be an issue

regnat (commented):
Thanks for the update, looking pretty great now.

Unfortunately the detection of a directory with no commit is still broken for non-english locales :/

(And to answer your question: I think it’s a useful thing to have it. I for one frequently create a git repository just to make it easy to filter out which files I want to have included in the flake by adding them or not to the index. Which is very handy and doesn’t require to create any commit)

src/libfetchers/git.cc
mschwaig: I have implemented a fix for this now in 0ef6df466f46a32f8c8677009303aea7d98ea2ee.

I have opted to explicitly setting the locale via an environment variable and still parsing the error messages, instead of running another `git` command, because I do not only want to distinguish the  no-repo and repo-but-no-commit cases, but also the case where `git` fails in some other unexpected error, which  should lead to fetch failing. I think looking at the error messages works better for that.

mschwaig:
I have fixed and added a tests for that locale related failure, but I have also converted this to a draft for now.

My reasons are:
* The tests are not properly working yet (they don't fail if I undo 0ef6df466f46a32f8c8677009303aea7d98ea2ee - not sure yet what's wrong).
* I had to add a dependency to `glibcLocales` just for those tests which is a bit ugly.
* I originally wanted to reproduce the locale issue with `builtins.fetchGit`, but was unable to, so I'm using `nix falke show` now and put them with the flake tests.

Generally I'm not sure if it is worthwhile to add those tests.

mschwaig:
Since I did not get the tests for that locale-related failure to work I have removed them from this PR.

You can still find them at https://github.com/mschwaig/nix/tree/git-workspace-dirty-detection-locale-tests

mschwaig:
I just rebased this onto current master, because I needed a more recent nix with those fixes.

mschwaig:
@thufschmitt can you or someone else take another look at this?

The `Changes requested` state here is outdated and it need another look, but I don't really know how to get out of that state.

mschwaig:
> Forgive my interjection: I'm affected by this PR via #5302. I see no problem with allowing fresh repos (no commits) to use flakes. In fact, I think that's a very valid use case for development purposes prior to an initial commit. The alternative is to `commit && echo "fix some stuff" && commit --amend` until everything is good - that UX is worse than the magical one (I agree it's kind of magical but it seems like its origin is [here](https://github.com/NixOS/nix/blob/fcddaa4b9b22a910a24a513bc05579a12eb1db41/src/nix/flake.cc#L725-L729)).
> 
> Generally, though, in the case where someone is both:
> 
>     1. starting a new repo
> 
>     2. developing a flake
> 
> 
> I think it's totally reasonable to expect that `nix` would need that `flake.nix` in the working tree and that, if it's not present, it will add it itself.
> 
> My apologies if I'm misunderstanding something and/or if my input isn't wanted, here.

@johnrichardrinehart I agree. I came to the same conclusion, so this PR does not change that behaviour. Thanks for the input.

thufschmitt (commented):
Thanks for the reminder @mschwaig .

This is all a bit dirty − esp the reliance of the error message from Git − but I guess we don’t have much choice (well, except maybe switching to libgit2, but that would be a much wider scope thing to do).

So :+1: from me (provided the CI agrees). Just leaving it open a couple extra days in case @edolstra wants to have another look

thufschmitt (approved):

Issue #5687, opened 2021-11-29
aarch64-darwin build fails see linked pr for more info


Artturin:
```
10f9a8e77d504354deafbcf7351a6d46d4528542 is the first bad commit
commit 10f9a8e77d504354deafbcf7351a6d46d4528542
Author: Eelco Dolstra <edolstra@gmail.com>
Date:   Fri Oct 15 16:52:37 2021 +0200

    Add a test for the non-blocking GC

 src/libstore/gc.cc       |  4 ++++
 tests/gc-non-blocking.sh | 33 +++++++++++++++++++++++++++++++++
 tests/local.mk           |  1 +
 3 files changed, 38 insertions(+)
 create mode 100644 tests/gc-non-blocking.sh
```

* https://github.com/NixOS/nix/pull/5149

cc @edolstra

_Originally posted by @bobrik in https://github.com/NixOS/nixpkgs/issues/147376#issuecomment-981323782_

mopleen:
Probably related https://github.com/NixOS/nix/issues/5502

DieracDelta:
Another datapoint: this is also failing to build for me on aarch64-darwin.

abathur:
I tried to nixpkgs-review this PR on intel Sunday night and got:

```console
ran test tests/referrers.sh... [PASS]
ran test tests/gc-non-blocking.sh... [FAIL]
    ++ [[ -n '' ]]
    + needLocalStore 'the GC test needs a synchronisation point'
    + [[ '' == \d\e\m\o
     ]]
    + clearStore
    + echo 'clearing store...'
    clearing store...
    + chmod -R +w /private/tmp/nix-build-nix-2.5pre20211126_55275fc.drv-0/nix-test/tests/gc-non-blocking/store
    + rm -rf /private/tmp/nix-build-nix-2.5pre20211126_55275fc.drv-0/nix-test/tests/gc-non-blocking/store
    + mkdir /private/tmp/nix-build-nix-2.5pre20211126_55275fc.drv-0/nix-test/tests/gc-non-blocking/store
    + rm -rf /private/tmp/nix-build-nix-2.5pre20211126_55275fc.drv-0/nix-test/tests/gc-non-blocking/var/nix
    + mkdir /private/tmp/nix-build-nix-2.5pre20211126_55275fc.drv-0/nix-test/tests/gc-non-blocking/var/nix
    + clearProfiles
[...]

PR #5770, opened 2021-12-13
Fix macOS GC failures

### LINKED ISSUE #5717: Investigate non-blocking GC random failures

edolstra:
Fixes #5717.

### DIFF
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -535,6 +535,12 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
                         }
                     });
 
+                    /* On macOS, accepted sockets inherit the
+                       non-blocking flag from the server socket, so
+                       explicitly make it blocking. */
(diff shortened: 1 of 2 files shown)


Artturin:
Did this fix https://github.com/NixOS/nix/issues/5687 ?

abathur:
xref above from 3605 is on an error message that contains an instance of debug added in d17d46c

src/libstore/gc.cc
```diff
@@ -126,7 +126,17 @@ void LocalStore::addTempRoot(const StorePath & path)
...
+            try {
+                nix::connect(state->fdRootsSocket.get(), socketPath);
+            } catch (SysError & e) {
+                /* The garbage collector may have exited, so we need to
+                   restart. */
+                if (e.errNo == ECONNREFUSED) {
+                    debug("GC socket connection refused");
+                    state->fdRootsSocket.close();
+                    goto restart;
+                }
```
andersk, line 138: This is missing a rethrow for other error codes (#6648).

PR #5788, opened 2021-12-16
backport - docker: fix image so that nix profile works

johnae:
nix profile will otherwise throw this error:

error: path '/nix/var/nix/profiles/default/manifest.nix' is not in the Nix store

That's not entirely true since manifest.nix is within a directory in
the nix store but nix profile seems to require the manifest.nix itself
to be a store path.

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -137,11 +137,8 @@ let
         name = "root-profile-env";
         paths = defaultPkgs;
       };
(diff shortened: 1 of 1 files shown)


johnae:
@garbas @edolstra This is the same fix to the docker image as earlier but for 2.4.

Issue #5794, opened 2021-12-17
Cannot make tests 


nmattia:
**Describe the bug**

I cannot run functional tests as per [the instructions](https://hydra.nixos.org/build/161545924/download/1/manual/contributing/hacking.html).

**Steps To Reproduce**

```bash
[nix-shell:~/nix]$ make tests
make: Nothing to be done for 'tests'.

[nix-shell:~/nix]$ make nix_tests
make: *** No rule to make target 'nix_tests'.  Stop.

[nix-shell:~/nix]$ make tests/lang/eval-okay-fromTOML.exp
make: Nothing to be done for 'tests/lang/eval-okay-fromTOML.exp'.

[nix-shell:~/nix]$ make tests/lang/eval-okay-fromTOML.nix
make: Nothing to be done for 'tests/lang/eval-okay-fromTOML.nix'.
```

**Additional context**

The build (with `make -j N`) went fine.

nmattia:
This works:

```
$ make tests/lang.sh.test
```

regnat:
The doc is still quite… sparse indeed.

[> nix-shell:~/nix]$ make tests
> make: Nothing to be done for 'tests'.
> 
> [nix-shell:~/nix]$ make nix_tests
> make: *** No rule to make target 'nix_tests'.  Stop.

That one is actually specified in the doc :wink: . The targets for running the tests are `check` (for the few unit tests) and `installcheck` for the functional tests

> [nix-shell:~/nix]$ make tests/lang/eval-okay-fromTOML.exp
> make: Nothing to be done for 'tests/lang/eval-okay-fromTOML.exp'.
> 
> [nix-shell:~/nix]$ make tests/lang/eval-okay-fromTOML.nix
> make: Nothing to be done for 'tests/lang/eval-okay-fromTOML.nix'.

Yes, the test targets are of the form `{stuff}.test` where stuff is an element of the `nix_tests` list defined in `tests/local.mk`. Definitely needs some more docs

tomberek:
Closing as stale. Re-open if needed.

Issue #5793, opened 2021-12-17
Cannot nix-build on macOS Monterey


nmattia:
**Describe the bug**

I cannot build the code with `nix-build`.

**Steps To Reproduce**

Clone the repo, then nix-build:

```
~/nix$ nix-build
these derivations will be built:
  /nix/store/9lg74h6qwi9sr32mhj6pcbdr2xi2rx65-nix-2.6.0pre19700101_dirty.drv
error: while setting up the build environment: getting attributes of path '/usr/lib/libSystem.B.dylib': No such file or directory
```

**Expected behavior**

The build should start.

**`nix-env --version` output**

~/nix$ nix-env --version
nix-env (Nix) 2.3.12

**Additional context**
 
Believe this is due to macOS having a dylib cache (introduced in Big Sur iirc)

virusdave:
This problem occurs on Big Sur as well, and currently prevents me from doing basically anything locally :(

toonn:
Does this happen for the stable release too?

virusdave:
Still happened in 2.3.10 for me.  2.5.x hit a different error when i tried, so it might(?) be fixed there already.  Will report back if i get anything working with that version.

virusdave:
I went ahead and fully reinstalled nix (package manager) on darwin at 2.6.0, and this no longer happens for me.  Whew.

PR #5795, opened 2021-12-17
Replace cpptoml with toml11

### LINKED ISSUE #5792: error: while parsing a TOML string at ... Arrays must be homogeneous
**Describe the bug**

The built-in `builtins.fromTOML` cannot parse heterogeneous arrays, whereas the TOML spec specifies [arrays may be mixed-type](https://toml.io/en/v1.0.0#array).

**Steps To Reproduce**

```
nix-repl> builtins.fromTOML ''foo = [ 1, "bar" ]''
error: while parsing a TOML string at (string):1:1: Arrays must be homogeneous at line 1

nix-repl> builtins.nixVersion
"2.3.12"
```

```bash
nix-env --version
nix-env (Nix) 2.3.12
```

---

EDIT: looking at the code, looks like the parsing is done by [cpptoml](https://github.com/skystrife/cpptoml). Unfortunately that library looks abandoned.

---

EDIT: This looks like a more lively, more compliant alternative: https://github.com/ToruNiina/toml11

nmattia:
Fixes #5792 

This replaces the (somewhat outdated) https://github.com/skystrife/cpptoml with the well maintained, more compliant https://github.com/ToruNiina/toml11.

This fixes issue #5792 by allowing heterogenous arrays:

```
nix-repl> :print builtins.fromTOML ''foo = [ "bar", 2, { echo = "hello" } ]''
{ foo = [ "bar" 2 { echo = "hello"; } ]; }
```

While I didn't manage to run the tests (#5794 #5793) it looks like previously accepted TOML strings are still accepted, and parsed as expected:

```
nix-repl> :print import ./tests/lang/eval-okay-fromTOML.nix
[ { clients = { data = [ [ "gamma" "delta" ] [ 1 2 ] ]; hosts = [ "alpha" "omega" ]; }; database = { connection_max = 5000; enabled = true; ports = [ 8001 8001 8002 ]; server = "192.168.1.1"; }; owner = { name = "Tom Preston-Werner"; }; servers = { alpha = { dc = "eqdc10"; ip = "10.0.0.1"; }; beta = { dc = "eqdc10"; ip = "10.0.0.2"; }; }; title = "TOML Example"; } { "1234" = "value"; "127.0.0.1" = "value"; a = { b = { c = { }; }; }; arr1 = [ 1 2 3 ]; arr2 = [ "red" "yellow" "green" ]; arr3 = [ [ 1 2 ] [ 3 4 5 ] ]; arr4 = [ "all" "strings" "are the same" "type" ]; arr5 = [ [ 1 2 ] [ "a" "b" "c" ] ]; arr7 = [ 1 2 3 ]; arr8 = [ 1 2 ]; bare-key = "value"; bare_key = "value"; bin1 = 214; bool1 = true; bool2 = false; "character encoding" = "value"; d = { e = { f = { }; }; }; dog = { "tater.man" = { type = { name = "pug"; }; }; }; flt1 = 1; flt2 = 3.1415; flt3 = -0.01; flt4 = 5e+22; flt5 = 1e+06; flt6 = -0.02; flt7 = 6.626e-34; flt8 = 9.22462e+06; fruit = [ { name = "apple"; physical = { color = "red"; shape = "round"; }; variety = [ { name = "red delicious"; } { name = "granny smith"; } ]; } { name = "banana"; variety = [ { name = "plantain"; } ]; } ]; g = { h = { i = { }; }; }; hex1 = -559038737; hex2 = -559038737; hex3 = -559038737; int1 = 99; int2 = 42; int3 = 0; int4 = -17; int5 = 1000; int6 = 5349221; int7 = 12345; j = { "ʞ" = { l = { }; }; }; key = "value"; key2 = "value"; name = "Orange"; oct1 = 342391; oct2 = 493; physical = { color = "orange"; shape = "round"; }; products = [ { name = "Hammer"; sku = 738594937; } { } { color = "gray"; name = "Nail"; sku = 284758393; } ]; "quoted \"value\"" = "value"; site = { "google.com" = true; }; str = "I'm a string. \"You can quote me\". Name\tJosé\nLocation\tSF."; table-1 = { key1 = "some string"; key2 = 123; }; table-2 = { key1 = "another string"; key2 = 456; }; x = { y = { z = { w = { animal = { type = { name = "pug"; }; }; name = { first = "Tom"; last = "Preston-Werner"; }; point = { x = 1; y = 2; }; }; }; }; }; "ʎǝʞ" = "value"; } { metadata = { "checksum aho-corasick 0.6.4 (registry+https://github.com/rust-lang/crates.io-index)" = "d6531d44de723825aa81398a6415283229725a00fa30713812ab9323faa82fc4"; "checksum ansi_term 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)" = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"; "checksum ansi_term 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)" = "23ac7c30002a5accbf7e8987d0632fa6de155b7c3d39d0067317a391e00a2ef6"; "checksum arrayvec 0.4.7 (registry+https://github.com/rust-lang/crates.io-index)" = "a1e964f9e24d588183fcb43503abda40d288c8657dfc27311516ce2f05675aef"; }; package = [ { dependencies = [ "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)" ]; name = "aho-corasick"; source = "registry+https://github.com/rust-lang/crates.io-index"; version = "0.6.4"; } { name = "ansi_term"; source = "registry+https://github.com/rust-lang/crates.io-index"; version = "0.9.0"; } { dependencies = [ "libc 0.2.42 (registry+https://github.com/rust-lang/crates.io-index)" "termion 1.5.1 (registry+https://github.com/rust-lang/crates.io-index)" "winapi 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)" ]; name = "atty"; source = "registry+https://github.com/rust-lang/crates.io-index"; version = "0.2.10"; } ]; } { a = [ [ { b = true; } ] ]; c = [ [ { d = true; } ] ]; e = [ [ 123 ] ]; } ]
```

For free, we also get nice ["rust-like"](https://github.com/ToruNiina/toml11#in-the-case-of-syntax-error) error messages:

Screenshot 2021-12-17 at 22 10 09

The license is MIT. I've vendored toml11 and only included the files relevant in the build, plus the README.

Apologies if the code is silly, I don't know C++ and I've mostly been following the compiler's error messages.

I've split this in two commits; the first commit is a huge diff replacing cpptoml with toml11; the second commit is more interesting.


src/libexpr/primops/fromTOML.cc
```diff
@@ -1,93 +1,87 @@
...
-        // containing a table.
-        else if (auto t2 = t->as_table_array()) {
-            size_t size = t2->get().size();
-
-            state.mkList(v, size);
-
-            for (size_t j = 0; j < size; ++j)
-                visit(*(v.listElems()[j] = state.allocValue()), t2->get()[j]);
-        }
+                        // TODO: note about creating children here in old code
```
nmattia, line 33: The previous code actually inline the "array" case here; it's unclear to me why

src/libexpr/primops/fromTOML.cc
```diff
@@ -1,93 +1,87 @@
...
+                mkFloat(v, toml::get<double>(t));
+                break;;
+            case toml::value_t::string:
+                mkString(v, toml::get<std::string>(t));
+                break;;
+            case toml::value_t::local_datetime:
+            case toml::value_t::offset_datetime:
+            case toml::value_t::local_date:
+            case toml::value_t::local_time:
+                // TODO: convert to string?
```
nmattia, line 64: How should we handle dates and times best? Failure is an option, but we could also convert them to strings.

src/libexpr/primops/fromTOML.cc
```diff
@@ -1,93 +1,87 @@
...
-                throw EvalError("unsupported value type in TOML");
         }
-
-        else abort();
     };
 
     try {
-        visit(v, parser(tomlStream).parse());
-    } catch (std::runtime_error & e) {
+        visit(val, toml::parse(tomlStream)); // TODO give filename
```
nmattia, line 75: Would be nice to give this a filename, right now it says "unknown file".

src/libexpr/primops/fromTOML.cc
```diff
@@ -1,93 +1,87 @@
...
+    } catch (std::exception & e) { // TODO: toml::syntax_error
```
nmattia, line 76: `toml11` has its own exception type, happy to use anything here.

nmattia:
Tests failed, here's a reproduction:

```
# on master
nix-repl> builtins.fromTOML ''hex = 0xDEADBEAF''
{ hex = 3735928495; }

# on this PR
nix-repl> builtins.fromTOML ''hex = 0xDEADBEEF''
{ hex = -559038737; }
```

nmattia:
@edolstra what do you think?

src/libexpr/primops/fromTOML.cc
edolstra: We should rethrow those exceptions as a subtype of `Error` since we don't really handle `std::exception` in a lot of places.

edolstra:
@nmattia Looks good to me, thanks!

src/libexpr/primops/fromTOML.cc
nmattia: Do you mean that `std::exception` is too general, and instead we should let those bubble up and only catch e.g. `toml::syntax_error`, which is `toml11`'s (`std::runtime_error` inheriting) exception type? If you mean something else I'm happy to implement it, I just need some pointers!

src/libexpr/primops/fromTOML.cc
nmattia: Works as is, so simpler is better.

nmattia:
Looks like another issue of hex parsing, I'll have a look:

Screenshot 2021-12-23 at 11 14 15

nmattia:
@edolstra potential fix here: https://github.com/NixOS/nix/pull/5827

balsoft:
This appears to have broken something. Reproducing: `nix build github:balsoft/simple-osd-daemons`. Error:

```
error: 'toPackageId' at /nix/store/9if9h9mmpnrvl94h2axhxv5dpw6a844j-source/tools.nix:181:23 called without required argument 'name'

       at /nix/store/9if9h9mmpnrvl94h2axhxv5dpw6a844j-source/tools.nix:257:45:

          256|             packagesWithoutLocal = builtins.filter (p: p ? source) packagesWithDuplicates;
          257|             packageById = package: { name = toPackageId package; value = package; };
             |                                             ^
          258|             packagesById = builtins.listToAttrs (builtins.map packageById packagesWithoutLocal);
```

(strangely, it looks like the `package` attrset does have a `name` attribute?)

Git bisect'ing:
```
The first bad commit could be any of:
c3f68b5db7b9d4b963102a2abc20e20445a8bec5
10a8b5d3ae65cbf842faa83a91e7805276cb799e
```

DavHau:
It is not clear which version of toml11 was vendored. It would be great if there was a version noted somewhere in the source or in a commit message.

PR #5807, opened 2021-12-20
Fix IFD with CA derivations

### LINKED ISSUE #5805: CA-derivations aren’t compatible with IFD
**Describe the bug**

Using IFD on a CA derivation (or depends on a ca derivation) fails with a cryptic error message like `error: access to absolute path '/0vfggackkifmvxhxpaapx26lswwg5xcr54dk1cmdgg90j4jw9njc' is forbidden in pure eval mode (use '--impure' to override)` or `error: getting status of '/1ix2zgscfhpnx492z7i2fr62fmipxcnw2ssjrhj0i802vliq88jv': No such file or directory`

**Steps To Reproduce**

```console
$ cat <<EOF > default.nix
let pkgs = import <nixpkgs> { config.contentAddressedByDefault = true; }; in

let
  drvToImport = pkgs.writeText "foo.nix" ''
    let pkgs = import <nixpkgs> {}; in pkgs.hello
  '';
in
import drvToImport
EOF
$ nix-build
error: getting status of '/1ix2zgscfhpnx492z7i2fr62fmipxcnw2ssjrhj0i802vliq88jv': No such file or directory
```

**Expected behavior**

Should work

**Additional context**

IFD works because `import` (and I guess `readFile`) have some magic to realise the context of their string argument before being called. But this isn’t enough with CA derivations because these strings don’t contain the output paths of the realisations they refer to, but only their placeholder (the `/<bigHash>` thing above). Hence the cryptic error.

It should be possible (and hopefully not too hard) to rewrite these strings before passing them to the primops

regnat:
Rewrite the string taken by the IFD-like primops to contain the actual output paths of the derivations rather than the placeholders

Fix #5805

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -35,9 +35,10 @@ namespace nix {
 InvalidPathError::InvalidPathError(const Path & path) :
     EvalError("path '%s' is not valid", path), path(path) {}
 
-void EvalState::realiseContext(const PathSet & context)
+StringMap EvalState::realiseContext(const PathSet & context)
 {
     std::vector<DerivedPath::Built> drvs;
+    StringMap res;
 
     for (auto & i : context) {
         auto [ctxS, outputName] = decodeContext(i);
@@ -46,10 +47,12 @@ void EvalState::realiseContext(const PathSet & context)
             throw InvalidPathError(store->printStorePath(ctx));
         if (!outputName.empty() && ctx.isDerivation()) {
             drvs.push_back({ctx, {outputName}});
+        } else {
+            res.insert_or_assign(ctxS, ctxS);
         }
     }
 
-    if (drvs.empty()) return;
+    if (drvs.empty()) return {};
 
     if (!evalSettings.enableImportFromDerivation)
         throw Error(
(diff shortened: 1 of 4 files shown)


src/libexpr/primops.cc
```diff
@@ -61,19 +64,43 @@ void EvalState::realiseContext(const PathSet & context)
...
-            }
+        for (auto & [_placeholder, outputPath] : res) {
+            allowPath(outputPath);
         }
     }
+
+    return res;
+}
+
+static Path realisePath(EvalState & state, const Pos & pos, Value & v, bool requireAbsolutePath = true)
```
edolstra, line 92: What does `requireAbsolutePath` mean? I.e. when is it okay for it to be a relative path, and what is it relative to?

regnat: `requireAbsolutePath` is here for the `findFile` primops (the desugaring of the angle brackets). Its first argument (the search path) can contain some strings with context, but these can be relative paths (which will then be relative to the cwd of the evaluator).

I’m not utterly happy having this flag just for that, but I’ve kept it to make sure to conserve the exact same semantics

(I had no idea that there could be strings-with-context here btw. Wonder whether that’s actually used anywhere)

tomberek:
I'm getting errors related to this commit:
```
error: access to absolute path '/.config/nixpkgs/overlays.nix' is forbidden in pure eval mode (use '--impure' to override)
```

The getEnv in impure.nix is resolving to the empty string, but is still trying to access the overlays.nix. It's related to imports/paths, but I'm not sure exactly what is the root cause.

PR #5808, opened 2021-12-20
Add shell completion for why-depends

### LINKED ISSUE #5806: Shell completion for `why-depends`
Some shell completion for `nix why-depends` would be nice. Making a PR in a minute but it will probably be horrible.

ncfavier:
Closes #5806.

I tried adding a `size_t` argument to `completeInstallable` and doing `.completer = completeInstallable` but I got scary errors about no known conversion from initialiser list to `nix::Args::ExpectedArg&&`. Not sure why it works with `completePath` but not `completeInstallable`.

### DIFF
--- a/src/nix/why-depends.cc
+++ b/src/nix/why-depends.cc
@@ -34,8 +34,21 @@ struct CmdWhyDepends : SourceExprCommand
 
     CmdWhyDepends()
     {
-        expectArg("package", &_package);
-        expectArg("dependency", &_dependency);
+        expectArgs({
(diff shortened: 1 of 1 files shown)


ncfavier:
Thanks!

edolstra:
Thanks! The error was probably because `completeInstallable` is a method and so requires an implicit `this` argument, which doesn't match the type of `.completer`.

ncfavier:
Ah, that makes sense.

PR #5809, opened 2021-12-20
small perf improvements

pennae:
nix allocates a *lot* of Values, and going through GC_malloc every time can be avoided by allocating a block of Values at once and using them one by one. it's not a world of change, but a nice improvement for a very small change. enabling LTO gives quite a good performance boost too.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -821,8 +821,23 @@ inline Value * EvalState::lookupVar(Env * env, const ExprVar & var, bool noEval)
 
 Value * EvalState::allocValue()
 {
+    /* We use the boehm batch allocator to speed up allocations of Values (of which there are many).
+       GC_malloc_many returns a linked list of objects of the given size, where the first word
+       of each object is also the pointer to the next object in the list. This also means that we
+       have to explicitly clear the first word of every object we take. */
+    if (!valueAllocCache) {
+        valueAllocCache = GC_malloc_many(sizeof(Value));
+        if (!valueAllocCache) throw std::bad_alloc();
+    }
+
+    /* GC_NEXT is a convenience macro for accessing the first word of an object.
+       Take the first list item, advance the list to the next item, and clear the next pointer. */
+    void * p = valueAllocCache;
+    GC_PTR_STORE_AND_DIRTY(&valueAllocCache, GC_NEXT(p));
+    GC_NEXT(p) = nullptr;
+
     nrValues++;
-    auto v = (Value *) allocBytes(sizeof(Value));
+    auto v = (Value *) p;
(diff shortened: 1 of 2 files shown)


pennae:
no clue why the test fails with LTO enabled. :confused: going to remove that then.

src/libexpr/eval.cc
```diff
@@ -821,8 +821,17 @@ inline Value * EvalState::lookupVar(Env * env, const ExprVar & var, bool noEval)
...
 Value * EvalState::allocValue()
 {
+    if (!valueAllocCache) {
+        valueAllocCache = GC_malloc_many(sizeof(Value));
+        if (!valueAllocCache) throw std::bad_alloc();
+    }
+
+    void * p = valueAllocCache;
+    GC_PTR_STORE_AND_DIRTY(&valueAllocCache, GC_NEXT(p));
+    GC_NEXT(p) = nullptr;
```
regnat, line 831: Isn’t this line equivalent to `valueAllocCache = nullptr`? I guess it’s not because that would make the whole thing quite inefficient, but I don’t see how it isn’t

regnat (changes requested):
I indeed seem to get a somewhat consistent 2% improvement (on `nix search nixpkgs hello`). So cool :)

Might be worth a comment explaining what’s going on though because it’s not (at least not to me) easy to understand what’s going on

regnat:
> no clue why the test fails with LTO enabled. confused going to remove that then.

Do you mind opening a separate (draft) PR for that? That would be a nice thing to have.

A possible explanation for the test failure is that IIRC Nix does (well, at least used to do) some discutable (not guaranteed by the standard) assumptions on the initialization order of some `static` values. Maybe lto breaks these assumptions and causes havoc somewhere because of that

src/libexpr/eval.cc
pennae: nope, because the line before is equivalent to `valueAllocCache = *valueAllocCache`. boehm malloc_many returns a list of objects in the form of a "first word is next pointer". it also explicitly says to clear the next pointer of an object taken from the list, which is what this line does.

pennae:
> Do you mind opening a separate (draft) PR for that? That would be a nice thing to have.

can do that tomorrow :)

might be worth noting there that other objects *could* also be batch-allocated (small Bindings come to mind, over 50% of Bindings allocations are for 2 attrs or less), but in our tests that didn't seem to help much at all.

src/libexpr/eval.cc
regnat: Oh yes sorry, I misread the thing. Makes sense now, thanks

regnat (approved):

PR #5812, opened 2021-12-21
improve parser performance a bit

pennae:
the current parser is not as efficient as it could be. this is an attempt at optimizing it at least a little, and we get around 4% improvement on nixos system builds out of it. pure parsing (such as loading the massive hackage registry) benefits much more, around 13%.

### DIFF
--- a/src/libexpr/eval.hh
+++ b/src/libexpr/eval.hh
@@ -181,8 +181,8 @@ public:
     Expr * parseExprFromFile(const Path & path, StaticEnv & staticEnv);
 
     /* Parse a Nix expression from the specified string. */
-    Expr * parseExprFromString(std::string_view s, const Path & basePath, StaticEnv & staticEnv);
-    Expr * parseExprFromString(std::string_view s, const Path & basePath);
+    Expr * parseExprFromString(std::string s, const Path & basePath, StaticEnv & staticEnv);
+    Expr * parseExprFromString(std::string s, const Path & basePath);
 
     Expr * parseStdin();
 
@@ -310,7 +310,7 @@ private:
     friend struct ExprAttrs;
     friend struct ExprLet;
 
-    Expr * parse(const char * text, FileOrigin origin, const Path & path,
+    Expr * parse(char * text, size_t length, FileOrigin origin, const Path & path,
         const Path & basePath, StaticEnv & staticEnv);
 
 public:
--- a/src/libexpr/parser.y
+++ b/src/libexpr/parser.y
@@ -273,9 +273,16 @@ void yyerror(YYLTYPE * loc, yyscan_t scanner, ParseData * data, const char * err
   nix::Formal * formal;
   nix::NixInt n;
   nix::NixFloat nf;
-  const char * id; // !!! -> Symbol
-  char * path;
-  char * uri;
+  // using C a struct allows us to avoid having to define the special
+  // members that using string_view here would implicitly delete.
+  struct StringToken {
+    const char * p;
+    size_t l;
+    operator std::string_view() const { return {p, l}; }
+  };
+  StringToken id; // !!! -> Symbol
+  StringToken path;
+  StringToken uri;
   std::vector<nix::AttrName> * attrNames;
   std::vector<std::pair<nix::Pos, nix::Expr *> > * string_parts;
 }
@@ -397,7 +404,7 @@ expr_select
 
 expr_simple
   : ID {
-      if (strcmp($1, "__curPos") == 0)
+      if (strncmp($1.p, "__curPos", $1.l) == 0)
           $$ = new ExprPos(CUR_POS);
       else
           $$ = new ExprVar(CUR_POS, data->symbols.create($1));
@@ -414,7 +421,7 @@ expr_simple
       $$ = new ExprConcatStrings(CUR_POS, false, $2);
   }
   | SPATH {
-      string path($1 + 1, strlen($1) - 2);
+      string path($1.p + 1, $1.l - 2);
       $$ = new ExprCall(CUR_POS,
           new ExprVar(data->symbols.create("__findFile")),
           {new ExprVar(data->symbols.create("__nixPath")),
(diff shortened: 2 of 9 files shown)


src/libexpr/eval.hh
```diff
@@ -310,7 +310,7 @@ private:
     friend struct ExprAttrs;
     friend struct ExprLet;
 
-    Expr * parse(const char * text, FileOrigin origin, const Path & path,
+    Expr * parse(char * text, size_t length, FileOrigin origin, const Path & path,
```
edolstra, line 313: It would be nicer to use a single `std::string_view` argument instead of separate `char *` and `size_t` arguments.

pennae: we chose to split them to signify that *not* any `string_view` would be acceptable as the parser will fail if the last two bytes of the input aren't 0, but we've no hard feelings either way.

src/libexpr/parser.y
```diff
@@ -674,7 +689,12 @@ Expr * EvalState::parseExprFromString(std::string_view s, const Path & basePath)
 Expr * EvalState::parseStdin()
 {
     //Activity act(*logger, lvlTalkative, format("parsing standard input"));
-    return parse(drainFD(0).data(), foStdin, "", absPath("."), staticBaseEnv);
+    auto buffer = drainFD(0);
+    // make sure drainFD has left some extra space for terminators
+    assert(buffer.capacity() > 2 && buffer.size() <= buffer.capacity());
```
edolstra, line 694: Why is this assertion necessary? The `+=` operator will grow the string if necessary. `buffer.size() <= buffer.capacity()` is an invariant guaranteed by `std::string`. `buffer.capacity() > 2` may fail if the input is shorter than 2 bytes.

pennae: to avoid copying the entire input string (which the generated parser previously did under the hood). when evaluating nixpkgs those copies add up to quite a bit, but for stdin it's admittedly not as important.

src/libexpr/parser.y
```diff
@@ -655,13 +660,23 @@ Expr * EvalState::parseExprFromFile(const Path & path)
...
 
 
 Expr * EvalState::parseExprFromString(std::string_view s, const Path & basePath, StaticEnv & staticEnv)
 {
-    return parse(s.data(), foString, "", basePath, staticEnv);
+    string buffer;
+    buffer.reserve(s.size() + 2);
+    buffer += s;
+    buffer += '\0';
+    buffer += '\0';
```
edolstra, line 678: Why do we need two null characters? Seems wasteful to make a copy of the input. If this is really needed, then instead of an `std::string_view`, it should take a `std::string &&` so we don't need to make a copy if the caller doesn't need to keep the string.

pennae: the generated parser demands them. the function nix previously used for parsing copied the input to add those two, so this is effectively just making behaviour visible that was hidden before.

src/libexpr/parser.y
pennae: also looks like we accidentally dropped some arithmetic, it really should be
```suggestion
    assert(buffer.capacity() > 2 && buffer.size() <= buffer.capacity() - 2);
```

src/libexpr/eval.hh
pennae: upon closer inspection `string_view` will not work. the modified parser needs ownership of the input string so it can modify it, and taking a `string_view` to const_cast its data feels horrible :/

src/libexpr/symbol-table.hh
```diff
@@ -70,15 +71,21 @@ public:
...
+    std::unordered_map<std::string_view, Symbol> symbols;
+    std::list<string> store;
 
 public:
     Symbol create(std::string_view s)
     {
-        // FIXME: avoid allocation if 's' already exists in the symbol table.
-        std::pair<Symbols::iterator, bool> res = symbols.emplace(std::string(s));
-        return Symbol(&*res.first);
+        // Most symbols are looked up more than once, so we trade off insertion performance
```
edolstra, line 80: It appears that in C++20, searching using an `std::string_view` in a set of `std::string`s now works. At least, this compiles:
```
-        // FIXME: avoid allocation if 's' already exists in the symbol table.
-        std::pair<Symbols::iterator, bool> res = symbols.emplace(std::string(s));
+        std::pair<Symbols::iterator, bool> res = symbols.emplace(s);
```

So that avoids storing a redundant `std::string_view`. But it should be checked that this doesn't introduce a lot of implicit conversions in every comparison (https://stackoverflow.com/a/53530846).

pennae: would've used those features if nix were C++20, but we didn't want to change it from 17 to 20 just for this. if that's a reasonable thing to do here we'll happily do it!

pennae:
@edolstra @thufschmitt may we ask this PR be checked? :) pretty much everything else we've not opened a PR for yet depends on this in one way or other.

src/libexpr/parser.y
```diff
@@ -273,9 +273,14 @@ void yyerror(YYLTYPE * loc, yyscan_t scanner, ParseData * data, const char * err
   nix::Formal * formal;
   nix::NixInt n;
   nix::NixFloat nf;
-  const char * id; // !!! -> Symbol
-  char * path;
-  char * uri;
+  struct StringToken {
```
thufschmitt, line 276: Might be worth explaining why we can’t use a `string_view` here

thufschmitt (approved):
Just one small comment, but LGTM otherwise

(note that I’m not really familiar with the parser code, so I might have missed something)

src/libexpr/parser.y
pennae: because the version of bison we use can't deal with it. we're planning on updating bison the 3.8 to get access to a proper C++ glr parser that *can* deal with that, but we're not entirely sure yet how to do that. currently it requires pulling *only* bison from a newer nixpkgs version than the flake uses because updating to a newer nixpkgs still breaks static build :slightly_frowning_face:

edolstra:
I still don't understand why the null terminators are necessary. That seems very hacky and infects other parts of the code (like `drainFD`).

pennae:
the generated parser demands them, no idea why (though we'd guess it's to do with lookahead in the absence of a separate tokenizer). we can either add an extra terminator ourselves or have the generated parser do it at the cost of an extra alloc+copy. and while we do agree that it's kinda terrible and it does infect anything it touches, adding that extra byte feels well worth it for avoiding potentially hundreds of megabytes of string copies that only happen to add that one extra byte :slightly_frowning_face:

edolstra:
Thanks for the explanation!

PR #5813, opened 2021-12-21
enable LTO in optimized builds

pennae:
gives 2-5% performance improvement across a board of tests.
unfortunately the flake checks fail for as yet unknown reasons 🙁


regnat:
I digged into the test failures, and I think I found the culprit:

Nix has several `Store` implementations, which are registered at startup by the constructor of `RegisterStoreImplementation`.
Each store implementation is associated with one (or several) uri schemes (`unix` for the local daemon, `ssh` for the legacy ssh store, `http(s)` for the http binary cache store, `file` for the local binary cache store, etc..) which are used at runtime to know which store implementation to use (from the value of `NIX_REMOTE` or the `--store` option).

Because spawning an http server for testing is (slightly) painful, there’s a magic environment variable that can be set to make the http cache also register the `file` uri scheme (using `libcurl` with `file://` urls behind the scenes). This environment also causes the local binary cache store not to register the `file` uri scheme to prevent any conflict.

The issue is that the http cache and the local cache don’t agree on the name of the variable: It’s called `_NIX_FORCE_HTTP` in the [http implementation](https://github.com/NixOS/nix/blob/dc83298449d5547926febe24c7bf973341127f60/src/libstore/http-binary-cache-store.cc#L74) and in the testsuite, but `_NIX_FORCE_HTTP_BINARY_CACHE_STORE` in the [local implementation](https://github.com/NixOS/nix/blob/dc83298449d5547926febe24c7bf973341127f60/src/libstore/local-binary-cache-store.cc#L109). So setting the `_NIX_FORCE_HTTP` environment variable causes both of them to register the same uri scheme, meaning that which one is chosen depends on the order in which they are registered (which itself depends on the order in which things are linked, and is very likely modified by lto). And because they behave slightly differently in some cases (the http implementation in particular has an intermediate local cache), this causes the tests to fail.

I’ve just pushed a commit making the local binary cache store honor `_NIX_FORCE_HTTP`, and it seems to fix the issue (at least it did locally)

pennae:
hope that's all it needed, having LTO would be so nice :crossed_fingers:

Kha:
Do you have any numbers on (full/incremental) build time overhead?

pennae:
just did some tests on compile time with gcc:

 - no lto: 66s (21min user time)
 - `-flto`: 2m52 (16min user)
 - `-flto=jobserver`: 58s (19min user)

lto build segfaults when using clang :(

pennae:
added a check to enable LTO only when using GCC for the moment, using clang+LTO crashes in the llvm linker plugin.

pennae:
on our current patch stack LTO actualy *decreases* performance significantly. :(

pennae:
try as we might, can't reproduce the slowdown any more. must've been a dodgy test setup, we're now seeing the 2-5% improvements mentioned before again.

edolstra:
@pennae What benchmark did you use? I tried `nix eval --json` on my NixOS system configuration and didn't get a statistically signifcant difference:
```
system CPU time:    median =      0.4195  mean =      0.4185  stddev =      0.0181  min =      0.3799  max =      0.4449  [rejected, p=0.00004, Δ=0.02573±0.01633]
user CPU time:      median =      2.6279  mean =      2.6322  stddev =      0.0461  min =      2.5441  max =      2.7097  [not rejected?, p=0.62874, Δ=-0.00515±0.03217]
elapsed time:       median =      3.4666  mean =      3.4743  stddev =      0.0702  min =      3.3715  max =      3.5834  [not rejected?, p=0.04096, Δ=0.03410±0.04862]
```

pennae:
@edolstra these two:

- `nix eval -f ../nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix`
- `nix eval --raw --impure --expr 'with import <nixpkgs/nixos> {}; system'`

parsing gained about 5%, system eval gained about 2%. the system config includes a large number of derivations in `systemPackages`.

pennae:
reran a set of benchmarks because this branch wasn't tested against zipAttrsWith before:

master (8f214a377):
```
Benchmark 1: nix search --no-eval-cache --offline nixpkgs hello
  Time (mean ± σ):      8.824 s ±  0.039 s    [User: 7.789 s, System: 0.680 s]
  Range (min … max):    8.749 s …  8.894 s    20 runs
 
Benchmark 2: nix eval -f ../nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix
  Time (mean ± σ):     416.0 ms ±   2.6 ms    [User: 380.2 ms, System: 35.8 ms]
  Range (min … max):   412.2 ms … 419.2 ms    20 runs
 
Benchmark 3: nix eval --raw --impure --expr 'with import <nixpkgs/nixos> {}; system'
  Time (mean ± σ):      2.824 s ±  0.034 s    [User: 2.561 s, System: 0.191 s]
  Range (min … max):    2.754 s …  2.878 s    20 runs
```

8f214a377 + lto:
```
Benchmark 1: nix search --no-eval-cache --offline nixpkgs hello
  Time (mean ± σ):      8.644 s ±  0.056 s    [User: 7.607 s, System: 0.690 s]
  Range (min … max):    8.504 s …  8.761 s    20 runs
 
Benchmark 2: nix eval -f ../nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix
  Time (mean ± σ):     391.0 ms ±   1.8 ms    [User: 353.4 ms, System: 37.5 ms]
  Range (min … max):   387.4 ms … 393.9 ms    20 runs
 
Benchmark 3: nix eval --raw --impure --expr 'with import <nixpkgs/nixos> {}; system'
  Time (mean ± σ):      2.755 s ±  0.027 s    [User: 2.489 s, System: 0.192 s]
  Range (min … max):    2.714 s …  2.794 s    20 runs
```

edolstra:
A downside of LTO is that it makes hacking on Nix a lot slower. For instance, without LTO, doing `touch src/libexpr/nixexpr.cc; make install -j16` takes 6.3s on my machine, but with LTO, it takes 86.0s.

edolstra:
Yeah that sounds good.

pennae:
turned the default-on into an `--enable-lto` switch to configure. kept the check that CXX is GCC when LTO is enabled though, since using LTO with clang crashes in the LLVM linker plugin for reasons we don't understand :(

lovesegfault:
@pennae Can you share how you benchmark this? 

There are some other compiler flags we could play with to see if we get perf improvements, namely:

* Graphite specific optimizations and other optimizations that are disabled at O3 but don't influence the compiler's judgement.
  * `-fgraphite-identity`
  *  `-floop-nest-optimize`
* `-fipa-pta`
  * This option increases compile times, but can potentially produce better binaries, especially with LTO. Essentially, it allows the compiler to look into called function bodies when performing alias analysis
* `-fdevirtualize-at-ltrans`
  * This allows GCC to perform devirtualization across object file boundaries using LTO. 
* `-fno-plt`
  * This option omits the PLT from the executable, making calls go through the GOT directly. It inhibits lazy binding, so this is not enabled by default. If you use prelink, this is strictly better than lazy binding. 
* Safe subset of `-ffast-math`, https://gcc.gnu.org/ml/gcc/2017-09/msg00079.html
  * `-fno-math-errno`
  * `-fno-trapping-math`

pennae:
we run hyperfine with the commands indicated above (https://github.com/NixOS/nix/pull/5813#issuecomment-1005589189). if the results are very noisy we pin hyperfine to one core with taskset and nix to a couple others. no dark magic happening here :)

roberth:
A couple percent change in performance is easily explained by compiler- or linker-induced entropy in memory layout per build configuration. See the linked resources in https://github.com/NixOS/nix/issues/5869.

It seems that micro-benchmarks can vary even more wildly, by tens of %. We're not microbenchmarking, but a couple of such arbitrary effects in hot code can presumably still average out to a noticable percentage.

pennae:
rebased to current master. we're still seeing 2-5% improvements across all our benchmarks :)

pennae:
we've resolved merge conflicts in the web UI so far for convencience, if that's not desirable for merging into master we can rebase everything properly.

thufschmitt (approved):
Thanks for keeping up with this :)

I think this is good to go, and very low risk since it's not enabled by default anyways.

I'd love to get some benchmarks see whether that can be enabled by default for the (gcc) nix build

mk/libraries.mk
```diff
@@ -91,7 +91,7 @@ define build-library
     $(1)_PATH := $$(_d)/$$($(1)_NAME).$(SO_EXT)
 
     $$($(1)_PATH): $$($(1)_OBJS) $$(_libs) | $$(_d)/
-	$$(trace-ld) $(CXX) -o $$(abspath $$@) -shared $$(LDFLAGS) $$(GLOBAL_LDFLAGS) $$($(1)_OBJS) $$($(1)_LDFLAGS) $$($(1)_LDFLAGS_PROPAGATED) $$(foreach lib, $$($(1)_LIBS), $$($$(lib)_LDFLAGS_USE)) $$($(1)_LDFLAGS_UNINSTALLED)
+	+$$(trace-ld) $(CXX) -o $$(abspath $$@) -shared $$(LDFLAGS) $$(GLOBAL_LDFLAGS) $$($(1)_OBJS) $$($(1)_LDFLAGS) $$($(1)_LDFLAGS_PROPAGATED) $$(foreach lib, $$($(1)_LIBS), $$($$(lib)_LDFLAGS_USE)) $$($(1)_LDFLAGS_UNINSTALLED)
```
Et7f3, line 94: Any reason why the + was added ? it make `make -n` ineffective. I don't see it linked with LTO

SuperSandro2000: Could it be that the jobserver is not working inside of nix build? I have seen a lot of messages in the logs around that lately.

SuperSandro2000: That patch does not apply to gcc12 and for gcc11 the jobserver is still written in c.


I switched to gcc12 and still get a lot of:


```
nix> lto-wrapper: warning: jobserver is not available: '--jobserver-auth=' is not present in 'MAKEFLAGS'
nix> lto-wrapper: note: see the '-flto' option documentation for more information
nix> lto-wrapper: warning: jobserver is not available: '--jobserver-auth=' is not present in 'MAKEFLAGS'
nix> lto-wrapper: note: see the '-flto' option documentation for more information
```

SuperSandro2000: If it is as easy as that, we should try that.

SuperSandro2000: Following overlay worked for applying the patch:

```nix
  nixVersions = prev.nixVersions // {
    stable = prev.nixVersions.stable.override {
      # stdenv = gcc12Stdenv;
      stdenv = overrideCC gcc12Stdenv (gcc.override {
        cc = gcc12.cc.overrideAttrs (oldAttrs: {
          patches = oldAttrs.patches or [ ] ++ [
            # dependency of patch below
            (fetchpatch {
              url = "https://github.com/gcc-mirror/gcc/commit/1270ccda70ca09f7d4fe76b5156dca8992bd77a6.patch";
              sha256 = "sha256-XkzktcHOuYNejFGjQYkmM+QU7/j3EHM8mUOLuczRSRY=";
            })
            # https://github.com/NixOS/nix/pull/5813#discussion_r1101203213
            # https://gcc.gnu.org/pipermail/gcc-patches/2022-August/599313.html
            (fetchpatch {
              url = "https://github.com/gcc-mirror/gcc/commit/53e3b2bf16a486c15c20991c6095f7be09012b55.patch";
              sha256 = "sha256-5AfTrsayft/bzSNsaTjqR7hB7tDq3UI4PufxQAjKG6c=";
            })
          ];
        });
      });
    }
  };
```

PR #5820, opened 2021-12-22
Don't insert spaces when completing attribute paths

ncfavier:
Adds a new completion mode `attrs` (and renames `no-filenames` to `normal`) for completing attribute paths, so that the shell doesn't add spaces.

Unfortunately I was only able to implement this for bash, even after reading the documentations for the zsh and fish completion systems. For zsh, there has to be a way since https://github.com/spwhitt/nix-zsh-completions does this (cc @hedning @matthewbauer). For fish, I have no idea (cc @haslersn @jtojnar).

### DIFF
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -39,7 +39,7 @@ void Completions::add(std::string completion, std::string description)
 bool Completion::operator<(const Completion & other) const
 { return completion < other.completion || (completion == other.completion && description < other.description); }
 
-bool pathCompletions = false;
+CompletionType completionType = ctNormal;
 std::shared_ptr<Completions> completions;
 
 std::string completionMarker = "___COMPLETE___";
@@ -277,7 +277,7 @@ Args::Flag Args::Flag::mkHashTypeOptFlag(std::string && longName, std::optional<
 
 static void _completePath(std::string_view prefix, bool onlyDirs)
 {
-    pathCompletions = true;
+    completionType = ctFilenames;
     glob_t globbuf;
     int flags = GLOB_NOESCAPE | GLOB_TILDE;
     #ifdef GLOB_ONLYDIR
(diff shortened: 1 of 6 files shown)


src/libutil/args.cc
```diff
@@ -39,7 +39,7 @@ void Completions::add(std::string completion, std::string description)
 bool Completion::operator<(const Completion & other) const
 { return completion < other.completion || (completion == other.completion && description < other.description); }
 
-bool pathCompletions = false;
+std::string completionType = "normal";
```
edolstra, line 42: Maybe it's better to change this to an enum.

haslersn:
I don't use fish anymore, so I won't be able to help with fish.

src/libutil/args.cc
ncfavier: Done

ncfavier:
Squeezed in an unrelated trivial fix.

jtojnar:
I am not sure if nospace is a good fit. In most cases, you just want a top-level attribute. Two attributes are still somewhat common and there are some three attribute paths but more is quite rare.

Maybe we could detect attribute sets containing `recurseForDerivations` and complete those with period in addition to the normal (e.g. `python3` and `python3.`). I asked fish for a feature that could make this more convenient: https://github.com/fish-shell/fish-shell/issues/8578

jtojnar:
According to https://github.com/fish-shell/fish-shell/issues/7832, fish should not insert a space after period so including a variant with a period into completion for attributes containing `recurseForDerivations` should fix it.

ncfavier:
`recurseForDerivations` isn't enough: my main use case for this is doing things like `nix build .#nixosConfigurations.foo.config.system.build.toplevel`. A better heuristic would be to add a dot *unless* the attribute is a derivation.

ncfavier:
Then again this probably depends on the actual command we're completing for: `nix eval` should probably have different semantics than `nix build`

jtojnar:
Also `python3` and `python3.pkgs.foo`.

PR #5821, opened 2021-12-22
Remove CPU locking

### LINKED ISSUE #3345: nix-shell inherits single CPU affinity when derivation has import from derivation

edolstra:
This was already accidentally disabled in ba87b08. It also no longer appears to be beneficial, and in fact slow things down, e.g. when evaluating a NixOS system configuration:
```
elapsed time:       median =      3.8170  mean =      3.8202  stddev =      0.0195  min =      3.7894  max =      3.8600  [rejected, p=0.00000, Δ=0.36929±0.02513]
```

Fixes #3345. Closes #3346.


layus:
LGTM. Nothing that can go wrong without being catched by the compiler. I made sure to look closely at the daemon protocol part. That one may go wrong.

Issue #5693, opened 2021-11-30
allow empty string interpolations 


milahu:
i use string interpolations to add "noop" comments to strings
noop = dont change the derivation

```nix
"a${/* comment in string */}b" # syntax error, unexpected '}'
''a${/* comment in string */""}b'' # ok
```

ideally both should work

Radvendii:
I'm going to close this since it's been stale for 2 years and it's not clear this is a good idea in the first place (see #5822). Feel free to reopen if you feel strongly and want to continue the discussion.

PR #5822, opened 2021-12-22
allow empty antiquotations

Radvendii:
`a${}b` or `a${/*comment*/}b`

Addresses #5693 

I don't personally have much use for this, but I've worked with that part of the code before so I figured I'd implement it. As in, I'm not weighing in on whether this PR should be accepted or not. It seems pretty harmless and like the behaviour I would expect, but the current behaviour is also perfectly consistent.

### DIFF
--- a/src/libexpr/parser.y
+++ b/src/libexpr/parser.y
@@ -288,7 +288,7 @@ void yyerror(YYLTYPE * loc, yyscan_t scanner, ParseData * data, const char * err
 %type <formal> formal
 %type <attrNames> attrs attrpath
 %type <string_parts> string_parts_interpolated ind_string_parts
-%type <e> path_start string_parts string_attr
+%type <e> path_start string_parts string_attr antiquotation
 %type <id> attr
 %token <id> ID ATTRPATH
 %token <e> STR IND_STR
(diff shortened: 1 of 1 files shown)


grahamc:
I’m not sure this is a good idea. Since nix is an expression oriented
language, shouldn’t ${}  contain an expression? This seems to add
surprising behavior. 

On December 22, 2021, GitHub ***@***.***> wrote:
> "a${}b" or "a${/comment/}b"
>
> Addresses #5693 <https://github.com/NixOS/nix/issues/5693>
>
> I don't personally have much use for this, but I've worked with that
> part of the code before so I figured I'd implement it. As in, I'm not
> weighing in on whether this PR should be accepted or not. It seems
> pretty harmless and like the behaviour I would expect, but the current
> behaviour is also perfectly consistent.
>

> You can view, comment on, or merge this pull request online at: 
> <https://github.com/NixOS/nix/pull/5822>
>
> Commit Summary
>  *  bbbc216
>     <https://github.com/NixOS/nix/pull/5822/commits/bbbc21680f2e1764ea0431c5ed9ebc2e7fc50ac3>
>     allow empty antiquotations
>

File Changes

Radvendii:
Alright, closing since I don't feel strongly about this and @grahamc does (and nobody else has responded)

PR #5824, opened 2021-12-22
nix repl: load flakes from cli args

nrdxp:
I think when working with flakes it is quite common to load the flake into a repl for inspecting and debugging. Therefore, when experimental feature "flakes" is enabled, it makes sense to consider the arguments as flake refs instead of files.
As a small bonus, I've implemented the ability for the `:load-flake` command to parse flake ref fragments (i.e. the part after the `#`).

I still need to modify the documentation, and I'm debating on whether I should add new `--files, -f` flag for when the old behavior is desired while using flakes, so I'm opening as draft for feedback. It would also be nice to include the same search behavior as other flake ref parsing commands, e.g. `nixpkgs#zlib` becomes `nixpkgs#legacyPackages.$current_system.zlib`.

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -611,11 +611,11 @@ void NixRepl::loadFile(const Path & path)
 
 void NixRepl::loadFlake(const std::string & flakeRefS)
 {
-    auto flakeRef = parseFlakeRef(flakeRefS, absPath("."), true);
+    auto [flakeRef, fragment] = parseFlakeRefWithFragment(flakeRefS, absPath("."), true);
     if (evalSettings.pureEval && !flakeRef.input.isImmutable())
         throw Error("cannot use ':load-flake' on mutable flake reference '%s' (use --impure to override)", flakeRefS);
 
-    Value v;
+    auto v = state->allocValue();
 
     flake::callFlake(*state,
         flake::lockFlake(*state, flakeRef,
@@ -624,8 +624,17 @@ void NixRepl::loadFlake(const std::string & flakeRefS)
                 .useRegistries = !evalSettings.pureEval,
                 .allowMutable  = !evalSettings.pureEval,
             }),
-        v);
-    addAttrsToScope(v);
+        *v);
+
+    auto f = v->attrs->get(state->symbols.create(fragment));
+
+    if (f == 0) {
+        warn("no attribute %s, nothing loaded", fragment);
+        return;
+    };
+
+    fragment != "" ? addAttrsToScope(*f->value) : addAttrsToScope(*v);
+
 }
 
 
@@ -654,7 +663,10 @@ void NixRepl::reloadFiles()
         if (!first) notice("");
         first = false;
         notice("Loading '%1%'...", i);
-        loadFile(i);
+
+        settings.isExperimentalFeatureEnabled(Xp::Flakes)
+        ? loadFlake(i)
+        : loadFile(i);
     }
 }
 


src/nix/repl.cc
```diff
@@ -611,11 +611,11 @@ void NixRepl::loadFile(const Path & path)
 
 void NixRepl::loadFlake(const std::string & flakeRefS)
 {
-    auto flakeRef = parseFlakeRef(flakeRefS, absPath("."), true);
+    auto [flakeRef, fragment] = parseFlakeRefWithFragment(flakeRefS, absPath("."), true);
     if (evalSettings.pureEval && !flakeRef.input.isImmutable())
         throw Error("cannot use ':load-flake' on mutable flake reference '%s' (use --impure to override)", flakeRefS);
 
-    Value v;
+    auto v = state->allocValue();
```
regnat, line 618: What’s the reason for the change? Having `v` as a plain stack-allocated `Value` is nice since it has a lexical lifetime.

src/nix/repl.cc
```diff
@@ -624,8 +624,17 @@ void NixRepl::loadFlake(const std::string & flakeRefS)
                 .useRegistries = !evalSettings.pureEval,
                 .allowMutable  = !evalSettings.pureEval,
             }),
-        v);
-    addAttrsToScope(v);
+        *v);
+
+    auto f = v->attrs->get(state->symbols.create(fragment));
```
regnat, line 629: I think this won’t work properly if `fragment` has more than one level (`foo.bar`). You probably need to split it first with `parseAttrPath` and recursively get the right value.

regnat (commented):
> As a small bonus, I've implemented the ability for the :load-flake command to parse flake ref fragments (i.e. the part after the #)

Yay! That’s cool :sunglasses: 

> when experimental feature "flakes" is enabled, it makes sense to consider the arguments as flake refs instead of files

That’s a tricky thing to decide. Because it means breaking (again) backwards-compatibility once flake aren’t experimental anymore.

Maybe a nicer path would be to

- Add a `--file` flag for loading a file
- Add a `--flake` flag for loading a flake
- Deprecate (but keep) the current behavior and recommend to use `--file` instead
- Eventually (after flakes are non-experimental anymore, and after at least 2 or 3 6w cycles) change the behavior

Possibly (dunno how much work that would be), the last step could be done by moving the cli of `nix repl` to use the same mechanism as the other commands (making it an instance of `InstallableCommand`) to make it fully consistent with the rest

> It would also be nice to include the same search behavior as other flake ref parsing commands, e.g. nixpkgs#zlib becomes nixpkgs#legacyPackages.$current_system.zlib

Yes, that sounds like that would be great (and probably also for `:lf`)

src/nix/repl.cc
nrdxp: Oh I think I copied that from somewhere else while try to debug a stack overflow. I'll change it back :+1:

src/nix/repl.cc
nrdxp: I'll give that a shot. While we are at it, is there a function or method somewhere used to traverse the multiple outputs to find a match?

src/nix/repl.cc
```diff
@@ -624,8 +624,17 @@ void NixRepl::loadFlake(const std::string & flakeRefS)
...
+
+    if (f == 0) {
```
edolstra, line 631:
```suggestion
    if (!f) {
```

src/nix/repl.cc
```diff
@@ -624,8 +624,17 @@ void NixRepl::loadFlake(const std::string & flakeRefS)
...
+        warn("no attribute %s, nothing loaded", fragment);
```
edolstra, line 632:
```suggestion
        warn("attribute '%s' does not exist, nothing loaded", fragment);
```

src/nix/repl.cc
```diff
@@ -624,8 +624,17 @@ void NixRepl::loadFlake(const std::string & flakeRefS)
...
+        return;
+    };
+
+    fragment != "" ? addAttrsToScope(*f->value) : addAttrsToScope(*v);
```
edolstra, line 636:
```suggestion
    addAttrsToScope(fragment != "" ? *f->value : *v);
```

src/nix/repl.cc
```diff
@@ -654,7 +663,10 @@ void NixRepl::reloadFiles()
         if (!first) notice("");
         first = false;
         notice("Loading '%1%'...", i);
-        loadFile(i);
+
+        settings.isExperimentalFeatureEnabled(Xp::Flakes)
+        ? loadFlake(i)
+        : loadFile(i);
```
edolstra, line 669:
```suggestion
        if (settings.isExperimentalFeatureEnabled(Xp::Flakes))
            loadFlake(i);
        else
            loadFile(i);
```
since it's not really idiomatic in C++ (IMHO) to have the conditional operator at statement level.

edolstra (approved):
Could you mention this in the release notes?

tomberek:
Added a bit of cleanup here: https://github.com/flox/nix/commit/da4dcb5c64b0b248e1dc730a6666f97378940071

After this i wanted to start using InstallableCommand,InstallablesCommand, or SourceExprCommand so that we are using the Installable concept instead of having a conditional on an experimental flag. Directly using that works, but also brings some unwanted default behavior:
```
nix repl .#
```
should probably NOT use a default attrpath of `defaultPackages.${system}`. At least my expectation is to preserve @nrdxp 's implemented behavior where that would load the output attrs. (Though it would also be good to have the inputs?)

tomberek:
Working on https://github.com/NixOS/nix/pull/6233

PR #5826, opened 2021-12-23
Properly return false on `builtins.pathExists /someNonAllowedPath`

regnat:
Follow-up from https://github.com/NixOS/nix/pull/5807 to fix https://github.com/NixOS/nix/pull/5807#issuecomment-1000135394

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -89,18 +89,28 @@ StringMap EvalState::realiseContext(const PathSet & context)
     return res;
 }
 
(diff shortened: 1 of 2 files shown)


balsoft (approved):

regnat:
actually the fix is wrong because the new catch is too eager:

```console
$ nix eval --expr 'builtins.pathExists (builtins.seq (builtins.readFile "/etc/hosts") /foo)'
false
```

^ that shouldn’t happen

PR #5827, opened 2021-12-23
Use int64_t and NixFloat in fromTOML types

nmattia:
This makes sure that values parsed from TOML have a proper size. Using
e.g. `double` caused issues on i686 where the size of `double` (32bit)
was too small to accommodate some values.


nmattia:
@edolstra any chance we can test this on i686?

edolstra:
@nmattia I tested it using `nix build .#defaultPackage.i686-linux` which also works on an `x86_64-linux` machine.

PR #5829, opened 2021-12-24
Push docker image to docker hub for each commit on master

garbas:


garbas:
I'll reopen this PR once I have a working version

.github/workflows/ci.yml
```diff
@@ -74,3 +74,33 @@ jobs:
...
+        authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+    - run: nix-build -A checks.$(nix-instantiate --eval -E 'builtins.currentSystem' --json).dockerImage
+    - run: docker load -i ./result/image.tar.gz
+    - run: docker tag nix:$NIX_VERSION nixos/nix:$NIX_VERSION
+    - name: Login to Docker Hub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+    - run: docker push nixos/nix:$NIX_VERSION
```
SuperSandro2000, line 106: I don't think that is exactly what we would want. The version number should be the release tag of that version while builds directly from master should be tagged with master or nightly or similar. ``latest`` should always point to the latest full release. Right now if the version number isn't updated right after a release and a merge goes to master before that we would override the release version with something potential very different/half broken.

.github/workflows/ci.yml
```diff
@@ -74,3 +74,33 @@ jobs:
...
+    - run: echo NIX_VERSION="$(nix-instantiate --eval -E '(import ./default.nix).defaultPackage.${builtins.currentSystem}.version' | tr -d \")" >> $GITHUB_ENV
+    - uses: cachix/cachix-action@v10
+      if: needs.check_cachix.outputs.secret == 'true'
+      with:
+        name: '${{ env.CACHIX_NAME }}'
+        signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+        authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+    - run: nix-build -A checks.$(nix-instantiate --eval -E 'builtins.currentSystem' --json).dockerImage
+    - run: docker load -i ./result/image.tar.gz
+    - run: docker tag nix:$NIX_VERSION nixos/nix:$NIX_VERSION
```
SuperSandro2000, line 100: Instead of retagging we could also overwrite the name https://github.com/NixOS/nix/blob/master/docker.nix#L3

.github/workflows/ci.yml
edolstra: When releasing from `master`, `officialRelease` is set to false, so the version will be something like `2.6.0pre<date>` rather than `2.6.0`. So it won't overwrite any official releases.

.github/workflows/ci.yml
garbas: I like to make think explicit in the deploy scripts and not rely too much on defaults that we would put in `docker.nix`

PR #5830, opened 2021-12-25
add zipAttrsWith primop

pennae:
nixpkgs can save a good bit of eval memory with this primop. zipAttrsWith is
used quite a bit around nixpkgs (eg in the form of recursiveUpdate), but the
most costly application for this primop is in the module system. it improves
the implementation of zipAttrsWith from nixpkgs by not checking an attribute
multiple times if it occurs more than once in the input list, allocates less
values and set elements, and just avoids many a temporary object in general.

nixpkgs has a more generic version of this operation, zipAttrsWithNames, but
this version is only used once so isn't suitable for being the base of a new
primop. if it were to be used more we should add a second primop instead.

see NixOS/nixpkgs#152046


pennae:
performance numbers for `nix eval` of a system config (using the linked PR for zipAttrsWith in the modules system):

| metric                  |  master    |  __zipAttrsWith | change |
|-------------------------|------------|-----------------|--------|
| cpuTime | 6.93235 | 6.81052 | -1.75% |
| envs.number             |    6282819 |    6013530      | -4.28% |
| envs.elements           |    8788832 |    8519543      | -3.06% |
| envs.bytes              |  170835760 |  164372824      | -3.78% |
| list.elements           |    2714619 |    2273941      | -16.2% |
| list.bytes              |   21716952 |   18191528      | -16.2% |
| values.number           |   11340948 |   11036880      | -2.68% |
| values.bytes            |  272182752 |  264885120      | -2.68% |
| sets.number             |    1789875 |    1619120      | -9.54% |
| sets.bytes              |  387449112 |  374397536      | -3.36% |
| sets.elements           |   14950463 |   14520484      | -2.87% |
| nrThunks                |    8363654 |    8071073      | -3.49% |
| nrAvoided               |    6674387 |    6216964      | -6.85% |
| nrPrimOpCalls           |    2733491 |    2511446      | -8.12% |
| nrFunctionCalls         |    5600747 |    5331458      | -4.80% |
| gc.heapSize             |  638251008 |  604696576      | -5.25% |
| gc.totalBytes           | 1041608032 | 1007749824      | -3.25% |

no significant changes in other metrics.

src/libexpr/primops.cc
```diff
@@ -2472,6 +2472,76 @@ static RegisterPrimOp primop_mapAttrs({
...
+
+    v.attrs->sort();
+}
+
+static RegisterPrimOp primop_zipAttrsWith({
+    .name = "__zipAttrsWith",
+    .args = {"f", "list"},
+    .doc = R"(
+      Merge sets of attributes and use the function f to merge attributes
+      values.
```
roberth, line 2534: Merging is a little vague, especially when used both for the individualized attributes and for f.

```suggestion
      Transpose a list of attribute sets into an attribute set of lists, then apply `mapAttrs`.

      `f` receives two arguments: the attribute name and a non-empty list of all values encountered for that attribute name.

      The result is an attribute set where the attribute names are the union of the attribute names in each element of `list`. The attribute values are the return values of `f`.
```

src/libexpr/primops.cc
```diff
@@ -2472,6 +2472,76 @@ static RegisterPrimOp primop_mapAttrs({
...
+
+      ```nix
+      builtins.zipAttrsWith (name: vs: vs) [{a = "x";} {a = "y"; b = "z";}]
```
roberth, line 2537:
```suggestion
      builtins.zipAttrsWith (name: vs: { inherit name vs; }) [{a = "x";} {a = "y"; b = "z";}]
```

This would make for a more complete example, if a little more verbose. (also update value below)

src/libexpr/primops.cc
```diff
@@ -2472,6 +2472,76 @@ static RegisterPrimOp primop_mapAttrs({
...
+
+    std::map<Symbol, size_t> attrOccurences;
+    std::map<Symbol, Value * *> lists;
+
+    state.forceFunction(*args[0], pos);
+    state.forceList(*args[1], pos);
+    const auto nrLists = args[1]->listSize();
+
+    for (unsigned int n = 0; n < nrLists; ++n) {
+        Value * vElem = args[1]->listElems()[n];
```
roberth, line 2492: I don't know if the compiler is clever enough to float out the `listElems()` logic, but it's also more readable if you call it once and put it in a nice variable, imo.
Also applies below.

src/libexpr/primops.cc
pennae: that documentation was lifted from nixpkgs, but your version is much better. :+1:

src/libexpr/primops.cc
```diff
@@ -2472,6 +2472,93 @@ static RegisterPrimOp primop_mapAttrs({
...
+{
+    // we will first count how many values are present for each given key.
+    // we then allocate a single attrset and pre-populate it with lists of
+    // appropriate sizes, stash the pointers to the list elements of each,
+    // and populate the lists. after that we replace the list in the every
+    // attribute with the merge function application. this way we need not
+    // use (slightly slower) temporary storage the GC does not know about.
+
+    std::map<Symbol, size_t> attrOccurences;
+    std::map<Symbol, Value * *> lists;
```
edolstra, line 2485: `std::unordered_map` is probably faster. Also, it should be possible to merge these maps (`std::unordered_map<Symbol, std::tuple<size_t, Value * * >>`, where the `Value * *` is filled in in the second pass).

pennae: `Symbol` isn't hashable yet, we'd have to add that. didn't do it because in our tests `unordered_map` was actually slower than `map`. (the sort at the end of the function is actually a leftover from those tests that snuck through)

edolstra:
@pennae Looks good, can you update the release notes?

pennae:
added a note similar to the recent `groupBy` :)

edolstra:
Thanks!

Issue #5736, opened 2021-12-06
`nix repl`: show trace


balsoft:
Currently, `nix repl --show-trace` doesn't work as expected: no trace is shown on error. Apart from being confusing, this is generally unfortunate: repl is the place where you're most likely to encounter errors, so it is extremely useful to be able to trace them.

I propose to fix `nix repl --show-trace` (and then show the trace for every error encountered), and also to add a `:trace` or `:show-trace` repl command to toggle this functionality while already in the repl.

PR #5840, opened 2021-12-28
nix repl: fix --show-trace and add the ability to set trace display

balsoft:
- BaseError::calcWhat: take loggerSettings.showTrace into account
- Add a test that nix repl --show-trace actually shows the trace
- Add ability to toggle show-trace from within the repl

Closes https://github.com/NixOS/nix/issues/5736

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -430,7 +430,8 @@ bool NixRepl::processLine(string line)
              << "  :t <expr>     Describe result of evaluation\n"
              << "  :u <expr>     Build derivation, then start nix-shell\n"
              << "  :doc <expr>   Show documentation of a builtin function\n"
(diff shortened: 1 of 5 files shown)


src/nix/repl.cc
```diff
@@ -572,6 +573,18 @@ bool NixRepl::processLine(string line)
             throw Error("value does not have documentation");
     }
 
+    else if (command == ":st" || command == ":show-trace") {
+        if (arg == "false" || (arg == "" && loggerSettings.showTrace)) {
+            std::cout << "not showing error traces\n";
+            loggerSettings.set("show-trace", "false");
```
edolstra, line 579: I think `loggerSettings.showTrace = false` should also work.

PR #5846, opened 2021-12-30
libexpr: printValue improvements

ckiee:
I was trying to trace an attrset of evaluated `evalConfig`s (NixOS) and Nix threw me the bad error produced by this function. Tidied it up a bit.

I've just been testing this by overriding my `nixUnstable` adding paths to the `patches` attribute and this codepath is pretty hard to hit so I guess that'll have to do.

(Oh also, this is my first Nix PR! Hi!)

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -157,6 +157,9 @@ void printValue(std::ostream & str, std::set<const Value *> & active, const Valu
     case tFloat:
         str << v.fpoint;
         break;
+    case tBlackhole:
+        str << "<BLACKHOLE>";
+        break;
     default:
         debug(format("invalid value %1%, aborting") % showType(v));
         abort();
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -158,6 +158,7 @@ void printValue(std::ostream & str, std::set<const Value *> & active, const Valu
         str << v.fpoint;
         break;
     default:
+        debug(format("invalid value %1%, aborting") % showType(v));
         abort();
     }
 
@@ -197,6 +198,7 @@ string showType(ValueType type)
         case nFloat: return "a float";
         case nThunk: return "a thunk";
     }
+    debug(format("invalid type id %1%, aborting") % type);
     abort();
 }
 


src/libexpr/eval.cc
```diff
@@ -144,8 +144,11 @@ void printValue(std::ostream & str, std::set<const Value *> & active, const Valu
     case tFloat:
         str << v.fpoint;
         break;
+    case tBlackhole:
+        str << "<BLACKHOLE>";
+        break;
     default:
-        throw Error("invalid value");
+        throw Error("invalid value %1%", showType(v));
```
edolstra, line 151: I think that this aborts if the type is unknown.

ckiee: Oh yeah, you're right: should I replace the `abort` with a `throw` in there?

ckiee: Ok, now it doesn't abort: (intentional bug for testing)
```
ckie@cookiemonster ~/git/nixfiles -> nix eval --expr 'builtins.trace (x: 1) 1'
error: invalid type id 100
```

edolstra: No, I think we should continue to abort, since invalid values are bugs in Nix, not user errors.

edolstra: So actually I would change this line to `abort()` just as in `showType()`.

ckiee: That makes it significantly more painful to debug: I ran into the error message randomly so I grepped for it and found the source.
Can we log *and* abort?

edolstra: `abort()` makes it easier to debug since in gdb it puts you right on the spot where the crash happened. Whereas if you throw an exception, you have to mess around with `catch throw`.

Ericson2314:
`abort` does seems better to me too. But the PR does this.

Did you encounter black holes, @ckiee?

ckiee:
Yes, though it was 2 years ago at this point and I was knee-deep debugging some weird thing with the module system's interaction with my config IIRC.

ckiee:
No reprod in 2023, closing.

PR #5865, opened 2022-01-05
be more memory friendly

pennae:
we do a lot of single-object allocations that needn't be done at all, are could at least be batched like the Value allocations are as of recently. singleton Envs are a good candidate since they make up the vast majority of envs allocated. we can also inline some allocation and check functions to avoid stack churn, and with memory batching get about 1.5% improvement on eval.

### DIFF
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -42,6 +42,7 @@ inline void * allocBytes(size_t n)
 [[gnu::always_inline]]
 Value * EvalState::allocValue()
 {
+#if HAVE_BOEHMGC
     /* We use the boehm batch allocator to speed up allocations of Values (of which there are many).
        GC_malloc_many returns a linked list of objects of the given size, where the first word
        of each object is also the pointer to the next object in the list. This also means that we
@@ -56,6 +57,9 @@ Value * EvalState::allocValue()
     void * p = *valueAllocCache;
     *valueAllocCache = GC_NEXT(p);
     GC_NEXT(p) = nullptr;
+#else
+    void * p = allocBytes(sizeof(Value));
+#endif
 
     nrValues++;
     return (Value *) p;
@@ -70,9 +74,8 @@ Env & EvalState::allocEnv(size_t size)
 
     Env * env;
 
-    if (size != 1)
-        env = (Env *) allocBytes(sizeof(Env) + size * sizeof(Value *));
-    else {
+#if HAVE_BOEHMGC
+    if (size == 1) {
         /* see allocValue for explanations. */
         if (!*env1AllocCache) {
             *env1AllocCache = GC_malloc_many(sizeof(Env) + sizeof(Value *));
@@ -83,7 +86,9 @@ Env & EvalState::allocEnv(size_t size)
         *env1AllocCache = GC_NEXT(p);
         GC_NEXT(p) = nullptr;
         env = (Env *) p;
-    }
+    } else
+#endif
+        env = (Env *) allocBytes(sizeof(Env) + size * sizeof(Value *));
 
     env->type = Env::Plain;
 
--- a/src/libexpr/eval-inline.hh
+++ b/src/libexpr/eval-inline.hh
@@ -24,6 +24,76 @@ LocalNoInlineNoReturn(void throwTypeError(const Pos & pos, const char * s, const
 }
 
 
+/* Note: Various places expect the allocated memory to be zeroed. */
+[[gnu::always_inline]]
+inline void * allocBytes(size_t n)
+{
+    void * p;
+#if HAVE_BOEHMGC
+    p = GC_MALLOC(n);
+#else
+    p = calloc(n, 1);
+#endif
+    if (!p) throw std::bad_alloc();
+    return p;
+}
+
+
+[[gnu::always_inline]]
+Value * EvalState::allocValue()
+{
+    /* We use the boehm batch allocator to speed up allocations of Values (of which there are many).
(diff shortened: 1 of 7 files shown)


src/libexpr/eval-inline.hh
```diff
@@ -30,6 +30,77 @@ LocalNoInlineNoReturn(void throwTypeError(const Pos & pos, const char * s, const
...
+
+[[gnu::always_inline]]
+Value * EvalState::allocValue()
+{
+    /* We use the boehm batch allocator to speed up allocations of Values (of which there are many).
+       GC_malloc_many returns a linked list of objects of the given size, where the first word
+       of each object is also the pointer to the next object in the list. This also means that we
+       have to explicitly clear the first word of every object we take. */
+    if (!valueAllocCache) {
+        valueAllocCache = GC_malloc_many(sizeof(Value));
```
andir, line 56: Should this have a macro guard as boehm might not be available?

pennae: indeed it should. forgot about that in the first instance and just duplicated the mistake ... :sweat_smile:

src/libutil/util.cc
```diff
@@ -678,31 +678,36 @@ void drainFD(int fd, Sink & sink, bool block)
...
+    if (!block) {
+        saved = fcntl(fd, F_GETFL);
+        if (fcntl(fd, F_SETFL, saved | O_NONBLOCK) == -1)
+            throw SysError("making file descriptor non-blocking");
+    }
+
     Finally finally([&]() {
         if (!block) {
+            #pragma GCC diagnostic push
+            // gcc detects `saved` as maybe uninitialized
```
edolstra, line 690: Easier to just do `int saved = 0` to shup up the warning.

pennae: changed to do that instead.

thufschmitt (approved):

pennae:
@edolstra except rebasing onto current master (which we're currently doing, plus rerunning benchmarks), is there anything this needs done? the boehm guard patches should definitely go into master, the others are optional but do help performance a bit.

edolstra:
I've added it to the 2.8 milestone since it's best not to merge a few days before release :-)

src/libutil/util.cc
```diff
@@ -691,24 +698,19 @@ void drainFD(int fd, Sink & sink, bool block)
...
     });
 
-    if (!block) {
-        saved = fcntl(fd, F_GETFL);
-        if (fcntl(fd, F_SETFL, saved | O_NONBLOCK) == -1)
-            throw SysError("making file descriptor non-blocking");
-    }
-
-    std::vector<unsigned char> buf(64 * 1024);
+    constexpr size_t BUF_SIZE = 64 * 1024;
```
edolstra, line 701: I don't think this matters for performance, and in general we should avoid large stack allocations. But if we must, then `std::array` or `boost::container::small_vector` are preferred for their safety.

src/libexpr/eval.hh
```diff
@@ -133,9 +133,14 @@ private:
     /* Cache used by prim_match(). */
     std::shared_ptr<RegexCache> regexCache;
 
+#if HAVE_BOEHMGC
     /* Allocation cache for GC'd Value objects. */
     std::shared_ptr<void *> valueAllocCache;
 
+    /* Allocation cache for size-1 Env objects. */
+    std::shared_ptr<void *> env1AllocCache = nullptr;
```
edolstra, line 141:
```suggestion
    std::shared_ptr<void *> env1AllocCache;
```

`std::shared_ptr` doesn't need an initializer.

src/libexpr/eval-inline.hh
```diff
@@ -24,6 +24,81 @@ LocalNoInlineNoReturn(void throwTypeError(const Pos & pos, const char * s, const
...
+    p = GC_MALLOC(n);
+#else
+    p = calloc(n, 1);
+#endif
+    if (!p) throw std::bad_alloc();
+    return p;
+}
+
+
+[[gnu::always_inline]]
```
edolstra, line 42: Is `always_inline` a good idea? Maybe this is best left to the compiler.

edolstra (changes requested):

src/libexpr/eval-inline.hh
pennae: in general no, but in this case it does seem to be. it does save the entire stack frame setup and teardown, which can be a significant part of allocation cost when taking the hot path using the cache. if the cache weren't there it'd definitely be better left to the compiler.

src/libutil/util.cc
pennae: it does, because the vector version includes an unnecessary memset. we could turn this into a `unique_ptr<unsigned char[]>` instead if the stack allocation is a no-go, but in terms of safety it's all a wash because we access the raw memory without bounds checks. (we're aware that this is much easier to exploit on the stack though)

pennae: okay, redid some benchmarks and can't see the difference we made this change for any more. would still prefer stack arrays or `unique_ptr<char[]>`, but we can also drop if that's what you'd prefer

edolstra: I would prefer idiomatic C++ (e.g. `buf.size()`) over C-style `BUF_SIZE` stuff. I expect that the cost of a heap allocation is dwarfed by the `read()` system call.

pennae: can't think of an wasy way to get that without undefined behaviour. reserving space in a vector instead of initializing it to a specific size is technically undefined, so that won't work. `std::array`s would default-initialize their storage as well.

will drop that commit.

PR #5871, opened 2022-01-06
Remove overly-strict ref check

amarshall:
- This claims to be a regex for a Git ref, but it is applied also to
  Mercurial. I was unable to find any docs for the syntax of a Mercurial
  ref, so seems better to just allow anything rather than disallow a
  valid one, blocking a valid use case.
- Again, this claims to be a regex for a Git ref, but it is extremely
  over-restrictive and disallows a variety of valid refs. Instead,
  remove the check entirely and allow either checks for badGitRefRegex
  or Git itself to say it is invalid instead.
- Tests added for some additional valid refs that previously failed.

I think that the badGitRefRegex check should also be removed, but want
feedback on just this first and that can be a follow-up.

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -546,6 +546,8 @@ struct GitInputScheme : InputScheme
                 // we're using --quiet for now. Should process its stderr.
                 try {
                     auto ref = input.getRef();
+                    if (std::regex_match(*ref, badGitRefRegex))
+                        throw BadURL("in input '%s', '%s' is not a branch/tag name", input.to_string(), *ref);
                     auto fetchRef = allRefs
                         ? "refs/*"
                         : ref->compare(0, 5, "refs/") == 0
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -44,7 +44,7 @@ struct GitArchiveInputScheme : InputScheme
         if (size == 3) {
             if (std::regex_match(path[2], revRegex))
                 rev = Hash::parseAny(path[2], htSHA1);
-            else if (std::regex_match(path[2], refRegex))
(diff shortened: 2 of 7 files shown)


edolstra:
I don't think we should allow everything, because it creates an opportunity for all sorts of injection shenanigans when we pass the `ref` to external programs like `git`.

bjornfor:
> I don't think we should allow everything, because it creates an opportunity for all sorts of injection shenanigans when we pass the `ref` to external programs like `git`.

Isn't that up the the external programs to protect themselves?

edolstra:
No, it's generally up to the caller (in this case Nix) not to pass arguments that could be misinterpreted (e.g. a ref that starts with a dash).

amarshall:
We should avoid causing argument injection or shell injection, but that is not necessarily the same as validating the format of the ref. What I think would be ideal as a follow-up (and I plan to get to in the next week) is:

- Add back the existing check for Mercurial only, since I don't really know anything about Mercurial and it may be better to leave its handling as-is.
- Either:
  - Add targeted checks when a Git ref is passed externally, aimed at sanitizing for that specific usage rather than checking the nuances of ref syntax (much of the restrictions are not iinjection problems)
  - Re-add the removed checks, but change them all to check that the string does not match badGitRefRegex

Thoughts?

xokdvium:
> Again, this claims to be a regex for a Git ref, but it is extremely
over-restrictive and disallows a variety of valid refs. Instead,
remove the check entirely and allow either checks for badGitRefRegex
or Git itself to say it is invalid instead.

Does https://github.com/NixOS/nix/pull/13729 solve the over-restrictive issue? That should be slightly more correct. We have also since switched to libgit2 for some things.

tomberek:

PR #5875, opened 2022-01-06
Fix segfault or stack overflow caused by large derivation fields

roberth:
This removes a dynamic stack allocation, making the derivation
unparsing logic robust against overflows when large strings are
added to a derivation.
Overflow behavior depends on the platform and stack configuration.

For instance, x86_64-linux/glibc behaves as (somewhat) expected:

$ (ulimit -s 20000; nix-instantiate tests/lang/eval-okay-big-derivation-attr.nix)
error: stack overflow (possible infinite recursion)

$ (ulimit -s 40000; nix-instantiate tests/lang/eval-okay-big-derivation-attr.nix)
error: expression does not evaluate to a derivation (or a set or list of those)

However, on aarch64-darwin:

$ nix-instantiate big-attr.nix                                                                                                                                                                                                                                                       ~
zsh: segmentation fault  nix-instantiate big-attr.nix

This indicates a slight flaw in the single stack protection page
approach that is not encountered with normal stack frames.
With this change, we avoid both types of failure.

### DIFF
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -272,7 +272,15 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
 
 static void printString(string & res, std::string_view s)
 {
-    char buf[s.size() * 2 + 2];
+    char * buf;
+    size_t bufSize = s.size() * 2 + 2;
+    std::unique_ptr<char[]> dynBuf;
+    if (bufSize < 0x10000) {
+        buf = (char *)alloca(bufSize);
+    } else {
+        dynBuf = decltype(dynBuf)(new char[bufSize]);
+        buf = dynBuf.get();
+    }
     char * p = buf;
     *p++ = '"';
     for (auto c : s)
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -4,6 +4,7 @@
 #include "util.hh"
 #include "worker-protocol.hh"
 #include "fs-accessor.hh"
+#include <boost/container/small_vector.hpp>
 
 namespace nix {
 
@@ -272,19 +273,27 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
 
 static void printString(string & res, std::string_view s)
 {
-    size_t bufSize = s.size() * 2 + 2;
-    withBuffer(bufSize, [&](char *buf) {
-        char * p = buf;
-        *p++ = '"';
-        for (auto c : s)
-            if (c == '\"' || c == '\\') { *p++ = '\\'; *p++ = c; }
-            else if (c == '\n') { *p++ = '\\'; *p++ = 'n'; }
-            else if (c == '\r') { *p++ = '\\'; *p++ = 'r'; }
-            else if (c == '\t') { *p++ = '\\'; *p++ = 't'; }
-            else *p++ = c;
-        *p++ = '"';
-        res.append(buf, p - buf);
-    });
+    // Large stack allocations can skip past the stack protection page.
+    const size_t stack_protection_size = 4096;
+    // We reduce the max stack allocated buffer by an extra amount to increase
+    // the chance of hitting it, even when `fun`'s first access is some distance
+    // into its *further* stack frame, particularly if the call was inlined and
+    // therefore not writing a frame pointer.
+    const size_t play = 64 * sizeof(char *); // 512B on 64b archs
+
+    boost::container::small_vector<char, stack_protection_size - play> buffer;
+    buffer.reserve(s.size() * 2 + 2);
+    char * buf = buffer.data();
+    char * p = buf;
+    *p++ = '"';
+    for (auto c : s)
+        if (c == '\"' || c == '\\') { *p++ = '\\'; *p++ = c; }
+        else if (c == '\n') { *p++ = '\\'; *p++ = 'n'; }
+        else if (c == '\r') { *p++ = '\\'; *p++ = 'r'; }
+        else if (c == '\t') { *p++ = '\\'; *p++ = 't'; }
+        else *p++ = c;
+    *p++ = '"';
+    res.append(buf, p - buf);
 }
 
 
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -273,7 +273,7 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
 static void printString(string & res, std::string_view s)
 {
     size_t bufSize = s.size() * 2 + 2;
-    withBuffer<void, char>(bufSize, [&](char buf[]) {
+    withBuffer(bufSize, [&](char *buf) {
         char * p = buf;
         *p++ = '"';
         for (auto c : s)
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -273,15 +273,7 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
 
 static void printString(string & res, std::string_view s)
 {
-    // Large stack allocations can skip past the stack protection page.
-    const size_t stack_protection_size = 4096;
-    // We reduce the max stack allocated buffer by an extra amount to increase
-    // the chance of hitting it, even when `fun`'s first access is some distance
-    // into its *further* stack frame, particularly if the call was inlined and
-    // therefore not writing a frame pointer.
-    const size_t play = 64 * sizeof(char *); // 512B on 64b archs
-
-    boost::container::small_vector<char, stack_protection_size - play> buffer;
+    boost::container::small_vector<char, 64 * 1024> buffer;
     buffer.reserve(s.size() * 2 + 2);
     char * buf = buffer.data();
     char * p = buf;
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -272,25 +272,19 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
 
 static void printString(string & res, std::string_view s)
 {
-    char * buf;
     size_t bufSize = s.size() * 2 + 2;
-    std::unique_ptr<char[]> dynBuf;
-    if (bufSize < 0x10000) {
-        buf = (char *)alloca(bufSize);
-    } else {
-        dynBuf = decltype(dynBuf)(new char[bufSize]);
-        buf = dynBuf.get();
-    }
-    char * p = buf;
-    *p++ = '"';
-    for (auto c : s)
-        if (c == '\"' || c == '\\') { *p++ = '\\'; *p++ = c; }
-        else if (c == '\n') { *p++ = '\\'; *p++ = 'n'; }
-        else if (c == '\r') { *p++ = '\\'; *p++ = 'r'; }
-        else if (c == '\t') { *p++ = '\\'; *p++ = 't'; }
-        else *p++ = c;
-    *p++ = '"';
-    res.append(buf, p - buf);
+    withBuffer<void, char>(bufSize, [&](char buf[]) {
+        char * p = buf;
+        *p++ = '"';
+        for (auto c : s)
+            if (c == '\"' || c == '\\') { *p++ = '\\'; *p++ = c; }
+            else if (c == '\n') { *p++ = '\\'; *p++ = 'n'; }
+            else if (c == '\r') { *p++ = '\\'; *p++ = 'r'; }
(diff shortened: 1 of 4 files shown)


src/libstore/derivations.cc
```diff
@@ -272,7 +272,15 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
 
 static void printString(string & res, std::string_view s)
 {
-    char buf[s.size() * 2 + 2];
+    char * buf;
+    size_t bufsize = s.size() * 2 + 2;
+    std::string dynbuf;
```
pennae, line 277: please use a `std::unique<char[]>` instead. zero-initializing (as string::resize does) is an unnecessary cost here

src/libstore/derivations.cc
```diff
@@ -272,7 +272,15 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
...
 static void printString(string & res, std::string_view s)
 {
-    char buf[s.size() * 2 + 2];
+    char * buf;
+    size_t bufSize = s.size() * 2 + 2;
+    std::unique_ptr<char[]> dynBuf;
+    if (bufSize < 0x10000) {
+        buf = (char *)alloca(bufSize);
+    } else {
+        dynBuf = decltype(dynBuf)(new char[bufSize]);
```
pennae, line 281:
```suggestion
        dynBuf = std::make_unique<char[]>(bufSize);
```

src/libstore/derivations.cc
```diff
@@ -272,7 +272,15 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
 
 static void printString(string & res, std::string_view s)
 {
-    char buf[s.size() * 2 + 2];
+    char * buf;
+    size_t bufSize = s.size() * 2 + 2;
+    std::unique_ptr<char[]> dynBuf;
+    if (bufSize < 0x10000) {
+        buf = (char *)alloca(bufSize);
```
edolstra, line 279: FYI, alloca is not standard C/POSIX, and definitely not idiomatic C++.

roberth: Ok, I can make it always allocate on the heap if you prefer, but that's going to be the obvious solution in case Nix is ported to such a rare platform.

roberth:
Interestingly, the new test case triggers #3605 when run with a daemon, but not when the derivation field is modified to be "just" 1MiB.
Is it possible that test suite on darwin doesn't run with a daemon?
Regardless, I have manually confirmed that the darwin segfault is triggered by the 1MiB test using `ulimit -s 100`, but not after the PR is applied.

src/libstore/derivations.cc
pennae: you could also move the actual buffer using code into a local lambda and either pass a stack-allocated buffer or a dynamically allocated buffer. avoiding heap allocs in this part of the code is worth it

src/libstore/derivations.cc
roberth: I try to conform to existing code, such as this pattern taken from `serialise.cc`.
I can change both if @edolstra agrees.

src/libutil/util.hh
```diff
@@ -655,5 +655,14 @@ template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
...
 
+template<typename R = void, typename T = char> inline R withBuffer(size_t size, std::function<R (T[])> fun) {
+    if (size < 0x10000) {
+        T buf[size];
+        return fun(buf);
+    } else {
+        auto buf = std::unique_ptr<T[]>(new T[size]);
+        return fun(buf.get());
+    }
+}
```
pennae, line 666: we can get away with not allocating a `function` here, but keep the signature checking:

```suggestion
template<typename T = char, typename Fn>
inline auto withBuffer(size_t size, Fn fun)
  -> std::invoke_result_t<Fn (T *)>
{
    if (size < 0x10000) {
        T buf[size];
        return fun(buf);
    } else {
        auto buf = std::unique_ptr<T[]>(new T[size]);
        return fun(buf.get());
    }
}
```

`T[]` parameters aren't all that useful usually since they'll immediately decay to pointers everywhere, and they have no range information either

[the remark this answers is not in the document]
roberth: I guess `T[]` was a gimmick anyway.

I've tried your code, but I had to remove the `invoke_result_t` part.

I wasn't impressed by the quality of the error message, and type checking still occurs, so do you think it's ok to omit that part?

```
src/libstore/derivations.cc: In function ‘void nix::printString(std::string&, std::string_view)’:
src/libstore/derivations.cc:287:6: error: no matching function for call to ‘withBuffer(size_t&, nix::printString(std::string&, std::string_view)::<lambda(char*)>)’
  287 |     });
      |      ^
In file included from src/libutil/serialise.hh:6,
                 from src/libutil/hash.hh:4,
                 from src/libstore/content-address.hh:4,
                 from src/libstore/path.hh:3,
                 from src/libstore/derivations.hh:3,
                 from src/libstore/derivations.cc:1:
src/libutil/util.hh:659:13: note: candidate: ‘template<class T, class Fn> std::invoke_result_t<Fn(T*)> nix::withBuffer(size_t, Fn)’
  659 | inline auto withBuffer(size_t size, Fn fun)
      |             ^~~~~~~~~~
src/libutil/util.hh:659:13: note:   template argument deduction/substitution failed:
In file included from /nix/store/sjhz1j2d1ssn59f66kqp92xj9mpsww2d-gcc-10.3.0/include/c++/10.3.0/variant:36,
                 from src/libstore/content-address.hh:3,
                 from src/libstore/path.hh:3,
                 from src/libstore/derivations.hh:3,
                 from src/libstore/derivations.cc:1:
/nix/store/sjhz1j2d1ssn59f66kqp92xj9mpsww2d-gcc-10.3.0/include/c++/10.3.0/type_traits: In substitution of ‘template<class _Fn, class ... _Args> using invoke_result_t = typename std::invoke_result::type [with _Fn = nix::printString(std::string&, std::string_view)::<lambda(char*)>(char*); _Args = {}]’:
src/libutil/util.hh:659:13:   required by substitution of ‘template<class T, class Fn> std::invoke_result_t<Fn(T*)> nix::withBuffer(size_t, Fn) [with T = char; Fn = nix::printString(std::string&, std::string_view)::<lambda(char*)>]’
src/libstore/derivations.cc:287:6:   required from here
/nix/store/sjhz1j2d1ssn59f66kqp92xj9mpsww2d-gcc-10.3.0/include/c++/10.3.0/type_traits:2957:11: error: no type named ‘type’ in ‘struct std::invoke_result<nix::printString(std::string&, std::string_view)::<lambda(char*)>(char*)>’
 2957 |     using invoke_result_t = typename invoke_result<_Fn, _Args...>::type;
```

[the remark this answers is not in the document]
pennae: yeah, it's okay to omit that part. we mainly put it there to document the interface, the compiler will run the same check in the function body anyway.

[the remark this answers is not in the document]
pennae: just checked the reference and it should've been `std::invoke_result_t<Fn, T*>`. we used the signature of `result_of`, which is not the signature of `invoke_result` :facepalm:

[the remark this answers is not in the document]
roberth: Done. Thank you for the review :)

dasJ:
I can confirm this fixes our Hydra Queue Runner crashing. It would be nice to somehow fix that Darwin situation

roberth:
> It would be nice to somehow fix that Darwin situation

I think you were referring to a flaky test result in CI and I agree.

roberth:
@edolstra This is ready for review

pennae:
we've run the current head with decreased buffer size through our benchmarks and haven't seen a significant impact. :+1:

roberth:
@thufschmitt I believe this is ready to be merged.

roberth:
Could you add backport labels?

thufschmitt (changes requested):
Just a small request for some extra doc, but :+1: otherwise

src/libstore/derivations.cc
```diff
@@ -272,17 +272,19 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
...
-    for (auto c : s)
-        if (c == '\"' || c == '\\') { *p++ = '\\'; *p++ = c; }
-        else if (c == '\n') { *p++ = '\\'; *p++ = 'n'; }
-        else if (c == '\r') { *p++ = '\\'; *p++ = 'r'; }
-        else if (c == '\t') { *p++ = '\\'; *p++ = 't'; }
-        else *p++ = c;
-    *p++ = '"';
-    res.append(buf, p - buf);
+    size_t bufSize = s.size() * 2 + 2;
+    withBuffer(bufSize, [&](char *buf) {
```
edolstra, line 276: Maybe we can use `boost::container::small_vector<char, N>` instead? See c9fc975259e27220caeb4291f3dff453e65f1965 for an example.

roberth: This causes a significant stack allocation even when we use the heap. Should we sacrifice that?

Alternatively I could change `withBuffer` to a RAII type instead of a higher order function.

edolstra: Stack allocations are essentially free (in a non-recursive function) so I don't think that's a problem.

roberth: Update: can't convert `withBuffer` to RAII without `alloca`, because the number of elements would need to be static. I've tried a couple of things but couldn't make it work, and it would make sense that structs and stack frames are different, even if they look similar.

@pennae What do you think? Should we sacrifice some stack space to use boost `small_vector`?

edolstra: What's the concern with stack space? A stack allocation is at most one machine instruction.

roberth: Oh, didn't see @edolstra's message; didn't mean to ignore him.

> Stack allocations are essentially free

I figured there's the cost of ~1 extra page in page tables or whatnot, as the `res.append` stack frame shifts, but I don't know if that's significant.

pennae: @roberth even if we do "waste" stack space with a `small_vector`, we'll still waste less than before with your current stack buffer size. page faults are amortized, and we're not in the territory where we have to worry a lot about TLB misses. if it makes the code more readable we'd say it's well worth it :)

roberth: @edolstra Done. Could you review?

src/libstore/derivations.cc
```diff
@@ -272,7 +273,17 @@ Derivation parseDerivation(const Store & store, std::string && s, std::string_vi
...
-    char buf[s.size() * 2 + 2];
+    // Large stack allocations can skip past the stack protection page.
+    const size_t stack_protection_size = 4096;
+    // We reduce the max stack allocated buffer by an extra amount to increase
+    // the chance of hitting it, even when `fun`'s first access is some distance
+    // into its *further* stack frame, particularly if the call was inlined and
+    // therefore not writing a frame pointer.
+    const size_t play = 64 * sizeof(char *); // 512B on 64b archs
+
+    boost::container::small_vector<char, stack_protection_size - play> buffer;
```
edolstra, line 284:
```suggestion
    boost::container::small_vector<char, 64 * 1024> buffer;
```
I don't think we need to worry too much about overflows for non-recursive functions. But if we're worried about this, it's probably better to use gcc's `-fstack-check` flag, which enables stack probe instructions automatically.

roberth: > I don't think we need to worry too much about overflows for non-recursive functions

_Some_ limit on stack usage is necessary though, because derivation attributes can be of disproportional size; multiple megabytes.

Considering the distribution of sizes, it really doesn't matter how big you go beyond the first page. Probably well over 95% already fits anyway. The benefit of staying within a page size is not needing `-fstack-check` in this location.

`-fstack-check` is definitely a good idea for improving the correct functioning of stack overflow detection and memory safety. We'll need to know the overhead though, and what to do for llc. Can we please scope this out?

I'd *really* like to move on with this and I hope you do too. If we really need a fourth iteration on this simple change, just tell me exactly what to do so we can be done with it.

edolstra: > just tell me exactly what to do so we can be done with it.

I ded provide a code suggestion, namely use 64 KB instead of a complicated calculation involving page sizes and char * sizes.

roberth: Alright, apparently I was focusing way too much on the regular comments. Thanks for being patient :)
I've committed it. Good to go now?

PR #5887, opened 2022-01-07
avoid streams

pennae:
istreams and ostreams have a performance cost associated with them that isn't always worth paying. avoiding them in favor of other methods of creating/parsing strings gives us about 2% performance improvement on system eval and a bit less on `nix search`, while also reducing memory pressure a bit.


src/libutil/json.cc
```diff
@@ -7,16 +7,38 @@ namespace nix {
...
+            put('u');
+            put(hex[(uint16_t(*i) >> 12) & 0xf]);
+            put(hex[(uint16_t(*i) >>  8) & 0xf]);
+            put(hex[(uint16_t(*i) >>  4) & 0xf]);
+            put(hex[(uint16_t(*i) >>  0) & 0xf]);
+        }
+        else put(*i);
+    }
+    put('"');
+    flush();
```
thufschmitt, line 41: That looks good (didn’t double-check the benchmarks, but I trust them). Just surprised: Isn’t there a native way of wrapping a stream to make it buffered (while keeping the `ostream` interface)?

src/libexpr/eval.cc
```diff
@@ -1660,13 +1684,31 @@ void EvalState::concatLists(Value & v, size_t nrLists, Value * * lists, const Po
...
+    const auto c_str = [&] {
+        char * result = allocString(sSize + 1);
+        char * tmp = result;
+        for (const auto & part : s) {
+            memcpy(tmp, part.c_str(), part.size());
+            tmp += part.size();
+        }
+        *tmp = 0;
+        return result;
+    };
```
thufschmitt, line 1710: I guess this isn’t defined as `str().c_str()` because we want to use `allocString` to remove a copy. If so, it’s probably worth a comment

thufschmitt (approved):
Looks good, thanks! (and nice performance improvements)

Just a couple of minor comments/questions, but nothing really blocking

src/libutil/json.cc
pennae: buffering is not the problem, streams are properly buffered already. the problem here is per-operation overhead of streams, eg the construction of sentry objects. those aren't free but do nothing we need done for each character, so it's overhead we should try to avoid.

src/libexpr/eval.cc
pennae: exactly. will add a explanation.

src/libutil/json.cc
thufschmitt: Oh, I see. A bit of a shame that we have to resort to something manual like that, but makes sense.

thufschmitt (approved):
👍 Just waiting for the CI to finish before merging.

(@pennae feel free to ping me if I forget)

src/libutil/json.cc
pennae: really is :slightly_frowning_face:

PR #5890, opened 2022-01-08
Allow CLI to pass environment variables to 

dramforever:
This is a WIP pull request implementing an option `environment`, which is a list of items, each being one of:

- `name=value`: Set environment variable `name` to `value`.
- `name`: Inherit environment variable `name` from current environment.

These environment variables will then be sent to the daemon side worker process, which will be used for daemon operations and `impureEnvVars`.

An example use case is for proxy environment variables. Currently to use a proxy in a multi-user setting one needs to edit e.g. `nix-daemon.service` (or provide an override file), or use `launchctl setenv`, and then restart the daemon. With this PR, one only needs to add an option:

```console
$ nix build --environment https_proxy=socks5://...
```

You can also put this in your `nix.conf`

```
environment = https_proxy
```

And all `nix` command will transparently honor `https_proxy`, even if it calls to a daemon.

Since environment variables can affect various aspects, this, as most settings, is available only to trusted users. Could there also be a `trusted-environment` list that allows arbitrary users to provide e.g. proxies? (Substitution should still be verifiable with public keys.) Could trusted users also be restricted from setting arbitrary variables?

Marked draft as I'm still playing around with this feature, and also for discussion.


dramforever:
I think this one is ready for bikeshedding

tests/environment.nix
```diff
@@ -0,0 +1,12 @@
+{ var, value }:
+
+derivation {
```
thufschmitt, line 3: You can use `mkDerivation` (from `config.nix`) here

src/libstore/globals.cc
```diff
@@ -252,6 +254,47 @@ void PluginFilesSetting::set(const std::string & str, bool append)
...
+        hasInherited = false;
+    }
+
+    StringMap newValues;
+
+    for (auto & elem : tokenizeString<Strings>(str)) {
+        size_t pos = elem.find('=');
+        if (pos == std::string::npos) {
+            // name
+            char * value = getenv(elem.c_str());
```
thufschmitt, line 278: Nix has its own `getenv` wrapper to directly manipulate c++ strings:

```suggestion
            std::optional<std::string> value = getEnv(elem);
```

(and changing the uses below)

thufschmitt (commented):
I quite like the broad idea of making it possible to pass stuff like `http_proxy` directly from the client.

One thing I’m not sure about is the fact that this changes then environment of the daemon itself rather than of the builders. This seems a bit dangerous (and might also break if the daemon stops forking for each connection)

src/libstore/globals.cc
```diff
@@ -83,6 +83,8 @@ void loadConfFile()
        ~/.nix/nix.conf or the command line. */
     globalConfig.resetOverridden();
 
+    settings.environment.overridden = settings.environment.hasInherited;
```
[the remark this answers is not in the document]
dramforever: Added a comment above this line. Basically if we get an element from the current environment we should consider it to be set dynamically so that when we contact the daemon we do send this over, even though we have the same config file as the daemon. Supports the `environment = https_proxy` use case.

This might be a bad idea, but this is the simplest thing that should work that I can think of.

src/libstore/globals.cc
dramforever: As you said :ok_hand:

tests/environment.nix
dramforever: As you said :ok:

src/libstore/globals.hh
```diff
@@ -966,6 +984,23 @@ public:
...
+          - `name`: Inherit environment variable `name` from current
+            environment.
+
+          If the user is trusted (see `trusted-users` option), the daemon will
+          also have these environment variables set.
+
+          This option is useful for, e.g., setting `https_proxy` for
+          fixed-output derivations and substituter downloads in a multi-user
+          Nix installation.
+        )"};
```
SuperSandro2000, line 1003: I think this is missing a warning that it could potentially break things. For example what happens if I would do a typo and overwrite important envs like HOME, TMP or NIX_STORE?

SuperSandro2000:
> An example use case is for proxy environment variables. Currently to use a proxy in a multi-user setting one needs to edit e.g. nix-daemon.service (or provide an override file), or use launchctl setenv, and then restart the daemon. 

You probably would want to use ``systemctl edit nix-daemon``. Also most of the time you can restart the daemon without aborting builds.

src/libstore/globals.hh
dramforever: A 'typo' here honestly feels unlikely, but I agree that a warning that this is potentially breaking.

dramforever:
> You probably would want to use `systemctl edit nix-daemon`. Also most of the time you can restart the daemon without aborting builds.

I personally feel this is more annoying than necessary. When I see e.g. that some FOD fails to fetch on my network, or cache.nixos.org is being super slow again, it's much more convenient to just retry with `https_proxy=socks5://... nix build` rather than having to `systemctl edit`, retry the build, and edit it back. On the other hand, I don't like having to have the proxy to be always set because usually `nix build` goes fine and the proxy also doesn't work 100% of the time...

This is likely relevant in places with restrictive networks, be it corporate networks or certain geographical areas.

edolstra:
I'm not really in favor of this since it will make it impossible to switch in the future to a daemon that uses threads rather than forking. It's also a giant security hole to allow users to override the environment of a root process. While trusted users on NixOS essentially have root access, this isn't (or shouldn't be) the case on non-NixOS systems. I think a better approach is to identify the use cases for this (e.g. overriding proxy settings) and introduce separate config settings for that.

thufschmitt:
@edolstra strongly agree with the issues of changing the environment of the daemon. That being said, I think we could get most of the usefulness of this without the drawbacks by only setting the variables in builder (well, the ones that are in `impureEnvVars`). That would solve the `http*_proxy` in particular with (as far as I can see) no serious drawback

dramforever:
I agree that this would be an improvement even if it only applied to `impureEnvVars`. It would also make the code cleaner. Should we just do that first and then figure out proxy settings for daemon-initiated downloads later on?

oxalica:
> I think a better approach is to identify the use cases for this (e.g. overriding proxy settings) and introduce separate config settings for that.

Besides the proxy config, I encountered another issue about remote builders. I use gpg-agent for SSH authentication and my key is in a CanoKey (like YubiKey). Since the key is exclusively occupied by gpg-agent, configuring another gpg-agent for root is unacceptable. One possible way is to create another SSH key stored in say `/root/.ssh`, which has trused-users, aka. root permission on the build machine. But this completely voids the security advantage of storing key in Canokey in the first place. A better way is to allow passing `SSH_AUTH_SOCK` to nix-daemon and it is covered by this PR.

fricklerhandwerk:
@dramforever are you still interested in going the `impureEnvVars` route? Nix maintainers would support that. If not, we should close this PR as there is no clear way forward.

PR #5903, opened 2022-01-12
Release Notes 2.4: add `--indirect` no-op change

Profpatsch:
Since
https://github.com/NixOS/nix/commit/00d25e84577659ccf0bc360c61c47b6cd25d1c26
which was first included in nix 2.4.

It is a backwards-compatible change since the flag will just be
ignored.

cc @edolstra

### DIFF
--- a/doc/manual/src/release-notes/rl-2.4.md
+++ b/doc/manual/src/release-notes/rl-2.4.md
@@ -276,6 +276,9 @@ more than 2800 commits from 195 contributors since release 2.3.
 
 * Plugins can now register `nix` subcommands.
 
+* The `--indirect` flag to `nix-store --add-root` has become a no-op.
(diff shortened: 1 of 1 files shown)


doc/manual/src/release-notes/rl-2.4.md
```diff
@@ -276,6 +276,9 @@ more than 2800 commits from 195 contributors since release 2.3.
 
 * Plugins can now register `nix` subcommands.
 
+* The `--indirect` flag to `nix-store --add-root` is became a no-op,
```
edolstra, line 279:
```suggestion
* The `--indirect` flag to `nix-store --add-root` has became a no-op.
```

Profpatsch: Seems like I couldn’t English today.

PR #5905, opened 2022-01-12
Custom commit messages when commiting the lockfile

lincolnauster:
This allows, for instance, the following command:

```bash
$ nix flake update --commit-lock-file --commit-summary "chore: update nixpkgs"
```

This would lead to a commit message like the following (taken from running the above on nixos/nix):
```
chore: update nixpkgs

Flake lock file updates:

• Updated input 'lowdown-src':
   'github:kristapsdz/lowdown/d2c2b44ff6c27b936ec27358a2653caaef8f73b8' (2021-10-06)
  → 'github:kristapsdz/lowdown/1de10c1d71bfb4348ae0beaec8b1547d5e114969' (2021-11-19)
• Updated input 'nixpkgs':
   'github:NixOS/nixpkgs/82891b5e2c2359d7e58d08849e4c89511ab94234' (2021-09-28)
  → 'github:NixOS/nixpkgs/c68e3678f42352cc4f0ff42f48944c292eef1297' (2022-01-09)

```

If a user wanted a custom commit message, they'd previously have to do something like this (untested, might be wrong):

```bash
$ nix flake update --commit-lock-file
$ git commit --ammend -m "$msg" # also gets rid of the extended commit message
```

There are a variety of reasons one may want to write their own commit message for a dep update, but the biggest one is to conform to the conventions used by a given repository (for instance, in the example, the use of the prefix `chore:` to convey that the commit is routine).

This PR contains two commits; one to update the file and one to document the changes in the changelog. I'm not too familiar with the code here, so (large amounts of) feedback is welcome and probably necessary - thanks!

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -9,3 +9,7 @@
   as `lib.zipAttrsWith` from nixpkgs, but much more efficient.
 * New command `nix store copy-log` to copy build logs from one store
   to another.
+* The `commit-lockfile-summary` option can be set to a non-empty string
+  to override the commit summary used when commiting an updated lockfile.
+  This may be used in conjunction with the nixConfig attribute in
+  `flake.nix` to better conform to repository conventions.
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -633,12 +633,24 @@ LockedFlake lockFlake(
 
                         newLockFile.write(path);
 
+                        std::optional<std::string> commitMessage = std::nullopt;
+                        if (lockFlags.commitLockFile) {
+                            std::string cm;
+
+                            cm = settings.commitLockFileSummary.get();
+
+                            if (cm == "") {
+                                cm = fmt("%s: %s", relPath, lockFileExists ? "Update" : "Add");
+                            }
+
+                            cm += "\n\nFlake lock file updates:\n\n";
+                            cm += filterANSIEscapes(diff, true);
+                            commitMessage = cm;
+                        }
+
                         topRef.input.markChangedFile(
                             (topRef.subdir == "" ? "" : topRef.subdir + "/") + "flake.lock",
-                            lockFlags.commitLockFile
-                            ? std::optional<std::string>(fmt("%s: %s\n\nFlake lock file changes:\n\n%s",
-                                    relPath, lockFileExists ? "Update" : "Add", filterANSIEscapes(diff, true)))
-                            : std::nullopt);
+                            commitMessage);
 
                         /* Rewriting the lockfile changed the top-level
                            repo, so we should re-read it. FIXME: we could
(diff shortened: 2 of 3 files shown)


edolstra:
I'm not sure about this. `--commit-lock-file` is a convenience mechanism that cannot accommodate every use case for what you want the commit message or the commit to look like. E.g. I often squash the change to the lock file with the change to `flake.nix`, but I wouldn't want to add a flag to Nix for that. It seems better to rely on the existing Git mechanism, like `git commit --amend` and `git rebase -i`.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -6,3 +6,6 @@
   set or toggle display of error traces.
 * New builtin function `builtins.zipAttrsWith` with same functionality
   as `lib.zipAttrsWith` from nixpkgs, but much more efficient.
+* The --commit-summary option can be passed wherever a lock file would be
```
edolstra, line 9:
```suggestion
* The `--commit-summary` option can be passed wherever a lock file would be
```

src/libcmd/installables.cc
```diff
@@ -74,6 +74,16 @@ MixFlakeOptions::MixFlakeOptions()
         .handler = {&lockFlags.commitLockFile, true}
     });
 
+    addFlag({
```
edolstra, line 77: It might be nice to change this to a config setting, since then it can be specified on a per-project basis in `flake.nix`'s `nixConfig` attribute.

lincolnauster:
> I'm not sure about this. `--commit-lock-file` is a convenience mechanism
> that cannot accommodate every use case for what you want the commit
> message or the commit to look like. E.g. I often squash the change to
> the lock file with the change to `flake.nix`, but I wouldn't want to add
> a flag to Nix for that.

The way I use this feature, I frequently use `--commit-lock-file` to
automatically generate a commit message while Nix changes the lock file.
It's helpful in that regard for automation: i.e., I can (and frequently
do) have a cron-triggered github workflow to update dependencies and
don't need to rely on unstable awk/grep/whatever hacks to get the old
and new versions out of the lockfiles for the message. However, the
summary doesn't fit with the conventions for the repository. When doing
this automatically, existing mechanisms would probably culminate in some
crazy `git-show` to `head` to `awk` to a file to `git-commit` pipeline.
I like pipelines, but that would not be a fun hack to maintain.

I do much prefer having it as a config setting; I'll look into that
today. Thanks for the feedback, and feel free to close this if it still
seems out of scope!

edolstra:
This gives a test failure in flakes.sh, can you have a look at that? Otherwise it looks good to me.

lincolnauster:
Yep, I'll take a look at this later today. Hopefully it's pretty simple. Thanks for reviewing this!

lincolnauster:
Behavior is fixed (`nix build` checks passing) and rebased onto current master

PR #5906, opened 2022-01-12
optimize primops and utils by caching more and copying less

pennae:
avoiding string copies in in places where string views are sufficient and using cached symbols for constant keys used by primops improves uncached nix search and nixos system eval by about 2%. there's a lot more in utils and the evaluator that can be made to use string views for copy avoidance, but that'll be a separate PR because those changes are based on #5812 (while this one does not).


lincolnauster:
Just for fun, I 'benchmarked' memory usage as measured by valgrind with `nix search --no-eval-cache --offline ../nixpkgs hello`. (It looks like some of the commits have runtime performance benchmarks already.) Here are the results:

```
Current master: 3,701,024,967 bytes allocated
This PR's HEAD: 3,580,387,562 bytes allocated
```

Assuming I did everything right, and I'm not confounding anything, that looks like a nice 3%-ish improvement :)

If anyone wants to try to reproduce, my nixpkgs clone is checked out to the old-ish a523706d02434f38c36d2f20cfdd972074ce8e1e.

src/libutil/types.hh
```diff
@@ -14,6 +14,7 @@ using std::list;
 using std::set;
 using std::vector;
 using std::string;
+using std::string_view;
```
edolstra, line 17: Let's not add more of these. In fact I'd rather get rid of the aliases for {set, vector, string}.

src/libutil/util.hh
```diff
@@ -368,22 +369,33 @@ MakeError(FormatError, Error);
...
 /* Concatenate the given strings with a separator between the
    elements. */
 template<class C>
-string concatStringsSep(const string & sep, const C & ss)
+string concatStringsSep(const string_view sep, const C & ss)
 {
     string s;
+    s.reserve(
+        std::accumulate(std::begin(ss), std::end(ss), string::size_type(0),
+            [] (auto size, string_view s) { return size + s.size(); }));
```
edolstra, line 383: Call me old-fashioned but I don't really like these STL algorithms like `std::accumulate` since they're verbose and non-functional (e.g. you have to mention `ss` multiple times, so you can't use it on an arbitrary expression). I prefer just
```
size_t size = 0;
for (auto & s : ss) size += s.size();
```

src/libutil/types.hh
pennae: we'll remove it. should `PathView` also go away to match, or would that be an exception?

edolstra: `PathView` is okay though maybe we should get rid of `Path` eventually, since it doesn't really serve a purpose.

src/libutil/util.hh
pennae: yeah, they are the opposite of pretty. at least C++20 has ranges, but even then `accumulate` is still pretty ugly :/

src/libexpr/primops.cc
```diff
@@ -2271,18 +2271,22 @@ static void prim_removeAttrs(EvalState & state, const Pos & pos, Value * * args,
...
-    for (auto elem : args[1]->listItems()) {
+    Symbol names[args[1]->listSize()];
+    for (auto [idx, elem] : enumerate(args[1]->listItems())) {
         state.forceStringNoCtx(*elem, pos);
-        names.insert(state.symbols.create(elem->string.s));
+        names[idx] = state.symbols.create(elem->string.s);
     }
 
+    auto it = &names[0], end = &names[args[1]->listSize()];
+    std::sort(it, end);
```
thufschmitt, line 2281: You’re not really `O(r+n)` (like f17684cca5c6c8afa469af9ad38026410de24da2 says) I assume since you have to sort. More `O(r*log(r)+n)`. Still faster than the existing implementation though.

pennae:
> If during the development you encountered any bug that feels like it should have been caught by the libutil testsuite

the only "bug" we've found is that util's `concatStringsSep` does not insert separators between leading empty strings, but at this point that probably have to be treated as part of its specification 😐

src/libexpr/primops.cc
pennae: yeah, the `r` there refers to the sorted set, not the unsorted input list. :sweat_smile:

src/libexpr/primops.cc
```diff
@@ -2271,18 +2271,22 @@ static void prim_removeAttrs(EvalState & state, const Pos & pos, Value * * args,
...
+
     /* Copy all attributes not in that set.  Note that we don't need
        to sort v.attrs because it's a subset of an already sorted
        vector. */
     auto attrs = state.buildBindings(args[0]->attrs->size());
     for (auto & i : *args[0]->attrs) {
-        if (!names.count(i.name))
+        while (it != end && *it < i.name) it++;
+        if (it == end || *it != i.name)
```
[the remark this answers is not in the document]
pennae: would prefer to use eg `std::set_difference` if that's also an option, although that requires modifications to `BindingsBuilder` so we can use `std::back_inserter(attrs)` as the target. (it's a small change, just adding a `push_back` alias for `insert`)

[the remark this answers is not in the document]
pennae: investigated this a bit and concluded that keeping the set is not worth it. the largest part of the cost the patch wants to remove from this function isn't in iterating over the set, it's in memory allocation. turning the set into a vector and reserving the required space is still more expensive than a stack array, but it does get us most of the way there.

perhaps this would be a good time to introduce a SmallVector wrapper into nix like llvm has, so we can store small amounts of stuff on the stack and large amounts on the heap without having to worry about stack overflows.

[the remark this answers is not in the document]
pennae: oh, neat! hadn't even checked boost yet, not having to build containers is nice :D looks like boost's small_vector is still slower than stack arrays, but not enough to not be worth using.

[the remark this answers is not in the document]
thufschmitt: I feared for a few moments that this would bring boost in the runtime closure of Nix (undoing the hack at https://github.com/NixOS/nix/blob/master/flake.nix#L304-L311), but looks like it doesn’t.

I’ve opened https://github.com/NixOS/nix/pull/5915 to enforce this in the future just in case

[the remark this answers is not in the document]
pennae: huh, hadn't even seen that hack yet. good catch!

PR #5915, opened 2022-01-14
Forbid runtime references to boost

thufschmitt:
We explicitly hack around to remove them, so might as well check that the hack is useful.

(Introduced because I feared that the changes of https://github.com/NixOS/nix/pull/5906#discussion_r784810238 would bring back some runtime references)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -312,6 +312,13 @@
                 chmod u+w $out/lib/*.so.*
                 patchelf --set-rpath $out/lib:${currentStdenv.cc.cc.lib}/lib $out/lib/libboost_thread.so.*
               ''}
+              ${lib.optionalString currentStdenv.isDarwin ''
+                for LIB in $out/lib/*.dylib; do
+                  chmod u+w $LIB
+                  install_name_tool -id $LIB $LIB
+                done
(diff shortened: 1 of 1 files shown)


thufschmitt:
Looks like this check isn’t totally spurious since it’s failing on OSX :D

I couldn’t dig this too deep since I don’t have a mac handy right now, but it looks like the boost libraries (that we selectively copy at build-time from the boost derivation to avoid having to pull in the whole library) have a self-reference which cause the dependency.

```
$ strings /nix/store/zkh34cv13gm60rl48g2xbq06nzdq31k3-nix-2.6.0pre20220113_a2811f8/lib/libboost_context.dylib | grep /nix/store/q8pcnpxg8i6ls44c3fxcn26pbxw5smxi-boost-1.69.0                          
/nix/store/q8pcnpxg8i6ls44c3fxcn26pbxw5smxi-boost-1.69.0/lib/libboost_context.dylib
```

I guess the right call of `install_name_tool` should fix it

Mic92:
@thufschmitt ping me on matrix and I can give you shell + root access on a m1 mac.

thufschmitt:
Thx @Mic92 . I found an old one in the meantime, but I’ll keep that in mind for a next time :)

PR #5916, opened 2022-01-14
Add git to the docker image

### LINKED ISSUE #5896: docker: add git to the nix image
This is just the same as https://github.com/NixOS/docker/issues/33 ; notably this was finally added in the original repository but was then subsequently lost again when that was made deprecated.

thufschmitt:
Fix #5896

See https://github.com/NixOS/docker/issues/33

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -21,6 +21,7 @@ let
     cacert.out
     findutils
     iana-etc
+    git
   ];
 
   users = {


garbas (approved):

Fuuzetsu:
Depending on impact to the image size, `gitMinimal` might be a better choice: I _think_ that's all that's needed for `nix` itself to work.

PR #5917, opened 2022-01-14
Make ProgressBar print build logs when TTY is not detected

### LINKED ISSUE #4402: Non-interactive progress indication in Nix 2
With `nix-command` enabled, running Nix in non-interactive environments (without a tty, only stdout, like in CI/CD environments) doesn't show nearly anything as it's all being hidden by the recent work to try and create an interactive progress bar. Even with `-L -v`, the only two options that seem relevant, there is nearly no output compared to `nix-build` and other Nix 1 commands. The only alternative seems to be to use `-vv` or `-vvv` which however show too much information, compared to Nix 1, especially evaluation progress.

I would expect there to be a mechanism for detecting interactive/non-interactive output (tty/file), and for the output to be very close to the original `nix-build` (`copying path...`, `building...`) if the output is non-interactive. Without a way to recover the old behavior which was informative in both interactive and non-interactive environments, the new output format seems like a step backwards.

armeenm:
Resolves #4402 until #4296 is finished. Nix will now behave as if `--log-format bar-with-logs` has been passed when `shouldANSI()` is false (i.e. stderr is not a TTY, `TERM` is `dumb` or unset, or `NO_COLOR` is set).

### DIFF
--- a/src/libmain/progress-bar.cc
+++ b/src/libmain/progress-bar.cc
@@ -85,7 +85,7 @@ private:
 public:
 
     ProgressBar(bool printBuildLogs, bool isTTY)
-        : printBuildLogs(printBuildLogs)
+        : printBuildLogs(printBuildLogs || !isTTY)
         , isTTY(isTTY)
     {
         state_.lock()->active = isTTY;


edolstra:
Let's close this following the discussion in #5918 (it's a bit unexpected behaviour that redirecting stderr causes a program to start generating a lot more output).

PR #5918, opened 2022-01-14
Print build logs when not connected to a TTY

balsoft:
When stderr is not connected to a tty, print out the build logs
instead of showing a bar.

Closes https://github.com/NixOS/nix/issues/4402

### DIFF
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -270,11 +270,15 @@ void mainWrapped(int argc, char * * argv)
         if (legacy) return legacy(argc, argv);
     }
 
-    verbosity = lvlNotice;
-    settings.verboseBuild = false;
     evalSettings.pureEval = true;
 
     setLogFormat("bar");
+    settings.verboseBuild = false;
+    if (isatty(STDERR_FILENO)) {
+        verbosity = lvlNotice;
+    } else {
+        verbosity = lvlInfo;
+    }
 
     Finally f([] { logger->stop(); });
 


balsoft:
Made during Nix UX hacking extravaganza

CC @tomberek

balsoft:
From meeting with Eelco:

- UNIX philosophy says we should be quiet when we have nothing important to say
  + But on the other hands some other build system-like tools, e.g. make, cabal and opam are not quiet at all
  + Some tools (like dune) also don't care about being interactive/non-interactive and always show the "interactive" output (but this is annoying and even breaks `nix log` sometimes)
- It may be surprising that when stderr redirected, we get a lot more output
  + On the other hand, when stderr is not redirected we get output shown as part of the bar, one line at a time. Maybe it makes sense that non-interactive output is basically the sum of all such lines shown in the interactive output?

Some discussion should follow

src/nix/main.cc
```diff
@@ -270,11 +270,17 @@ void mainWrapped(int argc, char * * argv)
...
-    setLogFormat("bar");
+    if (isatty(STDERR_FILENO)) {
+        verbosity = lvlNotice;
+        settings.verboseBuild = false;
+        setLogFormat("bar");
+    } else {
+        verbosity = lvlInfo;
+        settings.verboseBuild = true;
+        setLogFormat("raw-with-logs");
+    }
```
thufschmitt, line 283: Printing the build logs by default is quite verbose. I think it should be opt-in (`-L`) rather than opt-out

```suggestion
    setLogFormat("raw");
    settings.verboseBuild = false;
    if (isatty(STDERR_FILENO)) {
        verbosity = lvlNotice;
    } else {
        verbosity = lvlInfo;
    }
```

edolstra: :+1: but this code disables the progress bar in all cases which is probably not what you want?

thufschmitt: > this code disables the progress bar in all cases which is probably not what you want?

Uh yes I meant `bar` not `raw`, thanks. Fixed

ParetoOptimalDev:
Should this work from something like emacs `async-shell-command`?

I seem to still not get output with nix 2.7 using something like:

```
(async-shell-command "nixos-rebuild switch")
```

erikarvstedt:
`async-shell-command` attaches stderr to a TTY:
```
(async-shell-command "[[ -t 2 ]] && echo stderr is a tty") #=> stderr is a tty
```
Use this instead:
```
(async-shell-command "nixos-rebuild switch |& cat")
```

pshirshov:
I'm running nixos-rebuild under bash in a terminal and cannot get rid of the build logs.
Could you add an option to explicitly suppress them? (At the moment there is only an option to _enable_ the logs)

balsoft:
This PR ended up _not_ enabling log printing. What you're experiencing is likely a bug. Please, open an issue for it.

PR #5922, opened 2022-01-16
Add try/catch to queryJSON for assertion and errors

### LINKED ISSUE #5921: nix-env should do best effort to evaluate rather than failing
**Describe the bug**
I am seeing that `nix-env` is failing to completely evaluate `nixpkgs` when given `--json --out-path` due to `NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1`

You should be able to evaluate the whole `nixpkgs` or at least give a flag to `nix-env` to allow it to "best-effort" continue evaluation.

**Steps To Reproduce**

```
❯ NIXPKGS_ALLOW_BROKEN=1 NIXPKGS_ALLOW_INSECURE=1 NIXPKGS_ALLOW_UNFREE=1 ./nix/result/bin/nix-env -qaP --json --out-path > /tmp/out
warning: name collision in input Nix expressions, skipping '/home/fzakaria/.nix-defexpr/channels_root/nixpkgs'
error: Package ‘linux-4.14.165-172’ in /nix/store/gjjzmwrj1irv4hh1dv8l0c9iz4a3pbr0-nixpkgs-21.11pre332093.7fad01d9d5a/nixpkgs/pkgs/os-specific/linux/kernel/linux-hardkernel-4.14.nix:4 is not supported on ‘x86_64-linux’, refusing to evaluate.

       a) To temporarily allow packages that are unsupported for this system, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowUnsupportedSystem = true; }
       in configuration.nix to override this.

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowUnsupportedSystem = true; }
       to ~/.config/nixpkgs/config.nix.

```

**NOTE:** Adding `NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1` does not help because then you get stuck at an assert
```
❯ NIXPKGS_ALLOW_BROKEN=1 NIXPKGS_ALLOW_INSECURE=1 NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1 NIXPKGS_ALLOW_UNFREE=1 ./nix/result/bin/nix-env -qaP --json --out-path > /tmp/out
warning: name collision in input Nix expressions, skipping '/home/fzakaria/.nix-defexpr/channels_root/nixpkgs'
error: assertion '((stdenv).targetPlatform != (stdenv).hostPlatform)' failed

       at /nix/store/gjjzmwrj1irv4hh1dv8l0c9iz4a3pbr0-nixpkgs-21.11pre332093.7fad01d9d5a/nixpkgs/pkgs/top-level/all-packages.nix:11663:25:

        11662|   # built with, and use, that cross-compiled libc.
        11663|   gccCrossStageStatic = assert stdenv.targetPlatform != stdenv.hostPlatform; let
             |                         ^
        11664|     libcCross1 = binutilsNoLibc.libc;
```

**Expected behavior**

I would expect to generate as much JSON as possible for everything that can evaluate.

```
nix-env --version
nix-env (Nix) 2.6.0pre20220114_34f2eeb
```

fzakaria:
The table output for `nix-env -qaP` has a _try/catch_ for assertion & errors when parsing each individual derivation.
This functionality is missing for `queryJSON`.

I've added a similar `try/catch`.
**Note:** Perhaps ideally this should be a flag to control how "strict" `nix-env` should be.

I am finding that maintaing parity between the different output formats (_JSON_, _XML_ & _Tabular_) is very difficult.
I think a good issue and contribution would be the creation of a single struct that is then serialized to all 3 agnostically.

fixes #5921

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,7 @@ GTAGS
 compile_commands.json
 
 nix-rust/target
+
+result
+
+.vscode/
--- a/src/nix-env/nix-env.cc
+++ b/src/nix-env/nix-env.cc
@@ -911,36 +911,44 @@ static void queryJSON(Globals & globals, vector<DrvInfo> & elems, bool printOutP
 {
     JSONObject topObj(cout, true);
     for (auto & i : elems) {
-        JSONObject pkgObj = topObj.object(i.attrPath);
+        try {
+            if (i.hasFailed()) continue;
 
-        auto drvName = DrvName(i.queryName());
-        pkgObj.attr("name", drvName.fullName);
-        pkgObj.attr("pname", drvName.name);
-        pkgObj.attr("version", drvName.version);
-        pkgObj.attr("system", i.querySystem());
+            JSONObject pkgObj = topObj.object(i.attrPath);
 
-        if (printOutPath) {
-            DrvInfo::Outputs outputs = i.queryOutputs();
-            JSONObject outputObj = pkgObj.object("outputs");
-            for (auto & j : outputs) {
-                outputObj.attr(j.first, j.second);
-            }
-        }
+            auto drvName = DrvName(i.queryName());
+            pkgObj.attr("name", drvName.fullName);
+            pkgObj.attr("pname", drvName.name);
+            pkgObj.attr("version", drvName.version);
+            pkgObj.attr("system", i.querySystem());
 
-        if (printMeta) {
-            JSONObject metaObj = pkgObj.object("meta");
-            StringSet metaNames = i.queryMetaNames();
-            for (auto & j : metaNames) {
-                auto placeholder = metaObj.placeholder(j);
-                Value * v = i.queryMeta(j);
-                if (!v) {
-                    printError("derivation '%s' has invalid meta attribute '%s'", i.queryName(), j);
-                    placeholder.write(nullptr);
-                } else {
-                    PathSet context;
-                    printValueAsJSON(*globals.state, true, *v, noPos, placeholder, context);
+            if (printOutPath) {
+                DrvInfo::Outputs outputs = i.queryOutputs();
+                JSONObject outputObj = pkgObj.object("outputs");
+                for (auto & j : outputs) {
+                    outputObj.attr(j.first, j.second);
                 }
             }
+
+            if (printMeta) {
+                JSONObject metaObj = pkgObj.object("meta");
+                StringSet metaNames = i.queryMetaNames();
+                for (auto & j : metaNames) {
+                    auto placeholder = metaObj.placeholder(j);
+                    Value * v = i.queryMeta(j);
+                    if (!v) {
+                        printError("derivation '%s' has invalid meta attribute '%s'", i.queryName(), j);
+                        placeholder.write(nullptr);
+                    } else {
+                        PathSet context;
+                        printValueAsJSON(*globals.state, true, *v, noPos, placeholder, context);
+                    }
+                }
+            }
+        } catch (AssertionError & e) {
+            printMsg(lvlTalkative, "skipping derivation named '%1%' which gives an assertion failure", i.queryName());
+        } catch (Error & e) {
+            printMsg(lvlError, "skipping derivation named '%1%' which gives an error '%2%'", i.queryName(), e.msg());
         }
     }
 }
--- a/src/nix-env/nix-env.cc
+++ b/src/nix-env/nix-env.cc
@@ -948,7 +948,8 @@ static void queryJSON(Globals & globals, vector<DrvInfo> & elems, bool printOutP
         } catch (AssertionError & e) {
             printMsg(lvlTalkative, "skipping derivation named '%1%' which gives an assertion failure", i.queryName());
         } catch (Error & e) {
-            printMsg(lvlError, "skipping derivation named '%1%' which gives an error '%2%'", i.queryName(), e.msg());
+            e.addTrace(std::nullopt, "while querying the derivation named '%1%'", i.queryName());
+            throw;
         }
     }
 }


src/nix-env/nix-env.cc
```diff
@@ -911,36 +911,44 @@ static void queryJSON(Globals & globals, vector<DrvInfo> & elems, bool printOutP
...
-        pkgObj.attr("version", drvName.version);
-        pkgObj.attr("system", i.querySystem());
-
-        if (printOutPath) {
-            DrvInfo::Outputs outputs = i.queryOutputs();
-            JSONObject outputObj = pkgObj.object("outputs");
-            for (auto & j : outputs) {
-                outputObj.attr(j.first, j.second);
+        try {
+            if (i.hasFailed()) continue;
```
fzakaria, line 915: I noticed this check for `hasFailed` also in the table output which I copied over.

It's a bit unfortunate all this functionality has to be copied for JSON/XML/Table...
If I was a bit more experienced in the codebase, I think creating a single intermediate representation and then having that serialize to XML/Table/JSON would be the best.

I'm sure that's pretty obvious assessment though :P

src/nix-env/nix-env.cc
```diff
@@ -911,36 +911,44 @@ static void queryJSON(Globals & globals, vector<DrvInfo> & elems, bool printOutP
...
+                    } else {
+                        PathSet context;
+                        printValueAsJSON(*globals.state, true, *v, noPos, placeholder, context);
+                    }
                 }
             }
+        } catch (AssertionError & e) {
+            printMsg(lvlTalkative, "skipping derivation named '%1%' which gives an assertion failure", i.queryName());
+        } catch (Error & e) {
+            printMsg(lvlError, "skipping derivation named '%1%' which gives an error '%2%'", i.queryName(), e.msg());
```
fzakaria, line 951: The table format `throws` here instead.

.gitignore
```diff
@@ -120,3 +120,7 @@ GTAGS
 compile_commands.json
 
 nix-rust/target
+
+result
```
fzakaria, line 124: Every nix project should have _result_ in it's _gitignore_ :dancers:

fzakaria:
FWIW I seem to bet `37403` packages from nixpkgs commit: `63980ec2c90fbe3c70e2960b6bc7318a456093b1`

```console
❯ pwd
/home/fzakaria/code/github.com/NixOS/nixpkgs

❯ ../nix/result/bin/nix-env -qaP --json --out-path -f $(pwd) > /tmp/out

❯ cat /tmp/out | jq length
37403
```

fzakaria:
I think there are more bugs here; the derivations that are parsed based on the output do not look the same and occasionally I get an _abort_.

```console
❯ ../nix/result/bin/nix-env -qaP --json --out-path -f $(pwd) > /tmp/out
skipping derivation named 'lisp-djula-20211209-git' which gives an error 'error: value is a string with context while a list was expected

       at /home/fzakaria/code/github.com/NixOS/nixpkgs/pkgs/stdenv/generic/make-derivation.nix:262:22:

          261|          else                                     configureFlags)
          262|         ++ optional (elem "build"  configurePlatforms) "--build=${stdenv.buildPlatform.config}"
             |                      ^
          263|         ++ optional (elem "host"   configurePlatforms) "--host=${stdenv.hostPlatform.config}"'
skipping derivation named 'lisp-parser-combinators-cl-20131111-git' which gives an error 'error: value is a string with context while a list was expected

       at /home/fzakaria/code/github.com/NixOS/nixpkgs/pkgs/stdenv/generic/make-derivation.nix:262:22:

          261|          else                                     configureFlags)
          262|         ++ optional (elem "build"  configurePlatforms) "--build=${stdenv.buildPlatform.config}"
             |                      ^
          263|         ++ optional (elem "host"   configurePlatforms) "--host=${stdenv.hostPlatform.config}"'
skipping derivation named 'pew-1.2.0' which gives an error 'error: value is the built-in function 'filter' while a set was expected'


❯ ../nix/result/bin/nix-env -qaP --json --out-path -f $(pwd) > /tmp/out
[1]    523453 abort      ../nix/result/bin/nix-env -qaP --json --out-path -f $(pwd) > /tmp/out

❯ ../nix/result/bin/nix-env -qaP --json --out-path -f $(pwd) > /tmp/out
[1]    523808 abort      ../nix/result/bin/nix-env -qaP --json --out-path -f $(pwd) > /tmp/out

❯ ../nix/result/bin/nix-env -qaP --json --out-path -f $(pwd) > /tmp/out
skipping derivation named 'lisp-djula-20211209-git' which gives an error 'error: attribute 'baseName' missing

       at /home/fzakaria/code/github.com/NixOS/nixpkgs/pkgs/development/lisp-modules/define-package.nix:13:53:

           12|     outhash="''${outhash%%-*}"
           13|     config_script="$out"/lib/common-lisp-settings/${args.baseName}-shell-config.sh
             |                                                     ^
           14|     path_config_script="$out"/lib/common-lisp-settings/${args.baseName}-path-config.sh'
skipping derivation named 'lisp-parseq-20210531-git' which gives an error 'error: value is a list while a set was expected'
skipping derivation named 'lisp-parser-combinators-cl-20131111-git' which gives an error 'error: attribute 'baseName' missing

       at /home/fzakaria/code/github.com/NixOS/nixpkgs/pkgs/development/lisp-modules/define-package.nix:13:53:

           12|     outhash="''${outhash%%-*}"
           13|     config_script="$out"/lib/common-lisp-settings/${args.baseName}-shell-config.sh
             |                                                     ^
           14|     path_config_script="$out"/lib/common-lisp-settings/${args.baseName}-path-config.sh'
```

Might be tough to notice but the first invocation finds errors in: `lisp-djula-20211209-git`, `lisp-parser-combinators-cl-20131111-git` and `pew-1.2.0`.

The second invocation finds errors in: `lisp-djula-20211209-git`, `lisp-parser-combinators-cl-20131111-git` and `lisp-parseq-20210531-git`.

`lisp-djula-20211209-git` and `lisp-parseq-20210531-git` are respectively missing.

src/nix-env/nix-env.cc
edolstra: Yeah we should not ignore arbitrary errors.

edolstra:
@fzakaria I think the other errors were due to a GC bug that has now been fixed (#5964).

src/nix-env/nix-env.cc
fzakaria: Fixed in latest commit.

ncfavier:
This change goes in the right direction, but is not sufficient to evaluate all of nixpkgs with `pkgs/top-level/packages-config.nix`:

```
$ nix-env -f '<nixpkgs>' --arg config 'import <nixpkgs/pkgs/top-level/packages-config.nix>' -qa --json --meta --out-path
error: attribute 'mesos' missing

       at /home/n/git/nixpkgs/pkgs/development/haskell-modules/configuration-nix.nix:143:24:

          142|     # Pass _only_ mesos; the correct protobuf is propagated.
          143|     extraLibraries = [ pkgs.mesos ];
             |                        ^
          144|     preConfigure = "sed -i -e /extra-lib-dirs/d -e 's|, /usr/include, /usr/local/include/mesos||' hs-mesos.cabal";
```

This could of course be fixed in nixpkgs, but I wonder if it would be reasonable not to `throw` on evaluation errors? Maybe with a flag `--keep-going`?

FYI, this is relevant for nixos-search because we plan on supporting listing package outputs, which requires evaluation.

fzakaria:
I agree with you.

Adding a suppression for errors might be nice.

edolstra:
Not catching missing-attribute errors is intentional. Those are bugs in Nixpkgs that should be fixed, we don't want to quietly ignore them in nix-env.

ncfavier:
I see, so `nixpkgs-basic-release-checks.nix` should probably use `pkgs/top-level/packages-config.nix` to recurse into package sets and find errors there?

PR #5930, opened 2022-01-17
Document the concept of “experimental feature”

thufschmitt:
Not sure the Nix manual is the best place to put this, but I couldn’t think of any other.

This is just one take on it, and it’s obviously open for debate


doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,89 @@
...
+anything else that's considered as "public").
+
+For example, these changes need to be guarded by an experimental flag:
+
+-   Any language change (new builtin, syntactic or semantic change,
+    etc..),
+
+-   Any change to the (non-experimental) CLI interface
+
+-   A non-trivial internal refactoring that's likely to introduce some
```
edolstra, line 29: By their nature, I don't think it's feasible to make refactorings experimental. Because that would mean you have to keep the old and new code around, which goes against the idea of a refactoring. Also, refactorings shouldn't break anything by definition.

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,89 @@
...
+    bugs\[\^1\]
+
+\[\^1\]: Having to be weighted against the costs of course. If the
+refactoring changes a low-level interface and keeping both versions
+around essentially means duplicating half of the codebase, then it might
+not be worth doing that.
+
+# Life flow of an experimental flag
```
edolstra, line 37:
```suggestion
# Lifecycle of an experimental feature
```

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,89 @@
+This section aims at describing the notion of a Nix "experimental
+flag", and how it fits into the big picture of the developpement of Nix.
```
edolstra, line 2:
```suggestion
This section describes the notion of a Nix's "experimental
features", and how it fits into the big picture of the development of Nix.
```

And rename "experimental/xp flags" to "experimental features" throughout, because the config setting is `experimental-features` after all.

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,89 @@
...
+# What are these xp flags?
+
+Since Nix 2.4, some features of Nix can be flagged as
+experimental.Conceptually, this means that these features are not
+considered as part of Nix proper, but are only included in the code to
+make their developpement and testing easier.Practically, these require
+an explicit opt-in, by adding the name of the experimental feature to
+the `experimental-features` configuration option. This is really handy
+as it makes it harder to accidentally use an experimental feature
+without knowing it.
```
edolstra, line 13:
```suggestion
Since Nix 2.4, some features of Nix can be flagged as
experimental. This means that they can be changed or removed at any time.
Users must explicitly opt in to them by adding the name of the experimental feature to
the `experimental-features` configuration option. This is really handy
as it makes it harder to accidentally use an experimental feature
without knowing it.
```

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,89 @@
...
+
+# When should an experimental flag be used
+
+A change in the Nix codebase should be guarded by an experimental flag
+if it is likely to either introduce a regression of any kind, or if it
+changes the external interface of Nix (be it the language, the cli or
+anything else that's considered as "public").
```
edolstra, line 20: I don't think everything that changes the external interface requires an experimental feature. The main criterion is whether we're uncertain about whether the feature is a good idea at all, or about the exact interface, or the implementation is not sufficiently stable yet. But we wouldn't make (for instance) a new builtin function experimental unless we have doubts about it.

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,89 @@
...
+-   After at least a full 6w release cycle, the feature becomes enabled
+    by default
+
+-   After at lesat another full 6w release cycle, the old behavior is
+    removed
+
+This is obviously just indicative. In particular the third step is only
+useful if the feature is susceptible to introduce some significant and
+hard to work-around regression (for example the addition of a new
+language builtin). Conversely, the second is useless if people have no
```
edolstra, line 57: This is unclear, because why would the addition of a new builtin cause a regression?

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,89 @@
...
+compelling reason to switch the flag (for example a purely internal
+refactoring)
+
+# Relation to the RFC process
```
edolstra, line 61: I would leave out this section, since the relationship between RFCs and experimental features is itself something that should be clarified in a future RFC.

doc/manual/src/contributing/experimental-flags.md
thufschmitt: Fair point. What about:

```suggestion
A change in the Nix codebase should be guarded by an experimental flag
if it is likely to either introduce a regression of any kind, or if it
implies a non-trivial change in the external interface of Nix (be it the language, the cli or
anything else that's considered as "public").
```
?

(For the specific “builtin” case, I’d actually be in favor of making them experimental because that doesn’t cost much and can avoid having to commit to them forever if they prove not to be that useful. But that’s not central)

doc/manual/src/contributing/experimental-flags.md
thufschmitt: > shouldn’t

That’s where the devil lies :stuck_out_tongue: 

I guess it depends how you define “refactoring”. Something I had in mind in particular here is changes to the daemon protocol (say adding a new message type to make some specific operation faster), which _shouldn’t_ break anything, but happen to do so every once in a while (and for which we need to keep both versions anyways for backwards compatibility). Or just a re-implementation of a specific piece of functionality that’s well-contained (so cheap to duplicate for a short amount of time), but might be hiding some subtle bug

doc/manual/src/contributing/experimental-flags.md
thufschmitt: Indeed. I think I first wrote the sentence the other way, so that was actually an exemple of something that’s unlikely to warrant this third step :upside_down_face:

doc/manual/src/contributing/experimental-flags.md
thufschmitt: I’ve some mixed feelings about that.

That admittedly falls in a grey area. This whole “experimental features” thing is technically an implementation detail, so something that should just be of the resort of the Nix development team. But at the same time there’s no such team (since the “core team” got disbanded).

But in that matter you’re the doc, doc. So if you feel like it, I’ll re-submit the whole thing as an RFC

doc/manual/src/contributing/experimental-flags.md
thufschmitt, line 57:
```suggestion
hard to work-around regression (for example if it requires some non-trivial changes in existing codepaths). Conversely, the second is useless if people have no
```

doc/manual/src/contributing/experimental-flags.md
thufschmitt: Yes, I renamed it to “experimental flag” because it doesn’t have to be a “feature” strictly speaking (because of the section on refactorings). But we can keep the “feature” word for consistency

(Or we could add a new `xp-feature-is-flag` experimental feature to rename the option to `experimental-flags`. But that’s going a bit too meta :wink: )

Ericson2314 (approved):

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,68 @@
+This section describes the notion of *experimental features*, and how it fits into the big picture of the development of Nix.
+
+# What are experimental features?
+
+[Since Nix 2.4](../release-notes/rl-2.4.md), some features of Nix can be flagged as experimental.
+Experimental features can be changed or removed at any time, and users must [explicitly enable them].
+This allows accessing new functionality that is not yet stable without unwittingly relying on them.
+
+[explicitly enable them]: ../nix-conf.md#conf-expeirmental-features
```
edolstra, line 9:
```suggestion
[explicitly enable them]: ../nix-conf.md#conf-experimental-features
```

edolstra: BTW I'm really opposed to the "footnote" style of linking. It requires duplicating text like "explicitly enable them", meaning that if you change that text, you have to remember to update it in two places and not make a typo. It's much less verbose and error-prone to just write:

> ... and users must [explicitly enable them](../nix-conf.md#conf-experimental-features).

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,68 @@
...
+
+# When should a new feature be marked experimental?
+
+A change in the Nix codebase should be guarded by an experimental feature flag if it is likely to either introduce a regression of any kind, or if it implies a non-trivial change to the external interface of Nix.
```
edolstra, line 13: I would drop the "introduce a regression of any kind". Any change can introduce a regression, but that's a matter of testing.

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,68 @@
...
+
+Examples:
+
+- Changes to the Nix language, such as new built-ins, syntactic or semantic changes, etc.
+- Changes to the command-line interface
+- Some non-trivial internal changes that are likely to introduce bugs
```
edolstra, line 19:
```suggestion
```

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,68 @@
...
+
+# Lifecycle of an experimental feature
+
+Experimental features have to be treated on a case-by-case basis.
+However, the standard workflow for an experimental feature is as follows:
+
+- The feature is implemented in a pull request.
+- It is merged under an experimental flag, and is disabled by default.
+    - Being experimental, the feature can still be changed arbitrarily or removed again.
+- After at least one release cycle without any significant issues, the feature becomes enabled by default.
```
edolstra, line 29: This does not reflect our current process. We don't enable experimental features ever, that's for the user to do.

fricklerhandwerk: That's true. Then we'd need to specify when to simply make a feature not experimental any more.

Just an idea:

> When maintainers are confident that a feature will not require breaking changes in the future, the experimental flag is removed and the feature considered stable.

Ericson2314: I like the "cool down" period aspect, which is when we are ready to stabilize something we should make sure it hasn't changed since the last release. That means we might focus on something for stabilization, do some "finishing touches", and then wait a cycle.

tomberek: For that stabilization cycle it is is more of a "feature flag" than an "experimental flag". This would make the feature opt-out vs opt-in.

thufschmitt: Yeah, calling it “experimental” is wrong here. I think it's a worthwile phase to have for some changes though, where we really want to flip the switch but can't be sure that we won't break something. But maybe we can save it for later

doc/manual/src/contributing/experimental-flags.md
timfenney: Is it the intent that non-trivial API changes will be rolled out first as experimental features?

Ericson2314:
I would call it "experimental features" not "experimental flags" simply because I have the PR that allows making flags gated on experimental features, and I would think such flags would be called experimental flags not the experimental features themselves :).

thufschmitt:
> I would call it "experimental features" not "experimental flags"

I'm not fond of “experimental features” because they don't have to be “features” per se (`no-url-literals` for example). But maybe it's best to just stick to the established wording any way.

Ericson2314:
That's a good reason; I am open to other alternatives too --- just want to avoid the flags issue.

What about just "experiments"? :)

fricklerhandwerk:
experimental
- setting
- option
- behavior

Although I actually like "experiment".

Ericson2314:
In additional to allowing flags to be marked experimental, I also made a PR to allow settings/options to be marked experimental :).

doc/manual/src/contributing/experimental-flags.md
fricklerhandwerk:
```suggestion
A change in the Nix codebase should be guarded by an experimental feature flag if it implies a backwards-incompatible or non-trivial change to the external interface of Nix.
```

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,67 @@
...
+    - If any blocking issue gets reported the feature can be disabled by default again.
+- After at least another release cycle, the feature flag is removed and the feature is considered stable.
```
fricklerhandwerk, line 30:
```suggestion
- The experimental feature flag can be removed and the feature is considered stable if:
  - There is enough evidence of users having tried the feature, such as feedback, fixed bugs, demonstrations of how it is put to use
  - Maintainers feel confident that
    - The feature is designed and implemented sensibly, that it is fit for purpose
    - Potential interactions are well-understood
    - Stabilising the feature will not incur an outsized maintenance burden in the future
```

doc/manual/src/contributing/experimental-flags.md
```diff
@@ -0,0 +1,67 @@
...
+    trunk([pull request])
+    dropped([removed])
+
+
+    trunk --> |merge into master| xp_disabled
+    xp_disabled --> |At least one release without significant issues| xp_enabled
+    xp_enabled --> |At least one release without significant issues| stable
+    xp_enabled --> |Blocking issue reported| xp_disabled
+    xp_disabled --> |"(breaking) changes"| xp_disabled
+    xp_disabled --> |Decision against adding the feature| dropped
```
fricklerhandwerk, line 50: Update the diagram to match the text description

fricklerhandwerk (dismissed):
Reviewed in the Nix team meeting:

- fix up the comments
- rename the file to `experimental-features.md`

doc/manual/src/contributing/experimental-features.md
```diff
@@ -0,0 +1,59 @@
+This section describes the notion of *experimental features*, and how it fits into the big picture of the development of Nix.
+
+# What are experimental features?
+
+[Since Nix 2.4](../release-notes/rl-2.4.md), some features of Nix can be flagged as experimental.
+Experimental features can be changed or removed at any time, and users must [explicitly enable them](../nix-conf.md#conf-expeirmental-features).
```
thufschmitt, line 6:
```suggestion
Experimental features can be changed or removed at any time, and users must [explicitly enable them](../nix-conf.md#conf-experimental-features).
```

doc/manual/src/contributing/experimental-features.md
```diff
@@ -0,0 +1,59 @@
...
+
+- The feature is implemented in a pull request.
+- The feature is guarded by an experimental feature flag that is disabled by default, and the pull request is merged.
+    - Being experimental, the feature can still be changed arbitrarily or removed again.
+- The experimental feature flag can be removed and the feature is considered stable if:
+  - There is enough evidence of users having tried the feature, such as feedback, fixed bugs, demonstrations of how it is put to use
+  - Maintainers feel confident that
+    - The feature is designed and implemented sensibly, that it is fit for purpose
+    - Potential interactions are well-understood
+    - Stabilising the feature will not incur an outsized maintenance burden in the future
```
Ericson2314, line 31:
```suggestion
- The feature is implemented in a pull request, guarded by an experimental feature flag that is disabled by default, and the pull request is merged.
- The feature is *experimented* with by interested parties
    - Those that wish to experiment with the feature must explicitly enable it, signifying they are aware of the risk of using unstable ideas and are consenting to them.
    - Being experimental, the feature can still be changed arbitrarily.
- If the experience with the feature is negative, the feature is *removed*
  - The feature itself is removed, and the named experimental feature flag is also removed along with it.
- If the experience with the feature is positive, the feature is *stabilized*
  - The feature itself is kept, and just the experimental feature flag itself is removed, ungating the experimental feature so it is always enabled.
  - To reach this step:
    - There must be is enough evidence of users having tried the feature, such as feedback, fixed bugs, demonstrations of how it is put to use
    - Maintainers must feel confident that
      - The feature is designed and implemented sensibly, that it is fit for purpose
      - Potential interactions are well-understood
      - Stabilising the feature will not incur an outsized maintenance burden in the future
```

doc/manual/src/contributing/experimental-features.md
```diff
@@ -0,0 +1,59 @@
...
+This allows accessing new functionality that is not yet stable without unwittingly relying on them.
+
+# When should a new feature be marked experimental?
+
+A change in the Nix codebase should be guarded by an experimental feature flag if it is considered likely to be reverted or adapted in a backwards-incompatible manner after gathering more experience with it in practice.
+
+Examples:
+
+- Changes to the Nix language, such as new built-ins, syntactic or semantic changes, etc.
+- Changes to the command-line interface
```
Ericson2314, line 16:
```suggestion
A change in the Nix codebase should be guarded by an experimental feature flag if it needs the Nix community to gather experience with it before we can decide what we want it to be.

For most changes, experimenting with the pull request branch before it is merged should be good enough --- if the Nix maintainers are not confident in the design or implementation, then the PR must not be merged. But for some larger, highly visible, or important changes, this will not work:

  - The Nix maintainers need input from community at large need to help with the decision. The Nix community at large needs an easy and safe way to experiment and gain experience with the idea, and trying out pull request miscellaneous branches is neither easy or safe enough.

  - A long period of experimentation and gaining experience is necessary. Keeping a PR open for that length of time would result in and undue burden of fixing conflicts. With just a PR, we also loose out on the opportunity for the change (especially if the maintainers are leaning towards accepting it) influencing other development and thus gaining experience with the maintained costs and benefits of the features.
  
Conversely, just merging the feature and hoping it was designed correctly the first time is also impractical:
 - If we don't want the feature after all, removing it is a breaking change and prohibited without a major version bump. 
 - Even if we do mostly like the feature it may still some breaking changes to improve its design, again prohibited without a major version bump.

What we thus need is a middle-ground between merging a feature which instantly becomes stable, and requesting further changes to the PR without merging it. That is where experimental features come in:

 - The feature is merged so it is easy to try it out and gain more experience
 - The feature is not stable so breaking changes to the feature including removal are completely permissible
 - The feature is *guarded* so regular uses will not use it by mistake, and thus are shielded from any instability.

Examples:

- Changes to the Nix language, such as new built-ins, syntactic or semantic changes, etc.
- Changes to the command-line interface
```

fricklerhandwerk: I think we've already captured that in [one sentence:](https://github.com/NixOS/nix/pull/5930/files#diff-c60361f52f0e428ab378856cac200ca405b6f43f71a1c78621c67208ca4e3391R53)
> An experimental feature enables developers to iterate on and deliver a new idea without committing to it or requiring a costly long-running fork.

doc/manual/src/contributing/experimental-features.md
fricklerhandwerk:
```suggestion
- A new feature is implemented in a *pull request*
  - It is guarded by an experimental feature flag that is disabled by default
- The pull request is merged, the *experimental* feature ends up in a release
    - Using the feature requires explicitly enabling it, signifying awareness of the potential risks
    - Being experimental, the feature can still be changed arbitrarily
- The feature can be *removed*
  - The associated experimental feature flag is also removed
- The the feature can be made *stable*
  - The associated experimental feature flag is removed
  - There should be enough evidence of users having tried the feature, such as feedback, fixed bugs, demonstrations of how it is put to use
  - Maintainers must feel confident that:
    - The feature is designed and implemented sensibly, that it is fit for purpose
    - Potential interactions are well-understood
    - Stabilising the feature will not incur an outsized maintenance burden in the future
```

doc/manual/src/contributing/experimental-features.md
```diff
@@ -0,0 +1,59 @@
...
+
+```mermaid
+flowchart
+  idea([idea]) --> |discussion, design, implementation| pr([pull request])
+  experimental --> |"user feedback, (breaking) changes"| pr
+
+  pr --> |review| pr
+  pr --> |merge| experimental
+  experimental --> |decision to stabilise| stable
+  experimental --> |decision against keeping the feature| removed
```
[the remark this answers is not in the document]
fricklerhandwerk: I disagree that we should add more words to the diagram. It's only for illustration, just to get an overview. Stuffing it with information makes it unreadable.

doc/manual/src/contributing/experimental-features.md
Ericson2314: That looks good to me!

doc/manual/src/contributing/experimental-features.md
```diff
@@ -0,0 +1,64 @@
...
+
+# What are experimental features?
+
+[Since Nix 2.4](../release-notes/rl-2.4.md), some features of Nix can be flagged as experimental.
+Experimental features can be changed or removed at any time, and users must [explicitly enable them](../nix-conf.md#conf-experimental-features).
+This allows accessing new functionality that is not yet stable without unwittingly relying on them.
+
+# When should a new feature be marked experimental?
+
+A change in the Nix codebase should be guarded by an experimental feature flag if it is considered likely to be reverted or adapted in a backwards-incompatible manner after gathering more experience with it in practice.
```
Ericson2314, line 11:
```suggestion
Experimental features are needed for high stakes changes to Nix that benefit from broad community feedback.

Most of the time, experimentation happens prior to merging PRs; Nix maintainers merge changes when we are confident in those changes.
Merging changes we are not confident and then regretting those decisions leads to either:
 - Breaking our stability guarantee
 - Sticking with a design we think is bad

But for larger or highly visible changes, merging PRs only after we have become confident is not practical:
  - The PR would have to be open for a very long period of time, past the point of being an acceptable burden to the PR author
  - The barrier to installing an unofficial non-release version of Nix means we still wouldn't get enough community input.

Experimental features are a middle ground which, via a little extra code, avoids all these bad outcomes.
```

doc/manual/src/contributing/experimental-features.md
```diff
@@ -0,0 +1,64 @@
...
+
+Examples:
```
Ericson2314, line 13:
```suggestion
Examples of larger and highly visible requiring experimental features:
```

Ericson2314: (It was fine before, but if we add my longer description above, I think it stops becoming clear what this is an example of :)

thufschmitt (commented):
I can't approve since the PR is mine originally, but :+1: on the current content.
@fricklerhandwerk are there still things you want to change on this?

fricklerhandwerk:
The last changes introduced a lot of movement and noise into the text that we carefully avoided with the previous draft, and I need some time to concentrate to see through it. This week is very busy due the docs team project application, so please excuse another delay.

In think this piece is important enough to take that time, as experimental features say a lot about the development process and have repeatedly been a topic of debate in the past.

thufschmitt:
@fricklerhandwerk still planning to come back to this?

fricklerhandwerk:
@thufschmitt we paired with @Ericson2314 yesterday to fix this up. The change against what you and me prepared is now minimal to make a few things more concrete. I think we can merge this as is, as the spirit has not changed since the team's approval.

@Ericson2314 wants to follow up with a "why exp features" section, a glossary entry, and more links to this page, and I'd support that with pairing or reviews.

Ericson2314 (approved):

PR #5929, opened 2022-01-17
Add command to copy build logs between stores

edolstra:
TODO: add tests, implement for `BinaryCacheStore` so Hydra can use this.

Fixes #5222.

PR #5932, opened 2022-01-17
Remove shared strings

edolstra:
This gets rid of `std::shared_ptr<std::string>` and `ref<std::string>`. These were needed back in the pre-C++11 era because we didn't have move semantics. But now we do.

Extracted/expanded from #5929.

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -31,7 +31,7 @@ BinaryCacheStore::BinaryCacheStore(const Params & params)
 
     StringSink sink;
     sink << narVersionMagic1;
-    narMagic = *sink.s;
+    narMagic = sink.s;
 }
 
 void BinaryCacheStore::init()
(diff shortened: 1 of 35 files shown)


src/libstore/binary-cache-store.cc
```diff
@@ -509,7 +509,7 @@ std::shared_ptr<std::string> BinaryCacheStore::getBuildLog(const StorePath & pat
 
     debug("fetching build log from binary cache '%s/%s'", getUri(), logPath);
 
-    return getFile(logPath);
+    return std::move(*getFile(logPath));
```
thufschmitt, line 512: Any reason why you can’t return `getFile(logPath)` since it’s already an `optional<string>`?

(also, wouldn’t this throw if the log isn’t available?)

edolstra: Thanks, that was a SNAFU.

PR #5941, opened 2022-01-19
Optimize intersectAttrs performance

roberth:
- Add test for `intersectAttrs`.
 - Don't worry about `intersectAttrs` performance.
 - Improve performance where not used with `functionArgs`; primarily `haskell-modules/configuration-nix.nix`.
 - Support a more efficient `recursiveUpdateUntil` implementation (nixpkgs PR and stats TBD)

I've documented the choice of algorithm in the inline comment. I figured that info would be relevant for further optimization years from now, but it's not important for the implementation.

Regarding the implementation, if you're wondering if it could be represented more succinctly with a "big" and "small" bindings pointer, that's a no, because that actually increases the number of branches as the second argument's attribute value that has to be inserted could come from either the loop or the `find`.

TODO: benchmark nixpkgs with:

<details><summary>recursiveUpdateUntil</summary>

```diff
-  recursiveUpdateUntil = pred: lhs: rhs:
-    let f = attrPath:
-      zipAttrsWith (n: values:
-        let here = attrPath ++ [n]; in
-        if tail values == []
-        || pred here (head (tail values)) (head values) then
-          head values
-        else
-          f here values
-      );
-    in f [] [rhs lhs];
+  recursiveUpdateUntil = pred:
+    let
+      go = path: lhs: rhs:
+        if pred path lhs rhs
+        then rhs
+        else lhs // (
+          rhs //
+            mapAttrs
+              (k: go (path ++ [k]) lhs.${k})
+              (intersectAttrs lhs rhs)
+        );
+    in go [];
```

</details>


roberth:
The current master is not stable, and I have to rebase onto it, because @pennae's recent changes are relevant.

I will postpone benchmarking the suggested `recursiveUpdate` change, but I'm confident that the change in this PR is good in itself. The benchmark would only have served as a reinforcement for the PR's value.

roberth:
I've run a benchmark, with no statistically significant result. It was significant before pennae rewrite `recursiveUpdateUntil` with their new `zipAttrsWith` builtin, fwiw. I guess I should've hurried ;)
Without a practical result, this change can be judged by its asymptotic improvement instead: making it roughly proportional to the smallest set instead of the first set.

src/libexpr/primops.cc
```diff
@@ -2448,12 +2448,62 @@ static void prim_intersectAttrs(EvalState & state, const PosIdx pos, Value * * a
...
+            if (r != right.end())
+                attrs.insert(*r);
+        }
+    }
+    else {
+        for (auto & r : right) {
+            Bindings::iterator l = left.find(r.name);
+            if (l != left.end())
+                attrs.insert(r);
+        }
```
thufschmitt, line 2506: Could we replace the duplication by something like

```cpp
if (left.size() > right.size()) {
    std::swap(left, right);
}
```

?

thufschmitt (approved):

src/libexpr/primops.cc
roberth: That would interfere with the `insert`, which must take `right`.

thufschmitt: Oh, right

Issue #5912, opened 2022-01-13
`nix why-depends --derivation` output is very noisy, a little hard to read


lilyball:
**Describe the bug**

`nix why-depends` shows the location of the store hash in each file along the path. This can be interesting for runtime closures, but for `--derivation` output it's just very noisy as every "file" along the path is reported as `/` and the contents are a compact derivation, thus causing confusing output where every line starts with the end of the previous entry in the derivation. Also longer paths can cause the relevant output to be truncated, e.g. the `texinfo-interactive-6.8` line in the following output cuts off the output name(s).

Showing the file as `/` is weird, and feels like an artifact of how `nix why-depends` shows file dependencies in general, but using the actual derivation filename here would be a waste of space, and of course the derivation is listed immediately above each entry on the parent `→` line. Though the root entry doesn't have a parent `→` line anyway (which means this output doesn't contain the actual derivation path for the `package` argument).

**Steps To Reproduce**

1. `nix why-depends -a --derivation self#darwinConfigurations.DeerBook.system self#darwinConfigurations.DeerBook.pkgs.perl`

**Expected behavior**

This should be reasonably readable. Instead it looks like

```
/nix/store/i6ykhhkw7m1ssm5swwp17gkxgaksfiwz-darwin-system-22.05.20220110.81f05d8+darwin4.15635ae.drv
├───/: …-etc.drv",["out"]),("/nix/store/9zfsj9h3kx7d5y7s1llckv7hmryliki2-system-path.drv",["out"]),("/ni…
│   → /nix/store/9zfsj9h3kx7d5y7s1llckv7hmryliki2-system-path.drv
│   ├───/: …uild.drv",["out"]),("/nix/store/hxbsnrlbhzdl9bzxxl3c7gv9bl3rjdkg-perl-5.34.0.drv",["out"]),("/ni…
│   │   → /nix/store/hxbsnrlbhzdl9bzxxl3c7gv9bl3rjdkg-perl-5.34.0.drv
│   ├───/: …info.drv",["out"]),("/nix/store/lv12dxlngxnjxsaim3ybshnabnj7pjwd-texinfo-interactive-6.8.drv",["…
│   │   → /nix/store/lv12dxlngxnjxsaim3ybshnabnj7pjwd-texinfo-interactive-6.8.drv
[...]

PR #5942, opened 2022-01-19
Make `nix why-depends` quieter by default

thufschmitt:
Unless `--precise` is passed, make `nix why-depends` only show the dependencies between the store paths, without introspecting them to find the actual references.

This also makes it ~3x faster

/cc @lilyball 

Fix https://github.com/NixOS/nix/issues/5912

### DIFF
--- a/src/nix/why-depends.cc
+++ b/src/nix/why-depends.cc
@@ -31,6 +31,7 @@ struct CmdWhyDepends : SourceExprCommand
 {
     std::string _package, _dependency;
     bool all = false;
+    bool precise = false;
 
     CmdWhyDepends()
     {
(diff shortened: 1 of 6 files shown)


src/nix/why-depends.cc
```diff
@@ -56,6 +57,12 @@ struct CmdWhyDepends : SourceExprCommand
             .description = "Show all edges in the dependency graph leading from *package* to *dependency*, rather than just a shortest path.",
             .handler = {&all, true},
         });
+
+        addFlag({
+            .longName = "precise",
+            .description = "For each edge of the graph, inspect the parent node to display the exact location in the path that causes the dependency",
```
edolstra, line 63:
```suggestion
            .description = "For each edge of the graph, show the files in the parent that cause the dependency.",
```

edolstra (approved):

src/nix/why-depends.cc
thufschmitt: Thanks, I found the old description quite crappy but didn’t know how to put it right :)

PR #5944, opened 2022-01-19
some more parser improvements

pennae:
removing more allocations from the parser and not interning every string the parser sees increases eval performance a good bit. refactoring the "duplicate arguments" check for lambdas adds another bit, and all together we get about 5% improvement out of it on eval, and about 10% on pure parsing.


kvtb:
While it improves parsing, it could degrade string comparison 
parsing is performed only once.
code evaluation might have recursive loops and string comparison done thousand times

pennae:
@kvtb only nix *expressions* used symbols for strings, nix *values* (which are what is actually compared) do not. as such nix hasn't used the symbol nature of string literals to optimize comparisons before this change, so there's no degradation at runtime.

Issue #4560, opened 2021-02-19
`nix search` only highlights last regex when multiple regexes are supplied


lilyball:
**Describe the bug**

If I run `nix search` with multiple regex arguments, it searches for packages that match all regexes, but it only highlights the last regex in the results.

**Steps To Reproduce**

1. Run `nix search perl repl`

**Expected behavior**

The results should highlight both `perl` and `repl` in the results.

**Actual behavior**

Only the `repl` is highlighted. If you swap the arguments, e.g. `nix search repl perl`, you get the same results but this time `perl` is highlighted.

**`nix-env --version` output**

nix-env (Nix) 2.3.10

iFreilicht:
@thufschmitt this has been fixed in #5945 over a year ago already, I guess you just forgot to close the issue? Just tested it and confirmed it still works:

![Screenshot 2023-11-27 at 21 47 45](https://github.com/NixOS/nix/assets/9742635/15e0f586-bc20-4101-b52a-cfe1457e416d)

thufschmitt:
Indeed, thanks!
Let's close it then

PR #5945, opened 2022-01-19
Make `nix search` highlight all regexes and matches

afishhh:
This PR makes the `nix search` command highlight all matches of all regexes provided on the command line, previously only the first match of the last regex would be highlighted.
Fixes issue #4560.

This is my first pull request so feedback would be appreciated!

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -33,27 +33,27 @@ std::string hilite(const std::string & s, const std::smatch & m, std::string pos
           + std::string(m.suffix());
 }
 
-std::string hilite_all(const std::string &s, std::vector<std::smatch> matches, std::string postfix) {
-    // Don't waste time on trivial highlights
-    if (matches.size() == 0)
+template<class Iter>
+std::string hilite_all(const std::string &s, Iter matches_first, Iter matches_last, std::string postfix) {
+    // Avoid copy on zero matches
+    if (matches_first == matches_last)
         return s;
-    else if (matches.size() == 1)
-        return hilite(s, matches[0], postfix);
 
-    std::sort(matches.begin(), matches.end(), [](const auto &a, const auto &b) {
+    std::sort(matches_first, matches_last, [](const auto &a, const auto &b) {
         return a.position() < b.position();
     });
 
     std::string out;
     ssize_t last_end = 0;
-    for (size_t i = 0; i < matches.size(); i++) {
-        auto m = matches[i];
+
+    for (Iter it = matches_first; it != matches_last; ++it) {
+        auto m = *it;
         size_t start = m.position();
         out.append(s.substr(last_end, m.position() - last_end));
         // Merge continous matches
         ssize_t end = start + m.length();
-        while(i + 1 < matches.size() && matches[i+1].position() <= end) {
-            auto n = matches[++i];
+        while(it + 1 != matches_last && (*(it + 1)).position() <= end) {
+            auto n = *++it;
             ssize_t nend = start + (n.position() - start + n.length());
             if(nend > end)
                 end = nend;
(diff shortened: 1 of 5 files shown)


src/nix/search.cc
```diff
@@ -20,16 +21,52 @@ std::string wrap(std::string prefix, std::string s)
...
     return
         m.empty()
         ? s
         : std::string(m.prefix())
-          + ANSI_GREEN + std::string(m.str()) + postfix
+          + HILITE_COLOR + std::string(m.str()) + postfix
           + std::string(m.suffix());
 }
 
+std::string hilite_all(const std::string &s, std::vector<std::smatch> matches, std::string postfix) {
```
thufschmitt, line 36: That function starts being non-trivial. It’s probably worth moving into `libutil`, and add a few tests for it

src/nix/search.cc
```diff
@@ -110,37 +145,47 @@ struct CmdSearch : InstallableCommand, MixJSON
...
-                            || !descriptionMatch.empty())
-                            found++;
+                        found = false;
+                        auto add_all = [&found](std::sregex_iterator it, std::vector<std::smatch>& vec){
+                            const auto end = std::sregex_iterator();
+                            while(it != end) {
+                                vec.push_back(*it++);
+                                found = true;
+                            }
+                        };
```
thufschmitt, line 161: Rather than consuming the `sregex_iterator` and putting everything in a vector, isn’t it possible to keep the iterator and directly pass it to `hilite_all`? That would make the code both simpler and (I think) marginally more efficient

thufschmitt (changes requested):
Thanks for the PR!

I’ve added a couple of suggestions, but it’s good overall :+1: 

Do you mind adding a few end-to-end tests in `tests/search.sh`? (would have to grep the output to search for the highlight groups, so not utterly pretty, but better than nothing)

src/nix/search.cc
afishhh: That's a good idea but I think we would still need a vector of iterators and such a solution probably wouldn't be any simpler.

src/nix/search.cc
```diff
@@ -20,16 +21,52 @@ std::string wrap(std::string prefix, std::string s)
...
+    else if (matches.size() == 1)
+        return hilite(s, matches[0], postfix);
+
+    std::sort(matches.begin(), matches.end(), [](const auto &a, const auto &b) {
+        return a.position() < b.position();
+    });
+
+    std::string out;
+    ssize_t last_end = 0;
+    for (size_t i = 0; i < matches.size(); i++) {
```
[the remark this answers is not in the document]
afishhh: Good idea, I've changed `hilite_all` to take an iterator of matches and it seems to work fine and is a bit clearer.

src/nix/search.cc
```diff
@@ -20,16 +21,52 @@ std::string wrap(std::string prefix, std::string s)
...
-          + ANSI_GREEN + std::string(m.str()) + postfix
+          + HILITE_COLOR + std::string(m.str()) + postfix
           + std::string(m.suffix());
 }
 
+std::string hilite_all(const std::string &s, std::vector<std::smatch> matches, std::string postfix) {
+    // Don't waste time on trivial highlights
+    if (matches.size() == 0)
+        return s;
+    else if (matches.size() == 1)
```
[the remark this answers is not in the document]
afishhh: You're probably right I don't think it makes much of a difference so I've removed the second case.

src/nix/search.cc
afishhh: I'm certainly going to add tests.
I think moving the function into libutil makes sense but I'm not very familiar with the nix source code, should I put it in a new file or for example `fmt.hh`.

src/nix/search.cc
thufschmitt: Oh indeed you’re right. Would have loved to have some kind of combined iterator to transparently iterate over both, but since there’s nothing standard for that, let’s keep the vector

src/nix/search.cc
thufschmitt: I’d say either `fmt.hh` or `ansicolor.hh` would make sense. Tbh libutil would need some restructuring because it has grown a bit messy over time

afishhh: I moved it into `fmt.hh` and created an `fmt.cc` to avoid linker errors, this required changing the signature to accept a vector but I hope that's fine.

thufschmitt (commented):
Thanks :)

I’ve just added a small docstring to the `hiliteMatches` function, but looks great otherwise.

Will merge as soon as the CI is green

PR #5946, opened 2022-01-19
Remove dead field in NixArgs

Ericson2314:
This has been unused since 170e86dff5724264e0d3d25b9af1bd42df6aec74

CC @thufschmitt

### DIFF
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -59,7 +59,6 @@ struct HelpRequested { };
 
 struct NixArgs : virtual MultiCommand, virtual MixCommonArgs
 {
-    bool printBuildLogs = false;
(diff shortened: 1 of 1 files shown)


src/nix/main.cc
```diff
@@ -59,7 +59,6 @@ struct HelpRequested { };
 
 struct NixArgs : virtual MultiCommand, virtual MixCommonArgs
 {
-    bool printBuildLogs = false;
```
cole-h, line 62: Looks like it was removed, not renamed? :P

Ericson2314: Oops yes I typed the wrong thing.

thufschmitt (approved):

PR #5949, opened 2022-01-20
Add back `copyClosure` for plain `StorePath`s

Ericson2314:
This was removed in 2e199673a523fa81de31ffdd2a25976ce0814631 when
`copyPath` transitioned to use `RealisedPath`. But then in
e9848beca704d27a13e28b4403251725bd485bb2 we added it back just for
`realisedPath`.

I think it is a good utility function --- one can easily imagine it
becoming optimized in the future, and copying paths *violating* the
closure is a very niche feature.

So if we have `copyPaths` for both sorts of paths, I think we should
have `copyClosure` for both sorts too.

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -1042,6 +1042,21 @@ void copyClosure(
     copyPaths(srcStore, dstStore, closure, repair, checkSigs, substitute);
 }
 
+void copyClosure(
+    Store & srcStore,
+    Store & dstStore,
+    const StorePathSet & storePaths,
+    RepairFlag repair,
+    CheckSigsFlag checkSigs,
+    SubstituteFlag substitute)
+{
+    if (&srcStore == &dstStore) return;
+
+    StorePathSet closure;
+    srcStore.computeFSClosure(storePaths, closure);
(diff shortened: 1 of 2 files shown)


src/libstore/store-api.hh
```diff
@@ -794,6 +794,13 @@ void copyClosure(
     CheckSigsFlag checkSigs = CheckSigs,
     SubstituteFlag substitute = NoSubstitute);
 
+void copyClosure(
```
Ericson2314, line 797: No doc because documented above. Just like `copyClosure` right above that where only the first overload is documented.

src/libstore/store-api.cc
```diff
@@ -1042,6 +1042,21 @@ void copyClosure(
...
 
+void copyClosure(
+    Store & srcStore,
+    Store & dstStore,
+    const StorePathSet & storePaths,
+    RepairFlag repair,
+    CheckSigsFlag checkSigs,
+    SubstituteFlag substitute)
+{
+    if (&srcStore == &dstStore) return;
```
Ericson2314, line 1053: Not super reliable, but done in the other `copyClosure`

Ericson2314:
I used this in https://github.com/NixOS/hydra/pull/1125

PR #5954, opened 2022-01-21
nix-shell: print sourced shell file in higher levels

SuperSandro2000:
to easen debugs when nix-shell is not working in your shell

Due to unrelated things my ``nix-shell?`` broke. Right now you need to provide 4 ``-v`` to display where the temporary file is. At that level quite a bunch of other stuff is logged which is far less important in my opinion.


edolstra:
I think `debug` level (so one higher) is more appropriate, since I don't think there are many use cases other than debugging `nix-shell` where you want it to print the shell script.

SuperSandro2000:
amended and force pushed

SuperSandro2000:
bump

Ericson2314:
@edolstra?

Issue #5716, opened 2021-12-02
Document flake outputs


Radvendii:
**Is your feature request related to a problem? Please describe.**
It's hard to figure out what the set of commonly used flake outputs are.

**Describe the solution you'd like**
This should be documented somewhere, with an explanation of each output. In addition, there should be a flake template

This is a little bit of a tricky situation, because not all of these are officially recognized by the `nix` command in any way. So it's unclear whether they really should be documented, and if so where. A general user coming across flakes, though, is not going to care. They're going to want the information all in one obvious place.

CC: @tomberek @garbas (to be put in the Nix UX Project)

tomberek:
Started work on this here: https://github.com/NixOS/templates/pull/9

tomberek:
Closing as stale. Re-open if needed.

Issue #5532, opened 2021-11-10
Rename `defaultX` flake outputs to `X.default`


edolstra:
We have a couple of flake outputs that have a corresponding "default" flake output, e.g.

* `packages` and `defaultPackage`
* `devShells` and `devShell`
* `apps` and `defaultApp`
* `overlays` and `overlay`
* `nixosModules` and `nixosModule`
* `templates` and `defaultTemplate`
* `bundlers` and `defaultBundler`

There are also some flake outputs that lack a default, but could probably use one, like `nixosConfigurations`.

As can be seen above, the naming is inconsistent. It also requires a lot of special handling in commands like `nix flake show` and `nix flake check`. So I propose getting rid of all the singular flake outputs and using an attribute named `default` as the default. E.g. 

* `defaultPackage.<system>` => `packages.<system>.default`
* `devShell.<system>` => `devShells.<system>.default`
* `overlay` => `overlays.default`

Migration: we can deprecate the old singular flake outputs in 2.5 and remove support for them some time later.

regnat:
One issue with this is that we’ll then have a `default` in the same namespace as the rest, which isn’t very clean.
Like if I want something that iterates over all the packages exported by a flake (or my flake has `hydraJobs = self.packages`), then there will be a `default` package that’s probably just a duplicate of another one and doesn’t really make sense.

Also, if I want to reuse the outputs of flake as an overlay (do something like `import nixpkgs { overlays = [ (_: prev: nix.packages.${prev.system}) ]; }`, then I’ll also add a `default` package.

(not that these are really blockers of course, but it’s mildly annoying)

michaelpj:
Having suffered through lots of annoyance due to working with some nix code that mixes derivations with different meanings together: please keep collections homogeneous! Put the special attributes somewhere else.

jared-w:
This brings up a slightly unrelated point of annoyance for me. There's a solid amount of flake attributes that are system namespaced as well as ones that are not system namespaced. The separation prevents things like cleanly writing a `forAllSystems {}` flake generator (you end up with `forAllSystems {} // { the rest of the flake}` or, worse, `forAllSystems {} // ( if arm mac then {}) // (if x86_64 linux then {}) // ...` 

To reiterate on @michaelpj's point, one of the big appeals behind flakes (for me) was that:

- purity meant you could actually cache stuff
[...]

PR #5955, opened 2022-01-21
Add documentation for common flake outputs

Radvendii:
First attempt at addressing #5716 

Some issues with this:
1. Kind of verbose.
   - it would be nice if there were a standard conversion between singular / plural as discussed in #5532, as the documentation could be half as long.
   - I'm not sure if this is the right place for all the information that it contains. maybe we want something more slimmed down with references to where to find the rest of the information (for instance, `apps` could just have a link to the `nix run` page, and not tell people how `nix run` works.
1. Bundler interface is in flux right now, so that section is wating on #5456 
1. I've never used Hydra, so I don't know what to write for the `hydraJobs` section.
1. I'm worried that this will become out of date very quickly. How are we going to make sure it gets kept up to date?

### DIFF
--- a/src/nix/flake.md
+++ b/src/nix/flake.md
@@ -301,6 +301,11 @@ The following attributes are supported in `flake.nix`:
   value (e.g. `packages.x86_64-linux` must be an attribute set of
   derivations built for the `x86_64-linux` platform).
 
+* `nixConfig`: An attribute set specifying configuration options for nix the
+  same as you would set in [`nix.conf`](./how-do-i-reference/conf-file.md).
+  These options will be enabled when building outputs from this flake. Options
+  that may compromise security must be approved by the user.
+
 ## Flake inputs
 
 The attribute `inputs` specifies the dependencies of a flake, as an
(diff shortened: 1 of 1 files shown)


edolstra:
To prevent it from becoming outdated, maybe it should be restricted to flake outputs that are used/recognized by `nix` itself (including ones that are checked by `nix flake check`, like `nixosModules`).

The manpage for `nix flake check` contains some of this info already in the "Evaluation checks" section, so maybe we should merge that section into this list to avoid redundancy.

src/nix/flake.md
```diff
@@ -415,6 +420,244 @@ way. Most flakes provide their functionality through Nixpkgs overlays
...
+While the technical specification of a flake does not include any details about
+the outputs, there are some common conventions, and tools that expect certain
+outputs. These are documented here, including how they are typically used,
+whether they are recognized by [`nix flake check`](./flake-check.md), and what
+type they should be.
+
+### checks
+Used-by: `nix flake check`
+Checked: Yes
+Type: systemAttrs (attrsOf (derivation))
```
edolstra, line 434: This doesn't format in a nice way, namely "Used-by: nix flake check Checked: Yes Type: systemAttrs (attrsOf (derivation))".

tomberek:
- [ ] cross-reference with templates repo

tomberek:

PR #5957, opened 2022-01-21
convert more string copies into string_views

pennae:
most notably we don't have to copy strings that are already allocated in suitable memory (eg when pulling a string out of a tString or tPath Value). avoiding these copies gives us another ~1% on search and system eval, but improvements from copy avoidance look like they're starting to level out. time to go somewhere else :)

### current master

```
nix search --no-eval-cache --offline ../nixpkgs hello
  Time (mean ± σ):      8.907 s ±  0.053 s    [User: 6.814 s, System: 1.650 s]
  Range (min … max):    8.829 s …  9.049 s    20 runs
 
nix eval -f ../nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix
  Time (mean ± σ):     376.8 ms ±   1.6 ms    [User: 341.9 ms, System: 34.9 ms]
  Range (min … max):   375.3 ms … 381.2 ms    20 runs
 
nix eval --raw --impure --file expr.nix
  Time (mean ± σ):      2.839 s ±  0.037 s    [User: 2.538 s, System: 0.230 s]
  Range (min … max):    2.758 s …  2.907 s    20 runs
```

### this patchset

```
nix search --no-eval-cache --offline ../nixpkgs hello
  Time (mean ± σ):      8.820 s ±  0.047 s    [User: 6.744 s, System: 1.667 s]
  Range (min … max):    8.720 s …  8.882 s    20 runs
 
nix eval -f ../nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix
  Time (mean ± σ):     370.3 ms ±   2.5 ms    [User: 335.5 ms, System: 34.8 ms]
  Range (min … max):   368.0 ms … 379.8 ms    20 runs
 
nix eval --raw --impure --file expr.nix
  Time (mean ± σ):      2.806 s ±  0.028 s    [User: 2.503 s, System: 0.235 s]
  Range (min … max):    2.738 s …  2.847 s    20 runs
```

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1857,7 +1857,7 @@ void EvalState::forceFunction(Value & v, const Pos & pos)
 }
 
 
-string EvalState::forceString(Value & v, const Pos & pos)
+std::string_view EvalState::forceString(Value & v, const Pos & pos)
 {
     forceValue(v, pos);
     if (v.type() != nString) {
@@ -1866,7 +1866,7 @@ string EvalState::forceString(Value & v, const Pos & pos)
         else
             throwTypeError("value is %1% while a string was expected", v);
     }
-    return string(v.string.s);
+    return v.string.s;
 }
 
 
@@ -1901,17 +1901,17 @@ std::vector<std::pair<Path, std::string>> Value::getContext()
 }
 
 
-string EvalState::forceString(Value & v, PathSet & context, const Pos & pos)
+std::string_view EvalState::forceString(Value & v, PathSet & context, const Pos & pos)
 {
-    string s = forceString(v, pos);
+    auto s = forceString(v, pos);
     copyContext(v, context);
     return s;
 }
 
 
-string EvalState::forceStringNoCtx(Value & v, const Pos & pos)
+std::string_view EvalState::forceStringNoCtx(Value & v, const Pos & pos)
 {
-    string s = forceString(v, pos);
+    auto s = forceString(v, pos);
     if (v.string.context) {
         if (pos)
             throwEvalError(pos, "the string '%1%' is not allowed to refer to a store path (such as '%2%')",
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1,5 +1,6 @@
 #include "eval.hh"
 #include "hash.hh"
+#include "types.hh"
 #include "util.hh"
 #include "store-api.hh"
 #include "derivations.hh"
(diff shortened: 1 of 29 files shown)


src/libexpr/eval.cc
```diff
@@ -1943,57 +1946,58 @@ std::optional<string> EvalState::tryAttrsToString(const Pos & pos, Value & v,
     return {};
 }
 
-string EvalState::coerceToString(const Pos & pos, Value & v, PathSet & context,
-    bool coerceMore, bool copyToStore, bool canonicalizePath)
+std::string_view EvalState::coerceToString(const Pos & pos, Value & v, PathSet & context,
```
edolstra, line 1949: Returning a `string_view` here is too ugly/risky for me, since it requires the caller to manage a buffer. If only C++ had Rust's Cow type...

src/libexpr/primops.cc
```diff
@@ -3388,7 +3396,18 @@ static RegisterPrimOp primop_hashString({
 
 struct RegexCache
 {
-    std::unordered_map<std::string, std::regex> cache;
+    // TODO use C++20 transparent comparison when available
```
edolstra, line 3399: IIRC this already works with gcc 10.

src/libutil/util.hh
```diff
@@ -62,7 +62,7 @@ Path canonPath(PathView path, bool resolveSymlinks = false);
    everything before the final `/'.  If the path is the root or an
    immediate child thereof (e.g., `/foo'), this means `/'
    is returned.*/
-Path dirOf(const Path & path);
+PathView dirOf(const PathView path);
```
edolstra, line 65: In my experience returning string_views from functions is a great way to introduce subtle bugs when the lifetime of the returned value exceeds the lifetime of the argument... It's also not nice that you can't pass it temporary values.

src/libexpr/eval.cc
pennae: yeah, that'd be great. i guess we could add a simple variant of that?

src/libutil/util.hh
pennae: if only C++ had lifetimes, then this wouldn't be that much of an issue :disappointed: what's the best way forward here then? change, revert, something else? `dirOf` in particular isn't very hot and most of its users do make a copy of its return value, so undoing that wouldn't hurt much.

edolstra: Yeah, unless it's on a hot path it's probably not worth it. I have bad experience writing things like `auto s = dirOf(bar())` where `bar()` returns a temporary...

src/libexpr/eval.cc
edolstra: Hm yes, maybe a (wrapper around) `std::variant<std::string_view, std::string>` would be pretty nice.

PR #5967, opened 2022-01-23
nix-shell: run `set -x` when NIX_DEBUG is set

SuperSandro2000:


SuperSandro2000:
@edolstra ping

SuperSandro2000:
bump

edolstra:
Team discussion:

* This can be useful, but it can also be potentially spammy, since it will show all stdenv setup shell commands.
* Maybe this should be a separate option, so as not to overload `NIX_DEBUG`?
* Maybe the user can just set `preHook = "set -x"` in their derivation?

roberth:
More notes from the meeting today:

- @ericson2314: Rather have this be a flag. Just pinning it on some env var which was until now a Nixpkgs-only thing (despite the name) doesn't seem right.
  - Rather have it be a proper flag

- @edolstra: Can use `preHook` it runs pretty early
  - @ericson2314: Well, when one is debugging and feeling confused what is going on, it one wants to see *everything* just to rule it out.

Issue #5970, opened 2022-01-24
Set builtins.nixVersion to a constant?


edolstra:
`builtins.nixVersion` is an impurity that can cause evaluation results to differ between Nix versions, see https://github.com/NixOS/nixpkgs/pull/156527. So we could set it to a fixed value like `2.6`. It's not clear whether there is a legitimate use case for `builtins.nixVersion`.

bjornfor:
Did you see this use-case? https://github.com/nix-community/poetry2nix/pull/512/commits/5c06704d4584cbd17d66c214447d313a896811a2. (Arguably should not be needed, bit here we are.)

edolstra:
Yeah regressions should be fixed.

tomberek:
Closing as stale. Re-open if needed.

PR #5971, opened 2022-01-24
In pure mode, set builtins.nixVersion to a constant

edolstra:
This prevents issues like
https://github.com/NixOS/nixpkgs/pull/156527. Eventually we should do
this in impure mode as well. 

Issue #5970.

Depends on #8330

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -3822,7 +3822,7 @@ void EvalState::createBaseEnv()
         addConstant("__currentSystem", v);
     }
 
-    v.mkString(nixVersion);
+    v.mkString(evalSettings.pureEval ? "2.6.0" : nixVersion);
     addConstant("__nixVersion", v);
 
     v.mkString(store->storeDir);


edolstra:
Yes, but expressions like `builtins ? whatever` are unlikely to end up embedded in strings like in NixOS/nixpkgs#156527.

roberth:
Adding a `nixVersionAtLeast` function seems like a good compromise. It produces a boolean, so it's unlikely to end up in a build like that.

thufschmitt:
> Adding a `nixVersionAtLeast` function seems like a good compromise. It produces a boolean, so it's unlikely to end up in a build like that.

Yes, that sounds like a fairly good one

(and maybe just deprecating `builtins.nixVersion` for the time being to avoid breaking stuff too badly)

roberth:
Let's not forget about `builtins.langVersion`.

roberth:
> Adding a `nixVersionAtLeast` function seems like a good compromise.

This does make error messages less informative. Another solution is to poison the string context, similar to what we could do for secret strings, but slightly different allowed usages. String context poisoning has its own issue now:
 - https://github.com/NixOS/nix/issues/8388

fricklerhandwerk:
Discussed in the Nix team meeting 2023-06-05:

- @roberth proposes to add a comparison function instead and encourage its use, so version strings won't end up in builds
    - should deprecate `nixVersion` eventually of course
    - could be `impureNixVersionAtLeast` to signal
- @edolstra: could ask what's the use case of knowing the version. usually to determine if a primop is available. but that's what `?` is used 
    - @roberth: in case we happen to release a buggy builtin, it my 
- @fricklerhandwerk: should the constant be something yelling "don't use"?
    - @edolstra: could stop incrementing it at 2.16
- @ericson2314: what speaks against removing it in pure eval?
    - @edolstra: might break lots of consumers
    - @fricklerhandwerk: does Nixpkgs build in pure eval?
      - @edolstra: no, but possibly in restricted eval
      - @roberth: also flakes
- conclusion: merge https://github.com/NixOS/nix/pull/8330 first so we can document `nixVersion` to be discouraged, pin output 2.16

roberth:
Proposing something similar for a feature detection function: https://git.lix.systems/lix-project/lix/issues/607#issuecomment-9297

roberth (changes requested):
See earlier comments

PR #5992, opened 2022-01-25
Rename vendored nlohmann-json to avoid other-Linux collisions

### LINKED ISSUE #5981: Conflicts with system provided nlohman-json
**Describe the bug**

Nix 2.6 has a couple of files that conflict with system provided nlohman-json, which will cause issues with every Linux distribution.

This was introduced in this [PR](https://github.com/NixOS/nix/pull/5536) and this [commit](https://github.com/NixOS/nix/commit/307977963ca953ad11003bcd4c95f7d2308012f1).

**Steps To Reproduce**

Attempt to install Nix 2.6 and any distribution provided nlohman-json.

**Expected behavior**

A better solution.

**Additional context**

This breaks any addition of Nix 2.6+ to any Linux distribution.

roberth:
Closes #5981

I've renamed the directory, both in the source tree and `include` output.
I have not moved things to a new namespace at the C++ level, as it is not necessary for #5981 and it keeps the change simple.


edolstra:
I've created #5996 to stop vendoring nlohmann/json altogether.

roberth:
@edolstra, works for me.

@SuperSandro2000, what's up with the :-1: reaction? For this PR I really don't care, but next time you use it, please clarify, so people know how to improve. :heart:

SuperSandro2000:
I thought it was clear but I prefer the other PR because devendoring sounds to me like a better solution than renaming imports if we don't need to.

roberth:
Well, we're giving up some control. A bug in the JSON lib could affect values and therefore reproduciblity. Decoupling such a bug from the Nix version seems like a bad idea, but I have more pressing issues to worry about.
See, now we can have a discussion :)

Issue #5989, opened 2022-01-25
Allow flake `nixConfig` attribute to contain architecture-specific options


lilyball:
**Is your feature request related to a problem? Please describe.**
I want to have my flake specify e.g. `nixConfig.aarch64-darwin.extra-platforms = [ "x86_64-darwin" ]` without setting this for other platforms.

**Describe the solution you'd like**
Right now `nixConfig` doesn't allow nested attribute sets (if I'm reading the code correctly). It could support nested attribute sets by assuming the key is a platform and recursing only if the platform matches.

edolstra:
I'm not sure we should do this. It gives the flake file format a hard dependency on our current simplistic concept of systems (i.e. simple strings like "x86_64-linux"), which doesn't account for cross-compilation and doesn't match the richer definition in lib/systems in nixpkgs.

It also seems ugly/ambiguous to put system types in the same namespace as options (i.e. in `nixConfig.aarch64-darwin` it's not clear without domain knowledge that `aarch64-darwin` is a system type rather than an option name).

thufschmitt:
I agree with @edolstra. I think I see why you’d want that, but it’s a bit too ad-hoc and rigid to be freezed forever in the flake schema.

I think this kind of “twisted” use-case would be better served by a custom wrapper that would be more flexible imho

lilyball:
Regarding freezing in the flake schema, Nix already faces challenges with the idea of changing the system strings as `builtins.currentSystem` returns just such a string and changing this is likely to break a lot. And of course the current strings are used in a lot of the outputs already, with this use directly understood by the `nix` tools, meaning we've already baked this into the flake schema.

As for cross-compilation, we're talking about the options that affect Nix, not anything involved in the actual building of packages.

I have no objection to using `nixConfig.perSystem.<system>` for this. We should make sure list-typed config values will merge together, though I'm not sure right now if it already does that for any `nixConfig` entries that match things from `nix.conf` (I hope it does).

@thufschmitt I don't know what sort of "custom wrapper" you mean. I'm writing a flake, not writing a build system. My flake should work as-is with existing Nix tooling, and in particular with any existing wrappers around Nix tooling.

lincolnauster:
I have basically the same opinions and questions as @lilyball: I do
agree that we're kind of already stuck with host triples in the schema,
and I don't quite see the relevancy of cross-compilation, either.

> I have no objection to using `nixConfig.perSystem.<system>` for this.

[...]

PR #5993, opened 2022-01-26
Generalize nixConfig attributes to systems

lincolnauster:
See #5989 for context. The implementation here should be pretty straightforward.

Assuming no outstanding implementation issues, how can (should?) I write tests for these changes? Given the inherent host environment impurity, I'm not sure where to start with it.


edolstra:
See my comment here: https://github.com/NixOS/nix/issues/5989#issuecomment-1022005590

Regarding the implementation: `applyAttrs()` shouldn't depend on the system type. Rather, `ConfigFile` should be extended with a `std::map<std::string, std::map<std::string, ConfigValue>> settingsPerSystem` field.

The name `applyAttrs()` is confusing given the `apply()` method, since it doesn't apply any settings. Probably `parseAttrs()` is better.

lincolnauster:
@edolstra - I modified and force-pushed the patchest given your feedback; hopefully it's a bit cleaner now. The git history is similar to the previous iteration, but the perSystem attribute is implemented in a commit following the one allowing system specific changes in the first place.

lincolnauster:
Rebased on current master and added documentation.

Ericson2314:
CC @roberth I suspect this should be linked to the other issues trying to solve flake system problems?

roberth:
What is it for?

The original use case (`aarch64-darwin.extra-platforms = [ "x86_64-darwin" ];`) seems to have been fixed for many Nix releases now. Judging by the linked issue, I'm skeptical of the complexity this adds.

@Ericson2314 idk where to link for this specifically. It seems that this shares the problem of having to enumerate systems, and this only adds to the problem.

> Regarding the implementation: `applyAttrs()` shouldn't depend on the system type.

Maybe it used to be a function? Sad if true.

PR #5997, opened 2022-01-26
Fix the `store ping` test

thufschmitt:


edolstra:
Does this actually fix the issue though? `nix-daemon --version` prints `nix-daemon (Nix) 2.7.0` while `nix store ping` prints `2.7.0`, so grepping for the former will fail.

thufschmitt:
> Does this actually fix the issue though? `nix-daemon --version` prints `nix-daemon (Nix) 2.7.0` while `nix store ping` prints `2.7.0`, so grepping for the former will fail.

Duh you’re right :facepalm: I though the issue was just that it was picking the old version too, but indeed it’s not

PR #6000, opened 2022-01-27
Use the `nix` command (and flakes) in the CI

thufschmitt:
Apart from a slight simplification and a bit of dogfooding, this also make the cache behavior more predictable.
For example `nix build .` and `nix build nix/$(git rev-parse HEAD)` will yield the exact same path, while their “intuitive” non-flake equivalents (`nix-build` and `nix-build https://github.com/nixos/nix/archives/$(git rev-parse HEAD).tar.gz`) don’t.

This was a pain for example in https://github.com/NixOS/nix/pull/5059

Also, the `bar-with-logs` log format is imho nicer (even in an non-interactive context) because prefixing each log line with the name of the derivation that produced it makes it much easier to follow what’s going on.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -447,19 +447,7 @@
         installerScriptForGHA = installScriptFor [ "x86_64-linux" "x86_64-darwin" "armv6l-linux" "armv7l-linux"];
 
         # docker image with Nix inside
-        dockerImage = nixpkgs.lib.genAttrs linux64BitSystems (system:
-          let
-            pkgs = nixpkgsFor.${system};
-            image = import ./docker.nix { inherit pkgs; tag = version; };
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -606,6 +594,20 @@
 
           hardeningDisable = [ "pie" ];
         };
+        dockerImage =
```
edolstra, line 597: Since this adds an attribute to the overlay, it may be better to use a more descriptive name like `nixDockerImage`.

thufschmitt: That shouldn’t since I’ve only added it to `packages` and `overlay` doesn’t include these.

But I can move it to `nix.passthru.dockerImage` (and then include it in the overlay) or something like that if you prefer

Issue #5781, opened 2021-12-16
Build fails on DESTDIR= build environments: No rule to make target '/usr/bin/nix'


trofi:
Linux distributions usually install packages into sandboxed environments before moving packages into live system:

```
$ ./configure --prefix=/usr
$ make install DESTDIR=/sandbox
$ mv /sandbox/* /
```

`nix` fails at `make install` stage:

```
make: *** No rule to make target '/usr/bin/nix', needed by 'doc/manual/nix.json'.  Stop.
make: *** Waiting for unfinished jobs....
```

On `NixOS` it's easiest to reproduce by using inaccessible prefix:

```
$ nix develop -f ~/nixpkgs/master nixUnstable
$ ./bootstrap.sh
$ ./configure --prefix=/usr
$ make install DESTDIR=$PWD/__sandbox__
...
make: *** No rule to make target '/usr/bin/nix', needed by 'doc/manual/nix.json'.  Stop.
```

I sprinlked `s@$(bindir)/nix@$(DESTDIR)$(bindir)/nix@g` around with:

```diff
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -20,7 +20,7 @@ dummy-env = env -i \
        NIX_STATE_DIR=/dummy \
        NIX_CONFIG='cores = 0'

-nix-eval = $(dummy-env) $(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
+nix-eval = $(dummy-env) $(DESTDIR)$(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw

 $(d)/%.1: $(d)/src/command-ref/%.md
        @printf "Title: %s\n\n" "$$(basename $@ .1)" > $^.tmp
@@ -44,7 +44,7 @@ $(d)/src/SUMMARY.md: $(d)/src/SUMMARY.md.in $(d)/src/command-ref/new-cli
        $(trace-gen) cat doc/manual/src/SUMMARY.md.in | while IFS= read line; do if [[ $$line = @manpages@ ]]; then cat doc/manual/src/command-ref/new-cli/SUMMARY.md; else echo "$$line"; fi; done > $@.tmp
        @mv $@.tmp $@
```

But it's not enough as nix relies on it's own local libraries:

```
$ dev>LANG=C make install DESTDIR=$PWD/__sandbox__
[...]

PR #6005, opened 2022-01-27
doc/manual/local.mk: handle DESTDIR= for inplace builds

trofi:
Linux distributions usually install packages into sandboxed environments before moving packages into live system:

    $ ./configure --prefix=/usr
    $ make install DESTDIR=/sandbox
    $ mv /sandbox/* /

`nix` fails at `make install` stage:

    make: *** No rule to make target '/usr/bin/nix', needed by 'doc/manual/nix.json'.  Stop.

The change sprinkles enough $(DESTDIR) prefixes to make it work.
To allow nix inplace execution LD_LIBRARY_PATH was added into $DESTDIR location.

Closes: https://github.com/NixOS/nix/issues/5781

### DIFF
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -14,13 +14,14 @@ clean-files += $(d)/*.1 $(d)/*.5 $(d)/*.8
 # Provide a dummy environment for nix, so that it will not access files outside the macOS sandbox.
 # Set cores to 0 because otherwise nix show-config resolves the cores based on the current machine
 dummy-env = env -i \
+	LD_LIBRARY_PATH=$(DESTDIR)$(libdir) \
 	HOME=/dummy \
 	NIX_CONF_DIR=/dummy \
 	NIX_SSL_CERT_FILE=/dummy/no-ca-bundle.crt \
 	NIX_STATE_DIR=/dummy \
 	NIX_CONFIG='cores = 0'
 
-nix-eval = $(dummy-env) $(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
+nix-eval = $(dummy-env) $(DESTDIR)$(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
 
 $(d)/%.1: $(d)/src/command-ref/%.md
 	@printf "Title: %s\n\n" "$$(basename $@ .1)" > $^.tmp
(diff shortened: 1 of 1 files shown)


thkoch2001:
Thank you for working on this, but please do not merge this PR, as it is not fixing the root cause. See instead this patch:
https://github.com/NixOS/nix/pull/6015

The root cause is, that something (doc/manual/local.mk) in the build-phase has prerequisites on an install-phase target ($(bindir)/nix). The fix is to depend on a build-phase target instead.

trofi:
> Thank you for working on this, but please do not merge this PR, as it is not fixing the root cause. See instead this patch: #6015
> 
> The root cause is, that something (doc/manual/local.mk) in the build-phase has prerequisites on an install-phase target ($(bindir)/nix). The fix is to depend on a build-phase target instead.

Heh, I think I tried it before and @edolstra rejected it in https://github.com/NixOS/nix/pull/5185. I wonder how many more hour will people waste on it.

Does it work for you as is? I would expect `$(nix_DIR)/nix` not to be runnable as it links against inplace shared libraries (`libnixexpr.so` and so on). You would need to pass them through somehow. Say, via `LD_LIBRARY_PATH=` variable.

I suspect it works for you only because you already have nix installed with libraries in `/usr/lib` (or equivalent).

thkoch2001:
You can see the build log of my Debian package here:
https://buildd.debian.org/status/fetch.php?pkg=nix&arch=amd64&ver=2.6.0%2Bdfsg-1&stamp=1643544675&raw=0

When you search for "src/nix/nix" you find invocations of the inplace-build nix binary. They work. There's definitely no nix preinstalled on these machines.

I believe that this works because the nix build system is actually linking the nix binary two times for the inplace and the install target instead of linking it once and then copying it. See mk/programs.mk and these two rules:

inplace linking:

```Make
	$$(trace-ld) $(CXX) -o $$@ $$(LDFLAGS) $$(GLOBAL_LDFLAGS) $$($(1)_OBJS) $$($(1)_LDFLAGS) $$(foreach lib, $$($(1)_LIBS), $$($$(lib)_LDFLAGS_USE))
```

install-place linking:

```Make
      $(DESTDIR)$$($(1)_INSTALL_PATH): $$($(1)_OBJS) $$(_libs_final) | $(DESTDIR)$$($(1)_INSTALL_DIR)/
	$$(trace-ld) $(CXX) -o $$@ $$(LDFLAGS) $$(GLOBAL_LDFLAGS) $$($(1)_OBJS) $$($(1)_LDFLAGS) $$(foreach lib, $$($(1)_LIBS), $$($$(lib)_LDFLAGS_USE_INSTALLED))
```

trofi:
Ah, good point!

PR #6008, opened 2022-01-28
Fix systemctl kill no killing stuck child processes

SuperSandro2000:
For unknown reason I had some stuck daemon child processes today. I wanted to use ``systemctl kill nix-daemon`` which had no affect. This is because `KillMode` was set to `process` which only kills the main process, not child ones. I had to use htop to kill all childs to get my daemon back into a working condition. The systemd man page recommends not to set this.

See https://www.freedesktop.org/software/systemd/man/systemd.kill.html


edolstra:
This is intentional. It ensures that restarting the daemon (e.g. when doing an upgrade) does not kill builds that are already in progress.

SuperSandro2000:
And how are you supposed to kill a stuck child process then without searching for the pids and manually killing them? systemd can't do and the stuck child processes blocked everything for me

PR #6015, opened 2022-01-29
use nix from build phase in build phase

thkoch2001:
Fixes: https://github.com/NixOS/nix/issues/5781

For linux distributions it is important that a projects build and install
steps are clearly separated and that the install target can be overridden
with the DESTDIR variable:

https://www.gnu.org/software/make/manual/html_node/DESTDIR.html

Before this patch some targets in doc/manual/local.mk had prerequisites on
the INSTALLED nix binary which caused the BUILD step to depend on the INSTALL
step. This patch changes the prerequisites to depend on the built but not
yet installed nix binary.

### DIFF
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -20,7 +20,7 @@ dummy-env = env -i \
 	NIX_STATE_DIR=/dummy \
 	NIX_CONFIG='cores = 0'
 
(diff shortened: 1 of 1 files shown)


Ericson2314:
Closing as duplicate of https://github.com/NixOS/nix/pull/5145

PR #6018, opened 2022-01-29
canonPath: fix missing slash when resolving links

### LINKED ISSUE #6017: regression: string handling bug triggered via `nix profile remove`
**Describe the bug**

The bug I'm encountering is that presently when I run `nix profile remove` on anything (including items not installed), I get the following error: 

```
error: getting status of '/nix/store/2jq0a1260qidl0qf8f1rcshny2wc79rh-libreoffice-7.2.5.2/shareicons': No such file or directory
```

Of course this file does not actually exist, although there is a `/...-7.2.5.2/share/icons` (slash between).

I have an strace of what happens during this if that's useful (I thought maybe it was something weird following a bad link or something...).  Valgrind seems clear, just usual Nix stuff :).  It does follow a link from `/nix/store/j56jnpmqrlxky4g89v1n636cqlgq10nb-libreoffice-7.2.5.2/share` to `/nix/store/2jq0a1260qidl0qf8f1rcshny2wc79rh-libreoffice-7.2.5.2` or so, if that helps figure out where this is going wrong.

I noticed v2.5.1 doesn't have this problem, so used `git bisect run` to figure out when this started occurring with this little script: 
```shell
#!/bin/sh

set -e

rm asdf  || true
nix-build -o asdf

./asdf/bin/nix profile remove nopers
```

Which points to 44c92a1667ce829518178a77a9de0a53609d284f as 'the first bad commit' (which given the nature of what it changes, seems not unlikely :)).

**Steps To Reproduce**

Installing the package in question into a test profile (hence #6016, sorry) doesn't seem to trigger the issue.

So, unfortunately I'm not sure how to reproduce this at the moment.

Happy to help test things if that's useful, of course :+1: .

**Expected behavior**

`nix profile remove` doesn't error out on a non-existent file :).

`nix-env (Nix) 2.7.0pre20220124_0a70b37`

dtzWill:
Fixes #6017

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -147,7 +147,7 @@ Path canonPath(PathView path, bool resolveSymlinks)
                 path = {};
             } else {
                 s += path.substr(0, slash);
-                path = path.substr(slash + 1);
+                path = path.substr(slash);
             }
 
             /* If s points to a symlink, resolve it and continue from there */


dtzWill:
Thanks @pennae, this resolved the issue!

PR #6027, opened 2022-01-31
repl: `--option pure-eval true` actually enables pure eval mode

### LINKED ISSUE #5867: `nix repl` accepts but silently ignores `--option pure-eval true`
**Describe the bug**

I've used nix repl to test whether a feature works in pure-eval.
It seemed to work, but not outside of the repl, [causing](https://github.com/NixOS/nixpkgs/pull/153592#issuecomment-1005630910) an error for someone.

**Steps To Reproduce**

```
$ nix repl --option pure-eval true
Welcome to Nix 2.4. Type :? for help.

nix-repl> builtins.storePath /nix/store/90y23lrznwmkdnczk1dzdsq4m35zj8ww-bash-interactive-5.1-p8/bin/bash
"/nix/store/90y23lrznwmkdnczk1dzdsq4m35zj8ww-bash-interactive-5.1-p8/bin/bash"


$ nix-instantiate --option pure-eval true --expr --eval 'builtins.storePath /nix/store/90y23lrznwmkdnczk1dzdsq4m35zj8ww-bash-interactive-5.1-p8/bin/bash'
error: 'builtins.storePath' is not allowed in pure evaluation mode

       at «string»:1:1:

            1| builtins.storePath /nix/store/90y23lrznwmkdnczk1dzdsq4m35zj8ww-bash-interactive-5.1-p8/bin/bash
             | ^
```

**Expected behavior**

At least a warning. Ideally, `nix repl` _defaults_ to impure, but can be set to pure-eval.

**Additional context**

Seems to be caused by https://github.com/NixOS/nix/blob/70dfcbbb37c114550fe1228748d30fd60520fe10/src/nix/repl.cc#L890

Ma27:
To quote Eelco in #5867:

> Unfortunately we can't do
>
>     evalSettings.pureEval.setDefault(false);
>
> because then we have to do the same in main.cc (where
> pureEval is set to true), and that would allow pure-eval
> to be disabled globally from nix.conf.

Basically, the simplest way to work around this is by not setting `pureEval`
in `main.cc` in the first place if a `repl` is started. Considering that
similar special cases already exist below and that it's now possible to
interactively test the actual pure-mode, I figured that this is an OK
solution.

Closes #5867

cc @edolstra @roberth


edolstra:
This seems a bit ugly to me... The special cases that currently exist are all internal (non-user-visible) operations. Also, if I understand it correctly, this wouldn't handle the case where `repl` isn't the first argument (e.g. `nix -v repl`).

Ma27:
> This seems a bit ugly to me... The special cases that currently exist are all internal (non-user-visible) operations

@edolstra I see two alternative options:

* Make it impossible to override `pure-eval` in `nix.conf` by erroring out after reading from `.conf`-files.
* Only set `pureEval` by default to `true` if the command "says so": i.e. we could add another method (such as `doc()` or `description()` called `pureEvalByDefault` or sth. similar that returns `true` by default and can be overriden by e.g. the `ReplCommand`. We'd then have to `evalSettings.pureEval = true;` down to line ~350 on my branch, can you elaborate if that's "safe"?

> Also, if I understand it correctly, this wouldn't handle the case where repl isn't the first argument (e.g. nix -v repl).

Hmm yeah, good point.

Ma27:
@edolstra do my suggestions from my previous comments more sense? If yes, I'd upate the PR accordingly :)

Ma27:
@edolstra I'd really like to see this being part of the next Nix release and I'm happy to bring this PR into a better shape for it. However I'd need some input from you regarding my suggestions first :)

Ma27:
As Eelco stated in https://github.com/NixOS/nix/issues/5867#issuecomment-1005731506, this can't be done. Going for the second option then :)

Ma27:
Updated the PR accordingly. Ping @edolstra @Ericson2314 @thufschmitt .

Ma27:
cc @edolstra @thufschmitt anything else to do from my side? The macos failure seems unrelated to me.

thufschmitt (approved):
> cc @edolstra @thufschmitt anything else to do from my side? The macos failure seems unrelated to me.

I've restarted the OSX failure (95% chances i'ts unrelated indeed). I'll merge once everything is green :+1:

PR #6029, opened 2022-01-31
ssh-ng: also store build logs to make them accessible by `nix log`

### LINKED ISSUE #5079: Allow fetching remote build logs with ssh-ng://
`nix-build foo.drv --store ssh-ng://host` and then one of `nix log foo.drv --store ssh-ng://host` or `nix log foo.drv --option substituters ssh-ng://host` should show the build log of foo.drv

maybe add an option to copy build logs with `nix copy` ?

**Additional context**
If I grepped the code correctly, the operation `getBuildLog` is only implemented for local stores and http substituters.

Ma27:
Right now when building a derivation remotely via

    $ nix build -j0 -f . hello -L --builders 'ssh://builder'

it's possible later to read through the entire build-log by running
`nix log -f . hello`. This isn't possible however when using `ssh-ng`
rather than `ssh`.

The reason for that is that there are two different ways to transfer
logs in Nix through e.g. an SSH tunnel (that are used by `ssh`/`ssh-ng`
respectively):

* `ssh://` receives its logs from the fd pointing to `builderOut`. This
  is directly passed to the "log-sink" (and to the logger on each `\n`),
  hence `nix log` works here.
* `ssh-ng://` however expects JSON-like messages (i.e. `@nix {log data
  in here}`) and passes it directly to the logger without doing anything
  with the `logSink`. However it's certainly possible to extract
  log-lines from this format as these have their own message-type in the
  JSON payload (i.e. `resBuildLogLine`).

  This is basically what I changed in this patch: if the code-path for
  `builderOut` is not reached and a `logSink` is initialized, the
  message was successfully processed by the JSON logger (i.e. it's in
  the expected format) and the line is of the expected type (i.e.
  `resBuildLogLine`), the line will be written to the log-sink as well.

Closes #5079

cc @edolstra @thufschmitt @Ericson2314 @symphorien

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1207,12 +1207,14 @@ void DerivationGoal::handleChildOutput(int fd, std::string_view data)
     if (hook && fd == hook->fromHook.readSide.get()) {
         for (auto c : data)
             if (c == '\n') {
-                auto s = handleJSONLogMessage(currentHookLine, worker.act, hook->activities, true);
-                if (s && !isWrittenToLog && logSink) {
-                    auto json = nlohmann::json::parse(std::string(currentHookLine, 5));
-                    if (json["type"] == resBuildLogLine) {
-                        auto f = json["fields"];
-                        (*logSink)((f.size() > 0 ? f.at(0).get<std::string>() : "") + "\n");
+                auto json = parseJSONMessage(currentHookLine);
+                if (json) {
+                    auto s = handleJSONLogMessage(*json, worker.act, hook->activities, true);
+                    if (s && !isWrittenToLog && logSink) {
+                        if ((*json)["type"] == resBuildLogLine) {
+                            auto f = (*json)["fields"];
+                            (*logSink)((f.size() > 0 ? f.at(0).get<std::string>() : "") + "\n");
+                        }
                     }
                 }
                 currentHookLine.clear();
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1177,6 +1177,7 @@ bool DerivationGoal::isReadDesc(int fd)
 
 void DerivationGoal::handleChildOutput(int fd, std::string_view data)
 {
+    // local & `ssh://`-builds are dealt with here.
     auto isWrittenToLog = isReadDesc(fd);
     if (isWrittenToLog)
     {
@@ -1210,11 +1211,11 @@ void DerivationGoal::handleChildOutput(int fd, std::string_view data)
                 auto json = parseJSONMessage(currentHookLine);
                 if (json) {
                     auto s = handleJSONLogMessage(*json, worker.act, hook->activities, true);
-                    if (s && !isWrittenToLog && logSink) {
-                        if ((*json)["type"] == resBuildLogLine) {
-                            auto f = (*json)["fields"];
-                            (*logSink)((f.size() > 0 ? f.at(0).get<std::string>() : "") + "\n");
-                        }
+                    // ensure that logs from a builder using `ssh-ng://` as protocol
+                    // are also available to `nix log`.
+                    if (s && !isWrittenToLog && logSink && (*json)["type"] == resBuildLogLine) {
+                        auto f = (*json)["fields"];
+                        (*logSink)((f.size() > 0 ? f.at(0).get<std::string>() : "") + "\n");
                     }
                 }
                 currentHookLine.clear();
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1175,10 +1175,10 @@ bool DerivationGoal::isReadDesc(int fd)
     return fd == hook->builderOut.readSide.get();
 }
 
-
 void DerivationGoal::handleChildOutput(int fd, std::string_view data)
 {
-    if (isReadDesc(fd))
+    auto isWrittenToLog = isReadDesc(fd);
+    if (isWrittenToLog)
     {
         logSize += data.size();
         if (settings.maxLogSize && logSize > settings.maxLogSize) {
@@ -1207,7 +1207,14 @@ void DerivationGoal::handleChildOutput(int fd, std::string_view data)
     if (hook && fd == hook->fromHook.readSide.get()) {
         for (auto c : data)
             if (c == '\n') {
-                handleJSONLogMessage(currentHookLine, worker.act, hook->activities, true);
+                auto s = handleJSONLogMessage(currentHookLine, worker.act, hook->activities, true);
+                if (s && !isWrittenToLog && logSink) {
+                    auto json = nlohmann::json::parse(std::string(currentHookLine, 5));
+                    if (json["type"] == resBuildLogLine) {
+                        auto f = json["fields"];
+                        (*logSink)((f.size() > 0 ? f.at(0).get<std::string>() : "") + "\n");
+                    }
+                }
                 currentHookLine.clear();
             } else
                 currentHookLine += c;
--- a/src/libutil/logging.hh
+++ b/src/libutil/logging.hh
@@ -4,7 +4,7 @@
 #include "error.hh"
 #include "config.hh"
 
-#include <nlohmann/json.hpp>
+#include <nlohmann/json_fwd.hpp>
 
 namespace nix {
 
--- a/src/libutil/logging.hh
+++ b/src/libutil/logging.hh
@@ -4,6 +4,8 @@
 #include "error.hh"
 #include "config.hh"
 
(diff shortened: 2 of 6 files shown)


src/libstore/build/derivation-goal.cc
```diff
@@ -1207,7 +1207,14 @@ void DerivationGoal::handleChildOutput(int fd, const string & data)
     if (hook && fd == hook->fromHook.readSide.get()) {
         for (auto c : data)
             if (c == '\n') {
-                handleJSONLogMessage(currentHookLine, worker.act, hook->activities, true);
+                auto s = handleJSONLogMessage(currentHookLine, worker.act, hook->activities, true);
+                if (s && !isWrittenToLog && logSink) {
+                    auto json = nlohmann::json::parse(std::string(currentHookLine, 5));
```
Ma27, line 1212: Note: we're effectively parsing each message twice with this patch. Any suggestions how to restructure the code to avoid that in a reasonable way (not too experienced with the codebase admittedly)? Or is this neglectable?

Ericson2314: Good catch, yes I think we should fix this but I'll have to check not ob my phone to see how.

Ma27:
@edolstra @Ericson2314 if we want to avoid the double-parsing of JSON messages, we should IMHO add another signature to `handleJSONLogMessage` to accept already parsed JSON values. Then we could parse the message once and pass the parsed data to `handleJSONLogMessage` then. If this suggestion makes sense to you, I'd alter the patch accordingly.

Anything else to be fixed before this can be merged? %)

Ma27:
cc @Ericson2314 updated accordingly :)

src/libstore/build/derivation-goal.cc
```diff
@@ -1207,12 +1207,14 @@ void DerivationGoal::handleChildOutput(int fd, const string & data)
...
             if (c == '\n') {
-                auto s = handleJSONLogMessage(currentHookLine, worker.act, hook->activities, true);
-                if (s && !isWrittenToLog && logSink) {
-                    auto json = nlohmann::json::parse(std::string(currentHookLine, 5));
-                    if (json["type"] == resBuildLogLine) {
-                        auto f = json["fields"];
-                        (*logSink)((f.size() > 0 ? f.at(0).get<std::string>() : "") + "\n");
+                auto json = parseJSONMessage(currentHookLine);
+                if (json) {
+                    auto s = handleJSONLogMessage(*json, worker.act, hook->activities, true);
```
Ericson2314, line 1212: Can you add a comment saying this is for the `ssh-ng://` remote builder case?

src/libstore/build/derivation-goal.cc
```diff
@@ -1207,7 +1207,16 @@ void DerivationGoal::handleChildOutput(int fd, const string & data)
     if (hook && fd == hook->fromHook.readSide.get()) {
         for (auto c : data)
             if (c == '\n') {
-                handleJSONLogMessage(currentHookLine, worker.act, hook->activities, true);
+                auto json = parseJSONMessage(currentHookLine);
+                if (json) {
+                    auto s = handleJSONLogMessage(*json, worker.act, hook->activities, true);
+                    if (s && !isWrittenToLog && logSink) {
```
Ericson2314, line 1213: Can you add a comment saying this is for the `ssh-ng://` remote builder case?

src/libstore/build/derivation-goal.cc
```diff
@@ -1175,10 +1175,10 @@ bool DerivationGoal::isReadDesc(int fd)
     return fd == hook->builderOut.readSide.get();
 }
 
-
 void DerivationGoal::handleChildOutput(int fd, const string & data)
 {
-    if (isReadDesc(fd))
+    auto isWrittenToLog = isReadDesc(fd);
+    if (isWrittenToLog)
```
Ericson2314, line 1181: Is this for just the `ssh://` case, or also local building? Can you add a code comment with the answer?

Ericson2314:
As I mentioned in chat, this looks good to me now except that it would be great to have a test.

Later, I am hoping we can make do this logging stuff a bit more abstractly / separate it from the rest of the scheduler, but I don't want that refactor to cause a regression!

Ma27:
> As I mentioned in chat, this looks good to me now except that it would be great to have a test.

Yup, I started hacking on that one yesterday, but got too tired at some point %)

Ma27:
@Ericson2314 updated accordingly and added a test that failed for me on master and passes on this branch :) 

Details are in the commit messages %)

symphorien:
If I read the code correctly, this stores the log locally as the remote sends it during build. So I am correct that contrary to the http store used for cache.nixos.org, if I build remotely with `ssh nix-store foo.nix` and then request the log locally `nix log foo.drv --option substituter ssh-ng://host` it won't fetch the log?

Ma27:
That's correct. But I interpreted #5079 to be exactly what my problem (i.e. `ssh-ng://host` as remote builder doesn't create files for `nix log`) was.

symphorien:
Quite ironically my actual use case when I opened this issue was "fetch logs from remote CI with nix log --option substituters ssh-ng://host". I wrote it in the issue differently as it looked easier to describe.

Ericson2314 (approved):
Thanks so much!! Looking forward to seeing this merged very much

Ericson2314:
This definitely fixes a bug; should we take the discussion of @symphorien actual issue back to #5079? I don't yet follow.

symphorien:
You are right, I opened another issue #6146

Ma27:
Rebased onto latest master to resolve a merge conflict.

ping @Ericson2314 @edolstra

src/libutil/logging.hh
```diff
@@ -4,6 +4,8 @@
 #include "error.hh"
 #include "config.hh"
 
+#include <nlohmann/json.hpp>
```
edolstra, line 7: We should never include `nlohmann/json.hpp` in any header file, since (IIRC) just including it adds about half a second of compilation time per compilation unit. If really needed, you can use `json_fwd.hpp`.

Ericson2314 (approved):

Ericson2314:
@Ma27 Please rebase again? MacOS issue is fixed now.

Ma27:
@Ericson2314 done! %)

Ericson2314:
Thanks!

Ericson2314:
@edolstra After a few round of review, I have approved this. I hope you have time to take a look soon!

thufschmitt (changes requested):
Thanks, looks good overall, just a few comments/suggestions

tests/build-remote.sh
```diff
@@ -54,6 +54,14 @@ nix path-info --store $TEST_ROOT/machine3 --all \
   | grep -v builder-build-remote-input-2.sh \
   | grep builder-build-remote-input-3.sh
 
+
+if [[ -z "$CONTENT_ADDRESSED" ]]; then
```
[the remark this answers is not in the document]
Ma27: My observations are here: https://github.com/NixOS/nix/pull/6029/commits/287642f132d8141269bd7c23b0bc7c1ee512b171

tests/build-remote.sh
```diff
@@ -54,6 +54,14 @@ nix path-info --store $TEST_ROOT/machine3 --all \
...
+  for i in input1 input3; do
+    drv="$(nix-instantiate $file -A passthru.$i --store $TEST_ROOT/machine0 --arg busybox $busybox)"
+    nix log --store $TEST_ROOT/machine0 "$drv"
```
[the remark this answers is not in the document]
Ma27: Fair, implemented!

Ma27:
ping @edolstra @thufschmitt

tests/build-remote.sh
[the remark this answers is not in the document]
Ericson2314: It sounds like the issue is that `nix log` perhaps cannot follow derivation resolutions to find the actual log? If so, that is indeed an orthogonal problem deserving a separate PR. 

We could open a fresh issue (after someone confirms my theory enough to open an issue) and link it in the code.

(I think commit messages are insufficient for this: commit messages describe how the old and new code relate, but the question of "why is this this skipped with CA" is a question directly about the new code itself.)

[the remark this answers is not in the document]
thufschmitt: Yep’, I think that’s the right issue. I’ve opened https://github.com/NixOS/nix/issues/6209 to track it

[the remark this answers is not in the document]
thufschmitt:
```suggestion
# Temporarily disabled because of https://github.com/NixOS/nix/issues/6209
if [[ -z "$CONTENT_ADDRESSED" ]]; then
```

thufschmitt (approved):

thufschmitt:
... and the CI is green, let’s merge. Thanks!

PR #6051, opened 2022-02-06
`nix log` should also work if the log didn't provide any output

Ma27:
According to `git bisect` this was introduced by https://github.com/NixOS/nix/commit/50a35860ee9237d341948437c5f70a7f0987d393

To summarize: while trying to update Hydra to use Nix 2.6, I realized that [`yath t/queue-runner/notifications.t`](https://github.com/NixOS/hydra/blob/master/t/queue-runner/notifications.t) is failing because it initially builds a derivation (and the builder doesn't write anything to `stdout`/`stderr`). After that it's checked if e.g. `nix log` succeeds.

While this was the case on the Nix 2.4pre rev of Hydra, it isn't the case anymore on Nix 2.6.

While bisecting I found the problem, it seems as if `init()` wrongly sets `archive_read_support_filter_all` if `raw` would be `true`.

Also added a regression-test that I used for bisecting.

cc @edolstra

### DIFF
--- a/src/libutil/tarfile.cc
+++ b/src/libutil/tarfile.cc
@@ -39,28 +39,30 @@ void TarArchive::check(int err, const std::string & reason)
         throw Error(reason, archive_error_string(this->archive));
 }
 
-TarArchive::TarArchive(Source & source, bool raw)
-    : source(&source), buffer(4096)
+TarArchive::TarArchive(Source & source, bool raw) : buffer(4096)
 {
-    init();
-    if (!raw)
+    this->archive = archive_read_new();
+    this->source = &source;
+
+    if (!raw) {
+        archive_read_support_filter_all(archive);
         archive_read_support_format_all(archive);
-    else
+    } else {
+        archive_read_support_filter_all(archive);
         archive_read_support_format_raw(archive);
+        archive_read_support_format_empty(archive);
+    }
     check(archive_read_open(archive, (void *)this, callback_open, callback_read, callback_close), "Failed to open archive (%s)");
 }
 
+
 TarArchive::TarArchive(const Path & path)
 {
-    init();
+    this->archive = archive_read_new();
+
+    archive_read_support_filter_all(archive);
     archive_read_support_format_all(archive);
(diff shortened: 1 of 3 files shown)


Ma27:
@edolstra rebased onto latest master to resolve the merge-conflict. Anything else to fix here? Would love to see this bugfix being merged :)

src/libutil/tarfile.cc
```diff
@@ -39,28 +39,30 @@ void TarArchive::check(int err, const std::string & reason)
...
-    : source(&source), buffer(4096)
+TarArchive::TarArchive(Source & source, bool raw) : buffer(4096)
 {
-    init();
-    if (!raw)
+    this->archive = archive_read_new();
+    this->source = &source;
+
+    if (!raw) {
+        archive_read_support_filter_all(archive);
```
tomberek, line 48: factor out?

Ma27:
Rebased onto latest master to resolve the merge conflict.

May I ask if there was any other reason that this didn't land in 2.7 other than the merge-conflict?

> factor out?

Personally I'm not a big fan of delaying (trivial) bugfixes with these kinds of requests, especially if the fix is a single revert. But if you want me to, why not, otherwise I'd suggest to do this in a different PR.

Ma27:
ping @edolstra @thufschmitt :)

Ma27:
cc @edolstra @thufschmitt anything missing here? :)

tests/log.sh
```diff
@@ -0,0 +1,11 @@
...
+
+builder="$(mktemp)"
+echo -e "#!/bin/sh\nmkdir \$out" > "$builder"
+outp="$(nix-build -E \
+    'with import ./config.nix; mkDerivation { name = "fnord"; builder = '"$builder"'; }' \
+    --out-link "$(mktemp -d)/result")"
+
+test -d "$outp"
+
+nix log "$outp"
```
edolstra, line 11: Maybe merge this into tests/logging.sh? It's a bit confusing to have log.sh and logging.sh.

edolstra (approved):

Ma27:
@edolstra done! (also rebased against master)

edolstra:
Thanks!

PR #6054, opened 2022-02-06
repl/load-flake: throw error if path isn't specified

lincolnauster:
In the repl, `:lf` with no arguments currently throws an error saying that `'' is not a valid URL`. This writes in a special case for an empty URL, giving the imo nicer error "cannot use `:load-flake` without a path specified. (Use . for the current working directory.)".

The other option would be to use the current directory as the flake path, but that could incur a lot of extra copying if the user is unaware of that behavior.

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -623,6 +623,9 @@ void NixRepl::loadFile(const Path & path)
 
 void NixRepl::loadFlake(const std::string & flakeRefS)
 {
+    if (flakeRefS.empty())
+        throw Error("cannot use `:load-flake` without a path specified. (Use . for the current working directory.)");
+
     auto flakeRef = parseFlakeRef(flakeRefS, absPath("."), true);
     if (evalSettings.pureEval && !flakeRef.input.isImmutable())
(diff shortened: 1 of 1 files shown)


lincolnauster:
Ah, hit enter too early - editing with actual details...

src/nix/repl.cc
```diff
@@ -623,6 +623,9 @@ void NixRepl::loadFile(const Path & path)
 
 void NixRepl::loadFlake(const std::string & flakeRefS)
 {
+    if (flakeRefS.empty())
+        throw Error("cannot use `:load-flake` without a path specified. (Use . for the current working directory.)");
```
edolstra, line 627:
```suggestion
        throw Error("cannot use ':load-flake' without a path specified. (Use '.' for the current working directory.)");
```

lincolnauster: ah, thanks for catching that!

lincolnauster:
tests pass on my machine - is the failure just spurious?

edolstra:
Yeah, that's a random macOS failure.

PR #6087, opened 2022-02-11
nix-fmt: add command

kamadorueda:
Demo: https://github.com/kamadorueda/nix-fmt-demo

demo

### DIFF
--- a/src/nix/fmt.cc
+++ b/src/nix/fmt.cc
@@ -1,0 +1,53 @@
+#include "command.hh"
+#include "run.hh"
+
+using namespace nix;
+
+struct CmdFmt : SourceExprCommand {
+    std::vector<std::string> args;
+
+    CmdFmt() { expectArgs({.label = "args", .handler = {&args}}); }
+
+    std::string description() override {
+        return "reformat your code in the standard style";
+    }
+
+    std::string doc() override {
+        return
+          #include "fmt.md"
+          ;
+    }
+
+    Category category() override { return catSecondary; }
+
+    Strings getDefaultFlakeAttrPaths() override {
+        return Strings{"formatter." + settings.thisSystem.get()};
+    }
+
+    Strings getDefaultFlakeAttrPathPrefixes() override { return Strings{}; }
+
+    void run(ref<Store> store) {
+        auto evalState = getEvalState();
+        auto evalStore = getEvalStore();
+
+        auto installable = parseInstallable(store, ".");
+        auto app = installable->toApp(*evalState).resolve(evalStore, store);
+
+        Strings programArgs{app.program};
+
+        // Propagate arguments from the CLI
+        if (args.empty()) {
+            // Format the current flake out of the box
+            programArgs.push_back(".");
+        } else {
+            // User wants more power, let them decide which paths to include/exclude
+            for (auto &i : args) {
+                programArgs.push_back(i);
+            }
+        }
+
+        runProgramInStore(store, app.program, programArgs);
+    };
+};
+
+static auto r2 = registerCommand<CmdFmt>("fmt");
--- a/src/nix/fmt.md
+++ b/src/nix/fmt.md
@@ -1,0 +1,53 @@
+R""(
+
+# Examples
+
+With [nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt):
+
+```nix
+# flake.nix
+{
+  outputs = { nixpkgs, self }: {
+    formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.nixpkgs-fmt;
+  };
+}
+```
+
+- Format the current flake: `$ nix fmt`
+
+- Format a specific folder or file: `$ nix fmt ./folder ./file.nix`
+
+With [nixfmt](https://github.com/serokell/nixfmt):
+
+```nix
+# flake.nix
+{
+  outputs = { nixpkgs, self }: {
+    formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.nixfmt;
+  };
+}
+```
+
+- Format specific files: `$ nix fmt ./file1.nix ./file2.nix`
+
+With [Alejandra](https://github.com/kamadorueda/alejandra):
+
+```nix
+# flake.nix
+{
+  outputs = { nixpkgs, self }: {
+    formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.alejandra;
+  };
+}
+```
+
+- Format the current flake: `$ nix fmt`
+
+- Format a specific folder or file: `$ nix fmt ./folder ./file.nix`
+
+# Description
+
+`nix fmt` will rewrite all Nix files (\*.nix) to a canonical format
+using the formatter specified in your flake.
+
+)""
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -527,6 +527,16 @@ struct CmdFlakeCheck : FlakeCommand
                             }
                         }
 
+                        else if (name == "formatter") {
+                            state->forceAttrs(vOutput, pos);
+                            for (auto & attr : *vOutput.attrs) {
+                                checkSystemName(attr.name, *attr.pos);
+                                checkApp(
+                                    fmt("%s.%s", name, attr.name),
+                                    *attr.value, *attr.pos);
+                            }
+                        }
+
                         else if (name == "packages" || name == "devShells") {
                             state->forceAttrs(vOutput, pos);
                             for (auto & attr : *vOutput.attrs) {
@@ -1010,6 +1020,7 @@ struct CmdFlakeShow : FlakeCommand, MixJSON
                     || (attrPath.size() == 1 && (
                             attrPath[0] == "defaultPackage"
                             || attrPath[0] == "devShell"
+                            || attrPath[0] == "formatter"
                             || attrPath[0] == "nixosConfigurations"
                             || attrPath[0] == "nixosModules"
                             || attrPath[0] == "defaultApp"
@@ -1059,6 +1070,7 @@ struct CmdFlakeShow : FlakeCommand, MixJSON
 
                 else if (
                     (attrPath.size() == 2 && attrPath[0] == "defaultApp") ||
+                    (attrPath.size() == 2 && attrPath[0] == "formatter") ||
                     (attrPath.size() == 3 && attrPath[0] == "apps"))
                 {
                     auto aType = visitor.maybeGetAttr("type");
--- a/tests/fmt.sh
+++ b/tests/fmt.sh
@@ -1,0 +1,28 @@
+source common.sh
+
+set -o pipefail
+
+clearStore
+rm -rf $TEST_HOME/.cache $TEST_HOME/.config $TEST_HOME/.local
+
+cp ./simple.nix ./simple.builder.sh ./fmt.simple.sh ./config.nix $TEST_HOME
+
+cd $TEST_HOME
+
+nix fmt --help | grep "Format"
+
+cat << EOF > flake.nix
+{
+  outputs = _: {
+    formatter.$system = {
+      type = "app";
+      program = ./fmt.simple.sh;
+    };
+  };
+}
+EOF
+nix fmt ./file ./folder | grep 'Formatting: ./file ./folder'
+nix flake check
+nix flake show | grep -P 'x86_64-linux|x86_64-darwin'
+
+clearStore
--- a/tests/fmt.simple.sh
+++ b/tests/fmt.simple.sh
@@ -1,0 +1,1 @@
+echo Formatting: "${@}"
--- a/tests/local.mk
+++ b/tests/local.mk
@@ -77,6 +77,7 @@ nix_tests = \
   post-hook.sh \
   function-trace.sh \
   flake-local-settings.sh \
+  fmt.sh \
   eval-store.sh \
   why-depends.sh \
   import-derivation.sh \


Ma27:
Tbh I don't see much value over `nix run <flake-url for $formatter> -- $formatter ./folder`.

kamadorueda:
> Tbh I don't see much value over `nix run <flake-url for $formatter> -- $formatter ./folder`.

See it from the perspective that this is just the beginning 

There are some phases (we can swap points 2 and 3):
1. This PR 
1. RFC 0101 decides a tool, so we set a default for --formatter and thus the command is simplified to a simple `$ nix fmt $paths`, that would be much shorter than `$ nix run long-flake-uri -- $more_long_args` and therefore has better UX
1. We wrap the tool that RFC0101 decided, so for instance running `$ nix fmt` without arguments formats the current flake (as cargo fmt), and we can also add other niceties, which further improves UX

To me this is a good thing

src/nix/fmt.cc
```diff
@@ -0,0 +1,73 @@
+#include "command.hh"
+#include "run.hh"
+
+using namespace nix;
+
+struct CmdFmt : SourceExprCommand {
```
andir, line 6: This should require the experimental flakes feature or support non-flakes use cases.

kamadorueda: It will ask for the flakes flag before getting there

Other commands have the same behaviour and UX

```
[kamadorueda@nixos:/data/nix]$ ./outputs/out/bin/nix fmt

[kamadorueda@nixos:/data/nix]$ ./outputs/out/bin/nix --extra-experimental-features nix-command fmt
error: you must pass '--formatter <flake-uri>'
Try './outputs/out/bin/nix --help' for more information.

[kamadorueda@nixos:/data/nix]$ ./outputs/out/bin/nix --extra-experimental-features nix-command fmt --formatter nixpkgs#nixpkgs-fmt
error: experimental Nix feature 'flakes' is disabled; use '--extra-experimental-features flakes' to override

[kamadorueda@nixos:/data/nix]$ ./outputs/out/bin/nix --extra-experimental-features nix-command --extra-experimental-features flakes fmt --formatter nixpkgs#nixpkgs-fmt
(works)
```

It's intentional to make it work with flakes only, as other commands like `nix bundle` do

There is also a warning about experimental stuff:

image

or maybe I'm not understanding your suggestion, could you please help me understand?

terlar:
Perhaps the default formatter could also be configured via nix.conf. That would make the usability nice before community adapts a default formatter

SuperSandro2000:
> Perhaps the default formatter could also be configured via nix.conf. That would make the usability nice before community adapts a default formatter

Without such feature this subcommand does not make any sense because you could just strip everything before the repository name and would achieve the same result.

kamadorueda:
> Perhaps the default formatter could also be configured via nix.conf. That would make the usability nice before community adapts a default formatter

This is a very good idea, let's do it!

ElvishJerricco:
> RFC 0101 decides a tool, so we set a default for --formatter and thus the command is simplified to a simple $ nix fmt $paths, that would be much shorter than $ nix run long-flake-uri -- $more_long_args and therefore has better UX

So we intend to have a subcommand that automatically downloads and runs another program? Won't this likely cause fairly large downloads, as there's no reason to expect the formatter's flake to be using the same version of nixpkgs as the user? Or if we do get it to use the user's nixpkgs, then we have no expectation that it'll be cached, or even successfully build. And finally... implicitly downloading and running binaries for a command that appears as straightforward as `nix fmt` seems incredibly dangerous.

kamadorueda:
> So we intend to have a subcommand that automatically downloads and runs another program? Won't this likely cause fairly large downloads, as there's no reason to expect the formatter's flake to be using the same version of nixpkgs as the user? Or if we do get it to use the user's nixpkgs, then we have no expectation that it'll be cached, or even successfully build. And finally... implicitly downloading and running binaries for a command that appears as straightforward as `nix fmt` seems incredibly dangerous.

It seems that since three years ago the purpose is to ship such tool as part of Nix, so in the future we would be safe, cached and fast:

> edolstra- Great contributor, [May '19](https://discourse.nixos.org/t/job-implement-a-nix-fmt-formatter/2819/2?u=kamadorueda)
> Another goal to consider:
> 
> Write it in C++ or Rust so it can be integrated in (i.e. shipped as part of) Nix. Something written in Haskell or most other languages will pull in a lot of additional dependencies that would make this infeasible.

And in the present we are covered as well, there is nothing implicit as you mention, since users explicitly decide which tool to use (--formatter / nix.conf), the source code for the documented tools is available for anyone to read, audit, inspect and finally trust and use, and the flake-uri can be pinned to a git commit, branch and hash for integrity so that no malicious actor in this world can change the contents after you perform the exhaustive source code review. This is the same security model that Nixpkgs uses.

As per the download size in the present, yeah, this one will have to wait until we have a winner, so that we known which tool to embed directly into Nix

So in summary, we cool in the future, we cool (except download size) in the present, no need to worry I believe.

kamadorueda:
> But you said there would be a default. That's the main thing I found scary

Take into account that such default, does not mean pointing to `main`/`master`, it can be a default pointing to the commit, branch, hash. But anyway keep reading, it gets better

> Requiring yet another tool to be audited does not sound like a selling point to me

What's the difference between auditing 4000 lines of code already written on repo X, vs copy-pasting and auditing those 4000 lines of code on nixos/nix, or worst, vs re-writting those 4000 lines of code from scratch here? 

But sure, the copy paste is possible, we already copy-paste a few things like [toml11](https://github.com/NixOS/nix/commit/c3f68b5db7b9d4b963102a2abc20e20445a8bec5). In any place, we will need to audit, and the most economic mechanism seems to be just referencing upstream

I'm happy with any method anyway, and licenses of the formatters are very permissive on purpose as well, so we cool, plenty of options

---

But I understand the feedback from the various comments, thanks about that!

I suppose a more proper way forward is (points 2 and 3 can be swaped):

1. Improving this PR:
    - nothing implicit (done), --formatter is mandatory without a default
    - add nix.conf support, so that there is a (big) difference with a plain `$ nix run`, plus, people can configure it via nixConf in the flake.nix, so that you can define the formatter for the project in the very same flake.nix (this is an excellent UX) !!
2. Wait for RFC 101 to decide a tool, then we audit such tool, and we choose from any of the following mechanisms:
    - hard-code flake-uri pointing to upstream repo, commit and hash, ensuring there are substitutable outputs on cache.nixos.org (Alejandra is just 1.6MB when statically linked, so no big downloads will happen anyway)
    - referencing upstream, but compile and linking together with Nix so closure is minimal
    - copy-paste upstream code here, and compile and link together with Nix so closure is minimal
    - whatever really, there is no perfect solution, just trade-offs and preferences, we can decide later
3. Enhance interactions between `$ nix fmt` and the underlying tool implementation, so for instance running `$ nix fmt` without arguments formats the current flake (as cargo fmt), and we can also add other niceties like git-ignoring files, to further improve UX

Am I missing something? I believe this cover the overall feedback so far

There is something that I'm sure though, and is that we can deliver something that is: safe, secure, light-weight, and better than `$ nix run $long_flake_uri -- $more_args`, I'm sure of that

Thanks for helping me shape the future of this thing!

src/nix/fmt.md
```diff
@@ -0,0 +1,30 @@
+R""(
+
+# Examples
+
+- Format one or more files or directories:
+
+  With [nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt):
+
+  ```console
+  # nix fmt --formatter github:nixpkgs#nixpkgs-fmt ./folder ./file.nix
```
zimbatm, line 10: Is there a mechanism to select a default formatter for a project?

I don't really get how the current examples are different than `nix run github:nixpkgs#nixpkgs-fmt ./folder ./file.nix` for example.

garbas: as with any `nix.conf` value you should be able to set it via `nixConfig`. At least that is how I understand it. @kamadorueda correct?

```nix
# flake.nix
{
  description = "example";
  nixConfig.formatter = "github:foo/bar/commit?hash=123";
}
```

roberth: `?hash` is problematic as it suggests that we'll have to reinvent `inputs` for it.

It being outside `outputs` also means that it can't be abstracted over. I'd like to manage tooling configuration in a central flake and reference that in other projects.

Using `outputs.nixFormatter.${system}` will allow that.

It _must not_ be a `nix.conf` option, because that will inevitably lead to formatting churn as contributors with different versions reformat code in projects that do not pin the formatter. For the same reason, it must not be hardcoded in `nix`, because not everyone runs the same Nix version.

roberth:
I have some serious doubts about the usefulness of this command as the same can already be achieved better with tools like `pre-commit`, `treefmt` and `nixpkgs-fmt` itself.

Adding this code means that it has to be maintained and improved over time, *taking time away from other Nix features' maintenance.*

I can see a future where a command like this is an appropriate addition, but it is a distant one.

edolstra:
> I have some serious doubts about the usefulness of this command as the same can already be achieved better with tools like pre-commit, treefmt and nixpkgs-fmt itself.

The UX advantage is the same as commands like `cargo fmt`, namely that it provides a formatter that just works and is available to all Nix users without having to do any additional setup/configuration.

> Adding this code means that it has to be maintained and improved over time, taking time away from other Nix features' maintenance.

It doesn't, because the actual formatter isn't part of Nix. It's just a convenience command in `nix` to call an external formatter.

edolstra:
Ideally, `nix fmt` without any arguments should format all `.nix` files in the flake in the current directory, similar to `cargo fmt`. But that can be done in a future PR.

roberth:
> It's just a convenience command in `nix` to call an external formatter.

That doesn't imply that it doesn't need maintenance. Considering the lengthy discussion, there's a lot we can get wrong, and I am under no illusion that this PR will be the final implementation.

Glue code often looks innocent, but it is not.

I like the UX improvement and I'm not going to stop you. I just think there are more urgent issues, is all.

I will urge you to _never_ invoke a formatter version that isn't pinned alongside the files it formats.
No flake? No `fmt`. No pin in the flake? No `fmt`. See also https://github.com/NixOS/nix/pull/6087#discussion_r809009262

edolstra:
> I will urge you to never invoke a formatter version that isn't pinned alongside the files it formats.

I'm not sure whether pinning is essential. E.g. `cargo fmt` doesn't require the formatter to be pinned in `Cargo.lock`.

tomberek:
Perhaps: (due to https://github.com/NixOS/nix/issues/5532)
```
formatter.x86_64-linux.default
```
Not sure if this is needed, but raising this as a possibility. Should probably leave multiple languages support to something like treefmt, but we don't know what the future holds.

Ericson2314:
Honestly....stuff like this is why I want flakes factored into another layer. If we just throw in every convenient thing thing to Nix will will end up with a ginormous repo we have no hope of maintaining.

But we already have command plugins, so it would be very easy to add custom comands, and with good layering, we can have like a "batteries included" super plugin thingy that lives somewhere else than Nix itself.

edolstra:
> Honestly....stuff like this is why I want flakes factored into another layer. If we just throw in every convenient thing thing to Nix will will end up with a ginormous repo we have no hope of maintaining.

Except that we're *not* adding the formatter to the Nix repo. We're just adding a convenience command that fetches and runs the formatter.

src/nix/fmt.cc
```diff
@@ -0,0 +1,53 @@
...
+  std::vector<std::string> args;
+
+  CmdFmt() { expectArgs({.label = "args", .handler = {&args}}); }
+
+  std::string description() override {
+    return "reformat your code in the standard style";
+  }
```
edolstra, line 13: Speaking of formatting, please use the Nix formatting style :-)

E.g.

```suggestion
    std::string description() override 
    {
        return "reformat your code in the standard style";
    }
```

edolstra:
I think the interface should be more flake-centric. I.e. instead of 
```
{
  outputs = { nixpkgs, self }: {
    formatter.x86_64-linux =
      nixpkgs.legacyPackages.x86_64-linux.alejandra;
  };
}
```
do something similar to what `nix bundle` does, i.e. run `formatters.<system>.default` attribute from a specified flake. E.g.
```
{
  nixConfig.formatter = "github:kamadorueda/alejandra";
  # or: nixConfig.formatter = "github:kamadorueda/alejandra#some-formatter";
  outputs = ...;
}
```

Or if we want pinning:
```
{
  inputs.formatter.url = "github:kamadorueda/alejandra";
  outputs = ...;
}
```

roberth:
> more flake-centric

What I'm seeing in your suggestions is "more flake-restricted". If this functionality is only accessible through non-output attributes, you're removing most of the flexibility afforded by the Nix language.


> ```nix
> {
>   nixConfig.formatter = "github:kamadorueda/alejandra";
>   # or: nixConfig.formatter = "github:kamadorueda/alejandra#some-formatter";
>   outputs = ...;
> }
> ```

This might be ok _as a fallback_, if no flake-specific settings were found. (It'd work outside of a flake too, I presume)


> ```nix
> {
>   inputs.formatter.url = "github:kamadorueda/alejandra";
>   outputs = ...;
> }
> ```

Indeed pinning would be best practice. I'd be ok with this as a fallback, even if I don't like the idea of magic input names.


> ```nix
> {
>   outputs = { nixpkgs, self }: {
>     formatter.x86_64-linux =
>       nixpkgs.legacyPackages.x86_64-linux.alejandra;
>   };
> }
> ```

This is still the most powerful method. It allows the formatter to be defined in an input, in an input that isn't a dedicated formatter flake, like Nixpkgs, in the flake itself, or in a configuration helper flake, such as in a [flake-modules-core](https://github.com/hercules-ci/flake-modules-core) module.

kamadorueda:
I just pushed some changes to:
- Support `nix flake check`
- Support `nix flake show`
- Use c++ indentation with 4 spaces
- The formatter should be an application and not a derivation

I still believe having the formatter as an output is better because:
- it allows pinning. `inputs.formatter` also allows pinning but is a magic name, which can trigger surprising behaviours and magic interactions 
- it allows discoverability, since it will be printed at `nix flake show` time, to the side of other packages, dev shells and apps
- it allows for security.  `nix.conf` is not pinned in the flake.lock and therefore is susceptible to supply chain attacks
- it can be archived,  while `nix.conf` cannot
- it explicitly declares it works on a set of systems. `inputs.formatter` does not encapsulate this concept, which can make users wonder "I added the input, why it does not work"
- (arguably) it allows per-project formatters (hermeticity), while `nix.conf` may (by default) cause things like "code churn because one user is changing the style because its nix.conf formatter is a different version than the version of other users in the same project"

Proposed schema as of 2022-03-11:

```nix
# $ nix fmt
{
  outputs = {nixpkgs, ...}: {
    formatter.x86_64-linux = {
      type = "app";
      program = "${nixpkgs.legacyPackages.x86_64-linux.alejandra}/bin/alejandra";
    };
  };
}
```

tomberek:
I think the shape of the output should be as close as possible to the rest of the schema:
```
formatter.x86_64-linux.default
```
or 
```
apps.x86_64-linux.formatter
```
because just `formatter.x86_64-linux` is essentially the same as `defaultFormatter.x86_64-linux` which is a pattern just removed in 2.7: https://github.com/NixOS/nix/pull/6089

roberth:
> ```
> formatter.x86_64-linux.default
> ```

I don't think we have any intent of supporting multiple formatters and I don't think we should prioritize consistency over truthfulness.

I'll admit that I do like that I can return a bunch of thunks at `formatter.${whatever}` and then lazily determine "oh, we don't actually have a formatter that supports this `system`" and return a `{ }`. The same can be achieved by tolerating `null`s though, which is almost as good.


> ```
> apps.x86_64-linux.formatter
> ```

As apps are (primarily?) for external use, this suggests that the formatter is part of the purpose of the flake. Also the name `formatter` seems too generic for this context; it may collide and it isn't clear that it's for `nix fmt`.

Maybe there's a case for generic "development commands" that are for local use. These then won't pollute the `apps` namespace. Perhaps `nix dev <attr> <args>`? I'm writing an issue for that.

roberth:
Ok, wrote #6241 for the missing piece to the puzzle.

Now suggesting

    devTools.<system>.nixFormatter

Let me dissect that.
 - `devTools` because a formatter isn't the only dev tool #6241
 - `nix` to make clear that this is for formatting nix files
 - camelCase because it's not a package name

edolstra:
Thanks, merged! 

Note that because of #6389, formatters can no longer be both apps or packages, so I've changed it to packages for now.

PR #6088, opened 2022-02-11
hacking.md: allow for easier copying

tomberek:


edolstra:
Hm, not really in favor of this. I think it's inconsistent with most of the manual, where we show prompts explicitly to disambiguate between the commands entered by the user and the output of the commands.

PR #6089, opened 2022-02-11
Replace defaultBla.$system with bla.$system.default

### LINKED ISSUE #5532: Rename `defaultX` flake outputs to `X.default`

edolstra:
This also simplifies some `InstallableFlake` logic and fixes `nix bundle` parsing its installable twice.
    
Fixes #5532.

## TODO
- [x] make `nix flake check` warn about the old `defaultX` outputs.
- [x] update documentation
- [x] release note (bonus for examples)

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -1,5 +1,18 @@
 # Release X.Y (202?-??-??)
 
+* A number of "default" flake output attributes have been
+  renamed. These are:
+
+  * `defaultPackage.<system>` → `packages.<system>.default`
+  * `defaultApps.<system>` → `apps.<system>.default`
+  * `defaultTemplate` → `templates.default`
+  * `defaultBundler.<system>` → `bundlers.<system>.default`
+  * `overlay` → `overlays.default`
+  * `devShell.<system>` → `devShells.<system>.default`
+
+  The old flake output attributes still work, but `nix flake check`
+  will warn about them.
+
 * `nix bundle` breaking API change now supports bundlers of the form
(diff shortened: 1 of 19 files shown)


Pacman99:
How come this doesn't change `nixosModule` to `nixosModules.default`? Thats similar enough to `overlay` since their both not system-spaced and nix functions, but only `overlay` was changed not `nixosModule`.

rehno-lindeque:
I have the same question. I think a nice pattern for many people would be:

```nix
nixosModules = rec {
  default = {
    imports = [
      moduleA
      moduleB
      moduleC
    ];
  };
  moduleA = import ./nixos-modules/module-a;
  moduleB = import ./nixos-modules/module-b;
  moduleC = import ./nixos-modules/module-c;
}
```

Issue #5979, opened 2022-01-25
Support arbitrary fetching of files as Flake inputs


MatthewCroughan:
**Is your feature request related to a problem? Please describe.**
I want to make a flake that fetches people's keys from Github. It looks like this, but I cannot because fetching arbitrary files is not supported via flake inputs:
```nix
{
  inputs.matthewcroughanKeys = {
    url = "https://github.com/matthewcroughan.keys";
    flake = false;
  };
  outputs = { self, matthewcroughanKeys }: {
    myKey = matthewcroughanKeys.outPath;
  };
}
```

**Describe the solution you'd like**
I want flake inputs to support fetching arbitrary files and locking their hash in the `flake.lock`

**Describe alternatives you've considered**
Using a sha256 rather than `nix flake update` to manually fetch keys, maintaining a long list of sha256s for each person. Or aggregating the hashes of many keys in a set of fetchurl calls.

**Additional context**
https://discourse.nixos.org/t/fetching-ssh-public-keys/12076

tomberek:
Other use cases:
- fetch a patch and locking it automatically in a TOFU way via flake.lock
- fetch a remote resource and purposefully error out if it changes

One can currently point to a local path using `path:./somefile` so being unable to do this for remote seems to be an accidental restriction. Ugly hack: https://github.com/tomberek/system-arg/blob/master/systemFile/flake.nix#L4  (my intention is not specifically about the system parameter, but it is an example)

tomberek:
This seems to be working: https://github.com/tomberek/nix/tree/fetcher_url It should behave the same as the tarball fetcher, but with opposite boolean logic:

> In URL form, the schema must be http://, https:// or file:// URLs and the extension must NOT be .zip, .tar, .tar.gz, .tar.xz, .tar.bz2 or .tar.zst.

Kranzes:
When fetching keys from github, won't the hash change each time a key is added/removed, since it doesn't have a commit/rev?

tomberek:
> When fetching keys from github, won't the hash change each time a key is added/removed, since it doesn't have a commit/rev?

That particular example will have that annoyance, so it may not be the best use-case or motivating example. But generally being able to fetch a file instead of only trees is potentially useful.

Kranzes:
[...]

PR #6094, opened 2022-02-13
fetchers: able to fetch a single remote file instead of a tree

tomberek:
Proof-of-concept for #5979

Though it would be better to change tarball.cc into something like remote.cc and have an "unpack" boolean somewhere, resulting in far less code duplication.


tomberek:
stale and OBE, reopen if needed

Issue #4503, opened 2021-02-01
Follow up message when creating a new project


garbas:
At the time of writting when you create a new project using either `nix flake new project-A` or `nix flake init` nothing is displayed to the output.

The following content should be presented to the user:
- listing all the files that were created,
- link to flake.nix schema and tutorial how to configure a flake project,
- a message (that should come from a template itself) how to use this template and what are the first steps.

/cc @edolstra

tomberek:
For the written files?
```
diff --git a/src/nix/flake.cc b/src/nix/flake.cc
index 64fcfc000..cbd5ab1de 100644
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -682,6 +682,7 @@ struct CmdFlakeInitCommon : virtual Args, EvalCommand
                 else
                     throw Error("file '%s' has unsupported type", from2);
                 files.push_back(to2);
+                notice("wrote: %s",to2);
             }
         };
```

Link to https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#flake-format ? In other messages we link to the stable manual, so this would be a bit annoying.

For a "template message" there doesn't seem to be anything in https://github.com/NixOS/templates at the moment to utilize. Where would such a thing go? 
- A README.md could work at the cost of additional conflicts with existing files. 
- A README-flake.md wouldn't conflict as often
- detecting a README.md in a template and printing it when the user calls "init".

tomberek:
- optional implementation, readFile and trace in the templates?

tomberek:
Use welcomeText in the template schema.

tomberek:
Fixed by #6013

PR #6103, opened 2022-02-15
add descriptive output when creating templates

Radvendii:
this includes a `welcomeText` attribute which can be set in the
template, as well as outputing which files were created.

#4503

### DIFF
--- a/src/nix/flake-init.md
+++ b/src/nix/flake-init.md
@@ -37,6 +37,10 @@ A flake can declare templates through its `templates` and
 
 * `path`: The path of the directory to be copied.
 
+* `welcomeText`: A block of text to display when a user initializes a new flake
+  based on this template.
+
+
 Here is an example:
 
 ```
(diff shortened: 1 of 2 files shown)


edolstra:
I think the `welcomeText` should be in Markdown. It can be rendered easily by calling `renderMarkdownToTerminal()`.

src/nix/flake-init.md
```diff
@@ -45,6 +49,10 @@ outputs = { self }: {
   templates.rust = {
     path = ./rust;
     description = "A simple Rust/Cargo project";
+    welcomeText = ''
+      You've created a simple Rust/Cargo template.
+      Visit https://www.rust-lang.org/ for more info.
```
roberth, line 54: I don't think this is a useful link. It should point to a Nix-specific document.

I understand that it's just an example, but we shouldn't set a wrong precedent.

PR #6106, opened 2022-02-16
flake: Print removed inputs in yellow

lorenzleutgeb:
Red probably was chosen since this is somewhat related to `diff`, but for inputs. I still think that users will not grasp this, especially not when there is no input added in the same run that would provide a green line for contrast.

Instead, I'd argue that users will think that red indicates an error. Removal of an input however is not an error per se, but deserves a warning.

If you want to keep the green/red scheme, consider prepending this with a message like "Differences in inputs follows (<green>added</green>, <red>removed</red>):".


edolstra:
This doesn't compile because `ANSI_YELLOW` no longer exists. You can use `ANSI_WARNING` instead (which is magenta because yellow is hard to read for people with a white terminal background).

tomberek:

PR #6110, opened 2022-02-17
Create daemon-socket folder during install

### LINKED ISSUE #5909: Missing /nix/var/nix/daemon-socket when installed on Debian Sid
**Describe the bug**

When Nix is installed on Debian Sid, the file `/nix/var/nix/daemon-socket` is missing and the systemd daemon cannot start.

**Steps To Reproduce**

1. Install a fresh and minimal Debian
2. Setup Sid repositories and update/dist-upgrade
3. `sudo apt-get install --yes ca-certificates curl sudo systemd tar xz-utils` as sudo user
4. `sh <(curl -L https://nixos.org/nix/install) --daemon` as sudo user
5. Reboot

**Expected behavior**

`nix-daemon` running.

**Additional context**

As I test on fresh install, I get last versions of all, as today:
systemd version: 250.2-2
nix version: 2.5.1

Note that it works properly on Debian Bullseye, where:
systemd version: 247.3-6
nix version: 2.5.1

layus:
Fixes #5909. See comments there for details about the failure mode uncovered with systemd >= 250.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -609,7 +609,7 @@ EOF
         fi
     fi
     _sudo "to make the basic directory structure of Nix (part 1)" \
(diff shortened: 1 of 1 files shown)


garbas (approved):

garbas:
@edolstra Could we get this fix also in 2.6.1, or is it too late already?

PR #6119, opened 2022-02-18
libfetchers/path: set `lastModified` to path's mtime

Ma27:
When importing e.g. a local `nixpkgs` in a flake to test a change like

    {
      inputs.nixpkgs.url = path:/home/ma27/Projects/nixpkgs;
      outputs = /* ... */
    }

then the input is missing a `lastModified`-field that's e.g. used in
`nixpkgs.lib.nixosSystem`. Due to the missing `lastMoified`-field, the
mtime is set to 19700101:

    result -> /nix/store/b7dg1lmmsill2rsgyv2w7b6cnmixkvc1-nixos-system-nixos-22.05.19700101.dirty

With this change, the `path`-fetcher now sets a `lastModified` attribute
to the `mtime` just like it's the case in the `tarball`-fetcher already.
When building NixOS systems with `nixpkgs` being a `path`-input and this
patch, the output-path now looks like this:

    result -> /nix/store/ld2qf9c1s98dxmiwcaq5vn9k5ylzrm1s-nixos-system-nixos-22.05.20220217.dirty

cc @edolstra

### DIFF
--- a/src/libfetchers/path.cc
+++ b/src/libfetchers/path.cc
@@ -1,5 +1,7 @@
 #include "fetchers.hh"
 #include "store-api.hh"
+#include <filesystem>
+#include "archive.hh"
 
 namespace nix::fetchers {
 
@@ -80,8 +82,9 @@ struct PathInputScheme : InputScheme
         // nothing to do
     }
 
-    std::pair<StorePath, Input> fetch(ref<Store> store, const Input & input) override
+    std::pair<StorePath, Input> fetch(ref<Store> store, const Input & _input) override
     {
+        Input input(_input);
         std::string absPath;
         auto path = getStrAttr(input.attrs, "path");
 
@@ -111,9 +114,15 @@ struct PathInputScheme : InputScheme
         if (storePath)
             store->addTempRoot(*storePath);
 
-        if (!storePath || storePath->name() != "source" || !store->isValidPath(*storePath))
+        time_t mtime = 0;
+        if (!storePath || storePath->name() != "source" || !store->isValidPath(*storePath)) {
             // FIXME: try to substitute storePath.
-            storePath = store->addToStore("source", absPath);
+            auto src = sinkToSource([&](Sink & sink) {
+                mtime = dumpPathAndGetMtime(absPath, sink, defaultPathFilter);
+            });
+            storePath = store->addToStoreFromDump(*src, "source");
+        }
+        input.attrs.insert_or_assign("lastModified", uint64_t(mtime));
 
         return {std::move(*storePath), input};
     }
--- a/src/libfetchers/path.cc
+++ b/src/libfetchers/path.cc
@@ -1,6 +1,5 @@
 #include "fetchers.hh"
 #include "store-api.hh"
-#include <filesystem>
 #include "archive.hh"
 
 namespace nix::fetchers {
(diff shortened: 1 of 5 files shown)


Ma27:
...though it might actually be better to find the last modification inside the tree.... I'll check if that's feasible.

Ma27:
Any suggestions on how to get this to build on macos? %)

thufschmitt:
> Any suggestions on how to get this to build on macos? %)

You need to link against `c++fs` to use the `filepath` library on old clang (like [here](https://github.com/NixOS/nix/pull/5380/files#diff-568140f0a437f4a4742c695849d82368b69abf332616acae6fb3f51693e123c9R9-R11))

src/libfetchers/path.cc
```diff
@@ -108,6 +110,18 @@ struct PathInputScheme : InputScheme
         // FIXME: check whether access to 'path' is allowed.
         auto storePath = store->maybeParseStorePath(absPath);
 
+        auto srcSt = lstat(absPath);
```
thufschmitt, line 113: Since you’re using the filesystem api, it might be nicer to use it all the way down and replace `lstat` by something like [last_write_time](https://en.cppreference.com/w/cpp/filesystem/last_write_time)

(depends how clunky it is obviously. I didn’t really try using it)

edolstra:
I'm not really in favor of doing a second traversal of the input since it's inefficient/inelegant (and will give an incorrect result if we ever add filtering support). It would be cleaner to make `dumpPath()` return the maximum `lastModified` date, and use `addToStoreFromDump()` instead of `addToStore(path)`.

src/libfetchers/path.cc
```diff
@@ -80,8 +81,9 @@ struct PathInputScheme : InputScheme
         // nothing to do
     }
 
-    std::pair<StorePath, Input> fetch(ref<Store> store, const Input & input) override
+    std::pair<StorePath, Input> fetch(ref<Store> store, const Input & _input) override
     {
+        Input input(_input);
```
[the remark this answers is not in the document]
Ma27: This doesn't compile as `fetch` as a signature with a `const` input. In fact, other fetchers use the same workaround as I did here.

src/libfetchers/path.cc
Ma27: I implemented it in a different fashion now as @edolstra suggested.

Ma27:
@edolstra @thufschmitt the `path` fetcher now uses `addToStoreFromDump` with the `dump()` method now providing the mtime for a source :)

Ma27:
ping @edolstra @thufschmitt

Ma27:
@thufschmitt rebased onto latest master and implemented a small test. This doesn't check for the full date, but only for 2222-11-1* since lastModified is converted in `fetchTree` to UTC.

thufschmitt:
> @thufschmitt rebased onto latest master and implemented a small test. This doesn't check for the full date, but only for 2222-11-1* since lastModified is converted in `fetchTree` to UTC.

Ah indeed good catch :)

thufschmitt (approved):

thufschmitt:
Mh, this is breaking the i686 builds: https://hydra.nixos.org/eval/1749090 . I guess 2222 is too much for a 32 bit timestamp :upside_down_face:

PR #6134, opened 2022-02-20
Factoring out parts of the serve protocol for Hydra to share

Ericson2314:
Hydra uses the lock argument differently than Nix, so we un-hard-code it.

See https://github.com/NixOS/hydra/pull/1165, which takes advantage of this (and #9560)

Depends on #9560


src/libstore/serve-protocol-impl.hh
```diff
@@ -0,0 +1,25 @@
...
+namespace serve_proto {
+
+struct BasicConnection
+{
+    FdSink to;
+    FdSource from;
+    int remoteVersion;
+};
+
+StorePathSet queryValidPaths(
```
thufschmitt, line 19: Shouldn’t this be made a method of `BasicConnection` rather than a standalone function?

thufschmitt (commented):
There might be a very good reason that I’m missing, but wouldn’t it be simpler to directly use the legacy ssh store in hydra? That would also make it possible to incrementally make hydra store-agnostic by extending the store API to fit its needs

src/libstore/serve-protocol-impl.hh
Ericson2314: Sure. I put in a namespace like the `worker_proto` one we used earlier. Since this is sort of the "raw, secret" stuff, something about making it bare functions felt stylistically appropriate, but doesn't matter.

Ericson2314: Eh, I sort of feel in an "OOP is not actually good" way that it is best to just make methods functions which are intrinsically about the data type itself, and this feels to me more like something one does with the data type.

Let me know what you want @thufschmitt. If you still disagree I'll change it, but I would like this to not drag on, so Hydra can keep on being converted.

Ericson2314:
CI is not working because recent changes broke the macOS / clang builds.

Ericson2314:
@thufschmitt This has been rebaced.

Ericson2314:
Re what @thufschmitt wrote in https://github.com/NixOS/nix/pull/6134#issuecomment-1079199888:

Some steps like https://github.com/NixOS/hydra/pull/1180 are now done!

The big difference which makes it hard to reuse the `LegacySSHStore` as-is in Hydra is that every Nix operation grabs a new connection from the pool, whereas Hydra is resuing the same connection. And I am afraid this may make a difference for things like temp roots. So I think the sort of functions I am breaking out are still useful as a first more conserative step.

That said, I would be OK making `BasicClientConnection` part of `LegacySSHStore` and exposed `LegacySSHStore` as you say. (This would be similar to this PR but keeping more code in `legacy-ssh-store.cc`, where it is today, just moving it into these new functions.) How does that sound?

Ericson2314:
team meeting notes:

- The "slow incremental migration" process is approved by the team (i.e. doing all these steps). That is on me torturing myself with many intermediate steps, shouldn't burden the team. When everything is done, and hydra isn't even using the legacy protocol any more, we should arrive at the same destination as if we did the "fast jump migration" process.

- Still blocked on @thufschmitt taking a look because his concerns from a year ago that may or may not apply.

Ericson2314:
OK now this is rebased on top of the #9586 so we have the *option* of making these things part of `LegacySSHStore` for Hydra if @thufschmitt prefers!

Ericson2314:
(in/after the meeting @thufschmitt said he didn't care about reviewing this one in particular.)

src/libstore/legacy-ssh-store.cc
```diff
@@ -96,28 +61,21 @@ ref<LegacySSHStore::Connection> LegacySSHStore::openConnection()
...
-               close it. */
-            conn->sshConn->in.close();
-            auto msg = conn->from.drain();
-            throw Error("'nix-store --serve' protocol mismatch from '%s', got '%s'",
-                host, chomp(saved.s + msg));
+        conn->remoteVersion = ServeProto::BasicClientConnection::handshake(
+            conn->to, tee, SERVE_PROTOCOL_VERSION, host);
+    } catch (SerialisationError & e) {
+        /* In case the other side is waiting for our input,
+           close it. */
```
roberth, line 71: nitpick:

```suggestion
        // in.close(): Don't let the remote block on us not writing.
```

A bit more explicit, on 1 line.
We generally use `//` for this?

roberth (commented):
It does look equivalent. Will approve when this is tested.

Ericson2314:
OK added some tests!

src/libstore/serve-protocol-impl.hh
```diff
@@ -56,4 +57,101 @@ struct ServeProto::Serialise
...
+    ServeProto::Version remoteVersion;
+
+    /**
+     * Establishes connection, negotiating version.
+     *
+     * @return the version provided by the other side of the
+     * connection.
+     *
+     * @param to Taken by reference to allow for various error handling
+     * mechanismsm.
```
roberth, line 73:
```suggestion
     * mechanisms.
```

src/libstore/serve-protocol-impl.hh
```diff
@@ -56,4 +57,101 @@ struct ServeProto::Serialise
...
+     */
+    static ServeProto::Version handshake(
+        BufferedSink & to,
+        Source & from,
+        ServeProto::Version localVersion,
+        std::string_view host);
+
+    /**
+     * Coercion to `ServeProto::ReadConn`. This makes it easy to use the
+     * factored out serve protocol searlizers with a
```
roberth, line 90:
```suggestion
     * factored out serve protocol serializers with a
```

src/libstore/serve-protocol-impl.hh
```diff
@@ -56,4 +57,101 @@ struct ServeProto::Serialise
...
+    {
+        return ServeProto::ReadConn {
+            .from = from,
+            .version = remoteVersion,
+        };
+    }
+
+    /**
+     * Coercion to `ServeProto::WriteConn`. This makes it easy to use the
+     * factored out serve protocol searlizers with a
```
roberth, line 106:
```suggestion
     * factored out serve protocol serializers with a
```

src/libstore/serve-protocol-impl.hh
```diff
@@ -56,4 +57,101 @@ struct ServeProto::Serialise
...
+    StorePathSet queryValidPaths(
+        const Store & remoteStore,
+        bool lock, const StorePathSet & paths,
+        SubstituteFlag maybeSubstitute);
+
+    /**
+     * Just the request half, because Hydra may do other things between
+     * issuing the request and reading the `BuildResult` response.
+     */
+    void buildDerivationRequest(
```
roberth, line 129: If it doesn't go against a convention.

```suggestion
    void putBuildDerivationRequest(
```

src/libstore/serve-protocol-impl.hh
```diff
@@ -56,4 +57,101 @@ struct ServeProto::Serialise
...
+struct ServeProto::BasicServerConnection
+{
+    /**
+     * Establishes connection, negotiating version.
+     *
+     * @return the version provided by the other side of the
+     * connection.
+     *
+     * @param to Taken by reference to allow for various error handling
+     * mechanismsm.
```
roberth, line 144:
```suggestion
     * mechanisms.
```

tests/unit/libstore/serve-protocol.cc
```diff
@@ -401,4 +403,112 @@ VERSIONED_CHARACTERIZATION_TEST(
...
+
+        StringSource in { toClientLog };
+        auto clientResult = ServeProto::BasicClientConnection::handshake(
+            nullSink, in, defaultVersion, "blah");
+
+        EXPECT_EQ(clientResult, defaultVersion);
+    });
+}
+
+TEST_F(ServeProtoTest, handshake_client_trunated_replay_throws)
```
roberth, line 456:
```suggestion
TEST_F(ServeProtoTest, handshake_client_truncated_replay_throws)
```

roberth (approved):

PR #4289, opened 2020-11-28
darwin: encrypt nix volume if filevault is enabled

abathur:
This refactors WIP from #4181 to address things that became obvious:
1. The macOS install is getting too complex for single-user mode.
1. There's already frustration around difficulty of uninstalling multi-user on macOS.
1. I want to support upgrading existing macOS volumes to work the same as those made by this WIP, but simply tacking it on as an extra option at all relevant points was looking like a lot of extra conditional logic and calls to some fairly slow diskutils commands.
1. The greater number of Nix-related artifacts shook loose additional concern around avoiding state/artifact inconsistencies.
1. As I developed a curing process to address these inconsistencies, I realized it won't take too much more work to extend these concepts to uninstalling the other Darwin and general Nix multi-user artifacts. I'm pleasantly surprised how practical an uninstall mode feels from here.

The history had grown unruly even before this refactor, so for now I'm flattening this into an omnibus commit to start fresh. Since this touches on a lot, I've tried to better-document the high-altitude view by making three outlines:
- ~the [general process/logic in pseudocode](https://github.com/abathur/nix/blob/encrypt_darwin_volume/scripts/darwin_volume_reference.py) (as-is, and for the uninstaller I'm imagining)~ removed this for merge
- how high-level user-facing functionality changes here
- what lower-level implementation changes I'm making

## High-level functionality changes
- disable single-user installs on macOS
- if the user's root is read-only, create a volume (overridable default)
	- `--darwin-use-unencrypted-nix-store-volume` flag is now a no-op warning
	- if the boot volume has FileVault enabled, encrypt this volume
		- generate a random credential and place it in the system keychain
	- create a LaunchDaemon to mount it
		- still use fstab to specify mount opts (rw,noauto,nobrowse,suid,owners)
			- noauto so that the system *won't* auto-mount if our mounter breaks (goal: the user finds/presents for help)
		- the daemon isn't strictly necessary for unencrypted volumes, but I'm using it in both cases for consistency
- update volume naming/customization
	- default: use or create volume named "Nix Store" on same disk as /
		- I had changed this to "Nix", but I want to support up-converting current "Nix Store" volumes and can't justify the complexity of supporting both names.
		- no longer tries to match and use volumes labeled `[nN]ix.*` (see #4150)
	- overridable variables for the filesystem, label, and target disk
	- overridable variables to BYOV: volume special device, and UUID
- support curing any volume-related state (if present) on install
	- if Nix Store is encrypted
		- verify we can unlock it from keychain
[...]

PR #6144, opened 2022-02-22
doc: Add detailed uninstall section for macOS

toonn:
The multi-user installation on macOS, which is now the only option, has
gotten complicated enough that it discourages some users from checking
Nix out for fear of being left with a "dirty" system. Detailed
uninstallation instructions should make this less of an issue.

Note:
I left the Linux uninstall instructions unchanged but I suspect they could use a similar revamp wrt removing build users at least.

### DIFF
--- a/doc/manual/src/installation/installing-binary.md
+++ b/doc/manual/src/installation/installing-binary.md
@@ -84,7 +84,9 @@ The installer will modify `/etc/bashrc`, and `/etc/zshrc` if they exist.
 The installer will first back up these files with a `.backup-before-nix`
 extension. The installer will also create `/etc/profile.d/nix.sh`.
 
-You can uninstall Nix with the following commands:
+## Uninstalling
+
+### Linux
 
 ```console
 sudo rm -rf /etc/profile/nix.sh /etc/nix /nix ~root/.nix-profile ~root/.nix-defexpr ~root/.nix-channels ~/.nix-profile ~/.nix-defexpr ~/.nix-channels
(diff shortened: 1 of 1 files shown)


abathur:
Two thoughts (beyond how much this will help!):

1. This won't be universally true. I'm not entirely sure how to communicate the asterisks to the reader...
    1. I'm not 100% certain, but I don't think the installer will create the volume/synthetic.conf/fstab/mounter if the root is writable (so, pre-Catalina). 
    
        (I have thought it might be worth seeing if the older systems would support the components of this to wring more variety out of the install base, but unless it Just Works, it could mean a lot of effort for an unknown-but-probably-small number of installs--and we'd still have legacy installs.)
    2. Even on Catalina+, people with older installs (up through mid fall?) may not have the the mounter. Some inevitably still have single-user installs.
2. Almost noted this inline, but I'm dithering a little about placement (at the end of 5? at the end of 7? after the list as now?). It might help for the note on the empty mount dir to be more explicit? 

    Not sure this is right; just sketching...

    > Note: After you complete the steps here, you will still have an _empty_ `/nix` directory. This is an expected sign of a successful uninstall. The empty /nix directory will disappear the next time you reboot.
    > 
    > You do not _have_ to reboot to finish uninstalling Nix. The uninstall is complete, but macOS (Catalina+) directly controls root directories--its read-only root will prevent you from manually deleting the empty /nix mountpoint.

toonn:
Definitely valid points.

Do we want to explicitly address uninstall on older versions or do we leave that up to readers to figure out from the newer instructions? The main difference is the store so it wouldn't be too hard to figure out that can simply be removed.

That's a good clarification in the note. Should this be a footnote or inline?

abathur:
> Do we want to explicitly address uninstall on older versions or do we leave that up to readers to figure out from the newer instructions? The main difference is the store so it wouldn't be too hard to figure out that can simply be removed.

I'll chew on this a little. I just read the diff before, but I want to go back to read the full page to make sure I'm fresh on how it addresses single/multi linux/mac.

> That's a good clarification in the note. Should this be a footnote or inline?

Happy to go with your gut if you like the text. I started writing it for the end of step 5, but I started to worry that just mentioning reboot at the end of 5 would make some number of people misread and try rebooting, which would then break step 7. So, by the end, I was thinking it might replace the paragraph after 7?

doc/manual/src/installation/installing-binary.md
```diff
@@ -95,15 +97,87 @@ sudo systemctl stop nix-daemon.service
...
+   This will stop shells from sourcing the file and bringing everything you
+   installed using Nix in scope.
+
+2. Stop and remove the Nix daemon services:
+
+   ```console
+   sudo launchctl unload /Library/LaunchDaemons/org.nixos.nix-daemon.plist
+   sudo rm /Library/LaunchDaemons/org.nixos.nix-daemon.plist
+   sudo launchctl unload /Library/LaunchDaemons/org.nixos.activate-system.plist
+   sudo rm /Library/LaunchDaemons/org.nixos.activate-system.plist
```
abathur, line 135: This is the nix-darwin service, yeah? did you mean to use that one instead of the mount daemon? 

(I was reading over all of this and thinking maybe we could get away with adding a few qualifiers to a few of these steps to say something like "You may not have X if you if you are running macOS Mojave (or earlier), or if you installed Nix before Y.")

abathur:
> > Do we want to explicitly address uninstall on older versions or do we leave that up to readers to figure out from the newer instructions? The main difference is the store so it wouldn't be too hard to figure out that can simply be removed.
> 
> I'll chew on this a little. I just read the diff before, but I want to go back to read the full page to make sure I'm fresh on how it addresses single/multi linux/mac.

I am not sure we actually need to address it. I don't really know how many of these systems would plausibly be left? I think it could be as simple as a note at the beginning. Something like:

> *Note:*
> 1. If you are running macOS Mojave (10.14) or earlier, you will have a simple /nix root directory instead of a Nix Store volume. Specifically, this means:
>    - you will not have the nix-store LaunchDaemon removed in step 2
>    - you can skip steps 4, 5, and 7
>    - after step 6, you should also run `sudo rm -rf /nix`
> 2. If you first installed Nix before November 2021 (Nix versions earlier than 2.4) you will not have the nix-store LaunchDaemon removed in step 2.

That said, 0 complaints from me if you think these are too rare in the real-world to allocate this many vertical lines to... 🤿

edolstra:
Ideally, we would have an uninstall script that does these things automatically. Is there a reason why we can't put these commands in a script (possibly the installer, under a `--uninstall` flag)?

abathur:
> Ideally, we would have an uninstall script that does these things automatically. Maybe we can put these commands in a script (possibly the installer, under a `--uninstall` flag)?

@edolstra Strong agree on the "ideally" part, though I think ideal shouldn't be the enemy of better, here. (I already let that happen once myself.) :)

General thoughts:
- `--uninstall` seems *very* achievable to me. 
- In the process of #4289 I worked out a few "curing" patterns for macOS-specific artifacts that can readily be re-used as building blocks for **both** `--uninstall` and reinstall support. 

    (Caveat: these are just in the multi-user script; single user would need some separate effort.)
- These worked well enough that I was tempted to reach for `--uninstall` during #4289. I did an early prototype just to convince myself it was feasible. But I was got burnt out on how much effort that process was taking. I needed to cut scope to move on to other work. I'm also trying to avoid ending up ~owning the installer. 
- I nonetheless stubbed out some furniture for uninstall, mostly around https://github.com/NixOS/nix/commit/eab14a642cbcbc35f4473888d906f9de7deda07b#diff-e2c41a4fae89675f4eccc257282164c935ab616feb57e0dbd3ce45405b1bd016R51-R128
- I (naively) thought `--uninstall` would be exciting enough to almost sell itself, so I let ideal be the enemy of better and turned my own focus to trying to drum up interest in investments to improve our velocity on the installer in https://discourse.nixos.org/t/anyone-up-for-picking-at-some-nix-onboarding-improvements/13152 and later https://discourse.nixos.org/t/installer-test-suite-small-project-s-high-leverage-help-wanted/13662

I have some more thoughts but need to run for now...

doc/manual/src/installation/installing-binary.md
toonn: Ah, you're right of course. I was originally writing the instructions from a "I want to get rid of all things Nix," point of view but this would be more suited to the nix-darwin manual.

Maybe we can use the fstab step as the first differentiating step and add a note along the lines of "If you don't have this entry you probably have an older install on a pre-Catalina macOS, you won't have to do steps X, Y and Z and you should `rm /nix` instead?"

toonn: Ah, yes, similar to what you say in the very next comment but just localized to the first step where someone might notice a difference.

Alternatively we could make all the post-Catalina steps a sublist of a single step and then we can say to just skip the one step. Keeping them at the top-level list is slightly more flexible, the current order is based on not ending up with a broken setup at any point and going from easy to un-undo to harder to un-undo. However, I do think we could do all the post-Catalina specific stuff grouped.

doc/manual/src/installation/installing-binary.md
```diff
@@ -131,8 +131,6 @@ and `/etc/zshrc` which you may remove.
    ```console
    sudo launchctl unload /Library/LaunchDaemons/org.nixos.nix-daemon.plist
    sudo rm /Library/LaunchDaemons/org.nixos.nix-daemon.plist
-   sudo launchctl unload /Library/LaunchDaemons/org.nixos.activate-system.plist
-   sudo rm /Library/LaunchDaemons/org.nixos.activate-system.plist
```
abathur: Erg. Sorry. I think my previous comment didn't connect the dots.  I assume it was supposed to be the nix-store daemon instead of the nix-Darwin one?

toonn: Oh, is that the one that mounts the volume? What's that called again, don't have access to a machine with the newer setup currently?

toonn:
@edolstra, totally agree that an uninstall command would be the way to go. Not the easiest to version since ideally a newer installer could still uninstall an older Nix but a great goal none the less. I'm not so sure a simple uninstall script containing these commands would we be good enough, however, making it available to copy all at once might be useful, should I add this to the docs? I could make the entire section a script with comments if that's considered useful.

doc/manual/src/installation/installing-binary.md
abathur: Yep! I don't actually either, but it is: `/Library/LaunchDaemons/org.nixos.darwin-store.plist`

abathur (approved):

PR #6155, opened 2022-02-24
Fix Symbol::'<' to sort alphabetically rather than by symbol pointer

yorickvP:
The symbol table is currently sorted by pointer, which has some weird effects (available upon request from @puckipedia). This fixes it to sort by the string contents.

### DIFF
--- a/src/libexpr/symbol-table.hh
+++ b/src/libexpr/symbol-table.hh
@@ -42,7 +42,7 @@ public:
 
     bool operator < (const Symbol & s2) const
     {
-        return s < s2.s;
+        return *s < *s2.s;
     }
 
     operator const std::string & () const


puckipedia:
Simplest example caused by pointer ordering:

```
yatadera ~> nix-instantiate --parse -E '[ { a = 5; b = 6; } { b = 5; a = 6; } ]'
[ ({ a = 5; b = 6; }) ({ a = 6; b = 5; }) ]
yatadera ~> nix-instantiate --parse -E '[ { b = 5; a = 6; } { b = 5; a = 6; } ]'
[ ({ b = 5; a = 6; }) ({ b = 5; a = 6; }) ]
```

Note how the order of the first attrset changes the printing behavior for both attrsets.

edolstra:
Sorting is intentionally done by pointer because it's faster. If there are any issues, then please create an issue for those.

edolstra:
Generally, for anything that is externally observable, you need to use `lexicographicOrder()` to iterate through an attrset.

yorickvP:
Alright, we should come up with some cases where this is externally observable:
- nix-instantiate --parse
- attrset equality checking
- toJSON
- ...

PR #6158, opened 2022-02-24
Sort the tests by wall time

thufschmitt:
The tests are scheduled in the order they appear, so running the long
ones first slightly improves the scheduling.

On my machine, this decreases the time of `make install` from 40s to 36s


cole-h:
Should there maybe be a comment above this list of tests to that effect? I can see somebody down the line wanting to alphabetize the list or something.

PR #6164, opened 2022-02-25
refactor: remove verbose-build from docs

kamadorueda:
- From what I see it is an implementation detail
  but is no longer configurable from the settings


tomberek (approved):
If there is no user control over this, then it is an implementation detail.

thufschmitt (approved):
(I’m actually not even sure that `verboseBuild` is really used internally)

PR #6168, opened 2022-02-25
lockfile: catch and describe JSON parse errors

### LINKED ISSUE #6151: Error message on corrupted flake.lock is non-informative
**Describe the bug**

When flake.lock contains invalid JSON (for instance because of a currently ongoing Git merge), any Nix commands dealing with the flake will fail with an error that doesn't mention flake.lock in any way:
```
error: [json.exception.parse_error.101] parse error at line 116, column 1: syntax error while parsing object - unexpected end of input; expected '}'
```
In my case, I had to run `strace` on Nix to find out that my flake.lock was at the source of the error.

**Steps To Reproduce**

1. Create a simple flake.nix
2. Create an empty flake.lock
3. Run `nix flake metadata`

**Expected behavior**

The error message about the JSON parsing error should mention `flake.lock`

`nix-env (Nix) 2.7.0pre20220127_558c4ee`

lincolnauster:
Fixes #6151.

@alyaeanyx and/or maintainers, please let me know if this new error message is concise and detailed enough!

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -325,8 +325,7 @@ static void main_nix_build(int argc, char * * argv)
 
     state->printStats();
 
-    auto buildPaths = [&](const std::vector<StorePathWithOutputs> & paths0) {
-        auto paths = toDerivedPaths(paths0);
+    auto buildPaths = [&](const std::vector<DerivedPath> & paths) {
         /* Note: we do this even when !printMissing to efficiently
            fetch binary cache data. */
         uint64_t downloadSize, narSize;
@@ -348,7 +347,7 @@ static void main_nix_build(int argc, char * * argv)
         auto & drvInfo = drvs.front();
         auto drv = evalStore->derivationFromPath(drvInfo.requireDrvPath());
 
-        std::vector<StorePathWithOutputs> pathsToBuild;
+        std::vector<DerivedPath> pathsToBuild;
         RealisedPath::Set pathsToCopy;
 
         /* Figure out what bash shell to use. If $NIX_BUILD_SHELL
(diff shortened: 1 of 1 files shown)


lincolnauster:
`nix flake check`'ing is fine on my machine. Would this be another spurious failure?

src/libexpr/flake/lockfile.cc
```diff
@@ -185,8 +185,15 @@ std::string LockFile::to_string() const
...
 LockFile LockFile::read(const Path & path)
 {
-    if (!pathExists(path)) return LockFile();
-    return LockFile(nlohmann::json::parse(readFile(path)), path);
+    LockFile l;
+    if (!pathExists(path)) l = LockFile();
+    else try {
+        l = LockFile(nlohmann::json::parse(readFile(path)), path);
+    } catch (const nlohmann::json::parse_error & e) {
+        throw Error("the lockfile did not contain valid JSON.");
```
thufschmitt, line 193: The initial error message isn’t very helpful by itself, but it can still give some information as to the origin of the error, so I’d rather keep it.

```suggestion
        e.addTrace(std::nullopt, "while trying to read the flake lockfile");
```

thufschmitt (changes requested):
Suggested a small change to keep the original error, but looks good otherwise.

The CI issue is due to #6169 , I think it’s safe to ignore it until master gets fixed

lincolnauster:
> The initial error message isn’t very helpful by itself, but it can
> still give some information as to the origin of the error, so I’d
> rather keep it.
>
> ```suggestion
>         e.addTrace(std::nullopt, "while trying to read the flake lockfile");
> ```

The trace already says "...while updating the lock file of flake
'<path'>". This is helpful, but non-obvious (enough, apparently, to
cause an issue on the repo). Rather than making the trace more granular,
I think that, from a UX perspective, it may be more helpful to have the
first error be the most useful.

I do agree, though, that the exact JSON error may also be helpful. What
about placing it in parentheticals in the first error, as it doesn't fit
well in a trace? b9b42d9

Thanks!

PR #6036, opened 2022-02-02
Flake follows: resolve all follows to absolute

balsoft:
It's not possible in general to know in computeLocks, relative to
which path the follows was intended to be. So, we always resolve
follows to their absolute states when we encounter them (which can
either be in parseFlakeInput or computeLocks' fake input population).

Fixes https://github.com/NixOS/nix/issues/6013
Fixes https://github.com/NixOS/nix/issues/5609
Fixes https://github.com/NixOS/nix/issues/5697 (again)

Issue #6013, opened 2022-01-29
nix 2.6.0 transitive follows results in an error


dvzubarev:
**Describe the bug**
When using a chain of transitive `follows` an error is occurred when creating lock file.

**Steps To Reproduce**
Consider this chain of flakes z -> a -> b -> c. 
The goal is to use the same nixpkgs in all packages.
1. `git clone https://github.com/dvzubarev/flake-follows-bug-poc2.git`
2. cd flake-follows-bug-poc2/z
3. nix flake update .

```
error: input 'a/b/c/nixpkgs' follows a non-existent input 'b/nixpkgs'
(use '--show-trace' to show detailed location information)
```

**Expected behavior**
I expect that nixpkgs that was set in `z` package will be propagated to all its inputs.
It works for `a` package.
E.g.
```
cd flake-follows-bug-poc2/a
nix flake update .
nix flake metadata .

warning: Git tree '/tmp/temp/flake-follows-bug-poc2' is dirty
Resolved URL:  git+file:///tmp/temp/flake-follows-bug-poc2?dir=a
Locked URL:    git+file:///tmp/temp/flake-follows-bug-poc2?dir=a
Path:          /nix/store/768kr4kadd8ddkda0wva35cz9h0j9hrl-source
Last modified: 2022-01-29 20:37:22
Inputs:
├───b: github:dvzubarev/flake-follows-bug-poc2/91cf8b9a7fcbd349ac54961b0d580d0c98c2d96d?dir=b
│   ├───c: github:dvzubarev/flake-follows-bug-poc2/46c67f57ff38672f0fe30e7af33fc5395c5e0730?dir=c
│   │   └───nixpkgs follows input 'b/nixpkgs'
│   └───nixpkgs follows input 'nixpkgs'
└───nixpkgs: github:nixos/nixpkgs/5bb20f9dc70e9ee16e21cc404b6508654931ce41
```

**`nix-env --version` output**
nix-env (Nix) 2.6.0

amesgen:
Many flake-based haskell.nix projects are affected by this (see https://input-output-hk.github.io/haskell.nix/tutorials/getting-started-flakes.html#scaffolding).

Bisecting points to 2664a216e57169ec57d7f51be1b8383c1be83fd5 / #5839 introducing this bug.

lovesegfault:
cc. @balsoft

lilyball:
[...]

Issue #5790, opened 2021-12-17
Transitive inputs "not in registry"


NobbZ:
**Describe the bug**

I want to use `follows` on the input of an input.

Though I get an error, that this is not in the registry.

**Steps To Reproduce**

1. Add an arbitrary `nixpkgs` input to a `flake.nix`.
2. Add the following to your `flake.nix` at appropriate location:
   ```nix
   inputs.rnix-lsp.url = "github:nix-community/rnix-lsp/master";
   inputs.rnix-lsp.inputs.nixpkgs.follows = "nixpkgs";
   inputs.rnix-lsp.inputs.naersk.inputs.nixpkgs.follows = "nixpkgs";
   ```
3. Update the `flake.lock` and observe error:
   ```plain
   error: cannot find flake 'flake:naersk' in the flake registries
   (use '--show-trace' to show detailed location information)
   ```

**Expected behavior**

The `nixpkgs` input of the `naersk` input of `rnix-lsp` is properly overwritten and follows "my" `nixpkgs`.

**`nix-env --version` output**

```
$ nix-env --version
nix-env (Nix) 2.6.0pre20211214_18e4851
```

**Additional context**

I tried this as a way to workaround #5609.

The same problem happens on 2.4 and 2.5:

```plain
$ nix shell github:nixos/nix/2.4 -c nix flake metadata
warning: Git tree '/home/…/Projects/system-config' is dirty
error: cannot find flake 'flake:naersk' in the flake registries
(use '--show-trace' to show detailed location information)
$ nix shell github:nixos/nix/2.5.0 -c nix flake metadata
warning: Git tree '/home/…/Projects/system-config' is dirty
error: cannot find flake 'flake:naersk' in the flake registries
(use '--show-trace' to show detailed location information)
```

regnat:
I can indeed reproduce this. More self-contained repro: 

```bash
#!/usr/bin/env bash

set -euo pipefail
set -x

rm -rf A B C D
mkdir -p A B C D

cat <<EOF > A/flake.nix
{ outputs = _: {}; }
EOF
cat <<EOF > B/flake.nix
{
  inputs.A.url = "path:$PWD/A";
  outputs = _: {};
}
EOF
cat <<EOF > C/flake.nix
{
  inputs.B.url = "path:$PWD/B";
  outputs = _: {};
}
EOF
cat <<EOF > D/flake.nix
{
  inputs.C.url = "path:$PWD/C";
[...]

PR #6171, opened 2022-02-26
flakes: failing follows test case

edrex:
this is a bug report in the form of a failing test case

A has input B, B has input foobar, and 

- B: `inputs.D.follows = "foobar/D"`
- A: `inputs.B.inputs.foobar.follows` = "foobar"

Result:

`error: input 'B/D' follows a non-existent input 'B/foobar/D'`

Expected: `['foobar', 'D']`

Tested with v2.6.1 with #6036 cherry-picked, since master is giving me a `SIGBUS` right now.

### DIFF
--- a/tests/flakes.sh
+++ b/tests/flakes.sh
@@ -737,6 +737,7 @@ cat > $flakeFollowsB/flake.nix <<EOF
             url = "path:./flakeC";
             inputs.foobar.follows = "foobar";
         };
+	D.follows = "foobar/D";
     };
     outputs = { ... }: {};
 }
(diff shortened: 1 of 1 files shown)


edrex:
cc @balsoft

edrex:
Context: found when attempting to use https://github.com/srid/emanote/blob/master/flake.nix as an input. @srid

edrex:
I made a small change to the existing test case to trigger the failure, but it seems like factoring out the flake follows tests to be more explicit should make it easier to reason about expected behavior. 

Very similar to https://github.com/NixOS/nix/issues/6013 but not quite the same.

Maybe a common cause with https://github.com/NixOS/nix/issues/5790

tomberek:
fixed

PR #6173, opened 2022-02-27
Fix `mkString` for empty `string_view`

roberth:
The crux of the problem is that the empty `string_view` is {`null`, `0`} and `dupStringWithLen` crashes on the `null`.

Discovered in the hercules-ci-cnix-store Haskell binding test suite, so _possibly_ not an issue in Nix itself _yet_.

This might also improve performance a bit, by not allocating empty strings.

I've had to change the name, because unlike a `dup`-function, it does not always allocate a new string anymore.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -805,7 +805,13 @@ LocalNoInline(void addErrorTrace(Error & e, const Pos & pos, const char * s, con
 
 void Value::mkString(std::string_view s)
 {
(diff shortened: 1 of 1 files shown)


thufschmitt (approved):
Thx

thufschmitt:
Oh, didn’t check the CI. Do you mind rebasing on top of master to make Darwin happy?

roberth:
Done

Issue #4313, opened 2020-12-04
Substitute builtin url fetchers


roberth:
**Is your feature request related to a problem? Please describe.**

Unlike when using a derivation-based fetcher, the evaluation-time fetchers `builtins.fetchurl` and `builtins.fetchTarball` do not try to substitute. I need to use these fetchers because these are the ones that respect the user's `netrc`, letting them download authenticated resources.
We'd like to use the binary cache to give developers and deployments access to the data without duplicating credentials to a system these principals shouldn't have full access to.

**Describe the solution you'd like**

When fetching data in `builtins.fetchurl` and `builtins.fetchTarball`, query the substituters too.

**Describe alternatives you've considered**

Duplicate credentials everywhere and make the setup less secure.

**Additional context**

Here's an example of trying to use the data on a machine without credentials. The data is available on the binary cache, but it wasn't queried.

```
nix-instantiate -vvv --eval --expr 'builtins.fetchTarball { url = "https://private-data-repo.example.com/machine-learning-model.dat"; sha256 = "de2d3f3892ecf7000fe8cddb2ce0c801217ec0f31079ba56beff62f20a0b982f"; }'
did not find cache entry for '{"name":"machine-learning-model.dat","type":"file","url":"https://private-data-repo.example.com/machine-learning-model.dat"}'
downloading 'https://private-data-repo.example.com/machine-learning-model.dat'...
starting download of https://private-data-repo.example.com/machine-learning-model.dat
verify TLS: Nix CA file = '/etc/ssl/certs/ca-certificates.crt'
finished download of 'https://private-data-repo.example.com/machine-learning-model.dat'; curl status = 0, HTTP status = 401, body = 12 bytes
error: --- FileTransferError ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix-instantiate
unable to download 'https://private-data-repo.example.com/machine-learning-model.dat': HTTP error 401 ('')

response body:

Unauthorized
download thread shutting down
```

Tested with approximately Nix revision 05d9442f6 ie `nixUnstable` around Nov 26 2020.
This was also brought up in #2114, again in #3543, assumed resolved before #4149 but currently not working.

bjornfor:
Also needed for `builtins.fetchGit`.

AleXoundOS:
_`builtins.fetchGit` seems to be affected as well, but not sure_.

roberth:
From #6174

[...]

PR #6174, opened 2022-02-27
fetchTree: Do not re-fetch paths already present + refactor

roberth:
When the expected path is already in the store, we don't need to re-fetch it.

This reduces the amount of fetching required when using flakes repeatedly and beyond the `tarball-ttl`. I believe this can be quite significant, especially as builtin fetchers run sequentially, blocking evaluation.

It is also a first step towards #4313 and a partial solution.

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -185,6 +185,13 @@ static void prim_fetchTree(EvalState & state, const Pos & pos, Value * * args, V
 // FIXME: document
 static RegisterPrimOp primop_fetchTree("fetchTree", 1, prim_fetchTree);
 
+static void setStorePath(EvalState &state, const StorePath &storePath, Value &v) {
+    state.allowPath(storePath);
+
+    auto path = state.store->printStorePath(storePath);
+    v.mkString(path, PathSet({path}));
+}
+
 static void fetch(EvalState & state, const Pos & pos, Value * * args, Value & v,
     const std::string & who, bool unpack, std::string name)
 {
@@ -202,6 +209,8 @@ static void fetch(EvalState & state, const Pos & pos, Value * * args, Value & v,
             if (n == "url")
                 url = state.forceStringNoCtx(*attr.value, *attr.pos);
             else if (n == "sha256")
+                expectedHash = newHashAllowEmpty(state.forceStringNoCtx(*attr.value, *attr.pos), htSHA256);
+            else if (n == "narHash")
                 expectedHash = newHashAllowEmpty(state.forceStringNoCtx(*attr.value, *attr.pos), htSHA256);
             else if (n == "name")
                 name = state.forceStringNoCtx(*attr.value, *attr.pos);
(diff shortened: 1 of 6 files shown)


src/libexpr/primops/fetchTree.cc
```diff
@@ -248,7 +241,7 @@ static void fetch(EvalState & state, const Pos & pos, Value * * args, Value & v,
 
         auto validPaths = state.store->queryValidPaths({expectedPath}, NoSubstitute);
```
thufschmitt, line 242: Stupid question, but would passing `Substitute` here fix #4313 ?

roberth: Great question.
It would, but there would also be a performance cost.

Remote path querying has a significant latency, that I don't think we want to incur on every fetch, many of which will not be substitutable anyway. Ideally we could perform the remote path info query while the fetcher runs, and make some clever decisions regarding cancellation. That's not something I want to dive into, especially considering the non-zero risk of corrupting local files when cancelling the fetcher.

Instead I figured the next step could be to try substitution when the fetching fails, which also solves #4313, without performance impact. So now the question is; do you like big or small PRs?

thufschmitt:
Thanks for the explanation, that makes sense (I also wish we could do the same thing for the derivations. But that’s another story)

> So now the question is; do you like big or small PRs?

Just in case that’s not a rhetorical question, please please please keep them small :)

Do you mind rebasing on top of master to fix the OSX build? Apart from that that’s good for me

roberth:
> rhetorical question

It was an honest question. Better to ask than assume, but yeah, I guess I could have guessed.

> rebasing

Done!

src/libexpr/primops/fetchTree.cc
Ericson2314: Can we add a FIXME here linking #4313? Fine if just in the issue we add the details about latency.

thufschmitt (approved):

src/libexpr/primops/fetchTree.cc
edolstra: I think substituting is fine since the binary cache substituter caches negative lookup results. I.e. we can call `ensurePath` here instead of `queryValidPaths`.

BTW, why `queryValidPaths` instead of `store->isValidPath(expectedPath)`?

roberth: I don't see how negative caching helps here. When a path is not in the cache, it will be fetched, after which it won't have to be looked up again. I suppose it might be useful to start an asynchronous query and then rely on the cache, but then the real benefit is in positive caching anyway.

> why `queryValidPaths`

I glossed over it. Maybe it's because `valid` makes me uneasy. Validity just doesn't seem as strong a property as existence. Yet, I've found `queryValidPaths` so maybe I should stop analyzing myself :sweat_smile:

edolstra: You're right, negative caching doesn't matter. But it also means that the overhead of substitution is a non-issue since it will only happen once.

PR #6175, opened 2022-02-27
Fix temp dir variable in develop shell

ysndr:
Inside a `nix develop` shell nix commands that refer to a remote flake (that is not cached) construct a faulty tmp path (missing a `/`)

```
getting status of '/private/tmpnix-shell.cTvSGw': No such file or directory
```

should be
```
 '/private/tmp/nix-shell.cTvSGw'
              ^
```

This PR adds the trailing `/` back to `TMP*` variables inside the shell


ysndr:
This was an entirely wrong approach.
Converting this to an issue.

PR #6179, opened 2022-02-28
tests: Fix the start of the daemon

thufschmitt:
- Make sure that it starts even without the `nix-command` xp feature
- Fail if it doesn’t manage to start

This fixes a 30s wait for every test in `init.sh` as the daemon couldn’t start, but the code was just waiting 30s and continuing as if everything was all right.

### DIFF
--- a/tests/common.sh.in
+++ b/tests/common.sh.in
@@ -90,12 +90,18 @@ startDaemon() {
     # Start the daemon, wait for the socket to appear.  !!!
     # ‘nix-daemon’ should have an option to fork into the background.
     rm -f $NIX_DAEMON_SOCKET_PATH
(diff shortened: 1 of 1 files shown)


thufschmitt:
> @thufschmitt this overlaps a bit with my #5753, I guess I will rebase that but I hope we can do both.

Ah, I knew you had something touching the same bits of code :) (and looks like yours would fix the issue too).

If you’re OK rebasing, I’ll merge that one right now as the CI times are really a pain and that’s the quickest fix

Ericson2314:
@thufschmitt Yes I am totally fine rebasing. Sound good!

Ericson2314 (approved):
Ready to merge

PR #6181, opened 2022-02-28
Auto uid allocation -- update with latest master

Ericson2314:


edolstra:
Maybe better to update #3600 rather than have multiple open PRs for the same thing?

thufschmitt:
> Maybe better to update #3600 rather than have multiple open PRs for the same thing?

I think that’s intent of this PR − it’s opened against the `auto-uid-allocation` branch which @Ericson2314 doesn’t have write access to

edolstra:
Ah thanks, I didn't see that.

Ericson2314:
Thanks! Yeah I can't push there directly so I did this instead. The last few merges were also done this way.

Issue #5638, opened 2021-11-23
Triage and break up the global settings


Ericson2314:
Thinking through #5603, and the two `getDefaultSystemFeatures` reminded me of this:

1. We have many global settings that should be per-store`. `thisSytem` and `extraPlatforms` are good examples.
2. We have too much ad-hoc logic around build remotes that should be fixed up to use per-store settings instead, as we did before with system features in  4720853129b6866775edd9f90ad6f10701f98a3c
    3. We should probably make stores that can build a subclass of stores that just are read/written. Then we don't have to do silly things like say what "extraPlatforms" `s3://...` has. 
4. https://github.com/NixOS/nix/issues/5025#issuecomment-887444929 remains a very good idea. Then many of the remaining settings which aren't per-store would make sense as an additional config parameter to this now-freestanding (not `Store` method) function.

rickynils:
I'm dumping some references here, where discussions about options propagation to remote builders can be found:

 https://github.com/NixOS/nix/issues/5600
 https://github.com/NixOS/nix/pull/2994
 https://github.com/NixOS/nix/pull/4180
 https://github.com/NixOS/nix/issues/5638

The issue is that when using remote builders or remote stores, it is not at all clear what options will be in effect. For `ssh://` there is a small set of settings propagated, for `ssh-ng://` almost no options (I think `builders-use-substitutes` is the only one) are propagated. It doesn't make sense to propagate all Nix options indiscriminately, because many settings are related to aspects of the local machine.  If I understand this issue correctly, we could identify options that are related to specific types of stores and then propagate these when such remote stores are used?

Ericson2314:
The author of #6980, @virusdave, talks it about it being confusing that an *arbitrary* Nix invocation looks up the users home directory. I agree this seems off!

The reason for this is ultimately that our settings are stored in these giant global variables. The default value settings must be created and start-up time, and since profiles optionally got moved to home directories, that means looking up the user's home directory.

The new location is good, and the warning on looking up the home directory is also good. The problem is, once again, the global. Global settings mean a violation of "pay for what you use, only". We have to do all this work for settings we may well not end up using, simply because global variables mean all configuration must be done eagerly.

This is not good!

virusdave:
>   * illustrating
>     ```json
>     {
>       # bad according to Eelco
>       "gc": {
>         "auto": {
>           "enable": true
>         }
>       }
>       # good acording to Eelco
>       "gc-auto-enable": true,
[...]

PR #6183, opened 2022-03-01
Move some stuff from `Settings` to a new `FetchSettings`.

Ericson2314:
Starting work on #5638

The exact boundary between `FetchSettings` and `EvalSettings` is not
clear to me, but that's fine. First lets clean out `libstore`, and then
worry about what, if anything, should be the separation between those
two.


Ericson2314:
Thanks!

Issue #6152, opened 2022-02-23
`outputs.__functor` breaks `nix` CLI's target resolver


blaggacao:
- latest master

**Wrong**
```
nix repl> nix-repl> outputs.
outputs.__functor
outputs.foo

$ nix eval --json .#foo
error: the expression selected by the selection path 'packages.x86_64-linux.foo' should be a set but is a function
```

**Expected**
```
nix repl> nix-repl> outputs.
outputs.foo
$ nix eval --json .#foo
{"x86_64-linux":{ ... }}
```

**Observation**

- the presence of `outputs.__functor` breaks the `nix` CLI target resolver.

/cc @tomberek @cleverca22

tomberek:
Duplicate? https://github.com/NixOS/nix/issues/4415

tomberek:
Caused by a call to autoCallFunction and the subsequent detection of a "__functor" attribute:

Trace:
https://github.com/NixOS/nix/blob/94992a919636fb9df4c7441dacbde08592e4494d/src/libcmd/installables.cc#L614
https://github.com/NixOS/nix/blob/bd383d1b6f91c4fe7ac21c52771e92027f649fa0/src/libexpr/attr-path.cc#L59
https://github.com/NixOS/nix/blob/bd383d1b6f91c4fe7ac21c52771e92027f649fa0/src/libexpr/eval.cc#L1513

disassembler:
@tomberek @edolstra I think this is a major blocker for nix 2.7. can we get this fixed before that release?

cleverca22:
reproduced without involving flakes
```
[clever@amd-nixos:~]$ cat autocall.nix ; nix-instantiate --eval  --argstr arg2 bar --argstr arg1 foo autocall.nix -A attr1.attr2 --argstr arg3 baz
{ arg1 }: {
  attr1 = {
    __functor = self: arg2: {
      attr2 = { arg3 }: "${arg1}${arg2}${arg3}";
    };
    attr2 = "functor not followed ${arg1}";
  };
}
error: the expression selected by the selection path 'attr1.attr2' should be a set but is a function
```

based on https://gist.github.com/cleverca22/240c21dab528666c85a95c4edfdeae58

edolstra:
@disassembler Hm, why is it a blocker? It's not a regression right?

blaggacao:
@edolstra fair question and the answer is dynamic & social, not technical.

[...]

Issue #4415, opened 2021-01-01
`nix-instantiate --eval --strict -A` broken with __functor


Profpatsch:
```
> nix-instantiate --eval --strict -E '{ __functor = _: v: v; foo = {}; }.foo'
{ }
> echo '{ __functor = _: v: v; foo = {}; }' > abc.nix
> nix-instantiate --eval --strict abc.nix -A foo
error: the expression selected by the selection path 'foo' should be a set but is a function
```

I expect the last one to understand `__functor` and return `{}`.

edolstra:
This is fundamentally ambiguous because `nix-instantiate` auto-calls functions (and thus also functors), so it's not clear whether the `foo` attribute should refer to the output or input of the functor. So the current behaviour is not really wrong.

PR #6184, opened 2022-03-01
optionally dont run functors when finding along an attrpath

cleverca22:


blaggacao:
@tomberek @edolstra 

This closes https://github.com/NixOS/nix/issues/6152 & https://github.com/NixOS/nix/issues/4415

I have tested this successfully.

Thank you @cleverca22

kamadorueda:
Having a regression test would be awesome

It's just creating a file like: https://github.com/NixOS/nix/blob/master/tests/flake-bundler.sh
and then adding the file name here: https://github.com/NixOS/nix/blob/master/tests/local.mk

edolstra:
It is not clear whether this is desirable behaviour. Due to the interaction between functors and auto-call, it's ambiguous whether you want the attributes from the original set or the output of the functor. E.g. in 
```
outputs = { ... }: {
  x = 123;
  __functor = self: { }: { x = 456; };
};
```
it's not obvious whether `.#x` should evaluate to 123 or 456.

Some possibilities:
* Disallow functors here. I think this is okay since you can always call the functor explicitly.
* Always apply the functor (i.e. `.#x` evaluates to 456).

cleverca22:
the original fault in https://github.com/NixOS/nix/issues/6152 is because a `__functor` is used similarly to `.extend`, and it should only be called with a set of changes

so when you have:
```
{
__functor = ...;
key = 123;
}
```
nix flakes should never run the functor, and just directly use the `.key`

and the exact error, is because `autoCallFunction` ran the functor with the unexpected args, got the wrong result, and everything failed

this PR should just disable `__functor` any time a flake based code is doing an attrpath lookup

edolstra:
If the functor should not be called, then it shouldn't be a functor...

Especially for the flakes case, it seems better to simply forbid the `outputs` function to be a functor.

roberth:
It'd be least surprising if command line attribute syntax behaves close to expression language attribute syntax.

I've also run into this problem trying to define passthru-style tests for fetchers.

 - https://github.com/NixOS/nixpkgs/pull/136022#issuecomment-923944541

roberth:
I find no elegance in functor. It tends to violate single responsibility, and it obfuscates code due to its lack of visibility at the call site / ambiguity with regular application. I try to restrict my use of functor to backward compatibility.

tomberek:
I'm confused at this point. I can come up with instances to use __functor (eg: i'm using it for backward-compat [here](https://github.com/NixOS/bundlers/blob/master/flake.nix#L32)), but my presumptions would be that one can argue for either behavior.

As I was thinking about this I became even more confused about why autoCallFunction is even used. This seems to be a hold-over from trying to have files of this form easily used from CLI.
```
{ pkgs ? import <nixpkgs>{}):
blah
```
To ask another way, why are we auto-calling functions/functors here? Why does `findAlongAttrPath` call functions? Should it?

[another random and possibly horrible thought: there is some open syntax that can capture this distinction `someflake#usingFunctor` and `someflake#.bypassFunctorAndUseAttr` with the extra leading "."]

edolstra:
@tomberek Yeah, I think we shouldn't be auto-calling functions (including functors) at all for flakes. Auto-calling only makes sense if there are arguments to pass, which we don't support for flakes.

tomberek:
Perhaps we have findAlongAttrPath use a "callFunctions" (not sure what the right name should be) and remove the possibility of auto-calling either a function or a functor during attrPath traversal from the flake CLI. Normal attrPath auto-calling should not be impacted as it is relied upon all over the place.

disassembler:
So what's the solution @edolstra? From what I gather how it currently works is incorrect but you're not happy with how this resolves it. Do you have a suggestion that we can try to resolve the issue?

blaggacao:
- @edolstra Would it be possible to reconsider this for 2.8?
- @cleverca22 Can you rebase to make it cheaper to merge?

blaggacao:
superseded by #6399 (rebase)

blaggacao:
This is fixed in `2.8.0`. This can be closed. :rocket:

Ericson2314:
https://github.com/NixOS/nix/pull/6399 links to https://github.com/NixOS/nix/pull/6400 which links to https://github.com/NixOS/nix/pull/6404 which was merged.

@cleverca22 As author, do you agree this can be closed?

PR #6187, opened 2022-03-01
Acknowledge that the macOS tests are flaky

thufschmitt:
Restart the tests (at most once) on `unexpected EOF` errors on OSX.

This is truly ugly, but might prevent half of the CI runs to fail because of <https://github.com/NixOS/nix/issues/3605>.

Since the tests aren’t restarted more than once, we’re still gonna encounter some failures due to that, which will serve as a reminder for the underlying issue (but withouth blocking everything)

/cc @edolstra @roberth

### DIFF
--- a/mk/run_test.sh
+++ b/mk/run_test.sh
@@ -14,9 +14,27 @@ if [ -t 1 ]; then
     yellow="[33;1m"
     normal="[m"
 fi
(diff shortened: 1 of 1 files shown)


Ericson2314:
I am OK doing this for now, but once we have non-GHA CI, I rather experiment with always running the tests in a separate derivation, or even running them each in their own derivation.

roberth (approved):
CI failure fatigue is worse than this bug.

Issue #5729, opened 2021-12-05
Split `Store` class


Ericson2314:
The `Store` class is too broad. It would be nice to see it split up to separate concerns, and avoid `unsupported(...)` default methods.

Intuitively, we have stores than just read and write files, and store that also can do builds. The latter would have primary and secondary `system`s, `supportedFeatures`, etc. etc. in the `Config` class.

But it's good to think about this distinction in the context of https://github.com/NixOS/nix/issues/5025#issuecomment-887444929, making `buildPaths` a top level method that works with a variety of store for different purposes.

Now `buildPaths` as a top-level method is not perfect because sometimes we *do* want to be able to let a remote store schedule too. E.g. when there is a bunch of remote builders, many people want to send jobs to a single central scheduler and let *it* dispatch jobs. So we don't want to just loose the ability of today's `buildPaths` implementation in `RemoteStore` and `LegacySSHStore`.

That means maybe it stays a method in some class (`StoreThatCanSchedule`?) after all? Or is `StoreThatCanSchedule` really the same as `StoreThatCanBuild` in practice? Would `LocalStore` need to implement this method, or should we just downcast and use the freestanding `buildPaths` function if it succeeds?

I don't know the answers to these; there is a lot to think through. Perhaps there is another easier split we can start with first, to start untangling things and improve our understanding?

Ericson2314:
Thinking about it slightly more, I think a good first step is to split out a subclass for the GC methods. It doesn't achieve to much, but also seems much more straightforward than the issues above. That would therefore serve as a good "test run" for this sort of refactor, and still do the good work of getting rid of some `unimplemented` calls.

Ericson2314:
@edolstra How do you feel on `ReadStore` vs `WriteStore: public virtual ReadStore` as a split to do have `GcStore`?

On one hand, I think only `HttpBinaryCacheStore` can't implement `WriteStore`, so it is not that useful for categorizing implementations. On the other hand, I think it could well be used to organize with types how store are *used* (e.g. substitutors can be mere `ReadStore`s), and that as Rust roughly demonstrates this is a general useful thing to do.

Also, in longer term `ReadStore` vs `WriteStore` could fit into various authentication infrastructure. (Thinking of the nexus of https://gist.github.com/edolstra/afa5a41d4acbc0d6c8cccfede7fd4792 and CA trust maps, for example. Maybe you can subscribe to certain things one cannot read from?)

PR #6188, opened 2022-03-01
Factor out a `GcStore` interface

Ericson2314:
Starts progress on #5729.

The idea is that we should not have these default methods throwing
"unimplemented". This is a small step in that direction.

I kept `addTempRoot` because it is a no-op, rather than failure. Also,
as a practical matter, it is called all over the place, while doing
other tasks, so the downcasting would be annoying.

Maybe in the future I could move the "real" `addTempRoot` to `GcStore`,
and the existing usecases use a `tryAddTempRoot` wrapper to downcast or
do nothing, but I wasn't sure whether that was a good idea so with a
bias to less churn I didn't do it yet.

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -1,6 +1,7 @@
 #include "globals.hh"
 #include "shared.hh"
 #include "store-api.hh"
+#include "gc-store.hh"
 #include "util.hh"
 #include "loggers.hh"
 
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1,4 +1,5 @@
 #include "local-derivation-goal.hh"
+#include "gc-store.hh"
 #include "hook-instance.hh"
 #include "worker.hh"
 #include "builtins.hh"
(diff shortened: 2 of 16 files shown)


src/libstore/gc-store-api.cc
```diff
@@ -0,0 +1,13 @@
+#include "gc-store-api.hh"
```
edolstra, line 1: I would just call this `gc-store.hh` and `gc-store.cc`. API is redundant since every header is an API...

(Yes, `store-api.hh` is also a bad name.)

src/libstore/gc-store-api.cc
```diff
@@ -0,0 +1,13 @@
...
+
+namespace nix {
+
+GcStore & requireGcStore(Store & store0)
+{
+    auto * store = dynamic_cast<GcStore *>(&store0);
+    if (!store)
+        throw UsageError("Garbage collection not supported by this store");
+    return *store;
```
edolstra, line 10:
```suggestion
GcStore & requireGcStore(Store & store)
{
    auto * gcStore = dynamic_cast<GcStore *>(&store);
    if (!gcStore)
        throw UsageError("Garbage collection not supported by this store");
    return *gcStore;
```

and in general use `store` and `gcStore` instead of `store0` and `store`.

edolstra (approved):

src/libstore/gc-store-api.cc
Ericson2314: :) yes that was my thought too but didn't want to buck the pattern unless you agreed!

Ericson2314:
Glad you like this!

Ericson2314:
OK, both done!

edolstra:
Thanks!

PR #6196, opened 2022-03-02
Nix's own flake: Dedup and memoize more

Ericson2314:
- `nixpkgsFor` does all of native, static, cross, and the different stdenvs.

- The main Nix derivation is no longer duplicated for static.


Ericson2314:
Closing in favor of #6538.

Issue #6182, opened 2022-03-01
Build daemon without rest of Nix


Ericson2314:
From https://github.com/NixOS/rfcs/pull/68#issuecomment-722413588

> @Ericson2314 Would you be opposed to closing this RFC and creating a new feature request issue in the Nix repo? Having a minimal daemon is not really a potentially controversial or hard change that requires a shepherd team and all that :-) What do you think?

OK opening this. For me the requirements are this:

- Don't build `libexpr` `libfetchers`
- Expose modern `nix-daemon` and legacy `nix-daemon`

It seems like then we would have to make `libcmd` and `libmain` not depend on `libexpr`. Interesting!

edolstra:
IMHO this is fairly low-priority and shouldn't be done if it causes a lot of code churn, since the value of a minimal daemon isn't entirely obvious. (All that libexpr code doesn't actually get used by the daemon, so it shouldn't cause any problems.) We're also moving in the direction of providing Nix as a single binary, and this goes against that.

Ericson2314:
Last thing before I let this lie dormant:

> We're also moving in the direction of providing Nix as a single binary, and this goes against that.

So really, I would love to have libfetchers-less/libexpr-less versions of `nix hash` `nix copy` `nix build` and other such "core" commands. Even more churn, but also preservation of the separate binaries idea.

We could have a `nix-core` binary (name TBD) with just those sorts of commands, without a libfetchers or libexpr, and also the regular `nix` binary with everything as today. We could just install/build `nix-core` for regular end users, but offer a separate `nix-core` package for those of us interested in these sort of applications (big industrial, alternative frontend, and standardization).

I think I could this without duplicated code, but yes with lot's of churn.

tomberek:
Closing as stale. Re-open if needed.

PR #6197, opened 2022-03-02
nix profile: Support CA derivations

edolstra:
Also moved some installable-related functions from `command.hh` into `Installable` in `installables.hh`.

### DIFF
--- a/src/nix/profile.cc
+++ b/src/nix/profile.cc
@@ -61,6 +61,27 @@ struct ProfileElement
     {
         return std::tuple(describe(), storePaths) < std::tuple(other.describe(), other.storePaths);
     }
+
+    void updateStorePaths(ref<Store> evalStore, ref<Store> store, Installable & installable)
+    {
+        // FIXME: respect meta.outputsToInstall
+        storePaths.clear();
+        for (auto & buildable : getBuiltPaths(evalStore, store, installable.toDerivedPaths())) {
+            std::visit(overloaded {
+                [&](const BuiltPath::Opaque & bo) {
+                    storePaths.insert(bo.path);
+                },
+                [&](const BuiltPath::Built & bfd) {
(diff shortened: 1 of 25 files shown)


Ericson2314:
I see you moved the declarations to match the location of the definitions. But I must admit, I would prefer keeping ways to process installables separate from the definition of installables themselves. C.f. bcde5456cc3295061a0726881c3e441444dd6680 when I wanted to define derivation.

Less on principle, more for practical reasons, I have been thinking for https://github.com/NixOS/nix/issues/6182 maybe the churn of splitting `libcmd` might not be as bad as I feared. But I would want to keep these building utility function in the "upstream" package, while separating the libexpr-needing and libexpr-agnostic installables. If that still sounds like too much far future stuff, fine, but just something to think about --- things we do now would make it more or less churny accordingly!

src/nix/profile.cc
```diff
@@ -232,54 +253,26 @@ struct CmdProfileInstall : InstallablesCommand, MixDefaultProfile
...
         ProfileManifest manifest(*getEvalState(), *profile);
 
-        std::vector<DerivedPath> pathsToBuild;
+        auto builtPaths = Installable::build(getEvalStore(), store, Realise::Outputs, installables, bmNormal);
 
         for (auto & installable : installables) {
+            ProfileElement element;
+
             if (auto installable2 = std::dynamic_pointer_cast<InstallableFlake>(installable)) {
+                // FIXME: make build() return this?
```
Ericson2314, line 262: I do think what buildPaths returns could help with this or `updateStorePaths`? Each `BuiltPath` contains the paths for the outputs requested, so you don't need to do another `queryDerivationOutputMap` if you can use it.

edolstra: No, the issue here is that we need the resolved flakeref / attrpath.

PR #6198, opened 2022-03-02
Add EvalState::coerceToStorePath() helper

edolstra:
This is useful whenever we want to evaluate something to a store path (e.g. in `get-drvs.cc`).

Extracted from the lazy-trees branch (where we can require that a store path must come from a store source tree accessor).

### DIFF
--- a/src/nix-env/nix-env.cc
+++ b/src/nix-env/nix-env.cc
@@ -211,7 +211,7 @@ static long comparePriorities(EvalState & state, DrvInfo & drv1, DrvInfo & drv2)
 // at a time.
 static bool isPrebuilt(EvalState & state, DrvInfo & elem)
 {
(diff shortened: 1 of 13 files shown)


src/nix-env/nix-env.cc
```diff
@@ -957,6 +953,8 @@ static void queryJSON(Globals & globals, std::vector<DrvInfo> & elems, bool prin
 
 static void opQuery(Globals & globals, Strings opFlags, Strings opArgs)
 {
+    auto & store(*globals.state->store);
```
Ericson2314, line 956: Nit, but @thufschmitt and I have generally been doing the C++11 style
```suggestion
    auto & store { *globals.state->store };
```
which is supposed to be less ambiguous.

Ericson2314 (approved):
Great change!

Issue #6191, opened 2022-03-02
Missing error position pointer for UPDATE (`//`) operator.


DavHau:
**Is your feature request related to a problem? Please describe.**
When using the update operator with wrong types, there is no error position pointer in the output of `--show-trace`

Evaluating the following nix expression:
```nix

let
  s = "hello";

  func =
    arg1: arg2:
      let
        # some other invalid statements errors
        a = s + 1;
        b = s ++ 1;
        c = [] // s;
      in
      # invalid statement triggering the error
      {} // s;

in
  func 1 2
```

... the printed error looks like this:
```
error: value is "hello" while a set was expected

       … while evaluating 'func'

       at /home/grmpf/synced/projects/github/dream2nix/.temp/test.nix:6:11:

            5|   func =
            6|     arg1: arg2:
             |           ^
            7|       let

       … from call site

       at /home/grmpf/synced/projects/github/dream2nix/.temp/test.nix:17:3:

           16| in
           17|   func 1 2
             |   ^

       … while evaluating the file ...
```

The position pointer points at line 6, while the actual error happens in line 14.

This problem is specific to the `//` operator. When evaluating a similar invalid statement with the `++` operator, the error position is shown correctly:
```
error: value is a string while a list was expected

       at /home/grmpf/synced/projects/github/dream2nix/.temp/test.nix:14:7:

           13|       # invalid statement triggering the error
           14|       [] ++ s;
             |       ^
           15|

       … while evaluating 'func'

       at /home/grmpf/synced/projects/github/dream2nix/.temp/test.nix:6:11:

            5|   func =
            6|     arg1: arg2:
             |           ^
            7|       let

       … from call site

       at /home/grmpf/synced/projects/github/dream2nix/.temp/test.nix:17:3:

           16| in
           17|   func 1 2
             |   ^

       … while evaluating the file
```

**Describe the solution you'd like**
Print a stack trace showing the correct position of the error, similar to how it's done with `++` (see last example).

Artturin:
https://github.com/NixOS/nix/pull/6204 merged

Issue #6031, opened 2022-02-01
Immediate error context should be shown without `--show-trace`


lilyball:
**Describe the bug**

The Nixpkgs module system makes extensive use of `builtins.addErrorContext`. This is really helpful! But I don't get any of that info if I don't pass `--show-trace`. Instead I just get an error that points into the expression inside of the `builtins.addErrorContext` call.

I feel like usability here would be increased significantly if the error printing without `--show-trace` would still include the immediate error context if one exists. Which is to say, if showing the full trace would show a context added from `builtins.addErrorContext` as the next step, then go ahead and show that without `--show-trace`.

For example, I have an error right now that looks like

```
error: attribute 'inputs' missing

       at /nix/store/fbcgjqs34vllzzppa1y213fbxx01sxn7-source/lib/modules.nix:386:28:

          385|         builtins.addErrorContext (context name)
          386|           (args.${name} or config._module.args.${name})
             |                            ^
          387|       ) (lib.functionArgs f);
(use '--show-trace' to show detailed location information)
```

The location info here is not terribly helpful. If I pass `--show-trace` the very next entry in the trace is the error context

```
       … while evaluating the module argument `inputs' in "/nix/store/kxhwfcng3hv8kzvzgwaaqv1pv6kspa88-source/common/home-manager/lily/default.nix":
```

This one extra line provides significant value and I wish I didn't have to pass `--show-trace` to see it.

**Steps To Reproduce**

1. Use a module system such as NixOS or nix-darwin
2. Declare an argument on a module that doesn't actually exist
3. Evaluate your config

**Expected behavior**

The error should tell me where the bad argument is without having to use `--show-trace`.

**`nix --version` output**

nix (Nix) 2.5.1

layus:
Could be fixed by https://github.com/NixOS/nix/issues/6200

Artturin:
https://github.com/NixOS/nix/pull/6204 merged

PR #6204, opened 2022-03-04
Add context to better locate runtime coercions

### LINKED ISSUE #6200: Improve location reporting for type and type coercion errors
**Describe the bug**

Nix sometimes reports imprecise error location when a value has the wrong type for an operation. 

In this example, builtins.hasAttr expects a string and receives a boolean as first argument. But the caret points at the builtin itself, making the error message misleading.

```
error: value is a Boolean while a string was expected

       at /nix/store/1bicq9whypjqyy2j6b7qw58jqm9f1py3-source/lib.nix:75:12:

           74|         let inputName = declaredInputs."${name}"; in
           75|         if builtins.hasAttr inputName nickelInputs then
             |            ^
           76|           let input = nickelInputs."${inputName}"; in
```

In the code, we see that we coerce the first arg to a string, and pass as trace position the `pos` of the builtin itsefl (the only we can have at this point in evaluation).

https://github.com/NixOS/nix/blob/3144b373a486bf0d5ff35d3202a658311b26c6f0/src/libexpr/primops.cc#L2206 

**Expected behavior**

Each time we have a coercion, there is a potential for failure. We should ensure that the message is correct by pointing at the right location. From my analysis, it is next to impossible, as the information is lost by the time we know that we have to coerce.

The next best thing is to amend the error message to mention how the error relates to the error location. In this case, say it is the first argument to the builtin under evaluation. This way, the caret can remain on the builtin itself.

``error: in the first argument of builtins.hasAttr: value is a Boolean while a string was expected``

Thus, coercion functions such as `coerceToString`, `evalAttrs` and the like should take an extra, mandatory error message describing what the coerced value means for the current Expr node under evaluation.

We should also ensure that each `Expr` node that may fail in such a way gets a proper error context. That is, we should ensure that a caret points to the expression we refer to in the error message. This is the issue encountered by https://github.com/NixOS/nix/issues/6191.

**Additional context**

https://github.com/NixOS/nix/issues/6191

https://github.com/NixOS/nix/commit/9d67332e4b8d75a101b8cf6e1d7e4dc8e99e964a (thanks @greedy !)

layus:
Add tons of helpful error messages in locations with poor error reporting.
This includes mostly type errors. As these are runtime errors, the precise location of the faulty value is usually lost at that time. This translates textually the type expectations (requirements ?) hard-coded in the sources.

TODOs:
- [X] Also annotate coerceToString
- [X] Find and annotate remaining functions bearing expectations on the values. (I guess this is good enough)
- [X] Wrap some function calls with debug info

Fixes #6200 #6191 #6031


layus:
Have fun by starting with `./outputs/out/bin/nix-instantiate --eval --expr 'builtins.genericClosure false'` and ending with `./outputs/out/bin/nix-instantiate --eval --expr 'builtins.sort (a: b: a.key < b.key) (builtins.genericClosure {startSet = [{key = 1;}]; operator = x: if x.key < 5 then [{key = x.key + 1;} {key = x.key * 2;}] else [];})'` by trying to trigger all the possible type errors in between :-)

layus:
image

:tada:

layus:
More examples: https://maudoux.be/changes.html

layus:
Well, it has some impact on performance, but that remains under control:

```
This PR:
nixos.smallContainer.time     0.762153  0.975931  0.736349  0.775949
nixos.kde.time                1.44574   1.6254    1.48982   1.59174
nixos.lapp.time               1.1987    1.3634    1.19679   1.30787
nix-env.qa.time               6.18346   6.71418   6.32466   6.43977
nix-env.qaDrv.time            44.1625   50.7246   53.6632   46.6902
nix-env.qaAggressive.time     6.31606   6.67477   9.45704   6.60786
nix-env.qaDrvAggressive.time  44.3691   48.075    60.2891   45.648

Baseline:
nixos.smallContainer.time     0.71367  0.705955  0.71822  0.715239
nixos.kde.time                1.37252  1.40253   1.47962  1.38685
nixos.lapp.time               1.12598  1.14188   1.22673  1.12473
nix-env.qa.time               6.12947  6.16216   6.56867  6.22066
nix-env.qaDrv.time            42.1374  42.1028   54.5509  42.0571
nix-env.qaAggressive.time     6.79885  6.00412   9.02374  6.36788
nix-env.qaDrvAggressive.time  45.5595  42.1364   55.0709  42.3918
```

output

Ericson2314:
Did you consider using the existing call stack stuff?

Relatedly, it would be nice to use the positions of the subterm that is trying to be casted. But I see that was discussed at https://github.com/NixOS/nix/issues/6200

I would not be opposed to merging this as-is, and circling back to these issues, but others may disagree.

layus:
> Do you want to update TODOs, mark this draft, or save things for future PRs?

I guess this is as good as I can achieve in reasonable time.

> Did you consider using the existing call stack stuff?

Well, first I thought it would not work. Now I am not too sure anymore. It should be feasible, but it requires shifting around several `pos` and rework how we handle them. It is a bit more involved, and the try/catch syntax adds lots of clutter. Maybe recent C++ lambdas may alleviate this issue. Apparently the try{}catch may have zero cost when not used. That would definitely be the right way forward if it is the case.
So, yes, I think it could be superior if well used. It would certainly avoid mismatch between error messages and positions, as we have now.

That being said, the current version has the advantage that it requires any user of these functions to provide an explanation string, which ensures that we do not inadvertently omit them.
 
> Relatedly, it would be nice to use the positions of the subterm that is trying to be casted. But I see that was discussed at #6200

It is extremely complex in nix, as often there is no good position for values. We do keep some positions for the keys of attribute sets, and sometimes for functions, when feasible.

For example 

```
let
  list = [ 1 "a" {attr="value";} ];
  fun = builtins.map (x: x+1);
  afterList = builtins.seq list;
  getThird = l: builtins.elemAt l 2;
in
  afterList (getThird (fun list ++ [3]))
```

fails with 

```
   1 error: cannot coerce a set to a string
   2 
   3        at /home/layus/projects/nix/complex.nix:4:26:
   4 
   5             3|   list = [ 1 "a" {attr="value";} ];
   6             4|   fun = builtins.map (x: x+1);
   7              |                          ^
   8             5|   afterList = builtins.seq list;
   9 
  10        … while evaluating anonymous lambda
  11 
  12        at /home/layus/projects/nix/complex.nix:4:23:
  13 
  14             3|   list = [ 1 "a" {attr="value";} ];
  15             4|   fun = builtins.map (x: x+1);
  16              |                       ^
  17             5|   afterList = builtins.seq list;
  18 
  19        … from call site
  20 
  21        … while evaluating 'getThird'
  22 
  23        at /home/layus/projects/nix/complex.nix:6:14:
  24 
  25             5|   afterList = builtins.seq list;
  26             6|   getThird = l: builtins.elemAt l 2;
  27              |              ^
  28             7| in
  29 
  30        … from call site
  31 
  32        at /home/layus/projects/nix/complex.nix:8:14:
  33 
  34             7| in
  35             8|   afterList (getThird (fun list ++ [3]))
  36              |              ^
  37             9|
```

And nowhere is there any mention of the list itself, nor the value that failed to be a proper int. And that would be difficult. Would you point `list ++ [3]`, `list = [ 1 "a" {attr="value";} ]`, `[ 1 "a" {attr="value";} ]`, `{attr="value";}`, or something else ?

:bulb: One thing we could do is print the fautly values to help users get a ffeling of what happens.

> I would not be opposed to merging this as-is, and circling back to these issues, but others may disagree.

I think it is ready to merge. Better merge early and improve often :-).

Ericson2314:
> So, yes, I think it could be superior if well used. It would certainly avoid mismatch between error messages and positions, as we have now.

> That being said, the current version has the advantage that it requires any user of these functions to provide an explanation string, which ensures that we do not inadvertently omit them.

Yeah, I think the cast functions should take the argument, and do the try-catch themselves on behalf of the caller. That still ensures one doesn't forget to provide a reason for the cast!

src/libexpr/eval-inline.hh
```diff
@@ -52,26 +52,26 @@ void EvalState::forceValue(Value & v, Callable getPos)
 }
 
 
-inline void EvalState::forceAttrs(Value & v, const Pos & pos)
+inline void EvalState::forceAttrs(Value & v, const Pos & pos, const std::string & errorCtx)
```
pennae, line 55: this (and presumably most string params you've added) should be `const char *`s instead where at all possible. that saves a strlen and copy of the string to dynamic memory on each call, which is presumably makes for the majority of the performance impact.

Ericson2314: `std::string_view` should be fine too?

pennae: `string_view` also calls `strlen` when constructed from a string literal, so that's also mostly unnecessary overhead.

Ericson2314: https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view I don't think that's true?

pennae: constructing from a string literal will use overload 4 in the current code base, which calls a strlen-equivalent via the traits type. a strlen at runtime can only be avoided if the length operation of the traits type is itself constexpr, which doesn't seem to be part of the spec. if it is constexpr it's totally possible that you're right and the strlen is elided. :)

Ericson2314: Whew! I do agree the number of things that need to go right for this to work as intended is very frightening.

I guess always constexpr bind any constant string expression to be safe rather than plopping the expression right in the function argument slot to be safe.

pennae: as long as it doesn't eat our hard-won performance gains we won't complain :) would be nice to have an interface that guarantees things not wasting resources though. `const char *` would do that, or maybe a new type that only accepts `const char (&)[N]` on construction?

Ericson2314: There is `std::array` but I am not sure how to get `std::string_view` from it in a nice way.
`constexpr std::string_view msg = "asdfasdfds";` I think is good enough.

Ericson2314:
Very nice looking!

layus:
Except for painful conflicts... I guess this will have to wait the hackathon tomorrow.

layus:
@Ericson2314 Not sure I will be able to join the hacking session, but if you could find some time to review the changes, that would be much appreciated. I am quite happy with the result, but I lack real tests, and there are so many changes that typos can always slip in.

layus:
@Ericson2314 https://github.com/NixOS/nix/pull/6204/commits/963b8aa39b169bf5c054449ddce39d60faacf298 just for you :-). I hope you like it.

src/libutil/error.cc
```diff
@@ -185,15 +185,15 @@ void printAtPos(const ErrPos & pos, std::ostream & out)
...
                 break;
             }
             case foString: {
-                out << fmt(ANSI_BLUE "at " ANSI_WARNING "«string»:%s" ANSI_NORMAL ":", showErrPos(pos));
+                out << fmt(ANSI_BLUE "  at " ANSI_WARNING "«string»:%s" ANSI_NORMAL ":", showErrPos(pos));
                 break;
             }
             case foStdin: {
-                out << fmt(ANSI_BLUE "at " ANSI_WARNING "«stdin»:%s" ANSI_NORMAL ":", showErrPos(pos));
+                out << fmt(ANSI_BLUE "  at " ANSI_WARNING "«stdin»:%s" ANSI_NORMAL ":", showErrPos(pos));
```
Ericson2314, line 196: Still want this indent?

Ericson2314 (approved):
Other than the indent thing i just commented on, and the remaining `""` in flakes and the new commands someone should fill in (either before this is merged or right after) LGTM!

(It looks like @edolstra has self-assigned, which is good because I don't know much about the evaluator, but earlier I offered to review this for @layus when no one else ways, so I feel obliged to give it a final response.)

src/libutil/error.cc
layus: I changed the format from

```
  ... error foo
  
  at location x:y
  
```

to 


```
  ... error foo
      at location x:y
  
```

The indent does just that.

pennae:
would definitely be interesting to have another look at the performance impact now. not having gone over it again completely this doesn't *look* like it should cause performance regressions, so if it does there may still be room for optimizations.

src/libutil/error.cc
Ericson2314: OK to make this a bit more self documenting I might do
```c++
out << fmt(ELLIPSIS_INDENT ANSI_BLUE "at "
```
where `ELLIPSIS_INDENT` is
```c++
// Enough indent to align with with the `... `
// prepended to each element of thece trace
constexpr std::string_view ELLIPSIS_INDENT = "  ";
```

edolstra (commented):
My main concern with this PR is that it will produce an overwhelming amount of trace info, which may make it hard to find location info that is actually relevant. For instance, it's adding traces to if-then-else expressions, the built-in operators, etc. Do you have examples of how many trace levels this adds to typical Nixpkgs / NixOS evaluation errors?

Maybe some traces should only be shown at a sufficiently high verbosity level. E.g. critical trace messages are derivation evaluations, while an if-then-else is much less important.

tomberek:
Suggestions from NixUX:
- '--show-trace` can become overwhelming, even more than today. Perhaps respect `-v` verbosity flags?
- New messages can be a less severe level in order to limit the amount displayed by default.
- consider a new PR or doing it in the current PR
- It is hard to keep error message testing up to date.

- performance review @pennae?

pennae:
we've run this PR through our perf benchmarks, and while we do see an increase in instructions executed the real-world impact on walltime is in the noise. turning the string_view refs into plain string_views brings measurements even closer together.

performance-wise this looks pretty good. :)

tomberek:
From NixUX:
Thanks @pennae for the perf review. This still requires a way to control and limit the severity of the trace.

src/libcmd/installables.cc
```diff
@@ -527,7 +527,7 @@ ref<eval_cache::EvalCache> openEvalCache(
             auto vFlake = state.allocValue();
             flake::callFlake(state, *lockedFlake, *vFlake);
 
-            state.forceAttrs(*vFlake, noPos);
+            state.forceAttrs(*vFlake, noPos, "While evaluating a cached flake");
```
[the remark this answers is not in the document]
layus: Would you rather have "fetching" or "accessing" ? Beyond the usefulness of the messages, they also help finding the related source code position if someone wants to investigate, so correctness in strange cases like this one that should never happen it not that important.

src/libexpr/eval-cache.cc
```diff
@@ -336,7 +336,7 @@ Value & AttrCursor::getValue()
     if (!_value) {
         if (parent) {
             auto & vParent = parent->first->getValue();
-            root->state.forceAttrs(vParent, noPos);
+            root->state.forceAttrs(vParent, noPos, "While evaluating the parent attr set");
```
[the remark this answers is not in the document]
layus: Again, this is an issue in the eval cache, so the message quality is less important than it's unicity. It should never happen.

src/libexpr/eval.cc
```diff
@@ -1363,7 +1364,12 @@ void EvalState::callFunction(Value & fun, size_t nrArgs, Value * * args, Value &
                 env2.values[displ++] = args[0];
 
             else {
-                forceAttrs(*args[0], pos);
+                try {
+                    forceAttrs(*args[0], lambda.pos, "While evaluating the value passed for this lambda parameter");
```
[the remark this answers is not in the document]
layus: I am hoping that the surrounding context makes it more clear when reading the full trace. Will apply name change.

src/libexpr/eval.cc
```diff
@@ -724,75 +706,92 @@ std::optional<EvalState::Doc> EvalState::getDoc(Value & v)
...
+            const std::string_view & s2,
+            const Symbol & sym,
+            const Pos & p2,
+            const std::string_view & s3))
 {
-    throw EvalError(s, s2);
+    throw TypeError(ErrorInfo {
+        .msg = hintfmt(s, s2, sym),
+        .errPos = pos,
+    }).addTrace(p2, s3);
```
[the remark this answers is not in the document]
layus: :eyes:

src/libexpr/eval.cc
```diff
@@ -1488,8 +1507,12 @@ void EvalState::callFunction(Value & fun, size_t nrArgs, Value * * args, Value &
                heap-allocate a copy and use that instead. */
             Value * args2[] = {allocValue(), args[0]};
             *args2[0] = vCur;
-            /* !!! Should we use the attr pos here? */
-            callFunction(*functor->value, 2, args2, vCur, pos);
+            try {
+                callFunction(*functor->value, 2, args2, vCur, *functor->pos);
+            } catch (Error & e) {
+                e.addTrace(pos, "While calling a functor (an attribute set with a 'functor' attribute)");
```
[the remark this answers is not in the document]
layus: Right. But I think the error message should explain that, because it may not be obvious to a novice user following the trace why an attrset gets applied. These traces are not intended for expert users most of the time. Especially these basic type/usage errors that novices are more prone to do. This is a small atempt at making nix more discoverable by providing context when the error appears.

layus:
> Suggestions from NixUX:
> 
>     * '--show-trace`can become overwhelming, even more than today. Perhaps respect`-v` verbosity flags?

I am not sure this is true. Do we have examples ? The full trace is even bigger, but now it should be readable, with errors pointing at precise locations. See https://gist.github.com/layus/0c4203d41a67326367caf2086f20a2ab#file-basic-diff for an example where the added information allows to understand the error location.

https://gist.github.com/layus/0c4203d41a67326367caf2086f20a2ab#file-poetry-diff shows an example where the trace is made way bigger, and potentially unnecessarily verbose.

I still do not see what this PR should do about it. We had a situation where plain misleading messages were printed. Not it gets easier for newcomers to understand the errors and the required types they need in simple cases. Complex cases become more verbose, but also more easy to understand and to follow, as the increment between the trace steps is smaller. Less wild jumps in the code between two seemingly unrelated parts.

And by default you get the first parts of the trace without needing --show-trace. Meaning that if you really, really want to see the full trace, you should use more or less (pun intended).

I would argue that we should prefer correctness and precision over terseness. This PR cannot reasonably bundle a solution to the very complex issue of printing a meaningful trace in a lazy language.

> 
>     * New messages can be a less severe level in order to limit the amount displayed by default.

I would argue that the old messages should be less chatty ;-). Just kidding, but as above, the code could benefit from a redesign of the traces to show another projection of the full trace. Maybe expressions trace until the first function scope, and then only the function calls stack ? That is way out of context for this PR I think.

> 
>     * consider a new PR or doing it in the current PR
:+1: for new PR.

> 
>     * It is hard to keep error message testing up to date.

It depends on what you want to test. This PR made it sure that more error cases are documented via a mandatory text explanation. Maybe that is the "testing" we need. Any other testing is going to be brittle with respect to format changes.

> 
>     * performance review @pennae?

Thanks @pennae !

layus:
I will merge Eelco's comments, and rebase onto master. Wait for it :-).

src/libexpr/eval.cc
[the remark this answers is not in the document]
layus: Just checked, and the this refers to the exact code position we provide. So the full text reads "While evaluating the value passed for this lambda parameter\n at position... + code snippet"


Maybe "While evaluating the value passed for the lambda parameter\n at position... + code snippet". I guess so.

src/libexpr/eval-cache.cc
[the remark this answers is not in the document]
layus: Oh, this one is more complex. For one, I have to stick to static strings for performance. There is no reason to assemble a string for an error that may never occur. I changed it to "while searching for an attribute".

src/libexpr/primops/fetchTree.cc
```diff
@@ -195,16 +195,14 @@ static void fetch(EvalState & state, const Pos & pos, Value * * args, Value & v,
 
     if (args[0]->type() == nAttrs) {
 
-        state.forceAttrs(*args[0], pos);
-
         for (auto & attr : *args[0]->attrs) {
             std::string n(attr.name);
             if (n == "url")
-                url = state.forceStringNoCtx(*attr.value, *attr.pos);
+                url = state.forceStringNoCtx(*attr.value, *attr.pos, "While evaluating the url we should fetch");
```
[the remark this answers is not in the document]
layus: yes, but it seems fetch can be used elsewhere than in fetchurl, right ?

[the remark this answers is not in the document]
layus: It is at least in the backend of fetchurl, fetchTarball and fetchTree.

layus:
That is it for today. I think it should work now @bburdette :tada: :yum: 

See https://gist.github.com/layus/e10b4fab9938e9a1777530fe296f0623 for an example on how it behaves with a nixos module error.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -4,4 +4,8 @@
   defaults to `false`.  If it is set to `true`, the `hash` and `sha256`
   arguments will be ignored and the resulting derivation will have
   `__impure` set to `true`, making it an impure derivation.
+* Error traces have been reworked to provide detailed explanations and more
+  accurate error locations. A short excerpt of the trace is now shown by
+  default when an error occurs.
 
+>>>>>>> origin/master
```
Artturin, line 11: fail

fricklerhandwerk:
Discussed in Nix team meeting on 2022-11-25:

  - @thufschmitt: decided some time ago that we want to have it. @layus fixed merge conflicts, but a trivial one is still left
  - @edolstra: would this cause a major merge conflict with lazy trees?
      - @thufschmitt: have some some thoughts on that, but should discuss those in a different context

src/nix/eval.cc
```diff
@@ -111,7 +111,7 @@ struct CmdEval : MixJSON, InstallableCommand
 
         else if (raw) {
             stopProgressBar();
-            std::cout << *state->coerceToString(noPos, *v, context);
+            std::cout << *state->coerceToString(noPos, *v, context, "While generating the eval command output");
```
edolstra, line 114:
```suggestion
            std::cout << *state->coerceToString(noPos, *v, context, "while generating the eval command output");
```

edolstra:
@layus Merged, thanks!

roberth:
Unfortunately this PR has caused a regression in 2.13.0.

[Symptoms](https://gist.github.com/GrahamcOfBorg/fc8358d3707cca3d7aa55f38d6b9da4d) include infinite recursions and some sort of regression in the module system.
I've tried to find the root cause by
 - bisecting, but the git history of this change isn't conducive to it
 - reading through the change for two potential causes
   - accidentally forcing more evaluation than necessary
   - memory corruption

I haven't managed to find it, so for the time being, I've reverted this change.
Any help would be much appreciated, as we'd like to have this change back, both because it is a very welcome change and because other PRs would not to have conflicts resolved if we don't manage to merge this soon.

Ericson2314:
Do we have any test cases for these issues? I think a PR with the "revert of the revert" + some *failing* tests would be a good starting point, and then hopefully @layus or another person can pick up the baton from there.

tomberek:
@layus so far I have not seen a behavioral regression, only a change (and improvement in the error message). I added a reproducer to more quickly try out Nix versions: https://github.com/NixOS/nix/pull/7621#issuecomment-1387718654

layus:
So first there is indeed a change where I introduced "frame" traces (pardon the naming here). Think of them as more important trace points, like function calls and builtins.derivation calls.
This explains why the addErrorTrace thing is invisible: it happens after a function call in the trace. I have confirmed that removing this notion of frame traces fixes the nixpkgs tests errors. There are two solutions
1) disable frame traces
2) make addErrorTrace a frame trace.

I think 2 is better because we keep the terseness of the output, and it even makes sense, as a message that a human took time to write for debugging purposes deserves to be printed in the end :-).
Working on it.

tomberek:
Agreed, 2 is likely to be a better match for user expectations.

PR #6205, opened 2022-03-04
Add some suggestions in the CLI in case of possible typos

thufschmitt:
Try and be helpful when the user types something invalid in the cli by suggesting the closest valid possibilities.

This currently handle command names and flake attribute paths, but it can be extended to support other kind of inputs.
It’s also not cli-specific so it could be used in other places, like inside the evaluator (something like `attribute 'fob' missing. maybe you meant 'foo'`).

### examples

image


src/libutil/or-suggestions.hh
```diff
@@ -0,0 +1,64 @@
+#include "suggestions.hh"
+#include "error.hh"
+
+namespace nix {
+
+// Either a value of type `T`, or some suggestions
+template<typename T>
+class OrSuggestions {
```
edolstra, line 8: I'm not sure if this class adds much value over plain `std::variant<T, Suggestions>`. In fact it seems to add more cognitive overhead over the `std::variant`...

src/libutil/or-suggestions.hh
```diff
@@ -0,0 +1,64 @@
...
+        res.raw = s;
+        return res;
+    }
+
+    static OrSuggestions<T> failed()
+    {
+        return OrSuggestions<T>::failed(Suggestions{});
+    }
+
+    const Suggestions & get_suggestions()
```
edolstra, line 53:
```suggestion
    const Suggestions & getSuggestions()
```

src/libutil/suggestions.cc
```diff
@@ -0,0 +1,107 @@
+#include "suggestions.hh"
+#include "ansicolor.hh"
+#include "util.hh"
+#include <algorithm>
+
+namespace nix {
+
+int levenshtein_distance(std::string_view first, std::string_view second)
```
edolstra, line 8:
```suggestion
int levenshteinDistance(std::string_view first, std::string_view second)
```

src/libutil/suggestions.hh
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * A potential suggestion for the cli interface.
+ */
+class Suggestion {
+public:
+    int distance; // The smaller the better
+    std::string suggestion;
+
+    std::string pretty_print() const;
```
edolstra, line 19:
```suggestion
    std::string to_string() const;
```
or maybe `operator <<`, then it works in `fmt()` automatically.

src/libutil/suggestions.cc
```diff
@@ -0,0 +1,107 @@
...
+{
+    return ANSI_WARNING + filterANSIEscapes(suggestion) + ANSI_NORMAL;
+}
+
+std::string Suggestions::pretty_print() const
+{
+    switch (suggestions.size()) {
+        case 0:
+            return "";
+            break;
```
edolstra, line 78:
```suggestion
```

src/libutil/suggestions.cc
```diff
@@ -0,0 +1,107 @@
...
+        case 1:
+            return suggestions.begin()->pretty_print();
+            break;
```
edolstra, line 81:
```suggestion
```

src/libutil/suggestions.cc
```diff
@@ -0,0 +1,107 @@
...
+        default:
+            {
+                std::string res = "one of ";
```
edolstra, line 84:
```suggestion
        default: {
            std::string res = "one of ";
```
etc.

edolstra (approved):

src/libutil/or-suggestions.hh
thufschmitt: I started with a `std::variant`, but it was a bit annoying because it’s not possible to add methods to it, meaning that I had to resort to inline `std::visit(overload{})`or `std::get_if` everywhere, which was quite unfriendly.

Alternatively, I can switch back to a plain variant, and add some plain functions working on it (which is visually a bit less nice, but avoids this wrapping)

src/libutil/suggestions.hh
thufschmitt: Is it OK to have `to_string` or `<<` also print some ANSI escape codes?

thufschmitt: (I’ve made the change anyways since I guess the answer is yes)

src/libutil/or-suggestions.hh
edolstra: Just change it into an assertion since it's clearly a logic error if that happens.

thufschmitt: Right, makes sense.

I’ve actually replaced the whole thing by a call to `std::get` which will raise its own exception if the bad variant is provided

PR #6216, opened 2022-03-08
Add some suggestions to the evaluator

thufschmitt:
Make the evaluator show some suggestions when trying to access an
invalid field from an attrset.

```console
$ nix eval --expr '{ foo = 1; }.foa'
error: attribute 'foa' missing

       at «string»:1:1:

            1| { foo = 1; }.foa
             | ^
       Did you mean foo?
```

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -729,7 +729,7 @@ LocalNoInlineNoReturn(void throwEvalError(const char * s, const std::string & s2
 
 LocalNoInlineNoReturn(void throwEvalError(const Pos & pos, const Suggestions & suggestions, const char * s, const std::string & s2))
 {
-    throw EvalError({
+    throw EvalError(ErrorInfo {
         .msg = hintfmt(s, s2),
         .errPos = pos,
         .suggestions = suggestions,
@@ -738,7 +738,7 @@ LocalNoInlineNoReturn(void throwEvalError(const Pos & pos, const Suggestions & s
 
 LocalNoInlineNoReturn(void throwEvalError(const Pos & pos, const char * s, const std::string & s2))
 {
-    throw EvalError({
+    throw EvalError(ErrorInfo {
         .msg = hintfmt(s, s2),
         .errPos = pos
     });
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -782,6 +782,16 @@ LocalNoInlineNoReturn(void throwTypeError(const Pos & pos, const char * s, const
     });
 }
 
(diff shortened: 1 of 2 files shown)


bjornfor:
I love this!

>            1| { foo = 1; }.foa
>             | ^
>       Did you mean foo?

And I would love it even more if it pointed to the start of `foa` rather than the attrset. (At least that feels more natural to me.)

SuperSandro2000:
How would that handle the following cases?

```
$ nix eval --expr '{ foo = 1; foa = 2; fob = 3; foc = 4; }.fod'
```

```
$ nix eval --expr '{ foo = 1; }.goo'
```

thufschmitt:
Grmph failing darwin build :angry: 

> And I would love it even more if it pointed to the start of foa rather than the attrset. (At least that feels more natural to me.)

Yes, indeed :) (though that’s slightly orthogonal to this issue)

> How would that handle the following cases?

```console
$ nix run nix/eval-suggestions -- eval --expr '{ foo = 1; foa = 2; fob = 3; foc = 4; }.fod'
error: attribute 'fod' missing

       at «string»:1:1:

            1| { foo = 1; foa = 2; fob = 3; foc = 4; }.fod
             | ^
       Did you mean one of foa, fob, foc or foo?
$ nix run nix/eval-suggestions -- eval --expr '{ foo = 1; }.goo'                                                 
error: attribute 'goo' missing

       at «string»:1:1:

            1| { foo = 1; }.goo
             | ^
       Did you mean foo?
```

thufschmitt:
I’ve also added some suggestions for function calls on f6078e474d5fc41c8a7f683865d60490bf0c7040 (cc @Infinisil ).  this only includes extra arguments and only if there’s no missing one:

```console
# Gently displays some suggestions
$ nix eval --expr '({ foo ? 1 }: foo) { fob = 2; }'                
error: anonymous function at (string):1:2 called with unexpected argument 'fob'

       at «string»:1:1:

            1| ({ foo ? 1 }: foo) { fob = 2; }
             | ^
       Did you mean foo?
# Doesn’t because it first fails because of the missing argument
$ nix eval --expr '({ foo }: foo) { fob = 2; }'                
error: anonymous function at (string):1:2 called without required argument 'foo'

       at «string»:1:1:

            1| ({ foo }: foo) { fob = 2; }
             | ^
```

I’d like to have something like https://github.com/NixOS/nixpkgs/pull/79877, but I’m not sure that we can get some good error messages in general with this: The implementation in the linked PR suggests changing the formal argument names (rather than the actual arguments), which makes sense for `callPackage` but not always.

PR #6218, opened 2022-03-08
reduce the size of Attr from 3 pointers to 2 on 64 bit machines

pennae:
this changes position handling to use a table similar to SymbolTable for storing position information, and changes SymbolTable itself to require less memory per Symbol at the expense of lookup convenience. with both of these tables using 32 bit indices we can pack Attr more tightly, reducing GC heap size and allocations by 10-15%. reduced GC load also gives us an eval performance improvement on the same order for free.

depends on #5865


pennae:
rebased to current master now. :)

lovesegfault:
Do you have benchmarks on the eval perf improvement here?

pennae:
@lovesegfault benchmarks results are in the final commit message, copied here for convenience:
```

before (on memory-friendliness):

  Benchmark 1: nix flakes search --no-eval-cache --offline ../nixpkgs hello
    Time (mean ± σ):      6.960 s ±  0.028 s    [User: 5.832 s, System: 0.897 s]
    Range (min … max):    6.886 s …  7.005 s    20 runs

  Benchmark 2: nix flakes eval -f ../nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix
    Time (mean ± σ):     328.1 ms ±   1.7 ms    [User: 295.8 ms, System: 32.2 ms]
    Range (min … max):   324.9 ms … 331.2 ms    20 runs

  Benchmark 3: nix flakes eval --raw --impure --expr 'with import <nixpkgs/nixos> {}; system'
    Time (mean ± σ):      2.688 s ±  0.029 s    [User: 2.365 s, System: 0.238 s]
    Range (min … max):    2.642 s …  2.742 s    20 runs

after:

  Benchmark 1: nix flakes search --no-eval-cache --offline ../nixpkgs hello
    Time (mean ± σ):      6.902 s ±  0.039 s    [User: 5.844 s, System: 0.783 s]
    Range (min … max):    6.820 s …  6.956 s    20 runs

  Benchmark 2: nix flakes eval -f ../nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix
    Time (mean ± σ):     330.7 ms ±   2.2 ms    [User: 300.6 ms, System: 30.0 ms]
    Range (min … max):   327.5 ms … 334.5 ms    20 runs

  Benchmark 3: nix flakes eval --raw --impure --expr 'with import <nixpkgs/nixos> {}; system'
    Time (mean ± σ):      2.330 s ±  0.027 s    [User: 2.040 s, System: 0.234 s]
    Range (min … max):    2.272 s …  2.383 s    20 runs
```

src/libexpr/attr-set.hh
```diff
@@ -16,10 +16,10 @@ struct Value;
 struct Attr
 {
     SymbolPtr name;
-    Value * value;
     PosPtr pos;
+    Value * value;
```
andir, line 20: Maybe add a comment that the order of these was choosen purposely?

pennae: added a comment describing the reasoning here. not sure an assertion has much value though, do you have a scenario in mind that might interfere here?

pennae:
puzzling that the `eval-okay-xml` test fails on macos, it runs fine here on linux :confused:

pennae:
@edolstra this has been getting a lot of merge conflicts recently, does it make sense to keep resolving them or should we wait until this is officially approved?

tomberek:
NIX_SHOW_STATS trimmed for brevity
```
before                                             after
{                                                  {
  "envs": {                                          "envs": {
    "bytes": 146264296                                 "bytes": 146264296
  "list": {                                          "list": {
    "bytes": 26839376,                                 "bytes": 26839352,
  "values": {                                        "values": {
    "bytes": 362397240                                 "bytes": 362397456
  "symbols": {                                       "symbols": {
    "bytes": 3618409                                   "bytes": 1382817
  "sets": {                                          "sets": {
    "bytes": 1100289680,                               "bytes": 747502224,
  },                                                 },
  "sizes": {                                         "sizes": {
    "Env": 16,                                         "Env": 16,
    "Value": 24,                                       "Value": 24,
    "Bindings": 16,                                    "Bindings": 16,
    "Attr": 24                                         "Attr": 16
  },                                                 },
  "gc": {                                            "gc": {
    "heapSize": 1561522176,                            "heapSize": 1158864896,
    "totalBytes": 1960805392                           "totalBytes": 1566740096
  }                                                  }
```

pennae:
rebased again. also renamed `PosPtr` to `PosIdx` and `SymbolPtr` to `SymbolIdx` since neither of them really act like pointers at all.

Artturin:
@edolstra @thufschmitt @Ericson2314

Ericson2314:
I am not up to date on these things and so could not review very way, but perhaps @layus can after https://github.com/NixOS/nix/pull/6204. I think it's good to have the people that do performance and diagnostic quality reviewing each other, as those goals in the general case have some tension.

(To be clear, I don't see a sign of diagnostic quality regression in this case.)

src/libexpr/attr-set.hh
```diff
@@ -15,12 +15,16 @@ struct Value;
...
 {
-    Symbol name;
+    /* the placement of `name` and `pos` in this struct is important.
+       both of them are uint32 wrappers, they are next to each other
+       to make sure that Attr has no padding on 64 bit machines. that
+       way we keep Attr size at two words with not wasted space. */
+    SymbolIdx name;
+    PosIdx pos;
     Value * value;
-    ptr<Pos> pos;
```
roberth, line 20: `ptr<T>` was introduced for pos here. I don't think we currently have a use for a non-nullable regular pointer type, so it's possible to remove it.

pennae:
installer test has been stuck for nine days. nice.

src/libexpr/parser.y
```diff
@@ -369,7 +369,7 @@ expr_function
   : ID ':' expr_function
     { $$ = new ExprLambda(CUR_POS, data->symbols.create($1), 0, $3); }
   | '{' formals '}' ':' expr_function
-    { $$ = new ExprLambda(CUR_POS, data->symbols.create(""), toFormals(*data, $2), $5); }
+    { $$ = new ExprLambda(CUR_POS, {}, toFormals(*data, $2), $5); }
```
layus, line 372: This could use a comment for future readers. A dummy function with a proper name would help a lot.

src/libexpr/eval.hh
```diff
@@ -249,6 +249,45 @@ public:
...
+    void throwAssertionError(const Pos & pos, const char * s, const std::string & s1) const;
+    [[gnu::noinline, gnu::noreturn]]
+    void throwUndefinedVarError(const Pos & pos, const char * s, const std::string & s1) const;
+    [[gnu::noinline, gnu::noreturn]]
+    void throwMissingArgumentError(const Pos & pos, const char * s, const std::string & s1) const;
+
+    [[gnu::noinline]]
+    void addErrorTrace(Error & e, const char * s, const std::string & s2) const;
+    [[gnu::noinline]]
+    void addErrorTrace(Error & e, const Pos & pos, const char * s, const std::string & s2) const;
```
layus: This is the big pain point for me. Now these hacky functions leak into a much used, public header. I understand the purpose of these functions, but I would really like to find a more DRY alternative to this.

src/libexpr/eval-cache.cc
```diff
@@ -635,7 +625,7 @@ bool AttrCursor::isDerivation()
 
 StorePath AttrCursor::forceDerivation()
 {
-    auto aDrvPath = getAttr(root->state.sDrvPath, true);
+    auto aDrvPath = getAttr("drvPath", true);
     auto drvPath = root->state.store->parseStorePath(aDrvPath->getString());
```
layus, line 629: Do I read correctly that this disables the former optimization of turning common strings into symbols ?

layus (approved):
No real concern here for me, except for the slight readability decrease that often comes with performance tuning.

As this is a huge amount of work, I think a decision should be taken soon on whether we want to continue in this direction or not.

src/libexpr/parser.y
pennae: how about
```suggestion
    { $$ = new ExprLambda(CUR_POS, Symbol::none, toFormals(*data, $2), $5); }
```
which would be a new constant akin to `noPos`?

src/libexpr/eval-cache.cc
pennae: in a manner of speaking. the optimization doesn't have a great impact here since the symbol is soon turned into a string again and used in an sqlite query, so the extra symbol table lookup we may have to do isn't very expensive in comparison.

src/libexpr/eval.hh
pennae: could also make them private and make all the *Expr classes friends of EvalState, but that feels *even more* of a hack than this. we might be able to make them `inline` instead and define them only in `eval-state.cpp` to make all other uses compile time errors, but that only adds to the hackiness as well :confused:

src/libexpr/attr-set.hh
andir: Since it has a signitifcant impact I think the trivial issue could be that someone adds another field here without fully understanding the consequences. Also just some refactoring (prior to your comment) might have lead to someone moving them around (for various reasons) without knowing that the layout was chosen deliberately.

src/libexpr/eval-cache.cc
layus: Right, I guess the few places where you had to move back from symbols to strings are the ones where you do not have access to a symbol table ? I see hot paths like the one in pirmops.cc still use the state->sFooBar "global" symbols.

pennae: we do have access to the symbol in this location too (through `root->state`), in this case it just seemed more readable to not use it. wouldn't be a problem to use it here as well though if it bothers anyone.

src/libexpr/attr-set.hh
pennae: added an assertion for the expected size with a warning about why it's important.

src/libexpr/parser.y
pennae: added another constructor to ExprLambda instead because C++ really doesn't like constexpr data members of a type to be declared inside the type, and adding a `noSymbol` to the surrounding namespace for a single use seemed rather cluttering.

roberth:
Does the memory usage regress when evaluating on 32-bit platforms?

pennae:
haven't tested 32 bit, but it should also decrease. Pos now elides a Symbol (aka pointer) per instances at the cost of two ints and a std::string per file. that should break even at five positions per file on average for 32 bit machines, and if nixpkgs is anything to go by that shouldn't be hard to achieve :)

src/libexpr/eval-cache.cc
layus: Just trying to understand the code really, so not bothered by anything here. From the diff it looks like a de-optimisation, but it is indeed restricted to a few places. All fine for me.

src/libexpr/eval.hh
layus: How about going the other way around, and lookup the {Symbol,Position}Idx _before_ passing them to throw*Error ? I My understanding is that the stack usage should not change. If this machinery is indeed about reducing stack space, then it is about the same.

That would avoid having to expose all these dummy and ill-defined functions in the headers, and let them settle when we should find a way to get rid of them.

Obviously it would make the code a bit more verbose, due to the repeated lookups, but that is not a big deal for that code in particular. If we wanted it terse, we should first take care of the try/catch and re-throwing bits, that clutter the eval and primops files.

pennae: we did that at first, but it does clutters the call sites something fierce. this was mostly a readability choice as well and not something we're particularly attached to or did for runtime metrics. the symbols table is a public member of EvalState though, so we *could* make that parameter explicit and remove the declarations from the header again. (which then creates another problem: where does the argument go? first, last, somewhere in between?)

src/libexpr/nixexpr.hh
```diff
@@ -24,31 +27,113 @@ MakeError(RestrictedPathError, Error);
...
+
+    struct Offset {
+        uint32_t line, column;
+    };
+
+private:
+    static constexpr unsigned CHUNK_SIZE = 8192;
+
+    std::vector<Origin> origins;
+    std::vector<std::vector<Offset>> offsets;
```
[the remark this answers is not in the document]
thufschmitt: (and similar for the symbols ofc)

thufschmitt (commented):
Thanks, the speedup and memory reduction is much appreciated.

The diff is quite massive, so I might have missed something, but apart from the few comments I’ve left, looks good

src/nix/repl.cc
```diff
@@ -459,21 +459,21 @@ bool NixRepl::processLine(std::string line)
...
-            pos = v.lambda.fun->pos;
-        } else {
-            // assume it's a derivation
-            pos = findPackageFilename(*state, v, arg);
-        }
+        const auto [file, line] = [&] () -> std::pair<std::string, uint32_t> {
+            if (v.type() == nPath || v.type() == nString) {
+                PathSet context;
+                auto filename = state->coerceToString(noPos, v, context);
+                return {state->symbols.create(*filename), 0};
```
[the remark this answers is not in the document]
pennae: for the moment yes, but that's just to keep the visible diff small-ish. a [later commit](02f523717a5da15c11913932bd01603db2724858) removes the symbol creation here

src/libexpr/nixexpr.hh
```diff
@@ -24,31 +27,113 @@ MakeError(RestrictedPathError, Error);
...
+class PosTable
+{
+public:
+    class Origin {
+        friend PosTable;
+    private:
+        // must always be invalid by default, add() replaces this with the actual value.
+        // subsequent add() calls use this index as a token to quickly check whether the
+        // current origins.back() can be reused or not.
+        mutable uint32_t idx = std::numeric_limits<uint32_t>::max();
```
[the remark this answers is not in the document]
pennae: no, instances of this type double as origin information *within* the symbol table and as cookie values for *users* of the symbol table to enable reuse of origin entries. the idea is to have each site that creates positions allocate an `Origin` with an invalid `idx`, which the symbol table fills with a valid value when a position is created. each subsequent position creation need only check that the `idx` of the cookie is the `idx` of the last origin in the table and reuse that origin if it is, otherwise create a new origin and change the `idx` of the cookie. it turns out doing it this way is a lot more efficient than having each position creation pass the full file/origin pair and check those.

src/libexpr/nixexpr.hh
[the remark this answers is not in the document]
pennae: they do once you have a lot of them. allocating multiple fixed-size vectors is faster than reallocating a vector that keeps growing and has less overhead once the tables get large, at the expense of an additional indirection during lookup. extracting that to a separate type won't be a problem.

src/nix/repl.cc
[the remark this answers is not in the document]
thufschmitt: Oh yes sorry :facepalm:

src/libexpr/nixexpr.hh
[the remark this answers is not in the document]
thufschmitt: Oh yes, didn’t think about nesting the parsing (dunno whether that happens in practice, but it’s definitely safer to assume that it can)

pennae:
addressed @thufschmitt review comments :)

pennae:
and rebased to master once more, for good measure

thufschmitt (approved):
> addressed @thufschmitt review comments :)

Thanks :) I’ve pushed a couple of commits to move ChunkedVector to its own header and add some small tests. I’ll merge as soon as the CI is green

pennae:
not having high hopes for CI to be honest, the installer test has been hating this PR for weeks xD

pennae:
looks like the auto-merge was a bit over-eager again, the macos build failed:

```
 nix> In file included from src/libutil/tests/chunked-vector.cc:1:
nix> src/libutil/chunked-vector.hh:3:10: fatal error: 'bits/stdint-uintn.h' file not found
nix> #include <bits/stdint-uintn.h>
nix>          ^~~~~~~~~~~~~~~~~~~~~
nix> 1 error generated.
```

thufschmitt:
Duh, that thing is indeed totally broken :angry:

dtzWill:
Hi, I was wondering if `std::deque` was considered/would work instead of ChunkedVector? deque is basically a vector of vectors and is useful for same reasons I think ChunkedVector was introduced?

(modulo the forEach helper, but `std::for_each` will do the trick)

pennae:
@dtzWill was considered but (maybe mistakenly) deemed unsuitable because there are no allocation size guarantees for its chunks. part of the reason for chunking is to keep a constant maximum memory overhead for unused capacity, if we just didn't find the way to get that done with a deque then a deque would work just as well

src/nix/app.cc
```diff
@@ -85,9 +85,9 @@ UnresolvedApp Installable::toApp(EvalState & state)
 
     else if (type == "derivation") {
         auto drvPath = cursor->forceDerivation();
-        auto outPath = cursor->getAttr(state.sOutPath)->getString();
-        auto outputName = cursor->getAttr(state.sOutputName)->getString();
-        auto name = cursor->getAttr(state.sName)->getString();
+        auto outPath = cursor->getAttr("outPath")->getString();
+        auto outputName = cursor->getAttr("outputName")->getString();
+        auto name = cursor->getAttr("name")->getString();
```
edolstra, line 90: I don't understand this and similar changes. Not that this is on the hot path, but isn't it a worse for performance to replace a lookup of a symbol by a lookup of a string?

edolstra:
Question: we now have two ways to uniquely refer to symbols, namely `Symbol` and `SymbolIdx`, which is confusing. Which one should be used in APIs? Given that it's possible to efficiently convert a `SymbolIdx` to a `Symbol` but not the other way around, does it even make sense to expose `Symbol`?

src/libexpr/attr-path.cc
```diff
@@ -36,18 +36,18 @@ std::vector<Symbol> parseAttrPath(EvalState & state, std::string_view s)
 {
     std::vector<Symbol> res;
     for (auto & a : parseAttrPath(s))
-        res.push_back(state.symbols.create(a));
+        res.emplace_back(a);
```
edolstra, line 39: I'm not sure I understand this. How does this guarantee that the symbol is unique (i.e. pointer-comparable)?

src/nix/app.cc
pennae: we changed these for readability. ioo it was preferrable to do it like this instead of as
```
auto name = cursor->getAttr(state.symbols[state.sName])->getString();
```
but ultimately both of them are totally valid.

src/libexpr/attr-path.cc
pennae: it does not, as of this PR `Symbol` does not confer a uniqueness guarantee, only a "this is a nix identifier" semantics. we can remove `Symbol` and rename `SymbolIdx -> Symbol` to restore that guarantee to the name if that's preferrable

roberth:
@pennae what do you think of a scheme like this, that replaces the position table by a clever indexing scheme?
 - https://github.com/NixOS/nix/issues/8208

It would enable some improvements to error messages; see the github reverse links to the issues that come after it.

PR #6221, opened 2022-03-08
Add Store::buildPathsWithResults()

edolstra:
This function is like `buildPaths()`, except that it returns a vector of `BuildResult`s containing the exact statuses and output paths of each derivation / substitution. This is convenient for functions like `Installable::build()`, because they then don't need to do another series of calls to get the outputs of CA derivations. It's also a precondition to impure derivations, where we *can't* query the output of those derivations since they're not stored in the Nix database.

Note that `PathSubstitutionGoal` can now also return a `BuildStatus`.

### DIFF
--- a/src/libstore/build-result.hh
+++ b/src/libstore/build-result.hh
@@ -28,6 +28,7 @@ struct BuildResult
         LogLimitExceeded,
         NotDeterministic,
         ResolvesToAlreadyValid,
+        NoSubstituters,
     } status = MiscFailure;
     std::string errorMsg;
 
@@ -63,14 +64,25 @@ struct BuildResult
        non-determinism.) */
     bool isNonDeterministic = false;
 
+    /* For derivations, the derivation path and the wanted outputs. */
+    std::optional<StorePath> drvPath;
     DrvOutputs builtOutputs;
+
+    /* For substitutions, the substituted path. */
+    std::optional<StorePath> outPath;
 
     /* The start/stop times of the build (or one of the rounds, if it
        was repeated). */
     time_t startTime = 0, stopTime = 0;
 
-    bool success() {
+    bool success()
+    {
         return status == Built || status == Substituted || status == AlreadyValid || status == ResolvesToAlreadyValid;
+    }
+
+    void rethrow()
+    {
+        throw Error("%s", errorMsg);
     }
 };
 
--- a/src/libstore/build-result.hh
+++ b/src/libstore/build-result.hh
@@ -64,12 +64,12 @@ struct BuildResult
        non-determinism.) */
     bool isNonDeterministic = false;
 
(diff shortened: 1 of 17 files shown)


src/libstore/remote-store.hh
```diff
@@ -171,6 +176,9 @@ private:
 
     std::atomic_bool failed{false};
 
+    void copyDrvsFromEvalStore(
+        const std::vector<DerivedPath> & paths,
+        std::shared_ptr<Store> evalStore);
```
Ericson2314, line 181: I've been meaning to bring this up, can we call it "DrvStore" or similar in libstore? "Eval" is not a concept libstore should be aware of.

Also, can we do
```suggestion
    void copyDrvsFromEvalStore(
        const std::vector<DerivedPath> & paths,
        Store & evalStore);
```
I am hoping to cut down on the extraneous use of shared pointers (for ergonomics, more than efficiency).

Ericson2314:
@edolstra C.F. https://github.com/NixOS/nix/pull/4588 which was trying to get us closer to this this.

@thufschmitt was worry about the serialization helpers trapping us WRT protocol versions, but after starting #1165 I decided it would be better to completely separate the legacy and new protocols so protocol version could be passed to the serializers, ameliorating that issue.

edolstra:
Alternatively we could serialize `BuildResult` as JSON, like `Realisation`.

src/libstore/remote-store.hh
edolstra: The only problem with that is that if you need a `shared_ptr`, it's hard to get it from a `&`.

Ericson2314:
OK made the protocol versioning PR https://github.com/NixOS/nix/pull/6223

> Alternatively we could serialize `BuildResult` as JSON, like `Realisation`.

Sure, but I don't think it matters that much. JSON is nice enough, sure, but it doesn't save us from all versioning considerations, just adding new fields to objects. I am completely ambivalent.

src/libstore/build-result.hh
```diff
@@ -63,15 +64,26 @@ struct BuildResult
        non-determinism.) */
     bool isNonDeterministic = false;
 
+    /* For derivations, the derivation path and the wanted outputs. */
+    std::optional<StorePath> drvPath;
     DrvOutputs builtOutputs;
 
+    /* For substitutions, the substituted path. */
+    std::optional<StorePath> outPath;
+
```
Ericson2314, line 73: I don't really like this because it's possible for both to be `nullopt`. I was about to propose `std::variant<StorePath, StorePath>`, but why don't we just return the original `DerivedPath` used in the request? That's exactly correct!

edolstra: Thanks, done.

PR #6223, opened 2022-03-09
Give `nix daemon` and `nix-store --serve` protocols separate serializers with version info

Ericson2314:
This allows that the versioning parts for it to actually be correct.

Version agnostic serializers are deduped in `common-protocol.hh`/ `namespace nix::common_proto`, while version-specific ones written separately in `{worker,serve-proto.hh`/`namespace nix::{worker,serve}_proto`.

# Motivation

For me, these are main motivations for this:

1. No more perverse incentives. https://github.com/NixOS/nix/commit/fe1f34fa60ad79e339c38e58af071a44774663f7 did some awkward things because the serializers did not store the version. I don't want anyone making changes to be pushed towards keeping the serialization logic with the core data types because it's easier.

2. I am confident with the incremental approach I tried out in https://github.com/NixOS/nix/pull/6134, we can get Hydra using `ssh-ng://` (and the `Store` abstraction) pretty quickly. After that, I hope we can deprecate `ssh://` for a few releases and then remove it. I don't want to worry about accidentally changing it / breaking it during that time, and having separated codes ensures it stays out of the way!

3. In general, for anyone else implementing the daemon protocol, the more we factor out logic from `daemon.cc`, `remote-store.cc`, etc., the easier things will be to understand. There are a number of projects looking to reimplement the daemon protocol, and this helps them.

The last two commits are especially notable, because they use the serialization refactoring to dedup and factor out / improves factored out seralizers for:

 - `BuildResult`
 - `PathInfo`

This demonstrates the sort of cleanup that this is supposed t enable

# Context

The history in this PR has been delicately preserved to allow reviewing commit-by-commit. Please do so!

If a certain prefix of commits seems good, we can always merge them first too.

depends on #8360
depends on #9197

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`


Ericson2314:
Rebased this on top!

thufschmitt:
I’m quite scared by these raw `#include`-d files that have a different semantic depending on where they are included.

A nicer way to do that might be to
1. Replace the `worker_proto` and `serve_proto` namespaces by classes
2. Add these classes as argument to all the templates defined here (or maybe use plain subclassing?) so that they can be reused “normally”

Ericson2314:
@thufschmitt Those files are gone! Also, less boilerplate to handle versions now.

Combined with https://github.com/NixOS/nix/pull/6312 and likewise for valid path info, this will be much nicer.

thufschmitt:
I do like the overall idea, but I find this really complex, and I don’t see what justifies all this this complexity. Do we really need all that cpp and namespace mangling? I’d much rather use plain inheritance or templating, even it it comes with more boilerplate

Ericson2314:
@thufschmitt What CPP mangling is there now?

I agree the namespace tricks are unfortunate --- if we have just had Rust's Traits this would all be so much easier --- but I thought quite long about alternatives and I can't think of much different unless it become unextendable (e.g. fit set of method on some serializer, lots of virtual), which doesn't sit right with me but also is just feels like far more churn.

The namespace tricks are roughly how nlohmann json works, too, fwiw.

I would like to do this more more less as-is, and just hope we can get rid of the `nix-store --serve` protocol soon, after which things are simpler again.

thufschmitt:
> What CPP mangling is there now?

No CPP mangling, just namespace mangling ;).
But the `WRAP_META_PROTO`, `MAKE_PROTO` and friends intuitively feel like something that would better be replaced by a class holding both the `read` and `write` methods.

> I can't think of much different unless it become unextendable (e.g. fit set of method on some serializer, lots of virtual)

Is it that bad? After all we’re in control of everything, so it’s fine to have one master-class declaring all the methods (beside it’s easier to have a view of everything that’s implemented)

(Just to be clear: I’m not saying that this approach is wrong. There’s a lot of nasty complexity here and I’m all ready to believe that we don’t have a choice. But in that case it deserves a clear justification)

Ericson2314:
> Is it that bad? After all we’re in control of everything, so it’s fine to have one master-class declaring all the methods (beside it’s easier to have a view of everything that’s implemented)

You are right its not a problem in practice. Just the non-modularity of it sits very uncomfortably with me. If we ever want to make a greenfield new protocol that is e.g. Protobuf or JSON, and easier for other tools I understand, to those tools usually take the open world approach so I feel like it's good to match it too.

The tricky parts here are trying to share code between daemon protocol / serve protocol / hooks / import/export / whatever else. As soon as we stop/don't need to do that, it gets easier again.

thufschmitt:
> If we ever want to make a greenfield new protocol that is e.g. Protobuf or JSON, and easier for other tools I understand, to those tools usually take the open world approach so I feel like it's good to match it too.

Well, these new protocols wouldn’t share anything with the current one, would they?

> The tricky parts here are trying to share code between daemon protocol / serve protocol / hooks / import/export / whatever else. As soon as we stop/don't need to do that, it gets easier again.

Unfortunately that will probably have to stay for at least a couple years, so we’d better not treat this as too temporary

Ericson2314:
@elaforge Curious what you think of the last two commits here, as someone trying to reimplement `wopBuildDerivation`.

elaforge:
> @elaforge Curious what you think of the last two commits here, as someone trying to reimplement wopBuildDerivation.

It looks like you're moving the BuildResult serialize from inline code into the thing_proto::read "typeclass", yes?  Seems fine in general.  I did notice that nix is inconsistent about when it inlines and when it uses the typeclass thing, and mostly converted to typeclasses in haskell.  Actually, for me inline is easier to read since you don't have to grope around for the actual implementation, which c++ doesn't make easy since it's all overloading the same name.  But once I understood the (undocumented?) pseudo-typeclass style and gradually located where all the bits were scattered (some of them hidden behind typedefs, like `map<k, v>`), that wasn't the main problem, but rather stuff like undocumented processStderr and the custom BasicDerivation serialization (not just "here's the text of the drv").  But since I gather this is a refactor, not actually changing the protocol, not much can be done about those things except documentation.

I've only worked on the daemon protocol for now, not the `--serve` one, but of course reusing more code between them is good.  I expect the main difference between them will be that they present different bugs and I'll use the one that has fewer serious bugs.  For instance, I'm not sure if the 8k log limit is present in the daemon protocol or the `--serve` one or both or neither.

One thing I wonder, if this is a lipstick on a pig type thing?  The advantage of the ad-hoc protocol is it can be simple and direct.  If there's going to be a whole framework it eats away at the single advantage.  Then it seems more appealing to pick an existing framework, such as protobuf or CBOR or thrift or whatever animal from the zoo of semi-standards that does versioning well and has good language bindings.  Of course that's a big commitment with a high activation energy.  I see it's been discussed above already.

I'd go for documentation, which will be useful no matter what happens, and involves no complexity vs duplication tradeoff.

Ericson2314:
Thanks for the feedback, @elaforge

> and mostly converted to typeclasses in haskell.

Right, I figure most people would do that. Part of my thinking here is that unless we also structure the C++ code, there is no guarantee that approach works; there is nothing stopping the duplicated code that serializes the same thing from just "drifting apart" over time.

> One thing I wonder, if this is a lipstick on a pig type thing?

I do want to move us on to a better protocol too. But my thinking was first getting rid of the old one and ensuring the many regressions between them are fixed would "free up complexity budget" to do a new one --- supporting 3 protocols in tandems seems like a bit much.

See https://github.com/NixOS/nix/issues/4665 for what I had in mind if yo are curious.

elaforge:
Yeah, fair enough, this does seem like a reasonable strategy.  Open source is not amenable to my "burn it down and make a new one" strategy which I'm used to from company monorepos.  The current protocol will probably live for a long time, if not forever.

I gather you've been working on doxygen or at least normal comments for the c++ source.  I'll look around for a PR before trying to contribute any commentary.  Or maybe just another implementation would be docs enough... though it would be yet another incomplete 3rd party implementation to join the hnix and rust ones.

Ericson2314:
Yeah exactly, I wish I could burn it all down, but I can't. Have to move very cautiously/incrementally instead.

I am done with C++ API docs *en masse* at the moment --- existing comments are converted, so now I just try to "boyscott rule" with each PR rather than doing anything in bulk. So if there is something you'd like to do --- have at it!

fricklerhandwerk:
Discussed in Nix team meeting 2023-05-11:

- @Ericson2314
    - Big refactor of protocol code
    - Allows factored out "typeclass style" serialization logic to depend on version.
    - (Removes bad trade off between dedup and robustness to protocol evolution)
    - Uses that to deduplicate/organize `ValidPathInfo` and `BuildResult` serializations -- more complex types that have evolved over time
    - Would help with deduplicating the serialization logic with Hydra (and migrating to `ssh-ng://`) too --- a project I would still like to get to.
- @thufschmitt: also adds a lot of complexity, so not entirely convinced
- @fricklerhandwerk: the change is quite large, but organised very well by commits. we could do a team review
    - @thufschmitt: sounds good
    - @ericson2314 will schedule time for this

src/libstore/worker-protocol-impl.hh
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include "worker-protocol-impl.hh"
```
fricklerhandwerk:
```sugestion
```

src/libstore/daemon.cc
```diff
@@ -276,8 +276,14 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
     TrustedFlag trusted, RecursiveFlag recursive, unsigned int clientVersion,
     Source & from, BufferedSink & to, unsigned int op)
 {
-    worker_proto::ReadConn rconn { { .from = from } };
-    worker_proto::WriteConn wconn { { .to = to } };
+    worker_proto::ReadConn rconn {
```
fricklerhandwerk, line 279: New commit message:

```
Add protocol versions to {worker,serve}_protocol::*Conn

This will allow us to factor out logic, which is currently scattered inline, into several reusable instances
```

fricklerhandwerk (approved):
We reviewed this together today, with @roberth and @tomberek. This makes sense overall, even if the macro magic is quite involved, but it is reasonable and seems correct. Adding more explanatory comments should help with avoiding confusion and the appearance of magic.

Apart from that, the change *looks* like a pure refactoring for all intents and purposes, but it's scary large. I would predicate merging this on adding at least some significant amount of testing. Otherwise it feels like we're asking for trouble in the next release.

src/libstore/worker-protocol.hh
```diff
@@ -78,39 +27,144 @@ typedef enum {
...
 {
-    WorkerProto<T>::write(store, out, t);
+    return sink << (uint64_t) op;
 }
 
 /**
- * Declare a canonical serializer pair for the worker protocol.
+ * Convenience for debugging.
+ *
+ * @todo Perhaps render known opodes more nicely.
```
roberth, line 159:
```suggestion
 * @todo Perhaps render known opcodes more nicely.
```

src/libstore/length-prefixed-protocol-helper.hh
```diff
@@ -0,0 +1,132 @@
...
+ * Shorthand to refer to `Inner::Serializer` without hitting funky name resolution errors.
+ */
+
+#define LENGTH_PREFIXED_PROTO_HELPER(Inner, T) \
+    struct LengthPrefixedProtoHelper< Inner, T > \
+    { \
+        static T read(const Store & store, typename Inner::ReadConn conn); \
+        static void write(const Store & store, typename Inner::WriteConn conn, const T & str); \
+    private: \
+        template<typename U> using S = typename Inner::template Serialise<U>; \
```
thufschmitt, line 43: Maybe explain before that it's morally `using S = Inner::Serialise`

src/libstore/length-prefixed-protocol-helper.hh
```diff
@@ -0,0 +1,132 @@
...
+LengthPrefixedProtoHelper<Inner, std::vector<T>>::read(
+    const Store & store, typename Inner::ReadConn conn)
+{
+    std::vector<T> resSet;
+    auto size = readNum<size_t>(conn.from);
+    while (size--) {
+        resSet.push_back(S<T>::read(store, conn));
+    }
+    return resSet;
+}
```
thufschmitt, line 68: I that's not too hard, it would be great to have some test (rapidcheck?) for these to make sure that we keep the exact same serialization behavior

src/libstore/worker-protocol-impl.hh
```diff
@@ -9,70 +9,50 @@
...
-        WorkerProto::Serialise<T>::write(store, conn, key);
+#define WORKER_USE_LENGTH_PREFIX_SERIALISER(TEMPLATE, T) \
+    TEMPLATE T WorkerProto::Serialise< T >::read(const Store & store, WorkerProto::ReadConn conn) \
+    { \
+        return LengthPrefixedProtoHelper<WorkerProto, T >::read(store, conn); \
+    } \
+    TEMPLATE void WorkerProto::Serialise< T >::write(const Store & store, WorkerProto::WriteConn conn, const T & t) \
+    { \
+        LengthPrefixedProtoHelper<WorkerProto, T >::write(store, conn, t); \
     }
```
thufschmitt: Could we instead define the relevant instances as inheriting from `LenghtFixedProtoHelper` directly? Something like

```cpp
template <typename T>
struct WorkerProto::Serialise<std::set<T>> : LengthPrefixedProtocolHelper<WorkerProto, T> {};
```

src/libstore/build/local-derivation-goal.cc
```diff
@@ -10,8 +10,8 @@
 #include "archive.hh"
 #include "compression.hh"
 #include "daemon.hh"
-#include "worker-protocol.hh"
-#include "worker-protocol-impl.hh"
+#include "common-protocol.hh"
+#include "common-protocol-impl.hh"
```
thufschmitt, line 14: These can probably be removed

src/libstore/daemon.cc
```diff
@@ -646,16 +646,7 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
...
-        if (GET_PROTOCOL_MINOR(clientVersion) >= 29) {
-            to << res.timesBuilt << res.isNonDeterministic << res.startTime << res.stopTime;
-        }
-        if (GET_PROTOCOL_MINOR(clientVersion) >= 28) {
-            DrvOutputs builtOutputs;
-            for (auto & [output, realisation] : res.builtOutputs)
-                builtOutputs.insert_or_assign(realisation.id, realisation);
-            WorkerProto::write(*store, wconn, builtOutputs);
-        }
+        WorkerProto::write(*store, wconn, res);
```
Ericson2314, line 637: It would be good to factor out each protocol's in a separate commit.

thufschmitt (commented):
Paired review with @roberth and @Ericson2314 .

Great overall! Just a bunch of small-ish comments since this is a quite large change

src/libstore/length-prefixed-protocol-helper.hh
Ericson2314: Added this, and more to the documentation. (For others's, note that it is above because this is in a macro.

src/libstore/worker-protocol-impl.hh
Ericson2314: I tried that, but it didn't seem to be working (not sure why though). Moreover, it means `LengthPrefixedProtocolHelper` can't just be an implemnetation detail in the "impl" header, but must also be exposed in the main "interface" header.

edolstra (approved):

PR #6227, opened 2022-03-10
Impure derivations

edolstra:
This is a revived implementation of 18c512d3b86429d26d9b7b60f9c8bb0d60db1aff.

Impure derivations are derivations that can produce a different result every time they're built. Example:

```
stdenv.mkDerivation {
  name = "impure";
  __impure = true; # marks this derivation as impure
  #outputHashAlgo = "sha256"; # optional, default is sha256
  #outputHashMode = "recursive"; # optional, default is recursive
  buildCommand = "date > $out";
};
```

Some important characteristics:

* This requires the 'impure-derivations' experimental feature.

* Impure derivations are not "cached". Thus, running "nix-build" on the example above multiple times will cause a rebuild every time.

* They are implemented similar to CA derivations, i.e. the output is moved to a content-addressed path in the store. The difference is that we don't register a realisation in the Nix database.

* Only impure or fixed-output derivations are allowed to depend on impure derivations directly. "Pure" (i.e. input-addressed or floating CA derivations) can depend on impure derivations indirectly, if there is a fixed-output derivation in between. Thus the fixed-output derivation forms an "impurity barrier" in the dependency graph.

* When sandboxing is enabled, impure derivations can access the network in the same way as fixed-output derivations. In relaxed sandboxing mode, they can access the local filesystem.

A motivating example for impure derivations is the problem of using `fetchurl` on a dynamically generated tarball whose contents are deterministic, but where the tarball does not have a canonical form. Previously, this required `fetchurl` to do the unpacking in the same derivation. (That's what `fetchzip` does.) But now we can separate these two steps:

```
tarball = stdenv.mkDerivation {
  __impure = true;
  name = "tarball";
  buildInputs = [ curl ];
  buildCommand =
    "curl --fail -Lk https://github.com/NixOS/patchelf/tarball/c1f89c077e44a495c62ed0dcfaeca21510df93ef > $out";
};

unpacked = stdenv.mkDerivation {
  name = "unpacked";
  outputHashAlgo = "sha256";
  outputHashMode = "recursive";
  outputHash = "1jl8n1n36w63wffkm56slcfa7vj9fxkv4ax0fr0mcfah55qj5l8s";
  buildCommand =
    "mkdir $out; tar xvf ${tarball} -C $out";
};
```

Another application could be impure (final) steps in CA pipelines like Hydra jobsets, like a job that uploads a build result somewhere.

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -342,9 +342,9 @@ void DerivationGoal::gaveUpOnSubstitution()
     inputDrvOutputs.clear();
     if (useDerivation)
         for (auto & i : dynamic_cast<Derivation *>(drv.get())->inputDrvs) {
-            /* Ensure that pure derivations don't depend on impure
-               derivations. */
-            if (drv->type().isPure()) {
+            /* Ensure that pure, non-fixed-output derivations don't
+               depend on impure derivations. */
+            if (drv->type().isPure() && !drv->type().isFixed()) {
                 auto inputDrv = worker.evalStore.readDerivation(i.first);
                 if (!inputDrv.type().isPure())
                     throw Error("pure derivation '%s' depends on impure derivation '%s'",
@@ -993,7 +993,8 @@ void DerivationGoal::resolvedFinished()
                 auto newRealisation = realisation;
                 newRealisation.id = DrvOutput { initialOutputs.at(wantedOutput).outputHash, wantedOutput };
                 newRealisation.signatures.clear();
-                newRealisation.dependentRealisations = drvOutputReferences(worker.store, *drv, realisation.outPath);
+                if (!drv->type().isFixed())
+                    newRealisation.dependentRealisations = drvOutputReferences(worker.store, *drv, realisation.outPath);
                 signRealisation(newRealisation);
                 worker.store.registerDrvOutput(newRealisation);
             }
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -204,8 +204,33 @@ void DerivationGoal::haveDerivation()
 {
     trace("have derivation");
 
+    parsedDrv = std::make_unique<ParsedDerivation>(drvPath, *drv);
+
     if (!drv->type().hasKnownOutputPaths())
         settings.requireExperimentalFeature(Xp::CaDerivations);
+
+    if (!drv->type().isPure()) {
+        settings.requireExperimentalFeature(Xp::ImpureDerivations);
+
+        for (auto & [outputName, output] : drv->outputs) {
+            auto randomPath = StorePath::random(outputPathName(drv->name, outputName));
+            assert(!worker.store.isValidPath(randomPath));
+            initialOutputs.insert({
+                outputName,
+                InitialOutput {
+                    .wanted = true,
+                    .outputHash = impureOutputHash,
+                    .known = InitialOutputStatus {
+                        .path = randomPath,
+                        .status = PathStatus::Absent
+                    }
+                }
+            });
+        }
+
+        gaveUpOnSubstitution();
+        return;
+    }
 
     for (auto & i : drv->outputsAndOptPaths(worker.store))
         if (i.second.second)
@@ -229,9 +254,6 @@ void DerivationGoal::haveDerivation()
         done(BuildResult::AlreadyValid, std::move(validOutputs));
         return;
     }
-
-    parsedDrv = std::make_unique<ParsedDerivation>(drvPath, *drv);
-
 
     /* We are first going to try to create the invalid output paths
        through substitutes.  If that doesn't work, we'll build
@@ -265,6 +287,8 @@ void DerivationGoal::haveDerivation()
 void DerivationGoal::outputsSubstitutionTried()
 {
     trace("all outputs substituted (maybe)");
+
+    assert(drv->type().isPure());
 
     if (nrFailed > 0 && nrFailed > nrNoSubstituters + nrIncompleteClosure && !settings.tryFallback) {
         done(BuildResult::TransientFailure, {},
@@ -315,9 +339,21 @@ void DerivationGoal::outputsSubstitutionTried()
 void DerivationGoal::gaveUpOnSubstitution()
 {
     /* The inputs must be built before we can build this goal. */
+    inputDrvOutputs.clear();
     if (useDerivation)
-        for (auto & i : dynamic_cast<Derivation *>(drv.get())->inputDrvs)
+        for (auto & i : dynamic_cast<Derivation *>(drv.get())->inputDrvs) {
+            /* Ensure that pure derivations don't depend on impure
+               derivations. */
+            if (drv->type().isPure()) {
+                auto inputDrv = worker.evalStore.readDerivation(i.first);
+                if (!inputDrv.type().isPure())
+                    throw Error("pure derivation '%s' depends on impure derivation '%s'",
+                        worker.store.printStorePath(drvPath),
+                        worker.store.printStorePath(i.first));
+            }
+
             addWaitee(worker.makeDerivationGoal(i.first, i.second, buildMode == bmRepair ? bmRepair : bmNormal));
+        }
 
     /* Copy the input sources from the eval store to the build
        store. */
@@ -345,6 +381,8 @@ void DerivationGoal::gaveUpOnSubstitution()
 
 void DerivationGoal::repairClosure()
 {
+    assert(drv->type().isPure());
+
     /* If we're repairing, we now know that our own outputs are valid.
        Now check whether the other paths in the outputs closure are
        good.  If not, then start derivation goals for the derivations
@@ -452,22 +490,24 @@ void DerivationGoal::inputsRealised()
                        drvs. */
                     : true);
             },
+            [&](const DerivationType::Impure &) {
+                return true;
+            }
         }, drvType.raw());
 
-        if (resolveDrv)
-        {
+        if (resolveDrv && !fullDrv.inputDrvs.empty()) {
             settings.requireExperimentalFeature(Xp::CaDerivations);
 
             /* We are be able to resolve this derivation based on the
-               now-known results of dependencies. If so, we become a stub goal
-               aliasing that resolved derivation goal */
-            std::optional attempt = fullDrv.tryResolve(worker.store);
+               now-known results of dependencies. If so, we become a
+               stub goal aliasing that resolved derivation goal. */
+            std::optional attempt = fullDrv.tryResolve(worker.store, inputDrvOutputs);
             assert(attempt);
             Derivation drvResolved { *std::move(attempt) };
 
             auto pathResolved = writeDerivation(worker.store, drvResolved);
 
-            auto msg = fmt("Resolved derivation: '%s' -> '%s'",
+            auto msg = fmt("resolved derivation: '%s' -> '%s'",
                 worker.store.printStorePath(drvPath),
                 worker.store.printStorePath(pathResolved));
             act = std::make_unique<Activity>(*logger, lvlInfo, actBuildWaiting, msg,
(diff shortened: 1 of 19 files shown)


src/libstore/derivations.hh
```diff
@@ -44,13 +44,24 @@ struct DerivationOutputCAFloating
...
+/* Impure output which is moved to a content-addressed location (like
+   CAFloating) but isn't registered as a realization.
+ */
+struct DerivationOutputImpure
+{
+    /* information used for expected hash computation */
+    FileIngestionMethod method;
+    HashType hashType;
+};
+
```
Ericson2314, line 56: I don't think we need this? I didn't think we did, and I went through and indeed it seems this always handled the same way as a CA output.

src/libstore/derivations.hh
```diff
@@ -77,6 +88,7 @@ enum struct DerivationType : uint8_t {
     DeferredInputAddressed,
     CAFixed,
     CAFloating,
+    Impure
 };
```
Ericson2314, line 92: I think it might be better to bite the bullet and do:
```c++
struct DerivationType_InputAddressed {
    bool deferred;
};

struct DerivationType_ContentAddressed {
    bool pure;
    bool fixed;
};

typedef std::variant<
    DerivationType_InputAddressed,
    DerivationType_ContentAddressed
> _DerivationTypeRaw

struct DerivationType : _DerivationTypeRaw {
    using Raw = _DerivationTypeRaw;
    using Raw::Raw;
    using InputAddressed = DerivationType_InputAddressed;
    using ContentAddressed = DerivationType_ContentAddressed;
};
```
This fills out the missing "pure, fixed" use-case. My belief is while "pure, fixed" may not be be very useful, implementing "missing cases" is good to ensure the code stays nice, and also test the code better.

This also explains why we don't need `DerivationOutputImpure`: Only the fixed/floating axis affects the sort of output it is.

Ericson2314:
> Pure derivations are not allowed to depend on impure derivations. In the future fixed-output derivations will be allowed to depend on impure derivations, thus forming an "impurity barrier" in the dependency graph.

Actually, even that is overkill. Once a pure derivation is resolved, we don't where any of it's CA imputs came from --- since they are "just" data! --- and that means we don't care whether those imputs came from CA derivations. What this means is that pure unresolved derivations perhaps can only be *impurely resolved*, so we can use our in-memory resolution cache to finish the job once pure resolution gets stuck, at the cost of "tainting" any downstream resolutions.

If we squint and imagine a pure derivation is a function `a -> b`, what the above means is that impurity is a functor, and we are in effect mapping our derivation function to `impure a -> impure b`.

src/libstore/build/derivation-goal.cc
```diff
@@ -919,60 +953,53 @@ void DerivationGoal::buildDone()
...
-            worker.store.registerDrvOutput(newRealisation);
-            outputPaths.insert(realisation->outPath);
-            builtOutputs.emplace(realisation->id, *realisation);
-        } else {
-            // If we don't have a realisation, then it must mean that something
-            // failed when building the resolved drv
-            assert(!buildResult.success());
+            if (drv->type() != DerivationType::Impure) {
+                if (drv->type() != DerivationType::CAFixed)
+                    newRealisation.dependentRealisations = drvOutputReferences(worker.store, *drv, realisation.outPath);
```
edolstra, line 984: @thufschmitt @Ericson2314 This disables registering dependent realisations for fixed-output derivations. I'm not sure about this since I don't really understand dependent realisations very well... What are the consequences of this?

The reason for disabling it is that `drvOutputReferences()` gets realisations from the DB, and we don't have those for the impure inputs of `drv`.

thufschmitt (commented):
That’s great 😎

I managed to break a number of things (which is expected given that this adds a whole new derivation type, but better be aware of them). I don’t think these should be blockers for this PR (afaik, nothing breaks as long as there’s no impure derivation involved), just something to keep in mind for later:

- `nix-build`
    ```console
    $ nix-build impure-derivations.nix -A impure
    warning: unknown experimental feature 'ca-references'
      /tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv
    warning: unknown experimental feature 'ca-references'
    building '/tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv'...
    impure
    nix-build: src/nix-build/nix-build.cc:585: void main_nix_build(int, char**): Assertion `maybeOutputPath' failed.
    [1]    4184961 IOT instruction (core dumped)  nix-build impure-derivations.nix -A impure
    ```
- `nix-shell`
    ```console
    $ nix-shell impure-derivations.nix -A impureOnImpure
    warning: unknown experimental feature 'ca-references'
      /tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv
    warning: unknown experimental feature 'ca-references'
    building '/tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv'...
    impure
    warning: output 'out' of input '/tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv' missing, aborting the resolving
    nix-shell: src/nix-build/nix-build.cc:410: void main_nix_build(int, char**): Assertion `resolvedDrv && "Successfully resolved the derivation"' failed.
    [1]    7110 IOT instruction (core dumped)  nix-shell impure-derivations.nix -A impureOnImpure
    ```
- `nix develop`
    ```console
    $ nix develop -f impure-derivations.nix impureOnImpure
    warning: unknown experimental feature 'ca-references'
    error: pure derivation '/tmp/nix-test/default/store/gw8z2q3qnc6bx1ap3s8v81fbvd5qxlrr-impure-on-impure-env.drv' depends on impure derivation '/tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv'
    ```
- Trying to copy an impure derivation will cause Nix to segfault:
    ```console
    $ nix copy --to /tmp/nix --file impure-derivations.nix impure
    [1/0/1 built] building impure: impurenix: src/libstore/derived-path.cc:106: nix::BuiltPath::toRealisedPaths(nix::Store&) const::<lambda(const Built&)>: Assertion `thisRealisation' failed.
    [1]    4183600 IOT instruction (core dumped)  nix copy --to /tmp/nix --file impure-derivations.nix impure
    ```
- `nix why-depends` is broken too:
    ```console
    $ nix why-depends --file impure-derivations.nix inputAddressedAfterCA impure
    warning: unknown experimental feature 'ca-references'
      /tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv
    error: cannot operate on an output of unbuilt content-addressed derivation 'sha256:ddffc26775f61118da8597d1f2f62c544ec92461b31b5b1075ed8689a7ee8292!out'
    ```
- import-from-derivation (though I’m not sure that’s something that we want to support):
    ```console
    $ nix eval --impure --expr 'import (import ./impure-derivations.nix).impure'
    error: output 'out' of derivation '/tmp/nix-test/default/store/hhrsdj65kvf1n0d5y1yj6qc21dqr0gki-impure.drv' has no store path mapped to it

          … while realising the context of path '/135sba7ahqvizkd4xq2cbpwm2l4dy8dqli0ym1jmrsxjn975pz5w'

          at «string»:1:1:

                1| import (import ./impure-derivations.nix).impure
                | ^
    ```

(but as I said, these are just stuff to keep in mind for later, definitely not a criticism of the PR)

src/libstore/derivations.hh
Ericson2314: @edolstra would you mean if I took a stab at doing this, and then rebasing this PR getting rid of `DerivationOutputImpure` as outlined above?

I would like to do the cleanups first so the diff here is extra legible and easier to review.

tomberek: would "pure, fixed" be something like an enforced reproducibility check? or a cached CA outcome similar to builtins.storePath?

Either way, the "missing cases" viewpoint shows a good way to organize both the code and the underlying theory about how these types of derivations relate to each other.

Ericson2314: It's a fixed output derivation that has no special sandbox loopholes (e.g. it cannot access the network).

Glad you like the "missing cases" viewpoint too!

Ericson2314:
I am fixing the conflicts in this.

Ericson2314:
I must say, while the `BuildResult` returning is convenient, the fact that we are drifting away from the good "Command Query Responsibility Segregation" that we had before makes me nervous.

I think I would rather create a notion of transactions.
```c++
struct LayeredTrustMapStore: Store {
// some trust map functions,
// letting us know whether the results came from this store
// of the underlying store.
}

ref<LayeredTrustMapStore> Store::createTransacation();
```
The idea is that

1. All the commands @thufschmitt mentioned should work again.
2. Thanks to `virtual` we have the option of e.g. persisting impure trust maps in DB but segregated.
3. Interesting overlap with https://gist.github.com/edolstra/afa5a41d4acbc0d6c8cccfede7fd4792
4. Impure and pure flows re trust maps should be more similar: same *how* just different *where* the mapping goes.

Ericson2314:
https://github.com/NixOS/nix/compare/master...Ericson2314:impure-derivations-ng Here is the rebase.

tests/impure-derivations.sh
```diff
@@ -0,0 +1,55 @@
...
+
+path4=$(nix build -L --no-link --json --file ./impure-derivations.nix impureOnImpure | jq -r .[].outputs.out)
+[[ $(< $path4/n) = X3 ]]
+
+# Test that (self-)references work.
+[[ $(< $path4/symlink/bla) = 3 ]]
+[[ $(< $path4/self/n) = X3 ]]
+
+# Input-addressed derivations cannot depend on impure derivations directly.
+nix build -L --no-link --json --file ./impure-derivations.nix inputAddressed 2>&1 | grep 'depends on impure derivation'
```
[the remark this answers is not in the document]
edolstra: Thanks, I've done this by adding `set -o pipefail`.

src/libstore/derivations.cc
```diff
@@ -25,55 +25,98 @@ std::optional<StorePath> DerivationOutput::path(const Store & store, std::string
...
+
+bool DerivationType::isPure() const
+{
+    return std::visit(overloaded {
+        [](const InputAddressed & ia) {
+            return true;
+        },
         [](const ContentAddressed & ca) {
-            return !ca.pure;
+            return true;
```
Ericson2314, line 116: A `ContentAddressed` that is neither fixed-output or sandboxed is impure.

amjoseph-nixpkgs:
Is there a timeline for de-experimentalizing this feature?

Along with [NixOS/nixpkgs#188587](https://github.com/NixOS/nixpkgs/pull/188587) and #7052 it will allow us to eliminate all those "remember not to use `fetchpatch` here" comments all over the place in nixpkgs.

```
# Note: this package is used for bootstrapping fetchurl, and thus
# cannot use fetchpatch! All mutable patches (generated by GitHub or
# cgit) that are needed here should be included directly in Nixpkgs as
# files.
```

Some of these banners are extraneous, others are missing, and all of them are unchecked.  It would be great if we could just remove the restriction instead.

thufschmitt:
> Is there a timeline for de-experimentalizing this feature?

Afaik no, but one blocker (at least) for Nixpkgs is that hydra currently doesn't support it, and it's probably not a trivial change

amjoseph-nixpkgs:
> hydra currently doesn't support it, and it's probably not a trivial change

Hrm, what version of Nix does Hydra run (and is that documented anywhere)?

I assume from your comment that it must be running something older than 2.8, because since Nix 2.8 enabling impure derivations has been a trivial change (`--extra-experimental-features impure-derivations`).

thufschmitt:
Hydra is generally pretty close from Nix master, but it's not that simple unfortunately, because Hydra actually re-implements part of the Nix build loop, which means that features like that requires changing hydra.

PR #6229, opened 2022-03-11
Overhaul derivation hash modulo somewhat

Ericson2314:
These changes were taken from dynamic derivation (#4628). They somewhat
undo the the refactors I first did for floating CA derivations, as the
benefit of hindsight + requirements of dynamic derivations made me
reconsider some things.

They aren't to consequential, but I figured they might be good to land
first, before the more profound changes @thufschmitt has in the works.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -1207,32 +1207,37 @@ static void prim_derivationStrict(EvalState & state, const Pos & pos, Value * * 
         // hash per output.
         auto hashModulo = hashDerivationModulo(*state.store, Derivation(drv), true);
         std::visit(overloaded {
-            [&](Hash & h) {
-                for (auto & i : outputs) {
-                    auto outPath = state.store->makeOutputPath(i, h, drvName);
-                    drv.env[i] = state.store->printStorePath(outPath);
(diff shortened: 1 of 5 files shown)


Ericson2314:
@thufschmitt I did rebase your PR on top of https://github.com/Ericson2314/nix/commit/thufschmitt-hash-modulo (and take a few things I liked from it :)), but I didn't add it to this PR because yours changes behavior.

thufschmitt:
>  the resolution of input drvs no longer "flattens" the outputs of fixed output and and floating output derivations:

Oh thanks, good catch. Actually I think this is plain wrong because it means that only the last output will be taken into account (since it’s a map and they all share the same key).

> I think the moral of the story is we might need tests with hard-coded derivations, so we can catch when we actually change this stuff by mistake!

Yes, we do. There’s already a handful of these (one in the testsuite and a hydra job testing that evaluation of the whole of nixpkgs is unchanged), but afaik none of them uses CA derivations

(although in the case at hand it should probably have been caught with any test involving multiple-outputs derivations since we’re effectively erasing some dependencies in that case)

Ericson2314:
Thanks!

PR #6232, opened 2022-03-11
adr: 0001-condense-information-around-nix-3

kamadorueda:


edolstra:
Hm, what is the purpose of this PR? What is "adr"?

kamadorueda:
> What is "adr"?

ADR is just a term for documenting the why of things: https://adr.github.io/

> what is the purpose of this PR? 

What I'm trying to do with this PR is to document the why of Flakes

I believe the things Flakes solve are important and sound, but at the same time the why is rapidly lost in the discussions about the details and consensus is not reached because of this

This small document tries to summarize it all, so it's easy for the reader to see the problem and how Flakes solves it

Because this document also serves as a summary of issues, rfc, discourse, etc, we also have a document that we can later point people to in order to explain why it is as it is (more or less like a FAQ) and reach consensus by avoiding repetition

And since the why starts from the vision "Reproducible builds and deployments", which is something we all want to pursue, I believe it's much more simple to see that anything that gets us closer to that world is a good thing, and such thing is in this case is even already in the master branch: Flakes, so why not embracing Flakes?

> Hm

But honestly, this is just an idea of mine and I may be wrong

feel free to close; or if useful, I can continue working on this

thufschmitt:
I like the broad idea (though I agree that “adr” is a somewhat mysterious term) :smiley: 

I’m not sure it fits in the Nix repo (yet) though since flakes aren’t a thing yet (well they are, but “experimental”). It looks to me as this would better fit as an RFC (but maybe the end-goal here is different from what would be put in a RFC, feel free to correct me if it is).

On the content, I like how it’s concise and straight-to-the point. Probably needs to be extended because flakes aren’t just about reproducibility ([The intro of the old RFC](https://github.com/tweag/rfcs/blob/flakes/rfcs/0049-flakes.md#motivation) is probably a good starting point for getting all the underlying motivation).

fricklerhandwerk:
- @edolstra: shouldn't this information live in the in a wiki?
  - @fricklerhandwerk: no, then it won't be discoverable and not under version control

-> to discuss

fricklerhandwerk:
Closing this. It inspired https://nix.dev/concepts/flakes and many discussions the maintainers team had over the course of the past months.

PR #6233, opened 2022-03-11
Nix repl flakes

tomberek:
Based on: https://github.com/NixOS/nix/pull/5824

Add ability to `nix repl` to utilize installables syntax.

TODO: more tests

Fixes #5824


src/nix/repl.cc
```diff
@@ -892,10 +905,23 @@ struct CmdRepl : StoreCommand, MixEvalArgs
 
     void run(ref<Store> store) override
     {
+
         evalSettings.pureEval = false;
```
Ericson2314, line 909: It is interesting it seems this line doesn't work?

src/nix/repl.cc
```diff
@@ -865,17 +876,19 @@ std::ostream & NixRepl::printValue(std::ostream & str, Value & v, unsigned int m
...
+    }
 
     CmdRepl()
     {
-        expectArgs({
-            .label = "files",
-            .handler = {&files},
-            .completer = completePath
-        });
     }
```
Ericson2314, line 892: Should we be removing this flag? If we do, we can delete the entire constructor; the default one will be fine.

src/nix/repl.cc
```diff
@@ -59,13 +62,16 @@ struct NixRepl
...
+            std::function<AnnotatedValues()> getValues);
     ~NixRepl();
-    void mainLoop(const std::vector<std::string> & files);
+    void mainLoop();
     StringSet completePrefix(const std::string & prefix);
     bool getLine(std::string & input, const std::string &prompt);
     StorePath getDerivationPath(Value & v);
     bool processLine(std::string line);
+
+    void loadInstallable(Installable & installable);
```
Ericson2314, line 74: Dead code

src/nix/repl.cc
```diff
@@ -609,6 +613,11 @@ bool NixRepl::processLine(std::string line)
     return true;
 }
 
+void NixRepl::loadInstallable(Installable & installable)
+{
+    auto [val, pos] = installable.toValue(*state);
+    addAttrsToScope(*val);
+}
```
Ericson2314, line 621: Dead code.

src/nix/repl.cc
tomberek: thx...

src/nix/repl.cc
tomberek: it takes effect after the installables are parsed.... moving it to the constructor seems to do it

Ericson2314: Great!

Artturin (changes requested):

doc/manual/src/release-notes/rl-next.md
```diff
@@ -1 +1,6 @@
 # Release X.Y (202?-??-??)
+
+* `nix repl` now takes installables on the command line, unifying the usage
+  with other commands that use `--file` and `--expr`. Primary breaking change
+  is for the common usage of `nix repl '<nixpkgs>'` which can be recovered with
+  `nix repl nixpkgs` or `nix repl --expr 'import <nixpkgs>{}'`
```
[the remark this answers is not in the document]
tomberek: correct, it's not quite the same because it's not coming from NIX_PATH or from the expected ~/.nix-defexpr/channels/nixpkgs

Ericson2314:
n.b. `nix repl fooDir barDir bazDir` might also have a different interpretation with this?  That does make it a bigger breaking change.

I think we should consider this:

1. Deprecated features, the opposite of experimental features
2. an "old `nix repl` syntax" deprecated feature
3. Two different `CmdRepl`, `nix v1-repl`, `nix v2-repl` and we vary which one `nix repl` is using the deprecated feature.

`nix v2-repl` would require the "nix command" unstable feature so we are free to refine it.

src/nix/repl.cc
```diff
@@ -671,13 +673,15 @@ void NixRepl::reloadFiles()
...
     for (auto & i : old) {
-        if (!first) notice("");
-        first = false;
         notice("Loading '%1%'...", i);
         loadFile(i);
     }
+
+    for (auto & [i,what] : getValues()) {
+        notice("Loading Installable '%1%'...", what);
+        addAttrsToScope(*i);
```
roberth, line 683: So this effectively merges the attributes of multiple installables?

I don't like that "loadables" are loaded into the global scope anyway. It makes discovering which attributes exist harder than it should.

Perhaps flakes and installables could be put in a binding that's derived from the installable name?

```
# nix repl nixpkgs
Loading Installable 'flake:nixpkgs#'...
Added 1 variable.

nix-repl> nixpkgs.legacyPackages.x86_64-linux.emacs.name
"emacs-27.1"

nix-repl> nixpkgs.<TAB>
<UX delight ensues as all top-level flake attrs are listed>

```

Also interesting

```
# nix repl . .#inputs.nixpkgs
Added 2 variables

nix-repl> foo.<TAB>

nix-repl> nixpkgs.lib.concatStringsSep ":" ["testing" "it"]

```

edolstra: The only problem with this is that flakes don't have an obvious name, so it's not entirely straightforward what the top-level attribute name (e.g. `nixpkgs`) should be.

roberth: Right. I pulled `foo` out of thin air.
Most flake refs have a somewhat straightforward candidate for a name.
Using `dirname $PWD` for `.` has produced mixed success in the past though.

Maybe the attribute name should be provided by the user.

```
nix repl --bind foo . --bind nixpkgs .#inputs.nixpkgs
```

or

```
nix repl foo=. nixpkgs=.#inputs.nixpkgs
```

Adding a bare flake or installable to the top-level scope seems ok-ish as long as there's only one. Or maybe it should default to `flake=` for a flake and a name derived from the installable otherwise?

Ericson2314: We do have the previous behavior of multiple `--files` to worry about.

I wouldn't mind if it just made an array called `args`, hah! Then don't have to worry about names, and `:a` can put whatever one likes at the top level.

tomberek:
- [ ] more docs
- [ ] more tests
- [ ] improve error message for legacy usage `nix repl '<nixpkgs>'`  (or use --file?)

src/nix/repl.cc
```diff
@@ -873,17 +877,18 @@ std::ostream & NixRepl::printValue(std::ostream & str, Value & v, unsigned int m
...
-        expectArgs({
-            .label = "files",
-            .handler = {&files},
-            .completer = completePath
-        });
+    Strings getDefaultFlakeAttrPaths()
+    override {
+        return {""};
+    }
+    virtual bool useDefaultInstallables() {
```
thufschmitt, line 890:
```suggestion
    virtual bool useDefaultInstallables() override
    {
```

src/nix/repl.cc
```diff
@@ -873,17 +877,18 @@ std::ostream & NixRepl::printValue(std::ostream & str, Value & v, unsigned int m
...
-
-    CmdRepl()
-    {
-        expectArgs({
-            .label = "files",
-            .handler = {&files},
-            .completer = completePath
-        });
+    Strings getDefaultFlakeAttrPaths()
+    override {
```
thufschmitt, line 887:
```suggestion
    Strings getDefaultFlakeAttrPaths() override
    {
```

src/nix/repl.cc
```diff
@@ -900,10 +905,21 @@ struct CmdRepl : StoreCommand, MixEvalArgs
 
     void run(ref<Store> store) override
     {
-        evalSettings.pureEval = false;
-        auto repl = std::make_unique<NixRepl>(searchPath, openStore());
+        auto state = getEvalState();
+        auto repl = std::make_unique<NixRepl>(searchPath, openStore(),state
```
thufschmitt, line 909: It’d be nicer to split out the `getValue` function to make things easier to read imho

thufschmitt (changes requested):
Left a couple of small stylistic comments, but nothing really serious.

I’m a bit more concerned about the unexpect{ed,able} backwards-compatibility breakage. Maybe that could be eased by
1. Gating the change behind an xp feature flag (if possible making the non-experimental behavior just a layer on top of the new one, otherwise by just brutally duplicating everything)
2. Adding a flag `--file` to the “legacy” codepath (that would essentially be a no-op but allows a common interface between both versions)
3. Warn about usages of `nix repl foo.nix`, suggest using `--file`
4. After one (or two) releases, switch the default to the new behavior and remove the old codepath

Also (somewhat related to the above), `--file` apparently doesn’t auto-call functions properly anymore, so `nix repl --file '<nixpkgs>'` fails with `error: value is a function while a set was expected`

src/nix/repl.cc
```diff
@@ -92,8 +97,10 @@ std::string removeWhitespace(std::string s)
 }
 
 
-NixRepl::NixRepl(const Strings & searchPath, nix::ref<Store> store)
-    : state(std::make_unique<EvalState>(searchPath, store))
+NixRepl::NixRepl(const Strings & searchPath, nix::ref<Store> store,ref<EvalState> state,
```
edolstra, line 100:
```suggestion
NixRepl::NixRepl(const Strings & searchPath, nix::ref<Store> store, ref<EvalState> state,
```

src/libcmd/command.hh
```diff
@@ -114,12 +114,13 @@ struct InstallablesCommand : virtual Args, SourceExprCommand
...
 
     void prepare() override;
+    Installables load();
 
     virtual bool useDefaultInstallables() { return true; }
 
     std::optional<FlakeRef> getFlakeRefForCompletion() override;
 
-private:
+protected:
```
garbas, line 123: @edolstra ping!

src/nix/repl.cc
```diff
@@ -684,13 +686,15 @@ void NixRepl::reloadFiles()
...
 
-    bool first = true;
     for (auto & i : old) {
-        if (!first) notice("");
-        first = false;
         notice("Loading '%1%'...", i);
         loadFile(i);
     }
+
+    for (auto & [i,what] : getValues()) {
```
edolstra, line 694:
```suggestion
    for (auto & [i, what] : getValues()) {
```

src/libcmd/command.hh
tomberek, line 123: Is this reasonable? or is there a better way?

src/nix/repl.cc
```diff
@@ -684,13 +686,15 @@ void NixRepl::reloadFiles()
...
+        notice("Loading Installable '%1%'...", what);
```
edolstra, line 695:
```suggestion
        notice("Loading installable '%1%'...", what);
```

src/nix/repl.cc
```diff
@@ -913,10 +932,36 @@ struct CmdRepl : StoreCommand, MixEvalArgs
...
+            auto installables = load();
+            NixRepl::AnnotatedValues values;
+            for (auto & installable: installables){
+                auto what = installable->what();
+                if (!settings.isExperimentalFeatureEnabled(Xp::Flakes) && file){
+                    auto [val, pos] = installable->toValue(*state);
+                    auto what = installable->what();
+                    state->forceValue(*val, pos);
+                    auto autoArgs = getAutoArgs(*state);
+                    Value *valPost = state->allocValue();
```
edolstra, line 946:
```suggestion
                    auto valPost = state->allocValue();
```

src/nix/repl.cc
```diff
@@ -913,10 +932,36 @@ struct CmdRepl : StoreCommand, MixEvalArgs
...
+                    state->autoCallFunction(*autoArgs, *val, *valPost);
+                    state->forceValue(*valPost, pos);
+                    values.push_back( {valPost, what });
+                } else {
+                    auto [val, pos] = installable->toValue(*state);
+                    values.push_back( {val,what} );
```
edolstra, line 952:
```suggestion
                    values.push_back( {val, what} );
```

src/libcmd/installables.hh
```diff
@@ -131,6 +131,7 @@ struct Installable
         OperateOn operateOn,
         const std::vector<std::shared_ptr<Installable>> & installables);
```
edolstra, line 132:
```suggestion
        const Installables & installables);
```

and there are probably more places that can use the `Installables` type.

src/libcmd/installables.hh
```diff
@@ -131,6 +131,7 @@ struct Installable
...
 };
+typedef std::vector<std::shared_ptr<Installable>> Installables;
```
edolstra, line 134:
```suggestion

typedef std::vector<std::shared_ptr<Installable>> Installables;
```

edolstra (approved):
Just some style nitpicks, otherwise looks good.

src/nix/repl.cc
```diff
@@ -895,6 +895,22 @@ struct CmdRepl : InstallablesCommand
     CmdRepl(){
         evalSettings.pureEval = false;
     }
+    void prepare()
+    {
+        if (!settings.isExperimentalFeatureEnabled(Xp::Flakes) && !(file)) {
```
thufschmitt: Shouldn’t we have a separate XP feature for that? I’m imagining that this could become the default way before flakes become non-xp

src/nix/repl.cc
```diff
@@ -895,6 +895,22 @@ struct CmdRepl : InstallablesCommand
...
+            warn("future versions of Nix will require using `--file` to load a file");
+            if (this->_installables.size() > 1) {
+                warn("more than one input file is not currently supported");
+            }
```
thufschmitt: Afaik that’s just a limitation of the current CLI framework right?

(Not that we should change it now, but it’s good to keep in mind in case we need to revert this restriction)

src/nix/repl.cc
```diff
@@ -940,9 +937,20 @@ struct CmdRepl : InstallablesCommand
             auto installables = load();
             NixRepl::AnnotatedValues values;
             for (auto & installable: installables){
-                auto [val, pos] = installable->toValue(*state);
                 auto what = installable->what();
-                values.push_back( {val,what} );
+                if (!settings.isExperimentalFeatureEnabled(Xp::Flakes) && file){
```
thufschmitt, line 941: Don’t we want this regardless of the xp feature? It’d still be nice for `nix repl --file '<nixpkgs>'` to keep working

(maybe that could even be the case for the rest of the CLI which behaves a bit weirdly wrt autoargs)

tomberek: I can add it in, but there should be a consistent story across all of the CLI for the autoargs capability.

src/nix/repl.cc
tomberek: Yes, the current CLI takes "--file some-file" to change the meaning of all remaining installables to be attributes, thus we lost the ability to have multiple files. I doubt that is a high priority or common use case.

src/nix/repl.cc
tomberek: perhaps with "nix-command" instead?

src/libcmd/installables.hh
tomberek: I started to make this change, but the compiler was not happy with how I was doing it... we can leave this to another PR? or during hacking session?

src/nix/repl.cc
thufschmitt: Maybe yes. Let’s keep it as a follow-up for later

src/nix/repl.cc
thufschmitt: I’d rather have it use its own `repl-flake` experimental feature. It’s cheap and that will leave us with more freedom to flip the switch soon-ish

thufschmitt:
@tomberek I’ve added some tests in https://github.com/thufschmitt/nix/commit/1b00dcb3907c72c81bcd39071db0f8c2e723dc1a if you want to cherry-pick that

tests/repl.sh
```diff
@@ -71,3 +73,48 @@ testReplResponse '
...
+
+testReplResponse '
+drvPath
+' '".*-simple.drv"' \
+--file $testDir/simple.nix --experimental-features ''
+
+testReplResponse '
+drvPath
+' '".*-simple.drv"' \
+--file $testDir/simple.nix --experimental-features 'repl-flake'
```
thufschmitt, line 90:
```suggestion
--file $testDir/simple.nix --extra-experimental-features 'repl-flake'
```

To make it work with the CA-drvs test

And same below

tests/repl.sh
```diff
@@ -71,3 +73,48 @@ testReplResponse '
 :a { a = "2"; }
 "result: ${a}"
 ' "result: 2"
+
+testReplResponse '
+drvPath
+' '".*-simple.drv"' \
+$testDir/simple.nix --experimental-features ''
```
thufschmitt, line 80:
```suggestion
$testDir/simple.nix
```

(Otherwise the CA tests will fail because they require the `ca-derivations` xp feature)

And likewise below

thufschmitt (commented):
Thanks for keeping through my annoying reviews. Except for the small test failure, that’s looking good now!

bburdette:
@tomberek as far as I can tell its fine.

thufschmitt (approved):
Duh sorry, I though I had it already merged.

I have a couple suggestions for the non-code side, and there's a merge conflict, but looks good otherwise :+1:

thufschmitt:
@tomberek think you can have a look at the failing test?

thufschmitt (approved):
Thanks :)

The hardcoded `ca-derivations` xp feature in the test is a bit annoying, but that'll do for now

PR #6235, opened 2022-03-11
Create some type aliases for string Contexts

Ericson2314:


fricklerhandwerk (approved):

src/libexpr/value.hh
```diff
@@ -64,6 +64,8 @@ class JSONPlaceholder;
 
 typedef int64_t NixInt;
 typedef double NixFloat;
+typedef std::pair<Path, std::string> NixStringContextElem;
```
thufschmitt, line 67: Maybe remove this and use `NixStringContext::value_type` instead? Doesn’t change much, but I find it more explicit

thufschmitt (commented):
Just a small nit, but do what you want with it ;)

edolstra:
Not sure this is really an improvement. If I see `std::pair<std::string, std::string>` I know what it means, but if I see `NixStringContextElem` I have to look up what that type means. I.e. more cognitive overhead.

Ericson2314:
@edolstra The problem with `std::pair<std::string, std::string>` is that there are many other uses which *don't* have to do with the string context. Without having the alias (or better yet, writing a `struct`) one doesn't know which ones are for which purpose.

Concretely this also made #6237 easy (baring the GCC bug) because I could just change the definition in one spot.

Ericson2314:
Also c.f. `SearchPathElem` in this file.

src/libexpr/value.hh
Ericson2314: Or maybe it should be this way to be consistent with `SearchPathElem`. which I had completely forgotten about in the time since I first wrote this code.

fricklerhandwerk:
@edolstra Naming types by what they mean is always better for readability. There are so many arguments for creating (the right!) abstractions, and the idea is so commonplace, that any serious source I can think about citing would already sound corny and stereotyped.

Ericson2314:
#6237 This contains and motivates this, so I will close this in lieu of that.

PR #6236, opened 2022-03-11
Factor out `StoreDirConfig`

Ericson2314:
This takes out a bunch of non-virtual methods which only depend on the store dir.

I didn't move the implementation to a `store-dir-config.cc` because they were already split in two locations (`path.cc` and `store-api.cc`), and also because, I must admit, it will cause me grief later with more merge conflicts.

More progress on #5729.


edolstra:
I'm not really in favor of splitting things just because we can... Is there a specific use case for splitting off `StoreDirConfig`? (E.g. is there a plausible scenario where we have a `StoreDirConfig` that isn't a `Store`?)

Ericson2314:
This was working in Clang, but not GCC, due to that bug @thufschmitt ran into when first making `StoreConfig`. Hopefully I can fix soon.

Ericson2314:
Actually it's a double free instead. #6258 came from my attempts debugging this over the weekend. I rebase this on top, but still no dice.

Ericson2314:
I built with clang and the double free went away. Ugh!

edolstra:
> Yes there is, https://github.com/NixOS/nix/pull/6237 which depends on this.

That's a bit of a circular motivation, because I'm not sure what the reason for #6237 is...

Ericson2314:
Err the motivation of #6237 is simply to use `StorePath` where possible to demonstrate the invariant that the item in question is in fact a store path, as we have done many times before, e.g. your recent #6198.

edolstra:
Ah thanks, yeah that makes sense!

Ericson2314:
Still not sure how to fix this, but in the most recent commit found a bunch of locations that can be made `StoreDirConfig` instead of `Store`.

PR #6237, opened 2022-03-12
Decode string context straight to using StorePaths 

Ericson2314:
I gather decoding happens on demand, so I hope don't think this should
have any perf implications one way or the other.

### DIFF
--- a/src/libexpr/eval-cache.cc
+++ b/src/libexpr/eval-cache.cc
@@ -254,7 +254,7 @@ struct AttrDb
                 return {{rowId, attrs}};
             }
             case AttrType::String: {
-                std::vector<std::pair<Path, std::string>> context;
+                NixStringContext context;
                 if (!queryAttribute.isNull(3))
                     for (auto & s : tokenizeString<std::vector<std::string>>(queryAttribute.getStr(3), ";"))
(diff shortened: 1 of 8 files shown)


thufschmitt (commented):
The last commit looks good to me. Apart from the potential conflicts, is there any reason for this to depend on #6236 ?

Ericson2314:
@thufschmitt It is probably not necessary, but I was a bit nervous about making a `Store` more accessible to the eval cache, because I didn't want to risk some any bit of store state introducing impurities.

Ericson2314:
Squashed out the second commit so there is is no `StoreDirConfig` config in the history. I have kept it elsewhere instead.

Ericson2314:
@edolstra how is this? The core change is the switch to `StorePath`.

The auxillary change is new names for strong contexts, but I think this is good:

1. Matches `SearchPath` and `SearchPathElem`.

2. Allows changes like the `Path` to `StorePath` to be done easily, "catch" all the `std::pair`s that need to be so changed via variable substitution. For RFC 92 I will need to modify `NixStringContext`, so this isn't even that hypothetical a benefit after this one specific renaming.

Ericson2314:
Thank you!

PR #6238, opened 2022-03-12
fix precision loss when converting floating point numbers to strings

### LINKED ISSUE #5733: toJSON truncates floats
**Describe the bug**

toJSON unnecessarily truncates floats to six significant figures.

**Steps To Reproduce**

```
nix-repl> builtins.toJSON 1.000001
"1"

nix-repl> builtins.toJSON 1048576.1
"1.04858e+06"
```

**Expected behavior**

```
nix-repl> builtins.toJSON 1.000001
"1.000001"

nix-repl> builtins.toJSON 1048576.1
"1048576.1"
```

tested 2.4 and 2.5pre20211126_55275fc

**Additional context**

found while investigating https://github.com/NixOS/nixpkgs/issues/148824

pennae:
currently floating point conversions are done with 6 digits of precision, which is adequate for float but can lose precision for double. toString additionally uses the fixed format rather than the general format, so it will truncate numbers to 0 that have an absolute value less than 5e-7.

fixes #5733


edolstra:
This changes evaluation results, which is potentially a big problem...

pennae:
yeah, that is a potential problem :confused: there are a few nixos modules that use floats, but those shouldn't be impacted too much because upgrading nix would most likely rebuild service definitions anyway. in nixpkgs packages themselves floats seem to be pretty rare, running the ofborg outpaths check over all of nixpkgs calls `Value::mkFloat` only 29 times. scientific workloads using nix for reproducible data collection and analysis could benefit from better precision though.

the other option would be to prominently document that nix will not output floats with more than 6 fractional digits, *ever*, unless someone adds a builtin that can do that (which would likely involve versions of toJSON and toXML as well). it's all a bit messy :slightly_frowning_face:

src/libexpr/eval.cc
```diff
@@ -2017,7 +2018,11 @@ BackedStringView EvalState::coerceToString(const Pos & pos, Value & v, PathSet &
         if (v.type() == nBool && v.boolean) return "1";
         if (v.type() == nBool && !v.boolean) return "";
         if (v.type() == nInt) return std::to_string(v.integer);
-        if (v.type() == nFloat) return std::to_string(v.fpoint);
+        if (v.type() == nFloat) {
+            std::stringstream out;
+            out << std::setprecision(15) << v.fpoint;
```
andersk, line 2023: 15 digits does not suffice to distinguish some doubles, such as `1` and `1.0000000000000002`. This should be `std::numeric_limits<double>::max_digits10 == 17`.

src/libutil/json.cc
```diff
@@ -52,8 +52,8 @@ template<> void toJSON<long>(std::ostream & str, const long & n) { str << n; }
 template<> void toJSON<unsigned long>(std::ostream & str, const unsigned long & n) { str << n; }
 template<> void toJSON<long long>(std::ostream & str, const long long & n) { str << n; }
 template<> void toJSON<unsigned long long>(std::ostream & str, const unsigned long long & n) { str << n; }
-template<> void toJSON<float>(std::ostream & str, const float & n) { str << n; }
-template<> void toJSON<double>(std::ostream & str, const double & n) { str << n; }
+template<> void toJSON<float>(std::ostream & str, const float & n) { str << std::setprecision(7) << n; }
```
andersk, line 55: Should be `std::numeric_limits<float>::max_digits10 == 9`.

src/libutil/json.cc
```diff
@@ -52,8 +52,8 @@ template<> void toJSON<long>(std::ostream & str, const long & n) { str << n; }
...
+template<> void toJSON<double>(std::ostream & str, const double & n) { str << std::setprecision(15) << n; }
```
andersk, line 56: Should be `std::numeric_limits<double>::max_digits10 == 17`.

src/libexpr/eval.cc
pennae: that is true. in the interest of keeping at least somewhat close to the original implementation we chose to print floats with fewer digits than the binary format supports to avoid least significant digit artifacts like `1.1` printing as `1.1000000000000001`. it seemed like a good compromise between accuracy and potential impact of system-dependent rounding artifacts on the output.

andersk: If we’re going to risk changing evaluation results for this, maybe we should pull in the [Ryū](https://github.com/ulfjack/ryu) library and do it right, so we don’t have to change them again later?

In the future when we can rely on GCC ≥ 11, we’ll be able to use `std::to_chars`, which uses a Ryū implementation built into libstdc++.

pennae: that would indeed be much better. going to mark this as draft for now until we can give this the attention such a change deserves.

sersorrel:
this would be really really nice to have

I was _very_ confused earlier by nix telling me that 20.350000 and 20.350000 were not equal; the only way I worked out to see the actual value was to multiply by 1000000000 before printing, at which point I got to see that the first value is actually something like 20.349999999999996

infinisil:
I think for now we could at least introduce a `builtins.floatToString` that uses [Ryū](https://github.com/ulfjack/ryu) as suggested by @andersk, so that we don't need to change the evaluation result but at least have an alternate way to print floats correctly. Eventually we could deprecate `toString` for floats to remove the buggy behavior.

infinisil:
@pennae Fyi there's an RFC draft proposing something along those lines: https://github.com/NixOS/rfcs/pull/137

elaforge:
Just as a data point to show that the low precision does in fact affect people, about a year ago in our company's internal codebase I discovered people passing test arguments in strings like "0.1234*0.1234" which were then eval'ed in python.  They had (painfully) discovered that nix loses precision in floats and implemented a workaround.  I spent some time looking into it before I discovered it was documented and intentional and not a bug.

andersk:
A third alternative to pulling in Ryū or waiting for GCC ≥ 11 `std::to_chars` occurred to me: we could just binary search for the minimum precision needed to correctly round-trip the given float. This would be a bit slower, but I’m guessing float usage in Nix is rare enough that nobody will notice.

PR #6242, opened 2022-03-12
nix-env: always print output names in JSON and XML

ncfavier:
The current `--out-path` flag has two disadvantages when one is only concerned with querying the names of outputs:
- it requires evaluating every output's `outPath`, which takes significantly more resources and runs into more failures
- it destroys the information of the order of outputs so we can't tell which one is the main output

Added an `outputNames` field to the JSON output containing the ordered list of output names, and made the XML always contain `<output>` tags, only with a `path` attribute if `--out-path` is given.

Did not touch the column output.

cc @fzakaria 

The motivation is still https://github.com/NixOS/nixos-search/pull/419

### DIFF
--- a/src/libexpr/get-drvs.cc
+++ b/src/libexpr/get-drvs.cc
@@ -102,7 +102,7 @@ StorePath DrvInfo::queryOutPath() const
 }
 
 
-DrvInfo::Outputs DrvInfo::queryOutputs(bool onlyOutputsToInstall)
+DrvInfo::Outputs DrvInfo::queryOutputs(bool withPaths, bool onlyOutputsToInstall)
 {
     if (outputs.empty()) {
         /* Get the ‘outputs’ list. */
@@ -112,20 +112,24 @@ DrvInfo::Outputs DrvInfo::queryOutputs(bool onlyOutputsToInstall)
 
             /* For each output... */
             for (auto elem : i->value->listItems()) {
-                /* Evaluate the corresponding set. */
-                std::string name(state->forceStringNoCtx(*elem, *i->pos));
-                Bindings::iterator out = attrs->find(state->symbols.create(name));
-                if (out == attrs->end()) continue; // FIXME: throw error?
-                state->forceAttrs(*out->value, *i->pos);
+                std::string output(state->forceStringNoCtx(*elem, *i->pos));
 
-                /* And evaluate its ‘outPath’ attribute. */
-                Bindings::iterator outPath = out->value->attrs->find(state->sOutPath);
-                if (outPath == out->value->attrs->end()) continue; // FIXME: throw error?
-                PathSet context;
-                outputs.emplace(name, state->coerceToStorePath(*outPath->pos, *outPath->value, context));
+                if (withPaths) {
+                    /* Evaluate the corresponding set. */
+                    Bindings::iterator out = attrs->find(state->symbols.create(output));
+                    if (out == attrs->end()) continue; // FIXME: throw error?
+                    state->forceAttrs(*out->value, *i->pos);
+
+                    /* And evaluate its ‘outPath’ attribute. */
+                    Bindings::iterator outPath = out->value->attrs->find(state->sOutPath);
+                    if (outPath == out->value->attrs->end()) continue; // FIXME: throw error?
+                    PathSet context;
+                    outputs.emplace(output, state->coerceToStorePath(*outPath->pos, *outPath->value, context));
+                } else
+                    outputs.emplace(output, std::nullopt);
             }
         } else
-            outputs.emplace("out", queryOutPath());
+            outputs.emplace("out", withPaths ? std::optional{queryOutPath()} : std::nullopt);
     }
     if (!onlyOutputsToInstall || !attrs)
         return outputs;
--- a/src/nix-env/nix-env.cc
+++ b/src/nix-env/nix-env.cc
@@ -918,12 +918,17 @@ static void queryJSON(Globals & globals, std::vector<DrvInfo> & elems, bool prin
             pkgObj.attr("pname", drvName.name);
             pkgObj.attr("version", drvName.version);
             pkgObj.attr("system", i.querySystem());
+            pkgObj.attr("outputName", i.queryOutputName());
 
-            if (printOutPath) {
-                DrvInfo::Outputs outputs = i.queryOutputs();
+            {
+                DrvInfo::Outputs outputs = i.queryOutputs(printOutPath);
                 JSONObject outputObj = pkgObj.object("outputs");
-                for (auto & j : outputs)
-                    outputObj.attr(j.first, globals.state->store->printStorePath(j.second));
+                for (auto & j : outputs) {
+                    if (j.second)
+                        outputObj.attr(j.first, globals.state->store->printStorePath(*j.second));
+                    else
+                        outputObj.attr(j.first, nullptr);
+                }
             }
 
             if (printMeta) {
@@ -1154,13 +1159,16 @@ static void opQuery(Globals & globals, Strings opFlags, Strings opArgs)
                     columns.push_back(drvPath ? store.printStorePath(*drvPath) : "-");
             }
 
+            if (xmlOutput)
+                attrs["outputName"] = i.queryOutputName();
+
             if (printOutPath && !xmlOutput) {
                 DrvInfo::Outputs outputs = i.queryOutputs();
                 std::string s;
                 for (auto & j : outputs) {
                     if (!s.empty()) s += ';';
                     if (j.first != "out") { s += j.first; s += "="; }
-                    s += store.printStorePath(j.second);
+                    s += store.printStorePath(*j.second);
                 }
                 columns.push_back(s);
             }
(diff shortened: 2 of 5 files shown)


fzakaria:
Can you give an example of the JSON output?
If.i understand the change there are now two vectors for outputs and they are order dependent. 

Why not an object (key/value) instead?

ncfavier:
> Can you give an example of the JSON output?

Without `--out-path`:

```json
{
  "nix": {
    "name": "nix-2.7.0",
    "pname": "nix",
    "version": "2.7.0",
    "system": "x86_64-linux",
    "outputNames": [
      "out",
      "dev",
      "man",
      "doc",
      "debug"
    ]
  }
}
```

With `--out-path`:

```json
{
  "nix": {
    "name": "nix-2.7.0",
    "pname": "nix",
    "version": "2.7.0",
    "system": "x86_64-linux",
    "outputNames": [
      "out",
      "dev",
      "man",
      "doc",
      "debug"
    ],
    "outputs": {
      "debug": "/nix/store/yhqq6i9kpkpjrssgh4j7brm5jvwchd31-nix-2.7.0-debug",
      "dev": "/nix/store/rs3a3hlhdf084fkv3bp8rhr1ashv8h7f-nix-2.7.0-dev",
      "doc": "/nix/store/f0l7fvbzr914vklk23w0li1b6sy2xlhk-nix-2.7.0-doc",
      "man": "/nix/store/hqng1gmsc4s3j9954yr8ax16y790ra6k-nix-2.7.0-man",
      "out": "/nix/store/7c4cp4pj389j8rmrc40x5sq6csf0m8ks-nix-2.7.0"
    }
  }
}
```

> Why not an object (key/value) instead?

For the two reasons I mentioned: an object destroys the order of the keys (well, you could still print them in order but most deserialisers will read objects into an unordered map), and computing the values (paths) is expensive.

src/libexpr/get-drvs.cc
```diff
@@ -102,6 +102,24 @@ StorePath DrvInfo::queryOutPath() const
...
+
+    Bindings::iterator i;
+    if (attrs && (i = attrs->find(state->sOutputs)) != attrs->end()) {
+        state->forceList(*i->value, *i->pos);
+        for (auto elem : i->value->listItems()) {
+            std::string name(state->forceStringNoCtx(*elem, *i->pos));
+            outputNames.push_back(name);
+        }
+    } else
+        outputNames.push_back("out");
```
fzakaria, line 117: curly brace here my preference, maybe with a comment explaining why

fzakaria (commented):
the main output is encoded by order?
otherwise looks sensible the change with some nits

ncfavier:
Yes, this is documented:

https://github.com/NixOS/nix/blob/d5322698a2abbc6d141e1d244e17b0d226a2f18b/doc/manual/src/expressions/derivations.md?plain=1#L82

src/libexpr/get-drvs.cc
ncfavier: This is also documented

https://github.com/NixOS/nix/blob/d5322698a2abbc6d141e1d244e17b0d226a2f18b/doc/manual/src/expressions/derivations.md?plain=1#L50-L51

and it's the same logic as in `queryOutputs` so I'd rather not add a comment.

ncfavier: There's also no braces in `queryOutputs` so leaving it alone for symmetry.

edolstra:
I would go with 
```
"outputs": {
      "debug": null, 
      ...
}
```
instead of adding a superfluous `outputNames` attribute. The ordering of outputs doesn't really have much significance outside of builders.

thufschmitt:
I agree with @edolstra , a dict mapping to `null` if `--out-path` isn’t given is much nicer. With maybe another field giving the name of the default output if that’s needed

ncfavier:
Knowing the main output is important: if we're going to print outputs, the user should know that installing `glib` will install `glib.bin` and not `glib.out`.

I think having the order is also nice (although not essential) because outputs are often ordered in the source according to some sort of logic.

A third alternative is to print an `outputs` *list* without `--out-path` and an `outputs` *object* with `--out-path`. No redundancy, but we can't get the output order and the paths simultaneously.

Let me know what you prefer between this and the `{"outputs": {"out": null, ...}, "mainOutput": "out"}` solution.

Ericson2314:
I think only the question of what output is first matters? I would just have field `defaultOutput: "out"` and then do the object.

ncfavier:
`defaultOutput` it is then!

<details><summary>Example output</summary>

```
$ nix-env -qa -A glib --json
{
  "glib": {
    "name": "glib-2.70.3",
    "pname": "glib",
    "version": "2.70.3",
    "system": "x86_64-linux",
    "defaultOutput": "bin",
    "outputs": {
      "bin": null,
      "debug": null,
      "dev": null,
      "devdoc": null,
      "out": null
    }
  }
}
$ nix-env -qa -A glib --json --out-path
{
  "glib": {
    "name": "glib-2.70.3",
    "pname": "glib",
    "version": "2.70.3",
    "system": "x86_64-linux",
    "defaultOutput": "bin",
    "outputs": {
      "bin": "/nix/store/02qav4nwgc49hz6q1wrqlydfy0hf9zdi-glib-2.70.3-bin",
      "debug": "/nix/store/gfzp9jma5saybyzyd671ni7jjm0f29bc-glib-2.70.3-debug",
      "dev": "/nix/store/ibw8fhlmwj3b0xk5ar61qhs9pcwpw5vh-glib-2.70.3-dev",
      "devdoc": "/nix/store/bmapqjdmwf2p30w1k47nnkjsxbpaw9ms-glib-2.70.3-devdoc",
      "out": "/nix/store/pa5zrzvsblhlw3dkka4q4hkxh4yrbdab-glib-2.70.3"
    }
  }
}
$ nix-env -qa -A glib --xml
<?xml version='1.0' encoding='utf-8'?>
<items>
  <item attrPath="glib" defaultOutput="bin" name="glib-2.70.3" pname="glib" system="x86_64-linux" version="2.70.3">
    <output name="bin" />
    <output name="debug" />
    <output name="dev" />
    <output name="devdoc" />
    <output name="out" />
  </item>
</items>
$ nix-env -qa -A glib --xml --out-path
<?xml version='1.0' encoding='utf-8'?>
<items>
  <item attrPath="glib" defaultOutput="bin" name="glib-2.70.3" pname="glib" system="x86_64-linux" version="2.70.3">
    <output name="bin" path="/nix/store/02qav4nwgc49hz6q1wrqlydfy0hf9zdi-glib-2.70.3-bin" />
    <output name="debug" path="/nix/store/gfzp9jma5saybyzyd671ni7jjm0f29bc-glib-2.70.3-debug" />
    <output name="dev" path="/nix/store/ibw8fhlmwj3b0xk5ar61qhs9pcwpw5vh-glib-2.70.3-dev" />
    <output name="devdoc" path="/nix/store/bmapqjdmwf2p30w1k47nnkjsxbpaw9ms-glib-2.70.3-devdoc" />
    <output name="out" path="/nix/store/pa5zrzvsblhlw3dkka4q4hkxh4yrbdab-glib-2.70.3" />
  </item>
</items>
```
</details>

I fixed the JSON writer to print a final newline.

Also note that the caching of `outputs` in `queryOutputs` now relies on the assumption that we don't call `queryOutputs(false)` and then `queryOutputs(true)` on the same derivation.

src/nix-env/nix-env.cc
```diff
@@ -906,53 +906,57 @@ static VersionDiff compareVersionAgainstSet(
 
 static void queryJSON(Globals & globals, std::vector<DrvInfo> & elems, bool printOutPath, bool printMeta)
 {
```
thufschmitt, line 908: Any reason for this extra block? It’s making everything indented further but doesn’t seem to bring much

thufschmitt (changes requested):
Thanks!

A couple of small comments, but looks good overall.

I’ve also pushed a small test for the feature

src/libexpr/get-drvs.cc
```diff
@@ -102,7 +102,21 @@ StorePath DrvInfo::queryOutPath() const
 }
 
 
-DrvInfo::Outputs DrvInfo::queryOutputs(bool onlyOutputsToInstall)
+std::string DrvInfo::queryDefaultOutput()
```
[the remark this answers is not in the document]
ncfavier: I guess the semantics are slightly different? If you implement `defaultOutput` as `drv.outputName`, you're relying on `drv` to already be the default output (and not something like `drv.foo`). I don't know if that's a reasonable assumption in this context, so it seemed clearer to make a separate function.

src/nix-env/nix-env.cc
ncfavier: We need the destructor to run before the final newline is printed, because it prints the last `}`

ncfavier: I guess we could just put the newline after the `queryJSON` call

src/libexpr/get-drvs.cc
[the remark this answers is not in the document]
thufschmitt: That sounds rather sensible to me. Evaluating `glib.out` is morally equivalent to `glib` in which the default output is `out` (and the order of the outputs shouldn’t really matter besides initially setting `outputName`). So having it show `"defaultOutput": "out"` seems to be the right choice.

(now this is all very confusing in the context of `nix-env` because it will happily ignore all that in favor of `meta.outputsToInstall`. But that’s another matter)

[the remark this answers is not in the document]
ncfavier: OK; I suppose we can rename the field to `outputName` then too to avoid confusion.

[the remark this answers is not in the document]
thufschmitt: Yup’ indeed, that’s even better

thufschmitt (approved):
Looks all good now, thanks :+1:

PR #6244, opened 2022-03-12
nix-env: Add a suggestion for when there's a name collision in channels

Artturin:
help new users find a solution to their problem

./result/bin/nix-env -qa hello
warning: name collision in input Nix expressions, skipping '/home/artturin/.nix-defexpr/channels_root/master'
suggestion: remove 'master' from either the root channels or the user channels
hello-2.12
hello-2.12

Closes https://github.com/NixOS/nixpkgs/issues/136136


SuperSandro2000:
What if the first channel is named `a_nixos` and the second `nixos`. Wouldn't that give a false hint?

Artturin:
> What if the first channel is named `a_nixos` and the second `nixos`. Wouldn't that give a false hint?

then there won't be a collision and this code wont be reached

PR #6252, opened 2022-03-13
Enforce a global "offline" setting

ehmry:
Do not attempt to fetch anything when the "--offline" flag is
passed on the command line or when "offline" is set to true in
a configuration file.


src/nix/main.cc
```diff
@@ -356,12 +356,12 @@ void mainWrapped(int argc, char * * argv)
         && args.command->first != "upgrade-nix")
         settings.requireExperimentalFeature(Xp::NixCommand);
 
-    if (args.useNet && !haveInternet()) {
+    if (!settings.offline && !haveInternet()) {
         warn("you don't have Internet access; disabling some network-dependent features");
-        args.useNet = false;
+        settings.offline = false;
```
dtzWill, line 361: `= true;`, should be opposite of `useNet` and `false` doesn't make sense here anyway since it's only reached if it's already `false`.

dtzWill (changes requested):
Small fix for no internet case where `--offline` isn't set, flipped bool.

src/nix/main.cc
ehmry: Good catch!

src/libfetchers/fetchers.cc
```diff
@@ -131,6 +131,12 @@ std::pair<Tree, Input> Input::fetch(ref<Store> store) const
     }
 
     auto [storePath, input] = [&]() -> std::pair<StorePath, Input> {
+        if (settings.offline) {
+            auto url = this->toURL();
+            if (url.scheme != "file" && url.scheme != "git+file")
```
thufschmitt, line 136: There are more input types that can work offline (at least `path` and `hg+file` come to my mind).
It’d probably be better to defer this check to the fetcher − maybe have a `Fetcher::isLocal(std::string_view url)` method that we could call here, or just lift this check inside `Fetcher::fetch`

src/libfetchers/tarball.cc
```diff
@@ -35,9 +35,12 @@ DownloadFileResult downloadFile(
         };
     };
 
-    if (cached && !cached->expired)
+    if (cached && (!cached->expired || settings.offline))
         return useCached();
```
thufschmitt, line 39: Maybe log something (I’d say `warn`, but that might be too noisy) if we keep the expired cache because we’re offline

src/libstore/globals.hh
```diff
@@ -902,6 +902,12 @@ public:
           resolves to a different location from that of the build machine. You
           can enable this setting if you are sure you're not going to do that.
         )"};
+
+    Setting<bool> offline{
+        this, false, "offline",
+        R"(
+          If set to `true` then Nix will avoid using any network resources.
```
thufschmitt, line 909:
```suggestion
          If set to `true` then Nix will avoid using any network resources. In particular this will
          - Disable the substituters
          - Consider any previously downloaded file as up-to-date and refuse to download any new one
```

(well, maybe not the best wording, but I think explaining a bit more in details what’s going on is worth it)

thufschmitt (changes requested):
A few comments but looks good overall.

There’s also a few places where the cache still can expire in offline mode, I guess that’s what you meant to add as a follow-up in https://github.com/NixOS/nix/pull/6252#issuecomment-1066850849

src/libfetchers/fetchers.cc
ehmry: I tried putting the check in different places but decided doing it this dumb way was a reasonable baseline before modifiying multiple classes. I can move this check into `Fetcher`.

src/libfetchers/tarball.cc
ehmry: I think there are different intuitions on how Nix caching should work, but a message here is worthwhile for debugging.

src/libfetchers/fetchers.cc
Ericson2314: Agreed with @thufschmitt

src/libstore/globals.hh
```diff
@@ -902,6 +902,14 @@ public:
...
           can enable this setting if you are sure you're not going to do that.
         )"};
+
+    Setting<bool> offline{
+        this, false, "offline",
+        R"(
+          If set to `true` then Nix will avoid using any network resources. In particular this will
+          - Disable the substituters
+          - Consider any previously downloaded file as up-to-date and refuse to download any new one
+        )"};
```
Ericson2314, line 912: Can you move this to `FetchSettings`?

Ericson2314:
Draft because changes requested.

ehmry:
I­ should point out that I don't see any way to guarantee that an offline flag is being enforced everywhere. I'm probably missing some cases and it would be an iterative process to fully implement this.

PR #6258, opened 2022-03-14
Remove bug-avoiding `StoreConfig *` casts for settings

Ericson2314:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80431 has been fixed, the default GCC in Nixpkgs is now new enough, and the fix is trivial to backport for older GCCs too (I did it in https://github.com/NixOS/nixpkgs/pull/210816).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -68,6 +68,9 @@ case "$host_os" in
 esac
 
 
+ENSURE_NO_GCC_BUG_80431
+
+
 # Check for pubsetbuf.
 AC_MSG_CHECKING([for pubsetbuf])
 AC_LANG_PUSH(C++)
--- a/m4/gcc_bug_80431.m4
+++ b/m4/gcc_bug_80431.m4
@@ -1,0 +1,64 @@
+# Ensure that this bug is not present in the C++ toolchain we are using.
+#
+# URL for bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80431
+#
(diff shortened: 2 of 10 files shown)


edolstra:
Not opposed in theory but it's a huge diff for little gain... I've never really had a problem with this being a pointer so I'm not sure it really makes `Config` easier to use.

Ericson2314:
@edolstra Sure. So I was flailing around with the GCC bug and trying various things, and made this. The pointer vs ref thing is not very import, agreed. The function, however, is useful if due to that bug need to change what base class we are casting "from".

At this point, I am considering instead backporting patch to earlier GCCs (it is very small!), documenting the restriction in the read-me, and just deleting the workarounds. It is a frustrating bug.

How does that sound to you?

edolstra:
Hm, requiring GCC 12 or a patched GCC might be pretty inconvenient to downstream packages, so we probably shouldn't do that (yet).

Ericson2314:
Well, GCC 12, 5 line patched GCC, or any version of Clang. I do feel like requiring bug-free implementations is "morally" different from requiring the latest features, but yes they are harder to distinguish as  practical matter.

Ericson2314:
This PR was completely redone: rather than working around in a different way, we simply don't need to work around at all any more!

The only churn that is left is getting rid of hacks :).

roberth (approved):

PR #6264, opened 2022-03-15
Fix the tests on 32bits machines

thufschmitt:
year 2222 is too much for a 32 bit timestamp. So replace it by something smaller

### DIFF
--- a/tests/fetchPath.sh
+++ b/tests/fetchPath.sh
@@ -1,6 +1,6 @@
 source common.sh
 
 touch foo -t 202211111111
-# We only check whether 2222-11-1* **:**:** is the last modified date since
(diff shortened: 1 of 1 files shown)


tests/fetchPath.sh
```diff
@@ -1,6 +1,6 @@
 source common.sh
 
-touch foo -t 222211111111
+touch foo -t 202211111111
 # We only check whether 2222-11-1* **:**:** is the last modified date since
```
pennae, line 4:
```suggestion
# We only check whether 2022-11-1* **:**:** is the last modified date since
```

thufschmitt: Oops, thanks :)

Issue #6207, opened 2022-03-06
nix GC taking several hours typically


PAI5REECHO:
**Describe the bug**

Garbage collection typically takes several hours despite having only a few hundred packages installed from my `configuration.nix`. `neofetch` reports `Packages: 4527 (nix-system), 2 (nix-user), 2 (nix-default)`

**Steps To Reproduce**
```
# time nix store gc --dry-run
(will update once complete)
```

![flamegraph](https://user-images.githubusercontent.com/96688573/156940895-07131d08-8ad6-45f4-a6b8-8486cbff1c5c.svg)

**Expected behavior**
Quicker garbage collection time and more feedback regarding progress of GC operation.

**`nix-env --version` output**
`nix-env (Nix) 2.5.0pre20211206_d1aaa7e`

edolstra:
Looks like you have CA derivations enabled? It's spending a lot of time in `queryPartialDerivationOutputMap()` reading derivations. Maybe that could be memoized, or we could store the static output hashes in the database so we don't need to read the derivations at all. @thufschmitt

thufschmitt:
Indeed. I guess that having `keep-outputs` + ca-derivations can require the GC to read every derivation in the store, which is very costly.

> Maybe that could be memoized, or we could store the static output hashes in the database so we don't need to read the derivations at all

As far as I know, that’s already partially memoized (could probably be improved, but maybe not enough), but caching the hashes in the DB might be the best long-term solution (I’m generally a bit worried about putting too much stuff in the db as it’s the hardest part to change, but since that’s just a cache I don’t think it would be a big issue)

thufschmitt:
Although "several hours" is quite insanely high, even considering the need to read all the derivation files. Is your store particularly big (regardless of how many packages are actually “installed”)?

SuperSandro2000:
or is your store installed on a really slow hard drive or mounted over nfs or similar?

PAI5REECHO:
> Looks like you have CA derivations enabled?

> having `keep-outputs` + ca-derivations

Correct.
```
keep-outputs = true
keep-derivations = true
experimental-features = nix-command flakes ca-derivations recursive-nix
```
> Is your store particularly big (regardless of how many packages are actually “installed”)?

I'm not sure what metrics to check here. It's around ~100 GB I think

> or is your store installed on a really slow hard drive or mounted over nfs or similar?

No, it's located on a BTRFS mount point on a local SSD

PAI5REECHO:
[...]

PR #6266, opened 2022-03-16
src/libstore/derivations.cc: avoid istringstream on .drv parse path

trofi:
`nix store gc --dry-run` is a mostly CPU-bound workload that spends
visible time in istream::get and in memcmp (seen in `perf top`):

    8,20%  libc.so.6             [.] __memcmp_avx2_movbe
    5,74%  libstdc++.so.6.0.29   [.] std::istream::get
    3,02%  libc.so.6             [.] __libc_malloc
    3,01%  libsqlite3.so.0.8.6   [.] sqlite3VdbeExec
    2,99%  libc.so.6             [.] _int_malloc
    2,78%  libstdc++.so.6.0.29   [.] std::istream::sentry::sentry
    2,72%  libsqlite3.so.0.8.6   [.] pcache1Fetch
    ...

Looking closer at unexpected CPU usage most of calls to `expect()` in
`.drv` file parseer are 1-2 charasters long. The code always uses
`std::string` comparison for them where overhead is large:
`std::string` is constructed on heap and `memcmp()` is called. Similar
goes for `istream::get()` where we work one char at a time and call to
`istream::get()` takes more cycles than anything else hapens to a `char`.

The change turns most string comparisons to char comparisons effectively
eliminating most of memory allocation overheads and only leaving the
istream one.

Benchmark: my system has about 496MB, 150K `.drv` files,
keep-derivations = true`.

Performance before the change:

    $ time nix store gc --dry-run
    real    0m43,994s # +/- 1s

Performance after the change:

    $ time nix store gc --dry-run
    real    0m41,688s # +/- 1s

`perf top` output for comparison:

    8,20%  libc.so.6             [.] __memcmp_avx2_movbe
    4,72%  libc.so.6             [.] __memchr_avx2
    3,56%  libc.so.6             [.] _int_malloc
    3,56%  [kernel]              [k] copy_user_enhanced_fast_string
    3,23%  libc.so.6             [.] __libc_malloc
    3,08%  libsqlite3.so.0.8.6   [.] sqlite3VdbeExec
    2,77%  libsqlite3.so.0.8.6   [.] pcache1Fetch
    ...

That is a ~4% spedup. The overhead now shifted to sqlite queries
and to memory allocation/free somewhere else.

### DIFF
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -145,31 +145,66 @@ StorePath writeDerivation(Store & store,
 }
 
 
-/* Read string `s' from stream `str'. */
-static void expect(std::istream & str, std::string_view s)
-{
-    char s2[s.size()];
-    str.read(s2, s.size());
-    if (std::string(s2, s.size()) != s)
-        throw FormatError("expected string '%1%'", s);
-}
+struct DerivationParserContext {
+    /* input stream being parsed. */
+    std::string_view input;
 
+    DerivationParserContext(std::string_view i)
+        : input(i)
+    { }
 
-/* Read a C-style string from stream `str'. */
-static std::string parseString(std::istream & str)
+    char peek(void) {
+        if (input.empty())
+            throw FormatError("unexpected EOF");
+        return input.front();
+    }
+
+    char get() {
+        if (input.empty())
+            throw FormatError("unexpected EOF");
+        char r = input.front();
+        input.remove_prefix(1);
+        return r;
+    }
+
+    void expect(std::string_view s)
+    {
+        size_t sz = s.size();
+        if (input.size() < sz || s != input.substr(0, sz))
+            throw FormatError("expected string '%1%'", s);
+        input.remove_prefix(sz);
+    }
+
+    void expect(char c)
+    {
+        if (get() != c)
+            throw FormatError("expected char '%1%'", c);
+    }
+};
+
+static std::string parseString(DerivationParserContext & str)
 {
     std::string res;
-    expect(str, "\"");
-    int c;
-    while ((c = str.get()) != '"')
-        if (c == '\\') {
-            c = str.get();
-            if (c == 'n') res += '\n';
-            else if (c == 'r') res += '\r';
-            else if (c == 't') res += '\t';
-            else res += c;
-        }
+    str.expect('"');
+
+    /* Scan up to first '"' (closing string) or escape. Then append to a string. */
+    for (;;) {
+        std::string_view prefix = str.input.substr(0, str.input.find_first_of("\"\\"));
+        res.append(prefix);
+        str.input.remove_prefix(prefix.size());
+
+        int c = str.get();
+        if (c == '"')
+            break;
+
+        /* Handle '\\'. */
+        c = str.get();
+        if (c == 'n') res += '\n';
+        else if (c == 'r') res += '\r';
+        else if (c == 't') res += '\t';
         else res += c;
+    }
+
     return res;
 }
 
@@ -178,7 +213,7 @@ static void validatePath(std::string_view s) {
         throw FormatError("bad path '%1%' in derivation", s);
 }
 
-static Path parsePath(std::istream & str)
+static Path parsePath(DerivationParserContext & str)
 {
     auto s = parseString(str);
     validatePath(s);
@@ -186,7 +221,7 @@ static Path parsePath(std::istream & str)
 }
 
 
-static bool endOfList(std::istream & str)
+static bool endOfList(DerivationParserContext & str)
 {
     if (str.peek() == ',') {
         str.get();
@@ -200,7 +235,7 @@ static bool endOfList(std::istream & str)
 }
 
 
-static StringSet parseStrings(std::istream & str, bool arePaths)
+static StringSet parseStrings(DerivationParserContext & str, bool arePaths)
 {
     StringSet res;
     while (!endOfList(str))
@@ -253,12 +288,12 @@ static DerivationOutput parseDerivationOutput(const Store & store,
     }
 }
 
-static DerivationOutput parseDerivationOutput(const Store & store, std::istringstream & str)
+static DerivationOutput parseDerivationOutput(const Store & store, DerivationParserContext & str)
 {
-    expect(str, ","); const auto pathS = parseString(str);
-    expect(str, ","); const auto hashAlgo = parseString(str);
-    expect(str, ","); const auto hash = parseString(str);
-    expect(str, ")");
+    str.expect(','); const auto pathS = parseString(str);
+    str.expect(','); const auto hashAlgo = parseString(str);
+    str.expect(','); const auto hash = parseString(str);
+    str.expect(')');
 
     return parseDerivationOutput(store, pathS, hashAlgo, hash);
 }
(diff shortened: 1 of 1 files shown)


src/libstore/derivations.cc
```diff
@@ -103,31 +103,72 @@ StorePath writeDerivation(Store & store,
 }
 
 
-/* Read string `s' from stream `str'. */
-static void expect(std::istream & str, std::string_view s)
+/* Derivation parser local to this file.  Keep in unnabed namespace
+   to unblock aggressive inlining. */
```
edolstra, line 107:
```suggestion
/* Derivation parser local to this file.  Keep in unnamed namespace
   to unblock aggressive inlining. */
```

What does unblock mean? Prevent aggressive inlining? BTW pretty surprising that namespaces affect inlining...

src/libstore/derivations.cc
```diff
@@ -103,31 +103,72 @@ StorePath writeDerivation(Store & store,
...
+namespace {
+
+struct DerivationParserContext {
+    const char * ptr_;
+    const char * end_;
```
edolstra, line 112: Might be nicer to use an `std::string_view` here.

src/libstore/derivations.cc
```diff
@@ -103,31 +103,72 @@ StorePath writeDerivation(Store & store,
...
+
+    DerivationParserContext(const char * b, const char * e)
+      : ptr_(b) , end_(e)
+    {}
+    char peek(void) {
+        if (ptr_ == end_)
+            throw FormatError("Unexpected EOF");
```
edolstra, line 119:
```suggestion
            throw FormatError("unexpected EOF");
```

src/libstore/derivations.cc
```diff
@@ -103,31 +103,72 @@ StorePath writeDerivation(Store & store,
...
+        return *ptr_;
+    }
+    char get(void) {
```
edolstra, line 122:
```suggestion
    char get() {
```

src/libstore/derivations.cc
trofi: Oh, good point! I assumed C++ methods have interposition restrictions (effectively forbid inline) similar to C exported functions, but from superficial test it does not.

I'll drop `namespace{}` and re-test that it has no performance impact.

src/libstore/derivations.cc
```diff
@@ -103,31 +103,72 @@ StorePath writeDerivation(Store & store,
...
+struct DerivationParserContext {
+    const char * ptr_;
+    const char * end_;
+
+    DerivationParserContext(const char * b, const char * e)
+      : ptr_(b) , end_(e)
+    {}
+    char peek(void) {
+        if (ptr_ == end_)
+            throw FormatError("unexpected EOF");
```
21stce, line 119:
```suggestion
            throw FormatError("Unexpected EOF");
```

21stce (commented):
Removed by @grahamc.

src/libstore/derivations.cc
pennae: keeping the anonymous namespace around the struct would be a good idea, otherwise `DerivationParserContext` could in theory conflict with like-named declarations in other files. this isn't all that likely, but it seems like good practice to prevent such things.

src/libstore/derivations.cc
pennae: `string_view` also directly carries the size instead of the end pointer, which could save a load and a compare to check for EOF in peek and get. might also make the scan loop in `parseString` a bit easierto read (using `find_first_of("\"\\")`)

21stce:
@grahamc I am not sure what you want me to say. Some of these issues *are* unacceptable for me in any productive environment. 

Can you perhaps tell me what level of competence you are aiming for? 

How can one criticize without offending someone? Fact of the matter is that *I* would rewrite the whole thing, if a company asked me what I think about it. Do you want me to say I find it "sub-optimal" or that it "could use improvements"? That seems to contain far less information then the thing I want to communicate. At that point you are saying that it is illegal for me to express a review.

Or otherwise, please rewrite the review yourself (such that it still contains the same information), since you believe it can be written in a way that doesn't offend you and apparently someone that loved your comment.

The version I had written before already was the nice version, FYI.

src/libstore/derivations.cc
trofi: I dropped the `namespace{}` to be more in line with the rest of `nix`. After measuring the performance I don't see a slowdown (yay!).

src/libstore/derivations.cc
trofi: Sounds reasonable! Converted to `std::string_view`. The result is slightly more verbose in a few places, but `find_first_of()` is a lot less scary than initial version.

src/libstore/derivations.cc
trofi: Initial version used uppercase in both places. One of the suggestions applied was to lower-case the first occurence (with assumption that rest of occurences will be lower-cased after the other material changes).

All entries are lower-case now.

trofi:
After `std::string_view` change things are still faster. Before the change:

```
$ for i in `seq 1 3`; do time nix store gc --dry-run; done

real    4m38,797s
real    4m43,469s
real    4m46,344s

$ for i in `seq 1 3`; do time nix store gc --dry-run; done

real    3m54,591s
real    3m58,574s
real    4m4,834s
```

Only 15% difference. The input is slightly bigger than before, thus the different absolute times. I'd say it's still a useful change.

trofi:
> I am not sure what you want me to say. Some of these issues _are_ unacceptable for me in any productive environment.

I suggest avoiding vague judgement statements and focus on specific code at hand. I personally can't act on feedback similar to `rush job` or `whole module should be rewritten`. I need more specifics. Ideally if the changes you are thinking of are orthogonal to the change being proposed I suggest you to send a pull request that improves the module. Checking out written code review guide for other projects might also be useful. Here is a short one: https://google.github.io/eng-practices/review/reviewer/.

grahamc:
@21stce The issues I highlighted are not your concerns of the code, but the way you're treating the author of the PR. I don't necessarily agree with your concerns but treating contributors with respect is a fundamentally important part of participating in this community. Constructive criticism is always welcome, but courtesy and respect must also be present.

Regarding your request for me to rewrite your review: it isn't our job to teach you manners. However, it is our job to foster a functional, happy, and constructive environment for the project. It looks like @trofi has some useful resources for code review.

Please take care to engage constructively and with respect toward the contributors in the future.

edolstra:
BTW, I didn't get a statistically significant speedup (on a store of 107K .drv files):
```
# ~eelco/.nix-profile/bin/benchmark -f gc-after -c gc-before -- ~eelco/Dev/nix/outputs/out/bin/nix store gc --dry-run
run 0...
...
run 9...
note: discarding 10% highest outliers
maximum RSS:        median = 307904.0000  mean = 307798.2222  stddev =    222.4600  min = 307364.0000  max = 308020.0000  [not rejected, p=0.83382, Δ=20.88889±318.53455]
soft page faults:   median =  73789.0000  mean =  73789.5556  stddev =      4.8505  min =  73781.0000  max =  73795.0000  [not rejected, p=0.50456, Δ=1.66667±7.93906]
system CPU time:    median =      5.6485  mean =      5.6538  stddev =      0.0852  min =      5.4705  max =      5.7617  [not rejected, p=0.03068, Δ=-0.08244±0.11311]
user CPU time:      median =     11.3174  mean =     11.3107  stddev =      0.0711  min =     11.1950  max =     11.4020  [not rejected, p=0.62668, Δ=-0.01726±0.11321]
elapsed time:       median =     16.9751  mean =     17.0231  stddev =      0.0868  min =     16.9213  max =     17.1369  [not rejected, p=0.01790, Δ=-0.10562±0.13020]
```
("not rejected" meaning that the null hypothesis of there being no difference was not rejected).

GC is much faster on my system, only ~17s rather than several minutes. Maybe this is related to #6207?

trofi:
Yeah, I have `experimental-features = flakes nix-command ca-derivations` and build full `nixpkgs` from `staging` in CA mode frequently. `keep-outputs` / `keep-derivations` are defaults (I think it means `keep-derivations = true`).

trofi:
`strace`-ing `nix-daemon` while it's in garbage collection I see that it opens `.drv` files orders of magnitude more than I have (millons of opens for a system with 40K files).

That would explain why `.drv` parser for me gets stressed much more than amount of data available on disk. Should be fixable by avoiding visiting identical `.drv` files somewhere in `LocalStore::collectGarbage`.

My hunch it's a side-effect of multiple realisations available for one derivation path that causes repeated visits. I'll debug in more detail where they come from and will try to come up with another PR.

trofi:
One-liner was enough: https://github.com/NixOS/nix/pull/6281. Got GC performance down to 14s on my system.

trofi:
I wonder why you don't see any difference. Could it be that you have `keep-derivations = false`?

With https://github.com/NixOS/nix/pull/6281 applied the change is not as drastic:

```
# before:
$ for i in `seq 3`; do time nix store gc --dry-run; done

real    0m14,992s
real    0m14,749s
real    0m14,907s
```

```
# after:
$ for i in `seq 3`; do time nix store gc --dry-run; done

real    0m13,948s
real    0m13,458s
real    0m13,952s
```

I can measure at best 5% difference. I personally still think it's an ok change.

Worth applying or the benefit is too small?

Ericson2314:
BTW see 493961b6899e7f3471e7efa24ed251c7723adbcd which used to exist to solve this problem.

Ericson2314:
Also see https://github.com/obsidiansystems/nix/blob/6f6a169a285bf7ba7016d5d6418518523e089e7c/src/libutil/istringstream_nocopy.hh

edolstra:
@Ericson2314 Not sure `istringstream_nocopy` would help much here, since the code is already doing a move rather than copy of the input string.

Ericson2314:
Oh, i see the slow thing was `expect` itself. `expect` can just constructor a `std::string_view` instead of a `std::string` from it's VLA, so why don't we just do that instead?

src/libstore/derivations.cc
```diff
@@ -103,31 +103,66 @@ StorePath writeDerivation(Store & store,
 }
 
 
-/* Read string `s' from stream `str'. */
-static void expect(std::istream & str, std::string_view s)
-{
-    char s2[s.size()];
-    str.read(s2, s.size());
-    if (std::string(s2, s.size()) != s)
```
Ericson2314, line 111: Just make this `std::string_view`. After that, we can see if any further benchmarking is worth it, but I doubt it.

trofi: Can you clarify what I should change `str.read()` to? Note that `str` is the `std::istream`.

trofi: Also note that initial profile shows `istream::get()` as a big part of it. That probably means it's heavyweight implementation that includes exceptions handling does not inline well.

pennae: we've also seen that while optimizing the evaluator. istreams are *incredibly* expensive if what you really want is a char iterator. just getting rid of that saves a whole bunch of non-inlineable virtual calls and other overhead, that's definitely worth doing.

trofi:
Updated the commit to claim more modest ~4% performance change in `keep-derivations = true` systems.

trofi:
Assuming undesired.

src/libstore/derivations.cc
Ericson2314: The `istringstream_nocopy` part solved that.

PR #6268, opened 2022-03-16
Simplify the handling of the hash modulo

thufschmitt:
Rather than having four different but very similar types of hashes, make
only one, with a tag indicating whether it corresponds to a regular of
deferred derivation.

This implies a slight logical change: The original Nix+multiple-outputs
model assumed only one hash-modulo per derivation. Adding
multiple-outputs CA derivations changed this as these have one
hash-modulo per output. This change is now treating each derivation as
having one hash modulo per output.
This obviously means that we internally loose the guaranty that
all the outputs of input-addressed derivations have the same hash
modulo. But it turns out that it doesn’t matter because there’s nothing
in the code taking advantage of that fact (and it probably shouldn’t
anyways).

The upside is that it is now much easier to work with these hashes, and
we can get rid of a lot of useless `std::visit{ overloaded`.

### DIFF
--- a/src/libstore/derivations.cc
+++ b/src/libstore/derivations.cc
@@ -474,7 +474,7 @@ Sync<DrvHashes> drvHashes;
 /* Look up the derivation by value and memoize the
    `hashDerivationModulo` call.
  */
-static const DrvHashModulo pathDerivationModulo(Store & store, const StorePath & drvPath)
+static const DrvHash pathDerivationModulo(Store & store, const StorePath & drvPath)
 {
     {
         auto hashes = drvHashes.lock();
@@ -509,7 +509,7 @@ static const DrvHashModulo pathDerivationModulo(Store & store, const StorePath &
    don't leak the provenance of fixed outputs, reducing pointless cache
    misses as the build itself won't know this.
  */
-DrvHashModulo hashDerivationModulo(Store & store, const Derivation & drv, bool maskOutputs)
+DrvHash hashDerivationModulo(Store & store, const Derivation & drv, bool maskOutputs)
 {
     auto type = drv.type();
 
@@ -524,7 +524,10 @@ DrvHashModulo hashDerivationModulo(Store & store, const Derivation & drv, bool m
                 + store.printStorePath(dof.path(store, drv.name, i.first)));
             outputHashes.insert_or_assign(i.first, std::move(hash));
         }
-        return outputHashes;
+        return DrvHash{
+            .hashes = outputHashes,
+            .kind = DrvHash::Kind::Regular,
+        };
     }
 
     auto kind = std::visit(overloaded {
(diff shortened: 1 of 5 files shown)


Ericson2314:
Mmm first of all, I think (floating) CA derivations has nothing to do with this. `CaOutputHashes` I think is misnamed by me, because it *only* is used for fixed output derivations.

Ericson2314:
Aye, part of my is still partial to the original, because while it is more verbose, the `std::variant` helps me understand/remember how the two "different" strategies work, and what their invariants are.

Still, I agreed before we could do this that after #6229, and I shouldn't reneg on the deal, so :+1:.

src/libstore/derivations.cc
```diff
@@ -510,10 +510,9 @@ static const DrvHashModulo pathDerivationModulo(Store & store, const StorePath &
  */
 DrvHashModulo hashDerivationModulo(Store & store, const Derivation & drv, bool maskOutputs)
 {
-    auto kind = DrvHash::Kind::Regular;
+    auto kind = DrvHashModulo::Kind::Regular;
```
Ericson2314, line 513:
```suggestion
```

src/libstore/derivations.cc
```diff
@@ -523,15 +522,12 @@ DrvHashModulo hashDerivationModulo(Store & store, const Derivation & drv, bool m
...
-        break;
-    case DerivationType::DeferredInputAddressed:
-        break;
+        return DrvHashModulo{
+            .hashes = outputHashes,
+            .kind = kind,
+        };
+    } else if (drv.type() == DerivationType::CAFloating) {
+        kind = DrvHashModulo::Kind::Deferred;
     }
```
Ericson2314, line 531: I realized in my PRs that this definition could move down to all be in one place. I also realized this fairly subtle thing (that we already the case) that I think deserves a comment.
```suggestion
    }
    
     /* Deferred input addressed might be done "pessimistically", so we
        still do `Regular` in that case. If it is necessarily deferred,
        we will catch it processing inputDrvs. */
    auto kind = drv.type() == DerivationType::CAFloating
        ? DrvHashModulo::Kind::Deferred
        : DrvHashModulo::Kind::Regular;
```

src/libstore/derivations.cc
Ericson2314: If we do the below.

src/libstore/derivations.cc
thufschmitt: Yup’ indeed that’s slightly nicer, thanks :)

Ericson2314:
https://github.com/NixOS/nix/compare/master...Ericson2314:remove-the-variant-in-hashmodulo fixed conflicts again.

Ericson2314:
Note that #6227's `NoHash` conflicts with this, but I don't think we need the `NoHash` case.

thufschmitt:
> @thufschmitt Can you have a look at the merge conflict?

Oh thanks, didn’t notice it. Done

PR #6270, opened 2022-03-16
nix: allow using --file - to read from stdin

Artturin:
Supersedes https://github.com/NixOS/nix/pull/2583

`echo '{ foo = 1 + 1; }' | nix eval -f - foo`


thufschmitt (approved):
Nice, thanks :)

PR #6280, opened 2022-03-17
Fix mv in different filesystems

### LINKED ISSUE #6262: nix fails due to `rename` returning `EXDEV` in podman (non-root overlayfs)
I've been chasing down an error for running nix in podman (which uses overlayfs as a non-root user).

It started here: https://discourse.nixos.org/t/weird-error-when-running-nixos-nix-under-podman/18056
and continued here: https://github.com/containers/podman/discussions/13507

Podman can't seem to do anything and nix needs to gracefully handle `rename` returning an error to fix this.

```
> podman image inspect nixos/nix
[
     {
          "Id": "9583adf088c452bb8861cf2eb9c92574386ec78205847622dcf39f48f3385cb8",
          "Digest": "sha256:fc55b9bf9f61742a3fc262c0dc9ad62ea8ace014bb5bd4b11341da879e7b26ce",
          "RepoTags": [
               "docker.io/nixos/nix:latest"
          ],
<snip>
```

```
> podman run --rm -ti nixos/nix nix --version
nix (Nix) 2.7.0
```

thufschmitt:
Hopefully fixes #6262 (but I couldn’t manage to reproduce it, so I’m not sure).

This moves things around a bit (well, d4fc2b5a7325f6922c180bfde9562a299164b6ac mostly), but the gist of the PR is that calls to `rename` are replaced by a new `moveFile` function which tries the renaming and fallbacks to copying if the filesystems differ (as suggested in https://github.com/containers/podman/discussions/13507#discussioncomment-2363103 ).

I’d love to add some hydra tests for this (or more generally ensuring that the docker image works fine both in docker and podman), but that’ll take a bit more time, so I’ll leave this for later.


thufschmitt:
Oh right, always that OSX filesystem link issue. I’ll try to remember how I fixed it last time :unamused:

edolstra:
Note that we shouldn't just replace `rename()` with a "move-or-copy", since there is a general assumption that renames are atomic. E.g. profile upgrades are only atomic because `rename()` is.

I think generally we should just avoid moving across filesystems. Where are we doing that?

thufschmitt:
> I think generally we should just avoid moving across filesystems. Where are we doing that?

The root issue is https://github.com/NixOS/nix/issues/6262 in which Nix fails on podman because of some silly overlayfs behavior (https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html?highlight=overlayfs#renaming-directories). More generally, I think this kind of behavior might happen in unexpected places so that’s probably a good idea to have it to prevent Nix from crashing in such a case

> Note that we shouldn't just replace rename() with a "move-or-copy", since there is a general assumption that renames are atomic.

Yes you’re right, the fix should be more subtle − both in terms of only using the function in the right places and having a more atomic behavior (maybe copy to a temp location that’s guaranteed to be on the same fs, and then move from it or something like that).

Ericson2314:
I suppose we can a `renameNonatomic` that alone has this workaround? Agreeing with @thufschmitt that the fix could be a bit more clever, though it doesn't give us full atomicity. Using a name like this signals to the reader of the code what can go wrong, and caller-site comments can explain the "we don't need atomicity here, and we would like to not break with overlayfs here" reasoning.

I think the "bad news in the name (`renameNonatomic`), good news in the comment (overlayfs story)" is nicely conservative, by giving the bad parts a bit more emphasis than the silver lining. That conservatism in what is emphasized to future readers of the code makes me feel much better about this new complexity.

MagicRB:
I may have encountered this, I'm not sure but I'm getting
```
error: linking '/nix/store/pmv5v9q4x2655qms3cavq73l91g6lhkb-init.drv.chroot/nix/store/07y6c51dam7zliaznnnzvxsc6i7jws0q-2' to '/nix/store/07y6c51dam7zliaznnnzvxsc6i7jws0q-2': Invalid cross-device link
```
`/nix/store` is zfs and i assume that the chroot is tmpfs or something

thufschmitt:
@MagicRB That seems to be slightly different (your error is about linking, not moving), but strinkingly similar. What’s the exact context in which you encountered this? (in particular: which kind of command did you rur? and was it in a docker/podman container like in the original issue? )

> /nix/store is zfs and i assume that the chroot is tmpfs or something

The chroot should just be a subdir of `/nix/store` (its root is created [here](https://github.com/NixOS/nix/blob/d354fc30b9768ea3dc737a88b57bf5e26d98135b/src/libstore/build/local-derivation-goal.cc#L584-L585), and the `/nix/store` below it a few lines below).

It looks like the failure comes from [here](https://github.com/NixOS/nix/blob/d354fc30b9768ea3dc737a88b57bf5e26d98135b/src/libstore/build/local-derivation-goal.cc#L347-L361) (which is called while building the content of the sandbox). The code already falls back to copying for some error codes, but not for `EXDEV`.

MagicRB:
Nope this is on NixOS. I can get you the exact command with the rev and everything later. But it was `nix build git+https://gitea.redalder.org/RedAdler/systems#nixngSystems.hydra.config.system.build.toplevel` i think

thufschmitt:
@edolstra I've fixed (I think) the OSX failure. Does that look good?

configure.ac
```diff
@@ -282,15 +282,6 @@ AC_CHECK_FUNCS([setresuid setreuid lchown])
 AC_CHECK_FUNCS([strsignal posix_fallocate sysconf])
 
 
-# This is needed if bzip2 is a static library, and the Nix libraries
```
thufschmitt, line 285: @edolstra I've brutally commented that out because it was breaking the libc++fs link (dunno why, but I ended-up with a duplicate symbol error). Do you think it's still needed?

edolstra:
TBH I'm reluctant to start using `std::filesystem`, because it puts another layer between us and POSIX, while we often need to know exactly what's going on. E.g. how does `fs::copy()` deal with ACLs? Does `fs::rename()` have the same atomicity guarantees? And it introduces another path type with its own semantics...

configure.ac
edolstra: No idea, I'm fine with removing it, we'll see what breaks :-)

thufschmitt:
> TBH I'm reluctant to start using std::filesystem, because it puts another layer between us and POSIX, while we often need to know exactly what's going on

That's a good point (I was actually tripped over that while implementing this PR because although `std::filesystem::copy` can recursively copy directories, it will at least update the `mtime` in the copied version, so I had to roll ou my own implementation).
That being said, I think that despite its quirks, it's nicer than directly hitting the underlying C APIs for two reasons:

1. We need some kind of layer. Having all these inlined `c_str()` and error code checks is both painful to read and error-prone, and if we must have something on top of that, then I'd rather trust the one from the stl rather than a custom one

2. More importantly, it's a layer that doesn't just sit on top of POSIX, but also on top of OS-specific APIs (be it Unix, Windows, Linux or OSX), and it gives us a unified interface for all of these (through which we can break when needed like I did in this PR), which is very valuable – esp. if we consider ever having a windows support, which I think would be incredibly great

I'm not married to it though (and it wouldn't fundamentally change this PR), so if you're strongly against it I can rewrite it to use the good ol' Unix APIs (although that would be a non-trivial amount of work, so I'd rather not if possible :upside_down_face: )

thufschmitt:
@edolstra friendly ping on

> I'm not married to it though (and it wouldn't fundamentally change this PR), so if you're strongly against it I can rewrite it to use the good ol' Unix APIs (although that would be a non-trivial amount of work, so I'd rather not if possible upside_down_face )

Any final word on the matter? :)

thufschmitt:
@edolstra I'm merging this since the `filesystem` use is small and well-contained.
If it's too much of an issue, feel free to revert or ask me to rewrite the `filesystem` parts

PR #6283, opened 2022-03-17
Clean up `DerivationOutput`, and headers

Ericson2314:
1. `DerivationOutput` now as the `std::variant` as a base class. And the
   variants are given hierarchical names under `DerivationOutput`.

   In 8e0d0689be797f9e42f9b43b06f50c1af7f20b4a @matthewbauer and I
   didn't know a better idiom, and so we made it a field. But this sort
   of "newtype" is anoying for literals downstream.

   Since then we leaned the base class, inherit the constructors trick,
   e.g. used in `DerivedPath`. Switching to use that makes this more
   ergonomic, and consistent.

2. `store-api.hh` and `derivations.hh` are now independent.

   In bcde5456cc3295061a0726881c3e441444dd6680 I swapped the dependency,
   but I now know it is better to just keep on using incomplete types as
   much as possible for faster compilation and good separation of
   concerns.

### DIFF
--- a/src/libstore/repair-flag.hh
+++ b/src/libstore/repair-flag.hh
@@ -1,0 +1,7 @@
+#pragma once
+
+namespace nix {
+
+enum RepairFlag : bool { NoRepair = false, Repair = true };
+
+}
--- a/src/libexpr/get-drvs.cc
+++ b/src/libexpr/get-drvs.cc
@@ -1,6 +1,7 @@
 #include "get-drvs.hh"
 #include "util.hh"
 #include "eval-inline.hh"
+#include "derivations.hh"
 #include "store-api.hh"
 #include "path-with-outputs.hh"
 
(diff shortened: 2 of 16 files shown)


src/libstore/repair-flag.hh
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+namespace nix {
+
+enum RepairFlag : bool { NoRepair = false, Repair = true };
```
Ericson2314, line 5: This is sort of silly, but is the one definition used by `store-api.hh` and `derivations.hh`. I cannot make it abstract because the `NoRepair` default arguments.

Ericson2314:
@edolstra I know this will conflict heavily with #6227, so I am offering to rebase that one on top if this goes in first.

(This one changes absolutely no functionality, and will make the code easier to read in that one while we hash out the last few details (I would like to do the "derivation type" change) so I think it does make sense to land this first.)

thufschmitt (approved):
Looks good :+1: thanks :)

PR #6285, opened 2022-03-18
nix-daemon.conf.in: add tmpfiles file to create nix/daemon-socket directory

flokli:
nix-daemon.socket is used to socket-activate nix-daemon.service when
/nix/var/nix/daemon-socket/socket is accessed.

In container usecases, sometimes /nix/var/nix/daemon-socket is
bind-mounted read-only into the container.

In these cases, we want to skip starting nix-daemon.socket.

However, since systemd 250, `ConditionPathIsReadWrite` is also not met
if /nix/var/nix/daemon-socket doesn't exist at all. This means, a
regular NixOS system will skip starting nix-daemon.socket:

> [  237.187747] systemd[1]: Nix Daemon Socket was skipped because of a failed condition check (ConditionPathIsReadWrite=/nix/var/nix/daemon-socket).

To prevent this from happening, ship a tmpfiles file that'll cause the
directory to be created if it doesn't exist already.

In the case of NixOS, we can just add Nix to `systemd.tmpfiles.packages`
and have these files picked up automatically.

### DIFF
--- a/scripts/install-systemd-multi-user.sh
+++ b/scripts/install-systemd-multi-user.sh
@@ -9,6 +9,8 @@ readonly SERVICE_DEST=/etc/systemd/system/nix-daemon.service
 readonly SOCKET_SRC=/lib/systemd/system/nix-daemon.socket
 readonly SOCKET_DEST=/etc/systemd/system/nix-daemon.socket
 
+readonly TMPFILES_SRC=/lib/tmpfiles.d/nix-daemon.conf
+readonly TMPFILES_DEST=/etc/tmpfiles.d/nix-daemon.conf
 
 # Path for the systemd override unit file to contain the proxy settings
 readonly SERVICE_OVERRIDE=${SERVICE_DEST}.d/override.conf
@@ -83,6 +85,13 @@ EOF
 poly_configure_nix_daemon_service() {
     if [ -e /run/systemd/system ]; then
         task "Setting up the nix-daemon systemd service"
+
+        _sudo "to create the nix-daemon tmpfiles config" \
+              ln -sfn /nix/var/nix/profiles/default/$TMPFILES_SRC $TMPFILES_DEST
+
+        _sudo "to run systemd-tmpfiles once to pick that path up" \
+             sytemd-tmpfiles create --prefix=/nix/var/nix
+
         _sudo "to set up the nix-daemon service" \
               systemctl link "/nix/var/nix/profiles/default$SERVICE_SRC"
 
--- a/misc/systemd/local.mk
+++ b/misc/systemd/local.mk
@@ -1,7 +1,8 @@
 ifdef HOST_LINUX
 
   $(foreach n, nix-daemon.socket nix-daemon.service, $(eval $(call install-file-in, $(d)/$(n), $(prefix)/lib/systemd/system, 0644)))
+  $(foreach n, nix-daemon.conf, $(eval $(call install-file-in, $(d)/$(n), $(prefix)/lib/tmpfiles.d, 0644)))
 
-  clean-files += $(d)/nix-daemon.socket $(d)/nix-daemon.service
+  clean-files += $(d)/nix-daemon.socket $(d)/nix-daemon.service $(d)/nix-daemon.conf
 
 endif
--- a/misc/systemd/nix-daemon.conf.in
+++ b/misc/systemd/nix-daemon.conf.in
@@ -1,0 +1,1 @@
+d @localstatedir@/nix/daemon-socket 0755 root root - -


Ericson2314:
@flokli this line of thinking makes sense to me, but I am never up to date on the systemd details I don't take my own judgement here very seriously :).

Ericson2314:
As the the container stuff, I like how this ignores it.

It would be good to reach out the systemd people on this, but I think there ought to be something more explicit about when a service inside the container is to be provided by outside the container. E.g. perhaps the inner one should have a different `.service` file that says "host delegated" or similar? (I am also unclear whether running systemd *inside* containers is standard practice or not.)

flokli:
> It would be good to reach out the systemd people on this, but I think there ought to be something more explicit about when a service inside the container is to be provided by outside the container. E.g. perhaps the inner one should have a different .service file that says "host delegated" or similar? (I am also unclear whether running systemd inside containers is standard practice or not.)

`nixos-container create foo; nixos-container start foo; machinectl shell foo` will get you into a NixOS container, with its own systemd. That's nothing new, and what the `ConditionPathIsReadWrite=` check was also used for - in the case of a bind-ro directory, the nix-daemon startup was skipped. This also didn't regress, so we don't need to involve systemd into that.

What regressed was that the unit file is now getting skipped on the host, because `ConditionPathIsReadWrite=` now will also skip it the folder doesn't exist at all (which is reasonable, it previously just worked by accident).

The tmpfiles file here will now make sure a directory is present on the host, and will still properly skip as intended inside the container (and leave the read-only bind mount alone)

Ericson2314:
@flokli If the container has its own systemd, shouldn't it also have its own unit files (or "active" unit files?)? I would want the inner systemd to one of:

1. Not know about the `nix-daemon` service at all. The socket just exists, and socket activation will stil lwork.
2. Know it is provided by the external systemd, so in addition to the above, there could be a limited protocol also for the inside being able to request that the outside systemd start the service, recreate the socket, etc.

flokli:
>1. Not know about the `nix-daemon` service at all. The socket just exists, and socket activation will stil lwork.
>2. Know it is provided by the external systemd, so in addition to the above, there could be a limited protocol also for the inside being able to request that the outside systemd start the service, recreate the socket, etc.

Yes, we could introduce an option to not install the nix-daemon, and corresponding unit files, and default to using this in non-NixOS, and then possibly drop the "skip if something else provides the socket".

But this is somewhat orthogonal to this PR. With systemd 250, nix-daemon on the host gets skipped (on non-NixOS too), and the tmpfile config introduced here makes sure the folder is present, so the socket unit isn't skipping anymore.

Ericson2314:
Yeah agreed it is orthogonal. I leave it to @edolstra to review this, but this seems like a good thing to do. I made it so it is possible to disable the nix daemon in NixOS, and that should probably be used instead until systemd has the extra magic.

flokli:
> I made it so it is possible to disable the nix daemon in NixOS, and that should probably be used instead until systemd has the extra magic.

systemd doesn't need any extra magic for this. We just need to have something creating that folder, other than the `.socket` unit. For now, I moved this to a downstream `systemd.tmpfile.rules` in https://github.com/NixOS/nixpkgs/pull/164644 to unblock the channels.

Shipping a tmpfiles file via `lib/tmpfiles.d/nix-daemon.conf` in Nix itself is the more canonical way to describe this, though, especially when interacting with other distros - so we should move to that once its in.

flokli:
@edolstra can you review this? Something equivalent ended up [downstream in nixpkgs](https://github.com/NixOS/nixpkgs/pull/164644), but other distros packaging Nix might run into the same issues if they use our provided `nix-daemon.service` and bump to systemd 250.

edolstra:
Should `install-systemd-multi-user.sh` be updated to install this file?

flokli:
I wasn't aware of that script. I think so, yes. We should install it to `/etc/tmpfiles.d/`.

flokli:
No, I read `install-multi-user.sh`, which executes `install-systemd-multi-user.sh`.

Its `create_directories()` function already does create the daemon folder, so we don't need to extend it.

edolstra:
Shouldn't it run `systemd-tmpfiles --create` to create the socket directory?

Ericson2314:
@flokli I think @edolstra means, the installer does `systemctl link "/nix/var/nix/profiles/default$SERVICE_SRC"`, but will that install the this new file too? Or just the service file?

flokli:
install-multi-user.sh`'s `create_directories()` already creates the folder.

However, we you can duplicate this in `install-systemd-multi-user.sh` too.

I made an attempt, assuming `$out` from Nix ends up in `/nix/var/nix/profiles/default`, and pushed it to this PR.

I lack the tooling to test this, review welcome.

scripts/install-systemd-multi-user.sh
```diff
@@ -83,6 +85,13 @@ EOF
 poly_configure_nix_daemon_service() {
     if [ -e /run/systemd/system ]; then
         task "Setting up the nix-daemon systemd service"
+
+        _sudo "to create the nix-daemon tmpfiles config" \
+              cp /nix/var/nix/profiles/default/$TMPFILES_SRC $TMPFILES_DEST
```
Ericson2314, line 90:
```suggestion
              ln -s /nix/var/nix/profiles/default/$TMPFILES_SRC $TMPFILES_DEST
```
I think we want that, as IIRC that matches what `systemd link` will do, and also so upgrading Nix upogrades this file too.

flokli: I picked `ln -sfn`, to make this more idempotent if there's already an existing `$TMPFILES_DEST`. I assume `systemd link` does that too.

Ericson2314 (approved):
If the tempfiles thing is good (which I am still fundamentally not qualified to answer, but I think it is) this seems like the right way to do it.

flokli:
@edolstra can you make a call here?

flokli:
Thanks! I assume this should be backported to the `*-maintenance` branches?

What's the process here? Can you do a cherry-pick, or do I open individual PRs?

edolstra:
Before backporting, it would be great if somebody can test the installer with `--daemon`.

scripts/install-systemd-multi-user.sh
```diff
@@ -83,6 +85,13 @@ EOF
 poly_configure_nix_daemon_service() {
     if [ -e /run/systemd/system ]; then
         task "Setting up the nix-daemon systemd service"
+
+        _sudo "to create the nix-daemon tmpfiles config" \
+              ln -sfn /nix/var/nix/profiles/default/$TMPFILES_SRC $TMPFILES_DEST
+
+        _sudo "to run systemd-tmpfiles once to pick that path up" \
+             sytemd-tmpfiles create --prefix=/nix/var/nix
```
Artturin, line 93: typo
fixed in https://github.com/NixOS/nix/pull/6319

Issue #6269, opened 2022-03-16
Spaces before colon matter when they shouldn't


21stce:
**Describe the bug**

I'd expect that the command syntax is similar to that of e.g. `ghci`, but in `ghci` spaces don't matter, whereas in Nix version 2.3.16 they do. 

The screenshot should say enough. 

![image](https://user-images.githubusercontent.com/100212538/158636974-2a496b06-759b-4ada-b46a-704a0c2717a5.png)

Artturin:
the code that handles that: https://github.com/NixOS/nix/blob/fe1ad9613561ea78b6d97cf82c03d2dd77e43f69/src/nix/repl.cc#L405

PR #6292, opened 2022-03-20
nix: allow whitespace characters before command in repl

polykernel:
Before this change, processLine always uses the first character
as the start of the line. This cause whitespaces to matter at the
beginning of the line whereas it does not matter anywhere else.

This commit trims leading white spaces of the string line so that
subsequent operations can be performed on the string without explicitly
tracking starting and ending indices of the string.

Fixes https://github.com/NixOS/nix/issues/6269

### DIFF
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -396,6 +396,7 @@ StorePath NixRepl::getDerivationPath(Value & v) {
 
 bool NixRepl::processLine(std::string line)
 {
+    line = trim(line);
     if (line == "") return true;
 
     _isInterrupted = false;


thufschmitt (commented):
Looks good. Do you think it’s possible to use a `string_view` rather than keeping the index of the start? That would make the rest of the code slightly nicer (and prevent accidentally forgetting to pass the `s` argument somewhere)

src/nix/repl.cc
```diff
@@ -396,15 +396,16 @@ StorePath NixRepl::getDerivationPath(Value & v) {
 
 bool NixRepl::processLine(std::string line)
 {
-    if (line == "") return true;
+    size_t s = line.find_first_not_of(" \n\r\t");
+    if (s == std::string::npos) return true;
```
edolstra, line 400: Maybe just do `line = trim(line)`?

polykernel: I thought about this initially but I don't think the extra copy is necessary because both `arg` and `command` already remove white spaces and `line` only serves as an intermediate object where they are extracted from (please correct me if I am wrong as I am not very well-versed in C++).

polykernel:
> Do you think it’s possible to use a `string_view` rather than keeping the index of the start? 

Thank you for the suggestion. To clarify, do you mean changing processLine receive a string_view as opposed to a string or simply creating a view of the string `line` within the function?

src/nix/repl.cc
edolstra: No need to worry about efficiency here. The microsecond it takes to call `trim` isn't going to be noticeable in an interactive function :-)

thufschmitt:
> > Do you think it’s possible to use a `string_view` rather than keeping the index of the start?
> 
> Thank you for the suggestion. To clarify, do you mean changing processLine receive a string_view as opposed to a string or simply creating a view of the string `line` within the function?

I meant creating a view of the string within the function. But @edolstra ’s sugestion is as good, we don’t really care about copying the input here :)

(It would even be possible to make `trim` return a view of the original string, although I’m slightly worried of it silently breaking something else)

src/nix/repl.cc
polykernel: I see, I have applied your suggestion.

polykernel:
> I meant creating a view of the string within the function. But @edolstra ’s sugestion is as good, we don’t really care about copying the input here :)

I see, I don't know enough about the intricacies of string_view so I went with the safer option following edolstra's suggestion.

> (It would even be possible to make `trim` return a view of the original string, although I’m slightly worried of it silently breaking something else)

I also see some other functions that can benefit from string_view, but I think there is a hidden cost from tracking ownership based on my preliminary understanding. (gauging the impact of the change is definitely out of the breadth of my knowledge)

PR #6294, opened 2022-03-22
buildProfile(): Ignore manifest.{nix,json}

### LINKED ISSUE #6032: nix-env --upgrade silently converts profile to new experimental format (on Ubuntu)
**Describe the bug**

After running `nix-env -u`, my nix single-user profile apparently gets upgraded to a new format as if I used the experimental `nix profile` command (which I **never did**). All subsequent `nix-env` calls such as installing a package or `-u` report:

    error: profile '/nix/var/nix/profiles/per-user/vojtech/profile' is incompatible with 'nix-env'; please use 'nix profile' instead

**Steps To Reproduce**

1. Install nix based on the instructions [here](https://nixos.org/download.html), single-user variant.
2. Install a couple of packages...
3. Update them once in a while with `nix-env -u`/
4. The profile becomes unusable, `nix-env` reports the above error.
5. I have to rollback the profile, which works, but I can no longer upgrade with `nix-env -u` as that will (reproducibly) lead to the problem again.

**Expected behavior**

Profile stays in the non-experimental format?

`nix-env (Nix) 2.6.0`

**Additional context**

I didn't do anything special with the profile at all, basically the only commands I use are `nix-env -iA $package`, `nix-channel --update`,  `nix-env -u`, and `nix-collect-garbage --delete-older-than 30d`. (And the rollback one I guess after this started happening.)

edolstra:
If a package installs a file named `manifest.json`, it caused `nix-env` to consider the profile a new-style profile created by `nix profile`. Fixes #6032.

### DIFF
--- a/src/libstore/builtins/buildenv.cc
+++ b/src/libstore/builtins/buildenv.cc
@@ -47,9 +47,9 @@ static void createLinks(State & state, const Path & srcDir, const Path & dstDir,
             throw;
         }
 
-        /* The files below are special-cased to that they don't show up
-         * in user profiles, either because they are useless, or
-         * because they would cauase pointless collisions (e.g., each
+        /* The files below are special-cased to that they don't show
+         * up in user profiles, either because they are useless, or
+         * because they would cause pointless collisions (e.g., each
          * Python package brings its own
          * `$out/lib/pythonX.Y/site-packages/easy-install.pth'.)
          */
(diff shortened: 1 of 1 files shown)


ElvishJerricco:
It might have been worth adding some logic to `nix-env` to detect an incorrectly installed `manifest.json`. As is, if you had e.g. `element-web` installed with an older `nix-env` than this change, then even upgrading to a version *with* this change would still leave you with a broken profile.

PR #6296, opened 2022-03-22
Don't hide repeated values while generating manifest.nix

### LINKED ISSUE #6243: error: file 'REPEAT' was not found in the Nix search path
See https://github.com/NixOS/nixpkgs/issues/163374

Could nixpkgs commiters get issue management perms here so we could transfer issues from nixpkgs?

edolstra:
Fixes #6243.
Fixes https://github.com/NixOS/nixpkgs/issues/163374.


Artturin:
Thanks eelco

Issue #552, opened 2015-06-02
Document builtins.genericClosure


copumpkin:
Looks like it almost happened a while ago: https://github.com/NixOS/nix/blob/master/doc/manual/release-notes/rl-0.12.xml#L150

Seems like a generally useful primop!

sid-kap:
@copumpkin Would you happen to know where the source of `builtins.genericClosure` is? I did a Github code search for "org:NixOS genericclosure" but I could only find uses of `genericClosure`, not the definition. :/

copumpkin:
Yep! https://github.com/NixOS/nix/blob/master/src/libexpr/primops.cc#L285

sid-kap:
Thank you!

copumpkin:
From @edolstra on IRC, in case anyone's wondering:

```
but basically, it takes a "startSet" (a list of initial graph nodes) and a "operator" function that must return nodes reachable from a node
nodes are represented as attribute sets with a "key" attribute
so "key" could be the package name in this case
any other attributes are passed through
```

cdepillabout:
It would be nice to get this documented.

I put together a little example of `genericClosure` so that I could understand it better.  Hopefully this helps someone:

```nix
# `builtins.genericClosure` is great for computing transitive dependencies
# given a list of packages and the direct dependencies for each one.
#
# This file is a small example of that.

let
  # Imagine this is an attribute set of package names to dependency mappings.
  # Each attribute key represents a package name.  Each attribute value has
  # two fields: `depNames` and `val`:
  #
  # - `depNames` represents the names of all the dependencies for each
  #   package.  For instance, package `a` has dependencies on packages
  #   `b` and `c`.  Package `b` has no dependencies.
  #
  # - `val` represents any other value associated with a package.  This is
  #   mainly used just to show that values can be carried around in
  #   genericClosure.
  allDeps = {
    a = { depNames = [ "b" "c" ]; val = "val for a"; };
    b = { depNames = [ ]; val = "val for b"; };
    c = { depNames = [ "d" ]; val = "val for c"; };
    d = { depNames = [ "b" ]; val = "val for d"; };
    e = { depNames = [ "d" ]; val = "val for e"; };
    f = { depNames = [ "g" "h" ]; val = "val for f"; };
    g = { depNames = [ ]; val = "val for g"; };
    h = { depNames = [ "a" "b" "c" "d" "e" ]; val = "val for h"; };
[...]

PR #6305, opened 2022-03-24
docs: genericClosure

tomberek:
Fixes: #552

(If anyone finds a non-terminating example for that function, please let me know!)

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -694,7 +694,32 @@ static void prim_genericClosure(EvalState & state, const Pos & pos, Value * * ar
 
 static RegisterPrimOp primop_genericClosure(RegisterPrimOp::Info {
     .name = "__genericClosure",
+    .args = {"attrset"},
     .arity = 1,
+    .doc = R"(
+      Take an *attrset* with values named `startSet` and `operator` in order to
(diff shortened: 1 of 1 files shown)


src/libexpr/primops.cc
```diff
@@ -694,7 +694,28 @@ static void prim_genericClosure(EvalState & state, const Pos & pos, Value * * ar
 
 static RegisterPrimOp primop_genericClosure(RegisterPrimOp::Info {
     .name = "__genericClosure",
+    .args = {"attrset"},
     .arity = 1,
+    .doc = R"(
+      Take an *attrset* with elements named `startSet` and `operator` in order
+      to return an *attrset* by starting with the `startSet`, recursively
+      applying the `operator` to each element with a unique key. For example,
```
edolstra, line 702: This description doesn't mention the closure property, i.e. that it continues to call the function for each new key until it returns no new keys. Also worth mentioning that the keys must be comparable (so e.g. attrsets are not allowed).

tomberek: updated

thufschmitt:
> If anyone finds a non-terminating example for that function, please let me know!

```nix
builtins.genericClosure {
  startSet = [ { key = 1; } ];
  operator = item: [{
    key = item.key + 1;
  }];
}
```

?

PR #6311, opened 2022-03-24
Make buildPathsWithResults() only return info on wanted outputs

edolstra:
This fixes a regression in `nix build`  (mentioned [here](https://github.com/NixOS/nix/pull/6259#issuecomment-1075354762)) where it would print *all* outputs when building, but only the wanted outputs otherwise. Now it only prints the wanted outputs, as before.

It also fixes a potential infinite loop when retrying substitutions after an incomplete closure.

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -311,14 +311,11 @@ void DerivationGoal::outputsSubstitutionTried()
     gaveUpOnSubstitution();
 }
 
+
 /* At least one of the output paths could not be
    produced using a substitute.  So we have to build instead. */
 void DerivationGoal::gaveUpOnSubstitution()
 {
-    /* Make sure checkPathValidity() from now on checks all
-       outputs. */
-    wantedOutputs.clear();
-
     /* The inputs must be built before we can build this goal. */
     if (useDerivation)
         for (auto & i : dynamic_cast<Derivation *>(drv.get())->inputDrvs)
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -207,8 +207,6 @@ void DerivationGoal::haveDerivation()
     if (!drv->type().hasKnownOutputPaths())
         settings.requireExperimentalFeature(Xp::CaDerivations);
 
-    retrySubstitution = false;
-
     for (auto & i : drv->outputsAndOptPaths(worker.store))
         if (i.second.second)
             worker.store.addTempRoot(*i.second.second);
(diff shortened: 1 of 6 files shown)


src/libstore/build/derivation-goal.cc
```diff
@@ -311,14 +309,11 @@ void DerivationGoal::outputsSubstitutionTried()
...
 }
 
+
 /* At least one of the output paths could not be
    produced using a substitute.  So we have to build instead. */
 void DerivationGoal::gaveUpOnSubstitution()
 {
-    /* Make sure checkPathValidity() from now on checks all
-       outputs. */
-    wantedOutputs.clear();
```
edolstra, line 320: I couldn't see any reason why `checkPathValidity()` needs to check all outputs. In any case, messing with `wantedOutputs` seems dirty...

src/libstore/build/local-derivation-goal.cc
```diff
@@ -2613,7 +2613,8 @@ DrvOutputs LocalDerivationGoal::registerOutputs()
             signRealisation(thisRealisation);
             worker.store.registerDrvOutput(thisRealisation);
         }
-        builtOutputs.emplace(thisRealisation.id, thisRealisation);
+        if (wantedOutputs.empty() || wantedOutputs.count(outputName))
```
Ericson2314, line 2616: We have a function for this.
```suggestion
        if (wantOutput(outputName, wantedOutputs))
```

src/libstore/build/derivation-goal.cc
Ericson2314: I agree it is dirty, but the idea I think (I did not write that but remember reading it IIRC) that since we are now doing the build, we will be making all afresh outputs, and we should fail if the derivation did not produce all the outputs even if we already had the missing outputs before.

Without this, we run the risk of some wonky non-deterministic derivation that produces a different subset of the outputs each time not failing as it should.

Ericson2314: Oh, another reason is that if during a build we discover we in fact need other outputs, it is nice to know we don't need to restart a build that got so upgraded. See `DerivationGoal::addWantedOutputs`

Ericson2314: A consequence of this is that I think `buildResultsWithPaths` is going to need to filter the results based on the derived path.

edolstra: From what I can tell, what outputs get build/registered (or whose absence causes the build to fail) is unaffected by `wantedOutputs` or `initialOutputs.*.wanted`.  So I think it should be okay. Restarts are only needed in the substitution case, not when we're actually building.

Ericson2314: I will fix this in the other PR too, sigh.

edolstra: Think what? I'm not sure I understand the problem...

Ericson2314: See the long PR description in https://github.com/NixOS/nix/pull/6312

PR #6312, opened 2022-03-25
Shuffle `BuildResult` data definition, make state machine clearer, introduce `SingleDrvOutputs`

Ericson2314:
# Motivation

There are two changes going on here: changing `BuildResult` and changing the goal state machine logic.

## Introduce `Worker::makeGoal`

This takes a `DerivedPath` so the caller doesn't need to care about which sort of goal does what.

## `BuildResult` changes

Make `BuildResult` be like it was before, and instead made `KeyedBuildResult` be a subclass wit the path. Only `buildPathsWithResults` returns `KeyedBuildResult`s, everything else just becomes like it was before, where the "key" is unambiguous from context.

The idea here is that storing the path when it is already unambiguous from context *denormalizes* our data model, making two sources of truth that must be kept in sync. This isn't ideal.

This is especially notable for `DerivationGoal`, where two `DerivedPath` requests may correspond to the same `DerivationGoal`. Before, the `BuildResult` of that goal (stored, then returned) would just be the first one to create that goal. That is non-deterministic!

`KeyedBuildResult` *is* useful in the return type of `KeyedBuildResult`, but this is because it is conceptually returning an mapping of `DerivedPath` keys to `BuildResult` values. We could, for example, also use `std::map<BuildResult, BuildResult>` for this purpose (and probably also have the arguments be a `std::set` not `std::vector` for symmetry), in which case we would here also not need `KeyedBuildResult`.

## State machine changes

The `needRestart`, `retrySubstitution`, and `retriedSubstitution` bools are replaced with enums. The enum cases and conditional logic are exhaustively documented to try to make the code easier to understand.

An especially thorny case with the old code was that `needRestart` was set to `true` if the outputs changed for a goal that had progressed to building, but `needRestart` would never be read again. This is an in fact a good thing operationally, as a build will produce all outputs and so restarting is not necessary. But this (a) being how it worked, and (b) being a good thing was quite obfuscated just reading the code!

The new code and comments call out this case explicitly.

## Introduce `SingleDrvOutputs`

In many cases we are dealing with a collection of realisations, they are all outputs of the same derivation. In that case, we don't need "derivation hashes modulos" to be part of our map key, because the output names alone will be unique. Those hashes are still part of the realisation proper, so we aren't loosing any information, we're just "normalizing our schema" by narrowing the "primary key".

Besides making our data model a bit "tighter" this allows us to avoid a double `for` loop in `DerivationGoal::waiteeDone`. The inner `for` loop was previously just to select the output we cared about without knowing its hash. Now we can just select the output by name directly.

# Context

## Protocol non-changes

Note that neither protocol is changed as part of this: we are still transferring `DrvOutputs` over the wire for `BuildResult`s. I would only consider revising this once #6223 is merged, and we can mention protocol versions inside factored-out serialization logic. Until then it is better not change anything because it would come a the cost of code reuse.

## Initializers

Unfortunately, we need to avoid C++20 strictness on designated initializers.

(BTW https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2287r1.html
this offers some new syntax for this use-case. Hopefully this will be
adopted and we can eventually use it.)

No having that yet, maybe it would be better to not make `KeyedBuildResult` a subclass to just avoid this.

## History of this PR

In https://github.com/NixOS/nix/pull/6311#discussion_r834863823, I thought that since derivation goals' wanted outputs can "grow" due to overlapping dependencies (See `DerivationGoal::addWantedOutputs`, called by `Worker::makeDerivationGoalCommon`), the previous bug fix had an unfortunate side effect of causing more pointless rebuilds.

In particular, I was worried about this situation:

1. Goal made from `DerivedPath::Built { foo, OutputsSpec::Names { a } }`.

2. Goal gives up on on substituting, starts building.

3. Goal made from `DerivedPath::Built { foo, OutputsSpec::Names { b } }`, in fact is just modified original goal.

4. Though the goal had gotten as far as building, so all outputs were going to be produced, `addWantedOutputs` no longer knows that and so the goal is flagged to be restarted.

This might sound far-fetched with input-addressed drvs, where we usually basically have all our goals "planned out" before we start doing anything, but with CA derivation goals and especially RFC 92, where *drv resolution* means goals are created after some building is completed, it would be more likely to happen.

@edolstra, on the other hand, thought this was not an issue because even though `needsRestart = true` would be set, it would never be read again. Eelco might be right, but nonetheless this is still very confusing storing "needs restart" but in fact neither needing nor wanting a restart!

So the first thing I did was restore the clearing of `wantedOutputs` we used to do, and then filter the outputs in `buildPathsWithResults` to only get the ones we care about.

The `KeyedBuildResult` change technically was already needed due to the "modified derivational goal with second request" issue described in the motivation section, but become more urgent with this modification of `wantedOutputs`.

Eelco however (if I recalled correctly) in real time meeting (so no paper trail) also didn't like modifying `wantedOutputs` upon falling back to doing a build (as we did before), because we should only modify it in response to new requests --- *actual* wants --- and not because we are "incidentally" building all the outptus beyond what may have been requested.

That's a fair point, and the alternative is to replace the boolean soup with proper enums: Instead of modifying `wantedOuputs` som more, we'll modify `needsRestart` to indicate we are passed the need. That lead me to the `bool` -> `enum` changes, and then I did the substitution ones in addition too for consistency.

## "Keyed" data types in general.

I think separating the "primary key" field(s) from the other fields is good practical in general. (I would like to do the same thing for `ValidPathInfo`, for example.) Besides the example in the motivation, there are probably other cases where we would like to do `std::map<Key, ThingWithoutKey>` where the "without keys" is necessary to not contain duplicate keys and just precludes the possibility of those duplicate keys being out of sync.

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1013,7 +1013,7 @@ void DerivationGoal::resolvedFinished()
     auto resolvedDrv = *resolvedDrvGoal->drv;
     auto & resolvedResult = resolvedDrvGoal->buildResult;
 
-    DrvOutputs builtOutputs;
+    SingleDrvOutputs builtOutputs;
 
     if (resolvedResult.success()) {
         auto resolvedHashes = staticOutputHashes(worker.store, resolvedDrv);
@@ -1039,7 +1039,7 @@ void DerivationGoal::resolvedFinished()
                     worker.store.printStorePath(drvPath), wantedOutput);
 
             auto realisation = [&]{
-              auto take1 = get(resolvedResult.builtOutputs, DrvOutput { *resolvedHash, wantedOutput });
+              auto take1 = get(resolvedResult.builtOutputs, wantedOutput);
               if (take1) return *take1;
 
               /* The above `get` should work. But sateful tracking of
@@ -1064,7 +1064,7 @@ void DerivationGoal::resolvedFinished()
                 worker.store.registerDrvOutput(newRealisation);
             }
             outputPaths.insert(realisation.outPath);
-            builtOutputs.emplace(realisation.id, realisation);
+            builtOutputs.emplace(wantedOutput, realisation);
         }
 
         runPostBuildHook(
@@ -1189,7 +1189,7 @@ HookReply DerivationGoal::tryBuildHook()
 }
 
 
-DrvOutputs DerivationGoal::registerOutputs()
+SingleDrvOutputs DerivationGoal::registerOutputs()
 {
     /* When using a build hook, the build hook can register the output
        as valid (by doing `nix-store --import').  If so we don't have
@@ -1351,7 +1351,7 @@ OutputPathMap DerivationGoal::queryDerivationOutputMap()
 }
 
 
-std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
+std::pair<bool, SingleDrvOutputs> DerivationGoal::checkPathValidity()
 {
     if (!drv->type().isPure()) return { false, {} };
 
@@ -1364,7 +1364,7 @@ std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
             return static_cast<StringSet>(names);
         },
     }, wantedOutputs.raw());
-    DrvOutputs validOutputs;
+    SingleDrvOutputs validOutputs;
 
     for (auto & i : queryPartialDerivationOutputMap()) {
         auto initialOutput = get(initialOutputs, i.first);
@@ -1407,7 +1407,7 @@ std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
             }
         }
         if (info.wanted && info.known && info.known->isValid())
-            validOutputs.emplace(drvOutput, Realisation { drvOutput, info.known->path });
+            validOutputs.emplace(i.first, Realisation { drvOutput, info.known->path });
     }
 
     // If we requested all the outputs, we are always fine.
@@ -1431,7 +1431,7 @@ std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
 }
 
 
-DrvOutputs DerivationGoal::assertPathValidity()
+SingleDrvOutputs DerivationGoal::assertPathValidity()
 {
     auto [allValid, validOutputs] = checkPathValidity();
     if (!allValid)
@@ -1442,7 +1442,7 @@ DrvOutputs DerivationGoal::assertPathValidity()
 
 void DerivationGoal::done(
     BuildResult::Status status,
-    DrvOutputs builtOutputs,
+    SingleDrvOutputs builtOutputs,
     std::optional<Error> ex)
 {
     buildResult.status = status;
@@ -1498,11 +1498,11 @@ void DerivationGoal::waiteeDone(GoalPtr waitee, ExitCode result)
             .outputs = OutputsSpec::Names { outputName },
         });
         if (buildResult.success()) {
-            for (auto & [output, realisation] : buildResult.builtOutputs) {
+            auto i = buildResult.builtOutputs.find(outputName);
+            if (i != buildResult.builtOutputs.end())
                 inputDrvOutputs.insert_or_assign(
-                    { dg->drvPath, output.outputName },
-                    realisation.outPath);
-            }
+                    { dg->drvPath, outputName },
+                    i->second.outPath);
         }
     }
 }
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -145,8 +145,20 @@ void DerivationGoal::work()
 void DerivationGoal::addWantedOutputs(const OutputsSpec & outputs)
 {
     auto newWanted = wantedOutputs.union_(outputs);
-    if (!newWanted.isSubsetOf(wantedOutputs))
-        needRestart = true;
+    switch (needRestart) {
+    case NeedRestartForMoreOutputs::OutputsUnmodifedDontNeed:
+        if (!newWanted.isSubsetOf(wantedOutputs))
+            needRestart = NeedRestartForMoreOutputs::OutputsAddedDoNeed;
+        break;
+    case NeedRestartForMoreOutputs::OutputsAddedDoNeed:
+        /* No need to check whether we added more outputs, because a
+           restart is already queued up. */
+        break;
+    case NeedRestartForMoreOutputs::BuildInProgressWillNotNeed:
+        /* We are already building all outputs, so it doesn't matter if
+           we now want more. */
+        break;
+    };
     wantedOutputs = newWanted;
 }
 
@@ -297,12 +309,29 @@ void DerivationGoal::outputsSubstitutionTried()
         In particular, it may be the case that the hole in the closure is
         an output of the current derivation, which causes a loop if retried.
      */
-    if (nrIncompleteClosure > 0 && nrIncompleteClosure == nrFailed) retrySubstitution = true;
+    {
+        bool substitutionFailed =
+            nrIncompleteClosure > 0 &&
+            nrIncompleteClosure == nrFailed;
+        switch (retrySubstitution) {
+        case RetrySubstitution::NoNeed:
+            if (substitutionFailed)
+                retrySubstitution = RetrySubstitution::YesNeed;
+            break;
+        case RetrySubstitution::YesNeed:
+            // Should not be able to reach this state from here.
+            assert(false);
+            break;
+        case RetrySubstitution::AlreadyRetried:
+            debug("substitution failed again, but we already retried once. Not retrying again.");
+            break;
+        }
+    }
 
     nrFailed = nrNoSubstituters = nrIncompleteClosure = 0;
 
-    if (needRestart) {
-        needRestart = false;
+    if (needRestart == NeedRestartForMoreOutputs::OutputsAddedDoNeed) {
+        needRestart = NeedRestartForMoreOutputs::OutputsUnmodifedDontNeed;
         haveDerivation();
         return;
     }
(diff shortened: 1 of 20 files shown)


Ericson2314:
Failure is spurious.

Ericson2314:
@edolstra For reference, the `wantedOutputs` clearing is something you did in 8d8d47abd2a66898aa5d8999fcd75b29991e529d. Maybe looking at the commit will jog your memory on whether you were thinking of the same issue I was when you wrote that.

edolstra:
> Though the goal had gotten as far as building, so all outputs were
going to be produced, addWantedOutputs no longer knows that and so
the goal is flagged to be restarted.

Why would it be restarted? The only place where we restart is in `DerivationGoal::outputsSubstitutionTried()`. Once the build has actually started, it builds all outputs and it won't be restarted, as far as I can tell.

roberth:
Data model improvement and worker protocol no-ops lgtm.
Just need to look into the `src/libstore/build/entry-points.cc` changes still.

src/libstore/build/entry-points.cc
```diff
@@ -47,31 +47,57 @@ void Store::buildPaths(const std::vector<DerivedPath> & reqs, BuildMode buildMod
...
     std::shared_ptr<Store> evalStore)
 {
     Worker worker(*this, evalStore ? *evalStore : *this);
 
     Goals goals;
-    for (const auto & br : reqs) {
-        std::visit(overloaded {
-            [&](const DerivedPath::Built & bfd) {
-                goals.insert(worker.makeDerivationGoal(bfd.drvPath, bfd.outputs, buildMode));
+    std::vector<std::pair<GoalPtr, const DerivedPath &>> reqs2;
```
roberth, line 58:
```suggestion
    std::vector<std::pair<const DerivedPath &, GoalPtr>> state;
```

 - Flip the pair order because the `DerivationPath` is still the key. (plus index because it's a list )
 - `state` because it's not just the requests anymore. It covers the whole state.

src/libstore/build/entry-points.cc
```diff
@@ -47,31 +47,57 @@ void Store::buildPaths(const std::vector<DerivedPath> & reqs, BuildMode buildMod
...
+            [&](const DerivedPath::Built & bfd) -> GoalPtr {
+                return worker.makeDerivationGoal(bfd.drvPath, bfd.outputs, buildMode);
             },
-            [&](const DerivedPath::Opaque & bo) {
-                goals.insert(worker.makePathSubstitutionGoal(bo.path, buildMode == bmRepair ? Repair : NoRepair));
+            [&](const DerivedPath::Opaque & bo) -> GoalPtr {
+                return worker.makePathSubstitutionGoal(bo.path, buildMode == bmRepair ? Repair : NoRepair);
             },
-        }, br.raw());
+        }, req.raw());
```
roberth, line 68: This visit could be moved into `worker`.
`DerivedPath` is a very fundamental thing, so I think this would be appropriate.

`make` is a lie, but that's out of scope for now.

roberth: `Goal` is also a lie. `MutableWorkUnit`? Not sure if that name helps though. Let's not move the Goal posts for this PR.

src/libstore/build/entry-points.cc
```diff
@@ -47,31 +47,57 @@ void Store::buildPaths(const std::vector<DerivedPath> & reqs, BuildMode buildMod
...
-    for (auto & i : goals)
-        results.push_back(i->buildResult);
+    for (auto & [gp, req] : reqs2) {
+        auto & br = results.emplace_back(KeyedBuildResult {
+            gp->buildResult,
+            /* .path = */ req
+        });
+
+        auto pbp = std::get_if<DerivedPath::Built>(&req);
+        if (!pbp) continue;
```
roberth, line 84: It seems that we *could* return a `BuildResult` for opaque paths too. `Substituted`, `AlreadyValid`, `NoSubstituters` are appropriate statuses.

However, that might be somewhat of a breaking change. Maybe add a new method in a followup?

src/libstore/build/entry-points.cc
```diff
@@ -47,31 +47,57 @@ void Store::buildPaths(const std::vector<DerivedPath> & reqs, BuildMode buildMod
...
+        auto & bp = *pbp;
+
+        /* Because goals are in general shared between derived paths
+           that share the same derivation, we need to filter their
+           results to get back just the results we care about.
+         */
```
roberth, line 90: Aforementioned lies make this non-obvious. Short of that, I think we can simplify this a bit

```
        /* Goals are mutable shared state, that may contain more
           outputs than we requested.
         */
```

ie only the weird part. However, I think this code should be moved.

Part of the problem is the `gp->buildResult` is too easy. How about we isolate the nastiness of mutable state by encapsulating `gp`? Then we can provide a method that requests the right outputs and make this mistake impossible.

roberth (commented):
Suggesting some improvements to `buildPathsWithResults`.
Another issue I had was the heavy abbreviation of local variables, making the code harder to understand. I'm already suggesting some things that can be factored out, so perhaps it won't be as much of a problem anymore.

`req`/`reqs`: this is fine
`g` will only span 3 lines, but why not `goal`?
`reqs2` -> `state` as mentioned
`gp`: only used once, but `goalPtr` is still easier to read
`pbp`: might go away?
`bp`: contains multiple outputs. We've already established that that's an unnecessary grouping in most cases, so we'll probably revisit that anyway.
`bos`: move to goal method, change to `results`.

src/libstore/build/entry-points.cc
Ericson2314: The first commit of https://github.com/NixOS/nix/pull/4628 is I think just what you want! Added it here.

src/libstore/build/entry-points.cc
Ericson2314: Oh we are, that continue is after making a `KeyedBuildResult`.

This is a sign to make the `continue` is too tricky, so I am just doing a regular `if` block on the rest.

src/libstore/build/entry-points.cc
Ericson2314: Done!

src/libstore/build/derivation-goal.cc
```diff
@@ -1452,12 +1483,28 @@ void DerivationGoal::waiteeDone(GoalPtr waitee, ExitCode result)
...
+        if (buildResult.success()) {
+            for (auto & [output, realisation] : buildResult.builtOutputs) {
                 inputDrvOutputs.insert_or_assign(
-                    { bfd->drvPath, output.outputName },
+                    { dg->drvPath, output.outputName },
                     realisation.outPath);
+            }
+        }
+    }
 }
```
Ericson2314, line 1508: This became more awkward after `buildResult` was encapsulated because one cannot easily index `builtOutputs` without knowing the hash modulo.

I think what this means is that
```
DrvPaths builtOutputs;
```
is the wrong field for `BuildResult`; instead of that `std::map<DrvOutput, Realisation>` it should be
```
std::map<std::string, Realisation> builtOutputs;
Hash drvHash;
```
since we are only returning realizations for a single top-level derivation.

Ericson2314: I pushed one more commit which addresses this awkwardness, getting rid of the double `for` loop.

src/libstore/build/derivation-goal.cc
```diff
@@ -1411,7 +1442,7 @@ DrvOutputs DerivationGoal::assertPathValidity()
 
 void DerivationGoal::done(
     BuildResult::Status status,
-    DrvOutputs builtOutputs,
+    std::map<std::string, Realisation> builtOutputs,
```
roberth, line 1445: Isn't this a `SingleDrvOutputs`?

src/libstore/build/derivation-goal.hh
```diff
@@ -293,7 +329,7 @@ struct DerivationGoal : public Goal
 
     void done(
         BuildResult::Status status,
-        DrvOutputs builtOutputs = {},
+        std::map<std::string, Realisation> builtOutputs = {},
```
roberth, line 332: Same for header.

src/libstore/remote-store.cc
```diff
@@ -125,32 +125,54 @@ void write(const Store & store, Sink & out, const DrvOutput & drvOutput)
...
         << res.timesBuilt
         << res.isNonDeterministic
         << res.startTime
         << res.stopTime;
-    worker_proto::write(store, to, res.builtOutputs);
+    DrvOutputs builtOutputs;
+    for (auto & [output, realisation] : res.builtOutputs)
+        builtOutputs.insert_or_assign(realisation.id, realisation);
+    worker_proto::write(store, to, builtOutputs);
 }
```
roberth, line 176: Isn't this a breaking change to the protocol?

Ericson2314: Err my intent was to *not* change the protocol, converting it to the old way. (I should add that the commit message and PR description!)

Ericson2314: The top line is changed in a previous commit; that is the difference between `BuildResult` and `KeyedBuildResult`. The latter is now used everywhere needed to preserve the protocol.

Ericson2314: For what its worth, after #6223 is merged the factored out serialization logic can use protocol versions, and then I can deduplicate this stuff much more aggressively :)

roberth: I didn't really register the other changes to the `write` functions. This makes sense. Does the daemon compatibility test pass?

Ericson2314: Good question. Yes it does!

roberth (approved):

src/libstore/build/derivation-goal.cc
```diff
@@ -297,12 +309,29 @@ void DerivationGoal::outputsSubstitutionTried()
...
+            nrIncompleteClosure > 0 &&
+            nrIncompleteClosure == nrFailed;
+        switch (retrySubstitution) {
+        case RetrySubstitution::NoNeed:
+            if (substitutionFailed)
+                retrySubstitution = RetrySubstitution::YesNeed;
+            break;
+        case RetrySubstitution::YesNeed:
+            // Should not be able to reach this state from here.
+            assert(false);
```
roberth, line 323: Unfortunately, it has been reached
- https://github.com/NixOS/nix/issues/10092

PR #6314, opened 2022-03-25
Only provide builtins is the corresponding experimental feature is enabled

edolstra:
This allows writing fallback code like

```
if builtins ? fetchClosure then
  builtins.fetchClose { ... }
else
  builtins.storePath ...
```


Ericson2314 (approved):

Ma27:
@edolstra @Ericson2314 didn't we have this discussion about experimental settings vs. builtins in https://github.com/NixOS/nix/pull/5301#issuecomment-929161132 already? I still think that this is rather confusing to someone who's e.g. new to Nix.

Issue #6309, opened 2022-03-24
openStore takes too much time


ckiee:
**Describe the bug**

Among others, `nix-build` and `nix-instantiate` call `openStore` to initialize the store. This takes about 30% of the total runtime in `nix-instantiate '<nixpkgs>' -A hello`.

**Steps To Reproduce**

```sh
ckie@cookiemonster ~/git/nixfiles -> nix-shell -p inferno linuxKernel.packages.linux_zen.perf nixUnstable
 
[nix-shell:~/git/nixfiles]$ perf record --call-graph dwarf nix-instantiate '<nixpkgs>' -A hello && perf script | inferno-collapse-perf | inferno-flamegraph > flamegraph.svg
warning: you did not specify '--add-root'; the result might be removed by the garbage collector
/nix/store/0c8fsaaxqdamhk5dfp3xy66jyqv47ndz-hello-2.12.drv
[ perf record: Woken up 98 times to write data ]
[ perf record: Captured and wrote 26.430 MB perf.data (3264 samples) ]
 
[nix-shell:~/git/nixfiles]$ bokup flamegraph.svg
flamegraph.svg
      1,632,701 100%  169.54MB/s    0:00:00 (xfr#1, to-chk=0/1)
https://i.ckie.dev/lZfrtLe.svg
```
([svg](https://i.ckie.dev/lZfrtLe.svg))

**Expected behavior**

I expected most of the runtime to be spent on the actual evaluation in this already-long run:
```
ckie@cookiemonster ~ -> cat /tmp/h
#!/bin/sh
exec nix-instantiate '<nixpkgs>' -A hello
 
ckie@cookiemonster ~ -> hyperfine -- /tmp/h
Benchmark 1: /tmp/h
  Time (mean ± σ):     210.1 ms ±   3.6 ms    [User: 158.8 ms, System: 35.5 ms]
  Range (min … max):   203.3 ms … 216.1 ms    14 runs
```

**Additional context**

I don't use flakes but they are worse assuming the overhead from `rm` is minimal:
```
ckie@cookiemonster ~ -> cat /tmp/a
#!/bin/sh
rm -rf ~/.cache/nix/eval-cache-v2
nix build nixpkgs#hello
ckie@cookiemonster ~ -> hyperfine -- /tmp/a
Benchmark 1: /tmp/a
  Time (mean ± σ):     244.1 ms ±   3.2 ms    [User: 179.8 ms, System: 41.9 ms]
[...]

PR #6323, opened 2022-03-26
`nix eval`: Add option `--read-only`

erikarvstedt:
This option sets `Settings::readOnlyMode = true`.
Previously, eval'ing with `readOnlyMode` was only possible through `nix-instantiate --eval`. `readOnlyMode` is now also accessible to the `nix` command.

(Quick reminder: With `readOnlyMode = false` (the current default in `nix eval`), Nix instantiates each derivation whose output or drv path is referenced in the evaluation output; i.e. its `.drv` closure is copied to the store. This is required to support [Import From Derivation](https://nixos.wiki/wiki/Import_From_Derivation).)

`readOnlyMode` has the following advantages:

- Preventing unneeded writes to the store.

- Most importantly: Eval performance.
  Evaluation speed is a major bottleneck when developing or debugging NixOS system configurations.
  `nix eval`'ing a basic system takes ~40% more time [^1] compared to `nix eval --read-only` or `nix-instantiate --eval` (with warm caches).
  This issue has so far prevented me from moving my systems to flakes.
  (Using the much faster `nix-instantiate --eval` with flakes requires flake-compat and doesn't support pure evaluations.)

#### Todo
An optimized `nix eval` implementation would instantiate derivations only when they are accessed during evaluation (Import From Derivation).
In this case, `--read-only` would be obsolete and it could be removed.
Currently, `nix eval` serves as the `nix` CLI equivalent to `nix-instantiate` due to its implicit instantiation behavior.
To allow switching `nix eval` to a non-instantiating, optimized implementation in the future, this PR should also add explicit support for instantiating drvs to the `nix` CLI.
A possible choice is `nix build --derivation`. `--derivation` is currently a no-op when passed to `nix build` (`SourceExprCommand::operateOn` is unused).

#### Appendix

`nix eval nixpkgs#hello.outPath --apply builtins.unsafeDiscardStringContext` doesn't prevent the `hello` derivation from being instantiated, so this is no feasible workaround.

[^1]: You can verify this by running `./benchmark.sh nix_eval_vs_instantiate_eval` ([nix-benchmark](https://github.com/erikarvstedt/nix-benchmark)).
See also: #6309.

cc: @edolstra

### DIFF
--- a/src/libcmd/command.hh
+++ b/src/libcmd/command.hh
@@ -85,11 +85,12 @@ struct SourceExprCommand : virtual Args, MixFlakeOptions
 {
     std::optional<Path> file;
     std::optional<std::string> expr;
+    bool readOnlyMode = false;
 
     // FIXME: move this; not all commands (e.g. 'nix run') use it.
     OperateOn operateOn = OperateOn::Output;
 
-    SourceExprCommand();
+    SourceExprCommand(bool supportReadOnlyMode = false);
 
     std::vector<std::shared_ptr<Installable>> parseInstallables(
         ref<Store> store, std::vector<std::string> ss);
@@ -128,7 +129,7 @@ struct InstallableCommand : virtual Args, SourceExprCommand
 {
     std::shared_ptr<Installable> installable;
 
-    InstallableCommand();
+    InstallableCommand(bool supportReadOnlyMode = false);
 
     void prepare() override;
 
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -1,3 +1,4 @@
+#include "globals.hh"
 #include "installables.hh"
 #include "command.hh"
 #include "attr-path.hh"
@@ -129,7 +130,7 @@ MixFlakeOptions::MixFlakeOptions()
     });
 }
 
-SourceExprCommand::SourceExprCommand()
+SourceExprCommand::SourceExprCommand(bool supportReadOnlyMode)
 {
     addFlag({
         .longName = "file",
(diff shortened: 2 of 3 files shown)


thufschmitt (commented):
Thanks, I like the big picture :)

I wonder whether `supportReadOnlyMode` is really needed: `nix build --read-only` isn’t something I would run every day, but it doesn’t seem to hurt either.

Actually we could even make `readOnlyMode` a proper setting − which would implicitly make it a cli flag

erikarvstedt:
`supportReadOnlyMode` is only true (and thus activated) for command `eval`.
(`nix build --read-only` makes no sense because building requires instantiation.)

What do you think about using `nix build --derivation` for explicit instantiation?

thufschmitt:
Yes, `nix build` is probably a very bad example because `--read-only` doesn’t make a lot of sense here (although it’s always possible to come-up with some contrived examples where it would be useful). But my point was that it’s not a big deal to have it, while the `supportReadOnlyMode` parameter is imho adding some complexity to the code in a way that doesn’t feel right.

Just for completeness, an extra advantage of `--read-only` that I think you didn’t mention but is somewhat important (at least to me :wink: ) is that it allows working with stores that are actually read-only (as long as you don’t really need to write to it). For example `nix why-depends nixpkgs#hello nixpkgs#glibc --store https://cache.nixos.org` doesn’t work atm, but it could work in read-only mode

edolstra:
> An optimized nix eval implementation would instantiate derivations only when they are accessed during evaluation (Import From Derivation).

Yes, that's exactly the reason why I didn't add a `--read-only` flag to the new CLI. I didn't want to carry over technical debt from the old CLI. Ideally, in the future, we wouldn't materialize `.drv` files at all (since we can use `buildDerivation()` to send an in-memory representation of the derivation to the daemon).

thufschmitt:
> Ideally, in the future, we wouldn't materialize .drv files at all (since we can use buildDerivation() to send an in-memory representation of the derivation to the daemon).

That would be great, but that’s not gonna happen any time soon, is it?

erikarvstedt:
The new CLI is already exposing the concecpt of `.drv` files, e.g. in `nix copy --derivation`.
Note that `--read-only` can be removed in a fully backwards-compatible way by making it a no-op and hiding it from the documentation.

edolstra:
Should there also be a `--no-read-only` for `nix eval`? If I understand it correctly, `nix eval` will now always be be in read-only mode, so IFD won't work anymore without a way to turn it back on.

erikarvstedt:
This PR doesn't change the default behavior. `readOnlyMode` is still [off by default](https://github.com/NixOS/nix/blob/16860a0328094c04c3e877089119a56f9ddfbcd6/src/libcmd/command.hh#L88).

PR #6328, opened 2022-03-28
nix profile install: Don't use queryDerivationOutputMap()

### LINKED ISSUE #6286: nix profile install fails with *output*: No such file or directory
[**Describe the bug**]

Cannot install packages with `nix profile install`

**Steps To Reproduce**

1. `nix profile install "nixpkgs#bash"`
2. See error `error: opening directory '/nix/store/c49i1ggnr5cc8gxmk9xm0cn961z104dn-bash-5.1-p12-dev': No such file or directory`

**Expected behavior**
bash should be added to the profile 

```
nix --version
nix (Nix) 2.7.0
```
**Additional context**

originally reported in https://github.com/NixOS/nixpkgs/issues/164610

edolstra:
Instead get the outputs from `Installable::build()`. This will also allow `nix profile install` to support impure derivations.

Fixes #6286.

### DIFF
--- a/src/libstore/derived-path.hh
+++ b/src/libstore/derived-path.hh
@@ -25,6 +25,9 @@ struct DerivedPathOpaque {
     nlohmann::json toJSON(ref<Store> store) const;
     std::string to_string(const Store & store) const;
     static DerivedPathOpaque parse(const Store & store, std::string_view);
+
+    bool operator < (const DerivedPathOpaque & b) const
+    { return path < b.path; }
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/libstore/derived-path.hh
```diff
@@ -25,6 +25,9 @@ struct DerivedPathOpaque {
     nlohmann::json toJSON(ref<Store> store) const;
     std::string to_string(const Store & store) const;
     static DerivedPathOpaque parse(const Store & store, std::string_view);
+
+    bool operator < (const DerivedPathOpaque & b) const
+    { return path < b.path; }
```
Ericson2314, line 30: It might be good to standardize on @thufschmitt's `GENERATE_CMP` (from https://github.com/NixOS/nix/blob/master/src/libutil/comparator.hh) for these?

Ericson2314: In this case:
```c++
GENERATE_CMP(DerivedPathOpaque, me->path)
```

Artturin:
backport?

PR #6330, opened 2022-03-28
nix {run,shell}: Print a better error message if the store is not local

### LINKED ISSUE #6317: Flakes UX: nix run: unusable with `--store "{remote}"`
**Describe the bug**

I am trying to remove my devenv-driven bash scripts by instead wrapping these scripts into derivations and exposing them via nix run.

What I'd like to do is this:
```shell
nix run '.#flash-blueline-boot' \
  --eval-store auto \
  --store 'ssh-ng://colemickens@aarch64.nixos.community' \
  --override-input mobile-nixos "${HOME}/code/mobile-nixos"
```

But what I get is:
```
ls: cannot access '/nix/store/9cr0zplqxzpwcl6blmlz5dpclh3z3mbh-flash-blueline.sh': No such file or directory
```

Despite:
```
❯ ssh colemickens@aarch64.nixos.community \
  "ls -al /nix/store/9cr0zplqxzpwcl6blmlz5dpclh3z3mbh-flash-blueline.sh"
-r-xr-xr-x 1 root root 292 Jan  1  1970 /nix/store/9cr0zplqxzpwcl6blmlz5dpclh3z3mbh-flash-blueline.sh
```

This mirrors behavior of using `--store` in other scenarios where the remote built store path is not copied back.

With `nix build` I can understand it, but not `nix run`.

#### nix version
```
nix (Nix) 2.8.0pre20220322_d5d4d98
```

edolstra:
Closes #6317

### DIFF
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -38,9 +38,12 @@ void runProgramInStore(ref<Store> store,
        unshare(CLONE_NEWUSER) doesn't work in a multithreaded program
        (which "nix" is), so we exec() a single-threaded helper program
        (chrootHelper() below) to do the work. */
-    auto store2 = store.dynamic_pointer_cast<LocalStore>();
+    auto store2 = store.dynamic_pointer_cast<LocalFSStore>();
 
-    if (store2 && store->storeDir != store2->getRealStoreDir()) {
+    if (!store2)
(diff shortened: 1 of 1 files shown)


thufschmitt (approved):
Technically it’s not really fixing #6317, just making a nicer error when it fails. But that’s definitely better than the previous behavior :)

PR #6333, opened 2022-03-28
Add option `--apply` to all `SourceExprCommand`s.

erikarvstedt:
Previously, `--apply` was supported only by `nix eval`.

Advantages:

- Consistency.
  `nix eval` options that evaluate to a derivation can now be passed to any other nix cmd that accepts installables (like `nix build`, `nix shell`, ...).
  This property was lost when `--apply` got added to `eval` only.

- Unlocks pure, system-invariant build expressions.
  These are useful for quickly noting down modified derivations that work on all systems.
  With `--apply`:
  ```bash
  nix build nixpkgs/30d3d79b7d3607d56546dd2a6b49e156ba0ec634#luajit --apply 'p: p.override { enableGDBJITSupport = true; }'
  ```
  Without `--apply`:
  ```bash
  system=$(nix eval --impure --expr 'builtins.currentSystem')
  nix build --expr "(builtins.getFlake ''github:NixOS/nixpkgs/30d3d79b7d3607d56546dd2a6b49e156ba0ec634'').outputs.legacyPackages.$system.luajit.override { enableGDBJITSupport = true; }"
  ```
  Another example:
  ```
  nix shell nixpkgs#pkgs --apply 'pkgs: with pkgs; enableDebugging hello'
  ```


lilyball:
What's wrong with `--apply`[^1]? The only issue that comes to mind for me is the fact that it applies to all installables instead of having a per-installable function, but `--apply` is still useful for when you really do want to apply the function to all installables, and some per-installable syntax can be added later.

Removing `--apply` on `nix eval` in the name of consistency seems like absolutely the wrong approach unless we have a replacement ready already. Being able to purely evaluate installables and pass them through a function is extremely useful.

I also just filed #6526 asking for exactly what this PR does as I keep running into the problem of "I want to build python with a set of python packages" or "I want to build an overridden derivation as a once-off test". `--apply` would solve this for me (as I generally am not specifying multiple installables) and without it I've had to employ awkward workarounds.

[^1]: Are there minutes for the meeting? https://github.com/orgs/NixOS/projects/10/ has no note for 2022-03-31 Minutes, and https://pad.lassul.us/z9O-24cCQ2mFyGUwgGecxw from the calendar event also has no revision history for a 2022-03-31 meeting.

lilyball:
Also worth pointing out, any solution based on `builtins.getFlake` is going to fail for indirect flake references as that currently requires `--impure` even when the indirect reference resolves to `path:`. It also fails for `.` (or the equivalent absolute path).

I do think we should fix the problem where installables on the command-line can do things that `builtins.getFlake` cannot, but that is somewhat orthogonal to this particular issue and the ability to purely override derivations at the command-line should not be blocked by this.

tomberek:
still interested

SharzyL:
Still eager for this feature

PR #6334, opened 2022-03-30
nix-daemon.service: require mounts for /nix/var/nix/db

grahamc:
Users may want to mount a filesystem just for the Nix database, with
the filesystem's parameters specially tuned for sqlite. For example, on
ZFS you might set the recordsize to 64k after changing the database's
page size to 65536.


grahamc:
Note that system administrators need not change the database itself, and could tune the FS to use 4096 page size / record size.

PR #6337, opened 2022-03-30
libexpr: Throw the correct error in toJSON

### LINKED ISSUE #6336: `builtins.toJSON` exits the repl on error
**Describe the bug**

When `builtins.toJSON` encounters an error in `nix repl` it exits out of the repl. This differs from the behaviour of `builtins.toString`, which shows an error but stays inside the repl.

Example:

```
nix-repl> builtins.toString builtins.toJSON
error: cannot coerce the built-in function 'toJSON' to a string

       at «string»:1:1:

            1| builtins.toString builtins.toJSON
             | ^
            2|

nix-repl> builtins.toJSON builtins.toJSON
error: cannot convert the built-in function 'toJSON' to JSON

       at «string»:1:1:

            1| builtins.toJSON builtins.toJSON
             | ^
            2|
<exits out of repl>
```

**Steps To Reproduce**

1. Enter the repl using `nix repl`
2. Type `builtins.toJSON builtins.toJSON` and press enter
3. The repl exits with exit code 1

**Expected behavior**

Show the error message, but then stay inside the repl like **builtins.toString** does.

`nix-env (Nix) 2.5.0pre20211206_d1aaa7e`

**Additional context**

The problem is that a `nix::BaseError` gets thrown by `e.addTrace(...)` inside `printValueAsJSON(...)`, but it is not caught in `NixRepl::mainLoop(...)` because it shouldn't have been thrown in the first place. We actually want to throw a `TypeError` inside `printValueAsJSON(...)` instead.

I will soon submit a PR fixing this. Just putting this issue out there for visibility in case anyone else stumbles upon this bug.

danpls:
BaseError::addTrace(...) returns a BaseError, but we want to
throw a TypeError instead.

Fixes #6336.

### DIFF
--- a/src/libexpr/value-to-json.cc
+++ b/src/libexpr/value-to-json.cc
@@ -84,7 +84,8 @@ void printValueAsJSON(EvalState & state, bool strict,
                 .msg = hintfmt("cannot convert %1% to JSON", showType(v)),
                 .errPos = v.determinePos(pos)
             });
-            throw e.addTrace(pos, hintfmt("message for the trace"));
+            e.addTrace(pos, hintfmt("message for the trace"));
+            throw e;
     }
 }
 
--- a/src/libutil/error.cc
+++ b/src/libutil/error.cc
@@ -9,10 +9,9 @@ namespace nix {
 
 const std::string nativeSystem = SYSTEM;
 
-BaseError & BaseError::addTrace(std::optional<ErrPos> e, hintformat hint)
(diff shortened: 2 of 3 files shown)


thufschmitt (approved):
Thanks, nice catch :)

I guess it would be safer to make `e.addTrace` not return anything since that subtly changes the behavior of the exception, but that can be kept for later

PR #6338, opened 2022-03-30
add labeler workflow

Artturin:
Automatically add labels to pull requests based on the files they touch.

If a label exists in Nixpkgs then use that instead of making a new one.

https://github.com/actions/labeler

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,6 @@
+"documentation":
+  - doc/manual/*
+  - src/nix/**/*.md
+
+"tests":
+  - tests/**/*
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -1,0 +1,24 @@
+name: "Label PR"
+
+on:
+  pull_request_target:
+    types: [edited, opened, synchronize, reopened]
+
+# WARNING:
+# When extending this action, be aware that $GITHUB_TOKEN allows some write
+# access to the GitHub API. This means that it should not evaluate user input in
+# a way that allows code injection.
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  labels:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'NixOS'
+    steps:
+    - uses: actions/labeler@v4
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        sync-labels: true


.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+# if a label exists in nixpkgs then use that instead of making a new one
+
+"8.has: changelog":
+  - /doc/manual/src/release-notes/*
+
+"8.has: documentation":
```
fricklerhandwerk, line 6: @Artturin Thank you, I think this is the exact right way to improve discoverability of pull requests.

I also think it's generally good to converge on a labeling scheme with Nixpkgs (but not at all costs), and that top comment is valuable.

I don't feel qualified to decide on which labels to use exactly and how to organise them, and in any case would like to know what @edolstra thinks about this topic.

A few questions for me to get a better idea:

Why are the Nixpkgs labels structured this way (number.verb: noun)? Why isn't it just "with documentation", for example? (It doesn't seem right to say "has documentation" because often changes just update docs instead of adding it.)

How do we deal with semantically redundant labels? If we changed "documentation" to "8.has: documentation", this will break existing queries, and we'd have to update contributing guides for example.

In general I think it would be best to add a note to the contributing section in the manual when introducing such a workflow, such that contributors (and maintainers) know what all the colorful, weirdly named labels mean and how to best make use of them.

So, before going ahead I'd like to see more explicit deliberation and ensure it will also be visible for others, i.e., not just those participating in this pull request.

What do you think?

fricklerhandwerk:
- have to discuss the labeling convention in detail, but agree that we want this in principle

-> to review

.github/labeler.yml
Artturin: > Why are the Nixpkgs labels structured this way (number.verb: noun)? Why isn't it just "with documentation", for example?

it's for sorting
nixpkgs has [171 labels](https://github.com/NixOS/nixpkgs/labels)
nix has [37 labels](https://github.com/NixOS/nix/labels) so the numbering might not be necessary

image



> (It doesn't seem right to say "has documentation" because often changes just update docs instead of adding it.)

i think it it alright because even if docs are updated the pr still has docs

> How do we deal with semantically redundant labels? If we changed "documentation" to "8.has: documentation", this will break existing queries, and we'd have to update contributing guides for example.

the documentation label is for issues (nixpkgs has [`9.needs: documentation`](https://github.com/NixOS/nixpkgs/labels?q=documentation))  while ` 8.has: documentation` is for PRs

fricklerhandwerk: @Artturin thanks for the input. Is that documented anywhere else? I think part of this PR should be at least a short blurb in the [maintainers' handbook](https://github.com/NixOS/nix/tree/master/maintainers) or [contribution guide](https://github.com/NixOS/nix/tree/master/doc/manual/src/contributing).

On the [last team meeting](https://discourse.nixos.org/t/2022-12-12-nix-team-meeting-minutes-16/24119/1) we decided not to adopt the new labeling scheme yet, and the general sentiment is to keep the process overhead low. I think we should take another look at this given the new information.

fricklerhandwerk (changes requested):

  - @fricklerhandwerk: should discuss it again together, given new information on the labelling scheme used by Nixpkgs
  - @thufschmitt: this is relevant for maintainers and contributors, and we should delegate to someone to make a proposal giving more context and reasoning

---

@Artturin, since no one on the Nix team currently has much time to actually deal with the subject in depth, but we agree that developing a process involving labels would help us be more productive - would you like and do you have the capacity to help out here?

Here's what this could entail:

The Nixpkgs workflow is not documented anywhere except for your prior comment. Could you solicit input from maintainers and give more details how the labels are used exactly? And maybe gather some intelligence on other projects (on GitHub or similar) and their labelling schemes and workflows? Then make a PR to sketch rationale, alternatives considered, the proposed system, and how maintainers and contributors should make use of it. It should be lightweight to start with, and extensible with growing needs. I suggest the team would try out ideas from the proposal as we go, and merge it into the maintainer handbook or contributor guide once we settle on something that works.

I imagine that using more structured labels *could* help keeping an overview, given the procedure is well-documented and followed regularly. They have the advantage that state changes are logged by GitHub, as opposed to when they are moved around the GH project board where the team currently keeps track of issues and PRs. 

@NixOS/nix-team Until then I suggest we use the existing labels and merge the automation itself.

.github/labeler.yml
Artturin: i found the email thread from 2015 in which the label system was created by @nbp https://releases.nixos.org/nix-dev/2015-November/018714.html

.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+# if a label exists in nixpkgs then use that instead of making a new one
+
+"8.has: changelog":
+  - /doc/manual/src/release-notes/*
+
```
fricklerhandwerk, line 5:
```suggestion
```

Since we currently don't have a dedicated label and try to keep their number down, keep it out for now.

fricklerhandwerk (commented):
@Artturin are you still interested in pursuing this?

Artturin:
suggestions applied

here's the old `labeler.yml` incase its needed
```yml
# if a label exists in nixpkgs then use that instead of making a new one

"8.has: changelog":
  - /doc/manual/src/release-notes/*

"8.has: documentation":
  - doc/manual/*
  - src/nix/**/*.md

# does not exist in nixpkgs
"8.has: test":
  - tests/**/*
```



A another reason to use the same labels as nixpkgs (for issues at least) is that moving a issue uses the same labels when moved from nix to nixpkgs and vice versa
see https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Abug which are all moved from nix to nixpkgs

fricklerhandwerk (approved):

.github/workflows/labels.yml
```diff
@@ -0,0 +1,24 @@
...
+# a way that allows code injection.
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  labels:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'nix'
```
Ericson2314, line 19: based on https://github.com/NixOS/nixpkgs/blob/master/.github/workflows/labels.yml I think:
```suggestion
    if: github.repository_owner == 'NixOS'
```

Ericson2314:
We discussed this in the Nix team meeting today and agreed we want to do this. **Decision: Merge**.

We were going to do so after this CI failure went away, but I just noticed what I think is a typo. @Artturin can you fix this?

PR #6343, opened 2022-03-31
scripts/install-systemd-multi-user.sh: fix another typo

Artturin:
https://github.com/NixOS/nix/blob/master/.github/workflows/ci.yml#L62 should really be enabled for PRs instead of just pushes


thufschmitt (approved):

thufschmitt:
> https://github.com/NixOS/nix/blob/master/.github/workflows/ci.yml#L62 should really be enabled for PRs instead of just pushes

Doesn’t that require the origin repo to have cachix set up?

Artturin:
lets do the same thing as in nixpkgs
https://github.com/NixOS/nixpkgs/blob/master/.github/workflows/basic-eval.yml

perhaps a new job which only runs on PRs which uses a different cache from the one that runs on merged PRs

PR #6344, opened 2022-03-31
profile!: consistent use of url/uri. create new version

tomberek:
Bump version number for manifest.json in profiles. Use url and originalUrl as key names.

### DIFF
--- a/src/nix/profile.cc
+++ b/src/nix/profile.cc
@@ -97,19 +97,30 @@ struct ProfileManifest
             auto json = nlohmann::json::parse(readFile(manifestPath));
 
             auto version = json.value("version", 0);
-            if (version != 1)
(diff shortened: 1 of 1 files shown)


dtzWill:
Thanks! The version bump means older nix won't work with these new profiles (which are upgraded on write), which makes sense since we're changing the keys, but as a result of the bump will now print an error even on older nix making this clear (profile manifest has unsupported version) vs the mysterious json error from before (right)?

Sounds good to me, thanks!
(and glad the other uri/url is being changed too :))

dtzWill:
Haha, well FWIW anyone who has a profile with the intermediate profile format of originalUrl / uri (and version=1), as it was right before this PR, should probably recreate their profile or something, because now neither stable nor master work with it :).  All good for using unstable nix.

PR #6346, opened 2022-03-31
Get rid of `impureOutputHash`; fix possible bug

Ericson2314:
# Motivation

I do not believe there is any problem with computing `hashDerivationModulo` the normal way with impure derivations.

Conversely, the way this used to work is very suspicious because two almost-equal derivations that only differ in depending on different impure derivations could have the same drv hash modulo. That is very suspicious because there is no reason to think those two different impure derivations will end up producing the same content-addressed data!

# Context

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -185,41 +185,44 @@ Goal::Co DerivationGoal::haveDerivation()
     if (!drv->type().hasKnownOutputPaths())
         experimentalFeatureSettings.require(Xp::CaDerivations);
 
-    if (drv->type().isImpure()) {
-        experimentalFeatureSettings.require(Xp::ImpureDerivations);
-
-        for (auto & [outputName, output] : drv->outputs) {
-            auto randomPath = StorePath::random(outputPathName(drv->name, outputName));
-            assert(!worker.store.isValidPath(randomPath));
-            initialOutputs.insert({
-                outputName,
-                InitialOutput {
-                    .wanted = true,
-                    .outputHash = impureOutputHash,
-                    .known = InitialOutputStatus {
-                        .path = randomPath,
-                        .status = PathStatus::Absent
-                    }
-                }
-            });
-        }
-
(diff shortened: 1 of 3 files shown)


edolstra:
> The point of making DerivationType::ContentAddressed a record (and derivation type no longer an enum) was that we already handled this case and didn't need a new variant.

I don't agree that this already handles that case. An impure derivation is not conceptually the same as a derivation that isn't sandboxed *and* not fixed-output. You could imagine derivations that are not sandboxed and not fixed-output, but are still expected to have a reproducible result (e.g. a derivation that uses ccache to build something remotely).

Ericson2314:
We conversely, one might want to cache actually impure derivations more normally, in a "per transaction" substitution map.

Still, I will split this up so the second commit doesn't hold back the first.

src/libstore/build/derivation-goal.cc
```diff
@@ -246,6 +223,15 @@ void DerivationGoal::haveDerivation()
             }
         });
 
+    if (!drv->type().isPure()) {
+        settings.requireExperimentalFeature(Xp::ImpureDerivations);
+
+        /* We don't yet have any safe way to cache and impure derivation at
```
zseri, line 229:
```suggestion
        /* We don't yet have any safe way to cache an impure derivation at
```

Ericson2314:
In the meeting today @edolstra mentioned that, as we discussed in the past, we don't necessary want to make impure derivations work "just like" CA derivations.

A thing I forgot to point out was that *floating* CA derivations are *already* not handled specially. The only special case in `hashDerivationModulo` is for *fixed* CA derivations -- input address and floating CA are handled the same way.

Impure derivations are certainly not fixed output, so I don't think we are going to run into any problems here.

tomberek:
- Needs update.
- @Ericson2314 is this still an issue?

edolstra (approved):

Issue #6083, opened 2022-02-11
nix-shell as root cd's to /


dermetfan:
## Describe the bug

Running `nix-shell` or `nix develop` or `nix shell` as root drops you into `/` instead of staying in the current directory.

### `nix shell` example

```sh
sudo nix shell nixpkgs#hello
```

### `nix-shell` example

```nix
# shell.nix
with import <nixpkgs> {};
mkShell {}
```

```sh
sudo nix-shell
```

### `nix develop` example

```nix
# flake.nix
{
  outputs = { nixpkgs, ... }: {
    devShell.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.mkShell {};
  };
}
```

```sh
sudo nix develop
```

## Expected behavior

Drops you into a shell in the current directory.

## Additional context

`nix (Nix) 2.7.0pre20220210_5b809f9`

First noticed on 52f52319ad21bdbd7a33bb85eccc83756648f110.

We [worked around it](https://github.com/input-output-hk/cicero/blob/9cad514a6c0ad266e3bc089f55c2621583e40b8c/pkgs/cicero/evaluators/nix/lib.nix#L257-L259
) for now using `env -C "$PWD"`.

disassembler:
`nix run` also executes in context of `/` so relative paths passed as arguments no longer works.

cole-h:
I did a small bit of testing, and found that it was introduced between 2.4 and 2.5.0. I then bisected it to fcb8af550f5fca37458da0d9042a2b59523eb304 (by testing `nix build && doas result/bin/nix-shell -p`):

<details><summary>git bisect log</summary>

```
git bisect start
# good: [9d5c985aedcf7d11a5310eae6f27dcde8eb7f608] Set 2.4 release date
git bisect good 9d5c985aedcf7d11a5310eae6f27dcde8eb7f608
# bad: [a646dfcdd642df404ef1e172abdcddb3aa179d48] Set version
git bisect bad a646dfcdd642df404ef1e172abdcddb3aa179d48
# good: [5a0c8c6712ee31d85ac77b94707d3510f9383efa] Merge pull request #5348 from edolstra/chroot-addpath
git bisect good 5a0c8c6712ee31d85ac77b94707d3510f9383efa
# bad: [51ffc19f02e78d7bea31c2916bc18798183f9ca1] Merge branch 'add-docker-image-to-hydra-jobs' of https://github.com/garbas/nix
[...]

PR #6348, opened 2022-04-01
libutil: Fix restoring mount namespace

cole-h:
Fixes https://github.com/NixOS/nix/issues/6083.

---

Authored by @aszlig in https://github.com/NixOS/nix/issues/6083. I'd like to see this fixed ASAP, so I'm submitting the PR and will take on the burden of resolving discussions and issues :)

Since aszlig also mentioned it may not be desirable to depend on `std::filesystem`, I looked around the codebase and reworked some of the `std::filesystem`-dependent stuff to use the "typical" `getcwd` and `chdir` functions in 7f5caaa7c0f151520d05d4662415ac09d4cf34b0. I'd honestly prefer to use aszlig's `std::filesystem` variant because it is much cleaner, but the choice is yours.

---

Also, re: aszlig's commit message:

```
The reason for the latter is because running Nix as root means that we
can directly access the store, which makes sure we use a filesystem
namespace to make the store writable. XXX - REWORD!
```

I'm reasonably convinced this is true, but would like somebody to confirm before I remove the `XXX - REWORD!` remark (which is just because [aszlig wasn't sure if it was true](https://github.com/NixOS/nix/issues/6083#issuecomment-1086095212)).

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -10,7 +10,6 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <filesystem>
 #include <future>
 #include <iostream>
 #include <mutex>
@@ -1691,7 +1690,7 @@ void setStackSize(size_t stackSize)
 
 #if __linux__
 static AutoCloseFD fdSavedMountNamespace;
-std::optional<std::filesystem::path> savedCwd;
+std::optional<Path> savedCwd;
 #endif
 
 void saveMountNamespace()
@@ -1699,7 +1698,11 @@ void saveMountNamespace()
 #if __linux__
     static std::once_flag done;
     std::call_once(done, []() {
-        savedCwd.emplace(std::filesystem::current_path());
+        char* cwd = getcwd(NULL, 0);
+        if (cwd == NULL) throw SysError("getting cwd");
+        savedCwd.emplace(cwd);
+        free(cwd);
+
         AutoCloseFD fd = open("/proc/self/ns/mnt", O_RDONLY);
         if (!fd)
             throw SysError("saving parent mount namespace");
@@ -1717,11 +1720,8 @@ void restoreMountNamespace()
     } catch (Error & e) {
         debug(e.msg());
     }
-    try {
-        if (savedCwd)
-            std::filesystem::current_path(*savedCwd);
-    } catch (std::filesystem::filesystem_error const &e) {
-        debug(e.what());
+    if (savedCwd && chdir(savedCwd->c_str()) == -1) {
+        throw SysError("restoring cwd");
     }
 #endif
 }
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -1690,7 +1690,6 @@ void setStackSize(size_t stackSize)
 
 #if __linux__
 static AutoCloseFD fdSavedMountNamespace;
-static AutoCloseFD fdSavedCwd;
 #endif
 
 void saveMountNamespace()
@@ -1702,11 +1701,6 @@ void saveMountNamespace()
         if (!fd)
             throw SysError("saving parent mount namespace");
         fdSavedMountNamespace = std::move(fd);
-
-        fd = open("/proc/self/cwd", O_RDONLY);
-        if (!fd)
-            throw SysError("saving cwd");
-        fdSavedCwd = std::move(fd);
     });
 #endif
 }
@@ -1715,6 +1709,10 @@ void restoreMountNamespace()
 {
 #if __linux__
     try {
+        AutoCloseFD fdSavedCwd = open("/proc/self/cwd", O_RDONLY);
+        if (!fdSavedCwd) {
+            throw SysError("saving cwd");
+        }
         if (fdSavedMountNamespace && setns(fdSavedMountNamespace.get(), CLONE_NEWNS) == -1)
             throw SysError("restoring parent mount namespace");
         if (fdSavedCwd && fchdir(fdSavedCwd.get()) == -1) {
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -1698,20 +1698,7 @@ void saveMountNamespace()
 #if __linux__
     static std::once_flag done;
     std::call_once(done, []() {
-#ifdef __GNU__
-        // getcwd allocating its return value is a GNU extension.
-        char *cwd = getcwd(NULL, 0);
-        if (cwd == NULL)
-#else
-        char cwd[PATH_MAX];
-        if (!getcwd(cwd, sizeof(cwd)))
-#endif
-            throw SysError("getting cwd");
-        savedCwd.emplace(cwd);
-#ifdef __GNU__
-        free(cwd);
-#endif
-
+        savedCwd = absPath(".");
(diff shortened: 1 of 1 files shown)


src/libutil/util.cc
```diff
@@ -1688,13 +1688,21 @@ void setStackSize(size_t stackSize)
...
 static AutoCloseFD fdSavedMountNamespace;
+std::optional<Path> savedCwd;
+#endif
 
 void saveMountNamespace()
 {
 #if __linux__
     static std::once_flag done;
     std::call_once(done, []() {
+        char* cwd = getcwd(NULL, 0);
```
pennae, line 1701: `getcwd` allocating its return value is a GNU extension, so maybe we shouldn't rely on that working

cole-h: How's [`2a45cf5` (#6348)](https://github.com/NixOS/nix/pull/6348/commits/2a45cf54e4201a894254604676dcb51f4c1a471c) look to you?

pennae: would honestly prefer to not use the gnu extension at all, or define(/check?) that nix only supports glibc and musl (both of which can allocate in `getcwd`). doing both can cause (unlikely) odd failures where glibc can keep a very long pwd but other libcs can't :slightly_frowning_face: we can probably just use `std::filesystem`, it's included in at least one place and toml11 seems to use it as well.

cole-h: I'm in the same boat -- I'd prefer `std::filesystem` as well,  but I vaguely recall there being issues in the past. That's why I added the non-`std` stuff on top: to hopefully be in a mergeable state sooner if that was still the case. If `std::filesystem` *is* fine to use, then it's extremely easy to drop my commits and continue off of aszlig's work.

src/libutil/util.cc
```diff
@@ -1688,13 +1688,30 @@ void setStackSize(size_t stackSize)
...
+        if (cwd == NULL)
+#else
+        char cwd[PATH_MAX];
+        if (!getcwd(cwd, sizeof(cwd)))
+#endif
+            throw SysError("getting cwd");
+        savedCwd.emplace(cwd);
+#ifdef __GNU__
+        free(cwd);
+#endif
```
edolstra, line 1713:
```suggestion
        savedCwd = absPath(".");
```

src/libutil/util.cc
```diff
@@ -1712,6 +1729,9 @@ void restoreMountNamespace()
     } catch (Error & e) {
         debug(e.msg());
     }
+    if (savedCwd && chdir(savedCwd->c_str()) == -1) {
```
edolstra, line 1732: Rather than saving the current directory in `saveMountNamespace()`, it may be better to save a file descriptor to the current directory just before calling `setns()`, and then use `fchdir()` to restore it.

cole-h: The "just before calling `setns()`" part is tripping me up -- should I not just move `savedCwd` to an `AutoCloseFD` and modify the existing logic to reflect that, or does it _also_ need to be moved prior to the `setns()` call in `restoreMountNamespace()`? (Pardon my ignorance, but moving it into `restoreMountNamespace` just seems... wrong?)

cole-h: Switched to using an `fd` in [`e135d22` (#6348)](https://github.com/NixOS/nix/pull/6348/commits/e135d223f66f77f2b03a11b979d714e582364013).

Ericson2314:
Agree with @edolstra the FD method here is very nice!

cole-h:
Maybe I'm doing something wrong, but I was unable to get the FD method of saving the CWD to work after moving the logic into `restoreMountNamespace()` -- see the commit message of [`10b9c1b` (#6348)](https://github.com/NixOS/nix/pull/6348/commits/10b9c1b2b269b96dcb8b3d298491fa143d1663e8) for the issue I was facing (and the commit itself for the implementation).

I switched back to `absPath(".")` in [`56009b2` (#6348)](https://github.com/NixOS/nix/pull/6348/commits/56009b2639dc878be11f94d096fae56ac14dcd1d), which does work for both of my previously tested cases.

cole-h:
Did a bit of debugging on the FD issue and I came across this oddity:

```
1720            if (fdSavedMountNamespace && setns(fdSavedMountNamespace.get(), CLONE_NEWNS) == -1)
(gdb) info proc
process 2494958
cmdline = '/home/vin/workspace/vcs/nix/outputs/out/bin/nix --experimental-features nix-command flakes develop -c pwd'
cwd = '/home/vin/workspace/vcs/nix'
exe = '/home/vin/workspace/vcs/nix/outputs/out/bin/nix'
(gdb) n
1725            if (fd && fchdir(fd.get()) == -1) {
(gdb) info proc
process 2494958
cmdline = '/home/vin/workspace/vcs/nix/outputs/out/bin/nix --experimental-features nix-command flakes develop -c pwd'
cwd = '/'
exe = '/home/vin/workspace/vcs/nix/outputs/out/bin/nix'
(gdb) n
1711            Path cwd = absPath(".");
(gdb) info proc
process 2494958
cmdline = '/home/vin/workspace/vcs/nix/outputs/out/bin/nix --experimental-features nix-command flakes develop -c pwd'
cwd = '/workspace/vcs/nix'
exe = '/home/vin/workspace/vcs/nix/outputs/out/bin/nix'
```

`cwd` starts correct, then we `setns()` which change the `cwd` to `/` (as you expected), then after trying to "restore" the `cwd` from the stored `fd`, it loses `/home/vin` and becomes `/workspace/vcs/nix`. A bit out of my depth, so I have no idea how to fix this (aside from just using the `absPath(".")` workaround).

Ericson2314:
@cole-h can you say what both namespace roots are? I hope it's not a Linux bug...

cole-h:
How should I check that? Is it just under `/proc/pid/ns/` somewhere?

Ericson2314:
Not really sure. Maybe just breakpoint the libc functions and write down the args?

Ericson2314:
Or strace and grep. Strace following forks is much less annoying than GDB following forks.

pennae:
fchdir does *not* care about namespace boundaries. it drops you into the directory you got the fd for, not into an equivalent path of the current mount namespace. (this mirrors many other instances in linux where a file descriptor confers a *capability*, not a *link*)

Ericson2314:
@pennae that is certainly how I hope it should work!

cole-h:
I hacked together a little test program to confirm it's not a Linux bug, and it works as I'd expect (https://gist.github.com/cole-h/5b83465b840dac6fec63ef0a4a067cc9 -- pardon the ugly C).

I don't know how to debug this weird behavior with `setns` and `fchdir`, but at least it works with the current head of this PR...

Ericson2314:
@cole-h mind making a draft PR undoing that commit? It might be nice having that + issue for someone to return to later. I am very keen on using being able to use file descriptors as much as possible, in order for some grand Nix :handshake: capsicum future.

cole-h:
Sure, I can do that tomorrow. I'll make a note to myself.

Ericson2314:
Thanks!! I appreciate it.

cole-h:
Opened https://github.com/NixOS/nix/pull/6386 as a draft.

Issue #6170, opened 2022-02-26
Flake: nixConfig options don't get used by nix run


Kranzes:
**Describe the bug**

When using `nix run`, the command does not use the `nixConfig` options declared in the flake.nix. `nix build` does read it accordingly.

For example when using `nixConfig.extra-substituters` with `nix run` the build process won't use the binary cache, but when using `nix build` it does.

**Steps To Reproduce**

1. Execute `nix run github:AmitHalpert/contamination` 
2. ^ binary cache is **not** used ^
3. Execute `nix run github:AmitHalpert/contamination \
--option extra-substituters 'https://contamination.cachix.org/' \
--option extra-trusted-public-keys 'contamination.cachix.org-1:KmdW5xVF8ccKEb9tvK6qtEMW+lGa83seGgFyBOkeM/4='`
4. ^ binary **is** being used ^
5. Execute `nix build github:AmitHalpert/contamination` (binary cache is used and the options are prompted when running)

**Expected behavior**
When executing `nix run` on a flake with the `nixConfig` options set I expect `nix run` to read those options and use them accordingly.

**`nix-env --version` output**
`nix-env (Nix) 2.7.0pre20220127_558c4ee`

tomberek:
perhaps: https://github.com/flox/nix/commit/48483e7154462ee82e91e7059d48e094c9f69255 ?
I'm not sure what else might be impacted (or "double-locked") by adding that unconditionally there

thufschmitt:
The issue seems to be that `lockFlags.applyNixConfig` is false by default. Most commands overrides it − but `nix run` doesn’t 

So I think there’s two possibilities out of this:

1. Change the definition of `lockFlags` in [flake.cc](https://github.com/NixOS/nix/blob/c9a29d0d92529338026190374d0e7e886f486e47/src/nix/flake.cc#L672) to have `.applyNixConfig = true` since that seems to be the most sensible default. But maybe there are some catches to be aware of (thinking of completion in particular. I wouldn’t want `nix build .#<TAB>` to try and apply the nix config)
2. Just set `lockFlags.applyNixConfig = true;` in `nix run`’s `run()` function like it’s done for other commands

edrex:
Are there any other nix subcommands that will need to `applyNixConfig`?

PR #6350, opened 2022-04-02
fix(run): set applyNixConfig lockFlag

tomberek:
Fixes: #6170 

Per @thufschmitt 's comment: https://github.com/NixOS/nix/issues/6170#issuecomment-1085685028

### DIFF
--- a/src/nix/run.cc
+++ b/src/nix/run.cc
@@ -182,6 +182,7 @@ struct CmdRun : InstallableCommand
     {
         auto state = getEvalState();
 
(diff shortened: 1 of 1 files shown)


edrex (commented):
Tested, seems to work. Change seems straightforward and was recommended by @thufschmitt (but I don't know the codebase so can't evaluate possible side effects). +1

edolstra:
I agree with the comment https://github.com/NixOS/nix/issues/6170#issuecomment-1085685028 that it seems better to just enable `applyNixConfig` by default.

PR #6355, opened 2022-04-03
src/libstore/globals.hh: documentation: no root needed if userns

a-m-joseph:
On my Linux system with CONFIG_USER_NS=y and
/proc/sys/user/max_user_namespaces > 0, Nix is definitely doing
sandboxing.  I don't believe that giving root access to Nix is
required in order to get sandboxing in this case.

On the other hand, with the rate at which LPE exploits for
CONFIG_USER_NS are appearing it's not really clear that there is any
practical difference.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -406,10 +406,11 @@ public:
           not run in private network namespace to ensure they can access the
           network).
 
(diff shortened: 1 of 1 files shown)


a-m-joseph:
Sorry, apparently github automatically closes a PR if you rename its branch. They should warn people.

PR #6356, opened 2022-04-03
src/libstore/globals.hh: add option log-format

a-m-joseph:
This allows users to get the old `nix-build` behavior without having
to remember to pass the `-L` flag on every invocation of `nix build`.


a-m-joseph:
I have to say that I really like the new `bar-with-logs` format.  Especially the light-gray "package>" prefix on each line so I can tell where that log line came from.

a-m-joseph:
Sorry, apparently github automatically closes a PR if you rename its branch.  They should warn people.

PR #6357, opened 2022-04-04
src/libstore/globals.hh: add option log-format

a-m-joseph:
This allows users to get the old `nix-build` behavior without having
to remember to pass the `-L` flag on every invocation of `nix build`.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -120,7 +120,7 @@ public:
         R"(
           The default log format to display when buuilding; options are `raw`,
           `raw-with-logs`, `internal-json`, `bar`, and `bar-with-logs`.  `bar`
-          refers to the progress bar showing the build progress."
+          refers to the progress bar showing the build progress.
         )"};
 
     MaxBuildJobsSetting maxBuildJobs{
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -119,7 +119,7 @@ public:
     Setting<std::string> logFormat{this, "bar", "log-format",
         R"(
           The default log format to display when buuilding; options are `raw`,
-          `raw-with-logs`, `internal-json`, `bar`, and `bar-with-logs`.  'Bar'
+          `raw-with-logs`, `internal-json`, `bar`, and `bar-with-logs`.  `bar`
           refers to the progress bar showing the build progress."
         )"};
 
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -116,6 +116,13 @@ public:
         "The number of lines of the tail of "
         "the log to show if a build fails."};
 
+    Setting<std::string> logFormat{this, "bar", "log-format",
+        R"(
+          The default log format to display when buuilding; options are `raw`,
+          `raw-with-logs`, `internal-json`, `bar`, and `bar-with-logs`.  'Bar'
+          refers to the progress bar showing the build progress."
+        )"};
+
     MaxBuildJobsSetting maxBuildJobs{
         this, 1, "max-jobs",
         R"(
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -271,7 +271,7 @@ void mainWrapped(int argc, char * * argv)
 
     evalSettings.pureEval = true;
 
-    setLogFormat("bar");
+    setLogFormat(settings.logFormat);
     settings.verboseBuild = false;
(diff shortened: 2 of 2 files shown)


src/libstore/globals.hh
```diff
@@ -116,6 +116,13 @@ public:
         "The number of lines of the tail of "
         "the log to show if a build fails."};
 
+    Setting<std::string> logFormat{this, "bar", "log-format",
+        R"(
+          The default log format to display when buuilding; options are `raw`,
+          `raw-with-logs`, `internal-json`, `bar`, and `bar-with-logs`.  'Bar'
```
edolstra, line 122:
```suggestion
          `raw-with-logs`, `internal-json`, `bar`, and `bar-with-logs`.  `bar`
```

src/libstore/globals.hh
```diff
@@ -116,6 +116,13 @@ public:
...
+          refers to the progress bar showing the build progress."
```
edolstra, line 123: Maybe it's better to just document every logger instead of just `bar` in a list.

edolstra, line 123:
```suggestion
          refers to the progress bar showing the build progress.
```

edolstra (commented):
Shouldn't the `--log-format` option in src/libmain/common-args.cc be removed? It should be superfluous now.

Note that *not* having this as an option was intentional. It seems dangerous (e.g. for scripts) to have config settings that globally change the output format of a command.

src/libstore/globals.hh
a-m-joseph: > Shouldn't the `--log-format` option in src/libmain/common-args.cc be removed?

Sure, I will update this PR to remove it.

> Note that _not_ having this as an option was intentional. It seems dangerous (e.g. for scripts) to have config settings that globally change the output format of a command.

The output of `nix build` is on stderr.  I think it's sort of a convention that the stderr from any program is for humans, and you should not assume the ability to parse it.

[`NO_COLOR`](https://no-color.org/) is widely implemented and has this property.

a-m-joseph: Another possibility -- which I don't recommend -- would be to ignore `log-format` if [`isatty(3)`](https://man7.org/linux/man-pages/man3/isatty.3.html) reports that stderr is _not_ a tty.  This would give deterministic behavior for scripts running unattended.

Personally I'm not a big fan of programs that change their behavior just because I piped their output through `grep`.

a-m-joseph:
> You might also be interested in some of the discussion and work done in #3961.

> I refactored the loggers stuff to be pluggable (letting us drop the libmain <- libstore dependency).

Oh my, that's ambitious!  I just want to be able to move to `nix3` without losing the old `nix-build` behavior.  I hope it's okay for this PR to be unambitious.

cole-h:
That's how that PR started as well... ^^;

a-m-joseph:
Okay, I looked at this more closely.

The new `--log-format internal-json` is a very cool feature!  However nix clearly "thinks of" these json lines as being another style of log sent to stderr, which is troubling.  I think this will create problems in the long run: nix is encouraging other processes to parse its stderr, and doing so itself.  The "don't parse stderr" convention exists to ensure that software always has an output channel whose format is *never* stable, and which can be changed at any time to add more information or better error messages.

@edolstra, is there any possibility of reconsidering this policy decision to have nix's parent process parse its stderr?  If you are determined to send the json to stderr and cannot be talked out of it, then there does not seem to be any safe way to implement `--option log-format` and I should close this PR.

If this is something you are willing to consider then I would like to prepare, test, and submit a PR such that:

* a new option `--json` (which deliberately has no `--option` equivalent) is added.

* With `--json`, nix will close stderr immediately after parsing `argv` and will emit json events on its **stdout**.

* Without `--json`, nix will close stdout immediately after parsing `argv` and will log on **stderr** according to `log-format` (from `--option` or `nix.conf`).

edolstra:
>  I think this will create problems in the long run: nix is encouraging other processes to parse its stderr

I don't think this is a problem, since you have to ask for this. I.e. stderr only has a parseable format iff you set the logger to `internal-json`.

However, we can get into trouble if we call child processes (like `git`) that pollute stderr, so it's a good idea to provide an option to send the JSON log output to another file descriptor. We already use the environment variable `NIX_LOG_FD` inside builders, so we could honor this variable for Nix itself as well.

PR #6358, opened 2022-04-04
src/libstore/globals.hh: documentation: no root needed if userns

a-m-joseph:
On my Linux system with CONFIG_USER_NS=y and
/proc/sys/user/max_user_namespaces > 0, Nix is definitely doing
sandboxing.  I don't believe that giving root access to Nix is
required in order to get sandboxing in this case.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -406,10 +406,11 @@ public:
           not run in private network namespace to ensure they can access the
           network).
 
-          Currently, sandboxing only work on Linux and macOS. The use of a
-          sandbox requires that Nix is run as root (so you should use the
-          “build users” feature to perform the actual builds under different
-          users than root).
+          Currently, sandboxing only work on Linux and macOS. The use
+          of a sandbox requires that your system supports "user
+          namespaces" or else that Nix is run as root (so you should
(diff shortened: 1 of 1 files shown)


edolstra:
It does some sandboxing, but it can't be relied on a security feature since it doesn't run builds under a different uid.

a-m-joseph:
Thanks, I did not realize that nix was not using `CLONE_NEWUSER`.  I will revise this.

a-m-joseph:
So, the root reason for this PR was that I noticed that the result of some builds depended on whether sandboxing was enabled (see #166879).  Note that this involved derivations which would succeed _whether or not sandboxing was enabled_ -- but produce different results.  A real headache.

I assumed that sandboxing was an all-or-nothing bit that is either set or not.  Since that bit affects the results of a build but does not affect the derivation, I wanted to document as clearly as possible the conditions under which the bit gets set/cleared.

After digging more deeply into this, nix's sandboxing isn't just a bit.  There appear to be multiple levels of sandboxing, and conceivably a build could behave differently at each of those levels.

So, the problem isn't really a documentation problem.  Or at least it can't be dealt with using documentation changes.

Issue #6342, opened 2022-03-31
nix doctor: no output


davidak:
**Describe the bug**

When i run `nix doctor` i get no output.

Related to https://github.com/NixOS/nix/issues/3084 https://github.com/NixOS/nix/pull/3102

**Steps To Reproduce**

1. `nix run nixos-unstable.nix`
4. `nix doctor`

**Expected behavior**

Get the promised log output of the checks results.

I get that with `-v`:

```
[davidak@nixos:~]$ nix doctor -v
Running checks against store uri: daemon
[FAIL] Multiple versions of nix found in PATH:
  /nix/store/7fgvivmrhmimsba56p72sssn5qsdw00d-nix-2.3.16/bin
  /nix/store/h538v75x5xgwzy31s5ia3ah7k6cdyzmy-nix-2.7.0/bin

[PASS] All profiles are gcroots.
[FAIL] Warning: protocol version of this client does not match the store.
While this is not necessarily a problem it's recommended to keep the client in
sync with the daemon.

Client protocol: 1.33
Store protocol: 1.21
```

**`nix-env --version` output**

nix (Nix) 2.7.0

PR #6362, opened 2022-04-05
doctor: Always show the output

thufschmitt:
Fix https://github.com/NixOS/nix/issues/6342

### DIFF
--- a/src/nix/doctor.cc
+++ b/src/nix/doctor.cc
@@ -24,12 +24,12 @@ std::string formatProtocol(unsigned int proto)
 }
 
 bool checkPass(const std::string & msg) {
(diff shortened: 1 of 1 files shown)


src/nix/doctor.cc
```diff
@@ -24,12 +24,12 @@ std::string formatProtocol(unsigned int proto)
 }
 
 bool checkPass(const std::string & msg) {
-    logger->log(ANSI_GREEN "[PASS] " ANSI_NORMAL + msg);
+    logger->log(lvlNotice, ANSI_GREEN "[PASS] " ANSI_NORMAL + msg);
```
edolstra, line 27: We can use `notice()` here right?

thufschmitt: Indeed, thanks. Fixed

PR #6374, opened 2022-04-06
libfetchers: Fix assertion

### LINKED ISSUE #6195: "nix build" aborts with failed assertion when path contains "./"
"nix build" fails when given a path like "git+file:///tmp/./nixtest/". The unnecessary "./" seems to be the reason and causes it to abort.

$ nix --version
nix (Nix) 2.7.0pre20220302_1aca6b9

To reproduce:

$ mkdir /tmp/nixtest
$ cd /tmp/nixtest
$ git init .
$ nix build git+file:///tmp/./nixtest/
warning: Git tree '/tmp/./nixtest/' is dirty
nix: src/libfetchers/git.cc:265: nix::fetchers::GitInputScheme::fetch(nix::ref<nix::Store>, const nix::fetchers::Input&)::<lambda(const Path&)>: Assertion `hasPrefix(p, actualUrl)' failed.
Aborted (core dumped)

danpls:
The filter expects all paths to have a prefix of the raw `actualUrl`, but
`Store::addToStore(...)` provides absolute canonicalized paths.
To fix this create an absolute and canonicalized path from the `actualUrl` and
use it instead.

Fixes #6195.

### DIFF
--- a/tests/fetchGit.sh
+++ b/tests/fetchGit.sh
@@ -7,7 +7,9 @@ fi
 
 clearStore
 
-repo=$TEST_ROOT/git
+# Intentionally not in a canonical form
+# See https://github.com/NixOS/nix/issues/6195
+repo=$TEST_ROOT/./git
 
 export _NIX_FORCE_HTTP=1
 


thufschmitt (approved):
Thanks, looks good :)

I’ve just added a quick test for it, will merge as soon as the CI is green

(Interestingly I noted that `/foo//bar` isn’t recognized as a valid url by the Nix parser, which is also probably a bug)

danpls:
Yeah, you're right. I didn't even notice this one.
It seems that the regex is faulty:

https://github.com/NixOS/nix/blob/8b1e328d5d0ae7d3a4a8f6012ec065b59674ed4a/src/libutil/url.cc#L14-L19

You could easily allow empty path segments like in `/foo//bar`.
However, I don't want to accidentally allow invalid URIs in the process so I don't feel confident enough to do this yet.

PR #6376, opened 2022-04-06
don't assume that rev is a SHA1 hash

### LINKED ISSUE #6368: getRev seems to assume SHA1 hashes
Hello

I am making a fetcher plugin that uses sha256 hashes for revisions, but there is one thing holding me back:

https://github.com/NixOS/nix/blob/c0ad86f6817684efc2955dae904f548ea99ad6ee/src/libfetchers/fetchers.cc#L239-L243

Line 242 in `fetchers.cc`: it seems that it assumes that every hash is sha1.

I have also tried to simply not call `getRev` in plugin code and get the `Hash` manually, but I'm still getting errors like this:
```
$ src/nix/nix --extra-experimental-features 'nix-command flakes' flake metadata fsl+https://fossil.galkowski.xyz/nix-cl?ref=flake
fetching Fossil repository 'https://fossil.galkowski.xyz/nix-cl'
Resolved URL:  fsl+https://fossil.galkowski.xyz/nix-cl?ref=flake
Locked URL:    fsl+https://fossil.galkowski.xyz/nix-cl?ref=flake&rev=a89ab27eacb1736c22630d9824a9b1147683fa41fa2f2c482c9d4ea456c993e4
Description:   Common Lisp package sets and `withPackages` wrappers
Path:          /nix/store/nwbhqqwgi1phpxxhikrfff8gcypah0sp-source
error: hash 'a89ab27eacb1736c22630d9824a9b1147683fa41fa2f2c482c9d4ea456c993e4' has wrong length for hash type 'sha1'
```
I think this is because other parts of the code are using the version of `getRev` from `fetchers.cc`. By looking at the above stderr, I think it's line 202 in `flake.cc` that's erroring out:

https://github.com/NixOS/nix/blob/c0ad86f6817684efc2955dae904f548ea99ad6ee/src/nix/flake.cc#L199-L205

I have two questions:
- Is the `htSHA1` in `fetchers.cc` there on purpose, or should it be changed to allow for other hash types?
- Is there some smart way to make my plugin work, without changing Nix source code itself or without truncating my hash to 40 characters?

*full code here: https://gist.github.com/Uthar/8b360cc12153ecd827a33fe990e2eb82*

Uthar:
This was a problem when writing a fetcher that uses e.g. sha256 hashes
for revisions. This doesn't actually do anything new, but allows for
creating such fetchers in the future (perhaps when support for Git's
SHA256 object format gains more popularity).

fix #6368

### DIFF
--- a/src/libfetchers/fetchers.cc
+++ b/src/libfetchers/fetchers.cc
@@ -238,9 +238,18 @@ std::optional<std::string> Input::getRef() const
 
 std::optional<Hash> Input::getRev() const
 {
-    if (auto s = maybeGetStrAttr(attrs, "rev"))
-        return Hash::parseAny(*s, htSHA1);
-    return {};
+    std::optional<Hash> hash = {};
+
+    if (auto s = maybeGetStrAttr(attrs, "rev")) {
+        try {
+            hash = Hash::parseAnyPrefixed(*s);
+        } catch (BadHash &e) {
+            // Default to sha1 for backwards compatibility with existing flakes
+            hash = Hash::parseAny(*s, htSHA1);
+        }
+    }
+
+    return hash;
 }
 
 std::optional<uint64_t> Input::getRevCount() const
--- a/src/libutil/hash.cc
+++ b/src/libutil/hash.cc
@@ -155,7 +155,7 @@ static std::pair<std::optional<HashType>, bool> getParsedTypeAndSRI(std::string_
 {
     bool isSRI = false;
 
-    // Parse the has type before the separater, if there was one.
+    // Parse the hash type before the separator, if there was one.
     std::optional<HashType> optParsedType;
     {
         auto hashRaw = splitPrefixTo(rest, ':');
--- a/src/libutil/hash.hh
+++ b/src/libutil/hash.hh
@@ -93,13 +93,11 @@ public:
 
     std::string gitRev() const
     {
-        assert(type == htSHA1);
         return to_string(Base16, false);
     }
 
     std::string gitShortRev() const
     {
-        assert(type == htSHA1);
         return std::string(to_string(Base16, false), 0, 7);
     }
 


thufschmitt (changes requested):
Thanks for the PR :)

I’m a bit worried of just disambiguating on the hash length as that seems quite fragile (in particular just encoding the hash differently − base16 vs base32 vs base64 − would break).

I think it would me more reliable to just keep it as-it-is and expect the fetchers to explicitly prefix the hashes if they aren’t a sha1

Uthar:
Yes, thank you for catching that - let's not get rid of one assumption (SHA1) and introduce another (base16) :-).

I thought maybe something like this: https://github.com/NixOS/nix/commit/2c2fd4946f96e6839ecbfb4cf61318d8910e7e8f Allow plugins to pass their hash type with a prefix, but work unchanged for existing fetchers/flakes which pass a "naked" SHA1.

One idea is that eventually fetchers would be required to pass a "fully qualified" hash (with a prefix), and then try/catch could be removed. Or, maybe it's easy enough to make the required changes while we're at this already? If you think this is a good idea (Maybe it's better to just not touch it?)

Uthar:
Yes, thank you for catching that - let's not get rid of one assumption (SHA1) and introduce another (base16) :-).

I thought maybe something like this: https://github.com/NixOS/nix/commit/2c2fd4946f96e6839ecbfb4cf61318d8910e7e8f Allow plugins to pass their hash type with a prefix, but work unchanged for existing fetchers/flakes which pass a "naked" SHA1.

One idea is that eventually fetchers would be required to pass a "fully qualified" hash (with a prefix), and then try/catch could be removed. Or, maybe it's easy enough to make the required changes while we're at this already? If you think this is a good idea (Maybe it's better to just not touch it?)

thufschmitt:
> I thought maybe something like this: https://github.com/NixOS/nix/commit/2c2fd4946f96e6839ecbfb4cf61318d8910e7e8f Allow plugins to pass their hash type with a prefix, but work unchanged for existing fetchers/flakes which pass a "naked" SHA1.

I think you don’t even need that: `parseAny` already takes care of using the prefix if it is specified (the `HashType` that you pass as argument will only be used if the hash isn’t prefixed)

Uthar:
>  think you don’t even need that: parseAny already takes care of using the prefix if it is specified (the HashType that you pass as argument will only be used if the hash isn’t prefixed)

I tried this (calling just `Hash::parseAny(*s, htSHA1)`, where hash was `sha256:...`), but was getting different results - errors such as:
```
$ src/nix/nix --extra-plugin-files ~/projects/nix-fsl-fetcher/result/bin/libnix-fsl-fetcher.so --extra-experimental-features 'nix-command flakes' flake metadata fsl+https://fossil.galkowski.xyz/nix-fsl-fetcher
error: hash 'sha256:2ea84169c90d58c8f2fd4b810fe9ea70eff17e61d6b403000792c7e0466eb982' should have type 'sha1'
```

This is coming from line 198 in `hash.cc`, where the `htSHA1` is in conflict with the `sha256:` prefix:

https://github.com/NixOS/nix/blob/8b1e328d5d0ae7d3a4a8f6012ec065b59674ed4a/src/libutil/hash.cc#L193-L198

This is why I had to use the try-catch around `parseAnyPrefixed`: this way, we can first handle the case where the hash is "fully qualified" (such as from `plugin-files`, or from future fetchers in Nix core), and if it is not, fall back to the original SHA1 behavior for compatibility with existing fetchers, which pass the naked hash.

thufschmitt (approved):
> This is coming from line 198 in hash.cc, where the htSHA1 is in conflict with the sha256: prefix:

Oh yes you’re right, I read that too quickly! (maybe worth making the doc clearer on that front too, but that’s a thing for another time)

It’s all good then, thanks :)

Uthar:
> It’s all good then, thanks :)

Thank you :-)

edolstra:
I think for Git and Mercurial, we should verify that the rev is 20 bytes (or at least 20 bytes, if we want to be future-proof). Otherwise you can now pass an MD5 hash and Git will interpret it as a shortened rev, which is not what we want.

Uthar:
> I think for Git and Mercurial, we should verify that the rev is 20 bytes

Could this be done like this? https://github.com/NixOS/nix/pull/6387

Prevents e.g. `nix flake metadata git+https://github.com/nixos/nix?rev=md5:646af7325d93f98802b989f8a8e008a2`

PR #6380, opened 2022-04-07
Allow empty path segments in urls

thufschmitt:
Valid per https://datatracker.ietf.org/doc/html/rfc3986#section-3.3 (and also somewhat frequently happening for local paths).

I didn’t include it in the realease notes because I don’t think it’s worth it, but let me know if it needs to be

### DIFF
--- a/src/libutil/tests/url.cc
+++ b/src/libutil/tests/url.cc
@@ -178,7 +178,7 @@ namespace nix {
     }
 
     TEST(parseURL, parseFileURLWithQueryAndFragment) {
-        auto s = "file:///none/of/your/business";
+        auto s = "file:///none/of//your/business";
         auto parsed = parseURL(s);
 
         ParsedURL expected {
@@ -186,7 +186,7 @@ namespace nix {
             .base = "",
             .scheme = "file",
             .authority = "",
-            .path = "/none/of/your/business",
+            .path = "/none/of//your/business",
             .query = (StringMap) { },
             .fragment = "",
         };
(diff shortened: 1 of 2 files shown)


src/libexpr/lexer.l
```diff
@@ -115,7 +115,7 @@ ID          [a-zA-Z\_][a-zA-Z0-9\_\'\-]*
 INT         [0-9]+
 FLOAT       (([1-9][0-9]*\.[0-9]*)|(0?\.[0-9]+))([Ee][+-]?[0-9]+)?
 PATH_CHAR   [a-zA-Z0-9\.\_\-\+]
-PATH        {PATH_CHAR}*(\/{PATH_CHAR}+)+\/?
+PATH        {PATH_CHAR}*(\/{PATH_CHAR}*)+\/?
```
edolstra, line 118: Is there actually any advantage to allowing this? Presumably a Nix formatter would get rid of superfluous empty path segments, so I don't think we should allow this.

Also note that paths starting with // have special meaning even in POSIX (they're UNC paths), so that's a good reason to disallow them unless we're prepared to handle them.

thufschmitt: For the litteral paths, I just included it for consistency and because the test was failing otherwise, but I don’t think it’s very important.

For the url parser, I think it’s good to allow it, both because it’s easy to accidentally build a path with two slashes in a row (`ROOT=/foo/bar/; PATH=$ROOT/subdir`) and because you can actually have urls with two slashes in them

thufschmitt: Apparently this is breaking the parser anyways. So I’ll just keep it for the uri regex:

```
nix>     error: path has a trailing slash
nix>            at /build/source/tests/lang/eval-okay-types.nix:11:14:
nix>                10|   (isInt { x = 123; })
nix>                11|   (isInt (1 / 2))
nix>                  |              ^
nix>                12|   (isInt (1 + 1))
```

PR #6387, opened 2022-04-08
assert hash types for Git and Mercurial

Uthar:


src/libfetchers/git.cc
```diff
@@ -202,6 +202,7 @@ struct GitInputScheme : InputScheme
             -> std::pair<StorePath, Input>
         {
             assert(input.getRev());
+            assert(input.getRev()->type == htSHA1 || input.getRev()->type == htSHA256);
```
edolstra, line 205: This shouldn't be an assertion but an exception, because it can be triggered by user input.

Uthar: Alright. Changed to exceptions in 2769e43f61d827e1b8fe46257450986d76319243.

Tested like this:
```
nix flake metadata git+https://github.com/nixos/nix?rev=md5:646af7325d93f98802b989f8a8e008a2
nix flake metadata hg+https://hg.stevelosh.com/cl-chip8?rev=md5:205427feb9feb7297ca2c70cf6f0c86b
nix flake metadata hg+https://hg.stevelosh.com/cl-chip8?rev=sha256:a475283ffac2ae4cebe1a6bbccfe06db84f9973548aefeec13ee6ef3662a8ebc
```

PR #6389, opened 2022-04-09
Make flake show less lenient on apps

tomberek:
Fixes: https://github.com/NixOS/nix/pull/5652
Fixes: https://github.com/NixOS/nix/issues/5560

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -40,3 +40,7 @@
 
   As before, the old output will continue to work, but `nix flake check` will
   issue a warning about it.
+
+* `nix run` is now stricter wrt what it accepts:
+    * Members of `apps` are now required to be apps (as defined in [the manual](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-run.html#apps))
+    * Member of `packages` or `legacyPackages` cannot be of type "app" when used by `nix run`.
--- a/src/nix/app.cc
+++ b/src/nix/app.cc
@@ -65,6 +65,15 @@ UnresolvedApp Installable::toApp(EvalState & state)
 
     auto type = cursor->getAttr("type")->getString();
 
+    std::string expected;
+    if (hasPrefix(attrPath,"apps.")) {
+        expected = "app";
+    } else {
+        expected = "derivation";
+    }
+    if (type != expected) {
+        throw Error("Attribute '%s' should have type '%s'.", attrPath, expected);
+    }
     if (type == "app") {
         auto [program, context] = cursor->getAttr("program")->getStringWithContext();
 
--- a/tests/flakes-run.sh
+++ b/tests/flakes-run.sh
@@ -1,0 +1,29 @@
+source common.sh
+
+clearStore
+rm -rf $TEST_HOME/.cache $TEST_HOME/.config $TEST_HOME/.local
+cp ./shell-hello.nix ./config.nix $TEST_HOME
+cd $TEST_HOME
+
+cat <<EOF > flake.nix
+{
+    outputs = {self}: {
+      packages.$system.PkgAsPkg = (import ./shell-hello.nix).hello;
+      packages.$system.AppAsApp = self.packages.$system.AppAsApp;
+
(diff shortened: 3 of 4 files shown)


src/nix/app.cc
```diff
@@ -79,7 +88,7 @@ UnresolvedApp Installable::toApp(EvalState & state)
         }};
     }
 
-    else if (type == "derivation") {
+    else if (type == "derivation"){
```
edolstra, line 91:
```suggestion
    else if (type == "derivation") {
```

tomberek: thx!

tests/flakes-run.sh
```diff
@@ -0,0 +1,29 @@
...
+        program = "\${(import ./shell-hello.nix).hello}/bin/hello";
+      };
+    };
+}
+EOF
+nix run --no-write-lock-file .#AppAsApp
+nix run --no-write-lock-file .#PkgAsPkg
+
+! nix run --no-write-lock-file .#PkgAsApp || fail "should fail"
+! nix run --no-write-lock-file .#AppAsPkg || fail "should fail"
```
thufschmitt, line 26: Worth a slightly more detailed failure message imho. Maybe something like:

```suggestion
! nix run --no-write-lock-file .#PkgAsApp || fail  "'nix run' shouldn’t accept an 'app' defined under 'packages'"
! nix run --no-write-lock-file .#AppAsPkg || fail "Elements of 'apps' should be of type 'app'"
```

src/nix/app.cc
```diff
@@ -65,6 +65,15 @@ UnresolvedApp Installable::toApp(EvalState & state)
...
 
+    std::string expected;
+    if (hasPrefix(attrPath,"apps.")) {
+        expected = "app";
+    } else {
+        expected = "derivation";
+    }
+    if (type != expected) {
+        throw Error("Attribute '%s' should have type '%s'. Apps must be located in the apps output attribute. The apps output attribute must contain only apps.", attrPath, expected);
+    }
```
thufschmitt, line 76: I don’t find the error message very clear. I’d suggest to either stay very generic (`Attribute '%s' should have type '%s'`) or have a totally different error messages in both cases.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -40,3 +40,6 @@
 
   As before, the old output will continue to work, but `nix flake check` will
   issue a warning about it.
+
+* Items in the apps output schema must be of type "app" and other uses of `nix
+  run` must be of type derivation.
```
thufschmitt, line 45:
```suggestion
* `nix run` is now stricter wrt what it accepts:
    * Members of `apps` are now required to be apps (as defined in [the manual](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-run.html#apps))
    * Member of `packages` or `legacyPackages` must be plain derivations
```

Which also makes me realize that the [description](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-run.html#description) section in the manual would need to be updated

thufschmitt (changes requested):
Looks good overall, just a lot of tiny nitpicking on all the user-visible strings :stuck_out_tongue:

doc/manual/src/release-notes/rl-next.md
zseri: huh, but "member[s] (sic.) of `legacyPackages` must be plain derivations" is not really correct, because (nested) attrsets containing derivations are/should also be allowed there, which is afaik the purpose of `legacyPackages` in contrast to `packages`.

src/nix/app.cc
tomberek: I was trying to be explicit that:

>  Attributes shall be of type 'app' if and only if they are in the apps output schema in order to be correctly handled by `nix run` and `nix flake check`

Not sure of the best way to specify this.... reverting.....

doc/manual/src/release-notes/rl-next.md
tomberek: re-reading the src/nix/run.md and it seems the new behavior is already documented:

    An app is specified by a flake output attribute named `apps.<system>.<name>"...

Issue #6383, opened 2022-04-08
`resolvedResult.builtOutputs` doesn't contain expected output after successful multi-output build


zseri:
**Describe the bug**

Got the following error, but absolutely no information where exactly it occured (as-in no stacktrace or location info).
idk how to debug this.

```
openblas> patching script interpreter paths in /nix/store/fjkhb5vm7wdvs3piirjgvxd7gbpbq20b-openblas-0.3.20
openblas> checking for references to /build/ in /nix/store/fjkhb5vm7wdvs3piirjgvxd7gbpbq20b-openblas-0.3.20...
openblas> shrinking RPATHs of ELF executables and libraries in /nix/store/fqhw0blv04is1mc9r1ziqcgp9v4xc2s8-openblas-0.3.20-dev
openblas> strip is /nix/store/1kmkmwvs6d50z4jghnildm4cj3pzhvrq-gcc-wrapper-9.3.0/bin/strip
openblas> stripping (with command strip and flags -S) in /nix/store/fqhw0blv04is1mc9r1ziqcgp9v4xc2s8-openblas-0.3.20-dev/lib
openblas> patching script interpreter paths in /nix/store/fqhw0blv04is1mc9r1ziqcgp9v4xc2s8-openblas-0.3.20-dev
openblas> checking for references to /build/ in /nix/store/fqhw0blv04is1mc9r1ziqcgp9v4xc2s8-openblas-0.3.20-dev...
error: map::at
```

**Steps To Reproduce**

Unknown.

**Expected behavior**

The error should either not happen, or at least contain enough information to know where to start debugging, or a stacktrace...

**`nix-env --version` output**

nix-env (Nix) 2.8.0pre20220401_a24d077
but this might not be the version actually used, because this was hit when using `nixos-rebuild`, which first rebuilds itself.
the other rev involved is 8b1e328d5d0ae7d3a4a8f6012ec065b59674ed4a.
Also happens with nix-env (Nix) 2.8.0pre20220307_92b8d4d.
nix-env (Nix) 2.7.0pre20220225_df552ff: unable to test against, because it hits a known SIGBUS before gettings near this bug.

**Additional context**
`experimental-features = nix-command ca-derivations flakes`

zseri:
```
[zseri@orca:~/devel/nix]$ rg -w '[.>]at'
src/nix/hash.cc
185:        else if (*arg != "" && arg->at(0) == '-')

src/toml11/README.md
[...]

PR #6393, opened 2022-04-10
Get rid of most `.at` calls

zseri:
This a step towards fixing #6383, but currently in a sub-optimal state (exception text not good enough, too many changes at unrelated positions)

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -984,21 +984,28 @@ void DerivationGoal::resolvedFinished()
             realWantedOutputs = resolvedDrv.outputNames();
 
         for (auto & wantedOutput : realWantedOutputs) {
-            assert(initialOutputs.count(wantedOutput) != 0);
-            assert(resolvedHashes.count(wantedOutput) != 0);
-            auto realisation = resolvedResult.builtOutputs.at(
-                DrvOutput { resolvedHashes.at(wantedOutput), wantedOutput });
+            auto initialOutput = get(initialOutputs, wantedOutput);
+            auto resolvedHash = get(resolvedHashes, wantedOutput);
+            if ((!initialOutput) || (!resolvedHash))
+                throw Error(
+                    "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,resolve)",
+                    worker.store.printStorePath(drvPath), wantedOutput);
+            auto realisation = get(resolvedResult.builtOutputs, DrvOutput { *resolvedHash, wantedOutput });
+            if (!realisation)
+                throw Error(
+                    "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,realisation)",
+                    worker.store.printStorePath(resolvedDrvGoal->drvPath), wantedOutput);
             if (drv->type().isPure()) {
-                auto newRealisation = realisation;
-                newRealisation.id = DrvOutput { initialOutputs.at(wantedOutput).outputHash, wantedOutput };
+                auto newRealisation = *realisation;
+                newRealisation.id = DrvOutput { initialOutput->outputHash, wantedOutput };
                 newRealisation.signatures.clear();
                 if (!drv->type().isFixed())
-                    newRealisation.dependentRealisations = drvOutputReferences(worker.store, *drv, realisation.outPath);
+                    newRealisation.dependentRealisations = drvOutputReferences(worker.store, *drv, realisation->outPath);
                 signRealisation(newRealisation);
                 worker.store.registerDrvOutput(newRealisation);
             }
-            outputPaths.insert(realisation.outPath);
-            builtOutputs.emplace(realisation.id, realisation);
+            outputPaths.insert(realisation->outPath);
+            builtOutputs.emplace(realisation->id, *realisation);
         }
 
         runPostBuildHook(
@@ -1294,7 +1301,11 @@ std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
     DrvOutputs validOutputs;
 
     for (auto & i : queryPartialDerivationOutputMap()) {
-        InitialOutput & info = initialOutputs.at(i.first);
+        auto initialOutput = get(initialOutputs, i.first);
+        if (!initialOutput)
+            // this is an invalid output, gets catched with (!wantedOutputsLeft.empty())
+            continue;
+        auto & info = *initialOutput;
         info.wanted = wantOutput(i.first, wantedOutputs);
         if (info.wanted)
             wantedOutputsLeft.erase(i.first);
@@ -1309,7 +1320,7 @@ std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
                     : PathStatus::Corrupt,
             };
         }
-        auto drvOutput = DrvOutput{initialOutputs.at(i.first).outputHash, i.first};
+        auto drvOutput = DrvOutput{info.outputHash, i.first};
         if (settings.isExperimentalFeatureEnabled(Xp::CaDerivations)) {
             if (auto real = worker.store.queryRealisation(drvOutput)) {
                 info.known = {
--- a/src/libstore/build/worker.cc
+++ b/src/libstore/build/worker.cc
@@ -380,7 +380,10 @@ void Worker::waitForInput()
         std::set<int> fds2(j->fds);
         std::vector<unsigned char> buffer(4096);
         for (auto & k : fds2) {
-            if (pollStatus.at(fdToPollStatus.at(k)).revents) {
+            const auto fdPollStatusId = get(fdToPollStatus, k);
+            assert(fdPollStatusId);
+            assert(*fdPollStatusId < pollStatus.size());
+            if (pollStatus.at(*fdPollStatusId).revents) {
                 ssize_t rd = ::read(k, buffer.data(), buffer.size());
                 // FIXME: is there a cleaner way to handle pt close
                 // than EIO? Is this even standard?
--- a/src/libstore/derived-path.cc
+++ b/src/libstore/derived-path.cc
@@ -22,7 +22,9 @@ nlohmann::json DerivedPath::Built::toJSON(ref<Store> store) const {
     const auto knownOutputs = store->queryPartialDerivationOutputMap(drvPath);
     for (const auto& output : outputs) {
         auto knownOutput = get(knownOutputs, output);
-        res["outputs"][output] = knownOutput ? store->printStorePath(*knownOutput) : nullptr;
+        res["outputs"][output] = (knownOutput && *knownOutput)
+          ? store->printStorePath(**knownOutput)
+          : nullptr;
     }
     return res;
 }
--- a/src/libstore/derived-path.cc
+++ b/src/libstore/derived-path.cc
@@ -4,6 +4,8 @@
 
 #include <nlohmann/json.hpp>
 
+#include <optional>
+
 namespace nix {
 
 nlohmann::json DerivedPath::Opaque::toJSON(ref<Store> store) const {
@@ -17,12 +19,10 @@ nlohmann::json DerivedPath::Built::toJSON(ref<Store> store) const {
     res["drvPath"] = store->printStorePath(drvPath);
     // Fallback for the input-addressed derivation case: We expect to always be
     // able to print the output paths, so let’s do it
-    auto knownOutputs = store->queryPartialDerivationOutputMap(drvPath);
+    const auto knownOutputs = store->queryPartialDerivationOutputMap(drvPath);
     for (const auto& output : outputs) {
-        if (knownOutputs.at(output))
-            res["outputs"][output] = store->printStorePath(knownOutputs.at(output).value());
-        else
-            res["outputs"][output] = nullptr;
+        auto knownOutput = get(knownOutputs, output);
+        res["outputs"][output] = knownOutput ? store->printStorePath(*knownOutput) : nullptr;
     }
     return res;
 }
@@ -123,10 +123,15 @@ RealisedPath::Set BuiltPath::toRealisedPaths(Store & store) const
                 for (auto& [outputName, outputPath] : p.outputs) {
                     if (settings.isExperimentalFeatureEnabled(
                                 Xp::CaDerivations)) {
+                        auto drvOutput = get(drvHashes, outputName);
+                        if (!drvOutput)
+                            throw Error(
+                                "the derivation '%s' has unrealised output '%s' (derived-path.cc/toRealisedPaths)",
+                                store.printStorePath(p.drvPath), outputName);
                         auto thisRealisation = store.queryRealisation(
-                            DrvOutput{drvHashes.at(outputName), outputName});
-                        assert(thisRealisation);  // We’ve built it, so we must h
-                                                  // ve the realisation
+                            DrvOutput{*drvOutput, outputName});
+                        assert(thisRealisation);  // We’ve built it, so we must
+                                                  // have the realisation
                         res.insert(*thisRealisation);
                     } else {
                         res.insert(outputPath);
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -718,7 +718,11 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
                     // somewhat expensive so we do lazily
                     hashesModulo = hashDerivationModulo(*this, drv, true);
                 }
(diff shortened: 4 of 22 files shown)


src/libstore/build/derivation-goal.cc
```diff
@@ -985,21 +985,29 @@ void DerivationGoal::resolvedFinished()
...
+            if ((initialOutputs.end() == initialOutput) || (resolvedHashes.end() == resolvedHash))
+                throw Error(
+                    "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,resolve)",
+                    worker.store.printStorePath(drvPath), wantedOutput);
+            auto realisation = resolvedResult.builtOutputs.find(
+                DrvOutput { resolvedHash->second, wantedOutput });
+            if (resolvedResult.builtOutputs.end() == realisation)
+                throw Error(
+                    "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,realisation)",
+                    worker.store.printStorePath(resolvedDrvGoal->drvPath), wantedOutput);
```
zseri, line 999: This is the exception hit in #6383.

zseri: maybe `resolvedHash->second` mismatches? Is there some way to format `resolvedResult.builtOutputs` to include it in the error message?

thufschmitt (commented):
Thanks, I think that’s a good thing to do :) (and I must confess I’m probably guilty of a lot of them :/ ).

Maybe you could make the change nicer with something like

```cpp
template<typename C>
C::mapped_type getOrError(const C & container, const C::key_type & key, const Error & e)
{
    auto iter = container.find(key);
    if (iter == container.end())
        throw e;
    return iter->second;
}
```

so that in most cases you can replace `foo.at(x)` by `getOrError(foo, x, Error("'foo' doesn’t contain %s", x))`

zseri:
Although that would work, it would result in many unnecessary unconditional allocations... using a lambda to construct Error is probably a better idea.

edolstra:
I would rather add a variant of `get()` in util.hh that returns a pointer to the value, because then you can write
```
if (auto x = get(map, key))
  ...
else
  throw Error(...);
```

zseri:
why is there some code duplication between `libcmd/installables.cc` and `libstore/remote-store.cc`? (btw. is there some issue which collects all (semi-) code duplications?)

zseri:
well, that doesn't work.

zseri:
aa11ad912854c0b9170f4f8da38fde679589d5c5 is mergable. the rest isn't.

thufschmitt:
> well, that doesn't work.

Yeah, `std::optional<T&>` isn’t really a thing afaik. Maybe you could get around with a `std::shared_ptr<T>`…

Or the `OrError` version can also work with a bit of tweaks:

```cpp
template<typename C, typename E>
typename C::mapped_type & getOrError(C & container, const typename C::key_type & key, E e)
{
    auto iter = container.find(key);
    if (iter == container.end())
        throw e();
    return iter->second;
}
```

zseri:
Although it would work, I don't really think it's a good idea because it makes the code not really more readable, and the duplication it causes is well understood, because that's how iterators work in C++, and litters the code already in many places unavoidably. Without lambdas, it's somewhat ok, but incurs probably a perf hit in dozens of code paths, which is imo extremely undesirable, because performance is already a pain point of Nix... With lambdas, it gets obscure and complicated enough that I don't think the wrapper is worth it.

thufschmitt:
Fair point. I really don’t like the verbosity of this iterator-based way of working, but it’s good-enough. Let’s keep things as they are then

zseri:
Yesterday I also fixed most of the remaining afaik nitpicks:
* `stringifiedFeatures`: use `assert` instead `.at`
* `"internal error: [...]"`: use `assert` instead; condition already checked in previous loop of same function, we don't assume it changes underneath us (which we can't verify automatically in C++, but could in Rust...)

src/nix-build/nix-build.cc
```diff
@@ -235,7 +235,7 @@ static void main_nix_build(int argc, char * * argv)
         else if (*arg == "-")
             readStdin = true;
 
-        else if (*arg != "" && arg->at(0) == '-')
+        else if (!arg->empty() && (*arg)[0] == '-')
```
thufschmitt, line 238: Does that bring anything? A failing `at` call isn’t nice because it’s hard to track, but this is undefined if out-of-bounds which isn’t nicer

thufschmitt (commented):
I checked the various changes to make sure that there’s no accidental semantic change, looks good :+1: 

I’m not thrilled by these raw iterators everywhere though: A failing `at` call isn’t nice, but dereferencing a past-the-end iterator is undefined behavior (and from what I could see, both gcc and clang will happily return some garbage in most cases). So that look like a bomb waiting to explode.

src/libstore/build/worker.cc
```diff
@@ -380,7 +380,10 @@ void Worker::waitForInput()
         std::set<int> fds2(j->fds);
         std::vector<unsigned char> buffer(4096);
         for (auto & k : fds2) {
-            if (pollStatus.at(fdToPollStatus.at(k)).revents) {
+            auto fdps1 = fdToPollStatus.find(k);
+            if (fdToPollStatus.end() == fdps1)
+                throw Error("fdToPollStatus lookup failed for fd %1%", k);
+            if (pollStatus.at(fdps1->second).revents) {
```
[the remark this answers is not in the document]
zseri: this one is `vector::at`, it's distinct from the `map::at` (`map::at` is more difficult to debug, because the key is more complex than an index). But ok, this one should imo an assert instead.

src/nix-build/nix-build.cc
zseri: ok, might not be worth it (it's argument parsing and can be really slow without impacting anything much...)

zseri: also, this can't be out-of-bounds because C++ strings are always zero-terminated (`string_view`s aren't).

zseri:
maybe a better alternative would be to produce something which does resemble a stack trace by catching exceptions at more places and annotate them with more metadata, e.g. about which derivation was in progress, and such.

A problem is though, that `try { ...at() } catch(...) { ... }` is an anti-pattern, and afaik might also result in miscompilations unless `-O0` is given (at least that was the case 2016/2017)...

Another alternative (I would prefer this one) would be to modify `get(...)` to return a pointer for the mutable case (for consistency, it might be a good idea to always return a pointer), so that we get a "clean" deterministic SIGSEGV at null-pointer dereference, when some check is incorrect.

thufschmitt:
> Another alternative (I would prefer this one) would be to modify get(...) to return a pointer for the mutable case (for consistency, it might be a good idea to always return a pointer), so that we get a "clean" deterministic SIGSEGV at null-pointer dereference, when some check is incorrect.

I like this idea :) That’s actually very close to the `std::optional<T&>` you tried to do, but working and with potentially a better error in the end

zseri:
I think I don't really understand how to fix this error...
```
nix> In file included from src/libutil/util.hh:4,
nix>                  from src/libutil/serialise.hh:6,
nix>                  from src/libutil/archive.hh:4,
nix>                  from src/libexpr/primops.cc:1:
nix> src/libutil/error.hh: In instantiation of 'nix::ErrPos& nix::ErrPos::operator=(const P&) [with P = nix::PosIdx]':
nix> src/libutil/error.hh:97:15:   required from 'nix::ErrPos::ErrPos(const P&) [with P = nix::PosIdx]'
nix> /nix/store/l2iiwsrrscp3k15bl5bwmk07jjnb4zxs-gcc-10.3.0/include/c++/10.3.0/optional:225:47:   required from 'constexpr std::_Optional_payload_base<_Tp>::_Storage<_Up, false>::_Storage(std::in_place_t, _Args&& ...) [with _Args = {const nix::PosIdx&}; _Up = nix::ErrPos; _Tp = nix::ErrPos]'
nix> /nix/store/l2iiwsrrscp3k15bl5bwmk07jjnb4zxs-gcc-10.3.0/include/c++/10.3.0/optional:117:19:   required from 'constexpr std::_Optional_payload_base<_Tp>::_Optional_payload_base(std::in_place_t, _Args&& ...) [with _Args = {const nix::PosIdx&}; _Tp = nix::ErrPos]'
nix> /nix/store/l2iiwsrrscp3k15bl5bwmk07jjnb4zxs-gcc-10.3.0/include/c++/10.3.0/optional:358:42:   required from here
nix> src/libutil/error.hh:87:22: error: 'const class nix::PosIdx' has no member named 'origin'
nix>    87 |         origin = pos.origin;
nix>       |                  ~~~~^~~~~~
nix> src/libutil/error.hh:88:20: error: 'const class nix::PosIdx' has no member named 'line'
nix>    88 |         line = pos.line;
nix>       |                ~~~~^~~~
nix> src/libutil/error.hh:89:22: error: 'const class nix::PosIdx' has no member named 'column'
nix>    89 |         column = pos.column;
nix>       |                  ~~~~^~~~~~
nix> src/libutil/error.hh:90:20: error: 'const class nix::PosIdx' has no member named 'file'
nix>    90 |         file = pos.file;
nix>       |                ~~~~^~~~
nix> make: *** [mk/patterns.mk:3: src/libexpr/primops.o] Error 1
```

thufschmitt:
> I think I don't really understand how to fix this error...

That’s due to a change in master, `posDrvName` isn’t a `Pos` anymore, it’s a `PosIdx` (morally a pointer to `Pos`). You can get the corresponding `Pos` with `state.positions[posDrvName]`

src/nix-build/nix-build.cc
```diff
@@ -440,7 +440,8 @@ static void main_nix_build(int argc, char * * argv)
         env["NIX_STORE"] = store->storeDir;
         env["NIX_BUILD_CORES"] = std::to_string(settings.buildCores);
 
-        auto passAsFile = tokenizeString<StringSet>(get(drv.env, "passAsFile").value_or(""));
+        const auto passAsFile_ = get(drv.env, "passAsFile");
+        auto passAsFile = tokenizeString<StringSet>(passAsFile_ ? passAsFile_->c_str() : "");
```
thufschmitt, line 444: What’s the reason for this `c_str()` call?

src/libutil/util.hh
```diff
@@ -543,13 +543,31 @@ std::string stripIndentation(std::string_view s);
...
+    auto i = map.find(key);
+    if (i == map.end()) return nullptr;
+    return &i->second;
+}
+
+/* Get a value for the specified key from an associate container, or a default value if the key isn't present. */
+template <class T>
+const typename T::mapped_type & get_or(T & map,
+    const typename T::key_type & key,
+    const typename T::mapped_type & dflval)
```
thufschmitt, line 565:
```suggestion
    const typename T::mapped_type & defaultValue)
```

src/libutil/util.hh
```diff
@@ -543,13 +543,31 @@ std::string stripIndentation(std::string_view s);
...
+typename T::mapped_type * get(T & map, const typename T::key_type & key)
+{
+    auto i = map.find(key);
+    if (i == map.end()) return nullptr;
+    return &i->second;
+}
+
+/* Get a value for the specified key from an associate container, or a default value if the key isn't present. */
+template <class T>
+const typename T::mapped_type & get_or(T & map,
```
thufschmitt, line 563:
```suggestion
const typename T::mapped_type & getOr(T & map,
```

(that’s the general convention in the codebase)

src/libstore/local-store.cc
```diff
@@ -718,7 +718,11 @@ void LocalStore::checkDerivationOutputs(const StorePath & drvPath, const Derivat
                     // somewhat expensive so we do lazily
                     hashesModulo = hashDerivationModulo(*this, drv, true);
                 }
-                StorePath recomputed = makeOutputPath(i.first, hashesModulo->hashes.at(i.first), drvName);
+                auto hashifi = get(hashesModulo->hashes, i.first);
```
thufschmitt, line 721:
```suggestion
                auto currentOutputHash = get(hashesModulo->hashes, i.first);
```

(This recurrent `for(auto i : …)` pattern makes things unpleasantly hard to read, let’s not make it worse ;) )

src/libstore/build/worker.cc
```diff
@@ -380,7 +380,10 @@ void Worker::waitForInput()
         std::set<int> fds2(j->fds);
         std::vector<unsigned char> buffer(4096);
         for (auto & k : fds2) {
-            if (pollStatus.at(fdToPollStatus.at(k)).revents) {
+            auto fdps1 = get(fdToPollStatus, k);
+            assert(fdps1);
+            assert(*fdps1 < pollStatus.size());
+            if (pollStatus.at(*fdps1).revents) {
```
thufschmitt, line 386: I won’t suggest an alternative variable name for `fdps1` because I have no idea what this code is doing, but if you have one, that would be immensely helpful

thufschmitt (changes requested):
Left a couple of naming-related comments, but apart from that (and the build failure, but that’s hopefully easy to fix) :+1: , I think that’s a big improvement

src/libutil/util.hh
zseri: counterpoint: `remove_begin` (directly below this def.) also uses the STL naming convention.

zseri:
ok, one thing left: Can we check/make sure that the `getOr($container, $key, "")` doesn't return a dangling reference in the fallback/"default" case? (I'm always pretty skeptical about such stuff, because I've ran into dangling `std::string` usages way too many times, which is one of reasons why I usually prefer Rust, which fixes this problem at the root (that the programmer and the compiler might disagree about when a destructor should be called))... although in this case it's probably not that bad as usual, because "" is an empty string... if anywhere the used default argument is constructed inline and isn't an empty string, then this might run into another case of dangling references / basically use-after-free, except in cases where the data is directly copied.

zseri:
In `ExternalValueBase`, we neither have access to `state` nor `positions`...

zseri:
I think it would be a good idea to merge this as a squash merge if it gets merged, all the iterations to get single things working aren't worth putting into the commit log...

thufschmitt (approved):
Thanks a lot, hopefully that will make all these errors much more tractable

> ok, one thing left: Can we check/make sure that the getOr($container, $key, "") doesn't return a dangling reference in the fallback/"default" case?

What would be the point of writing C++ if there wasn’t all these funky footguns? :stuck_out_tongue:

src/libstore/derived-path.cc
```diff
@@ -123,10 +125,15 @@ RealisedPath::Set BuiltPath::toRealisedPaths(Store & store) const
                 for (auto& [outputName, outputPath] : p.outputs) {
                     if (settings.isExperimentalFeatureEnabled(
                                 Xp::CaDerivations)) {
+                        auto drvOutput = get(drvHashes, outputName);
+                        if (!drvOutput)
+                            throw Error(
+                                "the derivation '%s' has unrealised output '%s' (derived-path.cc/toRealisedPaths)",
```
edolstra, line 131: IMHO we shouldn't have references to source code in error messages (like `derived-path.cc/toRealisedPaths`) because they're meaningless to users.

If this is a "can't happen" logical error, then it's better to use `assert` (since that's much easier to debug using gdb).

zseri: I would, but the problem is that exactly that scenario happened which is what prompted me to do this PR... (I don't think an `assert` would print out enough context) ... and I didn't want to fudge the error message to make it unique so it can be associated with exactly one position in the source code where the error stems from...
An alternative would be that `Error` should get the ability to gather a full backtrace upon request (e.g. via an environment variable), and print it.

edolstra: Asserts are easier to debug (for me at least) because if you run in gdb, it stops exactly at the point where the assertion happens. Or you can load the coredump into gdb and get the same. Whereas with exceptions, you have to use `catch throw`, but there are a lot of other exceptions being thrown...

edolstra:
@thufschmitt Good point, I'm not sure if `getOr()` can be used safely. It returns a reference to the default value, so if that's a temporary, then it may be destroyed before we use it. I think that's actually likely to happen if default value is an `std::string` and we pass a string constant.

zseri:
We can mitigate that for most occasions by "simply" using string_view instead, and only supplying statically allocated strings.

src/libstore/derived-path.cc
zseri: follow-up PR: https://github.com/NixOS/nix/pull/6484 (males the error message more usable and fixes `getOr`)

zseri: btw. didn't change it to `assert` because originally it also wasn't an assert.

PR #6399, opened 2022-04-11
fix: don't call functors in a flake fragment

blaggacao:
- flake fragement evaluation is always hermetic
- calling a function underway is not only breaking things
  but also against that principle

This commit adds code path variant invokend on flake fragments that
abstains from auto-calling functions.


blaggacao:
#6400 is from correct branch. This was wrong branch, sry.

PR #6400, opened 2022-04-11
fix: don't call functors in a flake fragment

blaggacao:
- flake fragement evaluation is always hermetic
- calling a function underway is not only breaking things
  but also against that principle

This commit adds code path variant invokend on flake fragments that
abstains from auto-calling functions.

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -204,7 +204,7 @@ void SourceExprCommand::completeInstallable(std::string_view prefix)
             prefix_ = "";
         }
 
-        auto [v, pos] = findAlongAttrPath(*state, prefix_, *autoArgs, root);
+        auto [v, pos] = findAlongAttrPath(*state, prefix_, *autoArgs, root, false);
         Value &v1(*v);
         state->forceValue(v1, pos);
         Value v2;
@@ -452,7 +452,7 @@ struct InstallableAttrPath : InstallableValue
 
     std::pair<Value *, Pos> toValue(EvalState & state) override
     {
-        auto [vRes, pos] = findAlongAttrPath(state, attrPath, *cmd.getAutoArgs(state), **v);
+        auto [vRes, pos] = findAlongAttrPath(state, attrPath, *cmd.getAutoArgs(state), **v, false);
         state.forceValue(*vRes, pos);
         return {vRes, pos};
     }
@@ -624,7 +624,7 @@ std::pair<Value *, Pos> InstallableFlake::toValue(EvalState & state)
 
     for (auto & attrPath : getActualAttrPaths()) {
         try {
-            auto [v, pos] = findAlongAttrPath(state, attrPath, *emptyArgs, *vOutputs);
+            auto [v, pos] = findAlongAttrPath(state, attrPath, *emptyArgs, *vOutputs, false);
             state.forceValue(*v, pos);
             return {v, pos};
         } catch (AttrPathNotFound & e) {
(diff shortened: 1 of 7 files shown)


roberth:
Needs a test case.

src/libcmd/installables.cc
```diff
@@ -204,7 +204,7 @@ void SourceExprCommand::completeInstallable(std::string_view prefix)
             prefix_ = "";
         }
 
-        auto [v, pos] = findAlongAttrPath(*state, prefix_, *autoArgs, root);
+        auto [v, pos] = findAlongAttrPath(*state, prefix_, *autoArgs, root, false);
```
thufschmitt, line 207: This is for `InstallableAttrPath` which corresponds to the non-flake case. So we do want to call the functors here

src/libcmd/installables.cc
```diff
@@ -452,7 +452,7 @@ struct InstallableAttrPath : InstallableValue
 
     std::pair<Value *, Pos> toValue(EvalState & state) override
     {
-        auto [vRes, pos] = findAlongAttrPath(state, attrPath, *cmd.getAutoArgs(state), **v);
+        auto [vRes, pos] = findAlongAttrPath(state, attrPath, *cmd.getAutoArgs(state), **v, false);
```
thufschmitt, line 455: Like the above, I think we do want to call the functors here

src/libexpr/attr-path.hh
```diff
@@ -14,7 +14,8 @@ std::pair<Value *, Pos> findAlongAttrPath(
     EvalState & state,
     const std::string & attrPath,
     Bindings & autoArgs,
```
thufschmitt, line 16: These are only needed when `callFunctors` is true, aren’t they?
Maybe we could change the function signature to make `autoArgs` be a `std::optional<Bindings>` and only call he functors if the optional isn’t empty (or even better make this a proper sum type, but c++ makes them a bit painful to write and manipulate)

src/libexpr/eval.cc
```diff
@@ -1513,13 +1513,13 @@ void EvalState::incrFunctionCall(ExprLambda * fun)
...
 
-void EvalState::autoCallFunction(Bindings & args, Value & fun, Value & res)
+void EvalState::autoCallFunction(Bindings & args, Value & fun, Value & res, bool callFunctors)
 {
     auto pos = fun.determinePos(noPos);
 
     forceValue(fun, pos);
 
-    if (fun.type() == nAttrs) {
+    if (callFunctors && (fun.type() == nAttrs)) {
```
thufschmitt, line 1522: It looks like this will prevent from calling _functors_ but not plain functions. Is it intended?

thufschmitt (commented):
Thanks :)

A few comments/suggestions, but looks good on the principle.
And agree with @roberth, a test would be welcome

edolstra:
I've done a different fix in #6404, making `nix eval` use the same code path as `nix build` etc., so it won't use `findAlongAttrPath()`  and auto-call functions/functors. Does that fix the issue for you?

It would still be nice to make `findAlongAttrPath()` take a pointer to `Bindings` so you can pass a nullptr to disable calling functions/functors.

src/libexpr/attr-path.hh
edolstra: Maybe `Bindings *` instead of `std::optional<Bindings>`, otherwise the bindings are passed by value.

blaggacao:
@edolstra thanks, it's working

PR #6404, opened 2022-04-14
Make InstallableFlake::toValue() and toDerivation() behave consistently

### LINKED ISSUE #6152: `outputs.__functor` breaks `nix` CLI's target resolver

edolstra:
In particular, this means that `nix eval` (which uses `toValue()`) no longer auto-calls functions or functors (because
`AttrCursor::findAlongAttrPath()` doesn't).

Fixes #6152.

Also use `ref<>` in a few places, and don't return attrpaths from `getCursor()` because cursors already have a `getAttrPath()` method.

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -334,16 +334,16 @@ DerivedPath Installable::toDerivedPath()
     return std::move(buildables[0]);
 }
 
-std::vector<std::pair<std::shared_ptr<eval_cache::AttrCursor>, std::string>>
+std::vector<ref<eval_cache::AttrCursor>>
 Installable::getCursors(EvalState & state)
 {
     auto evalCache =
         std::make_shared<nix::eval_cache::EvalCache>(std::nullopt, state,
             [&]() { return toValue(state).first; });
-    return {{evalCache->getRoot(), ""}};
+    return {evalCache->getRoot()};
 }
 
-std::pair<std::shared_ptr<eval_cache::AttrCursor>, std::string>
+ref<eval_cache::AttrCursor>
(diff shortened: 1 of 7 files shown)


tomberek:
This more consistently uses the eval cache in situations where it currently doesn't seem to (eg: `nix eval`) I presume due to toValue now calling openEvalCache via getCursor.

Random question: 
```
$ nix eval .#nix --json
"/nix/store/7a8b4i4jy74azm9mliwzynnsczcxrzh4-nix-2.8.0pre20220414_dirty"
```
will not cache the outPath result, only the existence of attributes, while
```
$ nix eval .#nix.outPath --json
"/nix/store/7a8b4i4jy74azm9mliwzynnsczcxrzh4-nix-2.8.0pre20220414_dirty"
```
will cache the outPath into the sqlite at ~/.cache/nix/eval-cache-v2. It is often the outPath calculation that is most important to have cached.

edolstra:
This only superficially uses the eval cache (really only because it provides a `findAlongAttrPath()` that doesn't call functions/functors) The use of `forceValue()` means that we evaluate anyway.

PR #6411, opened 2022-04-15
nix-daemon.service: don't ignore sigpipe

Artturin:
to test this i removed the following lines from 
https://github.com/NixOS/nixpkgs/pull/151035/files#diff-989531da859adca1871d385292724e8e0c5b690758c7cd6188e47c869b2b9183R161-R163
and built busybox
additional context: https://lists.freedesktop.org/archives/systemd-devel/2014-August/021982.html

Closes https://github.com/NixOS/nix/issues/2803


edolstra:
It's probably better to do this in Nix itself rather than in the systemd service so we get predictable behaviour regardless of how Nix is executed. For reference, we used to ignore SIGPIPE and restore it in builders (see 7a65b2470eb53a320749d76746fbf65790183d9d, probably we want to bring back setting SIGPIPE to SIG_DFL in builders).

PR #6415, opened 2022-04-17
doc: document nix.conf connect-timeout default

aakropotkin:
Add note to `nix.conf` documentation and help message for `connect-timeout`, indicating the significance of the default value `0`.


src/libstore/filetransfer.hh
```diff
@@ -31,7 +31,7 @@ struct FileTransferSettings : Config
         R"(
           The timeout (in seconds) for establishing connections in the
           binary cache substituter. It corresponds to `curl`’s
-          `--connect-timeout` option.
+          `--connect-timeout` option. The default 0 means no limit.
```
cole-h, line 34:
```suggestion
          `--connect-timeout` option. A value of 0 means no limit.
```

PR #6416, opened 2022-04-18
feat: include openssh in docker image

Yajo:
When leveraging remote builders or cache in CI workloads, sometimes you need to configure nix to connect via SSH to a remote server.

It is the case for example when using nixbuild.net.

By including `openssh` package, CI should be able to reach remote builders when configured i.e. with environment variables.

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -22,6 +22,7 @@ let
     findutils
     iana-etc
     git
+    openssh
   ];
 
   users = {


edolstra:
How much does this increase the size of the Docker image?

Yajo:
I'd say that nothing. It should be already present as a requirement of git of some other dependency.

See these logs:

```
> podman run --rm docker.io/nixos/nix:latest nix-env -iA nixpkgs.openssh
installing 'openssh-8.6p1'
building '/nix/store/i6qcymrnl3ddzdb818syiijb9yjfc8s2-user-environment.drv'...
```

Doesn't seem to fetch anything new.

PR #6419, opened 2022-04-18
nix repl: make symlinks with the :bl command 

ckiee:
Make a new `nix repl` command build-'n-link (`:bl`) command that creates GC root symlinks
in the working directory after building.

-----

Requested by ppepino [on the Matrix](https://matrix.to/#/!KqkRjyTEzAGRiZFBYT:nixos.org/$Tb32BS3rVE2BSULAX4sPm0h6CDewX2hClOTGzTC7gwM?via=nixos.org&via=matrix.org&via=nixos.dev).

~~I didn't choose to make it toggleable since yet another command seems
bloaty, so it's just *there* which might actually be a bit annoying, but
I guess I'll see what other people think (^:~~

```
ckie@cookiemonster ~/git/nix -> ./outputs/out/bin/nix repl
Welcome to Nix 2.6.0. Type :? for help.

nix-repl> :l <nixpkgs>
Added 16118 variables.

nix-repl> :b runCommand "hello" {} "echo hi > $out"

This derivation produced the following outputs:
  ./repl-result-out -> /nix/store/kidqq2acdpi05c4a9mlbg2baikmzik44-hello
[1 built, 0.0 MiB DL]
ckie@cookiemonster ~/git/nix -> cat ./repl-result-out
hi
```

**Release Notes**

*Done*.

**Testing**

*N/A*.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ perl/Makefile.config
 /tests/shell.drv
 /tests/config.nix
 /tests/ca/config.nix
+/tests/repl-result-out
 
 # /tests/lang/
 /tests/lang/*.out
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -1,1 +1,4 @@
 # Release X.Y (202?-??-??)
+
+* `nix repl` has a new build-'n-link (`:bl`) command that builds a derivation
+  while creating GC root symlinks.
--- a/src/nix/repl.cc
+++ b/src/nix/repl.cc
@@ -33,6 +33,7 @@ extern "C" {
 #include "command.hh"
 #include "finally.hh"
 #include "markdown.hh"
+#include "local-fs-store.hh"
 
 #if HAVE_BOEHMGC
 #define GC_INCLUDE_NEW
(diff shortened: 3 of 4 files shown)


peperunas:
Fantastic, thanks!

ckiee:
Tests failing, gonna go chase that.

pennae:
would much prefer if this was a separate command (`:bl`?), we use `:b` to test little things quite a lot and would prefer it not litter the directory we happen to start the repl in with gcroots

Ericson2314:
I would also want this to be deduplicate with the other symlink root code, even though none of it is that complex.

ckiee:
> I would also want this to be deduplicate with the other symlink root code, even though none of it is that complex.

Where should I put it? (assuming I actually finish this PR, suddenly not motivated to work on it anymore for now)

edolstra:
I don't think deduplication is very useful here, since it will actually make the code bigger rather than smaller.

I agree adding a new command (`:bl`) is better.

ckiee:
Ugh, more conflicts to resolve..

ckiee:
@edolstra The tests aren't reproducible anymore since `tests/ca/repl.sh` sometimes runs at the same time as `tests/repl.sh`, both of which now make GC roots. Do you know how I can run them in parallel?

edolstra:
@ckiee "cd" to `$TEST_ROOT` before running the repl. Every test has its own `$TEST_ROOT` so they won't collide.

PR #6422, opened 2022-04-19
Don't log spam when all remote builders are busy

### LINKED ISSUE #6263: "Failed to find a machine for remote build" is still overly verbose
**Describe the bug**

59276244738ebec32ed0593ee4af615852ff564d reduced the verbosity to `lvlChatty` when `canBuildLocally` is true. However, `canBuildLocally` implies `amWilling`, which means the message will still be printed as a warning once `maxJobs` is hit. This is likely expected if `maxJobs` is zero, but for higher `maxJobs` means that there's still a large amount of noise in the logs.

**Steps To Reproduce**

1. Be on x86_64-linux
2. Add an aarch64-linux builder
3. Run a large build (leading to more parallel derivations than maxJobs)

**Expected behavior**

I expected no warning to be printed.

```
nix-env (Nix) 2.5.0pre20211206_d1aaa7e
```

**Additional context**

N/A

SuperSandro2000:
Closes #6263

### DIFF
--- a/src/build-remote/build-remote.cc
+++ b/src/build-remote/build-remote.cc
@@ -121,6 +121,7 @@ static int main_build_remote(int argc, char * * argv)
             mkdir(currentLoad.c_str(), 0777);
 
             while (true) {
+search:;
                 bestSlotLock = -1;
                 AutoCloseFD lock = openLockFile(currentLoad + "/main-lock", true);
                 lockFile(lock.get(), ltWrite, true);
@@ -154,7 +155,7 @@ static int main_build_remote(int argc, char * * argv)
                             }
                         }
                         if (!free) {
-                            continue;
+                            goto search;
                         }
                         bool best = false;
                         if (!bestSlotLock) {


src/build-remote/build-remote.cc
```diff
@@ -154,7 +155,7 @@ static int main_build_remote(int argc, char * * argv)
                             }
                         }
                         if (!free) {
-                            continue;
+                            goto search;
```
pennae, line 158: doesn't this cause a busy wait for a free builder?

SuperSandro2000: Yeah, I think now only one remote builder would get jobs.

thufschmitt:
I might be wrong, but I don’t think that’s the proper solution. What this will do is that each time it will encounter a “full” builder it will restart the search at the begining. Which means that it will essentially only use the very first compatible one from the list (and as @pennae mentionned, it will cause a busy loop endlessly trying to reach it). I’m not sure what the correct fix would be, but I suspect that it’s only a matter of tweaking when the error message is shown

SuperSandro2000:
> I suspect that it’s only a matter of tweaking when the error message is shown

Do you have an idea when it should be shown?

thufschmitt:
> Do you have an idea when it should be shown?

Nope :)

Looking at the code, there’s nothing obviously wrong. I’ll try to dig deeper once I can reproduce it properly

pennae:
it always happens for us when building with `-j0`. we'd suspect it's suffcient to ignore `canBuildLocally` when determining the log level and just using lvlWarn all the time?

PR #6423, opened 2022-04-19
libstore: enable fuse in a sandbox

cab404:
WIP, not yet tested

I need fuse in a build sandbox, so I can merge build layers (just like docker does) in multi-layer builds.
This will open path to easier stage caching, and allow for disk space optimizations in many cases.

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1666,6 +1666,7 @@ void LocalDerivationGoal::runChild()
                 ss.push_back("/dev/null");
                 ss.push_back("/dev/random");
                 ss.push_back("/dev/tty");
+                ss.push_back("/dev/fuse");
                 ss.push_back("/dev/urandom");
                 ss.push_back("/dev/zero");
                 createSymlink("/proc/self/fd", chrootRootDir + "/dev/fd");
--- a/src/libstore/sandbox-defaults.sb
+++ b/src/libstore/sandbox-defaults.sb
@@ -69,6 +69,7 @@
        (literal "/dev/stdin")
        (literal "/dev/stdout")
        (literal "/dev/tty")
+       (literal "/dev/fuse")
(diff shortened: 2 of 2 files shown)


edolstra:
Maybe this should be conditional on the "kvm" system feature, i.e. you would have to write `requiredSystemFeatures = [ "kvm" ]` to get `/dev/kvm`. That makes the behaviour more predictable.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -1666,6 +1666,7 @@ void LocalDerivationGoal::runChild()
                 ss.push_back("/dev/null");
                 ss.push_back("/dev/random");
                 ss.push_back("/dev/tty");
+                ss.push_back("/dev/fuse");
```
edolstra, line 1669: Note that this causes a build failure if `/dev/fuse` doesn't exist on the host filesystem.

dtzWill:
@edolstra this sounds good to me, although I think you meant fuse instead of kvm -- aka are suggesting this be hidden by a new "fuse" feature and that it acts like "kvm" presently does, but for /dev/fuse-- including checking if the device exists first?

edolstra:
Yeah, I meant fuse. In fact it's more important that fuse is conditional on a required system feature, since unlike with kvm there is no fallback if it's not available.

ShamrockLee:
Any progress on this PR?

It would make it possible to run Apptainer containers inside`runCommand` without `runInLinuxVM`.

ShamrockLee:
Any progress on this PR?

It would make it possible to run Apptainer containers inside`runCommand` without `runInLinuxVM`.

cab404:
> Any progress on this PR?
> 
> It would make it possible to run Apptainer containers inside`runCommand` without `runInLinuxVM`.

sadly, not :(
I didn't touch it for a long time now

tomberek:

PR #6434, opened 2022-04-21
docker: more params: extraPkgs, (trusted)Substituters and extraEnv

jlamur:
This allows for users to build nix docker images tuned for their substituters, ca and custom packages.

For example, I would use it this way:
```nix
{ nixpkgs }:

let
  nixRepo = ./nix;
  # nixRepo = builtins.fetchTarball { ... };
  proxyChannelURL = "https://github.proxy/channel-nixos-21.11.tar.gz";
  substituterURL = "https://nixos-cache.proxy";
  myCa = ./ca.crt;
in 
(import (nixRepo + "/docker.nix") {
  pkgs = nixpkgs;
  channelURL = proxyChannelURL;
  extraPkgs = [ nixpkgs.nixops ];
  extraEnv = [
    "SSL_CERT_FILE=${myCa}"
    "GIT_SSL_CAINFO=${myCa}"
    "NIX_SSL_CERT_FILE=${myCa}"
  ];
  substituters = [ substituterURL ];
  trustedSubstituters = [ substituterURL ];
})
```

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -4,6 +4,7 @@
 , tag ? "latest"
 , channelName ? "nixpkgs"
 , channelURL ? "https://nixos.org/channels/nixpkgs-unstable"
+, extraPkgs ? []
 }:
 let
   defaultPkgs = with pkgs; [
@@ -23,7 +24,7 @@ let
     iana-etc
     git
     openssh
-  ];
+  ] ++ extraPkgs;
 
   users = {
 
--- a/docker.nix
+++ b/docker.nix
@@ -5,6 +5,8 @@
 , channelName ? "nixpkgs"
 , channelURL ? "https://nixos.org/channels/nixpkgs-unstable"
 , extraPkgs ? []
+, substituters ? []
+, trustedSubstituters ? []
 }:
 let
   defaultPkgs = with pkgs; [
@@ -126,8 +128,12 @@ let
     sandbox = "false";
     build-users-group = "nixbld";
     trusted-public-keys = "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=";
-  };
-  nixConfContents = (lib.concatStringsSep "\n" (lib.mapAttrsFlatten (n: v: "${n} = ${v}") nixConf)) + "\n";
+  } // (if substituters == [] then {} else {
+    substituters = substituters;
+  }) // (if trustedSubstituters == [] then {} else {
+    trusted-substituters = trustedSubstituters;
+  });
+  nixConfContents = (lib.concatStringsSep "\n" (lib.mapAttrsFlatten (n: v: "${n} = ${toString v}") nixConf)) + "\n";
 
   baseSystem =
     let
--- a/docker.nix
+++ b/docker.nix
@@ -7,6 +7,7 @@
 , extraPkgs ? []
 , substituters ? []
 , trustedSubstituters ? []
+, extraEnv ? []
 }:
 let
   defaultPkgs = with pkgs; [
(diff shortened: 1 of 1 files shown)


thufschmitt:
Isn’t already exposed one way or another by `dockerTools`? With some kind of `override` or `overrideAttrs` function

jlamur:
I am quite new to nix & nixpkgs but I am pretty sure that it is impossible as most of these parameters are defined / used in intermediate variables (in `let` scope). As I said I am new so I may have missed a possibility.

Also, the certs override does not work in some CRI (an "old" one from a ubuntu repo on one of my servers fails to compute env variables correctly if they are duplicated). So I ended up re-creating an image from scratch for my own needs, it works pretty well for now. If that helps someone, it looks like this:
```nix
nixpkgs.dockerTools.buildLayeredImageWithNixDb {
  name = "nix";
  tag = "latest";

  contents = [
    pkgs.bashInteractive
    pkgs.coreutils-full
    pkgs.curl
    pkgs.git
    pkgs.nix
    pkgs.openssh
    pkgs.vim
    (pkgs.cacert.override {
      extraCertificateFiles = [ certificates.all ];
    })
    (pkgs.writeTextFile {
      name = "nix.conf";
      text = ''
        build-users-group =
        sandbox = false
        substituters = ${proxies.nixosCache}
        trusted-substituters = ${proxies.nixosCache}
      '';
      destination = "/etc/nix/nix.conf";
    })
    (pkgs.writeTextFile {
      name = ".nix-channels";
      text = "${proxies.nixpkgs "nixos-${version}"} nixpkgs";
      destination = "/.nix-channels";
    })
  ];

  extraCommands = ''
    mkdir -p tmp var/tmp
    chmod 1777 tmp var/tmp
  '';

  config = {
    Cmd = [ "bash" ];
    Env = [
      "USER=root"
      "PATH=/.nix-profile/bin:/bin"
      "NIX_PATH=/.nix-defexpr/channels"
      "SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt"
      "GIT_SSL_CAINFO=/etc/ssl/certs/ca-bundle.crt"
      "NIX_SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt"
    ];
  };
}
```

I am pretty sure the lack of a shadow & a passwd file may cause problems in some use-cases / for some softwares, but for my own needs (mainly `nix-build` / `nix build`) it works and is way easier to maintain.

thufschmitt:
> it is impossible as most of these parameters are defined / used in intermediate variables (in let scope).

Oh, that’s annoying :/

I’ll defer this to @garbas since he’s the author of this file. I’m not too fond of adding these kind of ad-hoc parameters but maybe he has a better solution to this issue

docker.nix
```diff
@@ -125,8 +129,12 @@ let
     sandbox = "false";
     build-users-group = "nixbld";
     trusted-public-keys = "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=";
-  };
-  nixConfContents = (lib.concatStringsSep "\n" (lib.mapAttrsFlatten (n: v: "${n} = ${v}") nixConf)) + "\n";
+  } // (if substituters == [] then {} else {
```
garbas, line 132: I think it would be better to extend `nix.conf` via some more general `extraNixConf` variable (as simple string) instead of having specialized arguments for each option in `nix.conf`.

garbas (changes requested):
@jlamur If you look at the initial PR when we added this file (`docker.nix`) it was said that this functionality should be actually moved to `nixpkgs`. Therefore we didn't really work on improving it. 

I left one comment, but generally this PR looks like improvement and I would merge once the comment gets addressed.

But if you have time I would love if you can move it to `nixpkgs`.

Thank you @jlamur for working on this. :rocket:

roberth:
> it was said that this functionality should be actually moved to nixpkgs.

https://github.com/NixOS/nixpkgs/pull/148456 demonstrates the refactoring of NixOS modules so that they can efficiently and accurately construct a "partial" configuration that is a container. (as well as be used in "complete" NixOS systems)
I believe this is the direction we should be taking, to avoid creating ever more interfaces that people have to learn (or more realistically, look up and translate).

tomberek:

Issue #5211, opened 2021-09-03
`builtins.eval`: like `import` but from strings


kvtb:
It would be nice to expose `state.parseExprFromString` as `builtins.eval`, similar to `builtins.import` but accepting string instead of file name.
One use case is pre-process importing files in Nix,  replacing `builtins.import path` with  `builtins.eval (f (builtins.readFile path))`
Another is working with files which might be modified during evaluation, avoiding caching of `builtins.import` 
Third is to keep configuration in Nix format , avoiding Json, Toml,  etc which are contaminating Nix ecosystem too (no wonder, Nix can do `builtins.fromJSON` and `builtins.fromTOML`)

L-as:
`builtins.eval = x: import (builtins.toFile "source" x)`?

PR #6436, opened 2022-04-21
fix: builtins.toFile adds path to allowedPaths

tomberek:
The produced path is then allowed be imported or utilized elsewhere:
```
assert (43 == import (builtins.toFile "source" "43")); "good"
```

This will still fail on read-only stores.

Can be used to implement "eval": Fixes https://github.com/NixOS/nix/issues/5211

Potential concerns:
- [ ] read-only stores
- [ ] distributed builds
- [ ] `--store /some/path` usage
- [ ] eval-store
- [ ] Hydra?
- [ ] interaction with lazy copy to store PR?

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -2,3 +2,7 @@
 
 * `nix repl` has a new build-'n-link (`:bl`) command that builds a derivation
   while creating GC root symlinks.
+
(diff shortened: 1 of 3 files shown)


thufschmitt (approved):
Thanks :)

As for your various concerns, they are essentially the same that all the builtins fetchers share, so it should be all right (although I’d love to see at least some of these tested − including for the other fetchers − but that’s orthogonal to this PR)

PR #6438, opened 2022-04-22
rename SymbolIdx -> Symbol, Symbol -> SymbolStr

pennae:
after https://github.com/NixOS/nix/pull/6218 `Symbol` no longer confers a uniqueness invariant on the
string it wraps, it is now possible to create multiple symbols that
compare equal but whose string contents have different addresses. this
guarantee is now only provided by `SymbolIdx`, leaving `Symbol` only as
a string wrapper that knows about the intricacies of how symbols need to
be formatted for output.

this change renames `SymbolIdx` to `Symbol` to restore the previous
semantics of `Symbol` to that name. we also keep the wrapper type and
rename it to `SymbolStr` instead of returning plain strings from lookups
into the symbol table because symbols are formatted for output in many
places. theoretically we do not need `SymbolStr`, only a function that
formats a string for output as a symbol, but having to wrap every symbol
that appears in a message into eg `formatSymbol()` is error-prone and
inconvient.

### DIFF
--- a/src/libexpr/symbol-table.hh
+++ b/src/libexpr/symbol-table.hh
@@ -12,82 +12,94 @@ namespace nix {
 /* Symbol table used by the parser and evaluator to represent and look
    up identifiers and attributes efficiently.  SymbolTable::create()
    converts a string into a symbol.  Symbols have the property that
-   they can be compared efficiently (using a pointer equality test),
+   they can be compared efficiently (using an equality test),
    because the symbol table stores only one copy of each string. */
 
-class Symbol
+/* This class mainly exists to give us an operator<< for ostreams. We could also
+   return plain strings from SymbolTable, but then we'd have to wrap every
+   instance of a symbol that is fmt()ed, which is inconvenient and error-prone. */
+class SymbolStr
 {
     friend class SymbolTable;
+
 private:
-    std::string s;
+    const std::string * s;
+
+    explicit SymbolStr(const std::string & symbol): s(&symbol) {}
 
 public:
-    Symbol(std::string_view s) : s(s) { }
-
-    // FIXME: remove
     bool operator == (std::string_view s2) const
     {
-        return s == s2;
+        return *s == s2;
     }
 
     operator const std::string & () const
     {
-        return s;
+        return *s;
     }
 
     operator const std::string_view () const
     {
-        return s;
+        return *s;
     }
 
-    friend std::ostream & operator << (std::ostream & str, const Symbol & sym);
+    friend std::ostream & operator <<(std::ostream & os, const SymbolStr & symbol);
 };
 
-class SymbolIdx
+class Symbol
 {
     friend class SymbolTable;
 
 private:
     uint32_t id;
 
-    explicit SymbolIdx(uint32_t id): id(id) {}
+    explicit Symbol(uint32_t id): id(id) {}
 
 public:
-    SymbolIdx() : id(0) {}
+    Symbol() : id(0) {}
 
     explicit operator bool() const { return id > 0; }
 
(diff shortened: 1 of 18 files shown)


Ericson2314:
This is sort of silly, but my first questions are about the commit message :).

I would markdown it,

> `Symbol` no longer guarantees pointer uniqueness of the contained string, only
> `SymbolIdx` does. This can be somewhat confusing, so we'll rename `SymbolIdx` to
> `Symbol` to restore uniqueness guarantees to the `Symbol` type.

I would add paragraph break here

> We also keep around
> the previous `Symbol` type (now called `SymbolStr`) because there are many instances
> of symbols being passed to `fmt`, which in turn uses the `operator <<` of the
> symbol. Removing the wrapper would require wrapping all symbols that are
> currently passed to `operator <<`, but this would be very error-prone and probably
> also surprising.

What is the wrapper? Removing the wrapper would require adding the wrapper? Huh?

> it also seems that the `Symbol` uniqueness guarantee was violated by the eval
> cache, though in a nondestructive way: ...

Before or just after your previous PR?

This being C++, "non-destructive" makes me think this something about mutation / in place update? but I think you merely meant "though in way that didn't yet cause problems:". Is that right?

> ...`AttrKey` contained a `Symbol` that could be
> directly constructed from a `std::string_view` instead of going through the symbol
> table. This was never a problem because this symbol would only be used for
> implicit conversions to string.

Now, I have been letting others do libexpr stuff largely, as I focus on libstore, so maybe if I was up to speed this all would be clear from context, but perhaps good to clear up on the message for people "paged out" like me anyways.

pennae:
looks like the message needs a good rewording. :sweat_smile: 

> What is the wrapper? Removing the wrapper would require adding the wrapper? Huh?

ugh, that part is really unclear. removing the previously-`Symbol` wrapper type and returning strings from the symbol table instead means we have to add eg `formatSymbol` calls around all symbols (strings) that are passed to `fmt()`. it's better to switch around the names to keep `Symbol` with an implied uniqueness, and now-`SymbolStr` as the thing that knows how a symbol should be formatted for output

> Before or just after your previous PR?

after. looks like it was fine before. "non-destructive" here means that while the guarantee that was once attached to the name didn't hold nothing bad would've happened.

Ericson2314:
Mind updating the PR description likewise? (I really wish they would automatically stay sync for single-commit PRs!)

pennae:
sure, done :)

src/libexpr/symbol-table.hh
```diff
@@ -12,82 +12,92 @@ namespace nix {
...
-class Symbol
+/* This class mainly exists to give us an operator<< for ostreams. We could also
+   return plain strings from SymbolTable, but then we'd have to wrap every
+   instance of a symbol that is fmt()ed, which is inconvenient and error-prone. */
+class SymbolStr
 {
-    friend class SymbolTable;
 private:
-    std::string s;
+    const std::string * s;
```
thufschmitt, line 24: Any reason for this change?

thufschmitt (commented):
Looks good, I’m just curious about the why of the new pointer indirection in the (new) `Symbol` class

src/libexpr/symbol-table.hh
thufschmitt: Oh I see, thanks.

The constructor below feels a bit dangerous now (it seems that we don’t construct a `SymbolStr` outside of a `SymbolTable` atm, but it would be tempting to do something like `Symbol blah("somestring")` which would lead to a dangling pointer. Any objection against either making it private or making it take a `std::string *` as argument to make it more explicit that it’s expecting the string to stay alive?

PR #6440, opened 2022-04-23
libfetchers/git: hardcode `--git-dir`

Ma27:
To demonstrate the problem:

* You need a `git` at 2.33.3 in your $PATH
* An expression like this in a git repository:

  ``` nix
  {
    outputs = { self, nixpkgs }: {
      packages.foo.x86_64-linux = with nixpkgs.legacyPackages.x86_64-linux;
        runCommand "snens" { } ''
          echo ${(builtins.fetchGit ./.).lastModifiedDate} > $out
        '';
    };
  }
  ```

Now, when instantiating the package via `builtins.getFlake`, it fails on
Nix 2.7 like this:

    $ nix-instantiate -E '(builtins.getFlake "'"$(pwd)"'").packages.foo.x86_64-linux'
    fatal: unsafe repository ('/nix/store/a7j3125km4h8l0p71q6ssfkxamfh5d61-source' is owned by someone else)
    To add an exception for this directory, call:

    	git config --global --add safe.directory /nix/store/a7j3125km4h8l0p71q6ssfkxamfh5d61-source

This breaks e.g. `nixops`-deployments using flakes with similar
expressions as shown above.

The cause for this is that `git(1)` tries to find the highest
`.git`-directory in the directory tree and if it finds a such a
directory, but with another owning user (root vs. the user who evaluates
the expression), it fails as above. This was changed recently to fix
CVE-2022-24765[1].

By explicitly specifying `--git-dir`, Git assumes to be in the top-level
directory and doesn't attempt to look for a `.git`-directory in the
parent directories and thus the code-path leading to said error is never
reached.

[1] https://lore.kernel.org/git/xmqqv8veb5i6.fsf@gitster.g/

cc @edolstra @thufschmitt @Ericson2314

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -23,7 +23,7 @@ const std::string gitInitialBranch = "__nix_dummy_branch";
 
 static std::string readHead(const Path & path)
 {
-    return chomp(runProgram("git", true, { "-C", path, "rev-parse", "--abbrev-ref", "HEAD" }));
+    return chomp(runProgram("git", true, { "-C", path, "--git-dir", ".git", "rev-parse", "--abbrev-ref", "HEAD" }));
 }
 
 static bool isNotDotGitDirectory(const Path & path)
@@ -152,11 +152,11 @@ struct GitInputScheme : InputScheme
         assert(sourcePath);
 
         runProgram("git", true,
-            { "-C", *sourcePath, "add", "--force", "--intent-to-add", "--", std::string(file) });
+            { "-C", *sourcePath, "--git-dir", ".git", "add", "--force", "--intent-to-add", "--", std::string(file) });
 
         if (commitMsg)
             runProgram("git", true,
-                { "-C", *sourcePath, "commit", std::string(file), "-m", *commitMsg });
+                { "-C", *sourcePath, "--git-dir", ".git", "commit", std::string(file), "-m", *commitMsg });
     }
 
     std::pair<bool, std::string> getActualUrl(const Input & input) const
@@ -259,7 +259,7 @@ struct GitInputScheme : InputScheme
                 if (hasHead) {
                     // Using git diff is preferrable over lower-level operations here,
                     // because its conceptually simpler and we only need the exit code anyways.
-                    auto gitDiffOpts = Strings({ "-C", actualUrl, "diff", "HEAD", "--quiet"});
+                    auto gitDiffOpts = Strings({ "-C", actualUrl, "--git-dir", ".git", "diff", "HEAD", "--quiet"});
                     if (!submodules) {
                         // Changes in submodules should only make the tree dirty
                         // when those submodules will be copied as well.
@@ -284,7 +284,7 @@ struct GitInputScheme : InputScheme
                 if (fetchSettings.warnDirty)
                     warn("Git tree '%s' is dirty", actualUrl);
 
-                auto gitOpts = Strings({ "-C", actualUrl, "ls-files", "-z" });
+                auto gitOpts = Strings({ "-C", actualUrl, "--git-dir", ".git", "ls-files", "-z" });
                 if (submodules)
                     gitOpts.emplace_back("--recurse-submodules");
 
@@ -314,7 +314,7 @@ struct GitInputScheme : InputScheme
                 // modified dirty file?
                 input.attrs.insert_or_assign(
                     "lastModified",
-                    hasHead ? std::stoull(runProgram("git", true, { "-C", actualPath, "log", "-1", "--format=%ct", "--no-show-signature", "HEAD" })) : 0);
+                    hasHead ? std::stoull(runProgram("git", true, { "-C", actualPath, "--git-dir", ".git", "log", "-1", "--format=%ct", "--no-show-signature", "HEAD" })) : 0);
 
                 return {std::move(storePath), input};
             }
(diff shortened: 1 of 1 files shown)


Ma27:
Note: this doesn't seem to fix `sudo nixos-rebuild switch --flake .#`. Workaround for that is `sudo nixos-rebuild switch --flake path:.#`. Any ideas on how to fix that?

bjornfor:
Related? https://github.com/NixOS/nix/issues/6443

Ma27:
yes :)

Ma27:
With the latest commit, `nixos-rebuild` works for me as well.

thufschmitt (changes requested):
Looks good, thanks :)

Since we’re adding more complexity to these `git` calls, do you think you could factor that out? Having one `runGit(Path path, std::vector<std::string> args)` command that would morally run `runProgram("git", true, { "-C", path, "--git-dir", getGitDir(), ..args}` (or something along those lines). Not sure whether the git calls are similar enough to warrant it, but it’s annoying to have all these calls with some very similar arguments everywhere.

Also, it might be a bit silly right now since the Git used in the flake doesn’t have the patch, but I’d love to see a regression test for this. Not sure how we could do it without having to resort to a costly vm test though since it involves medling with the permissions

Ma27:
> Also, it might be a bit silly right now since the Git used in the flake doesn’t have the patch, but I’d love to see a regression test for this. Not sure how we could do it without having to resort to a costly vm test though since it involves medling with the permissions

We could do that by updating the nixpkgs pin in this repo I guess.

Ma27:
> Since we’re adding more complexity to these git calls, do you think you could factor that out? Having one runGit(Path path, std::vector<std::string> args) command that would morally run runProgram("git", true, { "-C", path, "--git-dir", getGitDir(), ..args} (or something along those lines). Not sure whether the git calls are similar enough to warrant it, but it’s annoying to have all these calls with some very similar arguments everywhere.

The problem is, all of those calls are slightly different: e.g. different `RunOptions`, potentially a different `gitDir` (because of `repoDir = cacheDir` etc.). We could either factor out the creation of `{ "-C", repoDir, ... }` into a function (but that doesn't seem to be a benefit) or we could move the entire `runProgram`-call out, but then we'd have to take care of edge-case option for `runProgram` which also doesn't seem to be that beneficial.

@thufschmitt  In other words, I'd prefer to see this merged considering that it is an issue on unstable and 21.11 and would unblock a few people I guess.

thufschmitt (approved):
> @thufschmitt In other words, I'd prefer to see this merged considering that it is an issue on unstable and 21.11 and would unblock a few people I guess.

Fair point. I feared that trying to factor this out would make everything a bit awkward. Let’s merge this as-it-is then, I’ll just open an issue about the test

thufschmitt:
> @thufschmitt could you please add backport labels to 2.7 and 2.8

Both failed :unamused: 

I’ve manually created a backport PR for 2.8 (not sure why the bot failed). Do you think you can create one for 2.7 (there’s a small merge conflict) ?

src/libfetchers/git.cc
```diff
@@ -21,6 +21,15 @@ namespace nix::fetchers {
...
+    auto gitDir = getEnv("GIT_DIR");
+    if (!gitDir) {
+        return ".git";
+    }
+    return *gitDir;
+}
+
 static std::string readHead(const Path & path)
 {
     return chomp(runProgram("git", true, { "-C", path, "--git-dir", ".git", "rev-parse", "--abbrev-ref", "HEAD" }));
```
alexshpilkin, line 35: Should this have used getGitDir() as well?

alexshpilkin, line 35: Is there a reason why this shouldn’t have been changed as well?

```suggestion
    return chomp(runProgram("git", true, { "-C", path, "--git-dir", getGitDir(), "rev-parse", "--abbrev-ref", "HEAD" }));
```

edolstra:
@Ma27 I don't understand this part:
```
     auto gitDir = getEnv("GIT_DIR");
```
How does it make sense to have a global environment variable that overrides the `.git` directory *for all repositories*?

Ma27:
@edolstra that's a variable from Git itself. If it's declared, `.git` is named differently, so `--git-dir=.git` doesn't work.

edolstra:
I know that, but why does it make sense for Nix to use that variable? Git deals with only one repo, whereas Nix's Git fetcher deals with potentially multiple local repos, all of which may have a different `GIT_DIR` setting. If we really want to support this use case, then it should probably be a `git` input attribute. (And to be honest, overriding `.git` seems like one of those "just don't do that" things that only complicate the code...)

thufschmitt:
I confess didn’t think about it too much and I assumed that it was here for the `git+file` case (where we might indeed need to honor it since it’s a property of the workspace).

But indeed, even in that case it shouldn’t be blindly picked from the environment − and in any other case we probably shouldn’t respect it since the git repos are fully handled by Nix (so unless I’m missing something, there’s no good reason to use something else than the default `.git`)

Ma27:
> But indeed, even in that case it shouldn’t be blindly picked from the environment

Yeah thinking a bit more about this and I guess it may cause more problems than it solves, so we should replace `auto gitDir = getGitDir();` with `auto gitDir = ".git";` (the variable is still important because of `gitDir = ".";` for cached repos).

Ma27:
Followup in https://github.com/NixOS/nix/pull/6567 @edolstra :)

PR #6444, opened 2022-04-24
feat: add integration with zsh's run-help

midchildan:
run-help is a zsh widget that can be bound to a key. It mainly looks up the man page for the currently typed in command.

Although run-help works for any command without requiring special support, it can only deduce the right man page based solely on the name of the command. This change would allow Nix to provide better integration with run-help by helping zsh identify Nix subcommands and their corresponding man pages.

To actually use run-help on zsh, place the following lines in ~/.zshrc:

    (( $+aliases[run-help] )) && unalias run-help
    autoload -Uz run-help run-help-nix

Then assign run-help to any key of choice:

    bindkey '^[h' run-help

### DIFF
--- a/misc/zsh/run-help-nix
+++ b/misc/zsh/run-help-nix
@@ -1,0 +1,97 @@
+emulate -L zsh
+
+# run-help is a zsh widget that can be bound to a key. It mainly looks up the
+# man page for the currently typed in command.
+#
+# Although run-help works for any command without requiring special support,
+# it can only deduce the right man page based solely on the name of the
+# command. Programs like Nix provide better integration with run-help by
+# helping zsh identify Nix subcommands and their corresponding man pages. This
+# is what this function does.
+#
+# To actually use run-help on zsh, place the following lines in your .zshrc:
+#
+#     (( $+aliases[run-help] )) && unalias run-help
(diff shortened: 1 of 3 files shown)


misc/zsh/run-help-nix
```diff
@@ -0,0 +1,97 @@
...
+fi
+
+while [[ "$#" != 0 && "$1" == -* ]]; do
+  shift
+done
+
+case "$1" in
+  flake)
+    case "$2" in
+      archive|check|clone|info|init|lock|metadata|new|prefetch|show|update)
```
edolstra, line 33: These lists of commands will get out of sync quickly. Maybe they can be generated at build-time, e.g. using `nix __dump-args | jq '.commands | keys'`. Alternatively, it could use the completion code, e.g.

```
$ NIX_GET_COMPLETIONS=2 nix flake '' | tail +2
archive
check
clone
info
init
lock
metadata
new
prefetch
show
update
```

midchildan: I rewrote it to make it match subcommands based on `man -w` instead. This should work for future commands as long as its man pages are named in the form of `nix3-subcomands-joined-by-hypens`.

PR #6449, opened 2022-04-26
Allow selecting derivation outputs

edolstra:
This allows selecting the outputs of a derivation `nix` should operate on using the syntax `installable^outputs`, e.g. `nixpkgs#glibc^dev,static` or `nixpkgs#glibc^*`. By default, `nix` will use the outputs specified by the derivation's  `meta.outputsToInstall` attribute if it exists, or all outputs otherwise.

Depends on #6426.

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -14,3 +14,13 @@
 
 * `nix build` has a new `--print-out-paths` flag to print the resulting output paths.
   This matches the default behaviour of `nix-build`.
+
+* You can now specify which outputs of a derivation `nix` should
+  operate on using the syntax `installable^outputs`,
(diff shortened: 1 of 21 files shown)


edolstra:
I've changed the `!` character to `^` to make it easier on shell users.

Ericson2314:
@edolstra you still did not address and instead many of the issues in the other thread, once again.

 - `nixpkgs#glibc.dev` not working is terrible UI
 
 - Syntax is subtly inconsistent/ambiguous given accepted RFC 92

roberth:
It would be useful for this to operate at the expression level, rather than the store level, so that packages can be split across derivations. Here's a use case where it breaks https://github.com/NixOS/nixpkgs/pull/175785#discussion_r888798100

Another use case is for packages that use heavy tooling in their docs output. Splitting those docs out into a separate derivation reduces the build closure size.

See also #6507

Issue #6445, opened 2022-04-25
Nix does not degrade gracefully when a substituter dies


Gabriel439:
**Describe the bug**

If Nix successfully fetches a path from a substituter and then the substituter dies then all subsequent substitutions fail, even if `--fallback` is enabled and even if those substituters could have otherwise succeeded.

**Steps To Reproduce**

I created a NixOS test that reproduces the bug:

```nix
let
  nixpkgs = builtins.fetchTarball {
    url    = "https://github.com/NixOS/nixpkgs/archive/97471880609d786714a3d56fd7246f48144ab94c.tar.gz";
    sha256 = "0phz5fhhi74ydmikaiqf912vv1fbwys0jfrigjgd5nd8ps2ss83c";
  };

  pkgs = import nixpkgs { };

in
  pkgs.nixosTest {
    nodes = rec {
      client = {
        nix = {
          binaryCaches = [ "http://cache0:5000" "http://cache1:5000" ];

          binaryCachePublicKeys  = [
            "test-cache-resilience:Vs6rN2TKxsBNxzo/HWuQdspE90keHyeE3kYocldJv3g="
          ];
        };
      };

      cache0 = { config, ... }: {
        environment.etc."nix-serve/nix-serve.sec".text =
          "test-cache-resilience:Oz75MbefmJck5Yk+LcR3/Ngq33ZewcuNMHLj9Lx2PcpWzqs3ZMrGwE3HOj8da5B2ykT3SR4fJ4TeRihyV0m/eA==";

        networking.firewall.allowedTCPPorts =
          [ config.services.nix-serve.port ];

        services.nix-serve = {
          enable = true;

          secretKeyFile = "/etc/nix-serve/nix-serve.sec";
        };

        system.extraDependencies = [
          pkgs.cowsay
          pkgs.cowsay.drvPath
          pkgs.hello
          pkgs.hello.drvPath
        ];
      };

      cache1 = cache0;
    };

    testScript = ''
      start_all()

      client.succeed(
          "nix-store --realise ${pkgs.cowsay.drvPath}"
      )

      cache0.crash()

      client.succeed(
          "nix-store --fallback --option connect-timeout 1 --realise ${pkgs.hello.drvPath}"
      )
    '';
  }
```

What happens when you run that test is that the first substitution succeeds and the second substitution core dumps:

```
…
[...]

PR #6452, opened 2022-04-27
Remove almost all noexcept annotations in src/libstore

Gabriella439:
Fixes https://github.com/NixOS/nix/issues/6445

The `checkEnabled` function can throw a `SubstituterDisabled`
exception and that function is used by `getFile`, so any
function that transitively uses `getFile` can throw an
exception.  This means that we need to remove the `noexcept`
annotations for most of the methods underneath `src/libstore`
since they are not correct.

Otherwise what happens is that a client will core dump when
a substituter dies because the `SubstituterDisabled` exception
cannot propagate up to the `tryNext` functions and will instead
terminate the program when it his a `noexcept` annotation in the
callstack.

It is possible that not all of these annotations needed to be removed
as there might be a few functions that don't transitively depend on
`getFile`.  However, rather than carefully audit which ones needed
the annotation or not I removed all of the annotations.  My reasoning
was that even if some `noexcept` annotations might be technically
correct right now, they might no longer hold after a refactor.

Also, as noted here:

https://github.com/NixOS/nix/issues/6445#issuecomment-1110780076

… we can rely on the fact that the `Callback::operator()` is marked
`noexcept` anyway.

### DIFF
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -68,7 +68,7 @@ void BinaryCacheStore::upsertFile(const std::string & path,
 }
 
 void BinaryCacheStore::getFile(const std::string & path,
-    Callback<std::optional<std::string>> callback) noexcept
+    Callback<std::optional<std::string>> callback)
 {
     try {
         callback(getFile(path));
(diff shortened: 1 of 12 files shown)


zseri:
But `getFile` itself shouldn't be affected (that is, it's `noexcept` flag is correct), as it passes the exception right back to the callback, right?

zseri:
But `getFile` itself shouldn't be affected (that is, it's `noexcept` flag is correct), as it passes the exception right back to the callback, right?

Gabriella439:
If nobody is going to put up an alternative fix, then can we please merge this?  As far as I can tell the `noexcept` annotations are wrong as the original bug report illustrates and the current behavior is broken as a result

fricklerhandwerk:
Fixed by #6590

PR #6455, opened 2022-04-28
local-derivation-goal.cc: seccomp filter for powerpc on powerpc64

amjoseph-nixpkgs:
This commit should allow 32-bit binaries to execute in sandboxed     builders on big-endian powerpc64.  Unfortunately I am not in a position to test this code due to the fact that all of my powerpc64 machines are running little-endian kernels.

There is no corresponding entry for "powerpcle" on powerpc64le because     seccomp does not support 32-bit little-endian powerpc.

Little-endian 32-bit powerpc is soemthing of an exotic curiosity.     Although most (all?) 32-bit powerpc processors can boot in either     endianness mode, no commerical software system was ever based on the     little-endian variant.  The Linux kernel does have some level of     support for it, mainly due to a few people who had been running Linux     on 32-bit powerpc chips and ran their systems little-endian in order     to get faster emulation of x86_32.

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1529,6 +1529,11 @@ void setupSeccomp()
         seccomp_arch_add(ctx, SCMP_ARCH_ARM) != 0)
         printError("unable to add ARM seccomp architecture; this may result in spurious build failures if running 32-bit ARM processes");
 
+    /* big-endian only because seccomp does not support powerpcle (32-bit) */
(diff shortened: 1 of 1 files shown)


a-m-joseph:
Oh, I see, these `seccomp_arch_add()` calls are only needed in order to enable secondary ABIs, like x86_32 on an x86_64 processor.

Closing this PR as unnecessary.

amjoseph-nixpkgs:
I'm not likely to be in a position to test this anytime soon.

PR #6464, opened 2022-04-29
Add an owner check when searching for the flake root

### LINKED ISSUE #6408: restrict upward search behavior similar to git
Git is making it's searching upward behavior more strict right now due to `CVE-2022-24765` .
See https://github.blog/2022-04-12-git-security-vulnerability-announced/ for more info about that vulnerability.

Their new behavior is to stop traversal at the point where ownership of the directory changes away from the current user.
Searching upward is a tricky feature from a security POV, because as an attacker if
* you expect that some upward-searching command that basically executes code
  - will be run inside the folder `A/B` and
  - not find the file it's looking for inside `A/B` and at the same time
* you can write the file the command is looking for into `A`

That way you can turn being able to write a specific file into code execution.

This is a problem for `git` because while at its core git does not provide code execution as a feature, there is a lot of supplementary stuff like nice git-specific shell prompts that make it happen.

I would expect this issue to actually be worse for nix flakes, where code execution is the whole point of many operations.

_Originally posted by @mschwaig in https://github.com/NixOS/nix/issues/5720#issuecomment-1097870531_

thufschmitt:
# Motivation

When searching up the filesystem for the root of the flake (the directory that contains the `flake.nix`), don’t go anywhere up if we encounter a directory owned by a different user, as otherwise this other user could craft an arbitrary flake, potentially causing bad stuff to happen (shouldn’t in most cases since all it could do is run sandboxed builds, but there’s probably a lot of edge-cases that would make this very undesirable).

This is to fix Nix’s equivalent of CVE-2022-24765

This check is intentionally not applied to the exact directory specified since:
1. It’s up to the user to not point to an untrusted input
2. In multi-user Nix installations, that would prevent from using a flake in the Nix store (since it’s owned by root and not the current user)

# Context

Fix #6408

# Checklist for maintainers

 - [x] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
 - [x] documentation in the internal API docs
 - [x] code and comments are self-explanatory
 - [x] new feature or incompatible change: updated release notes

# Priorities

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -1,3 +1,5 @@
 # Release X.Y (202?-??-??)
 
-- [`nix-channel`](../command-ref/nix-channel.md) now supports a `--list-generations` subcommand
+* When searching upwards for the root of the flake, Nix doesn’t consider
+  anything that’s not owned by the current user anymore as it’s a
+  security hazard.
--- a/flake.nix
+++ b/flake.nix
@@ -598,6 +598,8 @@
           ["i686-linux" "x86_64-linux"]
           (system: runNixOSTestFor system ./tests/nixos/setuid.nix);
 
+        tests.cve-2022-24765 = runNixOSTestFor "x86_64-linux" ./tests/nixos/cve-2022-24765.nix;
+
 
         # Make sure that nix-env still produces the exact same result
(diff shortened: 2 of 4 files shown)


tomberek (approved):

tomberek:

PR #6465, opened 2022-04-29
tests: Distinguish crashes from expected failures

layus:


tests/lang.sh
```diff
@@ -10,32 +10,49 @@ nix-instantiate --show-trace --eval -E 'builtins.addErrorContext "Hello" (throw
 set +x
 
 fail=0
+just_failed=0
```
thufschmitt, line 13: Is this variable useful? Looks like you could directly set `failed` to 1 in `safe_expect` rather than indirecting through `just_failed`

tests/lang.sh
```diff
@@ -10,32 +10,49 @@ nix-instantiate --show-trace --eval -E 'builtins.addErrorContext "Hello" (throw
...
+
+safe_expect () {
```
thufschmitt, line 15: Maybe reuse (and extend?) the existing `expect` function in `tests/common.sh.in` since it does morally the same thing?

It’s marginally less user-friendly as-it-is because it doesn’t allow printing a different error message whether the command crashed or just returned an unexpected result, but I don’t think that’s really a problem

thufschmitt (changes requested):
Thanks, that’s indeed a good thing to do :D 

A couple of suggestions, but nothing really deep

tests/lang.sh
layus: We use it in the last tests to see if we should check the xml. That is why I had to introduce it.

tests/lang.sh
layus: Well, It was not possible at first, but now that I discovered that it always returns 1 or 0, `expect 0` and `expect 1` would work just as well, albeit with less explanations.

layus:
@thufschmitt A new attempt, dutifully based on expect. I am quite surprised that nix only fails with return code 1, but that makes this change way easier.

tests/lang.sh
```diff
@@ -4,6 +4,7 @@ export TEST_VAR=foo # for eval-okay-getenv.nix
 export NIX_REMOTE=dummy://
 
 nix-instantiate --eval -E 'builtins.trace "Hello" 123' 2>&1 | grep -q Hello
+nix-instantiate --eval -E 'builtins.addErrorContext "Hello" 123' 2>&1
```
thufschmitt, line 7: Accidentally committed?

thufschmitt (changes requested):
Looks good ! Just one strange change which I assume is accidental

tests/lang.sh
layus: not at all, the next commands again tests for errors. But such errors could be a crash, and we only care about the grep error. 
Unless the return code of pipelines handles that ?

thufschmitt: Oh you’re right, the tests don’t set `pipefail`. So that makes total sense

thufschmitt (approved):

Issue #6153, opened 2022-02-24
nix install breaks


ikuz:
**Describe the bug**

Tried to (re)install nix on mac os but it fails.

**Steps To Reproduce**

I followed instructions to remove a previous installation of nix: https://nixos.org/manual/nix/stable/installation/installing-binary.html

then did `sh <(curl -L https://nixos.org/nix/install)`

eventually it fails, because it looks like there are already _nixbld users on the system.

Log of the installation process:

```
% sh <(curl -L https://nixos.org/nix/install)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  4046  100  4046    0     0  12394      0 --:--:-- --:--:-- --:--:-- 12394
downloading Nix 2.6.1 binary tarball for aarch64-darwin from 'https://releases.nixos.org/nix/nix-2.6.1/nix-2.6.1-aarch64-darwin.tar.xz' to '/var/folders/6q/14m95v115x9b44w40cncnd8r0000gn/T/nix-binary-tarball-unpack.7ULme1Ndjy'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 8998k  100 8998k    0     0  4714k      0  0:00:01  0:00:01 --:--:-- 4728k
Switching to the Multi-user Installer
Welcome to the Multi-User Nix Installation

This installation tool will set up your computer with the Nix package
manager. This will happen in a few stages:

1. Make sure your computer doesn't already have Nix. If it does, I
   will show you instructions on how to clean up your old install.

2. Show you what I am going to install and where. Then I will ask
   if you are ready to continue.

3. Create the system users and groups that the Nix daemon uses to run
   builds.

4. Perform the basic installation of the Nix files daemon.

5. Configure your shell to import special Nix Profile files, so you
   can use Nix.

6. Start the Nix daemon.

Would you like to see a more detailed list of what I will do?
[y/n] n


---- let's talk about sudo -----------------------------------------------------
This script is going to call sudo a lot. Every time I do, it'll
output exactly what it'll do, and why.

Just like this:

---- sudo execution ------------------------------------------------------------
I am executing:

    $ sudo echo

to demonstrate how our sudo prompts look


This might look scary, but everything can be undone by running just a
[...]

PR #6466, opened 2022-04-29
WIP: installer: allow overriding nix user UID on darwin too

bjornfor:
Similar to f4d57aa490 ("installer: allow overriding nix user GID and
UIDs"), but for darwin.

(There's no corresponding *_GID in this script though, so skipping
that.)

Ref https://github.com/NixOS/nix/issues/6153.

### DIFF
--- a/scripts/install-darwin-multi-user.sh
+++ b/scripts/install-darwin-multi-user.sh
@@ -6,7 +6,7 @@ set -o pipefail
 readonly NIX_DAEMON_DEST=/Library/LaunchDaemons/org.nixos.nix-daemon.plist
 # create by default; set 0 to DIY, use a symlink, etc.
 readonly NIX_VOLUME_CREATE=${NIX_VOLUME_CREATE:-1} # now default
-NIX_FIRST_BUILD_UID="301"
+NIX_FIRST_BUILD_UID="${NIX_FIRST_BUILD_UID:-301}"
 NIX_BUILD_USER_NAME_TEMPLATE="_nixbld%d"
 
 # caution: may update times on / if not run as normal non-root user


abathur:
I don't have much time to look, but I think it may not be this simple. Iirc the variable is already set in the core multiuser script, so I think this would start inheriting that default?

bjornfor:
> I don't have enough time to look, but I think it may not be this simple. Iirc the variable is already set in the core multiuser script, so I think this would start inheriting that default?

You're right, landing this PR now would mean effectively changing the darwin default UID from 301 to 30001.

abathur:
#4531 and #4532 have some backstory on how it ended up at 301

scripts/install-darwin-multi-user.sh
```diff
@@ -6,7 +6,7 @@ set -o pipefail
 readonly NIX_DAEMON_DEST=/Library/LaunchDaemons/org.nixos.nix-daemon.plist
 # create by default; set 0 to DIY, use a symlink, etc.
 readonly NIX_VOLUME_CREATE=${NIX_VOLUME_CREATE:-1} # now default
-NIX_FIRST_BUILD_UID="301"
+NIX_FIRST_BUILD_UID="${NIX_FIRST_BUILD_UID:-301}"
```
bacchanalia, line 9: We can use a different var name to avoid it being overridden by the parent script.

```suggestion
NIX_FIRST_BUILD_UID="${NIX_FIRST_BUILD_UID_DARWIN:-301}"
```

bacchanalia:
I'd really like to get some version of this in. It's been causing me headaches.

bjornfor:
@bacchanalia: if you can, please make a PR.

bacchanalia:
@bjornfor sure. I just wasn't sure if you were interested in reviving this one or not.

scripts/install-darwin-multi-user.sh
abathur: While it is the simpler change, I think the potential-confusion would be lower if we moved the value-setting out into the ~speciated scripts (i.e., require the platform install scripts to be opinionated on this UID instead of having a default-that-isn't-a-default?)

bacchanalia: I'm not sure what you mean.

abathur: Something like:
- replace the default declaration up in the main multiuser script with a comment that says that the speciated ~variant (`-darwin-`, `-linux-`) scripts are responsible for setting it 
- move the default declaration out into the `-linux-` script
- do roughly what this current PR does for the `-darwin-` script

bacchanalia: gotcha. yeah, sounds good.

bjornfor:
Superseded by https://github.com/NixOS/nix/pull/9639.

PR #6469, opened 2022-04-29
Unlock output paths when a derivation is already built

### LINKED ISSUE #6468: nix build does not unlock when a derivation is already built
**Describe the bug**

We've been debugging some deadlocks in our builds, which we believe to be caused by nix builds not releasing locks in a particular case.

The scenario where we see this is as follows: we have multiple single user nix build processes running, building large, overlapping but not necessarily identical derivation graphs.  Occasionally, we find two or more of these processes will deadlock with each other.

After some investigation, we noticed that nix build processes will not drop the locks for derivation goals which have already been built by another process when the current process gets round to building them.

This occurs here in nix 2.3.11 which we're currently using (hoping to upgrade soon!): https://github.com/NixOS/nix/blob/2.3.11/src/libstore/build.cc#L1438-L1443, but the code seems to be similar on master: https://github.com/NixOS/nix/blob/master/src/libstore/build/derivation-goal.cc#L627-L632.

This means the locks are held until the process terminates.  If there are other nix build processes in a similar state, they will also try to acquire the same locks when they try to build the same derivation, and so will wait until the lock holder terminates (which might be a very long time if it has a lot to build). In some pathological cases, those processes might be holding their own locks on other derivations due to the same issue, and this can lead to deadlock.

I suspect the above scenario is not often seen in regular community usage of Nix, where multiple concurrent builds of overlapping build graphs like this are probably less common.

We believe a simple patch to add a call to `outputLocks.unlock()` here: https://github.com/NixOS/nix/blob/master/src/libstore/build/derivation-goal.cc#L630 resolves the problem, although it's hard to reproduce to know for sure.

I will raise a pull request with this change, but I would appreciate it if someone who knows the codebase better could comment on whether this seems reasonable.

`nix-env (Nix) 2.3.11pre20210707_02dd6bb`

gbpdt:
Without this change, nix build processes will not drop the locks for derivation goals
which have already been built by another process when the current process gets
round to building them. This means the locks are held until the process
terminates.

If there are other nix build processes in a similar state, they will also try to
acquire the same locks when they try to build the same derivation, and so will
wait until the lock holder terminates (which might be a very long time if it has
a lot to build). In some pathological cases, those processes might be holding
their own locks on other derivations due to the same issue, and this can lead to
deadlock.

Resolves #6468

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1456,6 +1456,7 @@ void DerivationGoal::done(
     SingleDrvOutputs builtOutputs,
     std::optional<Error> ex)
 {
+    outputLocks.unlock();
     buildResult.status = status;
     if (ex)
         buildResult.errorMsg = fmt("%s", normaltxt(ex->info().msg));


src/libstore/build/derivation-goal.cc
```diff
@@ -627,6 +627,7 @@ void DerivationGoal::tryToBuild()
     if (buildMode != bmCheck && allValid) {
         debug("skipping build of derivation '%s', someone beat us to it", worker.store.printStorePath(drvPath));
         outputLocks.setDeletion(true);
+        outputLocks.unlock();
```
gbpdt, line 630: It's possible this might be better placed in `done`, or perhaps `buildDone` could be used (I'm not too familiar with the code).

edolstra:
> This means the locks are held until the process terminates.

I don't think this is the case in general. `outputLocks` will be released via RAII from the goal's destructor. However, this may take a long time for top-level goals.

matthewbauer:
Goals are only destructed at the end of `Store::buildPaths`, so it will wait for all sibling goals to finish too IIUC. It seems like we should unlock these paths any time `AlreadyValid` comes up.

gbpdt:
> Goals are only destructed at the end of `Store::buildPaths`, so it will wait for all sibling goals to finish too IIUC.

Yes that matches the behaviour we are seeing.  We have a top-level goal with a fairly 'wide' graph of dependencies underneath it.  We see already-built goals in the middle of the graph hang around for some time while their siblings are being built, sometimes long enough to cause deadlock with other nix build processes building a similar dependency graph.

> It seems like we should unlock these paths any time `AlreadyValid` comes up.

I took a look through the other cases and I couldn't immediately see any others that actually take the locks, although I may well have missed something.  It looks like the other parts that might take locks (e.g. repairing) create new derivation goals which should go through the same code paths.

gbpdt:
We've been running this patch in our production Nix (2.3) builds for over a week now and have observed no deadlocks or other problems, so I'd support merging it.

bjornfor:
I've also seen deadlocks, with nix-2.7.0. I wasn't able to reproduce the issue outside of the production env, but hoping it's the same issue and that this fixes it.

gbpdt:
@edolstra any way we could get this merged for a future Nix release?

fricklerhandwerk:
Triaged in Nix team meeting:

@gbpdt please move the `unlock` to `done`, and we can merge it.

gbpdt:
> Triaged in Nix team meeting:
> 
> @gbpdt please move the `unlock` to `done`, and we can merge it.

Apologies @fricklerhandwerk, I missed this, thanks for taking a look! I've made the change (just moving to `done`, I didn't try to consolidate any other call sites of `outputLocks.unlock()`). PTAL

PR #6472, opened 2022-05-02
Draft: `#! nix-shell -i bash` use bash from shell.nix

Radvendii:
When nix-shell is non-interactive, we don't need to make sure we launch
an interactive bash, and it's more important to get the pinned bash

I'm a little worried this will break https://github.com/NixOS/nixpkgs/issues/27493

@domenkozar (from that nixpkgs issue)

@edolstra @regnat (discussed the problem & solution earlier)

TODO:
- [ ] communicate better what is going on to the user. if they `unset NIX_PATH`, they still get a scary error message despite it not really being an issue in shebang mode.

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -506,7 +506,7 @@ static void main_nix_build(int argc, char * * argv)
                 + structuredAttrsRC +
                 "\n[ -e $stdenv/setup ] && source $stdenv/setup; "
                 "%3%"
-                "PATH=%4%:\"$PATH\"; "
+                "%4%"
                 "SHELL=%5%; "
                 "BASH=%5%; "
                 "set +e; "
@@ -520,7 +520,7 @@ static void main_nix_build(int argc, char * * argv)
                 shellEscape(tmpDir),
                 (pure ? "" : "p=$PATH; "),
                 (pure ? "" : "PATH=$PATH:$p; unset p; "),
-                shellEscape(dirOf(*shell)),
+                (interactive && shell->substr(0, 1) == "/" ? "PATH=" + shellEscape(dirOf(*shell)) + ":\"$PATH\"; " : ""),
                 shellEscape(*shell),
                 (getenv("TZ") ? (std::string("export TZ=") + shellEscape(getenv("TZ")) + "; ") : ""),
                 envCommand);


Radvendii:
Oh. As I feared there's an error on Mac. Does anyone who uses mac have an insight into what's  going on?

Radvendii:
never mind? error seems to have magically disappeared.

thufschmitt:
> never mind? error seems to have magically disappeared.

ve restarted the CI ;)

Radvendii:
Right now `$BASH` and `$SHELL` are still set to the same value they were before. This was a while ago, so I don't remember our reasoning exactly. Was it deliberate that we left the `SHELL=` and `BASH=` lines in? 

It would be nice if we could go further and remove the error message (and the whole code-path that leads up to it) in the non-interactive case:

```
error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at «string»:1:9:

            1| (import <nixpkgs> {}).bashInteractive
             |         ^
will use bash from your environment
```

Since it's not really an error, and we're *not* using bash from the environment. I am wary though because this code is reused by a bunch of different commands.

I will add tests, and then maybe we should merge this PR as is, and change other things, like the error message, separately?

domenkozar:
+1 for removing that error message, I set `NIX_PATH` to be empty on all my machines/environments and it's annoying to see that error pop up.

fricklerhandwerk:

We'll wait for tests to be added. @Radvendii do you still have capacity to pursue this?

Radvendii:
So I stumbled across this again 2 years later and finally sat down to tackle it. I realised I'd gotten a couple things wrong.

1. What I thought had to do with interactive/non-interactive really had to do with whether we're in a shebang. If you run `nix shell --run "<command>"`, it might be relevant which `bash` is running that `<command>`. If you `#!nix-shell -i ruby`, it doesn't really matter which bash is executing `exec ruby`.
2. It still doesn't sit right to me that `nix-shell --run "<command>"` won't use the `bash` provided by your `shell.nix`. I don't think there's a great way to solve this though. We just don't have access to it at the right time.
3. The whole `PATH=" + shellEscape(dirOf(*shell)) + ":$PATH"` thing going on has nothing to do with anything else, and I'm starting to think it should just go away entirely.
   
   What it does is set it up so that if you run `bash` from inside a nix-shell, it will get you the same bash that you were already in. I'm guessing this is in place so you get `bashInteractive` rather than some non-interactive bash provided by `stdenv.mkDerivation`, but I would argue that's an issue with `mkShell`, which should provide an interactive bash by default.

Radvendii:
@thufschmitt @domenkozar @edolstra Sorry if you are no longer the people to ping about this, I know it's been a while I'm not sure who the right people are currently.

This issue has come up again recently. The error message can be quite confusing to people. In the case of a nix-shell shebang, I've removed the error message entirely. We will immediately execute the interpreter so we don't need it. In non-shebang cases, I've demoted the detailed error text to info level. People will still be notified that we're taking bash from the environment, and they can find the reason why with verbose flags, but in almost all cases this doesn't matter, so by default it will not be as loud.

As far as I'm aware the only thing holding this back was tests. I added those back in June 2024. I think I just didn't do a good job bumping it at the time.

CI is failing, but I tested `nix build .#checks.x86_64-linux.packaging-overriding` locally on `master` branch and it also failed, so it seems unrelated.

PR #6479, opened 2022-05-04
libfetchers: Add Fossil fetching support

MatthewCroughan:
Most of this work has been done by @Uthar via https://github.com/NixOS/nix/pull/6376 and https://github.com/NixOS/nix/issues/6368

I am just rebasing his work, and also added a check to make sure the input hashes are sha1 or sha256.

You can test this out by doing the following in this repo:
1. `nix shell .#nix nixpkgs#fossil`
2. `nix repl`
3. `builtins.fetchTree "fsl+https://sqlite.org/src?rev=sha256:9547e2c38a1c6f751a77d4d796894dec4dc5d8f5d79b1cd39e1ffc50df7b3be4"`

### DIFF
--- a/src/libexpr/flake/flakeref.cc
+++ b/src/libexpr/flake/flakeref.cc
@@ -180,6 +180,30 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
                             fragment);
                     }
 
+
+                    if (pathExists(flakeRoot + "/.fslckout")) {
+                        auto base = std::string("fsl+file://") + flakeRoot;
+
+                        auto parsedURL = ParsedURL{
+                            .url = base, // FIXME
+                            .base = base,
+                            .scheme = "fsl+file",
+                            .authority = "",
+                            .path = flakeRoot,
+                            .query = decodeQuery(match[2]),
+                        };
+
+                        if (subdir != "") {
+                            if (parsedURL.query.count("dir"))
+                                throw Error("flake URL '%s' has an inconsistent 'dir' parameter", url);
+                            parsedURL.query.insert_or_assign("dir", subdir);
+                        }
+
+                        return std::make_pair(
+                            FlakeRef(Input::fromURL(parsedURL), get(parsedURL.query, "dir").value_or("")),
+                            fragment);
+                    }
+
                     subdir = std::string(baseNameOf(flakeRoot)) + (subdir.empty() ? "" : "/" + subdir);
                     flakeRoot = dirOf(flakeRoot);
                 }
--- a/src/libfetchers/fossil.cc
+++ b/src/libfetchers/fossil.cc
@@ -220,9 +220,19 @@ struct FossilInputScheme : InputScheme
 
         }
 
+        auto checkHashType = [&](const std::optional<Hash> & hash)
+        {
+            if (hash.has_value() && !(hash->type == htSHA1 || hash->type == htSHA256))
+                throw Error("Hash '%s' is not supported by fossil. Supported types are sha1 and sha256.", hash->to_string(Base16, true));
+        };
+
+        checkHashType(input.getRev());
+
         if (!input.getRef()) input.attrs.insert_or_assign("ref", "trunk");
 
         auto revOrRef = input.getRev() ? input.getRev()->to_string(Base16, false) : *input.getRef();
+
+        auto revHashType = input.getRev()->type;
 
         Attrs unlockedAttrs({
             {"type", "fsl"},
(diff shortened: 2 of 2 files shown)


alexshpilkin:
I’m not sure this change should be applied, but at least it explains what changes between @Uthar’s approach in https://github.com/Uthar/nix/commit/d0091b482826fe52bef4bf7c442a20c791691bde (detecting hashes by length) and the one that was eventually merged in https://github.com/NixOS/nix/commit/2c2fd4946f96e6839ecbfb4cf61318d8910e7e8f (detecting them by prefix) that makes SHA-256 hashes not work. They work if you change `emitTreeAttrs()` to not strip the prefix off the hash:

```diff
diff --git a/src/libexpr/primops/fetchTree.cc b/src/libexpr/primops/fetchTree.cc
index d7c3c9918..0ab3932d9 100644
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -40,7 +40,7 @@ void emitTreeAttrs(
     if (!forceDirty) {
 
         if (auto rev = input.getRev()) {
-            attrs.alloc("rev").mkString(rev->gitRev());
+            attrs.alloc("rev").mkString(rev->to_string(Base16, rev->type != htSHA1));
             attrs.alloc("shortRev").mkString(rev->gitShortRev());
         } else if (emptyRevFallback) {
             // Backwards compat for `builtins.fetchGit`: dirty repos return an empty sha1 as rev
```

(I still find the approach of smuggling SHA3-256 hashes by saying they are SHA-256 hashes questionable.)

MatthewCroughan:
I've created an issue here in case it is helpful. https://github.com/NixOS/nix/issues/6496

alexshpilkin:
@MatthewCroughan I can’t speak with any degree of authority here, but yes, I think this is essentially a bug in the initial support for non-SHA-1 revision hashes (as done by @Uthar in #6376). Unfortunately, like the rest of that support it’s invisible without an actual user, such as this Fossil fetcher.

Uthar:
Yes, seems like a bug, I also can't `fetchTree` with nix 2.8.0 and an external fossil fetcher plugin (https://fossil.galkowski.xyz/nix-fsl-fetcher). (Have not tested it at the time, only tested fetching from an input in flake.nix/ ad hoc `nix flake` commands)

Now I noticed as well, somewhere between that merge and 2.8.0 `nix flake {metadata, update}` stopped working for a fsl input with no specified rev, using that plugin.

Uthar:
I Think that this won't work, because Nix understands only 52-bit sha256 hashes, these are 64-bit sha3-256 ones.

Uthar:
So it just fails here and silently tries to parse it as sha1:

https://github.com/NixOS/nix/blob/899878f77a1b29b337aca409ec3f7a420e574a6a/src/libfetchers/fetchers.cc#L244-L249

Maybe would be nice to have a warning here?

Uthar:
If we want fossil support we should teach Nix about sha3-256 hashes

Maybe here is a good starting point: https://github.com/NixOS/nix/blob/899878f77a1b29b337aca409ec3f7a420e574a6a/src/libutil/hash.cc#L31

Uthar:
nevermind, it seems to work with both 52 and 64 long sha256 hashes

Uthar:
Hmm i'm getting weird errors:

```
nix-repl> builtins.getFlake "fsl+https://fossil.galkowski.xyz/efsl"
warning: Defaulting to SHA1 for hash 84f4adc211c718773997f07b79724f3b34ee926005d9987304c4793ab1976e51 because of error: hash '84f4adc211c718773997f07b79724f3b34ee926005d9987304c4793ab1976e51' does not include a type
error: hash '84f4adc211c718773997f07b79724f3b34ee926005d9987304c4793ab1976e51' has wrong length for hash type 'sha1'
```

Uthar:
Seems to be stripping the prefix somewhere

Uthar:
got it working with this:

https://github.com/Uthar/nix-fsl-fetcher/commit/c1c67f31cfc1884974ff6e1523dd9d6b72f06b98

```
$ nix build github:uthar/nix-fsl-fetcher --override-input nix nix
$ nix --extra-plugin-files result/bin/libnix-fsl-fetcher.so repl
nix-repl> builtins.getFlake "fsl+https://fossil.galkowski.xyz/nix-fsl-fetcher"
```

Uthar:
see this pr https://github.com/NixOS/nix/pull/7272

I added the working version there

PR #6480, opened 2022-05-04
src/libutil/util.cc: add missing #include <nlohmann/json.hpp>

trofi:
Without the change nix fails to build on Gentoo against nlohmann_json-3.10.5:

      CXX    src/libutil/util.o
    src/libutil/util.cc: In function ‘const json* nix::get(const json&, const string&)’:
    src/libutil/util.cc:1591:14: error: invalid use of incomplete type ‘const json’ {aka ‘const class nlohmann::basic_json<>’}
     1591 |     auto i = map.find(key);
          |              ^~~
    In file included from src/libutil/config.hh:7,
                     from src/libutil/logging.hh:5,
                     from src/libutil/util.hh:5,
                     from src/libutil/util.cc:1:
    /usr/include/nlohmann/json_fwd.hpp:40:7: note: declaration of ‘using json = class nlohmann::basic_json<>’ {aka ‘class nlohmann::basic_json<>’}
       40 | class basic_json;
          |       ^~~~~~~~~~
    src/libutil/util.cc:1592:14: error: invalid use of incomplete type ‘const json’ {aka ‘const class nlohmann::basic_json<>’}
     1592 |     if (i == map.end()) return nullptr;
          |              ^~~

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -18,6 +18,7 @@
 
 #include <fcntl.h>
 #include <grp.h>
+#include <nlohmann/json.hpp>
 #include <pwd.h>
 #include <sys/ioctl.h>
 #include <sys/types.h>


edolstra:
What version of Nix is this? I don't see any references to `nlohmann::json` in `util.cc`.

edolstra:
Oh, I see this was changed by the `at()` PR.

edolstra:
Note that this also means that `util.hh` needs to include `json_fwd.hpp`, which I wouldn't be entirely fond of. Maybe it's better to move this into a `json-utils.hh`.

edolstra:
Thanks for alerting me to the build failure. I've fixed it in a slightly different way (#6482).

PR #6483, opened 2022-05-04
Fix the parsing of the sourcehut refs file

thufschmitt:
Since a26be9f3b89be2ee90c6358250b9889b37f95cf8, the same parser is used
to parse the result of sourcehut’s `HEAD` endpoint (coming from [git
dumb protocol]) and the output of `git ls-remote`. However, they are very
slightly different (the former doesn’t specify the current reference
since it’s implied to be `HEAD`).

Unify both, and make the parser a bit more robust and understandable (by
making it more typed and adding tests for it)

[git dumb protocol]: https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols#_the_dumb_protocol

Unbreaks the `sourcehutFlakes` hydra job


src/libutil/git.hh
```diff
@@ -0,0 +1,40 @@
...
+//
+//    where {target} is a commit id and {reference} is mandatory
+struct LsRemoteRefLine {
+    enum struct Kind {
+        Symbolic,
+        Object
+    };
+    Kind kind;
+    std::string target;
+    std::optional<std::string> reference;
```
zseri, line 33: I think it might be more appropriate to use just `std::string` here.

zseri, line 33: I think it might be more appropriate to use just `std::string` here.

thufschmitt: Why so? It’s actually optional in the file, so better make it explicit

thufschmitt: (deleting since it’s a duplicate comment. I guess github just got slightly crazy for a minute)

src/libutil/git.cc
```diff
@@ -0,0 +1,25 @@
...
+    std::match_results<std::string_view::const_iterator> match;
+    if (!std::regex_match(line.cbegin(), line.cend(), match, line_regex))
+        return std::nullopt;
+
+    return LsRemoteRefLine {
+        .kind = match[1].length() == 0
+            ? LsRemoteRefLine::Kind::Object
+            : LsRemoteRefLine::Kind::Symbolic,
+        .target = match[2],
+        .reference = match[3].length() == 0 ? std::nullopt : std::optional<std::string>{ match[3] }
```
zseri, line 20: I wouldn't make it optional based on this line (Is a present but empty reference valid at all? I don't think so... (but I could be wrong, if I am, the code doesn't properly represent this)). `optional(string())` shouldn't be represent-able. Especially since the usage also doesn't take advantage of the `optional`-ity at all.

PR #6484, opened 2022-05-04
debug #6383

fogti:
This PR inserts the missing debug output to better diagnose what went wrong in #6383; and fixes a potential use-after-free via `getOr`.

The main problem is that the error might take a long time to appear, but once it appears is reproducible if the build output is really required (and not just because the output got pulled in by a full build, but is not used by any other derivation)

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -988,13 +988,28 @@ void DerivationGoal::resolvedFinished()
             auto resolvedHash = get(resolvedHashes, wantedOutput);
             if ((!initialOutput) || (!resolvedHash))
                 throw Error(
-                    "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,resolve)",
+                    "derivation '%s' doesn't have expected output '%s'",
                     worker.store.printStorePath(drvPath), wantedOutput);
             auto realisation = get(resolvedResult.builtOutputs, DrvOutput { *resolvedHash, wantedOutput });
-            if (!realisation)
+            if (!realisation) {
+                std::string builtStr = "\tbuiltOutputs:\n";
+                for (const auto &i : resolvedResult.builtOutputs) {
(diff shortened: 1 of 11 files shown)


src/libstore/build/derivation-goal.cc
```diff
@@ -988,13 +988,28 @@ void DerivationGoal::resolvedFinished()
             auto resolvedHash = get(resolvedHashes, wantedOutput);
             if ((!initialOutput) || (!resolvedHash))
                 throw Error(
-                    "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,resolve)",
+                    "derivation '%s' doesn't have expected output '%s'",
                     worker.store.printStorePath(drvPath), wantedOutput);
             auto realisation = get(resolvedResult.builtOutputs, DrvOutput { *resolvedHash, wantedOutput });
```
zseri, line 993: example failure (reported by @trofi):
```
error: derivation '/nix/store/g7q24qrdzdcvsirxw8aw0alc9s0qsl77-shared-mime-info-unstable-2021-12-03.drv' doesn't have expected output 'dev @ sha256:094p4p14vbmqhhd2g4nr08phln6hr789rzqnjwll9q79kisc7n7g'
        builtOutputs:
                out @ sha256:094p4p14vbmqhhd2g4nr08phln6hr789rzqnjwll9q79kisc7n7g (key) = /nix/store/98f30ch5mldzkw0zznx2s2ll6r940jv3-shared-mime-info-unstable-2021-12-03
```
=> hash matches, but the wanted output is not part of the built outputs.

Ericson2314:
I am a little confused what is going with all the explicit construction; are you thinking what we did before is more unsafe than we think?

fogti:
Most of the explicit construction is to make sure that strings are still valid when used, and that the types used are correct (the only important cases are `Path`, which forces a copy, and `string_view`, which forces a proper string reference). But as this patch by now diverged anyways, I'm just going to close it. I haven't encountered the bug this is supposed to debug in the last few months.

Issue #3391, opened 2020-03-08
eval-okay-fromjson.nix contains an example from an RFC file


thkoch2001:
This issue was part of the rejection reasons in the first review of nix for entrance in the Debian archive, see https://salsa.debian.org/debian/nix/issues/5.

Please note that Debian copyright reviews are done by volunteers that are not lawyers. Sometimes the reviews might seem overly nitpicking. This however helps Debian to not have any troubles even with potential lawsuits.

tests/lang/eval-okay-fromjson.nix contains text from an RFC whose license does not permit derivative works (it is not explicitly marked as a code component in the RFC).

It would be trivial to replace the file with another test case that has no apparent similarity with the example from the RFC.

Thank you for your patience and help!

tomberek:
Closing as stale. Re-open if needed.

PR #6486, opened 2022-05-04
Change json example to be original

thkoch2001:
Closes: #3391


tests/lang/eval-okay-fromjson.nix
```diff
@@ -1,36 +1,35 @@
...
+          "Title":  "The Penguin Chronicles",
+          "Width":  1920,
+          "Height": 1080,
+          "EmbeddedData": [3.1415927, 23493,null, true  ,false, -10],
+          "Thumb": {
               "Url":    "http://www.example.com/image/481989943",
-              "Height": 125,
-              "Width":  100
+              "Width":  200,
+              "Height": 255,
```
edolstra, line 12: I think this comma is giving a parse error.

thkoch2001:
creating a new PR, thx.

edolstra:
Note that it's better to just update the PR rather than create a new one, since it creates less clutter.

PR #6505, opened 2022-05-08
libexpr: Fix manual link in error message

jtojnar:
It was changed to the old manual in https://github.com/NixOS/nix/commit/8895fa70a4b05ddebbb5a23ea96464d5e01345fb


thufschmitt (approved):
Thanks :)

Issue #6521, opened 2022-05-11
The progress bar should disappear on exit


ncfavier:
**Describe the bug**

When using `nix-build` (not `nix build`) with `--log-format bar` or `--log-format bar-with-logs`, if nix is interrupted by e.g. SIGINT or SIGTERM, the final line of the progress bar stays visible, and the prompt is printed on the same line.

**Steps To Reproduce**

1. Run a long build with `nix-build --log-format bar`
2. Interrupt the build with `^C` or `kill`
3. Observe something like this:
```
...
  /nix/store/l2iiwsrrscp3k15bl5bwmk07jjnb4zxs-gcc-10.3.0
copying path '/nix/store/l2iiwsrrscp3k15bl5bwmk07jjnb4zxs-gcc-10.3.0' from 'https://cache.nixos.org'...
error: interrupted by the user
[0/3 built, 1/0/5 copied (18.9/176.1 MiB), 5.4/40.7 MiB DL] fetching gcc-10.3.0 from https://cache.nixos.org~/git/nix $
```

**Expected behavior**

The same as when using `nix build`, i.e.:
```
error: interrupted by the user
~/git/nix $
```

**`nix-env --version` output**

Latest master.

**Additional context**

~I tried looking into this but I couldn't find why the behaviour differs between the two commands.~ EDIT: https://github.com/NixOS/nix/pull/6523

PR #6523, opened 2022-05-11
Stop the logger properly in legacy commands

ncfavier:
Fixes https://github.com/NixOS/nix/issues/6521

Ensures the logger is stopped on exit in legacy commands. Without this, when using `nix-build --log-format bar` and stopping nix with CTRL+C, the bar is not cleared from the screen.

Tested.


thufschmitt (approved):
So that’s why I occasionally had some weird glitches with the old CLI!

Thanks :)

PR #6535, opened 2022-05-16
Allow "self" in flake input paths

### LINKED ISSUE #4931: flake inputs cannot follow "self"
```bat
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: flake.nix
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ {
   2 ~ │   description = "Test Flake For Follows";
   3 ~ │   inputs =
   4 ~ │     {
   5 ~ │       nixos-generators.url = "github:nix-community/nixos-generators";
   6 ~ │       nixos-generators.inputs.nixpkgs.follows = "self";
   7   │     };
   8   │
   9 ~ │   outputs = { ... }: {};
  10   │ }
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```

```ShellSession
nix flake lock --update-input nixos-generators
warning: Git tree '/tmp/flake-test' is dirty
error: input 'nixos-generators/nixpkgs' follows a non-existent input 'self'
```

Imagine I wanted to ship a remote flake that depends on my local flake as part of my local flake ("engulfing").

With this bug present, I cant do that in flake interface land.

It is a legal operation.

alexshpilkin:
Fixes #4931 by making “self” in flake input paths the moral equivalent of “.” in filesystem paths. This takes the lazy approach and interprets any “self” components in the path at parse time.


thufschmitt:
I don’t have a very strong opinion on that topic, deferring to @edolstra since he might have one

roberth:
No particularly strong opinion here either, but if we create a keyword (`self`), we should disallow it as an input name to avoid potential for ambiguity.

alexshpilkin:
@roberth The point was (I thought) that `self` is already effectively a keyword in that even if you declare an input by that name you’ll still get your own outputs instead of it. I agree that you shouldn’t be able to declare one at all, though.

tomberek (approved):

alexshpilkin:
@roberth OK, that's fair. Except `tests/flakes/follow-paths.sh` looks more logical to me? And `tests/flakes/circular.sh` as well just to mirror the docs.

As for disallowing a "self" input, I'm a bit torn because in principle I see no reason why, for example, writing `inputs.self.inputs.foo.follows = "bar"` shouldn't work as an synonym of `inputs.foo.follows = "bar"`—it's silly but not nonsense—but trying to make it work seems extremely annoying, e.g. we then need to define how `inputs.self.inputs.foo`, `inputs.foo.inputs.self`, and `inputs.foo` are merged recursively if more than one of them is present. No other attributes of a "self" input seem to make sense, so I think I'll sacrifice this bit of orthogonality and just ban any mention of a `self` input in the `inputs` part of a flake.

Ericson2314:

Main takeaway:

- **Idea Approved, but requires tests and docs**

Discussion:

- @thufschmitt would prefer something like `.`, but doesn't care too much about the concrete syntax
- `self` isn't that bad

roberth:
> * @thufschmitt would prefer something like `.`, but doesn't care too much about the concrete syntax

Note that since then we've also accepted `flake#.foo` as a notation for absolute attribute paths (e.g. to avoid `packages.<system>.foo`), which is also inspired on unix path notation but with `.` instead of `/`.

Things still to do:

- [ ] Add test case
- [ ] Only operate on the root of the inputs tree. The `inputs` attribute, such as e.g. returned by `getFlake`, does not contain any self attributes either.
- [ ] Consider `.` as syntax for self


I'll mark this as draft for the time being. Please undraft when it's ready for review.

PR #6537, opened 2022-05-16
flake: use github: reference to nixpkgs

cole-h:


edolstra:
Does the static build still work? That was the reason why previous nixpkgs updates had to be reverted.

cole-h:
I see. Indeed, the static build is still broken.

I've pared this PR down to just change the references to the Nixpkgs flake to use the `github:` format, so I don't have to override the input. (I have my own flake registry, so the registry reference doesn't work.)

thufschmitt:
Really glad to see that the nixpkgs update fixed both the static build and the OSX reference checking :smiley: 

It’d be nice to run the full hydra jobset on the branch just in case, but that looks good otherwise. Thanks for that :)

cole-h:
Sorry for the confusion: this only changes the nixpkgs input from being indirect to being a GitHub input. I reverted the bump because static is still broken in 21.11, but looks like it will be fixed by 22.05.

Issue #6541, opened 2022-05-17
`builtins.getFlake` fails with "cannot write modified lock file"


ncfavier:
**Describe the bug**

`builtins.getFlake` fails with "cannot write modified lock file" when using it on a repository without a lock file, e.g. home-manager, even when passing `--no-write-lock-file`.

**Steps To Reproduce**

```
$ echo 'builtins.getFlake "github:nix-community/home-manager"' > f.nix
$ nix eval --no-write-lock-file -f f.nix
error: cannot write modified lock file of flake 'github:nix-community/home-manager' (use '--no-write-lock-file' to ignore)
(use '--show-trace' to show detailed location information)
```

**Expected behavior**

No such error, i.e. don't attempt to write a lock file in `builtins.getFlake`.

**`nix-env --version` output**

`nix (Nix) 2.9.0pre20220505_f4102de`

**Additional context**

https://github.com/NixOS/nix/blob/fdc97fc3f93c2afbd260b538b3546e14c0dcd134/src/libexpr/flake/flake.cc#L722-L729

We should probably add `.writeLockFile = false` here? I'll PR that in a while if there's no response here.

ncfavier:
Yes, the suggested fix doesn't change that

PR #6544, opened 2022-05-18
Do not attempt to write a lock file in builtins.getFlake

ncfavier:
Fixes https://github.com/NixOS/nix/issues/6541


thufschmitt:
I don’t think we can just do that, because that would break purity (since `getFlake` can be used in pure eval mode, and locking a flake is inherently impure). But it’s probably fine to set `writeLockFile` to `!evalSettings.pureEval`, so that it’s only available in impure mode.

For the pure use-case, maybe the interface of `getFlake` could be extended (to allow passing an external lockfile for example). Alternatively/in the meantime, `fetchTree`+[callLocklessFlake](https://github.com/NixOS/nixpkgs/pull/167947) might do the job too

ncfavier:
This should be covered by the lines above:

https://github.com/NixOS/nix/blob/681fc3c7996997aaabf645a3203ba7901053168a/src/libexpr/flake/flake.cc#L719-L720

edolstra:
@ncfavier That only checks whether the flake reference is locked (e.g. `github:nix-community/home-manager/32a7da69dc53c9eb5ad0675eb7fdc58f7fe35272` vs `github:nix-community/home-manager`). A locked flake reference can refer to a flake that doesn't have a `flake.lock`, which would be non-reproducible (e.g. its `flake.nix` can reference an unlocked `nixpkgs`).

thufschmitt:
Ah I should have looked better. The line you point to isn’t enough (it only checks whether the ref passed to `getFlake` is a locked reference, not that it contains a lockfile), but as @edolstra mentionned in https://github.com/NixOS/nix/issues/6541#issuecomment-1129762461, passing `.allowMutable = !evalSettings.pureEval` prevents it from updating the lockfile in pure mode. So it’s all right :)

thufschmitt:
> So it’s all right :)

Well, it’s all right, bit given how non-trivial it is, I think it’s definitely worth a test to ensure that it doesn’t get accidentally broken ;)

The whole thing actually deserves some tests imho

PR #6548, opened 2022-05-19
fetchTree: Allow fetching plain files

### LINKED ISSUE #3785: flakes: support downloading files (as opposed to github, git and tarball types)

thufschmitt:
Replace the `tarball` fetcher by a more generic `url` one, mostly similar except that it takes an extra `unpack` parameter (false by default) to specify whether the target is an archive to extract or a plain file.

Also adds a few convolutions for backwards compatibility:
- The `tarball` fetcher type is kept. It’s now an alias for `url`, but with `unpack` defaulting to `true`
- Flake references that used to resolve to a `tarball` fetcher (i.e. `http(s)://` and `file://` urls whose path ends with a known archive extension) still do, meaning that
  1. They are still extracted by default (which is both a backwards-compatibility measure, and very handy)
  2. The generated lockfiles are still compatible with older Nix versions who don’t know of the `url` fetcher

Fix #3785

Originally taken from #4153 (although most of it has been rewritten)

/cc @tomberek


thufschmitt:
One thing I’m really unhappy with is the `url` name. I’ve taken it for consistency with `fetchurl`, but it’s a highly vague and overloaded term, so I’d happily accept any suggestion for a better one :)

edolstra:
I would remove the "unpack" parameter and have two different input types: "tarball" and "file". Calling it "url" is not very meaningful since most fetchers use a URL, they just differ in what they do with the URL (e.g. in this case, unpack a tarball, or use a plain file).

thufschmitt:
> I would remove the "unpack" parameter and have two different input types: "tarball" and "file".

That’s actually what I had initially. I moved away from it because it was making the URL parsing slightly weird: I had the `tarball` URL parser accepting urls ending with a known archive format, the `file` parser accepting every curl-esque url, meaning that they were overlapping and and it only worked properly because they were registered in the right order.

… but now that I write it down, it feels much less dirty than the actual weird semantics of the current PR, so you’re right, I’ll switch back to that :slightly_smiling_face:

thufschmitt:
OK, I’ve made them separate input types, and added the `file+{blah}` and `tarball+{blah}` url schemes to resolve the ambiguity if needs be

tomberek:
Does a “file” type need to have the “flake” parameter? Or can it default to false because if you do not unpack it can’t be a flake at all. Perhaps the file it itself can be raw flake.nix file, but that seems too rare/edge case; I’d prefer the simpler situation where “file+” by default knows it cannot be a flake.

thufschmitt:
> Does a “file” type need to have the “flake” parameter? Or can it default to false because if you do not unpack it can’t be a flake at all. Perhaps the file it itself can be raw flake.nix file, but that seems too rare/edge case; I’d prefer the simpler situation where “file+” by default knows it cannot be a flake.

Ideally it probably shouldn’t indeed. In practice I couldn’t find an easy and clean way to do it ;)

src/nix/flake.md
```diff
@@ -181,9 +181,17 @@ Currently the `type` attribute can be one of the following:
 * `tarball`: Tarballs. The location of the tarball is specified by the
   attribute `url`.
 
-  In URL form, the schema must be `http://`, `https://` or `file://`
-  URLs and the extension must be `.zip`, `.tar`, `.tgz`, `.tar.gz`,
-  `.tar.xz`, `.tar.bz2` or `.tar.zst`.
+  In URL form, the schema must be `tarball+http://`, `tarball+https://` or `tarball+file://`.
+  If the extension corresponds to a known archive format (`.zip`, `.tar`,
+  `.tgz`, `.tar.gz`, `.tar.xz`, `.tar.bz2` or `.tar.zst`), then the `tarball+`
+  can be dropped.
```
zimbatm, line 187: Isn't that change breaking backwards compatibility?

src/nix/flake.md
```diff
@@ -181,9 +181,17 @@ Currently the `type` attribute can be one of the following:
...
+
+* `file`: Plain files or directory tarballs, either over http(s) or from the local
+  disk.
```
zimbatm, line 190: Might as well rename the function to `builtins.fetch` if it downloads arbitrary things, even non-tree ones.

src/nix/flake.md
thufschmitt: Nope, on the contrary: Currently http(s)/file urls are required to have such an extension and are treated as tarballs. The new behavior makes it so that what was already accepted still is, and with the same semantics, and the rest is treated as file urls

thufschmitt: The problematic bit though, is that adding a new tarball extension would break backwards-compat a bit: with this change, `https://foo.com/bar.tar.lzma` is fetched as a plain file, and if we wanted to add `.tar.lzma` as a recognized archive extension, it would become an directory tarball

src/nix/flake.md
thufschmitt: Might be an idea, indeed :)

I’d leave that for later though if you’re OK with that

src/libfetchers/tarball.cc
```diff
@@ -225,6 +234,42 @@ struct TarballInputScheme : InputScheme
...
+        auto parsedUrlScheme = parseUrlScheme(url.scheme);
+        return transportUrlSchemes.count(std::string(parsedUrlScheme.transport))
+            && (parsedUrlScheme.application
+                    ? parsedUrlScheme.application.value() == inputType()
+                    : !hasTarballExtension(url.path));
+    }
+
+    std::pair<StorePath, Input> fetch(ref<Store> store, const Input & input) override
+    {
+        auto file = downloadFile(store, getStrAttr(input.attrs, "url"), input.getName(), false);
```
edolstra, line 254: A problem with the use of `downloadFile()` is that it uses `FileIngestionMethod::Flat` instead of `FileIngestionMethod::Recursive`. Currently it's assumed that all flake inputs use recursive+sha256 with a name of "source". This allows inputs to be substituted using the `narHash` attribute in the lock file (see `Input::computeStorePath()`). However, the lazy trees branch will probably remove the ability to substitute inputs anyway...

thufschmitt: Oh indeed, I didn’t notice the `Flat` here. Would it be an issue to make `downloadFile` use `Recursive`? Afaik none of the other use sites really care about how the result is hashed since it’s only used internally by the fetchers but doesn’t leak outside.

(Alternatively I could just wrap that by adding an extra step that would copy the output to a `Recursive` CA location, but if that can be avoided it’s even better)

matthewbauer:
PR to add support to flake-compat:

https://github.com/edolstra/flake-compat/pull/44

Issue #6549, opened 2022-05-19
Propagating Flake inputs


ysndr:
**Problem Description**
Many flakes depend on local registry entries, often `nixpkgs`, either out of convenience or for not knowing the implications of it:

```nix 
{
  outputs = {nixpkgs}: { ... }
}
```

However, if flakes start to import other flakes (perhaps with their own dependencies, with their own ..), due to locking a single flake can pull in multiple versions of `nixpkgs`.

This is commonly mitigated by setting the dependencies inputs nixpkgs argument explicitly using the `follows` attribute. This method is explicit but does not scale very well as the number of inputs and the depth of the dependency tree grows. Currently it also seems have limits (#5790).

```nix
{
  inputs.nixpkgs.url = "..."; 
  
  inputs.dep.url = "...";
  inputs.dep.inputs.nixpkgs.follows = "nixpkgs"
  inputs.dep.inputs.subdep.inputs.nixpkgs.follows = "nixpkgs"
  # ^- #5790 suggests that follows deeper than two levels are currently not possible
  
  inputs.dep2.url = "...";
  inputs.dep2.inputs.nixpkgs.follows = "nixpkgs"
 
  ...

}
```

**Describe the solution you'd like**

Add a new flake input attribute `register` (TBD) which propagates the input to all other inputs of the flake recursively.

```nix
// parent/flake.nix

{
  inputs.nixpkgs.url = "special-nixpkgs";
  inputs.nixpkgs.register = true;

  inputs.dep.url = "...";
  inputs.dep2.url = "...";

  ...
}
```

```nix
// dep{,2}/flake.nix

{
  # no inputs (nixpkgs is not specified explicitly)
  
  outputs = { nixpkgs }: ...
#             ^
#             |
#             "special-nixpkgs"

}
```

Being an implicit method any kind of explicitly setting an input should be prioritized with the proposed priority being:

1. override (`.url`) 
2. `follows` 
3. some parents input with `register`
4. registry

The chosen flake to follow should be updated ass the dependency tree is walked downwards.

```nix
# root/flake.nix

{
  inputs.nixpkgs.url = "special-nixpkgs";
  inputs.nixpkgs.register = true;

  inputs.dep.url = "...";
}
```

```nix
# dep/flake.nix

{
  inputs.subdep.url = "...";
  outputs = { nixpkgs }: ...
#             ^
#             |
#             "special-nixpkgs" from root

}
```

```nix
# subdep/flake.nix

{
[...]

Issue #610, opened 2015-08-21
configure script uses bash'ism


blacklion:
`configure` script has these lines:

``` bash
# Expand all variables in config.status.
test "$prefix" = NONE && prefix=$ac_default_prefix
test "$exec_prefix" = NONE && exec_prefix='${prefix}'
for name in $ac_subst_vars; do
    declare $name="$(eval echo "${!name}")"
    declare $name="$(eval echo "${!name}")"
    declare $name="$(eval echo "${!name}")"
done
```

which fail on POSIX sh. POSIX sh doesn't understand `declare` and `${!name}` substitution.

tomberek:
Closing as stale. Re-open if needed.

PR #6550, opened 2022-05-19
Impl: Propagate Flake inputs

ysndr:
MVP implementation of #6549

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -370,7 +370,7 @@ LockedFlake lockFlake(
             const InputPath &lockRootPath,
             const Path &parentPath,
             bool trustLock,
-            std::map<Path, std::shared_ptr<LockedNode>> registered)>
+            std::map<Path, InputPath> registered)>
             computeLocks;
 
         computeLocks = [&](
@@ -381,9 +381,11 @@ LockedFlake lockFlake(
             const InputPath & lockRootPath,
             const Path & parentPath,
             bool trustLock,
-            std::map<Path, std::shared_ptr<LockedNode>> registered)
+            std::map<Path, InputPath> registered)
         {
             debug("computing lock file node '%s'", printInputPath(inputPathPrefix));
+
+            std::set<Path> registeredChilds;
 
             /* Get the overrides (i.e. attributes of the form
                'inputs.nixops.inputs.nixpkgs.url = ...'). */
@@ -393,6 +395,16 @@ LockedFlake lockFlake(
                     inputPath.push_back(id);
                     inputPath.push_back(idOverride);
                     overrides.insert_or_assign(inputPath, inputOverride);
+                }
+                if (input.registered) {
+                    // This is a stub node that is overriden in the input
+                    // auto childNode = std::make_shared<LockedNode>(
+                    //     *input.ref, *input.ref, input.isFlake);
+                    auto inputPath(inputPathPrefix);
+                    inputPath.push_back(id);
+                    debug("preregistering %s", id.c_str());
+                    registeredChilds.insert(id);
+                    registered.insert_or_assign(id, inputPath);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


MagicRB:
This kind of relates to #610 and #5497 and any other similar issue

thufschmitt:
I've mixed feelings about that. I see the problem that it's solving, but I'm not entirely convinced by the solution. It would probably be useful for some use-cases like a multi-flakes repository, but would potentially be fairly confusing when used with external flakes ‑ because of the fact that it can or override the lockfile depending on whether the flake has an explicit input specification.

Maybe restricting it to lockless flakes in a first time like you suggest in https://github.com/NixOS/nix/issues/6549 would be better until we can come up with a good solution.

charlesbaynham:
Can I vote against

> Maybe restricting it to lockless flakes in a first time like you suggest in https://github.com/NixOS/nix/issues/6549 would be better until we can come up with a good solution.

? I think a lot of the use of this will be overriding locks in input flakes which are locked in a way you don't like. An attribute to ignore lock files would be nice indeed. A big hammer would do lots of jobs, though not all 😁

ysndr:
I agree with @charlesbaynham. I think restricting this to only lockless flakes leaves too little use for it. I dont know if it would even work for local subflakes as their inputs are eventually locked as well by the importing flake.
Therefore, it would need to be constrained to completely lockless nix e.g. requiring `--no-write-lockfile` which has pretty limited application.

I agree people may be surprised if the nixpkgs that is locked with their flake is incompatible with the propagated / registered one. However I think that's acceptable as long as we make clear that you need explicitly add a nixpkgs input to keep it.
For third party flakes maybe its better to have an attribute to ignore the registered input instead:

*(<bikeshed>works better if we'd call it propagate </bikeshed>)*


```
inputs.nixpkgs.url = ...
inputs.nixpkgs.register = true;

inputs.someflake.url = ...; # <-- has registered nixpkgs

inputs.otherflake.url = ...;
inputs.otherflake.no-registered = true;

```

thufschmitt:
> I agree people may be surprised if the nixpkgs that is locked with their flake is incompatible with the propagated / registered one. However I think that's acceptable as long as we make clear that you need explicitly add a nixpkgs input to keep it.

That's not really my fear. My issue on the contrary is that this creates an incentive for publishing “incomplete” flakes (in that their inputs won't be defined) since that would be the only way to have their input transparently overriden by a `register` when imported from a parent flake. And that would be very limiting.

For example, if I'm the author of a public flake that is developped using `nixpkgs/nixos-22.05`, should I set `inputs.nixpkgs.url = "nixpkgs/nixos-22.05"` (in which case downstream users can't use the `register` mechanism to override it), or should I leave it blank (in which case I can't easily lock it to the right version since it's gonna pick the `nixpkgs` from the registry which might not be what I want)? Neither case is fully satsifying.

That's why I thought about limiting this to lockless flakes since these have that implicit contract that… well, they aren't locked :upside_down_face: 

> I dont know if it would even work for local subflakes as their inputs are eventually locked as well by the importing flake.

Eventually they are, ofc, but afaik Nix distinguishes between a flake that's locked “by itself” and locked as part of the parent flake

tomberek:
From discussion:
- can we limit this to lockless flakes? that and an experimental flag allows this to be easier to explore the design
- interacts with local flakes
- RFC?

charlesbaynham:
> For example, if I'm the author of a public flake that is developped using `nixpkgs/nixos-22.05`, should I set `inputs.nixpkgs.url = "nixpkgs/nixos-22.05"` (in which case downstream users can't use the `register` mechanism to override it), or should I leave it blank (in which case I can't easily lock it to the right version since it's gonna pick the `nixpkgs` from the registry which might not be what I want)? Neither case is fully satsifying.

Well true, but if my flake is picky enough about its inputs that using the latest nixpkgs will break it then specifying "22.05" helps communicate that to the user. If, on the other hand, I'm pinning my dependencies out of good practice but I don't expect severe compatibility problems in the future, I use "nixpkgs" so that later users can override my locked version with theirs.

nrdxp:
> That's not really my fear. My issue on the contrary is that this creates an incentive for publishing “incomplete” flakes

FWIW, I agree with this sentiment, but I would actually desire the implementation to go even further by not having `register` only act on for inputs that not definited explicitly, but for any input with the same name. So if I have a flake, and I set `inputs.nixpkgs.register = true`, I would want it to override every other input in the chain for any input named `nixpkgs` regardless if it's been specified or not. Maybe the change in semantic would require a change in name, so maybe something like `inputs.nixpkgs.override = true`?

In any case, something needs to be done here or there isn't much to do about #6626, assuming you do not have the authority to modify every upstream flake's input without forking.

roberth:
It seems that Nix has to finally incur a lot of the language-level dependency management complexity. These probably aren't problems that are solved by single PRs, but rather by collecting requirements and designing a _more_ comprehensive solution. Though it's just a thought.

blaggacao:
In the aforementioned spirit, I dropped this comment maybe in the wrong place: https://github.com/NixOS/nix/issues/6549#issuecomment-1169426244

PR #6564, opened 2022-05-23
distributed-builds: add link to source

asymmetric:
To point to additional arguments that can be passed as part of the store URI.

### DIFF
--- a/doc/manual/src/advanced-topics/distributed-builds.md
+++ b/doc/manual/src/advanced-topics/distributed-builds.md
@@ -22,6 +22,8 @@ specify an SSH identity file as part of the remote store URI, e.g.
 $ nix ping-store --store ssh://mac?ssh-key=/home/alice/my-key
 ```
 
+(For more store URI arguments, please consult the [source code](https://github.com/NixOS/nix/blob/master/src/libstore/ssh-store.cc))
+
 Since builds should be non-interactive, the key should not have a
 passphrase. Alternatively, you can load identities ahead of time into
 `ssh-agent` or `gpg-agent`.


edolstra:
Hm, I feel that telling people to read the source code is far from ideal...

thufschmitt:
There’s actually a `nix describe-stores` command that shows these arguments in a (very marginally atm, but it coud be greatly improved) nicer way, maybe you could suggest that instead

thufschmitt:
> Can the manual link to experimental commands though?

That's a good point, it wouldn't be wise indeed :)

> @thufschmitt the output of that command is.. interesting :)

The initial intent for that command was to use it in the build process to generate part of the documentation (like `__dump_builtins`). It never made it to that point – hence the half-baked markdown output and the fact that it's not documented – but if you feel like completing it that would be ideal

asymmetric:
@thufschmitt 

> but if you feel like completing it that would be ideal

I will look into that :) 

But that won't be a solution for the issue in this PR, so what should we do about that? Is a link to the source sufficient for now?

fricklerhandwerk:
@asymmetric here is an apt [comment on the subject of linking to source code by @roberth](https://github.com/NixOS/nixpkgs/issues/184982#issuecomment-1203718560). In this case I agree with @roberth that we should instead write down what there is instead of linking to the source.

asymmetric:
> we should instead write down what there is instead of linking to the source.

@fricklerhandwerk sure, that would be better. But in the meantime, doesn't this improve the status quo?

There's no harm in merging this and then replacing it with proper documentation IMO.

fricklerhandwerk:
I think writing down even part of those options should not take more time than discussing, and would prefer to set better examples instead of creating precedent.

asymmetric:
OK, closing this then :)

PR #6568, opened 2022-05-24
configure.ac: check for sandbox-shell's FEATURE_SH_STANDALONE

yorickvP:
See also: https://bugs.archlinux.org/task/73998.
Busybox's FEATURE_SH_STANDALONE feature causes other busybox applets to
leak into the sandbox, where system() calls will start preferring them over tools in $PATH. On arch, this even includes `ar`.

Let's check for this evil feature and disallow using this as a sandbox shell.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -294,7 +294,7 @@ esac
 AC_ARG_WITH(sandbox-shell, AS_HELP_STRING([--with-sandbox-shell=PATH],[path of a statically-linked shell to use as /bin/sh in sandboxes]),
   sandbox_shell=$withval)
 AC_SUBST(sandbox_shell)
-if ! test -z ${sandbox_shell+x}; then
+if test ${cross_compiling:-no} = no && ! test -z ${sandbox_shell+x}; then
   AC_MSG_CHECKING([whether sandbox-shell has the standalone feature])
   # busybox shell sometimes allows executing other busybox applets,
   # even if they are not in the path, breaking our sandbox
--- a/configure.ac
+++ b/configure.ac
@@ -294,6 +294,17 @@ esac
 AC_ARG_WITH(sandbox-shell, AS_HELP_STRING([--with-sandbox-shell=PATH],[path of a statically-linked shell to use as /bin/sh in sandboxes]),
   sandbox_shell=$withval)
 AC_SUBST(sandbox_shell)
+if ! test -z ${sandbox_shell+x}; then
+  AC_MSG_CHECKING([whether sandbox-shell has the standalone feature])
+  # busybox shell sometimes allows executing other busybox applets,
+  # even if they are not in the path, breaking our sandbox
+  if PATH= $sandbox_shell -c "busybox" 2>&1 | grep -qv "not found"; then
+    AC_MSG_RESULT(enabled)
+    AC_MSG_ERROR([Please disable busybox FEATURE_SH_STANDALONE])
+  else
+    AC_MSG_RESULT(disabled)
+  fi
+fi
 
 # Expand all variables in config.status.
 test "$prefix" = NONE && prefix=$ac_default_prefix


configure.ac
```diff
@@ -294,6 +294,17 @@ esac
 AC_ARG_WITH(sandbox-shell, AS_HELP_STRING([--with-sandbox-shell=PATH],[path of a statically-linked shell to use as /bin/sh in sandboxes]),
   sandbox_shell=$withval)
 AC_SUBST(sandbox_shell)
+if ! test -z ${sandbox_shell+x}; then
+  AC_MSG_CHECKING([whether sandbox-shell has the standalone feature])
+  # busybox shell sometimes allows executing other busybox applets,
+  # even if they are not in the path, breaking our sandbox
+  if PATH= $sandbox_shell -c "busybox" 2>&1 | grep -qv "not found"; then
```
alyssais, line 301: Won't this break cross?  Does busybox include its config file in a way we could check?  Otherwise, we'd need to have a way to opt out of this check or to disable it when cross-compiling.

alyssais (changes requested):

configure.ac
yorickvP: Good point! I'll disable this test when cross compiling.

yorickvP: Note that the sandbox-shell isn't even guaranteed to be busybox, but should be a statically linked single executable, so there's no way to really get any info from it without running it.

balsoft:
Just thinking out loud.

Is there even a point to having an actual shell as `/bin/sh` in the sandbox? Maybe we could get away with a stub that calls some real `sh` from PATH:

```C
#include <libgen.h>
#include <unistd.h>
#include <linux/limits.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <sys/types.h>
#include <sys/stat.h>

int main(int argc, char** argv) {
  char self[PATH_MAX] = {};
  char target[PATH_MAX + 3] = {}; // path/sh
  char realTarget[PATH_MAX] = {};
  char * nextPath = NULL;
  char * PATH = getenv("PATH");
  realpath("/proc/self/exe", self);
  char found = 0;

  nextPath = strtok(PATH, ":");

  while ( nextPath != NULL ) {

    strcpy(target, nextPath);
    strcat(target, "/sh");

    if (realpath(target, realTarget) != NULL) {
      struct stat st;
      if (stat(realTarget, &st) == 0 && st.st_mode & S_IXUSR) {
        if (strcmp(realTarget, self) != 0) {
          found = 1;
          break;
        }
      }
    }

    nextPath = strtok(NULL, ":");
  }

  if (!found) {
    fprintf(stderr, "%s: Failed to find an actual sh executable\n", argv[0]);
    return 1;
  }

  if (execv(realTarget, argv) == -1) {
    perror("Failed to exec a sh executable");
    return 1;
  }
}
```
This seems to be enough to build bash and stdenv (unless I'm missing something), and should work after bash is built.

alyssais:
> Is there even a point to having an actual shell as `/bin/sh` in the sandbox? Maybe we could get away with a stub that calls some real `sh` from PATH:

That sounds like it would cause even more impurities than the current approach (like the one Yorick is trying to fix here), no?

balsoft:
I don't think so; I'm talking about PATH that's set inside the sandbox, so it should be pure. As far as I can tell `/bin/sh` is actually never used for bootstrapping in nixpkgs, it is simply a convenience, so wherever it's used we almost universally have some `sh` in the `PATH` inside that derivation already, so why not reuse that instead of introducing an impurity?

yorickvP:
We'd still need to statically compile that binary, with musl or similar, so it wouldn't really save us a dependency (unless rustc can do this?).

balsoft:
The point is not "saving a dependency", the point is purity. WIth this, `/bin/sh` is now a pure part of the sandbox, since all Nix versions will now have the same file there. The exact shell being called will depend on the derivation being built, but that's precisely the idea here.

grawlinson:
Pinging @alerque as this also affects him.

This PR results in Arch having to remove `--with-sandbox-shell=/usr/bin/busybox` from compile flags and suggesting to users to insert `sandbox = false` in `nix.conf`. There are no statically linked shells in our repositories apart from busybox.

yorickvP:
I did find `dash-static-musl`, but it turns out to be in the AUR (and conflicts with `dash`). A possibility might be to depend the `nix` package on `musl` and build a static shell in `nix`'s `PKGBUILD`?

yorickvP:
Alternatively, it may be desirable to remove `FEATURE_STANDALONE` from the `busyboxx` package (afaict, it's always in `$PATH` when installed).

grawlinson:
I've added a `nix-busybox` package (just busybox without `FEATURE_STANDALONE`, installed to `/usr/lib/nix` exclusively for Nix), but it would be great if there was a better solution because this feels like a bit of a kludgy hack.

Issue #6509, opened 2022-05-09
Nixpkgs input is very out of date


cole-h:
**Describe the bug**

I make all of my flake inputs follow the Nixpkgs that my system is built against, so that I can be mostly sure everything is using the same package set. However, this recently broke. Specifically, when overriding the Nixpkgs input in this from nixos-21.05-small (!!! 21.11 has been out for a while, and 22.05 will be out "soon") to nixos-unstable, the `fetchGit` test fails.

**Steps To Reproduce**

1. Change the Nixpkgs input to point to nixos-unstable
2. Attempt to build
3. Observe test failure:

```
ran test tests/fetchGit.sh... [[31;1mFAIL[m]
    ++ [[ -n '' ]]
    ++ type -p git
    + [[ -z /nix/store/zx5pmkkgi0kszrcyc0mz7ypcsg7aggcv-git-2.36.0/bin/git ]]
    + clearStore
    + echo 'clearing store...'
    clearing store...
    + chmod -R +w /build/nix-test/tests/fetchGit/store
    + rm -rf /build/nix-test/tests/fetchGit/store
    + mkdir /build/nix-test/tests/fetchGit/store
    + rm -rf /build/nix-test/tests/fetchGit/var/nix
    + mkdir /build/nix-test/tests/fetchGit/var/nix
    + clearProfiles
    + profiles=/build/nix-test/tests/fetchGit/var/nix/profiles
    + rm -rf /build/nix-test/tests/fetchGit/var/nix/profiles
    + repo=/build/nix-test/tests/fetchGit/./git
    + export _NIX_FORCE_HTTP=1
    + _NIX_FORCE_HTTP=1
    + rm -rf /build/nix-test/tests/fetchGit/./git /build/nix-test/tests/fetchGit/./git-tmp /build/nix-test/tests/fetchGit/test-home/.cache/nix /build/nix-test/tests/fetchGit/worktree /build/nix-test/tests/fetchGit/shallow /build/nix-test/tests/fetchGit/minimal
    + git init /build/nix-test/tests/fetchGit/./git
    hint: Using 'master' as the name for the initial branch. This default branch name
    hint: is subject to change. To configure the initial branch name to use in all
    hint: of your new repositories, which will suppress this warning, call:
    hint: 
    hint: 	git config --global init.defaultBranch <name>
    hint: 
    hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
    hint: 'development'. The just-created branch can be renamed via this command:
[...]

PR #6582, opened 2022-05-27
Fix a segfault in the git fetcher

### LINKED ISSUE #6554: `nix build` for first-time home-manager build fails with address boundary error (SIGSEGV)
Original posts [here](https://www.reddit.com/r/NixOS/comments/ussfmg/nix_build_firsttime_homemanager_build_fails_with/) and [here](https://github.com/NixOS/nixpkgs/issues/173650) (very sorry about the spam)

### Describe the bug

So I recently wiped my laptop to change its filesystem from btrfs to ext4, hoping for a little performance improvement (every bit counts at its age). I was able to re-partition my disk and re-install NixOS smoothly on my laptop. (Here's [my NixOS config.](https://codeberg.org/harrisonthorne/nixos-config/))

But I cannot build my home-manager configuration.

I manage [my home-manager config](https://codeberg.org/harrisonthorne/dotfiles/) with flakes. To build it for the first time on a fresh system, I'm trying to use `nix build --no-link ~/.config/nixpkgs/#homeConfigurations.<machine_name>.activationPackage`. However, this fails with an address boundary error (SIGSEGV).

### Steps To Reproduce

1. You may need a fresh install NixOS with my system config (linked above). Also, enable `nix-command` and `flakes` experimental features.
2. Clone home-manager config (also linked above) to `~/.config/nixpkgs`
3. `cd ~/.config/nixpkgs`
4. `nix build --no-link ~/.config/nixpkgs#homeConfigurations.littlepony.activationPackage`

You should get something like `nix-build ... terminated by SIGSEGV (Address boundary error)`

### Expected behavior

I should be able to build my home-manager configuration successfully

### Additional context

```console
[user@system:~] $ nix-env --version
nix-env (Nix) 2.9.0pre20220512_d354fc3
```

I haven't found anything online about this. I've tried running it with `env GC_DONT_GC=1`, but that didn't fix the issue.

Here's [gdb output](https://pastebin.com/rHxqqc4S) and an [strace that might help.](https://pastebin.com/LUvUhPrP)

Thank you in advance for any help, workarounds, or suggestions :)

thufschmitt:
The git fetcher code used to dereference the (potentially empty) `ref`
input attribute. This was magically working, probably because the
compiler somehow outsmarted us, but is now blowing up with newer nixpkgs
versions.

Fix that by not trying to access this field while we don't know for sure
that it has been defined.

Fix #6554

@harrisonthorne if you have a way to reproduce the original issue, can you confirm that the fix is correct?

Related to #6509

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -449,11 +449,10 @@ struct GitInputScheme : InputScheme
             }
         }
 
-        const Attrs unlockedAttrs({
+        Attrs unlockedAttrs({
             {"type", cacheType},
             {"name", name},
             {"url", actualUrl},
-            {"ref", *input.getRef()},
         });
 
         Path repoDir;
@@ -466,6 +465,7 @@ struct GitInputScheme : InputScheme
                     head = "master";
                 }
                 input.attrs.insert_or_assign("ref", *head);
+                unlockedAttrs.insert_or_assign("ref", *head);
             }
 
             if (!input.getRev())
@@ -482,6 +482,7 @@ struct GitInputScheme : InputScheme
                     head = "master";
                 }
                 input.attrs.insert_or_assign("ref", *head);
+                unlockedAttrs.insert_or_assign("ref", *head);
             }
 
             if (auto res = getCache()->lookup(store, unlockedAttrs)) {


edolstra:
I fixed this differently in #6530 by making sure that we always have a `ref` at that point: https://github.com/NixOS/nix/blob/dfbb9a997fb1ab567f0dfbf2bf14263e0094d860/src/libfetchers/git.cc#L494

FlorianFranzen:
This is currently an issue in v2.8.1 and can be easily replicated e.g. by running:
```
nix eval --expr 'builtins.fetchGit { url = "https://github.com/mozilla/nixpkgs-mozilla.git"; rev = "15b7a05f20aab51c4ffbefddb1b448e862dccb7d"; }'
```

thufschmitt:
Yup' I agree with @zseri, #6530 doesn't look like it can be merged in a matter of days, so it seems important to have this fix land since it's pretty bad

Issue #3843, opened 2020-07-21
[flake]: provide builtins.currentSystem as a input to flakes


Mic92:
**Is your feature request related to a problem? Please describe.**

Right now one has to explicitly define system in flake outputs.
The nix flake itself already comes with boiler code like this:
 
```
outputs = { self } : let
   # ...
   forAllSystems = f: nixpkgs.lib.genAttrs systems (system: f system);
{
  defaultPackage = forAllSystems (system: self.packages.${system}.nix);;  
};
```

**Describe the solution you'd like**

```
outputs = { self, system } : {
  defaultPackage.${system} = self.packages.${system};  
};
```

**Describe alternatives you've considered**

Use boilercode in every project or rely on external libraries like https://github.com/numtide/flake-utils

If someone needs to explicitly specify platforms i.e. to build packages for different architectures with hydra this should be still possible.

Mic92:
cc @zimbatm who is involved in flake-utils

edolstra:
Not passing in `system` (or any other arguments) is intentional:

* It breaks hermetic evaluation: it makes the evaluation result depend on something external.
* It makes it impossible to enumerate the contents of the flake, since Nix doesn't know what the permissible values of the arguments are.
* The `packages` output should provide packages that work, not packages that might work.
* Evaluation caching becomes harder (not a huge problem for `system`, but would be a problem for arbitrary arguments).

Also, it's not clear that a `system` argument should be at the `outputs` level, since you may want to use outputs for different platforms in the same evaluation (e.g. on x86_64-linux, you might want to use some i686-linux derivations like `wine`).

zimbatm:
There are multiple usability issues with the current design:

The list of supported systems is currently internal to the flake. As a flake author, it is useful to be able to specify which systems are supported. As a flake user, it means that it is not possible to change that list without forking the repository. Especially once the flake ecosystem starts to grow, it will quickly become prohibitive to introduce new platforms.

The amount of boilerplate that is needed to specify all the permutations of systems for all the derivations. That's why I wrote flake-utils, to make that usage a bit less cumbersome. The outputs are mixing both system-dependent and independent keys so it's not as easy as passing the `system` as an argument.

Typing `x86_64-linux` is error-prone. When a typo happens, the feedback system is not able to point to the typo directly. The user will see that their system is not supported, and then have to make the mental connection to look for a typo.

Each flake ends-up re-initializing nixpkgs. This adds to the boilerplate and evaluation time.

---

[...]

PR #6583, opened 2022-05-27
Configurable derivations

edolstra:
WIP

To show all the options of a configuration:
```
# nix describe github:edolstra/configurables-test#nixosConfigurations2.test
```

To build a configuration:
```
# nix build github:edolstra/configurables-test#nixosConfigurations2.test
```

To override the host name from the command line:
```
# nix build github:edolstra/configurables-test#nixosConfigurations2.test --argstr networking.hostName xyzzy
```

To override the system type from the command line:
```
# nix build github:edolstra/configurables-test#nixosConfigurations2.test ---argstr nixpkgs.system i686-linux
```

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -599,14 +599,17 @@ InstallableFlake::InstallableFlake(
     Strings prefixes,
     const flake::LockFlags & lockFlags)
     : InstallableValue(state),
+      cmd(cmd),
       flakeRef(flakeRef),
       attrPaths(fragment == "" ? attrPaths : Strings{(std::string) fragment}),
       prefixes(fragment == "" ? Strings{} : prefixes),
       outputsSpec(std::move(outputsSpec)),
       lockFlags(lockFlags)
 {
+    #if 0
     if (cmd && cmd->getAutoArgs(*state)->size())
         throw UsageError("'--arg' and '--argstr' are incompatible with flakes");
+    #endif
 }
 
 std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableFlake::toDerivation()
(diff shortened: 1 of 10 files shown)


roberth:
Intriguing :)

### Mutually exclusive or mix-in?

I think the usability would be better of configurability was a mix-in interface rather than an entirely different `type`.
That way, expression can use the default configuration as usual. This is especially true if the wiring of the configuration happens in the flake fixpoint, so that `self.*` is always the configured thing.

### Overriding other parts of the flake

It'd also be more powerful to have this at the level of the flake fixpoint. This would allow the user to override a package in `packages` while deploying a `nixosConfigurations`.

A bonus feature would be to make these configurations first class, so they can be interpreted by the frameworks that build on Flakes, allowing the configuration of framework-specific concepts. This could be done by adding, say, `self.configurationOverrides.<attrName>.<configurableItem> = <configuredValue>`. Frameworks can declare that all possible `<attrName>`s are validated and handled by the framework by setting `outputs.nixConfig.checkConfigurationOverridesAttrNames = false` (or similar; maybe not `nixConfig`?), as this is already within their capability to do, without bothering the user.

### `nixosConfigurations` meaning

I've always understood the `nixosConfigurations` as the _final_ "real-world" configurations that are ready to be applied. Your example violates this understanding, making `nixosConfigurations` more like `nixosModules`.
We'd have:
 - `nixosModules`: services, image logic, profiles (I'd call them archetypes), complete but not instantiated system configs
 - `nixosConfigurations`: instantiated real systems, ***instantiated but not completely configured "real" systems***

### Configuring NixOS after the fact

While I'm not super happy about the fact that the `nixosConfigurations` are now almost conceptually redundant (if it wasn't for the choice of NixOS entrypoint (which source, which `lib.nixosSystem`-equivalent), I should point out a recent addition that will make your configurability a bit cleaner.

In the upcoming release (and recent backport to 21.11) it is possible to do

```nix
nixosConfigurations.foo = lib.nixosSystem { ..... };
nixosConfigurations.foo-with-hello = self.nixosConfigurations.extendModules { modules = [ { environment.systemPackages = [ pkgs.hello ]; } ]; };
```

This cleanly copies the effect of all `lib.nixosSystem` parameters to the new instance. That also fixes the little problem with `specialArgs` for you.

edolstra:
A bit off-topic, but I feel that `.extendModules` perpetuates the `.override` anti-pattern, where you have to call something (in this case a NixOS configuration) just to be able to override it.

> It'd also be more powerful to have this at the level of the flake fixpoint. This would allow the user to override a package in packages while deploying a nixosConfigurations.

I think this would require a change to the flakes spec, since it would mean that the flake `outputs` attribute has to be a configurable (i.e. implement an interface that allows its configuration options to be discoverable and settable). But it should be possible to do in a backwards-compatible way.

roberth:
> A bit off-topic, but I feel that `.extendModules` perpetuates the `.override` anti-pattern, where you have to call something (in this case a NixOS configuration) just to be able to override it.

In this case it is slightly better. It exists on a rather cheap value that does not require any options to be computed; not even imports to be resolved. Now this is going to look like a party trick, but it actually demonstrates that this extension mechanism is lazier than the usual `.override` ones. The crux: without a value for `later`, the module fixpoint can not be computed, yet we don't get an error, so it has not been computed before producing the final configuration.

```
nix-repl> ((lib.evalModules { prefix = []; modules = [ ({ later, ... }: { imports = [ later ]; }) ]; }).extendModules { specialArgs.later = { }; }).config
{ }
```

The reason why this mechanism is lazy is that unlike most overrides, the returned set off attribute names does not depend on the actual values. It is always `{ _module, config, options, extendModules, type }` so the only cost of overriding here is producing the other 4 thunks; next to nothing.

At first, I only intended `extendModules` for internal use in `specialisations` and such (through module arguments), but the surprisingly negligible cost made me reconsider that. In fact, this is the cheaper one, as when you use it through module arguments, you're by definition duplicating the configuration, as is required for such features as `specialisations`.

> But it should be possible to do in a backwards-compatible way.

While backcompat is generally nice, this should be no issue as flakes are experimental. I'm all for it though.

OT: A solution for https://github.com/NixOS/nix/issues/3843 can probably also be backwards compatible, but after accumulating a bunch of backcompat, it seems like a good idea to clear most of that out, before finalizing the RFC that specifies the first non-experimental flakes release.

roberth:
We've talked about generalizing this from output attributes to all outputs, but it could be generalized further to encompass the inputs as well. Doing so will solve https://github.com/NixOS/nixpkgs/issues/166220, or I guess this PR already does that, but only for the specific case of the `nixpkgs` flake; not yet other flakes that depend on `nixpkgs`.

At that point, we might consider setting `system` with the same means, thereby solving #3843 in a way, although such a solution would suggest being able to work with only a single system though, at least when ignoring the ability to invoke the good old nixpkgs function outside the flake paradigm. This could be a bit of a problem for dependency flakes that expose data (such as docs or a website) which isn't buildable everywhere, but can be used everywhere by virtue of being data (which is in a binary cache, presumably). There are other uses cases for cross-`system` dependencies as well, such as for natively built VM images.

edolstra:
> While backcompat is generally nice, this should be no issue as flakes are experimental. I'm all for it though.

Flake configurability is more of a flakes 2.0 feature. It shouldn't delay Nix 3.0.

domenkozar:
It would be cool if it accepted a JSON, as a file or a string.

src/nix/describe.cc
```diff
@@ -0,0 +1,85 @@
...
+        if (filter)
+            filterRegex = std::regex(*filter, std::regex::extended | std::regex::icase);
+
+        auto state = getEvalState();
+
+        auto cursor = installable->getCursor(*state);
+
+        auto type = cursor->getAttr(state->sType)->getString();
+
+        if (type != "configurable")
```
matthewbauer, line 45: Maybe it's overloading the concept, but it seems like you might expect `nix describe` to also work on flakes.

src/libcmd/installables.cc
```diff
@@ -615,42 +618,118 @@ std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableF
...
+                        if (auto pos2 = std::get_if<ValueTree>(&i->second))
+                            pos = pos2;
+                        else
+                            throw Error("definition of '%s' clashes with a previous definition", state->symbols[attr.name]);
+                    }
+                }
+            }
+        }
+
+        std::function<Value * (const ValueTree & tree)> buildAttrs;
```
shlevy, line 695: Might be worth a comment that this declared here for recursive calling, it surprised me.

src/libcmd/installables.cc
```diff
@@ -615,42 +618,118 @@ std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableF
...
 
-    return {attrPath, getLockedFlake()->flake.lockedRef, std::move(drvInfo)};
+        struct ValueTree
+        {
+            std::map<std::string, std::variant<Value *, ValueTree>> children;
+        };
+
+        ValueTree tree;
+
+        if (cmd) {
```
shlevy, line 673: If we want other things besides drvs/installables to be configurable, we should probably extract everything from here to [line 713](https://github.com/NixOS/nix/pull/6583/files#diff-71050ce7f0123d7c142fef6cf61d196f17109ebd49dbcf41211650d154cf5484R713) into a separate function that can be used for other interfaces.

src/libcmd/installables.cc
```diff
@@ -615,42 +618,118 @@ std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableF
...
+            }
+            auto vAttrs = state->allocValue();
+            vAttrs->mkAttrs(attrs);
+            return vAttrs;
+        };
+
+        auto vArgs = buildAttrs(tree);
+
+        auto vRes = state->allocValue();
+        state->callFunction(build, *vArgs, *vRes, noPos);
```
shlevy, line 713: Shouldn't the `pos` be wherever we found `attr`?

src/libcmd/installables.cc
```diff
@@ -615,42 +618,118 @@ std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableF
...
+            .outputsToInstall = std::move(outputsToInstall),
+            .priority = priority,
+        };
+
+        return {attrPath, getLockedFlake()->flake.lockedRef, std::move(drvInfo)};
     }
 
-    if (outputsToInstall.empty())
-        outputsToInstall.insert("out");
+    else if (type == "configurable") {
```
shlevy, line 660: If we want other things besides drvs/installables to be configurable, should we allow `type` to be some kind of structured value? E.g. `[ "configurable" "derivation" ]` or `{ configurable = true; configured = "derivation"; }` or something?

The alternative is to dynamically resolve this, i.e. check `type` after resolving the `configurable`.

shlevy (commented):
Overall looks pretty good! The one big picture question I have, referenced in a few of my comments: Why limit configurability to derivations? Why not configurable apps, configurable NixOS configurations, etc.?

I may pick this PR up and open a PR against your repo bringing it up to date with master and completing the work (documentation, testing, etc.)

ggPeti:
[made a workaround](https://github.com/ggPeti/param-pkg)

tomberek:

poelzi:
I'm using an ugly workaround. @edolstra can I adopt this ?

PR #6586, opened 2022-05-28
src/libstore/globals.hh: trusted-public-keys not needed for FODs

amjoseph-nixpkgs:
Nix will accept paths for FODs from a substituter without a signature. This is of course safe, since Nix knows and checks the hash of the path it is to receive.  Let's include this exception in the man page.

I noticed this while trying to build the `calibre` expression from nixpkgs.  It attempts to download [this patch](https://raw.githubusercontent.com/debian-calibre/calibre/debian/5.42.0%2Bdfsg-1/debian/patches/0007-Hardening-Qt-code.patch) which is no longer available at the specified URL (I get a 404).  I've had to do this several times in the past.

I prefer to build everything myself, from source.  Hydra isn't failing this build because it has the patch cached, possibly from a very long time ago.

From the manual page, I couldn't figure out how to tell Nix "you can substitute FODs, but nothing else".  It seems to me that setting `trusted-public-keys=""` but leaving the default `substituters=https://cache.nixos.org/` is how you accomplish this.  If this is incorrect please let me know and I will revise or close this PR.

Side question: is it necessary for the default `substituters` to use `https://` rather than `http://`?  Everything downloaded from `cache.nixos.org` ought to be FOD or else signed by the default `trusted-public-key`.  Just wondering.  Debian has always deliberately used `http://` for the same reason, to eliminate the TLS round-trip and to be robust in the face of certificate expiration and the large arsenal of X.509-infrastructure footguns.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -524,10 +524,9 @@ public:
         "trusted-public-keys",
         R"(
           A whitespace-separated list of public keys. When paths are
-          copied from another Nix store (such as a binary cache), and
-          are not part of a fixed-output derivation (such as a
-          fetcher), they must be signed with one of these keys. For
-          example:
+          copied from another Nix store (such as a binary cache), they must be
+          signed with one of these keys, unless they are content-addressed (like
+          the outputs of fixed-output derivations). For example:
           `cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
           hydra.nixos.org-1:CNHJZBh9K4tP3EKF6FkkgeVYsS3ohTl+oS0Qa8bezVs=`.
         )",
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -523,9 +523,11 @@ public:
         {"cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="},
         "trusted-public-keys",
         R"(
-          A whitespace-separated list of public keys. When paths are copied
-          from another Nix store (such as a binary cache), they must be
-          signed with one of these keys. For example:
+          A whitespace-separated list of public keys. When paths are
+          copied from another Nix store (such as a binary cache), and
+          are not part of a fixed-output derivation (such as a
+          fetcher), they must be signed with one of these keys. For
+          example:
           `cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
           hydra.nixos.org-1:CNHJZBh9K4tP3EKF6FkkgeVYsS3ohTl+oS0Qa8bezVs=`.
         )",


a-m-joseph:
> If this is incorrect please let me know and I will revise or close this PR.

I'm feeling a bit more sure of this, but it has been difficult to nail down the exact answers to (a) "when does nix consider something to be trusted" and (b) are fixed-output derivations considered to be content-addressed?

Starting from `LocalStore::pathInfoIsUntrusted()`, which calls `ValidPathInfo::checkSignatures()`.  That function short-circuits immediately reporting `maxSigs`-level trust if `isContentAddressed()`.  `ValidPathInfo::isContentAddressed()` mainly checks to see if `ca` is null; if so, immediately return `false`, if not, do some validation.  So now I check to see how `ca` acquires a non-`null` value.  The `PathInfo.ca` field is assigned at several points in the codebase, but always appears to be assigned by an expression of the form `ca = parseContentAddressOpt(...)`.  `parseContentAddressOpt()` calls `parseContentAddress()`, which invokes some kind of iterator (my C++ is rusty) which appears to accept `FixedOutputHashMethod`s, returning a valid `ContentAddress` for these.

Quite a journey!

So I am led to assume the following:

> Nix considers the outputs of a fixed-output derivation to be content-addressed.

It would probably help if the documentation stated this explicitly: fixed-output derivations' outputs are one particular kind of content-addressed store path.  This is true even if `--experimental-features content-addressed` is not provided.

This raises even more questions, however.  Running `nix-store make-content-addressable` on an FOD output will create a new output!  But running `nix-store make-content-addressable` again on that new output will *not* create a new store path; it's idempotent at that point.

Is there some command I've missed which shows me all the gory details of what Nix knows about a particular store-path?  Clearly it's keeping a lot of extra metadata, like [this](https://github.com/NixOS/nix/blob/ec07a70979a86cc436de7e46e03789b4606d25ab/src/libstore/store-api.cc#L81) but I couldn't find a `nix` command invocation that would show me all those internals.  Kind of like `btrfs inspect-internal` or `nix show-derivation`.

src/libstore/globals.hh
```diff
@@ -523,9 +523,11 @@ public:
...
         "trusted-public-keys",
         R"(
-          A whitespace-separated list of public keys. When paths are copied
-          from another Nix store (such as a binary cache), they must be
-          signed with one of these keys. For example:
+          A whitespace-separated list of public keys. When paths are
+          copied from another Nix store (such as a binary cache), and
+          are not part of a fixed-output derivation (such as a
+          fetcher), they must be signed with one of these keys. For
+          example:
```
thufschmitt, line 530:
```suggestion
          copied from another Nix store (such as a binary cache), they must be
          signed with one of these keys, unless they are content-addressed (like
          the outputs of fixed-output derivations). For example:
```

Nix doesn't really care about how the path was built, it just has to be content-addressed.
For instance you can `nix copy --from https://cache.ngi0.nixos.org/ /nix/store/yvk5yl9fid0zlxqk1xvvzn787d8gbh00-emacs-27.2` and it will work just fine, although the path doesn't come from a FO derivation

a-m-joseph:
> `nix path-info --json` is probably the closest to that

Ah, thank you so much, that's exactly what I was looking for.  I tried `nix path-info` but didn't know to add `--json`.

Ericson2314:
```git
<<<<<<< pr/docs/trusted-public-keys
          A whitespace-separated list of public keys. When paths are
          copied from another Nix store (such as a binary cache), they must be
          signed with one of these keys, unless they are content-addressed (like
          the outputs of fixed-output derivations). For example:
          `cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
          hydra.nixos.org-1:CNHJZBh9K4tP3EKF6FkkgeVYsS3ohTl+oS0Qa8bezVs=`.
=======
          A whitespace-separated list of public keys.

          At least one of the following condition must be met
          for Nix to accept copying a store object from another
          Nix store (such as a substituter):

          - the store object has been signed using a key in the trusted keys list
          - the [`require-sigs`](#conf-require-sigs) option has been set to `false`
          - the store object is [output-addressed](@docroot@/glossary.md#gloss-output-addressed-store-object)
>>>>>>> master
```
It would appear this has since been fixed? What do you think, @amjoseph-nixpkgs?

PR #6590, opened 2022-05-30
Fix noexcept violations

### LINKED ISSUE #6445: Nix does not degrade gracefully when a substituter dies

edolstra:
`HttpBinaryCacheStore::getFile()` shouldn't throw an exception since it's marked as `noexcept`. Fixes #6445.

Also ensure `RemoteStore::queryRealisationUncached()` doesn't throw an exception.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -348,7 +348,7 @@
 
             strictDeps = true;
 
-            passthru.perl-bindings = with final; currentStdenv.mkDerivation {
+            passthru.perl-bindings = with final; perl.pkgs.toPerlModule (currentStdenv.mkDerivation {
               name = "nix-perl-${version}";
 
               src = self;
@@ -378,7 +378,7 @@
               enableParallelBuilding = true;
 
               postUnpack = "sourceRoot=$sourceRoot/perl";
-            };
+            });
 
             meta.platforms = systems;
           };
--- a/flake.nix
+++ b/flake.nix
@@ -380,6 +380,7 @@
               postUnpack = "sourceRoot=$sourceRoot/perl";
             };
 
+            meta.platforms = systems;
           };
 
           lowdown-nix = with final; currentStdenv.mkDerivation rec {
--- a/src/libstore/http-binary-cache-store.cc
+++ b/src/libstore/http-binary-cache-store.cc
@@ -161,7 +161,12 @@ protected:
     void getFile(const std::string & path,
         Callback<std::optional<std::string>> callback) noexcept override
     {
-        checkEnabled();
+        try {
+            checkEnabled();
+        } catch (...) {
+            callback.rethrow();
+            return;
+        }
 
         auto request(makeRequest(path));
 
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -718,36 +718,34 @@ void RemoteStore::registerDrvOutput(const Realisation & info)
 void RemoteStore::queryRealisationUncached(const DrvOutput & id,
     Callback<std::shared_ptr<const Realisation>> callback) noexcept
 {
-    auto conn(getConnection());
-
-    if (GET_PROTOCOL_MINOR(conn->daemonVersion) < 27) {
-        warn("the daemon is too old to support content-addressed derivations, please upgrade it to 2.4");
-        try {
-            callback(nullptr);
-        } catch (...) { return callback.rethrow(); }
-    }
-
-    conn->to << wopQueryRealisation;
-    conn->to << id.to_string();
-    conn.processStderr();
-
-    auto real = [&]() -> std::shared_ptr<const Realisation> {
-        if (GET_PROTOCOL_MINOR(conn->daemonVersion) < 31) {
-            auto outPaths = worker_proto::read(
-                *this, conn->from, Phantom<std::set<StorePath>> {});
-            if (outPaths.empty())
-                return nullptr;
-            return std::make_shared<const Realisation>(Realisation { .id = id, .outPath = *outPaths.begin() });
-        } else {
-            auto realisations = worker_proto::read(
-                *this, conn->from, Phantom<std::set<Realisation>> {});
-            if (realisations.empty())
-                return nullptr;
-            return std::make_shared<const Realisation>(*realisations.begin());
-        }
-    }();
-
     try {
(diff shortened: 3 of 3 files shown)


thufschmitt (commented):
Looks good, except for the two apparently unrelated commits

Issue #4945, opened 2021-06-25
What language is flake.nix written in?


dasJ:
This issue may sound silly but it's actually not clear to me what language they are written in.
The syntax and file ending both look like Nix but the [manual](https://nixos.org/manual/nix/stable/#chap-writing-nix-expressions) doesn't apply most of the time.
The manual makes it seem like this should work:
```nix
{
  description = "Testflake";
  outputs = import ./outputs.nix;
}
```
but running that results in
```
error: expected a function but got a thunk at /nix/store/lm4pyqnkbiwvp2kxjd64nqb2xkdxmbfx-source/flake.nix:4:3
```

So what language are flake.nix files written in (is there any specification or documentation) and why is that language not Nix?

edolstra:
It's a subset of Nix that doesn't allow computation in the flake metadata attributes. So e.g. `outputs` cannot be a function application like `import ./outputs.nix`, it must be a function directly `outputs = { bla }: ...`. This is to prevent arbitrarily complex, possibly non-terminating computations while querying flake metadata.

This should be documented in the flake manpage.

dasJ:
Will care about this when the feature gets remotely usable for my use case so please leave this open

exarkun:
I ran into this and was quite surprised and confused for about half an hour.  I used `let ... in ...` at the top of my flake.nix and the resulting error message, "error: file '/nix/store/91d71n4sxkxnby3c3pbc8is4xm7a70r9-source/flake.nix' must be an attribute set" had me scratching my head (`builtins.typeOf` said it _evaluated_ to an attribute set) and looking all over for someplace I had put some non-set type into a set position.

I didn't believe the error message was worded accurately and precisely because I am used to Nix error messages that tell me one type was expected in place of another with little or no further location information and I thought this was one of those cases.

An error message something like "file '...' must be written using the limited flake expression language" might have helped me discover my mistake more quickly.

Also, none of the docs I read on my way to learning about flakes said anything about this limited expression language so apart from tripping over this error I'm not sure how I would have learned about this.

dzmitry-lahoda:
could be awesome if there would be turing decidable(always halt) security attack protected(like infinite reference circular import parsing) nix which allows to be flake input ` limited flake expression language` and that to be documented as flake input.

infinisil:
This issue has a better title and description, but it's a effectively a duplicate of https://github.com/NixOS/nix/issues/4945, which has a bunch more discussion.

rhendric:
[...]

PR #6592, opened 2022-05-30
Make flake metadata evaluation behaviour more discoverable

ckiee:
See commit messages. Builds and works for me in a `nix develop` shell.

### DIFF
--- a/src/nix/flake.md
+++ b/src/nix/flake.md
@@ -282,6 +282,14 @@ derivation):
 }
 ```
 
+
+
+`flake.nix` is written in a limited subset of the Nix language
+in order to prevent metadata lookups from hanging; values must
+be written directly and indirect evaluations such as `let..in`
+are not permitted. This does not apply to the body of
+the `outputs` function.
+
 The following attributes are supported in `flake.nix`:
 
 * `description`: A short, one-line description of the flake.
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1022,7 +1022,9 @@ void EvalState::cacheFile(
         // computation.
         if (mustBeTrivial &&
             !(dynamic_cast<ExprAttrs *>(e)))
-            throw EvalError("file '%s' must be an attribute set", path);
+            throw EvalError("file '%s' must not require evaluation;"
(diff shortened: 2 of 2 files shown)


ckiee:
cc #4945

tomberek:
These commits seem to imply the entire flake.nix cannot perform evaluation. Closer would be to say that a flake.nix contains up to three attributes at the top-level: `inputs`, `nixConfig`, and `outputs`. Where inputs and nixConfig must be simple key-value attrsets with no function evaluation. Outputs must be a function definition for which the function body can perform any pure Nix evaluation (modulo the usage of `--impure`,`--restrict-eval`, etc).

thufschmitt (changes requested):
Thanks, that's a great addition :)

I'm a bit annoyed by the pointer to a github issue, though. If there's some valuable information in the issue that's missing from the manual, then it should be ported to the manual rather than linked from the source code

ckiee:
> I'm a bit annoyed by the pointer to a github issue, though. If there's some valuable information in the issue that's missing from the manual, then it should be ported to the manual rather than linked from the source code

That would be nice and it did annoy me a bit too but I'm not sure how to do that. `flake.md` is a manual page, not a reference guide or something else. It seems weird to include a `## Flake metadata evaluation` or whatnot in what seems like more of a casual introduction. (*"As an example, [...]"*)

If you have any ideas on how to make that fit in that'd be nice.

src/nix/flake.md
```diff
@@ -282,7 +282,10 @@ derivation):
 }
 ```
 
-The following attributes are supported in `flake.nix`:
+The following attributes are supported in `flake.nix`, but all attributes other than `outputs`
```
edolstra, line 285: This does not read very well ("The following attributes..." followed by two sentences that are *not* the list of attributes).  Better to phrase it like "Flake attribute must be written using a limited subset bla bla... The following attributes are supported:".

ckiee:
> We definitely shouldn't link to GitHub issues in error messages, except maybe if there is an actual bug, **which is not the case here.**

[Are you sure?](https://github.com/NixOS/nix/issues/6631)

tomberek:

PR #6595, opened 2022-05-31
localstore: specified -> expected in output

asymmetric:
Better explains that the expected output is the one we verify against.

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1297,11 +1297,11 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
             auto hashResult = hashSink.finish();
 
             if (hashResult.first != info.narHash)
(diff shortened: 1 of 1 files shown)


asymmetric:
Build error seems unrelated?

asymmetric:
Any feedback on this?

fricklerhandwerk (approved):

edolstra:
This was actually specifically changed in 869c0321ff4829f56e10316f401a0f9268c1a8b0 from "wanted" to "specified" to make clear that it refers to the hash specified in the Nix input. Not sure "expected" is better (expected by who)?

fricklerhandwerk:
Same goes for "specified" - by who?

This hints at a deeper issue in presenting the problem. Would it be possible to show where exactly it was specified?

tomberek:

PR #6602, opened 2022-06-01
Explain exactly what nix-upgrade nix does

cw789:
Supersedes https://github.com/NixOS/nix/pull/6601.

Pleas check if I interpreted that correctly.

### DIFF
--- a/src/nix/upgrade-nix.cc
+++ b/src/nix/upgrade-nix.cc
@@ -34,7 +34,7 @@ struct CmdUpgradeNix : MixDryRun, StoreCommand
 
     std::string description() override
     {
-        return "upgrade Nix to the latest stable version";
+        return "upgrade Nix to the stable version declared in Nixpkgs";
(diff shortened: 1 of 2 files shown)


abathur:
> - Upgrade Nix to the latest available stable version:
> - The latest available version is determined by
 [nix-fallback-paths.nix](https://github.com/NixOS/nixpkgs/raw/master/nixos/modules/installer/tools/nix-fallback-paths.nix).
> - Means in general there may be a even newer Nix version available.

This is using `available` in two different ways that self-contradict.

Maybe something like:

> Upgrade Nix to the stable version declared in nixpkgs:
> ...
> This stable version is defined in [nix-fallback-paths.nix](https://github.com/NixOS/nixpkgs/raw/master/nixos/modules/installer/tools/nix-fallback-paths.nix) and updated manually. It may not always be the latest tagged release.

Issue #6594, opened 2022-05-31
debugger issue in nixpkgs


Artturin:
**Describe the bug**
```
$ nix build ".#bash" --debugger --impure
error: file 'nixpkgs-overlays' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at /nix/store/v5hrin57i9hdcpw9xc1by71fasrg2lr2-source/pkgs/top-level/impure.nix:41:36:

           40|       isDir = path: builtins.pathExists (path + "/.");
           41|       pathOverlays = try (toString <nixpkgs-overlays>) "";
             |                                    ^
           42|       homeOverlaysFile = homeDir + "/.config/nixpkgs/overlays.nix";

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0pre20220503_9489b4b. Type :? for help.
```

if i add --impure

```
$ nix build ".#bash" --debugger --impure
error: file 'nixpkgs-overlays' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at /nix/store/v5hrin57i9hdcpw9xc1by71fasrg2lr2-source/pkgs/top-level/impure.nix:41:36:

           40|       isDir = path: builtins.pathExists (path + "/.");
           41|       pathOverlays = try (toString <nixpkgs-overlays>) "";
             |                                    ^
           42|       homeOverlaysFile = homeDir + "/.config/nixpkgs/overlays.nix";

Starting REPL to allow you to inspect the current state of the evaluator.

Welcome to Nix 2.9.0pre20220503_9489b4b. Type :? for help.
```

**Steps To Reproduce**

1. `nix build ".#bash" --debugger` in nixpkgs

**`nix-env --version` output**
`nix (Nix) 2.9.0pre20220503_9489b4b`

roberth:
This exception is caught by the `builtins.try` on the same line.
Perhaps `nixpkgs` could avoid the attempt altogether in most (all?) cases when:
 - pure mode detected (`builtins?currentSystem` perhaps)
 - `nixpkgs-overlays` is not present in `builtins.nixPath` (needs precise reimplementation of the lookup logic)

Workaround: use `:c` debugger command to continue.

bburdette:
Like roberth said you have to use `:c` to continue to the exception you're actually interested in.  

Unfortunately this exception is normal and  is happening every time you run most nix scripts that aren't toy examples.  

[...]

PR #6608, opened 2022-06-02
--ignore-try flag for debugger mode

bburdette:
This is one way to address #6594.  

Added an --ignore-try flag, that causes the debugger to skip exceptions that occur in a tryEval.  

Also it prints an extra message is an exception occurs during tryEval, and --ignore-try is not used.

### DIFF
--- a/src/libcmd/command.cc
+++ b/src/libcmd/command.cc
@@ -91,6 +91,12 @@ EvalCommand::EvalCommand()
         .description = "start an interactive environment if evaluation fails",
         .handler = {&startReplOnEvalErrors, true},
     });
+
+    addFlag({
+        .longName = "ignore-try",
+        .description = "ignore exceptions in try clauses during debug",
+        .handler = {&ignoreExceptionsDuringTry, true},
+    });
 }
 
 EvalCommand::~EvalCommand()
@@ -120,7 +126,10 @@ ref<EvalState> EvalCommand::getEvalState()
             ;
 
         if (startReplOnEvalErrors) {
-            evalState->debugRepl = &runRepl;        
+            evalState->debugRepl = &runRepl;
+        };
+        if (ignoreExceptionsDuringTry) {
+            evalState->ignoreTry = ignoreExceptionsDuringTry;
         };
     }
     return ref<EvalState>(evalState);
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -467,6 +467,7 @@ EvalState::EvalState(
     , debugRepl(0)
     , debugStop(false)
     , debugQuit(false)
+    , ignoreTry(false)
     , regexCache(makeRegexCache())
 #if HAVE_BOEHMGC
     , valueAllocCache(std::allocate_shared<void *>(traceable_allocator<void *>(), nullptr))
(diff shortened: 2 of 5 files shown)


bburdette:
What it looks like without --ignore-try and then with --ignore-try

image

bburdette:
I'm not a huge fan of having an extra flag you have to use all the time.  Potentially we could --ignore-try by default, and then require a --no-ignore-try flag to debug inside tryEval.  But that wouldn't be as discoverable, so I like that less.

yorickvP:
I think it would make sense to make this the default

bburdette:
This actually was the default initially, but [a case came up right away](https://github.com/NixOS/nix/pull/5416#issuecomment-1132273716) of an error not being found.

--ignore-try is more cumbersome, but also more discoverable.  You see the unwanted exception and a message telling you how to skip it.  The other way you wouldn't know an exception had occurred.  

Maybe the way to go would be to continue execution by default, but print something like:

`an exception occurred within a tryEval.  use --debug-try to debug these`

src/libexpr/primops.cc
```diff
@@ -846,11 +846,32 @@ static RegisterPrimOp primop_floor({
     .fun = prim_floor,
 });
 
+class Counter
```
edolstra, line 849: There already is a `MaintainCount` class in util.hh that does the same thing.

src/libexpr/eval.cc
```diff
@@ -832,7 +834,14 @@ void EvalState::runDebugRepl(const Error * error, const Env & env, const Expr &
         : nullptr;
 
     if (error)
-        printError("%s\n\n" ANSI_BOLD "Starting REPL to allow you to inspect the current state of the evaluator.\n" ANSI_NORMAL, error->what());
+    {
+        printError("%s\n\n", error->what());
+
+        if (trylevel > 0 && error->info().level != lvlInfo)
+            printError("This exception occurred in a try clause. use " ANSI_GREEN "--ignore-try" ANSI_NORMAL " to skip these.\n");
```
edolstra, line 841:
```suggestion
            printError("This exception occurred in a try clause. Use " ANSI_GREEN "--ignore-try" ANSI_NORMAL " to skip these.\n");
```

The term "try clause" isn't really a thing in Nix. Maybe better to say something like "inside a 'tryEval' call".

src/libcmd/command.cc
```diff
@@ -91,6 +91,12 @@ EvalCommand::EvalCommand()
         .description = "start an interactive environment if evaluation fails",
         .handler = {&startReplOnEvalErrors, true},
     });
+
+    addFlag({
+        .longName = "ignore-try",
+        .description = "ignore exceptions in try clauses during debug",
+        .handler = {&ignoreExceptionsDuringTry, true},
+    });
```
edolstra, line 99: Maybe better to make this a setting inside `EvalSettings`. That way it can be set in `nix.conf`.

src/libexpr/eval.cc
bburdette: good point, changed.

src/libexpr/primops.cc
bburdette: switched to MaintainCount

PR #6612, opened 2022-06-03
Make nix copy parallel again

thufschmitt:
`nix copy` used to do stuff in parallel. This got removed in https://github.com/NixOS/nix/pull/5048 (to allow batching the copies in the ssh store case), but both can actually be kept in parallel with a little effort.

This is currently incomplete as it
- ~~Disables the batch ssh copy~~
- ~~Removes the ability to copy CA paths from one store prefix to another (I'm curious what the use-case for this is)~~

cc @connorbaker


thufschmitt:
> This is currently incomplete as it...

Also the code is a bit of a mess right now

src/libstore/store-api.cc
```diff
@@ -998,106 +1078,39 @@ std::map<StorePath, StorePath> copyPaths(
...
+            srcStore.narFromPath(missingPath, sink);
+        });
+        pathsToCopy.push_back(std::pair{*info, std::move(source)});
+    }
 
-    dstStore.addMultipleToStore(*source, repair, checkSigs);
+    dstStore.addMultipleToStore(pathsToCopy, act, repair, checkSigs);
 
     #if 0
-    std::atomic<size_t> nrDone{0};
```
edolstra, line 1024: You can remove the `#if 0...#endif`.

ConnorBaker:
Thank you so much for doing this @thufschmitt! This change has been invaluable in helping me quickly copy things to a binary cache I set up for CI/CD.

src/libstore/store-api.cc
Ericson2314: I am not sure we should remove this. When @edolstra did commented this out, we lost the ability to native `cp` for filesystem -> filesystem stores, and are forced to serialize and deserialize everything.

I think that is bad, and that functionality should be restored.

Overall, want something like "multiple dispatch" for `copyPaths`, where based on both the source and dest stores we can choose more efficient vs more general methods.

src/libstore/remote-store.cc
```diff
@@ -671,6 +671,23 @@ void RemoteStore::addToStore(const ValidPathInfo & info, Source & source,
 }
 
 
+void RemoteStore::addMultipleToStore(
```
edolstra, line 674: Shouldn't `LegacySSHStore` do the same thing?

edolstra:
Fixed a subtle hang in f0358ed4650e4608a383bd9f59ee23545f86c4ad.

Ericson2314:
Interesting bug! So this is something that `source`'s own destructor doesn't do? Is that intended?

edolstra:
The problem is that `source`'s destructor doesn't run until the unique_ptr is released.

Ericson2314:
Oh thanks I see now.

I think it might be better to *move* the unique pointer out of the map (leaving behind a null one, I suppose). And then we get the right destructor call by default.

edolstra:
@Ericson2314 You're right, in fact moving it was necessary to get it to build on macOS.

Issue #5759, opened 2021-12-10
[Flake] error: 'file:///tmp/测试' is not a valid URL


Vonfry:
**Describe the bug**
If a flake is under a dir with non english characters, the error is out when running `nix flake lock`.

**Steps To Reproduce**
```shell
cd /tmp
mkdir 测试
cd 测试
git init
echo '{ inputs.nixpkgs.url = "nixpkgs";  outputs = { ... }: {  }; }' > flake.nix
nix flake lock
```
or
```shell
cd /tmp
mkdir 测试
echo '{ inputs.nixpkgs.url = "nixpkgs";  outputs = { ... }: {  }; }' > flake.nix
nix flake lock ./测试 # P.S. nix flake lock ./test can work
```

**Expected behavior**
It can work.

**`nix-env --version` output**
nix-env (Nix) 2.4 (nixpkgs: c6fc79a2ac611b58ff93574cf787260ed2101d06)

**Additional context**
The problem is caused by flakeUrl and parseFlakeRefWithFragment probably

cc: @NickCao

regnat:
The error (`/tmp/测试 is not a valid URL`) is actually thrown in [parseURL](https://github.com/NixOS/nix/blob/c7b66caaca0833f0b88f3a93cbeaad73b4da1089/src/libutil/url.cc#L53). I tried changing the corresponding regex to replace the `a-zA-Z0-9` character ranges by `[:alnum:]` thinking that it would probably be utf-8 aware (as long as the locale is), but it didn’t seem to change anything.

kidonng:
> I tried changing the corresponding regex to replace the `a-zA-Z0-9` character ranges by `[:alnum:]` thinking that it would probably be utf-8 aware (as long as the locale is), but it didn’t seem to change anything.

Did some rudimentary test and it seems to work:
	
Run following code online: https://coliru.stacked-crooked.com/a/7966647daeafabab

Input:

```cpp
#include <iostream>
#include <locale>

int main() {
    const wchar_t c = L'\u86e4'; // Ideograph clam CJK
 
    std::locale loc1("C");
    std::cout << "isalnum('蛤', C locale) returned "
               << std::boolalpha << std::isalnum(c, loc1) << '\n';
 
    std::locale loc2("en_US.UTF-8");
    std::cout << "isalnum('蛤', Unicode locale) returned "
              << std::boolalpha << std::isalnum(c, loc2) << '\n';
}
```

Outputs:

```
[...]

Issue #4563, opened 2021-02-21
Can't have unicode characters in URL


luc65r:
**Describe the bug**

`parseURL` rejects unicode characters in URLs, which means you can't use `nix develop` for example.
```
error: --- BadURL -------------------------------------- nix
'file:///home/lucas/Documents/École/test' is not a valid URL
```

**Steps To Reproduce**

1. Make a `flake.nix` with a `devShell` in a path containing unicode characters
2. `nix develop`
3. Error

**Expected behavior**

I guess `parseURL` should allow unicode characters, at least in file urls.

**`nix-env --version` output**
`nix-env (Nix) 2.4pre20201205_a5d85d0`

**Additional context**

thufschmitt:
Is this fixed by https://github.com/NixOS/nix/pull/6614 ? I'm not entirely sure, but it might be

PR #6614, opened 2022-06-03
Implement support for percent encoded filepaths for flakerefs

### LINKED ISSUE #6394: [flakes / tryFetchGit] spaces not supported - error: "/home/user/project with space in name" is not a valid url
**Describe the bug**

Spaces in flake urls are not supported, leading to issues for local flakes that are located under a path containing spaces.

**Steps To Reproduce**

1. Create temporary project directory and initialise a git repo: `mkdir "/tmp/project name with space" && cd "/tmp/project name with space" && git init`
2. Create a `flake.nix` file (content does not matter):
```nix
{
  description = "demo broken flake with space in url";
  inputs.nixpkgs.url = "github:NixOs/nixpkgs";
  inputs.flake-utils.url = "github:numtide/flake-utils";
  inputs.flake-compat = {
    url = "github:edolstra/flake-compat";
    flake = false;
  };
  outputs = { self, nixpkgs, flake-utils, flake-compat, ... }@inputs:
    flake-utils.lib.eachDefaultSystem (system:
      let
        pkgs = import nixpkgs {
          inherit system;
        };
      in rec {
        devShell = pkgs.mkShell {
          buildInputs = [ ];
          nativeBuildInputs = (with pkgs; [
          ]);
        };
      }
    );
}
```
3. Add `flake.nix` to repo so nix finds it: `git add flake.nix`
4. Try to enter the development environment for the repo/flake: `nix develop` and observe following error:
```shell
error: 'file:///tmp/project name with space' is not a valid URL
```

**Expected behavior**

Get the development shell for the flake.

nix-env (Nix) 2.4

**Additional context**

- I found out the issue lies withing fetchGit by using a [flake-compat](https://github.com/edolstra/flake-compat) `nix.shell` file and running `nix-shell --show-trace`:
```shell
error: 'file:///tmp/project name with space' is not a valid URL

       … while evaluating 'tryFetchGit'

       at /nix/store/4gl05i14khqxwd5c1c1pfy18401x7l52-source/default.nix:93:19:

           92|     # tree is a valid git repository.
           93|     tryFetchGit = src:
             |                   ^
           94|       if isGit && !isShallow

       … from call site

       at /nix/store/4gl05i14khqxwd5c1c1pfy18401x7l52-source/default.nix:105:42:

          104|     { lastModified = 0; lastModifiedDate = formatSecondsSinceEpoch 0; }
          105|       // (if src ? outPath then src else tryFetchGit src);
             |                                          ^
          106|

       … while evaluating anonymous lambda

       at /nix/store/4gl05i14khqxwd5c1c1pfy18401x7l52-source/default.nix:134:13:

          133|     builtins.mapAttrs
          134|       (key: node:
             |             ^
          135|         let

       … from call site

       … while evaluating the attribute 'defaultNix'

       at /nix/store/4gl05i14khqxwd5c1c1pfy18401x7l52-source/default.nix:191:5:

          190|   rec {
          191|     defaultNix =
             |     ^
          192|       result
```
- the error occurs only if the flake is located in a git repository, it does work if the directory is *not* managed with git

RasmusRendal:
To support using nix flakes in paths with spaces, this introduces the convention that the `path` part of the URL should be percent-encoded when dealing with paths, and that it's the responsibility of fetchers that use a local path to decode it.

It follows the convention of firefox, which uses percent encoding when encoding a local path to an URL.

Eventually, this might also allow paths with arbitrary unicode characters (#4563, #5759), if the percent encoding and decoding methods are improved such that they can handle them.

This closes #6394.

I can get rid of the hijacked test if you'd prefer it.

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -1,1 +1,5 @@
 # Release X.Y (202?-??-??)
+
+- [URL flake references](@docroot@/command-ref/new-cli/nix3-flake.md#flake-references) now support [percent-encoded](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) characters.
+
+- [Path-like flake references](@docroot@/command-ref/new-cli/nix3-flake.md#path-like-syntax) now accept arbitrary unicode characters (except `#` and `?`).
--- a/src/libexpr/flake/flakeref.cc
+++ b/src/libexpr/flake/flakeref.cc
@@ -69,126 +69,129 @@ std::optional<FlakeRef> maybeParseFlakeRef(
     }
 }
 
-std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
+std::pair<FlakeRef, std::string> parsePathFlakeRefWithFragment(
     const std::string & url,
     const std::optional<Path> & baseDir,
     bool allowMissing,
     bool isFlake)
 {
-    using namespace fetchers;
+    std::string path = url;
+    std::string fragment = "";
+    std::map<std::string, std::string> query;
+    auto pathEnd = url.find_first_of("#?");
+    auto fragmentStart = pathEnd;
+    if (pathEnd != std::string::npos && url[pathEnd] == '?') {
+        fragmentStart = url.find("#");
+    }
+    if (pathEnd != std::string::npos) {
+        path = url.substr(0, pathEnd);
+    }
+    if (fragmentStart != std::string::npos) {
+        fragment = percentDecode(url.substr(fragmentStart+1));
+    }
+    if (pathEnd != std::string::npos && fragmentStart != std::string::npos) {
+        query = decodeQuery(url.substr(pathEnd+1, fragmentStart));
+    }
+
+    if (baseDir) {
+        /* Check if 'url' is a path (either absolute or relative
+            to 'baseDir'). If so, search upward to the root of the
+            repo (i.e. the directory containing .git). */
+
+        path = absPath(path, baseDir);
+
+        if (isFlake) {
+
+            if (!allowMissing && !pathExists(path + "/flake.nix")){
+                notice("path '%s' does not contain a 'flake.nix', searching up",path);
+
+                // Save device to detect filesystem boundary
+                dev_t device = lstat(path).st_dev;
+                bool found = false;
+                while (path != "/") {
+                    if (pathExists(path + "/flake.nix")) {
+                        found = true;
+                        break;
+                    } else if (pathExists(path + "/.git"))
+                        throw Error("path '%s' is not part of a flake (neither it nor its parent directories contain a 'flake.nix' file)", path);
+                    else {
+                        if (lstat(path).st_dev != device)
+                            throw Error("unable to find a flake before encountering filesystem boundary at '%s'", path);
+                    }
+                    path = dirOf(path);
+                }
+                if (!found)
+                    throw BadURL("could not find a flake.nix file");
+            }
+
+            if (!S_ISDIR(lstat(path).st_mode))
+                throw BadURL("path '%s' is not a flake (because it's not a directory)", path);
+
+            if (!allowMissing && !pathExists(path + "/flake.nix"))
+                throw BadURL("path '%s' is not a flake (because it doesn't contain a 'flake.nix' file)", path);
+
+            auto flakeRoot = path;
+            std::string subdir;
+
+            while (flakeRoot != "/") {
+                if (pathExists(flakeRoot + "/.git")) {
+                    auto base = std::string("git+file://") + flakeRoot;
+
+                    auto parsedURL = ParsedURL{
+                        .url = base, // FIXME
+                        .base = base,
+                        .scheme = "git+file",
+                        .authority = "",
+                        .path = flakeRoot,
+                        .query = query,
+                    };
+
+                    if (subdir != "") {
+                        if (parsedURL.query.count("dir"))
+                            throw Error("flake URL '%s' has an inconsistent 'dir' parameter", url);
+                        parsedURL.query.insert_or_assign("dir", subdir);
+                    }
+
+                    if (pathExists(flakeRoot + "/.git/shallow"))
+                        parsedURL.query.insert_or_assign("shallow", "1");
+
+                    return std::make_pair(
+                        FlakeRef(fetchers::Input::fromURL(parsedURL), getOr(parsedURL.query, "dir", "")),
+                        fragment);
+                }
+
+                subdir = std::string(baseNameOf(flakeRoot)) + (subdir.empty() ? "" : "/" + subdir);
+                flakeRoot = dirOf(flakeRoot);
+            }
+        }
+
+    } else {
+        if (!hasPrefix(path, "/"))
+            throw BadURL("flake reference '%s' is not an absolute path", url);
+        path = canonPath(path + "/" + getOr(query, "dir", ""));
+    }
+
+    fetchers::Attrs attrs;
+    attrs.insert_or_assign("type", "path");
+    attrs.insert_or_assign("path", path);
+
+    return std::make_pair(FlakeRef(fetchers::Input::fromAttrs(std::move(attrs)), ""), fragment);
+};
+
+
+/* Check if 'url' is a flake ID. This is an abbreviated syntax for
+    'flake:<flake-id>?ref=<ref>&rev=<rev>'. */
+std::optional<std::pair<FlakeRef, std::string>> parseFlakeIdRef(
+    const std::string & url,
+    bool isFlake
+)
+{
+    std::smatch match;
 
     static std::regex flakeRegex(
         "((" + flakeIdRegexS + ")(?:/(?:" + refAndOrRevRegex + "))?)"
         + "(?:#(" + queryRegex + "))?",
         std::regex::ECMAScript);
-
-    std::smatch match;
-
-    auto parsePathFlakeRef = [&]() {
-        std::string path = url;
-        std::string fragment = "";
-        std::map<std::string, std::string> query = {};
-        auto pathEnd = url.find_first_of("#?");
-        auto fragmentStart = pathEnd;
-        if (pathEnd != std::string::npos && url[pathEnd] == '?')
-            fragmentStart = url.find("#");
-        if (pathEnd != std::string::npos) {
-            path = url.substr(0, pathEnd);
-        }
-        if (fragmentStart != std::string::npos) {
-            fragment = percentDecode(url.substr(fragmentStart+1));
-        }
-        if (fragmentStart != std::string::npos && pathEnd != std::string::npos) {
-            query = decodeQuery(url.substr(pathEnd+1, fragmentStart));
-        }
-
-        if (baseDir) {
-            /* Check if 'url' is a path (either absolute or relative
-               to 'baseDir'). If so, search upward to the root of the
-               repo (i.e. the directory containing .git). */
-
-            path = absPath(path, baseDir);
-
-            if (isFlake) {
-
-                if (!allowMissing && !pathExists(path + "/flake.nix")){
-                    notice("path '%s' does not contain a 'flake.nix', searching up",path);
-
-                    // Save device to detect filesystem boundary
-                    dev_t device = lstat(path).st_dev;
-                    bool found = false;
-                    while (path != "/") {
-                        if (pathExists(path + "/flake.nix")) {
-                            found = true;
-                            break;
-                        } else if (pathExists(path + "/.git"))
-                            throw Error("path '%s' is not part of a flake (neither it nor its parent directories contain a 'flake.nix' file)", path);
-                        else {
-                            if (lstat(path).st_dev != device)
-                                throw Error("unable to find a flake before encountering filesystem boundary at '%s'", path);
-                        }
-                        path = dirOf(path);
-                    }
-                    if (!found)
-                        throw BadURL("could not find a flake.nix file");
-                }
-
-                if (!S_ISDIR(lstat(path).st_mode))
-                    throw BadURL("path '%s' is not a flake (because it's not a directory)", path);
-
-                if (!allowMissing && !pathExists(path + "/flake.nix"))
-                    throw BadURL("path '%s' is not a flake (because it doesn't contain a 'flake.nix' file)", path);
-
-                auto flakeRoot = path;
-                std::string subdir;
-
-                while (flakeRoot != "/") {
-                    if (pathExists(flakeRoot + "/.git")) {
-                        auto base = std::string("git+file://") + flakeRoot;
-
-                        auto parsedURL = ParsedURL{
-                            .url = base, // FIXME
-                            .base = base,
-                            .scheme = "git+file",
-                            .authority = "",
-                            .path = flakeRoot,
-                            .query = query,
-                        };
-
-                        if (subdir != "") {
-                            if (parsedURL.query.count("dir"))
-                                throw Error("flake URL '%s' has an inconsistent 'dir' parameter", url);
-                            parsedURL.query.insert_or_assign("dir", subdir);
-                        }
-
-                        if (pathExists(flakeRoot + "/.git/shallow"))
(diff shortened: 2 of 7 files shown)


src/libutil/url.hh
```diff
@@ -10,7 +10,7 @@ struct ParsedURL
     std::string base; // URL without query/fragment
     std::string scheme;
     std::optional<std::string> authority;
-    std::string path;
+    std::string path; // Percent encoded
```
edolstra, line 13: Wouldn't it be better to decode `path` in `parseURL()`? That way consumers of `path` don't have to remember to decode it, and avoids having to percent-encode in `parseFlakeRefWithFragment()`.

src/libutil/url.hh
```diff
@@ -22,6 +22,7 @@ struct ParsedURL
 MakeError(BadURL, Error);
 
 std::string percentDecode(std::string_view in);
+std::string percentEncode(std::string_view s, const char *skip="");
```
edolstra, line 25:
```suggestion
std::string percentEncode(std::string_view s, std::string_view skip);
```

and then use `skip.find(c)`.

src/libutil/url.hh
RasmusRendal: That is incredibly reasonable. Added it.

thufschmitt (approved):

thufschmitt:
Having another look at it, I'm a bit confused by the semantics: if `foo bar` is a flake, then `nix build "./foo bar"` will work (is that intended? You added ` ` to the allowed characters in a filepath but it's not documented), but not `nix build ./foo%20bar`. And conversely, `nix build "path:$PWD/foo bar"` won't work, but `nix build path:$PWD foo%20bar` will.

Also, can you add something to the release notes for this?

RasmusRendal:
> Having another look at it, I'm a bit confused by the semantics: if `foo bar` is a flake, then `nix build "./foo bar"` will work (is that intended? You added ` ` to the allowed characters in a filepath but it's not documented), but not `nix build ./foo%20bar`. And conversely, `nix build "path:$PWD/foo bar"` won't work, but `nix build path:$PWD foo%20bar` will.

These are my intended semantics at least. IMHO, users expect command-line tools to work with paths by default, and that is what they should do. If the user supplies something like `./directory`, then we should treat is as a directory, and not as a URL. If the user wants to work with URLs instead, then they should provide the prefix i.e. `path:$PWD/foo%20bar`.

thufschmitt:
> These are my intended semantics at least. IMHO, users expect command-line tools to work with paths by default, and that is what they should do. If the user supplies something like ./directory, then we should treat is as a directory, and not as a URL. If the user wants to work with URLs instead, then they should provide the prefix i.e. path:$PWD/foo%20bar.

Yes that kinda makes sense indeed… at least I can think of any better semantics. @edolstra any thought on that?

crasm:
Any update on if this can be merged? I was just hit by #6394 my first time trying to set up a flake-based project dev shell.

thufschmitt:
> Any update on if this can be merged?

Yes, I think it can, sorry about the long delay @RasmusRendal.

There are a few conflicts now, do you mind resolving them?

nomeata:
@RasmusRendal, are you still working on this PR or should someone take it over?

RasmusRendal:
> @RasmusRendal, are you still working on this PR or should someone take it over?

Yeah I don't think I'll have the time in the immediate future to get into the Nix codebase again. If somebody else would solve the merge conflicts, that would be very nice of them.

thufschmitt:
> > @RasmusRendal, are you still working on this PR or should someone take it over?
> 
> Yeah I don't think I'll have the time in the immediate future to get into the Nix codebase again. If somebody else would solve the merge conflicts, that would be very nice of them.

Ok, thanks for letting us know :+1: 

Unless someone is willing to beat me to it I can take over that one as it's an important fix (and also probably fixes #5759 which is of the same vein).

src/libexpr/flake/flakeref.cc
```diff
@@ -109,9 +103,14 @@ std::pair<FlakeRef, std::string> parseFlakeRefWithFragment(
             percentDecode(match.str(6)));
     }
 
-    else if (std::regex_match(url, match, pathUrlRegex)) {
-        std::string path = match[1];
-        std::string fragment = percentDecode(match.str(3));
+    else if (pathExists(url.substr(0, url.find("#")))) {
```
thufschmitt, line 106: Some notes (for myself mostly) on that change since it's a fairly crucial one:

- The `pathExists` call is potentially problematic because the flakeref should exist independently of what the filesystem contains. Maybe we could just split unconditionnally regardless of whether the path exist
- Splitting on `#` is probably not enough since the syntax of a path flakeref is of the form `<path>(\?<params>)?(#<fragments>)?`. So we'll want to split on `?` too
- I guess this isn't truly correct if multi-byte utf8 glyphs can contain `\63` (the ascii code for `?`) since we'd just split here. I think this isn't the case (precisely to allow this kind of things), but even if it's the case, it's probably better than the current situation

nomeata: > if multi-byte utf8 glyphs can contain \63 

I don't think they can, utf8 is carefully defined that the ascii bytes are only used for the coresponding character. All the other stuff has the highest bit set

thufschmitt:
I've updated this. @NixOS/nix-team, can someone have a look?

The actual change is 1c6d4514c93fea22e66d5d2adc7830b16a54facb, with 2b16210d02a2d675f9ce684d9dcb77c6c4c68691 being some extra cleanup that can be moved to a separate PR if needed (and the rest being tests and release notes).

fricklerhandwerk (commented):
This may need some user-visible documentation in the [flake references section](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#flake-references). One sentence and an example should be enough.

thufschmitt:
> This may need some user-visible documentation in the [flake references section](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#flake-references). One sentence and an example should be enough.

I started giving it a shot, but it turns out that the documentation for plain path flakerefs (`/foo/bar` or `./foo/bar`) is totally wrong already and would need to be rewritten. I've opened https://github.com/NixOS/nix/issues/8308 to track it.

Since rewriting that doc would have a slightly bigger scope, I'd be in favor of merging this PR first and fixing the docs later. Does that sound good?

fricklerhandwerk:
Yeah, I can live with a lower standard of documentation quality for experimental features.

nomeata:
I thought this was a bug fix, not an experimental features ;-)
(I'm eagerly looking forward to flakifying my shells in directories with spaces.)

thufschmitt:
I finally found the time to understand (and fix) the build failure and conflicts. Should be good to go now :boat:

TheSirC:
So we just need a reviewer to take a look at it now ?

fricklerhandwerk:
Triaged in Nix maintainers team meeting 2023-06-09:

- @fricklerhandwerk: please add a note to the flakeref docs. release notes are not enough
    - @thufschmitt: existing docs are already incorrect. will fix them up first in an independent PR
- @thufschmitt reviewed the code, but we're blocking this on https://github.com/NixOS/nix/issues/8482 to avoid piling up undocumented behavior

thufschmitt:
Rebased on top of #8640 to be able to document the new behavior (and marked as draft since it depends on it)

thufschmitt:
@fricklerhandwerk would you care to give this another look?

src/nix/flake.md
```diff
@@ -152,7 +153,7 @@ Currently the `type` attribute can be one of the following:
   [path:]<path>(\?<params)?
   ```
 
-  where *path* is an absolute path.
+  where *path* is an absolute path, percent-encoded according to the specifications of [rfc3986](https://github.com/NixOS/nix/issues/8482) if needed.
```
fricklerhandwerk, line 156:
```suggestion
  where *path* is an absolute path.
  Non-URI characters in the path are allowed in percent-encoding according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).
```

doc/manual/src/release-notes/rl-next.md
```diff
@@ -1 +1,4 @@
 # Release X.Y (202?-??-??)
+
+- URLs in Nix now support [percent-encoded](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) paths. The command line will percent-
+  encode non-URI characters automatically.
```
fricklerhandwerk, line 4:
```suggestion
- [Flake references](@docroot@/command-ref/new-cli/nix3-flake.md#flake-references) now support [percent-encoded](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) paths. The command line will percent-encode non-URI characters automatically.
```

src/nix/flake.md
```diff
@@ -98,6 +98,7 @@ Flakes corresponding to a local path can also be referred to by a direct path re
 The semantic of such a path is as follows:
 
 * If the directory is part of a Git repository, then the input will be treated as a `git+file:` URL, otherwise it will be treated as a `path:` url;
+* If the directory path contains characters which aren't allowed in uris, then they will be automatically percent-encoded (with the exception of `#` and `?`);
```
fricklerhandwerk, line 101:
```suggestion
* If the directory path contains characters that aren't allowed in URIs, then they will be automatically percent-encoded (with the exception of `#` and `?`) by the command line;
```

This is still a bit vague. Can we use references with non-URI-characters in `flake.nix`?

Also please add an example to the list of examples.

fricklerhandwerk (changes requested):

src/nix/flake.md
thufschmitt: The encoding doesn't happen in the CLI, but directly during the flakeref resolution (my bad, the release notes were wrong, I fixed them). So yes.

I've added an example of a contrieved path

doc/manual/src/release-notes/rl-next.md
thufschmitt: Included in a79affd6d (and fixed the bit about the command-line which is wrong)

src/nix/flake.md
thufschmitt: Included in a79affd6d

doc/manual/src/release-notes/rl-next.md
```diff
@@ -28,3 +28,5 @@
 - [`nix-store --query`](@docroot@/command-ref/nix-store/query.md) gained a new type of query: `--valid-derivers`. It returns all `.drv` files in the local store that *can be* used to build the output passed in argument.
 This is in contrast to `--deriver`, which returns the single `.drv` file that *was actually* used to build the output passed in argument. In case the output was substituted from a binary cache,
 this `.drv` file may only exist on said binary cache and not locally.
+
+- [Flake references](@docroot@/command-ref/new-cli/nix3-flake.md#flake-references) now support [percent-encoded](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) paths. [Path-like references](@docroot@/command-ref/new-cli/nix3-flake.md#path-like-syntax) percent-encode non-URI characters automatically.
```
fricklerhandwerk, line 32: To be honest, I don't fully understand what the new behavior is, and I'm worried this documentation is actively misleading.

So okay, we are now allowed to pass Unicode in file references. How is the percent encoding even relevant in that case? If I read that slice of code correctly it's an implementation detail that funnels the URL through from the CLI to the fetcher. We should not document that externally, it's just irrelevant to users.

Is that still correct?

```suggestion
- [Flake references](@docroot@/command-ref/new-cli/nix3-flake.md#flake-references) now support arbitrary unicode characters.
```

src/nix/flake.md
```diff
@@ -107,6 +108,7 @@ The semantic of such a path is as follows:
 
 * `.`: The flake to which the current directory belongs to.
 * `/home/alice/src/patchelf`: A flake in some other directory.
+* `./../sub directory/with Ûñî©ôδ€`: A flake in another relative directory
```
fricklerhandwerk, line 111:
```suggestion
* `./../sub directory/with Ûñî©ôδ€`: A flake in another relative directory that has Unicode characters in its name.
```

src/nix/flake.md
```diff
@@ -98,6 +98,7 @@ Flakes corresponding to a local path can also be referred to by a direct path re
 The semantic of such a path is as follows:
 
 * If the directory is part of a Git repository, then the input will be treated as a `git+file:` URL, otherwise it will be treated as a `path:` url;
+* If the directory path contains characters that aren't allowed in uris, then they will be automatically percent-encoded (with the exception of `#` and `?`);
```
fricklerhandwerk, line 101:
```suggestion
* The directory path can contain characters that aren't allowed in URIs.
  Those will be handled in [percent-encoding](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) internally (with the exception of `#` and `?`);
```

src/nix/flake.md
```diff
@@ -154,6 +156,8 @@ Currently the `type` attribute can be one of the following:
 
   where *path* is an absolute path.
 
+  Non-URI characters in the path are allowed in percent-encoding according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).
```
fricklerhandwerk, line 159:
```suggestion
  * The directory path can contain characters that aren't allowed in URIs.
  Those will be handled in [percent-encoding](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) internally (with the exception of `#` and `?`);
```

If the purpose is to say that the behavior is the same for all path references, we should use the same expression here.

fricklerhandwerk (commented):
Sorry for the back and forth, but it was really unclear to me what this changes. If I understand it correctly now, we can simplify a lot.

doc/manual/src/release-notes/rl-next.md
fricklerhandwerk: Yeah makes sense.

thufschmitt:
@fricklerhandwerk I gave the documentation and release notes another pass, PTAL.

If that's still not clear, I'd like to take ½h or 1h to pair with you on it so that we can get this through and not have it wait forever

src/nix/flake.md
```diff
@@ -67,6 +67,13 @@ inputs.nixpkgs = {
 };
 ```
 
+Following [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1),
+characters outside of the allowed range (i.e. are neither [reserved
+character](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) or
+[unreserved
+characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)) must be
+percent-encoded.
```
fricklerhandwerk, line 75:
```suggestion
Following [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1),
characters outside of the allowed range (i.e. neither [reserved characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2)
nor [unreserved characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3))
must be percent-encoded.
```

src/nix/flake.md
```diff
@@ -103,10 +110,15 @@ The semantic of such a path is as follows:
     2. The filesystem root (/), or
     3. A folder on a different mount point.
 
+Contrary to URL-like reference, path-like flake references can contain
+arbitrary unicode characters (except `#` and `?`).
```
fricklerhandwerk, line 114:
```suggestion
Contrary to URL-like references, path-like flake references can contain arbitrary unicode characters (except `#` and `?`).
```

fricklerhandwerk (approved):
Thanks a lot, makes sense now. Slight grammatical fixes, otherwise good to merge.

crasm:
Might just be a CI issue? Quite literally ran for an hour.

```
2023-09-25T07:57:06.1198552Z Requested labels: ubuntu-latest
[...]
2023-09-25T08:56:32.3792330Z building '/nix/store/jifwp7plfdh9kxqy30619k6j0a7py2n3-install-tests.drv'...
2023-09-25T08:57:25.0614911Z ##[error]The operation was canceled.
```

PR #6619, opened 2022-06-06
Add security.csm to ignored-acls

Jonpez2:
The security.csm ACL is, as far as I know, never reasonable to remove, so let's add it to the ignore-list in the vanilla nix image.  This makes this image usable on GKE.

I was considering adding a couple of others that seem to be generally added - security.selinux and system.nfs4_acl - but I thought I'd stick with the minimal.  Happy to add them if people think it's better!

Thanks!

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -125,7 +125,6 @@ let
     sandbox = "false";
     build-users-group = "nixbld";
     trusted-public-keys = "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=";
-    ignored-acls = "security.csm";
   };
   nixConfContents = (lib.concatStringsSep "\n" (lib.mapAttrsFlatten (n: v: "${n} = ${v}") nixConf)) + "\n";
 
--- a/docker.nix
+++ b/docker.nix
@@ -125,6 +125,7 @@ let
     sandbox = "false";
     build-users-group = "nixbld";
     trusted-public-keys = "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=";
+    ignored-acls = security.csm;
   };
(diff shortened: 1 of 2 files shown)


edolstra:
If `security.csm` is never reasonable to remove, then it's better to add it to the default value of `ignored-acls` in `globals.hh` (which already contains `security.selinux` and `system.nfs4_acl`).

Jonpez2:
Cool I will submit a diff to the base ignore list thank you!

Security.csm is (as I understand it) a thing that allows gke to detect when
a binary gets run that was not part of the original container. It’s part of
a system called container threat detection.

docker.nix
```diff
@@ -125,6 +125,7 @@ let
     sandbox = "false";
     build-users-group = "nixbld";
     trusted-public-keys = "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=";
+    ignored-acls = security.csm;
```
cole-h, line 128: Is this still necessary, now that it's been added as a default to the `ingored-acls` setting? If so, there is no `security` attribute in scope, so it should probably be stringified:
```suggestion
    ignored-acls = "security.csm";
```

docker.nix
```diff
@@ -125,6 +125,7 @@ let
     sandbox = "false";
     build-users-group = "nixbld";
     trusted-public-keys = "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=";
+    ignored-acls = "security.csm";
```
edolstra, line 128: Hm, as @cole-h said, we should remove this since it's not needed. In fact it has the effect of overriding the other built-in ignored ACLs (which could be avoided by using `extra-ignored-acls`).
```suggestion
```

Jonpez2:
Yes completely agree - I'm sorry, I got halfway through fixing this up and then dropped off the grid for a couple of days :(
I've committed your suggestion, thank you very much.

Issue #6561, opened 2022-05-22
print-missing option does not work with `nix build`


maralorn:
**Describe the bug**

The new experimental commands like `nix build` or `nix develop` do not call printMissing, even when `print-missing = true` (the default). 

**Steps To Reproduce**

1. Run `nix build --log-format raw` or `nix-build --log-format internal-json` for any package or flake.

**Expected behavior**

I would expect to see the list of packages which nix plans to build or download, since print-missing is set to true.

Alternatively, I would like the documentation of the print-missing option to state that it is not supposed to work with the new experimental commands, although that wouldn‘t solve my problem.

Note: When running `nix-build --log-format internal-json` the output of printMissing appears as a json message. So this is not an issue about the specific log-format, but about the difference between `nix-build` and `nix build`. The difference is, that in the default log-format of `nix build` the user wouldn’t notice the message, so they don’t miss it.

**`nix-env --version` output**

This happens on every nix version I know of.

**Additional context**

I am trying to make nix-output-monitor work with `nix build --log-format internal-json`. Sadly, when used with `nix build` the internal-json is missing a lot of information which the user could read from the output of nix-build. This makes figuring out the dependency tree which nix is planning to build impossible for nix-output-monitor.

thufschmitt:
> The new experimental commands like nix build or nix develop do not call printMissing, even when print-missing = true (the default).

Yup' it doesn't indeed. `--dry-run` does though if that can help.

> I am trying to make nix-output-monitor work with nix build --log-format internal-json. Sadly, when used with nix build the internal-json is missing a lot of information which the user could read from the output of nix-build. This makes figuring out the dependency tree which nix is planning to build impossible for nix-output-monitor.

Which kind of information are you missing? I think this could be added to the logger (and that might be a better long-term choice since the output of `nix build` is likely to change in the future)

maralorn:
With logger you mean the "internal-json" output or something else?

The information that nix-output-monitor needs, but currently cannot easily get from internal-json are in descending importance:

1. What are the targets of the current build, i.e. the derivations which don’t get build because they are depended on but because they are the end goal of the build? (It’s not important to know that they are the end goals, but I really need to know that they will be build because they are the roots of the build tree nix-output-monitor displays.)
2. What other derivations will be build?
3. What paths will be fetched?

If nix could improve the internal-json or another output method a little bit to give those information, that would be awesome!

thufschmitt:
[...]

PR #6620, opened 2022-06-06
respect print-missing variable in new-style build command

maralorn:
This addresses https://github.com/NixOS/nix/issues/6561

Currently nix-build prints the "printMissing" information by default,
nix build doesn’t. People generally don‘t notice this because the
standard log-format of nix build would not display the printMissing
output long enough to display the information.

Summoning @thufschmitt

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -919,6 +919,9 @@ std::vector<std::pair<std::shared_ptr<Installable>, BuiltPath>> Installable::bui
         break;
 
     case Realise::Outputs: {
+        if (settings.printMissing)
+          printMissing(store, pathsToBuild, lvlInfo);
+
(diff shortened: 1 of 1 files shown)


thufschmitt (approved):

edolstra:
I think this is more-or-less intended behaviour, in order to make the new CLI less spammy than the old one. The progress-bar branch makes it possible to interactively see what paths are still remaining to be built, which is probably better than *always* printing the missing paths.

maralorn:
@edolstra I don‘t think spammyness is an issue. I have tested this with all available log-formats and the missing builds message does not appear in any of them without -v. I think seeing this information with -v seems valuable and appropriate. Also, the progress-bar is not merged, yet.

edolstra:
Though I guess it's okay if it only happens at `-v`...

PR #6621, opened 2022-06-06
Fix nested flake input overrides

### LINKED ISSUE #5790: Transitive inputs "not in registry"

Kha:
Currently, `B` in `inputs.A.inputs.B.inputs.C...` is interpreted as registry flake, discarding any of its configuration in `A`, leading to an [unexpected](https://github.com/NixOS/nix/issues/5790) "cannot find flake 'flake:B' in the flake registries". This PR implements the expected behavior:
* input overrides (those nested inside `inputs`) are not defaulted to registry flakes.
* Input overrides do not outright replace the corresponding input of the nested flake, but are recursively merged into it, meaning e.g. `inputs.B.url` set in `A` is preserved by the above config.

Fixes #5790

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -90,11 +90,11 @@ static void expectType(EvalState & state, ValueType type,
 
 static std::map<FlakeId, FlakeInput> parseFlakeInputs(
     EvalState & state, Value * value, const PosIdx pos,
-    const std::optional<Path> & baseDir, InputPath lockRootPath);
+    const std::optional<Path> & baseDir, InputPath lockRootPath, unsigned depth);
 
 static FlakeInput parseFlakeInput(EvalState & state,
     const std::string & inputName, Value * value, const PosIdx pos,
-    const std::optional<Path> & baseDir, InputPath lockRootPath)
+    const std::optional<Path> & baseDir, InputPath lockRootPath, unsigned depth)
 {
     expectType(state, nAttrs, *value, pos);
 
@@ -118,7 +118,7 @@ static FlakeInput parseFlakeInput(EvalState & state,
                 expectType(state, nBool, *attr.value, attr.pos);
                 input.isFlake = attr.value->boolean;
             } else if (attr.name == sInputs) {
-                input.overrides = parseFlakeInputs(state, attr.value, attr.pos, baseDir, lockRootPath);
+                input.overrides = parseFlakeInputs(state, attr.value, attr.pos, baseDir, lockRootPath, depth + 1);
             } else if (attr.name == sFollows) {
                 expectType(state, nString, *attr.value, attr.pos);
                 auto follows(parseInputPath(attr.value->string.s));
@@ -163,7 +163,11 @@ static FlakeInput parseFlakeInput(EvalState & state,
             input.ref = parseFlakeRef(*url, baseDir, true, input.isFlake);
     }
 
-    if (!input.follows && !input.ref)
+    if (!input.follows && !input.ref && depth == 0)
+        // in `input.nixops.inputs.nixpkgs.url = ...`, we assume `nixops` is from
+        // the flake registry absent `ref`/`follows`, but we should not assume so
+        // about `nixpkgs` (where `depth == 1`) as the `nixops` flake should
+        // determine its default source
         input.ref = FlakeRef::fromAttrs({{"type", "indirect"}, {"id", inputName}});
 
     return input;
@@ -171,7 +175,7 @@ static FlakeInput parseFlakeInput(EvalState & state,
 
 static std::map<FlakeId, FlakeInput> parseFlakeInputs(
     EvalState & state, Value * value, const PosIdx pos,
-    const std::optional<Path> & baseDir, InputPath lockRootPath)
+    const std::optional<Path> & baseDir, InputPath lockRootPath, unsigned depth)
 {
     std::map<FlakeId, FlakeInput> inputs;
 
@@ -184,7 +188,8 @@ static std::map<FlakeId, FlakeInput> parseFlakeInputs(
                 inputAttr.value,
                 inputAttr.pos,
                 baseDir,
-                lockRootPath));
+                lockRootPath,
+                depth));
     }
 
     return inputs;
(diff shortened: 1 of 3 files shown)


src/libexpr/flake/flake.cc
```diff
@@ -404,6 +418,12 @@ LockedFlake lockFlake(
                         // Respect the “flakeness” of the input even if we
                         // override it
                         i->second.isFlake = input2.isFlake;
```
Kha, line 420: Note that this workaround could be replaced by a "proper" merge as now done with the other attributes below if `isFlake` had a value for "not specified", e.g. was changed to `std::optional<bool>`. But wanting to override the flakeness of an input does seem like an exotic use case.

tomberek:
Works as expected. There are a few changes in-flight to follows/overrides and the lazy-trees branch with which I'd wonder if there is any interaction or conflict.

tests/flakes.sh
```diff
@@ -845,3 +845,31 @@ nix store delete $(nix store add-path $badFlakeDir)
...
+cat <<EOF > $flakeFollowsA/flake.nix
+{
+  inputs.B.url = "path:./flakeB";
+  inputs.D.url = "path:./flakeD";
+  inputs.B.inputs.C.inputs.D.follows = "D";
+  outputs = _: {};
+}
+EOF
+
+nix flake lock $flakeFollowsA
```
thufschmitt, line 875: Can you move that to its own test file? The `flakes.sh` test is already infuriatingly slow (and totally sequential) and I wouldn't to expand it further

tests/flakes.sh
```diff
@@ -845,3 +845,31 @@ nix store delete $(nix store add-path $badFlakeDir)
...
+EOF
+cat <<EOF > $flakeFollowsC/flake.nix
+{
+  inputs.D.url = "path:nosuchflake";
+  outputs = _: {};
+}
+EOF
+cat <<EOF > $flakeFollowsB/flake.nix
+{
+  inputs.C.url = "path:../flakeC";
```
thufschmitt, line 862: That input doesn't seem right (`flakeC` is a subdir of `flakeB`). It's probably safer to use absolute paths (via the `$flakeFollows*` variables everywhere here

tests/flakes.sh
thufschmitt, line 875: I get the idea of the test (the flake lock can only succeed if the override is properly taken into account), but it would be nice to have a more explicit check, if only to make the intent clearer. Maybe something like

```
nix flake metadata $flakeFollowsA --json | jq --exit-status '.locks.nodes.C.inputs.D == [ "D" ]'
```

thufschmitt (changes requested):
Thanks for that, and sorry for the delay (I'm always a bit reluctant to touch that logic 😛).

Looks good overall, just a few suggestions, mostly around the tests

src/libexpr/flake/flake.cc
```diff
@@ -90,11 +90,11 @@ static void expectType(EvalState & state, ValueType type,
 
 static std::map<FlakeId, FlakeInput> parseFlakeInputs(
     EvalState & state, Value * value, const PosIdx pos,
-    const std::optional<Path> & baseDir, InputPath lockRootPath);
+    const std::optional<Path> & baseDir, InputPath lockRootPath, bool isOverride);
```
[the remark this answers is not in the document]
Kha: I can do that, but looking at the code again, it seemed more elegant to just pass the recursion depth given that `isOverride` corresponds to depth > 0. Does it look sensible?

tests/flakes.sh
Kha: Moved to the new `follows.sh`

tests/flakes.sh
Kha: done

tests/flakes.sh
Kha: done in the same style as in the other tests in the file

Kha:
I noticed and fixed a related bug, see new commit above. I added it to this PR to avoid merge conflicts in the test file, but happy to relocate.

tests/flakes/follow-paths.sh
```diff
@@ -178,3 +178,33 @@ EOF
...
+{
+  inputs.B.url = "path:$flakeFollowsB";
+  inputs.C.url = "path:$flakeFollowsC";
+  inputs.B.inputs.C.follows = "C";
+  outputs = _: {};
+}
+EOF
+
+# bug was not triggered without recreating the lockfile
+nix flake lock $flakeFollowsA --recreate-lock-file
```
Kha, line 208: It feels a bit concerning that one gets an entirely different lockfile without this flag, but that's probably a different problem

thufschmitt (approved):
Sorry for not coming back to this earlier. I'm only 95% confident that this isn't braking something (because I still couldn't wrap my head around the flake locking code), but since we're just at the beginning of a new release cycle, let's merge and revert if needs be. Thanks!

Issue #1548, opened 2017-09-01
nix-channel doesn't work as expected with sudo


LnL7:
On macOS it seems like nix-channel doesn't behave correctly when just using sudo.
I'm not sure if this also happens on other platforms, but it works fine on nixos.

```
$ sudo nix-channel --list
$ sudo -i nix-channel --list
nixpkgs http://nixos.org/channels/nixpkgs-unstable
```

and even more confusing

```
$ sudo nix-channel --add http://nixos.org/channels/foo
$ sudo nix-channel --update
unpacking channels...
created 3 symlinks in user environment
$ sudo nix-channel --list
foo http://nixos.org/channels/foo
$ sudo -i nix-channel --list
nixpkgs http://nixos.org/channels/nixpkgs-unstable
```

```
$ ls -l ~/.nix-defexpr
lrwxr-xr-x   1 root  staff   44 Sep  1 21:41 channels -> /nix/var/nix/profiles/per-user/root/channels
lrwxr-xr-x   1 root  staff   44 Jul 23 12:26 channels_root -> /nix/var/nix/profiles/per-user/root/channels
$ ls /nix/var/nix/profiles/per-user/root/channels
binary-caches manifest.nix  nixpkgs
$ ls /nix/var/nix/profiles/per-user/lnl/channels
manifest.nix
```

copumpkin:
cc @grahamc given that this changed with the multi-user setup

domenkozar:
cc@jmitchell

grahamc:
Doing some debugging, so far things are looking okay here:

```
nixos$ nix-channel --add https://nixos.org/channels/nixos-17.03 fizzbuzz
nixos$ nix-channel --update
downloading Nix expressions from ‘https://d3g5gsiof5omrk.cloudfront.net/nixos/17.03/nixos-17.03.1775.56da88a298/nixexprs.tar.xz’...
downloading ‘https://d3g5gsiof5omrk.cloudfront.net/nixos/17.03/nixos-17.03.1775.56da88a298/nixexprs.tar.xz’... [6975/9252 KiB, 6959.8 KiB/s]
unpacking channels...
nixos$ ls /nix/var/nix/profiles/per-user/grahamc/channels/           
binary-caches  fizzbuzz  manifest.nix



macos$ nix-channel --add https://nixos.org/channels/nixos-17.03 fizzbuzz
macos$ nix-channel --update
[...]

PR #6622, opened 2022-06-06
Don't read HOME from environment

mkenigs:
On darwin, sudo does not clear HOME, so HOME will still be set to a
user's home directory rather than root's.

util::getHome reads that value of HOME and is used in a number of
different places, which causes unexpected behavior in quite a few ways

See https://github.com/NixOS/nix/issues/1548 for some issues with
nix-channel

nix profile will symlink ~/.nix-profile to the default system profile

nix will create ~/.cache owned by root:
$ rm -r ~/.cache
$ sudo nix --extra-experimental-features "nix-command flakes" run nixpkgs#hello
Hello, world!
$ ls -ld ~/.cache
drwxr-xr-x 3 root staff 96 Jun  6 17:14 /Users/matthew/.cache

I'm guessing similar problems occur elsewhere getHome is used

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -573,17 +573,13 @@ Path getHome()
 {
     static Path homeDir = []()
     {
-        auto homeDir = getEnv("HOME");
-        if (!homeDir) {
-            std::vector<char> buf(16384);
-            struct passwd pwbuf;
-            struct passwd * pw;
-            if (getpwuid_r(geteuid(), &pwbuf, buf.data(), buf.size(), &pw) != 0
-                || !pw || !pw->pw_dir || !pw->pw_dir[0])
-                throw Error("cannot determine user's home directory");
-            homeDir = pw->pw_dir;
-        }
-        return *homeDir;
+        std::vector<char> buf(16384);
+        struct passwd pwbuf;
+        struct passwd * pw;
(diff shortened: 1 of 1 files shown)


mkenigs:
It is possible to just use `sudo -H` instead, so there is a workaround. Unless anyone is relying on overriding via `HOME` though it seems like it would be nicer for plain `sudo` to work

edolstra:
This will cause Nix to fail if the current user has no password entry, which could happen e.g. in a sandbox or user namespace. So I'm not sure this is a good idea.

mkenigs:
Would these be some better options?
1. use HOME if it is owned by current user, else fail
    - I think this is closest to preserving current behavior, but it would prevent users from doing something that's pretty clearly undesired like creating ~/.cache owned by root
    - Users would still have to use `sudo -H` on darwin
2. use HOME if it is owned by current user, else check for passwd entry, else fail
    - I don't love that this means the same command will behave differently depending on who invoked it. For option 1 you'll at least get an error explaining why it's behaving differently
    - Users could use plain `sudo`
3. check for passwd entry, if it doesn't exist use HOME if owned by current user, else fail
    - will still behave differently depending on the invoking user, but only if a user doesn't have a passwd entry

virusdave:
Ha, i was just about to start writing a similar patch.  Personally i'd vote for option (3), and this was what i was about to implement.

mkenigs:
Closing as https://github.com/NixOS/nix/pull/6676 was merged

PR #6628, opened 2022-06-08
More useful error when command not found on remote

balsoft:
On macOS and some versions of Linux, shells started in non-interactive
    mode don't execute their RC files, resulting in errors like

      zsh:1: command not found: nix-store

    This is confusing to new users.

    If the executable does exist somewhere we would expect the installer to
    put it, but that location is not in PATH, show a more detailed error
    instructing the user to add '&remote-program' to the URL.

### DIFF
--- a/src/libstore/legacy-ssh-store.cc
+++ b/src/libstore/legacy-ssh-store.cc
@@ -73,9 +73,12 @@ struct LegacySSHStore : public virtual LegacySSHStoreConfig, public virtual Stor
     ref<Connection> openConnection()
     {
         auto conn = make_ref<Connection>();
-        conn->sshConn = master.startCommand(
-            fmt("%s --serve --write", remoteProgram)
-            + (remoteStore.get() == "" ? "" : " --store " + shellEscape(remoteStore.get())));
+        Strings args = {"--serve", "--write"};
+        if (remoteStore.get() != "") {
+            args.push_back("--store");
+            args.push_back(remoteStore.get());
+        }
+        conn->sshConn = master.startCommand(remoteProgram, args);
         conn->to = FdSink(conn->sshConn->in.get());
         conn->from = FdSource(conn->sshConn->out.get());
 
--- a/src/libstore/ssh.cc
+++ b/src/libstore/ssh.cc
@@ -37,7 +37,7 @@ void SSHMaster::addCommonSSHOpts(Strings & args)
         args.push_back("-C");
 }
 
-std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command, Strings args)
+std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command, Strings & args)
 {
     Path socketPath = startMaster();
 
--- a/src/libstore/ssh.cc
+++ b/src/libstore/ssh.cc
@@ -37,7 +37,7 @@ void SSHMaster::addCommonSSHOpts(Strings & args)
         args.push_back("-C");
 }
 
-std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command)
+std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command, Strings args)
 {
     Path socketPath = startMaster();
 
@@ -62,21 +62,58 @@ std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string
         if (logFD != -1 && dup2(logFD, STDERR_FILENO) == -1)
             throw SysError("duping over stderr");
 
-        Strings args;
+        Strings execargs;
 
         if (fakeSSH) {
-            args = { "bash", "-c" };
+            execargs.push_back(command);
+            execargs.splice(execargs.end(), args);
         } else {
-            args = { "ssh", host.c_str(), "-x" };
-            addCommonSSHOpts(args);
+            std::string escapedCommand = shellEscape(command);
+
+            std::string script = {escapedCommand};
+
+            for (std::string arg: args) {
+                script.push_back(' ');
+                script.append(shellEscape(arg));
+            }
+
+            // This is really ugly, but it's less ugly then trying to catch the
+            // error thrown by this process when ssh it can't find the command.
+            // We would have to first of all catch this error (which is not
+            // trivial, since it will only be thrown to the parent process
+            // during blocking communication), then redirect the stderr, search
+            // it for "command not found" and then ssh again and try to find the
+            // command.
+            script.append(
+                fmt(
+                    " || ( \
+                    code=$?;\
+                    if ! [ $code -eq 127 ]; then exit $code; fi;\
+                    PATH=\"$PATH:$HOME/.nix-profile/bin:/nix/var/nix/profiles/default/bin\";\
+                    cmd=\"$(command -v %s)\";\
+                    if [ -n \"$cmd\" ]; then\
+                        printf \"Executable %s not found in PATH on the remote host, but it was found in %%s.\\nPerhaps try adding '\\e[35;1m&remote-program=%%s\\e[0m' at the end of your SSH url.\n\" \"$cmd\" \"$cmd\" > /dev/stderr; \
+                    else\
+                        printf \"Executable %s was not found on the remote host. Are you sure you have Nix installed there?\n\" > /dev/stderr;\
+                    fi;\
+                    exit $code;\
+                    )",
+                    escapedCommand,
+                    escapedCommand,
+                    escapedCommand
+                )
+            );
+
+            execargs = { "ssh", host.c_str(), "-x" };
+            addCommonSSHOpts(execargs);
             if (socketPath != "")
-                args.insert(args.end(), {"-S", socketPath});
+                execargs.insert(args.end(), {"-S", socketPath});
             if (verbosity >= lvlChatty)
-                args.push_back("-v");
+                execargs.push_back("-v");
+            execargs.push_back(script);
         }
 
-        args.push_back(command);
-        execvp(args.begin()->c_str(), stringsToCharPtrs(args).data());
+        execvp(execargs.begin()->c_str(), stringsToCharPtrs(execargs).data());
 
         // could not exec ssh/bash
         throw SysError("unable to execute '%s'", args.front());
--- a/src/libstore/ssh-store.cc
+++ b/src/libstore/ssh-store.cc
@@ -88,9 +88,12 @@ private:
 ref<RemoteStore::Connection> SSHStore::openConnection()
 {
     auto conn = make_ref<Connection>();
-    conn->sshConn = master.startCommand(
-        fmt("%s --stdio", remoteProgram)
-        + (remoteStore.get() == "" ? "" : " --store " + shellEscape(remoteStore.get())));
+    Strings args = {"--stdio"};
+    if (remoteStore.get() != "") {
+        args.push_back("--store");
+        args.push_back(remoteStore.get());
+    }
+    conn->sshConn = master.startCommand(remoteProgram, args);
     conn->to = FdSink(conn->sshConn->in.get());
     conn->from = FdSource(conn->sshConn->out.get());
     return conn;
--- a/src/libstore/ssh.hh
+++ b/src/libstore/ssh.hh
@@ -38,7 +38,7 @@ public:
         AutoCloseFD out, in;
     };
 
-    std::unique_ptr<Connection> startCommand(const std::string & command);
+    std::unique_ptr<Connection> startCommand(const std::string & command, Strings args);
 
     Path startMaster();
 };


edolstra:
Hm, I don't think it's a good idea to make assumptions about the remote machine like this, i.e. appending a bunch of paths to `$PATH` in the hope that one of them will work.

balsoft:
But currently we're just expecting that `PATH` contains the `nix-store` binary, which is often not true; This PR really does make `nix copy` and `nix-copy-closure` work in many situations where they didn't before, e.g. when copying to/from a macOS machine. I agree that the exact solution is not pretty and there are still edge cases and non-standard installations which it won't support, but it does help a lot of the time.

thufschmitt:
I agree that having it broken out-of-the-box like that is very annoying. And although the current solution is dirty, it has the merit of fixing it.

Alternatively, there's a `remote-program` store parameter that can specify explicitly the path to the `nix-{store,daemon}` program to run remotely. Maybe we can just improve the error message to point to it?

src/libstore/legacy-ssh-store.cc
```diff
@@ -74,7 +74,7 @@ struct LegacySSHStore : public virtual LegacySSHStoreConfig, public virtual Stor
     {
         auto conn = make_ref<Connection>();
         conn->sshConn = master.startCommand(
-            fmt("%s --serve --write", remoteProgram)
+            fmt("PATH=\"$PATH:$HOME/.nix-profile/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin\" %s --serve --write", remoteProgram)
```
SuperSandro2000, line 77: I don't think we need /run/current system here. That would only work if the remote machine is nixos and that has nix* in PATH.

bjornfor: Ping @balsoft ?

balsoft: Sure

thufschmitt (approved):
Yay!

@edolstra do you object at me merging this?

edolstra:
Maybe it's better to do something like `ssh ... sh --login exec nix-store --serve`? I don't think it's ideal to make assumptions about the location of Nix on the remote system. It's the job of the installer to make sure that `nix` can be found in a default shell environment. (Unfortunately that's currently not the case, requiring [hacks like here in the installer test](https://github.com/NixOS/nix/blob/0d4bf9c4d836f8e8570ad1d39245a5835ef4aaf1/tests/installer/default.nix#L197-L201).)

balsoft:
> Maybe it's better to do something like ssh ... sh --login exec nix-store --serve?

As far as I remember from my experimentation, `--login` doesn't help. Unless `sh` is started in interactive mode (with TTY attached and such), it will not run the RC files, and thus Nix won't be available. This is not great, but the proposed solution really does fix it and circumvents many other potential issues with the environment, without "hardcoding" anything (after all, it just adds to the search path, and does not subtract or override anything from it).

edolstra:
How about `sh --login -i`?

balsoft:
Doesn't seem to help.

Ericson2314:
@balsoft Hi, I attend the Nix team meeting where this was discussed.

There were a few things that came up, but a basic concern was that this is a workaround that is hard to remove once it is added, because people will start relying on it, and ideally we would want to change the installer so that this problem doesn't happen at all going forward. (I'll also add that PRs like yours to use XDG dirs would also means we need to change this yet again.)

What was decided might be good to do instead was something like this psuedocode:
1. First try sshing as today.
2. If that fails because the nix binary was not found, try sshing again with `PATH` set or other workarounds
3. Give the user an error message that instead of just saying what the failure is, tell them how they can set the `remote-program` query parameter to make it work.

The idea is that since we are just changing an error message, it is very easy to back out of later if we've fixed the underlying issues (installer etc.). Since manual effort (setting `remote-program`) is needed, and since that works either way, we can be sure that no one is accidentally relying on this will therefore be left "high and dry" if we remove it later.

balsoft:
Ok, I see. It would still be a bit annoying for users, but I'm guessing much less so. I'll try to rework it in this same PR.

Ericson2314:
Yes, manual effort is no fan, but manual fixes are a lot less painful than manual troubleshooting.

balsoft:
@Ericson2314 Rewritten as requested. It's really ugly, but seems to work.

src/libstore/legacy-ssh-store.cc
```diff
@@ -73,9 +73,12 @@ struct LegacySSHStore : public virtual LegacySSHStoreConfig, public virtual Stor
...
         auto conn = make_ref<Connection>();
-        conn->sshConn = master.startCommand(
-            fmt("%s --serve --write", remoteProgram)
-            + (remoteStore.get() == "" ? "" : " --store " + shellEscape(remoteStore.get())));
+        Strings args = {"--serve", "--write"};
+        if (remoteStore.get() != "") {
+            args.push_back("--store");
+            args.push_back(remoteStore.get());
+        }
+        conn->sshConn = master.startCommand(remoteProgram, args);
```
thufschmitt, line 81: That's a behaviour change I think because so far you can set `remote-program` to an arbitrary bash snippet (in particular `nix-store --some --arguments`). I don't mind this too much because not allowing that is much cleaner, but it would need to be documented.

src/libstore/ssh.cc
```diff
@@ -62,21 +62,58 @@ std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string
...
+                    else\
+                        printf \"Executable %s was not found on the remote host. Are you sure you have Nix installed there?\n\" > /dev/stderr;\
+                    fi;\
+                    exit $code;\
+                    )",
+                    escapedCommand,
+                    escapedCommand,
+                    escapedCommand
+                )
+            );
```
thufschmitt, line 105: I think it would be much cleaner (though a bit slower) to have that logic on the client side

thufschmitt (changes requested):

src/libstore/legacy-ssh-store.cc
balsoft: Ah, right. Well...

src/libstore/ssh.cc
balsoft: Which client side, though? As far as I understand, at this stage we haven't found the "client" yet.

balsoft: Ah, never mind, confusing client and server here. Do you want me to do what I described in the comment? Because I think it would be a lot more hacky due to having to catch an exception from another process :(

thufschmitt: Duh, I didn't read the comment :man_facepalming: 

Yeah, that would be hacky indeed, but afaik `ssh` forwards the return code, so checking for a return code `127` should work well-enough I think. But maybe that would still make it somewhat ugly :man_shrugging:

balsoft: Hm, that sounds good. Where would we actually `catch` the error though? This is a separate process from the parent one. There are a couple of options:

1. Wait some time to let ssh connect and fail if there's no such command, only then continue with the communcation. This is _not great_, obviously.
2. Write (in a blocking way) _something_ to the pipe, `catch` there. It's not clear what _something_ should be, it sounds hacky, and I'm not sure the exception thrown (probably "unexpected EOF") will actually have the exit code of ssh.

thufschmitt: Mh, I see. I'll defer this to @Ericson2314 as the owner of that PR :stuck_out_tongue:

Ericson2314: Yeah I would advocate for a hybrid approach where we do the `command -v %s` thing, and print out sentinel values `NIX YES` or `NIX NO` or whatever, but the rest of the logic is in C++. Does that make sense?

balsoft: You mean do the `command -v` dance before we run the command, then capture the output? Otherwise I can't see how it would work, given the limitations outlined in the previous comment.

Ericson2314: I mean something like this:

1. ssh normally
2. if failled, warn something like "${remoteProgram} on remote end not found, troubleshooting..."
3. begin trouble shooting loop over various additional path entries to try
   1. ssh again to run `PATH=$PATH:foo command -v; echo "COMMAND RESULT $!"` or similar, for some `foo`.
   2. parse the result in C++
   3. If something was found, print instructions to use for how to fix, and break out of loop.
 4. else If nothing was found, print something like "troubleshooting failed" and the original error.

balsoft: > 2. if failled, warn something like "${remoteProgram} on remote end not found, troubleshooting..."

It's not clear to me how we would catch this, and then decide that it failed because of a missing program and not some other problem.

> 3. if failled, warn something like "${remoteProgram} on remote end not found, troubleshooting..."

TBH this sounds like it would be even more hacky.

I don't think I want to spend the time to implement this, since I feel like it would be an ugly hackfest regardless.

Ericson2314: > It's not clear to me how we would catch this, and then decide that it failed because of a missing program and not some other problem.

That is why we are doing `echo "COMMAND RESULT $!"`, so we can grep for a specific sentinel value and not get confused about other errors.

> TBH this sounds like it would be even more hacky.

There is no fancy condition for that warning. that is simply to let the user know we are going to start poking at the remote some more. If we *do* hit another error and don't get `COMMAND RESULT <number>` back, it allows us to say something like "troubleshooting failed" indicating the issue is something else than what we are aware of.

Ericson2314: The point of all this is to keep the *happy* path very clean and simple, no messing with the `PATH` or anything else, but the *sad* path is free to try random stuff in a non-normative way to assist the user with debugging the issue.

balsoft: > The point of all this is to keep the happy path very clean and simple, no messing with the PATH or anything else, but the sad path is free to try random stuff in a non-normative way to assist the user with debugging the issue.


I think the happy path (before catching the error with `||`) is exactly as before with my solution as well. This ugly mess only happens when running the usual command fails.

Ericson2314: I don't mind there being a second ssh. The user can always control-C out if they don't want to do it; the "we're going to try trouble shooting now" message lets them know what's going on.

src/libstore/ssh.cc
```diff
@@ -37,7 +37,7 @@ void SSHMaster::addCommonSSHOpts(Strings & args)
         args.push_back("-C");
 }
 
-std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command)
+std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command, Strings args)
```
Ericson2314, line 40:
```suggestion
std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string & command, Strings & args)
```

Ericson2314:
@balsoft so my latest thought is that if the remote side unconditionally sends a message down one of the pipes, then the local side can unconditionally read that message. It can do that "synchronously", before yielding to the caller which uses the file descriptors at it sees fit.

This would be a pattern somewhat comparable to wrapping a packet in another sort of packet. E.g. how an IP packet can wrap a UDP packet, and code can "peel off and interpret" one packet at a time.

How does this sound?

balsoft:
I have been thinking about something like that. However, then we're violating the idea that "the happy path" should be devoid of hacks (since this sounds a lot like a hack to me). It's also more work then the current solution, and doesn't feel like it's a major improvement.

Ericson2314:
> However, then we're violating the idea that "the happy path" should be devoid of hacks (since this sounds a lot like a hack to me).

So the specific sort of hack I am most worried about is the "Postel's law creep", were we work harder and harder to make things work in ways that cannot be done done later without a breaking change. While it is unfortunate that this is work done in the happy path, it does *not* have that risk, in that we can add or remove the extra check at any time. (it is all client side code, and it doesn't effect whether the ssh will succeed since the program needs to exist to be run.)

> It's also more work then the current solution, and doesn't feel like it's a major improvement.

I do think it is better to through a C++ exception than rely on stderr being ferried along. The synchronous nature of this associates the message with the failure, which better for e.g. structured logging on the client side.

Ericson2314:
@balsoft If it helps, I would be open to landing some of your code quality improvements (like using a vector not space-splitting for arguments) first. Those are good in any event I'd think, and then it can de-clutter the rest.

Ericson2314:
@balsoft would you like to push what you've got?

balsoft:
No, not really. It didn't even work properly in the "happy path".

I think I'll just separate out the improvements you do agree on, and then leave the rest for now (there are changes that are more interesting to me personally, so I can't find the motivation to do a proper rewrite here).

Ericson2314:
To be clear, I would be happy to work on it myself from your point of attempting what I suggested (giving myself a taste of my own medicine :) ) but sure, we can start with the stuff that we both agree on.

thufschmitt:
> That let me to learn there is also a [RemoteCommand](https://man.openbsd.org/ssh_config#RemoteCommand) that I think is just what we want for what I was requesting with this!

I suspect that will break everyone using a restricted ssh access that only allows running `nix-daemon` or `nix-store` on the server

tomberek:

PR #6634, opened 2022-06-08
fix(libstore/lock): support users that belong to more than 10 groups

lovesegfault:
The manpage for `getgrouplist` says:

> If the number of groups of which user is a member is less than or
> equal to *ngroups, then the value *ngroups is returned.
>
> If the user is a member of more than *ngroups groups, then
> getgrouplist() returns -1.  In this case, the value returned in
> *ngroups can be used to resize the buffer passed to a further
> call getgrouplist().

In our original code, however, we allocated a list of size `10` and, if
`getgrouplist` returned `-1` threw an exception. In practice, this
caused the code to fail for any user belonging to more than 10 groups.

While unusual for single-user systems, large companies commonly have a
huge number of POSIX groups users belong to, causing this issue to crop
up and make multi-user Nix unusable in such settings.

The fix is relatively simple, when `getgrouplist` fails, it stores the
real number of GIDs in `ngroups`, so we must resize our list and retry.
Only then, if it errors once more, we can raise an exception.

This should be backported to, at least, 2.9.x.

**Release Notes**

**Testing**

### DIFF
--- a/src/libstore/lock.cc
+++ b/src/libstore/lock.cc
@@ -67,13 +67,26 @@ bool UserLock::findFreeUser() {
 #if __linux__
             /* Get the list of supplementary groups of this build user.  This
                is usually either empty or contains a group such as "kvm".  */
-            supplementaryGIDs.resize(10);
-            int ngroups = supplementaryGIDs.size();
-            int err = getgrouplist(pw->pw_name, pw->pw_gid,
-                supplementaryGIDs.data(), &ngroups);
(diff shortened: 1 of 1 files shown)


lovesegfault:
I debated a bit on whether to duplicate the call to `getgrouplist` or use a loop, but I figured since it's only called exactly twice that the duplication was better.

Happy to replace with a loop and a break/continue if others disagree.

PR #6637, opened 2022-06-09
nix-env: A small std::move() optimization

sidkshatriya:
(Found another location)

Avoids doing a O(n) copy of Strings i.e. `std::list<std::string>`


thufschmitt (approved):
Thanks :)

PR #6639, opened 2022-06-09
install-multi-user: check if selinux is enabled and if it is then abort

Artturin:
Prevents messed up installs and manual cleanup https://github.com/NixOS/nix/issues/2374

Not tested yet

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -638,6 +638,17 @@ place_channel_configuration() {
     fi
 }
 
+check_selinux() {
+    if command -v getenforce > /dev/null 2>&1; then
+        if ! [ "$(getenforce)" = "Disabled" ]; then
+            failure <<EOF
+Nix does not work with selinux enabled yet!
+see https://github.com/NixOS/nix/issues/2374
+EOF
+        fi
+    fi
+}
+
 welcome_to_nix() {
     ok "Welcome to the Multi-User Nix Installation"
 
@@ -865,6 +876,8 @@ Please do not run this script with root privileges. I will call sudo
 when I need to.
 EOF
     fi
+
+    check_selinux
 
     if [ "$(uname -s)" = "Darwin" ]; then
         # shellcheck source=./install-darwin-multi-user.sh


abathur:
If you create a cachix cache using the naming convention in https://github.com/NixOS/nix/pull/4577 and add a secret with the token, the CI run in your fork can generate and test the installer.

yawaramin (approved):

Artturin:
> If you create a cachix cache using the naming convention in #4577 and add a secret with the token, the CI run in your fork can generate and test the installer (well--confirm it doesn't cause trouble on ubuntu or macOS).

Thanks! could someone add it to https://github.com/NixOS/nix/blob/master/doc/manual/src/contributing/hacking.md so more people would know about it?

Artturin:
the installer tests succeeded https://github.com/Artturin/nix/actions/runs/2472299209

abathur:
If anyone around is set up to confirm (from a Fedora VM?), you can use the generated installer URL (and stable tarball-url-prefix) from the test: https://github.com/Artturin/nix/runs/6824433318?check_suite_focus=true#step:4:3

You just invoke the install like:
```
sh <(curl -L https://artturin-nix-install-tests.cachix.org/serve/wnnlsx7rygx1z5pj7lq65fhlrslr9vlg/install) --tarball-url-prefix https://artturin-nix-install-tests.cachix.org/serve
```

Artturin:
image

stelcodes (commented):
I think technically someone *could* create proper Nix SELinux rules before installation. It's not that Nix isn't compatible with SELinux, it's just that most users will not have the correct rules in place before trying to install. I think there should be a prompt that says something like "SELinux is enabled on your machine. The installation process is likely to fail due to permission errors since Nix does not include SELinux rules. Abort? (y/n)"

yawaramin:
I think the error message is fine as is. It's concise, to the point, and links to the relevant issue.

thomasjcf29:
I've raised #6881 as the solution provided in #6639 prevents install on a system with SELinux set to permissive. For all intents and purposes disabled and permissive should be treated the same (as far as nix is concerned).

I've gone into further detail on the PR.

PR #6645, opened 2022-06-10
systemd,launchd match nixos open files limit

Artturin:
it was bumped in https://github.com/NixOS/nixpkgs/pull/176558

### DIFF
--- a/misc/launchd/org.nixos.nix-daemon.plist.in
+++ b/misc/launchd/org.nixos.nix-daemon.plist.in
@@ -28,7 +28,7 @@
     <key>SoftResourceLimits</key>
     <dict>
       <key>NumberOfFiles</key>
-      <integer>4096</integer>
+      <integer>1048576</integer>
     </dict>
   </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


edolstra:
1 million file descriptors sounds rather excessive. If Nix really needs that many, it's clearly a bug.

roberth:
> 1 million file descriptors sounds rather excessive. If Nix really needs that many, it's clearly a bug.

Nix only needs like 3.
The question is what does a _user_'s derivation need? A million is only excessive until it is not, but _probably_ not enough to cause issues on the system.

> clearly a bug

Or a highly concurrent test suite that's heavy on fds.

lovesegfault:
@edolstra Can you take another look at this?

I can't build many derivations on macOS without manually patching the plist like this; otherwise it complains about too many open files.

mstone:
I also recently ran into (another) derivation that needed a higher limit than was provided by default [ IIRC, testing out a mass-rebuild-causing change that had been merged into staging ] and so would appreciate a fix along these lines ~regardless of what number is ultimately chosen.

Artturin:
Thanks thufschmitt & edolstra

Issue #6650, opened 2022-06-11
sourcehut flake input stopped working


wahjava:
**Describe the bug**

Input of type `sourcehut` git repository stopped working lately, while equivalent `git+https` works fine.

**Steps To Reproduce**

```
❯ mkdir /tmp/f
❯ cd /tmp/f
❯ cat >/tmp/f/flake.nix <<EOF
{
  inputs = { ff-overlay = { url = "sourcehut:~abbe/ff-overlay/mainline"; }; };
  outputs = { inputs, ... }: { };
}
EOF
❯ nix flake lock --recreate-lock-file --show-trace
error: in 'sourcehut:~abbe/ff-overlay/mainline', couldn't find ref 'mainline'

       … while fetching the input 'sourcehut:~abbe/ff-overlay/mainline'

       … while updating the flake input 'ff-overlay'

       … while updating the lock file of flake 'path:/tmp/f?lastModified=1654934958&narHash=sha256-d5my7Tnc39FLi6C1X1dqJrPEq9KshYoXm6p2QyY8OqM='

❯ curl -s https://git.sr.ht/\~abbe/ff-overlay/info/refs
7e6763635ec34443d04a640f3a0f41d45161914e	refs/heads/mainline

❯ git clone https://git.sr.ht/\~abbe/ff-overlay
Cloning into 'ff-overlay'...
remote: Enumerating objects: 482, done.
remote: Counting objects: 100% (18/18), done.
remote: Compressing objects: 100% (18/18), done.
remote: Total 482 (delta 10), reused 0 (delta 0), pack-reused 464
Receiving objects: 100% (482/482), 184.70 KiB | 327.00 KiB/s, done.
Resolving deltas: 100% (306/306), done.
❯ git -C ff-overlay rev-parse HEAD
7e6763635ec34443d04a640f3a0f41d45161914e
❯ git -C ff-overlay branch |more
* mainline
```

**Expected behavior**

Flake lock file should have been updated

**`nix-env --version` output**

```
❯ nix-env --version
nix-env (Nix) 2.9.0
```

**Additional context**

thufschmitt:
@misterio77 any idea what might be happening here?

wahjava:
`HEAD` and commitid works, but not the branch name.

Misterio77:
Hey folks!

[...]

PR #6653, opened 2022-06-11
fix sourcehut brach/tag resolving regression

Misterio77:
nixos/nix#6290 introduced a regex pattern to account for tags when resolving sourcehut refs.

nixos/nix#4638 (more specifically https://github.com/NixOS/nix/commit/9bf296c970bf33b7ed53d7e2d8fbe44197482518) reafactored the code, accidentally treating the pattern as a regular string, causing all non-HEAD ref resolving to break.

This fixes the regression and adds more test cases to avoid future breakage.

Fixes https://github.com/NixOS/nix/issues/6650.

The issue affects nix 2.9.0 and 2.9.1. Can we maybe backport the fix?

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -381,7 +381,7 @@ struct SourceHutInputScheme : GitArchiveInputScheme
 
         Headers headers = makeHeadersWithAuthTokens(host);
 
-        std::string ref_uri;
+        std::string refUri;
         if (ref == "HEAD") {
             auto file = store->toRealPath(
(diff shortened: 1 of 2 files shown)


thufschmitt (approved):
Thanks, great catch :)

thufschmitt:
> The issue affects nix 2.9.0 and 2.9.1. Can we maybe backport the fix?

I've merged the backport to the `2.9-maintenance` branch, should be picked-up by the next patch release

Misterio77:
Thank you, @thufschmitt!

PR #6663, opened 2022-06-13
flakes: throw an error if `follows`-declaration for an input is invalid

Ma27:
I recently got fairly confused why the following expression didn't have
any effect

    {
      description = "Foobar";
      inputs.sops-nix = {
        url = github:mic92/sops-nix;
        inputs.nixpkgs_22_05.follows = "nixpkgs";
      };
    }

until I found out that the input was called `nixpkgs-22_05` (please note
the dash vs. underscore).

IMHO it's not a good idea to not throw an error in that case and
probably leave end-users rather confused, so I implemented a small check
for that which basically checks whether `follows`-declaration from
overrides actually have corresponding inputs in the transitive flake.

In fact this was done by accident already in our own test-suite where
the removal of a `follows` was apparently forgotten[1].

Since the key of the `std::map` that holds the `overrides` is a vector
and we have to find the last element of each vector (i.e. the override)
this has to be done with a for loop in O(n) complexity with `n` being
the total amount of overrides (which shouldn't be that large though).

Please note that this doesn't work with nested expressions, i.e.

    inputs.fenix.inputs.nixpkgs.follows = "...";

which is a known problem[2].

For the expression demonstrated above, an error like this will be
thrown:

    error: sops-nix has a `follows'-declaration for a non-existant input nixpkgs_22_05!

[1] 2664a216e57169ec57d7f51be1b8383c1be83fd5
[2] https://github.com/NixOS/nix/issues/5790

cc @edolstra @thufschmitt @Ericson2314

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -353,26 +353,15 @@ LockedFlake lockFlake(
         std::vector<FlakeRef> parents;
 
         std::function<void(
-            const FlakeInputs & flakeInputs,
-            std::shared_ptr<Node> node,
             const InputPath & inputPathPrefix,
-            std::shared_ptr<const Node> oldNode,
-            const InputPath & lockRootPath,
-            const Path & parentPath,
-            bool trustLock)>
-            computeLocks;
+            const FlakeInputs & flakeInputs
+            )>
+            checkFollowsDeclarations;
 
-        computeLocks = [&](
-            const FlakeInputs & flakeInputs,
-            std::shared_ptr<Node> node,
+        checkFollowsDeclarations = [&](
             const InputPath & inputPathPrefix,
-            std::shared_ptr<const Node> oldNode,
-            const InputPath & lockRootPath,
-            const Path & parentPath,
-            bool trustLock)
-        {
-            debug("computing lock file node '%s'", printInputPath(inputPathPrefix));
-
+            const FlakeInputs & flakeInputs
+        ) {
             for (auto [inputPath, inputOverride] : overrides) {
                 auto inputPath2(inputPath);
                 auto follow = inputPath2.back();
@@ -394,6 +383,30 @@ LockedFlake lockFlake(
                     );
                 }
             }
+        };
+
+        std::function<void(
+            const FlakeInputs & flakeInputs,
+            std::shared_ptr<Node> node,
+            const InputPath & inputPathPrefix,
+            std::shared_ptr<const Node> oldNode,
+            const InputPath & lockRootPath,
+            const Path & parentPath,
+            bool trustLock)>
+            computeLocks;
+
+        computeLocks = [&](
+            const FlakeInputs & flakeInputs,
+            std::shared_ptr<Node> node,
+            const InputPath & inputPathPrefix,
+            std::shared_ptr<const Node> oldNode,
+            const InputPath & lockRootPath,
+            const Path & parentPath,
+            bool trustLock)
+        {
+            debug("computing lock file node '%s'", printInputPath(inputPathPrefix));
+
+            checkFollowsDeclarations(inputPathPrefix, flakeInputs);
 
             /* Get the overrides (i.e. attributes of the form
                'inputs.nixops.inputs.nixpkgs.url = ...'). */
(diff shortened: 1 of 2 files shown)


Ma27:
ping @edolstra @thufschmitt would you mind taking a look? :)

src/libexpr/flake/flake.cc
```diff
@@ -373,6 +373,29 @@ LockedFlake lockFlake(
...
+                    && flakeInputs.find(follow) == flakeInputs.end()
+                ) {
+                    std::string root;
+                    for (auto & element : inputPath2) {
+                        root.append(element);
+                        if (element != inputPath2.back()) {
+                            root.append(".inputs.");
+                        }
+                    }
+                    throw Error(
```
thufschmitt, line 391: I think a warning would be better. There's probably some somewhat legitimate scenarios where you want to override something that doesn't exist yet (say because you're generating some overrides on a set of flakes or because you're using `--override-input` to point an input to a new copy which has less inputs itself) so I wouldn't want it to just fail

src/libexpr/flake/flake.cc
```diff
@@ -373,6 +373,29 @@ LockedFlake lockFlake(
         {
             debug("computing lock file node '%s'", printInputPath(inputPathPrefix));
 
+            auto overrides2 = overrides;
+            for (auto & [inputPath, inputOverride] : overrides2) {
+                auto inputPath2(inputPath);
```
thufschmitt, line 378: It feels like you're doing some useless copies and move+copy. Could this be replaced by:

```suggestion
            for (auto [inputPath, inputOverride] : overrides) {
```

(and replacing `inputPath2` by `inputPath` below)

?

src/libexpr/flake/flake.cc
```diff
@@ -373,6 +373,29 @@ LockedFlake lockFlake(
...
+                        }
+                    }
+                    throw Error(
+                        "%s has a `follows'-declaration for a non-existant input %s!",
+                        root,
+                        follow
+                    );
+                }
+            }
+
```
thufschmitt, line 398: Can you move this whole logic into its own auxiliary function?

tests/flakes.sh
```diff
@@ -845,3 +846,21 @@ nix store delete $(nix store add-path $badFlakeDir)
...
+        inputs.invalid.follows = "D";
+    };
+    inputs.D.url = "path:./flakeD";
+    outputs = { ... }: {};
+}
+EOF
+
+git -C $flakeFollowsA add flake.nix
+
+nix flake lock $flakeFollowsA 2>&1 | grep "error: B has a \`follows'-declaration for a non-existant input invalid!"
```
thufschmitt, line 866: If it's not too much work, do you think you could move that out of the `flakes` test (along with the other `follows` tests)? This big file gets really hard to understand and slow to run

thufschmitt (changes requested):
Thanks :)

A few comments, but looks good overall

tests/flakes.sh
Kha: FWIW I moved all of the `flakeFollows*` test code into a separate `flake-follows.sh` file for #6621 but couldn't finish the changes because of a new bug, presumably from rebasing. I can push the incomplete state later today.

src/libexpr/flake/flake.cc
Ma27: I guess so, not too experienced with C++, so I'll just try it out :D

src/libexpr/flake/flake.cc
Ma27: Fine by me. I'm happy as long as there's an indication that something might be wrong :)

tests/flakes.sh
Ma27: Since @kha seems to have taken care of this already, I'd suggest to not do it here and to move it into `flakes-follows.sh` as soon as #6621 is ready if both of you are OK with that :)

thufschmitt: > FWIW I moved all of the flakeFollows* test code into a separate flake-follows.sh file for https://github.com/NixOS/nix/pull/6621
 
Uh [that's why](https://github.com/NixOS/nix/pull/6621#discussion_r913569695) I had this feeling of _dejà-vu_ when writing this comment. Sorry for the duplicate then, yes let's do the split in only one place, will be better :)

thufschmitt (approved):

PR #6673, opened 2022-06-15
libstore: improve warning message on missing sig

asymmetric:
Clarifies that the substitute will be ignored/skipped.


thufschmitt (approved):
Makes sense, thanks

Issue #4884, opened 2021-06-03
nix repl :e stdout not redirected to terminal


zanculmarktum:
This happened when I upgraded to `21.05`. It has the exact same `configuration.nix` from the previous version (`20.09`).

```
$ EDITOR=vim nix repl '<nixpkgs>'
:e hello
Vim: Warning: Output is not to a terminal
```

```
$ EDITOR="emacsclient -nw" nix repl '<nixpkgs>'
:e hello
/run/current-system/sw/bin/emacsclient: could not get terminal name
error: program 'emacsclient' failed with exit code 1
```

The behavior that we just saw is the exact same behavior when you redirect the stdout of an editor to somewhere else, i.e.

```
$ vim >/dev/null
Vim: Warning: Output is not to a terminal
```

```
$ emacsclient -nw >/dev/null 
/run/current-system/sw/bin/emacsclient: could not get terminal name
```

The normal behavior (the one from `20.09`) is that it fires up an editor instead of showing those error messages.

I use `nixpkgs.nixUnstable` (with flakes enabled) by the way.

zanculmarktum:
To be exact the version that works was from [e0ca98c2071b815578470e280df8fdb750c7e23b](https://github.com/NixOS/nix/tree/e0ca98c2071b815578470e280df8fdb750c7e23b).

virusdave:
I believe this is the problem: https://github.com/NixOS/nix/pull/6674

tomberek:
Closing as stale. Re-open if needed.

PR #6674, opened 2022-06-15
Don't capture `stdout` when launching subshells in `nix repl`

virusdave:
In [this commit](https://github.com/NixOS/nix/commit/9b9e703df41d75949272059f9b8bc8b763e91fce), the behavior of the `:s` and `:u` commands in `nix repl` changed by capturing `stdout` of the forked process.  This doesn't seem to be an intentional change, since the captured output is discarded.

This change reverts the behavior by having `nix repl`'s invocations of `nix` no longer redirect and capture `stdout` of the forked process.

### DIFF
--- a/src/libcmd/repl.cc
+++ b/src/libcmd/repl.cc
@@ -111,23 +111,20 @@ NixRepl::~NixRepl()
     write_history(historyFile.c_str());
 }
 
-std::string runNix(Path program, const Strings & args,
+void runNix(Path program, const Strings & args,
     const std::optional<std::string> & input = {})
 {
     auto subprocessEnv = getEnv();
     subprocessEnv["NIX_CONFIG"] = globalConfig.toKeyValue();
 
-    auto res = runProgram(RunOptions {
+    runProgram2(RunOptions {
         .program = settings.nixBinDir+ "/" + program,
         .args = args,
         .environment = subprocessEnv,
         .input = input,
     });
 
-    if (!statusOk(res.first))
-        throw ExecError(res.first, "program '%1%' %2%", program, statusToString(res.first));
-
-    return res.second;
+    return;
 }
 
 static NixRepl * curRepl; // ugly


virusdave:
I think this fixes https://github.com/NixOS/nix/issues/4884 .

```
$ EDITOR=vim nix repl '<nixpkgs>'
:e hello
# vim opens with the `hello` derivation correctly
```

virusdave:
@thufschmitt since no good deed goes unpunished, can I get your thoughts on this one as well?

src/libutil/util.cc
```diff
@@ -1093,6 +1093,14 @@ std::pair<int, std::string> runProgram(RunOptions && options)
     StringSink sink;
     options.standardOut = &sink;
 
+    int status = runProgramWithoutCapturingStdout(options);
+
+    return {status, std::move(sink.s)};
+}
+
+// Output = error code
+int runProgramWithoutCapturingStdout(const RunOptions & options)
```
virusdave, line 1102: I'm not a big fan of this name, and would rather have the "no capture" function be just plain `runProgram` and name the capturing version something else, but that's a more invasive change, so I'll defer to the maintainers/reviewers here on what they prefer.

src/libcmd/repl.cc
```diff
@@ -111,23 +111,23 @@ NixRepl::~NixRepl()
...
 
-std::string runNix(Path program, const Strings & args,
+void runNix(Path program, const Strings & args,
     const std::optional<std::string> & input = {})
 {
     auto subprocessEnv = getEnv();
     subprocessEnv["NIX_CONFIG"] = globalConfig.toKeyValue();
 
-    auto res = runProgram(RunOptions {
+    int res = runProgramWithoutCapturingStdout(RunOptions {
```
thufschmitt, line 120: Apart from the very untelling name, I think you could directly use `runProgram2` here, right?

virusdave:
By duplicating the "capture exception and grab edit code" logic that
doesn't exist in `runProgram2`? Sure, if that's preferred, let me know.

thufschmitt:
> By duplicating the "capture exception and grab edit code" logic that doesn't exist in `runProgram2`?

I might be missing something, but I'm under the impression that the exit code is only used to re-throw a similar exception a few lines later. Which means that the whole logic is "capture exception and grab edit code to immediately rebuild and throw the same (roughly) exception". Which sounds like a complex way of expressing the identity function :wink:

virusdave:
> > By duplicating the "capture exception and grab edit code" logic that doesn't exist in `runProgram2`?
> 
> I might be missing something, but I'm under the impression that the exit code is only used to re-throw a similar exception a few lines later. Which means that the whole logic is "capture exception and grab edit code to immediately rebuild and throw the same (roughly) exception". Which sounds like a complex way of expressing the identity function 😉

Ha, you are indeed correct.  I completely glossed over this, apparently.  I'll simplify this change to do what you're suggesting.  Thanks!

virusdave:
> > > By duplicating the "capture exception and grab edit code" logic that doesn't exist in `runProgram2`?
> > 
> > 
> > I might be missing something, but I'm under the impression that the exit code is only used to re-throw a similar exception a few lines later. Which means that the whole logic is "capture exception and grab edit code to immediately rebuild and throw the same (roughly) exception". Which sounds like a complex way of expressing the identity function 😉
> 
> Ha, you are indeed correct. I completely glossed over this, apparently. I'll simplify this change to do what you're suggesting. Thanks!

Done!

With latest changes:
```
dave @ davembp [/tmp]
$ nix-shell -p 'with pkgs; nix.overrideAttrs (self: super: {patches = fetchpatch {url="https://github.com/virusdave/nix/commit/32978e9c00b7287a6fef624aec8fa854d8062878.patch"; sha256="sha256-KuFThyYAOES/17lyzJLv55tMdeC43Oq60AVvcKj4g2g=";};})'

dave @ davembp [nix-shell:/tmp]
$ nix repl
Welcome to Nix 2.8.1. Type :? for help.

nix-repl> :l <nixpkgs>
Added 16467 variables.

nix-repl> :u hello     
dave @ davembp [nix-shell:/tmp]

$ echo 'Output still works!'
Output still works!

dave @ davembp [nix-shell:/tmp]
$ 
exit

```

thufschmitt (approved):
Simple and concise, like that :)

Thanks!

PR #6678, opened 2022-06-16
doc: update darwin uninstall instructions

abathur:
Fixing a few issues that might be misleading people:
- update the fstab example to match new installs
- update the volume removal step to account for an unmounted volume
- avoid suggesting that the installer can fix an old-style setup

### DIFF
--- a/doc/manual/src/installation/installing-binary.md
+++ b/doc/manual/src/installation/installing-binary.md
@@ -149,7 +149,7 @@ and `/etc/zshrc` which you may remove.
 
 4. Edit fstab using `sudo vifs` to remove the line mounting the Nix Store
    volume on `/nix`, which looks like
-   `UUID=<unique uuid> /nix apfs rw,noauto,nobrowse,suid,owners` or
+   `UUID=<uuid> /nix apfs rw,noauto,nobrowse,suid,owners` or
    `LABEL=Nix\040Store /nix apfs rw,nobrowse`. This will prevent automatic
    mounting of the Nix Store volume.
 
--- a/doc/manual/src/installation/installing-binary.md
+++ b/doc/manual/src/installation/installing-binary.md
@@ -148,7 +148,8 @@ and `/etc/zshrc` which you may remove.
    This will remove all the build users that no longer serve a purpose.
 
 4. Edit fstab using `sudo vifs` to remove the line mounting the Nix Store
-   volume on `/nix`, which looks like this,
+   volume on `/nix`, which looks like
+   `UUID=<unique uuid> /nix apfs rw,noauto,nobrowse,suid,owners` or
    `LABEL=Nix\040Store /nix apfs rw,nobrowse`. This will prevent automatic
    mounting of the Nix Store volume.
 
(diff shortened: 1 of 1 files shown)


doc/manual/src/installation/installing-binary.md
```diff
@@ -191,8 +202,7 @@ and `/etc/zshrc` which you may remove.
 <!-- Note: anchors above to catch permalinks to old explanations -->
 
 We believe we have ironed out how to cleanly support the read-only root
-on modern macOS. New installs will do this automatically, and you can
-also re-run a new installer to convert your existing setup.
+on modern macOS. New installs will do this automatically.
```
abathur, line 205: To clarify, the old statement was true for a while (and is technically true of the volume-setup parts of the macOS install), but installs can now break with a populated store since `rsync` was swapped out for `cp` in #5150. (I opened https://github.com/NixOS/nix/issues/6679 so that we have a chance to avoid losing the knowledge that it once worked this way.)

We've been getting a steady trickle of these reports since ~December (https://github.com/NixOS/nix/search?o=desc&q=%22cp%3A+cannot+overwrite+directory%22&s=created&type=issues). Not that I think fixing the language will actually help (I imagine most of these people are just _reasoning_ that a reinstall can/would/should fix whatever ails them).

doc/manual/src/installation/installing-binary.md
```diff
@@ -175,6 +176,16 @@ and `/etc/zshrc` which you may remove.
...
    store.
 
+   If this command indicates that it couldn't remove the volume, your Nix Store
+   volume may not be mounted. Run the following to double-check:
+
+   ```console
+   diskutil list
+   ```
+
+   If you see a Nix Store volume, re-run the diskutil deleteVolume command, but
```
toonn, line 186: This explains what to do if the volume is indeed mounted but not what to do if it's not. Probably as simple as mounting it but maybe there's some help we can offer here on how to figure out how to find the right volume to mount?

toonn (approved):
LGTM, one potential addition.

doc/manual/src/installation/installing-binary.md
toonn: Ah, I was the one who misunderstood. The "to double-check" made it sound to me like "it's probably mounted but the label may differ, use diskutil list to figure out the volume and used the bsd identifier."

Maybe it can say something along the lines of, "If there's no Nix Store volume, you can skip this step. Otherwise use the identifier to delete it?"

abathur: I've force-pushed some phrasing tweaks. WDYT?

toonn: Perfect!

doc/manual/src/installation/installing-binary.md
```diff
@@ -148,7 +148,8 @@ and `/etc/zshrc` which you may remove.
    This will remove all the build users that no longer serve a purpose.
 
 4. Edit fstab using `sudo vifs` to remove the line mounting the Nix Store
-   volume on `/nix`, which looks like this,
+   volume on `/nix`, which looks like
+   `UUID=<unique uuid> /nix apfs rw,noauto,nobrowse,suid,owners` or
```
fricklerhandwerk, line 152:
```suggestion
   `UUID=<uuid> /nix apfs rw,noauto,nobrowse,suid,owners` or
```

UUIDs are already unique.

doc/manual/src/installation/installing-binary.md
```diff
@@ -175,6 +176,18 @@ and `/etc/zshrc` which you may remove.
    This will remove the Nix Store volume and everything that was added to the
    store.
 
+   If this command indicates that it couldn't remove the volume, you should
```
fricklerhandwerk, line 179:
```suggestion
   If the output indicates that the command couldn't remove the volume, you should
```

domenkozar:
Can we merge this?

fricklerhandwerk (approved):

PR #6694, opened 2022-06-20
filetransfer: set Accept-Encoding to all values supported by libcurl for transparent compression

NickCao:
ref: https://discourse.nixos.org/t/announcing-harmonia-a-nix-binary-cache-written-in-rust/19855
This is especially helpful when fetching large nars from binary caches that does not support compression, or having difficulties implementing it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -190,6 +190,16 @@ PKG_CHECK_MODULES([SQLITE3], [sqlite3 >= 3.6.19], [CXXFLAGS="$SQLITE3_CFLAGS $CX
 # Look for libcurl, a required dependency.
 PKG_CHECK_MODULES([LIBCURL], [libcurl], [CXXFLAGS="$LIBCURL_CFLAGS $CXXFLAGS"])
 
+# Check whether libcurl supports required features
+PKG_CHECK_VAR([LIBCURL_SUPPORTED_FEATURES], [libcurl], [supported_features], [
+  m4_foreach([LIBCURL_FEATURE], [brotli, zstd], [
+    AC_MSG_CHECKING([whether libcurl supports feature LIBCURL_FEATURE])
+    AS_CASE($LIBCURL_SUPPORTED_FEATURES, [*LIBCURL_FEATURE*], [AC_MSG_RESULT(yes)],
+      [AC_MSG_RESULT(no)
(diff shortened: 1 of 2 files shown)


src/libstore/filetransfer.cc
```diff
@@ -160,7 +160,7 @@ struct curlFileTransfer : public FileTransfer
                 result.bodySize += realSize;
 
                 if (!decompressionSink) {
-                    decompressionSink = makeDecompressionSink(encoding, finalSink);
+                    decompressionSink = makeDecompressionSink("", finalSink);
```
edolstra, line 163: A "none" decompression sink is superfluous and can be removed.

IIRC the reason we're doing our own decompression here is that curl didn't have brotli support or wasn't compiled with it. We may want to check in our configure script that curl has support for brotli and zstd.

NickCao: Why should we prefer brotli, an algo optimized for html to compress nars that's mostly binary files? And properly implemented servers should not use encodings not explicitly advertised by the client to send responses, lacking support for a compression algo or two is not that critical from my point of view.

configure.ac
```diff
@@ -190,6 +190,16 @@ PKG_CHECK_MODULES([SQLITE3], [sqlite3 >= 3.6.19], [CXXFLAGS="$SQLITE3_CFLAGS $CX
...
+# Check whether libcurl supports required features
+PKG_CHECK_VAR([LIBCURL_SUPPORTED_FEATURES], [libcurl], [supported_features], [
+  m4_foreach([LIBCURL_FEATURE], [brotli, zstd], [
+    AC_MSG_CHECKING([whether libcurl supports feature LIBCURL_FEATURE])
+    AS_CASE($LIBCURL_SUPPORTED_FEATURES, [*LIBCURL_FEATURE*], [AC_MSG_RESULT(yes)],
+      [AC_MSG_RESULT(no)
+       AC_MSG_ERROR([Nix requires libcurl to support feature LIBCURL_FEATURE])])
+  ])
+], [AC_MSG_ERROR([Nix requires libcurl to provide a list of supported features])])
+
```
NickCao, line 202: I'm not an autoconf expert, feel free to point out better ways to do this.

PR #6698, opened 2022-06-21
Forbid the tilde expansion in pure eval mode

### LINKED ISSUE #6684: Possible to impurely read $HOME in flakes by using `~`-paths
**Describe the bug**

The nix language supports using paths that start with `~/`, which are relative to the user's home directory. It looks like this is implemented by using the $HOME environment variable:

```console
$ HOME=/tmp nix eval --expr '~/foo'
/tmp/foo
```

However, this functionality isn't disabled when evaluating a flake, meaning that flakes can produce different results depending on what $HOME is set to at evaluation time.

**Steps To Reproduce**

flake.nix:

```
{
  description = "A very basic flake";
  outputs = { self, nixpkgs }: {
    devShells.x86_64-linux.default = nixpkgs.legacyPackages.x86_64-linux.mkShell {
      shellHook = ''
        echo my path is: ${toString ./.}
        echo your home directory is: ${toString ~/.}
      '';
    };
  };
}
```

If I run `nix develop` as two different users, the flake prints different home directories:

```console
$ nix develop
my path is: /nix/store/c6c50m39qyfb94iix44352a8pj3yz22d-source
your home directory is: /home/ash
```

```console
$ sudo nix develop
[sudo] password for ash:
my path is: /nix/store/c6c50m39qyfb94iix44352a8pj3yz22d-source
your home directory is: /root
```

**Expected behavior**

Since `nix eval --expr 'builtins.getEnv "HOME"'` produces the empty string unless `--impure` is used, I would expect `~`-paths to be disallowed in flakes.

```
nix-env (Nix) 2.8.1
```

**Additional context**

I had a brief look through public flakes on github that might be affected by this (with the query `path:flake.nix " ~/"` in github code search), and I could only find one, kmein/niveum:

https://github.com/kmein/niveum/blob/db7cb8c74c5157b48c8ca9f95f4885f811284357/flake.nix#L96

This is a very similar example to what led me to realise this fact in the first place – I was running `sudo nixos-rebuild` and was confused why something using a `~/`-path wasn't working as I expected, until I realised that probably it was evaluating to `/root/...` rather than `/home/ash/...`.

thufschmitt:
Fix #6684 by throwing an error at parse time when encountering a path that starts with a `~`

### DIFF
--- a/src/libexpr/parser.y
+++ b/src/libexpr/parser.y
@@ -520,6 +520,12 @@ path_start
     $$ = new ExprPath(path);
   }
   | HPATH {
+    if (evalSettings.pureEval) {
+        throw Error(
+            "the path '%s' can not be resolved in pure mode",
+            std::string_view($1.p, $1.l)
+        );
+    }
     Path path(getHome() + std::string($1.p + 1, $1.l - 1));
     $$ = new ExprPath(path);
   }
--- a/tests/pure-eval.sh
+++ b/tests/pure-eval.sh
@@ -30,3 +30,5 @@ nix eval --store dummy:// --write-to $TEST_ROOT/eval-out --expr '{ x = "foo" + "
 
 rm -rf $TEST_ROOT/eval-out
 (! nix eval --store dummy:// --write-to $TEST_ROOT/eval-out --expr '{ "." = "bla"; }')
+
+(! nix eval --expr '~/foo')


thufschmitt:
I'm not sure I get your comment. The point of the pure evaluation mode is to make sure that the same expression evaluated in two different contexts will yield the same result. Right now `~/.` is a blatant exception to that rule, so forbidding it seems to be the logical choice

sersorrel:
I'm also not sure I understand. As far as I can see, this change is limited to enforcing that, during flake evaluation, you should not be able to do something differently depending on the location of the current user's home directory – instead of writing `~/something`, you should have to write e.g. `/home/ash/something`, because `~/something` could mean `/home/ash/something` or `/home/robert/something` or `/root/something` depending on which user evaluates it.

It's not "flakes should/shouldn't be able to read git config" or something (as far as I understand, it's not possible to invoke an external program like git during flake evaluation anyway), it's "when using flakes, I should get the same result whether I run `nixos-rebuild --use-remote-sudo` or `sudo nixos-rebuild`".

PR #6706, opened 2022-06-22
nar-info-disk-cache: refresh nix-cache-info weekly

lheckemann:
This allows changes to nix-cache-info to be picked up by existing
clients. Previously, the only way for this to happen would be for
clients to delete binary-cache-v6.sqlite, which is quite awkward for
users.

On the other hand, updates to nix-cache-info should be pretty rare,
hence the choice of a fairly long TTL. Configurability is probably not
useful enough to warrant implementing it.

### DIFF
--- a/src/libstore/nar-info-disk-cache.cc
+++ b/src/libstore/nar-info-disk-cache.cc
@@ -62,6 +62,9 @@ public:
     /* How often to purge expired entries from the cache. */
     const int purgeInterval = 24 * 3600;
 
+    /* How long to cache binary cache info (i.e. /nix-cache-info) */
+    const int cacheInfoTtl = 7 * 24 * 3600;
+
     struct Cache
     {
         int id;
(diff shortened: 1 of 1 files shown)


src/libstore/nar-info-disk-cache.cc
```diff
@@ -186,8 +189,11 @@ class NarInfoDiskCacheImpl : public NarInfoDiskCache
                 auto queryCache(state->queryCache.use()(uri));
                 if (!queryCache.next())
                     return std::nullopt;
+                if (queryCache.getInt(1) + cacheInfoTtl < time(0))
+                    return std::nullopt;
```
thufschmitt, line 193: Just for the sake of consistency, can you implement this check like it's done for the others (with the condition right in the SQL query)?

thufschmitt (changes requested):
Thanks,

A small suggestion, but looks good overall

thufschmitt (approved):
Thanks :)

(btw, I don't get a notification when you just push something new, so don't hesitate to re-request a review or send a comment when you do. I only noticed that you updated the PR accidentally because I clicked on the wrong link :) )

PR #6708, opened 2022-06-22
Enable tests for nix-static

edolstra:
The tests weren't running because `pkgsStatic` is apparently considered a cross environment, so `checkPhase` and `installCheckPhase` were disabled even though we enable them explicitly.

Also fix a few tests that were broken with musl.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -571,13 +571,18 @@
           nativeBuildInputs = nativeBuildDeps;
           buildInputs = buildDeps ++ propagatedDeps;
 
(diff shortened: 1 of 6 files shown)


Artturin:
fyi this was fixed in https://github.com/NixOS/nixpkgs/commit/41485e7337baca768dc542c553a9d66030df7b33

edolstra:
@Artturin Thanks, I've added a reminder to remove this in the future.

PR #6709, opened 2022-06-22
Fix build-remote in nix-static

edolstra:
`build-remote` is now executed via `/proc/self/exe` so it works out of the box when you only have the static binary.


src/nix/main.cc
```diff
@@ -266,6 +266,11 @@ void mainWrapped(int argc, char * * argv)
     programPath = argv[0];
     auto programName = std::string(baseNameOf(programPath));
 
+    if (argc > 0 && std::string_view(argv[0]) == "__build-remote") {
```
thufschmitt, line 269: Does that really work? It looks like the default `argv0` for the build hook is `readlink(/proc/self/exe) + "__build-remote"`, so something like `/usr/bin/nix__build-remote`, which means that this will always fail

PR #6710, opened 2022-06-22
Embed the sandbox shell into the statically linked 'nix' binary

edolstra:
With this, Nix will write a copy of the sandbox shell to `/bin/sh` in the sandbox rather than bind-mounting it from the host filesystem. This makes /bin/sh work out of the box with nix-static, i.e. you no longer get

```
/nix/store/qa36xhc5gpf42l3z1a8m1lysi40l9p7s-bootstrap-stage4-stdenv-linux/setup: ./configure: /bin/sh: bad interpreter: No such file or directory
```

This increases the size of the `nix` binary by 213 KiB.


src/libstore/local.mk
```diff
@@ -44,9 +44,19 @@ libstore_CXXFLAGS += \
  -DNIX_MAN_DIR=\"$(mandir)\" \
  -DLSOF=\"$(lsof)\"
 
+ifeq ($(embedded_sandbox_shell),yes)
+libstore_CXXFLAGS += -DSANDBOX_SHELL=\"__embedded_sandbox_shell__\"
+
+$(d)/build/local-derivation-goal.cc: $(d)/embedded-sandbox-shell.gen.hh
+
+$(d)/embedded-sandbox-shell.gen.hh: $(sandbox_shell)
+	$(trace-gen) hexdump -v -e '1/1 "0x%x," "\n"' < $< > $@.tmp
```
thufschmitt, line 53: I'm not sure it's available in stdenv, but if it is, `xxd` has a `--include` option to directly produce a C-compatible header

edolstra: Unfortunately, `xxd` is part of `vim` which is a pretty big dependency to pull in...

PR #6714, opened 2022-06-23
Automatically use a chroot store if /nix doesn't exist

edolstra:
Specifically, if we're not root and /nix and the daemon socket do not exist, then we use `~/.local/share/nix/root` as a chroot store. This enables non-root users to download nix-static and have it work out of the box, e.g.

```
ubuntu@ip-10-13-1-146:~$ ~/nix run nixpkgs#hello
warning: '/nix' does not exist, so Nix will use '/home/ubuntu/.local/share/nix/root' as a chroot store
Hello, world!
```

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -1326,9 +1326,9 @@ std::shared_ptr<Store> openFromNonUri(const std::string & uri, const Store::Para
                store in ~/.local/share/nix/root. */
             auto chrootStore = getDataDir() + "/nix/root";
             if (!pathExists(chrootStore))
-                warn("'/nix' does not exists, so Nix will use '%s' as a chroot store", chrootStore);
+                warn("'/nix' does not exist, so Nix will use '%s' as a chroot store", chrootStore);
             else
-                debug("'/nix' does not exists, so Nix will use '%s' as a chroot store", chrootStore);
+                debug("'/nix' does not exist, so Nix will use '%s' as a chroot store", chrootStore);
             Store::Params params2;
             params2["root"] = chrootStore;
             return std::make_shared<LocalStore>(params2);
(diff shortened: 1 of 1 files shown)


src/libstore/store-api.cc
```diff
@@ -1319,7 +1320,19 @@ std::shared_ptr<Store> openFromNonUri(const std::string & uri, const Store::Para
...
-        else
+        else if (!pathExists(stateDir) && params.empty() && getuid() != 0) {
+            /* If /nix doesn't exist, there is no daemon socket, and
+               we're not root, then automatically set up a chroot
+               store in ~/.local/share/nix/root. */
+            auto chrootStore = getDataDir() + "/nix/root";
+            if (!pathExists(chrootStore))
+                warn("'/nix' does not exists, so Nix will use '%s' as a chroot store", chrootStore);
+            else
+                debug("'/nix' does not exists, so Nix will use '%s' as a chroot store", chrootStore);
```
cole-h, line 1331: Nit:

```suggestion
                warn("'/nix' does not exist, so Nix will use '%s' as a chroot store", chrootStore);
            else
                debug("'/nix' does not exist, so Nix will use '%s' as a chroot store", chrootStore);
```

edolstra:
TODO: only do this on Linux.

src/libstore/store-api.cc
```diff
@@ -1319,6 +1320,21 @@ std::shared_ptr<Store> openFromNonUri(const std::string & uri, const Store::Para
             return std::make_shared<LocalStore>(params);
         else if (pathExists(settings.nixDaemonSocketFile))
             return std::make_shared<UDSRemoteStore>(params);
+        #if __linux__
+        else if (!pathExists(stateDir) && params.empty() && getuid() != 0) {
```
thufschmitt, line 1324: Could this instead try to create the state dir and only fallback to a chroot store if it can't? That would avoid issues like https://github.com/NixOS/nix/issues/6732

PR #6716, opened 2022-06-23
eval-cache: cast rowId to correct type

Mindavi:
Prevents (one of the) errors when running with UBSan:

/nix/store/j5vhrywqmz1ixwhsmmjjxa85fpwryzh0-gcc-11.3.0/include/c++/11.3.0/bits/stl_pair.h:353:4: runtime error: load of value 229, which is not a valid value for type 'AttrType'


edolstra:
Thanks!

PR #6718, opened 2022-06-24
nix develop: parse multi-line vars in get-env.sh

impl:
The `get-env.sh` script parses the output of `declare -p` line-by-line, but some shell variables can span multiple lines. In this case, if one of the variables is script-like (for example, `shellHook`), it can potentially pollute the derivation environment or fail outright. A trivial reproduction of the issue looks like:

```nix
{
  # ...
  shellHook = ''
    declare -x VAR=FOO
    declare -x "TEST_$VAR"=BAR
  '';
}
```

To avoid this problem, we teach `get-env.sh` to identify and ignore the content of multi-line string values when parsing variable declarations.

### DIFF
--- a/src/nix/get-env.sh
+++ b/src/nix/get-env.sh
@@ -38,9 +38,18 @@ __dumpEnv() {
     printf '  "variables": {\n'
     local __first=1
     while read __line; do
-        if ! [[ $__line =~ ^declare\ (-[^ ])\ ([^=]*) ]]; then continue; fi
+        if ! [[ $__line =~ ^declare\ (-[^ ])\ ([^=]*)(=(.*))?$ ]]; then continue; fi
         local type="${BASH_REMATCH[1]}"
         local __var_name="${BASH_REMATCH[2]}"
+        local __value="${BASH_REMATCH[4]}"
+
+        # Eat any string values (not arrays) that might span over multiple lines.
+        if [[ $__value == \"* ]]; then
+            __value="${__value:1}"
+            while ! [[ $__value =~ (^|[^\\])(\\\\)*\"$ ]]; do
+                IFS= read -r __value
+            done
+        fi
 
         if [[ $__var_name =~ ^BASH_ || \
               $__var_name = _ || \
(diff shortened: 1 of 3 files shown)


edolstra:
Hm, not sure I understand the problem description. `nix develop` already support multi-line strings, e.g. in Nix's own flake:
```
$ nix develop
$ echo "$shellHook"
PATH=$prefix/bin:$PATH
unset PYTHONPATH
export MANPATH=$out/share/man:$MANPATH

# Make bash completion work.
XDG_DATA_DIRS+=:$out/share
```

impl:
Hi Eelco,

Thanks for taking a look! Let me see if I can provide a bit more detail for you.

You can carefully craft a value for `shellHook` (or, in my case, incidentally craft, and discover this bug) that pollutes the variables found by the `get-env.sh` script. The problem is that the script parses each line as if it's at the top-level scope, without regard for quoting. For example, if I set:

```nix
{
  shellHook = ''
    echo foo
    declare -x I_SHOULD_NOT_BE_HERE=NOPE
  '';
}
```

Then we read, line-by-line,

```
declare -x shellHook="
```
```
echo foo
```
```
declare -x I_SHOULD_NOT_BE_HERE=NOPE
```
```
"
```

Normally, it isn't a problem because the [regex on line 41](https://github.com/NixOS/nix/blob/6cb41288ac8df00328865f4c66c324452e5a961c/src/nix/get-env.sh#L41) will skip over this out-of-scope multi-line content (this happens to the lines `echo foo` and `"` of our example), but if a line happens to also contain `declare ...`, then it gets treated as if it were part of the overall derivation environment! To validate:

```console
$ nix print-dev-env --json | jq '.variables | has("I_SHOULD_NOT_BE_HERE")'
true
```

The fix I've proposed just makes the script skip over the content of multi-line variables by parsing them as a string unto themselves, rather than allowing the outer loop to read their content line-by-line. Does that help clarify a bit?

impl:
Hi @edolstra,

Anything else I can do to help give clarity on this change? Happy to hop on Matrix or IRC to explain if it'd be more helpful to do it synchronously too. Let me know!

Thanks!

PR #6720, opened 2022-06-24
flake: add comment explaining when workaround for static compilation can be dropped

cole-h:
So we don't leave cruft like this lying around forever. Thanks @Artturin for pointing out that this was fixed, and providing a link to the commit that fixed it: https://github.com/NixOS/nix/pull/6708#issuecomment-1165912951.


cole-h:
Resolved in https://github.com/NixOS/nix/commit/749d914d10e7550fb26a3a1599e2d4d7ed59b00a.

PR #6721, opened 2022-06-24
Automatically format with clang-tools

grahamc:
Sort of a shitpost, but most times I read through Nix I wish it was consistently formatted.

### DIFF
--- a/src/libcmd/command.cc
+++ b/src/libcmd/command.cc
@@ -7,7 +7,7 @@
 
 #include <nlohmann/json.hpp>
 
-extern char * * environ __attribute__((weak));
+extern char ** environ __attribute__((weak));
 
 namespace nix {
 
@@ -20,7 +20,8 @@ nix::Commands RegisterCommand::getCommandsFor(const std::vector<std::string> & p
         if (name.size() == prefix.size() + 1) {
             bool equal = true;
             for (size_t i = 0; i < prefix.size(); ++i)
-                if (name[i] != prefix[i]) equal = false;
+                if (name[i] != prefix[i])
+                    equal = false;
             if (equal)
                 res.insert_or_assign(name[prefix.size()], command);
         }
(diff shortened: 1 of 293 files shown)


lovesegfault:
I was dying for something like this when I tried to write some code for Nix a few days ago. Hope it happens!

src/libcmd/command.cc
```diff
@@ -211,60 +207,58 @@ Strings editorFor(const Path & file, uint32_t line)
...
 {
-    addFlag({
-        .longName = "profile",
-        .description = "The profile to update.",
-        .labels = {"path"},
-        .handler = {&profile},
-        .completer = completePath
-    });
+    addFlag(
+        {.longName = "profile",
```
edolstra, line 221: The change to the formatting of designated initializers causes a huge diff, but I don't see a way to keep the old style...

src/nix/log.cc
```diff
@@ -17,11 +17,14 @@ struct CmdLog : InstallableCommand
     std::string doc() override
     {
         return
-          #include "log.md"
-          ;
+#include "log.md"
```
edolstra, line 20: This is ugly, it should follow the indentation.

thufschmitt (changes requested):
The diff is obviously scary, but overall I think it's a great thing!

One thing that would be needed imho to make this really useful is a way to enforce this in the long run as it will quickly diverge from that style again. Maybe a CI job checking the format of everything?

Also, if the expectation is to enforce this, we should make it clear and easy for contributors by updating the hacking docs accordingly and having a nice `make` target for it (I know there's already a script for it, but I don't think it's as discoverable and nice to use)

src/libcmd/command.cc
fricklerhandwerk: We could use [`.git-blame-ignore-revs`](https://github.com/community/community/discussions/5033#discussioncomment-2318478) as [we do in `nixpkgs`](https://github.com/NixOS/nixpkgs/blob/master/.git-blame-ignore-revs).

edolstra:
TODO: add a CI check to verify formatting.

Issue #6728, opened 2022-06-26
fetchTree is not documented


symphorien:
the nix manual only mentions fetchTree in release notes.

elikoga:
Looking into this: `nix __dump-builtins` doesn't do the right thing apparently

elikoga:
Found the fixme: https://github.com/NixOS/nix/blob/6cb41288ac8df00328865f4c66c324452e5a961c/src/libexpr/primops/fetchTree.cc#L185

elikoga:
Seems like this is a dupe of #5463

PR #6740, opened 2022-06-29
add fetchTree docs

elikoga:
Relates to #6728 and removes the fixme in https://github.com/NixOS/nix/blob/6cb41288ac8df00328865f4c66c324452e5a961c/src/libexpr/primops/fetchTree.cc#L185

Maybe someone can add a few words about what `fetchTree` actually does so we can expand the text

Otherwise I'll rework the text over the next few days

### DIFF
--- a/src/libexpr/primops/fetchTree.cc
+++ b/src/libexpr/primops/fetchTree.cc
@@ -182,8 +182,26 @@ static void prim_fetchTree(EvalState & state, const PosIdx pos, Value * * args, 
     fetchTree(state, pos, args, v, std::nullopt, FetchTreeParams { .allowNameArgument = false });
 }
 
-// FIXME: document
-static RegisterPrimOp primop_fetchTree("fetchTree", 1, prim_fetchTree);
+static RegisterPrimOp primop_fetchTree({
+    .name = "fetchTree",
+    .args = {"args"},
+    .arity = 1,
+    .doc = R"(
+      Introduced in 2.4
+
+      The fetcher infrastructure is exposed via the `fetchTree` built-in.
+
+      `builtins.fetchTree` allows fetching a source tree using any
+      backends supported by the fetcher infrastructure. It subsumes the
+      functionality of existing built-ins like `fetchGit`,
+      `fetchMercurial` and `fetchTarball`.
+
+      `builtins.fetchTree` can be used to fetch
+      plain files over the `http(s)` and `file` protocols in addition to
+      directory tarballs.
+    )",
+    .fun = prim_fetchTree,
+});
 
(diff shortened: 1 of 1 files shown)


elikoga:
Ouch, just saw https://github.com/NixOS/nix/pull/5556

Can rework this to that state if needed/requested

src/libexpr/primops/fetchTree.cc
```diff
@@ -182,8 +182,26 @@ static void prim_fetchTree(EvalState & state, const PosIdx pos, Value * * args,
...
-static RegisterPrimOp primop_fetchTree("fetchTree", 1, prim_fetchTree);
+static RegisterPrimOp primop_fetchTree({
+    .name = "fetchTree",
+    .args = {"args"},
+    .arity = 1,
+    .doc = R"(
+      Introduced in 2.4
+
+      The fetcher infrastructure is exposed via the `fetchTree` built-in.
+
```
edolstra, line 193:
```suggestion
```
No need to mention the "fetcher infrastructure", since that's an implementation detail and users don't know what that is.

src/libexpr/primops/fetchTree.cc
```diff
@@ -182,8 +182,26 @@ static void prim_fetchTree(EvalState & state, const PosIdx pos, Value * * args,
...
+      `builtins.fetchTree` allows fetching a source tree using any
+      backends supported by the fetcher infrastructure. It subsumes the
```
edolstra, line 195: Shouldn't mention "fetcher infrastructure". Instead it can refer to the docs on flake references (https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html#flake-references), specifically the attrset form.

src/libexpr/primops/fetchTree.cc
```diff
@@ -182,8 +182,26 @@ static void prim_fetchTree(EvalState & state, const PosIdx pos, Value * * args,
...
+      functionality of existing built-ins like `fetchGit`,
+      `fetchMercurial` and `fetchTarball`.
+
+      `builtins.fetchTree` can be used to fetch
+      plain files over the `http(s)` and `file` protocols in addition to
+      directory tarballs.
+    )",
```
edolstra, line 202: There should be one or two examples here of how to use `fetchTree`.

It should also mention that `fetchTree` is experimental and that it requires a locked flakeref in pure eval mode (which is the default).

edolstra (changes requested):

fricklerhandwerk:
Closing in favor of #8509. Sorry for the wasted time, everyone. I hope we'll continuously get get better at reviewing in a timely manner now that we have a maintainers team.

Issue #6250, opened 2022-03-13
build failure without boehm GC


nwns:
Trying to nix for Yocto and in an attempt at expedience, skipping all optional dependencies.

Build fails
```
| src/libexpr/eval.cc: In member function 'nix::Value* nix::EvalState::allocValue()':
| src/libexpr/eval.cc:861:28: error: 'GC_malloc_many' was not declared in this scope
|   861 |         *valueAllocCache = GC_malloc_many(sizeof(Value));
|       |                            ^~~~~~~~~~~~~~
| src/libexpr/eval.cc:868:47: error: 'GC_NEXT' was not declared in this scope
|   868 |     GC_PTR_STORE_AND_DIRTY(&*valueAllocCache, GC_NEXT(p));
|       |                                               ^~~~~~~
| src/libexpr/eval.cc:868:5: error: 'GC_PTR_STORE_AND_DIRTY' was not declared in this scope
|   868 |     GC_PTR_STORE_AND_DIRTY(&*valueAllocCache, GC_NEXT(p));
|       |     ^~~~~~~~~~~~~~~~~~~~~~
```

**Steps To Reproduce**

Presumably simply `./configure --disable-gc`

**Expected behavior**

Optional dependency to be actually be optional

Nix 2.6.1 with bellow patches on Yocto Honister:
```
Build Configuration:
BB_VERSION           = "1.52.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "universal"
TARGET_SYS           = "x86_64-poky-linux"
MACHINE              = "intel-skylake-64"
DISTRO               = "poky"
DISTRO_VERSION       = "3.4.1"
TUNE_FEATURES        = "m64 skylake"
TARGET_FPU           = ""
```
**Additional context**

Applied the following patches to source to reach this point

```diff
From dedd6c12cbce3081ca91866f9c41c378264ae55c Mon Sep 17 00:00:00 2001
From: Nigel Whillier <nigel.w@nosun.ca>
Date: Sun, 13 Mar 2022 14:26:12 -0600
Subject: [PATCH] No documentation

---
 Makefile | 1 -
 1 file changed, 1 deletion(-)

diff --git a/Makefile b/Makefile
index e6ce50c..88b0d27 100644
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,6 @@ makefiles = \
   misc/systemd/local.mk \
   misc/launchd/local.mk \
   misc/upstart/local.mk \
[...]

PR #6741, opened 2022-06-29
support building with --enable-gc=no

Mindavi:
Some minor changes fixing the build without boehm.
Fixes https://github.com/NixOS/nix/issues/6250.

Would be nice to test this (building without boehm) somehow (in hydra maybe), but haven't looked at that yet and IMO that can be added separately. Could help prevent or reduce regressions on this configuration.

I'm interested in this to possibly make ASAN builds more reliable. Not sure if this will really help, but since I was looking at it anyway and the fixes are low-hanging fruit, I've submitted it :smile: .

Not 100% sure about the AllocCache members, but in the cc file there's a path for both boehm enabled and disabled where the members are needed, so I just assumed that that was correct and this corresponds to that.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -45,7 +45,7 @@ static char * allocString(size_t size)
 #if HAVE_BOEHMGC
     t = (char *) GC_MALLOC_ATOMIC(size);
 #else
-    t = malloc(size);
+    t = (char *) malloc(size);
 #endif
     if (!t) throw std::bad_alloc();
     return t;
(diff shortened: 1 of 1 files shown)


Mindavi:
Hmm, not sure if this makes sense. The valueAllocCache only seems to be used when HAVE_BOEHMGC is defined. So the expression in the cc file should be fixed rather than this in the header :). Will revisit later (feel free to pick it up if I forget about it).

thufschmitt:
> The valueAllocCache only seems to be used when HAVE_BOEHMGC is defined. So the expression in the cc file should be fixed rather than this in the header :)

Yes, these alloc caches are only used to batch the allocations with the GC because allocating a lot of small values is fairly slow. That being said, it probably doesn't hurt to keep the exact same logic when not using the GC to keep things simple.

> Would be nice to test this (building without boehm) somehow (in hydra maybe)

Absolutely. A hydra job for that sounds like the best solution.

thufschmitt (approved):
Thanks, let's merge

PR #6749, opened 2022-07-01
Allow specification of extra packages, maxLayers in Docker image

centromere:
This change allows for deeper customization of the resulting Docker image. Specifically, one can set `maxLayers` as well as `extraPkgs` to include other packages.


garbas (approved):
This docker.nix will hopefully be moved to nixpkgs one day :)

But until then we can add this kind of stuff to it.

Issue #6753, opened 2022-07-02
A let-in expr is not accepted as flake outputs


lincolnauster:
**Describe the bug**

When writing a flake, outputs usually takes the form `{ self, and, deps }: ...`. I ran into a situation in which I wanted to write:

```nix
{
  # ...
  outputs = let helper = foo; in { self, and, deps}: makeAFlake;
}
```

Running this, however, gives you an error:

```
Error: expected a function but got a thunk at ...
```

**Steps To Reproduce**

Below is a pretty small demonstration. There's no reason to write this code exactly, but it demonstrates the issue.

```nix
# flake.nix
{
  inputs.nipkgs.url = "github:nixos/nixpkgs";

  outputs = let myHello = pkgs: pkgs.hello; in { self, nixpkgs }: {
    defaultPackage.x86_64-linux = myHello (import nixpkgs { system = "x86_64-linux"; });
  };
}
```

**Expected behavior**

A successful build.

**`nix-env --version` output**

2.9.1

---

If you'd like, I can probably patch this myself. I'm mostly wondering whether this is an intended feature/limitation or not.

cole-h:
You can get around this by doing

```nix
# flake.nix
{
  inputs.nipkgs.url = "github:nixos/nixpkgs";

  outputs = { self, nixpkgs }:
    let myHello = pkgs: pkgs.hello; in 
  {
    defaultPackage.x86_64-linux = myHello (import nixpkgs { system = "x86_64-linux"; });
  };
}
```

lincolnauster:
Wow, thanks for the quick response :)

I am getting around it like that, but it still feels sort of 'incorrect' that the prior (afaict equivalent) form doesn't work. Again, I can patch this whenever. (Looking at the source, it would probably be a pretty small patch, too). Is it worth reviewing?

tomberek:
Closing as stale. Re-open if needed.

Issue #3075, opened 2019-08-30
Require that flake attributes are not thunks


edolstra:
For example, `edition = 1 + 2` should not be valid but `edition = 3` should be. This prevents having to evaluate arbitrarily complex (possibly non-terminating) code to get flake metadata.

PR #6758, opened 2022-07-05
Make Value::isTrivial 'if-let aware'

oldaccountdeadname:
As it stands, isTrivial will not 'descend' through let expressions and treat an expression, e.g.,`let x = 1; in x`, as a non-trivial value. While this makes a bit of conceptual sense, it's got some unwelcome side effects. (See, for instance, #6753.)

Here's the commit message from the relevant change:

```
Value::isTrivial is used in a variety of cases, and mostly
descends throughout thunks. I.e., isTrivial returns true if a
value is a partially evaluated thunk that returns a 'trivially
trivial' value. This patch allows isTrivial to descend into let
expressions, so that if let expressions return a trivial value,
they're treated as such.
```

Hopefully this is applicable! Thanks!

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1,5 +1,6 @@
 #include "eval.hh"
 #include "hash.hh"
+#include "nixexpr.hh"
 #include "types.hh"
 #include "util.hh"
 #include "store-api.hh"
@@ -258,14 +259,23 @@ PosIdx Value::determinePos(const PosIdx pos) const
 
 bool Value::isTrivial() const
 {
-    return
-        internalType != tApp
-        && internalType != tPrimOpApp
-        && (internalType != tThunk
-            || (dynamic_cast<ExprAttrs *>(thunk.expr)
-                && ((ExprAttrs *) thunk.expr)->dynamicAttrs.empty())
-            || dynamic_cast<ExprLambda *>(thunk.expr)
-            || dynamic_cast<ExprList *>(thunk.expr));
+    if (internalType == tApp || internalType == tPrimOp) {
+        return false;
+    } else if (internalType == tThunk) {
+        auto expr = thunk.expr;
+        for (;;) {
+            if ((dynamic_cast<ExprAttrs *>(expr)
+                    && ((ExprAttrs *) expr)->dynamicAttrs.empty())
+                || dynamic_cast<ExprLambda *>(expr)
+                || dynamic_cast<ExprList *>(expr))
+                return true;
+            else if (dynamic_cast<ExprLet *>(expr)) {
+                expr = dynamic_cast<ExprLet *>(expr)->body;
+            } else {
+                return false;
+            }
+        }
+    } else return false;
 }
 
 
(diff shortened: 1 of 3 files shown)


edolstra:
In `let x = x; in x`, the let-body `x` is not a trivial value, so it should be fine.

lincolnauster:
> We can nearly do something like inputs.nixpkgs = let version = "22.05"; in { url = "nixpkgs/nixos-${version}"; }, but not really).

Oh, good point - there are a lot of scenarios that are disallowed here that I hadn't run into and therefore just wasn't thinking about.

I tracked down the original motivation of isTrivial to #3075, where the goal was to prevent reading metadata from hanging on infinite loops in flakes? (Correct me if I'm wrong.) I think preventing infinite loops is a pretty good UX step; it's a very unexpected failure mode for a user to face. My understanding is that, because halting problem, we're left manually whitelisting a non-turing-complete set of acceptable expressions. Because the whitelist is necessarily sub-turing-complete, the complexity of isTrivial can't get too bad, and it's probably okay to keep adding to it. Note that it's been added to it in the past (731edf0), and this patch would just continue the trend.

The only problem I see with the current solution is that it's not communicated to the user at all. I think it would help to change the error message from `Expected a <whatever> but got a thunk` to something like `Cannot evaluate non-trivial attribute which may infinitely loop`? Would this be a helpful improvement?

src/libexpr/eval.cc
```diff
@@ -258,14 +259,23 @@ PosIdx Value::determinePos(const PosIdx pos) const
...
-    return
-        internalType != tApp
-        && internalType != tPrimOpApp
-        && (internalType != tThunk
-            || (dynamic_cast<ExprAttrs *>(thunk.expr)
-                && ((ExprAttrs *) thunk.expr)->dynamicAttrs.empty())
-            || dynamic_cast<ExprLambda *>(thunk.expr)
-            || dynamic_cast<ExprList *>(thunk.expr));
+    if (internalType == tApp || internalType == tPrimOp) {
+        return true;
```
roberth, line 263: This clause isn't equivalent to the first two conjuncts in the original expression. Isn't it supposed to be?

roberth:
> something like `Cannot evaluate non-trivial attribute which may infinitely loop`?

It would be nice to avoid the undefined term "trivial" if we can, and provide some motivation.

I'd go for something like `Metadata must be easily retrievable, but the expression at ... is a ... so we can't be sure that it terminates soon enough.`

src/libexpr/eval.cc
lincolnauster: Whoops, thanks for catching that. Fixed up into aacc0af.

infinisil:
I do feel the need to point out the absurdity of having a Nix file be restricted to a loosely defined subset of Nix. It just smells like something you should never ever do.

lincolnauster:
On one hand, I agree with @infinisil - I can't imagine a scenario in which I'd a) write an infinite loop into `inputs.whatever.url`, b) ask Nix for metadata and c) not immediately realize my mistake after a second or two. On the other hand, an infinite loop is a pretty nix-y mistake to make that just doesn't show up too frequently elsewhere, and I can empathize with a new user cobbling together an expression just for nix to hang without explanation. A clear error message and a curated set of allowed expression types might be the way to go here: If part of the problem is that the subset is "loosely defined", clearly defining it in code and documentation (i.e., `nix flake help`) could be very useful. Is this the beginning of something merge-able?

tomberek:

PR #6762, opened 2022-07-05
Restrict the permissions of the CI

thufschmitt:
Follow-up from https://github.com/NixOS/nix/pull/6746

/cc @naveensrinivasan

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
   push:
 
+permissions: read-all
+
 jobs:
 
   tests:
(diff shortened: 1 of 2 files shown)


naveensrinivasan:
> Follow-up from #6746
> 
> /cc @naveensrinivasan

👍

naveensrinivasan (approved):
Thanks!

thufschmitt:
The hydra_status job was broken because of a yaml error. Fixed it

naveensrinivasan:
@thufschmitt  Something related to this. The actions aren't pinned by SHA which means someone can force a push tag which can be an attack vector. 

The downside of using an SHA is that it is hard to read for humans.  The SHA's are supported by dependabot.

PR #6766, opened 2022-07-05
fetchGit: allow fetching annotated tags

### LINKED ISSUE #3701: fetchGit can't handle annotated tags
## Description

Currently `builtins.fetchGit` does not seem to handle annotated tags and fails with:
```
error: cannot update ref 'refs/heads/xyz': trying to write non-commit object 098abc123 to branch 'refs/heads/xyz'
```

## Reproduction

You can see me reproduce this in https://github.com/status-im/status-react/issues/10825.

## Docs

This was reported once in #2385 and a fix for docs was applied in ae244af242ca3621e5a3b9196f27d9fcbf297266 by adding:
>Due to a bug #2385 only non-annotated tags can be fetched.

To documentation of `builtins.fetchGit`, but that change does not seem to be there in current docs:
https://nixos.org/nix/manual/#ssec-builtins
And there doesn't appear to be any fix in sight.

## Solutions

If the fix is difficult to implement for `builtins.fetchGit` could we at least update the documentation with this information and make `fetchGit` return an error that actually makes sense instead of this `trying to write non-commit object` nonsense?

balsoft:
Git doesn't like pointing a branch ref to an annotated tag, however the
other way around is fine. Use this to allow fetchGit to fetch annotated
tags, without specifically annotating them as such.

(hopefully) fixes https://github.com/NixOS/nix/issues/3701

**Please review with care! I'm not sure this doesn't break some important assumption elsewhere.**

Tested by poking around locally.

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -512,7 +512,7 @@ struct GitInputScheme : InputScheme
             Path localRefFile =
                 input.getRef()->compare(0, 5, "refs/") == 0
                 ? cacheDir + "/" + *input.getRef()
-                : cacheDir + "/refs/heads/" + *input.getRef();
+                : cacheDir + "/refs/tags/" + *input.getRef();
 
             bool doFetch;
             time_t now = time(0);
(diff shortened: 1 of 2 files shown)


thufschmitt:
After digging a bit into that, it turns out that the thing that matters is that annotated tags aren't commits (they are their own thing), and Git doesn't allow writing something that's not a commit hash under `/refs/heads` (check [here](https://github.com/git/git/blob/fe86abd7511a9a6862d5706c6fa1d9b57a63ba09/refs/files-backend.c#L1791-L1797)). Git being gentle-enough, it's nearly always fine to use a tag hash instead of a commit tag, so I think that fix is correct. Maybe for clarity we could use something else than `tags` though (since these refs don't have to be `tags`). I do like `nixrefs` :stuck_out_tongue: 

And obviously: Can you add some tests for this?

tomberek:

PR #6784, opened 2022-07-11
Add some tests for the CLI completion

thufschmitt:
Depends on #6693 (because it tests the new behavior it introduces)


tests/completions.sh
```diff
@@ -0,0 +1,48 @@
...
+    outputs = i: {
+        sampleOutput = 1;
+    };
+}
+EOF
+
+popd
+
+# Test the completion of a subcommand
+[[ $(printf "normal\nbuild\t\n") == $(NIX_GET_COMPLETIONS=1 nix buil) ]]
```
ncfavier, line 26:
```suggestion
[[ $(NIX_GET_COMPLETIONS=1 nix buil) == $'normal\nbuild\t\n' ]]
```

`==` isn't symmetrical, in that the LHS is treated as a single word so that you don't need to quote expansions, but the RHS is treated as a pattern so that unquoted glob characters will mess things up. Also, no need for a command substitution to get a fixed string.

tests/completions.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Filename completion
+[[ $(printf "filenames\n./foo\t\n") == $(NIX_GET_COMPLETIONS=2 nix build ./f) ]]
+[[ $(printf "filenames\n") == $(NIX_GET_COMPLETIONS=2 nix build ./nonexistent) ]]
+
+# Input override completion
+[[ $(printf "normal\na\t\n") == $(NIX_GET_COMPLETIONS=4 nix build ./foo --override-input '') ]]
+
+# Cli flag completion
+NIX_GET_COMPLETIONS=2 nix build --log-form | grep -- "--log-format"
```
ncfavier, line 37: `grep -q` maybe?

ncfavier (commented):
We should maybe also add tests for tilde expansion (`NIX_GET_COMPLETIONS=4 nix build ~/foo --override-input ''`) and completing inside a command line (`NIX_GET_COMPLETIONS=4 nix build flake1 --update-input '' flake2`)

tests/completions.sh
thufschmitt: I like having the `grep` output for this kind of things since it's hidden by default but it gives a direct feedback as to whether the grep call found something or not when debugging (as it can be a bit tricky to understand which command exactly failed in the middle of a complex pipeline). Not that I have a very strong opinion about it though

thufschmitt:
> We should maybe also add tests for tilde expansion (NIX_GET_COMPLETIONS=4 nix build ~/foo --override-input '') and completing inside a command line (NIX_GET_COMPLETIONS=4 nix build flake1 --update-input '' flake2)

Yep' done that.
The completion inside a command-line seems broken though (I didn't expect this to happen that soon :p ). Apparently it's still picking the inputs from the flake at `$PWD`. I've left the test, but commented out and with a pointer to #6693

ncfavier:
Two more test ideas:

- `nix flake show ./foo --override-input ''`
- multiple flakes: `nix build ./foo ./bar --override-input ''`

thufschmitt:
> That's because you used --override-input, which expects two arguments

Aaaah of course :facepalm: Fixed it, it works all right now

ncfavier (approved):

Ericson2314:
This PR has confused me somewhat while I have been working on https://github.com/NixOS/rfcs/pull/134. It appears we have gone from one two methods of doing completions? There is now both `completionHook` and `completeFlakeInput` which adds to the global completions variable.

It is not great having two methods of doing completions. We should have just one thing which can do everything we need.

ncfavier:
By "this PR" I guess you mean https://github.com/NixOS/nix/pull/6693?

`completeFlakeInput` is a helper, just like `completePath`, for completing a flake input.

`completionHook` was needed for reasons I documented in that PR.

You're welcome to rewrite the whole thing if you find a better way.

Ericson2314:
Ah yes whoops I did mean the other PR.

Ericson2314:
> You're welcome to rewrite the whole thing if you find a better way.

I might, but I am already in purgatory for making too many refactors.

PR #6794, opened 2022-07-12
Fix typo in flake.cc

eltociear:
non-existant -> non-existent


thufschmitt (approved):
Thanks :)

edolstra:
Note that this broke the flakes test, since it literally checked for "non-existant" in the output. This should be fixed as a side-effect of #6797.

We really should fix auto-merge, it seems to ignore the tests status entirely...

PR #6800, opened 2022-07-13
Ignore readonly variables used in ZSH for `nix print-dev-env`

ryanong:
EPOCHREALTIME, EPOCHSECONDS, and LINENO are read only variables in zsh.
When attempting to eval `nix print-dev-env` in ZSH it will raise an error saying the script is trying set read-only variables.


cole-h:
Should there be a note that they're zsh-specific variables?

ryanong:
After some more digging, it raises warnings but not errors

PR #6804, opened 2022-07-14
Disable auto-chroot if $NIX_STORE_DIR is set

### LINKED ISSUE #6732: building `lazy-options.json` fails (`NIX_STORE_DIR` usage broken?)
**Describe the bug**
```
building '/nix/store/ycczkwghxnz66qcvk244n22igy5crq9a-lazy-options.json.drv'...
lazy-options.json> warning: '/nix' does not exist, so Nix will use '/homeless-shelter/.local/share/nix/root' as a chroot store
lazy-options.json> error: creating directory '/homeless-shelter': Permission denied
lazy-options.json> Cacheable portion of option doc build failed.
lazy-options.json> Usually this means that an option attribute that ends up in documentation (eg `default` or `description`) depends on the restricted module arguments `config` or `pkgs`.
lazy-options.json> Rebuild your configuration with `--show-trace` to find the offending location. Remove the references to restricted arguments (eg by escaping their antiquotations or adding a `defaultText`) or disable the sandboxed build for the failing module by setting `meta.buildDocsInSandbox = false`.
```

The error seems to be related to the newly-introduced inference of a chroot store, which in this case seems errornous, because a presumably correct store location is given via `NIX_STORE_DIR` environment variable, set to `$TMPDIR/store`.

**Steps To Reproduce**

unknown; https://github.com/zseri/nixos-configs/commit/5366dea15f392b366e04284ec701b7f8424a05c6

**Expected behavior**

The error shouldn't happen, that is, building `lazy-options.json` should succeed.

nix-env (Nix) 2.10.0pre20220615_63df0fd

**Additional context**

edolstra:
Fixes #6732.

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -1321,7 +1321,7 @@ std::shared_ptr<Store> openFromNonUri(const std::string & uri, const Store::Para
         else if (pathExists(settings.nixDaemonSocketFile))
             return std::make_shared<UDSRemoteStore>(params);
         #if __linux__
-        else if (!pathExists(stateDir) && params.empty() && getuid() != 0) {
+        else if (!pathExists(stateDir) && params.empty() && getuid() != 0 && !getEnv("NIX_STORE_DIR").has_value()) {
             /* If /nix doesn't exist, there is no daemon socket, and
                we're not root, then automatically set up a chroot
                store in ~/.local/share/nix/root. */


NickCao:
This check should also be extended to `NIX_STATE_DIR`, see https://github.com/NixOS/nixpkgs/pull/181180#issuecomment-1185079527

Issue #6809, opened 2022-07-15
`nix print-dev-env` assumes that PATH and XDG_DATA_DIRS actually exist


jfly:
I use direnv with strict_env enabled, and with this [Hand-rolled Nix flakes integration](https://github.com/direnv/direnv/wiki/Nix#hand-rolled-nix-flakes-integration)

This demo.sh script effectively shows off what direnv does when I cd into a flake-enabled project:

    $ cat demo.sh
    #!/usr/bin/env bash

    set -euo pipefail
    eval "$(nix print-dev-env)"

This recently started failing for me after https://github.com/NixOS/nix/pull/6702/. Here's what started happening:

    $ ./demo
    ./demo.sh: line 6: XDG_DATA_DIRS: unbound variable

Note how `nix print-dev-env` now assumes that `XDG_DATA_DIRS` exists:

    $ nix print-dev-env
    unset shellHook
    nix_saved_PATH="$PATH"
    nix_saved_XDG_DATA_DIRS="$XDG_DATA_DIRS"
    ...

(There was already the same problem with `PATH`, but I've always had that
environment variable defined, I just don't have a `XDG_DATA_DIRS` set to
anything.)

I'll put together a PR to not assume that these environment variables are
actually set.

PR #6810, opened 2022-07-15
nix develop: do not assume that saved vars are set

jfly:
This fixes https://github.com/NixOS/nix/issues/6809


jfly:
Thanks for the quick turnaround!

Issue #5884, opened 2022-01-07
Invoking `nix` in a derivation's build script violates the macOS sandbox in Nix 2.4 and later


lilyball:
#### Describe the bug

Starting with Nix 2.4 and later, attempting to run `nix --version` (or `nix-env --version`) from within the macOS sandbox aborts with an uncaught exception caused by a `deny file-read-metadata`. The particular path it fails on differs between Nix 2.4 and Nix 2.5.1 but both fail. Nix 2.4 is failing on trying to read `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`

#### Steps To Reproduce

1. Turn on the sandbox in `/etc/nix/nix.conf` (in my case I have `sandbox = relaxed` but `true` works too)
2. `nix build --impure --expr 'with import (builtins.getFlake "github:NixOS/nixpkgs/77fda7f672726e1a95c8cd200f27bccfc86c870b") {}; runCommand "foo" { nativeBuildInputs = [ nix ]; } "nix --version"'` (the rev here is the current `nixpkgs/nixpkgs-unstable`)
3. Also try `nix build --impure --expr 'with import (builtins.getFlake "github:NixOS/nixpkgs/77fda7f672726e1a95c8cd200f27bccfc86c870b") {}; runCommand "foo" { nativeBuildInputs = [ nix_2_4 ]; } "nix --version"'` for Nix 2.4

#### Nix 2.5.1

Output:

```
libc++abi: terminating with uncaught exception of type nix::SysError: error: getting status of /System/Library/LaunchDaemons/com.apple.oahd.plist: Operation not permitted
/private/tmp/nix-build-foo.drv-0/.attr-0l2nkwhif96f51f4amnlf414lhl4rv9vh8iffyp431v6s28gsr90: line 1: 91536 Abort trap: 6           nix --version
```

The system log lists a number of sandbox denies, but the final one is

```
error	16:29:03.475029-0800	kernel	Sandbox: nix(91536) deny(1) file-read-metadata /System/Library/LaunchDaemons/com.apple.oahd.plist
```

#### Nix 2.4

Output:

```
libc++abi: terminating with uncaught exception of type nix::SysError: error: getting status of /nix/var/nix/profiles/per-user/root/channels/nixpkgs: Operation not permitted
[...]

PR #6812, opened 2022-07-15
fix(libstore): allow Nix to access all Rosetta 2 paths on MacOS

lovesegfault:
**Release Notes**

Fixes: #5884

I cherry-picked this from @centromere's tree, hopefully they're alright with
that.

This allows building `nix` and `cachix` on macOS with sandboxing enabled, among
other packages.

### DIFF
--- a/src/libstore/sandbox-defaults.sb
+++ b/src/libstore/sandbox-defaults.sb
@@ -98,7 +98,9 @@
 (allow file*
        (literal "/private/var/select/sh"))
 
-; Allow Rosetta 2 to run x86_64 binaries on aarch64-darwin.
+; Allow Rosetta 2 to run x86_64 binaries on aarch64-darwin (and vice versa).
 (allow file-read*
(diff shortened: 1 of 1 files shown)


lovesegfault:
It will need to be backported to 2.10

centromere:
Looks like I neglected to open a PR. Thank you @lovesegfault!

src/libstore/sandbox-defaults.sb
```diff
@@ -98,7 +98,9 @@
 (allow file*
        (literal "/private/var/select/sh"))
 
-; Allow Rosetta 2 to run x86_64 binaries on aarch64-darwin.
+; Allow Rosetta 2 to run x86_64 binaries on aarch64-darwin (and vice versa).
```
winterqt, line 101: > (and vice versa)

Rosetta 2 can't be run on x86_64 systems, maybe this comment should be updated to be more accurate?

PR #6813, opened 2022-07-16
libstore/globals.cc: Automatically set cores based on cgroup CPU limit

centromere:
By default, Nix sets the `cores` setting to the number of CPUs which are physically present on the machine. If cgroups are used to limit the CPU and memory consumption of a large Nix build, the OOM killer may be invoked.

For example, consider a GitLab CI pipeline which builds a large software package. The GitLab runner spawns a container whose CPU is limited to 4 cores and whose memory is limited to 16 GiB. If the underlying machine has 64 cores, Nix will invoke the build with `-j64`. In many cases, that level of parallelism will invoke the OOM killer and the build will completely fail.

This change sets the default value of `cores` to be ceil(cpu_quota / cpu_period), with a fallback to `std::thread::hardware_concurrency()` if cgroups v2 is not detected.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -11,11 +11,6 @@
 #include <dlfcn.h>
 #include <sys/utsname.h>
 
-#if __linux__
-#include <mntent.h>
-#include <cmath>
-#endif
-
 #include <nlohmann/json.hpp>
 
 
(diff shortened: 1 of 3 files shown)


src/libstore/globals.cc
```diff
@@ -114,7 +119,50 @@ std::vector<Path> getUserConfigFiles()
...
+        if (mountType == "cgroup2") {
+            cgPathParts.push_back(mountPath);
+            break;
+        }
+    }
+
+    fclose(fp);
+
+    if (cgPathParts.size() > 0 && pathExists("/proc/self/cgroup")) {
+        std::string currentCgroup = readFile("/proc/self/cgroup");
```
thufschmitt, line 147: I think this might blow-up if `/proc/self/cgroup` exists but is unreadable for whatever reason. Which is somewhat unlikely, but better be on the safe side

thufschmitt (approved):
Looks good overall :+1: A couple of suggestions:

- It would be nice to have the actual logic for this somewhere in `libutil` to keep the logic here a bit more high-level – in particular all these `#ifdef __linux__` inlined everywhere make everything harder to read, let's start to get rid of them :)
- There's – at least in theory – a few ways this code can blow up (if one of the file exists but is somehow unreadable or doesn't have the right format for whatever reason). It would be safer to wrap the whole thing in a `try-catch` block (Probably `catch (Error &) { ignoreException(); }` would be enough)

src/libstore/globals.cc
centromere: Thank you for the suggestions. I've made some changes. How does it look now?

thufschmitt (approved):
Thanks, that looks good now :)

It's a shame that we don't have the proper infrastructure to test this easily atm :/

PR #6814, opened 2022-07-16
local-derivation-goal.cc: improve error messages when sandboxing fails

amjoseph-nixpkgs:
The failure modes for nix's sandboxing setup are pretty complicated. When nix is unable to set up the sandbox, let's provide more detail about what went wrong.

Specifically, if sandbox setup fails:

* We make sure the error message includes the word "sandbox" so the user   knows that the failure was related to sandboxing.

* If `--option sandbox-fallback false` was provided, and removing it   would have allowed further attempts to make progress, remind the user that they have disabled `sandbox-fallback`.

* If `/proc/sys/user/max_user_namespaces` is missing or zero, `warn()` the user about this, since it is at least part of the reason for the failure.

* If `/proc/self/ns/user` does not exist, `warn()` the user about this, since it is likely an indication that the kernel was compiled without `CONFIG_USER_NS=y`.

* If `/proc/sys/kernel/unprivileged_userns_clone` is present (due to a [patch carried by Debian, Ubuntu, and other distributions](https://salsa.debian.org/kernel-team/linux/-/commit/d98e00eda6bea437e39b9e80444eee84a32438a6)) and is set to `0`, `warn()` the user about this, since it is at least part of the reason for the failure.

Personally, I've stumbled over all but the last of these at one point or another -- especially because there have been [a very large number](https://www.google.com/search?hl=en&q=cve%20%22max_user_namespaces%22) of local privilege escalation vulnerabilities in Linux's user namespaces for which the recommended mitigation was adding `echo 0 > /proc/sys/user/max_user_namespaces` early in the boot process, and I tend to forget about the fact that I've applied these mitigations.

I have exercised all of the above codepaths, including the `/proc/self/ns/user` check on a machine whose kernel did in fact lack `CONFIG_USER_NS=y`, and the `/proc/sys/kernel/unprivileged_userns_clone` check on a machine running a Debian-ified kernel.

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -862,6 +862,13 @@ void LocalDerivationGoal::startBuilder()
                             _exit(1);
                         if (!userNamespacesEnabled && errno==EPERM)
                             warn("user namespaces appear to be disabled; they are required for sandboxing; check /proc/sys/user/max_user_namespaces");
+                        if (userNamespacesEnabled) {
+                            Path procSysKernelUnprivilegedUsernsClone = "/proc/sys/kernel/unprivileged_userns_clone";
+                            if (pathExists(procSysKernelUnprivilegedUsernsClone)
+                                && trim(readFile(procSysKernelUnprivilegedUsernsClone)) == "0") {
+                                warn("user namespaces appear to be disabled; they are required for sandboxing; check /proc/sys/kernel/unprivileged_userns_clone");
+                            }
+                        }
                         Path procSelfNsUser = "/proc/self/ns/user";
                         if (!pathExists(procSelfNsUser))
                             warn("/proc/self/ns/user does not exist; your kernel was likely built without CONFIG_USER_NS=y, which is required for sandboxing");
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -855,11 +855,6 @@ void LocalDerivationGoal::startBuilder()
                 switch(errno) {
                     case EPERM:
                     case EINVAL: {
-                        /* Otherwise exit with EPERM so we can handle this in the
-                           parent. This is only done when sandbox-fallback is set
-                           to true (the default). */
-                        if (settings.sandboxFallback)
-                            _exit(1);
                         if (!userNamespacesEnabled && errno==EPERM)
                             notice("user namespaces appear to be disabled; they are required for sandboxing; check /proc/sys/user/max_user_namespaces");
                         if (userNamespacesEnabled) {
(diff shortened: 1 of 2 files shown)


src/libstore/build/local-derivation-goal.cc
```diff
@@ -845,18 +845,41 @@ void LocalDerivationGoal::startBuilder()
...
+                                warn("user namespaces appear to be disabled; they are required for sandboxing; check /proc/sys/kernel/unprivileged_userns_clone");
+                            }
+                        }
+                        Path procSelfNsUser = "/proc/self/ns/user";
+                        if (!pathExists(procSelfNsUser))
+                            warn("/proc/self/ns/user does not exist; your kernel was likely built without CONFIG_USER_NS=y, which is required for sandboxing");
+                        /* Mention sandbox-fallback in the error message so the user
+                           knows that having it disabled contributed to the
+                           unrecoverability of this failure */
+                        throw SysError("creating sandboxed builder process using clone(), without sandbox-fallback");
```
thufschmitt, line 878: I think this might not do what it should because the constructor of `SysError` reads `errno`, but at that point `errno` is probably no longer the right one

thufschmitt (changes requested):
Thanks, I think it can be quite helpful to have some basic builtin debugging :)
Left a couple of suggestions, but looks good overall

src/libstore/build/local-derivation-goal.cc
```diff
@@ -845,18 +845,41 @@ void LocalDerivationGoal::startBuilder()
...
+                switch(errno) {
+                    case EPERM:
+                    case EINVAL: {
+                        /* Otherwise exit with EPERM so we can handle this in the
+                           parent. This is only done when sandbox-fallback is set
+                           to true (the default). */
+                        if (settings.sandboxFallback)
+                            _exit(1);
+                        if (!userNamespacesEnabled && errno==EPERM)
+                            warn("user namespaces appear to be disabled; they are required for sandboxing; check /proc/sys/user/max_user_namespaces");
```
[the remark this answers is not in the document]
amjoseph-nixpkgs: I like that idea too.

99fcc91f67ece5a9646065665395f496d6a0cb84

src/libstore/build/local-derivation-goal.cc
amjoseph-nixpkgs: Thanks; I didn't see a version of `SysError()` that takes `errno` as an argument (rather than from the implicit global variable in the C runtime), so I added one... but my C++ is very stale; I'd appreciate it if you could look over 4e0c51786 in particular.

thufschmitt: Looks good :)

(What I actually had in mind was to construct the `SysError` at the start at the block when `errno` was still the right one and throw it at the end, but having an explicit `errno` argument is actually cleaner, so all the better)

thufschmitt (approved):
Thanks :)

I'll wait for the CI to complete, and if it doesn't complain I'll merge

src/libstore/build/local-derivation-goal.cc
amjoseph-nixpkgs: > (What I actually had in mind was to construct the `SysError` at the start at the block when `errno` was still the right one

Oh wow, that did not even occur to me!  If it had, I probably would've done that.

Issue #6507, opened 2022-05-09
A definition of "package" to clarify RFC-92 packages, and multi-drv packages that follow from the new definition


roberth:
**Describe the problem**

Packages and derivations are often the same thing. The lack of a definition and distinction between the two can not continue since RFC 92 (computed derivations, `outputOf`) and the conflation has unnecessarily made the concept of a multi-derivation package ill-formed / "unthinkable".

Nix doesn't really have a notion of "package". The term is only mentioned in a few places in the code, and only defined in the context of `buildenv` (ie legacy `nix-env`). This only related to the _usage_ of derivations in a profile, and therefore does not conflict with a definition of "package".

Nixpkgs on the other hand is all about packages, but it does not define precisely what a package is.

I propose the following definition:

A package is an attribute set with the following attributes:

 - `outputs`: list of strings
 - `${output}` for each `output` in `outputs`: store path string
 - `name`: string
 - `version`: string
 - `meta`: attribute set with specific optional(?) attributes
 - `tests`: optional attrset tree of derivations and/or packages

Notably absent from the definition of a package:

 - `drvPath`: this is an implementation detail. A package consumer really only cares about outputs, which don't need to be provided by a single derivation. Derivation path(s) can be recovered from the outputs.
 - `overrideAttrs`, `buildInputs`: attributes related to the construction of a derivation. These are implementation details
 - `passthru`: not related to the construction of a derivation, but an implementation detail for setting _package_ attrs as opposed to derivation attrs. This attribute owes its entire existence to `overrideAttrs`. Without `overrideAttrs`, `//` is sufficient to set package attributes.

**Steps To Reproduce**

1. Define a package where one output comes from a different derivation. You may want to do this to keep _derivation_ dependencies to a minimum (e.g. `doc` https://github.com/NixOS/nixpkgs/pull/172103 where it would be more desirable for the non-`doc` outputs not to depend on `texinfo`).
2. Be confused about what `drvPath` should be.
3. Install the package and note that the output from the non-`drvPath` output wasn't included.

**Expected behavior**

Nix represents packages by its outputs and metadata, not the `drvPath` implementation detail.

**`nix-env --version` output**

2.8. Or 2.x really. I would appreciate a major version increase for the (subtle) change in behavior.

**Additional context**

I came across this problem again in Nixpkgs today and figured I had to share my thoughts. I guess I should turn it into an RFC? I can do that later if y'all agree that we need something like this.

[...]

PR #6815, opened 2022-07-17
`OutputSpec` for `DerivationGoal` and `DerivedPath`, today's `OutputSpec` -> `ExtendedOutputSpec`

Ericson2314:
`DerivedPath::Built` and `DerivationGoal` were previously using a regular `std::set<std::string<` with the convention that the empty set means all outputs. But it is easy to forget about this rule when processing those sets. Using `OutputsSpec` forces us to get it right.

I did this because of the deduplication @edolstra requested in https://github.com/NixOS/nix/pull/4543#discussion_r922164687. We can't do much good code reuse when `DerivedPath` doesn't use `OutputsSpec` like it should.

We need to split `OutputsSpec` from `ExtendedOutputsSpec` because the "default" case only makes sense for high-level installables with an `outputsToInstall` field. For `DerivationGoal`, and `DerivedPath` itself, only concrete sets and the "all" wild-card make sense.

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -469,20 +469,14 @@ struct InstallableAttrPath : InstallableValue
 
         // Backward compatibility hack: group results by drvPath. This
         // helps keep .all output together.
-        std::map<StorePath, DerivedPath::Built> byDrvPath;
+        std::map<StorePath, OutputsSpec> byDrvPath;
 
         for (auto & drvInfo : drvInfos) {
             auto drvPath = drvInfo.queryDrvPath();
             if (!drvPath)
                 throw Error("'%s' is not a derivation", what());
 
-            auto derivedPath = byDrvPath.emplace(*drvPath, DerivedPath::Built {
-                .drvPath = *drvPath,
-                // Not normally legal, but we will merge right below
-                .outputs = OutputsSpec::Names { StringSet { } },
-            }).first;
-
-            derivedPath->second.outputs.merge(std::visit(overloaded {
+            auto newOutputs = std::visit(overloaded {
                 [&](const ExtendedOutputsSpec::Default & d) -> OutputsSpec {
                     std::set<std::string> outputsToInstall;
                     for (auto & output : drvInfo.queryOutputs(false, true))
@@ -492,12 +486,22 @@ struct InstallableAttrPath : InstallableValue
                 [&](const ExtendedOutputsSpec::Explicit & e) -> OutputsSpec {
                     return e;
                 },
-            }, extendedOutputsSpec.raw()));
+            }, extendedOutputsSpec.raw());
+
+            auto [iter, didInsert] = byDrvPath.emplace(*drvPath, newOutputs);
+
+            if (!didInsert)
+                iter->second = iter->second.union_(newOutputs);
         }
 
         DerivedPathsWithInfo res;
-        for (auto & [_, info] : byDrvPath)
-            res.push_back({ .path = { info } });
+        for (auto & [drvPath, outputs] : byDrvPath)
+            res.push_back({
+                .path = DerivedPath::Built {
+                    .drvPath = drvPath,
+                    .outputs = outputs,
+                },
+            });
 
         return res;
     }
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -446,19 +446,19 @@ struct InstallableAttrPath : InstallableValue
     SourceExprCommand & cmd;
     RootValue v;
     std::string attrPath;
-    OutputsSpec outputsSpec;
+    ExtendedOutputsSpec extendedOutputsSpec;
 
     InstallableAttrPath(
         ref<EvalState> state,
         SourceExprCommand & cmd,
         Value * v,
         const std::string & attrPath,
-        OutputsSpec outputsSpec)
+        ExtendedOutputsSpec extendedOutputsSpec)
         : InstallableValue(state)
         , cmd(cmd)
         , v(allocRootValue(v))
         , attrPath(attrPath)
-        , outputsSpec(std::move(outputsSpec))
+        , extendedOutputsSpec(std::move(extendedOutputsSpec))
     { }
 
     std::string what() const override { return attrPath; }
@@ -490,10 +490,10 @@ struct InstallableAttrPath : InstallableValue
 
             std::set<std::string> outputsToInstall;
 
-            if (auto outputNames = std::get_if<OutputNames>(&outputsSpec))
+            if (auto outputNames = std::get_if<OutputNames>(&extendedOutputsSpec))
                 outputsToInstall = *outputNames;
             else
-                for (auto & output : drvInfo.queryOutputs(false, std::get_if<DefaultOutputs>(&outputsSpec)))
+                for (auto & output : drvInfo.queryOutputs(false, std::get_if<DefaultOutputs>(&extendedOutputsSpec)))
                     outputsToInstall.insert(output.first);
 
             auto derivedPath = byDrvPath.emplace(*drvPath, DerivedPath::Built { .drvPath = *drvPath }).first;
@@ -581,7 +581,7 @@ InstallableFlake::InstallableFlake(
     ref<EvalState> state,
     FlakeRef && flakeRef,
     std::string_view fragment,
-    OutputsSpec outputsSpec,
+    ExtendedOutputsSpec extendedOutputsSpec,
     Strings attrPaths,
     Strings prefixes,
     const flake::LockFlags & lockFlags)
@@ -589,7 +589,7 @@ InstallableFlake::InstallableFlake(
       flakeRef(flakeRef),
       attrPaths(fragment == "" ? attrPaths : Strings{(std::string) fragment}),
       prefixes(fragment == "" ? Strings{} : prefixes),
-      outputsSpec(std::move(outputsSpec)),
+      extendedOutputsSpec(std::move(extendedOutputsSpec)),
       lockFlags(lockFlags)
 {
     if (cmd && cmd->getAutoArgs(*state)->size())
@@ -657,7 +657,7 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
             priority = aPriority->getInt();
     }
 
-    if (outputsToInstall.empty() || std::get_if<AllOutputs>(&outputsSpec)) {
+    if (outputsToInstall.empty() || std::get_if<AllOutputs>(&extendedOutputsSpec)) {
         outputsToInstall.clear();
         if (auto aOutputs = attr->maybeGetAttr(state->sOutputs))
             for (auto & s : aOutputs->getListOfStrings())
@@ -667,7 +667,7 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
     if (outputsToInstall.empty())
         outputsToInstall.insert("out");
 
-    if (auto outputNames = std::get_if<OutputNames>(&outputsSpec))
+    if (auto outputNames = std::get_if<OutputNames>(&extendedOutputsSpec))
         outputsToInstall = *outputNames;
 
     return {{
@@ -680,7 +680,7 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
             .originalRef = flakeRef,
             .resolvedRef = getLockedFlake()->flake.lockedRef,
             .attrPath = attrPath,
-            .outputsSpec = outputsSpec,
+            .extendedOutputsSpec = extendedOutputsSpec,
         }
     }};
 }
@@ -797,12 +797,12 @@ std::vector<std::shared_ptr<Installable>> SourceExprCommand::parseInstallables(
         }
 
         for (auto & s : ss) {
-            auto [prefix, outputsSpec] = OutputsSpec::parse(s);
+            auto [prefix, extendedOutputsSpec] = ExtendedOutputsSpec::parse(s);
             result.push_back(
                 std::make_shared<InstallableAttrPath>(
                     state, *this, vFile,
                     prefix == "." ? "" : prefix,
-                    outputsSpec));
+                    extendedOutputsSpec));
         }
 
     } else {
@@ -837,13 +837,13 @@ std::vector<std::shared_ptr<Installable>> SourceExprCommand::parseInstallables(
             }
 
             try {
-                auto [flakeRef, fragment, outputsSpec] = parseFlakeRefWithFragmentAndOutputsSpec(s, absPath("."));
+                auto [flakeRef, fragment, extendedOutputsSpec] = parseFlakeRefWithFragmentAndExtendedOutputsSpec(s, absPath("."));
                 result.push_back(std::make_shared<InstallableFlake>(
                         this,
                         getEvalState(),
                         std::move(flakeRef),
                         fragment,
-                        outputsSpec,
+                        extendedOutputsSpec,
                         getDefaultFlakeAttrPaths(),
                         getDefaultFlakeAttrPathPrefixes(),
                         lockFlags));
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -488,18 +488,23 @@ struct InstallableAttrPath : InstallableValue
             if (!drvPath)
                 throw Error("'%s' is not a derivation", what());
 
-            std::set<std::string> outputsToInstall;
+            auto derivedPath = byDrvPath.emplace(*drvPath, DerivedPath::Built {
+                .drvPath = *drvPath,
+                // Not normally legal, but we will merge right below
+                .outputs = OutputsSpec::Names { },
+            }).first;
 
-            if (auto outputNames = std::get_if<OutputNames>(&extendedOutputsSpec))
-                outputsToInstall = *outputNames;
-            else
-                for (auto & output : drvInfo.queryOutputs(false, std::get_if<DefaultOutputs>(&extendedOutputsSpec)))
-                    outputsToInstall.insert(output.first);
-
-            auto derivedPath = byDrvPath.emplace(*drvPath, DerivedPath::Built { .drvPath = *drvPath }).first;
-
-            for (auto & output : outputsToInstall)
-                derivedPath->second.outputs.insert(output);
+            derivedPath->second.outputs.merge(std::visit(overloaded {
+                [&](const ExtendedOutputsSpec::Default & d) -> OutputsSpec {
+                    std::set<std::string> outputsToInstall;
+                    for (auto & output : drvInfo.queryOutputs(false, true))
+                        outputsToInstall.insert(output.first);
+                    return OutputsSpec::Names { std::move(outputsToInstall) };
+                },
+                [&](const ExtendedOutputsSpec::Explicit & e) -> OutputsSpec {
+                    return e;
+                },
+            }, extendedOutputsSpec.raw()));
         }
 
         DerivedPathsWithInfo res;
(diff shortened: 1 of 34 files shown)


Ericson2314:
Hooray it is fixed!

Ericson2314:
@edolstra I rebased and in the final 4th commit cleaned up the parser, following up from #4543. This is ready for review!

src/libcmd/installables.cc
```diff
@@ -530,10 +530,12 @@ std::vector<InstallableValue::DerivationInfo> InstallableAttrPath::toDerivations
 
         std::set<std::string> outputsToInstall;
 
-        if (auto outputNames = std::get_if<OutputNames>(&extendedOutputsSpec))
+        auto * outputsSpec = std::get_if<ExtendedOutputsSpec::Explicit>(&extendedOutputsSpec);
```
edolstra, line 521: Is this variable used anywhere (other than in the `if`)?

src/libcmd/installables.cc
```diff
@@ -662,7 +664,9 @@ std::tuple<std::string, FlakeRef, InstallableValue::DerivationInfo> InstallableF
             priority = aPriority->getInt();
     }
 
-    if (outputsToInstall.empty() || std::get_if<AllOutputs>(&extendedOutputsSpec)) {
+    auto * outputsSpec = std::get_if<ExtendedOutputsSpec::Explicit>(&extendedOutputsSpec);
```
edolstra, line 655: Idem.

src/libstore/outputs-spec.cc
```diff
@@ -0,0 +1,159 @@
+#include "outputs-spec.hh"
+#include "store-api.hh"
+#include "nlohmann/json.hpp"
+
+#include <regex>
+
+namespace nix {
+
+bool OutputsSpec::contains(const std::string & outputName) const
```
roberth, line 9:
```suggestion
bool OutputsSpec::wants(const std::string & outputName) const
```

`contains` suggests that something exists, which we may not know yet. `wants` does not suggest this.

src/libstore/build/derivation-goal.cc
```diff
@@ -1322,7 +1320,14 @@ std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
...
     bool checkHash = buildMode == bmRepair;
-    auto wantedOutputsLeft = wantedOutputs;
+    auto wantedOutputsLeft = std::visit(overloaded {
+        [&](const OutputsSpec::All &) {
+            return StringSet {};
+        },
+        [&](const OutputsSpec::Names & names) {
+            return names;
+        },
+    }, wantedOutputs.raw());
```
edolstra, line 1330: This  is very similar to the code above (`realWantedOutputs`), so maybe this could be merged into a method on `OutputSpec`, e.g. `outputSpec.names_or(...)`.

src/libstore/build/derivation-goal.cc
```diff
@@ -1322,7 +1320,14 @@ std::pair<bool, DrvOutputs> DerivationGoal::checkPathValidity()
     if (!drv->type().isPure()) return { false, {} };
 
     bool checkHash = buildMode == bmRepair;
-    auto wantedOutputsLeft = wantedOutputs;
+    auto wantedOutputsLeft = std::visit(overloaded {
+        [&](const OutputsSpec::All &) {
+            return StringSet {};
```
tomberek, line 1325: presumably this is to minimize changes to the rest of this function. Should there be a TODO here?

src/libstore/misc.cc
```diff
@@ -301,4 +301,49 @@ std::map<DrvOutput, StorePath> drvOutputReferences(
     return drvOutputReferences(Realisation::closure(store, inputRealisations), info->references);
 }
 
+OutputPathMap resolveDerivedPath(Store & store, const DerivedPath::Built & bfd, Store * evalStore0)
```
edolstra, line 304: Maybe rename `evalStore0` to `evalStore_`. That's what we usually do for shadowed names.

edolstra: Actually why does `evalStore` need to be passed as a pointer instead of a reference?

src/libstore/outputs-spec.cc
```diff
@@ -6,57 +6,153 @@
...
-std::pair<std::string, ExtendedOutputsSpec> ExtendedOutputsSpec::parse(std::string s)
+bool OutputsSpec::contains(const std::string & outputName) const
 {
-    static std::regex regex(R"((.*)\^((\*)|([a-z]+(,[a-z]+)*)))");
+    return std::visit(overloaded {
+        [&](const OutputsSpec::All &) {
+            return true;
+        },
+        [&](const OutputsSpec::Names & outputNames) {
+            return outputNames.find(outputName) != outputNames.end();
```
edolstra, line 16: `return outputNames.count(outputName);`

src/libstore/outputs-spec.cc
```diff
@@ -6,57 +6,153 @@
...
+        [&](OutputsSpec::All &) {
+            /* If we already refer to all outputs, there is nothing to do. */
+            return false;
         },
-        [&](const ExtendedOutputsSpec::Names & outputNames) -> std::string {
-            return "^" + concatStringsSep(",", outputNames);
+        [&](OutputsSpec::Names & theseNames) {
+            return std::visit(overloaded {
+                [&](const OutputsSpec::All &) {
+                    *this = OutputsSpec::All {};
```
edolstra, line 98: I wonder if this is strictly legal C++...

src/libstore/outputs-spec.cc
```diff
@@ -6,57 +6,153 @@
...
-        [&](const ExtendedOutputsSpec::All &) -> std::string {
-            return "*";
+        [&](const ExtendedOutputsSpec::Explicit & outputSpec) -> std::string {
+            return "^" + outputSpec.print();
+        },
+    }, raw());
+}
+
+
+bool OutputsSpec::merge(const OutputsSpec & that)
```
edolstra, line 88: Maybe this can just return a new `OutputsSpec`?

src/libcmd/installables.cc
```diff
@@ -820,24 +808,46 @@ std::vector<std::shared_ptr<Installable>> SourceExprCommand::parseInstallables(
...
+                            // This handling predates the `^` syntax. The `^*` in
+                            // `/nix/store/hash-foo.drv^*` unambiguously means "do the
+                            // `DerivedPath::Built` case", so plain `/nix/store/hash-foo.drv` could
+                            // also unambiguously mean "do the DerivedPath::Opaque` case".
+                            //
+                            // Issue #7261 tracks reconsidering this `.drv` dispatching.
+                            return storePath.isDerivation()
+                                ? (DerivedPath) DerivedPath::Built {
+                                    .drvPath = std::move(storePath),
+                                    .outputs = {},
```
edolstra, line 836: What does `{}` as the outputs mean? I guess `AllOutputs`, but's it probably better to say that explicitly.

edolstra (approved):

src/libstore/outputs-spec.cc
roberth: fwiw, during the review, equality was mentioned, but this bool evaluates a subset relation instead.

src/libcmd/installables.cc
roberth: Can we delete that constructor?

Ericson2314: Ah! Good catch. So an issue is that `OutputNames` is a `typdef` and not it's own struct, so I am worried what is actually happening is that this is constructing an empty `std::set` which is then converted to an `OutputSpec`. In other words, I think this is a bug! I have at least seen this same bug developing the patch.

I think it would be better to make `OutputNames` its own struct (and we then yes can delete that constructor) to prevent this bug from sneaking in again. But I didn't do that because it makes some things, yes, more verbose.

src/libstore/misc.cc
Ericson2314: In the header it's has a default argument: `evalStore = nullptr`. I would like to do `evalStore = *this` and make it a reference but I don't think that is valid C++?

src/libstore/outputs-spec.cc
Ericson2314: For the record it is just used in one spot (I thought it might be used in more). I could change it or I could just inline it again as it was before whatever you all prefer.

src/libstore/outputs-spec.cc
Ericson2314: I have been thinking of `OutputSpec` as a merely the "finite power set of string ∪ ⊤", and so "contains" meant set contains for this notion of set of strings.

It is true that `OutputSpec` is usually *used* to represent the set of wanted outputs, but I feel a little odd calling the methods because I don't think it is the `OutputSpec` *itself* that does the wanting, it is merely used as means to convey something else's wants.

fricklerhandwerk:

- discussed how this fits with #6507
 - assigned @roberth for review

src/libstore/build/derivation-goal.cc
Ericson2314: Hehe I thought the same, but I think this is actually the natural way for this one specific check --- which is "did we refer to any outputs that don't exist?"

src/libstore/tests/outputs-spec.cc
```diff
@@ -0,0 +1,66 @@
...
+        auto [prefix, extendedOutputsSpec] = ExtendedOutputsSpec::parse("foo^out,bin");
+        ASSERT_EQ(prefix, "foo");
+        ASSERT_TRUE(std::get<OutputsSpec::Names>(std::get<ExtendedOutputsSpec::Explicit>(extendedOutputsSpec)) == OutputsSpec::Names({"out", "bin"}));
+    }
+
+    {
+        auto [prefix, extendedOutputsSpec] = ExtendedOutputsSpec::parse("foo^bar^out,bin");
+        ASSERT_EQ(prefix, "foo^bar");
+        ASSERT_TRUE(std::get<OutputsSpec::Names>(std::get<ExtendedOutputsSpec::Explicit>(extendedOutputsSpec)) == OutputsSpec::Names({"out", "bin"}));
+    }
```
roberth, line 63: Let's also test

```c++
ExtendedOutputsSpec::parse("*^bin")
ExtendedOutputsSpec::parse("*^*")
ExtendedOutputsSpec::parse("*^")
ExtendedOutputsSpec::parse("^*")
ExtendedOutputsSpec::parse("^")
ExtendedOutputsSpec::parse("bin^")
ExtendedOutputsSpec::parse("^bin")
ExtendedOutputsSpec::parse("drv!bin")
ExtendedOutputsSpec::parse("**")
```

"nonsensical" as they may be

src/libstore/outputs-spec.hh
```diff
@@ -0,0 +1,80 @@
...
+
+    inline const Raw & raw() const {
+        return static_cast<const Raw &>(*this);
+    }
+
+    /* Parse a string of the form 'prefix^output1,...outputN' or
+       'prefix^*', returning the prefix and the extended outputs spec. */
+    static std::pair<std::string_view, ExtendedOutputsSpec> parse(std::string_view s);
+
+    std::string print() const;
```
roberth, line 70: Not a big fan of `print` for a method that doesn't perform I/O.
The following aligns with existing code

```suggestion
    std::string to_string() const;
```

src/libstore/outputs-spec.hh
```diff
@@ -0,0 +1,80 @@
...
+};
+
+
+void to_json(nlohmann::json &, const OutputsSpec &);
+void from_json(const nlohmann::json &, OutputsSpec &);
+
+void to_json(nlohmann::json &, const ExtendedOutputsSpec &);
+void from_json(const nlohmann::json &, ExtendedOutputsSpec &);
```
roberth, line 78: These must be unit tested.

roberth (changes requested):
Some comments and suggestions for testing. I'm not too clever and I have the memory of a goldfish, so we have to do better with testing. This code in particular would benefit from unit tests, for the obvious reason I just stated, but also as a way to "document" the json interface, which should be stable. Lastly testing may reveal aspects that we might not otherwise think of.

Unofficial checklist.

 - [x] is the idea good? has it been discussed by the Nix team?
 - [x] unit tests
   - reused
    - is this purely refactoring? yes. was the documentation for outputs up to par? might be best to separate that out; scoped out for now
 - [x] documentation in the code (if necessary; ideally code is already clear)
    - will judge when the tests are written
 - [x] documentation in the commit message (why was this change made? for future reference when maintaining the code)
 - [x] n/a documentation in the changelog (to announce features and fixes to existing users who might have to do something to finally solve their problem, and to summarize the development history)

src/libstore/tests/outputs-spec.cc
```diff
@@ -0,0 +1,66 @@
...
+    {
+        auto outputsSpec = OutputsSpec::parse("out");
+        ASSERT_TRUE(std::get<OutputsSpec::Names>(outputsSpec) == OutputsSpec::Names({"out"}));
+    }
+
+    {
+        auto outputsSpec = OutputsSpec::parse("out,bin");
+        ASSERT_TRUE(std::get<OutputsSpec::Names>(outputsSpec) == OutputsSpec::Names({"out", "bin"}));
+    }
+
```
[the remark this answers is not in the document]
Ericson2314: Test that they fail?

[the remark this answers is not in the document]
roberth: I think so? Keeping the language small is a safe choice.

src/libstore/outputs-spec.hh
roberth: Doesn't have to be fancy; mostly to pin down the format.

Ericson2314:
I taught myself in https://github.com/NixOS/nix/pull/7543 how to write the tests, I'll return to this once that one is merge (they overlap slightly, so conflicts either way).

src/libcmd/installables.cc
Ericson2314: I rewrote this logic to hopefully be much easier to read.

src/libcmd/installables.cc
Ericson2314: I rewrote this logic to hopefully be much easier to read. There was some stuff that just seemed redundant? I deleted it but do check.

src/libstore/outputs-spec.cc
Ericson2314: @tfc? I am not sure.

tfc: It's not illegal, but I would certainly consider it a dangerous pattern.
Is it maybe possible to refactor this so we have a merge function that is not a member of one of the to-be-merged items?

src/libstore/outputs-spec.hh
```diff
@@ -1,32 +1,95 @@
...
 #include <variant>
 
-#include "util.hh"
-
 #include "nlohmann/json_fwd.hpp"
 
 namespace nix {
 
-typedef std::set<std::string> OutputNames;
+struct OutputNames : std::set<std::string> {
```
tfc, line 11: the STL container classes are *not* meant to be inherited from. Please don't.

src/libcmd/installables.cc
Ericson2314: They are removed.

src/libstore/tests/outputs-spec.cc
Ericson2314: Actually most of these do parse, because this really doesn't care what is to the right of a `^`.

src/libstore/outputs-spec.hh
Ericson2314: Is it a problem when it's a "newtype" like class --- exactly one superclass/field and so hopefully the same representation?

tfc: Hm... I see that you try to make it harder to create an empty set. 
Unfortunately I wouldn't deem this approach to be very effective since you can still write `auto myEmptyInstance = OutputNames({});` and it does not even look wrong, because the typename `OutputNames` does not convey that it's meant to never be empty. For this reason, one still has to put the same effort into the review to find out if we really never construct non-empty sets.

From a design perspective, this is not a case for inheritance.
I suggest to either drop this subclass which is from a design perspective more harm than good, or implement a class `template <typename T> class NonEmptySet` that *has* an `std::set` and that really implements the semantics of a non-empty data structure.

src/libstore/outputs-spec.hh
```diff
@@ -1,32 +1,91 @@
...
-    bool operator < (const AllOutputs & _) const { return false; }
+    /* This set should always be non-empty, so we delete this
+       constructor in order make creating empty ones by mistake harder.
+       */
+    OutputNames() = delete;
 };
 
-struct DefaultOutputs {
-    bool operator < (const DefaultOutputs & _) const { return false; }
+struct AllOutputs : std::monostate { };
```
tfc, line 28: I also suggest to not subclass this standard class. Looking over the code, it does not seem to bring much advantage.

src/libstore/outputs-spec.hh
```diff
@@ -1,32 +1,91 @@
...
+
+typedef std::variant<AllOutputs, OutputNames> _OutputsSpecRaw;
+
+struct OutputsSpec : _OutputsSpecRaw {
```
tfc, line 32: And here we have another subclass of STL standard classes... please, please, don't.

src/libstore/outputs-spec.hh
```diff
@@ -1,32 +1,91 @@
...
+    /* Parse a string of the form 'output1,...outputN' or
+       '*', returning the outputs spec. */
+    static OutputsSpec parse(std::string_view s);
+    static std::optional<OutputsSpec> parseOpt(std::string_view s);
+
+    std::string to_string() const;
 };
 
-typedef std::variant<DefaultOutputs, AllOutputs, OutputNames> OutputsSpec;
+struct DefaultOutputs : std::monostate { };
```
tfc, line 65: same pattern here. I have seen many code bases where developers subclassed stl classes - none of them aged well.

src/libstore/outputs-spec.hh
Ericson2314: For what it's worth, there was one "builder pattern" type situation where I did need to start with the empty set as an identity for merging proper ones in. So I dont really mind that there is an escape hatch. Of course yes the best thing to do would be to use a separate type for this one instance and make it ironclad.

src/libstore/outputs-spec.hh
Ericson2314: https://en.cppreference.com/w/cpp/utility/variant/monostate It gives me all the comparison functions and other such things I need.

src/libstore/outputs-spec.hh
Ericson2314: See https://github.com/NixOS/nix/issues/7479 we have an issue discussing this pattern. I am for it as the best that can be done for Sum types until we switch to Rust.

src/libstore/outputs-spec.cc
Ericson2314: It's gone!

src/libstore/outputs-spec.cc
Ericson2314: Done.

src/libstore/outputs-spec.hh
Ericson2314: Done.

src/libstore/outputs-spec.hh
roberth: It seems like that would be an appropriate use of `monostate`. It is a small class with a clear purpose, and by subclassing we make its meaning explicit.

What we're really doing here is declaring an algebraic data type, in a verbose, C++ way. Maybe that's why I don't see a problem, or may that's why there won't be a problem. Or maybe I'm misreading @tfc who isn't saying that this will be a problem per se and it isn't quite related to the comment about subclassing stl containers.

roberth (approved):
Looks really good, especially with the new tests, thanks @Ericson2314!
I'm also approving of using the established algebraic data type pattern, considering that it is better to discuss it as part of https://github.com/NixOS/nix/issues/7479. This new usage of it would not constitute any significant tech debt, as already a number of ADTs would have to be migrated to a different pattern.

roberth:
Merged to unblock further work. The remaining discussion wasn't critical and can be continued in #7479

edolstra:
@Ericson2314 Looks like this broke `build.aarch64-linux`, see https://hydra.nixos.org/build/205195243.

Ericson2314:
@edolstra @roberth told me about the problem and I made #7604 which I hope fixes it? Though I wasn't able to test it.

roberth:
I suspect that this PR has caused the regression:
 - #7655

PR #6824, opened 2022-07-20
s3-binary-cache-store: getFile streaming from S3

mupdt:
Before this change `nix copy --from s3://...` would download the entire object into memory before passing it on to the destination store. For bigger derivations this can lead to high memory usage and out-of-memory (OOM) errors. We experienced these OOM failures frequently.

This change adds support to stream objects from S3 to `nix copy`. This keeps the memory usage at a constant low level. We measured resident memory usage of roughly 40MB for packages of any size. To verify this patch works, we created a derivation with 10 GB of random data and did a before and after test. Without the patch we observed memory usage climbing to 10 GB. With the patch memory usage would settle at 40 MB.

We are now running this code in production and are no longer seeing OOM errors.

cc @thufschmitt

### DIFF
--- a/src/libstore/s3-binary-cache-store.cc
+++ b/src/libstore/s3-binary-cache-store.cc
@@ -183,6 +183,63 @@ S3Helper::FileTransferResult S3Helper::getObject(
 
     return res;
 }
+S3Helper::StreamResult S3Helper::getObjectStreamed(
+    const std::string & bucketName, const std::string & key, Sink & sink, size_t bufferSize)
+{
+    debug("streaming from 's3://%s/%s'...", bucketName, key);
+    auto now1 = std::chrono::steady_clock::now();
+    S3Helper::StreamResult res;
+
+    try {
+        size_t size = getObjectSize(bucketName, key);
+        size_t anchor = 0;
+        while (anchor < size)
+        {
+            const auto read_size = std::min(size - anchor, bufferSize);
+            size_t start = anchor;
+            size_t end = anchor + read_size;
+            anchor += read_size;
+
+            std::string range = boost::str(boost::format("bytes=%1%-%2%") % start % (end-1));
+            auto request =
+                Aws::S3::Model::GetObjectRequest()
+                .WithBucket(bucketName)
+                .WithKey(key)
+                .WithRange(range);
+
+            request.SetResponseStreamFactory([&]() {
+                return Aws::New<std::stringstream>("STRINGSTREAM");
+            });
+
+            auto result = checkAws(fmt("AWS error streaming '%s'", key), client->GetObject(request));
+            auto data = decompress(result.GetContentEncoding(),
+                dynamic_cast<std::stringstream &>(result.GetBody()).str());
+
+            sink(data);
+        }
+
+        res.streamSuccessful = true;
+        res.size = size;
+    } catch (S3Error & e) {
+        if ((e.err != Aws::S3::S3Errors::NO_SUCH_KEY) &&
+            (e.err != Aws::S3::S3Errors::RESOURCE_NOT_FOUND) &&
+            (e.err != Aws::S3::S3Errors::ACCESS_DENIED)) throw;
+    }
+
+    auto now2 = std::chrono::steady_clock::now();
+
+    res.durationMs = std::chrono::duration_cast<std::chrono::milliseconds>(now2 - now1).count();
+
+    return res;
+}
+
+size_t S3Helper::getObjectSize(const std::string & bucketName, const std::string & key){
+    Aws::S3::Model::HeadObjectRequest headObjectRequest;
+    headObjectRequest.WithBucket(bucketName).WithKey(key);
+    auto headOutcome = checkAws(fmt("AWS error checking object size '%s'", key), client->HeadObject(headObjectRequest));
+    long long len = headOutcome.GetContentLength();
+    return static_cast<size_t>(len);
+}
 
 S3BinaryCacheStore::S3BinaryCacheStore(const Params & params)
     : BinaryCacheStoreConfig(params)
@@ -202,7 +259,7 @@ struct S3BinaryCacheStoreConfig : virtual BinaryCacheStoreConfig
     const Setting<bool> multipartUpload{
         (StoreConfig*) this, false, "multipart-upload", "whether to use multi-part uploads"};
     const Setting<uint64_t> bufferSize{
-        (StoreConfig*) this, 5 * 1024 * 1024, "buffer-size", "size (in bytes) of each part in multi-part uploads"};
+        (StoreConfig*) this, 5 * 1024 * 1024, "buffer-size", "size (in bytes) of each part in multi-part uploads and downloads"};
 
     const std::string name() override { return "S3 Binary Cache Store"; }
 };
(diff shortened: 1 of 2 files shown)


src/libstore/s3-binary-cache-store.cc
```diff
@@ -184,6 +186,75 @@ S3Helper::FileTransferResult S3Helper::getObject(
...
+S3Helper::FileStreamResult S3Helper::getObjectStreamed(
+    const std::string & bucketName, const std::string & key, Sink & sink)
+{
+    debug("streaming from 's3://%s/%s'...", bucketName, key);
+
+    auto request =
+        Aws::S3::Model::GetObjectRequest()
+        .WithBucket(bucketName)
+        .WithKey(key)
+        .WithResponseContentEncoding("identity");
```
mupdt, line 229: Since we are streaming, we would need to get the content encoding before the download starts. Unfortunately, we were not able to find a way to inject the content encoding into the `streamtoSink` object before the streaming starts.

This is why we explicitly set the encoding to "identity" and are not applying the `decompress` step (unlike the old `getObject` function in L174 above).

Note that this is separate from nix's own compression of nar archives. Nix will still decompress nar archives if they were compressed when it uploaded them to S3.

src/libstore/s3-binary-cache-store.cc
```diff
@@ -184,6 +186,75 @@ S3Helper::FileTransferResult S3Helper::getObject(
...
+
+    auto now1 = std::chrono::steady_clock::now();
+
+    try {
+        checkAws(fmt("AWS error streaming '%s'", key), client->GetObject(request));
+        res.streamSuccessful = true;
+        res.size = streamToSink.totalSize;
+    } catch (S3Error & e) {
+        if ((e.err != Aws::S3::S3Errors::NO_SUCH_KEY) &&
+            (e.err != Aws::S3::S3Errors::RESOURCE_NOT_FOUND) &&
```
mupdt, line 247: This is another interesting difference from the old `getObject` function. AWS started to respond with `RESOURCE_NOT_FOUND` instead of `NO_SUCH_KEY` when the object does not exist in S3. We do not know why this started happening, but the only difference between our code and the old `getObject` function is the fact that we pin encoding to `identify`, perhaps internally somehow this causes AWS to change its response from `NO_SUCH_KEY` to `RESOURCE_NOT_FOUND`.

src/libstore/s3.hh
```diff
@@ -29,6 +30,16 @@ struct S3Helper
 
     FileTransferResult getObject(
```
mupdt, line 31: We left the old `getObject` function around because it's used in `filetransfer.cc`.

src/libstore/s3-binary-cache-store.cc
```diff
@@ -184,6 +186,75 @@ S3Helper::FileTransferResult S3Helper::getObject(
...
+    std::streamsize read(char* data, std::streamsize dataSize)
+    {
+        // Intentionally reports no data read. This struct is only intended for sinks.
+        return 0;
+    }
+
+    std::streamsize write(const char* data, std::streamsize dataSize)
+    {
+        auto strViewData = std::string_view(data, dataSize);
+        this->sink(strViewData);
```
mupdt, line 214: We discovered that `sink()` uses boost coroutines underneath (in `sinkToSource`). It turns out that the coroutine system can throw a `forced_unwind` exception. Surprisingly, this exception is caught and swallowed by `std::ostream::write`. However, this exception is supposed to be caught by the boost coroutine system, which will handle it and report any user-code exceptions properly. So, now we only get an obfuscated error log that looks like this:
```
nix: include/boost/context/detail/exception.hpp:37: boost::context::detail::forced_unwind::~forced_unwind(): Assertion `caught' failed.
```
We believe we have a fix for this, but we still have to run it through its paces. Stay tuned.

src/libstore/s3-binary-cache-store.cc
thufschmitt: My intuition is that it's a reasonable tradeoff since Nix already does its own compression most of the time, and most people will only use an `s3://` binary cache from inside aws (binary caches hosted on s3 but used outside of aws's datacenters will generally get an http proxy on front of them), so on a fast-enough network connection. @edolstra do you think there's a significant drawback in not having any compression on the transport layer here?

mupdt:
Quick update: First of all, apologies for the radio silence. We have since fixed the "corruption on retry" bug mentioned in the previous comment. We've been since using the patch in production.

Unfortunately, we have observed yet another a flake (rare, but persistent) which we believe to be a bug in our patch. We are in the process of tracking down the bug and will try to get a fix as soon as possible. Once we have the bugfix ready, we will use it internally until we are satisfied that there are no longer any bugs in the code. This might take a while, but we will comment on this PR when it happens and submit the final patch in this PR.

Apologies again and thank you for bearing with us!

mupdt:
Alright, this took much longer than expected, many apologies about that!

We managed to identify the cause for [the second set of flakes](https://github.com/NixOS/nix/pull/6824#issuecomment-1257925159). It turns out that AWS C++ SDK uses the `truncate` feature in streams (see [this issue in AWS C++ SDK](https://github.com/aws/aws-sdk-cpp/issues/715#issuecomment-341594388)). For example, if there's a retriable HTTP error response then the response's error body is written straight into the stream. AWS SDK code parses that error body, truncates the stream back to the end of the last successful body, and continues with the download streaming.

So, if we use a stream that doesn't support `truncate` (which we did) then we'd occasionally see retriable errors streaming garbage into our `nar` or `narinfo` streams. This would happen very rarely, hence the rare flakes.

We fixed this flake internally back on the 15th of October, stress-tested it for a while, and have then been using it in production since 25th of October. We haven't detected any flakes since and have downloaded in the order of ~100TiB (very rough estimate).

mupdt:
@thufschmitt: Could I ask you for another look? 🙏

roberth:
This PR needs a rebase.

Would you be interested in writing a test with minio? (https://github.com/NixOS/nix/issues/8239) While your practical testing is very encouraging, this would increase our confidence and help us maintain the S3 related code.

osnyx:
@mupdt You might want to have a look at the [Prometheus NixOS test](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests/prometheus.nix#L187), which integrates a separate node holding a minio S3 storage for the other test VM nodes to use.

src/libstore/s3-binary-cache-store.cc
```diff
@@ -183,6 +183,63 @@ S3Helper::FileTransferResult S3Helper::getObject(
...
 }
+S3Helper::StreamResult S3Helper::getObjectStreamed(
+    const std::string & bucketName, const std::string & key, Sink & sink, size_t bufferSize)
+{
+    debug("streaming from 's3://%s/%s'...", bucketName, key);
+    auto now1 = std::chrono::steady_clock::now();
+    S3Helper::StreamResult res;
+
+    try {
+        size_t size = getObjectSize(bucketName, key);
```
edolstra, line 194: This adds a `HEAD` request for every `GET` request, which I imagine adds a substantial amount of latency, especially for `.narinfo` files (which are very small). Do we really need to get the size in advance? Maybe we can start fetching from offset 0 and the response header will tell us what the full size is.

mupdt: I'll give it a try. Let's hope the S3 HTTP API supports this.

tomberek:

PR #6825, opened 2022-07-20
nix shell: document how to invoke multiple commands from the command …

rapenne-s:
while `nix-shell` allowed to run multiple commands in a shell environment like this: `nix-shell -p gnumake --run "cd src && make`, this isn't possible anymore with `nix shell`.

This PR adds an example for running multiple commands in a shell environment with `nix shell`

### DIFF
--- a/src/nix/shell.md
+++ b/src/nix/shell.md
@@ -23,6 +23,12 @@ R""(
   Hi everybody!
   ```
 
+* Run multiple commands in a shell environment:
+
+  ```console
+  # nix shell nixpkgs#gnumake -c /bin/sh -c "cd src && make"
+  ```
+
 * Run GNU Hello in a chroot store:
 
   ```console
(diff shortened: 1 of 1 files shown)


src/nix/shell.md
```diff
@@ -23,6 +23,12 @@ R""(
   Hi everybody!
   ```
 
+* Run multiple commands in a shell environment:
+
+  ```console
+  # nix shell nixpkgs#gnumake -c /bin/sh -c "cd src && make"
```
edolstra, line 29:
```suggestion
  # nix shell nixpkgs#gnumake -c sh -c "cd src && make"
```

Since this is Nix it would be nice to not depend on `/bin/sh` specifically.

fricklerhandwerk (approved):

rapenne-s:
Does it need further changes to be merged?

PR #6828, opened 2022-07-22
Introducing metadata fields to nix flake show’s json output for search.nixos.org

wyilio:
**Motivation**
I am working with @garbas on improving search.nixos.org and we are encountering difficulties with importing packages metadata. 
We are currently using some custom nix code that wraps each flake evaluation to export needed metadata. But this custom wrapper doesn’t work with repos that are the size of nixpkgs.
For nixpkgs we are currently resorting to using nix-env to be able to export all the metadata in an efficient way.
Our goal would be to deprecate the usage of nix-env and have nix flake show –json export all output information required for search in an efficient and standardized manner.

At this point I would only like to focus on metadata for packages as it is being defined (TODO: link) in nixpkgs. This means adding 3 more fields alongside a description field to JSON output (homepage, license, maintainers).

If possible I would like to bring this up for a discussion at some Nix related meeting.

**Current Progress:**
I have added a working solution to output JSON homepage, maintainers, and licenses. However, because the format of maintainers is a list of attributes, I had to write a `getListofAttrs` member function to `AttrCursor`. This currently does not use caching and returns a `std::vector` of `Bindings *`. Currently working on support for attribute lists nested in maintainers.


src/nix/flake.cc
```diff
@@ -1029,14 +1029,65 @@ struct CmdFlakeShow : FlakeCommand, MixJSON
...
+                        std::optional<std::vector<Symbol>> license;
                         if (auto aMeta = visitor.maybeGetAttr(state->sMeta)) {
                             if (auto aDescription = aMeta->maybeGetAttr(state->sDescription))
                                 description = aDescription->getString();
+                            if (auto aHomepage = aMeta->maybeGetAttr(state->sHomepage))
+                                homepage = aHomepage->getString();
+                            if (auto aMaintainers = aMeta->maybeGetAttr(state->sMaintainers))
+                                maintainers = aMaintainers->getListOfAttrs();
+                            if (auto aLicense = aMeta->maybeGetAttr(state->sLicense))
+                                license = aLicense->getAttrs();
```
thufschmitt, line 1043: Couldn't you rather have a "toplevel" `std::optional<AttrCursor> aLicense` that you set here and use [below](https://github.com/NixOS/nix/pull/6828/files#diff-98db1e334292594945051880fac0aecb52c9206898bd3cbc79730d647b260824R1079-R1090)?

```suggestion
                            aLicense = aMeta->maybeGetAttr(state->sLicense):
```

PR #6831, opened 2022-07-23
docker.nix: Pin nixpkgs in flake registry

centromere:
This patch allows the currently selected `nixpkgs` to be pinned in the Docker image's flake registry.

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -7,6 +7,7 @@
 , extraPkgs ? []
 , maxLayers ? 100
 , nixConf ? {}
+, sourceInfo ? null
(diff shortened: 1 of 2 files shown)


garbas (approved):

thufschmitt:
We discussed this in the Nix team meeting today.

This PR solves for a very specific use case. We think it would be sensible to, for example, instead use a generic post-build hook to add a registry to the image.

Also note that these docker builders are not officially supported and are only here as implementation details for building the official docker images.

We therefore decided to close this PR.

PR #6844, opened 2022-07-28
docker.nix: Allow Nix configuration to be customized

centromere:
This allows the default Nix configuration to be customized.


garbas (approved):
:+1:

PR #6845, opened 2022-07-28
manual: set -> attribute set

fricklerhandwerk:
also reword description to have shorter sentences.

addresses https://discourse.nixos.org/t/attribute-set-or-set/14253


fricklerhandwerk:
> Ideally, we would rename them to maps or dictionaries, since they're not actually sets.

Or maybe "records". But yes, too much of a change to rename the most prominent data type.

PR #6846, opened 2022-07-28
manual: use subheadings for primitive types

fricklerhandwerk:
this gives us HTML anchors for each of them.

### DIFF
--- a/doc/manual/src/expressions/language-values.md
+++ b/doc/manual/src/expressions/language-values.md
@@ -4,151 +4,161 @@
 
 Nix has the following basic data types:
 
-  - *Strings* can be written in three ways.
-    
-    The most common way is to enclose the string between double quotes,
-    e.g., `"foo bar"`. Strings can span multiple lines. The special
-    characters `"` and `\` and the character sequence `${` must be
-    escaped by prefixing them with a backslash (`\`). Newlines, carriage
-    returns and tabs can be written as `\n`, `\r` and `\t`,
-    respectively.
-    
-    You can include the result of an expression into a string by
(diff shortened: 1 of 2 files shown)


doc/manual/src/expressions/language-values.md
```diff
@@ -172,7 +180,7 @@ function and the fifth being a set.
 
 Note that lists are only lazy in values, and they are strict in length.
 
-## Attribute Sets
+## Attribute Set
 
 Attribute sets are collections of name-value-pairs (called *attributes*) enclosed in curly brackets (`{ }`).
```
cole-h, line 185: If you're de-pluralizing the heading, maybe the body should reflect that as well? (There may be more examples out of view of the diff.)

e.g.:
```suggestion
## Attribute Set

An attribute set is a collection of name-value-pairs (called *attributes*) enclosed in curly brackets (`{ }`).
```

thufschmitt (approved):
Thanks, anchors are cool :D 

If it's possible, I would have a (weak, do what you want with it) preference for encoding “definition lists” instead, by doing something like:

```markdown
- <a id="#type-string" href="#type-string">String</a>
    
    Strings can be writtten in three ways…
```

It's a bit cumbersome to write (and I'm not 100% sure that the md compiler will accept it), but the output is a bit more compact, and it's coherent with what we do in other places -- like for the [builtins](https://nixos.org/manual/nix/stable/expressions/builtins.html) or the [configuration options](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-accept-flake-config)

PR #6850, opened 2022-08-01
Fix Nix build --check flag

NinjaTrappeur:
See #2619 for more context.

Once a derivation goal has been completed, we check whether
this goal was meant to be repeated to check its output reproducibility.

An early return branch was preventing the worker to reach that repeat
code branch, hence breaking the --check command (#2619).

It seems like this early return branch is an artifact of a passed
refactoring. As far as I can tell, buildDone's main branch also
cleanup the tmp directory before returning.

The only difference here being that we are now deleting the output lock
files in a --check build. After reading a bit the code around this part, I think
this is the desired behavior, but I'm not 100% sure about that. If it's not the
case, we can guard the two lines doing that againt a `bmCheck` build.

-----------------------------------

How to test?

```
nix build "nixpkgs#hello" --rebuild --repeat 3
```

It'll only build hello once before this PR, 4 times after applying this patch.

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -914,12 +914,6 @@ void DerivationGoal::buildDone()
             outputPaths
         );
 
-        if (buildMode == bmCheck) {
-            cleanupPostOutputsRegisteredModeCheck();
-            done(BuildResult::Built, std::move(builtOutputs));
-            return;
-        }
-
         cleanupPostOutputsRegisteredModeNonCheck();
 
         /* Repeat the build if necessary. */
--- a/tests/check.sh
+++ b/tests/check.sh
@@ -40,6 +40,14 @@ nix-build check.nix -A deterministic --argstr checkBuildId $checkBuildId \
 if grep -q 'may not be deterministic' $TEST_ROOT/log; then false; fi
 checkBuildTempDirRemoved $TEST_ROOT/log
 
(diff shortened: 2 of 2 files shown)


NinjaTrappeur:
Could I introduce a regression test in https://github.com/NixOS/nix/blob/master/tests/check.sh or are these tests supposed to be for the "old" nix-build command only?

thufschmitt (commented):
Thanks!

I'd like @Ericson2314 to have a second look because I'm not 100% sure that it is _exactly_ the desired behavior, but in the worst case it's strictly better than having it outright broken.

> Could I introduce a regression test in https://github.com/NixOS/nix/blob/master/tests/check.sh or are these tests supposed to be for the "old" nix-build command only?

Yes, please do :)

NinjaTrappeur:
I added a regression test. Sadly, it seems disabled on the CI https://github.com/NixOS/nix/blob/8212adc15aad9b9c79e8cd6b06a0bf9f959e6e29/tests/check.sh#L4

I tested this regression test locally before and after applying this PR. It seems to behave as intended.

> I'd like @Ericson2314 to have a second look because I'm not 100% sure that it is exactly the desired behavior, but in the worst case it's strictly better than having it outright broken.

:+1:  As I said in the previous message. I'm not 100% sure this is the right thing to do wrt. 

```c++
outputLocks.setDeletion(true);
outputLocks.unlock();
```

thufschmitt:
Let's merge this since it's fixing something obviously broken

Issue #6359, opened 2022-04-04
tarball-ttl is documented as defaulting to 136 years in nix.conf(5)


alyssais:
**Describe the bug**

Something strange is happening in the manpage generation that results in an incorrect default value being recorded in the man page: (The actual default is 1 hour.)

```
       • tarball-ttl

         The  number of seconds a downloaded tarball is considered fresh. If
         the cached tarball is stale, Nix will check whether it is still  up
         to  date  using the ETag header. Nix will download a new version if
         the ETag header is unsupported, or the cached ETag doesn’t match.

         Setting the TTL to 0 forces Nix to always check if the  tarball  is
         up to date.

         Nix caches tarballs in $XDG_CACHE_HOME/nix/tarballs.

         Files fetched via NIX_PATH, fetchGit, fetchMercurial, fetchTarball,
         and fetchurl respect this TTL.

         Default: 4294967295
```

You can also see this in the nix.conf(5) man page for the nix package in Nixpkgs.

I couldn't reproduce when building the derivation in `nix-shell --pure default.nix` (with `genericBuild`).

**Steps To Reproduce**

1. <kbd>nix-build</kbd>
2. <kbd>man result/share/man/man5/nix.conf.gz</kbd>

**Expected behavior**

> Default: 3600

**`nix-env --version` output**

> nix-env (Nix) 3.0pre20200819_8a97b11

**Additional context**

n/a

cole-h:
Likely caused by these lines: https://github.com/NixOS/nix/blob/a4a1de69dcc3c6e0c40a093d67b5f20568a5f31e/src/nix/main.cc#L369-L370

alyssais:
Both this and #4622 (which is still an issue!) could be fixed by having e.g. a JSON file that is the source of truth for option names, descriptions, and defaults, rather than using C++ code as the source of truth.

patricksjackson:
This document mismatch also impacts `connect-timeout` (defined as `0` in code, but displays as `1` in the manual) and `download-attempts` (`tries` in the below code block, actual default `5`), so it is definitely due to the following block which only applies if there is no network.

https://github.com/NixOS/nix/blob/a4a1de69dcc3c6e0c40a093d67b5f20568a5f31e/src/nix/main.cc#L365-L375

PR #6854, opened 2022-08-02
fix(src/nix/main.cc): proper tarball-ttl default value

johnrichardrinehart:
# Description
https://github.com/NixOS/nix/issues/6359

The [manual](https://nixos.org/manual/nix/unstable/command-ref/conf-file.html?highlight=tarball-ttl#conf-tarball-ttl) reports a `MAXUINT32` for the default value `tarball-ttl` which is a very long time. The purported default is 1 hour. This changes the default value to match the value in [the docs](https://github.com/NixOS/nix/blob/780a4793866b2cac1040d522df514645d01c492f/src/libstore/globals.hh#L542).

### DIFF
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -375,7 +375,7 @@ void mainWrapped(int argc, char * * argv)
         if (!settings.useSubstitutes.overridden)
             settings.useSubstitutes = false;
         if (!settings.tarballTtl.overridden)
-            settings.tarballTtl = std::numeric_limits<unsigned int>::max();
+            settings.tarballTtl = 60 * 60;
         if (!fileTransferSettings.tries.overridden)
             fileTransferSettings.tries = 0;
         if (!fileTransferSettings.connectTimeout.overridden)


edolstra:
I'm not sure that an arbitrary 1 hour TTL is better than considering tarballs to be valid indefinitely. After all, if you *need* to get the latest, you should pass `--refresh`.

edolstra:
This breaks `--no-net`, which sets the TTL to infinity to ensure that we don't try to refresh tarballs when there is no network connection.

johnrichardrinehart:
So, is there a discrepancy between what's expressed at https://github.com/NixOS/nix/blob/7d1280bbaf7f4cd142c2259dec620c42bf6f96fd/src/libexpr/primops/fetchTree.cc#L305  and https://github.com/NixOS/nix/blob/7d1280bbaf7f4cd142c2259dec620c42bf6f96fd/src/libstore/globals.hh#L542 relative to what's indicated in the manual (link in original message)?

If so, how would you like to resolve it? It seems weird to me that this manual entry is generated assuming `--use-net=false`.

patricksjackson:
#7190 is an attempt to address the root of the problem.

Ericson2314:
Now that https://github.com/NixOS/nix/pull/7190 is merged, is this still needed?

johnrichardrinehart:
> Now that #7190 is merged, is this still needed?

No. I just checked the manual and the default values look great. Thanks for the work @patricksjackson !

PR #6865, opened 2022-08-04
manual: update docs when NIX_PATH is empty

zuzuleinen:
Update documentation how Nix resolves <nixpkgs> when `NIX_PATH` is not set

### DIFF
--- a/doc/manual/src/command-ref/env-common.md
+++ b/doc/manual/src/command-ref/env-common.md
@@ -46,9 +46,13 @@ Most Nix commands interpret the following environment variables:
 
     If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
 
-    1. `$HOME/.nix-defexpr/channels`
-    2. `nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
-    3. `/nix/var/nix/profiles/per-user/root/channels`
+      1. `$HOME/.nix-defexpr/channels`
+      2. `nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
+      3. `/nix/var/nix/profiles/per-user/root/channels`
+
+    If `NIX_PATH` is set to an empty string the above fallback will not happen and building will fail with:
+
+      `error: file 'nixpkgs' was not found in the Nix search path`
 
   - [`NIX_IGNORE_SYMLINK_STORE`]{#env-NIX_IGNORE_SYMLINK_STORE}\
     Normally, the Nix store directory (typically `/nix/store`) is not
--- a/doc/manual/src/command-ref/env-common.md
+++ b/doc/manual/src/command-ref/env-common.md
@@ -44,7 +44,7 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
-    If `NIX_PATH` environment variable is not set, Nix will fall back to the following list in impure and unrestricted evaluation mode:
+    If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
 
     1. `$HOME/.nix-defexpr/channels`
     2. `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
--- a/doc/manual/src/command-ref/env-common.md
+++ b/doc/manual/src/command-ref/env-common.md
@@ -44,7 +44,7 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
-    If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
+    If `NIX_PATH` is not set at all, Nix will fall back to the following list in [impure](./conf-file.md#conf-pure-eval) and [unrestricted](./conf-file.md#conf-restrict-eval) evaluation mode:
 
       1. `$HOME/.nix-defexpr/channels`
       2. `nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
--- a/doc/manual/src/command-ref/env-common.md
+++ b/doc/manual/src/command-ref/env-common.md
@@ -44,7 +44,7 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
-    If `NIX_PATH` is not set or empty, Nix will look up the following locations for a revision of `nixpkgs`:
+    If `NIX_PATH` environment variable is not set, Nix will fall back to the following list in impure and unrestricted evaluation mode:
 
     1. `$HOME/.nix-defexpr/channels`
     2. `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
--- a/doc/manual/src/command-ref/env-common.md
+++ b/doc/manual/src/command-ref/env-common.md
@@ -47,7 +47,7 @@ Most Nix commands interpret the following environment variables:
     If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
 
     1. `$HOME/.nix-defexpr/channels`
-    2. `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
+    2. `nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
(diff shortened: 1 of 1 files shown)


doc/manual/src/command-ref/env-common.md
```diff
@@ -44,6 +44,12 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
+    If `NIX_PATH` is not set or empty, Nix will look up the following locations for a revision of `nixpkgs`:
+
+    1. `$HOME/.nix-defexpr/channels`
+    2. `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
+    3. `/nix/var/nix/profiles/per-user/root/channels.`
```
edolstra, line 51:
```suggestion
    3. `/nix/var/nix/profiles/per-user/root/channels`
```

doc/manual/src/command-ref/env-common.md
```diff
@@ -44,6 +44,12 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
+    If `NIX_PATH` is not set or empty, Nix will look up the following locations for a revision of `nixpkgs`:
```
edolstra, line 47: This is not correct. It doesn't look for `nixpkgs` - it adds these directories to the search path, regardless of whether they contain `nixpkgs`. In particular, the first and third entries can contain non-nixpkgs channels.

jtojnar, line 47: The fallback also does not seem to be used with empty `NIX_PATH`:

```ShellSession
$ nix-channel --add https://nixos.org/channels/nixos-unstable nixpkgs
$ nix-channel --update
$ unset NIX_PATH
$ nix-instantiate -E '<nixpkgs>'
error: expression does not evaluate to a derivation (or a set or list of those)
$ NIX_PATH= nix-instantiate -E '<nixpkgs>'
error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)
```

fricklerhandwerk:
```suggestion
    If `NIX_PATH` is not set or empty, Nix will look up the following locations to set the search path:
```

We still have no explanations of the two mechanisms (`name=path` and how contents of a directory on the search path are mapped to `<name>` expressions), but that is out of scope here.
Eventually terms like "search path" should link to their definition, too.

zuzuleinen: I'm not too familiar with Nix. But locally, where I have an empty NIX_PATH, this command builds the go package for me:

`nix-build --no-out-link -E 'with import <nixpkgs> { }; go'`

jtojnar: Are you sure you have the variable set to an empty string rather than unset?

jtojnar: Maybe something like:

> If `NIX_PATH` environment variable is not set, Nix will fall back to the following list in impure and unrestricted evaluation mode:
> 
>     1. `$HOME/.nix-defexpr/channels`
>     2. `nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
>     3. `/nix/var/nix/profiles/per-user/root/channels`

zuzuleinen: I tried:

unset NIX_PATH && nix-build --no-out-link -E 'with import <nixpkgs> { }; go' 
env NIX_PATH="" && nix-build --no-out-link -E 'with import <nixpkgs> { }; go'

and both are building to `/nix/store/n449plwz10c5l53qhfnrywyns0m8g0qv-go-1.17.11`

fricklerhandwerk: @zuzuleinen You have to do
```console
 env NIX_PATH="" nix-build --no-out-link -E 'with import <nixpkgs> { }; go'
 ```
 instead. Note there is no `&&`, as just running `env` prints the environment, but running `env <assignments> <command>` runs the command with the assignments on top of the regular environment.

File: doc/manual/src/command-ref/env-common.md [hunk 3]
```diff
@@ -44,7 +44,7 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
-    If `NIX_PATH` is not set or empty, Nix will look up the following locations for a revision of `nixpkgs`:
+    If `NIX_PATH` environment variable is not set, Nix will fall back to the following list in impure and unrestricted evaluation mode:
```
fricklerhandwerk, line 47:
```suggestion
    If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
```

No reason to recall it's an environment variable (we have that from context). I suggest to emphasize that being set to empty and not being set is very different.

doc/manual/src/command-ref/env-common.md
zuzuleinen: Indeed running  `env NIX_PATH="" nix-build --no-out-link -E 'with import <nixpkgs> { }; go' ` will give me that error


```
error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at «string»:1:13:

            1| with import <nixpkgs> { }; go
             |             ^
```

doc/manual/src/command-ref/env-common.md
```diff
@@ -44,6 +44,12 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
+    If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
+
+    1. `$HOME/.nix-defexpr/channels`
+    2. `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```
jtojnar, line 50:
```suggestion
    2. `nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

doc/manual/src/command-ref/env-common.md
```diff
@@ -44,6 +44,16 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
+    If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
+
+      1. `$HOME/.nix-defexpr/channels`
+      2. `nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
+      3. `/nix/var/nix/profiles/per-user/root/channels`
+
+    If `NIX_PATH` is set to an empty string the above fallback will not happen and building will fail with:
```
fricklerhandwerk, line 53: Thanks for the quick response. Trying to be even more sharp:
```suggestion
    If `NIX_PATH` is set to an empty string, resolving search paths will always fail.
    For example, attempting to use `<nixpkgs>` will produce:
```

doc/manual/src/command-ref/env-common.md
```diff
@@ -44,6 +44,16 @@ Most Nix commands interpret the following environment variables:
...
+
+      `error: file 'nixpkgs' was not found in the Nix search path`
```
fricklerhandwerk, line 55:
```suggestion
Thanks for the quick response. Trying to be even more sharp:
```suggestion
    If `NIX_PATH` is set to an empty string, resolving search paths will always fail.
    For example, attempting to use `<nixpkgs>` will produce:

        error: file 'nixpkgs' was not found in the Nix search path
```

The error message should now render as code.

fricklerhandwerk (approved):
Great stuff, thanks for your patience @zuzuleinen!

amardeep:
Is this planned to be merged?

doc/manual/src/command-ref/env-common.md
```diff
@@ -44,6 +44,17 @@ Most Nix commands interpret the following environment variables:
     The Nix search path can also be extended using the `-I` option to
     many Nix commands, which takes precedence over `NIX_PATH`.
 
+    If `NIX_PATH` is not set at all, Nix will fall back to the following list in impure and unrestricted evaluation mode:
```
fricklerhandwerk, line 47:
```suggestion
    If `NIX_PATH` is not set at all, Nix will fall back to the following list in [impure](./conf-file.md#conf-pure-eval) and [unrestricted](./conf-file.md#conf-restrict-eval) evaluation mode:
```

fricklerhandwerk:
@amardeep thanks for the reminder. I did not have merge permissions back then, and do not keep track of things I can't do anything about.

PR #6265, opened 2022-03-15
Update glossary.md

stew:
add a definition of the nix database to the glossary

PR #6870, opened 2022-08-04
doc/manual: define {local,remote} store, binary cache, substituter

amjoseph-nixpkgs:
Nix veterans intuitively know what the following terms mean.  They are used in several places in the nix documentation, but never defined:

- local store
- remote store
- binary cache
- substituter

In particular, I found the last two terms to be confusingly similar. Let's give definitions for them.

This PR contains my attempt at giving a definition for each of these four terms, and mentioning the important-but-non-obvious fact that binary caches are kept in a different on-disk format than stores.  Not realizing this was one of the biggest obstacles back when I first started using nix -- I stumbled over this for several hours before realizing you can't just swap them back and forth.

If my definitions are wrong or just poorly worded please replace them.


edolstra:
This probably should go somewhere else since the manual is going to become a pure reference manual (and the "package management" section is outdated and will probably be removed).

amjoseph-nixpkgs:
> the manual is going to become a pure reference manual

Isn't that what a reference manual is for?  Providing definitions for product-specific jargon?  I wouldn't know where else to look for such information.

```
$ rg -e '((local|remote) store)|substituter|binary cache' doc/manual/ | wc -l
115
```

toraritte:
This is great and it should probably be put in the [Glossary](https://nixos.org/manual/nix/unstable/glossary.html).

doc/manual/src/package-management/terminology.md
```diff
@@ -0,0 +1,28 @@
...
+
+A *local store* exists on the local filesystem of the machine where
+Nix is invoked.  The `/nix/store` directory is one example of a
+local store.  You can use other local stores by passing the
+`--store` flag to `nix`.
+
+A *remote store* is a store which exists anywhere other than the
+local filesystem.  One example is the `/nix/store` directory on
+another machine, accessed via `ssh` or served by the `nix-serve`
+Perl script.
```
toraritte, line 11:
```suggestion
From   the  perspective   of   the  location   where  Nix   is
invoked<sup><b>1</b></sup>, the  Nix store can be  referred to
as a "_local_" or a "_remote_" one:

<sup>\[1]: Where "invoking Nix" means  an executing a Nix core
action/operation on  a Nix store.  For example, using  any CLI
commands from the `NixOS/nix` implementation.</sup>

+ A *local  store* exists  on the local  filesystem of
  the machine where Nix is  invoked. You can use other
  local stores  by passing  the `--store` flag  to the
  `nix` command.

+ A  *remote store*  exists  anywhere  other than  the
  local  filesystem. One  example is  the `/nix/store`
  directory on another machine,  accessed via `ssh` or
  served by the `nix-serve` Perl script.
```

I also removed the sentence "_The `/nix/store` directory is one example of a local store._" because the term "_Nix store_" should also be in the glossary. (The only place where it is defined is the [Introduction](https://nixos.org/manual/nix/unstable/introduction.html); planning to work on a PR about that.)

The footnote is mostly a placeholder that it should be explained what we mean by "invoking Nix".

doc/manual/src/package-management/terminology.md
```diff
@@ -0,0 +1,28 @@
...
+
+A *binary cache* is a remote store which is not the local store of
+any machine.  Examples of binary caches include S3 buckets and the
+[NixOS binary cache](https://cache.nixos.org).  Binary caches use a
+disk layout that is different from local stores; in particular, they
+keep metadata and signatures in `.narinfo` files rather than in
+`/nix/var/nix/db`.
```
toraritte, line 18:
```suggestion
A *binary cache* is a specialized Nix store whose metadata and
signatures are kept in `.narinfo` files rather than in the Nix
database. Examples of binary caches include S3 buckets and the
[NixOS binary cache](https://cache.nixos.org).
```
#6265 proposes to add the term "Nix database" to the [glossary](https://nixos.org/manual/nix/unstable/glossary.html) so we should take advantage of that.

As for the differences between a Nix store and a binary cache, I couldn't find a formal specification for the former, and the latter has an in-depth treatment by @fzakaria (even [a formal spec](https://fzakaria.github.io/nix-http-binary-cache-api-spec/)) which should definitely be incorporated.

doc/manual/src/package-management/terminology.md
```diff
@@ -0,0 +1,28 @@
...
+
+A *substituter* is a store other than `/nix/store` from which Nix will
+copy the realisation of a derivation instead of building it.  Nix will
+not copy a realisation from a remote store unless one of the following
+is true:
```
toraritte, line 23: > a store other than `/nix/store`
So a remote store? At least, as someone who has trouble to understand this part of Nix, I'm not sure what this refers to.

> the realisation of a derivation
I presume that this should be written as [store object](https://nixos.org/manual/nix/unstable/glossary.html#gloss-store-object).

amjoseph-nixpkgs: > So a remote store?

Or a local store:

```
mkdir ~/other-local-store
nix --store ~/other-local-store/ copy --from https://cache.nixos.org /nix/store/i46mrzinxi9a5incliwhksmk947ff4wn-stdenv-bootstrap-tools-mipsel-unknown-linux-gnu
nix copy --from ~/other-local-store/ /nix/store/i46mrzinxi9a5incliwhksmk947ff4wn-stdenv-bootstrap-tools-mipsel-unknown-linux-gnu
```



> > the realisation of a derivation
> 
> I presume that this should be written as [store object](https://nixos.org/manual/nix/unstable/glossary.html#gloss-store-object).

Yes!

amjoseph-nixpkgs: implemented: aae771c

amjoseph-nixpkgs:
> This is great and it should probably also be put in the [Glossary](https://nixos.org/manual/nix/unstable/glossary.html).

Wow, I don't know how I overlooked that.

Yes, that is exactly where this stuff belongs.

I will implement your feedback and then make a commit to move it (otherwise I can't use github's "commit suggestion" button)

doc/manual/src/package-management/terminology.md
amjoseph-nixpkgs: Since this stuff belongs in the glossary anyways, I propose to implement your suggestion above, and then move the entire chunk into the glossary as a starting point for the definition of "Nix store".

Obviously it isn't a complete entry since it still doesn't say what a "Nix store" is, but when that definition is added the two sub-cases will already be in the right place.

amjoseph-nixpkgs:
4de95f7f5 tries to at least acknowledge the question "so why do we need binary caches with a totally different on-disk format?"  If a better/longer answer to this question materializes later we can delete this text.  It's the other thing that drove me slightly crazy when I was first learning Nix.

It still seems a bit weird that we have two totally incompatible disk layouts for Nix stores, but I guess that's the only way to serve from S3 since it can't run a `nix-serve` daemon.

amjoseph-nixpkgs:
By the way, I've tried to match the existing formatting by hard-wrapping at 78 columns, but perhaps we shouldn't be hard-wrapping markdown text?  It makes the diffs really hard to read since github doesn't do `wdiff`.  Unlike code, editors can visually rewrap markdown pretty well.

doc/manual/src/package-management/terminology.md
toraritte: Thanks for the examples! I found "_a store other than `/nix/store` _" confusing, but not sure how to succinctly say "Nix store at a non-default location"...

doc/manual/src/package-management/terminology.md
toraritte: Sorry, meant to but subsequently forgot to open a placeholder issue for a new glossary entry for "Nix store". Here it is: #6875

doc/manual/src/glossary.md
```diff
@@ -11,6 +11,32 @@
     The location in the file system where store objects live. Typically
     `/nix/store`.
 
+    From   the  perspective   of   the  location   where  Nix   is
+    invoked<sup><b>1</b></sup>, the  Nix store can be  referred to
```
fricklerhandwerk, line 15:
```suggestion
    invoked[^1], the  Nix store can be  referred to
```

doc/manual/src/glossary.md
```diff
@@ -11,6 +11,32 @@
...
+    as a "_local_" or a "_remote_" one:
+
+    <sup>\[1]: Where "invoking Nix" means  an executing a Nix core
+    action/operation on  a Nix store.  For example, using  any CLI
+    commands from the `NixOS/nix` implementation.</sup>
```
fricklerhandwerk, line 20: Any reason we cannot use markdown footnotes (`[^1]` and `[^1]: ...`)? If no, please make sure the number is not otherwise used.

Other than that, I doubt the footnote helps. There is no such thing as a "Nix core action/operation", and if we want to reference the CLI commands, let us link the relevant manual section.

This also confuses me: Is there any other way to "invoke Nix" other than through commands? Sure, if you somehow call `libstore`, but does that really matter here? I think for the sake of getting the message across, invoking Nix should be enough.
```suggestion
    invoked, the  Nix store can be  referred to
    as a "_local_" or a "_remote_" one:
```

fricklerhandwerk:
@amjoseph-nixpkgs I agree on stopping the hard-wrapping, and have a [pull request for the nix.dev contribution guide](https://github.com/NixOS/nix.dev/pull/261/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R170), which is the [central place to coordinate documentation work](https://github.com/NixOS/nix.dev/issues/285#issuecomment-1177939217) and should (in the long run) aggregate best practices and set examples for other documentation resources to follow - something that should be easily visible but the [PR against nixos.org](https://github.com/NixOS/nixos-homepage/pull/875/files) is yet to be merged.

I suggest to apply that to new changes, and would be happy to see a PR against the contribution section. @edolstra objections?

Other than [my comment on the footnote](https://github.com/NixOS/nix/pull/6870#discussion_r940117881) this looks good.

fricklerhandwerk (approved):

kjeremy:
How does this reconcile the fact that the `binary-caches` option in `nix.conf` is deprecated in favor of `substituters` when they represent two distinct ideas?

doc/manual/src/glossary.md
```diff
@@ -29,6 +51,17 @@
     store object by downloading a pre-built version of the store object
     from some server.
 
+  - [substituter]{#gloss-substituter}\
+    A *substituter* is a store other than `/nix/store` from which Nix will
+    copy a store path instead of building it.  Nix will not copy a store
+    path from a remote store unless one of the following is true:
```
fricklerhandwerk, line 57:
```suggestion
    copy a store path instead of building it.  Nix will only copy a store
    path from a remote store if one of the following is true:
```

Ericson2314:
I like documenting these things in reference form, but this text is not right.

Here's what is going on.

1. There is a `Store` interface  (in `store-api.hh`). A few related interfaces extend it in various ways (e.g. `LocalFSStore`, in `local-fs-store.hh`, which means a Store that exposes stuff via the OS filesystem). There are various implementations of these interfaces.

2. By *convention* many commands / thing one can do can take a "main store", and also "other stores", in various capacities. "Substitors" are just stores used in a read-only way to get stuff we don't have in the main store.

3. "Binary cache" is/was another name for substitutor. In the implementation, it is also used for a family of stores which use compressed nars, narinfo files, etc. *as an implementation detail*. These implementation details are not very important.

So the thing to emphasize is the abstract interface of stores, versus the local store everyone is used to which has many "bonus features". "substitutor" sounds like an intrinsic property of a store, but it isn't, it is instead a term about *how the store is used*. It's thus somewhat a misnomer --- a classic example of natural language words literally referring to objects when the actual concept it notes is *a relation between objects*. "binary cache" is probably just too overloaded at this point and we should drop it. I would be fine renaming everything that is internally called `BinaryCache` to `NarinfoStore` or similar, for a fresh term without baggage.

fricklerhandwerk:
@Ericson2314 Then what about we consistently use`substitutor` and collapse the notes here into one?

doc/manual/src/glossary.md
```diff
@@ -11,6 +11,28 @@
...
+    + A *local  store* exists  on the local  filesystem of
+      the machine where Nix is  invoked. You can use other
+      local stores  by passing  the `--store` flag  to the
+      `nix` command.
+
+    + A  *remote store*  exists  anywhere  other than  the
+      local  filesystem. One  example is  the `/nix/store`
+      directory on another machine,  accessed via `ssh` or
+      served by the `nix-serve` Perl script.
+
```
Ericson2314, line 27: I would not start with this. We can do TCP to a local store in principle, avoiding use of the file system. We can use NFS and do file system access to storage that is far away. "local vs remote" is a misnomer from unjustified assumptions about how the file system is implemented.

Even in the implemented, the daemon on ones machine is communicated to with a `RemoteStore` subclass even though file system access is also used! (Should be called `RpcStore` or something instead.)

doc/manual/src/glossary.md
```diff
@@ -29,6 +51,17 @@
     store object by downloading a pre-built version of the store object
     from some server.
 
+  - [substituter]{#gloss-substituter}\
+    A *substituter* is a store other than `/nix/store` from which Nix will
```
Ericson2314, line 55:
```suggestion
    A *substituter* is an additional store from which Nix will
```

doc/manual/src/glossary.md
```diff
@@ -29,6 +51,17 @@
...
+    copy a store path instead of building it.  Nix will not copy a store
```
Ericson2314, line 56:
```suggestion
    copy store objects it doesn't have. Nix will not copy a store
```
Store path is just a reference, store object is the thing the reference refers to. We've been bad about this in the docs so far, so it's hardly your fault, but we should get better.

Also, it doesn't have to be something we'd otherwise build, it could be something we don't know how to build, that we *have* to substitute.

doc/manual/src/glossary.md
```diff
@@ -29,6 +51,17 @@
...
+    path from a remote store unless one of the following is true:
+
+    - the store object is signed by one of the `trusted-public-keys`
+    - the substituter is in the `trusted-substituters` list
+    - the `no-require-sigs` option has been set to disable signature checking
+    - the store object is a derivation
+    - the store object is the realisation of a fixed-output derivation
```
Ericson2314, line 63: This is a laundry list of special cases, we should break down the algorithm so it makes more sense.

- Content-addressed data is just data, and always "safe" to share. Derivations and fixed output derivations' outputs are just examples of content-addressed data.
- input-addressed things (everything is content-addressed or input-addressed) need at least one signature that corresponds to a trusted public key, unless we disabled this with `no-require-sigs`.

Ericson2314: Incidentally, this is a a great example of why @fricklerhandwerk's architecture docs are so important.

If we go straight into how user-facing knobs work, we have to either explain the data model on the fly (awkward diversion) or do a laundry list of familiar special cases (hard to remember, ad-hoc). But if we have a solid reference explaining the data model, we can hyperlink it instead, and not have to re-explain things or do laundry list. Yay!

Ericson2314 (approved):

doc/manual/src/glossary.md
Ericson2314: It is possible to sign content-addressed things, but this is an historical accident and it would be better if it didn't exist. In an ideal all-CA world, one only signs the "trust maps" which *relate* derviations and output.

Both the derivations and outputs themselves are plain old data and thus inherently trustworthy. Only the relationship between them is an unproven claim that needs to be authenticated with signatures.

edolstra, line 63: I don't think the glossary is the right place for this. A glossary should give brief definitions of terms, not a detailed description of the semantics of the store security model. Probably the manual needs a chapter on store path signing.

Ericson2314: Yeah it can be brief description + link to the section where the concept is describe in greater depth.

doc/manual/src/glossary.md
amjoseph-nixpkgs: > We can use NFS and do file system access to storage that is far away.

Okay; I have changed "local filesystem" to "filesystem".

> We can do TCP to a local store in principle, avoiding use of the file system.

Yes but you can't do builds in such a store.

Can you suggest a different term than "local" to express this distinction?

Note: the term "local store" is already in use to express this distinction:

https://github.com/NixOS/nix/blob/4823067247471aebdc24dceb21e13ee3d5411cb8/src/nix/run.cc#L44

Ericson2314: Bu you *can* do builds over TCP store, as long as the store "on the other end" can mount the items in some filesystem. That's how remote building works!

I'm with you that it would nice to break up the `Store` class further, but figuring out how to do this is surprisingly tricky.

doc/manual/src/glossary.md
amjoseph-nixpkgs: > This is a laundry list of special cases, we should break down the algorithm so it makes more sense.

What about `trusted-substituters`?

> * Derivations and fixed output derivations' outputs are just examples of content-addressed data.

Thanks, updated in f20a421c3 and then un-inlined in cb5ffeada.

I left the definition of "content-addressed derivation" as a FIXME because I'm not sure exactly how it should read.  Do you have a suggestion?  The obvious "has a `__contentAddressed=true` attr" conflicts with "all FODs are CADs", so I guess I'm not clear on the exact definition.

> a laundry list of familiar special cases (hard to remember, ad-hoc).

I don't think that's what this is.  It is meant to be an exhaustive list of non-overlapping conditions.  It is a security-critical detail; it needs to be specified somewhere, and at the moment it isn't.

The very simple question "when does Nix require a signature on something being added to the store" isn't answered anywhere.

> But if we have a solid reference explaining the data model, we can hyperlink it instead, and not have to re-explain things or do laundry list.

That's great; I look forward to the day when we have that and it provides a simple and direct answer to this important question.

amjoseph-nixpkgs: > I don't think the glossary is the right place for this. A glossary should give brief definitions of terms, not a detailed description of the semantics of the store security model. Probably the manual needs a chapter on store path signing.

I agree.  Can you suggest someplace where I can move this list until the indefinite time at which that chapter materializes?

doc/manual/src/glossary.md
amjoseph-nixpkgs: > Bu you can do builds over TCP store, as long as the store "on the other end" can mount the items in some filesystem.  That's how remote building works!

Come on.

You can *ask* a remote builder to do builds for you, in its local store on a filesystem that it has mounted.  And then you can ask it to send the build results back to you.

I agree it would be wonderful if we could erase any assumptions about things being on a local filesystem, and all uses of "local store" that currently exist in the source code and documentation.  Whenever that happens we should definitely delete this glossary entry.

amjoseph-nixpkgs:
> 3. "Binary cache" is/was another name for substitutor.

I got this terminology from `libstore`, which definitely uses the term "binary cache" (and `BinaryCacheStore`) in a way which does not encompass all substitut{o,e}rs.

In fact, a `BinaryCacheStore` isn't even a `RemoteStore`.  Neither extends the other.

This distinction isn't just internal to the codebase; it has leaked out onto user-facing surfaces.  I'm just trying to document the current situation, not advocate for it.

amjoseph-nixpkgs:
> How does this reconcile the fact that the `binary-caches` option in `nix.conf` is deprecated in favor of `substituters` when they represent two distinct ideas?

All binary caches are substituters but not vice versa.  Adding the more-general option `substituters` made the more specific option `binary-caches` unnecessary.

amjoseph-nixpkgs:
Boy I wish github would let me change peoples' non-repliable comments into repliable comments.  It's so much easier to read things with at least rudimentary threading...

doc/manual/src/glossary.md
amjoseph-nixpkgs: Apparently I can no longer "commit suggestion" on this, so I applied it manually (and moved the word "only" so it modifies "if" rather than modifying "copy").

doc/manual/src/glossary.md
amjoseph-nixpkgs: I've moved it to the description of `--option substituter` for now.  If there's a better place just let me know.

amjoseph-nixpkgs:
> 3\. "Binary cache" is/was another name for substitutor.

> This distinction isn't just internal to the codebase; it has leaked out onto user-facing surfaces.

Here's the one I stumbled over the same day I first tried nix:

https://github.com/NixOS/nix/blob/4823067247471aebdc24dceb21e13ee3d5411cb8/src/nix/copy.md?plain=1#L11
https://github.com/NixOS/nix/blob/4823067247471aebdc24dceb21e13ee3d5411cb8/src/nix/copy.md?plain=1#L12

Depending on whether you point `nix copy` at a `file://`-url or a bare path, it produces results in one of two totally different -- and totally incompatible formats.

Also, I will add an entry for "chroot store" since:

```
$ rg -i chroot.store | wc -l
11
```

edolstra:
> "Binary cache" is/was another name for substitutor.

That has never been the case. From the [very start](37f70984645b4fc5e6ed9f30a858ba6fbf402441), it was the name for a particular kind of substituter, namely the one using .narinfo etc. It replaced the manifest-based substituter (scripts/download-using-manifests.pl.in).  The confusion between the option names `binary-caches` / `substituters` probably comes from the fact that originally, `binary-caches` was an option of the binary cache substituter and so could *only* contain binary cache URLs. When substituters were generalized into store types (with the Perl -> C++ migration), the `binary-caches` option could be generalized into a list of store URLs.

I don't think calling it the "narinfo store" would be more descriptive to end users. "Binary cache" is not a great name but it is what is used everywhere now.

fricklerhandwerk:
> This distinction isn't just internal to the codebase; it has leaked out onto user-facing surfaces.

From the perspective of trying to present a sane interface in documentation, I think this is a bug. Such things should not be part of the manual, but documented in an issue.

fricklerhandwerk (approved):

PR #6871, opened 2022-08-05
docker.nix: Use JSON format for profile manifest

centromere:
This commit causes the profile manifest in the Docker image to be emitted in the newer `manifest.json` format instead of `manifest.nix`.

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -7,6 +7,7 @@
 , extraPkgs ? []
 , maxLayers ? 100
 , nixConf ? {}
+, sourceInfo ? null
 }:
 let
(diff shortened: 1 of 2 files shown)


thufschmitt:
What's the reason for that? Afaik using the old-style manifest is better (for the time being) since it works with `nix-env` and `nix develop` can migrate from it. But maybe I'm missing something?

Ericson2314:
Yes, the docker image should not be promoting unstable features.

centromere:
It is unclear to me how long it will take for flakes to graduate to stable, so I am going to close this for now.

PR #6880, opened 2022-08-07
Do not spam logs if the owned-homedir check results in a noop

virusdave:
Addresses log spam mentioned [here](https://github.com/NixOS/nix/pull/6676#issuecomment-1207414685)

Admittedly, this doesn't attempt to canonicalize the paths before comparison, so it's just a quick & dirty, best-effort way to reduce log spam.  Further improvement is probably possible.

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -577,6 +577,7 @@ Path getHome()
 {
     static Path homeDir = []()
     {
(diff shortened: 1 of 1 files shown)


virusdave:
@thufschmitt mind taking a look at this little change?

thufschmitt (approved):
Looks good.
Indeed an annoying edge-case worth fixing

PR #6882, opened 2022-08-08
Allow installing multi-user as root on macOS and Linux

### LINKED ISSUE #6801: Can't use the multi-user installer as root
**Describe the bug**

If you're logged in as root, `sh <(curl -L https://nixos.org/nix/install) --daemon` will not work because of the following lines:

https://github.com/NixOS/nix/blob/e1153069bd0e3320227c997c32e657f7e652a212/scripts/install-multi-user.sh#L873-L878

**Steps To Reproduce**

1. Log in as root
1. `sh <(curl -L https://nixos.org/nix/install) --daemon`
2. Notice that it bails out

**Expected behavior**

I'd expect to still be allowed to install the multi-user version of Nix (e.g. as a bootstrap step in setting up a new machine that doesn't yet have any other users) when running it as root.

N/A.

**Additional context**

It's possible to work around it by prefixing it with `env EUID=65534 ...` (or any non-zero `EUID`), but that's not ideal.

grahamc:
This may work fine on a mac, but I don't have one for testing. I haven't tested this exact change on Linux yet, but it operates on the idea that you can "just" set EUID=65535 as root and the installer works fine. Going to test this shortly.

Closes #6801, cc @cole-h.

### DIFF
--- a/doc/manual/src/installation/installing-binary.md
+++ b/doc/manual/src/installation/installing-binary.md
@@ -13,7 +13,7 @@ for your platform:
 - multi-user on macOS
 
   > **Notes on read-only filesystem root in macOS 10.15 Catalina +**
-  > 
+  >
   > - It took some time to support this cleanly. You may see posts,
   >   examples, and tutorials using obsolete workarounds.
   > - Supporting it cleanly made macOS installs too complex to qualify
@@ -75,7 +75,7 @@ should run this under your usual user account, *not* as root. The script
 will invoke `sudo` as needed.
 
 > **Note**
-> 
+>
 > If you need Nix to use a different group ID or user ID set, you will
 > have to download the tarball manually and [edit the install
 > script](#installing-from-a-binary-tarball).
@@ -167,7 +167,7 @@ and `/etc/zshrc` which you may remove.
    removed next.
 
 7. Remove the Nix Store volume:
-   
+
    ```console
    sudo diskutil apfs deleteVolume /nix
    ```
@@ -176,7 +176,7 @@ and `/etc/zshrc` which you may remove.
    store.
 
 > **Note**
-> 
+>
 > After you complete the steps here, you will still have an empty `/nix`
 > directory. This is an expected sign of a successful uninstall. The empty
 > `/nix` directory will disappear the next time you reboot.
--- a/doc/manual/src/installation/installing-binary.md
+++ b/doc/manual/src/installation/installing-binary.md
@@ -31,8 +31,8 @@ $ sh <(curl -L https://nixos.org/nix/install) --no-daemon
 ```
 
 This will perform a single-user installation of Nix, meaning that `/nix`
-is owned by the invoking user. You should run this under your usual user
-account, *not* as root. The script will invoke `sudo` to create `/nix`
+is owned by the invoking user. You can run this under your usual user
+account or root. The script will invoke `sudo` to create `/nix`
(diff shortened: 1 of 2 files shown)


scripts/install-multi-user.sh
```diff
@@ -313,10 +337,15 @@ __sudo() {
...
     local expl="$1"
     shift
-    if ! headless; then
+    if ! headless || is_root; then
         __sudo "$expl" "$*" >&2
     fi
-    sudo "$@"
+
+    if is_root; then
+        env "$@"
```
grahamc, line 345: Note this uses `env` because some _sudo callsites pass environment variables to the program being executed.

scripts/install-multi-user.sh
```diff
@@ -59,6 +59,30 @@ headless() {
...
+    if [ "$EUID" -eq 0 ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+is_os_linux() {
+    if [ "$(uname -s)" = "Linux" ]; then
+        return 0;
```
cole-h, line 72: Nit:

```suggestion
        return 0
```

scripts/install-multi-user.sh
```diff
@@ -59,6 +59,30 @@ headless() {
...
+    else
+        return 1
+    fi
+}
+
+is_os_darwin() {
+    if [ "$(uname -s)" = "Darwin" ]; then
+        return 0;
```
cole-h, line 80: Nit:

```suggestion
        return 0
```

doc/manual/src/installation/installing-binary.md
```diff
@@ -31,8 +31,8 @@ $ sh <(curl -L https://nixos.org/nix/install) --no-daemon
 ```
 
 This will perform a single-user installation of Nix, meaning that `/nix`
-is owned by the invoking user. You should run this under your usual user
-account, *not* as root. The script will invoke `sudo` to create `/nix`
+is owned by the invoking user. You can run this under your usual user
+account or root. The script will invoke `sudo` to create `/nix`
```
cole-h, line 35: "or root" seems to only be the case on Linux (as seen in the conditional on line 903 in this diff), so should that be explicitly called out here?

grahamc: Good point. I'd love to try this on an actual mac and see what happens...

cole-h: I just tried this on a Mac VM (x86), and it worked just fine running as root using `env EUID=65534`, so the check can probably be dropped altogether?

image

grahamc: Fixed up this review feedback by allowing installing as root on macOS too.

grahamc:
Tested the resulting installer, I think this is good to go.

biscuitPIMP:
You need a MacBook Air to test it? I. An fedex you one to try I have a few lying around I been hoarding them and raid ing costxo

edolstra:
It would be great to test this on macOS as root, but I'll merge it since it shouldn't break existing usage.

cole-h:
FYI: it was tested on an x86 macOS VM (not aarch64 or on real hardware, but hopefully a close enough approximation?): https://github.com/NixOS/nix/pull/6882#discussion_r941603130. So, ideally, everything should be fine.

grahamc:
I tested this on an aarch64 mac through MDM and it worked great!

PR #6884, opened 2022-08-09
Don't point to `nix --help` when the config is invalid

### LINKED ISSUE #6878: nix-build --help fails with invalid nix.conf
It looks like nix-build's help flag can't be used without a valid config (even though the message for an invalid config tells you to check the help). Clipping original comment below:

---

... I'm now receiving the following error after entering `nix-build -A plutus-playground.server`:

```
error: illegal configuration line 'https://cache.nixos.org/' in '/etc/nix/nix.conf'
Try 'nix-build --help' for more information.
```

And running `nix-build --help` produces the same result: 

```
error: illegal configuration line 'https://cache.nixos.org/' in '/etc/nix/nix.conf'
Try 'nix-build --help' for more information.
```

_Originally posted by @Dave-Lindberg in https://github.com/NixOS/nix/issues/6876#issuecomment-1207138311_

thufschmitt:
`nix --help` won't run with an invalid config file, so that's not a
really good suggestion. Point to the manpage instead.

This isn't totally ideal either since the man page is arguably less nice
and might not be here at all, but at least it's *something*

Fix #6878

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -382,6 +382,10 @@ int handleExceptions(const std::string & programName, std::function<void()> fun)
         }
     } catch (Exit & e) {
         return e.status;
+    } catch (ConfigError & e) {
+        logError(e.info());
+        printError("Try 'man nix.conf' for more information.");
+        return 1;
     } catch (UsageError & e) {
         logError(e.info());
         printError("Try '%1% --help' for more information.", programName);
--- a/src/libutil/config.cc
+++ b/src/libutil/config.cc
@@ -94,7 +94,7 @@ void AbstractConfig::applyConfig(const std::string & contents, const std::string
         if (tokens.empty()) continue;
 
         if (tokens.size() < 2)
-            throw UsageError("illegal configuration line '%1%' in '%2%'", line, path);
+            throw ConfigError("illegal configuration line '%1%' in '%2%'", line, path);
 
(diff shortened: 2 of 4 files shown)


edolstra:
I think the real issue is that `nix --help` shouldn't fail if the config is broken. We can't rely on the manpage existing (e.g. when using the statically-linked `nix` binary).

thufschmitt:
Yes, I totally agree. That one is just a stopgap fix because making `nix --help` work with an invalid config isn't trivial (also I imagine that at least the `plugin-files` option might change the behavior of `--help`, but that's too much of an edge-case to be worth it)

Ericson2314 (approved):
Even if we get the better fix, I think having the dedicated exception type is nice.

I therefore think this is a good stop-gap.

fricklerhandwerk:

- @edolstra: replaces a potentially unhelpful error message with another
    - we should fix that. `nix --help` should always work
    - @ericson2314: we're pretty far from fixing that
        - @thufschmitt: yes, config loading is pretty deep in the initialisation logic
    - we're already liberal with how we treat configuration files. cold just ignore a broken one
        - @ericson2314: this could have security implications. but should be fine if only used for `--help`
        - @edolstra: `git` prints an error and just proceeds, for example
- decision: catch error only for `--help`

tomberek:

PR #6887, opened 2022-08-10
nix-shell: specify which outputs from bashInteractive to build

winterqt:
This prevents downloading additional outputs that aren't needed for the shell.


edolstra:
I think it's better to rely on the default outputs specified by `bashInteractive` (currently `out` and `man), since then it won't break if we ever move bash to a `bin` output or something like that. `nix develop` does that, not sure about `nix-shell`.

winterqt:
`nix-shell` hardcodes `out`:
https://github.com/NixOS/nix/blob/af4e8b00fb986acf32d7e4cd4fff7218b38958df/src/nix-build/nix-build.cc#L444

PR #6906, opened 2022-08-15
add syntax overview from NixOS manual

fricklerhandwerk:
taken verbatim to keep track of required corrections, adding things as I go.

made it an HTML table to more easily change structure and keep diffs minimal.

### DIFF
--- a/doc/manual/src/language/index.md
+++ b/doc/manual/src/language/index.md
@@ -106,6 +106,14 @@ the rest.
   </tr>
   <tr>
     <td>
+       `~/.config`
+    </td>
+    <td>
+      A home path. Evaluates to the `"<user's home directory>/.config"`.
+    </td>
+  </tr>
+  <tr>
+    <td>
        <nixpkgs>
     </td>
     <td>
--- a/doc/manual/src/language/index.md
+++ b/doc/manual/src/language/index.md
@@ -73,7 +73,7 @@ the rest.
        `./foo.png`
     </td>
     <td>
-      A path (relative to the containing Nix expression)
+      A path relative to the file containing this Nix expression
     </td>
   </tr>
   <tr>
--- a/doc/manual/src/language/index.md
+++ b/doc/manual/src/language/index.md
@@ -403,7 +403,7 @@ This is an incomplete overview of language features, by example.
  <tr>
   <td>
 
-   `with pkgs.lib; head [ 1 2 3 ]`
+   `with builtins; head [ 1 2 3 ]`
 
   </td>
   <td>
--- a/doc/manual/src/language/index.md
+++ b/doc/manual/src/language/index.md
@@ -53,7 +53,7 @@ the rest.
        ```
     </td>
     <td>
-      A multiline string. Strips common prefixed whitespace. Evaluates to `"multi\n line\n  string"`.
+      A multi-line string. Strips common prefixed whitespace. Evaluates to `"multi\n line\n  string"`.
     </td>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


rapenne-s:
Looks good to me! It's helpful.

Could it be possible to add details about type casting? A few days ago, I searched for a way to use an integer variable into a multiline string, and it took me a while to find the solution on a random post on NixOS discourse :exploding_head:

doc/manual/src/language/index.md
```diff
@@ -11,3 +11,551 @@ packages, compositions of packages, and the variability within packages.
...
+  <td>
+
+   Variable definition
+
+  </td>
+ </tr>
+ <tr>
+  <td>
+
+   `with pkgs.lib; head [ 1 2 3 ]`
```
rapenne-s, line 406: I find this example puzzling, why not writing it `pkgs.lib.head [ 1 2 3]`?

maybe something like

> `with pkgs; [ awk git nix ]`
> Add all attributes from the given set (`pkgs` here) to the current scope. Otherwise, you would have to write  `[ pkgs.awk pkgs.git pkgs.nix ]`

fricklerhandwerk: This is from the original NixOS syntax overview. That line should go away, this is a pure Nix language overview and should not use stuff from `nixpkgs`.

fricklerhandwerk:
> Could it be possible to add details about type casting?

Out of scope for this one, it's just a migration from NixOS manual, with basic fixups and additions.

doc/manual/src/language/index.md
rapenne-s: I wasn't sure this could be possible, nevermind then :)

rapenne-s:
> > Could it be possible to add details about type casting?
> 
> Out of scope for this one, it's just a migration from NixOS manual, with basic fixups and additions.

ok, this may not be the right place, but this should be somewhere easy to find in the manual at least. For another PR.

rapenne-s (approved):
Looks good to me, thanks!

doc/manual/src/language/index.md
```diff
@@ -11,3 +11,551 @@ packages, compositions of packages, and the variability within packages.
...
+  <td>
+
+   A home path. Evaluates to the `"<user's home directory>/.config"`.
+
+  </td>
+ </tr>
+ <tr>
+  <td>
+
+   <nixpkgs>
```
edolstra, line 174: Should this be escaped in an HTML table, or does it just happen to work because of markdown's lack of a proper syntax?

doc/manual/src/language/index.md
edolstra: `pkgs.lib` is also from Nixpkgs though, as is the reference to `pkgs.bash` elsewhere. But I don't think it's a problem. If you want a pure Nix example, it could be something like:

```nix
with builtins; head (tail [1 2 3])
```

edolstra (approved):

doc/manual/src/language/index.md
fricklerhandwerk: Yes, it just happens to work. Everything between newlines is captured as markdown, so this is an indented code block.

doc/manual/src/language/index.md
```diff
@@ -403,7 +403,7 @@ This is an incomplete overview of language features, by example.
  <tr>
   <td>
 
-   `with pkgs.lib; head [ 1 2 3 ]`
+   `builtins.head [ 1 2 3 ]`
```
edolstra, line 406: Given that this is to illustrate `with`, I don't think this is intended?

rapenne-s (approved):

Issue #5612, opened 2021-11-20
`nix-instantiate --json` incorrectly uses store path


lapp0:
**Describe the bug**

Inconsistent output for`nix-instantiate` between use of `--json`, `--xml`, and no encoding argument. In the case of `--json` it uses the nix-store path.

**Steps To Reproduce**

Set `imports = [(modulesPath + "/installer/scan/not-detected.nix")];` in `my-file.nix`

Observe the output for
```
# no encoding
nix-instantiate --eval -E '(import my-file.nix {config = {}; pkgs = import <nixpkgs> {}; lib = {}; modulesPath = <nixpkgs/nixos/modules>;}).imports' --strict
[ /etc/nixos/nixpkgs/nixos/modules/installer/scan/not-detected.nix ]

# xml
nix-instantiate --xml --eval -E '(import my-file.nix {config = {}; pkgs = import <nixpkgs> {}; lib = {}; modulesPath = <nixpkgs/nixos/modules>;}).imports' --strict
<?xml version='1.0' encoding='utf-8'?>
<expr>
  <list>
    <path value="/etc/nixos/nixpkgs/nixos/modules/installer/scan/not-detected.nix" />
  </list>
</expr>

# json
nix-instantiate --json --eval -E '(import my-file.nix {config = {}; pkgs = import <nixpkgs> {}; lib = {}; modulesPath = <nixpkgs/nixos/modules>;}).imports' --strict
["/nix/store/5vc99s7pn69w6mnhh7rfnzcrdl8mf7dn-not-detected.nix"]%
```

**Expected behavior**

`--json` output should be `[ "/etc/nixos/nixpkgs/nixos/modules/installer/scan/not-detected.nix" ]`

**`nix-env --version` output**

`nix-env (Nix) 2.4pre20211006_53e4794`

milahu:
https://github.com/NixOS/nix/blob/9cd8cffefc65fe425c6aa07aa3bc67a8eec46087/src/libexpr/value-to-xml.cc#L79-L81

https://github.com/NixOS/nix/blob/9cd8cffefc65fe425c6aa07aa3bc67a8eec46087/src/libexpr/value-to-json.cc#L34-L36

... but i have no idea why json should print store paths

ncfavier:
Just ran into this, is there any reason the JSON printer behaves differently? I can't think of a reason to copy paths to the store when printing JSON.

ncfavier:
[...]

PR #6909, opened 2022-08-16
JSON: print paths as strings without copying them to the store

ncfavier:
Fixes https://github.com/NixOS/nix/issues/5612

### DIFF
--- a/src/libexpr/value-to-json.cc
+++ b/src/libexpr/value-to-json.cc
@@ -10,7 +10,7 @@
 namespace nix {
 
 void printValueAsJSON(EvalState & state, bool strict,
-    Value & v, const PosIdx pos, JSONPlaceholder & out, PathSet & context)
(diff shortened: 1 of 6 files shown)


edolstra:
This breaks `builtins.toJSON`, e.g. in `builtins.toJSON { foo = ./file; }`, it's essential that `./file` is copied to the Nix store and added to the context of the resulting string. Maybe for `nix eval --json`, `printValueAsJSON()` could take a flag that controls whether paths are copied to the store.

ncfavier:
Why should `toJSON` have this behaviour but not `toXML`?

ncfavier:
I've added a flag `copyToStore` to `printValueToJSON`, set to false in `nix eval --json`, `nix-instantiate --eval --json` and `nix-env --json`.

edolstra:
Not sure we should change `nix-env` / `nix-instantiate` behaviour. But it's fine to change `nix eval`.

ncfavier:
Done

Issue #5214, opened 2021-09-04
Disabling colors


teoljungberg:
**Is your feature request related to a problem? Please describe.**
I don't use much colors in my terminal environment, most tools adhere to the `NO_COLOR` environment variable or allows its users to disable coloring entirely if they choose to. I see Nix has support for it [1], but not spread across the entire toolchain.

This comes from https://no-color.org

[1]: https://github.com/nixos/nix/blob/90b2dd570cbd8313a8cf45b3cf66ddef2bb06e07/src/libutil/util.cc#L1377

**Describe the solution you'd like**
A configuration option to disable colors, or adhering to the NO_COLOR environment variable for all nix commands. Such as `nix repl`, `nix-shell`, or when a user enters a non-existing command to nix.

<img width="416" alt="image" src="https://user-images.githubusercontent.com/810650/132102737-cbee0930-dfaf-4ee3-bfcf-a00eaa41bfb4.png">

cole-h:
I believe https://github.com/NixOS/nix/pull/5090 implements this -- maybe give that a shot and report back on that PR?

teoljungberg:
Thanks for the link @cole-h! I'll follow up the discussion there.

antifob:
Why was this closed? Nothing came out of the PR and I'd love to see this implemented.

It makes some use cases impossible; such as simply capture `nix repl`'s output.

teoljungberg:
Re-opening this as #5090 was closed.

willbush:
![20240426_18h00m18s_grim](https://github.com/NixOS/nix/assets/2023546/42f92a1c-671a-460b-ac59-277dc136a74a)
This seems fixed when I try to repro it.

Looks similar to https://github.com/NixOS/nix/issues/9486, but I can repro that `NO_COLOR` doesn't work for `nix flake show`.

tomberek:
Closing as stale. Re-open if needed.

PR #6911, opened 2022-08-16
Make `nix log` respect NO_COLOR=1

knedlsepp:
Some parts of nix already respect NO_COLOR=1.
It seems however this functionality is desired, yet missing from `nix log`.
See: https://discourse.nixos.org/t/disabling-colors-for-nix-log-output/17480

### DIFF
--- a/src/nix/log.cc
+++ b/src/nix/log.cc
@@ -4,6 +4,7 @@
 #include "store-api.hh"
 #include "log-store.hh"
 #include "progress-bar.hh"
+#include "util.hh"
 
 using namespace nix;
 
@@ -53,6 +54,9 @@ struct CmdLog : InstallableCommand
             if (!log) continue;
             stopProgressBar();
             printInfo("got build log for '%s' from '%s'", installable->what(), logSub.getUri());
+            if (!shouldANSI()) {
+                log = filterANSIEscapes(*log, true);
+            }
             std::cout << *log;
             return;
         }


src/nix/log.cc
```diff
@@ -53,6 +54,9 @@ struct CmdLog : InstallableCommand
             if (!log) continue;
             stopProgressBar();
             printInfo("got build log for '%s' from '%s'", installable->what(), logSub.getUri());
+            if (!shouldANSI()){
```
edolstra, line 57:
```suggestion
            if (!shouldANSI()) {
```

knedlsepp:
> I wonder if there's a way to make the logger do this instead of having to litter around `shouldANSI` checks... I haven't looked at the Nix code in a while, but I imagine that would be better than the current situation.

How about wrapping `std::cout` in `nix::cout`?

knedlsepp:
> > I wonder if there's a way to make the logger do this instead of having to litter around `shouldANSI` checks... I haven't looked at the Nix code in a while, but I imagine that would be better than the current situation.
> 
> How about wrapping `std::cout` in `nix::cout`?

Hm. There is also `logger->cout`.

edolstra:
> Hm. There is also logger->cout.

Right, probably the cleanest solution is to make `logger->cout` strip the ANSI colors if needed.

sivizius:
I suggest to add a `--colour=<never|auto|always>` option, that defaults to `auto` which is the status quo. With `never`, `shouldANSI` must always return `false`, while with `always`, it must always return `true` ignoring `NO_COLOR` and whether the file descriptor is an actual terminal.
There should be an alias `--color` for this option (or the other way around `--colour` aliasing to `--color`). Perhaps an `--colour`/`--color` without an argument could be equal to `--colour=always`, but this might complicate argument parsing. 
The motivation behind `--colour=always` is to pipe the output to a pager like more or less. At the moment, because `isatty` returns `false` in this case, `shouldANSI` evaluates to `false` and any ANSI-colour-sequences are stripped.

n8henrie:
This would be really nice. For example, it's convenient to pipe `nix log` output to `gh gist create`, but as is (even with `NO_COLOR=1`) this creates a bunch of escape sequence mess.

Seems that one can pipe to a `sed -e $'s/\x1b\[[0-9;]*m//g'` intermediary for now ([source](https://superuser.com/a/380778)).

DavSanchez:
Any chance this gets some attention?

n8henrie:
This seems like a really small PR for a decent QoL win, particularly as we need to post logs for build errors that are currently littered with ansi escape codes unless people do manual workarounds.

knedlsepp:
> This seems like a really small PR for a decent QoL win, particularly as we need to post logs for build errors that are currently littered with ansi escape codes unless people do manual workarounds.

@n8henrie  I'm sorry I currently don't have the time to work in the suggested changes ( https://github.com/NixOS/nix/pull/6911#issuecomment-1222278160 ). But I agree that it might not be much work to do this properly. Maybe somebody else can give it a shot. :)
I'll close this for now.

drupol:
@knedlsepp I stumbled upon this issue myself today. Is there an issue open to keep track of the issue properly?

knedlsepp:
> @knedlsepp I stumbled upon this issue myself today. Is there an issue open to keep track of the issue properly?

Looks like https://github.com/NixOS/nix/issues/5214 would do.

PR #6913, opened 2022-08-17
lazy-trees cherrypicks

edolstra:
Some miscellaneous changes cherry-picked from the lazy-trees branch. See the individual commits.


tomberek:
Is [c3769c6](https://github.com/NixOS/nix/pull/6913/commits/c3769c68465bae971ab6bb48cfcdea85b61ea83a) still necessary? Removing the progress bar from `nix repl` solves the problem, but does remove it from the user experience. Can we bring back the progress bar?

PR #6652, opened 2022-06-11
doc: add install test info to hacking.md

abathur:
@Artturin [suggested](https://github.com/NixOS/nix/pull/6639#issuecomment-1151841848) _someone_ document installer testing (added in #4549, updated in at least #4577) in hacking.md.

I'm a little ambivalent about doing this because it underscores everything _else_ about generating and testing installers that isn't documented--and I don't grok the big picture here. I'm happy to field organization and phrasing nits/clarifications--but fair warning that I will be grumpy if there's pressure to turn this PR into a vehicle to fully document the status quo.

I did stub out a comment with what I can recall of the manual process in case it encourages someone else to flesh that part out, but I'm not sure if it'll cause any trouble in the manual-building pipeline.

PR #6916, opened 2022-08-17
Installer: Don't assume TMPDIR ends in a slash

edolstra:
Also create TMPDIR if it doesn't exist.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -349,6 +349,11 @@ _sudo() {
 }
 
 
+# Ensure that $TMPDIR exists if defined.
+if [[ -v TMPDIR ]]; then
+    mkdir -m 0700 -p "$TMPDIR"
+fi
+
 readonly SCRATCH=$(mktemp -d)
 finish_cleanup() {
     rm -rf "$SCRATCH"
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -349,7 +349,7 @@ _sudo() {
 }
 
 
-readonly SCRATCH=$(mktemp -d "${TMPDIR:-/tmp/}tmp.XXXXXXXXXX")
+readonly SCRATCH=$(mktemp -d)
 finish_cleanup() {
     rm -rf "$SCRATCH"
 }


scripts/install-multi-user.sh
```diff
@@ -349,7 +349,12 @@ _sudo() {
 }
 
 
-readonly SCRATCH=$(mktemp -d "${TMPDIR:-/tmp/}tmp.XXXXXXXXXX")
+# Ensure that $TMPDIR exists if defined.
+if [[ -n $TMPDIR ]]; then
+    mkdir -m 0700 -p "$TMPDIR"
```
abathur, line 354: I think this needs to either use an intermediate variable or use the same `${TMPDIR:-/tmp}` expansion to avoid this unbound variable error:

https://github.com/abathur/nix/runs/7883911756?check_suite_focus=true#step:4:90

edolstra: I've made a few changes: Use `mktemp -d` without any arguments, and use `[[ -v TMPDIR ]]`.

abathur: I've re-started CI with these in https://github.com/abathur/nix/actions/runs/2878089386

https://github.com/NixOS/nix/pull/6652 also documents what you need to set up the installer tests on your fork, should you want :)

grahamc:
It looks like dropping the explicit handling of TMPDIR from this is fine for macOS:


```
detsys@Mac-Studio ~ % env | grep  -i TMP 
TMPDIR=/var/folders/6t/7cxk8qlj6sd1dl6zvyv40c0r0000gn/T/

detsys@Mac-Studio ~ % export TMPDIR=/oeunoheu

detsys@Mac-Studio ~ % env
USER=detsys
LOGNAME=detsys
HOME=/Users/detsys
PATH=/Users/detsys/.nix-profile/bin:/nix/var/nix/profiles/default/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
SHELL=/bin/zsh
TERM=xterm-256color
SHLVL=1
PWD=/Users/detsys
OLDPWD=/Users/detsys
NIX_PROFILES=/nix/var/nix/profiles/default /Users/detsys/.nix-profile
NIX_SSL_CERT_FILE=/nix/var/nix/profiles/default/etc/ssl/certs/ca-bundle.crt
TMPDIR=/oeunoheu
_=/usr/bin/env

detsys@Mac-Studio ~ % mktemp -d              
/var/folders/6t/7cxk8qlj6sd1dl6zvyv40c0r0000gn/T/tmp.axJVgIzd
```

but on Linux, it does matter:

```
grahamc@hyperchicken:~/ > export TMPDIR=./foo   
grahamc@hyperchicken:~/ > mktemp -d          
mktemp: failed to create directory via template ‘./foo/tmp.XXXXXXXXXX’: No such file or directory
```

so I think this means if TMPDIR is set and the directory does not exist, we should create it?

grahamc:
In other words, I think this looks good to me. How about you @abathur?

scripts/install-multi-user.sh
abathur: Still failed (but new failure) at https://github.com/abathur/nix/runs/7889093146?check_suite_focus=true#step:4:90

abathur:
Still seeing a CI failure. Commented on the existing file thread. Getting dinner so I can't really poke at it atm.

scripts/install-multi-user.sh
grahamc: Good catch, Bash on macOS doesn't support -v:

```
bash-3.2$ if [[ -v OENUTHOEU ]]; then echo hi ; fi
bash: conditional binary operator expected
bash: syntax error near `OENUTHOEU'

bash-3.2$ if [[ -v $OENUTHOEU ]]; then echo hi ; fi
bash: conditional binary operator expected
bash: syntax error near `$OENUTHOEU'

bash-3.2$ help test
test: test [expr]
    Exits with a status of 0 (true) or 1 (false) depending on
    the evaluation of EXPR.  Expressions may be unary or binary.  Unary
    expressions are often used to examine the status of a file.  There
    are string operators as well, and numeric comparison operators.
    
    File operators:
    
        -a FILE        True if file exists.
        -b FILE        True if file is block special.
        -c FILE        True if file is character special.
        -d FILE        True if file is a directory.
        -e FILE        True if file exists.
        -f FILE        True if file exists and is a regular file.
        -g FILE        True if file is set-group-id.
        -h FILE        True if file is a symbolic link.
        -L FILE        True if file is a symbolic link.
        -k FILE        True if file has its `sticky' bit set.
        -p FILE        True if file is a named pipe.
        -r FILE        True if file is readable by you.
        -s FILE        True if file exists and is not empty.
        -S FILE        True if file is a socket.
        -t FD          True if FD is opened on a terminal.
        -u FILE        True if the file is set-user-id.
        -w FILE        True if the file is writable by you.
        -x FILE        True if the file is executable by you.
        -O FILE        True if the file is effectively owned by you.
        -G FILE        True if the file is effectively owned by your group.
        -N FILE        True if the file has been modified since it was last read.
    
      FILE1 -nt FILE2  True if file1 is newer than file2 (according to
                       modification date).
    
      FILE1 -ot FILE2  True if file1 is older than file2.
    
      FILE1 -ef FILE2  True if file1 is a hard link to file2.
    
    String operators:
    
        -z STRING      True if string is empty.
    
        -n STRING
        STRING         True if string is not empty.
    
        STRING1 = STRING2
                       True if the strings are equal.
        STRING1 != STRING2
                       True if the strings are not equal.
        STRING1 < STRING2
                       True if STRING1 sorts before STRING2 lexicographically.
        STRING1 > STRING2
                       True if STRING1 sorts after STRING2 lexicographically.
    
    Other operators:
    
        -o OPTION      True if the shell option OPTION is enabled.
        ! EXPR         True if expr is false.
        EXPR1 -a EXPR2 True if both expr1 AND expr2 are true.
        EXPR1 -o EXPR2 True if either expr1 OR expr2 is true.
    
        arg1 OP arg2   Arithmetic tests.  OP is one of -eq, -ne,
                       -lt, -le, -gt, or -ge.
    
    Arithmetic binary operators return true if ARG1 is equal, not-equal,
    less-than, less-than-or-equal, greater-than, or greater-than-or-equal
    than ARG2.
```

grahamc: Looks like we probably want -z with bash's default-if-unset syntax:

```
bash-3.2$ set -u
bash-3.2$ if [[ -z "${OENUTHOEU:-}" ]]; then echo hi ; fi
hi
```

abathur: Started a fresh CI job https://github.com/abathur/nix/actions/runs/2879326487

abathur: Advanced a line, at least: https://github.com/abathur/nix/runs/7890927130?check_suite_focus=true#step:4:91

I guess the unbound checks are parse-time. Since we're already checking that it isn't empty, I guess something like https://github.com/abathur/nix/commit/9250e20274dcc81055c78c288ecfe14c49728c35 is fine. 

New ci: https://github.com/abathur/nix/actions/runs/2879566885

abathur: Agh :)

`-v` -> `-z` inverted the condition (variable is set to string is empty). 

Running again with `-n`: https://github.com/abathur/nix/actions/runs/2879795805

abathur: macOS case still fell over at https://github.com/abathur/nix/runs/7892112794?check_suite_focus=true#step:4:90

It looks like this is a difference in the mkdir implementations, seemingly caused by the `-m` flag:

```console
$ mkdir -m 0700 -p "${TMPDIR:-}"

$ /bin/mkdir -p $TMPDIR

$ /bin/mkdir -m 0700 -p "${TMPDIR:-}"
mkdir: /var/folders/hg/pt5wsk1n54d1prjyg9pff5_r0000gn/T/: Operation not permitted
```

I guess the right next step is to look before we leap. Last swing for the night: https://github.com/abathur/nix/actions/runs/2880189989

abathur: (this one ran clean)

Issue #6681, opened 2022-06-17
a new setting to define a download speed limit


rapenne-s:
I'm using nixpkgs-unstable and I often upgrade to test latest versions, but I have a slow DSL connection with multiple users behind it. When I run an upgrade, I would like to limit nix download speed to avoid disrupting internet for everyone behind the router.

**Describe the solution you'd like**
This would be awesome to have a setting defining a maximum download speed.

**Describe alternatives you've considered**
I'm currently using `wondershaper` but this limits the network interface bandwidth so it's not really practical.

PR #6926, opened 2022-08-19
add a nix.conf option to set a download speed limit

rapenne-s:
Hello

This PR adds a new attribute to the options available in nix.conf to defines a download speed limit.
This is particularly useful for people like me behind a slow DSL line who regularly need to download gigabytes of packages with nix, but don't want to make the other LAN Internet users angry.

### DIFF
--- a/src/libstore/filetransfer.cc
+++ b/src/libstore/filetransfer.cc
@@ -308,6 +308,9 @@ struct curlFileTransfer : public FileTransfer
 
             curl_easy_setopt(req, CURLOPT_HTTPHEADER, requestHeaders);
 
+            if (settings.downloadSpeed.get() > 0)
+                curl_easy_setopt(req, CURLOPT_MAX_RECV_SPEED_LARGE, (curl_off_t) (settings.downloadSpeed.get() * 1024));
+
             if (request.head)
                 curl_easy_setopt(req, CURLOPT_NOBODY, 1);
 
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -746,6 +746,13 @@ public:
               /nix/store/xfghy8ixrhz3kyy6p724iv3cxji088dx-bash-4.4-p23`.
         )"};
 
+    Setting<unsigned int> downloadSpeed {
+        this, 0, "download-speed",
+        R"(
+          Specify the maxium transfer rate in kilobytes per second you want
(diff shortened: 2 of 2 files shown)


rapenne-s:
this would fix #6681

qbit:
This would help with my Starlink connection! :+1:

src/libstore/globals.hh
```diff
@@ -746,6 +746,13 @@ public:
               /nix/store/xfghy8ixrhz3kyy6p724iv3cxji088dx-bash-4.4-p23`.
         )"};
 
+    Setting<unsigned int> downloadSpeed {
+        this, 0, "download-speed",
+        R"(
+          Specify the maxium transfer rate in kilobytes per second you want
```
AndersonTorres, line 752: Typo

```suggestion
          Specify the maximum transfer rate in kilobytes per second you want
```

src/libstore/filetransfer.cc
```diff
@@ -308,6 +308,9 @@ struct curlFileTransfer : public FileTransfer
 
             curl_easy_setopt(req, CURLOPT_HTTPHEADER, requestHeaders);
 
+            if (settings.downloadSpeed.get() > 0)
+                curl_easy_setopt(req, CURLOPT_MAX_RECV_SPEED_LARGE, (curl_off_t) (settings.downloadSpeed.get() * 1024));
```
FRidh, line 312: If this is in bytes/second then a future improvement could be support for prefixes.

rapenne-s: indeed, this could be useful at some point. But I expect people needing this feature to have a low value less than ~10 MB/s so it still bearable. But this could be a further improvement after this PR.

src/libstore/globals.hh
```diff
@@ -746,6 +746,13 @@ public:
               /nix/store/xfghy8ixrhz3kyy6p724iv3cxji088dx-bash-4.4-p23`.
         )"};
 
+    Setting<unsigned int> downloadSpeed {
+        this, 0, "download-speed",
+        R"(
+          Specify the maximum transfer rate in kilobytes per second you want
+          nix to use for download.
```
edolstra, line 753:
```suggestion
          Nix to use for downloads.
```

edolstra (approved):

PR #6936, opened 2022-08-22
Search any attribute recursively if `recurseForDerivation` is set

### LINKED ISSUE #6938: Allow searching any attribute recursively

`nix search` currently only allows recursive searching through `legacyPackages`.
While most attributes are supposed to be flat, it is conceivable to collect packages out of the "official" ones.

I would like `nix search [flake]#myPackages` to recursively search through `myPackages` (given `recurseForDerivations` is set) rather than having to place `myPackages` into `legacyPackages`.

The solution proposed in #6936 removes the special case of `legacyPackages` being the only attribute that is searched recursively given `recurseForDerivations` is set.

Special cases for `packages` and `legacyPackages` to not need `recurseForDerivations` in the `system` attribute remain. 

-> always recurse two attributes deep?

ysndr:
closes #6938

### DIFF
--- a/src/nix/search.cc
+++ b/src/nix/search.cc
@@ -178,7 +178,7 @@ struct CmdSearch : InstallableCommand, MixJSON
                 else if (initialRecurse)
                     recurse();
 
-                else if (attrPathS[0] == "legacyPackages" && attrPath.size() > 2) {
+                else if (attrPath.size() >= 2) {
                     auto attr = cursor.maybeGetAttr(state->sRecurseForDerivations);
                     if (attr && attr->getBool())
                         recurse();


ysndr:
This change is specifically meant to enable you to do this expensive search on attributes other than `legacyPackages` and only if the attribute path matches.


A search on a flake like this that does not explicitly search `myCollection` (`nix search .#myCollection search-term`) does not incur any overhead.
Neither does the same search if there is no `legacyPackages.*.myCollection`.
That includes no attribute path at all as well; `nix search search-term` should remain searching `packages`/`legacyPackages` only.

```
{
   legacyPackages = { /* deep*/ };
   myCollection = { /* deep */ };
}
```

Please tell me if I'm wrong.

I see deep structures being a concern for `nix flake show` or similar things that would evaluate the entire flake and do not propose to change that here.
Yet `legacyPackages` is an exclusive place where derivations can be in a nested structure AND be searched for - I think this restriction could be loosened.

tomberek:
> ... it's very expensive to evaluate the `recurseForDerivation` attribute. It forces the evaluation of every derivation just to find out what the attribute names of packages are. ...

Perhaps this underlying problem can be addressed? Related? https://github.com/NixOS/nix/pull/4154

thufschmitt:
Discussed during the Nix team meeting.
This has too much of a performance impact right now to be acceptable. Another solution that doesn't require forcing everything might be acceptable (like having a `filedsToRecurseInto = [ "foo" "barr" ]` attribute in the parent). Alternatively, changing the way derivations are computed to make them more lazy (like what the nixpkgs modules WG is working on) might make this viable in the future in which case it would be worth re-discussing

PR #6946, opened 2022-08-24
Fix a misplaced parenthese in serve protocol check

rickynils:
This issue made it impossible for clients using a serve protocol of
version <= 2.3 to use the `cmdBuildDerivation` command of servers using
a protocol of version >= 2.6. The faulty version check makes the server
send back build outputs that the client is not expecting.


rickynils:
This issue exists in all Nix releases >= 2.4

edolstra:
Good catch, thanks!

PR #6953, opened 2022-08-24
Update install-systemd-multi-user.sh

### LINKED ISSUE #6952: Fix Typo In Multi User Install Script
https://github.com/NixOS/nix/blob/04e74f7c8bb5589dec578dd049013d3cd2554e65/scripts/install-systemd-multi-user.sh#L90

The symlink **should be referencing this** `default$TMPFILES_SRC` **_not this_** `default/$TMPFILES_SRC` based on how the variable was set:

https://github.com/NixOS/nix/blob/04e74f7c8bb5589dec578dd049013d3cd2554e65/scripts/install-systemd-multi-user.sh#L12

EzraSingh:
https://github.com/NixOS/nix/blob/04e74f7c8bb5589dec578dd049013d3cd2554e65/scripts/install-systemd-multi-user.sh#L90

The symlink **should be referencing this** `default$TMPFILES_SRC` **_not this_** `default/$TMPFILES_SRC` based on how the variable was set:

https://github.com/NixOS/nix/blob/04e74f7c8bb5589dec578dd049013d3cd2554e65/scripts/install-systemd-multi-user.sh#L12

---

See [Issue 6960](https://github.com/NixOS/nix/issues/6952)

Closes #6952

### DIFF
--- a/scripts/install-systemd-multi-user.sh
+++ b/scripts/install-systemd-multi-user.sh
@@ -87,7 +87,7 @@ poly_configure_nix_daemon_service() {
         task "Setting up the nix-daemon systemd service"
 
         _sudo "to create the nix-daemon tmpfiles config" \
-              ln -sfn /nix/var/nix/profiles/default$TMPFILES_SRC $TMPFILES_DEST
+              ln -sfn "/nix/var/nix/profiles/default$TMPFILES_SRC" "$TMPFILES_DEST"
 
         _sudo "to run systemd-tmpfiles once to pick that path up" \
              systemd-tmpfiles --create --prefix=/nix/var/nix
--- a/scripts/install-systemd-multi-user.sh
+++ b/scripts/install-systemd-multi-user.sh
@@ -87,7 +87,7 @@ poly_configure_nix_daemon_service() {
         task "Setting up the nix-daemon systemd service"
 
         _sudo "to create the nix-daemon tmpfiles config" \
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Please add a ``Closes #6952`` to close the issue when this is merged. Also you don't need to create an issue if you open a PR right after it.

Also as far as I can tell this is just looks and does not effect any function, right?

fricklerhandwerk:
@EzraSingh The amount of slashes in paths does not matter. `ls /usr/bin` is equivalent to `ls /////usr///bin` The additional slash is for readability.

Paiusco:
> @EzraSingh The amount of slashes in paths does not matter. `ls /usr/bin` is equivalent to `ls /////usr///bin`. The additional slash is for readability.

I agree with you here and think that readability is more important however, he does have a point. At `systemctl link "/nix/var/nix/profiles/default$SERVICE_SRC"` (and more) we follow a different pattern. If not this PR, a new one adding the trailing slashes would be nicer.

fricklerhandwerk:
@EzraSingh do you want to follow @Paiusco's proposal and make it consistent, or should we close this PR?

EzraSingh:
> @EzraSingh do you want to follow @Paiusco's proposal and make it consistent, or should we close this PR?

I added the additional quotes to make it consistent

EzraSingh:
@edolstra bump

fricklerhandwerk (approved):

PR #6962, opened 2022-08-25
Make 'logger' a std::unique_ptr

edolstra:
This prevents it from being leaked (see bb411e4ae16d6a5c61ea595c0c12e2ecee081ff9 for an example of this).

This also fixes a bug in the ProgressBar destructor when isTTY = false.

Depends on #12483.

### DIFF
--- a/src/libexpr-tests/primops.cc
+++ b/src/libexpr-tests/primops.cc
@@ -28,20 +28,15 @@ namespace nix {
     };
 
     class CaptureLogging {
-        Logger * oldLogger;
-        std::unique_ptr<CaptureLogger> tempLogger;
+        std::unique_ptr<Logger> oldLogger;
         public:
-            CaptureLogging() : tempLogger(std::make_unique<CaptureLogger>()) {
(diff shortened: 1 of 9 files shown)


src/libexpr/tests/primops.cc
```diff
@@ -110,7 +105,7 @@ namespace nix {
         CaptureLogging l;
         auto v = eval("builtins.trace \"test string 123\" 123");
         ASSERT_THAT(v, IsIntEq(123));
-        auto text = l.get();
+        auto text = ((CaptureLogger *) logger.get())->get();
```
Ericson2314, line 108: What's going on here?

Ericson2314 (approved):

fricklerhandwerk:

- @thufschmitt: attempted this multiple times but CI failed for various reasons
    - the bug it would fix is insignificant, not worth the efford right now
- postponed

fricklerhandwerk:
@edolstra do you want to take another look at it?

PR #6964, opened 2022-08-25
pathExists: Do not throw exception if lstat returns EPERM

centromere:
Nix fails to build in a sandboxed MacOS environment:

```
[ RUN      ] pathExists.bogusPathDoesNotExist
unknown file: Failure
C++ exception with description "error: getting status of /home/schnitzel/darmstadt/pommes: Operation not permitted" thrown in the test body.
[  FAILED  ] pathExists.bogusPathDoesNotExist (0 ms)
[----------] 3 tests from pathExists (0 ms total)
```

```
libc++abi: terminating with uncaught exception of type nix::SysError: error: getting status of /nix/var/nix/profiles/per-user/root/channels/nixpkgs: Operation not permitted
make: *** [mk/lib.mk:118: libexpr-tests_RUN] Abort trap: 6
```

This change ensures that an exception is not thrown if a file is unable to be accessed because of its permissions.

### DIFF
--- a/src/libutil/util.cc
+++ b/src/libutil/util.cc
@@ -252,7 +252,7 @@ bool pathExists(const Path & path)
     struct stat st;
     res = lstat(path.c_str(), &st);
     if (!res) return true;
-    if (errno != ENOENT && errno != ENOTDIR)
+    if (errno != ENOENT && errno != ENOTDIR && errno != EPERM)
         throw SysError("getting status of %1%", path);
     return false;
 }


centromere:
@thufschmitt The unit test can certainly be re-written, but what about this exception?

```
libc++abi: terminating with uncaught exception of type nix::SysError: error: getting status of /nix/var/nix/profiles/per-user/root/channels/nixpkgs: Operation not permitted
make: *** [mk/lib.mk:118: libexpr-tests_RUN] Abort trap: 6
```

thufschmitt:
Indeed you're right, the exception as it is reported is pretty bad. It 
should be transformed to a nice evaluation  error

edolstra:
I'm reluctant to change the semantics of `pathExists()`, since it's used in so many places. "Path existing but not being accessible" is not quite the same thing as "path not existing"...

centromere:
The exceptions above were preventing me from compiling on OSX. I wanted a quick fix (which I have achieved), so my intent is to put the issue on the radar rather than advocate for any particular solution.

Happy to close the PR in favor of a proper solution.

PR #6969, opened 2022-08-27
refactor rendering command documentation to markdown

fricklerhandwerk:
idea:
- make document structure visible, like in a template
- separate processing from formatting (structure/content vs. presentation)
- order functions by descending abstraction
- use self-explanatory names
- avoid broadly scoped or deeply nested let bindings

this allows readers to enter the code starting with what is visible from
the outside, instead of working themselves up from purely technical
details.

### DIFF
--- a/doc/manual/generate-manpage.nix
+++ b/doc/manual/generate-manpage.nix
@@ -8,12 +8,18 @@ let
   showCommand =
     { command, def, filename }:
     ''
-      **Warning**: This program is **experimental** and its interface is subject to change.
+      > **Warning**
+      > This program is **experimental** and its interface is subject to change.
+
+      # Name
+
+      `${command}` - ${def.description}
+
+      # Synopsis
+
+      ${showSynopsis { inherit command; args = def.args; }}
+
     ''
-    + "# Name\n\n"
-    + "`${command}` - ${def.description}\n\n"
-    + "# Synopsis\n\n"
-    + showSynopsis { inherit command; args = def.args; }
     + (if def.commands or {} != {}
        then
          let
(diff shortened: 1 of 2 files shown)


rapenne-s (approved):
looks good to me, generated documentation is iso with the current one

doc/manual/generate-manpage.nix
```diff
@@ -1,99 +1,110 @@
...
+  manpages = processCommand {
+    command = "nix";
+    details = builtins.fromJSON command;
+    filename = "nix";
+  };
+
+  tableOfContents = let
+    showEntry = page:
+      "    - [${page.command}](command-ref/new-cli/${page.name})";
+    in concatStringsSep "\n" (map showEntry manpages);
```
thufschmitt, line 108: This toc used to have an extra newline at the end (which it doesn't have anymore), meaning that the toplevel toc (`doc/manual/src/SUMMARY.md`, generated from `doc/manual/src/SUMMARY.md.in`) is missing a newline too: it contains

```
    - [nix why-depends](command-ref/new-cli/nix3-why-depends.md)  - [Files](command-ref/files.md)
    - [nix.conf](command-ref/conf-file.md)
```

Instead of

```
    - [nix why-depends](command-ref/new-cli/nix3-why-depends.md)
  - [Files](command-ref/files.md)
    - [nix.conf](command-ref/conf-file.md)
```

which in turn means that the `Files` section is ignored and the `conf-file` docs end-up directly in the `Experimental-commands section`.

thufschmitt (changes requested):
Thanks, the result is already quite nicer :)

I noticed just one small breakage, but everything seems correct otherwise. And removing the links from the manages is :1st_place_medal: 

(Also, it's very minor but having the “experimental feature” warning displayed as a quote makes it much more obvious :+1: )

thufschmitt (approved):

PR #6973, opened 2022-08-29
Only push Docker image when Docker secrets are set

Enzime:
When following the instructions in #6652 I ran into the issue that when Docker secrets aren't set, the `docker_push_image` step would fail and cause the workflow to fail and cancel the `installer_tests` steps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ permissions: read-all
 jobs:
 
   tests:
-    needs: [check_cachix]
+    needs: [check_secrets]
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


abathur:
A few notes (as a non-decider):
- At first I wondered why this was necessary because I haven't had this problem. After mulling this for a bit I imagine it's because you've tried to commit on your master branch, since that's the only place the docker flow is set to run: https://github.com/NixOS/nix/blob/ddb82ffda993d237d62d59578f7808a9d98c77fe/.github/workflows/ci.yml#L82-L87

- It does indeed look like your failed job was a merge to master: https://github.com/Enzime/nix/actions/runs/2942997871.
- The fix itself makes sense on a quick read. There's also a clean test run in the @enzime's repo: https://github.com/Enzime/nix/actions/runs/2949114988

    Merging to master in forks might be a bit of a footgun, but I guess there may be great reasons to do and I feel like the change makes things more explicit in a way that probably makes it less likely that mistakes are made in these Nix workflows going forward?

garbas (approved):

PR #6974, opened 2022-08-30
nix develop: use proper permissions for $TMPDIR

winterqt:
Before this change, unsandboxed builds that tried to use `passAsFile` would fail when run within a devshell.

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -298,7 +298,7 @@ struct Common : InstallableCommand, MixProfile
         for (auto & var : savedVars)
             out << fmt("%s=\"$%s:$nix_saved_%s\"\n", var, var, var);
 
-        out << "export NIX_BUILD_TOP=\"$(mktemp -d -t nix-shell.XXXXXX)\"\n";
+        out << "export NIX_BUILD_TOP=\"$(mktemp -d -t nix-shell.XXXXXX)\"\nchmod 1777 \"$NIX_BUILD_TOP\"\n";
         for (auto & i : {"TMP", "TMPDIR", "TEMP", "TEMPDIR"})
             out << fmt("export %s=\"$NIX_BUILD_TOP\"\n", i);
 


src/nix/develop.cc
```diff
@@ -298,7 +298,7 @@ struct Common : InstallableCommand, MixProfile
         for (auto & var : savedVars)
             out << fmt("%s=\"$%s:$nix_saved_%s\"\n", var, var, var);
 
-        out << "export NIX_BUILD_TOP=\"$(mktemp -d -t nix-shell.XXXXXX)\"\n";
+        out << "export NIX_BUILD_TOP=\"$(mktemp -d -t nix-shell.XXXXXX)\"\nchmod 777 \"$NIX_BUILD_TOP\"\n";
```
SuperSandro2000, line 301: You are really sure about 777? I would expect something more like 755.

winterqt: `/tmp` on Linux is usually 777, that's what I went off of.

bjornfor: Isn't the common perms for /tmp `1777` (note the sticky bit)? (At least it is on my systems.)

winterqt: You're right, I thought the sticky bit signified that it was a directory for whatever reason. Fixed.

edolstra:
I don't understand how this is supposed to fix `passAsFile`. The problem is that the environment variables recorded in the `getDerivationEnvironment()` refer to the sandbox, which doesn't exist anymore when we run the shell, e.g.
```
# nix develop
# echo $fooPath 
/build/.attr-1bp7cri8hplaz6hbz0v4f0nl44rl84q1sg25kgwqzipzd1mv89ic
```
Creating a world-writable directory definitely seems like something to be avoided.

Probably `makeRcScript()` should just process the `passAsFile` attributes from the original derivation and write them somewhere. The problem is that it's not entirely obvious where it should write them. E.g. it could write them to a temporary file that gets deleted when `nix develop` exits, but that wouldn't work for `nix print-dev-env`, and it might be annoying if the `passAsFile` paths change on every invocation. Or maybe it should just copy them to the Nix store...

winterqt:
> I don't understand how this is supposed to fix `passAsFile`.

This fixes `passAsFile` from (unsandboxed) builds run *within* a devshell, i.e.
```console
$ nix develop ...
$ nix build ...
```

fricklerhandwerk:

We couldn't figure out what problem this is solving. @winterqt will pick this up again if you add a test that reproduces the bug. Postponed for now.

- @edolstra: when you use `pass-as-file` it doesn't work in `nix develop` because those env vars will refer to files in the sandbox in `useDerviationEnvironment` which don't exist any more
    - didn't understand what the change is supposed to do though
    - especially how the restricted permissions break the build
- @Ericson2314: maybe the author should write a test and then we figure out what to do about it
- @thufschmitt: why do we set `NIX_BUILD_TOP` anyway?
    - @edolstra: to simulate a regular build as closely as possibler
- @edolstra: it's not even clear what exactly this fixes
- postponed, waiting for author to provide a reproducer

</details>

tomberek:

PR #6977, opened 2022-08-31
Try to fix boehmgc coroutine segfault

matthewbauer:
Trying to fix Issue https://github.com/NixOS/nix/issues/4919

Unsure if this completely fixes the issue.

Based off of this blogpost from the crystal project:

https://crystal-lang.org/2022/02/16/bdw-gc-coroutines-support.html

Sets GC_stackbottom when allocating coroutine.
Note this only works for single-threaded, which I think evaluation still is.

Anyway, I think I have a small patch that appears to fix the issue for me. Tested by doing:

```
rm -rf ~/.cache/nix && GC_INITIAL_HEAP_SIZE=1048576 ../nix/outputs/out/bin/nix develop -vvvvv
```

on a particularly large proprietary repo. Setting GC_INITIAL_HEAP_SIZE  to force a GC. Would be interested in anyone confirming this fixes their segmentation fault issues.

cc @edolstra @roberth


roberth:
`GC_stackbottom` must point to the correct stack whenever GC runs, but you're only modifying it around coroutine creation and destruction. We _should_ set `GC_stackbottom` in every yield.

Also all stacks must be marked (scanned). By changing `GC_stackbottom`, it seems that you stop the marking of the original stack. It's quite possible that your evaluation has read some garbage that was still intact.

The crystal-lang blog post describes a solution to the latter problem

> Addressing the bdw-gc set up part, we use GC_set_push_other_roots to hook before the GC attempts a collect. In this procedure we push all the stacks of the coroutines that are not the current one, i.e.: that are not running.

The switching logic should be removed from the allocation, and moved to the `yield` calls.
The `GC_add_roots` and `GC_remove_roots` should be removed; instead, we'll need to add part of the bookkeeping for our `GC_set_push_other_roots` implementation there.

roberth:
For a quick fix, it seems quite possible to port the patch to the analogous darwin file as well.

The patch isn't that problematic anyway. We're retaining the garbage part of the stack region when GCing during a coroutine, but this is kind of ok for a conservative GC like this anyway, and it only happens while a coroutine exists, and the coroutines are short-lived. (iirc only during source filtering, last time I checked)

matthewbauer:
> `GC_stackbottom` must point to the correct stack whenever GC runs, but you're only modifying it around coroutine creation and destruction. We _should_ set `GC_stackbottom` in every yield.

Thanks! I will try to do this.

> Also all stacks must be marked (scanned). By changing `GC_stackbottom`, it seems that you stop the marking of the original stack. It's quite possible that your evaluation has read some garbage that was still intact.

Hmmm... Yeah that's quite possible. I was thinking GC_add_roots would cover the existing roots.

> The crystal-lang blog post describes a solution to the latter problem
> 
> > Addressing the bdw-gc set up part, we use GC_set_push_other_roots to hook before the GC attempts a collect. In this procedure we push all the stacks of the coroutines that are not the current one, i.e.: that are not running.
> 
> The switching logic should be removed from the allocation, and moved to the `yield` calls. The `GC_add_roots` and `GC_remove_roots` should be removed; instead, we'll need to add part of the bookkeeping for our `GC_set_push_other_roots` implementation there.

Thanks! I will try to do this.

> For a quick fix, it seems quite possible to port the patch to the analogous darwin file as well.

So from what I can tell, that patch should be used on darwin as well as linux. `darwin_stop_world.c` is not used at all for the nixpkgs version of boehmgc on aarch64-darwin. Configure logs say:

```
checking for thread model used by GCC... posix
```

I will try to play around with the patch to see if maybe there's some assumption it makes that doesn't apply on macOS / darwin.

Perhaps I'm being naive, but this still doesn't seem that hard to get right... It would be really nice to get this right and I suspect we'll keep running into issues like missing platform support, having to keep the nixpkgs patch in sync, and making sure no one uses their own boehmgc dependency especially when building from source. Since it sounds like it's possible to do this correctly using bdwgc APIs, I think I'd like to continue working on this.

matthewbauer:
> So from what I can tell, that patch should be used on darwin as well as linux. `darwin_stop_world.c` is not used at all for the nixpkgs version of boehmgc on aarch64-darwin. Configure logs say:

Oh nevermind on that, it is being used, I was just patching something in an ifdef that never got included.

PR for this: https://github.com/NixOS/nix/pull/6987

Issue #6985, opened 2022-09-01
`nix flake show` should be filterable by the current system


NobbZ:
**Is your feature request related to a problem? Please describe.**

Often when one runs `nix flake show` nix tries to evaluate every well known output, though some of them depend on the system, and even worse, some of them use IFD, so to nix being able to show the information it needs to build for a foreign system.

This leads to errors like this:

```text
$ nix flake show github:nix-community/nix-doom-emacs
github:nix-community/nix-doom-emacs/ae22b4a3fe31ae31b3e8b415889f8c2c5a77d8dc
├───checks
│   ├───aarch64-darwin
error: a 'aarch64-darwin' with features {} is required to build '/nix/store/a5dhhm5nhn936q5177nlv3fn3b4jvqly-source.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
(use '--show-trace' to show detailed location information)
```

**Describe the solution you'd like**

`nix flake show` becomes system aware and does not try to evaluate foreign systems, but instead just shows an ellipsis as child for foreign systems. A fictive example could look like this:

```
$ nix flake show github:nobbz/nixos-config
github:nobbz/nixos-config/e460e5272935ea1e902c04e9c705e8dc938d6383
└───checks
    ├───aarch64-darwin
    │   └───... omitted (use '--foreign-systems' to show)
    └───x86_64-linux
        ├───alejandra: derivation 'alejandra-run-e460e5272935ea1e902c04e9c705e8dc938d6383'
        └───statix: derivation 'statix-run-e460e5272935ea1e902c04e9c705e8dc938d6383'
```

An extra flag could be introduced to "try anway" similar to how we use `--legacy` to get the full list of packages in `legacyPackages`, perhaps `--foreign-systems` as in the example above.

**Describe alternatives you've considered**

Fully forbid any kind of IFD in flake contexts.

**Additional context**

Actually I would totally prefer the alternative, though I see that the community isn't there yet and instead we need to raise awareness about the issue.

Kha:
> Actually I would totally prefer the alternative, though I see that the community isn't there yet and instead we need to raise awareness about the issue.

[...]

Issue #6454, opened 2022-04-27
Add an attribute that can be used to label custom attributes in `nix flake show`


NobbZ:
**Is your feature request related to a problem? Please describe.**

In a common flake there are some custom attributes used, that appear in `nix flake show` just as "unknown". One of the commonly used custom attributes is `lib` (used by `nixpkgs` itself). Also `homeConfigurations` and `homeModules` are rather common since `home-manager` got flake support. And I am sure there are equivalent attributes for `nix-darwin`.

**Describe the solution you'd like**

There should be a way to label these. This could be done by adding an extra attribute `labelers` to the flake schema.

A `labeler` then is a function that returns either a string or an arbitrary nested set of strings. `nix flake show` will use each attribute name as a "node" in the tree and increase nesting, while a string value will be used as label. As far as I can tell, this would pretty much match what one sees in the `nix flake show --json` output.

Whenever `nix flake show` discovers an unknown attribute `foo` it will call `self.labelers.foo self.foo`. The labeler then has to inspect the value and build the aforementioned label tree.

The following `show` snippet could have been created by the returned nix value below it:

```
├───homeConfigurations
│   ├───nobbz@enceladeus: Home Manager configuration
│   ├───nobbz@mimas: Home Manager configuration
│   └───nobbz@thetys: Home Manager configuration
```

```nix
{
  "nobbz@enceladeus" = "Home Manager configuration";
  "nobbz@mimas" = "Home Manager configuration";
  "nobbz@thetys" = "Home Manager configuration";
}
```

**Describe alternatives you've considered**

External programs that are required to increase discoverability of custom attributes

fricklerhandwerk:
Discussed in Nix team meeting on 2022-11-25:
- @edolstra: not in favor of adding ad hoc hacks like this
    - should have something like a module system instead, describing properties of flakes
        - this is of course a long-term thing
    - better not pollute flakes with ad-hoc stop-gaps, that would prevent solid long-term solutions

tomberek:
Closing as stale. Re-open if needed.

Issue #6453, opened 2022-04-27
Extend `nix flake check` to support custom attributes


NobbZ:
**Is your feature request related to a problem? Please describe.**

When running `nix flake check` one regularly sees warnings emitted by nix about unknown attributes, or `default*` moved to something else.

Especially the unknown attributes can be annoying, even more those that evolved into a quasi-standard, like `lib` (even used by `nixpkgs`) or `homeModules` and `homeConfigurations`.

**Describe the solution you'd like**

There might be yet another flake output which is not a derivation but instead a function that takes and arbitrary value and returns a list of string or set of string.

I initially wanted to use `checks` for this, though that is already used for something else. In lack of a better name, I therefore use `nixChecks` in the following examples.

Once `nix flake check` encounters an attribute `foo` it doesn't know how to check, it calls `self.nixChecks.foo self.foo`.

`nixChecks` can then use arbitrary nix functions to verify the value of `foo` to match the requirements and issue informative messages, warnings or errors. It should be avoided to use `trace`ing functions from the `nixChecks` though it lies in the nature of the checks, that they require at least partial evaluation of the values, and therefore might trigger `trace`ing calls within the value.

A message is either a set with `level` and `message` (where `level` is `"info"`, `"warning"`, or `"error"`.) or alternaively just a string, which is assumed to be the message for a warning then.

Whenever at least one error got emitted, the exit code of `nix flake check` should be non-zero.

**Describe alternatives you've considered**

External programs that would check validity of custom attributes while continue to ignore warnings issued by `nix flake check`, or not run it at all.

tomberek:
There are several similar situations where one would like the behavior of a well-known output path to be applied to another one or extended.

random brainstorm:
```
app2.behavesLike = "apps";
packages.behavesLike = "legacyPackages";
hydraJobs_feature_branch.behavesLike = "hydraJobs";
foo.behavesLike = {
   check = self: assert (1 == 1) true;
   build = self: packages;
};
```
I can see this quickly getting out of hand.

cyntheticfox:
Maybe the ability to disable the warning via a config setting like `warn-unknown-outputs` you could set in a flake's `nixConfig` might help?

PR #6988, opened 2022-09-01
nix flake show: don't evaluate derivations for foreign systems by default

max-privatevoid:
Simple implementation of the idea, based on how `legacyPackages` is handled already. Doesn't cover `hydraJobs`, unfortunately.

Fixes https://github.com/NixOS/nix/issues/6985


src/nix/flake.cc
```diff
@@ -957,6 +957,7 @@ struct CmdFlakeArchive : FlakeCommand, MixJSON, MixDryRun
 struct CmdFlakeShow : FlakeCommand, MixJSON
 {
     bool showLegacy = false;
+    bool showForeign = false;
```
thufschmitt, line 960: I'm not keen on is having `showForeign` false by default, because it would be bad for backwards-compatibility – esp. for the json output which is designed to be used in scripts and is more dangerous to break.

thufschmitt (commented):
I like this idea, although it conflicts a bit with the idea that the `system` hierarchy layer should be special-cased as least as can be. But probably that's worth it.

src/nix/flake.cc
NobbZ: I totally understand your sentiments, though having it defaulting to `true` is a major annoyance for the human.

1. Evaluation of `nix flake show` will take much longer than needed
2. Evaluation of `nix flake show` will show a lot of redundant information
3. Evaluation of `nix flake show` will fail for a majority of flakes

We can avoid all of these 3 problems by defaulting it to `false`.

At the same time, breaking the interface is fine, as long as we do it **now**. After 3.0 has been released we can not do it anymore.

Back when 2.4 got released, no one cared that it broke a lot of scripts. Back then it has been pointed out that its at your own risk to use an experimental API. So… If someone complaints, why did they use an experimental API if they do not want things to change?

Also, it is much easier to add `--foreign` to a handful of scripts, but annoying to have to remember to add it to each and every invocation to shave off **minutes** of eval time and annoying IFD errors…

I'd even suggest to make this a `nix.conf` option, such that every **user** can decide on their very own default in the systems or users `nix.conf`.

NobbZ:
> I like this idea, although it conflicts a bit with the idea that the system hierarchy layer should be special-cased as least as can be. But probably that's worth it.

I think we can easily agree on the fact that the way the system is currently handled has a lot of flaws, and doing special handling it will be the only way around it.

NobbZ:
And I also want to add:

Specialcasing the `system` within `nix`' source code wasn't necessary that mouch if we were doing checks and shows in nix expression language, as briefly suggested in #6453 and #6454.

Of course I see, that we would still need to tell the nix expressions what the current system is.

src/nix/flake.cc
edolstra: This should definitely not be a `nix.conf` option, as tempting as that is, since it's a bad idea for scripting to have options that fundamentally alter what the output of a command is. You can end up with a script that unexpectedly fails on another machine or user account because `foreign` is set to a different value...

yajo:
Or maybe `--systems=all` (default value for flag: `--systems=local`). That would let us also give a specific list of systems to eval (`--systems=x86_64-linux,aarch64-linux`).

L-as:
Is there any reason why we need this instead of just catching errors that happen due to not being able to build a drv and omitting those parts of the output?

yajo:
For `nix flake show`, probably it would be the same. However, for `nix flake check` we would need a way to check the flake only for the current arch. Otherwise an IFS would always become an error, even if it's not.

domenkozar:
Lost ~hour today due to this unexpected behavor while working on https://github.com/cachix/cachix/pull/476

adrian-gierakowski:
What is stopping this fix from being merged?

adrian-gierakowski:
What is stopping this fix from being merged?

thufschmitt:
> What is stopping this fix from being merged?

Nothing really.

@max-privatevoid I've merged master back into this and added a few tests. Let me know whether that looks good to you

thufschmitt:
Oh also: I've noticed that this doesn't apply to `legacyPackages`. Is it intentional?

(I don't think it should be a blocker, and I think it makes sense either way, just wants to make sure it's not an overlook)

max-privatevoid:
Thanks for writing the tests. I've added the functionality to `legacyPackages` as well now. I definitely meant to have the flag work there too, just forgot about it.

thufschmitt (approved):

thufschmitt:
Awesome, thanks. I've just pushed a fix for the tests because they were broken on non-x86_64-linux. If the CI is OK, it'll just merge

peterbecich:
Thanks for this PR, I am trying to use it to implement the same behavior for `nix flake check`: https://github.com/NixOS/nix/pull/7759

PR #6990, opened 2022-09-02
Don’t add a space after attrs completion in zsh

matthewbauer:
This matches the behavior of bash. We don’t want to add a space after
completion on attrs. Uses -S.

Switches to new compadd style comppletions instead of _describe.
Shouldn’t have any negative issues from what I can tell.

### DIFF
--- a/misc/zsh/completion.zsh
+++ b/misc/zsh/completion.zsh
@@ -10,14 +10,15 @@ function _nix() {
   local -a suggestions
   declare -a suggestions
   for suggestion in ${res:1}; do
-    # FIXME: This doesn't work properly if the suggestion word contains a `:`
-    # itself
-    suggestions+="${suggestion/	/:}"
+    suggestions+=("${suggestion%%	*}")
   done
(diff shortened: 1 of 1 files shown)


misc/zsh/completion.zsh
```diff
@@ -10,14 +10,15 @@ function _nix() {
...
+    args+=('-f')
+  elif [[ "$tpe" == attrs ]]; then
+    args+=('-S' '')
   fi
-  _describe 'nix' suggestions
+  compadd -J nix "${args[@]}" -a suggestions
 }
 
-_nix "$@"
+# _nix "$@"
```
knedlsepp, line 24: Happy little accident here?

matthewbauer:
```suggestion
_nix "$@"
```

matthewbauer: Oh yeah, forgot about that. That's the only way I could figure out to test the file with sourcing.

PR #6991, opened 2022-09-02
Sanity Manual Compilation

sgn:
This PR consists of 2 changes, the first change allows building manual this all target with touching anything outside build-directory. The second change adds support for cross compilation via qemu.

Current code base will always install nix into `$(prefix)` when `make` which is undesirable, since it may override the `nix` (assumed-to-be-good) installed in `$(prefix)` by the under-development `nix` (could be buggy) unconditionally.

The first change aimed to fix it, the second change is a bonus change that depends on the first change.

### DIFF
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -24,7 +24,13 @@ dummy-env = env -i \
 	NIX_STATE_DIR=/dummy \
 	NIX_CONFIG='cores = 0'
 
-nix-eval = $(dummy-env) $(nix_DIR)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
+ifneq (,$(findstring qemu,$(CROSS_EMULATOR)))
+  _nix-emu := QEMU_LD_PREFIX=${QEMU_LD_PREFIX} $(CROSS_EMULATOR)
+else
+  _nix-emu := $(CROSS_EMULATOR)
+endif
+
+nix-eval = $(dummy-env) $(_nix-emu) $(nix_DIR)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
 
 $(d)/%.1: $(d)/src/command-ref/%.md
 	@printf "Title: %s\n\n" "$$(basename $@ .1)" > $^.tmp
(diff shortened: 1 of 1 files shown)


sgn:
Downstream patches from https://github.com/void-linux/void-packages/pull/39050

doc/manual/local.mk
```diff
@@ -24,7 +24,13 @@ dummy-env = env -i \
 	NIX_STATE_DIR=/dummy \
 	NIX_CONFIG='cores = 0'
 
-nix-eval = $(dummy-env) $(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
+ifneq (,$(findstring qemu,$(CROSS_EMULATOR)))
+  _nix-emu := QEMU_LD_PREFIX=${QEMU_LD_PREFIX} $(CROSS_EMULATOR)
```
matthewbauer, line 28: if QEMU_LD_PREFIX is exported, shouldn't it just pass it through to CROSS_EMULATOR?

doc/manual/local.mk
```diff
@@ -24,7 +24,13 @@ dummy-env = env -i \
...
+else
+  _nix-emu := $(CROSS_EMULATOR)
+endif
+
+nix-eval = $(dummy-env) $(_nix-emu) $(nix_DIR)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
```
matthewbauer, line 33: Annoying that we have to do this, but makes sense for cross compilation.

doc/manual/local.mk
sgn: It will be unset if qemu was called with dummy-env.

matthewbauer: ah makes sense

thufschmitt (commented):
The qemu bits is very nice :)

The use of the locally built nix is more annoying because of an idiosyncrasy of the build system: `$(nix_DIR)/nix` and `$(bindir)/nix` are two separate executables (linked with different flags), and `make install` currently never builds `$(nix_DIR)/nix`. This change would make it necessary to build it too, which is annoying since the linking phase is far from cheap.

Ideally I'd really like to get rid of this weirdness and have `make install` just copy (and `patchelf` if needs be) the result of `make`, but that's another topic

sgn:
> I would rather get rid of the non-`make install` build to simplify things. IIRC simply copying the binary doesn't work because the installed binary needs a different RPATH.

But that would make life harder for those people who want to use a stable nix by distros channel.

thufschmitt:
> IIRC simply copying the binary doesn't work because the installed binary needs a different RPATH.

I'm no expert in the field, but isn't that a fairly common and mostly solved problem, either by using relative rpaths or patching the rpath at install time?

eli-schwartz:
The first commit is just #5145 and again #6015


Previous efforts that actually used a different tactic include:
#5185
#6005

Issues include:
#5140
#5184
#5781

eli-schwartz:
The actual fix will come as a side effect of #3160 since Meson internally handles both rpath and LD_LIBRARY_PATH and guarantees that build executables can be run by other build rules.

Ericson2314:
The first commit is handled by https://github.com/NixOS/nix/pull/5145, you should rebase on top for it.

PR #6993, opened 2022-09-02
Don’t readDerivation if impure derivations feature is disabled

matthewbauer:
readDerivation is pretty slow, and while it may not be significant for
some use cases, on things like ghc-nix where we have thousands of
derivations it really slows things down.

So, this just doesn’t do the impure derivation check if the impure
derivation experimental feature is disabled. Perhaps we could cache
the result of isPure() and keep the check, but this is a quick fix
of the slowdown introduced with impure derivations feature in 2.8.0.


thufschmitt (approved):

thufschmitt:
Note that this is just a stopgap fix as we should have it also fixed if the xp feature is enabled

Issue #6823, opened 2022-07-20
error: renaming '/nix/store/add-NUMBER/x' to '/nix/store/HASH-NAME'


bjornfor:
**Describe the bug**

I *think* this happens when auto-GC cannot free enough space.

Here it's represented by `nix flake update --commit-lock-file`, but it happens with `nix-build` too (the stable CLI).

```
+ nix flake update --commit-lock-file
running auto-GC to free 230875004928 bytes
waiting for the big garbage collector lock...
finding garbage collector roots...
deleting garbage...
deleting unused links...
note: currently hard linking saves -0.00 MiB
error: renaming '/nix/store/add-14603-0/x' to '/nix/store/zkjybwz3wlc0vsyjsnpgi5nmln8vb48g-source'
(use '--show-trace' to show detailed location information)
```

**Steps To Reproduce**

1. Configure Nix to do auto-GC, for example with `min-free = 17179869184  # 16 GiB` in `/etc/nix/nix.conf`. (I also have a `max-free = ` setting, but I don't think that matters.)
2. Fill up the disk.
3. Start a build and watch auto-GC kick in. (And additionally, if my theory is correct, auto-GC must fail to actually free any/enough space.)
3. Hopefully you'll see the above error.

**Expected behavior**
If the problem is that after auto-GC there isn't actually more/enough space available, then Nix should say that (e.g. `error: insufficient free disk space, aborting build`) instead of a very low-level `error: renaming A to B` message. (It's not like the disk is _full_, it just has less than `min-free =` available, so it doesn't make sense to me that a rename operation should fail.)

**`nix-env --version` output**

```
$ nix-env --version
nix-env (Nix) 2.9.0
```

agbrooks:
We've been encountering the same problem under similar conditions.

Our steps for reproducing the issue are:

1. `nix-collect-garbage -d`
2. Increase `min-free` above `total disk size - store size`, then restart Nix daemon to pick up config.
3. Use `builtins.fetchGit` to acquire some sources not currently in the store. 

Interestingly, neither `builtins.fetchTarball` nor `builtins.path` seems to exhibit this problem. I'm not sure if this problem is specific to `builtins.fetchGit`. In your situation, are you also using `builtins.fetchGit` somewhere or something else, @bjornfor ?

Not sure if it makes any difference, but we're on Nix 2.9.1 and running Nix on an Ubuntu installation.

[...]

PR #6994, opened 2022-09-02
Prevent tempdir from being GC-ed before addToStoreFromDump completes

agbrooks:
This fixes [issue 6823](https://github.com/NixOS/nix/issues/6823) by creating the temp directory used in `LocalStore::addToStoreFromDump` outside of the Nix store. Otherwise, it's possible for the temporary directory to be prematurely garbage-collected when automatic GC is enabled.

I've written up a few more details and a typical presentation of the bug in the linked issue.

### DIFF
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -622,9 +622,8 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
         /* There may be temp directories in the store that are still in use
            by another process. We need to be sure that we can acquire an
            exclusive lock before deleting them. */
-        AutoCloseFD tmpDirFd;
-        if (baseName.rfind("add-", 0) == 0) {
-            tmpDirFd = open(realPath.c_str(), O_RDONLY | O_DIRECTORY);
+        if (baseName.find("tmp-", 0) == 0) {
+            AutoCloseFD tmpDirFd = open(realPath.c_str(), O_RDONLY | O_DIRECTORY);
             if (tmpDirFd.get() == -1 || !lockFile(tmpDirFd.get(), ltWrite, false)) {
                 debug("skipping locked tempdir '%s'", realPath);
                 return;
--- a/src/libstore/gc.cc
+++ b/src/libstore/gc.cc
@@ -619,6 +619,18 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
         Path path = storeDir + "/" + std::string(baseName);
         Path realPath = realStoreDir + "/" + std::string(baseName);
 
+        /* There may be temp directories in the store that are still in use
+           by another process. We need to be sure that we can acquire an
+           exclusive lock before deleting them. */
+        AutoCloseFD tmpDirFd;
+        if (baseName.rfind("add-", 0) == 0) {
+            tmpDirFd = open(realPath.c_str(), O_RDONLY | O_DIRECTORY);
+            if (tmpDirFd.get() == -1 || !lockFile(tmpDirFd.get(), ltWrite, false)) {
+                debug("skipping locked tempdir '%s'", realPath);
+                return;
+            }
+        }
+
         printInfo("deleting '%1%'", path);
 
         results.paths.insert(path);
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1388,7 +1388,7 @@ StorePath LocalStore::addToStoreFromDump(Source & source0, std::string_view name
         StringSource dumpSource { dump };
         ChainSource bothSource { dumpSource, source };
 
-        auto tempDir = createTempDir(realStoreDir, "add");
+        auto tempDir = createTempDir("", "add");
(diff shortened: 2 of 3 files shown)


thufschmitt (changes requested):
Ahah good catch :)

I'm a bit annoyed by the solution because the code exploits the fact that `tempPath` and `realPath` will be on the same partition to be able to just rename the former to the latter, and avoid some spurious IO ([here](https://github.com/NixOS/nix/blob/master/src/libstore/local-store.cc#L1433)), and it would be a shame to loose that.

What I think we could do instead is:
- Use a proper (temporary) store path instead of an opaque temp dir with something like `makeStorePath("tmpAddToStore:" + someRandomBits, Hash(htSHA256), name)`
- Register it as a temporary gc root with `addTempRoot` (which will automatically be freed once the executable exits)

Thanks for the detailed write-up in the issue btw :)

edolstra:
I think a better solution is to acquire a lock that the garbage collector can test. E.g. if the garbage collector sees a file named `add.<something>`, it opens it and tries to get a non-blocking exclusive lock, which will fail if the file is still in use.

agbrooks:
Thank you both for the swift feedback!

I've pushed a commit that implements the approach described by @edolstra and appears to fix the bug. This codebase is pretty new to me, so any feedback/suggestions would be very welcome.

As part of this work, I took the liberty of updating [the existing `LocalStore::createTempDirInStore` function](https://github.com/NixOS/nix/blob/0c7f213c87c11c6c0f6b2fa07ba96548f6ea85ad/src/libstore/local-store.cc#L1511) to get exclusive access to an in-store temp dir. Previously, this function was unused (AFAIK) and broken (the [`parseStorePath` in `createTempDirInStore`](https://github.com/NixOS/nix/blob/0c7f213c87c11c6c0f6b2fa07ba96548f6ea85ad/src/libstore/local-store.cc#L1519) is doomed to fail because the generated temp dir path is too short to be a valid store path).

src/libstore/local-store.cc
```diff
@@ -1431,6 +1433,7 @@ StorePath LocalStore::addToStoreFromDump(Source & source0, std::string_view name
             } else {
                 /* Move the temporary path we restored above. */
                 moveFile(tempPath, realPath);
+                tempDirFd.close();
```
thufschmitt, line 1436: Nitpicking a bit, but since it's self closing, I'd rather not close it explicitly, just to keep the accidental complexity out of the way as much as possible

src/libstore/gc.cc
```diff
@@ -619,6 +619,18 @@ void LocalStore::collectGarbage(const GCOptions & options, GCResults & results)
         Path path = storeDir + "/" + std::string(baseName);
         Path realPath = realStoreDir + "/" + std::string(baseName);
 
+        /* There may be temp directories in the store that are still in use
+           by another process. We need to be sure that we can acquire an
+           exclusive lock before deleting them. */
+        AutoCloseFD tmpDirFd;
+        if (baseName.rfind("add-", 0) == 0) {
```
thufschmitt, line 626: Any reason the `AutoCloseFD` is declared outside of the block? Doesn't change a lot, but that make Nix create one instance for each store path rather than just the ones that match the pattern

```suggestion
        if (baseName.rfind("add-", 0) == 0) {
                AutoCloseFD tmpDirFd;
```

thufschmitt, line 626:
```suggestion
        if (baseName.find("add-", 0) == 0) {
```

(I think. You would want `add-add-foo` to match I imagine, so you need to start the search from the left)

src/libstore/local-store.cc
```diff
@@ -1507,18 +1510,24 @@ StorePath LocalStore::addTextToStore(
...
         /* There is a slight possibility that `tmpDir' gets deleted by
-           the GC between createTempDir() and addTempRoot(), so repeat
-           until `tmpDir' exists. */
-        tmpDir = createTempDir(realStoreDir);
-        addTempRoot(parseStorePath(tmpDir));
-    } while (!pathExists(tmpDir));
-    return tmpDir;
+           the GC between createTempDir() and when we acquire a lock on it.
+           We'll repeat until 'tmpDir' exists and we've locked it. */
+        tmpDirFn = createTempDir(realStoreDir, "add");
```
thufschmitt, line 1523: Since this function is (in the name at least) more generic than just the `addToStore` stuff, maybe we could change `add` to something more generic like `tmp`? (And sync that in the gc code)

thufschmitt (changes requested):
Thanks, that looks great overall :)

Left a few comments/suggestions/nitpicks but nothing really serious

src/libstore/local-store.cc
agbrooks: Agreed, `tmp` is a more descriptive name

thufschmitt (approved):
Thanks :)

It would have been great to have a way to test this, but I don't see a nice way of doing it, lo let's merge as it is

PR #6995, opened 2022-09-03
Disable SA_RESTART for some signals on macOS

matthewbauer:
Disables the SA_RESTART behavior on macOS which causes:

> Restarting of pending calls is requested by setting the SA_RESTART bit
> in sa_flags. The affected system calls include read(2), write(2),
> sendto(2), recvfrom(2), sendmsg(2) and recvmsg(2) on a communications
> channel or a slow device (such as a terminal, but not a regular file)
> and during a wait(2) or ioctl(2).

From: https://man.openbsd.org/sigaction#SA_RESTART

This being set on macOS caused a bug where read() calls on the client to the daemon
socket were blocking after a SIGINT was received. As a result,
checkInterrupt was never reached even though the signal was received
by the signal handler thread.

On Linux, SA_RESTART is disabled by default. This probably effects
other BSDs but I don’t have the ability to test it there right now.

cc @edolstra

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -208,13 +208,7 @@ void initNix()
     if (sigaction(SIGHUP, &act, 0)) throw SysError("handling SIGHUP");
     if (sigaction(SIGPIPE, &act, 0)) throw SysError("handling SIGPIPE");
     if (sigaction(SIGQUIT, &act, 0)) throw SysError("handling SIGQUIT");
-    if (sigaction(SIGILL, &act, 0)) throw SysError("handling SIGILL");
     if (sigaction(SIGTRAP, &act, 0)) throw SysError("handling SIGTRAP");
-    if (sigaction(SIGABRT, &act, 0)) throw SysError("handling SIGABRT");
-    if (sigaction(SIGFPE, &act, 0)) throw SysError("handling SIGFPE");
-    if (sigaction(SIGBUS, &act, 0)) throw SysError("handling SIGBUS");
-    if (sigaction(SIGXCPU, &act, 0)) throw SysError("handling SIGXCPU");
-    if (sigaction(SIGXFSZ, &act, 0)) throw SysError("handling SIGXFSZ");
 #endif
 
     /* Register a SIGSEGV handler to detect stack overflows. */
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -194,9 +194,16 @@ void initNix()
     /* HACK: on darwin, we need can’t use sigprocmask with SIGWINCH.
      * Instead, add a dummy sigaction handler, and signalHandlerThread
      * can handle the rest. */
-    struct sigaction sa;
-    sa.sa_handler = sigHandler;
-    if (sigaction(SIGWINCH, &sa, 0)) throw SysError("handling SIGWINCH");
+    act.sa_handler = sigHandler;
+    if (sigaction(SIGWINCH, &act, 0)) throw SysError("handling SIGWINCH");
+
+    // Disable SA_RESTART for interrupts, so that system calls on this thread
+    // error with EINTR like they do on Linux, and we don’t hang forever.
+    act.sa_handler = SIG_DFL;
+    if (sigaction(SIGINT, &act, 0)) throw SysError("handling SIGINT");
+    if (sigaction(SIGTERM, &act, 0)) throw SysError("handling SIGTERM");
+    if (sigaction(SIGHUP, &act, 0)) throw SysError("handling SIGHUP");
+    if (sigaction(SIGPIPE, &act, 0)) throw SysError("handling SIGPIPE");
 #endif
 
     /* Register a SIGSEGV handler to detect stack overflows. */
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -181,8 +181,9 @@ void initNix()
     /* Reset SIGCHLD to its default. */
     struct sigaction act;
     sigemptyset(&act.sa_mask);
-    act.sa_handler = SIG_DFL;
     act.sa_flags = 0;
+
+    act.sa_handler = SIG_DFL;
     if (sigaction(SIGCHLD, &act, 0))
         throw SysError("resetting SIGCHLD");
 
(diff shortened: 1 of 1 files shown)


abathur:
What were the symptoms of this? (there's a traumatized part of my brain wondering if this might be the root cause of #3605).

matthewbauer:
> What were the symptoms of this? (there's a traumatized part of my brain wondering if this might be the root cause of #3605).

I don't think so. This would only happen when you're sending a signal, and it happens to be during a syscall. So you'd just get a "^C" and the process never exits.

matthewbauer:
Trying to figure out the difference between Linux & macOS here, I created this program:

```c
#include <signal.h>
#include <stdio.h>

int main() {
    struct sigaction act;
    int signals[] = {
      SIGHUP,
      SIGINT,
      SIGQUIT,
      SIGILL,
      SIGTRAP,
      SIGABRT,
      SIGFPE,
      SIGKILL,
      SIGBUS,
      SIGSEGV,
      SIGSYS,
      SIGPIPE,
      SIGALRM,
      SIGTERM,
      SIGURG,
      SIGSTOP,
      SIGTSTP,
      SIGCONT,
      SIGCHLD,
      SIGTTIN,
      SIGTTOU,
      SIGIO,
      SIGXCPU,
      SIGXFSZ,
      SIGVTALRM,
      SIGPROF,
      SIGWINCH,
      SIGUSR1,
      SIGUSR2
    };
    const char* signal_names[] = {
      "SIGHUP",
      "SIGINT",
      "SIGQUIT",
      "SIGILL",
      "SIGTRAP",
      "SIGABRT",
      "SIGFPE",
      "SIGKILL",
      "SIGBUS",
      "SIGSEGV",
      "SIGSYS",
      "SIGPIPE",
      "SIGALRM",
      "SIGTERM",
      "SIGURG",
      "SIGSTOP",
      "SIGTSTP",
      "SIGCONT",
      "SIGCHLD",
      "SIGTTIN",
      "SIGTTOU",
      "SIGIO",
      "SIGXCPU",
      "SIGXFSZ",
      "SIGVTALRM",
      "SIGPROF",
      "SIGWINCH",
      "SIGUSR1",
      "SIGUSR2"
    };
    for (int i = 0; i < sizeof(signals) / sizeof(int); i++) {
      sigaction(signals[i], NULL, &act);
      printf("%s default sa_mask: %x\n", signal_names[i], act.sa_mask);
      printf("%s default sa_flags: %x\n", signal_names[i], act.sa_flags);
      if (act.sa_handler == SIG_DFL) {
        printf("%s default sa_handler: SIG_DFL\n", signal_names[i]);
      } else if (act.sa_handler == SIG_ERR) {
        printf("%s default sa_handler: SIG_ERR\n", signal_names[i]);
      } else if (act.sa_handler == SIG_IGN) {
        printf("%s default sa_handler: SIG_IGN\n", signal_names[i]);
      } else {
        printf("%s default sa_handler: custom function\n", signal_names[i-1]);
      }
    }
}
```

and diffed the output on macOS and Linux resulting in:

```diff
5c5
< SIGINT default sa_flags: 2
---
> SIGINT default sa_flags: 0
8c8
< SIGQUIT default sa_flags: 2
---
> SIGQUIT default sa_flags: 0
11c11
< SIGILL default sa_flags: 2
---
> SIGILL default sa_flags: 0
14c14
< SIGTRAP default sa_flags: 2
---
> SIGTRAP default sa_flags: 0
17c17
< SIGABRT default sa_flags: 2
---
> SIGABRT default sa_flags: 0
20c20
< SIGFPE default sa_flags: 2
---
> SIGFPE default sa_flags: 0
23c23
< SIGKILL default sa_flags: 2
---
> SIGKILL default sa_flags: 0
26c26
< SIGBUS default sa_flags: 2
---
> SIGBUS default sa_flags: 0
29c29
< SIGSEGV default sa_flags: 2
---
> SIGSEGV default sa_flags: 0
32c32
< SIGSYS default sa_flags: 2
---
> SIGSYS default sa_flags: 0
35c35
< SIGPIPE default sa_flags: 2
---
> SIGPIPE default sa_flags: 0
41c41
< SIGTERM default sa_flags: 2
---
> SIGTERM default sa_flags: 0
44c44
< SIGURG default sa_flags: 2
---
> SIGURG default sa_flags: 0
47c47
< SIGSTOP default sa_flags: 2
---
> SIGSTOP default sa_flags: 0
50c50
< SIGTSTP default sa_flags: 2
---
> SIGTSTP default sa_flags: 0
53c53
< SIGCONT default sa_flags: 2
---
> SIGCONT default sa_flags: 0
59c59
< SIGTTIN default sa_flags: 2
---
> SIGTTIN default sa_flags: 0
62c62
< SIGTTOU default sa_flags: 2
---
> SIGTTOU default sa_flags: 0
65c65
< SIGIO default sa_flags: 2
---
> SIGIO default sa_flags: 0
68c68
< SIGXCPU default sa_flags: 2
---
> SIGXCPU default sa_flags: 0
71c71
< SIGXFSZ default sa_flags: 2
---
> SIGXFSZ default sa_flags: 0
74c74
< SIGVTALRM default sa_flags: 2
---
> SIGVTALRM default sa_flags: 0
77c77
< SIGPROF default sa_flags: 2
---
> SIGPROF default sa_flags: 0
83c83
< SIGUSR1 default sa_flags: 2
---
> SIGUSR1 default sa_flags: 0
86c86
< SIGUSR2 default sa_flags: 2
---
> SIGUSR2 default sa_flags: 0
```

src/libmain/shared.cc
```diff
@@ -194,9 +195,26 @@ void initNix()
...
+    if (sigaction(SIGHUP, &act, 0)) throw SysError("handling SIGHUP");
+    if (sigaction(SIGPIPE, &act, 0)) throw SysError("handling SIGPIPE");
+    if (sigaction(SIGQUIT, &act, 0)) throw SysError("handling SIGQUIT");
+    if (sigaction(SIGILL, &act, 0)) throw SysError("handling SIGILL");
+    if (sigaction(SIGTRAP, &act, 0)) throw SysError("handling SIGTRAP");
+    if (sigaction(SIGABRT, &act, 0)) throw SysError("handling SIGABRT");
+    if (sigaction(SIGFPE, &act, 0)) throw SysError("handling SIGFPE");
+    if (sigaction(SIGBUS, &act, 0)) throw SysError("handling SIGBUS");
+    if (sigaction(SIGXCPU, &act, 0)) throw SysError("handling SIGXCPU");
+    if (sigaction(SIGXFSZ, &act, 0)) throw SysError("handling SIGXFSZ");
```
edolstra, line 217: Some of these seem unnecessary, namely `SIGILL`, `SIGABRT`, `SIGFPE`, `SIGBUS`, `SIGXCPU` and `SIGXFSZ`. Really we only care about getting EINTR for a ctrl-c or client disconnect.

PR #6997, opened 2022-09-04
libfetchers: avoid api.github.com ratelimit

### LINKED ISSUE #5409: Add option to use github.com instead of api.github.com (which may be blocked) for `github:` flake ref
My company recently **_blocked_** access to `https://api.github.com`, but we can still access most parts of `https://github.com`.

When referencing flake inputs with `github:numtide/devshell` (for example), the download URL uses the blocked URL 😢. Thus downloading the tarballs fails.

Related code:
https://github.com/NixOS/nix/blob/fda4efff873730f0065cf35f0bc5b36983d675c7/src/libfetchers/github.cc#L260-L267
and
https://github.com/NixOS/nix/blob/fda4efff873730f0065cf35f0bc5b36983d675c7/src/libfetchers/github.cc#L243-L247

It also blocks my use of `nix run foo#bar` where `foo` is a flake reference to a github branch,
it tries to access `https://api.github.com/repos/nixos/nixpkgs/commits/nixpkgs-unstable` but fails..

Related code: (thanks @Kha for the find)
https://github.com/NixOS/nix/blob/37cc50f2c88b50086a1a8ed20631d1427d445b7e/src/libfetchers/github.cc#L246-L247

A config option to choose `github.com` host to download tarballs.

For example, the equivalent of:
`https://api.github.com/repos/numtide/devshell/tarball/4b5ac7cf7d9a1cc60b965bb51b59922f2210cbc7`
is:
`https://github.com/numtide/devshell/tarball/4b5ac7cf7d9a1cc60b965bb51b59922f2210cbc7`
(which does work at my company)

Use `git+https://github.com/numtide/devshell` for my own dependencies.
⚠️ BUT if one of my dependency uses a `github:...` flake reference, I'm back to square one and I can't use that dependency.

**Additional context**
`nix --version`: `nix (Nix) 2.4pre20210326_dd77f71`

SuperSandro2000:
Closes #5409

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -262,13 +262,8 @@ struct GitHubInputScheme : GitArchiveInputScheme
 
     DownloadUrl getDownloadUrl(const Input & input) const override
     {
-        // FIXME: use regular /archive URLs instead? api.github.com
-        // might have stricter rate limits.
         auto host = maybeGetStrAttr(input.attrs, "host").value_or("github.com");
-        auto url = fmt(
-            host == "github.com"
-            ? "https://api.%s/repos/%s/%s/tarball/%s"
-            : "https://%s/api/v3/repos/%s/%s/tarball/%s",
+        auto url = fmt("https://%s/%s/%s/archive/%s.tar.gz",
             host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
             input.getRev()->to_string(Base16, false));
 


Kha:
See also https://github.com/NixOS/nix/issues/5409 (in particular, https://github.com/NixOS/nix/issues/5409#issuecomment-949507854)

edolstra:
Doesn't this break support for private repositories?

It would be nice to somehow detect that a repo is public and use the non-rate-limited endpoint in that case.

Mic92:
> Doesn't this break support for private repositories?
> 
> It would be nice to somehow detect that a repo is public and use the non-rate-limited endpoint in that case.

Private repos would only ever work if there is a github api token set. So how about using the public archive by default when no github api token is set? Than we get best of both worlds.

fricklerhandwerk:
Triaged in Nix team meeting:

@SuperSandro2000 are you still interested in finishing this? Please implement @Mic92's suggestion https://github.com/NixOS/nix/pull/6997#issuecomment-1244389233 and we can merge.

Mic92:
Was this not already merged in a different pr?

edolstra:
Yes indeed, this was already done in d9abce4ad4.

PR #6998, opened 2022-09-04
narHash examples and documentation

nrdsp:


knedlsepp:
I'm slightly confused. If no `rev` is specified, but only the `narHash`, how does nix know which `rev` to fetch? I would assume this will only fetch `HEAD` and then break for anybody that fetches the input when the remote's HEAD has advanced.
Is there something I'm misunderstanding here?

thufschmitt:
> I would guess that anyone who would want to use the narHash atribute would do it as a substitute for rev, meaning that each commit has its own sha256 hash that can be used instead

There's actually a big difference between both: Using `rev` causes Nix to use the commits it points to, but with `narHash` Nix has no way of knowing which commit has the right hash (short of traversing the entire history graph and archiving each commit to get its narHash which is prohibitely expensive). So given `github:edolstra/dwarffs?narHash=sha256-kq5dejftKMkHMBhwyocis9wZrkdCIZUFm2IfO6KLYjU=`, all Nix can do is detch the HEAD of the repo, and check whether it has the right nar hash (and fail otherwise)

edolstra:
Note that #6530 largely removes the `narHash` attribute except for some flake input types (like `path`) since it's too expensive to compute.

fricklerhandwerk:
@nrdsp shall we close this then?

Issue #7000, opened 2022-09-05
Remove 'Experimental Features' warning


mpscholten:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when nix doesn't do anything when I use "experimental features" such as nix flake or nix develop.

I always get this error:

```bash
error: experimental Nix feature 'nix-command' is disabled; use '--extra-experimental-features nix-command' to override
```

Instead I expect nix to just run it's thing. I don't care that this feature is experimental, I just want nix to run the command I told it to run. Why does everything in nix needs to be so frustrating?

**Describe the solution you'd like**

The best solution would be to remove the warning. I know what I'm doing, I don't need the program to stop me.

Another solution would be to better describe how to fix the problem. E.g. in my case I was checking out an internal repo used at my company which contained a `.envrc` file which uses the direnv nix flake integration. So in that case I cannot pass any parameters to nix, as the nix command is called by direnv itself.

**Describe alternatives you've considered**

I've tried to get rid of this error a few times in the past when trying out `nix develop`. I never got `nix develop` working (the error message is not very helpful on NixOS as well) and just gave up.

**Additional context**
OS: macOS Monterey
Nix version: `nix (Nix) 2.10.3`

SuperSandro2000:
> Instead I expect nix to just run it's thing. I don't care that this feature is experimental, I just want nix to run the command I told it to run.

Just add ``experimental-features = flakes nix-command`` to ``nix.conf`` to always have it enabled.

> Why does everything in nix needs to be so frustrating?

Because nix is not pure magic and can do everything perfect first try but the community is working on it one step at the time.

> The best solution would be to remove the warning. I know what I'm doing, I don't need the program to stop me.

It will be, when the features are no longer experimental.

mpscholten:
Thanks 👍 

> Just add experimental-features = flakes nix-command to nix.conf to always have it enabled.

It would be useful to have this hint in the error message itself.

> Because nix is not pure magic and can do everything perfect first try but the community is working on it one step at the time.

👍 Sorry, I was very frustrated here. I hope my intent to improve nix is clear here as well :)

> It will be, when the features are no longer experimental.

I understand these are experimental features, but what's the reason users need to opt in to use them? It feels like it just adds trouble to the workflow. Especially as these experimental nix features are now used more often.

sellout:
I agree that the message should describe how to enable them (or at least include a link to a place that gives more context and explains how to enable them while also explaining why they're experimental.

[...]

PR #7003, opened 2022-09-05
Improve experimental-features error wording

SuperSandro2000:
Motivation #7000

### DIFF
--- a/src/libutil/experimental-features.cc
+++ b/src/libutil/experimental-features.cc
@@ -52,7 +52,7 @@ std::set<ExperimentalFeature> parseFeatures(const std::set<std::string> & rawFea
 }
 
 MissingExperimentalFeature::MissingExperimentalFeature(ExperimentalFeature feature)
-    : Error("experimental Nix feature '%1%' is disabled; pass '--extra-experimental-features %1%' as an argument to enable it", showExperimentalFeature(feature))
+    : Error("experimental Nix feature '%1%' is disabled; add '--extra-experimental-features %1%' to enable it", showExperimentalFeature(feature))
     , missingFeature(feature)
 {}
 
--- a/src/libutil/experimental-features.cc
+++ b/src/libutil/experimental-features.cc
@@ -52,7 +52,7 @@ std::set<ExperimentalFeature> parseFeatures(const std::set<std::string> & rawFea
 }
 
 MissingExperimentalFeature::MissingExperimentalFeature(ExperimentalFeature feature)
-    : Error("experimental Nix feature '%1%' is disabled; use '--extra-experimental-features %1%' to override", showExperimentalFeature(feature))
+    : Error("experimental Nix feature '%1%' is disabled; pass '--extra-experimental-features %1%' as an argument to enable it", showExperimentalFeature(feature))
     , missingFeature(feature)
 {}
 


thufschmitt (commented):
Thanks for the upgrade :)

I'm not sure we want to push the message in this direction (at least not according to the original issue): The problem seems to be not that people don't get that it should be passed as an argument, but rather that it's not obvious that it can be set in stone (well, in the config).

Maybe we could have something like that instead?

> The experimental Nix feature '%1%' is disabled.
> You can enable it by adding '%1%'  to the 'experimental-features' option in your Nix config or passing '--extra-experimental-features %1%'

rapenne-s:
I'm adding the message isn't helpful for `nix flake`

```
> nix flake
```

> Ok, let's add the parameter

```
> nix flake --extra-experimental-features "nix-command"
error: 'nix flake' requires a sub-command
Try 'nix --help' for more information
```

> Ok! I just forgot to pass a parameter to nix flake

```
> nix flake show --extra-experimental-features "nix-command"
error: experimental Nix feature 'nix-command' is disabled; use '--extra-experimental-features flakes' to override
error: could not find a flake.nix file
```

This is highly confusing

- `nix flake` requires `flakes` feature to be enabled?
- `nix flake show` without `flakes feature` reports an error that an experimental feature is missing, but at the same time it reports an error about the missing `flake.nix` file if there are any in the directory

It's also tempting to remove "nix-command" to replace with "flakes" thinking it would be enough

thufschmitt:
> To enable it permanently, add
>
>    experimental-features = %1%
>    
> to ~/.config/nix/nix.conf (for the current user)
> or /etc/nix/nix.conf (for all users).

There's still the risk of people adding `experimental-features = foo` to `~/.config/nix/nix.conf` and override another setting in `/etc/nix/nix.conf`. So I'd suggest suggesting using `extra-experimental-features` here too. But I like it otherwise. And it being verbose is probably not an issue.

> NixOS could patch this string to add appropriate instructions.

That sounds harder. But even without that it's better.

SuperSandro2000:
Yes, it should use ``extra-experimental-features``.

> NixOS could patch this string to add appropriate instructions.

Just hardcode the instructions for NixOS in.

> One could make it easier by first changing that behavior to consider values from all lines containing `experimental-features`.

Touching the config parser will probably break other things and making this unique to one option just makes things needlessly complicated.

fricklerhandwerk:
> Just hardcode the instructions for NixOS in.

No, absolutely no. Beginners already have severe trouble understanding that there even is a difference between Nix and NixOS. And this error is one people will encounter in their first hour or couple of days, depending on how they get introduced to Nix. Telling them to do something they don't understand how it even belongs together will just create more confusion.

@thufschmitt Why is it hard for NixOS to use a patched version of Nix where this string is different? It's just another package, isn't it?

thufschmitt:
> @thufschmitt Why is it hard for NixOS to use a patched version of Nix where this string is different? It's just another package, isn't it?

It's just that it requires maintaining a custom patch for a small edge-case. Especially given that (ideally 😒) most people shouldn't have to fiddle with experimental features unless they know what they are doing

thufschmitt:
(And having a different package for Nix-on-nixos and Nix-on-nixpkgs which is a bit strange)

SuperSandro2000:
We shouldn't require to add NixOS specific patches to our own package manager.

fricklerhandwerk:
It would not be required, but NixOS (or any other system that automates configuration such that users don't have to write stuff to obscure files, such as nix-darwin) could improve their specific user experience by adapting the hint to the interface they provide to configure Nix.

Not mentioning whatever external way of dealing with Nix configuration is still better than confusing beginners with stuff they may never come to use if they are confused by information that will misguide them. I already see people on Ubuntu and Arch going "where is that configuration.nix? why is this so hard to begin with?" and eventually ending up with "Nix documentation is just horrible".

Ericson2314:
There is a lot of discussion on things beyond what the PR is doing. Is there a small step we could take in the right direction?

fricklerhandwerk:
It seems we agreed on having a more instructive error message, but disagreed over how to deal with the case where Nix is configured declaratively.

I'm still convinced Nix should not assume or even concern itself with how it is used and should focus on documenting itself consistently.

There may be less invasive ways to patch the error message contents for NixOS or other declarative systems, such as wrapping Nix to add an environment variable that will determine the location, but that will require thinking it through. We could merge the hard coded proposal and accept sensible PRs that allow for a more flexible setup.

roberth:
Thanks @SuperSandro2000 for the more accurate phrasing.

PR #7010, opened 2022-09-06
nix develop: Ignore stdenv's $SHELL

edolstra:
Stdenv sets this to a bash that doesn't have readline/completion support, so running `nix shell` inside a `nix develop` gives you a crippled shell. So let's just ignore the derivation's `$SHELL`.

This could break interactive use of build phases that use `$SHELL`, butthey appear to be fairly rare.

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -246,6 +246,7 @@ struct Common : InstallableCommand, MixProfile
         "NIX_LOG_FD",
         "NIX_REMOTE",
         "PPID",
+        "SHELL",
         "SHELLOPTS",
         "SSL_CERT_FILE", // FIXME: only want to ignore /no-cert-file.crt
         "TEMP",
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -483,12 +483,12 @@ LockedFlake lockFlake(
                                 } else if (auto follows = std::get_if<1>(&i.second)) {
                                     if (! trustLock) {
                                         // It is possible that the flake has changed,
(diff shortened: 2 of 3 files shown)


src/nix/develop.cc
```diff
@@ -246,6 +246,7 @@ struct Common : InstallableCommand, MixProfile
         "NIX_LOG_FD",
         "NIX_REMOTE",
         "PPID",
+        "SHELL",
```
thufschmitt, line 249: Should we instead set `$SHELL` to the interactive bash we fetch from `nixpkgs`? that way most stuff will just work as expected since it will still point to bash

ncfavier:
What does this fix exactly? The only place I see `SHELL` used in the Nix codebase is in `nix run`.

edolstra:
@ncfavier It fixes this:
```
[eelco@localhost:~/Dev/nix]$ nix develop
[eelco@localhost:~/Dev/nix]$ nix shell
\[\][eelco@localhost:~/Dev/nix]$\[\] 
```
i.e. the inner shell doesn't have readline/completion support.

ShamrockLee:
@ncfavier It would also fix the misinterpreted VSCode terminal IIUC.

Hoverbear:
I encountered this particular bug often when using `starship` and using `nix develop` then launching either vscode or neovim inside, and referring to their terminals.

Eg. (If you use `starship)
```bash
nix develop
nvim .
```

Then run `:terminal`


image

truh:
I've just upgraded from nix 2.18 to 2.24.10 with the upcoming NixOS release and it looks to me like this change hasn't made it into nix versions >=2.20.0. Has this been reverted intentionally? Will this be incorporated into newer versions?

Issue #6202, opened 2022-03-03
nix develop --build and phases confusion


thufschmitt:
When buildPhase is defined, `nix develop --build` does not work due to confusion between the variable and function during stdenv eval.

```
{
  description = "A flake that uses Hello World";
  # inputs.flake01.url = "/tmp/flake01";
  outputs = { self, nixpkgs }: {
    defaultPackage.x86_64-linux =
      with import nixpkgs { system = "x86_64-linux"; };
      stdenv.mkDerivation {
        name = "flake02";
        src = ./d;
        buildInputs = [ hello ];
        buildPhase = ''
          hello >> ./hello.txt
        '';
        installPhase = ''
          mkdir -p $out
          install -m 644 -t $out/ hello.txt
        '';
      };
  };
}
```

need feedback:
https://github.com/flox/nix/commit/c03252cab70379214fabcd839cf432d447a13c42

Or change the order here in stdenv? https://github.com/NixOS/nixpkgs/blob/master/pkgs/stdenv/generic/setup.sh#L72-L80

tomjnixon:
I ran into this too, and it was a bit confusing. This also applies to using `nix-shell`. Three possible resolutions for this:

1) Add a function in the default builder that you could run with `runPhase buildPhase` (or `runPhase build`?), which would have the same behavior as [this line](https://github.com/NixOS/nixpkgs/blob/7eb1a32324f039230261195ee8f913bfb4617f91/pkgs/stdenv/generic/setup.sh#L1345), and update all the documentation to say to use this rather than `buildPhase`.

2) Change the default phase definitions to evaluate the contents of the corresponding variable, and have default definitions of those variables with the current behavior.

3) Just update the documentation to say that you might need to run `eval  "$buildPhase"` rather than `buildPhase` if it's overridden, and add a change in nix to make `--build` etc. work properly.

IMO 1 is the cleanest, but is annoying to type and there will be lots of tutorials etc. that will never be fixed. 2 is pragmatic, but still results in some confusion with user-defined (or non-default) phases, or where `buildPhase` is re-defined as a function. 3 is lowest risk but is a bad user experience.

Ninlives:
Any progress? I ran into this problem too, which is very confusing to me.

Ninlives:
@thufschmitt I've created a [PR](https://github.com/NixOS/nixpkgs/pull/230468) for nixpkgs in preparation of fixing this issue, please take a look at it.

PR #7023, opened 2022-09-11
evaluate phases in nix develop the same way as genericBuild

arianvp:
This way `nix develop` works with  meson, cmake etc; which set the `xxxPhase=` environment variable instead of overriding the `xxxPhase` function.

open question. I think in these cases `foundMakefile=1` should not be set. But we'd need to fix that in  https://github.com/NixOS/nixpkgs/blob/master/pkgs/stdenv/generic/setup.sh first I think

### DIFF
--- a/src/nix/develop.cc
+++ b/src/nix/develop.cc
@@ -482,7 +482,7 @@ struct CmdDevelop : Common, MixEnvironment
             // FIXME: foundMakefile is set by buildPhase, need to get
             // rid of that.
             script += fmt("foundMakefile=1\n");
-            script += fmt("runHook %1%Phase\n", *phase);
+            script += fmt("eval \"${!%1%Phase:-$1%1Phase}\";\n", *phase);
         }
 
         else if (!command.empty()) {


kira-bruneau (commented):
Thanks for this fix @arianvp! I noticed this problem too, but never really looked into it before :sweat_smile: 

> open question. I think in these cases foundMakefile=1 should not be set. But we'd need to fix that in https://github.com/NixOS/nixpkgs/blob/master/pkgs/stdenv/generic/setup.sh first I think

Hmm, yeah. It looks like `checkPhase` and `installCheckPhase` should check for the makefile themselves so they aren't dependent on the `buildPhase`. I wonder how many other places we have dependencies between phases like this?

firestack:
uhhhh, related? https://github.com/flox/nix/commit/c03252cab70379214fabcd839cf432d447a13c42

https://github.com/NixOS/nix/issues/6202

thufschmitt (changes requested):
Duh, I though that's what `runHook` was meant to do :confused: 

Thanks for the fix then :) That looks good (well, it's literally what the stdenv does, so it better be…)

Can you just add a test for that? I guess you can add a new derivation (with a custom `buildPhase`) in `tests/shell.nix` and run `nix develop --build` on it in `tests/nix-shell.sh`.

arianvp:
It feels more elegant to fix stdenv. Let me give that a try first

PR #7029, opened 2022-09-12
libfetchers: avoid api.github.com ratelimit if no github token is set

Mic92:
If we don't have any github token, we won't be able to fetch private repos, but we are also more likely to run into API limits since we don't have a token. To mitigate this only ever use the github api if we actually have a token.

This is an alternative to https://github.com/NixOS/nix/pull/6997

closes https://github.com/NixOS/nix/issues/5409

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -262,17 +262,19 @@ struct GitHubInputScheme : GitArchiveInputScheme
 
     DownloadUrl getDownloadUrl(const Input & input) const override
     {
-        // FIXME: use regular /archive URLs instead? api.github.com
-        // might have stricter rate limits.
         auto host = maybeGetStrAttr(input.attrs, "host").value_or("github.com");
-        auto url = fmt(
-            host == "github.com"
-            ? "https://api.%s/repos/%s/%s/tarball/%s"
-            : "https://%s/api/v3/repos/%s/%s/tarball/%s",
-            host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
+        Headers headers = makeHeadersWithAuthTokens(host);
+        const char* urlFmt;
+        if (host == "github.com")
+            // If we have no auth headers then we default to the public archive
+            // urls so we do not run into rate limits.
+            urlFmt = headers.empty() ? "https://%s/%s/%s/archive/%s.tar.gz" : "https://api.%s/repos/%s/%s/tarball/%s";
+        else
+            urlFmt = "https://%s/api/v3/repos/%s/%s/tarball/%s";
+
+        auto url = fmt(urlFmt, host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
             input.getRev()->to_string(Base16, false));
 
-        Headers headers = makeHeadersWithAuthTokens(host);
         return DownloadUrl { url, headers };
     }
 


src/libfetchers/github.cc
```diff
@@ -265,14 +265,18 @@ struct GitHubInputScheme : GitArchiveInputScheme
...
         auto host = maybeGetStrAttr(input.attrs, "host").value_or("github.com");
-        auto url = fmt(
-            host == "github.com"
-            ? "https://api.%s/repos/%s/%s/tarball/%s"
-            : "https://%s/api/v3/repos/%s/%s/tarball/%s",
-            host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
+        Headers headers = makeHeadersWithAuthTokens(host);
+        const char* urlFmt;
+        if (host == "github.com")
+            // If we have no auth headers than we default to the public archive
```
knedlsepp, line 271: than -> then

src/libfetchers/github.cc
```diff
@@ -265,14 +265,18 @@ struct GitHubInputScheme : GitArchiveInputScheme
         // FIXME: use regular /archive URLs instead? api.github.com
```
knedlsepp, line 265: FIXME can be dropped now

src/libfetchers/github.cc
```diff
@@ -265,14 +265,18 @@ struct GitHubInputScheme : GitArchiveInputScheme
...
         // might have stricter rate limits.
         auto host = maybeGetStrAttr(input.attrs, "host").value_or("github.com");
-        auto url = fmt(
-            host == "github.com"
-            ? "https://api.%s/repos/%s/%s/tarball/%s"
-            : "https://%s/api/v3/repos/%s/%s/tarball/%s",
-            host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
+        Headers headers = makeHeadersWithAuthTokens(host);
+        const char* urlFmt;
```
knedlsepp, line 269: Any reason to not opt for a std::string here?

Mic92: afaik, this would create an unnecessary string allocation. Also I cannot use const in this case.

edolstra: I think `std::string_view` should work.

edolstra: You can probably also write this as
```
auto urlFmt =
    host == "github.com"
    ? headers.empty() 
       ? "https://%s/%s/%s/archive/%s.tar.gz" 
       : "https://api.%s/repos/%s/%s/tarball/%s"
    : "https://%s/api/v3/repos/%s/%s/tarball/%s"
```

Mic92: > You can probably also write this as
> 
> ```
> auto urlFmt =
>     host == "github.com"
>     ? headers.empty() 
>        ? "https://%s/%s/%s/archive/%s.tar.gz" 
>        : "https://api.%s/repos/%s/%s/tarball/%s"
>     : "https://%s/api/v3/repos/%s/%s/tarball/%s"
> ```

I find this quite hard to read to be honest.

Mic92: https://github.com/NixOS/nix/pull/7039

Mic92: Closed this pr and move ahead in https://github.com/NixOS/nix/pull/7039

PR #7039, opened 2022-09-14
libfetchers: avoid api.github.com ratelimit if no github token is set

Mic92:
If we don't have any github token, we won't be able to fetch private repos, but we are also more likely to run into API limits since we don't have a token. To mitigate this only ever use the github api if we actually have a token.

### DIFF
--- a/src/libfetchers/github.cc
+++ b/src/libfetchers/github.cc
@@ -262,17 +262,20 @@ struct GitHubInputScheme : GitArchiveInputScheme
 
     DownloadUrl getDownloadUrl(const Input & input) const override
     {
-        // FIXME: use regular /archive URLs instead? api.github.com
-        // might have stricter rate limits.
         auto host = maybeGetStrAttr(input.attrs, "host").value_or("github.com");
-        auto url = fmt(
-            host == "github.com"
-            ? "https://api.%s/repos/%s/%s/tarball/%s"
-            : "https://%s/api/v3/repos/%s/%s/tarball/%s",
-            host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
+        Headers headers = makeHeadersWithAuthTokens(host);
+        // If we have no auth headers then we default to the public archive
+        // urls so we do not run into rate limits.
+        const auto urlFmt =
+            host != "github.com"
+            ? "https://%s/api/v3/repos/%s/%s/tarball/%s"
+            : headers.empty()
+            ? "https://%s/%s/%s/archive/%s.tar.gz"
+            : "https://api.%s/repos/%s/%s/tarball/%s";
+
+        const auto url = fmt(urlFmt, host, getStrAttr(input.attrs, "owner"), getStrAttr(input.attrs, "repo"),
             input.getRev()->to_string(Base16, false));
 
-        Headers headers = makeHeadersWithAuthTokens(host);
         return DownloadUrl { url, headers };
     }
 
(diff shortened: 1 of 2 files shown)


Mic92:
Anything else that needs to be done here?

src/libfetchers/github.cc
```diff
@@ -265,14 +265,19 @@ struct GitHubInputScheme : GitArchiveInputScheme
         // FIXME: use regular /archive URLs instead? api.github.com
         // might have stricter rate limits.
         auto host = maybeGetStrAttr(input.attrs, "host").value_or("github.com");
-        auto url = fmt(
+        Headers headers = makeHeadersWithAuthTokens(host);
+        // If we have no auth headers than we default to the public archive
+        // urls so we do not run into rate limits.
+        const auto urlFmt =
```
pleshevskiy, line 271:
```suggestion
        const auto urlFmt =
            host != "github.com"
            ? "https://%s/api/v3/repos/%s/%s/tarball/%s"
            : headers.empty()
            ? "https://%s/%s/%s/archive/%s.tar.gz"
            : "https://api.%s/repos/%s/%s/tarball/%s"
```

src/libfetchers/github.cc
```diff
@@ -265,14 +265,19 @@ struct GitHubInputScheme : GitArchiveInputScheme
         // FIXME: use regular /archive URLs instead? api.github.com
         // might have stricter rate limits.
         auto host = maybeGetStrAttr(input.attrs, "host").value_or("github.com");
-        auto url = fmt(
+        Headers headers = makeHeadersWithAuthTokens(host);
+        // If we have no auth headers than we default to the public archive
```
knedlsepp, line 269: than -> then

src/libfetchers/github.cc
```diff
@@ -265,14 +265,19 @@ struct GitHubInputScheme : GitArchiveInputScheme
         // FIXME: use regular /archive URLs instead? api.github.com
```
knedlsepp, line 265: The FIXME can be dropped now.

SuperSandro2000:
Can this please be merged for the next nix release? Since more and more people switched to flakes we are almost always hitting that rate limit.

SuperSandro2000:
bump

Mic92:
> This seems important to get in, is there anything blocking it?

The lack of Nix maintainers to merge.

thufschmitt (changes requested):
Looks good at a glance, but you'll have to update `tests/github-flakes.nix` as it as a mock of the github API and will need to be extended to support this change.

You can run it with `nix build .\#hydraJobs.tests.githubFlakes`

SuperSandro2000:
https://github.com/Mic92/nix-1/pull/1

Mic92:
The tests have been extended to cover both cases (download via public archive and download via api)

thufschmitt (approved):
Thanks, looks good :)

And thanks for improving the readability of the github test too :)

(Ideally it'd be nice to serve `github.com` and `api.github.com` separately in it, but that's out of scope here)

Kha:
> To mitigate this only ever use the github api if we actually have a token.

This doesn't seem to be true, the `getRevFromRef` function just above `getDownloadUrl` still uses the API. I just ran into a rate limit error from that. Not sure there's a better way to avoid that than `git ls-remote`.

Mic92:
At least this lets you download archives, which is more often used in CIs than updating flakes.

PR #7043, opened 2022-09-14
Add an installer test

edolstra:
This is loosely based on the Nix installer matrix (https://github.com/grahamc/nix-install-matrix), but it runs inside a Nix build using QEMU rather than outside of it using Vagrant. It also doesn't rely on import-from-derivation, unlike the old installer test that ran inside a declaratively built RedHat/Debian image.


tests/installer/default.nix
```diff
@@ -0,0 +1,221 @@
...
+        hash = "sha256-iUUXyRY8iW7DGirb0zwGgf1fRbLA7wimTJKgP7l/OQ8=";
+      };
+      rootDisk = "box-disk1.vmdk";
+      system = "x86_64-linux";
+    };
+    */
+
+    "ubuntu-16-04" = {
+      image = import <nix/fetchurl.nix> {
+        url = https://app.vagrantup.com/generic/boxes/ubuntu1604/versions/4.1.12/providers/libvirt.box;
```
cole-h, line 50: Should all these plain URLs be quoted (especially due to RFC45: https://github.com/NixOS/rfcs/pull/45)?

tests/installer/default.nix
```diff
@@ -0,0 +1,221 @@
...
+          [[ \$(cat \$out) = foobar ]]
+        EOF
+
+        echo "Done!"
+        touch $out
+      '';
+
+in
+
+builtins.mapAttrs (imageName: image:
```
Hoverbear, line 216: It would be quite convenient to have an `all` task that ran them for a specific architecture...

tests/installer/default.nix
Hoverbear: I wasn't able to get it working with a plain string: 

```
❯  nix build .#hydraJobs.installerTests.rhel-8.x86_64-linux.install-force-daemon --rebuild -L
warning: Git tree '/home/ana/git/nixos/nix' is dirty
error: some outputs of '/nix/store/m54sb60nb2nnrmfmfp5ai2dakc54ry5c-installer-test-rhel-8-install-force-daemon.drv' are not valid, so checking is not possible
```

cole-h: If you try without `--rebuild`, it should succeed.

tests/installer/default.nix
```diff
@@ -0,0 +1,221 @@
...
+          if $ssh -- true; then
+            echo "[ssh] Connected!"
+            break
+          fi
+          if ! kill -0 $qemu_pid; then
+            echo "qemu died unexpectedly"
+            exit 1
+          fi
+          sleep 1
+        done
```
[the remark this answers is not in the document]
bjornfor: How do you continue from there? I tried `vagrant up`, but it requires libvirt. I tried to run libvirtd inside a derivation and got into trouble[1]. There's https://github.com/billyan2018/vagrant-qemu which looks interesting though.

[1] After fixing the unix socket path in the Vagrantfile

```
Vagrant.configure("2") do |config|
  config.vm.box = "testVM"
  config.vm.provider :libvirt do |libvirt|
    libvirt.uri = "qemu+unix:///session?socket=/run/user/$UID/libvirt/libvirt-sock"
  end
end
```

libvirt failed with some permission / path errors.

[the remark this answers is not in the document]
edolstra: Pulling in vagrant and libvirt adds a lot of complexity compared to calling qemu directly that I'd rather avoid...

PR #7046, opened 2022-09-14
dockerImage: fix root shell

mkenigs:
Currently root's shell is set to a path that does not exist; this change sets it to the correct path to bash


thufschmitt (approved):
Looks good, thanks!

PR #7048, opened 2022-09-15
Guard uses of lutimes, for portability

### LINKED ISSUE #7045: lutimes not optional
**Describe the bug**

Although [configure.ac](https://github.com/NixOS/nix/blob/master/configure.ac#L81-L83) states that the lutimes function is optional, I can't build nix on a system without it. Compilation fails with
```
src/libutil/filesystem.cc:76:13: error: use of undeclared identifier 'lutimes'; did you mean 'utimes'?
        if (lutimes(link.c_str(), times))
            ^~~~~~~
            utimes
```

**Steps To Reproduce**

1. try to build nix on a system without lutimes

**Expected behavior**

It should just work around the missing lutimes, like it's done in [some parts](https://github.com/NixOS/nix/blob/master/src/libstore/local-store.cc#L568) of the code already.

**Additional context**

I'm currently trying to package nix for termux, which does not have lutimes at the moment, see [this issue](https://github.com/termux/termux-packages/issues/6176)

puffnfresh:
Fixes #7045

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -41,21 +41,6 @@ bool isCacheFileWithinTtl(time_t now, const struct stat & st)
     return st.st_mtime + settings.tarballTtl > now;
 }
 
-bool touchCacheFile(const Path & path, time_t touch_time)
-{
-#ifndef _WIN32 // TODO implement
-    struct timeval times[2];
-    times[0].tv_sec = touch_time;
-    times[0].tv_usec = 0;
-    times[1].tv_sec = touch_time;
-    times[1].tv_usec = 0;
-
-    return lutimes(path.c_str(), times) == 0;
-#else
-    return false;
-#endif
-}
-
 Path getCachePath(std::string_view key, bool shallow)
 {
     return getCacheDir()
@@ -594,8 +579,11 @@ struct GitInputScheme : InputScheme
                     warn("could not update local clone of Git repository '%s'; continuing with the most recent version", repoInfo.url);
                 }
 
-                if (!touchCacheFile(localRefFile, now))
-                    warn("could not update mtime for file '%s': %s", localRefFile, strerror(errno));
+                try {
+                    setWriteTime(localRefFile, now, now);
+                } catch (Error & e) {
+                    warn("could not update mtime for file '%s': %s", localRefFile, e.msg());
+                }
                 if (!originalRef && !storeCachedHead(repoInfo.url, ref))
                     warn("could not update cached head '%s' for '%s'", ref, repoInfo.url);
             }
--- a/src/libutil/file-system.hh
+++ b/src/libutil/file-system.hh
@@ -157,6 +157,30 @@ inline void createDirs(PathView path)
 }
 
 /**
+ * Set the access and modification times of the given path, not
+ * following symlinks.
+ *
+ * @param accessTime Specified in seconds.
+ *
+ * @param modificationTime Specified in seconds.
+ *
+ * @param isSymlink Whether the file in question is a symlink. Used for
+ * fallback code where we don't have `lutimes` or similar. if
+ * `std::optional` is passed, the information will be recomputed if it
+ * is needed. Race conditions are possible so be careful!
+ */
+void setWriteTime(
+    const std::filesystem::path & path,
+    time_t accessedTime,
+    time_t modificationTime,
+    std::optional<bool> isSymlink = std::nullopt);
+
+/**
+ * Convenience wrapper that takes all arguments from the `struct stat`.
+ */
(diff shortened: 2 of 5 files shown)


src/libfetchers/git.cc
```diff
@@ -33,13 +33,17 @@ bool isCacheFileWithinTtl(const time_t now, const struct stat & st)
...
+#if HAVE_LUTIMES
   struct timeval times[2];
   times[0].tv_sec = touch_time;
   times[0].tv_usec = 0;
   times[1].tv_sec = touch_time;
   times[1].tv_usec = 0;
 
   return lutimes(path.c_str(), times) == 0;
+#else
+  return false;
```
puffnfresh, line 45: Emits a warning when `false` but I have assumed that's acceptable.

src/libutil/filesystem.cc
```diff
@@ -76,6 +77,7 @@ void createSymlink(const Path & target, const Path & link,
         if (lutimes(link.c_str(), times))
             throw SysError("setting time of symlink '%s'", link);
     }
+#endif
```
thufschmitt, line 80: That guard (and the one below which acts on essentially duplicated code :/ ) is a bit annoying because it changes the semantics and affects reproducibility. Do you know whether there's a way to emulate `lutimes` if it's absent?

Or maybe we can just ignore setting the time for the symlinks if `lutimes` isn't available and pray that it doesn't matter in practice

thufschmitt (changes requested):
Mh, that's sad (and apologies, at least some of these are my doing).

Dropping these is annoying because it can make things non-reproducible (and in a way that Nix might very much not like). Not sure whether there's a good way around this

src/libutil/filesystem.cc
Et7f3: > futimes() is available since glibc 2.3. lutimes() is available since glibc 2.6, and is implemented using the [utimensat](https://linux.die.net/man/2/utimensat)(2) system call, which is supported since kernel 2.6.22.

Et7f3: @jugendhacker Does termux has the mentioned function ?

Ericson2314:

- @thufschmitt: `lutimes` is a syscall that's not available everywhere 
    - its for getting file timestamps while observing symlinks
    - this is relevant for termux
- @edolstra: don't really believe in `autoconf` checks and its philosophy
    - should just have conditionals
- assigned to @Ericson2314 to fix the `autoconf` checks           
  - related: https://github.com/NixOS/nix/issues/7045

Ericson2314:
https://github.com/NixOS/nix/pull/8504 removed the dead code, now there are just two `lutimes` to deduplicate left.

fricklerhandwerk:
This should be good to go, @thufschmitt @Ericson2314 do you want to take another look?

src/libutil/file-system.hh
```diff
@@ -165,6 +166,15 @@ inline Paths createDirs(PathView path)
     return createDirs(Path(path));
 }
 
+/**
+ * Set the access and modification times of the given path, not
+ * following symlinks.
+ *
+ * Whether or not the path is a symlink is taken from the mode of `st`
+ * in fallback code.
```
thufschmitt, line 174: I don't think that precision really makes sense without the context of that PR (what's “fallback mode”?)

Besides that, I don't think we should trust `st` to tell us whether the file is a symlink or not. Barring possible race conditions (which I think we can't avoid any way), there's no guaranty that `st` refers to the path at `path` (might have been a totally made up struct just to set the time, or taken from another file). We should rather re-check within the function ([std::filesystem::is_symlink](https://en.cppreference.com/w/cpp/filesystem/is_symlink) being our friend here I guess)

thufschmitt (changes requested):

src/libutil/file-system.hh
Ericson2314: fallback mode is without `lutimes`, with just `utimes` (yes, this should be reworded)

> Besides that, I don't think we should trust `st` to tell us whether the file is a symlink or not. 

Note that this is not a new thing. I agree in principle, but how would you update `canonicaliseTimestampAndPermissions` in practice? It is, already, very geared about working from a prior stat (in a context that we don't think things would change without them looking.

Unless we want to scope creep and overhaul `canonicaliseTimestampAndPermissions`, I think it would be good to just document the race condition on this function and say callers must be confident `st` is up to date.

Ericson2314: OK this is done now. the `struct stat` way still exists, but is not forced upon eveyone else.

Ericson2314 (approved):
@thufschmitt's request is addressed. All the compat is a bit gnarly, but future improvements are possible.

PR #7052, opened 2022-09-16
libexpr/fetchurl.nix: allow __impure fetch

amjoseph-nixpkgs:
This commit adds an optional `__impure` parameter to fetchurl.nix, which allows the caller to use `libfetcher`'s fetcher in an impure derivation.  This allows nixpkgs' patch-normalizing fetcher (fetchpatch) to be rewritten to use nix's internal fetchurl, thereby eliminating the awkward "you can't use fetchpatch here" banners scattered all over the place.

See also: https://github.com/NixOS/nixpkgs/pull/188587

### DIFF
--- a/src/libexpr/fetchurl.nix
+++ b/src/libexpr/fetchurl.nix
@@ -12,7 +12,7 @@
 , executable ? false
 , unpack ? false
 , name ? baseNameOf (toString url)
-, __impure ? false
+, impure ? false
 }:
 
 derivation ({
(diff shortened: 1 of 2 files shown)


src/libexpr/fetchurl.nix
```diff
@@ -12,13 +12,13 @@
 , executable ? false
 , unpack ? false
 , name ? baseNameOf (toString url)
+, __impure ? false
```
edolstra, line 15:
```suggestion
, impure ? false
```

The `__` prefix is appropriate for builtins like `derivation` to prevent them from clashing with user attributes, but that's not a danger here.

edolstra:
Can you add this to the release notes?

Issue #7064, opened 2022-09-19
Error: schema is corrupt


squalus:
**Describe the bug**

I have seen multiple cases of "schema is corrupt" error messages in a production environment. This tends to happen on NixOS systems that have unexpected power cuts.

```
$ nix store verify --all
error: '/mnt/nix/var/nix/db/schema' is corrupt
```

In this case, it's an ext4 file system and the schema file is empty.

**Steps To Reproduce**

I have a minimal test case that simulates a power cut with NixOS tests and reproduces the problem here: https://github.com/squalus/nix-durability-tests. It can be run on several different file system.

```
nix -L build github:squalus/nix-durability-tests#corrupt-schema-tests.xfs
```

This will hopefully print a "schema is corrupt" error message.

**Expected behavior**

The schema file should never be invalid, even if there's an unexpected power cut.

**`nix-env --version` output**
`nix-env (Nix) 2.8.1`

**Additional context**

Some possible causes:

  1. Errors from `close(2)` are ignored in `nix::writeFile`. (From `man close`: Failing to check the return value when closing a file may lead to silent loss of data.)
  2. `fsync(2)` is not run on the file after writing the contents. This means the data may not be fully flushed to disk.
  3. `fsync(2)` is not run on the parent directory after closing the file. This means the directory may have outdated contents. (This wouldn't cause an empty file, but it could cause a mismatch. I haven't yet observed this problem.)
  4. The file is not written atomically. It could instead be written with a temporary file and a call to `rename(2)`, like in https://github.com/google/renameio.

Point 2 was addressed in this PR, but it was never merged: https://github.com/NixOS/nix/pull/1956

squalus:
#7065 takes care of 1-3, but I'll keep this open because the atomic file write (point 4) could still be done to improve this.

PR #7065, opened 2022-09-19
Improve durability of schema version file writes

squalus:
This makes some durability improvements to the schema version file writes and fixes the "schema is corrupt" errors from the test case described in #7064.

I haven't tested on Darwin yet.

- call close explicitly in writeFile to prevent the close exception from being ignored
- fsync after writing schema file to flush data to disk
- fsync schema file parent to flush metadata to disk

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -158,7 +158,7 @@ void migrateCASchema(SQLite& db, Path schemaPath, AutoCloseFD& lockFd)
             txn.commit();
         }
 
-        writeFile(schemaPath, fmt("%d", nixCASchemaVersion));
+        writeFile(schemaPath, fmt("%d", nixCASchemaVersion), 0666, true);
         lockFile(lockFd.get(), ltRead, true);
     }
 }
@@ -281,7 +281,7 @@ LocalStore::LocalStore(const Params & params)
     else if (curSchema == 0) { /* new store */
         curSchema = nixSchemaVersion;
         openDB(*state, true);
-        writeFile(schemaPath, (format("%1%") % nixSchemaVersion).str());
+        writeFile(schemaPath, (format("%1%") % nixSchemaVersion).str(), 0666, true);
     }
 
     else if (curSchema < nixSchemaVersion) {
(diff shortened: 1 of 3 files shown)


src/libutil/util.cc
```diff
@@ -364,10 +364,13 @@ void writeFile(const Path & path, std::string_view s, mode_t mode)
         e.addTrace({}, "writing file '%1%'", path);
         throw;
     }
+    if (sync)
+        fd.fsync();
+    fd.close();
```
thufschmitt, line 369: I thinks that's worth a comment explaining why it's explicitly closed

src/libstore/local-store.cc
```diff
@@ -158,7 +158,8 @@ void migrateCASchema(SQLite& db, Path schemaPath, AutoCloseFD& lockFd)
             txn.commit();
         }
 
-        writeFile(schemaPath, fmt("%d", nixCASchemaVersion));
+        writeFile(schemaPath, fmt("%d", nixCASchemaVersion), 0666, true);
+        syncParent(schemaPath);
```
thufschmitt, line 162: Shouldn't that be done automatically by `writeFile` when `sync` is set to true instead?

thufschmitt (changes requested):
Thanks,

A couple small suggestions but looks good overall.

Note that the CI is failing on darwin (doesn't like `F_FULLSYNC`. Maybe a missing `#include`?)

src/libstore/local-store.cc
squalus: That makes sense. Changed `writeFile` to call `syncParent`.

src/libutil/util.cc
squalus: Added an extra comment here.

squalus:
> Note that the CI is failing on darwin (doesn't like `F_FULLSYNC`. Maybe a missing `#include`?)

Oops, it's `F_FULLFSYNC`. Fixed the typo and tested a build on Darwin.

thufschmitt (approved):
Great, thanks :)

PR #7066, opened 2022-09-19
manual: architecture overview

fricklerhandwerk:
this continues the effort from #6420, which originally began with #4280.


abathur:
Aside: will there be a ~style guide with notes on some of the substitutions like action->task and perform->run? (Unclear from casual observation if these are just cases where you're being consistent about choosing simpler words or if this is more about using a consistent, precise vocabulary.)

fricklerhandwerk:
Good question. I think this is not yet worth a firm rule. Here it's just about finding more appropriate terminology and being consistent about it.

If people find it reasonable we can just continue using it, and watching out in reviews to *keep* it consistent.

We discussed this up and down in the making of #6420, and I'm fairly convinced now that this is both correct *and* appealing to common sense. Of course still open to constructive disagreement!

abathur:
Fine answers for now :)

The consistency of the edits just made me curious.

doc/manual/src/architecture/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+[purely functional]: https://en.m.wikipedia.org/wiki/Purely_functional_programming
+
+Command line interface and Nix language are what users interact with most.
+
+> **Note**
+> The Nix language itself does not have a notion of *packages* or *configurations*.
+> As far as we are concerned here, the inputs and results of a build plan are just data.
+
+Underlying these is the [Nix store](../glossary.md#gloss-store), a mechanism to keep track of build plans, data, and references between them.
+It can also execute build plans to produce new data.
```
bryanhonof, line 46:
```suggestion
Underlying the command line interface and the Nix language is the [Nix store](../glossary.md#gloss-store), a mechanism to keep track of build plans, data, and references between them.
It can also execute build plans to produce new data.
```

My tiny brain already forgot what came before... 😅

doc/manual/src/architecture/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+At the top is the [command line interface](../command-ref/command-ref.md), translating from invocations of Nix executables to interactions with the underlying layers.
+
+Below that is the [Nix language](../language/index.md), a [purely functional] configuration language.
+It is used to compose expressions which ultimately evaluate to self-contained *build plans*, used to derive *build results* from referenced *build inputs*.
+
+[purely functional]: https://en.m.wikipedia.org/wiki/Purely_functional_programming
+
+Command line interface and Nix language are what users interact with most.
```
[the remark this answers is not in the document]
fricklerhandwerk:
```suggestion
The command line interface and the Nix language are what users interact with most.
```

doc/manual/src/architecture/architecture.md
fricklerhandwerk: Probably has nothing to do with your brain specifically.

doc/manual/src/architecture/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+The command line interface and the Nix language are what users interact with most.
+
+> **Note**
+> The Nix language itself does not have a notion of *packages* or *configurations*.
+> As far as we are concerned here, the inputs and results of a build plan are just data.
+
+Underlying the command line interface and the Nix language is the [Nix store](../glossary.md#gloss-store), a mechanism to keep track of build plans, data, and references between them.
+It can also execute build plans to produce new data.
+
+A build plan is a series of *build tasks*.
```
asymmetric, line 51: I find it a bit counterintuitive that the first term introduced is "build task", only then for the term to be redefined in a "boxed paragraph".

Also, by introducing this term ("build task"), I think you're increasing the cognitive load, as it is not going to be useful in the broader Nix context.

My suggesiton is instead 

```suggestion
A build plan is a series of "build tasks", which in Nix are called [derivations](../glossary#gloss-derivation).
```

doc/manual/src/architecture/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+|                         evaluates to             |            |
+|                              |                   V            |
+| +----------------------------|------------------------------+ |
+| | store                      |                              | |
+| |            referenced by   V       builds                 | |
+| | .-------------.      .------------.      .--------------. | |
+| | | build input |----->| build plan | ---->| build result | | |
+| | '-------------'      '------------'      '--------------' | |
+| +-----------------------------------------------------------+ |
++---------------------------------------------------------------+
```
asymmetric, line 34: Not sure if I'm misunderstanding, but `referenced by` seems to go in the opposite direction of the arrow.

SuperSandro2000:
Could we use mermaid to draw the diagrams instead of manually doing them?

fricklerhandwerk:
@SuperSandro2000 I tried that first, since *of course it's insane to maintain ASCII diagrams* but unfortunately it doesn't work:
- Mermaid doesn't offer enough control over presentation. It very quickly looks unnecessarily messy.
- There is no ASCII rendering for Mermaid, which we need for terminal output (`man` and `--help`), and I can't imagine how that would be less messy if it existed.
- Mermaid requires adding a huge closure the manual build to display the diagrams at all.

`svgbob` renders these things good enough, and allows us to have both presentations. Using it is optional since just having the ASCII is already better than no visualisation at all. We can always upgrade to a more convenient method once we find one.

@edolstra said he doesn't like the brittle custom syntax, and I concur with a smirk that

> It's like markdown brought to its bitter conclusion.

I'm agnostic to the tools we use, as long as they get the job done. Right now I don't see a meaningful alternative. A few considerations:

- LaTeX could work, but this is only slightly less cumbersome and needs tons of specialised knowledge to maintain.
- Building something with some SVG or diagramming tool and committing both the source and the final rendering is possible in principle. But it makes the contribution process really cumbersome, due to the manual steps, requires contributors to install and learn using a very specific tool, and I know none that would give us ASCII output as well.

Do you have other suggestions?

doc/manual/src/architecture/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+Its expressions ultimately evaluate to self-contained *build plans*, used to derive *build results* from referenced *build inputs*.
+
+The command line interface and the Nix language are what users interact with most.
+
+> **Note**
+> The Nix language itself does not have a notion of *packages* or *configurations*.
+> As far as we are concerned here, the inputs and results of a build plan are just data.
+
+Underlying the command line interface and the Nix language is the [Nix store](../glossary.md#gloss-store), a mechanism to keep track of build plans, data, and references between them.
+It can also execute build plans to produce new data.
```
edolstra, line 49: "It" here seems to refer to the Nix store, which cannot be said to execute build plans.

doc/manual/src/architecture/architecture.md
edolstra: Yes, maybe it's better to use "derivation" instead of "build task", which is not a term that appears in the Nix source.

doc/manual/src/architecture/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+[hierarchical layers]: https://en.m.wikipedia.org/wiki/Multitier_architecture#Layers
+
+```
++---------------------------------------------------------------+
+| Nix             .-------------------------.                   |
+|                 | commmand line interface |------.            |
+|                 '-------------------------'      |            |
+|                              |                   |            |
+|                          evaluates               |            |
```
edolstra, line 18: This suggests that the CLI layer evaluates the Nix language, but that's the job of the `libexpr` layer.

doc/manual/src/architecture/architecture.md
edolstra, line 34: It would be good if this picture includes the names of the components: `libstore`, `libexpr`, `libcmd / src/nix`, ...

fricklerhandwerk: Build inputs are referenced by build tasks.

Yes, the arrow directions can be ambiguous. As we discussed in yesterday's maintainer meeting, the point of these diagrams is to convey the right general idea in order to help shape the first approximation of a mental
model, not to be mathematically precise.

There is not enough space in such a diagram and not enough capacity on the reader's end to capture all the details, and this introductory document is not meant to hold that information.

The necessary details will follow follow-up PRs adding sub-chapters, which will be linked from the respective terms.

doc/manual/src/architecture/architecture.md
fricklerhandwerk: We still don't have a definition of what exactly a derivation even is. I'd argue the term derivation should be reserved for Nix's specific implementation of build tasks, whatever it is we define it to be - store derivation or the Nix language concept.

doc/manual/src/architecture/architecture.md
asymmetric: Can’t we just invert the arrow and say “build inputs reference build tasks”?

asymmetric: Alright, just wanted to double check. Consider my question answered. Thanks!

fricklerhandwerk:
Discussed on Nix team meeting on 2022-12-12:

Decision: fix up wording and continue async with merging
  
 - @edolstra: ASCII diagrams are okay for documentation that is not user-facing
      - @fricklerhandwerk: the only reason this is not in the man pages is that no one set it up yet
          - @edolstra: we probably don't want it there either
              - @fricklerhandwerk: why not? having offline documentation doesn't cost us much and can be invaluable in some situations.
      - ability to render ASCII art should not be a blocker to include graphics
          - @fricklerhandwerk: agreed, but it should be kept as consideration. it would be good to include diagrams in the terminal output if we have them, and they are simple enough for it to be realistic
  - @Eriscon2314: summary: we want to have SVG for the HTML rendering, and edit conveniently
      - @edolstra: a declarative format would be ideal, but is not a requirement. could also edit SVG in a graphical editor
          - @fricklerhandwerk: graphical editors produce very noisy output
              - @edolstra: diffability or manual editing is not required, GitHub shows SVG in previews
                  - Rust Book uses plain [SVG which is fairly minimal](https://github.com/rust-lang/book/blob/a60f4316ec923a5ac2ed6a2eba6960edb832d855/src/img/trpl04-03.svg?short_path=7c153e2)
                  - @fricklerhandwerk: that's almost surely touched up by hand
  - @thufschmitt: PlantUML also renders ASCII https://plantuml.com/en/ascii-art
      - @fricklerhandwerk: apparently only for sequence diagrams.
      - @thufschmitt: also requires a huge closure
          - @fricklerhanderk: any tool that does these transformations will; choose your poison
  - decision: leave the ASCII for now, consider moving to SVG later
  - @edolstra: which level of architecture is supposed to be documented here?
      - @Ericson2314: it's more user-oriented and is supposed to show behavior
      - @fricklerhandwerk: will rename configuration language to evaluator
  - @edolstra: graph arrows and labels are not entirely clear about what they mean
      - @fricklerhandwerk: the point is to have a general overview, and the subsections will go into details on each term and concept
  - (some more discussion over wording and intent)
  - decision: fix up wording and continue async with merging
  - assigned to @edolstra
    
</details>

PR #7070, opened 2022-09-20
nix repl: warn if creating dir for history fails

alyssais:
The history is not critical to the functionality of nix repl, so it's enough to warn here, rather than refuse to start if the directory Nix thinks the history should live in can't be created.


thufschmitt (approved):
Indeed, no reason to fail here. Thanks!

PR #7078, opened 2022-09-22
"valid signature" -> "signature by a trusted key"

Ericson2314:
I just had a colleague get confused by the previous phrase for good reason. "valid" sounds like an *objective* criterion, e.g. and *invalid signature* would be one that would be trusted by no one, e.g. because it misformatted or something.

What is actually going is that there might be a signature which is perfectly valid to *someone else*, but not to the user, because they don't trust the corresponding public key. This is a *subjective* criterion, because it depends on the arbitrary and personal choice of which public keys to trust.

~~I therefore think "trustworthy" is a better adjective to use. Whether something is worthy of trust is clearly subjective, and then "trust" within that word nicely evokes `trusted-public-keys` and friends.~~

"signature by a trusted key" is unambiguous.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -560,9 +560,15 @@ public:
         R"(
           If set to `true` (the default), any non-content-addressed path added
           or copied to the Nix store (e.g. when substituting from a binary
-          cache) must have a trustworthy signature, that is, be signed using one of
-          the keys listed in `trusted-public-keys` or `secret-key-files`. Set
-          to `false` to disable signature checking.
+          cache) must have a signature by a key we trust. A trusted key is one
+          listed in `trusted-public-keys`, or a public key counterpart to a
+          private key stored in a file listed in `secret-key-files`.
+          
+          Set to `false` to disable signature checking and trust all
+          non-content-addressed paths unconditionally.
+          
+          (Content-addressed paths are inherently trustworthy and thus
+          unaffected by this configuration option.)
         )"};
 
     Setting<StringSet> extraPlatforms{
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -560,7 +560,7 @@ public:
         R"(
           If set to `true` (the default), any non-content-addressed path added
           or copied to the Nix store (e.g. when substituting from a binary
-          cache) must have a signature by a key we trust. A trusted key is one
+          cache) must have a signature by a trusted key. A trusted key is one
           listed in `trusted-public-keys`, or a public key counterpart to a
           private key stored in a file listed in `secret-key-files`.
           
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -560,7 +560,7 @@ public:
         R"(
           If set to `true` (the default), any non-content-addressed path added
           or copied to the Nix store (e.g. when substituting from a binary
-          cache) must have a valid signature, that is, be signed using one of
+          cache) must have a trustworthy signature, that is, be signed using one of
           the keys listed in `trusted-public-keys` or `secret-key-files`. Set
           to `false` to disable signature checking.
         )"};
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -751,7 +751,7 @@ void LocalStore::registerDrvOutput(const Realisation & info, CheckSigsFlag check
     if (checkSigs == NoCheckSigs || !realisationIsUntrusted(info))
         registerDrvOutput(info);
     else
-        throw Error("cannot register realisation '%s' because it lacks a trustworthy signature", info.outPath.to_string());
+        throw Error("cannot register realisation '%s' because it lacks a signature by a trusted key", info.outPath.to_string());
 }
 
 void LocalStore::registerDrvOutput(const Realisation & info)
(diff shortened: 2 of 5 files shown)


fricklerhandwerk:
I agree with the reasoning, it just reads clunky. Would it be wrong to say "trusted signature"? That would align even more nicely, and read more fluently.

Ericson2314:
I didn't want to use "trusted" because I was worried that people might think "trusted signatures" and "trusted public keys" are two independent "entry points" of trust. Well-formed Signatures themselves are always "trusted" in the sense that in

> trusted signature & trusted public key => trusted store object

the signature is trivially trusted, and trust flows from the public key to the store object if the signature and public key match.

`trustworthy` vs `trusted` was a (perhaps feeble) way to indicate public keys and signatures work differently.

Ericson2314:
Maybe I am overthinking the English :).

src/libstore/local-store.cc
```diff
@@ -751,7 +751,7 @@ void LocalStore::registerDrvOutput(const Realisation & info, CheckSigsFlag check
     if (checkSigs == NoCheckSigs || !realisationIsUntrusted(info))
         registerDrvOutput(info);
     else
-        throw Error("cannot register realisation '%s' because it lacks a valid signature", info.outPath.to_string());
+        throw Error("cannot register realisation '%s' because it lacks a trustworthy signature", info.outPath.to_string());
```
Ericson2314, line 754:
```suggestion
        throw Error("cannot register realisation '%s' because it lacks a signature by a trusted public key", info.outPath.to_string());
```

src/libstore/local-store.cc
```diff
@@ -1266,7 +1266,7 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
     RepairFlag repair, CheckSigsFlag checkSigs)
 {
     if (checkSigs && pathInfoIsUntrusted(info))
-        throw Error("cannot add path '%s' because it lacks a valid signature", printStorePath(info.path));
+        throw Error("cannot add path '%s' because it lacks a trustworthy signature", printStorePath(info.path));
```
Ericson2314, line 1269:
```suggestion
        throw Error("cannot add path '%s' because it lacks a signature by a trusted public key", printStorePath(info.path));
```

tests/signing.sh
```diff
@@ -81,7 +81,7 @@ info=$(nix path-info --store file://$cacheDir --json $outPath2)
 [[ $info =~ 'cache1.example.org' ]]
 [[ $info =~ 'cache2.example.org' ]]
 
-# Copying to a diverted store should fail due to a lack of valid signatures.
+# Copying to a diverted store should fail due to a lack of trustworthy signatures.
```
Ericson2314, line 84:
```suggestion
# Copying to a diverted store should fail due to a lack of signatures by trusted public keys.
```

src/nix/make-content-addressed.md
```diff
@@ -22,7 +22,7 @@ R""(
 
   ```console
   # nix copy --to /tmp/nix --trusted-public-keys '' nixpkgs#hello
-  cannot add path '/nix/store/zy9wbxwcygrwnh8n2w9qbbcr6zk87m26-libunistring-0.9.10' because it lacks a valid signature
+  cannot add path '/nix/store/zy9wbxwcygrwnh8n2w9qbbcr6zk87m26-libunistring-0.9.10' because it lacks a trustworthy signature
```
Ericson2314, line 25:
```suggestion
  cannot add path '/nix/store/zy9wbxwcygrwnh8n2w9qbbcr6zk87m26-libunistring-0.9.10' because it lacks a signature by a trusted public key
```

src/libstore/globals.hh
```diff
@@ -560,7 +560,7 @@ public:
         R"(
           If set to `true` (the default), any non-content-addressed path added
           or copied to the Nix store (e.g. when substituting from a binary
-          cache) must have a valid signature, that is, be signed using one of
+          cache) must have a trustworthy signature, that is, be signed using one of
           the keys listed in `trusted-public-keys` or `secret-key-files`. Set
           to `false` to disable signature checking.
```
Ericson2314, line 565:
```suggestion
          cache) must have a signature by a key we trust. A trusted key is one
          listed in `trusted-public-keys`, or the public key counterpart to the
          private key listed in `secret-key-files`.
          
          Set to `false` to disable signature checking and trust all non-content-addressed paths unconditionally.
          
          (Content-addressed paths are inherently trustworthy and thus unaffected by this configuration option.)
```

Ericson2314:
3542d4fe16822bcd94433bf9393fbe17a5297948 is prior art.

src/nix/verify.cc
```diff
@@ -41,7 +41,7 @@ struct CmdVerify : StorePathsCommand
         addFlag({
             .longName = "sigs-needed",
             .shortName = 'n',
-            .description = "Require that each path has at least *n* valid signatures.",
+            .description = "Require that each path has is signed by *n* different keys.",
```
runeksvendsen, line 44: I guess it should still say "at least" here? Also the "has" from the previous sentence is still there.

```suggestion
            .description = "Require that each path is signed by at least *n* different keys.",
```

runeksvendsen (changes requested):
Thank you for writing this up @Ericson2314

src/libstore/globals.hh
```diff
@@ -560,9 +560,15 @@ public:
         R"(
           If set to `true` (the default), any non-content-addressed path added
           or copied to the Nix store (e.g. when substituting from a binary
-          cache) must have a valid signature, that is, be signed using one of
-          the keys listed in `trusted-public-keys` or `secret-key-files`. Set
-          to `false` to disable signature checking.
+          cache) must have a signature by a key we trust. A trusted key is one
```
fricklerhandwerk, line 563:
```suggestion
          cache) must have a signature from a trusted key. A trusted key is one
```

Ericson2314:
Thanks, both of you!

fricklerhandwerk (approved):

PR #7079, opened 2022-09-22
Allow pass max-silent-time and build-poll-interval to daemon untrusted

matthewbauer:
These settings seem harmless, they control the same polling functionality that timeout does, but with different behavior. Should be safe for untrusted users to pass in.


thufschmitt (approved):
Looks good, thanks :)

PR #7082, opened 2022-09-23
add ccacheStdenv

rapenne-s:
when using ccache, rebuild time has been measured
89% faster while not slowing the speed of cold builds

Using `make -j 6` with an i7-1165G7, I got the following results:
- ccache first build: 428 seconds
- ccache rebuild: 48s (= 89% speedup)
- no ccache build: 426 seconds
- another no ccache build: 424 seconds

Just use `nix develop .#ccacheStdenv` and you have ccache enabled, using GCC. A bit more work could be done to have a clang ccache if desired.


thufschmitt (approved):
Thanks, looks good :)

PR #7089, opened 2022-09-24
Mention `--accept-flake-config` in the related warning

### LINKED ISSUE #7086: need a way to trust flake nixConfig -- `nix flake trust` or `nix flake metadata --trust`

Yes. I want to use nixConfig from my flake in CI. I'd like to be able to run a command on initialization that instructions nix to trust my flake nixconfig.

A way to run `nix flake trust` or `nix flake metadata --trust` to automatically opt-in to the settings specified in the current-context's flake.

Not sure.

**Additional context**
n/a

thufschmitt:
Make sure that people who run Nix in non-interactive mode (and so don't have the possibility to interactively accept the individual flake configuration settings) are aware of this flag.

Fix #7086

cc @cole-h

### DIFF
--- a/src/libexpr/flake/config.cc
+++ b/src/libexpr/flake/config.cc
@@ -68,7 +68,7 @@ void ConfigFile::apply()
                 }
             }
             if (!trusted) {
-                warn("ignoring untrusted flake configuration setting '%s'", name);
+                warn("ignoring untrusted flake configuration setting '%s'.\nPass '%s' to trust it", name, "--accept-flake-config");
                 continue;
             }
         }


lheckemann (approved):

cole-h:
(cc @colemickens)

colemickens:
Thank you @thufschmitt for the response with direct links and for the PR! I do see that it's documented, I was looking in the wrong place (ie, the help output for `nix flake` and/or looking for a command that would permanently trust, rather than a flag)

In fact, now I'm wondering if I need to keep supplying the flag to subsequent nix commands..... I almost hope so given the flag-y nature of it... I can figure this out with some experimentation though, I suppose.

Thanks again!

colemickens:
Actually, can we just go ahead and make this message really explicit? Maybe something more along the lines of:

>pass `--accept-flake-config` to permanently trust and use the flake's nixConfig

or

>pass `--accept-flake-config` with each nix invocation to use the flake's nixConfig

depending on which is true? (Though, I'm probably going to file another bug if it's the former...)

PR #7099, opened 2022-09-26
Remove repeat option

### LINKED ISSUE #1911: repeat option is ignored
If I run…
```
#!/usr/bin/env sh
set -e
drv=$(nix-instantiate -E 'with import <nixpkgs> {}; runCommand "test" {} "echo hello | tee $out"')
nix-store -r $drv
nix-store -r $drv --check --repeat 5
```

I get the output:
```
hello
/nix/store/zls05i4bq533y4yh30c68d15qz81zbq5-test
checking outputs of '/nix/store/2yf6gj4qz0wns9bl53hlcjrfmg42nc88-test.drv'...
hello
/nix/store/zls05i4bq533y4yh30c68d15qz81zbq5-test
```

While I expect it to run the build (and print "hello") a total of 7 times.

Deleting the path and running `nix-store -r --repeat 5` runs the build 6 times as expected, but deleting a path isn't always feasible when trying to check determinism.

On a side note, I'm guessing the current absence of a `--check`-like flag for `nix build` is an intentional omission for now?

lheckemann:
This was only functioning if a very narrow combination of conditions held:

- The result path does not yet exist (--check did not result in repeated builds), AND
- The result path is not available from any configured substituters, AND
- No remote builders that can build the path are available.

If any of these do not hold, a derivation will be build 0 or 1 times regardless of the repeat option. Thus, remove it to avoid confusion.

Fixes #1911


Ericson2314:
Good idea. Longer time, I think we could make a new CA-derivation-only repeat that will make much more sense eventually, and it would be far easier to understand. Until then, this is just barely working code cluttering up the build logic.

tomberek:
Seems that `enforce-determinism` is only used in the section deleted and would be lingering as a remaining vestige.

tomberek:
Notes from Nix team meeting:

- @edolstra: think we should keep the repeat option. when we move to CA drvs, we need a way for people to check if their builds are binary reproducible
- @Ericson2314: agree we should keep the feature, but it's badly implemented, as @lheckemann identified. rather rip it out and re-imlement later.
- right now it does something, not clear what's the benefit of ripping it out
- @Ericson2314: it's a reputation issue: people will find something in the manual and then trying it will encounter it to be broken. better not to have it at all then.
    - propose the following heuristic: if we didn't have the feature, how would we prioritise imlementing it?
    - but we're using that feature in Hydra, for example
- @Ericson2314: Hydra is special
- @thufschmitt: removing it is tougher than pretending it never existed. maybe we should try to fix it instead.
- @tomberek: have to ask why @lheckemann bothered to remove it in the first place. agree with @Ericson2314 that removing it may unblock progress
- @thufschmitt: this plays into the problem that we did
- @edolstra: alright, it let's remove it. the PR should update the release notes and make using the option throw an error

## Decision
- remove the feature
- it is okay that this removal will cause an error if the option is completely removed
- update release notes

src/libmain/progress-bar.cc
```diff
@@ -180,10 +180,8 @@ class ProgressBar : public Logger
             auto machineName = getS(fields, 1);
             if (machineName != "")
                 i->s += fmt(" on " ANSI_BOLD "%s" ANSI_NORMAL, machineName);
-            auto curRound = getI(fields, 2);
-            auto nrRounds = getI(fields, 3);
-            if (nrRounds != 1)
-                i->s += fmt(" (round %d/%d)", curRound, nrRounds);
+            getI(fields, 2); // Used to be curRound, but build repetition hasn't really worked in ages
+            getI(fields, 3); // Used to be nrRounds, but ^
```
Ericson2314, line 184: Warn if we get something we don't expect?

src/nix-store/nix-store.cc
```diff
@@ -808,14 +808,13 @@ static void opServe(Strings opFlags, Strings opArgs)
         if (GET_PROTOCOL_MINOR(clientVersion) >= 2)
             settings.maxLogSize = readNum<unsigned long>(in);
         if (GET_PROTOCOL_MINOR(clientVersion) >= 3) {
-            settings.buildRepeat = readInt(in);
+            readInt(in); // this used to be build-repeat, but it never worked anyway
```
Ericson2314, line 811: I think no longer supported `buildRepeat` values should be an error, just as `--repeat` on the CLI will become an error.

src/libmain/progress-bar.cc
Ericson2314: Still on the fence about what to do about that

src/libstore/build/derivation-goal.cc
```diff
@@ -564,12 +560,11 @@ void DerivationGoal::started()
...
         buildMode == bmRepair ? "repairing outputs of '%s'" :
         buildMode == bmCheck ? "checking outputs of '%s'" :
-        nrRounds > 1 ? "building '%s' (round %d/%d)" :
-        "building '%s'", worker.store.printStorePath(drvPath), curRound, nrRounds);
+        "building '%s'", worker.store.printStorePath(drvPath));
     fmt("building '%s'", worker.store.printStorePath(drvPath));
     if (hook) msg += fmt(" on '%s'", machineName);
     act = std::make_unique<Activity>(*logger, lvlInfo, actBuild, msg,
-        Logger::Fields{worker.store.printStorePath(drvPath), hook ? machineName : "", curRound, nrRounds});
+        Logger::Fields{worker.store.printStorePath(drvPath), hook ? machineName : ""});
```
Ericson2314, line 567: Oh hmm should we be doing
```suggestion
        Logger::Fields{worker.store.printStorePath(drvPath), hook ? machineName : "", 1, 1});
```
here for compat with old log readers at the other end?

src/libmain/progress-bar.cc
Ericson2314: I decided this is good, because it is tantamount to asserting that we took a certain control flow in the old code.
```suggestion
            // Used to be `nrRounds` and `curRound`, but build repetition hasn't really worked in ages.
            // `nrRounds` being something other than 1 indicated repetition before, so we assert that it is 1.
            assert(getI(fields, 3), 1);
            // `curRounds` was ignored if `nrRounds` was `1` (no repetition) so we also ignore it.
            getI(fields, 2);
```

Ericson2314 (approved):

src/libstore/build/local-derivation-goal.cc
```diff
@@ -2688,11 +2687,6 @@ DrvOutputs LocalDerivationGoal::registerOutputs()
                 debug("unreferenced input: '%1%'", worker.store.printStorePath(i));
         }
 
-        if (curRound == nrRounds) {
-            localStore.optimisePath(actualPath, NoRepair); // FIXME: combine with scanForReferences()
-            worker.markContentsGood(newInfo.path);
```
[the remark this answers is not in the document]
Ericson2314: (Sorry I missed that!)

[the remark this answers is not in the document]
lheckemann: Fixed!

andir:
Could this be related to a build error I am getting when remote building from Nix 2.11 to 2.13pre (from yesterday) with the following error?

```
Dec 10 01:16:44 ...[2521297]: building '/nix/store/rf2b02am8f6dsn2fkkiggpw63ypbka2y-....drv' on 'ssh://xxx@x.x.x.x'...
Dec 10 01:16:44 ...[2521297]: copying 0 paths...
Dec 10 01:16:44 ...[2521297]: error: client requested enforcing determinism, but this is not currently implemented
Dec 10 01:16:44 ...[2521297]: error: unexpected end-of-file
Dec 10 01:16:44 ...[2521297]: error: builder for '/nix/store/rf2b02am8f6dsn2fkkiggpw63ypbka2y-....drv' failed with exit code 1;
Dec 10 01:16:44 ...[2521297]:        last 1 log lines:
Dec 10 01:16:44 ...[2521297]:        > error: client requested enforcing determinism, but this is not currently implemented
```

In my case this is a systemd job that calls `nix build .#somePackage` and the system has a remote builder configured 

```
cat $ /etc/nix/machines
ssh://xxx@xx.xx.xx.xx x86_64-linux /root/.ssh/id_ed25519 8 1 nixos-test,kvm,big-parallel,benchmark - -
```

andir:
Yeah, this seems to be it:
```
nix-repl> p = derivation { name = "foo";  system = builtins.currentSystem; builder = "/bin/sh";  args = []; }          

nix-repl> :b p
error: unexpected end-of-file
error: builder for '/nix/store/9d794zbx10n43fa6zaqf33pfymdrm4r2-foo.drv' failed with exit code 1;
       last 1 log lines:
       > error: client requested enforcing determinism, but this is not currently implemented 1
       For full logs, run 'nix log /nix/store/9d794zbx10n43fa6zaqf33pfymdrm4r2-foo.drv'.
```

The setting `enforceDeterminism` used to be `true` by default (thus sending a != 0 value over the wire). This means you probably want to ignore that flag unless `nrRounds` is `> 0` as well?

https://github.com/NixOS/nix/pull/7099/files#diff-9412152b7224a524f972b70909301342567c69c9f391625677764859063c48ebL580-L583

PR #7109, opened 2022-09-28
__allowNetwork when sandbox==relaxed

jcumming:
This patch allows for derivations to be build with privateNetworking distabled when sandbox == relaxed, much in the same way as __noChroot works.

This is motivated by the desire to run EDA tools that need network access to checkout license tokens, but still have filesystem and process isolation from the host.

I'm not sure this is the *best* solution, and the builders have no resolver setup (which is maybe a good thing?).

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -379,9 +379,13 @@ void LocalDerivationGoal::startBuilder()
     /* Are we doing a chroot build? */
     {
         auto noChroot = parsedDrv->getBoolAttr("__noChroot");
+        auto allowNetwork = parsedDrv->getBoolAttr("__allowNetwork");
         if (settings.sandboxMode == smEnabled) {
             if (noChroot)
                 throw Error("derivation '%s' has '__noChroot' set, "
+                    "but that's not allowed when 'sandbox' is 'true'", worker.store.printStorePath(drvPath));
+            if (allowNetwork)
+                throw Error("derivation '%s' has '__allowNetwork' set, "
                     "but that's not allowed when 'sandbox' is 'true'", worker.store.printStorePath(drvPath));
 #if __APPLE__
             if (additionalSandboxProfile != "")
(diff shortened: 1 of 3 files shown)


Ericson2314:
Could be a perma-unstable feature too, FWIW.

Ericson2314 (approved):

newAM (approved):
I have the same problem (working with closed source software that needs to talk to license servers), and this looks like a great solution!

@Ericson2314 are you a maintainer for `nix`?  What are the next steps to get this merged?

Ericson2314:
I do not have merge access, sorry. See how it is enqueued on a GitHub project board. Hopefully that means it will receive more attention soon.

edolstra:
Maybe this use case could use impure derivations (which already exist as an experimental feature and which IIRC allow network access)?

Ericson2314:
@jcumming Have you tried using the new experimental impure derivations? There was some interest in that being used instead of this escape hatch for this.

newAM:
I did not know about impure derivations!

For my usecase impure derivations worked for accessing license servers, but only on NixOS where the `impure-derivations` and `ca-derivations` experimental-features are available (the [wiki has a TODO](https://nixos.wiki/wiki/Ca-derivations) and my own attempts were unsuccessful, I assume it is still WIP).

This PR is still useful as a smaller hammer than `__noChroot` for non-NixOS systems; at least until `ca-derivations` is available for non-NixOS.

Ericson2314:
@newAM non-NixOS is a read herring. I have updated the Wiki with instructions.

jcumming:
I didn't know about impure derivations either! Back to reading release notes for me. 

I like the `__impure` solution better. The workflow I envision:

 - build derivation as `__impure`
 - measure output
 - back-annotate derivation with:
 
 ```
outputHashMode = "recursive";
outputHashAlgo = "sha256";
outputHash = "...";
```

In the meantime, for the derivations I worry about, I can use this patch locally.

Ericson2314:
Glad you like the idea of impure derivations!

> In the meantime, for the derivations I worry about, I can use this patch locally.

Would you like to close this PR then?

andreasfelix:
AFAIU `impure-derivations` are not cached and normal derivations cannot depend on it, so imo they don't cover all use cases this patch would cover. IMO this PR would still be useful as a smaller escape-hatch than `__noChroot`. (e.g. one use case is a derivation that only has pure inputs and is deterministic but has to perform a side-effect - like uploading source maps to a service like sentry). 

any chance you would reconsider accepting this PR?

PR #7121, opened 2022-09-30
Print common flags in --help

Hoverbear:
Currently, the output of commands like `nix develop --help` do not necessarily list all the possible options a user could pass to the command. For example, `--print-build-logs`. This also happens in https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html, which new users may believe is a full reference.

This obscures discoverability for newer users. A new user may need to consult multiple help locations to find options such as `--refresh` or `--cores`, if they discover them at all.

While speaking with @grahamc and @edolstra about the topic, we pondered if we should display all common args.

I believe this PR does that. I have some worries that it may list irrelevant options though. I haven't yet tested them all.

### DIFF
--- a/doc/manual/generate-manpage.nix
+++ b/doc/manual/generate-manpage.nix
@@ -1,11 +1,11 @@
-{ command }:
+{ toplevel }:
 
 with builtins;
 with import ./utils.nix;
 
 let
 
-  showCommand = { command, details, filename }:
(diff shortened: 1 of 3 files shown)


grahamc:
(crosslinking: DS-335)

edolstra:
Thanks, merged!

Issue #1218, opened 2017-02-04
nix-env being aborted bricks try to run it again


nh2:
I'm running NixOS in a KVM virtual machine to show this, but this problem should also exist on machines with little memory, or machines that suddenly suffer a power outage.

If I have configured too little memory (e.g. running `kvm` without the `-m` flag), this can happen:

```
[root@nixos:~]# nix-env -i htop
installing ‘htop-2.0.2’
these paths will be fetched (0.07 MiB download, 0.19 MiB unpacked):
  /nix/store/6dbi3g4hhnpc1r3rmnmj9ivxd3hzfypv-htop-2.0.2
fetching path ‘/nix/store/6dbi3g4hhnpc1r3rmnmj9ivxd3hzfypv-htop-2.0.2’...

*** Downloading ‘https://cache.nixos.org/nar/1hccfhb7m810b0ix2afjw6j830iigbb36lchshzpl1sb2wclhmjl.nar.xz’ (signed by ‘cache.nixos.org-1’) to ‘/nix/store/6dbi3g4hhnpc1r3rmnmj9ivxd3hzfypv-htop-2.0.2’...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 72180  100 72180    0     0  50005      0  0:00:01  0:00:01 --:--:-- 50020

building path(s) ‘/nix/store/ddz54781490iywc75drap5a15nv2ggc9-user-environment’
error: unable to fork: Cannot allocate memory
```

Afterwards, if I run it again, even if I have used `-m 2048` now:

```
[root@nixos:~]# nix-env -i htop
installing ‘htop-2.0.2’
error: error parsing derivation ‘/nix/store/kvdgrl0mrkdz8k1yvwya2zyh70bhmkac-htop-2.0.2.drv’: expected string ‘Derive([’
```

I suppose that's because `/nix/store/kvdgrl0mrkdz8k1yvwya2zyh70bhmkac-htop-2.0.2.drv` is simply an empty dir.

It seems that a `nix-env` install is non-atomic this way (dir created, but without contents), and thus the machine somehow failing after the `mkdir` can create this situation.

edolstra:
Did you reset the VM between the first and the second run, or otherwise uncleanly shut down the VM? That's the only situation I can think of that would cause an empty file in the Nix store registered as a valid path.

[...]

PR #7126, opened 2022-10-02
Add fsync-store-paths option

squalus:
#1218

This adds an `fsync-store-paths` option to perform durable writes before registering paths. This prevents data loss when there's an unexpected power cut. The data loss manifests as store corruption errors later. I have observed this problem in a production environment.

This PR covers the writes performed by `LocalStore` (e.g. extracting NARs). Local builds can still result in non-fsynced writes to the nix store.

It's higher performance than `sync-before-registering`, where `sync(2)` is called in `registerValidPath`. `sync(2)` can block on writing unrelated things to disk. And `registerValidPaths` can be called when there are no store path writes, like in `nix-store --load-db`.

Enabling the option causes a negative performance impact, since normally, Nix doesn't wait for store path data to be written to disk. On Linux, we use `sync_file_write(2)` to mitigate the slowdown. PostgreSQL and dpkg also use this technique. On Darwin there's a large performance impact, since there's no equivalent of `sync_file_write(2)` to initiate an asynchronous disk write.

My store corruption tests (https://github.com/squalus/nix-durability-tests) pass with this option enabled.

The option is disabled by default. There's no change in behavior when it's disabled.

---

- Add `recursiveFsync` function to flush a directory tree to disk

- Add `AutoCloseFD::startFsync` to initiate an asynchronous fsync without waiting for the result

- Initiate an asynchronous fsync while extracting NAR files

- Implement the `fsync-store-paths` option in `LocalStore`

---

Benchmark results for `time nix copy /nix/store/wd9bwh6552vvn00q99wj7qnh2gn5ndy4-go-1.18.5 --to ./teststore`

**NixOS x86_64, Haswell, SATA SSD, XFS**

```
fsync-store-paths = false

real    0m2.126s
user    0m1.273s
sys     0m0.524s

fsync-store-paths = true

real    0m2.767s
user    0m1.409s
sys     0m0.915s

fsync-store-paths = true (without sync_file_range)

real    0m27.232s
user    0m1.755s
sys     0m2.933s

fsync-store-paths = true (dont fsync directories separately, see below)

real    0m4.056s
user    0m1.438s
sys     0m1.055s

```

**NixOS x86_64, Haswell, SATA Spinning Disk, XFS**
```
fsync-store-paths = false

real    0m2.977s
user    0m1.353s
sys     0m0.567s

fsync-store-paths = true

real    0m9.517s
user    0m1.461s
sys     0m0.903s
```

**macOS, MacBookPro14,1, APFS**

```
fsync-store-paths = false

real    0m12.366s
user    0m4.245s
sys     0m8.558s

fsync-store-paths = true

real    0m39.306s
user    0m5.421s
sys     0m17.082s

```


squalus:
I'm not sure yet if it's necessary to fsync the directories in a depth-first manner. Maybe it's fine if you just do it in directory traversal order?

squalus:
Postgresql also uses this depth-first recursive fsync approach: https://github.com/postgres/postgres/blob/bdf9b60085bca2dbb0af920d50f1a0917dda0cbb/src/backend/storage/file/fd.c#L3323

They fsync the directories during traversal instead of afterwards. I tried this and found that it's a pretty big slowdown on Linux/XFS. It's a lot faster to fsync all the files, then all the directories. Added timings in the top ("dont fsync directories separately"). It's at the expense of memory though, because all directories need to be stored in memory in my implementation. Only timed on XFS so far.

edolstra:
Since this is faster than the current implementation of `sync-before-registering`, I would suggest *replacing* the implementation of `sync-before-registering` with the fsync-based implementation of this PR, but not adding a new option.

squalus:
> Since this is faster than the current implementation of `sync-before-registering`, I would suggest _replacing_ the implementation of `sync-before-registering` with the fsync-based implementation of this PR, but not adding a new option.

To do that, this PR should cover all the "non-fsynced" Nix store write cases. In particular, `local-derivation-goal.cc` will need to be updated. I originally kept the scope limited to LocalStore. I planned to do the rest in a follow-up.

Not a problem though. I can work on this.

amjoseph-nixpkgs:
We really should do this, and should probably have this option default to `true` if `/nix/store` is on btrfs (and probably bcachefs too), or perhaps on any filesystem except a curated list (ZFS, ext4) for which the recovery semantics happen to coincide with Nix's assumptions.

I've always been bewildered by the Nix community's bizarre aversion to btrfs.  Until I encountered Nix I'd never heard anybody claim btrfs lost any of their data, ever.  I believe it's due to Nix declaring "corruption! corruption! corruption!" when unsynchronized outpath registration leads to zero-sized files in the store after an unclean shutdown.

#### What's happening

BTRFS journals metadata and data separately, and therefore during recovery from an unclean shutdown it takes the *entirely POSIX-valid* perspective that a builder's `cp stuff $out/bin/stuff` consists of two separate steps:

1. Create an inode and directory entry for `$out/bin/stuff` with zero size (metadata)
2. Write a bunch of bytes into the file (data)

Without `sync-before-registering` or `fsync-store-paths`, it's entirely possible for the stream of writes to `/nix/var/nix/db/db.sqlite` to become interleaved with the two-operation sequence above before being flushed to the disk.  So you can end up with:

1. Create an inode and directory entry for `$out/bin/stuff` with zero size (metadata)
2. Register the store path in `/nix/var/nix/db/db.sqlite`
3. Write a bunch of bytes into the file (data)

#### ZFS, bcachefs

ZFS is more forgiving on these synchronization issues, partly because it has a simpler disk structure (no extent tree, since it can't reshape the array like btrfs can).  I think that "works on ZFS, blame $OTHER_FS" may have caused this bug to become rather entrenched in Nix.

I suspect that bcachefs will be as unforgiving as btrfs, and possibly moreso.

nyabinary:
Any status on this?

src/libstore/globals.hh
```diff
@@ -234,6 +234,13 @@ public:
           default is `true`.
         )"};
 
+    Setting<bool> fsyncStorePaths{this, false, "fsync-store-paths",
+        R"(
+        "Whether to call `fsync()` on store paths before registering them, to
+         flush them to disk. This improves robustness in case of system crashes,
+         but reduces performance. The default is `false`.
+        )"};
```
Ericson2314, line 242: `LocalStore` should use a local store option, not global option.

DaniD3v:
@squalus are you still working on this?

squalus:
I haven't worked on this in a while, unfortunately. It's still a problem though, and I still see nix store corruption from time to time because of this.

src/libstore/globals.hh
edolstra: Unfortunately, AFAIK, we don't have a convenient way to configure settings for the store. You don't want to set 
```
store = local?fsync-store-paths=1
```
in `nix.conf` because that will override the default value of `auto`, breaking the use of the daemon.

Also, overriding this setting when building via the daemon would be impossible.

edolstra:
I've synced this with master.

edolstra:
@squalus I've taken the liberty of adding a VM test based on your https://github.com/squalus/nix-durability-tests. Otherwise, this looks good to go.

In a follow-up PR, it would be nice to fsync paths after they have been built and to remove the `sync-before-registering` option, since that should be redundant then.

edolstra (approved):

shlevy:
@edolstra thoughts on whether this qualifies for a backport? Fixes a bug but its kind of a new feature.

src/libutil/file-system.cc
```diff
@@ -318,6 +319,50 @@ void syncParent(const Path & path)
...
+        if (!fd)
+            throw SysError("opening file '%1%'", path);
+        fd.fsync();
+        return;
+    }
+
+    /* Otherwise, perform a depth-first traversal of the directory and
+       fsync all the files. */
+    std::deque<fs::path> dirsToEnumerate;
+    dirsToEnumerate.push_back(path);
```
CertainLach, line 337: `path` also might be a symlink, not a directory: https://github.com/NixOS/nix/issues/12099

PR #7130, opened 2022-10-03
flake.nix: Remove patchNetrcRegression curl override

centromere:
Now that the netrc regression has been fixed upstream, this override is no longer needed.

See: https://github.com/NixOS/nixpkgs/pull/189083

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -13,16 +13,17 @@
       "original": {
         "owner": "kristapsdz",
         "repo": "lowdown",
+        "rev": "d2c2b44ff6c27b936ec27358a2653caaef8f73b8",
         "type": "github"
       }
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


edolstra:
Has that curl fix been backported to 22.05? Otherwise we shouldn't get rid of this override yet.

AmandaCameron:
The curl version bump wasn't backported, however it's worth noting using this flake's overlay on an unstable system causes the system to not build, as the hydra flake does. A [version check like this](https://github.com/NixOS/hydra/blob/master/flake.nix#L166-L168) from hydra for the json library could be implemented, but I'm not sure if that'd fix everything or not.

zowoq:
This was done in https://github.com/NixOS/nix/pull/7367.

centromere:
Thank you @zowoq!

PR #7132, opened 2022-10-03
Fix fish shell MANPATH creation

### LINKED ISSUE #7131: Single/multi user installation nix.fish produces bad MANPATH
**Describe the bug**

When installing Nix it installs a nix.fish script which produces a bad MANPATH environment variable. In the bash version, MANPATH is set only if it is currently not set, but in the fish version, it is always set and causes the default man location /opt/local/share/man to be excluded from the man path.

**Steps To Reproduce**

1. Install single or multi-user Nix with fish shell installed
2. Try runnning `man <executable>` for anything not installed by Nix
3. See that `man` produces an error message like `dirname: missing operand ...`

**Expected behavior**
- Default man pages for packages installed by "legacy" package managers (like `dnf`) work.

`nix-env (Nix) 2.11.1`

**Additional context**

stelcodes:
Previously the MANPATH was set even if MANPATH was empty beforehand which resulted in a MANPATH of only ~/.nix-profile/share/man which omitted the default man page directory (commonly /opt/local/share/man) from man page results.

Fixes #7131

### DIFF
--- a/scripts/nix-profile.fish.in
+++ b/scripts/nix-profile.fish.in
@@ -28,7 +28,9 @@ if test -n "$HOME" && test -n "$USER"
     # Only use MANPATH if it is already set. In general `man` will just simply
     # pick up `.nix-profile/share/man` because is it close to `.nix-profile/bin`
     # which is in the $PATH. For more info, run `manpath -d`.
-    set --export --prepend --path MANPATH "$NIX_LINK/share/man"
+    if set --query MANPATH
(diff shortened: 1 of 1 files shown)


thufschmitt (approved):
Looks good, thanks. I'll just ask @grahamc to double-check since I don't know anything about fish

patricksjackson (approved):
As a fish user, LGTM and matches the behavior of the bash equivalent in https://github.com/NixOS/nix/blob/master/scripts/nix-profile.sh.in#L29-L31

PR #7134, opened 2022-10-05
Temporarily disable the debugger during completion evaluation

### LINKED ISSUE #6692: nix-eval --debugger segfault
**Describe the bug**
If autocompleting errors out, and the `--debugger` flag is used, a new repl is started and exiting that repl causes a segfault

**Steps To Reproduce**
- Make a `flake.nix` with the content:
```
{
  outputs = x: rec {
    y = "".z;
  };
}
```
- `nix eval .#y --debugger`
- `y.<TAB>`
- `:q`
- segfault

**Expected behavior**
Either no new shell should open, or it should exit normally (the first one makes the most sense to me, but I'm not too experienced with nix)

`nix-env (Nix) 2.9.1`

**Additional context**
It says `(core dumped)`, but doesn't specify where. Considering how easy it is to reproduce, I'm not going to look for it, though if things turn out to be more complicated than they should be (which happens way too often) I'll look for it and upload it

yorickvP:
readline/editline is not re-entrant, so entering the debugger from the `completioncallback` results in an eventual segfault.

The workaround is to temporarily disable the debugger when searching for possible completions.

Fixes #6692

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -258,9 +258,8 @@ void SourceExprCommand::completeInstallable(std::string_view prefix)
                 getDefaultFlakeAttrPaths(),
                 prefix);
         }
-    } catch (EvalError& e) {
-        // swallow eval error
-        (void)e;
+    } catch (EvalError&) {
+        // Don't want eval errors to mess-up with the completion engine, so let's just swallow them
     }
 }
 
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -207,55 +207,60 @@ Strings SourceExprCommand::getDefaultFlakeAttrPathPrefixes()
 
 void SourceExprCommand::completeInstallable(std::string_view prefix)
 {
-    if (file) {
-        completionType = ctAttrs;
+    try {
+        if (file) {
+            completionType = ctAttrs;
 
-        evalSettings.pureEval = false;
-        auto state = getEvalState();
-        Expr *e = state->parseExprFromFile(
-            resolveExprPath(state->checkSourcePath(lookupFileArg(*state, *file)))
-        );
+            evalSettings.pureEval = false;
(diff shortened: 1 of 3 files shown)


thufschmitt:
Nice catch :)

I didn't try the fix so I might have overlook something, but I think this wouldn't work for the CLI completion, only for the one inside `nix repl` (which is different). The CLI completion is implemented [here](https://github.com/NixOS/nix/blob/ac0fb38e8a5a25a84fa17704bd31b453211263eb/src/libcmd/installables.cc#L208), and probably the same trick could be used in that function.

At any rate, it would be nice to have a small test for it. I think you can just extend https://github.com/NixOS/nix/blob/master/tests/completions.sh with some (failing or not) examples using `--debugger`

domenkozar:
cc @bburdette

yorickvP:
The CLI completion doesn't trigger the debugger, even with the --debugger flag. It does print the error to stderr, which might be a problem. I'll fix that and add a test.

src/libcmd/installables.cc
```diff
@@ -207,55 +207,60 @@ Strings SourceExprCommand::getDefaultFlakeAttrPathPrefixes()
...
-        completeFlakeRefWithFragment(
-            getEvalState(),
-            lockFlags,
-            getDefaultFlakeAttrPathPrefixes(),
-            getDefaultFlakeAttrPaths(),
-            prefix);
+    } catch (EvalError& e) {
+        // swallow eval error
+        (void)e;
     }
```
thufschmitt, line 264: Just for consistency with how it's done elsewhere:

```suggestion
    } catch (EvalError&) {
        // Don't want eval errors to mess-up with the completion engine, so let's just swallow them
    }
```

thufschmitt (approved):
Sorry for the delay. Just one small nitpick, but looks good otherwise

PR #7135, opened 2022-10-05
use shellspec for integration tests

rapenne-s:
Hi,

I've started working at improving the way integrations tests (the shell scripts in `./tests/`) are handled. They are currently mapped as Make target and run through `local/run_tests.sh`. This PR contains a change showing the migration for the test `search.sh`.

I propose to use the framework `shellspec` instead of the current shell wrapping and boilerplate.

https://github.com/shellspec/shellspec/

The benefits of this move would be:

- tests can be more accurate, instead of checking for a number of line in the output, we can look for pattern in stderr/stdout, return status, number of line in stderr/stdout
- tests initialization are contained in the test themselves, it's currently scattered into makefiles include and `run_test.sh`
- shellspec DSL is easy to understand and well documented, it may appear verbose but at least anyone can understand what the tests are doing, and easy to build on
- it's more fun to write them

I leave this PR as a draft, I'll be happy to receive feedback and opinions about this change. The goal would be to rewrite tests gradually into shellspec tests to make tests more sustainable and even more useful.


YorikSar:
If we're changing the language that is used to write these tests, then why `shellspec` instead of any other test DSL? We could use `rspec`, `pytest`, `Test::Spec` or anything else. I feel like using huge shell library anywhere is more fragile that pretty much anything else.

thufschmitt:
> I feel like using huge shell library anywhere is more fragile that pretty much anything else.

I agree that it's worse than most of the alternatives by most conceivable metrics. But the existing test suite is already in bash, so I suspect it'll be much simpler to migrate them to a shell-based framework.

@rapenne-s do you think it's possible to run the existing tests unmodified with `shellspec` (by changing `run_tests.sh` to run the test inside a shellspec test)? That way we can have a more unified interface

rapenne-s:
> If we're changing the language that is used to write these tests, then why `shellspec` instead of any other test DSL? We could use `rspec`, `pytest`, `Test::Spec` or anything else. I feel like using huge shell library anywhere is more fragile that pretty much anything else.

This is actually a good question. I don't have much experience with tests in general, I'd be happy to be proven wrong or shown better alternatives.

For the nix tests, we are running commands in a shell, and we need to look at outputs and return status, and also prepare / clean the environment.

It seems more common framework that are tide to a language require a lot of boilerplate to run commands and handle the results. And they may had even more dependencies into the project. Meanwhile, `shellspec` is only one extra package, and it's tailored to run and check many outputs / results / status / files resulting of the command, natively. It seems a great choice for me because of that.

rapenne-s:
> @rapenne-s do you think it's possible to run the existing tests unmodified with `shellspec` (by changing `run_tests.sh` to run the test inside a shellspec test)? That way we can have a more unified interface

I don't think anything prevent doing so, I'll take a look.

abathur:
> We could use `rspec`, `pytest`, `Test::Spec` or anything else.

I initially implemented tests for resholve's CLI in pytest but ended up (fairly early on in its life) converting these to bats (a different shell testing framework) because of how slow the tests were. I like pytest and am familiar with its idioms (and find bats a little frustrating at times), but my commit message (https://github.com/abathur/resholve/commit/7888f16feafb4d85b7be8ef6798ee0e62df4f7c8) asserts that it sped up resholve's tests by at least an OOM.

Don't let this close off avenues of exploration--especially since I can only compare pytest against a _different_ shell testing framework (bats) and am not actually sure if these integration tests do now or will ever take a meaningful amount of time--but if so it's probably worth doing a small pilot in each to test performance before any big conversion?

rapenne-s:
As of the content of the tests, with shellspec I added as much tests as possible like stdout/stderr lines, pattern, return status.

I personally feel it's better to check the result as much possible to detect any behavior change, but this may not be desirable. This can be changed obviously.

rapenne-s:
I included all the existing tests in shellspec.

However, speed isn't good: shellspec with 4 jobs takes 110 seconds and the current framework with 4 jobs takes 44 seconds.

I'm going to try bats to compare.

rapenne-s:
Hey! I think we can discuss on this topic now. I provided 2 PR: one providing shellspec to handle the tests, including a entire rewrite of the `search.sh` test into a full shellspec file, and another PR providing `bats` as the tool to handle the tests, but it's currently just running the existing tests, I didn't reimplement any test in bats itself.

Let's start with the bad news, compared to the current test framework with a i7-1165G7:

- shellspec is 47% slower (33.6s vs 22.2s)
- bats is 23% slower (28s vs 22.2s)
(numbers available at the end of the comment)

It's slower, but that still run fast enough to not be slower by a magnitude of order.

However, the good news are that they provide benefits over the current framework:

- test readability: it's easier to onboard and write tests
- more controls on the tests: we can check the output of stdout / stderr independently, status code, check patterns exist or not
- more granular tests: instead of having a whole file working/failing, we can define tests running a single command
- granular test running: you can just run a few tests instead of the whole test suite
- make reports: bats and shellspec can generate nice reports if it's something we can make use of
- timing: this may be worth investigating how we could keep track of each test timing, to see if one get slower after a commit, instead of relying on users when it's too late

I'd be really happy to get feedback about this, even if you disagree, please share your thoughts

## Numbers

```
Benchmark 1: make installcheck -j 8
  Time (mean ± σ):     22.231 s ±  2.484 s    [User: 94.175 s, System: 41.957 s]
  Range (min … max):   19.413 s … 24.219 s    5 runs

Benchmark 1: make installcheck -j 10
  Time (mean ± σ):     22.734 s ±  0.924 s    [User: 111.894 s, System: 50.114 s]
  Range (min … max):   21.333 s … 23.897 s    5 runs

----

Benchmark 1: shellspec --jobs 8
  Time (mean ± σ):     34.640 s ±  0.390 s    [User: 146.304 s, System: 60.679 s]
  Range (min … max):   33.944 s … 34.838 s    5 runs

Benchmark 1: shellspec --jobs 10
  Time (mean ± σ):     33.568 s ±  1.254 s    [User: 155.975 s, System: 63.575 s]
  Range (min … max):   31.339 s … 34.341 s    5 runs

-----

Benchmark 1: bats --jobs 8 spec/migration*sh
  Time (mean ± σ):     30.903 s ±  0.359 s    [User: 107.321 s, System: 48.304 s]
  Range (min … max):   30.487 s … 31.480 s    5 runs

Benchmark 1: bats --jobs 10 spec/migration*sh
  Time (mean ± σ):     27.930 s ±  1.182 s    [User: 119.651 s, System: 53.543 s]
  Range (min … max):   25.829 s … 28.657 s    5 runs
```

roberth:
> Let's start with the bad news, compared to the current test framework with a i7-1165G7:
> 
> * shellspec is 47% slower (33.6s vs 22.2s)
> * bats is 23% slower (28s vs 22.2s)

The specs only run the existing test scripts. This should cause almost no change, as the bulk of the work is in those scripts that are now called by the framework. The only framework based assertions check the exit status.

Based on the line counts, this PR adds ~500 lines, so in the best case, _bats_, we get about a 11 ms penalty for each of those lines. At a generous 5 lines per command, that's 45ms of overhead per tested command.

Some .sh files contain only one or two tested commands, some contain 10, some contain many more tested commands. So if we'd use bats everywhere, we may expect the overhead to increase tenfold, adding about a minute of overhead to the entire test suite. That's close to an order of magnitude slowdown.

> However, the good news are that they provide benefits over the current framework

That would be good news, but I don't think we should do it at such a high cost.
Perhaps there's another library or tool out there that doesn't suffer from such a high overhead? Maybe something from the C++, Go or Python ecosystems?

I'm surprised that @abathur got a speedup moving from python to bash. I wonder what went wrong on the python side and whether our use case differs.

-------

Nitpick:

> anyone can understand what the tests are doing

Understanding what properties they test would be even better. This isn't always easy to infer, and a misunderstood test case can turn out to be useless.

abathur:
> I'm surprised that @abathur got a speedup moving from python to bash. I wonder what went wrong on the python side and whether our use case differs.

It's been long enough that I've forgotten most of the detail, but it looks like the pytest-based testing was still intact at https://github.com/abathur/resholve/tree/176a2c2fdd181a26a1a056b91a9db65586940efb/tests if you want to look for obvious differences or antipatterns that would invalidate the observation. It's plausible that most of the overhead I observed was in pytest_shell and not in python/pytest broadly.

(I'll also reiterate that I'm not trying to argue against using python/pytest--just for doing a pilot to ensure performance is acceptable before investing a lot of conversion effort.)

thufschmitt:
We've discussed that (along with #7232) during the Nix team meeting.

Full meeting notes below, but the tl;dr is that

- There are issues with the current test suite
- There's no agreement whether using a proper testing framework would help
- The slowdown is a blocker for `bats` and `shellspec`
- We're not sure what the best solution would be, need to experiment further

<details><summary>Meeting notes</summary>

- @Ericson2314: This doesn't seem high priority
  - The current test is annoying because the logic is scattered around in make, runTest.sh, etc., but we could simply *not* scatter it around that way
  - If we switch to Meson per the RFC in discussion, this opens up more ways to test
  - THere are perf issues

- Manually running tests is hard to do, @theophane and @ericson2314 agree.
  - We should definitely fix this.
  - @edolstra: should not be a problem, just a bunch of commands
      - apart from performance, learning some obscure framework is an additional burden on contributors
      - @fricklerhandwerk could make a requirement of framework contenders to have good documentation. structurally all [xUnit](https://en.m.wikipedia.org/wiki/XUnit) frameworks are the same anyway.
  - @fricklerhandwerk: not documented, not discoverable.  if everything else turns out to be not promising enough, we could documenent it properly and refactor the existing testing scripts to be more palatable.

  - @Ericson2314: willing to work on refactoring [+1: tomberek]

- @Theophane: Agree the two frameworks we saw so far are no good because perf.
    - @fricklerhandwerk: other possibilities:
        - use a different language
        - refactor the current testsuite without using any special framework
        - move as many tests as possible to the C++ test suite

- @tomberek: would like a set of "pure" tests, and then a set of "in-real-world-with-network" tests as well
  - unit tests (C++)
  - CLI tests  (Bash functional tests)
  - integration tests (some existing NixOS, this distinction might be removed)
  - real-world "check to make sure everything is still good in a non-isolated world"
  - 
</details>

PR #7139, opened 2022-10-06
Improve error message for non-existing store path

runeksvendsen:
If the store path doesn't exist then say that, instead of the more generic "invalid".


edolstra:
Is this actually correct? What happens if the path exists but is invalid?

I think `didExist()` is a misnomer and should be named `wasValid()`.

fricklerhandwerk:
Should error messages incorporate documentation when run with `-v`? In this case, printing a definition of "validity" from the glossary.

tomberek:

PR #7140, opened 2022-10-06
Make warning about chroot store location more accurate

crawford:
While trying to use an alternate directory for my Nix installation, I noticed that nix's output didn't reflect the updated state directory. This patch corrects that and now prints the warning before attempting to create the directory (if the directory creation fails, it wouldn't have been obvious why nix was attempting to create the directory in the first place).

With this patch, I now get the following warning:

    warning: '/home/deck/.var/app/org.nixos.nix/var/nix' does not
    exist, so Nix will use '/home/deck/.local/share/nix/root' as a
    chroot store

### DIFF
--- a/src/libstore/store-api.cc
+++ b/src/libstore/store-api.cc
@@ -1363,9 +1363,9 @@ std::shared_ptr<Store> openFromNonUri(const std::string & uri, const Store::Para
                 } catch (Error & e) {
                     return std::make_shared<LocalStore>(params);
                 }
-                warn("'/nix' does not exist, so Nix will use '%s' as a chroot store", chrootStore);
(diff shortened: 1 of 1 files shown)


src/libstore/store-api.cc
```diff
@@ -1358,14 +1358,14 @@ std::shared_ptr<Store> openFromNonUri(const std::string & uri, const Store::Para
                store in ~/.local/share/nix/root. */
             auto chrootStore = getDataDir() + "/nix/root";
             if (!pathExists(chrootStore)) {
+                warn("'%s' does not exist, so Nix will use '%s' as a chroot store", stateDir, chrootStore);
```
edolstra, line 1361: Not sure the warning should be moved. That's misleading if an error occurs, because then it *doesn't* use a chroot store.

PR #7147, opened 2022-10-08
Remove a couple legacy logger->stop() calls

thomasjm:
These were removed by 1461e6cdda06f7f461114cce5b415f6d50381311, and then re-added by 502d7d9092ccf792a27088f31571dbace96f1962.

I noticed a logger getting double-stopped due to this.

Not sure why they were re-added, CC @ncfavier ?

### DIFF
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -569,8 +569,6 @@ static void main_nix_build(int argc, char * * argv)
 
         restoreProcessContext();
 
-        logger->stop();
-
         execvp(shell->c_str(), argPtrs.data());
 
         throw SysError("executing shell '%s'", *shell);
@@ -628,8 +626,6 @@ static void main_nix_build(int argc, char * * argv)
 
             outPaths.push_back(outputPath);
         }
-
-        logger->stop();
 
         for (auto & path : outPaths)
             std::cout << store->printStorePath(path) << '\n';


ncfavier:
See https://github.com/NixOS/nix/pull/6651

ncfavier:
> I noticed a logger getting double-stopped due to this.

Could you elaborate? What error did you get? After running what command?

thomasjm:
I have a fork with a custom logger. I had inadvertently made the `stop` method non-idempotent and I noticed a crash there.

I made a branch with prints to demo this: https://github.com/codedownio/nix/tree/debug-double-stop. I see double-stops with `nix-build`, coming from both `main.cc` and `nix-build.cc` when running this command:

```
./nix-build -E 'with import /home/tom/tools/nixpkgs {}; hello'
```

(I'm not a C++ person, but isn't a more reliable way to destroy the logger to use its destructor? From some googling it seems like it qualifies as having "static duration" and should automatically get destructed after the main function. I think this currently doesn't happen because the loggers are allocated with `new`...)

ncfavier:
> I have a fork with a custom logger. I had inadvertently made the `stop` method non-idempotent and I noticed a crash there.

Seems like you found your problem, then?

> I'm not a C++ person, but isn't a more reliable way to destroy the logger to use its destructor?

`logger` is a pointer (so that we can change the type of the logger dynamically), so its destructor is not called on program exit. Instead, we use `Finally` which is just a destructor that runs the given function at the end of the scope. But we don't always want to stop the logger right at the end of the program.

The `logger->stop()` call in `nix-build` is needed to print a new line before the paths are printed. The `logger->stop()` call in `nix-shell` is needed because the `Finally` destructor is not called before the call to `execve`.

Making sure that `stop` will be called *exactly* once in all branches of execution sounds like a bit of work, and it seems easier to just make it idempotent.

thomasjm:
All right, closing.

PR #7149, opened 2022-10-09
parseDrvName: remove doc/impl discrepancy, add test covering the gap

amjoseph-nixpkgs:
The documentation for `parseDrvName` does not agree with the implementation when the derivation name contains a dash which is followed by something that is neither a letter nor a digit.

The first commit corrects the documentation so it agrees with the implementation.

The second commit adds a test covering the case where there had previously been a discrepancy.


riking:
Please also add the UTF-8 test case:

`name-with-leading-unicode-é-1.0`
`name-with-nonleading-unicode-eé-1.0`

PR #7156, opened 2022-10-11
Create a vision for Nix

thufschmitt:
First very rough draft of a vision for Nix as discussed in [last week's Nix team meeting](https://discourse.nixos.org/t/nix-team-meeting-minutes-oct-7-2022/22369).

This is merely the output of a 45mins brainstorming session so far, and a roadmap should follow once we have refined this.

cc @edolstra @fricklerhandwerk @roberth @tomberek

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,4 +1,8 @@
-# Nix is everywhere and invisible
+# Nix is everywhere
+
+Nix is the universal build and [configuration management] tool.
+
+[configuration management]: https://www.sebokwiki.org/wiki/Configuration_Management
 
 Software developers use Nix as a matter of course every day, mostly without even noticing.
 Nix runs trivially, anywhere.
@@ -6,13 +10,15 @@ Nix runs trivially, anywhere.
 For individuals to large organizations, Nix underpins the entire software supply chain:
 - Developer tooling
 - Build automation
-- Binary distribution.
+- Binary distribution
 
 To this end, the Nix team will work towards the following goals.
 
 ## Make Nix easy to adopt
 
   - Well-defined target user base
+      - anyone who wants to manage the complexity of - and build - software
+      - transform bits in a declarative and reproducible way
   - Well-defined core user stories
     - Ad hoc environments
         - One-liner setup (nix-shell)
@@ -77,8 +83,3 @@ To this end, the Nix team will work towards the following goals.
       - Capnproto for RPC
       - Bazel RBE protocol
       - ...
-
-    
-
-
-"Likewise, we want Nix to be everywhere. We want a world where finding a flake.nix at the root of a software project is not a pleasant surprise but something expected and common. And we want the people who design tomorrow’s software distribution mechanism to at least know and recognize the Nix model, and take that into account."
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,0 +1,87 @@
+# Nix is everywhere and invisible
+
+Software developers use Nix as a matter of course every day, mostly without even noticing.
+Nix runs trivially, anywhere.
+
+For individuals to large organizations, Nix underpins the the entire software supply chain:
+- Developer tooling
+- Build automation
+- Binary distribution.
+
+To this end, the Nix team will work towards the following goals.
+
+## Make Nix easy to adopt
+
+  - Well-defined target user base
+  - Well-defined core user stories
+    - Ad hoc environments
+        - One-liner setup (nix-shell)
+    - Declarative environments
+        - One-liner setup ("templates")
+        - Easy modification/extension
+        - Easy and transparent usage ("direnv"/"lorri")
+    - Secret management as first-class citizen
+    - Configurations as first-class citizens
+        - Configuration/Modules/Nickel/etc.
+    - Language bindings
+    - Supply chain trust solution
+        - Content-addressed derivation
+        - Build result signing, key distribution
+        - SBOM/SLSA
+    - Unprivileged installation and use
+        <!-- valentin: this needs clarification, I still don't know what rewriting is -->
+        - Portable store?
+        - restricted-root
+        - ACLs
+        - rewriting
+    - ...
+  - Linux, MacOS and Windows support at feature parity
+
+## Make Nix a tool that users can rely on
+
+  - Reliable installer
+  - Effective testing
+    - Test coverage for all major use cases
+    - Memory safety validation (sanitizers, ...)
+    - Benchmarking infrastructure
+    - Test reports published and accessible
+  - Executable language specification
+ 
+## Make Nix a good investment for users
+
+  - Explicit compatibility guarantees (and non-guarantees)
+    - Commitment to uncompromising reproducibility
+    - Well-defined release process
+    - Feature support status
+    - Deprecation strategy
+    - LTS commitments
+    - Close Flakes schism, remove uncertanity/ambiguity/confusion
+
+  - Exemplary contributor and maintainer experience
+    - Recommended development setup
+    - Testing guidance
+    - Formalize review criteria
+    - Formalize design criteria (technical invariants)
+    - Well-defined architecture of isolated components
+        - Swappable store
+          - Formalize store protocol
+        - Swappable Nix language evaluator
+        - Swappable scheduler and remote-build system
+            - Integrate Hydra (modulo UI) into Nix
+            - Remote protocol speed and reliability improvements
+            - Binary cache protocol speed improvements
+
+    - Minimal custom code base (proven off-the-shelf components where possible)
+      - Git file hashing
+      - Sandboxing, containers
+      - Capnproto for RPC
+      - Bazel RBE protocol
+      - ...
+
+    
+<!-- roberth after meeting, feel free to remove when processed: be the binary distribution method of choice for dev tooling, such as IDE plugins that bundle their own binaries (needs Nix (libstore? installer?) to run trivially, anywhere) -->
+<!-- valentin: see the top-level goal above, does that match your idea? -->
+<!-- valentin: the architectural aspects seem to be dispersed a bit. I still think all the component notes should belong to one goal and rearranged accordingly. -->
+
+
+"Likewise, we want Nix to be everywhere. We want a world where finding a flake.nix at the root of a software project is not a pleasant surprise but something expected and common. And we want the people who design tomorrow’s software distribution mechanism to at least know and recognize the Nix model, and take that into account."
--- a/roadmap.md
+++ b/roadmap.md
@@ -3,7 +3,7 @@
 Software developers use Nix as a matter of course every day, mostly without even noticing.
 Nix runs trivially, anywhere.
 
-For individuals to large organizations, Nix underpins the the entire software supply chain:
+For individuals to large organizations, Nix underpins the entire software supply chain:
 - Developer tooling
 - Build automation
 - Binary distribution.
--- a/roadmap.md
+++ b/roadmap.md
@@ -79,9 +79,6 @@ To this end, the Nix team will work towards the following goals.
       - ...
 
     
-<!-- roberth after meeting, feel free to remove when processed: be the binary distribution method of choice for dev tooling, such as IDE plugins that bundle their own binaries (needs Nix (libstore? installer?) to run trivially, anywhere) -->
-<!-- valentin: see the top-level goal above, does that match your idea? -->
-<!-- valentin: the architectural aspects seem to be dispersed a bit. I still think all the component notes should belong to one goal and rearranged accordingly. -->
 
 
 "Likewise, we want Nix to be everywhere. We want a world where finding a flake.nix at the root of a software project is not a pleasant surprise but something expected and common. And we want the people who design tomorrow’s software distribution mechanism to at least know and recognize the Nix model, and take that into account."


roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+      - Git file hashing
+      - Sandboxing, containers
+      - Capnproto for RPC
+      - Bazel RBE protocol
+      - ...
+
+    
+<!-- roberth after meeting, feel free to remove when processed: be the binary distribution method of choice for dev tooling, such as IDE plugins that bundle their own binaries (needs Nix (libstore? installer?) to run trivially, anywhere) -->
+<!-- valentin: see the top-level goal above, does that match your idea? -->
+<!-- valentin: the architectural aspects seem to be dispersed a bit. I still think all the component notes should belong to one goal and rearranged accordingly. -->
```
roberth, line 84:
```suggestion
```

I think we can consider this done. We have "underpins dev tooling" at the start and we don't have to make all dependencies explicit in this document. We can't even begin to do that.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+For individuals to large organizations, Nix underpins the the entire software supply chain:
+- Developer tooling
+- Build automation
+- Binary distribution.
+
+To this end, the Nix team will work towards the following goals.
+
+## Make Nix easy to adopt
+
+  - Well-defined target user base
```
edolstra, line 15: Not sure what this means...

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+  - Well-defined core user stories
```
edolstra, line 16: It's not really clear what the intent of this section is. E.g. we *have* ad hoc and declarative environments, so we should say here what we actually want to improve about them.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+    - Ad hoc environments
+        - One-liner setup (nix-shell)
+    - Declarative environments
+        - One-liner setup ("templates")
+        - Easy modification/extension
+        - Easy and transparent usage ("direnv"/"lorri")
+    - Secret management as first-class citizen
```
edolstra, line 23:
```suggestion
```

Not sure whether this is realistic in the roadmap time frame (except maybe via the ACL work).

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+    - Configurations as first-class citizens
+        - Configuration/Modules/Nickel/etc.
+    - Language bindings
```
edolstra, line 26:
```suggestion
```
Much more important to have JSON output on all relevant commands.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+        - Build result signing, key distribution
+        - SBOM/SLSA
+    - Unprivileged installation and use
+        <!-- valentin: this needs clarification, I still don't know what rewriting is -->
+        - Portable store?
+        - restricted-root
+        - ACLs
+        - rewriting
+    - ...
+  - Linux, MacOS and Windows support at feature parity
```
edolstra, line 38: Don't think it's realistic to promise this, especially the Windows support.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Make Nix a tool that users can rely on
+
+  - Reliable installer
+  - Effective testing
+    - Test coverage for all major use cases
+    - Memory safety validation (sanitizers, ...)
+    - Benchmarking infrastructure
+    - Test reports published and accessible
+  - Executable language specification
```
edolstra, line 48:
```suggestion
```
Sounds like a lot of work for little benefit.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+ 
+## Make Nix a good investment for users
+
+  - Explicit compatibility guarantees (and non-guarantees)
+    - Commitment to uncompromising reproducibility
+    - Well-defined release process
+    - Feature support status
+    - Deprecation strategy
+    - LTS commitments
```
edolstra, line 57: Not sure we should commit to this right now.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+            - Integrate Hydra (modulo UI) into Nix
+            - Remote protocol speed and reliability improvements
+            - Binary cache protocol speed improvements
+
+    - Minimal custom code base (proven off-the-shelf components where possible)
+      - Git file hashing
+      - Sandboxing, containers
+      - Capnproto for RPC
+      - Bazel RBE protocol
+      - ...
```
edolstra, line 79:
```suggestion
```

Likewise this doesn't really fit under "Make Nix a good investment for users".

roadmap.md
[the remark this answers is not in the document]
roberth: Think of it as a literate programming for the test suite. Goals:
 - Improve the test suite
 - Produce a detailed specification document (readable, linkable, etc)
 - Interoperability; allow the test suite to be used in other implementations
 - Potentially, dual-purpose the spec runner as a test framework for Nixlang libraries

I brought this up in the meeting.
We're not in a stage where we're planning implementation time frames and such. I'd describe this document as a detailed mission statement, rather than a roadmap. More on that in a separate thread.

roadmap.md
```diff
@@ -0,0 +1,87 @@
+# Nix is everywhere and invisible
```
roberth, line 1: The file (and PR) call this a roadmap, but it does not outline a medium term planning.
May I suggest "mission statement" instead, for the file name and for the title:

```suggestion
# Detailed Mission Statement
```

I'm not corporate enough to come up with a better term ;), but I think this would be an improvement.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+    - Formalize review criteria
+    - Formalize design criteria (technical invariants)
+    - Well-defined architecture of isolated components
+        - Swappable store
+          - Formalize store protocol
+        - Swappable Nix language evaluator
+        - Swappable scheduler and remote-build system
+            - Integrate Hydra (modulo UI) into Nix
+            - Remote protocol speed and reliability improvements
+            - Binary cache protocol speed improvements
```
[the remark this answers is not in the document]
roberth: End users are not directly on the receiving end of this section, but by improving our ability to improve the product, we do improve the end user experience long term. (> Exemplary contributor and maintainer experience)

Componentization is good for local reasoning, for testing and it can improve build times.
(For example, having libstore and libexpr as a derivation-level dependencies helps when rebuilding (in Nix) a change to the CLI)
It also fosters other Nix implementation projects and specific use cases. For instance, `hnix` could more easily talk to the daemon, and an application of Nix in IOT could do store-level update distribution without ever shipping an evaluator to devices.

roadmap.md
roberth: We'll want to manage expectations. This isn't quite a roadmap; see https://github.com/NixOS/nix/pull/7156/files#r992369142

roadmap.md
roberth: Some of these are better not to be implemented in `nix` itself, but in places where the ecosystem can iterate on solutions. `nix` itself should be a source of stability. Places like `nixpkgs` and other repos are much more suitable for solving open-ended problems because those, unlike `nix` itself, can be pinned.

roadmap.md
roberth: I'd say equally important. A language binding can do strictly more than what JSON output achieves, but JSON is an easier and less invasive interface. We'll need both.

Ericson2314:
Really love a lot of the items here, great work everyone!

roadmap.md
Ericson2314: +1. "mission statement", "vision statement", "project objectives" / "project goals", all sound good to me.

roadmap.md
Ericson2314: Another thing could be implementation-agnostic conformance tests that allow reimplementing things in other languages, if C++ FFI is too hard.

roadmap.md
Ericson2314: Agreed it is a lot for short term, but excellent for long term.

If get the Meson build system, we can iterate on Windows incrementally I think. E.g., goals like "libnixutil builds on Windows with both MinGW and LLVM in MSVC mode" can be tackled in isolation.

https://github.com/nix-windows/nix after all has things that are too messy right now, but do work. If we pick and choose when tackling those initial goals and clean up, it need not even be that hard.

roadmap.md
Ericson2314: Ah, I see in the language case this is mentioned below.

roadmap.md
fricklerhandwerk: Wo is the target audience? What are these users' needs? This may entail sketching sobering like user personas.

roadmap.md
fricklerhandwerk: Also these user stories should be clearly defined so we can prioritise tasks to improve those specifically instead of something else.

roadmap.md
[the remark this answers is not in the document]
Ericson2314: Oh this is great! I was already talking to @tazjin about these things too. "Literate implementation-agnostic conformance test suite" sounds amazing!!

roadmap.md
fricklerhandwerk: This is not a roadmap, this is a set of objectives. It does not have or imply a schedule. A roadmap is a sequence of sub-objectives to fulfill any may have a schedule.

roadmap.md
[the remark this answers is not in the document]
Ericson2314: This might not matter to beginners, but it absolutely does matter for end users / institutional users. Maybe chop up the section to indicate that split but please don't drop this!

roadmap.md
fricklerhandwerk: While the specific points should be subject to debate and were supposed to be random brainstorm ideas, the point here is to reduce maintenance surface. We discussed quite intensely that it needs a lot of deliberation what to depend on and what to purpose-build. But the general idea would be to do things the way outsiders would expect and only customize what is special.

roadmap.md
fricklerhandwerk: Yes, I ninjedited the PR description to call it a vision.

```suggestion
# Our Vision for Nix
```

roadmap.md
Ericson2314: I think tests are actually easier than *formal* spec. Anyone can submit a test and if it passes CI it is good, but need masters or PhD skills to write a formal language spec.

roberth: I don't think @fricklerhandwerk meant to say _formal_ as in _formal proof_ or any other highly technological meaning of _formal_. _Formal_ can refer to formal language, as in natural language.

I think the other thread about this line illustrates the value of it, which, I agree, is not immediately tangible for users, but is beneficial to the broader community. ("_broader_ community"? Just the fact that such a term is relevant in this context implies that there's work to be done)

roadmap.md
roberth: We already make long term commitments for reproducibility, Nixlang compatibility, a degree of worker protocol compatibility, etc. Have we made those commitments explicit?

roadmap.md
fricklerhandwerk: This is not a promise, but an ambition. The roadmap towards that ambition will be a promise.

roadmap.md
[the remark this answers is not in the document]
tazjin: Hm, it does sound like a good idea but I have my doubts that it will be possible to create a literate spec like that which would also still be something a human would want to read.

Nix language has a huge amount of non-trivial edge cases (many of which are not covered by the C++ Nix test suite, but which have gained tests in Tvix that we run against both our interpreter and C++ Nix). While their behaviour is sometimes straightforward to describe (but not necessarily implement) in "spec-language", actually testing conformance is a verbose affair.

My gut feeling is that rather than trying specifically to make a literate spec, it would be more useful to have a spec and an accompanying conformance test suite (that can be read at one's leisure).

Either way, the hard part of this will be figuring out which things should go in the spec and which things are bugs in C++ Nix. In Tvix, our approach has been to consider "What does nixpkgs do?" as the primary guideline for this, which is the only way we see to actually cut down on some of the stuff that exists in the language. This is also the criteria by which we are working on our spec.

roadmap.md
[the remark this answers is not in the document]
tazjin: > This is a lot of work with no benefit to end users

Not true, there are lots of ways in which this is beneficial to end users (this part of the doc is basically the Tvix architecture statement from a while ago, there are reasons people keep converging on these thoughts!).

As a concrete example for an end-user benefit, having pluggable store and substitution implementations lets users adapt to local constraints. If you're in a country like Egypt (where some of us live on and off), using Nix is currently prohibitively expensive due to the huge traffic wasted in central substitution and the price of internet in those countries. Having a switchable implementation there would let us plug in something like a P2P substituting stores with Trustix-verification without impacting the workflows of people that use `cache.nixos.org` or Cachix through the standard protocols.

Easy to come up with more examples like this for each of the points ... this really is an important topic!

roadmap.md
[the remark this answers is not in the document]
roberth: I agree that not all parts of the spec will be fun to read, and the existence of executable test cases in the spec is not excuse not to write the necessary prose that belongs in a spec.

I don't have the illusion that turning a test suite into a literate program automatically creates a good spec, but it does help, and bringing the spec and the conformance suite closer together can only be a good thing. For example, some test cases can be in / near the main prose, whereas others are delegated to a detailed conformance suite section.

> Either way, the hard part of this will be figuring out which things should go in the spec and which things are bugs in C++ Nix.

> cut down on some of the stuff that exists in the language

I highly value Nix's ability to evaluate old expressions. Without this, what is the point of using Nix for reproducibility long term?

For me the choice seems obvious: put it in the spec. Now I don't have the experience of reimplementing Nix, so I'd like to know what behaviors you've considered to be bugs, but let's save that for later. We could have optional levels of conformance, to accommodate for other projects like Tvix.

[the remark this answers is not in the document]
tazjin: > I highly value Nix's ability to evaluate old expressions. Without this, what is the point of using Nix for reproducibility long term?

For some behaviours of the C++ evaluator it'd be **extremely** surprising if you had any expressions depending on it, but unfortunately as an implementor (and doubly so as a spec writer) you need to investigate every little corner (including edge-cases). That means that there's likely some behaviour we can simplify classify as bugs in C++ Nix (and eventually fix) without including them in the spec.

> I'd like to know what behaviors you've considered to be bugs, but let's save that for later

We've [started a document](https://cs.tvl.fyi/depot/-/blob/tvix/eval/docs/language-issues.md) on our end where we will collect a lot of these things, and we'll also be investigating (or have already done so) each of them to figure out whether expressions in nixpkgs or other large publicly known code-bases (e.g. home-manager, or TVL itself) depend on the behaviour. Note that this is very early form, with some things already added by @sternenseemann but lots of stuff we already know about missing. It might be useful to work on this in a shared Google Doc or similar!

roadmap.md
Ericson2314: Maybe `roadmap.md` -> `vision.md` once the other items are taken care of too (so as not to loose line comments)?

roadmap.md
thufschmitt: > we should say here what we actually want to improve about them.

I think the point of this item is that we need to know precisely how these can be used to solve concrete problems so that we know how we can meaningfully improve them

roadmap.md
```diff
@@ -0,0 +1,84 @@
...
+
+  - Reliable installer
+  - Effective testing
+    - Test coverage for all major use cases
+    - Memory safety validation (sanitizers, ...)
+    - Benchmarking infrastructure
+    - Test reports published and accessible
+  - Executable language specification
+ 
+## Make Nix a good investment for users
```
thufschmitt, line 50: It looks like the intended meaning of “good investment” needs to be explained given the misunderstandings on that section

```suggestion
## Make Nix a good investment for users

_Give potential users the assurance that Nix will keep working reliably for them in the long run._
```

roadmap.md
thufschmitt: Dunno about "LTS" in particular, but the whole "compatibility guarantees" is a very important point in my opinion. As @roberth says, we already have most of these implicitly, and it's important to make them explicit so that users know what to expect

roadmap.md
thufschmitt: > But the general idea would be to do things the way outsiders would expect and only customize what is special.

100% agree. I think there's three motivations here:
- By only focusing on what makes Nix Nix, we can reduce the maintenance burden (with the caveats that we must be very deliberate in choosing what we depend on as Valentin said)
- By reusing as much as possible off-the-shelf components/standards we remove as much accidental complexity from the face of new users as we can (because although the core of Nix would remain something new to learn, all the _stuff_ around it would potentially be already boring technology). And we send the message that Nix isn't a weird exotic technology, but something firmly grounded in the state-of-the-art technological landscape
- In terms of pure expressiveness, speaking mainstream protocols means that Nix can integrate with existing tool in a much nicer way. Like what flakes do with the Git repositories for example

roadmap.md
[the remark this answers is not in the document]
thufschmitt: Whether (and when) in practice we'll go all the way down there is something for the roadmap (and I tend to agree with Eelco that given the state of things it might be too costly for the short/medium term), but I think that it is a desirable thing, at least as an internal distinction (a stronger one than the `libstore`/`libexpr` split which unfortunately has grown to be very leaky and I think isn't modular enough)

> Having a switchable implementation there would let us plug in something like a P2P substituting stores with Trustix-verification without impacting the workflows of people that use cache.nixos.org or Cachix through the standard protocols.

Thinking about this (and most of the other use-cases that come to my mind), it feels like the most important part to properly distinguish from the rest (and that's kinda missing here) is an official and properly documented binary cache API. Does that seem right?

roadmap.md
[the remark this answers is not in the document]
fricklerhandwerk: >  That means that there's likely some behaviour of X we can simplify classify as bugs (and eventually fix) without including them in the spec for X.

I've seen many instances where this principle could be applied. Documenting things with a bit of wishful thinking helps a lot with thinking straight in the first place.

roadmap.md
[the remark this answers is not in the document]
Ericson2314: @thufschmitt I think it is somewhat backwards to think of this as "tool costly". Fixing this is in fact how we *pay* for things like "Minimal custom code base"; only by cleaning up our own code can we figure out how to slot in off-the-shelf well.

This definitely has benefits on its own, but is is also necessary to tackle other goals without drowning in complexity. *Not* doing this is more costly if we are to tackle those other goals.

------

Also agreed with @tazjin. high quality network interfaces, whether for substituting or RPC, leverage domain-specific concerns in lots of ways. The current HTTP interface and the current RPC protocols are equally crummy in this regard.

I will say however that the "having a switchable implementation" would be fine to put under "Minimal custom code base" rather than this. I think that effectively clears up the gaps. The flow of benefits is: good layering -> easy swapping of bespoke current things for external things -> easy to experiment with P2P strategies.

Ericson2314:
(FYI to anyone reading this there are replies to @thufschmitt's comments but they are only showing up in the diff view, not conversations view.)

roadmap.md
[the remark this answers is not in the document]
7c6f434c: Well, purely as a user, and assuming _only_ layer separation and not the further development enabled by it, I see layer separation as the only chance to close weird gaps in CLI such as impossible to find in docs (at least to me) operation «.drv + output name → output path» in a systematic way.

roadmap.md
thufschmitt: > invisible

Is that the right word? Nix can't really be "invisible" since people actually interact with it

roadmap.md
thufschmitt: (Eelco) should that be defined before we can have a clear vision? So that we can know where we want to lead Nix

roadmap.md
thufschmitt: (eelco): Might be too low-level for the vision
(Robert): The ecosystem already has solutions for that, so probably not mandatory for Nix

roadmap.md
```diff
@@ -0,0 +1,84 @@
...
+  - Well-defined core user stories
+    - Ad hoc environments
+        - One-liner setup (nix-shell)
+    - Declarative environments
+        - One-liner setup ("templates")
+        - Easy modification/extension
+        - Easy and transparent usage ("direnv"/"lorri")
+    - Secret management as first-class citizen
+    - Configurations as first-class citizens
+        - Configuration/Modules/Nickel/etc.
```
thufschmitt, line 25:
```suggestion
        - Have everything Nix be configurable, discoverable and introspectable
        - Ability to configure arbitrary systems (incl. secrets)
```

roadmap.md
thufschmitt: (roberth): A number of projects rely on or would benefit from unofficial bindings

thufschmitt: On the Nix side, I think that the important bit would be to provide a stable FFI interface that bindings can build against. Whether the different bindings live inside or outside of the Nix repo is more of a detail imho

roadmap.md
```diff
@@ -0,0 +1,84 @@
...
+    - Supply chain trust solution
+        - Content-addressed derivation
+        - Build result signing, key distribution
+        - SBOM/SLSA
+    - Unprivileged installation and use
+        <!-- valentin: this needs clarification, I still don't know what rewriting is -->
+        - Portable store?
+        - restricted-root
+        - ACLs
+        - rewriting
```
thufschmitt, line 36: Could be moved to a sub item. The important bit is that Nix runs everywhere

roadmap.md
```diff
@@ -0,0 +1,84 @@
...
+  - Exemplary contributor and maintainer experience
+    - Recommended development setup
+    - Testing guidance
+    - Formalize review criteria
+    - Formalize design criteria (technical invariants)
+    - Well-defined architecture of isolated components
+        - Swappable store
+          - Formalize store protocol
+        - Swappable Nix language evaluator
+        - Swappable scheduler and remote-build system
```
thufschmitt, line 69: “swappable” might not be the right term here. I don't think we should care too much about it being swappable, but more about having clear and well-defined interfaces (the "swappable" thing can be a side-effect of this, but not the first goal)

roadmap.md
```diff
@@ -0,0 +1,84 @@
...
+            - Integrate Hydra (modulo UI) into Nix
+            - Remote protocol speed and reliability improvements
+            - Binary cache protocol speed improvements
+
+    - Minimal custom code base (proven off-the-shelf components where possible)
+      - Git file hashing
+      - Sandboxing, containers
+      - Capnproto for RPC
+      - Bazel RBE protocol
```
thufschmitt, line 78:
```suggestion
      For example, use one of these internally:
      - Git file hashing
      - Sandboxing, containers
      - Capnproto for RPC
      - Bazel RBE protocol
```

thufschmitt (commented):
Comments from a group review with @edolstra and @roberth

roadmap.md
roberth: some ideas
 - and works transparently (- as possible?)
 - and gets out of the way
 - and is a robust utility (in the sense of electricity etc)
 - and is an expected but boring part of modern computing

roadmap.md
SuperSandro2000: Windows is not even a Unix or *nix. Software is often build, distributed and developed very different and a lot of software on Windows is freeware or proprietary especially the entire foundation and core parts. I don't think it is realistic to get nix on feature parity on Windows.

roadmap.md
SuperSandro2000: How do we want to realize that other than giving our word? Sell service contracts and insurance?

roadmap.md
SuperSandro2000: Right now nixpkgs and so a bit distant also nix is the exact opposite of LTS and I know for a fact that many people I know that use nix choose nix because of the fast development cycle like Arch but with reproducibility and it if it falls apart you can rescue it without a reinstall. Converting nix(pkgs) to LTS would alienate a lot of users.

If LTS means compatibility guarantees (which is not the same thing) then the wording here must be improved.

roadmap.md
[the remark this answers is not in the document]
SuperSandro2000: If we can foster improvements (like already mentioned above) by splicing parts up into smaller pieces I am all for it but we should be careful that we not blindly start and end up with a thousand pieces in the end with no plan or goal and everything just got more complex.

```suggestion
    - Well-defined architecture of isolated components
        - Swappable store and formalized store protocol
        - Swappable Nix language evaluator
        - Swappable remote-build system including it's scheduler
```

I wouldn't merge hydra in the short term to keep complexity lower and the amount of code lower. Also improvements should go to a different section.

roadmap.md
SuperSandro2000:
```suggestion
    - Try to use proven off-the-shelf components where possible like already done with:
      - editline
      - libarchive
      - lowdown
      - sqlite
```

roadmap.md
Ericson2314: I really like "utility" I will leave some comments on that.

roadmap.md
```diff
@@ -0,0 +1,84 @@
+# Nix is everywhere and invisible
+
+Software developers use Nix as a matter of course every day, mostly without even noticing.
+Nix runs trivially, anywhere.
```
Ericson2314, line 4:
```suggestion
# Our vision for Nix: *the* utility behind all software development

Utilities, like water, electricity, or the internet are not like other products or inputs.
They are intentionally neutral, unopinionated, or even *boring*.
No one complains electricity or water is "misdesigned" or a the wrong tool for the job --- "a square peg in a round hole".
They also don't care what you use them for, and they are ubiquitous.
There are too many tasks that rely on them for anyone to keep track of.
When they do try to try to step outside their division of labor, like violating net neutrality, users don't like it.
Most of the time they just work, and during those times, you don't even think about them.
They just recede in the background.

All software development should be able to rely on Nix this same way.
Whatever you are trying to do, Nix should be an essential tool to support what you do, and yet one you don't even notice.
Whatever you opinions are on how packages should be written, dependencies specified, Nix should be able to help out.
Whatever machines you have at your disposal to build or deploy software, Nix should be able to work with them.

There should never be a reason not to use Nix.
```

roadmap.md
7c6f434c: Should there be an acknowledgement that this is at high risk (at least) to contradict the minimisation of opinionated decisions, or some idea how to manage this balance?

roadmap.md
7c6f434c: Hmm, is there any kind of a principle in the vision what should be allowed to evolve in the ecosystem and what ends up in the core?

roadmap.md
fricklerhandwerk: We decided to remove the "invisible" aspect.

roadmap.md
fricklerhandwerk: We decided not to go into that level of detail in the vision statement.

roadmap.md
fricklerhandwerk: This is about the development model, e.g. determining how we deal with deprecation and sticking to that religiously. On the other hand, keeping Nix compatible with its surrounding hardware and software ecosystem, which continues to evolve, is indeed something that needs sustained maintenance.

roadmap.md
fricklerhandwerk:
```suggestion
    - Compatibility guarantees
```

roadmap.md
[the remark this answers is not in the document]
fricklerhandwerk:
```suggestion
    - Well-defined architecture of isolated components
        - Swappable store and formalized store protocol
        - Swappable Nix language evaluator
        - Swappable remote build system and scheduler
```

roadmap.md
fricklerhandwerk: > The ecosystem already has solutions for that, so probably not mandatory for Nix

@robeth Which are those solutions? There was a brief lunch break debate at NixCon that e.g. a primitive in the Nix language   for dealing with secret values would be very helpful to enable more sophisticated solutions on top of that.

> is there any kind of a principle in the vision what should be allowed to evolve in the ecosystem and what ends up in the core?

@7c6f434c not yet, but I think we absolutely have to answer that question, somehow. For example, it could be implied by the vision statement, i.e. with everything that is not specifically defined by it. In any case we have to make explicit that we draw such a boundary, and make clear where that is.

roadmap.md
fricklerhandwerk: This strikes me as highly ambiguous. What is "everything in Nix" and what does it mean for that to have these properties? Do you mean the source code, or stuff at runtime?

roadmap.md
```diff
@@ -0,0 +1,85 @@
...
+- Build automation
+- Binary distribution
+
+To this end, the Nix team will work towards the following goals.
+
+## Make Nix easy to adopt
+
+  - Well-defined target user base
+      - anyone who wants to manage the complexity of - and build - software
+      - transform bits in a declarative and reproducible way
```
fricklerhandwerk, line 21: I remembered this phrase I coined when trying to describe what got me into Nix:

```suggestion
      - anyone who wants to make computers work reliably in the long run
```

Advantages:
- not limited to software development, but still focused on making computers do something
- captures the reproducibility part without using the technical term

One could play with the "want to" part. To make it softer or broader:

```suggestion
      - anyone who cares for computers to work reliably in the long run
```

```suggestion
      - anyone who has an interest in computers working reliably in the long run
```

To make it stricter and narrower:

```suggestion
      - anyone who needs computers to work reliably in the long run
```

```suggestion
      - anyone who requires computers to work reliably in the long run
```

roadmap.md
fricklerhandwerk: >  this is at high risk to contradict the minimisation of opinionated decisions

@7c6f434c what would you charactarise as an opinionated decision in the context of determining first-class user stories? Maybe I'm misunderstanding what you're trying to say.

roadmap.md
7c6f434c: Manage complexity of _software_ is more fitting than making _computers_ work reliably, the latter also needs stuff like easily swappable power supply components etc.

And indeede the value of Nix is in uniform handling of one aspect of complexity of managing software systems, not just reproducibility: I don't care about rebulds from scratch quite as much as about atomic updates and reduction of conflict surface.

(Moreover, for some notions of «run reliably» Nix does nothing with the problem that either you update and get new upstream bugs, or don't and could just as well boot a read-only Ubuntu live image or something)

roadmap.md
7c6f434c: Well, drawing a line between first-class and second-class usages is already an opinionated decision, and, for example, the «mass market» user stories probably don't use much package overrides so any story-based redesign will probably impair override facilities. (Without which facilities I probably find Nixpkgs _much_ less useful). And look at unfree-predicate vs flakes to see how well things will go. Another thing is platform support: for first-class user stories you probably say that something either mostly works or not and this fact is recorded in the code base, for exotic cross-compilation with a unique set of tunables things might be broken by upstream changes and patched back to work too often for «broken» in a package to even make sense.

roadmap.md
7c6f434c: What currently says why pinning part of flakes should not be external add-on on equal footing with niv that people used for many of the same aims? I think that any «define basic logic and derive from there» kind of boundary will end up demanding a large redrawing of lines, so you probably need to define at least a _process_ for redrawing the boundary (which Nix currently lacks, judging from NixOS/RFCs#0049)

roadmap.md
7c6f434c: And even at runtime, is it in terms of Nix lib* as a library, or in terms of reflection available to Nix code?

roadmap.md
7c6f434c: Keeping Nix compatible with ecosystem — would it include accepting a patch with an exploit for PlayStation 3 if somebody had provided it after working on OtherOS support there and Sony fraudulently disabling an advertised feature?

roadmap.md
```diff
@@ -0,0 +1,84 @@
...
+  - Reliable installer
+  - Effective testing
+    - Test coverage for all major use cases
+    - Memory safety validation (sanitizers, ...)
+    - Benchmarking infrastructure
+    - Test reports published and accessible
+  - Executable language specification
+ 
+## Make Nix a good investment for users
+
```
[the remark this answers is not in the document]
7c6f434c: Does this intend to forbid any hardcoded reference to git in flakes code?

roadmap.md
7c6f434c: Swappable could also be a criterion of this, though.

roadmap.md
roberth: > > The ecosystem already has solutions for that, so probably not mandatory for Nix
> 
> @roberth Which are those solutions? There was a brief lunch break debate at NixCon that e.g. a primitive in the Nix language for dealing with secret values would be very helpful to enable more sophisticated solutions on top of that.

Existing solutions: NixOps `deployment.keys`, sops-nix, agenix.
Potential solutions:
 - NixOS: make NixOps keys part of NixOS. For example `nixos-rebuild` could take secrets from `/root` and install them.
 - Nix: extend the string context to allow marking of secret values so that they can not be written to the store.
 - Nix: complicate the store concept. Not a fan of this.

roadmap.md
fricklerhandwerk:
```suggestion
      - anyone who requires software to work reliably both right now *and* in the long run, at scale
```

roberth: I've thought about the problem some more and it seems that we (or at least me) are not clear on the purpose of defining the Nix audience.
Is it to attract people? That's better handled by the marketing team.
Is it to reject PRs? We should probably judge them on merit instead.
Is it to inform a roadmap? Then define one or two personas to focus on, which aren't the whole audience.
Is it something else? Then please enlighten me.

edolstra: >  Is it to reject PRs? We should probably judge them on merit instead.

Ideally, having a vision+roadmap makes it easier to decide which PRs are out of scope, i.e. should be closed regardless of their merit because they would take Nix into a direction we don't want.

So the audience shouldn't be too broad or too vague. I'd rather say "Nix is for software developers / DevOps" than "Nix is for anyone who want reliable software".

roberth: I doubt that we need to reject PRs based on who a person "is". Making it personal (or about groups) is a recipe for disaster.

> Ideally, having a vision+roadmap makes it easier to decide which PRs are out of scope, i.e. should be closed regardless of their merit because they would take Nix into a direction we don't want.

So then we can define the project's scope and direction instead.

> So the audience shouldn't be too broad or too vague.

This only follows if we assume that we take all of the audience's PRs as orders, which can't be the goal.

fricklerhandwerk: > Is it to attract people?

@roberth No. It is to inform which of the many possible use cases to elevate into the vision statement.

> Is it to reject PRs? We should probably judge them on merit instead.

Yes, and that merit needs an intended receiver.

> Is it to inform a roadmap? Then define one or two personas to focus on, which aren't the whole audience.

That would indeed be for the roadmap, and a roadmap will always only cover a subset of the vision at a time.

> the audience shouldn't be too broad or too vague.

@edolstra yes, this is why I push for finding a concise and clear-cut phrasing. It must not be too narrow either.

After all, this is simply a decision who we, as maintainers, commit to make Nix useful for, eventually. So maybe we should ask ourselves that question differently.

> I'd rather say "Nix is for software developers / DevOps"

We could say that, and decide if we want to deliberately exclude data scientists for example. But we know that scientists already greatly benefit from using Nix. And while DevOps is something people use Nix for, if Nix had become more popular 15 years prior, that job description may never have emerged in the first place.

That target audience is also to some extent an economic decision: What kind of organisation will support Nix development in the future?

> "Nix is for anyone who want reliable software"

This is explicitly not what I'm proposing, because as you say, that's everyone. Nix already is for people who have some pressing need to rely on software, or work on making it reliable and can make use of Nix as a means to that end.

Okay, here's two proposals for next time:
1. Brainstorm concrete groups which fall into our target audience and those who don't.

   Agree on these two sets, remove everything that is controversial. Maybe an apt description will emerge. Otherwise  leave the two sets as the outcome. We did this partially last time, but did not really deal with the non-audience, and did not do the agreement part. We can still update the vision if someone comes up with an obviously better phrase.

2. Set aside the audience topic, and brainstorm use cases that we clearly agree we want to support.
   T
   As said, this will play into defining user personas, and probably the list we will come up with will inform that.

roadmap.md
fricklerhandwerk: @7c6f434c note that this is only about Nix itself, not Nixpkgs. The user stories would tell what Nix allows immediately, and Nixpkgs would use that for the extended use cases it enables in turn.

>  And look at unfree-predicate vs flakes to see how well things will go.

What do exactly you mean by that?

roadmap.md
roberth: > After all, this is simply a decision who we, as maintainers, commit to make Nix useful for, eventually. So maybe we should ask ourselves that question differently.

This should include our current user base.

> > Is it to attract people?
> 
> No. It is to inform which of the many possible use cases to elevate into the vision statement.

Good. Publishing a decision about audience is going to be broad or divisive. Let's not publish one.

> > Is it to reject PRs? We should probably judge them on merit instead.
> 
> Yes, and that merit needs an intended receiver.

Our current user base would be a good starting point. We know quite well who they are.

> > Is it to inform a roadmap? Then define one or two personas to focus on, which aren't the whole audience.
> 
> That would indeed be for the roadmap, and a roadmap will always only cover a subset of the vision at a time.

So, kind of, but not really?

> After all, this is simply a decision who we, as maintainers, commit to make Nix useful for, eventually. So maybe we should ask ourselves that question differently.

I don't think picking favorites in public is something you do "simply".

> > I'd rather say "Nix is for software developers / DevOps"
> 
> We could say that, and decide if we want to deliberately exclude data scientists for example. But we know that scientists already greatly benefit from using Nix. And while DevOps is something people use Nix for, if Nix had become more popular 15 years prior, that job description may never have emerged in the first place.
> 
> That target audience is also to some extent an economic decision: What kind of organisation will support Nix development in the future?

Let's not make badly informed economic decisions.

> I'm fine with just saying "software developers". Anyone else may not have or want to use Nix directly anyway.

We do support indirect use of Nix though, via `--json` for example. Should we exclude them? I think not. Projects that facilitate such use are very committed and if successful they'll want to support the project because it's an essential part of theirs.

> > "Nix is for anyone who want reliable software"
> 
> This is explicitly not what I'm proposing, because as you say, that's everyone. Nix already is for people who have some pressing need to rely on software, or work on making it reliable and can make use of Nix as a means to that end.
> 
> Okay, here's two proposals for next time:
> 
> 1. Brainstorm concrete groups which fall into our target audience and those who don't.
>    Agree on these two sets, remove everything that is controversial. Maybe an apt description will emerge. Otherwise  leave the two sets as the outcome. We did this partially last time, but did not really deal with the non-audience, and did not do the agreement part. We can still update the vision if someone comes up with an obviously better phrase.

I'm not convinced that there's a real need for this.

> 2. Set aside the audience topic, and brainstorm major use cases.
>    Leave all that we clearly agree on, and prioritise roughly.
>    As said, this will play into defining user personas, and probably the list we will come up with will inform that.

Sounds better.

I'd like to see us work toward practically achievable goals and show the community that we're making real progress.

Ericson2314: I think it's fairly hard to write a vision statement for Nix as a whole, but much easier to rewrite ones for specific layers like libnixstore and libnixexpr. This should be a sign that divvying up responsibilities should make Nix project management easier and avoid clashing opinions or uncertainty.

fricklerhandwerk: I agree with everything you wrote. Just one note, so we don't misunderstand each other.

> I don't think picking favorites in public is something you do "simply".

I don't see how any of this is about picking favorites. The "simple" here was supposed to refer to what we talked about last week: It should be simple to answer for users or contributors if their intent to use or modify Nix is covered by the vision statement. That is primarily to communicate what kind of support to expect from maintainers.

We should by all means lead the discussion in public, to make sure we're on the right track. That is naturally not simple or easy, because it means saying "no" to some people in order to be able to clearly say "yes" to other people. It always hurts.

At the very least, the rationale behind that decision must be absolutely transparent, in my opinion.

I'm also fine with not determining this and instead skipping to user stories, but I suspect that will amount roughly to the same thing. Nix will not be able to solve everyone's problems, and we have to focus on something in the long term to be able to pick something to focus on in the short term.

7c6f434c: > It should be simple to answer for users or contributors if their intent to use or modify Nix is covered by the vision statement.

Stated like this, it is «what Nix does» not «what use cases Nix serves» — which is of course better.

(And of course Nix does not help with reliability of software itself, but it is a good tool to reduce cross-contamination of different parts when such interactions are undesirable)

roadmap.md
7c6f434c: With non-flake Nixpkgs, you can quickly and easily relax unfree/broken/platform checks for a single invocation. Flakes make it complicated with no clear way how to change Nixpkgs to have the same ease back. It's a Nix change that makes Nixpkgs less useful for some use cases.

roadmap.md
Ericson2314: > Stated like this, it is «what Nix does» not «what use cases Nix serves»

@7c6f434c It sounds like we are thinking alike :) I started writing a bunch on what I am calling "Supply-" versus "demand-driven" project management.

7c6f434c: I mean, I myself am more or less Unix-maximalist (composable scope-able predictable tools, all that). Also I sincerely believe that there are many more tools useful _despite_ their intended use cases but thanks to their predictability and friendliness to integrating into larger toolchains — than tools that are actually mainly useful as intended.

PR #7167, opened 2022-10-13
Don't incorrectly assign aarch64-darwin to x86_64-darwin machines

grahamc:
Users on an aarch64 Mac should install an aarch64 copy of Nix. If they'd like to use an x86_64 Nix on an aarch64 Mac and still get the extra-platforms support, they should configure it manually instead of incorrectly configuring x86_64 Macs.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -156,9 +156,7 @@ StringSet Settings::getDefaultExtraPlatforms()
     // always exec with their own binary preferences.
     if (pathExists("/Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist") ||
         pathExists("/System/Library/LaunchDaemons/com.apple.oahd.plist")) {
-        if (std::string{SYSTEM} == "x86_64-darwin")
-            extraPlatforms.insert("aarch64-darwin");
-        else if (std::string{SYSTEM} == "aarch64-darwin")
+        if (std::string{SYSTEM} == "aarch64-darwin")
             extraPlatforms.insert("x86_64-darwin");
     }
 #endif


src/libstore/globals.cc
```diff
@@ -156,9 +156,7 @@ StringSet Settings::getDefaultExtraPlatforms()
     // always exec with their own binary preferences.
     if (pathExists("/Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist") ||
         pathExists("/System/Library/LaunchDaemons/com.apple.oahd.plist")) {
-        if (std::string{SYSTEM} == "x86_64-darwin")
-            extraPlatforms.insert("aarch64-darwin");
-        else if (std::string{SYSTEM} == "aarch64-darwin")
+        if (std::string{SYSTEM} == "aarch64-darwin")
```
Hoverbear, line 159: Hmm this seems to still be wrong, and also only assign this for aarch64-darwin now?

grahamc: An aarch64-darwin with Rosetta indeed does have x86_64-darwin as a supported (extra) platform.

grahamc: Nix doesn't need the current system to be an "extra" platform :).

cole-h: I did a brief test with an M1 and Intel Mac I have access to, and the `/System/Library/LaunchDaemons/com.apple.oahd.plist` path existed on both (including the Intel Mac), which I think is why the "if x86, add aarch" branch was removed -- it ends up activating on non-Rosetta machines.

grahamc:
Different approach incoming.

PR #7168, opened 2022-10-13
Improve Rosetta detection

edolstra:
Turns out that one of those *.plist files can exist even if Rosetta is not installed. So let's just try to run an x86_64-darwin binary directly.

### DIFF
--- a/src/libstore/globals.cc
+++ b/src/libstore/globals.cc
@@ -154,12 +154,12 @@ StringSet Settings::getDefaultExtraPlatforms()
     // machines. Note that we can’t force processes from executing
     // x86_64 in aarch64 environments or vice versa since they can
     // always exec with their own binary preferences.
-    if (pathExists("/Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist") ||
-        pathExists("/System/Library/LaunchDaemons/com.apple.oahd.plist")) {
-        if (std::string{SYSTEM} == "x86_64-darwin")
-            extraPlatforms.insert("aarch64-darwin");
-        else if (std::string{SYSTEM} == "aarch64-darwin")
+    if (std::string{SYSTEM} == "aarch64-darwin") {
+        if (runProgram(RunOptions {.program = "arch", .args = {"-arch", "x86_64", "/bin/pwd"}, .mergeStderrToStdout = true}).first == 0) {
+            debug("Rosetta detected");
             extraPlatforms.insert("x86_64-darwin");
+        } else
+            debug("Rosetta not detected");
     }
 #endif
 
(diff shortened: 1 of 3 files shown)


grahamc:
Looks pretty good to me relative to the behaviors we were seeing on the various macs.

src/libstore/globals.cc
```diff
@@ -154,12 +154,12 @@ StringSet Settings::getDefaultExtraPlatforms()
     // machines. Note that we can’t force processes from executing
     // x86_64 in aarch64 environments or vice versa since they can
     // always exec with their own binary preferences.
-    if (pathExists("/Library/Apple/System/Library/LaunchDaemons/com.apple.oahd.plist") ||
-        pathExists("/System/Library/LaunchDaemons/com.apple.oahd.plist")) {
-        if (std::string{SYSTEM} == "x86_64-darwin")
-            extraPlatforms.insert("aarch64-darwin");
-        else if (std::string{SYSTEM} == "aarch64-darwin")
+    if (std::string{SYSTEM} == "aarch64-darwin") {
+        if (runProgram(RunOptions {.program = "arch", .args = {"-arch", "x86_64", "/bin/pwd"}, .mergeStderrToStdout = true}).first == 0) {
```
grahamc, line 158: We could also use true (thanks @cole-h):

```
ephemeraladmin@mac-obliging-starfish> file /usr/bin/true                                                                                                                                          ~
/usr/bin/true: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64
- Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e
- Mach-O 64-bit executable arm64e]
/usr/bin/true (for architecture x86_64):	Mach-O 64-bit executable x86_64
/usr/bin/true (for architecture arm64e):	Mach-O 64-bit executable arm64e

$ arch -arch x86_64 /usr/bin/true; echo $?                                                                                                                    ~
0

```

grahamc:
@andir mentioned this might trigger a pop-up. We should test this interactively and see what happens.

grahamc:
I ran this over SSH and inside a gui based terminal on the mac:

```
ephemeraladmin@mac-epic-turducken ~ % arch -arch x86_64 /usr/bin/true
arch: posix_spawnp: /usr/bin/true: Bad CPU type in executable
```

and neither way caused a popup. Thankfully :).

andir:
> @andir mentioned this might trigger a pop-up. We should test this interactively and see what happens.

I specifically meant first-use of rosettta binaries on a Mac. I remember having been promted with a "rosetta installer" popup the first time running an x86_64 binary on my M1 device.

grahamc:
Yep, this machine is aarch64 and doesn't have Rosetta. I think this makes this a safe/correct/not annoying test. Right?

edolstra:
Yeah looks like `arch` doesn't trigger a popup.

divanorama:
Also can confirm that on x86_64-darwin `/System/Library/LaunchDaemons/com.apple.oahd.plist` can be present but it can't execute aarch64-darwin binaries of course

PR #7172, opened 2022-10-14
libmain: Make stack overflow handler configurable

roberth:
This make the code to run in the stack overflow handler configurable.
It is useful for libmain / libexpr consumers that want to report the stack overflow differently.
The doc describes some of the risks and requirements for writing your own handler. It is the responsibility of the caller to use it appropriately.

### DIFF
--- a/src/libmain/shared.hh
+++ b/src/libmain/shared.hh
@@ -113,14 +113,15 @@ struct PrintFreed
 /* Install a SIGSEGV handler to detect stack overflows. */
 void detectStackOverflow();
 
-/* Pluggable behavior to run before _exit(1) in case of a stack overflow.
+/* Pluggable behavior to run in case of a stack overflow.
 
-   Default value: do nothing, return immediately.
+   Default value: defaultStackOverflowHandler.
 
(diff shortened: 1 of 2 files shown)


src/libmain/shared.hh
```diff
@@ -113,5 +113,25 @@ struct PrintFreed
...
+   This is called by the handler installed by detectStackOverflow().
+
+   This gives Nix library consumers a limit opportunity to report the error
+   condition. The handler should exit the process.
+   See defaultStackOverflowHandler() for a reference implementation.
+
+   NOTE: Use with diligence, because this runs in the signal handler, with very
+   limited stack space and a potentially a corrupted heap, all while the failed
+   thread is blocked indefinitely. All functions called must be reentrant. */
+extern std::function<void(siginfo_t * info, void * ctx)> stackOverflowHandler;
```
edolstra, line 129: I think this should work: 

```suggestion
static constexpr auto stackOverflowHandler = defaultStackOverflowHandler;
```

This removes the need for the definition of `stackOverflowHandler` in `stack.cc` which could get out of sync.

PR #7176, opened 2022-10-15
Don’t override flake ref if no url or type provided

matthewbauer:
We have a situation where we want to override an input’s input, but not the input itself. It looks something like:

```
inputs.a = {
  type = "github";
  owner = "example";
  repo = "a";
  inputs.b.inputs.nixpkgs.follows = "nixpkgs";
};
```

The problem is that Nix thinks we are overriding "b", when we are really only wanting to override one of "b"’s inputs. The flake url should remain the same, but in the past behavior, Nix assigns "b" to an indirect input, and ignores the "b" from "a"s inputs. We want it to behave more like the "follows" and not treat it as a full override of "b".

### DIFF
--- a/src/libexpr/flake/flake.cc
+++ b/src/libexpr/flake/flake.cc
@@ -163,9 +163,6 @@ static FlakeInput parseFlakeInput(EvalState & state,
             input.ref = parseFlakeRef(*url, baseDir, true, input.isFlake);
     }
 
-    if (!input.follows && !input.ref)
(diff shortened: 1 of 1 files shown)


Kha:
That sounds like #6621. I'm still waiting for #6530 to be merged first.

matthewbauer:
> That sounds like #6621. I'm still waiting for #6530 to be merged first.

Oh wow! I didn’t see that! Yours looks pretty similar to mine but it looks like you handle some edge cases better. I’m not sure if I understand why it had to be reverted though?

PR #7188, opened 2022-10-18
Allow missing binary caches by default

### LINKED ISSUE #4383: More helpful behavior when caches are broken/missing

- I've tried to build my package, but an unrelated cache is failing

    ```
    building '/nix/store/8z0aq6pnim0v2rfnl97m5wqq505v3qaj-cabal2nix-.drv'...
    installing
    warning: unable to download 'https://my-self-hosted-unreliable-cache/1p34mip9xdfr5zp12895z7wxj9pgq83m.narinfo': HTTP error 502; retrying in 314 ms
    warning: unable to download 'https://my-self-hosted-unreliable-cache/1p34mip9xdfr5zp12895z7wxj9pgq83m.narinfo': HTTP error 502; retrying in 521 ms
    warning: unable to download 'https://my-self-hosted-unreliable-cache/1p34mip9xdfr5zp12895z7wxj9pgq83m.narinfo': HTTP error 502; retrying in 1390 ms
    warning: unable to download 'https://my-self-hosted-unreliable-cache/1p34mip9xdfr5zp12895z7wxj9pgq83m.narinfo': HTTP error 502; retrying in 2140 ms
    error: unable to download 'https://my-self-hosted-unreliable-cache/1p34mip9xdfr5zp12895z7wxj9pgq83m.narinfo': HTTP error 502
    ```

- I can realise what's happening and the obvious desired behavior is to disable that cache for this build but I don't know how to do that

- I have to go and hunt down `--option substitute false`, and then do a build with no caching at all

Ideally nix could be configured to always treat cache errors as misses, this is almost always the behavior I want.

Failing that, the least it could do would be to print on stderr `Looks like a cache is failing, perhaps you want '--option ignore-cache https://my-self-hosted-unreliable-cache/'` (or even just prompt `--option substitute false`).

### LINKED ISSUE #6901: Nix fails if my local cache (substituer) is offline. Even when everything is available on the next one: cache.nixos.org
I have many machines on my local network that are using NixOS and they used to be pulling all of their dependencies from the internet (`cache.nixos.org`).

Since they are all using the almost same configuration, I setup my NAS to act as a local cache:
```nix
  nix.settings = {
    substituters = [
      "http://192.168.1.1:5000"
      "https://cache.nixos.org"
    ];
    trusted-public-keys = [
      "192.168.1.1:QwhwNrClkzxCvdA0z3idUyl76Lmho6JTJLWplKtC2ig="
    ];
  };
```

It works great, saves a lot of time, bandwidth, and resources on `cache.nixos.org`. I just need to update the NAS first.

My problem is that when the NAS is unavailable, `nix` stops working on all my machines. Same issue when I use my laptop outside of my local network.

For example:
```
$ nix shell nixos#konsole 
warning: error: unable to download 'http://192.168.1.1:5000/kbixjq5b2ddnv1vzj01knvrc5j0cbkyv.narinfo': Couldn't connect to server (7); retrying in 307 ms
warning: error: unable to download 'http://192.168.1.1:5000/kbixjq5b2ddnv1vzj01knvrc5j0cbkyv.narinfo': Couldn't connect to server (7); retrying in 520 ms
warning: error: unable to download 'http://192.168.1.1:5000/kbixjq5b2ddnv1vzj01knvrc5j0cbkyv.narinfo': Couldn't connect to server (7); retrying in 1195 ms
warning: error: unable to download 'http://192.168.1.1:5000/kbixjq5b2ddnv1vzj01knvrc5j0cbkyv.narinfo': Couldn't connect to server (7); retrying in 2116 ms
error: unable to download 'http://192.168.1.1:5000/kbixjq5b2ddnv1vzj01knvrc5j0cbkyv.narinfo': Couldn't connect to server (7)
```
and the command fails without installing `konsole`.

Then if I make the NAS available again, `nix` will successfully see that `konsole` was not present on the NAS and use `cache.nixos.org` instead.

This means that the logic to "try the next substituer" is already there, but only works when the error on the first one is a `404` but not when it's a failed connection.

I'd be happy to make a PR if someone could give me a pointer or two about where the offending code is.

### LINKED ISSUE #7127: Ignore failing binary caches
I run my own binary cache, but this is not always operational and responding. When I run nix commands when it is not available they fail and I need to manually supply substitutors

The nix commands to ignore a failing binary cache.

An alternative would be to declare only some binary caches as failable.

arcuru:
This change allows for individual substituters to be unavailable even when `fallback = false` (default). Prior to this change, _any_ unavailable substituters would cause a build failure even if others were reachable. This causes issues when people try to setup occasionally accessible binary caches.

The [fallback](https://nixos.org/manual/nix/unstable/command-ref/conf-file.html#conf-fallback) setting states that it allows Nix to fallback to building from source if a binary substitute fails, and this change makes the behavior now match the description. Prior to this change it actually did two things:

1. Allows for missing/erroring binary substituters
2. Fallback to building from source if no binary substitute is available

Allowing unreachable binary substituters should probably be the expected behavior with or without the `fallback` flag, and is what this change allows.

I've tested simple scenarios locally, but it's not clear to me how to write a test for this. I think it needs a network failure to trigger the new paths, and I'm not sure how to set that up in the tests.

Closes #6901, Closes #7127, Closes #4383

### DIFF
--- a/src/libstore/build/substitution-goal.cc
+++ b/src/libstore/build/substitution-goal.cc
@@ -110,18 +110,12 @@ void PathSubstitutionGoal::tryNext()
         tryNext();
         return;
     } catch (SubstituterDisabled &) {
-        if (settings.tryFallback) {
-            tryNext();
-            return;
-        }
-        throw;
+        tryNext();
+        return;
     } catch (Error & e) {
-        if (settings.tryFallback) {
-            logError(e.info());
-            tryNext();
-            return;
-        }
-        throw;
+        logError(e.info());
+        tryNext();
+        return;
     }
 
     if (info->path != storePath) {
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1158,10 +1158,7 @@ void LocalStore::querySubstitutablePathInfos(const StorePathCAMap & paths, Subst
             } catch (InvalidPath &) {
             } catch (SubstituterDisabled &) {
             } catch (Error & e) {
-                if (settings.tryFallback)
-                    logError(e.info());
-                else
-                    throw;
+                logError(e.info());
             }
         }
     }
--- a/src/libstore/http-binary-cache-store.cc
+++ b/src/libstore/http-binary-cache-store.cc
@@ -82,7 +82,7 @@ protected:
     void maybeDisable()
     {
         auto state(_state.lock());
-        if (state->enabled && settings.tryFallback) {
+        if (state->enabled) {
             int t = 60;
             printError("disabling binary cache '%s' for %s seconds", getUri(), t);
             state->enabled = false;


patricksjackson:
Friendly ping. I am constantly seeing this come up, most recently in https://github.com/NixOS/nix/pull/7424#issuecomment-1341629923.

Ericson2314:
Oh yes, things like this should not languish.

Ericson2314:
@edolstra based on our very brief convo this morning, I think there is a misunderstanding. The goal is *not* to fallback on building from source, but just to try all the substituters before giving up.

patricksjackson:
Thanks @Ericson2314 :)

Just for completeness:

From #7424:

> https://github.com/NixOS/nix/pull/7188 still tries them sequentially, so a 60 second timeout can be annoying. We would need to them concurrently for it to not be annoying.

If we're ok with the changes in this PR, then rewriting this codepath to check substituters in parallel is fairly straightforward. I'd be willing to rewrite this PR to include fetching the substituters in parallel if that's desired.

Also happy to wait and make that change separately (or at least log it for later), which is probably the better option.

Ericson2314:
@patricksjackson thank you very much for that offer, but I would say let's leave that for a follow-up PR? This is good a a minimal thing to discuss what behavior we want, after which doing things in parallel is mostly an optimization.

Ericson2314:
You're free to work on PR concurrently if you want, though :)

domenkozar:
Cachix was down 10 days ago for 30min, half of the time was due to this not being fixed in Nix yet.

Ironically, users of Cachix had the same issue because of which Cachix was down.

Hope that helps motivate merging/fixing this finally.

domenkozar:
@roberth @Ericson2314 could you cover this one for the next Nix team meeting?

roberth:
How can there be no test change and no test failure?

domenkozar:
It's likely that @patricksjackson is a first time contributor and needs a bit of guidance.

Ericson2314:
My understanding of the semantics:

Ignoring `querySubstitutablePathInfos` for now

### Background

query path from substitutior results:

 - Success
 - Throws `InvalidPath`: Cache is accessible, and for sure does *not* have the store object
 - Throws `SubstituterDisabled`: not sure?
 - Throws `Error`: substitutor is inaccessible, it may or may not have the store object
 
 `SubstituterDisabled` being subtle/unclear to me isn't so bad, because `SubstituterDisabled` and other `Error` are always handled the same way.
 
 ### Today
 
 The semantics are very order dependent, hard to specify except for as the exact argument:
 
 ```
 for each substitutor:
   if it has the path:
     return path
   else if InvalidPath:
     continue search
   else if SubstituterDisabled or other Error:
     if not fallback:
       give up
     else:
       keep searching
     end
   end
 end
 
 build from source
 ```
 
 Note that this means that if one subsitutor is down (`Error` or `SubstituterDisabled`) but the next one has the path, it will fail. This is not very user friendly!
 
 ## New semantics
 
 These are presented in a a more "declarative" and order-independent way.
 
New definitions for querying on sets of substitutors (note that sets are order-independent!)

1. If *any* substutitor returns success, the set of substituters has the path.
2. If *all* substitutors returns `InvalidPath`, the set of substiters doesn't have the path.

Note the "any" vs "all" duality.

Otherwise, we have a mix of `Error`/`SubstituterDisabled` and `InvalidPath`, in which case the set substitutors *maybe* has the path.

New policy based on the above defs:
```
case result on query on set of substituters:
  case set has path:
    substitute path
  case set does not have path:
    build path
  case set may have path:
    if fallback:
      build path
    else:
      error
    end
  end
end
```

fricklerhandwerk:
Discussed in Nix team meeting on 2023-01-09:

@patricksjackson thanks for your contribution! We agree that we want to fix this, but haven't decided yet what the overall process of picking substituters should look like. Therefore no decision to merge, since we don't know if that would incur changing behavior down the line, or other redundant work.
@Ericson2314 volunteered to write up a proposal and, when done, put it back on the agenda for the team to decide. We will then come back to this PR.

  - @thufschmitt: the problem with failing only after trying all substituters is that most non-trivial Nix builds will have a last wrapper that is not cached anywhere (such as `nix-shell`)
      - @Ericson2314: yes, this solution only gets us half way to fulfilling the user story
  - @roberth: how about adding a parameter for each substituter to tell if it's required?
      - @fricklerhandwerk: that requires more syntax for the configuration file
      - @roberth: we could use `?` parameters. it's not great but a simple way to do it
      - @fricklerhandwerk: more configuration is probably okay, but we should decide more deliberately on how to add more parameters. would be better if we could solve the issue in way that just works in a way that is unsurprising
      - @thufschmitt: there are two kinds of caches: the "authoritative" cache.nixos.org which one uses for 95% of the cases without which nothing works, and the "convenience" caches such as Cachix.
          - this speaks in favor of having a parameter
  - @thufschmitt: another question is about warning users on failures. verbose output may drown the messages.
  - @thufschmitt: there are two types of failures: not reachable or actual error. they should be handled differently
  - @Ericson2314: the worst thing about the current state is that order of substituters matters, because it fails on the first one. order should not matter
      - @thufschmitt: ideally we should parallelise susbtituter access, even if the code currently doesn't allow for it
      - @roberth: speculative fetching of `.narinfo` files and removing order from the configuration are different things. the latter is more complicated actually, because it adds more corner cases.
  - @fricklerhandwerk: how to move on with this PR? it seems we don't really know what the desired behavior even is
      - @roberth: let's write a specification for what the behavior should be overall
      - @Ericson2314: yes, should go backwards from desired behavior and guide by proposing test cases
      - @thufschmitt: agree, the team should make a proposal for the semantics
  - agreement: @Ericson2314 will write up a proposal and, when done, put it back on the agenda for the team to decide over. then continue discussing the implementation with @patricksjackson 

</details>

Ericson2314:
The thing I wrote above is the proposal. We should be able to resume progress on this.

Ericson2314:
@fricklerhandwerk Can you move this back to "in discussion" since I wrote the above proposal?

fricklerhandwerk:

 - @roberth: would be slightly more clear if we wrote "unknown" instead of "maybe has path"
 - @tomberek: does the unknown state provide feedback to the user about what's happening?
     - @Ericson2314: yes, we could provide a structured error message as we did with the profiles lately
     - @fricklerhandwerk: and that's up to the lower-level implementation to determine, so somewhat below the level of abstraction we care about here
     - @tomberek: some information should be visible somewhere why something is happening, such as why we're building or picking a certain substitutor
         - @fricklerhandwerk: would be debugging information for verbose output
- agreement
    - @Ericson2314 will help out getting the PR to implement the proposal

patricksjackson:
It should be obvious, but just to be clear about my feelings as the original PR author, and since I wasn't there to discuss things with you:

**I don't feel any ownership over this change,** and I expect it's easier for any one of you to handle this PR yourself anyways. Feel free to steal/close/whatever you want with this PR.

Thanks for thinking about a holistic solution!

Ericson2314:
@patricksjackson So I was requested to make this proposal by the rest of the Nix Team to help us clarify our own thoughts. It wasn't supposed to be in contrast to the PR -- quite on the contrary I it was your PR description that helped me write that post!

I think this PR is close to doing that (by "today" I meant not your PR but Nix `master`, edited to make that explicit), and any divergence is not intentional on your part. I much rather help you finish this off if you are still interested than implement myself.

Sorry again this got stuck in the team backlog for so long before we returned to it.

patricksjackson:
I'll take a look soon.

Ericson2314:
Glad to hear it!

patricksjackson:
Alright, I've now had a chance to come back and reload the context here. I have a couple of questions before I can make progress again: @Ericson2314 

First, it's unclear to me if your [design proposal](https://github.com/NixOS/nix/pull/7188#issuecomment-1375652870) is describing how you want this code to work at a high level or if you are describing what you want the actual code flow to look like.

* If it's just a high level description, then you're basically describing what the code already does after my changes? I can take a second look and clean up the error messages and the code a bit, which I think is what you're suggesting, but that's pretty much the intended existing logic after the changes already in this PR.

* If you do want the code flow to work exactly as described (querying all the substituters first then applying the logic):
  * It might make it clearer, however there are several questions I would need clarified from the Nix team concerning implementation details, so let me know if that's the approach you want to take. (for example: is it ok to leak store paths to lower-priority substituters, since that doesn't happen today? are there concerns about the extra network traffic? should we add a flag to choose substituters by latency instead of/in addition to priority? etc)
  * I would likely still need to apply this logic one substituter response at a time for performance reasons, so the benefits to readability would mainly come from refactoring.
  * This is a change I'd planned to propose as a followup anyways, before the current PR stalled; factoring out this set of code and then async querying the substituters in an intelligent way. I'd be happy to ditch the current PR and just move to that one if you preferred, but it will be more work.

Lastly, as I mentioned in the original comment, I'm unable to find how to test this in the current codebase. There appear to be no unit tests covering these modules, so do you want me to write the unit test infrastructure to test this? There are a few integration tests but even those I'm not sure how to trigger a network error within them, especially to trigger all the paths that this change would use. I suspect I'm missing something here, but I'm not very familiar with the codebase still.

patricksjackson:
> This is a change I'd planned to propose as a followup anyways, before the current PR stalled; factoring out this set of code and then async querying the substituters in an intelligent way. I'd be happy to ditch the current PR and just move to that one if you preferred, but it will be more work.

I said screw it and wrote an MVP for querying them all async, since I had this problem on my mind. It's in a separate branch on my fork: https://github.com/patricksjackson/nix/tree/substitution-set

I won't be offended if it's not needed/wanted/feasible, I felt like writing it anyways.

Ericson2314:
@patricksjackson Sorry I missed this!! :( Feel free -- encoraged -- to ping me on matrix if I am not responding to a PR to which I am assigned.

> First, it's unclear to me if your https://github.com/NixOS/nix/pull/7188#issuecomment-1375652870 is describing how you want this code to work at a high level or if you are describing what you want the actual code flow to look like.

Just at a high level :). It is supposed to be a specification which admits many compliant implementations.

> If it's just a high level description, then you're basically describing what the code already does after my changes?

I believe I have described your *intent*, but the code doesn't quite do what it says. I think the divergence is just an accident.

> I said screw it and wrote an MVP for querying them all async

Oh that's very cool! I think we can wrap this up, but then try to do that after. How's that sound?

src/libstore/build/substitution-goal.cc
```diff
@@ -110,18 +110,12 @@ void PathSubstitutionGoal::tryNext()
         tryNext();
         return;
     } catch (SubstituterDisabled &) {
-        if (settings.tryFallback) {
-            tryNext();
-            return;
-        }
-        throw;
+        tryNext();
```
Ericson2314, line 113: So I think what we need to do is stash the error here. And then at the end of the function, we re-throw the error instead of returning if nothing is found.

Does it make sense how that brings us in alignment with the spec I wrote?

src/libstore/build/substitution-goal.cc
```diff
@@ -110,18 +110,12 @@ void PathSubstitutionGoal::tryNext()
...
+        return;
     } catch (Error & e) {
-        if (settings.tryFallback) {
-            logError(e.info());
-            tryNext();
-            return;
-        }
-        throw;
+        logError(e.info());
```
Ericson2314, line 116: ditto, not sure if we should log here, or log at the end if we don't end up re-throwing.

src/libstore/local-store.cc
```diff
@@ -1158,10 +1158,7 @@ void LocalStore::querySubstitutablePathInfos(const StorePathCAMap & paths, Subst
             } catch (InvalidPath &) {
             } catch (SubstituterDisabled &) {
             } catch (Error & e) {
-                if (settings.tryFallback)
-                    logError(e.info());
-                else
-                    throw;
+                logError(e.info());
```
Ericson2314, line 1161: ditto

SuperSandro2000:
bump

bouk:
I'm picking up this changes at https://github.com/NixOS/nix/pull/8983

Magkoooh (approved):

Ericson2314:
Closed because replaced by #13301

PR #7190, opened 2022-10-18
Fix default values in the manual

### LINKED ISSUE #6359: tarball-ttl is documented as defaulting to 136 years in nix.conf(5)

patricksjackson:
The default values that show up in the manual are generated by dumping the _current_ settings using `nix show-config`. When the manual is built, that command is run in a sandbox with no internet access which caused some of the default settings to be altered before we dump the config. See #6359 for discussion and alternative options.

This change just special cases the `show-config` command to run prior to any settings being altered.

This may not be what we want to do if `nix show-config` needs to show those altered settings. If that's the case I'd suggest we add a `nix show-config --defaults` command.

Closes #6359

### DIFF
--- a/doc/manual/generate-options.nix
+++ b/doc/manual/generate-options.nix
@@ -11,16 +11,16 @@ concatStrings (map
     + concatStrings (map (s: "    ${s}\n") (splitLines option.description)) + "\n\n"
     + (if option.documentDefault
        then "    **Default:** " + (
-       if option.value == "" || option.value == []
+       if option.defaultValue == "" || option.defaultValue == []
        then "*empty*"
-       else if isBool option.value
-       then (if option.value then "`true`" else "`false`")
+       else if isBool option.defaultValue
+       then (if option.defaultValue then "`true`" else "`false`")
        else
          # n.b. a StringMap value type is specified as a string, but
          # this shows the value type.  The empty stringmap is "null" in
          # JSON, but that converts to "{ }" here.
-         (if isAttrs option.value then "`\"\"`"
-          else "`" + toString option.value + "`")) + "\n\n"
+         (if isAttrs option.defaultValue then "`\"\"`"
+          else "`" + toString option.defaultValue + "`")) + "\n\n"
        else "    **Default:** *machine-specific*\n")
     + (if option.aliases != []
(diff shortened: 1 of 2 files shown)


roberth:
> This may not be what we want to do if `nix show-config` needs to show those altered settings. If that's the case I'd suggest we add a `nix show-config --defaults` command.

The changelog says: `nix show-config` shows all configuration options and their current values.
I've used it for retrieving current values.




> add a `nix show-config --defaults` command.

Sounds good.

We'll also want to improve the documentation:

```diff
--- a/src/nix/show-config.cc
+++ b/src/nix/show-config.cc
@@ -12,5 +12,5 @@ struct CmdShowConfig : Command, MixJSON
     std::string description() override
     {
-        return "show the Nix configuration";
+        return "show the current Nix configuration";
     }
```

patricksjackson:
The implementation for this completely switched, so I'd recommend squashing this PR on merge.

On a further look, `nix show-config --json` already dumps the default values but the manpages were just being generated off the current values instead of the default. This changes the generation code to use the defaults instead.

roberth (approved):
This solves the problem.
Tested manually by building and then searching through the nix.conf manpage.

Please squash before merge

PR #7191, opened 2022-10-19
Explain how Nix handles antiquotation of paths

jherland:
Specifically, explain why Nix does not _re_evaluate paths during a `nix repl` session. This is a thing that bit me while playing around with paths and antiquotation in `nix repl` while reading the Nix language tutorial at https://nix.dev/tutorials/nix-language.

Co-authored-by: Valentin Gagarin <valentin.gagarin@tweag.io>

This straddles the boundary between _language_ and _behavior_, so I'm not sure where this is best placed. @fricklerhandwerk suggested here. See https://github.com/NixOS/nix.dev/pull/343#discussion_r998768965 for context.

### DIFF
--- a/doc/manual/src/language/values.md
+++ b/doc/manual/src/language/values.md
@@ -150,6 +150,20 @@
   recognized as a path. `a.${foo}/b.${bar}` is a syntactically valid division
   operation. `./a.${foo}/b.${bar}` is a path.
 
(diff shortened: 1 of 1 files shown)


thufschmitt (approved):
Duh, I can't believe this was missing from the manual :astonished: 

The bit about the repl is a bit a bit off indeed (esp. since that behavior is not specific to antiquotations), but we can move it later.

PR #7196, opened 2022-10-19
tests/lang: re-enable eval-okay-eq.nix tests

sternenseemann:
This is a really old test case (which was originally written before the proper Nix syntax). The tested deep comparison behavior was implemented and reverted soon after due to performance problems, but it has been restored in today's Nix again (thanks to the derivation comparison optimization, presumably).


thufschmitt (approved):
Thanks, must have been a nice bit of archeology to get back to why this test got disabled :)

PR #7203, opened 2022-10-22
Proposal: Use C++20

graham33:
Would there be an objection to using C++20 for the Nix codebase? We are working on a future patch (related to S3) that might benefit from C++ 20, so thought it was worth raising a PR to upgrade to see if it's feasible. I'm not sure what the constraints are for compatibility in various environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,4 +36,4 @@ endif
 
 include mk/lib.mk
 
-GLOBAL_CXXFLAGS += -g -Wall -include config.h -std=c++17 -I src
+GLOBAL_CXXFLAGS += -g -Wall -include config.h -std=c++20 -I src
--- a/perl/Makefile
+++ b/perl/Makefile
@@ -1,6 +1,6 @@
 makefiles = local.mk
 
-GLOBAL_CXXFLAGS += -g -Wall -std=c++17 -I ../src
+GLOBAL_CXXFLAGS += -g -Wall -std=c++20 -I ../src
 
 -include Makefile.config
 
--- a/src/libcmd/nix-cmd.pc.in
+++ b/src/libcmd/nix-cmd.pc.in
@@ -6,4 +6,4 @@ Name: Nix
 Description: Nix Package Manager
 Version: @PACKAGE_VERSION@
 Libs: -L${libdir} -lnixcmd
-Cflags: -I${includedir}/nix -std=c++17
+Cflags: -I${includedir}/nix -std=c++20
--- a/src/libexpr/nix-expr.pc.in
+++ b/src/libexpr/nix-expr.pc.in
@@ -7,4 +7,4 @@ Description: Nix Package Manager
 Version: @PACKAGE_VERSION@
 Requires: nix-store bdw-gc
 Libs: -L${libdir} -lnixexpr
-Cflags: -I${includedir}/nix -std=c++17
+Cflags: -I${includedir}/nix -std=c++20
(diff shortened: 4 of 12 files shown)


thufschmitt:
@edolstra which compilers are you refeering to? The darwin builds used to have a very old clang, but it is now using clang 11 Afaik (which technically [doesn't support the whole of c++20](https://en.cppreference.com/w/cpp/compiler_support/20), but close-enough). Is there anything else blocking us there?

edolstra:
I was mostly thinking about gcc, but it looks like the support for C++20 in gcc/libstdc++ 11 is pretty good now:

https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html
https://gcc.gnu.org/projects/cxx-status.html#cxx20

graham33:
Yes I think gcc 11's support for C++20 is pretty good. See also https://en.cppreference.com/w/cpp/compiler_support

edolstra:
Thanks, I just needed this for `std::span` on the lazy-trees branch!

roberth:
Looks like CI doesn't run on a merge commit, as this merge has broken master.

Bors could help us prevent this.

roberth:
Maybe it only happens on clang.
The build failure came from the darwin job (but the linux one was cancelled because of that fwiw).

edolstra:
It also fails on aarch64-linux, though with a different error:

> g++: error: unrecognized command line option '-std=c++20'; did you mean '-std=c++2a'?

Looks like we're using an older version of gcc on aarch64-linux...

roberth:
Not compatible with boehmgc, it seems. https://github.com/NixOS/nixpkgs/pull/174282#issuecomment-1166357537

That's rather non-optional. Maybe we're not quite ready for C++20 then. Might `c++2a` be a viable solution, despite being insufficiently descriptive? Otherwise we'll have to revert this. That may actually be the best option until the boehmgc compatibility is resolved.

edolstra:
Or we use gcc11Stdenv on aarch64-linux.

It would be rather sad to revert this because aarch64-linux is stuck indefinitely on an old version of gcc. That cannot be a maintainable situation for nixpkgs either. It kind of defeats the purpose of a *standard* environment if stdenv on some architectures has a different version of gcc...

roberth:
- https://github.com/NixOS/nixpkgs/pull/209870 lists gcc on arm as its first immediate benefit; also of interest to @Ericson2314

roberth:
I'm afraid that this upgrade will send us down a very counterproductive rabbit hole. Is there a reasonable alternative to `std::span` you could use?

tpwrules:
This might be a more relevant tracking/goal issue: https://github.com/NixOS/nixpkgs/issues/208412

TL;DR: We have a solution to upgrade GCC on `aarch64-linux` for now (upgrade the bootstrap tarball) and intend to have it in by NixOS 23.05. But it doesn't fix the underlying problems and is likely to cause the version to get stuck again. We are putting in significant effort on more comprehensive solutions so we can fix the issue for good but the solutions are quite non-trivial.

My vote would be to revert this until a solution is merged into nixpkgs. I would worry about mixing the GCC version of such a fundamental package, and there is the possibility of rare and hard to detect bugs being induced with GCC 11 (which is why it hasn't been upgraded yet).

vcunat:
The mixing sounds problematic just for packages that link against nix libs.  They will probably be forced to use the same compiler version (because C++ ABI).

PR #7206, opened 2022-10-23
Defer to SSH config files for ForwardAgent option

akiekintveld:
Currently, Nix passes `-a` when it runs commands on a remote machine via SSH, which disables agent forwarding. This causes issues when the `ForwardAgent` option is set in SSH config files, as the command line operation always overrides those.

In particular, this causes issues if the command being run is `sudo` and the remote machine is configured with the equivalent of NixOS's `security.pam.enableSSHAgentAuth` option. Not allowing SSH agent forwarding can cause authentication to fail unexpectedly.

This can currently be worked around by setting `NIX_SSHOPTS="-A"`, but we should defer to the options in the SSH config files to be least surprising for users.

### DIFF
--- a/src/libstore/ssh.cc
+++ b/src/libstore/ssh.cc
@@ -67,7 +67,7 @@ std::unique_ptr<SSHMaster::Connection> SSHMaster::startCommand(const std::string
         if (fakeSSH) {
             args = { "bash", "-c" };
         } else {
-            args = { "ssh", host.c_str(), "-x", "-a" };
+            args = { "ssh", host.c_str(), "-x" };
             addCommonSSHOpts(args);
             if (socketPath != "")
                 args.insert(args.end(), {"-S", socketPath});


akiekintveld:
To make this a bit more searchable for others, this manifested when I tried to run `nixos-rebuild switch --flake .#charon --target-host charon --use-remote-sudo`, where `charon` is configured with the NixOS `security.pam.enableSSHAgentAuth` option and has no passwords set for its users, and has `ForwardAgent yes` in its match block in my `~/.ssh/config` file. This is the error message that results when it tries to copy the closure:
```
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
```

I verified that a lack of SSH agent forwarding was the culprit by setting `NIX_SSHOPTS="-G"`, which confirms that the agent is not getting forwarded:
```
error: 'nix-store --serve' protocol mismatch from 'charon', got 'user akiekintveld
       hostname charon
[snip]
       forwardagent no
[snip]
```

thufschmitt:
Git-blaming this `-a` argument leads to https://github.com/NixOS/nix/commit/7b351da680cad41623c3f9853ff2eea3cb8e3545 (after quite some hops) which introduces it amongst some other changes without any special justification, so I assume it's mostly here as a "nice-to-have" and we can get rid of it if it breaks anything.

However I'm not sure this is the cause of your problem because Nix itself will never `sudo`. `nixos-rebuild` will if `--use-remote-sudo` is set, but that will directly call `ssh` without going through Nix (like [here](https://github.com/NixOS/nixpkgs/blob/14709714e88e7bfba52df154d7606ba961e3d897/pkgs/os-specific/linux/nixos-rebuild/nixos-rebuild.sh#L178) and a few others cases below). So if you only experience that with `nixos-rebuild` the culprit might be there. Also `nixos-rebuild` [honors `NIX_SSHOPTS`](https://github.com/NixOS/nixpkgs/blob/14709714e88e7bfba52df154d7606ba961e3d897/pkgs/os-specific/linux/nixos-rebuild/nixos-rebuild.sh#L422) so that would explain why your workaround works.

Can you actually reproduce this with plain Nix? Maybe I'm overlooking something

akiekintveld:
It looks like the `nixos-rebuild` script uses SSH in two ways, first being thru `nix-copy-closure`, where this `-a` flag would be applied, and second being as just plan `ssh` commands in the script like you saw.

I also noticed they setup connection sharing (`-o ControlMaster` and related options get set in the script). I **think** this is causing a bad interaction where it calls `nix-copy-closure` first, which opens the initial SSH connection with the `-a` option set. Then, when it does the plain `ssh` the second time it reuses the connection, so even though `-a` isn't set on that command, the `ForwardAgent` config option won't apply. This would also explain why setting `NIX_SSHOPTS="-A"` would fix it, as it would override the `-a` flag in the `nix-copy-closure` and then the reused connection would continue forwarding correctly.

If that is the case, this could also be fixed by removing the connection sharing options from `nixos-rebuild`. Personally, I'd rather remove the `-a` here as the connection sharing seems beneficial.

thufschmitt:
Urgh yes, the `ControlMaster` probably causes this to leak outside of Nix itself :/
Let's remove it then since it's indeed causing troubles.

thufschmitt (approved):

PR #7207, opened 2022-10-23
Mark flakes with .type = "flake".

### LINKED ISSUE #7186: Type tag for flakes, `_type = "flake";`

By convention, `_type` can be used to perform nominal type checking and/or matching.

This allows expression logic to improve error messages, allow extra flexibility (for compatibility or ease of use) and it allows code to express their intent more clearly.

All flakes automatically get an attribute `_type = "flake";`.

Guess based on other standard attributes, such as `sourceInfo` and `inputs`. This isn't as clear and could result in false positives and strange errors.

**Additional context**

Example from https://github.com/NixOS/nixpkgs/issues/196584; ideally it'd know for sure that the thing is a flake, so it can report an appropriate error when `nixosModules.default` is not defined.

```
  prepareImport = x:
    if x?nixosModules.default && x?inputs && x.sourceInfo
    then x.nixosModules.default
    else x;
```

shlevy:
Fixes #7186

### DIFF
--- a/src/libexpr/flake/call-flake.nix
+++ b/src/libexpr/flake/call-flake.nix
@@ -43,7 +43,7 @@ let
 
           outputs = flake.outputs (inputs // { self = result; });
 
-          result = outputs // sourceInfo // { inherit inputs; inherit outputs; inherit sourceInfo; };
+          result = outputs // sourceInfo // { inherit inputs; inherit outputs; inherit sourceInfo; type = "flake"; };
         in
           if node.flake or true then
             assert builtins.isFunction flake.outputs;


src/libexpr/flake/call-flake.nix
```diff
@@ -43,7 +43,7 @@ let
 
           outputs = flake.outputs (inputs // { self = result; });
 
-          result = outputs // sourceInfo // { inherit inputs; inherit outputs; inherit sourceInfo; };
+          result = outputs // sourceInfo // { inherit inputs; inherit outputs; inherit sourceInfo; type = "flake"; };
```
roberth, line 46:
```suggestion
          result = outputs // sourceInfo // { inherit inputs; inherit outputs; inherit sourceInfo; _type = "flake"; };
```

In Nixpkgs/lib, `isType`, `setType`, and the module system `mk*` constructors all use `._type`.
We also have `pkgs._type = "pkgs"` in Nixpkgs.
`.type` is only used for derivations, afaict.

roberth (changes requested):

src/libexpr/flake/call-flake.nix
shlevy: Yes, but derivations are the only one of these that are generated by Nix itself. I think the reason the others use `_type` is precisely that `type` is reserved for Nix.

roberth: Maybe `_type` was chosen to avoid interfering with domain-specific `type` attributes, such as `(mkOption foo).type`.
Nix also uses `type` in a domain-specific way (as opposed to poor man's nominal typing) in the `fetchTree` arguments.
To illustrate this idea, I think it would make sense for `fetchTree` to return `{ _type = "tree"; type = "git"; ... }`.

shlevy: I defer to @edolstra on the attr name here.

edolstra: I guess `_type` is better because it has a smaller chance of clashing with user identifiers. Probably `derivation` should have used that as well.

blaggacao:
cc @gytis-ivaskevicius :smile:

PR #7212, opened 2022-10-24
Improve --profile description

wentasah:
The description of the `--profile` option talks about the "update" operation. This is probably meant for operations such as `nix profile install`, but the same option is reused in other subcommands, which do not update the profile, such as `nix profile {list,history,diff-closures}`.

We update the description to make sense in both contexts.


thufschmitt (approved):
Thanks, that's better indeed :)

PR #7217, opened 2022-10-25
Fix `nix __build-remote`

### LINKED ISSUE #7193: nix build is failing with: 'nix __build-remote' is not a recognised command
**Describe the bug**

When running nix build to evaluate a NixOS system config in an aarch64 container, it fails with

```
error: 'nix __build-remote' is not a recognised command
Try '/nix/store/0lrf6danhdqjsrhala134ak8vn0b9ghj-nix-2.11.1/bin/nix --help' for more information.

error: unexpected EOF reading a line
         … while reading the response from the build hook
```

The output of `nix show-config | grep build-hook` is
```
build-hook = /nix/store/0lrf6danhdqjsrhala134ak8vn0b9ghj-nix-2.11.1/bin/nix __build-remote
```

The command `__build-remote` does not exist.

**Steps To Reproduce**

The only environment where I can reproduce this behaviour is in my github action where I install nix in an aarch64 container, see https://github.com/Gerschtli/nix-config/actions/runs/3280828039/jobs/5402088955.

**Expected behavior**

No error should occur and the build should succeed.

nix-env (Nix) 2.11.1

**Additional context**

Related change: https://github.com/NixOS/nix/pull/6709

thufschmitt:
Because of a wrong index, `nix __build-remote` wasn't working.

Fix the index to restore the command (and the build hook).

@edolstra any idea why the tests (and apparently most stuff since no one complained until last week) were still OK despite that? Are we using something else in the general case?

Fix #7193 

cc @Gerschtli

### DIFF
--- a/src/libstore/build/hook-instance.cc
+++ b/src/libstore/build/hook-instance.cc
@@ -16,11 +16,11 @@ HookInstance::HookInstance()
     buildHookArgs.pop_front();
 
     Strings args;
+    args.push_back(std::string(baseNameOf(buildHook)));
 
     for (auto & arg : buildHookArgs)
         args.push_back(arg);
 
-    args.push_back(std::string(baseNameOf(settings.buildHook.get())));
     args.push_back(std::to_string(verbosity));
 
     /* Create a pipe to get the output of the child. */
--- a/src/nix/main.cc
+++ b/src/nix/main.cc
@@ -270,7 +270,7 @@ void mainWrapped(int argc, char * * argv)
     programPath = argv[0];
     auto programName = std::string(baseNameOf(programPath));
 
-    if (argc > 0 && std::string_view(argv[0]) == "__build-remote") {
+    if (argc > 1 && std::string_view(argv[1]) == "__build-remote") {
         programName = "build-remote";
         argv++; argc--;
     }


edolstra:
It works because `HookInstance::HookInstance()` does pass `__build-remote` as `argv[0]`:
```
starting build hook '/nix/store/pxd89fpbcssdl2pbbf0yaggs7rn3pfrr-nix-2.12.0pre20220916_a38a55b/bin/nix __build-remote'
[pid 3789649] execve("/nix/store/pxd89fpbcssdl2pbbf0yaggs7rn3pfrr-nix-2.12.0pre20220916_a38a55b/bin/nix", ["__build-remote", "nix __build-remote", "7"], 0x7ffea5f83ac8 /* 94 vars */) = 0
```

BTW the second argument (`"nix __build-remote"`) should probably be removed, i.e. the line
```
    args.push_back(std::string(baseNameOf(settings.buildHook.get())));
```

thufschmitt:
Ooooh, I see. Is that intended or is that accidental? Because from what I see, what happens is that

- The `buildHook` setting defaults to `/nix/store/abc-nix __build-remote` (more precisely `getSelfExe() + __build-remote`)
- The logic to run it is roughly:
  - The setting is tokenized into a list of space-separated words
  - The exe to run is the first element of that list
  - The other elements are pushed into `argv` (but not the executable itself)
  - Then the basename of the whole hook setting (`nix __build-remote`) is pushed
  - Then extra arguments are passed by Nix

I imagine the intended logic is something along the lines of
- The setting is tokenized into a list of space-separated words
- The exe to run is the first element of that list
- Then the basename of this first element (`nix`) is pushed as `argv[0]`
- The other elements are pushed into `argv`
- Then extra arguments are passed by Nix

Which would led to call `execve("/nix/store/abc-.../nix", ["nix", "__build-remote", ... ])` which seems more logical.

Does that make sense? Or is there another reason for the way we call the build hook?

thufschmitt:
> To be honest, I don't remember why I did it this way, but probably part of it was to ensure that libexec/nix/build-remote (which is a symlink to bin/nix) still works.

Makes sense :)

I think it was broken nonetheless, right?

At any rate, I've changed the logic to call `["nix", "__build-remote", ...]` which seems better. Let me know whether that looks good

PR #7222, opened 2022-10-26
builtins.substring: fix int overflow

yorickvP:
repro: `builtins.substring 4294967296 1 "umu"`

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -170,10 +170,16 @@ static I forceIntChecked(EvalState & state, Value & v, const PosIdx pos, std::st
 
     try {
         return boost::numeric_cast<I>(result);
-    } catch(boost::bad_numeric_cast&) {
-        state.error("integer %1% is out of bounds", result)
-            .withTrace(pos, errorCtx)
-            .debugThrow<TypeError>();
+    } catch(boost::bad_numeric_cast& e) {
+        if (dynamic_cast<boost::numeric::negative_overflow*>(&e)) {
+            state.error("integer %1% is too low", result)
+                .withTrace(pos, errorCtx)
+                .debugThrow<TypeError>();
+        } else {
+            state.error("integer %1% is too high", result)
+                .withTrace(pos, errorCtx)
+                .debugThrow<TypeError>();
+        }
     }
 }
 
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -25,6 +25,7 @@
 #include <dlfcn.h>
 
 #include <cmath>
+#include <boost/numeric/conversion/cast.hpp>
 
 
 namespace nix {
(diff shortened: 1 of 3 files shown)


yorickvP:
cc @puckipedia

thufschmitt (changes requested):
Care to add a little testcase for it? ([here](https://github.com/nixos/nix/blob/ac0fb38e8a5a25a84fa17704bd31b453211263eb/src/libexpr/tests/primops.cc) is probably the best place to do it)

Looks good otherwise :+1:

src/libexpr/primops.cc
```diff
@@ -3424,7 +3424,7 @@ static void prim_substring(EvalState & state, const PosIdx pos, Value * * args,
             .errPos = state.positions[pos]
         }));
 
-    v.mkString((unsigned int) start >= s->size() ? "" : s->substr(start, len), context);
+    v.mkString((unsigned NixInt) start >= s->size() ? "" : s->substr(start, len), context);
```
edolstra, line 3427: This causes a compilation failure on macOS. Wouldn't it make more sense to cast it to `size_t`, since we're comparing to `s->size()`?

yorickvP: That's annoying, a cast to size_t would overflow on 32-bit systems. I've now added a forceIntChecked function that uses `boost::numeric_cast` to do a checked cast to size_t where applicable in the primops.

thufschmitt (commented):
The change looks good overall, but it feels like the error messages are slightly worse than what they used to be (when evaluating something like `builtins.genList (x: x) (-1) for example`). 

It might be nicer to make `forceIntChecked` return an optional or equivalent rather than throwing, and keeping the error handling on the client side where we can get some better messages

fricklerhandwerk (approved):
The additions to the error messages are sensible, very good. Only added suggestions on phrasing, but this is cherry on top.

thufschmitt:
On the topic of error messages, I'd just like to point out a marginal regression wrt what we have on master (I don't have a strong opinion as to whether that should be a blocker or not, just want to make sure that's not an overlook):

```console
$ # On master                                   
$ nix eval --expr 'builtins.genList (x: x) (-1)'
error:
       … while calling the 'genList' builtin

         at «string»:1:1:

            1| builtins.genList (x: x) (-1)
             | ^

       error: cannot create list of size -1
$ # On this branch
$ nix run github:yorickvP/nix/fix-substring-int -- eval --expr 'builtins.genList (x: x) (-1)'
error:
       … while calling the 'genList' builtin

         at «string»:1:1:

            1| builtins.genList (x: x) (-1)
             | ^

       … while evaluating the second argument passed to builtins.genList

         at «none»:0: (source not available)

       error: integer -1 is too low
```

roberth:
Are you still interested to work on this problem?
The code has changed quite a bit due to error message improvements. We now also have a type wrapper for checked arithmetic in `src/libutil/checked-arithmetic.hh`.
I'd be happy to review it.

tomberek:

PR #7229, opened 2022-10-27
Auto-assign reviewers by file

Mic92:
This was discussed in the documentation meeting to get quicker reviewers on manual related things.

### DIFF
--- a/.github/workflows/assign-reviewer.yml
+++ b/.github/workflows/assign-reviewer.yml
@@ -1,0 +1,12 @@
+name: "Auto Assign"
+on:
+  - pull_request
+
+jobs:
+  assign_reviewer:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: shufo/auto-assign-reviewer-by-files@v1.1.4
+        with:
+          config: ".github/assign-by-files.yml"
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/assign-by-files.yml
+++ b/.github/assign-by-files.yml
@@ -1,0 +1,5 @@
+---
+# This files is used by https://github.com/marketplace/actions/auto-assign-reviewer-by-files
+# to assign maintainers
+"doc/**/*":
+  - fricklerhandwerk


.github/workflows/assign-reviewer.yml
```diff
@@ -0,0 +1,12 @@
+name: "Auto Assign"
+on:
+  - pull_request
```
Mic92, line 3: Not sure if this token can even assign reviewers.
@zowoq I guess I need pull_request_target here?

zowoq: https://github.com/shufo/auto-assign-reviewer-by-files#use-of-team-reviewers-results-in-could-not-resolve-to-a-node-with-the-global-id-of-error

I think `pull_request` should work, only looks like it's problematic with teams. As it's `pull_request` should be able to test it in this PR?

Like the rest of the actions in this repo `pull_request` won't run automatically for new contributors so CODEOWNERS may be a better option here if you want consistent notifications.

fricklerhandwerk: > `pull_request` won't run automatically for new contributors

What exactly does this mean? Will it not run if a new contributor makes a pull request? Why?

zowoq: https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/

Mic92: > What exactly does this mean? Will it not run if a new contributor makes a pull request? Why?

Github action CI are skipped for first-time contributors.

Mic92: > https://github.com/shufo/auto-assign-reviewer-by-files#use-of-team-reviewers-results-in-could-not-resolve-to-a-node-with-the-global-id-of-error
> 
> I think `pull_request` should work, only looks like it's problematic with teams. As it's `pull_request` should be able to test it in this PR?
> 
> Like the rest of the actions in this repo `pull_request` won't run automatically for new contributors so CODEOWNERS may be a better option here if you want consistent notifications.

Yes but CODEOWNERS requires write access to the repo, which @fricklerhandwerk does not have. So this is the second best solution. At least once someone allows CI to be run for a PR than it will assign him.

fricklerhandwerk: In my case I do have merge access, but we want more maintainers for documentation, who can review changes and want to be notified for that purpose. But they don't necessarily have to merge them themselves.

It may be fine to have this as an intermediate solution if someone wants click on the button for every PR. Not sure if I can and want to afford the time.

@edolstra what do you think about giving the current docs team lead merge access and make it their responsibility? @LucPerkins would it be okay for you to take this on?

fricklerhandwerk (approved):
- @thufschmitt: we need this because `CODEOWNERS` requires commit access
    - @edolstra: also it makes an implication of ownership
    - @fricklerhandwerk: why is that a problem?
        - @edolstra: don't want a too large group of people to have veto power over the code
    - @thufschmitt: we can note this in the file, that it does not mean actual onership?
    - @fricklerhandwerk: we need both code ownership (and being on the Nix team should come with the right and duty to own code) and a means to notify people who care about changes but don't have merge access
        - @tomberek: agreed, we need both
- @thufschmitt: it's quite an expensive routine, but this is GitHub's cost
    - also it's silly that maintainers have to press the button on PRs by outside contributors, but we should be able to handle it
   - @edolstra: can we have the same thing as in Nixpkgs?
     - @fricklerhandwerk: there a bot reads the `maintainers` attribute on packages, we can't do that here
     - @thufschmitt: a bot could read `git blame` and assign automatically based on that
       - @fricklerhandwerk: that would not help notifying people interested in tracking changes and reviewing code *before* they got anything merged
- @edolstra: this is a good change, let's merge this

# Decision
undraft and merge

thufschmitt:
We've just discussed that as part of the weekly Nix team meeting. Long-story short: This is great, let's merge it.

- We all want something like that.
- `CODEOWNERS` could be great, but
    - The requirement of giving everyone commit access isn't awesome
    - It implies "ownership" which is a different concept than what we want here (just ping potential reviewers)
- The requirement of spinning up a whole ubuntu machine just to assign people is a bit silly, but so be it
- We can add more rules to the `assign-by-files` file later

@Mic92 is there anything you'd like to do on this PR before we merge it? Otherwise feel free to un-draft it and I'll merge

thufschmitt:
Personal comment: Could you document it somewhere?

I'm not sure what would be the best place for that (cc @fricklerhandwerk). That probably deserves a custom section under https://github.com/NixOS/nix/tree/master/doc/manual/src/contributing/

Mic92:
> @Mic92 is there anything you'd like to do on this PR before we merge it? Otherwise feel free to un-draft it and I'll merge

Ok from my side.

edolstra:
This doesn't work yet, I guess we need to configure some token or whatever (https://github.com/NixOS/nix/actions/runs/3445124965/jobs/5748478961).

.github/workflows/assign-reviewer.yml
zowoq: Sorry, looks like this did need to be `pull_request_target`. 

https://github.com/NixOS/nix/pull/7398

PR #7232, opened 2022-10-28
use bats for integration tests

rapenne-s:
this is a PR to modify the makefiles to use `bats` to handle integration tests, this PR can be used to compare with `shellspec` testing framework PR available at https://github.com/NixOS/nix/pull/7135


Ericson2314:
I don't get what is good abou tthis or #7135. It is very nice to have tests as separate make jobs so that parallelism is handled in a consistent way. The old setup where the integration tests did not run in parallel was terrible!

rapenne-s:
The current PRs are just for transitioning the tests toward a new framework, the tests will have to be rewritten. The PRs mostly just run the existing tests through a new framework, that's just the foundation to build upon it new tests.

You will get more explanation in the first message of https://github.com/NixOS/nix/pull/7135#issue-1397878268

Ericson2314:
OK I get the motivation for individual tests somewhat, thanks, but I would still prefer to leverage make to be able to run the tests in parallel.

PR #7234, opened 2022-10-28
Fix user namespace detection with old kernels

rodarima:
The current mechanism to detect the presence of user namespaces assumes they are not available when the file `/proc/sys/user/max_user_namespaces` doesn't exist. This file is only present since Linux 4.9, so older versions which still support namespaces cause the check to return false. In particular, our machine has the version 4.4.

This patch adds another detection check when the max_user_namespaces fails, by testing the presence of `/proc/self/ns/user`, which should exist since 3.8 (see namespaces(7)).

The particular problem occured to @aleixrocks when running some python tests that call `getgrgid`. The group id doesn't have an entry in /etc/group when the user namespaces are not in use, which causes an error in getgrgid.

This is a reproducer we used to test (with an old nixpkgs version), where we use getent instead, to get the information of the effective group:

```nix
let
  commit = "1614b96a68dd210919865abe78bda56b501eb1ef";

  # Pin the nixpkgs
  nixpkgsPath = builtins.fetchTarball {
    # Descriptive name to make the store path easier to identify
    name = "nixpkgs-${commit}";
    # Commit hash for nixpkgs as of 2021-09-01
    url = "https://github.com/nixos/nixpkgs/archive/${commit}.tar.gz";
    # Hash obtained using `nix-prefetch-url --unpack <url>`
    sha256 = "17c4a6cg0xmnp6hl76h8fvxglngh66s8nfm3qq2iqv6iay4a92qz";
  };

  pkgs = import nixpkgsPath { };

  test = pkgs.stdenv.mkDerivation {
    name = "test";
    buildInputs = [ pkgs.glibc.bin ];
    src = ./repro.nix;
    dontUnpack = true;
    buildPhase = ''
      set -x
      cat /etc/group
      cat /etc/passwd
      id -g
      id -gr
      cat /proc/self/uid_map
      cat /proc/self/gid_map
      getent group $(id -g) || echo not found
      getent group 100 || echo not found
      set +x
    '';
  };

in test
```

The difference of the output is as follows when applying the patch:

```diff
--- bad 2022-10-28 17:39:41.389428541 +0200
+++ ok  2022-10-28 17:40:33.740893984 +0200
@@ -11,20 +11,19 @@
 nogroup:x:65534:
 ++ cat /etc/passwd
 root:x:0:0:Nix build user:/build:/noshell
-nixbld:x:30001:30000:Nix build user:/build:/noshell
+nixbld:x:1000:100:Nix build user:/build:/noshell
 nobody:x:65534:65534:Nobody:/:/noshell
 ++ id -g
-30000
+100
 ++ id -gr
-30000
+100
 ++ cat /proc/self/uid_map
-         0          0 4294967295
+      1000      30001          1
 ++ cat /proc/self/gid_map
-         0          0 4294967295
+       100      30000          1
 +++ id -g
-++ getent group 30000
-++ echo not found
-not found
+++ getent group 100
+nixbld:!:100:
 ++ getent group 100
 nixbld:!:100:
 ++ set +x

```
The uid and gid maps are set properly after the patch and now the getent command succeeds.

This test also suggest that running the build without user namespaces may result in non reproducible builds, as the user id of the nix builder is visible from the build sandbox.

### DIFF
--- a/src/libutil/namespaces.cc
+++ b/src/libutil/namespaces.cc
@@ -18,19 +18,28 @@ bool userNamespacesSupported()
         }
 
         Path maxUserNamespaces = "/proc/sys/user/max_user_namespaces";
-        if (!pathExists(maxUserNamespaces) ||
-            trim(readFile(maxUserNamespaces)) == "0")
-        {
-            debug("user namespaces appear to be disabled; check '/proc/sys/user/max_user_namespaces'");
-            return false;
+        if (pathExists(maxUserNamespaces) {
+            /* If the file exists, we should respect what it says. */
+            if (trim(readFile(maxUserNamespaces)) == "0") {
+                debug("user namespaces appear to be disabled; check '/proc/sys/user/max_user_namespaces'");
+                return false;
+            }
+        } else {
+            /* Older Linux kernels never produce this file, whether or
+               not user namespaces are available. Assume we are running
+               an older kernel and continue. */
         }
 
         Path procSysKernelUnprivilegedUsernsClone = "/proc/sys/kernel/unprivileged_userns_clone";
-        if (pathExists(procSysKernelUnprivilegedUsernsClone)
-            && trim(readFile(procSysKernelUnprivilegedUsernsClone)) == "0")
-        {
-            debug("user namespaces appear to be disabled; check '/proc/sys/kernel/unprivileged_userns_clone'");
-            return false;
+        if (pathExists(procSysKernelUnprivilegedUsernsClone) {
+            /* If the file exists, we should respect what it says. */
+            if (trim(readFile(procSysKernelUnprivilegedUsernsClone)) == "0")
+            {
+                debug("user namespaces appear to be disabled; check '/proc/sys/kernel/unprivileged_userns_clone'");
+                return false;
+            }
+        } else {
+            /* Assume all good, just like above. */
         }
 
         try {


rodarima:
I modified the patch to only check once (following the previous behavior) and also to only attempt to check for `/proc/self/ns/user` when there is no `/proc/sys/user/max_user_namespaces` (ignoring when it exists and is set to zero).

CC @edolstra

src/libstore/build/local-derivation-goal.cc
```diff
@@ -803,9 +803,21 @@ void LocalDerivationGoal::startBuilder()
...
         Path maxUserNamespaces = "/proc/sys/user/max_user_namespaces";
-        static bool userNamespacesEnabled =
-            pathExists(maxUserNamespaces)
-            && trim(readFile(maxUserNamespaces)) != "0";
+        Path procSelfNsUser = "/proc/self/ns/user";
+        static bool userNamespacesEnabled = false;
+        static bool userNamespacesTested = false;
+
+        /* Only check once */
+        if (!userNamespacesTested) {
```
thufschmitt, line 811: Could this be replaced by a `static` variable?

src/libstore/build/local-derivation-goal.cc
```diff
@@ -803,9 +803,21 @@ void LocalDerivationGoal::startBuilder()
...
+            if (pathExists(maxUserNamespaces)) {
+                userNamespacesEnabled = (trim(readFile(maxUserNamespaces)) != "0");
```
thufschmitt, line 813: Do we still need to check that or is `/proc/self/ns/user` enough?

thufschmitt (commented):
Thanks :)

A couple of questions, but looks good overall

src/libstore/build/local-derivation-goal.cc
rodarima: Not sure what you mean, it is already declared as static variable in line 808, so the check should only run once (as far as I know, as I'm assuming the same behavior as in C). Do you mean a static member of `LocalDerivationGoal`?

thufschmitt: Yes sorry that wasn't very clear :upside_down_face: 

What I meant is moving all that logic in the initialization of the `static` variable. Something like :

```c++
static bool userNamespacesEnabled = [&](){
    if (pathExists(maxUserNamespaces)) {
        return trim(readFile(maxUserNamespaces)) != "0";
    } else {
        /* Linux kernel versions < 4.9 lack
         * /proc/sys/user/max_user_namespaces */
        return pathExists(procSelfNsUser);
    } 
}();
```

(If you're more used to C, the `[&](){ blah }()` syntax is just creating an anonymous lambda and applying it on the fly, which is the less-wrong way of initializing variables with a complex expression in C++)

src/libstore/build/local-derivation-goal.cc
rodarima: I think checking only for `/proc/self/ns/user` at the beginning seems reasonable. It would also solve the problem when the max namespaces initially are set to 0 and then become available again. I will modify the patch accordingly.

src/libstore/build/local-derivation-goal.cc
rodarima: Oh, I see. Well removing the max namespaces check as commented below results in a much simpler initialization that doesn't require the lambda :-)

thufschmitt (approved):
Thanks, that looks good.

I wouldn't mind someone else (@edolstra) having a second look and confirming that we can skip the check for `max_user_namespaces` since that kind of things is barely documented (or I just can't find the right places)

src/libstore/build/local-derivation-goal.cc
```diff
@@ -802,10 +802,8 @@ void LocalDerivationGoal::startBuilder()
 
         userNamespaceSync.create();
 
-        Path maxUserNamespaces = "/proc/sys/user/max_user_namespaces";
-        static bool userNamespacesEnabled =
-            pathExists(maxUserNamespaces)
-            && trim(readFile(maxUserNamespaces)) != "0";
```
thufschmitt, line 808: If the file exists, then we should check that its content is non zero as it would otherwise mean that the namespaces are enabled in the kernel but disabled for the current user

src/libstore/build/local-derivation-goal.cc
```diff
@@ -803,8 +803,13 @@ void LocalDerivationGoal::startBuilder()
         userNamespaceSync.create();
 
         Path procSelfNsUser = "/proc/self/ns/user";
+        Path maxUserNamespaces = "/proc/sys/user/max_user_namespaces";
+
         static bool userNamespacesEnabled = pathExists(procSelfNsUser);
 
+        if (userNamespacesEnabled && pathExists(maxUserNamespaces))
+            userNamespacesEnabled = trim(readFile(maxUserNamespaces)) != "0";
```
edolstra: Maybe simplify this to
```
static bool userNamespacesEnabled = 
    pathExists(procSelfNsUser)
    && (!pathExists(maxUserNamespaces) 
       || trim(readFile(maxUserNamespaces)) != "0");
```

src/libstore/build/local-derivation-goal.cc
```diff
@@ -802,10 +802,13 @@ void LocalDerivationGoal::startBuilder()
...
+        Path procSelfNsUser = "/proc/self/ns/user";
         Path maxUserNamespaces = "/proc/sys/user/max_user_namespaces";
-        static bool userNamespacesEnabled =
-            pathExists(maxUserNamespaces)
-            && trim(readFile(maxUserNamespaces)) != "0";
+
+        static bool userNamespacesEnabled = pathExists(procSelfNsUser);
+
+        if (userNamespacesEnabled && pathExists(maxUserNamespaces))
+            userNamespacesEnabled = trim(readFile(maxUserNamespaces)) != "0";
```
fricklerhandwerk, line 811:
```suggestion
        Path procSelfNsUser = "/proc/self/ns/user";
        Path maxUserNamespaces = "/proc/sys/user/max_user_namespaces";

        static bool userNamespacesEnabled =
            // old kernels support this file, so this check always works
            pathExists(procSelfNsUser)
            // `max_user_namespaces` does not exist on some old kernels, but respect it if it does
            && (!pathExists(maxUserNamespaces) || trim(readFile(maxUserNamespaces)) != "0");
```

rodarima:
On Mon, Jun 19, 2023 at 04:34:22AM -0700, Eelco Dolstra wrote:
>@edolstra commented on this pull request.
>
>
>
>>          static bool userNamespacesEnabled = pathExists(procSelfNsUser);
>
>+        if (userNamespacesEnabled && pathExists(maxUserNamespaces))
>+            userNamespacesEnabled = trim(readFile(maxUserNamespaces)) != "0";
>
>Maybe simplify this to
>```
>static bool userNamespacesEnabled =
>    pathExists(procSelfNsUser)
>    && (!pathExists(maxUserNamespaces)
>       || trim(readFile(maxUserNamespaces)) != "0");
>```

AFAIK that would compute the value at the start and not perform the test 
more than once. My intention was to always check for the content of the 
max_user_namespaces, as it can vary during the execution of the 
nix-daemon (it will also help debugging problems via strace).

But let's not stagnate this PR more, as I was trying to solve a 
particular issue with old kernels and is not affected by this.

I applied your patch.

rodarima:
So... I was rebasing with NixOS/nix:master to pass the tests and it 
seems that it is already fixed with the way `bool 
userNamespacesSupported()` checks for namespaces:

```cpp
bool userNamespacesSupported()
{
     static auto res = [&]() -> bool
     {
         if (!pathExists("/proc/self/ns/user")) {
             debug("'/proc/self/ns/user' does not exist; your kernel was likely built without CONFIG_USER_NS=y");
             return false;
         }

         Path maxUserNamespaces = "/proc/sys/user/max_user_namespaces";
         if (!pathExists(maxUserNamespaces) ||
             trim(readFile(maxUserNamespaces)) == "0")
         {
             debug("user namespaces appear to be disabled; check '/proc/sys/user/max_user_namespaces'");
             return false;
         }

         Path procSysKernelUnprivilegedUsernsClone = "/proc/sys/kernel/unprivileged_userns_clone";
         if (pathExists(procSysKernelUnprivilegedUsernsClone)
             && trim(readFile(procSysKernelUnprivilegedUsernsClone)) == "0")
         {
             debug("user namespaces appear to be disabled; check '/proc/sys/kernel/unprivileged_userns_clone'");
             return false;
         }

         ...
```

The machine where this was a problem got updated to a recent kernel, so 
I cannot check there easily.

Feel free to close this PR then, I will open another PR it the issue 
persists.

Ericson2314:
Uh oh! I am afraid us reviewers forgot to check what happened in the meantime: this code was factored out and cleaned up much in the way you described.

I discovered this only now when rebasing and trying to fix conflicts.

I pushed something else trying to keep the comments you might like, feel free to change it how you see fit, or just close this.

Sorry for wasting your time, @rodarima.

rodarima:
No problem, thanks for your comments :-)

PR #7240, opened 2022-10-30
Add doas support to the installation script.

ForeverNooob:
Previously the install errored out and told you to manually add `/nix/` and set the perms if you had `doas` installed instead of `sudo`. Well, not any more!

### DIFF
--- a/scripts/install-nix-from-closure.sh
+++ b/scripts/install-nix-from-closure.sh
@@ -137,7 +137,7 @@ echo "performing a single-user installation of Nix..." >&2
 if ! [ -e "$dest" ]; then
     cmd="mkdir -m 0755 $dest && chown $USER $dest"
     echo "directory $dest does not exist; creating it by running '$cmd' using sudo" >&2
-    if ! sudo sh -c "$cmd"; then
+    if ! (sudo sh -c "$cmd" || doas sh -c "$cmd") ; then
         echo "$0: please manually run '$cmd' as root to create $dest" >&2
         exit 1
     fi


scripts/install-nix-from-closure.sh
```diff
@@ -137,7 +137,7 @@ echo "performing a single-user installation of Nix..." >&2
 if ! [ -e "$dest" ]; then
     cmd="mkdir -m 0755 $dest && chown $USER $dest"
     echo "directory $dest does not exist; creating it by running '$cmd' using sudo" >&2
-    if ! sudo sh -c "$cmd"; then
+    if ! (sudo sh -c "$cmd" || doas sh -c "$cmd") ; then
```
LunNova, line 140: nit: if you have both sudo and doas and sudo fails it'll move on to doas, is that right?

Not sure if even matters.

niksingh710: > Shouldn't we check if there is sudo or doas executable and then use that rather than trying one and then the other?

Yes, That will be correct way.

```
hash sudo 2>/dev/null && rootmgr="sudo"
hash doas 2>/dev/null && rootmgr="doas"
[ -z $rootmgr ] && {
    echo "How you are running the system without any rootmgr!!"
    echo "Are you a wizard?"
    exit 1
}
```

knl: Just want to add that in my company we don't have sudo nor doas, but dzdo. Could it be possible to specify root manager during the installation?

154pinkchairs: yeah, POSIX sh has support for `read -r`, which is roughly equivalent to `scanf()` in C. Then we can simply use `which`and try searching the path for the provided binary. 

Though giving too much leeway here feels instinctively dangerous. Without a predefined list of root managers which would be required to sanitize the input, we'd potentially open the floodgates to arbitrary code execution, or exploits where if the script doesn't implement proper access controls, an attacker could potentially gain unauthorized access by specifying a root manager they have control over. 

It may also break the least privilege principle, and actually I believe if doas is found in the system the script should make it the default, with an option to use sudo annotated with a warning that it is potentially less secure. Though that'd require checks to ensure required level of robustness with doas, since many commands and programs, like yay perform a fatal exit when an incorrect password is entered even once, which even though the script does not take much time to run on most computers, would probably annoy most users.

bjornfor: > yeah, POSIX sh has support for `read -r`, which is roughly equivalent to `scanf()` in C. Then we can simply use `which`and try searching the path for the provided binary.

`which` isn't part of POSIX (so not guaranteed to exist), but `command -v` is.

tomberek:

Issue #7107, opened 2022-09-27
Flakes: Not including untracked files is confusing for unaware users.


ElvishJerricco:
As flakes become more and more common, so too is the issue that untracked files aren't included when your flake is copied to the nix store. Personally, I consider this behavior a good thing, as it prevents me from accidentally copying `result` symlinks or `nixos.qcow2` images into the store unnecessarily. But in my experience assisting users on Matrix and Discourse, it's a common error to not track a necessary file with git. This manifests with users being very confused why their shiny new code seems to not exist at all.

I'm not sure what the answer is here. Again, I personally prefer the current behavior, but clearly it's a problem for people who are unaware of the cause. I figured there should be an open issue to at least discuss the matter.

silky:
I'd honestly be happy/expect to see this pop up when running adding some verbosity to the commands; it's what I tried to do to debug the issue, and was left confused when additional `v's` didn't yield me any particularly relevant hints ....

Kha:
Relevant: https://github.com/NixOS/nix/pull/6858#issuecomment-1203872793 and following comments

squalus:
For anyone else running into this, here's how to include untracked files:
`nix build path:///path/to/myflake#mypackage`.

The entire flake directory gets copied to the Nix store in this case.

Atemu:
With https://github.com/NixOS/nix/pull/6530, the problem of accidentally copying large untracked files should no longer exist.

I think it'd make sense to disable the unintuitive git integration then.

I'm all for integrating different tools with another but that should always be an optional feature. A build tool should not care about which VCS I'm using or what its state is; especially not by default. (Being able to fetch a build expression via git is a great feature though.)

andrewhamon:
I too was pretty surprised at the tight git integration, and it makes certain things impossible, like conditionally importing an untracked file that may or may not exists (i.e. poor mans secret management, or a way to let folks customize a nix shell in a large shared repo).

silky:
Depending on your setup @andrewhamon you might be interested in something like this - https://github.com/hasura/graphql-engine/blob/master/.envrc#L59 - to source a local .envrc file.

bergkvist:
Even being familiar with this behavior, I find it cumbersome/unintuitive, and it is one of my biggest barriers to moving from traditional nix over to flakes.

[...]

PR #7244, opened 2022-10-31
Warn about untracked files in git repositories

rti:
This adds a warning about untracked files in fetched git repositories. Addresses #7107 and hopefully helps to avoid some confusion. Maybe this can help as a simple fix before #6530 gets merged.

### DIFF
--- a/src/libfetchers/git.cc
+++ b/src/libfetchers/git.cc
@@ -179,6 +179,21 @@ WorkdirInfo getWorkdirInfo(const Input & input, const Path & workdir)
     bool clean = false;
     bool hasHead = exitCode == 0;
 
+    auto statusResult = runProgram(RunOptions {
+        .program = "git",
+        .args = { "-C", workdir, "--git-dir", gitDir, "status", "--untracked-files=normal", "--short"},
+        .environment = env,
+        .mergeStderrToStdout = true
+    });
+
+    if (WEXITSTATUS(statusResult.first) != 0) {
+        throw Error("getting status of Git repository '%s'", workdir);
+    }
+
+    if (std::regex_search(statusResult.second, gitUntrackedRegex)) {
+        warn("Git tree '%s' has untracked files that are ignored by nix", workdir);
+    }
+
     try {
         if (hasHead) {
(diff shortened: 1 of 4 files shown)


SuperSandro2000:
Would this be disabled with ``warn-dirty = false``?

rti:
> Would this be disabled with `warn-dirty = false`?

At the moment, not. Should it?

edolstra:
I'd like to postpone this until after https://github.com/NixOS/nix/pull/6530 is merged to avoid creating merge conflicts.

Artturin:
> > Would this be disabled with `warn-dirty = false`?
> 
> At the moment, not. Should it?

yeah that'd be nice, some users have `warn-dirty = false` because we don't need these warnings.

rti:
As I understand it, https://github.com/NixOS/nix/pull/6530 introduces a warning for files not under version control anyway, doesn't it?

SuperSandro2000:
> yeah that'd be nice, some users have `warn-dirty = false` because we don't need these warnings.

I don't consider this nice to have but absolutely required to not spam even more warnings when building.

roberth:
I'd consider dirtiness and having untracked files to be rather different scenarios.
If the user has presumably committed all their files, the outcome is a non-dirty build, as it has a valid commit id. Nonetheless, their evaluation may try to access untracked files.
I think warning about dirty repos might be useful to some, when referencing repositories that aren't the current working directory, although personally I'd like to silence them anyway, as they occur too often and I ignore them anyway.
Attempts to access untracked files on the other hand are indicative of a real problem that can be resolved by adding the file to git or to .gitignore.

I would avoid warning about files that aren't accessed though. Some people keep untracked files around and unnecessary warnings could again lead to warning fatigue, just like we have with the dirty warning.

fricklerhandwerk:

This was discussed briefly: the approach is too eager, we should only warn on access. No concrete decision was made how to proceed. @rti are you still interested in pursuing this?

rti:
@fricklerhandwerk Thanks a lot for the update. Unfortunately I don't have the resources for this any time soon. So I will just close this PR for now.

PR #7248, opened 2022-11-01
nix-shell: Support --out-link

### LINKED ISSUE #2208: nix-shell dependencies can be garbage collected any time now / persistent nix-shell envs

jsoo1:
Closes #2208.

`nix-shell` previously had no way of creating gcroots for shell derivation. This can cause problems if a garbage collection happens concurrently with a `nix-shell` incantation that runs long enough.  `nix develop` solves this, as far as I understand it, but `--out-link` functionality would still be useful in some circumstances.

### DIFF
--- a/doc/manual/src/command-ref/nix-shell.md
+++ b/doc/manual/src/command-ref/nix-shell.md
@@ -13,6 +13,8 @@
   [`--exclude` *regexp*]
   [`--pure`]
   [`--keep` *name*]
+  [`--no-out-link`]
+  [{`--out-link` | `-o`} *outlink*]
   {{`--packages` | `-p`} {*packages* | *expressions*} … | [*path*]}
 
 # Disambiguation
@@ -100,6 +102,16 @@ All options not listed here are passed to `nix-store
   - `--keep` *name*\
     When a `--pure` shell is started, keep the listed environment
     variables.
+
+  - [`--no-out-link`]{#opt-no-out-link}\
+    Do not create a symlink to the output path. This is the default.
+    Note that as a result the output does not become a root of the
+    garbage collector, and so might be deleted by `nix-store --gc`.
+
+  - [`--out-link`]{#opt-out-link} / `-o` *outlink*\
+    Create a symlink to the output path of the shell environment. The
+    output will then become a root of the garbage collector and will
+    not be deleted by `nix-store --gc`.
 
 The following common options are supported:
 
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -5,3 +5,4 @@
   arguments will be ignored and the resulting derivation will have
   `__impure` set to `true`, making it an impure derivation.
 
+* `nix-shell` now accepts `--out-link` and `--no-out-link` flags.
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -366,6 +366,14 @@ static void main_nix_build(int argc, char * * argv)
             store->buildPaths(paths, buildMode, evalStore);
     };
 
+    auto writeOutLink = [&](std::string_view prefix, const StorePath & outputPath, std::string_view outputName) {
+        if (auto store2 = store.dynamic_pointer_cast<LocalFSStore>()) {
+            std::string symlink(prefix);
+            if (outputName != "out") symlink += "-" + outputName;
+            store2->addPermRoot(outputPath, absPath(symlink));
+        }
+    };
+
     if (runEnv) {
         if (drvs.size() != 1)
             throw UsageError("nix-shell requires a single derivation");
@@ -620,11 +628,7 @@ static void main_nix_build(int argc, char * * argv)
             assert(maybeOutputPath);
             auto outputPath = *maybeOutputPath;
 
-            if (auto store2 = store.dynamic_pointer_cast<LocalFSStore>()) {
-                std::string symlink = drvPrefix;
-                if (outputName != "out") symlink += "-" + outputName;
-                store2->addPermRoot(outputPath, absPath(symlink));
-            }
+            writeOutLink(drvPrefix, outputPath, outputName);
 
             outPaths.push_back(outputPath);
         }
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -99,7 +99,7 @@ static void main_nix_build(int argc, char * * argv)
 
     AutoDelete tmpDir(createTempDir("", myName));
 
-    std::string outLink = "./result";
+    std::optional<std::string> outLink = runEnv ? std::nullopt : std::optional{"./result"};
 
     // List of environment variables kept for --pure
     std::set<std::string> keepVars{
@@ -155,7 +155,7 @@ static void main_nix_build(int argc, char * * argv)
             ; // obsolete
 
         else if (*arg == "--no-out-link" || *arg == "--no-link")
-            outLink = (Path) tmpDir + "/result";
+            outLink = std::nullopt;
 
         else if (*arg == "--attr" || *arg == "-A")
             attrPaths.push_back(getArg(*arg, arg, end));
@@ -610,7 +610,7 @@ static void main_nix_build(int argc, char * * argv)
 
         for (auto & [drvPath, outputName] : pathsToBuildOrdered) {
             auto & [counter, _wantedOutputs] = drvMap.at({drvPath});
-            std::string drvPrefix = outLink;
+            std::string drvPrefix = outLink.value_or((Path) tmpDir + "/result");
             if (counter)
                 drvPrefix += fmt("-%d", counter + 1);
 
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -453,10 +453,7 @@ static void main_nix_build(int argc, char * * argv)
 
         if (shellDrv) {
             auto shellDrvOutputs = store->queryPartialDerivationOutputMap(shellDrv.value());
-            auto shellDrvOut = shellDrvOutputs.at("out").value();
-            shell = store->printStorePath(shellDrvOut) + "/bin/bash";
-
-            if (outLink) writeOutLink(*outLink + "-shell", shellDrvOut, "out");
+            shell = store->printStorePath(shellDrvOutputs.at("out").value()) + "/bin/bash";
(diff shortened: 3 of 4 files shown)


jsoo1:
My only concern with this change is the following scenario:

1. `default.nix` builds both the shell and the primary artifact of some project
2. `nix-shell --out-link <some-path>` fails to build because the project is in a broken state

The user could omit `--out-link` in this case. But there is currently no indication why the shell can't be accessed to work on the project in a failiing state.

roberth:
Wouldn't a temporary root solve the problem of retaining a long running shell's dependencies? This would be a cleaner solution where you don't need to put in the file system what seem to be useless links.

roberth:
It seems that `execvp` might stop us from holding on to temporary roots. We might want to run a child process instead.

jsoo1:
> Wouldn't a temporary root solve the problem of retaining a long running shell's dependencies? This would be a cleaner solution where you don't need to put in the file system what seem to be useless links.

I think probably `nix-shell` would ideally always create a temp root, what do you think?

`--out-link` mirrors `nix-build`, though, and was definitely easier to implement :sweat_smile:

roberth:
> I think probably `nix-shell` would ideally always create a temp root, what do you think?

:+1: 

> `--out-link` mirrors `nix-build`, though, and was definitely easier to implement sweat_smile

`nix-shell` is quite different from `nix-build` though. A shell doesn't really produce outputs, so it shouldn't mirror `nix-build` like that.

> > It seems that execvp might stop us from holding on to temporary roots. We might want to run a child process instead.
> 
> Would that be a breaking change to `nix-shell`?

We're definitely past the point where any detail could become a breaking change, but I'd expect even a naive implementation to fix more problems than it creates. One potential problem we can solve is passing posix signals along to the child process. The `siginfo_t` sending pid will still be different, but now I'm talking about a niche in a niche. We could add an `--exec` flag as a contingency.

roberth:
I'm not much of an expert either, so I don't think I'd be super helpful, but I can do a pass of review.

jsoo1:
Wait a second. I think temproots shouldn't even be required.  GC reads /proc/{environ,maps} etc to check for live roots anyways:

https://github.com/NixOS/nix/blob/ffca3e34cb59a5cd18669805d968eaeaa158c0f0/src/libstore/gc.cc#L348

and:

```ShellSession
$ nix-shell -p tmux --run 'nix-store --query --roots $(type -p tmux) | grep $BASHPID'                                                                               
these paths will be fetched (0.34 MiB download, 0.97 MiB unpacked):
  /nix/store/zfwyd14l1z0sg54d6m7jw43469si1pdw-tmux-3.3a
copying path '/nix/store/zfwyd14l1z0sg54d6m7jw43469si1pdw-tmux-3.3a' from 'https://cache.nixos.org'...
/proc/3277340/environ -> /nix/store/zfwyd14l1z0sg54d6m7jw43469si1pdw-tmux-3.3a
```

jsoo1:
@roberth https://github.com/NixOS/nix/pull/7281 for temp roots

jsoo1:
@thufschmitt
> Mh, I couldn't find anything really relevant in https://github.com/NixOS/nix/issues/2208, it seems to really be related to keeping the gc root once the shell is closed (which is also something important that we've left aside for too long).

Do you think `--out-link` is the right flag for this?

thufschmitt:
> @thufschmitt
> 
> > Mh, I couldn't find anything really relevant in #2208, it seems to really be related to keeping the gc root once the shell is closed (which is also something important that we've left aside for too long).
> 
> Do you think `--out-link` is the right flag for this?

Indeed, it is. The current approach isn't exactly right though, because:
1. It requires the derivation to build, which often won't be the case (`mkShell` for example produces a failing derivation by design)
2. It will only keep the runtime references of the output, which is the subset of the dependencies of the shell (in the pathological case where the builder is `touch $out` like you've added it for `-p`, it probably won't keep anything)

What `nix develop` (which supports an `--out-path` flag) and I think [lorri](https://github.com/nix-community/lorri) do is that they build a modified version of the derivation which retains all the build-time inputs (something along the lines of `env > $out` might be a good approximation of that).

bjornfor:
> It requires the derivation to build, which often won't be the case (mkShell for example produces a failing derivation by design)

`mkShell` was thankfully fixed (made buildable) in https://github.com/nixos/nixpkgs/commit/1e910209ae414755415ed8ef990fa3cbb07c8cd9.

jsoo1:
> > Do you think `--out-link` is the right flag for this?
> 
> Indeed, it is. The current approach isn't exactly right though, because:
> 
>     1. It requires the derivation to build, which often won't be the case (`mkShell` for example produces a failing derivation by design)

I mentioned this [above](https://github.com/NixOS/nix/pull/7248#issuecomment-1307429988) as a UX problem . I am not sure this is entirely a problem. For back-compat I could see there being problem cases. But in the back-compat case `--no-out-link` is still the default.

Two additional things about this:
1. I'm not sure that's true (see https://github.com/NixOS/nixpkgs/blob/0106a6855599fbe26a7c331782d9672f829bd1e1/pkgs/build-support/mkshell/default.nix#L46).
2. As mentioned in mkshell/default.nix, the (non)existence of an output is a nixpkgs implementation detail. I would say nix probably should be less coupled to nixpkgs if at all possible. (I know there is a lot of nixpkgs-specific code in nix already, but I worry those cause other, more problematic back-compat problems).

>     2. It will only keep the runtime references of the output, which is the subset of the dependencies of the shell (in the pathological case where the builder is `touch $out` like you've added it for `-p`, it probably won't keep anything)
> 
> 
> What `nix develop` (which supports an `--out-path` flag) and I think [lorri](https://github.com/nix-community/lorri) do is that they build a modified version of the derivation which retains all the build-time inputs (something along the lines of `env > $out` might be a good approximation of that).

I think this is generally a very hard problem to solve with the tools nix has and with the existence of `nuke-references` and friends. I would like there to be a solution that would also capture build-time inputs. Does it need to go in as part of this set of patches, though?

thufschmitt:
> 1. I'm not sure that's true (see https://github.com/NixOS/nixpkgs/blob/0106a6855599fbe26a7c331782d9672f829bd1e1/pkgs/build-support/mkshell/default.nix#L46).

Indeed. Looks like I'm nearly [one year late to the party](https://github.com/NixOS/nixpkgs/pull/153194) :stuck_out_tongue: 
 
> 2. As mentioned in mkshell/default.nix, the (non)existence of an output is a nixpkgs implementation detail. I would say nix probably should be less coupled to nixpkgs if at all possible. (I know there is a lot of nixpkgs-specific code in nix already, but I worry those cause other, more problematic back-compat problems).

Do you mean that there's no guaranty that `out` will exist? That's true, but we'd want to set `outputs = [ "out" ]` anyway because we want to fix the outputs list.

> > 2. It will only keep the runtime references of the output, which is the subset of the dependencies of the shell (in the pathological case where the builder is `touch $out` like you've added it for `-p`, it probably won't keep anything)
> > 
> > What `nix develop` (which supports an `--out-path` flag) and I think [lorri](https://github.com/nix-community/lorri) do is that they build a modified version of the derivation which retains all the build-time inputs (something along the lines of `env > $out` might be a good approximation of that).
> 
> I think this is generally a very hard problem to solve with the tools nix has and with the existence of `nuke-references` and friends. I would like there to be a solution that would also capture build-time inputs. Does it need to go in as part of this set of patches, though?

I don't think it is. Things like `nuke-references` are external programs that can't interfere here. The only potential limitation would be #7087, but even then we can just make sure that we unset `__visibleReferences`.

> Edit: Another thought: this `--out-link` works the same way that `--out-link` from `nix-build` does. I think including build-time inputs might violate the principle of least surprise. If build-time inputs were also kept as gc roots for every other derivation (say by `nix-build` or `nix-store --realize --add-root ...`) then I would be for that change here.

Quite the contrary, I would say. `nix-shell` and `nix-build` are conceptually very different (despite sharing the same code), and having `nix-shell`'s `--out-path` flag mirror `nix-build`'s wouldn't make that much sense imho.

At any rate, this new flag is just combining `nix-build --out-path foo && nix-shell` into `nix-shell --out-path foo`, which I think is not worth it given the confusion about the semantics of `--out-path`.

roberth:
Adding `--out-link` to `nix-shell` is like giving Frankenstein's monster a tail fin. I really appreciate the effort to try to help the monster, but there's no fixing it.
`nix-shell` already has conflicting requirements, and adding `--out-link` will only make that worse. The new flag doesn't even really solve the root cause of @jsoo1's problem.
What we should do instead is

1. Fix bugs instead of adding workarounds

2. Deprecate nix-shell in favor of specific commands, some of which are `nix run`, and some others which require new nix-daemon features. Here's an [analysis](https://github.com/NixOS/nix/issues/4715#issuecomment-917069357) (probably should be its own issue) of the problem space that nix-shell necessarily fails to cover.

jsoo1:
I think this change is still worthwhile since #7281 won't merge, but I don't have time to do the work and continue this conversation. Thank you for your consideration @roberth !

PR #7251, opened 2022-11-01
docs: drop shell prompt character for consistency

msagarpatel:
In addition to consistency, the fancy "Copy to clipboard" button on the website will copy the prompt character. Retaining the prompt character would mean having to edit each command after pasting in the terminal.

### DIFF
--- a/doc/manual/src/installation/installing-binary.md
+++ b/doc/manual/src/installation/installing-binary.md
@@ -3,7 +3,7 @@
 The easiest way to install Nix is to run the following command:
 
 ```console
-$ sh <(curl -L https://nixos.org/nix/install)
+sh <(curl -L https://nixos.org/nix/install)
 ```
 
 This will run the installer interactively (causing it to explain what
@@ -27,7 +27,7 @@ you can authenticate with `sudo`.
 To explicitly select a single-user installation on your system:
 
 ```console
-$ sh <(curl -L https://nixos.org/nix/install) --no-daemon
+sh <(curl -L https://nixos.org/nix/install) --no-daemon
 ```
 
 This will perform a single-user installation of Nix, meaning that `/nix`
@@ -37,8 +37,8 @@ if it doesn’t already exist. If you don’t have `sudo`, you should
 manually create `/nix` first as root, e.g.:
 
 ```console
-$ mkdir /nix
-$ chown alice /nix
+mkdir /nix
+chown alice /nix
 ```
 
 The install script will modify the first writable file from amongst
@@ -50,7 +50,7 @@ the install script to disable this behaviour.
 You can uninstall Nix simply by running:
 
 ```console
-$ rm -rf /nix
+rm -rf /nix
 ```
 
 # Multi User Installation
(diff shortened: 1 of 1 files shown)


edolstra:
This makes it hard to distinguish between commands and their output in examples. In these particular cases, there is no output, but for consistency we should include a prompt character everywhere.

fricklerhandwerk:
I agree with @msagarpatel. See [this comment on a similar issue](https://github.com/NixOS/nixos-summer/pull/41#discussion_r953222745) for details.

Distinguishing input and output can and should be done in CSS. Since there are apparently no examples with outputs, we're not in a hurry there.

fricklerhandwerk (approved):

Et7f3:
> This makes it hard to distinguish between commands and their output in examples. In these particular cases, there is no output, but for consistency we should include a prompt character everywhere.

What about `:;` for prompt ?

fricklerhandwerk:
Related: https://github.com/NixOS/nix.dev/pull/390#issuecomment-1311969633

Does the codeblock element behave the same way in mdBook?

Ericson2314:
@fricklerhandwerk mentioned that Sphinx can rendering this very nicely. I just learned sphinx supports markdown, and I have long thought Sphinx does a better job than mdbook. Perhaps issues like this are best addressed with such a switch.

fricklerhandwerk:

   - @fricklerhandwerk this could be something for the documentation team to decide on autonomously
       - @thufschmitt, @edolstra: agreed

fricklerhandwerk:
Discussed on the @NixOS/documentation-team. Will merge this as-is, and follow-up with an issue to change the presentation according to https://github.com/NixOS/nix.dev/pull/390#issuecomment-1311969633

@Ericson2314 opening the platform discussion again is not worthwhile any time soon, in my opinion. We should instead fix the various issues with structure and contents.

Ericson2314:
@fricklerhandwerk I am confused.

1. This makes the docs *less* consistent. If ones does `git grep -A3 '```console` it is clear there are many other examples of the `$ ` being used.

2. The `$` is good to clarify the contents. It is a bummer mdbook does not handle it well, but it reflects our semantic intent.

3. The I bring up Sphinx thing because it feels backwards to me to make our markdown worse because the tool we used to render it is bad. It seems like firstly, we should write the markdown we want to write, and secondly, if the rendering is bad we should either improve the tool or switch tools.

   There was lots of drama/bikeshedding switching from docbook to markdown but very little discussion of the tool. If we use sphinx for this and nix.dev we are move consistent / fewer tools, so that is good.
   
   But regardless, I think we should keep `$` even if it sucks to cut and paste, because consistency and expressive power.

So I really don't understand the decision or the PR title.

Ericson2314:
At the very least, if we don't include the prompt I think `bash` is more appropriate than `console`. Tracking the convention with the syntax highlighting tag makes it easier to update later.

infinisil:
The docs will be read and used by thousands of people, but only written or updated by very few. Because of this I think it makes sense to make a small sacrifice (it's really small, a single character) for doc writers at the benefit of all users.

But yes the inconsistency isn't great, we should use the same decision everywhere.

PR #7252, opened 2022-11-01
tarfile: set directory mode to at least 0500, don't extract fflags and perms

### LINKED ISSUE #7175: `builtins.fetchTree` and `builtins.fetchTarball` crash for tarballs with bogus `u-x` directory perms.
**Describe the bug**

The builtin fetchers crash and burn for tarballs that carry evil permission bits on directories. I'm sure this is a side effect of running `tar c` on a wonky system or something, but if a tarball has a directory with perms like this you get a permissions failure during the post-unpack check routine in `libutil/tarfile.cc`.
```
~/repos/nix % tar tfv ./tarball.tgz
drw-rw-rw- 0/0               0 2020-02-18 03:50 package
-rw-rw-rw- 0/0             297 2020-02-18 03:50 package/index.d.ts
-rw-rw-rw- 0/0            1920 2020-02-18 03:50 package/index.js
-rw-rw-rw- 0/0            1092 2020-01-31 04:31 package/LICENSE
-rw-rw-rw- 0/0             937 2020-02-18 03:51 package/package.json
-rw-rw-rw- 0/0             713 2020-02-18 03:50 package/README.md
```

People shouldn't be publishing releases with permissions like this; but unfortunately it effects ~1/3,000 tarballs in the NPM registry. This really threw a wrench in a big Node.js+Flakes effort I've been working on.

I'm not sure if the `umask` settings of the user or daemon or builder are the ones being applied, but I suspect that it's using the `nix-daemon` perms as `root` because  otherwise `libarchive` is supposed to respect `umask`. Notice in the perms above the group and user IDs are set to `0/0` which I expect is relevant.

**Steps To Reproduce**

```
nix eval --impure --expr 'builtins.fetchTarball { url = "https://registry.npmjs.org/char-regex/-/char-regex-1.0.2.tgz"; }'
error (ignored): error: end of string reached
error: getting status of '/private/tmp/nix-42779-0/package/LICENSE': Permission denied
# or
nix eval --impure --expr 'builtins.fetchTree { type = "tarball"; url = "https://registry.npmjs.org/char-regex/-/char-regex-1.0.2.tgz"; }'
error (ignored): error: end of string reached
error: getting status of '/private/tmp/nix-42786-0/package/LICENSE': Permission denied
```

**Expected behavior**

untarred dirs should be readable by the user who invoked `nix`.

nix-env (Nix) 2.10.3

**Additional context**

I'm running on Darwin, but IIRC this same issue effected my Linux boxes. They're powered off now so I can confirm that later.

I tried a bunch of different `ARCHIVE_EXTRACT_*` combinations with no success but I'm far from a `libarchive` expert. A simple fix might just be to reset `u+x` before attempting to read.

I'm happy to help resolve this but I'm hesitant to PR anything effecting permissions without a more expert set of eyes on it.

yorickvP:
We don't need SGID, or any ACL's. We also want to keep every dir +rx.

Fixes #7175

### DIFF
--- a/src/libutil/tarfile.cc
+++ b/src/libutil/tarfile.cc
@@ -77,9 +77,7 @@ TarArchive::~TarArchive()
 
 static void extract_archive(TarArchive & archive, const Path & destDir)
 {
-    int flags = ARCHIVE_EXTRACT_FFLAGS
-        | ARCHIVE_EXTRACT_PERM
-        | ARCHIVE_EXTRACT_TIME
+    int flags = ARCHIVE_EXTRACT_TIME
         | ARCHIVE_EXTRACT_SECURE_SYMLINKS
         | ARCHIVE_EXTRACT_SECURE_NODOTDOT;
 
@@ -97,6 +95,10 @@ static void extract_archive(TarArchive & archive, const Path & destDir)
 
         archive_entry_copy_pathname(entry,
             (destDir + "/" + name).c_str());
+
+        // sources can and do contain dirs with no rx bits
+        if (archive_entry_filetype(entry) == AE_IFDIR && (archive_entry_mode(entry) & 0500) != 0500)
+            archive_entry_set_mode(entry, archive_entry_mode(entry) | 0500);
 
         // Patch hardlink path
         const char *original_hardlink = archive_entry_hardlink(entry);


edolstra:
Does this preserve the executable bit on regular files?

aakropotkin:
Thank you so much

yorickvP:
> Does this preserve the executable bit on regular files?

Yes

Issue #6647, opened 2022-06-10
error: the string '...' is not allowed to refer to a store path


yu-re-ka:
**Describe the bug**

I'm getting an eval error for an expression that used to evaluate and I don't exactly know why

**Steps To Reproduce**

```
$ git clone https://git.neulandlabor.de/Yuka/c3h-nixfiles/
$ cd c3h-nixfiles
$ git checkout f8686374176d7875846a0ee71a2c2890ef7d4a85
$ nix-instantiate "<nixpkgs/nixos>" -A config.system.build.toplevel -I "$(nix-build nix/sources-dir.nix --no-out-link)" -I "nixos-config=$PWD/hosts/nuc/configuration.nix"
error: the string 'nixos-system-nuc-22.05pre-git' is not allowed to refer to a store path (such as '')
```

**Expected behavior**

This should evaluate, or throw a more helpful error. It did evaluate with 2.3, and users expect their old Nix expressions that used only stable features to evaluate with new versions of Nix.

**`nix-env --version` output**

nix-env (Nix) 2.9.0

**Additional context**

The same expression was already failing to build in the initial 2.4 release, which got fixed in 2.5: #5410
Overall this is the behaviour with different Nix versions:

nix 2.3: ✅ 
nix 2.4: `error: store path '/nix/store/dbfkhagz0s9246jgahm913vypi0sv6qf-sources' is not allowed to have references`
nix 2.5: ✅ 
nix 2.6-2.8: `error: the string 'nixos-system-nuc-22.05pre-git' is not allowed to refer to a store path (such as '/nix/store/8x56z90r0h9iwgba93hy2iws4z9rkrgl-nixpkgs-src')`
nix 2.9: `error: the string 'nixos-system-nuc-22.05pre-git' is not allowed to refer to a store path (such as '')`

I tried to come up with a simpler example to reproduce it, but I don't understand the underlying issue enough. It would be great if this could be added as a test so that future Nix versions don't break the behavior again.

thufschmitt:
I'm not sure what is the right behavior here, but at a glance the issue comes from the fact that your `nixpkgs` is itself a derivation output (the output of the `runCommand` call in `nix/sources-dir.nix`), and that's causing some troubles.

yorickvP:
Small testcase:
```nix
let
  nixpkgs = fetchTarball {
    url = channel:nixos-22.05;
  };

  pkgs = import nixpkgs { overlays = []; };

[...]

PR #7258, opened 2022-11-04
Fix printing of eval errors with two format placeholders

ncfavier:
With this fix, the error in https://github.com/NixOS/nix/issues/6647#issue-1267677835 becomes slightly more useful:

```
error: the string 'nixos-system-nuc-22.05pre-git' is not allowed to refer to a store path (such as '/nix/store/8x56z90r0h9iwgba93hy2iws4z9rkrgl-nixpkgs-src')
```

It would be nice to enable warnings for unused ~variables~ parameters but it looks like a lot of work to silence them all.


ncfavier:
ping @thufschmitt

thufschmitt (approved):
Thanks :)

PR #7260, opened 2022-11-04
Restrict `readFile` context to references that appear in the string

ncfavier:
When calling `builtins.readFile` on a store path, the references of that path are currently added to the resulting string's context (https://github.com/NixOS/nix/pull/1643).

This change makes those references the *possible* context of the string, but filters them to keep only the references whose hash actually appears in the string, similarly to what is done for determining the runtime references of a path.

Fixes https://github.com/NixOS/nix/issues/6647 and probably https://github.com/NixOS/nixpkgs/issues/175125 (not sure how to test that one) while keeping the `readfile-context` test successful.

~I wasn't sure how to generalise `dumpPath(path, sink)` and `sink << string` so I gave in to my primal Haskeller instincts and used a lambda. We could make `scanForReferences` accept a source, but then we'd need to turn `λ sink. dumpPath(path, sink)` into a source and the simplest way to do this [seems](https://github.com/NixOS/nix/blob/499e99d099ec513478a2d3120b2af3a16d9ae49d/src/libutil/serialise.cc#L245-L292) to involves coroutines...~


Ericson2314:
This is a nice idea, thanks for looking into this!

src/libstore/references.cc
```diff
@@ -81,6 +81,18 @@ StorePathSet scanForReferences(
...
+    auto fun = [&](Sink & refsSink) {
+        /* Look for the hashes in the NAR dump of the path. */
+        TeeSink sink { refsSink, toTee };
+        dumpPath(path, sink);
+    };
+    return scanForReferences(fun, refs);
+}
+
+StorePathSet scanForReferences(
+    std::function<void(Sink &)> fun,
```
Ericson2314, line 94:
```suggestion
    Source & source,
```
I feel like this should make it less awkward? Note that there are various source to sink and sink to source functions.

ncfavier: Yes, I noticed `sinkToSource` (see OP), but I feel like it's a shame to use coroutines for something so simple...

Ericson2314: Actually, what we really want is a `PathRefScanSink`

ncfavier: That probably makes more sense

ncfavier: Ugh, this is apparently too much for C++. Tried `PathRefScanSink : public RefScanSink`, but it doesn't seem possible to do the preprocessing (for each path, add its hash to `hashes` and insert it into the `backMap`) *before* calling the parent constructor. I'd like to avoid having an inner `RefScanSink * sink;` member and wiring it up explicitly...

I guess I'll have to make a "smart constructor" function that does the preprocessing and calls the constructor.

Ericson2314: @ncfavier Here's a trick, you can make static method (or just free-floating function) which the constructor calls!

ncfavier: But this needs to happen in the member initialiser list, so we can't feed one output of the function (hashes) to the base constructor and the other output (backMap) to initialise a member...

Ericson2314: @ncfavier I was thinking about this, see https://stackoverflow.com/questions/4162021/is-it-ok-to-call-a-function-in-constructor-initializer-list I think we can make a blank back map, then make the other one set by mutating the first one.

Ericson2314: @ncfavier I replied on matrix with details.

ncfavier: Made the constructor private and inlined `scanStringForReferences` as requested

Ericson2314: Thanks!

Ericson2314 (approved):
This change is looking quite good. It is the best of both worlds between not just dropping the references but also having less of a breaking change than 2.3.

Thank you very much @ncfavier!

src/libstore/references.hh
```diff
@@ -27,6 +27,22 @@ public:
...
 };
 
+class PathRefScanSink : public RefScanSink
+{
+    std::map<std::string, StorePath> backMap;
+
+    PathRefScanSink(StringSet && hashes, std::map<std::string, StorePath> backMap)
+        : RefScanSink(std::move(hashes))
+        , backMap(backMap)
+    { }
```
Ericson2314, line 37:
```suggestion
    PathRefScanSink(StringSet && hashes, std::map<std::string, StorePath> && backMap)
        : RefScanSink(std::move(hashes))
        , backMap(std::move(backMap))
    { }
```

Ericson2314: I would also move it out of the header into the implementation file with
```suggestion
    PathRefScanSink(StringSet && hashes, std::map<std::string, StorePath> backMap);
```
in just the header,

That leaks the least amount of what is basically implementation details given the "smart constructor".

src/libstore/references.cc
```diff
@@ -67,20 +67,7 @@ void RefScanSink::operator () (std::string_view data)
...
-    auto found = scanForReferences(hashSink, path, refs);
-    auto hash = hashSink.finish();
-    return std::pair<StorePathSet, HashResult>(found, hash);
-}
-
-StorePathSet scanForReferences(
-    Sink & toTee,
-    const Path & path,
-    const StorePathSet & refs)
+PathRefScanSink PathRefScanSink::fromPaths(const StorePathSet & refs)
```
Ericson2314, line 70: if you do opt to merely declare in the header as I suggested
```suggestion
PathRefScanSink::PathRefScanSink(StringSet && hashes, std::map<std::string, StorePath> && backMap)
    : RefScanSink(std::move(hashes))
    , backMap(std::move(backMap))
{ }
    
PathRefScanSink PathRefScanSink::fromPaths(const StorePathSet & refs)
```

src/libexpr/primops.cc
```diff
@@ -1542,6 +1543,9 @@ static void prim_readFile(EvalState & state, const PosIdx pos, Value * * args, V
             refs = state.store->queryPathInfo(state.store->toStorePath(path).first)->references;
         } catch (Error &) { // FIXME: should be InvalidPathError
         }
+        PathRefScanSink refsSink = PathRefScanSink::fromPaths(refs);
```
Ericson2314, line 1546:
```suggestion
        // Re-scan refs to filter down to just the ones that actually occur in the file.
        PathRefScanSink refsSink = PathRefScanSink::fromPaths(refs);
```

Ericson2314 (approved):

src/libexpr/primops.cc
```diff
@@ -1542,6 +1543,10 @@ static void prim_readFile(EvalState & state, const PosIdx pos, Value * * args, V
             refs = state.store->queryPathInfo(state.store->toStorePath(path).first)->references;
         } catch (Error &) { // FIXME: should be InvalidPathError
         }
+        // Re-scan references to filter down to just the ones that actually occur in the file.
+        PathRefScanSink refsSink = PathRefScanSink::fromPaths(refs);
```
edolstra, line 1547:
```suggestion
        auto refsSink = PathRefScanSink::fromPaths(refs);
```

thufschmitt (approved):
Thanks!

@ncfavier I've pushed some release notes and a test. I'll let you confirm that they look all right

tests/readfile-context.nix
```diff
@@ -10,8 +10,17 @@ let
...
+    builder = ./readfile-context.builder.sh;
+    input = "read-dependent"; # No reference in it
+  };
+
   readDependent = mkDerivation {
-    name = "read-dependent";
+    # Will evaluate correctly because `readFile
+    # discardableDependent` won't keep any context (since it will
+    # refscan its content, and this one doesn't have any reference)
+    name = builtins.readFile discardableDependent;
```
ncfavier, line 23: Apparently this succeeds even if there's a context.

thufschmitt: How did you see that? I tried replacing `discardableDependent` by `dependent` and the evaluation fails because `the string '/tmp/nix-test/tests/readfile-context/store/dk4iy6vpksymyjpm1wl0646hh5p3vm3c-simple/hello' is not allowed to refer to a store path `.

I also tried with `builtins.substring 0 0 (builtins.readFile dependent) + "foo"` just to make sure that I had the same error with a valid `name`, and I indeed have it

ncfavier: Ah, I see, the problem is that `discardableDependent` has no references at all so the test succeeds even without this patch.

What we should do to test this properly is have a derivation that writes *two* files, one with a reference and one without, and read the one without a reference.

thufschmitt: Oh yes, of course! I assumed that `discardableDepend` would refer to itself, but it isn't by default

ncfavier: Updated the test.

thufschmitt (approved):

thufschmitt:
Just waiting for the CI to get green, and I'll merge

ncfavier:
It's green now!

thufschmitt:
> It's green now!

Yeah sorry, I kinda new when writing my message that I would forget about it :unamused:

PR #7277, opened 2022-11-08
add CODEOWNERS

fricklerhandwerk:
as a supplement to https://github.com/NixOS/nix/pull/7229 as [discussed on the Nix team](https://github.com/NixOS/nix/pull/7229#pullrequestreview-1168504260)

@edolstra @thufschmitt feel free to add more specifics. Also, who else has merge access and may want to get notified?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,7 +4,7 @@
 # Merge permissions are required for maintaining an entry in this file.
 # For documentation on this mechanism, see https://help.github.com/articles/about-codeowners/
 
-# Notified if nothing else matches
+# Default reviewers if nothing else matches
 * @edolstra @thufschmitt
 
 # This file
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,15 @@
+# Pull requests concerning the listed files will automatically invite the respective maintainers as reviewers.
+# This file is not used for denoting any kind of ownership, but is merely a tool for handling notifications.
+#
+# Merge permissions are required for maintaining an entry in this file.
+# For documentation on this mechanism, see https://help.github.com/articles/about-codeowners/
+
+# Notified if nothing else matches
+* @edolstra @thufschmitt
+
+# This file
+.github/CODEOWNERS @edolstra
+
+# Public documentation
+/doc @fricklerhandwerk
+*.md @fricklerhandwerk


thufschmitt (approved):
👍 from me (although I wouldn't merge it without an agreement from the whole Nix team).

In a "small steps" fashion, I'll refrain from making any addition to the file itself, leaving that for follow-up PRs

.github/CODEOWNERS
```diff
@@ -0,0 +1,15 @@
+# Pull requests concerning the listed files will automatically invite the respective maintainers as reviewers.
+# This file is not used for denoting any kind of ownership, but is merely a tool for handling notifications.
+#
+# Merge permissions are required for maintaining an entry in this file.
+# For documentation on this mechanism, see https://help.github.com/articles/about-codeowners/
+
+# Notified if nothing else matches
```
fricklerhandwerk, line 7:
```suggestion
# Default reviewers nothing else matches
```

fricklerhandwerk:
Discussed on Nix team meeting 2022-11-11:

- @edolstra: this seems to do a subset of the auto-assign bot's work. let's make sure that works first. no point having two different files denoting responsibilities
    - @thufschmitt: if the bot requires reviewers to be collaborators too, let's stick to CODEOWNERS
    - @edolstra: will [try the bot](https://github.com/NixOS/nix/pull/7293) with someone who has no commit access

fricklerhandwerk:
@edolstra does the review assignment workflow work already? The [test PR](https://github.com/NixOS/nix/pull/7293) still fails, but I [got assigned](https://github.com/NixOS/nix/pull/3600#event-7877771808) to #3600 by _some_ GitHub Action.

edolstra:
It seems to work sometimes. On the test PR it fails with "Resource not accessible by integration" whatever that means.

thufschmitt:
I'd suspect it works for PRs that come from the `NixOS/nix` repo and fail for the others

fricklerhandwerk:
I that case it's useless, since everyone except maintainers will make PRs from their forks.

thufschmitt:
> I that case it's useless, since everyone except maintainers will make PRs from their forks.

Yes, totally agreed

> PS: Actually I think everyone _including_ maintainers should do this, to keep only branches which serve a clear purpose so they are discoverable.

I don't think that's too problematic. I agree that using the fork as the base branch is a good default, but it's occasionally useful to create PRs directly from the NixOS repo. For stacked PRs in particular, there's still no good way to do them if all the branches aren't local.

PR #7278, opened 2022-11-08
antiquotation -> string interpolation

fricklerhandwerk:
as proposed by @mkaito[1] and @tazjin[2]

[1]: https://github.com/NixOS/nix.dev/pull/267#issuecomment-1270076332
[2]: https://github.com/NixOS/nix.dev/pull/267#issuecomment-1270201979

### DIFF
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -60,7 +60,7 @@
     cache](https://cache.nixos.org).
 
   - [store path]{#gloss-store-path}\
-    The location in the file system of a store object, i.e., an
+    The location of a [store object] in the file system, i.e., an
     immediate child of the Nix store directory.
 
     Example: `/nix/store/a040m110amc4h71lds2jmr8qrkj2jhxd-git-2.38.1`
@@ -178,3 +178,12 @@
 
   - [`ε`]{#gloss-epsilon}\
     The epsilon symbol. In the context of a package, this means the version is empty. More precisely, the derivation does not have a version attribute.
+
+  - [string interpolation]{#gloss-string-interpolation}\
+    Expanding expressions enclosed in `${ }` within a [string], [path], or [attribute name].
+
+    See [String interpolation](./language/string-interpolation.md) for details.
+
+    [string]: ./language/values.md#type-string
+    [path]: ./language/values.md#type-path
+    [attribute name]: ./language/values.md#attribute-set
--- a/doc/manual/src/language/string-interpolation.md
+++ b/doc/manual/src/language/string-interpolation.md
@@ -1,0 +1,82 @@
+# String interpolation
+
+String interpolation is a language feature where a [string], [path], or [attribute name] can contain expressions enclosed in `${ }` (dollar-sign with curly brackets).
+
+Such a string is an *interpolated string*, and an expression inside is an *interpolated expression*.
+
+Interpolated expressions must evaluate to one of the following:
+
+- a [string]
+- a [path]
+- a [derivation]
+
+[string]: ./values.md#type-string
+[path]: ./values.md#type-path
+[attribute name]: ./values.md#attribute-set
+[derivation]: ../glossary.md#gloss-derivation
+
+## Examples
+
+### String
+
+Rather than writing
+
+```nix
+"--with-freetype2-library=" + freetype + "/lib"
+```
+
+(where `freetype` is a [derivation]), you can instead write
+
+```nix
+"--with-freetype2-library=${freetype}/lib"
+```
+
+The latter is automatically translated to the former.
+
+A more complicated example (from the Nix expression for [Qt](http://www.trolltech.com/products/qt)):
+
+```nix
+configureFlags = "
+  -system-zlib -system-libpng -system-libjpeg
+  ${if openglSupport then "-dlopen-opengl
+    -L${mesa}/lib -I${mesa}/include
+    -L${libXmu}/lib -I${libXmu}/include" else ""}
+  ${if threadSupport then "-thread" else "-no-thread"}
+";
+```
+
+Note that Nix expressions and strings can be arbitrarily nested;
+in this case the outer string contains various interpolated expressions that themselves contain strings (e.g., `"-thread"`), some of which in turn contain interpolated expressions (e.g., `${mesa}`).
+
+### Path
+
+Rather than writing
+
+```nix
+./. + "/" + foo + "-" + bar + ".nix"
+```
+
+or
+
+```nix
+./. + "/${foo}-${bar}.nix"
+```
+
+you can instead write
+
+```nix
+./${foo}-${bar}.nix
+```
+
+### Attribute name
+
+Attribute names can be created dynamically with string interpolation:
+
+```nix
+let name = "foo"; in
+{
+  ${name} = "bar";
+}
+```
+
+    { foo = "bar"; }
--- a/doc/manual/src/language/values.md
+++ b/doc/manual/src/language/values.md
@@ -13,41 +13,9 @@
   returns and tabs can be written as `\n`, `\r` and `\t`,
   respectively.
 
-  You can include the result of an expression into a string by
-  enclosing it in `${...}`, a feature known as *antiquotation*. The
-  enclosed expression must evaluate to something that can be coerced
-  into a string (meaning that it must be a string, a path, or a
-  derivation). For instance, rather than writing
+  You can include the results of other expressions into a string by enclosing them in `${ }`, a feature known as [string interpolation].
 
-  ```nix
-  "--with-freetype2-library=" + freetype + "/lib"
-  ```
-
-  (where `freetype` is a derivation), you can instead write the more
-  natural
-
-  ```nix
-  "--with-freetype2-library=${freetype}/lib"
-  ```
-
-  The latter is automatically translated to the former. A more
-  complicated example (from the Nix expression for
-  [Qt](http://www.trolltech.com/products/qt)):
-
-  ```nix
-  configureFlags = "
-    -system-zlib -system-libpng -system-libjpeg
-    ${if openglSupport then "-dlopen-opengl
-      -L${mesa}/lib -I${mesa}/include
-      -L${libXmu}/lib -I${libXmu}/include" else ""}
-    ${if threadSupport then "-thread" else "-no-thread"}
-  ";
-  ```
-
-  Note that Nix expressions and strings can be arbitrarily nested; in
-  this case the outer string contains various antiquotations that
-  themselves contain strings (e.g., `"-thread"`), some of which in
-  turn contain expressions (e.g., `${mesa}`).
+  [string interpolation]: ./string-interpolation.md
 
   The second way to write string literals is as an *indented string*,
   which is enclosed between pairs of *double single-quotes*, like so:
@@ -75,7 +43,7 @@
   Note that the whitespace and newline following the opening `''` is
   ignored if there is no non-whitespace text on the initial line.
 
-  Antiquotation (`${expr}`) is supported in indented strings.
+  Indented strings support [string interpolation].
 
   Since `${` and `''` have special meaning in indented strings, you
   need a way to quote them. `$` can be escaped by prefixing it with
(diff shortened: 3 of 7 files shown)


doc/manual/src/language/values.md
```diff
@@ -14,8 +14,11 @@
   respectively.
 
   You can include the result of an expression into a string by
-  enclosing it in `${...}`, a feature known as *antiquotation*. The
-  enclosed expression must evaluate to something that can be coerced
+  enclosing it in `${...}`, a feature known as *string interpolation*.
+  Such a string with placeholders is an *interpolated string*, and a
```
edolstra, line 18: I think we shouldn't use "placeholders", because that sounds like Python/Rust/SQL placeholders (e.g. `{}`` or `format!("foo {}", x)`). Nix doesn't have placeholders in that sense.

doc/manual/src/language/values.md
```diff
@@ -238,20 +241,20 @@ You can use arbitrary double-quoted strings as attribute names:
 { "foo ${bar}" = 123; "nix-1.0" = 456; }."foo ${bar}"
 ```
 
-This will evaluate to `123` (Assuming `bar` is antiquotable). In the
-case where an attribute name is just a single antiquotation, the quotes
-can be dropped:
+This will evaluate to `123` (Assuming `bar` can be coerced to a string). In the
```
edolstra, line 244:
```suggestion
This will evaluate to `123` (assuming `bar` can be coerced to a string). In the
```

doc/manual/src/language/values.md
fricklerhandwerk:
```suggestion
  Such a string is an *interpolated string*, and a
```

doc/manual/src/language/values.md
```diff
@@ -14,8 +14,11 @@
   respectively.
 
   You can include the result of an expression into a string by
-  enclosing it in `${...}`, a feature known as *antiquotation*. The
-  enclosed expression must evaluate to something that can be coerced
+  enclosing it in `${...}`, a feature known as *string interpolation*.
+  Such a string with placeholders is an *interpolated string*, and a
+  a placeholder is an *interpolation part*.
```
fricklerhandwerk, line 19: @Ericson2314 suggests
```suggestion
  a placeholder is an *interpolation point*.
```

doc/manual/src/glossary.md
```diff
@@ -150,7 +152,23 @@
...
     links.  NARs are generated and unpacked using `nix-store --dump`
     and `nix-store --restore`.
+
   - [`∅`]{#gloss-emtpy-set}\
     The empty set symbol. In the context of profile history, this denotes a package is not present in a particular version of the profile.
+
   - [`ε`]{#gloss-epsilon}\
     The epsilon symbol. In the context of a package, this means the version is empty. More precisely, the derivation does not have a version attribute.
+
+  - [string interpolation]{#gloss-string-interpolation}\
```
edolstra, line 162: I don't think this should be in the glossary. It should be a section in the language chapter. The glossary should just have brief definitions that link elsewhere for full details.

fricklerhandwerk: Done

doc/manual/src/glossary.md
```diff
@@ -46,7 +48,7 @@
     cache](https://cache.nixos.org).
 
   - [store path]{#gloss-store-path}\
-    The location in the file system of a store object, i.e., an
+    The of a [store object] location in the file system, i.e., an
```
Ericson2314, line 51:
```suggestion
    The location of a [store object] in the file system, i.e., an
```

Ericson2314 (approved):
Except for the typo I think found, LGTM

doc/manual/src/language/string-interpolation.md
```diff
@@ -0,0 +1,98 @@
...
+- [path]
+- [derivation]
+
+[string]: ./language/values.md#type-string
+[path]: ./language/values.md#type-path
+[attribute name]: ./language/values.md#attribute-set
+[derivation]: ../glossary.md#gloss-derivation
+
+When an interpolated expression evaluates to a path, and is thus coerced into a string, the path is first copied into the Nix store and the resulting string is
+the [store path].
```
fricklerhandwerk, line 19:
```suggestion
When an interpolated expression evaluates to a path, and is thus coerced into a string, the path is first copied into the Nix store and the resulting string is the [store path].
```

doc/manual/src/language/string-interpolation.md
```diff
@@ -0,0 +1,97 @@
+# String interpolation
+
+String interpolation is a language feature where a [string], [path], or [attribute name] can contain expressions enclosed in `${ }` (dollar-sign with curly brackets).
+
+Such a string is an *interpolated string*, and an expression inside is an *interpolated expression*.
+
+Interpolated expressions must evaluate to something that can be coerced to a string:
```
tazjin, line 7: I'd add `, such as:` here at the end because the list is not exhaustive.

doc/manual/src/language/string-interpolation.md
```diff
@@ -0,0 +1,97 @@
...
+- [string]
+- [path]
+- [derivation]
+
+[string]: ./language/values.md#type-string
+[path]: ./language/values.md#type-path
+[attribute name]: ./language/values.md#attribute-set
+[derivation]: ../glossary.md#gloss-derivation
+
+When an interpolated expression evaluates to a path, and is thus coerced into a string, the path is first copied into the Nix store and the resulting string is the [store path].
```
tazjin, line 18: "thus" is causal, but I think you want temporal relation here: "and is *then* coerced"

doc/manual/src/language/string-interpolation.md
```diff
@@ -0,0 +1,97 @@
...
+### Attribute name
+
+Attribute names can be created dynamically with string interpolation:
+
+```nix
+let
+  names = [ "foo" "bar" "baz" ];
+  attrs = map (name: { ${name} = true; }) names;
+in
+builtins.foldl' (x: y: x // y) {} attrs
```
tazjin, line 92: I think this is an unnecessarily complex example for someone just trying to understand string interpolation.

fricklerhandwerk:
@tazjin thanks for the review. Could you please make your suggestions actual GitHub suggestions? That's a lot clearer and can be committed immediately.

doc/manual/src/language/string-interpolation.md
fricklerhandwerk: I'd like to make the list exhaustive. What else is missing?

doc/manual/src/language/string-interpolation.md
fricklerhandwerk: Agreed. I simply couldn't come up with a simpler one on the spot. How would you demonstrate dynamic attribute names?

doc/manual/src/language/string-interpolation.md
fricklerhandwerk: Actually, the coercion part is irrelevant for that statement anyway.

```suggestion
When an interpolated expression evaluates to a path, the path is first copied into the Nix store and the resulting string is the [store path].
```

doc/manual/src/language/string-interpolation.md
tazjin: Most important additional thing are attribute sets with the `__toString` attribute (which is a function `self -> string`):

```
nix-repl> "${{ __toString = _: "toString"; }}"
"toString"
```

This is sometimes used to make some kind of "smart value" that has e.g. `__toString` and `__functor` to pretend to be an object or whatever.

Technically saying that `derivations` are coercible to string is also not precise, as it is all attribute sets with an `outPath` attribute:

```
nix-repl> "${{ outPath = "outPath"; }}"        
"outPath"
```

This is the behaviour for what we call "weak coercion" in Tvix (equivalent to `!coerceMore` in C++ Nix), there's slightly different behaviour for "strong coercion" but I'm pretty sure it doesn't apply to interpolation.

doc/manual/src/language/string-interpolation.md
tazjin: I think it can be completely logicless, e.g.:

```
let
  key = "keyName";
in {
  ${key} = 42;
}
```

as we're not trying to demonstrate anything else here.

tazjin:
@fricklerhandwerk I'll do that next time, not sure how to do it (I don't use Github much) and currently don't have time for more than drive-by comments while traveling :)

doc/manual/src/language/string-interpolation.md
fricklerhandwerk: Yes, of course. Now I remember why I didn't put it there: I wasn't sure if we want to expose this feature.

This PR was originally only to rename antiquotation to string interpolation.

In the Nix team meeting we agreed to add a glossary entry to have something to point to.

Then @edolstra asked to extract that into its own section.

I tried to keep changes minimal and preserve the contents as far as possible.

Describing what a derivation is and how it works specifically is a completely different scope, and the changes here are already a bit too much to review easily.

I'd rather postpone going into the `__toString` business to a different PR.

fricklerhandwerk:
@edolstra please check again so we can merge this.

Eventually I would like to re-use the examples with the [`{{#include }}` feature of mdBook](https://rust-lang.github.io/mdBook/format/mdbook.html#including-files). Right now it doesn't work because the Basic Values section is indented in a list, and the inclusion is just dumb string replacement, which breaks the list.

Reworking that would be a different PR and needs a bit more thought and possibly discussion.

tazjin (approved):

doc/manual/src/language/values.md
```diff
@@ -143,26 +111,26 @@
...
-
-  Note that the Nix language assumes that all input files will remain
-  _unchanged_ during the course of the Nix expression evaluation.
-  If you for example antiquote a file path during a `nix repl` session, and
-  then later in the same session, after having changed the file contents,
-  evaluate the antiquotation with the file path again, then Nix will still
-  return the first store path. It will _not_ reread the file contents to
-  produce a different Nix store path.
+  When an [interpolated string][string interpolation] evaluates to a path, the path is first copied into the Nix store and the resulting string is the [store path] of the newly crea
+ted [store object].
```
edolstra, line 115:
```suggestion
  When an [interpolated string][string interpolation] evaluates to a path, the path is first copied into the Nix store and the resulting string is the [store path] of the newly created [store object].
```

doc/manual/src/language/values.md
```diff
@@ -143,26 +111,26 @@
...
+
+  [store path]: ../glossary.md#gloss-store-path
+  [store object]: ../glossary.md#gloss-store-object
+
+  For instance, evaluating `"${./foo.txt}"` will cause `foo.txt` in the current directory to be copied into the Nix store and result in the string `"/nix/store/<hash>-foo.txt"`.
+
+  Note that the Nix language assumes that all input files will remain _unchanged_ while  evaluating a Nix expression.
+
```
edolstra, line 123:
```suggestion
```

doc/manual/src/language/values.md
```diff
@@ -143,26 +111,26 @@
...
+  For example, assume you used a file path in an interpolated string during a `nix repl` session.
+  Later in the same session, after having changed the file contents, evaluating the interpolated string with the file path again will still return the first store path.
+  Nix will _not_ re-read the file contents to produce a different store path.
```
edolstra, line 126:
```suggestion
  Later in the same session, after having changed the file contents, evaluating the interpolated string with the file path again might not return a new store path, since Nix might not re-read the file contents.
```

We shouldn't say that this caching is guaranteed behaviour.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -1,5 +1,7 @@
 # Release X.Y (202?-??-??)
 
+* Instead of "antiquotation", the more common term [string interpolation](../language/string-interpolation.md) is now used consistently.
+  Historical release notes were not changed.
```
edolstra, line 4:
```suggestion
```

doc/manual/src/language/string-interpolation.md
```diff
@@ -0,0 +1,82 @@
...
+Interpolated expressions must evaluate to something that can be coerced to a string, such as:
+
+- [string]
+- [path]
+- [derivation]
+
+[string]: ./values.md#type-string
+[path]: ./values.md#type-path
+[attribute name]: ./values.md#attribute-set
+[derivation]: ../glossary.md#gloss-derivation
```
edolstra, line 16:
```suggestion
- [a string](./values.md#type-string)
- [a path](./values.md#type-path)
- [a derivation](./values.md#attribute-set)
```

doc/manual/src/language/string-interpolation.md
```diff
@@ -0,0 +1,82 @@
+# String interpolation
+
+String interpolation is a language feature where a [string], [path], or [attribute name] can contain expressions enclosed in `${ }` (dollar-sign with curly brackets).
+
+Such a string is an *interpolated string*, and an expression inside is an *interpolated expression*.
+
+Interpolated expressions must evaluate to something that can be coerced to a string, such as:
```
edolstra, line 7:
```suggestion
Interpolated expressions must evaluate to one of the following:
```

doc/manual/src/release-notes/rl-next.md
fricklerhandwerk: Why remove that? Do you think it's self-evident? I added the note because the term _antiquotation_ still appears in the release notes, but one could have changed that as well.

doc/manual/src/language/string-interpolation.md
```diff
@@ -0,0 +1,82 @@
...
+
+String interpolation is a language feature where a [string], [path], or [attribute name] can contain expressions enclosed in `${ }` (dollar-sign with curly brackets).
+
+Such a string is an *interpolated string*, and an expression inside is an *interpolated expression*.
+
+Interpolated expressions must evaluate to one of the following:
+
+- [string]
+- [path]
+- [derivation]
```
fricklerhandwerk, line 11: Counter-proposal to [this](https://github.com/NixOS/nix/pull/7278/files#r1039732574):
```suggestion
- a [string]
- a [path]
- a [derivation]
```

Link titles should reflect the contents, and in this case we're linking to the definition of the type, not a particular instance.

fricklerhandwerk:
@edolstra rebased and applied all suggestions except the line about release notes. Please check.

edolstra:
@fricklerhandwerk Please be careful, this PR deleted several entries from `rl-next.md`...

fricklerhandwerk:
Oh sorry, apparently I made a mistake while rebasing. :( Will be more careful in the future.

PR #7281, opened 2022-11-09
nix-shell: Add temp roots for outputs

jsoo1:
The garbage collector uses heuristics (i.e. /proc/*/environ) to check for live roots - meaning that a nix-shell should theoretically be safe from gc. In practice this detection does not always work. This change makes nix-shell spawn a child process and creates temproots for the shell.

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -5,3 +5,7 @@
   arguments will be ignored and the resulting derivation will have
   `__impure` set to `true`, making it an impure derivation.
 
+* `nix-shell` now spawns a child process instead of `exec`ing immediately.
+
+* `nix-shell` creates temp roots for as long as the executed shell
+  process is alive.
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -573,7 +573,25 @@ static void main_nix_build(int argc, char * * argv)
 
         execvp(shell->c_str(), argPtrs.data());
 
-        throw SysError("executing shell '%s'", *shell);
+        pid_t pid = fork();
+
+        if (pid == -1)
+            throw SysError("forking shell");
+
+        if (pid == 0) {
+            execvp(shell->c_str(), argPtrs.data());
+
+            throw SysError("executing shell '%s'", *shell);
+        }
+
+        siginfo_t shellStatus;
+
+        auto shellExited = waitid(P_PID, pid, &shellStatus, WEXITED);
+
+        if (shellExited == -1)
+            throw SysError("waiting on shell");
+
+        throw Exit(shellStatus.si_status);
     }
 
     else {
--- a/src/nix-build/nix-build.cc
+++ b/src/nix-build/nix-build.cc
@@ -571,14 +571,15 @@ static void main_nix_build(int argc, char * * argv)
 
         logger->stop();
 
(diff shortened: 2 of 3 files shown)


jsoo1:
This was very hard to make a test case for without `ps` at least:
```ShellSession
$ git rev-parse HEAD
239f1e5249a9894a491a9a273f9f32d4ac88a330

$ $(nix-build --no-out-link)/bin/nix-shell -p jq --run '
  ps -eo pid,command | awk "/nix-daemon[[:space:]]+$PPID/ { print \$1 }" | xargs -i_ stat /nix/var/nix/temproots/_
'
  File: /nix/var/nix/temproots/3377944
  Size: 53206           Blocks: 104        IO Block: 4096   regular file
Device: 254,1   Inode: 143941455   Links: 1
Access: (0600/-rw-------)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2022-11-08 18:59:38.294816016 -0800
Modify: 2022-11-08 18:59:38.615815786 -0800
Change: 2022-11-08 18:59:38.615815786 -0800
 Birth: 2022-11-08 18:59:38.294816016 -0800

$ /run/current-system/sw/bin/nix-shell -p jq --run '
  ps -eo pid,command | awk "/nix-daemon[[:space:]]+$PPID/ { print \$1 }" | xargs -i_ stat /nix/var/nix/temproots/_
'
=> empty
```

thufschmitt:
> In practice this detection does not always work

Do you have some examples of it not working? Because if so that's annoying beyond just `nix-shell`

jsoo1:
> > In practice this detection does not always work
> 
> Do you have some examples of it not working? Because if so that's annoying beyond just `nix-shell`

I think https://github.com/NixOS/nix/issues/2208 mentions a few. In our CI system we run into nix-shell interpreter script dependencies being gc'd pretty often when they overlap with a garbage collection. I mentioned that this might be papering over a bug here: https://github.com/NixOS/nix/pull/7248#issuecomment-1308089070

Nonetheless I think temproots are still worth creating for nix-shells since i'd rather not rely on the gc heuristic to catch everything in the shell.

jsoo1:
I think this change also needs an ifdef for darwin.

thufschmitt:
Mh, I couldn't find anything really relevant in #2208, it seems to really be related to keeping the gc root once the shell is closed (which is also something important that we've left aside for too long).

> Nonetheless I think temproots are still worth creating for nix-shells since i'd rather not rely on the gc heuristic to catch everything in the shell.

I don't think quoting this as an heuristic is really fair here. It's heuristic in that a program could refer to a store path without it being in one of the searched paths, but it is a fairly reliable mechanism otherwise.

A possible explanation could be that this interacts with the non-blocking GC mechanism. It kinda requires new roots to be explicitely registered when the GC is running, and maybe the `/proc` scanning doesn't do it properly. @edolstra does that sound possible?

edolstra:
I'm not in favor of this. Keeping the `nix-shell` parent around is potentially very expensive, since a Nix evaluation can consume gigabytes of memory. It would be much better to investigate/fix whatever the GC issue is.

jsoo1:
> a Nix evaluation can consume gigabytes of memory.

Ahh, right. 

>  It would be much better to investigate/fix whatever the GC issue is.

I agree. If nix-shell is to be deprecated, perhaps this change is irrelevant anyways

Issue #6572, opened 2022-05-25
nix-build: requires non-existent output 'out' from input derivation


Izorkin:
**Describe the bug**

Packages build fails very often:
```
checking for true... true
error (ignored): error: cannot unlink '/tmp/nix-build-python3-minimal-3.9.12.drv-1/Python-3.9.12': Directory not empty
error: derivation '/nix/store/6xjmvr03428342gs1432ac0mn8myxnpc-grub-config.xml.drv' requires non-existent output 'out' from input derivation '/nix/store/88vs9d74w69j14s611vb5v7cn0
...
ran test tests/config.sh... [PASS]
installCheckPhase completed in 40 seconds
error: derivation '/nix/store/cqmxz06ggrnhr7wzf6s1kf39hykmsh6g-local-cmds.drv' requires non-existent output 'out' from input derivation '/nix/store/dw0x90b23k67mkav4bf24r63nvn97dir-nix-2.8.1.drv'
...
error: derivation '/nix/store/shnrbk0l8qdsfmvf4x6qp6nlvc0kdfjc-libmodsecurity-3.0.6.drv' requires non-existent output 'out' from input derivation '/nix/store/nrpsq5ld8fjp3adpimdiwkmg7wbhqsq3-lmdb-0.9.29.drv'
...
checking for references to /build/ in /nix/store/90wv2xdvci79wnj8nxlrxy73j1hiksjh-bind-9.18.3-host...
error: derivation '/nix/store/irc43zk5n8w2bp7ig0d5hlfi4ybm3vs9-system-path.drv' requires non-existent output 'host' from input derivation '/nix/store/ll61r00n2f62qkxlll7y0ba45mzby2m4-bind-9.18.3.drv'
```

I have to restart the build process manually each time.
The problem happens on nix versions 2.8.1 and 2.9.0pre20220512_d354fc3
On version 2.8.0 the problem rarely occurred

NickCao:
My guess would be nix starting the next build too early before the outputs of the previous build is properly registered.

milahu:
aka `build input X does not exist` #1744

in something simple as

```nix
{ pkgs }:
with pkgs;
let
  x = fetchFromGitHub {
    # ...
  };
in
stdenv.mkDerivation {
  buildInputs = [ x ];
  # ...
}
```

this time, restarting the build does not help

<details>

[...]

PR #7283, opened 2022-11-09
tests: Reproduce #6572

roberth:
Only `issue_6572_dependent_outputs` fails, suggesting that dependencies between the outputs of a multi-output drv are involved.

```
    + nix-store --delete /run/user/1000/nix-test/tests/build/store/5fv3k7lg7cgnswgiw43xarw07xg1z1wh-multiple-outputs-a-second
    finding garbage collector roots...
    deleting '/run/user/1000/nix-test/tests/build/store/5fv3k7lg7cgnswgiw43xarw07xg1z1wh-multiple-outputs-a-second'
    deleting unused links...
    note: currently hard linking saves 0.00 MiB
    1 store paths deleted, 0.00 MiB freed
    ++ nix build -f multiple-outputs.nix use-a --no-link --print-out-paths
      /run/user/1000/nix-test/tests/build/store/f1qc3m41ffxjfz51rs3ix17xxvihix3r-multiple-outputs-a.drv
      /run/user/1000/nix-test/tests/build/store/p101iqkw9wwqpb7f81h5hxjficl7kizi-use-a.drv
    building '/run/user/1000/nix-test/tests/build/store/f1qc3m41ffxjfz51rs3ix17xxvihix3r-multiple-outputs-a.drv'...
    error: derivation '/run/user/1000/nix-test/tests/build/store/p101iqkw9wwqpb7f81h5hxjficl7kizi-use-a.drv' requires non-existent output 'first' from input derivation '/run/user/1000/nix-test/tests/build/store/f1qc3m41ffxjfz51rs3ix17xxvihix3r-multiple-outputs-a.drv'
```


Ericson2314 (approved):

dasJ:
@Artturin about the backport label, the test fails for me on 2.11:
```
    warning: you don't have Internet access; disabling some network-dependent features
    true
    + testNormalization
    + clearStore
    + echo 'clearing store...'
    clearing store...
    + chmod -R +w /build/nix-test/tests/build/store
    + rm -rf /build/nix-test/tests/build/store
    + mkdir /build/nix-test/tests/build/store
    + rm -rf /build/nix-test/tests/build/var/nix
    + mkdir /build/nix-test/tests/build/var/nix
    + clearProfiles
    + profiles=/build/nix-test/tests/build/var/nix/profiles
    + rm -rf /build/nix-test/tests/build/var/nix/profiles
    ++ nix-build ./simple.nix --no-out-link
      /build/nix-test/tests/build/store/g8zy4ay6vc6wqcli8l4flbzp5g30raxb-simple.drv
    building '/build/nix-test/tests/build/store/g8zy4ay6vc6wqcli8l4flbzp5g30raxb-simple.drv'...
    PATH=
    + outPath=/build/nix-test/tests/build/store/fggyi1r9mxi0xzvx6afhjxfv998nknjs-simple
    ++ stat -c %Y /build/nix-test/tests/build/store/fggyi1r9mxi0xzvx6afhjxfv998nknjs-simple
    + test 1 -eq 1
    + issue_6572_independent_outputs
    + nix build -f multiple-outputs.nix --json independent --no-link
    warning: you don't have Internet access; disabling some network-dependent features
      /build/nix-test/tests/build/store/2w1aqwk5ivsk27w1mab3mm87a9000hbh-multiple-outputs-independent.drv
    building '/build/nix-test/tests/build/store/2w1aqwk5ivsk27w1mab3mm87a9000hbh-multiple-outputs-independent.drv'...
    ++ nix build -f multiple-outputs.nix use-independent --no-link
    warning: you don't have Internet access; disabling some network-dependent features
      /build/nix-test/tests/build/store/rp7hidn5nns87dfwc81al0bizm4cv2ml-multiple-outputs-a.drv
      /build/nix-test/tests/build/store/963i9wavg9m06hl0cvhqh5vrlw8b55bl-use-independent.drv
    building '/build/nix-test/tests/build/store/rp7hidn5nns87dfwc81al0bizm4cv2ml-multiple-outputs-a.drv'...
    building '/build/nix-test/tests/build/store/963i9wavg9m06hl0cvhqh5vrlw8b55bl-use-independent.drv'...
    + p=
    + nix-store --delete ''
    error: path '/build/source/tests' is not in the Nix store
make: *** [mk/lib.mk:123: tests/build.sh.test] Error 1
```

dasJ:
@roberth I agree, but the test relies on `--print-out-paths` which is only available on 2.9+ [source](https://nixos.org/manual/nix/stable/release-notes/rl-2.9.html)

roberth:
I'll fix that in the backport prs.

thufschmitt (approved):
Thanks a lot for digging into this, it's **really** great to see this issue fixed.

Having to query the store again is a bit sad, but hopefully one day we'll have a proper structured control flow, and we won't have to rely on this any more :crossed_fingers:

src/libstore/build/derivation-goal.cc
```diff
@@ -528,13 +528,31 @@ void DerivationGoal::inputsRealised()
...
+                   in itself */
+                if (auto outPath = get(inputDrvOutputs, { depDrvPath, j })) {
                     worker.store.computeFSClosure(*outPath, inputPaths);
-                else
-                    throw Error(
-                        "derivation '%s' requires non-existent output '%s' from input derivation '%s'",
-                        worker.store.printStorePath(drvPath), j, worker.store.printStorePath(depDrvPath));
+                }
+                else {
+                    auto outMap = worker.evalStore.queryDerivationOutputMap(depDrvPath);
```
Kha, line 546: Is this something expensive that should be evaluated at most once in the loop?

roberth: In the vast majority of cases, `inputDrvOutputs` will have the result. Otherwise Nix would have been completely unusable since 2.8.

We can just lift it out of the loop, because that would create many more derivation file queries; for each inputDrv, instead for only very rare cases. Even then the chance of reusing the map is fairly low, because derivations typically don't have many outputs, and this only happens for multi output derivations.

It is important to keep in mind though when implementing the suggestion in the comment above, so I'll update that.

Ericson2314:
@thufschmitt 

> Having to query the store again is a bit sad, but hopefully one day we'll have a proper structured control flow, and we won't have to rely on this any more 🤞

We should talk about this at some point because I feel the exact opposite, that the store should be the single source of truth and we should have as little in-memory state as possible.

Issue #7286, opened 2022-11-11
mix mac install assumed gnu diff and other /etc/synthetic.conf issues


diekhans:
MacOS Ventura, M1
I want to have nix installed under 

**Steps To Reproduce**

Create an /etc/synthetic.conf with /nix

```
nix       System/Volumes/Data/nix
```

and reboot then
```
sh <(curl -L https://nixos.org/nix/install) --daemon
```

Install then tries to edit out the nix line from synthetic.conf.
I don't want it to do this.  I don't want a new disk volume,
I just want nix in System/Volumes/Data/nix

As part of insisting on edit the file, it calls diff with invalid arguments
for MacOs

```
~~> Fixing any leftover Nix volume state                                                                                                
Before I try to install, I'll check for any existing Nix volume config
and ask for your permission to remove it (so that the installer can
start fresh). I'll also ask for permission to fix any issues I spot.

During install, I add 'nix' to /etc/synthetic.conf, which instructs
macOS to create an empty root directory for mounting the Nix volume.

---- warning! ------------------------------------------------------------------                                                        

/etc/synthetic.conf already contains a line instructing your system
to make '/nix' as a symlink:
    3:nix       System/Volumes/Data/nix

This may mean your system has/had a non-standard Nix install.

The volume-creation process in this installer is *not* compatible
with a symlinked store, so I'll have to remove this instruction to
continue.

If you want/need to keep this instruction, answer 'n' to abort.



Nix isn't the only thing in /var/folders/3k/fvyxn7p97976wmlmtkd7ltfr0000z8/T/tmp.XagVkQCxZv/synthetic.conf.edit,
but I think I know how to edit it out.
Here's the diff:
diff: unrecognized option `--unchanged-group-format=\033[2m%=\033[0m'
usage: diff [-aBbdilpTtw] [-c | -e | -f | -n | -q | -u] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            [-I pattern] [-F pattern] [-L label] file1 file2
       diff [-aBbdilpTtw] [-I pattern] [-L label] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            [-F pattern] -C number file1 file2
       diff [-aBbdiltw] [-I pattern] [--ignore-case] [--no-ignore-case]
            [--normal] [--strip-trailing-cr] [--tabsize] -D string file1 file2
[...]

PR #7288, opened 2022-11-11
darwin-install: use abspath for diff

abathur:
Addresses error message from nonstandard diff in #7286.


abathur:
Per ongoing discussion in linked issue this wasn't actually a path issue. Looks like Ventura uses a new Apple diff that we'll need to adapt to...

abathur:
Closing in favor of #7948 (which actually addresses the issue :)

PR #7294, opened 2022-11-12
libstore: link to aws-crt-cpp

tobim:
This change is needed to support aws-sdk-cpp 1.10 and newer.

I opted not to make this dependent on the sdk version because the crt dependency has been in the interface of the older sdk as well, and it was only coincidence that libstore didn't make use of any privately defined symbols directly.


roberth (approved):
> C++ wrapper around the aws-c-* libraries. Provides Cross-Platform Transport Protocols and SSL/TLS implementations for C++.

-- https://awslabs.github.io/aws-crt-cpp/

Builds with current aws sdk version too. Probably a bit redundant in the current version, but at least it keeps the makefile a bit simple.

thufschmitt (approved):
Thanks :)

PR #7300, opened 2022-11-14
doc/book.toml: Improve config

roberth:
- `title` value will be added to the HTML as in `<title>${content.title} - Nix Package Manager Manual</title>`

 - `git-repository-url` adds a link to the GitHub repo in the top right corner

 - `edit-url-template` adds an edit link, inviting contributions
 
 You can see the buttons in action at [flake.parts](https://flake.parts) (exception in the options reference).

### DIFF
--- a/doc/manual/book.toml
+++ b/doc/manual/book.toml
@@ -1,6 +1,11 @@
+[book]
+title = "Nix Package Manager Manual"
+
 [output.html]
 additional-css = ["custom.css"]
 additional-js = ["redirects.js"]
+edit-url-template = "https://github.com/NixOS/nix/tree/master/doc/manual/{path}"
+git-repository-url = "https://github.com/NixOS/nix"
 
 [preprocessor.anchors]
 renderers = ["html"]
--- a/doc/manual/book.toml
+++ b/doc/manual/book.toml
@@ -1,5 +1,5 @@
 [book]
-title = "Nix Package Manager Manual"
+title = "Nix Reference Manual"
 
 [output.html]
 additional-css = ["custom.css"]


doc/manual/book.toml
```diff
@@ -1,6 +1,11 @@
+[book]
+title = "Nix Package Manager Manual"
```
fricklerhandwerk, line 2:
```suggestion
title = "Nix Manual"
```

- Long titles don't fit title bars anyway
- we discussed narrowing Nix down to a package manager may mischaracterise the project
- Nixpkgs and NixOS have the same pattern

fricklerhandwerk (approved):
This is small and awesomely valuable. Thank you so much for taking time.

doc/manual/book.toml
roberth: > * we discussed narrowing Nix down to a package manager may mischaracterise the project

It narrows down the scope of the document, not the scope of the ecosystem. I think it's important to signal, to the extent we reasonably can, that this manual does not document the Nix ecosystem, but the Nix package manager.

> Long titles don't fit title bars anyway

Fair enough, but the full title is shown on hover and, importantly, in the bookmark when creating one.

fricklerhandwerk: > I think it's important to signal, to the extent we reasonably can, that this manual does not document the Nix ecosystem, but the Nix package manager.

I agree and think that signaling and clear distinction should be done externally when people get introduced to Nix ecosystem, say on the web site and materials that refer to more than just Nix itself.

For me this is a matter of separation of concerns. I'd suggest to deliberately pretend that, from the perspective of the Nix manual, nothing else exists. It's only the reference manual, it doesn't have to explain anything, it just has to be unambiguous and correct, and ideally complete. The guiding and explanation is a different problem domain.

The moment we say "Nix Package Manager" all kinds of distracting questions arise:
- Is that what NPM stands for?
- But isn't it also a build system?
- It seems to be so much more, what is it even exactly?

The correct and unambiguous short circuit to that is to call it what it is: Nix. And leave the explanation for a different context.

roberth: > deliberately pretend

The situation with the multiple manuals is far from ideal in the first place. The last thing we should be doing is pretending.

If you call this the Nix Manual, indeed a new reader's association may be with the whole ecosystem. In this case, the least you could do is explain what this manual _does not_ document, to rectify the well intended but unnecessary lie.

> The correct and unambiguous short circuit to that is to call it what it is: Nix.

I completely agree with your ambitious desire to fundamentally improve the documentation, but until we have an overarching, unified documentation system for all core Nix projects, including Nix, Nixpkgs, NixOS, we can not pretend to have a single manual for "Nix".

If your goal is to merge the manuals, I 100% agree with your reasoning, but it only makes sense when we actually implement such a merged manual.

(I believe antora.org now supports markdown, but it'd be a *big* leap still. Maybe there's another comprehensive documentation system out there that's a better match? idk)

fricklerhandwerk: I understand the potential for confusing Nix with the Nix ecosystem.

But we seem to have misunderstood each other. I don't intend to merge all manuals or call
the result Nix, but would rather treat the different projects separately (which to me, for cultural and technical reasons, they clearly are) in their manuals - this is what I meant by pretending in the Nix manual: that for instance there is no Nixpkgs, but just the tools Nix itself offers. Then there is no need to distinguish oneself from the rest. It would be firmly sticking to our concern at hand.

I think those spurious references to Nixpkgs which still float around in the Nix manual are what produced some
of the confusion in the first place (although the major culprit is the contents of nixos.org), as there was no clear boundary as to what comes from where and how the tools interact exactly.

Nixpkgs and NixOS of course should continue to mention Nix as that is what they are based on, and reference the relevant manual sections where appropriate.

And of course we should also have a note on the first page that links to Nixpkgs and NixOS manuals for clarity. But only there.

And then, maybe it's clearer now, enlightening users about the distinction should happen somewhere else. We agreed with @lucperkins and @domenkozar at the docs team meeting last week that nix.dev should eventually end up to be that place where.

roberth: I agree with this principle about only describing yourself and linking to what you use as relevant.

However, I would certainly make an exception for the introductory text of a manual (and possibly an exception for reverse links in clearly distinguishable call-outs).

The latter I completely understand if that's too much for now, but the first is crucial. It takes very little space and doesn't have to set a tone for, let alone permeate the rest of the document, while it adds a ton of value for readers who visit the manual for the first time. Anyone else won't even read the intro, so it's perfectly fine to make this exception.

roberth: (and same exception applies to the title)

fricklerhandwerk: Let's bounce this against the Nix team?

fricklerhandwerk:
  - @thufschmitt: prefer "Nix Package Manager Manual"
  - @edolstra: "Nix Reference Manual"
      - @infinisil, @thufschmitt, @Ericson2314: agree

doc/manual/book.toml
fricklerhandwerk, line 2:
```suggestion
title = "Nix Reference Manual"
```

As [discussed on the team meeting](https://github.com/NixOS/nix/pull/7300#issuecomment-1332911864).

fricklerhandwerk (approved):

PR #7307, opened 2022-11-15
libstore/derivation-goal: Elaborate a TODO for performance concern

roberth:
Addresses https://github.com/NixOS/nix/pull/7283/files/c279ddb18cf3a34b0f6d4e3adcf9455da5397ad7#r1022993437


thufschmitt:
I actually don't think we want to implement that with lookups. That's some info that we should have, we only can't rely on it right now because the control-flow is impossible to understand. And if we have to rely on it, then we probably shouldn't cache it since we'd risk running into the same kind of problem that we've just solved here.

fricklerhandwerk:
Discussed in Nix team meeting 2023-01-06:

- @fricklerhandwerk: sounds like a low-level design issue worth discussing in more detail
    - @Ericson2314 will follow up with a written issue and put it up for triage
- merged

PR #7312, opened 2022-11-16
Allow system certs access to fixed-output derivations

layus:
Allow building fixed-output derivations behind an SSL proxy

This change mounts the system certificates bundle in the sandbox of
fixed-output derivations. This makes sense given that they are already given
network access, and network is pretty much unusable nowadays if you do not have
the right certificates.

The implementation mounts the certificates in a well-known location,
independently of the real location passed to the daemon. This should make it
visible to most tools by default, and still allow overrides if needed.

Other options have been considered, but this seems the most logical one, as it
does not require modification to the nix expressions themselves for something
that is essentially a configuration quirk of the build machine.

Comments ?

**Release Notes**

**Testing**

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1715,6 +1715,8 @@ void LocalDerivationGoal::runChild()
                 for (auto & path : { "/etc/resolv.conf", "/etc/services", "/etc/hosts" })
                     if (pathExists(path))
                         ss.push_back(path);
+
+                dirsInChroot.emplace(settings.caFile, "/etc/ssl/certs/ca-certificates.crt");
             }
 
             for (auto & i : ss) dirsInChroot.emplace(i, i);


zoranbosnjak:
@layus could you please describe how to test this patch. I am trying to rebuild and use this branch, to workaround the SSL cert problem behind proxy, but I am getting the same error.

```
# ... behing the SSL proxy

$ git clone this/branch ~/nix
$ cd ~/nix
$ git rev-parse --short HEAD
a7d2a3d08
$ nix-build 
/nix/store/crb5h86yjz6bgi0b1ywdn2z6qvvca7a1-nix-2.12.0pre20221116_a7d2a3d
$ nix-env -i $(readlink result)
$ nix --version
nix (Nix) 2.12.0pre20221116_a7d2a3d

$ cd ~/hello
$ cat hello.nix 
with (import <nixpkgs> {});

stdenv.mkDerivation rec {
  name = "test";
  src = fetchgit {
    url = "https://git.savannah.gnu.org/git/hello.git";
    rev = "fab5dca74c099b1663dce30d9e11c2d83ecf3078";
    sha256 = "1p0pnwcbgf5aks0zd52k9xkkjn8pfmk475aq640cyd05f06zzlgi";
    fetchLFS = false;
    fetchSubmodules = false;
    deepClone = false;
    leaveDotGit = false;
  };

  installPhase = ''
    mkdir -p $out
    cat ${src}/README > $out/README
  '';
}

$ nix-build hello.nix 
  /nix/store/p1vkixln7dy1am13k429ca2rpqbk1jz8-hello-fab5dca.drv
  /nix/store/gq2v3z5bvwd3a8wxvcfs8kmviwqakza6-test.drv
building '/nix/store/p1vkixln7dy1am13k429ca2rpqbk1jz8-hello-fab5dca.drv'...
exporting https://git.savannah.gnu.org/git/hello.git (rev fab5dca74c099b1663dce30d9e11c2d83ecf3078) into /nix/store/5x5qhmxly7vnpr5ldc0q3wz5z3f9h3q5-hello-fab5dca
Initialized empty Git repository in /nix/store/5x5qhmxly7vnpr5ldc0q3wz5z3f9h3q5-hello-fab5dca/.git/
fatal: unable to access 'https://git.savannah.gnu.org/git/hello.git/': SSL certificate problem: unable to get local issuer certificate
fatal: unable to access 'https://git.savannah.gnu.org/git/hello.git/': SSL certificate problem: unable to get local issuer certificate
fatal: unable to access 'https://git.savannah.gnu.org/git/hello.git/': SSL certificate problem: unable to get local issuer certificate
Unable to checkout fab5dca74c099b1663dce30d9e11c2d83ecf3078 from https://git.savannah.gnu.org/git/hello.git.
error: builder for '/nix/store/p1vkixln7dy1am13k429ca2rpqbk1jz8-hello-fab5dca.drv' failed with exit code 1
error: 1 dependencies of derivation '/nix/store/gq2v3z5bvwd3a8wxvcfs8kmviwqakza6-test.drv' failed to build
```

It looks like the `/etc/ssl/certs/ca-certificates.crt` is setup correctly, since the system has internet connectivity:

```
$ git clone https://git.savannah.gnu.org/git/hello.git/
Cloning into 'hello'...  inside the shell works as expected
```

fricklerhandwerk:

We agree the idea is reasonable, and while the implementation may need more care or follow-up design decisions, we want to move forward with this and @edolstra will review the PR.

- @thufschmitt: change seems reasonable
    - implementation may need refinement
- @ericson2314: is there something like `extraSandboxPaths` but for FODs?
    - such a more general approach might be more scalable
    - perhaps should not try to anticipate all possible specific needs
    - @thufschmitt: for specific things one could repurpose the `pre-build-hook`, but this use case is generic enough to have it there by default
- @edolstra: would just merge it, tests seem to be really tricky
    - @Ericson2314: tests are probably quite feasible
- decision: idea approved, assigned to @edolstra for review

</details>

zoranbosnjak:
@edolstra @thufschmitt 
I will appreciate the fix. But just a quick note that the original fix from this pull request did not work for me. 
This works:
https://github.com/zoranbosnjak/nix/commit/161e6c477cb881b46b7905426247ba61a299e960

Please consider/review/use this change too.

zoranbosnjak:
@edolstra @thufschmitt 
I will appreciate the fix. But just a quick note that the original fix from this pull request did not work for me. 
This works:
https://github.com/zoranbosnjak/nix/commit/161e6c477cb881b46b7905426247ba61a299e960

Please consider/review/use this change too.

zoranbosnjak:
@edolstra @thufschmitt 
I will appreciate the fix. But just a quick note that the original fix from this pull request did not work for me. 
This works:
https://github.com/zoranbosnjak/nix/commit/161e6c477cb881b46b7905426247ba61a299e960

Please consider/review/use this change too.

layus:
Hi @zoranbosnjak, thanks for having a look at this PR. You are right that it does not works as-is. It is an attempt at fixing a not-so trivial problem. The first step would be to merge this into nix. Then, nixpkgs can adapt to it by using this path in the fixed-output derivations.

That is why the patch you have linked does work, but is not complete. https://github.com/zoranbosnjak/nix/commit/161e6c477cb881b46b7905426247ba61a299e960#diff-08a5e40e2243b11f7d074881064ea221e03448cce5bdf0faee0c2fe0f197571dR1769 only works if the certificates are in `/etc/ssl/certs/ca-certificates.crt`, which is not the case on some platforms where nix runs. This cannot be a nixos only solution.

As for the second part https://github.com/zoranbosnjak/nix/commit/161e6c477cb881b46b7905426247ba61a299e960#diff-08a5e40e2243b11f7d074881064ea221e03448cce5bdf0faee0c2fe0f197571dR1139, it is git-specific, and would requires a PR in nix for any other tool that uses a custom env var to find SSL certificates. And yet there is nothing in nix that requires these certificates. `builtins.fetchurl` does not use that sandbox for example. So it's only uselful in nixpkgs, once this PR gets merged. In nixpkgs, every fixed-output derivation will be free to use and setup the SSL certificates the way it needs.

So yes, this is not usable as-is, but I think it is the right patch to merge into nix, and and the other part, also needed, should go into nixpkgs, with a nix version check.

Ericson2314:
It looks like at least part of the problem is that the case where there is no cert file is not covered. I think changing `caFile` from `Setting<Path>` to `Setting<std::optional<Path>>` would help ensure this is not missed. There might be other path settings that could also benefit from this treatment.

layus:
I would assume the path is only mounted if it is found, I did not check that.
Also, is master CI different from PR CI ?
How can I help fixing this ?

Ericson2314:
@layus I think it was just new tests added since this PR. If you do a `path != ""` check in `local-derivation-goal.cc`, I have already started on the `std::optional` clean up. I will just rebase mine on top.

layus:
I would prefer to stick with the `std::optional` solution, so we do not silently ignore a value set to "". But in this case the issue is different. The path is set to a hard-coded value that does not happen to exist on CI apparently. So the fix is a bit more complex. Shall we silently ignore ?

Ericson2314:
Fix: https://github.com/NixOS/nix/pull/8375

PR #7249, opened 2022-11-01
Parse JSON using simdjson

yorickvP:
# Description
 This PR changes the json-to-value function to use [simdjson](https://github.com/simdjson/simdjson), an optimized json library.
It additionally adds an ExprJSON expression in order to generate json thunks, which lazily convert the rest of the document from the simdjson representation to nix Values (allocating nix values turns out to be expensive). These are generated for complex attrset values.

# Motivation
- Some real world nix evaluations spend a significant fraction of the eval time parsing JSON (I've seen up to 50%).
- There has been a [simdjson plugin](https://github.com/serokell/scratch/tree/master/nix-plugins/simdjson) for a while. At NixCon, some people expressed interest in having it upstream.

This mainly saves time on json-heavy libs like haskell.nix and nix-pypi-fetcher, which can be quite slow. The added laziness helps lot for the real world big-attrset-of-pkg-descriptions json files.

# Downsides
- simdjson can't fully replace nlohmann, since it doesn't generate any JSON.
- The generated syntax errors are worse. There is no position information and the error messages are generic. (e.g. "The JSON document has an improper structure: missing or superfluous commas, braces, missing keys, etc.")

# Current state
- I'm not happy with the autoconf code
- Everything should work.

# Benchmarks

<details>
<summary>Real world performance difference (~30%)</summary>

## This PR:

```json
{
  "cpuTime": 3.17696,
  "envs": {
    "number": 9101249,
    "elements": 13673660,
    "bytes": 255009264
  },
  "list": {
    "elements": 3733294,
    "bytes": 29866352,
    "concats": 375405
  },
  "values": {
    "number": 16694476,
    "bytes": 400667424
  },
  "symbols": {
    "number": 114647,
    "bytes": 1468921
  },
  "sets": {
    "number": 2499791,
    "bytes": 474100000,
    "elements": 27131459
  },
  "sizes": {
    "Env": 16,
    "Value": 24,
    "Bindings": 16,
    "Attr": 16
  },
  "nrOpUpdates": 892412,
  "nrOpUpdateValuesCopied": 19229361,
  "nrThunks": 12884099,
  "nrAvoided": 10452895,
  "nrLookups": 5494571,
  "nrPrimOpCalls": 4107432,
  "nrFunctionCalls": 7952752,
  "gc": {
    "heapSize": 1610612736,
    "totalBytes": 1589656096
  }
}
```

## Nix 2.11.0:
```json
{
  "cpuTime": 4.70948,
  "envs": {
[...]

PR #7313, opened 2022-11-16
Replace src/libutil/json.cc with nlohmann

yorickvP:
Spinoff from https://github.com/NixOS/nix/pull/7249

Currently, nix uses two different ways to generate json strings. They collide at one point in `ParsedDerivation::prepareStructuredAttrs`. This PR removes the old code and changes everything to use nlohmann.

A side effect of this is that objects keys are now ordered alphabetically everywhere and that `NIX_SHOW_SYMBOLS` no longer prints a duplicate "symbols" key. It is possible to keep the original ordering where this is desired, but I'm not sure where that's the case.


yorickvP:
cc @edolstra

edolstra:
This makes the output of commands like `nix-env -qa --json` non-streaming right? I.e. it builds the entire `nlohmann::json` tree in memory before writing it to stdout. Not sure if anybody cares about the streaming behaviour though...

thufschmitt:
> Not sure if anybody cares about the streaming behaviour though...

I would say that it's OK to drop it since:
- In the very worst case it's “only” a performance regression for the consumers and doesn't really change the behaviour;
- If it happens to really be a blocker for someone, we can always restore it by manually writing the top-level JSON structure and letting `nlohmann-json` handle the formatting of the inner objects;
- I find that streaming JSON is always a bit clunky because the format isn't really designed for that. So if we need that in the long term we could gain from switching to something like [jsonlines](https://jsonlines.org/) (just a fancy name for a stream of newline-delimited JSON objects) or similar.

The only worry I have regarding the loss of the streaming behaviour is that building the entire object in memory could be expensive. But I expect that's generally not meaningful compared to the memory used for the evaluation.

srhb:
This changed the representation of `builtins.toJSON 1.000001` from `"1"` to `"1.000001"` -- was that intentional? I couldn't find anything in the changelog while debugging some scary changes locally, though I might be missing something.

Issue #7302, opened 2022-11-14
ubsan errors


Mindavi:
**Describe the bug**

I've built Nix with UBSAN on my system (added `NIX_CFLAGS_COMPILE = -fstack-protector-all -fsanitize=undefined -fsanitize-recover=all -fno-common -fno-omit-frame-pointer -O1 -fno-optimize-sibling-calls` to be precise). Whenever I start a build I get these errors:

```
src/libexpr/eval.cc:2579:23: runtime error: member call on address 0x7f2108816340 which does not point to an object of type 'BaseSetting'
0x7f2108816340: note: object has invalid vptr
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
src/libexpr/eval.cc:2579:53: runtime error: member call on address 0x7f21088163c0 which does not point to an object of type 'BaseSetting'
0x7f21088163c0: note: object has invalid vptr
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
src/libexpr/primops.cc:2468:36: runtime error: variable length array bound evaluates to non-positive value 0
src/libexpr/primops.cc:2797:35: runtime error: variable length array bound evaluates to non-positive value 0
src/libexpr/primops.cc:3174:24: runtime error: variable length array bound evaluates to non-positive value 0
```

**Steps To Reproduce**

1. Rebuild with listed flags (main thing is `-fsanitize=undefined`).
2. Start a build with `nix-build` or `nix build` (shouldn't matter).
3. Observe listed errors.

**Expected behavior**

I expect that nix doesn't rely on undefined behavior to work correctly.

**`nix-env --version` output**

```
src/libexpr/eval.cc:2579:23: runtime error: member call on address 0x7f6b8cc16340 which does not point to an object of type 'BaseSetting'
0x7f6b8cc16340: note: object has invalid vptr
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
[...]

PR #7315, opened 2022-11-17
Don't create zero-lengh arrays in the primops

thufschmitt:
Partial fix for https://github.com/NixOS/nix/issues/7302

Some primops take a list as one of their argument, and internally create an array of the same length as the list.
This is UB when the size is zero. It just happens to work because the behavior of the primop is trivial in that case, and we never touch that array.

Fix this by special-casing the empty list case and bailing out early when it happens.

cc @Mindavi


Mindavi (approved):
Seems sane to me. Did not test.

edolstra:
What actually is the problem with zero-length arrays if we never access them? Seems very annoying to have to special-case the zero length case.

thufschmitt:
@edolstra the problem is that creating them is UB. Probably fine in practice, but who knows.
But I've actually opened #7348 because I realized that they just weren't valid C++, so better get rid of them altogether

PR #7318, opened 2022-11-18
add "instantiate" to glossary

fricklerhandwerk:
add a minimal definition and refer to `nix-instantiate`.

### DIFF
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -19,6 +19,13 @@
 
     [store derivation]: #gloss-store-derivation
 
+  - [instantiate]{#gloss-instantiate}, instantiation\
+    Translate a [derivation] into a [store derivation].
+
+    See [`nix-instantiate`](./command-ref/nix-instantiate.md).
+
+    [instantiate]: #gloss-instantiate
+
   - [realise]{#gloss-realise}, realisation\
     Ensure a [store path] is [valid][validity].
 


doc/manual/src/glossary.md
```diff
@@ -7,6 +7,59 @@
     translated into low-level *store derivations* (implicitly by
     `nix-env` and `nix-build`, or explicitly by `nix-instantiate`).
 
+    [derivation]: #gloss-derivation
+
+  - [store derivation]{#gloss-store-derivation}\
+    A [derivation] represented as a `.drv` file in the [store].
+    It has a [store path], like any [store object].
+
+    See [`nix show-derivation`](./command-ref/new-cli/nix3-show-derivation.md) (experimental) for displaying the contents of store derivations.
```
sir4ur0n, line 16: These new links are inconsistent with others in this file.

E.g. on line 141, the link is `command-ref/conf-file.html#conf-substituters`:
* no prefix `./`
* ends in `.md` vs `.html`

Is there a reason to have different link syntaxes?

sir4ur0n, line 16: I don't see any `new-cli` directory in https://github.com/fricklerhandwerk/nix/tree/doc-instantiate/doc/manual/src/command-ref

When I click on the link from a Github page, I land - as expected - on a 404

Is this intended? If so, would you mind explain to me why this will work _in fine_ on the documentation/website?

doc/manual/src/glossary.md
```diff
@@ -7,6 +7,59 @@
     translated into low-level *store derivations* (implicitly by
     `nix-env` and `nix-build`, or explicitly by `nix-instantiate`).
 
+    [derivation]: #gloss-derivation
+
+  - [store derivation]{#gloss-store-derivation}\
+    A [derivation] represented as a `.drv` file in the [store].
+    It has a [store path], like any [store object].
```
sir4ur0n, line 14: What do you think about giving an example of `drv` path here? E.g. `/nix/store/0a2qvbq8b9xvrijp83p5m81xcw2xpgig-x509-validation-1.6.12.tar.gz.drv`

doc/manual/src/glossary.md
```diff
@@ -7,6 +7,59 @@
...
+    ```
+
+    See [`nix-instantiate`](./command-ref/nix-instantiate.md).
+
+    [instantiate]: #gloss-instantiate
+
+  - [realise]{#gloss-realise}, realisation\
+    Produce a [store object] from a [store derivation].
+
+    This means either running the `builder` executable as specified the corresponding [derivation] or fetching a pre-built [store object] from a [substituter].
```
sir4ur0n, line 39:
```suggestion
    This means either running the `builder` executable as specified in the corresponding [derivation] or fetching a pre-built [store object] from a [substituter].
```

doc/manual/src/glossary.md
fricklerhandwerk: Fixed.

doc/manual/src/glossary.md
fricklerhandwerk: Done.

doc/manual/src/glossary.md
fricklerhandwerk: These files are generated in the [Makefile](https://github.com/NixOS/nix/blob/f91dc023f27ea17362274fcdf30e362423b080f8/doc/manual/local.mk#L51-L53) through a [Nix expression](https://github.com/NixOS/nix/blob/f91dc023f27ea17362274fcdf30e362423b080f8/doc/manual/generate-manpage.nix) that is fed by running the `nix` executable.

sir4ur0n: I can't resolve threads (thanks Github permissions), but this now looks good

sir4ur0n: For my own culture: is there a CI check that all links are valid in the generated documentation?

fricklerhandwerk:
Took out all the unrelated parts (covered in other PRs), now this only adds a term to the glossary.

Ericson2314 (approved):

Ericson2314:
I do still hope per #6507 we will drop "store derivation", but this is correct for now. If/when we do, we will just revise this.

roberth (approved):
I've used the term loosely for adding sources to the store, which is required for adding derivations to the store, as derivations reference sources. This usage might be controversial and/or not worth adding to the definition here, so +1 on the current change from me.

tomberek (approved):
LGTM

PR #7320, opened 2022-11-18
restructure attribute listing to `derivation`

fricklerhandwerk:
# Motivation

The most important part of the language is still documented somewhat sparsely, in somewhat dense language. This is a first step to provide some more predictable structure that is easier to parse.

# Context

We still need many more examples and cross-links. This PR starts out on just the structure and the essentials.

# Checklist for maintainers

### DIFF
--- a/doc/manual/src/language/derivations.md
+++ b/doc/manual/src/language/derivations.md
@@ -71,11 +71,15 @@ It outputs an attribute set, and produces a [store derivation](@docroot@/glossar
   Examples:
 
   Imagine a library package that provides a dynamic library, header files, and documentation.
-  A program that links against the library doesn’t need the header files and documentation at runtime, and it doesn’t need the documentation at build time.
+  A program that links against such a library doesn’t need the header files and documentation at runtime, and it doesn’t need the documentation at build time.
   Thus, the library package could specify:
 
   ```nix
-  outputs = [ "lib" "dev" "doc" ];
+  derivation {
+    # ...
+    outputs = [ "lib" "dev" "doc" ];
+    # ...
+  }
   ```
 
   This will cause Nix to pass environment variables `lib`, `dev`, and `doc` to the builder containing the intended store paths of each output.
@@ -90,22 +94,12 @@ It outputs an attribute set, and produces a [store derivation](@docroot@/glossar
 
   for an Autoconf-style package.
 
-  You can refer to each output of a
-  derivation by selecting it as an attribute, e.g.
-
-  ```nix
-  buildInputs = [ pkg.lib pkg.dev ];
-  ```
-  <!-- FIXME: move this to the output attributes section when we have one -->
+  You can refer to each output of a derivation by selecting it as an attribute, e.g. `myPackage.lib` or `myPackage.doc`.
 
   The first element of `outputs` determines the *default output*.
-  Thus, you could also write
+  Therefore, in the given example, `myPackage` is equivalent to `myPackage.lib`.
 
-  ```nix
-  buildInputs = [ pkg pkg.dev ];
-  ```
-
-  since `pkg` is equivalent to `pkg.lib`.
+  <!-- FIXME: refer to the output attributes when we have one -->
 
 - See [Advanced Attributes](./advanced-attributes.md) for more, infrequently used, optional attributes.
 
@@ -115,9 +109,9 @@ It outputs an attribute set, and produces a [store derivation](@docroot@/glossar
   Attribute values are translated to environment variables as follows:
 
     - Strings are passed unchanged.
-    
+
     - Integral numbers are converted to decimal notation.
-    
+
     - Floating point numbers are converted to simple decimal or scientific notation with a preset precision.
 
     - A *path* (e.g., `../foo/sources.tar`) causes the referenced file
--- a/doc/manual/src/language/derivations.md
+++ b/doc/manual/src/language/derivations.md
@@ -24,7 +24,7 @@ It outputs an attribute set, and produces a [store derivation](@docroot@/glossar
 
   The system type on which the [`builder`](#attr-builder) executable is meant to be run.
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
@Ericson2314 [commented](https://github.com/NixOS/nix/pull/7332#discussion_r1029317693):
> Building is strictly speaking store derivation -> store object (without using cache). I wouldn't include evaluation. nix build doing evaluation too is a porcelain convenience that shouldn't effect the definition of "building".

doc/manual/src/glossary.md
```diff
@@ -1,11 +1,95 @@
...
+
+  - [build]{#gloss-build}\
+    Produce a [store object].
+
+    See [`nix-build`](./command-ref/nix-build.md).
+
+    See [`nix build`](./command-ref/new-cli/nix3-build.md) (experimental).
+
+    Nix uses a two-stage build process:
+    A Nix expression that evaluates to a [derivation] is [instantiate]d to a [store derivation], which is then [realise]d to a [store object].
```
sir4ur0n, line 16: What do you think about making this two-stage build more visible with a numbered list?
Incidentally I think it makes it easier to read and grasp the explanation
E.g.

```suggestion
    Nix uses a two-stage build process:
    1. A Nix expression that evaluates to a [derivation] is [instantiate]d to a [store derivation]
    2. This [store derivation] is then [realise]d to a [store object].
```

sir4ur0n: Minor English comment: "realised to" looks weird to me (non-native English speaker), I would have intuitively said "realised into", but I fail to find a convincing source of truth about either

doc/manual/src/glossary.md
```diff
@@ -1,11 +1,95 @@
...
+
+    ```
+    +--------------------+
+    | Nix language       |
+    | .----------------. |
+    | | Nix expression | |
+    | '----------------' |
+    |         |          |
+    |      evaluate      |
```
sir4ur0n, line 25: I find this explanation and diagram slightly confusing because "evaluate" is not a definition in this document, unlike `instantiate` and `realise`: it looks like it is yet another command and stage.

I would have expected "Nix expression of type `derivation`" rather than "Nix expression - evaluate -> derivation".

Ericson2314:
More broadly, I think if we have https://github.com/NixOS/nix/pull/4543 then the `DerivedPath` type (in the code, and in my NixCon talk) will finally be directly use-visible, and I think that in turn will unlock what I think are some big improvements in being able to explain things.

doc/manual/src/glossary.md
fricklerhandwerk: There is no such thing as a type `derivation`, and even the Nix language concept of a derivation is vague. It's an attribute set with two magic attributes `drvPath` and `outPath`, where the latter is also used as the value in string interpolation. See https://github.com/NixOS/nix/pull/7332#discussion_r1029343720

I can add a definition for evaluation though, but maybe in a different PR to keep this one small.

sir4ur0n (approved):

tomberek:
notes from meeting:

### realization
"Ensure a store path is valid (it and all of its transitive dependencies are in the store); either a derivation's output, or the store path itself if provided directly. When used with a derivation, Nix will either build or substitute the necessary store paths. When used directly with a store path, the necessary store paths will only be substituted, if available."

Two cases:
1) derivation ---(via building or substitution) ----> storePath is valid in a store
2) store path ---(via substitution) ---------------------> storePath is valid in a store

This operation will fail if the supplied derivation/store paths cannot be created by either building or substitution.

valid: this is a property of a store path in a store which ensures it and all transitive references to other store paths are readable using that store (for example: on a local FS store this means "on disk").... WIP

fricklerhandwerk:
Discussed in Nix team meeting 2022-12-19:

- @edolstra: there should instead be a separate section about the store
    - @fricklerhandwerk: that's what the Architecture chapter will have
- @edolstra: merge it as is and move around the information when there is a better place
- (some discussion over wording of the precise definition of validity)
- @tomberek: we should focus on what's a blocker to merging and split out tangential considerations into issues or follow-up PRs
- @fricklerhandwerk: will rework the wording of definitions according to discussion outcome

Ericson2314:
I think you have another PR's text stuck in here?

fricklerhandwerk:
Ah yeah the force push didn't get through. Will clean it up when I get back to a keyboard.

fricklerhandwerk:
The build failing on linkcheck is very weird, as the highlighted link is actually correct. Checked with @roberth.

doc/manual/src/language/derivations.md
```diff
@@ -1,161 +1,201 @@
...
+
+  `system = "x86_64-linux";`
+
+  `system = builtins.currentSystem;`
+
+  [`builtins.currentSystem`](@docroot@/language/builtin-constants.md#builtins-currentSystem) has the value of the [`system` configuration option], and defaults to the system type of the current Nix installation.
+
+  [`system` configuration option]: @docroot@/command-ref/conf-file.md#conf-system
+
+- [`builder`]{#attr-builder} ([Path](@docroot@/language/values.md#type-path) | [String](@docroot@/language/values.md#type-string))
```
tomberek, line 39: specify that "/bin/sh" is somewhat special?

roberth: I don't think we want people to use `/bin/sh` except through `system(2)` and such, so I'd rather push it away into a section of platform quirks.

OT: I think we should give derivation authors the option of specifying their own `/bin/sh` as a derivation attribute, at least on Linux where this is certainly feasible. Not sure about darwin.

fricklerhandwerk: What's the change suggestion here?

fricklerhandwerk:
Any blockers for merging this?

fricklerhandwerk:
Merging this after addressing all comments and observing that the review seems to converge to minutiae.

Issue #4758, opened 2021-05-01
builtins.match regex support "broken" on aarch64-darwin


andir:
**Describe the bug**

I just tried to use `poetry2nix` on `aarch64-darwin` and the eval failed `invalid regular expression`. Apparently this is an issue with the aarch64-darwin build of Nix. I'm using Nixpkgs from https://github.com/NixOS/nixpkgs/pull/105026.

The problem appears to only exist on aarch64-darwin and not on any of the platforms. Also nixUnstable vs stable nix doesn't make a difference.

## aarch64-darwin

```
[aarch64-darwin] $ nix repl
Welcome to Nix version 2.3.10. Type :? for help.

nix-repl> builtins.match "^.*?(egg|tar|tar.bz2|tar.gz|tar.lz|tar.lzma|tar.xz|tbz|tgz|tlz|txz|whl|zip)" "asd"
error: invalid regular expression '^.*?(egg|tar|tar.bz2|tar.gz|tar.lz|tar.lzma|tar.xz|tbz|tgz|tlz|txz|whl|zip)', at (string):1:1

[aarch64-darwin] $ nix repl
Welcome to Nix version 2.4pre20210326_dd77f71. Type :? for help.

nix-repl> builtins.match "^.*?(egg|tar|tar.bz2|tar.gz|tar.lz|tar.lzma|tar.xz|tbz|tgz|tlz|txz|whl|zip)" "asd"
error: invalid regular expression '^.*?(egg|tar|tar.bz2|tar.gz|tar.lz|tar.lzma|tar.xz|tbz|tgz|tlz|txz|whl|zip)'

       at «string»:1:1:

            1| builtins.match "^.*?(egg|tar|tar.bz2|tar.gz|tar.lz|tar.lzma|tar.xz|tbz|tgz|tlz|txz|whl|zip)" "asd"
             | ^
            2|

```

## x86_64-linux

```
[x86_64-linux] $ nix repl
Welcome to Nix version 2.3.10. Type :? for help.

nix-repl> builtins.match "^.*?(egg|tar|tar.bz2|tar.gz|tar.lz|tar.lzma|tar.xz|tbz|tgz|tlz|txz|whl|zip)" "asd"
null

nix-repl> builtins.currentSystem
"x86_64-linux"


[x86_64-linux] $ nix repl
Welcome to Nix version 2.4pre20210326_dd77f71. Type :? for help.

[...]

Issue #2147, opened 2018-05-09
builtins.match regression: stack overflow on large strings


xzfc:
NixOS 18.03:
```
Welcome to Nix version 2.0. Type :? for help.

nix-repl> large-string = let self = a: if a == 0 then "a" else let s = self (a - 1); in s+s; in self 19 # string of 524288 bytes

nix-repl> builtins.match ".*" large-string
error: stack overflow (possible infinite recursion)
```

NixOS 17.09:
```
$ nix-repl
Welcome to Nix version 1.11.16. Type :? for help.

nix-repl> large-string = let self = a: if a == 0 then "a" else let s = self (a - 1); in s+s; in self 19 # string of 524288 bytes

nix-repl> builtins.match ".*" large-string
[ ]
```

-----

Due to this regression, `nixpkgs-mozilla` overlay stoped working on NixOS 18.03. https://github.com/mozilla/nixpkgs-mozilla/issues/90

I've reduced the case to the following:
```
$ nix repl
Welcome to Nix version 2.0. Type :? for help.

nix-repl> builtins.match ".*[\n]([0-9a-f]*)  linux-x86_64/en-US/firefox-60.0.tar.bz2..*" (builtins.readFile (builtins.fetchurl http://download.cdn.mozilla.net/pub/firefox/releases/60.0/SHA512SUMS ))
error: stack overflow (possible infinite recursion)
```

xzfc:
I've bisected this to the commit b05b98df7544d02387f583ca5434f33f3e9cb471 (replace own regex class with std::regex) that merged in #1098.

And it seems it's a bug in `libstdc++`.
```
#include <regex>
#include <string>
#include <iostream>

int main (int argc, const char * argv[]) {
    std::regex regex{".*"};
    std::string str{"a"};
    for (int i = 0; i < 19; i++)
	    str += str;
    std::cerr << str.length() << std::endl; // 524288
    std::cerr << std::regex_match(str, regex) << std::endl; // Segmentation fault
}
```

copumpkin:
In case anyone was wondering, the libc++ version works fine on Nix 2.0 on macOS.

dtzWill:
> In case anyone was wondering, the libc++ version works fine on Nix 2.0 on macOS.

Ah, that's probably why I wasn't able to reproduce either (libc++, musl, linux).  Curious.

zimbatm:
[...]

PR #7336, opened 2022-11-23
Use PCRE for `builtins.match` and `builtins.split`

yorickvP:
Fixes https://github.com/NixOS/nix/issues/2147, fixes https://github.com/NixOS/nix/issues/4758
See also https://github.com/NixOS/nix/pull/3826.

This is not technically fully compatible, but I can't find a builtins.match invocation in the wild that doesn't work, since programs could only rely on quite a limited subset anyways.
Performance is similar because no one really used a lot of regexes, but this should theoretically be faster.

Additionally adds support for named captures, so you can do
```
nix-repl> builtins.match "(?<date>(?<year>(\\d\\d)?\\d\\d) - (?<month>\\d\\d) - (?<day>\\d\\d))" "2020 - 10 - 10"
{ date = "2020 - 10 - 10"; day = "10"; month = "10"; year = "2020"; }
```


src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+}
+
+static RegisterPrimOp primop_split({
+    .name = "__split",
+    .args = {"regex", "str"},
+    .doc = R"s(
+      Returns a list composed of non matched strings interleaved with the
+      lists of the [extended POSIX regular
+      expression](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04)
+      *regex* matches of *str*. Each item in the lists of matched
```
oxalica, line 330: PCRE is not POSIX ERE.

Also worth mentioning that PCRE supports arbitrary look-around, which can cause (maybe accidental) exponential run time. Maybe we should limit the input to a reasonable subset.

oxalica (changes requested):

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+    pcre2_code* code;
+    size_t usage = 0;
+    std::vector<std::pair<std::string_view, uint32_t>> name_table;
+public:
+    Regex(std::string_view re)
+    {
+        int errorcode;
+        PCRE2_SIZE erroffset;
+
+        code = pcre2_compile((const unsigned char*)re.data(), re.length(), 0, &errorcode, &erroffset, nullptr);
```
tfc, line 29:
```suggestion
        code = pcre2_compile(static_cast<const unsigned char*>(re.data()), re.length(), 0, &errorcode, &erroffset, nullptr);
```

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+    return rc;
+};
+
+};
+
+struct RegexCache
+{
+    // TODO use C++20 transparent comparison when available
+    std::unordered_map<std::string_view, PCRE::Regex> cache;
+    std::list<std::string> keys;
```
tfc, line 150: if this is performance sensitive, i suggest just using `std::vector<std::string>`

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+        return cache.emplace(keys.back(), keys.back()).first->second;
+    }
+};
+
+std::shared_ptr<RegexCache> makeRegexCache()
+{
+    return std::make_shared<RegexCache>();
+}
+
+size_t regexCacheSize(std::shared_ptr<RegexCache> cache)
```
tfc, line 167:
```suggestion
size_t regexCacheSize(const std::shared_ptr<RegexCache> &cache)
```

this should help avoiding incrementing and decrementing the reference counter just for looking up the keys size.

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+{
+    return cache->keys.size();
+}
+
+static void extract_matches(EvalState & state, const PCRE::MatchData& match, Value & v)
+{
+    auto nameTable = match.regex().nameTable();
```
tfc, line 174: sprinkling const in front of all these local variables would be helpful when assessing what is part of changing state and what isn't.

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+    if (nameTable.size()) {
+        // try to extract named bindings
+        auto bindings = state.buildBindings(nameTable.size());
+        for (auto i : nameTable) {
```
tfc, line 178: without `const&` we're needlessly copying these items which probably has performance impact

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+            Value & elem = bindings.alloc(i.first);
+            if (!match[i.second].has_value()) elem.mkNull();
+            else elem.mkString(*match[i.second]);
+        }
+        v.mkAttrs(bindings);
+    } else {
+        // the first match is the whole string
+        const size_t len = match.size() - 1;
```
tfc, line 186: it feels like asserting that the size is `> 0` would increase the safety of this function

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+        state.mkList(v, len);
+        for (size_t i = 0; i < len; ++i) {
+            if (!match[i+1].has_value())
+                (v.listElems()[i] = state.allocValue())->mkNull();
+            else
+                (v.listElems()[i] = state.allocValue())->mkString(*match[i + 1]);
```
tfc, line 192:
```suggestion
            auto *val = state.allocValue()
            v.listElems()[i] = val;
            if (!match[i+1].has_value())
                val->mkNull();
            else
                val->mkString(*match[i + 1]);
```

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+}
+
+void prim_match(EvalState & state, const PosIdx pos, Value * * args, Value & v)
+{
+    auto re = state.forceStringNoCtx(*args[0], pos);
```
tfc, line 199: can we add some asserts here?

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+      Evaluates to `[ "FOO" ]`.
+    )s",
+    .fun = prim_match,
+});
+
+/* Split a string with a regular expression, and return a list of the
+   non-matching parts interleaved by the lists of the matching groups. */
+void prim_split(EvalState & state, const PosIdx pos, Value * * args, Value & v)
+{
+    auto re = state.forceStringNoCtx(*args[0], pos);
```
tfc, line 264: same here, asserts would improve debuggability if we ever end up with code that inserts nullptrs

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+        size_t lastmatchend = 0;
+        size_t newmatchstart = 0;
+        while (rc != PCRE2_ERROR_NOMATCH && newmatchstart <= str.length()) {
+            // Add a string for non-matched characters.
+            Value * prefix = state.allocValue();
+            prefix->mkString(str.substr(lastmatchend, match.startPos() - lastmatchend));
+            result.push_back(prefix);
+
+            // Add a list for matched substrings.
+            auto elem = state.allocValue();
```
tfc, line 295: the code is a bit inconsistent in choosing the type for `allocValue` calls. It's either `Value*` or `auto` and it would be nice to choose one variant and stick to it.

tfc (changes requested):

src/libexpr/primops/regex.cc
edolstra: So it should point to https://www.pcre.org/original/doc/html/pcrepattern.html.

edolstra:
Looks good to me. The main objection is that this makes the Nix language specification depend on "whatever PCRE does". That's not a real problem for Nix, but might be a problem for people who want to reimplement it without having a dependency on a C library.

src/libexpr/primops/regex.cc
```diff
@@ -0,0 +1,360 @@
...
+            v.mkNull();
+            return;
+        }
+        extract_matches(state, match, v);
+
+    } catch (RegexError & e) {
+        state.debugThrowLastTrace(EvalError({
+            .msg = hintfmt("error while evaluating regex '%s': ", re, e.what()),
+            .errPos = state.positions[pos]
+        }));
```
thufschmitt, line 220: Actually the above doesn't look so great, maybe rather something like

```suggestion
        e.addTrace(state.positions[pos], "while evaluating regex '%s'", re);
        state.debugThrowLastTrace(e);
```

thufschmitt (changes requested):
I like this overall :+1: left a few inline comments, but nothing too big.

Since this is a potential breaking change, I would like to see it feature-gated for at least a release cycle. Maybe on by default, but with an easy off switch in case it's hurting people. That would require keeping the old code around of course, but that'd just be temporary.

alyssais:
Is there no portable library we could use that implements POSIX ERE? That would avoid compatibility problems between Nix versions, avoid the footguns of backtracking etc., and be something more reasonable to find multiple implementations of for Nix reimpls.

oxalica:
> Is there no portable library we could use that implements POSIX ERE? That would avoid compatibility problems between Nix versions, avoid the footguns of backtracking etc., and be something more reasonable to find multiple implementations of for Nix reimpls.

`boost` also supports regex in many flavor, [including POSIX ERE](https://www.boost.org/doc/libs/1_52_0/libs/regex/doc/html/boost_regex/syntax/basic_extended.html). Could it be a choice?

edolstra:
`boost::regex` is essentially the same as `std::regex`, except that it has a dependency on `icu4c` which makes it very bloated.

oxalica:
> `boost::regex` is essentially the same as `std::regex`, except that it has a dependency on `icu4c` which makes it very bloated.

Technically it should not, if we don't use the Unicode part. The issue is that currently `boost` is dynamically linked and have `--with-icu` by default...

tfc:
> While `std::regex` is known (at least for me) to be fragile and broken at different toolchain/platforms.

`std::regex` is also known to be really slow. I was involved in a project where we, based on significant differences in performance after benchmarking it changed to google's re2 library. (not suggesting that it should be re2, but `std::regex` is simply slow)

edolstra:
> The issue is that currently boost is dynamically linked and have --with-icu by default...

If we can leave out icu support using an override on boost, I'm fine with that.

Ericson2314:
One option is that we support both regexes during a migration period.

alyssais:
Another option for a portable POSIX ERE implementation might be [musl's](https://git.musl-libc.org/cgit/musl/tree/src/regex)?  It's permissively licensed, and only a few thousand lines total, so it would be reasonable for Nix to just bundle it, IMO.

fricklerhandwerk:
Discussed in the Nix team meeting on 2022-12-05:

- decision: @yorickvP please try building `boost` without `icu4c`. the requirement is full backwards compatibility and minimal closure size.
  - (@alyssais that was before your comment. @edolstra anything speaking against `musl`?)

  - @edolstra: if we did ran tests against public flakes, we could use that to detect breakages on regexes
  - @Ericson2314: could have both implementations
      - @edolstra: strongly against having multiple implementations. this would be a bad architectural idea in terms of maintainability and elegance
      - @thufschmitt: could have both and a policy that one of them will be turned off at some point in the future
  - (more argument over reverting Nix versions and compatibility policy)
  - decision: ask author to try building `boost` without `icu4c`
      - in that case, we get around both constraints: compatibility and closure 

</details>

yorickvP:
boost builds without icu (it's mostly used for Boost.Locale afaict), but it requires an ugly `overrideAttrs` or a nixpkgs patch
```nix
(boost.override { icu = null; }).overrideAttrs (o: {
  configureFlags = (lib.take 4 o.configureFlags) ++ [ "--without-icu" ];
})
```

roberth:
While I do think fixing the stack behavior is a higher priority than extending the language with PCRE-specific features, adding PCRE may be feasible.

I believe PCRE was designed to process all EREs correctly, but I haven't found definitive evidence of this. Finding this would greatly help, so if anyone knows better where to look...

To some degree, any change in behavior could be a breaking change. (cue spacebar heating) However, I'd be willing to assume that nobody is validating ERE regexes by trying to use them in builtins.match regex "", in large part because tryEval can't catch a bad regex anyway. This should make the "boolean" whether the extended functionality is present unobservable from within the language.

So we have two possible approaches: either "prove" that PCRE processes EREs correctly, or use a good ERE library instead.

thufschmitt:
@roberth afaik, PCRE is _mostly_ a superset of Posix EREs, but not strictly.
An example of an (obscure) ERE regex construct that is not PCRE one is [character equivalents](https://www.regular-expressions.info/posixbrackets.html#eq):

```console
$ echo e | grep --extendes-regexp '[[=e=]]'
e
$ echo e | grep --perl-regexp '[[=e=]]'
grep: POSIX collating elements are not supported
```

Now the question is whether this really matters, and I'm not really convinced it does given how niche it is

roberth:
> An example of an (obscure) ERE regex construct that is not PCRE one is [character equivalents](https://www.regular-expressions.info/posixbrackets.html#eq):

Ooh, that's locale dependent. I didn't manage to exploit that as a potential impurity (but maybe I did it wrong). Maybe it's just not implemented in GNU stdc++, but it may be implemented in a library that does.

What helps in this case is that libpcre prints an error, and doesn't continue with a garbage regex. If anyone does rely on regexes with collating elements, they'll be able to detect the problem in time, and work around it by applying regexes to their, presumably foreign input, regexes. If character equivalents are already truly broken in Nix (which they _should_ be, as it's impure), this is hardly a regression.
Similarly, collating sequences must not be supported; also locale dependent.

May have to dig through this...
https://gist.github.com/CMCDragonkai/6c933f4a7d713ef712145c5eb94a1816#feature-comparison
`[:alpha:]` looked like a potential deal breaker, but seems ok in grep.
So the bracket classes (bottom of page) seem kind of ok. Haven't looked at the rest yet.

SuperSandro2000:
Just my 2 cents on this:
boost is notorious for massively increasing compile times and for the 50k lines c++ code nix has, it takes very long to compile.
Not sure if buying more into boost is a good considering compile time.

yorickvP:
Superseded by https://github.com/NixOS/nix/pull/7762

PR #7337, opened 2022-11-23
Fix why-depends for CA derivations

Radvendii:
why-depends assumed that we knew the output path of the second argument. For CA derivations, we might not know until it's built. One way to solve this would be to build the second installable to get the output path.

In this case we don't need to, though. If the first installable (A) depends on the second (B), then getting the store path of A will necessitate having the store path B. The contrapositive is, if the store path of B is not known (i.e. it's a CA derivation which hasn't been built), then A does not depend on B.

Fix the remaining issue in #4661

### DIFF
--- a/src/nix/why-depends.cc
+++ b/src/nix/why-depends.cc
@@ -84,18 +84,34 @@ struct CmdWhyDepends : SourceExprCommand
         auto package = parseInstallable(store, _package);
         auto packagePath = Installable::toStorePath(getEvalStore(), store, Realise::Outputs, operateOn, package);
         auto dependency = parseInstallable(store, _dependency);
-        auto dependencyPath = Installable::toStorePath(getEvalStore(), store, Realise::Derivation, operateOn, dependency);
-        auto dependencyPathHash = dependencyPath.hashPart();
+        auto derivedDependency = dependency->toDerivedPath();
+        auto optDependencyPath = std::visit(overloaded {
+            [](const DerivedPath::Opaque & nodrv) -> std::optional<StorePath> {
+                return { nodrv.path };
+            },
+            [&](const DerivedPath::Built & hasdrv) -> std::optional<StorePath> {
+                if (hasdrv.outputs.size() != 1) {
+                    throw Error("argument '%s' should evaluate to one store path", dependency->what());
+                }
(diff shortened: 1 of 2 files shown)


src/nix/why-depends.cc
```diff
@@ -84,19 +84,35 @@ struct CmdWhyDepends : SourceExprCommand
         auto package = parseInstallable(store, _package);
         auto packagePath = Installable::toStorePath(getEvalStore(), store, Realise::Outputs, operateOn, package);
         auto dependency = parseInstallable(store, _dependency);
-        auto dependencyPath = Installable::toStorePath(getEvalStore(), store, Realise::Derivation, operateOn, dependency);
-        auto dependencyPathHash = dependencyPath.hashPart();
+        auto derivedDependency = dependency->toDerivedPath();
```
thufschmitt, line 87: Maybe add a comment explaining why we don't want to build it?

thufschmitt (changes requested):
A minor request, but :+1: otherwise. Thanks :)

Can you just add a test? I think it's enough to do like in https://github.com/NixOS/nix/blob/master/tests/ca/build-dry.sh, just re-source the `why-depends.sh` test but with `NIX_TESTS_CA_BY_DEFAULT=1`

Radvendii:
Is it worth wrapping this:
```
std::visit(overloaded {
            [](const DerivedPath::Opaque & nodrv) -> std::optional<StorePath> {
                return { nodrv.path };
            },
            [&](const DerivedPath::Built & hasdrv) -> std::optional<StorePath> {
                if (hasdrv.outputs.size() != 1) {
                    throw Error("argument '%s' should evaluate to one store path", dependency->what());
                }
                auto outputMap = store->queryPartialDerivationOutputMap(hasdrv.drvPath);
                auto maybePath = outputMap.find(*hasdrv.outputs.begin());
                if (maybePath == outputMap.end()) {
                    throw Error("unexpected end of iterator");
                }
                return maybePath->second;
            },
        }, derivedDependency.raw())
 ```
 
 into a function? This seems like a common thing to want "give me the store path if you already have it, otherwise nullopt"

thufschmitt (approved):

PR #7339, opened 2022-11-23
`exportReferencesGraph` does not work

### LINKED ISSUE #7330: `__unsafeDiscardOutputDependency` used with `exportReferencesGraph` does not work
**Describe the bug**

`__unsafeDiscardOutputDependency` used with `exportReferencesGraph` cause a build failure when the output paths of the drv File used with cannot be found in the sandbox because they are not yet built.

**Steps To Reproduce**

TODO make exact example; it can be used as a test.

**Expected behavior**

drv outputs are ignored, only the drv file itself matters.

**Additional context**

1. `__unsafeDiscardOutputDependency` is not unsafe at all, it is just used to make a derivation depend on a drv file itself and *not* its outputs. (Just depending on the drv itself should probably be the default, but isn't for historical reasons.) How does it work? It removes any `=` the string context starts with. Implemented in https://github.com/NixOS/nix/blob/62960f32915909a5104f2ca3a32b25fb3cfd34c7/src/libexpr/primops/context.cc#L34-L44

2. In https://github.com/NixOS/nix/blob/62960f32915909a5104f2ca3a32b25fb3cfd34c7/src/libexpr/primops.cc#L1202-L1211 the `=` is handled. It causes the newly-built derivation to have far more dependencies -- in addition to the drv file itself, *all* its drv file dependencies *and* all their outputs are depended upon.
  
4. `exportReferencesGraph` is implemented in large part by `Store::eexporReferences` in https://github.com/NixOS/nix/blob/62960f32915909a5104f2ca3a32b25fb3cfd34c7/src/libstore/store-api.cc#L823-L836. If a derivation is encountered, it also adds its outputs to the closure.

It seems that the export references code is wrong:

1. It is assuming all drv paths got the `=` treatmeant but this is not necessarily the case. For ones that didn't (using `__unsafeDiscoardOutputDependency`), we cannot start processing their outputs because those are not necessarily dependencies of the derivation.

2. The code is unnecessary, because the treatment of `=` in the evaluator will *already* ensure all those paths are in `inputPaths`/`inputDrvs`, and https://github.com/NixOS/nix/blob/62960f32915909a5104f2ca3a32b25fb3cfd34c7/src/libstore/build/derivation-goal.cc#L532-L560 snures that the closure of those is already in `inputPaths`

As such we should be able to do this:

1. Write test confirming the current failure in `tests/export-graph.nix`. See there is one for a `drvPath` without `__unsafeDiscardOutputDependency` aready.

2. Delete the code in `Store::exportReferences` that seems pointless, and confirm the bug is fixed.

See 28972864879a63c4110211d9852b692672869a2a for what I think introduced the bug.

aciceri:
Closes #7330 (hopefully)

### DIFF
--- a/tests/export-graph.sh
+++ b/tests/export-graph.sh
@@ -22,9 +22,7 @@ nix-store --gc # should force rebuild of input-1
 
 outPath=$(nix-build ./export-graph.nix -A 'foo."bar.buildGraph"' -o $TEST_ROOT/result)
 
-checkRef input-1
 checkRef input-1.drv
-checkRef input-2
 checkRef input-2.drv
 
 for i in $(cat $outPath); do checkRef $i; done
--- a/tests/export-graph.sh
+++ b/tests/export-graph.sh
@@ -26,3 +26,14 @@ checkRef input-1.drv
 checkRef input-2.drv
 
 for i in $(cat $outPath); do checkRef $i; done
+
+# Test the export of the dependency graph when discarding outputs
+
+nix-store --gc
+
+outPath=$(nix-build ./export-graph.nix -A 'foo."bar.buildGraphWithoutOutDeps"' -o $TEST_ROOT/result)
+
+! [[ $(checkRef "dep-a$") ]]
+checkRef "dep-a.drv"
+! [[ $(checkRef "dep-b") ]]
+checkRef "dep-b.drv"
--- a/tests/export-graph.sh
+++ b/tests/export-graph.sh
@@ -4,7 +4,7 @@ clearStore
 clearProfiles
 
 checkRef() {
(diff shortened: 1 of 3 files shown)


tests/export-graph.sh
```diff
@@ -22,9 +22,18 @@ nix-store --gc # should force rebuild of input-1
 
 outPath=$(nix-build ./export-graph.nix -A 'foo."bar.buildGraph"' -o $TEST_ROOT/result)
 
-checkRef input-1
 checkRef input-1.drv
-checkRef input-2
```
aciceri, line 27: @Ericson2314 Why had I to remove these two checks? This test should build also outputs even after removing that piece of code, right?

Issue #4910, opened 2021-06-12
Built-in YAML support


mweinelt:
**Is your feature request related to a problem? Please describe.**
Currently in nixpkgs YAML is often render as JSON and `pkgs.formats.yaml` denotes 

```
# # YAML has been a strict superset of JSON since 1.2
```

which glances over the fact that JSON is only a subset of YAML and continues to render it as JSON.

Specifically we have no proper way to denote builtin and user-defined types, which are unquoted strings prefixed with two or one exclamation mark.

**Describe the solution you'd like**
A more complete mapping from Nix to YAML.

**Describe alternatives you've considered**
Rendering an attrset to JSON, then using remarshals `json2yaml` to convert it to YAML and then apply a regular expression to unquote strings that start with an exclamation mark.

Home-assistant uses these to substitute secrets and do other includes at runtime.

```nix
{ pkgs, ... }:
let
  configFile = pkgs.runCommand "configuration.yaml" { preferLocalBuild = true; } ''
    ${pkgs.remarshal}/bin/json2yaml -i ${configJSON} -o $out
    # Hack to support custom yaml objects,
    # i.e. secrets: https://www.home-assistant.io/docs/configuration/secrets/
    sed -i -e "s/'\!\([a-z_]\+\) \(.*\)'/\!\1 \2/;s/^\!\!/\!/;" $out
  '';
in
{}
```

**Additional context**
https://en.wikipedia.org/wiki/YAML#Advanced_components

roberth:
You would have to represent these advanced components in the Nix language somehow; perhaps:

```nix
{ type = "yaml-advanced-component"; value = "..."; }
```

With such a representation, nothing should stop you from converting that to proper YAML in a derivation.

There's some concern about keeping Nix itself easy to bootstrap, so adding a yaml library dependency comes at a cost.

For TOML we only have a parsing function, so we don't need IFD to read those files into Nix values. The same might apply for YAML, but that's not the use case you're describing.

mweinelt:
I still care about this feature.

DavHau:
I would be very happy to have a `builtins.fromYAML`.
YAML is commonly used in other package ecosystems to express package metadata.
This metadata could be converted to derivations via nix without IFD or any extra steps if nix had a parsing function.
To give some examples:
  - yarn.lock (yarn v2) for nodejs
  - stack.yaml and stack.yaml.lock for haskell

ners:
[...]

PR #7340, opened 2022-11-23
Add support for parsing YAML

NaN-git:
# Why `builtins.fromYAML`?

YAML is widely used amongst other package managers and deployment tools. If we want better compatibility to these ecosystems, the ability to parse yaml efficiently in nix is useful, as it is for TOML and JSON, which we already support. Further discussion can be found, i.a., in [#4910](https://github.com/NixOS/nix/issues/4910#issuecomment-1296261500).

# Description

YAML 1.2 is a complex standard and nix has a limited set of data types. Thus only a subset of YAML can be represented in nix. For example attribute sets require String keys, i.e. attribute sets can represent YAML maps with String keys only, and nix has no data types for binary data or dates. If `builtins.fromYAML` encounters YAML with incompatible data types, then it fails similar to `builtins.fromTOML`.

First, the implementation uses [rapidyaml](https://github.com/biojppm/rapidyaml) to parse the YAML string and afterwards the nix objects are created while traversing the YAML tree similar to `builtins.fromTOML`. Tags are mostly ignored and affect only scalars. Custom tags are always ignored and I don't see how custom tags could be handled by nix.

# Why rapidyaml?

As part of nix robustness and safety of the implementation of `builtins.fromYAML` are important.

Some reasons why rapidyaml was chosen:

- C++11 library with limited usage of dynamic memory allocations
- tested with a wide range of compilers and on different platforms
- extensive set of unit tests
- heavy usage of assertions within the library
- available as single header file, i.e. easy to bootstrap, so that no build tool could link another version of rapidyaml
- good speed in comparison to yaml-cpp and other YAML libraries

## limitations of rapidyaml

rapidyaml has a few [limitations](https://github.com/biojppm/rapidyaml#known-limitations). Most of these limitations are not really relevant for the nix use case because of the limitations of nix.

Some comments with respect to the limitations:
- The macro `RYML_WITH_TAB_TOKENS` is defined for `builtins.fromYAML`.
- Contrary to the listed bullets rapidyaml seems to support anchor names terminated with a colon.
- Otherwise the only issue that I found is that a string/block containing only tab-, space- and new line-characters might be incorrectly parsed as empty string ([test case](https://github.com/yaml/yaml-test-suite/blob/main/src/Y79Y.yaml)). I don't think that this limits the real world usage.

Also rapidyaml parses some invalid YAML successfully, but that is actually helpful. Otherwise the valid JSON `{"a":"b"}`, which is emitted by `builtins.toJSON`, could not be parsed by `builtins.fromYAML` because it is [no valid YAML](https://yaml.org/spec/1.2.2/#separation-spaces) due to the missing separation white space after the `:`-token.

# Tested platforms
- x86_64-linux
- aarch64-linux


DavHau:
Like mentioned in https://github.com/NixOS/nix/issues/4910#issuecomment-1214148665, we might want to consider to encode the yaml version into the function somehow. Instead of `fromYAML` it could be named `fromYAML1_2`. In case the YAML spec ever gets updated in an incompatible way, we could make a second function, like for example `fromYAML2_0` and won't have to update/break the existing builtin.

NaN-git:
> Like mentioned in [#4910 (comment)](https://github.com/NixOS/nix/issues/4910#issuecomment-1214148665), we might want to consider to encode the yaml version into the function somehow. Instead of `fromYAML` it could be named `fromYAML1_2`. In case the YAML spec ever gets updated in an incompatible way, we could make a second function, like for example `fromYAML2_0` and won't have to update/break the existing builtin.

If this should be implemented, then a decision about the versioning scheme is needed and how the versioning should be done, i.e. whether the builtins should have different names or whether a function with additional version argument should be used instead. Having explicitly different function names would be advantageous with respect to reproducibility.  

The versioning should apply to different code versions, not only to different YAML standards, so that bug compatibility can be retained. The YAML standard is ugly and we need to parse real world data, i.e. the "YAML" might be invalid, so that some tradeoffs are required...

I'm wondering how updates of `builtins.fromJSON` and `builtins.fromTOML` are handled. `builtins.fromJSON` uses nlohmann-json, but if nix isn't built by nix, then the linked library version isn't fixed. Did this cause any problems yet?

tfc (changes requested):

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,155 @@
...
+
+namespace nix {
+
+struct NixContext {
+    EvalState & state;
+    const PosIdx pos;
+    std::string_view yaml;
+};
+
+static void s_error [[ noreturn ]] (const char* msg, size_t len, ryml::Location loc, void *nixContext)
```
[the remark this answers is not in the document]
NaN-git: That's the signature required by rapidyaml... `loc` is not used because `msg` contains the location of the error.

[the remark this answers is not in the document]
tfc: Ok, can you then please drop the parameter's name? it would signal to the reader "this is the signature i need, but i don't intend to use `loc` so it is *not* a mistake that it is never accessed."

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,155 @@
...
+{
+    auto context = (const NixContext *) nixContext;
```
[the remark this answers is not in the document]
NaN-git: You shouldn't look at other parts of nix... In this case `static_cast` is not really safer than a C-type cast because `void*` has not much type information. So it's more a matter of convention, but of course I can apply your suggestion.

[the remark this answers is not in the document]
tfc: Of course i have no authority so define what's right if there is enough historic precedence of the C-way but it also looks like it gets worse if we just stay at it. It would be awesome to gradually get back to C++

DavHau:
> Like mentioned in [#4910 (comment)](https://github.com/NixOS/nix/issues/4910#issuecomment-1214148665), we might want to consider to encode the yaml version into the function somehow. Instead of `fromYAML` it could be named `fromYAML1_2`. In case the YAML spec ever gets updated in an incompatible way, we could make a second function, like for example `fromYAML2_0` and won't have to update/break the existing builtin.

I discussed this once more with @NaN-git and it doesn't seem like a good idea to encode the yaml spec version in the function name. `rapidyaml` doesn't distinguish between yaml versions. It's hard to forsee the reason for which we would have to introduce a new `fromYaml` in the future, if ever.

Whatever naming schema we would come up right now, there is a high chance that it won't be meaningful. Therefore I think the builtin should just be introduced as a plain `fromYaml`. If it ever needs to be updated then we can still make a `fromYaml_2` or whatever makes sense in that situation.

domenkozar:
@edolstra any reason not to merge this?

tomberek:
## Decisions
- [x] agreement on not vendoring
- [x] agreement on making this an experimental feature

## Issues to clarify
- [ ] Nix's compile time impact?
- [ ] closure size?
- [ ] static link?
- [ ] show various features and how they map to Nix values
    - [ ] add tests for the features
- [ ] testing framework improvements
    - [ ] add explanation of testing
    - [ ] convert bash to google-test approach?

## Versioning
If the underlying library returns different results over time, this impacts reproducibility

NaN-git:
> ## Issues to clarify
> [...]
>     * [ ]  show various features and how they map to Nix values
>       
>       * [ ]  add tests for the features
> 
>     * [ ]  testing framework improvements
>       
>       * [ ]  add explanation of testing
>       * [ ]  convert bash to google-test approach?
Regarding the tests:
- The tests from [yaml-test-suite](https://github.com/yaml/yaml-test-suite) are converted to gtest tests by a bash script. The resulting test suite is part of the PR.
- The tests contain a JSON representation ("`json`") of the YAML ("`yaml`") test cases. Thus the tests check whether `builtins.fromJSON json == builtins.fromYAML json == builtins.fromYAML yaml` holds true by serializing the nix expressions. Several hundred test cases are executed and compatibility between `builtins.fromJSON` and `builtins.fromYAML` is verified at least for these cases. If `yaml` should be invalid then the test case is skipped because rapidyaml will often parse this without error.
- `builtins.fromYAML` converts a YAML file to the following nix types: maps, lists, strings, integers (int64) and floats (double). Conversion to custom data types or interpreting data as nix expression would be very dangerous.
Alias nodes (references) are automatically resolved, e.g.
```nix
builtins.fromYAML ''
---
values:
  - &value someValue
--- 
# some YAML template
template: *value
''
```
evaluates to (I'm not very happy with the handling of multiple documents)
```nix
[
  {
    values = [ "someValue" ];
  }
  {
    template = "someValue";
  }
]
```

## Build questions

rapidyaml uses cmake as build system. I don't think that this a good choice for nix because bootstrapping of nix has to be easy. The only dependency of rapidyaml is [c4core](https://github.com/biojppm/c4core) from the same author.
Thus it should be easy to statically link the library. At the moment the library is included as single header file similar to `builtins.fromTOML`.
What is the best way to include this library into nix?

Even with rapidyaml as single header file the build time of `builtins.fromYAML` seems to be faster than the build time of `builtins.fromTOML` and the size of the object file is much smaller than the size of the `builtins.fromTOML` object file.

Ericson2314:
> What is the best way to include this library into nix?

I would package it and its dependency in Nixpkgs. I am happy to help with that part if you want.

------

I personally am not to concerned about the bootstrapping because we could always make YAML a (compile time of Nix) optional feature.

NaN-git:
> I would package it and its dependency in Nixpkgs. I am happy to help with that part if you want.
> 
> I personally am not to concerned about the bootstrapping because we could always make YAML a (compile time of Nix) optional feature.

Packaging rapidyaml is rather easy:
```nix
{ cmake
, fetchFromGitHub
, git
, stdenv
, enableStatic ? true
}:
stdenv.mkDerivation rec {
  pname = "rapidyaml";
  version = "0.5.0";

  src = fetchFromGitHub {
    owner = "biojppm";
    repo = pname;
    rev = "v${version}";
    fetchSubmodules = true;
    hash = "sha256-1/P6Szgng94UU8cPFAtOKMS+EmiwfW/IJl2UTolDU5s=";
  };

  nativeBuildInputs = [ cmake git ];
  cmakeFlags = [
    "-DRYML_WITH_TAB_TOKENS=ON"
    "-DBUILD_SHARED_LIBS=${if enableStatic then "OFF" else "ON"}"
  ];
}
```
I prefer to statically link this because the static library is rather small (~500 KB).

In my opinion not only bootstrapping of nix has to be easy, but compiling it for other distributions should be easy, too. I don't know how to add rapidyaml as optional dependency to nix, so that builds without nix aren't messed up.

NaN-git:
I removed rapidyaml as single header file and made it an optional library instead. If the library cannot be found then `builtins.fromYAML` is not available. Rapidyaml is not part of nixpkgs yet, so that I added the package to `flake.nix` and it will be linked statically.
With `builtins.fromYAML` the size of libnixexpr increases by 800KB and in my tests on an AMD 5800U the build time of nix increased by 4-5 seconds to ~176 seconds when compiling with 16 threads.

Regarding the testing framework: I need more information what's actually required/wanted. I commented some point in https://github.com/NixOS/nix/pull/7340#issuecomment-1409605494. Of course it would be possible to preprocess the tests differently so that less logic is needed in the testing code and it would be more obvious which tests are executed actually.

fricklerhandwerk:
@edolstra what do you think about an optional built-in? Doesn't sound right to me.

roberth:
> > If the library cannot be found then builtins.fromYAML is not available.
> 
> @edolstra what do you think about an optional built-in? Doesn't sound right to me.

Supporting a build without yaml library might be useful for some build-from-source bootstrapping process, but in this case it must be an explicit choice to remove the yaml feature and create an incomplete Nix. It must not happen by accident.

roberth:
Whether that's even a worthwhile effort, I don't know. Personally I think bootstrapping Nix from source is not an important use case, but packaging by other distros is. Those distros must not package a Nix without yaml though! I'd rather require rapidyaml unconditionally.

NaN-git:
@roberth I implemented option no. 2. Now `configure` fails if rapidyaml is not available unless it is disabled explicitly.

NaN-git:
### Lazy evaluation?
I have another thought: Would it make sense to make the implementation lazy like in #7249? I think that it won't have a large impact on the runtime of `builtins.fromYAML` because in my large test cases most of the runtime was actually spent in parsing the YAML. Of course the runtime of the YAML parser depends a lot on the structure of the parsed YAML and its format because as example, block scalars, double-quoted and single-quoted scalars use different code paths.
In case of large, deeply nested YAML (or JSON) lazy evaluation might speed up nix in general, if the YAML tree is only evaluated partially because many strings might not be created and the GC would track less objects.

Ericson2314:
It is nice it is not vendored, but we should still have the dependency in Nixpkgs. Please make a Nixpkgs PR for that, and I will ensure it is merged and backported to 22.11 so it can be used in this PR right away.

`enableStatic` can go away, and instead one can pass it a static stdenv which will have the same effect.

The optional dependency alleviates the bootstrapping concerns.

roberth:
### Reproducibility...

This is a little concerning. The rapidyaml documentation [says](https://github.com/biojppm/rapidyaml#known-limitations):

> Also, ryml tends to be on the permissive side where the YAML standard dictates there should be an error; in many of these cases, ryml will tolerate the input. This may be good or bad, but in any case is being improved on (meaning ryml will grow progressively less tolerant of YAML errors in the coming releases). So we strongly suggest to stay away from those dark corners of YAML which are generally a source of problems, which is a good practice anyway.

What that means for us is that expressions that happen to depend on flawed yaml will become impossible to reproduce, and we have no way of knowing that in advance; not without invoking a more precise implementation first.
Perhaps we should use a more compliant library?

NaN-git:
> ### Reproducibility...
> 
> This is a little concerning. The rapidyaml documentation [says](https://github.com/biojppm/rapidyaml#known-limitations):
> 
> > Also, ryml tends to be on the permissive side where the YAML standard dictates there should be an error; in many of these cases, ryml will tolerate the input. This may be good or bad, but in any case is being improved on (meaning ryml will grow progressively less tolerant of YAML errors in the coming releases). So we strongly suggest to stay away from those dark corners of YAML which are generally a source of problems, which is a good practice anyway.
> 
> What that means for us is that expressions that happen to depend on flawed yaml will become impossible to reproduce, and we have no way of knowing that in advance; not without invoking a more precise implementation first. Perhaps we should use a more compliant library?

This point was already discussed previously and one of the reasons why I vendored the library initially. In my opinion rapidyaml is one of the most compliant YAML libraries when parsing valid YAML. You can have a look at [YAML playground](https://play.yaml.io/main/parser?input=IyBFZGl0IE1lIQoKJVlBTUwgMS4yCi0tLQpmb286IEhlbGxvLCBZQU1MIQpiYXI6IFsxMjMsIHRydWVdCmJhejoKLSBvbmUKLSB0d28KLSBudWxsCg==) and see that for some edge cases basically each library returns a different result. YAML is a complex standard and it's neither the best standard. Therefore tradeoffs have to be made. Additionally only a subset of YAML can be mapped easily to nix.

I don't like to make promises regarding non-valid YAML, which implies undefined behaviour and undefined behaviour should be avoided (I don't want to rant too much about YAML). Nix is written in C++, so there are big chances that nix relies on UB and it is therefore not reproducible without restrictions.
In #7249 simdjson was considered as JSON library. It has another issue of reproducibility because it introduces different code paths for different CPU architectures, i.e. willingness to trade reproducibility against performance seems to exist.
One of the major use cases of `builtins.fromYAML` is to read (possible large) lock files, i.e. as exchange format and only a subset of (hopefully valid) YAML is required for this purpose. Using YAML just because it's possible is probably a bad idea for nix.
The performance of the library cannot be neglected. If it would be 10x or 100x slower than rapidyaml, then its usefulness would be limited.
One of my most important criteria was the security and safety of the library because it is part of nix and it might be exposed to untrusted input. Also it should not introduce more memory leaks into nix because the inputs might be huge.

roberth:
> I don't like to make promises regarding non-valid YAML,

So then we need a very strict parser.
Accepting a file today is a promise to keep accepting that file tomorrow.

> Nix is written in C++, so there are big chances that nix relies on UB

We're doing a decent job at avoiding UB, and I don't think UB is a contributor to evaluation reproducibility problems in practice.
The existence of UB is not an excuse to choose to undermine reproducibility.

> In #7249 simdjson was considered as JSON library. It has another issue of reproducibility because it introduces different code paths for different CPU architectures, i.e. willingness to trade reproducibility against performance seems to exist.

This PR was rejected by the team because of the error messages and the extra maintenance. Why would it cause extra maintenance? Partly because it's more code, and partly because we need a high bar for correctness, and that's because of reproducibility.

> One of the major use cases of `builtins.fromYAML` is to read (possible large) lock files, i.e. as exchange format and only a subset of (hopefully valid) YAML is required for this purpose.

I'd be ok with parsing some useful subset of YAML, as long as we can guarantee to keep parsing at least that subset in the same way in all future versions of Nix.
What we can not do is have two subsets: one that we intend to support and one that we actually support. Now this seems like a familiar software engineering problem, but our forward compatibility needs flip the importance of the differences around. Most applications would only care about intended behavior not being supported, aka bugs. In our case however, the opposite difference is equally bad or worse - unintended behavior being supported. That's because this breaks our promise that non-reproducible evaluations are rejected. That's not just a bug, but an unintended increase in the surface area that we have to support. It may even lead to conflicting requirements about how certain cases should be handled.

> Using YAML just because it's possible is probably a bad idea for nix. The performance of the library cannot be neglected. If it would be 10x or 100x slower than rapidyaml, then its usefulness would be limited. One of my most important criteria was the security and safety of the library because it is part of nix and it might be exposed to untrusted input. Also it should not introduce more memory leaks into nix because the inputs might be huge.

These are of course also valid requirements.

Regarding performance and lock files, we have a promising path to an alternative solution using RFC 92 dynamic derivations. That allows the lock parsing and fixed-output derivation generation to happen in a derivation, so that the lock parsing, processing and instantiation work is cached. Instantiating such a derivation is certainly faster than parsing and generating many derivations at evaluation time, and it avoids the reproducibility issues by having the expression pin the YAML implementation that's used.

DavHau:
There is [StrictYAML](https://hitchdev.com/strictyaml/) that parses and validates a [restricted subset](https://hitchdev.com/strictyaml/features-removed) of the [YAML](https://hitchdev.com/strictyaml/what-is-yaml) specification.
The reference implementation is in python, but there is also one [in rust](https://github.com/fralalonde/strict-yaml-rust). I could not find a C++ implementation.

roberth:
StrictYAML parses booleans into strings. While it makes sense to require a schema for booleans support when you're writing a new application, I think this may be a deal breaker for a Nix builtin.
It is not a meaning-preserving subset of YAML and not a superset of JSON either.

src/libexpr/tests/yaml.cc
```diff
@@ -0,0 +1,161 @@
...
+                            fromYAML(state, noPos, &pYaml, yamlVal); // should throw exception, if fail is asserted, and has to throw, if emptyJSON is asserted
+                        } catch (const EvalError &e) {
+                            continue;
+                        }
+                    }
+                    EXPECT_FALSE(emptyJSON) << "Testcase[" + std::to_string(ctr) + "]: Parsing empty YAML has to fail";
+                    // ryml parses some invalid YAML successfully
+                    // EXPECT_FALSE(fail) << "Testcase[" << ctr << "]: Invalid YAML was parsed successfully";
+                }
+                return "OK";
```
edolstra, line 151: Isn't it cleaner to return a boolean? Then the call sites can just be `ASSERT(execYAMLTest(...))`.

src/libexpr/tests/yaml.cc
```diff
@@ -0,0 +1,161 @@
...
+#include "libexpr.hh"
+
+// Ugly, however direct access to the SAX parser is required in order to parse multiple JSON objects from a stream
+#include "json-to-value.cc"
+
+
+namespace nix {
+// Testing the conversion from YAML
+
+    static std::string replaceUnicodePlaceholders(std::string_view str) {
```
edolstra, line 12: What is `replaceUnicodePlaceholders()` needed for? Maybe add a comment to clarify.

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+    auto valTypeCheck = [=] (ryml::YamlTag_e tag, bool defaultVal = true) {
+        auto valTag = ryml::TAG_NONE;
+        if (t.has_val_tag()) {
+            auto tag = t.val_tag();
+            valTag = tag == "!" ? ryml::TAG_STR : ryml::to_tag(tag);
+        }
+        return valTag == ryml::TAG_NONE ? defaultVal : valTag == tag;
+    };
+
+    v.mkBlackhole();
```
edolstra, line 47:
```suggestion
```
We shouldn't rely on `mkBlackhole()` leaving the value as a thunk. It's better to handle the fall-through case (i.e when the YAML cannot be represented) in another way (maybe by simplifying the logic into a single "if ... else" chain).

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+        if (valTypeCheck(ryml::TAG_STR) && v.type() == nThunk) {
+            std::string_view value(val.begin(), val.size());
+            v.mkString(value);
+        }
+    }
+    if (v.type() == nThunk) {
+        auto val = t.has_val() ? t.val() : "";
+        auto tag = t.has_val_tag() ? t.val_tag() : "";
+        auto msg = ryml::formatrs<std::string>(
+            "Error: The YAML value '{}' with '{}' tag cannot be represented as Nix data type", val, tag);
```
edolstra, line 106: The test suite should have some tests for YAML that cannot be represented (i.e. asserting that we get an error message).

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+      ```
+
+      returns the value `{ x = [ 1 2 3 ]; y = "null"; z = null; }`.
+
+      Maps are converted to attribute sets, but attribute sets require String keys, so that no other key data types are supported.
+
+      Scalars are converted to the type specified by their optional value tag and parsing fails, if a conversion is not possible.
+      Not all YAML types are supported by Nix, e.g. Nix has no binary and timestamp data types, so that parsing of YAML with any of these types fails.
+      Custom tags are ignored and a stream with multiple documents is mapped to a list except when the stream contains a single document.
+    )",
```
edolstra, line 128: This should be an experimental primop initially to gather user feedback, i.e.:

```suggestion
    )",
    .experimentalFeature = Xp::FromYaml,
```

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+            auto msg = ryml::formatrs<std::string>(
+                "Error: Nix supports string keys only, but the key '{}' has the tag '{}'", t.key(), t.key_tag());
+            s_error(msg.data(), msg.size(), {}, &context);
+        }
+    }
+    if (t.is_map()) {
+        if (t.num_children() == 1) { // special case for YAML string ":"
+            auto child = t.child(0);
+            if (child.key().empty() && !child.is_key_quoted() && child.has_val() && child.val().empty() && !child.is_val_quoted()) {
+                v.mkNull();
```
edolstra, line 60: What does 'special case for YAML string ":"' mean? Why does it map to null (rather than, say, an error)?

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+    } else if (valTypeCheck(ryml::TAG_NULL) && t.val_is_null()) {
+        v.mkNull();
+    } else if (t.has_val()) {
+        bool _bool;
+        NixFloat _float;
+        NixInt _int;
+        bool isQuoted = t.is_val_quoted();
+        auto val = t.val();
+        // Caution: ryml is able to convert integers into booleans and ryml::from_chars might ignore trailing chars
+        if (t.has_val_tag() && t.val_tag() == "!" && val.empty() && !isQuoted) { // special case for YAML string "!"
```
edolstra, line 88: Same thing, why does `!` map to null?

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+    .doc = R"(
+      Convert a YAML 1.2 string to a Nix value, if a conversion is possible. For example,
+
+      ```nix
+      builtins.fromYAML ''{x: [1, 2, 3], y: !!str null, z: null}''
+      ```
+
+      returns the value `{ x = [ 1 2 3 ]; y = "null"; z = null; }`.
+
+      Maps are converted to attribute sets, but attribute sets require String keys, so that no other key data types are supported.
```
edolstra, line 123:
```suggestion
      Maps are converted to attribute sets, but only strings are supported as keys.
```

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+
+      Scalars are converted to the type specified by their optional value tag and parsing fails, if a conversion is not possible.
```
edolstra, line 125:
```suggestion
      Scalars are converted to the type specified by their optional value tag. Parsing fails if a conversion is not possible.
```

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,156 @@
...
+      Not all YAML types are supported by Nix, e.g. Nix has no binary and timestamp data types, so that parsing of YAML with any of these types fails.
+      Custom tags are ignored and a stream with multiple documents is mapped to a list except when the stream contains a single document.
```
edolstra, line 127: Should custom tags be ignored or should they throw an error? Or maybe map to the "type" attribute?

src/libexpr/tests/yaml.cc
NaN-git: The test files contain some Unicode characters to indicate the presence/absence of some characters like spaces, tabs and newlines: https://github.com/yaml/yaml-test-suite/tree/main#special-characters
`replaceUnicodePlaceholders()` replaces these characters.

src/libexpr/primops/fromYAML.cc
NaN-git: Good catch. Actually this should throw an error. Now I recognize that I'm parsing the test files incorrectly. YAML `:` is equivalent to YAML `null: null` (see [Example 8.18](https://yaml.org/spec/1.2.2/#822-block-mappings)), but this cannot be converted to JSON, i.e. it's not equivalent to JSON `{ "null" :null}` or `null`.

brianmcgee:
Adding my bump, it would be really handy if this got merged.

71g3pf4c3:
Bumping this. Must-have option

adrian-gierakowski:
this is badly needed

jim3692:
I have a hacky solution to that, but I am not sure how to implement it in nixpkgs, or how to expose it as a function from a flake.

My idea for parsing YAML consists of two parts:

1. A package that takes a YAML file path as an argument, and converts it to JSON, like this:
```nix
{ stdenv, pkgs, inputFile, ... }:

stdenv.mkDerivation {
  name = "yaml-to-json";
  dontUnpack = true;
  dontBuild = true;

  installPhase = ''
    mkdir -p "$out"
    cd "$out"
    cp "${inputFile}" input.yml
    ${pkgs.yq-go}/bin/yq -p yaml -o json ./input.yml >./result.json
  '';
}
```

2. A call to the package and parsing the JSON, like this:
```nix
let
	jsonResult = pkgs.callPackage ./yaml-to-json.nix { inputFile = ./the/yaml/file.yml; };
	parsedYaml = (builtins.fromJSON (builtins.readFile "${jsonResult}/result.json"));
in { ... }
```

Generating a YAML file is also possible by calling `yq -p json -o yaml ./input.json >./result.yml`

bandithedoge:
> I have a hacky solution to that, but I am not sure how to implement it in nixpkgs, or how to expose it as a function from a flake.
> 
> My idea for parsing YAML consists of two parts:
> 
>     1. A package that takes a YAML file path as an argument, and converts it to JSON, like this:
> 
> 
> ```nix
> { stdenv, pkgs, inputFile, ... }:
> 
> stdenv.mkDerivation {
>   name = "yaml-to-json";
>   dontUnpack = true;
>   dontBuild = true;
> 
>   installPhase = ''
>     mkdir -p "$out"
>     cd "$out"
>     cp "${inputFile}" input.yml
>     ${pkgs.yq-go}/bin/yq -p yaml -o json ./input.yml >./result.json
>   '';
> }
> ```
> 
>     2. A call to the package and parsing the JSON, like this:
> 
> 
> ```nix
> let
> 	jsonResult = pkgs.callPackage ./yaml-to-json.nix { inputFile = ./the/yaml/file.yml; };
> 	parsedYaml = (builtins.fromJSON (builtins.readFile "${jsonResult}/result.json"));
> in { ... }
> ```
> 
> Generating a YAML file is also possible by calling `yq -p json -o yaml ./input.json >./result.yml`

Many projects (including nixpkgs) disallow [IFD](https://nix.dev/manual/nix/2.22/language/import-from-derivation) so your solution isn't exactly suitable.

jim3692:
> > I have a hacky solution to that, but I am not sure how to implement it in nixpkgs, or how to expose it as a function from a flake.
> > My idea for parsing YAML consists of two parts:
> > ```
> > 1. A package that takes a YAML file path as an argument, and converts it to JSON, like this:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```nix
> > { stdenv, pkgs, inputFile, ... }:
> > 
> > stdenv.mkDerivation {
> >   name = "yaml-to-json";
> >   dontUnpack = true;
> >   dontBuild = true;
> > 
> >   installPhase = ''
> >     mkdir -p "$out"
> >     cd "$out"
> >     cp "${inputFile}" input.yml
> >     ${pkgs.yq-go}/bin/yq -p yaml -o json ./input.yml >./result.json
> >   '';
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > 2. A call to the package and parsing the JSON, like this:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```nix
> > let
> > 	jsonResult = pkgs.callPackage ./yaml-to-json.nix { inputFile = ./the/yaml/file.yml; };
> > 	parsedYaml = (builtins.fromJSON (builtins.readFile "${jsonResult}/result.json"));
> > in { ... }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Generating a YAML file is also possible by calling `yq -p json -o yaml ./input.json >./result.yml`
> 
> Many projects (including nixpkgs) disallow [IFD](https://nix.dev/manual/nix/2.22/language/import-from-derivation) so your solution isn't exactly suitable.

Thank you for that note. I didn't realize I can call `derivation {}` directly instead of `callPackage`, which is also IFD, but it can simplify the implementation for my projects.

NikOverflow:
How is this going?

src/libexpr/primops/fromYAML.cc
NaN-git: With the latest commit inputs, which shall throw an error, are tested, too, although some tests are skipped because rapidyaml still parses these inputs. I think that these cases could be fixed upstream in rapidyaml over the next months. At least I can spent some time on it.

src/libexpr/primops/fromYAML.cc
Atry: I think throwing an error might be a good starting point, assuming it would not be a breaking change to turning an error into a non-error in the future.

NaN-git: Timestamps are parsed as strings at the moment. YAML 1.1 has base60 encoded numbers (no longer mentioned in YAML 1.2), which don't have any prefix, but have `:` at specific positions. This is very questionable and it's not supported by rapidyaml anyhow.
In my opinion it makes no sense to parse timestamps, unless the format is well specified. And even this would be difficult without explicit tag. For example `1.23` is parsed as float. With correct rounding this can still be converted to the string `"1.23"`. Of course this property breaks in case of `1.230`. If something would be parsed as timestamp, then it is probably impossible to recover the original representation, unless the representation would be saved. But then equivalent values can have different representations...
I propose to support JSON schema tags only. If another tag is encountered, then an error can be thrown. I'm not sure whether lazy errors are a good solution. The tree generation is not lazy and thus on completion I expect that nix was able to parse the input. Throwing an error immediately is more restrictive than lazy errors. Thus lazy errors could be introduced later, too.

NaN-git:
@roberth @Atry I think that the PR is now ready for merging.

I made some larger changes:
- `builtins.fromYAML` got a second argument. This attribute set can be used to control the behavior of the parser. For example many applications, e.g. Kubernetes, are still supporting YAML 1.1 features not supported by YAML 1.2 and for compatibility I added an option to enable YAML 1.1 booleans.
Futhermore tag resolution can be application specific and in the future it might be possible to extend the tag resolution with nix code.
Backward incompatible changes could be enabled by this argument, too.
- Tag resolution is done according to the YAML 1.2 core schema (`map`, `seq`, `null`, `bool`, `int`, `float` and `str`). Any tag not defined by the YAML 1.2 core schema will result in an error. Any plain scalar, which neither matches the `null`, `bool`, `int` nor `float` regex, is parsed as string.
- Strictness is improved and more error conditions are checked. Also the error messages are improved.
- The tests are improved. Tests of the `yaml-test-suite`, which shall generate an error, are no longer ignored. Sadly, some invalid YAML strings are still parsed successfully by rapidyaml (these tests are skipped), but the situation has improved and it seems that only a handful of cases need to be fixed. The author of rapidyaml seems to be aware about this and there might be fixes in the next few months.
I added tests for codepaths added in `fromYAML.cc`.

## Binary size and performance
-  Static linking of rapidyaml is no problem.
- Additional compile time due to the `form-yaml` feature: approx. +5s out of 6 minutes for `autoreconf`, `configure` and `make` measured on a Ryzen 7 Pro 5850U. Compiling the tests takes longer than compiling the builtin function.
- Binary size overhead (with `nix build`): +110 KiB for the nix binary and less than +650 KiB for libnixexpr and nix.
- Performance: similar to `builtins.fromJSON` and sometimes even faster, but of course with larger memory footprint.

Mastermindaxe:
Just thought about this feature on NixCon and now seeing actual progress being made made me super happy! :tada: Thanks for working on this! I really like the approach of including the version in the function call! Any way I can support this being merged?

NaN-git:
> Any way I can support this being merged?

I talked to @roberth today and he created a milestone: https://github.com/NixOS/nix/milestone/57
This link has to be added to the source code and there is a meson merge conflict again, which needs to be resolved.
Probably someone should review the meson integration, too, because it simply defines `from-yaml` as optional feature, which is slightly different than the `configure`/`make` built.
Hopefully there are no other hurdles left.

gabyx:
This makes me also very happy, as we can move away from our ugly IFD troubles 😀

nakoo:
It seems the PR is ready, but it still hasn't been merged. Is there an underlying reason that hasn't been mentioned here?

gabyx:
@roberth Could you maybe give this another chance, loading yaml is so utterly needed :). Thanks a lot. Even if it becomes experimental feature, I am happy about it.

src/libexpr-tests/yaml.cc
```diff
@@ -0,0 +1,425 @@
...
+        while (ss.peek() != EOF && json.size() - ss.tellg() > 1) {
+            parseJSON(state, ss, refJson);
+            list.emplace_back(refJson);
+            // sanity check: builtins.fromJSON and builtins.fromYAML should return the same result when applied to a
+            // JSON string
+            root.mkString(std::string_view(json.begin() + start, ss.tellg() - start));
+            Value rymlJson = fromYAML(state, root, {});
+            EXPECT_EQ(printValue(state, refJson), printValue(state, rymlJson));
+            start = ss.tellg() + std::streampos(1);
+        }
```
gabyx, line 83: sugg: maybe a comment why this empty catch here

src/libexpr-tests/yaml.cc
```diff
@@ -0,0 +1,425 @@
...
+}
+
+TEST_F(FromYAMLTest, TrueYAML1_1)
+{
+    Value options;
+    auto bindings = state.buildBindings(1);
+    bindings.alloc("useBoolYAML1_1").mkBool(true);
+    options.mkAttrs(bindings);
+
+    Value val = parseYAML("[ true, True, TRUE, y, Y, on, On, ON, yes, Yes, YES ]", options);
```
gabyx, line 356: question: is this really yaml ( then I didnt know) or does Nix parse these as booleans?

src/libexpr/primops/fromYAML.cc
```diff
@@ -0,0 +1,451 @@
...
+
+} /* namespace */
+
+namespace nix {
+
+static RegisterPrimOp primop_fromYAML(
+    {.name = "__fromYAML",
+     .args = {"e", "attrset"},
+     .doc = R"(
+       Convert a YAML 1.2 string *e* to a Nix value, if a conversion is possible.
```
gabyx, line 401: Nice doc!

src/libexpr-tests/yaml.cc
NaN-git: Sadly, `YAML 1.1` specifies that this should be parsed as booleans, e.g. [kubernetes](https://kubernetes.io/docs/concepts/configuration/overview/) relies on this behavior. By default Nix uses `YAML 1.2`, where this is parsed as strings.

NaN-git: It's a misconception that YAML would be simple. Instead the standard is very complex and making `YAML 1.2` a superset of JSON didn't simplify it.

gabyx: Hmm. jeah thats why people probably want toml over yaml sometimes. I am happy when this gets merged, not sure whats taking so long, ok its a new dependency? I guess the ecosystem benefits from such a function a lot. 

Does that yaml parser also support references?

samos667:
I can't wait to test this new experimental features!

Do we know if the merge is planned, and if so, when?

emilazy:
The pain of things like https://github.com/NixOS/nix/pull/13741 and https://github.com/NixOS/nixpkgs/pull/433710 would make me extremely hesitant to bake another complex and underspecified serialization format into the evaluation semantics of the language.

NaN-git:
> The pain of things like #13741 and [NixOS/nixpkgs#433710](https://github.com/NixOS/nixpkgs/pull/433710) would make me extremely hesitant to bake another complex and underspecified serialization format into the evaluation semantics of the language.

I don't think that something like #13741 can happen here. Rapidyaml parses the YAML tree only. The application (nix) is responsible to chose the type of leaf nodes and to parse them.

The type of scalars is chosen according to the rather simple YAML core schema (keys are always parsed as strings because nix cannot handle anything else).

The number parsing has an external dependency, so that some risk remains, but neither the nix parser vendors number parsing.

No YAML serialization is done.

emilazy:
I looked at the changelog for the most recent version of Rapid YAML. It seems like the fix for https://github.com/biojppm/rapidyaml/issues/484 would cause a breaking change in Nix evaluation semantics, right?

NaN-git:
> I looked at the changelog for the most recent version of Rapid YAML. It seems like the fix for [biojppm/rapidyaml#484](https://github.com/biojppm/rapidyaml/issues/484) would cause a breaking change in Nix evaluation semantics, right?

Yes, a bug in the library can affect nix and before this PR is merged I would update the minimum required version (still it would be an experimental feature).
A bug in `nlohmann::json` could affect nix, too, because no specific version is pinned, although `nlohmann::json` is mature and the complexity of the JSON standard is low.

Would your quoted bug affect builds?
Probably not assuming that this feature would be used to parse YAML lock files, which should (TM) require a smaller subset of the YAML standard.
The big elephant in the room is: Which YAML subset is actually required? I don't know.

The concerns have to be tackled before this feature can become non-experimental.

gabyx:
@NaN-git : Isnt it that we can trust this function to be deterministic to the future by locking the dependencies for `nlohman` and `rapidyaml` etc ? But maybe that is not enough: Cause I guess @emilazy probably points out that only a real subset should be enabled (no anchors etc ? no other arcane things?) 
which makes any update of this yaml parsing in the future easier and bugfixes in `rapidyaml` (and deps) do not trickle down into the evaluation and making everything non-determ when `nix` is upgraded etc? Do I understand correctly?

xokdvium:
> Isnt it that we can trust this function to be deterministic to the future by locking the dependencies for nlohman and rapidyaml etc

This would effectively constitute forking/vendoring those libraries and maintaining them ourselves for all eternity to be compatible with newer compilers e.t.c.

It's also doubtful that we'd want to bake problematic and buggy behaviour and declare that to be stable for all eternity.

The problem is that such underspecified and hard to implement stuff as YAML is too much of an API surface to effectively guarantee stability. Doing this would absolutely require adding comprehensive fuzzing and a gigantic corpus of tests to regular regression tests. This is something nix is totally lacking at the moment.

gabyx:
Aha: Then the only way forward is ditch this PR, which is really sad given the amount of work went in and implement a simple yaml parser in Nix which nobody wants to do, resulting in the same problem

x10an14:
> Aha: Then the only way forward is ditch this PR, which is really sad given the amount of work went in and implement a simple yaml parser in Nix which nobody wants to do, resulting in the same problem

If so, you could always roll your own, like I did here: https://git.sr.ht/~x10an14/nix-conf-langs/tree/main/item/languages.nix#L32

gabyx:
@NaN-git Any chance this PR gets merged and experimental status?

roberth:
I can't speak for the nix team, but:

> Doing this would absolutely require adding comprehensive fuzzing and a gigantic corpus of tests to regular regression tests.

This could be done if there's sufficient interest in refining a YAML spec and (possibly) maintaining a fork of `rapidyaml` so that it continues to implement the extra detailed spec.
I wouldn't recommend that the team divides its attention this way, but the :+1: and :heart: suggestions suggest to me that at least the maintenance part could be covered by the community. There may also be interest in this kind of thing outside the Nix/NixOS community. If someone's interested in doing the fuzzing, I believe that could make this happen.

samos667:
~~Tbh I don't understand why this take so long to merge a experimental feature ? Could we consider to address previously mentioned issues in the PR that will make this feature "prod-ready"?~~

~~Just my honnest opinion, but I'll have lose my mind if I was @NaN-git. This PR was initiated three years ago.~~

Edit: Never mind. I thought the author was waiting for the team to merge.

xokdvium:
FWIW, last this was discussed during a team meeting @Mic92 mentioned that even @NaN-git is unsure if this patch *can* be merged.

Ericson2314:
@samos667 I would also like to remind you that every line of code is a liability, and every new feature is a liability. Remember https://www.hyrumslaw.com/ . It might sound silly, but this is the reality of maintaining a widely use piece of software, trying to preserve simplicity, and maintain stability

gabyx:
> > Aha: Then the only way forward is ditch this PR, which is really sad given the amount of work went in and implement a simple yaml parser in Nix which nobody wants to do, resulting in the same problem
> 
> If so, you could always roll your own, like I did here: [git.sr.ht/~x10an14/nix-conf-langs/tree/main/item/languages.nix#L32](https://git.sr.ht/~x10an14/nix-conf-langs/tree/main/item/languages.nix#L32)

I need deserialization.

PR #7346, opened 2022-11-25
Use boost small vectors instead of VLAs in the primops

thufschmitt:
VLAs are a dangerous feature, and their usage triggers an undefined
behavior since theire size can be zero in some cases.
So replace them with `boost::small_vector`s which fit the same goal but
are safer.

It's also incidentally consistently 1% faster on the benchmarks.

The optimal static size of the `small_vectors` still needs to be benchmarked, currently it's only a wild guess (and probably too big).

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -26,7 +26,6 @@
 
 #include <cmath>
 
-
 namespace nix {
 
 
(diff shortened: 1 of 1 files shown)


edolstra (approved):

bb010g:
Subsumed by #9430, which has been merged. This should be closed.

thufschmitt:
Indeed, thanks

PR #7348, opened 2022-11-25
Remove the usage of VLAs in the code

thufschmitt:
Although Variable-length arrays are allowed by the C standard, they are frowned
upon, and aren't valid C++ (`gcc` accepts them a [non-standard extension](https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html)) and `clang` apparently does the same which is why they work here). So better get rid of them altogether.

Depends on <https://github.com/NixOS/nix/pull/7346>.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1766,8 +1766,13 @@ void ExprCall::eval(EvalState & state, Env & env, Value & v)
     Value vFun;
     fun->eval(state, env, vFun);
 
-
-    boost::container::small_vector<Value *, 64> vArgs(args.size());
+    // Empirical arity of Nixpkgs lambdas by regex e.g. ([a-zA-Z]+:(\s|(/\*.*\/)|(#.*\n))*){5}
+    // 2: over 4000
+    // 3: about 300
+    // 4: about 60
+    // 5: under 10
+    // This excluded attrset lambdas (`{...}:`). Contributions of mixed lambdas appears insignificant at ~150 total.
+    boost::container::small_vector<Value *, 4> vArgs(args.size());
     for (size_t i = 0; i < args.size(); ++i)
         vArgs[i] = args[i]->maybeThunk(state, env);
 
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1,6 +1,7 @@
 #include "eval.hh"
 #include "eval-settings.hh"
 #include "hash.hh"
+#include "primops.hh"
 #include "types.hh"
 #include "util.hh"
 #include "store-api.hh"
@@ -2015,7 +2016,7 @@ void ExprConcatStrings::eval(EvalState & state, Env & env, Value & v)
         return result;
     };
 
-    boost::container::small_vector<Value, 16> values(es->size());
+    boost::container::small_vector<Value, conservativeStackReservation> values(es->size());
     Value * vTmpP = values.data();
 
     for (auto & [i_pos, i] : *es) {
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -2015,7 +2015,7 @@ void ExprConcatStrings::eval(EvalState & state, Env & env, Value & v)
         return result;
     };
 
-    boost::container::small_vector<Value, 64> values(es->size());
+    boost::container::small_vector<Value, 16> values(es->size());
     Value * vTmpP = values.data();
 
     for (auto & [i_pos, i] : *es) {
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -723,6 +723,23 @@ void EvalState::addConstant(const std::string & name, Value * v, Constant info)
 }
 
 
+void PrimOp::check()
+{
+    if (arity > maxPrimOpArity) {
+        throw Error("primop arity must not exceed %1%", maxPrimOpArity);
+    }
+}
+
+
+void Value::mkPrimOp(PrimOp * p)
+{
+    p->check();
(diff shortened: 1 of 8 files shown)


src/libexpr/eval.cc
```diff
@@ -1709,7 +1710,8 @@ void EvalState::callFunction(Value & fun, size_t nrArgs, Value * * args, Value &
                 /* We have all the arguments, so call the primop with
                    the previous and new arguments. */
 
-                Value * vArgs[arity];
+                assert(arity < 64);
```
cole-h, line 1713: Should we maybe have a constant for this 64? I see it used a lot here.

That, or a comment explaining where it came from wherever it's used.

src/libexpr/primops.cc
```diff
@@ -2470,8 +2469,8 @@ static void prim_catAttrs(EvalState & state, const PosIdx pos, Value * * args, V
     auto attrName = state.symbols.create(state.forceStringNoCtx(*args[0], pos));
     state.forceList(*args[1], pos);
 
-    Value * res[args[1]->listSize()];
-    unsigned int found = 0;
+    boost::container::small_vector<Value *, 256> res(args[1]->listSize());
```
cole-h, line 2472: Same with this 256?

edolstra:
I think statically sized arrays or `small_vector` should be avoided in functions that can be executed recursively. E.g.
```
                Value * vArgs[64];
```
in `callFunction()` adds 512 bytes of stack space to function calls, which adds up quickly for deeply recursive Nix code.

To be honest I don't really share the concerns over VLAs. We've used them for years without any issues. The zero-length VLA issue seems a bit language lawyering.

tomberek:
TODO (from meeting):
- [ ] check eval impacts
- [ ] update stack overflow test

src/libexpr/eval.cc
roberth: Done. See `maxPrimOpArity`.

src/libexpr/primops.cc
roberth: This is properly arbitrary. I think it's safe to assume that these numbers are guesswork, whether that's actual guesswork or an outdated benchmark.

roberth:
> I think statically sized arrays or `small_vector` should be avoided in functions that can be executed recursively.

I've reduced the numbers significantly. Most significantly, 4 is enough for normal lambda calls. I expect the primops to contribute less to stack depth by number of nested invocations, so these use larger numbers, albeit reduced from what was originally in the PR.
This reduction is fine, because for large numbers the allocation is amortized to insignificance, and in many cases the data size is small anyway.



> TODO (from meeting):
> 
> * [ ]  check eval impacts

This is infinitely faster on expressions where these VLAs caused stack overflows.

> * [ ]  update stack overflow test

I find this somewhat pointless since the code does not allocate variable amounts of stack space anymore. Yes, tests are good, but these would be expensive, especially in terms of maintaining them because it involves developing non-trivial test support code for it. Silvan could attest that this is non-trivial. I'd rather focus that effort on more impactful testing improvements.

roberth:
bug label: Fixes a stack overflow: https://github.com/NixOS/nixpkgs/pull/214021#discussion_r1394896936

src/libexpr/eval.cc
```diff
@@ -1748,11 +1767,16 @@ void ExprCall::eval(EvalState & state, Env & env, Value & v)
     Value vFun;
     fun->eval(state, env, vFun);
 
-    Value * vArgs[args.size()];
+    // Nixpkgs lambdas by regex e.g. ([a-zA-Z]+:(\s|(/\*.*\/)|(#.*\n))*){5}
```
thufschmitt, line 1770:
```suggestion
    // Empirical arity of Nixpkgs lambdas by regex e.g. ([a-zA-Z]+:(\s|(/\*.*\/)|(#.*\n))*){5}
```

(took me a few seconds to understand what that meant)

src/libexpr/eval.hh
```diff
@@ -18,6 +18,12 @@
 
 namespace nix {
 
+/**
+ * We put a limit on primop arity because it lets us use a fixed size array on
+ * the stack. 16 is already an impractical number of arguments. Use an attrset
```
thufschmitt, line 23:
```suggestion
 * the stack. 8 is already an impractical number of arguments. Use an attrset
```

I expect this should match the actual value

thufschmitt (commented):
Thanks for taking over this @roberth !

Looks good to me, although I think this deserves some benchmarks (IIRC, the original version had a small impact depending on the sizes chosen)

roberth:
> deserves some benchmarks

That would be great, but I can only treat this as a bugfix unfortunately.
Based on your earlier measurement I feel confident that this is not a performance regression.

src/libexpr/eval.cc
roberth: Thanks!

src/libexpr/primops.cc
roberth: I did name and document this after all. See `nonRecursiveStackReservation`

thufschmitt:
> Based on your earlier measurement I feel confident that this is not a performance regression.

I wasn't, but I hyperfined it and there's zero visible difference on `nix search nixpkgs firefox`. So I should have :stuck_out_tongue:

edolstra:
This broke the `buildNoGC` job: https://hydra.nixos.org/build/241067941 which is blocking the new release.

roberth:
> This broke the `buildNoGC` job: https://hydra.nixos.org/build/241067941 which is blocking the new release.

- Fixed in #9374

Issue #7347, opened 2022-11-25
Out of bounds read in `void Value::mkString(std::string_view)`


NaN-git:
**Describe the bug**

There is an out of bounds read caused by `strlen` in `void Value::mkString(std::string_view)`, which calls [GC_STRNDUP](https://github.com/NixOS/nix/blob/master/src/libexpr/eval.cc#L76). [This function](https://github.com/ivmai/bdwgc/blob/master/mallocx.c#L594) calls `strlen` on the pointer returned by `std::string_view::data()`, which results in an out of bounds read because `data()` does not return a zero-terminated C-string. The underlying string might also not be zero-terminated, which could cause more trouble.

**Steps To Reproduce**

I noticed the bug when I parsed a very large input string and created short strings from string views of the input string. Nix executed `strlen` for minutes.

**Additional context**

I think that this bug has a kind of denial of service attack surface.

PR #7351, opened 2022-11-25
cleanup eval.hh/eval.cc

NaN-git:
- Remove unused function `makeImmutableStringWithLen` because its interface is not self explanatory and this could have been one of the reasons of  #7347 .
- Remove another undefined function.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -67,20 +67,17 @@ static char * dupString(const char * s)
 
 // When there's no need to write to the string, we can optimize away empty
 // string allocations.
-// This function handles makeImmutableStringWithLen(null, 0) by returning the
-// empty string.
-static const char * makeImmutableStringWithLen(const char * s, size_t size)
+// This function handles makeImmutableString(std::string_view()) by returning
+// the empty string.
+static const char * makeImmutableString(std::string_view s)
 {
+    const size_t size = s.size();
     if (size == 0)
         return "";
     auto t = allocString(size + 1);
-    memcpy(t, s, size);
-    t[size] = 0;
+    memcpy(t, s.data(), size);
+    t[size] = '\0';
     return t;
-}
-
-static inline const char * makeImmutableString(std::string_view s) {
-    return makeImmutableStringWithLen(s.data(), s.size());
 }
 
 
--- a/src/libexpr/eval.hh
+++ b/src/libexpr/eval.hh
@@ -60,7 +60,6 @@ void copyContext(const Value & v, PathSet & context);
 typedef std::map<Path, StorePath> SrcToStore;
 
 
-std::ostream & printValue(const EvalState & state, std::ostream & str, const Value & v);
 std::string printValue(const EvalState & state, const Value & v);
 std::ostream & operator << (std::ostream & os, const ValueType t);
 


src/libexpr/eval.cc
```diff
@@ -65,26 +65,18 @@ static char * dupString(const char * s)
...
     if (size == 0)
         return "";
-#if HAVE_BOEHMGC
-    t = GC_STRNDUP(s, size);
-#else
-    t = strndup(s, size);
-#endif
-    if (!t) throw std::bad_alloc();
+    char * t = allocString(size + 1);
+    memcpy(t, s.data(), size);
```
NaN-git, line 75: `strncpy` could be used instead and an additional check can be added, which ensures that the string_view did contain no zeroes.
The drawback might be reduced performance

src/libexpr/eval.cc
```diff
@@ -65,26 +65,18 @@ static char * dupString(const char * s)
...
 // string allocations.
-// This function handles makeImmutableStringWithLen(null, 0) by returning the
-// empty string.
-static const char * makeImmutableStringWithLen(const char * s, size_t size)
-{
-    char * t;
+// This function handles makeImmutableString(std::string_view()) by returning
+// the empty string.
+static inline const char * makeImmutableString(std::string_view s) {
+    size_t size = s.size();
```
tfc, line 71: why not:

```cpp
const size_t size {s.size()};
auto destination = allocString(size + 1);
memcpy(destination, s.data(), size);
t[size] = '\0';
return t;
```

the size = 0 case is included.

src/libexpr/eval.cc
yorickvP: I worked around this [a while ago](https://github.com/NixOS/nix/pull/7249/commits/742654c5b09f824167b374cc66647e5d0abb1df1) using `strnlen`

src/libexpr/eval.cc
NaN-git: @tfc According to the [C++-standard](https://en.cppreference.com/w/cpp/string/byte/memcpy) the behaviour of `memcpy` is undefined, if the source pointer is an invalid or null pointer, even if count is zero. Thus this codes prevents that potential misuse.
Also the empty string does not use a dynamic memory allocation and each invocation of the function returns the same address. I don't know whether other parts of nix rely on this behaviour...

src/libexpr/eval.cc
NaN-git: @yorickvP `strnlen` and `memcpy` is still slower than a single `memcpy` and probably even slower than `strncpy` because in the normal case `strnlen` would still scan the whole string.

src/libexpr/eval.cc
tfc: you're right, that sounds sensible. thank you for the explanation.

fricklerhandwerk:

@NaN-git the original bug was already fixed in https://github.com/NixOS/nix/pull/7350. We can move this forward if you rework your PR to be pure a refactoring of that fix, i.e., resolve the merge conflict.

NaN-git:
@fricklerhandwerk Thanks for the feedback. I'll create a new PR with these changes and more cleanup in other places.

fricklerhandwerk:
It should be fine to continue with this PR and rename it accordingly, but as you wish.

NaN-git:
@fricklerhandwerk I updated the PR. Now it's pretty trivial.

thufschmitt (approved):
Thanks @NaN-git . Let's merge then!

Issue #6866, opened 2022-08-04
''nix flake check'' requires modules be in a form unamenable to deduplication


charles-dyfis-net:
**Describe the bug**

If two or more modules included into a single system `import` the same dependent module from a flake, these two separate imports will conflict with each other, because `nix flake check` requires that a flake's `nixosModules` be functions, not locations of files from which functions can be imported; and only the latter form can be deduplicated.

**Steps To Reproduce**

See https://gist.github.com/charles-dyfis-net/7b7c68575fbd5fcb68b9160f6f189bc8, setting up a system definition wherein a module is imported twice. In that test case the extra imports are next to each other; in a more realistic case they would be indirect (f/e, requiring two different modules, each of which independently requires the common module).

Within those two, there are two separate tests to be performed:

- Checking if the container build succeeds: `nixos-container create flake-test --flake '.#doubleImportTest'`
- Checking if `nix flake check` succeeds: `nix flake check`

One or the other of these will succeed, but not both:

 - If the following line is uncommented:
   ```
   testModule = ./module.nix;
   ```
  ...then `nix flake check` fails, but the NixOS system build succeeds.

 - If the above line is commented out, and the following line is uncommented in its place:
   ```
   testModule = import ./module.nix;
   ```
   ...then the NixOS system build fails due to the duplicate imports that cannot be deduplicated.

**Expected behavior**

Whichever mechanism for importing modules `nix flake check` encourages should allow the full set of common use cases, including those where a module is imported via multiple paths.

**`nix-env --version` output**

nix-env (Nix) 2.9.1

**Additional context**

N/A.

thufschmitt:
It's not ideal, but you can have

```nix
testModule = { imports = [ ./module.nix ]; };
```

which will work both for `check` and building the system configuration.

roberth:
> It's not ideal, but you can have
> 
> ```nix
> testModule = { imports = [ ./module.nix ]; };
> ```
> 
> which will work both for `check` and building the system configuration.

This form is bad for error messages, because the location information is lost. Errors will be reported in "anonymous module" or worse.

ncfavier:
https://github.com/NixOS/nix/pull/7356

ncfavier:
https://github.com/NixOS/nix/pull/7559

PR #7356, opened 2022-11-27
flake check: allow modules to be paths

ncfavier:
Fixes https://github.com/NixOS/nix/issues/6866

### DIFF
--- a/src/nix/flake.cc
+++ b/src/nix/flake.cc
@@ -380,12 +380,19 @@ struct CmdFlakeCheck : FlakeCommand
 
         auto checkModule = [&](const std::string & attrPath, Value & v, const PosIdx pos) {
             try {
+                Value m;
                 state->forceValue(v, pos);
-                if (v.isLambda()) {
-                    if (!v.lambda.fun->hasFormals() || !v.lambda.fun->formals->ellipsis)
+                if (v.type() == nPath) {
+                    PathSet context;
+                    auto path = state->coerceToPath(pos, v, context);
+                    state->evalFile(path, m);
+                } else
+                    m = v;
+                if (m.isLambda()) {
+                    if (!m.lambda.fun->hasFormals() || !m.lambda.fun->formals->ellipsis)
                         throw Error("module must match an open attribute set ('{ config, ... }')");
-                } else if (v.type() == nAttrs) {
-                    for (auto & attr : *v.attrs)
+                } else if (m.type() == nAttrs) {
+                    for (auto & attr : *m.attrs)
                         try {
                             state->forceValue(*attr.value, attr.pos);
                         } catch (Error & e) {
--- a/tests/flakes/check.sh
+++ b/tests/flakes/check.sh
@@ -78,6 +78,30 @@ EOF
 cat > $flakeDir/flake.nix <<EOF
 {
   outputs = { self }: {
+    nixosModules.default = ./module.nix;
+  };
+}
+EOF
+
+! nix flake check $flakeDir
+
+cat > $flakeDir/module.nix <<EOF
+{
+  a.b.c = 123;
+}
+EOF
+
+nix flake check $flakeDir
+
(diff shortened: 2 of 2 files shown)


src/nix/flake.cc
```diff
@@ -381,6 +381,11 @@ struct CmdFlakeCheck : FlakeCommand
         auto checkModule = [&](const std::string & attrPath, Value & v, const PosIdx pos) {
             try {
                 state->forceValue(v, pos);
+                if (v.type() == nPath) {
+                    PathSet context;
+                    auto path = state->coerceToPath(pos, v, context);
+                    state->evalFile(path, v);
```
roberth, line 387: Since `v` is a reference pointing into the flake, `evalFile` will mutate the flake attribute.
That could lead to surprising behavior down the road.
You could copy `v` into a temporary `Value` after forcing `v` and work with that instead of `v`.

ncfavier: Made a pointer to avoid copying.

ncfavier: Actually that was silly, copying `Value`s is cheap.

edolstra:
Not in favor of this. In general we should avoid magic overloads. Imagine if Nix had a type system - what would the type of `nixosModules.*` be? (It would have to be "a module or a file that when imported evaluates to a module".)

Also, this seems to be entirely to work around a deficiency in the NixOS module system (namely that it uses file names as keys). It would be better to fix this in the module system.

roberth:
> In general we should avoid magic overloads. 

Following this line of reasoning, Nix should be a compiled language, because otherwise, it'd have to use runtime type tags.
The module system, like Nix, is a DSL, which means that it takes a few liberties to make common use cases easier. Furthermore, its implementation is restricted by the primitives that are available in the Nix language.

> Also, this seems to be entirely to work around a deficiency in the NixOS module system (namely that it uses file names as keys). It would be better to fix this in the module system.

The module system has good reason not to hide path imports from its implementation. Removing path imports from the module system would break everything and cause a huge regression in error message quality.

So how? I've maintained the module system for three years. I've dealt with this very part of the system. What you're suggesting is wishful thinking.

I'll be happy to see a replacement for the module system that's _actually_ better, but that is not a project we can start before making Flakes release-worthy.

You can't change the module system by disallowing this, so please don't. You're only making the Flakes user experience worse.

NixOS has been a construct that exists entirely within the language, and that Nix did not have to be aware of. Arguably, `nix flake check` is overreaching by performing NixOS's work, and should instead expose a general eval-time checking facility that not just NixOS but any module implementation can use. This could be supported by `flake-parts` or similar. All Nix would have to do is define an attribute that contains the error messages. Lacking such a feature, and as a fallback for "manually" written flakes, Nix should check what it reasonably can, without reporting false error messages.

Artturin (approved):
`nix flake check` does not work in nixpkgs without this

what @roberth said is true

fricklerhandwerk:

Agreement: Maintaining knowledge about NixOS modules is a layer violation. This is outside of Nix's responsibility.

Decision: Close the pull request.

   - @fricklerhandwerk: confused that `nix flake check` requires Nix to know so much about flake consumers
        - @thufschmitt: agree, it's a layer violation
        - @edolstra: in the absence of a type system that's the best there is right now
        - @roberth: this is what [`flake-parts`](https://flake.parts/) solves in an opt-in way and makes `nix flake check` redundant
    - @edolstra: alternative: remove checking for NixOS modules altogether
    - @fricklerhandwerk: would like to have a more principled approach, like inversion of control where consumers determine the output attributes
        - @edolstra: that was [discussed in the past](https://github.com/NixOS/nix/commit/e5ea01c1a8bbd328dcc576928bf3e4271cb55399). agree in general, but that would require opening the discussion on the general design again. that should be done before stabilising flakes.

</details>

roberth:
> Decision: Close the pull request.

The proposed alternative is to remove this check altogether.

PR #7358, opened 2022-11-28
repl: print a newline on ctrl-D

ncfavier:
It's annoying to get the shell prompt on the same line as the `nix-repl>` prompt when exiting the REPL.


src/libcmd/repl.cc
```diff
@@ -270,6 +270,7 @@ void NixRepl::mainLoop()
             // ctrl-D should exit the debugger.
             state->debugStop = false;
             state->debugQuit = true;
+            std::cout << std::endl;
```
edolstra, line 273:
```suggestion
            logger->cout("");
```
in order to avoid potential problems with the progress bar.

PR #7367, opened 2022-11-29
Bump nixpkgs to 22.11

lheckemann:
Allows removing some patches!


cole-h:
FYI: 22.11 is no longer beta :) https://github.com/NixOS/nixos-org-configurations/pull/228

flake.nix
```diff
@@ -127,13 +127,9 @@
           });
 
         propagatedDeps =
-          [ ((boehmgc.override {
+          [ (boehmgc.override {
               enableLargeConfig = true;
-            }).overrideAttrs(o: {
-              patches = (o.patches or []) ++ [
-                ./boehmgc-coroutine-sp-fallback.diff
```
bb010g, line 134: `./boehmgc-coroutine-sp-fallback.diff` is no longer used & can be deleted.

bb010g (approved):

edolstra:
@lheckemann Looks like this broke some of the VM tests, e.g. https://hydra.nixos.org/build/203389412. Can you have a look at that? I guess we need some more packages in `virtualisation.additionalPaths`.

lheckemann:
https://github.com/NixOS/nix/pull/7570 fixes them

flake.nix
roberth: What? How? Our `BoehmGCStackAllocator` solution does not work correctly with vanilla boehmgc!

Issue #3245, opened 2019-11-26
Nix build processes on non-NixOS have `nogroup` in their supplementary groups


Twey:
Nix on non-NixOS seems to add `nogroup` to the build process's supplementary groups, even though `nixbld` isn't actually a member of `nogroup` in the sandbox.

NixOS (note the lack of `64434(nogroup)` at the end of the `id` line):
```
[twey@uruz:/tmp]$ nix-build -E 'with import <nixpkgs> { }; runCommand "foo" { } "id"'
these derivations will be built:
  /nix/store/cw1c5slf25gjjzpmlwdv44gwdx8kqivc-foo.drv
building '/nix/store/cw1c5slf25gjjzpmlwdv44gwdx8kqivc-foo.drv'...
uid=1000(nixbld) gid=100(nixbld) groups=100(nixbld)
```

Arch:
```
> nix-build -E 'with import <nixpkgs> { }; runCommand "foo" { } "id"'
these derivations will be built:
  /nix/store/scia1pcb1bpwsb026dr9vjzcr4qvc1zr-foo.drv
building '/nix/store/scia1pcb1bpwsb026dr9vjzcr4qvc1zr-foo.drv'...
uid=1000(nixbld) gid=100(nixbld) groups=100(nixbld),65534(nogroup)
```

Ubuntu 18.04:
```
$ nix-build -E 'with import <nixpkgs> { }; runCommand "foo" { } "id"'
these derivations will be built:
  /nix/store/pqjsbxhs0nmgnxfz3l5yn7jdycvvsp4l-foo.drv
building '/nix/store/pqjsbxhs0nmgnxfz3l5yn7jdycvvsp4l-foo.drv'...
uid=1000(nixbld) gid=100(nixbld) groups=100(nixbld),65534(nogroup)
```

We're all using Nix 2.3.1.

This breaks tests for Go (e.g. `go_bootstrap` in nixpkgs): the Go test suite includes a test `TestChown` that attempts to change the group for each supplementary group given by `getgroups`, but this fails for `nogroup`.

zimbatm:
Can you also `cat /etc/group` and `cat /etc/passwd` in the derivation?

Twey:
I edited the description to add some more information about the test failure and sandbox environment.  :)

7c6f434c:
[...]

PR #7374, opened 2022-11-30
Warn when building without nix daemon

yshui:
We need root permission to drop supplementary groups, and if we don't do that, some builds can fail in user namespace, most notably go.

Related: #3245

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -913,8 +913,13 @@ void LocalDerivationGoal::startBuilder()
             int flags = CLONE_NEWPID | CLONE_NEWNS | CLONE_NEWIPC | CLONE_NEWUTS | CLONE_PARENT | SIGCHLD;
             if (privateNetwork)
                 flags |= CLONE_NEWNET;
-            if (usingUserNamespace)
+            if (usingUserNamespace) {
+                if (getgroups(0, NULL) != 0) {
+                    warn("user namespace enabled, but we weren't able to drop supplementary groups; "
+                        "this can break some builds. consider using the nix daemon.");
+                }
                 flags |= CLONE_NEWUSER;
+            }
 
             pid_t child = clone(childEntry, stack + stackSize, flags, this);
             if (child == -1 && errno == EINVAL) {


Ericson2314:

- @ericson2314: PR description perhaps sells it short. If the user enabled sandboxing but is only getting "semi-sandboxing", that deservers a warning.
- Would be good to good to perhaps improve the error message or at least add more comments.
- To discuss more thoroughly later

- @roberth this will run for each build, that would log too often
- @edolstra: why is `getgroups(0, NULL)` sufficient to determine this condition?
  - it's skipped, see above comment
- @edolstra: probably only go is affected (https://github.com/golang/go/issues/42525)
- @theophane, edolstra: wontfix -- don't want to spam non-daemon users with a warning they can't do anything about 
- @ericson2314: Offers to email kernel mailing list and read LWN 
  - Original comment: https://lwn.net/Articles/626665/
  - @tomberek also found: https://lwn.net/Articles/855943/
- shadow?
- @ericson2314: Would be good to document what the builder is doing, what is in scope / out of scope, at some point.

yshui:
> don't want to spam non-daemon users with a warning

what about only print this warning if the build fails?

yshui:
Or, add a way to detect non-daemon builds from nix scripts (is there already a way?), and print a warning from go nix script?

thufschmitt:
> Or, add a way to detect non-daemon builds from nix scripts (is there already a way?), and print a warning from go nix script?

Yes, that sounds sensible since as far as I can tell, this essentially concerns `go` (also, maybe we could fix all the standard go builds like I did for one of them in https://github.com/NixOS/nixpkgs/pull/144398). I assume we can just reuse the same logic that caused the test to fail to detect whether the sandbox has the right groups or not.

fricklerhandwerk:
Triaged in Nix team meeting:

- @Ericson2314 we probably don't need this any more with the [`require-drop-supplementary-groups` setting](https://hydra.nixos.org/build/256426636/download/1/manual/command-ref/conf-file.html?highlight=supplement#conf-require-drop-supplementary-groups)
  - maybe there could be an additional option that issues a warning rather than silently allowing or failing with an error

@yshui feel free to open a new PR if needed

PR #7376, opened 2022-12-01
clarify definition of "installable"

fricklerhandwerk:
the term was hard to discover, as its definition and explanation were in
  a very long document lacking an overview section.
  search did not help because it occurs so often.

  - clarify wording in the definition
  - add an overview of installable types
  - add "installable" to glossary
  - link to definition from occurrences of the term

the first commit makes descriptions of each installable type its own subsection.
it makes for a messy diff but does not change contents. read the second commit to see the actual changes.

### DIFF
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -193,6 +193,11 @@
     A symlink to the current *user environment* of a user, e.g.,
     `/nix/var/nix/profiles/default`.
 
+  - [installable]{#gloss-installable}\
+    Something that can be realised in the Nix store.
+
+    See [installables](./command-ref/new-cli/nix.md#installables) for [`nix` commands](./command-ref/new-cli/nix.md) (experimental) for details.
+
   - [NAR]{#gloss-nar}\
     A *N*ix *AR*chive. This is a serialisation of a path in the Nix
     store. It can contain regular files, directories and symbolic
--- a/src/libcmd/command.hh
+++ b/src/libcmd/command.hh
@@ -22,7 +22,7 @@ static constexpr Command::Category catSecondary = 100;
 static constexpr Command::Category catUtility = 101;
 static constexpr Command::Category catNixInstallation = 102;
 
-static constexpr auto installablesCategory = "Options that change the interpretation of installables";
+static constexpr auto installablesCategory = "Options that change the interpretation of [installables](@docroot@/command-ref/new-cli/nix.md#installables)";
 
 struct NixMultiCommand : virtual MultiCommand, virtual Command
 {
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -153,7 +153,7 @@ SourceExprCommand::SourceExprCommand()
         .longName = "file",
         .shortName = 'f',
         .description =
-            "Interpret installables as attribute paths relative to the Nix expression stored in *file*. "
+            "Interpret [*installables*](@docroot@/command-ref/new-cli/nix.md#installables) as attribute paths relative to the Nix expression stored in *file*. "
             "If *file* is the character -, then a Nix expression will be read from standard input. "
             "Implies `--impure`.",
         .category = installablesCategory,
@@ -164,7 +164,7 @@ SourceExprCommand::SourceExprCommand()
 
     addFlag({
         .longName = "expr",
-        .description = "Interpret installables as attribute paths relative to the Nix expression *expr*.",
+        .description = "Interpret [*installables*](@docroot@/command-ref/new-cli/nix.md#installables) as attribute paths relative to the Nix expression *expr*.",
         .category = installablesCategory,
         .labels = {"expr"},
         .handler = {&expr}
--- a/src/nix/nix.md
+++ b/src/nix/nix.md
@@ -55,7 +55,6 @@ The following types of installable are supported by most commands:
 
 - [Flake output attribute](#flake-output-attribute)
 - [Store path](#store-path)
-- [Store derivation](#store-derivation)
 - [Nix file](#nix-file), optionally qualified by an attribute path
 - [Nix expression](#nix-expression), optionally qualified by an attribute path
 
@@ -126,22 +125,17 @@ Example: `/nix/store/v5sv61sszx301i0x6xysaqzla09nksnd-hello-2.10`
 
 These are paths inside the Nix store, or symlinks that resolve to a path in the Nix store.
 
-### Store derivation
+A [store derivation] is also addressed by store path.
 
 Example: `/nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv`
 
-By default, if you pass a [store derivation] path to a `nix` subcommand other than [`show-derivation`](./nix3-show-derivation.md), the command will operate on the [output path]s of the derivation.
+If you want to refer to an output path of that store derivation, add the output name preceded by a caret (`^`).
 
-[output path]: ../../glossary.md#gloss-output-path
+Example: `/nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv^out`
 
-For example, [`nix path-info`](./nix3-path-info.md) prints information about the output paths:
+All outputs can be referred to at once with the special syntax `^*`.
 
-```console
-# nix path-info --json /nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv
-[{"path":"/nix/store/v5sv61sszx301i0x6xysaqzla09nksnd-hello-2.10",…}]
-```
-
-If you want to operate on the store derivation itself, pass the `--derivation` flag.
+Example: `/nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv^*`
 
 ### Nix file
 
--- a/src/nix/nix.md
+++ b/src/nix/nix.md
@@ -48,22 +48,26 @@ manual](https://nixos.org/manual/nix/stable/).
 
 # Installables
 
-Many `nix` subcommands operate on one or more *installables*. These are
-command line arguments that represent something that can be built in
-the Nix store.
+Many `nix` subcommands operate on one or more *installables*.
+These are command line arguments that represent something that can be realised in the Nix store.
 
-For most commands, if no installable is specified, the default is `.`,
-i.e. Nix will operate on the default flake output attribute of the
-flake in the current directory.
+The following types of installable are supported by most commands:
 
-Here are the recognised types of installables:
+- [Flake output attribute](#flake-output-attribute)
+- [Store path](#store-path)
+- [Store derivation](#store-derivation)
+- [Nix file](#nix-file), optionally qualified by an attribute path
+- [Nix expression](#nix-expression), optionally qualified by an attribute path
 
-## Flake output attributes
+For most commands, if no installable is specified, `.` as assumed.
+That is, Nix will operate on the default flake output attribute of the flake in the current directory.
+
+### Flake output attribute
 
 Example: `nixpkgs#hello`
 
 These have the form *flakeref*[`#`*attrpath*], where *flakeref* is a
-flake reference and *attrpath* is an optional attribute path. For
+[flake reference](./nix3-flake.md#flake-references) and *attrpath* is an optional attribute path. For
 more information on flakes, see [the `nix flake` manual
 page](./nix3-flake.md).  Flake references are most commonly a flake
 identifier in the flake registry (e.g. `nixpkgs`), or a raw path
@@ -116,46 +120,46 @@ subcommands, these are `packages.`*system*,
 attributes `packages.x86_64-linux.hello`,
 `legacyPackages.x86_64-linux.hello` and `hello`.
 
-## Store paths
+### Store path
 
 Example: `/nix/store/v5sv61sszx301i0x6xysaqzla09nksnd-hello-2.10`
 
-These are paths inside the Nix store, or symlinks that resolve to a
-path in the Nix store.
+These are paths inside the Nix store, or symlinks that resolve to a path in the Nix store.
 
-## Store derivations
+### Store derivation
 
 Example: `/nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv`
 
-By default, if you pass a [store derivation] path to a `nix` subcommand, the command will operate on the [output path]s of the derivation.
+By default, if you pass a [store derivation] path to a `nix` subcommand other than [`show-derivation`](./nix3-show-derivation.md), the command will operate on the [output path]s of the derivation.
 
 [output path]: ../../glossary.md#gloss-output-path
 
-For example, `nix path-info` prints information about the output paths:
+For example, [`nix path-info`](./nix3-path-info.md) prints information about the output paths:
 
 ```console
 # nix path-info --json /nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv
 [{"path":"/nix/store/v5sv61sszx301i0x6xysaqzla09nksnd-hello-2.10",…}]
 ```
 
-If you want to operate on the store derivation itself, pass the
-`--derivation` flag.
+If you want to operate on the store derivation itself, pass the `--derivation` flag.
 
-## Nix attributes
+### Nix file
 
 Example: `--file /path/to/nixpkgs hello`
 
-When the `-f` / `--file` *path* option is given, installables are
-interpreted as attribute paths referencing a value returned by
-evaluating the Nix file *path*.
+When the option `-f` / `--file` *path* \[*attrpath*...\] is given, installables are interpreted as the value of the expression in the Nix file at *path*.
+If attribute paths are provided, commands will operate on the corresponding values accessible at these paths.
+The Nix expression in that file, or any selected attribute, must evaluate to a derivation.
 
-## Nix expressions
+### Nix expression
 
-Example: `--expr '(import <nixpkgs> {}).hello.overrideDerivation (prev: { name = "my-hello"; })'`.
+Example: `--expr 'import <nixpkgs> {}' hello`
 
-When the `--expr` option is given, all installables are interpreted
-as Nix expressions. You may need to specify `--impure` if the
-expression references impure inputs (such as `<nixpkgs>`).
+When the option `--expr` *expression* \[*attrpath*...\] is given, installables are interpreted as the value of the of the Nix expression.
+If attribute paths are provided, commands will operate on the corresponding values accessible at these paths.
+The Nix expression, or any selected attribute, must evaluate to a derivation.
+
+You may need to specify `--impure` if the expression references impure inputs (such as `<nixpkgs>`).
 
 ## Derivation output selection
 
--- a/src/nix/nix.md
+++ b/src/nix/nix.md
@@ -50,100 +50,112 @@ manual](https://nixos.org/manual/nix/stable/).
 
 Many `nix` subcommands operate on one or more *installables*. These are
 command line arguments that represent something that can be built in
-the Nix store. Here are the recognised types of installables:
-
-* **Flake output attributes**: `nixpkgs#hello`
-
-  These have the form *flakeref*[`#`*attrpath*], where *flakeref* is a
-  flake reference and *attrpath* is an optional attribute path. For
-  more information on flakes, see [the `nix flake` manual
-  page](./nix3-flake.md).  Flake references are most commonly a flake
-  identifier in the flake registry (e.g. `nixpkgs`), or a raw path
-  (e.g. `/path/to/my-flake` or `.` or `../foo`), or a full URL
-  (e.g. `github:nixos/nixpkgs` or `path:.`)
-
-  When the flake reference is a raw path (a path without any URL
-  scheme), it is interpreted as a `path:` or `git+file:` url in the following
-  way:
-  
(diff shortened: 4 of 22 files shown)


doc/manual/src/glossary.md
```diff
@@ -1,5 +1,15 @@
 # Glossary
 
+  - [installable]{#gloss-installable}\
+    Argument type that can be passed to [`nix-build`](./command-ref/nix-build.md) or [`nix build`](./command-ref/new-cli/nix3-build.md) (experimental):
```
edolstra, line 4:
```suggestion
    Argument type that can be passed to [`nix build`](./command-ref/new-cli/nix3-build.md):
```

doc/manual/src/glossary.md
```diff
@@ -1,5 +1,15 @@
...
+
+    - path to Nix expression
+    - store derivation path
+    - [store path]
+    - [flake reference](./command-ref/new-cli/nix3-flake.md#flake-references) (only for [`nix build`](./command-ref/new-cli/nix3-build.md), experimental)
```
edolstra, line 9:
```suggestion
    - [flake reference](./command-ref/new-cli/nix3-flake.md#flake-references)
```

doc/manual/src/glossary.md
edolstra: The entry should try to give some intuition about what the intent of "installable" is. E.g. from the `nix` manpage:

> Many nix subcommands operate on one or more *installables*. These are command line arguments that represent something that can be built in the Nix store.

src/libcmd/command.hh
```diff
@@ -22,7 +22,7 @@ static constexpr Command::Category catSecondary = 100;
 static constexpr Command::Category catUtility = 101;
 static constexpr Command::Category catNixInstallation = 102;
 
-static constexpr auto installablesCategory = "Options that change the interpretation of installables";
+static constexpr auto installablesCategory = "Options that change the interpretation of [installables](../../glossary.md#gloss-installable)";
```
edolstra, line 25: Instead of linking to the glossary, it would be better to link to the "Installables" section of the `nix` manpage, which is much more complete.

doc/manual/src/glossary.md
Ericson2314: We should probably also rename them (in a follow-up PR) as they really are not about installing at all.

Once we have https://github.com/NixOS/nix/pull/4543 we can document this better as derived paths are the *ur installable*, everything else is just a more elaborate way of specifying a derived path (with some commands doing things with those extra details).

src/libcmd/command.hh
Ericson2314: Shouldn't everything in the man page be also in the manual? I would think the man page info can go in the manual, and the glossary can be a 1-sentance thing that points to it.

I am not sure which we should link, but that is a general issue when the same info (once we do the above!) is available two different ways.

fricklerhandwerk: @edolstra the idea was to link to an at-a-glance overview of what the term means. The full page requires scrolling just to see all the options.

But sure, I can also add an overview on the full page.

doc/manual/src/glossary.md
```diff
@@ -1,5 +1,15 @@
 # Glossary
 
+  - [installable]{#gloss-installable}\
+    Argument type that can be passed to [`nix-build`](./command-ref/nix-build.md) or [`nix build`](./command-ref/new-cli/nix3-build.md) (experimental):
+
+    - path to Nix expression
```
amjoseph-nixpkgs, line 6: Does this mean

- filesystem path of a `.nix` file (e.g. `~/nixpkgs/default.nix`) or directory containing `default.nix` or
- attribute path (e.g. `pkgsStatic.busybox`)

I guess both are correct, depending on which command-line flags (`-f`, `-A`) and cli (`nix-build`, `nix build`) you use.

Aside: it has always seemed weird to me that `nix` needs flags in order to refer to an attrpath within a `.nix` file instead of being able to write `nix build ~/nixpkgs/default.nix/pkgsStatic.busybox`.

doc/manual/src/glossary.md
fricklerhandwerk: @Edolstra why does the concept not apply to `nix-build`?

doc/manual/src/glossary.md
fricklerhandwerk: The specifics of how things may be weird are not a concern right here, I'm just trying to state the facts.

You see, writing this down can help determining what's weird and which ways there may be to change that. But let's first focus on on the part of writing down clearly and correctly.

Ericson2314: Agree completely, I *don't* want to see that fixes in this docs PR!

doc/manual/src/glossary.md
edolstra: I mean, it's just not a concept in the old CLI (in the source code, error messages, docs...). Obviously the old CLI has a similar overloading of command line arguments, but different (e.g. `nix-build foo.nix -A bar` vs `nix build -f foo.nix bar` - so we can't say that `foo.nix` is an installable).

src/libcmd/command.hh
edolstra: @Ericson2314 The man page is part of the manual (https://nixos.org/manual/nix/stable/command-ref/new-cli/nix.html#installables) so it's fine to link to that.

@fricklerhandwerk The at-a-glance overview should ideally (also) be part of the installables section of the `nix` man page, since that's what most users will see (e.g. when they run `nix help`).

doc/manual/src/glossary.md
Ericson2314: @fricklerhandwerk yeah fwiw the code in the implemementation called `Installable` isn't used by the old CLI either. The old CLI is quite ad-hoc and doesn't have these reusable concepts in its implementation.

src/libcmd/command.hh
Ericson2314: Sounds good. I was sort of thinking the regular manual info shouldn't need to refer to the man pages (even though the regular manual does contain the man page) but then it turned out that the updated sections were all man pages anyways!

doc/manual/src/glossary.md
amjoseph-nixpkgs: Sorry, I should've resisted the urge to bikeshed there.  I've crossed out the distracting part of my comment.

amjoseph-nixpkgs (commented):
I think that trying to mash together "anything accepted by any nix command is an installable and vice versa" is not only confusing but also incorrect.  See below, a lot of nix commands which "take installables" reject some kinds of installables.

We shouldn't try to hide the difference between an outpath and its deriver from users.

This "everything is an installable and there are nothing but installables" papers over the many-to-one relationship between derivers and store objects.  That will come back to bite us later, just like the "referring to a derivation always means referring to its realisation, except when it doesn't" problem that led to the `--derivation` flag did.

fricklerhandwerk:
@amjoseph-nixpkgs I think you're right, and all this hints at is that the CLI has some design issues. We should try to be as precise as possible, but I feel it's only worthwhile up to a point. Saying "it doesn't actually work for all the commands" has to suffice for now. Indeed we'd have to untangle the underlying issues first, which here is the overloading of argument types, but this is not the PR and, in my opinion, not yet the time to do so.

I'll fix up what seems reasonable and would then try to get closer to merging.

src/nix/nix.md
```diff
@@ -48,11 +48,18 @@ manual](https://nixos.org/manual/nix/stable/).
...
+These are command line arguments that represent something that can be built in the Nix store.
 
-* **Flake output attributes**
+The following types of installable are supported:
+
+- [Flake output attribute](#flake-output-attribute)
+- [Store path](#store-path)
+- [Store derivation](#store-derivation)
+- [Attribute path](#attribute-path)
+- [Nix expression](#nix-expression)
```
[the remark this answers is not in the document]
fricklerhandwerk: Watered it down with "most commands".

fricklerhandwerk:
@amjoseph-nixpkgs updated according to your suggestions. @Nixos/nix-team ready for review.

src/nix/nix.md
```diff
@@ -48,102 +48,116 @@ manual](https://nixos.org/manual/nix/stable/).
...
-  subcommands, these are `packages.`*system*,
-  `legacyPackages.*system*` and the empty prefix. Thus, on
-  `x86_64-linux` `nix build nixpkgs#hello` will try to build the
-  attributes `packages.x86_64-linux.hello`,
-  `legacyPackages.x86_64-linux.hello` and `hello`.
+For most commands, if no installable is specified, `.` as assumed.
+That is, Nix will operate on the default flake output attribute of the flake in the current directory.
 
-* **Store paths**: `/nix/store/v5sv61sszx301i0x6xysaqzla09nksnd-hello-2.10`
+### Flake output attribute
```
amjoseph-nixpkgs, line 65:
```suggestion
### Evaluatable Installables

Of the five types of installables, the first (flake output attributes) and last (nix expressions) are considered to be *evaluatable installables*.  Some commands accept only these two specific types of installables.

### Flake output attribute
```

src/nix/search.md
```diff
@@ -62,10 +62,10 @@ R""(
 
 # Description
 
-`nix search` searches *installable* (which must be evaluatable, e.g. a
-flake) for packages whose name or description matches all of the
+`nix search` searches [*installable*](./nix.md#installables) (which can be evaluated, that is, a
+flake or Nix expression, but not a store path or store derivation path) for packages whose name or description matches all of the
```
amjoseph-nixpkgs, line 66:
```suggestion
`nix search *evaluatable-installable* *regex*` searches its first argument (an [*evaluatable installable*](./nix.md#installables) for attributes whose values are installables whose name or description matches the regular expression *regex*.
```

src/nix/show-derivation.md
```diff
@@ -39,7 +39,7 @@ R""(
 # Description
 
 This command prints on standard output a JSON representation of the
-[store derivation]s to which *installables* evaluate. Store derivations
+[store derivation]s to which [*installables*](./nix.md#installables) evaluate. Store derivations
```
amjoseph-nixpkgs, line 42:
```suggestion
store derivation.  Store derivations
```

amjoseph-nixpkgs: `nix show-deriation` doesn't work on outpaths.

amjoseph-nixpkgs (commented):
This is a significant improvement, especially the identification of an installable which "can be evaluated, that is, a flake or Nix expression, but not a store path or store derivation path" as an important subcategory of installables.

Most of my comments below push towards defining a new term *evaluatable installable* for "flakes or expressions" as a subset of installables.  You're welcome to pick a different term if you like.  But I think it's worth having a name for it, because then you can unambiguously specify the allowed arguments for every command except `nix show-derivation`.

`nix show-derivation` is the oddball because it takes any installable *except* an outpath.  It's also the only command that doesn't automatically "dereference" a derivation to its output by default, or have a `--derivation` flag to prevent this from happening.

src/nix/nix.md
```diff
@@ -48,102 +48,116 @@ manual](https://nixos.org/manual/nix/stable/).
...
-    of the supplied path that contains a `flake.nix` file (within the same file-system).
-    If no such directory exists, then Nix will error-out.
-    
-    For example, if `/foo/bar/flake.nix` exists, then `/foo/bar/baz/` will resolve to
-   `path:/foo/bar`
+- [Flake output attribute](#flake-output-attribute)
+- [Store path](#store-path)
+- [Store derivation](#store-derivation)
+- [Nix file](#nix-file), optionally qualified by an attribute path
+- [Nix expression](#nix-expression), optionally qualified by an attribute path
```
amjoseph-nixpkgs, line 60:
```suggestion
- [Nix expression](#nix-expression), optionally qualified by an attribute path, if it evaluates to a derivation
```

This is actually a pretty important distinction.  This is not an installable:

```nix
3 + 4 * (throw "cats")
```

This *is* an installable:

```nix
builtins.derivation { name = "frogs"; builder = ./ribbit.sh; }
```

src/nix/nix.md
amjoseph-nixpkgs: You might consider re-ordering the list so you can say "the first two" instead of "first and last".

src/nix/nix.md
```diff
@@ -48,102 +48,116 @@ manual](https://nixos.org/manual/nix/stable/).
...
 
-  - If the supplied path is not a git repository, then the url will have the form
-    `path:FLAKE_DIR_PATH` where `FLAKE_DIR_PATH` is the closest parent
-    of the supplied path that contains a `flake.nix` file (within the same file-system).
-    If no such directory exists, then Nix will error-out.
-    
-    For example, if `/foo/bar/flake.nix` exists, then `/foo/bar/baz/` will resolve to
-   `path:/foo/bar`
+- [Flake output attribute](#flake-output-attribute)
+- [Store path](#store-path)
```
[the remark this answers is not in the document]
fricklerhandwerk: @Ericson2314 would it be correct to say this is a *derived path*? IIUC this is your proposed terminology for store paths that are created by realising derivations.

src/nix/run.md
```diff
@@ -35,7 +35,7 @@ R""(
 
 # Description
 
-`nix run` builds and runs *installable*, which must evaluate to an
+`nix run` builds and runs [*installable*](./nix.md#installables), which must evaluate to an
 *app* or a regular Nix derivation.
```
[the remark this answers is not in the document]
fricklerhandwerk: Nope. It is some notion defined in the source, but I didn't understand what it means when I last looked at it. Not touching it here, this is out of scope.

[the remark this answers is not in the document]
fricklerhandwerk: I know it would be nice changing it while we're at it, but I'd like avoid scope creep to reduce the time in flight of PRs, therefore I'd ask you to make a separate PR to address this you want it fixed.

src/nix/search.md
fricklerhandwerk: I agree this stuff needs to be more precise, and your suggestion is definitely an improvement. But this is not what this PR is about, so it would be great if you could make a follow-up.

Also I'm not sure why that aspect `nix search` behavior is even described here and not on its respective page. I also noticed this while rewording the description of the different installable types. Especially documentation for `--file` and `--expr`  should rather go into the option documentation, but this needs to be a different PR.

src/nix/nix.md
fricklerhandwerk: This is correct, but too long for the overview. Noted this in the individual descriptions.

src/nix/show-derivation.md
fricklerhandwerk: Applied the related suggestion.

src/nix/nix.md
fricklerhandwerk: I agree there needs to be a clarification, but don't want to go down that path, because I think the notion of installable is confusing enough as it is, and that is because the relation between derivation, store derivation, store path, output path, and the process how one is transformed into the other, is more important to untangle.

Related: https://github.com/NixOS/nix/issues/6507

amjoseph-nixpkgs:
So it turns out that `nix copy foo.drv` doesn't do the same thing as `nix copy $(nix-store -r foo.drv)`.

In other words, "operations on derivations operate on their realisation except when `--derivation` is used" isn't correct.  There is a third behavior.

Ericson2314:
@amjoseph-nixpkgs the one difference for `nix-copy` I know of is CA derivation realizations. Is that what you hit?

amjoseph-nixpkgs:
> @amjoseph-nixpkgs the one difference for `nix-copy` I know of is CA derivation realizations. Is that what you hit?

I don't think so; I don't have them in my `experimental-features` at the moment.

I think the problem is actually in the `$(nix-store -r)`.   Here is an example that doesn't need `nix copy`, both in prose and bash:

### Prose


In prose: if your local store has neither particular outpath (say, `/nix/store/outpath`) nor its deriver (say, `/nix/store/deriver.drv`), then running

```
nix-store -r /nix/store/outpath
```

will copy the outpath, **but not its deriver,** to your local machine.

If you instead run

```
nix-store -r /nix/store/deriver.drv
```

then **both the deriver *and* the outpath** will be copied to your local machine.

This contradicts the "using a derivation as an installable produces exactly the same result as using its realisation as an installable".

### Bash

```
$ nix-store -r /nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools
this path will be fetched (122.16 MiB download, 122.16 MiB unpacked):
  /nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools
copying path '/nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools' from 'http://substituter'...
/nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools

$ ls $(nix-store -qd /nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools)
ls: cannot access '/nix/store/iqwd30cadrajfampslvykwbv8wps1jsx-bootstrap-tools.drv': No such file or directory

$ nix-store -r $(nix-store -qd /nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools)
don't know how to build these paths:
  /nix/store/iqwd30cadrajfampslvykwbv8wps1jsx-bootstrap-tools.drv
copying path '/nix/store/iqwd30cadrajfampslvykwbv8wps1jsx-bootstrap-tools.drv' from 'http://substituter'...
/nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools

$ ls $(nix-store -qd /nix/store/dhkv3d1a4hjkfabhi91m6kzpp9x40vpp-bootstrap-tools)
/nix/store/iqwd30cadrajfampslvykwbv8wps1jsx-bootstrap-tools.drv
```

src/libcmd/command.hh
```diff
@@ -22,7 +22,7 @@ static constexpr Command::Category catSecondary = 100;
 static constexpr Command::Category catUtility = 101;
 static constexpr Command::Category catNixInstallation = 102;
 
-static constexpr auto installablesCategory = "Options that change the interpretation of installables";
+static constexpr auto installablesCategory = "Options that change the interpretation of [installables](./nix.md#installables)";
```
fricklerhandwerk, line 25:
```suggestion
static constexpr auto installablesCategory = "Options that change the interpretation of [installables](@docroot@/nix.md#installables)";
```

src/libcmd/installables.cc
```diff
@@ -161,15 +161,15 @@ SourceExprCommand::SourceExprCommand(bool supportReadOnlyMode)
...
+        .description = "Interpret *installables* as attribute paths relative to the Nix expression *expr*.",
         .category = installablesCategory,
         .labels = {"expr"},
         .handler = {&expr}
     });
 
     addFlag({
         .longName = "derivation",
-        .description = "Operate on the [store derivation](../../glossary.md#gloss-store-derivation) rather than its outputs.",
+        .description = "If *installables* is a [store derivation](../../glossary.md#gloss-store-derivation), operate on that rather than its outputs.",
```
fricklerhandwerk, line 172:
```suggestion
        .description = "If *installables* is a [store derivation](@docroot@/glossary.md#gloss-store-derivation), operate on that rather than its outputs.",
```

src/nix/nix.md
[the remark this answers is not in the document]
fricklerhandwerk: @amjoseph-nixpkgs I'm hesitant to add this immediately, because we don't have neither of the terminology established. Rather than blocking this PR on another dependency that requires lots of scrutiny and possibly discussion, I'd keep it that way even if it's still slightly confusing, and postpone fixing that to a follow-up. You're very welcome to make PRs in that direction which I will gladly review.

fricklerhandwerk:
@amjoseph-nixpkgs I also encountered this inconsistency and think it's a valid point we should address with a design decision. Please open an issue.

@thufschmitt While it's still not perfect, I think it's ready to merge. Could you please take look?

src/libcmd/command.hh
```diff
@@ -22,7 +22,7 @@ static constexpr Command::Category catSecondary = 100;
 static constexpr Command::Category catUtility = 101;
 static constexpr Command::Category catNixInstallation = 102;
 
-static constexpr auto installablesCategory = "Options that change the interpretation of installables";
+static constexpr auto installablesCategory = "Options that change the interpretation of [installables](@docroot@/command-ref/new-clii/nix.md#installables)";
```
fricklerhandwerk, line 25:
```suggestion
static constexpr auto installablesCategory = "Options that change the interpretation of [installables](@docroot@/command-ref/new-cli/nix.md#installables)";
```

Ericson2314 (approved):
This is looking very nice!

fricklerhandwerk:
@thufschmitt please take a look.

src/nix/nix.md
```diff
@@ -48,102 +48,118 @@ manual](https://nixos.org/manual/nix/stable/).
...
+
+Example: `/nix/store/v5sv61sszx301i0x6xysaqzla09nksnd-hello-2.10`
+
+These are paths inside the Nix store, or symlinks that resolve to a path in the Nix store.
+
+### Store derivation
+
+Example: `/nix/store/p7gp6lxdg32h4ka1q398wd9r2zkbbz2v-hello-2.10.drv`
+
+By default, if you pass a [store derivation] path to a `nix` subcommand other than [`show-derivation`](./nix3-show-derivation.md), the command will operate on the [output path]s of the derivation.
```
Ericson2314, line 133: No longer true! :)

fricklerhandwerk: How time flies...

thufschmitt (approved):

PR #7378, opened 2022-12-01
doc rendering: add functions to scope explicitly

fricklerhandwerk:
this especially helps beginners with code readability, since the origin
of names is always immediately visible.

no changes to semantics, only renamed and reordered, and made the imported names explicit.


Ericson2314:
We briefly discussed this in the Nix team meeting today, and **assigned** it to @roberth for final review.

Ericson2314 (approved):

PR #7381, opened 2022-12-01
doc: listToAttrs: document repeated keys

aakropotkin:
Previously I had thought this was implemented as the first example below ( I was wrong ):
```
# WRONG
listToAttrs = attrs:
  builtins.foldl' ( acc: key: acc // { ${key} = attrs.${key}; } ) {}
                  ( builtins.attrNames attrs );
# RIGHT
listToAttrs = attrs:
  builtins.foldl' ( acc: key: { ${key} = attrs.${key}; } // acc ) {}
                  ( builtins.attrNames attrs );
```
I was browsing the `primops.cc` sources and bumped into an inline comment that mentioned that it was actually the opposite behavior: `listToAttrs` preserves the "first instance" of a key.

This change adds a note to the docstring, and extends the existing example to make the behavior clear to users.

### DIFF
--- a/src/libexpr/primops.cc
+++ b/src/libexpr/primops.cc
@@ -2420,8 +2420,12 @@ static RegisterPrimOp primop_listToAttrs({
       Construct a set from a list specifying the names and values of each
       attribute. Each element of the list should be a set consisting of a
       string-valued attribute `name` specifying the name of the attribute,
-      and an attribute `value` specifying its value. In case of duplicate
-      occurrences of the same name, the first takes precedence. Example:
+      and an attribute `value` specifying its value.
+
+      In case of duplicate occurrences of the same name, the first
+      takes precedence.
+
+      Example:
 
       ```nix
       builtins.listToAttrs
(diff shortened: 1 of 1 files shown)


src/libexpr/primops.cc
```diff
@@ -2420,12 +2420,14 @@ static RegisterPrimOp primop_listToAttrs({
       Construct a set from a list specifying the names and values of each
       attribute. Each element of the list should be a set consisting of a
       string-valued attribute `name` specifying the name of the attribute,
-      and an attribute `value` specifying its value. Example:
+      and an attribute `value` specifying its value. In case of duplicate
+      occurrences of the same name, the first takes precedence. Example:
```
fricklerhandwerk, line 2424: Please add whitespace, if text is too dense it's hard to find the bits of information by scanning.
```suggestion
      and an attribute `value` specifying its value.
      
      In case of duplicate occurrences of the same name, the first takes precedence.
      
      Example:
```

fricklerhandwerk (approved):
Great catch, thank you!

src/libexpr/primops.cc
aakropotkin: Will do.

PR #7382, opened 2022-12-01
move documentation on `auto-allocate-uids` to options docs

fricklerhandwerk:
this is where it belongs and can be found together with the other
options.

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -284,7 +284,21 @@ public:
         )"};
 
     Setting<bool> autoAllocateUids{this, false, "auto-allocate-uids",
-        "Whether to allocate UIDs for builders automatically."};
+        R"(
+          Whether to allocate UIDs for builders automatically.
+
+          These UIDs are allocated starting at 872415232 (0x34000000) on Linux and 56930 on macOS.
+
+          > **Warning**
+          > This is an experimental feature.
+
+          To enable it, add the following to [`nix.conf`](#):
+
+          ```
+          extra-experimental-features = auto-allocate-uids
+          auto-allocate-uids = true
+          ```
+        )"};
 
     Setting<uint32_t> startId{this,
         #if __linux__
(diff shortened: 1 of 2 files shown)


src/libstore/globals.hh
```diff
@@ -308,11 +322,21 @@ public:
     Setting<bool> useCgroups{
         this, false, "use-cgroups",
         R"(
-          Whether to execute builds inside cgroups. Cgroups are
-          enabled automatically for derivations that require the
-          `uid-range` system feature.
-        )"
-    };
+          Whether to execute builds inside cgroups.
+          Only on Linux with systemd.
```
edolstra, line 326:
```suggestion
          This is only supported on Linux.
```

src/libstore/globals.hh
```diff
@@ -308,11 +322,21 @@ public:
...
+
+          cgroups are required and enabled automatically for derivations that require the `uid-range` system feature.
```
edolstra, line 328:
```suggestion
          Cgroups are required and enabled automatically for derivations 
          that require the `uid-range` system feature.
```

src/libstore/globals.hh
```diff
@@ -284,7 +284,21 @@ public:
         )"};
 
     Setting<bool> autoAllocateUids{this, false, "auto-allocate-uids",
-        "Whether to allocate UIDs for builders automatically."};
+        R"(
+          Whether to allocate UIDs for builders automatically.
```
edolstra, line 288:
```suggestion
          Whether to select UIDs for builds automatically, instead of using the
          users in `build-users-group`.
```

src/libstore/globals.hh
```diff
@@ -284,7 +284,21 @@ public:
...
+
+          These UIDs are allocated starting at 872415232 (0x34000000) on Linux and 56930 on macOS.
```
edolstra, line 290:
```suggestion
          UIDs are allocated starting at 872415232 (0x34000000) on Linux and 56930 on macOS.
```

PR #7388, opened 2022-12-01
allow setting cgroup memory limits via both nix.conf and pre-build hooks

cleverca22:
allows setting resource limits with `--option use-cgroups true --option memory-max $((1024*1024*1024*5))`
allows a pre-build hook to set limits:
```bash
#!/bin/sh

if [[ -d $NIX_CGROUP ]]; then
  echo $((1024*1024*1024*1)) > $NIX_CGROUP/memory.max
fi
```
such a pre-build hook could fork out a daemon, that watches `memory.events` and dynamically raises `memory.high` while co-operating with a resource manager

note, currently requires that nix-daemon be ran with an extra dir in its cgroup path, such as `/system.slice/nix-daemon.service/actual-daemon`

### DIFF
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -154,6 +154,7 @@ void LocalDerivationGoal::killSandbox(bool getStats)
         if (getStats) {
             buildResult.cpuUser = stats.cpuUser;
             buildResult.cpuSystem = stats.cpuSystem;
+            buildResult.memoryHigh = stats.memoryHigh;
         }
         #else
         abort();
@@ -415,6 +416,7 @@ void LocalDerivationGoal::startBuilder()
             throw Error("cannot determine cgroup name from /proc/self/cgroup");
 
         auto ourCgroupPath = canonPath("/sys/fs/cgroup/" + ourCgroup);
+        ourCgroupPath = dirOf(ourCgroupPath);
 
         if (!pathExists(ourCgroupPath))
             throw Error("expected cgroup directory '%s'", ourCgroupPath);
@@ -720,6 +722,15 @@ void LocalDerivationGoal::startBuilder()
             chownToBuilder(*cgroup);
             chownToBuilder(*cgroup + "/cgroup.procs");
             chownToBuilder(*cgroup + "/cgroup.threads");
+            auto parentCgroup = dirOf(*cgroup);
+            writeFile(parentCgroup + "/cgroup.subtree_control", "+memory");
+            writeFile(*cgroup + "/memory.oom.group", "1");
+            if (settings.memoryHigh) {
+              writeFile(*cgroup + "/memory.high", fmt("%d", settings.memoryHigh));
+            }
+            if (settings.memoryMax) {
+              writeFile(*cgroup + "/memory.max", fmt("%d", settings.memoryMax));
+            }
             //chownToBuilder(*cgroup + "/cgroup.subtree_control");
         }
 
@@ -751,7 +762,14 @@ void LocalDerivationGoal::startBuilder()
             stExtraChrootDirs
         };
         auto state = stBegin;
-        auto lines = runProgram(settings.preBuildHook, false, args);
+        auto hookEnv = getEnv();
+        if (cgroup) hookEnv["NIX_CGROUP"] = *cgroup;
+        auto res = runProgram(RunOptions {.program = settings.preBuildHook, .searchPath = false, .args = args, .environment = hookEnv, .input = {} });
+
+        if (!statusOk(res.first))
+            throw ExecError(res.first, "program '%1%' %2%", settings.preBuildHook, statusToString(res.first));
+
+        auto lines = res.second;
         auto lastPos = std::string::size_type{0};
         for (auto nlPos = lines.find('\n'); nlPos != std::string::npos;
                 nlPos = lines.find('\n', lastPos))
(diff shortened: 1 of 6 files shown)


src/libstore/globals.hh
```diff
@@ -313,6 +313,22 @@ public:
...
+          sets the cgroup memory.high limit, causing the build to throttle if exceeded
+          depends on use-cgroups=true
+        )"
+    };
+
+    Setting<uint64_t> memoryMax{
+        this, 0, "memory-max",
+        R"(
+          sets the cgroup memory.max limit, causing the build to be killed if exceeded
+          depends on use-cgroups=true
```
SuperSandro2000, line 329: I am missing some documentation here in what relation to each other you should set memory-high and memory-max. I would assume something like memory-high 3GB and memory-max 4GB?

Artturin: https://facebookmicrosites.github.io/cgroup2/docs/memory-controller.html

cleverca22: depends on how much a specific derivation needs, the goal is to keep track of how much memory each derivation is using, and not schedule too many on one machine
then use cgroups to enforce that it stays within the reserved amounts

thufschmitt:
Could it make sense to have a more general `cgroup-config` parameter accepting a list of key-value pairs?

src/libstore/build/local-derivation-goal.cc
```diff
@@ -415,6 +416,7 @@ void LocalDerivationGoal::startBuilder()
             throw Error("cannot determine cgroup name from /proc/self/cgroup");
 
         auto ourCgroupPath = canonPath("/sys/fs/cgroup/" + ourCgroup);
+        ourCgroupPath = dirOf(ourCgroupPath);
```
edolstra, line 419: This is very confusing naming. Doing `dirOf` means it *isn't* our cgroup, but the parent's. And it's not clear that we want to be putting stuff in the parent cgroup, since it causes nix to break out of whatever resource controls are in place for its own cgroup. It especially wouldn't be nice if Nix builds escape from `system.slice/nix-daemon.service`.

src/libstore/cgroup.cc
```diff
@@ -113,6 +113,11 @@ static CgroupStats destroyCgroup(const Path & cgroup, bool returnStats)
             }
         }
 
+        auto memoryhighPath = cgroup + "/memory.high";
+        if (pathExists(memoryhighPath)) {
+          stats.memoryHigh = string2Int<uint64_t>(trim(readFile(memoryhighPath)));
```
edolstra, line 118: Indentation is off here.

src/libstore/globals.hh
```diff
@@ -313,6 +313,22 @@ public:
           `uid-range` system feature.
         )"
     };
+
+    Setting<uint64_t> memoryHigh{
+        this, 0, "memory-high",
+        R"(
+          sets the cgroup memory.high limit, causing the build to throttle if exceeded
+          depends on use-cgroups=true
```
edolstra, line 321: This needs proper capitalization/spelling/Markdown formatting.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -720,6 +722,15 @@ void LocalDerivationGoal::startBuilder()
             chownToBuilder(*cgroup);
             chownToBuilder(*cgroup + "/cgroup.procs");
             chownToBuilder(*cgroup + "/cgroup.threads");
+            auto parentCgroup = dirOf(*cgroup);
+            writeFile(parentCgroup + "/cgroup.subtree_control", "+memory");
```
edolstra, line 726: This will fail if the cgroup does not have `memory` in `cgroup.controllers`, or if it's not a leaf cgroup or one that only has `pids` in its children (see the "no internal processes" rule in `cgroups(7)`). In particular, this may cause non-root chroot builds to fail (i.e. `--store /path/to/my/nix` as a non-root user) since we can't count on `memory` being enabled.

It's not clear what the best solution is, but some options:

* When `nix-daemon` starts, it can create a sub-cgroup (e.g. `system.slice/nix-daemon.service/manager`) for itself and move into it, and create sub-cgroups for the builds in the original cgroup (e.g. `system.slice/nix-daemon.service/<build>`). That way the original cgroup doesn't violate the "no internal processes" rule.
* That doesn't work for the non-root case, since the original cgroup will likely have other processes in it (e.g. `user.slice/user-1000.slice/user@1000.service/app.slice/app-org.kde.konsole-7edbbf6bbf864ef9a259f86f31b0fd90.scope`). So we can't easily convert it into a cgroup that doesn't violate the "no internal processes" rule. I guess we could search upwards for the nearest parent that has the `memory` controller enabled and has no internal processes (e.g. `/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/app.slice` in this example) but that's a bit naughty...
* Allow the parent cgroup for builds to be specified in `nix.conf`.

Whatever the solution, Nix shouldn't fail if writing to `cgroup.subtree_control` fails.

edolstra (changes requested):

src/libstore/build/local-derivation-goal.cc
cleverca22: i had modified things so the daemon runs in `system.slice/nix-daemon.service/actual-daemon`

the problem is that you cant certain things on `system.slice/nix-daemon.service` if it contains any process, everything must be in a child c-group

i'm not sure what the best solution to that is, and this currently needs an extra script around `nix-daemon` to move it to a child cgroup

tomberek:

PR #7389, opened 2022-12-02
do not use shell prompt character, add highlighting

fricklerhandwerk:
follow-up on #7251, merged on behalf of @NixOS/documentation-team.
@Ericson2314 correctly pointed out that it was by far incomplete.

Screenshot 2022-12-02 at 03 56 21

this helps distinguishing inputs and outputs clearly, both in the web
rendering as well as in the markdown source.

input-output-pairs do not render as nicely on the terminal, because
`lowdown` does not understand indented blocks. therefore those appear to
be demoted to regular text paragraphs there.


fricklerhandwerk:
@Ericson2314 after tinkering with this setup for a while, I fully agree that Sphinx+MyST is the highly superior alternative. It  would clearly be a Pareto improvement over mdBook (except probably closure size and build time, but I doubt that really matters for the manual).

Ericson2314 (commented):
Yay we are consistent again!

edolstra:
I'm really not in favor of this. It's a pretty marginal improvement in the output, and it makes the markdown source a mess, e.g. I find this hard to read/write:

<pre>
  An example output of this configuration:

  ```console
  nix-build ./test.nix -A unstable
  ```

        /nix/store/ch6llwpr2h8c3jmnf3f2ghkhx59aa97f-unstable.drv
      building '/nix/store/ch6llwpr2h8c3jmnf3f2ghkhx59aa97f-unstable.drv' (round 1/2)...
      building '/nix/store/ch6llwpr2h8c3jmnf3f2ghkhx59aa97f-unstable.drv' (round 2/2)...
      output '/nix/store/6xg356v9gl03hpbbg8gws77n19qanh02-unstable' of '/nix/store/ch6llwpr2h8c3jmnf3f2ghkhx59aa97f-unstable.drv' differs from '/nix/store/6xg356v9gl03hpbbg8gws77n19qanh02-unstable.check' from previous round
      /nix/store/6xg356v9gl03hpbbg8gws77n19qanh02-unstable
</pre>

A better approach would be to have some syntax highlighting that e.g. colors lines that start with `#` differently.

domenkozar:
I think this is a step backwards for two reasons:

1) It moves away from the real usage of the terminal, so my expectation to see the terminal prompt isn't fulfilled and as I read I can get confused if it's input or output that I'm seeing.

2) Ideally we'd use tool like [tesh](https://github.com/OceanSprint/tesh) that we built at the oceansprint to make sure these examples are tested. The tool relies on the prompt to understand what to feed into the shell for testing.

edolstra:
Since highlight.js already detects prompts in `console` code blocks, we just need to add some CSS like this to `custom.css`:
```
.hljs-meta.prompt_ {
    color: grey;
    font-style: italic;
    user-select: none;
}
```

Ericson2314:
For the record, that comes from https://github.com/highlightjs/highlight.js/commit/6b8c831f00c4e87ecd2189ebbd0bb3bbdde66c02

PR #7391, opened 2022-12-02
feat: add pre-http-request-hooks config entry and handle it

olebedev:
## Motivation 

Currently, the model of how resources are being pulled/pushed via HTTP is not supposed to use short-living tokens for private endpoints. Only a `netrc` file with basic access auth credentials is supported. At the company, I am working for we would like to have this access model extended in order to make use of short-living tokens, issues per URL, or even more, one-off disposable tokens per request to make the secure posture stronger.

The problem with that approach is that it's file-based, ie. a physical file needs to be stored on the machine and it serves for the basic access auth (`Authorization: Basic <credentials>`) so it's pretty limited and the last but not least, it is hard to implement rotation of such files on developer's machine.

## In this PR

This PR introduces another global option in `nix.conf` that allows defining tuples of host and a path to the program to execute, in order to obtain additional HTTP headers for a particular URL and action, that is being requested by the Nix process. 

The hooks are defined as a map with a hook per host, for example:

```
pre-http-request-hooks = nix.private.big-corp.com=/nix/store/g47h1ll8vmc4ip5lm8lmnhibhbxbn7j9-pre-http-request-hook
```

Where the `/nix/store/g47h1ll8vmc4ip5lm8lmnhibhbxbn7j9-pre-http-request-hook` is a bash executable:
```shell
#!/nix/store/y3hpcrb8gnx6b6pb5b6s247rs13i1rgv-bash-5.1-p16/bin/bash

set -euo pipefail

url="${1}"  # the current URL that the hook is being executed for
verb="${2}" # the verb, can be `upload` or `download`

token="$(/nix/store/g47h1ll8vmc4ip5lm8lmnhibhbxbn7j9-cf-creds-helper "${url} "${verb}")"

echo "Cf-Access-Token: ${token}"
echo "Origin: nix.private.big-corp.com"
```

These hooks will be executed for each HTTP request that matches the `host`. By that, it's become possible to have short-living access tokens per each URL,action tuple which is a more secure access model and it allows managing short-living tokens with ease.

This PR is a proof of concept that does work for us. I am happy to discuss possible options and implement something that can address the original issue.

### DIFF
--- a/src/libstore/filetransfer.cc
+++ b/src/libstore/filetransfer.cc
@@ -116,6 +116,26 @@ struct curlFileTransfer : public FileTransfer
             for (auto it = request.headers.begin(); it != request.headers.end(); ++it){
                 requestHeaders = curl_slist_append(requestHeaders, fmt("%s: %s", it->first, it->second).c_str());
             }
+
+            for (auto &[host, prog] : fileTransferSettings.preHTTPRequestHooks.get())
+            {
+                if (hasPrefix(request.uri, fmt("https://%s/", host)) || hasPrefix(request.uri, fmt("http://%s/", host)))
+                {
+                    auto verb = request.data ? "upload" : "download";
+                    debug("pre-http-request-hooks: '%s' for '%s' (%d)", prog, request.uri, verb);
+                    auto lines = runProgram(prog, false, Strings({ request.uri, verb }));
+                    auto lastPos = std::string::size_type{0};
+                    for (auto nlPos = lines.find('\n'); nlPos != std::string::npos; nlPos = lines.find('\n', lastPos))
+                    {
+                        auto line = lines.substr(lastPos, nlPos - lastPos);
+                        lastPos = nlPos + 1;
+                        auto hn = line.substr(0, line.find(':'));
+                        debug("add header: '%s: *****' to '%s' (%d)", hn, request.uri, verb);
+                        requestHeaders = curl_slist_append(requestHeaders, line.c_str());
+                    }
+                    break;
+                }
+            }
         }
 
         ~TransferItem()
--- a/src/libstore/filetransfer.hh
+++ b/src/libstore/filetransfer.hh
@@ -26,6 +26,35 @@ struct FileTransferSettings : Config
         )",
         {"binary-caches-parallel-connections"}};
 
+    Setting<StringMap> preHTTPRequestHooks{this, {}, "pre-http-request-hooks",
+        R"(
+          Optional. A whitespace-separated list of tuples (`host=<path to a program>`) where 
+          the program can set extra URL-specific HTTP headers. This is used for URLs
+          that can't be accessed publically.
+
+          The hook is passed wth the resource URL and the verb (`upload` or `download`).
+          It can then add https headers to the HTTP request by send them to stdout. 
+          
+          When using the nix-daemon, the daemon executes the hook as `root`.
+          If the nix-daemon is not involved, the hook runs as the user
+          executing the nix-build.
+
+          Example:
+              `#!/usr/bin/env bash
+
+              source /my/lib.sh
+
+              url=$1
+              verb=$2
+
+              host=$(parse_host ${url})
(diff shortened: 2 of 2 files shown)


olebedev:
Hi @fricklerhandwerk,

Would you mind having a look, please? I would love to hear general feedback and maybe you have some ongoing conversations around auth problems for Nix when run inside companies with strict security requirements.

We were discussing this area with @domenkozar the other day, so maybe this (or a solution like this) would also benefit Cashix's customers.

fricklerhandwerk:
@olebedev Thanks! I will drop it into the triage pipeline for the Nix team. Not sure why you pinged me though. I have no authority or qualification to decide anything, and only happen to publish meeting notes on behalf of the Nix team and sometimes have opinions on documentation.

Please note that we're [still figuring out how to prioritise PRs and issues](https://discourse.nixos.org/t/2022-12-02-nix-team-meeting-minutes-13/23731#discussion-3), and cannot state an ETA when yours will be handled.

edolstra:
It's probably good to enumerate the intended use cases first. A hook is powerful but is also hard to use. Plus it's pretty ad hoc. It might be better to have an option to specify headers directly.

olebedev:
> It's probably good to enumerate the intended use cases first. A hook is powerful but is also hard to use. Plus it's pretty ad hoc. 

@edolstra, I agree that it's a bit an ad-hoc and probably not what we want to see in the Nix longer term. I am open to discussion around possible options and would love to follow your guidance since I am relatively new to the Nix development practices. Do you think we can make use of it affordable for a large audience? What do you think is hard within this approach, from your perspective?

Our workflow requirements stand as:
- The user from their laptop/workstation has to be able to access in-house protected HTTP resources, such as binary caches and mirrors through the Nix processes. This comes from the fact that in-house-made Nix packages contain intellectual property and we can afford them to leak to the public.
- The protected resources accept access tokens that need to be defined as HTTP headers, standard and/or non-standard ones. That is, `Authorisation: Bearer <token>`, `Cf-auth-token: <token>`, and like.
- Because of the company's security posture, such tokens need to be rotated frequently, in most cases, TTL is about an hour.
- Because of the company's security posture, tokens need to be issued per host
- We can't use files on the user's file system storing any credential there, because we have to maintain a state of such files, which is non-trivial if you don't control the whole system
- Make the access model work for background processes, running on the machine.
- Keep nix performance sane, that is, understand slowdowns that will be introduced by an auth mechanism and if it affects the overall developer experience. This should not be a problem given Nix will be looking at protected resources only for the Nix packages that don't exist in the Nix store.

Optional requirements:
- Make only the Nix process access such protected HTTP resources
- Issue access tokens per URL & HTTP Method tuple

---

> It might be better to have an option to specify headers directly.

Initially, I've been trying to think about the problem in that way but it turned out that static headers specifying don't address the dynamic nature of the workflow above.

In our code base, Nix is being used in all possible ways, we install packages via `nix-env`, we do `nix-build` execution and we use `nix-shell` shebang a lot, even stack them in a sequence, calling it from, for example, `IDE -> Nix shell -> Bazel -> Nix shell` and so on. Sometimes there is no access to Nix CLI directly since it's wrapped in multiple execution contexts.

@edolstra, what kind of idea do you have in mind that would get us into the state where we can specify headers without involving the hook mechanism, what do you think?

olebedev:
> Please note that we're [still figuring out how to prioritise PRs and issues](https://discourse.nixos.org/t/2022-12-02-nix-team-meeting-minutes-13/23731#discussion-3), and cannot state an ETA when yours will be handled.

@fricklerhandwerk, that is fine :-) I think I can wait for some time, however, if possible, I would love to get feedback about the current implementation and get some time to make changes if required. We already using a patched version of Nix but would prefer to make the change upstream, generic enough, and well documented.

fricklerhandwerk:

The general sentiment is that the feature has low priority at the moment. We will take time to decide how to predictably deal with such pull requests in the future, given maintainer time is limited.

olebedev:
Closing for now, feel free to reopen if needed.

PR #7392, opened 2022-12-02
Revert "docs: drop shell prompt character for consistency"

edolstra:
This reverts commit c068cce107b94f23c5288521770b2c807729a052 because it makes the docs *less* consistent.

See comments https://github.com/NixOS/nix/pull/7389#issuecomment-1335070679, https://github.com/NixOS/nix/pull/7251#issuecomment-1334322973.

### DIFF
--- a/doc/manual/src/installation/installing-binary.md
+++ b/doc/manual/src/installation/installing-binary.md
@@ -3,7 +3,7 @@
 The easiest way to install Nix is to run the following command:
 
 ```console
-sh <(curl -L https://nixos.org/nix/install)
+$ sh <(curl -L https://nixos.org/nix/install)
 ```
 
 This will run the installer interactively (causing it to explain what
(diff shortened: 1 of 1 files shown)


Ericson2314 (approved):
Yay we are consistent again!

That is also what I said in https://github.com/NixOS/nix/pull/7389 too :) but it's true. We can make it consistent either way

infinisil:
Can we also add your suggestion from https://github.com/NixOS/nix/pull/7389#issuecomment-1335109486 then?

fricklerhandwerk (approved):

infinisil:
@Ericson2314 https://github.com/NixOS/nix/blob/master/doc/manual/custom.css looks like it might be it

Ericson2314:
Great! Thanks for sleuthing it out!

PR #7394, opened 2022-12-02
Lower verbosity of exceptions in getMaxCPU()

### LINKED ISSUE #7268: nix tries to read `cpu.max` which is non-readable
**Describe the bug**

In the [nix-on-droid](https://github.com/t184256/nix-on-droid) project, we spin up a proot environment for non-privileged users on android devices. Unfortunately, nix tries to read `cpu.max` from the path mounted with `cgroup2` which happens to be non-readable by the unprivileged user: 

```
❯ grep cgroup2 /proc/mounts
none /dev/cg2_bpf cgroup2 rw,nosuid,nodev,noexec,relatime 0 0
❯ ls /dev
ls: cannot open directory '/dev': Permission denied
```

This produces errors like these on every invocation of any nix command
```
error (ignored): error: getting status of /dev/cg2_bpf/cpu.max: Permission denied
```

Everything works perfectly fine, but this error is always showing up.

Why is nix reading `cpu.max` and why is it throwing an error but resuming nonetheless?

**Steps To Reproduce**

1. Download `nix-on-droid` app on an android device
2. Start app and bootstrap with default settings
3. Multiple errors occur

**Expected behavior**

There should not be any errors which are ignored by default.

nix-env (Nix) 2.8.1

**Additional context**

Related issue: https://github.com/t184256/nix-on-droid/issues/213
Source code reference: https://github.com/NixOS/nix/blob/9dadb5481cdf848c409e6a7e4953acf4861923a5/src/libutil/util.cc#L726-L772

edolstra:
Fixes #7268.

Also refactors cgroup handling a bit.


src/libutil/util.cc
```diff
@@ -727,45 +727,22 @@ unsigned int getMaxCPU()
...
-
-                if (quota != "max")
-                    return std::ceil(std::stoi(quota) / std::stof(period));
-            }
-        }
-    } catch (Error &) { ignoreException(); }
+        auto cgroupFS = getCgroupFS();
+        if (!cgroupFS) return 0;
+
+        auto cgroups = getCgroups("/proc/self/cgroupp");
```
Mic92, line 733: Has this code been tested? It should be `/proc/self/cgroup`.

PR #7395, opened 2022-12-02
Change "while evaluating <fun>" to "while *calling*" in trace

Ericson2314:
The old way was not correct.

Here is an example:
```
 $ nix-instantiate --eval --expr 'let x = a: throw "asdf"; in x 1' --show-trace
error: asdf

       … while evaluating 'x'

       at «string»:1:9:

            1| let x = a: throw "asdf"; in x 1
             |         ^

       … from call site

       at «string»:1:29:

            1| let x = a: throw "asdf"; in x 1
             |                             ^
```
and yet also:
```
 $ nix-instantiate --eval --expr 'let x = a: throw "asdf"; in x' --show-trace
<LAMBDA>
```

Here is the thing: in both cases we are evaluating `x`!

Nix is a higher-order languages, and functions are a sort of value. When we write `x = a: ...`, `a: ...` is the expression that `x` is being defined to be, and that is already a value. Therefore, we should *never* get an trace that says "while evaluating `x`", because evaluating `a: ...` is *trival* and nothing happens during it!

What is actually happening here is we are applying `x` and evaluating its *body* with arguments substituted for parameters. I think the simplest way to say is just "while *calling* `x`", and so that is what I changed it to.

### DIFF
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1646,7 +1646,7 @@ void EvalState::callFunction(Value & fun, size_t nrArgs, Value * * args, Value &
                 auto dts = debugRepl
                     ? makeDebugTraceStacker(
                         *this, *lambda.body, env2, positions[lambda.pos],
-                        "while evaluating %s",
+                        "while calling %s",
                         lambda.name
                         ? concatStrings("'", symbols[lambda.name], "'")
                         : "anonymous lambda")
@@ -1655,7 +1655,7 @@ void EvalState::callFunction(Value & fun, size_t nrArgs, Value * * args, Value &
                 lambda.body->eval(*this, env2, vCur);
             } catch (Error & e) {
                 if (loggerSettings.showTrace.get()) {
-                    addErrorTrace(e, lambda.pos, "while evaluating %s",
+                    addErrorTrace(e, lambda.pos, "while calling %s",
                         (lambda.name
                             ? concatStrings("'", symbols[lambda.name], "'")
                             : "anonymous lambda"));


Ericson2314:
I have in fact been legitimately confused by this in the past, but another way to think of this is that we are making is better for pedants like me :), and no worse for non-pedants, so the change is Pareto-optimal.

roberth (approved):

Ericson2314:
Another example is that today we have
```shell-session
 $ nix-instantiate --eval --expr 'let x = a: let y = throw "asdf"; in y; in x 1' --show-tra
ce
error: asdf

       … while evaluating 'x'

       at «string»:1:9:

            1| let x = a: let y = throw "asdf"; in y; in x 1
             |         ^

       … from call site

       at «string»:1:43:

            1| let x = a: let y = throw "asdf"; in y; in x 1
             |                                           ^
```
but it would be nice if we could also do extra verbosity and print:
```shell-session
 $ nix-instantiate --eval --expr 'let x = a: let y = throw "asdf"; in y; in x 1' --show-tra
ce
error: asdf

       … while evaluating 'y'

       at «string»:1:20:

            1| let x = a: let y = throw "asdf"; in y; in x 1
             |                     ^

       … from use at

       at «string»:1:37:

            1| let x = a: let y = throw "asdf"; in y; in x 1
             |                                     ^

       … while calling 'x'

       at «string»:1:9:

            1| let x = a: let y = throw "asdf"; in y; in x 1
             |         ^

       … from call site

       at «string»:1:43:

            1| let x = a: let y = throw "asdf"; in y; in x 1
             |                                           ^
```
By adding trace frames to track mere variable usage, we provide a *valid* use for "while evaluating <name>`: `x` is being called, but `y` is merely being evaluated.

Similarly today
```shell-session
 $ nix-instantiate --eval --expr 'let x = throw "asdf"; in x 1' --show-trace
error: asdf
```
*doesn't* get any trace, because the error happens we call any function (while we are trying to get a function to call) but with more verbosity we can do
```shell-session
 $ nix-instantiate --eval --expr 'let x = throw "asdf"; in x 1' --show-trace
error: asdf

       … while evaluating 'x'

       at «string»:1:9:

            1| let x = throw "asdf"; in x1
             |         ^

       … from use at

       at «string»:1:17:

            1| let x = throw "asdf"; in x 1
             |                          ^
```
we before we can evaluate function application this is just a regular use of `x`. So we say "evaluating `x`" because we are erroring before we reach the point of doing anything particular with `x` such as calling it.

PR #7398, opened 2022-12-03
.github/workflows/assign-reviewer: switch to pull_request_target

zowoq:


ncfavier:
I don't have strong opinions about this, but see https://github.com/actions/first-interaction/issues/10#issuecomment-1278884148

zowoq:
> I don't have strong opinions about this, but see [actions/first-interaction#10 (comment)](https://github.com/actions/first-interaction/issues/10#issuecomment-1278884148)

comment != review

ncfavier:
I'd expect both to fall under the `pull-requests` scope?

Anyway, I was rather pointing at the potential security risk with `pull_request_target`. Maybe we should try with

```yaml
permissions:
  contents: read
  pull-requests: write
```

first, see if that works?

zowoq:
>  Maybe we should try with ...

Please do.

fricklerhandwerk:
Please add a descriptive commit message that does not state what the code change itself already informs us about, i.e. *why* that change was made.

zowoq:
Please

Issue #5557, opened 2021-11-14
`unsafeGetAttrPos` is not unsafe. `tryGetAttrPos`?


roberth:
**Describe the bug**

Its unsafety is an exaggeration. It cannot guarantee to return a (good?) position for any arbitrary attribute, but that does not make it unsafe.
It could even be said that `unsafeGetAttrPos` is safer than `readFile` because it doesn't even throw an exception, but I'd consider them to be in the same class of reliability.

**Steps To Reproduce**

1. Read https://github.com/NixOS/nixpkgs/pull/114647#issuecomment-968069395
2. See that people don't want to use it because it's "unsafe"

**Expected behavior**

`unsafeGetAttrPos` causes a crash. I mean, that's what people would expect, but we ought to just fix the name by introducing an alias for the function. What about `tryGetAttrPos`?

**`nix-env --version` output**

**Additional context**

This is probably "cultural heritage" from Haskell's `unsafePerformIO`: a function that isn't inherently unsafe, but allows things to be done that shouldn't be possible within the language semantics. It can cause issues through interactions with other parts of the language that rely on referential transparency for memory safety, so the `unsafe` qualifier is appropriate there. However, it is quite safe to use in predictable circumstances, which I believe leads people to use the same qualifier for functions that are actually safe.

tomberek:
Closing as stale. Re-open if needed.

eclairevoyant:
Stale bots are a scourge on software dev.

xokdvium:
> Its unsafety is an exaggeration. It cannot guarantee to return a (good?) position for any arbitrary attribute, but that does not make it unsafe.

I disagree somewhat. It is unsafe in the sense that position information is not stable and one needs to think really hard when using this primop. Otherwise you risk leaking its result into the hashes and that's going to end very badly - as we saw current position computation doesn't work at all when multi-byte code points are involved and produces wrong results.

Keeping it as "unsafe" and explicitly for not in use in derivations is good IMO.

Issue #5897, opened 2022-01-11
Syntax that evaluates to the path to the current source file


lilyball:
**Is your feature request related to a problem? Please describe.**
I sometimes need to include a reference to the current file (e.g. when defining an inline NixOS module I want to say `_file = ./thisfile.nix` to ensure the option definition has location information). I can do this by writing `./filename.nix`, but this means the code now has a dependency on the current filename, making it harder to move code around.

**Describe the solution you'd like**
The simplest solution I can think of is to add a new builtin `builtins.currentFile` that evaluates to the path to the current file. I don't know enough about how builtins work though so I don't know if lazy evaluation is an issue here.

**Describe alternatives you've considered**
No obvious syntax for this comes to mind besides the builtin.

thufschmitt:
That already indirectly exists with `builtins.unsafeGetAttrPos`:

```console
$ cat > foo.nix <<EOF
> (builtins.unsafeGetAttrPos "x" { x = 1; }).file                                                                                             
> EOF                              
$ nix eval --file foo.nix
"/tmp/tmp.8s0eR6X2ee/foo.nix"
```

(note that it’s “unsafe” as it breaks referential transparency because it means that the same expression at two different places will have a different value)

lilyball:
Oof, that's an awkward solution (though it does demonstrate that this is perfectly achievable with builtins).

As for referential transparency, the expression `./.` already has a different value depending on which file you put that expression in. Here I figured that the `unsafe` in `unsafeGetAttrPos` was just "we don't guarantee that this actually has the expected value, it's a best-effort thing".

lilyball:
@pennae Oh that's neat. I notice that there's only one usage of this in all of nixpkgs (to calculate a module `_file` attribute, which is precisely my use-case), which strikes me as a little curious. It's also not documented outside of the release notes. But it certainly sounds like precisely what I want.

tomberek:
Closing as stale. Re-open if needed.

PR #7399, opened 2022-12-03
Rename `unsafeGetAttrPos` to `tryGetAttrPos`

ncfavier:
"Unsafe" looks scarier than it should, when the only unsafe thing about this is that it violates referential transparency (but so does `./.`).

Fixes https://github.com/NixOS/nix/issues/5557, see also https://github.com/NixOS/nix/issues/5897#issuecomment-1010125472

cc @lilyball @roberth

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -52,3 +52,6 @@
 
 * `nix build --json` now prints some statistics about top-level
   derivations, such as CPU statistics when cgroups are enabled.
+
+* `builtins.unsafeGetAttrPos` was renamed to `builtins.tryGetAttrPos`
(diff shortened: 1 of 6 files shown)


edolstra:
"Unsafe" conveys that you should generally avoid this function, since it's not guaranteed to provide reproducible results. So I think the current name is fine. Even if the new name is better, it's not really worth the hassle of breaking forward compatibility.

willbush:
> breaking forward compatibility.

Randomly found this from searching why this builtin is unsafe. I was just wondering if this PR went through and old name is an alias to new, then how does that break compatibility?

eclairevoyant:
Older versions of nix wouldn't know about `tryGetAttrPos`, which means they wouldn't be able to evaluate newer code, and that breakage isn't worthwhile.

PR #7401, opened 2022-12-04
getMaxCPU: fix cgroup path

Mic92:
Given this typo I am not sure if it has been tested. Introduced in https://github.com/NixOS/nix/pull/7394


edolstra:
Oops, that was a typo I introduced to check whether it still worked if cgroups were not enabled...

PR #7404, opened 2022-12-05
tests: don't refer to TMPDIR

ncfavier:
`TMPDIR` is unset when using nix-direnv to enter the development shell automatically, because nix-direnv [ignores](https://github.com/nix-community/nix-direnv/pull/31#discussion_r1039588647) the `TMPDIR` variable set by `print-dev-env`. This causes `make tests/check.sh.test` to fail.

Set it globally so we don't have to worry about it.

### DIFF
--- a/tests/check.nix
+++ b/tests/check.nix
@@ -44,7 +44,7 @@ with import ./config.nix;
   };
 
   hashmismatch = import <nix/fetchurl.nix> {
-    url = "file://" + builtins.getEnv "TMPDIR" + "/dummy";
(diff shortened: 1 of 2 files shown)


thufschmitt:
Shouldn't we rather change `tests/check.sh` to not refer to `$TMPDIR`? There's no real reason for it to do since the test environment already sets `TEST_ROOT` which is semantically a better choice.

ncfavier:
Fair

thufschmitt (approved):
Thanks :)

PR #7390, opened 2022-12-02
check the store for input before failing (hopefully fix #6383)

Radvendii:
This is a fix very similar to #7283 but in a slightly different place, which will hopefully address #6383. I can't reproduce the issue, so I'm not sure it fixes it.

The idea is that `inputDrvOutputs` is statefully constructed and is somehow not getting set properly. Rather than track down that bug (which would also be good to do at some point), this piece of duct-tape just checks the actual store as a backup when `inputDrvOutputs` doesn't have it.

PR #7409, opened 2022-12-06
check the store for input before failing (hopefully fix #6383)

### LINKED ISSUE #6383: `resolvedResult.builtOutputs` doesn't contain expected output after successful multi-output build

Radvendii:
This applies the same basic idea as #6572 and #7390, but applied to the problem occurring in #6383.

I am not super familiar with the ++ part of C++, so some things that I'm not sure about:

I have a `std::shared_ptr` that I'm turning into a regular pointer with `.get()`. The only thing happening with this pointer is accessing / copying values shortly thereafter. Is this safe? What is the canonical way of doing this?

I had to change `auto` to `const Realisation *` because otherwise it was defaulting to a regular `Realisation *` and then failing later when it needed to be `const`. I'm not sure if this is the idiomatic way of dealing with this problem.

### DIFF
--- a/src/libstore/build/derivation-goal.cc
+++ b/src/libstore/build/derivation-goal.cc
@@ -1017,30 +1017,33 @@ void DerivationGoal::resolvedFinished()
                     "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,resolve)",
                     worker.store.printStorePath(drvPath), wantedOutput);
 
-            const Realisation * realisation = get(resolvedResult.builtOutputs, DrvOutput { *resolvedHash, wantedOutput });
-            if (!realisation) {
+            auto realisation = [&]{
+              auto take1 = get(resolvedResult.builtOutputs, DrvOutput { *resolvedHash, wantedOutput });
+              if (take1) return *take1;
+
               /* The above `get` should work. But sateful tracking of
                  outputs in resolvedResult, this can get out of sync with the
                  store, which is our actual source of truth. For now we just
                  check the store directly if it fails. */
-              realisation = worker.evalStore.queryRealisation(DrvOutput { *resolvedHash, wantedOutput }).get();
-              if (!realisation) {
-                throw Error(
-                    "derivation '%s' doesn't have expected output '%s' (derivation-goal.cc/resolvedFinished,realisation)",
-                    worker.store.printStorePath(resolvedDrvGoal->drvPath), wantedOutput);
-              }
-            }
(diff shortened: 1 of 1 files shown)


Radvendii:
Okay this should be cleaner. It took a while because I was running into errors testing this, but it turns out they were unrelated and I just had to do some combination of `git clean -f -d -x`, `make clean` and `sudo nix-collect-garbage -d`. :woman_facepalming:

Radvendii:
@thufschmitt I fixed the problems we discussed about this PR. Care to take another look?

thufschmitt (approved):
Thanks, looks good :+1:

PR #7411, opened 2022-12-06
Remove GPG-signing of releases

edolstra:
This makes it easier for others to make releases, and probably few people care about GPG signatures anyway.

### DIFF
--- a/maintainers/upload-release.pl
+++ b/maintainers/upload-release.pl
@@ -115,10 +115,6 @@ sub downloadFile {
 
     write_file("$tmpFile.sha256", $sha256_actual);
 
-    if (! -e "$tmpFile.asc") {
-        system("gpg2 --detach-sign --armor $tmpFile") == 0 or die "unable to sign $tmpFile\n";
-    }
-
     return $sha256_expected;
 }
 
@@ -194,7 +190,7 @@ for my $fn (glob "$tmpDir/*") {
         my $configuration = ();
         $configuration->{content_type} = "application/octet-stream";
 
-        if ($fn =~ /.sha256|.asc|install/) {
+        if ($fn =~ /.sha256|install/) {
             # Text files
             $configuration->{content_type} = "text/plain";
         }


thufschmitt (approved):

ncfavier:
[These instructions](https://nixos.org/download.html#nix-verify-installation) should be updated.

ncfavier:
https://github.com/NixOS/nixos-homepage/pull/975

milkknife:
Any chance this pull request/decision can be reverted?
- https://www.reddit.com/r/NixOS/comments/gbnmk0/does_nixos_support_verifying_the_iso_disk_images/?rdt=38517
- https://blog.linuxmint.com/?p=2994

I believe that cryptographically verifying the authenticity and integrity of .iso releases using GPG (GNU Privacy Guard) is an essential step in ensuring the security and trustworthiness of the downloaded files. Here are a few reasons why this verification process is important:

1. Ensuring Authenticity: Cryptographic verification with GPG allows you to confirm that the .iso release you have downloaded has not been tampered with or modified by unauthorized parties. By verifying the digital signature attached to the .iso file, you can be confident that it was created and signed by the trusted entity or organization responsible for its distribution.

2. Protecting Against Malicious Downloads: Verifying the authenticity of .iso releases helps protect against the risk of downloading malicious or compromised files. Without cryptographic verification, there is a possibility that the downloaded .iso file may have been modified to include malware, viruses, or other malicious code. By using GPG to verify the digital signature, you can ensure that the file has not been tampered with and comes from a trusted source.

3. Maintaining Data Integrity: Cryptographic verification also ensures the integrity of the downloaded .iso release. It allows you to verify that the file has not been corrupted during the download process or while in transit. By comparing the computed hash or checksum of the downloaded .iso file with the one provided by the trusted source, you can confirm that the file has not been altered or modified.

4. Trust in the Software Supply Chain: Cryptographic verification is crucial for establishing trust in the software supply chain. It enables users to validate the authenticity and integrity of software releases, ensuring that they are using genuine and unmodified versions of the software. This is particularly important for operating systems, where compromised or tampered installations can have serious security implications.

5. Compliance and Regulatory Requirements: In certain industries or organizations, cryptographic verification of software releases may be a requirement to comply with regulatory standards or internal security policies. By following best practices for verifying the authenticity and integrity of .iso releases, organizations can demonstrate due diligence in ensuring the security of their software deployments.

In summary, cryptographic verification with GPG provides assurance that the downloaded .iso release is authentic, unmodified, and trustworthy. It helps protect against the risk of downloading malicious files, ensures data integrity, and establishes trust in the software supply chain.

FlyingWombat:
So I'm a bit late here, but ... WTF?

Can someone please explain to me why all of the sudden signed releases don't matter anymore?

> This makes it easier for others to make releases ...

The whole point of signed releases is that not just anyone can do it -- that's what provides authenticity. 
Really, how much of a burden is it for the release team to make signatures?

> ... and probably few people care about GPG signatures anyway.

In other words: "Other people don't care about cybesecurity, so neither should we." 
This is like removing seat-belts from cars, and banning motorcycle helmets, because some people don't want to bother with them -- you are making sure no one has the choice to use it.

If what you actually meant is that "GPG signatures actually don't matter in the first place", then that is a very different argument, and something I'd be very interested to hear about.

PR #7414, opened 2022-12-06
.version: Add newline

roberth:
Make the release process a bit less error prone.
Most files do end in newlines, some editors will add one, and some humans are inclined to add one too.


infinisil (commented):
Whoops, forgot to submit this review

PR #7418, opened 2022-12-07
add template for installer issues

fricklerhandwerk:
since the installer prompts users to file issues, labelling them
automatically should reduce triaging effort significantly.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -97,7 +97,8 @@ is_os_darwin() {
 }
 
 contact_us() {
-    echo "You can open an issue at https://github.com/nixos/nix/issues"
+    echo "You can open an issue at"
+    echo "https://github.com/NixOS/nix/issues/new?labels=installer&template=installer.md"
     echo ""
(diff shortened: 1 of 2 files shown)


scripts/install-multi-user.sh
```diff
@@ -97,7 +97,8 @@ is_os_darwin() {
 }
 
 contact_us() {
-    echo "You can open an issue at https://github.com/nixos/nix/issues"
+    echo "You can open an issue at"
+    echo "https://github.com/NixOS/nix/issues/new?labels=installer&template=installer.md"
```
thufschmitt, line 101: Nitpick: is the `labels=intaller` parameter needed since the template already specifies it?

thufschmitt (approved):
Just a very minor comment, but good idea overall :+1:

scripts/install-multi-user.sh
fricklerhandwerk: Tested this, yes you need it. You get an even longer URL with empy fields from the GitHub UI when filing an issue.

PR #7419, opened 2022-12-07
let installer blurb to to community page

fricklerhandwerk:
being too specific about it requires more maintenance (or otherwise
produced more confusion and churn), since these points of contact change
over time.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -99,11 +99,7 @@ is_os_darwin() {
 contact_us() {
     echo "You can open an issue at https://github.com/nixos/nix/issues"
     echo ""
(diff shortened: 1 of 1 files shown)


thufschmitt:
I don't have a strong opinion on that. I'll just point out that https://nixos.org/community/ points to a different Matrix channel than the one in the current message (`#community:nixos.org` versus the more specifix `#nix:nixos.org`)

fricklerhandwerk:
I don't think we want everyone who the installer fails on to post to the Nix channel. The purpose of that message should be to direct beginners to the established support paths which are more accessible and easier to maintain than this obscure piece of code in the installer.

PR #7423, opened 2022-12-07
Support flake references in the old CLI

### LINKED ISSUE #7026: Support flake references in NIX_PATH
Allow flake references in `NIX_PATH` and `-I`, e.g.
```
NIX_PATH=nixpkgs=flake:nixpkgs
```
(similar to the `channel:...` syntax).

This allows the legacy CLI to handle `<nixpkgs>` references on an otherwise fully flakified system, without having to rely on channels. In particular, it allows `<nixpkgs>` to refer to the version of Nixpkgs used to build the current system. Currently you can put something like
```
nix.nixPath = [ "nixpkgs=${nixpkgs}" ];
```
in the NixOS system configuration, but that requires `${nixpkgs}` to be materialized in the Nix store (breaking the use of e.g. zip file accessors in #6530). Also, we may want to pin nixpkgs without making it part of the system closures and making it un-GC-able.

edolstra:
Fixes #7026.

Extracted from the lazy-trees branch.

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -2,3 +2,12 @@
 
 * The `repeat` and `enforce-determinism` options have been removed
   since they had been broken under many circumstances for a long time.
+
+* You can now use flake references in the old CLI, e.g.
+
+  ```
+  # nix-build flake:nixpkgs -A hello
+  # nix-build -I nixpkgs=flake:github:NixOS/nixpkgs/nixos-22.05 \
+      '<nixpkgs>' -A hello
+  # NIX_PATH=nixpkgs=flake:nixpkgs nix-build '<nixpkgs>' -A hello
+  ```
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -3,7 +3,10 @@
 * The `repeat` and `enforce-determinism` options have been removed
   since they had been broken under many circumstances for a long time.
 
-* You can now use flake references in the old CLI, e.g.
+* You can now use [flake references] in the [old command line interface], e.g.
+
+   [flake references]: ../command-ref/new-cli/nix3-flake.md#flake-references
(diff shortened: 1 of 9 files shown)


src/libexpr/eval.cc
```diff
@@ -2583,6 +2584,23 @@ Strings EvalSettings::getDefaultNixPath()
     return res;
 }
 
+bool EvalSettings::isPseudoUrl(std::string_view s)
```
thufschmitt, line 2587: Any reason for moving that out of `filetransfer.hh`?

doc/manual/src/release-notes/rl-next.md
```diff
@@ -2,3 +2,15 @@
...
+
+   [flake references]: ../command-ref/new-cli/nix3-flake.md#flake-references
+   [old command line interface]: ../command-ref/main-commands.md
+
+  ```
+  # nix-build flake:nixpkgs -A hello
+  # nix-build -I nixpkgs=flake:github:NixOS/nixpkgs/nixos-22.05 \
+      '<nixpkgs>' -A hello
+  # NIX_PATH=nixpkgs=flake:nixpkgs nix-build '<nixpkgs>' -A hello
+  ```
```
thufschmitt, line 16: Shouldn't that also be documented in https://nixos.org/manual/nix/stable/language/values.html?#type-path ? In particular the two last examples are not directly tied to the old interface but rather to the semantics of `<foo>`

thufschmitt (changes requested):
That's a great change :)

A couple of small comments, but I like it overall

src/libexpr/eval.cc
edolstra: Yes, these "URLs" are very specific to the Nix CLI (e.g. the `channel:` and `flake:` schema) and are not used by `filetransfer.cc`.

doc/manual/src/release-notes/rl-next.md
edolstra: Added doss form the lazy-trees branch.

PR #7424, opened 2022-12-07
Configurable substituter disable time

cidkidnix:
Allow to change how long substituters are disabled for based on a nix.conf config option
Still need to add a better explanation to globals.hh

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -515,6 +515,10 @@ public:
         this, true, "enforce-determinism",
         "Whether to fail if repeated builds produce different output. See `repeat`."};
 
+    Setting<unsigned int> substituterDisableTime{
+        this, 60, "substituter-disable-time",
+            "How long to disable substituters on fail"};
+
     Setting<Strings> trustedPublicKeys{
         this,
         {"cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="},
--- a/src/libstore/http-binary-cache-store.cc
+++ b/src/libstore/http-binary-cache-store.cc
@@ -83,7 +83,7 @@ protected:
     {
         auto state(_state.lock());
         if (state->enabled && settings.tryFallback) {
-            int t = 60;
+            int t = settings.substituterDisableTime;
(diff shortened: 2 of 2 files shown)


Ericson2314 (approved):
Yeah I think most magic numbers like this should be config options. There is no globally optimal choice.

edolstra:
Just because there is no globally optimal choice, doesn't mean adding a configuration option is a good idea. The existence of options like these is generally a sign of bad design, since it punts the problem to the user.

What's the use case for a different interval?

cidkidnix:
We only disable a substituter temporarily (assuming fallback and others are enabled) at the moment (from what I can see), and it's hard set at 60sec. I'd like to be able to configure this so we don't re-trigger a substituter being checked "too-soon" for my use-case. Basically I want to be able to change how long in-between that happens.

thufschmitt:
I agree that adding an option doesn't look like a nice solution here. Ideally Nix should just behave in a way that's satisfying for everyone in this kind of scenario. Would some kind of exponential back off work better for your use-case?

Ericson2314:
The current code is absolutely atrocious for our use-case because if one substituter is inaccessible, it won't try any other.

I agree good design means one shouldn't *need* to change configuration options. E.g. if it will try the substitutors concurrently a 60 second timeout isn't so bad.

But just cause it isn't necessary to change the magic numbers doesn't mean having a config object is bad. IMO there is a big gap between on one hand obscurist endlessly configurable consultantware, and on the other patronizingly configuration-free Apple-style software. A middle ground of "there are settings but you shouldn't need to change them" feels fine to me.

cidkidnix:
Though yes what @Ericson2314 said is 100% a problem (although workaround able with #7188 or enabling fallback currently), this further allows you to configure how long substituters that are down/missing will stay disabled locally.

Ericson2314:
#7188 still tries them sequentially, so a 60 second timeout can be annoying. We would need to them concurrently for it to not be annoying.

domenkozar:
See https://github.com/NixOS/nix/pull/7188#issuecomment-1357505188

tomberek:

PR #7426, opened 2022-12-08
Avoid poly_user_note_set twice

tianyuanhao:
f06f810 incorrectly introduces a boolean flip, resulting in a senseless
 poly_user_note_set even though the user comment has been set correctly.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -575,7 +575,7 @@ EOF
     # to extract _just_ the user's note, instead it is prefixed with
     # some plist junk. This was causing the user note to always be set,
     # even if there was no reason for it.
-    if ! poly_user_note_get "$username" | grep -q "Nix build user $coreid"; then
+    if poly_user_note_get "$username" | grep -q "Nix build user $coreid"; then
         row "              Note" "Nix build user $coreid"
     else
         poly_user_note_set "$username" "Nix build user $coreid"


abathur:
I'm not sure this is right, either. 

The existing code is trying to print the note if it is already set (grep exits 0), and otherwise set the note. That seems like the correct logic, yeah?

tianyuanhao:
@abathur In _Bash_, `0` is __true__, `1` is __false__. This results in __redundant__ `poly_user_note_set`.

@fricklerhandwerk Fine, I will rewrite the commit message. I thought the commit title is fine, because it just tell people what the commit did.

abathur:
> @abathur In _Bash_, `0` is **true**, `1` is **false**.

Yes, I know. This looks correct--not sure how I mis-read your diff on the first pass. Apologies for the noise.

Ericson2314:
Wouldn't that `!` only apply to `poly_user_note_get` itself? I would think that should never return non-`0` so yeah this just seems wrong.

Do we have automatic tests for the installer?

abathur:
> Do we have automatic tests for the installer?

There is a golden-path kind of test, but you need to do some one time cachix setup to enable it. See https://nixos.org/manual/nix/stable/contributing/hacking.html

> Wouldn't that `!` only apply to `poly_user_note_get` itself? I would think that should never return non-`0` so yeah this just seems wrong.

The most relevant bit from https://www.gnu.org/software/bash/manual/bash.html#Pipelines is:

> The exit status of a pipeline is the exit status of the last command in the pipeline, unless the pipefail option is enabled (see [The Set Builtin](https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin)). If pipefail is enabled, the pipeline’s return status is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands exit successfully. If the reserved word ‘!’ precedes the pipeline, the exit status is the logical negation of the exit status as described above. The shell waits for all commands in the pipeline to terminate before returning a value.

Ericson2314:
OK thanks. Then yes this PR is correct. We don't want to flip the result of the whole pipeline, and we want to set the note if it is not already set.

Ericson2314 (approved):

thufschmitt (approved):

PR #7430, opened 2022-12-08
Ca/fix nix log

### LINKED ISSUE #6209: `nix log` doesn’t work properly with ca derivations
From https://github.com/NixOS/nix/commit/287642f132d8141269bd7c23b0bc7c1ee512b171

**Describe the bug**

`nix log foo#bar` doesn’t work with content-addressed derivations

**Steps To Reproduce**

```console
$ nix build github:regnat/nixpkgs/c049134e307cdc2e981c98f0fe89ca58e2ca4a41#hello # just a plain `nixpkgs#hello`, but content-addressed
$ nix log github:regnat/nixpkgs/c049134e307cdc2e981c98f0fe89ca58e2ca4a41#hello
error: build log of 'github:regnat/nixpkgs/c049134e307cdc2e981c98f0fe89ca58e2ca4a41#hello' is not available
```

(The log is actually here, and `nix log ./result` will show it properly)

**Expected behavior**

`nix log <installable>` should work

```console
$ nix --version   
nix (Nix) 2.7.0pre20220126_4f24a33
```

**Additional context**

This is likely due to a missing resolving phase: when building a content-addressed `hello.drv`, we actually build `resolved(hello.drv)`, so there’s no log attached to `hello.drv`, only to `resolved(hello.drv)` (and so we need to actually show the log for the resolved derivation)

Radvendii:
Fix #6209 

There was one minor tweak needed to rebase onto master, but other than that it seems to work when I test it with simple derivations. Is there a reason this wasn't merged before?

### DIFF
--- a/src/libstore/log-store.hh
+++ b/src/libstore/log-store.hh
@@ -11,7 +11,14 @@ struct LogStore : public virtual Store
 
     /* Return the build log of the specified store path, if available,
        or null otherwise. */
-    virtual std::optional<std::string> getBuildLog(const StorePath & path) = 0;
+    std::optional<std::string> getBuildLog(const StorePath & path) {
+      auto maybePath = getBuildDerivationPath(path);
+      if (!maybePath)
+          return std::nullopt;
+      return getBuildLogExact(maybePath.value());
+    }
+
+    virtual std::optional<std::string> getBuildLogExact(const StorePath & path) = 0;
 
     virtual void addBuildLog(const StorePath & path, std::string_view log) = 0;
 
--- a/src/libstore/binary-cache-store.cc
+++ b/src/libstore/binary-cache-store.cc
@@ -504,18 +504,10 @@ void BinaryCacheStore::addSignatures(const StorePath & storePath, const StringSe
 
 std::optional<std::string> BinaryCacheStore::getBuildLog(const StorePath & path)
 {
-    auto drvPath = path;
-
-    if (!path.isDerivation()) {
-        try {
-            auto info = queryPathInfo(path);
-            // FIXME: add a "Log" field to .narinfo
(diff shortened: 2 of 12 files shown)


Ericson2314:
Otherwise, this looks good. I don't think there was any reason it wasn't merged before.

Radvendii:
Yes, done. Thank you for bringing my attention to #6215, I hadn't seen that.

Ericson2314 (approved):
Thanks for making the change!

Radvendii:
Anything left to do on this? Should it be merged? @thufschmitt

src/libstore/log-store.hh
```diff
@@ -11,7 +11,14 @@ struct LogStore : public virtual Store
 
     /* Return the build log of the specified store path, if available,
        or null otherwise. */
-    virtual std::optional<std::string> getBuildLog(const StorePath & path) = 0;
+    std::optional<std::string> getBuildLog(const StorePath & path) {
+      auto maybePath = getBuildDerivationPath(path);
+      if (!maybePath)
+          return std::nullopt;
+      return getBuildLogExact(maybePath.value());
+    }
```
thufschmitt, line 19: Can you move the implementation to the `.cc` file? That makes the interface file easier to read and improves the build speed a tiny bit.

thufschmitt (commented):
> Anything left to do on this? Should it be merged?

No, sorry.

I had left a comment on it, but it stayed in a “pending” state. I've just published it, but it's not a blocker, so let's merge!

thufschmitt (approved):

src/libstore/log-store.hh
thufschmitt: Opened #7597 for that

PR #7439, opened 2022-12-09
feat: add :clean command to Nix REPL

aldoborrero:
## Rationale

When playing with the Nix REPL, as a user you can pollute your scope of variables. Right now, to start fresh whilst inside the REPL you have to:

1) Type `:q`
2) `$ nix repl`

I think adding a `:clear` command (with a `:cc` shortcut) may improve the usability of the REPL a little bit.

### DIFF
--- a/src/libcmd/repl.cc
+++ b/src/libcmd/repl.cc
@@ -497,6 +497,7 @@ bool NixRepl::processLine(std::string line)
              << "  :p <expr>     Evaluate and print expression recursively\n"
              << "  :q            Exit nix-repl\n"
              << "  :r            Reload all files\n"
+             << "  :cc           Clean the Nix REPL environment\n"
              << "  :sh <expr>    Build dependencies of derivation, then start nix-shell\n"
              << "  :t <expr>     Describe result of evaluation\n"
              << "  :u <expr>     Build derivation, then start nix-shell\n"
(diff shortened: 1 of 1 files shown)


Kranzes:
Why `cc` and not just `c`?

aldoborrero:
@Kranzes mostly because of the [continue command in debug mode](https://github.com/aldoborrero/nix/blob/feat/nix-repl-clean-command/src/libcmd/repl.cc#L516):

```c++
if (state->debugRepl) {
             std::cout
             << "\n"
             << "        Debug mode commands\n"
             << "  :env          Show env stack\n"
             << "  :bt           Show trace stack\n"
             << "  :st           Show current trace\n"
             << "  :st <idx>     Change to another trace in the stack\n"
             << "  :c            Go until end of program, exception, or builtins.break\n"
             << "  :s            Go one step\n"
             ;
        }
```

Maybe better to switch it?

thufschmitt:
Is `:reload` not enough? It has precisely the semantics of closing and reopening the repl.

aldoborrero:
@thufschmitt, what if I want to start fresh without variable pollution? 

Say, for example, I brought to scope a flake, and I want to clean up the REPL env; with the `:reload` command, it will reload everything that was already there, bringing to scope the flake.

roberth:
> Is `:reload` not enough? It has precisely the semantics of closing and reopening the repl.

Only for `:lf`. `:l` reloads properly.

```
nix-repl> :l .
warning: Git tree '/home/user/h/nix' is dirty
Added 16 variables.

nix-repl> packages?x86_64-linux
true

nix-repl> :r                    
Loading '.'...
warning: Git tree '/home/user/h/nix' is dirty
Added 16 variables.

```

thufschmitt:
Discussed during the Nix team meeting: This doesn't seem warranted given that restarting the repl is nearly instantaneous. @roberth 's comment is hinting at a separate but real bug with repl, so opened https://github.com/NixOS/nix/issues/8753 to track it.

PR #7441, opened 2022-12-10
Ignore the enforceDeterminism value

andir:
We used to set enforceDeterminism to true in the settings (by default) and thus did send a non-zero value over the wire. The value should probably be ignored as it should only matter if nrRounds is non-zero as well.

Having the old code here where the value is expected to be zero only works with the same version of Nix where we are sending zero. We should always test this against older Nix versions being client or server as otherwise upgrade in larger networks might be a pain.

Fixes https://github.com/NixOS/nix/commit/8e0946e8df968391d1430af8377bdb51204e4666

See https://github.com/NixOS/nix/pull/7099#issuecomment-1344943739 for further discussion.


Ericson2314 (approved):
Thanks so much for fixing!

thufschmitt (approved):
Thanks for the fix @andir !

> We should always test this against older Nix versions being client or server as otherwise upgrade in larger networks might be a pain.

Yes! We do test against an older Nix daemon, but not an older Nix client. All the logic is there, we “just” need to add it and look at all the tests failures in that scenario, see whether they are justified or not.

Would you feel like doing that by any chance? :innocent:

Ericson2314:
I have PRs that did that but they were never merged. It would be great @andir if you want to resurect them.

andir:
> Thanks for the fix @andir !
> 
> > We should always test this against older Nix versions being client or server as otherwise upgrade in larger networks might be a pain.
> 
> Yes! We do test against an older Nix daemon, but not an older Nix client. All the logic is there, we “just” need to add it and look at all the tests failures in that scenario, see whether they are justified or not.
> 
> Would you feel like doing that by any chance? 😇

I tried to figure out how that stuff currently works but cut off because I was lost in the mixture of bash and Nix.

Could probably figure it out but I did time box this as it already distracted me from other things.

> I have PRs that did that but they were never merged. It would be great @andir if you want to resurect them.

@Ericson2314  Can you link to those? I could try rebasing / updating them.

PR #7442, opened 2022-12-11
install-multi-user: chown bootstrap store contents

winterqt:
Before this change, the bootstrap store contents would be owned by the user who ran the script.

This is not ideal, and leads to confusing issues like zsh's compaudit being fired.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -810,6 +810,24 @@ install_from_extracted_nix() {
         _sudo "to make the new store non-writable at $NIX_ROOT/store" \
               chmod -R ugo-w "$NIX_ROOT/store/"
 
+        # This is copied from create_directories, see it for why we do all this stuff just to find chown.
+        local get_chr_own="$(PATH="$(getconf PATH 2>/dev/null)" command -vp chown)"
+        if [[ -z "$get_chr_own" ]]; then
+            get_chr_own="$(command -v chown)"
+        fi
+
+        if [[ -z "$get_chr_own" ]]; then
+            reminder <<EOF
+I wanted to change ownership of new Nix store files,
+but I couldn't locate 'chown'. (You may need to fix your PATH.)
+To manually change file ownership, you can run:
+    sudo chown -R 'root:$NIX_BUILD_GROUP_NAME' '$NIX_ROOT/store'
+EOF
+        else
+            _sudo "to change ownership of Nix store files" \
+                  "$get_chr_own" -R "root:$NIX_BUILD_GROUP_NAME" "$NIX_ROOT/store" || true
+        fi
+
         if [ -d "$NIX_INSTALLED_NIX" ]; then
             echo "      Alright! We have our first nix at $NIX_INSTALLED_NIX"
         else


scripts/install-multi-user.sh
```diff
@@ -810,6 +810,23 @@ install_from_extracted_nix() {
...
+            reminder <<EOF
+I wanted to change ownership of new Nix store files,
+but I couldn't locate 'chown'. (You may need to fix your PATH.)
+To manually change file ownership, you can run:
+    sudo chown -R 'root:$NIX_BUILD_GROUP_NAME' '$NIX_ROOT'
+EOF
+        else
+            _sudo "to change ownership of Nix store files" \
+                  "$get_chr_own" -R "root:$NIX_BUILD_GROUP_NAME" "$NIX_ROOT" || true
+        fi
```
winterqt, line 828: @abathur Do you think it would be worthwhile to factor this out into a function out, since it's also done in the case that the store already exists?

abathur: I'm of two minds:
1. Broadly/normally: yes
2. A restructuring will probably take more testing. There's an effort afoot to make a binary installer. It isn't a given it'll succeed and be adopted, but I'm personally inclined to make minimum-effort fixes until it's clear how that effort will shake out. I think it makes sense to leave a comment and wait for a third repetition to actually extract it?

winterqt: > A restructuring will probably take more testing.

We could just make a function like `chown_store` and pass a parameter to adjust the message accordingly if we really want, though -- I don't think it would be that much of a stretch.

> I think it makes sense to leave a comment and wait for a third repetition to actually extract it?

Sure, makes sense. How's what I just pushed?

scripts/install-multi-user.sh
```diff
@@ -810,6 +810,25 @@ install_from_extracted_nix() {
         _sudo "to make the new store non-writable at $NIX_ROOT/store" \
               chmod -R ugo-w "$NIX_ROOT/store/"
 
+        # This is copied from the implementation in create_directories, see
+        # it for why we do all this stuff just to find chmod.
```
abathur, line 814:
```suggestion
        # it for why we do all this stuff just to find chown.
```

Yeah? :)

winterqt: Fixed :)

winterqt:
Oh, so that... undos code owner reviews too... 🙃

abathur:
I don't feel like I have a handle on the high-level correctness of the change, but we can start by setting up testing as a backstop. 

If you follow the steps in https://nixos.org/manual/nix/stable/contributing/hacking.html#installer-tests it'll enable the installer test job in your fork's GA workflow. 

It's not a full test suite, but it at least confirms that the happy-path isn't broken. It'll also be possible to use the  x86_64-darwin installer it generates outside of the workflow to confirm.

winterqt:
I manually tested the change before I submitted this PR, and it worked as expected.

I'll setup the tests in a few hours, though, and we can go from there.

winterqt:
> It'll also be possible to use the x86_64-darwin installer it generates outside of the workflow to confirm.

This isn't a Darwin specific issue, I was also able to reproduce this on Linux.

abathur:
Sorry--jumping to conclusions since most of my installer work is for macOS fixes. It'll generate an `x86_64-linux` installer as well.

winterqt:
Yup, was just making sure you were aware that this affects both Darwin and Linux.

I've tested that this works on aarch64-darwin (thanks to Virtualization.framework), and I'll generate an installer for aarch64-linux and test there. (Maybe we can even add this as a case for installer_test, that is, making sure the store permissions are setup correctly?)

(Side note, but we should probably enable aarch64-linux cross builds for the installer workflow... I'll test + do that in another PR :)

winterqt:
@abathur Installer tests succeeded: https://github.com/winterqt/nix/actions/runs/3671329069.

scripts/install-multi-user.sh
```diff
@@ -810,6 +810,24 @@ install_from_extracted_nix() {
         _sudo "to make the new store non-writable at $NIX_ROOT/store" \
               chmod -R ugo-w "$NIX_ROOT/store/"
 
+        # This is copied from create_directories, see it for why we do all this stuff just to find chown.
```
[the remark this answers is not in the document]
winterqt: See https://github.com/NixOS/nix/pull/7442#discussion_r1045152339 for @abathur and I's previous discussion on this.

scripts/install-multi-user.sh
```diff
@@ -810,6 +810,24 @@ install_from_extracted_nix() {
...
+        if [[ -z "$get_chr_own" ]]; then
+            reminder <<EOF
+I wanted to change ownership of new Nix store files,
+but I couldn't locate 'chown'. (You may need to fix your PATH.)
+To manually change file ownership, you can run:
+    sudo chown -R 'root:$NIX_BUILD_GROUP_NAME' '$NIX_ROOT'
+EOF
+        else
+            _sudo "to change ownership of Nix store files" \
+                  "$get_chr_own" -R "root:$NIX_BUILD_GROUP_NAME" "$NIX_ROOT" || true
```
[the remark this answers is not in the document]
winterqt: (Would it really have a practical effect at this point in the install process. though -- are the per-user directories even created? Just curious, I'll still change it for correctness + the reminder would need to be fixed anyways.)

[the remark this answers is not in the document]
winterqt: Fixed.

edolstra:
@abathur Do you have any opinion on this? 

Maybe instead of doing a `chown -R` on the store, we can do a `chown -R` on the unpacked tarball.

abathur:
I said I wasn't sure about high-level correctness before, and I guess that's because I'm a little perplexed about why it's only showing up now if this has been happening on normal installs in the wild. It seems like anything running later under the installing user being able to fiddle with the bootstrapped files is probably Bad, so this makes me feel like there's an unknown variable somewhere.

I guess an explanation that would make sense is if this shifted as a result of https://github.com/NixOS/nix/pull/5150

Idiomatically, this addition mirrors the chmod to disable write permissions above it. If the change above did cause ownership to shift, I guess the ownership of the unpacked files may not get preserved during the recursive copy anyways?

I can't fiddle around with the commands to test that thesis just now. Maybe this evening.

abathur:
Though the `ls` output in https://github.com/NixOS/nix/pull/7603#issuecomment-1411124619 does indicate root root for something that's presumably in the bootstrap. 

@iFreilicht did you happen to notice the installing user owning any store files while working through #7603?

abathur:
Got out a spare laptop to test. I do see this behavior on modern installers. Can verify with something like `find /nix -user $USER` 

The first release with the rsync -> cp change from 5150 was 2.4, and the behavior is present there. I do _not_ see the behavior on 2.3.16.

I think this is because the `-p` flag to rsync preserves permissions but not ownership, while the `-p` flag to cp preserves permissions, ownership, and more. Since it looks like the `cp` command should preserve the ownership, I think @edolstra is right that we should be able to chown the unpacked store instead.

iFreilicht:
> @iFreilicht did you happen to notice the installing user owning any store files while working through https://github.com/NixOS/nix/pull/7603?

I did not, but I didn't look particularly closely at that. The issues I fixed were not caused by an ownership issue.

tomberek:

PR #7446, opened 2022-12-11
Document ability to disable the global registry

### LINKED ISSUE #4874: Allow empty config `flake-registry = ""`, to disable default global flakes

bew:
Closes #4874 
Replaces #5420 

While updating my old PR #5420, I realized it is already possible to disable the global flake registry, due to this line:
https://github.com/NixOS/nix/blob/145e9a81230604baae706371cd19861ba9da3766/src/libfetchers/registry.cc#L19-L20
Which is called in `getGlobalRegistry`, when `path` starts with a `/`:
https://github.com/NixOS/nix/blob/145e9a81230604baae706371cd19861ba9da3766/src/libfetchers/registry.cc#L164

So instead of adding ability to set it to `""` (which I can still add if you think it's a better UX), I figured I'd add some tests and document it in the description of `flake-registry`.

### DIFF
--- a/src/libfetchers/fetch-settings.hh
+++ b/src/libfetchers/fetch-settings.hh
@@ -71,7 +71,10 @@ struct FetchSettings : public Config
         "Whether to warn about dirty Git/Mercurial trees."};
 
     Setting<std::string> flakeRegistry{this, "https://channels.nixos.org/flake-registry.json", "flake-registry",
-        "Path or URI of the global flake registry."};
+        R"(
+          Path or URI of the global flake registry.
+          When the path does not exist, disables the global flake registry.
+        )"};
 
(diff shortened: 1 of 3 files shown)


thufschmitt:
I actually like #5420. @edolstra anything keeping it back (except the small merge conflict)?

edolstra:
Not that I know.

Issue #7314, opened 2022-11-16
Improve `readDir` performance


roberth:
**Is your feature request related to a problem? Please describe.**

`builtins.readDir` currently performs stat calls for each directory entry. This is bad enough that @aakropotkin works around `readDir` with a generated file. A fast `readDir` may allow Nixpkgs to replace `all-packages.nix` over time; a project that the Nixpkgs Architecture Team is researching.

**Describe the solution you'd like**

Implementation 1, posix compatible:
Fill the attrset with thunks that call `pathType` https://github.com/NixOS/nix/issues/3096

Implementation 2, linux optimized:
Use [`getdents` or `getdents64`](https://man7.org/linux/man-pages/man2/getdents.2.html)

**Describe alternatives you've considered**

Avoid `readDir` on large directories, generate a list of files. The `readDir` function is not crucial for the architecture team proposal.

**Additional context**

https://github.com/nixpkgs-architecture/auto-called-packages/issues/17#issue-1450633492

edolstra:
I'm a bit skeptical about `auto-called-packages`. It inherently makes any Nix evaluation take time linear in the size of Nixpkgs, rather than the size of the packages used by the derivation you're building. A full traversal would be deadly on non-SSD drives (and even on SSD it takes >2 seconds on a cold cache), but maybe we don't need to care about that anymore.

Having said that, `readDirectory` is supposed to avoid `stat` calls by returning a `type` hint (which may be `DT_UNKNOWN` depending on the filesystem/OS).

roberth:
> It inherently makes any Nix evaluation take time linear in the size of Nixpkgs

Technically this is already the case, but with a small constant factor.
We won't be scanning recursively, but if performance suffers, we will use a generated list in a file instead.

> Having said that, `readDirectory` is supposed to avoid `stat` calls by returning a `type` hint (which may be `DT_UNKNOWN` depending on the filesystem/OS).

It seems that we've overlooked this. "Implementation 2" seems to be already implemented then. The `.type` thunk optimization could still be implemented.

Ericson2314:
Yeah it is already linear time today. And even if Nixpkgs continues to not use it, it should still be better if we can make it better. Finally, if we ever do https://github.com/NixOS/nix/issues/4090 we can simply stat paths (or the lazy trees equivalent) when attributes are accessed, fully lazy again!

edolstra:
[...]

Issue #3096, opened 2019-09-18
New builtin: pathType


roberth:
I'm proposing to add a new `builtins.pathType` that returns the node type, similar to the values returned by `builtins.readDir`.

### Motivation

Avoid having to read a directory when all you need is a single path's node type.

### Spec

Like `builtins.pathExists`, its argument is a path or string that may or may not exist on the filesystem.

Given argument `p` return value is
 - `null` when the path `p` does not point to an existing filesystem node
 - a value equal to `(builtins.readDir (dirOf p))."$(baseNameOf p)"`, such as "directory", "regular", "symlink"

### Advantages

 - Performance: no need to read the entire directory (may be big, like `/nix/store`)
 - Less strict: `readDir (dirOf p)` may not be allowed by dir permissions
 - Decreases the file watch scope for tools like `lorri`

roberth:
`pathType` on a store path is generally not something you'll want to do, because it blocks the evaluator, causing the evaluation heap to exist for longer than necessary (wasting memory not usable by builds) and creating a choke point for parallelism.

Most of the time, it's possible to check the type of the store path in the builder that uses that information.

For example, don't do

```nix
runCommand "foo" { } ''
  ${optionalString (pathType dep == "directory") ''
    echo dep is a directory
  ''}
''
```

but instead do

```nix
runCommand "foo" { } ''
  if [[ -d ${dep} ]]; then
    echo dep is a directory;
  fi
''
```

RuRo:
Unfortunately, this doesn't work for me (and probably won't work for everybody). There are cases, where `pathType` is needed at expression evaluation time, not at build/run time, so it can't be offloaded to the build step. For example, running `lib.filesystem.listFilesRecursive` on some source tree and `import`ing the `.nix` files.

Also, IMHO, it doesn't matter whether getting the `pathType` of store paths is a good practice or not. You should be able to get the type of `whatever/path/foo` without listing the contents of `whatever/path`.

I am currently working around this issue by adding an extra directory level to my flake, so I am not too invested in this issue, but I still think, that this ought to work.

roberth:
Right, it used to be that you could just avoid putting your project sources in the store (except for derivation sources, by necessity). I hope #6530 makes that work a little better, but I don't assume it does. Reopening.

mkenigs:
+1 I ran into this as well

thufschmitt:
Fixed by https://github.com/NixOS/nix/pull/7447

roberth:
[...]

PR #7447, opened 2022-12-12
Read file type

aakropotkin:
Adds a new primop `readFileType`, allowing `readDir` to create thunks for unknown types when reading large directories.


aakropotkin:
@Ericson2314 @infinisil

Ericson2314:
CC @roberth, it was his idea in #7314

aakropotkin:
As a note to @roberth it looks like on Linux `readents` is already used.

With that in mind the changes to `readDir` largely effect Darwin systems.

Benchmarking is TODO.

src/libexpr/primops.cc
```diff
@@ -1638,6 +1638,29 @@ static RegisterPrimOp primop_hashFile({
     .fun = prim_hashFile,
 });
 
+
+static void prim_readFileType(EvalState & state, const PosIdx pos, Value * * args, Value & v)
```
roberth, line 1642:
```suggestion
static void prim_pathType(EvalState & state, const PosIdx pos, Value * * args, Value & v)
```

Reading is an implementation detail, and it's not necessarily a file.
See https://github.com/NixOS/nix/issues/3096
Also I think it would be nice to let the caller handle the missing file case themselves
> `null` when the path p does not point to an existing filesystem node

Otherwise, some functions will need to do `pathExists` first, which amounts to another `stat`.

src/libexpr/primops.cc
```diff
@@ -1638,6 +1638,29 @@ static RegisterPrimOp primop_hashFile({
...
+        type == DT_DIR ? "directory" :
+        type == DT_LNK ? "symlink" :
+        "unknown");
+}
+
+static RegisterPrimOp primop_readFileType({
+    .name = "__readFileType",
+    .args = {"p"},
+    .doc = R"(
+      Determine the directory entry type of a file, being
```
roberth, line 1658:
```suggestion
      Determine the directory entry type of a file system node, being
```

src/libexpr/primops.cc
```diff
@@ -1648,13 +1671,20 @@ static void prim_readDir(EvalState & state, const PosIdx pos, Value * * args, Va
...
-            ent.type == DT_REG ? "regular" :
-            ent.type == DT_DIR ? "directory" :
-            ent.type == DT_LNK ? "symlink" :
-            "unknown");
+        auto & attr = attrs.alloc(ent.name);
+        if (ent.type == DT_UNKNOWN) {
+            auto epath = state.allocValue();
+            Path path2 = path + "/" + ent.name;
+            epath->mkString(path2);
+            attr.mkApp(&state.getBuiltin("readFileType"), epath);
```
roberth, line 1679: `state.getBuiltin` can be floated outside of the loop.

That's not great when the list is empty, so an early return would be nice.

roberth (changes requested):

src/libexpr/primops.cc
```diff
@@ -1638,6 +1638,29 @@ static RegisterPrimOp primop_hashFile({
...
+static void prim_readFileType(EvalState & state, const PosIdx pos, Value * * args, Value & v)
+{
+    auto path = realisePath(state, pos, *args[0]);
+    /* Enum DT_(DIR|LNK|REG|UNKOWN) */
+    auto type = getFileType(path);
+    v.mkString(
+        type == DT_REG ? "regular" :
+        type == DT_DIR ? "directory" :
+        type == DT_LNK ? "symlink" :
+        "unknown");
```
edolstra, line 1651: This should be deduplicated with the same code in `readDir`. That's already done in the lazy-trees branch: https://github.com/NixOS/nix/pull/6530/files#diff-0f59bb6f197822ef9f19ceae9624989499d170c84dfdc1f486a8959bb4588cafR1654-R1662

aakropotkin:
> > This should probably be changed into a `stat` primop that also returns file size and whether a regular file is executable.
> 
> I agree, except `stat` is just a terrible awful non-name.
> 
> How about `pathInfo`?
> 

`stat` should return the full set of `stat` info, most importantly permissions. I see that as a useful built-in but separate from our goal here.

Ericson2314:
@aakropotkin I think the advantage of this "reduced stat" is that it returns *just* the attributes we care about in the store's model of the file system.

roberth:
I think `pathType` and the performance improvement combine really well into a single small change that does everything we need it to.

Changing it to include any extra information would be a more significant change to the language, that would be less coherent with the `readDir` builtin. If we choose to implement a reduced stat, we should "change" `readDir` to reflect this change.

```python
# completing the current interface
pathType f
  = (readDir (dirOf f)).${baseNameOf f}
  = "file" or ...

# a possible future interface
pathInfo f
  = (directoryNodes (dirOf f)).${baseNameOf f}
  = { type = "file"; executable; size; }
 or { type = "directory"; }
 or ...
```

I don't think we have a feature request for the latter (except an outlier optimization bug that @aakropotkin subtly referenced here, that we should probably fix in some other way) whereas we do for `pathType`.

I suggest we complete the current interface, which does not interfere with the possible future interface, thereby achieving the goals in the _simplest_ way.
 - #7314 
 - #3096

This way we avoid having to answer questions that we really don't need to answer, such as whether the extra attrsets impact performance, which attributes should even be in there, etc. (+ discussions, extra review rounds that we can better spend on the larger issues)

There is a time for `pathInfo`, but I don't think it should be a priority now. `pathType` otoh is low hanging fruit. That will unblock the Nixpkgs Architecture Team and solve a validated use case.

aakropotkin:
I'm on board with @roberth aside from the function name. I'd personally have a lot of use cases for `pathInfo` but don't want to weigh down the performance improvements we're chasing in `readDir` by smuggling those features into this PR.

I'll take another pass at the notes we have here and update the PR.

roberth (approved):

roberth:
Field testing the a PR checklist, because I messed up. Apologies for not doing this sooner. The Nix team is improving.

* [x]  is the idea good? has it been discussed by the Nix team?
* [x]  unit tests
      - I don't think a unit test would add value here, as the logic is quite simple and heavily reliant on OS interfaces. Mocking this would not be an effective use of our time and code complexity budget.
* [ ]  functional tests (`tests/**.sh`)
       - Please add a test to invoke the new primop
* [x]  documentation in the manual
       - Done through generated `builtins.md`
* [ ]  documentation in the code (if necessary; ideally code is already clear)
* [ ]  documentation in the commit message (why was this change made? for future reference when maintaining the code)
       - Please squash
* [ ]  documentation in the changelog (to announce features and fixes to existing users who might have to do something to finally solve their problem, and to summarize the development history)
       - Please document both the new builtin and the performance improvement, separately

src/libexpr/primops.cc
```diff
@@ -1647,14 +1677,20 @@ static void prim_readDir(EvalState & state, const PosIdx pos, Value * * args, Va
...
     for (auto & ent : entries) {
-        if (ent.type == DT_UNKNOWN)
-            ent.type = getFileType(path + "/" + ent.name);
-        attrs.alloc(ent.name).mkString(
-            ent.type == DT_REG ? "regular" :
-            ent.type == DT_DIR ? "directory" :
-            ent.type == DT_LNK ? "symlink" :
-            "unknown");
+        auto & attr = attrs.alloc(ent.name);
+        if (ent.type == DT_UNKNOWN) {
```
roberth, line 1684:
```suggestion
        if (ent.type == DT_UNKNOWN) {
          // some fs or os may not be able to return detailed node info quickly
          // in this case we produce a thunk to query the file type lazily
```

fricklerhandwerk:

 - original goal of improving performance is good
 - `readFileType` is nice side catch; no need make a design change to the file type / directory primops API yet
 - assigned to @roberth for merging

fricklerhandwerk:

- @fricklerhandwerk: release notes are missing, this is a significant feature for improving performance in some use cases
- @thufschmitt: this changes the language API, we currently don't have a process to handle that - we can't change that after the fact.
    - would be good if we could make such added features experimental for some time to make sure to get proper feedback if this was the right design idea
        - it may turn out someone will come around and argue that a `stat` syscall would be better
    - @edolstra: historically we didn't have this because for builtins so far they were obviously useful, typically for performance reasons
        - in this specific case it's tricky to make it experimental because it's used unconditionally by `readDir`

Ericson2314:
> in this specific case it's tricky to make it experimental because it's used unconditionally by `readDir`

Not saying we should do that (I think it is probably fine to just stabilize this) but I think we could make an "anonymous primop" that doesn't require a stable feature but also can't be accessed by user code.

roberth:
We definitely could, but we don't need to go there because we have a valid use case #3096 and we've discussed the design, deciding not to over-engineer and stick to the valid use case, as it is sufficient and coherent with the existing `readDir` interface. Reverting the new interface would be wasteful.

Ericson2314:
I agree. I just wanted to point that a middle ground does exist if others insisted -- not advocate for it!

aameen-tulip:
@roberth I made the changes you requested. I pushed an empty commit with the message to be used on the squashed form of this PR.

aameen-tulip:
Merged 2.13.0, squashed, and ready to merge.

Ericson2314 (approved):

roberth (approved):

aakropotkin:
> > `readFileType` is nice side catch; no need make a design change to the file type / directory primops API yet
> 
> Why was this renamed from `readPathType`/`pathType`? There's no arguments written down in any comments or meeting notes.
> 
> I think `readFileType` is inaccurate and misleading, because it also works on directories, not just files. I know that @fricklerhandwerk argued before that in Unix, directories are also technically files, but I don't think most people think of it that way.
> 
> We now have this situation:
> 
>     * `builtins.readDir` exists, requires a directory, not a file
> 
>     * `builtins.readFile` exists, requires a file, not a directory
> 
>     * `builtins.readFileType` exists, but it works on files and directories
> 
> 
> Also there's already a `lib.pathType` [in Nixpkgs](https://nixos.org/manual/nixpkgs/stable/#function-library-lib.sources.pathType) that works on both files and directories. Related is [NixOS/nixpkgs#224834](https://github.com/NixOS/nixpkgs/pull/224834)
> 
> I think this should really have been `builtins.pathType` or `builtins.readPathType`, I prefer the latter because of @aakropotkin's argument [here](https://github.com/NixOS/nix/pull/7447#discussion_r1045934197).

I think I was intentionally trying to avoid confusing "path type" ( the Nix construct ) with "filesystem path". I felt that `builtins.isPath ./foo` and `builtins.pathType ./foo` could both be grokked as "a predicate that checks if `./foo` is of Nix type `"path"`. Similarly `builtins.readPathType ./foo` could be interpreted as "read the file `./foo` which must be of Nix type `"path"` ( more of a stretch ).

In retrospect `builtins.readPathType` was probably a better name.

roberth:
I wasn't super happy about the name, but we had already bikeshedded towards this and at least two people agreed on this name. I guess I have to be more annoying then.
Instead of worrying or breaking something, let's just establish that the builtins suck. They're our equivalent of the kernel syscall interface. Have you seen the number of stat calls?

So `lib` can choose its own sensible name; problem solved.

> inaccurate and misleading

Good. I hope this is the final drop in the bucket and we actively start avoiding direct use of builtins.
Slightly trolling, but I do mean it.

infinisil:
Could we backport this to the 2.13 release since that will [be the default for NixOS 23.05](https://github.com/NixOS/nixpkgs/pull/233439#issuecomment-1557416142) and because `builtins.readFileType` is the only way to get the path type of a Nix store path with pure evaluation?

roberth:
> the only way to get the path type of a Nix store path with pure evaluation?

Is it? I thought the `readDir` solution worked fine.
I suppose maybe not in cases where `dirOf` leaves the approved paths, but isn't that more of a restrict-eval thing anyway?

I do support a backport; I just don't understand the urgency.

infinisil:
@roberth The more specific problem I ran into was this:

```nix
{
  outputs = { nixpkgs, ... }: {
    test = nixpkgs.lib.pathType ./.;
  };
}
```

```
$ nix eval .#test
error: access to absolute path '/nix/store' is forbidden in pure eval mode (use '--impure' to override)
```

This is caused because the old `pathType` before [this PR](https://github.com/NixOS/nixpkgs/pull/224834) relied on `readDir (dirOf path)`, which in this case, since flakes get copied to the store before evaluation, `readDir (dirOf /nix/store/...) == readDir /nix/store`, therefore trying to list the entire Nix store, obviously disallowed in pure evaluation mode.

This is not specific to flakes though, pure evaluation of any `./.` inside a fetched path at the root would cause the same.

PR #7461, opened 2022-12-14
Improve warning when an untrusted user is using an untrusted substituter

### LINKED ISSUE #6672: when a substituter is untrusted, give a solution in the warning message
**Describe the bug**

When running a flakes program through `nix run` that is providing a substituter, the warnings messages in the current code are not really helpful.

**Steps To Reproduce**

1. run `nix run github:tweag/nickel -- --help`
2. answers `y` to the multiple questions about a substituter
3. you will see a message `warning: ignoring untrusted substituter 'https://nickel.cachix.org'`

Because you accept all the questions about trust, it's confusing to see the substituter is ignored because it is untrusted.

**Expected behavior**

I think the last warning message should add a solution: `[...], if you want to use this substituter, either add your user to the trusted-users list, or add the substituter to the trusted-substituters list in your Nix configuration.`

**`nix-env --version` output** : nix-env (Nix) 2.9.1

**Additional context**

My user is not a trusted user and the substituter wasn't listed in the trusted-substituters. Setting one is enough to have the cache enabled.

rapenne-s:
This change brings a huge warning message with explanations when
an untrusted user (wrt nix) is trying to use a substituter that isn't
trusted in nix.conf, this is a very common case nowadays with flakes
containing substituter informations.

The point is to help the user by giving the following information:

- why is this happening
- what are the solutions to fix it
- what are the risks
- how to learn more about the explanation

fix #6672

**Release Notes**

Does it deserve an entry in release notes?

**Testing**

All the tests are still passing.

**Priorities**

### DIFF
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -222,7 +222,8 @@ struct ClientSettings
                     else if (!hasSuffix(s, "/") && trusted.count(s + "/"))
                         subs.push_back(s + "/");
                     else
-                        warn("ignoring untrusted substituter '%s'", s);
+                        warn("ignoring untrusted substituter '%s', you are not a trusted user.\n"
+                             "More information about 'trusted-substituters' option in nix.conf man page", s);
                 res = subs;
                 return true;
             };
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -223,7 +223,7 @@ struct ClientSettings
                         subs.push_back(s + "/");
                     else
                         warn("ignoring untrusted substituter '%s', you are not a trusted user.\n"
-                             "More information about 'trusted-substituters' option in nix.conf man page", s);
+                             "Run `man nix.conf` for more information on the `substituters` configuration option.", s);
                 res = subs;
                 return true;
             };
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -570,11 +570,15 @@ public:
         {"cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="},
         "trusted-public-keys",
         R"(
-          A whitespace-separated list of public keys. When paths are copied
-          from another Nix store (such as a binary cache), they must be
-          signed with one of these keys. For example:
-          `cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
-          hydra.nixos.org-1:CNHJZBh9K4tP3EKF6FkkgeVYsS3ohTl+oS0Qa8bezVs=`.
+          A whitespace-separated list of public keys.
+
+          At least one of the following condition must be met
+          for Nix to accept copying a store object from another
+          Nix store (such as a substituter):
+
+          - the store object has been signed using a key in the trusted keys list
+          - the [`require-sigs`](#conf-require-sigs) option has been set to `false`
+          - the store object is [output-addressed](@docroot@/glossary.md#gloss-output-addressed-store-object)
         )",
         {"binary-cache-public-keys"}};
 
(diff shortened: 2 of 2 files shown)


src/libstore/daemon.cc
```diff
@@ -222,7 +222,15 @@ struct ClientSettings
...
                         subs.push_back(s + "/");
                     else
-                        warn("ignoring untrusted substituter '%s'", s);
+                        warn("warning: trying to use an untrusted substituter '%s', but you are not a trusted user.\n\n"
+                             "/!\\ A trusted but malicious actor could hack your system through this substituter.\n\n"
+                             "If you trust this substituter, either\n"
+                             " - run the command again as a trusted user, or mark your user as trusted in /etc/nix/nix.conf:\n"
+                             "   trusted-users = <users>\n"
+                             " - add the substituter and its public key to nix.conf:\n"
+                             "   substituters = %s\n"
```
tomberek, line 231:
```suggestion
                             "   trusted-substituters = %s\n"
```

src/libstore/daemon.cc
```diff
@@ -222,7 +222,15 @@ struct ClientSettings
                     else if (!hasSuffix(s, "/") && trusted.count(s + "/"))
                         subs.push_back(s + "/");
                     else
-                        warn("ignoring untrusted substituter '%s'", s);
+                        warn("warning: trying to use an untrusted substituter '%s', but you are not a trusted user.\n\n"
+                             "/!\\ A trusted but malicious actor could hack your system through this substituter.\n\n"
```
tomberek, line 226: This being a warning has also been frustrating, often it is critical to use the provided substituter.

edolstra: "warning:" is superfluous because `Logger::warn()` already adds that (see `logging.cc`).

edolstra:
Rather than having a long warning message, it might be better to link to some section in the manual.

jonringer:
> Rather than having a long warning message, it might be better to link to some section in the manual.

Or go the route of the rustc compiler, in which you get a short message, and a link to a long explanation.

src/libstore/daemon.cc
rapenne-s: I've drop the extra warning word.
As for the critical level, I'm not sure it's fitting because that doesn't prevent Nix to continue (and maybe compile everything). However, the behavior could be changed to make nix die when in this situation, instead of only displaying a warning.

rapenne-s:
I'm convinced it's important to get immediate information, that doesn't even require Internet or a web browser to read, I think we could add a link to an deeper explanation (if we had it), but we should keep a local explanation.

rapenne-s:
@fricklerhandwerk I made the changes, however I don't know if we want to make a standardized wording to refer to documentation. I don't think we have any at the moment, but it would be practical for further similar changes.

src/libstore/daemon.cc
```diff
@@ -222,7 +222,8 @@ struct ClientSettings
                     else if (!hasSuffix(s, "/") && trusted.count(s + "/"))
                         subs.push_back(s + "/");
                     else
-                        warn("ignoring untrusted substituter '%s'", s);
+                        warn("ignoring untrusted substituter '%s', you are not a trusted user.\n"
+                             "More information about 'trusted-substituters' option in nix.conf man page", s);
```
eordano, line 226:
```suggestion
                             "More information about 'trusted-substituters' option can be found in the nix.conf man page or at \nhttps://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-trusted-substituters", s);
```

fricklerhandwerk: @eordano we [discussed this recently](https://github.com/NixOS/nix/issues/6672#issuecomment-1357976553) and decided not to add a URL, because that would be in there forever, while quite literally everything about that URL may change in the future. Only the man pages are guaranteed to be in sync with the code.

eordano, line 226:
```suggestion
                             "More information about 'trusted-substituters' option can be found in the nix.conf man page", s);
```

eordano: I've suggested a new change then, yes, I also felt strange about `stable` -- having to detect what channel the user is using could be complex.

rapenne-s: I tried to keep the wording short, I'm not convinced your suggestion adds any value.
Is my current wording incorrect in English? I'm not a native speaker, I can't tell.

fricklerhandwerk: In any case, you probably want to give instructions here, and those should be written in imperative.
```suggestion
                             "Run `man nix.conf` for more information on the `substituters` configuration option.", s);
```

@rapenne-s please add the detailed instructions to that section and link to it from the docs on the related options. (I don't really care where you put it, but `substituters` is the general concept, and the other things are more specific, so I suppose that is what people would stumble upon first when skimming around.)

tomberek (approved):

thufschmitt (approved):

PR #7462, opened 2022-12-14
Inform user instead of warning them when using a trusted substituter

rapenne-s:
When an user is trusted by nix and has accepted a substituter, there is
a Warning message in fuschia displayed at each use of this substituter
because it's trusted.

This change the behavior by downgrading the message at the info level,
only displayed with nix in verbose mode.

This is actually scaring people because it looks like a problem, while
it's just plain information about something they did.

### DIFF
--- a/src/libexpr/flake/config.cc
+++ b/src/libexpr/flake/config.cc
@@ -56,7 +56,7 @@ void ConfigFile::apply()
             auto tlname = get(trustedList, name);
             if (auto saved = tlname ? get(*tlname, valueS) : nullptr) {
                 trusted = *saved;
-                printMsg(lvlInfo, "Using saved setting for '%s = %s' from ~/.local/share/nix/trusted-settings.json.", name,valueS);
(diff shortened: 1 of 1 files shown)


src/libexpr/flake/config.cc
```diff
@@ -56,7 +56,7 @@ void ConfigFile::apply()
             auto tlname = get(trustedList, name);
             if (auto saved = tlname ? get(*tlname, valueS) : nullptr) {
                 trusted = *saved;
-                warn("Using saved setting for '%s = %s' from ~/.local/share/nix/trusted-settings.json.", name,valueS);
+                printMsg(lvlInfo, "Using saved setting for '%s = %s' from ~/.local/share/nix/trusted-settings.json.", name,valueS);
```
thufschmitt, line 59:
```suggestion
                printInfo("Using saved setting for '%s = %s' from ~/.local/share/nix/trusted-settings.json.", name,valueS);
```

thufschmitt (approved):

Issue #3898, opened 2020-08-05
nix-build yields SQLiteError due to failed constraint when inserting into NARs


terlar:
**Describe the bug**

Running on a CI pipeline with two parallel nix-build breaks with SQLiteError sporadically when using nix-2.4pre7805_984e5213.
(/nix/store/i188a271sni33gqxs0nwrkgnpljfhwwf-nix-2.4pre7805_984e5213.drv)

I see errors such as this:
```
$ nix-build -A eu-west-1.platform --argstr env dev
error (ignored): error: --- SQLiteError --- nix-build
executing SQLite statement 'insert or replace into NARs(cache, hashPart, namePart, url, compression, fileHash, fileSize, narHash, narSize, refs, deriver, sigs, ca, timestamp, present) values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, 1)': constraint failed (in '/home/jenkins/.cache/nix/binary-cache-v6.sqlite')
while evaluating the attribute 'paths' of the derivation 'cbpf-platform-env' at /nix/store/sp6h92qaz2rdppy43g0k36w1chp857ss-nixpkgs-0cfa467f8771ca585b3e122806c8337f5025bd92/pkgs/build-support/trivial-builders.nix:7:14:
while evaluating the attribute 'paths' of the derivation 'cbpf' at /nix/store/sp6h92qaz2rdppy43g0k36w1chp857ss-nixpkgs-0cfa467f8771ca585b3e122806c8337f5025bd92/pkgs/build-support/trivial-builders.nix:7:14:
while evaluating the attribute 'buildCommand' of the derivation 'cbpf-deps' at /nix/store/sp6h92qaz2rdppy43g0k36w1chp857ss-nixpkgs-0cfa467f8771ca585b3e122806c8337f5025bd92/pkgs/build-support/trivial-builders.nix:7:14:
while evaluating 'concatMapStrings' at /nix/store/sp6h92qaz2rdppy43g0k36w1chp857ss-nixpkgs-0cfa467f8771ca585b3e122806c8337f5025bd92/lib/strings.nix:31:25, called from /nix/store/sp6h92qaz2rdppy43g0k36w1chp857ss-nixpkgs-0cfa467f8771ca585b3e122806c8337f5025bd92/pkgs/build-support/trivial-builders.nix:329:9:
[...]

PR #7473, opened 2022-12-17
Improve sqlite error messages

roberth:
The exception did not include the detailed error message, losing essential information for understanding #3898. I'm also cherry-pick this into Nixpkgs.

Example message:

    warning: creating statement 'insert or rplace into NARs(cache, hashPart, namePart, url, compression, fileHash, fileSize, narHash, narSize, refs, deriver, sigs, ca, timestamp, present) values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, 1)': at offset 10: SQL logic error, near "rplace": syntax error (in '/tmp/nix-shell.grQ6f7/nix-test/tests/binary-cache/test-home/.cache/nix/binary-cache-v6.sqlite')

It's not the best example, but more important information will be in the message for e.g. a constraint violation.

I don't see why this specific error is printed as a warning, but that's for another PR.

### DIFF
--- a/src/libstore/sqlite.cc
+++ b/src/libstore/sqlite.cc
@@ -8,12 +8,15 @@
 
 namespace nix {
 
-SQLiteError::SQLiteError(const char *path, int errNo, int extendedErrNo, hintformat && hf)
-  : Error(""), path(path), errNo(errNo), extendedErrNo(extendedErrNo)
+SQLiteError::SQLiteError(const char *path, const char *errMsg, int errNo, int extendedErrNo, int offset, hintformat && hf)
+  : Error(""), path(path), errMsg(errMsg), errNo(errNo), extendedErrNo(extendedErrNo), offset(offset)
 {
-    err.msg = hintfmt("%s: %s (in '%s')",
+    auto offsetStr = (offset == -1) ? "" : "at offset " + std::to_string(offset) + ": ";
+    err.msg = hintfmt("%s: %s%s, %s (in '%s')",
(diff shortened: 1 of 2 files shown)


thufschmitt (approved):
Thanks!

I don't know how many times I've implemented something like that for debugging purposes, then thought that it would be worth upstreaming, and forgot about it… So it's great to have it at last!

abathur:
Just dot-connecting. I spotted a report on reddit with an install failure on:

```console
/nix/store/lsr79q5xqd9dv97wn87x12kzax8s8i1s-nix-2.13.2/bin/nix-store: symbol lookup error: /nix/store/lsr79q5xqd9dv97wn87x12kzax8s8i1s-nix-2.13.2/lib/libnixstore.so: undefined symbol: sqlite3_error_offset
/tmp/nix-binary-tarball-unpack.AOBpnqbDFf/unpack/nix-2.13.2-x86_64-linux/install: unable to register valid paths
```

https://www.reddit.com/r/Nix/comments/117m3nw/installation_error_on_ubuntu_2204_libnixstoreso/

It looks like `sqlite3_error_offset` was added early last year in https://www.sqlite.org/releaselog/3_38_0.html, so I assume it didn't make the ship for 22.04 :)

roberth:
I think the solution is to not query the sqlite library in an impure location such as `/usr`.
I don't know whether they are to blame for configuring their environment that way, or whether we should lock down our outputs more.

PR #7474, opened 2022-12-17
Make install command in documentation compatible with fish shell

doubledup:
See the description in [NixOS/nix.dev/pull/422](https://github.com/NixOS/nix.dev/pull/422) for more context.


edolstra:
Shouldn't we just use
```
curl -L https://nixos.org/nix/install | sh
```
since that works in both shells?

doubledup:
@edolstra That's much better, thanks! I've updated the PR and the description. I've also updated some other install commands I found, but left the release notes alone. Is it ok to update the release notes as well?

fricklerhandwerk (approved):

doubledup:
@fricklerhandwerk @edolstra The [nix.dev PR](https://github.com/NixOS/nix.dev/pull/422) has been merged, but this one seems blocked on some tests that haven't run yet. Is there anything I can do to get the tests running?

fricklerhandwerk:
I was on vacation, sorry for the delay and thanks for hanging tight. Looks good, I'll merge this.

doubledup:
No problem at all. Thanks for all your work on Nix!

zegelin:
This change has removed `--daemon` from the quick-start guide's invocation of `install.sh`.

I'm new to Nix/NixOS, yet it seems like a lot of common internet advice expects Nix to be installed in multi-user mode. Even various Nix wiki pages mention editing `/etc/nix/nix.conf` and restarting `nix-daemon.service` (eg: https://nixos.wiki/wiki/NixOS_on_ARM#Compiling_through_binfmt_QEMU, see the Note box for Arch Linux users).

That change also makes the paragraph following the install instructions make no sense:
> On Linux, `--daemon` can be omitted for a single-user install.

There is no `--daemon` in the invocation to omit!

This change seems to have also missed the https://nixos.org/manual/nix/stable/installation/installation page, which still uses the `bash <(curl -L https://nixos.org/nix/install) --daemon` style invocation.

PR #7478, opened 2022-12-19
libstore: Make sure that initNix has been called

roberth:
Prevent bugs like https://github.com/cachix/cachix/pull/477

### DIFF
--- a/src/libstore/globals.hh
+++ b/src/libstore/globals.hh
@@ -984,4 +984,12 @@ std::vector<Path> getUserConfigFiles();
 
 extern const std::string nixVersion;
 
+/* NB: This is not sufficient. You need to call initNix() */
+void initLibStore();
+
+/* It's important to initialize before doing _anything_, which is why we
+   call upon the programmer to handle this correctly. However, we only add
+   this in a key locations, so as not to litter the code. */
+void assertLibStoreInitialized();
+
 }
--- a/src/libexpr/tests/libexprtests.hh
+++ b/src/libexpr/tests/libexprtests.hh
@@ -12,6 +12,7 @@ namespace nix {
     class LibExprTest : public ::testing::Test {
         public:
             static void SetUpTestSuite() {
+                initLibStore();
                 initGC();
             }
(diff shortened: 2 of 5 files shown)


src/libstore/globals.hh
```diff
@@ -984,4 +984,12 @@ std::vector<Path> getUserConfigFiles();
...
 extern const std::string nixVersion;
 
+/* NB: This is not sufficient. You need to call initNix() */
+void initLibStore();
+
+/* It's important to initialize before doing _anything_, which is why we
+   call upon the programmer to handle this correctly. However, we only add
+   this in a key locations, so as not to litter the code. */
+void assertLibStoreInitialized();
+
```
roberth, line 994: I would have done this simply at the `initNix` level, but the library layering does not allow this.

Perhaps the common parts of initialization should be moved into `libutil`?

thufschmitt (approved):

fricklerhandwerk:

 - @roberth: call structure is a bit convoluted because a dependency inversion was required
 - @edolstra: looks ok
 - @roberth opened issue to resolve the exposed tech debt: #7502
 - assigned to @edolstra for merging

fricklerhandwerk:

- @fricklerhandwerk: please add the rationale to the commit message, otherwise it's not clear why the change was made 
    - @roberth: we're trying to prevent using the SQLite in the wrong way. if we don't load the config file, it will use the default settings
    - @thufschmitt: could we instead use the dirty static variable trick we have in other places, so the code is run whenever the library is loaded?
    - @edolstra: what we actually want is that the `Config` object is initialised, and that's what we should assert somewhere
        - @roberth: opened an [issue to follow up](https://github.com/NixOS/nix/issues/7502) and improve the factoring
            - @edolstra: this thing is not designed to be used for anybody else, this is where all the problems come from. so there is much to be improved
- decision: merge

Ma27:
FYI this appears to break the perl bindings that are used by e.g. Hydra. Fix is in #7705.

PR #7486, opened 2022-12-20
reword definition of "reference"

fricklerhandwerk:
this corrects the definition to make a clearer distinction between store paths and store objects. it's not the paths that have references to each other, but the store objects.

### DIFF
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -115,7 +115,7 @@
     then be built.
 
   - [reference]{#gloss-reference}\
-    A [store object] `O` is said to have a *reference* to a store object `P` if the [store path] of `P` appears in the contents of `O`.
+    A [store object] `O` is said to have a *reference* to a store object `P` if a [store path] to `P` appears in the contents of `O`.
     The *references* of a store object `O` are the set of store objects to which `O` has a reference.
 
     Store objects can refer to both other store objects and themselves.
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -138,7 +138,7 @@
     files could be missing. The command `nix-store -qR` prints out
     closures of store paths.
 
-    As an example, if the store object at path `P` contains a [reference]
+    As an example, if the [store object] at path `P` contains a [reference]
     to store object `Q`, then `Q` is in the closure of `P`. Further, if `Q`
     references `R` then `R` is also in the closure of `P`.
 
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -115,14 +115,12 @@
     then be built.
 
   - [reference]{#gloss-reference}\
-    A store path `P` is said to have a reference to a store path `Q` if
-    the store object at `P` contains the path `Q` somewhere. The
-    *references* of a store path are the set of store paths to which it
-    has a reference.
+    A [store path] `P` is said to have a *reference* to a store path `Q` if the string `Q` appears in the [store object] at `P`.
+    The *references* of a store path `P` are the set of store paths to which `P` has a reference.
 
-    A derivation can reference other derivations and sources (but not
-    output paths), whereas an output path only references other output
-    paths.
+    A [derivation] can reference other derivations and source files, but not [output path]s, whereas an output path can only reference other output paths.
+
+    [reference]: #gloss-reference
 
   - [reachable]{#gloss-reachable}\
     A store path `Q` is reachable from another store path `P` if `Q`
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -140,7 +140,7 @@
     closures of store paths.
 
     As an example, if the store object at path `P` contains a [reference]
-    to path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
+    to store object `Q`, then `Q` is in the closure of `P`. Further, if `Q`
     references `R` then `R` is also in the closure of `P`.
 
   - [output path]{#gloss-output-path}\
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -118,9 +118,9 @@
     A [store object] `O` is said to have a *reference* to a store object `P` if the [store path] of `P` appears in the contents of `O`.
     The *references* of a store object `O` are the set of store objects to which `O` has a reference.
 
-    Source files have no references.
-    A [store derivation] can only reference source files and other store derivations, including itself.
-    In contrast, a store object that was produced from a [derivation] can only reference other "derived" store objects.
+    Store objects can refer to both other store objects and themselves.
+    References from a store object to itself are called *self-references*.
+    References other than a self-reference must not form a cycle.
 
     [reference]: #gloss-reference
 
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -116,7 +116,6 @@
 
   - [reference]{#gloss-reference}\
     A [store object] `O` is said to have a *reference* to a store object `P` if a [store path] to `P` appears in the contents of `O`.
-    The *references* of a store object `O` are the set of store objects to which `O` has a reference.
 
     Store objects can refer to both other store objects and themselves.
     References from a store object to itself are called *self-references*.
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -115,10 +115,12 @@
     then be built.
 
   - [reference]{#gloss-reference}\
-    A [store path] `P` is said to have a *reference* to a store path `Q` if the string `Q` appears in the [store object] at `P`.
-    The *references* of a store path `P` are the set of store paths to which `P` has a reference.
+    A [store object] `O` is said to have a *reference* to a store object `P` if the [store path] of `P` appears in the contents of `O`.
+    The *references* of a store object `O` are the set of store objects to which `O` has a reference.
 
-    A [derivation] can reference other derivations and source files, but not [output path]s, whereas an output path can only reference other output paths.
+    Source files have no references.
+    A [store derivation] can only reference source files and other store derivations, including itself.
+    In contrast, a store object that was produced from a [derivation] can only reference other "derived" store objects.
 
     [reference]: #gloss-reference
 
@@ -137,7 +139,7 @@
     files could be missing. The command `nix-store -qR` prints out
     closures of store paths.
 
-    As an example, if the store object at path `P` contains a reference
+    As an example, if the store object at path `P` contains a [reference]
     to path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
     references `R` then `R` is also in the closure of `P`.
 
--- a/doc/manual/src/glossary.md
+++ b/doc/manual/src/glossary.md
@@ -139,7 +139,7 @@
     closures of store paths.
 
     As an example, if the [store object] at path `P` contains a [reference]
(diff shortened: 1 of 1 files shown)


Ericson2314:
Yes this isn't quite right. I would say:

1. Store paths don't have references. Store paths point to at most a single store object, and that's it. (or no object at all, a dangling reference, e.g. if you just mashed the keyboard for the hash part).

2. Store objects have references to other store objects.

3. Store objects' references are *usually* (but not always) computed by scanning the score object for (the hash parts of) (certain, drawn from a whitelist) store paths.

We might need to go through the manual with a comb and change "store path" to "store object" more widely.

doc/manual/src/glossary.md
```diff
@@ -95,14 +103,12 @@
     then be built.
 
   - [reference]{#gloss-reference}\
-    A store path `P` is said to have a reference to a store path `Q` if
-    the store object at `P` contains the path `Q` somewhere. The
-    *references* of a store path are the set of store paths to which it
-    has a reference.
+    A [store path] `P` is said to have a *reference* to a store path `Q` if the string `Q` appears in the [store object] at `P`.
```
fricklerhandwerk, line 106:
```suggestion
    A [store object] `O` is said to have a *reference* to a store object `P` if the store path of `P` appears in the contents of`O`.
```

@Ericson2314 @edolstra does that sound about right?

Ericson2314: This sounds good to me!

doc/manual/src/glossary.md
```diff
@@ -115,14 +115,14 @@
...
-    has a reference.
+    A [store object] `O` is said to have a *reference* to a store object `P` if the [store path] of `P` appears in the contents of `O`.
+    The *references* of a store object `O` are the set of store objects to which `O` has a reference.
 
-    A derivation can reference other derivations and sources (but not
-    output paths), whereas an output path only references other output
-    paths.
+    Source files have no references.
+    A [store derivation] can only reference source files and other store derivations, including itself.
+    In contrast, a store object that was produced from a [derivation] can only reference other "derived" store objects.
```
Ericson2314, line 123: I wouldn't get too much into the source file vs build result distinction here. It is not a very deep thing in Nix. One can depend on all sorts of store objects with e.g. `builtins.storePath` --- getting around these distinction.

> including itself

A derivation never references itself. The outputs of a derivation however can reference themselves.

fricklerhandwerk: So do you suggest removing the paragraph altogether?

Ericson2314: Heh I guess I didn't leave too much left. But maybe I can suggest and alternative.

Ericson2314:
```suggestion
    Store objects can refer to both other store objects and themselves.
    References from a store object to itself are called *self-references*.
    References other than self-reference are not allowed to form a cycle.
```

fricklerhandwerk: Maybe we can leave it out since that's essentially what will be in the architecture chapter on references anyway.

Ericson2314 (approved):
Looks good!

Ericson2314:
Merge away, @fricklerhandwerk!

doc/manual/src/glossary.md
```diff
@@ -115,14 +115,14 @@
     then be built.
 
   - [reference]{#gloss-reference}\
-    A store path `P` is said to have a reference to a store path `Q` if
-    the store object at `P` contains the path `Q` somewhere. The
-    *references* of a store path are the set of store paths to which it
-    has a reference.
+    A [store object] `O` is said to have a *reference* to a store object `P` if the [store path] of `P` appears in the contents of `O`.
```
edolstra, line 118: This is not quite correct. References are not store objects but store paths.

There is no such thing as "the store path of P" since store objects don't have a uniquely determined store path. (E.g. there can be multiple store paths with the same contents.)

edolstra, line 118: This is not quite correct. References are not store objects but store paths.

There is no such thing as "the store path of P" since store objects don't have a uniquely determined store path. (E.g. there can be multiple store paths with the same contents.)

doc/manual/src/glossary.md
```diff
@@ -115,14 +115,14 @@
...
+    The *references* of a store object `O` are the set of store objects to which `O` has a reference.
 
-    A derivation can reference other derivations and sources (but not
-    output paths), whereas an output path only references other output
-    paths.
+    Store objects can refer to both other store objects and themselves.
```
edolstra, line 121: Can a store object refer to itself if a store object doesn't know its own store path?

edolstra (changes requested):

edolstra:
I'm increasingly beginning to think that the store object / store path distinction is unhelpful and we should just drop "store object" from the docs. We've survived for 20 years without "store object" in the docs so we'll be fine.

doc/manual/src/glossary.md
Ericson2314: I would say store paths are you unique, and those are not the same store object. I don't think I am "right' and you are "wrong" though, I think these are both valid opinions as of today, and we decide which one we want to be correct and formalize it.

In other words, this is an interesting question, not something to be shoved underneath the rug!

doc/manual/src/glossary.md
Ericson2314: Yes.

- Some of the time does know its own store path, so there is no issue. For example, when the store object is mounted in the file system

- Some of the time it does not know its own store path, so self-references are made with a sentinel value. For example, when the store object is to be content-addressed and is being hashed.

Ericson2314:
@edolstra This is like saying "arrays vs pointers are confusing, let's add implicit conversions". Well, C is still quite confusing, and Rust, which does not do this conflation is not.

As I just wrote above in https://github.com/NixOS/nix/pull/7486#discussion_r1083328591, there are multiple interpretations here, and it behooves us to take the time to think a bit and decide on one, and then document it. Strictly defined semantics is good, and wishy-washy hand-waving is not.

Indeed, https://github.com/NixOS/nix/pull/3746 gets into these issues.

doc/manual/src/glossary.md
Ericson2314: Here is one way to fix it
```suggestion
    A [store object] `O` is said to have a *reference* to a store object `P` if a [store path] of `P` appears in the contents of `O`.
```

Ericson2314: here is another way
```suggestion
    Store objects can *refer* to other store objects by [store path].
    A [store object] `O` is said to have a *reference* `P` to some other store object if `P` appears in the contents of `O`.
```

Ericson2314: These are valid with both interpretations.

I do have a preference for not saying "store path of" even if the paths are unique, on the "things shouldn't know their own name" principle.

fricklerhandwerk: > store objects don't have a uniquely determined store path

> I would say store paths are unique, and [two store objects with same contents but different paths] are not the same store object.

@edolstra @Ericson2314 Good points, both. Let's determine and document the mental model we want users to apply. Intuitively I'm inclined towards the latter, because input addressing is a concept that's easy to explain and understand. It's the derivations that produce store objects and define their store paths. That different derivations may produce the same store object is incidental. Deduplicating store objects with the same contents and aliasing those under different paths appears to be an implementation detail that is not even necessarily a concern of Nix but that of the underlying persistence layer.

Until this is resolved, I would go with the following, because it's brief and not wrong w.r.t. the current implementation:

```
A [store object] `O` is said to have a *reference* to a store object `P` if a [store path] to `P` appears in the contents of `O`.
```

doc/manual/src/glossary.md
fricklerhandwerk: We should add these specifics to the relevant architecture documentation section.

fricklerhandwerk:
@edolstra I agree that this type of documentation does not perfectly match the implementation. It never did, and not forcing it to do so (yet) is deliberate: I think at this point it's more important to give users the right ideas how to think about what Nix is doing. Omitting details here and there is a (hopefully temporary) trade-off I'm trying to make very carefully. This is until we have the proper place to expand on them, in this case the architecture chapter. I will make the next PR in that direction soon™.

doc/manual/src/glossary.md
```diff
@@ -139,7 +139,7 @@
     files could be missing. The command `nix-store -qR` prints out
     closures of store paths.
 
-    As an example, if the store object at path `P` contains a reference
+    As an example, if the store object at path `P` contains a [reference]
     to path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
```
Ericson2314, line 143:
```suggestion
    to store path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
```

doc/manual/src/glossary.md
```diff
@@ -115,14 +115,14 @@
     then be built.
 
   - [reference]{#gloss-reference}\
-    A store path `P` is said to have a reference to a store path `Q` if
-    the store object at `P` contains the path `Q` somewhere. The
-    *references* of a store path are the set of store paths to which it
-    has a reference.
+    A [store object] `O` is said to have a *reference* to a store object `P` if a [store path] to `P` appears in the contents of `O`.
+    The *references* of a store object `O` are the set of store objects to which `O` has a reference.
```
Ericson2314, line 119: I sort of just want to drop this sentence. Refernces = set of references themselves is arguably better English than set of things referred to. But it is nice to still not talk about store paths as that is just one means of referencing.

The references of a store object are a lot like the string contexts of a string.

Ericson2314: The references of a store object are a lot like the string contexts of a string. I might also make sense to call them both references (string references, store object references; string contex, store object context). The glossary for reference could refer to both kinds of references.

fricklerhandwerk:
```suggestion
```

With pleasure. I also found it weird and barely helpful, but didn't want to change too much.

doc/manual/src/glossary.md
```diff
@@ -139,8 +138,8 @@
     files could be missing. The command `nix-store -qR` prints out
     closures of store paths.
 
-    As an example, if the store object at path `P` contains a reference
-    to path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
+    As an example, if the store object at path `P` contains a [reference]
```
fricklerhandwerk, line 141:
```suggestion
    As an example, if the [store object] at path `P` contains a [reference]
```

doc/manual/src/glossary.md
```diff
@@ -139,8 +138,8 @@
     files could be missing. The command `nix-store -qR` prints out
     closures of store paths.
 
-    As an example, if the store object at path `P` contains a reference
-    to path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
+    As an example, if the [store object] at path `P` contains a [reference]
+    to store object `Q`, then `Q` is in the closure of `P`. Further, if `Q`
```
fricklerhandwerk, line 142:
```suggestion
    As an example, if the store object at path `P` contains a reference
    to path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
```

fricklerhandwerk, line 142:
```suggestion
    to a store object at path `Q`, then `Q` is in the closure of `P`. Further, if `Q`
```

Ericson2314 (approved):
LGTM

There is more we can do about store path vs store object consistency in other adjacent sections, but that can just be a separate PR.

PR #7492, opened 2022-12-21
Preserve TMPDIR on macOS in 'nix shell'

### LINKED ISSUE #7491: nix unsets TMPDIR on macOS in subshell
**Describe the bug**

On macOS the Nix tools [unsets TMPDIR](https://github.com/NixOS/nix/blob/1437582ccd3c4af51f34e43a77df5c8622e24d6c/src/libmain/shared.cc#L229-L235) which means that a shell spawned with `nix shell` also has it unset. Then, if you use a tool (ctags in my example) that requires it to be set you'll get an error.

**Steps To Reproduce**

1. `echo $TMPDIR`
1. `nix shell`
2. `echo $TMPDIR`

**Expected behavior**

TMPDIR keeps a value when using `nix shell`

`nix-env (Nix) 2.11.1`

bouk:
See for context: https://github.com/NixOS/nix/issues/7491

Closes https://github.com/NixOS/nix/issues/7491

### DIFF
--- a/src/libmain/shared.cc
+++ b/src/libmain/shared.cc
@@ -35,6 +35,10 @@ namespace nix {
 
 char * * savedArgv;
 
+#if __APPLE__
+std::optional<std::string> savedTmpDir;
+#endif
+
 static bool gcWarning = true;
 
 void printGCWarning()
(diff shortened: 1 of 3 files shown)


thufschmitt (commented):
I think the fix is correct, but having to save tmpdir in a global variable like that is a bit ugly :/

@edolstra if I understand correctly, the problem with `TMPDIR` is only for the builders, right? Could we only unset it in the build loop then?

Ericson2314:

- @edolstra: agree that changing it in the builders would be better
- assigned to @thufschmitt

tomberek:

PR #7494, opened 2022-12-21
Move the error message to the bottom when showing traces

domenkozar:
Given `error.nix`:

```nix
with (import <nixpkgs> {});

let 
  a = b: git.foo;
in a 3
```

Before:

```
❯ /nix/store/rz6a733dfhhycsf81mc6nf8jh3h729n8-nix-2.13.0pre20221221_b1223e1/bin/nix-instantiate error.nix -I ~/dev --show-trace
error:
       … while evaluating call site

       at /home/domen/dev/nix/error.nix:5:4:

            4|   a = b: git.foo;
            5| in a 3
             |    ^
            6|

       … while calling 'a'

       at /home/domen/dev/nix/error.nix:4:7:

            3| let
            4|   a = b: git.foo;
             |       ^
            5| in a 3

       error: attribute 'foo' missing

       at /home/domen/dev/nix/error.nix:4:10:

            3| let
            4|   a = b: git.foo;
             |          ^
            5| in a 3
       Did you mean doc?

```

After:

```
❯ ./outputs/out/bin/nix-instantiate error.nix -I ~/dev --show-trace
error:
       … while evaluating call site

       at /home/domen/dev/nix/error.nix:5:4:

            4|   a = b: git.foo;
            5| in a 3
             |    ^
            6|

       … while calling 'a'

       at /home/domen/dev/nix/error.nix:4:7:

            3| let
            4|   a = b: git.foo;
             |       ^
            5| in a 3

       at /home/domen/dev/nix/error.nix:4:10:

            3| let
            4|   a = b: git.foo;
             |          ^
            5| in a 3

       error: attribute 'foo' missing

       Did you mean doc?
```


Ericson2314:
I'm on board with the general goal here, but I am not so sure about the inconsistency of going from a bunch of frames in "thing at" order to a final item in "at thing" order.

roberth:
Another problem is that it says `error: ` at the beginning, leading at least some people to believe that the error description is on the next line. Looking for the first error is a good practice, so you really can't blame them.
Example: https://discourse.nixos.org/t/error-in-the-left-operand-of-the-update-operator/29901

If we keep `error: ` at the start, we should at least start with something like "original error and most recent call appear last"

domenkozar:
I'd love if someone finishes this and ships it, it's a huge UX improvement.

Maybe @bburdette?

bburdette:
What about this?

- 'error:' at top replaced by 'an error occurred!'
- ...while-at pattern continues with '... error location:'
- 'error location:' makes clear that this is where the actual error is.
- error message (almost) at the end, so you can scroll to the bottom and reliably find it.

```
❯ ./outputs/out/bin/nix-instantiate error.nix -I ~/dev --show-trace
an error occurred!

       … while evaluating call site

       at /home/domen/dev/nix/error.nix:5:4:

            4|   a = b: git.foo;
            5| in a 3
             |    ^
            6|

       … while calling 'a'

       at /home/domen/dev/nix/error.nix:4:7:

            3| let
            4|   a = b: git.foo;
             |       ^
            5| in a 3

       … error location:

       at /home/domen/dev/nix/error.nix:4:10:

            3| let
            4|   a = b: git.foo;
             |          ^
            5| in a 3

       error: attribute 'foo' missing

       Did you mean doc?
```

Would need a bit of thinking through for the different verbosities - error, warn, debug, etc.

domenkozar:
Closing as not wanted

Issue #7338, opened 2022-11-23
Should path values be deprecated?


infinisil:
The Nix language has path expressions which evaluate to path values. I believe it might be good to deprecate path values and eventually make path expressions return string values instead.

Reasons why path values aren't great:
- Their store import semantics are implicit and non-obvious. This makes it really hard for especially new people to understand paths, this relates to https://github.com/NixOS/nix/issues/7327:
  - "${some/path}" imports the path into the store
  - "${toString some/path}" doesn't import it into the store
- `..` path components get normalised for no good reason (which can be wrong if symlinks are involved, which Nix doesn't check for):
  ```
  nix-repl> /does/not/exist                 
  /does/not/exist

  nix-repl> /does/not/exist + "/foo/../bar"
  /does/not/exist/bar
  ```
- Are unintuitive to work with due to trailing `/` getting normalised, see also https://github.com/NixOS/nix/issues/7301:
  ```
  nix-repl> /foo + "/" + "bar"
  /foobar

  nix-repl> /foo + ("/" + "bar")
  /foo/bar
  ```

Meanwhile, there's essentially no good use cases for path values:
- `builtins.path` can be used to explicitly import a path into the store, which would make it much more obvious when it happens and it can't happen accidentally.
- Builtins don't require path values, they work with strings too, and even more correctly due to the `..` not getting normalised away

Ignoring deprecation for now, I think ideally path expressions should return string values instead, so like an implicit `toString`. This then requires people to be very explicit about what gets imported into the store, either using `builtins.path` or some library function built on top (like `lib.cleanSource`).

See also https://github.com/NixOS/nix/issues/7335 and https://github.com/NixOS/nixpkgs/pull/200718, ping @roberth @fricklerhandwerk

edolstra:
While there are a few warts with path values, I don't see how that justifies removing them. How would you distinguish between the derivation attributes
```
builder = ./builder.sh;
```
vs
```
builder = "./builder.sh";
```

And if the idea is to keep the path syntax but not path values: that doesn't work for lazy trees, where `./builder.sh` is a tuple that points to the source tree accessor and the path relative to the root of that accessor. 

> builtins.path can be used to explicitly import a path into the store, which would make it much more obvious when it happens and it can't happen accidentally.

We're not going to prefix `builtins.path` to every file reference in Nixpkgs. Nix is a DSL for writing Nix packages concisely, and this change would make Nix packages much more annoying to write.

roberth:
[...]

PR #7498, opened 2022-12-23
refactor documentation of operators, document `+` for strings and paths

fricklerhandwerk:
- move annotations out of the column

  this form is easier to maintain, displays better because it needs less space, and allows
  for more details on each operator that requires explanations.

- document semantics of overloaded `+` operator

Related: https://github.com/NixOS/nix/issues/7338#issuecomment-1328229411


edolstra:
I'm not convinced that this format is better for readers than a table. Whether the table is less maintainable is not super important given that we rarely ever add operators.

For comparison, the Rust language reference also uses a table: https://doc.rust-lang.org/book/appendix-02-operators.html

fricklerhandwerk:
@edolstra now this is closer to how it's done in the Rust docs. I added explanations in external notes, such that the table is narrower and we still have enough space to add explanations where needed.

Issue #6752, opened 2022-07-02
I'm an untrusted user and nix asks me to set `extra-trusted-public-keys` from a flakes `nixConfig` and then ignores me


NobbZ:
**Describe the bug**

Given a flake with a `nixConfig` attribute, `nix` asks me to confirm all the settings, just to tell me that it ignores what I just did, as I am not a trusted user.

So why does `nix` bother me at all?

**Steps To Reproduce**

1. run `nix run github:viperML/dotfiles#resolve-march-native`
2. Accept the substitutor and the public key
3. See that "untrusted substituter" will be ignored

**Expected behavior**

Either adhere to my choice and use the substituters or do not bother me with questions.

**`nix-env --version` output**

```
$ nix --version
nix (Nix) 2.10.0pre20220629_b7eb4ac
```

NobbZ:
Perhaps some nix config would be nice?

```
flake-disallow-config = *
```

To generally assume "no" for all `nixConfig` leys

```
flake-disallow-config = extra-substituters substituters
```

To disallow only those 2?

Or allow whitelisting in `flake-allow-config` or how its called, rathaer than a black/white switch?

thufschmitt:
> Perhaps some nix config would be nice?

I could probably help, indeed. Although the main issue here I think is people getting utterly confused when they see that for the first time, and that wouldn't be enough for that.

Maybe what we could do is add the ability to query the daemon for the options it accepts, and only prompt the user for these. But that would also mean knowing which options are used only by the daemon, which we don't atm

thufschmitt:
Closing in favour of #6672 since they are a duplicate of each other

PR #545, opened 2015-05-25
Support non-bash shells in nix-shell.

gebner:
This adds support for non-bash shells in the interactive mode of nix-shell.  I have tested it with zsh and fish.  Usage with bash works exactly the same as before.

Since the stdenv/setup script requires bash, the user shell is executed as a command from bash, after PATH, etc. have been configured.

I have also set the correct environment variable for --pure support in zsh.

PR #7515, opened 2022-12-26
Add Store::isTrustedClient()

MatthewCroughan:
This PR extends the worker protocol in [`worker-protocol-hh`](https://github.com/NixOS/nix/blob/a4604f19284254ac98f19a13ff7c2216de7fe176/src/libstore/worker-protocol.hh) by adding a new operation (47) which permits a client to check if they are trusted. I have needed this for some scripts I'm trying to write which use Nix,  that need to check.

In addition, it also adds a new command `nix store check-trusted` which will make use of this addition. It will return exit code 0 if trusted, 1 if not trusted or 2 if an unexpected error occurs.

I have also added a new `checkInfo` function to [`doctor.cc`](https://github.com/NixOS/nix/blob/9a640afc1e63bed28926cb44fe85137fb7d2d2d1/src/nix/doctor.cc) which will print an inconsequential blue `[INFO]` telling the user whether they are trusted. This does not change the exit status of the `nix doctor` command.

image

The following Bash function is an example of a fallback mechanism I've implemented in Bash but do not have the skills to implement in C++.

```bash
function isTrustedUser {
  function writeDummyBuildLog {
    {
      echo 0x6378696e 0x00000000; # WORKER_MAGIC_1
      echo 0x0a010000 0x00000000; # PROTOCOL_VERSION
      echo 0x2d000000 0x00000000; # wopAddBuildlog (45)
      echo 0x26000000 0x00000000; # Length of Store Path (38)
      echo -n "$(echo -n aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-a.drv | xxd -p -c0)00000000000000000000"; # Store Path
    } \
    | xxd -p -r \
    | socat - UNIX-CONNECT:/nix/var/nix/daemon-socket/socket \
    | tr -d '\0'
  }
  x=$(writeDummyBuildLog)
  if (grep -aq "error:" <<< "$x") && (grep -aq "not privileged" <<< "$x")
  then
    echo "You are not a trusted user"
    return 1
  elif (grep -aq "error:" <<< "$x")
    echo "An unknown error occurred whilst checking if you are a trusted user"
    return 2
  else
    :
    return 0
  fi
}
isTrustedUser
```

Thanks a lot to @roberth who introduced me to the C++ codebase and answering a lot of questions. And thanks to @tomberek + @pkharvey + @helicalbytes for helping me understand the daemon protocol.

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -39,3 +39,9 @@
 
   * `man nix-store-<operation>` and `man nix-env-<operation>`
   * `nix-store --help --<operation>` and `nix-env --help --<operation>`.
+
+* Nix when used as a client now checks whether the store (the server) trusts the client.
+  (The store always had to check whether it trusts the client, but now the client is informed of the store's decision.)
+  This is useful for scripting interactions with (non-legacy-ssh) remote Nix stores.
+
+  `nix store ping` and `nix doctor` now display this information.
--- a/src/libstore/build/local-derivation-goal.cc
+++ b/src/libstore/build/local-derivation-goal.cc
@@ -1415,6 +1415,9 @@ struct RestrictedStore : public virtual RestrictedStoreConfig, public virtual Lo
 
     virtual void addBuildLog(const StorePath & path, std::string_view log) override
     { unsupported("addBuildLog"); }
+
+    std::optional<TrustedFlag> isTrustedClient() override
+    { return NotTrusted; }
 };
 
 
@@ -1467,7 +1470,7 @@ void LocalDerivationGoal::startDaemon()
                 FdSink to(remote.get());
                 try {
                     daemon::processConnection(store, from, to,
-                        daemon::NotTrusted, daemon::Recursive);
+                        NotTrusted, daemon::Recursive);
                     debug("terminated daemon connection");
                 } catch (SysError &) {
                     ignoreException();
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -1032,6 +1032,15 @@ void processConnection(
     if (GET_PROTOCOL_MINOR(clientVersion) >= 33)
         to << nixVersion;
 
+    if (GET_PROTOCOL_MINOR(clientVersion) >= 35) {
+        // We and the underlying store both need to trust the client for
+        // it to be trusted.
+        auto temp = trusted
+            ? store->isTrustedClient()
+            : std::optional { NotTrusted };
+        worker_proto::write(*store, to, temp);
+    }
+
     /* Send startup error messages to the client. */
     tunnelLogger->startWork();
 
--- a/src/libstore/dummy-store.cc
+++ b/src/libstore/dummy-store.cc
@@ -39,6 +39,14 @@ struct DummyStore : public virtual DummyStoreConfig, public virtual Store
         callback(nullptr);
     }
 
+    /**
+     * The dummy store is incapable of *not* trusting! :)
+     */
+    virtual std::optional<TrustedFlag> isTrustedClient() override
+    {
+        return Trusted;
+    }
+
     static std::set<std::string> uriSchemes() {
         return {"dummy"};
     }
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -42,6 +42,40 @@ void write(const Store & store, Sink & out, const StorePath & storePath)
 }
 
 
+std::optional<TrustedFlag> read(const Store & store, Source & from, Phantom<std::optional<TrustedFlag>> _)
+{
+    auto temp = readNum<uint8_t>(from);
+    switch (temp) {
+        case 0:
+            return std::nullopt;
+        case 1:
+            return { Trusted };
+        case 2:
+            return { NotTrusted };
+        default:
+            throw Error("Invalid trusted status from remote");
+    }
+}
+
+void write(const Store & store, Sink & out, const std::optional<TrustedFlag> & optTrusted)
+{
+    if (!optTrusted)
+        out << (uint8_t)0;
+    else {
+        switch (*optTrusted) {
+        case Trusted:
+            out << (uint8_t)1;
+            break;
+        case NotTrusted:
+            out << (uint8_t)2;
+            break;
+        default:
+            assert(false);
+        };
+    }
+}
+
+
 ContentAddress read(const Store & store, Source & from, Phantom<ContentAddress> _)
 {
     return parseContentAddress(readString(from));
@@ -224,6 +258,13 @@ void RemoteStore::initConnection(Connection & conn)
         if (GET_PROTOCOL_MINOR(conn.daemonVersion) >= 33) {
             conn.to.flush();
             conn.daemonNixVersion = readString(conn.from);
+        }
+
+        if (GET_PROTOCOL_MINOR(conn.daemonVersion) >= 35) {
+            conn.remoteTrustsUs = worker_proto::read(*this, conn.from, Phantom<std::optional<TrustedFlag>> {});
+        } else {
+            // We don't know the answer; protocol to old.
+            conn.remoteTrustsUs = std::nullopt;
         }
 
         auto ex = conn.processStderr();
@@ -1082,6 +1123,11 @@ unsigned int RemoteStore::getProtocol()
     return conn->daemonVersion;
 }
 
+std::optional<TrustedFlag> RemoteStore::isTrustedClient()
+{
+    auto conn(getConnection());
+    return conn->remoteTrustsUs;
+}
 
 void RemoteStore::flushBadConnections()
 {
--- a/src/libstore/store-api.hh
+++ b/src/libstore/store-api.hh
@@ -89,6 +89,7 @@ const uint32_t exportMagic = 0x4558494e;
 
 
 enum BuildMode { bmNormal, bmRepair, bmCheck };
+enum TrustedFlag : bool { NotTrusted = false, Trusted = true };
 
 struct BuildResult;
 
@@ -814,6 +815,17 @@ public:
     {
         return 0;
     };
+
+    /**
+     * @return/ whether store trusts *us*.
+     *
+     * `std::nullopt` means we do not know.
+     *
+     * @note This is the opposite of the StoreConfig::isTrusted
+     * store setting. That is about whether *we* trust the store.
+     */
+    virtual std::optional<TrustedFlag> isTrustedClient() = 0;
+
 
     virtual Path toRealPath(const Path & storePath)
     {
--- a/src/nix/doctor.cc
+++ b/src/nix/doctor.cc
@@ -33,6 +33,10 @@ bool checkFail(const std::string & msg) {
     return false;
 }
 
+void checkInfo(const std::string & msg) {
+    notice(ANSI_BLUE "[INFO] " ANSI_NORMAL + msg);
+}
+
 }
 
 struct CmdDoctor : StoreCommand
@@ -63,6 +67,7 @@ struct CmdDoctor : StoreCommand
             success &= checkProfileRoots(store);
         }
         success &= checkStoreProtocol(store->getProtocol());
+        checkTrustedUser(store);
 
         if (!success)
             throw Exit(2);
@@ -137,6 +142,14 @@ struct CmdDoctor : StoreCommand
         }
 
         return checkPass("Client protocol matches store protocol.");
+    }
+
+    void checkTrustedUser(ref<Store> store)
+    {
+        std::string_view trusted = store->isTrustedClient()
+            ? "trusted"
+            : "not trusted";
+        checkInfo(fmt("You are %s by store uri: %s", trusted, store->getUri()));
     }
 };
 
--- a/tests/remote-store.sh
+++ b/tests/remote-store.sh
@@ -5,7 +5,18 @@ clearStore
 # Ensure "fake ssh" remote store works just as legacy fake ssh would.
 nix --store ssh-ng://localhost?remote-store=$TEST_ROOT/other-store doctor
 
+# Ensure that store ping trusted works with ssh-ng://
+nix --store ssh-ng://localhost?remote-store=$TEST_ROOT/other-store store ping --json | jq -e '.trusted'
+
 startDaemon
+
+if isDaemonNewer "2.15pre0"; then
+    # Ensure that ping works trusted with new daemon
+    nix store ping --json | jq -e '.trusted'
+else
+    # And the the field is absent with the old daemon
+    nix store ping --json | jq -e 'has("trusted") | not'
+fi
 
 # Test import-from-derivation through the daemon.
 [[ $(nix eval --impure --raw --expr '
--- a/src/libstore/daemon.hh
+++ b/src/libstore/daemon.hh
@@ -6,7 +6,6 @@
 
 namespace nix::daemon {
 
-enum TrustedFlag : bool { NotTrusted = false, Trusted = true };
 enum RecursiveFlag : bool { NotRecursive = false, Recursive = true };
 
 void processConnection(
--- a/src/libstore/http-binary-cache-store.cc
+++ b/src/libstore/http-binary-cache-store.cc
@@ -194,6 +194,18 @@ protected:
             }});
     }
 
+    /**
+     * This isn't actually necessary read only. We support "upsert" now, so we
+     * have a notion of authentication via HTTP POST/PUT.
+     *
+     * For now, we conservatively say we don't know.
+     *
+     * \todo try to expose our HTTP authentication status.
+     */
+    std::optional<TrustedFlag> isTrustedClient() override
+    {
+        return std::nullopt;
+    }
 };
 
 static RegisterStoreImplementation<HttpBinaryCacheStore, HttpBinaryCacheStoreConfig> regHttpBinaryCacheStore;
--- a/src/libstore/legacy-ssh-store.cc
+++ b/src/libstore/legacy-ssh-store.cc
@@ -389,6 +389,15 @@ public:
         return conn->remoteVersion;
     }
 
+    /**
+     * The legacy ssh protocol doesn't support checking for trusted-user.
+     * Try using ssh-ng:// instead if you want to know.
+     */
+    std::optional<TrustedFlag> isTrustedClient() override
+    {
+        return std::nullopt;
+    }
+
     void queryRealisationUncached(const DrvOutput &,
         Callback<std::shared_ptr<const Realisation>> callback) noexcept override
     // TODO: Implement
--- a/src/libstore/local-binary-cache-store.cc
+++ b/src/libstore/local-binary-cache-store.cc
@@ -95,6 +95,10 @@ protected:
         return paths;
     }
 
+    std::optional<TrustedFlag> isTrustedClient() override
+    {
+        return Trusted;
+    }
 };
 
 void LocalBinaryCacheStore::init()
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1685,6 +1685,11 @@ unsigned int LocalStore::getProtocol()
     return PROTOCOL_VERSION;
 }
 
+std::optional<TrustedFlag> LocalStore::isTrustedClient()
+{
+    return Trusted;
+}
+
 
 #if defined(FS_IOC_SETFLAGS) && defined(FS_IOC_GETFLAGS) && defined(FS_IMMUTABLE_FL)
 
--- a/src/libstore/local-store.hh
+++ b/src/libstore/local-store.hh
@@ -204,6 +204,8 @@ public:
 
     unsigned int getProtocol() override;
 
+    std::optional<TrustedFlag> isTrustedClient() override;
+
     void vacuumDB();
 
     void repairPath(const StorePath & path) override;
--- a/src/libstore/remote-store.hh
+++ b/src/libstore/remote-store.hh
@@ -144,6 +144,8 @@ public:
 
     unsigned int getProtocol() override;
 
+    std::optional<TrustedFlag> isTrustedClient() override;
+
     void flushBadConnections();
 
     struct Connection
@@ -151,6 +153,7 @@ public:
         FdSink to;
         FdSource from;
         unsigned int daemonVersion;
+        std::optional<TrustedFlag> remoteTrustsUs;
         std::optional<std::string> daemonNixVersion;
         std::chrono::time_point<std::chrono::steady_clock> startTime;
 
--- a/src/libstore/s3-binary-cache-store.cc
+++ b/src/libstore/s3-binary-cache-store.cc
@@ -509,6 +509,16 @@ struct S3BinaryCacheStoreImpl : virtual S3BinaryCacheStoreConfig, public virtual
         return paths;
     }
 
+    /**
+     * For now, we conservatively say we don't know.
+     *
+     * \todo try to expose our S3 authentication status.
+     */
+    std::optional<TrustedFlag> isTrustedClient() override
+    {
+        return std::nullopt;
+    }
+
     static std::set<std::string> uriSchemes() { return {"s3"}; }
 
 };
--- a/src/libstore/worker-protocol.hh
+++ b/src/libstore/worker-protocol.hh
@@ -10,7 +10,7 @@ namespace nix {
 #define WORKER_MAGIC_1 0x6e697863
 #define WORKER_MAGIC_2 0x6478696f
 
-#define PROTOCOL_VERSION (1 << 8 | 34)
+#define PROTOCOL_VERSION (1 << 8 | 35)
 #define GET_PROTOCOL_MAJOR(x) ((x) & 0xff00)
 #define GET_PROTOCOL_MINOR(x) ((x) & 0x00ff)
 
@@ -103,6 +103,7 @@ MAKE_WORKER_PROTO(, DerivedPath);
 MAKE_WORKER_PROTO(, Realisation);
 MAKE_WORKER_PROTO(, DrvOutput);
 MAKE_WORKER_PROTO(, BuildResult);
+MAKE_WORKER_PROTO(, std::optional<TrustedFlag>);
 
 MAKE_WORKER_PROTO(template<typename T>, std::vector<T>);
 MAKE_WORKER_PROTO(template<typename T>, std::set<T>);
--- a/src/nix/ping-store.cc
+++ b/src/nix/ping-store.cc
@@ -28,15 +28,20 @@ struct CmdPingStore : StoreCommand, MixJSON
             store->connect();
             if (auto version = store->getVersion())
                 notice("Version: %s", *version);
+            if (auto trusted = store->isTrustedClient())
+                notice("Trusted: %s", *trusted);
         } else {
             nlohmann::json res;
             Finally printRes([&]() {
                 logger->cout("%s", res);
             });
+
             res["url"] = store->getUri();
             store->connect();
             if (auto version = store->getVersion())
                 res["version"] = *version;
+            if (auto trusted = store->isTrustedClient())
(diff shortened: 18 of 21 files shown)


MatthewCroughan:
CI fails because it is using the old daemon protocol version, this is an expected failure.

NobbZ:
This is a good first step into a solution for #6752/#6672.

Can't say much codewise though.

MatthewCroughan:
This PR will also improve https://github.com/NixOS/nix/pull/7461

src/nix/doctor.cc
```diff
@@ -130,6 +135,15 @@ struct CmdDoctor : StoreCommand
 
         return checkPass("Client protocol matches store protocol.");
     }
+
+    void checkTrustedUser(ref<Store> store)
+    {
+        bool trusted = store->checkTrustedUser();
```
max-privatevoid, line 141: This will throw an error if the remote store is too old, possibly prematurely terminating the doctor, which feels a little surprising. I wouldn't call it a dealbreaker though.

max-privatevoid (changes requested):

src/libstore/dummy-store.cc
```diff
@@ -56,6 +56,11 @@ struct DummyStore : public virtual DummyStoreConfig, public virtual Store
     void queryRealisationUncached(const DrvOutput &,
         Callback<std::shared_ptr<const Realisation>> callback) noexcept override
     { callback(nullptr); }
+
+    bool checkTrustedUser() override
+    {
+        throw Error("You cannot query trusted-user on a dummy store");
```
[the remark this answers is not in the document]
MatthewCroughan: On the advice of @roberth I decided not to do this. If I was going to do this, I'd just make the default implementation in the `virtual` function in `store-api.cc` say "This store type is not supported" for example.

src/nix/doctor.cc
MatthewCroughan: It doesn't change the exit status, and I think it's the correct way to fail. Failure is not a form of `[INFO]`, only the successful case of figuring out if we're trusted or not is. Anything else should just be printed as-is, which is how it works at the moment.

MatthewCroughan: Apparently you need to bit shift by 8 bytes, this was incredibly confusing to me because:
- `GET_PROTOCOL_MAJOR(conn->daemonVersion)` will return `256`
- `GET_PROTOCOL_MAJOR(conn->daemonVersion) >> 8` will return `1`

Anyway.. I've implemented it now as follows.

```diff
diff --git a/src/libstore/remote-store.cc b/src/libstore/remote-store.cc
index 154afa726..5cfd432dd 100644
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -1092,7 +1092,8 @@ bool RemoteStore::checkTrustedUser()
         conn.processStderr();
         return readInt(conn->from);
     } else {
-      throw Error("daemon protocol %d is too old (< 2.13) to ask whether we're a trusted user",
+      throw Error("daemon protocol %d.%d is too old (< 2.13) to ask whether we're a trusted user",
+          GET_PROTOCOL_MAJOR(conn->daemonVersion) >> 8,
           GET_PROTOCOL_MINOR(conn->daemonVersion));
     }
 }
```

src/libstore/remote-store.cc
```diff
@@ -1084,6 +1084,18 @@ unsigned int RemoteStore::getProtocol()
...
+bool RemoteStore::checkTrustedUser()
+{
+    auto conn(getConnection());
+    if (GET_PROTOCOL_MINOR(conn->daemonVersion) >= 35) {
+        conn->to << wopCheckTrustedUser;
+        conn.processStderr();
+        return readInt(conn->from);
+    } else {
+      throw Error("daemon protocol %d is too old (< 2.13) to ask whether we're a trusted user",
+          GET_PROTOCOL_MINOR(conn->daemonVersion));
```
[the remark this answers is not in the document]
MatthewCroughan: Apparently you need to bit shift by 8 bytes to read `GET_PROTOCOL_MAJOR`. This was incredibly confusing to me because:
- `GET_PROTOCOL_MAJOR(conn->daemonVersion)` will return `256`
- `GET_PROTOCOL_MAJOR(conn->daemonVersion) >> 8` will return `1`

Anyway.. I've implemented it now as follows.

```diff
diff --git a/src/libstore/remote-store.cc b/src/libstore/remote-store.cc
index 154afa726..5cfd432dd 100644
--- a/src/libstore/remote-store.cc
+++ b/src/libstore/remote-store.cc
@@ -1092,7 +1092,8 @@ bool RemoteStore::checkTrustedUser()
         conn.processStderr();
         return readInt(conn->from);
     } else {
-      throw Error("daemon protocol %d is too old (< 2.13) to ask whether we're a trusted user",
+      throw Error("daemon protocol %d.%d is too old (< 2.13) to ask whether we're a trusted user",
+          GET_PROTOCOL_MAJOR(conn->daemonVersion) >> 8,
           GET_PROTOCOL_MINOR(conn->daemonVersion));
     }
 }
```

MatthewCroughan:
I somehow accidentally removed the review request for @edolstra @fricklerhandwerk and @thufschmitt just by requesting a re-review from @max-privatevoid, this seems to be a bug in the way reviews are handled on Github.

src/libstore/dummy-store.cc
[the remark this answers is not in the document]
max-privatevoid: A default implementation was what I had in mind, yes. My bad for being unclear.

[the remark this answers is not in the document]
MatthewCroughan: @roberth seems not to like default implementations, and I also think it's a good idea to avoid default implementations, generally. It's a small nitpick and something that seems to vary between programmers.

MatthewCroughan:
I marked the PR as a draft, then unmarked it as a draft. It seems that the GitHub review bug is still in effect.

src/libstore/remote-store.cc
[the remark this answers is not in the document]
roberth: https://github.com/NixOS/nix/pull/7523

max-privatevoid (approved):

MatthewCroughan:
Just pushed to resolve conflicts with the release notes.

src/libstore/daemon.cc
```diff
@@ -970,6 +970,14 @@ static void performOp(TunnelLogger * logger, ref<Store> store,
         break;
     }
 
+    case wopCheckTrustedUser: {
+        logger->startWork();
+        bool result = trusted && store->checkTrustedUser();
```
roberth, line 975:
```suggestion
    case wopIsTrustedUser: {
        logger->startWork();
        bool result = trusted && store->isTrustedUser();
```

These two symbols are only about retrieving the information, but not necessarily acting upon it, so it's not really a check.
The `nix store check-trusted` command does not need renaming, because returning an exit code is more of a side effect.

src/libstore/build/local-derivation-goal.cc
```diff
@@ -1464,6 +1464,11 @@ struct RestrictedStore : public virtual RestrictedStoreConfig, public virtual Lo
 
     virtual void addBuildLog(const StorePath & path, std::string_view log) override
     { unsupported("addBuildLog"); }
+
+    bool checkTrustedUser() override
+    {
+        throw Error("You cannot query trusted-user with recursive-nix");
```
roberth, line 1470: It may make sense to call recursive nix not trusted.

MatthewCroughan: In which case, should we just return false?

src/libstore/daemon.cc
MatthewCroughan: I've kept the local function name `checkTrustedUser()` in check-trusted.cc and doctor.cc, but otherwise each store implementation to `isTrustedUser`. Thanks for teaching me more about naming conventions!

src/libstore/store-api.hh
```diff
@@ -635,6 +635,9 @@ public:
         return 0;
     };
 
+    /* Check if the user making the connection is a trusted-user. */
+    virtual bool isTrustedUser() = 0;
```
thufschmitt, line 639: I think it's fine to define this to `unsupported` by default like we do for a bunch of other methods and remove all the custom implementations for the stores that don't support it:

```suggestion
    virtual bool isTrustedUser()
    { unsupported("isTrustedUser"); }
```

src/libstore/remote-store.cc
```diff
@@ -1084,6 +1084,19 @@ unsigned int RemoteStore::getProtocol()
...
+{
+    auto conn(getConnection());
+    if (GET_PROTOCOL_MINOR(conn->daemonVersion) >= 35) {
+        conn->to << wopIsTrustedUser;
+        conn.processStderr();
+        return readInt(conn->from);
+    } else {
+      throw Error("daemon protocol %d.%d is too old (< 2.13) to ask whether we're a trusted user",
+          (GET_PROTOCOL_MAJOR(conn->daemonVersion) >> 8),
+          GET_PROTOCOL_MINOR(conn->daemonVersion));
```
thufschmitt, line 1097: I actually don't think that printing the daemon protocol version is useful here since it's mostly an internal detail. It would be better to point to the minimal working Nix version (2.13 in that instance).

src/nix/doctor.cc
```diff
@@ -130,6 +135,15 @@ struct CmdDoctor : StoreCommand
 
         return checkPass("Client protocol matches store protocol.");
     }
+
+    void checkTrustedUser(ref<Store> store)
+    {
+        bool trusted = store->isTrustedUser();
+        if (trusted)
+            checkInfo("You are a trusted-user on store uri: " + store->getUri());
```
thufschmitt, line 143: Do we need to print the store uri here? It doesn't seem very useful (or if we want to print it, we could have it as its own line somewhere else)

src/nix/store-check-trusted.cc
```diff
@@ -0,0 +1,61 @@
...
+    {
+        bool trusted = store->isTrustedUser();
+        if (!trusted) {
+            return checkFail("You are not a trusted-user on store uri: " + store->getUri());
+        }
+        return checkPass("You are a trusted-user on store uri: " + store->getUri());
+    }
+};
+
+static auto rCmdCheckTrusted = registerCommand2<CmdCheckTrusted>({"store", "check-trusted"});
```
thufschmitt, line 61:
```suggestion
static auto rCmdCheckTrusted = registerCommand2<CmdCheckTrusted>({"store", "check-trusted-user"});
```

`trusted` has a bunch of different meanings in Nix, better be explicit

src/nix/store-check-trusted.cc
```diff
@@ -0,0 +1,61 @@
...
+struct CmdCheckTrusted : StoreCommand
+{
+    bool success = true;
+
+    Category category() override { return catUtility; }
+
+    std::string description() override
+    {
+        return "ask a Nix store if you are a trusted-user, return exit code 0 if trusted, 1 if not trusted, 2 if unexpected error";
+    }
```
thufschmitt, line 39: Better keep the description short and add some proper doc of how this works, something like:

```suggestion
    std::string description() override
    {
        return "check whether you are a trusted-user";
    }
    
    std::string doc() override
    {
          return                                                                                                                                     
              #include "store-check-trusted.md"                                                                                                                     
              ;    
    }
```

Where `store-check-trusted.md` would contain some more docs for the command (similar to https://github.com/NixOS/nix/blob/master/src/nix/ping-store.md for example)

src/nix/store-check-trusted.cc
```diff
@@ -0,0 +1,61 @@
...
+
+    void run(ref<Store> store) override
+    {
+        logger->log("Running checks against store uri: " + store->getUri());
+        success &= checkTrustedUser(store);
```
thufschmitt, line 44:
```suggestion
        bool success = checkTrustedUser(store);
```

Unless there's a good reason to make it a class variable, I'd rather keep it local as it makes things much more readable

src/nix/store-check-trusted.cc
```diff
@@ -0,0 +1,61 @@
...
+    if (isatty(fileno(stdin)))
+      notice(ANSI_GREEN "[PASS] " ANSI_NORMAL + msg);
+    return true;
+}
+
+bool checkFail(const std::string & msg) {
+    if (isatty(fileno(stdin)))
+      notice(ANSI_RED "[FAIL] " ANSI_NORMAL + msg);
+    return false;
+}
```
thufschmitt, line 26: I don't think these colors bring a lot here. This is low-level plumbing, we should probably stick to the most sober solution

thufschmitt (changes requested):
Thanks, that's a useful command to have I think.

Left some suggestions on the code and docs, but that looks good overall.

(That also makes me realize with horror that we don't have any test for the "non-trusted-user" case :grimacing: )

src/libstore/remote-store.cc
MatthewCroughan: I also felt that this was unnecessary, but just copied the wisdom/pattern from other parts of the code. I'll remove it.

src/nix/doctor.cc
MatthewCroughan: I think it's useful for debug, especially in doctor. If you do `nix doctor` and my store is a remote store, I want to know. It helps me debug my configuration. Granted this could also be done by increasing Nix verbosity with  `-v`. My preference is to keep printing it.

src/nix/store-check-trusted.cc
MatthewCroughan: Then we should be removing that in another PR, for `nix doctor` itself. But for this PR to remain a simple addition, I'm not going to touch the existing code/pattern that doctor.cc implements.

src/nix/store-check-trusted.cc
MatthewCroughan: This is a pattern I copied from doctor.cc, nothing intentional.

src/nix/store-check-trusted.cc
MatthewCroughan: > proper doc of how this works

The proper doc of how this works is that it returns 1, 0 or 2. I don't think it's possible to make a dedicated document, unless you want it to describe the implementation details, which I think would not be suitable.

MatthewCroughan: Are there other commands with similarly very short and possibly useless manual pages? I think this would actually hide the information? `nix doctor` does not have its own man page, for the same reason, it's too small to document.

src/nix/store-check-trusted.cc
thufschmitt: No, I think that's fine for `doctor` which is intended as a higher-level “check for obvious configuration issues” command. I just think we don't need it in this command

src/nix/store-check-trusted.cc
MatthewCroughan: What other meanings does `trusted` have in Nix? And is this not disambiguated already by the fact that it is a subcommand of `store`?

thufschmitt: The three that come to my mind are:

- A path is trusted if it has a valid signature or has been built locally
- A public key is trusted or not
- A substituter can also be marked as [trusted](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-trusted-substituters) 

None of these would _totally_ make sense here (except maybe the third one), but they are closely related to the store and I can imagine them being misleading for a user

src/nix/store-check-trusted.cc
MatthewCroughan: Ah, I misread the source file name, I thought the code highlighted was from doctor.cc

In this case, it's not displayed unless your input `isatty`. This means it handles the case of scripting where you don't want any stdout, only an exit code, as well as also telling the user whether they passed or failed trust checking.

It is only in the interactive case where the input is a tty, where you would want any output from the command at all. So what would you rather it print? The same text without the pass/fail context text prepended?

src/nix/store-check-trusted.cc
thufschmitt: > The proper doc of how this works is that it returns 1, 0 or 2. I don't think it's possible to make a dedicated document, unless you want it to describe the implementation details, which I think would not be suitable.

Fair, a separate document might be overkill. But keeping the description short and moving the exact behaviour separate (even a simple one like here) is important for things like `nix --help` (which will list all the commands with their description).

src/nix/store-check-trusted.cc
MatthewCroughan: That makes sense, I'll change it.

src/libstore/store-api.hh
MatthewCroughan: That definitely makes it a smaller diff too.

src/nix/store-check-trusted.cc
MatthewCroughan: After some thinking, I think I agree now. I was just copying Nix doctor and liked the pretty colors. But it's not sober, as you mention.

src/nix/store-check-trusted.cc
MatthewCroughan: After breaking it out, I found that showing an example with `--store` is beneficial anyway and have created the md file.

src/nix/store-check-trusted.cc
MatthewCroughan: I've removed the color and fail/pass text.

src/libstore/store-api.hh
MatthewCroughan: Done. I've left the implementation for legacy ssh though, since it's more helpful to print it.

MatthewCroughan:
@thufschmitt I believe I've adjusted the code to meet all of your suggestions, except for `nix doctor` printing the Store URI, which I think is quite helpful. I've re-requested a review from you on GitHub, sadly this has the unintended side-effect of removing @edolstra due to what seems like either repo misconfiguration or a GitHub bug. Could you please add him back as a reviewer if he's required to get this merged, after you do your next review? Thank you!

src/libstore/build/local-derivation-goal.cc
MatthewCroughan: I've taken the suggestion from @thufschmitt to include a default implementation of `unsupported("foo")` instead. If we should return false, that is something we can do in another PR, if this gets merged soon.

roberth:
> add him back as a reviewer if he's required to get this merged

This isn't required if we follow the Nix team "project board" process that we've started to work with. Technically we should approve the idea first, but 3/5 Nix team members have already been involved, so we'll probably move this to final review and assign a reviewer in the next meeting.

Question for the team: is the process always required?

src/libstore/build/local-derivation-goal.cc
roberth: > just return false?

Yes.

MatthewCroughan: I've returned `false` for the case of recursive nix now.

src/libstore/store-api.hh
thufschmitt: @roberth since the default behaviour is to fail loudly, I think it's OK because it's a conservative approach, and we can always refine the implementation after the fact if we realize that we've been missing one

doc/manual/src/release-notes/rl-next.md
```diff
@@ -21,3 +21,17 @@
 * Error traces have been reworked to provide detailed explanations and more
   accurate error locations. A short excerpt of the trace is now shown by
   default when an error occurs.
+
+* On Linux, `nix develop` now sets the
+  [*personality*](https://man7.org/linux/man-pages/man2/personality.2.html)
+  for the development shell in the same way as the actual build of the
+  derivation. This makes shells for `i686-linux` derivations work
+  correctly on `x86_64-linux`.
```
thufschmitt, line 29: Merge failure? :upside_down_face:

doc/manual/src/release-notes/rl-next.md
```diff
@@ -21,3 +21,17 @@
...
+
+* New command: `nix store check-trusted-user`, which checks if you are a
+  trusted-user, returning 1 or 0. This is useful for scripting interactions with
+  remote Nix stores.
+
+* `nix doctor` now prints `[INFO]` instead of just `[PASS]` or `[FAIL]` and will
+  use the `nix store check-trusted-user` function to print whether you are a
+  trusted-user or not.
```
thufschmitt, line 37: I find this point a bit hard to read (and technically misleading since `nix doctor` doesn't really call `nix store check-trusted-user`). What about simply:

```suggestion

* `nix doctor` now displays whether you are a trusted user or not.
```

thufschmitt (changes requested):
Small glitch in the release notes, but that's looking good otherwise :+1: 

Question for the crowd (@NixOS/nix-team): Should we follow #5930 and feature-gate this for some time?

fricklerhandwerk:
@roberth this needs discussion by the team and is out of scope here. Opened https://github.com/NixOS/nix/pull/7579 for that purpose.
@thufschmitt I'd be in favor of making it experimental, for the exact reason stated in the relevant PR. It would give us at least some escape hatch if a substantial change is needed. For this specific feature, while I'm on board with the use case it unblocks, I'd still like to be cautious with growing the API surface.

<details><summary>Side notes</summary>

Both cases constitute additional [red tape](https://en.m.wikipedia.org/wiki/Red_tape). We want to make the contribution process easier, but also be much more deliberate about the overall design. Right now, due to many design decisions and strategies lacking, these may appear as conflicting goals; we already had multiple discussions over adding features where the use case was deemed worthy of support, but the specific change seemed too risky in terms of keeping options for the future. As a result, progress on these issues is stalling. Examples: https://github.com/NixOS/nix/pull/7188#issuecomment-1377200466, https://github.com/NixOS/nix/pull/545#issuecomment-1377208634.

I think the team should strive towards providing clear (written) guidance on how to approach new feature development in order to make it more likely to succeed.

</details>

doc/manual/src/release-notes/rl-next.md
MatthewCroughan: Derp

doc/manual/src/release-notes/rl-next.md
MatthewCroughan: I like that more. Fixed.

doc/manual/src/release-notes/rl-next.md
MatthewCroughan: Fixed

MatthewCroughan:
I've rebased this PR on 2.14 after the Nix 2.13 release yesterday.

tests/remote-store.sh
```diff
@@ -5,8 +5,16 @@ clearStore
 # Ensure "fake ssh" remote store works just as legacy fake ssh would.
 nix --store ssh-ng://localhost?remote-store=$TEST_ROOT/other-store doctor
 
+# Ensure that check-trusted-user works with ssh-ng://
+nix --store ssh-ng://localhost?remote-store=$TEST_ROOT/other-store store check-trusted-user
+
 startDaemon
 
+# Ensure that check-trusted-user works with daemon
+if isDaemonNewer "2.13pre0"; then
```
Ericson2314, line 14: This can be updated, but fine if you don't bother fixing this until we are about to merge so you aren't stuck doing it again.

MatthewCroughan: This means the daemon needs to be newer than 2.13, so it is the correct value. When 2.13 was released, this was bumped from 2.12 to 2.13.

Ericson2314:
@MatthewCroughan Do note with https://github.com/NixOS/nix/pull/7723 merged the `nix store ping` changes will be less work!

MatthewCroughan:
Have dropped the ball on this temporarily, I'm a bit ill after FOSDEM, will pick this up when I am better, anyone else feel free to contribute by making a PR to this PR!

fricklerhandwerk:
Discussed in the Nix team meeting 2023-04-03:

- @Ericson2314 rewrote the PR based on the team's requested changes, needs a different reviewer
- assigned to @thufschmitt

domenkozar:
Would be great to have this in for 2.15.

doc/manual/src/release-notes/rl-next.md
```diff
@@ -39,3 +39,8 @@
 
   * `man nix-store-<operation>` and `man nix-env-<operation>`
   * `nix-store --help --<operation>` and `nix-env --help --<operation>`.
+
+* Nix now checks whether the the store trusts you, the client.
+  This is useful for scripting interactions with (non-legacy-ssh) remote Nix stores.
```
thufschmitt, line 44: This message feels a bit weird. Nix obviously already checked that. There just wasn't a way to get this information from the client side

src/libstore/remote-store.cc
```diff
@@ -226,6 +226,23 @@ void RemoteStore::initConnection(Connection & conn)
...
         }
 
+        if (GET_PROTOCOL_MINOR(conn.daemonVersion) >= 35) {
+            auto temp = readNum<uint8_t>(conn.from);
+            switch (temp) {
+                case 0:
+                    conn.remoteTrustsUs = std::nullopt;
+                    break;
+                case 1 ... 2:
+                    conn.remoteTrustsUs = (TrustedFlag)(temp - 1);
```
thufschmitt, line 236: Let's not try to be confusingly smart here

```suggestion
                case 1:
                    conn.remoteTrustsUs = NotTrusted;
                    break;
                case 2:
                    conn.remoteTrustsUs = Trusted;
```

src/libstore/daemon.cc
```diff
@@ -1032,6 +1032,15 @@ void processConnection(
     if (GET_PROTOCOL_MINOR(clientVersion) >= 33)
         to << nixVersion;
 
+    if (GET_PROTOCOL_MINOR(clientVersion) >= 35) {
+        // We and the underlying store both need to trust the client for
+        // it to be trusted.
+        auto temp = trusted
+            ? store->isTrustedClient()
+            : std::optional { NotTrusted };
+        to << (temp ? 1 + (uint8_t)*temp : (uint8_t)0);
```
thufschmitt, line 1041: :exploding_head: 

Can we rather have a proper well-defined serialization scheme for `std::optional<TrustedFlag>` (in `worker-protocol.hh` like the rest I would assume)? This inlined arithmetic trick is too smart for its own good I think

src/libstore/remote-store.cc
thufschmitt: (also, I'd rather move this to a proper `worker_proto::read(..., TrustedFlag)` function like the rest)

thufschmitt (changes requested):
The change is good overall, but the wire encoding/decoding logic is a bit too smart imho. Let's keep things simple as much as we can.

Ericson2314:
Good point, I did sort of forget about that mechanism! :flushed:

src/libstore/remote-store.cc
```diff
@@ -42,6 +42,40 @@ void write(const Store & store, Sink & out, const StorePath & storePath)
...
+    auto temp = readNum<uint8_t>(from);
+    switch (temp) {
+        case 0:
+            return std::nullopt;
+        case 1:
+            return { Trusted };
+        case 2:
+            return { NotTrusted };
+        default:
+            throw Error("Invalid trusted status from remote");
```
thufschmitt, line 56: Saving it for a potential follow-up (and fairly low-prio), but it could be nicer to just warn and return `std::nullopt` in that case. Just to make potential future evolutions of the protocol nicer.

thufschmitt (approved):
Yay!

Just a small suggestion, but fairly minor and not blocking. So let's merge

src/libstore/remote-store.cc
Ericson2314: I had thought about that, but I think it's just better to bunl the protocol version on that case? Maybe I'm forgetting, but I couldn't think of another time we tried to do this sort of week *forward* compat before.

roberth: We currently have more or less one forward compat mechanism, which is the protocol version handshake.
If we're going to improve the forward compat mechanism, it would make more sense to do it structurally, with a self-describing or schema based serialization where we can add fields as we please.

PR #7518, opened 2022-12-27
separate man pages for `nix-store` and `nix-env` subcommands

fricklerhandwerk:
# Motivation

Documentation on "classic" commands with many sub-commands are notoriously hard to discover due to lack of overview and anchor links. Additionally the information on common options and environment variables is not accessible offline in `man` pages, and therefore often overlooked by readers.

With this change, each sub-command of `nix-store` and `nix-env` gets its own page in the manual (listed in the table of contents), and each own `man` page.

Also, `man` pages for each subcommand now (again) list common options and environment variables. While this makes each page quite long and some common parameters don't apply, this should still make it easier to navigate as that additional information was not accessible on the command line at all.

It is now possible to run `nix-store --<subcommand> --help` to display help pages for the given subcommand.

## Outline of implementation

- split the large `nix-store` and `nix-env` pages into separate pages, one for each sub-command living in a sub-directory
  - no changes to contents, only fixing links appropriately
- add nested entries to table of contents, add overview on each main command page
  - add note how to access these pages with `man` and `--help`
- convert sub-command pages  (changes to the Makefile `doc/src/local.mk`)
- change how the `--help` option works:
  - `nix-store --help --realise` will show the help page for the `realise` operation.
  - NOTE: currently `nix-store --realise --help` **does not** work
- re-use common options and environment variables on each command page using the include directive from mdBook
  - re-implement the include directive in the Makefile such that it also works for the terminal output (`man` and `--help` pages) generated with `lowdown`

Ideas for the future: include hints to use `man nix-*` and `nix* --help` conditionally depending on presentation.

## Alternatives

- refer readers to the other pages, e.g. from the command documentation to environment variable documentation
  - (+) smaller diff
  - (-) more interactions for the reader
  - (-) probably harder to discover, because it adds more text that is *not* actual documentation

## How to read the diff

- start from the third commit, which has the relevant changes to the makefile
- the following are changes following the initial review

# Checklist for maintainers

 - [x] agreed on idea
 - [ ] ~~unit tests~~ (no substantial change in behavior)
 - [ ] ~~functional tests (`tests/**.sh`)~~ (no substantial change in behavior)
 - [x] new feature or bug fix: updated release notes


doc/manual/local.mk
```diff
@@ -4,83 +4,119 @@ MANUAL_SRCS := \
...
+  NIX_CONF_DIR=/dummy \
+  NIX_SSL_CERT_FILE=/dummy/no-ca-bundle.crt \
+  NIX_STATE_DIR=/dummy \
+  NIX_CONFIG='cores = 0'
 
 nix-eval = $(dummy-env) $(bindir)/nix eval --experimental-features nix-command -I nix/corepkgs=corepkgs --store dummy:// --impure --raw
 
+$(d)/nix-env-%.1: $(d)/src/command-ref/nix-env/%.md
+  @printf "Title: %s\n\n" "$(subst nix-env-,nix-env --,$$(basename "$@" .1))" > $^.tmp
+  $(render-subcommand)
```
Ericson2314, line 38: Make needs this to be a tab.

Ericson2314: I think you did tabs -> spaces in this whole file which do don't want for make files.

fricklerhandwerk: It was mixed to begin with, for some reason. Why tabs not spaces?

Ericson2314: https://retrocomputing.stackexchange.com/questions/20292/why-does-make-only-accept-tab-indentation

Ericson2314: https://www.gnu.org/software/make/manual/html_node/Rule-Introduction.html

fricklerhandwerk: Will fix that. Apparently my Make foo got a bit dusty.

fricklerhandwerk:
@roberth now ready for review. Would be great if we could merge this soon so we don't diverge from master, because it's essentially impossible to track changes automatically, and we had to catch up twice already.

src/nix-env/nix-env.cc
```diff
@@ -1426,37 +1428,59 @@ static int main_nix_env(int argc, char * * argv)
...
+                opName = arg->substr(1);
+            }
+            else if (*arg == "--list-generations") {
                 op = opListGenerations;
-            else if (*arg == "--delete-generations")
+                opName = arg->substr(1);
+            }
+            else if (*arg == "--delete-generations") {
                 op = opDeleteGenerations;
+                opName = arg->substr(1);
```
roberth, line 1482: I would've preferred hardcoded names everywhere, as that's robust to addition of short flags, but those are unlikely to be added.

src/nix-store/nix-store.cc
```diff
@@ -1024,62 +1024,104 @@ static int main_nix_store(int argc, char * * argv)
...
-            else if (*arg == "--delete")
+            else if (*arg == "--delete") {
                 op = opDelete;
-            else if (*arg == "--query" || *arg == "-q")
+                opName = arg->substr(1);
+            }
+            else if (*arg == "--query" || *arg == "-q") {
                 op = opQuery;
-            else if (*arg == "--print-env")
+                opName = "-query";
```
roberth, line 1057: This one has further subcommands, but let's ignore that for now.

roberth (approved):
I've sampled `nix-* --help` `man nix-*` and the manual and it all seems to work. I've added a small commit and some comments, but found no blockers.

Issue #6666, opened 2022-06-13
enabling ca-derivations can cause deadlock on "waiting for exclusive access to the Nix store for ca drvs..."


elaforge:
**Describe the bug**

After adding `ca-derivations` to `experimental-features` in `nix.conf`, many servers have all nix operations stuck on "waiting for exclusive access to the Nix store for ca drvs...".

It looks like this is a deadlock when upgrading the sqlite schema, but I don't know when it happens.  It seems to have happened on most of our servers though so it must be easy to trigger.  Perhaps it happens when nix-daemon is restarted with the setting added and builds are currently in progress?  In any case, everything is then permanently deadlocked at that point, so existing builds completing don't fix it.

Aside from the deadlock, tools to inspect the sqlite version and status, and possibly upgrade it independently would be less error-prone than attempting to upgrade the schema implicitly on the fly.  I wasn't expecting that merely enabling a feature would be a risky operation.

**`nix-env --version` output**

2.4.1

**Additional context**

This has been separately reported in https://github.com/NixOS/nixops/issues/1520, but I don't think this is a nixops-only problem.

elaforge:
I've worked around this by forcing the upgrade (update nix.conf, force a build of any drv, even if it fails) on machines when they're idle.  I didn't get a verified reproduction, but it seems pretty easy to make it deadlock because I still got it several times when I tried without the initial drain.  Perhaps if the machine is doing any build at the same time as the upgrade it will deadlock.

thufschmitt:
I also encountered this. My hunch is that it's a NixOS issue (https://github.com/NixOS/nixpkgs/issues/177142): Somehow `nixos-rebuild switch` seems to keep an instance of the old Nix running while it runs the new one, hence the deadlock. Not totally sure, but it seems to be the most likely culprit (esp. given that the updates seem to work fine when running outside of the script)

thufschmitt:
> Aside from the deadlock, tools to inspect the sqlite version and status, and possibly upgrade it independently would be less error-prone than attempting to upgrade the schema implicitly on the fly.

There's an interesting tradeoff here. Upgrading the schema on-the-fly means that just using a new version might make the store backwards-incompatible. Otoh making that more independent would make things more complex both implementation-wise (because the code would have do deal with old db versions) and for users (because they would have to think about it).

> I wasn't expecting that merely enabling a feature would be a risky operation.

[...]

PR #7521, opened 2022-12-27
Release shared lock before acquiring exclusive lock

ncfavier:
See https://github.com/NixOS/nix/issues/6666#issuecomment-1356284834

In principle, this should avoid deadlocks where two instances of Nix are holding a shared lock on big-lock and are both waiting to get an exclusive lock.

However, it seems like `flock(2)` is supposed to do this automatically, so it's not clear whether this is actually where the problem comes from.


thufschmitt (approved):
Thanks,

I think it's worth merging and see whether the problem reappears :crossed_fingers:

PR #7523, opened 2022-12-28
Improve protocol major macro

roberth:
## Why

 - Prevent future surprises like https://github.com/NixOS/nix/pull/7515#discussion_r1057882208
 - Fix an exception message in `writeDerivedPaths`

## What

Most significant change: add `PROTOCOL_MAJOR`. All other changes follow from this. From the commit message:

`PROTOCOL_MAJOR` returns a simple major number, rather than the
number multiplied by 256. This is what one would reasonably
expect from `GET_PROTOCOL_MAJOR`, but we can't just change a
macro in a public header; hence the "rename".

### DIFF
--- a/src/libstore/serve-protocol.hh
+++ b/src/libstore/serve-protocol.hh
@@ -6,6 +6,9 @@ namespace nix {
 #define SERVE_MAGIC_2 0x5452eecb
 
 #define SERVE_PROTOCOL_VERSION (2 << 8 | 7)
+#define PROTOCOL_MAJOR(x) (((x) & 0xff00) >> 8)
+#define PROTOCOL_MINOR(x) ((x) & 0x00ff)
+// Deprecated:
 #define GET_PROTOCOL_MAJOR(x) ((x) & 0xff00)
 #define GET_PROTOCOL_MINOR(x) ((x) & 0x00ff)
 
--- a/src/libstore/worker-protocol.hh
+++ b/src/libstore/worker-protocol.hh
@@ -10,6 +10,9 @@ namespace nix {
 #define WORKER_MAGIC_2 0x6478696f
 
 #define PROTOCOL_VERSION (1 << 8 | 34)
+#define PROTOCOL_MAJOR(x) (((x) & 0xff00) >> 8)
+#define PROTOCOL_MINOR(x) ((x) & 0x00ff)
+// Deprecated:
 #define GET_PROTOCOL_MAJOR(x) ((x) & 0xff00)
 #define GET_PROTOCOL_MINOR(x) ((x) & 0x00ff)
 
--- a/src/libstore/daemon.cc
+++ b/src/libstore/daemon.cc
@@ -114,7 +114,7 @@ struct TunnelLogger : public Logger
         if (!ex)
             to << STDERR_LAST;
         else {
-            if (GET_PROTOCOL_MINOR(clientVersion) >= 26) {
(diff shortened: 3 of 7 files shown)


src/libstore/serve-protocol.hh
```diff
@@ -6,6 +6,9 @@ namespace nix {
 #define SERVE_MAGIC_2 0x5452eecb
 
 #define SERVE_PROTOCOL_VERSION (2 << 8 | 7)
+#define PROTOCOL_MAJOR(x) (((x) & 0xff00) >> 8)
+#define PROTOCOL_MINOR(x) ((x) & 0x00ff)
```
tfc, line 10: Does anything speak against implementing that as a nicely typed function like this:

```cpp
constexpr size_t protocol_minor(size_t x) { return x & 0xff; }
```

This way this happens at compile time if possible and no one will ever have to handle problems that occur from entering the wrong signed/unsigned type.

src/libstore/worker-protocol.hh
```diff
@@ -10,6 +10,9 @@ namespace nix {
 #define WORKER_MAGIC_2 0x6478696f
 
 #define PROTOCOL_VERSION (1 << 8 | 34)
+#define PROTOCOL_MAJOR(x) (((x) & 0xff00) >> 8)
+#define PROTOCOL_MINOR(x) ((x) & 0x00ff)
```
tfc, line 14: it seems like it would make sense to give the different major/minor macros (or functions if the other advise is followed) differing names as this also makes it easier to grep for their usage etc.

thufschmitt (changes requested):
Duh, that was indeed a nice footgun.

In addition to https://github.com/NixOS/nix/pull/7523#discussion_r1059656172, could we also deprecate the old macros? Deprecating a macro isn't really possible as far as I know, but we can hack something like:

```cpp
constexpr size_t deprecated(size_t x) { return x }
#define GET_PROTOCOL_MAJOR(x) (deprecated((x) & 0xff00))
#define GET_PROTOCOL_MINOR(x) (deprecated((x) & 0x00ff))
```

edolstra:
I would just remove the old macros. I think the only use is in Hydra which can be updated easily.

roberth:
The old macro should stick around for compatibility, as Hydra isn't the only consumer. The new macros are also necessary for library-consuming code that needs to work around certain issues (shouldn't have to, but escape hatches are good).

TODO: clean up the implementation code to use a struct `version.minor` is nicer than a ton of macro calls. (thanks @Ericson2314)

Ericson2314:
@roberth and I discussed this, and we think it would be good for the C++ code to fully parse the version once into a struct, and then pass that around instead.

@roberth also mentioned it would still be good to have a macro to --- in effect --- document the binary format itself separate from Nix's parser implementing it (which is the spirit of these `foo-protocol.hh` headers anyways), and that sounds fine to me.

roberth:
> fully parse the version once into a struct

Have tried this but failed. I accidentally broke something.

I would prefer to just merge this, to provide an alternative to the footgun, and then move on to some more relevant problem solving.

Ericson2314:
@roberth Sounds good. Do you think you could push your attempt somewhere? I would be happy to take a stab at the follow-up if that helps.

roberth:
It looks like I did a proper rage quit deleting my edits. It's mostly just a bunch of find and replace, so it's probably best to start over anyway. Sorry for dropping the actual struct bit though :/

tomberek:

Issue #7437, opened 2022-12-09
cli: read "paths" from stdin for commands that take them


nrdxp:
**Is your feature request related to a problem? Please describe.**
If the user wants to pass a list of paths to say `nix path-info` or `nix-build`, they run the risk of recieving a "too many args" OS error.

Since passing derivations around can be very efficient when handled in batches it would be nice not to have a cap on this. The current workaround is `xargs`, which I guess is fine, but not exactly intuitive or necessarily efficient.

**Describe the solution you'd like**
I could see it two ways, if the command recieves a non-empty stdin, simply add it to whatever list of paths already exist in the cli, or we could make it explicit and require something like `nix-build -` to specify reading paths from stdin.

**Describe alternatives you've considered**
Basically just keep using xargs, but even if the arg limit is not an issue, shell pipelines are far more readable than nested process substititions, e.g:
```console
$ which svn | nix-store -qR | nix-store --realize
# is nicer than
$ nix-build $(nix-store -qR $(which svn))
```

**Additional context**
Hit this a couple of times when trying to ensure the entire build closure is cached by parsing the results of a `--dry-run`: [divnix/std-action/run/build.sh#14](https://github.com/divnix/std-action/blob/b01efeb31dca2df9f7672233ac1de92068be3671/run/build.sh#L14)

nrdxp:
Just found `nix-store --export` and it's probably the most important usecase for this, since using `xargs` wouldn't wok in this case to serialise a potentially large list of derivations at once to another store.

PR #7526, opened 2022-12-29
nix-store: implement --query --unsubstitutable

### LINKED ISSUE #3946: support building only drvs that lack substitutes (aka `nix-build-uncached`)

nrdxp:
Resolves #3946 by providing an efficient means of querying cache status of the passed paths.

For every path passed, it will be returned back on standard output only if it is not available in any of the configured substituters.

~~A threadpool is used to make this as quick as possible.~~

Included in the documentation is a sample command showing how users can now trivially capture any and all uncached build & runtime dependencies of a given derivation. This information is highly useful in CI environments, for example.

Only caveat is that we might hit the OS arg limit until #7437 is fixed.

#### TODO
- [x] implement `nix-store` query
- [ ] write tests

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`

### DIFF
--- a/src/nix-store/nix-store.cc
+++ b/src/nix-store/nix-store.cc
@@ -373,7 +373,22 @@ static void opQuery(Strings opFlags, Strings opArgs)
                storePaths.emplace(store->followLinksToStorePath(str));
             }
 
-            auto substitutablePaths = store->querySubstitutablePaths(storePaths);
+            StringSet pathsS;
+            // query each configured substituter
+            for (auto & sub : getDefaultSubstituters()) {
+              if (sub->storeDir != store->storeDir) continue;
+              if (!sub->wantMassQuery) continue;
+
+              auto paths = sub->queryValidPaths(StorePathSet(storePaths.begin(), storePaths.end()));
+              for (auto & path : paths) {
+                pathsS.emplace(store->printStorePath(path));
+              }
+            }
+
+            StorePathSet substitutablePaths;
+            for (auto & str : pathsS) {
+              substitutablePaths.emplace(store->followLinksToStorePath(str));
+            }
 
             for ( auto & path : storePaths) {
                 if (substitutablePaths.find(path) == substitutablePaths.end())
--- a/src/nix-store/nix-store.cc
+++ b/src/nix-store/nix-store.cc
@@ -276,7 +276,7 @@ static void opQuery(Strings opFlags, Strings opArgs)
     enum QueryType
         { qDefault, qOutputs, qRequisites, qReferences, qReferrers
         , qReferrersClosure, qDeriver, qBinding, qHash, qSize
-        , qTree, qGraph, qGraphML, qResolve, qRoots };
+        , qTree, qGraph, qGraphML, qResolve, qRoots, qUnsubstitutable };
     QueryType query = qDefault;
     bool useOutput = false;
     bool includeOutputs = false;
@@ -300,6 +300,7 @@ static void opQuery(Strings opFlags, Strings opArgs)
         }
         else if (i == "--hash") query = qHash;
         else if (i == "--size") query = qSize;
+        else if (i == "--unsubstitutable" || i == "-U") query = qUnsubstitutable;
         else if (i == "--tree") query = qTree;
         else if (i == "--graph") query = qGraph;
         else if (i == "--graphml") query = qGraphML;
@@ -363,6 +364,24 @@ static void opQuery(Strings opFlags, Strings opArgs)
                 cout << fmt("%s\n", info->deriver ? store->printStorePath(*info->deriver) : "unknown-deriver");
             }
             break;
+
+        case qUnsubstitutable: {
+            if (opArgs.empty()) throw UsageError("no paths passed");
+
+            StorePathSet storePaths;
+            for (auto const & str : opArgs) {
+               storePaths.emplace(store->followLinksToStorePath(str));
+            }
+
+            auto substitutablePaths = store->querySubstitutablePaths(storePaths);
+
+            for ( auto & path : storePaths) {
+                if (substitutablePaths.find(path) == substitutablePaths.end())
+                    std::cout << '\n' << store->printStorePath(path);
+            }
+            std::cout << std::endl;
+            break;
+        }
 
         case qBinding:
             for (auto & i : opArgs) {
@@ -679,7 +698,6 @@ static void opExport(Strings opFlags, Strings opArgs)
     store->exportPaths(paths, sink);
     sink.flush();
 }
-
 
 static void opImport(Strings opFlags, Strings opArgs)
 {
--- a/doc/manual/src/command-ref/nix-store.md
+++ b/doc/manual/src/command-ref/nix-store.md
@@ -289,7 +289,7 @@ error: cannot delete path `/nix/store/zq0h41l75vlb4z45kzgjjmsjxvcv1qk7-mesa-6.4'
   {`--outputs` | `--requisites` | `-R` | `--references` |
   `--referrers` | `--referrers-closure` | `--deriver` | `-d` |
   `--graph` | `--tree` | `--binding` *name* | `-b` *name* | `--hash` |
-  `--size` | `--roots`}
+  `--size` | `--roots` | `--unsubstitutable` | `-U`}
   [`--use-output`] [`-u`] [`--force-realise`] [`-f`]
   *paths…*
 
@@ -405,6 +405,10 @@ symlink.
     Prints the garbage collector roots that point, directly or
     indirectly, at the store paths *paths*.
 
+  - `--unsubstitutable`; `-U`\
+    Outputs a list of the store paths *paths* which are not available in
+    any configured substituters.
+
 ## Examples
 
 Print the closure (runtime dependencies) of the `svn` program in the
@@ -440,6 +444,16 @@ $ nix-store -q --tree $(nix-store -qd $(which svn))
 +---/nix/store/fmzxmpjx2lh849ph0l36snfj9zdibw67-bash-3.0.drv
 |   +---/nix/store/570hmhmx3v57605cqg9yfvvyh0nnb8k8-bash
 |   +---/nix/store/p3srsbd8dx44v2pg6nbnszab5mcwx03v-builder.sh
+...
+```
+
+Show unsubstitutable build & runtime dependencies:
+
+```console
+$ drv=$(nix-store -qd $(which svn))
+$ nix-store -qU $(nix-store -qR --include-outputs $drv)
+/nix/store/5mbglq5ldqld8sj57273aljwkfvj22mc-subversion-1.1.4
+/nix/store/9lz9yc6zgmc0vlqmn2ipcpkjlmbi51vv-glibc-2.3.4
 ...
 ```
 


mlyxshi:
Are you interested in this relative issue?
https://github.com/NixOS/nix/issues/7527

nrdxp:
> Are you interested in this relative issue? #7527

You could use this PR to accomplish your goal by simply processing you derivations through `nix-store -qU` before sending them to `nix copy`. As Sandro rightly pointed out though, it will still result in a few redundant dependencies, since if the entire dependency closure does not exist in a single cache, Nix throws an error. To improve the situation we just have to allow Nix to continue instead of erroring in this situation, and you could then just pass a `--no-recursive` to `nix copy` and cache only those that don't exist elsewhere.

In general, my recent experience with Nix in CI has kind of changed my perspective somewhat, and I'm starting to feel like we should take a more "git-like" approach to our cli, rather than replacing the older, more verbose commands outright, we should stabalize them as much as possible and use them as the "plumbing" to the higher level commands.

nrdxp:
Because there isn't really a nice standalone solution for this atm, I went ahead and extracted the logic as a standalone tool for now until this is accepted:
https://github.com/divnix/nix-uncached

edolstra:
I think functionality like this should be added to the new CLI rather than the legacy CLI. It seems to me that what we really want here is a CLI interface to `Store::queryMissing()` (ideally with JSON output).

Also "uncached" is a vague term. What does it mean for a path to be uncached? Maybe "unsubstitutable" is better.

nrdxp:
> I think functionality like this should be added to the new CLI rather than the legacy CLI. It seems to me that what we really want here is a CLI interface to `Store::queryMissing()` (ideally with JSON output).

I can add it to `nix path-info` I think, fairly easily.

My first implementation did just use queryMissing, but that isn't _exactly_ what I wanted, since I don't care if the path exists locally or not, I just want to know if it exists in a substituter. If you want to include the local store as part of the equation you can add `--extra-substituters auto` to the invocation to make that explicit.

> Also "uncached" is a vague term. What does it mean for a path to be uncached? Maybe "unsubstitutable" is better.

Yeah I was thinking that myself, for now I just used the terminology that has been used elsewhere in the community so far, but I was also thinking of using unsubstitutable instead just last night, only thing is that it's quite long, although I guess with a short flag it would still be fine.

nrdxp:
@edolstra, I implemented the same logic for `nix path-info`, including allowing it the `--json` to add a new boolean output showing whether a path is cached or not.

However there are a few downsides to the `nix path-info` implementation compared to the `nix-store` one so far. In order to explain that, understand that my main usecase for this is querying information about the paths before they have been built, to decide if they need to be built, etc. I the paths I need from a flake using `nix eval` which produces a special json we then want to manipulate based on information we are going to query, like the cache status.

Importantly, I only want to run evaluation once, and only once, for a given CI run, so passing flake refs around isn't desirable. Instead we query information on the derivations directly. This just doesn't work well with `nix path-info`.

With the `nix-store` implementation it works perfectly, I can just query the derivation with `nix-store -qR --include-outputs` and then filter that through my new `nix-store -qU` and I've got the information I need, minus some additional post-processing which can be easily done with `jq` and `nix show-derivation` (to remove locally built derivations from the equation, etc). 

Specifically these are the quircks I've encountered with the `nix path-info` implementation so far:
* as already mentioned, there is no equivalent to the `--include-outputs` flag, so its either querying buildtime _or_ runtime deps, not both.
* If I pass a derivation (which is all I have), then the command will not return information about it's outpath. I have to explicitly pass the outpath for that, but that is a non-starter for my usecase since I don't have them built yet (and don't want to necessarily).

For these reasons I propose we keep both implementations until both are more in-line, and as mentioned in my previous comment. As mentioned in my previous comment, we may want to consider a more "plumbing" and "porceline" approach for flakes 2.0, to keep drift between the old and new commands from becoming a big issue.

src/nix/path-info.cc
```diff
@@ -87,41 +96,98 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
+                    auto info = store->queryPathInfo(store->followLinksToStorePath(storePathS));
+                    auto storePath = info->path;
+
+                    SubstitutablePathInfos infos;
+                    store->querySubstitutablePathInfos({{storePath, std::nullopt}}, infos);
+
+                    clone["substitute"] = !infos.empty();
+
+                    auto jsonList(jsonList_.lock());
+                    jsonList->emplace_back(clone);
```
tfc, line 127: in order to really emplace it at the back without unneeded copies i suggest this:
```suggestion
                    jsonList->emplace_back(std::move(clone));
```

src/nix-store/nix-store.cc
```diff
@@ -364,6 +369,32 @@ static void opQuery(Strings opFlags, Strings opArgs)
...
+                if (infos.empty()) {
+                    cout << store->printStorePath(req) << '\n';
+                }
+            };
+
+            for (auto & i : opArgs) {
+                auto path = store->followLinksToStorePath(i);
+
+                pool.enqueue(std::bind(queryPath, path));
+            }
```
tfc, line 391: Does anything speak against leaving out the `std::function` and `std::bind` by writing it like this: ?

```cpp
            auto queryPathThunk = [&](const StorePath & req) {
                return [&]{
                    SubstitutablePathInfos infos;
                    store->querySubstitutablePathInfos({{req, std::nullopt}}, infos);

                    if (infos.empty()) {
                        cout << store->printStorePath(req) << '\n';
                    }
                };
            };

            for (auto & i : opArgs) {
                pool.enqueue(queryPathThunk(store->followLinksToStorePath(i)));
            }
```

This would remove some typing information that is both unneeded and fixed otherwise.

src/nix/path-info.cc
```diff
@@ -87,41 +96,98 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
+
+                Sync<nlohmann::json> jsonList_;
+
+                {
+                    auto jsonList(jsonList_.lock());
+                    *jsonList = nlohmann::json::array();
+                }
+
+                std::function<void(nlohmann::json::basic_json::value_type)> checkSubs;
+                checkSubs = [&](const nlohmann::json::basic_json::value_type & value) {
```
tfc, line 115: same suggestion with the nested lambda expression here. all the type info is relatively noisy and can be reduced.

src/nix/path-info.cc
```diff
@@ -87,41 +96,98 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
+                    auto clone = value;
+                    std::string storePathS = value["path"];
+                    auto info = store->queryPathInfo(store->followLinksToStorePath(storePathS));
+                    auto storePath = info->path;
```
tfc, line 119: it seems like some `const ... &` could help the reader understand what really needs to be a (mutable) copy and what not, as well as reduce some redundant copying.

src/nix/path-info.cc
```diff
@@ -87,41 +96,98 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
         }
 
         else {
+            ThreadPool pool(fileTransferSettings.httpConnections);
 
-            for (auto & storePath : storePaths) {
+            Sync<std::set<std::string>> results_;
+
+            std::function<void(StorePath)> printPath;
+            printPath = [&](const StorePath & storePath) {
```
tfc, line 147: Here, the parameter type from the `std::function` and the actual assignment differ by having and not having `const ... &` qualifiers. i suggest using the same nested lambda expression approach here, too.

src/nix/path-info.cc
```diff
@@ -87,41 +96,98 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
                     if (info->ultimate) ss.push_back("ultimate");
                     if (info->ca) ss.push_back("ca:" + renderContentAddress(*info->ca));
                     for (auto & sig : info->sigs) ss.push_back(sig);
-                    std::cout << concatStringsSep(" ", ss);
+                    storePathS.append(concatStringsSep(" ", ss));
                 }
 
-                std::cout << std::endl;
+                auto results(results_.lock());
+                results->emplace(storePathS);
```
tfc, line 179: using `std::move` will help avoiding a needless copy.

src/nix/path-info.cc
```diff
@@ -87,41 +96,98 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
+
+            for (auto & storePath : storePaths) {
+                pool.enqueue(std::bind(printPath, storePath));
             }
+            pool.process();
 
+            auto results(results_.lock());
+            for (auto & str : *results) {
+                std::cout << std::endl << str;
+            }
```
tfc, line 190: `std::endl` means to emit a new line character *and* flush the terminal. But this code suggests that you mean to separate all strings by newlines and then flush only once at the end, so i suggest this:

```suggestion
            auto results(results_.lock());
            for (const auto & str : *results_.lock()) {
                std::cout << '\n' << str;
            }
            std::cout << std::endl;
```

nrdxp:
Thanks for the review, I'm moving to draft for now as the implementation needs work. After some benchmarks, the threadpool didn't actually help anything and is about the same as using a single thread, even when setting the threadpool limit to something super high like 1000.

Its possible that an async implementation may be superior or there is some bottleneck that needs to be tracked down. Until that is worked out, I'll leave in draft.

nrdxp:
Found the performance bottleneck. Turns out we don't need a threadpool at all but can get really great performance from the `Store::querySubstitutablePaths` function itself. Also implemented the same logic for `nix path-info` in a separate PR: #7587 as requested by @edolstra 

I split the two up so that one does not block merging of the other if we decide not to keep both.

src/nix-store/nix-store.cc
```diff
@@ -364,6 +365,39 @@ static void opQuery(Strings opFlags, Strings opArgs)
             }
             break;
 
+        case qUnsubstitutable: {
+            if (opArgs.empty()) throw UsageError("no paths passed");
+
+            StorePathSet storePaths;
+            for (auto const & str : opArgs) {
+               storePaths.emplace(store->followLinksToStorePath(str));
```
edolstra, line 373: The indentation here and below is wrong (should be 4 characters per level).

src/nix-store/nix-store.cc
```diff
@@ -364,6 +365,39 @@ static void opQuery(Strings opFlags, Strings opArgs)
...
+              }
+            }
+
+            StorePathSet substitutablePaths;
+            for (auto & str : pathsS) {
+              substitutablePaths.emplace(store->followLinksToStorePath(str));
+            }
+
+            for ( auto & path : storePaths) {
+                if (substitutablePaths.find(path) == substitutablePaths.end())
```
edolstra, line 394:
```suggestion
                if (!substitutablePaths.count(path))
```

ncfavier: This does not short-circuit

edolstra: In what way? On an `std::map` it should be equally efficient.

But since we're fine with C++20 now, you can also use `substitutablePaths.contains(path)`.

nrdxp:
closing in favour of the aforementioned PR

tomberek:
Miss click

PR #7531, opened 2023-01-01
Fix typo in error message

steshaw:
Tiny fix to insert a space between "how" and "to" in the error message. i.e.

from:

> if using 'nix profile' to find out howto change the priority of one of the conflicting packages

to:

> if using 'nix profile' to find out how to change the priority of one of the conflicting packages


thufschmitt (approved):
Thanks :)

PR #7539, opened 2023-01-02
Fix `nix why-depends --derivation`

### LINKED ISSUE #7496: `nix why-depends --derivation` gives false negatives on Nix 2.12
**Describe the bug**

The `why-depends` command on Nix 2.12 claims that two derivations do not depend on each other, when in reality they do (and  Nix 2.11 shows the dependency correctly).

**Steps To Reproduce**

In a nixpkgs checkout:
```
$ git checkout nixos-unstable
Already on 'nixos-unstable'

$ git rev-parse HEAD
04f574a1c0fde90b51bf68198e2297ca4e7cccf4

$ nix run -f . nixVersions.nix_2_11 -- why-depends --derivation -f . hello stdenv
/nix/store/7vf0d0j7majv1ch1xymdylyql80cn5fp-hello-2.12.1.drv
└───/nix/store/ap9g09fxbicj836zm88d56dn3ff4clxl-stdenv-linux.drv

$ nix run -f . nixVersions.nix_2_12 -- why-depends --derivation -f . hello stdenv
'hello' does not depend on 'stdenv'
```

Runtime dependencies are correctly detected. For example, `why-depends -f hello glibc.out` gives the correct result in both Nix versions.

**Expected behavior**

`why-depends --derivation` should correctly determine if and how two derivations depend on each other.

nix-env (Nix) 2.12.0

**Additional context**

**Priorities**

thufschmitt:
Reverts #7337 and re-implements it in a different way in order to not break `nix why-depends --derivation`.

Fix #7496

Cc @raphaelr

### DIFF
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -931,10 +931,7 @@ std::vector<std::pair<std::shared_ptr<Installable>, BuiltPathWithResult>> Instal
                                 DrvOutput outputId { *outputHash, output };
                                 auto realisation = store->queryRealisation(outputId);
                                 if (!realisation)
-                                    throw Error(
-                                        "cannot operate on an output of the "
-                                        "unbuilt derivation '%s'",
-                                        outputId.to_string());
+                                    throw MissingRealisation(outputId);
                                 outputs.insert_or_assign(output, realisation->outPath);
                             } else {
                                 // If ca-derivations isn't enabled, assume that
--- a/src/libstore/realisation.hh
+++ b/src/libstore/realisation.hh
@@ -93,4 +93,14 @@ struct RealisedPath {
     GENERATE_CMP(RealisedPath, me->raw);
 };
 
+class MissingRealisation : public Error
+{
+public:
+    MissingRealisation(DrvOutput & outputId)
+        : Error( "cannot operate on an output of the "
+                "unbuilt derivation '%s'",
+                outputId.to_string())
+    {}
+};
+
 }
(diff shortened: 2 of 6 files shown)


Radvendii:
Using `Realise::Derivation` here tells it to *only* realise the derivation, and not go further (i.e. it won't try to resolve the output if it hasn't already been resolved). `operateOn` tells it to return the store path **of** the derivation if invoked with `--derivation`. Is that right?

raphaelr:
Can confirm that this fixes the issue:
```
$ nix run 'github:NixOS/nix/pull/7539/head' -- why-depends --derivation -f . libsForQt5.qtbase nasm
/nix/store/6mr2fs8z9arz8wfc357szvcnz3vvmxv4-qtbase-5.15.7.drv
└───/nix/store/9kzlffpxzj7zf3fmz637y5glkqnawj99-libjpeg-turbo-2.1.4.drv
    └───/nix/store/wpd81rvxfd3lc4qf0l5dls5fq3p0vdny-nasm-2.15.05.drv
```

thufschmitt:
> * Using `Realise::Derivation` here tells it to _only_ realise the derivation, and not go further (i.e. it won't try to resolve the output if it hasn't already been resolved).
> 
>     * `operateOn` tells it to return the store path **of** the derivation if invoked with `--derivation`.
> 
> 
> Is that right?

Not exactly: `Realise::Derivation` is some kind of `dry-run` mode. Unless you also pass `OperateOn::Derivation`, it will still try to return the output path (despite not building it). So it fails if the derivation is an unbuilt CA derivation (hence the `try-catch` block).

PR #7541, opened 2023-01-03
Check links in the manual

### LINKED ISSUE #7535: [chore] look into link checking for the manual
Make documentation easier to review.

**Priorities**

roberth:
- [x] is the idea good? has it been discussed by the Nix team?
   - Closes #7535 
 - [x] unit tests
   - Part of the build
 - [x] functional tests (`tests/**.sh`)
   - N/A
 - [x] documentation in the code (if necessary; ideally code is already clear)
   - Makefile inline scripts are not conducive to such comments
     - So use individual commands instead of inline scripts
 - [x] documentation in the commit message (why was this change made? for future reference when maintaining the code)
 - [x] documentation in the changelog (to announce features and fixes to existing users who might have to do something to finally solve their problem, and to summarize the development history)
    - N/A

### DIFF
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -108,7 +108,10 @@ doc/manual/generated/man1/nix3-manpages: $(d)/src/command-ref/new-cli
 $(docdir)/manual/index.html: $(MANUAL_SRCS) $(d)/book.toml $(d)/anchors.jq $(d)/custom.css $(d)/src/SUMMARY.md $(d)/src/command-ref/new-cli $(d)/src/command-ref/conf-file.md $(d)/src/language/builtins.md
 	$(trace-gen) \
 	  set -euo pipefail; \
-	  RUST_LOG=warn mdbook build doc/manual -d $(DESTDIR)$(docdir)/manual 2>&1 \
+	  RUST_LOG=warn mdbook build doc/manual -d $(DESTDIR)$(docdir)/manual.tmp 2>&1 \
 		  | { grep -Fv "because fragment resolution isn't implemented" || :; }
+	@rm -rf $(DESTDIR)$(docdir)/manual
+	@mv $(DESTDIR)$(docdir)/manual.tmp/html $(DESTDIR)$(docdir)/manual
+	@rm -rf $(DESTDIR)$(docdir)/manual.tmp
 
 endif
--- a/doc/manual/local.mk
+++ b/doc/manual/local.mk
@@ -50,11 +50,14 @@ $(d)/src/SUMMARY.md: $(d)/src/SUMMARY.md.in $(d)/src/command-ref/new-cli
 
 $(d)/src/command-ref/new-cli: $(d)/nix.json $(d)/generate-manpage.nix $(bindir)/nix
 	@rm -rf $@
-	$(trace-gen) $(nix-eval) --write-to $@ --expr 'import doc/manual/generate-manpage.nix { toplevel = builtins.readFile $<; }'
+	$(trace-gen) $(nix-eval) --write-to $@.tmp --expr 'import doc/manual/generate-manpage.nix { toplevel = builtins.readFile $<; }'
+	$(trace-gen) sed -i $@.tmp/*.md -e 's^@docroot@^../..^g'
+	@mv $@.tmp $@
 
 $(d)/src/command-ref/conf-file.md: $(d)/conf-file.json $(d)/generate-options.nix $(d)/src/command-ref/conf-file-prefix.md $(bindir)/nix
 	@cat doc/manual/src/command-ref/conf-file-prefix.md > $@.tmp
-	$(trace-gen) $(nix-eval) --expr 'import doc/manual/generate-options.nix (builtins.fromJSON (builtins.readFile $<))' >> $@.tmp
+	$(trace-gen) $(nix-eval) --expr 'import doc/manual/generate-options.nix (builtins.fromJSON (builtins.readFile $<))' \
+	  | sed -e 's^@docroot@^..^g'>> $@.tmp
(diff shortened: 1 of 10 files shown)


doc/manual/src/contributing/hacking.md
```diff
@@ -249,3 +249,22 @@ search/replaced in it for each new build.
...
+make html -j $NIX_BUILD_CORES
+```
+
+When iterating on the makefile, run:
+
+```console
+rm $(git ls-files doc/manual/ -o | grep -F '.md') && rmdir doc/manual/src/command-ref/new-cli && make html -j $NIX_BUILD_CORES
+```
+
+If a broken link occurs in a snippet that was inserted into multiple generated files in different directories, use `@docroot@` to reference the `doc/manual/src` directory.
```
fricklerhandwerk, line 268: I'm not sure what this instruction is trying to tell me. In which case does this happen? I suppose you encountered this during debugging, but I'm lacking the context to follow. Maybe `@docroot@` should be documented in the Makefile itself (saying why it exists and when to use it)?

src/libexpr/primops.cc
```diff
@@ -253,7 +254,7 @@ static RegisterPrimOp primop_import({
       >
       > Unlike some languages, `import` is a regular function in Nix.
       > Paths using the angle bracket syntax (e.g., `import` *\<foo\>*)
-      > are [normal path values](language-values.md).
+      > are normal path values.
```
fricklerhandwerk, line 257:
```suggestion
      > are normal [path values](@docroot@/language/values.md#type-path).
```

Does anything speak against making the link useful instead of removing it?

fricklerhandwerk (commented):
This is really good, especially because it solves the problem of having to figure out where to point relative paths from in-code documentation thanks to the `@docroot@` convention.

doc/manual/src/contributing/hacking.md
roberth: It doesn't have a single place in the Makefile either, so this seems like a central place for it.

src/libexpr/primops.cc
roberth: I couldn't find this somehow. Problem between keyboard and chair.

doc/manual/src/contributing/hacking.md
roberth: I've linked this from the makefile in 4917ca17ce70c31ee5c0b16a707bfa5ec647f1e1

doc/manual/local.mk
```diff
@@ -102,6 +109,12 @@ doc/manual/generated/man1/nix3-manpages: $(d)/src/command-ref/new-cli
 	@touch $@
 
 $(docdir)/manual/index.html: $(MANUAL_SRCS) $(d)/book.toml $(d)/anchors.jq $(d)/custom.css $(d)/src/SUMMARY.md $(d)/src/command-ref/new-cli $(d)/src/command-ref/conf-file.md $(d)/src/language/builtins.md
-	$(trace-gen) RUST_LOG=warn mdbook build doc/manual -d $(DESTDIR)$(docdir)/manual
+	$(trace-gen) \
+	  set -euo pipefail; \
+	  RUST_LOG=warn mdbook build doc/manual -d $(DESTDIR)$(docdir)/manual.tmp 2>&1 \
+		  | { grep -Fv "because fragment resolution isn't implemented" || :; }
```
fricklerhandwerk, line 115:
```suggestion
		  | { grep -Fv "because URI fragment resolution isn't implemented" || :; }
```

doc/manual/src/contributing/hacking.md
fricklerhandwerk:
```suggestion
```

Now that you have the sentence in the new subsection.

doc/manual/src/contributing/hacking.md
```diff
@@ -249,3 +249,36 @@ search/replaced in it for each new build.
...
+[`mdbook-linkcheck`]: https://github.com/Michael-F-Bryan/mdbook-linkcheck
+[URI fragments]: https://en.m.wikipedia.org/wiki/URI_fragment
+
+#### `@docroot@` variable
+
+`@docroot@` provides a base path for links that occur in reusable snippets or other documentation that doesn't have a base path of its own.
+
+If a broken link occurs in a snippet that was inserted into multiple generated files in different directories, use `@docroot@` to reference the `doc/manual/src` directory.
+
+If the `@docroot@` literal appears in an error message from the `mdbook-linkcheck` tool, the `@docroot@` replacement needs to be applied to the (probably) generated source file that mentions it. See existing `@docroot@` logic in the [Makefile]. Regular markdown files that are not copied to other markdown files do have a base path of their own and they can use relative paths instead of `@docroot@`.
```
fricklerhandwerk, line 284:
```suggestion
If the `@docroot@` literal appears in an error message from the `mdbook-linkcheck` tool, the `@docroot@` replacement needs to be applied to the generated source file that mentions it.
See existing `@docroot@` logic in the [Makefile].
Regular markdown files used for the manual have a base path of their own and they can use relative paths instead of `@docroot@`.
```

fricklerhandwerk (approved):

doc/manual/local.mk
roberth: This is a filter to remove useless messages, not documentation.

fricklerhandwerk: 🤦🏻‍♂️ Yes, right.

doc/manual/src/contributing/hacking.md
fricklerhandwerk: @roberth just this fix and you can merge yourself. I already approved.

Issue #7522, opened 2022-12-28
Unwanted verbose output about nix fish init scripts when starting fish


tabbyrobin:
**Describe the bug**

After installing nix, there is verbose output whenever I start fish shell.  I'm not sure whether these are errors or just verbose output. Generally nix seems to work fine. This is what shows at every launch:

```shell
/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish (line 34): 
fish_add_path --prepend --global "/nix/var/nix/profiles/default/bin"
^
from sourcing file /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish
	called on line 185 of file /usr/share/fish/config.fish
in function '.' with arguments '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish'
	called on line 4 of file /etc/fish/conf.d/nix.fish
from sourcing file /etc/fish/conf.d/nix.fish
	called on line 294 of file /usr/share/fish/config.fish
from sourcing file /usr/share/fish/config.fish
	called during startup
/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish (line 35): 
fish_add_path --prepend --global "$HOME/.nix-profile/bin"
^
from sourcing file /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish
	called on line 185 of file /usr/share/fish/config.fish
in function '.' with arguments '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish'
	called on line 4 of file /etc/fish/conf.d/nix.fish
from sourcing file /etc/fish/conf.d/nix.fish
	called on line 294 of file /usr/share/fish/config.fish
from sourcing file /usr/share/fish/config.fish
	called during startup
Welcome to fish, the friendly interactive shell
Type `help` for instructions on how to use fish
```

**Steps To Reproduce**

1. Have fish installed on debian 10.
2. Install nix multiuser via `sh <(curl -L https://nixos.org/nix/install) --daemon` as per https://nixos.org/download.html (Yes to use sudo)
3. After install succeeds, restart shells.
4. Start fish, see verbose output before fish prompt appears.

**Expected behavior**

Same as observed behavior, except without the verbose output when starting fish.

**`nix-env --version` output**

nix-env (Nix) 2.12.0

**Additional context**

I am using a qubes standalone vm based on the debian-10 template.

abathur:
I think this is the same issue reported in #7546

edolstra:
[...]

PR #7547, opened 2023-01-03
Fix Nix installation on older versions of fish

### LINKED ISSUE #7546: Installing does not work with older fish versions (<3.2.0)
**Describe the bug**
Fish shell support [was recently added](https://github.com/NixOS/nix/pull/7014), however it only works on recent versions of fish. In particular, it does not work on debian-stable (bullseye), where fish version [is currently 3.1.2](https://packages.debian.org/stable/shells/fish).

The installer currently relies on the `fish_add_path` function ([new in fish 3.2.0](https://fishshell.com/docs/current/relnotes.html#fish-3-2-0-released-march-1-2021)) to update the PATH:
* in [nix-profile.fish.in](https://github.com/NixOS/nix/blob/15341334b50b2b432830a8e78abc331d0a47de13/scripts/nix-profile.fish.in#L35)
* in [nix-profile-daemon.fish.in](https://github.com/NixOS/nix/blob/15341334b50b2b432830a8e78abc331d0a47de13/scripts/nix-profile-daemon.fish.in#L34)

I would suggest changing fish_add_path by something like:
```
# in nix-profile.fish.in
set -g fish_user_paths "$NIX_LINK/bin" $fish_user_paths
```
when `fish_add_path` is not available.

**Steps To Reproduce**
Install nix and open a fish shell with version < 3.2.0. You will see the following error message, and will not be able to run any nix command:
```
~/.nix-profile/etc/profile.d/nix.fish (line 35):
    fish_add_path --prepend --global "$NIX_LINK/bin"
    ^
from sourcing file ~/.nix-profile/etc/profile.d/nix.fish
        called on line 185 of file /usr/share/fish/config.fish
in function '.' with arguments '/home/username/.nix-profile/etc/profile.d/nix.fish'
        called on line 2 of file ~/.config/fish/conf.d/nix.fish
from sourcing file ~/.config/fish/conf.d/nix.fish
        called on line 294 of file /usr/share/fish/config.fish
from sourcing file /usr/share/fish/config.fish
        called during startup
```

**Expected behavior**
No error message should be displayed, and `$PATH` should contain `$NIX_LINK/bin`, so that nix commands work as expected.

`nix-env (Nix) 2.12.0`

alxthm:
The `fish_add_path` function is only available for fish 3.2.0 or newer, and not on older versions (e.g. 3.1.2, current version on debian stable, bullseye).
This commit adds an alternative way to update the PATH when `fish_add_path` does not exist.

Fixes #7546

### DIFF
--- a/scripts/nix-profile-daemon.fish.in
+++ b/scripts/nix-profile-daemon.fish.in
@@ -1,3 +1,15 @@
+function add_path --argument-names new_path
+    if type -q fish_add_path
+        # fish 3.2.0 or newer
+	fish_add_path --prepend --global $new_path
+    else
+        # older versions of fish
+        if not contains $new_path $fish_user_paths
+            set --global fish_user_paths $new_path $fish_user_paths
+        end
+    end
+end
+
 # Only execute this file once per shell.
 if test -n "$__ETC_PROFILE_NIX_SOURCED"
   exit
@@ -31,5 +43,7 @@ else
   end
 end
 
-fish_add_path --prepend --global "@localstatedir@/nix/profiles/default/bin"
-fish_add_path --prepend --global "$HOME/.nix-profile/bin"
+add_path "@localstatedir@/nix/profiles/default/bin"
+add_path "$HOME/.nix-profile/bin"
+
+functions -e add_path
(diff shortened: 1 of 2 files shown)


abathur:
It looks like this would also close #7522

abathur:
Not technically a requirement AFAIK, but https://nixos.org/manual/nix/stable/contributing/hacking.html#installer-tests details how to enable the installer tests in your fork and how to manually test with the generated installer.

PR #7551, opened 2023-01-04
Include macOS sandbox files in the Nix binary

edolstra:
This basically reverts https://github.com/NixOS/nix/commit/6e5165b77370c76bfa39d4b55e9f83673f3bd466. It fixes errors like

    sandbox-exec: <internal init prelude>:292:47: unable to open sandbox-minimal.sb: not found

when trying to run a development Nix installed in a user's home directory.

Also, we're trying to minimize the number of installed files to make it possible to deploy Nix as a single statically-linked
binary.


src/libmain/shared.cc
```diff
@@ -363,6 +363,7 @@ void printVersion(const std::string & programName)
             << "\n";
         std::cout << "Store directory: " << settings.nixStore << "\n";
         std::cout << "State directory: " << settings.nixStateDir << "\n";
+        std::cout << "Data directory: " << settings.nixDataDir << "\n";
```
thufschmitt, line 366: Is that related to the rest?

thufschmitt (approved):
Just one weird apparently unrelated change, but looks good otherwise :+1:

src/libmain/shared.cc
edolstra: Not really related, it was just useful during debugging this issue.

PR #7557, opened 2023-01-05
On macOS with auto-uid-allocation and sandboxing, use the correct gid

### LINKED ISSUE #7529: auto-allocate-uids broken when sandbox enabled on macOS
**Describe the bug**

auto-allocate-uids broken when sandbox enabled on macOS

**Steps To Reproduce**

```console
$ cat /etc/nix/nix.conf
# WARNING: this file is generated from the nix.* options in
# your nix-darwin configuration. Do not edit it!
allowed-users = *
auto-optimise-store = false
build-users-group = nixbld
builders =
cores = 0
extra-sandbox-paths =
max-jobs = auto
require-sigs = true
sandbox = true
sandbox-fallback = false
substituters = https://cache.nixos.org/
trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
trusted-substituters =
trusted-users = root
experimental-features = nix-command flakes repl-flake auto-allocate-uids
auto-allocate-uids = true
builders-use-substitutes = true
builders = @/etc/nix/machines


$ nix-info
system: "aarch64-darwin", multi-user?: no, version: nix-env (Nix) 2.12.0, channels(psondej): "", channels(root): "nixpkgs", nixpkgs: /var/root/.nix-defexpr/channels/nixpkgs


$ nix repl nixpkgs
Welcome to Nix 2.12.0. Type :? for help.

Loading installable 'flake:nixpkgs#'...
Added 5 variables.
nix-repl> :b legacyPackages.aarch64-darwin.stdenvNoCC.mkDerivation { name = "pure"; dontUnpack = true; installPhase = '' date > $out ''; }
error: builder for '/nix/store/xyjf1d8mdy31ikkj3byrizy4bv000p4k-pure.drv' failed with exit code 1;
       last 8 log lines:
       > patching sources
       > updateAutotoolsGnuConfigScriptsPhase
       > configuring
       > no configure script, doing nothing
       > building
       > no Makefile, doing nothing
       > installing
       > /nix/store/s5rvpdf77v47cqy1n56wr5xlj3b7vzmm-stdenv-darwin/setup: line 1406: /nix/store/clibgpvcmav6br1b8pqz3lr0r2fk5561-pure: Permission denied
       For full logs, run 'nix log /nix/store/xyjf1d8mdy31ikkj3byrizy4bv000p4k-pure.drv'.



$ nix repl nixpkgs --no-sandbox
Welcome to Nix 2.12.0. Type :? for help.

Loading installable 'flake:nixpkgs#'...
Added 5 variables.
nix-repl> :b legacyPackages.aarch64-darwin.stdenvNoCC.mkDerivation { name = "pure"; dontUnpack = true; installPhase = '' date > $out ''; }

This derivation produced the following outputs:
  out -> /nix/store/clibgpvcmav6br1b8pqz3lr0r2fk5561-pure

```

**Expected behavior**

**Additional context**

**Priorities**

edolstra:
macOS doesn't have user namespacing, so the gid of the builder needs to be `nixbld`. The logic got "has sandboxing enabled" confused with "has user namespaces".

Fixes #7529.

### DIFF
--- a/src/libstore/lock.cc
+++ b/src/libstore/lock.cc
@@ -123,8 +123,12 @@ struct AutoUserLock : UserLock
 
     std::vector<gid_t> getSupplementaryGIDs() override { return {}; }
 
-    static std::unique_ptr<UserLock> acquire(uid_t nrIds, bool useChroot)
+    static std::unique_ptr<UserLock> acquire(uid_t nrIds, bool useUserNamespace)
     {
+        #if !defined(__linux__)
+        useUserNamespace = false;
+        #endif
+
         settings.requireExperimentalFeature(Xp::AutoAllocateUids);
         assert(settings.startId > 0);
         assert(settings.uidCount % maxIdsPerBuild == 0);
@@ -157,7 +161,7 @@ struct AutoUserLock : UserLock
                 auto lock = std::make_unique<AutoUserLock>();
                 lock->fdUserLock = std::move(fd);
                 lock->firstUid = firstUid;
-                if (useChroot)
+                if (useUserNamespace)
                     lock->firstGid = firstUid;
                 else {
                     struct group * gr = getgrnam(settings.buildUsersGroup.get().c_str());
@@ -174,10 +178,10 @@ struct AutoUserLock : UserLock
     }
 };
 
-std::unique_ptr<UserLock> acquireUserLock(uid_t nrIds, bool useChroot)
+std::unique_ptr<UserLock> acquireUserLock(uid_t nrIds, bool useUserNamespace)
 {
     if (settings.autoAllocateUids)
-        return AutoUserLock::acquire(nrIds, useChroot);
+        return AutoUserLock::acquire(nrIds, useUserNamespace);
     else
         return SimpleUserLock::acquire();
 }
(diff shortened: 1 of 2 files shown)


src/libstore/lock.cc
```diff
@@ -123,8 +123,12 @@ struct AutoUserLock : UserLock
 
     std::vector<gid_t> getSupplementaryGIDs() override { return {}; }
 
-    static std::unique_ptr<UserLock> acquire(uid_t nrIds, bool useChroot)
+    static std::unique_ptr<UserLock> acquire(uid_t nrIds, bool useUserNamespace)
     {
+        #if !defined(__linux__)
+        useUserNamespace = false;
+        #endif
```
thufschmitt, line 130: Isn't it a bit weird to conditionally hardcode the value of the parameter? Shouldn't we instead move that logic at the (unique) call site?

edolstra: No, because then the call site can screw it up. That's fine if there is only one call site, but you could imagine multiple users of this class.

thufschmitt: Couldn't we make it an assertion instead then? It's not too big a deal, but that looks quite confusing

PR #7559, opened 2023-01-05
Don't check NixOS modules

ncfavier:
Fix https://github.com/NixOS/nix/issues/6866, act II

Previously: https://github.com/NixOS/nix/pull/7356#issuecomment-1346191130


thufschmitt (approved):
Thanks

PR #7567, opened 2023-01-06
mk: Fix installation of static libraries

### LINKED ISSUE #7021: Statically built Nix doesn't install its libraries
**Describe the bug**

```console
$ nix-build -A pkgsStatic.nix
$ ls result/lib
systemd  tmpfiles.d
```

However, the still installed pkg-config files claim libraries are available:

```console
$ nix-build -A pkgsStatic.nix.dev
$ cat result-dev/lib/pkgconfig/nix-main.pc 
prefix=/nix/store/756yh330xipm9fdncislpp46cxfh33an-nix-static-x86_64-unknown-linux-musl-2.11.0
libdir=/nix/store/756yh330xipm9fdncislpp46cxfh33an-nix-static-x86_64-unknown-linux-musl-2.11.0/lib
includedir=/nix/store/sssclazqrxw13h362k5afq729160zqw3-nix-static-x86_64-unknown-linux-musl-2.11.0-dev/include

Name: Nix
Description: Nix Package Manager
Version: 2.11.0
Libs: -L${libdir} -lnixmain
Cflags: -I${includedir}/nix -std=c++17
```

I think the static archive should be installed when static linking for the benefit of the packages that link against Nix. In nixpkgs of course we should use a separate output for this, so we don't redundantly ship parts of Nix to users of just the executable.

See also https://github.com/NixOS/nixpkgs/issues/188401#issuecomment-1242671304.

zhaofengli:
This PR fixes the installation of static libraries and adds the private linker flags to libnixstore's `.pc` file so it can be more easily used. Fixes #7021.

Tested that the [Attic](https://github.com/zhaofengli/attic) client, which links against libnixstore, builds statically with this change.

### DIFF
--- a/mk/libraries.mk
+++ b/mk/libraries.mk
@@ -130,7 +130,15 @@ define build-library
 
     $(1)_LDFLAGS_USE += $$($(1)_PATH) $$($(1)_LDFLAGS)
 
-    $(1)_INSTALL_PATH := $$(libdir)/$$($(1)_NAME).a
+    $(1)_INSTALL_PATH := $(DESTDIR)$$($(1)_INSTALL_DIR)/$$($(1)_NAME).a
+
+    $$(eval $$(call create-dir, $$($(1)_INSTALL_DIR)))
+
+    $$($(1)_INSTALL_PATH): $$($(1)_OBJS) | $(DESTDIR)$$($(1)_INSTALL_DIR)/
+	+$$(trace-ld) $(LD) -Ur -o $$(_d)/$$($(1)_NAME).o $$^
+	$$(trace-ar) $(AR) crs $$@ $$(_d)/$$($(1)_NAME).o
+
+    install: $$($(1)_INSTALL_PATH)
 
   endif
 
--- a/src/libstore/local.mk
+++ b/src/libstore/local.mk
@@ -69,6 +69,13 @@ $(d)/build.cc:
 
 clean-files += $(d)/schema.sql.gen.hh $(d)/ca-specific-schema.sql.gen.hh
 
+$(d)/nix-store.pc: $(d)/nix-store.pc.in
+	$(trace-gen) rm -f $@ && ./config.status --quiet --file=$@
+ifeq ($(BUILD_SHARED_LIBS), 1)
+	sed -i 's|@LIBS_PRIVATE@||' $@
+else
+	sed -i 's|@LIBS_PRIVATE@|Libs.private: $(libstore_LDFLAGS) $(libstore_LDFLAGS_PROPAGATED) $(foreach lib, $(libstore_LIBS), $($(lib)_LDFLAGS))|' $@
+endif
 $(eval $(call install-file-in, $(d)/nix-store.pc, $(libdir)/pkgconfig, 0644))
 
 $(foreach i, $(wildcard src/libstore/builtins/*.hh), \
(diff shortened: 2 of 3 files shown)


src/libstore/local.mk
```diff
@@ -69,6 +69,13 @@ $(d)/build.cc:
 
 clean-files += $(d)/schema.sql.gen.hh $(d)/ca-specific-schema.sql.gen.hh
 
+$(d)/nix-store.pc: $(d)/nix-store.pc.in
+	$(trace-gen) rm -f $@ && ./config.status --quiet --file=$@
+ifeq ($(BUILD_SHARED_LIBS), 1)
+	sed -i 's|@LIBS_PRIVATE@||' $@
+else
+	sed -i 's|@LIBS_PRIVATE@|Libs.private: $(libstore_LDFLAGS) $(libstore_LDFLAGS_PROPAGATED) $(foreach lib, $(libstore_LIBS), $($(lib)_LDFLAGS))|' $@
```
zhaofengli, line 77: I'm not sure if this is the correct approach. It also exposes the `-L` paths which may be undesirable.

thufschmitt (approved):
That sounds reasonable. I'll let @edolstra have a second look if he wants to since I'm not utterly confident here. Feel free to ping me again in a week if nothing happened.

Ericson2314:

- assigned for @ericson2314 to review

mk/libraries.mk
```diff
@@ -130,7 +130,15 @@ define build-library
 
     $(1)_LDFLAGS_USE += $$($(1)_PATH) $$($(1)_LDFLAGS)
 
-    $(1)_INSTALL_PATH := $$(libdir)/$$($(1)_NAME).a
+    $(1)_INSTALL_PATH := $(DESTDIR)$$($(1)_INSTALL_DIR)/$$($(1)_NAME).a
+
+    $$(eval $$(call create-dir, $$($(1)_INSTALL_DIR)))
```
Ericson2314, line 135: Can we move outside so both branches can use just one of these?

mk/libraries.mk
```diff
@@ -130,7 +130,15 @@ define build-library
...
+
+    $$($(1)_INSTALL_PATH): $$($(1)_OBJS) | $(DESTDIR)$$($(1)_INSTALL_DIR)/
+	+$$(trace-ld) $(LD) -Ur -o $$(_d)/$$($(1)_NAME).o $$^
+	$$(trace-ar) $(AR) crs $$@ $$(_d)/$$($(1)_NAME).o
+
+    install: $$($(1)_INSTALL_PATH)
```
Ericson2314, line 141: Shouldn't it be under a `ifdef $(1)_FORCE_INSTALL`? like above?

mk/libraries.mk
```diff
@@ -130,7 +130,15 @@ define build-library
 
     $(1)_LDFLAGS_USE += $$($(1)_PATH) $$($(1)_LDFLAGS)
 
-    $(1)_INSTALL_PATH := $$(libdir)/$$($(1)_NAME).a
+    $(1)_INSTALL_PATH := $(DESTDIR)$$($(1)_INSTALL_DIR)/$$($(1)_NAME).a
```
Ericson2314, line 133:
```suggestion
    $(1)_INSTALL_PATH := $(DESTDIR)$$($(1)_INSTALL_DIR)/$$($(1)_NAME).$(LIB_EXT)
```
and factor out for both branches

src/libstore/local.mk
Ericson2314: Do we have private libraries? I didn't think so. `--static` can already compute transitive closure just fine, right?

Ericson2314:
Marking as draft until items above are resolved

zhaofengli:
Superseded by the efforts in #2503.

PR #7570, opened 2023-01-08
tests: fix for nixpkgs 22.11

lheckemann:
runCommand now uses stdenvNoCC by default, so that needs to be included instead of the regular stdenv.

### DIFF
--- a/tests/containers.nix
+++ b/tests/containers.nix
@@ -16,7 +16,7 @@ makeTest ({
         { virtualisation.writableStore = true;
           virtualisation.diskSize = 2048;
           virtualisation.additionalPaths =
-            [ pkgs.stdenv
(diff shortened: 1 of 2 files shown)


roberth:
> runCommand now uses stdenvNoCC by default,

It's been that way for years, but `runCommandNoCC` has been demoted to a `aliases.nix` alias after an unintentional brownout.

Did you experience an error? If so, something else must have caused it, and I'd be interested to know what the error was.

That said, this change is probably ok.

roberth (approved):
Ok, found the error in the linked thread; just an attempt to bootstrap a lot in the NixOS test sandbox, because *something* is missing.
I guess we could figure why some derivation output is missing, but on second though that does not seem like a very useful exercise.

PR #7579, opened 2023-01-11
review process: all PRs must be triaged

fricklerhandwerk:
This is only a proposal picking up [@roberth's question](https://github.com/NixOS/nix/pull/7515#issuecomment-1368988316) in order to lead a traceable discussion and come to a decision the team can follow:

in order to make the development process more transparent for everyone,
all pull requests should go through the triage process before getting
merged.

this ensures that all team members are aware of what is going on, and
that rationale for decisions is kept track of in the meeting notes for
posterity. (ideally all that should go into the commit history, but this
is a more invasive process change that needs further deliberation.)

having all team members take at least a brief but conscious look at each
change may also help with keeping our quality standards, as more
reviewers are more likely to remind each other of our shared values.

### DIFF
--- a/maintainers/README.md
+++ b/maintainers/README.md
@@ -79,7 +79,8 @@ Items on the board progress through the following states:
 
   If there is disagreement on the general idea behind an issue or pull request, it is moved to _To discuss_, otherwise to _In review_.
 
-  To ensure process quality and reliability, all pull requests must be triaged before merging.
+  To ensure process quality and reliability, all non-trivial pull requests must be triaged before merging.
+  What constitutes a trivial pull request is up to maintainers' judgement.
 
 - To discuss
 
--- a/maintainers/README.md
+++ b/maintainers/README.md
@@ -56,7 +56,7 @@ Meeting notes are collected on a [collaborative scratchpad](https://pad.lassul.u
 
 The team uses a [GitHub project board](https://github.com/orgs/NixOS/projects/19/views/1) for tracking its work.
 
-Issues on the board progress through the following states:
+Items on the board progress through the following states:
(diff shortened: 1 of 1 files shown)


thufschmitt:
That seems quite counter-productive to me: The goal of the team is to remove a bottleneck in the review process, not to add another one that would barely scale better (or maybe scale even worse).

Quite on the contrary, the team should trust however it appoints as a maintainer to just merge stuff that is _a priori_ not controversial (with the possibility to revert the changes if needs be). And only the rest should be discussed with the team.

> ideally all that should go into the commit history, but this is a more invasive process change that needs further deliberation

I don't think that having the rationale for changes in the history is in any way controversial, is it? We don't really enforce it in general, but it's a common good practice that's broadly followed.

fricklerhandwerk:
@thufschmitt fully agree, we just need a guideline to decide what's controversial and what's not. So far I don't see that at all. This is why as a temporary measure it seems meaningful to force everything through triage, and from collective experience boil down those guidelines. Ideally these documents would also immediately help contributors to make pull requests that are less controversial, or make more controversial changes in a way that are easier to deal with for everyone.

blaggacao:
"Norming. Storming. Performing.", as they say.

It may be interpreted this way:

- Norming &rarr; Architecture & Design Document
- Storming &rarr; Being watched and shaped on the first few PRs ("bootcamp")
- Performing &rarr; being trusted


It may make sense for the Nix Team to identify and graduate those new contributors in their storming phase. But I also don't know where the Nix Team is currently at, at all.

thufschmitt (approved):

PR #7587, opened 2023-01-11
path-info: implement --filter-substitutable

### LINKED ISSUE #3946: support building only drvs that lack substitutes (aka `nix-build-uncached`)

nrdxp:
Fixes #3946 

Alternative to #7526 which implements the same logic in the new cli instead of the legacy `nix-store`

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`

### DIFF
--- a/src/nix/path-info.cc
+++ b/src/nix/path-info.cc
@@ -16,6 +16,7 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
     bool showClosureSize = false;
     bool humanReadable = false;
     bool showSigs = false;
+    bool showSubStatus = false;
 
     CmdPathInfo()
     {
@@ -44,6 +45,14 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
             .longName = "sigs",
             .description = "Show signatures.",
             .handler = {&showSigs, true},
+        });
+
+        addFlag({
+            .longName = "filter-substitutable",
+            .description =
+                "Query path availability in the configured substituters, printing only those that are not available. "
+                "When used with `--json`, a `substitutable` boolean is added to the output.",
+            .handler = {&showSubStatus, true},
         });
     }
 
@@ -86,14 +95,36 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
         for (auto & storePath : storePaths)
             pathLen = std::max(pathLen, store->printStorePath(storePath).size());
 
+        StorePathSet substitutablePaths;
+        if (showSubStatus)
+            substitutablePaths = store->querySubstitutablePaths(StorePathSet(storePaths.begin(), storePaths.end()));
+
         if (json) {
-            std::cout << store->pathInfoToJSON(
+            auto json = store->pathInfoToJSON(
                 // FIXME: preserve order?
                 StorePathSet(storePaths.begin(), storePaths.end()),
-                true, showClosureSize, SRI, AllowInvalid).dump();
+                true, showClosureSize, SRI, AllowInvalid);
+
+            if (showSubStatus) {
+                for (auto & v : json) {
+                    std::string const storePathS = v["path"];
+                    v["substitutable"] = (bool) substitutablePaths.count(store->parseStorePath(storePathS));
+                }
+            }
+            std::cout << json.dump();
         }
 
         else {
+            StorePaths missingPaths;
+
+            if (showSubStatus) {
+                for (auto & path : storePaths) {
+                    if (!substitutablePaths.count(path))
+                      missingPaths.emplace_back(std::move(path));
+                }
+
+                storePaths = missingPaths;
+            }
 
             for (auto & storePath : storePaths) {
                 auto info = store->queryPathInfo(storePath);
--- a/src/nix/path-info.md
+++ b/src/nix/path-info.md
@@ -39,6 +39,15 @@ R""(
 
   ```
 
+* Print all dependant paths which do not exist in any configured binary cache:
+
+  ```console
+  # nix path-info -r --filter-substitutable /nix/store/blzxgyvrk32ki6xga10phr4sby2xf25q-geeqie-1.5.1
+  /nix/store/blzxgyvrk32ki6xga10phr4sby2xf25q-geeqie-1.5.1
+  /nix/store/w2vgpk3rdl6smqz7ixxywqapgagsarjf-fbida-2.14
+  ...
+  ```
+
 * Print the 10 most recently added paths (using --json and the jq(1)
   command):
 


nrdxp:
@roberth, it seems this implementation is fast, but it suffers from only every querying the official cache (cache.nixos.org). Trying to nail down why exactly that is has been sort of a PITA. Do you happen to have any insight into that?

src/nix/path-info.cc
```diff
@@ -86,14 +94,58 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
+
+            if (showSubStatus) {
+                auto jsonNew = nlohmann::json::array();
+
+                for (auto v : json) {
+                    std::string const storePathS = v["path"];
+                    substitutablePaths.find(store->followLinksToStorePath(storePathS)) ==
+                            substitutablePaths.end()
+                        ? v["substitutable"] = false
+                        : v["substitutable"] = true;
```
edolstra, line 130:
```suggestion
                    v["substitutable"] = (bool) substitutablePaths.count(store->followLinksToStorePath(storePathS));
```

src/nix/path-info.cc
```diff
@@ -86,14 +94,58 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
...
+            }
+            else std::cout << json.dump();
         }
 
         else {
+            StorePaths missingPaths;
+
+            if (showSubStatus) {
+                for (auto & path : storePaths) {
+                    if (substitutablePaths.find(path) == substitutablePaths.end())
```
edolstra, line 143:
```suggestion
                    if (!substitutablePaths.count(path))
```

src/nix/path-info.cc
edolstra: Maybe instead of a `substitutable` bool, it should return an array of substituters?

src/nix/path-info.cc
```diff
@@ -86,14 +94,58 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
         for (auto & storePath : storePaths)
             pathLen = std::max(pathLen, store->printStorePath(storePath).size());
 
+        StorePathSet substitutablePaths;
+        if (showSubStatus) {
+            StringSet pathsS;
+            // query each configured substituter
+            for (auto & sub : getDefaultSubstituters()) {
+              if (sub->storeDir != store->storeDir) continue;
```
edolstra, line 102: Indentation is wrong here.

src/nix/path-info.cc
```diff
@@ -86,14 +94,58 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
         for (auto & storePath : storePaths)
             pathLen = std::max(pathLen, store->printStorePath(storePath).size());
 
+        StorePathSet substitutablePaths;
+        if (showSubStatus) {
+            StringSet pathsS;
+            // query each configured substituter
```
edolstra, line 100: This should use `Store::querySubstitutablePaths()`.

edolstra (changes requested):

src/nix/path-info.cc
```diff
@@ -45,6 +46,13 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
             .description = "Show signatures.",
             .handler = {&showSigs, true},
         });
+
+        addFlag({
+            .longName = "--query-substitutes",
+            .shortName = 'U',
```
edolstra, line 52: Not sure whether we should introduce a `shortName` for this. Those should only be used for flags that will be used a lot.

src/nix/path-info.cc
nrdxp: Good idea

src/nix/path-info.cc
nrdxp: Should be fine. I will mostly be using this in scripts anyway, where the long name makes the code easier to follow.

src/nix/path-info.cc
nrdxp: So I tried overriding it but its unreliable it seems unless I override both the `RemoteStore` and `LocalStore` implementations. Also it starts a threadpool in the amount of `fileTransferSettings.httpConnections` and then immediate shuts it down before even starting the queries, not sure why that is as I didn't do that in my override code.

src/nix/path-info.cc
nrdxp: So now that I've figured out why `Store::querySubstitutablePaths` wasn't working: https://github.com/NixOS/nix/pull/7587#discussion_r1068472573, I'm not sure how to easily do this. If I stick with defining the logic myself it should be easy though.

Attempts at using `querySubstitutablePathInfos` result in much slower queries which is why I moved away from that implementation. One possibility would be to add a flag to `queryMissing` to ignore the local store, but that would break the API.

nrdxp:
I'm starting to think that I am just gonna have to make a new sub-command for this functionality, as the `BuiltPathsCommand` doesn't seem like the right choice. It completely removes paths that don't exist locally from the input, which is sorta counterproductive to what we want here.

Also have some async refactor work [here](https://github.com/divnix/nix-uncached/commit/e098900e1522b79babf3a7de2dadfdd47463c17e) that I'd like to include, but that might be better off as a separate PR.

domenkozar:
That's awesome work @nrdxp

src/nix/path-info.cc
```diff
@@ -45,6 +46,14 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
             .description = "Show signatures.",
             .handler = {&showSigs, true},
         });
+
+        addFlag({
+            .longName = "query-substitutes",
+            .description =
+                "Query path availability in the configured substituters, printing only those that are not available."
+                "with `--json` adds a `substitutable` boolean to the output.",
```
edolstra, line 54:
```suggestion
                "Query path availability in the configured substituters, printing only those that are not available. "
                "When used with `--json`, a `substitutable` boolean is added to the output.",
```

src/nix/path-info.cc
```diff
@@ -45,6 +46,14 @@ struct CmdPathInfo : StorePathsCommand, MixJSON
             .description = "Show signatures.",
             .handler = {&showSigs, true},
         });
+
+        addFlag({
+            .longName = "query-substitutes",
```
edolstra, line 51: Maybe this should be named `--filter-substitutable` or something like that? Since it doesn't really query what  the substitutes are.

fricklerhandwerk:
@nrdxp please add a release note.

tomberek:
Oops… misclick.

PR #7593, opened 2023-01-12
install-multi-user: ignore profile_target backups that have no change

miallo:
If there was a prior nix installation that created this backup file and then you tried to install it again, it would stop to tell you there is this file. But if the file and its backup are identical in content, there is no harm in continuing and in a later step overwriting the existing backup file with the identical one. This is just a convenience feature (I ran into the same issue at least three times now).

I did not test this, because I am not sure how to.

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -445,6 +445,14 @@ EOF
         #       a row for different files.
         if [ -e "$profile_target$PROFILE_BACKUP_SUFFIX" ]; then
             # this backup process first released in Nix 2.1
+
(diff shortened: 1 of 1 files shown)


thufschmitt:
Thanks :)

I've pushed it to https://github.com/NixOS/nix/tree/ignore-identical-rc-backup-files-for-installation to get a CI that runs the installer test (bit overkill for just running that script, but that's low-effort for me :stuck_out_tongue: ) to make sure that at least the "normal" install works fine

Ericson2314:

- @thufschmitt will merge this

thufschmitt:
> @thufschmitt will merge this

Sorry, this went out of the radar. But it's merged now

PR #7594, opened 2023-01-12
feat: read installable paths from stdin

### LINKED ISSUE #7437: cli: read "paths" from stdin for commands that take them

nrdxp:
Resolves #7437 for new `nix` commands and for `nix-store` by adding a new flag for these commands `--stdin`.

I decided to add this functionality to the legacy `nix-store` command for primarily one reason: passing a large amount of paths (more than os arg limit) to `nix-store --export`.

If paths are also passed on the cli they will be combined with the ones from standard input.


doc/manual/src/command-ref/nix-store.md
```diff
@@ -54,6 +54,11 @@ have an effect.
     created by sequentially numbering symlinks beyond the first one
     (e.g., `foo`, `foo-2`, `foo-3`, and so on).
 
+  - <span id="opt-stdin">[`--stdin`](#opt-stdin)</span>
+
+    Read *paths…* from the standard input. Useful for chaining
+    nix-store commands.
```
fricklerhandwerk, line 60:
```suggestion
    Read *paths…* from the standard input.
    Useful for chaining `nix-store` commands.
```

nrdxp:
> Looks good, add to release notes?

Sounds good, but it seems the notes are empty. Is there a standard format these should be in? I can find anything in the git log.

tomberek:
> > Looks good, add to release notes?
> 
> Sounds good, but it seems the notes are empty. Is there a standard format these should be in? I can't find a good example in the git log. Something similar to the nixpkgs release notes I imagine.

We place things into https://github.com/NixOS/nix/blob/master/doc/manual/src/release-notes/rl-next.md

nrdxp:
Apologies for delay, release note was added.

blaggacao:
This is an increasingly common issue for us when scripting large closures.

At least according to GH, this is ready for merging?

nrdxp:
> this is ready for merging?

looks like we have a merge conflict now, but I will rebase and it should be ready from there.

thufschmitt (approved):
Yup' that's good.

Thanks @nrdxp, that should be merged as soon as the CI is happy

tomberek:
@nrdxp it seems that this means that:
```
true | nix COMMAND --stdin
```
will have the same impact as `nix COMMAND`, which is a bit surprising.

Should we check for the case of zero installables via stdin and error out?

PR #7595, opened 2023-01-12
nix/show-config: allow getting the value of a specific setting

### LINKED ISSUE #7586: `nix show-config system` should print the value of `system`
None.

It would be nice to not need `jq` or the evaluator to look for e.g. the current system (`nix show-config --json | jq -r .system.value` and `nix eval --impure --raw --expr builtins.currentSystem`), but just do `nix show-config system` directly.

Basically, `nix show-config <key>` should print the value of  `<key>`, without needing to reach for extra tools just to get the value.

None.

**Additional context**
None.

**Priorities**

cole-h:
Instead of needing to run `nix show-config --json | jq -r '."warn-dirty".value'` to view the value of `warn-dirty`, you can now run `nix show-config warn-dirty`.

Fixes #7586.

---

I'm sure this isn't the most beautiful solution (and I couldn't figure out a good way to get JSON support for the single-option mode at this point in time), but it works!

### DIFF
--- a/src/nix/show-config.cc
+++ b/src/nix/show-config.cc
@@ -9,15 +9,44 @@ using namespace nix;
 
 struct CmdShowConfig : Command, MixJSON
 {
+    std::optional<std::string> name;
+
+    CmdShowConfig() {
+        expectArgs({
+            .label = {"name"},
+            .optional = true,
+            .handler = {&name},
+        });
+    }
+
     std::string description() override
     {
-        return "show the Nix configuration";
+        return "show the Nix configuration or the value of a specific setting";
     }
 
     Category category() override { return catUtility; }
 
     void run() override
     {
+        if (name) {
+            if (json) {
+                throw UsageError("'--json' is not supported when specifying a setting name");
+            }
+
+            std::map<std::string, Config::SettingInfo> settings;
+            globalConfig.getSettings(settings);
+            auto setting = settings.find(*name);
+
+            if (setting == settings.end()) {
+                throw Error("could not find setting '%1%'", *name);
+            } else {
+                const auto & value = setting->second.value;
+                logger->cout("%s", value);
+            }
+
+            return;
+        }
+
         if (json) {
             // FIXME: use appropriate JSON types (bool, ints, etc).
             logger->cout("%s", globalConfig.toJSON().dump());
--- a/tests/config.sh
+++ b/tests/config.sh
@@ -51,3 +51,8 @@ exp_features=$(nix show-config | grep '^experimental-features' | cut -d '=' -f 2
 [[ $prev != $exp_cores ]]
 [[ $exp_cores == "4242" ]]
 [[ $exp_features == "flakes nix-command" ]]
+
+# Test that it's possible to retrieve a single setting's value
+val=$(nix show-config | grep '^warn-dirty' | cut -d '=' -f  2 | xargs)
+val2=$(nix show-config warn-dirty)
+[[ $val == $val2 ]]


src/nix/show-config.cc
```diff
@@ -9,15 +9,44 @@ using namespace nix;
...
+            }
+
+            std::map<std::string, Config::SettingInfo> settings;
+            globalConfig.getSettings(settings);
+            auto setting = settings.find(*name);
+
+            if (setting == settings.end()) {
+                throw Error("could not find setting '%1%'", *name);
+            } else {
+                const auto & value = setting->second.value;
```
cole-h, line 43: Honestly, I have no idea when I should `const auto &` and when I shouldn't, so I just did it here since it doesn't seem to hurt :shrug:

src/nix/show-config.cc
```diff
@@ -9,15 +9,44 @@ using namespace nix;
...
+        return "show the Nix configuration or the value of a specific setting";
     }
 
     Category category() override { return catUtility; }
 
     void run() override
     {
+        if (name) {
+            if (json) {
+                throw UsageError("'--json' is not supported when specifying a setting name");
```
thufschmitt, line 33: Shouldn't we allow that? Both for consistency and because some settings are themselves a structured value (`substituters` for example) so it's worth having them already parsed.

src/nix/show-config.cc
thufschmitt: I think that's just fine. As you say, `const`-ing it can't hurt

thufschmitt (changes requested):
Nice, I actually had a talk just yesterday about the fact that it would be handy to have something like that :D 

Can you add some tests for it? `tests/config.sh` seems the right place for that

src/nix/show-config.cc
thufschmitt: Ah, I apparently didn't see the second half of the PR description :man_facepalming: 

Wouldn't it be enough to just print `setting->second.toJSON()` in [here](https://github.com/NixOS/nix/blob/1fc74afbbae1ae3e78e3c0d2096e81c1e24f9d52/src/nix/show-config.cc#L43-L44) ?

cole-h: Sadly, `SettingInfo` doesn't have `toJSON()`... Which is why I left it unsupported :sweat_smile: 

Also, `SettingInfo` is stringly-typed, so I'm not sure if we can easily get (useful) structured values out of it?

cole-h: Though, if `configRegistrations` was somehow exposed, maybe we could go through that, since that appears to have `toJSON` _with_ proper JSON types (that's how `globalConfig.toJSON()` is implemented from what I can see).

cole-h: Derp, it is exposed, just under `globalConfig`. Still, I did a bit of playing and it's real ugly, trying to get the proper `toJSON` stuff. Basically need to be able to do the following:

1. Iterate through all of the `Config`s in `configRegistrations`
1. Gain access to `Config::_settings` (which is `private`) so we can look up the setting by key there
2. Gain access to `AbstractSetting::toJSON()` (which is `protected`)

If there was a general-purpose "C++ value to appropriate JSON value" function somewhere, we could just call that, but I don't know if that exists (all the json files appears to be under libexpr).

edolstra: I don't think we need JSON support here, because you can always get a specific flag by doing `nix show-config --json | jq .<name>.value`.

edolstra (approved):

src/nix/show-config.cc
Ericson2314: I might follow an issue. Whether or not this an important feature (and I readily believe the `jq` way is good enough), it would be nice if our configuration infrastructure didn't get in the way of this.

cole-h: I'd be happy to see json supported, but my attempt was a bit of a pain, and I was in over my head.

PR #7598, opened 2023-01-13
Add escape for systemd service in installer script

yangsheng6810:
Among all the characters that are allowed in a URL, both the percentage sign "%" and the single quotation mark "'" needs escaping when written as a environment variable in a systemd service file. While the single quotation mark may be rare, the percentage sign is widely used to escape characters in a URL. This is especially common in proxy setting, where username and password may contain special characters that need percentage escaping. This patch applies the following replacements:
```
  % -> %%
  ' -> \'
```

### DIFF
--- a/scripts/install-systemd-multi-user.sh
+++ b/scripts/install-systemd-multi-user.sh
@@ -24,12 +24,17 @@ $1
 EOF
 }
 
+escape_systemd_env() {
+    temp_var="${1//\'/\\\'}"
+    echo "${temp_var//\%/%%}"
+}
+
 # Gather all non-empty proxy environment variables into a string
 create_systemd_proxy_env() {
     vars="http_proxy https_proxy ftp_proxy no_proxy HTTP_PROXY HTTPS_PROXY FTP_PROXY NO_PROXY"
     for v in $vars; do
         if [ "x${!v:-}" != "x" ]; then
-            echo "Environment=${v}=${!v}"
+            echo "Environment=${v}=$(escape_systemd_env ${!v})"
         fi
     done
 }


thufschmitt:
Could `systemd-escape` be used instead? Since this is a systemd install, I imagine it's fine to assume that it's present and that would make the logic more robust and future-proof

yangsheng6810:
Actually that was what I tried first, and unfortunately the answer is a "NO". `systemd-escape` is intended for escaping things to be included in unit **names**, a totally different scenario. For example
```
$ systemd-escape /mnt/some-dir
-mnt-some\x2ddir
$ systemd-escape http://127.0.0.1:8080
http:--127.0.0.1:8080
```
And the escaped URL is no longer valid.

Just another thought: while a single quote is technically legal in a URL according to [RFC3986](http://www.faqs.org/rfcs/rfc3986.html), it is rarely used in real life. It may be a good idea **not** to support it, to guard against potential escaping bugs. I do not expect someone to use a proxy with single quote somewhere in its URL. Should it happen, I would expect it already being escaped as `"%27"` in `HTTP_PROXY`.

thufschmitt:
> Actually that was what I tried first, and unfortunately the answer is a "NO". systemd-escape is intended for escaping things to be included in unit names, a totally different scenario. 

Argh, too bad.

> Just another thought: while a single quote is technically legal in a URL according to [RFC3986](http://www.faqs.org/rfcs/rfc3986.html), it is rarely used in real life. It may be a good idea not to support it, to guard against potential escaping bugs.

I don't have a strong opinion here, but I think supporting it is a good idea. It's edge-casy and all, but should work OK

thufschmitt (approved):

PR #7599, opened 2023-01-13
Move `ValidPathInfo` defintions to `path-info.cc`

Ericson2314:
Originally there was no `path-info.*`, then there was `path-info.hh`, then there was `path-info.cc`, but only for new things. Moving this stuff over makes everything consistent.


roberth (approved):
Identical except for an added newline between functions for consitency.

Issue #7261, opened 2022-11-04
`--derivation` considered harmful, remove it in lieu of `^` and new commands


Ericson2314:
#### No single meaning

`--derivation` is a shared flag between all commands, but only some commands care about it at all:
```bash
src/nix/diff-closures.cc:        auto beforePath = Installable::toStorePath(getEvalStore(), store, Realise::Outputs, operateOn, before);
src/nix/diff-closures.cc:        auto afterPath = Installable::toStorePath(getEvalStore(), store, Realise::Outputs, operateOn, after);
src/nix/why-depends.cc:        auto packagePath = Installable::toStorePath(getEvalStore(), store, Realise::Outputs, operateOn, package);
src/nix/why-depends.cc:        auto dependencyPath = Installable::toStorePath(getEvalStore(), store, Realise::Derivation, operateOn, dependency);
```

Other commands force one way or the other:
```bash
 $ git grep OperateOn src/nix
src/nix/develop.cc:                getEvalStore(), store, Realise::Nothing, OperateOn::Output, {installable});
src/nix/develop.cc:            for (auto & path : Installable::toStorePaths(getEvalStore(), store, Realise::Outputs, OperateOn::Output, {bashInstallable})) {
src/nix/run.cc:        auto outPaths = Installable::toStorePaths(getEvalStore(), store, Realise::Outputs, OperateOn::Output, installables);
```

Still other commands force something like `--derivation` by hand, like `nix log`:
https://github.com/NixOS/nix/blob/499e99d099ec513478a2d3120b2af3a16d9ae49d/src/nix/log.cc#L45-L52

This makes for unpredictable behavior, which is a frustrating user experience.

#### Doesn't scale to CA derivations

In the CA derivation world, we cannot assume a path has one unique builder. This is true both for unstable floating CA derivations, and *stable* fixed CA derivations (fixed output derivations). `--derivation` fundamentally is meaningless, or non-determinstic (picking one of many builders) in an arbitrary way. This is no good.

```bash
nix build myFixedOutput0 # produces /nix/store/asdfasdfdasf-foo
nix build myFixedOutput1 # Also produces /nix/store/asdfasdfdasf-foo
nix log /nix/store/asdfasdfdasf-foo # does what?!
```

### Doesn't scale to RFC 92

[...]

PR #7600, opened 2023-01-14
Get rid of `.drv` special-casing for store path installable

Ericson2314:
We no longer treat a store path different based on whether it ends in `.drv`.

To quote the release notes:

> In light of the new `^` syntax for store paths allowing one to be fully explicit, stop special-casing whether a store path in an installable ends in `.drv`.
>
> For example,
> ```shell-session
> $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv
> ```
> gives info about the derivation itself, while
> ```shell-session
> $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv^*
> ```
> gives info about each of its outputs.

# Motivation

## Plumbing CLI should be simple

Store derivation installations are intended as "plumbing": very simple utilities for advanced users and scripts, and not what regular users interact with. (Similarly, regular Git users will use branch and tag names not explicit hashes for most things.)

The plumbing CLI should prize simplicity over convenience; that is its raison d'etre. If the user provides a path, we should treat it the same way not caring what sort of path it is.

## Scripting

This is especially important for the scripting use-case. when arbitrary paths are sent to e.g. `nix copy` and the script author wants consistent behavior regardless of what those store paths are. Otherwise the script author needs to be careful to filter out `.drv` ones, and then run `nix copy` again with those paths and `--derivation`. That is not good!

## Surprisingly low impact

Only two lines in the tests need changing, showing that the impact of this is pretty light.

Many command, like `nix log` will continue to work with just the derivation passed as before. This because we used to:
 - Special case the drv path and replace it with it's outputs (what this gets rid of).
 - Turn those output path *back* into the original drv path.

Now we just skip that entire round trip!

# Context

Issue #7261 lays out a broader vision for getting rid of `--derivation`, and has this as one of its dependencies. But we can do this with or without that.

`Installable::toDerivations` is changed to handle the case of a  `DerivedPath::Opaque` ending in `.drv`, which is new: it simply doesn't need to do any extra work in that case. On this basis, commands like `nix {show-derivation,log} /nix/store/...-foo.drv` still work as before, as described above.

When testing older daemons, the post-build-hook will be run against the old CLI, so we need the old version of the post-build-hook to support that use-case.

# Checklist for maintainers

 - [x] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests

See the change log for user-facing details.

Only two lines in the tests need changing, showing that the impact of this is pretty light.

Progress towards #7261

### DIFF
--- a/doc/manual/src/release-notes/rl-next.md
+++ b/doc/manual/src/release-notes/rl-next.md
@@ -1,2 +1,18 @@
 # Release X.Y (202?-??-??)
 
+* The special handling of an [installable](../command-ref/new-cli/nix.md#installables) with `.drv` suffix being interpreted as all of the given [store derivation](../glossary.md#gloss-store-derivation)'s output paths is removed, and instead taken as the literal store path that it represents.
+
+  The new `^` syntax for store paths introduced in Nix 2.13 allows explicitly referencing output paths of a derivation.
+  Using this is better and more clear than relying on the now-removed `.drv` special handling.
+
+  For example,
+  ```shell-session
+  $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv
+  ```
+
+  now gives info about the derivation itself, while
+
+  ```shell-session
+  $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv^*
+  ```
+  provides information about each of its outputs.
--- a/src/libcmd/installables.cc
+++ b/src/libcmd/installables.cc
@@ -677,9 +677,12 @@ StorePathSet Installable::toDerivations(
         for (const auto & b : i->toDerivedPaths())
             std::visit(overloaded {
                 [&](const DerivedPath::Opaque & bo) {
-                    if (!useDeriver)
-                        throw Error("argument '%s' did not evaluate to a derivation", i->what());
-                    drvPaths.insert(getDeriver(store, *i, bo.path));
+                    drvPaths.insert(
+                        bo.path.isDerivation()
+                            ? bo.path
+                        : useDeriver
+                            ? getDeriver(store, *i, bo.path)
+                        : throw Error("argument '%s' did not evaluate to a derivation", i->what()));
                 },
                 [&](const DerivedPath::Built & bfd) {
                     drvPaths.insert(bfd.drvPath);
--- a/src/libcmd/installable-derived-path.cc
+++ b/src/libcmd/installable-derived-path.cc
@@ -31,27 +31,24 @@ InstallableDerivedPath InstallableDerivedPath::parse(
     ExtendedOutputsSpec extendedOutputsSpec)
 {
     auto derivedPath = std::visit(overloaded {
-        // If the user did not use ^, we treat the output more liberally.
+        // If the user did not use ^, we treat the output more
+        // liberally: we accept a symlink chain or an actual
+        // store path.
         [&](const ExtendedOutputsSpec::Default &) -> DerivedPath {
-            // First, we accept a symlink chain or an actual store path.
             auto storePath = store->followLinksToStorePath(prefix);
-            // Second, we see if the store path ends in `.drv` to decide what sort
-            // of derived path they want.
-            //
-            // This handling predates the `^` syntax. The `^*` in
-            // `/nix/store/hash-foo.drv^*` unambiguously means "do the
-            // `DerivedPath::Built` case", so plain `/nix/store/hash-foo.drv` could
-            // also unambiguously mean "do the DerivedPath::Opaque` case".
-            //
-            // Issue #7261 tracks reconsidering this `.drv` dispatching.
-            return storePath.isDerivation()
-                ? (DerivedPath) DerivedPath::Built {
-                    .drvPath = std::move(storePath),
-                    .outputs = OutputsSpec::All {},
-                }
-                : (DerivedPath) DerivedPath::Opaque {
-                    .path = std::move(storePath),
+            // Remove this prior to stabilizing the new CLI.
+            if (storePath.isDerivation()) {
+                auto oldDerivedPath = DerivedPath::Built {
+                    .drvPath = storePath,
+                    .outputs = OutputsSpec::All { },
                 };
+                warn(
+                    "The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '%s'",
+                    oldDerivedPath.to_string(*store));
+            };
+            return DerivedPath::Opaque {
+                .path = std::move(storePath),
+            };
         },
         // If the user did use ^, we just do exactly what is written.
         [&](const ExtendedOutputsSpec::Explicit & outputSpec) -> DerivedPath {
--- a/tests/binary-cache.sh
+++ b/tests/binary-cache.sh
@@ -23,7 +23,7 @@ nix log $outPath 2>&1 | grep 'is not available'
 nix log --substituters file://$cacheDir $outPath | grep FOO
 
 # Test copying build logs from the binary cache.
-nix store copy-log --from file://$cacheDir $(nix-store -qd $outPath)
+nix store copy-log --from file://$cacheDir $(nix-store -qd $outPath)^'*'
 nix log $outPath | grep FOO
 
 basicDownloadTests() {
--- a/tests/ca/substitute.sh
+++ b/tests/ca/substitute.sh
@@ -28,6 +28,12 @@ nix realisation info --file ./content-addressed.nix transitivelyDependentCA
 nix realisation info --file ./content-addressed.nix dependentCA
 # nix realisation info --file ./content-addressed.nix rootCA --outputs out
 
+if isDaemonNewer "2.13"; then
+    pushToStore="../push-to-store.sh"
+else
+    pushToStore="../push-to-store-old.sh"
+fi
+
 # Same thing, but
 # 1. With non-ca derivations
 # 2. Erasing the realisations on the remote store
@@ -37,7 +43,7 @@ nix realisation info --file ./content-addressed.nix dependentCA
 #
 # Regression test for #4725
 clearStore
-nix build --file ../simple.nix -L --no-link --post-build-hook ../push-to-store.sh
+nix build --file ../simple.nix -L --no-link --post-build-hook "$pushToStore"
 clearStore
 rm -r "$REMOTE_STORE_DIR/realisations"
 nix build --file ../simple.nix -L --no-link --substitute --substituters "$REMOTE_STORE" --no-require-sigs -j0
@@ -52,7 +58,7 @@ if [[ -z "$(ls "$REMOTE_STORE_DIR/realisations")" ]]; then
 fi
 
 # Test the local realisation disk cache
-buildDrvs --post-build-hook ../push-to-store.sh
+buildDrvs --post-build-hook "$pushToStore"
 clearStore
 # Add the realisations of rootCA to the cachecache
 clearCacheCache
--- a/tests/post-hook.sh
+++ b/tests/post-hook.sh
@@ -9,8 +9,14 @@ echo 'require-sigs = false' >> $NIX_CONF_DIR/nix.conf
 
 restartDaemon
 
+if isDaemonNewer "2.13"; then
+    pushToStore="$PWD/push-to-store.sh"
+else
+    pushToStore="$PWD/push-to-store-old.sh"
+fi
+
 # Build the dependencies and push them to the remote store.
-nix-build -o $TEST_ROOT/result dependencies.nix --post-build-hook $PWD/push-to-store.sh
+nix-build -o $TEST_ROOT/result dependencies.nix --post-build-hook "$pushToStore"
 
 clearStore
 
--- a/tests/push-to-store-old.sh
+++ b/tests/push-to-store-old.sh
@@ -7,4 +7,4 @@ set -e
 [ -n "$DRV_PATH" ]
 
 echo Pushing "$OUT_PATHS" to "$REMOTE_STORE"
-printf "%s" "$DRV_PATH"^'*' | xargs nix copy --to "$REMOTE_STORE" --no-require-sigs
+printf "%s" "$DRV_PATH" | xargs nix copy --to "$REMOTE_STORE" --no-require-sigs


doc/manual/src/release-notes/rl-next.md
```diff
@@ -21,3 +21,15 @@
...
+
+* In light of the new `^` syntax for store paths allowing one to be fully explicit, stop special-casing whether a store path in an installable ends in `.drv`.
+
+  For example,
+  ```shell-session
+  $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv`
+  ```
+  gives info about the derivation itself, while
+  ```shell-session
+  $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv^*`
```
abathur, line 33: The test case uses `^'*'`; should it be the same here?

doc/manual/src/release-notes/rl-next.md
```diff
@@ -21,3 +21,15 @@
...
+  ```
+  gives info about of its outputs.
```
abathur, line 35:
```suggestion
  gives info about each of its outputs.
```

src/libcmd/installables.cc
```diff
@@ -1053,9 +1040,12 @@ StorePathSet Installable::toDerivations(
         for (const auto & b : i->toDerivedPaths())
             std::visit(overloaded {
                 [&](const DerivedPath::Opaque & bo) {
-                    if (!useDeriver)
-                        throw Error("argument '%s' did not evaluate to a derivation", i->what());
-                    drvPaths.insert(getDeriver(store, *i, bo.path));
+                    drvPaths.insert(
+                        bo.path.isDerivation()
+                            ? bo.path
```
Ericson2314, line 1045: This allows `nix show-derivation /nix/store/asdfasdfasdf-foo.drv` to work. It would be silly to require `^*` for that!

If we do end up doing all of #7261, I am pretty sure the `useDriver` flag, and possibly this entire function, will disappear altogether.

doc/manual/src/release-notes/rl-next.md
Ericson2314: This is quoting for bash, but I want to say what the syntax is independent of which shell the user is using / which quoting is needed.

edolstra:
Arguably this creates an inconsistency between store derivations and expressions that evaluate to a derivation. E.g.
```
nix path-info nixpkgs#hello
```
operates on the default outputs of `nixpkgs#hello`, *not* the .drv file. For the latter you have to specify `--derivation`.

roberth:
This highlights the importance of being explicit about store level derivations and the expression-level "derivations" (which I propose to call packages instead, #6507).

doc/manual/src/release-notes/rl-next.md
```diff
@@ -1,2 +1,13 @@
 # Release X.Y (202?-??-??)
 
+* In light of the new `^` syntax for store paths allowing one to be fully explicit, stop special-casing whether a store path in an installable ends in `.drv`.
+
+  For example,
+  ```shell-session
+  $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv`
```
Ericson2314, line 7:
```suggestion
  $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv
```

doc/manual/src/release-notes/rl-next.md
Ericson2314:
```suggestion
  $ nix path-info /nix/store/gzaflydcr6sb3567hap9q6srzx8ggdgg-glibc-2.33-78.drv^*
```

src/libcmd/installables.cc
```diff
@@ -815,27 +815,24 @@ std::vector<std::shared_ptr<Installable>> SourceExprCommand::parseInstallables(
...
+                            // Remove this prior to stabilizing the new CLI.
+                            if (storePath.isDerivation()) {
+                                auto oldDerivedPath = DerivedPath::Built {
+                                    .drvPath = storePath,
+                                    .outputs = OutputsSpec::All { },
                                 };
+                                warn(
+                                    "The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '%s'",
+                                    oldDerivedPath.to_string(*store));
+                            };
```
Ericson2314, line 832: @thufschmitt says mention `--derivation` too

roberth:
A possible resolution for the `nix build` verb expectation issue:
 - Add `nix produce <installable>` which does not necessarily build (the same way `nix copy` doesn't)
 - Change `nix build` to build the thing it gets. If the installable produces outputs, realise those; if it produces a drv, realise its outputs.

fricklerhandwerk:
Discussed in the Nix team meeting 2023-01-20:

- @regnat: agree we should have it eventually, but this is too radical
    - should start with a warning
    - @Ericson2314: against of another cycle, would like to have this functionality before any of the related commands are stable
 - to be discussed further by the team

fricklerhandwerk:
Discussed in the Nix team meeting 2023-01-23:

- Change in behavior
    - Currently has a warning telling that the behavior changed
    - But doesn't give people a grace period to do the change more incrementally
        - Maybe we could advise for using `--derivation` for now or switch to the `^*` syntax
    - @edolstra: Inconsistency between passing in the drv path and an higher-level installable (`nix build nixpkgs#hello` will behave differently than `nix build /nix/store/...-hello.drv`)
        - @Ericson2314: Not really inconsitent. It's fine to consider these as different things: the `.drv` is a file while the result of the evaluation is a derivation value
- Digression: Should we rename `nix build` to make it more intuitive with the new behavior?
    - `nix build ...-foo.drv` doesn't _build_ anything
    - Maybe should be called `nix get` instead
        - And also have a `nix build` command that forcibly builds stuff
    - Some commands want to act on the store path (`nix copy`), others directly on the output paths (`nix build`, `nix profile install`)
- no decision

doc/manual/src/release-notes/rl-next.md
```diff
@@ -1,2 +1,13 @@
 # Release X.Y (202?-??-??)
 
+* In light of the new `^` syntax for store paths allowing one to be fully explicit, stop special-casing whether a store path in an installable ends in `.drv`.
```
fricklerhandwerk, line 3:
```suggestion
* The new `^` syntax for store paths allows explicitly referencing output paths of a derivation.
  With this release, the special handling of an [installable](./command-ref/new-cli/nix.md#installables) with `.drv` suffix being interpreted as all of the given [derivation](./glossary#gloss-derivation)'s output paths is removed, and instead taken as the literal store path that it represents. 
```

fricklerhandwerk:
Discussed in the Nix team meeting 2022-02-27:

- Didn't reach an agreement before
    - Broad agreement on that we want the new explicit syntax to take over eventually
    - Contention on how to pull off a smooth transition
- alternatives:
    - don't change the behavior but issue a warning
    - change the behavior and add a warning that it changed
        - warning recommends to use `--derivation` although it will ultimately go away
        - eventually `--derivation` is a no-op
- @thufschmitt: probably people use this in scripts calling `path-info` on many derivations, and in that case I wouldn't want to flood them with warnings
- @fricklerhandwerk: it's experimental, we can just break the behavior
    - @tomberek: it gets us to the end state quicker
        - adding a warning is a concession to existing users of experimental features
        - @fricklerhandwerk: the after-the-fact-warning sounds like interactive release notes. why not put it in the release notes?
    - @Ericson2314: we have been encouraging people to use experimental features in the past, and that was probably a mistake. but those who followed the advice should not pay for our mistakes, so in this case it is sensible to make smaller steps
- agreed on implementation
- @fricklerhandwerk will fix up release notes and merge once tests pass

cole-h:
Hi, what is the right way to use e.g. `nix path-info` on a `.drv` file now? `--derivation` is apparently bad and will go away eventually, so I don't want to reach for that, but the `!*` suggestion from the daemon does not work for me (nor the `^*` suggestion as documented in this PR).

```
$ nix-env --version
nix-env (Nix) 2.15.0pre20230320_1de5b0e

$ nix build nixpkgs#hello && realpath ./result
/nix/store/v02pl5dhayp8jnz8ahdvg5vi71s8xc6g-hello-2.12.1

$ nix path-info -r /nix/store/v02pl5dhayp8jnz8ahdvg5vi71s8xc6g-hello-2.12.1
/nix/store/qmnr18aqd08zdkhka695ici96k6nzirv-libunistring-1.0
/nix/store/vv6rlzln7vhxk519rdsrzmhhlpyb5q2m-libidn2-2.3.2
/nix/store/76l4v99sk83ylfwkz8wmwrm4s8h73rhd-glibc-2.35-224
/nix/store/v02pl5dhayp8jnz8ahdvg5vi71s8xc6g-hello-2.12.1

$ nix-store --query --deriver /nix/store/v02pl5dhayp8jnz8ahdvg5vi71s8xc6g-hello-2.12.1
/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv

$ nix path-info -r /nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv
warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv!*'
don't know how to build these paths:
  /nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv
error: path '/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv' is not valid

$ nix path-info -r '/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv!*'
path '/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv!*' does not contain a 'flake.nix', searching up
error: getting status of '/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv!*': No such file or directory

$ nix path-info -r '/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv^*'
don't know how to build these paths:
  /nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv
error: path '/nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv' is not a valid store path

$ nix path-info --derivation -r /nix/store/v02pl5dhayp8jnz8ahdvg5vi71s8xc6g-hello-2.12.1 | head
/nix/store/001gp43bjqzx60cg345n2slzg7131za8-nix-nss-open-files.patch
/nix/store/01n3wxxw29wj2pkjqimmmjzv7pihzmd7-which-2.21.tar.gz.drv
/nix/store/1qrnbw8xsww3vydd71lwfp32ylgx9i8g-make-wrapper.sh
/nix/store/6xg259477c90a229xwmb53pdfkn6ig3g-default-builder.sh
/nix/store/819fzxfwzp7zhhi4wy5nkapimkb1bsx5-die.sh
/nix/store/5yzw0vhkyszf2d179m0qfkgxmp5wjjx4-move-docs.sh
/nix/store/cickvswrvann041nqxb0rxilc46svw1n-prune-libtool-files.sh
/nix/store/ckzrg0f0bdyx8rf703nc61r3hz5yys9q-builder.sh
/nix/store/d0vsy7cjz9d5vzrndl8962pqq0mv02s3-patch-shebangs.sh
/nix/store/d275wzmimzi3xp4j3vbkvxscmc79q088-strip.sh
```

That last command (with `--derivation`) has the information I was looking for, but I don't want to rely on it because it will eventually go away. Is this just pending work to bring parity between `nix path-info -r --derivation /nix/store/aaaaa-hello` and `nix path-info -r /nix/store/ddd-hello.drv`?

cole-h:
Well, the drv path doesn't exist on my machine, and I can't seem to fetch it with `nix-store -r /nix/store/25i5yk3xxr0g54rab62jfmi2hpmcapiw-hello-2.12.1.drv`, so I guess that might mean it's invalid? It's strange that `nix path-info --derivation` works just fine though.

No eval store or any other funk. The most "strange" my setup is, is that I enable the `flakes` and `nix-command` features.

I only posted here in case I was just holding it wrong, but it might be more valuable to move this to an issue if that is not the case.

raphaelr:
I think the `!*` syntax that's suggested by the new warning message should be a `^*` instead, because `nix build /nix/store/whatever.drv!*` fails:

```
$ nix --version
nix (Nix) 2.16.0pre20230411_ef0b483
$ mkdir tmpstore

$ nix eval nixpkgs#hello.drvPath --store ./tmpstore/
"/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv"

$ nix build --print-out-paths '/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv' --store ./tmpstore/
warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing
try again with '/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv!*'
/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv

$ nix build --print-out-paths '/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv!*' --store ./tmpstore/
path '/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv!*' does not contain a 'flake.nix', searching up
error: getting status of '/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv!*': No such file or directory

$ nix build --print-out-paths '/nix/store/42ipbdz8z12ax0aqchv0rmsy2qb5wi1i-hello-2.12.1.drv^*' --store ./tmpstore/
/nix/store/cfbhfgvn525zg8var9fr67wvmg2z5szl-hello-2.12.1
```

`nix` is from `github:nixos/nix/ef0b48377d0fc79d70455c402ed4df4b18cb93dd` (but I also reproduced this on 2.15.0), and `nixpkgs` is `github:nixos/nixpkgs/ea96b4af6148114421fda90df33cf236ff5ecf1d`.

fricklerhandwerk:
@raphaelr fixing this would be a great first contribution. Feel free to ping me in a PR or on Matrix if you get stuck. We're trying to make onboarding easier, and both the fix and feedback on the process are highly appreciated.

Issue #7417, opened 2022-12-07
allow installables without drvPath


tomberek:
Installables that point to a builtins.storePath or builtins.fetchClosure cannot be used from the command line due to not having a drvPath, yet they can be used if wrapped in a builtins:buildenv or any other derivation wrapper.

**Describe the solution you'd like**
Allow usage of installables (even flake references) without requiring them to have drvPath's.

**Describe alternatives you've considered**
- Wrapping each such installable in a builtins:buildenv, which works.
- I tried following the source. It seems that goals need a name, and for attribute/flakerefs it needs a drvPath to "name" the goal, but otherwise it looks like an artificial limitation.
- `nix build $(nix eval --impure .#some-attr-with-builtins-storePath)` works, which is odd because one would think this could be done automatically

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

Ericson2314:
So concretely today we have
```shell-session
$ nix build --impure --expr 'builtins.storePath "/nix/store/4ywap29s7i8irjz95iz8hakvyc2836bf-neovim-0.7.0/bin/nvim"'
error: expression does not evaluate to a derivation (or a set or list of those)
```
yet
```shell-session
$ nix build /nix/store/4ywap29s7i8irjz95iz8hakvyc2836bf-neovim-0.7.0/bin/nvim
```
works fine.

Strictly speaking, the thing that doesn't work is still a bona fide installable in that the command line parsed successfully, but that it didn't work because the evaluation-based ones are very tricky over what sort of values they allow I think is a bit silly.

Ericson2314:
I think strings which are paths / placeholders (for CA derivation) that match their context should also be allowed.

To take our example above:

```
nix-repl> :p builtins.getContext (builtins.storePath "/nix/store/4ywap29s7i8irjz95iz8hakvyc2836bf-neovim-0.7.0/bin/nvim"){ "/nix/store/4ywap29s7i8irjz95iz8hakvyc2836bf-neovim-0.7.0" = { path = true; }; }
```
the context and string match, so this should be valid.

For another example:
```
nix-repl> hello.out.outPath
"/nix/store/jppfl2bp1zhx8sgs2mgifmsx6dv16mv2-hello-2.12"

nix-repl> :p builtins.getContext hello.out.outPath
[...]

PR #7601, opened 2023-01-14
Make more string values work as installables

Ericson2314:
# Motivation

As discussed in #7417, it would be good to make more string values work as installables. That is to say, if an installable refers to a value, and the value is a string, it used to not work at all, since #7484, it works somewhat, and this PR make it work some more.

The new cases that are added for `BuiltPath` contexts:

- Fixed input- or content-addressed derivation:

  ```
  nix-repl> hello.out.outPath
  "/nix/store/jppfl2bp1zhx8sgs2mgifmsx6dv16mv2-hello-2.12"

  nix-repl> :p builtins.getContext hello.out.outPath
  { "/nix/store/c7jrxqjhdda93lhbkanqfs07x2bzazbm-hello-2.12.drv" = { outputs = [ "out" ]; }; }

  The string matches the specified single output of that derivation, so it should also be valid.

- Floating content-addressed derivation:

  ```
  nix-repl> (hello.overrideAttrs (_: { __contentAddressed = true; })).out.outPath
  "/1a08j26xqc0zm8agps8anxpjji410yvsx4pcgyn4bfan1ddkx2g0"

  nix-repl> :p builtins.getContext (hello.overrideAttrs (_: { __contentAddressed = true; })).out.outPath
  { "/nix/store/qc645pyf9wl37c6qvqzaqkwsm1gp48al-hello-2.12.drv" = { outputs = [ "out" ]; }; }
   ```

  The string is not a path but a placeholder, however it also matches the context, and because it is a CA derivation we have no better option. This should also be valid.

# Context

See #7417, where these ideas were discussed.

The history is broken out to be legible. One can review commit-by-commit to see all the support infra put in place before it is finally used in the last commit.

We may also want to think about richer attrset based values (also discussed in that issue and #6507), but this change "completes" our string-based building blocks, from which the others can be desugared into or at least described/document/taught in terms of.

~~Depends on #7710~~

# Checklist for maintainers

<!-- Contributors: please leave this as is -->

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests

### DIFF
--- a/src/libcmd/installable-attr-path.cc
+++ b/src/libcmd/installable-attr-path.cc
@@ -46,7 +46,15 @@ std::pair<Value *, PosIdx> InstallableAttrPath::toValue(EvalState & state)
 
 DerivedPathsWithInfo InstallableAttrPath::toDerivedPaths()
 {
-    auto v = toValue(*state).first;
+    auto [v, pos] = toValue(*state);
+
+    if (std::optional derivedPathWithInfo = trySinglePathToDerivedPaths(
+        *v,
+        pos,
+        fmt("while evaluating the attribute '%s'", attrPath)))
+    {
+        return { *derivedPathWithInfo };
+    }
 
     Bindings & autoArgs = *cmd.getAutoArgs(*state);
 
--- a/src/libcmd/installable-value.hh
+++ b/src/libcmd/installable-value.hh
@@ -98,6 +98,24 @@ struct InstallableValue : Installable
 
     static InstallableValue & require(Installable & installable);
     static ref<InstallableValue> require(ref<Installable> installable);
+
+protected:
+
+    /**
+     * Handles either a plain path, or a string with a single string
+     * context elem in the right format. The latter case is handled by
+     * `EvalState::coerceToDerivedPath()`; see it for details.
+     *
+     * @param v Value that is hopefully a string or path per the above.
+     *
+     * @param pos Position of value to aid with diagnostics.
+     *
+     * @param errorCtx Arbitrary message for use in potential error message when something is wrong with `v`.
+     *
+     * @result A derived path (with empty info, for now) if the value
+     * matched the above criteria.
+     */
+    std::optional<DerivedPathWithInfo> trySinglePathToDerivedPaths(Value & v, const PosIdx pos, std::string_view errorCtx);
 };
 
 }
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -2318,6 +2318,80 @@ StorePath EvalState::coerceToStorePath(const PosIdx pos, Value & v, NixStringCon
 }
 
 
+std::pair<DerivedPath, std::string_view> EvalState::coerceToDerivedPathUnchecked(const PosIdx pos, Value & v, std::string_view errorCtx)
+{
+    NixStringContext context;
+    auto s = forceString(v, context, pos, errorCtx);
+    auto csize = context.size();
+    if (csize != 1)
+        error(
+            "string '%s' has %d entries in its context. It should only have exactly one entry",
+            s, csize)
+            .withTrace(pos, errorCtx).debugThrow<EvalError>();
+    auto derivedPath = std::visit(overloaded {
+        [&](NixStringContextElem::Opaque && o) -> DerivedPath {
+            return DerivedPath::Opaque {
+                .path = std::move(o.path),
+            };
+        },
+        [&](NixStringContextElem::DrvDeep &&) -> DerivedPath {
+            error(
+                "string '%s' has a context which refers to a complete source and binary closure. This is not supported at this time",
+                s).withTrace(pos, errorCtx).debugThrow<EvalError>();
+        },
+        [&](NixStringContextElem::Built && b) -> DerivedPath {
+            return DerivedPath::Built {
+                .drvPath = std::move(b.drvPath),
+                .outputs = OutputsSpec::Names { std::move(b.output) },
+            };
+        },
+    }, ((NixStringContextElem &&) *context.begin()).raw());
+    return {
+        std::move(derivedPath),
+        std::move(s),
+    };
+}
+
+
+DerivedPath EvalState::coerceToDerivedPath(const PosIdx pos, Value & v, std::string_view errorCtx)
+{
+    auto [derivedPath, s_] = coerceToDerivedPathUnchecked(pos, v, errorCtx);
+    auto s = s_;
+    std::visit(overloaded {
+        [&](const DerivedPath::Opaque & o) {
+            auto sExpected = store->printStorePath(o.path);
+            if (s != sExpected)
+                error(
+                    "path string '%s' has context with the different path '%s'",
+                    s, sExpected)
+                    .withTrace(pos, errorCtx).debugThrow<EvalError>();
+        },
+        [&](const DerivedPath::Built & b) {
+            // TODO need derived path with single output to make this
+            // total. Will add as part of RFC 92 work and then this is
+            // cleaned up.
+            auto output = *std::get<OutputsSpec::Names>(b.outputs).begin();
+
+            auto drv = store->readDerivation(b.drvPath);
+            auto i = drv.outputs.find(output);
+            if (i == drv.outputs.end())
+                throw Error("derivation '%s' does not have output '%s'", store->printStorePath(b.drvPath), output);
+            auto optOutputPath = i->second.path(*store, drv.name, output);
+            // This is testing for the case of CA derivations
+            auto sExpected = optOutputPath
+                ? store->printStorePath(*optOutputPath)
+                : downstreamPlaceholder(*store, b.drvPath, output);
+            if (s != sExpected)
+                error(
+                    "string '%s' has context with the output '%s' from derivation '%s', but the string is not the right placeholder for this derivation output. It should be '%s'",
+                    s, output, store->printStorePath(b.drvPath), sExpected)
+                    .withTrace(pos, errorCtx).debugThrow<EvalError>();
+        }
+    }, derivedPath.raw());
+    return derivedPath;
+}
+
+
 bool EvalState::eqValues(Value & v1, Value & v2, const PosIdx pos, std::string_view errorCtx)
 {
     forceValue(v1, noPos);
--- a/src/libexpr/eval.cc
+++ b/src/libexpr/eval.cc
@@ -1047,6 +1047,27 @@ void EvalState::mkStorePathString(const StorePath & p, Value & v)
 }
 
 
+void EvalState::mkOutputString(
+    Value & value,
+    const StorePath & drvPath,
+    const std::string outputName,
+    std::optional<StorePath> optOutputPath)
+{
+    value.mkString(
+        optOutputPath
+            ? store->printStorePath(*std::move(optOutputPath))
+            /* Downstream we would substitute this for an actual path once
+               we build the floating CA derivation */
+            : downstreamPlaceholder(*store, drvPath, outputName),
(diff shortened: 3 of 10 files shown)


roberth:
Seems like this would solve #7654? Maybe allow the whole string to be printed in the repl and CLI?

src/libcmd/installables.cc
```diff
@@ -617,17 +617,49 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
...
+                           throw Error(
+                               "flake output attribute '%s' evaluates to the string '%s' which should be '%s' but is not",
+                               attrPath, s, sExpected);
+                       return DerivedPath::Opaque {
+                           .path = std::move(o.path),
+                       };
+                    },
+                    [&](NixStringContextElem::DrvDeep & o) -> DerivedPath {
+                       throw Error(
+                           "flake output attribute '%s' evaluates to the string '%s' which unsupported String Context",
```
tomberek, line 640:
```suggestion
                           "flake output attribute '%s' evaluates to the string '%s' which contains a String Context for a derivation file",
```

src/libcmd/installables.cc
```diff
@@ -617,17 +617,49 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
...
+                           attrPath, s);
+                    },
+                    [&](NixStringContextElem::Built & o) -> DerivedPath {
+                       auto drv = state->store->readDerivation(o.drvPath);
+                       auto i = drv.outputs.find(o.output);
+                       if (i == drv.outputs.end())
+                           throw Error("derivation '%s' does not have output '%s'", state->store->printStorePath(o.drvPath), o.output);
```
Ericson2314, line 647:
```suggestion
                           throw Error("flake output attribute '%s' refers to derivation '%s' and output '%s', does the derivation does not have that output", state->store->printStorePath(o.drvPath), o.output);
```

src/libcmd/installables.cc
```diff
@@ -617,17 +617,49 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
...
+                       auto optOutputPath = i->second.path(*state->store, drv.name, o.output);
+                       auto sExpected = optOutputPath
```
tomberek, line 649:
```suggestion
                       # This is testing for the case of CA derivations
                       auto sExpected = optOutputPath
```
and we should confirm that this derivation is indeed CA.

src/libcmd/installables.cc
```diff
@@ -617,17 +617,49 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
...
+                           ? state->store->printStorePath(*optOutputPath)
+                           : downstreamPlaceholder(*state->store, o.drvPath, o.output);
+                       if (s != sExpected)
+                           throw Error(
+                               "flake output attribute '%s' evaluates to the string '%s' which should be '%s' but is not",
```
tomberek, line 654:
```suggestion
                               "flake output attribute '%s' evaluates to the string '%s' which should be '%s', usage as an installable requires the string and its context to be equal",
```

src/libcmd/installables.cc
```diff
@@ -617,17 +617,49 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
...
+                    "flake output attribute '%s' evaluates to the string '%s' which does not denote exactly one store",
+                    attrPath, s);
+            return {{
+                .path = std::visit(overloaded {
+                    [&](NixStringContextElem::Opaque & o) -> DerivedPath {
+                       auto sExpected = state->store->printStorePath(o.path);
+                       if (s != sExpected)
+                           throw Error(
+                               "flake output attribute '%s' evaluates to the string '%s' which should be '%s' but is not",
+                               attrPath, s, sExpected);
```
Ericson2314, line 633:
```suggestion
                               "flake output attribute '%s' evaluates to the path string '%s', but its context refers to something else not '%s' as expected",
                               attrPath, s, sExpected);
```

src/libcmd/installables.cc
Ericson2314, line 654:
```suggestion
                               "flake output attribute '%s' evaluates to the string '%s' referring to '%s', but the string is not the right place holder for this derivation output. It should be should be '%s'",
```

src/libcmd/installables.cc
```diff
@@ -617,17 +617,49 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
...
+                       if (i == drv.outputs.end())
+                           throw Error("derivation '%s' does not have output '%s'", state->store->printStorePath(o.drvPath), o.output);
+                       auto optOutputPath = i->second.path(*state->store, drv.name, o.output);
+                       auto sExpected = optOutputPath
+                           ? state->store->printStorePath(*optOutputPath)
+                           : downstreamPlaceholder(*state->store, o.drvPath, o.output);
+                       if (s != sExpected)
+                           throw Error(
+                               "flake output attribute '%s' evaluates to the string '%s' which should be '%s' but is not",
+                               attrPath, s, sExpected);
```
Ericson2314, line 655:
```suggestion
                               attrPath, s, o.to_string(*state.store), sExpected);
```

src/libcmd/installables.cc
```diff
@@ -617,17 +617,49 @@ DerivedPathsWithInfo InstallableFlake::toDerivedPaths()
...
-                        .path = std::move(*storePath),
-                    }
-                }};
-            } else
-                throw Error("flake output attribute '%s' evaluates to the string '%s' which is not a store path", attrPath, s);
+            auto context = v.getContext(*state->store);
+            std::string_view s = state->forceString(v, noPos, fmt("while evaluating the flake output attribute '%s'", attrPath));
+            if (context.size() != 1)
+                throw Error(
+                    "flake output attribute '%s' evaluates to the string '%s' which does not denote exactly one store",
```
Ericson2314, line 624:
```suggestion
                    "flake output attribute '%s' evaluates a string '%s' which which has multiple entries in its context. It should only have one entry",
```

src/libcmd/installables.cc
Ericson2314:
```suggestion
                           "flake output attribute '%s' evaluates to the string '%s' which has an unsupported context",
```

Ericson2314:
```suggestion
                           "flake output attribute '%s' evaluates to the string '%s' whose context context refers to a complete source and binary closure. This is not supported at this time",
```

src/libcmd/installables.cc
Ericson2314:
```suggestion
                       // This is testing for the case of CA derivations
                       auto sExpected = optOutputPath
```

src/libcmd/installable-attr-path.cc
```diff
@@ -48,6 +48,26 @@ DerivedPathsWithInfo InstallableAttrPath::toDerivedPaths()
...
+            .info = make_ref<ExtraPathInfo>(),
+        }};
+    }
+
+    else if (v->type() == nString) {
+        return {{
+            .path = state->coerceToDerivedPath(
+                noPos,
+                *v,
+                ""),
```
roberth, line 66: > `""`

Shouldn't this include the attribute path?
We can't assume that it's obvious, because there could be multiple paths, and even a single path may not be apparent to the user when used in a script.

If the attr path is added elsewhere, then the equivalent for the flake attribute case might be redundant, unless it's only redundant when `--show-trace` is enabled.

src/libcmd/installable-attr-path.cc
```diff
@@ -48,6 +48,26 @@ DerivedPathsWithInfo InstallableAttrPath::toDerivedPaths()
...
+            .info = make_ref<ExtraPathInfo>(),
+        }};
+    }
```
roberth, line 69: These two conditionals duplicate the equivalent flake attribute logic, except for the `errorCtx` string, which should remain a parameter.
`std::optional<DerivedPathsWithInfo> trySinglePathToDerivedPaths(Value *v, string_view errorCtx)`?

src/libexpr/eval.cc
```diff
@@ -2296,6 +2317,57 @@ StorePath EvalState::coerceToStorePath(const PosIdx pos, Value & v, NixStringCon
     error("path '%1%' is not in the Nix store", path).withTrace(pos, errorCtx).debugThrow<EvalError>();
 }
 
+DerivedPath EvalState::coerceToDerivedPath(const PosIdx pos, Value & v, std::string_view errorCtx, bool validate)
```
roberth, line 2320: `validate` seems to be for unit testing only.

Is that really necessary? It's not good for coverage either.

roberth (changes requested):
Needs a bit of cleanup.

src/libexpr/eval.cc
```diff
@@ -2296,6 +2317,75 @@ StorePath EvalState::coerceToStorePath(const PosIdx pos, Value & v, NixStringCon
     error("path '%1%' is not in the Nix store", path).withTrace(pos, errorCtx).debugThrow<EvalError>();
 }
 
+std::pair<DerivedPath, std::string_view> EvalState::coerceToDerivedPathUnchecked(const PosIdx pos, Value & v, std::string_view errorCtx)
+{
+    NixStringContext context;
+    auto s = forceString(v, context, pos, errorCtx);
+    if (context.size() != 1)
+        error("string '%s' has multiple entries in its context. It should only have one entry", s)
```
tomberek, line 2325: Error reads a bit odd when the supplied string has zero context.

Ericson2314: Now has a number instead, and we test for both 0 and 2

src/libcmd/installable-value.hh
```diff
@@ -98,6 +98,24 @@ struct InstallableValue : Installable
...
     static ref<InstallableValue> require(ref<Installable> installable);
+
+protected:
+
+    /**
+     * Handles either a plain path, or a string with a single string
+     * context elem in the right format. The latter case is handled by
+     * `EvalState::coerceToDerivedPath()`; see it for details.
+     *
+     * @param v Value that is a string or path per the above.
```
roberth, line 109:
```suggestion
     * @param v Value that is hopefully a string or path per the above.
```

src/libcmd/installable-value.hh
```diff
@@ -98,6 +98,24 @@ struct InstallableValue : Installable
...
+     *
+     * @param pos Position of value to aid with diagnostics.
+     *
+     * @param errorCtx Arbitrary message for diagnostics.
```
roberth, line 113:
```suggestion
     * @param errorCtx Arbitrary message for use in potential error message when something is wrong with `v`.
```

roberth (commented):
Almost there :rocket:

tests/build.sh
```diff
@@ -57,6 +57,30 @@ nix build -f multiple-outputs.nix --json 'e^*' --no-link | jq --exit-status '
...
+  | grepQuiet "has 0 entries in its context. It should only have exactly one entry"
+
+# Too much string context
+expectStderr 1 nix build --impure --expr 'with (import ./multiple-outputs.nix).e.a_a; "${drvPath}${outPath}"' --no-link \
+  | grepQuiet "has 2 entries in its context. It should only have exactly one entry"
+
+nix build --impure --json --expr 'builtins.unsafeDiscardOutputDependency (import ./multiple-outputs.nix).e.a_a.drvPath' --no-link | jq --exit-status '
+  (.[0] | .path | match(".*multiple-outputs-e.drv"))
+'
+
```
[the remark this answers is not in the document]
roberth: Possibly in a different file where ca derivations is enabled

roberth (approved):

Issue #7215, opened 2022-10-25
Nix Installer fails on macOS if /etc/bash.bashrc.backup-before-nix already exists


mpscholten:
**Is your feature request related to a problem? Please describe.**
I've seen a couple times that the nix installer fails when a file at `/etc/bash.bashrc.backup-before-nix` already exists. E.g. yesterday a colleague tried to install nix, and as he had nix installed previously at that macbook he got that error saying that the nix installer cannot proceed until we delete the `/etc/bash.bashrc.backup-before-nix` file. After that file was deleted, we reran the installer and  now got the same error but for the `zshrc.backup-before-nix` file.

**Describe the solution you'd like**
The installer should just take care of this automatically. Either ask the user what to do, or even better just automatically rename e.g. `zshrc.backup-before-nix` to `zshrc.backup-before-nix2`.

abathur:
They'll have to follow the uninstall instructions before reinstalling it: https://nixos.org/manual/nix/stable/installation/installing-binary.html#macos

If they only tried reinstalling because nix wasn't on PATH after taking a macos update, they don't have to reinstall. The update is overwriting the shell hook, but they can just replace it. See #3616 for more.

mpscholten:
This specific case was not directly related to a macos update. It just looked like nix was not installed on that machine (likely it got broken by some macos update in the past) and we tried to run the installer.

Running the uninstaller only comes to mind when you're aware that nix is already installed.

iFreilicht:
One potential way of fixing this could be check whether the desired lines are already in `/etc/bash` and just not modifying it if that is the case.

Issue #7069, opened 2022-09-20
/etc/bash.bashrc.backup-before-nix contains NIX information 


VanessaPC:
**Describe the bug**

According to the instructions I am getting on failure I get: 

2. Ensure /etc/bashrc.backup-before-nix does not have anything
   Nix-related in it. If it does, something is probably quite
   wrong. Please open an issue or get in touch immediately.

However,the files contains this lines 

```
# Nix
if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
  . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
fi
# End Nix
```

**Steps To Reproduce**

**Expected behavior**

**`nix-env --version` output**

**Additional context**

abathur:
It's a sign nix was installed before the backups were taken, so you may need to clean it up manually.

I'd try:
1. Run `grep "End Nix" /etc/*` and manually remove those lines from any matching files that don't have the backup-before-nix extension.
2. Remove all of the files with the backup-before-nix extension.
3. Follow the remaining appropriate macOS/Linux uninstall instructions before reinstalling it: https://nixos.org/manual/nix/stable/installation/installing-binary.html#uninstalling

VanessaPC:
Thank you. 

I did that, as well as uninstalling from mac, it then worked.

Issue #7013, opened 2022-09-07
Failed re-install on Mac OS 12.5.1


savil:
**Describe the bug**

1. I followed the install steps at https://nixos.org/download.html#nix-install-macos
2. I had an old nix install from a few weeks ago, and the install script prompted me to backup the bashrc and zshrc files. I did the backups and re-ran the install script.
3. The install script then did a fair amount (see output below), and then failed.

**Steps To Reproduce**

1. Have an old install on mac os
2. run the install script
3. Hopefully (not!) that will fail too?

**Expected behavior**

expected a full install to succeed.

**Additional context**

```➜ sh <(curl -L https://nixos.org/nix/install)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  4064  100  4064    0     0  15402      0 --:--:-- --:--:-- --:--:-- 3968k
downloading Nix 2.11.0 binary tarball for x86_64-darwin from 'https://releases.nixos.org/nix/nix-2.11.0/nix-2.11.0-x86_64-darwin.tar.xz' to '/var/folders/zv/r3sx92_94gq86_rq3yn1ky2h0000gn/T/nix-binary-tarball-unpack.XXXXXXXXXX.e1DdTxR2'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 23.4M  100 23.4M    0     0  47.0M      0 --:--:-- --:--:-- --:--:-- 47.7M
Switching to the Multi-user Installer
Welcome to the Multi-User Nix Installation

This installation tool will set up your computer with the Nix package
manager. This will happen in a few stages:

1. Make sure your computer doesn't already have Nix. If it does, I
   will show you instructions on how to clean up your old install.

2. Show you what I am going to install and where. Then I will ask
   if you are ready to continue.

3. Create the system users and groups that the Nix daemon uses to run
   builds.

4. Perform the basic installation of the Nix files daemon.

5. Configure your shell to import special Nix Profile files, so you
   can use Nix.

6. Start the Nix daemon.

Would you like to see a more detailed list of what I will do?
[y/n] n


---- let's talk about sudo -----------------------------------------------------
This script is going to call sudo a lot. Every time I do, it'll
output exactly what it'll do, and why.

Just like this:

[...]

Issue #6912, opened 2022-08-17
Nix not working after Monterey update


camilafracaro:
**Describe the bug**

My nix stopped working after the migration to Monterey. I've run `curl -L https://nixos.org/nix/install | sh` and followed the steps 1-3. and now I have this error:

```
~~> Installing Nix
cp: cannot overwrite directory /nix/store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Resources with non-directory ./store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Resources
cp: cannot overwrite directory /nix/store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Versions/A/Resources with non-directory ./store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Versions/A/Resources
cp: cannot overwrite directory /nix/store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Versions/Current with non-directory ./store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Versions/Current
cp: cannot overwrite directory /nix/store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Headers with non-directory ./store/3h7vf3h57xp29p892iqvvq54h8nkv297-apple-framework-IOKit/Library/Frameworks/IOKit.framework/Headers
cp: cannot overwrite directory /nix/store/5crz0l3klajnqchl0zll1ia19n0ljs32-swift-corefoundation-unstable-2018-09-14/Library/Frameworks/CoreFoundation.framework/Versions/Current with non-directory ./store/5crz0l3klajnqchl0zll1ia19n0ljs32-swift-corefoundation-unstable-2018-09-14/Library/Frameworks/CoreFoundation.framework/Versions/Current
[...]

PR #7603, opened 2023-01-15
Make multi-user installer idempotent

### LINKED ISSUE #6679: multi-user-installer: restore copy-to-store idempotence?
Before #5150, this section of the multi-user installer used rsync and was (IIRC) ~idempotent:

https://github.com/NixOS/nix/blob/7e301fd74e929af916116a04d84e130f9e0112de/scripts/install-multi-user.sh#L759-L768

Dropping the rsync dependency to support debian makes sense, but it would be nice to figure out _whether_ we can portably make this step idempotent again--and do it, if it's simple enough.

Shooting from the hip:
- this might be as simple as using cp's `-f` flag, if it's portable?
- it's probably not as simple as just ignoring the exit code from cp, since there could be other issues
- it's probably not as simple as just diffing the directories, since an existing store will contain more than the store seed

(If we can't do this without, say, wiping the store or restoring the rsync dependency, we should probably figure out how to fail faster until we can find another approach.)

### LINKED ISSUE #7548: Reinstall doesnt work
## i got this message and i did all the instructions, but it still doesnt work

I back up shell profile/rc scripts before I add Nix to them.
I need to back up /etc/bashrc to /etc/bashrc.backup-before-nix,
but the latter already exists.

Here's how to clean up the old backup file:

1. Back up (copy) /etc/bashrc and /etc/bashrc.backup-before-nix
   to another location, just in case.

2. Ensure /etc/bashrc.backup-before-nix does not have anything
   Nix-related in it. If it does, something is probably quite
   wrong. Please open an issue or get in touch immediately.

3. Once you confirm /etc/bashrc is backed up and
   /etc/bashrc.backup-before-nix doesn't mention Nix, run:
   mv /etc/bashrc.backup-before-nix /etc/bashrc

We'd love to help if you need it.

You can open an issue at https://github.com/nixos/nix/issues

Or feel free to contact the team:
 - Matrix: #nix:nixos.org
 - IRC: in #nixos on irc.libera.chat
 - twitter: @nixos_org
 - forum: https://discourse.nixos.org

iFreilicht:
This will fix #6679 and #7215. 

Additionally, this PR

* should fix #7548 and #2445
* will at least fix the cause of (but not necessarily the user's acute problem)
  * #7069
  * #6617
* should help resolve by allowing re-installation
  * #6117
  * #3616
* Will fix all `cp: cannot overwrite directory ... with non-directory` errors
  * #7573
  * #7177
  * #7030
  * #7013
  * #6970
  * #6912
  * #6670
  * #6584
  * #6511
  * #5959

### DIFF
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -441,19 +441,31 @@ $(uninstall_directions)
 EOF
     fi
 
-    # TODO: I think it would be good for this step to accumulate more
-    #       knowledge of older obsolete artifacts, if there are any.
-    #       We could issue a "reminder" here that the user might want
-    #       to clean them up?
-
+    problematic_profile_targets=()
     for profile_target in "${PROFILE_TARGETS[@]}"; do
-        # TODO: I think it would be good to accumulate a list of all
-        #       of the copies so that people don't hit this 2 or 3x in
-        #       a row for different files.
-        if [ -e "$profile_target$PROFILE_BACKUP_SUFFIX" ]; then
-            # this backup process first released in Nix 2.1
-            failure <<EOF
+        # this backup process first released in Nix 2.1
+        if [ -e "$profile_target$PROFILE_BACKUP_SUFFIX" ] && ! contains_source_lines "$profile_target"; then
+            # Identical backup and original happens especially on macOS as it overwrites /etc/zshrc on every update
+            if cmp -s "$profile_target" "$profile_target$PROFILE_BACKUP_SUFFIX"; then
+                rm "$profile_target$PROFILE_BACKUP_SUFFIX"
+            else
+                problematic_profile_targets+=("$profile_target")
+            fi
+        fi
+    done
+
+    if [ "${PROFILE_TARGETS[@]}" ]; then
+        resolution_explanation=$(cat <<EOF
 I back up shell profile/rc scripts before I add Nix to them.
+However, at least one of them was already backed up.
+Clean them up manually to continue.
+EOF
+)
+
+        for profile_target in "${problematic_profile_targets[@]}"; do
+            resolution_explanation+=$(cat <<EOF
+
+
 I need to back up $profile_target to $profile_target$PROFILE_BACKUP_SUFFIX,
 but the latter already exists.
 
@@ -470,8 +482,11 @@ Here's how to clean up the old backup file:
    $profile_target$PROFILE_BACKUP_SUFFIX doesn't mention Nix, run:
    mv $profile_target$PROFILE_BACKUP_SUFFIX $profile_target
 EOF
-        fi
-    done
+)
+        done
+
+        failure "$resolution_explanation"
+    fi
 
     if is_os_linux && [ ! -e /run/systemd/system ]; then
         warning <<EOF
@@ -862,6 +877,11 @@ end
 EOF
 }
 
+contains_source_lines() {
+    # This works for both POSIX shells and fish
+    grep -q "# Nix" "$1" && grep -q "# End Nix" "$1"
+}
+
 configure_shell_profile() {
     task "Setting up shell profiles: ${PROFILE_TARGETS[*]}"
     for profile_target in "${PROFILE_TARGETS[@]}"; do
@@ -877,7 +897,7 @@ configure_shell_profile() {
             fi
         fi
 
-        if [ -e "$profile_target" ]; then
+        if [ -e "$profile_target" ] && ! contains_source_lines "$profile_target"; then
             shell_source_lines \
                 | _sudo "extend your $profile_target with nix-daemon settings" \
                         tee -a "$profile_target"
--- a/scripts/install-multi-user.sh
+++ b/scripts/install-multi-user.sh
@@ -816,7 +816,7 @@ install_from_extracted_nix() {
         cd "$EXTRACTED_NIX_PATH"
 
         _sudo "to copy the basic Nix files to the new store at $NIX_ROOT/store" \
-              cp -RPp ./store/* "$NIX_ROOT/store/"
+              cp -RPnp ./store/* "$NIX_ROOT/store/"
 
         _sudo "to make the new store non-writable at $NIX_ROOT/store" \
               chmod -R ugo-w "$NIX_ROOT/store/"
--- a/tests/installer/default.nix
+++ b/tests/installer/default.nix
@@ -131,7 +131,8 @@ let
     with nixpkgsFor.${image.system}.native;
     runCommand
       "installer-test-${imageName}-${testName}"
-      { buildInputs = [ qemu_kvm openssh ];
+      {
+        buildInputs = [ qemu_kvm openssh ];
         image = image.image;
         postBoot = image.postBoot or "";
         installScript = installScripts.${testName}.script;
@@ -230,15 +231,40 @@ let
           [[ \$(nix-instantiate --eval --expr 'builtins.readFile <myChannel/someFile>') = '"someContent"' ]]
         EOF
 
+        echo "Running installer again to test for idempotency..."
+        $ssh "set -eux; $installScript"
+
+        echo "Testing Nix installation..."
+        $ssh <<EOF
+          set -ex
+
+          # FIXME: get rid of this; ideally ssh should just work.
+          source ~/.bash_profile || true
+          source ~/.bash_login || true
+          source ~/.profile || true
+          source /etc/bashrc || true
+
+          nix-env --version
+          nix --extra-experimental-features nix-command store ping
+
+          out=\$(nix-build --no-substitute -E 'derivation { name = "foo"; system = "x86_64-linux"; builder = "/bin/sh"; args = ["-c" "echo foobar > \$out"]; }')
+          [[ \$(cat \$out) = foobar ]]
+        EOF
+
         echo "Done!"
         touch $out
       '';
 
 in
 
-builtins.mapAttrs (imageName: image:
-  { ${image.system} = builtins.mapAttrs (testName: test:
-      makeTest imageName testName
-    ) installScripts;
-  }
-) images
+builtins.mapAttrs
+  (imageName: image:
+    {
+      ${image.system} = builtins.mapAttrs
+        (testName: test:
+          makeTest imageName testName
+        )
+        installScripts;
+    }
+  )
+  images


abathur:
There are instructions at https://nixos.org/manual/nix/stable/contributing/hacking.html#installer-tests for setting up the installer jobs to run in your own fork.

abathur:
I guess you'd need to manually test the generated installer or maybe tweak the install test workflow to run the job twice.

iFreilicht:
Alright, I made the multi-user installer fully idempotent now, and the test was successful! 

```
$ nix-build -A hydraJobs.installerTests.ubuntu-22-04.x86_64-linux.install-force-daemon
  /nix/store/dv4dmxf445xpqydbg8pp533ld3yajgaz-nix-2.13.0pre19700101_61734d2.drv
  /nix/store/h48ibf3hi96r4hwjjfimajmnrva94dsf-closure-info.drv
  /nix/store/ralvjga518dld7swlgdg8hpwwkm2lql0-nix-binary-tarball-2.13.0pre19700101_61734d2.drv
  /nix/store/85b3kqs3dhy78j0r9bfjlcr9lm9vpi7f-installer-test-ubuntu-22-04-install-force-daemon.drv
building '/nix/store/dv4dmxf445xpqydbg8pp533ld3yajgaz-nix-2.13.0pre19700101_61734d2.drv'...
unpacking sources
[...]
building '/nix/store/85b3kqs3dhy78j0r9bfjlcr9lm9vpi7f-installer-test-ubuntu-22-04-install-force-daemon.drv'...
Unpacking Vagrant box /nix/store/pkkfzasqwc8lay1m360xlc9nvixbaqk4-libvirt.box...
Vagrantfile
box.img
info.json
metadata.json
Formatting './disk.qcow2', fmt=qcow2 cluster_size=65536 extended_l2=off compression_type=zlib size=137438953472 backing_file=./box.img backing_fmt=qcow2 lazy_refcounts=off refcount_bits=16
Starting qemu...
Waiting for SSH...
[ssh] Trying to connect...
cSeaBIOS (version rel-1.16.0-0-gd239552ce722-prebuilt.qemu.org)


iPXE (http://ipxe.org) 00:03.0 CA00 PCI2.10 PnP PMM+BFF911E0+BFEF11E0 CA00



Booting from Hard Disk...
[ssh] Trying to connect...
[ssh] Trying to connect...
[ssh] Trying to connect...
[ssh] Trying to connect...
[ssh] Trying to connect...
[ssh] Connected!
Copying installer...
Running installer...
+ tar -xf ./nix.tar.xz
+ mv ./nix-2.13.0pre19700101_61734d2-x86_64-linux nix
+ ./nix/install --daemon --no-channel-add
Note: a multi-user installation is possible. See https://nixos.org/manual/nix/stable/installation/installing-binary.html#multi-user-installation
Switching to the Multi-user Installer
Welcome to the Multi-User Nix Installation

This installation tool will set up your computer with the Nix package
manager. This will happen in a few stages:
[...]
---- Reminders -----------------------------------------------------------------
[ 1 ]
Nix won't work in active shell sessions until you restart them.

Testing Nix installation...
+ source /home/vagrant/.bash_profile
-bash: line 4: /home/vagrant/.bash_profile: No such file or directory
+ true
+ source /home/vagrant/.bash_login
-bash: line 5: /home/vagrant/.bash_login: No such file or directory
+ true
+ source /home/vagrant/.profile
++ '[' -n '5.1.16(1)-release' ']'
++ '[' -f /home/vagrant/.bashrc ']'
++ . /home/vagrant/.bashrc
+++ case $- in
+++ return
++ '[' -d /home/vagrant/bin ']'
++ '[' -d /home/vagrant/.local/bin ']'
+ source /etc/bashrc
++ '[' -e /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh ']'
++ . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh
+++ '[' -n 1 ']'
+++ return
+ nix-env --version
nix-env (Nix) 2.13.0pre19700101_dirty
+ nix --extra-experimental-features nix-command store ping
Store URL: daemon
Version: 2.13.0pre19700101_dirty
++ nix-build --no-substitute -E 'derivation { name = "foo"; system = "x86_64-linux"; builder = "/bin/sh"; args = ["-c" "echo foobar > $out"]; }'
  /nix/store/9qb0l9n1gsmcyynfmndnq3qpmlvq8rln-foo.drv
building '/nix/store/9qb0l9n1gsmcyynfmndnq3qpmlvq8rln-foo.drv'...
+ out=/nix/store/sivbjmsgqj95sxw35iqgvqd64grp5q91-foo
++ cat /nix/store/sivbjmsgqj95sxw35iqgvqd64grp5q91-foo
+ [[ foobar = foobar ]]
Running installer again to test for idempotency...
+ tar -xf ./nix.tar.xz
+ mv ./nix-2.13.0pre19700101_dirty-x86_64-linux nix
+ ./nix/install --daemon --no-channel-add
Note: a multi-user installation is possible. See https://nixos.org/manual/nix/stable/installation/installing-binary.html#multi-user-installation
Switching to the Multi-user Installer
Welcome to the Multi-User Nix Installation

This installation tool will set up your computer with the Nix package
manager. This will happen in a few stages:
[...]
---- Reminders -----------------------------------------------------------------
[ 1 ]
Nix won't work in active shell sessions until you restart them.

Testing Nix installation...
+ source /home/vagrant/.bash_profile
-bash: line 4: /home/vagrant/.bash_profile: No such file or directory
+ true
+ source /home/vagrant/.bash_login
-bash: line 5: /home/vagrant/.bash_login: No such file or directory
+ true
+ source /home/vagrant/.profile
++ '[' -n '5.1.16(1)-release' ']'
++ '[' -f /home/vagrant/.bashrc ']'
++ . /home/vagrant/.bashrc
+++ case $- in
+++ return
++ '[' -d /home/vagrant/bin ']'
++ '[' -d /home/vagrant/.local/bin ']'
+ source /etc/bashrc
++ '[' -e /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh ']'
++ . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh
+++ '[' -n 1 ']'
+++ return
+ nix-env --version
nix-env (Nix) 2.13.0pre19700101_dirty
+ nix --extra-experimental-features nix-command store ping
Store URL: daemon
Version: 2.13.0pre19700101_dirty
++ nix-build --no-substitute -E 'derivation { name = "foo"; system = "x86_64-linux"; builder = "/bin/sh"; args = ["-c" "echo foobar > $out"]; }'
+ out=/nix/store/sivbjmsgqj95sxw35iqgvqd64grp5q91-foo
++ cat /nix/store/sivbjmsgqj95sxw35iqgvqd64grp5q91-foo
+ [[ foobar = foobar ]]
Done!
qemu-kvm: terminating on signal 15 from pid 1 ()
/nix/store/i9ln2c3wqm23vh5i0k63gazik4kbckmv-installer-test-ubuntu-22-04-install-force-daemon
```

So that's looking good.

I saw that a lot of issues on here are related to this, so I'll make a proper list, squash the commits and add references to all of them so they get closed when this is merged.

I'm not sure if making the github CI-pipeline test for this as well makes sense, as long as hydra is doing it.

iFreilicht:
I tested this again, and it does seem like 250c11895bb2e1 from this PR kind-of fixes the `cp: cannot overwrite directory with non-directory` crashes. The issue here is that if the Nix store is genuinely broken (which is likely the original cause of that error, otherwise cp would overwrite the directory fine), a re-install wouldn't fix it.
So instead, I'll add an additional line that `rm`s all the directories that are about to be copied into the store. I'll also deliberately break the nix store in the installer test to ensure this won't break again in the future.

Potentially, we could also fully delete /nix/* first (which might also resolve other issues), but I feel this is way too nuclear and has potential to break and/or delete other parts of the installation like per-user profiles and their generations, which ideally should be left alone during a re-install.

iFreilicht:
YES! I finally figured it out! In bb0c4b9f25b6d064d91aedd71940f14b1b624291, this change was made:

```diff
@@ -741,2 +741,2 @@
         _sudo "to copy the basic Nix files to the new store at $NIX_ROOT/store" 
\
-              cp -RLp ./store/* "$NIX_ROOT/store/"
+              cp -RPp ./store/* "$NIX_ROOT/store/"
```

Which made sure symlinks are copied verbatim and not followed. While technically correct, this created an incompatibility with installs made BEFORE this commit was released.

Before, symlinks to directories would be followed and deep-copied. For example, libkrb contains a symlink `sbin -> bin`. So for installations made with the pre-bb0c4b9 installer, it looked like this:

```shell
$ ls -lF /nix/store/5sxcmklgrgl7lsij8bp9a98iws4q8fw0-libkrb5-1.19.2
total 4
dr-xr-xr-x 1 root root   44 Jan  1  1970 bin/
dr-xr-xr-x 1 root root 1108 Jan  1  1970 lib/
lrwxrwxrwx 1 root root    3 Jan  1  1970 sbin/
dr-xr-xr-x 1 root root   10 Jan  1  1970 share/
```

But when running the post-bb0c4b9 installer, it tries to create this:

```shell
$ ls -lF /nix/store/5sxcmklgrgl7lsij8bp9a98iws4q8fw0-libkrb5-1.19.2
total 4
dr-xr-xr-x 1 root root   44 Jan  1  1970 bin/
dr-xr-xr-x 1 root root 1108 Jan  1  1970 lib/
lrwxrwxrwx 1 root root    3 Jan  1  1970 sbin -> bin/
dr-xr-xr-x 1 root root   10 Jan  1  1970 share/
```

Which is not possible and fails with the dreaded message

```
cp: cannot overwrite directory /nix/store/h9z5lncphgm9if86wxrfqg7w7fv7khbh-libkrb5-1.19.3/sbin with non-directory ./store/h9z5lncphgm9if86wxrfqg7w7fv7khbh-libkrb5-1.19.3/sbin
```

Now I can confidently say that this PR will fix all the `cp: cannot overwrite directory ... with non-directory` errors. I thus updated the list of issues this PR would fix.

I also found two related outdated issues that I believe were caused by 475fc109e72e494039b253131314e3a3ea5723c6, cannot occur since bb0c4b9f25b6d064d91aedd71940f14b1b624291 and will not regress after this PR. They're basically the opposite problem, trying to overwrite a link with a directory:

* #5703
* #5522

I will rebase the PR and add a short version of and link to this explanation to the commit that fixes the issue. After that it is ready to merge from my side.

Ericson2314:

- @thufschmitt: really hard to test on CI due to 
- @tomberek: there have been multiple attempts at various changes and refactorings
  - we should probably treat the installer as a subsystem and deal with it differently
- to discuss a strategy how to address the broader issue, and how that relates to the Determinate installer
  - @thufschmitt: will bring it up in the installer work group

fricklerhandwerk:
@abathur could you take a look?

tests/installer/default.nix
```diff
@@ -43,14 +43,14 @@ let
   images = {
 
     /*
-    "ubuntu-14-04" = {
+      "ubuntu-14-04" = {
```
SuperSandro2000, line 46: The intendation changes here seem odd

scripts/install-multi-user.sh
```diff
@@ -441,19 +441,30 @@ $(uninstall_directions)
...
+            else
+                problematic_profile_targets+=("$profile_target")
+            fi
+        fi
+    done
+
+    if [ "${PROFILE_TARGETS[@]}" ]; then
+        resolution_explanation=$(cat <<EOF
 I back up shell profile/rc scripts before I add Nix to them.
+However, at least one of them was already backed up, so you have to clean up manually.
```
SuperSandro2000, line 460:
```suggestion
However, at least one of the files was already backed up, to continue you need to clean them up manually.
```

fricklerhandwerk:
```suggestion
However, at least one of the files was already backed up.
Clean them up manually to continue.
```

abathur:
> @abathur Ah, thanks for the pointer! I found that page but it didn't seem to be what I was looking for. I got that setup working, which is of course nice, but running the CI just takes forever. What I was really looking for was this:
> 
> ```shell
> $ nix-build -A hydraJobs.installerTests.ubuntu-22-04.x86_64-linux.install-force-daemon
> ```
> 
> This allows me to build and test locally automatically. Very convenient.

Yes, the GA tests aren't great for velocity--but they're still a big step up from the status quo, and I think it's ideal to lock in whatever gains we can get here with both sets of tests.

> But yes, good point, I have to adapt the CI as well to avoid a regression on this in the future. However, the CI is using the [cachix/install-nix](https://github.com/actions-marketplace-validations/cachix_install-nix-action/blob/master/install-nix.sh) action right now, which does a single-user install 

I'm pretty sure it uses multiuser and has for a while as long as you're on macOS or have systemd.

> AND skips the installation if it finds the nix command

Ah. Yes. That's annoying (for our case...). 

I spent some time trying to find a ~cheap way to force this to work and do have a candidate: https://github.com/abathur/nix/commit/100c3d5400f315eec04cfdbeb99a20a51b66932c#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR120-R129

The few lines above the highlight are a failed attempt to just overwrite GITHUB_PATH (which had no effect), but removing nix from the default profile did let the cachix action try to go ahead. (There might be a better way to do this?)

I added a set of jobs that try this with the current release installer, and a set that try it on top of your PR.

Here's the CI run https://github.com/abathur/nix/actions/runs/5329640965

The sanity-check jobs fail where I think we expect them to:
- linux: https://github.com/abathur/nix/actions/runs/5329640965/jobs/9655524481
- macos: https://github.com/abathur/nix/actions/runs/5329640965/jobs/9655524554

The install-test jobs both also fail, so I'm curious what you think about where they tip over (maybe these are out of scope?):
- The linux job fails on linking the systemd unit: https://github.com/abathur/nix/actions/runs/5329640965/jobs/9656408834
- The macOS job fails on a bunch of `cp: cannot overwrite directory ... with non-directory ...` errors: https://github.com/abathur/nix/actions/runs/5329640965/jobs/9656408907

tests/installer/default.nix
iFreilicht: Oh whoops, that was the default autoformatter. I'll revert that.

iFreilicht:
@abathur Thanks for taking the time! Yes, the sanity-checks are exactly what we expect.

But actually, the failing installer-test on macOS is also expected. See the issues I linked in the initial post:

> * Will fix all `cp: cannot overwrite directory ... with non-directory` errors

This is what was fixed by [adding the -n flag to cp](https://github.com/NixOS/nix/pull/7603/files#diff-7d207e64957e3b0c8de11d6ac4464bd4633d326c9706f1961a20a2f5c6c957a4R833).

The unit file on linux I don't know about, I didn't test that. I assume it happens [here, on line 101](https://github.com/NixOS/nix/blob/559fd7ffe7a8cb61b11bec081f14ce3c3ffb210d/scripts/install-systemd-multi-user.sh#L101), but can't really explain why. I also searched the issue tracker, and it seems [not a single person has ever reported this issue before](https://github.com/NixOS/nix/issues?q=is%3Aissue+%22Failed+to+link+unit%3A+No+such+file+or+directory%22+).

tests/installer/default.nix
iFreilicht: Fixed now.

scripts/install-multi-user.sh
iFreilicht: I went with @fricklerhandwerk's suggestion.

abathur:
> But actually, the failing installer-test on macOS is also expected. See the issues I linked in the initial post:
> 
> > * Will fix all `cp: cannot overwrite directory ... with non-directory` errors
> 
> This is what was fixed by [adding the -n flag to cp](https://github.com/NixOS/nix/pull/7603/files#diff-7d207e64957e3b0c8de11d6ac4464bd4633d326c9706f1961a20a2f5c6c957a4R833).

I am not sure I understand why this failure is expected. The latter set of tests includes your code (https://github.com/abathur/nix/blame/100c3d5400f315eec04cfdbeb99a20a51b66932c/scripts/install-multi-user.sh#L833), so it doesnt appear that the new flag stops the error?

iFreilicht:
> I am not sure I understand why this failure is expected. The latter set of tests includes your code (https://github.com/abathur/nix/blame/100c3d5400f315eec04cfdbeb99a20a51b66932c/scripts/install-multi-user.sh#L833), so it doesnt appear that the new flag stops the error?

Ah sorry, I misunderstood. Yeah I guess. I checked the installer the pipeline downloaded from your cachix, and it seems to indeed be the correct version, so I have no idea why it doesn't work in CI but works when running the hydra tests locally.

abathur:
I tested this installer on a spare MacBook with an existing nix 2.13.2 install and it too failed at the cp step with overwrite directory errors.

I'm not very familiar with the vm tests, so I don't have much of a mental model for why they might be differing. Are you sure the macOS variant ran?

tomberek:

PR #7604, opened 2023-01-15
Try to fix build failure

Ericson2314:
Failure: https://hydra.nixos.org/build/205357257/nixlog/1

The problem seems to be trying to `std::visit` a derived class of `std::variant`. Per
https://stackoverflow.com/questions/63616709/incomplete-type-stdvariant-used-in-nested-name-specifier certain C++ standard library implementations allow this, but others do not.

The solution is simply to call the `raw` method, which upcasts the reference back to the `std::variant`.

### DIFF
--- a/src/libstore/misc.cc
+++ b/src/libstore/misc.cc
@@ -319,7 +319,7 @@ OutputPathMap resolveDerivedPath(Store & store, const DerivedPath::Built & bfd, 
         [&](const OutputsSpec::Names & names) {
             return static_cast<std::set<std::string>>(names);
         },
-    }, bfd.outputs);
+    }, bfd.outputs.raw());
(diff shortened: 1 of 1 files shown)


roberth (approved):

roberth:
Perhaps we shouldn't subclass `std::variant` then?

Ericson2314:
We could private subclass It which will force the use of `raw` but still allow `using` for the constructors I think, as another option.

edolstra:
Still broken: https://hydra.nixos.org/build/205946224

Ericson2314:
OK will keep looking.

Ericson2314:
https://github.com/NixOS/nix/pull/7613

PR #7605, opened 2023-01-16
add comments

Radvendii:
@thufschmitt @fricklerhandwerk and I are going through the process by which nix builds things. These are some comments / points of clarification that came up while doing that.

### DIFF
--- a/src/nix-store/nix-store.cc
+++ b/src/nix-store/nix-store.cc
@@ -72,11 +72,13 @@ static PathSet realisePath(StorePathWithOutputs path, bool build = true)
         Derivation drv = store->derivationFromPath(path.path);
         rootNr++;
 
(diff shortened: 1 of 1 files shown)


src/nix-store/nix-store.cc
```diff
@@ -141,6 +143,7 @@ static void opRealise(Strings opFlags, Strings opArgs)
         toDerivedPaths(paths),
         willBuild, willSubstitute, unknown, downloadSize, narSize);
 
+    /* filter out unkown paths from `paths` */
```
edolstra, line 146:
```suggestion
    /* Filter out unknown paths from `paths`. */
```

And similarly capitalize/punctuate the other comments.

Radvendii: Done.

PR #7609, opened 2023-01-17
Hide experimental settings

Ericson2314:
# Motivation

Settings for experimental features should not be accessible without that experimental feature enabled. For example, if `nix-command` is enabled, but `flakes` is disabled, the CLI should refuse to:

 - accept flags which would modify flake-specific settings
 - list flake-specific settings in `help` and `show-config`
 - etc.

This commit allows marking settings experimental, like we do for built-ins already, in a declarative way.
It also allows flags to likewise be marked experimental, which is needed to implement the above but also useful in its own right.

Even if we were to stabilize `flakes` and `nix-command` at once (and I hope not!), it is still nice to have this functionality for future experimental features. For example, we might end up needing settings and flags that exist for `ca-derivations`. We should likewise be able to hide those from the stable CLI.

# Context

In #7608 @roberth started investigating what would be needed to make a very simple command `nix store gc`, Hiding these `flake` settings was one of the items in his check list.

The first commit message as a test that fails, and the latter two fix the test, until it passes. The latter two have somewhat in-depth commit messages explaining the implementation strategy.

Needed for #7608.

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests

### DIFF
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -52,7 +52,7 @@ std::shared_ptr<Completions> completions;
 
 std::string completionMarker = "___COMPLETE___";
 
-std::optional<std::string> needsCompletion(std::string_view s)
+static std::optional<std::string> needsCompletion(std::string_view s)
 {
     if (!completions) return {};
     auto i = s.find(completionMarker);
@@ -120,6 +120,12 @@ void Args::parseCmdline(const Strings & _cmdline)
 
     if (!argsSeen)
         initialFlagsProcessed();
+
+    /* Now that we are done parsing, make sure that any experimental
+     * feature required by the flags is enabled */
+    for (auto & f : flagExperimentalFeatures)
+        experimentalFeatureSettings.require(f);
+
 }
 
 bool Args::processFlag(Strings::iterator & pos, Strings::iterator end)
@@ -128,12 +134,18 @@ bool Args::processFlag(Strings::iterator & pos, Strings::iterator end)
 
     auto process = [&](const std::string & name, const Flag & flag) -> bool {
         ++pos;
+
+        if (auto & f = flag.experimentalFeature)
+            flagExperimentalFeatures.insert(*f);
+
         std::vector<std::string> args;
         bool anyCompleted = false;
         for (size_t n = 0 ; n < flag.handler.arity; ++n) {
             if (pos == end) {
                 if (flag.handler.arity == ArityAny || anyCompleted) break;
-                throw UsageError("flag '%s' requires %d argument(s)", name, flag.handler.arity);
+                throw UsageError(
+                    "flag '%s' requires %d argument(s), but only %d were given",
+                    name, flag.handler.arity, n);
             }
             if (auto prefix = needsCompletion(*pos)) {
                 anyCompleted = true;
@@ -152,7 +164,11 @@ bool Args::processFlag(Strings::iterator & pos, Strings::iterator end)
             for (auto & [name, flag] : longFlags) {
                 if (!hiddenCategories.count(flag->category)
                     && hasPrefix(name, std::string(*prefix, 2)))
+                {
+                    if (auto & f = flag->experimentalFeature)
+                        flagExperimentalFeatures.insert(*f);
                     completions->add("--" + name, flag->description);
+                }
             }
             return false;
         }
@@ -172,7 +188,8 @@ bool Args::processFlag(Strings::iterator & pos, Strings::iterator end)
         if (prefix == "-") {
             completions->add("--");
             for (auto & [flagName, flag] : shortFlags)
-                completions->add(std::string("-") + flagName, flag->description);
+                if (experimentalFeatureSettings.isEnabled(flag->experimentalFeature))
+                    completions->add(std::string("-") + flagName, flag->description);
         }
     }
 
@@ -219,6 +236,8 @@ nlohmann::json Args::toJSON()
     auto flags = nlohmann::json::object();
 
     for (auto & [name, flag] : longFlags) {
+        /* Skip experimental flags when listing flags. */
+        if (!experimentalFeatureSettings.isEnabled(flag->experimentalFeature)) continue;
         auto j = nlohmann::json::object();
         if (flag->aliases.count(name)) continue;
         if (flag->shortName)
--- a/src/libutil/config.cc
+++ b/src/libutil/config.cc
@@ -70,10 +70,17 @@ void AbstractConfig::reapplyUnknownSettings()
         set(s.first, s.second);
 }
 
+// Whether we should process the option. Excludes aliases, which are handled elsewhere, and disabled features.
+static bool applicable(const Config::SettingData & sd)
+{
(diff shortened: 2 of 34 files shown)


src/libutil/config.cc
```diff
@@ -70,10 +70,17 @@ void AbstractConfig::reapplyUnknownSettings()
         set(s.first, s.second);
 }
 
+static bool regular(const Config::SettingData & sd)
```
roberth, line 73:
```suggestion
// Whether we should process the option. Excludes aliases, which are handled elsewhere, and disabled features.
static bool applicable(const Config::SettingData & sd)
```

I think this name is a bit more meaningful. Where are aliases handled?

src/libutil/config.hh
```diff
@@ -353,4 +359,23 @@ struct GlobalConfig : public AbstractConfig
...
+
+    /* null ptr means no feature, which means there is nothing that could be
+     * disabled, and so the function returns true.
+     */
+    bool isEnabled(const ExperimentalFeature *) const;
+
+    void require(const ExperimentalFeature &) const;
+};
+
+// FIXME: don't use a global variable.
```
roberth, line 378:
```suggestion
```

I agree, but I don't think this is an isolated problem.

tests/experimental-features.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Simple case, the configuration effects the running command
+both_ways show-config
+
+# Complicated case, earlier args effect later args
+
+both_ways store gc --help
+
+! nix --experimental-features 'nix-command' show-config --flake-registry 'https://no'
+nix --experimental-features 'nix-command flakes' show-config --flake-registry 'https://no'
```
roberth, line 19: A unit test would be more future proof, but this will do.

src/libutil/config.hh
```diff
@@ -353,4 +359,23 @@ struct GlobalConfig : public AbstractConfig
...
 
+
+struct ExperimentalFeatureSettings : Config {
+
+    Setting<std::set<ExperimentalFeature>> experimentalFeatures{this, {}, "experimental-features",
+        "Experimental Nix features to enable."};
+
+    bool isEnabled(const ExperimentalFeature &) const;
+
+    /* null ptr means no feature, which means there is nothing that could be
```
roberth, line 370:
```suggestion
    /* nullptr means no feature, which means there is nothing that could be
```

or _null pointer_

roberth (commented):
Looks decent; just a suggestion and two nitpicks.
I don't think I know enough about how the initialization order of the settings to judge that aspect, but otherwise lgtm.

src/libutil/config.hh
Ericson2314:
```suggestion
    /* null pointer means no feature, which means there is nothing that could be
```

src/libutil/config.cc
Ericson2314: Not sure, but I didn't that.

src/libutil/config.cc
```diff
@@ -147,34 +154,38 @@ nlohmann::json Config::toJSON()
 {
     auto res = nlohmann::json::object();
     for (auto & s : _settings)
-        if (!s.second.isAlias) {
+        if (regular(s.second))
```
Ericson2314, line 157:
```suggestion
        if (applicable(s.second))
```

src/libutil/config.cc
```diff
@@ -147,34 +154,38 @@ nlohmann::json Config::toJSON()
...
-        }
     return res;
 }
 
 std::string Config::toKeyValue()
 {
     auto res = std::string();
     for (auto & s : _settings)
-        if (!s.second.isAlias) {
+        if (regular(s.second))
```
Ericson2314, line 166:
```suggestion
        if (applicable(s.second))
```

src/libutil/config.hh
Ericson2314: It isn't, but another such global already had this FIXME, so I just copied it for consistency.

tests/experimental-features.sh
Ericson2314: Thanks. Yeah I would have to mock up a whole command thingy and parse it, at which point I'd be worried the mocking and the real world got out of sync.

src/libutil/config.hh
roberth: I must have made a mistake grepping for it. sgtm

src/libutil/args.cc
```diff
@@ -124,10 +124,21 @@ void Args::parseCmdline(const Strings & _cmdline)
 
 bool Args::processFlag(Strings::iterator & pos, Strings::iterator end)
 {
+    auto expFlag = [](const Args::Flag & f) -> bool {
+        auto & xp = f.experimentalFeature;
+        return experimentalFeatureSettings.isEnabled(xp ? &*xp : nullptr);
+    };
+
     assert(pos != end);
```
roberth, line 132: I'm getting
```
$ nix shell ~/h/nix --accept-flake-config
nix: src/libutil/args.cc:132: virtual bool nix::Args::processFlag(std::__cxx11::list<std::__cxx11::basic_string<char> >::iterator&, std::__cxx11::list<std::__cxx11::basic_string<char> >::iterator): Assertion `pos != end' failed.
Aborted (core dumped)
```

tests/experimental-features.sh
```diff
@@ -0,0 +1,19 @@
+source common.sh
+
+# Without flakes, flake options should not show up
+# With flakes, flake options should show up
+
+function both_ways {
+    nix --experimental-features 'nix-command' "$@" | grepQuietInverse flake
+    nix --experimental-features 'nix-command flakes' "$@" | grepQuiet flake
```
roberth, line 8: Flags used to be declarative. This currently breaks:

```suggestion
    nix --experimental-features 'nix-command' "$@" | grepQuietInverse flake
    nix --experimental-features 'nix-command flakes' "$@" | grepQuiet flake

    # Also, the order should not matter
    nix "$@" --experimental-features 'nix-command' | grepQuietInverse flake
    nix "$@" --experimental-features 'nix-command flakes'  | grepQuiet flake
```

src/libutil/config.cc
```diff
@@ -444,4 +462,30 @@ GlobalConfig::Register::Register(Config * config)
...
+
+bool ExperimentalFeatureSettings::isEnabled(const ExperimentalFeature * feature) const
+{
+    return !feature || isEnabled(*feature);
+}
+
+void ExperimentalFeatureSettings::require(const ExperimentalFeature * feature) const
+{
+    if (feature) require(*feature);
+}
```
roberth, line 489: Would be nicer to use `std::optional<>&`, but otherwise the pr lgtm!

fricklerhandwerk:
Discussed in the Nix team meeting 2023-03-13:

 - Annoying because of the churn coming with it
 - But also very useful as long as we have experimental features.
 - Idea approved, assigned to @roberth

PR #7611, opened 2023-01-17
Stuctured command stability

Ericson2314:
Prior to this, there was an ad-hoc whitelist in `main.cc`. Now, every command states its stability.

In a future PR, we will adjust the manual to take advantage of this new
information in the JSON. It will be easier to do that once we have some experimental feature docs to link too; see #5930 and #7798.)

### DIFF
--- a/src/libutil/args.cc
+++ b/src/libutil/args.cc
@@ -236,8 +236,6 @@ nlohmann::json Args::toJSON()
     auto flags = nlohmann::json::object();
 
     for (auto & [name, flag] : longFlags) {
-        /* Skip experimental flags when listing flags. */
-        if (!experimentalFeatureSettings.isEnabled(flag->experimentalFeature)) continue;
         auto j = nlohmann::json::object();
         if (flag->aliases.count(name)) continue;
         if (flag->shortName)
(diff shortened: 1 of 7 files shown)


fricklerhandwerk:
The diff is quite large and it's not clear at a glance what the change actually is about. Can you elaborate in the PR description what exactly changed in terms of user-visible behavior?

Ericson2314:
The diff is large because of the other PRs this depends on.

Ericson2314:
Draft because https://github.com/apps/dpulls is not cutting it, and the chained diffs are confusing.

fricklerhandwerk:
Discussed in the Nix team meeting 2023-04-03:

- @edolstra: not in favor of dynamically generated documentation
    - it should be just markdown files, and the raw string technique is just a hack
    - the flag is good, but should rather use it to mark experimental features rather than hiding it
    - we can also put the information into the prose
    - @fricklerhandwerk: agreed, to gather experience we have to expose the functionality in documentation
- agreement:
    - merge the infrastructure first
    - deal with rendering experimental stuff later
    - `--help` language on installable and flakes should be static and can be an independent PR

Ericson2314:
This is now just the infrastructure, as discussed.

src/libutil/args.cc
```diff
@@ -249,6 +247,11 @@ nlohmann::json Args::toJSON()
             j["arity"] = flag->handler.arity;
         if (!flag->labels.empty())
             j["labels"] = flag->labels;
+        // FIXME With C++23 use std::optional::tranform
```
roberth, line 250:
```suggestion
        // TODO With C++23 use std::optional::tranform
```

Not an actual problem, just something we may do.

roberth (approved):
Just the nit, otherwise lgtm!

PR #7616, opened 2023-01-17
Fix foreign key error inserting into NARs #3898

### LINKED ISSUE #3898: nix-build yields SQLiteError due to failed constraint when inserting into NARs

roberth:
# Motivation

 - Fixes #3898 

# Context

I've conjectured that this might be a SQLite bug, but the infrequent behavior was caused by the timestamp ttl instead.

# Checklist for maintainers

 - [ ] agreed on idea
 - [ ] unit tests
   - relies on functional tests
   - the code is exercised by the binary-caches.sh test
   - n/a


src/libstore/nar-info-disk-cache.cc
```diff
@@ -189,23 +216,42 @@ class NarInfoDiskCacheImpl : public NarInfoDiskCache
...
 
-            state->insertCache.use()(uri)(time(0))(storeDir)(wantMassQuery)(priority).exec();
-            assert(sqlite3_changes(state->db) == 1);
-            state->caches[uri] = Cache{(int) sqlite3_last_insert_rowid(state->db), storeDir, wantMassQuery, priority};
+            Cache ret {
+                // id: to be set in the conditional below
+                .id = -1,
+                // lastCacheInfoFetch: always the current time, because this is
+                // only called when a cached Cache wasn't available, so the parameters
+                // are always fresh.
```
roberth, line 227: I would prefer for this code not to make an assumption, but rather refactor these methods so that `NarInfoDiskCache` only exposes a single method that's an upsert with a lazy default. I've opted for a less invasive fix for now.

src/libstore/sqlite.cc
```diff
@@ -58,6 +62,11 @@ SQLite::SQLite(const Path & path, bool create)
     if (sqlite3_busy_timeout(db, 60 * 60 * 1000) != SQLITE_OK)
         SQLiteError::throw_(db, "setting timeout");
 
+    if (false) {
```
cole-h, line 65: I don't profess to be a C/C++ master, but I've seen other places in the codebase where we use `#if 0` and `#endif` to guard these kinds of things, rather than  an `if (false)`. Is there a reason we don't do the same here?

To my naive understanding, the difference is that `#if 0` would not be parsed / compiled / etc at all, while `if (false)` might? Maybe compilers are smart enough to deal with that, though.

roberth: I did try `#if 0`, but that resulted in a warning about unused `traceSQL`. I figured this was the least obtrusive solution.

I don't like the alternative of adding more machinery to take care of `traceSQL`. If we're not going minimal, I'd rather improve the logger situation so that it's more configurable at runtime.

cole-h: Gotcha, that makes sense. Thanks for trying my silly suggestion!

cransom:
I've stuck this in our CI system and it's been smooth sailing so far. pre-patch, a fresh instance would come up and fail it's first few builds regularly. I haven't seen a failure yet. Thank you!

src/libstore/nar-info-disk-cache.cc
```diff
@@ -166,19 +179,33 @@ class NarInfoDiskCacheImpl : public NarInfoDiskCache
         return i->second;
     }
 
+private:
+
+    bool cacheIsUpToDate(const Cache &cache) {
+        return cache.lastCacheInfoFetch > (time(0) - NarInfoDiskCacheImpl::cacheInfoTtl);
```
thufschmitt, line 185:
```suggestion
        return cache.lastCacheInfoFetch > (time(0) - cacheInfoTtl);
```

(I think, correct me if we're not in the right context)

src/libstore/nar-info-disk-cache.cc
```diff
@@ -189,23 +216,42 @@ class NarInfoDiskCacheImpl : public NarInfoDiskCache
...
+                .priority = priority,
+            };
+            if (cache) {
+                ret.id = cache->id;
+                state->updateCache.use()(uri)(ret.lastCacheInfoFetch)(storeDir)(wantMassQuery)(priority)(ret.id).exec();
+                assert(sqlite3_changes(state->db) == 1);
+            } else {
+                state->insertCache.use()(uri)(ret.lastCacheInfoFetch)(storeDir)(wantMassQuery)(priority).exec();
+                assert(sqlite3_changes(state->db) == 1);
+                ret.id = (int) sqlite3_last_insert_rowid(state->db);
```
thufschmitt, line 240:
```suggestion
                ret.id = (int) state->db.getLastInsertedRowId();
```

(mostly cosmetic, but since we have a more idiomatic way of doing that, we might as well use it)

thufschmitt (changes requested):
A few minor comments, but looks good overall :+1: 

Quick extra question: Does this require bumping the cache db schema version number? I think it doesn't but would just like to make sure that we're not forgetting something and exposing us to subtle bugs because we accidentally change the exact meaning of the db

roberth:
> Quick extra question: Does this require bumping the cache db schema version number? I think it doesn't but would just like to make sure that we're not forgetting something and exposing us to subtle bugs because we accidentally change the exact meaning of the db

This doesn't change the meaning or format of the stored data, so I see no reason to invalidate it.

edolstra:
So what's happening here is that [we were incorrectly using `INSERT OR REPLACE`](https://stackoverflow.com/questions/418898/sqlite-upsert-not-insert-or-replace), since that will *delete* rather than update the previously existing row. As a side-effect, all cached NAR entries were deleted as well (due to 'on delete cascade'). This probably started happening as a result of d533a885465846e7512ff976d3599685c90316eb.

I think the simpler fix is to replace the `INSERT OR REPLACE` with an upsert, i.e.
```
insert into BinaryCaches(url, timestamp, storeDir, wantMassQuery, priority) values (?1, ?2, ?3, ?4, ?5) 
  on conflict(url) do update set timestamp = ?2, storeDir = ?3, wantMassQuery = ?4, priority = ?5;
```

roberth:
> and replace the use of `sqlite3_last_insert_rowid()` with a SELECT since it doesn't work in case of a conflict.

So then we're almost always doing two statements.
It looks like there's a `RETURNING` clause we could use to return the row id. I'll look into it.

src/libstore/sqlite.cc
```diff
@@ -41,6 +41,10 @@ SQLiteError::SQLiteError(const char *path, const char *errMsg, int errNo, int ex
         throw SQLiteError(path, errMsg, err, exterr, offset, std::move(hf));
 }
 
+static void traceSQL (void *x, const char *sql){
+    fprintf(stderr, "SQL<[%s]>\n", sql);
```
edolstra, line 45: Change to `printError()`?

src/libstore/sqlite.cc
```diff
@@ -41,6 +41,10 @@ SQLiteError::SQLiteError(const char *path, const char *errMsg, int errNo, int ex
         throw SQLiteError(path, errMsg, err, exterr, offset, std::move(hf));
 }
 
+static void traceSQL (void *x, const char *sql){
```
edolstra, line 44:
```suggestion
static void traceSQL(void * x, const char * sql)
{
```

edolstra (approved):

PR #7617, opened 2023-01-17
local-store.cc: suggest `nix store repair` on hash mismatch

amjoseph-nixpkgs:
Recently my buildfarm tripped a circuit breaker and I wound up with a bunch of zero-length files in /nix/store.  This is no big deal, of course, and easy to fix.

However I spent quite a lot of time thinking that Nix's hash mismatch error was a message from my local `nix`, informing me that the remote builder had sent back invalid data.  In fact my local `nix` was informing me that it had discovered corrupted paths in `/nix/store`.

Let's add a suggestion to this error message.  This will make it clear that the problem is (at least partly) an issue with the local store, and also put the user on course towards the most likely solution to their problems.

### DIFF
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1306,11 +1306,11 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
             auto hashResult = hashSink.finish();
 
             if (hashResult.first != info.narHash)
-                throw Error("hash mismatch importing path '%s';\n  specified: %s\n  got:       %s",
+                throw Error("hash mismatch importing path '%s' (try `nix store repair`);\n  specified: %s\n  got:       %s",
                     printStorePath(info.path), info.narHash.to_string(Base32, true), hashResult.first.to_string(Base32, true));
 
             if (hashResult.second != info.narSize)
-                throw Error("size mismatch importing path '%s';\n  specified: %s\n  got:       %s",
+                throw Error("size mismatch importing path '%s' (try `nix store repair`);\n  specified: %s\n  got:       %s",
                     printStorePath(info.path), info.narSize, hashResult.second);
 
             if (info.ca) {
--- a/src/libstore/local-store.cc
+++ b/src/libstore/local-store.cc
@@ -1306,11 +1306,11 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
             auto hashResult = hashSink.finish();
 
             if (hashResult.first != info.narHash)
-                throw Error("hash mismatch importing path '%s' (try `nix store repair`);\n  specified: %s\n  got:       %s",
+                throw Error("hash mismatch importing path '%s' (try: nix store repair);\n  specified: %s\n  got:       %s",
                     printStorePath(info.path), info.narHash.to_string(Base32, true), hashResult.first.to_string(Base32, true));
 
             if (hashResult.second != info.narSize)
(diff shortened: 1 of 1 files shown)


amjoseph-nixpkgs:
Note: I did not add the suggestion to the two similar error messages for `ca-derivations` because I'm not sure it is as certain to be the right course of action there.  If it is, let me know, and I can add suggestions to those errors as well:

https://github.com/NixOS/nix/blob/ab9dcf3477541633f14e72323eb66d093b7ded92/src/libstore/local-store.cc#L1324

https://github.com/NixOS/nix/blob/ab9dcf3477541633f14e72323eb66d093b7ded92/src/libstore/local-store.cc#L1333

src/libstore/local-store.cc
```diff
@@ -1306,11 +1306,11 @@ void LocalStore::addToStore(const ValidPathInfo & info, Source & source,
             auto hashResult = hashSink.finish();
 
             if (hashResult.first != info.narHash)
-                throw Error("hash mismatch importing path '%s';\n  specified: %s\n  got:       %s",
+                throw Error("hash mismatch importing path '%s' (try `nix store repair`);\n  specified: %s\n  got:       %s",
```
edolstra, line 1309:
```suggestion
                throw Error("hash mismatch importing path '%s' (try 'nix store repair');\n  specified: %s\n  got:       %s",
```

Error messages are not markdown.

roberth: The other day I ran into an error message that along the lines of

    'something is wrong, did you mean mapAttrs'

That reads a lot like `mapAttrs'`, a valid identifier and existing function. Backticks would resolve that confusion.
I also think that people read unrendered markdown quite often, when browsing markdown-documented code for example. Maybe the backtick is becoming a natural way to quote code.

That's a larger decision that shouldn't be made here though, so I do agree with the suggestion.

thufschmitt:
Is this is a relevant message in the general case?
I assume this mostly happens when importing an invalid path, in which case `nix store repair` isn't the best advice.

I'm curious how you managed to trigger this btw

amjoseph-nixpkgs:
> I'm curious how you managed to trigger this btw

BTRFS is known to leave zero-length files after a power failure because it journals metadata and data separately.

> Is this is a relevant message in the general case?
> I assume this mostly happens when importing an invalid path, in which case `nix store repair` isn't the best advice.

When using remote builders, Nix prints error messages produced *by* the remote builder *on* the build-driver (the machine running `nix-build`), with no extra tag to explain which machine the error came from (this is something that annoys me).

So if you have corrupt store paths on your `nix-build` machine, you'll see these messages when the remote builder complains about the bogus store paths being fed to it over the remote builder protocol.

I think this is what I experienced.  I do know for certain that the corrupted files were on the `nix-build` machine, not the `nix-daemon` machines.  They are (or rather were) all on a single circuit breaker.

thufschmitt:
> If you have corrupt store paths on your nix-build machine, you'll see these messages when the remote builder complains about the bogus store paths being fed to it over the remote builder protocol. I think this is what I experienced

Ah yes, that makes sense, thanks.

Maybe we could check the validity of paths when _sending_ them then :thinking: I expect that it wouldn't be too costly since we're serializing them any way, and that way we could have an error message that points to the correct location

amjoseph-nixpkgs:
> Maybe we could check the validity of paths when _sending_ them then thinking I expect that it wouldn't be too costly since we're serializing them any way, and that way we could have an error message that points to the correct location

I like that idea; I'll submit a PR for that instead.  Might take a few weeks, I am pretty backlogged right now.

PR #7621, opened 2023-01-17
Revert #6204 to fix regression, add nixpkgs/lib/tests as regression test

roberth:
# Motivation

The latest release causes an undue infinite recursion error and another undue error in a select expression (but possibly not caused by ExprSelect)

This reverts the PR that causes the problem and adds the nixpkgs/lib test suite to help make sure we don't break the evaluator or nixpkgs/lib as easily.

<details><summary>Troubleshooting, bisect, etc</summary>

TODO
 - [x] ~fix~ "fix" the issue
    - bisect details below
    - Appears to be introduced by https://github.com/NixOS/nix/pull/6204
      - hypothesis: Nix appears to have become stricter?
        - no extra `force*` calls in the diff
    - Reverted everything

Nixpkgs diff

```diff
diff --git a/lib/tests/release.nix b/lib/tests/release.nix
index f67892ab962..568e82b3b02 100644
--- a/lib/tests/release.nix
+++ b/lib/tests/release.nix
@@ -1,11 +1,11 @@
 { # The pkgs used for dependencies for the testing itself
   # Don't test properties of pkgs.lib, but rather the lib in the parent directory
-  pkgs ? import ../.. {} // { lib = throw "pkgs.lib accessed, but the lib tests should use nixpkgs' lib path directly!"; }
+  pkgs ? import ../.. {} // { lib = throw "pkgs.lib accessed, but the lib tests should use nixpkgs' lib path directly!"; },
+  nix ? pkgs.nix,
 }:
 
 pkgs.runCommand "nixpkgs-lib-tests" {
   buildInputs = [
-    pkgs.nix
     (import ./check-eval.nix)
     (import ./maintainers.nix {
       inherit pkgs;
@@ -19,8 +19,11 @@ pkgs.runCommand "nixpkgs-lib-tests" {
       inherit pkgs;
     })
   ];
+  nativeBuildInputs = [
+    nix
+  ];
 } ''
-    datadir="${pkgs.nix}/share"
+    datadir="${nix}/share"
     export TEST_ROOT=$(pwd)/test-tmp
     export NIX_BUILD_HOOK=
     export NIX_CONF_DIR=$TEST_ROOT/etc
```

```
$ git bisect good 2.12.0
$ git bisect bad upstream/master
$ cat >../bisecter
#!/usr/bin/env bash
nix-build ../nixpkgs/lib/tests/release.nix --arg nix '(builtins.getFlake "git+file://${toString ./.}").packages.x86_64-linux.default'
$ chmod a+x ../bisecter
$ git bisect run ../bisecter
```

```
git bisect start
# status: waiting for both good and bad commits
# good: [ef800f1e73602c0f10951dd789b97e750f37afc0] Mark official release
git bisect good ef800f1e73602c0f10951dd789b97e750f37afc0
# status: waiting for bad commit, 1 good commit known
# bad: [98f57f44bbeca3b555bd732771eac4c07d54576b] Merge pull request #7620 from NixOS/bump-2.14.0
git bisect bad 98f57f44bbeca3b555bd732771eac4c07d54576b
# good: [eece14dce42bcadb3137b27197f4807f0d696c01] Merge pull request #7410 from edolstra/release-notes
git bisect good eece14dce42bcadb3137b27197f4807f0d696c01
# good: [fb8fc6fda61b44fec161958531ab69934c5f2f4d] Merge pull request #7478 from hercules-ci/make-sure-initNix-called
git bisect good fb8fc6fda61b44fec161958531ab69934c5f2f4d
# bad: [1a4a02cff9f5d474e7085c80a0c6ef58cf50a335] Merge pull request #7559 from ncfavier/no-check-modules
git bisect bad 1a4a02cff9f5d474e7085c80a0c6ef58cf50a335
# bad: [d6f5734c630f893e8695bd37ca2b0dbf2bb24daa] Complete genericClosure tests
git bisect bad d6f5734c630f893e8695bd37ca2b0dbf2bb24daa
# bad: [5ef88457b8bdef957fcbb3cd0e7740595fad1949] Better document error location indent
git bisect bad 5ef88457b8bdef957fcbb3cd0e7740595fad1949
# bad: [e6d07e0d89d964cde22894fca57a95177c085c8d] Refactor to use more traces and less string manipulations
git bisect bad e6d07e0d89d964cde22894fca57a95177c085c8d
# bad: [57684d62475c9ec37e2abc2b7df2ec3581d0f011] fixup! s/forceValue/forceFunction/ where applicable
git bisect bad 57684d62475c9ec37e2abc2b7df2ec3581d0f011
# bad: [be1f0697468bd6c0f2be4f7e058270c161098e9f] Add error context for most basic coercions
git bisect bad be1f0697468bd6c0f2be4f7e058270c161098e9f
# bad: [00e242feed5ac848f5948dd2731bfabe603999ce] Add some context to coercion error strings
git bisect bad 00e242feed5ac848f5948dd2731bfabe603999ce
# first bad commit: [00e242feed5ac848f5948dd2731bfabe603999ce] Add some context to coercion error strings
```

</details>

# Context

 - This test breaks when upgrading to the 2.13 release. https://github.com/NixOS/nixpkgs/pull/211290#event-8250211956
   Log: https://gist.github.com/GrahamcOfBorg/fc8358d3707cca3d7aa55f38d6b9da4d

# Checklist for maintainers

 - [ ] ~agreed on idea~
    - hotfix. easy enough to revert the revert and fix properly. Ideally we shouldn't have to agree and release is automated so I can complete the hotfix myself.
    - `nixpkgsFor` includes our nix by overlay 
 - [x] ~unit~ tests
    - regression test added
    - tbd when root cause is found
    - n/a
   - could be improved slightly, but I'm going to favor speed and definitely not merge without CI to regain that speed, because I do make mistakes.
 - [x] new feature or bug fix: updated release notes


thufschmitt:
Probably worth pinging @layus since he's the author of #6204 

On top of your head, do you have any idea what might have gone wrong?

thufschmitt:
Other than that, Do we really want to revert this PR on master? It's obviously important on the release branch, but keeping it for a couple days on master and try to directly fix the bug might have been simpler given the size of the change (and the likelyness of merge conflicts once we want to re-merge it). Maybe we should un-revert it if there's a reasonable guaranty that the bug can be fixed quickly?

edolstra:
I'm not happy about reverting this on master, instead of actually fixing the bug. This applies the code churn of the original PR all over again. For instance, I spent a non-trivial amount of time last week syncing master into lazy-trees after the original PR got merged. The revert probably will cause a similar amount of merge conflicts to be resolved, and presumably again once we re-apply the PR.

flake.nix
```diff
@@ -532,6 +532,12 @@
               mkdir $out
             '';
 
+        tests.nixpkgsLibTests =
+          nixpkgs.lib.genAttrs systems (system:
+            import (nixpkgs + "/lib/tests/release.nix")
+              { pkgs = nixpkgsFor.${system}; }
+          );
+
```
edolstra, line 540: I don't think this test works, because it's evaluated by the Nix of the CI, not the Nix we're building. 

Maybe this should just extend `tests.evalNixpkgs`?

roberth:
> Do we really want to revert this PR on master?

That is a good question to ask. I think it's important to keep master in a regression free state.

> but keeping it for a couple days on master

I'd prefer for this to happen on a feature branch. In the end it's the commits that matter, not really the name of the branch.

> the likelyness of merge conflicts once we want to re-merge it

Reverting the PR only took only two or three conflicts. I don't expect many conflicts reverting the revert.

> I'm not happy about reverting this on master, instead of actually fixing the bug.

Neither am I. I've spent about an hour trying to find the problem after bisecting.

> I spent a non-trivial amount of time last week syncing master into lazy-trees after the original PR got merged.

It should be in a mergeable state as soon as the revert is reverted, shouldn't it?

flake.nix
roberth: > > * `nixpkgsFor` includes our nix by overlay

Just confirmed that it catches the regression. I've also merged a PR upstream so that we can pass it directly, which is more explicit.

> Maybe this should just extend `tests.evalNixpkgs`?

Reusing the whole upstream expression helps with making sure that the release can make its way into Nixpkgs.

roberth: > the Nix of the CI

The real testing happens in the sandbox. The Nix of the CI only builds the new Nix and provides the environment, but the new Nix is used for the actual testing.

lf-:
@tomberek while it is true that it is a regression in the error message, it seems to be a regression that causes the error context to not be displayed, which is definitely a notable regression since it would break custom errors in nixpkgs modules, regressing UX.

tomberek:
> @tomberek while it is true that it is a regression in the error message, it seems to be a regression that causes the error context to not be displayed, which is definitely a notable regression since it would break custom errors in nixpkgs modules, regressing UX.

@lf- Please confirm. Is the resulting error worse? The relevant PR was specifically improvements to error messages.

layus:
@roberth Thanks for catching this, and sorry for the fuss.
The nixos modules traces not being printed is definitely an issue, but thankfully a very easy one to fix, see https://github.com/NixOS/nix/pull/6204#issuecomment-1396617754.

I could not reproduce the infinite execution / strictness issue you mentioned. Does it happen in the same tests ?

lf-:
> 

I am just going off the output posted in the ofborg failure, which should say which module was being evaluated due to the added error context. I've not done any tests myself.

tomberek:
The difference in output re-ordering and the following line with `--show-trace`. This is not a difference in evaluation output.

```
 … while evaluating the module argument `custom' in "/home/tom/nixpkgs/lib/tests/modules/import-custom-arg.nix":
```

I'm pretty sure the original 

## 2.12.0
```
error: infinite recursion encountered

       at /nix/store/1y58r47hvvn2dbwm4vrsj95krdnmfdl5-source/lib/modules.nix:483:28:

          482|         builtins.addErrorContext (context name)
          483|           (args.${name} or config._module.args.${name})
          484|       ) (lib.functionArgs f);

       … while evaluating the module argument `custom' in "/home/tom/nixpkgs/lib/tests/modules/import-custom-arg.nix":

       … while calling anonymous lambda

       at /nix/store/1y58r47hvvn2dbwm4vrsj95krdnmfdl5-source/lib/modules.nix:481:44:

          480|       context = name: ''while evaluating the module argument `${name}' in "${key}":'';
          481|       extraArgs = builtins.mapAttrs (name: _:
             |                                            ^
          482|         builtins.addErrorContext (context name)

       … from call site

       … while calling 'optional'

       at /nix/store/1y58r47hvvn2dbwm4vrsj95krdnmfdl5-source/lib/lists.nix:255:20:

          254|   */
          255|   optional = cond: elem: if cond then [elem] else [];
             |                    ^
          256|

       … from call site
```

and
## 2.13.0
```
         at /nix/store/1y58r47hvvn2dbwm4vrsj95krdnmfdl5-source/lib/lists.nix:255:20:

          254|   */
          255|   optional = cond: elem: if cond then [elem] else [];
             |                    ^
          256|

       … while calling anonymous lambda

         at /nix/store/1y58r47hvvn2dbwm4vrsj95krdnmfdl5-source/lib/modules.nix:481:44:

          480|       context = name: ''while evaluating the module argument `${name}' in "${key}":'';
          481|       extraArgs = builtins.mapAttrs (name: _:
             |                                            ^
          482|         builtins.addErrorContext (context name)

       error: infinite recursion encountered

       at /nix/store/1y58r47hvvn2dbwm4vrsj95krdnmfdl5-source/lib/modules.nix:483:28:

          482|         builtins.addErrorContext (context name)
          483|           (args.${name} or config._module.args.${name})
          484|       ) (lib.functionArgs f);
```

layus:
https://github.com/NixOS/nix/pull/7641 is my best fix right now. And passes the old and the new tests.

PR #7629, opened 2023-01-18
doc: update language/index.md

phip1611:
- make `<nixpkgs>` visible (was blank in the rendered version)
- add more details

### DIFF
--- a/doc/manual/src/language/index.md
+++ b/doc/manual/src/language/index.md
@@ -191,12 +191,12 @@ This is an incomplete overview of language features, by example.
  <tr>
   <td>
 
-   <nixpkgs>
+   `<nixpkgs>`
 
   </td>
   <td>
 
-   Search path. Value determined by [`$NIX_PATH` environment variable](../command-ref/env-common.md#env-NIX_PATH).
+   Search path for Nix files. Value determined by [`$NIX_PATH` environment variable](../command-ref/env-common.md#env-NIX_PATH).
 
   </td>
  </tr>


doc/manual/src/language/index.md
```diff
@@ -191,12 +191,12 @@ This is an incomplete overview of language features, by example.
  <tr>
   <td>
 
-   <nixpkgs>
+   `<path>`, `<nixpkgs>`
```
edolstra, line 194:
```suggestion
   `<path>`
```

fricklerhandwerk: Either that or the concrete thing many people will still encounter in the wild:

```suggestion
   `<nixpkgs>`
```

doc/manual/src/language/index.md
```diff
@@ -191,12 +191,12 @@ This is an incomplete overview of language features, by example.
...
 
   </td>
   <td>
 
-   Search path. Value determined by [`$NIX_PATH` environment variable](../command-ref/env-common.md#env-NIX_PATH).
+   Search path for Nix expressions; e.g., `"${(import <nixpkgs> {}).nodejs}/bin/node"`. Value determined by [`$NIX_PATH` environment variable](../command-ref/env-common.md#env-NIX_PATH).
```
fricklerhandwerk, line 199:
```suggestion
   Search path for Nix expressions. Value determined by [`$NIX_PATH` environment variable](../command-ref/env-common.md#env-NIX_PATH).
```

The relevant documentation should instead be made more accessible, such as in #7518

fricklerhandwerk:
@phip1611 thanks a lot, good catch!

doc/manual/src/language/index.md
phip1611: so, what now? I'm confused by your two suggestions.

doc/manual/src/language/index.md
phip1611: I agree that documentation should live where it makes the most sense, however, it is helpful especially at this place IMHO. 

From my personal view, it would have helped me a lot when I started learning Nix.

> The relevant documentation should instead be made more accessible

Can you point me to a specific file so that I can include it in the same PR, please?

fricklerhandwerk: > The relevant documentation should instead be made more accessible

Ah, the linked PR already does that work, by splicing in documentation for environment variables and option flags into each `nix-env` and `nix-store` subcommand.

> Can you point me to a specific file so that I can include it in the same PR but a better location, please?

Your other PR #7630 addresses this by making a more elaborate example. I don't know how much time you want to invest, but that part on `NIX_PATH` and the [section on search paths](https://github.com/NixOS/nix/blob/ee1372d2467bba3f40fc849c4a5a9e1d782ee502/doc/manual/src/language/values.md?plain=1#L110-L113) definitely needs more examples to illustrate how all that actually works, and link to each other.

The overview in this table should be brief and just link to one of those.

phip1611:
I updated the PR.

fricklerhandwerk (approved):

PR #7514, opened 2022-12-26
remove incorrect reference to `NIX_PATH` documentation

fricklerhandwerk:
the semantics are not explained in the referenced section any more.
they have been moved to the documentation for common options in the new CLI in https://github.com/NixOS/nix/pull/7421.

unfortunately I was under the [incorrect impression](https://github.com/NixOS/nix/pull/7421#pullrequestreview-1209302202) that there was only one source for the documentation of common options, so in fact the change made discoverability of that piece of documentation strictly worse.

while I know that de facto the classic command line is not maintained any more, I still suggest to fix that properly. who knows how long it will be around...

This work is sponsored by [Antithesis](https://antithesis.com/) ✨

PR #7630, opened 2023-01-18
doc: update command-ref/env-common.md

phip1611:
Add more details to make clearer how it works.

### DIFF
--- a/doc/manual/src/command-ref/env-common.md
+++ b/doc/manual/src/command-ref/env-common.md
@@ -10,8 +10,11 @@ Most Nix commands interpret the following environment variables:
     A colon-separated list of directories used to look up the location of Nix
     expressions using [paths](../language/values.md#type-path)
     enclosed in angle brackets (i.e., `<path>`),
-    e.g. `/home/eelco/Dev:/etc/nixos`. It can be extended using the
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
While there was a [recent change](https://github.com/NixOS/nix/pull/7421) that improved the wording and examples, but moved the information in this section to documentation pages for each `nix` command, e.g. [`nix build`](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-build.html). It left a dangling verbal reference (fixed by https://github.com/NixOS/nix/pull/7514).

I think we should still fix that in a better way, and instead move back the description of how search paths work back to `NIX_PATH` documentation, such that it can be linked to from both old-style and new-style commands without duplicating the source. Then we can continue working on examples.

phip1611:
I moved an additional example from https://github.com/NixOS/nix/pull/7629 to this PR, as it fits better into the new location. What do you think?

fricklerhandwerk:
Yes, I'll close this one in favor of a more elaborate solution as described above. Feel free to propose one if you find time, this will surely make it happen faster than waiting for me or anyone else. Thanks a lot for your time, and for giving things a thought.

PR #7631, opened 2023-01-18
OutputSpec: Allow all valid output names

### LINKED ISSUE #7624: Error: `Invalid outputs specifier` since 2.13.0
**Describe the bug**

Since version 2.13.0 underscores seem to not be allowed anymore in output specifiers. This breaks compatibility between nix versions.
[This commit](https://github.com/NixOS/nix/commit/ce2f91d356438297fd795bd3edb8f9f4536db7da) introduced the error message which I'm seeing [here](https://github.com/holochain/holonix/actions/runs/3944584583/jobs/6750650168#step:4:543)

edolstra:
Fixes #7624.

### DIFF
--- a/src/libstore/outputs-spec.cc
+++ b/src/libstore/outputs-spec.cc
@@ -21,7 +21,8 @@ bool OutputsSpec::contains(const std::string & outputName) const
 
 std::optional<OutputsSpec> OutputsSpec::parseOpt(std::string_view s)
 {
-    static std::regex regex(R"((\*)|([a-z]+(,[a-z]+)*))");
+    // See checkName() for valid output name characters.
+    static std::regex regex(R"((\*)|([a-zA-Z\+\-\._\?=]+(,[a-zA-Z\+\-\._\?=]+)*))");
 
     std::smatch match;
     std::string s2 { s }; // until some improves std::regex
@@ -42,7 +43,7 @@ OutputsSpec OutputsSpec::parse(std::string_view s)
 {
     std::optional spec = parseOpt(s);
     if (!spec)
-        throw Error("Invalid outputs specifier: '%s'", s);
+        throw Error("invalid outputs specifier '%s'", s);
     return *spec;
 }
 
@@ -65,7 +66,7 @@ std::pair<std::string_view, ExtendedOutputsSpec> ExtendedOutputsSpec::parse(std:
 {
     std::optional spec = parseOpt(s);
     if (!spec)
-        throw Error("Invalid extended outputs specifier: '%s'", s);
+        throw Error("invalid extended outputs specifier '%s'", s);
     return *spec;
 }
 
--- a/tests/build.sh
+++ b/tests/build.sh
@@ -42,20 +42,21 @@ nix build -f multiple-outputs.nix --json 'a^*' --no-link | jq --exit-status '
 nix build -f multiple-outputs.nix --json e --no-link | jq --exit-status '
   (.[0] |
     (.drvPath | match(".*multiple-outputs-e.drv")) and
-    (.outputs | keys == ["a", "b"]))
+    (.outputs | keys == ["a_a", "b"]))
 '
 
 # But not when it's overriden.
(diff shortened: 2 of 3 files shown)


tests/build.sh
```diff
@@ -88,7 +89,7 @@ nix build "$drv^first,second" --no-link --json | jq --exit-status '
     (.outputs |
       (keys | length == 2) and
       (.first | match(".*multiple-outputs-a-first")) and
-	  (.second | match(".*multiple-outputs-a-second"))))
+          (.second | match(".*multiple-outputs-a-second"))))
```
Ericson2314, line 92: What is going on here?

tests/build.sh
```diff
@@ -97,14 +98,14 @@ nix build "$drv^*" --no-link --json | jq --exit-status '
...
```
Ericson2314, line 101: idem

tests/build.sh
```diff
@@ -42,20 +42,21 @@ nix build -f multiple-outputs.nix --json 'a^*' --no-link | jq --exit-status '
 nix build -f multiple-outputs.nix --json e --no-link | jq --exit-status '
   (.[0] |
     (.drvPath | match(".*multiple-outputs-e.drv")) and
```
Ericson2314, line 44: I would update the unit tests too.

src/libstore/outputs-spec.cc
```diff
@@ -21,7 +21,8 @@ bool OutputsSpec::contains(const std::string & outputName) const
 
 std::optional<OutputsSpec> OutputsSpec::parseOpt(std::string_view s)
 {
-    static std::regex regex(R"((\*)|([a-z]+(,[a-z]+)*))");
+    // See checkName() for valid output name characters.
+    static std::regex regex(R"((\*)|([a-zA-Z\+\-\._\?=]+(,[a-zA-Z\+\-\._\?=]+)*))");
```
Ericson2314, line 25: Can be in a future PR, but it would be great for these two to share code so they are automatically in sync. Maybe we could `constexpr` build the regex from one to just match a single output name used in `checkName`, for example.

tests/build.sh
edolstra: Stray tabs converted to spaces.

Ericson2314: Ah I see. My bad with that, but those should be 4 spaces not 8. (Vim...)

src/libstore/outputs-spec.cc
roberth: I've checked the regex and it looks accurate.

tests/build.sh
Ericson2314:
```diff
diff --git a/src/libstore/tests/outputs-spec.cc b/src/libstore/tests/outputs-spec.cc
index c9c2cafd0..836ba7e82 100644
--- a/src/libstore/tests/outputs-spec.cc
+++ b/src/libstore/tests/outputs-spec.cc
@@ -40,6 +40,13 @@ TEST(OutputsSpec, names_out) {
     ASSERT_EQ(expected.to_string(), str);
 }

+TEST(OutputsSpec, names_underscore) {
+    std::string_view str = "a_b";
+    OutputsSpec expected = OutputsSpec::Names { "a_b" };
+    ASSERT_EQ(OutputsSpec::parse(str), expected);
+    ASSERT_EQ(expected.to_string(), str);
+}
+
 TEST(OutputsSpec, names_out_bin) {
     OutputsSpec expected = OutputsSpec::Names { "out", "bin" };
     ASSERT_EQ(OutputsSpec::parse("out,bin"), expected);
```
should do it

roberth (commented):
Good incremental improvement, and CI has passed.

tests/build.sh
edolstra: Thanks, added!

tests/build.sh
edolstra: Fixed!

Ericson2314 (approved):

Issue #7633, opened 2023-01-18
Property-based testing


Ericson2314:
To get the most value out of our unit tests, I think we need to be able to write some property based tests. #7624 is something that slipped through the cracks. @edolstra's fix pointed out the relationship with `checkName` --- an excellent point and one that I hadn't thought of when working on `OutputSpec` (and causing the bug). It would be great to have a property test explicitly relating the two.

I did some quick searching and https://github.com/emil-e/rapidcheck/blob/master/doc/gtest.md looks like something that works with googletest. I guess we can try that?

CC @roberth

roberth:
Lowering the barrier to write good tests is always a good thing. That includes property tests.

Ericson2314:
#7639 adds property a test.

Ericson2314:
Since ecd3e4ebd777dca694d0c60431641f054ec2d1ce we have a few, so I'll close this as complete!

PR #7639, opened 2023-01-19
Test store paths, with property tests, fix bug

Ericson2314:
# Motivation

We want mores tests in general. We also have separate code parsing store paths names (when parsing `StorePath` and when parsing `OutputsSpec`) that want to keep in sync. 

The property test in fact found a bug: we were excluding numbers!

The commits are broken up so that reviewing commit by commit should make for easy reading.

# Context

#7633 is an issue to start making property tests for Nix (a la QuickCheck). This kicks off that process, adding `rapidcheck` as a dependency, and using it for these new tests.

# Checklist for maintainers

 - [x] agreed on idea
 - [x] tests, as appropriate
   - unit tests - `src/*/tests`
   - integration tests
    - link to rapidcheck
 - [x] code and comments are self-explanatory
   - we don't have a habit of mentioning fixes or documenting patch releases properly yet; postponed

### DIFF
--- a/doc/manual/src/contributing/hacking.md
+++ b/doc/manual/src/contributing/hacking.md
@@ -92,7 +92,8 @@ $ nix develop
 
 The unit-tests for each Nix library (`libexpr`, `libstore`, etc..) are defined
 under `src/{library_name}/tests` using the
-[googletest](https://google.github.io/googletest/) framework.
+[googletest](https://google.github.io/googletest/) and
+[rapidcheck](https://github.com/emil-e/rapidcheck) frameworks.
 
 You can run the whole testsuite with `make check`, or the tests for a specific component with `make libfoo-tests_RUN`. Finer-grained filtering is also possible using the [--gtest_filter](https://google.github.io/googletest/advanced.html#running-a-subset-of-the-tests) command-line option.
 
--- a/src/libstore/derivations.hh
+++ b/src/libstore/derivations.hh
@@ -13,6 +13,7 @@
 
 namespace nix {
 
+class Store;
 
 /* Abstract syntax of derivations. */
 
--- a/src/libstore/path.hh
+++ b/src/libstore/path.hh
@@ -16,6 +16,8 @@ public:
     /* Size of the hash part of store paths, in base-32 characters. */
     constexpr static size_t HashLen = 32; // i.e. 160 bits
 
+    constexpr static size_t MaxPathLen = 211;
+
     StorePath() = delete;
 
     StorePath(std::string_view baseName);
--- a/src/libstore/path.hh
+++ b/src/libstore/path.hh
@@ -5,7 +5,6 @@
 
 namespace nix {
 
-class Store;
 struct Hash;
 
 class StorePath
--- a/src/libstore/tests/path.cc
+++ b/src/libstore/tests/path.cc
@@ -1,0 +1,144 @@
+#include <regex>
+
+#include <nlohmann/json.hpp>
+#include <gtest/gtest.h>
+#include <rapidcheck/gtest.h>
+
+#include "path-regex.hh"
+#include "store-api.hh"
+
+#include "libstoretests.hh"
+
+namespace nix {
+
+#define STORE_DIR "/nix/store/"
+#define HASH_PART "g1w7hy3qg1w7hy3qg1w7hy3qg1w7hy3q"
+
+class StorePathTest : public LibStoreTest
+{
+};
+
+static std::regex nameRegex { std::string { nameRegexStr } };
+
+#define TEST_DONT_PARSE(NAME, STR)                           \
+    TEST_F(StorePathTest, bad_ ## NAME) {                    \
+        std::string_view str =                               \
+            STORE_DIR HASH_PART "-" STR;                     \
+        ASSERT_THROW(                                        \
+            store->parseStorePath(str),                      \
+            BadStorePath);                                   \
+        std::string name { STR };                            \
+        EXPECT_FALSE(std::regex_match(name, nameRegex));     \
+    }
+
+TEST_DONT_PARSE(empty, "")
+TEST_DONT_PARSE(garbage, "&*()")
+TEST_DONT_PARSE(double_star, "**")
+TEST_DONT_PARSE(star_first, "*,foo")
+TEST_DONT_PARSE(star_second, "foo,*")
+TEST_DONT_PARSE(bang, "foo!o")
+
+#undef TEST_DONT_PARSE
+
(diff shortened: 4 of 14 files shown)


src/libstore/path.hh
```diff
@@ -5,7 +5,6 @@
 
 namespace nix {
 
-class Store;
```
Ericson2314, line 8: This header doesn't refer to `Store`

src/libstore/derivations.hh
```diff
@@ -13,6 +13,7 @@
 
 namespace nix {
 
+class Store;
```
Ericson2314, line 16: idem

doc/manual/src/contributing/hacking.md
```diff
@@ -92,7 +92,8 @@ $ nix develop
 
 The unit-tests for each Nix library (`libexpr`, `libstore`, etc..) are defined
 under `src/{library_name}/tests` using the
-[googletest](https://google.github.io/googletest/) framework.
+[googletest](https://google.github.io/googletest/) and
+[rapidtest](https://github.com/emil-e/rapidcheck) frankworks.
```
cole-h, line 96:
```suggestion
[rapidtest](https://github.com/emil-e/rapidcheck) frameworks.
```

Unless `frankworks` is a new term I haven't come across yet...? :laughing:

src/libutil/regex-combinators.hh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+static inline std::string group(std::string_view a)
+{
+    return std::string { "(" } + a + ")";
+}
+
+static inline std::string many(std::string_view a)
+{
+    return std::string { a } + "*";
```
roberth, line 22: This isn't algebraic, as it only works for groups. (Or for the last subexpression, but that can't be right)
Can we add a non-capturing group around `a`?
Same for `either` as well.

src/libutil/regex-combinators.hh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+static inline std::string list(std::string_view a)
+{
+    return std::string { a } + many(group("," + a));
```
roberth, line 27: This seems to be a domain-specific thing instead of a regex thing. Separate file?

src/libstore/tests/path.cc
```diff
@@ -0,0 +1,144 @@
...
+
+}
+
+namespace rc {
+using namespace nix;
+
+template<>
+struct Arbitrary<StorePath> {
+    static Gen<StorePath> arbitrary();
+};
```
roberth, line 79: I think we'll we want to reuse this in other tests. It's annoying to have that churn in the "next" pr. Do we have a testsupport directory yet?

Ericson2314: Right now all the tests are independent executables.

src/libutil/regex-combinators.hh
Ericson2314: Ah nice, I forgot about those.

Ericson2314:
```suggestion
    return std::string { "(?:" } + a + ")*";
```

src/libutil/regex-combinators.hh
Ericson2314: I don't mind that, but I was a little scared I was making too many headers already :).

doc/manual/src/contributing/hacking.md
```diff
@@ -92,7 +92,8 @@ $ nix develop
 
 The unit-tests for each Nix library (`libexpr`, `libstore`, etc..) are defined
 under `src/{library_name}/tests` using the
-[googletest](https://google.github.io/googletest/) framework.
+[googletest](https://google.github.io/googletest/) and
+[rapidtest](https://github.com/emil-e/rapidcheck) frameworks.
```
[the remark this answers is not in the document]
roberth: Was going to commit this and approve, but it seems that maintainers don't have access to your branch.

roberth (approved):

yorickvP:
This PR breaks buliding in `nix develop` with at least clangStdenv, and possibly everywhere.

`./configure $configureFlags --prefix=$(pwd)/outputs/out` now expands to `./configure CXXFLAGS=-I/nix/store/l6z73w3nqkjk5rq6913sf9vigx1vyv05-rapidcheck-unstable-2020-12-19/extras/gtest/include --with-boost=/nix/store/lf8ghiy89rfhqqyh9j98270afwjqzyrb-boost-1.79.0/lib --with-sandbox-shell=/nix/store/7b943a2k4amjmam6dnwnxnj8qbba9lbq-busybox-static-x86_64-unknown-linux-musl-1.35.0/bin/busybox LDFLAGS=-fuse-ld=gold --prefix=/home/yorick/tweag/nix/outputs/out`

Which interprets everything as CXXFLAGS. The configure script then errors out about boost, but looking in config.status has a whole bunch of `clang-11: error: no such file or directory: 'LDFLAGS=-fuse-ld=gold' `

Reordering to put the new CXXFLAGS last makes it work. This is the documented behavior for ./configure

```
Usage: ./configure [OPTION]... [VAR=VALUE]...
```

Ericson2314:
@yorickvP I am seeing
```
  $ ./configure ... CXXFLAGS=-I/nix/store/l6
z73w3nqkjk5rq6913sf9vigx1vyv05-rapidcheck-unstable-2020-12-19/extras/gtest/include --with-
boost=/nix/store/lf8ghiy89rfhqqyh9j98270afwjqzyrb-boost-1.79.0/lib --with-sandbox-shell=/n
ix/store/7b943a2k4amjmam6dnwnxnj8qbba9lbq-busybox-static-x86_64-unknown-linux-musl-1.35.0/
bin/busybox LDFLAGS=-fuse-ld=gold

ac_cv_env_CXXFLAGS_set=set
ac_cv_env_CXXFLAGS_value=-I/nix/store/l6z73w3nqkjk5rq6913sf9vigx1vyv05-rapidcheck-unstable-2020-12-19/extras/gtest/include
```
in a local `config.log`, which is to say `CXXFLAGS=...` is not the last configure argument and yet not too much got gobbled up, so I am not sure what is going on.

yorickvP:
@Ericson2314 you're right.
It seems that fish was passing `$configureFlags` as a single argument to ./configure, which worked before but not anymore.
I've changed my configure command to `./configure (string split ' ' $configureFlags) --prefix=$(pwd)/outputs/out` to work around it.
Not sure how common this case will be (direnv+fish).

Ericson2314:
@yorickvP Ah yes. I use fish too but not in my Nix development shell so I never noticed that. Glad it's working for you now!

Issue #7553, opened 2023-01-04
Show `addErrorContext` traces by default


roberth:
**Is your feature request related to a problem? Please describe.**

Module system stack traces are too long. The low information density makes users believe that they won't learn anything from reading through it, but essential information is in the stack trace, thanks to builtins.addErrorContext.

**Describe the solution you'd like**

`addErrorContext` traces are sufficiently valuable and terse that the should always be shown.

Optionally, add `--no-trace`, to disable that, but I don't think we even want that.

**Describe alternatives you've considered**

See the following and its alternatives

 - #7552

**Additional context**

# Example

This would have been a **700 line** stack trace with `--show-trace` (or `--full-trace` if you're confused by the other issue).

```
error: infinite recursion encountered

       … while evaluating definitions from `/home/user/h/nixpkgs/pkgs/top-level/all-packages.nix':

       … while evaluating the option `services.postgresql.enable':

       … while evaluating definitions from `/home/user/h/nixpkgs/pkgs/top-level/all-packages.nix':

       … while evaluating the option `services.rabbitmq.enable':

       … while evaluating definitions from `/home/user/h/nixpkgs/nixos/modules/services/amqp/rabbitmq.nix':

       … while evaluating the option `services.epmd.enable':

       … while evaluating definitions from `/home/user/h/nixpkgs/nixos/modules/services/networking/epmd.nix':

       … while evaluating the option `assertions':
```

See the full example in #7552

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

roberth:
We don't have "spammy addErrorContext" in NixOS, but we might want to restrict to a certain number of lines in case some other library does have it.

If spamming does happen in the future, we could add measures similar to what logging libraries use: log levels and/or labeling by module. This is completely hypothetical for now.

edolstra:
Note from the team discussion: it might be nice to be able to associate priorities with messages.

PR #7334, opened 2022-11-22
WIP: Don't reverse stack trace when showing

chaoflow:

PR #7641, opened 2023-01-19
Coerce strings fixups

layus:
Fix for #6204, which was reverted in #7621.

The last commit makes builtins.addErrorTrace a so called "frame" trace, meaning that it is never hidden. If a message was written to guide debugging, we better print it. And it unbreaks the nixpkgs-lib-tests which check for the presence of certain handcrafted traces in the output.

Now, this PR also contains important fixes that were forgotten in the original one for some reason.
https://github.com/NixOS/nix/pull/7641/commits/ca7c5e08c10d3ebd5a491a52c76f29b1dc102375 in particular is pretty big, as apparently I managed to pass arguments on the wrong order to coerceToString, and the type system happily takes a boolean for a string, and vice-versa. This means that the previous version was severely broken with respect to the boolean flags of some coerceToString calls.
Sadly, the diff is quite ugly due to the huge revert commit at the base. Take a chance at investigating individual commits.

- [X] This reverts the revert patch, as it was simpler than merges/rebases.
- [X] This keeps https://github.com/NixOS/nix/pull/7621/commits/620e4fb89bd08fb35e83d18d84f5f635a83239d8
- [x] Local `nix flake check` 

I am still a bit puzzled by the infinite recursion error, but this whole patchset seems to get rid of it. Not sure it's worth investigating further.


edolstra:
> If a message was written to guide debugging, we better print it.

Is this actually true? The same could be said about the trace messages from the evaluator. There could be a lot of spammy debug messages that shouldn't be shown by default.

roberth:
> > If a message was written to guide debugging, we better print it.
> 
> Is this actually true? The same could be said about the trace messages from the evaluator. There could be a lot of spammy debug messages that shouldn't be shown by default.

With `--show-trace`, absolutely, and without it, there's a good case to be made; see #7553.
We don't have spammy `addErrorContext` in NixOS, but we might want to restrict to a certain number of lines in case some other library does have it.

roberth:
@tomberek 

> I’m pretty sure the referenced “infinitive recursion error” is only the error message changes, not that something is now infinite and was not before.

That is correct.

> The extra frame is useful to add back in, but I’d be willing to change the Nixpkgs test regardless because overall the output is far more useful.

Not sure what you mean? I would not be willing to change it because the added error context is crucial.

> The diff is only in “—show-trace” of the module evaluation which is likely inscrutable to nearly all of our users.

Long traces are very unattractive, but anyone who's motivated could figure out which module options are involved in an error.
In the old days it was very easy to see which trace items were about options, because the messages were close together and visually aligned.

> While this PR makes that output much more useful, with or without the frame trace.

Wait, are we solving #7553 and referring to that?

src/libcmd/repl.cc
```diff
@@ -397,7 +397,7 @@ StringSet NixRepl::completePrefix(const std::string & prefix)
             Expr * e = parseString(expr);
             Value v;
             e->eval(*state, *env, v);
-            state->forceAttrs(v, noPos);
+            state->forceAttrs(v, noPos, "nevermind, it is ignored anyway");
```
roberth, line 400:
```suggestion
            state->forceAttrs(v, noPos, ""); // it is ignored anyway
```

tomberek:
Sample output for the same tests as referred to [earlier](https://github.com/NixOS/nix/pull/7621#issuecomment-1387718654). 


```
       … while calling 'optional'

         at /nix/store/zipyafg6vvg329q2qg5dpql7bh6da1f2-source/lib/lists.nix:255:20:

          254|   */
          255|   optional = cond: elem: if cond then [elem] else [];
             |                    ^
          256|

       … while calling anonymous lambda

         at /nix/store/zipyafg6vvg329q2qg5dpql7bh6da1f2-source/lib/modules.nix:481:44:

          480|       context = name: ''while evaluating the module argument `${name}' in "${key}":'';
          481|       extraArgs = builtins.mapAttrs (name: _:
             |                                            ^
          482|         builtins.addErrorContext (context name)

       … while evaluating the module argument `custom' in "/home/tom/nixpkgs/lib/tests/modules/import-custom-arg.nix":

       error: infinite recursion encountered

       at /nix/store/zipyafg6vvg329q2qg5dpql7bh6da1f2-source/lib/modules.nix:483:28:

          482|         builtins.addErrorContext (context name)
          483|           (args.${name} or config._module.args.${name})
          484|       ) (lib.functionArgs f);
```

Of specific note is the the new output: `… while evaluating the module argument `custom' in "/home/tom/nixpkgs/lib/tests/modules/import-custom-arg.nix":` and that the order is a bit more clear - no more having to scroll to the beginning in long module evaluations!

roberth:
> no more having to scroll

That's thanks to #7334. These changes make a great combination.

src/libcmd/repl.cc
layus: I prefer my version, as in the event where this strange string gets printed, and not ignored as expected, users have a chance to find the location where it comes from.

roberth: Then let's just fix the message today. This seems accurate and unique enough.

```suggestion
            state->forceAttrs(v, noPos, "while evaluating an attrset for the purpose of completion");
```

layus:
@edolstra

>>    If a message was written to guide debugging, we better print it.
>
> Is this actually true? The same could be said about the trace messages from the evaluator. There could be a lot of spammy debug messages that shouldn't be shown by default.

There could be indeed. But keep in mind that they are still hidden behind --show-trace. So this change is not about always printing all the addErrorContext messages. It's about always having them in the full trace, and visible with --show-trace. It could get spammy at some point, and we could craft better rules about these, but the old rule was definitely too strict.

src/libcmd/repl.cc
roberth:
```suggestion
            state->forceAttrs(v, noPos, "while evaluating an attrset for the purpose of completion (this error should not be displayed; file an issue?)");
```

src/libutil/error.cc
```diff
@@ -200,13 +200,125 @@ std::ostream & showErrorInfo(std::ostream & out, const ErrorInfo & einfo, bool s
 
     auto noSource = ANSI_ITALIC " (source not available)" ANSI_NORMAL "\n";
 
-    // traces
-    if (showTrace && !einfo.traces.empty()) {
+    /*
+     * Traces
+     * ------
+     *
+     *  The semantics of traces is a bit weird. We have only one option to
```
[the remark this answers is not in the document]
layus: It depends where you stand from I think. Its the semantics of traces that is weird in a lazy & declarative programming language with respect to imperative traces. They do not mean the same thing.
But yes, the implementation of traces here is weird, just because we cannot do it differently given the nix language features.

src/libutil/error.hh
```diff
@@ -86,6 +86,7 @@ void printCodeLines(std::ostream & out,
 struct Trace {
     std::shared_ptr<AbstractPos> pos;
     hintformat hint;
+    bool frame;
```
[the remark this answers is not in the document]
layus: What about "importance", or "scope" ? The thing is, I produce a lot of trace elements about details of the nix expressions. If this is done everywhere, it gets too long to be useful. but if I print details until I reach a broader scope trace (like a function call), I stop printing the details and only print function calls.

All this fuss happened because I considered that addErrorTrace was a local trace, but it makes more sense to have it also in the stack trace. It is extremely difficult to strike the right balance here, and this is just a naive attempt.

Also, frame kind of makes sense, because it corresponds to a trace about a stack frame (well, initially, before adding addErrorTrace). I know technically all of the expressions correspond to calling functions, but I have decided to consider builtins calls as normal expressions, including addErrorTrace. This is just a work of printing lots of expressions and getting them all look good.

To make progress on this, we need a collection of error-generating expressions. Then we can work on comparing them, and finding what is the best info to display. This PR only tries to add context to some coercions, not to fix all the traces at once.

tomberek:
> Now, this PR also contains important fixes that were forgotten in the original one for some reason.
> [ca7c5e0](https://github.com/NixOS/nix/commit/ca7c5e08c10d3ebd5a491a52c76f29b1dc102375) in particular is pretty big, as apparently I managed to pass arguments on the wrong order to coerceToString, and the type system happily takes a boolean for a string, and vice-versa. This means that the previous version was severely broken with respect to the boolean flags of some coerceToString calls.
> Sadly, the diff is quite ugly due to the huge revert commit at the base. Take a chance at investigating individual commits.

@layus I'm doing a review of the "forgotten commit". Can you clarify where we had the incorrect usage of coerceToString and an example of the incorrect behavior? I'm just struggling to characterize exactly what is being fixed.

src/libutil/error.hh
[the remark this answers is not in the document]
roberth: It seems this is going to be tech debt then.

src/libutil/error.hh
```diff
@@ -152,15 +155,22 @@ public:
...
     template<typename... Args>
-    void addTrace(std::shared_ptr<AbstractPos> && e, const std::string & fs, const Args & ... args)
+    void addTrace(std::shared_ptr<AbstractPos> && e, std::string_view fs, const Args & ... args)
     {
-        addTrace(std::move(e), hintfmt(fs, args...));
+        addTrace(std::move(e), hintfmt(std::string(fs), args...));
     }
 
-    void addTrace(std::shared_ptr<AbstractPos> && e, hintformat hint);
+    void addTrace(std::shared_ptr<AbstractPos> && e, hintformat hint, bool frame = false);
```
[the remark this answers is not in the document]
roberth: Probably not, it seems.

src/libexpr/primops.cc
```diff
@@ -801,8 +790,10 @@ static void prim_addErrorContext(EvalState & state, const PosIdx pos, Value * *
         v = *args[1];
     } catch (Error & e) {
         PathSet context;
-        e.addTrace(nullptr, state.coerceToString(pos, *args[0], context,
-                "while evaluating the error message passed to builtins.addErrorContext").toOwned());
+        auto message = state.coerceToString(pos, *args[0], context,
+                "while evaluating the error message passed to builtins.addErrorContext",
+                false, false).toOwned();
+        e.addTrace(nullptr, message);
```
layus: @tomberek: here for example: state.coerceToString(pos, *args[0], context, "while evaluating the error message passed to builtins.addErrorContext"). The string ("while ...") was passed in place of the first bool argument (copyToStore) while it should really be passed after three bools corresponding to the default values of the three parameters.

src/libexpr/primops/fetchTree.cc
```diff
@@ -125,7 +125,7 @@ static void fetchTree(
             if (attr.name == state.sType) continue;
             state.forceValue(*attr.value, attr.pos);
             if (attr.value->type() == nPath || attr.value->type() == nString) {
-                auto s = state.coerceToString(attr.pos, *attr.value, context, false, false, "").toOwned();
+                auto s = state.coerceToString(attr.pos, *attr.value, context, "", false, false).toOwned();
```
layus, line 128: @tomberek: here it was missing one bool, and the empty string was passed in lieu of the third bool.
Overall, it was silly to provide a default value for that error context, which I wanted to be required. So it makes sense to move it before optional arguments.

src/libexpr/eval.cc
```diff
@@ -2154,13 +2159,23 @@ BackedStringView EvalState::coerceToString(const PosIdx pos, Value & v, PathSet
...
+        }
+        return coerceToString(pos, *i->value, context, errorCtx,
+                              coerceMore, copyToStore, canonicalizePath);
     }
 
-    if (v.type() == nExternal)
-        return v.external->coerceToString(positions[pos], context, coerceMore, copyToStore, errorCtx);
+    if (v.type() == nExternal) {
+        try {
+            return v.external->coerceToString(positions[pos], context, coerceMore, copyToStore);
```
layus, line 2173: @tomberek: This one is very special, as I did not want to change the interface of external values. That is the reason why I first introduced the context as an extra, optional argument. Now it's a required argument, and thus to keep the external values interface the same, we just omit it there, catch errors, and add the context ourselves.

tomberek (approved):
From meeting we discussed only merging the revert and fix to addErrorContext. Upon further review I'm more confident in the full PR as it fixes incorrect calls; for example to builtins.addErrorContext which was involved in the original issue. This meant that the intended error context was interpreted as a bool and not used.

[1]: https://github.com/NixOS/nix/pull/7641/commits/ca7c5e08c10d3ebd5a491a52c76f29b1dc102375#diff-0f59bb6f197822ef9f19ceae9624989499d170c84dfdc1f486a8959bb4588cafL804-L805

layus:
@tomberek Thank you for reviewing this one in detail. :heart:

thufschmitt (approved):

thufschmitt:
Agreed with @tomberek. And the “forgotten commit” looks very sensible too. So let's merge

